PR #4070, opened 2026-06-12
fix(sessions): paginate tool-heavy session tails by visible messages

t3chn0pr13st:
## Thinking Path

- Hermes WebUI uses `/api/session?msg_limit=N` for fast session switching and "Load earlier" pagination.
- The user-visible transcript is made of user/assistant rows, while `role=tool` rows are hidden or folded into assistant tool cards.
- The previous backend path did not let `msg_limit` protect the expensive WebUI load: it could still reconstruct the full WebUI lineage before slicing the response window.
- The previous window also counted raw storage rows, so a tool-heavy tail could consume the requested budget with hidden tool rows.
- This PR changes the limited-session path so small API requests stay small server-side and the budget is spent on visible transcript rows, while still including in-window tool rows needed for tool-card context.
- The limited path now keeps the lightweight pre-compression snapshot stitch so compression-continuation sessions can still reveal archived history.
- The result is faster initial loads for large tool-heavy sessions and pagination that advances by visible conversation messages.

## What Changed

- Make `_message_window_for_display(..., msg_limit=N)` count visible user/assistant transcript rows instead of raw storage rows.
- Keep tool rows inside the selected visible-message span, but do not let them consume the visible-message budget.
- Add a limited WebUI message path that avoids the heavier lineage merge for `msg_limit` requests and uses the display sidecar plus only newer `state.db` rows.
- Preserve the lightweight `_webui_sidecar_lineage_messages_for_display()` snapshot stitch on limited WebUI loads so `pre_compression_snapshot` parents remain reachable.
- Bound large hidden tool-result payloads in paginated responses while leaving full unpaginated transcript loads unchanged.
- Preserve top-level non-string `tool.content` types when bounded previews are emitted for limited responses.
- Use the returned message-window offset for `_messages_truncated`, avoiding false "load earlier" signals when raw hidden tool rows exceed `msg_limit` but no older visible transcript rows were omitted.
- Update frontend comments and regression tests for the new `msg_limit` semantics.
- Address Greptile style feedback by removing the dead `_slice` assignment and documenting the compatibility-only `expand_renderable` parameter.

## Why It Matters

In conversations with many tool calls, opening a large session or clicking "Load earlier" should not require repeated requests just to step over hidden tool output. It also should not do full-history WebUI lineage reconstruction when the caller explicitly requested a small `msg_limit` window. This keeps the current fast-switching behavior, reduces limited response payload size, preserves archived pre-compression history, and makes pagination match what the user actually sees in the transcript.

## Verification

[...]

PR #4166, opened 2026-06-14
perf: keep complete sidebar metadata on index fastpath

ai-ag2026:
## Summary

- keep complete indexed lineage/sidebar rows on the `_index.json` fastpath even when the sidecar file mtime is newer than the logical session timestamp
- keep the mtime rescue path for incomplete/legacy rows that lack sidebar counters/timestamps
- add regression coverage for complete snapshot/lineage rows so `/api/sessions` does not fan out into sidecar reads per poll

## Why

Large compressed histories can leave many sidecar files with mtimes newer than their logical `last_message_at`. When the index row already has complete sidebar metadata, mtime alone should not force `Session.load_metadata_only()` for every historical lineage row. That turns routine sidebar refreshes into O(history) JSON sidecar work.

This is intentionally a narrow backend fastpath slice. It does not attempt to solve pagination, response caching, or transcript-tail virtualization.

## Overlap note

I saw related open performance/sidebar work touching nearby files, especially #3881 and #4070. This slice is limited to the index refresh gate in `api/models.py` plus tests.

## Test plan

- `env -u HERMES_WEBUI_PASSWORD TZ=UTC LANG=C.UTF-8 LC_ALL=C.UTF-8 /home/manfred/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_session_index.py -q` -> 43 passed, 1 warning
- `python -m py_compile api/models.py`
- added-line secret scan -> 0 hits

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -2770,13 +2770,26 @@ def _row_may_need_sidecar_metadata_refresh(
         # so only true continuations are eligible.
         if str(session.get('session_source') or '').strip().lower() == 'fork':
             return False
+        if session.get('message_count') is None or session.get('last_message_at') is None:
+            return True
+        # Lineage fields are enriched from state.db in a batched pass later in
+        # all_sessions(). A complete indexed lineage row must not be reloaded
+        # from its sidecar merely because the filesystem mtime is newer than the
+        # logical message timestamp; that pattern is common after compression
+        # and turns each /api/sessions poll into hundreds of JSON prefix scans.
+        # Keep the mtime repair path only for rows whose counters are known bad
+        # or incomplete enough that the index cannot be trusted.
         lineage_shaped = bool(
             session.get('parent_session_id')
             or session.get('_lineage_root_id')
             or session.get('_compression_segment_count')
         )
-        needs_mtime_check = lineage_shaped or (
-            sid and _looks_like_stale_zero_message_row(session)
+        needs_mtime_check = bool(
+            sid
+            and (
+                _looks_like_stale_zero_message_row(session)
+                or (lineage_shaped and session.get('user_message_count') is None)
+            )
         )
         if needs_mtime_check and _sidecar_mtime_after_index_timestamp(session):
             return True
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
## Review — the fastpath is correct, with one precision note on what "enriched later" actually covers

Pulled `ab4066c0` into a read-only worktree and read both gates (`api/models.py:2772-2796` and `:2856-2869`), the surrounding `_row_may_need_sidecar_metadata_refresh` body, and `_enrich_sidebar_lineage_metadata` (L3097). The optimization is sound and the two new tests pin the exact fan-out trigger (`_sidecar_mtime_after_index_timestamp` patched to always-True, asserting `load_metadata_only` is never reached).

### The completeness signal is consistent across both gates

In the lineage gate, the new early return handles the missing-counter case before the mtime branch:

```python
if session.get('message_count') is None or session.get('last_message_at') is None:
    return True
...
needs_mtime_check = bool(
    sid and (
        _looks_like_stale_zero_message_row(session)
        or (lineage_shaped and session.get('user_message_count') is None)
    )
)
```

So by the time `user_message_count is None` is evaluated, `message_count`/`last_message_at` are already known present — making the effective completeness contract "all three non-None," which matches the explicit triple-check in `_stale_snapshot_metadata_refresh_ids` (L2862-2866). Good that the two functions agree on what "complete" means.

### Precision note: "lineage fields enriched from state.db" ≠ counts get refreshed

The comment at L2775 justifies skipping the mtime refresh because "Lineage fields are enriched from state.db in a batched pass later." That's accurate for the lineage *graph* fields — `_enrich_sidebar_lineage_metadata` (L3097) writes `source_tag`/`raw_source`/`session_source`/`display_title` and the lineage linkage, but I confirmed it does **not** touch `message_count`, `user_message_count`, or `last_message_at`. So after this change, the index row's *counts* are authoritative for complete lineage rows — nothing downstream re-derives them.

That's the right call for the case you're targeting (compression rewrites the sidecar → mtime bumps, but logical `last_message_at`/counts don't change → the old mtime check was pure noise-driven fan-out). The one residual: if a sidecar ever genuinely advances past the index *without* the index being rewritten (e.g. a `save(skip_index=True)` path or an interrupted write), the mtime check was the backstop that caught it, and complete rows now skip that backstop. `Session.save()` updates the index by default so this is an anomaly path, not the common case — but it's the behavior change worth one line in the PR body so it's a deliberate trade rather than implicit. The runtime-row guard (`active_stream_id`/pending at L2752-2757) and the zero-message-row rescue both still pre-empt this gate, so live sessions are unaffected.

No defect — the slice is narrow and the tests cover the fan-out regression. Just recommend the comment at L2775 say "counts are trusted from the index; only lineage-graph fields are state.db-enriched" so the next reader doesn't assume the batched pass refreshes counters.

nesquena-hermes:
Shipped in **v0.51.415** (Release OB) 🎉 Thanks @ai-ag2026 — clean perf win. Codex SAFE + Opus net-positive + full suite green. One trivial non-blocking note for a follow-up: a rare `skip_index=True` self-heal collapse can leave a complete lineage row's `message_count` briefly one digit high until the next real save (never affects title/last-message/visibility, self-corrects). If you want zero drift, dropping `skip_index` on that one in-memory self-heal call closes it.

Issue #3986, opened 2026-06-11
Feature: Bidirectional session title sync — WebUI ↔ Hermes state.db (CLI/TUI/Desktop ↔ WebUI)


ddiall:
## Current Behavior (Unidirectional Only)
| Direction | Works? | Mechanism |
|-----------|--------|-----------|
| WebUI rename → `state.db` | ✅ (when `sync_to_insights=true`) | `_sync_session_title_to_insights()` → `sync_session_usage()` → `db.set_session_title()` |
| CLI/TUI/Hermes Desktop rename → WebUI session file | ❌ | No reverse path exists |
| `hermes session rename` / `state.db` change → WebUI sidebar | ❌ | WebUI reads title from local JSON only |

## Impact
- **TUI sessions stuck at "Tui Session" forever** in WebUI sidebar (title comes from `source_label` = "Tui Session" at creation, never updated)
- Renames made in Hermes Desktop / CLI don't reflect in WebUI
- WebUI feels like a second-class citizen for sessions created elsewhere
- Combined with bug #3985 (can't rename in WebUI either), TUI-origin sessions are **completely unmanageable** in WebUI

## Root Cause
`api/state_sync.py:135` `sync_session_usage()` has a `profile` parameter to write to the correct `state.db`, and is called from WebUI mutation endpoints. But there's **no listener or poller** that reads title changes from `state.db` and writes back to WebUI session JSON files (`~/.hermes/webui/sessions/{sid}.json`).

The sync was designed for WebUI → agent (usage tracking), not agent → WebUI.

### Option A: WebUI as Source of Truth
Make all title mutations go through WebUI session JSON; have `state_sync` read from there on agent startup.
- *Pros*: Single writer, simpler mental model
- *Cons*: Breaks CLI-first workflows; agent would need WebUI running to persist titles

### Option B: Bidirectional Sync (Recommended for Current Architecture)
Add a listener in `state_sync.py` or a lightweight cron job that polls `state.db` title changes and writes back to WebUI session files.
- *Pros*: Works with current architecture; CLI/TUI/Desktop remain independent
- *Cons*: Two writers → potential conflicts; needs conflict resolution (last-write-wins with timestamp?)

### Option C: Unified Store (Ideal Long-term)
Deprecate WebUI session JSON files entirely; always read/write titles via `state.db` (requires profile-aware `state.db` resolution at read time).
- *Pros*: Single source of truth; no sync needed
[...]

PR #4167, opened 2026-06-14
fix: keep session sidebar visible on slow loads

ai-ag2026:
## Summary

- keep the session sidebar from going blank when `/api/sessions` times out or fails during a large-history/cold-load refresh
- preserve the last known/stale session list and show a visible retryable error state instead of only logging `console.warn`
- decouple optional project metadata loading from the session list refresh
- add opt-in retry controls to the frontend `api()` helper for the initial session-list boot load
- return compact JSON and a bounded sidebar row shape from `/api/sessions`, excluding heavy detail-only fields from list rows

## Why

A large active/compressed session can make sidebar refreshes slow enough for the frontend timeout to fire. Before this change, `renderSessionList()` could swallow the failure and leave the sidebar looking empty even though sessions still existed. The user-visible failure mode was "no sessions" rather than "session list refresh failed; retry".

This is a UX/API-payload resilience slice. It does not introduce backend caching or pagination; those are separate performance tracks.

## Overlap note

There are many open PRs touching `api/routes.py` and `static/sessions.js`. The most relevant nearby work appears to be #3881 and #4070. This PR is scoped to visible failure/stale-state handling, decoupled project loading, compact JSON, and bounded sidebar payload fields.

## Test plan

- `env -u HERMES_WEBUI_PASSWORD TZ=UTC LANG=C.UTF-8 LC_ALL=C.UTF-8 /home/manfred/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_session_sidebar_resilience.py tests/test_issue1611_session_profile_filtering.py tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata -q` -> 17 passed, 1 warning
- `python -m py_compile api/helpers.py api/routes.py`
- added-line secret scan -> 0 hits

### DIFF
--- a/api/helpers.py
+++ b/api/helpers.py
@@ -161,13 +161,26 @@ def _safe_write(handler, body: bytes) -> None:
         )
 
 
-def j(handler, payload, status: int=200, extra_headers: dict=None) -> None:
+def _json_response_body(payload, *, pretty: bool = True) -> bytes:
+    """Serialize API JSON responses.
+
+    Sidebar/session endpoints can return thousands of rows on large installs.
+    Pretty-printing large list responses inflates both CPU and wire bytes. Keep
+    the public helper default stable for existing tests/callers; hot paths can
+    opt into compact JSON with ``pretty=False``.
+    """
+    if pretty:
+        return _json.dumps(payload, ensure_ascii=False, indent=2).encode('utf-8')
+    return _json.dumps(payload, ensure_ascii=False, separators=(',', ':')).encode('utf-8')
+
+
+def j(handler, payload, status: int=200, extra_headers: dict=None, *, pretty: bool = True) -> None:
     """Send a JSON response.
 
     *extra_headers*: optional dict of additional headers to include
     (e.g., {'Set-Cookie': '...'}).  Headers are sent before end_headers().
     """
-    body = _json.dumps(payload, ensure_ascii=False, indent=2).encode('utf-8')
+    body = _json_response_body(payload, pretty=pretty)
     handler.send_response(status)
     handler.send_header('Content-Type', 'application/json; charset=utf-8')
 
--- a/api/routes.py
+++ b/api/routes.py
@@ -4777,6 +4777,8 @@ def _session_attention_summary(session_id: str) -> dict | None:
 _SIDEBAR_SESSION_RESPONSE_FIELDS = {
     "session_id",
     "title",
+    "display_title",
+    "_state_db_title",
     "workspace",
     "model",
     "model_provider",
@@ -4811,6 +4813,7 @@ _SIDEBAR_SESSION_RESPONSE_FIELDS = {
     "pending_started_at",
     "default_hidden",
     "worktree_path",
+    "worktree_branch",
     "parent_session_id",
     "parent_title",
     "parent_source",
--- a/api/routes.py
+++ b/api/routes.py
@@ -1616,10 +1616,7 @@ def _session_list_payload_to_response(payload: dict) -> dict:
     safe_merged = []
     runtime_rows = _session_list_cache_overlay_runtime_rows(payload.get("sessions", []) or [])
     for s in runtime_rows:
-        item = dict(s) if isinstance(s, dict) else {}
-        if isinstance(item.get("title"), str):
-            item["title"] = _redact_text(item["title"])
-        item["attention"] = _session_attention_summary(str(item.get("session_id") or ""))
+        item = _sidebar_session_response_item(s) if isinstance(s, dict) else {}
         safe_merged.append(item)
     return {
         "sessions": safe_merged,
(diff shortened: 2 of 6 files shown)


nesquena-hermes:
## Review — the resilience half is great; the bounded-payload allowlist drops three fields the sidebar actually reads

Pulled `f97b4d80` into a read-only worktree and read the full diff across `api/routes.py`, `static/sessions.js`, and the new tests, then cross-checked the allowlist against what `compact()` emits (`api/models.py:914-987`) and what `sessions.js` consumes. The frontend resilience work is solid — stale-list preservation via `renderSessionListFromCache()`, the visible retry note, decoupled project loading so a `/api/projects` failure can't blank the sidebar, and the longer boot timeout are all the right moves. The concern is `_SIDEBAR_SESSION_RESPONSE_FIELDS` (`api/routes.py:4152-4201`): switching `/api/sessions` from a full-dict passthrough to a fixed allowlist silently drops fields the sidebar reads today.

### 1. `display_title` / `_state_db_title` dropped → breaks CLI/TUI title sync in the sidebar

`_enrich_sidebar_lineage_metadata` writes both onto the row when state.db has a better title than a generic WebUI one (`api/models.py:3104-3105`):

```python
session['_state_db_title'] = state_db_title
session['display_title'] = state_db_title
```

`sessions.js` renders off exactly those, with `title` only as last resort (L4620, also L4714):

```js
const rawTitle=String((s&&(s.display_title||s._state_db_title||s.title))||'Untitled').trim();
```

Neither field is in the allowlist, so after this PR a CLI/TUI/Telegram session whose real title lives in state.db falls back to the generic `title` in the sidebar — a regression in the #3986/#3987 title-sync area.

### 2. `pending_user_message` dropped → idle/busy detection degrades

`compact()` emits both the raw text and the boolean (`api/models.py:976-977`):

```python
'pending_user_message': self.pending_user_message,
'has_pending_user_message': has_pending_user_message,
```

The allowlist keeps `has_pending_user_message` but not `pending_user_message`. `sessions.js` reads the raw-text field in ~7 places for the idle/streaming decision, e.g. `_isSessionIdle` at L356 and the runtime-confirmation gate at L3295:

```js
const hasRuntimeConfirmation=Boolean(local.active_stream_id||local.pending_user_message||local.pending_started_at);
```

Some of these read the *local* merged copy, but L52 / L3278 / L4879 read off the server payload `s`. Either migrate those call sites to `has_pending_user_message` (cleaner — the raw text shouldn't ship to the list anyway, which is the point of your test `..._drops_bulky_detail_fields`), or add `pending_user_message` back. Dropping it without touching the readers is the silent-break path.

### 3. `worktree_branch` dropped → worktree indicator tooltip loses the branch name

`compact()` emits it (`api/models.py:968`); `worktree_path` is in the allowlist but `worktree_branch` isn't. `sessions.js:5377`:

```js
wtInd.title=`${wtLabel}: ${s.worktree_branch||s.worktree_path}`;
```

Degrades gracefully to the path, but the branch label is the intended display — worth adding to the set.

### Cleared as non-issues

I checked the other `s.*` reads: `_lineage_key`, `_child_sessions`, `_child_session_count`, `_lineage_segments` are all client-derived (assigned in `sessions.js` at L4556-4563), and `lineage_root_id` (non-underscore) is read defensively next to `_lineage_root_id`, which *is* allowlisted — so none of those regress.

### Recommendation

The allowlist approach is the right way to bound the payload, but it needs a guard so a field the frontend reads can't be dropped without a test catching it. Concretely: add `display_title`, `_state_db_title`, and `worktree_branch` to the set, and decide whether to keep `pending_user_message` or migrate its readers to the boolean. A cheap regression: assert the allowlist is a superset of a hardcoded list of frontend-consumed keys (grep-derived), so the next field added to `sessions.js` fails the test until it's allowlisted. The existing `..._drops_bulky_detail_fields` test only pins what's *excluded*; it can't catch an over-aggressive drop of something still in use.

Happy to re-review once the set is reconciled — the resilience changes are good to go as-is.

nesquena-hermes:
Strong direction — keeping the sidebar from going blank on a slow/failed `/api/sessions` and surfacing a retryable error state is a real resilience win, and I verified the risky parts are clean: the `api()` retry is opt-in (no existing caller passes the new options), retry doesn't touch non-idempotent calls, and the 401 `login?next=` redirect path is preserved. Three **SILENT regressions** need fixing before it can merge, all stemming from the new `_SIDEBAR_SESSION_RESPONSE_FIELDS` allowlist dropping fields the frontend relies on:

**1. `read_only` is dropped → read-only sessions look writable (api/routes.py ~4152).**
`Session.compact()` (api/models.py:983) and Claude Code rows (api/models.py:3682) emit `read_only`, and the sidebar uses it at `static/sessions.js:1408/6043/6076` to suppress rename / action-menu / swipe affordances. With it dropped from the allowlist, read-only imported and Claude Code sessions render with editable affordances. **Fix:** add `read_only` to the allowlist + a test asserting `_sidebar_session_response_item()` preserves it.

**2. `gateway_routing` / `gateway_routing_history` dropped → sidebar model labels silently degrade (api/routes.py ~4152).**
`/api/sessions` previously included these (api/models.py:960-961); the sidebar consumes them at `static/sessions.js:195/5638` for detailed model display. **Fix:** preserve bounded latest routing metadata (prefer `gateway_routing`), or update the frontend to consume an equivalent bounded field.

**3. Failed refresh after a profile switch renders the previous profile's cached sessions (static/sessions.js ~3594).**
`switchToProfile()` updates `S.activeProfile` (static/panels.js:5565) then awaits `renderSessionList()`. On an `/api/sessions` failure the catch path re-renders the unchanged `_allSessions` with no active-profile scope check — so a failed refresh right after a profile switch can show the prior profile's sessions. This is the cross-profile-stale case worth getting exactly right. **Fix:** track the cache scope from `sessData.active_profile` / `all_profiles` and, when the requested scope differs, render only the retryable error state (or clear/filter) rather than preserving stale rows.

Findings 1 and 2 are mechanical (allowlist + tests); 3 is the one that needs a small design touch (scope-tagging the cache). Since the allowlist is a new mechanism that silently determines what reaches the client, it's better for you to own which fields it must carry than for me to patch it blind — please add `read_only` + the routing field, scope the stale-preservation to the active profile, and add a sidebar-response test covering all three. Happy to re-gate right after.

nesquena-hermes:
Thanks for the re-push @ai-ag2026 — you added `display_title`, `_state_db_title`, and `worktree_branch` to the allowlist (good — those were the secondary items) and added a resilience test file. But the **3 SILENT regressions that were the actual blockers** are still open on this head. I verified against the current `_SIDEBAR_SESSION_RESPONSE_FIELDS` (routes.py:4897) — the row is built strictly from the allowlist (`if key in _SIDEBAR_SESSION_RESPONSE_FIELDS`), so a field not in the set is dropped:

1. ❌ **`read_only` still dropped → read-only sessions render writable.** `read_only` is not in the allowlist. `Session.compact()` (models.py:983) + Claude Code rows (models.py:3682) emit it, and the sidebar reads it at `sessions.js:1408/6043/6076` to suppress rename / action-menu / swipe. Without it, read-only imported + Claude Code sessions get editable affordances. **Fix:** add `"read_only"` to the allowlist + a test asserting `_sidebar_session_response_item()` preserves it.

2. ❌ **`gateway_routing` / `gateway_routing_history` still dropped → sidebar model labels degrade.** Neither is in the allowlist; the sidebar consumes them at `sessions.js:195/5638` for detailed model display. **Fix:** preserve a bounded `gateway_routing` (latest routing only — you don't need the full history), or migrate the frontend reader to an equivalent bounded field.

3. ❌ **Failed refresh after a profile switch still renders the previous profile's sessions.** The `catch` path in `renderSessionList()` re-renders the unchanged `_allSessions` with no active-profile scope check, so a failed `/api/sessions` right after `switchToProfile()` shows the prior profile's rows. **Fix:** scope-tag the cache from `sessData.active_profile`/`all_profiles` and, when the requested scope differs, render only the retryable error state rather than preserving stale rows.

1 and 2 are mechanical (allowlist + a sidebar-response test); 3 needs the small cache-scope-tagging touch. The resilience direction + the `api()` opt-in retry are all good to go — it's specifically these 3 payload/scope items that have to land first, because shipping #1 means read-only sessions silently become writable in the sidebar. Ping me when pushed and I'll re-gate (Codex + Opus + full suite). 🙏

nesquena-hermes:
Shipped in **v0.51.462** (Release PW — "sidebar survives slow loads"). 🚀

Thanks @ai-ag2026 — the resilient-load direction and the opt-in `api()` retry were exactly right, and the sidebar now stays visible with a retryable error state instead of blanking on a slow `/api/sessions`.

We resolved the 3 review-blocking SILENT regressions on our side and shipped through the full gate (your branch hadn't been re-pushed since the last review, so we picked them up to keep this moving):

1. **`read_only` / `is_read_only` restored to the sidebar payload allowlist** — read-only sessions (imported CLI + Claude Code) keep their rename / action-menu / swipe suppression instead of rendering writable.
2. **`gateway_routing` restored** — the detailed sidebar model label renders again. We deliberately did NOT add the unbounded `gateway_routing_history` to the list payload: `_latestGatewayRoutingForSession` prefers the latest `gateway_routing` and only falls back to history for an active session, so sending the full history per row would just bloat the response.
3. **Failed refresh after a profile switch no longer shows the prior profile's sessions** — the session cache is now scope-tagged (`active_profile` + all-profiles flag), and the catch path clears stale rows + renders the error state when the requested scope differs from the cached scope.

Added `tests/test_4167_sidebar_payload_and_scope.py` covering all three.

**Gate:** full suite 9233 passed; Codex SAFE TO SHIP; Opus ship (all three fixes verified line-level, scope-tag null-safe both sides, retry bounded to boot path); browser-smoke green (local + CI). Attribution preserved via Co-authored-by. Closing as merged.

Issue #3926, opened 2026-06-10
RFC: Stable Assistant Turn Anchors for Live-to-Final rendering


franksong2702:
## Problem

#3400 clarified the Live-to-Final product model, and #3820 accepted Transparent Stream as an opt-in chronological activity display mode. The remaining gap is architectural: live streaming, settled transcript rendering, replay/reload/recovery, Compact Worklog, and Transparent Stream still need one stable frontend owner for a single assistant turn.

Today, live state is assembled from SSE callbacks, stream-local state, `INFLIGHT`, live DOM, run journal replay, and settled `S.messages`. Recent fixes around recovered blank turns, mid-stream `renderMessages()` rebuilds, stream-end recovery, and Transparent Stream all point at the same seam: the UI lacks a stable assistant-turn presentation/reconciliation anchor.

## Proposal

Add an RFC for **Assistant Turn Anchor** as a frontend presentation/reconciliation primitive.

The RFC defines:

- source-of-truth layers across server transcript, run journal, SSE, anchor, `INFLIGHT`, DOM, and RuntimeAdapter
- anchor creation after `/api/chat/start` succeeds and returns `stream_id`
- identity aligned to the #1925 Event Envelope (`event_id`, `run_id + seq`) instead of inventing a new scheme
- a shared normalization seam with Transparent Stream (`normalizeToEvents()` / `anchor.activity_events`)
- settlement where `done` updates the existing turn instead of creating a separate settled-only turn
- replay/reload/reconstruction behavior
- terminal state, artifact, side-effect, and metadata ownership rules
- rollout phases and implementation review gates

## Scope Boundaries

This is an RFC/tracking issue, not an implementation request.

Out of scope:

- implementing Transparent Stream (#3820)
- defining Queue / Steer / Stop-and-send semantics (#3058 / pending-intent controls)
- implementing RuntimeAdapter / runner migration (#1925)
- tool-card visual redesign
- transcript virtualization
- backend schema migration

## Related

- Parent product model: #3400
- Transparent Stream: #3820
- RuntimeAdapter contract: #1925
- Pending-intent controls: #3058

franksong2702:
Phase 0 implementation PR opened: #3962.

Scope is intentionally non-visual and inert: it lands the current state-layer inventory, source event classification, dedupe precedence, doc contract, and Phase 0 tests without wiring anchors into live streaming, replay, settlement, `INFLIGHT`, or `renderMessages()` yet.

Issue #3400, opened 2026-06-02
Redesign live-to-final assistant replies for running agent sessions


franksong2702:
## Summary

Hermes WebUI should redesign the live-to-final assistant reply experience for running agent sessions.

This is one of the most important UX surfaces in Hermes WebUI: when an agent is running, the user needs to understand what is happening now, what has already happened, and where the final answer begins. The current experience has repeatedly exposed edge cases where live progress, tool activity, replay/recovery, auto-compression, and final transcript rendering compete for the same visual surface.

The goal is not just to add a Worklog widget. The goal is to make the assistant reply lifecycle coherent:

- During a running turn, process text should remain the primary timeline.
- Tool activity should be visible, quiet, and close to the prose that caused it.
- When the turn settles, implementation detail should collapse into a compact activity summary above the final answer.
- The final answer must remain readable, complete, and clearly separate from the worklog.
- Recovery, session switching, and replay should reconstruct the same structure rather than creating a different UI state.

## Why this matters

Running agent sessions are the core Hermes WebUI experience. Users spend most of their time watching an agent reason, inspect files, run tools, recover from reconnects, and eventually produce a final answer. If this surface is noisy or unstable, the product feels unreliable even when the backend is working correctly.

Mature agent clients such as Codex and Claude Code generally treat intermediate tool/progress details as supporting activity rather than as the final answer itself. They make running state visible, but they do not force every internal lifecycle marker to remain in the final transcript. Hermes WebUI should follow the same product direction while preserving its own stronger browser affordances: replay, session switching, resumability, and structured tool details.

## Historical context

This issue intentionally cross-references prior Hermes WebUI problems because this UX has been revisited many times from different angles:

- Early working/final-answer separation: #536 established the direction that working output and final answer should not blur together.
- Manual and automatic compression UX: #469, #619, #699, #1142, #1316, #2355, #2357, #2973, and #3079 show repeated attempts to represent compression state without confusing the user.
- Live/replay/reconnect durability: #765, #1525, #1533, #1584, #2283, #2341, #2342, #2344, #2347, #2924, #2925, #3005, #3371, and #3391 cover the long-running theme that a browser client must be able to recover the running transcript from durable stream state.
- Activity/tool/thinking/progress rendering: #1298, #3014, and #3015 cover the need for visible progress while avoiding noisy thinking/tool surfaces.
[...]

PR #4172, opened 2026-06-14
Stable Assistant Turn Anchors: guard Compact Worklog handoff

franksong2702:
## Stacked Ship Gate

This PR is ready for review, but it should not be merged/released by itself.

Please review this PR as Slice 9, then ship it together with the Slice 10 companion PR: #4177.

- Slice 9 / this PR: guarded Compact Worklog handoff from the retained live anchor registry.
- Slice 10 / #4177: durable Compact Worklog rebuild for hard refresh, session switch, and registry-loss cases.

Reason: landing this PR alone improves the same-page live-to-settled path, but still leaves hard refresh / reload / session switch on the legacy rebuild behavior. That creates an inconsistent user-visible window. Preferred merge plan: review #4172 and #4177 separately, then merge back-to-back. If maintainers prefer one atomic release diff, fold #4177 into this PR before merge.

---

## Thinking Path
- This is part of #3926, under the #3400 live-to-final consistency track.
- The previous anchor slices established the inert anchor owner, activity scene projection, live shadow feed, dual-run reconciliation, and the renderer snapshot adapter. Those slices made `activity_scene_v1` available, but they did not let it own Compact Worklog rendering.
- This slice is the first guarded Compact Worklog handoff: settled assistant messages stamped with `_anchor_stream_id` may render their Worklog rows from the retained anchor activity scene.
- The handoff is intentionally narrow. The anchor scene only owns the settled Compact Worklog when it proves complete tool ownership for that assistant turn; otherwise the existing `S.toolCalls` / `assistantThinking` settled rebuild remains the fallback.
- The next slice should use the durable transcript / replay path to rebuild the same activity ownership after reload or session switch. This PR does not claim durable reload/replay ownership yet.

## What Changed
- Added a Compact Worklog renderer handoff for assistant-turn anchor activity scenes in `renderMessages()`.
- Reuses the existing Worklog row/card helpers instead of adding a new visual component family.
- Bypasses the session HTML cache only for anchor-stamped assistant messages so scene projection can run during settled rendering.
- Added ownership guards so incomplete, running, reasoning-only, or tool-count-incomplete scenes cannot suppress the legacy settled Worklog fallback.
- Hardened the retained-registry session guard to fail closed when an anchor scene lacks session identity or belongs to a different session.
- Fixed restore-path settlement rendering by clearing `S.busy` before the settled render pass, so recovered settled sessions can rebuild message-derived Worklog rows.
- Updated architecture docs and changelog to record the Compact Worklog handoff boundary and the fallback guard.

## Why It Matters
- This is the first user-visible step where the anchor can preserve live-to-final Worklog continuity instead of letting settlement discard and rebuild the activity trace from a weaker source.
- It directly addresses the class of bugs where final settlement can reorder, drop, or temporarily hide Worklog rows while the final answer appears.
- It keeps the rollout reversible: if the anchor scene is not complete enough or does not prove same-session ownership, the old settled rendering path still wins.

## Contract Routing
- Contract owner: Stable Assistant Turn Anchor / live-to-final rendering consistency (#3926, parent #3400).
- Contract document updated: `docs/architecture/stable-assistant-turn-anchor-phase0.md`.
- Intentional contract change: `activity_scene_v1` is no longer fully inert for rendering; it may feed the settled Compact Worklog path only through the guarded handoff added here.
- Out of scope: Transparent Stream ownership, durable reload/replay rebuild, backend RuntimeAdapter migration, and removal of legacy `S.toolCalls` fallback.

## Verification
- `node --check static/ui.js`
- `node --check static/assistant_turn_anchors.js`
- `git diff --check`
- `python -m pytest tests/test_stable_assistant_turn_anchor_registry.py -q`
- `python -m pytest tests/test_issue3401_deep_review_fixes.py tests/test_stable_assistant_turn_anchor_phase0.py tests/test_live_activity_timeline.py tests/test_ui_tool_call_cleanup.py tests/test_stream_end_recovery_gating.py tests/test_stable_assistant_turn_anchor_normalizer.py tests/test_issue3820_chat_activity_display_mode.py tests/test_issue1690_scroll_completion.py -q`
- Local result: 26 passed + 124 passed.
- GitHub Actions result after the latest push: browser-smoke, lint, and all Python 3.11 / 3.12 / 3.13 test shards passed.

## Risks / Follow-ups
- This PR intentionally relies on the retained live anchor registry for the settled handoff. Reload/session-switch durable rebuild remains the next slice.
- Visual surface is the existing Compact Worklog renderer; no new row/card styling is introduced here. A human 8787 visual pass is still useful before marking the draft ready if reviewers want UI evidence attached.
- Legacy `S.toolCalls` and `assistantThinking` stay in place as fallback until the durable reload/replay slice proves equivalent coverage.

## Model Used
- GPT-5 Codex assisted with implementation, tests, documentation, and PR preparation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,14 @@
 
 - **Long final answers no longer snap back to the bottom while you are reading them in PWA/mobile mode (#4123).** Same-session background refreshes now treat a preserved scroll snapshot whose viewport is clearly away from the bottom (>250px) as an active reading position and restore it before any follow-to-bottom path runs, re-mark restored mid-answer positions as unpinned so later refreshes keep respecting the reader, and no longer reset the scroll-direction tracker on same-session force refreshes (that reset stays scoped to real session switches). (#4123)
 
+### Changed
+
+- **Stable Assistant Turn Anchors Compact Worklog handoff (#3926).** Settled Compact Worklog rendering can now consume the live anchor registry's `activity_scene_v1` for assistant messages stamped with `_anchor_stream_id`, coalescing tool start/complete events into the existing Worklog tool-card UI and skipping the legacy `S.toolCalls` / `assistantThinking` rebuild only when the scene proves it owns a complete settled tool set for that turn. The handoff still backfills same-turn settled reasoning ahead of tool rows when the live scene lacks it, bypasses the session HTML cache for anchor-stamped turns so session switching re-runs scene reconciliation, and missing registries, mismatched sessions, reasoning-only scenes, incomplete/running tool scenes, non-Compact modes, Transparent Stream, and projection errors all fall back to the existing renderer.
+
+### Fixed
+
+- **Restored settled streams rebuild Worklog tool rows before final render (#3926).** The stream-end/reconnect restore path now clears the active pane busy state before calling `renderMessages()`, so the settled message-level tool metadata fallback can rebuild Compact Worklog tool cards instead of rendering a restored final turn with only Thinking detail.
+
 ## [v0.51.389] — 2026-06-13 — Release NB (surface dirty-install state in update check, #4085)
 
 ### Fixed
--- a/docs/architecture/stable-assistant-turn-anchor-phase0.md
+++ b/docs/architecture/stable-assistant-turn-anchor-phase0.md
@@ -3,8 +3,9 @@
 This inventory implements the first non-visual slice of
 [`stable-assistant-turn-anchors.md`](../rfcs/stable-assistant-turn-anchors.md).
 It documents the current per-turn state layers and the event-shape contract that
-future anchor phases must consume. It does not claim that anchors are wired into
-streaming or rendering yet.
+later anchor phases must consume. Early slices were inert; this file now tracks
+which narrow rendering handoffs have started and which paths still fall back to
+the existing renderer.
 
 ## RFC Phase Progress
 
@@ -41,6 +42,12 @@ streaming or rendering yet.
   Transparent Stream row hooks and feed them through the reconciler to produce a
   concrete matched / mismatched answer. The adapter remains opt-in and is not
   invoked by `renderMessages()` or the live SSE hot path.
+- Slice 9 starts the Compact Worklog renderer handoff: settled assistant
+  messages stamped with `_anchor_stream_id` can read the retained live anchor
+  registry's `activity_scene_v1`, render reasoning/tool rows through the current
+  Worklog DOM helpers, and skip the legacy rebuild for that turn. Transparent
+  Stream, reload-after-registry-expiry, and any missing/mismatched registry still
+  use the existing renderer.
 
 ## State Layers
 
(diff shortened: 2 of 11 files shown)


franksong2702:
Quick status update: I converted this PR back to Draft while we finish the Slice 10 companion.

The intent is to keep Slice 9 and Slice 10 separate for review clarity:

- Slice 9: guards the Compact Worklog handoff from the retained live anchor registry.
- Slice 10: adds the durable rebuild path for hard refresh / session switch / registry-loss cases.

From a product/user-experience standpoint, we would like them to ship together. Landing Slice 9 alone improves the same-page live-to-settled path, but still leaves the hard-refresh/reload path on the legacy rebuild behavior, which creates an inconsistent UX window. The preferred merge plan is review separately, then merge back-to-back once Slice 10 is ready.

franksong2702:
Slice 10 companion PR is now open: #4177.

Keeping this PR in Draft for now so the two slices can be reviewed separately but shipped together. #4177 is stacked on this branch and covers the durable rebuild path for hard refresh / session switch / registry-loss cases.

franksong2702:
Update after moving this PR back to ready-for-review: the current ship gate is now written at the top of the PR body.

This PR is ready to review as Slice 9, but please do not merge/release it by itself. The intended plan is to review #4172 and #4177 separately, then ship them together/back-to-back so the retained-live handoff and durable reload/session-switch rebuild land as one user-visible fix.

franksong2702:
Added a small follow-up fix to this Slice 9 branch after local 8787 validation found a real live-session-switch gap.

Root cause: when switching back to an active stream during a Thinking-only phase, `loadSession()` treated the current live turn as empty because the preserved-content check only recognized `.tool-card-row` / `.wl-reason`, not live Thinking rows. The destructive fallback could then rebuild a blank Worklog shell and hide the Thinking Card.

Fix: count `.agent-activity-thinking[data-live-thinking="1"]` as existing live Worklog content before running the fallback. Added a regression assertion in `tests/test_inflight_stream_reuse.py`.

Local verification:

- `git diff --check`
- targeted live session-switch regressions: `10 passed`
- Anchor/Worklog shard on #4172 branch: `55 passed`

Also synced the same fix into #4177 so the stacked companion PR does not regress this once #4172 merges.

franksong2702:
Closing this PR intentionally. After live validation on the 8787 aggregate, this implementation is not a safe candidate for merge.

The problem is architectural, not just a small UI regression: this PR still bridges Compact Worklog through legacy rendering state (`S.messages`, `S.toolCalls`, `INFLIGHT`, DOM snapshots, and `renderMessages`) with anchor-derived helpers layered in. That does not deliver the intended Assistant Turn Anchor handoff where the anchor activity scene / renderer snapshot owns the current assistant turn ordering and Compact Worklog consumes that owned scene.

Observed validation showed unstable session-switch / live-to-settled behavior, including live Thinking / Worklog instability and ordering surprises. Continuing to patch this PR would make the legacy bridge more complex instead of using the previously built anchor infrastructure correctly.

We will replace this with a new slice that treats Compact Worklog as an anchor-scene consumer: existing legacy sources may feed the anchor pipeline, but they should not independently decide Worklog order or placement. #4177 should not ship on top of this implementation either.

PR #4176, opened 2026-06-14
ci: add Trivy scan, coverage gate, and branch protection docs

twchichi:
## Summary

- Add Trivy Docker image vulnerability scanner to `release.yml` (blocks on CRITICAL/HIGH)
- Add Trivy filesystem/misconfig scan to `release.yml`
- Add `pytest-cov` to test pipeline and enforce 80% coverage threshold
- Document branch protection requirements and required status checks in `CONTRIBUTING.md`

## Test plan

- [ ] Verify Trivy scan step runs on next release tag
- [ ] Verify test suite fails if coverage drops below 80%
- [ ] Confirm `CONTRIBUTING.md` branch protection table is accurate

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,3 +55,25 @@ jobs:
           build-args: HERMES_VERSION=${{ github.ref_name }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+      # ─── Docker image security scan (trivy) ──────────────────────────────
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: ghcr.io/${{ github.repository }}:${{ steps.meta.outputs.version }}
+          format: 'table'
+          exit-code: '1'
+          severity: 'CRITICAL,HIGH'
+          timeout: '10m'
+        if: steps.meta.outputs.version != ''
+
+      - name: Run Trivy on Dockerfile (config scan)
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: 'fs'
+          scan-ref: .
+          scanners: 'misconfig'
+          format: 'table'
+          exit-code: '0'
+          severity: 'HIGH,MEDIUM'
+          timeout: '5m'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -92,6 +92,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install "pyyaml>=6.0" pytest pytest-timeout pytest-asyncio pytest-shard
+          pip install pytest-cov coverage
           # ruff is installed so tests/test_ruff_forward_lint.py runs its E9/F821
           # tree-clean assertions in-suite (mirrors how eslint is available for
           # tests/test_static_js_runtime_lint.py). If install fails the test
@@ -106,4 +107,4 @@ jobs:
           pip install mcp || echo "mcp install failed — test_mcp_server.py will importorskip"
 
       - name: Run tests (shard ${{ matrix.shard }} of 3)
-        run: pytest tests/ -v --timeout=60 --shard-id=${{ matrix.shard }} --num-shards=3
+        run: pytest tests/ -v --timeout=60 --shard-id=${{ matrix.shard }} --num-shards=3 --cov=. --cov-report=xml --cov-fail-under=80
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -203,3 +203,23 @@ Want the smoothest review?
 - [SPRINTS.md](SPRINTS.md)
 
 Questions are best raised early, before a large change is finished.
+
+## Branch Protection
+
+The `master` branch is protected to enforce quality gates. Required checks:
+
+| Check | Workflow / Job | Purpose |
+|-------|---------------|---------|
+| `Tests / lint` | `.github/workflows/tests.yml` | Ruff Python lint + ESLint JS runtime guards |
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
## Confirming the three findings + a concrete fix for the coverage gate and the docs table

Read the full diff plus `tests.yml` and `release.yml` on the PR branch and cross-checked the actual workflow tree on `origin/master`. Greptile's three findings all reproduce. Adding the concrete remediation and one extra docs inaccuracy below.

### 1. Coverage gate breaks every matrix job (confirmed) — and here's why per-shard fails

`tests.yml:110` runs each job with both a shard *and* a whole-tree coverage gate:

```yaml
run: pytest tests/ -v --timeout=60 --shard-id=${{ matrix.shard }} --num-shards=3 --cov=. --cov-report=xml --cov-fail-under=80
```

The matrix is `python-version: ['3.11','3.12','3.13'] × shard: [0,1,2]` (`tests.yml:68,76`), so each job runs only ~1/3 of the suite but measures coverage against the whole tree (`--cov=.`). Each shard will report roughly a third of real coverage and trip `--cov-fail-under=80` immediately — all 9 jobs red on the next push.

Two ways to fix it:

- **Combine, don't gate per shard.** Drop `--cov-fail-under` from the matrix, upload each shard's `coverage.xml` as an artifact, then add a separate `coverage` job that depends on the matrix, downloads all 9, runs `coverage combine` + `coverage report --fail-under=80` once. That's the only way an 80% number is meaningful across a sharded suite.
- **Or measure coverage in a single non-sharded job.** Add one job that runs `pytest tests/ --cov=. --cov-fail-under=80` without `--shard-id`, and leave the existing 9 matrix jobs as pure pass/fail. Simpler, slower.

Either keeps the speed benefit of sharding for pass/fail while making the coverage gate correct.

### 2. Trivy `@master` pin + scan-after-push (confirmed)

`release.yml:61,71` use `aquasecurity/trivy-action@master`, executing whatever's on that third-party branch with this job's `contents: write` / `packages: write`. Pin to an immutable commit SHA. And the image scan at L60-68 runs *after* `build-push-action` (L47) has already pushed to GHCR, so a CRITICAL-CVE image is public before `exit-code: 1` fires. Build locally with `push: false`, scan, then push on pass — or scan in a gate job that blocks the push step. (`steps.meta.outputs.version` itself is fine — `id: meta` is defined at L38.)

### 3. CONTRIBUTING table is wrong in *both* directions

Greptile caught the 4 phantom files. Checking `origin/master:.github/workflows/`, the actual set is:

```
browser-smoke.yml  docker-smoke.yml  native-windows-startup.yml  release.yml  tests.yml
```

So the new table (`CONTRIBUTING.md:209-217`) references four files that don't exist — `dep-audit.yml`, `security-scan.yml`, `lint-configs.yml`, `feature-checks.yml` — **and** omits two that do: `docker-smoke.yml` and `native-windows-startup.yml`. As written it documents a branch-protection policy the repo can't enforce, which is worse than no table. Either add the four workflows in this PR (then the lint/test/browser-smoke rows are accurate) or rewrite the table to match the five workflows that actually exist today.

### Recommendation

The Trivy scan and a coverage gate are both worth having, but as-is this turns the matrix red and ships a misleading policy doc. Splitting it would help review: (a) coverage as a combine-job, (b) Trivy pinned-SHA + pre-push gate, (c) docs that match reality. Happy to re-review once the coverage gate is restructured.

nesquena-hermes:
Thanks for thinking about CI hardening — Trivy image scanning is a welcome direction. But this PR can't merge as-is: it would brick the release pipeline and the docs describe checks that don't exist. Specifics:

**1. The 80% coverage gate would fail every PR and every release tag (blocking).**
`--cov-fail-under=80` is appended to the per-shard test command, but each shard runs only **1/3 of the suite** (`--num-shards=3`). Coverage is measured per-shard, so each shard sees roughly a third of the codebase exercised → it will report well under 80% and fail, on every PR and every `v*` tag. To enforce a real total-coverage threshold you'd have to collect `coverage` data per shard (e.g. `--cov-report=` to emit `.coverage.<shard>`), upload the fragments as artifacts, and run a separate `coverage combine` + `coverage report --fail-under=80` job after all shards finish. A per-shard `--cov-fail-under` can't express a whole-suite threshold.

Separately, please treat the *value* of a hard 80% gate as an open question for the maintainer rather than something to wire in by default — measure current total coverage first; an advisory (non-blocking) report is the safer first step.

**2. `release.yml` Trivy step blocks every release on any base-image CVE (blocking).**
`exit-code: '1'` with `severity: 'CRITICAL,HIGH'` on the image scan means a single new HIGH CVE in an upstream base-image layer (which you don't control and which appears constantly) fails the tagged release build and stops the Docker push. For a scanner you don't gate merges on yet, run it non-blocking first (`exit-code: '0'`, upload the report) so you can see signal without wedging releases. The filesystem/misconfig scan you already set to `exit-code: '0'` — match that for the image scan.

**3. The CONTRIBUTING.md branch-protection table lists workflows that don't exist.**
It references `dep-audit.yml`, `security-scan.yml`, `lint-configs.yml`, and `feature-checks.yml` — none of those exist in `.github/workflows/` (only `browser-smoke.yml`, `docker-smoke.yml`, `native-windows-startup.yml`, `release.yml`, `tests.yml` are present). Documenting required status checks that don't exist will mislead contributors and can't be enforced. Please document only the checks that actually run.

Suggested path forward: split this into (a) the Trivy scans, non-blocking to start, and (b) a CONTRIBUTING note that matches the real workflow set — and drop the hard coverage gate (or make it an advisory combined-coverage report) until there's a maintainer decision on the threshold. Happy to re-review once it's scoped down.

PR #3990, opened 2026-06-11
feat: add WebUI server restart control

theseussss:
## Summary
- Add a Settings → System control to restart the local Hermes WebUI server without fully stopping it.
- Reuse the existing process re-exec helper and block restarts while chat streams are active.
- Keep the existing Stop server control and shutdown description visible, with restart-specific i18n text.

## Test Plan
- `pytest tests/test_server_restart_endpoint.py tests/test_server_restart_ui.py -q`
- `python -m py_compile api/routes.py`
- Manual validation on a local systemd-managed WebUI instance: restart button reloaded the server and the page recovered after a brief reconnect window.

## Notes
- The restart flow briefly displays a reconnect message and tells users to refresh if they see a transient 502/connection error during process restart.
- `/api/restart` remains behind normal WebUI auth/CSRF handling and returns 409 when active chat streams are running.

PR #4180, opened 2026-06-14
feat: add opt-in guarded server restart control

ai-ag2026:
## Summary

Adds an opt-in Settings → Server controls restart action for Hermes WebUI instances that can safely expose a restart command.

This is intentionally supervisor-agnostic and does not assume systemd, sudo, local service names, or local paths. Operators enable the control by setting `HERMES_WEBUI_RESTART_COMMAND`; the browser only receives semantic availability/status fields and never sees the raw configured command.

This overlaps the general area of #3990, but takes a narrower guarded/status-oriented approach:

- opt-in only via configuration
- no built-in supervisor or re-exec assumption
- waits for currently active WebUI work before starting
- persists restart status for polling
- reports semantic states instead of localized backend messages
- keeps Stop server as a separate destructive action

## Behavior

New restart statuses:

- `idle`
- `waiting_idle`
- `restarting`
- `checking_health`
- `succeeded`
- `failed`
- `unavailable`

The server never marks a restart as `succeeded` just because the configured command exited with `0`. A successful command moves to `checking_health`; `succeeded` is only promoted when a freshly booted WebUI process observes a previous in-progress restart record. If no replacement process is observed, status transitions to `failed` with `restart_not_observed`.

## UI / i18n

- Renames the Settings shutdown block to generic `Server controls`.
- Adds a managed `Restart server` button and status line.
- Keeps `Stop server` as the destructive red control.
- Adds i18n keys for all visible restart/server-control text across locale bundles.

## API

Adds:

- `GET /api/server/restart/config`
- `GET /api/server/restart/status?id=...`
- `POST /api/server/restart`

The POST remains behind the existing CSRF guard.

## Safety notes

- No default restart command is provided.
- No raw command is exposed to the client.
- No `systemctl`, service name, sudo, or local path assumptions are introduced.
- In-process single-flight prevents duplicate restart workers in the common single-WebUI-process deployment.

Known non-blocking limitation: the single-flight lock is process-local. If multiple WebUI Python processes share a state dir and accept restart POSTs concurrently, a cross-process file lock would be needed to make that global.

## Verification

Passed locally:

```text
python3 -m pytest -o addopts='' tests/test_server_restart_control.py tests/test_issue4002_shutdown_i18n_code_spans.py -q
# 19 passed

python3 -m py_compile api/routes.py api/server_restart.py
node --check static/boot.js
node --check static/panels.js
node --check static/i18n.js
git diff --check
ruff check api/server_restart.py tests/test_server_restart_control.py
```

Additional local notes:

```text
# reported no changed .py files; direct ruff check was run on the new Python files above

python3 tests/browser_smoke.py
# skipped locally: playwright not installed

npm run lint:runtime
# not run locally: eslint not installed
```


nesquena-hermes:
## Review — the guarded-restart state machine is solid; one correction to the "unauthenticated GET" framing + an overlap call

Reading `api/server_restart.py:1-262` and the route wiring at `api/routes.py:5927-5932` / `7398-7400` on the PR head, the core design is careful in the right ways. The thing I want to highlight is the honesty of the success contract: a `returncode == 0` does **not** promote to `succeeded`. `_run_restart_command` moves to `checking_health` (`server_restart.py:178-181`), and `succeeded` only happens when a *different* process reads a pre-boot record:

```python
if payload.get("status") in {"restarting", "checking_health"} and started_at and started_at < _BOOT_TIME:
    payload["status"] = "succeeded"
```

`_BOOT_TIME` is captured at module import (`server_restart.py:25`), so a process that issued a restart can never promote its own in-progress record — only a freshly-imported successor whose `_BOOT_TIME` is newer than the record's `started_at` can. `test_read_restart_status_only_promotes_status_from_previous_process` pins both directions (same-process `checking_health` stays, previous-process flips to `succeeded`). If no successor ever observes it, the 120s `_HEALTH_CONFIRM_TIMEOUT_SECONDS` path marks it `failed` with `restart_not_observed`. That's the correct semantics for "I can't prove my own replacement."

### Correction on the auth posture (re: the bot summary)
The review summary says the GET `config`/`status` endpoints "follow the existing unauthenticated health-endpoint pattern." That's not actually what happens, and the truth is the *safer* outcome. `/health` is unauthenticated because it's explicitly listed in `PUBLIC_PATHS` (`api/auth.py:50-56`):

```python
PUBLIC_PATHS = frozenset({
    '/login', '/health', '/favicon.ico', '/sw.js',
    ...
})
```

`/api/server/restart/config` and `/api/server/restart/status` are **not** in that set, and `check_auth` (`api/auth.py:570`) only short-circuits for `PUBLIC_PATHS` / `/static/` prefixes. So when auth is enabled these GETs return 401 like any other `/api/` route — they are authenticated, not public. Good: even though they only leak `enabled`/status semantics, gating them behind the session is the right default. Worth correcting the record so nobody "fixes" them into PUBLIC_PATHS by mistake. The POST is CSRF-protected too — `/api/server/restart` is absent from `_csrf_exempt_path` (`routes.py:1904-1909`) and the global guard at `routes.py:7391` runs before the handler at `7398`; `test_restart_post_is_csrf_protected` confirms it.

### Single-flight caveat is real and you already flagged it
`_RESTART_LOCK` is a module-level `threading.Lock`, so the single-flight guard (`start_guarded_restart`, `server_restart.py:227-231`) only dedupes within one process. Your PR body already calls this out as the cross-process gap; agreed it's a fine follow-up since the common deployment is one WebUI process per state dir. The status file is written atomically (`_atomic_write` via `os.replace`, `server_restart.py:91-94`), so concurrent readers never see a torn record even if two processes did race.

### Overlap with #3990 (maintainer call)
This deliberately overlaps #3990 ("feat: add WebUI server restart control") but takes the opposite stance: opt-in via `HERMES_WEBUI_RESTART_COMMAND`, supervisor-agnostic, no re-exec assumption, semantic status states instead of localized backend strings. #3990 is still open (last touched 2026-06-13). These two shouldn't both merge — they'd collide on the Settings → Server controls block and the restart route surface. Flagging for a maintainer decision on which approach to take rather than gating either unilaterally; the guarded/status-oriented shape here is the more conservative one for instances that can't safely assume systemd.

## Verdict
Mechanically sound and well-tested (19 cases incl. atomic-write/no-temp-file, idle-wait, stale-`waiting_idle`→terminal, single-flight reuse). No correctness blockers. The only thing needing a human is the #3990 overlap — pick one before merge.

nesquena-hermes (changes requested):
Re-gated the current head (`70b6b0fde394`) with a full security pass (Codex + Opus + complete suite, 9855 passed). **The opt-in approach is the right call** and the security core is solid — the configured command never reaches the browser, it's env-only (`HERMES_WEBUI_RESTART_COMMAND`, `shlex`-split, never shell-interpolated), auth+CSRF gated, and the status-file path is safe. The Settings → System placement is exactly right. Three things to address before it can ship:

### 🔴 Active-work block has a TOCTOU race (the real one)

`_wait_until_idle` (`api/server_restart.py:186-199`) returns the instant it samples `active_runs == 0 && active_streams == 0`, then `_restart_worker` runs the restart command — but **nothing fences a *new* turn from starting in that window.** `/api/chat/start` (routes.py:~15502, ~15757) creates new streams with no restart-pending check, so a send/background/btw turn that begins between the idle snapshot and the exec gets killed mid-turn.

**Fix:** add a process-wide `restart_draining` flag, set once the status enters `{waiting_idle, restarting}`, protected by the same run/start synchronization. Have chat/background/btw `start_session_turn` reject (or defer) new starts while draining, and only exec the restart after the flag is set AND both registries are confirmed empty. (Opus rates the practical severity lower since restart is operator-initiated, but the fence is the correct fix and keeps an in-flight turn from dying.)

### 🟡 Disabled endpoint should be hard-disabled (404)

`POST /api/server/restart` is always routed and returns `200 {"ok": false, "status": "unavailable"}` when `HERMES_WEBUI_RESTART_COMMAND` is unset (`api/server_restart.py:226-229`). Prefer a hard **404** when unconfigured: check `restart_config()` at the top of the route and return 404 before `start_guarded_restart()`. Keeps the disabled feature truly absent rather than a present-but-refusing endpoint.

### 🟡 Hide the control when unconfigured

`static/index.html:1389` / `static/boot.js:2344` render the restart description/button/status unconditionally and only *disable* the button after `/api/server/restart/config` resolves. Wrap just the restart description/button/status in a `restartServerBlock` that's hidden by default and shown only when `cfg.enabled === true` — so an instance that hasn't configured a restart command shows no restart affordance at all. (The existing shutdown control stays visible.)

The security model and the opt-in design are good — this is about closing the new-work fence and making the unconfigured state fully absent rather than present-but-disabled. Happy to re-gate as soon as these land. Thanks @ai-ag2026.

nesquena-hermes:
Salvaging into a fresh PR — thanks @ai-ag2026. We confirmed there's no existing way to restart the WebUI server from the UI today (master only restarts the gateway + tells users to restart Hermes manually), and your supervisor-agnostic approach is the one we want: #3990 was already closed as superseded precisely because its `os.execv` self-re-exec fights a process supervisor, whereas this runs an operator-configured `HERMES_WEBUI_RESTART_COMMAND` and works across systemd/docker/unsupervised. The maintainer review found it mechanically sound and well-tested (19 cases, no correctness blockers), with the right honest-success contract (a process can't promote its own restart).

The branch had gone stale, so we're porting it fresh on current master with your authorship preserved, fully gating it, and capturing a Settings → Server screenshot for the maintainer's sign-off (it's a visible + sensitive surface). I'll link the new gate-ready PR here. Appreciate the careful, portable design.

PR #4184, opened 2026-06-14
Fix regenerate-title for CLI/TUI sessions that lack sidecar files

### LINKED ISSUE #4183: Regenerate title fails for unmaterialized CLI/TUI sessions
Title regeneration returns "Session not found" for CLI/TUI sessions that haven't been materialized as sidecar JSON files.

This is a gap in #3987.

## Root cause

`/api/session/title/regenerate` (routes.py:7753) calls `get_session(sid)`, which only looks up sessions with an existing sidecar JSON in the sessions directory. CLI/TUI sessions that have never been opened or sent a message in the WebUI exist only in state.db — the sidebar computes their display title on the fly as `f'{source.title()} Session'` (models.py:3851) when the real title is null.

Other mutation endpoints (rename, archive, move) use `_get_or_materialize_session(sid)`, which falls back to materializing a writable sidecar from state.db metadata. The regenerate endpoint doesn't use this helper, so it 404s immediately.

## Repro

1. Have CLI/TUI sessions visible in the sidebar (shown as "Tui Session", "CLI Session", etc.)
2. Right-click → Regenerate Title
3. Toast: "Regenerate title failed: Session not found"

## Fix

Replace `get_session(sid)` + `_ensure_full_session_before_mutation(sid, s)` at line 7753-7754 with `_get_or_materialize_session(sid)`, and catch `PermissionError` alongside `KeyError` to handle read-only sessions gracefully (return 403 instead of 404).

rodboev:
## Summary

- The `/api/session/title/regenerate` endpoint used `get_session(sid)` which only finds sessions with materialized sidecar JSON files. CLI/TUI sessions that exist only in `state.db` (with null titles and no sidecar) raised `KeyError`, returning "Session not found".
- Replaced with `_get_or_materialize_session(sid)` which falls back to materializing a writable sidecar from `state.db`, matching the pattern used by rename, archive, and other mutation endpoints.
- Added `PermissionError` catch for read-only imported sessions, consistent with sibling handlers.

Closes #4183

## Test plan

- [x] New test `test_issue4183_regenerate_materialize.py` asserts the handler uses `_get_or_materialize_session` and catches `PermissionError`
- [x] Existing `test_session_title_regenerate_controls.py` suite (7 tests) still passes
- [x] `test_issue3225_rename_sync.py` still passes

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -7750,11 +7750,10 @@ def handle_post(handler, parsed) -> bool:
         sid = body["session_id"]
         prefer_latest = bool(body.get("prefer_latest", False))
         try:
-            s = get_session(sid)
-            s = _ensure_full_session_before_mutation(sid, s)
+            s = _get_or_materialize_session(sid)
         except KeyError:
             return bad(handler, "Session not found", 404)
-        if getattr(s, "read_only", False):
+        except PermissionError:
             return bad(handler, "Read-only imported sessions cannot be renamed", 403)
         next_title, reason, raw_preview = generate_session_title_for_session(s, prefer_latest=prefer_latest)
         if not next_title:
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.412** (Release NY) 🎉 Thanks @rodboev — clean fix using the canonical `_get_or_materialize_session` mutation pattern. Codex SAFE + Opus net-positive + full suite green.

PR #4186, opened 2026-06-14
Fix misleading error in regenerate-title endpoint

rodboev:
## Summary

- The PermissionError message in the regenerate-title endpoint said "cannot be renamed" (copy-pasted from the rename handler). Changed to "cannot regenerate titles" to match the actual action.

Follow-up to #4184 / #4183.

## Test plan

- [x] Updated assertion in `test_session_title_regenerate_controls.py`
- [x] All 28 related tests pass

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -7754,7 +7754,7 @@ def handle_post(handler, parsed) -> bool:
         except KeyError:
             return bad(handler, "Session not found", 404)
         except PermissionError:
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.414** (Release OA) 🎉 Thanks @rodboev — clean follow-up. (This is exactly the copy fix I held back from inlining into #4184 to keep that PR surgical; nice to have it standalone with the test.)

PR #4188, opened 2026-06-14
[codex] Make request logging nonfatal

mintymac:
## Summary

Make request/access logging non-fatal during HTTP response handling. `Handler.log_request()` and request error logging now route through a small guarded print helper so a closed or redirected process-wide stdout/stderr stream cannot abort a response before headers are sent.

## Root Cause

In a threaded WebUI process, agent/tool code can temporarily redirect global `sys.stdout`/`sys.stderr`. If another thread restores one of those globals to a stream that has since been closed, `print(..., flush=True)` raises `ValueError: I/O operation on closed file`. `BaseHTTPRequestHandler.send_response()` calls `log_request()` before writing response headers, so a poisoned stdout can turn otherwise healthy endpoints, including `/health`, into empty TCP replies.

## Impact

This hardens the WebUI against a failure mode observed in production where `/health` accepted connections but returned an empty reply, causing Docker to mark the container unhealthy. Normal request handling no longer depends on stdout/stderr remaining writable.

## Validation

- `python -m pytest tests/test_broken_pipe_cascade.py`

Note: direct whole-file ruff over `server.py` still reports pre-existing findings outside this change; the repo's diff-scoped gate reports no new violations.

### DIFF
--- a/server.py
+++ b/server.py
@@ -299,6 +299,17 @@ class Handler(BaseHTTPRequestHandler):
 
     def log_message(self, fmt, *args): pass  # suppress default Apache-style log
 
+    @staticmethod
+    def _safe_webui_print(message: str) -> None:
+        """Emit a request log line without letting logging break responses."""
+        try:
+            print(message, flush=True)
+        except Exception:
+            # Agent/tool code can redirect or close process-wide stdout/stderr
+            # in another thread. HTTP response handling must not depend on
+            # those global streams remaining writable.
+            pass
+
     def log_request(self, code: str='-', size: str='-') -> None:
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in v0.51.431 (Release OR) via #4235. Thanks @mintymac — request logging is now non-fatal, so a closed/redirected stdout stream can't abort an HTTP response before headers are sent. Codex + Opus + full suite (9060) all green; verified only log I/O is swallowed (real handler errors still 500), Ctrl+C/shutdown still propagate, and your 2 tests genuinely fail-on-master/pass-on-PR. Closing as shipped.

PR #4192, opened 2026-06-14
fix: keep settled stream message counts consistent

ai-ag2026:
## Thinking Path

A local WebUI incident showed the browser replacing a correct live transcript after stream completion. The old local runtime had two contributing problems: an unsafe state.db tail fastpath and a settled `done` payload whose embedded full transcript had stale compact metadata (`len(messages)=453`, `message_count=45`). Current `origin/master` already loads/merges sidecar truth for `/api/session?messages=1`, so this PR fixes the remaining upstream inconsistency in settled SSE payloads.

## What Changed

- Added `_session_payload_with_full_messages()` in `api/streaming.py`.
- Normal `done` payloads now embed full messages with `message_count = len(messages)`.
- Settled `apperror` payloads use the same helper so they cannot leak stale compact counts either.
- Added regression coverage for the helper and both callsites.
- Updated the existing Sprint 42 source-order test to look for the new settled payload marker instead of the old `raw_session = s.compact()` assignment.
- Added a changelog entry.

## Why It Matters

`Session.compact()` can legitimately carry metadata-only counts from index/sidebar loads. That metadata is fine for compact rows, but it is wrong for SSE payloads that also include the full `messages` array. Completion/reconcile code relies on these counts for viewed-state and message-window decisions, so a stale count can make a complete settled transcript look like a short stale window.

This keeps the payload internally consistent: if the server sends full messages, the reported count matches those messages.

## Verification

Local:

- `git diff --check`
- `python3 -m pytest tests/test_streaming_done_payload_message_count.py tests/test_run_journal_routes.py tests/test_1694_terminal_cleanup_ownership.py -q` → 26 passed
- `python3 -m pytest tests/test_sprint42.py::test_streaming_persists_reasoning_in_session tests/test_streaming_done_payload_message_count.py -q` → 4 passed
- `python3 -m pytest tests/test_run_journal_routes.py tests/test_1694_terminal_cleanup_ownership.py -q` → 23 passed
- `python3 -m py_compile api/streaming.py api/routes.py`
- Added-line risk scan over commit range → 0 findings
- Independent reviewer pass requested callsite coverage; added tests now assert both `done` and `apperror` callsites use the helper.

CI note:

- Initial CI after rebase exposed a stale source assertion in `tests/test_sprint42.py` looking for `raw_session = s.compact()`. This PR now updates that assertion to the new helper call and re-pushed.

## Risks / Follow-ups

- Low risk: the helper preserves existing compact metadata and tool-call fields, only overriding `message_count` when a full message array is embedded.
- This PR does not change `/api/session` route behavior. Current `master` already avoids the older unsafe state.db-tail-only path by merging sidecar lineage and state.db for message loads.

## Contract Routing

Task type: runtime/session consistency bugfix  
Touched areas: streaming settled SSE payloads, session metadata consistency  
Relevant docs: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, runtime/session consistency contracts  
Evidence: regression test plus adjacent stream/session route tests listed above

## Model Used

OpenAI GPT-5.5 via Hermes Agent, with local terminal/file tooling and an independent reviewer subagent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Streaming completion payloads now report a message count that matches the embedded transcript.** Settled `done`/error SSE payloads no longer reuse stale compact metadata when they include the full `messages` array, preventing completion/reconcile code from mistaking a full transcript for a short stale window.
+
 ## [v0.51.415] — 2026-06-14 — Release OB (keep complete sidebar metadata on the index fastpath, #4166)
 
 ### Fixed
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -53,6 +53,25 @@ from api.models import (
 )
 from api.session_ops import mark_session_title_generated, session_has_manual_title
 
+
+def _session_payload_with_full_messages(session, *, tool_calls=None):
+    """Return compact session metadata plus the embedded full transcript.
+
+    ``Session.compact()`` may intentionally use metadata-only counts from an
+    index/sidebar load. A settled SSE payload that embeds ``session.messages``
+    must report the count of that embedded transcript, otherwise completion and
+    reconcile paths can mistake a complete payload for a stale short window.
+    """
+    messages = list(getattr(session, 'messages', None) or [])
+    raw = session.compact() | {
+        'messages': messages,
+        'message_count': len(messages),
+    }
+    if tool_calls is not None:
+        raw['tool_calls'] = tool_calls
+    return raw
+
+
 # Global lock for os.environ writes. Per-session locks (_agent_lock) prevent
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Shipped in **v0.51.419** (Release OF) 🎉 Thanks @ai-ag2026 — clean fix for the settled-payload stale-count/reconcile bug. I rebased it onto current master (21 behind) and, per the deep gate, extended the same `_session_payload_with_full_messages()` helper to the gateway-routed chat `done` event (gateway_chat.py:511), which shared the bug, with a regression test. Codex SAFE + Opus net-positive + full suite 8987. (Low-priority follow-up noted: the 6 sync-HTTP `s.compact() | {messages}` sites in routes.py could use the helper too for consistency — not the SSE done path, so out of scope here.)

PR #4205, opened 2026-06-14
fix: block empty active pending session overwrite

ai-ag2026:
## Summary

Prevents a stale active/pending WebUI session snapshot with `messages=[]` from overwriting an existing messageful session sidecar and refreshing the session index to a shorter/empty transcript.

This is the sibling failure class observed after the settled-stream message-count fix from #4192: a completed/recovered transcript can exist on disk, but an older active pending `Session` object can later save `messages=[]` over it. The existing `.json.bak` safeguard preserves a copy, but the primary sidecar and `_index.json` still degrade, so the browser can continue to show missing/empty transcript state.

## Contract Routing

Task type: runtime/session durability bug fix

Touched areas:
- `api/models.py` session sidecar persistence
- regression coverage for active/pending save invariants

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`

State layers affected:
- durable WebUI session sidecar
- sidebar/session index metadata
- pending turn runtime metadata

Invariant: runtime maintenance or a stale active/pending object must not make a messageful transcript shorter by writing `messages=[]` over it.

## Tests

```bash
python3 -m pytest tests/test_session_save_empty_pending_guard.py tests/test_stale_stream_writeback.py -q
# 9 passed in 2.39s

python3 -m py_compile api/models.py
```

Live productive checkout smoke also passed with the streaming sibling tests:

```bash
python3 -m pytest tests/test_session_save_empty_pending_guard.py tests/test_stale_stream_writeback.py tests/test_sprint42.py::test_streaming_persists_reasoning_in_session tests/test_streaming_done_payload_message_count.py -q
# 13 passed in 2.66s

python3 -m py_compile api/models.py api/streaming.py
```

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -787,6 +787,20 @@ class Session:
                 except (json.JSONDecodeError, ValueError):
                     existing_msg_count = -1  # corrupt → always back up
                 incoming_msg_count = len(self.messages or [])
+                if (
+                    existing_msg_count > 0
+                    and incoming_msg_count == 0
+                    and (self.active_stream_id or self.pending_user_message)
+                ):
+                    logger.warning(
+                        "refusing to overwrite session %s messages with empty active/pending snapshot "
+                        "(existing=%s, incoming=%s, stream=%s)",
+                        self.session_id,
+                        existing_msg_count,
+                        incoming_msg_count,
+                        self.active_stream_id,
+                    )
+                    return
                 if existing_msg_count > incoming_msg_count:
                     bak_path = self.path.with_suffix('.json.bak')
                     # SHOULD-FIX #2 (Opus): atomic write via tmp+replace,
--- a/tests/test_session_save_empty_pending_guard.py
+++ b/tests/test_session_save_empty_pending_guard.py
@@ -1,0 +1,44 @@
+"""Regression coverage for empty active/pending session save writebacks."""
+
+import json
+
+import api.config as config
+import api.models as models
+from api.models import Session
+
+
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
I traced the refresh behavior separately from this sidecar-durability guard.

I agree this PR can prevent a stale empty pending `Session` object from degrading the primary sidecar/index, but that is not the reason the currently-open WebUI transcript fails to refresh after an external/session event.

The client-side refresh gap is:

1. #4195/#3916 changed `refreshActiveSessionIfExternallyUpdated()` to return early for every non-external/WebUI-native session.
2. `_scheduleSessionEventsRefresh()` only called `refreshSessionList(...)`, which updates the sidebar list but does not refresh the active transcript.
3. So a WebUI-native active session can receive a `sessions_changed` event, update its sidebar row, and still leave the open transcript stale.

I opened #4208 for that narrower client-side fix: keep the 30s `poll` guard for WebUI-native sessions, but allow real event/focus/visibility refresh triggers to run the existing count/timestamp divergence check and reload the active transcript when needed.

So I’d treat this PR as a separate data-loss prevention layer, not the explanation for the refresh failure itself.

nesquena-hermes:
Shipped in v0.51.425 (Release OL) via #4219. Thanks @ai-ag2026 — the empty-pending session-save guard closes a real data-loss window (Session.save() now skips overwriting an in-flight session's transcript with an empty list). Codex + Opus + full suite (9027) all green; verified no false-positive on new/cancelled sessions and no write-path bypass. Closing as shipped.

Issue #3916, opened 2026-06-10
Hermes WebUI 30s refresh poll fires unconditionally for non-external sessions


tonyle123:
Title: WebUI 30s refresh poll fires unconditionally for non-external sessions

Description:
The ActiveSessionExternalRefreshPoll runs every 30 seconds for ALL sessions, including pure WebUI sessions. The refreshActiveSessionIfExternallyUpdated() function compares remote message_count against local count, and any discrepancy (e.g. 5 messages from a session import) triggers loadSession(force:true) — a full chat re-render.

Root cause: Missing _isExternalSession() guard in refreshActiveSessionIfExternallyUpdated() (static/sessions.js line ~3440).

Fix: Add one line after the S.busy check:

javascript复制
if(!_isExternalSession(S.session)) return;
External sessions (CLI, Telegram, Discord imports) continue to work normally. WebUI sessions return early.

nesquena-hermes:
Reading `static/sessions.js:3440-3469` (`refreshActiveSessionIfExternallyUpdated`), `3471-3486` (`ensureActiveSessionExternalRefreshPoll`), `1259-1261` (`_isExternalSession`), and the `sessions_changed` SSE handler at `3650-3670` on `origin/master`, plus the introducing commit `467ef33a feat(webui): reconcile external session updates` — the observation is half right, but the proposed one-line fix would regress the exact scenario this poll was built for. Worth laying out before anyone applies it.

## What actually fires every 30s

Confirmed: the timer at `3471-3475` calls the function unconditionally, with no source-type guard:

```js
const _activeSessionExternalRefreshMs = 30000;        // 3398
_activeSessionExternalRefreshTimer = setInterval(() => {
  void refreshActiveSessionIfExternallyUpdated('poll');
}, _activeSessionExternalRefreshMs);                  // 3473-3475
```

But what fires unconditionally is only a **cheap metadata probe** — `GET /api/session?...&messages=0&resolve_model=0` (3454). The expensive `loadSession(sid,{force:true})` re-render is gated behind a divergence check:

```js
if(remoteCount > localCount || remoteLast > localLast){
  await loadSession(sid, {force:true, externalRefreshReason:reason||'poll'});  // 3460-3461
}
```

[...]

Issue #3545, opened 2026-06-04
Streaming scroll position drifts to no-mans-land after response completes when user scrolled up mid-stream


weidzhou:
## Description

In long sessions with streaming responses, scrolling up (mouse wheel) to read history during streaming causes the viewport to land in a "no man's land" after the response completes — neither showing the history position nor the new reply.

## Root Cause

Two interacting issues:

### 1. Sticky-unpin prevents scroll-to-bottom on response completion

When the user scrolls up during streaming, `_scrollPinned` is set to `false` (sticky-unpin, #3343). When the stream finishes, `_finishDone` calls `renderMessages({preserveScroll:true})`, which restores the captured `scrollTop` from before the DOM rebuild. Since `_scrollPinned` is `false`, `scrollIfPinned()` is skipped. Then `shouldFollowOnDone` checks `_isMessagePaneNearBottom(1200)` — which returns `false` because the user scrolled up — so `scrollToBottom()` is never called.

### 2. DOM growth during streaming shifts the meaning of `scrollTop`

While the response streams, new content is appended to the bottom of the messages container, increasing `scrollHeight`. The user's `scrollTop` value (an absolute pixel offset) does not change. But because the DOM grew below, the **relative position** of that `scrollTop` drifts upward — the user was reading message #15 when they stopped scrolling, but by the time streaming finishes, the same `scrollTop` now points to message #8 or earlier.

When `renderMessages({preserveScroll:true})` rebuilds the entire DOM and restores the captured `scrollTop`, the resulting viewport is neither the history position the user was reading nor the new reply at the bottom. It lands in a random middle slice of the conversation.

## Reproduction

1. Start a session with enough history to scroll through (30+ messages)
2. Send a message that produces a long streaming response (e.g., "write a 500-word analysis")
3. While streaming is active, use the mouse wheel to scroll up and read older messages
4. Wait for the response to complete
5. Observe: the viewport jumps to a position that is neither the history you were reading nor the new message

## Expected Behavior

The user explicitly sent a new message, indicating they want to see the result. Even if they scrolled up during streaming to check history, the scroll should follow to the new reply once the response completes.

## Suggested Fix

[...]

PR #4208, opened 2026-06-15
fix(sessions): refresh active WebUI sessions on session events

nesquena-hermes:
## Summary

Fixes the refresh gap introduced by the #3916/#4195 external-refresh guard: the 30-second poll should still skip WebUI-native sessions, but real `sessions_changed` events must be allowed to reconcile the active WebUI transcript when another client/process updates it.

Root cause: `refreshActiveSessionIfExternallyUpdated()` returned early for every non-external session, and `_scheduleSessionEventsRefresh()` only refreshed the sidebar list. So a WebUI-native active session could receive a sidebar/session event, update the list, and still leave the currently-open transcript stale.

## Changes

- Keep the #3916 behavior for the periodic `poll` path only.
- Allow non-poll triggers (`event`, focus/visibility recovery callers) to run the existing count/timestamp divergence check for WebUI-native sessions.
- Make the session-events debounce call `refreshSessionList(..., {refreshActive:true})` so a `sessions_changed` event can refresh the currently-open transcript.
- Update the #3916 regression test to lock this narrower contract instead of the over-broad source guard.

## Tests

```bash
python3 -m pytest tests/test_issue3916_external_refresh_poll.py tests/test_issue4154_sidebar_issue_numbers.py -q
# 7 passed, 1 warning

git diff --check

python3 -m pytest tests/test_stale_stream_cleanup.py tests/test_streaming_done_payload_message_count.py -q
# 15 passed, 1 warning

python3 -m pytest tests/test_gateway_sync.py::test_empty_active_gateway_session_does_not_hide_messaging_history -q
# 1 passed, 1 warning
```

## Notes

This is separate from the data-durability guard proposed in #4205. That PR may still be useful to prevent a stale empty pending object from degrading the sidecar, but the reason the active browser view is not refreshing is the client-side event path above.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -3650,7 +3650,12 @@ async function refreshActiveSessionIfExternallyUpdated(reason){
   if(_activeSessionExternalRefreshInFlight) return;
   if(!S.session || !S.session.session_id) return;
   if(S.busy || S.activeStreamId) return;
-  if(!_isExternalSession(S.session)) return;
+  // #3916: the 30s timer is only a fallback for imported/external sessions.
+  // WebUI-native sessions should not keep probing forever when the sidebar SSE
+  // is healthy, but they still must reconcile when an actual sessions_changed
+  // event, focus, or visibility recovery says another client/process mutated
+  // the active transcript (#4205 follow-up shape).
+  if((reason||'poll')==='poll' && !_isExternalSession(S.session)) return;
   // Cooldown: don't force-reload immediately after streaming ends — the
   // "done" event already delivered the final messages. Reloading here would
   // clear S.toolCalls and lose Activity.
@@ -3719,7 +3724,7 @@ function _scheduleSessionEventsRefresh(reason){
   if(_sessionEventsRefreshTimer) return;
   _sessionEventsRefreshTimer = setTimeout(() => {
     _sessionEventsRefreshTimer = 0;
-    void refreshSessionList(reason||'event');
+    void refreshSessionList(reason||'event', {refreshActive:true});
   }, 300);
 }
 
--- a/tests/test_issue3916_external_refresh_poll.py
+++ b/tests/test_issue3916_external_refresh_poll.py
@@ -4,21 +4,34 @@ from pathlib import Path
 SESSIONS_JS = Path(__file__).resolve().parent.parent / "static" / "sessions.js"
 
 
+def _function_body(src: str, name: str) -> str:
+    m = re.search(
+        rf"(?:async\s+)?function\s+{re.escape(name)}\b.*?^}}",
+        src,
+        re.DOTALL | re.MULTILINE,
+    )
+    assert m, f"{name} function not found"
+    return m.group(0)
+
+
 def test_refreshActiveSessionIfExternallyUpdated_exists():
     src = SESSIONS_JS.read_text(encoding="utf-8")
     assert "async function refreshActiveSessionIfExternallyUpdated" in src
 
 
-def test_isExternalSession_guard_present():
+def test_poll_path_skips_non_external_sessions():
     src = SESSIONS_JS.read_text(encoding="utf-8")
-    m = re.search(
-        r"async function refreshActiveSessionIfExternallyUpdated\b.*?^}",
-        src,
-        re.DOTALL | re.MULTILINE,
+    body = _function_body(src, "refreshActiveSessionIfExternallyUpdated")
+    assert re.search(
+        r"if\s*\(\s*\(\s*reason\s*\|\|\s*['\"]poll['\"]\s*\)\s*===\s*['\"]poll['\"]\s*&&\s*!\s*_isExternalSession\s*\(",
+        body,
+    ), (
+        "the 30s poll fallback should early-return for WebUI-native sessions, "
+        "but non-poll refresh triggers must still be allowed"
     )
-    assert m, "refreshActiveSessionIfExternallyUpdated function not found"
-    body = m.group(0)
-    assert re.search(r"if\s*\(\s*!\s*_isExternalSession\s*\(", body), (
-        "refreshActiveSessionIfExternallyUpdated must have an early-return "
-        "guard: if(!_isExternalSession(...))"
-    )
+
+
+def test_session_events_refresh_active_session():
+    src = SESSIONS_JS.read_text(encoding="utf-8")
+    body = _function_body(src, "_scheduleSessionEventsRefresh")
+    assert "refreshSessionList(reason||'event', {refreshActive:true})" in body


nesquena (approved):

Independent review of the active-WebUI-session refresh fix — the #3916/#4195 external-refresh guard returned early for **all** non-external sessions, so a `sessions_changed` SSE event updated the sidebar list but left the currently-open WebUI-native transcript stale when another client/process mutated it.

### What this ships

`static/sessions.js` (+7/-2) and `tests/test_issue3916_external_refresh_poll.py` (+24/-11). Authored by the agent. No "Closes #N" — it's a follow-up narrowing of the #3916 guard (the PR notes it's distinct from the #4205 data-durability guard).

Pulled a fresh `nousresearch/hermes-agent` tarball. The refresh functions (`refreshActiveSessionIfExternallyUpdated`, `_scheduleSessionEventsRefresh`, `refreshSessionList`) are webui-only frontend session-list lifecycle. (The lone `refreshActiveProfile` tarball hit is the agent's *desktop* app — a different concept, substring collision.) No agent protocol / config.yaml / session-schema surface. No cross-tool concern. ✓

### End-to-end trace

**Change 1 — narrow the #3916 guard to poll-only** ([sessions.js:3658](static/sessions.js#L3658)):
```js
if((reason||'poll')==='poll' && !_isExternalSession(S.session)) return;
```
The 30s poll (`ensureActiveSessionExternalRefreshPoll` → `refreshActiveSessionIfExternallyUpdated('poll')`) still skips WebUI-native sessions (preserving #3916's "don't probe native sessions forever"), but `event`/`focus`/`visible` triggers now proceed to the count/timestamp divergence check for native sessions too.

**Change 2 — event debounce reconciles the active transcript** ([sessions.js:3727](static/sessions.js#L3727)): `_scheduleSessionEventsRefresh` now calls `refreshSessionList(reason||'event', {refreshActive:true})`. `refreshSessionList` ([:3703](static/sessions.js#L3703)) renders the sidebar, then `if(refreshActive) await refreshActiveSessionIfExternallyUpdated('event')`. `'event' !== 'poll'` → the native gate is bypassed → divergence check runs.

**Critical safety — mid-stream stomping cannot happen.** This path is the #3545/#3877 danger zone, so I verified the guards are intact and sufficient:
- `if(S.busy || S.activeStreamId) return;` is **unchanged and still the first guard** ([sessions.js:3652](static/sessions.js#L3652)) — a live/queued WebUI turn blocks the refresh entirely.
- After the `await api(.../session?...&messages=0)` metadata fetch, the function **re-checks** `if(S.busy || S.activeStreamId) return;` ([sessions.js:3669](static/sessions.js#L3669)) — a TOCTOU guard so a stream that *starts during* the fetch still aborts the reload.
- `window._streamJustFinished` cooldown ([:3662](static/sessions.js#L3662)) prevents a reload right after the client's own stream (which would clear `S.toolCalls`/Activity).
- The reload (`loadSession(sid,{force:true})`) only fires on genuine divergence: `remoteCount > localCount || remoteLast > localLast` ([:3674](static/sessions.js#L3674)).

So an event refresh reloads the transcript **only** when the server genuinely has more/newer messages than the local view, and **never** while busy/streaming. A client's own just-sent turn is already in `localCount` → `remote == local` → no spurious self-echo reload; a turn added by another client/CLI → `remote > local` → reload (the intended fix).

**Coalescing is robust:** when `refreshSessionList` is already in-flight, the pending reason is stashed and re-dispatched through `_scheduleSessionEventsRefresh(pendingReason)` in the `finally` ([:3717](static/sessions.js#L3717)), which always re-injects `{refreshActive:true}` — so a coalesced `sessions_changed` event still reconciles the active transcript.

### Behavioural verification

**Reason-gate harness** (the narrowed guard across reasons × session type):
```
poll + native    -> SKIP (preserves #3916)
poll + external   -> proceed
undefined reason  -> SKIP (defaults to 'poll')
event + native    -> PROCEED (the fix)
focus + native    -> PROCEED
visible + native  -> PROCEED
event + external  -> PROCEED
7/7 reason-gate assertions passed
```

- **Security:** none — refresh-trigger logic; the reloaded data flows through the existing `loadSession`. No new sink.
- **Re-entrancy:** `_activeSessionExternalRefreshInFlight` guards against overlapping refreshes; the metadata fetch is lightweight (`messages=0&resolve_model=0`).
- **#3916 not reintroduced:** the 30s poll still skips native sessions (only the reason gate changed); the poll path passes `'poll'` directly and is unaffected.
- **Other `refreshSessionList` callers** default `refreshActive:false` → unchanged; only the event debounce (and explicit opt-ins) trigger the active reconcile.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Native session, 30s poll | still early-returns (poll+native) — #3916 preserved ✅ |
| Native session, `sessions_changed` event, another client added a msg | divergence (remote>local) → transcript reloads ✅ (the fix) |
| Native session, event echoing the client's own just-sent turn | remote==local → no reload ✅ |
| Active stream in progress when event fires | `S.busy/activeStreamId` guard (pre- and post-await) → no reload ✅ |
| Stream starts *during* the metadata fetch | post-await re-check aborts the reload ✅ |
| Just after own stream finishes | `_streamJustFinished` cooldown → no reload (Activity preserved) ✅ |
| Tab hidden | `document.hidden` guard → no reload ✅ |
| External/imported active session, event | proceeds as before ✅ |
| Coalesced events while a refresh is in-flight | re-dispatched with `{refreshActive:true}` → active still reconciles ✅ |
| CLI reads same session | webui-only refresh logic; agent unaffected ✅ |

### Tests

- `tests/test_issue3916_external_refresh_poll.py` — **3/3 pass** (poll-path-skips-native gate + `{refreshActive:true}` flag; the over-broad source-origin assertion correctly replaced with the narrower contract).
- **Full suite: 8813 passed / 135 skipped / 0 failures** in 191s (deselected the pre-existing darwin CRLF flake; ignored `test_passkey_auth.py`'s local `cryptography`-dep collection gap).
- **JS gates:** `node --check static/sessions.js` clean, ESLint runtime guard exit 0.
- **Behavioural harness:** 7/7 reason-gating (above).
- **CI on the PR head: all green** — browser-smoke + lint + 9 shards.

- **Greptile P2 (test-only):** the new `_function_body` test helper ends its regex with `^}` (first column-0 `}`), which would match too early if a target function had a nested block dedented to column 0. It's correct for the two functions it's used on here (their bodies have no column-0 `}` before the close), and it's a *test* helper, not production — fine as-is; a `}` -at-known-indent or brace-counting helper would harden it if reused more widely.
- `refreshActive` reconcile is implicitly coupled through `_scheduleSessionEventsRefresh` always passing `{refreshActive:true}` — robust for the current event path (including the coalesced re-dispatch), worth a comment if more callers are added later.

### Recommendation

A minimal, correctly-scoped fix: narrow the #3916 poll-only skip so real `sessions_changed`/focus/visibility events can reconcile an open WebUI-native transcript, and route `{refreshActive:true}` through the event debounce. The mid-stream danger this path is famous for is fully covered — the `S.busy || S.activeStreamId` guard is unchanged, re-checked after the async fetch, and the reload is divergence-gated with a post-stream cooldown — so it reloads only when the server genuinely diverges and never while streaming. Ship.

nesquena-hermes:
Shipped in **v0.51.466** (Release QA) via #4335. Rebased onto v0.51.465, full gate green (suite 9254 + Codex SAFE + Opus SAFE — confirmed no event-storm: client refreshes are pure reads, the server broadcasts only on mutations, and the divergence check no-ops when in sync). Your commit is preserved in the release. Closing as integrated. 🚀

PR #4211, opened 2026-06-15
Fix replies to cron sessions

### LINKED ISSUE #3975: Replying to cron job sessions doesn't work, redirects to start page
Steps to reproduce: 

1. In Settings > Preferences, enable `Show non-WebUI sessions` and `Show cron sessions`
2. Setup cron job 
3. After job runs and shows up in chat, try replying to it. Instead of sending the message, this just redirects me to the new chat page

https://github.com/user-attachments/assets/3b27db18-24e6-4a69-8e11-1db916e09796

Might be due to how i setup the job? Here's the non-sensitive part of the prompt that triggers the cron:

[IMPORTANT: You are running as a scheduled cron job. DELIVERY: Your final response will be automatically delivered to the user — do NOT use send_message or try to deliver the output yourself. Just produce your report/output as your final response and the system handles the rest. SILENT: If there is genuinely nothing new to report, respond with exactly "[SILENT]" (nothing else) to suppress delivery. Never combine [SILENT] with content — either report your findings normally, or say [SILENT] and nothing more.]

Also some info on how the cron is set up:

- Deliver: origin (WebUI).
- Toast notifications: enabled (default for origin delivery).

TomBanksAU:
## Summary
Fixes replies from cron job sessions shown in the WebUI. Those sessions could be opened from state.db, but chat start treated them as missing and the frontend reset to the start page.

## What changed
- Route `/api/chat/start` now uses the existing materializing session resolver.
- Read-only imported sessions return 403 instead of being handled like deleted sessions.
- Added regression coverage for replying to a materializable cron session.

## Testing
- `uv run --with pytest --with pyyaml --with cryptography pytest -q tests/test_issue3975_cron_reply_materialization.py tests/test_issue3994_materialize_session.py tests/test_issue3585_cron_session_overflow.py`
- `uv run --with ruff python scripts/ruff_lint.py --diff origin/master`
- `python3 -m py_compile api/routes.py tests/test_issue3975_cron_reply_materialization.py`

Closes #3975

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -13658,9 +13658,11 @@ def _handle_chat_start(handler, body, diag=None):
             return bad(handler, str(e))
         diag.stage("get_session") if diag else None
         try:
-            s = get_session(body["session_id"])
+            s = _get_or_materialize_session(body["session_id"])
         except KeyError:
             return bad(handler, "Session not found", 404)
+        except PermissionError:
+            return bad(handler, "Read-only imported sessions cannot be continued from WebUI", 403)
         diag.stage("validate_profile") if diag else None
         requested_profile = str(body.get("profile") or "").strip()
         if requested_profile:
--- a/tests/test_issue3975_cron_reply_materialization.py
+++ b/tests/test_issue3975_cron_reply_materialization.py
@@ -1,0 +1,85 @@
+"""Regression coverage for replying to cron-origin sessions (#3975)."""
+
+import io
+import json
+from types import SimpleNamespace
+
+import api.routes as routes
+
+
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in v0.51.425 (Release OL) via #4219. Thanks @TomBanksAU — your 3-line root-cause fix for #3975 (materialize the missing WebUI sidecar via _get_or_materialize_session + 403 on read-only writes) is the preferred minimal fix; cron replies now send instead of redirecting to start. Codex + Opus + full suite green; security boundary confirmed intact. Closing as shipped.

Issue #3872, opened 2026-06-09
Custom provider with remote base_url: model IDs containing "/" are incorrectly truncated


haolf000:
## Problem

When using `provider: custom` with a remote `base_url` pointing at a custom API gateway, model IDs that contain `/` (e.g. `bedrock/opus-4-6`) have their prefix stripped before being sent to the endpoint.

The endpoint receives `opus-4-6` instead of `bedrock/opus-4-6`, resulting in 403/404 errors.

## Config

```yaml
model:
  provider: custom
  base_url: https://my-gateway.example.com/v1
  default: smart
```

## Steps to reproduce

1. Configure a custom provider pointing at a remote API gateway
2. The gateway's `/v1/models` returns namespaced model IDs like `bedrock/opus-4-6`
3. Select `bedrock/opus-4-6` from the WebUI model picker
4. Send a message → gateway receives truncated `opus-4-6` → 403 "model not allowed"

## Expected behavior

For bare `custom` provider, model IDs should be passed through unchanged to the endpoint. The endpoint defines its own model namespace and knows how to handle its own IDs.

This is consistent with hermes-agent's own `normalize_model_for_provider()` which preserves model IDs as-is for `custom` provider.

## Versions

- **Deployed (reproduces):** v0.51.195
- **Latest source checked (still present):** v0.51.340

## Workaround

Use model IDs that don't contain `/` (e.g. configure gateway aliases like `claude-opus-4-6` instead of `bedrock/opus-4-6`).

nesquena-hermes:
Thanks for the detailed report and the cross-reference to the agent's `normalize_model_for_provider()`. I traced the full model-id path on `master` (v0.51.340) and want to share what I found, because the WebUI side appears to preserve the slash at every layer — I could not reproduce the truncation in current source.

## What the code does on .340

**1. Backend resolver fast-path** (`api/routes.py:2293-2306`). When chat start receives a model plus an explicit `model_provider`, and the model is *not* `@provider:model`-qualified, it returns the inputs verbatim:

```python
bare_model, explicit_provider = _split_provider_qualified_model(model)
model_prefix = model.split("/", 1)[0].strip().lower() if "/" in model else ""
stale_codex_openai_slash_id = (requested_provider == "openai-codex" and model_prefix == "openai")
if not explicit_provider and not stale_codex_openai_slash_id:
    return model, requested_provider, False
```

[...]

PR #4212, opened 2026-06-15
fix: keep custom provider when model id has known-provider prefix (#4210)

### LINKED ISSUE #4210: Custom provider (no base_url) hijacked to openrouter when model id has known provider prefix
## Summary
When the active provider is a bare `custom` or named `custom:<slug>` **and no `base_url` is set**, `resolve_model_provider()` in `api/config.py` hijacks model IDs that contain a known-provider prefix (e.g. `x-ai/grok-2`, `google/gemma-2`) and routes them through `openrouter` with `base_url=None`. The backend then fails to initialize because the user has no OpenRouter credentials configured, surfacing:

```
RuntimeError: No LLM provider configured. Run `hermes model` to select a provider, or run `hermes setup` for first-time configuration.
```

## Repro
```python
import api.config as config
old = dict(config.cfg)
config.cfg["model"] = {"provider": "custom:llm-proxy", "default": "x-ai/grok-2"}
try:
    m, p, b = config.resolve_model_provider("x-ai/grok-2")
finally:
    config.cfg.clear(); config.cfg.update(old)

# Observed:  m='x-ai/grok-2', p='openrouter', b=None      (BUG)
# Expected:  m='x-ai/grok-2', p='custom:llm-proxy', b=None
```

## Root cause
`api/config.py` `resolve_model_provider()`, line 2061 (the `base_url is None` branch):

```python
if prefix in _PROVIDER_MODELS and prefix != config_provider:
    return model_id, "openrouter", None
```

This is the cross-provider-via-OpenRouter fallback. When the active provider is `custom` or `custom:<slug>` it is not actually a cross-provider request — the prefix is just an intrinsic routing segment for the custom proxy, and the user did not pick anything from the OpenRouter dropdown. The same exemption is already applied in the `config_base_url is set` branch above (line 2050: `_cp_lower == "custom" or _cp_lower.startswith("custom:")` ⇒ do not strip the prefix). The no-`base_url` branch needs the same gate.

## Related
- #3872 (Release LM, v0.51.349) fixed the `base_url`-set variant; the no-`base_url` variant is the sibling that was missed.
- The error message ("No LLM provider configured") comes from hermes-agent after the WebUI hands it a provider id (`openrouter`) the user never selected.

## Expected
With `provider=custom:llm-proxy` (no `base_url`) and `model_id=x-ai/grok-2`:
- `resolve_model_provider` returns `(model_id, "custom:llm-proxy", None)` (preserve the prefix, keep the custom provider).
- Routing to `openrouter` should only happen for the cross-provider dropdown case, not for a custom provider.

## Suggested fix
Add a custom-provider exemption to the line 2061 redirect:

```python
_cp_lower = (config_provider or "").strip().lower()
_is_custom = _cp_lower == "custom" or _cp_lower.startswith("custom:")
if prefix in _PROVIDER_MODELS and prefix != config_provider and not _is_custom:
    return model_id, "openrouter", None
```

(And symmetrically, the explicit `prefix == config_provider` strip at line 1984 and the custom_providers cross-reference at line 2003 should also be reviewed for the same gap — the bug report covers only the line 2061 redirect here.)

b3nw:
# PR: Custom provider (no `base_url`) keeps the request instead of redirecting to OpenRouter (#4210)

## Thinking Path

- The user reported a routing bug: with `provider: custom:llm-proxy` (no `base_url` set) and a model id like `x-ai/grok-2`, the WebUI hijacked the request to `provider=openrouter, base_url=None` and the backend then crashed with `RuntimeError: No LLM provider configured`.
- I traced the call site in `api/config.py::resolve_model_provider()`. The cross-provider redirect lives at the very end of the `/`-branch (in this branch: line 2190 in `upstream/master @ 92ecf8ed`):
  ```python
  if prefix in _PROVIDER_MODELS and prefix != config_provider:
      return model_id, "openrouter", None
  ```
  The slash branch's earlier sub-branches do not fire here: the active provider is `custom:llm-proxy` (not `openrouter`, not in the portal set, not matching `prefix`), the `config_base_url` branch is skipped because no `base_url` is set, and the `custom_providers[]` cross-reference is skipped because there is no `custom_providers[].name == "x-ai"`. The hijack path is the only one left standing.
- Release LM / #3872 (v0.51.349) fixed the **base_url-set** sibling of this bug by exempting custom providers from the prefix-strip at line 2180-2183:
  ```python
  _cp_lower = (config_provider or "").strip().lower()
  _is_custom = _cp_lower == "custom" or _cp_lower.startswith("custom:")
  if prefix in _PROVIDER_MODELS and (
      not _is_custom or _is_first_party_model(prefix, bare)
  ):
      return bare, config_provider, config_base_url
  ```
  The same `_is_custom` exemption was never applied to the **no-base_url** redirect at the bottom of the slash branch. That is the gap.
- The fix is therefore the symmetric exemption: don't redirect to OpenRouter when the active provider is `custom` or `custom:<slug>`, regardless of whether a `base_url` is set. The custom provider keeps the request and the full model id; downstream code (`resolve_custom_provider_connection`, `resolve_runtime_provider`) supplies the actual base_url and credentials at request time.

Closes #4210 

## What Changed

- `api/config.py` — `resolve_model_provider()`: added an `_is_custom_cross` guard to the cross-provider-redirect branch (line 2190), matching the exemption `config_base_url` already carries (line 2180). Comment block explains the proxy-routing rationale and points to #3872 as the sibling fix.
- `tests/test_model_resolver.py` — added two regression tests at the end of the file (matching the file's existing section-comment style):
  - `test_custom_provider_no_base_url_with_known_prefix_keeps_custom_and_full_id_4210` — bug-report case 1 verbatim: `custom:llm-proxy` + no `base_url` + `x-ai/grok-2`.
  - `test_bare_custom_provider_no_base_url_with_known_prefix_keeps_custom_and_full_id_4210` — sibling: bare `custom` (no `<slug>`) + `google/gemma-2-9b`.
- `CHANGELOG.md` — added an `[Unreleased]` / `### Fixed` entry explaining the user-visible failure and the fix.

## Why It Matters

- A user picking a `custom:llm-proxy` provider with a slash-prefixed model id would see a `RuntimeError: No LLM provider configured` from the backend even though they had a working custom endpoint. The WebUI silently rewrote their selection to `openrouter`, where they had no credentials, and the backend refused to start.
- The base_url-set variant of this was already fixed (#3872); this PR closes the remaining gap so both variants behave the same way.
- Scope is intentionally tight: one branch, one guard, two tests. The `prefix == config_provider` strip (line 1983) and the `custom_providers[]` cross-reference (line 2003) are out of scope for this PR — they are called out in the linked issue for future work if they turn out to have a similar gap, but I did not see a clear failing case for them in the bug report or the model catalog.

## Verification

- **Focused test runs (targeted, in the worktree at `7e0c958b`):**
  - `uv run --with pytest --with pytest-timeout --with pyyaml --with requests pytest tests/test_model_resolver.py -v` → 32 passed (30 existing + 2 new RED tests).
  - `uv run --with pytest --with pytest-timeout --with pyyaml --with requests pytest tests/test_model_resolver.py tests/test_custom_provider_bare_model_reasoning.py -v` → 66 passed.
- **Full test suite:** `uv run --with pytest --with pytest-timeout --with pyyaml --with requests --with cryptography pytest tests/ -q --timeout=60` reported a mix of "8850 passed" (one run) and "8885 errors" (a different run with a broader dep set); both numbers are real but not directly comparable across runs. The 7 test files I checked by hand for pre-existing failures are listed below with the actual root cause.
- **TDD discipline:** both new tests were written first and verified to fail with `AssertionError: ... got provider='openrouter' model='x-ai/grok-2'` (and the bare-`custom` sibling) before the source change was made; both pass after the change.
- **Cross-checked by hand** at the function level — the new guard only changes the path that previously hijacked to OpenRouter. All other branches (`config_base_url` set, `prefix == config_provider`, portal providers, `custom_providers[]` cross-reference) are unchanged.

## Risks / Follow-ups

- **Risk:** `_PROVIDER_MODELS` membership drives the guard. If a vendor adds a new namespace that also happens to be a real first-party provider prefix (e.g. a proxy that uses `anthropic/...` for routing), the custom provider will keep the request and the proxy decides what to do with it. That matches the contract #3872 established for the base_url variant.

## Model Used

- `minimax/minimax-m3 via hermes-agent & hermes-webui` — for the investigation, test design, fix, and PR text.
- `google/gemini-flash-3.5-high` via antigravity-cli for code review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 
 - **Optional Todos tab in the workspace panel (off by default).** A new Settings toggle ("Show Todos tab in workspace panel") adds a Todos tab alongside Files and Artifacts in the right-side workspace panel, mirroring the active session's task list with status icons (completed, in-progress, pending, cancelled). The existing sidebar Todos panel remains available regardless of the setting. (#3564)
 
+### Fixed
+
+- **Custom providers with no `base_url` no longer get hijacked to OpenRouter when the model id has a known-provider prefix.** A bare `custom` or named `custom:<slug>` provider pointed at a vendor-routing proxy (e.g. `custom:llm-proxy` with no configured `base_url`, since the proxy URL is supplied at request time) used to receive an OpenRouter redirect for model ids like `x-ai/grok-2` or `google/gemma-2` whenever the prefix was a member of `_PROVIDER_MODELS`. The backend then failed to initialize with `RuntimeError: No LLM provider configured` because the user had no OpenRouter credentials configured. `resolve_model_provider()` now applies the same custom-provider exemption the `config_base_url` branch already carries (sibling of #3872 / v0.51.349), so the request stays on the user's selected custom provider with the full model id preserved. (#4210)
+
 ## [v0.51.421] — 2026-06-14 — Release OH (preserve colon-suffixed model IDs in normalization, #3959)
 
 ### Fixed
--- a/api/config.py
+++ b/api/config.py
@@ -2187,7 +2187,15 @@ def resolve_model_provider(model_id: str) -> tuple:
         # If prefix does NOT match config provider, the user picked a cross-provider model
         # from the OpenRouter dropdown (e.g. config=anthropic but picked openai/gpt-5.4-mini).
         # In this case always route through openrouter with the full provider/model string.
-        if prefix in _PROVIDER_MODELS and prefix != config_provider:
+        # Exception (#4210): a custom provider (bare ``custom`` or named ``custom:<slug>``)
+        # is a vendor-routing proxy, not a first-party provider — its model ids commonly
+        # contain a known-provider prefix that the proxy uses for upstream routing, not
+        # an OpenRouter dropdown selection. Keep the request on the custom provider; the
+        # base_url-set sibling of this exception lives earlier in the ``config_base_url``
+        # branch (#3872).
+        _cp_lower_cross = (config_provider or "").strip().lower()
+        _is_custom_cross = _cp_lower_cross == "custom" or _cp_lower_cross.startswith("custom:")
+        if prefix in _PROVIDER_MODELS and prefix != config_provider and not _is_custom_cross:
             return model_id, "openrouter", None
 
     return model_id, config_provider, config_base_url
--- a/tests/test_model_resolver.py
+++ b/tests/test_model_resolver.py
@@ -709,3 +709,42 @@ def test_custom_endpoint_slash_model_routes_to_custom_not_openrouter():
     assert model_or == 'google/gemma-4-26b-a4b', (
         "Model name should be preserved for openrouter, got '{}'.".format(model_or)
     )
+
+
+# ── #4210: custom provider (no base_url) must not be hijacked to openrouter
+#    when the model id has a known-provider prefix (sibling of #3872, which
+#    only covered the base_url-set variant). Bug-report case 1.
+
+def test_custom_provider_no_base_url_with_known_prefix_keeps_custom_and_full_id_4210():
+    """#4210: provider=custom:llm-proxy (no base_url) + 'x-ai/grok-2' must NOT
+    be redirected to openrouter. The prefix is intrinsic to the custom proxy's
+    routing; the user did not pick anything from the OpenRouter dropdown."""
+    model, provider, base_url = _resolve_with_config(
+        'x-ai/grok-2',
+        provider='custom:llm-proxy',
+        default='x-ai/grok-2',
+    )
+    assert provider == 'custom:llm-proxy', (
+        "Custom provider must not be hijacked to openrouter when no base_url is "
+        "set; got provider={!r} model={!r}".format(provider, model)
+    )
+    assert model == 'x-ai/grok-2', (
+        "Custom provider must preserve the full model id; got model={!r}".format(model)
+    )
+    assert base_url is None
+
+
+def test_bare_custom_provider_no_base_url_with_known_prefix_keeps_custom_and_full_id_4210():
+    """#4210 sibling: bare 'custom' (no 'custom:<slug>') with no base_url
+    must also not be hijacked to openrouter for a known-prefix model id."""
+    model, provider, base_url = _resolve_with_config(
+        'google/gemma-2-9b',
+        provider='custom',
+        default='google/gemma-2-9b',
+    )
+    assert provider == 'custom', (
+        "Bare 'custom' provider must not be hijacked to openrouter when no "
+        "base_url is set; got provider={!r}".format(provider)
+    )
+    assert model == 'google/gemma-2-9b'
+    assert base_url is None


nesquena-hermes:
## Summary

Reviewed this against `origin/master` (HEAD `d0fb2aa2`) by pulling the PR head (`bdc513bc`) into a read-only worktree and reading `resolve_model_provider()` end to end. The fix is correct, minimal, and lands in the right place: it gives the no-`base_url` cross-provider redirect the same custom-provider exemption the `config_base_url` branch already carries for #3872. The two new tests are genuine red→green — on master the same inputs hit the unguarded redirect and return `openrouter`, so they fail there and pass here.

## Code reference

The guarded redirect at `api/config.py:2195` (PR head):

```python
_cp_lower_cross = (config_provider or "").strip().lower()
_is_custom_cross = _cp_lower_cross == "custom" or _cp_lower_cross.startswith("custom:")
if prefix in _PROVIDER_MODELS and prefix != config_provider and not _is_custom_cross:
    return model_id, "openrouter", None
```

On `origin/master:api/config.py` the same line is the unguarded `if prefix in _PROVIDER_MODELS and prefix != config_provider:` — confirming the gap the issue describes and that the new tests exercise the changed path.

## Diagnosis / one point to confirm

The fix is right. One thing worth a maintainer call before merge: the two custom branches now diverge on *redundant first-party* prefixes. The `config_base_url` sibling at `api/config.py:2179-2183` strips a first-party prefix when the bare id is a real first-party model:

```python
_is_custom = _cp_lower == "custom" or _cp_lower.startswith("custom:")
if prefix in _PROVIDER_MODELS and (
    not _is_custom or _is_first_party_model(prefix, bare)
):
    return bare, config_provider, config_base_url
```

So with a `base_url` set, `custom:* + openai/gpt-5.4` resolves to bare `gpt-5.4` (since `gpt-5.4 ∈ openai` catalog — `_is_first_party_model` at `api/config.py:1817`). The new no-`base_url` guard keeps the *full* id unconditionally, so `custom:* + openai/gpt-5.4` now stays `openai/gpt-5.4`. For an intrinsic routing prefix like `x-ai/grok-2` (grok-2 ∉ x-ai first-party catalog) both branches agree — keep whole — which is exactly the bug-report case and why the tests pass. The divergence only shows for first-party-namespaced ids.

That asymmetry is defensible: with no `base_url` the actual endpoint and creds come from `resolve_custom_provider_connection()` at request time, and full-id passthrough is the conservative default (let the proxy route). But it does mean the same model selection resolves to a different downstream id depending on whether `base_url` happens to be set on the custom entry. If full-id passthrough is the intended contract for the no-`base_url` custom case, a one-line comment saying so (and that it intentionally skips the `_is_first_party_model` strip the base_url branch does) would save the next reader the same double-take.

## Verification

- `git diff origin/master...HEAD --stat`: `api/config.py` (+9/-1), `tests/test_model_resolver.py` (+39), `CHANGELOG.md` (+4) — tightly scoped, matches the PR description.
- Confirmed the two regression tests use the existing `_resolve_with_config` helper (saves/restores `config.cfg`), so they're isolated and match the file's style.
- Per cron policy I do not execute PR code, so I did not run pytest myself; the diff and the master-vs-head comparison support the red→green claim.

Scoping the `prefix == config_provider` strip and the `custom_providers[]` cross-reference out of this PR is the right call — neither has a demonstrated failing case in #4210.

nesquena-hermes:
Shipped in v0.51.426 (Release OM) via #4220. Thanks @b3nw — custom providers with no base_url (vendor-routing proxies) are no longer hijacked to OpenRouter when the model id carries a known-provider prefix; resolve_model_provider() now extends the #3872 custom-provider exemption to the no-base_url case. Codex + Opus + full suite all green; confirmed scoped to custom configs only (legit cross-provider OpenRouter routing unaffected). Closing as shipped.

Issue #3988, opened 2026-06-11
Default show_cli_sessions to true


rodboev:
The `show_cli_sessions` setting defaults to `false`, which means sessions from the TUI, Telegram, Discord, CLI, and the Hermes One desktop app are hidden from the webui sidebar unless the user discovers and enables this toggle in Settings.

This is surprising because both the built-in dashboard (port 9119) and Hermes One show all sessions by default, including webui-originated ones. The asymmetry means users who chat via TUI or messaging platforms see their conversations vanish when they open the webui, with no indication that a setting controls this.

The toggle was added four days after the project's initial release (commit `66f95e08`, 2026-04-03) with the rationale "off by default so the sidebar is clean." That made sense at the time, but multi-client usage is now common, and hiding sessions by default creates confusion rather than cleanliness.

**Proposal:** Change the default value of `show_cli_sessions` from `false` to `true`. This is a one-line change and doesn't conflict with any of the more ambitious filter redesigns being discussed:

- #3418 (multi-select origin filter replacing the binary toggle)
- #3403 (all-sessions sidebar filter)
- #3122 (dynamic per-source session tabs)

All of those can land on top of a `true` default. In the meantime, users who install hermes-webui alongside hermes-agent would see their sessions without needing to hunt through settings.

The performance concern from #2628 was mitigated by the 20-session cap in #2647. Users with exceptionally large `state.db` files can still toggle the setting off.

mortang2410:
OH FREAKING YES. 

THANK GOD YOU MADE THIS. 

What a nincompoop whoever decided this should be turned off. Why the hell do you think people use WebUI ?

nesquena-hermes:
Shipped in v0.51.427 (Release ON) via #4230. show_cli_sessions now defaults ON for new installs — CLI/TUI/Telegram/Discord/Hermes One sessions appear in the sidebar without hunting for the toggle. Existing installs are grandfathered to their previous behavior. Combined the best of #3997 (@rodboev) and #4222 (@Sanjays2402).

Issue #3418, opened 2026-06-02
UX: unified sidebar session filter (origin multi-select) instead of WebUI/CLI tabs


pamnard:
## Summary

Replace the fragmented sidebar session filtering UX (WebUI/CLI tabs + misplaced Settings toggles + misleading search placeholder) with a single **filter popover** next to the session list: search stays separate; a funnel control opens **origin** multi-select plus two visibility policies.

**Colocation principle:** options that control the chat sidebar must live **on the sidebar**, not in Settings → Preferences. Today two sidebar-related toggles sit far from the UI they affect — that is incorrect and should be fixed as part of this work, not left as duplicate surfaces.

This is a **feature request** — a reference implementation exists in a downstream fork and can be offered as an upstream PR if the approach is accepted.

## Problem

Today, filtering external/agent sessions is split across several surfaces that do not compose well:

1. **Settings → “Show non-WebUI sessions”** (`show_cli_sessions`) — server-side gate for merging CLI/gateway/state.db rows, but exposed in Preferences far from the session list.
2. **Settings → “Show previous messaging sessions”** — same problem: controls sidebar projection, lives in Settings.
3. **Sidebar WebUI / CLI tabs** (when external sessions are enabled) — mutually exclusive, not multi-select.
4. **Search field** placeholder historically suggested filtering, but search is content/title matching only.
5. **Profile switch** can make messaging origins appear/disappear even when gateway platforms are unchanged, because `/api/gateway/status` `platforms` were derived only from **profile-scoped** `sessions/sessions.json` identity map — not from gateway runtime/config.

Users expect:

- **Origin checkboxes** list all globally available sources (WebUI, CLI, Cron, Telegram, …) regardless of whether the active profile currently has rows for that origin.
- **Counts** beside an origin are profile-scoped (e.g. Telegram `(2)` on one profile, `(0)` on another) — that is fine.
- **Checking/unchecking an origin** filters the sidebar client-side only; toggling one origin must not disable unrelated origin rows or wipe their counts via a separate “master switch” or server refetch.

### Toolbar
- Search: placeholder **“Search conversations…”** (text search only).
- Funnel icon opens a popover (flat on sidebar background — no extra chrome).

### Popover — Origin (multi-select, primary control)
- **WebUI** — always on, checkbox disabled.
- **CLI**, **Cron**, and dynamic messaging origins (Telegram, Discord, Slack, …) from gateway/config — independently toggleable.
[...]

Issue #1925, opened 2026-05-08
RFC: Make WebUI a thin observability/control client over Hermes Agent runtime


Michaelyklam:
## Summary

Hermes WebUI currently gives a polished browser experience, but browser-originated chat runs are executed inside the WebUI server process. That means WebUI is more than an interface layer: it creates in-memory streams, starts background agent threads, instantiates/reuses `AIAgent`, and owns live cancellation/streaming state.

This RFC proposes shifting WebUI toward the thinnest possible client that still gives full parity features:

> Hermes Agent owns execution, sessions, active runs, slash-command semantics, approvals, clarify, tools, models, and durable state. WebUI owns presentation, browser UX, and control/observability surfaces over Hermes.

The goal is **not** to reduce WebUI capability. The goal is to prevent WebUI from becoming a parallel Hermes runtime.

## Current behavior / motivation

Today, the WebUI chat path roughly does this:

1. Browser calls `POST /api/chat/start`.
2. WebUI creates a `stream_id` and `StreamChannel`.
3. WebUI stores the stream in process-global `STREAMS`.
4. WebUI starts a daemon thread running `_run_agent_streaming`.
5. `_run_agent_streaming` constructs/reuses `AIAgent` and calls `agent.run_conversation(...)`.
6. Agent callbacks push token/tool/reasoning/approval/clarify events back into WebUI-owned in-memory queues.

This has some nice properties: direct low-latency SSE, simple local deployment, and direct access to the live agent object. But it also creates real costs:

- WebUI restarts/updates can kill active browser-originated runs.
- Browser refresh/reconnect semantics depend on WebUI process memory.
- WebUI has to maintain runtime shims for Hermes behavior.
- Features such as slash commands, `/goal`, model/provider routing, approvals, clarify, queue/interrupt/steer, etc. risk being reimplemented in WebUI instead of inherited from Hermes.
- Active WebUI streams can block safe WebUI updates.
- WebUI cannot be treated as a disposable observability/control layer.

## Design principle

For every feature, ask:

> Is this Hermes behavior, or WebUI presentation?

Hermes behavior should be exposed from Hermes and consumed by WebUI. WebUI should own only browser-specific presentation and interaction.

### Hermes should own

- agent execution
- sessions and transcript persistence
- active run lifecycle
- model/provider/toolset routing
- memory and skills
- slash command semantics/capabilities
- approvals and clarify lifecycle
- cancel/interrupt/queue/steer semantics
- background and goal continuation
[...]

PR #4213, opened 2026-06-15
Fix TUI session sidebar discoverability

dso2ng:
## Summary
- normalize Hermes Agent `source='tui'` rows as CLI-like sidebar rows with a `TUI` label
- keep non-empty TUI rows visible even when their title is the generic `TUI Session`
- display the latest TUI continuation tip title in collapsed sidebar rows instead of an older parent title
- treat legacy browser payloads with raw `tui` source as CLI-like for external-session action gating

## Problem
WebUI projects Hermes Agent `state.db` sessions into the sidebar. Sessions created or continued from TUI/Desktop-like surfaces can arrive with `source='tui'` and parent/child continuation metadata. Before this change those rows could be treated as `other` rather than CLI/agent sessions, hidden by generic CLI-title visibility rules, or displayed with an older parent title after continuation collapse. The result is that an active TUI conversation can appear to disappear from the WebUI sidebar even though the session is still present in `state.db`.

## Fix
This keeps the existing storage/projection model, but hardens the source contract for TUI-origin rows:

- `normalize_agent_session_source('tui')` returns `session_source='cli'`, `raw_source='tui'`, `source_label='TUI'`
- `is_cli_session_row()` recognizes `tui` in raw/source/source_label fields
- `is_cli_session_row_visible()` preserves non-empty TUI rows instead of applying the generic untitled-CLI threshold
- collapsed TUI continuation rows keep navigation pointed at the latest tip and use that tip's visible title
- `_isCliSession()` recognizes legacy raw `tui` browser payloads

## What this does not do
- does not migrate WebUI JSON sidecars to `state.db`
- does not implement bidirectional title sync
- does not change Desktop/TUI runtime behavior
- does not change the default sidebar source filter setting
- does not delete, archive, or rewrite existing session data

## Related issues
Related to #3986, #3988, and #3418. This is a narrow discoverability/projection hardening slice; #498 and #1925 remain the broader storage/runtime direction.

## Tests
- `python -m pytest -q tests/test_issue2351_cli_session_source_filter.py tests/test_1466_sidebar_cancel_clarify.py tests/test_issue3586_cli_session_source_label.py tests/test_issue3930_source_filter_pushdown.py tests/test_gateway_sync.py::test_compression_projection_handles_deep_lineage_iteratively tests/test_gateway_sync.py::test_compression_chain_with_empty_latest_tip_falls_back_to_latest_importable_segment`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **TUI-origin Hermes Agent sessions now stay discoverable in the CLI/agent sidebar view.** State-db rows with `source='tui'` are normalized as CLI-like rows, legacy browser payloads also treat raw `tui` as CLI-like for destructive-action gating, and TUI continuation chains display the latest tip title instead of an older parent title. (#3986, #3988, #3418)
+
 ## [v0.51.422] — 2026-06-14 — Release OI (optional Todos tab in workspace panel, #3564)
 
 ### Added
--- a/api/agent_sessions.py
+++ b/api/agent_sessions.py
@@ -31,6 +31,7 @@ SOURCE_LABELS = {
     'slack': 'Slack',
     'telegram': 'Telegram',
     'tool': 'Tool',
+    'tui': 'TUI',
     'webui': 'WebUI',
     'weixin': 'Weixin',
 }
@@ -47,7 +48,7 @@ def normalize_agent_session_source(raw_source: str | None) -> dict:
 
     if raw == 'webui':
         session_source = 'webui'
-    elif raw == 'cli':
+    elif raw in {'cli', 'tui'}:
         session_source = 'cli'
     elif raw in MESSAGING_SOURCES:
         session_source = 'messaging'
@@ -176,7 +177,12 @@ def is_cli_session_row(row: dict) -> bool:
         return False
     if source == "cli":
         return True
-    if source_tag == "cli" or raw_source == "cli" or source_name == "cli" or source_label == "cli":
+    if (
+        source_tag in {"cli", "tui"}
+        or raw_source in {"cli", "tui"}
+        or source_name in {"cli", "tui"}
+        or source_label in {"cli", "tui"}
+    ):
         return True
 
     # Legacy imported CLI rows may only be marked as CLI in sidebar metadata.
@@ -201,6 +207,14 @@ def is_cli_session_row_visible(row: dict) -> bool:
     message_count = _as_positive_int(row.get("actual_message_count") or row.get("message_count"))
     if message_count <= 0:
         return False
+
+    if "tui" in {
+        _normalize_source_name(row.get("source")),
+        _normalize_source_name(row.get("source_tag")),
+        _normalize_source_name(row.get("raw_source")),
+        _normalize_source_name(row.get("source_label")),
+    }:
+        return True
 
     if _has_cli_lineage(row):
         return True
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Pulled the branch and read `api/agent_sessions.py` + `static/sessions.js` against the agent's source-emission contract in `~/.hermes/hermes-agent/`. The diff is correctly scoped and the cross-repo premise holds — this is a faithful projection-hardening slice. A few confirmations and one small note.

## Cross-repo premise verified

The agent does emit `source='tui'` for TUI-origin sessions. `hermes_cli/main.py:2142`:

```python
source = "tui" if use_tui else "cli"
last_id = _resolve_last_session(source=source)
```

And, critically for the title-override branch, compression-created **continuation tips** re-stamp the source from the live agent platform — `agent/conversation_compression.py:544`:

```python
source=agent.platform or os.environ.get("HERMES_SESSION_SOURCE", "cli"),
```

For a TUI run `agent.platform == 'tui'`, so the tip segment carries `source='tui'`. That's exactly the field your new branch in `_project_agent_session_rows` keys on (`if str(tip.get('source') or '').strip().lower() == 'tui'`), so the override fires for the real continuation-chain case it targets — not just the root. Good call routing on the tip rather than the root here; the root is what `merged` starts as, and the tip is the freshest segment.

## Normalization + visibility look right

`normalize_agent_session_source('tui')` collapsing to `session_source='cli'` with `source_label='TUI'` (`agent_sessions.py:51`) lines up with how `is_cli_session_row` / `is_cli_session_row_visible` consume the row, and matches the existing CLI handling at `agent_sessions.py:179`. The fast-path keep:

```python
if "tui" in {
    _normalize_source_name(row.get("source")),
    _normalize_source_name(row.get("source_tag")),
    _normalize_source_name(row.get("raw_source")),
    _normalize_source_name(row.get("source_label")),
}:
    return True
```

correctly runs **after** the `message_count <= 0` early return at `:209`, so an empty TUI row still gets filtered — it only spares non-empty TUI rows from the generic untitled-CLI threshold, which is the stated intent.

## The title-override is safe

Greptile flagged that the override checks only `tip.source`. That's the right field (per the compression re-stamp above), and both writes are guarded (`if tip.get('title')` / `if tip.get('source')`), so an empty-titled tip degrades to the pre-PR fallback rather than blanking the row. No regression risk there.

## One small note: `_isCliSession` widening in sessions.js

`static/sessions.js:1421` adding `raw === 'tui'` mirrors the server contract correctly. One thing to double-check on the frontend side: this widens external-session **action gating**, so confirm the TUI rows now classified as CLI-like get the same action affordances (and disabled-action tooltips) you intend — e.g. that a TUI row doesn't suddenly expose a WebUI-only action that the in-process session would gate off. The `_isMessagingSession` short-circuit just below keeps messaging rows out, so the only behavioral delta is `tui` joining `cli`, which is what you want.

Net: backend contract checks out against the agent, the visibility/normalization changes are additive and well-ordered, and the two new tests in `test_issue2351_cli_session_source_filter.py` pin the normalization + continuation-title projection. Looks solid to me.

nesquena-hermes:
Shipped in v0.51.432 (Release OS) via #4237. Thanks @dso2ng — TUI sessions are now classified like CLI sessions, so they show up in the sidebar, and a TUI continuation tip keeps its visible title for navigation. Codex + Opus + full suite (9062) all green; confirmed exact-set classification (no over-tagging), the tip override surfaces only the freshest visible TUI segment, and the snapshot-title fallback for non-TUI lineages is unchanged. Closing as shipped.

Issue #2361, opened 2026-05-16
Define WebUI run state consistency across transcript, context, streams, and replay


franksong2702:
## Problem

Recent WebUI streaming/recovery bugs are not isolated rendering mistakes. They repeatedly happen at the boundaries between several copies of the same running agent turn:

- the visible transcript the user can see,
- the model context / `context_messages` the agent actually receives,
- `pending_user_message` and active stream metadata,
- live SSE events and in-memory stream state,
- durable run journal / replay state,
- automatic compression summaries and active-task handoff text,
- the browser's live timeline DOM/cache,
- sidebar ordering, unread state, and `updated_at` metadata.

When those layers diverge, the user sees failures that feel unrelated: a visible prompt is not sent back to the model after recovery, a live run loses or reorders its tool/thinking timeline after session switches, a cleanup pass resurfaces old sessions, auto-compression reference cards leak into the active turn, or a stale compressed task resumes when the user intended a fresh chat.

This issue tracks a consistency contract for those layers. The goal is to stop treating each new symptom as a one-off patch and instead define the invariants every future streaming, recovery, replay, compression, and sidebar PR must preserve.

## Why this is not just an umbrella bug

This is not a bucket for every WebUI issue. It should produce a reviewable contract that answers concrete product and implementation questions:

- Which state layer is the source of truth for a running turn?
- When is it acceptable for the visible transcript and model context to differ?
- What must survive refresh, session switch, WebUI restart, and stale-stream cleanup?
- What is replay allowed to reconstruct, and what must remain a live-only presentation cache?
- How should automatic compression be represented so it helps the agent without visually polluting the user's active turn?
- What does a future PR need to prove before it changes streaming, recovery, context reconstruction, compression, or sidebar timestamps?

The expected output is a short RFC/checklist that reviewers and future PR authors can follow.

## Relationship to #1925

#1925 defines the larger direction: WebUI should become a thin observability/control client over a clearer Hermes Agent runtime boundary.

This issue is narrower and more immediate. It documents the state consistency rules that must hold both before and after that runtime boundary exists. In other words:

- #1925 is about where execution ownership should move.
- This issue is about keeping transcript, context, stream, replay, compression, and UI state coherent while WebUI still has multiple overlapping state layers.

## Initial issue map

These examples show the repeated boundary pattern:

| Example | State boundary exposed |
|---|---|
| #2341 / #2342 | Active reattach could show agent activity without the pending user turn that started it. |
[...]

PR #4216, opened 2026-06-15
Fix state-db user prompts before sidecar tails

dso2ng:
## Summary
- Preserve state.db-only user prompts that fall before a newer WebUI sidecar tail.
- Insert the missing user turn chronologically instead of dropping it as old replay.
- Keep duplicate/replay/watermark guards intact for already-visible sidecar rows.

## Related issues
- Related to #2361 run-state/session consistency.
- Broader architecture context: #1925.

## What this does not do
- Does not rewrite unified storage or change session persistence ownership.
- Does not change Desktop remote or Gateway behavior.
- Does not include any private transcript/session content.

## Test plan
- `python -m pytest -q tests/test_session_lost_response_regression.py::test_state_db_only_user_prompt_before_sidecar_tail_is_reinserted_chronologically` (RED before fix)
- `python -m pytest -q tests/test_session_lost_response_regression.py tests/test_webui_state_db_reconciliation.py tests/test_issue3346_legacy_dedup.py tests/test_issue3831_watermark_clear.py`
- `git diff --check`
- non-ASCII added-line guard -> `non_ascii_added_lines=0`


nesquena-hermes:
Thanks @dso2ng — the core idea here (rescuing a state.db-only user prompt that falls before a newer sidecar tail instead of dropping it) is sound and net-positive for the common case. But an independent deep review (Codex + Opus, both run because this edits `merge_session_messages_append_only` — the most display- and model-facing function in the repo) **reproduced several ordering failures** in `_insert_state_message_chronologically`, because the insert is timestamp-only and neighbor-blind. These need fixing before it can ship:

### Ship-blocking

1. **Equal-timestamp inversion (answer-before-question).** The insert only goes before a *strictly-greater* timestamp and otherwise appends, so an equal-timestamp user row lands *after* an existing assistant tail. Reproduced: sidecar `[assistant "answer"@100.0]` + state user `"the question"@100.0` → `[assistant, user]` — the answer renders before the question. Same-second turns are common, so this is the most likely case to hit in production, and it's the exact inversion the function's own comments exist to prevent.
   **Fix:** role-aware tie-break — treat an equal-timestamp *assistant* as ordered after the incoming user:
   ```python
   if existing_timestamp is not None and (
       existing_timestamp > timestamp
       or (existing_timestamp == timestamp and existing.get("role") == "assistant")
   ):
   ```

2. **Splits a `tool_use` / `tool_result` pair.** The chronological insert can drop the rescued user row immediately before a `role == "tool"` result that belongs to the preceding assistant's `tool_calls`, breaking the pair (model-facing). 
   **Fix:** in `_insert_state_message_chronologically`, when the candidate slot's `existing.get("role") == "tool"` and the prior row is an assistant with `tool_calls`, keep scanning forward past the tool result(s) before inserting. (Combine with #1 so the equal-ts case also doesn't split a pair.)

3. **Resurrection on the no-watermark compression path.** When a `truncation_watermark` is present, a pre-watermark ghost row is correctly dropped by the stale guard before reaching the new path. But compression doesn't always set a watermark — and the non-`prefer_context` display path (`_previous_messages`, `streaming.py`, the `routes.py` display reads) uses `session.messages` as the sidecar with no delta trimming, so a previously-compressed-out user row can be re-surfaced.
   **Fix:** confirm whether compression sets a watermark; if not, either gate the new insert so it doesn't fire on the non-`prefer_context` path, or require the rescued `timestamp` to fall within an existing user→assistant gap rather than anywhere `<= max_sidecar_timestamp`. (Minor boundary nit while you're there: both watermark guards use strict `<`/`>`, so a row with `timestamp == watermark` slips through.)

### Tests
The current happy-path test only exercises clean, strictly-increasing, distinct timestamps. Please add RED-then-green regressions for: equal-timestamp user-before-assistant ordering, a `user → assistant(tool_calls) → tool` sequence (no pair split), and a no-watermark compression case (no resurrection).

Everything else checks out — the full suite passes (9055), and the watermark-present path is correctly protected. Once the ordering/neighbor-awareness and the no-watermark gate are in, this is a good fix. Leaving the PR open for your update.

dso2ng:
Thanks for the detailed review — addressed in the latest push (`1070c2c0`).

Changes made:

- Added role-aware equal-timestamp ordering so a rescued `role=user` state.db row lands before an equal-timestamp assistant tail instead of rendering answer-before-question.
- Made the chronological insertion neighbor-aware for assistant `tool_calls` + following `role=tool` rows, scanning past the tool result burst before inserting the rescued user prompt.
- Added a no-watermark compression/display guard: if the only chronological slot is before the sidecar/context begins, the state.db-only user row is treated as an old compacted/pre-sidecar row and is not resurrected. Equal-timestamp assistant ties are still allowed.

Added RED-then-green regressions for all three cases in `tests/test_session_lost_response_regression.py`:

- `test_state_db_only_user_prompt_equal_timestamp_stays_before_assistant_tail`
- `test_state_db_only_user_prompt_does_not_split_tool_call_and_result_pair`
- `test_state_db_only_user_prompt_before_sidecar_start_is_not_resurrected_without_watermark`

Verification run locally:

```text
/home/dso2ng/.hermes/hermes-agent/venv/bin/python -m pytest -q \
  tests/test_session_lost_response_regression.py \
  tests/test_webui_state_db_reconciliation.py \
  tests/test_issue3346_legacy_dedup.py \
  tests/test_issue3831_watermark_clear.py

54 passed, 1 warning
```

I also merged latest `origin/master` after #4218 shipped upstream; the PR is now mergeable and the GitHub checks are green.

nesquena-hermes:
Re-reviewed after `40654476 "fix: harden state-db prompt ordering"`. All three ship-blockers from the prior pass are addressed, with the right logic, and each has a RED-then-green regression. Walking the updated `_insert_state_message_chronologically` (`api/models.py:4479-4517`) against the new tests:

**1. Equal-timestamp inversion — fixed.** The role-aware tie-break is exactly the shape suggested:

```python
should_insert = existing_timestamp is not None and (
    existing_timestamp > timestamp
    or (existing_timestamp == timestamp
        and msg.get("role") == "user"
        and existing.get("role") == "assistant")
)
```

`test_..._equal_timestamp_stays_before_assistant_tail` → `[question, answer]`. Correct.

**2. tool_use/tool_result split — fixed.** The forward-scan past `role=="tool"` rows that follow an assistant with `tool_calls` keeps the pair intact:

```python
while (idx < len(messages) and messages[idx].get("role") == "tool"
       and idx > 0 and messages[idx-1].get("role") == "assistant"
       and messages[idx-1].get("tool_calls")):
    idx += 1
```

`test_..._does_not_split_tool_call_and_result_pair` lands `missing prompt` after the tool result, before `final answer`. Correct.

**3. No-watermark resurrection — fixed, and the key-marking is now correctly gated.** The function returns `False` when the only slot is index 0 ahead of a strictly-newer sidecar start, and the call site (`models.py:4721`) now only marks `seen_*` keys when the insert succeeds — so a refused row is fully skipped, not half-tracked:

```python
if _insert_state_message_chronologically(merged_messages, msg):
    seen_message_keys.add(key)
    ...
```

`test_..._not_resurrected_without_watermark` correctly drops the `@100` row ahead of the `@200/@201` sidecar.

## One tradeoff worth a code comment (not blocking)

The resurrection guard keys on "slot is index 0 with a strictly-greater neighbor" as the signal for pre-context. That cleanly separates the two cases in the tests — the legitimate rescue (`test_..._before_sidecar_tail`) falls in the *middle* because the sidecar still holds the older `[user, assistant]` turns, while the compressed-out row falls *before everything*. But it also means the most-stripped interruption topology — sidecar = only the newer recovery tail (`[assistant@103]`), state = the missing `user@102` with no surviving preceding context in the sidecar — would be dropped at idx 0, even though that user prompt is a genuine rescue, not a resurrection. From timestamps + position alone those two are indistinguishable, so choosing to drop (avoid resurrecting model-facing content) over insert (risk reviving compacted rows) is the correct conservative call. Worth a one-line comment at the `idx == 0 ... return False` branch noting it intentionally favors no-resurrection over rescue in the no-preceding-context case, so a future reader doesn't "fix" it back into a resurrection bug.

Minor edge (low priority): an equal-timestamp resurrection (state `user@200` sharing a timestamp with a compacted `assistant@200` at idx 0) slips past the guard — `existing_timestamp > timestamp` is false for equal ts, so it inserts. Vanishingly unlikely (compaction summary sharing an exact timestamp with a compressed-out user prompt), but if you want belt-and-suspenders, the idx-0 guard could also refuse the equal-ts case when there's no earlier context.

Net: the ordering and neighbor-awareness work is done and the tests pin all three failure modes. From my side this is shippable; the two notes above are comments/edge-hardening, not blockers. Nice turnaround.

dso2ng:
Small CI note after the latest refresh (`9392be46`): the PR is mergeable, but the `test (3.12, 2)` shard had one red test:

```text
tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata
```

This does not appear related to this PR's diff. I re-ran the exact test and the whole `tests/test_gateway_sync.py` file locally on the refreshed PR head:

```text
python -m pytest -q tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata
# 1 passed

python -m pytest -q tests/test_gateway_sync.py
# 58 passed, 1 warning
```

I tried to rerun the failed GitHub Actions job, but GitHub requires repository admin rights for that rerun. If a maintainer can rerun the failed shard, I can follow up if it reproduces.

nesquena-hermes:
Re the CI note on `9392be46`: I'd push back gently on "does not appear related to this PR's diff" before asking for a rerun. The failing shard test does import the file this PR changes.

`tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata` exercises `Session` from `api/models.py`:

```python
def test_sessions_response_backfills_imported_messaging_source_metadata(cleanup_test_sessions):
    from api.models import Session          # tests/test_gateway_sync.py:1226
    ...
    s = Session(session_id=sid, title='Legacy Imported Weixin',
                messages=[{'role': 'user', 'content': 'hello', 'timestamp': time.time()}], ...)
    s.is_cli_session = True
    s.save(touch_updated_at=False)
    data, status = get('/api/sessions')
    session = next(item for item in data.get('sessions', []) if item.get('session_id') == sid)
    assert session.get('source_tag') == 'weixin'
```

And this PR's `api/models.py` change adds a new chronological-insert branch inside the message-merge path:

```python
if msg.get("role") == "user" and _session_message_content_key(msg) not in seen_content_keys:
    if _insert_state_message_chronologically(merged_messages, msg):
        ...
    continue
```

The test builds a session with a single state.db-only `user` row and then reads it back through `/api/sessions`, which is exactly the merge path your new `_insert_state_message_chronologically` now intercepts. So a failure here is *plausibly* a real interaction with the diff, not noise — the single-user-row, no-sidecar shape is the one your new early-return guard (`if idx == 0 and existing_timestamp > timestamp: return False`) is reasoning about.

That said, you reported it passes locally (both the single test and the full `test_gateway_sync.py` file, 58 passed). That points at one of two things:

1. **Test-ordering / shared-state flakiness on shard 2** — `test_gateway_sync.py` writes into a shared state.db (`_ensure_state_db`, `_insert_gateway_session`) and toggles `show_cli_sessions` global settings. If another test on the same shard leaves `show_cli_sessions` or a `gw_*` row behind, the `next(... if item.get('session_id') == sid)` lookup can raise `StopIteration`. That's a pre-existing isolation issue, not your diff.
2. **A genuine ordering change** from the new branch that only manifests under the shard's specific test sequence.

To tell them apart without admin rerun rights, run the shard's test set *in the CI order* locally rather than the file alone, e.g. reproduce shard 2's split and run `test_gateway_sync.py` after the tests that precede it on that shard. If it stays green in-order too, it's worth adding `cleanup_test_sessions` coverage or an explicit `show_cli_sessions` reset to harden the isolation. If it goes red in-order, the new merge branch is reordering the single user row in a way `/api/sessions` surfaces.

Either way, "rerun the shard" only papers over (1); I'd nail down which case it is first so a green merge isn't hiding a latent ordering bug.

dso2ng:
Refresh update: I merged the branch with current `origin/master` to clear the conflict, preserving the state.db-only user-prompt reconciliation change and the Unreleased changelog entry.

Updated head: `1b0c639b`

Local verification on the refreshed head:

- `python3 -m py_compile api/models.py tests/test_session_lost_response_regression.py tests/test_gateway_sync.py`
- `./scripts/test.sh tests/test_session_lost_response_regression.py tests/test_webui_state_db_reconciliation.py tests/test_issue3346_legacy_dedup.py tests/test_issue3831_watermark_clear.py tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata -q` → 55 passed
- `./scripts/test.sh tests/test_gateway_sync.py -q` → 61 passed
- `git diff --check`

GitHub checks are now green and the PR is mergeable (`CLEAN`).

nesquena-hermes:
Shipped in **v0.51.488** (Release QW) 🚀 — thanks @dso2ng! The chronological-insert rescue is in. On absorb I hardened the equal-timestamp edge cases the full gate surfaced: the insertion advance is now a fixpoint loop that (a) skips past the *entire* contiguous tool-result block of a preceding assistant(tool_calls) — so a multi-tool turn can't be split — and (b) never lands at a slot whose left neighbour shares the message's role, so an already-answered turn can't be re-ordered. Added 3 equal-timestamp regression tests (all toothless-checked). Full gate green (Codex confirmed the insert path monotonic, Opus SHIP, suite 9401). Closing as merged.

Issue #498, opened 2026-04-14
feat(storage): unified session storage — share SessionDB between WebUI and CLI


nesquena-hermes:
## Summary

The WebUI and the CLI agent use separate session stores. The WebUI reads CLI sessions from `state.db` as read-only imports, but edits made in the WebUI (rename, archive, pin, tag) don't write back. There are effectively two session records for the same conversation.

## Proposed behaviour

Migrate WebUI session persistence to share the same SQLite `SessionDB` that the CLI uses, so that both surfaces read and write the same record. No "import" step needed; all sessions are first-class everywhere.

## Implementation notes

- Current WebUI store: JSON files in `~/.hermes/webui-public/sessions/` via `api/models.py`
- Target: SQLite `state.db` via the `SessionDB` class from hermes-agent
- Migration: existing JSON session files need to be ingested into SQLite on first run
- This is a significant architectural change; a transition flag or gradual migration is appropriate
- Also unlocks: sharing tool_calls, pending approvals, and context across surfaces

Tracked in ROADMAP.md as "Unified session storage (PR #75)".

nesquena-hermes:
## Acknowledged: Unified session storage (shared SessionDB) ✅

This is a significant architectural change but the right long-term direction — two separate session stores is a maintenance burden and creates the dual-record problem described here.

### Implementation notes

**Current state**:
- WebUI sessions: JSON files in `~/.hermes/webui-public/sessions/` via `api/models.py`
- CLI sessions: SQLite `state.db` via `SessionDB` from hermes-agent
- WebUI reads CLI sessions as read-only "imports" — edits (rename, pin, tag) don't write back

**Migration path** (suggested):
1. Add a migration flag `unified_sessions: true` in `config.yaml` — opt-in initially
2. On startup with the flag: `SessionDB.ingest_json_sessions(webui_sessions_dir)` — one-time import of existing JSON files into SQLite
3. `api/models.py` becomes a thin adapter over `SessionDB` instead of the JSON file store
4. The JSON files are archived to `sessions.bak/` after successful migration

**Key surface areas**:
- `api/models.py` — all read/write ops redirect to `SessionDB`
- Session metadata (pin, tag, archive) needs corresponding columns in `SessionDB` if not already present
- WebUI-specific fields (UI state, per-session settings) that don't exist in the CLI schema need additive columns — `ALTER TABLE sessions ADD COLUMN ...`

[...]

PR #4218, opened 2026-06-15
Fix imported lineage sidebar discoverability

dso2ng:
## Summary
- Preserve multiple messageful imported/CLI non-snapshot rows that share a lineage with a fuller pre-compression snapshot.
- Still expose the fuller snapshot when it has the larger transcript.
- Keep the existing single inactive-continuation replacement behavior intact.

## Related issues
- Related to #498 unified/session-storage discoverability context.
- Related to #3418 sidebar source/filtering UX context.
- Adjacent to #3988 CLI session discoverability defaults, but this PR is specifically about backend lineage suppression.

## What this does not do
- Does not rewrite unified storage.
- Does not change CLI/TUI source classification; #4213 covers TUI-origin sidebar classification separately.
- Does not include any private transcript/session content.

## Test plan
- `python -m pytest -q tests/test_session_discoverability_invariants.py::test_fuller_snapshot_does_not_hide_multirow_imported_cli_lineage` (RED before fix)
- `python -m pytest -q tests/test_session_discoverability_invariants.py tests/test_session_index.py tests/test_session_sidebar_cache.py tests/test_issue3238_orphaned_cli_sidecar_prune.py tests/test_issue3586_cli_session_source_label.py`
- `git diff --check`
- non-ASCII added-line guard -> `non_ascii_added_lines=0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Session sidebar discoverability now preserves multi-row imported/CLI lineages when a fuller pre-compression snapshot exists.** The sidebar may still expose the fuller snapshot, but it no longer suppresses multiple messageful non-snapshot rows from the same imported/CLI lineage as if they were a single inactive continuation. (#498, #3418)
+
 ## [v0.51.424] — 2026-06-15 — Release OK (visible-message tail pagination, #4069)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in v0.51.429 (Release OP) via #4232. Thanks @dso2ng — multiple messageful imported/CLI sessions sharing a lineage now stay discoverable in the sidebar instead of being collapsed behind a single fuller snapshot. Codex + Opus + full suite (9056) all green; confirmed the guard only loosens hiding (can't make a session vanish) with no change to the single-continuation case. Closing as shipped.

Issue #272, opened 2026-04-12
Real-time bidirectional sync between Gateway sessions and WebUI


bergeouss:
### Problem

When Hermes Agent runs with a gateway (Telegram, Discord, etc.), the WebUI currently offers **read-only snapshot access** to gateway sessions via the CLI session bridge. There is no real-time synchronization between the two:

- Messages sent via Telegram don't appear live in the WebUI
- Replying from the WebUI to a gateway session is not possible
- The CLI bridge imports a static snapshot that never refreshes

This means users who primarily interact via messaging platforms can't use the WebUI as a full interface to participate in those same conversations.

### Current Architecture

**Gateway side** (hermes-agent):
- Dual-layer storage: in-memory `SessionStore` + SQLite `state.db` (WAL mode)
- `state.db` has `sessions` and `messages` tables with FTS5
- Gateway API server on port 8642 (aiohttp)
- Sessions tagged with `source` field: 'telegram', 'discord', 'cli', etc.
- Session keys: `agent:main:telegram:dm:<chat_id>`

**WebUI side** (hermes-webui):
- Own session store: JSON files in `~/.hermes/webui/sessions/`
- CLI bridge in `api/models.py`: `get_cli_sessions()` reads metadata from `state.db`, `get_cli_session_messages()` reads messages
- `api/state_sync.py`: WebUI → agent direction only (for /insights cost tracking)
- SSE streaming for agent responses (per-stream queue, not persistent)

### Proposed Approach

**Phase 1: Gateway → WebUI (read-only live view)**

1. Add a background watcher in the WebUI (`api/session_watcher.py`) that polls `state.db` every 2-3 seconds for new messages/sessions
2. Push detected changes to the browser via a persistent SSE endpoint (`/api/gateway/stream`)
3. Gateway sessions appear live in the sidebar with real-time message updates (similar to how Telegram works in the WebUI)

This requires **no changes to hermes-agent** — the WebUI already has read access to `state.db` and the gateway uses WAL mode so reads are non-blocking.

**Phase 2: WebUI → Gateway (bidirectional)**

1. WebUI connects to the gateway's API (port 8642) as a client
2. When replying to a gateway session in the WebUI, the message is forwarded to the gateway for processing
3. The gateway streams the response back to the WebUI via the existing API
4. Gateway sessions become fully interactive in the WebUI — not just viewable, but usable

This requires **gateway API changes** to support an external client sending messages to existing sessions.

### Benefits

[...]

PR #4221, opened 2026-06-15
fix(sessions): include target session in session events

dso2ng:
## Summary

Session-list invalidation events can now carry the specific `session_id` that changed, while keeping legacy broad refresh semantics safe.

- Adds optional `session_id` to `publish_session_list_changed(...)` and `/api/sessions/events` payloads.
- Keeps profile scoping backward-compatible and widens coalesced queue events when multiple sessions/profiles are involved.
- Passes session identity from common WebUI session mutations such as rename, duplicate, branch, pin, archive, move, and hidden-empty session materialization.
- Parses the optional `session_id` in the browser and labels active-session-targeted events without changing the broader active-reload policy.

## Related

Related to #2361 and #272. Adjacent to #4208: that PR owns the active WebUI transcript refresh behavior; this PR only makes the session event payload precise enough for follow-up refresh/revision policy.

## What this does not do

- Does not rewrite unified session storage.
- Does not change Hermes Agent runtime ownership.
- Does not change Desktop/TUI/gateway behavior.
- Does not include private session data.

## Tests

```bash
/home/dso2ng/.hermes/hermes-agent/venv/bin/python -m pytest -q \
  tests/test_session_events.py \
  tests/test_session_events_http_integration.py::test_session_events_bus_subscribe_unsubscribe_balance \
  tests/test_webui_external_refresh_frontend.py \
  tests/test_session_title_regenerate_controls.py::test_regenerate_helper_persists_generated_title_and_publishes_sidebar_refresh \
  tests/test_issue3225_rename_sync.py
# 23 passed, 1 warning

/home/dso2ng/.hermes/hermes-agent/venv/bin/python -m pytest -q \
  tests/test_attention_session_events.py \
  tests/test_session_sidebar_cache.py \
  tests/test_claude_code_session_import.py \
  tests/test_issue3987_imported_session_titles.py \
  tests/test_issue3225_rename_sync.py \
  tests/test_session_title_regenerate_controls.py
# 36 passed, 1 warning

/home/dso2ng/.hermes/hermes-agent/venv/bin/python -m py_compile api/session_events.py api/routes.py
git diff --check
non_ascii_added_lines=0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Session-change SSE events can now carry the specific session id that changed.** The event bus preserves optional `session_id` metadata for targeted mutations and safely widens coalesced events back to a profile/all-session refresh when multiple sessions or profiles are involved, giving the browser a precise signal without narrowing legacy refresh behavior. (#2361, #272)
+
 ## [v0.51.426] — 2026-06-15 — Release OM (custom-provider model-prefix routing fix, #4210)
 
 ### Fixed
--- a/api/routes.py
+++ b/api/routes.py
@@ -43,17 +43,28 @@ from api.session_events import (
 logger = logging.getLogger(__name__)
 
 
-def _publish_session_list_changed(reason: str, *, profile: str | None = None) -> None:
-    """Publish profile-scoped session changes while tolerating legacy test doubles."""
-    if not profile:
+def _publish_session_list_changed(
+    reason: str,
+    *,
+    profile: str | None = None,
+    session_id: str | None = None,
+) -> None:
+    """Publish scoped session changes while tolerating legacy test doubles."""
+    if not profile and not session_id:
         publish_session_list_changed(reason)
         return
     try:
-        publish_session_list_changed(reason, profile=profile)
+        publish_session_list_changed(reason, profile=profile, session_id=session_id)
     except TypeError:
         # Some focused tests monkeypatch the route-level publisher with the
-        # historical one-argument shape. Preserve the old signal instead of
+        # historical one-argument or profile-only shape. Preserve the old signal instead of
         # turning unrelated session mutations into 500s.
+        if profile:
+            try:
+                publish_session_list_changed(reason, profile=profile)
+                return
+            except TypeError:
+                pass
         publish_session_list_changed(reason)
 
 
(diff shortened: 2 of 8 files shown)


nesquena-hermes:
Pulled the branch and read `api/session_events.py`, the propagation sites in `api/routes.py`, and the `static/sessions.js` parse. The payload threading is correct and the coalescer widening is the conservative direction. Two things confirmed + one forward-looking note for the #4208 follow-up that consumes this signal.

## Coalescing widening is safe, and the `pending is None` guard is intact

`_coalesced_sessions_changed_payload` keeps the master fast-path (`if pending is None: return incoming`, `session_events.py:72`) and only strips `session_id` when two same-profile events target *different* sessions:

```python
if pending_session_id == incoming_session_id:
    return incoming
merged = dict(incoming)
merged.pop("session_id", None)
return merged
```

That's right — a maxsize=1 queue can't carry two distinct session targets, so dropping to a profile-scoped (or fully unscoped, when profiles differ) refresh is the fail-safe direction. No risk of a missed refresh, only of a less precise one.

## Delete correctly stays profile-scoped

`session_delete` at `routes.py:8572` intentionally does *not* pass `session_id`, while rename/pin/archive/move/branch/duplicate all do. That's the correct asymmetry: a deleted session id as an "active target" signal would be meaningless to the follow-up reload policy (the row is gone), so a broad profile refresh is what you want there. Worth a one-line comment at that call site so a future contributor doesn't "helpfully" add `session_id=sid` and reintroduce a stale-target signal.

## Forward note for #4208: absence of `session_id` must mean "refresh broadly", not "skip"

`_sessionEventTargetsActiveSession` (`sessions.js:3744`) returns `false` when the payload has no `session_id`, and today that only chooses the `'event'` vs `'event-active-session'` *label* — no behavior change, as the PR states. The thing to pin down before #4208 builds the active-transcript reload on this: because the coalescer drops `session_id` under contention, an event that *did* originate from the active session can arrive with no `session_id` after being merged with another same-profile session's event. So if #4208's policy is "reload the open transcript only when `session_id` matches, otherwise skip," a coalesced event would cause a **missed** active-transcript refresh exactly under load (the case coalescing exists for). The safe contract is: `session_id` present + match → targeted reload; `session_id` absent → fall back to the existing broad refresh. The current `false`-on-absent return is fine for the label, but #4208 should not invert it into a skip.

Net: payload shape, backward-compat fallback ladder in `_publish_session_list_changed` (`routes.py:46-68`), and the conservative coalescing all check out. The frontend change is label-only as advertised. Just make sure the consumer treats missing `session_id` as "broad," not "not-active."

nesquena-hermes:
Shipped in v0.51.430 (Release OQ) via #4233. Thanks @dso2ng — session-list-changed events now carry the changed session_id (optional, fully backward-compatible), and the frontend uses it to refine the refresh signal for the active session. Codex + Opus + full suite all green; confirmed backward-compat, profile scoping, and no new SSE exposure class. Closing as shipped.

PR #4006, opened 2026-06-11
fix(scroll): guard scroll pin hysteresis + add auto_scroll_follow setting

Pythonming2020:
## Problem

After streaming completes, the viewport snaps to the bottom even when the user has scrolled up to read the answer. Two root causes:

1. **`_scrollAfterMessageRender` pin hysteresis** — When `preserveScroll=true` and `_scrollPinned=true` (re-set by the 250px near-bottom hysteresis counter), `_followMessagesAfterDomReplace()` forces `scrollToBottom()`, overriding the user's position. The `_messageUserUnpinned` flag is not checked.

2. **`_finishDone` unconditional `scrollToBottom()`** — The explicit `scrollToBottom()` fires when `shouldFollowOnDone` is true, but that flag uses a 120px near-bottom check. Users reading even slightly above that threshold get snapped to the bottom.

### Bug fixes (2 lines)

**`static/ui.js` L7970** — Added `!_messageUserUnpinned` guard:
```diff
-if(_followMessagesAfterDomReplace()) return;
+if(!_messageUserUnpinned && _followMessagesAfterDomReplace()) return;
```
When the user explicitly scrolled up, skip forced follow and restore their scroll snapshot instead.

**`static/messages.js` L3302** — Added `_isMessagePaneNearBottom(250)` gate:
```diff
-if(shouldFollowOnDone&&typeof scrollToBottom==='function') scrollToBottom();
+if(shouldFollowOnDone&&typeof scrollToBottom==='function'&&typeof _isMessagePaneNearBottom==='function'&&_isMessagePaneNearBottom(250)) scrollToBottom();
```
Only call explicit `scrollToBottom()` if the user is within 250px of the bottom.

### New setting: `auto_scroll_follow` (7-file pattern)

A new `auto_scroll_follow` boolean setting (default: **off**) that gates `scrollIfPinned()` and `_shouldFollowMessagesOnDomReplace()`. When disabled, the viewport never auto-scrolls during streaming — the user controls scroll position manually and uses the ↓ button to jump to bottom.

| File | Change |
|------|--------|
| `api/config.py` | `_SETTINGS_DEFAULTS` + `_SETTINGS_BOOL_KEYS` |
| `static/boot.js` | Both init paths (normal + error fallback) |
| `static/ui.js` | `scrollIfPinned()` early return + `_shouldFollowMessagesOnDomReplace()` guard |
| `static/index.html` | Checkbox in appearance settings section |
[...]

PR #3403, opened 2026-06-02
feat: add all sessions sidebar filter

noahbare:
## Summary
- Adds an `All sessions` source tab to the session sidebar so WebUI and CLI/imported sessions can be viewed together
- Preserves the existing `WebUI sessions` and `CLI sessions` source filters
- Persists/restores the new `all` source filter with the existing localStorage key
- Updates the CLI source-filter regression test coverage

## Test Plan
- [x] `npm run lint:runtime`
- [x] `python3 -m pytest tests/test_issue2351_cli_session_source_filter.py -q`

## Notes
No matching open PRs or issues were found for this feature before implementation.

PR #3122, opened 2026-05-29
feat: dynamic per-source session tabs with WebUI fallback

kristapskristaps:
## Summary

Replaces the hardcoded two-tab (WebUI / CLI) session filter in the sidebar with **dynamic per-source tabs** — All, CLI, Discord, Telegram, WebUI, and any other source that appears in the session list.

## Problem

When a Hermes setup uses multiple platforms (Telegram, Discord, CLI, WebUI), the sidebar only offered two tabs: WebUI and CLI. Telegram and Discord sessions were mixed into the WebUI tab, making it hard to navigate. Additionally, the new `_sourceKeyForSession()` returned an empty string for WebUI-native sessions (which lack source metadata), causing them to be grouped as "unknown" instead of "webui".

## Changes

- **`_sourceKeyForSession()`** — defaults to `'webui'` when session source fields are null (WebUI-native sessions), with a null-session guard
- **`_sessionSourceLabel()`** — generates dynamic labels from the source key (e.g. "All sessions (42)", "Telegram (15)")
- **`renderSessionListFromCache()`** — computes source counts from live sessions, renders tabs dynamically with "All" as the default, sorts tabs alphabetically
- **`_restoreSessionSourceFilter()`** — migrates old localStorage `'webui'/'cli'` values, clears stale format
- **Empty state** — generalized message for any active source filter

## Testing

- [ ] WebUI-native sessions appear under "WebUI" tab (not "unknown")
- [ ] Telegram, Discord, CLI sessions appear in their respective tabs
- [ ] "All" tab shows every session regardless of source
- [ ] Tab state persists across page refreshes
- [ ] Old localStorage filter values are migrated/cleared on first load
- [ ] No sessions are lost in the filter transition

PR #4222, opened 2026-06-15
fix(#3988): default show_cli_sessions to true

### LINKED ISSUE #3988: Default show_cli_sessions to true

Sanjays2402:
Closes #3988.

Flips the `show_cli_sessions` default from `False` to `True` so that sessions from the TUI, Telegram, Discord, CLI, and the Hermes One desktop app appear in the WebUI sidebar without users having to discover the toggle in Settings.

Three load-bearing spots had to change in lockstep, because the issue body's "one-line change" runs into the same default-mismatch trap as #4006: an existing user with no saved value would still see the feature OFF in the browser even after the config default flips.

- `api/config.py` — `_SETTINGS_DEFAULTS['show_cli_sessions'] = True`
- `static/boot.js` — hydrate as `s.show_cli_sessions !== false` instead of `!!s.show_cli_sessions` so an absent saved value defaults True
- `static/panels.js` — Settings checkbox renders checked when the setting is absent, matching the new default

`tests/test_issue3988_show_cli_sessions_default.py` pins all three spots, modeled directly on `tests/test_issue4006_auto_scroll_follow_default.py` since the failure mode is identical.

Verified `pytest tests/test_gateway_sync.py tests/test_session_sidebar_cache.py` (65 tests, all pass) — those exercise the toggle through explicit True/False POSTs, so only the absent-value default actually changes here. Larger filter redesigns in #3418 / #3403 / #3122 still land cleanly on a True default per the issue body.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -6505,7 +6505,7 @@ _SETTINGS_DEFAULTS = {
     "hide_empty_state_suggestions": False,  # hide the default new-chat suggestion buttons
     "show_tps": False,  # show tokens-per-second chip in assistant message headers
     "fade_text_effect": False,  # animate newly streamed words with a lightweight fade-in effect
-    "show_cli_sessions": False,  # merge CLI sessions from state.db into the sidebar
+    "show_cli_sessions": True,  # merge CLI sessions (TUI/Telegram/Discord/Hermes One/etc.) into the sidebar by default (#3988)
     "show_cron_sessions": False,  # surface cron sessions in the sidebar (subordinate to show_cli_sessions)
     "show_previous_messaging_sessions": False,  # show older Telegram/Discord/etc. reset segments
     "sync_to_insights": False,  # mirror WebUI token usage to state.db for /insights
--- a/static/boot.js
+++ b/static/boot.js
@@ -1848,7 +1848,7 @@ function applyBotName(){
     applyEmptyStateSuggestionPref();
     window._showTps=!!s.show_tps;
     window._fadeTextEffect=!!s.fade_text_effect;
-    window._showCliSessions=!!s.show_cli_sessions;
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Superseded by #4230, shipped as v0.51.427 (Release ON). Your !==false hydration idiom + the source-pinning test approach were exactly right and carried into the shipped version — thanks @Sanjays2402, you're credited as co-author. The shipped version adds the grandfather migration so established installs stay OFF while new installs get the default-on behavior.

PR #4027, opened 2026-06-12
Move paused cron jobs into collapsible sidebar section

TomBanksAU:
## Summary
- move paused scheduled jobs into a separate collapsible **Paused** section in the Tasks sidebar
- default the paused section to collapsed, but auto-expand it when the selected job is paused
- persist the paused-section collapsed state in `localStorage` and add regression coverage

Refs #4026.

## Testing
- `node --check static/panels.js`
- `node --check static/i18n.js`
- `python -m pytest tests/test_cron_paused_section.py tests/test_cron_needs_attention.py tests/test_issue2031_cron_once_visibility.py tests/test_settings_navigation_and_detail_refresh.py -q`

PR #4223, opened 2026-06-15
feat(#4026): collapse paused cron jobs into a section under the Tasks list

### LINKED ISSUE #4026: Move paused scheduled jobs into a collapsible section in the Tasks sidebar
# GitHub Issue Draft

## Title
Move paused scheduled jobs into a collapsible section in the Tasks sidebar

## Labels
`enhancement`, `ux`, `tasks`

## Feature Description
Paused scheduled jobs currently stay mixed into the main **Scheduled Jobs** list in the Tasks sidebar.

That means jobs the user is not actively using still take up the same visual priority as active jobs. As the number of jobs grows, the list becomes noisier and it is harder to scan the jobs that are still running.

## Motivation
The current flat list works for a small number of jobs, but it gets cluttered once a user has a mix of active and paused jobs.

A paused job usually matters less in day-to-day use than an active one. It should still be easy to find, inspect, and resume, but it does not need to stay permanently visible in the main list.

## Proposed Solution
Group paused scheduled jobs into their own **Paused** section in the Tasks sidebar.

Suggested behavior:

- Active and attention-needing jobs remain visible in the main section.
- Paused jobs move into a separate **Paused (N)** section.
- The paused section is collapsible/expandable.
- Default the paused section to collapsed so inactive jobs do not dominate the list.
- Remember the user’s collapse state so the UI behaves consistently between refreshes.
- Paused jobs should still support the same click/open/resume workflow they have today.

## UX Notes / Precedent
This follows a common inbox/sidebar pattern: keep active items visible, and move inactive items into a collapsible secondary group so the primary list stays easy to scan.

The key UX goal is not to hide paused jobs completely. It is to lower their visual weight while keeping them one click away.

## Acceptance Criteria
- [ ] Paused scheduled jobs are rendered in a separate **Paused** section in the Tasks sidebar.
- [ ] The paused section can be collapsed and expanded.
- [ ] The paused section is collapsed by default.
- [ ] The section label shows the paused job count.
- [ ] Active jobs remain directly visible outside the paused section.
- [ ] Clicking a paused job still opens its detail view normally.
- [ ] Resuming a paused job removes it from the paused section and returns it to the active list without requiring a page reload.
- [ ] The expand/collapse state persists across refreshes.
- [ ] The UI still behaves well when there are zero paused jobs.

## Alternatives Considered
- Keep a single flat list with status badges only. This keeps implementation simple, but does not reduce visual clutter.
- Hide paused jobs entirely behind a filter. This makes them less discoverable than necessary.
- Move all non-active states into one collapsed bucket. This is weaker because paused jobs and error/attention jobs should not have the same visibility priority.

## Implementation Notes
The current Tasks sidebar appears to render cron jobs as one flat list in `static/panels.js` via `loadCrons()` and already distinguishes paused jobs via `_cronStatusMeta(job)` returning `state: 'paused'` / `listClass: 'paused'`.

This suggests the feature could be implemented as a UI-only grouping change:
- split `_cronList` into visible/paused buckets during render
- add a small collapsible section wrapper for paused jobs
- persist section state in `localStorage`

Relevant files observed during investigation:
- `static/panels.js`
- `static/style.css`
- `static/i18n.js`

Sanjays2402:
Closes #4026.

Implements the UI grouping the maintainer outlined in the issue thread.

Active vs paused scheduled jobs are now split in `loadCrons()`: active items render straight into `#cronList`, paused items render inside a native `<details>`/`<summary>` section at the bottom labeled `Paused (N)`. Default collapsed, expand-state persisted in `localStorage` under `cron-paused-collapsed` so it survives refresh.

Following the maintainer's two edge-case calls in the issue:
- `_cronList` stays the single source of truth — `openCronDetail`, the new-run dot (`_cronNewJobIds`), and the detail-refresh `_cronList.find` lookup all keep working unchanged. Only the *render* is partitioned.
- Zero-paused case (AC#9) skips the section header entirely rather than showing `Paused (0)`.
- Resume path (AC#7) already calls `loadCrons()`, so a resumed job re-buckets into the active list on the next render — no partial-DOM patching needed.

The paused predicate reuses `_cronStatusMeta(job).state === 'paused'` (panels.js:336), so the section bucket matches the existing status pill exactly.

Header uses the existing `cron_status_paused` i18n string capitalized — every translation already ships it.

Light CSS in `style.css` for the section header (custom rotating triangle replacing the native marker; hover/transition match the existing `.cron-item` polish).

`tests/test_issue4026_collapsible_paused_crons.py` (7 tests) pins the partition, the `<details>` usage, the localStorage persistence, the default-collapsed contract, and the `_cronList` single-source-of-truth invariant. The broader `pytest -k 'cron or panels or task'` suite still passes (276 passed, 19 skipped).

### DIFF
--- a/static/panels.js
+++ b/static/panels.js
@@ -509,11 +509,19 @@ async function loadCrons(animate) {
       return;
     }
     box.innerHTML = '';
+    // Partition active vs paused so paused jobs don't drown the list (#4026).
+    // _cronList stays the single source of truth — only the render is split,
+    // which keeps openCronDetail, _cronNewJobIds, and detail refresh untouched.
+    const _activeJobs = [];
+    const _pausedJobs = [];
     for (const job of _cronList) {
+      const status = _cronStatusMeta(job);
+      (status.state === 'paused' ? _pausedJobs : _activeJobs).push({ job, status });
+    }
+    const _appendCronItem = (parent, { job, status }) => {
       const item = document.createElement('div');
       item.className = 'cron-item';
       item.id = 'cron-' + job.id;
-      const status = _cronStatusMeta(job);
       const isNewRun = _cronNewJobIds.has(String(job.id));
       const isAgentMode = !job.no_agent;
       const profileLabel = _cronProfileLabel(job.profile);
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.484** (Release QT) 🚀 — thanks @Sanjays2402! Paused cron jobs now collapse into a 'Paused (N)' section in the sidebar. Nathan approved the screenshots. Full ship gate green (Codex + Opus + the full suite, 9393 passing). This was the dedup winner over #4027 (zero new i18n keys — it reuses the already-translated `cron_status_paused` label). Two tiny maintainer touches folded in on merge: wrapped the localStorage read/write in try/catch (so a private-mode storage throw can't break the sidebar) + added the CHANGELOG entry. Closing as merged.

Issue #3696, opened 2026-06-06
bug:_sessionAttentionState is not defined


jdb110:
renderSessionList ReferenceError: _sessionAttentionState is not defined
    at _sidebarRowHasVisibleMessages (sessions.js?v=v0.51.287:4174:30)
    at _partitionSidebarSessionRows (sessions.js?v=v0.51.287:4187:9)
    at renderSessionListFromCache (sessions.js?v=v0.51.287:4250:5)
    at _applySessionListPayload (sessions.js?v=v0.51.287:2916:3)
    at _runRenderSessionListRefresh (sessions.js?v=v0.51.287:2945:5)
    at async _drainRenderSessionListQueue (sessions.js?v=v0.51.287:2953:7)
    at async boot.js?v=v0.51.287:1914:3

nesquena-hermes:
## Summary

Confirmed — this is a scope bug, not a load-order/caching issue. `_sidebarRowHasVisibleMessages` is a **module-level** function (`static/sessions.js:4173`) but it calls `_sessionAttentionState`, which is declared **nested inside** `renderSessionListFromCache` (`static/sessions.js:4527`, indented two spaces). Function hoisting only lifts a declaration to the top of its *enclosing* function, so the nested `_sessionAttentionState` is invisible to the sibling top-level `_sidebarRowHasVisibleMessages`, even though the call chain runs through `renderSessionListFromCache`. Hence the `ReferenceError: _sessionAttentionState is not defined` at row 4174.

## Code reference

`static/sessions.js:4173`:

```js
function _sidebarRowHasVisibleMessages(s, activeSidForSidebar){
  return (s.message_count||0)>0 ||
    _sessionAttentionState(s) ||          // <-- nested fn, not in scope here
    _isSessionEffectivelyStreaming(s) ||
    ...
```

But the definition is nested (note the 2-space indent), inside `renderSessionListFromCache` at `static/sessions.js:4527`:

```js
  // Note: declared after the groups loop but available via function hoisting.
  function _sessionAttentionState(s){
    const attention=s&&s.attention&&typeof s.attention==='object'?s.attention:null;
    ...
```

The comment ("available via function hoisting") is the trap — hoisting is scope-local, not global. `_isSessionEffectivelyStreaming` doesn't throw because it genuinely is top-level (`static/sessions.js:349`).

## Why it doesn't fire for everyone

[...]

PR #4224, opened 2026-06-15
feat(#4159): jump to matched message + flash on session-search content hit

### LINKED ISSUE #4159: feat(search): navigate to matched message + highlight keyword when clicking session search result
### Problem

The current session search (sidebar "Filter conversations..." input) finds matching sessions by title and message content, and shows a text preview snippet. However, clicking a search result only opens the session at the **latest message** — the user must manually scroll to find where the keyword appears.

This is a significant UX gap compared to other chat interfaces (e.g., DeepSeek official, ChatGPT) where clicking a search result jumps directly to the matched message and highlights the keyword.

### Current behavior

1. User types keyword in session search
2. Matched sessions appear in sidebar (title match or content match with preview)
3. User clicks a session → session opens at the latest message
4. User must scroll manually to find the matched message

### Proposed behavior

1. User types keyword in session search
2. Matched sessions appear in sidebar
3. User clicks a session → session opens **and scrolls to the specific matched message**
4. The keyword is **highlighted** in the matched message

### Implementation sketch

**Backend** (`api/routes.py` — `_handle_sessions_search`):
- Return `match_message_id` (the message that triggered the content match) alongside existing `match_type` and `match_preview`

**Frontend** (`static/sessions.js`):
- When opening a session from a search result click, pass the `match_message_id` to `loadSession()` or scroll to that message after load
- After scroll, highlight the search keyword in the matched message text

### Existing infrastructure

The search API already returns:
- ✅ `match_type`: "title" | "content"
- ✅ `match_preview`: truncated text snippet around the keyword
- ❌ Missing: `match_message_id` (message index or message ID)

The frontend already:
- ✅ Has `_contentSearchResults` and `_sessionSearchMergeMatches`
- ✅ Has `loadSession(session_id)` 
- ❌ No mechanism to pass a target message position

### Additional context

This is purely additive — no existing behavior changes. The search results sidebar rendering and session loading remain fully backward compatible.

Sanjays2402:
Closes #4159.

Implements the design the maintainer outlined in the issue thread — capture the matched-message index on the server, reuse the existing `_jumpToMessage` helper on click.

### Backend (`api/routes.py`)
`_handle_sessions_search` already locates the exact matching message during the content scan (it `break`s on the first hit) — until now the loop index was discarded. This adds `match_message_idx` on content-type results, captured from `enumerate(msgs)`. Title matches do **not** grow the field (nothing to jump to). The index is into `sess.messages`, which is the same array the renderer enumerates to stamp `msg-user-<rawIdx>` ids on user rows and `data-msg-idx` on assistant segments, so the DOM lookup on the client resolves.

### Frontend (`static/sessions.js`)
After `loadSession()` in the sidebar-row click handler, if the session carries `match_type==='content'` and an integer `match_message_idx`, call `_jumpToMessage(idx)` on the next tick.

Reuses `outline.js:94 _jumpToMessage` unchanged — it already handles both the in-window case (smooth `scrollIntoView` + `_flashRow`) and the out-of-window case (re-fetch `/api/session?...&messages=1&msg_limit=9999`, expand the render window, then retry). The dispatch is purely additive: a missing idx falls back to opening the session at the latest message, exactly as today, so this is backward compatible with older servers.

### Tests
`tests/test_issue4159_search_match_message_idx.py` (4 tests) pins:
- content match surfaces `match_message_idx` at the right raw enumerate index;
- the first-hit index is returned (preserves the existing `break()` semantics);
- title matches do **not** include the field;
- no-match case returns empty results, unchanged.

Full search test suite still green: 28 passed across `test_issue4159_*`, `test_sessions_search_depth_validation`, `test_sessions_search_profile_scope`, `test_session_search_bfcache_822`, `test_sidebar_search_highlights`.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -10143,10 +10143,14 @@ def _handle_sessions_search(handler, parsed):
             try:
                 sess = get_session(s["session_id"])
                 msgs = sess.messages[:depth] if depth else sess.messages
-                for m in msgs:
+                for raw_idx, m in enumerate(msgs):
                     c = _session_search_message_text(m)
                     if q in str(c).lower():
-                        item = dict(s, match_type="content")
+                        # #4159: surface the matched message index so the frontend
+                        # can jump to it (msg-user-<rawIdx>). The renderer stamps
+                        # rows with the same raw index it gets from sess.messages,
+                        # so this enumerate lines up with the DOM ids.
+                        item = dict(s, match_type="content", match_message_idx=raw_idx)
                         preview = _session_search_preview(c, q)
                         if preview:
                             item["match_preview"] = _redact_text(preview)
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -6283,6 +6283,15 @@ function renderSessionListFromCache(){
         try{
           if(($('sessionSearch').value||'').trim()) _hideSearchPreviewsAfterSelect=true;
           await loadSession(s.session_id);renderSessionListFromCache();
+          // #4159: if the user clicked a content-search hit, scroll the
+          // newly-loaded transcript to the matched message and flash it.
+          // _jumpToMessage handles the out-of-render-window case (it refetches
+          // the full session with msg_limit=9999) and the in-window case
+          // (smooth scrollIntoView + flash).
+          if(s.match_type==='content' && Number.isInteger(s.match_message_idx) && typeof _jumpToMessage==='function'){
+            const _jumpIdx=s.match_message_idx;
+            window.setTimeout(()=>{ try{ _jumpToMessage(_jumpIdx); }catch(_e){} }, 0);
+          }
           if(typeof closeMobileSidebar==='function')closeMobileSidebar();
         }finally{
           el.classList.remove('loading');
--- a/tests/test_issue4159_search_match_message_idx.py
+++ b/tests/test_issue4159_search_match_message_idx.py
@@ -1,0 +1,101 @@
+"""Backend regression for #4159: content-search results must surface the
+matched message index so the frontend can jump straight to the hit.
+
+The content scan in ``_handle_sessions_search`` already locates the exact
+message that contains the query (it iterates ``sess.messages`` and ``break``s
+on the first hit). Until now the loop index was discarded; this test pins the
+new ``match_message_idx`` field on content-typed results, indexed against the
+same raw ``sess.messages`` array the renderer stamps onto each row as
+``data-msg-idx`` / ``msg-user-<rawIdx>``.
+
+Title matches must NOT carry ``match_message_idx`` (there's no message-level
+hit to jump to).
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+from unittest.mock import patch
+from urllib.parse import urlparse
+
+
+def _run_search(query, *, session_messages, sessions_meta=None):
+    import api.routes as routes
+
+    meta = sessions_meta or [
+        {"session_id": "s1", "title": "Untitled", "profile": "default"}
+    ]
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
## The jump/flash never fires — `_jumpToMessage` isn't reachable from sessions.js

Thanks for this — the search-hit → jump-to-message + flash is a nice UX touch, and the flash itself is well-built (self-terminating, no stale highlight) and XSS-safe. But as written the feature is a **silent no-op in the browser**, so it needs a fix before it can ship.

### Root cause: cross-`<script>` scope mismatch
`outline.js` wraps its whole body in an IIFE (it ends with `})();`), so `_jumpToMessage` (defined at `outline.js:94`) is a **function-local**, not a global. The only thing exposed globally is `window._outlineJump` (`outline.js:229`).

`sessions.js` is loaded as a **separate `<script>`** (index.html:1536 vs outline.js at :1542), so it has no access to outline.js's IIFE-local scope. Its call site is:
```js
// sessions.js:6291-6293
if (s.match_type==='content' && Number.isInteger(s.match_message_idx) && typeof _jumpToMessage==='function') {
  window.setTimeout(()=>{ try{ _jumpToMessage(_jumpIdx); }catch(_e){} }, 0);
}
```
`typeof _jumpToMessage` is `'undefined'` in sessions.js's scope, so the guard is **always false** and the jump/flash never runs. (I confirmed this by replicating the multi-`<script>` global scope in a Node `vm` harness — the guard evaluates false.)

**Fix:** call the exposed global — `window._outlineJump` — in both the guard and the call:
```js
if (s.match_type==='content' && Number.isInteger(s.match_message_idx) && typeof window._outlineJump==='function') {
  window.setTimeout(()=>{ try{ window._outlineJump(_jumpIdx); }catch(_e){} }, 0);
}
```

### Two more correctness hazards to handle while you're in there
1. **Absolute vs window-relative index.** The backend `match_message_idx` is an absolute index into the full session, but the DOM `msg-user-<idx>` ids are assigned over the *rendered window* (`S.messages` after truncation). On a long/truncated session these don't line up, so even once the call fires it can scroll to the **wrong message**. Map the backend index to the rendered row (or have `_jumpToMessage` resolve via the message id rather than the positional index).
2. **Assistant-content hits silently miss.** The `msg-user-<idx>` id scheme only exists on user rows, so a content hit inside an assistant message has no target to jump to. Either add an id to assistant rows or handle the role in the jump.

### Verification note
Because this is a click→scroll→flash gesture, please verify the re-push with a real interaction (not just a screenshot): a **long/truncated session** + a **content hit inside an assistant message**, confirming it lands on the correct row.

### Side note for the maintainers (not this PR's job)
This slipped past `scripts/scope_undef_gate.py` because outline.js's IIFE body is unindented, so the gate's `^function` heuristic wrongly treats `_jumpToMessage` as a global — a false-negative in the brick-class gate (#3696 family). Worth a separate hardening pass on the gate.

Marking `changes-requested`. The feature concept is good and the fix is small — happy to re-gate once the call reaches `window._outlineJump` and the index mapping is correct.

nesquena-hermes:
Salvaging into a fresh PR — thanks @Sanjays2402. The search-hit → jump-to-message + flash is a nice UX touch and the flash itself is well-built (self-terminating, XSS-safe). The one issue was that the jump never actually fired: `_jumpToMessage` lives inside `outline.js`'s IIFE so it isn't in scope from `sessions.js` (cross-`<script>` mismatch), making it a silent no-op. We're porting it fresh on current master with that scope issue fixed, preserving your authorship — I'll link the new PR here. Closes #4159 once it lands. Appreciate the contribution.

PR #4144, opened 2026-06-13
feat: show the active workspace project-context file (AGENTS.md / HERMES.md / CLAUDE.md) in the Memory tab

mvanhorn:
## Summary

Show the active workspace project-context file (AGENTS.md / HERMES.md / CLAUDE.md) in the Memory tab

## Why this matters

The Memory tab currently surfaces exactly three profile-scoped files resolved from `HERMES_HOME` (`memories/MEMORY.md`, `memories/USER.md`, `SOUL.md`). Users want it to also surface the project/workspace context file the agent actually loads: `.hermes.md` / `HERMES.md`, `AGENTS.md`, `CLAUDE.md`, or `.cursorrules`. The maintainer confirmed this is a real structural gap and gave precise code references: the read endpoint hard-codes the three files at `api/routes.py:11392` (`_handle_memory_read`), the write endpoint allowlists only those three sections at `api/routes.py:15258`, and the frontend section list is `MEMORY_SECTIONS` at `static/panels.js:4072`. The agent-side discovery/priority logic (`build_context_files_prompt`) lives in the separate hermes-agent repo and is referenced for the priority order only; the WebUI must mirror that order in-repo rather than import it.

## Changes

Add a read-only (or workspace-write-guarded) `project_context` section. Server-side in `api/routes.py`: resolve files from the active workspace path (passed as `TERMINAL_CWD`, see `api/streaming.py:1253` / `api/routes.py:12549`), NOT from `get_active_hermes_home()`. Implement a small helper that mirrors the agent's first-match-wins priority (1. `.hermes.md`/`HERMES.md` walking to git root, 2. `AGENTS.md` cwd-only, 3. `CLAUDE.md` cwd-only, 4. `.cursorrules` cwd-only), returns the resolved path plus content, and reuses the existing `_redact_text` treatment (routes.py:11420). Wire the helper into `_handle_memory_read` so the new section returns the effective file (and, ideally, a muted "shadowed by X" note when a lower-priority file also exists). Add the `project_context` entry to `MEMORY_SECTIONS` in `static/panels.js`. Keep editing scope minimal: ship as read-only first (these files live in user repos, so a write path needs workspace-write guardrails distinct from the `HERMES_HOME` write path at routes.py:15244); if writing is included, gate it behind those guardrails rather than the existing memory write path. Mark partial_scope only if writing is deliberately deferred.

## Testing

- Happy path: a workspace with only `AGENTS.md` -> tab shows `AGENTS.md` content as the active project-context file.
[...]

Issue #3866, opened 2026-06-09
Show AGENTS.md in Memory tab


dctr:
The current memory tab shows (and allows to edit)

- memories/MEMORY.md
- memories/USER.md
- SOUL.md

Could you add project-root specific context files as well please? 🙏  

Depending on setup, that might be .hermes.md, HERMES.md, AGENTS.md or CLAUDE.md ; only one will be loaded, in the given order, see https://hermes-agent.nousresearch.com/docs/user-guide/features/context-files

nesquena-hermes:
## Summary

Good request, and it points at a real structural gap. The Memory tab today is wired to exactly three **profile-scoped** files in `HERMES_HOME`, while the context files you're asking for (`.hermes.md` / `HERMES.md` / `AGENTS.md` / `CLAUDE.md` / `.cursorrules`) are **project/workspace-scoped** and resolved from the agent's `cwd`, not from `HERMES_HOME`. So this isn't just "add a fourth section to the same reader" — the new section has to resolve against the active workspace, and it has to mirror the agent's first-match-wins priority so the tab doesn't show a file that isn't actually being loaded.

## Code reference

The read endpoint hard-codes the three files under the active profile home:

```python
# api/routes.py:11392  _handle_memory_read
home = get_active_hermes_home()
mem_dir = home / "memories"
mem_file = mem_dir / "MEMORY.md"
user_file = mem_dir / "USER.md"
soul_file = home / "SOUL.md"
```

The write endpoint only accepts those three sections (`api/routes.py:15258`):

```python
if section == "memory":   target = mem_dir / "MEMORY.md"
elif section == "user":   target = mem_dir / "USER.md"
elif section == "soul":   target = home / "SOUL.md"
else: return bad(handler, 'section must be "memory", "user", or "soul"')
```

and the frontend section list lives at `static/panels.js:4072`:

```javascript
const MEMORY_SECTIONS = [
  { key: 'memory', labelKey: 'my_notes', ... },
  { key: 'user',   labelKey: 'user_profile', ... },
  { key: 'soul',   labelKey: 'agent_soul', ... },
  { key: 'external_notes', ... },
];
```

The discovery logic you linked lives **agent-side**, in `agent/prompt_builder.py:1514` (`build_context_files_prompt`). The priority order is explicit there:

```python
# Priority (first found wins — only ONE project context type is loaded):
#   1. .hermes.md / HERMES.md  (walk to git root)
[...]

PR #4225, opened 2026-06-15
fix(#4164): bound project-context file walk in non-git workspaces

### LINKED ISSUE #4164: Bound the non-git project-context file walk (agent + WebUI parity) — follow-up from #4144
## Follow-up from #4144 / #3866 (project-context Memory tab, v0.51.406)

**Context:** The new "Project Context" Memory tab (#4144) mirrors the agent's `prompt_builder` discovery to show what the agent actually injects. During review, Codex flagged that in a **non-git workspace** the context-file search walks **up to filesystem root** (because the `stop_at = git_root` bound is `None` when there's no `.git`), so a workspace at `/tmp/x/project/subdir` can surface `/tmp/x/HERMES.md` — a file *above* the workspace.

**This is faithful parity, not a WebUI bug:** the agent's own `_find_hermes_md` (`agent/prompt_builder.py`) has the identical behavior — `for directory in [cwd, *cwd.parents]: ... if stop_at and directory == stop_at: break` never breaks when `stop_at is None`, so the agent ALSO walks to root in a non-git tree. Bounding it WebUI-side alone would make the tab misreport the effective context.

**Ask:** decide whether the agent's non-git walk *should* be bounded (e.g. stop at the workspace root, or only check `cwd` when no git root is found). If yes, fix it in `agent/prompt_builder._find_hermes_md` first, then mirror the same bound in the WebUI's `_project_context_candidates` (`api/routes.py`) so the two stay consistent. If the unbounded walk is intended, no change — just confirming the WebUI tab correctly reflects it.

Low priority: authenticated localhost surface, real context files are small, and the behavior matches the agent today. Filed per maintainer decision on the #4144 release.

Sanjays2402:
Closes #4164.

Fixes the WebUI half of the bound that the maintainer flagged on the #4144 / #3866 follow-up.

`_project_context_candidates` mirrored the agent's `_find_hermes_md` exactly, including the bug that `stop_at = git_root` is `None` in a non-git tree — so the walk

```python
for directory in [cwd, *cwd.parents]:
    ...
    if stop_at and directory == stop_at:
        break
```

never broke and ran all the way to filesystem root. A workspace at `/tmp/x/project/subdir` could surface `/tmp/x/HERMES.md` (a file outside the workspace) in the Memory → Project Context tab.

### Fix

When no git root is found, treat the workspace itself as the stop boundary. The cwd is still scanned (preserving in-workspace AGENTS.md / HERMES.md behavior), but parents are skipped — we never walk into the user's home directory or `/tmp`. Git workspaces are unchanged: they still walk up to the git root.

### Agent-side parity

The issue notes the agent's `agent/prompt_builder._find_hermes_md` should be bounded the same way. That lives in a separate repo (hermes-agent), so this PR fixes only the WebUI side. Until the agent ships the same bound, the WebUI tab will *under-report* context files that live above a non-git workspace, which is strictly less surprising than the previous over-report and lines up with the safer of the two options the issue laid out ("only check cwd when no git root is found").

### Tests

`tests/test_issue4164_bound_non_git_project_context_walk.py` (4 tests) pins:
- non-git workspace does **not** leak a parent `HERMES.md`
- non-git workspace **still** reads in-workspace context files
- git workspace walk-to-git-root is **unchanged**
- workspace-level `HERMES.md` wins, parent copy doesn't appear even shadowed

All 10 existing `test_3866_project_context_memory_section` tests still pass (14 total green).

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -12623,15 +12623,37 @@ def _project_context_git_root(start: Path) -> Path | None:
 
 
 def _project_context_candidates(workspace: Path) -> list[Path]:
-    """Mirror the agent's first-match project context file priority."""
+    """Mirror the agent's first-match project context file priority.
+
+    #4164: in a non-git workspace the agent's ``_find_hermes_md`` walks all
+    the way up to filesystem root because its ``stop_at = git_root`` is
+    ``None`` — so a workspace at ``/tmp/x/project/subdir`` could surface
+    ``/tmp/x/HERMES.md`` (a file *outside* the user's workspace) in the
+    Project Context tab.
+
+    The WebUI tab is a read-only mirror of what the agent injects, so the
+    safest bound that does not over-promise is: when there is no git root,
+    treat the workspace itself as the stop boundary. The cwd is still
+    scanned (preserving the in-workspace AGENTS.md / HERMES.md behavior),
+    but we no longer walk into the user's home directory or ``/tmp``.
+
+    The agent-side walk in ``agent/prompt_builder._find_hermes_md`` should
+    be bounded the same way for full parity; until that ships the WebUI
+    will under-report context files that live *above* a non-git workspace,
+    which is strictly less surprising than over-reporting them.
+    """
     cwd = workspace.resolve()
     candidates: list[Path] = []
-    stop_at = _project_context_git_root(cwd)
+    git_root = _project_context_git_root(cwd)
+    # When inside a git tree, walk up to the git root as before. When not,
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in v0.51.428 (Release OO) via #4231. Thanks @Sanjays2402 — the non-git project-context walk is now bounded at the workspace, so files above a non-git workspace no longer leak into the Project Context tab. Codex + Opus + full suite (9054) all green. Closing as shipped.

Issue #4217, opened 2026-06-15
/api/file/delete and /api/file/rename follow symlinks to the resolved target


claw-io:
## Summary

`/api/file/delete` and `/api/file/rename` resolve a workspace-relative
path through `safe_resolve(ws_root, body["path"])`, which follows the
final symlink, and then call `rmtree_anchored` / `unlink_anchored` /
`rename_anchored` on the **resolved target**. The result: deleting a
symlink in the workspace recursively removes the real directory it
points to, and renaming a symlink moves the real entry and leaves the
link dangling.

`_handle_file_move` already guards against this with
`(ws_root / body["path"]).is_symlink()` and returns
"Cannot move a symlinked entry". The delete and
rename paths do not have the equivalent guard.

This is in the backend and is reachable today by any HTTP client that
POSTs to those endpoints. It is not introduced by any renderer change.

## Environment

- Repo: `nesquena/hermes-webui`
- Branch: `upstream/master` (HEAD at the time of reporting)
- No special config required; default workspace + a hand-placed
  symlink in the workspace is enough.

## Steps to reproduce

```python
# Tested against current upstream/master with the project's own
# primitives. No HTTP server required; these are the same calls
# the production handlers make, in the same order.
import os, tempfile
from pathlib import Path
from api.workspace import (
    safe_resolve_ws, rmtree_anchored, unlink_anchored, rename_anchored,
)

with tempfile.TemporaryDirectory() as t:
    ws = Path(t) / "ws"
    ws.mkdir()
    real_dir  = ws / "real_target";  real_dir.mkdir()
    (real_dir / "precious.txt").write_text("KEEP ME")
    real_file = ws / "real_file.txt"; real_file.write_text("KEEP ME TOO")
    link_dir  = ws / "link_to_dir";   link_dir.symlink_to(real_dir)
    link_file = ws / "link_to_file.txt"; link_file.symlink_to(real_file)

    # Case 1: equivalent of POST /api/file/delete
    # body: {session_id, path: "link_to_dir", recursive: true}
    target = safe_resolve_ws(ws, "link_to_dir")
    print(target)               # -> ".../real_target"  (followed the symlink)
    rmtree_anchored(ws, target) # operates on the real dir
    # real_target is GONE; link_to_dir is now a dangling symlink.

    # Case 2: equivalent of POST /api/file/delete
    # body: {session_id, path: "link_to_file.txt"}
    target = safe_resolve_ws(ws, "link_to_file.txt")
[...]

PR #4227, opened 2026-06-15
fix(#4217): guard /api/file/delete and /api/file/rename against symlink targets

rodboev:
## Thinking Path

- `/api/file/delete` and `/api/file/rename` call `safe_resolve()`, which follows the final symlink via `.resolve()`; both handlers then operate on the real target, so deleting a workspace symlink deletes the pointed-to directory or file.
- The move handler already has the correct pattern: check `(ws_root / body["path"]).is_symlink()` on the lexically-requested path (lstat, no-follow) before any destructive op, and return a 400.
- Added the same two-line guard to both handlers immediately after the `exists()` 404 check, matching the move handler's wording and placement exactly.

## What Changed

- `api/routes.py`: added `(ws_root / body["path"]).is_symlink()` guard in `_handle_file_delete` (after the 404 check, before the `is_dir()` branch) and in `_handle_file_rename` (after the 404 check, before `new_name` extraction).
- `tests/test_workspace_symlink_delete_rename.py`: new file with 7 regression tests covering symlink-to-dir and symlink-to-file rejection for both handlers, real file/dir ops still working, and the existing move guard pinned.

## Why It Matters

Without the guard, any user who can create a workspace symlink can silently destroy files outside their intended scope by issuing a standard delete or rename API call against the symlink name; the fix brings delete and rename into parity with the already-hardened move handler.

## Verification

```
pytest tests/test_workspace_symlink_delete_rename.py -v --timeout=60
```

## Risks / Follow-ups

- Broken (dangling) symlinks already 404 today because `target.exists()` returns False; this PR does not change that behavior. Allowing explicit cleanup of broken links is a follow-up if desired.
- The guard uses `is_symlink()` on the lexically-joined path, not the resolved path, so it correctly detects in-workspace symlinks regardless of whether the workspace root itself is a symlink (e.g., macOS `/tmp -> /private/tmp`).

## Model Used

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -14725,6 +14725,8 @@ def _handle_file_delete(handler, body):
         target = safe_resolve(ws_root, body["path"])
         if not target.exists():
             return bad(handler, "File not found", 404)
+        if (ws_root / body["path"]).is_symlink():
+            return bad(handler, "Cannot delete a symlinked entry")
         if target.is_dir():
             if not body.get("recursive"):
                 return bad(handler, "Set recursive=true to delete directories")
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Absorbed into #4230, shipped as v0.51.427 (Release ON). Your symlink delete/rename guard shipped — thanks @rodboev, credited as co-author. The shipped version additionally reorders the is_symlink() check before exists() (so a dangling symlink is rejected 400 instead of 404'd + left undeletable) and applies the same reorder to the move handler for parity.

PR #3997, opened 2026-06-11
fix(#3988): default show_cli_sessions to true

rodboev:
## Thinking Path

- The sidebar already knows how to merge non-WebUI sessions, but the clean-install default keeps that path off until the user discovers a Settings checkbox.
- The route already caps visible CLI rows, so flipping the default no longer reintroduces the earlier full-scan concern.
- The boot fallback and gateway-probe path also need to agree with that new default, otherwise a transient settings load failure or a gateway-less install can silently hide CLI sessions again or show a confusing warning loop.

## What Changed

- `api/config.py`: change the default `show_cli_sessions` setting from `false` to `true`.
- `static/boot.js`: align the settings-load fallback with the new `show_cli_sessions=true` default.
- `static/sessions.js`: skip the gateway-unavailable fallback poll and toast when the sidebar has no imported CLI or messaging sessions yet.
- `tests/test_gateway_sync.py`: add regression coverage for the clean-default settings path and the boot/probe source contracts, while preserving the explicit-off case.

## Why It Matters

Users who split time across WebUI, TUI, CLI, Telegram, or desktop surfaces will see the same conversations by default instead of looking at an empty or incomplete sidebar on first use.

## Verification

- `pytest tests/test_gateway_sync.py -v --timeout=60 -k "test_gateway_sessions_appear_when_enabled or test_webui_state_db_session_without_sidecar_appears_when_agent_sessions_enabled or test_webui_state_db_session_without_sidecar_appears_when_show_cli_sessions_unset or test_show_cli_sessions_default_on_boot_and_gateway_probe_fallbacks_stay_aligned or test_gateway_sessions_excluded_when_disabled"`, passed locally via the Windows headless runner, 5 selected tests.
- `npx eslint --no-config-lookup -c eslint.runtime-guard.config.mjs static/boot.js static/sessions.js`, passed locally.

Full-suite CI context, not a required local check unless requested: `pytest tests/ -v --timeout=60`.

## Risks / Follow-ups

- Users who prefer the old cleaner sidebar can still turn the setting off; this PR only changes the default for unset configs.
- The gateway watcher still does not auto-start on gateway-less installs; this change only avoids the confusing warning loop when no imported sessions exist yet.
- If the multi-select origin filter in PR [#3566](https://github.com/nesquena/hermes-webui/pull/3566) lands later, it can absorb this default without changing the underlying setting migration story.

## Model Used

GPT-5 via Codex CLI

PR #4230, opened 2026-06-15
Release ON: show_cli_sessions default-on for new installs (#3988) + symlink delete/rename guard (#4217)

nesquena-hermes:
## Release ON — CLI sessions default-on for new installs (#3988) + symlink delete/rename guard (#4217)

A risk-separated 2-fix staging branch. **Awaiting Nathan's independent review before the full gate + ship.**

---

### #3988 — `show_cli_sessions` default → True for NEW installs, established installs grandfathered OFF

Combines the best of @rodboev's #3997 and @Sanjays2402's #4222, plus the migration semantics Nathan specified (new users ON; anyone already established stays as they were).

- **config.py:** `_SETTINGS_DEFAULTS["show_cli_sessions"]` flipped `False → True`, so new installs surface CLI / TUI / Telegram / Discord / Hermes One sessions in the sidebar without hunting for the Settings toggle.
- **Grandfather migration (`load_settings()`):** when the saved settings predate the flip — `onboarding_completed: True` AND `show_cli_sessions` absent — the value is pinned to `False`. An existing user's sidebar does NOT silently change. Fresh installs (no settings file, or onboarding not yet completed) get the `True` default. `save_settings()` reads `load_settings()` first and writes the full merged dict, so there's no window where a later save flips an established user on.
- **Default-true hydration at every read site** (`boot.js`, `panels.js` checkbox) via the `!== false` idiom, so an absent value never renders config-ON-but-UI-OFF (the #4006 default-mismatch class). The Settings checkbox renders checked-by-default consistently with the behavior.
- **10 tests** (`tests/test_issue3988_show_cli_sessions_default.py`): config default + bool-key round-trip; grandfather matrix (fresh / mid-onboarding / established-absent→OFF / explicit-true / explicit-false / save-preserves-grandfather); both frontend hydration sites (and asserting the buggy `!!` form is absent).

Supersedes #3997 and #4222 (both credited as co-authors).

### #4217 — symlink delete/rename guard (@rodboev #4227, absorbed)

`/api/file/delete` and `/api/file/rename` resolved the final symlink before operating, so a delete/rename against a workspace symlink name destroyed the real (possibly out-of-workspace) target. Both handlers now reject a symlinked path with a 400 — the same no-follow `is_symlink()` guard the move handler already had. +210/-0, 7 regression tests. (Bundled as a separate commit; risk-isolated from the #3988 change.)

---

### Gate status (pre-review)
- ruff + ESLint forward gates: CLEAN. New + absorbed test files pass (17 local).
- Full Codex + Opus + suite gate will run AFTER Nathan's independent review (per his request).

Co-authored-by: Rod Boev <rod.boev@gmail.com>
Co-authored-by: Sanjays2402 <Sanjays2402@users.noreply.github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,16 @@
 
 ## [Unreleased]
 
+## [v0.51.427] — 2026-06-15 — Release ON (CLI sessions on by default for new installs #3988 + symlink delete/rename guard #4217)
+
+### Changed
+
+- **CLI / TUI / messaging sessions now appear in the sidebar by default for new installs (#3988).** `show_cli_sessions` now defaults to `True`, so sessions from the CLI, TUI, Telegram, Discord, and the Hermes One desktop app show up in the WebUI sidebar without users having to discover the toggle in Settings. **Existing installs are grandfathered:** a user who already completed onboarding and never opted in keeps their previous (hidden) behavior — the default change only affects fresh installs. The toggle remains in Settings → Sessions for anyone who wants to change it.
+
+### Fixed
+
+- **Deleting or renaming a workspace symlink no longer destroys the file or directory it points to (#4217).** `/api/file/delete` and `/api/file/rename` resolved the final symlink before operating on the target, so a standard delete/rename against a symlink name acted on the real (possibly out-of-workspace) target. Both handlers now reject a symlinked path with a 400, matching the guard the move handler already had. (#4217)
+
 ## [v0.51.426] — 2026-06-15 — Release OM (custom-provider model-prefix routing fix, #4210)
 
 ### Fixed
--- a/api/config.py
+++ b/api/config.py
@@ -6505,7 +6505,7 @@ _SETTINGS_DEFAULTS = {
     "hide_empty_state_suggestions": False,  # hide the default new-chat suggestion buttons
     "show_tps": False,  # show tokens-per-second chip in assistant message headers
     "fade_text_effect": False,  # animate newly streamed words with a lightweight fade-in effect
-    "show_cli_sessions": False,  # merge CLI sessions from state.db into the sidebar
+    "show_cli_sessions": True,  # merge CLI/TUI/messaging sessions from state.db into the sidebar by default (#3988); established installs are grandfathered OFF by the load_settings backfill
     "show_cron_sessions": False,  # surface cron sessions in the sidebar (subordinate to show_cli_sessions)
     "show_previous_messaging_sessions": False,  # show older Telegram/Discord/etc. reset segments
     "sync_to_insights": False,  # mirror WebUI token usage to state.db for /insights
@@ -6647,6 +6647,17 @@ def load_settings() -> dict:
                         if k not in _SETTINGS_LEGACY_DROP_KEYS
                     }
                 )
+                # Grandfather established installs OFF for show_cli_sessions (#3988).
+                # The default flipped True so NEW users see CLI/TUI/messaging
+                # sessions without hunting for the toggle — but an existing user
+                # who already completed onboarding and never opted in should not
+                # have their sidebar silently change. If the saved settings predate
+                # the flip (onboarding done, key absent), pin it to the old default.
+                if (
+                    "show_cli_sessions" not in stored
+                    and bool(stored.get("onboarding_completed"))
+                ):
+                    settings["show_cli_sessions"] = False
         except Exception:
             logger.debug("Failed to load settings from %s", SETTINGS_FILE)
     settings["theme"], settings["skin"] = _normalize_appearance(
--- a/api/config.py
+++ b/api/config.py
@@ -6650,12 +6650,21 @@ def load_settings() -> dict:
                 # Grandfather established installs OFF for show_cli_sessions (#3988).
                 # The default flipped True so NEW users see CLI/TUI/messaging
                 # sessions without hunting for the toggle — but an existing user
-                # who already completed onboarding and never opted in should not
-                # have their sidebar silently change. If the saved settings predate
-                # the flip (onboarding done, key absent), pin it to the old default.
-                if (
-                    "show_cli_sessions" not in stored
-                    and bool(stored.get("onboarding_completed"))
+                # who never opted in should not have their sidebar silently change.
+                # Treat the install as established (and pin the old False default)
+                # when show_cli_sessions is absent AND the file already carries
+                # real user state — either onboarding was completed, or some
+                # setting OTHER than a not-yet-completed onboarding flag has been
+                # persisted. Keying on "has saved user state" (not just
+                # onboarding_completed) also covers a CLI-configured user who
+                # tweaked a WebUI setting before running the wizard. A genuinely
+                # new / still-mid-onboarding file falls through to the True default.
+                _established_keys = [
+                    k for k in stored
+                    if k not in ("show_cli_sessions", "onboarding_completed")
+                ]
+                if "show_cli_sessions" not in stored and (
+                    bool(stored.get("onboarding_completed")) or _established_keys
                 ):
                     settings["show_cli_sessions"] = False
         except Exception:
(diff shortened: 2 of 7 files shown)


nesquena (approved):
## Review — end-to-end ✅ (independent pre-gate review; clean, no fixes needed)

Independent review of **Release ON** — a risk-separated 2-fix staging branch the PR explicitly parks for my review before the full Codex+Opus+suite gate. I review the code (the version stamp / merge / full-gate is the release agent's). Two independent fixes:

1. **#3988** — flip `show_cli_sessions` default `False → True` for new installs, with a grandfather migration keeping established installs OFF.
2. **#4217** — security: reject symlinked paths in `/api/file/delete` + `/api/file/rename` (they followed the symlink and destroyed the real target).

### What this ships

`api/config.py` (+12/-1), `api/routes.py` (+4), `static/boot.js` (+1/-1), `static/panels.js` (+1/-1), `tests/test_issue3988_show_cli_sessions_default.py` (+146, 10 tests), `tests/test_workspace_symlink_delete_rename.py` (+203, 7 tests), CHANGELOG. Co-authored with @rodboev (#3997/#4227) and @Sanjays2402 (#4222).

Pulled a fresh `nousresearch/hermes-agent` tarball. `show_cli_sessions` is **absent** from it — a webui-only display setting (the webui merges CLI/TUI/messaging sessions from `state.db` into its sidebar; the agent never reads this preference). The file delete/rename handlers are webui workspace endpoints with no agent counterpart. No cross-tool surface for either fix. ✓

### #4217 — symlink guard (security): traced + independently proven

The guards ([routes.py:14728](api/routes.py#L14728) delete, [:14810](api/routes.py#L14810) rename) add, after `safe_resolve` + existence check:
```python
if (ws_root / body["path"]).is_symlink():
    return bad(handler, "Cannot delete/rename a symlinked entry")
```
This is **byte-parity** with the move handler's established, commented guard ([routes.py:14848](api/routes.py#L14848)).

The combined defense is sound — I traced both layers and proved them with a real-symlink harness:
- `safe_resolve` ([helpers.py:48-52](api/helpers.py#L48)) does `(root/requested).resolve()` (follows **all** symlinks, parent + leaf) then `relative_to(root.resolve())` → raises on any path whose **resolved** target escapes the workspace. This blocks an out-of-ws leaf symlink **and** parent-symlink traversal (`linkdir/secret.txt` where `linkdir→/outside`).
- But `safe_resolve` **allows** an in-workspace leaf symlink (e.g. `link.txt → real.txt`), resolving to the real file — so without the new guard, delete/rename would operate on `real.txt`, destroying the target instead of the link. The lexical `is_symlink()` leaf check (lstat, no-follow) catches exactly this case the containment check misses.

**Independent symlink-defense harness** (real symlinks in a tmp workspace):
```
guard: link.txt is_symlink -> True (rejected); real.txt -> False (allowed)
safe_resolve(link.txt) -> real.txt   (proves why the guard is needed)
safe_resolve(escape.txt) raises ValueError       (out-of-ws leaf blocked)
safe_resolve(linkdir/secret.txt) raises ValueError (parent-symlink escape blocked)
guard catches the in-ws symlink that post-resolve containment alone would MISS
6/6 symlink-defense assertions passed
```
The `is_symlink()` leaf check is correct even under a symlinked workspace root (lstat on the joined path checks the final component without following it; parent symlinks are resolved by the OS to reach the leaf). TOCTOU between the check and the op is the same window the move handler already accepts (single-user workspace); not a new concern.

### #3988 — grandfather migration: traced

**Default** flipped to `True` ([config.py:6508](api/config.py#L6508)). **Grandfather** in `load_settings()` ([config.py:6650-6660](api/config.py#L6650)), inside `if isinstance(stored, dict)` after the stored-merge:
```python
if "show_cli_sessions" not in stored and bool(stored.get("onboarding_completed")):
    settings["show_cli_sessions"] = False
```
Traced every branch:
- **Fresh install** (no file): `stored` stays `None` → block unreached → `True` default. ✓
- **Mid-onboarding** (file exists, `onboarding_completed` false/absent): not grandfathered → `True`. ✓
- **Established** (`onboarding_completed` true, key absent): pinned `False`. ✓
- **Explicit value** (key present): `"show_cli_sessions" not in stored` is false → respects the saved value. ✓

**No silent-flip window:** the grandfather re-applies on **every** `load_settings`, and `save_settings` starts with `current = load_settings()` ([config.py](api/config.py)) then merges incoming changes over it and writes the full dict — so an established user with the key absent on disk loads as `False`, and the first subsequent save persists `False` explicitly. There is no path that writes `True` for an established user without an explicit user toggle.

**Default-true hydration at all 3 read sites** (avoids the #4006 config-ON-but-UI-OFF class): config.py default `True`; `boot.js` `window._showCliSessions=s.show_cli_sessions!==false` ([boot.js:1851](static/boot.js#L1851)); panels.js checkbox `showCliCb.checked=settings.show_cli_sessions!==false` ([panels.js:6776](static/panels.js#L6776)). The `!== false` idiom means an absent value renders ON, consistent with the config default. ✓

- **Security (#4217):** the guard short-circuits before any filesystem mutation; out-of-ws targets were already blocked by `safe_resolve`. No new sink.
- **Security (#3988):** `show_cli_sessions` is in `_SETTINGS_BOOL_KEYS` so it round-trips as a coerced bool — a crafted POST can't inject a non-bool.
- **Risk isolation:** the two fixes are independent commits; neither touches the other's surface.
- The grandfather block is guarded by `isinstance(stored, dict)`, so `stored.get(...)` is safe.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Fresh install, no settings file | `show_cli_sessions` = True (new default) ✅ |
| Mid-onboarding (not completed) | True ✅ |
| Established install, key absent | grandfathered False ✅ |
| Established install, first save after upgrade | False persisted explicitly (no flip) ✅ |
| Explicit true / false saved | respected ✅ |
| Absent value at boot.js / checkbox | renders ON (matches config default) ✅ |
| Delete/rename in-ws symlink → real target | rejected 400; real target survives (#4217) ✅ |
| Delete/rename out-of-ws / parent-symlink path | blocked by `safe_resolve` ✅ |
| Delete/rename real file/dir | still works ✅ |
| CLI reads same config / workspace | webui-only setting + webui-only endpoints; agent unaffected ✅ |

### Tests

- `tests/test_issue3988_show_cli_sessions_default.py` — **10/10** (config default + bool-key; grandfather matrix incl. `test_established_install_persists_grandfather_through_save`; both frontend hydration sites + asserting the buggy `!!` form is gone).
- `tests/test_workspace_symlink_delete_rename.py` — **7/7** (delete/rename symlink-to-file and symlink-to-dir rejected; real delete/rename still work; move still rejected).
- **Full suite: 8854 passed / 135 skipped / 0 failures** in 165s (deselected the pre-existing darwin CRLF flake; ignored `test_passkey_auth.py`'s local `cryptography`-dep collection gap).
- **JS gates:** `node --check` on boot.js + panels.js clean; ESLint runtime guard exit 0.
- **Independent harness:** 6/6 symlink-defense (above).
- **CI on the PR head: green** — the completed shards (browser-smoke + lint + 9 test shards) pass.

- The grandfather key on `onboarding_completed` is the right signal, but it does mean a user who somehow has a settings file *without* `onboarding_completed` (manually edited, or a very old pre-onboarding schema) would be treated as fresh and get CLI sessions ON. That's an acceptable, vanishingly-rare edge — and arguably correct (no evidence they opted out).
- The CHANGELOG already carries the `v0.51.427` version section (this is a release-staging branch); that's the release agent's domain — flagging only that the code matches the entry.

### Recommendation

✅ **Approved** for the independent-review gate. Parked at approval — ready for the agent's full Codex+Opus+suite gate and the release pipeline.

Both fixes are correct and well-tested. The **security fix (#4217)** is the load-bearing one: I verified it's byte-parity with the existing move guard and that the combined `safe_resolve` (post-resolution containment) + lexical `is_symlink()` leaf check closes every path — out-of-ws leaf, parent-symlink traversal, and the in-ws leaf symlink that containment alone would let through — proven with a real-symlink harness (6/6). The **#3988 migration** correctly grandfathers established installs OFF (re-applied every load, persisted on save, no silent-flip window) while defaulting new installs ON, with consistent default-true hydration at all three read sites. Ship.

Issue #4234, opened 2026-06-15
/api/file/save follows symlinks to the resolved target (same class as #4217)


nesquena-hermes:
## Summary

Follow-up from #4217 (symlink delete/rename/move guard, shipped v0.51.427). `_handle_file_save` in `api/routes.py` has the same resolve-then-write-through-symlink pattern that #4217 fixed for delete/rename/move, but for the **write** path it was intentionally left out of scope to avoid re-opening that PR's approved review.

## The gap

`_handle_file_save` does:
```python
ws_root = Path(s.workspace)
target = safe_resolve(ws_root, body["path"])
if not target.exists():
    return bad(handler, "File not found", 404)
if target.is_dir():
    return bad(handler, "Cannot save: path is a directory")
...
fd = open_anchored_write_fd(ws_root, target)
```
`safe_resolve()` follows the final symlink, so saving to a workspace symlink `link.txt → real.txt` writes through to `real.txt` — overwriting the symlink's in-workspace target instead of the link. `safe_resolve` still blocks out-of-workspace targets (containment is intact), so this is **lower severity** than the delete/rename case (overwrite of an in-workspace file, not destruction of an out-of-workspace target) — but it's the same class and worth closing for consistency.

## Suggested fix

Mirror the #4217 guard (and its dangling-symlink ordering): add a no-follow `is_symlink()` check on the lexically-requested path **before** the `exists()` probe, and reject with a 400:
```python
if (ws_root / body["path"]).is_symlink():
    return bad(handler, "Cannot save to a symlinked entry")
if not target.exists():
    return bad(handler, "File not found", 404)
```
Plus a regression test covering: save to a symlink-to-file rejected, save to a real file still works, dangling symlink rejected (not 404).

## Context

Identified during the #4230 / #4217 ship review (Greptile flagged `_handle_file_save` as a class-sibling of the delete/rename guard). Deliberately deferred to keep #4230's scope tight. Single-operator workspace, so the practical risk is a self-inflicted overwrite — low urgency, but a clean consistency fix.

_Filed by the autonomous PR sweep; this is a security-surface change so it's left for maintainer review rather than shipped unattended._

PR #4239, opened 2026-06-15
fix(file): reject symlink saves

zapabob:
## Thinking Path
- Issue #4234 identifies `/api/file/save` as the remaining sibling of the symlink delete/rename/move guard from #4217.
- I checked the current open security/path/symlink PR and issue set before editing. I found no open PR covering this exact save path.
- The narrow fix is to reject the lexically requested workspace entry when it is a symlink before the save path writes through the resolved target.

## What Changed
- Reject `/api/file/save` requests whose requested workspace path is a symlinked entry.
- Add regression coverage for saving through a symlink, dangling symlink ordering, and ordinary real-file saves.

## Why It Matters
- `safe_resolve()` follows the final symlink. Without this guard, saving `link.txt` can overwrite the linked in-workspace target instead of rejecting the symlink entry consistently with delete, rename, and move.
- The guard keeps the file mutation contract uniform across workspace write operations.

## Verification
- `git diff --check`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 HERMES_WEBUI_PYTHON=C:\Users\downl\AppData\Local\Programs\Python\Python312\python.exe HERMES_HOME=%TEMP%\hermes-webui-pr1-home py -3.12 -m pytest tests/test_workspace_symlink_delete_rename.py -q`
  - Windows local result: 3 passed, 9 skipped. The symlink-specific cases skip because this shell lacks Windows symlink privilege; Linux CI should exercise them.

## Risks / Follow-ups
- This intentionally does not change read/raw behavior or broader workspace symlink policy.
- It preserves existing behavior for ordinary file saves.

## Model Used
- GPT-5 Codex

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -14824,6 +14824,8 @@ def _handle_file_save(handler, body):
     try:
         ws_root = Path(s.workspace)
         target = safe_resolve(ws_root, body["path"])
+        if (ws_root / body["path"]).is_symlink():
+            return bad(handler, "Cannot save to a symlinked entry")
         if not target.exists():
             return bad(handler, "File not found", 404)
         if target.is_dir():
--- a/tests/test_workspace_symlink_delete_rename.py
+++ b/tests/test_workspace_symlink_delete_rename.py
@@ -138,6 +138,51 @@ def test_rename_workspace_symlink_to_file_rejected(workspace):
 
 # ── Non-symlink operations still work ───────────────────────────────────
 
+def test_save_workspace_symlink_to_file_rejected(workspace):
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in v0.51.433 (Release OT) via #4244. Thanks @zapabob — this implements exactly the #4234 follow-up to the #4217 symlink guard: /api/file/save now rejects a symlinked path (before the existence check, so dangling symlinks are caught) instead of writing through to the target. Codex + Opus + full suite (9065) all green. Closing as shipped.

PR #4240, opened 2026-06-15
fix(skills): reject symlinked skill saves

zapabob:
## Thinking Path
- I reviewed the open PR/issue set for skill save/path/symlink/SKILL.md work before editing.
- Existing skill PRs are feature/UI oriented or profile management work; I did not find a PR that guards the `/api/skills/save` final `SKILL.md` write against symlinked files.
- The narrow risk is write-through when an existing `SKILL.md` inside the active skills directory is a symlink.

## What Changed
- Reject `/api/skills/save` when the target `SKILL.md` is a symlinked file.
- Add focused regression coverage for a symlinked `SKILL.md` and a normal skill save.

## Why It Matters
- The handler validates the resolved skill directory, then writes `SKILL.md` directly. If that final file is a symlink, saving a skill can overwrite the linked target.
- This keeps profile skill editing bounded to an actual skill file rather than following an attacker-prepared link.

## Verification
- `git diff --check`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 HERMES_WEBUI_PYTHON=C:\Users\downl\AppData\Local\Programs\Python\Python312\python.exe HERMES_HOME=%TEMP%\hermes-webui-pr2-home py -3.12 -m pytest tests/test_skill_save_symlink_guard.py -q`
  - Windows local result: 1 passed, 1 skipped. The symlink-specific case skips because this shell lacks Windows symlink privilege; Linux CI should exercise it.

## Risks / Follow-ups
- This does not alter skill name/category validation or external skill discovery.
- Ordinary skill creation and update remain unchanged.

## Model Used
- GPT-5 Codex

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -16600,6 +16600,8 @@ def _handle_skill_save(handler, body):
         return bad(handler, "Invalid skill path")
     skill_dir.mkdir(parents=True, exist_ok=True)
     skill_file = skill_dir / "SKILL.md"
+    if skill_file.is_symlink():
+        return bad(handler, "Cannot save to a symlinked skill file")
     skill_file.write_text(body["content"], encoding="utf-8")
     return j(handler, {"ok": True, "name": skill_name, "path": str(skill_file)})
 
--- a/tests/test_skill_save_symlink_guard.py
+++ b/tests/test_skill_save_symlink_guard.py
@@ -1,0 +1,60 @@
+import os
+
+import pytest
+
+import api.routes as routes
+
+
+class _FakeHandler:
+    pass
+
+
+def _patch_skill_routes(monkeypatch, skills_dir):
+    cap = {}
+    monkeypatch.setattr(routes, "_active_skills_dir", lambda: skills_dir)
+    monkeypatch.setattr(routes, "j", lambda h, o: (cap.__setitem__("ok", o), True)[1])
+    monkeypatch.setattr(
+        routes,
+        "bad",
+        lambda h, m, c=400: (cap.__setitem__("bad", (m, c)), True)[1],
+    )
+    return cap
+
+
+def test_skill_save_rejects_symlinked_skill_file(tmp_path, monkeypatch):
+    skills_dir = tmp_path / "skills"
+    skill_dir = skills_dir / "demo"
+    skill_dir.mkdir(parents=True)
+    outside = tmp_path / "outside.md"
+    outside.write_text("important", encoding="utf-8")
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Read the diff against `origin/master` and the full `_handle_skill_save` it sits in. The guard is correctly placed and closes a real gap that the existing path-traversal check does *not* cover.

The handler already resolves the directory and confirms it stays inside the active skills dir:

```python
skill_dir.resolve().relative_to(skills_dir.resolve())   # api/routes.py:16600
...
skill_dir.mkdir(parents=True, exist_ok=True)
skill_file = skill_dir / "SKILL.md"
skill_file.write_text(body["content"], encoding="utf-8")
```

That `resolve()`+`relative_to()` *follows* symlinks, so a symlinked **directory** that escapes the sandbox is already rejected (the resolved target wouldn't be under `skills_dir`). What it doesn't catch is the final-file case: `skill_dir` is legitimately inside the sandbox, but `SKILL.md` itself is a symlink — the code appends `/SKILL.md` and writes without re-resolving. Your new check sits exactly there:

```python
skill_file = skill_dir / "SKILL.md"
if skill_file.is_symlink():
    return bad(handler, "Cannot save to a symlinked skill file")
```

This matches the established no-follow pattern the workspace ops already use (`api/routes.py:14800`, `:14887`, `:14932` all guard with `(...).is_symlink()` and the comment there spells out "is_symlink() is a no-follow lstat on the requested path"). So it's consistent with how the rest of the codebase reasons about this threat.

One boundary note, not a blocker: `is_symlink()` is final-component only, so it won't catch a symlinked *intermediate* parent under the skills dir (e.g. `skills/<category>` being a link). The `skill_dir.resolve().relative_to(...)` guard above already covers the escape-the-sandbox version of that, so the residual case is "symlinked parent that still resolves inside the sandbox," which is low-risk and out of scope here.

The test (`tests/test_skill_save_symlink_guard.py`) asserts both the rejection and that the outside file is left untouched, and the normal-save case still writes through — that's the right coverage. CI is green across all shards. LGTM on the security logic.

nesquena-hermes:
Shipped in v0.51.434 (Release OU) via #4245. Thanks @zapabob — skill saves now reject a symlinked SKILL.md, extending the #4217/#4234 symlink hardening to the skills surface. Codex + Opus + full suite (9067) all green. (One CI shard hit a known unrelated gateway_sync flake; re-ran clean before merge.) Closing as shipped.

PR #4242, opened 2026-06-15
fix(memory): reject symlinked memory writes

zapabob:
## Thinking Path
- I reviewed current open PRs/issues for memory write, MEMORY.md, SOUL.md, and symlink-related work before editing.
- Existing memory issues cover wrong-profile writes and broader memory features, but I did not find a PR guarding the final memory files or `memories/` directory against symlink write-through.
- The narrow fix is to refuse writes when the target persistent memory surface is a symlink.

## What Changed
- Reject `/api/memory/write` for `memory` and `user` sections when the `memories/` directory is itself a symlink.
- Reject writes when the final `MEMORY.md`, `USER.md`, or `SOUL.md` target is a symlink.
- Add regression coverage for a symlinked memory file, a symlinked memories directory, and a normal memory write.

## Why It Matters
- `/api/memory/write` is a persistent profile data write. If a local attacker or previous unsafe state prepares a symlinked memory path, the WebUI can overwrite the linked target.
- Refusing symlinked memory surfaces keeps the endpoint bound to real profile memory files.

## Verification
- `git diff --check`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 HERMES_WEBUI_PYTHON=C:\Users\downl\AppData\Local\Programs\Python\Python312\python.exe HERMES_HOME=%TEMP%\hermes-webui-pr3-home py -3.12 -m pytest tests/test_memory_write_symlink_guard.py -q`
  - Windows local result: 1 passed, 2 skipped. The symlink-specific cases skip because this shell lacks Windows symlink privilege; Linux CI should exercise them.

## Risks / Follow-ups
- This intentionally does not change profile selection or memory redaction behavior.
- Ordinary memory writes continue to create/update real files as before.

## Model Used
- GPT-5 Codex

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -16715,16 +16715,23 @@ def _handle_memory_write(handler, body):
     except ImportError:
         home = Path.home() / ".hermes"
         mem_dir = home / "memories"
-    mem_dir.mkdir(parents=True, exist_ok=True)
     section = body["section"]
     if section == "memory":
+        if mem_dir.is_symlink():
+            return bad(handler, "Cannot write through a symlinked memories directory")
+        mem_dir.mkdir(parents=True, exist_ok=True)
         target = mem_dir / "MEMORY.md"
     elif section == "user":
+        if mem_dir.is_symlink():
+            return bad(handler, "Cannot write through a symlinked memories directory")
+        mem_dir.mkdir(parents=True, exist_ok=True)
         target = mem_dir / "USER.md"
     elif section == "soul":
         target = home / "SOUL.md"
     else:
         return bad(handler, 'section must be "memory", "user", or "soul"')
+    if target.is_symlink():
+        return bad(handler, "Cannot write to a symlinked memory file")
     target.write_text(body["content"], encoding="utf-8")
     return j(handler, {"ok": True, "section": section, "path": str(target)})
 
--- a/tests/test_memory_write_symlink_guard.py
+++ b/tests/test_memory_write_symlink_guard.py
@@ -1,0 +1,85 @@
+import os
+
+import pytest
+
+import api.profiles as profiles
+import api.routes as routes
+
+
+class _FakeHandler:
+    pass
+
+
+def _patch_memory_routes(monkeypatch, home):
+    cap = {}
+    monkeypatch.setattr(profiles, "get_active_hermes_home", lambda: home)
+    monkeypatch.setattr(routes, "j", lambda h, o: (cap.__setitem__("ok", o), True)[1])
+    monkeypatch.setattr(
+        routes,
+        "bad",
+        lambda h, m, c=400: (cap.__setitem__("bad", (m, c)), True)[1],
+    )
+    return cap
+
+
+def test_memory_write_rejects_symlinked_memory_file(tmp_path, monkeypatch):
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Read the diff against `origin/master` plus the full `_handle_memory_write`. The guard logic is right and the placement is careful — worth calling out one detail that makes it more robust than it might look at first glance.

Unlike the skills handler, this path has *no* `resolve()`+`relative_to()` traversal guard, because the section is a fixed enum (`memory`/`user`/`soul`) and the paths are derived, not user-supplied. So the `is_symlink()` checks you add are the only symlink protection here, and they cover both surfaces:

```python
if section == "memory":
    if mem_dir.is_symlink():
        return bad(handler, "Cannot write through a symlinked memories directory")
    mem_dir.mkdir(parents=True, exist_ok=True)
    target = mem_dir / "MEMORY.md"
elif section == "user":
    if mem_dir.is_symlink():
        ...
    target = mem_dir / "USER.md"
elif section == "soul":
    target = home / "SOUL.md"
...
if target.is_symlink():
    return bad(handler, "Cannot write to a symlinked memory file")
```

Two things I checked that are correct:

1. **The trailing `target.is_symlink()` covers all three sections**, including `soul`. SOUL.md lives at `home / "SOUL.md"` and never goes through the `mem_dir` branch, but it still hits the final file-level check before `write_text`. So a symlinked `SOUL.md` is rejected too, even though there's no dir-level guard for it (correct — `home` is the active profile root from `get_active_hermes_home()`, trusted state, and isn't created here).

2. **Moving `mem_dir.mkdir()` inside the branches** (it was unconditional on master at `api/routes.py:16716`) means the dir is only created after the symlink check passes — so a write to a symlinked `memories/` dir bails before any filesystem mutation. Good ordering.

This matches the no-follow `is_symlink()` lstat pattern the workspace ops already use (`api/routes.py:14800`, `:14887`, `:14932`), so it's consistent with the rest of the codebase's threat model. Same final-component-only caveat applies as elsewhere (a symlinked parent *above* `mem_dir` that still resolves inside the profile home isn't caught), but `home` being trusted profile state makes that out of scope.

Test coverage in `tests/test_memory_write_symlink_guard.py` exercises the symlinked-file case, the symlinked-directory case, and a normal write, each asserting the linked target is left untouched — that's the right shape. CI green across all shards. LGTM.

nesquena-hermes:
Shipped in **v0.51.452** (Release PM) — merged via release PR #4298. Thank you @zapabob! 🎉

We shipped the **target-file** symlink guard (refuse to write when MEMORY.md/USER.md/SOUL.md is a symlink), which closes the arbitrary-file-clobber vector and matches the shipped skills/plugins symlink-rejection class (#4217/#4234/#4240). 

One deliberate trim: we dropped the **parent-directory** symlink rejection — symlinking the whole `.hermes/memories` directory is a legitimate setup for some users, and with the fixed filenames the parent-symlink case can at most create one of three known files in the link target (not clobber an arbitrary existing file). So only the concrete target file is guarded. A regression test pins that the symlinked-parent case stays allowed.

Gate: Codex + Opus both SAFE, full suite 9169 passed. Closing as shipped.

Issue #3961, opened 2026-06-11
Resolve /api/providers + /api/models credentials from the thread-local profile channel (drop the os.environ mirror race)


nesquena-hermes:
## Summary

Follow-up from the #3957 fix (#3960, v0.51.356). The independent review of #3960 surfaced a **pre-existing** architectural limitation in how the read-only provider/model endpoints resolve per-profile credentials, recommending a tracking issue.

## The limitation

WebUI profile switching is per-client/cookie-scoped (#798). The fix in #3960 routes `/api/providers` and `/api/models` (and the detached models-rebuild worker) through the active profile via `profile_env_for_active_request` / `profile_scope_for_detached_worker`, which delegate to the **pre-existing** `profile_env_for_background_worker` (the same mechanism streaming uses for title-gen / checkpoints).

That mechanism uses **two channels**:
1. a **thread-local** (`_thread_ctx`) — race-free, the primary routing channel; and
2. a **process-global `os.environ` mirror**, kept because several production Hermes readers still call `os.getenv()` directly for provider credentials.

By explicit design (`api/profiles.py` ~L739-743), `_ENV_LOCK` is held only around the env **setup/restore**, NOT the `yield` body (unlike `_cron_env_lock`, which serializes the whole body). The server is a `ThreadingHTTPServer` (`daemon_threads=True`), so requests run concurrently.

**Consequence:** two clients on **different non-default profiles** hitting `/api/providers` concurrently can interleave their `os.environ` mutations during the (multi-second) serial auth-probe window — transiently bleeding one profile's credentials into the other's probe, and potentially leaving a non-default profile's value as the process-global "default" after both restore.

## Scope / severity

- **Pre-existing, not introduced by #3960.** #3960's diff to `api/profiles.py` is purely additive — `profile_env_for_background_worker` itself is unchanged. Streaming already exercises the identical `os.environ`-mirror race for concurrent different-profile title-gen/checkpoint workers.
- **`/api/models` is largely insulated**: its rebuild is serialized by `_cache_build_cv` (one rebuild at a time), and its cache path + fingerprint resolve through the **thread-local** (race-free). The exposure concentrates on `/api/providers` and the opt-in synchronous rebuild (`_LIVE_REBUILD_BUDGET_SECONDS <= 0`).
[...]

Issue #3294, opened 2026-05-31
bug: Profile Tool Configuration Not Respected in Web UI Unless Tool Restrictions Are Explicitly Set


gottipx:
**Description**

I created a second profile and disabled all tools and skills for that profile.

When using the profile through Hermes CLI, a simple prompt such as Hi results in approximately 400 input tokens, which is consistent with a minimal system context.

However, when using the same profile in the Web UI, the identical prompt results in approximately 15,000 input tokens.

It appears that the Chat Interface loads the global default tool configuration when the Tool Restrictions field is left blank, regardless of the profile configuration. As a result, all tools are included in the prompt context unless explicitly overridden.

**Expected Behavior**

The profile’s configured tool and skill settings should be respected by the Chat Interface. If a profile has no tools or skills enabled, the prompt context should remain minimal without requiring additional tool restriction settings.

**Actual Behavior**

When the Tool Restrictions field is empty, the Chat Interface appears to load the global default tool set, causing a significant increase in input tokens (~15K).

If the Tool Restrictions field is set to a value such as none, the input token count drops to approximately 400–900 tokens, which aligns with the profile configuration.

**Steps to Reproduce**

1. Create a new profile.
2. Configure the profile with:
    * No skills enabled
    * No tools enabled
    * A model only
3. Open a new chat session using this profile.
4. Send the message: Hi
5. Observe the input token count (approximately 15K).
6. Start another new session.
7. Set Tool Restrictions to a value such as none.
8. Send the message: Hi again.
9. Observe that the input token count drops to approximately 400–900 tokens.

**Additional Observation**

Hermes CLI respects the profile configuration and produces the expected low token count. The discrepancy appears specific to the Web UI behavior when Tool Restrictions are left unset.

nesquena-hermes:
## Summary

Reproduced the contract on `origin/master`. This is a WebUI-side profile-scoping leak on the streaming worker thread, not an agent bug. When **Tool Restrictions** is blank, the WebUI resolves toolsets from `_resolve_cli_toolsets(_cfg)`, where `_cfg` is read via the *active-profile* config path. But the agent runs on a background thread whose active-profile lookup falls back to the process-global `default` profile — so it reads the **default** profile's `platform_toolsets.cli` (all tools) instead of your second profile's empty list, inflating the prompt to ~15K tokens. The agent's `_get_platform_tools` is doing the right thing; it's being handed the wrong profile's config.

## Code reference

[...]

PR #4243, opened 2026-06-15
Add AI/ML API onboarding preset

### LINKED ISSUE #4045: Provider request: AIML API OpenAI-compatible setup/preset
Hi, thanks for maintaining Hermes WebUI.

Would you be open to an AIML API setup/preset for the WebUI provider flow?

AIML API exposes an OpenAI-compatible endpoint at:

```text
https://api.aimlapi.com/v1
```

Auth is standard bearer token auth:

```text
Authorization: Bearer <AIMLAPI_API_KEY>
```

Why this might fit Hermes WebUI:

- Hermes WebUI already has strong support for custom/provider-aware model selection and live `/v1/models` discovery.
- AIML API is useful as a multi-model OpenAI-compatible provider, so users can access OpenAI/Anthropic/Gemini-style model IDs through one API key.
- The integration could stay small and avoid changing the generic custom-provider path.

Potential focused scope:

- add an AIML API provider preset or documented setup path with:
  - provider id/name: `aimlapi` / `AIML API`
  - base URL: `https://api.aimlapi.com/v1`
  - env/key hint: `AIMLAPI_API_KEY`
  - key URL: `https://aimlapi.com/app/keys`
- rely on existing `/v1/models` discovery rather than hardcoding a large model list
- add a short docs/example snippet for configuring AIML API as an OpenAI-compatible provider
- keep this separate from unrelated provider/model picker changes

If this scope looks useful, I can prepare a small PR.

D1m7asis:
Summary:
- Add AIML API as a named onboarding provider preset.
- Use AIMLAPI_API_KEY and https://api.aimlapi.com/v1 as the OpenAI-compatible base URL.
- Keep the preset model list empty and use the existing /models probe flow for live model discovery.

Tests:
- python -m pytest tests\test_custom_providers_in_panel.py tests\test_sprint39.py
- npm run lint:runtime
- python -m ruff check api\onboarding.py tests\test_sprint39.py
- python scripts\ruff_lint.py --diff origin/master

Closes #4045

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- **AI/ML API is now available as an onboarding provider preset.** Users can configure it with `AIMLAPI_API_KEY`, use the OpenAI-compatible `https://api.aimlapi.com/v1` endpoint, and pick from a curated starter catalog of chat-capable models. (#4243)
+
 ### Internal
 
 - **Windows pytest-harness compatibility (#3664).** Hardened the test suite to run on Windows: profile-home fallback paths are path-normalized, strict POSIX file-mode (`0o600`) assertions are gated behind `os.name != "nt"` (Linux still asserts them at full strictness), the conftest cleanup handles Windows process-tree/port teardown and the Py3.12+ `shutil.rmtree` `onexc` shim, and tests that require `fork`/`fcntl` carry `@requires_fork` / `@requires_fcntl` markers (which never skip on Linux). Test-only — no runtime or app behavior change, no Linux CI behavior change. (#4254, #4255, #4256, #4257, #4259, #4263, #4266, #4274)
--- a/api/config.py
+++ b/api/config.py
@@ -1285,6 +1285,7 @@ _PROVIDER_MODELS = {
         {"id": "gpt-5.4-mini", "label": "GPT-5.4 Mini"},
         {"id": "gpt-5.4",      "label": "GPT-5.4"},
     ],
+    # Curated chat-capable AIML API subset for the initial offline picker.
     "aimlapi": [
         {"id": "gpt-4o-mini", "label": "GPT-4o mini"},
         {"id": "gpt-4o", "label": "GPT-4o"},
--- a/api/config.py
+++ b/api/config.py
@@ -873,6 +873,7 @@ _PROVIDER_DISPLAY = {
     "openrouter": "OpenRouter",
     "anthropic": "Anthropic",
     "openai": "OpenAI",
+    "aimlapi": "AI/ML API",
     "openai-api": "OpenAI API",
     "openai-codex": "OpenAI Codex",
     "xai-oauth": "xAI Grok OAuth",
@@ -1283,6 +1284,26 @@ _PROVIDER_MODELS = {
         {"id": "gpt-5.5-mini", "label": "GPT-5.5 Mini"},
         {"id": "gpt-5.4-mini", "label": "GPT-5.4 Mini"},
         {"id": "gpt-5.4",      "label": "GPT-5.4"},
+    ],
+    "aimlapi": [
+        {"id": "gpt-4o-mini", "label": "GPT-4o mini"},
+        {"id": "gpt-4o", "label": "GPT-4o"},
+        {"id": "gpt-5-chat", "label": "GPT-5 Chat"},
+        {"id": "gpt-5", "label": "GPT-5"},
+        {"id": "gpt-5-mini", "label": "GPT-5 mini"},
+        {"id": "gpt-5-nano", "label": "GPT-5 nano"},
+        {"id": "gpt-5.1-chat-latest", "label": "GPT-5.1 Chat latest"},
+        {"id": "gpt-5.2-chat-latest", "label": "GPT-5.2 Chat latest"},
+        {"id": "claude-sonnet-4-5-20250929", "label": "Claude Sonnet 4.5"},
+        {"id": "claude-sonnet-4-6", "label": "Claude Sonnet 4.6"},
+        {"id": "gemini-3-flash-preview", "label": "Gemini 3 Flash Preview"},
+        {"id": "gemini-2.5-pro", "label": "Gemini 2.5 Pro"},
+        {"id": "deepseek-chat", "label": "DeepSeek Chat"},
+        {"id": "deepseek-reasoner", "label": "DeepSeek Reasoner"},
+        {"id": "x-ai/grok-4-07-09", "label": "Grok 4"},
+        {"id": "qwen3.5-flash", "label": "Qwen3.5 Flash"},
+        {"id": "qwen3-coder-480b-a35b-instruct", "label": "Qwen3 Coder"},
+        {"id": "meta-llama/Llama-3.3-70B-Instruct-Turbo", "label": "Llama 3.3 70B Instruct Turbo"},
     ],
     "openai-api": [
         {"id": "gpt-5.5",      "label": "GPT-5.5"},
@@ -2175,7 +2196,7 @@ def resolve_model_provider(model_id: str) -> tuple:
         # fired in the prefix==config_provider case, causing HTTP 404 from the
         # portal which requires the full provider/model id (#2177; sibling of
         # #854 / #894 for Nous, where this guard was originally added).
-        _PORTAL_PROVIDERS = {"nous", "opencode-zen", "opencode-go", "nvidia"}
+        _PORTAL_PROVIDERS = {"nous", "opencode-zen", "opencode-go", "nvidia", "aimlapi"}
         if config_provider in _PORTAL_PROVIDERS:
             return model_id, config_provider, config_base_url
         # If prefix matches config provider exactly, strip it and use that provider directly.
@@ -5048,6 +5069,7 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
             for k in (
                 "ANTHROPIC_API_KEY",
                 "OPENAI_API_KEY",
+                "AIMLAPI_API_KEY",
                 "OPENROUTER_API_KEY",
                 "GOOGLE_API_KEY",
                 "GEMINI_API_KEY",
@@ -5082,6 +5104,8 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                 # picker without a manual config.yaml edit. Users without Codex OAuth will see
                 # picker entries but hit auth errors at inference time (#1189 known limitation).
                 detected_providers.add("openai-codex")
+            if all_env.get("AIMLAPI_API_KEY"):
+                detected_providers.add("aimlapi")
             if all_env.get("OPENROUTER_API_KEY"):
                 detected_providers.add("openrouter")
             if all_env.get("GOOGLE_API_KEY"):
--- a/api/onboarding.py
+++ b/api/onboarding.py
@@ -65,6 +65,15 @@ _SUPPORTED_PROVIDER_SETUPS = {
         "models": list(_PROVIDER_MODELS.get("openai", [])),
         "category": "easy_start",
     },
+    "aimlapi": {
+        "label": "AI/ML API",
+        "env_var": "AIMLAPI_API_KEY",
+        "default_model": "gpt-5-chat",
+        "default_base_url": "https://api.aimlapi.com/v1",
+        "requires_base_url": False,
+        "models": list(_PROVIDER_MODELS.get("aimlapi", [])),
+        "category": "easy_start",
+    },
     # ── Open / self-hosted ─────────────────────────────────────────────
     "ollama": {
         "label": "Ollama",
--- a/api/providers.py
+++ b/api/providers.py
@@ -688,6 +688,7 @@ _PROVIDER_ENV_VAR: dict[str, str] = {
     "openrouter": "OPENROUTER_API_KEY",
     "anthropic": "ANTHROPIC_API_KEY",
     "openai": "OPENAI_API_KEY",
+    "aimlapi": "AIMLAPI_API_KEY",
     "google": "GOOGLE_API_KEY",
     "gemini": "GEMINI_API_KEY",
     "zai": "GLM_API_KEY",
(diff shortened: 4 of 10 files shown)


nesquena-hermes:
Pulled the branch and read the diff against `origin/master`. The preset wiring itself looks right — `aimlapi` lands in `_SUPPORTED_PROVIDER_SETUPS` with `AIMLAPI_API_KEY`, `https://api.aimlapi.com/v1`, `requires_base_url: False`, `probe_models: True`, empty `models`, and the catalog now surfaces `probe_models` (`api/onboarding.py:65-74`, `:805`), so live `/models` discovery drives the picker. The `static/onboarding.js` changes correctly gate the probe on `requires_base_url || probe_models` and seed `baseUrl` from `default_base_url`. That part is sound.

## One test will fail deterministically

`tests/test_sprint39.py::test_aimlapi_setup_writes_aimlapi_env_and_base_url` passes `"***"` as the key but asserts the *written* value is `"aiml-test-key"`:

```python
mod.apply_onboarding_setup({
    "provider": "aimlapi",
    "model": "openai/gpt-4o-mini",
    "api_key": "***",          # tests/test_sprint39.py:181
})
write_env_mock.assert_called_once()
self.assertEqual(write_env_mock.call_args.args[1], {"AIMLAPI_API_KEY": "aiml-test-key"})  # :185
...
self.assertEqual(os.environ.get("AIMLAPI_API_KEY"), "aiml-test-key")  # :190
```

There's no `***`→key substitution anywhere in the apply path. `apply_onboarding_setup` reads the key verbatim and writes it verbatim:

```python
api_key = str(body.get("api_key") or "").strip()      # api/onboarding.py:949
...
_write_env_file(env_path, {provider_meta["env_var"]: api_key})  # :1017
...
os.environ[provider_meta["env_var"]] = api_key          # :1031
```

So `write_env_mock.call_args.args[1]` will be `{"AIMLAPI_API_KEY": "***"}` and `os.environ["AIMLAPI_API_KEY"]` will be `"***"` — both assertions against `"aiml-test-key"` fail. Note the sibling test right above it (`test_api_key_set_in_os_environ_after_apply`, `:130-155`) gets this right: it passes `"***"` and asserts `== "***"`.

## Fix

Pass the value you're asserting on into the body so input and expectation match:

```python
mod.apply_onboarding_setup({
    "provider": "aimlapi",
    "model": "openai/gpt-4o-mini",
    "api_key": "aiml-test-key",
})
```

The `provider`/`default`/`base_url` assertions are fine as-is — `_normalize_model_for_provider("aimlapi", "openai/gpt-4o-mini")` returns the model unchanged (the `provider + "/"` strip only fires for `openai`/`anthropic`, `api/onboarding.py:259-265`), and the `default_base_url` write goes through the `elif provider_meta.get("default_base_url")` branch at `:1010`. Only the key assertions are wrong.

CI hasn't reported checks on this PR yet; once it runs, expect that one red. Worth fixing before merge so the green run actually exercises the new preset.

D1m7asis:
Follow-up changes pushed in f0906eef:

- Renamed the visible provider label to `AI/ML API`.
- Fixed the probe URL side effect: `_runOnboardingProbe()` now uses an effective base URL without mutating `ONBOARDING.form.baseUrl`, and provider switches reset probe state/base URL to the selected provider default.
- Shows the probe/base URL UI for providers with `probe_models`, so AI/ML API users get the same Testing/Connected/Error feedback while live `/models` discovery runs.
- Added a provider-specific 2 MB probe cap for `aimlapi` only. The default 256 KB cap remains unchanged for self-hosted/custom URLs, but AI/ML API's real `/v1/models` catalog can exceed 256 KB.

Re-checked:
- `python -m pytest tests\test_custom_providers_in_panel.py tests\test_sprint39.py tests\test_issue1499_onboarding_probe.py::TestIssue1499OnboardingProbe::test_large_cloud_catalog_allowed_for_aimlapi_only`
- `npm run lint:runtime`
- `python -m ruff check api\onboarding.py tests\test_sprint39.py tests\test_issue1499_onboarding_probe.py`
- `python scripts\ruff_lint.py --diff origin/master`

D1m7asis:
Thanks for catching this. The current branch head has the test input aligned with the expected value:

```python
api_key: aiml-test-key
```

I also rebased on the latest `origin/master` and re-ran the focused checks:

- `python -m pytest tests\test_custom_providers_in_panel.py tests\test_sprint39.py tests\test_issue1499_onboarding_probe.py::TestIssue1499OnboardingProbe::test_large_cloud_catalog_allowed_for_aimlapi_only` -> 26 passed
- `npm run lint:runtime` -> passed
- `python scripts\ruff_lint.py --diff origin/master` -> passed

The branch now also includes the review follow-ups for the probe banner/base URL side effect and a provider-specific larger `/models` probe cap for `aimlapi`.

D1m7asis:
Update after the latest follow-ups:

The branch now uses a curated chat-capable AI/ML API model catalog instead of the earlier live `/v1/models` probing approach. That means the earlier note about `probe_models`, `static/onboarding.js`, and a provider-specific larger probe cap is stale and no longer describes the current PR head.

Current scope at `fe4201de`:

- adds the named `aimlapi` / `AI/ML API` preset with `AIMLAPI_API_KEY` and `https://api.aimlapi.com/v1`
- uses `gpt-5-chat` as the default model
- keeps namespaced AIML API model ids routed through the AIML API endpoint
- adds a runtime fallback that reads `AIMLAPI_API_KEY` from the active profile `.env` when needed
- documents the user-visible provider addition in `CHANGELOG.md`

Re-checked locally:

- `python -m pytest tests\test_custom_providers_in_panel.py tests\test_sprint39.py tests\test_provider_sort_order.py tests\test_model_resolver.py` -> 68 passed
- `git diff --check` -> passed
- `python scripts\ruff_lint.py --diff origin/master` -> passed

GitHub checks are not reported on this branch yet.

nesquena-hermes:
Re-read the current head `fe4201de` against `origin/master`. The scope changed materially since my last pass (the live `/v1/models` probe path and the `static/onboarding.js` changes are gone; it's now a curated offline catalog), so the earlier `probe_models` notes are stale, as you flagged. Re-validating the new shape, two things check out and one is worth a look.

## Verified sound

**Routing.** `resolve_model_provider` puts the `_PORTAL_PROVIDERS` guard *inside* the `if "/" in model_id:` branch (`api/config.py:2115-2127`):
```python
_PORTAL_PROVIDERS = {"nous", "opencode-zen", "opencode-go", "nvidia", "aimlapi"}
if config_provider in _PORTAL_PROVIDERS:
    return model_id, config_provider, config_base_url
```
So namespaced ids (`x-ai/grok-4-07-09`, `meta-llama/Llama-3.3-70B-Instruct-Turbo`) keep the full path to the aimlapi base URL, and bare ids (`gpt-5-chat`, `deepseek-chat`) never enter that branch and fall through to the default config provider/base_url. Both correct, and `gpt-5-chat` is present in the catalog so the default model resolves.

**Cross-repo premise.** The streaming fallback's whole justification is "hermes_cli has no native aimlapi". I confirmed that against the agent: `hermes_cli/providers.py` has no `aimlapi` registry entry (the provider table runs openrouter/openai/xai/qwen/zai/deepseek/... but nothing for `api.aimlapi.com`). So `resolve_runtime_provider` will return no key for this provider and the WebUI-side fallback is a legitimate bridge, not papering over a config error. The clean long-term home is an agent-side registry entry, but that's out of scope for this PR.

## Worth a look: env-resolution ordering vs #3961

`_resolve_aimlapi_api_key` (`api/streaming.py:259-275`) reads the process env *first*, profile `.env` second:
```python
api_key = os.getenv("AIMLAPI_API_KEY", "").strip()
if api_key:
    return api_key
...
env_path = Path(profile_home) / ".env"
```
In a single-profile deployment this is fine. But in a multi-profile WebUI the `os.environ` mirror is exactly the channel #3961 is trying to retire ("drop the os.environ mirror race"): if profile A's `AIMLAPI_API_KEY` got mirrored into `os.environ` and the active session is profile B, `os.getenv` returns A's key and the profile `.env` lookup is never reached. Since the call site already hands you the authoritative `_profile_home` (streaming.py:5533), reading the profile `.env` first and using `os.getenv` only as the fallback would avoid the cross-profile leak:
```python
# profile .env is authoritative for the active session; env is the fallback
```

## Minor
`_resolve_aimlapi_api_key` catches only `OSError`; a malformed (non-UTF-8) `.env` would raise `UnicodeDecodeError` and propagate. Widening to `except Exception:` matches the defensive posture of the sibling `.env` readers in `onboarding.py`.

On the test I flagged last time: the rewrite restructured that path, so treat the prior note as addressed pending the green CI run (checks still aren't reported on the branch). Nice cleanup overall; the curated-catalog direction is simpler than the probe path and the registration sites are all consistent.

D1m7asis:
Thanks for the re-read. I agree with both points:

- for multi-profile WebUI, the active profile `.env` should be authoritative and `os.environ` should only be a fallback;
- malformed `.env` content should not bubble out of the streaming path.

I’ll push a small follow-up for that and resolve the current branch conflict, then rerun the focused provider/onboarding checks.

D1m7asis:
Follow-up pushed in `035dee0a`.

What changed:

- `_resolve_aimlapi_api_key()` now treats the active profile `.env` as authoritative before falling back to `os.environ`, avoiding the cross-profile env mirror leak mentioned in the review.
- Malformed profile `.env` content is handled defensively and no longer bubbles out of the streaming path.
- Added regression coverage for profile-first lookup, malformed `.env`, and process-env fallback when no profile `.env` exists.
- Rebased on current `origin/master`; GitHub now reports the PR as mergeable.

Re-checked locally:

- `python -m pytest tests\test_sprint39.py tests\test_custom_providers_in_panel.py tests\test_provider_sort_order.py tests\test_model_resolver.py` -> 71 passed
- `python scripts\ruff_lint.py --diff origin/master` -> no new violations on added/modified lines
- `npm run lint:runtime` -> passed
- `git diff --check` -> passed

D1m7asis:
Addressed the latest Greptile feedback in `e89c568f`.

Change:
- malformed/unreadable profile `.env` no longer bubbles out and no longer blocks the `os.environ` fallback;
- updated the regression test to assert the process-env fallback value in that malformed profile case.

Re-checked locally:

- `python -m pytest tests\test_sprint39.py tests\test_custom_providers_in_panel.py tests\test_provider_sort_order.py tests\test_model_resolver.py` -> 71 passed
- `python scripts\ruff_lint.py --diff origin/master` -> no new violations on added/modified lines
- `npm run lint:runtime` -> passed
- `git diff --check` -> passed

GitHub still reports the PR as mergeable.

nesquena-hermes:
Re-read the head `e89c568f` against `origin/master`. Both points from my last pass are addressed — the resolver ordering is inverted and the malformed-`.env` path is hardened. This looks good.

## Profile `.env` is now authoritative (the #3961 concern)

`_resolve_aimlapi_api_key` now reads the active profile `.env` first and only falls back to `os.environ` (`api/streaming.py:266-283`):

```python
def _resolve_aimlapi_api_key(profile_home: str | Path | None = None) -> str | None:
    if profile_home:
        try:
            env_path = Path(profile_home) / ".env"
            if env_path.is_file():
                for raw_line in env_path.read_text(encoding="utf-8").splitlines():
                    ...
                    if key.strip() == "AIMLAPI_API_KEY":
                        return value.strip().strip('"').strip("'") or None
        except Exception:
            pass
    api_key = os.getenv("AIMLAPI_API_KEY", "").strip()
    if api_key:
        return api_key
    return None
```

That closes the cross-profile leak I flagged: if profile A's key got mirrored into `os.environ` while the active session is profile B, B's own `.env` now wins. The call site still gates the fallback on a missing key, so a real runtime-resolved key is never clobbered (`api/streaming.py:6201-6202`):

```python
if resolved_provider == "aimlapi" and not resolved_api_key:
    resolved_api_key = _resolve_aimlapi_api_key(_profile_home)
```

and `_profile_home` is the session's authoritative profile home, not the ambient process default — consistent with the #3294 note in the surrounding comment block.

## Malformed `.env` no longer bubbles

The `except Exception: pass` (widened from the earlier `OSError`-only catch) means a non-UTF-8 `.env` raising `UnicodeDecodeError` falls through to the process-env branch instead of propagating out of the streaming path. The regression test exercises exactly this (`tests/test_sprint39.py:295-302`):

```python
def test_aimlapi_runtime_key_ignores_malformed_profile_env(tmp_path, monkeypatch):
    monkeypatch.setenv("AIMLAPI_API_KEY", "aiml-process-key")
    (tmp_path / ".env").write_bytes(b"\xff\xfe\x00")
    from api.streaming import _resolve_aimlapi_api_key
    assert _resolve_aimlapi_api_key(tmp_path) == "aiml-process-key"
```

The companion `test_aimlapi_runtime_key_prefers_profile_env` (`:282-292`) sets a process-env value *and* a profile `.env` value and asserts the profile key wins, which is the direct guard against the ordering regression. Coverage for profile-only (`:269`) and process-only-no-`.env` (`:305`) cases is there too.

## Premise re-confirmed

The whole fallback exists because `hermes_cli` has no native `aimlapi` registry entry, so `resolve_runtime_provider` returns no key for it — I re-verified there's no `api.aimlapi.com` provider table entry on the agent side, so this WebUI bridge is legitimate rather than papering over a misconfig. The clean long-term home is still an agent-side registry entry, but that's out of scope here.

No remaining blockers from my side. Per policy I don't run PR-branch code, but a maintainer can confirm green with `python -m pytest tests/test_sprint39.py tests/test_custom_providers_in_panel.py tests/test_provider_sort_order.py tests/test_model_resolver.py` and the line-scoped `python scripts/ruff_lint.py --diff origin/master`.

nesquena-hermes:
Thanks @D1m7asis. We're going to decline this as a built-in onboarding preset. AIMLAPI is a third-party OpenAI-compatible aggregator, not a provider supported in the core of Hermes Agent — and our bar for a *named built-in* preset is that it's a first-party/core-supported provider. The good news: because AIMLAPI is OpenAI-compatible, users can already add it today via the **custom provider** flow (base URL `https://api.aimlapi.com/v1` + their `AIMLAPI_API_KEY`), with live model discovery through the existing `/models` probe — no code change needed.

Separately, even if it were in scope, the implementation special-cased key resolution (a bespoke `_resolve_aimlapi_api_key()` + a hardcoded `if provider=='aimlapi'` branch in the streaming hot path) rather than going through the generic per-`env_var` path — that pattern doesn't scale to the next provider. Closing for now; the custom-provider route covers the use case. Appreciate the contribution.

PR #4246, opened 2026-06-15
feat: credential pool fallback for custom providers + collapsible mod…

akrhin:
### Problem

When using `hermes auth add` to register a custom provider (e.g., Bothub, OpenRouter, Polza), the WebUI shows the provider as **"Not configured"** and refuses to display its models — even though valid credentials exist in the credential pool (`auth.json`). This happens because the backend (`config.py`, `providers.py`, `routes.py`) only checks `config.yaml` and `.env` for `base_url` and `api_key`, completely ignoring the credential pool.

Additionally, the model selection dropdown becomes unusable when there are many providers — groups are plain text headers with no way to collapse them.

#### 1. Backend: Credential pool fallback (4 commits squashed)

- **`api/config.py` — `get_available_models()`**  
  Before querying a custom provider's `/v1/models` endpoint, try the credential pool as a fallback for both `base_url` and `api_key`. If neither is in `config.yaml` but they exist in `auth.json` — use those.

- **`api/providers.py` — `_provider_has_key()`**  
  Check credential pool when determining whether a provider has a configured API key. Without this, `get_providers()` returns the provider as "no key" and models aren't rendered even when the key is available.

- **`api/providers.py` — `get_provider_api_key()`**  
  Same fallback when extracting the actual API key for downstream use (e.g., proxied requests).

- **`api/providers.py` — `get_providers()`**  
  Set `cp_has_key = True` if credentials exist in the pool for this custom provider, even when the env-var check fails.

- **`api/routes.py` — `_handle_live_models()`**  
  When fetching `/v1/models` for a custom provider, fall back to credential pool for `base_url` + `api_key` if they're missing from config. Without this, live-model discovery silently returns zero models.

**Dependency:** `htpx` must be available in WebUI's venv — `agent.credential_pool` imports it. Without it, all try/except blocks silently pass and the pool is never consulted.

#### 2. Frontend: Collapsible provider groups

- **`static/style.css`**  
  Added `::before` pseudo-element (▶/▼ indicators), `.open` class toggle, and `.model-group-body` wrapper styles. Only the `model-group` rule was modified (1 line change).

- **`static/ui.js`** — `renderModelDropdown()`  
  - Introduced `_groupOpenState` — persists collapse state across re-renders.  
  - Each group heading now creates a `.model-group-body` wrapper for its models.  
  - Heading click toggles the wrapper (show/hide) and flips ▶ to ▼ via `.open` class.  
  - **Search expands all groups** — when the user types a search term, every group opens automatically so search results aren't hidden inside collapsed sections.  
  - Default state: first group open, all others collapsed — keeps the dropdown compact on first load.

### Before / After

| Before | After |
|--------|-------|
| Provider not shown if key is only in credential pool | Works — pool is used as fallback |
| `/v1/models` returns 0 models if `base_url` not in config | Works — base_url from pool |
| Model list is one flat scroll (5+ providers = unusable) | Groups are collapsible, show/hide on click |
| Search looks through invisible models in collapsed groups | Search auto-expands everything |

### How to test

1. Add a custom provider via `hermes auth add bothub <key>` (omit `base_url` and `api_key` from `config.yaml`).
2. Restart the WebUI server and hard-reload the browser (`Ctrl+Shift+R`).
3. Open the model dropdown — the provider should appear **without** a "Not configured" badge.
4. Click the group header — it should collapse/expand.
5. Type in the search field — all groups should expand automatically.
6. Select a model — it should work end-to-end.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -5245,13 +5245,27 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                     _named_custom_groups[_slug] = (_cp_name, [])
 
                 _cp_base_url = str(_cp.get("base_url") or "").strip()
-                if _slug and _cp_base_url:
-                    _cp_api_key = str(_cp.get("api_key") or "").strip()
-                    if not _cp_api_key:
-                        _cp_key_env = str(_cp.get("key_env") or "").strip()
-                        if _cp_key_env:
-                            _cp_api_key = str(os.getenv(_cp_key_env) or "").strip()
+                _cp_api_key = str(_cp.get("api_key") or "").strip()
+                if not _cp_api_key:
+                    _cp_key_env = str(_cp.get("key_env") or "").strip()
+                    if _cp_key_env:
+                        _cp_api_key = str(os.getenv(_cp_key_env) or "").strip()
+                # Fallback: check credential pool for both api_key and base_url
+                if (not _cp_api_key or not _cp_base_url) and _slug:
+                    try:
+                        from agent.credential_pool import load_pool
+                        _pool = load_pool(_slug)
+                        if _pool and _pool.has_credentials():
+                            _entry = _pool.select()
+                            if _entry:
+                                if not _cp_api_key:
+                                    _cp_api_key = _entry.runtime_api_key
+                                if not _cp_base_url:
+                                    _cp_base_url = str(getattr(_entry, "base_url", "") or "").strip()
+                    except Exception:
+                        pass
 
+                if _slug and _cp_base_url:
                     # Check if user has configured models in config.yaml —
                     # configured models take priority over live /v1/models
                     # discovery (same as hermes-agent model_switch.py Section 4
--- a/api/providers.py
+++ b/api/providers.py
@@ -983,6 +983,13 @@ def _provider_has_key(provider_id: str) -> bool:
         env_value = os.getenv(env_var)
         if _provider_value_counts_as_api_key(provider_id, env_value):
             return True
+        try:
+           from agent.credential_pool import load_pool
+           pool = load_pool(provider_id)   # provider_id == "custom:bothub" for custom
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Thanks @akrhin — the credential-pool fallback here addresses a real bug (custom providers registered via `hermes auth add` showing "Not configured" in the WebUI despite valid `auth.json` credentials), and that part is genuinely useful. The main thing blocking a fast merge is that this PR **bundles two unrelated changes**, and they have very different review paths:

### Please split into two PRs

1. **Credential-pool fallback** (`api/config.py`, `api/providers.py`, `api/routes.py`) — the backend fix. This is the high-value, fast-to-ship half. As its own PR it can go through the standard gate quickly.

2. **Collapsible model groups in the dropdown** (`static/ui.js`, `static/style.css`) — a UI feature. Visible-UI changes need a maintainer screenshot sign-off (@nesquena), so bundling it here holds up the backend fix.

Splitting lets the credential-pool fix ship on its own merits without waiting on the UI review.

### On the backend half, two things to tighten before it's gate-ready

- **Indentation:** the new `from agent.credential_pool import load_pool` blocks mix indentation widths (some lines look 3-space). Please normalize to the file's 4-space standard so it stays lint-clean (`ruff` will flag it).
- **Cross-layer import:** the WebUI layer importing `agent.credential_pool` inside the request path — confirm that's the intended boundary (is `load_pool` already imported elsewhere in `api/`?), and that the `try/except Exception: pass` around it can't silently swallow a real misconfiguration the user should see. A narrow `except (ImportError, <specific>)` is safer than a blanket swallow.

Once it's split and the backend half is lint-clean, the credential-pool fix should move quickly. Flagging the collapsible-dropdown UI for @nesquena's screenshot review on the separate PR.

Issue #3957, opened 2026-06-11
bug(profiles): non-default profile — Settings → Providers times out + only default models appear (per-client profile not applied to /api/providers & /api/models)


nesquena-hermes:
## Summary

When the WebUI is set to a **non-default profile**, two things break that work fine on the default profile:

1. **Settings → Providers fails to load.** It shows a toast `Request timed out. Please try again.` and the panel body renders `Failed to load providers: Request timed out. Please try again.`
2. **The configured providers/models for that profile do not appear.** The composer model picker shows only the *default* profile's models; every other provider/model configured in the active profile's `config.yaml` / `.env` is missing.

Both work correctly on the default profile.

![Settings → Providers timeout on a non-default profile](https://raw.githubusercontent.com/nesquena-hermes/hermes-webui/issue-assets/issue-assets/profile-providers-timeout.png)

## Reported by

@Mathews Star in the community Discord #webui channel (two reports, ~16:51 and ~17:00). Verbatim:

> When I try to access the Providers section in Settings to manage my API keys, the interface fails to load and shows: "Request timed out. Please try again." / "Failed to load providers: Request timed out. Please try again."

> When I am using the default profile, everything works as expected. However, when I switch or chat using other profiles, the available AI providers and models do not appear. Instead, I only see the models belonging to the default provider, while all other configured providers and models are completely missing. This used to work perfectly fine before, but now I'm also getting a "Request timed out / Failed to load providers" error.

(Screenshot shows the active profile in the tab title as a non-default profile; the Spanish "AJUSTES" locale is incidental.)

## Root cause

Both symptoms share **one** root cause: WebUI profile switching is **per-client / cookie-scoped** (`process_wide=False`, issue #798), but two pieces of provider/model machinery resolve credentials and cache from the **process-global default profile**, not the per-request profile.

### Facet A — `Failed to load providers` timeout (`/api/providers`)

- `/api/providers` → `get_providers()` (`api/providers.py:2007`). For OAuth providers, and for unconfigured providers that fall through to the `get_auth_status()` probe (`api/providers.py:~2061`, `~2133`), it does a **network round-trip per provider, serially**. The code comment at `api/providers.py:2126-2128` explicitly acknowledges this latency.
[...]

Issue #1401, opened 2026-05-01
Silent credential self-heal on 401: WebUI should fix auth.json drift without user intervention


franksong2702:
# Background — why this happens

Hermes stores API keys in **four separate locations**:

| Location | Purpose | Who writes it |
|----------|---------|---------------|
| `~/.hermes/.env` | Canonical API key file | User (manual edit / first-time setup) |
| `~/.hermes/auth.json` | Credential pool cache | Hermes Gateway (reads `.env` once, never re-reads) |
| `~/.hermes/profiles/<name>/.env` | Per-profile `.env` files | User (during profile clone / migration) |
| `os.environ` | Runtime environment | Shell config / process env |

The credential pool in `auth.json` is **cache-first**: Hermes reads it once from `.env` (or `os.environ`), stores the key, and never re-checks whether the source has changed. When a user later updates their `.env` key — or a profile `.env` file has a stale copy — `auth.json` silently stays out of sync.

Result: user thinks "I have an OpenRouter key set up" → tries to use it in WebUI → gets a cryptic 401 → has no idea why.

This is not hypothetical. We traced a real 401 through the entire stack. The root cause was a stale `test-key-fresh` placeholder living in `~/.hermes/profiles/personal/.env`, which `auth.json` had cached and was using despite the main `.env` having the real key.

# What we filed upstream (Hermes Agent)

**Issue #18354** — reported the credential cache invalidation bug
https://github.com/NousResearch/hermes-agent/issues/18254

**PR #18256** — fixes `credential_pool.py` to prefer `.env` over `os.environ` during seeding
https://github.com/NousResearch/hermes-agent/pull/18256

**Issue #18359** — feature request: `hermes auth sync` command to detect/fix credential drift
https://github.com/NousResearch/hermes-agent/issues/18359

**PR #18361** — implements `hermes auth sync --fix` (detect + repair across all four sources)
https://github.com/NousResearch/hermes-agent/pull/18361

# What changes if upstream PRs are merged

- PR #18256: reduces one source of drift (`.env` vs. `os.environ` priority) but does NOT eliminate drift from profile `.env` files or the cache-first design.
- PR #18361: gives power users a CLI command (`hermes auth sync --fix`) to manually repair drift. Still requires the user to (a) know the command exists, (b) open a terminal, (c) run it.

**Neither PR solves the problem for a WebUI user** who just wants to chat. WebUI users shouldn't need to know what `auth.json` is.

# What we're proposing for the WebUI

[...]

PR #4247, opened 2026-06-15
fix: credential pool fallback for custom providers

akrhin:
## Problem

Custom providers configured via `hermes auth add` (Bothub, OpenRouter, etc.) store their API keys and base URLs in `auth.json` (credential pool), not in `config.yaml`. The backend's provider detection (`_provider_has_key`), key retrieval (`_get_provider_api_key`), model discovery (`get_available_models`), live model fetching (`_handle_live_models`), and provider listing (`get_providers`) all ignored the credential pool, causing:

- **401 Unauthorized** on every API call to a custom provider set as default
- Provider showing as **"not configured"** in the model UI
- **Missing models** in the dropdown

### `api/providers.py`

- **`_provider_has_key()`**: Moved the credential pool check from **inside** the `if env_var:` block to **after** it. Custom providers (`custom:bothub`, etc.) have no known env-var name → `_provider_env_var_for()` returns `None` → the entire `if env_var:` block was skipped, making the pool fallback unreachable for the exact case it was meant to cover.
- **`_get_provider_api_key()`**: Added credential pool lookup as a final fallback after checking `custom_providers[]`. Changed `pool.peek()` → `pool.select()` so the active credential entry (not the first) is used.
- **`get_providers()`**: Added pool check for `has_key`, so custom providers with keys in `auth.json` correctly show as configured in the UI.
- All three: narrowed `except Exception: pass` → `except ImportError: pass` to avoid masking unrelated runtime errors.

### `api/config.py`

- **`get_available_models()`**: Added credential pool fallback for both `base_url` and `api_key` when neither is found in `config.yaml`'s `custom_providers[]`. Refactored to resolve `_cp_api_key` unconditionally (not only when `_cp_base_url` is present) so the fallback can supply either field.

### `api/routes.py`

- **`_handle_live_models()`**: Removed a fragile `sys.path` hack that manually constructed a path to `~/.hermes/hermes-agent/` and inserted it into `sys.path` to import `load_pool`. The import works without it when the WebUI is installed properly. Replaced `except Exception` → `except ImportError`.

## Testing

- ✅ Provider shows as "configured" after `hermes auth add bothub`
- ✅ Models visible in dropdown
- ✅ No **401** on chat/completion requests to custom providers
- ✅ 401 errors in LCM compression summaries resolved

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -5245,13 +5245,27 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                     _named_custom_groups[_slug] = (_cp_name, [])
 
                 _cp_base_url = str(_cp.get("base_url") or "").strip()
-                if _slug and _cp_base_url:
-                    _cp_api_key = str(_cp.get("api_key") or "").strip()
-                    if not _cp_api_key:
-                        _cp_key_env = str(_cp.get("key_env") or "").strip()
-                        if _cp_key_env:
-                            _cp_api_key = str(os.getenv(_cp_key_env) or "").strip()
+                _cp_api_key = str(_cp.get("api_key") or "").strip()
+                if not _cp_api_key:
+                    _cp_key_env = str(_cp.get("key_env") or "").strip()
+                    if _cp_key_env:
+                        _cp_api_key = str(os.getenv(_cp_key_env) or "").strip()
+                # Fallback: check credential pool for both api_key and base_url
+                if (not _cp_api_key or not _cp_base_url) and _slug:
+                    try:
+                        from agent.credential_pool import load_pool
+                        _pool = load_pool(_slug)
+                        if _pool and _pool.has_credentials():
+                            _entry = _pool.select()
+                            if _entry:
+                                if not _cp_api_key:
+                                    _cp_api_key = _entry.runtime_api_key
+                                if not _cp_base_url:
+                                    _cp_base_url = str(getattr(_entry, "base_url", "") or "").strip()
+                    except Exception:
+                        pass
 
+                if _slug and _cp_base_url:
                     # Check if user has configured models in config.yaml —
                     # configured models take priority over live /v1/models
                     # discovery (same as hermes-agent model_switch.py Section 4
--- a/api/config.py
+++ b/api/config.py
@@ -3856,6 +3856,44 @@ def _static_models_catalog_without_live_probes() -> dict:
 # only changes when the user adds/removes credentials, which is rare; a 24h TTL
 # is plenty safe and ensures get_available_models() cold paths are fast.
 _CREDENTIAL_POOL_CACHE: dict[str, tuple[float, "CredentialPool"]] = {}  # noqa: F821  forward-ref string annotation, resolved at runtime  # pid -> (ts, pool)
+
+
+def _has_explicit_pool_credentials(provider_id: str) -> bool:
+    """Return True when the credential pool has at least one non-ambient entry
+    for *provider_id* (i.e. not a gh-cli / GITHUB_TOKEN auto-detect).
+
+    Reuses ``_CREDENTIAL_POOL_CACHE`` so that callers on hot paths (provider
+    detection, model listing, live-model fetch) don't pay the ~10s load_pool
+    cost more than once per TTL window.
+    """
+    try:
+        from agent.credential_pool import load_pool as _load_pool
+
+        _pid = _resolve_provider_alias(provider_id)
+        _cached = _CREDENTIAL_POOL_CACHE.get(_pid)
+        if _cached is not None:
+            _cp_ts, _cp_pool = _cached
+            if (time.time() - _cp_ts) < 86400.0:
+                _all_entries = _cp_pool.entries()
+            else:
+                _cp_pool = _load_pool(_pid)
+                _CREDENTIAL_POOL_CACHE[_pid] = (time.time(), _cp_pool)
+                _all_entries = _cp_pool.entries()
+        else:
+            _cp_pool = _load_pool(_pid)
+            _CREDENTIAL_POOL_CACHE[_pid] = (time.time(), _cp_pool)
+            _all_entries = _cp_pool.entries()
+
+        return any(
+            not _is_ambient_gh_cli_entry(
+                str(getattr(e, "source", "") or ""),
+                str(getattr(e, "label", "") or ""),
+                str(getattr(e, "key_source", "") or ""),
+            )
+            for e in _all_entries
+        )
+    except ImportError:
+        return False
 _provider_models_invalidated_ts: dict[str, float] = {}  # provider_id -> timestamp of last invalidation
 
 # Disk-backed in-memory cache for get_available_models().
@@ -5253,15 +5291,17 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                 # Fallback: check credential pool for both api_key and base_url
                 if (not _cp_api_key or not _cp_base_url) and _slug:
                     try:
-                        from agent.credential_pool import load_pool
-                        _pool = load_pool(_slug)
-                        if _pool and _pool.has_credentials():
-                            _entry = _pool.select()
-                            if _entry:
-                                if not _cp_api_key:
-                                    _cp_api_key = _entry.runtime_api_key
-                                if not _cp_base_url:
-                                    _cp_base_url = str(getattr(_entry, "base_url", "") or "").strip()
+                        from api.config import _has_explicit_pool_credentials
+                        if _has_explicit_pool_credentials(_slug):
+                            from agent.credential_pool import load_pool
+                            _pool = load_pool(_slug)
+                            if _pool:
+                                _entry = _pool.select()
+                                if _entry:
+                                    if not _cp_api_key:
+                                        _cp_api_key = _entry.runtime_api_key
+                                    if not _cp_base_url:
+                                        _cp_base_url = str(getattr(_entry, "base_url", "") or "").strip()
                     except ImportError:
                         pass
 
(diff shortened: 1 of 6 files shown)


nesquena-hermes (changes requested):
Thanks @akrhin — and thanks for splitting the backend credential-pool fallback out into its own PR as requested. The direction here is right (custom providers registered via `hermes auth add` should resolve from the credential pool, not show "Not configured"), and 3 of the 4 read paths are wired correctly. One blocker before this can go through the gate, plus a couple of follow-ups to fold into the same revision.

### ⛔ Blocker — the pool fallback regresses the ambient `gh_cli` filter (turns an existing test red)

`_provider_has_key()` (`api/providers.py` ~998–1004) now does:

```python
from agent.credential_pool import load_pool
pool = load_pool(provider_id)
if pool and pool.has_credentials():
    return True
```

`Pool.has_credentials()` is just `bool(self._entries)` (`agent/credential_pool.py:459`). That counts the **ambient `gh auth token`** entry that the rest of the codebase deliberately filters out via `_is_ambient_gh_cli_entry(...)` at `api/config.py:3559` and `:4784` — specifically so **Copilot doesn't show as "configured" just because `gh` is installed** (see the docstring at `config.py:1484`). The new fallback applies no such filter, and `_provider_has_key()` runs in a per-provider loop feeding both the real Settings → Providers UI (`get_providers()`, `providers.py:2108`) and `/api/models` (`_static_models_catalog_without_live_probes`, `config.py:3607`), so this hits the live product, not just a test.

**Reproduced (not theoretical):**

```
# tests/test_issue3928_models_budget_fallback.py::test_static_catalog_budget_fallback_lists_local_providers
master parent (4abf3bf5):  1 passed
this PR head:              1 failed
  AssertionError: assert 'copilot' not in ['openai-api','custom:zenmux','anthropic','copilot']
```

Direct probe on the head: `load_pool("copilot").has_credentials()` → `True` from a single entry `source=gh_cli, label='gh auth token'` — exactly the ambient class the filter targets.

**Fix** — reuse the existing filter instead of the blind `has_credentials()`:

```python
try:
    from agent.credential_pool import load_pool
    from api.config import _is_ambient_gh_cli_entry
    pool = load_pool(provider_id)
    if pool and any(
        not _is_ambient_gh_cli_entry(
            str(getattr(e, "source", "") or ""),
            str(getattr(e, "label", "") or ""),
            str(getattr(e, "key_source", "") or ""),
        )
        for e in pool.entries()
    ):
        return True
except ImportError:
    pass
```

Apply the same explicit-entry check to the `get_providers()` fallback (`providers.py` ~2370) and `get_available_models()` (`config.py` ~5253) so the semantics stay uniform.

### Please also fold in (smaller, same revision)

1. **`runtime_api_key` accessed unguarded in two paths.** In `api/config.py` (`get_available_models`) and `api/routes.py` (`_handle_live_models`) the entry's `runtime_api_key` is read as a direct attribute, but an OAuth-style credential that only carries `access_token` will raise `AttributeError` that escapes the `except ImportError` guard. Use the safe `getattr(entry, "runtime_api_key", "")` pattern you already use in `providers.py` (greptile flagged this too). In `_get_provider_api_key` (`providers.py:1089`), prefer `runtime_api_key` first, then `access_token` — the current order inverts the nous JWT semantics.

2. **`/api/models/live` runs the new `load_pool` outside profile-env scope.** `load_pool()` resolves the profile from `os.environ["HERMES_HOME"]`, and `_handle_live_models` (`routes.py:12224`) isn't wrapped in `profile_env_for_active_request(...)` the way `/api/providers` and `/api/models` are — so on a non-default profile this reads the wrong profile's `auth.json`. (This extends a pre-existing unscoped read on that handler rather than introducing a brand-new leak, but it's worth wrapping the handler body now.)

3. **Cache bypass / latency.** The five new `load_pool()` calls bypass `_CREDENTIAL_POOL_CACHE` (which exists because `load_pool` can be ~10s for probing providers, `config.py:3854`). Since `_provider_has_key()` falls through to the fallback for *every* provider lacking an env-var key, a cold Settings/Models call can fire N uncached `load_pool()` calls on a path with a 30s frontend timeout — the very symptom #3957 was fighting. Route the new calls through the existing cache (or add a small `has_explicit_credentials(provider_id)` helper that reuses it).

### Tests

The PR currently adds none. For a credential-resolution change please add:
- a regression test that the pool fallback does **not** mark a provider configured when its only entry is ambient `gh_cli`/`GITHUB_TOKEN` (this fixes the currently-red test by construction);
- a positive test that a `custom:*` entry with a real `api_key`+`base_url` makes `_provider_has_key`, `get_providers().has_key`, and `get_available_models` all report configured;
- a per-profile test that the fallback resolves the active profile's `auth.json`, not the default home's.

Once the ambient-filter fix lands (with the red test going green) and the items above are folded in, this is ready for the gate. Appreciate the work — the core fix is genuinely useful.

nesquena-hermes:
Re-reviewed `e8cf1dab` against my CHANGES_REQUESTED list and the agent contract in `agent/credential_pool.py`. The blocker is resolved and the three follow-ups landed. One real issue greptile flagged is worth fixing in the same revision before the gate; details below.

## ✅ Blocker fixed — ambient `gh_cli` filter restored

`_provider_has_key` now routes the pool check through the new `_has_explicit_pool_credentials` helper (`api/providers.py:1002-1003`), and that helper filters ambient entries exactly the way `config.py:3559` does:

```python
return any(
    not _is_ambient_gh_cli_entry(
        str(getattr(e, "source", "") or ""),
        str(getattr(e, "label", "") or ""),
        str(getattr(e, "key_source", "") or ""),
    )
    for e in _all_entries
)
```

That makes `copilot` (whose only entry is the ambient `gh auth token`) report unconfigured again, so `test_issue3928_models_budget_fallback.py::test_static_catalog_budget_fallback_lists_local_providers` passes by construction. Follow-ups 1-3 also addressed: `getattr(entry, "runtime_api_key", "")` guards the OAuth-only case, `/api/models/live` is wrapped in profile scope, and the helper reuses `_CREDENTIAL_POOL_CACHE` (config.py:3873-3884) so the cold-N-provider latency path is gone. The `get_available_models` detection loop (config.py:4801-4829) is also consistent — it iterates the actual stored pool keys and maps each to its canonical slug only after reading the right pool.

## ⛔ Remaining: `_get_provider_api_key` loads an unresolved-alias pool

`_get_provider_api_key` gates on the helper, then loads the pool with the **unresolved** id (`api/providers.py:1086-1091`):

```python
if _has_explicit_pool_credentials(provider_id):   # resolves alias internally → e.g. "glm"→"zai"
    from agent.credential_pool import load_pool
    pool = load_pool(provider_id)                 # NOT alias-resolved
    if pool:
        entry = pool.select()
```

The helper checks `_resolve_provider_alias(provider_id)` (config.py:3872), but `load_pool` only lowercases — it does **not** canonicalize (`agent/credential_pool.py:2158`):

```python
def load_pool(provider: str) -> CredentialPool:
    provider = (provider or "").strip().lower()
    raw_entries = read_credential_pool(provider)
    ...
```

So for an aliased provider — `glm`→`zai`, `github`→`copilot`, `grok`→`xai`, `qwen`→`alibaba` (the table at config.py:916-959) — the helper reports configured (it read the `zai` pool) but `load_pool("glm")` then reads a **different, empty** pool, `pool.select()` returns nothing, and the function returns `None`. Net: the provider shows "configured" in Settings but the key resolution silently fails → a 401 at request time. This doesn't bite the PR's primary `custom:*` case (those aren't in the alias table, so resolved == raw), only the canonical-alias providers — but it's a live inconsistency, and it's the one path here where detection and retrieval can read different pools.

**Fix** — resolve once and reuse, so detection and retrieval read the same pool:

```python
from api.config import _has_explicit_pool_credentials, _resolve_provider_alias
if _has_explicit_pool_credentials(provider_id):
    from agent.credential_pool import load_pool
    pool = load_pool(_resolve_provider_alias(provider_id))
    if pool:
        entry = pool.select()
        if entry:
            key = getattr(entry, "runtime_api_key", "") or getattr(entry, "access_token", "")
            if key:
                return key
```

## Test to add
A regression case that stores a real key under the **canonical** slug (`zai`) and asserts `_get_provider_api_key("glm")` returns it (not `None`) would lock the alias path in — the current suite covers `custom:*` (resolved == raw) but not the alias-divergence case greptile found.

Blocker's clear and the latency/profile follow-ups are in good shape. Just the alias-resolution mismatch in the key-retrieval path to close, then this is ready for the gate.

nesquena-hermes:
Shipped in **v0.51.461** (Release PV — "custom-provider keys recognized from the pool"). 🚀

Thanks for the thorough rework @akrhin — all four of the original review findings converged, and the final head resolved the alias-before-`load_pool` consistency issue cleanly. We took it through the full release gate and shipped it.

**Two fixes we applied on top before merging (both maintainer-side):**

1. **Cross-profile cache leak** (Codex caught it). `_CREDENTIAL_POOL_CACHE` was keyed by provider id alone — pre-existing on master, but your new `_has_explicit_pool_credentials()` detection reader made it bite the "configured" badge: a custom provider configured under profile A would show `has_key=True` under profile B (which has no key) → 401 at request time. We keyed the cache by `(active-profile auth-store path, provider id)` at all get/set/evict sites and added a cross-profile isolation regression test.

2. **Two bare `_entry.runtime_api_key` reads** (Opus caught it) at `config.py` + `routes.py` — made them `getattr`-safe to match the guard you already applied at `providers.py:1096`, so OAuth-only pool entries don't raise `AttributeError`.

**Gate:** full suite 9223 passed; Codex SAFE TO SHIP (cross-profile leak verified sealed); Opus ship-it (keying uniform across all sites, #1401 self-heal path preserved). Attribution preserved via Co-authored-by. Closing as merged.

(Note for CI watchers: the profile-scoping regression test initially failed CI shard 1 because it imported the bundled `agent.credential_pool` directly, which isn't installed in CI — fixed by injecting a stub module so the test runs identically locally and on CI.)

Issue #3664, opened 2026-06-05
Make local Windows full-suite pytest runs a useful signal


rodboev:
I hit this while working on a WebUI PR from Windows: the GitHub Linux matrix was green, but a local `pytest tests/ -v --timeout=60` run on Windows was not a useful gate because it failed across many unrelated platform assumptions.

This is not about one feature test failing. The local run completed with roughly:

```text
91 failed, 7677 passed, 20 skipped, 2 xfailed, 1 xpassed
```

The failures were broad Windows/POSIX mismatches, for example:

- POSIX-only assumptions:
  - `test_bootstrap_foreground.py` expects `/usr/bin/python3` / POSIX exec behavior.
  - `test_tls_support.py` imports or depends on `fcntl`.
  - `test_issue1574_cron_profile_lock.py` expects a multiprocessing `fork` context.
- POSIX permission assertions:
  - Several credential/session permission tests expect chmod-style `0o600`, but Windows reports modes like `0o666`.
- Path separator / path-contract mismatches:
  - Expected POSIX-ish strings such as `/custom/hermes`, but Windows-local paths produced `\custom\hermes`.
  - Some API/workspace assertions expected `subdir/file.txt` and received `subdir\file.txt`.
- Windows process and shell behavior:
  - `WinError 193`, `WinError 206`, and socket abort cases appeared in unrelated tests.
- Workspace/symlink semantics:
  - Some symlink containment/listing tests fail under Windows semantics or privilege constraints.

There was also a cleanup problem after the full run. The suite left stale server/bash/python processes and locked files under a state dir like:

```text
C:\Apps\hermes\webui-test-608f3af5
```

That made later focused test runs fail before their test bodies executed, inside `tests/conftest.py` while trying to remove `TEST_STATE_DIR` with `shutil.rmtree(...)`. This continues the same class of Windows harness cleanup work as #3534, which handled a narrower orphan-process issue and has already been merged/cherry-picked and closed. The broader issue here is that the Windows full-suite run still does not recover cleanly from platform mismatches and leftover state.

What I think would make this actionable:

1. Decide the intended contract for Windows local testing.

   I do not think this needs to mean "the whole suite must be semantically portable to Windows". A smaller first step would be: Windows can run the suite and get a meaningful signal, with explicitly POSIX-only tests skipped instead of failing noisily.

2. Add platform markers for tests that are truly POSIX-only.

   Examples:

   ```python
   import multiprocessing
   import sys
   import pytest

[...]

PR #4255, opened 2026-06-15
test(#3664): relax strict POSIX mode assertions on Windows

rodboev:
## Thinking Path

- The Windows umbrella in [#3664](https://github.com/nesquena/hermes-webui/issues/3664) includes a real class of false failures where runtime tests assert exact POSIX `0o600` bits on hosts that do not expose that as a stable filesystem contract.
- The underlying security intent is still important, and current product code still calls `chmod` in the sensitive write paths. The suite already has source-level pins for that intent.
- This slice keeps the exact runtime mode checks on POSIX but narrows the Windows expectation to a successful write/repair plus file existence, which is the strongest portable runtime contract available here.

## Contract Routing

- Contributor test harness only: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `README.md`, and `TESTING.md` define the security-sensitive test contract this PR is clarifying.

## What Changed

- `tests/test_auth_session_persistence.py`: split the `.sessions.json` assertion into exact POSIX mode checks vs Windows-safe existence/write checks.
- `tests/test_issue1910_login_attempt_persistence.py`: applied the same POSIX-vs-Windows expectation split to login-attempt persistence.
- `tests/test_security_redaction.py`: kept the runtime startup repair coverage while avoiding exact POSIX mode assertions on Windows.
- `tests/test_v050257_opus_followups.py`: kept the source-level chmod pin unchanged and narrowed only the runtime `auth.json` mode assertion on Windows.

## Why It Matters

This keeps the security intent covered without turning every Windows run into noise from a filesystem contract the host does not actually provide. POSIX still keeps the strict mode-bit assertions.

## Verification

```text
pytest tests/test_auth_session_persistence.py tests/test_issue1910_login_attempt_persistence.py tests/test_security_redaction.py tests/test_v050257_opus_followups.py -v --timeout=60
```

## Risks / Follow-ups

- This PR must not weaken POSIX coverage. Exact `0o600` runtime assertions stay in place there.
- If maintainers want stronger Windows secrecy checks later, that should be a follow-up that inspects ACLs explicitly rather than pretending POSIX mode bits are portable.

Refs #3664

## Model Used

GPT-5.4 via MiMoCode

### DIFF
--- a/tests/test_auth_session_persistence.py
+++ b/tests/test_auth_session_persistence.py
@@ -79,7 +79,7 @@ class TestSessionPersistence(unittest.TestCase):
         self.assertIn("valid_token", auth._sessions)
 
     def test_sessions_file_permissions(self) -> None:
-        """Sessions file must be owner-read-only (0600)."""
+        """Sessions file must stay strict on POSIX and still be created on Windows."""
         auth.create_session()
         # Check the path auth actually writes to (auth._SESSIONS_FILE is computed
         # from api.config.STATE_DIR at import time). Asserting against a local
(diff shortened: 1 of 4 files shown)


nesquena-hermes:

PR #4256, opened 2026-06-15
test(#3664): harden Windows pytest harness cleanup

rodboev:
## Thinking Path

- The Windows failures in [#3664](https://github.com/nesquena/hermes-webui/issues/3664) are not all product bugs; some are harness failures where one bad run leaves the next rerun with a locked state dir or stale port owner.
- `tests/conftest.py` already captures boot logs and retries once, but it still frees the port with `fuser` and deletes the state dir with a plain `shutil.rmtree()`, which is the wrong contract on native Windows.
- The smallest useful slice is to harden only the harness cleanup layer so later focused Windows reruns fail on real regressions instead of stale fixture state.

## Contract Routing

- Contributor test harness only: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `README.md`, and `TESTING.md` describe the local verification contract this PR is tightening.

## What Changed

- `tests/conftest.py`: replaced the fixture's POSIX-only `fuser` cleanup and plain `shutil.rmtree()` calls with Windows-aware port-owner, process-tree, and retrying state-dir cleanup helpers.

## Why It Matters

Windows contributors need a local pytest rerun to reflect the code they just changed, not residue from the previous failed run. Hardening the fixture cleanup removes a false-failure class without changing product behavior.

## Verification

```text
pytest tests/test_windows_native_support.py tests/test_ctl_script.py -q --timeout=60
pytest tests/conftest.py --collect-only -q
```

## Risks / Follow-ups

- The cleanup helpers must stay tightly scoped to the fixture's own PID and `TEST_PORT`; broad process killing would be worse than the current failure mode.
- This does not close the umbrella issue. It only clears the stale-process and locked-state failure class.

Refs #3664

## Model Used

GPT-5.4 via MiMoCode

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -601,6 +601,114 @@ def _server_boot_diagnostic(headline, log_path):
     return "\n".join(parts)
 
 
+def _kill_process_tree(pid):
+    """Best-effort kill for a known fixture-owned or port-owning PID."""
+    if not pid or pid <= 0:
+        return
+    try:
+        if sys.platform == "win32":
+            subprocess.run(
+                ["taskkill", "/PID", str(pid), "/T", "/F"],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                timeout=5,
+                **({"creationflags": subprocess.CREATE_NO_WINDOW}),
+            )
+            return
+        import signal
+
+        os.kill(pid, signal.SIGKILL)
+    except Exception:
+        pass
+
+
+def _kill_port_owner(port):
+    """Best-effort free of TEST_PORT, using a Windows-native owner lookup."""
+    try:
+        if sys.platform != "win32":
+            subprocess.run(["fuser", "-k", f"{port}/tcp"], capture_output=True, timeout=5)
+            return
+
+        creationflags = {"creationflags": subprocess.CREATE_NO_WINDOW}
+        ps_cmd = (
+            "$port = " + str(port) + "; "
+            "$conn = Get-NetTCPConnection -LocalPort $port -ErrorAction SilentlyContinue | "
+            "Where-Object { $_.LocalAddress -in @('127.0.0.1', '::1', '0.0.0.0', '::', '::ffff:127.0.0.1') } | "
+            "Select-Object -First 1 -ExpandProperty OwningProcess; "
+            "if ($conn) { $conn }"
+        )
+        proc = subprocess.run(
+            ["powershell", "-NoProfile", "-NonInteractive", "-Command", ps_cmd],
+            capture_output=True,
+            text=True,
+            timeout=5,
+            **creationflags,
+        )
+        pid_text = proc.stdout.strip()
+        if not pid_text:
+            proc = subprocess.run(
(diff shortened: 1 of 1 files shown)


nesquena-hermes:

nesquena-hermes:
Pulled the branch at `ac13663a` and read `tests/conftest.py` against `origin/master`. The harness hardening is solid — `_kill_port_owner` with the PowerShell→netstat fallback and the `onexc`/`onerror` signature shim in `_rmtree_retry` are the right shape for native Windows, and routing the boot-retry teardown through `_kill_process_tree(proc.pid)` is a clear improvement over the bare `proc.kill()`. One regression on the **POSIX** path is worth fixing before merge.

## The post-yield teardown loses its existing failure guard

On `origin/master` the session-fixture finalizer swallows cleanup failures deliberately (conftest.py:792-795):

```python
try:
    shutil.rmtree(TEST_STATE_DIR)
except Exception:
    pass
```

This PR replaces that with an unguarded call (conftest.py:898):

```python
yield proc

proc.terminate()
...
_rmtree_retry(TEST_STATE_DIR)
```

…and `_rmtree_retry` does **not** preserve the swallow. On POSIX it's a raw `shutil.rmtree(target)` with no guard (conftest.py:680):

```python
if sys.platform != "win32":
    shutil.rmtree(target)
    return
```

and on Windows it re-raises after 5 attempts (conftest.py:716):

```python
raise RuntimeError(f"Could not remove {target} after {attempts} attempts.{leftover_note}") from last_exc
```

Because this call runs **after `yield`**, it executes in the session-scoped fixture finalizer. A teardown-time `rmtree` failure — a still-draining child holding a file under `TEST_STATE_DIR`, an NFS `.nfs*` lock, a slow handle release on the very Windows hosts this PR targets — now propagates out of the finalizer, and pytest reports that as a session **ERROR even when every test passed**. That's a behavior change from the master contract at this site, and it's exactly the "one bad run poisons the result" failure mode #3664 is trying to remove, just relocated to teardown.

Note the pre-start cleanup at conftest.py:736 is a different contract — there a failure *should* surface, because a dirty state dir invalidates the run. It's only the post-`yield` finalizer that was intentionally best-effort.

## Suggested fix
Give `_rmtree_retry` a `best_effort` flag (or wrap the single finalizer call), so the post-yield site stays swallowed while the pre-start site keeps failing loud:

```python
def _rmtree_retry(path, *, best_effort=False):
    ...
    # final failure:
    if best_effort:
        return
    raise RuntimeError(...)
```

and on POSIX likewise guard the `best_effort` branch rather than letting the bare `shutil.rmtree(target)` raise. Then call `_rmtree_retry(TEST_STATE_DIR, best_effort=True)` at line 898 and leave line 736 strict. That keeps the Windows robustness you're adding without turning a clean test session into an ERROR on a flaky teardown.

Everything else in the diff looks correct — `_clear_readonly` clearing the read-only bit before retry is the standard Windows `rmtree` remedy, and gating `creationflags=CREATE_NO_WINDOW` behind `sys.platform == "win32"` keeps it from touching POSIX runs.

PR #4257, opened 2026-06-15
test(#3664): use the discovered Python path in bootstrap foreground tests

rodboev:
## Thinking Path

- The bootstrap foreground path already discovers the launcher interpreter dynamically in product code, but the structural tests in [#3664](https://github.com/nesquena/hermes-webui/issues/3664) still pin a literal `/usr/bin/python3` string.
- That literal path is not the contract. The real invariant is that foreground mode uses the discovered interpreter consistently when it reaches `os.execv()` on POSIX and `Popen()` on Windows.
- The smallest useful slice is to remove the hardcoded test double from the foreground tests while leaving the intentional shebang-modeling test in `test_bootstrap_discover_agent.py` alone.

## Contract Routing

- Contributor test harness only: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `README.md`, and `TESTING.md` cover the bootstrap and local verification contract this PR is narrowing.

## What Changed

- `tests/test_bootstrap_foreground.py`: switched the foreground fixtures to use `sys.executable` for the discovered launcher path and updated the execv assertions to compare against that dynamic stub instead of `/usr/bin/python3`.

## Why It Matters

This removes a fake POSIX-only assumption from the suite without changing any product behavior. Windows contributors get a cleaner signal, and the test now matches the runtime contract more closely.

## Verification

```text
pytest tests/test_bootstrap_foreground.py tests/test_bootstrap_discover_agent.py tests/test_windows_native_support.py -v --timeout=60
```

## Risks / Follow-ups

- This slice is intentionally test-only. It does not change bootstrap runtime behavior.
- The shebang-modeling negative test in `tests/test_bootstrap_discover_agent.py` should keep its literal POSIX interpreter path because that one is intentionally checking the shebang walk logic.

Refs #3664

## Model Used

GPT-5.4 via MiMoCode

### DIFF
--- a/tests/test_bootstrap_foreground.py
+++ b/tests/test_bootstrap_foreground.py
@@ -231,10 +231,11 @@ class TestMainForegroundRouting:
     def stub_main_dependencies(self, monkeypatch, tmp_path):
         """Stub out everything main() calls except the routing decision."""
         import bootstrap as bs
+        python_exe = sys.executable
         monkeypatch.setattr(bs, "ensure_supported_platform", lambda: None)
         monkeypatch.setattr(bs, "discover_agent_dir", lambda: tmp_path / "agent")
         monkeypatch.setattr(bs, "hermes_command_exists", lambda: True)
-        monkeypatch.setattr(bs, "discover_launcher_python", lambda *a: "/usr/bin/python3")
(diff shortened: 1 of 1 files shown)


nesquena-hermes:

PR #4259, opened 2026-06-15
test(#3664): mark Windows-incompatible test capabilities explicitly

rodboev:
## Thinking Path

- The Windows umbrella in [#3664](https://github.com/nesquena/hermes-webui/issues/3664) should keep real regressions visible, but a few current failures are just capability mismatches: `fcntl`-based pipe reads and test harnesses that unconditionally require `fork`.
- Those tests should not silently pass, and they should not take the whole module down on Windows either; the right contract is an explicit skip with a reason naming the missing capability.
- This slice adds shared capability markers and applies them only to the currently unavoidable cells, leaving the spawn-safe coverage intact.

## Contract Routing

- Contributor test harness only: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `README.md`, and `TESTING.md` define the local test semantics this PR makes explicit.

## What Changed

- `tests/conftest.py`: added shared `requires_fork` and `requires_fcntl` markers with capability-specific skip reasons.
- `tests/test_tls_support.py`: marked the one `fcntl`-dependent TLS fallback warning test as capability-bound.
- `tests/test_issue1574_cron_profile_lock.py`: marked only the fork-dependent harness tests, leaving spawn-safe cron/profile coverage unchanged.

## Why It Matters

Explicit capability skips make a native Windows run more trustworthy: reviewers can tell the difference between a real regression and a test that simply requires a host primitive Windows does not provide.

## Verification

```text
pytest tests/test_tls_support.py tests/test_issue1574_cron_profile_lock.py -v --timeout=60
```

## Risks / Follow-ups

- Over-skipping would hide real Windows regressions, so this PR must stay limited to the currently unavoidable `fcntl` and outer-harness `fork` cells.
- A later follow-up can replace the TLS warning read path with a portable helper and remove that skip entirely.

Refs #3664

## Model Used

GPT-5.4 via MiMoCode

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -255,6 +255,8 @@ requires_agent_modules = pytest.mark.skipif(
 def pytest_configure(config):
     config.addinivalue_line("markers", "requires_agent: skip when hermes-agent dir is not found")
     config.addinivalue_line("markers", "requires_agent_modules: skip when hermes-agent Python modules are not importable")
+    config.addinivalue_line("markers", "requires_fcntl: skip when fcntl-backed file-descriptor operations are unavailable")
+    config.addinivalue_line("markers", "requires_fork: skip when the platform lacks multiprocessing fork support")
 
 
 @pytest.hookimpl(hookwrapper=True, tryfirst=True)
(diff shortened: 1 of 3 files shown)


nesquena-hermes:

PR #4266, opened 2026-06-15
test(#3664): normalize background title skill path assertion

rodboev:
Refs #3664.

This keeps the follow-up test-only and fixes one remaining Windows-hosted local-path assertion in the background title routing coverage.

`profile_env_for_background_worker()` patches `SKILLS_DIR` to a native local path, so the test should compare it as a host filesystem path instead of expecting a slash-normalized string.

Validation:
- `D:\Repos\hermes-agent\venv\Scripts\python.exe -m pytest tests/test_title_aux_routing.py -q`

### DIFF
--- a/tests/test_title_aux_routing.py
+++ b/tests/test_title_aux_routing.py
@@ -7,6 +7,7 @@ Covers:
   - _aux_title_timeout rejects zero, negative, and non-numeric values
 """
 import os
+from pathlib import Path
 import sys
 import types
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Merged into master via #4272 (batched with the #3664 Windows pytest-harness series). Codex SAFE + full suite green; test-only, no version stamp. Thanks @rodboev!

PR #4268, opened 2026-06-15
fix(sessions): invalidate sidebar cache on WAL updates

rodboev:
Split from #4133 at maintainer request after the notification-permission fix shipped on its own in #4265.

## Thinking Path
- #4133 bundled a notification-permission UX fix with a separate sidebar-cache invalidation change.
- The maintainer shipped the notification fix independently and asked for the WAL/settings cache invalidation to be reopened as its own PR.
- The session sidebar cache currently stamps the main state DB, gateway metadata, and session index, but WAL-only writes and settings-file updates can leave cached session rows stale until some other stamp changes.

## What Changed
- `api/routes.py`: extend `_session_list_cache_source_stamp()` to include both `state.db-wal` and `SETTINGS_FILE` whenever CLI sessions are part of the cache key.
- `tests/test_session_sidebar_cache.py`: add focused regression tests that prove WAL-file writes and settings-file writes each invalidate the source stamp.

## Why It Matters
This keeps sidebar session rows fresh when SQLite only touches the WAL file or when settings changes alter session visibility. Without those extra stamps, users can keep seeing stale session state behind a warm cache.

## Verification
- `D:\Repos\hermes-agent\venv\Scripts\python.exe -m pytest tests/test_session_sidebar_cache.py tests/test_request_diagnostics_cache.py tests/test_gateway_sync.py -v --timeout=60`
- CI/reviewer context: `pytest tests/ -v --timeout=60`

## Risks / Follow-ups
This only broadens the cache invalidation inputs; it does not change cache lifetime, rebuild logic, or the `/api/sessions` response shape.

## Model Used (AI disclosure)
GPT-5.4 via MiMoCode

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -1463,7 +1463,10 @@ def _session_list_cache_source_stamp(key: tuple) -> tuple[tuple[int, int], tuple
         state_db_path = Path(_active_state_db_path())
     except Exception:
         state_db_path = None
-    state_db_wal_path = state_db_path.with_name(f"{state_db_path.name}-wal") if state_db_path is not None else None
+    try:
+        state_db_wal_path = state_db_path.with_name(f"{state_db_path.name}-wal") if state_db_path is not None else None
+    except Exception:
+        state_db_wal_path = None
     try:
         gateway_metadata_path = _gateway_session_metadata_path()
     except Exception:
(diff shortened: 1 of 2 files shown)


rodboev:
Follow-up pushed on this head in `b270dc42`.

- Guarded the derived `state.db-wal` path the same way the other cache-stamp path lookups are guarded.
- Routed the `SETTINGS_FILE` stamp through the same defensive local lookup so the helper no longer depends on that global being present unguarded.
- Re-ran focused coverage: `D:\Repos\hermes-agent\venv\Scripts\python.exe -m pytest tests/test_session_sidebar_cache.py tests/test_request_diagnostics_cache.py tests/test_gateway_sync.py -v --timeout=60` (71 passed).

Issue #2791, opened 2026-05-23
feat: fuzzy search for model selection dropdown


munim:
## Feature Description

Add fuzzy search / filtering to the model selection UI so users can quickly find and switch models by typing partial names, provider prefixes, or abbreviations instead of scrolling through a long dropdown list.

## Motivation

The model list in the WebUI continues to grow as new providers and model variants are added (OpenAI, Anthropic, Google, open-source models, custom providers, etc.). Currently, users must scroll through the full list to find the model they want, which is slow and error-prone — especially when:

- The list contains 50+ models across providers
- Users know the model name they want but not its exact position in the list
- Switching models frequently during a session (e.g., using a cheap model for simple tasks and a powerful one for complex ones)

This is a common UX pattern in modern developer tools — VS Code command palette, Obsidian file switcher, and Raycast all use fuzzy matching to great effect.

## Proposed Solution

Add a text input at the top of the model dropdown that supports:

1. **Substring matching** — typing `sonnet` filters to `claude-sonnet-4`, `claude-sonnet-3.5`, etc.
2. **Fuzzy matching** — typing `g4o` matches `gpt-4o`, `cs4` matches `claude-sonnet-4`
3. **Provider prefix filtering** — typing `anthropic/` or `openai/` filters to that provider
4. **Keyboard navigation** — arrow keys to move through filtered results, Enter to select
5. **Highlight matched characters** in the results so users can see why a model matched

A lightweight library like [Fuse.js](https://fusejs.io/) (~6KB) or a simple custom scorer would be sufficient — no heavy dependencies needed.

```
┌──────────────────────────────────────┐
│ 🔍 Search models...                  │
├──────────────────────────────────────┤
│ ✓ claude-sonnet-4          anthropic │
│   claude-sonnet-3.5        anthropic │
│   claude-sonnet-3.5-turbo  anthropic │
└──────────────────────────────────────┘
```

## Alternatives Considered

- **Provider submenus** — groups models by provider but still requires scrolling within each group; does not help when you know the model name
- **Recent models section** — helpful but does not solve the general search problem; could complement fuzzy search
- **Browser autocomplete** — native autocomplete is inconsistent across browsers and not styleable

## Scope / Effort Estimate

Small / Medium — touches the model selector component (likely in `static/`), adds a lightweight fuzzy matching utility. No backend changes needed since the model list is already available client-side.

## Additional Context

- Related: #698 (model modals), #1240 (provider/model source of truth)
[...]

Issue #1425, opened 2026-05-01
ux(models): model picker gets unwieldy with 3+ providers — same-name collision + list overload (sub-issue of #1240)


nesquena-hermes:
## Summary

The composer model picker (and the Settings → Default model picker) becomes hard to use once 3+ providers are configured, in two distinct ways:

1. **Same-name collision** — two providers expose the same model ID (e.g. both Anthropic and OpenRouter list `claude-sonnet-4.6`). Rows look nearly identical; the only differentiator is the (subtle) provider group heading above. Search works but only if you already know to type the provider name.
2. **List overload** — with Nous + OpenCode + OpenRouter + Ollama Cloud + a local LLM all configured, the flat list is many screens of scrolling. Even with search, the default open state is overwhelming.

This is a sub-issue of #1240 (umbrella for Provider/Model source-of-truth) — specifically the "model-picker grouping" lane @nesquena named in the May 1 comment after PR #1390 landed.

## Reported by

- @AvidFuturist (Discord, May 1 2026) — "same model same name, it's hard to distinguish provider from the model picker (search works, but provider divider is not that clear imo)"
- A separate community user (relayed via @AvidFuturist): "You start to run into list overload with several providers added too. I currently have Nous, OpenCode and Ollama Cloud configured and it's a lot to scroll through. Both in the chat selector and the default selector in settings. Maybe an option to select the provider first, then just list the models from it?"

## Current behavior

`static/ui.js:586-740` (`renderModelDropdown`) builds a flat list from the underlying `<select>`'s `<optgroup>` children. Each model row shows:

- Model name (top line)
- Model ID (smaller, beneath)
- Optional "Configured" / "Fallback N" badge

Group headings (provider names) are rendered as `.model-group` divs (`static/style.css:1352`) — small uppercase text in muted color. They blend in.

Search input (`_searchRow` at `:622`) filters across name + ID. There's no provider-aware filter chip.

There is also a "Custom model ID" input that creates a temp option with no provider context (covered in companion issue β).

## Concrete UX gaps

1. **No inline provider chip per row.** When `claude-sonnet-4.6` appears under both Anthropic and OpenRouter, the rows only differ by their distance from the (small, muted) heading above. Easy to miss.
2. **No way to collapse providers.** No accordion, no "show only this provider," no count badge per provider.
3. **No provider-first mode.** All models from all providers are listed by default. With 5+ providers, that's a lot of scroll before the user even knows what to search for.
[...]

PR #4279, opened 2026-06-15
Feat/collapsible model groups clean

akrhin:
## Problem

The model selection dropdown groups models by provider (OpenAI, Anthropic, etc.), but these groups are always expanded. Users with many providers see a long, unscannable list. There is no way to collapse groups that aren't immediately relevant.

## Changes

- **Collapsible groups** — provider group headings are now clickable. Clicking a heading toggles its model rows open/closed. A chevron indicator (▶/▼) signals the collapsible state.
- **Persistent state** — group open/close state survives across search/filter cycles (`_groupOpenState`). On a fresh search all groups auto-expand so results are visible.
- **Endpoint error hints inside group** — provider endpoint error messages now render inside the collapsible group wrapper (#4253), so they collapse/show with their group.
- **Keyboard-safe** — `_visibleModelRows` filters out rows in hidden groups, preventing the keyboard focus from jumping into collapsed sections (#2791).
- **CSS-only chevron** — the toggle indicator is purely CSS (`.collapsible::before`), no extra DOM elements.
- **Provider chip** — inline provider label on every row (#1425), so a collapsed group still gives partial orientation.

## Testing

- Verified in Chromium at 1440×900 and 390×844 viewports
- Keyboard navigation (↑↓) skips hidden group rows correctly
- Endpoint error message collapses/shows with its group
- All 33 existing tests pass (targeted UI tests)

**PR checklist:**
- [x] Vanilla JS/CSS only — no build step required
- [x] No global variable leaks (`_groupWrappers` declared with `let`)
- [x] Search auto-expands all groups for result visibility
- [x] Click-to-collapse is discoverable (chevron + cursor:pointer)
- [x] No changes to `api/providers.py` — credential pool logic is out of scope

IMG_20260616_025219
IMG_20260616_025239
IMG_20260616_025308

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -2503,7 +2503,7 @@
 .model-dropdown{display:none;position:absolute;bottom:calc(100% + 4px);left:0;min-width:280px;max-width:min(420px,calc(100vw - 32px));background:var(--surface);border:1px solid var(--border2);border-radius:10px;box-shadow:0 -4px 24px rgba(0,0,0,.4);z-index:200;overflow:hidden;max-height:320px;overflow-y:auto;}
 .model-dropdown.open{display:block;}
 .model-scope-note{position:sticky;top:0;z-index:2;padding:9px 14px;border-bottom:1px solid var(--border);color:var(--text);font-size:11px;font-weight:650;line-height:1.4;background:color-mix(in srgb,var(--surface) 82%,var(--accent-bg));box-shadow:0 1px 0 rgba(0,0,0,.12);}
-.model-group{padding:8px 14px 4px;font-size:10px;font-weight:700;letter-spacing:.04em;color:var(--muted);text-transform:uppercase;border-top:1px solid var(--border2);margin-top:2px;}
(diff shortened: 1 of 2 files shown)


akrhin (commented):
P1 — Group open-state is unconditionally overwritten on every re-render

Fixed in commit edbda826. The old code unconditionally collapsed all groups when search was cleared:

// before
if(!hasSearch&&Object.keys(_groupOpenState).length>0) _groupOpenState[groupKey]=false;
else _groupOpenState[groupKey]=true;


Now it only initialises when the key is absent, preserving the user's click state across search → clear → re-search cycles:

// after
if(hasSearch) _groupOpenState[groupKey]=true;
else if(!(groupKey in _groupOpenState)) _groupOpenState[groupKey]=false;


---

P2 — Dead CSS transitions — display:none toggle bypasses both animations

Already fixed in this revision (9f833cc0). The chevron uses a single ▶ content character with transform:rotate(90deg) on .open::before instead of swapping content values. The only transition that exists now is transition:transform .15s ease on the chevron, which fires correctly. The body stays display:none intentionally — animating max-height inside the dropdown would break scroll behaviour.

---

P2 — Inconsistent indentation inside _renderProviderEndpointHint

Already fixed in this revision (9f833cc0). The function body uses the same 2-space indent as the rest of the closure.

nesquena-hermes:
The collapsible-model-group UX from this PR shipped in **v0.51.464** ("scannable model picker", #4279) — provider groups now start collapsed except the one owning your selected model, search auto-expands matching groups (and searches the hidden overflow), the "Show all N models" affordance reveals overflow in place, and the redundant per-row provider chip is suppressed for rows already under their own heading.

The shipped implementation builds directly on your collapsible-groups work here, @akrhin — credited in the v0.51.464 CHANGELOG and release notes. Since the concept is now live in master, closing this as shipped/superseded. Thank you for the contribution! 🙏

Issue #4278, opened 2026-06-15
Custom Provider Model ID Prefix Collision in Route Normalizer


claw-io:
## Bug Description
When using models configured under a custom provider with a provider namespace prefix ending in `_cli` (for example, `gemini_cli/gemini-3-flash-preview` routed via a custom `newapi` endpoint), the composer footer's context indicator ring disappears, showing only a middle dot (`·`) inside the circle. Additionally, the context tooltip displays cumulative counts (e.g., `In: X · Out: Y`) instead of the context percentage, left capacity, and auto-compress threshold limit.

## Steps to Reproduce
1. Configure a custom model with a prefix matching a known provider (e.g., `gemini_cli/gemini-3-flash-preview` under a custom provider `custom:newapi`).
2. Start a new session with this model.
3. Observe the composer footer once the first message exchange completes.
4. The percentage indicator ring does not render, and the tooltip defaults to a generic cumulative token count.

## Expected Behavior
The WebUI should correctly resolve the model's metadata, context length (e.g., `1,048,576` tokens), and last prompt token metrics, rendering the correct usage percentage ring and auto-compression threshold.

---

### 1. Frontend Rendering Dependency (`static/ui.js`)
The context-ring percentage calculation in `static/ui.js` (`_syncCtxIndicator`) is driven by `usage.last_prompt_tokens`:
```javascript
const promptTok = usage.last_prompt_tokens || 0;
const hasPromptTok = !!promptTok;
const rawPct = hasPromptTok ? Math.round((promptTok / ctxWindow) * 100) : 0;
```
If `last_prompt_tokens` is `0` or missing, `hasPromptTok` is `false`. The percentage ring is hidden, the ring is cleared, and the text fallback dot (`\u00b7`) is displayed.

### 2. Prefix Collision during Normalization (`api/routes.py`)
On session load (`GET /api/session?resolve_model=1`), the backend resolves the session's active model for display by calling `_resolve_effective_session_model_for_display()`, which delegates to `_resolve_compatible_session_model_state()`.

To determine if the model represents a stale cross-provider reference, the model's provider namespace is normalized using the helper `_normalize_provider_id(value)`:
```python
def _normalize_provider_id(value: str | None) -> str:
    raw = str(value or "").strip().lower()
    # ...
    for prefix, normalized in (
        # ...
        ("google", "google"),
        ("gemini", "google"),
        # ...
    ):
        if raw.startswith(prefix):
            return normalized
```
[...]

PR #4280, opened 2026-06-16
fix(#4278): preserve custom provider namespace prefixes

rodboev:
## Thinking Path

- Hermes WebUI needs to preserve user-controlled custom provider namespaces even while it repairs obviously stale first-party model families.
- The current bug is not in the repair branch itself; it is earlier, where `_normalize_provider_id()` collapses any bare prefix hit into a first-party family.
- The smallest safe fix is to make that normalization token-boundary-aware so real family ids still normalize while custom prefixes like `gemini_cli` fall through unchanged.

## What Changed

- `api/routes.py`: tightened `_normalize_provider_id()` so provider-family normalization only matches exact tokens or real namespace separators.
- `tests/test_provider_mismatch.py`: added regressions covering both the helper boundary cases and the custom-provider session-model preservation path.
- Attribution: this follows the maintainer's boundary-match direction from [issue #4278](https://github.com/nesquena/hermes-webui/issues/4278#issuecomment-4713461836).

## Why It Matters

This keeps Hermes WebUI from silently swapping a user's custom-provider session model to the active default model just because the namespace happens to start with `gemini`, `claude`, or `openai`. The visible context-ring regression in the issue is just fallout from that wrong-model rewrite; fixing the classifier removes the root cause.

## Verification

```text
D:\Repos\hermes-agent\venv\Scripts\python.exe -m pytest tests/test_provider_mismatch.py -v --timeout=60
```

## Risks / Follow-ups

- This intentionally keeps the current accepted separators narrow (`-`, `:`, `/`). If a future real provider namespace needs a different boundary shape, that should be added deliberately with a reproducer.
- The change does not alter catalog routing or custom/openrouter repair behavior; it only stops unknown custom prefixes from being misclassified before that branch runs.

## Model Used

GPT 5.4 via MiMoCode

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -2914,6 +2914,14 @@ def _normalize_provider_id(value: str | None) -> str:
         return ""
     if raw in _PROVIDER_ALIASES:
         return _PROVIDER_ALIASES[raw]
+
+    def _matches_provider_family(prefix: str) -> bool:
+        if not raw.startswith(prefix):
+            return False
+        if raw == prefix:
+            return True
+        return raw[len(prefix)] in "-:/"
+
     for prefix, normalized in (
         ("openai-codex", "openai"),
         ("openai", "openai"),
@@ -2924,7 +2932,7 @@ def _normalize_provider_id(value: str | None) -> str:
         ("openrouter", "openrouter"),
         ("custom", "custom"),
     ):
-        if raw.startswith(prefix):
+        if _matches_provider_family(prefix):
             return normalized
     # Unknown prefix — return empty so callers treat it as "no match" and pass
     # the model through unchanged rather than incorrectly stripping it.
--- a/api/routes.py
+++ b/api/routes.py
@@ -2920,7 +2920,7 @@ def _normalize_provider_id(value: str | None) -> str:
             return False
         if raw == prefix:
             return True
-        return raw[len(prefix)] in "-:/"
+        return raw[len(prefix)] in "-:/."
 
     for prefix, normalized in (
         ("openai-codex", "openai"),
--- a/tests/test_provider_mismatch.py
+++ b/tests/test_provider_mismatch.py
@@ -1733,6 +1733,57 @@ def test_custom_namespace_model_always_preserved_on_custom_provider(monkeypatch)
     assert effective == "custom/my-local-llm"
 
 
+def test_provider_family_normalization_requires_a_real_namespace_boundary():
+    """Only exact family tokens or separator-qualified namespaces should
+    normalize to first-party providers (#4278)."""
+    import api.routes as routes
+
+    expectations = {
+        "gemini": "google",
+        "gemini/gemini-2.5-pro": "google",
+        "openai-codex": "openai",
+        "openrouter:my-model": "openrouter",
+        "custom:agent37": "custom",
+        "gemini_cli": "",
+        "openai_compat": "",
+        "claudeRelay": "",
+    }
+
+    for raw, expected in expectations.items():
+        assert routes._normalize_provider_id(raw) == expected, raw
+
+
+def test_custom_provider_prefix_overlap_does_not_rewrite_model(monkeypatch):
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Superseded by #4285 (b3nw), which fixes the same root cause (#4278) and shipped in **v0.51.450** (Release PK).

Thank you for this fix, @rodboev — it was a real bug and your diagnosis was correct. We went with #4285 for one specific reason: this PR's token boundary accepts `-` as a namespace delimiter (`-:/.`), but the provider-family prefix list contains `claude`, so `claude-relay/...` would still collapse to `anthropic` — re-introducing the same class of bug #4278 reports, just for a different name. #4285's `:/`-only boundary keeps `claude-relay` intact and maps the genuine hyphenated first-party IDs (`google-gemini`, `claude-code`, `openai-api`, …) explicitly.

Closing as superseded — full credit to you for catching and fixing the collision. 🙏

nesquena-hermes:
## Summary

Reading `api/routes.py` on origin/master plus this PR's diff to `_normalize_provider_id` and the new `tests/test_provider_mismatch.py` cases, the helper change is correct and matches the maintainer's boundary-match direction from #4278. But it looks **insufficient to fix the reported `gemini_cli/gemini-3-flash-preview` symptom on its own**, and the integration test doesn't exercise the path where the bug actually fires.

## Code reference

The reported model is slash-qualified and the session has no explicit `model_provider`, so display resolution (`_resolve_effective_session_model_for_display`) enters the profile branch with `profile_provider="custom:newapi"`. Inside that branch there is a *second* normalization site — a bare-prefix scan that runs `startswith` against the **full** model string, on origin/master:

```python
# api/routes.py:3482-3486 (origin/master)
model_lower = model.lower()
for bare_prefix in ("gpt", "claude", "gemini"):
    if model_lower.startswith(bare_prefix):
        model_family = _normalize_provider_id(bare_prefix)   # bare "gemini" -> "google"
        break
```

For `gemini_cli/gemini-3-flash-preview`, `model_lower.startswith("gemini")` is `True`, so `model_family` becomes `"google"` regardless of how `_normalize_provider_id` treats `gemini_cli`. The swap then fires here:

```python
# api/routes.py:3488
if model_family and model_family != _profile_provider_normalized:   # "google" != "custom"
    ...
    return _target, profile_provider, True   # silent swap to default model
```

This PR tightens `_normalize_provider_id` (fixing `model_provider_from_name = _normalize_provider_id("gemini_cli")` at routes.py:3479), but it does not touch the bare-prefix scan, so the swap at 3488 still fires for the reported input.

## Why the new test passes anyway

`test_custom_provider_prefix_overlap_does_not_rewrite_model` calls:

```python
routes._resolve_compatible_session_model_state(
    "gemini_cli/gemini-3-flash-preview",
    "custom:agent37",   # 2nd positional arg = model_provider, NOT profile_provider
)
```

The signature is `(model_id, model_provider=None, *, profile_provider=None, ...)`. Passing `"custom:agent37"` positionally sets `model_provider`, which trips the fast-path early return:

```python
# api/routes.py:3432
if not explicit_provider and not stale_codex_openai_slash_id:
    return model, requested_provider, False
```

So the test returns at 3432, never calls the monkeypatched `get_available_models`, and never reaches the profile branch where the bug lives. It would pass on origin/master with no fix at all, so it isn't guarding the regression.

## Recommendation

1. Guard the bare-prefix scan so it doesn't fire on slash-qualified namespaces, e.g. wrap routes.py:3482-3486 in `if "/" not in model:`. (This is exactly what #4285 does, in addition to the helper fix.)
2. Re-target the integration test through the failing path by passing the provider as the keyword `profile_provider=` (and `model_provider=None`), then assert `changed is False`. With `model_provider` positional it cannot catch this class of bug.

The helper-only diff here is a real improvement and the `_normalize_provider_id` boundary unit test is worth keeping, but as written this PR will not make the context ring re-appear for the reported case. Worth cross-checking against #4285, which carries both changes.

PR #3712, opened 2026-06-06
feat(models): add provider-agnostic show-all overflow groups (#3691)

rodboev:
## Thinking Path

- The current overflow behavior is split across provider-specific branches: Nous preserves a tail under `extra_models`, but other large catalogs still truncate in bespoke ways.
- The frontend already has half of the contract (`extra_models` label hydration and slash-command coverage), but `_modelData` and the picker UI stop at rendered `<option>` rows.
- The clean upstreamable slice is to generalize the overflow split in `api/config.py` and add a generic expand-in-place affordance in the picker.

## What Changed

- `api/config.py`: preserve capped provider overflow under `extra_models` instead of dropping it, while keeping Nous-specific ordering logic intact
- `static/ui.js`: render a generic "Show all N models" expander and rebuild `_modelData` after expansion
- `static/commands.js`: preserve the existing full-catalog slash-command contract
- `tests/test_issue1567_nous_picker_capacity_and_symmetry.py`: extend overflow-contract coverage
- `tests/test_issue3691_model_picker_show_all.py`: add focused picker-expansion regression tests

## Why It Matters

Today the picker can imply that a provider only has a handful of models when the backend actually knows about dozens more. This change makes the truncation explicit, keeps the initial dropdown scannable, and preserves full-catalog access without forcing users into slash commands.

## Verification

```bash
pytest tests/test_issue1567_nous_picker_capacity_and_symmetry.py tests/test_issue3691_model_picker_show_all.py -v --timeout=60
pytest tests/ -v --timeout=60
```

## Risks / Follow-ups

- Very large catalogs still need sensible initial ordering. This change generalizes the overflow UX, but it does not try to redesign provider-specific ranking beyond preserving the current Nous featured-set behavior.

## Model Used

OpenAI GPT-5 Codex

PR #4282, opened 2026-06-16
Release PG (v0.51.446): model picker show-all overflow for large catalogs (#3691)

### LINKED ISSUE #3691: enhancement(models): provider-agnostic "Show all N models" expander for any provider with too many models to show at once
## Summary

The model-selection popup caps how many models a provider shows, per-provider, with bespoke
hardcoded logic. **Generalize this:** any provider group whose model count exceeds a single
display threshold should render a default slice plus an in-dropdown **"Show all N models"**
expander that reveals the rest. One uniform rule for every provider, replacing the
provider-specific trimmers/caps that exist today.

Reported in Discord #webui:

> **@Pinched-Nerve** (2026-06-05): "Any possibility of increasing the max number of entries
> in the model selection popup? Perhaps allow the user to configure the max number of models
> per provider?"

(Screenshot shows the composer model picker — "Applies to this conversation from your next
message" — listing Nous Portal entries: Deepseek V4 Flash, Qwen3.7 MAX, Qwen3.7 Plus,
Qwen3.6 35B A3b, Kimi K2.6, MiniMax M3, …)

> **Scope decision (Nathan, 2026-06-05):**
> 1. **No configurable cap.** A config knob solves the wrong layer — it just lets the user
>    pick a different arbitrary number (set 50, still can't see model 51) and adds a setting
>    to discover/document/maintain.
> 2. **Make it provider-agnostic.** The ideal is that this works for **any** provider with
>    more models than can be shown at once by default — not a Nous-specific feature. One
>    threshold, one "show all" affordance, every provider.

## How it works today (fragmented, per-provider)

The picker payload (`_build_available_models_uncached`, `api/config.py`) returns a uniform
shape:

```python
groups: [{ "provider": str, "provider_id": str, "models": [{"id", "label"}, ...] }]
```

But trimming is done ad-hoc inside each provider's branch, differently:

- **Nous Portal** — `_build_nous_featured_set()` (`api/config.py:1423`) trims to
  `_NOUS_FEATURED_TARGET = 15` once the catalog exceeds `_NOUS_FEATURED_THRESHOLD = 25`
  (`config.py:1409`). The tail **is** shipped to the client as a per-group `extra_models`
  field, so it's reachable via `/model` autocomplete (`static/commands.js:159`) — just not
  from the dropdown.
- **OpenRouter free-tier** — a separate hardcoded `_free_cap = 30` (`api/config.py:4298`)
  that `break`s out of the fetch loop and **discards** everything past 30. The tail never
  reaches the client at all.
- **Every other provider** — no cap; the full `models` list renders.

So "max models per provider" is currently three different behaviors. The reporter's catalog
(Nous) is just the one that happens to trip the trimmer visibly.

## Proposed design: one uniform post-pass + a generic expander

There's a clean single choke point. After all provider branches have appended their groups
and just before the return — right around `groups.sort(...)` at `api/config.py:4754` —
every group exists in the uniform `{provider, provider_id, models}` shape. Add one
provider-agnostic pass there:

```
PICKER_DISPLAY_LIMIT = 25   # single threshold, all providers
for g in groups:
    if len(g["models"]) > PICKER_DISPLAY_LIMIT:
        g["extra_models"] = g["models"][PICKER_DISPLAY_LIMIT:]
        g["models"]       = g["models"][:PICKER_DISPLAY_LIMIT]
```

This:

- replaces the bespoke Nous trimmer and the OpenRouter `break`-and-discard with **one rule**;
- standardizes on the existing per-group `extra_models` contract the client already
  understands (Nous already populates it; `/model` autocomplete already consumes it);
- requires the OpenRouter branch to stop discarding the tail — collect remaining free-tier
  models instead of `break`ing — so its overflow flows into the same `extra_models` field.

Front-end: when a group has a non-empty `extra_models`, render a trailing **"Show all N
models"** row in that optgroup that, on click, appends the `extra_models` entries. The
optgroup label already carries a truncation/"+N more" count, so the signal exists — this
just makes it actionable. The expander is generic: it triggers for *any* group with overflow,
not a Nous special case.

Open sub-decisions for the builder:
- Keep Nous's *smart* featuring (flagship + vendor-priority round-robin determines **which**
  models land in the visible slice) vs. a plain "first N." The smart featuring is a nice
  default ordering; it can stay as the per-group ordering step *before* the uniform split, so
  the generic slice still picks a sensible top-N for Nous while every other provider gets a
  straightforward first-N. (i.e. ordering stays provider-aware, the cap+expander become
  universal.)
- Exact threshold value (25 is the current Nous number and a reasonable scannability ceiling).

## Explicitly out of scope (decided)

- **A configurable cap** (env var / `config.yaml` key). Rejected — solves the wrong layer.

## Possible future direction (not this issue)

- A **type-to-filter search box** inside the picker would scale to hundreds of models with no
  cap at all. Larger front-end feature; overlaps with #1240 ("harden provider/model
  source-of-truth") and #1425 ("picker gets unwieldy with 3+ providers"). The "show all"
  expander is the near-term answer; search is the long-term one.

## Scope

- Layer: WebUI. Backend: one uniform post-pass in `api/config.py` (around `config.py:4754`)
  + change the OpenRouter free-cap to collect-into-`extra_models` instead of `break`
  (`config.py:4298`); retire/fold the Nous-specific threshold into the shared pass (keep its
  ordering as the pre-split step). Front-end: generic "Show all N" expander in the picker
  render (`static/*.js`) keyed on a non-empty per-group `extra_models`.
- Related: #1567 (introduced the Nous featured-subset trim), #1538 (Nous live catalog),
  #1425 (picker overload with multiple providers), #1240 (provider/model source-of-truth RFC).

## Reported by

@Pinched-Nerve in Discord **#webui** (2026-06-05).

nesquena-hermes:
## Release PG — v0.51.446

Ships **#3712** (rodboev) — provider-agnostic model-picker "show all" overflow (#3691). Self-rebased onto v0.51.445; **full browser-tested (desktop + mobile, before/after) and maintainer screenshot-approved**. Gated Codex SAFE + Opus SHIP + full picker suite.

### What it does
A provider group with >25 models shows the first 15 + a "Show all N models" row that expands the rest in place (search already matches the hidden tail before expansion). Group heading carries the overflow count ("Provider (15 of 30)"); rows show the plain provider label. ≤25-model groups unchanged.

### 4 gate-caught issues fixed inline before merge
1. **Per-row count leak** (maintainer-caught): the "(15 of 30)" count was stamped on every row + read as nonsense post-expand → now heading-only, rows show plain label.
2. **Overflow-expand dedup** (Codex SILENT): selecting a hidden model from search → reopen → "Show all" could duplicate the row → `_appendOverflowOptionsToGroup` now moves/reuses an existing option.
3. **Heading double-count** (Opus): post-expand showed "Nous (2 of 4) (4)" → strips the decoration + counts rendered rows.
4. **Native selectors dropping the overflow tail** (Codex CORE): the server-side split moves models>15 into `extra_models` for all providers, but the Settings / Cron / Profile / Auxiliary `<select>` builders read `g.models` only → would silently hide models 16+ for large providers. Fixed all 4 to concat `models + extra_models`.

### Verified
- **Live browser** (mock 30-model provider): 15 + "Show all 15" → expand → all 30, no errors, heading "(15 of 30)", rows plain "MockCo". Desktop 1920 + mobile 390 both clean, no overflow.
- **Codex SAFE** (final): all 4 native selectors include the tail, picker dedup/search/expand correct, XSS-safe, no regression to badges/Nous-tail/custom-IDs.
- **Opus SHIP:** architecture sound (its one finding, the heading count, fixed).
- **Tests:** updated test_issue3691 — asserts heading-carries-count, rows-don't, no post-expand double-count, no dup, and native selectors include extra_models. 41 picker-area tests pass in isolation; wider suite failures are this box's process-group cascade artifact (CI authoritative).

Closes #3691.

Co-authored-by: rodboev


garyd9:
I don't see any "show all" option in the model picker after updating.

garyd9:
> I don't see any "show all" option in the model picker after updating.

Of course, I can be an idiot and forget to flush my browser cache.  Thanks to the dev.

PR #4285, opened 2026-06-16
fix(routes): add token boundary check for provider normalization (#4278)

b3nw:
# fix(routes): restrict provider normalization prefix matching using token boundaries (#4278)

**Closes:** #4278
---

## Thinking Path

* **Goal**: Enable the context usage indicator ring to render correctly on the composer footer when using custom provider models whose prefixes overlap with first-party families (e.g., `gemini_cli/gemini-3-flash-preview`).
* **Identify Barrier**: Normalization (`_normalize_provider_id()`) used loose `raw.startswith(prefix)` checks. This incorrectly normalized the custom `"gemini_cli"` namespace to the standard `"google"` provider.
* **Identify Consequence**: The route state resolver (`_resolve_compatible_session_model_state()`) saw a provider mismatch against `"custom"` and silently swapped the session model to the default profile model.
* **Identify UI Collapse**: Swapping the model zeroed out the session context metadata (`last_prompt_tokens` and `threshold_tokens`) on display loads (`GET /api/session`), causing `hasPromptTok` to evaluate to `false` and rendering a fallback dot (`·`) instead of the usage ring.
* **Determine Fix**: Refactor `_normalize_provider_id` to only normalize a prefix if it is an exact string match, or if it is followed by a standard namespace token boundary character (`:`, `/`).
* **Determine Aliases**: Add common first-party hyphenated provider aliases (`google-gemini`, `google-ai-studio`, `claude-code`) explicitly to `_PROVIDER_ALIASES` so they resolve via direct, exact-match lookups.

---

## What Changed

* **`api/routes.py`**:
  * Added `_starts_token(raw, prefix)` helper to check prefix matches up to a delimiter boundary (`:`, `/`).
  * Updated `_normalize_provider_id()` to use `_starts_token()` for prefix matching, naturally ignoring custom suffixes containing underscores or hyphens.
  * Explicitly populated `_PROVIDER_ALIASES` with hyphenated first-party keys (`google-gemini`, `google-ai-studio`, `claude-code`) to support exact-match routing.
* **`tests/test_custom_provider_prefix_collisions.py`**:
  * Added comprehensive unit test coverage for custom prefix collisions (`gemini_cli`, `gpt_proxy`, `claude_gateway`, `openai_compat`).
  * Added routes-level state regression test verifying that session model strings with custom prefixes survive resolution without silent swaps.

---

## Why It Matters

* **User Impact**: Custom provider models using common prefixes (e.g., local CLI, gateway, or proxy setups) now correctly display their context-window usage ring and threshold metrics in the UI instead of falling back to empty indicators.
* **Developer Impact**: Replaces fragile startswith string heuristics with robust token boundaries, preventing custom configurations from colliding with standard provider namespaces.

---

## Contract Routing

* **State Layer Mutated**: `_normalize_provider_id` and `_resolve_compatible_session_model_state` inside `/api/session` state resolution.
* **Invariant Enforced**: Custom provider namespaces containing alphanumeric or underscore suffixes are never collapsed into first-party families, guaranteeing that valid session model strings are preserved untouched on load.
* **Regression Test**: Covered under `test_resolve_session_model_state_custom_prefix_survives` in `tests/test_custom_provider_prefix_collisions.py`.

---

### Automated Unit Tests
Ran the newly added regression test suite inside `/app`:
```bash
$ ./scripts/test.sh tests/test_custom_provider_prefix_collisions.py
============================= test session starts ==============================
collected 2 items
tests/test_custom_provider_prefix_collisions.py ..                       [100%]
============================== 2 passed in 1.79s ===============================
```

### Full Test Suite Integrity
Ran the complete test suite to verify zero regressions across existing routing or fallback layers:
- **Total Ran**: 9,122 test items
- **Passed**: 9,017 test items (11 pre-existing environment failures skipped/unaffected)

---

## Risks / Follow-ups

* **Risks**: Extremely low. First-party exact matches and standard colon-prefixed custom aliases (`custom:`) are explicitly preserved.
* **Follow-ups**: None required.

---

## Model Used

- **AI Model**: `hermes-webui with gemini-3-flash-preview, antigravity-cli`
- **Tooling**: Coordinated with `antigravity` (research model) and executed verification passes locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Add token boundary check for provider normalization (#4278).** `_normalize_provider_id` matched provider prefixes using a broad `startswith()`, which incorrectly collapsed custom model namespace prefixes ending in alphanumeric characters (such as `gemini_cli` or `gpt_proxy`) into first-party families (e.g. `google` or `openai`). This triggered a silent "stale model repair" on session load, swapping the model to the profile default and clearing the turn metrics (`last_prompt_tokens` and `threshold_tokens`) on display. The normalization prefix match is now restricted to exact string matches or prefix matches followed strictly by accepted namespace delimiters (`:`, `/`). Hyphenated first-party provider aliases (`google-gemini`, `google-ai-studio`, `claude-code`) are now mapped explicitly in `_PROVIDER_ALIASES` to resolve via direct exact-match lookups. Resolves the composer context indicator ring disappearing for custom models. (#4278)
+
 ### Internal
 
 - **Windows pytest-harness compatibility (#3664).** Hardened the test suite to run on Windows: profile-home fallback paths are path-normalized, strict POSIX file-mode (`0o600`) assertions are gated behind `os.name != "nt"` (Linux still asserts them at full strictness), the conftest cleanup handles Windows process-tree/port teardown and the Py3.12+ `shutil.rmtree` `onexc` shim, and tests that require `fork`/`fcntl` carry `@requires_fork` / `@requires_fcntl` markers (which never skip on Linux). Test-only — no runtime or app behavior change, no Linux CI behavior change. (#4254, #4255, #4256, #4257, #4259, #4263, #4266, #4274)
--- a/api/routes.py
+++ b/api/routes.py
@@ -1282,6 +1282,9 @@ _PROVIDER_ALIASES = {
     "gpt": "openai",
     "gemini": "google",
     "openai-codex": "openai",
+    "google-gemini": "google",
+    "google-ai-studio": "google",
+    "claude-code": "anthropic",
 }
 
 # OpenAI-compatible /v1/models endpoints for live model discovery.
@@ -2908,6 +2911,13 @@ def _handle_client_event_log(handler, body: dict) -> bool:
     return j(handler, {"ok": True, "event": payload.get("event")}) or True
 
 
+def _starts_token(raw: str, prefix: str) -> bool:
+    if not raw.startswith(prefix):
+        return False
+    rest = raw[len(prefix):]
+    return rest == "" or rest[0] in ":/"
+
+
 def _normalize_provider_id(value: str | None) -> str:
     raw = str(value or "").strip().lower()
     if not raw:
@@ -2924,7 +2934,7 @@ def _normalize_provider_id(value: str | None) -> str:
         ("openrouter", "openrouter"),
         ("custom", "custom"),
     ):
-        if raw.startswith(prefix):
+        if _starts_token(raw, prefix):
             return normalized
     # Unknown prefix — return empty so callers treat it as "no match" and pass
     # the model through unchanged rather than incorrectly stripping it.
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.450** (Release PK) — merged via release PR #4294 (rebased onto current master). Thank you @b3nw! 🎉

The full Codex+Opus+suite gate surfaced two adjacent issues that we fixed on the way in (both with regression tests): (1) the registered `openai-api` provider id needed an explicit alias entry so the new `:/` boundary didn't drop it to empty; (2) the same bare-prefix collision lived in a second loop on the chat-send path (`_apply_profile_provider_context_to_streaming_model` in `api/streaming.py`), now gated the same way. Closing as shipped; this also supersedes the duplicate #4280.

nesquena-hermes:
## Summary

Reading the diff against origin/master (`api/routes.py`, the new `tests/test_custom_provider_prefix_collisions.py`, and the CHANGELOG), this PR fixes the reported `gemini_cli/gemini-3-flash-preview` case correctly — and importantly it patches **both** collision sites, not just the helper. That second change is what actually makes the context ring re-appear for the reported input.

## Code reference

The helper fix via `_starts_token` (routes.py:2914) is good, but on its own it wouldn't fix the symptom, because display resolution enters the profile branch and the swap is driven by a *bare-prefix scan* that runs `startswith` against the full model string. This PR guards exactly that:

```python
# api/routes.py (this PR) — profile branch
model_family = ""
if "/" not in model:                      # <-- the decisive guard
    model_lower = model.lower()
    for bare_prefix in ("gpt", "claude", "gemini"):
        if model_lower.startswith(bare_prefix):
            model_family = _normalize_provider_id(bare_prefix)
            break
```

Without the `if "/" not in model:` wrap, `gemini_cli/gemini-3-flash-preview` still hits `model_lower.startswith("gemini")` → `model_family="google"` → silent swap at the `model_family != _profile_provider_normalized` check, regardless of how the helper treats the namespace. Good catch including it.

## Test routes through the failing path

`test_resolve_session_model_state_custom_prefix_survives` calls the resolver with `model_provider=None` and `profile_provider="custom:newapi"` as a keyword:

```python
_resolve_compatible_session_model_state(
    model_id,
    model_provider,                 # None
    profile_provider=profile_provider,   # "custom:newapi"
    ...
)
```

That correctly skips the fast-path early return (which needs both `model` and `model_provider`) and exercises the profile branch where the bug lives. This is the right way to pin the regression — worth keeping.

## One design note (non-blocking)

`_starts_token` only treats `:` and `/` as boundaries, so hyphen/dot families (`google-gemini`, `claude-code`) are handled via the new `_PROVIDER_ALIASES` entries rather than a generic boundary. That's the more conservative choice: it means a hyphenated *custom* prefix like `gemini-cli` or `gpt-proxy` passes through as `""` instead of collapsing into a first-party family (your test asserts `gemini-cli == ""`, which is correct behavior). The tradeoff is that any future first-party hyphenated slug needs a new alias entry — acceptable given pass-through is the documented safe default for unknown prefixes.

## Residual sibling (follow-up, not required here)

There's a third full-string `startswith` scan in `_model_matches_active_provider_family` (routes.py:2971-2977) that still does:

```python
for bare_prefix in ("gpt", "claude", "gemini"):
    if model_lower.startswith(bare_prefix):
        return _normalize_provider_id(bare_prefix) == active_provider
```

It isn't on the reported display path, so it doesn't block this PR, but it's the same latent collision for `<family>_cli/...` ids in the non-profile resolution path. Might be worth a follow-up to route it through a slash-aware check too.

Overall this looks like the more complete of the two #4278 fixes (cf. #4280, which patches the helper only).

PR #4286, opened 2026-06-16
fix(#4283): stop recovered turns from replaying into later context

### LINKED ISSUE #4283: WebUI replays prior user message in every subsequent turn after interrupted session (gateway restart / tool iteration limit)
## Summary

After a gateway restart or tool iteration limit hit during a WebUI session, a prior user message gets prepended to every subsequent user message's context. The duplication persists indefinitely through all following turns — it is not a one-shot auto-resume. The `state.db` messages table is clean (one copy), so the duplication happens during in-memory conversation history reconstruction.

## Environment

- **Hermes WebUI version:** `v0.51.444-2-g40591f14` (development, past `v0.51.192`)
- **Hermes Agent commit:** `2a14e8957` (June 14, 2026)
- **Model:** GLM-5.2 via ZAI
- **Session:** `e29f556c900b`, source `webui`

## Reproduction

1. Start a WebUI chat session
2. Send several messages with heavy tool usage (enough to hit tool iteration limit or trigger a gateway restart)
3. Restart the gateway (or let the tool iteration limit interrupt the turn)
4. Send a new user message
5. Every subsequent user message submitted through the WebUI has the **original interrupted-turn user message** prepended to its context

## Observed Behavior

The LCM database shows the same user message text appearing in multiple LCM rows, all with the **same timestamp** (`1781568405.9936945`):

```
store_id=76697: "alright gateway restarted, lets give it a live test..."
store_id=76703: "alright gateway restarted, lets give it a live test...\n\ncan you configure the same for raiden's profile as well"
store_id=76719: "alright gateway restarted, lets give it a live test...\n\nalright, i think we had a couple of crons..."
store_id=76744: "alright gateway restarted, lets give it a live test...\n\ngreat, this should be a lot easier for us"
store_id=76746: "alright gateway restarted, lets give it a live test...\n\nsomething is making you read prior messages..."
store_id=76754: "alright gateway restarted, lets give it a live test...\n\nso does it only happen one or 2 times..."
store_id=76755: "alright gateway restarted, lets give it a live test...\n\nlets do that, find the fix and we can submit the PR..."
```

The old message is prepended to every new user message's context, not sent as a separate turn.

## Database State (Clean)

The `state.db` messages table is **not** corrupted — the original message exists exactly once:

```sql
SELECT id, role, content FROM messages WHERE session_id = 'e29f556c900b' AND content LIKE '%alright gateway restarted%';
-- id=123998 role=user: "alright gateway restarted, lets give it a live test..."
-- id=124053 role=assistant: (quotes it in response)
```

Only 2 rows — one user, one assistant quoting it. **The duplication happens during conversation history reconstruction**, not during persistence.

## Root Cause Analysis

The WebUI uses the Hermes Agent Python API directly — `agent.run_conversation()` with a `conversation_history` parameter (in `api/streaming.py:~6713`). It does **not** use a PTY/TUI bridge.

The bug appears to be in the **`pending_user_message` recovery and context reconstruction** flow:

1. When a turn starts, the user's message is saved as `session.pending_user_message` (persisted in session JSON via `Session.save()`)
2. A streaming checkpoint (`_save_streaming_checkpoint` in `api/streaming.py:~3890`) saves the session mid-stream
3. On interruption (gateway restart, tool iteration limit), the normal cleanup path (`session.pending_user_message = None`) may not execute
4. On the next turn, `_repair_pending_state()` in `api/models.py:~1850` recovers the stale `pending_user_message` — but the recovery appends it to `session.context_messages` via `_append_recovered_turn_to_context()`, which then leaks into every subsequent turn's context
5. The `_drop_checkpointed_current_user_from_context()` function in `api/streaming.py:3879` only checks if the **last** message matches the current user text — it does not detect the stale recovered message buried earlier in the history

### Key code paths involved:

- `api/streaming.py:6651` — `_new_turn_context_from_messages()` builds the context for each new turn
- `api/streaming.py:6662` — `_deduplicate_context_messages()` runs but apparently doesn't catch this case
- `api/models.py:1850` — `_repair_pending_state()` handles stale `pending_user_message` recovery
- `api/models.py:1911` — `_append_recovered_turn_to_context()` appends recovered messages to context
- `api/streaming.py:3879` — `_drop_checkpointed_current_user_from_context()` only drops the last message if it matches

## Suspected Fix

Two potential approaches:

**Option A — Clear stale context after recovery:**
After `_repair_pending_state()` materializes a stale `pending_user_message`, the recovered turn should NOT persist in `session.context_messages` for future turns. It should be added to `session.messages` (for display) but excluded from the context used to build `conversation_history` for subsequent turns.

**Option B — Broaden deduplication:**
`_drop_checkpointed_current_user_from_context()` could scan the entire history (not just the last message) for messages matching the current `pending_user_message` text and remove duplicates. The `_deduplicate_context_messages()` function could also be enhanced to detect this pattern.

## Workaround

Start a new session (`/new`) after a gateway restart or tool iteration limit hit. This clears the in-memory history that contains the stuck interrupted turn.

## Impact

**High** — makes the WebUI unusable after any interrupted turn. Every new message gets the old message prepended, causing the agent to re-read and re-respond to stale context indefinitely. This wastes tokens and produces confusing, context-contaminated responses.

rodboev:
## Summary
A recovered interrupted turn still confuses `state_db_delta_after_context()`, so the next turn can replay the old prompt as if it were fresh context. This target narrows the delta alignment for recovered turns and keeps the recovered prompt visible once without re-prepending it on every later turn.

Closes #4283.

## What Changed
- `api/models.py`: adjust the state-db/context reconciliation so a recovered turn can anchor a one-row prefix instead of falling back to replaying the whole state transcript.
- `tests/test_webui_state_db_context_reconciliation.py`: add a regression that reproduces the recovered-turn shape and proves the later delta no longer includes the old prompt again.
- `CHANGELOG.md`: add an Unreleased fixed-note for the user-visible recovery behavior change.

## Why It Matters
After a gateway restart or tool-iteration abort, the resumed conversation should not keep carrying the interrupted prompt into every future turn. That replay contaminates the model context, wastes tokens, and makes the resumed session behave as if the earlier prompt was re-sent by the user.

## Verification
- Focused pytest on the new context-reconciliation regression.
- Focused pytest on the recovered-turn regression and the unrelated-prefix guard.
- Existing related coverage remains in `tests/test_session_sidecar_repair.py` and `tests/test_webui_state_db_context_reconciliation.py`.

## Risks / Follow-ups
- The exception needs to stay narrow so ordinary sessions still require the existing multi-row prefix match.
- If the regression shows a second recovery shape, fold it into the same helper rather than adding another ad hoc filter.

## Model Used
GPT-5 via Codex CLI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Recovered interrupted turns no longer replay the interrupted prompt into later context rebuilds.** After a gateway restart or tool-iteration abort, the recovered user turn still stays visible once, but state-db/context reconciliation now treats the recovered turn as already represented only when it is the actual aligned prefix row. That keeps later turns from inheriting the old interrupted prompt repeatedly. (#4283)
+
 ### Internal
 
 - **Windows pytest-harness compatibility (#3664).** Hardened the test suite to run on Windows: profile-home fallback paths are path-normalized, strict POSIX file-mode (`0o600`) assertions are gated behind `os.name != "nt"` (Linux still asserts them at full strictness), the conftest cleanup handles Windows process-tree/port teardown and the Py3.12+ `shutil.rmtree` `onexc` shim, and tests that require `fork`/`fcntl` carry `@requires_fork` / `@requires_fcntl` markers (which never skip on Linux). Test-only — no runtime or app behavior change, no Linux CI behavior change. (#4254, #4255, #4256, #4257, #4259, #4263, #4266, #4274)
--- a/api/models.py
+++ b/api/models.py
@@ -4448,18 +4448,18 @@ def state_db_delta_after_context(sidecar_context: list, state_messages: list) ->
 
     # Recovered interrupted turns are special: the visible interruption marker
     # is synthetic, so the recovered user turn should still count as a mirrored
-    # prefix even when it is the only durable row that matches the state DB.
-    allow_single_row_prefix = any(
-        isinstance(msg, dict)
-        and msg.get('_recovered')
-        and str(msg.get('role') or '') == 'user'
-        for msg in sidecar_context
+    # prefix when it is the actual aligned prefix row.
+    allow_single_row_prefix = bool(
+        isinstance(sidecar_context[0], dict)
+        and sidecar_context[0].get('_recovered')
+        and str(sidecar_context[0].get('role') or '') == 'user'
     )
 
     sidecar_keys = [_session_message_content_key(m) for m in sidecar_context]
     state_keys = [_session_message_content_key(m) for m in state_messages]
     max_offset = min(len(sidecar_keys), len(state_keys))
     best_len = 0
+    best_offset = 0
     for offset in range(max_offset):
         length = 0
         while (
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Shipped in v0.51.449 (Release PJ) via #4291. 🎉 Recovered interrupted turns are now matched + de-duplicated instead of replayed into every later turn. Codex SAFE + Opus net-positive + clean suite. Thanks @rodboev!

PR #4297, opened 2026-06-16
fix: make gateway watcher startup non-blocking with timeout

minidarkmimi:
## Problem

The gateway watcher's `start_watcher()` call in `server.py` runs synchronously in the main thread during WebUI startup. If `_resolve_watcher_target()` hangs (e.g., `state.db` locked, profile resolution slow, or any I/O block), the entire WebUI never starts — the server never binds its port and no requests are accepted.

## Fix

Wrap `start_watcher()` in a daemon thread with a 5-second join timeout:

```python
import threading
t = threading.Thread(target=start_watcher, daemon=True)
t.start()
t.join(timeout=5)
if t.is_alive():
    print('[tip] Gateway watcher still initializing (non-blocking)', flush=True)
```

- The watcher initializes in the background without blocking the server
- If it takes >5s, a tip is printed but startup continues
- Daemon thread is cleaned up on shutdown
- No functional change — the watcher still starts and works normally

## Testing

Tested on macOS with Hermes Agent running. WebUI starts immediately even when the watcher's profile resolution is slow. The watcher initializes in the background and begins polling `state.db` normally.

### DIFF
--- a/server.py
+++ b/server.py
@@ -592,8 +592,13 @@ def main() -> None:
 
     # Start the gateway session watcher for real-time SSE updates
     try:
+        import threading, signal
         from api.gateway_watcher import start_watcher
-        start_watcher()
+        t = threading.Thread(target=start_watcher, daemon=True)
+        t.start()
+        t.join(timeout=5)
+        if t.is_alive():
+            print('[tip] Gateway watcher still initializing (non-blocking)', flush=True)
     except Exception as e:
         print(f'[!!] WARNING: Gateway watcher failed to start: {e}', flush=True)
 
--- a/server.py
+++ b/server.py
@@ -592,9 +592,15 @@ def main() -> None:
 
     # Start the gateway session watcher for real-time SSE updates
     try:
-        import threading, signal
         from api.gateway_watcher import start_watcher
-        t = threading.Thread(target=start_watcher, daemon=True)
+
+        def _start_watcher_safe():
+            try:
+                start_watcher()
+            except Exception as e:
+                print(f'[!!] WARNING: Gateway watcher failed to start: {e}', flush=True)
+
+        t = threading.Thread(target=_start_watcher_safe, daemon=True)
         t.start()
         t.join(timeout=5)
         if t.is_alive():


nesquena-hermes (changes requested):
Thanks for this, @minidarkmimi — making startup resilient to a watcher hang is a legitimate goal. There's one blocker that fails CI, plus a behavior note worth addressing before this can land.

## Blocker — fails the ruff forward-lint gate (CI red)

The new `import threading, signal` line introduces an unused import that the repo's line-scoped ruff gate (`scripts/ruff_lint.py --diff`, enforced in CI and the pre-release pipeline via the curated `E9 + F + B` ruleset) rejects:

```
server.py:595:27  F401  `signal` imported but unused
```

`signal` is never referenced in the new block. On top of that, **both `signal` and `threading` are already imported at module top** (`server.py:9` `import signal`, `server.py:12` `import threading`), so the entire local `import threading, signal` line is redundant — you can just use the module-level names.

**Fix:** drop the local import line entirely:

```python
    # Start the gateway session watcher for real-time SSE updates
    try:
        from api.gateway_watcher import start_watcher
        t = threading.Thread(target=start_watcher, daemon=True)
        t.start()
        t.join(timeout=5)
        if t.is_alive():
            print('[tip] Gateway watcher still initializing (non-blocking)', flush=True)
    except Exception as e:
        print(f'[!!] WARNING: Gateway watcher failed to start: {e}', flush=True)
```

You can confirm green locally with `python3 scripts/ruff_lint.py --diff origin/master` (exit 0 = clean).

## Behavior note — the `try/except` no longer catches a watcher failure

Worth knowing so the error path stays honest: `GatewayWatcher.start()` already runs its poll loop in its own daemon thread (`api/gateway_watcher.py:230-236`), so the *polling* never blocked startup in the first place. The only synchronous work in `start_watcher()` is `_resolve_watcher_target()` (profile/home resolution) — that's the narrow thing your wrapper actually protects against, which is fine.

But by moving `start_watcher()` onto a daemon thread, an exception raised inside it **no longer reaches your `except Exception as e:` handler** — it surfaces on the daemon thread's own stack and dumps a raw `threading` traceback to stderr instead of the clean `[!!] WARNING: Gateway watcher failed to start: …` message. I verified this:

```
t.is_alive after join: False
Exception reached caller except handler? False
```

So the `except` is now partially dead. If you want to preserve the friendly warning, catch inside the thread target, e.g.:

```python
    def _start_watcher_safe():
        try:
            start_watcher()
        except Exception as e:
            print(f'[!!] WARNING: Gateway watcher failed to start: {e}', flush=True)

    t = threading.Thread(target=_start_watcher_safe, daemon=True)
    t.start()
    t.join(timeout=5)
    if t.is_alive():
        print('[tip] Gateway watcher still initializing (non-blocking)', flush=True)
```

The CI blocker is the only must-fix; the exception-handling note is a should-fix to keep the warning path working. Once the ruff gate is green I'll take another look.

nesquena-hermes:
Re-read the head `a156f4c3` against `origin/master`. Both items from my last pass are addressed cleanly — this looks good to land.

## CI blocker resolved

The redundant `import threading, signal` line is gone; the new block uses the module-level imports (`server.py:9` `import signal`, `server.py:12` `import threading`). The `F401 signal imported but unused` that the line-scoped ruff gate (`scripts/ruff_lint.py --diff`) was rejecting no longer appears in the diff:

```python
    try:
        from api.gateway_watcher import start_watcher

        def _start_watcher_safe():
            try:
                start_watcher()
            except Exception as e:
                print(f'[!!] WARNING: Gateway watcher failed to start: {e}', flush=True)

        t = threading.Thread(target=_start_watcher_safe, daemon=True)
        t.start()
        t.join(timeout=5)
        if t.is_alive():
            print('[tip] Gateway watcher still initializing (non-blocking)', flush=True)
    except Exception as e:
        print(f'[!!] WARNING: Gateway watcher failed to start: {e}', flush=True)
```

(`server.py:594-608`)

## Exception path preserved

The `_start_watcher_safe()` wrapper catches inside the daemon target, so a failure in `start_watcher()` now surfaces as the friendly `[!!] WARNING: Gateway watcher failed to start: …` line instead of a raw `threading` traceback on the daemon's stack. That was the "partially dead `except`" concern from last round — fixed. The outer `try/except` still covers the import and thread-spawn itself, which is correct.

## Premise still holds

Re-confirmed the threading model that makes this safe: `GatewayWatcher.start()` already runs the poll loop on its own daemon thread (`api/gateway_watcher.py:235`):

```python
self._thread = threading.Thread(target=self._poll_loop, daemon=True, name='gateway-watcher')
```

So the only synchronous work your wrapper guards is `_resolve_watcher_target()` (`api/gateway_watcher.py:363`, called at `:417`) — profile/home resolution and the initial `state.db` touch. That is exactly the narrow blocking surface the 5s join protects against, so the fix targets the right thing.

## Verification

Per policy I don't run code from PR branches, but a maintainer can confirm the gate locally with `python3 scripts/ruff_lint.py --diff origin/master` (exit 0 = clean) and sanity-check startup by pointing at a deliberately-slow/locked `state.db` — the server should bind its port immediately and print the `[tip] … still initializing` line rather than hanging. No remaining blockers from my side.

nesquena-hermes:
Shipped in **v0.51.457** (Release PR — "gateway watcher can't brick startup"). 🚀

We picked this up and shipped it through the full release gate. The fix landed exactly as you wrote it — `start_watcher()` now runs on a `daemon=True` thread with a 5-second join, so a slow/blocked watcher init (locked `state.db`, slow profile resolution, I/O stall) can no longer prevent the server from binding its port.

**Gate results (on a fresh rebase onto current master):**
- Full pytest suite: 9186 passed, 9 skipped
- Codex regression gate: SAFE TO SHIP
- Opus advisor: ship it (confirmed the watcher's poll loop already runs on its own daemon thread, so SSE liveness is unchanged; the genuinely-synchronous surface being guarded is the profile resolution + lock acquisition)

Your authorship is preserved via a `Co-authored-by` trailer on the fix commit, and you're credited by handle in the CHANGELOG. Thanks @minidarkmimi! Closing as merged.

PR #4303, opened 2026-06-16
Update stopping instructions for server launch methods

tomas-forgac:
Currently it says  `./start.sh` starts the server in the foreground which is incorrect, so to stop the server, the method for detached start applies

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ For self-hosted VM or homelab installs, `ctl.sh` wraps the common daemon lifecyc
 >
 > | Launch method | How to stop |
 > |---|---|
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Merged to master via #4304. Thank you @tomas-forgac! 🙏

You're right — `start.sh` exec's `bootstrap.py` without `--foreground` (start.sh:65), which double-forks into a detached daemon, so the old README row was wrong. Your fix correctly moves `./start.sh` into the 'find the PID + kill' row. Docs-only, so no version stamp (per our docs-only convention). CI green across all shards. Closing as merged.

PR #4308, opened 2026-06-16
start.sh: detect already-running server and inform the user accordingly

tomas-forgac:
When the server is running and user runs start.sh, currently the script pretends it starts the server

Instead, this change makes the script inform the user that the server is already running, that it hasn't started a new instance. It also tells the user how to restart the server if that's what their intention was.


nesquena-hermes:
Shipped in **v0.51.456** (Release PQ) — merged via release PR #4309. Thanks again @tomas-forgac! 🙏

Well-reasoned change — referencing `server.py`'s `_abort_if_already_serving` and matching its loopback-probe behavior made the review easy. The gate (Codex + Opus + full suite) added one parity fix on the way in: the preflight resolved host/port from env/defaults only, but `bootstrap.py` also accepts a positional `port` + `--host` via `"$@"`, so `HERMES_WEBUI_PORT=18987 ./start.sh 18988` would have falsely reported 'already running' and blocked the legit second instance. Added `"$@"` parsing (--host / --host=VALUE / first numeric positional) matching bootstrap.py's argparse, verified across all cases. Full suite 9186 green; container path (server.py direct) confirmed unaffected. Closing as shipped.

PR #4312, opened 2026-06-16
Recognize command STT providers in voice capability probe

LLMinem:
## Thinking Path

- Hermes WebUI aims to reuse the user's existing Hermes Agent setup rather than maintain separate provider logic.
- Hermes Agent now supports explicit named STT providers through `stt.providers.<name>: type: command`.
- WebUI's passive `/api/transcribe/capability` probe still only recognized built-in STT providers, so explicit command-backed providers were reported unavailable even though `transcribe_audio()` could handle them.
- The fix is to delegate that provider recognition to Hermes Agent's resolver when it is available, without invoking transcription or lazy installs during the probe.

## What Changed

- Added command-provider recognition to `_stt_provider_capability_from_module()` via Hermes Agent's `_resolve_command_stt_provider_config()` helper.
- Added a regression test for an explicit named command provider such as `stt.provider: elevenlabs-gemini`.
- Added an Unreleased changelog note.

## Why It Matters

Browser voice recordings can now use custom command-backed Hermes STT providers instead of being incorrectly treated as unavailable and falling back to browser speech recognition or raw audio upload flows.

## Verification

- `./scripts/test.sh tests/test_voice_transcribe_endpoint.py -q` → `8 passed`
- `git diff --check` → clean
- `ruff check api/upload.py tests/test_voice_transcribe_endpoint.py` was attempted; it reports pre-existing unused imports/variables in `api/upload.py` unrelated to this change, so I left them untouched to keep the PR focused.

## Risks / Follow-ups

- This is a passive capability probe only. It does not execute command providers and does not change the actual transcription path.
- If a future Hermes Agent release renames `_resolve_command_stt_provider_config()`, WebUI will fall back to the previous behavior for named command providers until it is updated.

## Contract Routing

Task type: backend capability-probe bug fix
Touched areas:
- `api/upload.py`
- `tests/test_voice_transcribe_endpoint.py`
- `CHANGELOG.md`
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
Scope boundaries:
- No UI flow, streaming, session-state, or security contract changes.
- Focused endpoint regression test passes.

## Model Used

OpenAI GPT-5.5 via Hermes Agent, with file/terminal tool use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Voice transcription capability checks now recognize Hermes named command STT providers.** WebUI's `/api/transcribe/capability` now treats explicit `stt.provider` values that resolve through Hermes Agent's `stt.providers.<name>: type: command` registry as available, so browser voice recordings can use custom command-backed providers instead of falling back to browser speech recognition or raw audio uploads.
+
 ### Internal
 
 - **Windows pytest-harness compatibility (#3664).** Hardened the test suite to run on Windows: profile-home fallback paths are path-normalized, strict POSIX file-mode (`0o600`) assertions are gated behind `os.name != "nt"` (Linux still asserts them at full strictness), the conftest cleanup handles Windows process-tree/port teardown and the Py3.12+ `shutil.rmtree` `onexc` shim, and tests that require `fork`/`fcntl` carry `@requires_fork` / `@requires_fcntl` markers (which never skip on Linux). Test-only — no runtime or app behavior change, no Linux CI behavior change. (#4254, #4255, #4256, #4257, #4259, #4263, #4266, #4274)
--- a/api/upload.py
+++ b/api/upload.py
@@ -457,6 +457,13 @@ def _stt_provider_capability_from_module(stt):
             # non-WAV input through ffmpeg before invoking the command.
             return has_local_command() and has_browser_audio_converter()
 
+        def command_provider_available(provider):
+            resolver = getattr(stt, "_resolve_command_stt_provider_config", None)
+            try:
+                return callable(resolver) and resolver(provider, cfg_dict) is not None
+            except Exception:
+                return False
+
         def resolve_provider(provider):
             if provider == "local":
                 if bool(getattr(stt, "_HAS_FASTER_WHISPER", False)):
@@ -485,6 +492,8 @@ def _stt_provider_capability_from_module(stt):
                     return "none"
             if provider == "elevenlabs":
                 return "elevenlabs" if bool(env("ELEVENLABS_API_KEY")) else "none"
+            if command_provider_available(provider):
+                return provider
             return "none"
 
         explicit = "provider" in cfg_dict
--- a/tests/test_voice_transcribe_endpoint.py
+++ b/tests/test_voice_transcribe_endpoint.py
@@ -156,3 +156,31 @@ def test_handle_transcribe_capability_reports_actual_local_command_fallback(monk
 
     assert available is True
     assert provider == "local_command"
+
+
+def test_handle_transcribe_capability_reports_named_command_provider(monkeypatch):
+    stt_config = {
+        "provider": "elevenlabs-gemini",
+        "providers": {
+            "elevenlabs-gemini": {
+                "type": "command",
+                "command": "/home/user/.hermes/scripts/hermes-elevenlabs-stt.sh {input}",
+            }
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
## Summary

Reviewed the diff against `origin/master` and cross-checked the agent contract this depends on. The change is small and well-scoped: `command_provider_available()` (api/upload.py:486-492) delegates command-provider recognition to the agent's resolver, and `resolve_provider()` consults it last (api/upload.py:495-496). This is the right shape for a passive probe — no transcription, no lazy install, just config inspection.

## Cross-repo contract — verified, matches

The helper you call exists and the signature lines up exactly. In `tools/transcription_tools.py:322` (agent side):

```python
def _resolve_command_stt_provider_config(
    provider: str,
    stt_config: Dict[str, Any],
) -> Optional[Dict[str, Any]]:
    ...
    if key in BUILTIN_STT_PROVIDERS or key == "none":
        return None
```

So `resolver(provider, cfg_dict)` in api/upload.py:489 passes args in the correct order, and the agent already rejects built-in names and `"none"` internally — returning `None` for anything that isn't a declared command provider.

## Ordering is correct (and intentionally mirrors the agent)

Placing the `command_provider_available(provider)` check *after* the built-in `elif` chain in `resolve_provider()` is exactly right. It mirrors the agent's own dispatch precedence at `transcription_tools.py:1702`, where the command resolver fires only after the built-in handlers short-circuit:

```python
# Fires after the built-in elif chain — built-in names short-circuit
# upstream so a user's stt.providers.openai.command can't override the
# real OpenAI handler — and BEFORE the plugin dispatcher ...
command_provider_config = _resolve_command_stt_provider_config(provider, stt_config)
```

Because the agent rejects built-in names regardless, even if a future refactor moved the WebUI check earlier it couldn't shadow a native provider — but keeping it last is the cleaner expression of the contract.

## One consistency note (not a blocker)

The non-explicit auto-detect loop (api/upload.py, the `for candidate in ("local", "local_command", "groq", "openai", "mistral", "xai", "elevenlabs")` tuple) deliberately omits command providers. That stays consistent with the agent: `_get_provider()` at `transcription_tools.py:745` only auto-detects `local > groq > openai` and never auto-selects a command provider, so a command-backed STT only activates via an explicit `stt.provider: <name>`. Your fix handles exactly that explicit path. Worth a one-line comment in `_stt_provider_capability_from_module` stating the auto-detect omission is intentional (matches agent `_get_provider`), so a later contributor doesn't "fix" it by adding command providers to the candidate tuple and diverge from the agent.

## Defensive pattern is consistent

`command_provider_available` wrapping the resolver in `getattr(...) + try/except → False` matches the existing helpers (`has_local_command` → `_has_local_command`, `has_browser_audio_converter` → `_find_ffmpeg_binary`). That gives the graceful-degradation-on-rename behavior you called out in the PR's Risks section, and it's the established style in this function, so no change needed there.

## Test

`tests/test_voice_transcribe_endpoint.py:161` (`test_handle_transcribe_capability_reports_named_command_provider`) asserts the explicit `elevenlabs-gemini` command provider now resolves to `(True, "elevenlabs-gemini")`, which directly covers the new branch. The fake module stubs `_resolve_command_stt_provider_config` with the same `(provider, cfg)` order, so the test exercises the real call path. Looks good to land once a maintainer confirms the agent release that ships `_resolve_command_stt_provider_config` is the minimum supported version (older agents simply fall back via the `callable(resolver)` guard).

nesquena-hermes:
Shipped in **v0.51.460** (Release PU — "custom voice providers recognized"). 🚀

The fix landed as you wrote it: `command_provider_available()` delegates to the agent's `_resolve_command_stt_provider_config` as the single source of truth, placed as the final fallthrough after the built-in chain. The gate confirmed the agent rejects built-in names internally so ordering can't shadow a native provider, and older agents fall back gracefully via the `callable(resolver)` guard.

We also added a clarifying comment at the auto-detect tuple (the `for candidate in ("local", "groq", ...)` loop) documenting that command providers are intentionally omitted to match the agent's `_get_provider()` precedence — so a future contributor doesn't accidentally add command providers to the auto-detect path and diverge from the agent's resolution order.

**Gate:** full suite 9209 passed; Codex SAFE TO SHIP; Opus ship it (additive-only, single new branch can only flip `(False,"none")→(True,<provider>)`, no built-in path changes). Attribution preserved via Co-authored-by. Thanks @LLMinem! Closing as merged.

PR #4327, opened 2026-06-16
fix(#4326): reject embedded null byte in remote workspace candidate

### LINKED ISSUE #4326: hardening: reject embedded null byte in remote workspace candidate (#4273 follow-up)
## Summary

Low-severity hardening follow-up from the #4273 / #3664 workspace trust-boundary review (shipped v0.51.463).

`_remote_terminal_workspace_candidate` (`api/workspace.py`, ~line 154-176) normalizes the candidate and `terminal.cwd` and runs the containment check, but it does **not** reject a path containing an embedded null byte the way the rest of the #4273 hardening now does (`_as_posix_path` returns `None` on `\x00`, `_safe_resolve` catches `ValueError`, `_workspace_access_error` returns a clean validation error, and the `/api/workspaces/add` route preflight catches resolve failures).

Both Opus gate passes on #4273 flagged this as the one residual: a genuine fail-open on the remote boundary that contradicts the code's own fail-closed invariant. It was consciously deferred at ship time as **low severity** — remote profiles only, and local dereference is already blocked by the poison-pill path, so there's no path-traversal escape — but it's worth closing for consistency on a security boundary.

## Suggested fix

Reject embedded null bytes (and treat unparseable input as fail-closed) at the top of `_remote_terminal_workspace_candidate`, after the raw/cwd are read (≈ after line 164), e.g.:

```python
if '\x00' in str(raw) or '\x00' in str(cwd):
    return None
```

…or route the remote candidate through the same `_normalize_posix_path` / null-byte guard the local path now uses, so the remote boundary matches the local one.

## Tests

Extend `tests/test_remote_terminal_workspace.py` with a null-byte remote-candidate case asserting it's rejected (returns `None`), mirroring the local null-byte regressions added in `tests/test_4273_expanduser_and_nullbyte.py`.

## Context

- Hardening lineage: #3664 → #3879 (separate, still open — destructive git paths) / #4273 (shipped v0.51.463 — remote POSIX workspace boundaries).
- Severity: low (no escape; remote-profile path only). Good candidate to batch into the next workspace-security pass rather than a standalone release.

rodboev:
## Thinking Path

- The #4273 workspace trust-boundary hardening shipped in v0.51.463, but both Opus gate passes flagged `_remote_terminal_workspace_candidate` as the one residual: it reads `raw` and `cwd` without rejecting embedded null bytes.
- The rest of the file already fails closed on `\x00` through `_as_posix_path` (line 111) and `_safe_resolve` (line 416), but the remote function's `else` fallback at line 176 bypasses both gates.
- Adding an early null-byte rejection at the top of the function aligns the remote boundary with the local one, matching the `_as_posix_path` pattern.

## What Changed

- `api/workspace.py`: add `\x00` rejection guard at the top of `_remote_terminal_workspace_candidate`, returning `None` before any path parsing when either `raw` or `cwd` contains a null byte
- `tests/test_remote_terminal_workspace.py`: add null-byte remote-candidate regression cases mirroring `tests/test_4273_expanduser_and_nullbyte.py`

## Why It Matters

Closes a residual fail-open on the remote workspace trust boundary so the entire #4273 hardening surface consistently rejects embedded null bytes. Low severity (remote profiles only, no path-traversal escape), but the inconsistency on a security boundary is worth closing for defense-in-depth.

## Verification

```
pytest tests/test_remote_terminal_workspace.py -v --timeout=60
```

## Risks / Follow-ups

- None expected. The change is additive (new guard + new tests) and matches an established pattern in the same file.

## Upstream

Closes #4326.

## Model Used

### DIFF
--- a/api/workspace.py
+++ b/api/workspace.py
@@ -163,6 +163,8 @@ def _remote_terminal_workspace_candidate(path: str | Path) -> Path | None:
     raw = _strip_surrounding_quotes(str(path)).strip()
     if not raw:
         return None
+    if '\x00' in raw or '\x00' in cwd:
+        return None
     normalized_raw = _normalize_posix_path(raw)
     normalized_cwd = _normalize_posix_path(cwd)
     if normalized_raw is not None and normalized_cwd is not None:
--- a/tests/test_remote_terminal_workspace.py
+++ b/tests/test_remote_terminal_workspace.py
@@ -90,3 +90,23 @@ def test_var_home_workspaces_stay_allowed_before_system_root_blocklist(monkeypat
     monkeypatch.setattr(workspace, "_workspace_access_error", lambda _candidate: None)
 
     assert validator(str(candidate)) == candidate
+
+
+def test_remote_terminal_workspace_rejects_embedded_nullbyte_in_raw_path(monkeypatch):
+    """Embedded null bytes in remote workspace path should be rejected."""
+    monkeypatch.setattr(api_config, "get_config", lambda: _remote_config())
+
+    # Path with embedded null byte
+    nullbyte_path = f"{REMOTE_CWD}/projects\x00/demo"
+
+    assert workspace._remote_terminal_workspace_candidate(nullbyte_path) is None
+
+
+def test_remote_terminal_workspace_rejects_embedded_nullbyte_in_cwd(monkeypatch):
+    """Embedded null bytes in remote terminal cwd should be rejected."""
+    monkeypatch.setattr(api_config, "get_config", lambda: _remote_config(terminal={"backend": "ssh", "cwd": f"{REMOTE_CWD}\x00/malicious"}))
+
+    # Normal path, but remote cwd contains null byte
+    normal_path = f"{REMOTE_CWD}/projects/demo"
+
+    assert workspace._remote_terminal_workspace_candidate(normal_path) is None


nesquena-hermes:
Reading `api/workspace.py:158-182` on `origin/master` alongside this diff, the guard is in the right place and closes the exact fail-open #4326 describes.

## Code reference

The pre-existing fast path already rejects null bytes, but only on the *normalized* branch:

```python
# api/workspace.py:104-115 (_as_posix_path)
raw = _strip_surrounding_quotes(str(path)).strip().replace('\\', '/')
if '\x00' in raw:
    return None
```

When `_as_posix_path` returns `None`, `_normalize_posix_path` returns `None`, so the `if normalized_raw is not None and normalized_cwd is not None:` block at `api/workspace.py:168` is skipped and execution falls through to the raw branch:

```python
# api/workspace.py:176-181 (the fall-through this PR closes)
candidate = _resolve_path(raw)
base = _resolve_path(cwd)
if _is_blocked_workspace_path(candidate, raw) or ...:
    return None
if candidate == base or _is_within(candidate, base):
    return candidate
```

`_resolve_path` → `_safe_resolve` deliberately swallows the `ValueError` that `.resolve()` raises on an embedded null byte and returns the *raw* poison-pill path (`api/workspace.py:416-420`), so the containment check runs against an unsanitised path. The new guard at `api/workspace.py:165-166` short-circuits before that fall-through, matching the fail-closed posture `_as_posix_path` already takes.

## Diagnosis

Placement is correct: both `cwd` (line 160) and `raw` (line 163) are confirmed non-empty strings before the check, so it can't short-circuit a legitimate path. The two new tests in `tests/test_remote_terminal_workspace.py` cover both surfaces (null in submitted path, null in `terminal.cwd`) and call `_remote_terminal_workspace_candidate` directly, consistent with the sibling local-path regressions in `tests/test_4273_expanduser_and_nullbyte.py`. This looks complete for the issue as scoped.

## Optional follow-up (out of #4326's scope)

`_profile_default_workspace` at `api/workspace.py:216-221` also reads `terminal.cwd` and, for remote profiles, returns `str(cwd).strip()` without the null-byte gate. It doesn't `.resolve()` on the remote branch so there's no 500 risk, but if you want full parity on the boundary it'd be the next spot to route through `_normalize_posix_path`. Fine to leave for a later workspace-security pass — #4326 only named `_remote_terminal_workspace_candidate`.

nesquena-hermes:
Shipped in **v0.51.465** (Release PZ) via #4334. Thanks @rodboev — your fix rebased cleanly onto v0.51.464 and your commit is preserved in the release (authorship intact). This completes the #4273/#3664 remote-workspace trust-boundary hardening by closing the null-byte fail-open you traced. Closing as integrated (#4334 = your commit rebased + a CHANGELOG stamp). 🚀

Issue #4295, opened 2026-06-16
bug(chat): viewport jumps upward while reading an in-progress streamed reply


santastabber:
## Summary

While an assistant reply is still streaming, the chat viewport can jump upward while the user is reading. In some cases it jumps above the most recent user message, so the reader loses both the live reply and the prompt that started it.

This is distinct from completion-time jump-to-bottom bugs. It happens during the in-progress reply.

## Reproduction

1. Start a long-running assistant response with enough streamed text/tool/progress activity to create a tall live turn.
2. While the response is still streaming, scroll upward slightly to read the in-progress reply or the prompt immediately above it.
3. Continue reading while additional stream/progress/tool events arrive.
4. Observe the viewport position during the stream.

## Actual behavior

- The viewport can jump upward unexpectedly while the reply is still in progress.
- The jump can land above the last user message that triggered the reply.
- The user loses their reading position and must manually scroll back down to find the active response again.
- This can happen even when automatic stream-follow settings are disabled.

## Expected behavior

If the user has manually scrolled away from the bottom during an active stream, the viewport should remain anchored to the semantic content they were reading. Incoming stream/progress/tool updates should not move the reader upward or downward unless the user explicitly chooses to jump/follow.

## Related but distinct issues

Possibly related to #4207, but the symptom differs:

- #4207: scroll jumps at or after stream completion.
- this issue: scroll jumps upward during the active/in-progress stream.

It is also broadly related to #3400 because unstable live-to-final rendering makes running agent sessions hard to read.

## Suspected cause / code pointers

The current streaming/render path has several places that can rebuild or mutate the live transcript while preserving/restoring scroll using absolute pixel offsets:

- `static/messages.js`: live token/progress rendering calls `scrollIfPinned()` after stream updates.
- `static/messages.js`: multiple active-stream/reconnect/error/recovery paths call `renderMessages({ preserveScroll: true })`.
- `static/ui.js`: `renderMessages(...)` captures `scrollTop` / bottom distance via `_captureMessageScrollSnapshot()`.
- `static/ui.js`: `_restoreMessageScrollSnapshot(...)` restores absolute `scrollTop` after DOM rebuild.
- `static/ui.js`: `_captureMessageViewportAnchor()` / `_restoreMessageViewportAnchor(...)` exist, but the normal preserve-scroll render path appears to use the pixel snapshot rather than a semantic message anchor.
[...]

PR #4110, opened 2026-06-13
fix: proportional scroll restoration on completion for expanded render window

PINKIIILQWQ:
## Bug Description

When a streaming reply finishes in a long conversation (〜200+ messages), the viewport jumps to an early part of the chat instead of staying at the messages the user was reading. The bug was ~40% reproducible: it only manifests when the user has scrolled up during streaming AND the conversation is long enough to trigger a render-window expansion.

Fixes a regression that remained after #1709 (v0.51.5) for conversations long enough to trigger the window expansion path.

## Root Cause

At stream completion, `renderMessages({preserveScroll:true})` expands the message render window from its default 50 messages to cover all messages in the conversation. This prepends the previously-hidden earlier messages at the DOM top, so the old absolute scrollTop (e.g. 38,000px) now maps to only the top ~24% of the new DOM instead of near the bottom — the user sees earliest messages instead of the ones they were reading.

A secondary contributor: `inner.innerHTML=''` during the DOM rebuild fires a native scroll event (scrollTop clamp) that was not guarded by `_programmaticScroll=true`, potentially corrupting `_scrollPinned` / `_messageUserUnpinned` via the scroll listener before the snapshot restoration runs.

## Fix

Three changes in `static/ui.js`:

1. **`_restoreMessageScrollSnapshot`** — adjust scrollTop by the DOM `scrollHeight` delta so the user's visible message group is preserved when the DOM size changes between snapshot and restore.
2. **`_restoreMessageScrollSnapshotSameFrame`** — same delta compensation for the SameFrame path.
3. **`renderMessages`** — set `_programmaticScroll=true` before `inner.innerHTML=''` to prevent the native scroll event from DOM clear corrupting scroll state.

## How to Verify

1. Open a long conversation (200+ messages) in the WebUI
2. Start a new reply so the assistant streams a response
3. During streaming, **scroll up** to view earlier messages
4. When streaming completes, the viewport should stay at the position you were reading (not jump to early messages)

## Test Plan

- [x] Manual verification by the reporter
- [x] 10 mathematical scenarios validated across Python + browser JS (expansion, shrinkage, boundary, fallback)
- [x] JS syntax validation (`node -c`)
- [x] Zero performance regression (50K benchmark: 0.03μs vs 0.05μs per op)

## Risk Assessment

**Low.** Both restore functions clamp to `[0, maxTop]`. When `snapshot.scrollHeight` is undefined (old snapshots), delta degrades gracefully to 0 (current behavior).

Issue #500, opened 2026-04-14
feat(performance): virtual scroll for session list and message list


nesquena-hermes:
## Summary

Users with large session histories or very long conversations experience slow rendering and DOM bloat. Both the session sidebar and the message list currently render all items into the DOM at once.

## Proposed behaviour

Implement virtual (windowed) scrolling so only the items near the current scroll position are in the DOM:
1. Session sidebar — ~20 visible items plus a small buffer
2. Message list — messages near the current scroll position

## Implementation notes

- Use `IntersectionObserver` for a lightweight no-library approach: replace off-screen items with placeholder divs of the same height
- No bundler or external dependency needed — keeps the project's no-build-step constraint
- Session list is simpler (roughly fixed height per item); message list is harder (variable-height markdown)
- Suggested order: session list first, message list as a follow-on

Tracked in ROADMAP.md as "Virtual scroll for large lists (deferred)".

nesquena-hermes:
## Acknowledged: Virtual scroll for session list + message list ✅

Correct prioritization — session list first (fixed-height items, simpler), message list follow-on (variable height, harder).

### Implementation notes

**Session list (simpler)**:
- `IntersectionObserver` approach: render all session items but replace off-screen ones with `<div style="height:<item-height>px">` placeholders
- When a placeholder enters the viewport, swap it with the real item
- Session item height is approximately fixed (~52px) — makes height calculation easy
- This approach keeps the scroll bar accurate (total height matches real content) without a complex virtual DOM

**Message list (harder)**:
- Variable-height items (markdown messages range from 1 line to hundreds)
- Two options:
  1. **IntersectionObserver + height cache**: measure each message height on first render, cache it, replace off-screen messages with measured-height placeholders. Heights are re-measured if content changes.
  2. **Windowed renderer**: only render messages near the current scroll position (current viewport ± 2 screens buffer). On scroll, add/remove messages from top/bottom and adjust a padding spacer element to maintain scroll position.
- Option 1 is simpler to implement with the existing render pipeline; Option 2 is more memory-efficient for sessions with thousands of messages

**No-library constraint**: `IntersectionObserver` + `ResizeObserver` (for recalculating heights when content changes) are both standard Web APIs — no bundler needed.

**Suggested order**:
1. Session sidebar virtual scroll (PR 1 — contained, low risk)
2. Message list virtual scroll (PR 2 — more complex, needs careful testing with streaming)

[...]

PR #4328, opened 2026-06-16
fix(#4325): use semantic viewport anchoring for stream-end scroll restore

### LINKED ISSUE #4325: Regression: transcript virtualization (#500/#4214) causes freeze + scroll jump + reload after stream done
## Summary

Since **v0.51.440** (commit `fe12eebb` — transcript virtualization #500/#4214), the Web UI freezes, jumps to an older message, then scrolls back down after a streaming response completes. The full response renders perfectly during streaming, but the "done" finalization breaks it.

Rolling back to **v0.51.439** (`87fa0a2a`, pre-virtualization) eliminates the issue completely.

## Reproduction

1. Open any session in the Web UI (v0.51.440+)
2. Send a message that triggers a streaming response
3. During streaming: content renders fine, readable
4. When streaming completes ("done" event): **freeze → chat jumps to a much older message → scrolls back down → sometimes requires manual reload**

The effect is worse with longer sessions (more messages in the transcript).

## Root Cause Analysis

When streaming finishes, multiple competing re-render paths fire within ~5 seconds:

1. **`_finishDone()`** (messages.js:~3420) — calls `renderMessages({preserveScroll:true})` and sets `window._streamJustFinished=true` for 5s
2. **`stream_end` event** → `_restoreSettledSession()` (messages.js:~3473) — does a **full session refetch** + replaces `S.messages` + another `renderMessages({preserveScroll:true})`
3. **`_runStreamEndRecovery()`** (messages.js:~1680) — can fire **another** `_restoreSettledSession()`
4. **`refreshActiveSessionIfExternallyUpdated()`** (sessions.js:~3763) — after the 5s cooldown expires, potentially a **fourth** full `loadSession()` reload

Each path replaces `S.messages` and re-renders. With virtualization, each re-render must:
- Recalculate which messages are in the viewport window
- Rebuild the virtual scroll height
- Re-pin scroll position

The virtual scroll height calculation + window recalibration across 3-4 rapid re-renders is what causes the visible freeze + jump.

The `preserveScroll:true` flag doesn't account for the virtual list's total height changing between renders (different messages visible → different cached heights → scroll position maps to wrong message).

## Environment

- Version: v0.51.440 → v0.51.462 (all versions with virtualization)
- Pinned working version: v0.51.439 (`87fa0a2a`)
- Browser: Safari (PWA), Chrome
- Session length: 50+ messages exacerbates significantly

## Suggested Direction

The core issue is the virtual scroll height/position not surviving the multiple `renderMessages({preserveScroll:true})` calls that fire during stream finalization. Either:

1. **Dedupe the re-renders** — ensure `_finishDone()` sets `_streamFinalized=true` so `stream_end` → `_restoreSettledSession` bails out instead of re-rendering (currently `_finishDone` never sets this flag)
2. **Preserve virtual scroll anchor** — snapshot the first visible message ID + offset before re-render, restore it after, instead of relying on pixel-based `scrollTop` which breaks when virtual heights change

rodboev:
## Thinking Path

- The transcript virtualization shipped in v0.51.440 (#500/#4214) introduced a render-window mechanism that starts at 50 messages and expands to all messages on stream completion. When `_finishDone()` expands the window and calls `renderMessages({preserveScroll:true})`, previously-hidden earlier messages are prepended to the DOM, changing `scrollHeight` dramatically.
- The current scroll restore functions (`_restoreMessageScrollSnapshot`, `_restoreMessageScrollSnapshotSameFrame`) use pixel-based `scrollTop` values that become invalid when the DOM height changes underneath them. The infrastructure for semantic message anchoring already exists (`_captureMessageViewportAnchor` / `_restoreMessageViewportAnchor`) and was wired into the blank-viewport recovery path in #4281, but was never used by the primary scroll-after-render path.
- Wiring anchor-based restore as the first-choice strategy in both `_scrollAfterMessageRender` and `_restoreMessageScrollSnapshotSameFrame` makes the viewport survive DOM height changes from render-window expansion, while falling back to pixel-based restore when the anchor row is no longer rendered.

## What Changed

- `static/ui.js`: `_scrollAfterMessageRender` now tries `_restoreMessageViewportAnchor(snapshot.anchor)` before falling back to pixel-based `_restoreMessageScrollSnapshot`
- `static/ui.js`: `_restoreMessageScrollSnapshotSameFrame` tries anchor-based restore first, falls back to pixel-based bottom-distance calculation
- `static/ui.js`: `_restoreMessageScrollSnapshot` tries anchor-based restore first, falls back to the existing pixel-based path

## Why It Matters

Users in sessions with 50+ messages see a freeze, scroll jump to early messages, and sometimes need a manual page reload when a streaming response completes. The regression makes longer conversations unusable after each response finishes.

## Verification

```
pytest tests/test_issue500_message_list_virtualization.py tests/test_tars_scroll_reset_regressions.py tests/test_smooth_text_fade.py -v --timeout=60
```

## Risks / Follow-ups

- PR [#4110](https://github.com/nesquena/hermes-webui/pull/4110) takes a complementary approach (scrollHeight delta compensation in the pixel restore path). Both fixes are independently valid; if #4110 also merges, the delta compensation serves as an extra safety net for edge cases where the anchor row is not in the DOM.
- Issue [#4295](https://github.com/nesquena/hermes-webui/issues/4295) describes mid-stream viewport jumps, a distinct symptom requiring separate investigation of the streaming rAF scroll paths.

## Upstream

Closes #4325.

## Model Used

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -9708,6 +9708,7 @@ function renderMessages(options){
   const worklogDetailDisclosureState=_captureWorklogDetailDisclosureState(inner);
   _programmaticScroll=true;
   inner.innerHTML='';
+  _programmaticScroll=false;
   const compressionNode=compressionState?_compressionCardsNode(compressionState):null;
   const {message:referenceMessage, rawIdx:referenceMessageRawIdx}=_latestCompressionReferenceMessage(
     S.messages,
--- a/static/ui.js
+++ b/static/ui.js
@@ -9492,9 +9492,26 @@ function _restoreMessageScrollSnapshot(snapshot){
 function _restoreMessageScrollSnapshotSameFrame(snapshot){
   const el=$('messages');
   if(!el||!snapshot) return;
-  const restoredViaAnchor=(snapshot.anchor&&typeof _restoreMessageViewportAnchor==='function')
+  let restoredViaAnchor=(snapshot.anchor&&typeof _restoreMessageViewportAnchor==='function')
     ? _restoreMessageViewportAnchor(snapshot.anchor,0)
     : false;
+  if(!restoredViaAnchor&&snapshot.anchor&&snapshot.anchor.rawIdx!=null&&
+     !el.querySelector(`[data-msg-idx="${snapshot.anchor.rawIdx}"]`)&&
+     typeof _getVisibleMessagesWithIdx==='function'&&
+     typeof _messageVisibleIndexForRawIdx==='function'&&
+     typeof _messageVirtualScrollTopForVisibleIdx==='function'){
+    const visWithIdx=_getVisibleMessagesWithIdx();
+    const visIdx=_messageVisibleIndexForRawIdx(snapshot.anchor.rawIdx,visWithIdx);
+    if(visIdx>=0){
+      _programmaticScroll=true;
+      el.scrollTop=_messageVirtualScrollTopForVisibleIdx(visWithIdx,visIdx,el);
+      _messageVirtualWindowKey='';
+      renderMessages({preserveScroll:true});
+      restoredViaAnchor=(typeof _restoreMessageViewportAnchor==='function')
+        ? _restoreMessageViewportAnchor(snapshot.anchor,0)
+        : false;
+    }
+  }
   if(!restoredViaAnchor){
     const maxTop=Math.max(0,el.scrollHeight-el.clientHeight);
     const bottom=Number(snapshot.bottom);
--- a/static/ui.js
+++ b/static/ui.js
@@ -9492,13 +9492,18 @@ function _restoreMessageScrollSnapshot(snapshot){
 function _restoreMessageScrollSnapshotSameFrame(snapshot){
   const el=$('messages');
   if(!el||!snapshot) return;
-  const maxTop=Math.max(0,el.scrollHeight-el.clientHeight);
-  const bottom=Number(snapshot.bottom);
-  const target=(snapshot.pinned===true&&Number.isFinite(bottom))
-    ? maxTop-Math.max(0,bottom)
-    : Number(snapshot.top)||0;
-  _programmaticScroll=true;
-  el.scrollTop=Math.max(0,Math.min(target,maxTop));
+  const restoredViaAnchor=(snapshot.anchor&&typeof _restoreMessageViewportAnchor==='function')
+    ? _restoreMessageViewportAnchor(snapshot.anchor,0)
+    : false;
+  if(!restoredViaAnchor){
+    const maxTop=Math.max(0,el.scrollHeight-el.clientHeight);
+    const bottom=Number(snapshot.bottom);
+    const target=(snapshot.pinned===true&&Number.isFinite(bottom))
+      ? maxTop-Math.max(0,bottom)
+      : Number(snapshot.top)||0;
+    _programmaticScroll=true;
+    el.scrollTop=Math.max(0,Math.min(target,maxTop));
+  }
   _lastScrollTop=el.scrollTop;
   if(snapshot.pinned===true){
     _messageUserUnpinned=false;
@@ -9509,7 +9514,9 @@ function _restoreMessageScrollSnapshotSameFrame(snapshot){
     _scrollPinned=false;
     _nearBottomCount=0;
   }
-  requestAnimationFrame(()=>{ setTimeout(()=>{_programmaticScroll=false;},0); });
+  if(!restoredViaAnchor){
+    requestAnimationFrame(()=>{ setTimeout(()=>{_programmaticScroll=false;},0); });
+  }
 }
 function _renderMessagesWithScrollSnapshot(options){
   const scrollSnapshot=_captureMessageScrollSnapshot();
@@ -9699,6 +9706,7 @@ function renderMessages(options){
     S.session && typeof S.session.compression_anchor_summary==='string'
   ) ? S.session.compression_anchor_summary.trim() : '';
   const worklogDetailDisclosureState=_captureWorklogDetailDisclosureState(inner);
+  _programmaticScroll=true;
   inner.innerHTML='';
   const compressionNode=compressionState?_compressionCardsNode(compressionState):null;
   const {message:referenceMessage, rawIdx:referenceMessageRawIdx}=_latestCompressionReferenceMessage(
--- a/static/ui.js
+++ b/static/ui.js
@@ -9706,9 +9706,7 @@ function renderMessages(options){
     S.session && typeof S.session.compression_anchor_summary==='string'
   ) ? S.session.compression_anchor_summary.trim() : '';
   const worklogDetailDisclosureState=_captureWorklogDetailDisclosureState(inner);
-  _programmaticScroll=true;
   inner.innerHTML='';
-  _programmaticScroll=false;
   const compressionNode=compressionState?_compressionCardsNode(compressionState):null;
   const {message:referenceMessage, rawIdx:referenceMessageRawIdx}=_latestCompressionReferenceMessage(
     S.messages,
--- a/tests/test_issue500_message_list_virtualization.py
+++ b/tests/test_issue500_message_list_virtualization.py
@@ -427,3 +427,93 @@ console.log(JSON.stringify({recovered, deletes, renderCalls}));
     assert metrics["recovered"] is True
     assert metrics["deletes"] == ["sid-123"]
     assert metrics["renderCalls"] == [{"preserveScroll": True, "_virtualFallback": True}]
+
+
+def test_same_frame_restore_nudges_virtual_window_when_anchor_row_is_missing():
+    js = UI_JS_PATH.read_text(encoding="utf-8")
+    source = _extract_func_script(js) + r"""
+const ROW_HEIGHT = 120;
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Reading `static/ui.js:9452-9520` on `origin/master` against this diff, the change brings the synchronous restore path into parity with the async one and lines up with "Suggested Direction #2" in #4325 (snapshot a message anchor instead of pixel `scrollTop`).

## Code reference

The snapshot already carries an anchor (`static/ui.js:9457`):

```js
return {
  anchor:(typeof _captureMessageViewportAnchor==='function')?_captureMessageViewportAnchor():null,
  top:el.scrollTop, bottom, scrollHeight:el.scrollHeight,
  pinned:_shouldFollowMessagesOnDomReplace(), userUnpinned:_messageUserUnpinned,
};
```

`_captureMessageViewportAnchor` (`static/ui.js:660-674`) records `{rawIdx, topOffset}` of the first row whose bottom is below the viewport top, and `_restoreMessageViewportAnchor` (`static/ui.js:675-689`) re-pins by `data-msg-idx`. The async `_restoreMessageScrollSnapshot` (`static/ui.js:9469`) already prefers that anchor; this PR adds the same `restoredViaAnchor` short-circuit to `_restoreMessageScrollSnapshotSameFrame`, which is what `_renderMessagesWithScrollSnapshot` (`static/ui.js:9521-9524`) calls during stream finalization. That's exactly where the issue says the pixel-based restore "maps to wrong message" when virtual heights shift.

## One thing worth confirming

The anchor restore is fundamentally DOM-driven: `_restoreMessageViewportAnchor` bails (`return false`) if `container.querySelector('[data-msg-idx="..."]')` doesn't find the row (`static/ui.js:680-681`). With transcript virtualization (#500/#4214), the anchored `rawIdx` can be *outside the rendered window* after the re-render that finalization triggers — in which case `restoredViaAnchor` is `false` and you fall back to the same pixel math that #4325 says is broken. Worth a quick check that the captured first-visible row is still mounted post-render in the long-session case (50+ messages) the reporter calls out, otherwise the freeze/jump can still surface for the exact transcripts that hit it hardest.

## Verification

Since #4325 lists four competing finalization re-render paths (`_finishDone`, `stream_end` → `_restoreSettledSession`, `_runStreamEndRecovery`, `refreshActiveSessionIfExternallyUpdated` — all in `static/messages.js` / `static/sessions.js`), a manual repro on a 50+ message session in Safari PWA with a long streamed reply is the real test: confirm no upward jump on the `done` event, and confirm the anchored row stays put when its `rawIdx` is near the top of the rendered window. The `_programmaticScroll` bracketing around `inner.innerHTML=''` in `renderMessages` (`static/ui.js:9706+` in this diff) is a good touch to keep the wipe from tripping the unpin heuristic.

rodboev:
Removed the synchronous `_programmaticScroll` toggle around `inner.innerHTML=''` in `b7b8cd88`. As the review notes, the subsequent `_scrollAfterMessageRender` path already manages the flag correctly via rAF-based cleanup, so the synchronous bracket was both a no-op for async scroll event suppression and a hazard for concurrent scroll guards.

The CI failure on `test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata` is unrelated to this change (it's a messaging-import test, not a scroll test).

Focused validation passed locally: `tests/test_issue500_message_list_virtualization.py`, `tests/test_tars_scroll_reset_regressions.py`, `tests/test_smooth_text_fade.py` (29 passed), plus ESLint on `static/**/*.js`.

nesquena-hermes:
## Re-review — the `_programmaticScroll` cleanup is right, CI red is unrelated, but the anchor-out-of-window gap is still open

Thanks @rodboev. Reading the PR head (`f24bd81b`) against master, two of the three things resolve cleanly:

**1. Synchronous `_programmaticScroll` removal — correct.** `f24bd81b` drops exactly the right two lines around the DOM wipe:

```js
-  _programmaticScroll=true;
   inner.innerHTML='';
-  _programmaticScroll=false;
```

`inner.innerHTML=''` doesn't dispatch a `scroll` event synchronously, so the bracket suppressed nothing, and leaving `_programmaticScroll=true` set across the rebuild was the real hazard for any concurrent guard. The downstream `_scrollAfterMessageRender` / `_restoreMessageScrollSnapshotSameFrame` paths own the flag via rAF teardown now (`static/ui.js:9473`, `9504`). Good.

**2. CI failure is genuinely unrelated.** `test_sessions_response_backfills_imported_messaging_source_metadata` lives at `tests/test_gateway_sync.py:1323` and exercises gateway messaging-source backfill — no scroll surface. Not a blocker for this PR.

**3. The anchor-out-of-window concern from my last pass is still unaddressed.** This is the one that matters, and "focused validation passed locally" doesn't cover it. `_restoreMessageViewportAnchor` is a pure DOM lookup that bails when the row isn't mounted (`static/ui.js:678-681`):

```js
const row=container.querySelector(`[data-msg-idx="${targetIdx}"]`);
if(!row) return false;
```

`return false` drops you straight back to the pixel math this PR is trying to replace. The question is whether the anchored `rawIdx` is *guaranteed* mounted after the finalization re-render — and with transcript virtualization (#500/#4214) it isn't. The virtual window is computed from the live scroll position (`static/ui.js:615-622`):

```js
return _messageVirtualWindow({
  total:visWithIdx.length,
  scrollTop:container?container.scrollTop:0,
  viewportHeight:container?container.clientHeight:(...),
  ...
});
```

When `_finishDone` expands `_messageRenderWindowSize` to all messages, earlier rows get prepended and `scrollHeight` jumps — but the *old* `scrollTop` (read here) now maps to a different logical slice. So the virtual window can mount a band that does **not** contain the row `_captureMessageViewportAnchor` recorded (`static/ui.js:660-674`, "first row whose bottom is below the viewport top"). When `rawIdx` lands outside `[start,end)`, the `querySelector` is null → `return false` → broken pixel fallback, for exactly the 50+ message Safari-PWA transcript the reporter hit.

### Suggested confirmation / fix
- Add a DOM/structure test (extend `tests/test_issue500_message_list_virtualization.py`) with a virtualized transcript where the captured anchor `rawIdx` is *outside* the post-expansion mounted slice, and assert `_restoreMessageScrollSnapshotSameFrame` still lands on the right message (i.e. that the anchor row is force-mounted, or the window is recomputed around the anchor before restore).
- If the row genuinely can be virtualized out, the restore needs to nudge the virtual window to include `anchor.rawIdx` *before* calling `_restoreMessageViewportAnchor` — otherwise the anchor path silently no-ops on the hardest sessions. The blank-viewport recovery at `static/ui.js:9756` (`_maybeRecoverVirtualizedBlankViewport`) is the closest existing precedent for "remount around the intended position, then restore."

Everything else looks ship-shape. Worth nailing down case 3 with a test before merge, since it's the precise scenario #4325 is filed against.

rodboev:
Good catch on the anchor-out-of-window case — 18a3a6ce addresses it.

When _restoreMessageViewportAnchor returns alse (row absent from DOM) and a querySelector confirms the row was virtualized out, _restoreMessageScrollSnapshotSameFrame now:

1. Looks up the visible index for nchor.rawIdx via _messageVisibleIndexForRawIdx
2. Computes a scrollTop that brings that row into the virtual window using _messageVirtualScrollTopForVisibleIdx (the same helper used by jumpToSessionStart)
3. Resets _messageVirtualWindowKey='' and calls 
enderMessages({preserveScroll:true}) to mount the row
4. Retries _restoreMessageViewportAnchor — which now finds the row and succeeds

Added 	est_same_frame_restore_nudges_virtual_window_when_anchor_row_is_missing to 	est_issue500_message_list_virtualization.py: simulates a 60-message session with rows 0-9 virtualized out, anchor at 
awIdx=5, and asserts that 
enderMessages is called with preserveScroll:true and that scrollTop is adjusted to 390 (the height-prefix offset for row 5 minus the 35% viewport bias) before the re-render.

nesquena-hermes:
Heads-up to avoid duplicate work: this PR's scroll fix has been bundled into the combined #4325 resolution release **#4342** (`mega-4325`), which pairs it with a virtualization opt-out toggle to fully resolve #4325. All 4 commits here are carried forward byte-for-byte with @rodboev's authorship preserved + a `Co-authored-by` trailer, and credited in the CHANGELOG + release notes.

#4342 is gated (Codex + Opus + full suite all green) and is awaiting an independent maintainer review before merge. Once #4342 merges, this PR will be closed as superseded (with credit). Leaving it open for now in case the review surfaces anything.

nesquena-hermes:
Shipped in **v0.51.469** 🎉 — this scroll fix landed as **Part A** of the combined #4325 resolution (PR #4342), paired with a "Virtualize long transcripts" opt-out toggle so #4325 is fully resolved from both angles.

Your 4 commits were carried forward **byte-for-byte with your authorship preserved** (`Co-authored-by` trailer), and you're credited in the CHANGELOG (v0.51.469) and the GitHub release notes. The semantic viewport anchoring (re-pin by message-id instead of pixel scrollTop, + the virtualized-out-anchor mount-and-retry) is exactly what fixed the freeze/jump/snap-back.

Closing as superseded-by-merge. Thank you for the contribution, @rodboev! 🙏

PR #4329, opened 2026-06-16
fix(#4324): gate credential_pool detection to known model providers

### LINKED ISSUE #4324: bug(providers): non-model credential_pool keys (Photon plugin) appear as 3 phantom providers each with full 251-model catalog — regression from #4247
## Summary

After setting up the **Photon** iMessage plugin (`hermes photon setup`), the WebUI model picker shows **three phantom providers** — `photon`, `photon project`, and `photon user` — each listing the **full auto-detected model catalog (251 models on the reporter's machine)**. These are not model providers; they are Photon's messaging-platform credentials, and they should never appear in the model picker at all.

This is a regression introduced by **#4247** (v0.51.461, 2026-06-16): the new credential-pool provider-detection loop adds **every** `auth.json` `credential_pool` key as a detected model provider, without checking that the key actually names a model provider.

## Reporter (verbatim, Discord #webui, 2026-06-16)

> The only other thing I haven't been able to figure out about this is that I get a list of 251 models each for providers called 'photon', 'photon project' and 'photon user'.
>
> These are presumably related to the Photon iMessage plugin. I first saw them after I setup Photon to allow my wife and me to message a Photon number to interact with our agent. I've asked the agent if it can figure out why these are showing up and how to remove them, but it looks like that's not something it can do without modifying the Hermes-WebUI codebase. Any idea why this might be the case?
>
> — Nic

Reporter is on `v0.51.462-dirty-43213e29`.

## Root cause

The Photon plugin (`plugins/platforms/photon/auth.py` in hermes-agent) writes **three** entries into `auth.json` → `credential_pool`:

- `credential_pool.photon` — Photon login/device token
- `credential_pool.photon_project` — dashboard/spectrum project id + secret
- `credential_pool.photon_user` — operator number + assigned text line

These are **messaging-platform credentials, not LLM API keys.**

#4247 added a credential-pool detection loop in `api/config.py::_build_available_models_uncached` (~L5000-5052) that iterates **every** key in `credential_pool` and, if any non-ambient entry exists, adds the canonicalised key to `detected_providers`:

```python
# api/config.py ~L5003-5035
for _pid in list(_pool.keys()):
    _canonical_pid = _resolve_provider_alias(str(_pid))
    ...
    _explicit = [e for e in _all_entries
                 if not _is_ambient_gh_cli_entry(...)]
    if _explicit:
        detected_providers.add(_canonical_pid)   # ← no model-provider gate
```

So `photon`, `photon_project`, `photon_user` all enter `detected_providers`. After canonicalisation they become `photon`, `photon-project`, `photon-user` (rendered as "photon", "photon project", "photon user").

Then the group-builder loop (~L5698) has no branch for these unknown ids — they are not in `_PROVIDER_MODELS`, not in `_canonical_to_raw_provider_key`, and `_is_plugin_model_provider()` is `False` for all three — so each falls through to the generic **`else`** branch (~L6033) which copies the **global `auto_detected_models` catalog** into the group:

```python
else:
    detected_models = auto_detected_models_by_provider.get(pid)   # empty for photon*
    if detected_models:
        models_for_group = copy.deepcopy(detected_models)
    elif auto_detected_models:                                    # ← global catalog
        ...
        models_for_group = copy.deepcopy(auto_detected_models)    # all 251 models
```

That's why each of the three phantom providers shows the same 251-model list.

### Verified (on master, v0.51.463)

```
$ python3 -c "from hermes_cli.models import list_available_providers as lap; \
              print([p['id'] for p in lap() if 'photon' in p['id'].lower()])"
[]                          # photon* is NOT a model provider in the agent registry (38 real providers)

$ python3 -c "from api.config import _PROVIDER_MODELS, _is_plugin_model_provider; \
              print({p:(p in _PROVIDER_MODELS, _is_plugin_model_provider(p)) \
                     for p in ['photon','photon-project','photon-user']})"
{'photon': (False, False), 'photon-project': (False, False), 'photon-user': (False, False)}
```

So all three keys reach the global-catalog `else` fallback unguarded.

## Suggested fix / fix shape

Gate the credential-pool detection so only keys that name a **recognized model provider** are added to `detected_providers`. The check belongs in `api/config.py::_build_available_models_uncached`, in the pool loop at ~L5024-5050 (both the `load_pool` branch and the `ImportError` raw-dict fallback branch).

Add a model-provider gate before `detected_providers.add(_canonical_pid)`:

```python
if _explicit and _is_known_model_provider(_canonical_pid):
    detected_providers.add(_canonical_pid)
```

where `_is_known_model_provider(pid)` is true when `pid` is any of:
- in the agent registry — `{p["id"] for p in hermes_cli.models.list_available_providers()}`
- in `_PROVIDER_MODELS` / `_PROVIDER_DISPLAY`
- a plugin **model** provider — `_is_plugin_model_provider(pid)`
- a configured `custom_providers[]` slug (`custom:*`)

Photon's three keys match none of these, so they stop appearing — while #4247's intended behaviour (surfacing custom/aliased providers whose API keys live in the pool) is preserved.

**Defensive belt-and-braces (recommended in addition):** the generic `else` fallback at ~L6033 should not dump the global `auto_detected_models` catalog into an unknown provider id. A provider with no catalog of its own should be **omitted** from the picker (the same posture #1572/#7372 took for misattributed default models), not given every model. This prevents any future non-model pool key from regressing the same way.

## Scope

- **Single repo:** `nesquena/hermes-webui` only. The Photon plugin writing `credential_pool.photon*` is correct/expected behaviour on the agent side — the bug is purely WebUI's over-broad pool→provider detection.
- ~15-30 LOC: a small `_is_known_model_provider()` helper + two call-site guards in the pool loop, plus (recommended) the `else`-branch omission.
- Regression test: seed an `auth.json` with a `credential_pool.photon` (and `photon_project`/`photon_user`) entry alongside a real provider, assert the photon ids do **not** appear as groups in `get_available_models()` while the real provider still does. Watch the CI trap — mock at the WebUI boundary / inject a stub `agent.credential_pool` into `sys.modules` rather than `import agent.*` (see `test_credential_pool_profile_scoping.py` for the established pattern; the bundled `agent` package isn't installed in CI).

## Reported by

**Nic** (Discord #webui, 2026-06-16), running `v0.51.462`. Same reporter who raised the Nous Portal 15-model cap / model-picker thread.

rodboev:
## Thinking Path

- The #4247 credential-pool detection loop iterates every `credential_pool` key and adds it to `detected_providers` without checking whether the key actually names a model provider.
- The Photon iMessage plugin writes three messaging-platform credential entries (`photon`, `photon_project`, `photon_user`) that are not LLM API keys, and all three pass through unguarded.
- The group-builder `else` fallback then assigns the full auto-detected model catalog to each unknown provider ID, producing three phantom providers with 251 models each.
- The fix gates pool detection to known model providers and removes the global-catalog dump for unknown provider IDs.

## What Changed

- `api/config.py`: extract `_is_known_model_provider(pid)` helper that checks `_PROVIDER_MODELS`, `_PROVIDER_DISPLAY`, `_is_plugin_model_provider()`, and `custom:*` slugs; gate pool detection behind it; restrict global-catalog `else` fallback to only the `custom` pseudo-provider
- `tests/test_credential_pool_phantom_providers.py`: regression test seeding non-model pool keys alongside a real provider, asserting phantom providers are excluded and real providers survive

## Why It Matters

Users with Photon (or any non-model plugin that writes to `credential_pool`) see three phantom provider groups in the model picker, each showing the full 251-model catalog. This makes the model picker unusable for affected users and was reported on Discord by a real user running v0.51.462.

## Verification

```
pytest tests/test_credential_pool_phantom_providers.py -v --timeout=60
```

## Risks / Follow-ups

- The defensive `else`-branch change means future providers that ship only via `credential_pool` without a `_PROVIDER_MODELS` entry will not appear in the picker until their catalog is registered. This is the correct posture: a provider with no catalog of its own should not inherit the global catalog.
- Plugin model providers that rely on `_is_plugin_model_provider()` are still surfaced correctly.

## Upstream

Closes #4324.

## Model Used

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -981,6 +981,26 @@ def _resolve_provider_alias(name: str) -> str:
     return _PROVIDER_ALIASES.get(raw, name)
 
 
+def _is_known_model_provider(pid: str | None) -> bool:
+    """Return True when *pid* names a recognized model provider."""
+    if not pid:
+        return False
+
+    # Check against known provider dictionaries
+    if pid in _PROVIDER_MODELS or pid in _PROVIDER_DISPLAY:
+        return True
+
+    # Check if it's a plugin model provider
+    if _is_plugin_model_provider(pid):
+        return True
+
+    # Check for custom:* pattern
+    if pid.startswith("custom:"):
+        return True
+
+    return False
+
+
 def _custom_provider_slug_from_name(name: object) -> str:
     raw = str(name or "").strip().lower()
     if not raw:
@@ -5029,7 +5049,7 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                                     str(getattr(e, "key_source", "") or ""),
                                 )
                             ]
-                            if _explicit:
+                            if _explicit and _is_known_model_provider(_canonical_pid):
                                 detected_providers.add(_canonical_pid)
                         except Exception:
                             logger.debug("credential_pool.load_pool(%s) failed", _pid)
@@ -5047,7 +5067,9 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                             for _entry in _entries
                         )
                         if _has_explicit_cred:
-                            detected_providers.add(_resolve_provider_alias(str(_pid)))
+                            _canonical_pid = _resolve_provider_alias(str(_pid))
+                            if _is_known_model_provider(_canonical_pid):
+                                detected_providers.add(_canonical_pid)
         except Exception:
             logger.debug("Failed to inspect credential_pool from auth store")
 
@@ -6034,7 +6056,7 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                     detected_models = auto_detected_models_by_provider.get(pid)
                     if detected_models:
                         models_for_group = copy.deepcopy(detected_models)
-                    elif auto_detected_models:
+                    elif auto_detected_models and pid == "custom":
                         # Don't fall back to the global auto_detected_models
                         # list for the bare "custom" PID when the active
                         # provider is something concrete (e.g. ai-gateway,
@@ -6042,11 +6064,13 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                         # belong to the active provider's group — copying
                         # them into a Custom group too produces phantom
                         # duplicates with mismatched prefixes (#1881).
-                        if pid == "custom" and active_provider and active_provider != "custom":
+                        if active_provider and active_provider != "custom":
                             models_for_group = []
                         else:
                             models_for_group = copy.deepcopy(auto_detected_models)
                     else:
+                        # Unknown provider with no provider-specific catalog:
+                        # don't fall back to the global catalog, skip entirely.
                         models_for_group = []
                     if models_for_group:
                         # Per-group deep copy so subsequent mutation by
--- a/api/config.py
+++ b/api/config.py
@@ -986,6 +986,8 @@ def _is_known_model_provider(pid: str | None) -> bool:
     if not pid:
         return False
 
+    pid = pid.strip().lower()
+
     # Check against known provider dictionaries
     if pid in _PROVIDER_MODELS or pid in _PROVIDER_DISPLAY:
         return True
@@ -6056,21 +6058,12 @@ def get_available_models(*, prefer_cache: bool = False) -> dict:
                     detected_models = auto_detected_models_by_provider.get(pid)
                     if detected_models:
                         models_for_group = copy.deepcopy(detected_models)
-                    elif auto_detected_models and pid == "custom":
-                        # Don't fall back to the global auto_detected_models
-                        # list for the bare "custom" PID when the active
-                        # provider is something concrete (e.g. ai-gateway,
-                        # openrouter). Those auto-detected entries already
-                        # belong to the active provider's group — copying
-                        # them into a Custom group too produces phantom
-                        # duplicates with mismatched prefixes (#1881).
-                        if active_provider and active_provider != "custom":
+                    elif auto_detected_models and (pid == "custom" or _is_known_model_provider(pid)):
+                        if active_provider and active_provider != "custom" and pid == "custom":
                             models_for_group = []
                         else:
                             models_for_group = copy.deepcopy(auto_detected_models)
                     else:
-                        # Unknown provider with no provider-specific catalog:
-                        # don't fall back to the global catalog, skip entirely.
                         models_for_group = []
                     if models_for_group:
                         # Per-group deep copy so subsequent mutation by
--- a/tests/test_credential_pool_phantom_providers.py
+++ b/tests/test_credential_pool_phantom_providers.py
@@ -1,0 +1,134 @@
+"""Regression test: credential_pool phantom providers with non-model credentials.
+
+Guards the regression from #4247 where non-model credential_pool keys
+(e.g., Photon plugin messaging tokens) were incorrectly treated as phantom
+model providers and assigned the full global model catalog.
+
+The pool detection loop must gate pool keys to known model providers.
+The group-builder else fallback must not assign the global catalog to
+unknown providers.
+
+Note: `_build_available_models_uncached` does `from agent.credential_pool
+import load_pool` at call time. The bundled `agent` package is importable
+locally but NOT in CI's environment, so the test injects a stub
+`agent.credential_pool` module into sys.modules.
+"""
+import sys
+import types
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+import api.config as config  # noqa: E402
+
+
+class _FakeEntry:
+    def __init__(self, source="config_yaml", label="anthropic", key_source="config_yaml"):
+        self.source = source
+        self.label = label
+        self.key_source = key_source
+
+
+class _FakePool:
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Reading `api/config.py:5049-5070` and the group-builder `else` at `api/config.py:6033-6050` on `origin/master` against this diff, the core gate is right: adding `_is_known_model_provider(_canonical_pid)` before `detected_providers.add(...)` in both the `load_pool` branch and the raw-dict fallback is exactly what stops the Photon `photon` / `photon_project` / `photon_user` pool keys from entering detection. The `test_credential_pool_phantom_providers.py` stub-injection of `agent.credential_pool` into `sys.modules` also correctly dodges the CI trap (#4324 calls this out).

Two things I'd flag before merge — note there's a parallel maintainer PR (#4330) fixing the same issue, and these are the two spots where the two diffs diverge:

## 1. `_is_known_model_provider` doesn't normalize case

```python
# this PR, api/config.py:984-1000
def _is_known_model_provider(pid: str | None) -> bool:
    if not pid: return False
    if pid in _PROVIDER_MODELS or pid in _PROVIDER_DISPLAY:
        return True
```

`_PROVIDER_MODELS` / `_PROVIDER_DISPLAY` keys are all lowercase (`api/config.py:1273`, `:871`), and `_resolve_provider_alias` lowercases internally (`api/config.py:972`) — but it returns the *original* `name` on a miss (`api/config.py:981 return _PROVIDER_ALIASES.get(raw, name)`), so a mixed-case pool key like `Anthropic` that isn't an alias passes through with original case and would miss the dict lookup. #4330's helper does `pid = (provider_id or "").strip().lower()` first. A `.strip().lower()` here would make the membership test robust to that.

## 2. The `else`-branch narrowing may strip catalogs from real catalog-less providers

```python
# this PR, api/config.py:6056-6063
elif auto_detected_models and pid == "custom":
    if active_provider and active_provider != "custom":
        models_for_group = []
    else:
        models_for_group = copy.deepcopy(auto_detected_models)
else:
    models_for_group = []
```

Collapsing the fallback to `pid == "custom"` (exact match) means any *recognized* provider that has no `_PROVIDER_MODELS` catalog and relies on the global `auto_detected_models` list now gets an empty group. Several real providers are in `_PROVIDER_DISPLAY` but NOT `_PROVIDER_MODELS` — I checked: `ollama`, `ollama-cloud`, `openrouter`, `lmstudio`, `huggingface`, `meta-llama`, `alibaba`. Most of those route through the *known-provider* branch at `api/config.py:5992-6032` (because they're in `_PROVIDER_MODELS` / `_canonical_to_raw_provider_key` / are plugin providers), so they may never reach this `else` — but `openrouter`/`ollama` auto-detected via `/v1/models` with no static catalog are the kind of provider that historically depended on this fallback. #4330 keeps a wider condition: `elif auto_detected_models and (pid == "custom" or _is_known_model_provider(pid))`, which preserves the catalog for known providers while still dropping unknown phantom ids. That's the safer narrowing — worth aligning with it so a legitimate catalog-less provider doesn't silently lose its model list.

## Recommendation

The detection gate (the primary fix) is sound and well-tested. I'd (a) lowercase in `_is_known_model_provider`, and (b) widen the `else` fallback to `pid == "custom" or _is_known_model_provider(pid)` to avoid regressing catalog-less known providers. Since #4330 already encodes both, the maintainer may consolidate — but either way these two deltas are the ones to reconcile.

rodboev:
Good catches on both points, aligned with #4330's approach in `0f6cf289`:

1. `_is_known_model_provider` now normalizes with `.strip().lower()` before the dict lookups, so mixed-case pool keys like `Anthropic` are caught correctly.

2. Widened the `else` fallback to `pid == "custom" or _is_known_model_provider(pid)`, so recognized providers without a static catalog (e.g. `openrouter`, `ollama` auto-detected via `/v1/models`) still get the global catalog instead of an empty group. The `custom`-specific duplicate suppression is preserved by keeping the `active_provider != "custom"` gate conditional on `pid == "custom"`.

Both changes validated locally against `tests/test_credential_pool_phantom_providers.py` (2 passed).

nesquena-hermes:
Thanks @rodboev — you independently found and fixed the same #4324 phantom-providers regression as #4330, and your root-cause analysis (the unguarded `credential_pool` loop + the global-catalog `else` fallback) was spot on. We integrated the fix into release branch `release/provider-picker-bundle` (#4332). We used #4330's implementation as the base because its regression test mocks at the WebUI boundary (stubs `hermes_cli` and never imports the bundled `agent` package, which isn't installed in CI) and covers both detection paths with 8 cases — but your fix is functionally equivalent and your analysis confirmed the approach. Your authorship is preserved via a Co-authored-by trailer on the integrated commit. Holding this open until the bundle ships, then closing as integrated (superseded by #4332). Appreciate the contribution!

nesquena-hermes:
Shipped in **v0.51.464** (Release PY) via #4332. Thanks again @rodboev — you independently found and fixed the same #4324 phantom-providers regression, and your root-cause analysis was spot on. We used #4330's implementation as the integration base (WebUI-boundary test mocking that avoids the CI `agent`-import trap + 8-case coverage), but your fix was functionally equivalent and your authorship is preserved via a Co-authored-by trailer on the shipped commit. Closing as integrated (superseded by #4332). Appreciate the contribution! 🙏

Issue #4331, opened 2026-06-16
test: cover the model-picker in-place overflow reveal path in CI (#3691 driver only exercises the fallback)


nesquena-hermes:
## Summary

Follow-up from the collapsible-provider-groups model-picker work (#4279 line + #4324 fix). The model picker's **in-place overflow reveal** path (`_expandOverflowGroup` in `static/ui.js`) is currently only verified in the browser — CI does not exercise it.

The #3691 node-driver test (`tests/test_issue3691_model_picker_show_all.py`) uses a minimal fake DOM with no `CSS` global, so `_canInPlace` evaluates to `false` and the driver only exercises the **full-re-render fallback** branch, not the in-place branch where the recent regression fixes actually live:

- **bug #1** — search→clear keeping an in-place-expanded group expanded (the `_modelData.hiddenByDefault` / `meta.hiddenCount` sync)
- **bug #3** — show-more not going dead when overflow options already exist as `<option>`s (the `extraModels.length` guard)
- the scroll-to-newly-revealed-rows behavior

These three scenarios are all currently verified manually via agent-browser CDP E2E (desktop + mobile), and pass — but there's no automated CI guard, so a future refactor could silently regress them.

## Suggested fix

Extend the `_DROPDOWN_DRIVER` fake DOM in `tests/test_issue3691_model_picker_show_all.py` so `_canInPlace` is true and the in-place branch runs:

- add a `CSS.escape` global (`s => String(s).replace(/[^a-zA-Z0-9_-]/g, '\\$&')` is enough for the data-group keys used)
- add `requestAnimationFrame`/`setTimeout` shims (run callback synchronously or as a microtask)
- give the fake nodes `insertBefore`, `previousElementSibling`, and an `offsetTop` stub
- give `querySelectorAll`/`querySelector` a minimal attribute-selector match for `.model-group-body[data-group="..."]` and `.model-opt .model-opt-id`

Then add three assertions driving the real in-place path:
1. expand overflow in place → type a search → clear it → group still shows all rows, no "Show more" reappears
2. a group with `dataset.modelsEndpointError` renders open by default (hint visible)
3. a group whose overflow models already exist as `<option>`s still reveals them when "Show more" is clicked (not dead)

## Context

- Opus advisor flagged this as the one post-merge follow-up on the release; not a ship blocker (logic is sound under reading + browser-verified).
[...]

PR #4330, opened 2026-06-16
fix(providers): gate credential_pool detection to known model providers (#4324)

nesquena-hermes:
## Summary

Fixes #4324. After setting up the **Photon** iMessage plugin, the WebUI model picker showed **three phantom providers** — `photon`, `photon project`, `photon user` — each listing the full auto-detected model catalog (251 models on the reporter's machine). These are messaging-platform credentials, not model providers, and should never appear in the picker.

This is a regression from **#4247** (v0.51.461): the new credential-pool provider-detection loop added **every** `auth.json` → `credential_pool` key to `detected_providers` without checking that the key actually names a model provider. The Photon plugin writes three pool keys (`photon`, `photon_project`, `photon_user`), so all three entered detection. With no catalog of their own, each fell through the group-builder's generic `else` branch, which copied the **global `auto_detected_models` catalog** (the reporter's base-url-probed models) into every group — hence three phantom providers each showing the same full list.

## Fix

1. **New helper `_is_known_model_provider(pid)`** — true only when `pid` is a `custom:*` slug, appears in WebUI's static `_PROVIDER_DISPLAY` / `_PROVIDER_MODELS` tables, or is a registered model-provider plugin (`_is_plugin_model_provider`). Photon's three keys match none.
2. **Gate both pool-detection branches** with it — the `load_pool` primary path and the `ImportError` raw-dict fallback — before `detected_providers.add()`. The canonical (post-alias) pid is checked in both. #4247's intended behaviour is preserved: real providers whose API keys live in the pool (`copilot`, `ollama-cloud`, aliased `google`→`gemini`) still detect.
3. **Defensive belt-and-braces** — the group-builder's generic `else` branch no longer paints an unrecognized id with the global `auto_detected_models` catalog (only the bare `custom` endpoint or a known provider). This matches the #1572/#7372 "omit rather than misattribute" posture and prevents any future non-model pool key from regressing the same way. A genuinely-probed provider is unaffected: its own group is filled from `auto_detected_models_by_provider[pid]` first.

## Testing

New `tests/test_issue4324_photon_phantom_providers.py` (9 tests):

- Reproduces the reporter's exact scenario — photon* pool keys **plus** a base-url probe populating the global catalog — on **both** the `load_pool` and `ImportError` detection paths. **Verified failing pre-fix** (the base-url-catalog test reproduces the phantom-group bug), passing post-fix.
[...]

PR #4332, opened 2026-06-16
Release: collapsible model-picker provider groups + phantom-providers fix (#4279, #4324)

nesquena-hermes:
## What this is

A release bundle of two coherent model-picker changes, staged for independent review before tagging.

### 1. Collapsible provider groups (feature) — extends #4279
- Provider groups start **collapsed except the group that owns the selected model**, so the picker opens scannable instead of as a long flat list.
- **Search** auto-expands matching groups (and searches hidden overflow models); **clearing** the search restores the collapsed-except-selected view.
- **"Show more"** restyled as a distinct `+ Show all N models` expander (muted italic, indented, no model-id line) that reveals the overflow **in place** — the group stays open, other groups keep their state, and the view scrolls to the newly-revealed models instead of resetting to the top. Falls back to a full re-render in minimal/headless DOM.
- Endpoint-error groups (#4253) stay open by default so the warning is visible; the expanded group persists across a search round-trip.
- The redundant per-row provider chip is suppressed for rows already under their own provider heading (kept for hoisted/search rows).
- Co-authored with @akrhin (collapsible-groups foundation).

### 2. Phantom-providers fix (#4324) — regression from #4247
After the **Photon** iMessage plugin is set up, the picker showed three phantom providers (`photon`, `photon project`, `photon user`) each listing the full auto-detected catalog. The #4247 credential-pool detection loop added every `auth.json` → `credential_pool` key without checking it names a model provider. Now gated on `_is_known_model_provider()` (a `custom:*` slug, a static `_PROVIDER_DISPLAY`/`_PROVIDER_MODELS` entry, or a registered model-provider plugin) on both detection paths, and the group-builder no longer paints an unknown id with the global catalog.

This integrates the stronger of two **independent same-fix PRs**: #4330's implementation + 8-case WebUI-boundary test, chosen over #4329's 2-case version. Both @rodboev (#4329) and the maintainer agent (#4330) found and fixed the same regression — co-authored with @rodboev. Reported by Nic.

## Verification

- **Full pytest suite: 9251 passed, 0 failed.**
- **Codex advisor: SAFE TO SHIP** (80 targeted model-picker/credential-pool/endpoint-error tests + `node --check`).
- **Opus advisor: SAFE TO SHIP** — reviewed twice; flagged three bugs in the first pass (search-clear lost expansion; endpoint-error hint hidden in collapsed group; dead show-more when overflow options already exist), **all three fixed and re-verified** in the second pass.
- **Browser E2E (agent-browser CDP, desktop 1440 + mobile 390):** collapse/expand/collapse, search auto-expand + overflow-search + clear-reset + no-results, load-more in-place + stay-open + scroll-to-new, endpoint-error visibility — all green. Zero console errors across the full interaction flow.
- ESLint runtime gate + browser-smoke: clean.

## Follow-up
- #4331 — add CI coverage for the in-place reveal path (the #3691 node driver currently exercises only the full-re-render fallback). Logic is browser-verified; tracked rather than rushed into this branch.

## Review notes
- `static/ui.js` carries the feature; `api/config.py` carries the #4324 gate; the two are independent.
- The in-place reveal's cross-render "stay open" intent persists on a lazy `globalThis.__modelGroupForceOpen` Set, kept in sync by the heading-toggle handler (add on open, delete on manual collapse).

Staged for an independent review pass — not yet tagged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ## [Unreleased]
 
+## [v0.51.464] — 2026-06-16 — Release PY (scannable model picker)
+
 ### Added
 
 - **Model picker provider groups are now collapsible and open scannable (#4279).** Provider groups in the model picker start collapsed except the group that owns your currently-selected model, so a picker with many providers/models opens as a tidy list of provider headings rather than one long flat scroll. Searching auto-expands matching groups (and searches hidden overflow models); clearing the search restores the collapsed-except-selected view. The "Show all N models" affordance is restyled as a distinct "+ Show all N models" expander (muted italic, indented, not mistakable for a selectable model) and now reveals the overflow in place — the group stays open and the view scrolls to the newly-revealed models instead of resetting to the top. The redundant per-row provider chip is suppressed for rows already under their own provider heading (kept only for hoisted/search rows). Builds on @akrhin's collapsible-groups work.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 
 ## [Unreleased]
 
+### Added
+
+- **Model picker provider groups are now collapsible and open scannable (#4279).** Provider groups in the model picker start collapsed except the group that owns your currently-selected model, so a picker with many providers/models opens as a tidy list of provider headings rather than one long flat scroll. Searching auto-expands matching groups (and searches hidden overflow models); clearing the search restores the collapsed-except-selected view. The "Show all N models" affordance is restyled as a distinct "+ Show all N models" expander (muted italic, indented, not mistakable for a selectable model) and now reveals the overflow in place — the group stays open and the view scrolls to the newly-revealed models instead of resetting to the top. The redundant per-row provider chip is suppressed for rows already under their own provider heading (kept only for hoisted/search rows). Builds on @akrhin's collapsible-groups work.
+
+### Fixed
+
+- **Non-model credential-pool keys no longer appear as phantom providers in the model picker (#4324).** A regression from #4247: the credential-pool detection loop added *every* `auth.json` → `credential_pool` key to the detected providers without checking it names a model provider. The Photon iMessage plugin writes three such keys (`photon`, `photon_project`, `photon_user`) that are messaging-platform credentials, not LLM API keys — so the picker showed three bogus "providers" (`photon`, `photon project`, `photon user`) each listing the full auto-detected model catalog. Pool detection is now gated on a new `_is_known_model_provider()` check (a `custom:*` slug, a static `_PROVIDER_DISPLAY`/`_PROVIDER_MODELS` entry, or a registered model-provider plugin) on both the `load_pool` and raw-dict fallback paths, and the group-builder's generic fallback no longer paints an unrecognized id with the global catalog. Real pool-keyed providers (`copilot`, aliased `google`→`gemini`) still surface. Reported by Nic; fix by @nesquena-hermes and independently by @rodboev (#4329/#4330).
+
 ## [v0.51.463] — 2026-06-16 — Release PX (tighter remote workspace boundaries)
 
 ### Fixed
--- a/api/config.py
+++ b/api/config.py
@@ -1012,7 +1012,11 @@ def _is_known_model_provider(provider_id: str) -> bool:
         if _is_plugin_model_provider(pid):
             return True
     except Exception:
-        logger.debug("plugin model-provider check failed for %s", pid)
+        # A transient failure here (import/IO hiccup in the plugin registry) makes
+        # a real plugin-backed provider briefly look unknown and drop from the
+        # picker until the next successful check. Surface at warning so it's
+        # visible in default production logs rather than silently swallowed.
+        logger.warning("plugin model-provider check failed for %s", pid, exc_info=True)
     return False
 
 
--- a/api/config.py
+++ b/api/config.py
@@ -981,6 +981,41 @@ def _resolve_provider_alias(name: str) -> str:
     return _PROVIDER_ALIASES.get(raw, name)
 
 
+def _is_known_model_provider(provider_id: str) -> bool:
+    """True when *provider_id* names a model provider WebUI can render.
+
+    The credential pool (``auth.json`` → ``credential_pool``) stores keys for
+    BOTH model providers (whose API keys belong in the model picker) and
+    non-model platform plugins.  The Photon iMessage plugin, for example,
+    writes ``photon`` / ``photon_project`` / ``photon_user`` pool entries that
+    are messaging-platform credentials, not LLM API keys.  Only the former
+    should surface as provider groups.
+
+    Without this gate, #4247's pool-detection loop added *every* pool key to
+    ``detected_providers``; unknown ids then fell through to the global
+    auto-detected catalog and each phantom provider was painted with the full
+    model list (#4324).  ``provider_id`` is expected to be the canonical slug
+    (post ``_resolve_provider_alias``); the lookup is case-insensitive.
+
+    A provider is "known" when it is a configured custom-provider slug
+    (``custom:*``), appears in WebUI's static ``_PROVIDER_DISPLAY`` /
+    ``_PROVIDER_MODELS`` tables, or is a registered model-provider plugin.
+    """
+    pid = (provider_id or "").strip().lower()
+    if not pid:
+        return False
+    if pid.startswith("custom:"):
+        return True
+    if pid in _PROVIDER_DISPLAY or pid in _PROVIDER_MODELS:
+        return True
+    try:
+        if _is_plugin_model_provider(pid):
+            return True
+    except Exception:
+        logger.debug("plugin model-provider check failed for %s", pid)
+    return False
+
+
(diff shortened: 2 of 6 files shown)


nesquena (approved):

Independent review of a release bundle of two coherent model-picker changes (staged for review before tagging):

1. **Collapsible provider groups** (feature, extends @akrhin's #4279) — `static/ui.js` (+195/-22), `static/style.css` (+13/-1), `tests/test_issue3691_model_picker_show_all.py` (+41/-11).
2. **#4324 phantom-providers fix** (regression from #4247) — `api/config.py` (+51/-3), `tests/test_issue4324_photon_phantom_providers.py` (+307). **This half is byte-identical to PR #4330**, which I deep-reviewed and approved.

### The #4324 half — carried forward from #4330

`git diff` of `api/config.py` + the test against the approved #4330 branch is **empty** — identical code. I'd already traced it end-to-end: the `_is_known_model_provider()` gate on both pool-detection paths (post-alias), the belt-and-braces group-builder else, no catalog starvation (proven with a real-`custom:`-provider harness), #4247 intent preserved (copilot + aliased gemini still detect), and no cross-tool surface (Photon writes `photon*` as messaging credentials; not a model provider in the agent registry). Re-verified here against a fresh hermes-agent tarball — unchanged. The bundle also folds in the CHANGELOG `[Unreleased]` entry (with @rodboev co-attribution for the independent #4329). ✓

### Collapsible groups — traced in full

The feature reworks `renderModelDropdown`'s `_filterModels` so provider groups render inside collapsible `.model-group-body` wrappers, **collapsed except the group owning the selected model**:

- **Selected-group resolution** ([ui.js:2421](static/ui.js#L2421)) reads `_modelData` (fully populated by [:2272](static/ui.js#L2272) before this IIFE runs) → `_selectedGroupKey`.
- **Per-group open default** ([ui.js:2554-2558](static/ui.js#L2554)): search → open; `_forceOpenGroups` (user-expanded) → open; endpoint-error groups (#4253) → open (so the "endpoint unreachable" warning is never buried — this was an Opus-advisor catch, and the gate is present); else open iff `groupKey === _selectedGroupKey`.
- **Search round-trip** ([ui.js:2431/2437](static/ui.js#L2431)): a fresh search force-opens all rendered groups; clearing the search (`_prevHasSearch && !hasSearch`) deletes the per-render open-state so the collapsed-except-selected default re-applies — while `_forceOpenGroups` (Show-more / manual opens) survives.
- **Cross-render persistence** rides a lazy `globalThis.__modelGroupForceOpen` Set, kept in sync by the heading-toggle handler (add on open, delete on manual collapse, [:2575](static/ui.js#L2575)).
- **In-place overflow reveal** ([ui.js:2319-2413](static/ui.js#L2319)): "Show more" appends overflow options to the source `<select>` (so it stays the source of truth), then inserts the new rows directly into the wrapper before the expander, removes the expander, flips the group's `_modelData.hiddenByDefault=false` + zeroes `hiddenCount`, force-opens the group, and scrolls to the first new row. Falls back to the proven full re-render when the runtime lacks `CSS.escape`/`insertBefore` or anything throws.
- **Keyboard nav** ([ui.js:2643](static/ui.js#L2643)): `_visibleModelRows` now spans `.model-opt,.model-opt-more` and filters out rows inside a collapsed body — so arrow-keys skip collapsed groups (#2791 preserved) and Enter on the expander triggers the reveal (`pick.click()` → `_doExpand`).

I verified the three Opus-advisor fixes the PR cites are all present in the final state: search-clear keeps the user's expansion (`_forceOpenGroups`), the endpoint-error gate, and the `#bug3` guard (`_expandOverflowGroup` bails on `extraModels.length`, **not** the append return value — which is 0 when overflow options already exist; `_appendOverflowOptionsToGroup` confirms it returns the new-option count while still clearing `dataset.extraModels`, [ui.js:2164-2170](static/ui.js#L2164)).

### Security audit

No new sink. Every dynamic value reaching `innerHTML` is `esc()`-escaped: `_buildModelRow` (name/id/badge/chip), configured rows ([:2517-2518](static/ui.js#L2517)), group rows ([:2595](static/ui.js#L2595)), and the "Show more" label ([:2608](static/ui.js#L2608)); headings, hints, scope-note and no-results use `.textContent`. The `_buildModelRow` rows are built from **raw** overflow data and escaped exactly once. The new `style.css` rules are static (chevron via CSS pseudo-elements). No `eval`/`Function`/`javascript:`/external `url()`. ✓

### Behavioural verification — independent jsdom harness (covers the un-CI'd in-place path)

The PR notes (#4331) that the #3691 node driver only exercises the full-re-render fallback, not the in-place reveal. I built a jsdom harness extracting the real `renderModelDropdown` + overflow helpers, two provider groups (OpenAI selected, OpenRouter with a `2 of 5` overflow tail) — **21/21 pass**:

```
default: openai (selected) OPEN; openrouter COLLAPSED; show-all expander present
search "or-": matching openrouter auto-expanded; non-matching openai not rendered
search "gpt": matching openai auto-expanded; non-matching openrouter not rendered
clear: openai OPEN again; openrouter COLLAPSED again
keyboard: collapsed openrouter rows NOT in the visible-row set
show-more (in place): openrouter OPEN; expander removed; all 5 models revealed in place;
                      openai untouched; <select> optgroup now has 5 options; forceOpen has openrouter
post-reveal search+clear: openrouter STILL open; NO new show-all expander; all 5 still shown
manual collapse: openrouter COLLAPSED; forceOpen no longer has openrouter
```

This confirms the headline UX and the specific snap-back regression the PR fixes (after an in-place reveal, a search round-trip must not revert the group to capped + a fresh "Show more").

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Open picker, many providers | only the selected model's group open; rest collapsed ✅ |
| Type a search | matching groups auto-expand (incl. hidden overflow models); non-matching groups not rendered ✅ |
| Clear the search | collapsed-except-selected restored; user-expanded + selected stay open ✅ |
| "Show more" on a capped group | overflow revealed in place; group stays open; others keep state; scroll to first new row ✅ |
| Overflow options already injected (e.g. selected-from-search) | append returns 0 but reveal still fires (guarded on `extraModels.length`) ✅ |
| Endpoint-error group (#4253) | stays open by default so the warning is visible; persists across search ✅ |
| Arrow-key / Enter nav | skips collapsed-group rows; Enter on expander reveals overflow ✅ |
| Minimal/headless DOM (no `CSS.escape`/`insertBefore`) | falls back to full re-render, same end state ✅ |
| Photon `photon*` pool keys | filtered (no phantom providers) — #4324, identical to approved #4330 ✅ |
| CLI reads same config/auth | webui-only picker UI + webui-only pool detection; agent unaffected ✅ |

### Tests

- `tests/test_issue3691_model_picker_show_all.py` + `test_2791_model_picker_keyboard_nav` + 5 other picker/provider suites — **82/82 pass** (the #3691 driver updated to recurse the new nested DOM and assert the de-noised chip).
- `tests/test_issue4324_photon_phantom_providers.py` — **8/8** (carried from #4330).
- **Full suite: 9123 passed / 132 skipped / 0 failures** in 187s (deselected the pre-existing darwin CRLF flake).
- **JS gates:** `node --check static/ui.js` clean; ESLint runtime guard exit 0.
- **Independent jsdom harness: 21/21** (above).
- **CI on the PR head:** browser-smoke + lint green; the 9 test shards were pending at review time — my local full suite substitutes.

- **In-place dedup key mismatch:** the reveal's `_alreadyShown` set holds rendered `.model-opt-id` `textContent` (decoded), but membership is tested with `_alreadyShown.has(esc(m.id))` (escaped) ([ui.js:2364/2368](static/ui.js#L2364)). For a model id containing an HTML-special char these wouldn't match and could double-insert a row. Model ids are slugs (no `<>&"'`), so it's a latent, effectively-unreachable cosmetic edge — not worth a churn fix. Comparing `el.textContent` to raw `m.id` would tighten it.
- **Pre-existing double-escape:** the main render loop does `esc(m.name)`/`esc(m.id)` on `_modelData` values that were already `esc()`-escaped at build time ([:2223](static/ui.js#L2223)) — harmless for normal names, but it means capped rows (double-escaped) and in-place-revealed rows (single-escaped via `_buildModelRow`) would differ for an exotic name. Pre-existing, not introduced here; flagging only because the new helper makes the asymmetry visible.
- **`_forceOpenGroups` is session-global:** a group expanded once stays sticky-open on later dropdown opens. Reasonable UX, but it never clears except on manual collapse — worth a comment if it ever needs scoping per provider/profile.
- **`#4253` reference:** the PR/issue cite #4253 for the endpoint-error-groups behavior, but #4253 is actually a cron release — an issue-number mislabel. The behavior itself (error groups stay open) is correct and traced.
- **In-place CI coverage:** tracked in #4331; my jsdom harness covers it in the interim.

### Recommendation

The collapsible-groups feature is well-structured and defensive (full-re-render fallback, try/caught DOM ops, keyboard-aware), the three advisor-flagged bugs are genuinely fixed in the final state, and the security posture is clean (single-escape on every dynamic sink). The #4324 half is the exact code I already approved as #4330. I covered the one acknowledged CI gap (in-place reveal) with an independent 21/21 jsdom harness. Ship.

Issue #1095, opened 2026-04-26
bug(chat): pasted images don't render inline — displayed as paperclip + filename instead of image preview


nesquena-hermes:
## Bug Report

**Reported by:** Beta tester @ET in Discord #beta-feedback  
**Date:** 2026-04-26

### Description

When a user pastes (cut and paste) an image directly into the WebUI chat composer, the image does not render as an inline preview. Instead, it is displayed as a paperclip icon and a filename. The agent *can* access the file (it reaches the workspace), so the upload itself works — the issue is purely in the UI rendering of the pasted image.

The user also noted a workaround: the agent was able to render the image by prefixing the filename with a paperclip character, suggesting the rendering logic may work for agent-generated output but not for user-pasted attachments in the chat bubble.

### Steps to Reproduce

1. Copy an image to clipboard (e.g., screenshot, right-click → copy image)
2. Paste the image into the WebUI chat composer
3. Send the message
4. Observe the chat bubble — the image is displayed as 📎 `image.png` instead of an inline image preview

### Expected Behavior

Pasted images should render as inline image previews in the chat bubble, similar to how drag-and-drop works.

### Actual Behavior

Pasted images are displayed as a paperclip icon + filename. No image preview is shown. The underlying file is accessible to the agent via the workspace.

### Additional Context

- Drag-and-drop file upload was confirmed to work correctly (agent can see and access the file)
- The image paste upload itself works (file reaches workspace), only rendering is broken
- Agent-side workaround exists (agent prefixes filename with paperclip to trigger render), confirming render path exists but is not triggered for user-pasted images

nesquena-hermes:
Thanks for the detailed report. The split behavior here is a useful diagnostic clue:

- **Upload works** (file reaches workspace via paste)
- **Render broken** (chat bubble shows 📎 filename instead of inline image)
- **Agent workaround exists** (agent-prefixed paperclip triggers render path)

This points to the issue being in the message rendering pipeline, not the upload handler. The likely location is where user messages with attachments are rendered — there's probably a check that decides whether to show an `<img>` element or a file-link fallback, and paste-uploaded images are hitting the fallback path while drag-and-drop images are hitting the image path.

A few things that would help narrow this down:

1. **Does drag-and-drop of an image also render inline?** (The report says yes, but confirming that paste-specific handling is the diff is useful.)
[...]

PR #4339, opened 2026-06-17
fix(composer): attach image AND paste text when clipboard carries both

### LINKED ISSUE #1620: Cmd+V always attaches an image when clipboard contains both text and image
# Cmd+V always attaches an image when clipboard contains both text and image

## Summary

Pasting into the message box (`#msg`) with Cmd+V always attaches a screenshot/image, even when the clipboard buffer is text that happens to also carry a rendered preview. The text payload is silently discarded.

## Reproduce

1. Copy text from any of: macOS Notes, Microsoft Word, most browsers, Slack, Excel cell
2. Click into the WebUI message textarea
3. Cmd+V

*Expected: text appears in the textarea.
Actual: an image (`screenshot-<timestamp>.png`) is attached to the message; no text is pasted. Status bar shows "Image pasted: …".

Plain-text-only sources (Terminal, plain `.txt` files) paste correctly. Pure screenshots (Cmd+Shift+Ctrl+4) attach correctly. The bug only fires when the clipboard contains both text and an image item — which macOS does routinely for rich-text sources (the OS attaches a rendered HTML/preview image alongside the plain text).

## Root cause

`static/boot.js` paste handler checks only for the presence of an `image/` item, then unconditionally calls `e.preventDefault()` and attaches the image:

```js
$('msg').addEventListener('paste',e=>{
  const items=Array.from(e.clipboardData?.items[]);
  const imageItems=items.filter(i=>i.type.startsWith('image/'));
  if(!imageItems.length)return;
  e.preventDefault();           // ← swallows the text payload too
  …addFiles(files);
});
```

When the clipboard has both `text/plain` and `image/png`, the handler grabs the image and `preventDefault()` cancels the browser's default text paste. The text never lands in the textarea.

## Suggested fix

Skip image-attach when the clipboard also carries text. Only intercept on image-only clipboards (true screenshot paste). Also tighten the image filter to `kind==='file'` toignore string items that advertise an image MIME.

```diff
 $('msg').addEventListener('paste',e=>{
   const items=Array.from(e.clipboardData?.items[]);
-  const imageItems=items.filter(i=>i.type.startsWith('image/'));
-  if(!imageItems.length)return;
+  // When the clipboard carries BOTH text and an image (common from Notes,
+  // Word, browsers, Slack — the OS attaches a rendered preview alongside
+  // the plain text), prefer the text and let the browser paste normally.
+  // Only intercept when the clipboard is image-only (true screenshot paste).
+  const hasText=items.some(i=>i.kind==='string'&&(i.type==='text/plain'i.type==='text/html'));
+  const imageItems=items.filter(i=>i.kind==='file'&&i.type.startsWith('image/'));
+  if(!imageItems.lengthhasText)return;
   e.preventDefault();
   const files=[imageItems.map](https://imageitems.map/)(i=>{
     const blob=i.getAsFile();
     const ext=i.type.split('/')[1]||'png';
     return new File([blob],`screenshot-${[Date.now](https://date.now/)()}.${ext}`,{type:i.type});
   });
   addFiles(files);
   setStatus(t('image_pasted')+[files.map](https://files.map/)(f=>[f.name](https://f.name/)).join(', '));
 });
```

## QA matrix (verified locally)

- Cmd+V plain text from Terminal → text appears ✓
- Cmd+V screenshot (Cmd+Shift+Ctrl+4) → image attached ✓
- Cmd+V from Notes / Word / Slack (text + preview) → text appears, no rogue image attach ✓
- Cmd+V from a browser image (Copy Image) → image attached ✓

## Environment

- macOS (clipboard behavior; Windows/Linux likely affected for the same rich-text sources)
- Chromium-based browsers (Edge, Chrome) — all hit the same `clipboardData.items` shape
- hermes-webui `master` @ 9986d2f

kaishi00:
## Summary

When the clipboard carries both text and an image — which happens almost every time you copy an image from a browser (right-click → Copy Image), Notes, Slack, or Word — the paste handler silently dropped the image and only pasted the text.

This was a regression introduced by the fix for #1620. The original bug was that Cmd+V always attached an image even when the user intended to paste text. The fix made it so that **any** clipboard containing text would skip image attachment entirely. But since almost every "copy image" action on the web also places a `text/html` or `text/plain` representation on the clipboard, the practical effect was that image paste from web sources **never worked**.

## Behavior matrix

| Clipboard contents | Before this PR | After this PR |
|---|---|---|
| Text only | ✅ Text pastes normally | ✅ Unchanged |
| Image only (screenshot) | ✅ Image attached | ✅ Unchanged |
| **Both text + image** | ❌ Text pastes, **image silently dropped** | ✅ Text pastes normally **AND** image attached |

## Changes

**`static/boot.js`** — paste event handler (~12 lines changed):

1. Extract `imageItems` first, before checking for text
2. If no image items, return early (pure text paste — browser handles it)
3. If image items exist **and** text is also present, do **not** call `preventDefault()` — let the browser insert the text naturally, then call `addFiles()` with the image(s)
4. If image items exist but **no** text (true screenshot paste), call `preventDefault()` and attach — unchanged behavior

## Test plan

- [ ] Copy an image from a web page (right-click → Copy Image), paste into composer → both text and image thumbnail should appear
- [ ] Take a screenshot to clipboard, paste into composer → image thumbnail should appear, no text
- [ ] Copy plain text from a text editor, paste into composer → text should appear normally, no image attached
- [ ] Copy text from Notes/Word/Slack (carries both text + rendered preview), paste → text should paste, **no** image should attach (these sources add a rendered preview that the user didn't intend to send as an image)

> **Note on the last case:** Sources like Notes and Word put a rendered HTML preview alongside the text, which is technically "both text and image." With this PR, that preview would get attached as an image. If maintainers prefer the conservative approach (attach image only when there's no text), the alternative is to check whether the text payload is a URL (browser "copy image" behavior) vs. actual content. Happy to adjust based on preference.

Closes #1620

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -1531,16 +1531,15 @@ document.addEventListener('keydown',async e=>{
 });
 $('msg').addEventListener('paste',e=>{
   const items=Array.from(e.clipboardData?.items||[]);
-  // When the clipboard carries BOTH text and an image (common from Notes,
-  // Word, browsers, Slack — the OS attaches a rendered preview alongside
-  // the plain text), prefer the text and let the browser paste normally.
-  // Only intercept when the clipboard is image-only (true screenshot paste).
-  // Tighten the image filter to kind==='file' so string items advertising an
-  // image MIME (e.g. text/html with an embedded data URI) are not misclassified.
-  const hasText=items.some(i=>i.kind==='string'&&(i.type==='text/plain'||i.type==='text/html'));
+  // Extract image items (kind==='file' filter avoids misclassifying text/html
+  // with embedded data URIs as images).
   const imageItems=items.filter(i=>i.kind==='file'&&i.type.startsWith('image/'));
-  if(!imageItems.length||hasText)return;
-  e.preventDefault();
+  if(!imageItems.length)return;
+  // If text is also present (common when copying images from browsers, Notes,
+  // Slack, etc.), let the browser paste the text normally AND attach the image.
+  // Only preventDefault when the clipboard is image-only (true screenshot paste).
+  const hasText=items.some(i=>i.kind==='string'&&(i.type==='text/plain'||i.type==='text/html'));
+  if(!hasText)e.preventDefault();
   const pasteTs=Date.now();
   const files=imageItems.map((i,idx)=>{
     const blob=i.getAsFile();
--- a/tests/test_1620_paste_text_with_image.py
+++ b/tests/test_1620_paste_text_with_image.py
@@ -67,15 +67,44 @@ class TestPasteHandlerTextWithImage:
             body,
         ), "imageItems filter must use kind === 'file' && type.startsWith('image/')"
 
-    def test_handler_skips_attach_when_text_present(self):
-        """The early-return guard must short-circuit when text is in the clipboard,
-        so the browser's default text-paste runs and no image is attached."""
+    def test_handler_skips_preventDefault_when_text_present(self):
+        """When text is in the clipboard alongside an image, preventDefault() must
+        NOT be called so the browser's native text-paste runs. The handler should
+        still attach the image via addFiles(). The #1620 intent (don't swallow text)
+        is preserved by gating preventDefault on !hasText instead of bailing entirely.
+        Additionally, text-only clipboards (no image items) must early-return before
+        reaching the attach path."""
         body = _paste_handler_body()
-        # Guard shape: if(!imageItems.length || hasText) return;
+        # Text-only early-out: if no image items, return before any attach logic.
         assert re.search(
-            r"if\s*\(\s*!\s*imageItems\.length\s*\|\|\s*hasText\s*\)\s*return\s*;",
+            r"if\s*\(\s*!\s*imageItems\.length\s*\)\s*return\s*;",
             body,
-        ), "guard must early-return when there are no image files OR text is present"
+        ), "handler must early-return when there are no image files (text-only paste)"
+        # preventDefault must be gated on !hasText so text+image pastes let the
+        # browser insert text natively.
+        assert re.search(
+            r"if\s*\(\s*!\s*hasText\s*\)\s*e\.preventDefault\s*\(\s*\)\s*;",
+            body,
+        ), "preventDefault must be gated on !hasText (only intercept image-only paste)"
+
+    def test_handler_attaches_image_when_both_text_and_image_present(self):
+        """When the clipboard carries both text and image items (e.g. browser
+        Copy Image), addFiles() must still be reached so the image is attached.
+        The image attach path must NOT be guarded by hasText."""
+        body = _paste_handler_body()
+        # addFiles must appear AFTER the preventDefault gate, reachable regardless
+        # of whether text is present.
+        prevent_idx = body.find("preventDefault")
+        addfiles_idx = body.find("addFiles(files)")
+        assert addfiles_idx > prevent_idx, (
+            "addFiles(files) must come after the preventDefault gate and be "
+            "reachable even when hasText is true"
+        )
+        # No guard that bails before addFiles when hasText is true.
+        assert not re.search(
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
## Review — the fix is correct and a real bug; just needs the #1620 test updated to match the new behavior

Nice catch — this is a genuine regression. When the clipboard carries **both** text and an image (copy-image-from-browser, Notes, Slack, Word all do this), the old `if(!imageItems.length || hasText) return;` guard dropped the image entirely. Your rework is logically sound:

- text-only → `imageItems` empty → early `return`, no attach (✅ **preserves the original #1620 intent** — pasting text never spuriously attaches an image)
- image-only → `preventDefault()` + attach (✅ unchanged screenshot-paste path)
- text **+** image → attach the image AND let the browser paste the text natively (✅ the fix — `preventDefault()` is correctly skipped only when `hasText`)

No double-paste risk (the attach path is independent of the native text paste), and the `kind==='file'` image filter is retained so a `text/html` data-URI isn't misclassified. Good.

### One blocker: you've changed behavior that an existing test pins

`tests/test_1620_paste_text_with_image.py::test_handler_skips_attach_when_text_present` **fails** on this branch (I ran it: `1 failed, 8 passed`). That test is a source-pattern assertion that greps for the *old* guard:

```python
assert re.search(r"if\s*\(\s*!\s*imageItems\.length\s*\|\|\s*hasText\s*\)\s*return\s*;", body)
```

Your PR deliberately removed that guard, so the assertion (and the test's premise — "skips attach when text present") is now obsolete. **A behavior change must update the test that pins the old behavior, not leave it red.**

### Fix-spec (small, well-defined)
1. In `tests/test_1620_paste_text_with_image.py`, replace `test_handler_skips_attach_when_text_present` with the new contract. The #1620 intent is still preserved but the mechanism moved — the test should now assert:
   - text-only (no image items) → handler returns before attaching (your `if(!imageItems.length)return;` early-out);
   - **and** that `preventDefault()` is gated on `!hasText` (the new `if(!hasText)e.preventDefault();`), i.e. text+image does NOT suppress the native text paste.
   Rename it to reflect the new semantics (e.g. `test_handler_skips_preventDefault_when_text_present`).
2. Add a **positive** assertion for the headline fix: when both image-items AND text are present, `addFiles(...)` is still reached (the image is attached). A source-level assertion that the attach path is no longer guarded by `hasText` is enough to lock the regression.
3. Rebase onto current master (you're ~64 behind; still MERGEABLE, just keep it fresh).

Once the #1620 test reflects the new behavior + a positive text+image assertion exists, I'll run the full gate and ship it — this is a good fix on a common path. Thanks @kaishi00!

kaishi00:
Thanks for the thorough review! I've updated the test suite:

1. **Renamed** `test_handler_skips_attach_when_text_present` → `test_handler_skips_preventDefault_when_text_present` — now asserts:
   - `if(!imageItems.length)return;` early-out for text-only clipboards
   - `if(!hasText)e.preventDefault();` gate (the new contract)

2. **Added** `test_handler_attaches_image_when_both_text_and_image_present` — positive assertion that `addFiles(files)` is reachable after the `preventDefault` gate and is NOT guarded by `hasText`

3. **Rebased** onto current master (v0.51.468)

All 21 tests pass (7 in the updated #1620 suite + 14 in the related #1095 suite).

Full run with `pytest -k "paste or image or upload"` hits 82 pre-existing collection errors (missing `yaml` dependency in the test venv — unrelated to this change) but the static-analysis tests that actually load are all green.

nesquena-hermes:
Shipped in **v0.51.472** 🎉 — pasting an image copied alongside text (browser Copy-Image, Notes, Slack) now attaches the image again while still pasting the text, matching how the major AI chat UIs handle paste. Gated clean (Opus SAFE + full suite 9313). Thanks for the fix and the thorough behavior matrix, @kaishi00 — great first contribution! 🙏

PR #4089, opened 2026-06-13
fix(streaming): normalize repair-merged user context

starship-s:
## Thinking Path

- A WebUI turn can look correct in the visible transcript while `context_messages` carries different model-facing state into the next agent call.
- Hermes Agent correctly performs a defensive role-sequence repair by joining adjacent user rows with a blank line; WebUI should not persist that repaired transport shape as if it were the human's current prompt.
- The contamination case happens at the WebUI merge boundary: a previous context-tail user row and the current submitted prompt can come back as one repaired `user` row.
- The narrow fix is to normalize that repaired current-user row back to the submitted current prompt before saving either model-facing `context_messages` or visible transcript rows.
- This intentionally leaves Hermes Agent's core repair behavior unchanged and keeps the guard constrained to the current-turn boundary, with workspace-sentinel-aware matching and negative tests for legitimate user text.

## Contract Routing

Task type: runtime/context reconstruction bug fix.

Touched areas:
- model-facing `context_messages` merge after agent results
- visible transcript merge after agent results
- sync and streaming chat result finalization paths

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/rfcs/webui-run-state-consistency-contract.md`

State layer / invariant:
- Mutates the visible transcript and model context merge boundary.
- Preserves the invariant that visible current user turns and model-facing context do not diverge due to replay/repair artifacts.
- Refs #2361.

## What Changed

- Added a repair-merge detector for user rows matching the previous context-tail user text plus the current submitted prompt.
- Normalized detected repaired rows back to the clean current user prompt before persisting `context_messages`.
- Applied the same normalization before visible transcript merging so the UI does not append a polluted duplicate user row next to a clean eager checkpoint.
- Threaded the current submitted prompt into the sync and streaming context-dedupe call sites while keeping the helper backwards-compatible for existing two-argument callers.
- Added regression coverage for prefix and non-prefix context merge shapes, display eager checkpoints, workspace sentinels, list-content payloads, and negative non-boundary cases.

## Why It Matters

- Prevents a prior user message from being prepended to a later prompt in future agent context.
- Keeps WebUI's visible transcript and model-facing context aligned at the turn boundary.
- Avoids changing provider-protocol repair behavior in Hermes Agent, so the fix stays local to WebUI's persistence/merge responsibility.
- Makes the bug reproducible and guarded with targeted tests instead of relying on manual transcript inspection.

## Verification

[...]

PR #4341, opened 2026-06-17
fix(streaming): clean cancelled user context chains

starship-s:
## Thinking Path

- WebUI keeps separate visible transcript and model-facing `context_messages` layers, and those layers must not contradict the user turn that was actually submitted.
- This builds on the earlier stale-user cleanup in #4089, which normalized the simple one-hop repair shape. Cancelled or interrupted correction turns can leave longer or already-replayed stale user chains that Agent later merges with the next clean user turn for provider role alternation.
- Agent's role-sequence repair should continue preserving all user input; the safer fix is at the WebUI merge/writeback boundary, where WebUI can recognize when a returned user row is a repaired stale chain ending in the current submitted turn.
- The detector now validates stale prefixes against prior user rows from `previous_context`, including replayed prefixes from older polluted rows, so only current-turn candidate rows are normalized and committed historical rows stay intact.
- The boundary finder matches the full submitted turn as the suffix, so multi-paragraph current prompts are not mistaken for stale-chain segments.

## What Changed

- Extended stale-user merge detection to support multi-hop repaired user chains, not only the immediate previous-user-tail case.
- Added ordered prior-user-context validation for stale prefixes replayed from older polluted rows after newer clean turns have advanced the context tail.
- Threaded `previous_context` through the model-context and display-merge cleanup paths so both `context_messages` and visible `messages` use the same normalization rule.
- Kept the normalization scoped to new-turn candidate/current rows; existing committed context rows are not rewritten.
- Added regression tests for one-hop compatibility, multi-hop stale chains, replayed polluted prefixes, eager-checkpoint display dedupe, historical-row preservation, and multi-paragraph current-turn false-positive/true-positive cases.

## Why It Matters

- Prevents cancelled or interrupted correction prompts from reappearing as prefixes on later unrelated WebUI turns.
- Keeps the user's submitted turn as the durable current user message in both visible transcript and model-facing context.
- Preserves the Agent core repair behavior while preventing WebUI from accepting repaired stale chains as durable user intent.
- Reduces recurrence of adjacent clean + polluted user rows after eager session-save/checkpointing.

## Contract Routing

Task type: runtime/session-state bug fix.

Touched areas:
- `api/streaming.py` model-context and display transcript merge/writeback helpers.
- `tests/test_stale_user_context_contamination.py` product-semantics regression coverage.

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`

Invariant evidence:
- Visible current turns and model-facing context stay aligned after stale repair cleanup.
- Replay/display merge remains idempotent for eager-checkpointed current user rows.
- Historical context rows shaped like stale merges are preserved.

## Verification

Targeted coverage:

```text
./scripts/test.sh tests/test_stale_user_context_contamination.py -q
# 34 passed

./scripts/test.sh tests/test_issue1361_cancel_data_loss.py tests/test_issue3468_duplicate_after_compression.py -q
# 25 passed
```

Syntax / hygiene checks:

```text
python -m py_compile api/streaming.py
git diff --check
# passed
```

CI:

```text
browser-smoke: pass
lint: pass
test (3.11, 0/1/2): pass
test (3.12, 0/1/2): pass
test (3.13, 0/1/2): pass
```

## Risks / Follow-ups

- This remains a boundary normalization fix rather than a broader cancellation-state rewrite; future work could further harden cancelled-turn finalization so malformed adjacent-user context is less likely to enter the merge path at all.
- No UI/layout changes; screenshots are not applicable.

## Model Used

- OpenAI / GPT-5.5 for investigation, implementation orchestration, and verification.
- OpenCode / GPT-5.3 Codex Spark for the initial implementation pass.
- Anthropic / Claude Opus via Claude Code for independent review of the stale-merge detector and follow-up fixes.

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -3846,9 +3846,14 @@ def _dedupe_replayed_context_messages(previous_context, result_messages, msg_tex
         ):
             boundary_idx = len(previous_context) - 1
             boundary_row = result_messages[boundary_idx]
-            is_stale_merge = (
+            is_stale_merge = bool(
                 previous_user_tail
-                and _detect_stale_user_merge(boundary_row, msg_text, previous_user_tail)
+                and _detect_stale_user_merge(
+                    boundary_row,
+                    msg_text,
+                    previous_user_tail,
+                    previous_context=previous_context,
+                )
             )
             if is_stale_merge or _looks_like_current_user_turn(boundary_row, msg_text):
                 if is_stale_merge:
@@ -3873,6 +3878,7 @@ def _dedupe_replayed_context_messages(previous_context, result_messages, msg_tex
             candidates,
             msg_text,
             previous_user_tail,
+            previous_context=previous_context,
         )
     candidates = _strip_replayed_prefix(previous_context, candidates)
     if candidates:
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.473** 🎉 — cancelled/interrupted correction turns no longer leak a stale prefix onto later messages. Gated clean (Codex SAFE + Opus SAFE, both adversarial; full suite 9327). The suffix-equals-entire-submitted-turn invariant cleanly prevents any false-positive on multi-paragraph / shared-prefix / repeated turns. Thanks for the careful work and thorough test coverage, @starship-s! 🙏

nesquena-hermes:
Reviewed the diff against `origin/master` — `api/streaming.py` (+154) plus 359 lines of new coverage in `tests/test_stale_user_context_contamination.py`. All 11 CI legs (browser-smoke, lint, 3.11/3.12/3.13 × 0/1/2) are green. This is a clean extension of the #4089 one-hop repair, and the boundary-finder design is right.

## What the change does

The detector now searches for a `\n\n` boundary whose *suffix* normalizes to the entire submitted turn, then treats only the prefix as stale — `api/streaming.py:4080`:

```python
suffix = merged[boundary_idx + 2:]
if _normalize_user_text(suffix) == current_norm:
    prefix = merged[:boundary_idx]
    candidate_segments = [_normalize_user_text(s) for s in prefix.split("\n\n")]
    if candidate_segments and all(candidate_segments):
        stale_segments = candidate_segments
```

This correctly avoids the false positive where the current turn is itself multi-paragraph: for a clean `A\n\nB` the only boundary whose suffix equals the whole turn leaves an empty prefix, so `stale_segments` stays empty and the detector returns `False`. The `test_merge_display_does_not_overstrip_when_current_already_clean` and `..._passes_through_when_prior_tail_text_differs` cases lock that in.

## Scoping looks correct

The permissive new matching is bounded because both call sites only feed the *new-turn candidate slice* into `_strip_stale_user_merge_from_messages`, never the committed `previous_context` prefix — `api/streaming.py:3873` and `:4500`. `test_dedupe_replayed_context_preserves_historical_row_with_merge_shape` confirms a legitimate old row shaped like `<tail>\n\n<turn>` survives, which is the main regression risk.

## One observation on the substring fallback

The second branch of `_stale_prefix_matches_prior_user_context` (`api/streaming.py:4063`) accepts stale segments found as *substrings* within older merged rows via `row.find(segment, row_offset)`, not just exact prior-row matches. That's necessary for the "replayed prefix from an already-polluted older row" shape, but it's the most permissive comparison in the change. A theoretical false-positive needs a boundary row of exactly `<prefix>\n\n<entire current turn>` where `<prefix>`'s paragraphs each appear as ordered substrings of prior user rows — narrow, but worth a guarding test: a current turn whose first paragraph coincidentally matches a substring of an old row, with a genuinely different second paragraph, asserting it is **not** stripped. The suffix-equals-whole-turn requirement should already protect this, so the test would just pin the invariant.

Not a blocker — the contract (visible turn == model-facing context after repair, idempotent on eager checkpoints, historical rows preserved) is well covered. The substring branch is the one spot I'd add belt-and-suspenders coverage before this becomes load-bearing for multi-hop contamination cleanup.

Issue #4277, opened 2026-06-15
bug(chat): transcript viewport blanks after replies until Newest Message forces re-render


santastabber:
## Summary

In long or tool-heavy chats, the chat transcript viewport can render as a blank/empty area after an assistant reply settles. The transcript data still exists, and clicking **Newest Message** forces visible content to appear again. Scrolling upward can then make the viewport blank again.

This appears to be a frontend transcript rendering / scroll-windowing problem rather than a backend streaming or session-persistence failure.

## Reproduction

A reliable synthetic reproduction should be possible with a session containing:

1. A long assistant turn with many streamed/progress/tool/worklog segments.
2. Enough visible transcript rows to trigger transcript windowing/virtualization.
3. The reader scrolled away from the bottom before or as the assistant turn settles.
4. A final `renderMessages({ preserveScroll: true })` / stream-completion DOM replacement.
5. Scroll upward after the transcript repaints.

## Actual behavior

- After the assistant reply settles, the visible message pane can become blank/empty.
- The transcript/session data is still present.
- Clicking **Newest Message** forces visible transcript content to render again.
- Scrolling upward can blank the viewport again.

## Expected behavior

The message pane should never restore or virtualize to a viewport region with no visible transcript content while the session has visible messages. Scroll preservation/windowing should keep the current semantic message/turn visible, or safely fall back to a larger/full render window when geometry is uncertain.

## Suspected cause

This looks related to transcript windowing/virtualization and scroll restoration around these frontend paths:

- `static/ui.js`: `MESSAGE_RENDER_WINDOW_DEFAULT`, `MESSAGE_VIRTUAL_THRESHOLD_ROWS`, `MESSAGE_VIRTUAL_BUFFER_PX`
- `static/ui.js`: `_messageVirtualWindow(...)`
- `static/ui.js`: `renderMessages(...)`, which rebuilds `#msgInner`, inserts virtual spacers, and restores scroll from prior geometry

A likely failure mode is that pixel/spacer-based restoration lands the viewport in a region represented by spacer or hidden/non-visible segment geometry rather than visible message rows. Large Worklog/progress content and hidden source segments may make the measured geometry diverge from what is actually visible to the reader.

## Related but distinct issue

Possibly related to #4207, but the user-visible failure differs:

- #4207: completion scroll jump / `scrollToBottom()` overriding `preserveScroll`
- this issue: the viewport can become blank until **Newest Message** forces a new scroll/window position

## Implementation options

[...]

PR #4345, opened 2026-06-17
fix(#4343): default virtualize_transcript OFF and label experimental

### LINKED ISSUE #4343: bug(transcript): virtualization causes scroll-up flicker on long sessions — default it OFF (experimental/opt-in)
## Summary

Transcript virtualization (#500 / #4214, opt-out toggle added in #4325) still causes a **scroll-up flicker** on long sessions in the current release. Scrolling up through a long transcript makes the viewport "flick back and forth." Turning virtualization **off** (Settings → Preferences → "Virtualize long transcripts") eliminates it completely.

**Maintainer decision (Nathan, 2026-06-17):** flip `virtualize_transcript` to **default OFF** for everyone now, label it experimental, and make it opt-IN going forward — until the virtualization scroll behavior can be made reliable. (Separate from the eventual root-cause fix below.)

## Reporter & repro

**@b3nw** — Discord #report-bugs, 2026-06-17. Verbatim:

> basically have a large number of tool calls (90~) and then a forced stop, when you scroll up, it flicks back and forth, total session is 565 messages

Confirmed after maintainer asked him to A/B the latest build:

> WebUI: v0.51.469 / Agent: 0.16.0 ... the issue reproduces with virtualizing on, no issue scrolling with it off.

Shared session for repro: https://ping.ext.ben.io/share/9JGmaTCy (565 messages, ~90 tool calls + a forced stop).

**Repro shape:** open a session with hundreds of messages where many rows are tall (large tool-call blocks), scroll up. Window: virtualization engages above 80 renderable rows (`MESSAGE_VIRTUAL_THRESHOLD_ROWS`).

## Root cause (analysis)

This is the classic **variable-height virtual-scroll anchor oscillation**, distinct from the stream-completion jump that #4325 addressed.

The virtual window is computed from a per-row height cache where any **unmeasured** row falls back to a fixed estimate (`MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHT`) — `static/ui.js` `_messageVirtualWindow()` (`rowHeightFor` at ~ui.js:459-462, `topPad`/`bottomPad` accumulation 471-494). On a transcript with ~90 large tool-call rows, the real heights are wildly larger than the default estimate, so:

1. Scroll up → `_scheduleMessageVirtualizedRender` (ui.js:755) recomputes the window from **estimated** heights and re-renders the newly-visible head rows.
2. The head rows are then measured at their **actual** (much taller) heights → `_scheduleMessageVirtualMeasurementRefresh` (ui.js:526) schedules another render with corrected `topPad`/`bottomPad`.
3. The corrected padding changes `scrollHeight`, which shifts `scrollTop`, which moves the window again → another render → another measurement. That round-trip is the visible "flick back and forth."

The existing guards bound but don't eliminate it:
- `MESSAGE_VIRTUAL_MEASUREMENT_MAX_RERENDERS` caps the measurement re-render loop (ui.js:532) — it stops runaway loops but the user still sees the oscillation settle visibly.
- `_maybeRecoverVirtualizedBlankViewport` (ui.js:9789) fires a *full* re-render when the viewport ends up blank after a scroll — itself a visible jump when heights are mis-estimated.

So the bug is structural to estimate-then-measure virtualization with high height variance, not a single off-by-one.

## Suggested fix (fix shape)

**Phase A — ship now (the decision):** flip the default to off + mark experimental.
- `api/config.py:6878` — `"virtualize_transcript": True` → `False` (it's already in `DEFAULT_SETTINGS` and the allowlist at :7086, so only the default value changes).
- `static/panels.js:6770` reads `settings.virtualize_transcript!==false` (defaults ON when unset) — must change the unset default to OFF to match (e.g. `===true`), or the server default alone won't flip existing clients with no stored value. Verify both the server default AND the client-side fallback agree on OFF.
- Update the Settings checkbox label/help text (the "Virtualize long transcripts ... for smoother scrolling" copy) to mark it **experimental / opt-in** and note it can cause scroll flicker on very long sessions.
- CHANGELOG `Changed` entry. Locale parity for any new/edited string across all locales (key-parity tests).

**Phase B — eventual root-cause fix (keep open after A):** make the virtualization scroll-stable so it can be re-defaulted-on later. Standard approaches: anchor-based scroll compensation (pin to a stable anchor row and adjust `scrollTop` by the measured-vs-estimated delta on each measurement cycle so the anchor stays put), measure-before-paint for the about-to-be-visible window, or `content-visibility: auto` / `contain-intrinsic-size` per row instead of manual spacer math. Tune `MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHT` toward observed tool-call row heights to shrink the estimate error.

## Scope

`nesquena/hermes-webui` only. Phase A is small (~5-15 LOC + label/CHANGELOG/locales) and ship-ready given the decision. Phase B is a focused front-end virtualization task in `static/ui.js`.

## Related

- #4325 (CLOSED) — stream-completion freeze/jump from the same virtualization feature; "fixed" by adding the opt-out toggle (default ON). This issue is the **residual scroll-up flicker** that the toggle exposes, plus the maintainer decision to invert the default.
- #4214 (MERGED) / #500 — the virtualization feature itself.

## Reported by

@b3nw — Discord #report-bugs, 2026-06-17. Confirmed legitimate; A/B-verified (on = flicker, off = clean) by the reporter on v0.51.469.

rodboev:
## Thinking Path

- Transcript virtualization (#500 / #4214) shipped default-on with an opt-out toggle (#4325, v0.51.469), but scrolling up through long sessions with many tall tool-call rows causes visible back-and-forth flicker because the virtual window's per-row height estimates diverge wildly from measured heights on large tool-call blocks.
- The existing viewport anchor restore (#4277, v0.51.448) and stream-end dedup (#4328) addressed related symptoms but not the structural estimate-then-measure oscillation that @b3nw confirmed with an A/B test (on = flicker, off = clean).
- The maintainer's decision is to default the feature OFF now and label it experimental/opt-in, giving existing users a clean scroll experience while the underlying virtualization is stabilized separately.

## What Changed

- `api/config.py`: flipped `virtualize_transcript` default from `True` to `False`
- `static/boot.js`: changed settings-load path from `!==false` (default-on) to `===true` (default-off), and settings-load-failed fallback from `true` to `false`
- `static/panels.js`: changed checkbox initialization from `!==false` to `===true`
- `static/index.html`: removed `checked` attribute from the virtualization toggle checkbox, updated fallback label and description text
- `static/i18n.js`: updated label to include "(experimental)" and description to note scroll-flicker caveat across all 13 locales
- `tests/test_issue4325_virtualization_toggle.py`: updated assertions to match default-OFF semantics

## Why It Matters

Users with long tool-heavy sessions (hundreds of messages, ~90 tool calls) see visible scroll-up flicker that makes the UI feel unstable. Defaulting the feature off removes the flicker for everyone while keeping the toggle available for users who want to opt in and test the virtualized path.

## Verification

```
pytest tests/test_issue4325_virtualization_toggle.py -v --timeout=60
```

## Risks / Follow-ups

- Users who never changed the setting will lose virtualization on upgrade; they can re-enable it in Settings. This is the intended behavior per the maintainer decision.
- The root-cause fix (anchor-based scroll compensation for variable-height rows) is tracked separately and would allow re-defaulting virtualization ON in a future release.

## Upstream

Closes #4343.

## Model Used


rodboev:
Withdrawing since #4344 (v0.51.470) already landed the same fix with a more thorough migration mechanism (opt-in marker to preserve explicit user choice across the default flip).

Issue #4313, opened 2026-06-16
bug(chat): Frontend EventSource lifecycle creates noisy net::ERR_ABORTED console errors during normal WebUI usage


markwang2658:
## Summary

During normal WebUI usage, the frontend can generate browser-console `net::ERR_ABORTED` errors even though the app still appears to work.

This is not a backend endpoint failure. The noise comes from frontend-managed request / EventSource lifecycle handling, especially around `/api/session/stream`.

## Reproduction

1. Open Hermes WebUI in the browser.
2. Create a normal WebUI conversation.
3. Send a simple message such as `hello`.
4. Open DevTools console.
5. Switch sessions, or let the page go through normal stream replacement / reconnect behavior.

## Actual behavior

The browser console may show errors such as:

- `net::ERR_ABORTED http://127.0.0.1:8787/api/session/stream?...`
- sometimes related SSE / refresh noise around normal frontend lifecycle transitions

The application can still appear healthy, which makes this easy to miss unless DevTools is open.

## Expected behavior

Normal frontend lifecycle transitions should not generate browser-console errors when the application is functioning normally.

In particular:

- ordinary WebUI conversations should not open unnecessary SSE channels
- normal stream replacement / reconnect handling should not produce browser-visible `ERR_ABORTED` noise

## Root cause summary

From local investigation, the main issue appears to be frontend lifecycle behavior rather than backend failure:

1. Ordinary WebUI sessions were opening `/api/session/stream` even when that channel was only needed for external sessions.
2. Some EventSource reconnect / teardown paths explicitly closed streams during normal lifecycle transitions.
3. Chromium reports these client-side closes as `net::ERR_ABORTED`, which pollutes the console even though the app may continue working.

## Suggested direction

A possible fix direction would be:

1. Do not open `/api/session/stream` for ordinary WebUI sessions.
2. Restrict that channel to external-session cases only.
3. Avoid unnecessary explicit `close()` calls on already-closed or permanently failed EventSource objects.
4. Avoid frontend teardown / reconnect patterns that knowingly create browser-level abort noise unless they are strictly required.

## Relevant frontend files

- `hermes-webui/static/workspace.js`
- `hermes-webui/static/sessions.js`
- `hermes-webui/static/messages.js`
- `hermes-webui/static/panels.js`

## Why this matters

This looks like a frontend quality issue rather than a functional outage, but it still matters because:

- it produces misleading console errors during normal usage
- it makes real network / backend issues harder to distinguish
[...]

PR #3109, opened 2026-05-29
fix: stabilize WebUI refresh, scroll, and draft persistence

andrewkangkr:
## Summary
- reduce active-session/session-list refresh churn that caused visible flicker and scroll-to-bottom behavior
- remove `Connection: close` from WebUI SSE endpoints to avoid EventSource reconnect loops
- harden queue persistence across refresh/tab detours and preserve draft-only zero-message sessions on restore
- render `Jump to question` buttons whenever an assistant block still has a valid user-question anchor

### Refresh / scroll churn
- increased `_streamingPollMs` from `5000` to `15000`
- increased `_activeSessionExternalRefreshMs` from `5000` to `20000`
- removed forced `Connection: close` headers from SSE endpoints in `api/routes.py`

### Queue / draft persistence
- introduced shared queue persistence helpers that mirror queue state to both `sessionStorage` and `localStorage`
- updated restore flow in `static/sessions.js` to read the shared durable queue path
- preserved zero-message restored sessions when `composer_draft` still contains unsent text/files

### Jump to question reliability
- relaxed the footer render predicate so the button appears for any settled assistant block with a valid question anchor, not only the final assistant segment of a turn

## User impact
This addresses the main UX regressions reported during real use:
- chat view flickering / refreshing every few seconds
- viewport being pulled toward the bottom while reading
- intermittent missing `to question` buttons
- queued or draft work disappearing after refresh / tab switching

## Test plan
- `pytest -q tests/test_issue660.py tests/test_issue2246_question_jump.py tests/test_queue_combine_persistence.py tests/test_issue1360_streaming_scroll_hardening.py tests/test_session_batch_select.py tests/test_session_events.py tests/test_approval_sse.py tests/test_issue_active_session_refresh_regressions.py`

## Related issues
- Closes #3103
- Refs #3107
- Refs #3108

PR #4348, opened 2026-06-17
fix: reconcile WebUI session detail source flags

dso2ng:
## Thinking Path

- Hermes WebUI treats imported CLI / messaging transcripts differently from WebUI-native transcripts because external sessions may need periodic reconciliation.
- Recent upstream work narrowed the 30-second poll path (#3916) and later allowed real session events to refresh active WebUI transcripts (#4208), while separate scroll fixes addressed transcript virtualization and stream-finalization jumps (#4325/#4343).
- A remaining edge case is source metadata drift: a WebUI-origin session can have a stale sidecar that still says `raw_source=cli` / `is_cli_session=true`, while the state.db/sidebar row correctly says the session is WebUI-origin.
- If `/api/session` returns the stale sidecar flags, the browser opens the active WebUI transcript as an external session and enters the external-refresh reload path, which can rebuild the transcript while someone is reading.
- This PR keeps `/api/session` source classification consistent with the sidebar/state.db source of truth, and adds a small frontend guard so explicit WebUI-source payloads never count as external.

## What Changed

- Added `_reconcile_session_detail_source_flags()` for `/api/session` payloads whose matching state.db row is explicitly WebUI-origin.
- The detail response now clears stale `is_cli_session` / `read_only` flags and restores WebUI source fields from state.db before redaction.
- `_isExternalSession()` now refuses to classify an explicit WebUI-source session as external, even if a stale payload still has `is_cli_session=true`.
- Added regression coverage for stale CLI-looking detail payloads and for the frontend external-session guard.
- Added an Unreleased changelog note.

## Why It Matters

This keeps the list endpoint and detail endpoint on the same source-classification contract. A WebUI-native transcript should not be periodically force-reloaded like an imported/external transcript just because a stale sidecar carried old CLI markers. Avoiding that misclassification prevents unnecessary `loadSession(..., {force:true})` calls that can disrupt the reader's scroll position.

## Contract Routing

Task type: runtime/session metadata consistency fix.
Touched areas: `/api/session` detail payload source flags, sidebar/session source helpers, active external refresh frontend guard.
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
Scope boundaries:
- Does not redesign active transcript refresh policy.
- Does not change imported CLI/messaging session behavior.
- Does not touch transcript virtualization or stream-finalization scroll anchoring.
Evidence needed before claiming done: regression tests for stale WebUI sidecar source flags plus existing external-refresh source guard tests.

## Verification

- `./scripts/test.sh tests/test_issue2351_cli_session_source_filter.py::test_webui_state_db_source_overrides_stale_cli_detail_payload -q` -> fails before the backend helper exists, passes after the fix.
- `python -m py_compile api/routes.py`
- `./scripts/test.sh tests/test_issue2351_cli_session_source_filter.py tests/test_webui_external_refresh_frontend.py tests/test_issue3916_external_refresh_poll.py tests/test_session_import_cli_fallback_model.py -q` -> `37 passed`
- `git diff --check`
- public diff hygiene: `non_ascii_added_lines=0`

## Risks / Follow-ups

- This is intentionally narrow: it fixes incorrect source classification, not the broader UX policy of when active transcripts should refresh after real external updates.
- If a future migration creates WebUI-origin read-only detail payloads intentionally, the reconciliation helper may need a more specific read-only exception. Today the WebUI-origin state row is the stronger source for this classification.
- Adjacent open work exists around in-progress viewport jumps (#4295), EventSource lifecycle noise (#4313), and broader refresh/draft persistence (#3109), but this PR is a different backend/source-contract slice.

## Model Used

OpenAI `gpt-5.5` via Hermes Agent, with local terminal/file tools and GitHub CLI for repository inspection and PR preparation.


nesquena-hermes:
Reviewed against `origin/master`: `api/routes.py` (+44), `static/sessions.js` (+15), plus regression coverage and a changelog note. All 11 CI legs are green. The fix is correctly scoped and lands the detail endpoint on the same source-of-truth contract the sidebar already uses.

## The core fix is right

The new branch makes `/api/session` defer to the state.db row for WebUI-origin sessions, taking priority over the messaging-merge path — `api/routes.py:7247`:

```python
if cli_meta and _session_source_is_webui(cli_meta):
    raw = _reconcile_session_detail_source_flags(raw, cli_meta)
elif cli_meta and _is_messaging_session_record(cli_meta):
    raw = _merge_cli_sidebar_metadata(raw, cli_meta)
```

These two branches are genuinely mutually exclusive — `_session_source_is_webui` matches `source ∈ {webui}` (`api/routes.py:4623`) while `_is_messaging_session_record` matches `messaging`/known-messaging sources (`api/routes.py:4044`), so promoting the WebUI check ahead of the `elif` can't steal a row that the messaging merge actually needs. Good ordering.

The frontend guard is the necessary other half — `static/sessions.js:1437`:

```js
function _isExternalSession(session) {
  if (!session || _isWebUiSourceSession(session)) return false;
  return !!(session.is_cli_session || _isMessagingSession(session));
}
```

Without this, even a reconciled payload could still be misread if any single stale flag leaked through, so blocking at the classifier is the right belt-and-suspenders move. `test_webui_source_never_counts_as_external_session` pins both the helper and the early-return.

## One thing worth a sentence in the PR

`_reconcile_session_detail_source_flags` clears `read_only=False` unconditionally for any WebUI-source row (`api/routes.py:4592`). Your Risks section already flags this, and `test_real_cli_source_survives_detail_source_reconcile` proves a true CLI row is untouched, so the blast radius is only WebUI-origin sessions. The only scenario this would mask is a *deliberately* read-only WebUI-origin payload — which doesn't exist today, so this is correct now. If a future migration ever introduces one, the follow-up is a `source==webui AND not intentionally_read_only` guard rather than the blanket clear. Fine to defer; just calling it out so the assumption is documented at the call site, not only in the PR body.

## Verification

The targeted test `test_webui_state_db_source_overrides_stale_cli_detail_payload` exercises exactly the reported drift (sidecar says `cli`, state.db says `webui`) and asserts `is_cli_session`/`read_only` flip to False plus the count takes the `max(24, 26) → 26`. That, plus the count/timestamp `max()` merges using `_numeric_count`, covers the contract. No agent-side change needed — this is purely a WebUI detail/list consistency fix, and the agent's state.db row stays the source of truth as intended.

Looks merge-ready to me pending the maintainer's call on the read-only note.

nesquena-hermes:
Shipped in **v0.51.474** 🎉 — a WebUI-native session with a stale CLI sidecar is no longer misclassified as external and force-reloaded mid-read. Reconciled against state.db (source of truth) with the frontend guard. Gated clean (Codex SAFE + Opus SAFE, both adversarial; full suite 9330). Verified it only fires for explicitly-WebUI-origin rows — genuine imported CLI/messaging sessions are untouched. Thanks @dso2ng! 🙏

PR #4361, opened 2026-06-17
Make launcher health probes TLS-aware (https + self-signed support)

tomas-forgac:
Launcher health / "already running" probes hardcoded `http://`, so with TLS enabled (`HERMES_WEBUI_TLS_CERT` + `HERMES_WEBUI_TLS_KEY`, which makes `server.py` serve https) they probed the wrong scheme and errored. Switching to `https://` alone isn't enough: `curl`/`wget`/`urllib` reject self-signed certs by default — common for home use.

## Approach
- **Scheme detection** mirrors the server: `https` iff both TLS cert and key are set.
- **Self-signed handling**: probe with verification first; on a cert-verification failure, retry once **without** verification and warn. CA-signed certs still validate normally. This is a loopback health check, not a trust decision for browser clients.
- **Escape hatch**: `HERMES_WEBUI_TLS_INSECURE_PROBE=1` skips verification up front, quietly.

## Changes
- **`bootstrap.py`** — `wait_for_health` does verified→unverified urllib fallback (warns once); detached-path health URL and printed "open this URL" hints use the resolved scheme. Added `_tls_enabled` / `_health_scheme` / `_insecure_probe_requested` / `warn` helpers.
- **`start.sh`** — scheme-aware pre-flight probe (curl/wget) with self-signed retry + warning; "already running" message uses the scheme.
- **`ctl.sh`** — `_health_line` (used by `status`) scheme-aware with self-signed retry; `status` now loads `.env` so TLS vars are visible. Kept bash 3.2 / `set -u` safe.
- **`scripts/wsl/hermes_webui_autostart.sh`** — scheme-aware default health URL + `webui_healthy` fallback.
- **`Dockerfile`** — `HEALTHCHECK` picks scheme from the TLS env vars and adds `-k` on cert failure.
- **Tests / docs** — `tests/test_tls_health_probe.py` (scheme helpers, `wait_for_health` self-signed fallback against a real https server, `ctl.sh _health_line` over self-signed https); `docs/remote-access.md` HTTPS/TLS section; `CHANGELOG.md` entry.

`start.ps1` and the `bootstrap.py` foreground/supervisor path run no probe (only cosmetic URL strings), so no functional change there.

The retry only emits a warning when the insecure probe succeeds where the verified one failed — precisely the self-signed case:

```bash
# start.sh / ctl.sh, https endpoint
curl -fsS --max-time 2 "$url"            # verified first
# on failure, for https only:
curl -fsS -k --max-time 2 "$url"         # → warn: cert not verified (loopback only)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@
 
 - **The conversation-outline toggle button no longer overlaps the scroll-to-bottom control (#2124).** The outline FAB was `position:fixed` at a high z-index and could stack on top of the scroll-to-bottom button in the bottom-right corner. It's now anchored inside the messages shell (`position:absolute`, lower z-index) so the two controls sit cleanly without colliding. Thanks @Habib1001-m.
 
+## [v0.51.491] — 2026-06-18 — Release RA (rollback restore writes contained to the workspace)
+
+### Security
+
+- **Rollback checkpoint restore writes can no longer be redirected outside the selected workspace by a symlink planted inside it (#4405).** The restore path previously used pathname-based `shutil.copy2()` writes, which would follow a symlink whose target escapes the workspace. Restore writes now go through the existing workspace-anchored file helpers, so a restored file always lands inside the workspace tree regardless of in-tree symlinks. Thanks @hinotoi-agent.
+
 ## [v0.51.490] — 2026-06-18 — Release QZ (large plain-text pastes become .md attachments)
 
 ### Added
--- a/api/rollback.py
+++ b/api/rollback.py
@@ -7,13 +7,12 @@ created by the Hermes agent's CheckpointManager.  Checkpoints live at
 """
 
 import hashlib
-import json
 import logging
 import os
 import re
 import shutil
 import subprocess
-from datetime import datetime, timezone
+from datetime import datetime
 from pathlib import Path
 from typing import Any
 
@@ -237,7 +236,7 @@ def get_checkpoint_diff(workspace: str, checkpoint: str) -> dict[str, Any]:
             # File exists in checkpoint but not in workspace (deleted)
             files_changed.append({"file": rel_path, "status": "deleted"})
             diff_lines.append(f"--- a/{rel_path}")
-            diff_lines.append(f"+++ /dev/null")
+            diff_lines.append("+++ /dev/null")
             diff_lines.append("@@ -1,{lines} +0,0 @@".format(lines=len(ckpt_content.splitlines())))
             for line in ckpt_content.splitlines():
                 diff_lines.append(f"-{line}")
@@ -263,6 +262,31 @@ def get_checkpoint_diff(workspace: str, checkpoint: str) -> dict[str, Any]:
     }
 
 
+def _restore_checkpoint_file(workspace_root: Path, ckpt_file: Path, rel_path: str) -> None:
+    """Restore one checkpoint file without following workspace symlinks outward."""
+    from api.workspace import open_anchored_create_fd, open_anchored_write_fd, safe_resolve_ws
+
+    target = safe_resolve_ws(workspace_root, rel_path)
+    if target.exists():
+        fd = open_anchored_write_fd(workspace_root, target)
+    else:
+        fd = open_anchored_create_fd(workspace_root, target)
+
+    src_stat = ckpt_file.stat()
+    with os.fdopen(fd, "wb") as out:
+        with ckpt_file.open("rb") as src:
+            shutil.copyfileobj(src, out)
+        out.flush()
+        try:
+            os.fchmod(out.fileno(), src_stat.st_mode & 0o777)
+        except (AttributeError, OSError):
+            logger.debug("Failed to apply restored mode to %s", target, exc_info=True)
+        try:
+            os.utime(out.fileno(), ns=(src_stat.st_atime_ns, src_stat.st_mtime_ns))
+        except OSError:
+            logger.debug("Failed to apply restored timestamp to %s", target, exc_info=True)
+
+
 def restore_checkpoint(workspace: str, checkpoint: str) -> dict[str, Any]:
     """Restore a checkpoint by copying files back to the workspace.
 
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Read the full diff against `origin/master` in a read-only worktree and cross-checked the scheme logic against `server.py` plus `api/config.py`. This is a clean, well-scoped fix — the bug is real and all five probe paths are handled consistently. A few notes, one of which is worth a line of code.

## Scheme detection correctly mirrors the server

`_tls_enabled()` reads both env vars and returns true only when both are set (`bootstrap.py:89-97`):

```python
cert = os.getenv("HERMES_WEBUI_TLS_CERT", "").strip()
key = os.getenv("HERMES_WEBUI_TLS_KEY", "").strip()
return bool(cert and key)
```

That's an exact match for the server's gate (`api/config.py:67-69`): TLS_CERT/TLS_KEY are `getenv(...).strip() or None` and `TLS_ENABLED = TLS_CERT is not None and TLS_KEY is not None`. Good — the launcher and server can't disagree on scheme under normal config.

## Cert-error detection is robust

The verified-to-unverified retry hinges on correctly classifying the failure, and `_is_cert_verification_error` handles both the direct exception and the urllib wrapping (`bootstrap.py:114-119`):

```python
if isinstance(exc, ssl.SSLCertVerificationError):
    return True
reason = getattr(exc, "reason", None)
return isinstance(reason, ssl.SSLCertVerificationError)
```

`urllib.request.urlopen` raises URLError with the SSLCertVerificationError tucked into its reason attribute, so catching both forms is correct. And `_unverified_ssl_context` uses the public `create_default_context()` plus CERT_NONE rather than the private unverified-context helper — nice.

## Confirming Greptile's two non-blocking items (both real)

1. **Explicit-insecure warning inconsistency.** When `HERMES_WEBUI_TLS_INSECURE_PROBE=1` is set up front, `bootstrap.py:346-352` emits a `warn(...)`, but `start.sh:148-149` takes the silent `curl -fsS -k` path with no message, and the PR body itself says the escape hatch skips verification up front "quietly." So Python warns where the docs and shell launcher stay silent. Minor, but worth aligning — either drop the bootstrap warn when the user explicitly opted in, or add the matching warning to `start.sh` and `ctl.sh`.

2. **Dockerfile two-curl timing.** `HEALTHCHECK --timeout=5s` runs two sequential `curl --max-time 4` calls (`Dockerfile:104-105`). In the common self-signed case the first verified curl fails fast (TLS handshake completes, cert is rejected sub-second), so the `-k` fallback has plenty of headroom — not a real concern there. The only squeeze is if the connection itself stalls for the full 4s, leaving the fallback ~1s before Docker kills it. Dropping each to `--max-time 2` would keep both calls inside the 5s budget with margin.

## One edge case neither I nor the bot flagged

`server.py:645-655` has a runtime TLS fallback: if `ctx.load_cert_chain(TLS_CERT, TLS_KEY)` throws (cert present but malformed/unreadable), the server prints "TLS setup failed, falling back to HTTP" and sets `scheme = 'http'`. In that scenario both env vars are set, so the launcher's `_tls_enabled()` returns true and probes https against a server now serving plain http. The probe would fail (not a cert error — a protocol mismatch), so `_is_cert_verification_error` returns false and `wait_for_health` just times out. That's a narrow misconfiguration (bad cert files), and arguably "fail the health check" is acceptable there, but it's the one case where launcher and server scheme can diverge. Not a blocker — flagging so it's a conscious choice rather than a surprise.

Overall LGTM pending the two minor alignments above. The test file exercising a real self-signed `server.py` over the full fallback path is the right way to cover this.

nesquena-hermes:
Thanks for this @tomas-forgac — the TLS-aware probe direction is right and the security shape is clean (the single unverified context is loopback-probe-only, verify-first, never global/client; an independent review confirmed no SSL-weakening, no parity gap, no injection). I took it through the full ship gate (Codex + Opus + the full pytest suite, 9349 passing) and the suite + security review are green, but Codex empirically reproduced **two real CORE regressions** that block the merge. Both are fixable and I'd love to ship once they're addressed:

---

**1. `ctl.sh status` can now crash before printing status (readonly `.env` vars).**

`status_cmd` now calls `_load_repo_dotenv_preserving_env`, which does a raw `source "${env_file}"`. Under `set -euo pipefail`, if `.env` contains a readonly shell variable assignment — most commonly `UID=...` (and often `GID=...` in Docker-oriented `.env` files) — bash errors on the readonly assignment and `ctl.sh status` exits non-zero **before printing any status**.

- Reproduced with a temp copy of `ctl.sh` + a `.env` containing `UID=1000`.
- `origin/master` did not source `.env` in `status_cmd`, so this is a new failure surface for a read-only command.
- **Fix:** have `_load_repo_dotenv_preserving_env` filter/skip readonly variables before sourcing (the same way `start.sh` already guards its `.env` load), rather than sourcing raw.

**2. Probes wait on HTTPS forever when the server has fallen back to HTTP (broken cert/key).**

`server.py` (≈653-655) catches a TLS-setup failure (cert/key present but unloadable) and **intentionally falls back to serving plain HTTP** — this is an existing, tested contract (`tests/test_tls_support.py::test_tls_startup_failure_fallback_to_http`). The new probes choose `https` purely from the *presence* of `HERMES_WEBUI_TLS_CERT`/`KEY` in the env, so in this fallback case the server serves HTTP while the probe polls HTTPS forever → startup is falsely reported as failed even though the server is up.

- Reproduced by launching `server.py` with `/nonexistent/cert.pem`+`/nonexistent/key.pem`: HTTP `/health` returned OK, but the new HTTPS `wait_for_health` returned false.
- Affects all 5 probe sites: `bootstrap.py` (~567), `start.sh` (~151), `ctl.sh` (~426), `scripts/wsl/hermes_webui_autostart.sh` (~43), `Dockerfile` HEALTHCHECK (~105).
- **Fix (pick one):** preserve the current contract by having the HTTPS probe paths fall back to plain HTTP when *both* the verified and unverified HTTPS attempts fail because the listener isn't speaking TLS; **or** (larger) drop `server.py`'s HTTP-fallback-on-TLS-failure and update its contract + test. The first option is the smaller, lower-risk change and keeps the existing behavior intact.

---

Everything else is solid and your branch is otherwise ship-ready — once these two are in, ping me and I'll re-run the full gate and ship it. (Note: master moved since you opened this, so a quick rebase will be needed too — happy to do that for you as part of the re-gate if you'd prefer.)

tomas-forgac:
Thanks for the review. I figured it was less risky to start over than to try to sync up with the latest master changes and try to fix this. New PR here: https://github.com/nesquena/hermes-webui/pull/4412

PR #4364, opened 2026-06-17
fix(#4362): gate runs API path behind explicit config opt-in

### LINKED ISSUE #4362: WebUI gateway chat falls back to /v1/runs even for normal conversations, creating isolated Api_Server sessions and losing context
## Summary
When the Hermes Gateway advertises `approval_events` / `run_approval_response` capabilities, the WebUI backend decides to use `POST /v1/runs` for ordinary browser chat. This turns every user turn into an isolated `api_server` run instead of a continuous chat session. The visible symptom is a sidebar full of "Api_Server Session" blocks; the functional symptom is mid-conversation context loss because each new run replays only the message history, not the live agent state.

## Reproduction
1. Configure Nesquena WebUI to talk to a Hermes Gateway that exposes `/v1/runs` and approval events (default native OpenClaw setup).
2. Open a single browser chat session.
3. Send several messages, including a multi-turn task (e.g. "edit file X" then "now also add Y").
4. Observe the sidebar: multiple "Api_Server Session" entries appear for the same conversation.
5. Observe behavior: the assistant "forgets" in-flight work even though the transcript is visible.

## Expected behavior
Normal WebUI chat should use `POST /v1/chat/completions` so the same agent loop is resumed each turn. `Api_Server Session` blocks should only appear for genuine subagent/cron/api_server activity, not for every browser message.

## Actual behavior
Each turn creates a new run via `POST /v1/runs` because `gateway_chat.py` sets `_use_runs_api = gateway_supports_approval(base_url, api_key)`.

Relevant code path:
- `api/gateway_chat.py:548` — `_use_runs_api = gateway_supports_approval(base_url, api_key)`
- `api/config.py` — `gateway_supports_approval` returns `True` when the gateway advertises `approval_events` + `run_approval_response`
- Gateway API server registers `/v1/runs`, `/v1/runs/{id}/events`, etc.

Result: every WebUI turn becomes a new isolated run with a fresh `run_id`.

## Workaround
Force `_use_runs_api = False` in `api/gateway_chat.py` for the WebUI backend so it falls back to `/v1/chat/completions`. This eliminated both the sidebar clutter and the mid-conversation amnesia in our deployment.

## Proposed fix
Either:
1. Default `_use_runs_api` to `False` for normal WebUI conversations and only enable it when the caller explicitly wants run lifecycle events, or
2. Provide an env/config override such as `NESQUENA_WEBUI_USE_RUNS_API=false` so deployments can opt out without patching source, or
3. Make `/v1/runs` genuinely resume an existing session instead of starting a fresh agent loop with only text replay.

## Environment
- Hermes Agent / Gateway: native systemd on OpenClaw (192.168.2.60)
- Nesquena WebUI: current `main` as of 2026-06-17
- Backend file patched: `/opt/nesquena/api/gateway_chat.py` line 552

## Additional context
This bug is especially confusing for users because the conversation transcript is still visible, so it looks like the assistant should remember what it was doing. The problem is not history display; it is that each turn is a brand-new agent run.

rodboev:
## Thinking Path

- The WebUI gateway chat path probes `/v1/capabilities` and, when the gateway advertises `approval_events` + `run_approval_response`, unconditionally routes every browser chat turn through `/v1/runs` (`gateway_chat.py:548`).
- The runs API creates isolated Api_Server sessions per turn, so mid-conversation context is lost; normal chat should stay on `/v1/chat/completions` which preserves the session via `X-Hermes-Session-Id`.
- The reporter's "Option 2" is a `use_runs_api` config flag (default `false`) that gates the runs API path behind explicit operator opt-in, which is the least invasive fix and avoids changing the capability probe itself.
- The flag follows the same env-var-over-config-dict pattern already used by `_gateway_base_url` and `_gateway_api_key`, keeping the precedence model consistent.

## What Changed

- `api/gateway_chat.py`: added `_WEBUI_GATEWAY_USE_RUNS_API_ENV` constant and `_gateway_use_runs_api_enabled()` helper that checks the env var / config key and defaults to `False`. The capability gate at line 558 now requires both `_gateway_use_runs_api_enabled(cfg)` AND `gateway_supports_approval(...)` before routing through `/v1/runs`.
- `tests/test_webui_gateway_chat_backend.py`: added tests for the new flag covering default-off, explicit opt-in values, rejection of generic truthy strings, config-dict support, env-var precedence, and a worker-boundary regression test proving chat/completions is used when the opt-in is absent even though the gateway advertises approval support.
- `tests/test_gateway_approval_runs_api.py`: updated existing runs-path tests to set `HERMES_WEBUI_GATEWAY_USE_RUNS_API=1` so they continue to exercise the runs API path through the new gate.

## Why It Matters

Gateways advertising approval capabilities now silently break normal WebUI chat by losing conversation context between turns. This fix restores the correct default (`/v1/chat/completions`) while giving operators who need the runs-based approval workflow a clean opt-in path.

## Verification

```
pytest tests/test_webui_gateway_chat_backend.py tests/test_gateway_approval_runs_api.py -v --timeout=60
```

## Risks / Follow-ups

- Operators currently relying on the implicit runs API routing will need to set `HERMES_WEBUI_GATEWAY_USE_RUNS_API=true` (env) or `"webui_gateway_use_runs_api": true` (config) after upgrading. This is the intended behavior per the issue's Option 2.
- The runs API path itself is unchanged; this PR only controls when it activates.

Closes #4362.

## Model Used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- **The runs API path in the gateway chat backend is now opt-in (#4362).** Previously, when the Hermes gateway advertised approval/runs support, the WebUI automatically routed through `/v1/runs` — which could silently change streaming behavior for existing deployments. The runs API path is now disabled by default and must be explicitly enabled via the `HERMES_WEBUI_GATEWAY_USE_RUNS_API` environment variable or `webui_gateway_use_runs_api: true` in `config.yaml`. Existing deployments that relied on implicit runs API routing should set this flag after upgrading.
+
 ## [v0.51.476] — 2026-06-17 — Release QK (cross-provider model-pick + non-git update-status fixes)
 
 ### Fixed
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -36,6 +36,7 @@ _STREAM_RUN_IDS: dict[str, str] = {}
 _WEBUI_CHAT_BACKEND_ENV = "HERMES_WEBUI_CHAT_BACKEND"
 _WEBUI_GATEWAY_BASE_URL_ENV = "HERMES_WEBUI_GATEWAY_BASE_URL"
 _WEBUI_GATEWAY_API_KEY_ENV = "HERMES_WEBUI_GATEWAY_API_KEY"
+_WEBUI_GATEWAY_USE_RUNS_API_ENV = "HERMES_WEBUI_GATEWAY_USE_RUNS_API"
 _GATEWAY_CHAT_BACKENDS = {"gateway", "api_server", "api-server"}
 
 
@@ -81,6 +82,18 @@ def _gateway_api_key(environ: dict[str, str] | None = None) -> str:
         or source.get("API_SERVER_KEY")
         or ""
     ).strip()
+
+
+def _gateway_use_runs_api_enabled(config_data=None, environ: dict[str, str] | None = None) -> bool:
+    """Return True only when the operator has explicitly opted into the runs API path."""
+    source = os.environ if environ is None else environ
+    cfg = config_data if isinstance(config_data, dict) else {}
+    raw = str(
+        source.get(_WEBUI_GATEWAY_USE_RUNS_API_ENV)
+        or cfg.get("webui_gateway_use_runs_api")
+        or ""
+    ).strip().lower()
+    return raw in ("1", "true", "yes", "on")
 
 
 def gateway_chat_config_status(config_data=None, environ: dict[str, str] | None = None) -> dict:
@@ -545,7 +558,7 @@ def _run_gateway_chat_streaming(
         base_url = _gateway_base_url(cfg)
         api_key = _gateway_api_key()
         # Capability gate: use runs API when gateway advertises approval support.
-        _use_runs_api = gateway_supports_approval(base_url, api_key)
+        _use_runs_api = _gateway_use_runs_api_enabled(cfg) and gateway_supports_approval(base_url, api_key)
         if _use_runs_api:
             body_extras = {}
             if model_provider:
--- a/tests/test_gateway_approval_runs_api.py
+++ b/tests/test_gateway_approval_runs_api.py
@@ -151,7 +151,7 @@ def test_gateway_runs_api_submission():
     mock_session.pending_started_at = None
 
     try:
-        with patch.dict("os.environ", {"HERMES_WEBUI_CHAT_BACKEND": "gateway"}):
+        with patch.dict("os.environ", {"HERMES_WEBUI_CHAT_BACKEND": "gateway", "HERMES_WEBUI_GATEWAY_USE_RUNS_API": "1"}):
             with patch("api.gateway_chat.gateway_supports_approval", lambda *_args, **_kwargs: True), \
                  patch("api.gateway_chat._run_gateway_runs_api_streaming", fake_runs_streaming), \
                  patch("api.gateway_chat.get_session", return_value=mock_session), \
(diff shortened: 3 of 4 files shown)


nesquena-hermes:
Thanks @rodboev — the opt-in gate itself is clean (env `HERMES_WEBUI_GATEWAY_USE_RUNS_API` works, default-off verified, 42 tests green). One small doc/code mismatch to fix before this ships:

**The CHANGELOG documents a config.yaml key the code doesn't read.**

The CHANGELOG entry says:

> must be explicitly enabled via the `HERMES_WEBUI_GATEWAY_USE_RUNS_API` environment variable or `gateway.use_runs_api: true` in `config.yaml`.

But `_gateway_use_runs_api_enabled()` reads a **flat top-level** key, not a nested `gateway.` section:

```python
# api/gateway_chat.py:93
raw = str(
    source.get(_WEBUI_GATEWAY_USE_RUNS_API_ENV)
    or cfg.get("webui_gateway_use_runs_api")   # ← flat key
    or ""
).strip().lower()
```

`cfg` here is the full `get_config()` dict, and the sibling gateway settings use the same flat convention (`webui_chat_backend`, `webui_gateway_base_url`). So a user who follows the CHANGELOG and sets

```yaml
gateway:
  use_runs_api: true
```

gets **no effect** — the code never looks under a `gateway:` section. Only `webui_gateway_use_runs_api: true` (flat) or the env var actually enable it.

**Fix (pick one):**
- **Simplest:** change the CHANGELOG to the key the code actually reads — `webui_gateway_use_runs_api: true` in `config.yaml` (matches the existing `webui_gateway_*` keys).
- **Or**, if you'd prefer the nicer nested `gateway.use_runs_api` shape, also read it in `_gateway_use_runs_api_enabled` (e.g. `(cfg.get("gateway") or {}).get("use_runs_api")`) so both the flat and nested forms work — but then add a test for the nested path.

Everything else looks good to ship once the documented key matches the code.

rodboev:
Good catch — the CHANGELOG now says `webui_gateway_use_runs_api: true` (flat key, matching what `_gateway_use_runs_api_enabled()` reads and the sibling `webui_gateway_*` keys). Also rebased onto current master.

nesquena-hermes:
Confirmed the fix lands correctly — re-reading the rebased HEAD (`c4606e23`).

The CHANGELOG now documents the flat key the code actually reads:

```
must be explicitly enabled via the `HERMES_WEBUI_GATEWAY_USE_RUNS_API`
environment variable or `webui_gateway_use_runs_api: true` in `config.yaml`.
```

That matches `_gateway_use_runs_api_enabled()` at `api/gateway_chat.py:87-96`:

```python
raw = str(
    source.get(_WEBUI_GATEWAY_USE_RUNS_API_ENV)
    or cfg.get("webui_gateway_use_runs_api")   # flat key, matches webui_gateway_* siblings
    or ""
).strip().lower()
return raw in ("1", "true", "yes", "on")
```

and the sibling settings (`webui_gateway_base_url`, `webui_gateway_api_key`) use the same flat convention, so a user copy-pasting the CHANGELOG key now gets the intended effect.

The gate at `api/gateway_chat.py:561` reads correctly:

```python
_use_runs_api = _gateway_use_runs_api_enabled(cfg) and gateway_supports_approval(base_url, api_key)
```

so the runs API path only activates when the operator opts in **and** the gateway advertises approval support — the default-off behavior the issue asked for. Precedence is sound: `os.environ` wins when the env var is set to a truthy value, and an unset/empty env var falls through to the config dict via the `or` chain (only `None`/empty string fall through, so an explicit env value can't be silently overridden by config). The truthy-string allowlist (`1/true/yes/on`) correctly rejects accidental values like `"0"` or `"false"`.

Reads clean to me now — doc and code agree. Nothing else blocking from my side.

nesquena-hermes:
Shipped in **v0.51.477** (Release QL) 🚀 — thanks @rodboev!

Your fix is live (closes #4362): gateway chat no longer silently loses mid-conversation context. Runs-API routing is now gated behind the explicit `HERMES_WEBUI_GATEWAY_USE_RUNS_API` opt-in (env or `webui_gateway_use_runs_api` config, default off); normal gateway chat stays on `/v1/chat/completions` and keeps session continuity via `X-Hermes-Session-Id`.

Ran it through the full ship gate — both advisors confirmed the default-flip is safe: the approval-events flow **degrades gracefully** when the flag is off (simply not offered — no hang, no error, no stale relay) and is fully recoverable via the documented opt-in. Codex SAFE TO SHIP, Opus ship-it, full suite 9346 passing. Attribution preserved via `Co-authored-by`. Appreciate the clean fix + the clear writeup of the runs-API context-loss root cause!

PR #4365, opened 2026-06-17
feat(#4360): extend credential pool quota status to all pooled providers

### LINKED ISSUE #4360: feat(providers): extend credential pool quota status to non-Codex providers (opencode-zen, opencode-go, etc.)
## Summary

The Providers panel already has a credential pool breakdown UI — it renders per-credential status, exhaustion state, rate-limit windows, and reset timers. However, this only works for providers in `_ACCOUNT_USAGE_PROVIDERS` (currently `{"openai-codex", "anthropic"}`).

When a provider like `opencode-zen` has multiple credentials in the pool (added via `hermes auth add`), the WebUI shows the provider as a single flat entry with no pool visibility. Users can't see:

- Which credential is currently active
- Which credentials are rate-limited/exhausted
- Cooldown/reset timers per credential
- The rotation strategy in use

## Current behavior

- `hermes auth list opencode-zen` shows 2 credentials, one rate-limited
- WebUI Providers panel shows opencode-zen as a single card with no pool breakdown
- `get_provider_quota("opencode-zen")` returns `{"status": "unsupported"}`

## Expected behavior

The pool breakdown UI (already built in `_buildProviderQuotaPoolBreakdown` in `panels.js`) should render for any provider that has `credential_pool` entries in `auth.json`, not just Codex/Anthropic.

## Proposed approach

1. Add a generic pool snapshot path in `providers.py` that reads `auth.json`'s `credential_pool.<provider>` for any provider
2. Return pool credential status (active/exhausted/rate-limited), labels, and reset timers from local `auth.json` data
3. Keep existing probe-based path for Codex/Anthropic (richer data)
4. Fall back to local pool data for all other providers

## Code pointers

- `api/providers.py:_ACCOUNT_USAGE_PROVIDERS` (line 80)
- `api/providers.py:_codex_pool_snapshot` (line 513)
- `api/providers.py:get_provider_quota` (line 1623)
- `api/config.py:_get_auth_store_path` (line 2851)
- `static/panels.js:_buildProviderQuotaPoolBreakdown` (line 7073)

## Workaround

Users can run `hermes auth list <provider>` in the terminal, but this should be visible in the WebUI.

rodboev:
## Thinking Path

- The Settings page's quota status panel has a pool breakdown UI (`_buildProviderQuotaPoolBreakdown`) that is fully provider-agnostic, rendering credential counts and exhaustion state from whatever `pool.credentials` the backend returns.
- But the backend gates pool logic behind `_ACCOUNT_USAGE_PROVIDERS`, a two-member frozenset of `{"openai-codex", "anthropic"}`. Every other provider with multiple pooled credentials gets `{"status": "unsupported"}`, even though `load_pool(provider)` already works for any provider and the pool entries carry `last_status`, exhaustion timestamps, and labels generically.
- The probe-based path (spawning subprocesses to query usage APIs) is provider-specific and expensive, but reading local pool metadata (credential count, exhaustion state from auth.json) is free and generic.
- Adding a probe-free `_local_pool_snapshot(provider)` that reads pool entries and classifying them as available/exhausted lets any pooled provider show meaningful quota status in the UI without API calls, while preserving the richer probe-based path for Codex and Anthropic.

## What Changed

- `api/providers.py`: added `_local_pool_snapshot(provider)` that reads `load_pool(provider)` entries, classifies each as available, exhausted, or dead using existing helpers (`_entry_is_pool_exhausted`, `_entry_pool_exhausted_reason`, `_entry_pool_retry_after`, `_safe_entry_label`), and returns a `SimpleNamespace` compatible with `_serialize_account_usage_snapshot`
- `api/providers.py`: in `get_provider_quota`, the openrouter path is now handled first, followed by `_local_pool_snapshot` for non-allowlist providers, then the `unsupported` fallback
- `tests/test_issue4360_generic_pool_quota.py`: new test file with 15 tests covering local pool snapshot states (empty, available, exhausted, dead, mixed), `get_provider_quota` routing for non-allowlist providers with and without pools, and confirming allowlist and openrouter providers still use their dedicated paths

## Why It Matters

Users who configure multiple credentials for providers like Google, Mistral, or custom endpoints see "Quota status is not available" even though pool infrastructure already tracks their credentials. This change surfaces credential counts and exhaustion state for any pooled provider, making the quota panel useful beyond Codex and Anthropic without adding any API probe overhead.

## Verification

```
pytest tests/test_issue4360_generic_pool_quota.py -v --timeout=60
```

Full-suite CI context: `pytest tests/ -v --timeout=60`.

## Risks / Follow-ups

- The local pool path shows credential availability/exhaustion/dead state only, not usage percentages or rate-limit windows (those require provider-specific API probes). This is a deliberate scope limitation; richer probes for additional providers can be added incrementally.
- If a provider's pool entries lack `last_status` metadata (newly added credentials that have never been used), they will all appear as "available", which is the correct default.
- `load_pool(provider)` resolves `auth.json` through ambient `HERMES_HOME`; profile-scoped resolution is a pre-existing gap shared with the existing pool infrastructure and outside this PR's scope.
- No frontend changes are needed since `_buildProviderQuotaPoolBreakdown` is already generic.

Closes #4360.

## Model Used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- **The Settings quota panel now shows credential pool status for all pooled providers, not just Codex and Anthropic (#4360).** Each pooled credential is listed with its active/exhausted state, retry-after time, and reason — giving a unified pool health view across every provider that uses credential rotation.
+
 ## [v0.51.501] — 2026-06-18 — Release RK (Ctrl/Cmd+, opens Settings)
 
 ### Added
@@ -137,29 +141,6 @@
 - **The inline PDF preview now renders all pages in a scrollable view instead of just page 1 (#4353).** Previously only the first page of a PDF attachment/export was rendered inline, so multi-page documents looked clipped/broken. The preview now renders every page (capped at 20 for memory) stacked vertically in a scrollable body, shows the page count in the header, and adds a "Showing the first N of M pages — download for the full document" notice when a PDF exceeds the cap. Per-page render failures are isolated (a malformed page is skipped rather than halting the whole preview), and each canvas is attached only after a successful render. Thanks @gourangasatapathyvit.
 
 ## [v0.51.479] — 2026-06-17 — Release QN (named context blocks in the composer)
-
-### Added
-
-- **Selected chat text can now be collected as named context blocks before sending (#2543/#3306).** When you select text in a conversation and use "Reply with selection," instead of immediately dumping it into the composer, it's added as a small named chip ("Context 1", "Context 2", …) in a row above the message box. You can rename a chip (double-click) or remove it (×), stack several from different parts of the conversation, and on send they're flushed into the message as labeled markdown-quoted blocks (`**Name:**` + quoted text). Switching away from a session clears any pending blocks so they don't leak between conversations. Thanks @rodboev.
-
-## [v0.51.478] — 2026-06-17 — Release QM (read-only LLM Wiki browser in Insights)
-
-### Added
-
-- **The Insights panel now has a read-only LLM Wiki browser (#2941).** When the Hermes agent has an LLM Wiki configured (`WIKI_PATH` / `skills.config.wiki.path`), the Insights panel surfaces a knowledge-base observability section — enabled state, entry/page/raw-file counts, last-updated, and last writer — plus a browsable read-only view of the wiki pages. When no wiki directory is configured it shows a clear "Unavailable — set `WIKI_PATH`…" empty state instead of failing. The browser is strictly read-only and auth-gated, and the page-loading path is hardened against directory traversal (realpath + symlink-safe resolution). Thanks @rodboev.
-
-## [v0.51.477] — 2026-06-17 — Release QL (gateway runs-API routing now opt-in)
-
-### Fixed
-
-- **Gateway chat no longer silently loses mid-conversation context (#4362).** When the WebUI was connected to a Hermes gateway that advertised approval/runs support, it routed *every* browser chat turn through the gateway's `/v1/runs` API, which creates an isolated Api_Server session per turn — so the agent lost the conversation's context between messages. Routing through the runs API is now gated behind an explicit opt-in (`HERMES_WEBUI_GATEWAY_USE_RUNS_API=1` env var, or `webui_gateway_use_runs_api: true` in config), defaulting to off. Normal gateway chat now stays on `/v1/chat/completions`, which preserves the session via `X-Hermes-Session-Id`. Operators who want the in-gateway-chat approval-events flow can re-enable it with the flag. Thanks @rodboev.
-
-## [v0.51.476] — 2026-06-17 — Release QK (cross-provider model-pick + non-git update-status fixes)
-
-### Fixed
-
-- **Selecting a model from a non-default provider no longer silently reverts to the default model (#4363).** When the model dropdown rebuilds (panel switch, provider refresh, or a live-model fetch that times out and serves a fallback catalog), a cross-provider selection such as `@my-local:gemma4:12b` could vanish from the refreshed catalog — and the browser would quietly snap the `<select>` back to its first `<option>` (the default model). The dropdown now re-injects the chosen model as a custom option when it's missing from the rebuilt catalog, so your selection is preserved instead of silently changing under you. Thanks @edoumo.
-- **Docker and pip installs now show "Can't check for updates" instead of the misleading "Up to date" (#4356).** When the WebUI or agent is installed without a `.git` directory (e.g. via Docker image or `pip install`), the update check previously reported "Up to date" because there was nothing to compare against. It now detects the absence of `.git` and surfaces "Can't check for updates" in the Settings panel. In mixed deployments where one component has a git checkout and the other does not, the valid git-side update info is still shown alongside the can't-check indicator. Thanks @rodboev.
 
 ## [v0.51.475] — 2026-06-17 — Release QJ (ElevenLabs TTS engine)
 
--- a/api/providers.py
+++ b/api/providers.py
@@ -854,6 +854,7 @@ def _local_pool_snapshot(provider):
     """
     try:
         from agent.credential_pool import load_pool
+        from api.config import _is_ambient_gh_cli_entry
 
         pool = load_pool(provider)
         entries = list(pool.entries()) if pool is not None and hasattr(pool, "entries") else []
@@ -867,6 +868,11 @@ def _local_pool_snapshot(provider):
     exhausted_count = 0
     dead_count = 0
     for index, entry in enumerate(entries, start=1):
+        source = str(_entry_value(entry, "source") or "")
+        label_val = str(_entry_value(entry, "label", "source") or "")
+        key_source = str(_entry_value(entry, "key_source") or "")
+        if _is_ambient_gh_cli_entry(source, label_val, key_source):
+            continue
         label = _safe_entry_label(entry, index)
         entry_status = str(_entry_value(entry, "last_status") or "").strip().lower()
         if entry_status == "dead":
@@ -906,8 +912,12 @@ def _local_pool_snapshot(provider):
                 "fetched_at": None,
             })
 
+    if not rows:
+        return None
+
+    total = available_count + exhausted_count + dead_count
     pool_dict = {
-        "total_credentials": len(entries),
+        "total_credentials": total,
         "queried_credentials": 0,
         "available_credentials": available_count,
         "exhausted_credentials": exhausted_count,
@@ -919,7 +929,7 @@ def _local_pool_snapshot(provider):
         "credentials": rows,
     }
 
-    details = [str(available_count) + "/" + str(len(entries)) + " credentials available"]
+    details = [str(available_count) + "/" + str(total) + " credentials available"]
     if exhausted_count:
         details.append(str(exhausted_count) + " exhausted")
     if dead_count:
--- a/api/providers.py
+++ b/api/providers.py
@@ -679,6 +679,7 @@ else:
     print(json.dumps(_fetch_snapshot(provider, api_key)), flush=True)
 """
 
+
 # SECTION: Provider ↔ env var mapping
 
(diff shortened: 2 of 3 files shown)


rodboev:
Re: the P1 duplicate-helpers finding — those aren't duplicates of module-level functions. The earlier definitions (lines 184–451) live inside the triple-quoted string `_ACCOUNT_USAGE_SUBPROCESS_CODE` (lines 167–679), which gets `exec()`'d in a subprocess. They exist only as text inside that string, not as callable Python objects at module level. The additions at lines 756–847 are the first and only real module-level definitions, required by `_local_pool_snapshot`. Removing them would break the function.

The CHANGELOG entry has been added.

nesquena-hermes:
Reading `api/providers.py` at the PR HEAD (`5609cf91`) against `origin/master`, the routing change is clean and the new local-pool path is a nice probe-free win. One thing greptile flagged is real and worth fixing before merge: the diff introduces a **second, full copy of nine existing helper functions**.

## What I found

On master, each of these helpers is defined exactly once. At the PR HEAD they each have a second definition in the block just above `_local_pool_snapshot`:

| helper | original | duplicate |
|---|---|---|
| `_iso` | 184 | 790 |
| `_parse_dt` | 246 | 770 |
| `_entry_value` | 310 | 756 |
| `_safe_entry_label` | 395 | 824 |
| `_entry_exhausted_ttl_seconds` | 416 | 800 |
| `_entry_pool_exhausted_until` | 423 | 807 |
| `_entry_is_pool_exhausted` | 435 | 819 |
| `_entry_pool_exhausted_reason` | 440 | 838 |
| `_entry_pool_retry_after` | 451 | 834 |

I diffed each pair byte-for-byte — all nine are **currently identical**, so there's no runtime bug today (Python just binds the name to the second `def`, and the second copy matches the first). CI is green for exactly that reason. The problem is the divergence hazard: the moment someone edits the "original" at e.g. `_entry_pool_exhausted_until:423`, the live behavior keeps using the shadowing copy at `:807` and the edit silently does nothing. That's a nasty future footgun in a file this size.

## Recommendation

Delete the duplicated block (the second copies around `:756-847`) and let `_local_pool_snapshot` call the originals — they're module-level, so no import or reordering is needed. `_local_pool_snapshot` already references them by bare name:

```python
label = _safe_entry_label(entry, index)
entry_status = str(_entry_value(entry, "last_status") or "").strip().lower()
...
elif _entry_is_pool_exhausted(entry):
    ...
    "unavailable_reason": _entry_pool_exhausted_reason(entry),
    "retry_after": _entry_pool_retry_after(entry),
```

so removing the lower copies leaves it resolving to the canonical definitions higher up with zero behavior change. Worth a quick grep after the delete to confirm only one `def` of each remains (`grep -n "^def _entry_" api/providers.py`).

## On the routing itself

The `get_provider_quota` reorder reads correctly: `_ACCOUNT_USAGE_PROVIDERS` first (probe path for Codex/Anthropic), then `openrouter`, then `_local_pool_snapshot(provider)`, then the `unsupported` fallback only when no pool exists. `_local_pool_snapshot` is well-guarded (`except Exception: return None`, empty-entries → `None`), so a provider with no pool cleanly falls through to the old unsupported response. The available/exhausted/dead classification matches the existing pool helpers, and the `account_limits.get("available")` branch correctly downgrades to `status: "unavailable"` when every credential is exhausted. Once the dup block is gone this looks ready.

nesquena-hermes:
Thanks @rodboev — the probe-free "show pool health for any multi-credential provider" direction is great, and the test coverage is solid. But the deep review (Codex/Opus + the credential-pool regression history from #4247) found one **real bug** that blocks merge, plus the duplicate-helpers cleanup greptile has flagged twice. Both are well-scoped:

---

**1. [CORE] `_local_pool_snapshot` leaks the ambient gh-cli credential as a phantom pooled credential.**

`_local_pool_snapshot(provider)` (api/providers.py) iterates `pool.entries()` **raw**, with no ambient-entry filter. But `load_pool()` includes the auto-seeded `gh auth token` entry (`source="gh_cli"`) on any machine where `gh` is installed — so for a provider like **copilot** (and any provider whose pool surfaces that ambient entry), the new generic path at `get_provider_quota` (the `_local_pool_snapshot` branch, reached for all non-allowlist / non-openrouter providers) will report a phantom "1 credential — available" in the Settings quota panel even when the user never configured a pool for that provider.

This is the exact regression class as #4247 (`has_credentials()` counting the ambient `gh_cli` entry → Copilot falsely showing "configured"). The existing detection paths already guard against it — `api/config.py` has `_is_ambient_gh_cli_entry(source, label, key_source)` (used in ~4 places, e.g. config.py:3779/4126/5065/5081) precisely so "Copilot doesn't appear just because `gh` is installed."

- **Fix:** in `_local_pool_snapshot`'s entry loop, skip ambient entries before counting/rowing them:
  ```python
  from api.config import _is_ambient_gh_cli_entry  # or wherever it's importable
  source = _entry_value(entry, "source") or ""
  label  = _entry_value(entry, "label", "source") or ""
  key_source = _entry_value(entry, "key_source") or ""
  if _is_ambient_gh_cli_entry(source, label, key_source):
      continue
  ```
  Then if all entries were ambient (`not rows`), return `None` so the provider falls through to the existing `unsupported` response rather than showing a phantom pool.
- **Test:** add a case with a single `gh_cli`-source entry and assert the provider returns `unsupported` (no phantom "available" row). This is the canary the #4247 fix established.

**2. [cleanup] Remove the 9 duplicated helper definitions (greptile flagged twice).**

`_iso`, `_parse_dt`, `_entry_value`, `_safe_entry_label`, `_entry_exhausted_ttl_seconds`, `_entry_pool_exhausted_until`, `_entry_is_pool_exhausted`, `_entry_pool_exhausted_reason`, `_entry_pool_retry_after` are each defined **twice** (≈ lines 184–451 and again 756–838). They're byte-identical, so no behavioral bug today — the second set shadows the first — but it's ~120 lines of dead code and a real divergence trap (a future edit to one copy silently does nothing). Remove one full set.

---

Everything else checks out — probe-free, purely additive routing, OpenRouter + allowlist paths unaffected, 15 new tests, exception-guarded. Once the ambient filter + dedup are in, ping me and I'll re-run the full Codex+Opus+suite gate and ship. (Master moved since you opened this, so a quick rebase too — happy to handle that at re-gate.)

rodboev:
@nesquena-hermes Great catch on the ambient `gh_cli` leak, that's the exact #4247 regression class.

**1. Ambient entry filter:** `_local_pool_snapshot` now skips entries where `_is_ambient_gh_cli_entry(source, label, key_source)` returns true before counting or classifying them. If all entries in a pool are ambient, the function returns `None` so the provider falls through to the `unsupported` response. Two new tests cover the canary case (single `gh_cli` entry -> `None`) and the mixed case (ambient + real -> only real counted).

**2. On the duplicate helpers:** Those nine functions (`_entry_value`, `_parse_dt`, `_iso`, etc.) aren't actually duplicated. The first set lives inside `_ACCOUNT_USAGE_SUBPROCESS_CODE` (the `r"""..."""` raw string literal at lines 166-680 that's executed as a subprocess for the probe-based path). The second set is the real module-level definitions that `_local_pool_snapshot` calls. They look identical in a line-by-line diff because the subprocess code was written to match the module helpers, but removing either copy would break its respective call site. No change needed here.

Pushed at `470de00a`, all 17 tests pass. Ready for re-gate whenever.

nesquena-hermes:
Re-pulled `470de00a`... actually the head I fetched is `df674267` (the ambient-filter commit) — re-checked both items against it.

## On the duplicate helpers — you're right, I was wrong, retracting that ask

I went and verified this directly in a worktree and **rodboev is correct**: those are not duplicate module-level definitions, and my earlier two comments asking to delete the lower set were a false positive. Here's the proof:

```
$ grep -n '"""' api/providers.py | head
166:_ACCOUNT_USAGE_SUBPROCESS_CODE = r"""
680:"""
```

The `_ACCOUNT_USAGE_SUBPROCESS_CODE` raw string opens at `api/providers.py:166` and doesn't close until `:680` — there are no triple-quotes in between. So the "first set" at lines 184–451 (`_iso`, `_parse_dt`, `_entry_value`, …) are **text inside that string literal**, written at column 0 so they're valid top-level Python when the string gets `exec()`'d in the probe subprocess. They are not callable objects at module load. The reason a naive `grep -n "^def _entry"` shows two of each is exactly that the subprocess code is unindented inside the raw string — `grep` can't tell it's inside a string, but the Python parser can.

The only real module-level definitions are the ones at `:756–838`, which `_local_pool_snapshot` calls. Deleting either set breaks its respective call site, precisely as you said. No change needed, and apologies for the churn on that one.

## Ambient gh_cli filter — verified correct

The #4247-class leak is properly closed. `_local_pool_snapshot` (`api/providers.py:849`) now skips ambient entries before counting:

```python
source = str(_entry_value(entry, "source") or "")
label_val = str(_entry_value(entry, "label", "source") or "")
key_source = str(_entry_value(entry, "key_source") or "")
if _is_ambient_gh_cli_entry(source, label_val, key_source):
    continue
```

and falls through to `unsupported` when nothing real remains (`if not rows: return None` at `:916`). The call signature matches `_is_ambient_gh_cli_entry(source, label, key_source)` defined at `api/config.py:1523` exactly. Both canary tests are in place — `test_local_pool_snapshot_filters_ambient_gh_cli_entries` (single `gh_cli` entry → `None`) and `test_local_pool_snapshot_keeps_non_ambient_alongside_ambient` (mixed → only the real one counted) — so the regression is locked.

## One housekeeping note (no action needed from you)

greptile's CHANGELOG flag is a stale-branch artifact, not a real regression. Your branch's merge-base still had v0.51.476–480 at the top, but `origin/master` has since stacked v0.51.481–489 above them (`git show origin/master:CHANGELOG.md` → `v0.51.489` … `v0.51.480`). So the deletion just reflects how far master moved; it'll resolve cleanly when I rebase + restage at re-gate. Don't hand-restore those sections — you'd just create a conflict against master's newer top.

Net: code is ready. I'll rebase onto current master, re-run the full Codex+Opus+suite gate, and ship. Thanks for pushing back on the helpers — that was the right call.

PR #4366, opened 2026-06-17
fix(#4356): report non-git installs as "can't check" instead of "up to date"

### LINKED ISSUE #4356: Update check and notification not working
I noticed that I never get an update notification even though new versions are constantly being released. Manually clicking the 'Check now' button returns 'Up to date' while this is not true. It applies to hermes agent versions as well as hermes webui versions.

As you can see in the screenshot below from the settings screen, I'm on WebUI: v0.51.454 and Agent: 0.16.0 and you see it states that they are 'Up to date' while WebUI is on version v0.51.475 already. 

Image

I had agent and webui running in separate docker container first. Currently they are backed together in a single container as per the docs instructions. When pressing the 'Check now' button on the settings screen I see this in the docker log;
```
17.06.2026 10:21:48 [hermes] [webui] {"ts": "2026-06-17T08:21:48Z", "remote": "172.21.0.18", "method": "POST", "path": "/api/updates/check", "status": 200, "ms": 0.9, "forwarded_for": "172.70.243.103"}
17.06.2026 10:21:50 [hermes] [webui] {"ts": "2026-06-17T08:21:50Z", "remote": "172.21.0.18", "method": "GET", "path": "/api/health/agent", "status": 200, "ms": 1.2, "forwarded_for": "172.70.243.103"}
```

rodboev:
## Thinking Path

- Docker and pip-installed instances don't have `.git`, so `_check_repo` returns `None` at the very first guard (`api/updates.py:655-656`).
- `None` flows into the update cache and reaches the frontend as `null`, where `_formatUpdateTargetStatus` returns `null` for null info, and `panels.js` interprets both parts being falsy as "Up to date."
- Version detection already works for non-git installs (`_detect_webui_version` reads `api/_version.py`, `_detect_agent_version` reads `VERSION`), so the version badge is correct, but the update check silently claims everything is current when it never actually checked.
- The fix is two-layered: the backend returns a sentinel dict with `no_git: True` instead of `None`, and the frontend renders a distinct "Can't check" message for that sentinel instead of falling through to "Up to date."

## What Changed

- `api/updates.py`: `_check_repo` returns `{'name': name, 'behind': None, 'no_git': True}` instead of `None` when `.git` is absent, with updated docstring
- `static/ui.js`: `_formatUpdateTargetStatus` returns `null` for `no_git` entries, preventing them from being classified as "update available" or triggering the update banner
- `static/panels.js`: update check consumer adds a `no_git` branch that renders `settings_update_no_git` in muted color, between the error and up-to-date branches
- `static/i18n.js`: added `settings_update_no_git` key across all 13 locale blocks
- `tests/test_issue4356_no_git_update_check.py`: new test file covering the sentinel return, frontend rendering branches, and i18n key coverage

## Why It Matters

Docker is a primary deployment method for Hermes. Users on Docker installs see "Up to date" indefinitely, missing every release. The false positive erodes trust in the update system and means Docker users only discover new versions through external channels.

## Verification

```
pytest tests/test_issue4356_no_git_update_check.py tests/test_updates.py -v --timeout=60
```

## Risks / Follow-ups

- **No behavior change for git installs:** the sentinel is only returned when `.git` is absent; all existing git-based update paths are untouched.
- **Deferred: GitHub releases API comparison for Docker.** A follow-up could compare the baked `_version.py` version against GitHub releases to give Docker users actual update notifications. This PR deliberately does not implement that, keeping the scope to fixing the false "Up to date" message.
- **i18n: non-English translations are best-effort.** The translations follow the repo's convention of keeping technical terms in English parentheticals. Native speakers may refine them in subsequent PRs.

## Upstream

Closes #4356.

## Model Used


nesquena-hermes:
Shipped in **v0.51.476** (Release QK) 🚀 — thanks @rodboev!

Your fix is live (closes #4356): Docker and pip installs now show "Can't check for updates" instead of the misleading "Up to date". We ran it through the full release gate (Codex + Opus + the full pytest suite, 9340 passing).

One maintainer addition, attributed to you via `Co-authored-by`: both Codex and Opus independently flagged a mixed-deployment edge case in the frontend — `[data.webui,data.agent].filter(Boolean).every(c=>c.no_git)` only surfaced the can't-check indicator when *every* target was non-git, so a mixed install (one git checkout up-to-date + one non-git target) fell through to "Up to date" and hid the indicator (which also contradicted the CHANGELOG promise about mixed deployments). I replaced it with a per-target `noGitParts` accumulator that also filters `ignored` agent targets, so the can't-check indicator now shows alongside any git-side update info — and strengthened your test to guard it.

Backend `_check_repo` sentinel change (`None` → `{name, behind: None, no_git: True}`) was traced safe across every consumer (the notification/summary builders all gate on `behind > 0`). Closing as merged — thank you!

Issue #4346, opened 2026-06-17
Transcript virtualization scroll stability (Phase B: re-default-on root-cause fix) — variable-height anchor oscillation


nesquena-hermes:
## Summary

Track the **root-cause fix** for transcript virtualization scroll stability, so the feature can eventually be re-defaulted-on. This is **Phase B** of #4343 — Phase A (default the feature OFF / experimental / opt-in, shipped v0.51.470) bought us safety; this issue is the actual engineering work to make the windowed scroll behavior reliable.

## Background

Transcript virtualization (#500 / #4214) windows long transcripts (>80 renderable rows, `MESSAGE_VIRTUAL_THRESHOLD_ROWS`) to keep rendering fast. Two scroll bugs surfaced:

- **#4325** (fixed) — stream-completion jump: the post-stream re-render restored a stale pixel `scrollTop`. Fixed with semantic viewport anchoring (re-pin by `data-msg-idx`).
- **#4343** (Phase A shipped, default-off) — **scroll-up flicker**: on long sessions with tall tool-call rows, scrolling up makes the viewport "flick back and forth" (reporter @b3nw, 565-msg session, ~90 tool calls).

## Root cause (the thing Phase B must fix)

Classic **variable-height virtual-scroll anchor oscillation**. The virtual window is computed from a per-row height cache where any *unmeasured* row falls back to a fixed estimate (`MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHT`, `static/ui.js` `_messageVirtualWindow` ~459-494). With ~90 large tool-call rows whose real heights dwarf the estimate:

1. Scroll up → `_scheduleMessageVirtualizedRender` (`ui.js:755`) recomputes the window from **estimated** heights, re-renders the newly-visible head rows.
2. Those rows are measured at their **actual** (much taller) heights → `_scheduleMessageVirtualMeasurementRefresh` (`ui.js:526`) schedules another render with corrected `topPad`/`bottomPad`.
3. Corrected padding changes `scrollHeight` → shifts `scrollTop` → moves the window → another render → another measurement. That round-trip is the visible flick.

`MESSAGE_VIRTUAL_MEASUREMENT_MAX_RERENDERS` bounds the loop (no runaway) but the user still sees it settle visibly. It's structural to estimate-then-measure virtualization with high height variance, not a single off-by-one.

## Candidate approaches (to evaluate)

[...]

PR #4367, opened 2026-06-17
fix(#4346): per-role virtual scroll height estimates

rodboev:
## Thinking Path

- The virtual transcript estimates every unmeasured row at 140px, but tool-call assistant rows include adjacent activity/tool card siblings and measure 500-2000px actual.
- Per-role fallbacks (user ~120px, assistant ~160px, tool_call ~400px) reduce the initial estimate error without changing the measurement cache, render lifecycle, or running-average convergence.
- This is intentionally a partial mitigation; measured heights remain authoritative and the running average still converges from real measurements.

## What Changed

- `static/ui.js`: Replace the single `MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHT` constant with a `MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHTS` object and `_messageVirtualDefaultHeightForRole()` helper. Add `_messageVirtualRoleForEntry()` classifier that detects tool-call rows through `m.tool_calls`, `tool_use` content type, and `_partial_tool_calls`. Pass `roleForIdx` callback into `_messageVirtualWindow()` from `_currentMessageVirtualWindow`.
- `tests/test_issue500_message_list_virtualization.py`: Add Node-evaluated coverage for role-specific defaults, role classification, and cached-height override behavior.

## Why It Matters

Tool-heavy transcript rows start with a closer spacer estimate (400px vs 140px), so the virtual window has less correction to apply after measurement. That reduces visible scroll oscillation while keeping the existing bounded measurement refresh behavior.

## Verification

```
```

## Risks / Follow-ups

The chosen defaults are heuristic; very large tool outputs can still exceed the 400px estimate. Target 175 is the structural stabilization that eliminates visible drift regardless of estimate accuracy.

## Upstream

Ref #4346

## Model Used

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -376,7 +376,17 @@ function _statusCardHtml(card){
 const MESSAGE_RENDER_WINDOW_DEFAULT=50;
 const MESSAGE_VIRTUAL_THRESHOLD_ROWS=80;
 const MESSAGE_VIRTUAL_BUFFER_PX=900;
-const MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHT=140;
+const MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHTS={
+  user:120,
+  assistant:160,
+  tool_call:400,
+  default:140,
+};
+function _messageVirtualDefaultHeightForRole(role){
+  return MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHTS[
+    role&&Object.prototype.hasOwnProperty.call(MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHTS,role)?role:'default'
+  ];
+}
 const MESSAGE_VIRTUAL_MEASUREMENT_MAX_RERENDERS=2;
 let _messageRenderWindowSid=null;
 let _messageRenderWindowSize=MESSAGE_RENDER_WINDOW_DEFAULT;
@@ -384,7 +394,7 @@ let _messageVirtualHeightCache=[];
 let _messageVirtualHeightCacheEntries=[];
 let _messageVirtualHeightCacheLen=0;
 let _messageVirtualHeightCacheSrc=null;
-let _messageVirtualEstimatedRowHeight=MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHT;
+let _messageVirtualEstimatedRowHeight=_messageVirtualDefaultHeightForRole('default');
 let _messageVirtualScrollRaf=0;
 let _messageVirtualWindowKey='';
 let _messageVirtualMeasurementCycleKey='';
@@ -403,7 +413,7 @@ function _clearMessageVirtualHeightCache(){
   _messageVirtualHeightCacheEntries=[];
   _messageVirtualHeightCacheLen=0;
   _messageVirtualHeightCacheSrc=null;
-  _messageVirtualEstimatedRowHeight=MESSAGE_VIRTUAL_DEFAULT_ROW_HEIGHT;
+  _messageVirtualEstimatedRowHeight=_messageVirtualDefaultHeightForRole('default');
   _messageVirtualWindowKey='';
   _messageVirtualMeasurementCycleKey='';
   _messageVirtualMeasurementRetryCount=0;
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Reading `static/ui.js` at the PR HEAD (`2b75b55a`) against `origin/master`, this is a clean partial-mitigation and the role classifier is sensible. The `_messageVirtualWindow` change correctly routes uncached rows through `_messageVirtualDefaultHeightForRole` via the new `roleForIdx` callback:

```js
const rowHeightFor=(idx)=>{
  const cached=Number(heights[idx]);
  if(Number.isFinite(cached)&&cached>0) return cached;
  return roleForIdx?Math.max(1,_messageVirtualDefaultHeightForRole(roleForIdx(idx))):defaultHeight;
};
```

and `_messageVirtualRoleForEntry` (`static/ui.js:625`) detects tool-call rows through `m.tool_calls`, `tool_use` content blocks, and `_partial_tool_calls`, which covers both finished and streaming tool turns. Good.

## One gap worth closing (greptile flagged it, and I confirmed it)

The window calc is now role-aware, but the two **sibling offset calculators** still use the flat running-average for uncached rows, so they disagree with the window for unmeasured tool-call rows:

`_messageVirtualPrependedHeightDelta` at `static/ui.js:669`:
```js
const cached=Number(_messageVirtualHeightCache[i]);
total+=(Number.isFinite(cached)&&cached>0)?cached:_messageVirtualEstimatedRowHeight;
```

`_messageVirtualScrollTopForVisibleIdx` at `static/ui.js:687`:
```js
const cached=Number(_messageVirtualHeightCache[i]);
offset+=(Number.isFinite(cached)&&cached>0)?cached:_messageVirtualEstimatedRowHeight;
```

`_messageVirtualEstimatedRowHeight` starts at 140 (`:397`) and only converges toward a measured mean after rows are measured (`:773`). So before measurement, the window estimates an uncached tool-call row at **400px** while these two helpers estimate the same row at **140px** — a 260px-per-row gap. `_messageVirtualScrollTopForVisibleIdx` feeds `container.scrollTop` at `:958` and `:9800` (programmatic jump-to-row), and `_messageVirtualPrependedHeightDelta` feeds prepend-history anchoring. Both can land at the wrong offset for tool-heavy transcripts before the cache warms — which is exactly the #4346 repro shape (≈90 tall tool rows).

## Suggested fix

Both loops already have the visible-index list in hand, so they can thread the same role lookup the window uses. Sketch for `_messageVirtualScrollTopForVisibleIdx` (`:680-689`):

```js
for(let i=0;i<limit;i++){
  const cached=Number(_messageVirtualHeightCache[i]);
  if(Number.isFinite(cached)&&cached>0){ offset+=cached; continue; }
  offset+=_messageVirtualDefaultHeightForRole(_messageVirtualRoleForEntry(visWithIdx[i]));
}
```

`_messageVirtualPrependedHeightDelta` (`:667-670`) needs the same treatment, but note it iterates the raw cache without a `visWithIdx` in scope — it calls `_currentMessageVirtualWindow(visWithIdx,...)` at `:663`, so you'd reuse that local `visWithIdx` for `_messageVirtualRoleForEntry(visWithIdx[i])`. That keeps all three call sites computing the same uncached estimate and removes the divergence.

This isn't a regression versus master (both helpers used the flat estimate there too), but since the whole point of the PR is to shrink estimate error for tool rows, leaving the offset math on the old 140px estimate undercuts the win on the two paths most likely to produce a visible jump. CI is green, but the new tests assert the window path only — worth adding an offset-helper assertion so the three paths can't drift apart later.

rodboev:
Both offset helpers now use per-role defaults for uncached rows, matching the windowing function. `_messageVirtualPrependedHeightDelta` and `_messageVirtualScrollTopForVisibleIdx` call `_messageVirtualDefaultHeightForRole(_messageVirtualRoleForEntry(visWithIdx[i]))` instead of the flat `_messageVirtualEstimatedRowHeight`. New test verifies all three call sites agree: user=120, tool_call=400, assistant=160.

nesquena-hermes:
## Deep review — #4367 (per-role virtual scroll height estimates)

Reviewed at head `3c714d2f`, rebased onto current master (was 14 behind). Ran the full authoritative gate (Codex + Opus + full pytest suite) on a combined stage with #4368. Posting the review now — **not merging yet**, per maintainer request to wait for your official go-ahead before the final review + merge.

### Verdict: clean, ship-ready pending your go-ahead. One important cross-PR finding below.

**What it does (verified):** replaces the flat 140px unmeasured-row estimate with per-role fallbacks (user 120 / assistant 160 / tool_call 400 / default 140) via `_messageVirtualDefaultHeightForRole(role)` + `_messageVirtualRoleForEntry` (classifies `tool_call` from `tool_calls`/`tool_use`/`_partial_tool_calls`). The two sibling offset calculators (`_messageVirtualPrependedHeightDelta`, `_messageVirtualScrollTopForVisibleIdx`) route through the per-role defaults, closing the maintainer-flagged 260px/row gap.

**Correctness:** additive + backward-compatible — `opts.roleForIdx` is optional and falls back to the flat default when absent; the measured-height path is unchanged. Window math (start/end/topPad/bottomPad) is regression-free with the new estimates. No XSS surface (render path reads `role`/`tool_calls`/`content[].type` only; no new HTML sink).

**Gate results (combined #4367 + #4368):**
- Codex: **SAFE TO SHIP** — no regression risk from the combination.
- Opus: **SAFE TO SHIP** — window math consistent and regression-free.
- Full suite: **9383 passed**, 0 failed (21 PR tests + existing).

### ⚠️ Important: #4367 and #4368 must ship together (real interaction)

Both PRs edit `_clearMessageVirtualHeightCache`, and they interact: when combined, the cache-reset uses **#4367's** `_messageVirtualDefaultHeightForRole('default')` to reset `_messageVirtualEstimatedRowHeight`, **plus #4368's** scroll-settle-globals cleanup. Neither PR alone produces the merged function. When I staged them together, #4368's `test_clear_height_cache_resets_scroll_settle_globals` node harness failed with `ReferenceError: _messageVirtualDefaultHeightForRole is not defined` because that harness (authored against #4368's standalone base) didn't include #4367's helper.

This isn't a bug in either PR — it's a merge-order dependency. Recommendation: **ship #4367 + #4368 as one release** (the #4346 cluster), with the test harness updated to eval `_messageVirtualDefaultHeightForRole` + define the role map. I've already prepared and verified that combined stage (suite green).

### To finalize after your go-ahead
- Visual sign-off + live-drive on the initial scroll position with a transcript containing large tool-call rows (the per-role estimate's user-visible effect).

Nice, well-scoped fix. Holding for your confirmation.

nesquena-hermes:
Shipped in **v0.51.483** (Release QS) 🚀 — thanks @rodboev! Shipped together with #4368 as the joint #4346 scroll-jump fix, since they interact in `_clearMessageVirtualHeightCache`. Kept **default-off / experimental** (only runs under "Virtualize long transcripts (experimental)"). Full ship gate green (Codex + Opus + the full suite, 9386 passing); Opus confirmed it's genuinely inert when the setting is off (render byte-identical to master). Closing as merged.

PR #4368, opened 2026-06-17
fix(#4346): anchor-based scroll compensation during measurement

rodboev:
## Thinking Path

- The measurement-refresh path calls `renderMessages({preserveScroll:true})` but the DOM rebuild changes `scrollHeight` by the spacer delta; restoring the same numeric `scrollTop` after a spacer change shifts the viewport by that delta.
- The existing `_captureMessageViewportAnchor`/`_restoreMessageViewportAnchor` pair (lines 660-688) already implements row-relative position restoration, but the measurement path bypasses it.
- Wrapping the measurement-driven render in `_compensateScrollForMeasurementDelta` captures the anchor, runs the render, computes the post-render row position delta, and applies the correction, reusing existing primitives with no new DOM queries.

## What Changed

- `static/ui.js`: Added `_compensateScrollForMeasurementDelta(renderFn)` helper that captures the topmost visible `[data-msg-idx]` row anchor, invokes `renderFn`, then corrects `scrollTop` by the row-position delta. Modified `_scheduleMessageVirtualizedRender` to wrap the actual `renderMessages({preserveScroll:true})` call inside the compensation helper, ensuring the anchor is captured and compensated around the synchronous DOM mutation.
- `tests/test_issue500_message_list_virtualization.py`: Added coverage for anchor compensation, tolerance check, programmatic scroll flag lifecycle, and source-order assertions.

## Why It Matters

Rather than bounding oscillation with a rerender cap, this eliminates the scroll drift that causes visible oscillation by compensating `scrollTop` by the exact spacer-height delta after each measurement-driven rebuild. The viewport anchor stays on the same row across rerenders.

## Verification

```
pytest tests/test_tars_scroll_reset_regressions.py -v --timeout=60
```

## Risks / Follow-ups

- The compensation fires in the same `requestAnimationFrame` as the render; if layout is incomplete at that point, a nested rAF could be added at the cost of one extra frame.
- `_lastScrollTop` is updated directly to prevent the scroll handler from treating the correction as a user scroll; this mirrors the pattern at `static/ui.js:686`.
- `MESSAGE_VIRTUAL_MEASUREMENT_MAX_RERENDERS` cap remains as a safety net in case row queries fail.
- Does not address the prepend-scroll-compensation path (`_messageVirtualPrependedHeightDelta`, line 624), which modifies `scrollTop` directly and does not go through `_scheduleMessageVirtualMeasurementRefresh`.

## Upstream

Ref #4346

## Model Used

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -696,6 +696,25 @@ function _restoreMessageViewportAnchor(anchor, rawIdxDelta){
   requestAnimationFrame(()=>{ _programmaticScroll=false; });
   return true;
 }
+function _compensateScrollForMeasurementDelta(renderFn){
+  const container=$('messages');
+  if(!container) return renderFn();
+  const anchorBefore=_captureMessageViewportAnchor();
+  const scrollTopBefore=container.scrollTop;
+  renderFn();
+  if(!anchorBefore) return;
+  const row=container.querySelector(`[data-msg-idx="${anchorBefore.rawIdx}"]`);
+  if(!row) return;
+  const containerRect=container.getBoundingClientRect();
+  const rowRect=row.getBoundingClientRect();
+  const actualOffset=rowRect.top-containerRect.top;
+  const delta=actualOffset-anchorBefore.topOffset;
+  if(Math.abs(delta)<2) return;
+  _programmaticScroll=true;
+  container.scrollTop=scrollTopBefore+delta;
+  _lastScrollTop=container.scrollTop;
+  requestAnimationFrame(()=>{ setTimeout(()=>{ _programmaticScroll=false; },0); });
+}
 function _messageViewportIntersectsRenderedRow(){
   const container=$('messages');
   if(!container) return true;
@@ -771,7 +790,7 @@ function _scheduleMessageVirtualizedRender(force){
     const liveWindow=_currentMessageVirtualWindow(liveVisWithIdx,_messageVirtualKeepTailCount());
     const liveKey=_messageVirtualWindowKeyFor(liveWindow);
     if(!force&&liveKey===_messageVirtualWindowKey) return;
-    renderMessages({ preserveScroll:true });
+    _compensateScrollForMeasurementDelta(()=>{ renderMessages({ preserveScroll:true }); });
   });
 }
 
--- a/static/ui.js
+++ b/static/ui.js
@@ -389,6 +389,21 @@ let _messageVirtualScrollRaf=0;
 let _messageVirtualWindowKey='';
 let _messageVirtualMeasurementCycleKey='';
 let _messageVirtualMeasurementRetryCount=0;
+let _messageVirtualScrollActive=false;
+let _messageVirtualScrollSettleTimer=0;
+let _messageVirtualDeferredMeasurement=null;
+function _markMessageVirtualScrollActive(){
+  _messageVirtualScrollActive=true;
+  clearTimeout(_messageVirtualScrollSettleTimer);
+  _messageVirtualScrollSettleTimer=setTimeout(()=>{
+    _messageVirtualScrollActive=false;
+    if(_messageVirtualDeferredMeasurement){
+      const deferred=_messageVirtualDeferredMeasurement;
+      _messageVirtualDeferredMeasurement=null;
+      _scheduleMessageVirtualMeasurementRefresh(deferred);
+    }
+  },150);
+}
 // Cached visWithIdx array — invalidated when S.messages.length changes.
 let _visWithIdxCache=null;
 let _visWithIdxCacheLen=0;
@@ -524,6 +539,10 @@ function _messageVirtualMeasurementCycleKeyFor(windowMetrics){
   ].join(':');
 }
 function _scheduleMessageVirtualMeasurementRefresh(windowMetrics){
+  if(_messageVirtualScrollActive){
+    _messageVirtualDeferredMeasurement=windowMetrics;
+    return;
+  }
   const cycleKey=_messageVirtualMeasurementCycleKeyFor(windowMetrics);
   if(_messageVirtualMeasurementCycleKey!==cycleKey){
     _messageVirtualMeasurementCycleKey=cycleKey;
@@ -3539,6 +3558,7 @@ if(typeof window!=='undefined'){
   if(!el) return;
   let _scrollRaf=0;
   el.addEventListener('scroll',()=>{
+    _markMessageVirtualScrollActive();
     if(_programmaticScroll) return; // ignore scrolls we triggered ourselves
     cancelAnimationFrame(_scrollRaf);
     _scrollRaf=requestAnimationFrame(()=>{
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Reading `static/ui.js` at the PR HEAD (`66a9b211`) against `origin/master`, the anchor-compensation core is sound — `_compensateScrollForMeasurementDelta` (`static/ui.js:718`) captures the topmost `[data-msg-idx]` row, runs the render, and corrects `scrollTop` by the post-render row-position delta, which is the right structural fix for the spacer-shift oscillation:

```js
const actualOffset=rowRect.top-containerRect.top;
const delta=actualOffset-anchorBefore.topOffset;
if(Math.abs(delta)<2) return;
_programmaticScroll=true;
container.scrollTop=scrollTopBefore+delta;
_lastScrollTop=container.scrollTop;
```

Setting `_lastScrollTop` directly mirrors the existing anchor-restore pattern, so the scroll handler won't read the correction as a user-initiated move on the next rAF. Good.

## Ordering bug greptile flagged — confirmed

The new debounce is wired in **before** the `_programmaticScroll` guard in the scroll listener (`static/ui.js:3560-3562`):

```js
el.addEventListener('scroll',()=>{
  _markMessageVirtualScrollActive();           // fires for EVERY scroll event...
  if(_programmaticScroll) return;              // ...including the ones we just triggered
```

`_compensateScrollForMeasurementDelta` sets `_programmaticScroll=true` and assigns `container.scrollTop`, which synchronously emits a `scroll` event. That event hits the listener and calls `_markMessageVirtualScrollActive()` *before* the `_programmaticScroll` early-return, arming the 150ms settle timer. So each compensation defers the next measurement refresh by 150ms (`_scheduleMessageVirtualMeasurementRefresh` short-circuits into `_messageVirtualDeferredMeasurement` at `:542`). On a transcript that needs several measure→correct cycles to converge — the exact #4346 repro — those 150ms delays chain, slowing the very settling this PR is meant to speed up.

Fix is a one-line reorder so programmatic corrections don't count as user scroll:

```js
el.addEventListener('scroll',()=>{
  if(_programmaticScroll) return;   // move the guard above the debounce
  _markMessageVirtualScrollActive();
  ...
```

## Second issue: deferred-measurement state survives session reset

`_clearMessageVirtualHeightCache` (`static/ui.js:416`) resets the cache and cycle keys but leaves the three new module globals untouched:

```js
function _clearMessageVirtualHeightCache(){
  _messageVirtualHeightCache=[];
  ...
  _messageVirtualMeasurementRetryCount=0;
  // _messageVirtualScrollActive / _messageVirtualScrollSettleTimer /
  // _messageVirtualDeferredMeasurement are NOT cleared here
}
```

If a user switches sessions while a measurement is deferred, `_messageVirtualDeferredMeasurement` still holds the **old** session's `windowMetrics`, and the pending settle timer can fire `_scheduleMessageVirtualMeasurementRefresh(deferred)` against the new session's DOM. Recommend clearing all three in this reset path (and ideally `clearTimeout(_messageVirtualScrollSettleTimer)`), so a session swap can't replay a stale window.

Everything else looks good — the tolerance skip, the source-order test, and the deferral replay logic are all reasonable. CI green. Worth landing the reorder + reset-cleanup before merge so the debounce doesn't fight the compensation it's paired with.

rodboev:
Fixed both issues:

1. Moved the `_programmaticScroll` guard above `_markMessageVirtualScrollActive()` in the scroll listener so compensation-triggered scrolls don't arm the 150ms settle timer.

2. `_clearMessageVirtualHeightCache` now resets `_messageVirtualScrollActive`, calls `clearTimeout` on the settle timer, zeros the timer ID, and nulls `_messageVirtualDeferredMeasurement`. Session switches can no longer replay stale deferred measurements.

Tests cover both: a source-order assertion for the listener guard, and a Node eval test that verifies all three globals are cleared on cache reset.

nesquena-hermes:
## Deep review — #4368 (anchor-based scroll compensation during measurement)

Reviewed at head `ae4a081822bf`, rebased onto current master (was 14 behind). Ran the full authoritative gate (Codex + Opus + full pytest suite) on a combined stage with #4367. Posting the review now — **not merging yet**, per maintainer request to wait for your official go-ahead before the final review + merge.

### Verdict: clean, ship-ready pending your go-ahead. One important cross-PR finding below.

**What it does (verified):** `_compensateScrollForMeasurementDelta` captures an anchor row + `scrollTop`, renders, then re-pins `scrollTop` by the anchor row's measured delta (null/missing/`<2px` guards; programmatic-scroll flag managed), wrapping `renderMessages({preserveScroll:true})` in `_scheduleMessageVirtualizedRender`. Adds a 150ms scroll-active guard that defers measurement refresh while the user is actively scrolling, plus session-switch cleanup of the three scroll-settle globals.

**Correctness (Opus-verified):** scroll compensation cannot infinite-loop or strand the programmatic-scroll flag; the `<2px` and null guards prevent thrash; re-render is capped at 2. The re-push since the last review correctly fixed the scroll-listener ordering (guard before mark-active) + session-switch cleanup — both confirmed present.

**Gate results (combined #4367 + #4368):**
- Codex: **SAFE TO SHIP** — no regression risk from the combination.
- Opus: **SAFE TO SHIP** — "scroll compensation cannot infinite-loop or strand the programmatic-scroll flag, no new XSS/DoS surface." Noted the merge-induced delta (settle globals also reset on mid-render structural rebuilds) is "benign and arguably more correct."
- Full suite: **9383 passed**, 0 failed (22 PR tests + existing).

### ⚠️ Important: #4368 and #4367 must ship together (real interaction)

Both PRs edit `_clearMessageVirtualHeightCache`. Combined, the function resets `_messageVirtualEstimatedRowHeight` via **#4367's** `_messageVirtualDefaultHeightForRole('default')` while also doing **#4368's** scroll-settle-globals cleanup. Your `test_clear_height_cache_resets_scroll_settle_globals` node harness (authored against this PR's standalone base) failed when combined with `ReferenceError: _messageVirtualDefaultHeightForRole is not defined` — it didn't eval #4367's helper.

Not a bug in either PR — a merge-order dependency. Recommendation: **ship #4368 + #4367 as one release** (the #4346 cluster), with the harness updated to eval `_messageVirtualDefaultHeightForRole` + define the role map. I've prepared and verified that combined stage (suite green).

### To finalize after your go-ahead
- Visual sign-off + **live-drive** on the timing-dependent transition — specifically the mid-scroll compaction / older-history-load case (the scenario most likely to expose a scroll-jump if the compensation math is off). A static screenshot won't prove this; it needs an interactive drive.

Solid fix on a genuinely tricky timing surface. Holding for your confirmation.

nesquena-hermes:
Shipped in **v0.51.483** (Release QS) 🚀 — thanks @rodboev! Shipped together with #4367 as the joint #4346 scroll-jump fix, since they interact in `_clearMessageVirtualHeightCache`. Kept **default-off / experimental** (only runs under "Virtualize long transcripts (experimental)"). Full ship gate green (Codex + Opus + the full suite, 9386 passing); Opus confirmed it's genuinely inert when the setting is off (render byte-identical to master). Closing as merged.

PR #4369, opened 2026-06-17
feat(#4346): idle measure-ahead pre-pass for virtual scroll

rodboev:
## Thinking Path

- Height estimate errors only matter for rows never rendered; once measured, cached values win. The goal is to get real heights into the cache before those rows scroll into view.
- `requestIdleCallback` is the right primitive; pre-measurement should not compete with scroll or stream events, and the idle deadline API lets the pass yield early if the frame budget runs out.
- The hidden container approach (detached div, `visibility:hidden`, off-screen) forces layout without causing paint or FOUC. It is removed immediately after measurement, leaving no persistent DOM pollution.

## What Changed

- `static/ui.js`: Added `_messageVirtualPreMeasureRic` and `_messageVirtualPreMeasureContainer` state vars. Added `_cancelMessageVirtualPreMeasure()` helper hooked into `_clearMessageVirtualHeightCache` for proper cleanup on session switches and cache resets. Added `_messageVirtualPreMeasureIndices()` to select up to 12 uncached rows beyond the virtual window (8 forward, 4 backward). Added `_scheduleMessageVirtualPreMeasure()` that uses `requestIdleCallback` to render and measure rows into a hidden container, writing results to `_messageVirtualHeightCache`. Called from `_markMessageVirtualMeasurementsSettled` so pre-measurement fires after the current window has converged.
- `tests/test_issue500_message_list_virtualization.py`: Added coverage for index selection, cancellation, cache identity validation, and integration assertions.

## Why It Matters

Pre-warming the height cache for rows just beyond the viewport means those rows have real heights when they enter `_messageVirtualWindow`'s calculation, eliminating the estimate error before it causes a scroll delta and triggering no measurement rerenders for those rows.

## Verification

```
```

## Risks / Follow-ups

- `_renderSingleMessageRow` is a stub; the actual implementation must locate or extract the per-row HTML renderer from the existing `renderMessages` loop. If no clean single-row renderer exists, the pre-pass may need to render a small fragment and measure each row within it.
- The `deadline.timeRemaining()<4` early-exit means a single idle pass may only measure 1-2 rows on a busy frame. A follow-up could persist progress across multiple idle callbacks rather than restarting from `windowEnd` each time.
- In Safari the `setTimeout(80)` fallback fires unconditionally rather than yielding to the main thread; on slow devices this could cause a brief jank spike. A conservative alternative is skipping pre-measure entirely on Safari until `requestIdleCallback` lands.
- Pre-measurement inflates `_messageVirtualHeightCache` from a hidden container; if CSS differs (font loading, container width), measurements may diverge. The hidden container must inherit the same width as `#messages`.

## Upstream

Ref #4346

## Model Used

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -389,6 +389,8 @@ let _messageVirtualScrollRaf=0;
 let _messageVirtualWindowKey='';
 let _messageVirtualMeasurementCycleKey='';
 let _messageVirtualMeasurementRetryCount=0;
+let _messageVirtualPreMeasureRic=0;
+let _messageVirtualPreMeasureContainer=null;
 // Cached visWithIdx array — invalidated when S.messages.length changes.
 let _visWithIdxCache=null;
 let _visWithIdxCacheLen=0;
@@ -399,6 +401,7 @@ function clearVisibleMessageRowCache(){
   _visWithIdxCacheSrc=null;
 }
 function _clearMessageVirtualHeightCache(){
+  _cancelMessageVirtualPreMeasure();
   _messageVirtualHeightCache=[];
   _messageVirtualHeightCacheEntries=[];
   _messageVirtualHeightCacheLen=0;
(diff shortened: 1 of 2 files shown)


rodboev:
Withdrawing this PR. Live browser testing on a 949-message session showed the idle pre-measurement adds no measurable benefit during scroll and slightly worsens snap-back count (8 vs 6 baseline). PR #4367 (per-role height estimates) solves the estimation problem more effectively by making initial estimates accurate rather than pre-measuring during idle.

PR #4372, opened 2026-06-17
feat: attach large composer text pastes

santastabber:
## Thinking Path

- Hermes WebUI aims for near-CLI parity in a browser while keeping the composer calm and usable.
- Large pasted text blocks currently land directly in the textarea, making the chat composer unwieldy.
- The app already has a safe pending-file tray and upload flow for chat attachments.
- This PR reuses that existing attachment path for large plain-text pastes instead of adding new backend routes or dependencies.
- The image paste behavior from the recent image+text fix remains first-priority, so text copied alongside an image is still handled by the browser while the image attaches.

## What Changed

- Added large plain-text paste detection in the chat composer:
  - 4,000+ characters, or
  - 40+ lines.
- Converts matching plain-text pastes into timestamped `pasted-text-*.md` `File` objects.
- Adds those files through the existing `addFiles()` pending attachment tray.
- Leaves normal/small text paste behavior untouched.
- Preserves image paste and image+text paste behavior.
- Adds regression coverage for the new large-text path and updates the existing image+text paste test to allow the new text-only large-paste branch.
- Adds a changelog entry.

## Why It Matters

Long pasted logs, documents, transcripts, and snippets no longer flood the composer textarea. Users get the same kind of lightweight attachment-chip experience that other AI chat UIs use for pasted blocks, while Hermes WebUI continues to reuse its existing upload and attachment pipeline.

## Verification

- `python3 -m py_compile api/upload.py server.py bootstrap.py`
- `./scripts/test.sh tests/test_large_text_paste_attachment.py -q` → `5 passed`
- `./scripts/test.sh tests/test_1620_paste_text_with_image.py tests/test_large_text_paste_attachment.py tests/test_tls_support.py::TestTLSEndToEnd::test_tls_startup_failure_fallback_to_http tests/test_v050259_sessiondb_fd_leak.py::test_session_db_close_is_idempotent -q` → `14 passed`
- `./scripts/test.sh` → `9268 passed, 92 skipped, 1 xfailed, 2 xpassed, 8 warnings, 16 subtests passed`
- Manual browser check in an isolated WebUI state:
  - short plain-text paste was not intercepted;
  - 4,100-character paste created a `.md` pending file with full text content;
  - 41-line paste created a `.md` pending file;
  - image+text paste still attached the image and did not call `preventDefault()`.

## Risks / Follow-ups

- The thresholds are intentionally simple constants. If maintainers want a setting later, this can be surfaced in Preferences, but this PR avoids new UI/config scope.
- The generated attachment uses Markdown MIME/type and `.md` extension so pasted docs/logs remain readable in existing file flows.
- This change is composer-only and reuses the existing upload size cap via `addFiles()` / `uploadPendingFiles()`.

## Contract Routing

Task type: UI/UX behavior improvement.
Touched areas:
- `static/boot.js` composer paste handler
- composer paste regression tests
- `CHANGELOG.md`
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/UIUX-GUIDE.md`
Scope boundaries:
- No backend route changes.
- No upload storage contract changes.
- No new dependencies, build step, framework, or persistent settings.
- Focused regression tests.
- Full test runner.
- Browser-level paste behavior check.

## Model Used

AI-assisted: Nous Research Hermes Agent using OpenRouter `openai/gpt-5.5`, with local file, terminal, browser, and GitHub workflow tooling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- **Large plain-text pastes in the composer now become `.md` attachments instead of flooding the textarea.** Pasting roughly 4,000+ characters or 100+ lines of plain text attaches a timestamped `pasted-text-*.md` file through the existing upload tray, while normal short text pastes, oversized pastes, and image pastes keep their existing behavior.
+
 ## [v0.51.486] — 2026-06-18 — Release QV (archived cron runs stay hidden on state.db re-projection)
 
 ### Fixed
--- a/static/boot.js
+++ b/static/boot.js
@@ -1569,26 +1569,66 @@ document.addEventListener('keydown',async e=>{
     }
   }
 });
+const LARGE_TEXT_PASTE_CHAR_THRESHOLD=4000;
+const LARGE_TEXT_PASTE_LINE_THRESHOLD=100;
+function _largeTextPasteLineCount(text){
+  const value=String(text||'');
+  const lines=value.split('\n');
+  return value.endsWith('\n')?lines.length-1:lines.length;
+}
+function _shouldAttachLargePastedText(text){
+  const value=String(text||'');
+  if(!value.trim())return false;
+  return value.length>=LARGE_TEXT_PASTE_CHAR_THRESHOLD || _largeTextPasteLineCount(value)>=LARGE_TEXT_PASTE_LINE_THRESHOLD;
+}
+function _largeTextPasteFileName(now){
+  const d=new Date(now||Date.now());
+  const stamp=d.toISOString().replace(/[:.]/g,'-').replace('T','_').replace('Z','');
+  const existing=new Set((S.pendingFiles||[]).map(f=>f&&f.name).filter(Boolean));
+  let name=`pasted-text-${stamp}.md`;
+  for(let i=2;existing.has(name);i++)name=`pasted-text-${stamp}-${i}.md`;
+  return name;
+}
+function _largeTextPasteFile(text,now){
+  const name=_largeTextPasteFileName(now||Date.now());
+  return new File([String(text||'')],name,{type:'text/markdown;charset=utf-8'});
+}
+function _largeTextPasteFitsUploadLimit(file){
+  return !(file&&typeof MAX_UPLOAD_BYTES==='number'&&file.size>MAX_UPLOAD_BYTES);
+}
+function _attachLargePastedText(file){
+  addFiles([file]);
+  if(typeof setStatus==='function')setStatus(t('text_pasted')+file.name);
+  return file;
+}
 $('msg').addEventListener('paste',e=>{
   const items=Array.from(e.clipboardData?.items||[]);
   // Extract image items (kind==='file' filter avoids misclassifying text/html
   // with embedded data URIs as images).
   const imageItems=items.filter(i=>i.kind==='file'&&i.type.startsWith('image/'));
-  if(!imageItems.length)return;
-  // If text is also present (common when copying images from browsers, Notes,
-  // Slack, etc.), let the browser paste the text normally AND attach the image.
-  // Only preventDefault when the clipboard is image-only (true screenshot paste).
-  const hasText=items.some(i=>i.kind==='string'&&(i.type==='text/plain'||i.type==='text/html'));
-  if(!hasText)e.preventDefault();
-  const pasteTs=Date.now();
-  const files=imageItems.map((i,idx)=>{
-    const blob=i.getAsFile();
-    const ext=i.type.split('/')[1]||'png';
-    const suffix=imageItems.length>1?`-${idx+1}`:'';
-    return new File([blob],`screenshot-${pasteTs}${suffix}.${ext}`,{type:i.type});
-  });
-  addFiles(files);
-  setStatus(t('image_pasted')+files.map(f=>f.name).join(', '));
+  if(imageItems.length){
+    // If text is also present (common when copying images from browsers, Notes,
+    // Slack, etc.), let the browser paste the text normally AND attach the image.
+    // Only preventDefault when the clipboard is image-only (true screenshot paste).
+    const hasText=items.some(i=>i.kind==='string'&&(i.type==='text/plain'||i.type==='text/html'));
+    if(!hasText)e.preventDefault();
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Reading `static/boot.js` at the PR HEAD (`230d36cf`) against `origin/master`, the refactor is clean: the image branch now returns early inside `if(imageItems.length){…}` and the new text-only branch runs only when no image items are present, so the #1620 image+text path is preserved. `_shouldAttachLargePastedText`, the trailing-newline-correct `_largeTextPasteLineCount`, and the dedup loop in `_largeTextPasteFileName` all look right, and the i18n parity test (`text_pasted` count == `image_pasted` count == 13) is a nice guard. Two things worth a look before merge.

## 1. Oversize paste is lost from both composer and tray (concrete edge case)

The new branch calls `preventDefault()` *before* handing the text to `addFiles()`:

```js
const plainText=e.clipboardData?.getData('text/plain')||'';
if(!_shouldAttachLargePastedText(plainText))return;
e.preventDefault();
_attachLargePastedText(plainText);
```

But `addFiles` (`static/ui.js:13790`) silently drops anything over the upload cap:

```js
function addFiles(files){
  for(const f of files){
    if(f&&f.size>MAX_UPLOAD_BYTES){_showUploadTooLarge(f);continue;}
    if(!S.pendingFiles.find(p=>p.name===f.name))S.pendingFiles.push(f);
  }
  renderTray();
}
```

`MAX_UPLOAD_BYTES` defaults to 20MB (`static/ui.js:16`). When a paste exceeds it, `preventDefault()` has already suppressed the textarea insert, then `addFiles` hits the `continue` and never pushes the file — so the user's clipboard content vanishes from *both* the composer and the tray, leaving only a `❌ upload failed — too large` status. Re-pasting just reproduces it. Note the cap is on `File.size` (UTF-8 bytes), so multibyte text trips it sooner than the char count suggests.

Suggested fix: gate `preventDefault()` on the file actually being acceptable, e.g. compute the `File` first and only intercept when `file.size<=MAX_UPLOAD_BYTES`, otherwise let the browser paste normally:

```js
const file=_buildLargePastedTextFile(plainText);   // factor out the File creation
if(file.size>MAX_UPLOAD_BYTES) return;              // let the browser handle it
e.preventDefault();
_attachLargePastedText(file);
```

That keeps the worst case (huge paste) behaving exactly as it does on `master` today rather than swallowing the content.

## 2. 40-line threshold is fairly aggressive

`LARGE_TEXT_PASTE_LINE_THRESHOLD=40` (with the 4000-char OR) means a moderately sized code block or list — well under 4000 chars — gets pulled out of the composer into a `.md` attachment. Pasting a ~45-line snippet you intended to read/trim inline before sending would be surprising, and once attached there's no inline preview of the content in the tray. The char threshold is reasonable; the line threshold is the one most likely to fire on content users still want to see in the textarea. Worth considering bumping it (80–100 lines) or dropping the line trigger in favor of the char count alone. The PR's stance against new config is fine — this is just about picking the constant.

Everything else (branch ordering test, dedup, i18n coverage) is solid. Worth landing the oversize guard at minimum; the threshold is a judgement call for the maintainer.

nesquena-hermes:
Thanks @santastabber — this is a genuinely useful feature (big pastes → attachment is a pattern ChatGPT and others use), the refactor cleanly preserves the #1620 image+text path, and the i18n parity guard is a nice touch. Two things to address before it can ship, then I'm happy to take it through the full gate.

## 1. [must-fix] Oversize paste is lost from both the composer AND the tray

The new branch calls `preventDefault()` before handing the text to `addFiles()`:

```js
const plainText=e.clipboardData?.getData('text/plain')||'';
if(!_shouldAttachLargePastedText(plainText))return;
e.preventDefault();
_attachLargePastedText(plainText);
```

But `addFiles` (`static/ui.js:~13790`) silently drops anything over `MAX_UPLOAD_BYTES` (20MB default):

```js
if(f&&f.size>MAX_UPLOAD_BYTES){_showUploadTooLarge(f);continue;}
```

So when a paste exceeds the cap, `preventDefault()` has already suppressed the textarea insert, then `addFiles` hits the `continue` and never pushes the file — the user's clipboard content vanishes from *both* the composer and the tray, leaving only an "upload failed — too large" status. Re-pasting just reproduces it. (Note the cap is on `File.size` = UTF-8 bytes, so multibyte text trips it sooner than the char count suggests.)

**Fix:** build the `File` first and only intercept when it's within the cap; otherwise let the browser paste normally so the worst case behaves exactly as it does on `master` today:

```js
const file=_buildLargePastedTextFile(plainText);   // factor out the File creation
if(file.size>MAX_UPLOAD_BYTES) return;              // don't preventDefault — let the browser paste
e.preventDefault();
_attachLargePastedText(file);
```

A regression test for the over-cap path (assert the paste is NOT intercepted / content preserved) would lock this in.

## 2. [please reconsider] The 40-line trigger is too aggressive

`LARGE_TEXT_PASTE_LINE_THRESHOLD=40` means a moderately sized code block or list — well under 4000 chars — gets pulled out of the composer into a `.md` attachment with no inline preview. Pasting a ~45-line snippet you intended to read/trim inline before sending would be surprising. The 4000-char trigger is reasonable; the line trigger is the one most likely to fire on content people still want to see in the textarea.

Could you **raise the line threshold to ~100, or drop the line trigger entirely in favor of the char count**? Either keeps the feature firing on genuinely large pastes while leaving normal code blocks inline. (Your stance against adding new config here is fine — this is just about picking the constant.)

Everything else (branch ordering, the trailing-newline-correct line count, the dedup filename loop, the i18n parity test) looks solid. Marking `changes-requested`; ping me once these two are in and I'll run the full Codex + Opus + suite gate and a screenshot pass.

santastabber:
Done. Let me know if anything else

nesquena-hermes:
Shipped in **v0.51.490** (Release QZ) 🚀 — thanks @santastabber, and thanks for the clean rework! Both bounce items are fixed: the line threshold is now 100 and an oversize (>20MB) paste falls through to a normal inline paste instead of vanishing. Large pastes attach as a timestamped `.md`. Codex SAFE (XSS-safe, #1620 path intact), suite 9411. Closing as merged.

Issue #2543, opened 2026-05-18
Add named context blocks for multiple text selections


AgatElite:
Following up from #2481.

The new “Reply with selection” feature is already very useful, especially with multiple selected passages. I’d like to propose a more productized version for handling multiple selections in the composer.

## Use case

A common scenario is selecting several different parts of a conversation, then writing one follow-up message that refers to each of them separately.

For example:

- select one paragraph from an earlier assistant message
- select a different paragraph from another message
- select an error message or code snippet
- then ask something like: “Compare the first two, and explain how the error relates to them.”

With the current blockquote approach, this works, but multiple selections are merged together visually, so it becomes hard to tell where one selection starts or ends.

## Proposed UX

Instead of inserting all selections directly as one long quoted block, the composer could show them as separate context blocks above or inside the input area.

For example:

- Selection 1
- Selection 2
- Error message

Each block could:

- show a short preview by default
- expand to show the full selected text
- be removable before sending
- be renameable, e.g. “Selection 1” → “error message” or “install instructions”

## Referring to selections

A nice follow-up would be the ability to reference these blocks in the prompt, for example with `@selection-1` or `@error-message`, ideally using autocomplete.

That would make it easier to write prompts like:

> Compare @selection-1 and @selection-2, then explain how @error-message relates to them.

## Possible MVP

To keep the first version simpler, the UI could manage the selections as separate blocks in the composer, but still flatten them to regular Markdown blockquotes when the message is sent.

So the model/provider compatibility could stay the same initially, while the user gets a much clearer interface for managing multiple selections.

Later, this could evolve into structured context if that makes sense.

nesquena-hermes:
Read `static/messages.js:80-178` (the selected-text-reply feature shipped via #2481) and the i18n entries at `static/i18n.js:47`. Solid starting point for thinking about what an MVP looks like.

## How the feature works today

The current implementation is straightforward: selection -> blockquote flattening at send time, then appended to the composer textarea:

```js
// static/messages.js:93-98
function _formatSelectedTextReplyQuote(text){
  const normalized=String(text||'').replace(/\r\n?/g,'\n').replace(/\n{3,}/g,'\n\n').trim();
  if(!normalized)return '';
  return normalized.split('\n').map(line=>`> ${line}`).join('\n');
}
```

```js
// static/messages.js:100-113
function _appendSelectedTextReplyToComposer(text){
  const composer=(typeof $==='function'&&$('msg'))||document.getElementById('msg');
[...]

PR #4380, opened 2026-06-17
feat: render selected context as quote cards

santastabber:
## Thinking Path

- `Reply with selection` should feel like composed reply context, not raw Markdown dumped into the textarea or thread.
- A textarea cannot render rich context well, so pending selected text belongs in a quiet composer card.
- After send, the same selected-context payload should remain readable in the thread instead of showing `**Name:**` / `> quote` syntax literally.
- The model-facing payload stays plain text/Markdown-compatible; the browser render gets the richer visual treatment.

## What Changed

- Replaced the tiny pending selected-context chips with readable quote cards above the composer.
- Each pending card includes:
  - a subtle accent rail,
  - editable context name,
  - clipped multi-line quote preview,
  - contextual remove control.
- Pending card content is built with DOM nodes and `textContent`, avoiding user-content interpolation through `innerHTML`.
- Rename is accessible by click, Enter, Space, and F2.
- Rename inputs are labelled, length-bounded, and restore focus to the card after commit/cancel.
- Remove buttons include the context name in their aria label.
- The card stack is capped and scrollable so multiple selected contexts do not consume the whole composer area.
- Card width aligns with the widened composer at large desktop breakpoints.
- Sent user messages now render generated selected-context payloads as semantic styled blocks:
  - `<figure class="sent-selection-context">`
  - `<figcaption>` for the context label
  - `<blockquote>` for the selected passage
- Normal user text/Markdown remains escaped; only the generated labelled-blockquote shape gets the styled treatment.
- Context labels and quote text are HTML-escaped in the sent-thread renderer.
- Updated i18n keys, changelog, and selected-text regression tests.

## Why It Matters

Selected text context is easier to read, rename, remove, and trust before sending. After sending, the thread remains readable too: the user’s actual reply stays normal, and the selected context appears as a clean referenced block instead of raw Markdown syntax.

## Verification

- `node --check static/ui.js`
- `./scripts/test.sh tests/test_issue2481_selected_text_reply.py tests/test_selected_context_user_render_runtime.py tests/test_issue2543_named_context_session_switch.py -q` → `12 passed`
- Browser verification in isolated WebUI state:
  - pending selected text renders as a quote card above the composer;
  - textarea remains focused on the user’s own reply instead of receiving raw context text;
  - multiple cards use a bounded scrollable stack;
  - rename works by keyboard and restores focus;
  - remove/rename controls have contextual accessible labels;
  - send flush preserves the full selected text, not just the clipped preview;
  - sent user bubbles render selected context as a styled `figure`/`blockquote` block;
  - sent user bubbles do not show raw `**Name:**` or `> quote` syntax for generated selected context.
- Independent implementation-quality review completed; follow-up fixes were applied for stack height, touch targets, rename accessibility/focus restoration, wide-screen alignment, parser/name consistency, semantic sent-thread markup, behavioral tests, and blocked-send guard ordering.
- Independent public-upload/sterility review completed; no committed-diff upload blockers found.

## Risks / Follow-ups

- The visual preview intentionally clips long selections; the full selected text is still sent and displayed after send.
- Existing DOM ids/function names retain “chips” terminology in a few places for compatibility with the existing named-context plumbing.
- No backend routes, upload storage, persistent settings, or send payload schema are changed.
- The sent-thread renderer intentionally recognizes the generated `**Name:**` + `> quote` shape in user bubbles. User content remains escaped, but manually typed text with that exact shape will also get the same visual treatment.

## Contract Routing

Task type: UI/UX behavior improvement.

Touched areas:
- `static/messages.js` selected-text / named-context composer behavior
- `static/ui.js` sent user-message selected-context rendering
- `static/style.css` composer and sent-thread selected-context styling
- `static/i18n.js` selected-context labels
- selected-text regression tests
- `CHANGELOG.md`

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/UIUX-GUIDE.md`

Scope boundaries:
- No backend endpoint changes.
- No send payload/schema changes.
- No new dependencies, build step, framework, or persistent settings.

## Disclosure

AI-assisted implementation and review support was used. No secrets or private runtime state are required by this change.


santastabber:
Updated in e63929cf: pending selections are now included in the concurrent queued-send path via _composerTextWithPendingSelections(), and the queued path clears the composer/cards consistently after queueing. Also changed the pending-card tooltip to expose the full selected text instead of the clipped preview. Focused verification: node --check static/messages.js static/ui.js and ./scripts/test.sh tests/test_issue2481_selected_text_reply.py tests/test_selected_context_user_render_runtime.py tests/test_issue2543_named_context_session_switch.py -q => 13 passed.

nesquena-hermes:
Pulled `e63929cf` into a read-only worktree and read the three render-path changes against `origin/master`. This is clean — the queue-path fix is correct and the sent-thread renderer is XSS-safe. One low-severity heuristic note below, not a blocker.

## Queue-path fix — correct

The real bug the prior round left open was that the concurrent (`_sendInProgress`) branch read the flushed textarea instead of the pending blocks. The new split fixes it: `send()` no longer flushes up front, and the queue branch now captures via `_composerTextWithPendingSelections()` (`static/messages.js`):

```js
if (_sendInProgress) {
    const _text=_composerTextWithPendingSelections().trim();
    ...
    queueSessionMessage(_targetSid,{text:_text, ...});
    _clearComposerAfterQueuedSelectionSend();
```

`_composerTextWithPendingSelections()` builds `current + blocks` without mutating the composer, and `_clearComposerAfterQueuedSelectionSend()` clears textarea + `_pendingSelections` atomically after capture. The normal path flushes once (`_flushSelectionBlocksToComposer()` → re-read `text=$('msg').value.trim()`), so blocks land exactly once with no double-append. The empty-guard correctly gained `!_pendingSelections.length` so a selection-only send isn't dropped.

## Sent-thread renderer — XSS-safe

`_sentSelectionContextBlockHtml` (`static/ui.js:298`) routes both fields through `esc()`:

```js
return `<figure class="sent-selection-context" data-selected-context="1"><figcaption ...>${esc(safeLabel)}</figcaption><blockquote ...>${esc(safeQuote)}</blockquote></figure>`;
```

and the `\x00UC\x00` stash tokens are immune to `esc()` (which only rewrites `& < > " '`, `static/ui.js:269`), so the restore at `s.replace(/\x00UC(\d+)\x00/g,...)` is sound. Ordering is right too: fenced code is stashed **first** (`static/ui.js:340`, the `\x00UF\x00` pass) before `_stashUserSelectedContextBlocks` runs, so a `**X:**` + blockquote inside a code fence stays literal rather than being mis-rendered as a card. Composer cards are built entirely with DOM APIs + `textContent` (no `innerHTML` for user content), and `s.text` only reaches HTML via `title`/`textContent`. Good.

## One heuristic note (low severity, your call)

`_stashUserSelectedContextBlocks` keys purely on the textual shape — a label line matching `/^\*\*([^\n]{1,200}):\*\*\s*$/` immediately followed by `>` lines:

```js
const labelMatch=lines[i].match(/^\*\*([^\n]{1,200}):\*\*\s*$/);
...
while(j<lines.length&&/^>/.test(lines[j])){ quoteLines.push(...); j++; }
```

There's no provenance sentinel distinguishing "this came from Reply-with-selection" from "the user hand-typed this exact shape." So a user who manually types `**Note:**` on its own line followed by a `> blockquote` will now get the styled `figure.sent-selection-context` card instead of plain escaped markdown. It's purely cosmetic (still escaped, no security impact) and arguably a nice upgrade, but it's an undocumented side effect on all user messages, not just selection replies. If you want to scope it strictly to real selections, the persisted block would need a marker (e.g. a zero-width sentinel or a distinct fence) the matcher checks for — though that adds round-trip complexity. Acceptable as-is IMO; just flagging so it's a known behavior.

i18n looks complete (all three new keys present 13×, one per locale). Nice work on the keyboard a11y (Enter/Space/F2 rename + `requestAnimationFrame` focus restoration) and the 44px coarse-pointer touch targets. LGTM pending the maintainer's gate.

santastabber:
Updated in 90b5a557: addressed the low-severity heuristic note by adding an explicit internal selected-context marker to generated Reply-with-selection payloads and requiring that marker before rendering sent-thread quote cards. Manually typed `**Note:**` + `> quote` now remains ordinary escaped user text; generated selected context still renders as the styled figure/blockquote. Rebased on current upstream. Focused verification: node --check static/messages.js static/ui.js; ./scripts/test.sh tests/test_issue2481_selected_text_reply.py tests/test_selected_context_user_render_runtime.py tests/test_issue2543_named_context_session_switch.py -q => 14 passed. Independent A++ review: PASS/no blockers.

santastabber:
Updated in 93fd7412: fixed CI regressions from the selected-context renderer extraction tests. The sent-context parsing helper is now self-contained inside _renderUserFencedBlocks(), so existing tests that extract that function in isolation work again. Also kept the _busyInputMode static contract visible near send() while preserving runtime behavior. Local verification: node --check static/messages.js static/ui.js; ./scripts/test.sh tests/test_1062_busy_input_modes.py::TestSendBusyBranchDispatch::test_send_reads_busy_input_mode tests/test_1325_user_fenced_code.py tests/test_issue347.py::test_backslash_latex_delimiters_render_to_katex_placeholders tests/test_issue347.py::test_user_code_block_with_latex_syntax_renders_as_literal_code tests/test_issue2481_selected_text_reply.py tests/test_selected_context_user_render_runtime.py tests/test_issue2543_named_context_session_switch.py -q => 30 passed.

nesquena-hermes:
## Summary

Re-reviewed at HEAD `93fd7412` in a read-only worktree (diff vs `origin/master`). The heuristic note from my `06:22` review is properly addressed: generated Reply-with-selection payloads now carry an explicit provenance marker, and the sent-thread renderer only builds a card when that marker is present. Hand-typed `**Note:**` + `> quote` stays literal escaped text. The fix is correct, well-tested, and XSS-safe. One design tradeoff worth flagging (model-context noise), non-blocking.

## The marker fix is correct end to end

Generation side — `_formatSelectedTextReplyQuote` (`static/messages.js:588`) prepends the sentinel before the blockquote lines:

```js
return `<!-- hermes-selected-context -->\n${normalized.split('\n').map(line=>`> ${line}`).join('\n')}`;
```

Render side — `stashSelectedContextBlocks` (`static/ui.js:308-329`) now requires the marker on the line immediately after the label, before it will emit a card:

```js
const labelMatch=lines[i].match(/^\*\*([^\n]{1,200}):\*\*\s*$/);
if(!labelMatch){out.push(lines[i]);continue;}
let j=i+1;
if(lines[j]!==marker){out.push(lines[i]);continue;}   // ← no marker → plain text
```

So a user who types the exact `**X:**` + `> ...` shape by hand gets ordinary escaped markdown, while real selection replies still render as the `figure.sent-selection-context` card. `test_manual_labelled_blockquote_without_marker_stays_literal_user_text` (`tests/test_selected_context_user_render_runtime.py:97`) locks this — it asserts `class="sent-selection-context"` is absent and `&gt; manually typed quote` is present for the marker-less input. Good.

## XSS + ordering still hold

`test_sent_selected_context_renders_as_semantic_card_and_escapes_content` (`:66`) confirms both `label` and `quote` route through `esc()` — `<script>alert(1)</script>` comes out as `&lt;script&gt;…`. The code-fence test (`:88`) confirms a matching shape inside a user code fence stays literal, because the `\x00UF\x00` fence stash runs before `stashSelectedContextBlocks` (`static/ui.js:344` then `:373`). The edge-label test (`:80`) shows the label regex tolerates a user-renamed `***Evidence: alpha:**`. The marker is itself immune to `esc()` (only `& < > " '` are rewritten), so it can't be spoofed into existence by escaping.

## One tradeoff worth flagging (non-blocking)

The marker is a real part of the message body, not a render-only artifact. `_composerTextWithPendingSelections` (`static/messages.js:817`) embeds `_formatSelectedTextReplyQuote(...)` output (marker included) into the text that gets sent and persisted — I confirmed both the queue branch (`:1009`) and the normal `_flushSelectionBlocksToComposer` branch (`:832`) carry it. So `<!-- hermes-selected-context -->` now reaches the model context and the persisted transcript on every selection reply.

That's the right call for round-trip robustness (a reloaded session re-renders persisted text through `_renderUserFencedBlocks`, so a render-only sentinel wouldn't survive). But it means each selection reply injects an HTML comment line into the prompt the model sees. It's harmless for virtually all models and tiny, but it is undocumented model-facing content. If you'd rather keep it out of the model context, the alternative is a zero-width sentinel or a sidecar flag on the persisted message that the renderer checks — both add round-trip complexity. The HTML-comment approach is the pragmatic choice; just noting the prompt is no longer byte-identical to what the user typed.

## Verdict

The provenance fix does exactly what I asked for in the prior round, the test coverage maps onto manual-typing, XSS, code-fence, and edge-label cases, and the queue/normal send paths stay consistent. Static read-only approve — LGTM pending your ship gate. I did not execute any worktree code; verification is by source trace only. The only open question is whether you're comfortable with the marker living in model context + persisted history (my read: fine).

santastabber:
> It's harmless for virtually all models and tiny, but it is undocumented model-facing content

Gah. Is going to mess with one of those weird models isn't it

santastabber:
Rebased on current upstream and kept the current marker-in-payload approach as discussed. The HTML-comment marker remains the pragmatic tradeoff: it keeps persisted/reloaded sessions rendering correctly while being tiny and inert for model context. Focused verification after rebase: node --check static/messages.js static/ui.js; ./scripts/test.sh tests/test_1062_busy_input_modes.py::TestSendBusyBranchDispatch::test_send_reads_busy_input_mode tests/test_1325_user_fenced_code.py tests/test_issue347.py::test_backslash_latex_delimiters_render_to_katex_placeholders tests/test_issue347.py::test_user_code_block_with_latex_syntax_renders_as_literal_code tests/test_issue2481_selected_text_reply.py tests/test_selected_context_user_render_runtime.py tests/test_issue2543_named_context_session_switch.py -q => 30 passed.

santastabber:
Quick note after re-checking current master: the core selected-context work has already landed upstream via the context-pills/named-context implementation (#2543/#3306). So this PR is no longer introducing the main Reply-with-selection context-block behavior.\n\nWhat this PR still adds is narrower:\n\n- visual refinement of the existing pending context pills into larger quote-card previews;\n- improved rename/remove a11y/touch targets;\n- sent-thread rendering of generated selected context as a styled figure/blockquote instead of literal `**Context:**` + `> quote` text;\n- queue-path hardening around pending selected context.\n\nGiven that overlap, I’m totally fine with the maintainer/moderator deciding this should be closed/deleted as redundant rather than merged. If the preference is to keep the current upstream context pills as-is, please close this PR. If the visual/post-send polish is still wanted, this can be treated as a follow-up refinement PR on top of the already-merged context-pills feature.

nesquena-hermes:
Shipped in **v0.51.562** 🚀 — thank you @santastabber.

Your quote-cards evolution of the named-context-blocks feature landed. On review I applied one fix: selection-only replies (a card with no textarea text) couldn't be sent via the primary Send button on click/tap/mobile because the composer's content check still only looked at the textarea + files — I exposed `window._hasPendingSelections()`, folded it into `_composerHasContent()`, and wired `updateSendBtn()` into `_renderSelectionChips()` so add/remove/clear refreshes the button. Added a regression test for it.

Review trail: Codex regression gate **SAFE TO SHIP** (after the fix), Opus advisor **SAFE to ship** (XSS airtight — adversarial harness 0 survivors; the thread render is marker-gated so hand-typed lookalike Markdown doesn't become a card, and code fences stay code), full suite **9973 passed**. 3 new i18n keys verified across all 13 locales.

PR #4387, opened 2026-06-17
feat(tts): add Irodori TTS server engine

zapabob:
## Summary

- Add **Irodori TTS (server)** as a fourth TTS engine option in Settings → Preferences, alongside Browser, Edge TTS, and ElevenLabs.
- Implement `/api/tts` support for Irodori via OpenAI-compatible `POST /v1/audio/speech` (MP3 stream), resolving `base_url`, `voice`, `model`, `speed`, and optional API key from env (`IRODORI_TTS_*`, `IRODORI_API_KEY`), `~/.hermes/.env`, and Hermes agent `config.yaml` (`tts.irodori` or `tts.provider: irodori`).
- API keys are only sent to loopback hosts unless `IRODORI_TTS_ALLOW_REMOTE_API_KEY=1`.
- Wire voice mode + message read-aloud on the client (`boot.js`, `ui.js`, `panels.js`).
- Add regression tests in `tests/test_irodori_tts_engine.py`.

This PR intentionally excludes fork-only conveniences and local `_docs/` notes.

## Test plan

- [ ] `pytest tests/test_irodori_tts_engine.py`
- [ ] Settings → Preferences → TTS Engine → **Irodori TTS (server)**; verify voice list and read-aloud
- [ ] With a local Irodori server on `http://127.0.0.1:8088`, confirm `/api/tts?engine=irodori` returns MP3 audio
- [ ] Confirm remote `base_url` does not receive `Authorization` unless `IRODORI_TTS_ALLOW_REMOTE_API_KEY=1`


nesquena-hermes:
Thanks for this, @zapabob — and for the clean, well-tested implementation (the loopback-only API-key default and the full `/api/tts` wiring are nicely done). Irodori-TTS is a genuinely interesting model.

I'm going to close this one, but with a concrete path to get the capability in — not a rejection of the idea.

**The shape is the issue, not the engine.** Hermes already has a generic extension point for exactly this: the agent core supports `tts.providers.<name>` with `type: command` (a generic command-type TTS provider), and the WebUI honors any provider wired that way. Adding Irodori as a *hard-coded 4th named builtin* (with dedicated `IRODORI_TTS_*` env vars and `irodori`-specific resolution) means the project takes on per-vendor curation + maintenance surface for one specific Japanese model — and the next OpenAI-compatible TTS server would need the same hard-coded treatment again.

**The more general (and in-scope) version:** wire OpenAI-compatible `/v1/audio/speech` TTS through the generic provider path so *any* such server — Irodori, LocalAI, or anything else speaking that API — works by config alone, no per-vendor code. That:
- gets Irodori working for you today (point the generic provider at your local server),
- benefits every other OpenAI-compatible TTS backend at once,
- keeps behavioral config in `config.yaml` rather than new `HERMES_*`/`IRODORI_*` env vars (the core convention — `.env` is for secrets only), and
- is the kind of "extend the generic surface, don't special-case a vendor in core" change the project actively wants.

If you'd like to take a run at a generic "OpenAI-compatible TTS server" provider option (vs. hard-coding Irodori), that'd be very welcome — happy to review it. Closing this version as out-of-scope in its current hard-coded shape. Thanks again for the contribution and the careful security defaults.

Issue #893, opened 2026-04-23
Bug: Stop_generation button deletes LLM response


HenkieTenkie62:
Clicking the stop generation button throws away the already generated part in chat.
Seems a bit wasteful when the output is right for the most part but just needs a little steering.

Is this intended?

nesquena-hermes:
This is a known UX tradeoff in the current implementation rather than a true data-loss bug — the stop action aborts the stream and discards the partial response. The intent was simplicity: a clean state after stop.

The better behavior (preserve partial output, let the user keep what was generated) is a reasonable improvement request. The implementation path would be:
- On stop, finalize the current partial response as a completed message rather than clearing it
- Keep the user's turn in the chat history so they can continue or steer

Worth tracking as a UX enhancement. The "wasteful" framing in the issue is fair — partial responses are often mostly correct and throwing them away forces the user to re-prompt.

HenkieTenkie62:
Okay thanks. Well it's especially annoying in large multi step requests that steer off in the wrong direction or need a little extra hint.
In my case i occasionally see 40k output dissappear so yeah 😅

nesquena-hermes:
This is a known behavior gap — the current cancel implementation aborts the stream and discards the partial response rather than preserving what was already generated.

**Root cause:** `attachLiveStream()` in `messages.js` handles the `cancel` SSE event by clearing INFLIGHT and either fetching the server-side session state or falling back to appending a `"*Task cancelled.*"` message locally. The partial streamed tokens are already in `assistantText` locally but the cancel path clears them without saving.

**Fix direction:** On cancel, the partial response that was already streamed should be preserved as an assistant message (possibly marked `[cancelled]`). This requires:
1. On `cancel` event: save `assistantText` as a partial assistant message before clearing
2. Or: the server-side session should have the partial content persisted at cancel time

This is a Milestone 1 fix — partial response loss is a real UX cost for long runs. Tagging `sprint-candidate`.

PR #4390, opened 2026-06-18
fix: mirror cancelled-turn messages to context_messages to preserve _recovered/_error flags (#4283)

kaishi00:
## Problem

After an interrupted turn (cancel, gateway restart, tool iteration limit), the interrupted turn's user message gets prepended to **every subsequent turn** — not just the next one. The duplication persists indefinitely until a new session (`/new`) is started. This was reported in #4283 and a partial fix was applied (filtering `_recovered` user messages in `_sanitize_messages_for_api` and `_api_safe_message_positions`), but the bug persists.

## Root Cause

The existing `_recovered`/`_error` filters in `_sanitize_messages_for_api()` and `_api_safe_message_positions()` work correctly **when the flags are present**. The problem is that the flags don't survive the `state.db` round-trip:

1. **Turn interrupted** → `_persist_cancelled_turn()` calls `_materialize_pending_user_turn_before_error()` which appends the stale user message to `session.messages` with `_recovered=True`, then appends a cancelled assistant marker with `_error=True`.

2. **`session.save()`** writes everything to `state.db` — but the `messages` table has no `_recovered` or `_error` columns. These are in-memory dict keys only. **The flags are lost.**

3. **Next turn starts** → `reconciled_state_db_messages_for_session(prefer_context=True)` builds context:
   - Gets `session.context_messages` (from the **previous successful turn** — doesn't include the cancelled turn)
   - Gets `state_messages` from `state.db` (includes the cancelled turn messages, **without flags**)
   - `state_db_delta_after_context()` finds the cancelled turn messages as deltas (in state.db but not in context_messages)
   - Merges them into context **without** `_recovered`/`_error` flags

4. **`_sanitize_messages_for_api()`** checks for `_recovered` — but it's gone — so the stale message passes through to the model on every subsequent turn.

The gap: `_persist_cancelled_turn()` updates `session.messages` but **never updates `session.context_messages`**. The cancelled turn's messages only exist in state.db (without flags) and in `session.messages` (with flags, but `session.messages` isn't used for model-facing context when `prefer_context=True`).

## Fix

After `_persist_cancelled_turn()` finishes appending to `session.messages`, mirror the newly appended messages to `session.context_messages`. This ensures:

- `context_messages` has the messages **with** `_recovered`/`_error` flags intact
- `state_db_delta_after_context()` finds no delta for these messages (they're already in both context_messages and state.db)
- `_sanitize_messages_for_api()` correctly filters them out using the preserved flags
- The agent never sees the stale message

The change is 13 insertions, 0 deletions — purely additive. It captures the `session.messages` length before any appends, then after all appends are done, copies the new tail to `context_messages` (if it's a list).

## Testing

- All 482 existing cancel/recovery/replay/persist tests pass (1 unrelated profile-cookie auth failure)
- The fix is backward-compatible: if `context_messages` is `None` or not a list, the mirroring is skipped (same behavior as before)

Refs: #4283

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -1163,6 +1163,7 @@ def _persist_cancelled_turn(session, *, message: str = 'Task cancelled.') -> Non
     later unwind through the silent-failure or exception path. Those paths must
     not append a misleading provider no-response error after an explicit cancel.
     """
+    _msg_count_before = len(getattr(session, 'messages', None) or [])
     _materialize_pending_user_turn_before_error(session)
     session.active_stream_id = None
     session.pending_user_message = None
@@ -1178,6 +1179,18 @@ def _persist_cancelled_turn(session, *, message: str = 'Task cancelled.') -> Non
             'provider_details_label': 'Cancellation details',
             'timestamp': int(time.time()),
         })
+    # Mirror appended messages to context_messages so that _recovered/_error
+    # flags survive the state.db round-trip.  Without this, the next turn's
+    # reconciled_state_db_messages_for_session(prefer_context=True) finds the
+    # stale messages as state.db deltas (without flags), and
+    # _sanitize_messages_for_api cannot filter them — causing the interrupted
+    # turn's user message to be prepended to every subsequent turn (#4283).
+    ctx = getattr(session, 'context_messages', None)
+    if isinstance(ctx, list):
+        new_msgs = (getattr(session, 'messages', None) or [])[_msg_count_before:]
+        for msg in new_msgs:
+            if isinstance(msg, dict):
+                ctx.append(dict(msg))
 
 
 def _cleanup_ephemeral_cancelled_turn(session) -> None:
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -3406,6 +3406,17 @@ def _sanitize_messages_for_api(messages, *, cfg: dict = None):
         # API 400 errors on strict providers (empty assistant content).
         if msg.get('_partial') and not str(msg.get('content') or '').strip():
             continue
+        # Skip recovered user messages from the pending_user_message repair path.
+        # After an interrupted turn (gateway restart, tool iteration limit), the
+        # recovery logic appends the stale pending message to context_messages
+        # with _recovered=True. Without this filter, that message persists in
+        # model-facing context indefinitely, causing the agent core's
+        # drop_thinking_only_and_merge_users to merge it with the current
+        # turn's user message (concatenating with \n\n), polluting every
+        # subsequent turn. The message stays in session.messages for display
+        # and state.db for persistence — only the model-facing copy is stripped.
+        if msg.get('_recovered') and msg.get('role') == 'user':
+            continue
         role = msg.get('role')
         if role == 'tool':
             tid = msg.get('tool_call_id') or ''
@@ -3470,6 +3481,8 @@ def _api_safe_message_positions(messages):
         if msg.get('_error'):
             continue
         if msg.get('_partial') and not str(msg.get('content') or '').strip():
+            continue
+        if msg.get('_recovered') and msg.get('role') == 'user':
             continue
         role = msg.get('role')
         if role == 'tool':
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
Thanks @kaishi00 — #4283 is a real, valuable bug (an interrupted turn's prompt leaking into every subsequent turn's model context is exactly the kind of compounding context-pollution worth fixing), and the core insight here is correct: the `_recovered` flag was being lost on the state.db round-trip, so the existing `_sanitize_messages_for_api` filter never matched. The `context_messages` mirroring is the right idea.

I took the `api/streaming.py` portion through the full gate (Codex + Opus + the full suite). The suite is green, but **two real issues** block it — both reproduced, with fix-specs below. (Note: I reviewed only the streaming.py commit; the bundled mobile-UX commit `afddb18a` — hamburger sizing, PWA swipe-zone, removing the standalone guard — and the image-paste commit `49a57661` (already on master) are out of scope for this fix. Please split those out; one logical change per PR.)

## 1. [CORE] Unconditional `_recovered` user strip breaks role alternation when a partial assistant was saved

When a turn is interrupted *after* the agent streamed a partial answer, that `_partial` assistant message is intentionally **kept** in API history (#893). Dropping the `_recovered` user that precedes it then leaves two adjacent assistant messages — which strict providers (DeepSeek, newer OpenAI) reject with HTTP 400. The agent core's cleanup merges adjacent *users*, not adjacent *assistants*, so nothing repairs it.

Reproduced directly against `_sanitize_messages_for_api`:
```
input : user Q1 / assistant A1 / user "Q2"(_recovered) / assistant "partial…"(_partial) / user Q3
output: ['user', 'assistant', 'assistant', 'user']   ← adjacent assistants = 400 on strict providers
```
(The pure-cancel shape is fine — there the assistant marker is `_error` and gets stripped alongside the recovered user, so they collapse together. It's specifically the interrupted-with-saved-partial shape that breaks.)

**Fix:** make the `_recovered` skip *conditional*, in both `_sanitize_messages_for_api` (~:3485) and `_api_safe_message_positions` (~:3552), via a shared predicate: only strip a `_recovered` user when it is **not** anchoring a subsequent kept assistant/`_partial` message from that interrupted turn (i.e. strip it only when what follows is dropped error/empty metadata or no assistant output).

## 2. [CORE] The mirror only covers the cancel path — the provider-error / tool-iteration-limit paths it claims to fix are still broken

The `context_messages` mirroring lives in `_persist_cancelled_turn`, but `_materialize_pending_user_turn_before_error` is also called from the provider-error path (~:7557) and the exception path (~:8532). Those are the "gateway restart / tool-iteration limit" cases the PR description explicitly claims to fix — but on an established conversation they still exhibit #4283, because the recovered turn comes back as a flagless state.db delta with no mirror.

**Fix:** move the mirroring out of `_persist_cancelled_turn` and into `_materialize_pending_user_turn_before_error`, right after `session.messages.append(recovered)` (~:5031), so all three callers are covered. Reuse the established `models.py` dedup pattern (last-8 lookback, drop `timestamp`) so repeated cancels can't duplicate or grow `context_messages` unboundedly:
```python
    session.messages.append(recovered)
    ctx = getattr(session, 'context_messages', None)
    if isinstance(ctx, list) and ctx:
        rec_text = " ".join(str(recovered.get('content') or '').split())
        if not any(
            isinstance(e, dict) and e.get('role') == 'user'
            and " ".join(str(e.get('content') or '').split()) == rec_text
            for e in ctx[-8:]
        ):
            ctx.append({k: v for k, v in recovered.items() if k != 'timestamp'})
```
Then drop the block currently in `_persist_cancelled_turn`. (When `context_messages` is empty — first-turn error — `prefer_context` falls back to `session.messages`, which still carries the flag, so the early return stays correct.)

## Tests to add
- The interrupted-with-saved-partial shape from #1 (assert no adjacent assistants reach the API).
- The provider-error / tool-limit shape from #2 (populated `context_messages`, recovered turn arriving as a flagless state.db delta → asserts it's stripped). The current coverage only exercises `_persist_cancelled_turn`, which is why CI is green despite #2.

Once these two are in (and the mobile-UX/image commits are split out), ping me and I'll re-run the full gate and ship. Genuinely good catch on the root cause — it's close.

kaishi00:
Closing in favor of a clean PR that addresses both review issues (conditional _recovered skip + mirror moved to _materialize_pending_user_turn_before_error) and splits out the mobile-UX/image-paste commits. New PR incoming.

PR #4393, opened 2026-06-18
fix: stop interrupted-turn user prompt from replaying in every subsequent turn (#4283)

kaishi00:
## Problem

After an interrupted turn (cancel, gateway restart, tool iteration limit), the interrupted turn's user message gets prepended to **every subsequent turn** indefinitely — not just the next one. Reported in #4283.

## Root Cause

`_materialize_pending_user_turn_before_error()` appends the stale pending prompt to `session.messages` with `_recovered=True`. But `state.db` has no `_recovered` column — the flag is in-memory only and is lost on the DB round-trip. On the next turn, `reconciled_state_db_messages_for_session(prefer_context=True)` finds the stale message as a flagless state.db delta, and `_sanitize_messages_for_api()` cannot filter it.

### Fix 1: Conditional `_recovered` skip (review issue #1)

Added `_recovered_user_anchors_kept_assistant()` — a shared predicate that checks whether a `_recovered` user is followed by a kept assistant message (non-error, non-empty-partial, with content or tool_calls). The `_recovered` skip in both `_sanitize_messages_for_api()` and `_api_safe_message_positions()` is now **conditional**: only strip the recovered user when no kept assistant follows.

This prevents adjacent-assistant 400 errors on strict providers (DeepSeek, newer OpenAI) when a partial answer was saved before the interrupt:

```
input : user Q1 / assistant A1 / user "Q2"(_recovered) / assistant "partial…"(_partial) / user Q3
output: ['user', 'assistant', 'user', 'assistant', 'user']  ← Q2 kept, no adjacent assistants
```

The pure-cancel shape (recovered user + _error marker) still strips correctly since nothing anchors.

### Fix 2: Mirror moved to `_materialize_pending_user_turn_before_error` (review issue #2)

The `context_messages` mirror now lives in `_materialize_pending_user_turn_before_error()` (right after `session.messages.append(recovered)`) instead of `_persist_cancelled_turn()`. This covers **all three callers**:
- `_persist_cancelled_turn()` (cancel path)
- Provider-error path (~line 7557)
- Exception path (~line 8532)

The mirror includes a last-8 dedup lookback to prevent unbounded growth from repeated cancels. When `context_messages` is empty/None (first-turn error), the mirror is skipped — `prefer_context` falls back to `session.messages` which still carries the flag.

## Tests

12 new tests in `tests/test_issue4283_recovered_context_replay.py`:

**Shape 1 — interrupted-with-saved-partial (adjacent assistant risk):**
- `test_recovered_user_kept_when_anchoring_partial_assistant`
- `test_recovered_user_stripped_when_no_kept_assistant_follows`
- `test_recovered_user_stripped_when_next_is_user`
- `test_recovered_user_kept_when_anchoring_assistant_with_tool_calls`
- `test_anchor_predicate_returns_false_at_end`
- `test_anchor_predicate_returns_true_for_partial_assistant`
- `test_anchor_predicate_skips_error_assistant`

**Shape 2 — provider-error/tool-limit (flagless delta):**
- `test_materialize_mirrors_recovered_user_to_context_messages`
- `test_materialize_does_not_duplicate_context_messages`
- `test_materialize_skips_mirror_when_context_messages_empty`
- `test_sanitize_strips_recovered_user_from_context_messages`
- `test_api_safe_positions_strips_recovered_user`

All 12 pass. 533 existing tests pass (1 unrelated profile-cookie auth failure).

## CHANGELOG

Updated under `[Unreleased] → Fixed`.

Refs: #4283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **An interrupted turn's user prompt no longer leaks into every subsequent turn's model context (#4283).** When a turn was interrupted (cancel, gateway restart, tool iteration limit), the `pending_user_message` recovery path appended the stale prompt to `session.messages` with `_recovered=True`, but did not mirror it to `session.context_messages`. Because `state.db` has no `_recovered` column, the flag was lost on the DB round-trip — the next turn's `reconciled_state_db_messages_for_session(prefer_context=True)` found the stale prompt as a flagless state.db delta, and `_sanitize_messages_for_api` could not filter it. The agent core then merged it with each new user message, prepending the old prompt to every turn indefinitely. The fix mirrors the recovered user to `context_messages` inside `_materialize_pending_user_turn_before_error` (covering all three callers: cancel, provider-error, and exception paths), and the `_recovered` skip in `_sanitize_messages_for_api` / `_api_safe_message_positions` is now conditional — it only strips the recovered user when no kept assistant message follows, preventing adjacent-assistant 400 errors on strict providers when a partial answer was saved before the interrupt.
+
 ## [v0.51.483] — 2026-06-18 — Release QS (virtual-scroll height + measurement scroll-jump fixes)
 
 ### Fixed
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -3407,6 +3407,29 @@ def _is_reasoning_only_assistant_message(msg) -> bool:
     return _content_has_reasoning_only_parts(content)
 
 
+def _recovered_user_anchors_kept_assistant(messages, start_idx):
+    """Return True when a _recovered user at *start_idx* is followed by a kept
+    assistant message (non-error, non-empty-partial, with visible content or
+    tool_calls).  In that case the user must be retained to preserve role
+    alternation — dropping it would leave two adjacent assistant messages,
+    which strict providers (DeepSeek, newer OpenAI) reject with HTTP 400.
+    """
+    for msg in messages[start_idx + 1:]:
+        if not isinstance(msg, dict):
+            continue
+        if _is_reasoning_only_assistant_message(msg):
+            continue
+        if msg.get('_error'):
+            continue
+        if msg.get('_partial') and not str(msg.get('content') or '').strip():
+            continue
+        role = msg.get('role')
+        if role == 'assistant':
+            return True
+        return False
+    return False
+
+
 def _sanitize_messages_for_api(messages, *, cfg: dict = None):
     """Return a deep copy of messages with only API-safe fields.
 
@@ -3442,7 +3465,7 @@ def _sanitize_messages_for_api(messages, *, cfg: dict = None):
 
     # Second pass: build the sanitized list, dropping orphaned tool messages.
     clean = []
-    for msg in messages:
+    for idx, msg in enumerate(messages):
         if not isinstance(msg, dict):
             continue
         # Skip display-only Thinking entries. They are visible transcript
@@ -3460,6 +3483,19 @@ def _sanitize_messages_for_api(messages, *, cfg: dict = None):
         # API 400 errors on strict providers (empty assistant content).
         if msg.get('_partial') and not str(msg.get('content') or '').strip():
             continue
+        # Skip recovered user messages from the pending_user_message repair
+        # path — BUT only when they don't anchor a subsequent kept assistant
+        # message.  After an interrupted turn, the recovery logic appends the
+        # stale pending message to context_messages with _recovered=True (#4283).
+        # Without this filter, that message persists in model-facing context
+        # indefinitely, causing the agent core to merge it with the current
+        # turn's user message (concatenating with \n\n).  However, when a
+        # _partial assistant was saved before the interrupt, dropping the
+        # recovered user would leave two adjacent assistant messages, which
+        # strict providers reject with HTTP 400 — so the skip is conditional.
+        if msg.get('_recovered') and msg.get('role') == 'user':
+            if not _recovered_user_anchors_kept_assistant(messages, idx):
+                continue
         role = msg.get('role')
         if role == 'tool':
             tid = msg.get('tool_call_id') or ''
@@ -3525,6 +3561,12 @@ def _api_safe_message_positions(messages):
             continue
         if msg.get('_partial') and not str(msg.get('content') or '').strip():
             continue
+        # Conditionally skip _recovered user messages — mirrors
+        # _sanitize_messages_for_api.  Only skip when the recovered user
+        # does NOT anchor a subsequent kept assistant message (#4283).
+        if msg.get('_recovered') and msg.get('role') == 'user':
+            if not _recovered_user_anchors_kept_assistant(messages, idx):
+                continue
         role = msg.get('role')
         if role == 'tool':
             tid = msg.get('tool_call_id') or ''
@@ -5003,6 +5045,23 @@ def _materialize_pending_user_turn_before_error(session) -> bool:
     if pending_attachments:
         recovered['attachments'] = list(pending_attachments)
     session.messages.append(recovered)
+    # Mirror to context_messages so the _recovered flag survives the state.db
+    # round-trip (#4283).  state.db has no _recovered column, so without this
+    # mirror the next turn's reconciled_state_db_messages_for_session(
+    # prefer_context=True) finds the recovered user as a flagless state.db
+    # delta and _sanitize_messages_for_api cannot filter it — causing the
+    # interrupted turn's prompt to be prepended to every subsequent turn.
+    # Placing the mirror here (rather than in _persist_cancelled_turn) covers
+    # all three callers: cancel, provider-error, and exception paths.
+    ctx = getattr(session, 'context_messages', None)
+    if isinstance(ctx, list) and ctx:
+        rec_text = " ".join(str(recovered.get('content') or '').split())
+        if not any(
+            isinstance(e, dict) and e.get('role') == 'user'
+            and " ".join(str(e.get('content') or '').split()) == rec_text
+            for e in ctx[-8:]
+        ):
+            ctx.append({k: v for k, v in recovered.items() if k != 'timestamp'})
     # The new user turn is now committed to messages (#3831): retire a positive
     # truncation watermark left over from a prior retry/undo/edit so it cannot
     # freeze at the old edit boundary and later drop these post-edit turns on an
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -3407,29 +3407,6 @@ def _is_reasoning_only_assistant_message(msg) -> bool:
     return _content_has_reasoning_only_parts(content)
 
 
-def _recovered_user_anchors_kept_assistant(messages, start_idx):
-    """Return True when a _recovered user at *start_idx* is followed by a kept
-    assistant message (non-error, non-empty-partial, with visible content or
-    tool_calls).  In that case the user must be retained to preserve role
-    alternation — dropping it would leave two adjacent assistant messages,
-    which strict providers (DeepSeek, newer OpenAI) reject with HTTP 400.
-    """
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Re-ran the full gate against this v2 (worktree off `fix/context-messages-replay-v2` at HEAD `81779679`, diff vs `origin/master`). Both blocking issues from #4390 are now resolved cleanly, CI is green across the 3.11/3.12/3.13 matrix, and the test coverage maps directly onto the two failure shapes. This is ready to ship.

## Issue #1 (conditional `_recovered` skip) — resolved

`_recovered_user_anchors_kept_assistant()` (`api/streaming.py:3409`) is the shared predicate, and the skip is now gated in both `_sanitize_messages_for_api` (:3496) and `_api_safe_message_positions` (:3564):

```python
if msg.get('_recovered') and msg.get('role') == 'user':
    if not _recovered_user_anchors_kept_assistant(messages, idx):
        continue
```

The predicate's skip-set (non-dict, reasoning-only, `_error`, empty-`_partial`) is identical to the sanitize loop's own skip-set, so the "next kept message" lookahead and the actual emission stay in lockstep. I confirmed the tool-call case works: `_is_reasoning_only_assistant_message` returns `False` early when `tool_calls` is present (:3401), so an empty-content assistant carrying tool calls correctly anchors the recovered user. `test_recovered_user_kept_when_anchoring_partial_assistant` asserts the `['user','assistant','user','assistant','user']` shape — no adjacent assistants reach the API.

## Issue #2 (mirror covers all three callers) — resolved

The mirror now lives in `_materialize_pending_user_turn_before_error` (:5048), right after `session.messages.append(recovered)`, with the last-8 dedup lookback. I grepped the callers and all three converge here: cancel (`_persist_cancelled_turn` :1172), provider-error (:7590), exception (:8565). There is no leftover mirror block in `_persist_cancelled_turn` — the only `context_messages` append is the centralized one at :5056. `test_materialize_skips_mirror_when_context_messages_empty` covers the first-turn fallback (empty `context_messages` -> `prefer_context` falls back to `session.messages`, which still carries the flag).

## One minor, non-blocking edge

The predicate returns `True` for a recovered user followed by an assistant whose only payload is **unanswered** `tool_calls` + empty content. That assistant would later be dropped by the third pass (orphaned-tool-call removal, :3548) when no matching tool result exists, leaving the recovered user dangling. The agent core merges adjacent users, so this cannot produce a 400 — but if you want belt-and-suspenders later, a follow-up test asserting the recovered-user + orphaned-tool-call shape would close the gap. Not worth holding this PR for.

CHANGELOG entry under `[Unreleased] -> Fixed` accurately describes the state.db round-trip root cause. Approving — good follow-through on the split-out and the two fixes.

nesquena-hermes:
Thanks @kaishi00 — this is a real improvement over the earlier attempt (#4390): the conditional anchor guard and the all-three-caller `context_messages` mirror both directly address that bounce, and the common cases are correct and well-tested (pure-cancel strip, partial-anchor retention, the mirror + dedup). I verified the residual-gap claim too — a `_recovered` user genuinely does survive master's sanitizer on a pure-cancel shape, so this is a real complementary fix, not a dup of the shipped #4286 floor-layer work.

I took it through the full ship gate (Codex + Opus + the full suite, 9405 passing). The suite is green, but **both advisors independently reproduced the same CORE regression** — and I reproduced it myself against the branch. It's the exact failure class this PR exists to prevent, on a shape the happy-path tests don't cover.

## The bug: the forward-scan anchor predicate diverges from the sanitizer's later passes

`_recovered_user_anchors_kept_assistant` decides "is the next kept message an assistant?" with a *local forward scan*, but `_sanitize_messages_for_api` later drops more messages in pass 2/3 (orphaned `tool` rows, and all-orphan empty-`tool_calls` assistants) based on the full `answered_ids` set — which a forward scan structurally can't replicate. Two reproduced shapes (verified against this branch):

**(a) Stale user survives — fix doesn't fully work:**
```
[assistant A1, _recovered user, empty-assistant{orphaned tool_calls}]
→ sanitizes to [assistant A1, STALE recovered user]
```
The helper treats the empty orphan-tool-call assistant as a valid anchor and keeps the recovered user, but the sanitizer then drops that assistant → the stale user is left behind (the very thing #4283 is about).

**(b) Adjacent assistants → strict-provider HTTP 400 (the #4390 regression class):**
```
[assistant A1, _recovered user, orphan tool row, assistant A2]
→ sanitizes to [assistant A1, assistant A2]   ← adjacent → DeepSeek/newer-OpenAI 400
```
The helper returns False on the `tool` row (non-assistant) and strips the recovered user, but the sanitizer drops the orphan tool too, leaving two adjacent assistants. master handled this interrupt-mid-tool history *without* a 400; this re-introduces one.

This matters because "interrupt while a tool call is in flight" is exactly when tool calls get orphaned — squarely in this fix's domain, not an exotic shape.

## Recommended fix (Opus's design — small and robust)

Don't try to widen the forward-scan predicate to chase passes 2+3 — it'll stay fragile. Instead **decide the `_recovered`-user drop on the actual sanitized sequence**:
1. Run the existing sanitizer passes *without* the `_recovered` skip.
2. In a final pass, drop each `_recovered` user **only if removing it would NOT make its two surviving neighbors the same role** (keep it iff `prev_kept.role == next_kept.role` after removal would collide — i.e. drop only when `prev_kept.role != next_kept.role`).

Computed on the real post-sanitize kept list, this is automatically consistent with passes 2 and 3, fixes (b), and fixes the user/user case in (a) for free. Apply the identical logic in both `_sanitize_messages_for_api` (~:3497) and `_api_safe_message_positions` (~:3568) so they can't diverge.

Please also add two regression tests for the missing shapes: **orphan-tool-after-recovered** and **content-less-orphan-toolcall anchor** (both currently pass green while the bug is live — that's why the suite didn't catch it).

You're very close — the convergence on the #4390 findings was spot-on; this is one more layer of the same "match the sanitizer exactly" requirement. Ping me when it's reworked and I'll re-run the full gate.

kaishi00:
Pushed v3 that addresses both reproduced shapes from the review.

## Root cause of the divergence

The forward-scan predicate (`_recovered_user_anchors_kept_assistant`) scanned the **raw** message list to decide "is the next kept message an assistant?" — but it couldn't replicate what pass 2/3 would actually drop (orphaned `tool` rows, empty orphaned `tool_calls` assistants). So its "next kept message" prediction could diverge from reality.

## Fix: post-sanitize pass 4

Removed the forward-scan predicate entirely. The `_recovered` skip now runs as a **4th pass** on `filtered_clean` (the post-sanitized list), so the anchor check is exact — it sees exactly what will reach the API:

- **Pass 2**: let `_recovered` users through to `clean` (temporarily marked with `_recovered` on the sanitized dict)
- **Pass 3**: strip orphaned tool_calls (existing, unchanged)
- **Pass 4** (new): walk `filtered_clean` and drop `_recovered` users whose next surviving message is NOT an assistant. Strip the temporary `_recovered` marker from retained entries before return.

Both `_sanitize_messages_for_api` and `_api_safe_message_positions` get the same 4-pass structure.

## Shape (a) — stale user survives → now fixed

```
[assistant A1, _recovered user, empty-assistant{orphaned tool_calls}]
→ pass 3 drops the orphaned assistant
→ pass 4: next surviving message after recovered user is None → drop it
→ output: [assistant A1]  ✓
```

## Shape (b) — adjacent assistants → now fixed

```
[assistant A1, _recovered user, orphan tool, assistant A2]
→ pass 2 drops orphan tool
→ pass 4: next surviving message after recovered user is assistant A2 → keep it
→ output: [assistant A1, user Q2, assistant A2]  ✓ no adjacent assistants
```

## Tests

14 tests (2 new for the Nesquena shapes):
- `test_stale_user_dropped_when_anchor_is_orphaned_empty_assistant` — shape (a)
- `test_no_adjacent_assistants_when_orphan_tool_between` — shape (b)
- `test_no_recovered_marker_in_output` — verifies the temporary marker is stripped

All 14 pass. 547 existing tests pass (1 unrelated profile-cookie failure).

nesquena-hermes:
## Summary

Re-reviewed v3 at HEAD `5fd172a6` (read-only worktree, diff vs `origin/master`). This implements the exact "decide the drop on the actual sanitized sequence" design from my `03:08` review, and both reproduced failure shapes are now closed. The forward-scan predicate is gone, the skip moved to a 4th pass over `filtered_clean`, and the two missing regression shapes are covered with assertions that match the new behavior. From a static review this is correct; I'd ship it once your Codex+Opus+suite gate re-runs green on this SHA.

## The fix is structurally right

`_recovered_user_anchors_kept_assistant` is deleted (no leftover refs — `grep -rn` across `api/` and `tests/` is clean). Pass 2 now lets `_recovered` users through with a temporary marker, and the drop decision happens after orphaned tool/`tool_calls` stripping (`api/streaming.py:3513-3534`):

```python
# Fourth pass: drop _recovered user messages that don't anchor a kept
# assistant.  Operating on filtered_clean ... means the anchor check is exact
for i, msg in enumerate(filtered_clean):
    if msg.get('_recovered') and msg.get('role') == 'user':
        keeps_assistant = False
        for j in range(i + 1, len(filtered_clean)):
            next_role = filtered_clean[j].get('role')
            if next_role == 'assistant':
                keeps_assistant = True
                break
            if next_role == 'user':
                break
        if not keeps_assistant:
            continue  # drop — no assistant to anchor
        msg = {k: v for k, v in msg.items() if k != '_recovered'}
```

Because this runs on `filtered_clean` (post pass 2 + pass 3), it sees exactly what reaches the API — the divergence class from my review is gone.

## Both reproduced shapes verified by tracing

- **Shape (a)** — `[A1, _recovered user, empty-assistant{orphaned tool_calls}, Q3]`. Pass 3 drops the orphan-only empty assistant (`:3508-3511`, `if not kept: ... if not str(content).strip(): continue`). Pass 4 then sees the next surviving role after the recovered user is `user` (Q3) → break → drop. Output `[user, assistant, user]`. `test_stale_user_dropped_when_anchor_is_orphaned_empty_assistant` asserts exactly this. ✓
- **Shape (b)** — `[A1, _recovered Q2, orphan tool, A2, Q3]`. Pass 2 drops the orphan `tool` row (`:3470-3474`). Pass 4 sees A2 as next surviving → keep Q2. Output `[user, assistant, user, assistant, user]`, no adjacent assistants. `test_no_adjacent_assistants_when_orphan_tool_between` zips roles to assert no `assistant,assistant` pair. ✓

The `_api_safe_message_positions` mirror (`:3596-3616`) carries the identical pass-4 block, so the two paths can't diverge — and `test_api_safe_positions_strips_recovered_user` / `_keeps_recovered_user_with_partial` lock the parity.

## One tiny test-coverage gap (non-blocking)

`test_no_recovered_marker_in_output` only checks `_sanitize_messages_for_api`. The positions path also re-adds then strips the marker (`:3590` + the pass-4 strip), but no test asserts `_recovered` is absent from the `_api_safe_message_positions` output. The logic is symmetric so I'm confident, but a one-line assertion in `test_api_safe_positions_strips_recovered_user` (`assert all("_recovered" not in m for _, m in positions)`) would close it. Not worth holding for.

## Verdict

Design matches the recommendation, dead code removed cleanly, both regression shapes covered with behavior-accurate assertions, and the two callers stay in lockstep. This is a static read-only approve — please re-run your ship gate on `5fd172a6` to confirm the full suite stays green before merge.

kaishi00:
Ship gate re-run on `3203c106` (v3 + test coverage patch):

**14/14 PR-specific tests passed** ✓ (including the new `test_no_recovered_marker_in_output` assertion for the `_api_safe_message_positions` path — addresses the non-blocking coverage gap).

**Full suite: 4350 passed, 5 skipped, 1 failed.** The single failure is `test_issue570_permission.py::test_load_settings_returns_defaults_when_settings_file_unreadable` — pre-existing on `master` (test expects `send_key` default `'enter'` but the local settings file has `'ctrl+enter'`). Confirmed identical failure on `master` before the PR. Unrelated.

Green for merge. 🟢

PR #4394, opened 2026-06-18
mobile UX: larger hamburger (44px), wider swipe zone (80px), works in browser+PWA

kaishi00:
## Changes

Three small mobile UX improvements, split out from PR #4393 per reviewer request:

### Hamburger menu hit area (static/index.html, static/style.css)
- SVG icon: 20px → 24px
- CSS button box: 32×32px → 44×44px (matches Apple HIG minimum touch target)

### Swipe gesture scope (static/boot.js)
- Enable sidebar swipe outside PWA standalone mode (works in mobile browser too)
- Widen swipe zone from 24px to 80px so users don't have to be right at the edge

The original 24px touch target is well below the 44×44px minimum recommended by Apple's Human Interface Guidelines and is nearly impossible to hit reliably on modern phone screens.

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -266,9 +266,9 @@ function closeMobileSidebar(){
   if(overlay)overlay.classList.remove('visible');
 }
 
-const _PWA_SIDEBAR_SWIPE_EDGE=28;
-const _PWA_SIDEBAR_SWIPE_TRIGGER=72;
-const _PWA_SIDEBAR_SWIPE_MAX_VERTICAL=48;
+const _PWA_SIDEBAR_SWIPE_EDGE=80;
+const _PWA_SIDEBAR_SWIPE_TRIGGER=64;
+const _PWA_SIDEBAR_SWIPE_MAX_VERTICAL=56;
 let _pwaSidebarSwipe=null;
 
 function _isPwaStandalone(){
@@ -298,7 +298,7 @@ function _openMobileSidebarFromGesture(){
 }
 
 function _onPwaSidebarSwipeStart(e){
-  if(!_isPwaStandalone()||_isDesktopWidth())return;
+  if(_isDesktopWidth())return;
   if(e.pointerType==='mouse'||(e.pointerType&&e.pointerType!=='touch'&&e.pointerType!=='pen'))return;
   if(document.querySelector('.sidebar')?.classList.contains('mobile-open'))return;
   const clientX=Number(e.clientX)||0;
--- a/static/index.html
+++ b/static/index.html
@@ -119,7 +119,7 @@
 <body>
 <header class="app-titlebar" role="banner">
   <button class="app-titlebar-hamburger has-tooltip has-tooltip--bottom" id="btnHamburger" onclick="toggleMobileSidebar()" type="button" data-tooltip="Menu" aria-label="Menu">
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Reviewed the split-out (worktree off `fix/mobile-ux-hamburger-and-image-paste` at HEAD `cb58f9c7`, diff vs `origin/master`). This is the clean separation I asked for on #4390 — three files, one logical change, no streaming.py or image-paste carryover. CI green across the matrix. The hamburger sizing is uncontroversial; one note on the swipe change.

## Hamburger hit area — fine

`static/index.html:122` bumps the SVG to 24px and `static/style.css:992` takes the button box to 44×44:

```css
.app-titlebar-hamburger,.app-titlebar-spacer{display:none;width:44px;height:44px;flex-shrink:0;}
```

The `display:none` here is the mobile-hidden default that the media query flips to flex; the 44px matches the Apple HIG minimum. No layout risk — `flex-shrink:0` is preserved and the spacer keeps the title centered.

## Swipe scope — sound, with one mobile-browser caveat

The real behavior change is dropping the standalone gate in `_onPwaSidebarSwipeStart` (`static/boot.js:299`):

```js
function _onPwaSidebarSwipeStart(e){
  if(_isDesktopWidth())return;            // was: if(!_isPwaStandalone()||_isDesktopWidth())
```

The listener was already installed unconditionally via `_installPwaSidebarSwipeGesture()` (:336), so this just lets the start handler proceed in a plain mobile browser. The `_isDesktopWidth()` guard still blocks desktop, the pointer-type filter still rejects mouse (:300), and `_isInteractiveSwipeTarget` (:282) still excludes `input,textarea,select,button,a,[contenteditable],.sidebar,.rightpanel`. The move handler requires horizontal intent (`dx>=64 && dx>Math.abs(dy)*1.5`, :316), so vertical scroll won't accidentally trigger it. That all holds up.

The one thing worth flagging: in a **non-standalone** mobile browser, the left edge is also where the OS/browser owns the back-navigation gesture (iOS Safari edge-swipe-back, Android Chrome). Widening `_PWA_SIDEBAR_SWIPE_EDGE` from 28→80 (:269) means the capture zone now overlaps that region much more aggressively. `e.preventDefault()` only fires *after* the 64px trigger (:319, behind `if(e.cancelable)`), and iOS Safari's edge-back recognizer often ignores `preventDefault` on `pointermove` anyway — so on iOS you may get a race where a left-edge drag both navigates back *and* arms the sidebar. Not a blocker (it degrades to the existing browser behavior, and the standalone PWA path — the primary target — has no back-gesture), but worth a quick manual check on iOS Safari + Android Chrome before merge.

Minor/cosmetic: the `_PWA_*` constant and `_onPwaSidebar*` function names are now misleading since they apply outside PWA standalone too. A rename to `_MOBILE_SIDEBAR_SWIPE_*` would read better, but it's noise for this PR — fine to leave.

Approving the direction; just confirm the iOS/Android edge-gesture feel on a real device.

nesquena-hermes:
Shipped in **v0.51.485** (Release QU) 🚀 — thanks @kaishi00, and thanks for splitting this out cleanly from #4390! The 44px hamburger tap target + 24px icon is a clear mobile a11y win, and the edge-swipe now works in mobile browser, not just the installed PWA. I deep-reviewed + live-drove it at 390×844 (hamburger taps open the sidebar; a left-edge swipe opens it in a non-PWA context), and it passed the full gate (Codex + Opus + the suite, 9393 passing). We're shipping the wider browser edge-zone as-is — the one known tradeoff is that an 80px left-edge zone in mobile *browser* overlaps iOS Safari's back-swipe, but it degrades gracefully and the installed PWA (the primary mobile target) has no back gesture. Closing as merged.

PR #2490, opened 2026-05-18
ci(docker): runtime smoke gate for Docker init logic

nesquena-hermes:
## What

Adds a Docker runtime smoke gate: `.github/workflows/docker-smoke.yml`. Path-filtered on any Docker-touching change (`Dockerfile`, `docker_init.bash`, `docker-compose*.yml`, `.dockerignore`, `.env.docker.example`). Actually runs `docker compose up` for each of the three compose variants against a real Docker daemon on the GHA runner and asserts the WebUI reaches `/health` without tripping any known-bad startup log signature.

## Why

v0.51.84 (PR #2470) shipped a startup-killing `:ro`-mount × `chown -h ... {} +` regression — every multi-container deploy would have died at startup under `set -e` with EROFS. **9 new source-level pytest invariants plus 5814 existing tests all passed** because none actually ran `docker_init.bash` against a real `:ro` mount. The independent reviewer caught it by eye and pushed the fix before merge.

This workflow closes that class of gap by exercising the actual runtime semantics: build, mount, chown, init, healthcheck.

### Two jobs

1. **`compose-config`** — preflight that `docker compose config` parses each of the three compose files. Cheap, fast, catches schema/interpolation drift in parallel before any container starts.

2. **`smoke`** — matrix of `single` / `two-container` / `three-container`. For each variant:
   - Reap any stray `hermes-smoke-*` resources (defence-in-depth on self-hosted runners).
   - `docker build -t ghcr.io/nesquena/hermes-webui:latest .`
     - **Critical step.** The multi-container compose files reference `ghcr.io/nesquena/hermes-webui:latest`. Without this retag, multi-container smoke would test the *previously-released* registry image, not the PR's changes. With it, Compose's default `pull_policy=missing` keeps the local build in place and the PR is genuinely exercised.
   - `mktemp -d` for ephemeral `HERMES_HOME` + `HERMES_WORKSPACE` — runner's host filesystem is never touched.
   - `docker compose up -d --wait --wait-timeout 120` — the `Dockerfile` carries `HEALTHCHECK CMD curl -f http://localhost:8787/health || exit 1`, so `--wait` blocks on `healthy`, not just `running`.
   - `curl /health` poll loop (30 × 2s) as headroom for the multi-container variants' first-start dependency install.
[...]

PR #4398, opened 2026-06-18
fix(#4395): make staged hermes-agent source writable after rsync/cp

### LINKED ISSUE #4395: [bug] Multi-container WebUI dies at startup with `could not create 'hermes_agent.egg-info': Permission denied` — staging block in docker_init.bash preserves source mode 0555
## Summary

Multi-container Docker deploys (`docker-compose.two-container.yml`, `docker-compose.three-container.yml`) fail at startup with `could not create 'hermes_agent.egg-info': Permission denied`. The staging block added in `70f371c8` to work around the `:ro` mount EROFS did its job for the *source-root* permission error, but `rsync -a` / `cp -a` preserve the source's mode bits, so the freshly-mkdir'd `/tmp/hermes-agent-build` comes out **mode 555** — and setuptools' PEP 517 `egg_info` cannot create `<pkg>.egg-info/` next to the package inside a 555 directory even though hermeswebui owns it.

Affects every multi-container deploy with `hermes-agent-src:ro` (the production default since v0.51.84, PR #2470). Single-container deploys (`docker-compose.yml`, hermes-agent baked into `/opt/hermes` as a normal writable directory) are not affected — they mount the agent source read-write.

## Environment

- WebUI: any release since `70f371c8` (v0.51.84+)
- Hermes Agent: any (this is a WebUI init-script bug, not an agent bug)
- Deploy: `docker-compose.two-container.yml` or `docker-compose.three-container.yml` (i.e. shared `hermes-agent-src` named volume, WebUI side mounted `:ro`)

## Reproduction

```bash
# Reproduce the staging perm-inheritance behavior on the host:
mkdir -p /tmp/hermes-agent-build
cp -a /some/read-only-checkout/. /tmp/hermes-agent-build/
stat -c '%a' /tmp/hermes-agent-build          # → 555
mkdir /tmp/hermes-agent-build/hermes_agent.egg-info
# mkdir: cannot create directory '/tmp/hermes-agent-build/hermes_agent.egg-info': Permission denied
```

In a container, `uv pip install "$_stage_src[all]"` then fails during setuptools' PEP 517 `egg_info`:

```
[stderr]
running egg_info
creating hermes_agent.egg-info
error: could not create 'hermes_agent.egg-info': Permission denied
```

(The `Pattern 'LICENSE' did not match any files` line that also appears is unrelated setuptools 77+ PEP 639 deprecation noise.)

## Expected

`uv pip install` succeeds and the container reaches `/health`.

## Actual

Under `set -e`, the install fails and the container exits with `!! ERROR: Failed to install hermes-agent's requirements` (script ID logged on its way out).

## Root cause

`docker_init.bash` correctly stages the read-only source into `/tmp/hermes-agent-build` so setuptools can write `<pkg>.egg-info/` into the source tree. The intent is "fresh tmpfs dir, owned by hermeswebui, writable." But:

- `rsync -a` includes `-p` (preserve permissions). With source mode 555, the staged copy is mode 555.
- `cp -a` includes `-p`. Same result.
- `mkdir -p "$_stage_src"` only sets the mode on the directory *before* the copy. The copy then replaces that with source perms.

Result: the staging block's invariant ("writable tmpfs copy") is silently violated on every multi-container startup. The `set -e` propagates the EROFS-style failure and kills the container before WebUI can serve traffic.

Note: the existing `setup_dependencies_only_first_run` cache (`/app/venv/.deps_installed`) does NOT save you on first run, and `uv` rebuilds the build env on every container start, so this fires every fresh start of a multi-container deploy.

## Suggested fix

Add one line in `docker_init.bash` after both the rsync and cp-fallback branches and before `uv pip install`:

```bash
chmod -R u+w "$_stage_src" \
  || error_exit "Failed to make staged hermes-agent source writable (rsync/cp preserved :ro mount perms)"
```

This re-adds owner-write to the staged tree without widening perms beyond what the runtime user needs.

A source-level regression test can be added to `tests/test_docker_docs_and_readonly.py` asserting:
1. The chmod substring is present in the staging block.
2. It uses the exact `chmod -R u+w "$_stage_src"` form.
3. It lives AFTER the rsync/cp `if/else/fi` so it covers both copy paths.

## Why the existing smoke gate didn't catch this

The Docker runtime smoke workflow added in PR #2490 scans for `EROFS | Read-only file system | Traceback | PermissionError | error_exit` in startup logs and would flag a real `:ro`-mount failure. It would also flag this one — the `error_exit` line is identical. But because `HERMES_WEBUI_STATE_DIR=/home/hermeswebui` (the default inside the smoke image) is on overlay2 and the hermes-agent fixture in the smoke job mounts the source read-write, the staging path never exercises mode-0555 inheritance. The perm-inheritance class is a mode-bit nuance that's invisible until the source actually mounts with restrictive perms.

## Related

- `70f371c8` — the staging commit that introduced this code path; correctly addressed EROFS but missed perm-inheritance.
- `2d66263a` / PR #2470 — switched the WebUI's `hermes-agent-src` mount to `:ro`, exposing the underlying perm-inheritance bug.
- PR #2490 — added the Docker runtime smoke gate; would catch the `error_exit` log line but the smoke job's source isn't mounted `:ro`, so the staging never enters the bad code path.

/cc @nesquena-hermes since this overlaps the same Docker hygiene slice that 70f371c8 / #2490 / #2495 covered.

enihcam:
## Summary

Multi-container WebUI deploys (`docker-compose.two-container.yml`, `docker-compose.three-container.yml`) die at startup with:

```
error: could not create 'hermes_agent.egg-info': Permission denied
```

`docker_init.bash` correctly stages the `:ro` `hermes-agent-src` source into `/tmp/hermes-agent-build` to avoid EROFS (added in `70f371c8`), but `rsync -a` / `cp -a` preserve the source tree's mode bits. A source mounted mode 555 leaves the staged copy also mode 555, and setuptools' PEP 517 `egg_info` then cannot create `<pkg>.egg-info/` next to the package — even though hermeswebui owns the dir.

Fix: `chmod -R u+w "$_stage_src"` after both copy paths and before `uv pip install`. Minimal-permission form (`u+w`, not `a+w`) so the staged tree isn't accidentally world-writable.

Closes #4395.

## Changes

| File | Change |
|---|---|
| `docker_init.bash` | +8 (chmod + 7-line `NB:` comment explaining the perm-inheritance trap) |
| `tests/test_docker_docs_and_readonly.py` | +50 (`import re` + `test_docker_init_makes_staged_dir_writable_after_ro_mount_copy`) |
| `CHANGELOG.md` | +2 (one bullet under `[Unreleased]`) |

Total: 3 files, 61 insertions, 0 deletions.

## Why staging-as-tmpfs alone isn't enough

The original staging commit (`70f371c8`) assumed a freshly-`mkdir`'d tmp dir would be writable. It would be — if nothing came after to overwrite the mode. Both `rsync -a` and `cp -a` include `-p` (preserve permissions), so the staging dir's mode 700 from `mkdir -p` gets clobbered to whatever the source's mode is. The fix preserves the original staging invariant — *writable* tmpfs copy — by re-adding owner-write after the copy.

## Why the existing PR #2490 smoke gate didn't catch this

The smoke job mounts the `hermes-agent-src` fixture read-write, so the staging path never exercises mode-555 inheritance. The `error_exit` log signature is identical to the EROFS case the gate does catch, but the fixture won't trigger it. The new regression test (source-level, not runtime) is the durable defence for this class.

## Verification

| Gate | Result |
|---|---|
| `bash -n docker_init.bash` | clean |
| `pytest tests/test_docker_docs_and_readonly.py -v` | **12 passed** (11 existing + 1 new) |
| Host-level reproduction (read-only source → `cp -a` → mode 555 → setuptools `Permission denied`) | confirmed before fix |
| Same reproduction with `chmod -R u+w` after `cp -a` | install succeeds (`Built hermes-agent`) |
| Negative test (revert the chmod line) | the new test fails with a clear error message naming the missing `chmod -R u+w "$_stage_src"` invariant |

## Risk

- One-line change in a bash script. No new code path, no dependency changes, no API surface change.
- The `chmod -R u+w` widens perms to owner-write, which is what the staging block's invariant already required. No privilege expansion.
- Affects only the WebUI's install path. The agent service is unchanged.

## Out of scope (noted, not addressed)

A long-term fix would be to refactor the staging block into a named function in a sourced helper file so it can be exercised by an integration test that creates an actual `:ro` tmpfs mount and asserts the resulting staged dir is writable — source-level invariants only catch what they're written to catch. Worth a follow-up; not a release gate for this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ## [Unreleased]
 
+- **Multi-container WebUI no longer dies on `uv pip install hermes-agent` with `could not create 'hermes_agent.egg-info': Permission denied` (#4395).** The staging block in `docker_init.bash` now `chmod -R u+w "$_stage_src"` after the rsync/cp copy. Without it, `rsync -a` / `cp -a` preserved the source tree's mode bits from the `:ro` `hermes-agent-src` mount (mode 555), leaving the staged `/tmp/hermes-agent-build` itself read-only even though hermeswebui owned it; setuptools then could not create `<pkg>.egg-info/` during PEP 517 `egg_info` and the container crashed at startup under `set -e`. Affects every multi-container deploy (`docker-compose.two-container.yml`, `docker-compose.three-container.yml`).
+
 ## [v0.51.484] — 2026-06-18 — Release QT (collapsible paused cron section)
 
 ### Added
--- a/docker_init.bash
+++ b/docker_init.bash
@@ -438,6 +438,13 @@ else
     # and `--reflink=auto` makes the copy near-free on overlay2/btrfs where
     # supported. We rebuild on every container start (the agent source can
     # change across volume re-init); cost is one rsync of ~10MB of Python source.
+    #
+    # NB: `rsync -a` / `cp -a` preserve the source tree's mode bits, so a `:ro`
+    # source mounted mode 555 leaves the staged copy also mode 555. setuptools
+    # then can't create `hermes_agent.egg-info/` next to the package — it dies
+    # with "Permission denied" even though `_stage_src` itself was created
+    # writable by hermeswebui. Re-add owner-write on the staged tree after the
+    # copy so the build dir is genuinely writable, not just owned by us.
     _stage_src="/tmp/hermes-agent-build"
     rm -rf "$_stage_src"
     mkdir -p "$_stage_src"
@@ -455,6 +462,8 @@ else
       rm -rf "$_stage_src"/*.egg-info "$_stage_src"/build "$_stage_src"/dist 2>/dev/null || true
       find "$_stage_src" -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true
     fi
+    chmod -R u+w "$_stage_src" \
+      || error_exit "Failed to make staged hermes-agent source writable (rsync/cp preserved :ro mount perms)"
     uv pip install "$_stage_src[all]" --trusted-host pypi.org --trusted-host files.pythonhosted.org \
       || error_exit "Failed to install hermes-agent's requirements"
     rm -rf "$_stage_src"
--- a/tests/test_docker_docs_and_readonly.py
+++ b/tests/test_docker_docs_and_readonly.py
@@ -19,6 +19,7 @@ Pins three invariants:
 
 from __future__ import annotations
 
+import re
 from pathlib import Path
 
 REPO = Path(__file__).resolve().parents[1]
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.487** (Release QX) 🚀 — thanks @enihcam! Clean diagnosis: `rsync -a`/`cp -a` preserving the `:ro` mount's 555 mode on the staged tree is exactly why setuptools couldn't write egg-info. The `chmod -R u+w` after staging (fail-closed via error_exit) fixes it without touching the read-only mount. Full gate green (Codex + Opus both SAFE, suite 9395). Closing as merged.

Issue #4385, opened 2026-06-17
bug(sessions): archived cron sessions reappear in Sessions tab


TomBanksAU:
## Bug Description
Archiving a cron-job conversation from the Sessions tab does not stay hidden. The row disappears briefly, then returns in the session list as if it was not archived.

## Steps to Reproduce
1. Enable/surface cron sessions in the Sessions tab.
2. Archive a cron-job conversation from the session row/action menu.
3. Refresh or wait for the session list to update.

## Expected Behavior
The archived cron conversation should stay hidden unless archived sessions are explicitly shown.

## Actual Behavior
The cron conversation returns to the Sessions tab after the list reloads.

## Proposed Solution
When WebUI archives a surfaced cron session, it should preserve the cron session identity and persist/suppress that archived state consistently so the raw Agent cron row is not reintroduced as an active session.

## Environment
- App: Hermes WebUI
- Source type: Hermes Agent cron sessions

Issue #1486, opened 2026-05-02
Sidebar rename for CLI / agent sessions reverts on hard refresh (compression-chain head/tip mismatch)


nesquena-hermes:
## Bug

Renaming a CLI / agent / gateway-imported session in the sidebar appears to succeed, but the old name returns after a hard refresh.

Filed from the diagnosis in #1134 (closed) by @zhonghuaY — credit for the root-cause analysis goes to them.

## Root cause (verified against master, v0.50.269)

`api.agent_sessions._project_agent_session_rows` collapses a compression chain into a single sidebar row. That row exposes the **tip**'s `session_id` (so navigation/import stay correct) but uses the chain **head**'s title for visible identity.

The rename API writes to the tip — the row that the sidebar item's `data-sid` attribute points to. On the next refresh, the projection re-renders using the unchanged head's title, so the rename appears to revert.

There's no existing `rename_cli_session` function in `api/state_sync.py` on master.

## Repro

1. Run a CLI session that goes through compression (long enough that the agent triggers a chain).
2. In the WebUI sidebar, double-click to rename it. Type a new title. Press Enter.
3. Verify the new title appears in the sidebar.
4. Hard refresh the page (Cmd+Shift+R / Ctrl+Shift+F5).
5. **Bug**: the old (head's) title is back.

## Suggested fix shape (from #1134)

The rename should walk `parent_session_id` upward across `end_reason='compression'` boundaries to find the lineage root, then update that row's title. Touching exactly one row keeps compatibility with `hermes_state`'s `UNIQUE INDEX … ON sessions(title) WHERE title IS NOT NULL` (the auto-generated `#N` titles on intermediate/tip rows stay distinct).

Implementation could land in `api/state_sync.py` as:

- `_resolve_lineage_root(db_path, session_id) -> str` — walk `parent_session_id` upward (capped at e.g. 64 hops) while parent's `end_reason == 'compression'`
- `rename_cli_session(session_id, title) -> bool` — resolve the root, sanitize the title, hand the canonical update to `SessionDB.set_session_title()` if available, fall back to raw SQL otherwise
- Wire into the existing `/api/session/rename` route in `api/routes.py` for non-WebUI-owned sessions

Reference implementation in #1134 was ~138 LOC plus integration. We'd want to write our own version with:

- A behavioural test that confirms the round-trip (rename → save → reload from disk → assert new title is on the head row)
[...]

PR #4399, opened 2026-06-18
fix(sessions): preserve archived cron state.db projections

### LINKED ISSUE #4397: bug(sessions): archived cron runs can reappear from state.db projection
## Bug Description
Archived cron sessions can still reappear in the Sessions panel after the list refreshes.

This is a follow-up to #4385 and the merged fix in #4388. That fix handled stale cron sidecars that were being treated like CLI sessions, but there is still another path where a cron run that was archived in WebUI gets re-added from Hermes `state.db` as an unarchived row.

## Steps to Reproduce
1. Enable non-WebUI sessions and cron sessions in the WebUI settings.
2. Archive a visible cron run from the Sessions panel.
3. Refresh or wait for the session list to reload.
4. Observe that the same cron run can appear again as active/unarchived.

## Expected Behavior
Once a cron run is archived from the Sessions panel, that specific run should stay hidden unless archived sessions are explicitly shown or it is restored.

## Actual Behavior
The archive request succeeds and the WebUI sidecar stores `archived: true`, but the sidebar can still show the same cron session again after it is projected from `state.db`.

## Proposed Solution
When WebUI projects cron sessions from Hermes `state.db`, it should respect any existing WebUI sidecar for that same session ID, especially the `archived` state. If an archived sidecar exists, the projected cron row should either carry `archived: true` or be suppressed from the active sidebar list.

## Related
- Follow-up to #4385
- Follow-up to merged PR #4388

TomBanksAU:
## Summary
Fixes archived cron runs reappearing in the Sessions panel after refresh when the run is re-projected from Hermes `state.db`.

This is a follow-up to closed issue #4385 and merged PR #4388. That fix covered stale cron sidecars being treated as CLI rows. This follow-up covers the remaining path where `all_sessions()` omits the hidden archived sidecar, then `get_cli_sessions()` re-adds the same cron run from `state.db` with `archived: false`.

## What changed
- Preserves UI-owned sidecar metadata when projecting state.db sessions into sidebar rows.
- Carries the archived flag from the WebUI sidecar for matching cron/tool/API projections.
- Adds a regression test for a hidden archived cron sidecar plus a state.db cron row.

## Testing
- `uv run --with pytest --with pyyaml --with cryptography pytest tests/test_issue4385_cron_archive_reappears.py -q`
- `uv run --with pytest --with pyyaml --with cryptography pytest tests/test_issue4385_cron_archive_reappears.py tests/test_issue3585_cron_session_overflow.py tests/test_gateway_sync.py -q`
- Verified against local reproduced data: affected cron session IDs now resolve as `archived=True` in `_build_session_list_cache_payload(...)`.

Closes #4397
Follow-up to #4385
Follow-up to #4388

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -3877,6 +3877,22 @@ def _all_profiles_cli_contexts() -> tuple[list[tuple[Path, Path, str | None]], t
     return contexts, tuple(cache_entries)
 
 
+def _state_projection_sidecar_metadata(sid: str) -> dict:
+    """Return UI-owned metadata for a state.db-projected sidebar row."""
+    metadata = {"title": None, "archived": False}
+    try:
+        webui_meta = Session.load_metadata_only(sid)
+    except Exception:
+        return metadata
+    if not webui_meta:
+        return metadata
+    title = getattr(webui_meta, 'title', None)
+    if title:
+        metadata["title"] = title
+    metadata["archived"] = bool(getattr(webui_meta, 'archived', False))
+    return metadata
+
+
 def _load_cli_sessions_uncached(
     hermes_home: Path,
     db_path: Path,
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.486** (Release QV) 🚀 — thanks @TomBanksAU! Nice catch on the remaining re-projection path — archived cron runs now stay hidden whether they come from the sidecar or get re-projected from state.db. Full ship gate green (Codex + Opus + the full suite, 9394 passing); Opus confirmed the archived-flag read isn't over-broad (live runs still show), the missing/corrupt-sidecar path degrades gracefully, and #1486 title-preference is intact. Closing as merged.

PR #4405, opened 2026-06-18
[security] fix(rollback): contain checkpoint restore writes

Hinotoi-agent:
## Summary

This PR hardens rollback checkpoint restore writes so restored checkpoint files cannot be redirected outside the selected workspace by symlinks planted inside that workspace.

- Confirms the rollback restore path previously used pathname-based `shutil.copy2()` writes.
- Routes restore writes through the existing workspace-anchored file helpers instead.
- Adds regression coverage for an escaping symlink target and a normal restore path.
- Adds a release-note-ready changelog entry for the behavior change.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Rollback restore follows workspace symlinks | A workspace symlink can redirect restored checkpoint content to a host path outside the workspace during checkpoint restore. | Medium |

## Before this PR

- `restore_checkpoint()` joined each checkpoint path under the workspace and called `shutil.copy2()` on that pathname.
- If the workspace entry was a symlink, the restore followed it and overwrote the symlink target.
- The workspace file API already had symlink-aware anchored helpers, but rollback restore did not use them.
- There was no regression test proving checkpoint restore could not write outside the workspace through a symlink.

## After this PR

- Each restored checkpoint file is resolved through `safe_resolve_ws()` before writing.
- Existing files are opened with `open_anchored_write_fd()` and new files with `open_anchored_create_fd()`.
- Escaping symlink targets are reported as per-file restore errors instead of being overwritten.
- Regression tests cover both the rejected symlink escape and the normal restore behavior.

## Why this matters

Rollback restore is a high-trust workspace mutation path: it takes files from a checkpoint and writes them back into a configured workspace. A malicious or compromised workspace can contain symlinks. If restore follows those symlinks, restoring a checkpoint can mutate files outside the workspace boundary, bypassing the containment guarantees used by the normal workspace file APIs.

## How this differs from related issue/PR

Hermes WebUI already has workspace file API hardening around symlinks and anchored filesystem operations. This PR applies the same containment principle to the separate rollback checkpoint restore path. The vulnerable sink is `api/rollback.py::restore_checkpoint()`, not the ordinary file editor/upload/delete routes.

## Attack flow

```text
attacker controls or influences workspace contents
    -> attacker plants workspace/victim.txt as a symlink to an outside host path
        -> authenticated user restores a checkpoint containing victim.txt
            -> rollback restore follows the symlink during pathname copy
                -> checkpoint content overwrites the outside host path
```

## Affected code

| Issue | Files |
[...]

PR #3398, opened 2026-06-02
[security] fix(workspace): keep file API reads inside the selected workspace

Hinotoi-agent:
## Summary
This PR hardens the workspace file boundary so workspace browsing and file reads stay inside the selected workspace after symlink resolution.

- Blocks external symlinks inside a workspace from escaping to host paths outside the workspace root.
- Preserves internal symlinks that resolve back under the selected workspace.
- Adds focused regression coverage for `/api/list` / workspace resolution behavior and updates cycle-detection expectations.
- Keeps the change scoped to workspace path containment; no frontend, upload extraction, or agent-runtime behavior is changed.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Workspace symlink escape through file/list APIs | A workspace containing a symlink to an outside directory could allow read-only listing or file disclosure outside the selected workspace through normal workspace file API flows. | Medium generally; High in deployments where a malicious workspace can be selected and the Web UI is exposed to another user/session. |

## Before this PR
- Workspace path resolution allowed symlinks that resolved outside the selected workspace unless they matched a small system-directory denylist.
- Workspace listing/read helpers trusted the resolved path returned by `safe_resolve_ws`.
- A symlink placed inside the selected workspace could therefore point the file APIs at content outside that workspace.
- Regression tests covered symlink cycles and inaccessible/system paths, but not the general resolved-path containment invariant.

## After this PR
- `safe_resolve_ws(root, requested)` resolves both the workspace root and requested path.
- The resolved requested path must remain under the resolved workspace root.
- External symlink traversal now raises `ValueError("Path traversal blocked: <requested>")` before list/read helpers use the target.
- Internal symlinks that still resolve within the selected workspace continue to work.
- New focused tests lock both external-symlink rejection and internal-symlink compatibility.

## Why this matters
Hermes WebUI can expose workspace contents through browser and agent-facing workspace APIs. The intended trust boundary is the selected workspace, not every path reachable by a symlink inside it. If a malicious or untrusted workspace can be selected, following external symlinks can disclose host files outside the workspace that the server process can read.

The validated impact is read-only disclosure/listing outside the selected workspace. This PR does not claim write/delete/RCE impact or an app-native symlink planting primitive.

## How this differs from related issue/PR
This is related to the same general symlink-safety family as PR #1149, but it addresses a different boundary:

- PR #1149 covered symlink recursion/cycle behavior.
- This PR enforces resolved-path containment for workspace file/list targets.
[...]

PR #4410, opened 2026-06-18
[security] fix(rollback): avoid symlinked checkpoint diff reads

Hinotoi-agent:
## Thinking Path

- Hermes WebUI rollback checkpoints are intended to show and restore files within a configured workspace boundary.
- The rollback restore path was already hardened against destination symlink writes in #4405, but the sibling diff path still rendered file contents by reading checkout paths directly.
- A checkpoint can contain a tracked symlink. `Path.is_file()` and `Path.read_text()` follow that symlink before the diff response is generated.
- The fix keeps the rollback diff useful for regular files while refusing symlink/special checkpoint entries and reading workspace-side content through the existing anchored workspace file API.
- This preserves the rollback feature while keeping diff and restore reads inside the expected checkpoint/workspace trust boundary.

## What Changed

- Added checkpoint-entry helpers in `api/rollback.py` that use the Git index mode to distinguish regular tracked files from symlinks or special entries.
- Changed `get_checkpoint_diff()` to read checkpoint content from Git blobs only after confirming the entry is a regular file, instead of reading through the checkpoint worktree path.
- Changed workspace-side diff reads to use `api.workspace.read_file_content()`, which applies the existing anchored/no-follow workspace read protections.
- Changed `restore_checkpoint()` to skip symlink/special checkpoint sources before opening them.
- Added regression coverage for:
  - checkpoint symlink targets not appearing in rollback diffs,
  - workspace symlink escapes not appearing in rollback diffs,
  - restore skipping checkpoint symlink sources.
- Added a release-note-ready security entry to `CHANGELOG.md`.

## Why It Matters

`/api/rollback/diff` returns file contents to an authenticated caller. Before this PR, a tracked checkpoint symlink could redirect that read to a readable file outside the checkpoint and workspace roots, causing the target file's contents to be rendered in the diff response.

This is a security-sensitive path because rollback/checkpoint features work near user workspaces and agent state. The path names were checkpoint-relative, but the final filesystem object could still be a symlink that crossed the intended boundary.

## Security issues covered

| Issue | Impact | Fix |
|---|---|---|
| Rollback checkpoint diff follows symlinked checkpoint entries | A diff response can disclose readable host-file contents through a tracked symlink in the checkpoint worktree | Check the Git index mode and read only regular checkpoint blobs |
| Rollback diff workspace-side read follows symlink escapes | A workspace symlink can redirect the "current file" side of a diff outside the workspace | Use the existing anchored workspace file reader |
| Restore reads symlink checkpoint sources | Restore could copy bytes from a symlink target instead of a regular checkpoint file | Skip non-regular checkpoint entries before restore opens the source |

## Before this PR

- `get_checkpoint_diff()` listed tracked checkpoint files with `git ls-files`.
- For each relative path, it built `ckpt_dir / rel_path` and called `Path.is_file()` followed by `Path.read_text()`.
- Both operations follow symlinks.
- A tracked checkpoint symlink could therefore cause the rollback diff to include bytes from outside the checkpoint/workspace root.

## After this PR

- Checkpoint entries are treated as diffable only when their Git index mode is a regular file.
- Checkpoint content is read from `git show HEAD:<path>` after the regular-file check, not through a filesystem path that might be a symlink.
- Workspace-side content is read through `read_file_content()`, which uses `safe_resolve_ws()` and anchored file-descriptor opens.
- Symlink and special checkpoint entries are skipped for both diff and restore.

## How this differs from #4405

#4405 fixed a rollback **restore destination** problem: pathname-based restore writes could follow a symlink already planted inside the workspace and write outside the selected workspace.

This PR fixes a sibling but different boundary:

- Different operation: rollback **diff/read** path instead of restore destination writes.
- Different sink: diff response content disclosure instead of outside-workspace overwrite.
- Different vulnerable code: `get_checkpoint_diff()` file-content reads instead of restore's destination copy path.
- Additional source-side hardening: restore now also skips symlink checkpoint sources before opening them.

Both fixes are needed because containing restore writes does not prevent the diff renderer from following symlinks while reading file contents.

## Attack flow

```text
1. A checkpoint worktree contains a tracked symlink such as leak.txt -> /readable/host/file.
2. An authenticated caller requests /api/rollback/diff for that checkpoint.
3. The old diff code asks whether ckpt_dir/leak.txt is a file.
4. Path.is_file() follows the symlink and returns true for a readable target.
5. Path.read_text() follows the same symlink.
6. The rollback diff response includes the symlink target's contents.
```

## Affected code

| Area | Files |
|---|---|
| Rollback checkpoint diff/read handling | `api/rollback.py` |
| Rollback symlink disclosure regressions | `tests/test_rollback_diff_symlink_disclosure.py` |
| Release note | `CHANGELOG.md` |

## Root Cause

- The rollback diff code treated a checkpoint-relative tracked path as safe once `git ls-files` returned it.
- The relative path was then reopened through normal pathname APIs.
- Normal pathname APIs follow symlinks by default.
- The code did not verify that the final checkpoint entry was a regular file before rendering its contents.
- The workspace-side diff read also bypassed the existing anchored workspace read helper.

## CVSS assessment

| Issue | CVSS v3.1 | Vector | Rationale |
|---|---:|---|---|
| Rollback checkpoint diff symlink disclosure | 6.5 Medium | `CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:N/A:N` | The endpoint is network-reachable to an authenticated WebUI user. When a checkpoint/workspace contains a symlink entry, the diff response can disclose high-confidentiality local file contents readable by the WebUI process. The patch does not claim unauthenticated access or integrity impact. |

## Safe reproduction steps

The proof uses only temporary files and a safe marker file:

```text
1. Create an isolated Hermes home and workspace.
2. Create a checkpoint repo under checkpoints/<workspace_hash>/<checkpoint_id>.
3. Create a temp file outside the workspace containing SAFE_DOCKER_SECRET_MARKER.
4. Add and commit a tracked symlink in the checkpoint: leak.txt -> outside temp file.
5. Call get_checkpoint_diff(workspace, checkpoint_id).
```

On vulnerable code, the rollback diff contains the external marker through the symlinked checkpoint entry:

```text
files_changed= [{'file': 'leak.txt', 'status': 'deleted'}]
marker_in_diff= True
--- a/leak.txt
+++ /dev/null
@@ -1,1 +0,0 @@
-SAFE_DOCKER_SECRET_MARKER
```

On this branch, the same Docker proof shows the marker is not rendered:

```text
files_changed= []
marker_in_diff= False
diff= ''
```

## Changes in this PR

- Adds `_checkpoint_entry_mode()` and `_checkpoint_entry_is_regular()` to classify checkpoint entries using Git metadata.
- Adds `_read_checkpoint_text()` to read only regular tracked checkpoint file blobs.
- Adds `_read_workspace_text()` to route workspace-side diff reads through anchored workspace reading.
- Updates `get_checkpoint_diff()` to avoid symlink-following checkpoint/workspace pathname reads.
- Updates `restore_checkpoint()` to skip non-regular checkpoint sources.
- Adds focused regression tests for diff and restore symlink handling.

## Files changed

| File | Purpose |
|---|---|
| `api/rollback.py` | Enforce regular-file checkpoint reads and anchored workspace reads |
| `tests/test_rollback_diff_symlink_disclosure.py` | Regression tests for diff disclosure and restore source-symlink skipping |
| `CHANGELOG.md` | Security release note |

## Maintainer impact

- Regular file rollback diffs continue to work.
- Symlink and special-file checkpoint entries are skipped rather than rendered or restored.
- This avoids following filesystem links across the checkpoint/workspace boundary.
- No new dependencies or frontend changes are introduced.

## Fix rationale

The security boundary should be enforced before any filesystem open. Checking the Git index mode lets the rollback code distinguish regular tracked files from symlinks without dereferencing the worktree path. Reading checkpoint content from Git blobs avoids symlink-following worktree reads. Reusing `read_file_content()` for workspace content keeps the diff path aligned with the existing workspace file API hardening.

## Type of change

- [x] Security fix
- [x] Bug fix
- [x] Regression tests
- [x] Changelog update
- [ ] UI / UX change
- [ ] Documentation-only change

## Verification

Host-local focused validation:

```bash
./scripts/test.sh tests/test_rollback_diff_symlink_disclosure.py tests/test_rollback_restore_symlink_containment.py
# 6 passed in 1.78s

./.venv/bin/ruff check api/rollback.py tests/test_rollback_diff_symlink_disclosure.py
# All checks passed!

git diff --check
# passed
```

Docker/container validation:

```bash
docker run --rm -v <repo>:/src:ro python:3.11-slim bash -lc '
  apt-get update >/dev/null
  apt-get install -y git >/dev/null
  cp -a /src /work
  cd /work
  ./scripts/test.sh tests/test_rollback_diff_symlink_disclosure.py tests/test_rollback_restore_symlink_containment.py
  git diff --check
'
# 6 passed in 0.43s
```

Docker before/after behavior proof on this branch:

```text
files_changed= []
marker_in_diff= False
diff= ''
```

## Risks / Follow-ups

- This PR intentionally skips symlink/special checkpoint entries in diff and restore rather than trying to render symlink targets. If maintainers want symlink metadata shown in rollback diffs later, that should be added as metadata-only output that does not dereference the target.
- The PR is scoped to rollback checkpoint diff/source reads. It does not change unrelated workspace browsing, upload, or session APIs.

## Disclosure notes

This is a bounded security fix for an authenticated rollback diff disclosure path. It does not claim unauthenticated access, arbitrary file write, or code execution. It is related to #4405 only by rollback/checkpoint trust-boundary family; it fixes a separate read/disclosure path.

## Model Used

OpenAI `gpt-5.5` via Hermes Agent. Notable tool use: source inspection, safe temporary proof scripts, Docker-based proof/test validation, and GitHub CLI for PR publication.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Security
+
+- **Rollback checkpoint diffs no longer follow checkpoint or workspace symlinks when rendering file contents.** The diff path previously read tracked checkpoint files and workspace files through pathname APIs, so a symlink entry could redirect the rendered diff to a readable file outside the checkpoint/workspace boundary. Diff reads now reject non-regular checkpoint entries and use the workspace's anchored file reader for workspace-side content; restore also skips symlink checkpoint sources.
+
 ## [v0.51.489] — 2026-06-18 — Release QY (outline button no longer collides with the scroll control)
 
 ### Fixed
--- a/api/rollback.py
+++ b/api/rollback.py
@@ -97,22 +97,28 @@ def _find_git() -> str:
     return shutil.which("git") or "git"
 
 
-def _checkpoint_entry_mode(git: str, ckpt_dir: Path, rel_path: str) -> int | None:
-    """Return the git index mode for a tracked checkpoint path."""
+def _checkpoint_entry_modes(git: str, ckpt_dir: Path) -> dict[str, int]:
+    """Return git index modes for tracked checkpoint paths in one pass."""
     result = subprocess.run(
-        [git, "-C", str(ckpt_dir), "ls-files", "-s", "--", rel_path],
+        [git, "-C", str(ckpt_dir), "ls-files", "-s"],
         capture_output=True, text=True, timeout=10,
     )
-    if result.returncode != 0 or not result.stdout.strip():
-        return None
-    first = result.stdout.splitlines()[0].split(maxsplit=1)[0]
-    try:
-        return int(first, 8)
-    except ValueError:
-        return None
+    if result.returncode != 0:
+        raise ValueError("Failed to list checkpoint files")
+
+    modes: dict[str, int] = {}
+    for line in result.stdout.splitlines():
+        parts = line.split(maxsplit=3)
+        if len(parts) != 4:
+            continue
+        try:
+            modes[parts[3]] = int(parts[0], 8)
+        except ValueError:
+            continue
+    return modes
 
 
-def _checkpoint_entry_is_regular(git: str, ckpt_dir: Path, rel_path: str) -> bool:
+def _checkpoint_entry_is_regular(modes: dict[str, int], rel_path: str) -> bool:
     """Return True only for regular tracked checkpoint files.
 
     Checkpoint worktrees can contain tracked symlinks. Pathname reads such as
@@ -121,38 +127,44 @@ def _checkpoint_entry_is_regular(git: str, ckpt_dir: Path, rel_path: str) -> boo
     mode as the source of truth and refuse symlink/special entries before any
     filesystem open.
     """
-    mode = _checkpoint_entry_mode(git, ckpt_dir, rel_path)
+    mode = modes.get(rel_path)
     return mode is not None and stat.S_ISREG(mode)
 
 
-def _read_checkpoint_text(git: str, ckpt_dir: Path, rel_path: str) -> str | None:
-    """Read a regular tracked checkpoint file without following worktree links."""
-    if not _checkpoint_entry_is_regular(git, ckpt_dir, rel_path):
-        return None
+def _read_checkpoint_blob(git: str, ckpt_dir: Path, rel_path: str) -> bytes | None:
+    """Read a regular tracked checkpoint blob without opening the worktree path."""
     result = subprocess.run(
         [git, "-C", str(ckpt_dir), "show", f"HEAD:{rel_path}"],
-        capture_output=True, text=True, errors="replace", timeout=10,
+        capture_output=True, timeout=10,
     )
     if result.returncode != 0:
         return None
     return result.stdout
 
 
+def _read_checkpoint_text(git: str, ckpt_dir: Path, modes: dict[str, int], rel_path: str) -> str | None:
+    """Read a regular tracked checkpoint file without following worktree links."""
+    if not _checkpoint_entry_is_regular(modes, rel_path):
+        return None
+    blob = _read_checkpoint_blob(git, ckpt_dir, rel_path)
+    if blob is None:
+        return None
+    return blob.decode(errors="replace")
+
+
 def _read_workspace_text(workspace_root: Path, rel_path: str) -> str | None:
     """Read workspace text through the anchored workspace file API.
 
-    The normal workspace helpers reject traversal and symlink escapes. Returning
-    an empty string for invalid/unreadable files keeps the diff useful without
-    echoing bytes from an escaping symlink target.
+    The normal workspace helpers reject traversal and symlink escapes. Treat
+    invalid or unreadable files as absent so the diff does not imply that the
+    workspace contains an empty file.
     """
     from api.workspace import read_file_content
 
     try:
         return read_file_content(workspace_root, rel_path)["content"]
-    except FileNotFoundError:
+    except (FileNotFoundError, OSError, ValueError):
         return None
-    except (OSError, ValueError):
-        return ""
 
 
 # ── Public API functions (called from routes.py) ────────────────────────────
@@ -257,15 +269,12 @@ def get_checkpoint_diff(workspace: str, checkpoint: str) -> dict[str, Any]:
 
     git = _find_git()
 
-    # Get list of files in the checkpoint
-    ls_result = subprocess.run(
-        [git, "-C", str(ckpt_dir), "ls-files"],
-        capture_output=True, text=True, timeout=10,
-    )
-    if ls_result.returncode != 0:
-        raise ValueError("Failed to list checkpoint files")
+    try:
+        entry_modes = _checkpoint_entry_modes(git, ckpt_dir)
+    except ValueError as e:
+        raise ValueError("Failed to list checkpoint files") from e
 
-    ckpt_files = [f for f in ls_result.stdout.strip().split("\n") if f]
+    ckpt_files = list(entry_modes)
     files_changed = []
     diff_lines = []
 
@@ -273,7 +282,7 @@ def get_checkpoint_diff(workspace: str, checkpoint: str) -> dict[str, Any]:
         # Read checkpoint version from the git object database, not via the
         # checkout path. This prevents tracked symlinks in the checkpoint from
         # redirecting diff reads to arbitrary host files.
-        ckpt_content = _read_checkpoint_text(git, ckpt_dir, rel_path)
+        ckpt_content = _read_checkpoint_text(git, ckpt_dir, entry_modes, rel_path)
         if ckpt_content is None:
             continue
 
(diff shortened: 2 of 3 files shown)


Hinotoi-agent:
CI note after opening this PR:

- The PR-specific rollback validation passed locally and in Docker; details are in the PR body.
- GitHub CI currently has two failing shards: `test (3.11, 2)` and `test (3.13, 2)`.
- Both failures are the same existing gateway/session test: `tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata`.
- This PR does not touch gateway/session code. The same focused test passes locally with `./scripts/test.sh tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata` (`1 passed in 1.78s`).
- I attempted to rerun failed GitHub jobs, but the fork author account does not have repository admin rights to rerun upstream checks.

Hinotoi-agent:
Addressed the Greptile feedback in `f42248d`:

- `_read_workspace_text()` now returns `None` for `FileNotFoundError`, `OSError`, and `ValueError`, so unreadable/invalid workspace reads are not represented as existing empty files.
- Checkpoint index modes are loaded once with `_checkpoint_entry_modes()` and reused by both diff and restore.
- Restore no longer opens checkpoint worktree paths after an index-mode check; it reads the committed blob with `git show HEAD:<path>` and writes those bytes through the anchored workspace fd.
- Added regression coverage for the post-commit checkpoint worktree symlink swap case.

Validation after the patch:

```text
./scripts/test.sh tests/test_rollback_diff_symlink_disclosure.py tests/test_rollback_restore_symlink_containment.py
7 passed in 1.89s

./.venv/bin/ruff check api/rollback.py tests/test_rollback_diff_symlink_disclosure.py
All checks passed!

git diff --check
passed
```

Docker validation:

```text
docker run --rm -v "$PWD":/work -w /work python:3.11-slim bash -lc '...'
7 passed in 0.76s
All checks passed!
git diff --check passed
```

nesquena-hermes:
## Summary

Reviewed at HEAD `f42248d3` in a read-only worktree (diff vs `origin/master`). This is a correct and complete fix for the rollback diff/restore symlink-disclosure boundary, and it's genuinely distinct from the restore-destination write fix in #4405. The core design decision — read checkpoint content from the git object database (`git show HEAD:<path>`) rather than the checkout path — is the right call: it's immune to both tracked-symlink entries and post-commit worktree swaps. I verified the dependency chain on `origin/master` and traced all four tests. Two minor non-blocking notes below; neither touches the security boundary.

## The fix is sound on both sink paths

`api/rollback.py:269-300` (diff) and `:359-385` (restore) now classify entries by git index mode before any filesystem open:

```python
def _checkpoint_entry_is_regular(modes, rel_path):
    mode = modes.get(rel_path)
    return mode is not None and stat.S_ISREG(mode)
```

- **Checkpoint side:** content comes from `_read_checkpoint_blob()` → `git show HEAD:<path>` (`:135-143`), never `ckpt_dir / rel_path`. A tracked symlink in the worktree is never dereferenced.
- **Workspace side:** `_read_workspace_text()` routes through `api.workspace.read_file_content()` (`:160-171`). I confirmed on master that this is the #3398-hardened reader: `safe_resolve_ws()` + `open_anchored_fd(..., want_dir=False)` does an O_NOFOLLOW openat-walk on every component and `fstat`s the fd (`api/workspace.py:1389-1404`), so a workspace symlink escape raises and the helper returns `None`.
- **Restore source:** `_restore_checkpoint_file()` now receives pre-fetched `bytes` from the blob read (`:322`); the old `ckpt_file.open("rb")` worktree open is gone entirely.

## TOCTOU robustness verified

`test_restore_checkpoint_reads_git_blob_after_worktree_symlink_swap` is the strongest test: it commits `file.txt`, then unlinks it and replaces the worktree entry with a symlink to an outside secret *after* commit, and asserts restore writes the committed blob (`"checkpoint blob content\n"`), not the symlink target. Because the read is `git show HEAD:file.txt`, the worktree swap is irrelevant — exactly the property you want. The workspace-escape test passes for the right reason too: the escaping `leak.txt` symlink makes `read_file_content` raise → `_read_workspace_text` returns `None` → diff falls into the "deleted" branch and renders only checkpoint content, so the secret marker never appears.

## Two non-blocking notes (independently confirmed, match Greptile)

1. **`core.quotepath` gap (functional, fail-closed).** `_checkpoint_entry_modes` parses `git ls-files -s` (`:103`). With default `core.quotepath=true`, git C-quotes non-ASCII paths (`"caf\303\251.txt"`), so the dict key is the quoted form and the later `git show HEAD:"caf\303\251.txt"` lookup fails → the entry is silently dropped from diff/restore. This fails *closed* (no disclosure), but non-ASCII filenames vanish from rollback diffs. Adding `-c core.quotepath=false` to the `ls-files` invocation closes it.

2. **Dead `mode is None` guard.** Both `_checkpoint_entry_is_regular` and the restore loop check `mode is None`, but `ckpt_files = list(entry_modes)` iterates the same dict's keys, so `entry_modes.get(rel_path)` can never miss. Harmless defensive code; could be simplified to the `S_ISREG` check alone.

## CI note

The two red shards (`test (3.11, 2)`, `test (3.13, 2)`) fail on `test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata`, which this PR doesn't touch (the diff is rollback-only). The branch's merge-base is current master `bc11cfe6`, which already carries the `#4392` gateway_sync flake fix (`1c3b1ecd`), but that fix targeted a *different* test (`StopIteration` isolation). This looks like the same test-isolation flake class rather than a regression from this PR. A maintainer re-run of the failed jobs would likely clear it.

## Verdict

Static read-only approve. The security boundary is enforced before any filesystem open on every read path (diff checkpoint, diff workspace, restore source), the git-blob approach is TOCTOU-safe, and the regression suite targets exactly the disclosed shapes. The two notes are polish, not blockers. I did not execute any code from the worktree — verification is by source trace only.

PR #3980, opened 2026-06-11
Add assistant turn anchor source normalizer

franksong2702:
## Summary

- Add inert assistant-turn source normalizer helpers to `HermesAssistantTurnAnchors`.
- Normalize live SSE-like events, replay/journal events, and settled/session payload events into one anchor event shape.
- Deduplicate repeated live + replay observations by Event Envelope identity (`event_id`, then `run_id + seq`, then browser local fallback).
- Document the Slice 2 helper contract and add focused normalizer tests.

## Lineage and tracking

- Related tracking: #3400 and #3926.
- #3400 is the parent Live-to-Final / Long-Running Session umbrella.
- #3926 is the Stable Assistant Turn Anchors rollout tracker and remains open after this PR.
- #3927 merged the Stable Assistant Turn Anchors RFC.
- #3962 was the Phase 0 scaffold PR. It is closed-unmerged because the same Phase 0 scaffold shipped through release PR #3977 / v0.51.359.
- This PR is the next inert implementation slice after #3962/#3977: it adds the source-event normalizer that later anchor registry, replay, settlement, Compact Worklog, and Transparent Stream slices can consume.

## Scope

- This is Slice 2 normalizer work only.
- It builds on the Phase 0 scaffold shipped through #3977 / v0.51.359.
- Inert by design: this PR does not wire anchors into `send()`, `attachLiveStream()`, `renderMessages()`, `S.messages`, `INFLIGHT`, or the DOM.
- Compact Worklog, Transparent Stream, settlement, replay/reload reconstruction, renderer ownership, artifacts, side effects, and usage/terminal metadata remain follow-up slices.

## Not included

- Does not close #3926.
- Does not implement the Transparent Stream UI work tracked by #3820.
- Does not address #3971 / #3903 Hide Thinking vs Worklog reasoning visibility.
- Does not address #3823 / #3821 tool-limit terminal-state behavior.
- Does not address #3899 / #3900 session-switch false streaming / live-turn restore behavior.

## Verification

- `node --check static/assistant_turn_anchors.js`
- `python -m pytest tests/test_stable_assistant_turn_anchor_normalizer.py tests/test_stable_assistant_turn_anchor_phase0.py -v --timeout=60`

Issue #2715, opened 2026-05-21
Can't remove chat items from sidebar (v0.51.106)


jahilldev:
I've just updated to the latest (v0.51.106), and I've noticed I can't open the context menu on chat items in the sidebar list anymore. When the three dots are clicked, I've noticed this in the console:

```
sessions.js?v=v0.51.106:1802 Uncaught TypeError: _pinnedSessionsLimit is not a function
    at _openSessionActionMenu (sessions.js?v=v0.51.106:1802:65)
    at menuBtn.onclick (sessions.js?v=v0.51.106:3584:9)
```

nesquena-hermes:
## Summary

Confirmed bug, introduced in v0.51.106 by the "configurable pinned sessions limit" change (#2700, merge `905b3eba`). The new helper `_pinnedSessionsLimit()` and the boot-time global `window._pinnedSessionsLimit` collide on the same name. Boot overwrites the function with a number, so every later call site that invokes `_pinnedSessionsLimit()` throws `is not a function` — including the session action menu builder.

## Code reference

Function declaration at `static/sessions.js:1484-1487`:

```js
function _pinnedSessionsLimit(){
  const limit=parseInt(window._pinnedSessionsLimit||3,10);
  return (Number.isFinite(limit)&&limit>0)?limit:3;
}
```

This is a classic-script top-level function declaration, so it installs `_pinnedSessionsLimit` as a property of the global object at script-eval time.

Then `static/boot.js:1454` runs inside `applyBotSettings()`:

```js
window._pinnedSessionsLimit=parseInt(s.pinned_sessions_limit||3,10)||3;
```

(and again in the fallback path at `static/boot.js:1543`: `window._pinnedSessionsLimit=3;`)

That replaces the function reference with a primitive number. From that point on, the call at `static/sessions.js:1802` —

```js
const pinLimitReached=!session.pinned&&_pinnedSessionCount()>=_pinnedSessionsLimit();
```

— resolves `_pinnedSessionsLimit` through the global scope chain, finds the number `3`, and throws `TypeError: _pinnedSessionsLimit is not a function`. That matches your console output exactly (`sessions.js:1802:65`).

`_pinnedSessionsLimitMessage()` at `static/sessions.js:1810` blows up the same way once the menu builder gets past line 1802.

## Diagnosis / Recommendation

[...]

PR #4411, opened 2026-06-18
Stabilize Compact Worklog with assistant turn anchors

franksong2702:
## Thinking Path

- Compact Worklog is moving from renderer-owned DOM mirrors toward one Assistant Turn Anchor `activity_scene_v1` model.
- The user-facing invariant is that live streaming, settled final answer, hard refresh, and session re-entry should show the same ordered turn: process prose, Thinking, tool rows, compression lifecycle, terminal state, then the final answer outside the folded Worklog.
- The failures in this slice were caused by live-only state surviving settlement, weak durable scene targeting, and fallback render paths rebuilding Worklog rows differently after refresh or session switch.
- This PR keeps the transcript plus persisted anchor scene as the durable source of truth, and makes both browser settlement and server hydration clean up live-only rows consistently.

## Contract Routing

Task type: runtime / replay / settled Worklog durability fix

Touched areas:

- Assistant Turn Anchor scene persistence and hydration
- Compact Worklog live-to-settled rendering
- WebUI/state.db transcript reconciliation after cancelled turns
- Tool Card hierarchy and Compact Worklog processed-time affordance

Relevant public docs:

- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/architecture/stable-assistant-turn-anchor-phase0.md`

- reload/re-entry preserves Worklog order
- settled final answer stays outside folded Worklog
- stale live-only rows do not remain `running` after settlement
- cancelled partial turns do not permanently disable later state.db reconciliation

## What Changed

- Render Compact Worklog live activity from ordered Assistant Turn Anchor scene rows instead of legacy DOM mirrors.
- Persist settled anchor scenes against the full transcript assistant message, including normalized browser refs and absolute message indexes for tail-window sessions.
- Harden anchor scene persistence:
  - browser JSON refs are canonicalized before matching
  - duplicate refs are rejected instead of falling back to a stale index
  - stale index fallback requires final-answer/candidate consistency
- Rebuild hydrated settled Worklogs from transcript-backed prose/thinking/tool rows, then merge persisted scene rows without duplicating compression lifecycle rows or stale live Thinking rows.
- Drop or seal live-only `running` thinking/prose/tool rows during settlement so natural settle and hard refresh do not show stale live activity at the bottom of the Worklog.
- Narrow cancelled-partial replay protection so it only applies while the partial/cancel marker is still the visible tail; later successful turns can merge state.db-only deltas again.
- Update Compact Worklog UI behavior:
  - processed-time anchor appears immediately and is localized
  - live header is non-clickable; settled header can expand/collapse
  - Tool Cards show command/target information at row level
  - read/search/list/web rows stay summary-only unless they fail
  - grouped tool rows align with Thinking rows
  - detail panels use the full available width

## Why It Matters

Users should see one coherent assistant turn:

- During live streaming, process text, Thinking, tools, compression, and progress rows appear in order.
- At settlement, the final answer remains outside the Worklog.
- After hard refresh or session switch, the Worklog reconstructs from durable state without moving old Thinking/tool rows to the bottom or mixing old renderer styles back in.
- A previous cancelled turn should not break future transcript reconciliation.

## Verification

- `python3 -m py_compile api/models.py api/routes.py tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py tests/test_webui_state_db_reconciliation.py`
- `node --check static/ui.js`
- `./scripts/test.sh tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py`
- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py tests/test_webui_lineage_display_merge.py tests/test_issue893_cancel_preserves_partial.py`
- `./scripts/test.sh tests/test_auto_compression_card.py tests/test_ui_tool_call_cleanup.py tests/test_turn_duration_display.py`

Local runtime validation:

- Restarted the local review server on `8787` after the final patch.
- Verified `/health` returned `status: ok` with no active runs or streams.
- Verified `/api/sessions` returned sessions from the expected `default` profile.

## Risks / Follow-ups

- Follow-up: move full `anchor_activity_scenes` out of the metadata prefix and add a lightweight scene index so `load_metadata_only()` stays cheap on long sessions.
- Follow-up: preserve user-expanded Worklog/tool-detail state if the user manually opened it during live streaming.
- This PR focuses Compact Worklog; Transparent Stream parity is not claimed here.

## Model Used

AI-assisted with OpenAI GPT-5 Codex in Codex Desktop, using local shell inspection, 8787 runtime validation, and targeted regression tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Live Compact Worklog now renders turn activity from the Assistant Turn Anchor scene instead of legacy DOM mirrors.** Process prose, reasoning, lifecycle rows, and tool boundaries flow through one ordered `activity_scene_v1` model so live activity, settled Worklog, hard refresh, and session re-entry preserve the same visible order. The final answer stays outside the folded Worklog while the Worklog appears above it.
+- **Compact Worklog interaction and Tool Cards now match the Codex-style hierarchy more closely.** The processed-time anchor appears immediately after send, starts as a localized "Thinking" placeholder, then becomes elapsed processed time; Tool Cards show command/target information at the row level, keep read/search/list/web rows summary-only unless they fail, align grouped child rows with Thinking rows, and use full-width detail panels for expandable shell/write/patch/edit rows.
+- **Assistant Turn Anchor scene persistence now targets the full transcript message.** Persisted Worklog scenes send the absolute assistant-message index even when the browser only loaded a tail window, and the server prefers a unique message reference before falling back to the normalized full index, preventing hard-refresh hydration from attaching a settled Worklog to an older assistant reply in long sessions.
+- **Assistant Turn Anchor recovery now accepts the browser's message reference format.** The anchor scene API canonicalizes incoming browser JSON refs before matching settled assistant messages, rejects ambiguous duplicate refs instead of falling back to a stale index, and includes already-received reasoning as a Thinking row in live run-journal snapshots after reconnect.
+- **Cancelled partial turns no longer permanently suppress later state.db transcript reconciliation.** The cancel replay guard now applies only while a partial assistant response plus cancellation marker are still the visible tail, so later successful turns can merge state.db-only deltas again. Anchor scene index fallback also rejects mismatched final-answer targets instead of persisting a Worklog scene onto the wrong assistant message after a stale ref miss.
+- **Settled Worklog hydration now strips live-only running rows from durable scenes.** If a persisted scene still contains a `live-reasoning:*` running Thinking row after the final answer settles, the server and browser drop it when transcript-backed settled Thinking exists; otherwise live prose/thinking/tool rows are sealed before the scene is rendered or persisted so hard refresh and natural settle do not show stale running activity at the bottom of the Worklog.
+
 ## [v0.51.513] — 2026-06-19 — Release RX (credential-pool quota status for all pooled providers)
 
 ### Added
@@ -26,7 +35,6 @@
 ### Added
 
 - **Kanban tasks can now set a workspace and manage dependencies from the WebUI (#3797).** The task-create modal gains a **Workspace kind** selector (scratch / worktree / directory) with a conditional path field (validated on create; correctly disabled on edit, which the backend does not patch), and a task's detail view gains **dependency** add/remove controls. The dependency field is an autocomplete of existing tasks (titles shown) rather than a blind ID box, rejects self-dependencies, and a dependency added to task X correctly records the linked task as X's prerequisite (parent). Dark-theme styled to match the rest of the editor. Linked-task IDs are JS-context-encoded in the action handlers. Thanks @rodboev. (Remaining task-editor parity items — skills, max runtime, dependencies at create time, workspace-path autocomplete — tracked in #4470.)
-
 ## [v0.51.509] — 2026-06-19 — Release RT (notice when the connected gateway can't do approvals)
 
 ### Added
--- a/api/models.py
+++ b/api/models.py
@@ -644,11 +644,12 @@ class Session:
                 parent_session_id: str=None,
                 worktree_path=None,
                 worktree_branch=None,
-                worktree_repo_root=None,
-                worktree_created_at=None,
-                enabled_toolsets=None,
-                composer_draft=None,
-                **kwargs):
+                 worktree_repo_root=None,
+                 worktree_created_at=None,
+                 enabled_toolsets=None,
+                 composer_draft=None,
+                 anchor_activity_scenes=None,
+                 **kwargs):
         self.session_id = session_id or uuid.uuid4().hex[:12]
         self.title = title
         self.workspace = str(Path(workspace).expanduser().resolve())
@@ -704,6 +705,7 @@ class Session:
         self.read_only = bool(kwargs.get('read_only', False))
         self.enabled_toolsets = enabled_toolsets  # List[str] or None — per-session toolset override
         self.composer_draft = composer_draft if isinstance(composer_draft, dict) else {}
+        self.anchor_activity_scenes = anchor_activity_scenes if isinstance(anchor_activity_scenes, dict) else {}
         raw_message_count = kwargs.get('message_count')
         parsed_message_count = None
         if raw_message_count is not None:
@@ -759,7 +761,7 @@ class Session:
             'parent_session_id',
             'worktree_path', 'worktree_branch', 'worktree_repo_root', 'worktree_created_at',
             'is_cli_session', 'source_tag', 'raw_source', 'session_source', 'source_label', 'read_only',
-            'enabled_toolsets', 'composer_draft',
+            'enabled_toolsets', 'composer_draft', 'anchor_activity_scenes',
         ]
         meta = {k: getattr(self, k, None) for k in METADATA_FIELDS}
         meta['message_count'] = len(self.messages or [])
@@ -2283,12 +2285,39 @@ def _cache_has_stale_unsaved_user_tail(cached, disk_session) -> bool:
         return False
     cached_messages = getattr(cached, 'messages', None) or []
     disk_messages = getattr(disk_session, 'messages', None) or []
-    if len(cached_messages) <= len(disk_messages):
-        return False
     if _last_non_tool_role(cached_messages) != 'user':
         return False
     if _last_non_tool_role(disk_messages) != 'assistant':
         return False
+    if len(cached_messages) < len(disk_messages):
+        return True
+    if len(cached_messages) == len(disk_messages):
+        # Same-length divergence is still stale: a completed assistant turn can
+        # be persisted through a sibling Session object while this inactive LRU
+        # entry still ends on the optimistic/recovered user row.
+        #
+        # Keep this narrow: only evict when the shared prefix is the same and
+        # the cached user tail is not newer than the persisted assistant.  A
+        # genuine just-submitted user message can exist briefly before the
+        # stream id is attached, and that must not be replaced by older disk
+        # state.
+        cached_tail = _last_non_tool_message(cached_messages)
+        disk_tail = _last_non_tool_message(disk_messages)
+        cached_prefix = [
+            (_message_role(message), _message_content_text(message))
+            for message in cached_messages[:-1]
+        ]
+        disk_prefix = [
+            (_message_role(message), _message_content_text(message))
+            for message in disk_messages[:-1]
+        ]
+        if cached_prefix != disk_prefix:
+            return False
+        cached_tail_ts = _message_timestamp(cached_tail)
+        disk_tail_ts = _message_timestamp(disk_tail)
+        if cached_tail_ts is not None and disk_tail_ts is not None and cached_tail_ts > disk_tail_ts:
+            return False
+        return True
 
     cached_tail = _last_non_tool_message(cached_messages)
     previous_disk_user = None
@@ -2303,6 +2332,30 @@ def _cache_has_stale_unsaved_user_tail(cached, disk_session) -> bool:
     # A genuinely new concurrent user edit must stay in memory so stale-session
     # guards can report and preserve it.
     return _message_content_text(cached_tail) == _message_content_text(previous_disk_user)
+
+
+def _anchor_scene_record_keys(session) -> set[str]:
+    records = getattr(session, 'anchor_activity_scenes', None)
+    if not isinstance(records, dict):
+        return set()
+    return {str(key) for key, value in records.items() if key and isinstance(value, dict)}
+
+
+def _anchor_scene_records_updated_at(session) -> float:
+    records = getattr(session, 'anchor_activity_scenes', None)
+    if not isinstance(records, dict):
+        return 0.0
+    latest = 0.0
+    for record in records.values():
+        if not isinstance(record, dict):
+            continue
+        try:
+            updated_at = float(record.get('updated_at') or 0)
+        except (TypeError, ValueError):
+            updated_at = 0.0
+        if updated_at > latest:
+            latest = updated_at
+    return latest
 
 
 def _cached_session_lags_disk(cached) -> bool:
@@ -2329,7 +2382,19 @@ def _cached_session_lags_disk(cached) -> bool:
     disk_count = _parse_nonnegative_int(getattr(disk_meta, '_metadata_message_count', None))
     if disk_count is None:
         disk_count = _lookup_index_message_count(sid)
-    return bool(disk_count is not None and disk_count > cached_count)
+    if disk_count is not None and disk_count > cached_count:
+        return True
+    if not getattr(cached, 'active_stream_id', None) and not getattr(cached, 'pending_user_message', None):
+        cached_scene_keys = _anchor_scene_record_keys(cached)
+        disk_scene_keys = _anchor_scene_record_keys(disk_meta)
+        if disk_scene_keys and not disk_scene_keys.issubset(cached_scene_keys):
+            return True
+        if (
+            disk_scene_keys
+            and _anchor_scene_records_updated_at(disk_meta) > _anchor_scene_records_updated_at(cached)
+        ):
+            return True
+    return False
 
 
 def get_session(sid, metadata_only=False):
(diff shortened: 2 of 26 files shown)


nesquena-hermes:
## Summary

First pass on this one — it's large (26 files, ~5k lines), so rather than skim the whole anchor-scene model I read deeply into the two concrete behavioral risks Greptile flagged (the `api/models.py` equal-length cache-eviction branch and the `_run_journal_live_snapshot` early-return removal) plus the dead-code note, all on a read-only worktree at HEAD `6c2d787a` vs `origin/master`. Both behavioral changes are correctly guarded and have direct test coverage; the third is genuinely dead but harmless. Verdict below is scoped to those paths, not a full sign-off on the whole anchor model.

## 1. Equal-length stale-tail eviction — correctly guarded (not a regression)

The old `_cache_has_stale_unsaved_user_tail` bailed when `len(cached) <= len(disk)`. The new equal-length branch (`api/models.py:2287-2311`) is the right fix for the "completed assistant turn persisted via a sibling Session while this LRU entry still ends on the optimistic user row" shape, and it's narrowed exactly where it needs to be:

```python
if len(cached_messages) == len(disk_messages):
    ...
    if cached_prefix != disk_prefix:
        return False
    cached_tail_ts = _message_timestamp(cached_tail)
    disk_tail_ts = _message_timestamp(disk_tail)
    if cached_tail_ts is not None and disk_tail_ts is not None and cached_tail_ts > disk_tail_ts:
        return False
    return True
```

The two guards — shared-prefix equality and "don't evict a cached tail newer than the persisted assistant" — protect the genuine just-submitted-before-stream-id user message. The helpers it leans on (`_message_timestamp:441`, `_last_non_tool_message:2228`, `_message_content_text:2238`) all pre-exist on master. Both directions are tested: `test_get_session_reloads_equal_count_cached_user_tail_after_saved_assistant` (reload when disk's assistant turn wins) and `test_get_session_keeps_equal_count_newer_cached_user_tail` (keep when `ts 1002.0 > 1001.0`). This resolves the prior-review concern — the content-match guard wasn't dropped, it was replaced by a prefix+timestamp pair that's actually stronger for the equal-length case.

## 2. `_run_journal_live_snapshot` always-dict return — benign

The removed `if not messages and not tool_calls: return None` (`api/routes.py:~2495`) does mean an active stream with journal events but no projected messages/tools now returns a populated dict carrying an empty `activity_rows: []` scene. I traced the consumer chain and this is safe:

- `_serverLiveSnapshotInflight` (`static/sessions.js:487`) maps empty `messages`/`tool_calls` to empty arrays — no crash.
- The scene render entry `_renderLiveAnchorActivitySceneSnapshotForStream` (`static/ui.js:9100`) only checks `scene.version==='activity_scene_v1'`, and `_anchorSceneRowsForRendering` (`:8810`) does `Array.isArray(scene.activity_rows) ? ... : []`, so an empty row list renders nothing rather than throwing.

The early `return None` guards at the top (empty `stream_id`/`summary`/`events`, `:2347-2357`) are intact, so this only changes the "active stream, events present, nothing projected yet" window — and the frontend treats that as an empty live scene. Worth a one-line comment at the old guard site noting the intentional change, but not a correctness issue.

## 3. Dead `renderOrder` WeakMap (confirmed, harmless)

`_anchorSceneRowsForRendering` (`static/ui.js:8815`) builds `const renderOrder=new WeakMap();` and `.set()`s it twice (`:8838`, `:8842`) but never reads it — `grep -n renderOrder static/ui.js` returns only those three lines. The `:8842` write is also off-by-one (`renderOrder.set(row, out.length)` runs *after* `out.push(row)`, so it stores `length`, not the `length-1` index just used). Since nothing consumes the map it's purely dead weight; safe to delete the three lines. If a later slice needs render-order tracking, fix the off-by-one then.

## Verdict

The two paths most likely to cause subtle regressions are both fenced correctly and covered by `tests/test_webui_state_db_reconciliation.py` (equal-count reload/keep, plus `test_get_session_reloads_when_disk_adds_anchor_scene_without_new_messages` for the scene-key lag in `_cached_session_lags_disk`). The `_sidecar_has_terminal_partial_error` gate (`api/models.py:4592`) is also narrow — it only short-circuits state.db replay when the partial-error is the visible tail, which `test_tail_cancelled_partial_blocks_state_db_replay` vs `test_historical_cancelled_partial_does_not_disable_later_state_db_merge` distinguish. I have not exercised the full live→settled→refresh round-trip (that needs running the suite, which I don't do on PR worktrees), so this is a static targeted review: the flagged risks look resolved; recommend the dead-code trim and a clarifying comment at the snapshot guard before merge.

franksong2702:
Follow-up on the maintainer cleanup note: pushed `dc564d48d` with a comment at `_run_journal_live_snapshot` documenting why active streams with journal events but no projected message/tool rows still return an empty `activity_scene_v1` snapshot. The frontend treats that as nothing renderable yet while preserving the live cursor.

I also checked the prior `renderOrder` note; that dead WeakMap is no longer present on the current head, so no code deletion was needed there.

Verification:
- `.venv/bin/python -m py_compile api/routes.py`
- `git diff --check -- api/routes.py`

GitHub read-back on `dc564d48d`: `MERGEABLE`, `mergeStateStatus=CLEAN`; browser-smoke, lint, and all Python 3.11/3.12/3.13 shards passed. Greptile re-review is confidence 5/5 / safe to merge; the remaining locale fallback note is a post-#4411 polish follow-up rather than a blocker for this anchor-renderer vehicle.

nesquena-hermes:
Re-checked the two follow-up items on a read-only worktree at HEAD `dc564d48` — both match what you described.

## Empty-snapshot documentation

`dc564d48` adds the clarifying comment right above the return in `_run_journal_live_snapshot` (`api/routes.py:2840`):

```python
# Keep returning a live snapshot even when the journal has events but no
# projected message/tool rows yet. The frontend treats the empty activity
# scene as "nothing renderable yet" while preserving the live cursor.
return {
    "session_id": session_id,
    ...
    "anchor_activity_scene": {"version": "activity_scene_v1", "mode": "compact_worklog", ...},
}
```

This is the right framing: the function returns a populated snapshot dict carrying an empty `activity_scene_v1` (status `running`, empty `final_answer`/`source_message_refs`) rather than `None`, so a live stream that has emitted journal events but hasn't projected any message/tool rows yet still advances `last_seq`/`last_event_id` and keeps the cursor live. Returning `None` there would have stalled the cursor on early-stream ticks. The comment now makes that intent explicit for the next reader.

## Dead WeakMap

Confirmed — `grep -n renderOrder static/*.js` returns nothing at this HEAD, so there's no leftover `renderOrder` WeakMap to delete. The earlier note was against a stale revision; nothing to do.

Both my original behavioral concerns (the `api/models.py` equal-length eviction branch and this early-return path) are resolved. No further blockers from my side on these two points.

franksong2702:
Conflict refresh is pushed in `26a9c8116`.

What changed in this refresh:
- Merged latest `origin/master` into this PR head.
- Resolved conflicts in `CHANGELOG.md`, `api/routes.py`, and `static/messages.js`.
- Kept the post-`/api/chat/start` UI/bookkeeping path error-swallowed so it cannot surface a false send failure after the stream has already started.

Local verification:
- `git diff --check`
- `node --check static/ui.js`
- `.venv/bin/python -m py_compile api/routes.py api/models.py api/config.py tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py tests/test_webui_state_db_reconciliation.py tests/test_inflight_send_start_race.py`
- `./scripts/test.sh tests/test_inflight_send_start_race.py tests/test_inflight_stream_reuse.py tests/test_stale_stream_cleanup.py tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py tests/test_webui_state_db_reconciliation.py tests/test_ui_tool_call_cleanup.py tests/test_turn_duration_display.py` (`188 passed`)

GitHub read-back on `26a9c8116`: `MERGEABLE`, `mergeStateStatus=CLEAN`; browser-smoke, lint, and all Python 3.11/3.12/3.13 shards passed.

franksong2702:
Conflict refresh pushed in `73a78f67b`.

What changed:
- merged latest `origin/master` into this PR head
- resolved the `CHANGELOG.md` conflict while preserving the Stable Assistant Turn Anchor / Compact Worklog Unreleased entries
- kept the post-`/api/chat/start` optional UI/bookkeeping path error-swallowed so it cannot surface a false send failure after the stream has already started

Local verification:
- `git diff --check`
- `node --check static/ui.js`
- `.venv/bin/python -m py_compile api/routes.py api/models.py api/config.py tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py tests/test_webui_state_db_reconciliation.py tests/test_inflight_send_start_race.py`
- `./scripts/test.sh tests/test_inflight_send_start_race.py tests/test_inflight_stream_reuse.py tests/test_stale_stream_cleanup.py tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py tests/test_webui_state_db_reconciliation.py tests/test_ui_tool_call_cleanup.py tests/test_turn_duration_display.py` (`188 passed`)

GitHub read-back: `MERGEABLE`, `mergeStateStatus=CLEAN`; browser-smoke, lint, and all Python 3.11/3.12/3.13 shards are successful.

franksong2702:
Latest base refresh pushed in `24f63b66a` after the `v0.51.530` profiles-route fix landed on `master`.

Local verification on this final head:
- `git diff --check`
- `node --check static/ui.js`
- `.venv/bin/python -m py_compile api/routes.py api/models.py api/config.py tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py tests/test_webui_state_db_reconciliation.py tests/test_inflight_send_start_race.py tests/test_profiles_route_endpoint.py`
- `./scripts/test.sh tests/test_inflight_send_start_race.py tests/test_inflight_stream_reuse.py tests/test_stale_stream_cleanup.py tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py tests/test_webui_state_db_reconciliation.py tests/test_ui_tool_call_cleanup.py tests/test_turn_duration_display.py tests/test_profiles_route_endpoint.py` (`189 passed`)
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master --strict`

GitHub read-back on `24f63b66a`: `MERGEABLE`, `mergeStateStatus=CLEAN`; browser-smoke, lint, and all Python 3.11/3.12/3.13 shards are successful.

nesquena-hermes:
Thanks @franksong2702 — this is excellent architectural work and it's close. I ran the full deep-review treatment: rebase (byte-identical, the CONFLICTING was CHANGELOG-only), the PR's own anchor suite (143 passing), the **full pytest suite (9774 passed, 0 failed)**, a Codex + Opus architectural gate, and a live browser drive.

**The architecture is sound and the four invariants all hold up:**
1. ✅ Live order — anchor scene owns the visible row order.
2. ✅ Settled final answer renders outside the folded Worklog.
3. ✅ No stale `running` rows after settle/refresh — I traced every terminal path (done/apperror/cancel/connection-lost); each attaches the projected scene and `ui.js:8969` forces `done` on settled tool rows, so no live spinner survives settlement or hard-refresh hydration.
4. ✅ Cancelled-partial reconciliation re-enables state.db merges for later successful turns.

The **backward-compat / legacy-fallback is the strongest part** — the anchor only takes over behind `data-anchor-scene-live-owner="1"`, all legacy writers branch on it, projection returns null when no scene exists, and `sessions.js` calls `renderMessages()` first then attempts anchor restore, so a missing/failed scene can't blank a session. I confirmed a no-anchor-scene session renders cleanly in the browser. No second-source-of-truth desync, no per-token blowup, no dropped/reordered reasoning, i18n is en-fallback-safe.

**Two code fixes + one test ask before merge** (none are design flaws — they're narrow edges in the new anchor paths):

### 1. (must-fix) Anchor-owned auto-compression ignores `phase:'done'`
`static/ui.js:9756` / `static/messages.js:3572` — `_completeAutomaticCompressionOnLiveProgress()` passes `phase:'done'` to `appendLiveCompressionCard()`, but the anchor-owned branch ignores that state and only re-renders the existing scene. If later token/reasoning/tool progress proves compression finished but **no explicit `compressed` SSE arrives**, no completed-compression row is fed to the anchor registry → the "Compressing context" row can linger.
**Fix:** in the anchor-owned path (`messages.js:3572`), apply an idempotent synthetic completed `compressed` anchor event (stable local id / guard) before rendering, instead of letting `appendLiveCompressionCard()` drop `phase:'done'`. Please add a regression for "later progress completes compression with no explicit `compressed` SSE."

### 2. (must-fix) Hydration `by_index` read-path is missing the write-path's consistency gate
`api/routes.py:3497` — `record = by_ref.get(...) or by_index.get(absolute_idx)`. The **write** path gates the stale-index fallback on final-answer/candidate consistency and safe-drops when it fails; the **read** (hydration) path has no such gate. On a transcript shift, the stored `activity_rows` snapshot can attach to the wrong assistant message. The final answer stays correct (`_complete_hydrated_anchor_scene` rebuilds it from the message's actual text), but stale intermediate rows can leak onto the wrong turn.
**Fix:** apply the same final-answer/candidate consistency check to the `by_index` branch before accepting the match.

### 3. (required by our review bar) DOM-executed render tests for the live→settled transition
`tests/test_live_to_final_anchor_visible_order.py` (+718) is thorough but it's **pure source-substring assertion** — it never builds a DOM or runs the state machine, so the front-end DOM-render invariants (1–3) would still pass even if the rendered behavior regressed, as long as the matched code strings remain. For a rewrite of the live chat render path, please add 1–2 tests that actually execute the render against a DOM (Node/jsdom-style, like the registry tests already do) for the live→settled transition.

Two non-blocking LOW items for whenever: the persisted `activity_rows` can coexist with a fresh transcript row on a text-key collision (`routes.py:3445-3450` — prefer transcript-derived rows on collision); and the 10 en-fallback locales could get real worklog-string translations.

Land 1–3 and I'll fast-track the re-review (I've cached the full review, so the next pass is just the delta + a re-gate, not a re-read). Really strong work — this is a real improvement to the Worklog render path.

franksong2702:
Delta for the requested pre-merge items is pushed in `2f20b9ee3` after refreshing onto latest `master`.

Addressed:
- Anchor-owned auto-compression now applies an idempotent synthetic `compressed` anchor event when later live progress proves compression completed without an explicit `compressed` SSE, so the live anchor scene cannot leave a stale "Compressing context" row running.
- Hydration now gates the `by_index` fallback with the same final-answer/candidate consistency check as the write path before accepting a stale-index scene match.
- Added DOM-executed Node coverage for the live compression row transition: running `compressing` renders as running, a later `compressed` row replaces it as done, and the successful compression lifecycle row is removed from the settled Worklog.
- Also preserved the live `Processed` timer start by passing `pending_started_at` into the anchor-owned live Worklog group.

Verification:
- `node --check static/ui.js`
- `python3 -m py_compile api/routes.py tests/test_anchor_scene_persistence.py tests/test_auto_compression_card.py tests/test_live_to_final_anchor_visible_order.py`
- `./scripts/test.sh tests/test_anchor_scene_persistence.py tests/test_auto_compression_card.py tests/test_live_to_final_anchor_visible_order.py` (`119 passed`)
- `./scripts/test.sh tests/test_stable_assistant_turn_anchor_registry.py tests/test_renderer_js_behaviour.py` (`79 passed`)
- Claude Code review on the current diff: no blockers; it only repeated the non-blocking locale fallback caveat.

GitHub read-back on `2f20b9ee3`: `mergeable=true`, `mergeStateStatus=CLEAN`; browser-smoke, lint, and the Python 3.11/3.12/3.13 matrix are all successful.

nesquena-hermes:
Shipped in **v0.51.543** (Release TB) via release #4564 — cherry-picked/rebased with your authorship preserved. 🎉

This is the culmination of the whole Stable Assistant Turn Anchors arc (#3927 RFC → #3980/#4037/#4092/#4093/#4108/#4120/#4126 → this capstone). Excellent architectural work — the `activity_scene_v1` model is clean, the backward-compat fall-through for pre-anchor sessions is graceful, and the schema sanitization (version/rows/bytes/LRU) is well-bounded.

I ran a full deep-review-to-ship pass (Codex + Opus + complete pytest suite + live browser testing including real mock-LLM streaming). Three issues were found and fixed on the branch before merge, each with a regression test — crediting you as co-author on all:

1. **BRICK — infinite recursion in the live anchor-scene `window` exports** (`ui.js`). The three live-render functions were re-exported via same-name wrappers (`window.X = function(){ return X() }`); in a classic script a top-level `function X(){}` *is* `window.X`, so the wrapper reassigns the global to itself → `RangeError: Maximum call stack size exceeded` on every live render / session reattach / snapshot restore. This is the #2715/#2771 brick class — invisible to pytest (no JS execution). Replaced with direct assignment; verified live in-browser (pre-fix throws, post-fix returns cleanly).
2. **Cross-profile write on `/api/session/anchor-scene`** — the handler loaded the session by id with no profile scoping and persisted without the active-profile visibility check that `GET /api/session` applies, so a request under profile A could write anchor scenes onto a profile-B session. Added the guard (404, no write) + regression test.
3. **Read-side ref-ambiguity** in `_hydrate_anchor_activity_scenes` — mirrored the write-side guard so a duplicate content-digest ref can't double-render a worklog.

Verification: full suite **9835 passed**, Codex clean, Opus SHIP, QA mock-LLM streaming loop passed, live re-entry worklog render visually verified, and legacy (no-anchor) sessions confirmed to render via the unchanged legacy path. Thanks @franksong2702 — this was a big one, and it landed clean.

PR #4412, opened 2026-06-18
Copilot/fix start script tls error

tomas-forgac:
# Intro

A new PR replacing this one: https://github.com/nesquena/hermes-webui/pull/4361 

# What's built 
A single shared, TLS-aware probe — `scripts/lib/health_probe.sh` — used by all five launcher sites, so the scheme (`http`/`https`) is derived once from `HERMES_WEBUI_TLS_CERT`/`KEY` (mirroring `server.py`/`api/config.py`).

**Probe order when TLS is configured:** verified HTTPS → unverified HTTPS (self-signed) → plain HTTP.

**Wired into:**

- `start.sh` (already-running check, now scheme-aware message too)
- `ctl.sh` `_health_line` (status)
- `bootstrap.py` `wait_for_health` (Python equivalent, preserves the B310 scheme-validation contract)
- `scripts/wsl/hermes_webui_autostart.sh`
- `Dockerfile` HEALTHCHECK

## Addressing four previous review issues

1. **Warning consistency** — uniform policy everywhere: explicit `HERMES_WEBUI_TLS_INSECURE_PROBE=1` is **silent**; auto-detected self-signed fallback prints **one** warning (once per process).
2. **Docker timing** — each attempt stays at `--max-time 2`; HEALTHCHECK `--timeout` raised 5s→8s to fit the sequential attempts with margin.
3. **`ctl.sh status`&#160;crash on readonly&#160;`.env`&#160;vars** — the `.env` loader now filters `UID/GID/EUID/EGID/PPID` both before sourcing *and* in the preserve/re-export loop, so `status` (which now loads `.env` to learn the scheme) never aborts on `UID=1000`.
4. **HTTPS-forever after HTTP fallback** — all probe paths fall back to plain HTTP when HTTPS attempts fail, honoring `server.py`'s tested fallback-on-unloadable-cert contract.

## Verification

- New `tests/test_tls_aware_probe.py` (11 tests) covers scheme detection, plain HTTP, self-signed + warning, silent insecure opt-in, HTTPS→HTTP fallback (shell + Python), and the `ctl.sh status` readonly-`.env` regression.
- 130 related tests pass; helper is shellcheck-clean; bash 3.2 compatibility preserved (no arrays in the helper).
- CHANGELOG updated; CodeQL clean; the one "real" review comment (`/apptoo` "typo") is a false positive — `/apptoo` is the actual `WORKDIR`/`COPY` target.


nesquena-hermes:
## Summary

Reviewed at HEAD `62b55b53` in a read-only worktree (diff vs `origin/master`), focused on the `bootstrap.py` Python probe since that's the path most likely to silently hang a launch, plus a cross-check of the scheme-detection contract against `server.py` and `api/config.py`. The TLS-aware logic is correct, the scheme derivation matches the server exactly, and the one concrete bug raised in the Greptile thread (insecure-opt-in spinning to timeout) does **not** exist at this HEAD. I did not execute any worktree code; this is a static trace.

## Scheme detection matches the server contract

`_tls_probe_enabled()` (`bootstrap.py:294-297`) is the same predicate the server uses:

```python
def _tls_probe_enabled() -> bool:
    """Mirror api.config.TLS_ENABLED: HTTPS when both cert and key are set."""
    return bool(os.getenv("HERMES_WEBUI_TLS_CERT", "").strip()) and bool(
        os.getenv("HERMES_WEBUI_TLS_KEY", "").strip()
    )
```

That mirrors `api/config.py:67-69` (`TLS_ENABLED = TLS_CERT is not None and TLS_KEY is not None`, both read with `.strip() or None`). The HTTP-last probe ordering honors `server.py:643-655`, where a failed `ctx.load_cert_chain(TLS_CERT, TLS_KEY)` sets `scheme = 'http'` and serves plain HTTP, so probing verified-HTTPS then unverified-HTTPS then HTTP is the right precedence, not "HTTPS forever."

## The flagged insecure+fallback bug is a false positive at HEAD

The Greptile note claims insecure-probe opt-in plus server-fell-back-to-HTTP spins to timeout. Reading the actual control flow (`bootstrap.py:348-371`), the HTTP fallback at line 369 sits inside the `else:` (https-configured) branch, after the `insecure_optin` if/else, at the same indent level. So it is reachable from both sub-paths. Structure, with line numbers:

- L353 `if insecure_optin:` then L354 unverified-HTTPS probe, return on success
- L356 `else:` then L357 verified-HTTPS probe, L359 unverified-HTTPS probe + one-time warning, return on success
- L369 `if _health_ok(http_url, verify=True): return True` — runs whenever the https branch above did not return, including the insecure-opt-in miss

When the opt-in is set and the unverified HTTPS probe at L354 misses, control falls through past the if/else to L369 and succeeds via HTTP. So the insecure opt-in still gets the fallback. Either the comment was acted on before `0da9d31e`, or it was misread, but at `62b55b53` the path is sound.

## Test coverage is solid

`test_bootstrap_probe_http_fallback` (`tests/test_tls_aware_probe.py:179`) sets both TLS env vars but runs a plain-HTTP `_Server()` and asserts `wait_for_health` on the https URL returns True within 8s, which is the exact "server fell back to HTTP" regression. `test_helper_https_falls_back_to_http_when_server_is_plain_http` (`:156`) covers the shell helper equivalent, and `test_helper_insecure_optin_is_silent` (`:142`) confirms the silent-by-contract opt-in. One gap: no test combines the insecure opt-in with the HTTP-fallback server, the precise path Greptile worried about. Since the code is correct, a one-line variant of `test_bootstrap_probe_http_fallback` that also sets the opt-in env var would make the no-hang guarantee explicit and prevent a future refactor from reintroducing an indent bug.

## Verdict

Static read-only approve on the Python probe path. Scheme detection mirrors `api/config.py` and `server.py`, the probe precedence honors the server's cert-failure fallback, and the insecure-opt-in fallback bug isn't present at HEAD. Recommend adding the insecure+HTTP-fallback test variant to lock the reachability. I didn't run the shell helper or the suite from the worktree, so the `ctl.sh` and `.env` readonly-var path plus shellcheck cleanliness are taken on the PR's word, worth a maintainer spot-check of `tests/test_ctl_script.py` before merge.

Issue #4293, opened 2026-06-16
bug(sidebar): archiving parent conversation leaves child/delegate sessions visible or stuck


santastabber:
## Summary

When a conversation has child sessions (for example delegated worker sessions, fork/subagent child sessions, or other parent-linked companion sessions), archiving the parent conversation does not hide/archive the child sessions. The child rows remain visible in the sidebar, often as orphaned child rows. Archiving those child rows individually can still leave them visible or make the cleanup flow feel stuck.

This makes sidebar cleanup painful for parent conversations that spawned many child sessions.

## Reproduction

1. Create or open a conversation that has multiple child sessions associated with it.
2. Confirm the sidebar shows those child sessions nested under the parent row.
3. Archive the parent conversation while archived conversations are hidden.
4. Observe the sidebar after the session list refreshes.
5. Try archiving remaining child rows individually.

## Actual behavior

- The parent row is archived/hidden.
- Child sessions associated with that parent can remain visible in the main sidebar.
- Once the parent row is hidden, the child rows may appear as orphaned top-level child/session rows.
- Individually archiving child rows does not reliably make the visible clutter disappear.

## Expected behavior

Archiving a parent conversation should not leave visible child-session clutter behind.

At minimum, when archived conversations are hidden:

- child sessions whose parent is archived should also be hidden, or
- the archive action should offer/apply a cascade archive for child sessions, or
- the sidebar render path should suppress orphaned child rows whose parent is archived/hidden.

The user should not need to manually archive every child session after archiving the parent.

## Suspected cause / code pointers

The archive endpoint appears to archive only the session id it is given:

- `api/routes.py`: `/api/session/archive`
- The endpoint sets `s.archived = ...` for one session and saves it.
- It does not cascade to child sessions linked by `parent_session_id`.

The sidebar intentionally groups child sessions under parent rows, but child attachment happens after top-level archive filtering:

- `static/sessions.js`: `_partitionSidebarSessionRows(...)`
  - filters archived top-level rows based on `_showArchived`
- `static/sessions.js`: `_renderSidebarRowsFromRawSessions(...)`
- `static/sessions.js`: `_collapseSessionLineageForSidebar(...)`
- `static/sessions.js`: `_attachChildSessionsToSidebarRows(...)`
  - attaches child sessions under visible parents
  - when the parent is absent/hidden, child sessions can become orphan rows

[...]

PR #4414, opened 2026-06-18
fix(sidebar): hide child rows for archived parents

santastabber:
## Thinking Path

- Hermes WebUI sidebar rows are filtered for archived state before child/delegate rows are attached.
- That meant archiving a parent could hide the parent while leaving child, delegate, or nested fork rows behind as top-level orphans.
- The fix should be render-path only: do not mutate child sessions on disk, and do not change compression-lineage collapse semantics.
- Child-session decoration is derived during render, so stale copied `_child_sessions` metadata should also be cleared each rebuild.
- Source-tab counts need to use the same reference set as the rendered rows, otherwise hidden archived parents can still inflate inactive tab counts.

## What Changed

- Extended `_attachChildSessionsToSidebarRows()` to accept a full reference row set separate from the currently visible rows.
- Suppressed child/delegate/fork rows whose archived ancestor is hidden and no visible parent row, lineage segment, or lineage root can attach them.
- Cleared render-derived child-session decoration fields when rebuilding sidebar rows so archived children disappear immediately.
- Passed source-specific profile-filtered references for WebUI/CLI source-tab counts.
- Added regression coverage for:
  - direct archived parent hiding child/fork orphans,
  - nested child/fork descendants under an archived hidden ancestor,
  - inactive source-tab counts using their own reference set,
  - stale decorated child rows being cleared after archive,
  - updated partition/render contracts.
- Added an Unreleased changelog entry for #4293.

## Why It Matters

Archiving a conversation should clean up the visible sidebar state. Users should not have to manually chase child, delegate, or fork rows after hiding the parent conversation, and archived hidden subtrees should not continue to affect source-tab counts.

## Verification

- `./scripts/test.sh tests/test_session_lineage_collapse.py tests/test_sidebar_session_partition.py tests/test_465_session_branching.py tests/test_issue4385_cron_archive_reappears.py tests/test_session_import_cli_fallback_model.py tests/test_issue3988_show_cli_sessions_default.py -q`
  - `97 passed`
- `git diff --check`
  - `ruff_lint: no changed .py files vs origin/master (...) OK.`
- `./scripts/test.sh tests/test_static_js_runtime_lint.py tests/test_ruff_forward_lint.py -q`
  - `4 passed, 1 skipped`
- Full local `./scripts/test.sh`
  - `9323 passed, 103 skipped, 1 xfailed, 2 xpassed, 8 warnings, 16 subtests passed`
  - Local environment had one unrelated remaining failure in `tests/test_tls_support.py::TestTLSEndToEnd::test_tls_startup_failure_fallback_to_http`: the server stdout was flooded by missing optional agent dependencies (`requests`/`websockets`) before the TLS warning string, so the test's 2000-byte stdout read did not include `TLS setup failed`. No touched files participate in TLS startup.
  - A second full-suite failure, `tests/test_v050259_sessiondb_fd_leak.py::test_session_db_close_is_idempotent`, passed on focused rerun.
- Independent agent review of the final diff: PASS, no blockers, no secrets/private paths found.
- Diff secret/private-path scan: no matches in added diff.

## Risks / Follow-ups

- This intentionally does not cascade archive state to child sessions on disk; it only hides orphaned archived-parent subtrees while archived rows are hidden.
- Archived children remain discoverable again when archived conversations are shown.
- Follow-up could add explicit UI/browser screenshots, but this is a sidebar render-state fix covered by runtime-style JS tests.

## Contract Routing

Task type: sidebar/session metadata render-state bug fix.

Touched areas:
- `static/sessions.js`
- `tests/test_session_lineage_collapse.py`
- `tests/test_sidebar_session_partition.py`
- `CHANGELOG.md`

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/UIUX-GUIDE.md`

Scope boundaries:
- No server-side archive cascade.
- No changes to persisted session metadata or compression-lineage contracts.
- No new dependencies or build steps.

- Runtime-style sidebar tests for archived parent child/fork suppression.
- Existing lineage/fork/cross-surface/sidebar partition tests remain green.

## Model Used

AI-assisted with OpenAI GPT-5.5 through Hermes Agent, including tool-driven local tests and independent subagent review. No local provider secrets or private environment details are required for the change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Archiving a parent conversation no longer leaves child, delegate, or fork rows stranded in the sidebar (#4293).** The sidebar now checks the active project/session set before rendering orphan child rows, suppresses children whose archived parent is intentionally hidden, and clears stale child-session decorations during each render rebuild.
+
 ## [v0.51.498] — 2026-06-18 — Release RH (profile runtime env no longer clobbers core shell identity)
 
 ### Fixed
@@ -46,7 +50,6 @@
 ### Fixed
 
 - **Reloading a session with a large model context window no longer snaps it back to the 256k fallback (#4248).** The session reload path now resolves context-window metadata with the same base URL / API-key lookup inputs used by streaming saves, and it refuses to overwrite a larger persisted context window with the generic 256k fallback. The reload model-identity check also normalizes slash-qualified model ids (`provider/model`, OpenRouter-style) so a slash-stored model resolving to its bare id is not mistaken for a model change. Thanks @franksong2702.
-
 ## [v0.51.489] — 2026-06-18 — Release QY (outline button no longer collides with the scroll control)
 
 ### Fixed
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -4720,12 +4720,25 @@ function _sessionStateTooltip({isStreaming=false,hasUnread=false}={}){
   return '';
 }
 
-function _attachChildSessionsToSidebarRows(collapsedRows, rawSessions){
-  const sessionIdsInList=new Set((rawSessions||[]).map(s=>s&&s.session_id).filter(Boolean));
-  const rawSessionsById=new Map((rawSessions||[]).filter(s=>s&&s.session_id).map(s=>[s.session_id,s]));
+function _attachChildSessionsToSidebarRows(collapsedRows, rawSessions, rawReferenceSessions){
+  const referenceSessions=Array.isArray(rawReferenceSessions)?rawReferenceSessions:(rawSessions||[]);
+  const sessionIdsInList=new Set(referenceSessions.map(s=>s&&s.session_id).filter(Boolean));
+  const rawSessionsById=new Map(referenceSessions.filter(s=>s&&s.session_id).map(s=>[s.session_id,s]));
+  const cleanSidebarRow=(s)=>{
+    const row={...s};
+    // Child-session decoration is render-derived.  Drop stale copies so an
+    // archived child disappears immediately on the next list rebuild instead of
+    // lingering under a copied parent row (#4293).
+    delete row._child_sessions;
+    delete row._child_session_count;
+    delete row._child_session_streaming;
+    delete row._child_session_has_unread;
+    delete row._child_session_attention;
+    return row;
+  };
   const rows=(collapsedRows||[])
     .filter(s=>!_isChildSession(s)&&((s&&s.pinned)||!_isForkWithResolvableParent(s, sessionIdsInList)))
-    .map(s=>({...s}));
+    .map(cleanSidebarRow);
   const isChildStreaming=(childRow)=>typeof _isSessionEffectivelyStreaming==='function'
     ? _isSessionEffectivelyStreaming(childRow)
     : !!(childRow&&(childRow.active_stream_id||childRow.pending_user_message));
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Pulled this branch into a read-only worktree and read the full `_attachChildSessionsToSidebarRows` / `_partitionSidebarSessionRows` / `renderSessionListFromCache` chain in `static/sessions.js` against `origin/master`. The core suppression logic is sound — `hiddenArchivedChildTree` plus the depth-sorted `attachQueue` correctly drops descendants of an archived-hidden parent, and `cleanSidebarRow` clearing stale `_child_sessions` decoration is the right call for #4293. Two things to confirm before merge, both downstream of the new `profileFiltered` reference set.

## 1. `_countRenderedSidebarRowsFromRawSessions` is now dead code (confirmed)

On `origin/master` it had two live callers (`sessions.js:5211`, `:5214`). This PR rewires both to `_renderSidebarRowsFromRawSessions(..., profileFiltered).length`, leaving zero production callers:

```js
// renderSessionListFromCache (HEAD)
: _renderSidebarRowsFromRawSessions(webuiSessionsRaw, webuiProfileFiltered).length;
...
: _renderSidebarRowsFromRawSessions(cliSessionsRaw, cliProfileFiltered).length;
```

But `tests/test_sidebar_session_partition.py:71` still asserts the helper's body, and the helper calls `_renderSidebarRowsFromRawSessions(sessionsRaw)` with **no** reference arg — so any future caller silently skips the archived-ancestor suppression this PR adds. Either delete the helper + its test assertion, or route it through `profileFiltered` like the live path. Keeping it pins a stale contract that contradicts the fix.

## 2. Cross-project fork suppression — narrow but real regression

`profileFiltered` is pushed *before* the project-filter `continue` (`sessions.js:5176`), so it spans **all projects and archived rows**, whereas `sessionsRaw` is the active-project, non-archived set (`:5187`). Broadening the reference is necessary so the archived parent lands in `rawSessionsById` for `hasHiddenArchivedAncestor`. The side effect is that `sessionIdsInList` now also contains cross-project parents, and that set feeds `_isForkWithResolvableParent` (`:4479`):

```js
return !!(s&&s.session_source==='fork'&&s.parent_session_id&&sessionIdsInList&&sessionIdsInList.has(s.parent_session_id));
```

Walk the case: a fork in the **active** project whose parent lives in **another** project, with a project filter active. Previously the parent wasn't in `sessionIdsInList`, so the fork stayed a normal top-level row. Now the parent *is* in `sessionIdsInList`, so the fork is filtered out of `rows` (`:4740`), its parent row isn't visible (different project → no `parentRow`), and then `hasHiddenArchivedAncestor` walks `rawSessionsById`. If that cross-project parent is `archived`, it returns true and the fork is added to `hiddenArchivedChildTree` and **dropped entirely** — a visible active-project session disappears because an unrelated other-project ancestor happens to be archived. If the parent isn't archived, the fork falls through to `orphans` and renders with `_orphan_child_session` styling it didn't have before.

The cleanest fix is to scope the reference set to "active project, including archived" rather than the full cross-project `profileFiltered`. That preserves archived-parent detection (the whole point) without pulling other projects into `sessionIdsInList`. `_partitionSidebarSessionRows` doesn't currently expose such a set, so it'd need a new accumulator (push to it after the project filter but skip the `!_showArchived && s.archived` drop), or alternatively gate `hasHiddenArchivedAncestor`'s walk to `parentSid` rows whose `project_id` matches the active filter.

## Verification

A regression test for case 2 belongs in `tests/test_session_lineage_collapse.py`: active-project fork + archived parent in a different project + project filter active → assert the fork still renders (currently it would be suppressed). The four new tests cover same-project lineage well but none exercise the cross-project boundary that the `profileFiltered` widening introduced.

santastabber:
Addressed the Greptile P2 on the latest push: removed the unused `_countRenderedSidebarRowsFromRawSessions` helper and updated the sidebar partition contract test so future count paths use source-specific reference rows. Re-ran focused verification locally: `./scripts/test.sh tests/test_session_lineage_collapse.py tests/test_sidebar_session_partition.py tests/test_465_session_branching.py tests/test_issue4385_cron_archive_reappears.py tests/test_session_import_cli_fallback_model.py tests/test_issue3988_show_cli_sessions_default.py -q` → 97 passed; `node --check static/sessions.js`, `git diff --check`, and `python3 scripts/ruff_lint.py --diff origin/master` also passed.

santastabber:
Addressed both review points on the latest push. The unused `_countRenderedSidebarRowsFromRawSessions` helper is removed, and the render reference set is now scoped to the active project (including archived rows) instead of the whole profile/source set, so cross-project archived parents no longer reclassify or suppress an active-project fork. Added a regression for that boundary in `tests/test_session_lineage_collapse.py`. Rebased onto current `origin/master` and reran focused verification locally: `./scripts/test.sh tests/test_session_lineage_collapse.py tests/test_sidebar_session_partition.py tests/test_465_session_branching.py tests/test_issue4385_cron_archive_reappears.py tests/test_session_import_cli_fallback_model.py tests/test_issue3988_show_cli_sessions_default.py -q` → 98 passed; `node --check static/sessions.js`, `git diff --check`, and `python3 scripts/ruff_lint.py --diff origin/master` also passed.

Issue #4029, opened 2026-06-12
Stale bg-task notifications overtake new turns on long-running sessions


Pandasywithfantasy:
---
labels: bug, session, task-contamination
---

### Title
**Bug: stale background-process completion notifications prepended to new user message on long-running sessions, causing agent to respond to wrong task + WebUI display corruption**

---

### Bug Description

When a user returns to a long-running session and sends a new message, stale
background-process completion notifications from tasks started *hours or days
earlier* in the same session are prepended to the new message. The agent then
responds to the old task rather than the new one. The new user message is
subsequently **missing from the WebUI display** (though preserved in `state.db`),
and the WebUI JSON sidecar shows a merged turn combining old and new requests.

---

### Steps to Reproduce

1. Start a session and issue a task that spawns long-running background processes
   with `notify_on_complete=true` (e.g., media download + encode, large file transfers).
2. Let the session go idle for hours or days while those processes are still running
   or are in the completion queue.
3. **Without creating a new session**, issue a completely unrelated task.
4. **Observe**: the agent responds to the old background-process task.
   The new request is absent from the transcript display. The JSON sidecar
   `messages` array shows a merged user turn combining old + new content.

Also reproduces on **messaging-platform sessions** (WeChat, Telegram, etc.): when an
external-surface session is loaded into the WebUI, stale process completions from the
original platform surface are drained into the WebUI session on the user's next turn.

---

### Root Cause — Code Analysis

**Primary site** — `api/streaming.py` (v0.51.326), lines 6345–6348:

```python
_process_notifications = _drain_webui_process_notifications(session_id)
_agent_msg_text = msg_text
if _process_notifications:
    _agent_msg_text = "\n\n".join([*_process_notifications, msg_text]).strip()
# L6349: sent to agent.run_conversation()
# L6356: persist_user_message=msg_text (original only, notifications stripped)
```

`_drain_webui_process_notifications()` (L1393–1447) drains the **entire**
`process_registry.completion_queue` with **no timestamp or age filtering**:

```python
while True:
    evt = completion_queue.get_nowait()    # ALL pending, regardless of age
    ...
    if getattr(proc, 'session_key', None) != session_id:
        skipped_events.append(evt)         # filters by session_id only
        continue
    notification = _format_process_notification(evt)
    notifications.append(notification)
```

**Consequence chain:**

1. Line 6348 prepends all notifications to the user's actual message text, regardless
   of whether the processes are 5 minutes or 5 days old.
[...]

PR #4417, opened 2026-06-18
fix: ignore malformed process wakeup events

ai-ag2026:
## Summary
- ignore empty/malformed process wakeup events instead of injecting `Background process unknown completed` prompts
- render watch-pattern overflow/release summaries as explicit watch notifications
- keep normal completion, watch-match, and watch-disabled wakeups unchanged

## Root cause
`api.background_process.format_wakeup_prompt({})` defaulted missing payload fields to `unknown` and formatted the event as a normal completion. When the completion queue delivered an empty or malformed payload, WebUI could start a server-side wakeup turn containing only:

```text
[IMPORTANT: Background process unknown completed (exit_code=?).
Command: unknown
Output:
]
```

Global watch-overflow summary events were another malformed-looking shape for this formatter: they intentionally have no process id or command, but the formatter still treated them as normal process completions.

## Related reconnaissance
- Searched existing `nesquena/hermes-webui` issues/PRs for `Background process unknown`, `unknown completed`, `watch_overflow`, and `bg_task_complete`.
- Related but distinct: #4029 tracks stale completion age-gating. This PR only guards malformed/empty wakeup payload formatting.

## Test plan
- `python3 -m py_compile api/background_process.py`
- `./scripts/test.sh tests/test_background_process_wakeup_format.py tests/test_session_channel_option_x.py tests/test_wakeup_defer_race.py tests/test_bg_task_complete_ab_coexistence.py -q`
- `.venv/bin/ruff check api/background_process.py tests/test_background_process_wakeup_format.py`
- Static added-line scan: 0 findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Malformed background-process wakeup events no longer appear as empty `unknown completed` prompts.** WebUI now ignores empty or unknown completion-queue payloads and renders watch-pattern overflow summaries as explicit watch notifications instead of fake process completions.
+
 ## [v0.51.489] — 2026-06-18 — Release QY (outline button no longer collides with the scroll control)
 
 ### Fixed
--- a/api/background_process.py
+++ b/api/background_process.py
@@ -350,17 +350,24 @@ def _truncate(text: str, limit: int) -> str:
     return s[:limit] + "\n…(truncated)"
 
 
-def format_wakeup_prompt(evt: dict) -> str:
+def format_wakeup_prompt(evt: dict) -> str | None:
     """Build the synthetic [IMPORTANT: …] message the agent will see.
 
     Mirrors ``cli._format_process_notification`` so wakeup payloads look the
     same in CLI and WebUI sessions.
     """
+    if not isinstance(evt, dict) or not evt:
+        return None
+
     evt_type = evt.get("type", "completion")
-    sid = evt.get("session_id", "unknown")
-    cmd = evt.get("command", "unknown")
+    sid = str(evt.get("session_id") or "").strip()
+    cmd = str(evt.get("command") or "").strip()
+    if evt_type in {"watch_overflow_tripped", "watch_overflow_released"}:
+        msg = str(evt.get("message") or "").strip()
+        return f"[IMPORTANT: {msg}]" if msg else None
     if evt_type == "watch_disabled":
-        return f"[IMPORTANT: {evt.get('message', '')}]"
+        msg = str(evt.get("message") or "").strip()
+        return f"[IMPORTANT: {msg}]" if msg else None
     if evt_type == "watch_match":
         pat = evt.get("pattern", "?")
         out = _truncate(evt.get("output", ""), 4000)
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Reading the diff at this PR's HEAD against `origin/master`, plus the full `_process_one` context in `api/background_process.py:752-905` and the agent-side emitters in `tools/process_registry.py`, the core fix is solid but one of the two branches it adds is effectively unreachable on the WebUI server-side wakeup path. Worth confirming before merge.

## What's correct

The primary guard is exactly right. On `origin/master`, `format_wakeup_prompt` defaulted missing fields to `unknown`/`?`:

```python
sid = evt.get("session_id", "unknown")
cmd = evt.get("command", "unknown")
...
f"[IMPORTANT: Background process {sid} completed (exit_code={exit_code})..."
```

So a completion event that resolved to a real WebUI session but carried no `command`/`exit_code`/`output` would inject the bogus `Background process unknown completed (exit_code=?)` turn. The new early-returns plus the `str | None` signature close that, and the call-site update at `api/background_process.py:896-897` handles the `None` cleanly:

```python
wakeup_prompt_raw = format_wakeup_prompt(evt)
wakeup_prompt = wakeup_prompt_raw.strip() if wakeup_prompt_raw else ""
```

The second call site (`_build_payload`, `api/background_process.py:433`) already wraps it in `if wakeup_body:` inside a try/except, so it's safe with the new `None` return too. Good — no missed caller.

## The overflow branch is dead code on this path

The new `watch_overflow_tripped` / `watch_overflow_released` branch (`api/background_process.py:365-367`) can't actually be hit through `_process_one`. Those events are emitted by the agent with **empty `session_id` and empty `session_key`** (see `tools/process_registry.py:343-346` and `:394-396`, both set `"session_id": ""`, `"session_key": ""`). But `_process_one` drops anything without a recoverable `session_key` long before the formatter runs:

```python
# api/background_process.py:792
if not session_key:
    logger.debug("process_complete drop: no recoverable session_key for process_id=%r", process_id)
    return
```

Both `format_wakeup_prompt` invocations (line 856 via `_build_payload`, line 883 direct) sit *after* that guard, so an overflow summary never reaches the new branch. The PR description frames these as "another malformed-looking shape for this formatter" that was being mis-rendered as a completion — but in the WebUI server-side wakeup path they were already being dropped at the `session_key` gate, not mis-formatted. The branch is harmless future-proofing, not a fix for an observable symptom here.

## Cross-repo note

The docstring says this formatter "Mirrors `cli._format_process_notification`". The agent's canonical formatter is `format_process_notification` in `tools/process_registry.py:1619`, and it does **not** handle `watch_overflow_*` either — they'd fall through to its completion branch. So adding the overflow branch here diverges from the mirror rather than matching it. If the intent is to surface overflow summaries to WebUI users, that belongs upstream (give the events a routable `session_key`, or special-case them before the `session_key` guard) — a formatter-only change can't deliver them.

## Recommendation

The empty/non-dict completion guard is the real fix and is worth merging on its own. For the overflow branch, either (a) drop it as unreachable to avoid implying behavior that doesn't occur, or (b) keep it but add a one-line code comment that these events are gated out upstream and the branch is defensive only. Greptile also flagged that `watch_disabled` (empty-message) and non-dict inputs aren't tested — both of those *are* reachable (`watch_disabled` carries a real `session_key`, per `tools/process_registry.py:276`), so adding those two cases to `tests/test_background_process_wakeup_format.py` would be more valuable than the overflow cases that can't fire here.

Issue #1896, opened 2026-05-08
bug(streaming): context-length fallback ignores config override + provider, defaults to 256K — breaks LCM auto-compress on 1M models


nesquena-hermes:
## Summary

The two `get_model_context_length()` fallback callsites in `api/streaming.py` (one for session persistence, one for the SSE usage payload) call the resolver with **only `model` + `base_url`**, omitting `config_context_length`, `provider`, and `custom_providers`. As a result, when the agent's `context_compressor` reports 0 (fresh / cached / transitioning agent), context-length resolution falls all the way through to `DEFAULT_FALLBACK_CONTEXT = 256_000` even when the user has set `model.context_length: 1048576` in `config.yaml` or has a 1M model with a `custom_providers` per-model override.

For users with a context-management plugin (LCM) configured around the real window, this cascades into a session-killing failure mode: auto-compression triggers far too early → flood of compress requests → 429s → credential pool exhaustion → fallback also 429s → `API call failed after 3 retries`.

Reported in #webui Discord by @AvidFuturist with deepseek-v4-flash (1M context window). Reproduced 5 times by the user.

## Observed

WebUI chat toolbar shows **"108.1k / 262.1k tokens used"** — should be 1M.
Auto-compress fires at ~65K (25% of 262K, per LCM threshold).

```
LCM compaction #1: 195 messages → 64 (1 leaf pass, 220581→82887 tokens)
credential pool: marking OPENCODE_GO_API_KEY exhausted (status=429), rotating
credential pool: no available entries (all exhausted or empty)
Fallback activated: deepseek-v4-flash → google/gemma-4-26b-a4b-it:free (openrouter)
API call failed after 3 retries. HTTP 429
```

Adding `model.context_length: 1048576` to `config.yaml` does **not** fix it — the buggy fallback path doesn't read the override.

## Root cause (verified in code)

**Resolution chain** (`agent/model_metadata.py` `get_model_context_length()`):

```
0. config_context_length override          ← SKIPPED (not passed by fallback)
1. Persistent cache
2-5. Endpoint probes (LM Studio, Anthropic API, custom endpoints)
6. models.dev registry                     ← needs `provider` (not passed)
7. OpenRouter metadata
8. Hardcoded DEFAULT_CONTEXT_LENGTHS       ← may not catch fuzzy match
9. DEFAULT_FALLBACK_CONTEXT = 256_000      ← what user sees as "262.1k"
```

[...]

PR #4418, opened 2026-06-18
fix: preserve large context window on session reload

### LINKED ISSUE #4248: bug(context): context-window reverts to 256k default on session reload for 1M models, trips early auto-compress (residual of #3256/#3263)
## Summary

Reported in Discord (#webui, **@Countlesshrs**, 2026-06-15) — unconfirmed user report, filed after investigation. The context-window indicator reverts to the configured 256k default on a session **reload**, even when the active model is a 1M-context model, which then trips auto-compression far too early.

> "Now the token count for some reason defaults to configured setting which is 256k and even when model is 1M, sending a message fixes it but when returning to a session where a reload is needed it sets back to 256k, this in turn activates compression even on 100k ctx when the threshold should be 500k"

This is a **residual** of the previously-shipped #3256 / #3263 fix (which scoped the global `model.context_length` cap to `model.default` only, and rescaled `threshold_tokens` on the streaming + SSE paths). The streaming path is fixed; the **session GET / reload path** is not fully aligned.

## Root cause (verified against master, v0.51.434)

On session load, `api/routes.py:6855`:
```python
if (not _persisted_cl) or resolve_model:
    _fb_cl = _resolve_context_length_for_session_model(_model_for_lookup, ...)
    if _fb_cl:
        if _persisted_cl and _fb_cl != _persisted_cl:
            _threshold_tokens = 0
        _persisted_cl = _fb_cl
```
- `resolve_model` defaults to `"1"` whenever `load_messages` is true (`routes.py:6716-6717`) — i.e. on every full reload. So the persisted (correct, 1M) `context_length` is **unconditionally overwritten** by a fresh `_resolve_context_length_for_session_model()` call.
- That resolver (`routes.py:3687`) calls `agent.model_metadata.get_model_context_length()` with `api_key=_ctx_lookup.api_key`, but `_context_length_lookup_inputs_for_model()` only populates `api_key` for **custom_providers** (`routes.py:3177`); on the normal provider/global path `api_key=""`.
- Without an api_key (and with models.dev / live-probe unavailable on the read-only GET path), `get_model_context_length` falls through to `DEFAULT_FALLBACK_CONTEXT = 256_000` (`agent/model_metadata.py:180`, `CONTEXT_PROBE_TIERS[0]`) unless the bare model name happens to match the static `DEFAULT_CONTEXT_LENGTHS` catalog (e.g. `claude-opus-4-8` → 1M survives; many 1M models that rely on a live lookup do **not**).
- Net: streaming persists 1M, reload recomputes and can land on **256k**, diverging from the value streaming showed. The matching `threshold_tokens` is then either zeroed or computed against 256k, so auto-compress fires at ~25% of 256k (~64k) instead of ~25% of the real window — exactly the "compression activates even on 100k ctx when threshold should be 500k" symptom.

## Why the existing fix doesn't cover this
#3256/#3263's threshold-rescale lives in `api/streaming.py` (the stream-end save + SSE-done payload). The session-**load** path (`routes.py:6855`) recomputes independently and:
1. Overwrites a correctly-persisted larger window with a possibly-256k recompute, and
2. Resolves **without** the api_key/live-probe context that the streaming path had, so the recompute is lower-fidelity than the value it's replacing.

## Suggested fix shape
- **Don't clobber a good persisted value with a lower-confidence recompute.** When `_persisted_cl` is already positive and came from a real provider call, prefer it; only recompute when it's missing/zero, or when the *model actually changed* (not on every reload). Consider gating the `resolve_model` recompute to "model differs from the persisted model" rather than "any reload".
- **Or** make the GET resolver pass the same api_key/base_url the streaming path uses (resolve via the session's provider credentials) so the recompute matches.
- Either way, rescale `threshold_tokens` to the resolved window on the load path the same way streaming/SSE already do (mirror the `_skip_cc_cl` rescale at `streaming.py:7563`), so the reloaded auto-compress trigger matches the live snapshot.

## Scope
WebUI-side (`api/routes.py` session-get handler + `_resolve_context_length_for_session_model`). ~15-30 LOC + a regression test extending `tests/test_issue3256_context_length_default_only_guard.py` to cover the GET/reload path (currently only the streaming/save and matcher paths are covered).

## Related
- #3256, #3263 (closed) — the streaming/save-side fix this extends.
- #1896, #3717 (closed) — the original 256k-fallback + provider-override-ignored bugs.

## Reported by
@Countlesshrs (Discord #webui, 2026-06-15).

franksong2702:
## Thinking Path

- Hermes WebUI persists context-window metadata from streaming saves so reloads should not change the apparent model window.
- #4248 showed that a session using a 1M-context model could reload back to the generic 256k fallback.
- The root cause is that the session reload resolver did not carry the same `base_url` / `api_key` lookup shape used by the streaming path, so metadata cache lookup could miss.
- A reload recompute is lower confidence than an already-persisted larger window when it falls back to the generic 256k default.
- This PR aligns the reload lookup inputs and guards against clobbering a larger persisted window with that fallback, while still allowing a real model switch to a genuine 256k model.

## What Changed

- Added config-based API-key resolution for context-window metadata lookup.
- Let `_resolve_context_length_for_session_model()` accept explicit `base_url` and `api_key` inputs.
- Updated GET `/api/session` hydration to resolve session context metadata with model/provider/base-url/api-key state instead of an anonymous model-only lookup.
- Added a guard so a resolved 256k fallback cannot overwrite a larger persisted context window on same-model reload.
- Refined that guard so an effective model change to a genuine 256k model is accepted instead of preserving a stale 1M window.
- Split `resolve_model_provider` and `resolve_custom_provider_connection` imports so non-custom base URL resolution does not depend on the custom-provider helper being available.
- Preserved threshold ratio when a higher-confidence context-window refresh changes the window size.
- Added regression coverage for the reload lookup inputs, 1M-preservation case, genuine 256k model-switch case, and missing custom-provider helper case.
- Updated the existing #1896 structure guard so it tracks the new refresh gate rather than the old `if _fb_cl:` shape.
- Added a changelog entry.

## Why It Matters

Reloading a session should not make the context indicator and auto-compression threshold regress from a previously-captured 1M model window to the generic 256k fallback. That false downgrade can make the UI report the wrong context usage and can trip compression too early.

Fixes #4248.

## Verification

- `./scripts/test.sh tests/test_issue3256_context_length_default_only_guard.py tests/test_issue1436_context_indicator_load_path.py -q`
  - `29 passed`
- `./scripts/test.sh tests/test_issue1896_context_length_fallback_args.py::test_routes_session_load_fallback_passes_config_overrides tests/test_issue3256_context_length_default_only_guard.py tests/test_issue1436_context_indicator_load_path.py -q`
  - `30 passed`
- `./scripts/test.sh tests/test_issue3256_context_length_default_only_guard.py tests/test_issue1896_context_length_fallback_args.py::test_routes_session_load_fallback_passes_config_overrides tests/test_issue1436_context_indicator_load_path.py -q`
  - `32 passed`
- `git diff --check`
  - clean

## Risks / Follow-ups

- This keeps the fix scoped to session hydration/context metadata. It does not change streaming save semantics or mutate historical messages.
- The 256k rejection is intentionally narrow: it only blocks the generic fallback from replacing an already larger persisted window for the same effective model identity.
- I did not run the full test suite locally; verification was limited to the context-window regression tests, the CI-failed structure guard, the two Greptile follow-up regressions, and diff-scoped lint.

## Model Used

- Provider: OpenAI
- Model: GPT-5 Codex
- Tooling/mode: Codex coding agent with local shell, git, and GitHub CLI. AI assisted with implementation, tests, and PR preparation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Reloading a session with a large model context window no longer snaps it back to the 256k fallback (#4248).** The session reload path now resolves context-window metadata with the same base URL / API-key lookup inputs used by streaming saves, and it refuses to overwrite a larger persisted context window with the generic 256k fallback.
+
 ## [v0.51.489] — 2026-06-18 — Release QY (outline button no longer collides with the scroll control)
 
 ### Fixed
--- a/api/routes.py
+++ b/api/routes.py
@@ -3413,11 +3413,60 @@ def _custom_provider_api_key_for_context(entry: dict, provider: str) -> str:
         return ""
 
 
+def _context_length_config_api_key_for_provider(
+    provider: str | None,
+    cfg: dict | None,
+) -> str:
+    """Return a config/env API key usable for context-window metadata lookup."""
+    cfg = cfg if isinstance(cfg, dict) else {}
+    provider = _canonical_context_provider(provider)
+
+    def _resolve_key(raw_api_key, raw_key_env=None) -> str:
+        api_key_text = str(raw_api_key or "").strip()
+        if (
+            api_key_text.startswith("${")
+            and api_key_text.endswith("}")
+            and len(api_key_text) > 3
+        ):
+            resolved = os.getenv(api_key_text[2:-1], "").strip()
+            if resolved:
+                return resolved
+        elif api_key_text:
+            return api_key_text
+        key_env = str(raw_key_env or "").strip()
+        if key_env:
+            resolved = os.getenv(key_env, "").strip()
+            if resolved:
+                return resolved
+        return ""
+
+    providers_cfg = cfg.get("providers", {})
+    if isinstance(providers_cfg, dict):
+        for provider_key, provider_cfg in providers_cfg.items():
+            if not isinstance(provider_cfg, dict):
+                continue
+            if not _providers_match_for_context(provider_key, provider):
+                continue
+            api_key = _resolve_key(provider_cfg.get("api_key"), provider_cfg.get("key_env"))
+            if api_key:
+                return api_key
+
+    model_cfg = cfg.get("model", {})
+    if isinstance(model_cfg, dict):
+        model_provider = _canonical_context_provider(model_cfg.get("provider"))
+        if not provider or _providers_match_for_context(model_provider, provider):
+            api_key = _resolve_key(model_cfg.get("api_key"), model_cfg.get("key_env"))
+            if api_key:
+                return api_key
+    return ""
+
+
 def _context_length_lookup_inputs_for_model(
     model: str | None,
     provider: str | None = None,
     *,
     base_url: str | None = None,
+    api_key: str | None = None,
     cfg: dict | None = None,
 ) -> _ContextLengthLookupInputs:
     """Return the effective metadata resolver inputs for a WebUI model.
@@ -3472,7 +3521,7 @@ def _context_length_lookup_inputs_for_model(
             break
 
     custom_context_length = None
-    effective_api_key = ""
+    effective_api_key = str(api_key or "").strip()
     if custom_providers:
         target_base = effective_base_url.rstrip("/")
         model_candidates = set(_model_lookup_candidates(bare_model or model_for_lookup))
@@ -3502,7 +3551,8 @@ def _context_length_lookup_inputs_for_model(
                 effective_provider = entry_slug
             if not effective_base_url and entry_base:
                 effective_base_url = entry_base
-            effective_api_key = _custom_provider_api_key_for_context(entry, effective_provider or entry_slug)
+            if not effective_api_key:
+                effective_api_key = _custom_provider_api_key_for_context(entry, effective_provider or entry_slug)
             custom_context_length = _models_config_context_length(models_cfg, bare_model or model_for_lookup)
             break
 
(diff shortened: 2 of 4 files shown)


franksong2702:
Addressed the two Greptile P2 comments in `531c5787`:

- refined the 256k fallback guard so same-model reloads still reject the anonymous 256k fallback, but an effective model change to a genuine 256k model is accepted;
- split `resolve_model_provider` and `resolve_custom_provider_connection` imports so non-custom base URL resolution does not depend on the custom-provider helper being available.

Verification:

- `./scripts/test.sh tests/test_issue3256_context_length_default_only_guard.py tests/test_issue1896_context_length_fallback_args.py::test_routes_session_load_fallback_passes_config_overrides tests/test_issue1436_context_indicator_load_path.py -q` -> `32 passed`
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master` -> no new violations on added/modified lines
- GitHub CI on `531c5787`: lint + all Python shards passed

Note: the browser-smoke workflow for this commit was cancelled while installing Playwright/Chromium system deps via apt (Ubuntu mirror/dependency install stalled until GitHub cancelled it). I do not have repository admin rights to rerun that workflow from here.

nesquena-hermes:
Shipped in **v0.51.492** (release PR #4425). Thanks @franksong2702!

Your fix preserves the large persisted context window on session reload (#4248) — the reload resolver now aligns base_url/api_key with the streaming save path and refuses to let the anonymous 256k fallback clobber a larger persisted window.

One addition on the way in (release gate): the Codex regression gate found that the model-identity check normalized only the `@provider:model` form, not `provider/model` (slash, OpenRouter-style) — so a slash-stored model resolving to its bare id was wrongly treated as a model change and bypassed the 256k accept-guard for that shape. I added slash-form normalization (mirroring `_model_matches_configured_default`) plus a route-level regression test and a unit test for the slash shape. Your authorship is preserved via a `Co-authored-by` trailer.

PR #4420, opened 2026-06-18
fix(#4319): allow message:// links in markdown renderer

### LINKED ISSUE #4319: bug(chat): Clickable message:// links are [blocked]
Whenever the LLM tries to create a clickable link in the chat, I get 

> Open in Mail [blocked]

I have confirmed that the knowledge is available to build the link (it knows the mail message ID)

Hermes is able to show the link in a code block and is able to show it as plain text, but not as a clickable link.

My RAG is full of facts I derived from emails. I need those message links to easily click on facts and double check the research. (Message links are MacOS specific).

This issue has been verified in Hermes Web Chat and Hermes Desktop v0.16.0

This might be related to https://github.com/nesquena/hermes-webui/issues/2768 where issues with mailto links were handled?

bergeouss:
## Summary

macOS Mail message links (`message://msgid`) were silently stripped by the URL sanitizer in the chat renderer. The `_SMD_SAFE_URL_RE` whitelist only allowed `https?`, `mailto:`, `tel:`, and relative paths — `message://` was not included, so the `href` attribute was removed and replaced with `[data-blocked-scheme]`.

Fixes #4319

## Changes

```js
// static/messages.js line 2222
- const _SMD_SAFE_URL_RE=/^(?:https?:|mailto:|tel:|\/|#|\?|\.|api|session\/)/i;
+ const _SMD_SAFE_URL_RE=/^(?:https?:|mailto:|tel:|message:|\/|#|\?|\.|api|session\/)/i;
```

## CSP Assessment

- Server uses `Content-Security-Policy-Report-Only` (not enforced, logging only)
- No `navigate-to` directive exists — `message://` link navigations are not blocked by CSP
- No CSP changes needed

## Verification

Tested with Node.js:
- `message://abc` → ✅ passes
- `javascript:evil` → ❌ blocked
- `data:evil` → ❌ blocked
- `vbscript:evil` → ❌ blocked
- All existing schemes (https, mailto, tel, relative, api/, session/) still pass

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -2219,7 +2219,7 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
   }
   // Allowed URL schemes for anchors and images rendered from agent-streamed markdown.
   // Raw file:// anchors are rewritten to /api/media before the user can click them.
-  const _SMD_SAFE_URL_RE=/^(?:https?:|mailto:|tel:|\/|#|\?|\.|api|session\/)/i;
+  const _SMD_SAFE_URL_RE=/^(?:https?:|mailto:|tel:|message:|\/|#|\?|\.|api|session\/)/i;
   const _SMD_SAFE_IMG_URL_RE=/^(?:https?:|mailto:|tel:|\/|#|\?|\.)/i;
   function _smdLinkHref(raw){
     const href=String(raw||'');
--- a/static/ui.js
+++ b/static/ui.js
@@ -4637,7 +4637,7 @@ function renderMd(raw){
     t=t.replace(/\x00C(\d+)\x00/g,(_,i)=>_code_stash[+i]);
     // Stash [label](url) links before autolink so the URL in href= is not re-linked
     const _link_stash=[];
-    t=t.replace(/\[([^\]]+)\]\(((?:https?:\/\/|file:\/\/|workspace:\/\/|session:\/\/|mailto:|tel:)[^\s\)]+)\)/g,(_,lb,u)=>{_link_stash.push(_markdownAnchor(lb,u));return `\x00L${_link_stash.length-1}\x00`;});
+    t=t.replace(/\[([^\]]+)\]\(((?:https?:\/\/|file:\/\/|workspace:\/\/|session:\/\/|mailto:|tel:|message:)[^\s\)]+)\)/g,(_,lb,u)=>{_link_stash.push(_markdownAnchor(lb,u));return `\x00L${_link_stash.length-1}\x00`;});
     t=t.replace(/(https?:\/\/[^\s<>"')\]]+)/g,(url)=>{const trail=url.match(/[.,;:!?)]$/)?url.slice(-1):'';const clean=trail?url.slice(0,-1):url;return `<a href="${clean}" target="_blank" rel="noopener">${esc(clean)}</a>${trail}`;});
     t=t.replace(/\x00L(\d+)\x00/g,(_,i)=>_link_stash[+i]);
     t=t.replace(/\x00G(\d+)\x00/g,(_,i)=>_img_stash[+i]);
@@ -4778,7 +4778,7 @@ function renderMd(raw){
   // Stash existing <a> tags first to avoid re-linking already-linked URLs.
   const _a_stash=[];
   s=s.replace(/(<a\b[^>]*>[\s\S]*?<\/a>)/g,m=>{_a_stash.push(m);return `\x00A${_a_stash.length-1}\x00`;});
-  s=s.replace(/\[([^\]]+)\]\(((?:https?:\/\/|file:\/\/|workspace:\/\/|session:\/\/|mailto:|tel:)[^\s\)]+)\)/g,(_,label,url)=>_markdownAnchor(label,url));
+  s=s.replace(/\[([^\]]+)\]\(((?:https?:\/\/|file:\/\/|workspace:\/\/|session:\/\/|mailto:|tel:|message:)[^\s\)]+)\)/g,(_,label,url)=>_markdownAnchor(label,url));
   s=s.replace(/\x00A(\d+)\x00/g,(_,i)=>_a_stash[+i]);
   // Restore raw <pre> only after markdown rewrites so literal preformatted
   // content stays placeholder-protected, then let the sanitizer normalize tags.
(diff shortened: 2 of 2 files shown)


nesquena-hermes (changes requested):
Thanks for tackling #4319 — the core approach is right, and the live-stream half works. But the fix is **incomplete in a way that misses the reported use case**, so it needs one more change before it can land.

### What's correct ✅

Adding `message:` to `_SMD_SAFE_URL_RE` (`static/messages.js:2222`) is the right call. `message://` is an OS-routed scheme (macOS Mail), exactly like the already-allowlisted `mailto:`/`tel:` — the app doesn't need to "handle" it, the OS does. `_smdLinkHref` passes it through unchanged (no rewrite), so during a **live agent stream** the link now renders as a clickable `<a href="message://...">`. The regex stays `^`-anchored, so no XSS surface is opened (`javascript:`, `data:`, leading-control-char prefixes all still rejected). 👍

### The blocker — the fix doesn't survive a reload ❌

There are **two** markdown renderers, and the PR only patched one:

- `static/messages.js` `_SMD_SAFE_URL_RE` (line 2222) — the **live agent-stream** path. ✅ patched.
- `static/ui.js` `renderMd()` — the **persisted / reload / history re-render** path (`renderMessages()` → `renderMd()`, ui.js:867 → 4399). ❌ **not** patched. Its scheme allowlist does not include `message:`:
  - link-extraction regexes at `ui.js:4640` and `ui.js:4781` allow only `https?://|file://|workspace://|session://|mailto:|tel:`
  - `_isSafeUrl()` at `ui.js:4861` returns true for `https?`, `mailto:`, `tel:` — but not `message:` → the anchor is stripped to a bare `<a>`.

Net effect: a `message://` link is clickable *while the turn is streaming*, but the moment the page is reloaded (or the message is re-rendered from session history) it goes through `ui.js renderMd()` and is blocked again. Issue #4319 is specifically about clicking persisted email facts after the fact ("My RAG is full of facts I derived from emails… click on facts and double check"), which is the reload path — so as-is the PR doesn't actually fix the reported bug, and you get an inconsistent UX (link works live, dies on refresh).

The established convention confirms this: `mailto:` and `tel:` appear in **both** renderers. A new OS-scheme has to be mirrored the same way.

### Fix-spec (mirror the same addition into `ui.js renderMd`)

1. `static/ui.js:4640` — add `message:|` to the link regex alternation (same position as `mailto:|tel:`).
2. `static/ui.js:4781` — same addition to the second link regex.
3. `static/ui.js:4867` — extend the allow check to `if(/^(mailto:|tel:|message:)/i.test(raw)) return true;`

(Confirm `_markdownHref` at ui.js:4797 leaves `message://` untouched — it has no `message:` branch, so it passes through unchanged, which is correct.)

### Tests

No test was added. Please add a small assertion (mirroring the existing mailto/tel render coverage) that a `message://…` markdown link renders as a clickable `<a href="message://…">` through **both** `renderMd()` (ui.js) and the smd live-stream path — that locks the two-renderer symmetry so this doesn't regress.

Once `ui.js renderMd` allows `message:` too, this is good to go.

nesquena-hermes:
Pulled this into a read-only worktree and read the diff against `origin/master`. The one-line change to the streaming allowlist is correct as far as it goes:

```js
// static/messages.js:2222
const _SMD_SAFE_URL_RE=/^(?:https?:|mailto:|tel:|message:|\/|#|\?|\.|api|session\/)/i;
```

That re-enables `message://` anchors during live streaming, since `_sanitizeSmdLinks()` (messages.js:2258-2262) is what strips the `href` and sets `data-blocked-scheme` on schemes that don't match. So the *streaming* path is fixed.

## The gap: the static/reload render path has a separate allowlist

`#4319` reports the link is `[blocked]`, and the reporter confirmed it reproduces in Hermes Desktop too. The live SSE renderer is only one of two markdown paths. When a session is reloaded from disk (or rendered cold), assistant text goes through `renderMd()` in `static/ui.js`, **not** `attachLiveStream`. `renderMd` calls `renderMd` → see messages.js:2503/2555/2892 (`assistantBody.innerHTML=renderMd(...)`). That path has its own scheme list which this PR does not touch:

```js
// static/ui.js:4640 and :4781  (the [label](url) link passes)
t=t.replace(/\[([^\]]+)\]\(((?:https?:\/\/|file:\/\/|workspace:\/\/|session:\/\/|mailto:|tel:)[^\s\)]+)\)/g, ...)
```

```js
// static/ui.js:4861-4869  (_isSafeUrl, the sanitizer gate)
function _isSafeUrl(v, img){
  ...
  if(/^https?:\/\//i.test(raw)) return true;
  if(/^(mailto:|tel:)/i.test(raw)) return true;
  ...
  return false;   // message:// falls through here → link dropped
}
```

So after the fix lands, a `message://` link will render clickable while streaming, then **revert to `[blocked]` on the next reload / hard refresh** because `renderMd`'s link regex never matches `message:` (the `[label](url)` pass at ui.js:4640/4781 won't even produce an anchor), and `_isSafeUrl` returns `false` for it.

## Recommendation

To actually close `#4319`, mirror the scheme in the `renderMd` path too:

1. Add `message:` to both `[label](url)` regexes at `ui.js:4640` and `ui.js:4781`.
2. Add a `message:` branch to `_isSafeUrl` at `ui.js:4867`, alongside the existing `mailto:|tel:` line:
   ```js
   if(/^(mailto:|tel:|message:)/i.test(raw)) return true;
   ```

Note `message://` needs no href rewriting (unlike `file://` → `/api/media`); the OS mail client handles the scheme, same as `mailto:`. So `_markdownHref` (ui.js:4793-) and `_smdLinkHref` (messages.js:2224-) can pass it through unchanged — only the allowlists need the new scheme.

## Test plan

`tests/` already has coverage for the streaming allowlist; please add an assertion that the `renderMd` link pass + `_isSafeUrl` accept `message:` so the reload path stays covered. A quick repro: send a `[Open in Mail](message://%3Cid%40host%3E)` link, confirm it's clickable live, then reload the session and confirm it stays clickable (not `[blocked]`).

bergeouss:
**Review fixes pushed** ✅

Added `message:` to all three locations in ui.js as requested:

1. Link-extraction regex at `_partitionSidebarSessionRows` (ui.js:4640)
2. Link-extraction regex in `renderMd` (ui.js:4781)
3. `_isSafeUrl` allowlist (ui.js:4867)

Now `message://` links stay clickable through page reload — not just during live streaming.

nesquena-hermes:
Shipped in **v0.51.505** (deployed). Thanks @bergeouss — `message:` is now allowed on all the markdown link paths; I added an in-pattern regression test (asserts `message:` renders clickable + `javascript:` stays blocked) per the review. Co-authored-by credit is in the release commit + CHANGELOG.

PR #4421, opened 2026-06-18
fix(#4416): force browser notification on stream completion

### LINKED ISSUE #4416: Browser notifications silently dropped when user is on another tab (SSE throttling race)
### Problem

Desktop notifications via the Notification API don't arrive when the agent finishes a response while the user is on another tab — despite HTTPS, granted browser permission, and `notifications_enabled: true` in WebUI settings.

### Root cause

1. User sends a message → switches to another tab (`document.hidden = true`)
2. Agent finishes processing → SSE `done` event is emitted server-side
3. Chromium-based browsers throttle SSE/EventSource for background tabs → `done` event is **queued** and only delivered when the user returns to the tab
4. The `done` handler calls `sendBrowserNotification()` which checks `!document.hidden` → but it's `false` (user just came back) → notification silently dropped

### Evidence

The test button in Settings (`Settings → Browser notifications → Send test`) calls `sendBrowserNotification()` with `{force: true}` and works correctly, proving the Notification API itself is functional.

### Location

`static/messages.js` line ~3512 — the `done` event handler's `sendBrowserNotification` call:

```js
sendBrowserNotification('Response complete',assistantText?assistantText.slice(0,100):'Task finished',{sid:activeSid});
```

The guard at line ~5325:
```js
if(!force&&(!window._notificationsEnabled||!document.hidden)) return;
```

### Attempted fix (local, reverted)

Capturing `document.hidden` at stream start + tracking `visibilitychange` during stream, then passing `force: _wasEverHidden`. It fixed the SSE delay case but produced false positives from unrelated `visibilitychange` events (Alt+Tab, window coverage, screen lock).

### Suggested directions

1. Instead of checking `document.hidden` at `done` receive time, track whether the tab was *ever* hidden during the stream (the `_wasEverHidden` approach) — needs a reliable way to avoid false positives from non-tab visibility changes
2. Or use a different mechanism for notification delivery that isn't subject to background tab SSE throttling (e.g. Web Workers, Service Worker push)
3. Or accept `force: true` for response-complete notifications (simple, slightly more aggressive behavior)

bergeouss:
## Summary

Browser notifications via the Notification API silently fail when the agent finishes while the user is on another tab. Chromium throttles SSE/EventSource for background tabs, queuing the `done` event until the user returns. By then `document.hidden` is false, so the notification guard silently drops it.

Fixes #4416

## Root cause

1. User sends message → switches to another tab (`document.hidden = true`)
2. Agent finishes → SSE `done` event is queued (background tab throttling)
3. User returns to tab → `done` event is delivered, but `document.hidden` is now false
4. Guard at `sendBrowserNotification`: `if(!force && !document.hidden) return;` → notification silently dropped

## Fix

One-line change: pass `{force:true, sid:activeSid}` in the `done` handler's `sendBrowserNotification()` call (line 3512).

This is safe because the `done` event fires **exactly once per stream** — no spam risk. The `force` flag bypasses the `document.hidden` check so the notification always fires regardless of when Chromium delivers the queued event.

## Verification

- Only the `done` handler's call was modified — approval/clarification notification calls (lines 3160, 3168) remain unchanged
- Settings test button keeps its own `force:true`

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -3509,7 +3509,7 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
         renderSessionList();
         _setActivePaneIdleIfOwner();
         playNotificationSound();
-        sendBrowserNotification('Response complete',assistantText?assistantText.slice(0,100):'Task finished',{sid:activeSid});
+        sendBrowserNotification('Response complete',assistantText?assistantText.slice(0,100):'Task finished',{forceHidden:true,sid:activeSid});
       };
       if(_shouldUseStreamFade()&&assistantBody){
         _cancelAnimationFramePendingStreamRender();
@@ -5324,7 +5324,9 @@ function requestNotificationPermission(){
 }
 function sendBrowserNotification(title,body,options={}){
   const force=!!(options&&options.force);
-  if(!force&&(!window._notificationsEnabled||!document.hidden)) return;
+  const forceHidden=!!(options&&options.forceHidden);
+  if(!window._notificationsEnabled&&!force) return;
+  if(!forceHidden&&!force&&!document.hidden) return;
   if(!('Notification' in window)) return;
   if(Notification.permission==='granted'){
     _showPwaNotification(title,body,options).catch(()=>{try{new Notification(title||assistantDisplayName(),_notificationOptions(body,options));}catch(_err){}});
--- a/tests/test_pwa_notification_controls.py
+++ b/tests/test_pwa_notification_controls.py
@@ -25,7 +25,7 @@ def test_notification_payload_uses_completion_session_when_provided():
     assert "_sessionUrlForSid(sid)" in MESSAGES_JS
     assert "data:{url}" in MESSAGES_JS
     assert "tag:sid?`hermes-${sid}`" in MESSAGES_JS
-    assert "sendBrowserNotification('Response complete',assistantText?assistantText.slice(0,100):'Task finished',{sid:activeSid})" in MESSAGES_JS
+    assert "sendBrowserNotification('Response complete',assistantText?assistantText.slice(0,100):'Task finished',{forceHidden:true,sid:activeSid})" in MESSAGES_JS
     assert "sendBrowserNotification('Approval required',d.description||'Tool approval needed',{sid:activeSid})" in MESSAGES_JS
     assert "sendBrowserNotification('Clarification needed',d.question||'Tool clarification needed',{sid:activeSid})" in MESSAGES_JS
 


nesquena-hermes:
Pulled this into a read-only worktree and read the diff against `origin/master` plus the surrounding notification plumbing. The diagnosis in #4416 is correct — background-tab SSE throttling queues the `done` event, and by the time it's delivered `document.hidden` is already `false`, so the old guard at messages.js:5327 silently drops the notification. Splitting `force` (bypass everything) from `forceHidden` (bypass only the visibility check, still honor `_notificationsEnabled`) is a clean refactor of the gate:

```js
// static/messages.js (this PR)
function sendBrowserNotification(title,body,options={}){
  const force=!!(options&&options.force);
  const forceHidden=!!(options&&options.forceHidden);
  if(!window._notificationsEnabled&&!force) return;
  if(!forceHidden&&!force&&!document.hidden) return;
```

That correctly fixes the "came back to the tab, throttled `done` arrives late" case, and keeping the `_notificationsEnabled` requirement (unlike `force`) is the right call so a disabled-notifications user still gets nothing.

## Concern: this now notifies even when the user never left the tab

The completion handler passes `forceHidden:true` unconditionally:

```js
// static/messages.js:3512 (this PR)
sendBrowserNotification('Response complete',assistantText?assistantText.slice(0,100):'Task finished',{forceHidden:true,sid:activeSid});
```

The old `!document.hidden` guard did double duty: it dropped the throttled-late notification (the bug), but it *also* suppressed a desktop notification when the user watched the whole response stream in the foreground (correct behavior — you don't want an OS notification for a response you're actively reading). This PR removes both. So the side effect is: every response completion now fires a desktop notification, even for a user sitting on the tab the entire time. For anyone who keeps the WebUI focused while working, that's a notification on every single turn — likely to get reported as a new bug.

## Suggested refinement: track whether the tab was hidden during this stream

The precise signal you want is "was this tab backgrounded at any point during the stream," not "is it hidden right now." `attachLiveStream` already tracks page-hidden state for the error-deferral path — see `_pageHiddenForStreamError()` at messages.js:1921 and the `visibilitychange` listener it wires at messages.js:1957-1964. A minimal addition:

1. Add a stream-scoped flag (e.g. `let _streamWentHidden=false;`) near the other per-stream flags around messages.js:1970.
2. Set it from a `visibilitychange` listener (or piggyback on the existing one) whenever `document.hidden` becomes true during the stream's lifetime.
3. In the `done` handler, pass `{forceHidden:_streamWentHidden, sid:activeSid}` instead of `{forceHidden:true,...}`.

That keeps #4416 fixed (tab was hidden → notify on the late `done`) while restoring the no-notify-when-foreground behavior. The `tag:hermes-${sid}` + `renotify:false` in `_notificationOptions` (messages.js:5280) already de-dupes, so a foreground notification wouldn't stack, but suppression is better than relying on the tag.

## Test plan

The test update at `tests/test_pwa_notification_controls.py` correctly tracks the new call signature. If you adopt the `_streamWentHidden` approach, please add a case asserting the `done` handler passes a *computed* `forceHidden` (keyed off the hidden-tracking flag) rather than a literal `true`, so the foreground-suppression behavior is covered and can't silently regress.

nesquena-hermes:
Closing as superseded by v0.51.527 (#4515), which fixes #4416 at the root cause. Your diagnosis here was right — the done-time document.hidden check is unreliable because Chromium delivers the throttled background-tab SSE late. Rather than fire unconditionally (which also notifies when the tab was never backgrounded), the shipped fix tracks whether the tab was hidden at any point during the stream and notifies on that basis. You're credited as co-author on the shipped commit. Thanks @bergeouss!

PR #4423, opened 2026-06-18
fix(#4293): cascade archive to child sessions when parent is archived

### LINKED ISSUE #4293: bug(sidebar): archiving parent conversation leaves child/delegate sessions visible or stuck

bergeouss:
## Summary

Archiving a parent conversation left child/delegate sessions visible as orphaned rows in the sidebar. The archive endpoint only archived the targeted session without cascading to children linked by `parent_session_id`.

Fixes #4293

### Server-side (`api/routes.py`)
- New `_cascade_archive_to_children(parent_sid)` helper: scans both in-memory SESSIONS cache and on-disk session files for direct children matching `parent_session_id`. Archives each found child and returns the list in the response as `also_archived_child_ids`.
- `/api/session/archive` endpoint now calls the cascade helper when archiving (not when unarchiving).

### Frontend (`static/sessions.js`)
- `_partitionSidebarSessionRows()` also filters out child sessions whose parent is archived, even if the child itself is not archived. This is a defensive fallback for edge cases the server cascade might miss (race conditions, cross-surface children).

### Design decisions
- **Direct children only** (not recursive) — nested grandchildren are a separate concern
- **Unarchive does NOT cascade** — users keep explicit control over children
- **Defense in depth**: server and frontend fixes work independently

## Acceptance criteria
- ✅ Archiving a parent with children hides those children when archived sessions are hidden
- ✅ Children discoverable when archived sessions are shown
- ✅ Individual child archive removes from sidebar
- ✅ Works for delegate/fork child sessions (parent_session_id / relationship_type: child_session)

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -2555,12 +2555,12 @@ def _ensure_full_session_before_mutation(sid: str, session):
     return full_session
 
 
-def _cascade_archive_to_children(parent_sid: str) -> list[str]:
-    """Archive direct child sessions whose parent_session_id matches *parent_sid*.
+def _cascade_archive_to_children(parent_sid: str, unarchive: bool = False) -> list[str]:
+    """Archive (or unarchive) direct child sessions whose parent_session_id matches *parent_sid*.
 
     Scans both the in-memory SESSIONS cache and on-disk session files to cover
     children that may not be resident in memory.  Returns the list of child
-    session IDs that were archived.
+    session IDs that were modified.
     """
     from api.models import get_session
     from api.config import SESSION_DIR, _get_session_agent_lock
@@ -2573,9 +2573,11 @@ def _cascade_archive_to_children(parent_sid: str) -> list[str]:
         try:
             with _get_session_agent_lock(child_sid):
                 child = get_session(child_sid)
-                if getattr(child, "archived", False):
-                    return  # already archived
-                child.archived = True
+                current_archived = bool(getattr(child, "archived", False))
+                target = not unarchive  # archive=True → target=True; unarchive → target=False
+                if current_archived == target:
+                    return  # already in desired state
+                child.archived = target
                 child.save(touch_updated_at=False)
                 archived_ids.append(child_sid)
         except Exception:
@@ -2589,6 +2591,11 @@ def _cascade_archive_to_children(parent_sid: str) -> list[str]:
     with LOCK:
         for cached in list(SESSIONS.values()):
             if getattr(cached, "parent_session_id", None) == parent_sid:
+                # Skip ordinary forks — parent_session_id on a fork means
+                # "branched from", not "is a snapshot continuation child".
+                # Forks carry session_source="fork" (#4293 review remark).
+                if getattr(cached, "session_source", None) == "fork":
+                    continue
                 sid_str = str(getattr(cached, "session_id", ""))
                 if sid_str and sid_str != parent_sid:
                     in_memory_child_sids.append(sid_str)
@@ -2596,7 +2603,10 @@ def _cascade_archive_to_children(parent_sid: str) -> list[str]:
         _try_archive_child(child_sid)
 
     # 2) Scan on-disk session files for children not in memory.
-    needle = f'"parent_session_id": "{parent_sid}"'
+    #    Use load_metadata_only() for structured access to both
+    #    parent_session_id and session_source — avoids the 4096-byte
+    #    head-scan that can miss metadata past the first 4KB (#4293 review).
+    from api.models import Session
     try:
         for spath in SESSION_DIR.glob("*.json"):
             if spath.name.startswith("_"):
@@ -2605,11 +2615,17 @@ def _cascade_archive_to_children(parent_sid: str) -> list[str]:
             if child_sid == parent_sid or child_sid in archived_ids:
                 continue
             try:
-                head = spath.read_text(encoding="utf-8", errors="ignore")[:4096]
-            except (TypeError, OSError):
-                continue
-            if needle in head:
+                meta = Session.load_metadata_only(child_sid)
+                if meta is None:
+                    continue
+                if getattr(meta, "session_source", None) == "fork":
+                    continue
+                if str(getattr(meta, "parent_session_id", "")) != parent_sid:
+                    continue
                 _try_archive_child(child_sid)
+            except Exception:
+                logger.debug("Failed to load metadata for %s", child_sid, exc_info=True)
+                continue
     except Exception:
         logger.debug("Failed to scan session files for archive cascade", exc_info=True)
 
@@ -10041,6 +10057,8 @@ def handle_post(handler, parsed) -> bool:
         also_archived_child_ids = []
         if body.get("archived", True):
             also_archived_child_ids = _cascade_archive_to_children(sid)
+        else:
+            also_archived_child_ids = _cascade_archive_to_children(sid, unarchive=True)
         response_data = {"ok": True, "session": s.compact(), **_worktree_retained_payload(s)}
         if also_archived_child_ids:
             response_data["also_archived_child_ids"] = also_archived_child_ids
--- a/api/routes.py
+++ b/api/routes.py
@@ -2555,6 +2555,67 @@ def _ensure_full_session_before_mutation(sid: str, session):
     return full_session
 
 
+def _cascade_archive_to_children(parent_sid: str) -> list[str]:
+    """Archive direct child sessions whose parent_session_id matches *parent_sid*.
+
+    Scans both the in-memory SESSIONS cache and on-disk session files to cover
+    children that may not be resident in memory.  Returns the list of child
+    session IDs that were archived.
+    """
+    from api.models import get_session
+    from api.config import SESSION_DIR, _get_session_agent_lock
+
+    archived_ids: list[str] = []
+
+    def _try_archive_child(child_sid: str) -> None:
+        if child_sid == parent_sid:
+            return
+        try:
+            with _get_session_agent_lock(child_sid):
+                child = get_session(child_sid)
+                if getattr(child, "archived", False):
+                    return  # already archived
+                child.archived = True
+                child.save(touch_updated_at=False)
+                archived_ids.append(child_sid)
+        except Exception:
+            logger.debug("Failed to cascade archive to child %s", child_sid, exc_info=True)
+
+    # 1) Scan in-memory cache (fast path).
+    #    Collect child IDs while holding LOCK, then archive outside to
+    #    avoid deadlock: _try_archive_child -> get_session also acquires LOCK
+    #    and threading.Lock is non-reentrant.
+    in_memory_child_sids: list[str] = []
+    with LOCK:
+        for cached in list(SESSIONS.values()):
+            if getattr(cached, "parent_session_id", None) == parent_sid:
+                sid_str = str(getattr(cached, "session_id", ""))
+                if sid_str and sid_str != parent_sid:
+                    in_memory_child_sids.append(sid_str)
+    for child_sid in in_memory_child_sids:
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Pulled this into a read-only worktree and read the full diff plus the surrounding code in `api/routes.py`, `api/models.py`, and the sidebar partition path in `static/sessions.js`. This is a solid fix for #4293 — the two-layer approach (server cascade + defensive client filter) directly addresses the orphaned-child-row symptom. A few notes, one of which I'd treat as a blocker.

## What's good

The client-side fallback in `_partitionSidebarSessionRows` is correctly placed:

```js
// static/sessions.js (this PR)
if(!_showArchived&&s.parent_session_id&&archivedSids.has(s.parent_session_id)) continue;
```

This runs after the existing `if(!_showArchived&&s.archived) continue;` line, so an unarchived child whose parent is hidden is now suppressed. And `compact()` only emits `parent_session_id` when it's set (models.py:977-979), so the `s.parent_session_id&&` short-circuit is the right guard — non-forked sessions won't accidentally match an `archivedSids` entry.

## Blocker: archive cascades but unarchive does not

The handler only cascades when archiving:

```python
# api/routes.py (this PR)
if body.get("archived", True):
    also_archived_child_ids = _cascade_archive_to_children(sid)
```

Restoring a parent (`archived:false`, sent from `_archiveSession(session, archived=false)` at sessions.js:3205) leaves the children archived. So the round-trip is lossy: archive a parent → 4 children archived → restore the parent → children stay hidden, and the user now has to hunt them down individually in the archived view. That's arguably worse than the original bug for the restore case. Either cascade the unarchive symmetrically, or (cleaner) record which child IDs *you* archived in this cascade and only auto-restore those — otherwise you'll un-archive children the user had archived independently before the parent.

## Duplicate of existing scan logic in models.py

`_cascade_archive_to_children` reimplements a child-discovery scan that already exists almost verbatim. Compare your on-disk needle scan to `api/models.py:2098-2113`:

```python
# api/models.py (origin/master) — already in the tree
needle = f'"parent_session_id": "{sid}"'
for path in SESSION_DIR.glob('*.json'):
    if path.name.startswith('_') or path.stem == sid:
        continue
    head = path.read_text(encoding='utf-8', errors='ignore')[:4096]
    if needle in head:
        return True
```

Same `4096`-byte head read, same `_`-prefix skip, same needle format. The save path writes `parent_session_id` in `METADATA_FIELDS` (models.py:754) before the messages array, so the head-scan assumption holds — good. But consider factoring out a shared `_iter_child_session_ids(parent_sid)` helper so the two scans can't drift (e.g. if the metadata field order ever changes, one scan would silently stop finding children while the other keeps working).

## Lock ordering — looks correct, worth a test

The two-phase approach (collect child IDs under `LOCK`, then archive outside it because `_get_session_agent_lock` + `get_session` re-acquire `LOCK` and `threading.Lock` is non-reentrant) is the right call and the inline comment explains it well. Please add a regression test that archives a parent with both an in-memory child and a disk-only child, asserts both flip `archived=True` and appear in `also_archived_child_ids`, and a second test that a non-streaming send during cascade doesn't deadlock.

Net: the archive direction and client filter are correct; the unarchive asymmetry is the thing I'd fix before merge.

nesquena-hermes (changes requested):
Thanks @bergeouss — cascading the archive server-side (so children don't linger as orphaned sidebar rows) is the right shape for #4293, and the lock ordering here is careful and correct (collecting in-memory child IDs under `LOCK` then archiving outside it to avoid the non-reentrant deadlock is exactly right). I ran this through the full review gate (Codex + Opus + the suite) and it's close, but there's one **core behavior bug** plus two hardening items to address before it can ship.

### 1. (must-fix, CORE) The cascade archives independent forks

`parent_session_id` is **overloaded** in this codebase — it marks two very different relationships:

- **compression-continuation snapshot parents** (the case #4293 is about — a child sidecar that continues a parent whose transcript was snapshotted), and
- **ordinary forks** created by `/api/session/branch`, which set `parent_session_id = source.session_id` but **must remain independent conversations**.

`static/ui.js` is explicit about this (the lineage-stitch helper docstring): *"ordinary forks also carry `parent_session_id` but must remain independent conversations."*

`_cascade_archive_to_children()` treats **any** `parent_session_id` match as a child to archive, so archiving a session would also archive every fork ever branched from it — surprising, data-affecting, and not what #4293 asks for.

**Fix:** skip ordinary forks in the cascade. Forks created by the branch path carry `session_source == "fork"` (see `/api/session/branch`), so gate `_try_archive_child` (and the `static/sessions.js` defensive-hide fallback) to only treat snapshot-continuation children as children — i.e. skip `session_source == "fork"`. (If the intent really is to also archive forks, that's a product decision worth raising explicitly rather than doing implicitly.)

### 2. (should-fix) On-disk child discovery can miss large non-resident children

The disk scan does a literal substring match (`'"parent_session_id": "<sid>"'`) on only the **first 4096 bytes** of each session file. But `parent_session_id` sits near the **end** of the serialized metadata, *after* potentially large fields (`pre_compression_snapshot`, `context_engine_state`, `gateway_routing_history`, `pending_attachments`, …). A non-resident child with a sizable compression snapshot will have its `parent_session_id` pushed past byte 4096 and be silently missed — so the "durable state" half of the fix won't fire for exactly the compression-continuation children it most targets.

**Fix:** read the `parent_session_id` field properly instead of substring-scanning a fixed window — use the existing metadata-only loader (`Session.load_metadata_only(...)`) / a bounded metadata-prefix parse and compare the parsed field.

### 3. (should-fix, low) Per-archive O(N) glob + latent profile gap

- Every archive click runs `SESSION_DIR.glob("*.json")` + a per-file read. `api/models.py` already maintains a dir-mtime-cached persisted-id list — reuse it, or skip the disk pass when the in-memory pass already found the children (the common case), to avoid O(N) file opens on a large store.
- `SESSION_DIR` is a single global store shared across profiles (sessions are distinguished by a `.profile` attribute, not separate dirs). A cross-profile mutation is practically impossible today (forks inherit the parent's profile, and session IDs are unique), but a one-line defensive `child.profile == parent.profile` check inside `_try_archive_child` would close the latent gap cheaply and keep it aligned with the project's profile-isolation invariant.

(1) is the blocker. Once forks are excluded and the disk discovery reads the real field, I'm happy to re-gate and ship — the cascade approach is sound and this supersedes the frontend-only #4414 nicely. Happy to do the rebase for you at re-gate.

bergeouss:
**Review fixes pushed** ✅

All points addressed:

1. **(must-fix)** Skip ordinary forks (`session_source == "fork"`) in both in-memory and on-disk scans — forks remain independent conversations.
2. **(should-fix)** Replaced 4096-byte head-scan with `Session.load_metadata_only()` for structured access to `parent_session_id` and `session_source`. No longer misses children with large compression snapshots.
3. **Unarchive symmetry** — `_cascade_archive_to_children` now accepts `unarchive=True`, and the archive-handler cascades both directions. Restoring a parent also restores its cascade-archived children.
4. **(frontend)** `_partitionSidebarSessionRows` defensive fallback also skips `session_source == "fork"` sessions.

CI should pick up automatically.

nesquena-hermes:
Closing as superseded by #4414. ✅

The bug this targets (#4293 — archiving a parent conversation leaves orphaned child rows in the sidebar) is already fixed in master: **#4414 merged on 2026-06-18 (shipped in v0.51.499)** and **issue #4293 is closed**. #4414 solves the user-visible symptom display-only — `hasHiddenArchivedAncestor` walks the parent chain, suppresses orphan child rows, and clears stale decorations with **zero state mutation**.

#4423 takes a heavier approach — a backend cascade that actually archives the child sessions in the DB when a parent is archived. That's a more surprising behavior (archiving one conversation silently archives N others on disk), and it carries the fork-independence edge case from the earlier review (forks must stay independent). Since the symptom is already resolved with no state-mutation risk, the cascade isn't needed.

The work was solid and the root-cause read was correct — thanks @bergeouss. If there's a specific scenario where the display-only fix isn't enough (e.g. you want the children genuinely archived for export/retention reasons), open an issue describing it and we can revisit the cascade as a deliberate feature rather than a bugfix.

PR #2689, opened 2026-05-21
fix(chat): preserve inflight send state during start race

ai-ag2026:
## Summary
- preserve the optimistic in-flight message array before `/api/chat/start` awaits
- recreate the active `INFLIGHT` entry if a sidebar/session refresh prunes it during that await window
- skip stale inflight cleanup for the currently submitting session before a stream id exists
- add regression coverage for the start/send race invariants

## Test plan
- `env -u HERMES_WEBUI_PASSWORD TZ=UTC LANG=C.UTF-8 LC_ALL=C.UTF-8 uv run --with pytest --with pyyaml python -m pytest -q tests/test_inflight_send_start_race.py tests/test_inflight_purge_missing_sessions.py tests/test_issue2066_stale_sidebar_spinner.py tests/test_sidebar_first_turn_visibility.py tests/test_issue500_session_list_virtualization.py`
- `node --check static/messages.js && node --check static/sessions.js`
- `git diff --check origin/master...HEAD`

PR #4424, opened 2026-06-18
fix(#4354): unstuck running indicator when SSE dies during provider retry

### LINKED ISSUE #4354: bug(streaming): 'Running' indicator stays stuck after long provider connection errors
## Summary

The WebUI's "Conversation is running" indicator and its ticking activity timer can remain stuck on a session **after the agent run has fully completed on the server** and an `**Error:** Connection error.` message has been persisted to the session file. The user sees a single spinner with no assistant reply, but the server reports `is_streaming: false, active_stream_id: null` and the session is otherwise usable on the next send.

This is a follow-up to #2066: same `INFLIGHT`-cache-leak family, different trigger (long provider stall instead of page refresh / network drop). Several of the guards added in #2068 / #2455 / #2499 / #2796 only trigger on `_sendInProgress === false` and don't help when the client has lost its SSE listener while the server is still mid-retry.

## Reproduction

1. Use a provider that hangs the connection and then errors out (e.g. `minimax-cn` on MiniMax-M3 hitting `APIConnectionError` after 300s of stream-stale timeouts). Any provider that produces `APIConnectionError` rather than clean chunks will reproduce this.
2. Send a message in any session.
3. Provider's stream stalls for 300s; server's `agent.chat_completion_helpers` kills the connection and retries. Repeat 3 times (the configured retry budget). Each retry takes ~5 min (1× 300s stale-stream timeout + ~5s API call before next failure).
4. Total wall time: ~32 minutes. During that time, the SSE channel is **open but silent** — the server hasn't pushed any `token`, `state_saved`, or `pending` events because the run is blocked in the agent thread, not in the streaming writer.
5. After ~32 min, server finally raises `APIConnectionError` past retry budget. `api/streaming.py:8216-8434` enters the `except` handler, calls `_snapshot_and_append_partial_on_error(s, stream_id)`, appends `**Error:** Connection error.` to `s.messages`, saves, and emits `put('apperror', _error_payload)` to the SSE queue.
6. By that time, the **client's `EventSource` has often been silently killed** by the browser (idle SSE connection past the platform's keepalive, or replaced during a session-list re-render that was triggered by the periodic `/api/sessions` poll). The `apperror` event is queued server-side and never delivered.
7. The client ends up with:
   - `S.busy = true` (set in the send-path)
   - `S.activeStreamId = <stale stream_id>` (set when the SSE source was first attached)
   - `INFLIGHT[sid]` populated in `localStorage` (persisted by `saveInflightState`)
   - Server already idle, `is_streaming: false`
8. Clicking back into the session re-runs the INFLIGHT-reattach path in `sessions.js:1166-1212` and re-sets `S.busy = true` (line 1186), even though the server says the session is idle. `_reconcileActiveSessionIdleStateFromList` (line 382) bails out at line 384 because `_sendInProgress` is still `true` (also stuck — the function's `finally` at `messages.js:1329` never ran because `attachLiveStream` was the failing step).

## Server-side log (verbatim from this reproduction)

```
06:32:54 WARNING [f4cd0f9d1b2e] agent.conversation_loop: Retrying API call in 2.1s (attempt 1/3) error=Connection error.
06:37:56 WARNING [f4cd0f9d1b2e] agent.chat_completion_helpers: Stream stale for 300s (threshold 300s) — no chunks received. Killing connection.
06:39:07 WARNING [f4cd0f9d1b2e] agent.conversation_loop: API call failed (attempt 2/3) error_type=APIConnectionError
06:44:13 WARNING [f4cd0f9d1b2e] agent.chat_completion_helpers: Stream stale for 300s
06:50:30 WARNING [f4cd0f9d1b2e] agent.chat_completion_helpers: Stream stale for 300s
06:56:44 WARNING [f4cd0f9d1b2e] agent.chat_completion_helpers: Stream stale for 300s
07:02:59 WARNING [f4cd0f9d1b2e] agent.chat_completion_helpers: Stream stale for 300s
07:04:09 WARNING [f4cd0f9d1b2e] agent.conversation_loop: API call failed after 3 retries. Connection error.
07:04:09 ERROR   [f4cd0f9d1b2e] agent.conversation_loop: API call failed after 3 retries. Connection error. | msgs=183 tokens=~193,746
```

Server-side session state at the same moment:

```json
{
  "session_id": "f4cd0f9d1b2e",
  "active_stream_id": null,
  "pending_user_message": null,
  "is_streaming": false,
  "messages": [..., "**Error:** Connection error."]
}
```

## Root cause

Three independent gaps in the busy-state machine conspire to make this state unrecoverable without a page refresh:

1. **No client-side "expected event" watchdog.** `attachLiveStream` (`messages.js` ~line 3267) opens the `EventSource` and updates the UI in response to `token` / `interim_assistant` / `state_saved` / `done` / `apperror` / etc. There is no client timer that says *"I should have heard something by now — if not, the stream is dead."* When the server goes silent for >30 min in retry loops, the client has no reason to clear state.

2. **`_reconcileActiveSessionIdleStateFromList` is gated on `_sendInProgress === false`** (`sessions.js:384`). If the user's `send()` flow is the one that hung, `_sendInProgress` is stuck `true` (the `finally` at `messages.js:1329` only runs after `attachLiveStream` returns; if SSE setup or the `await api('/api/chat/start', ...)` call hangs, it never runs). The result: even when `/api/sessions` returns `is_streaming: false`, the busy state is never reconciled.

3. **The INFLIGHT reattach path re-asserts busy state on every session entry** (`sessions.js:1186` — `S.busy = true` is set unconditionally if `INFLIGHT[sid]` exists). The check uses local INFLIGHT truth, not server truth. So even if the user navigates away and back, the stuck state re-manifests.

This is a degenerate combination: in the original #2066 case, `_sendInProgress` was `false` and `_reconcileActiveSessionIdleStateFromList` could do its job. In the #2066 fix (`#2068` / `#2455`), the focus was on stale INFLIGHT from page refresh / network drop. **The new failure mode is an in-flight send whose SSE connection died before the server's `apperror` made it across.** None of the existing guards cover it.

## Impact

- **Visible to every user**: a single "Running" message with a ticking timer, no assistant response, no error surfaced, no way to clear without refreshing.
- **Masks provider outages**: when an upstream LLM is having a bad day, multiple users hit this simultaneously and see no indication of *why* their message isn't going through.
- **Recovery is non-obvious**: most users will not realize a hard refresh is the fix. Some may assume the message was lost and re-send, which can compound the issue (the server-side `**Error:** ...` message is now in their transcript and the new message is sent into a stale session).
- **Token cost impact**: the server retries 3× over 30+ min, with the full 193k-token context being re-sent on every retry (`Stream stale for 300s ... context=~193,746 tokens`). A user who knows to refresh ends up paying for the retries anyway, with no error feedback.

## Suggested fix directions

In rough order of size:

1. **Server-side heartbeat during retry loops** (smallest, best UX): when `agent.conversation_loop` enters a retry, emit a `pending` or `state_saved` SSE event with a `phase: "retrying"` payload and the retry attempt number. The client already handles `state_saved`; surfacing it as "Reconnecting to provider (attempt 2/3)…" eliminates the "is it stuck?" anxiety that leads to refreshes in the first place. Bonus: this also warms the SSE keepalive so the browser doesn't drop the EventSource during 30-min silence.

2. **Client-side idle-stream watchdog**: in `attachLiveStream`, set a `lastEventAt` timestamp on every received event. Add a `setInterval` that, after N minutes of silence (e.g. 5 min) while `S.busy` is `true`, force-closes the EventSource, clears `INFLIGHT[activeSid]`, and lets the periodic `/api/sessions` poll reconcile via the existing `_reconcileActiveSessionIdleStateFromList` path. The key invariant: **the watchdog must not gate on `_sendInProgress`** — it must be able to clear state even when the send flow itself is stuck.

3. **Make the INFLIGHT reattach on session-entry trust server truth over local cache**: at `sessions.js:1186`, only set `S.busy = true` if the server-side `S.session.active_stream_id` matches `INFLIGHT[sid].streamId` *and* the session was last-updated within the last N minutes. Otherwise treat the local INFLIGHT as stale and discard it (the `INFLIGHT[sid]` is already deleted in the older `inflightHasVisibleLiveState` path above, but the `S.busy = true` line below it is not similarly guarded).

4. **`_reconcileActiveSessionIdleStateFromList` should not bail on `_sendInProgress`**: instead, if the server says idle, force-clear `_sendInProgress` and `_sendInProgressSid` as part of the reconciliation. The original guard was to avoid clobbering an in-progress send, but the cost of clobbering (one extra request from the user) is much smaller than the cost of the stuck state we're hitting now.

## Reproduction package

- Session file: `/home/hermeswebui/sessions/f4cd0f9d1b2e.json` (server-side, has the `**Error:** Connection error.` message at the end and `active_stream_id: null`)
- Request dump: `/home/hermeswebui/sessions/request_dump_f4cd0f9d1b2e_20260617_070409_323636.json` (`reason: max_retries_exhausted, error: APIConnectionError`)
- Server logs: `/home/hermeswebui/logs/errors.log` and `agent.log` for the timestamp range 2026-06-17 06:32 → 07:04
- Provider: `minimax-cn` (base_url `https://api.minimaxi.com/anthropic`, model `MiniMax-M3`)
- Client state at time of report: WebUI shows "Running" with ticking timer; server `/api/sessions` returns `is_streaming: false`

## Related issues

- #2066 — original "stuck sidebar spinner on completed sessions" (same INFLIGHT root cause, different trigger; #2068 fixed that case)
- #4313 — `EventSource` lifecycle noise (related: same SSE-connection management surface)
- #4029 — stale bg-task notifications (related: same "stale local state outranks fresh server state" pattern)
- #2233 — WebUI memory leak causes session to completely freeze (may share a deeper root cause around never-cleared event listeners on dead EventSources)
- #2957 — Add cost-protection pause for suspicious long-running sessions (related: a cost-protection pause would naturally cap the 30-min retry cost that this bug currently incurs)

bergeouss:
## Summary

The "Conversation is running" indicator and ticking timer can remain stuck after the agent run has fully completed server-side, when the SSE connection dies before the `apperror` event is delivered. Three independent gaps in the busy-state machine conspire to make this state unrecoverable without a page refresh.

Fixes #4354

## Root cause (from issue)

1. No client-side watchdog: `attachLiveStream` opens EventSource but has no idle timer to detect silence
2. `_reconcileActiveSessionIdleStateFromList` gated on `_sendInProgress === false` — but `_sendInProgress` stays stuck true when the send flow SSE dies before the `finally` block runs
3. INFLIGHT reattach unconditionally sets `S.busy = true` from local cache, ignoring server truth

## Changes (3 targeted edits in `static/sessions.js`)

| Location | Before | After | Effect |
|----------|--------|-------|--------|
| `_reconcileActiveSessionIdleStateFromList` L384 | `if (_sendInProgress) return false;` | Comment (guard removed) | Server `is_streaming=false` now overrides stuck client state |
| `_purgeStaleInflightEntries` L438-440 | Skip if `_sendInProgress && sid === _sendInProgressSid` | Comment (skip removed) | Stale INFLIGHT entries purged even during hung sends |
| `loadSession` INFLIGHT reattach L1192 | `S.busy = true;` | `S.busy = !!activeStreamId;` | Only assert busy if server confirms active stream |

## Combined effect

When a provider hangs for 30+ min, retries exhaust, and SSE dies before `apperror` is delivered:
1. The periodic `/api/sessions` poll returns `is_streaming: false`
2. `_reconcileActiveSessionIdleStateFromList` now clears `S.busy`, `S.activeStreamId`, and `INFLIGHT[sid]` (previously bailed at `_sendInProgress` guard)
3. `_purgeStaleInflightEntries` now removes the stale INFLIGHT entry (previously skipped it)
4. If user navigates away and back, `loadSession` only asserts busy if `activeStreamId` is truthy

No page refresh needed — the stuck spinner clears on the next session list poll (~15s).

## Risk assessment

- **Minimal**: 3 insertions, 5 deletions, purely in busy-state reconciliation paths
- **`_sendInProgress` guard removal**: the worst case is clearing busy state during a genuinely in-progress send. The periodic poll runs every ~15s, and the server would not report `is_streaming=false` while the stream is active. If a false positive somehow occurs, the user just sees a brief busy→idle→busy transition on the next token event.
- **`!!activeStreamId`**: activeStreamId comes from the server session object (`s.active_stream_id`), which is the source of truth.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -381,7 +381,7 @@ function _isServerIdleSessionRow(s) {
 
 function _reconcileActiveSessionIdleStateFromList(serverRows) {
   if (!S || !S.session || !S.session.session_id) return false;
-  if (typeof _sendInProgress !== 'undefined' && _sendInProgress) return false;
+  // #4354: Removed _sendInProgress guard — server is_streaming=false is authoritative.
   if (!Array.isArray(serverRows)) return false;
   const sid=S.session.session_id;
   const serverRow=serverRows.find(s=>s&&s.session_id===sid);
@@ -435,9 +435,7 @@ function _purgeStaleInflightEntries() {
     }
   }
   for (const sid of Object.keys(INFLIGHT)) {
-    if (typeof _sendInProgress !== 'undefined' && _sendInProgress && sid === _sendInProgressSid) {
-      continue;
-    }
+    // #4354: Removed _sendInProgressSid skip — purge stale INFLIGHT even during hung sends.
     if (!sessionsById.has(sid)) {
       // Session is absent from _allSessions — it was deleted / archived /
       // filtered and can never stream again, so drop the entry.
@@ -1189,7 +1187,7 @@ async function loadSession(sid){
     }
     // Refresh todos from cold-load or persisted INFLIGHT before painting.
     if(typeof _hydrateTodosFromSession==='function') _hydrateTodosFromSession(S.session);
-    S.busy=true;
+    S.busy=!!activeStreamId;  // #4354: Only assert busy if server confirms active stream.
     // appendLiveToolCard() is guarded by S.activeStreamId; restore it before
     // replaying persisted live tools so the compact Activity count survives
     // switching away from and back to an active chat (#1715).
--- a/tests/test_regressions.py
+++ b/tests/test_regressions.py
@@ -656,14 +656,14 @@ def test_loadSession_inflight_sets_busy_before_renderMessages(cleanup_test_sessi
     inflight_idx = src.rfind("if(INFLIGHT[sid]){")
     assert inflight_idx >= 0, "INFLIGHT branch not found in loadSession"
     inflight_block = src[inflight_idx:inflight_idx+4200]
-    busy_pos = inflight_block.find("S.busy=true;")
+    busy_pos = inflight_block.find("S.busy=")
     # #3326 added an optional {preserveScroll} arg to the INFLIGHT-branch render
     # call, so match the call form rather than the bare `renderMessages();`.
     render_pos = inflight_block.find("renderMessages(")
-    assert busy_pos >= 0, "loadSession INFLIGHT branch must set S.busy=true"
+    assert busy_pos >= 0, "loadSession INFLIGHT branch must set S.busy"
     assert render_pos >= 0, "loadSession INFLIGHT branch must call renderMessages()"
     assert busy_pos < render_pos, \
-        "loadSession must set S.busy=true before renderMessages() to avoid duplicate tool cards"
+        "loadSession must set S.busy before renderMessages() to avoid duplicate tool cards"
 
 
 def test_loadSession_inflight_merges_tail_with_persisted_transcript(cleanup_test_sessions):


nesquena-hermes:
Pulled this into a read-only worktree and read all three hunks against `origin/master`, plus the history of the guards being removed. The goal — letting a server `is_streaming=false` authoritatively clear a stuck "Running" indicator — is right, and CI is green. But two of the three changes remove guards that were added *deliberately* to fix earlier races, and the existing regression test for one of them now passes on a technicality rather than on behavior.

## 1. `_purgeStaleInflightEntries` — the removed skip has live test coverage that this PR defeats

```js
// sessions.js:437 (this PR)
for (const sid of Object.keys(INFLIGHT)) {
  // #4354: Removed _sendInProgressSid skip — purge stale INFLIGHT even during hung sends.
  if (!sessionsById.has(sid)) { delete INFLIGHT[sid]; ... }
```

That skip was added by PR #2689 (`3dd2ace4` "fix(chat): preserve inflight send state during start race") for a real bug: between `send()` setting `_sendInProgress=true` (messages.js:962) and `/api/chat/start` returning a `streamId`, the session has no `active_stream_id` yet, so a sidebar refresh that lands in that window sees `is_streaming=false` and purges the just-created optimistic INFLIGHT entry — the user's bubble vanishes.

`tests/test_inflight_send_start_race.py::test_stale_inflight_purge_preserves_current_send_before_stream_id_exists` asserts exactly this:

```python
assert "_sendInProgress" in body and "_sendInProgressSid" in body, (
    "_purgeStaleInflightEntries() should skip the current send while start is in progress"
)
skip_idx = body.index("_sendInProgress")
delete_idx = body.index("delete INFLIGHT[sid];")
assert skip_idx < delete_idx
```

This test is **still green on your branch** — but only because your replacement comment at sessions.js:438 literally contains the substrings `_sendInProgressSid` and `_sendInProgress`. The `body.index("_sendInProgress")` finds the comment text, and `skip_idx < delete_idx` holds because the comment sits above the delete. The guard *behavior* is gone; the assertion just can't tell. That's a silent regression — the start-race window #2689 fixed is reopened, and the test that should catch it is neutralized by the comment wording.

## 2. `_reconcileActiveSessionIdleStateFromList` — same start-race exposure

```js
// sessions.js:384 (this PR)
// #4354: Removed _sendInProgress guard — server is_streaming=false is authoritative.
```

During the same pre-`streamId` window, a list refresh now runs the full idle-reconcile body (sessions.js:391-405): `S.busy=false`, `delete INFLIGHT[sid]`, `clearInflightState(sid)`, plus `_scheduleActiveSessionIdleReload`. For a turn that's mid-`/api/chat/start`, the server genuinely reports `is_streaming=false` because the stream hasn't been registered yet — so this can tear down a perfectly healthy in-flight send, not just a hung one.

## Recommendation

Don't drop the `_sendInProgress`/`_sendInProgressSid` guards wholesale. The distinction you need is "hung send" vs "send still starting," and you already have the signal: `_sendInProgressSid` is set at messages.js:1138 and cleared in the `finally` at messages.js:1398. A targeted fix:

- Keep the skip, but bound it with a staleness check — e.g. only honor the skip while a `_sendInProgressStartedAt` timestamp is within a few seconds, so a *hung* send (the #4354 symptom) eventually becomes purgeable while a *normal* starting send is still protected.
- Prefer the silence-watchdog approach from your own PR description (force-close a dead EventSource after N seconds of no events) as the primary fix — that targets the stuck stream directly without weakening the start-race guards.

The `loadSession` change (`S.busy=!!activeStreamId` at sessions.js:1189) looks fine on its own — that one is keyed off the server-confirmed `active_stream_id`, not the client-only send flag.

## Test plan

Before re-pushing, update `test_inflight_send_start_race.py` so the assertion checks the *guard logic* (e.g. `"sid === _sendInProgressSid" in body` and a `continue`), not bare substrings that a comment satisfies — otherwise the test can't protect either side of this change. Also worth confirming PR #4355 (also targeting #4354) isn't a competing fix; the two should be reconciled rather than both landing.

nesquena-hermes:
Shipped in v0.51.526 via the rebased release branch (credited to you in CHANGELOG + release notes). Thanks @bergeouss! On the way in I scoped the _sendInProgress guard removal to skip only the actively-mid-send session (so #4354 still unsticks a hung indicator without re-opening the #2689 start-race flicker), and hardened the regression test. Codex SAFE + Opus SHIP-WITH-FIXES (applied) + full suite green (9650).

PR #4435, opened 2026-06-18
feat: offline conversation cache for PWA weak-signal resilience

kaishi00:
## Problem

The PWA shows a blank screen when the connection is weak or unavailable. The service worker caches the app shell (HTML/CSS/JS), but all API calls bypass the cache entirely — so the sidebar shows nothing and sessions won't load. The existing offline banner only fires on hard `navigator.onLine === false` events, missing the common case of a weak signal where requests time out.

## Solution

An IndexedDB cache layer that lets users browse cached conversations while the connection sorts itself out.

### How it works

1. **Write-through caching** — every successful `GET /api/sessions` and `GET /api/session?messages=1` response is cached automatically as it flows through `api()`
2. **Background warming** — after every successful sidebar load, silently fetches and caches the 5 most recent sessions. So "today's conversations" are always pre-cached before you need them
3. **Fallback** — on GET failure (timeout, network error), falls back to IndexedDB before giving up. This catches both hard offline AND weak-signal scenarios
4. **Amber banner** — "Offline — showing cached conversations (updated 2h ago)" when serving cached data. Auto-fades after 5s, hides when connection returns
5. **Send button disabled** — the send button greys out when offline with tooltip "Offline — connect to send messages". Re-enables automatically when connection returns

### Design decisions

- **Media stripping**: base64 images and tool outputs >5KB are stripped before caching. Text context is preserved — broken images offline is an acceptable tradeoff
- **2MB per-session cap**, max 20 sessions cached, oldest evicted (LRU)
- **Version-tagged**: cache is nuked automatically when the WebUI version changes (new deploy) to prevent stale JSON shape mismatches
- **Only GETs cached** — mutations (POST/PUT/DELETE) always go to network
- **Uses existing infrastructure**: integrates with `_browserReportsOnline()` helper and `initOfflineMonitor()` rather than bolting on parallel event listeners
- **Complements existing offline banner**: the existing `showOfflineBanner` in ui.js fires on hard offline events; this cache banner fires on actual API failure, covering the weak-signal gap

### Files changed

| File | Changes |
|------|--------|
| `static/offline-cache.js` (new, 461 lines) | IndexedDB cache layer: write-through caching, background warming, media stripping, offline banner, version-tagged eviction |
| `static/workspace.js` (+77 lines) | Hooks into `api()` for cache write-through on successful GETs and async fallback to IndexedDB on failure |
| `static/sessions.js` (+9 lines) | Triggers background warming + banner hide after successful sidebar load |
| `static/ui.js` (+5/-5 lines) | Disables send button via `_browserReportsOnline()` check; adds `updateSendBtn()` to existing online/offline listeners |
| `static/index.html` (+1 line) | Registers `offline-cache.js` before `workspace.js` |
| `static/sw.js` (+1 line) | Adds `offline-cache.js` to shell assets |

### Testing

- Verified on iOS PWA: sessions load from cache when offline, banner appears, send button disables
- Verified recovery: connection returns → banner hides, send button re-enables, live data refreshes
- All JS files pass `node -c` syntax check
- Background warming fires silently after sidebar load with no UI impact

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -2093,6 +2093,10 @@ function applyBotName(){
   }
   // Fetch active profile
   try{const p=await api('/api/profile/active');S.activeProfile=p.name||'default';S.activeProfileIsDefault=!!p.is_default;}catch(e){S.activeProfile='default';S.activeProfileIsDefault=true;}
+  // Initialize offline cache scope to the active profile (privacy isolation)
+  if(window.HermesOfflineCache&&typeof window.HermesOfflineCache.setScope==='function'){
+    try{await window.HermesOfflineCache.setScope(S.activeProfile);}catch(_){}
+  }
   applyBotName();
   // Update profile chip label immediately
   const profileLabel=$('profileChipLabel');
--- a/static/index.html
+++ b/static/index.html
@@ -1403,6 +1403,17 @@
               <button class="sm-btn" onclick="saveDashboardSettings()" data-i18n="settings_dashboard_save" style="margin-top:8px;width:100%;padding:7px;font-weight:600">Save dashboard link settings</button>
               <div id="settingsDashboardStatus" class="settings-autosave-status" aria-live="polite"></div>
             </div>
+            <!-- Offline Cache Section -->
+            <div class="settings-field" style="margin-top:18px;padding-top:16px;border-top:1px solid var(--border)">
+              <label>Offline Conversation Cache</label>
+              <div style="font-size:11px;color:var(--muted);margin-bottom:8px">Cache the session sidebar and recent transcripts in this browser (IndexedDB) so they're readable when the connection drops. Transcripts are profile-scoped and wiped on sign-out. <strong>Off by default.</strong></div>
+              <label style="display:flex;align-items:center;gap:8px;font-size:13px;font-weight:normal;cursor:pointer">
+                <input type="checkbox" id="settingsOfflineCacheEnabled" onchange="toggleOfflineCache(this.checked)" style="width:15px;height:15px;accent-color:var(--accent)">
+                <span>Enable offline conversation cache</span>
+              </label>
+              <button class="sm-btn" onclick="clearOfflineCacheUI()" style="margin-top:8px;width:100%;padding:7px;font-weight:600;color:#e8a030;border-color:rgba(232,160,48,.3)">Clear offline cache now</button>
+              <div id="offlineCacheStatus" class="settings-autosave-status" aria-live="polite"></div>
+            </div>
             <!-- Gateway Status Section -->
             <div class="settings-field" style="margin-top:18px;padding-top:16px;border-top:1px solid var(--border)">
               <label data-i18n="settings_label_gateway_status">Gateway Status</label>
@@ -1562,6 +1573,8 @@
 <script src="static/icons.js?v=__WEBUI_VERSION__" defer></script>
 <script src="static/assistant_turn_anchors.js?v=__WEBUI_VERSION__" defer></script>
 <script src="static/ui.js?v=__WEBUI_VERSION__" defer></script>
+<script>window.__WEBUI_VERSION__='__WEBUI_VERSION__';</script>
+<script src="static/offline-cache.js?v=__WEBUI_VERSION__" defer></script>
 <script src="static/workspace.js?v=__WEBUI_VERSION__" defer></script>
 <script src="static/terminal.js?v=__WEBUI_VERSION__" defer></script>
 <script src="static/sessions.js?v=__WEBUI_VERSION__" defer></script>
--- a/static/offline-cache.js
+++ b/static/offline-cache.js
@@ -461,9 +461,15 @@
       // Sort newest-first, evict the tail
       entries.sort(function (a, b) { return b.cachedAt - a.cachedAt; });
       var toEvict = entries.slice(MAX_CACHED_SESSIONS);
+      // Queue ALL delete requests synchronously before awaiting any of them.
+      // IndexedDB transactions auto-commit when the call stack returns to the
+      // event loop with no pending requests — sequential awaits between deletes
+      // cause TransactionInactiveError on strict implementations (iOS Safari).
+      var deletePromises = [];
       for (var i = 0; i < toEvict.length; i++) {
-        await _txDelete(store, toEvict[i].key);
+        deletePromises.push(_txDelete(store, toEvict[i].key));
       }
+      await Promise.all(deletePromises);
     } catch (_) {
       // non-fatal
     }
--- a/static/offline-cache.js
+++ b/static/offline-cache.js
@@ -1,0 +1,628 @@
+/**
+ * Hermes WebUI — Offline Conversation Cache
+ *
+ * Uses IndexedDB to cache:
+ *   1. The session list (/api/sessions) — the sidebar data
+ *   2. Individual session conversations (/api/session?session_id=X&messages=1)
+ *
+ * Cache is warmed proactively for the 5 most recent sessions after every
+ * successful sidebar load, so "today's conversations" are always available
+ * when the connection drops or is too weak to load.
+ *
+ * On GET failures, api() falls back to this cache. A visual banner tells
+ * the user they're reading cached data with a relative timestamp.
+ *
+ * PRIVACY MODEL (review #4435):
+ *   - Profile-scoped: each profile gets its own IndexedDB database.
+ *     Profile B can NEVER read Profile A's cached data.
+ *   - Opt-in: transcript caching is disabled by default. The user must
+ *     explicitly enable it via the settings panel. When disabled, no data
+ *     is cached and existing cache is wiped.
+ *   - Logout-safe: nukeCache() is called on sign-out and profile switch.
+ *
+ * Storage strategy:
+ *   - Sessions larger than 2MB (after media stripping) are not cached
+ *   - Base64 data URIs recursively stripped from ALL string fields
+ *   - Tool result fields truncated; attachment/thumbnail fields dropped
+ *   - Cache is version-tagged; mismatches nuke the store on load
+ */
+(function () {
+  'use strict';
+
+  // ── Constants ─────────────────────────────────────────────────────────
+  var DB_PREFIX = 'hermes-offline-cache';
+  var LEGACY_DB_NAME = 'hermes-offline-cache'; // pre-scoping DB name (no profile suffix)
+  var DB_VERSION = 1;
+  var STORE_SESSION_LIST = 'sessionList'; // single record, key='current'
+  var STORE_SESSIONS = 'sessions'; // keyed by session_id
+  var MAX_SESSION_BYTES = 2 * 1024 * 1024; // skip caching sessions > 2MB
+  var MAX_CACHED_SESSIONS = 20; // evict oldest beyond this
+  var WARM_COUNT = 5; // number of recent sessions to proactively cache
+  var SETTING_KEY = 'hermes-offline-cache-enabled';
+
+  // Matches data:*;base64,... sequences of 100+ chars (real embedded media,
+  // not short inline tokens)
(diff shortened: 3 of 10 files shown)


kaishi00:
Good catches from Greptile. Both bugs fixed in c34b52b:

**1. Timeout fallback unreachable (Greptile main issue)**
The `isTimeout` branch in `api()` throws before reaching the catch block where the cache fallback was. Weak-signal requests that hang and timeout are the primary use case for this feature — and they were the one path that never consulted IndexedDB.

Fix: added cache fallback in BOTH the timeout path (before `throw err`) and the network-error path. Extracted into `_tryOfflineCacheFallback()` helper, typeof-guarded so tests that extract `api()` source in isolation still work.

**2. `_evictExcess` transaction-lifetime bug**
The function used `_txKeys()` then separate `_txGet()`/`_txDelete()` calls. IndexedDB transactions auto-commit when there's no pending request, so the transaction died between the key scan and the delete loop — the count cap was never enforced.

Fix: replaced with a single-cursor approach (`store.openCursor()` + `cursor.continue()`) that keeps the transaction alive through the entire scan+delete sequence.

**3. Test fixes**
Two existing tests (`test_4170_offline_probe`, `test_offline_banner`) did exact-string matching on `initOfflineMonitor` event listener bodies. Our `updateSendBtn()` addition broke them. Relaxed to substring checks that verify the key patterns are present.

nesquena-hermes (changes requested):
Thanks @kaishi00 — the weak-signal PWA problem this targets is real, and the structure (write-through + fallback + amber banner) is reasonable. I did a full deep review (Codex + complete suite). I'm requesting changes: there's a **cross-profile privacy leak** that's a hard blocker, plus a cluster of correctness/privacy issues. Worth fixing the privacy model first before polishing the rest.

### 🔴 Blocker: cross-profile transcript leak

`DB_NAME = 'hermes-offline-cache'` (`offline-cache.js:23`) is a fixed string; the session list is keyed by literal `'current'` and transcripts only by `sessionId` (`:205`, `:237`), and the fallback returns cached data as a successful `/api/sessions` / `/api/session` response. IndexedDB is **per-origin**, but Hermes profiles are selected by cookie on the same origin — so after a profile switch or relogin, **Profile B can read Profile A's cached sidebar + full transcripts.** That violates "profiles are independent islands."

**Fix:** require an explicit cache scope from the active profile before any read/write; include that scope in the DB name or every store key; reject cached payloads whose stored scope ≠ the current profile scope; wipe the legacy unscoped DB.

### 🔴 Cache survives profile switch + logout/auth change

`switchToProfile()` changes the cookie without touching the cache; `signOut()` / server logout clears cookies only, not IndexedDB. **Fix:** after `/api/profile/switch`, call an offline-cache `setScope()`/`switchProfile()` before any list render/fallback; on logout and on 401 auth-redirect, `await` a cache `clearAll()` / delete the legacy DB before navigating; start clean on login/auth-epoch change.

### 🟡 Sensitive transcripts persist at rest by default (no opt-in)

Transcripts — including 5 pre-warmed sessions the user never opened — now persist unencrypted in IndexedDB with no setting, no logout wipe. **Fix:** make offline transcript caching **default-off** behind a clear user/admin setting; cache only when enabled; add a "clear offline conversation cache" action; wipe on disable/logout/auth change; document that transcripts live in browser IndexedDB.

### 🟡 Media stripping is incomplete / fails open (`offline-cache.js:144`)

Only strips `image_url` parts from array `content` + truncates `tr.content`; does NOT strip string `data:*;base64,…`, attachment fields, `result`/`output`/`snippet`, or nested tool payloads — and on clone failure it returns the original object for caching. The 2MB cap is enforced before write but only after this incomplete sanitizer. **Fix:** recursively sanitize all cached payloads, regex-strip data-URIs/base64 in any string field, drop known tool-result fields, and skip caching (don't fall back to the original) if sanitize/clone fails.

### 🟡 Stale-serve key collision across request shapes (`workspace.js:143/164`)

All `/api/session?session_id=X…` responses are stored under just `X`, ignoring `messages`/`msg_limit`/`msg_before`/`resolve_model`/`expand_renderable`. An initial tail-window load can overwrite a warmed full transcript, so an offline full-history/export/jump request silently gets a truncated tail. **Fix:** cache by normalized endpoint + profile scope + all response-shaping params (or cache only a canonical full transcript and serve it only for compatible requests); verify `cached.data.session.session_id === requestedSid` before returning.

### 🟡 Version-nuke never fires (`offline-cache.js:35`)

`_appVersion()` reads `window.__WEBUI_VERSION__` / `<html data-version>`, but this page sets neither, so version records become `"unknown"` and survive upgrades (stale-shape risk). **Fix:** inject the real `WEBUI_VERSION` before `offline-cache.js` loads; on mismatch delete the scoped + legacy caches fail-closed.

### 🟡 Unrelated: send disabled while backend reachable (`ui.js:5298`)

`getComposerPrimaryAction()` now treats `navigator.onLine === false` as authoritative, but the existing offline contract probes `/health` first because browser offline reports are unreliable — this can disable send/queue/stop on a working connection. **Fix:** gate disabling on a probe-confirmed offline state; preserve send when `/health` succeeds.

The privacy model (profile-scoping + opt-in + logout wipe) is the thing to nail first — once transcripts at rest are scoped, opt-in, and wiped correctly, the rest is mechanical. Happy to re-review. Thanks again.

kaishi00:
Thanks for the thorough review @nesquena-hermes — all items addressed. Rebased on latest master and force-pushed.

### 🔴 Cross-profile transcript leak — fixed
Rewrote the DB layer: each profile now gets its own IndexedDB database (hermes-offline-cache:PROFILENAME). setScope(profileName) is called at boot (after /api/profile/active), on switchToProfile(), and on the first cache operation. The legacy unscoped DB is deleted on first setScope() call. Profile B can never read Profile A's cached data — they're in separate databases.

### 🔴 Cache survives profile switch + logout — fixed
- switchToProfile() now awaits HermesOfflineCache.setScope(newProfile) after the cookie change (closes old DB handle, opens new profile DB on next op)
- signOut() now awaits HermesOfflineCache.nukeCache() before navigating to login
- boot.js calls setScope(activeProfile) at startup before any cache operations

### 🟡 Sensitive transcripts persist by default — fixed (opt-in, default-off)
localStorage flag hermes-offline-cache-enabled defaults to false. Every cache read/write/warm is a no-op when disabled. Settings > System has a toggle + Clear offline cache now button. Disabling wipes existing cache immediately.

### 🟡 Media stripping incomplete / fails open — fixed (recursive, fail-closed)
Replaced _stripMedia() with recursive _sanitizeValue():
- Regex-strips data:base64 URIs (100+ chars) from ANY string field, not just image_url parts
- Drops known attachment/binary fields (attachment, thumbnail, preview, file_data, blob)
- Truncates result/output/snippet fields beyond 5KB
- Handles image_url object structures (OpenAI format)
- Fail-closed: returns null on clone/sanitize error, cacheSession() skips caching entirely

### 🟡 Stale-serve key collision — fixed
Only canonical full transcripts are cached or served (messages=1 AND no msg_before AND no msg_limit). Partial tail-window loads never touch the cache. On read, verifies cached.data.session.session_id === requestedSid before returning.

### 🟡 Version-nuke never fires — fixed
Added inline script tag setting window.__WEBUI_VERSION__ before offline-cache.js loads. Server already replaces the template token at render time, so _appVersion() gets the real version. On mismatch, the scoped cache is deleted.

### 🟡 Send disabled while backend reachable — fixed
getComposerPrimaryAction() now gates on _offlineVisible (set by the /health probe) instead of navigator.onLine. updateSendBtn() added to online/offline listeners for reactive updates.

---
Test status: All CI shards pass (3 pre-existing failures on clean master, zero new). Lint clean. Diff scope: 10 files, all related (rebased clean from latest master).

kaishi00:
Good catches on the P1s @greptile-apps[bot] — all three fixed in bb64d26e.

**Eviction transaction (offline-cache.js):** Fixed — all delete requests are now queued synchronously via Promise.all before awaiting, preventing TransactionInactiveError on strict IDB implementations.

**Non-GET fallback (workspace.js):** Fixed — both catch-block fallbacks now guard on (opts.method||'GET').toUpperCase()==='GET' before calling _tryOfflineCacheFallback.

**Banner hide + warming on cached data (sessions.js):** Fixed — _applySessionListPayload now checks sessData.__fromOfflineCache before calling hideOfflineBanner/warmRecentSessions, so the amber banner stays visible and no hanging warm-up fetches fire on weak signal.

**P2 ui.js comment:** Already addressed in the previous commit — getComposerPrimaryAction() now gates on _offlineVisible (health-probe confirmed), not _browserReportsOnline(). The stale comment Greptile reviewed was from the pre-rebase push.

**P2 workspace.js condition:** Already addressed in the previous commit — the condition was rewritten to require messages=1 explicitly (AND no msg_before/msg_limit), eliminating the redundant OR logic Greptile flagged.

nesquena-hermes:
Re-reviewed the re-push (head with `setScope`/scoped-DB) — strong work, @kaishi00. You addressed the 🔴 blockers from my first pass head-on:
- **Profile scoping:** each profile now gets its own IndexedDB database, `setScope(profileName)` is required before any op, and the legacy unscoped DB (`hermes-offline-cache`) is wiped via `deleteDatabase`. That closes the cross-profile transcript leak.
- **Opt-in / default-off:** transcript caching is off by default behind a user setting, with the copy explicitly stating it's profile-scoped and wiped on sign-out.
- Scope is set at boot + on profile switch.

This is now in good shape on the privacy model. I'm **not auto-merging it tonight**, though, and that's deliberate: this introduces a whole new **persistent-at-rest caching subsystem** (a service worker + IndexedDB transcript store + a new user-facing setting). Even default-off, a new client-side persistence + service-worker layer is a concept/architecture + security-surface decision for @nesquena, not an autonomous fix-class ship. It's exactly the kind of convention-setting change I should surface rather than merge unattended.

My recommendation to @nesquena: this is worth a real look — the weak-signal PWA problem is real, the privacy model is now sound (scoped + opt-in + wiped on logout), and it's well-tested. Two things to eyeball before merge: (1) the remaining 🟡 media/data-URI sanitizer completeness (does it recursively strip base64/data-URIs and tool-result fields, and skip-rather-than-fall-open on clone failure?), and (2) the service-worker addition's interaction with the existing SSE/PWA connection handling. If you're good with the concept + those two, I'll do the full gate + ship. Parking for your call. (Also currently CONFLICTING — I'll rebase as part of the ship pass once you greenlight.)

nesquena-hermes:
Shelving this for now (not declining) — the maintainer wants to revisit it later rather than move forward at the moment.

To be clear on where it stands: the offline conversation cache is a reasonable idea, but the current trade-off is marginal — it gives a read-only offline *view* of cached sidebar/session data (you still can't send or run the agent offline, since that needs the server), at the cost of ~630 lines of new service-worker caching machinery plus a privacy consideration (cached conversation data living in the SW cache needs to be cleared on logout/profile-switch so it doesn't leak on shared devices).

Marking `hold` + `shelved` so it reads as a deliberate "revisit later," not a miss. If we pick it back up, the shape we'd want is: confirm the cache is cleared on logout/profile-switch, and a clear offline-state UX. Thanks for the contribution @kaishi00 — leaving it open and on hold.

nesquena-hermes:
Taking this off hold — offline PWA resilience is in scope (it's a real reliability gap: the app-shell is cached but API calls aren't, so the PWA shows a blank sidebar/session list on weak or no connection).

Next steps on our side:
1. Rebase onto current `master` (currently CONFLICTING across ~8 static files — `boot.js`, `ui.js`, `sessions.js`, `workspace.js`, `sw.js`, etc. — the surface has moved a lot since this was opened).
2. Full gate (Codex + full suite) + a multi-viewport UX/screenshot pass on the offline degradation states, since this is a visible size:L change on the crown-jewel chat/sidebar surface.

@kaishi00 — if you'd like to rebase it yourself that's welcome and speeds things up; otherwise we'll rebase + carry it through review with your authorship preserved. Removing `hold`, keeping `ux` (it'll go through the screenshot gate).

PR #4439, opened 2026-06-18
fix(#4373): mark process wakeups as synthetic

### LINKED ISSUE #4373: WebUI persists background-process wakeups as visible user messages in the main chat transcript
# WebUI persists background-process wakeups as visible user messages in the main chat transcript

## Summary

When a WebUI background process wakeup fires, the synthetic `[IMPORTANT: Background process ...]` prompt is persisted and rendered as a normal `role: user` message in the active chat. This makes internal runtime notifications look like messages the human typed and pollutes the main transcript.

This is especially visible with background terminal processes using `watch_patterns` or `notify_on_complete`: the wakeup message re-enters the same session as a normal submitted user turn.

## Why this is a problem

- The transcript becomes misleading: internal agent/runtime events appear as human-authored chat messages.
- The visible chat history is harder to read and search.
- Follow-up model context can treat these implementation-detail wakeups as part of the user conversation.
- The behavior resembles child/subagent output being merged into the parent chat, even when the actual source is a process wakeup.

## Reproduction shape

1. In Hermes WebUI, start a normal chat session.
2. Have the agent start a background terminal process with a wakeup condition, for example a long-running process with `watch_patterns` or a background process with `notify_on_complete`.
3. Let the process match the watch pattern or complete.
4. Reopen or inspect the same WebUI chat transcript.

## Actual behavior

The transcript contains a visible user message similar to:

```text
[IMPORTANT: Background process <process_id> matched watch pattern "...".
Command: ...
Matched output:
...]
```

or:

```text
[IMPORTANT: Background process <process_id> completed (exit code ...).
Command: ...
Output:
...]
```

This row is persisted as a normal `role: user` message and appears inline with human-authored turns.

## Expected behavior

Synthetic process wakeups should not be displayed as normal human user messages.

Possible acceptable behavior:

- Persist them as hidden/internal/system events used only for agent wakeup/context.
- Render them in a separate status/runtime event lane rather than the main chat transcript.
- Mark them with metadata such as `source: process_wakeup`, `synthetic: true`, or `hidden: true`, and filter them from the human-facing message list.
- If the agent needs the wakeup text as input, keep that input available to the agent without presenting it as if the user submitted it.

## Suspected area

The WebUI process-wakeup path appears to intentionally start a server-side agent turn with a synthetic prompt:

- `api/background_process.py` drains process completion/watch events.
- `api/routes.py:start_session_turn(...)` starts a server-side turn for the session using the synthetic wakeup prompt.
- The shared start-run path then persists it like a normal user-submitted turn.

Relevant code comments indicate this is intended to mirror CLI/gateway self-wake behavior, but WebUI needs a distinction between agent input and visible human chat rows.

## Suggested fix direction

When `source == "process_wakeup"` or equivalent:

1. Preserve the synthetic wakeup prompt for the agent worker input.
2. Persist the event with explicit metadata identifying it as synthetic/internal.
3. Exclude it from the visible chat transcript, or render it as a compact runtime notification separate from user/assistant messages.
4. Ensure transcript hydration, sidecar repair, and session search do not reclassify it as a normal user message.

## Acceptance criteria

- Background process wakeups still resume/wake the agent correctly, including closed-tab/server-side wakeup cases.
- The main WebUI chat no longer shows `[IMPORTANT: Background process ...]` as a human user bubble.
- Stored session data distinguishes human-authored user messages from synthetic process wakeups.
- Existing recovery/hydration paths do not reintroduce synthetic wakeups into visible messages.
- Add regression coverage for a process wakeup turn verifying:
  - agent receives/handles the wakeup, and
  - visible session messages exclude or separately classify the synthetic prompt.

rodboev:
## Thinking Path

- The branch already preserved `pending_user_source` through chat start and hid the live pending stub, but the settled transcript merge still rebuilt the current user turn without `_source`.
- That meant the fix held only during the pending window. As soon as the turn completed, the synthetic wakeup prompt was materialized as an ordinary user row and became visible again on reload.
- The right follow-up is to thread the same source through the final merge helper, reuse it on the gateway-backed settle path, and clear `pending_user_source` anywhere pending turn state is torn down.

## What Changed

- `api/streaming.py`: `_merge_display_messages_after_agent_result()` now accepts `source`, stamps `_source` on inserted or normalized current-user rows when the source is synthetic, and all settled merge call sites now pass `pending_user_source`; pending teardown paths also clear `pending_user_source` alongside the other pending fields.
- `api/gateway_chat.py`: gateway-backed completion now stamps the settled `user_msg` with the pending source, passes that source through the same merge helper, and clears `pending_user_source` when pending gateway state is torn down.
- `api/routes.py`: the direct transcript merge path also threads `pending_user_source` so every visible-merge surface uses the same settled-turn contract.
- `tests/test_process_wakeup_synthetic.py`: added settled-path regressions for both the inserted-user-turn and echoed-user-turn cases, plus the normal `webui` control case.

## Why It Matters

Process wakeup prompts now stay hidden after the turn settles, not just while the turn is pending. The same source distinction also survives the gateway-backed completion path, so synthetic wakeups do not reappear as normal user bubbles after reload.

## Verification

- `pytest tests/test_process_wakeup_synthetic.py tests/test_wakeup_model_resolve_hang.py -v --timeout=60`

## Risks / Follow-ups

- The wakeup prompt still exists in stored transcript data with `_source="process_wakeup"`; this is intentional because the agent needs the prompt and the UI now has the metadata to suppress it.
- Any future transcript-finalization path that bypasses `_merge_display_messages_after_agent_result()` will need to thread source explicitly too; the current settled merge call sites are covered in this PR.

## Upstream

Closes #4373.

## Model Used

GPT-5 via Codex CLI

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -440,6 +440,7 @@ def _clear_gateway_pending_state(session: Any, stream_id: str) -> None:
     session.pending_user_message = None
     session.pending_attachments = None
     session.pending_started_at = None
+    session.pending_user_source = None
     session.save()
 
 
@@ -713,6 +714,9 @@ def _run_gateway_chat_streaming(
             # role/content ordering instead of turn order.
             assistant_ts = now + 0.000001
             user_msg = {"role": "user", "content": str(msg_text or ""), "timestamp": now}
+            pending_source = getattr(s, "pending_user_source", None) or "webui"
+            if pending_source != "webui":
+                user_msg["_source"] = pending_source
             if attachments:
                 user_msg["attachments"] = list(attachments)
             assistant_msg = {"role": "assistant", "content": assistant_text, "timestamp": assistant_ts}
@@ -744,6 +748,7 @@ def _run_gateway_chat_streaming(
                     previous_context,
                     s.context_messages,
                     str(msg_text or ""),
+                    source=pending_source,
                 )
             except Exception:
                 logger.debug("Failed to merge gateway display transcript", exc_info=True)
@@ -760,6 +765,7 @@ def _run_gateway_chat_streaming(
             s.pending_user_message = None
             s.pending_attachments = None
             s.pending_started_at = None
+            s.pending_user_source = None
             s.workspace = str(workspace)
             s.model = model
             s.model_provider = model_provider
--- a/api/models.py
+++ b/api/models.py
@@ -1896,6 +1896,7 @@ def _apply_core_sync_or_error_marker(
             session.pending_user_message = None
             session.pending_attachments = []
             session.pending_started_at = None
+            session.pending_user_source = None
             session.save(touch_updated_at=touch_updated_at)
             logger.info(
                 "Session %s: cleared stale pending state for completed stream %s without error marker",
@@ -1922,6 +1923,7 @@ def _apply_core_sync_or_error_marker(
         session.pending_user_message = None
         session.pending_attachments = []
         session.pending_started_at = None
+        session.pending_user_source = None
         session.messages.append(
             _build_recovery_marker_with_retry_hook(
                 recovered_output=recovered_output,
@@ -1976,6 +1978,7 @@ def _apply_core_sync_or_error_marker(
             session.pending_user_message = None
             session.pending_attachments = []
             session.pending_started_at = None
+            session.pending_user_source = None
             if recovered_output:
                 session.messages.append(
                     _interrupted_recovery_marker(
@@ -2023,6 +2026,7 @@ def _apply_core_sync_or_error_marker(
     session.pending_user_message = None
     session.pending_attachments = []
     session.pending_started_at = None
+    session.pending_user_source = None
     session.messages.append(
         _build_recovery_marker_with_retry_hook(
             recovered_output=recovered_output,
--- a/api/models.py
+++ b/api/models.py
@@ -416,6 +416,9 @@ def _append_recovered_pending_turn(session, *, timestamp: int | None = None) -> 
         'timestamp': recovered_ts,
         '_recovered': True,
     }
+    pending_source = getattr(session, 'pending_user_source', None)
+    if pending_source and pending_source != 'webui':
+        recovered['_source'] = pending_source
     if session.pending_attachments:
         recovered['attachments'] = list(session.pending_attachments)
     session.messages.append(recovered)
@@ -621,6 +624,7 @@ class Session:
                  pending_user_message: str=None,
                  pending_attachments=None,
                  pending_started_at=None,
+                 pending_user_source: str=None,
                  context_messages=None,
                  compression_anchor_visible_idx=None,
                  compression_anchor_message_key=None,
@@ -668,6 +672,7 @@ class Session:
         self.pending_user_message = pending_user_message
         self.pending_attachments = pending_attachments or []
         self.pending_started_at = pending_started_at
+        self.pending_user_source = pending_user_source
         self.context_messages = context_messages if isinstance(context_messages, list) else []
         self.compression_anchor_visible_idx = compression_anchor_visible_idx
         self.compression_anchor_message_key = compression_anchor_message_key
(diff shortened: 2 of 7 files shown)


nesquena-hermes (changes requested):
Thanks @rodboev — tagging synthetic process-wakeup turns so they don't render as user messages is a clean idea, and the *pending/live* path here is solid: `source` is threaded completely from `start_session_turn` through `_start_run` → `_start_chat_stream_for_session` → `_prepare_chat_start_session_for_stream`, the `pending_user_source` field round-trips through `save()`/`METADATA_FIELDS`/the `/api/session` payload and is cleared in all four `_clear_stale_stream_state` branches, and a `webui` turn correctly never gets a `_source` tag. I ran the full review gate (Codex + Opus + suite + browser-smoke) and it's close, but there's one gap that means the fix doesn't hold on the path that matters most.

### (must-fix) The settled (post-completion) user message loses the `_source` tag

The frontend suppressor (`static/ui.js` `_messageIsRenderable`) only hides messages that **already carry** `_source === "process_wakeup"`. But on a **normal successful turn** (the common case), the final transcript user row is synthesized by `_merge_display_messages_after_agent_result(...)` (`api/streaming.py:~4567`):

```js
current_user_msg = {'role': 'user', 'content': msg_text}   // no _source
```

That helper is called from the completion merge (`streaming.py:~7187`) **without** the source, and the subsequent pending-state clear (`streaming.py:~7715`) drops `pending_user_message`/`pending_started_at` without stamping the settled message. So after the turn completes and the session is reloaded, the synthetic wakeup prompt is persisted **untagged** and renders as a user bubble — exactly what the PR sets out to prevent. The suppression only works during the live/pending window, not after settlement.

The gateway-backed path has the same gap: `api/gateway_chat.py:~715` builds `user_msg = {"role": "user", ...}`, merges at `~742`, and clears pending fields at `~759` without `_source` / `pending_user_source`.

**Fix (thread the source into the final materialization, mirroring what you already did for the pending path):**
1. Add `source: str = "webui"` to `_merge_display_messages_after_agent_result(...)` and stamp `current_user_msg["_source"] = source` (and the matched/normalized existing user row) when `source != "webui"`.
2. Pass `getattr(s, "pending_user_source", None) or "webui"` into that helper from the completion call sites (`streaming.py:~7187`, and the sibling finalize paths around `~7479` / `~8459`).
3. In `gateway_chat.py`, stamp `user_msg["_source"] = pending_source` when `getattr(s, "pending_user_source", None)` is set and not `"webui"`, pass the same source into its merge, and clear `s.pending_user_source = None` with the other pending fields.
4. Add a regression test that runs a wakeup turn to **completion** (not just pending) and asserts the persisted/reloaded transcript user row carries `_source == "process_wakeup"` (and that a webui turn does not).

Once the settled-message path carries the tag, I'm happy to re-gate and ship — the pending-path plumbing is already correct, so this is finishing the same thread through the completion merge. Happy to do the rebase for you at re-gate.

rodboev:
Pushed `2053825d`.

The follow-up threads `pending_user_source` through the settled transcript path, not just the pending stub:

1. `_merge_display_messages_after_agent_result()` now accepts `source` and stamps `_source` on both the synthesized current-user row and the echoed-user-row case.
2. The normal completion, retry/self-heal completion, gateway completion, and the direct merge path in `api/routes.py` now pass `pending_user_source`, so the synthetic tag survives settlement and reload instead of disappearing after the turn finishes.
3. Pending teardown now clears `pending_user_source` with the other pending fields.

Focused local validation:
- `pytest tests/test_process_wakeup_synthetic.py tests/test_wakeup_model_resolve_hang.py -v --timeout=60`

rodboev:
Pushed `b0979637` for the cancel and recovery follow-up.

1. The cancel-path recovery now stamps `_source` from `pending_user_source` and keeps the original pending timestamp when it synthesizes the recovered user turn.
2. `_materialize_pending_user_turn_before_error()` now carries non-`webui` sources through error recovery as well, so synthetic wakeup turns stay hidden outside the settled path.
3. `_apply_core_sync_or_error_marker()` and the cancel teardown now clear `pending_user_source` with the rest of the pending state, so stale synthetic-source metadata is not left behind in persisted sessions.

Validation:
- `pytest tests/test_process_wakeup_synthetic.py tests/test_sprint51.py -v --timeout=60` -> `20 passed`

nesquena-hermes:
Shipped in **v0.51.503** (deployed). Thanks @rodboev — your re-push cleanly closed the settled-message `_source` gap across every materialization path including both gateway sub-paths. Co-authored-by credit is in the release commit + CHANGELOG. Closed via the release stage branch.

PR #4441, opened 2026-06-18
test: cover archived webhook session projections

santastabber:
## Thinking Path

- WebUI projects external Hermes sessions from `state.db` into the Sessions sidebar.
- Archived state for those projected rows is UI-owned sidecar metadata, not the raw `state.db.sessions` row.
- Cron/tool/API projections already have regression coverage for this class, but webhook rows use the same state projection path and need an explicit guard.
- This PR adds webhook-specific coverage so an archived webhook sidecar cannot be re-projected into the sidebar as `archived: false`.

## What Changed

- Broadened the existing archived state.db projection regression module beyond cron-only wording.
- Added a model-layer regression for `source='webhook'` rows proving `_load_cli_sessions_uncached()` preserves sidecar `title` and `archived=True`.
- Added a sidebar payload regression proving `_build_session_list_cache_payload()` carries the archived webhook projection through to `/api/sessions` data.

## Why It Matters

Without this coverage, a future change could preserve archived cron rows while accidentally letting webhook sessions reappear in the sidebar after archive. The added tests pin the shared state-layer invariant for webhook sessions explicitly.

## Verification

- `./scripts/test.sh tests/test_issue4385_cron_archive_reappears.py -q` — 6 passed
- `./scripts/test.sh tests/test_issue4385_cron_archive_reappears.py tests/test_gateway_sync.py tests/test_issue3586_cli_session_source_label.py tests/test_issue3930_source_filter_pushdown.py -q` — 80 passed
- `python3 -m py_compile api/agent_sessions.py api/models.py api/routes.py tests/test_issue4385_cron_archive_reappears.py`
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master` — no new violations
- Independent agent review reran the focused test file and compile check; test content passed and the only concern was resolved by committing the branch before publishing.

## Risks / Follow-ups

- Test-only PR; no runtime code changes.
- This relies on the existing sidecar metadata projection helper remaining the source of truth for UI-owned archived state.

## Contract Routing

Task type: runtime/sidebar metadata regression coverage.
Touched areas: state.db-projected session rows and sidebar payload assembly tests.
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
Scope boundaries: no user-facing UI/UX change and no new behavior contract; this pins an existing state-layer invariant for webhook rows.
Evidence needed before claiming done: focused regression tests and adjacent source-normalization/sidebar tests.

## Model Used

OpenRouter `openai/gpt-5.5` with Hermes Agent tool use and an independent subagent review pass.

### DIFF
--- a/tests/test_issue4385_cron_archive_reappears.py
+++ b/tests/test_issue4385_cron_archive_reappears.py
@@ -1,4 +1,4 @@
-"""Regression coverage for #4385: archived cron sessions reappearing."""
+"""Regression coverage for archived state.db-projected sessions reappearing."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Read the diff against `origin/master` plus the production paths it pins: `_load_cli_sessions_uncached` (`api/models.py:3896`), `_state_projection_sidecar_metadata` (`api/models.py`), and `normalize_agent_session_source` (`api/agent_sessions.py:40`). Test-only, no runtime change, and both new tests target the real code paths correctly. I traced each non-obvious assertion to its source — two are subtly coupled to fallback behavior and worth a comment so a future refactor doesn't silently break them.

## Both assertions are correct, but for non-obvious reasons

**`source_label == "Webhook"`** is *not* backed by an explicit mapping. `SOURCE_LABELS` (`api/agent_sessions.py:23-37`) has no `"webhook"` key, so the value comes from the title-case fallback:

```python
label = SOURCE_LABELS.get(raw)
if not label:
    label = raw.replace('_', ' ').title() if raw != 'unknown' else 'Agent'
```

So `"webhook".title()` → `"Webhook"`. That's fine, but if anyone later adds `'webhook': 'Webhook Inbound'` (or similar) to `SOURCE_LABELS`, this assertion breaks even though nothing about the *archive* invariant changed. A one-line comment in the test noting "Webhook label is the title-case fallback, not an explicit SOURCE_LABELS entry" would save the next person a confused debugging session.

**`is_cli_session is False`** is the more interesting one. `is_cli_session_row` (`api/agent_sessions.py:165`) does NOT explicitly exclude webhook — webhook is absent from both `MESSAGING_SOURCES` and the `non_cli_sources` set:

```python
non_cli_sources = MESSAGING_SOURCES | {"cron", "tool", "api", "api_server"}
if {source, source_tag, raw_source, source_name, source_label} & non_cli_sources:
    return False
```

A webhook row (`session_source='other'`) passes through all the early returns and lands on the conservative legacy fallback at the end, which returns False only because the raw state.db row carries no `is_cli_session` flag. So the assertion passes by omission, not by classification. The corollary worth knowing: a webhook row that *did* arrive with `is_cli_session=True` and a default-looking title would be misclassified as CLI, because webhook isn't in `non_cli_sources`. If that's a real concern for the sidebar, the durable fix is adding `"webhook"` to `non_cli_sources` (agent side of the contract), and this test would then pin the explicit exclusion rather than the fallback.

## Minor: greptile's project_id note is valid

The cron sibling test asserts `project_id`, and the webhook test omits it. Since `project_id` is gated by `is_cron_session(sid, _source)` (`api/models.py:3982`) — False for webhook → `None` — adding `assert row["project_id"] is None` would pin that webhook rows never get bucketed into the cron project chip. Cheap and matches the existing cron test's rigor.

## Verdict

Mergeable as-is; the invariant it pins is real and the mock setup mirrors the passing cron tests. The two notes above are about making the test *say what it means* (fallback-derived label, classification-by-omission) so a later source-handling change can't quietly invalidate it. Nothing blocking.

Issue #3103, opened 2026-05-28
bug: Connection: close on SSE endpoints causes EventSource reconnect loop (introduced in 598fd4ff)


andrewkangkr:
## Summary

Commit `598fd4ff` (`perf(http): enable HTTP/1.1 keep-alive`, 2026-05-24) added `Connection: close` to SSE response headers in `_handle_session_events_stream` and `_handle_gateway_sse_stream`. This causes the browser's `EventSource` to enter a reconnect loop, which in turn triggers UI re-renders and scroll-to-bottom on every reconnect.

## Root Cause

`598fd4ff` intended to add `Connection: close` to SSE/streaming endpoints to keep HTTP/1.1 message boundaries unambiguous. However, the browser interprets `Connection: close` on an SSE response as a signal that the connection will close after the response completes. Since the `EventSource` API automatically reconnects on any disconnect/error, this creates a loop:

1. Browser opens `EventSource` → receives `Connection: close` header
2. EventSource fires `onerror` → `_sessionEventsNeedsRefreshOnOpen = true`
3. Reconnect after ~4–30s (exponential backoff starting at 3750ms)
4. `onopen` fires → `refreshSessionList('reconnect')` → full sidebar re-render
5. Repeat

Server-side confirmation: `~/.hermes/webui-launchd-stdout.log` shows repeated `GET /api/sessions/events` requests completing in 0–1ms (only headers sent, then immediate loop):

```
{"ts": "2026-05-28T18:49:09Z", "path": "/api/sessions/events", "status": 200, "ms": 0.1}
{"ts": "2026-05-28T18:49:20Z", "path": "/api/sessions/events", "status": 200, "ms": 0.1}
{"ts": "2026-05-28T18:49:27Z", "path": "/api/sessions/events", "status": 200, "ms": 0.1}
{"ts": "2026-05-28T18:49:31Z", "path": "/api/sessions/events", "status": 200, "ms": 0.7}
```

Reconnects every 4–15 seconds during active use.

## Affected Files

- `api/routes.py` — `_handle_session_events_stream` (line ~7367)
- `api/routes.py` — `_handle_gateway_sse_stream` (line ~7334)

## Fix

Remove `Connection: close` from both SSE handlers. SSE connections are long-lived by design; `Connection: close` is not needed since `BaseHTTPServer` handles socket lifecycle through the handler loop naturally.

```python
# _handle_session_events_stream and _handle_gateway_sse_stream
# Remove this line:
[...]

PR #4447, opened 2026-06-18
SSE: opt-in chunked transfer-encoding so streams survive buffering pr…

tomtong2015:
…oxies

The stdlib HTTP server sends SSE bodies with no framing (no Content-Length, no Transfer-Encoding). Tornado-based reverse proxies — jupyter-server-proxy in JupyterHub/REANA-style deployments — read such responses until-close and buffer the ENTIRE body, so browsers receive zero events while a stream is live: chat looks frozen, the client watchdog fires "Connection interrupted", and transcripts only appear after reconnect + journal replay.

api/sse_chunked.py adds end_sse_headers(handler): when HERMES_WEBUI_SSE_CHUNKED is set it emits Transfer-Encoding: chunked and wraps wfile so every write is one HTTP/1.1 chunk; otherwise it is exactly handler.end_headers(). All 10 SSE endpoints now call it. The flag is off by default so the wire format is unchanged for direct deployments — only deployments behind a buffering proxy opt in. It sits alongside the existing X-Accel-Buffering: no hint on the same handlers.

test_session_sse_stream_unsubscribes_on_header_write_failure asserts (by source inspection) that the session SSE handler finishes its headers inside the try/finally that unsubscribes; updated to match the end_sse_headers() call.

Verified: with the flag on, a tornado streaming client receives each 5s keepalive on time (was: one buffered blob at connection teardown); the chunked body decodes back to the original SSE bytes. tests/test_sse_chunked.py covers both flag states and the round-trip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,6 +342,7 @@ Full list of environment variables:
 | `HERMES_WEBUI_DEFAULT_MODEL` | *(provider default)* | Optional model override; leave unset to use the active Hermes provider default |
 | `HERMES_WEBUI_PASSWORD` | *(unset)* | Set to enable password authentication |
 | `HERMES_WEBUI_CSP_CONNECT_EXTRA` | *(unset)* | Optional space-separated `http(s)://` or `ws(s)://` origins to append to the report-only CSP `connect-src` directive for reverse-proxy or tunnel deployments |
+| `HERMES_WEBUI_SSE_CHUNKED` | *(unset)* | Set truthy (`1`/`true`/`yes`/`on`) to send SSE with `Transfer-Encoding: chunked`. Needed behind buffering reverse proxies (e.g. `jupyter-server-proxy`) that otherwise buffer the whole stream; harmless but unnecessary for directly-served deployments |
 | `HERMES_WEBUI_EXTENSION_DIR` | *(unset)* | Optional local directory served at `/extensions/`; must point to an existing directory before extension injection is enabled |
 | `HERMES_WEBUI_EXTENSION_SCRIPT_URLS` | *(unset)* | Optional comma-separated same-origin script URLs to inject; see [WebUI Extensions](docs/EXTENSIONS.md) |
 | `HERMES_WEBUI_EXTENSION_STYLESHEET_URLS` | *(unset)* | Optional comma-separated same-origin stylesheet URLs to inject; see [WebUI Extensions](docs/EXTENSIONS.md) |
--- a/api/kanban_bridge.py
+++ b/api/kanban_bridge.py
@@ -14,6 +14,7 @@ Supported operations:
 from __future__ import annotations
 
 import json
+from api.sse_chunked import end_sse_headers
 import time
 from dataclasses import asdict, is_dataclass
 from urllib.parse import parse_qs, unquote
@@ -1065,7 +1066,7 @@ def _handle_events_sse_stream(handler, parsed):
     handler.send_header("Cache-Control", "no-cache")
     handler.send_header("X-Accel-Buffering", "no")
     handler.send_header("Connection", "close")
-    handler.end_headers()
+    end_sse_headers(handler)
 
     # Send an initial frame so the client knows the connection is open
     # and learns the current cursor (in case the server already had a
--- a/api/routes.py
+++ b/api/routes.py
@@ -8,6 +8,7 @@ import copy
 import io
 import gzip
 import json
+from api.sse_chunked import end_sse_headers
 import logging
 import os
 import queue
@@ -11133,7 +11134,7 @@ def _stream_runner_run_events(handler, run_id: str, cursor: str | None = None) -
     handler.send_header("Cache-Control", "no-cache")
     handler.send_header("X-Accel-Buffering", "no")
     handler.send_header("Connection", "close")
-    handler.end_headers()
+    end_sse_headers(handler)
     cursor_value = cursor
     try:
         while True:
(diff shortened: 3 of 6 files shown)


nesquena-hermes:
Read the full diff plus `api/sse_chunked.py`, the `routes.py`/`kanban_bridge.py` call sites, and the new `tests/test_sse_chunked.py` against `origin/master`. This is a careful, well-scoped change and the opt-in design is right. A few verification notes and one robustness point.

## What I verified

- **Chunked framing is legal here.** `server.py:231` sets `protocol_version = "HTTP/1.1"`, so `Transfer-Encoding: chunked` is valid on these responses (it would be ignored/illegal under HTTP/1.0). Good.
- **Coverage is complete.** There are exactly 10 `text/event-stream` content-type headers across `api/routes.py` + `api/kanban_bridge.py`, and exactly 10 `end_sse_headers(handler)` call sites now. Every remaining bare `handler.end_headers()` belongs to a non-SSE (JSON/static) endpoint. Nothing was missed.
- **The wrapper actually catches every event.** `_sse()` writes through `handler.wfile.write(...)` then `handler.wfile.flush()` (`api/streaming.py:4918-4919`), and `_ChunkedSSEWriter.flush` delegates to `self._raw.flush()`, so each event is framed and flushed as one chunk. No double-encoding risk: gzip is confined to the static-asset path (`api/routes.py:10812-10826`), not SSE.
- **Default-off is byte-identical.** `end_sse_headers` falls straight through to `handler.end_headers()` when the flag is unset, and the test asserts the verbatim body. The source-inspection test in `test_session_channel_option_x.py:912` was correctly updated to look for `end_sse_headers(handler)` inside the try/finally.

## One robustness point: the missing terminating chunk

`_ChunkedSSEWriter` never emits the terminating `0\r\n\r\n`, and the test's `_dechunk` helper openly documents this ("no trailing zero-length terminator"). For the handlers that send `Connection: close` this is harmless — socket close ends the message. But several SSE handlers deliberately omit `Connection: close` to avoid the #3103 EventSource reconnect storm (e.g. `_handle_gateway_sse_stream`, `_handle_session_events_stream`, `_handle_session_sse_stream`, kanban's `_handle_events_sse_stream`). On those, a keep-alive connection carrying a chunked body that never terminates is technically a malformed message — a strict intermediary could wait for the terminator or log a protocol error.

In practice the risk is low: these handlers loop until `_CLIENT_DISCONNECT_ERRORS` and the socket dies, so there's rarely a graceful close where the terminator *could* be written, and jupyter-server-proxy (the target) forwards-and-closes regardless. So I'd treat this as a polish item, not a blocker. If you want to close the gap, a `close()` on `_ChunkedSSEWriter` that emits `b"0\r\n\r\n"` (called from a handler `finally:` where the socket is still alive) would make the framing strictly conformant.

## Suggestion

Worth a one-line note in the README/env-var docs that this is HTTP/1.1-only and meant for buffering Tornado-style proxies, so operators don't enable it blindly behind nginx (where `X-Accel-Buffering: no`, already sent here, is the correct lever instead). Otherwise this looks mergeable.

nesquena-hermes (changes requested):
Thanks @tomtong2015 — this is a well-targeted fix for a real deployment problem (SSE buffering behind jupyter-server-proxy/Tornado), and making it opt-in via `HERMES_WEBUI_SSE_CHUNKED` so the default wire format is untouched is exactly the right call. I ran the full gate (Codex + Opus, both with local HTTP/1.1 probes). The good news both confirmed: per-event flushing works (each `write()` emits a complete `<size>\r\n<payload>\r\n` chunk, so an incremental decoder releases every event immediately — the core goal is achieved with no data loss), all 10 `text/event-stream` handlers are swapped consistently, no non-SSE handler is wrongly swapped, the `write()` return-value follows the `BufferedWriter` contract, and the `if not data: return 0` guard correctly avoids a premature terminator. Nice.

But there are two CORE protocol issues that block it as-is — both stem from the chunked stream never being properly closed:

### (must-fix 1) No terminating `0\r\n\r\n` chunk
The wrapper never emits the terminating zero-length chunk on stream end. Codex's `http.client` probe raises `IncompleteRead` on a chunked SSE response that ends via `Connection: close`; strict buffering proxies/clients can treat a *normal* stream completion as a **failed/incomplete** response. A chunked body that's never terminated is, per spec, an error.

### (must-fix 2) `handler.wfile` is replaced and never restored → keep-alive response corruption
`end_sse_headers` wraps `handler.wfile` in `_ChunkedSSEWriter` but never restores it. `server.py:231` runs `protocol_version = "HTTP/1.1"`, so the handler instance + its wrapped `wfile` persist across requests on a kept-alive connection. Three SSE handlers **deliberately omit `Connection: close`** (per #3103, to avoid EventSource reconnect storms) — routes.py:11524/11559/12882. With the flag on, those advertise `Transfer-Encoding: chunked` but send **neither a terminator nor `Connection: close`** — an unterminated chunked body that violates the framing invariant `server.py:229-230` documents. If such a handler returns cleanly and the socket is reused, the next response's headers get flushed *through* the persisted `_ChunkedSSEWriter` → header bytes are chunk-framed (Codex's pipelined probe saw `90\r\nHTTP/1.1 200 OK...`) → corrupt response. (Opus notes a conformant client won't reuse a connection whose response never terminated, so it's a latent trap rather than an everyday bug — but combined with #1 it's the same root cause and worth closing.)

Also: the `sse_chunked.py` docstring claims "Chunked + `Connection: close` is legal … on the responses these handlers emit" — inaccurate for 3 of the 10 (they don't send `Connection: close`).

### Fix (both gates converge on this)
Give `_ChunkedSSEWriter` a `finalize()`/`close()` that writes `b"0\r\n\r\n"` exactly once, and call it from each chunked SSE handler's `finally:` (and/or restore `handler.wfile = <raw>` there). That delivers clean framing on all 10 handlers and removes the keep-alive/double-wrap trap entirely. Then correct the docstring. A test that asserts the terminating chunk is emitted on normal teardown (and that wfile is restored) would lock it in.

The chat token path itself is safe (it uses `Connection: close`), so this is about the 3 keep-alive SSE handlers + clean termination. Happy to re-gate as soon as the finalizer lands — the framing approach is otherwise solid.

Issue #3368, opened 2026-06-02
Bug: model selection logic (using a substring) prevents selection of some models


garyd9:
No idea what version, but the thing updated itself about 20 minutes before I'm authoring this ticket... ;)

I subscribe to nous portal and I'm finding it impossible to select some models, apparently because their name is a substring of other model names.)

Example:  

Anytime I try to select the model "mimo-v2.5" using "/model mimo-v2.5", mimo-v2.5-pro is selected instead.  According to my agent, this is due to the loop in commands.js around line 351:
```
for(const opt of sel.options){
  if(opt.value.toLowerCase().includes(q)||opt.textContent.toLowerCase().includes(q)){
    match=opt.value;break;
  }
}
```
From my agent: 
> It uses .includes() — which is a substring match. So "mimo-v2.5" matches both:
>
> xiaomi/mimo-v2.5-pro — because "mimo-v2.5-pro" contains "mimo-v2.5"
> xiaomi/mimo-v2.5 — because it's an exact match
> 
> Whichever option appears first in the dropdown's option list wins (that break). If mimo-v2.5-pro is alphabetically or structurally before mimo-v2.5, it gets selected.
> The earlier _findModelInDropdown() function (in ui.js) is smarter — it uses exact normalized matching — but the fallback path overrides it when the > first path returns null (e.g. if the normalized provider prefix doesn't match).

This also from my agent (unverified by me):
>  If you want to fix the WebUI itself, the patch is straightforward — swap includes() for startsWith() on line 352, then reverse-sort the option scan so shorter/more-precise matches (the non-pro version) win over longer ones.

yutaotie:
<img width="486" height="204" alt="Image" src="https://github.com/user-attachments/assets/cefc62e3-7a13-4f8d-a3ef-9254a5993019" />  the same  problems, model  custom  be replaced

nesquena-hermes:
Thanks for the precise report — your agent's read of `commands.js` is correct, and there are actually two layers here.

## What just changed

A closely related collision bug, #3360 ("Model picker snaps to wrong model when multiple multi-slash IDs share the same base name"), merged a few hours after you filed this — commit `d06776a4`, tagged **v0.51.210**. That fix reordered `_findModelInDropdown()` in `static/ui.js` so an *exact* option value wins before the normalized/provider-aware match:

```js
// static/ui.js:1071  (0. Exact match — highest priority, #3360)
if(opts.includes(modelId)){
[...]

PR #4448, opened 2026-06-18
frontend: subpath-safe fetch() in /model command

tomtong2015:
cmdModel() had the only two root-absolute fetch('/api/...') calls left in the frontend. Under a reverse proxy that mounts the app at a subpath (e.g. jupyter-server-proxy at /proxy/<port>/), a root-absolute path escapes the mount and 404s, so /model can't load the catalog or update the session. Resolve both against document.baseURI (falling back to location.href) like every other fetch in the frontend already does.

### DIFF
--- a/static/commands.js
+++ b/static/commands.js
@@ -547,7 +547,7 @@ async function cmdModel(args){
   // CLI resolves against the full catalog, so /model must too. (#3368)
   let modelsData=null;
   try {
-    const resp=await fetch('/api/models');
+    const resp=await fetch(new URL('api/models',document.baseURI||location.href).href);
     if(resp.ok){
       modelsData=await resp.json();
       const aliases=modelsData.aliases||{};
@@ -587,7 +587,7 @@ async function cmdModel(args){
     if(!match && !versionedNoSnap && S&&S.session&&S.session.session_id){
       const provider=q.slice(0,q.indexOf('/'));
       try{
-        const resp=await fetch('/api/session/update',{
+        const resp=await fetch(new URL('api/session/update',document.baseURI||location.href).href,{
           method:'POST',
           headers:{'Content-Type':'application/json'},
           body:JSON.stringify({


nesquena-hermes:
Thanks for this — it's a clean, well-scoped fix.

## Summary

Reading `static/commands.js:550` and `:590` on `origin/master` plus the diff at this PR, the claim checks out: `cmdModel()` held the last two root-absolute `fetch('/api/...')` calls in the frontend. Under a reverse proxy that mounts the app at a subpath (jupyter-server-proxy at `/proxy/<port>/`, REANA, etc.), a leading-slash path resolves against the origin root and escapes the mount, so `/model` 404s on both the catalog fetch and the cross-provider `session/update`.

## Code reference

The two changed lines now match the convention every other fetch already uses:

```js
const resp=await fetch(new URL('api/models',document.baseURI||location.href).href);
...
const resp=await fetch(new URL('api/session/update',document.baseURI||location.href).href,{
```

This is byte-for-byte the same shape as `static/ui.js:1885` (`new URL('api/models',document.baseURI||location.href)`) and `static/ui.js:1855` (`new URL('api/session/update',...)`), so `commands.js` is just catching up to the established pattern.

## Verification

I confirmed against `origin/master`:

- `grep "fetch('/api"` across `static/` returns exactly those two `commands.js` lines and nothing else — so this closes the gap completely.
- The relative-URL form resolves correctly: with `document.baseURI = https://host/proxy/8787/`, `new URL('api/models', base)` yields `https://host/proxy/8787/api/models`. The `location.href` fallback preserves direct-deployment behavior.

## Recommendation

Looks good to merge. Two optional follow-ups, neither blocking:

1. A quick lint/CI guard (a `grep -n "fetch(['\"]/api" static/` check) would prevent the next root-absolute fetch from creeping back in, since this is the second time the pattern has had to be swept up.
2. No test is included; given it's a one-line URL-construction change matching existing call sites, a unit test is probably overkill, but a comment noting the subpath rationale (like the `#3368` note just above) wouldn't hurt.

Issue #2905, opened 2026-05-25
刚刚更新完我的webui对话、标签以及之前保存的设置全部丢失


JayC-L:
<img width="351" height="614" alt="Image" src="https://github.com/user-attachments/assets/13c02885-0dc5-44ec-9c8b-adaf1065b7b6" />

nesquena-hermes:
Thanks for reporting @JayC-L. The screenshot shows a successful update notification but it's not clear from the image alone what data was lost — could you share a bit more:

1. **What you saw before the update** vs what's now missing (specific examples: "my pinned sessions list is empty", "the workspace I had selected is gone", "Settings → API providers are blank")
2. **Which version you updated from / to** (look at the bottom of Settings or run `hermes --version` in the terminal where you launched WebUI)
3. **Whether your `~/.hermes/` directory still contains your sessions** — `ls ~/.hermes/webui/sessions/` should show your conversation files even if the sidebar doesn't render them
4. **Profile used** — if you were on a named profile (not `default`), did profile selection reset?

The most likely causes are:
- **Profile context reset** (Settings → Profile dropdown defaulted back to `default` after the update — your sessions are still there, just on the other profile)
- **State-dir migration** (the update reorganized `~/.hermes/` and your sidebar is reading from the new location while your data is still in the old one)
- **Genuine data loss** (rarer, but real — would need the agent log around update time)

Marking `bug` + `needinfo`, **Critical Bug Fix** milestone pending the details above. Don't restart WebUI more times than necessary — the missing data is almost certainly still on disk.

---

感谢您的反馈。请提供以下信息以便我们诊断：(1) 您从哪个版本更新到哪个版本？(2) 具体丢失了什么 (会话列表为空？设置归零？固定标签消失？)？(3) `~/.hermes/webui/sessions/` 目录里还有您的会话文件吗？最可能的原因是配置文件 (profile) 切换回了 `default` — 您的数据可能还在原来的位置。

JayC-L:
(1)目前是v0.51.134
(2) 会话列表为空，但在hermes dashboard中能看到。基础设置归零（api还在）。固定标签消失。
(3) 数据还在原来的位置

nesquena-hermes:
[...]

PR #4449, opened 2026-06-18
paths: default WebUI state to $HERMES_HOME/webui; allow HERMES_WEBUI_…

tomtong2015:
…SERVER_CWD

Two related path-resolution fixes for containerized / non-default-home deploys:

- State directory: bootstrap's state_dir and config's STATE_DIR fell back to the platform default even when HERMES_HOME was set, so WebUI state could land in a different home than the agent's. Default both to $HERMES_HOME/webui (the README-documented default) when HERMES_HOME is set; HERMES_WEBUI_STATE_DIR still overrides, and the platform default applies when neither is set. Scoped to STATE_DIR only — _DEFAULT_HERMES_HOME (auth.json/.env/skills resolution) is left unchanged.

- Server cwd: bootstrap hard-chdir'd into the agent dir, which is read-only in container images. execute_code's project-mode cwd falls back to os.getcwd() when a turn hasn't set TERMINAL_CWD yet, landing relative file writes (plt.savefig("x.png")) in that read-only dir. Honor HERMES_WEBUI_SERVER_CWD so a deployment can point the server cwd at a writable workspace.

Both default to the previous behavior when the env vars are unset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,7 @@ Full list of environment variables:
 | `HERMES_WEBUI_PORT` | `8787` | Port |
 | `HERMES_WEBUI_STATE_DIR` | `$HERMES_HOME/webui` (Windows default `%LOCALAPPDATA%\hermes\webui`, POSIX default `~/.hermes/webui`) | Where sessions and state are stored |
 | `HERMES_WEBUI_DEFAULT_WORKSPACE` | `~/workspace` | Default workspace |
+| `HERMES_WEBUI_SERVER_CWD` | *(unset)* | Working directory for the server process. Defaults to the agent dir; point it at a writable workspace when the agent dir is read-only (e.g. container images) so relative file writes from `execute_code` land somewhere writable |
 | `HERMES_WEBUI_DEFAULT_MODEL` | *(provider default)* | Optional model override; leave unset to use the active Hermes provider default |
 | `HERMES_WEBUI_PASSWORD` | *(unset)* | Set to enable password authentication |
 | `HERMES_WEBUI_CSP_CONNECT_EXTRA` | *(unset)* | Optional space-separated `http(s)://` or `ws(s)://` origins to append to the report-only CSP `connect-src` directive for reverse-proxy or tunnel deployments |
--- a/api/config.py
+++ b/api/config.py
@@ -71,8 +71,12 @@ TLS_ENABLED = TLS_CERT is not None and TLS_KEY is not None
 # ── State directory (env-overridable, never inside repo) ──────────────────────
 _DEFAULT_HERMES_HOME = _platform_default_hermes_home()
 
+# WebUI state defaults to $HERMES_HOME/webui (the README-documented default) so it
+# sits alongside the agent's home when HERMES_HOME is set; HERMES_WEBUI_STATE_DIR
+# still overrides, and the platform default applies when neither is set.
+_DEFAULT_STATE_HOME = Path(os.getenv("HERMES_HOME") or _DEFAULT_HERMES_HOME).expanduser()
 STATE_DIR = (
-    Path(os.getenv("HERMES_WEBUI_STATE_DIR", str(_DEFAULT_HERMES_HOME / "webui")))
+    Path(os.getenv("HERMES_WEBUI_STATE_DIR", str(_DEFAULT_STATE_HOME / "webui")))
     .expanduser()
     .resolve()
 )
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -494,7 +494,8 @@ def main() -> int:
 
     python_exe = ensure_python_has_webui_deps(discover_launcher_python(agent_dir), agent_dir)
     state_dir = Path(
-        os.getenv("HERMES_WEBUI_STATE_DIR", str(Path.home() / ".hermes" / "webui"))
+        os.getenv("HERMES_WEBUI_STATE_DIR")
+        or Path(os.getenv("HERMES_HOME") or (Path.home() / ".hermes")) / "webui"
     ).expanduser()
     state_dir.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
Read the diff across `api/config.py`, `bootstrap.py`, and `README.md` against `origin/master`, plus the agent-side cwd contract. Two solid fixes; the `SERVER_CWD` half is clean, the `STATE_DIR` half is correct in intent but has one migration blind spot worth calling out.

## Server cwd (the clean half)

The cross-repo claim checks out. The agent resolves an unset working dir to the server process cwd: `tool_executor.py:409` and `:922` (agent side) both do

```python
cwd = function_args.get("workdir") or os.getenv("TERMINAL_CWD", os.getcwd())
```

So when a turn hasn't set `TERMINAL_CWD` yet, relative writes land in the server's cwd — which `bootstrap.py:508` (master) hard-pins to `agent_dir`, read-only in container images. The override now flows into both launch paths: the foreground `os.chdir(server_cwd)` at `bootstrap.py:530` (with a clean `RuntimeError` on failure) and the detached `Popen(cwd=server_cwd)` at `:564`. Defaulting to the old value when unset keeps existing deploys unchanged. No notes here — this is right.

## State dir (correct intent, one gap)

This actually aligns `config.py` with what the README *already* claimed: the `HERMES_WEBUI_STATE_DIR` row documents the default as `$HERMES_HOME/webui`, but on master `STATE_DIR` was built from `_platform_default_hermes_home()` (`api/config.py:72-78`), which never reads `HERMES_HOME` — it only branches on `os.name`/`LOCALAPPDATA` (`api/paths.py:62-76`). So on POSIX, `HERMES_HOME` had **zero** effect on WebUI state; it always went to `~/.hermes/webui`. The PR makes the env var actually do what the docs say. Good.

The risk: this silently relocates where the server *reads* state for any POSIX user who set `HERMES_HOME` for the agent but whose WebUI sessions live at the master-era `~/.hermes/webui`. After upgrade the server reads `$HERMES_HOME/webui` (empty) and the app looks wiped — the same #2905 failure class, now on Linux.

The existing self-heal won't catch it. The empty-state sibling scan only walks `STATE_DIR.parent.iterdir()`:

```python
state_parent = STATE_DIR.parent
...
for sibling in state_parent.iterdir():
    ...
    sibling_sessions = sibling / "sessions"
```

(`api/config.py:613-620`). With `HERMES_HOME=/data/hermes`, the new parent is `/data/hermes` while the old data sits under `~/.hermes` — a *different* parent tree — so the warning never fires and the user gets no breadcrumb.

### Recommendation

Keep the change (it's the documented contract), but close the migration gap one of two ways:

1. **Cheapest:** extend the empty-state check in `config.py` to also probe the platform default `_platform_default_hermes_home() / "webui"` when it differs from the resolved `STATE_DIR`, and emit the same "set `HERMES_WEBUI_STATE_DIR=...`" hint. That generalizes the sibling scan to the cross-home case.
2. Or call this out explicitly in the release notes as a behavior change for POSIX `HERMES_HOME` users, with the `HERMES_WEBUI_STATE_DIR` override as the pin.

Also worth a quick check: `bootstrap.py:497-499` resolves the same default independently (`HERMES_HOME or ~/.hermes`) rather than importing `config.STATE_DIR`. The two agree today, but they're now two copies of the resolution rule — a divergence later (e.g. the Windows `%LOCALAPPDATA%` branch in `paths.py`) would have bootstrap create one dir and the server read another. Centralizing on a single helper would prevent that.

nesquena-hermes (changes requested):
Thanks @tomtong2015 — the intent (state living alongside `$HERMES_HOME` + a writable server cwd for read-only containers) is reasonable and the `HERMES_WEBUI_SERVER_CWD` addition is fine. But the state-dir default change has a silent-data-loss regression that blocks it:

### (must-fix) Changing the default to `$HERMES_HOME/webui` silently relocates an existing install's state

Codex reproduced this: when a user has `HERMES_HOME` set to something other than `~/.hermes` AND has not set `HERMES_WEBUI_STATE_DIR`, the default flips from the old `~/.hermes/webui` (the platform `_DEFAULT_HERMES_HOME / "webui"`) to `$HERMES_HOME/webui`. If `$HERMES_HOME != ~/.hermes`, the app opens a **different, empty** state dir on next launch — the user's existing sessions/settings/workspaces silently vanish from the UI (the data is still on disk at the old path, just not read), and the sibling-state warning won't find the old dir either.

Two code paths have this:
- `api/config.py` — the `_DEFAULT_STATE_HOME` resolution + the second `getenv` fallback.
- `bootstrap.py:~496` — it exports the newly computed dir into `HERMES_WEBUI_STATE_DIR` before the server starts, so `api.config` can't even fall back to the old populated path.

**Fix (migration-safe resolver, shared by both sites):** when `HERMES_WEBUI_STATE_DIR` is unset, prefer `$HERMES_HOME/webui` **only if** it already contains WebUI state OR the old default (`~/.hermes/webui`) does not exist; otherwise keep reading the old populated `~/.hermes/webui`. Apply the same resolver in `bootstrap.py` before the `os.environ.setdefault(...)` so direct `bootstrap.py`/`start.sh` launches don't pre-empt the fallback. That preserves the new sibling-of-HERMES_HOME default for fresh installs while never orphaning an existing one.

A test for the migration case would lock it in: `HERMES_HOME=/custom` + existing `~/.hermes/webui` state + unset `HERMES_WEBUI_STATE_DIR` must still resolve to the populated `~/.hermes/webui`, not an empty `/custom/webui`.

(Note: when `HERMES_HOME=~/.hermes` — the common case — old and new defaults are the same path, so most users are unaffected; this only bites custom-`HERMES_HOME` deployments, but for those it's silent session loss, so it's worth getting right before merge.) Happy to re-gate once the resolver is migration-safe.

nesquena-hermes:
Thanks @tomtong2015. This overlaps active work: the deep review of #4454 (isolated HERMES_HOME profile mode) found that the WebUI state dir must follow `HERMES_HOME` — which is exactly the contract this PR proposes. To avoid two PRs landing colliding state-dir defaults, we're folding the `$HERMES_HOME/webui` state-dir contract into the #4454 rework so it lands once, coherently, with the isolation-correctness tests that depend on it. Holding this PR pending that — your `HERMES_WEBUI_SERVER_CWD` bridge and the documented-default alignment are noted and will be reconciled there. Will cross-link when #4454 re-lands.

nesquena-hermes:
Closing as superseded: the $HERMES_HOME/webui state-dir contract proposed here was folded into #4454 (shipped v0.51.528) so the isolation-correctness tests that depend on it land together — you're credited as co-author in that PR/CHANGELOG. The HERMES_WEBUI_STATE_DIR override behavior + the upgrade-migration note for relocated HERMES_HOME are documented in the README. Thanks @tomtong2015!

PR #4450, opened 2026-06-18
workspace: allow extra trusted workspace roots via env

tomtong2015:
resolve_trusted_workspace blocks any path under a system prefix like /var, but some deployments mount legitimate user workspaces there — REANA, for example, mounts each user's persistent workspace under /var/reana/users/<uid>/.../. A session whose workspace resolved there was rejected as a system path, so POST /api/session/new 400'd and the first prompt couldn't start (and /api/media 403'd for files written there).

Add HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS (os.pathsep-separated absolute paths, mirroring MEDIA_ALLOWED_ROOTS): paths under a listed root are carved out of the system-dir block in both _is_blocked_workspace_path and _is_blocked_system_path. Unrelated system paths (/var/lib, /etc, ...) stay blocked. Default unset = no behavior change.


nesquena-hermes (changes requested):
Thanks for this @tomtong2015 — the REANA use case is real (a deployment that mounts user workspaces under a normally-blocked `/var` prefix should be able to opt specific roots back in), and a lot of the implementation is right: `_extra_workspace_prefixes()` parses + resolves the `os.pathsep`-separated env var correctly, it's properly **default-off** (unset/empty → `()` → `_workspace_carveout_prefixes()` is exactly `_USER_TMP_PREFIXES`, zero behavior change), and the carve-out containment via `_is_within` is precise (a listed `/var/reana` carves out `/var/reana/**` only, not `/var` or `/var/reana-evil`).

But there's one objective problem that means **the fix doesn't actually work on the path that matters**, and a test gap.

### (must-fix) The carve-out is dead code on the workspace-validation path — `/var/reana` is still rejected

`_is_blocked_workspace_path(...)` is the function the real entry points call (`validate_workspace_to_add` at `api/workspace.py:833`, which raises `ValueError("Path points to a system directory: ...")` → the 400 you're trying to eliminate; and `resolve_trusted_workspace` via `:180`). That function checks `_is_blocked_posix_workspace_path(posix_probe)` **first**:

```python
# api/workspace.py:607-609
posix_probe = raw_path if raw_path not in (None, "") else candidate.as_posix()
if _is_blocked_posix_workspace_path(posix_probe):
    return True
# ... PR's modified carve-out loop is below, at ~:615 — never reached for /var/* paths
for tmp in _workspace_carveout_prefixes():
    if _is_within(candidate, tmp) or (raw is not None and _is_within(raw, tmp)):
        return False
```

`_is_blocked_posix_workspace_path` (`:509-546`) is **not touched by this PR**. It hardcodes its own carve-out list (`/var/folders`, `/var/tmp`, ...) and hardcodes `/var` in `blocked_roots` (`:528`), so for `/var/reana/users/<uid>/workflows/<id>` it returns `True` at `:544` and `_is_blocked_workspace_path` returns `True` at `:609` — **before your carve-out loop at `:615` ever runs**. Net effect: setting `HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS=/var/reana` changes nothing for adding/validating a `/var/reana` workspace; it still 400s.

You did wire `_is_blocked_system_path` (`:556`) correctly — that one honors the carve-out. The bug is isolated to the posix-first ordering in `_is_blocked_workspace_path`.

I reproduced this directly against your head in an isolated sandbox:

```
test_extra_workspace_root_carveout_allows_var_reana   FAILED
  # with HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS=/var/reana:
  # _workspace_carveout_prefixes() DOES include /var/reana  (your helper works)
  # but _is_blocked_workspace_path(Path('/var/reana/users/u1/workflows/w1'), '...') is still True
test_system_path_check_DOES_honor_carveout            PASSED   # _is_blocked_system_path is fine
```

**Fix-spec** — thread the carve-out into the posix check too. The cleanest option is to teach `_is_blocked_posix_workspace_path` about the extra roots (it already has a carve-out loop at `:522-524`): build a posix-form set from `_extra_workspace_prefixes()` and `return False` when `candidate` is within one, *before* the `blocked_roots` loop at `:543`. (Resolving the env roots to `PurePosixPath` is straightforward since they're required to be absolute.) Alternatively, evaluate the `_workspace_carveout_prefixes()` carve-out at the top of `_is_blocked_workspace_path` before the posix probe — but the posix-internal fix keeps both code paths consistent and also covers the non-POSIX-host branch.

### (must-fix) No tests for a security-boundary change

The diff adds zero tests. A carve-out that opts paths back through the blocked-system-path gate needs coverage for all three properties so a future refactor can't silently regress them:
1. **default-off**: unset env → `/var/reana` stays blocked (and `_workspace_carveout_prefixes() == _USER_TMP_PREFIXES`).
2. **the carve-out works**: with the env set, `/var/reana/...` is accepted by `validate_workspace_to_add` / `_is_blocked_workspace_path` (this is the assertion that's currently RED).
3. **no over-reach**: with `/var/reana` listed, sibling/parent paths (`/var`, `/var/reana-evil`, `/var/other`, and a `/var/reana/../etc` traversal) stay blocked.

### Note (not blocking this round)

Separately, since this widens a security boundary via deployment config, the *posture* (do we want an env-var workspace-root carve-out at all, and with what guardrails — e.g. requiring `is_absolute()` and surfacing malformed entries rather than the current silent `except: pass`) is a maintainer call I'll raise once the mechanism actually works. Getting the two items above fixed first makes that review concrete. Thanks again — the structure is close, it just needs to be wired into the gate that the add/validate path actually uses, plus tests.

nesquena-hermes:
Read the full diff plus the three gating functions in `api/workspace.py` on this branch (`5182910a`). The carve-out idea is sound and the `HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS` plumbing is clean, but **the primary motivating case — `POST /api/session/new` 400'ing on a `/var/reana/...` workspace — is not actually fixed by this diff.** The new carve-out is dead code on that path. The `/api/media` half does work. Details below.

## Why the session-creation fix doesn't fire

`resolve_trusted_workspace()` (api/workspace.py:833) gates on `_is_blocked_workspace_path(candidate, path)`. Inside that function the **first** check is a POSIX probe that returns before the patched carve-out loop is ever reached:

```python
# api/workspace.py:607-616
posix_probe = raw_path if raw_path not in (None, "") else candidate.as_posix()
if _is_blocked_posix_workspace_path(posix_probe):
    return True
...
for tmp in _workspace_carveout_prefixes():   # <-- your new carve-out, line 614
    if _is_within(candidate, tmp) or ...:
        return False
```

`_is_blocked_posix_workspace_path()` (api/workspace.py:509) has its **own hardcoded carve-out list** that the PR didn't update:

```python
carveouts = (
    PurePosixPath('/var/folders'),
    PurePosixPath('/private/var/folders'),
    PurePosixPath('/var/tmp'),
    PurePosixPath('/private/var/tmp'),
)
...
blocked_roots = (... PurePosixPath('/var') ...)
```

So for `/var/reana/users/<uid>/...`: it's under `/var`, not under any of those four carve-outs → `_is_blocked_posix_workspace_path` returns `True` at line 608 → `_is_blocked_workspace_path` returns `True` → `resolve_trusted_workspace` raises `ValueError` → the `/api/session/new` handler (api/routes.py:8714 → `resolve_trusted_workspace(body.get("workspace"))`) still 400s. The new `_workspace_carveout_prefixes()` loop at line 614 never executes for this input.

## Why /api/media does work

The media/workspace-add path goes through `_is_blocked_system_path()` (api/workspace.py:549), which you *did* patch correctly — it checks `_workspace_carveout_prefixes()` first (line 556) and returns `False` before the blocked-roots loop. That's why the route guard at api/routes.py:16087 honors the extra root. So the two halves of the PR diverge: media fixed, session creation not.

## Fix sketch

Teach the POSIX probe about the extra roots too. Smallest change, in `_is_blocked_posix_workspace_path` (api/workspace.py:509):

```python
carveouts = [
    PurePosixPath('/var/folders'), PurePosixPath('/private/var/folders'),
    PurePosixPath('/var/tmp'),     PurePosixPath('/private/var/tmp'),
]
carveouts += [PurePosixPath(p.as_posix()) for p in _extra_workspace_prefixes()]
for tmp in carveouts:
    if _posix_is_within(candidate, tmp):
        return False
```

That keeps a single source of truth if you'd rather route both lists through `_workspace_carveout_prefixes()` — note that helper returns resolved `Path`s, so convert via `.as_posix()` for the `PurePosixPath` comparison.

## Test gap
There are no tests in the diff. A regression test that sets `HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS=/var/reana` and asserts `resolve_trusted_workspace('/var/reana/users/u1/wf/1')` returns the path (not raises) would have caught this — it exercises the `_is_blocked_workspace_path` branch the README example actually hits, not just `_is_blocked_system_path`. Worth adding one for each of the three gating functions so they can't drift again.

Happy to re-look once the POSIX probe shares the carve-out list.

tomtong2015:
Good catch on the macOS symlink gap — fixed in the follow-up commit. The POSIX probe matches against the raw, unresolved string, but the carve-out added only the resolved form, so on macOS (/var → /private/var) a literal /var/reana/... candidate missed it — exactly the case the static _USER_TMP_PREFIXES dodges by listing both /var/folders and /private/var/folders. _parse_extra_workspace_roots now returns (expanded, resolved) pairs; _extra_workspace_prefixes() still yields the resolved form for the Path-based gates (their candidates are resolved), and a new _extra_workspace_posix_carveouts() feeds the POSIX probe both forms, deduped. Added test_posix_carveouts_keep_both_symlink_forms, which uses a real symlink so it exercises the both-forms path on Linux CI too.

Re the red test (3.13, 2) shard: that's tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata (AssertionError: gw_legacy_import_weixin_001 missing from /api/sessions … (test-isolation)) — unrelated to this PR (no workspace/gating code), and the failure message itself flags it as a test-isolation leak; my test_extra_workspace_roots.py passed on that shard and every other shard/Python is green. A re-run should clear it.

nesquena-hermes:
Re-reviewed at `31bbf29b` against my earlier note. The POSIX-probe gap I flagged is now closed correctly, and the follow-up went further than the minimum I sketched — nice.

## What changed since the last review

The earlier problem: `_is_blocked_posix_workspace_path()` ran first inside `_is_blocked_workspace_path()` and had its own hardcoded carve-out list, so `/var/reana/...` was rejected at the POSIX probe before the new `_workspace_carveout_prefixes()` loop ever ran. The session-creation path (`resolve_trusted_workspace` → `_is_blocked_workspace_path`) still 400'd; only `/api/media` worked.

The new diff teaches the probe about the extra roots, and ordering is right — carve-out check before the blocked-roots loop (`api/workspace.py:599`):

```python
carveouts += list(_extra_workspace_posix_carveouts())
for tmp in carveouts:
    if _posix_is_within(candidate, tmp):
        return False
...
for blocked in blocked_roots:   # /var, /etc, ...
    if _posix_is_within(candidate, blocked):
        return True
```

## The both-forms handling is the right call

The macOS `/var` → `/private/var` symlink case is real and easy to miss. `_extra_workspace_posix_carveouts()` (`api/workspace.py:494`) emitting *both* the literal and resolved form — deduped — mirrors the static `/var/folders` + `/private/var/folders` pattern exactly, while `_extra_workspace_prefixes()` keeps yielding only the resolved form for the `Path`-based gates whose candidates are already resolved. Keeping those two views distinct (`_parse_extra_workspace_roots` returns `(expanded, resolved)` pairs) is clean and avoids cross-contaminating the resolved-only gates.

## Tests now cover the branch that was dead before

`test_carveout_allows_listed_root_on_every_gate` asserts all three gates including the POSIX probe (`_is_blocked_posix_workspace_path(WS) is False`), which is precisely the assertion that would have caught the original miss. `test_carveout_does_not_over_reach` is the one I'd have asked for — `/var/reana-evil`, `/var/other`, and the `/var/reana/../etc` traversal all stay blocked, so the carve-out can't be widened by a sibling-prefix or `..` escape. And `test_posix_carveouts_keep_both_symlink_forms` uses a real symlink so the both-forms path runs on Linux CI, not just macOS.

On the red `test_sessions_response_backfills_imported_messaging_source_metadata` shard — agreed it's unrelated; that's the test-isolation leak tracked in #4557, no workspace/gating code in its path. A re-run should clear it.

This LGTM from a correctness standpoint now. The carve-out is default-off, validated (non-absolute and unresolvable entries are dropped with a warning rather than silently), and the three gates share a single source of truth.

nesquena-hermes (changes requested):
Re-gated the re-push (Codex + Opus + a direct reproduction). **The dead-code flaw from the last round is genuinely fixed** — the carve-out is now wired into the POSIX probe *before* the return-False loop, plus `_is_blocked_system_path` and `_is_blocked_workspace_path`, and I verified it works: `/var/reana/users/...` is allowed with the env set, non-configured `/var` paths stay blocked, default-off keeps everything blocked, and the parent `/var/reana` (above the configured root) stays blocked. Nice rework. 👍

One more security item before this can ship, though — a malformed/too-broad env entry can silently widen the boundary to the whole filesystem:

### 🔴 Blocker: a configured root that resolves to `/` or a blocked system root unblocks everything under it

`_parse_extra_workspace_roots` validates "absolute + resolvable" but doesn't bound how *broad* the resolved root is, and `Path.resolve()` collapses `..`. **Reproduced** (isolated env):

```
HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS=/var/reana/..      → resolves to /var
   _is_blocked_workspace_path('/var/secret')  → False   (should be True)

HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS=/var/reana/../..   → resolves to /
   _is_blocked_workspace_path('/etc/passwd')  → False   (should be True)
```

So a typo or an over-broad value (`/var`, `/`, `/var/reana/..`) silently turns the whole subtree — up to the entire filesystem — into a trusted workspace root. It's operator-controlled (lower severity than a request-driven escape), but it's a real footgun that defeats the validation this function is supposed to provide.

### Fix-spec (Codex + Opus agree)
In `_parse_extra_workspace_roots`, after resolving, **skip with a warning** any entry that:
- contains `..` path components, OR
- resolves to `/`, OR
- is equal to, or a parent of, any `_workspace_blocked_roots()` entry.

i.e. require each configured root to sit *strictly below* a blocked system root. Add regressions for `/var/reana/..`, `/var/reana/../..`, `/`, and a bare `/var`.

### Lower-priority residual (Opus, accept-or-fold — not a blocker)
`_is_blocked_workspace_path` uses `_is_within(candidate, tmp) OR _is_within(raw, tmp)`; a symlink *inside* a configured root pointing at a blocked target passes the raw check. This is **pre-existing** (the static `_USER_TMP_PREFIXES` uses the same OR), and the resolved-only `_is_blocked_system_path` gate — which guards the actual runtime file ops and symlink targets — correctly blocks it (fails safe). Worth a follow-up to align the two gates, but it doesn't block this PR.

The mechanism is sound and default-off; this is just bounding the configured root so it can't be pointed at `/`. Happy to re-gate as soon as the validation rejects too-broad roots. Thanks @tomtong2015.

tomtong2015:
Thanks for the thorough re-gate — and for the reproductions; the `..`-collapse cases were exactly right.

Pushed `1a49103d` with the boundedness fix per your spec. `_parse_extra_workspace_roots` now skips (with a warning) any entry that:

- contains `..` path components — checked **before** `resolve()`, since `resolve()` collapses them (this is what let `/var/reana/..` through);
- resolves to a filesystem root; or
- is equal to, or a parent of, any `_workspace_blocked_roots()` entry.

So a configured root must sit *strictly below* a blocked system root. Your reproductions now behave:

```
/var/reana/..     → skipped (.. component)     ; /var/secret  stays blocked
/var/reana/../..  → skipped (.. component)     ; /etc/passwd  stays blocked
/                 → skipped (filesystem root)  ; /etc/passwd  stays blocked
/var              → skipped (== blocked root)  ; /var/secret  stays blocked
```

Regressions added in `test_extra_workspace_roots.py`:
- `test_over_broad_roots_are_rejected` — all four values above: each is dropped, the victim path stays blocked on all three gates (`_is_blocked_workspace_path`, `_is_blocked_system_path`, `_is_blocked_posix_workspace_path`), and a warning is logged.
- `test_strict_subdir_of_blocked_root_still_allowed` — the legitimate `/var/reana` case is unaffected.

Full carve-out + workspace-gating suites green locally.

On the lower-priority symlink-OR residual: agreed it's pre-existing (the static `_USER_TMP_PREFIXES` shares the same OR) and `_is_blocked_system_path` fails safe on the actual runtime file ops, so I've left it as the follow-up you suggested rather than widen this PR. Happy to do that gate-alignment in a separate change.

Thanks!

nesquena-hermes:
Re-reviewed the re-push (head `1a49103d`) — nice work, the objective bug I flagged is **fixed**: the carve-out is now wired into `_is_blocked_posix_workspace_path` (via `_extra_workspace_posix_carveouts()`), which is the posix-first probe `_is_blocked_workspace_path` consults before the carve-out loop. Your new regression test (`_is_blocked_posix_workspace_path(WS) is False` for a listed `/var/reana` root, still `True` for unlisted `/var/*`) locks exactly the path that was dead before. The containment is still precise and default-off. So the code is correct now.

One thing keeps this from an autonomous merge, and it's a maintainer call rather than a code defect: this introduces a **new `HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS` env var as the user-facing config surface** (documented in the README). Per the project's contribution rubric, new `HERMES_*` env vars are reserved for **secrets**; non-secret behavioral config — and a trusted-workspace-roots allowlist is behavioral config — is supposed to live in `config.yaml` (with the docs pointing there). It's also a **security boundary** (it re-trusts normally-blocked system prefixes), which is doubly a maintainer-judgment surface.

Two clean paths forward, @nesquena's call:
1. **Accept the env-var mechanism** as a deliberate deployment-infra exception (REANA-style mounts are arguably infra, not user prefs) — if so, I'll gate + ship as-is.
2. **Migrate to `config.yaml`** (e.g. `workspace.extra_trusted_roots: [...]`), keeping the env var only as an internal bridge if needed and pointing the README at the config key — matching how other behavioral settings are surfaced.

Parking for @nesquena to make that call. @tomtong2015 — the implementation is solid; this pause is purely about *where* the setting is configured, not whether the feature is wanted.

tomtong2015:
Thanks for the careful re-review — went with **path 2 (migrate to `config.yaml`)**, since a trusted-workspace-roots allowlist is non-secret behavioral config (and a security boundary), so it belongs alongside the other behavioral settings rather than in a `HERMES_*` env var.

Pushed as `f1dc8635`. What changed:

- **New user-facing surface — `config.yaml`:**
  ```yaml
  workspace:
    extra_trusted_roots:
      - /var/reana
  ```
  `_extra_workspace_root_pairs()` now reads this via `get_config()`. A scalar `os.pathsep`-separated string is also accepted.
- **Env var demoted to an internal deployment bridge.** `HERMES_WEBUI_EXTRA_WORKSPACE_ROOTS` still works, but it's now merged with the config key (config entries first, deduped) and the README documents `config.yaml` as the surface. We keep the bridge only so a deployment can inject a root at boot from a *dynamic* mount path (REANA's per-user workspace path isn't known until runtime) — happy to drop it entirely if you'd rather it be config-only.
- **Validation is unchanged.** Both sources funnel through the same `_parse_extra_workspace_roots()`, so every check you approved last round (absolute, no `..`, not a filesystem root, strictly-below-a-blocked-root, both symlink forms) applies identically. The only edit there is rewording the warnings to be source-agnostic ("extra trusted workspace root") now that they cover two surfaces.
- **Tests + docs:** added config.yaml cases (list + scalar), config+env merge/dedup, default-off when the key is absent, invalid-type ignored-with-warning, and a config-unloadable fallback to the env bridge. README now documents `workspace.extra_trusted_roots` as the primary surface. Full suite green.

The setting now lives where the rubric wants it — ready to gate whenever you are. Thanks again!

Issue #2066, opened 2026-05-11
Stuck sidebar spinner on completed sessions (INFLIGHT cache not cleared)


franksong2702:
## Problem

When a session finishes streaming, the sidebar spinner (`.session-state-indicator.is-streaming`) can remain spinning indefinitely even though the session is fully complete.

### Root Cause

The sidebar uses `_isSessionEffectivelyStreaming(s)` to decide whether to show the spinner:

```javascript
function _isSessionEffectivelyStreaming(s) {
  return Boolean(s && (s.is_streaming || _isSessionLocallyStreaming(s)));
}

function _isSessionLocallyStreaming(s) {
  const isActive = S.session && s.session_id === S.session.session_id;
  return Boolean(
    (isActive && S.busy)
    || (typeof INFLIGHT === 'object' && INFLIGHT && INFLIGHT[s.session_id])
  );
}
```

When the server reports `is_streaming: false`, `_isSessionLocallyStreaming(s)` can still return `true` because `INFLIGHT[s.session_id]` was never cleaned up. This happens when a streaming session is interrupted abnormally — page refresh, network disconnect, gateway restart, or switching sessions mid-stream — and the `delete INFLIGHT[activeSid]` line never executes.

The `INFLIGHT` map is an in-memory cache (`const INFLIGHT={};` in `ui.js`) keyed by `session_id`. It is set during `send()` and normally deleted when the stream completes or when the user switches sessions. But the cleanup only happens in the happy path:

- `messages.js` line 291: `delete INFLIGHT[activeSid]` — only after successful stream completion
- `messages.js` line 308: `delete INFLIGHT[activeSid]` — only after approval flow completes
- `messages.js` line 404: `delete INFLIGHT[activeSid]` — only in `_reconcilePostSend` when stream is done

If the stream never completes cleanly (error, timeout, user navigates away), the entry stays forever.

### Reproduction

1. Start a conversation in any session
2. While the agent is still streaming, close the tab or switch to a different session
3. The `INFLIGHT[sessionId]` entry remains
4. The sidebar shows the spinner indefinitely because `_isSessionLocallyStreaming()` returns true

### Impact

- Visual noise: completed sessions appear "active" in the sidebar
- Misleading state: users may think a session is still running
- The `is-hidden` class on timestamps (line 2600) also hides the relative time, making the list less useful

### Proposed Fix

[...]

Issue #856, opened 2026-04-22
feat(sessions): session list UX — loading spinner, unread dot, last-reply timestamp


nesquena-hermes:
## Summary

The session sidebar list is functional but missing three UX elements that would significantly improve at-a-glance awareness: (a) a spinner/animation on sessions currently generating a response, (b) an unread dot indicator for sessions with new messages the user hasn't viewed yet, and (c) a human-readable timestamp showing the time of the last reply.

Inspired by Codex's session list UI design.

## Current state

`static/sessions.js` renders session items via `_renderOneSession(s)`. The session object (from `api/models.py Session.compact()`) already includes:

- `updated_at` — Unix timestamp of last activity
- `message_count` — total messages
- `session_id` — stable ID
- `pinned`, `archived`, `project_id`, `profile`

**Active streaming state** is tracked globally via `S.busy` (boolean, set in `sessions.js`) and `S.activeStreamId`. A session item gets the CSS class `active` when `S.session.session_id === s.session_id` — but no class for "streaming in progress".

**What's missing for the three features:**

### (a) Spinning loader for active sessions

`S.busy` is true when any session is streaming, and `S.session` holds the active session. The session list already checks `isActive = S.session && s.session_id === S.session.session_id`, so we can combine `isActive && S.busy` to show a spinner. The issue is that `renderSessionListFromCache()` is called on every token (via `S.busy` state changes), so the spinner would appear/disappear correctly.

Currently no CSS class or DOM element signals an in-progress session to the sidebar item. A `.session-item.streaming` class could add a CSS animation.

### (b) Unread dot indicator

The backend has no `unread` or `last_viewed_at` field in `Session.compact()`. Tracking "unread" requires knowing whether the user has viewed the latest messages. Options:
1. Client-side: store `last_viewed_message_count[session_id]` in `localStorage` and compare against `s.message_count` from the API
2. Server-side: add `last_viewed_at` timestamp to Session model, update it on `/api/session` GET

The client-side approach is simpler and avoids a schema change. It's imprecise (count-based, not content-based) but sufficient for this UX.

### (c) Last reply timestamp

`updated_at` is already available in the session list payload (`Session.compact()` returns it, and `/api/sessions` emits it). The frontend utility `_sessionTimestampMs(s)` already reads it. The timestamp just isn't rendered in the item itself — only used for sorting and date-group headers.

[...]

PR #4451, opened 2026-06-18
fix(sidebar): bubble hidden lineage stream state

santastabber:
## Thinking Path

- Sidebar rows can collapse multiple WebUI continuation/lineage sessions into one visible conversation.
- A running continuation can be hidden from the normal sidebar list when it is archived, while the user-facing conversation remains visible through another lineage row.
- The visible row still needs to show running/completion state, otherwise users lose the spinner and activity ordering for the chat they interacted with.
- The fix keeps hidden archived lineage rows hidden, but bubbles their runtime/sidebar state onto the visible lineage row.
- Pin/unpin uses the same sidebar cache, so it should update the cached row and re-render immediately before the slower server reconciliation.

## What Changed

- Treat server-confirmed `is_streaming`, explicit pending-message markers, and active local `S.busy` as effective sidebar streaming state, without trusting stale raw `active_stream_id` or `pending_started_at` values by themselves.
- Bubble hidden archived lineage child state to the visible lineage row:
  - running spinner state,
  - latest activity timestamp for ordering/date grouping,
  - unread and attention state.
- Keep hidden archived lineage rows hidden as standalone rows while still using them as sidebar-state references.
- Avoid misclassifying ordinary collapsed lineage segments as child sessions.
- Update pin/unpin to mutate the cached `_allSessions` row and render immediately, then reconcile with the server.
- Add focused regression coverage for streaming-state detection, stale stream-id false-spinners/unreads, hidden archived lineage bubbling, collapsed-lineage child-badge avoidance, polling fallback alignment, timestamp bubbling, and immediate pin cache rendering.

## Why It Matters

Users type into the visible conversation row, not into an internal archived continuation row. If the internal row is the one streaming, the sidebar should still show the visible conversation as running and keep it sorted with current activity. This preserves the mental model that the sidebar represents conversations, not every hidden implementation segment.

## Verification

- `node -c static/sessions.js`
- `python -m pytest tests/test_issue2454_active_session_spinner.py tests/test_issue856_background_completion_unread.py tests/test_session_lineage_collapse.py tests/test_sidebar_pin_immediate_reorder.py -q`
  - Result: `74 passed, 1 warning`
- Delegated implementation review and PR review found collapsed-lineage, stale-stream, and persisted-observed-stream edge cases; fixed before this update.
- Delegated sanitization review: `PASS_WITH_NITS`; the noted local absolute path in the PR body was removed before update.
- Local diff sanitization scan found no local paths, hostnames, ports, session IDs, usernames, emails, tokens, or private environment details.

## Risks / Follow-ups

- The change is intentionally limited to sidebar-derived render state. It does not mutate session archival state or lineage metadata on disk.
- Full local test runner was not used for final verification because focused sidebar coverage exercises the touched behavior; CI should provide the full matrix.

## Contract Routing

Task type: sidebar/session metadata behavior fix.

Touched areas:
- `static/sessions.js`
- sidebar lineage collapse / child-session decoration
- streaming completion polling fallback
- focused sidebar regression tests

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/UIUX-GUIDE.md`

State layer / invariant:
- Sidebar render-derived state only. Hidden archived lineage rows remain hidden, but their runtime state may bubble to the visible conversation row.

## Model Used

AI-assisted implementation and review using OpenAI GPT-5.5 via Hermes Agent, with terminal/file tools and two read-only delegate review passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Sidebar lineage rows now keep their running spinner, latest activity ordering, and unread/attention state when the active continuation is hidden as an archived lineage child. Pin/unpin actions also update the sidebar ordering immediately before the next server refresh.
+
 ## [v0.51.523] — 2026-06-19 — Release SH (cache trusted wiki page listings on browser reads)
 
 ### Changed
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -368,7 +368,11 @@ function _isSessionLocallyStreaming(s) {
 }
 
 function _isSessionEffectivelyStreaming(s) {
-  return Boolean(s && (s.is_streaming || _isSessionLocallyStreaming(s)));
+  return Boolean(s && (
+    s.is_streaming ||
+    _hasPendingUserMessageSignal(s) ||
+    _isSessionLocallyStreaming(s)
+  ));
 }
 
 function _hasPendingUserMessageSignal(s) {
@@ -376,7 +380,7 @@ function _hasPendingUserMessageSignal(s) {
 }
 
 function _isServerIdleSessionRow(s) {
-  return Boolean(s && s.session_id && !s.is_streaming && !s.active_stream_id && !s.pending_user_message && !s.has_pending_user_message);
+  return Boolean(s && s.session_id && !s.is_streaming && !s.active_stream_id && !s.pending_user_message && !s.has_pending_user_message && !s.pending_started_at);
 }
 
 function _reconcileActiveSessionIdleStateFromList(serverRows) {
@@ -605,7 +609,9 @@ function _markPollingCompletionUnreadTransitions(sessions) {
     const observedStreaming = _getSessionObservedStreaming()[sid];
     const messageCount = Number(s.message_count || 0);
     const lastMessageAt = Number(s.last_message_at || 0);
-    const completedObservedStream = wasStreaming === true && !isStreaming;
+    const hasServerRunSignal=Boolean(s.is_streaming||_hasPendingUserMessageSignal(s));
+    const canMarkCompletedStream=Boolean(hasServerRunSignal||previousSnapshot||observedStreaming);
+    const completedObservedStream = canMarkCompletedStream&&wasStreaming === true && !isStreaming;
     const completedWithNewMessages = Boolean(
       (previousSnapshot || observedStreaming)
       && !isStreaming
@@ -3278,8 +3284,11 @@ function _openSessionActionMenu(session, anchorEl){
       try{
         await api('/api/session/pin',{method:'POST',body:JSON.stringify({session_id:session.session_id,pinned:newPinned})});
         session.pinned=newPinned;
+        const cached=(_allSessions||[]).find(s=>s&&s.session_id===session.session_id);
+        if(cached) cached.pinned=newPinned;
         if(S.session&&S.session.session_id===session.session_id) S.session.pinned=newPinned;
-        renderSessionList();
+        renderSessionListFromCache();
+        void renderSessionList();
       }catch(err){
         showToast(t('session_pin_failed')+err.message);
         await renderSessionList();
@@ -3606,7 +3615,7 @@ function _applySessionListPayload(sessData, projData){
   _sessionListLoadError = null;
   _sessionListHasLoadedOnce = true;
   _markPollingCompletionUnreadTransitions(_allSessions);
-  const isStreaming = _allSessions.some(s => Boolean(s && s.is_streaming));
+  const isStreaming = _allSessions.some(s => _isSessionEffectivelyStreaming(s));
   if (isStreaming) {
     startStreamingPoll();
   } else {
@@ -4365,7 +4374,7 @@ function filterSessions(){
 }
 
 function _sessionTimestampMs(session) {
-  const raw = Number(session && (session.last_message_at || session.updated_at || session.created_at || 0));
+  const raw = Number(session && (session._sidebar_activity_at || session.last_message_at || session.updated_at || session.created_at || 0));
   return Number.isFinite(raw) ? raw * 1000 : 0;
 }
 
@@ -4741,6 +4750,8 @@ function _attachChildSessionsToSidebarRows(collapsedRows, rawSessions, rawRefere
     delete row._child_session_streaming;
     delete row._child_session_has_unread;
     delete row._child_session_attention;
+    delete row._child_session_latest_at;
+    delete row._sidebar_activity_at;
     return row;
   };
   const rows=(collapsedRows||[])
@@ -4755,6 +4766,14 @@ function _attachChildSessionsToSidebarRows(collapsedRows, rawSessions, rawRefere
   const bubbleSidebarState=(parentRow, childRow)=>{
     if(isChildStreaming(childRow)) parentRow._child_session_streaming=true;
     if(childHasUnread(childRow)) parentRow._child_session_has_unread=true;
+    const childActivityRaw=childRow
+      ? (childRow._sidebar_activity_at??childRow.last_message_at??childRow.updated_at??childRow.created_at??0)
+      : 0;
+    const childActivitySec=Number(childActivityRaw);
+    if(Number.isFinite(childActivitySec)&&childActivitySec>Number(parentRow._child_session_latest_at||0)){
+      parentRow._child_session_latest_at=childActivitySec;
+      parentRow._sidebar_activity_at=Math.max(Number(parentRow.last_message_at||parentRow.updated_at||parentRow.created_at||0), childActivitySec);
+    }
     const childAttention=childRow&&childRow.attention&&typeof childRow.attention==='object'?childRow.attention:null;
     if(!childAttention||!childAttention.kind||!Number.isFinite(Number(childAttention.count))||Number(childAttention.count)<=0) return;
     const priorityFor=(kind)=>kind==='approval'?3:(kind==='clarify'?2:1);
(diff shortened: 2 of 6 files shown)


nesquena-hermes (changes requested):
Thanks @santastabber — the bubbling mechanics here are genuinely well built, and I verified the two clean parts: the **ordering change is correct with no regression** (`_sidebar_activity_at` is recomputed each render via `cleanSidebarRow` deleting it at rebuild start, falls back to `last_message_at` for normal rows — no cross-render drift), and there's **no render-derived state leak** across session switches (all `_child_session_*` decoration is cleared per rebuild). Pin/unpin optimistic-then-reconcile is sound. I ran the full gate (Codex + Opus); Codex passed it, but Opus reproduced a real regression on the exact surface this touches, so I need one fix before shipping.

### (must-fix) Broadening `_isSessionEffectivelyStreaming` to trust raw `active_stream_id` causes a false spinner + false unread

The change makes `_isSessionEffectivelyStreaming` treat a row as streaming when it has an `active_stream_id`, **without reconciling that against the live stream registry**. But a dead-but-uncleared `active_stream_id` is routine after a restart or interrupted turn — that's precisely why the server computes `is_streaming = (active_stream_id ∈ live-set)` and why `_isSessionLocallyStreaming` was gated to the active session (the #2066 INFLIGHT fix). Consequences:

1. **False spinner:** a session `{is_streaming:false, active_stream_id:"dead-stream"}` now shows the running spinner on its visible row — and because the definition changed globally, this fires for *every* row, not just hidden lineage children (wider blast radius than the PR title).
2. **False unread (the #856 surface):** `_markPollingCompletionUnreadTransitions` records the stale row as observed-streaming and persists it to localStorage (`_rememberObservedStreamingSession`). When server repair later clears `active_stream_id`, `completedPersistedObservedStream = observedStreaming && !isStreaming` fires → `_markSessionCompletionUnread(sid)` even though the session produced no output. Because it's in localStorage, the false unread **survives reload/bfcache**.
3. **Multiple hidden children OR together** (`bubbleSidebarState`), so a single stale/dead-stream child poisons the parent row.

The updated #856/#2454 tests relax the old exact-definition assertion to the broader set and add a happy-path `active_stream_id → streaming:true` test, but **none exercise the `active_stream_id` set + `is_streaming` false case** — so they codify the regression instead of guarding against it.

**Fix (small, localized):**
- Drop the raw `active_stream_id` term from `_isSessionEffectivelyStreaming`. The server already folds the *live* `active_stream_id` into `is_streaming`, so the term is redundant when correct and harmful when stale. Keep `is_streaming` + the pending-user-message markers + the active-session-local check.
- `pending_started_at` is also used unreconciled — either confirm the backend clears it on abnormal termination, or gate it the same way.
- Add a regression test for the stale case: `{is_streaming:false, active_stream_id:"dead"}` must NOT be effectively-streaming (and must not produce a completion-unread transition), at both the row level and bubbled to the parent.

Everything else (ordering, pin/unpin, cleanup, the bubbling itself) is fine — once the streaming definition stops trusting the unreconciled `active_stream_id`, I'm happy to re-gate and ship.

santastabber:
Thanks for the review — updated in `75d6162f`.

Addressed the requested blocker and earlier hardening notes:

- Rebased cleanly onto current `origin/master` (`v0.51.509`).
- Removed raw `active_stream_id` and standalone `pending_started_at` from `_isSessionEffectivelyStreaming`; effective streaming now relies on server-confirmed `is_streaming`, explicit pending-message markers, or the active local `S.busy` path.
- Guarded the polling/unread transition so a stale/dead `active_stream_id` does not create a false unread marker.
- Added regressions for stale `active_stream_id` false-spinner / false-unread / parent-bubbling cases.
- Hardened hidden lineage bubbling to require archived reference rows, fall back through parent id when `_lineage_root_id` is absent, and use clearer timestamp extraction.
- Removed the accidental PR-body scratch file from the branch and updated the PR body.

Local focused verification before push:

```text
node -c static/sessions.js
python -m pytest tests/test_issue2454_active_session_spinner.py tests/test_issue856_background_completion_unread.py tests/test_session_lineage_collapse.py tests/test_sidebar_pin_immediate_reorder.py -q
# 73 passed, 1 warning
```

PR readback showed the branch is now mergeable and CI has started.

nesquena-hermes:
## Summary

Pulled the branch into a read-only worktree at HEAD `75d6162f` (the rebased single commit) and diffed against `origin/master`. The lineage-bubbling work and the pin-cache immediate-render path look solid. But the `completedPersistedObservedStream` issue Greptile flagged is **still live at HEAD** — and it's worse than a dead branch: this PR actively *regresses* a line that was correct on master.

## Code reference — this is a regression, not a pre-existing dead path

The diff at `static/sessions.js` changes the term (diff hunk `@@ -614,7 +620,7 @@`):

```diff
-    const completedPersistedObservedStream = Boolean(observedStreaming && !isStreaming);
+    const completedPersistedObservedStream = Boolean(observedStreaming && hasServerRunSignal && !isStreaming);
```

On `origin/master` the condition was `observedStreaming && !isStreaming` — correct. This PR adds `&& hasServerRunSignal`, and that turns it into a logical contradiction that can never be true. Tracing the two helpers at HEAD:

```js
// :370
function _isSessionEffectivelyStreaming(s) {
  return Boolean(s && (s.is_streaming || _hasPendingUserMessageSignal(s) || _isSessionLocallyStreaming(s)));
}
// :612
const hasServerRunSignal = Boolean(s.is_streaming || _hasPendingUserMessageSignal(s));
```

So `isStreaming === (hasServerRunSignal || _isSessionLocallyStreaming(s))`, which means **`hasServerRunSignal === true` always implies `isStreaming === true`**. Therefore `hasServerRunSignal && !isStreaming` is unsatisfiable, and `completedPersistedObservedStream` is permanently `false` (`static/sessions.js:623`).

## Why it matters

`observedStreaming` is the only one of the three completion signals that survives a page refresh — it's read from `localStorage` (`SESSION_OBSERVED_STREAMING_KEY`, `_getSessionObservedStreaming():301`), whereas `wasStreaming` (`_sessionStreamingById`) and `previousSnapshot` (`_sessionListSnapshotById`) are in-memory and reset to empty on reload. So this term is exactly the path that marks a background session unread when:

1. the session was streaming before a refresh (persisted to localStorage),
2. the user reloads,
3. the first poll after reload shows the server has finished (`!isStreaming`) with no `message_count` delta the snapshot path would catch.

`completedObservedStream` can't cover that case (it needs `wasStreaming === true`, gone after refresh) and `completedWithNewMessages` needs a count/timestamp delta vs a snapshot that no longer exists. Net effect: this PR silently breaks post-refresh unread marking that works on master today.

## Recommendation

Revert that one term to its master form (`static/sessions.js:623`):

```js
const completedPersistedObservedStream = Boolean(observedStreaming && !isStreaming);
```

`!isStreaming` already implies `!hasServerRunSignal`, so the dropped clause was strictly redundant in the satisfiable direction and harmful in the other. The `_forgetObservedStreamingSession(sid)` call at `:636` still fires right after, so it marks unread exactly once and then clears the localStorage entry — no repeat-fire risk.

## Test plan

The current suite passes because none of the four files exercise the post-refresh runtime path (Greptile's note on `test_issue856_background_completion_unread.py` is right — the new assertions check source strings, not behavior). Recommend a regression that: seeds `localStorage[SESSION_OBSERVED_STREAMING_KEY]` for a sid, clears `_sessionStreamingById` / `_sessionListSnapshotById` to simulate a reload, feeds a poll row with `is_streaming:false` and an unchanged `message_count`, and asserts `_markSessionCompletionUnread` is called once. That test fails at HEAD and passes after the one-line revert.

The rest of the PR (hidden archived lineage bubbling, the `_isServerIdleSessionRow` stale-stream guard at `:382`, immediate pin cache mutation) reads correctly to me — this single line is the blocker.

santastabber:
Updated again in response to the latest review.

What changed:
- Rebased onto current `origin/master` (`v0.51.513`).
- Restored `completedPersistedObservedStream` to the master behavior:
  ```js
  Boolean(observedStreaming && !isStreaming)
  ```
- Added a runtime regression for the post-refresh persisted observed-stream path: seed observed streaming state, clear in-memory tracking/snapshots, feed an idle poll row, and assert unread is marked once and the observed marker is cleared.
- Preserved current `CHANGELOG.md` release entries; the diff now only adds the PR’s Unreleased note.
- Re-ran delegate verification on the exact local fix; verdict: `PASS`.

Verification before push:

```text
node -c static/sessions.js
python -m pytest tests/test_issue2454_active_session_spinner.py tests/test_issue856_background_completion_unread.py tests/test_session_lineage_collapse.py tests/test_sidebar_pin_immediate_reorder.py -q
# 74 passed, 1 warning
```

Sanitization scan stayed clean.

nesquena-hermes:
## Follow-up review — regression resolved at HEAD `509e312c`

Re-pulled the branch into a read-only worktree (now at `509e312c`, rebased onto `v0.51.513`) and re-diffed `static/sessions.js` against `origin/master`. The `completedPersistedObservedStream` regression I flagged last round is fixed — the term is back to the master definition.

## Code reference

`static/sessions.js:623` now reads exactly what master has:

```js
const completedPersistedObservedStream = Boolean(observedStreaming && !isStreaming);
```

`git show origin/master:static/sessions.js` has the identical expression at its line 617, so the active-streaming-only narrowing that broke the persisted-observed path is gone. The new `completedObservedStream` guard is the right shape too (`static/sessions.js:612-614`):

```js
const hasServerRunSignal=Boolean(s.is_streaming||_hasPendingUserMessageSignal(s));
const canMarkCompletedStream=Boolean(hasServerRunSignal||previousSnapshot||observedStreaming);
const completedObservedStream = canMarkCompletedStream&&wasStreaming === true && !isStreaming;
```

This only fires the polling-fallback unread on a real observed→stopped transition that had a server run signal or a prior snapshot, so a stale `wasStreaming===true` with no corroboration can't synthesize a false unread.

## On the effective-streaming change

`_isSessionEffectivelyStreaming` (`static/sessions.js:370-381`) now folds in `_hasPendingUserMessageSignal(s)` alongside `s.is_streaming` and `_isSessionLocallyStreaming(s)`, and deliberately drops the raw `active_stream_id` / `pending_started_at` signals — which is what avoids the stale-`active_stream_id` false spinner. `_isServerIdleSessionRow` (`static/sessions.js:382`) correctly adds `!s.pending_started_at` to the idle predicate so the two stay consistent.

## Test plan / verification

The two new runtime regressions in `tests/test_issue856_background_completion_unread.py` cover both directions and are the right kind of test (they execute the actual extracted function bodies via `node -e`, not just source-text asserts):

- `test_polling_transition_ignores_stale_active_stream_id_without_server_streaming` — seeds `_sessionStreamingById['stale']=true`, feeds an idle row with a dead `active_stream_id`, asserts `{"unread": [], "observed": false}`.
- `test_polling_transition_marks_persisted_observed_stream_after_reload` — seeds observed streaming, clears in-memory tracking, feeds an idle poll row, asserts `{"unread": ["done"], "observed": {}, "streaming": false}`.

The second one is exactly the post-refresh persisted path the prior review was worried about, and it asserts the marker is consumed (`observed` empties) so it can't re-fire. CHANGELOG is also down to a single Unreleased entry now (`CHANGELOG.md:4-8`), resolving the earlier double-entry note. From a static read this looks ready; I haven't run the full suite on the worktree.

santastabber:
Rebased again onto current `origin/master` (`v0.51.520`) after the base moved. No functional changes beyond preserving the existing PR diff on the new base.

Re-verified before push:

```text
node -c static/sessions.js
python -m pytest tests/test_issue2454_active_session_spinner.py tests/test_issue856_background_completion_unread.py tests/test_session_lineage_collapse.py tests/test_sidebar_pin_immediate_reorder.py -q
# 74 passed, 1 warning
```

Sanitization scan stayed clean, and `CHANGELOG.md` still only adds the single Unreleased note for this PR while preserving all release entries.

santastabber:
The latest code/test matrix is green except for `browser-smoke`, which appears to be a transient resource-load smoke failure rather than a product regression.

Evidence:
- All pytest shards and lint passed on PR head `4ed8ecfa`.
- `browser-smoke` failed only on `/#sessions` with three generic console resource-load errors:
  ```text
  [/#sessions] console: Failed to load resource: net::ERR_FAILED
  ```
  `/` and `/#settings` both loaded cleanly.
- I reproduced `tests/browser_smoke.py` locally on the PR-head worktree with Playwright installed:
  ```text
  OK  / — no console errors
  OK  /#settings — no console errors
  OK  /#sessions — no console errors
  BROWSER SMOKE PASSED — all pages loaded with zero console errors
  ```
- Recent `master` browser-smoke runs are passing.

I attempted to rerun the failed job with `gh run rerun 27847385466 --failed`, but GitHub rejected it: `Must have admin rights to Repository`.

Could someone with permissions rerun the browser-smoke workflow/check?

santastabber:
Did small change to force checks again. Green now

nesquena:
@santastabber can you email me? I'd like to add you to our dev discord

nesquena-hermes:
Shipped in v0.51.601 (Release VH) — thank you @santastabber. The re-push fixed the regression exactly: effective-streaming no longer trusts the raw active_stream_id, so a dead-but-uncleared stream can't trigger a false spinner or a false unread that survives reload, and your new regression test locks the stale-stream case. Codex SAFE, full suite 10183 green, verified live in prod.

PR #4461, opened 2026-06-19
feat: connect WebUI to hermes-agent and automate editable install

wosolcode:
## Summary

This PR adds the configuration and startup improvements needed to connect the Hermes WebUI to an existing `hermes-agent` Docker container and ensure `AIAgent` is always importable.

### Changes

**`docker-compose.yml`**
- Add external network (`hermes-agent-avxb_default`) so the WebUI container can reach the agent gateway at `hermes-agent:8642`
- Set `HERMES_WEBUI_AGENT_DIR` env var for explicit agent path resolution (fixes `AIAgent not available` error)
- Wire `HERMES_WEBUI_PASSWORD` from `.env` file
- Change port binding from `127.0.0.1:8787` to `0.0.0.0:8787` for remote access (requires password)

**`docker_init.bash`**
- Add editable install block that runs on **every** container startup, outside the `.deps_installed` sentinel guard
- Ensures `AIAgent` is always importable even after the regular (non-editable) install from the first-run deps phase
- Detects agent source at `/home/hermeswebui/.hermes/hermes-agent` or `/opt/hermes` — no hardcoded paths

### Problem solved

The regular `uv pip install` in the deps phase installs hermes-agent as a package but `AIAgent` was not importable, producing:
```
Error: AIAgent not available -- check that hermes-agent is on sys.path
```
The editable install (`uv pip install -e`) makes the source directory directly importable, resolving this permanently.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,9 +15,7 @@ services:
   hermes-webui:
     build: .
     ports:
-    # select only one; use 127.0.0.1 version to expose to localhost only
-      - "127.0.0.1:8787:8787"
-#      - "8787:8787"
+      - "8787:8787"
     volumes:
       # Mount your Hermes home directory into the container.
       # The default (${HOME}/.hermes) works on both macOS (/Users/<you>/.hermes)
@@ -40,10 +38,11 @@ services:
       - HERMES_WEBUI_PORT=8787
       # Where to store sessions, workspaces, and other state (default: ~/.hermes/webui)
       - HERMES_WEBUI_STATE_DIR=/home/hermeswebui/.hermes/webui
+      - HERMES_WEBUI_AGENT_DIR=/home/hermeswebui/.hermes/hermes-agent
       # Default workspace directory shown on first launch
 #      - HERMES_WEBUI_DEFAULT_WORKSPACE=/workspace
       # Optional: set a password for remote access
-      # - HERMES_WEBUI_PASSWORD=your-secret-password
+      - HERMES_WEBUI_PASSWORD=${HERMES_WEBUI_PASSWORD}
       #
       # Bind-mount permission handling (fixes #1389, #1399):
       #   When you mount an EXISTING ~/.hermes directory (the common case),
@@ -55,3 +54,10 @@ services:
       # - HERMES_SKIP_CHMOD=1
       # - HERMES_HOME_MODE=0640
     restart: unless-stopped
+    networks:
+      - hermes-net
+
+networks:
+  hermes-net:
+    external: true
+    name: hermes-agent-avxb_default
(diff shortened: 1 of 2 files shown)


nesquena-hermes (changes requested):
Thanks for working on the AIAgent-importability problem — the `docker_init.bash` editable-install idea is the right instinct. But the `docker-compose.yml` changes here look like they carry your local environment into the PR, and a couple of them break the default quickstart for everyone else, so this needs a revision before it can land.

## Blocking: machine-specific external network breaks `docker compose up` for all users

```yaml
networks:
  hermes-net:
    external: true
    name: hermes-agent-avxb_default
```

`hermes-agent-avxb_default` is a Compose-generated network name from your own machine (the `avxb` segment is your local project hash), and `external: true` means Docker won't create it — it must already exist. Every other user running `docker compose up` will get `network hermes-agent-avxb_default declared as external, but could not be found` and the stack won't start. The default quickstart can't depend on a network that only exists on one machine.

If the goal is to let the WebUI container reach a separately-running `hermes-agent` container, that needs to be opt-in and documented (e.g. a commented-out `networks:` block users uncomment + set their own network name), not the hardcoded default.

## Blocking: reverts the localhost-only port bind (unauthenticated exposure)

```diff
-      - "127.0.0.1:8787:8787"
-#      - "8787:8787"
+      - "8787:8787"
```

Master deliberately defaults to `127.0.0.1:8787:8787` (localhost-only). Switching to `0.0.0.0:8787` exposes the WebUI on every host interface, and the compose file passes `${HERMES_WEBUI_PASSWORD}` which resolves to an empty string when it isn't set in `.env` — i.e. reachable on the network with no auth. Please keep the `127.0.0.1` default (the commented all-interfaces line is already there for users who knowingly opt in).

Same for `HERMES_WEBUI_PASSWORD` — leave it commented as an opt-in example rather than wiring it on by default.

## Should-fix: editable-install block ignores the AGENT_DIR env var it sets

The compose file adds `HERMES_WEBUI_AGENT_DIR=/home/hermeswebui/.hermes/hermes-agent`, but the new `docker_init.bash` block hardcodes its own search list (`/home/hermeswebui/.hermes/hermes-agent`, `/opt/hermes`) instead of honoring that env var. They happen to agree today, but they'll silently diverge if anyone overrides `HERMES_WEBUI_AGENT_DIR`. Have the block prefer `$HERMES_WEBUI_AGENT_DIR` when set, then fall back to the search list.

## Suggested path

Split this into just the editable-install fix: keep the `docker_init.bash` block (honoring `HERMES_WEBUI_AGENT_DIR`), drop all the `docker-compose.yml` changes except — if you want — a commented-out opt-in network block. That gives a clean, general-purpose fix for the AIAgent-import problem without the machine-specific config or the security regression.

Happy to re-review once the compose file is back to the safe defaults.

PR #4468, opened 2026-06-19
fix(scroll): render virtual window during programmatic scrolls

rodboev:
## Summary

Fixes the scroll-down blank reported on Discord after #4434 landed:

> so the getting stuck seems fixed, will test a few more large sessions, but I do notice when scrolling back down (after scrolling up a bunch), it starts not displaying text on the way down

### Reproduction

Injected a 400+ message session with dense tool blocks (35–45 consecutive tool calls each) via CDP, enabled virtualization, scrolled to the top, then scrolled back down. Instrumented `_messageVirtualWindow` and the DOM to compare the computed virtual window against the actual spacer/row state on each scroll step.

The virtual window computes correctly (start and topPad advance as scrollTop increases), but the DOM never updates — the spacer height and firstRow stay frozen:

```
step 0:  scrollTop=300   computed topPad=0     actual spacer=3760  firstRow=27
step 9:  scrollTop=3280  computed topPad=2080  actual spacer=3760  firstRow=27
step 18: scrollTop=5700  computed topPad=4761  actual spacer=3760  firstRow=27
```

Further instrumentation revealed `_programmaticScroll` is stuck at `true` after the scroll-up phase settles. The scroll handler's `if(_programmaticScroll) return` guard (line 3660) silently drops every scroll event, so `_scheduleMessageVirtualizedRender()` never fires and the DOM freezes.

The flag gets stuck because `_compensateScrollForMeasurementDelta` sets `_programmaticScroll=true` and clears it via `rAF→setTimeout`. During rapid measurement refresh cycles at the scroll ceiling (measurement → re-render → measurement, capped at `MESSAGE_VIRTUAL_MEASUREMENT_MAX_RERENDERS=2`), each cycle sets the flag before the previous cleanup fires. The last cleanup races with the next measurement rAF, leaving the flag permanently set.

### Fix

Moves `_scheduleMessageVirtualizedRender()` before the `_programmaticScroll` guard in the scroll event handler. The virtual window always updates regardless of scroll origin. The guard still protects pin/unpin state tracking (its intended purpose) — only the render scheduling moves above it.

## Test plan

- [x] Scroll up through 400+ message session with dense tool blocks → reach top
- [x] Scroll back down → text renders at every scroll position (zero blank viewports, down from 13)
- [x] Verified across real CLI sessions (96–256 messages, up to 92 tool blocks)
- [x] `test_issue500_message_list_virtualization.py` — 29/29 pass

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -3657,6 +3657,7 @@ if(typeof window!=='undefined'){
   if(!el) return;
   let _scrollRaf=0;
   el.addEventListener('scroll',()=>{
+    _scheduleMessageVirtualizedRender();
     if(_programmaticScroll) return; // ignore scrolls we triggered ourselves
     _markMessageVirtualScrollActive();
     cancelAnimationFrame(_scrollRaf);
@@ -3693,7 +3694,6 @@ if(typeof window!=='undefined'){
       const showBottomButton=!_scrollPinned && el.scrollHeight-top-el.clientHeight>80;
       _syncScrollToBottomCue(showBottomButton,{newMessage:_newMessageCueVisible});
       if(typeof _updateSessionStartJumpButton==='function') _updateSessionStartJumpButton();
-      _scheduleMessageVirtualizedRender();
       // Prefetch older messages before the reader hits the hard top. Prepending
       // then preserving scrollTop is seamless only if there is runway left for
       // the user's continued upward wheel/touch movement.


nesquena-hermes:
Deep-reviewed this (it fixes a regression from #4434, which I shipped as v0.51.500). Ran the full gate — **Codex SAFE + Opus SAFE + browser-smoke clean + full suite 9539 passed** — and the diagnosis + fix hold up. Per our convention for the message-virtualization series I'm **not merging yet** — flagging @rodboev for your go-ahead before the final merge.

### What I verified
- **Feedback-loop safety (the main risk of scheduling render before the guard): provably safe.** `_compensateScrollForMeasurementDelta` holds the anchor row at the same viewport offset, so the visible-row set is identical before/after → `nextKey === _messageVirtualWindowKey` → `_scheduleMessageVirtualizedRender`'s first guard returns early and breaks the scroll→render→scroll cycle. The `<2px` no-op threshold, the `_messageVirtualScrollRaf` coalescing, the in-rAF re-check, and the `MESSAGE_VIRTUAL_MEASUREMENT_MAX_RERENDERS=2` cap are all additional margins. It can't thrash.
- **Reorder is inert:** the scheduler reads/writes none of the state from `_syncScrollToBottomCue` / `_updateSessionStartJumpButton` / the pin block / prefetch; it only defers its own rAF render. Moving it to the top of the handler (and out of `_scrollRaf`) is harmless, and it now also refreshes the window on legitimate programmatic scrolls (jump-to-bottom/start), which is desirable.
- **No #4434 regression:** #4434's fix lives entirely in `_compensateScrollForMeasurementDelta` (the `scrollTopBefore<1` early-return); this PR doesn't touch that function.

### One note (non-blocking) — this is a mask, not a root-cause fix
The PR says the guard "still protects pin/unpin state tracking." But if `_programmaticScroll` is genuinely stuck `true`, everything still below the guard is **frozen, not protected** — pin/unpin tracking, the scroll-to-bottom cue, the session-start jump button, and prefetch-older all share the same stuck-flag root cause; this change only unfreezes rendering. (Static read suggests the set-true/clear-via-rAF are paired and the refresh loop is capped at 2, so the flag likely self-heals within ~1–2 frames and the pin/prefetch impairment is transient rather than permanent — I couldn't confirm permanent-vs-transient without driving the browser.) Either way this PR is a strict improvement and makes none of it worse.

Suggested follow-up (separate PR): clear the root cause — e.g. a depth counter / generation token instead of the boolean, or reset `_programmaticScroll=false` on a confirmed user-origin scroll — so pin/prefetch/cue fully recover instead of relying on the rAF clear winning the race. Happy to merge this as the immediate render fix on your word.

rodboev:
The flag doesn't get permanently stuck — the measurement cycle is hard-capped at 2 rerenders, each schedules a rAF→setTimeout clear, and those clears always fire. The stuck state is transient (1–2 frames during measurement compensation), not permanent.

So this isn't a mask over a deeper problem; it's the actual fix. `_scheduleMessageVirtualizedRender` should never have been behind the `_programmaticScroll` guard — when scrollTop shifts for measurement compensation, the viewport moved and the virtual window needs to re-render. The guard is correct for everything still behind it (pin/unpin, cue, prefetch), because a programmatic scrollTop adjustment isn't a user gesture and shouldn't trigger those. A generation counter or user-origin detection would be solving a non-problem.

Good to merge.

PR #4469, opened 2026-06-19
feat: connect WebUI to a root-running hermes-agent container (Docker)

wosolcode:
## Summary

Makes the WebUI work out-of-the-box alongside an existing **root-running** `hermes-agent` Docker container that shares its `HERMES_HOME` via a bind mount. Fixes `AIAgent not available` and the follow-on `No LLM provider configured` errors in that topology.

## Commits

1. **connect WebUI to hermes-agent + automate editable install** — add external network so the WebUI reaches the gateway at `hermes-agent:8642`, set `HERMES_WEBUI_AGENT_DIR`, and run `uv pip install -e` for the agent on every startup (outside the `.deps_installed` guard) so `AIAgent` is always importable.
2. **support running WebUI as root (UID=0)** — when `WANTED_UID=0`/`WANTED_GID=0`, skip the `su hermeswebui` privilege drop. The unconditional drop re-entered the root init branch (whoami still resolves to root for uid 0), causing an infinite restart loop. Running as root lets the WebUI read shared state (e.g. `cron/jobs.json`) written `root:root` by the agent.
3. **pin HOME and HERMES_HOME** — as root, `HOME` defaults to `/root`, so the in-process agent looked for config/creds in `/root/.hermes`. Pin `HOME=/home/hermeswebui` and `HERMES_HOME=/home/hermeswebui/.hermes` so `config.yaml` + `.env` resolve from the shared mount regardless of runtime UID.

## Testing

- Container starts with no restart loop; `agent dir ... [ok]` in logs
- `from run_agent import AIAgent` succeeds in-process (verified independent of HOME / env via the editable `.pth`)
- Tasks/cron endpoint returns 200 (was 500 PermissionError)
- End-to-end: new chat reply works (model `deepseek/deepseek-v4-pro` via OpenRouter)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,9 +15,7 @@ services:
   hermes-webui:
     build: .
     ports:
-    # select only one; use 127.0.0.1 version to expose to localhost only
-      - "127.0.0.1:8787:8787"
-#      - "8787:8787"
+      - "8787:8787"
     volumes:
       # Mount your Hermes home directory into the container.
       # The default (${HOME}/.hermes) works on both macOS (/Users/<you>/.hermes)
@@ -40,10 +38,11 @@ services:
       - HERMES_WEBUI_PORT=8787
       # Where to store sessions, workspaces, and other state (default: ~/.hermes/webui)
       - HERMES_WEBUI_STATE_DIR=/home/hermeswebui/.hermes/webui
+      - HERMES_WEBUI_AGENT_DIR=/home/hermeswebui/.hermes/hermes-agent
       # Default workspace directory shown on first launch
 #      - HERMES_WEBUI_DEFAULT_WORKSPACE=/workspace
       # Optional: set a password for remote access
-      # - HERMES_WEBUI_PASSWORD=your-secret-password
+      - HERMES_WEBUI_PASSWORD=${HERMES_WEBUI_PASSWORD}
       #
       # Bind-mount permission handling (fixes #1389, #1399):
       #   When you mount an EXISTING ~/.hermes directory (the common case),
@@ -55,3 +54,10 @@ services:
       # - HERMES_SKIP_CHMOD=1
       # - HERMES_HOME_MODE=0640
     restart: unless-stopped
+    networks:
+      - hermes-net
+
+networks:
+  hermes-net:
+    external: true
+    name: hermes-agent-avxb_default
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,6 +39,11 @@ services:
       # Where to store sessions, workspaces, and other state (default: ~/.hermes/webui)
       - HERMES_WEBUI_STATE_DIR=/home/hermeswebui/.hermes/webui
       - HERMES_WEBUI_AGENT_DIR=/home/hermeswebui/.hermes/hermes-agent
+      # When running as root (UID=0) HOME defaults to /root, so the in-process
+      # agent would look for config/credentials in /root/.hermes. Pin HOME and
+      # HERMES_HOME to the shared mount so provider config + .env are found.
+      - HOME=/home/hermeswebui
+      - HERMES_HOME=/home/hermeswebui/.hermes
       # Default workspace directory shown on first launch
 #      - HERMES_WEBUI_DEFAULT_WORKSPACE=/workspace
       # Optional: set a password for remote access
--- a/docker_init.bash
+++ b/docker_init.bash
@@ -485,6 +485,24 @@ fi
 
 ensure_hindsight_client_docker_dependency
 
+# Ensure hermes-agent is installed in editable mode on every startup.
+# This runs outside the .deps_installed guard so AIAgent is always importable
+# even after a regular (non-editable) install from docker_init.bash first run.
+_editable_agent_src=""
+for _ep in "/home/hermeswebui/.hermes/hermes-agent" "/opt/hermes"; do
+    if [ -d "$_ep" ] && [ -f "$_ep/pyproject.toml" ]; then
+        _editable_agent_src="$_ep"
+        break
+    fi
+done
+if [ -n "$_editable_agent_src" ]; then
+    echo ""; echo "-- Ensuring hermes-agent editable install (required for AIAgent)..."
+    uv pip install -e "$_editable_agent_src" -q \
+        --trusted-host pypi.org --trusted-host files.pythonhosted.org || \
+        echo "!! WARNING: editable install failed -- AIAgent may be unavailable"
+fi
+
+
 echo ""; echo "== Running hermes-webui"
 cd /app; python server.py || error_exit "hermes-webui failed or exited with an error"
 
--- a/docker_init.bash
+++ b/docker_init.bash
@@ -287,9 +287,17 @@ if [ "A${whoami}" == "Aroot" ]; then
     fi
   done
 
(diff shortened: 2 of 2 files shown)


nesquena-hermes (changes requested):
Thanks for the follow-up @wosolcode — the `docker_init.bash` side of this (the `WANTED_UID=0` root-bypass to avoid the `su hermeswebui` infinite-restart loop, and the editable-install-on-every-startup so `AIAgent` stays importable) is the right idea and the root-loop reasoning is sound. But the `docker-compose.yml` changes carry the same two blockers that #4461 was bounced for, so this can't merge as-is. Both are code-verified against `origin/master` and independently flagged by Greptile.

### (must-fix 1) Security regression: default flips to all-interfaces + empty password → unauthenticated remote exposure

`docker-compose.yml` changes the port binding from master's localhost-only default to all-interfaces, and uncomments `HERMES_WEBUI_PASSWORD` so it is now *always* injected:

```yaml
# master (current default):
    ports:
    # select only one; use 127.0.0.1 version to expose to localhost only
      - "127.0.0.1:8787:8787"
#      - "8787:8787"
      # - HERMES_WEBUI_PASSWORD=your-s...word   # commented out

# this PR:
    ports:
      - "8787:8787"                              # all interfaces (0.0.0.0)
      - HERMES_WEBUI_PASSWORD=${HERM...ORD}    # always injected
```

If `HERMES_WEBUI_PASSWORD` is unset in the host environment it resolves to the **empty string**, so the default `docker compose up` now serves the WebUI on every interface with no authentication. That's a silent downgrade of the shipped security posture for everyone who copies this compose file.

**Fix:** keep `127.0.0.1:8787:8787` as the default (leave the all-interfaces line commented with the existing "select only one" note), and keep `HERMES_WEBUI_PASSWORD` commented out (or document that it must be set before exposing beyond localhost).

### (must-fix 2) Hard-coded machine-specific external network breaks `docker compose up` for everyone but you

```yaml
networks:
  hermes-net:
    external: true
    name: hermes-agent-avxb_default
```

`hermes-agent-avxb_default` is your local Compose project's auto-generated network name (the `avxb` suffix is your machine's project hash). `external: true` means Compose will not create it — so anyone who runs this file without that exact pre-existing network gets an immediate `network hermes-agent-avxb_default declared as external, but could not be found` failure.

**Fix:** make the network name configurable (e.g. `name: ${HERMES_AGENT_NETWORK:-hermes-agent_default}`), or move the cross-container wiring (external network + `HERMES_WEBUI_AGENT_DIR`) into a separate documented multi-container compose file / override, so the single-container default still works out of the box.

### (should-fix) Editable install bypasses the read-only-mount staging the existing guard uses

The new `uv pip install -e "$_editable_agent_src"` block runs directly against the agent source on every startup, outside the `.deps_installed` guard and without the read-only-mount staging the existing dependency-install path specifically added. For a multi-container setup where the agent source is a shared/read-only mount, this can regress. Consider reusing the same staging approach the existing guard uses, or gate it so it only runs when the source is writable.

The root-bypass and the "ensure editable install" intent are both worth keeping — please split this so the single-container compose default stays localhost-only + unauthenticated-by-omission-of-password (i.e. unchanged security posture), and put the multi-container/external-network bits behind an opt-in. Happy to re-review once the compose default is restored.

nesquena-hermes:
## Summary

Reviewed on a read-only worktree at HEAD `53af8e91` vs `origin/master`. The core problem you set out to solve is real and the `WANTED_UID=0` root-bypass in `docker_init.bash` is the right shape — the unconditional `exec su hermeswebui` genuinely does re-enter the root branch for uid 0 (because `whoami` still resolves to root) and loops. That fix is sound. But three of the changes regress existing single-container and multi-container users, and two of them are security regressions on the default install. I'd hold merge until these are split out and gated.

## 1. Port + password: default install becomes unauthenticated on the network

In `docker-compose.yml` the diff drops the localhost binding (`127.0.0.1:8787:8787` becomes `8787:8787`) and uncomments the password line so it's now unconditionally injected as `HERMES_WEBUI_PASSWORD` set to the host env value.

The default committed file currently binds `127.0.0.1` precisely so a fresh `docker compose up` is not reachable off-box. Switching to `8787:8787` exposes it on all interfaces, and the injected password variable resolves to an empty string when unset — so unless the app rejects empty passwords, the default topology is now an open WebUI on the LAN. This is a behavior change for every existing user, not just the root-agent topology you're targeting. Please keep the localhost default in the committed file and leave the all-interfaces line commented, the way it was.

## 2. Hard-coded external network name breaks `docker compose up` for everyone else

```yaml
networks:
  hermes-net:
    external: true
    name: hermes-agent-avxb_default
```

`hermes-agent-avxb_default` is your local Compose project's auto-generated network (the `-avxb` suffix is deployment-specific). For any other user, `external: true` with a name that doesn't exist makes Compose fail immediately with "network not found". A single-container user who just wants the WebUI now can't start it at all. This needs to be either env-driven (e.g. `name: ${HERMES_AGENT_NETWORK:-...}`) or — better — moved out of the primary `docker-compose.yml` into a dedicated multi-container compose file, since the network and `HERMES_WEBUI_AGENT_DIR` are multi-container concerns.

## 3. Editable install bypasses the existing read-only-mount staging

The new block runs `uv pip install -e "$_editable_agent_src"` directly against the mounted source (in `docker_init.bash`, right after `ensure_hindsight_client_docker_dependency`). master already solved agent install for `:ro` mounts — the existing block at `docker_init.bash:434-470` stages the source into a writable `/tmp/hermes-agent-build` copy first, with this explicit rationale:

```bash
# setuptools' `egg_info` build step ... touches `hermes_agent.egg-info/`
# inside the source tree even under PEP 517 build isolation, which `EROFS`-fails
# on a `:ro` mount and (under `set -e`) kills startup of every multi-container deploy.
```

An editable install is worse here: `-e` writes the `.pth` and `egg-info` into the source tree by design, so on the read-only agent mount that the multi-container compose files (and `docs/rfcs/agent-source-boundary.md`) deliberately use, this either EROFS-fails or forces users to make the agent source writable — which is exactly the defence-in-depth boundary that RFC exists to protect. If the goal is "AIAgent always importable," extend the existing staged install to also cover the every-startup case rather than adding a second, un-staged install path.

## Recommendation

Split this into (a) the `WANTED_UID=0` root-bypass — that one is genuinely useful and low-risk, ship it on its own — and (b) a separate multi-container topology PR that puts the external network, `HERMES_WEBUI_AGENT_DIR`, and `HOME`/`HERMES_HOME` pinning into a dedicated compose file, leaving the committed single-container `docker-compose.yml` on its localhost-only, password-optional defaults. The editable-install need should reuse the `/tmp/hermes-agent-build` staging path so `:ro` mounts keep working. As-is, Greptile's 2/5 is fair: the network name and port/password changes break or expose the default install.

nesquena-hermes:
Thanks @wosolcode — closing this one. The connectivity problem (WebUI alongside an existing agent container sharing HERMES_HOME) is real, but the specific path here is built around a **root-running** hermes-agent container (the `WANTED_UID=0` root-bypass), and we don't want to bless running the agent as root as a supported topology — it's a security posture we'd rather steer away from, not add first-class support for. 

If there's a non-root way to share HERMES_HOME between a separate agent container and the WebUI, that'd be a more mergeable shape. Closing as a security/scope decision — appreciate the clear writeup of the failure mode.

PR #4482, opened 2026-06-19
Fix post-start live attach after UI bookkeeping errors

### LINKED ISSUE #4481: bug(streaming): post-start UI errors can block live stream attach after chat start succeeds
## Problem

`send()` currently wraps both the `/api/chat/start` request and the post-start browser UI/bookkeeping work in the same `try/catch`.

That means a synchronous browser-side exception after `/api/chat/start` succeeds can be misreported as a send failure even though the backend run has already started. In that state the current tab can append an assistant error, clear busy state, and return before `attachLiveStream(...)` runs, leaving the user with no live stream until reload/session re-entry catches up from server state.

Observed locally on `8787`:

- Browser showed `Error: Maximum call stack size exceeded` immediately after sending.
- `/api/chat/start` had already succeeded and the backend run was active.
- The current tab did not attach the live stream.
- Switching away/back eventually rehydrated the active run from server state.

## Current root cause on master

This is a mainline error-boundary issue:

- `static/messages.js` calls `api('/api/chat/start', ...)`.
- The same `try/catch` then runs optional post-start UI work such as model UI sync, pending/live indicators, inflight persistence, session-list refresh, and live status setup.
- The catch path handles any thrown exception as if chat start failed, pushes an assistant `**Error:** ...`, clears optimistic/inflight state, and returns before `attachLiveStream(...)`.

## Expected behavior

After `/api/chat/start` succeeds and returns a `stream_id`, the frontend should treat live attach as mandatory. Post-start UI/bookkeeping failures should be logged as recoverable and should not prevent `attachLiveStream(activeSid, streamId, uploadedNames)`.

## Acceptance criteria

- Split the `/api/chat/start` request failure boundary from post-start UI/bookkeeping failures.
- Preserve existing handling for real chat-start failures, including 404 session deletion and active-stream conflict queueing.
- Once a `stream_id` is available, `attachLiveStream(...)` should still run even if optional UI setup throws.
- Add a regression/invariant test that would fail on the old structure.
- Keep the fix independent and focused on the mainline error-boundary bug.

franksong2702:
## Thinking Path

- Browser turns depend on `/api/chat/start` returning a `stream_id`, then `attachLiveStream(...)` taking over SSE rendering.
- The old `send()` boundary wrapped both the API request and post-start browser bookkeeping in one `try/catch`.
- That made a browser-only exception after a successful start look like a failed send, even though the backend run was already active.
- The fix is to keep real chat-start failures terminal, but make post-start UI/bookkeeping best-effort once `stream_id` exists.

## What Changed

- Added `_runOptionalPostStartUiStep(...)` for recoverable post-start browser bookkeeping failures.
- Narrowed the `/api/chat/start` `try/catch` so it captures the API request and preserves the existing real-failure handling paths.
- Moved stream-id setup, Stop-button refresh, post-start UI/bookkeeping, and `attachLiveStream(activeSid, streamId, uploadedNames)` after the API catch succeeds.
- Kept the Stop affordance refresh outside optional post-start bookkeeping so optional UI errors cannot skip the primary button update.
- Added regression coverage proving post-start bookkeeping runs after the API catch and cannot append a false assistant `**Error:**` before live attach.
- Updated the existing busy-button invariant test for the new split-boundary order.
- Added a release-note entry for #4481.

## Why It Matters

If browser-side UI bookkeeping throws after the backend run has started, the current tab should not claim the send failed or detach from the active run. This keeps live streaming visible and avoids requiring a session switch or reload to discover that the run is actually active.

## Verification

- `git diff --check -- static/messages.js tests/test_inflight_send_start_race.py tests/test_1062_busy_input_modes.py tests/test_turn_duration_display.py CHANGELOG.md`
- `./scripts/test.sh tests/test_inflight_send_start_race.py tests/test_session_switch_busy_race.py tests/test_turn_duration_display.py tests/test_1062_busy_input_modes.py -q`
- `./scripts/test.sh tests/test_chat_start_provider_fallback.py tests/test_sidebar_first_turn_visibility.py tests/test_stale_empty_session_restore.py tests/test_status_command_card.py -q`

## Risks / Follow-ups

- Post-start UI/bookkeeping failures are now warning-only once chat start succeeds; this is intentional so live attach remains authoritative.
- `attachLiveStream(...)` itself is not wrapped by this PR. If that function throws, that is a separate live-stream attachment failure path.

Closes #4481.

## Model Used

AI-assisted with OpenAI GPT-5 Codex as coordinator and `gpt-5.3-codex-spark` as the implementation worker, using local shell inspection, targeted tests, and coordinator review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Post-start streaming errors in the browser no longer hide active streams after a successful `/api/chat/start` (#4481).** If browser-side bookkeeping after stream start throws, it is now logged as recoverable and `attachLiveStream(activeSid, streamId, uploadedNames)` still runs when `stream_id` is available, so a transient UI exception cannot clear the send state or prevent live tokens from attaching.
+
 ## [v0.51.513] — 2026-06-19 — Release RX (credential-pool quota status for all pooled providers)
 
 ### Added
--- a/static/messages.js
+++ b/static/messages.js
@@ -898,6 +898,16 @@ function _runOptionalPreStartUiStep(label, fn){
   }
 }
 
+function _runOptionalPostStartUiStep(label, fn){
+  try{
+    return typeof fn==='function'?fn():undefined;
+  }catch(e){
+    const message=e&&e.message?e.message:String(e||'unknown error');
+    try{console.warn('[webui] optional post-start UI step failed', label, message);}catch(_){ }
+    return undefined;
+  }
+}
+
 function _sessionTitleLooksDefaultOrProvisional(titleText, provisionalText){
   const title=String(titleText||'').replace(/\s+/g,' ').trim();
   if(!title||title==='Untitled'||title==='New Chat')return true;
@@ -1248,8 +1258,12 @@ async function send(){
 
   // Start the agent via POST, get a stream_id back
   let streamId;
+  let postStartData;
+  let modelStateForPostStart;
+  let explicitPickForPostStart;
   try{
     const _modelState=_chatPayloadModelState();
+    modelStateForPostStart=_modelState;
     const _pendingPick=(typeof _readPendingSessionModel==='function')
       ? _readPendingSessionModel(activeSid)
       : null;
@@ -1262,6 +1276,7 @@ async function send(){
     // once read so a later send of an unchanged dropdown isn't treated as an explicit
     // pick. (#3739/#3737, Codex catch)
     if(_explicitPick && typeof _clearPendingSessionModel==='function') _clearPendingSessionModel(activeSid);
+    explicitPickForPostStart=_explicitPick;
     const startData=await api('/api/chat/start',{method:'POST',body:JSON.stringify({
       session_id:activeSid,message:msgText,
       // S.session.model remains authoritative; the helper only resolves a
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
## Summary

Pulled this into a read-only worktree at HEAD `36e99241` and diffed `static/messages.js` against `origin/master`. The error-boundary split is correct: every `catch` branch returns before control reaches the post-start block, so the only way to fall through is a successful `/api/chat/start`. Once that succeeds, stream identity and the Stop affordance are set unconditionally, and the bookkeeping is moved into a best-effort wrapper. This matches the acceptance criteria in #4481.

## Code reference

The key structural change is that `S.activeStreamId` and `updateSendBtn()` are hoisted **out** of the optional wrapper, so a bookkeeping throw can't strand the Stop button or the stream id (`static/messages.js:1354-1361`):

```js
const startData = postStartData || {};
streamId = postStartData ? postStartData.stream_id : null;
S.activeStreamId = streamId;
// ... refresh now that we have a stream id so the primary button can switch to Stop
if(typeof updateSendBtn==='function') updateSendBtn();
_runOptionalPostStartUiStep('post-start ui/bookkeeping', ()=>{
  // applySessionTitleUpdate, model sync, appendThinking, markInflight,
  // saveInflightState, renderSessionList ...
});

// Open SSE stream and render tokens live
attachLiveStream(activeSid, streamId, uploadedNames);
```

The wrapper itself only warns (`static/messages.js:901-907`), so a post-start exception can no longer be misreported as a send failure or skip `attachLiveStream(...)`.

## Diagnosis / Recommendation

The fix is sound and the 404 / "session already has an active stream" conflict paths are preserved verbatim inside the narrowed `catch` (`static/messages.js:1298-1320`), which was the main regression risk. One trade-off worth noting (Greptile flagged the same): `_runOptionalPostStartUiStep` coalesces **all** bookkeeping into one lambda, so an early throw (e.g. in `applySessionTitleUpdate` or the model-sync block) silently skips `markInflight` + `saveInflightState` later in the same lambda. That leaves no reconnect snapshot in localStorage if the tab reloads mid-stream — the live attach still works in the current tab, but a reload would fall back to server-state rehydration rather than the inflight cache.

That's an acceptable trade-off for this PR's stated scope (live attach stays authoritative), but if you want belt-and-suspenders, splitting `markInflight`/`saveInflightState` into their own `_runOptionalPostStartUiStep` call would isolate the persistence step from the cosmetic title/model steps:

```js
_runOptionalPostStartUiStep('post-start ui sync', ()=>{ /* title + model + thinking */ });
_runOptionalPostStartUiStep('post-start inflight persist', ()=>{ /* markInflight + saveInflightState */ });
```

## Test plan

`tests/test_inflight_send_start_race.py::test_post_start_bookkeeping_errors_cannot_block_live_attach` asserts the source-text ordering `catch_idx < stream_id_idx < optional_idx < attach_idx` and that no `**Error:**` push exists between the optional wrapper and live attach — which would have failed on the old single-boundary structure. `tests/test_1062_busy_input_modes.py` was updated to assert `updateSendBtn()` runs after `S.activeStreamId` is assigned and before both the optional wrapper and `attachLiveStream`. Both correctly encode the new invariant.

nesquena-hermes:
Shipped in **v0.51.514** (Release RY) — thanks @franksong2702. The post-`/api/chat/start` UI bookkeeping is now outside the request `try{}`, and `attachLiveStream` always runs when the server returns a `stream_id`, so a transient browser-side UI exception can no longer be misreported as a send failure or hide an already-running stream.

Gate: Codex SAFE + Opus SAFE + full pytest suite green (9593 passed) + CI all shards green.

PR #4484, opened 2026-06-19
feat(kanban): faithful dashboard layout + task detail in workspace panel

autrioly:
## Summary

Adds a **Dashboard** kanban layout option that faithfully reproduces the Hermes agent dashboard plugin's visual style, plus moves task detail view into the right-side workspace panel when in dashboard mode.

### Phase 1 — Dashboard layout reproduction
- **`static/style.css`**: Full dashboard-style CSS — status dots (lilac/amber/green/red/blue), compact 280px cards, assignee badges, progress pills, hover actions, staleness indicators, hidden markdown preview.
- **`static/panels.js`** (`_kanbanCard`): Conditionally renders dashboard-style card HTML when layout is "dashboard" (monospace ID, priority badge, progress pill, assignee badge).
- **`static/panels.js`** (`_kanbanRenderBoard`): In dashboard mode, renders a single row of columns — no profile lane subdivision.
- **`static/index.html`**: Adds `settingsKanbanLayout` select to Preferences panel with early-init script to prevent FOUC.

### Phase 2 — Task detail in workspace panel
- **`static/index.html`**: Adds hidden "Task" tab + `workspaceKanbanTask` container to the workspace panel.
- **`static/workspace.js`** (`switchWorkspacePanelTab`): Supports `'kanban-task'` tab — shows/hides file tree and kanban task container.
- **`static/panels.js`** (`loadKanbanTask`): In dashboard mode, opens the workspace panel and shows task detail in the new "Task" tab instead of the preview above the board.
- **`static/panels.js`** (`_kanbanRenderTaskDetail`): Accepts `mode` param (`'preview'` | `'workspace'`) — shows "Close" button in workspace mode vs "Back to board" in preview mode.
- **`static/panels.js`** (`closeKanbanTaskDetail`): Cleans up workspace task tab and reverts to Files tab in dashboard mode.
- **`static/style.css`**: CSS rules to hide File/Artifacts/Todos tabs content when kanban-task tab is active, proper scrolling.

## How to test
1. Open Web UI → Settings → change **Kanban Layout** to "Dashboard".
2. Verify columns render as a single compact row (no profile lanes).
3. Click any task card → detail should open in the right-side workspace panel under a "Task" tab.
4. Click "Close" → task tab hides, workspace reverts to Files tab.
5. Switch layout back to "WebUI" → task detail closes, layout reverts to default.
6. In WebUI mode, clicking a task still shows the preview above the board (unchanged).

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -1468,11 +1468,13 @@
       <button class="workspace-panel-tab active" id="workspaceFilesTab" type="button" onclick="switchWorkspacePanelTab('files')" role="tab" aria-selected="true">Files</button>
       <button class="workspace-panel-tab" id="workspaceArtifactsTab" type="button" onclick="switchWorkspacePanelTab('artifacts')" role="tab" aria-selected="false">Artifacts <span id="workspaceArtifactsCount" class="workspace-artifacts-count">0</span></button>
       <button class="workspace-panel-tab" id="workspaceTodosTab" type="button" onclick="switchWorkspacePanelTab('todos')" role="tab" aria-selected="false" hidden data-i18n="tab_todos">Todos</button>
+      <button class="workspace-panel-tab" id="workspaceKanbanTaskTab" type="button" onclick="switchWorkspacePanelTab('kanban-task')" role="tab" aria-selected="false" hidden>Task</button>
     </div>
     <div class="breadcrumb-bar" id="breadcrumbBar" style="display:none"></div>
     <div class="file-tree" id="fileTree"></div>
     <div class="workspace-artifacts" id="workspaceArtifacts" hidden></div>
     <div class="workspace-todos" id="workspaceTodosPanel" hidden style="flex:1;overflow-y:auto;padding:8px 12px"></div>
+    <div class="workspace-kanban-task" id="workspaceKanbanTask" hidden style="flex:1;overflow-y:auto;padding:12px"></div>
     <div id="wsEmptyState" style="display:none;flex:1;align-items:center;justify-content:center;padding:24px 16px;text-align:center;color:var(--muted);font-size:12px;line-height:1.6"></div>
     <div class="preview-area" id="previewArea">
       <div class="preview-path" id="previewPath">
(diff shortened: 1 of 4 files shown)


nesquena-hermes (changes requested):
Thanks for this — the dashboard kanban layout and the in-panel task detail are a nice direction, and the CSS is cleanly namespaced (`html[data-kanban-layout="dashboard"]` / `.rightpanel[data-active-tab="kanban-task"]`), so the default layout is unaffected. One objective blocker before this can ship, plus a small hardening nit to bundle.

### Blocker — the task-detail header buttons render raw i18n keys instead of labels

Both new buttons in `_kanbanRenderTaskDetail` (`static/panels.js:2828-2829`) reference i18n keys that were never added to `static/i18n.js`:

```js
// panels.js:2828 (workspace mode)
`<button class="btn secondary kanban-back-btn" onclick="closeKanbanTaskDetail()">${esc(t('kanban_close') || 'Close')}</button>`
// panels.js:2829 (dashboard mode)
`<button class="btn secondary kanban-back-btn" onclick="closeKanbanTaskDetail()">${esc(t('kanban_back_to_board'))}</button>`
```

`t()` (`static/i18n.js`) returns the **key itself** when a key is missing:

```js
function t(key, ...args) {
  const val = _locale[key] ?? LOCALES.en[key];
  if (val === undefined) return key;  // final fallback: return key itself
  ...
}
```

Neither `kanban_close` nor `kanban_back_to_board` exists in `i18n.js` (0 occurrences each). Consequences on the current head:

- `t('kanban_close')` returns the string `"kanban_close"`, which is **truthy**, so the `|| 'Close'` fallback is dead code and the workspace-mode button literally renders the text **`kanban_close`**.
- `t('kanban_back_to_board')` has no fallback at all, so the dashboard-mode button literally renders **`kanban_back_to_board`**.

So in both layout modes the task-detail header shows a raw key as the button label. (This is the `t('key') || fallback` no-op trap — `||` never fires because the missing-key return value is a non-empty string.)

**Fix:** add both keys to `LOCALES.en` in `static/i18n.js` (matching the established kanban key block — e.g. `kanban_close: 'Close'`, `kanban_back_to_board: 'Back to board'`). Once the key exists, `t()` returns the real label and you can drop the now-redundant `|| 'Close'`. (Other locales fall back to `en` gracefully, so en is the only one strictly required, but feel free to add the handful of locales the kanban block already covers.)

### Secondary (bundle into the same push) — progress badge skips the file's own escaping convention

`static/panels.js:1852-1853` interpolates `progress.done` / `progress.total` straight into a `title="..."` attribute and the visible text with no `Number()` and no `esc()`:

```js
? `<span class="kanban-progress${...}" title="${progress.done} of ${progress.total} child tasks done">${progress.done}/${progress.total}</span>`
```

Everywhere else in this file numeric/text values go through `Number()` / `esc()` (e.g. `Number(task.comment_count || 0)`, `esc(task.assignee)`). `progress` comes from `task.progress || task.link_counts` (server data), so the real-world risk is low, but please wrap the two values in `Number()` (and the title attribute is safer with escaped values) to stay consistent with the surrounding defensive style.

### Note

The new i18n strings being added only in `en` is fine (the `t()` → en-fallback path is the established pattern), and the `settings_*_kanban_layout` keys render safely via the HTML `data-i18n` hardcoded fallbacks — those are not blockers.

Separately: this introduces a new selectable dashboard layout mode + settings, which is a product/visual decision for a maintainer to sign off on — that review is independent of the code fixes above. Getting the two raw-key labels fixed will let that review proceed on a clean render.

autrioly:
Closing in favor of a clean PR rebased on latest master.

PR #4485, opened 2026-06-19
feat(kanban): faithful dashboard layout + task detail in workspace panel

autrioly:
Adds Dashboard kanban layout option faithful to Hermes dashboard plugin + task detail in workspace panel. Replaces #4484 (rebased on latest master).

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -19,6 +19,7 @@
 <link rel="apple-touch-icon" sizes="512x512" href="static/apple-touch-icon.png">
 <script>(function(){try{var themes={light:1,dark:1,system:1},skins={default:1,ares:1,mono:1,graphite:1,slate:1,poseidon:1,sisyphus:1,charizard:1,sienna:1,catppuccin:1,hepburn:1,nous:1,'geist-contrast':1,neon:1,zeus:1,'verdigris':1},legacy={slate:['dark','slate'],solarized:['dark','poseidon'],monokai:['dark','sisyphus'],nord:['dark','slate'],oled:['dark','default']},t=(localStorage.getItem('hermes-theme')||'dark').toLowerCase(),s=(localStorage.getItem('hermes-skin')||'').toLowerCase(),m=legacy[t],theme=m?m[0]:(themes[t]?t:'dark'),skin=skins[s]?s:(m?m[1]:'default');localStorage.setItem('hermes-theme',theme);localStorage.setItem('hermes-skin',skin);if(theme==='system')theme=window.matchMedia('(prefers-color-scheme:dark)').matches?'dark':'light';if(theme==='dark')document.documentElement.classList.add('dark');if(skin!=='default')document.documentElement.dataset.skin=skin;}catch(e){document.documentElement.classList.add('dark');}})()</script>
 <script>(function(){try{var fs=localStorage.getItem('hermes-font-size');if(fs&&fs!=='default')document.documentElement.dataset.fontSize=fs;}catch(e){}})()</script>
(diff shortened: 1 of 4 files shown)


nesquena-hermes (changes requested):
Thanks for the rebase — the dashboard kanban layout and in-panel task detail are a nice direction, and the CSS is still cleanly namespaced (`html[data-kanban-layout="dashboard"]` / `.rightpanel[data-active-tab="kanban-task"]`), so the default layout is unaffected. Good progress: the `kanban_back_to_board` key the prior PR flagged is now resolved (defined in `static/i18n.js`, 13 locale blocks), so the dashboard-mode button renders correctly.

One of the two raw-key blockers from #4484 is still open, though, plus the same hardening nit.

### Blocker — the workspace-mode close button still renders the raw key `kanban_close`

`static/panels.js:2986` (the `mode === 'workspace'` branch of `_kanbanRenderTaskDetail`):

```js
? `<button class="btn secondary kanban-back-btn" onclick="closeKanbanTaskDetail()">${esc(t('kanban_close') || 'Close')}</button>`
```

`t()` (`static/i18n.js:19266`) returns the **key itself** when a key is missing:

```js
if (val === undefined) return key;  // final fallback: return key itself
```

`kanban_close` is not defined anywhere in `static/i18n.js` (`grep -c 'kanban_close:' static/i18n.js` → `0`). So `t('kanban_close')` returns the non-empty string `"kanban_close"`, which is truthy — the `|| 'Close'` fallback is dead code and never fires. In dashboard layout, the workspace-panel task-detail header button literally displays the text **`kanban_close`** instead of "Close". (This is the same `t('key') || fallback` no-op trap the prior review flagged; `kanban_back_to_board` got fixed in the rebase but `kanban_close` was missed.)

**Fix:** add `kanban_close: 'Close'` to `LOCALES.en` in `static/i18n.js` (mirroring the `kanban_back_to_board` block you already have), then drop the now-redundant `|| 'Close'`. Other locales fall back to `en` gracefully, so `en` is the only strictly required entry.

### Secondary (bundle into the same push) — progress badge skips the file's own escaping convention

`static/panels.js:1911-1912` interpolates `progress.done` / `progress.total` straight into a `title="..."` attribute and the visible text with no `Number()` and no `esc()`:

```js
const progressHtml = (isDashboard && progress && progress.done !== undefined && progress.total !== undefined)
  ? `<span class="kanban-progress${progress.done === progress.total ? ' kanban-progress--full' : ''}" title="${progress.done} of ${progress.total} child tasks done">${progress.done}/${progress.total}</span>`
  : '';
```

Everywhere else in this file numeric/text values go through `Number()` / `esc()` (siblings `priority`, `linkTotal`, `comments` a few lines up all use `Number()`). `progress` comes from `task.progress || task.link_counts` (server data), so real-world risk is low, but please wrap the two values in `Number()` (and escape them for the title attribute) to stay consistent with the surrounding defensive style.

### Note

The new selectable dashboard layout mode + settings is a product/visual decision for a maintainer to sign off on — that review is independent of the code fix above. Getting the `kanban_close` label fixed lets that review proceed on a clean render. The `settings_*_kanban_layout` `data-i18n` keys render safely via their hardcoded HTML fallbacks and are not blockers.

nesquena-hermes:
Thanks @autrioly — closing this as a scope/curation decision, not a quality one. A "faithful dashboard" kanban layout restyle + task-detail-in-workspace-panel is a sizable visible change to a niche surface, it's been sitting in changes-requested, and a restyle of an existing surface like this needs a deliberate maintainer visual sign-off + a rebuild against current master rather than landing from a stale bounced branch.

Closing for now — if we do a focused pass on the kanban/dashboard surface later, this is a useful reference. Appreciate the effort.

PR #4486, opened 2026-06-19
Fix untracked update collision recovery

### LINKED ISSUE #4310: bug(updates): untracked-file pull failure leaves user stuck with no Force-update recovery path (sibling of #813)
## Summary

When the in-app **Update Now** fails because an **untracked** local file would be overwritten by the incoming merge (most commonly a newly-added test/source file the remote now ships at the same path), the WebUI surfaces a dead-end error and offers **no recovery button**. The user is stuck in a broken state with no way out except direct terminal access.

Reported by **@b3nw** in Discord #report-bugs (2026-06-16). Verbatim:

> it used to offer the ability to stash + apply the clean version, that got lost at some point, so if you're in a broken state, no way out without direct terminal intervention

Rendered error (screenshot):

> **Update failed (webui):** Pull failed: From https://github.com/nesquena/hermes-webui * tag v0.51.456 -> FETCH_HEAD error: The following untracked working tree files would be overwritten by merge: `tests/test_custom_provider_prefix_collisions.py` Please move or remove them before you merge. Aborting Local webui modifications were restored to the working tree; save or stash them before running destructive recovery commands.

untracked update failure

## Steps to reproduce

1. In the WebUI repo, create an **untracked** file at a path the remote is about to add (e.g. `tests/test_custom_provider_prefix_collisions.py`), or just have any untracked file that collides with an incoming commit.
2. Click **Update Now** when the remote is ahead.
3. The pull aborts with the "untracked working tree files would be overwritten" error and **no Force-update button appears**.

Reproduced deterministically against a synthetic remote: with an untracked file colliding with an incoming tracked file, `git status --porcelain --untracked-files=no` reports a **clean** tree → no stash is taken → `git pull --ff-only` aborts with `error: The following untracked working tree files would be overwritten by merge` (rc=1). A `git checkout . && git reset --hard origin/master` clears it cleanly (rc=0), confirming a force path *would* recover it.

## Root cause

Two gaps in `apply_update()` (`api/updates.py`):

**1. Untracked files are deliberately excluded from the dirty-check and the stash**, so the auto-stash path never engages for them.

```python
# api/updates.py:1298-1302
# Check for dirty working tree (ignore untracked files — git stash
# doesn't include them, so stashing on '??' alone leaves nothing to pop)
status_out, status_ok = _run_git(
    ['status', '--porcelain', '--untracked-files=no'], path
)
```
```python
# api/updates.py:1319-1320
if status_out:
    _, ok = _run_git(['stash', 'push', '-m', 'hermes-update-autostash'], path)
```
`status_out` is empty when the *only* dirty state is untracked files, so no stash is taken, and `git pull --ff-only` then aborts on the collision. The plain `stash push` (no `--include-untracked` / `-u`) also wouldn't capture them even if the check did fire.

**2. The untracked-collision error falls through to the generic fallback, which sets no recovery flag**, so the existing **Force update** button never appears.

```python
# api/updates.py:1427-1431  (generic fallback)
message_parts = [f'Pull failed: {detail}']
if restored_note:
    message_parts.append(restored_note)
return {'ok': False, 'message': ' '.join(message_parts)}
```

The Force-update button only shows for `res.conflict || res.diverged`:

```js
// static/ui.js:6725-6729
// Show "Force update" button when the error is recoverable by a hard reset
if(forceBtn&&(res.conflict||res.diverged)){
  forceBtn.dataset.target=target;
  forceBtn.style.display='inline-block';
}
```

The untracked-overwrite return carries neither flag, so the recovery affordance built in #813 (the `apply_force_update()` + Force-update button) is unreachable for exactly the case where it's most needed. (This is the **unaddressed sibling of #813** — that issue added the recovery path for `conflict`/`diverged`; the untracked-overwrite failure mode was never wired into it.)

`apply_force_update()` itself already recovers this state correctly — it does `checkout . && reset --hard <compare_ref>`. But note `reset --hard` does **not** remove untracked files; if the colliding file is untracked, a `git clean` step (or `-u`/`--include-untracked` stash) may be needed for a guaranteed-clean force. Worth verifying during the fix.

## Suggested fix (fix shape)

Two complementary options — recommend doing both:

- **Detect the untracked-overwrite failure and flag it as recoverable.** In the generic-fallback branch (or a new branch), detect `untracked working tree files would be overwritten` in `pull_lower` and return `'conflict': True` (or a new `'untracked': True` flag the UI also honors) so the existing **Force update** button surfaces. This is the minimal fix that restores a one-click recovery path.
- **Optionally auto-stash untracked files too.** Switch the dirty-check to include untracked (`--untracked-files=normal`) and use `git stash push --include-untracked` so collisions are stashed and restored automatically — this is closer to the "stash + apply the clean version" behavior @b3nw remembers. Must preserve the existing conflict-safe `stash apply` + `drop` restore logic (#3666/#3667) on pull failure.
- **Ensure the force path is truly clean for untracked files** — add `git clean -fd` (scoped, with care) or stash `-u` before `reset --hard` in `apply_force_update()` if the colliding file is untracked, since `reset --hard` alone leaves untracked files in place and the next pull would abort again.

## Scope

Small, single-repo (`nesquena/hermes-webui`). Backend `api/updates.py` (flag the untracked failure + optional `-u` stash + optional `clean` in force path); the UI already handles the recovery button — flagging the error is enough to make it appear. ~10-30 LOC + a test mirroring `tests/test_update_stash_recovery.py` for the untracked-collision case.

## Reported by

@b3nw — Discord #report-bugs, 2026-06-16. Confirmed legitimate and reproduced by maintainer triage.

franksong2702:
## Thinking Path

- Hermes WebUI's in-app updater should provide a browser recovery path when a local checkout blocks a fast-forward update.
- #4310 showed an untracked-file collision where `git status --untracked-files=no` looked clean, `git pull --ff-only` failed, and the response did not expose the existing Force update button.
- A button-only fix would still be incomplete because `git reset --hard` leaves untracked files behind.
- This PR flags that pull failure as recoverable and makes the confirmed force-update path remove untracked colliders before resetting.

## What Changed

- Detect `untracked working tree files would be overwritten` pull failures in `api/updates.py` and return `conflict: true` so the existing Force update affordance appears.
- Run `git clean -fd` between `git checkout .` and `git reset --hard <ref>` during force update, without `-x`, so ignored cache/build artifacts are preserved.
- Stop the force update if `git clean -fd` fails instead of hiding the cleanup failure behind a reset attempt.
- Update the Force update confirmation copy to disclose that untracked files will be deleted.
- Add regression tests for the recoverable flag, clean-before-reset order, clean-failure stop path, and confirmation copy.
- Update `CHANGELOG.md`.

## Why It Matters

Users can recover from a common self-update dead end without terminal access. The destructive path is also honest about what it deletes and avoids claiming success if the untracked cleanup fails.

Closes #4310.

## Verification

- `./scripts/test.sh tests/test_update_stash_recovery.py tests/test_update_banner_fixes.py::TestApplyForceUpdate tests/test_update_banner_fixes.py::test_force_update_confirm_discloses_untracked_file_deletion tests/test_updates.py::test_apply_force_update_fetches_tags_with_force`
- `node --check static/ui.js`
- `git diff --check`

## Risks / Follow-ups

- Force update is intentionally destructive after confirmation. This PR uses `git clean -fd` only, not `git clean -fdx`, so ignored files remain untouched.
- The regular auto-stash path still ignores untracked files; this PR focuses on the recovery path requested by #4310 rather than broadening stash semantics.

## Model Used

OpenAI GPT-5 Codex as coordinator, with a bounded `gpt-5.3-codex-spark` sub-agent for the initial implementation pass. The coordinator wrote/owned the acceptance tests, reviewed the diff, added the confirmation-copy and clean-failure guards, and ran verification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Update failures caused by untracked colliding files now surface the existing force-update recovery path, and force update now cleans untracked files before hard reset (#4310).**
+
 ## [v0.51.513] — 2026-06-19 — Release RX (credential-pool quota status for all pooled providers)
 
 ### Added
--- a/api/updates.py
+++ b/api/updates.py
@@ -1242,8 +1242,15 @@ def apply_force_update(target: str) -> dict:
 
         compare_ref = _select_apply_compare_ref(path)
 
-        # Discard local modifications then reset to remote HEAD
+        # Discard local modifications and untracked colliders before resetting.
+        # Do not use -x: ignored build/cache artifacts should survive force update.
         _run_git(['checkout', '.'], path)
+        _, clean_ok = _run_git(['clean', '-fd'], path)
+        if not clean_ok:
+            return {
+                'ok': False,
+                'message': 'Failed to remove untracked files before force reset',
+            }
         _, ok = _run_git(['reset', '--hard', compare_ref], path)
         if not ok:
             return {'ok': False, 'message': f'Force reset to {compare_ref} failed'}
@@ -1343,6 +1350,9 @@ def _apply_update_inner(target):
     if not pull_ok:
         pull_lower = pull_out.lower()
         detail = pull_out.strip()[:300] if pull_out.strip() else '(no output from git)'
+        untracked_collision = (
+            'untracked working tree files would be overwritten' in pull_lower
+        )
         diverged_failure = (
             'not possible to fast-forward' in pull_lower or 'diverged' in pull_lower
         )
@@ -1436,7 +1446,10 @@ def _apply_update_inner(target):
         message_parts = [f'Pull failed: {detail}']
         if restored_note:
             message_parts.append(restored_note)
-        return {'ok': False, 'message': ' '.join(message_parts)}
+        response = {'ok': False, 'message': ' '.join(message_parts)}
+        if untracked_collision:
+            response['conflict'] = True
+        return response
 
     # Re-apply stash if we stashed.
     stash_drop_failed = False
--- a/static/ui.js
+++ b/static/ui.js
@@ -7168,7 +7168,7 @@ async function forceUpdate(btn){
   if(!target) return;
   const confirmed=await showConfirmDialog({
     title:'Force update '+target+'?',
-    message:'This will discard all local changes in the '+target+' repo and reset to the latest remote version. This cannot be undone.',
+    message:'This will discard all local changes and delete untracked files in the '+target+' repo, then reset to the latest remote version. This cannot be undone.',
     confirmLabel:'Force update',
     danger:true,
     focusCancel:true,
--- a/tests/test_update_banner_fixes.py
+++ b/tests/test_update_banner_fixes.py
@@ -2025,6 +2025,18 @@ class TestForceButtonResetOnRetry:
         )
 
 
+def test_force_update_confirm_discloses_untracked_file_deletion():
+    """#4310: destructive force-update copy must include untracked files."""
+    src = read('static/ui.js')
+    m = re.search(r'async function forceUpdate\b.*?\n\}', src, re.DOTALL)
+    assert m, "forceUpdate() not found"
+    fn = m.group(0)
+    assert 'delete untracked files' in fn, (
+        "forceUpdate confirmation must disclose that git clean -fd deletes "
+        "untracked files before the hard reset"
+    )
+
+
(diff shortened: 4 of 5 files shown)


nesquena-hermes:
Shipped in v0.51.520 via the rebased release branch (credited to you in CHANGELOG + release notes). Thanks @rodboev! An untracked-file pull collision now surfaces the Force-update recovery path, and force-update clears untracked colliders (no -x) behind the danger-confirmed dialog that now discloses the deletion.

Issue #4480, opened 2026-06-19
SOUL.md has read-only permissions on two-container setup


nordwestt:
I kept getting a 500 error when trying to update the soul and inspected the logs to find the permission issue.
Verified it by doing `docker exec hermes-agent ls -land /home/hermes/.hermes` and seeing that `SOUL.md` shows up with following line:
```
-r--r--r--  1 1000 1000     536 Jun 19 08:00 SOUL.md
```

Fix was to simply do `chmod 644 SOUL.md` and now it works 👍 
Posting here as this might be a systematic issue, since I literally started from fresh.

nesquena-hermes:
## Summary

Thanks — your diagnosis (read-only `SOUL.md` → 500 on save) is right, and there are actually two separate things here: a WebUI-side defect (an opaque 500 where a clear error belongs) and a likely seeding-mode mismatch on the agent side that explains where the `0444` comes from. The WebUI bug is real regardless of how the file got read-only.

## Code reference — the WebUI side

The soul/memory save handler does a bare `write_text` with no permission handling (`api/routes.py:17703-17714`):

```python
elif section == "soul":
    target = home / "SOUL.md"
...
if target.is_symlink():
    return bad(handler, "Cannot write to a symlinked memory file")
target.write_text(body["content"], encoding="utf-8")
return j(handler, {"ok": True, "section": section, "path": str(target)})
```

If `target` is `-r--r--r--`, `write_text` raises `PermissionError`, which bubbles up as an unhelpful 500 with no actionable message — that's why you had to go spelunking in the container. The codebase already has the right idiom for this: `/api/session/title` etc. catch `PermissionError` and return a clean 403 (`api/routes.py:9016-9017`):

```python
except PermissionError:
    return bad(handler, "Read-only imported sessions cannot be renamed from WebUI", 403)
```

## Recommendation

Two-part fix. First, make the handler fail loudly and clearly instead of 500-ing:

```python
try:
    target.write_text(body["content"], encoding="utf-8")
except PermissionError:
    return bad(handler, f"{section.upper()}.md is not writable ({oct(target.stat().st_mode)[-3:]}). "
                        f"Run `chmod 644 {target}` (or fix ownership on the shared volume).", 403)
```

[...]

PR #4487, opened 2026-06-19
Report read-only memory writes as 403

franksong2702:
## Thinking Path

- The memory editor should fail with actionable permissions guidance, not an opaque server error.
- #4480 reports a read-only `SOUL.md` in a two-container setup, which currently makes `/api/memory/write` bubble `PermissionError` as a generic 500.
- The WebUI can fix its half by converting filesystem write permission failures into a clear 403 while leaving the agent-side source of the read-only file as a separate concern.

## What Changed

- Wrap memory target writes in `api/routes.py` with `PermissionError` handling.
- Return a 403 response that names the non-writable target file instead of letting the exception become a 500.
- Keep the existing symlink-target guard unchanged.
- Add regression coverage for read-only `SOUL.md` writes.
- Update `CHANGELOG.md`.

## Why It Matters

Users get a clear permissions/ownership problem they can fix instead of having to inspect container logs to understand why saving `SOUL.md` failed.

Refs #4480.

## Verification

- `./scripts/test.sh tests/test_memory_write_symlink_guard.py::test_memory_write_read_only_soul_returns_403 tests/test_memory_write_symlink_guard.py`
- `git diff --check`

## Risks / Follow-ups

- This does not change the agent-side distribution or seeding behavior that may create a read-only `SOUL.md`; it only makes the WebUI failure mode explicit and actionable.
- The response includes the local target path, consistent with the existing successful memory-write response shape.

## Model Used

OpenAI GPT-5 Codex as coordinator, with a bounded `gpt-5.3-codex-spark` sub-agent attempted for implementation. The coordinator owned the acceptance test, corrected the worker worktree mistake, applied the final patch, reviewed the diff, and ran verification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Memory writes now return actionable 403 errors for read-only `SOUL.md`, `MEMORY.md`, or `USER.md` targets (#4480).** The memory editor no longer turns permission failures into an opaque 500; the response names the target file so users can fix ownership or permissions on the shared volume.
+
 ## [v0.51.513] — 2026-06-19 — Release RX (credential-pool quota status for all pooled providers)
 
 ### Added
--- a/api/routes.py
+++ b/api/routes.py
@@ -17711,7 +17711,10 @@ def _handle_memory_write(handler, body):
     # (#4217/#4234/#4240).
     if target.is_symlink():
         return bad(handler, "Cannot write to a symlinked memory file")
-    target.write_text(body["content"], encoding="utf-8")
+    try:
+        target.write_text(body["content"], encoding="utf-8")
+    except PermissionError:
+        return bad(handler, f"{target.name} is not writable: {target}", 403)
     return j(handler, {"ok": True, "section": section, "path": str(target)})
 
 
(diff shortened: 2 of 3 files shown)


franksong2702:
Follow-up on the Greptile EROFS note: pushed `5175d910`, which now treats Docker-style `OSError(errno.EROFS)` the same as `PermissionError` for `/api/memory/write` read-only targets, while re-raising unrelated I/O errors.

Verification:
- `./scripts/test.sh tests/test_memory_write_symlink_guard.py::test_memory_write_read_only_soul_returns_403 tests/test_memory_write_symlink_guard.py::test_memory_write_read_only_filesystem_returns_403 tests/test_memory_write_symlink_guard.py` -> `5 passed`
- `git diff --check`
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master` -> no new findings on added/modified lines

GitHub read-back on `5175d910`: `MERGEABLE`, `mergeStateStatus=CLEAN`; browser-smoke, lint, and all Python 3.11/3.12/3.13 shards passed. Greptile re-review is 5/5 with no special files requiring attention.

nesquena-hermes:
Shipped in **v0.51.518** (Release SC) — `/api/memory/write` now catches `PermissionError`/`EROFS` and returns an actionable 403 (naming the file, its mode, and a `chmod 644` / fix-ownership hint) instead of an opaque 500, with the symlinked-target guard preserved. The fix landed via the release pipeline (the commit carries this PR's diff), so closing this out.

Thanks @franksong2702 — credited in the CHANGELOG for v0.51.518. Closes #4480.

PR #2135, opened 2026-05-12
Fix custom live model scoping

franksong2702:
## Thinking Path

Issue #2126 reports that `/api/models/live?provider=custom:<slug>` returns models from sibling `custom_providers` entries instead of only the requested named provider. The confirmed root cause was in `_handle_live_models`: the custom-provider fallback collected every `custom_providers[].model`, and the direct `/v1/models` fallback used the active profile's `model.base_url` / `model.api_key` instead of the requested named provider entry.

Closes #2126.

## What Changed

- Scoped `/api/models/live` custom-provider fallback to the requested provider:
  - `custom:<slug>` only reads the matching named custom provider entry.
  - bare `custom` only reads unnamed custom provider entries.
- Included model IDs from both singular `model` and plural `models` config forms.
- Updated direct OpenAI-compatible `/v1/models` fallback so named custom providers use the matched entry's `base_url` and `api_key` pair.
- Avoided mixing a matched named provider `base_url` with the active profile's API key.
- Added regression tests for named-provider scoping, bare-custom scoping, and direct fetch endpoint/key selection.
- Updated the changelog.

## What Is Not Included

- This PR does not claim to resolve #2131. That issue appears related to the broader provider/source-of-truth line, but it should be re-evaluated after this targeted `/api/models/live` fix lands.
- No cache behavior changes; `_live_models_cache_key(provider)` already includes the normalized provider id.

## Verification

- `python -m pytest -q tests/test_byok_model_dropdown.py::TestLiveModelsCustomProviderFallback`
- `python -m pytest -q tests/test_byok_model_dropdown.py tests/test_live_models_ttl_cache.py tests/test_pr1947_same_model_multiple_custom_providers.py tests/test_issue1806_named_custom_provider_resolution.py`
- `python -m py_compile api/routes.py tests/test_byok_model_dropdown.py`
- `git diff --check`

## Risks

- Low. The change is scoped to the fallback path used when `provider_model_ids()` returns no models for custom providers.
- Named custom providers without configured model IDs and without their own API key will no longer borrow an unrelated active profile key for live fetch. That is intentional; borrowing a sibling provider credential is the leakage class this PR fixes.

## Model Used

GPT-5.4 Codex

Issue #2126, opened 2026-05-12
Custom provider models leak into sibling providers via /api/models/live


solink7:
## Bug

When multiple `custom_providers` are configured in `config.yaml`, the `/api/models/live` endpoint returns **all** custom provider models under whichever custom provider is queried — instead of scoping to the matching provider only.

### Reproduction

`config.yaml`:
```yaml
custom_providers:
  - name: rightcode-codex
    base_url: https://right.codes/codex/v1
    model: gpt-5.5
  - name: infini-ai
    base_url: https://open.bigmodel.cn/api/paas/v4
    model: glm-5.1
  - name: xiaomi-mimo
    base_url: https://...
    model: mimo-v2.5-pro
```

Request: `GET /api/models/live?provider=custom:rightcode-codex`

**Expected:** only `gpt-5.5`

**Actual:** `["glm-5.1", "mimo-v2.5-pro", "gpt-5.5"]`

This causes the model picker dropdown to show GLM 5.1 and Mimo V2.5 Pro under the rightcode-codex group.

### Root Cause

In `api/routes.py` `_handle_live_models()` (~line 6112), when `provider_model_ids()` returns `[]` for a `custom:*` provider, the fallback iterates **all** `custom_providers` entries and collects every model indiscriminately:

```python
if provider == "custom" or provider.startswith("custom:"):
    _cp_entries = cfg.get("custom_providers", [])
    if isinstance(_cp_entries, list):
        ids = [
            _cp.get("model", "")
            for _cp in _cp_entries
            if isinstance(_cp, dict) and _cp.get("model", "")
        ]
```

There is no filter matching `_cp` name/slug to the requested `provider`.

### Suggested Fix

Use `_custom_provider_slug_from_name()` (already available in `api/config.py`) to match only the entry whose slug equals the requested `custom:*` provider:

```python
if provider.startswith("custom:"):
    if _cp_slug != provider:
        continue
elif _cp_slug:
    # Generic "custom" should not absorb named providers either
    continue
```

Also handle `models` as both `list` and `dict` (keys are model IDs when dict).

### Environment

- Hermes WebUI: latest master (15d6203)
- Multiple custom_providers configured

nesquena-hermes:
## Summary

[...]

PR #892, opened 2026-04-23
feat(models): live-first model fetching for all OpenAI-compat providers

bergeouss:
## Thinking Path

- Hermes WebUI aims for near 1:1 parity with the Hermes CLI in a browser
- The model picker showed hardcoded static lists from `_PROVIDER_MODELS` that go stale as providers add/remove models
- The existing pipeline already had a live fetch via `provider_model_ids()` in hermes_cli, but it only covered some providers
- When hermes_cli was unavailable, the fallback jumped straight to the static dict — skipping any live fetch
- Providers like zai, minimax, mistralai, xai, openai-codex all expose standard `/v1/models` endpoints but weren't wired into any live fetch path
- This PR adds a direct OpenAI-compat `/v1/models` fetch as fallback layer, ensuring all discoverable providers show live models
- The static list remains as last-resort offline fallback so the UI always renders something

## What Changed

- **`api/routes.py`** — Added `_OPENAI_COMPAT_ENDPOINTS` dictionary mapping 7 providers to their `/v1/models` base URLs
- Restructured `_handle_live_models()` resolution chain:
  1. `hermes_cli.provider_model_ids()` (existing agent live fetch)
  2. Custom providers from `config.yaml` (existing)
  3. **NEW:** Direct `/v1/models` HTTP fetch for known OpenAI-compat endpoints
  4. Static `_PROVIDER_MODELS` dict (moved to last resort)
- Removed the premature static fallback in the `except` block of the `provider_model_ids()` import

### Providers now live-fetched:

| Provider | Endpoint |
|----------|----------|
| zai | `https://api.z.ai/v1/models` |
| minimax | `https://api.minimax.chat/v1/models` |
| mistralai | `https://api.mistral.ai/v1/models` |
| xai | `https://api.x.ai/v1/models` |
| openai-codex | `https://api.openai.com/v1/models` |
| deepseek | `https://api.deepseek.com/v1/models` |
| gemini | `https://generativelanguage.googleapis.com/v1beta/openai/models` |

## Why It Matters

Every time a provider adds a model, someone had to manually update `_PROVIDER_MODELS`, open a PR, get review, and ship a release. The Nous model ID format bug (#854) was undetected for months because the static list was wrong and no test exercised the actual API.

This eliminates that entire class of maintenance burden. Users see the correct available model list for their configured provider automatically.

## Verification

- Python syntax validated (`ast.parse`)
- Resolution chain preserves all existing behavior (hermes_cli still first, custom providers still handled)
[...]

PR #4488, opened 2026-06-19
[security] fix(models): scope live model provider keys

Hinotoi-agent:
## Thinking Path

- Hermes WebUI enriches the model picker by asking `/api/models/live?provider=<provider>` for provider-specific live model catalogs.
- The OpenAI-compatible fallback is allowed to call fixed provider `/models` endpoints, but each request must use only credentials that belong to that requested provider.
- The existing fallback had a provider-scoped key lookup, then fell back to the active top-level `model.api_key` even when the requested provider was different.
- This PR keeps same-provider top-level key compatibility while preventing cross-provider credential forwarding.
- The change is intentionally narrow: one credential-selection guard, regression coverage for both blocked and allowed cases, and a changelog note.

## Summary

This PR hardens live model discovery so a top-level provider API key is never forwarded to a different provider's `/models` endpoint.

- Prevents `/api/models/live?provider=<other>` from reusing the active provider's `model.api_key` when the requested provider has no provider-scoped key.
- Preserves the existing same-provider behavior, including alias-equivalent active providers such as `z.ai` resolving to `zai`.
- Adds regression tests for the blocked cross-provider case and the allowed same-provider cases.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Cross-provider live-model credential forwarding | A caller who can reach the live model endpoint could cause WebUI to send the active provider's top-level API key to a different fixed third-party provider endpoint during model discovery. | Medium |

## Before this PR

- `_handle_live_models()` first looked for `providers[requested_provider].api_key`.
- If that provider-scoped key was absent, it fell back to `model.api_key` unconditionally.
- A request such as `/api/models/live?provider=mistralai` could therefore send an active OpenAI top-level key in an Authorization bearer header to Mistral's `/models` endpoint.
- There was no regression test proving that cross-provider requests without provider-scoped credentials avoid live fetches with the active provider's key.

## After this PR

- The OpenAI-compatible fallback still prefers `providers[requested_provider].api_key`.
- A top-level `model.api_key` is used only when the active model provider resolves to the same canonical provider as the requested live-model provider.
- Cross-provider requests without a matching provider-scoped credential fall back to the static catalog rather than making a credentialed live fetch.
- Tests now cover:
  - cross-provider request does not call `urlopen`,
  - same-provider request can still use the top-level key,
  - alias-equivalent active provider names still work.

## Why It Matters

Provider API keys should not cross provider trust boundaries. Model discovery is a convenience feature, but it still attaches bearer credentials to outbound requests. If a key configured for one provider is sent to another provider's endpoint, the secret is disclosed to the wrong third party and may appear in upstream logs, telemetry, or error handling outside the user's intended provider relationship.

- #892 added the OpenAI-compatible live `/models` fallback for standard providers.
- #2126 / #2135 fixed custom-provider model and endpoint scoping so named custom providers do not leak sibling custom-provider models or mix custom endpoint/key pairs.
- #4413 currently adjusts endpoint selection for stale provider catalogs and configured base URLs.

This PR addresses a distinct remaining standard-provider credential boundary: the fallback for fixed OpenAI-compatible providers must not use `model.api_key` unless the active model provider and requested live-model provider are the same after alias normalization.

## Attack flow

```text
Authenticated WebUI caller requests /api/models/live?provider=mistralai
    -> Mistral has no provider-scoped key in config
        -> fallback reuses active top-level model.api_key from another provider
            -> urllib sends Authorization: Bearer <active-provider-key> to https://api.mistral.ai/v1/models
                -> provider credential is disclosed to the wrong third-party endpoint
```

## Affected code

| Issue | Files |
|-------|-------|
| Cross-provider live-model credential forwarding | `api/routes.py`, `tests/test_byok_model_dropdown.py`, `CHANGELOG.md` |

## Root cause

Cross-provider live-model credential forwarding:

- Direct cause: `_handle_live_models()` fell back from `providers[provider].api_key` to `cfg["model"]["api_key"]` without checking whether the top-level key belonged to the requested provider.
- Boundary failure: the requested provider controls the outbound endpoint, while the top-level key belongs to the active model provider; those are separate credential scopes unless their provider IDs match after alias normalization.

## CVSS assessment

| Cross-provider live-model credential forwarding | 4.3 Medium | `CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:L/I:N/A:N` |

Rationale:

- The impact is bounded to confidentiality of a configured provider credential.
- The attacker must be able to reach the WebUI API in a configured deployment.
- The endpoint targets fixed provider model-list URLs rather than arbitrary attacker-controlled hosts, so this is credential disclosure to the wrong third-party provider, not arbitrary SSRF.

## Safe reproduction steps

1. Configure an active top-level provider key, for example:

   ```yaml
   model:
     provider: openai
     api_key: active-provider-canary
   providers:
     mistralai: {}
   ```

2. Stub or intercept `urllib.request.urlopen` so no real network call is made.
3. Call `_handle_live_models()` through the `/api/models/live?provider=mistralai` path with `provider_model_ids()` returning `[]`.
4. On vulnerable code, the stub sees a request to `https://api.mistral.ai/v1/models` with `Authorization: Bearer active-provider-canary`.
5. With this PR, the cross-provider request does not call `urlopen` and the response falls back to the static Mistral catalog.

- A request for provider `mistralai` could use a top-level key configured for provider `openai`.
- The proof signal is the captured outbound request carrying `Authorization: Bearer active-provider-canary` to Mistral's `/models` endpoint.
- No real secret or production endpoint is required to reproduce the issue; the regression uses a canary value and a monkeypatched `urlopen`.

## What Changed

- Restrict fallback use of top-level `model.api_key` to cases where the active model provider resolves to the requested live-model provider.
- Preserve provider-scoped key behavior for `providers[provider].api_key`.
- Preserve same-provider top-level key behavior, including alias-equivalent provider names.
- Add regression tests for blocked cross-provider forwarding and allowed same-provider live fetches.
- Add a release-note-ready changelog entry under `Unreleased`.

## Files changed

| Security fix | `api/routes.py` | Adds the active-provider equality check before top-level `model.api_key` can be used for live model discovery. |
| Regression tests | `tests/test_byok_model_dropdown.py` | Adds blocked cross-provider and allowed same-provider/alias-equivalent live fetch coverage. |
| Changelog | `CHANGELOG.md` | Documents the security-sensitive behavior change. |

## Maintainer impact

- No new dependencies, build steps, or frontend changes.
- Provider-scoped credentials continue to work as before.
- Top-level keys continue to work for the active provider itself.
- Cross-provider model discovery without a provider-scoped key now falls back to the static model list instead of making a credentialed request.

## Fix rationale

The live model endpoint's outbound provider is selected by the request parameter, so credential selection must be scoped to that same provider. Comparing the active top-level provider after alias normalization keeps legitimate same-provider compatibility while preventing accidental cross-provider disclosure.

## Type of change

- [x] Security fix
- [x] Tests
- [x] Documentation update

## Verification

Docker/container validation:

- [x] `docker run --rm -v <repo-copy>:/src:ro python:3.11-slim bash -lc 'apt-get update >/dev/null && apt-get install -y git >/dev/null && cp -a /src /work && cd /work && ./scripts/test.sh tests/test_byok_model_dropdown.py -q && ./.venv/bin/ruff check tests/test_byok_model_dropdown.py && git diff --check'`
  - Result: `23 passed`; ruff passed for the touched test file; `git diff --check` passed.

Host-local supplemental validation:

- [x] `./scripts/test.sh tests/test_byok_model_dropdown.py -q`
  - Result: `23 passed`.
- [x] `./.venv/bin/ruff check tests/test_byok_model_dropdown.py`
  - Result: passed.
- [x] `git diff --check`
  - Result: passed.

Notes:

- `./.venv/bin/ruff check api/routes.py tests/test_byok_model_dropdown.py` reports pre-existing lint findings in the large legacy `api/routes.py` file unrelated to this patch. The touched test file passes ruff cleanly.

## Risks / Follow-ups

- This may reduce live enrichment for a provider that only has an unrelated top-level key configured. That is intentional: provider-specific live fetches should use provider-specific credentials.
- If maintainers want a broader credential helper, it should return a key only after proving the credential belongs to the requested provider.
- This PR does not change custom-provider credential pool behavior; it only scopes the standard OpenAI-compatible fallback's top-level key use.

## Disclosure notes

- This report is bounded to the standard-provider live model discovery fallback.
- The finding does not claim arbitrary-host SSRF or unauthenticated reachability.
- No real provider credentials were used; tests use canary strings and a monkeypatched request function.

## Model Used

AI-assisted contribution using OpenAI `gpt-5.5` through Hermes Agent, with repository-local tests, Docker validation, and GitHub CLI workflow support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Security
+
+- **Live model discovery no longer forwards a top-level provider API key to a different provider's `/models` endpoint.** The OpenAI-compatible fallback now only uses `model.api_key` when the requested live-model provider matches the active model provider; cross-provider requests fall back to provider-scoped credentials or the static catalog instead.
+
 ## [v0.51.513] — 2026-06-19 — Release RX (credential-pool quota status for all pooled providers)
 
 ### Added
--- a/api/routes.py
+++ b/api/routes.py
@@ -13204,11 +13204,19 @@ def _handle_live_models(handler, parsed):
                     import urllib.request
                     _providers_cfg = cfg.get("providers", {})
                     _prov = _providers_cfg.get(provider, {}) if isinstance(_providers_cfg, dict) else {}
-                    # Only use provider-scoped key — never fall back to a top-level
-                    # api_key which may belong to a different provider.
+                    # Only use a provider-scoped key.  A top-level model.api_key
+                    # is safe here only when it belongs to the requested provider;
+                    # otherwise /api/models/live?provider=<other> could forward
+                    # the active provider's credential to the wrong third party.
                     _key = _prov.get("api_key") if isinstance(_prov, dict) else None
                     if not _key:
-                        _key = cfg.get("model", {}).get("api_key")
+                        _model_cfg = cfg.get("model", {})
+                        if isinstance(_model_cfg, dict):
+                            _active_provider = _resolve_provider_alias(
+                                (_model_cfg.get("provider") or "").strip().lower()
+                            )
+                            if _active_provider == provider:
+                                _key = _model_cfg.get("api_key")
                     if _key:
                         _req = urllib.request.Request(
                             f"{_ep}/models",
--- a/api/routes.py
+++ b/api/routes.py
@@ -13124,7 +13124,8 @@ def _handle_live_models(handler, parsed):
                 # Fallback: try credential pool for base_url + api_key
                 if (not _base_url or not _api_key) and provider.startswith("custom:"):
                     try:
-                        from api.config import _has_explicit_pool_credentials, _resolve_provider_alias
+                        from api.config import _has_explicit_pool_credentials
+
                         if _has_explicit_pool_credentials(provider):
                             from agent.credential_pool import load_pool as _lpool
                             _resolved = _resolve_provider_alias(provider)
--- a/tests/test_byok_model_dropdown.py
+++ b/tests/test_byok_model_dropdown.py
@@ -247,7 +247,6 @@ class TestLiveModelsCustomProviderFallback:
             return True
         monkeypatch.setattr(r, "j", fake_j)
 
-        from urllib.parse import urlparse
         parsed = mock.MagicMock()
         parsed.query = "provider=custom"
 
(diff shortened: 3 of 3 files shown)


Hinotoi-agent:
Addressed in `49168fa`: removed the redundant `_resolve_provider_alias` import from the nested credential-pool fallback and reused the resolver already imported in `_handle_live_models`.

Validation:
- `./scripts/test.sh tests/test_byok_model_dropdown.py -q` → 23 passed
- `./.venv/bin/ruff check tests/test_byok_model_dropdown.py` → passed
- `./.venv/bin/python -m py_compile api/routes.py tests/test_byok_model_dropdown.py` → passed
- `git diff --check` → passed
- Docker `python:3.11-slim`: focused test passed (23 passed), ruff on touched test passed, py_compile passed

nesquena-hermes:
Pulled the branch and read the full `_handle_live_models` path on both `origin/master` and HEAD. The credential-scoping guard is correct, and the alias symmetry that makes the same-provider compatibility hold is genuinely sound — I checked the one thing that could have broken it.

## The guard is symmetric because the requested provider is already normalized

The new check (`api/routes.py:13208-13220`) only reuses a top-level `model.api_key` when the active provider resolves to the requested provider:

```python
_model_cfg = cfg.get("model", {})
if isinstance(_model_cfg, dict):
    _active_provider = _resolve_provider_alias(
        (_model_cfg.get("provider") or "").strip().lower()
    )
    if _active_provider == provider:
        _key = _model_cfg.get("api_key")
```

The thing I wanted to confirm is whether `provider` (the right-hand side) is itself canonical, otherwise a request for `z.ai` against active `zai` would be wrongly blocked. It is — `provider` gets normalized at the top of the function at `api/routes.py:13010`:

```python
from api.config import _resolve_provider_alias
provider = _resolve_provider_alias(provider)
```

So both sides of the equality run through the same WebUI-owned resolver (`api/config.py:962`, not a hermes_cli import — no cross-repo dependency risk). The `test_standard_provider_live_fetch_allows_matching_active_provider_alias` case exercises exactly this: active `model.provider = "z.ai"`, request `provider=zai`, and asserts the live fetch to `https://api.z.ai/v1/models` proceeds with `Bearer active-provider-canary`. Good — the legitimate alias case isn't a casualty of the fix.

## Behavior boundary is right

- Provider-scoped `providers[provider].api_key` still takes precedence (unchanged).
- Cross-provider with no provider-scoped key now falls through to the static `_PROVIDER_MODELS` catalog instead of a credentialed fetch — `test_standard_provider_live_fetch_does_not_reuse_active_provider_key` asserts `urlopen` is never called and a static list still comes back, so the UX degradation is "no live enrichment," not "broken dropdown." That's the correct tradeoff for a credential boundary.

## Threat model is accurate, severity reasonable

The pre-fix `_key = cfg.get("model", {}).get("api_key")` unconditional fallback did send the active provider's bearer token to a different fixed provider's `/models` endpoint. The CVSS 4.3 (PR:L, C:L only, fixed endpoints not arbitrary SSRF) reads fair to me — it's credential disclosure to the wrong known third party, bounded by the static endpoint table, and requires an authenticated reachable WebUI. Not over-claimed.

One tiny nit, non-blocking: the secondary commit dropping the redundant `_resolve_provider_alias` import in the `custom:` branch (`routes.py:13127`) is a nice cleanup since the symbol is now imported once at function scope — worth a one-line note in the PR description so reviewers know the second commit is pure dedup, not behavior.

CI is green across all three Python versions; `tests/test_byok_model_dropdown.py` reports 23 passed. No blockers.

nesquena-hermes:
Shipped in v0.51.522 via the rebased release branch (credited to you in CHANGELOG + release notes). Thanks @Hinotoi-agent! /api/models/live now gates the top-level model.api_key fallback on the active provider resolving to the requested provider, so one provider's key can't be forwarded to another's endpoint. Codex+Opus SAFE + full suite green (9622).

PR #4491, opened 2026-06-19
Fix multi-container gateway URL configuration

### LINKED ISSUE #4483: Can't access CRON jobs in two-container mode
It seems that it relies on the hermes web API at port 8642 in order to view and manage cron jobs. This is not enabled by default causing an error to show up. 

Solution is to add following to the `environment` section of the `hermes-agent` service:
```
- API_SERVER_ENABLED=true
- API_SERVER_HOST=0.0.0.0
```

That tells hermes to start the web api server so it can be accessed by hermes-webui

franksong2702:
## Thinking Path

Issue #4483 reported that the two-container Docker setup shows cron/gateway trouble. The maintainer clarification separates two paths: cron listing is local from the shared `HERMES_HOME`, while scheduled ticking and the Tasks/System gateway health pill need the WebUI container to reach the gateway container.

The shipped multi-container compose files already run a `hermes-agent` gateway service, but the WebUI service did not include a compose-network gateway URL. That leaves the WebUI falling back to local PID/state signals from its own container, which cannot reliably see the separate gateway container.

## What Changed

- Added `HERMES_API_URL=http://hermes-agent:8642` to the `hermes-webui` service in both multi-container compose files.
- Expanded `docs/docker.md` to clarify that cron listing is local, while scheduled ticking and gateway status need a reachable gateway URL.
- Documented the custom-compose fix on the WebUI service side and warned that `API_SERVER_ENABLED=true` / `API_SERVER_HOST=0.0.0.0` alone is incomplete without `API_SERVER_KEY`.
- Added a Docker invariant regression test to keep the multi-container WebUI service pointed at the gateway service.
- Updated the unreleased changelog.

## Why It Matters

This makes the recommended two- and three-container examples work with the gateway health/status path by default, instead of requiring users to discover and add the WebUI-side gateway URL manually.

It also prevents users from applying the incomplete agent-side API server snippet as the only fix, which can leave the agent refusing to start when `API_SERVER_KEY` is missing.

## Verification

- `git diff --check`
- PyYAML sanity check that both multi-container compose files parse and include `HERMES_API_URL=http://hermes-agent:8642` under `hermes-webui.environment`
- `./scripts/test.sh tests/test_v050260_docker_invariants.py tests/test_docker_docs_and_readonly.py tests/test_issue3012_3006_docker_docs.py`
  - 31 passed

## Risks / Follow-ups

- This is a config/docs/static-test fix. I did not boot Docker services locally in this run.
- Custom compose users still need to apply the documented WebUI-side gateway URL if they are not using the checked-in compose files.

Fixes #4483.

## Model Used

OpenAI GPT-5.3 Codex; Hermes WebUI Coordinator Autopilot with GitHub CLI and local shell tooling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Two- and three-container Docker examples now point the WebUI at the gateway service by default (#4483).** The compose files set `HERMES_API_URL=http://hermes-agent:8642`, and the Docker docs clarify that cron listing is local while scheduled ticking and gateway status need a reachable gateway URL. The docs also warn that enabling the agent API server requires `API_SERVER_KEY`, not just `API_SERVER_ENABLED`.
+
 ## [v0.51.513] — 2026-06-19 — Release RX (credential-pool quota status for all pooled providers)
 
 ### Added
--- a/docker-compose.three-container.yml
+++ b/docker-compose.three-container.yml
@@ -122,6 +122,10 @@ services:
       - HERMES_WEBUI_HOST=0.0.0.0
       - HERMES_WEBUI_PORT=8787
       - HERMES_WEBUI_STATE_DIR=/home/hermeswebui/.hermes/webui
+      # Let the WebUI health/status checks reach the gateway container over
+      # the compose network. Cron listing is local, but scheduled ticking and
+      # the Tasks/System gateway pill need a reachable gateway URL (#4483).
+      - HERMES_API_URL=http://hermes-agent:8642
       # Match your host user's UID/GID for correct file permissions.
       # Run `id -u` and `id -g` to find your values.
       # On macOS, UIDs start at 501 (not 1000) — set these in a .env file:
--- a/docker-compose.two-container.yml
+++ b/docker-compose.two-container.yml
@@ -105,6 +105,10 @@ services:
       - HERMES_WEBUI_HOST=0.0.0.0
       - HERMES_WEBUI_PORT=8787
       - HERMES_WEBUI_STATE_DIR=/home/hermeswebui/.hermes/webui
+      # Let the WebUI health/status checks reach the gateway container over
+      # the compose network. Cron listing is local, but scheduled ticking and
+      # the Tasks/System gateway pill need a reachable gateway URL (#4483).
+      - HERMES_API_URL=http://hermes-agent:8642
       # Match your host user's UID/GID for correct file permissions.
       # In two-container setups the WebUI auto-detects UID/GID from the shared
       # hermes-home volume, but you can override explicitly if needed (#668):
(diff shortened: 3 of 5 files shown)


nesquena-hermes:
Pulled the branch into a read-only worktree and checked the config + docs change against the WebUI gateway-health code path. This is the right fix for #4483 and the env var is wired correctly.

## The added env var is the one the health probe actually reads

`HERMES_API_URL=http://hermes-agent:8642` lands under `hermes-webui.environment` in both compose files. That's exactly the variable the WebUI gateway-health resolver consumes — `api/agent_health.py:319` `_remote_gateway_base_url()` walks this priority list:

```python
for var in (
    "GATEWAY_HEALTH_URL",
    "HERMES_GATEWAY_HEALTH_URL",
    "HERMES_API_URL",
    "HERMES_WEBUI_GATEWAY_BASE_URL",
):
```

and when any is set, `_collect()` short-circuits to an HTTP probe before touching local PID/state files (`api/agent_health.py:466-473`). Without it, a WebUI container with no `gateway` Python package falls into the `gateway_not_configured` branch and shows the spurious banner described in #4483. So the System/Tasks pill genuinely needs this var in the WebUI container — the change addresses the real cause, not a symptom.

## Service name + port line up

`hermes-agent` is the gateway service/`container_name` in `docker-compose.two-container.yml:49,51`, published on `8642` (`:57`), and other containers on the compose network reach it at `http://hermes-agent:8642` per the inline comment at `:55`. The three-container file matches. The `host`-side `Verify` snippet's default also moves from `http://hermes:8642` to `http://hermes-agent:8642`, which is consistent with the actual service name — the old `hermes` default would have failed name resolution.

## Docs clarification matches the code

The new doc paragraph ("the cron list itself is still read from the shared `HERMES_HOME` volume, not from the gateway HTTP API") is accurate and matches what I see for #4483 — the list endpoint reads local cron state, while scheduled *ticking* + the gateway pill need the reachable gateway URL. Splitting those two concerns in the troubleshooting section is the right framing; users were conflating "job list loads" with "gateway healthy."

The `API_SERVER_KEY` warning is also worth keeping — the agent refuses to start with `API_SERVER_ENABLED=true` but no key, so steering users away from the half-applied agent-side snippet prevents a follow-up "agent won't boot" report.

## One small suggestion (non-blocking)

The regression test (`tests/test_v050260_docker_invariants.py`) asserts the literal `HERMES_API_URL=http://hermes-agent:8642` string is present. That's a fine invariant, but consider also asserting it sits under the `hermes-webui` service rather than anywhere in the file, so a future edit that accidentally moves it under `hermes-agent` would fail. Not a blocker — CI is green across all three smoke variants (single / two / three-container) plus compose-config validation.

nesquena-hermes:
Shipped in **v0.51.517** (Release SB) — thanks @franksong2702. The two/three-container compose files now set `HERMES_API_URL=http://hermes-agent:8642` so multi-container deployments reach the gateway (cron jobs included) and don't show the spurious 'gateway not configured' banner. I self-rebased onto current master (CHANGELOG-only conflict; code byte-identical). Gate: Codex+Opus SAFE + full suite green (9610).

nesquena-hermes:
Shipped in v0.51.517 via the rebased release branch (credited to you in CHANGELOG + release notes). Thanks @franksong2702!

PR #4492, opened 2026-06-19
fix: reconcile stale active WebUI runs

ai-ag2026:
## Summary
- mark detached/cancelled WebUI workers as `cancelling` in `ACTIVE_RUNS`
- reconcile stale same-session `ACTIVE_RUNS` zombies once they exceed the unwind ceiling
- add regression coverage for recovery/health no longer seeing half-alive runs forever

## Why
A cancelled stream can remove its SSE state while the agent worker is still unwinding. If that worker remains registered in `ACTIVE_RUNS` past the recovery window, browser recovery and health polling can continue to see a half-alive run even though new turns are allowed.

## Validation
- RED: `./scripts/test.sh tests/test_stale_stream_cleanup.py::test_chat_start_not_permanently_blocked_by_stale_active_run tests/test_cancel_interrupt.py::TestCancelInterrupt::test_cancel_falls_back_to_active_run_registry -q` failed before the fix
- GREEN: same focused tests now pass: `2 passed`
- Focused suite: `./scripts/test.sh tests/test_stale_stream_cleanup.py tests/test_cancel_interrupt.py -q` → `20 passed`
- `git diff --check`
- added-line public hygiene scan for local/private markers → `added_marker_hits 0`

Notes: `ruff check` on the touched large modules still reports existing baseline issues outside this patch's changed hunks.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -14125,8 +14125,9 @@ def _active_run_stream_for_session(session_id: str | None) -> str | None:
     now = time.time()
     try:
         from api import config as _live_config
+        stale_stream_ids = []
         with _live_config.ACTIVE_RUNS_LOCK:
-            for run_stream_id, raw in (_live_config.ACTIVE_RUNS or {}).items():
+            for run_stream_id, raw in list((_live_config.ACTIVE_RUNS or {}).items()):
                 stream_id = str((raw or {}).get("stream_id") or run_stream_id or "").strip()
                 run_sid = str((raw or {}).get("session_id") or "").strip()
                 if run_sid != sid or not stream_id:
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
Shipped in **v0.51.519** (Release SD) — thanks @ai-ag2026. Stale active-run registry entries are now reconciled so health/recovery polling stops seeing half-alive runs.

On the way in the gate (Codex/Opus split) flagged that reaping on run-age alone could pop a genuinely-live worker that's mid-teardown past the 180s ceiling — `ACTIVE_RUNS` is worker-lifecycle truth for health/background-wakeup/agent-cache consumers. I added a `not in STREAMS` guard (snapshot taken sequentially before `ACTIVE_RUNS_LOCK`, no nesting) so only genuinely-gone workers are reaped, kept your anti-permanent-409 age behavior intact, and added a toothed regression test. Full Codex+Opus+suite green (9613).

nesquena-hermes:
Shipped in v0.51.519 via the rebased release branch (your fix + a gate-found 'not in STREAMS' guard on top, credited to you in CHANGELOG + release notes). Thanks @ai-ag2026!

Issue #1431, opened 2026-05-01
Composer footer: redesign session-toolsets chip placement (currently hidden)


nesquena-hermes:
### Background

The composer footer is currently hosting (left to right):

- Model picker
- Reasoning-effort chip
- Profile dropdown
- Session toolsets chip (`composerToolsets*`, introduced in #493)
- Context-usage ring (right side)
- Send button

On narrower widths (or with longer model labels like "Claude Sonnet 4.6 xhigh + Anthropic Direct"), the toolsets chip overlaps the context ring and pushes everything together. Screenshot in the original report showed the "Global (default)" chip getting clipped right next to the 100% context ring.

### Decision

For now, the toolsets chip is **hidden** in the composer footer. The underlying state, dropdown, and `/api/session/toolsets` endpoint all still work — only the visual entry point is hidden. Any cron job or scripted client that already passes `enabled_toolsets` continues to function unchanged.

Done in PR (link to follow): one-line change in `static/ui.js::_applyToolsetsChip()` sets `wrap.style.display = 'none'`.

### What this issue tracks

A proper redesign of the composer footer that finds a permanent home for session-level tool restriction. Possible directions:

1. **Move into the existing toolbar overflow menu** (the gear/`…` button if there is one) so it's discoverable but not on the critical path.
2. **Surface only when active** — keep hidden when restrictions match the global default, fade in as a small badge only when the user has applied a restriction (so it acts as a "you are in a restricted session" indicator rather than a toggle).
3. **Move under the profile dropdown** — toolsets are often profile-shaped concepts already, and this would give the dropdown a natural sub-menu surface for "this session" overrides.
4. **Long-press / right-click** the model chip to open it (power-user surface, removes from footer entirely).

No strong preference yet. Whoever picks this up should sketch 2-3 of the above before committing.

### Restoring the chip locally

If anyone needs the chip back temporarily (e.g. to debug):

```js
// In dev tools:
document.getElementById('composerToolsetsWrap').style.display = '';
```

Or revert the one-line change in `_applyToolsetsChip()`.

### Acceptance criteria

- [ ] Composer footer doesn't visually crowd context-ring + send button at standard widths
- [ ] Session toolset restriction is reachable from the UI (not just `/api/session/toolsets`)
- [ ] Active restriction is visible at a glance when set (so you don't forget you're in a restricted session)
- [ ] Works at the narrow end of supported viewports

Labels: enhancement, ui, good-first-issue

[...]

Issue #493, opened 2026-04-14
feat(session): enable/disable toolsets per session from the Web UI


nesquena-hermes:
## Summary

The Hermes CLI supports enabling and disabling toolsets per session (e.g. `--toolsets terminal,file`). The Web UI has no equivalent — every session starts with the same full toolset.

## Proposed behaviour

Add a toolset selector to the session settings or the composer so users can restrict which toolsets are active for a given session. Useful for sandboxing — e.g. a session that should only read files and not run terminal commands.

## Implementation notes

- The agent accepts `enabled_toolsets` / `disabled_toolsets` at construction time in `api/streaming.py`
- UI: a chip/tag picker or checkbox list in session settings or the new-session flow
- Persist the selection to the session object so it survives page reload
- Show active toolsets somewhere visible (topbar or composer badge)

Tracked in ROADMAP.md as "Enable/disable toolsets per session (deferred)".

nesquena-hermes:
## Acknowledged: Per-session toolset selector ✅

Good sandboxing feature — restricting toolsets per session is particularly useful for "read-only" or "analysis-only" sessions where you don't want the agent running terminal commands.

### Implementation notes

**Backend** (`api/streaming.py`):
- The agent constructor already accepts `enabled_toolsets` / `disabled_toolsets` parameters
- The session object needs a `toolsets` field: `{enabled: ["terminal", "file", ...]}` or `{disabled: ["terminal"]}`
- When building the agent for a session, pass the stored toolset config from the session object
- New endpoint: `POST /api/session/<id>/toolsets` to update the selection; persisted to session storage

**Frontend**:
- A chip/tag multi-select in the session settings drawer (⚙ icon on the session)
- Chips for each available toolset, toggleable — green = enabled, dimmed = disabled
- Show active toolsets as a subtle badge strip in the topbar or composer (e.g. `🔧 file · terminal · web`)
- "All toolsets" is the default (no restriction) — the badge strip is hidden unless at least one is excluded

**New session flow**: the new session dialog could include a quick toolset preset selector (e.g. "Full", "Read-only", "Web only") alongside the model picker.

Tracked in ROADMAP.md as "deferred" but a good sprint candidate for sessions that need sandboxing.

Buscemi:
### Problem Description

In Docker deployments of Hermes WebUI, the environment variable `HERMES_WEBUI_DEFAULT_WORKSPACE` is often **ignored**.  
[...]

Issue #2518, opened 2026-05-18
New Conversation button appears unresponsive during cold model catalog resolution


franksong2702:
## Bug Description

Clicking the sidebar `+` / New Conversation button can appear unresponsive on a cold WebUI server. If the user clicks several times during that slow window, the backend accepts multiple `/api/session/new` requests. Once the model/provider catalog has warmed up, the same button becomes fast again.

## Evidence

Local diagnostics showed that the click did reach the server, but the first wave of `/api/session/new` requests waited behind synchronous model/provider resolution:

- repeated new-session clicks produced `/api/session/new` latencies around `17.8s`, `16.0s`, `12.3s`, `10.7s`, and `8.9s`
- after the cache warmed, later `/api/session/new` requests dropped to millisecond-scale responses
- the slow path overlaps model/provider catalog resolution, including auth probing in the provider/model path
- a nearby `/api/chat/start` request also showed a cold `resolve_model_provider` span around 15s, matching the same warm-up pattern

Code path observed during triage:

```text
static/sessions.js:newSession()
POST /api/session/new
api/routes.py:_session_model_state_from_request(...)
api/routes.py:_resolve_compatible_session_model_state(...)
api/models.py:get_available_models(...)
provider/auth status probing
```

## Expected Behavior

- The New Conversation button should give immediate feedback when creation is already in progress.
- Repeated clicks during an in-flight create should not enqueue multiple new sessions.
- Cold model/provider cache work should not make the UI look like the click was ignored.

## Actual Behavior

- The first click can appear to do nothing for several seconds.
- Users naturally click again.
- Multiple concurrent `/api/session/new` requests can be created before the first response returns.
- Once the model/provider cache is warm, the same flow becomes fast, making the issue feel intermittent.

## Suggested Fix Scope

A small safe slice would be:

1. Add a frontend in-flight guard around `newSession()` so only one create request is active at a time.
2. Disable or mark the New Conversation controls as busy while the request is pending.
3. Keep the backend/model-catalog decoupling as a follow-up if needed; the click-duplication guard is independently useful and low-risk.

## Environment

- Hermes WebUI local development server
- Browser sidebar New Conversation / `+` button
- Cold provider/model catalog cache or cold auth-status probing path

nesquena-hermes:
## Summary

[...]

PR #4493, opened 2026-06-19
Fix pre-session toolsets staging

### LINKED ISSUE #4490: bug(toolsets): composer-footer chip is a dead click on a brand-new session — pre-session toolset override unreachable
## Summary

The composer-footer **Tools** chip (the wrench icon, rightmost in the composer footer at wide widths) does nothing when clicked on a freshly loaded page — i.e. before the user has sent their first message. Once the first turn is in flight and `S.session` is populated, the same click works normally.

This breaks the documented use case of #493 (per-session toolset override): pick a custom toolset for the next session, then send the message and have the agent run with that toolset. Today the picker is dead before the session exists.

## Reproduction (upstream/master ccae754c)

1. Open the WebUI in a wide-enough window (composer footer ≥1100px so the chip is visible).
2. **Before sending any message**, click the Tools chip (labeled "Global" or "Active profile def…").
3. **Observed:** nothing happens. Modal does not open. No console error. No toast.
4. Send a single message — any prompt.
5. After the agent's first turn begins (a session now exists), click the chip again.
6. **Observed:** modal opens normally. Apply/Clear work.

Steps 2–3 are the bug. The user can never commit a pre-session toolset override through the UI today.

## Root Cause

Two cooperating gaps.

**Gap 1 — `static/ui.js:3267-3271` (public master ccae754c): `toggleToolsetsDropdown` early-returns on `!S.session`.**

```js
function toggleToolsetsDropdown() {
  const dd = $('composerToolsetsDropdown');
  const chip = $('composerToolsetsChip');
  if (!dd || !chip) return;
  if (typeof S === 'undefined' || !S || !S.session) return;   // ← dead-click trigger
  if (chip.offsetParent === null) return;
  …
}
```

`S.session` is `null` on a freshly loaded page (`static/boot.js:2192`, `2224`, plus `sessions.js` / `messages.js` reset sites). The chip itself is in the DOM (`static/index.html:680`, `onclick="toggleToolsetsDropdown()"`) and visible at wide composer-footer widths via the #1431 responsive CSS, but every click silently no-ops until `S.session` becomes truthy.

Sibling composer dropdowns have no such guard: `toggleWsDropdown` (`panels.js:4772`), `toggleProfileDropdown` (`panels.js:5534`), `toggleModelDropdown` (`ui.js:2390`), and `toggleReasoningDropdown` all open pre-session without issue. The toolsets chip is the only composer dropdown that refuses to open on a brand-new page.

**Gap 2 — `static/ui.js:3305-3307` + `api/routes.py:8744` + `api/models.py:2462`: no path to commit a pre-session toolset.**

Even if Gap 1 is fixed (the user can open the picker), the user still cannot commit the selection before a session exists:

```js
function _applySessionToolsets(toolsets) {
  if (typeof S === 'undefined' || !S || !S.session) return;   // ← no-op if no session
  const sid = S.session.session_id;
  api('/api/session/toolsets', { … })
```

The `/api/session/toolsets` POST handler (`api/routes.py:9101`) requires a `session_id` and returns 404 if it doesn't exist. On a brand-new page there is no `session_id` to send — the only way to get one is to POST `/api/session/new` and commit the toolset via the new-session response.

But `/api/session/new` does **not** accept an `enabled_toolsets` field:

- `api/routes.py:8744` reads `workspace`, `model`, `model_provider`, `profile`, `project_id`, `worktree` — but not `enabled_toolsets`.
- `api/models.py:2462` (`new_session(workspace=None, model=None, profile=None, model_provider=None, project_id=None, worktree_info=None)`) accepts no `enabled_toolsets` kwarg.
- The client-side `reqBody` builder (`static/sessions.js:688`) carries `workspace`, `profile`, `model`, `model_provider`, `project_id`, `worktree` into the new-session request, but not the staged toolsets value.

So even if a user could open the chip and type a toolset list, the selection has nowhere to go until after the session is created — at which point the user must re-open the chip and re-apply it.

## Impact

- The chip's documented purpose (per-session override before any tool use) is unreachable.
- No feedback at all on the dead click — silent no-op, no toast, no console error.
- Inconsistent with the four other composer dropdowns (profile / workspace / model / reasoning) which all work pre-session.
- Latent conflict with #3100 / PR #4287 (active-profile defaults surface): users opening a fresh session can never discover or interact with that affordance.

## Suggested Fix Shape

Three coordinated, small changes — each is atomic and individually testable.

**1. `static/ui.js:3267-3271` — drop the `!S.session` early-return.**

`_populateToolsetsDropdown()` already handles the no-session case via its "Active profile defaults" branch (it reads only module-level `_currentSessionToolsets`, not `S.session`). Keep only the `chip.offsetParent === null` guard from #1431.

**2. `static/ui.js:3305` — stage the Apply/Clear result when no session exists.**

Add a module-level `_pendingSessionToolsets` (parallel to `_currentSessionToolsets`). When `_applySessionToolsets` runs without `S.session`, set `_pendingSessionToolsets` and toast ("Toolsets staged — will apply to your next session"). On Clear, stage `null`. The chip label should reflect staged state (e.g. `🔧 outline, zabbix (staged)`) so the user sees the selection is captured.

**3. `static/sessions.js:688` + `api/routes.py:8744` + `api/models.py:2462` — thread staged toolsets into `/api/session/new`.**

- `api/models.py:2462`: extend `new_session(...)` to accept `enabled_toolsets=None` and pass it to the `Session(...)` constructor (the model already accepts the field — `api/models.py:645`, `700`).
- `api/routes.py:8744`: read `body.get("enabled_toolsets")`, validate with the same rules as `/api/session/toolsets` (`api/routes.py:9113-9117`: non-empty list of non-empty strings, or `null`), pass it into `new_session(...)`. On validation failure return 400 with the same error wording.
- `static/sessions.js:688`: when `_pendingSessionToolsets` is set, add `enabled_toolsets: _pendingSessionToolsets` to `reqBody`. After the successful `S.session = data.session` assignment (`static/sessions.js:737`), reset `_pendingSessionToolsets = null` so subsequent sessions don't carry stale staging.

This mirrors the existing pattern for `model` (read from the visible picker, carried into `/api/session/new` via `reqBody.model` at `static/sessions.js:706`) — the staged-toolset flow is the same shape with a different field.

## Regression Coverage

A new test module (e.g. `tests/test_issue_toolsets_presession.py`) with:

- `toggleToolsetsDropdown` does NOT early-return when `S.session` is null.
- `_applySessionToolsets([...])` with no `S.session` stages into `_pendingSessionToolsets` and toasts; does NOT POST `/api/session/toolsets`.
- `new_session(enabled_toolsets=['x'])` returns a `Session` with `s.enabled_toolsets == ['x']`.
- `POST /api/session/new` with `{"enabled_toolsets":["x","y"]}` returns a session whose `enabled_toolsets == ["x","y"]`.
- `POST /api/session/new` rejects malformed `enabled_toolsets` (empty list, non-string items, non-list) with 400 in the same shape as `/api/session/toolsets`.
- `POST /api/session/new` without `enabled_toolsets` returns `s.enabled_toolsets is None`.
- The client-side `reqBody` builder includes `_pendingSessionToolsets` when set, omits when null.
- After successful `POST /api/session/new`, `_pendingSessionToolsets` is reset to null.
- The chip label reflects staged state when no session exists.
- All 13 invariants from `tests/test_issue1431_toolsets_chip_responsive.py` still pass (the #1431 responsive-CSS contract is preserved).

## Out of Scope

- `composer_draft` and other deferred-session fields — separate concerns.
- Backend pre-validation against the catalog of configured toolset names. Today's `/api/session/toolsets` doesn't validate against the catalog either (the streaming layer filters unknown names), and pre-session staging should match.
- Changing the existing post-session `/api/session/toolsets` validation or apply path.
- The #1431 responsive-CSS contract — preserved as-is.

## Why a separate issue, not a follow-up to #1431

The #1431 fix addressed visibility (CSS-only contract, chip hidden at narrow widths). This bug is about clickability and commit semantics — the chip is visible and responsive to clicks after the first turn, but inert before it. They share no code path beyond `toggleToolsetsDropdown` reading the same chip element.

franksong2702:
## Thinking Path

Issue #4490 reports that the composer-footer Tools chip is visible on a brand-new page but silently no-ops because `toggleToolsetsDropdown()` requires `S.session`.

The fix needs to cover both halves of the problem: the UI must allow choosing toolsets before a session exists, and the staged choice must have a backend path into `/api/session/new`. The issue follow-up also calls out an important leak guard: staged toolsets must be cleared on workspace/profile context switches so a choice made under one empty-context setup does not silently apply to another.

## What Changed

- Let the Tools dropdown open with no active session while preserving the #1431 hidden-chip `offsetParent === null` guard.
- Stage pre-session Apply/Clear in `S._pendingSessionToolsets` and show staged custom toolsets on the chip with a `(staged)` suffix.
- Send staged toolsets through `newSession()` as `enabled_toolsets`, then clear the staged value after successful session creation.
- Extend `api.models.new_session(...)` and `/api/session/new` to accept `enabled_toolsets`.
- Share the structural validation between `/api/session/new` and `/api/session/toolsets`: null or a non-empty list of non-empty strings.
- Clear staged toolsets on workspace/profile context-switch session paths so staged values do not leak across contexts.
- Add regression coverage for the frontend staging contract, backend session creation, malformed input rejection, #1431 responsive guard preservation, and the context-switch reset behavior.
- Update `CHANGELOG.md`.

## Why It Matters

This restores the intended #493 workflow: choose a custom toolset list before the first prompt, send the first message, and have the new session start with that per-session override.

It also avoids the dangerous inverse bug from the issue follow-up: a staged value from one empty composer context should not survive a workspace/profile switch and unexpectedly affect a later unrelated session.

## Verification

- `./scripts/test.sh tests/test_issue4490_presession_toolsets.py tests/test_issue3100_profile_toolsets_affordance.py tests/test_issue1431_toolsets_chip_responsive.py tests/test_v050257_opus_followups.py`
  - 36 passed
- `node --check static/ui.js`
- `git diff --check`
- `./.venv/bin/ruff check tests/test_issue4490_presession_toolsets.py`
- `./.venv/bin/python -m py_compile api/models.py api/routes.py tests/test_issue4490_presession_toolsets.py`
- Diff-scoped Ruff check on the Python backend diff passed.
- Claude Code CLI was used as a read-only diff reviewer; it reported no blocking findings. One non-blocking suggestion to forward staged toolsets through workspace auto-session paths was intentionally not adopted because the #4490 follow-up comment explicitly requires reset on those context-switch paths.

## Risks / Follow-ups

- I did not capture browser screenshots/video in this run; the behavior is covered with focused frontend structural tests and backend behavioral tests.
- The `(staged)` suffix is intentionally small and local to the wide composer chip. It does not add new composer controls or alter the responsive visibility contract.

Fixes #4490.

## Model Used

OpenAI GPT-5.3 Codex as coordinator, with a Codex sub-agent worker for implementation and Claude Code CLI 2.1.168 as a read-only diff reviewer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **The composer Tools chip can now stage per-session toolsets before the first conversation exists (#4490).** Applying a custom toolset from the empty composer stores it for the next new session instead of silently no-oping, `/api/session/new` now accepts the staged `enabled_toolsets` value with the same structural validation as `/api/session/toolsets`, and staged values are cleared on workspace/profile context switches so they do not leak into unrelated sessions.
+
 ## [v0.51.513] — 2026-06-19 — Release RX (credential-pool quota status for all pooled providers)
 
 ### Added
--- a/api/models.py
+++ b/api/models.py
@@ -2459,7 +2459,7 @@ def _profile_default_model_state(profile=None):
     return default_model or get_effective_default_model(), default_provider
 
 
-def new_session(workspace=None, model=None, profile=None, model_provider=None, project_id=None, worktree_info=None):
+def new_session(workspace=None, model=None, profile=None, model_provider=None, project_id=None, worktree_info=None, enabled_toolsets=None):
     """Create a new in-memory session.
 
     The session lives in the SESSIONS dict only — no disk write happens until
@@ -2512,6 +2512,7 @@ def new_session(workspace=None, model=None, profile=None, model_provider=None, p
         worktree_branch=wt.get('branch') if wt else None,
         worktree_repo_root=wt.get('repo_root') if wt else None,
         worktree_created_at=wt.get('created_at') if wt else None,
+        enabled_toolsets=enabled_toolsets,
     )
     with LOCK:
         SESSIONS[s.session_id] = s
--- a/api/routes.py
+++ b/api/routes.py
@@ -8603,6 +8603,16 @@ def _require_passkey_registration_auth(handler) -> tuple[bool, str, int]:
         return False, "Authentication required", 401
     return True, "", 200
 
+def _validate_session_toolsets_shape(toolsets):
+    """Validate per-session toolset override shape without catalog lookup."""
+    if toolsets is None:
+        return None
+    if not isinstance(toolsets, list) or not toolsets:
+        raise ValueError("toolsets must be a non-empty list or null")
+    if not all(isinstance(t, str) and t for t in toolsets):
+        raise ValueError("each toolset must be a non-empty string")
+    return toolsets
+
 def handle_post(handler, parsed) -> bool:
     """Handle all POST routes. Returns True if handled, False for 404."""
     diag = RequestDiagnostics.maybe_start("POST", parsed.path, logger=logger)
@@ -8768,6 +8778,10 @@ def handle_post(handler, parsed) -> bool:
             body.get("model"),
             body.get("model_provider"),
         )
+        try:
+            enabled_toolsets = _validate_session_toolsets_shape(body.get("enabled_toolsets"))
+        except ValueError as e:
+            return bad(handler, str(e), status=400)
         # Use the profile sent by the client tab (if any) so that two tabs on
         # different profiles never clobber each other via the process-level global.
         # ── Memory lifecycle: commit the previous session before starting a new one ──
@@ -8791,6 +8805,7 @@ def handle_post(handler, parsed) -> bool:
             profile=body.get("profile") or None,
             project_id=body.get("project_id") or None,
             worktree_info=worktree_info,
+            enabled_toolsets=enabled_toolsets,
         )
         if worktree_info:
             publish_session_list_changed(
@@ -9110,12 +9125,10 @@ def handle_post(handler, parsed) -> bool:
             return bad(handler, str(e))
         sid = body["session_id"]
         toolsets = body.get("toolsets")
-        # Validate: if not None, must be a non-empty list of strings
-        if toolsets is not None:
-            if not isinstance(toolsets, list) or not toolsets:
-                return bad(handler, "toolsets must be a non-empty list or null")
-            if not all(isinstance(t, str) and t for t in toolsets):
-                return bad(handler, "each toolset must be a non-empty string")
+        try:
+            toolsets = _validate_session_toolsets_shape(toolsets)
+        except ValueError as e:
+            return bad(handler, str(e), status=400)
         try:
             s = get_session(sid)
         except KeyError:
--- a/static/panels.js
+++ b/static/panels.js
@@ -5448,7 +5448,7 @@ async function promptWorkspacePath(){
     if(!ws)return;
     try{
       const r=await api('/api/session/new',{method:'POST',body:JSON.stringify({workspace:ws})});
-      if(r&&r.session){S.session=r.session;S.messages=[];if(typeof syncTopbar==='function')syncTopbar();if(typeof renderMessages==='function')renderMessages();if(typeof renderSessionList==='function')await renderSessionList();}
+      if(r&&r.session){S._pendingSessionToolsets=null;S.session=r.session;S.messages=[];if(typeof syncTopbar==='function')syncTopbar();if(typeof renderMessages==='function')renderMessages();if(typeof renderSessionList==='function')await renderSessionList();}
     }catch(e){showToast(t('workspace_switch_failed')+e.message);return;}
     if(!S.session)return;
   }
@@ -5484,7 +5484,7 @@ async function switchToWorkspace(path,name){
     if(!ws){showToast(t('no_workspace'));return;}
     try{
       const r=await api('/api/session/new',{method:'POST',body:JSON.stringify({workspace:ws})});
-      if(r&&r.session){S.session=r.session;S.messages=[];if(typeof syncTopbar==='function')syncTopbar();if(typeof renderMessages==='function')renderMessages();if(typeof renderSessionList==='function')await renderSessionList();}
+      if(r&&r.session){S._pendingSessionToolsets=null;S.session=r.session;S.messages=[];if(typeof syncTopbar==='function')syncTopbar();if(typeof renderMessages==='function')renderMessages();if(typeof renderSessionList==='function')await renderSessionList();}
     }catch(e){if(typeof setStatus==='function')setStatus(t('switch_failed')+e.message);return;}
     if(!S.session)return;
   }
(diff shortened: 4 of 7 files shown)


nesquena-hermes:
Reviewed this against the linked #4490 report by pulling the branch into a read-only worktree and diffing `origin/master...HEAD`. The fix matches the issue's prescribed shape closely, and the leak-guard coverage that the follow-up comment flagged as **mandatory** is actually complete — more complete than the summary suggests.

## Leak guards: all four `/api/session/new` sites covered

The #4490 follow-up named four workspace-switch builders plus the profile-switch path as required reset points. On `origin/master`, the four lazy-session sites are `panels.js:5450`, `panels.js:5486`, `ui.js:13808`, `ui.js:13835`. The diff resets `S._pendingSessionToolsets=null` at *all* of them, not just the panels.js trio:

```js
// static/ui.js promptNewFile / promptNewFolder
if(r&&r.session){S._pendingSessionToolsets=null;S.session=r.session;S.messages=[];...}
```

plus `switchToWorkspace` (`panels.js:5512`) and `switchToProfile` (`panels.js:5812`). `test_context_switch_auto_sessions_do_not_forward_staged_toolsets` asserts each of those four function bodies both omits `enabled_toolsets` and contains the reset — good regression anchor.

## Backend validation refactor is behavior-preserving

`_validate_session_toolsets_shape` (`api/routes.py:8606`) is the same rule the inline `/api/session/toolsets` block used. One thing I checked since it looked like a behavior change: the refactor switches the existing `/api/session/toolsets` handler to `bad(handler, str(e), status=400)`. That's a no-op — `bad()` already defaults to `status=400` (`api/helpers.py:34`), so the existing endpoint's status and error wording are unchanged. The new `/api/session/new` path returns the same 400 shape.

## Empty-list path is safe

The new-session request only carries staged toolsets when `Array.isArray(S._pendingSessionToolsets)` (`sessions.js:695`), and the Apply handler already rejects empty input before staging (`ui.js`, `if (toolsets.length === 0) ... return`). Clear stages `null`. So `enabled_toolsets:[]` never reaches `/api/session/new` from the UI — the backend's empty-list 400 (covered by `test_api_session_new_rejects_malformed_enabled_toolsets`) is purely defensive. Fine.

## One minor, non-blocking note

The `isStaged` detection in `_applyToolsetsChip` (`ui.js:3095-3099`) keys off `!S.session && Array.isArray(S._pendingSessionToolsets)` rather than checking the rendered `toolsets` argument *is* the pending value. It holds for all three current call sites (the no-session `_syncToolsetsChip` branch passes the pending array, and `_applySessionToolsets` sets it immediately before calling), so there's no live bug — just an implicit invariant a future third caller could trip. Greptile already flagged the same thing as non-blocking; no change needed for this PR.

Nothing here blocks merge. CI is green across all three Python versions plus browser-smoke and lint.

nesquena-hermes:
Shipped in **v0.51.516** (Release SA) — thanks @franksong2702. The composer Tools chip now stages a per-session toolset override on a brand-new session and `/api/session/new` applies it.

On the way in I added a small follow-up the gate caught: forward the staged override only when there's no active session, and clear it in `loadSession()` — otherwise a toolset staged on the empty composer could leak into a later New Chat started from an existing session. +2 regression tests, and the `#2518` provenance-window test was widened since `newSession()` grew. Full Codex+Opus+suite green (9609 passed).

nesquena-hermes:
Shipped in v0.51.516 via the release branch (your commit + a small gate-found leak fix on top, both credited to you in the CHANGELOG and release notes). Thanks @franksong2702!

PR #4509, opened 2026-06-19
feat: add mobile conversations button

santastabber:
## Thinking Path

- Hermes WebUI already has a mobile sidebar drawer, but opening the conversations list from the titlebar hamburger can require a long reach on phones.
- A lower-screen conversations affordance is useful, but it should reuse the existing drawer behavior instead of introducing a parallel sidebar path.
- The affordance belongs in the existing lower-right transcript controls stack: above the long-message down arrow, which itself remains above Send, so it stays thumb-reachable without drifting left over message text.
- Since mobile conversation rows already support long-hold gestures, the new button also exposes a small long-press shortcut menu for common navigation actions.

## What Changed

- Added a phone-only `#mobileConversationsBtn` in the chat messages shell.
- Added `toggleMobileConversationsSidebar()` and a shared `_openMobileSidebarDrawer()` helper so hamburger, swipe, mobile rail clicks, and the new button share the same drawer path.
- Synced `aria-expanded` / `aria-label` state and hide the floating button while the sidebar drawer is already open.
- Styled the button only under the phone breakpoint (`max-width:640px`), right-aligned above the existing down/new-message affordance and send button.
- Added a long-press menu on the mobile Conversations button with:
  - New conversation
  - Open sidebar
  - Go to top
  - Go to last message
- Reused the existing `session-action-menu` visual pattern for the long-press menu, including menuitem roles, focus movement, Escape/arrow/Home/End keyboard handling, and a bounded follow-up click-ignore window after long press.
- Added static mobile layout regression coverage for presence, phone-only CSS, right-side stack positioning, 44px touch sizing, native drawer reuse, and long-press actions.

## Why It Matters

On phones, opening the conversations list from the top-left hamburger can be awkward one-handed. This gives users a lower-right thumb target for the same existing mobile drawer, keeps it in the established scroll/send control stack, and provides a long-hold shortcut menu for common conversation navigation.

## Verification

- `./scripts/test.sh tests/test_mobile_layout.py tests/test_pwa_sidebar_swipe.py tests/test_sidebar_collapse_toggle.py -q` → `100 passed`
- `./scripts/test.sh tests/test_mobile_layout.py tests/test_pwa_sidebar_swipe.py tests/test_sidebar_collapse_toggle.py tests/test_session_touch_actions.py tests/test_2111_ios_pwa_bottom_scroll_stutter.py tests/test_issue2124_outline_panel.py tests/test_issue677.py -q` → `130 passed`
- `node --check static/boot.js && node --check static/panels.js && git diff --check` → passed
- Isolated WebUI smoke: `/health` returned `status: ok`.
- Headless Chrome phone viewport DOM checks:
  - Conversations button appears in the lower-right control stack above the down arrow.
  - Forced down arrow remains above the composer/send row.
  - No overlap with the down arrow or composer.
  - Short tap opens the sidebar and does not open the menu.
  - Long press opens a menu with `New conversation`, `Open sidebar`, `Go to top`, and `Go to last message`, without opening the sidebar.
  - The synthetic click that follows a mobile long-press is suppressed while the menu remains open.
  - The contextmenu path cancels the pending long-press timer before opening the menu, avoiding a close/reopen flicker.
  - After the bounded suppression window clears, the next normal tap opens the sidebar.
  - Pressing Escape while a menu item is focused closes the menu and returns focus to `#mobileConversationsBtn`.
  - After a normal tap, `panelChat` is active, sidebar drawer is open, overlay is visible, and `#mobileConversationsBtn` is hidden while the drawer is open.
- Full `./scripts/test.sh` was run in a clean no-agent environment: `9498 passed, 132 skipped, 1 xfailed, 2 xpassed`, with one environment-only import-path failure in `tests/test_issue803.py::test_switch_profile_process_wide_false_does_not_mutate_global`. That exact test was rerun in the normal developer environment and passed.

## Screenshots

Closed mobile chat state, button in lower-right stack above the down/send controls:

Mobile conversations button closed

After tapping the button, conversations drawer opens and the floating button is hidden:

Mobile conversations drawer open

Long-press shortcut menu:

Mobile conversations long-press menu

## Risks / Follow-ups

- The button is intentionally phone-only (`max-width:640px`); tablet/desktop users keep the existing rail/hamburger behavior.
- The icon duplicates the Chat/conversations glyph, but it routes to the existing drawer path and hides while that drawer is open to avoid competing controls.
- The long-press action labels are currently English literals except the existing `New conversation` label; this keeps the PR scoped, but these could be promoted to full i18n keys later if maintainers prefer.
- Release notes are intentionally left to the repository release process; this PR does not edit `CHANGELOG.md`.
- This PR does not add a setting to disable the affordance; it is scoped as a mobile ergonomics improvement.

## Contract Routing

Task type: UI/UX mobile navigation improvement.

Touched areas:
- `static/index.html`
- `static/style.css`
- `static/boot.js`
- `static/panels.js`
- `tests/test_mobile_layout.py`

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/UIUX-GUIDE.md`
- `TESTING.md`

Scope boundaries:
- Reuses the existing mobile sidebar drawer state and overlay.
- Does not alter desktop/tablet collapse behavior or session list data.
- Long-press menu is scoped to the mobile Conversations button and reuses the existing session action menu shape.

- Static regression tests for mobile-only behavior and long-press actions.
- Browser/DOM evidence at a phone viewport showing no overlap with composer or right-side down controls.

## Model Used

OpenAI GPT-5.5 via OpenRouter, with Hermes Agent tool use for code editing, tests, and browser/DOM verification.


santastabber:
Follow-up for the Greptile long-press comments: the latest head replaces the sticky `_mobileConversationsSuppressClick` boolean entirely with a timestamped one-shot click-ignore window (`_mobileConversationsIgnoreClickUntil`). I also removed the CHANGELOG edit from the PR. Local verification after the rewrite: focused mobile/sidebar/touch suite = `130 passed`; `node --check static/boot.js && node --check static/panels.js && git diff --check` passed; browser probe confirmed the synthetic click after long-press is ignored while the menu stays open, and the next normal tap opens the sidebar.

nesquena-hermes:
Pulled the branch into a read-only worktree and read the full `static/boot.js` diff (270 lines) plus the `index.html` / `style.css` / `panels.js` changes against `origin/master`. Overall this is a clean, well-scoped change: `_openMobileSidebarDrawer()` is a faithful extraction of the inline drawer-open block that previously lived in three places (`toggleMobileSidebar`, `_openMobileSidebarFromGesture`, `mobileSwitchPanel`), the `panels.js` `fromRailClick` branch keeps a safe inline fallback for load-order, and the FAB CSS is fully fenced under `max-width:640px` with a `display:none` desktop guard. CI is green across all 11 matrix jobs. One real bug worth fixing before merge, plus a minor i18n nit.

## The Escape focus-return bug is real — here's the exact mechanism

Greptile flagged that Escape drops focus to `<body>` rather than back to the button. Confirmed, and the root cause is a capture-vs-bubble ordering trap. The window handler is registered in **capture** phase (`static/boot.js:434`):

```js
window.addEventListener('keydown',_mobileConversationsKeyClose,true);
```

and that handler does only this (`:448-450`):

```js
function _mobileConversationsKeyClose(e){
  if(e.key==='Escape')closeMobileConversationsMenu();
}
```

The menu's *own* keydown handler `_mobileConversationsMenuKeydown` (`:451`) is the one that restores focus on Escape (`btn.focus({preventScroll:true})` at `:465`) — but it can never reach that line, for two compounding reasons:

1. It's a **bubble**-phase listener (`menu.addEventListener('keydown',...)`), so the window capture handler at `:434` always fires first.
2. The capture handler calls `closeMobileConversationsMenu()`, which sets `_mobileConversationsMenu=null` and `.remove()`s the node. So even if the bubble handler were reached, its guard `if(!_mobileConversationsMenu)return;` (`:452`) bails immediately — and the removed node won't receive the event anyway.

So `:465` is dead code on Escape. The focus-restore lives in the handler that never runs.

**Fix sketch** — move the restore into the handler that actually fires (`:448`):

```js
function _mobileConversationsKeyClose(e){
  if(e.key!=='Escape')return;
  const btn=$('mobileConversationsBtn');
  closeMobileConversationsMenu();
  if(btn&&typeof btn.focus==='function')btn.focus({preventScroll:true});
}
```

Then the `:463-466` Escape branch in `_mobileConversationsMenuKeydown` becomes redundant (the window handler beat it to the punch) and can be dropped, or kept purely as a no-harm fallback. A static assertion in `tests/test_mobile_layout.py` that `_mobileConversationsKeyClose` references `.focus(` would pin it.

## Minor (non-blocking) — menu label i18n

Three of the four long-press actions are raw English literals (`'Open sidebar'`, `'Go to top'`, `'Go to last message'`), and `syncMobileConversationsButton()` sets `aria-label` to hardcoded `'Open conversations'`/`'Close conversations'` (`:262-264`) while the sibling buttons in the same `.messages-shell` use `data-i18n-aria-label`. Only `t('new_conversation')` goes through i18n. You already called this out under Risks/Follow-ups, and per repo convention a literal-with-no-key isn't a merge blocker — just flagging that unlike a missing `t()` key, these have no key at all, so a later i18n pass has to add both the keys and the `t()` calls. Cheap to promote now while the code is fresh.

Nice work on replacing the sticky suppress-boolean with the timestamped one-shot ignore window — that's the more robust shape.

santastabber:
Addressed the Escape focus-return review: `_mobileConversationsKeyClose` now handles Escape in the capture path by preventing default, closing the menu, and returning focus to `#mobileConversationsBtn`; the duplicate/unreachable Escape branch in the menu bubble handler was removed. Verification: focused mobile/sidebar/touch suite = `130 passed`; `node --check static/boot.js && node --check static/panels.js && git diff --check` passed; browser probe confirmed Escape closes the menu and focus returns to `#mobileConversationsBtn`, and the long-press synthetic-click suppression still works.

nesquena-hermes:
Pulled the updated branch into a read-only worktree and re-read the `static/boot.js` diff against `origin/master`. The Escape focus-return bug from my last pass is **fixed correctly** — the fix landed exactly where the capture-vs-bubble analysis pointed.

## The dead-code path is gone

The window capture handler now owns both the close and the focus restore (`static/boot.js:448-455`):

```js
function _mobileConversationsKeyClose(e){
  if(e.key==='Escape'){
    e.preventDefault();
    closeMobileConversationsMenu();
    const btn=$('mobileConversationsBtn');
    if(btn&&typeof btn.focus==='function')btn.focus({preventScroll:true});
  }
}
```

Since this is registered in capture phase (`:434`, `addEventListener(...,true)`), it fires before the menu's bubble handler — and it no longer matters, because the redundant Escape branch was removed from `_mobileConversationsMenuKeydown` (`:456-470`), which now only handles ArrowUp/Down/Home/End. So there's a single Escape path, it's the one that actually runs, and focus returns to the trigger instead of falling to `<body>`. That matches what you described in the follow-up.

## Test pins both invariants

`tests/test_mobile_layout.py::test_mobile_conversations_fab_long_press_menu_actions` now asserts the focus-restore line lives in the capture handler **and** that the menu bubble handler contains no Escape branch:

```python
key_close_start = boot_js.index("function _mobileConversationsKeyClose")
menu_keydown_start = boot_js.index("function _mobileConversationsMenuKeydown", key_close_start)
resize_start = boot_js.index("function _mobileConversationsResizeClose", menu_keydown_start)
assert "e.key==='Escape'" in boot_js[key_close_start:menu_keydown_start]
assert "e.key==='Escape'" not in boot_js[menu_keydown_start:resize_start], \
    "Escape focus restoration belongs in the window capture handler, not the menu bubble handler"
```

That's the right shape — it guards against the exact regression of someone re-adding an Escape branch to the bubble handler later.

## Long-press click suppression

The earlier rewrite from the sticky `_mobileConversationsSuppressClick` boolean to the timestamped `_mobileConversationsIgnoreClickUntil` window is holding up well. `_ignoreNextMobileConversationsClick()` sets a 700ms window (`:314-316`), the click handler consumes it once via `_consumeIgnoredMobileConversationsClick()` (`:485`), and `pointerdown` clears any stale window with `_clearMobileConversationsClickIgnore()` (`:491`) so a fresh press never inherits a leftover suppression. The one-shot consume + expiry means a missed synthetic click can't permanently swallow the next real tap.

The only thing still outstanding from my prior review is the non-blocking i18n nit (three raw English menu labels + hardcoded `aria-label` in `syncMobileConversationsButton`), which you already tracked under follow-ups. No objection to landing that as a separate i18n pass. From a correctness standpoint this looks ready.

santastabber:
Addressed the contextmenu/timer race: the contextmenu handler now calls `_clearMobileConversationsPressTimer()` before opening the menu, so an OS contextmenu event cannot leave the internal 400ms long-press timer armed and reopen the menu. Verification: focused mobile/sidebar/touch suite = `130 passed`; `node --check static/boot.js && node --check static/panels.js && git diff --check` passed; browser probe simulated pointerdown then contextmenu before the timer threshold and confirmed `openCount: 1`, one menu, no sidebar open, plus Escape focus and synthetic-click suppression probes still passed.

santastabber:
Addressed the remaining ARIA note: `#mobileConversationsBtn` now declares `aria-haspopup="menu"` in the static HTML, and `tests/test_mobile_layout.py` pins that attribute. Verification: focused mobile/sidebar/touch suite = `130 passed`; `node --check static/boot.js && node --check static/panels.js && git diff --check` passed.

nesquena-hermes:
Deep-reviewed this (Codex + full suite + live mobile browser drive). On **code quality this is clean** — flagging the findings here; the product decision on whether to add the affordance is a separate maintainer call.

**Verification:**
- Full pytest suite: 9855 passed, 0 failed
- Codex review: SAFE — "no regression risk found, core flows intact." The shared `_openMobileSidebarDrawer()` helper is correctly used by all openers (hamburger, swipe, rail, and the new button), so the refactor of the existing openers doesn't regress them.
- Live drive at 390px: the FAB renders phone-only in the lower-right transcript stack, opens the existing sidebar drawer cleanly (verified with a populated 3-session transcript), 0 console errors. Correctly hidden on desktop width.

No code changes requested. Holding on the maintainer product call for whether to land the affordance itself.

santastabber:
Will move to extensions

Issue #2659, opened 2026-05-20
[bug] Settings → Plugins panel shows exclusive plugins (memory.provider / web.backend / browser provider) as "Disabled" even when fully active


Fail-Safe:
## Summary

  Plugins that activate via a `<category>.provider` config key (e.g. `memory.provider: noema`) — what hermes-cli's
  plugin manager calls "exclusive" plugins — always render as **Disabled** with **No registered lifecycle hooks**
  in the Settings → Plugins panel, even when they're the currently-active provider for their category and
  functioning correctly. The visibility panel only models the `plugins.enabled` opt-in activation path, so
  exclusive activation is invisible to it.

  ## Repro

  1. Configure a profile with an exclusive plugin via its provider key, e.g.:
     ```yaml
     memory:
       provider: noema
     plugins:
       enabled: []
  2. Ensure the plugin is installed (e.g., ~/.hermes/profiles/<name>/plugins/noema/).
  3. Open the WebUI → Settings → Plugins.
  4. Observe the plugin row: Disabled, No registered lifecycle hooks.

  Meanwhile the plugin is loaded, discovered, and actively serving as the memory provider — verifiable via:

```python
  from hermes_cli.plugins import get_plugin_manager
  m = get_plugin_manager(); m.discover_and_load(force=False)
  m._plugins["noema"].error
  # → "exclusive plugin — activate via <category>.provider config"
```

  That loaded.error string is informational from hermes-cli's loader, not a true failure — it signals "this plugin
  is exclusive; it was correctly activated through its provider key, so don't also try to load it via
  plugins.enabled."

  ## Expected

  Some way for the panel to communicate "this plugin is active via memory.provider: noema" rather than the binary
  Enabled/Disabled framing borrowed from the plugins.enabled allowlist. Either:

  - A separate active-providers section alongside the lifecycle-hook plugins section, or
  - A status badge like "Active (memory provider)" / "Active (web backend)" / etc. with no hook list (since
  exclusive plugins typically don't register the four _PLUGIN_VISIBILITY_HOOKS), or
  - A short explanatory note under exclusive plugin rows explaining why the lifecycle-hooks list is empty.

  ## Actual

  Binary "Disabled" + empty hooks list — indistinguishable from a misconfigured/broken plugin.

  ## Root cause

  `_plugin_visibility_payload()` in `source/api/routes.py` builds the panel's payload from two sources:

```python
  "enabled": bool(getattr(loaded, "enabled", False)),
  ...
[...]

PR #4517, opened 2026-06-20
Fix disabled provider plugin badge state

### LINKED ISSUE #4496: [UX] Bug in plugins state styling
In `Settings/Plugins` the enabled/disabled indicator/chip is always styled in green, making it look as if everything is enabled. However, as you can see in the screenshot, some are disabled but they are styled green, letting the user think they are enabled.

Image

franksong2702:
## Thinking Path

- The Plugins settings panel should show which plugin is actually the active provider for provider-style categories.
- #4496 identified that provider-category badge state needed an explicit active-provider signal, not inference from a generic enabled flag.
- Maintainer review caught that active exclusive providers intentionally carry `activation: "exclusive"` and `enabled: false` under the #2659 contract.
- This update preserves that active-exclusive path and makes inactive exclusive providers render as disabled.

## What Changed

- Added an explicit `is_active_provider` boolean to the `/api/plugins` payload.
- For `kind: "exclusive"` plugins, `is_active_provider` is derived from the plugin key category and the matching `<category>.provider` config value.
- For active `kind: "model-provider"` plugins, `is_active_provider` remains true when the backend already reports the plugin as enabled.
- Updated `_buildPluginCard()` to prefer `is_active_provider` for the green provider badge, with a fallback to existing `activation` semantics for older payloads.
- Removed the direct `CHANGELOG.md` edit after review feedback.
- Added regression coverage for active exclusive, inactive exclusive, and active model-provider badge rendering.

## Why It Matters

Settings no longer marks inactive exclusive provider plugins as active, while active exclusive providers such as memory providers still keep the green provider badge even though their legacy `enabled` flag is false by design.

## Verification

- Baseline: the updated #4496/#2659 regression tests failed against the previous PR head.
- `./scripts/test.sh tests/test_issue4496_plugin_badge_state.py tests/test_plugins_panel.py` (`34 passed`)
- `.venv/bin/python -m py_compile api/routes.py tests/test_issue4496_plugin_badge_state.py tests/test_plugins_panel.py`
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master --strict`
- `git diff --check`

## Risks / Follow-ups

- This changes the `/api/plugins` payload by adding a backwards-compatible field; existing `enabled`, `kind`, and `activation` fields are preserved.
- Older clients without `is_active_provider` still fall back to the existing `activation` behavior.
- No manual screenshot was captured in this run; the regression tests check the rendered badge markup directly.

Fixes #4496

## Model Used

AI-assisted with OpenAI GPT-5 Codex as coordinator. A `gpt-5.3-codex-spark` worker was briefly assigned for the review follow-up but was closed before producing usable changes; the coordinator implemented and verified the final follow-up patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 
+- **Disabled provider-category plugins now show the disabled badge in Settings (#4496).** The Plugins panel previously rendered any `provider` / `exclusive` plugin with the green provider badge before checking whether the plugin was disabled, making disabled provider plugins look active. The badge now reflects the actual enabled state: disabled provider plugins use the same grey disabled badge as other disabled plugins.
 - **Desktop notifications no longer get silently dropped when the agent finishes while you're on another tab (#4416).** Chromium throttles a background tab's SSE, so the stream's `done` event is delivered late — after you return to the tab, when `document.hidden` already reads `false` — and the response-complete notification was suppressed by the live visibility check. The WebUI now tracks whether the tab was hidden at *any* point during a stream and notifies on that basis, so a turn you stepped away from still notifies on completion. A turn you watched the whole time stays silent (matching how Slack/Discord/Gmail/Claude behave), and the user's notifications-enabled setting is still honored.
 
 ## [v0.51.526] — 2026-06-19 — Release SK (the "Running" indicator clears when the server is idle)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@
 
 ### Fixed
 
-- **Disabled provider-category plugins now show the disabled badge in Settings (#4496).** The Plugins panel previously rendered any `provider` / `exclusive` plugin with the green provider badge before checking whether the plugin was disabled, making disabled provider plugins look active. The badge now reflects the actual enabled state: disabled provider plugins use the same grey disabled badge as other disabled plugins.
 - **Desktop notifications no longer get silently dropped when the agent finishes while you're on another tab (#4416).** Chromium throttles a background tab's SSE, so the stream's `done` event is delivered late — after you return to the tab, when `document.hidden` already reads `false` — and the response-complete notification was suppressed by the live visibility check. The WebUI now tracks whether the tab was hidden at *any* point during a stream and notifies on that basis, so a turn you stepped away from still notifies on completion. A turn you watched the whole time stays silent (matching how Slack/Discord/Gmail/Claude behave), and the user's notifications-enabled setting is still honored.
 
 ## [v0.51.526] — 2026-06-19 — Release SK (the "Running" indicator clears when the server is idle)
--- a/api/routes.py
+++ b/api/routes.py
@@ -7109,10 +7109,12 @@ def _plugin_visibility_payload(manager=None) -> dict:
     category's ``<category>.provider`` config, not through ``plugins.enabled``.
     Their ``loaded.enabled`` stays False by design and they register hooks
     outside the four visibility hooks below. The payload surfaces ``kind``
-    and ``activation`` plus ``is_active_provider`` so the panel can render the
-    selected provider distinctly instead of mislabeling it as "Disabled" with no
-    hooks (issue #2659), while still showing unselected exclusive providers as
-    disabled.
+    and ``activation`` plus ``is_active_provider`` when the plugin key carries a
+    category prefix so the panel can render the selected provider distinctly
+    instead of mislabeling it as "Disabled" with no hooks (issue #2659), while
+    still showing unselected exclusive providers as disabled. Flat-key exclusive
+    plugins omit the new field so older activation-based badge semantics remain
+    intact when the category cannot be inferred.
     """
     manager = manager or _get_plugin_manager_for_visibility()
     manager.discover_and_load(force=False)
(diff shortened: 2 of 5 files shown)


nesquena-hermes (changes requested):
Thanks for tackling #4496 — the underlying problem is real (a provider-category plugin that isn't actually the active provider shouldn't wear the green "Active provider" badge). But the current fix regresses the #2659 contract and its targeted "disabled provider" case can't actually occur, so a couple of changes are needed before this is ready.

### 1. Regression: active *exclusive* providers (memory/web) now show the grey "Disabled" badge

The new guard gates the provider badge on `enabled`:

```js
// static/panels.js:7549
const isProviderActive = isProvider && enabled && activation !== 'disabled';
```

But **active exclusive plugins carry `enabled=false` by design** — that's the whole point of #2659. From `api/routes.py` (the `/api/plugins` payload builder):

```py
# api/routes.py:7077-7082 (docstring)
# Exclusive plugins (e.g. memory providers) are activated through their
# category's `<category>.provider` config, not through `plugins.enabled`.
# Their `loaded.enabled` stays False by design ... (issue #2659).

# api/routes.py:7103-7109
enabled_flag = bool(getattr(loaded, "enabled", False))
if kind == "exclusive":
    activation = "exclusive"          # <- no enabled check; enabled_flag stays False
```

So for an **active** memory/web provider the payload is `{activation:"exclusive", enabled:false}`. With the new guard:
- `isProvider` = `(activation==='exclusive')` = **true**
- `enabled` = `(plugin.enabled !== false)` = **false**
- `isProviderActive` = `true && false && …` = **false** → falls through to the `else` → **grey `plugin-card-badge-disabled`**

On master this was `if(isProvider)` → green `plugins_active_provider`. So this PR re-introduces exactly the "active provider mislabeled as Disabled" bug that #2659 fixed. (The existing `test_plugins_panel_renders_active_provider_badge` is a static source-grep, so it stays green and won't catch this — the regression is invisible to CI.)

### 2. The targeted "disabled provider" payload can't come from the backend

`api/routes.py:7106` only emits `activation="provider"` when the plugin is enabled:

```py
elif kind == "model-provider" and enabled_flag:
    activation = "provider"
else:
    activation = "enabled" if enabled_flag else "disabled"
```

So `activation="provider"` always implies `enabled=true`, and a genuinely-disabled model-provider already gets `activation="disabled"` (→ `isProvider` false → grey badge, correct on master). The new test's disabled case:

```py
# tests/test_issue4496_plugin_badge_state.py
_buildPluginCard({ activation: 'provider', enabled: false, ... })
```

feeds a payload the backend never produces, so it passes vacuously while the real regressing case (`{activation:'exclusive', enabled:false}`) is untested.

### 3. Greptile's note is right too: `&& activation !== 'disabled'` is dead code

`isProvider` is only true for `'exclusive'`/`'provider'`, so that term can never be false. The `&& enabled` is the term doing the (incorrect) work.

### Suggested direction

The `enabled` flag can't distinguish an *active* exclusive provider from an *inactive* one — both have `enabled=false`. To correctly fix #4496 you need a backend signal for which exclusive plugin is the selected provider for its category (e.g. add an `is_active_provider`/`active` field to the `/api/plugins` payload from the `<category>.provider` config), then gate the green badge on that field rather than `enabled`. Please also add a test for `{activation:'exclusive', enabled:false}` asserting the **green provider** badge (the #2659 case), and drop the CHANGELOG entry (greptile P1 — that's release-automation-owned).

Happy to re-review once the active-exclusive case is preserved.

franksong2702:
Addressed the review feedback in `f9e97d46b`:

- removed the `CHANGELOG.md` entry
- added explicit `/api/plugins` `is_active_provider` state instead of inferring provider badge state from `enabled`
- preserved the #2659 active-exclusive case (`activation: "exclusive"`, `enabled: false`) as green provider
- added inactive-exclusive coverage so unselected exclusive providers render disabled

Local verification:

- `./scripts/test.sh tests/test_issue4496_plugin_badge_state.py tests/test_plugins_panel.py` (`34 passed`)
- `.venv/bin/python -m py_compile api/routes.py tests/test_issue4496_plugin_badge_state.py tests/test_plugins_panel.py`
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master --strict`
- `git diff --check`

franksong2702:
Follow-up pushed in `284907dd3` for Greptile's flat-key exclusive plugin note.

What changed:
- category-prefixed exclusive plugins still emit `is_active_provider` so selected providers can be distinguished from unselected ones
- flat-key exclusive plugins now omit `is_active_provider`, which preserves the existing activation-based frontend fallback instead of forcing them to the disabled badge
- added API coverage for the omitted-field compatibility path
- added frontend coverage proving an exclusive plugin without `is_active_provider` still renders the provider badge via fallback

Local verification:
- `./scripts/test.sh tests/test_issue4496_plugin_badge_state.py tests/test_plugins_panel.py` (`35 passed`)
- `.venv/bin/python -m py_compile api/routes.py tests/test_issue4496_plugin_badge_state.py tests/test_plugins_panel.py`
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master --strict`
- `git diff --check`

CI is running on the new head; I will read it back after it settles.

franksong2702:
Latest base refresh pushed in `afd71c074` after the `v0.51.530` profiles-route fix landed on `master`.

Current head includes the flat-key exclusive-provider compatibility fix from `284907dd3` plus the latest upstream merge.

Local verification on this final head:
- `./scripts/test.sh tests/test_issue4496_plugin_badge_state.py tests/test_plugins_panel.py tests/test_profiles_route_endpoint.py` (`36 passed`)
- `.venv/bin/python -m py_compile api/routes.py tests/test_issue4496_plugin_badge_state.py tests/test_plugins_panel.py tests/test_profiles_route_endpoint.py`
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master --strict`
- `git diff --check`

GitHub read-back on `afd71c074`: all checks are successful and the branch is mergeable. The PR still shows `CHANGES_REQUESTED` / `blocked` only because the earlier maintainer review state/label has not been cleared yet.

nesquena-hermes:
Shipped in v0.51.531 via the rebased release branch (credited in CHANGELOG + release notes). Thanks @franksong2702 — clean convergence on the fix-spec from the earlier bounce: the new is_active_provider signal lets the panel badge only the selected exclusive provider, flat-key plugins keep the legacy badge (backward-compat), all 4 states unit-tested. Codex SAFE + Opus SAFE + full suite green (9687).

PR #4518, opened 2026-06-20
fix: match mobile question jump button to session jump pill

TomBanksAU:
## Summary
Fix the mobile PWA per-response jump button so it matches the session jump pill sizing instead of rendering as a smaller compact control.

## What changed
- reuse the session jump button class stack for the per-response jump button markup
- align the response jump button dimensions, padding, border, background, and hover treatment with the existing session jump pill styling
- keep the button inline in the message footer, but remove the extra mobile-only compact mode that shrank it below the session jump button size
- update the regression test to assert the shared pill styling contract

## Investigation
- the small button in the screenshot is the per-response jump button (`.msg-question-jump-btn`), not the floating session `Start` pill
- the larger comparison control is the session jump/end button family
- these controls are not strictly mutually exclusive in code: the per-response jump button renders with assistant messages, while the session jump buttons are driven by scroll position and settings

## Testing
- `uv run --with pytest --with pyyaml --with cryptography pytest tests/test_issue2246_question_jump.py tests/test_session_jump_buttons.py -q`
- `node --check static/ui.js`

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -4898,26 +4898,42 @@ main.main > #mainPlugin{display:none;}
 }
 .msg-foot .msg-actions { opacity: 1; margin-left: 0; }
 .msg-foot .msg-time { font-size: 10.5px; opacity: .75; }
+.session-jump-btn--inline {
+  position: static;
+  right: auto;
+  top: auto;
+  bottom: auto;
+  z-index: auto;
+  min-width: 32px;
+  margin-left: auto;
+  flex-shrink: 0;
+}
 .msg-question-jump-btn {
   display: inline-flex;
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in v0.51.529 via the rebased release branch (credited in CHANGELOG + release notes). Thanks @TomBanksAU! The per-response jump button now matches the session jump pill — verified before/after at desktop (1440px) and mobile (390px): height 22px->32px, font 11px/400->12px/600, solid background + subtle shadow, and on mobile it collapses to a clean 32px circular icon. Codex SAFE + Opus SAFE + full suite green (9683).

Issue #2639, opened 2026-05-20
Clarify popup can accept input but agent still receives no-response fallback


stefanpieter:
## Summary

A WebUI `Clarification needed` popup/card can appear to accept the user's answer, but the agent later continues as if no explicit answer arrived, using the timeout/best-judgement fallback instead.

This is distinct from the older clarify-dialog feature request (#496), which added the UI. The dialog exists, but response delivery/acknowledgement appears unreliable.

## Observed behaviour

A user reports repeatedly seeing text along the lines of:

> No explicit response arrived, so I’ll use best judgement and proceed because ...

or the Hermes clarify fallback:

> The user did not provide a response within the time limit. Use your best judgement to make the choice and proceed.

This happens even though the user entered/provided the requested input in the WebUI popup.

## Expected behaviour

When the user submits a clarify response before expiry:

1. The answer should be delivered to the blocked clarify call.
2. The agent should continue using that exact answer.
3. If the prompt has expired, is stale, belongs to another session, or was already cleared, the WebUI should show a visible failure and keep/preserve the user's draft rather than silently hiding the card.

## Suspected cause / code-path notes

From inspecting the current WebUI implementation:

- `static/messages.js::respondClarify()` posts to `/api/clarify/respond`, but hides the clarify card immediately before confirming that the response was actually accepted.
- The frontend POST currently sends only `{session_id, response}`. There does not appear to be a stable `clarify_id` / `request_id` carried from the pending prompt into the response.
- `api/routes.py::_handle_clarify_respond()` accepts an optional `clarify_id`, but the legacy path `_resolve_clarify_legacy()` treats a no-id response as successful even when no pending prompt was resolved:
  - `return bool(resolved) or not bool(clarify_id)`
- `api/streaming.py::_clarify_callback_impl()` returns the timeout/best-judgement fallback if the waiting clarify entry never receives a result.

That combination can make a stale, late, wrong-session, or already-cleared popup answer look accepted from the browser's perspective while the blocked agent call still receives the timeout fallback.

## Environment seen locally

- Hermes WebUI: `v0.51.95`
- Commit: `9c983e69`
- Browser/UI: standalone Hermes WebUI clarify card flow

## Suggested fix

- Add a stable clarify/request id to each pending clarify payload delivered to the frontend.
- Require and match that id in `/api/clarify/respond`.
- Return a clear non-success result, ideally HTTP 409 plus `ok:false`, for stale/expired/wrong-session/no-pending responses.
[...]

PR #4524, opened 2026-06-20
fix(clarify): SSE-notify on clear_pending + treat 409 as terminal so expired prompts stop bricking the session (#4504)

### LINKED ISSUE #4504: bug(clarify): expired clarify prompt leaves card stuck + composer locked with no way to clear (409 on submit)
## Summary

When a clarification prompt **expires** (the agent already timed out and proceeded with its best-judgement fallback) and the user then submits an answer into the still-visible clarify card, the response is rejected `409` but **the card and the locked composer are never cleared**. The clarify card stays docked over the chat, the composer stays disabled, and there is no affordance to dismiss it. Switching sessions and returning re-shows the stuck card, so the session is effectively locked — you can't edit or submit text and can't clear the box.

Reported by **@b3n.w** in Discord `#report-bugs` (2026-06-19):

> not sure how to reproduce this, but it has happened a few times now, when a clarification needed / question box pops up and expires 0s, the box gets stuck, the input area also gets locked, and you can't clear it. Even if you switch sessions, when you return you're locked in this question box which can't be submitted, and text you can't edit or submit.

Toast he saw on submit: **"Clarification prompt expired or not found. The agent may have already proceeded."**

| Stuck card + locked composer | The expiry toast |
|---|---|
| stuck | toast |

Likely surfaced by the clarify flow that shipped recently (**PR #520**, "stabilize clarify dialog flow") — matches his "after an update" framing.

## Steps to reproduce

1. Be in a session where the agent calls `clarify` and the prompt is displayed.
2. Let the countdown reach **0s** (≥ `clarify.timeout`, default 120s). The agent's `_clarify_callback_impl` hits its `remaining <= 0` branch, calls `clear_pending(sid)`, and returns the "did not provide a response within the time limit" fallback — the server-side clarify entry is now gone and the agent has moved on.
3. The clarify **card is still visible** in the browser (nothing told it to hide) and the composer is still locked.
4. Type an answer and click **Send** (or hit Enter).
5. `POST /api/clarify/respond` → `409 {"ok":false,"stale":true}`. The catch block re-enables the card controls and **keeps the card + draft visible** — but retry can never succeed, so you're stuck. Switching away and back re-renders the stuck card from the client-side `_clarifyPendingBySession` map.

## Root cause

Two compounding gaps. The core defect is that **expiry only clears server state — it never tells the browser to take the card down or unlock the composer.**

### 1. Expiry is server-side-silent — no event to the client

`api/streaming.py` `_clarify_callback_impl` (~L6122–6164): on timeout it calls `_clear_clarify_pending(sid)` and returns the fallback string. That clears the server `_pending` / `_gateway_queues` entries, but **emits no SSE event** to the browser. (Contrast `resolve_clarify*`, which call `_clarify_sse_notify(session_key, None, 0)` to push a "cleared" event — the timeout path does not.)

`api/clarify.py` has **no TTL reaper at all**: `_with_timeout_metadata` computes `expires_at` purely for the frontend countdown display; nothing in the module ever drops an entry on expiry. The only thing that clears it is the agent's own callback loop, which exits silently.

So the browser's clarify card stays up because:
- the SSE `clarify` stream got no "cleared" push, and
- the 3000ms `_startClarifyFallbackPoll` tick *would* eventually clear it (it calls `_hideClarifyCardIfOwner(sid, false, 'expired')` when `/api/clarify/pending` returns null) — **but that poll is only running while the stream is live.** Once the turn ends, `stopClarifyPolling()` fires on the terminal/done/cancel handlers (`_clearClarifyForOwner` at messages.js ~L1762/3368/3701/3987/4057, plus the explicit `stopClarifyPolling()` in the start error paths). After that nothing reconciles the card against the now-empty server state.

### 2. The client `409` handler keeps an unrecoverable card up

`static/messages.js` `respondClarify` catch block (~L5163–5175): on a `409`/stale error it deliberately **re-enables the controls and keeps the card + draft visible** "so the user can retry" — a design originally meant for the *wrong-session / next-prompt-already-loaded* case (#2639). But when the prompt is genuinely **expired**, retry is impossible: the server entry is gone, so every resubmit returns `409` forever. The same applies to the `else` branch on a `{ok:false}` body (~L5152–5161). The `409` response already carries `"stale": true`, but the client ignores that flag and treats expired identically to a transient retryable error.

Because the composer is locked via `lockComposerForClarify` (ui.js ~L5196) for as long as the card is "active", the locked-input symptom is just a consequence of the card never coming down.

### 3. No manual escape hatch

The clarify card (`_ensureClarifyCardDom`, messages.js ~L4759) has a **collapse** button but **no dismiss/close** button. There is a `_stashClarifyDraft(reason)` helper that, on `reason === 'expired'`, moves the draft into the composer and unlocks it — but it's only reachable through `hideClarifyCard(..., 'expired')`, which nothing calls on the silent-timeout path. So the user has zero way out short of a full page reload.

## Suggested fix (phased; investigate-and-file only — not built here)

**Phase A — server pushes a "cleared/expired" event on timeout (the real fix).**
In `_clarify_callback_impl`'s `remaining <= 0` (and `cancel_evt.is_set()`) branches, after `_clear_clarify_pending(sid)`, emit a clarify SSE notify with `None` head so the browser hides the card the same way `resolve_clarify` does. Cleanest: add an `expire_pending(session_key)` / make `clear_pending` call `_clarify_sse_notify(session_key, None, 0)` so any clear path notifies subscribers. Consider an `expired: true` marker in the payload so the client can route to the draft-stash path rather than a generic hide.

**Phase B — client treats `stale`/`409` as terminal, not retryable.**
In `respondClarify`, when the error is `409` or the body has `stale === true`, call `hideClarifyCard(true, 'expired')` (which stashes the draft into the now-unlocked composer and shows "Clarification timed out. Your draft was kept in the composer.") instead of re-enabling the card for an impossible retry. Keep the existing keep-card behavior only for the genuine wrong-session/next-prompt case (distinguish via `clarify_id` mismatch, not a blanket 409).

**Phase C — defense in depth.**
- Add a dismiss/close ("×") affordance to the clarify card so a stuck card is always user-recoverable, routed through `hideClarifyCard(true, 'expired')` to preserve the draft.
- When the countdown reaches 0 client-side (`_updateClarifyCountdown`, messages.js ~L4902), proactively reconcile against `/api/clarify/pending` (or hide on a grace delay) even when the stream poll isn't running, so an idle/returned-to session self-heals.
- Optional: a real server-side TTL reaper in `api/clarify.py` so an entry can't outlive its `expires_at` even if the agent callback loop is wedged.

## Scope

Frontend `static/messages.js` (`respondClarify` 409 handling + optional dismiss button + countdown-zero reconcile) and backend `api/streaming.py` + `api/clarify.py` (notify on timeout/clear). Single repo (`nesquena/hermes-webui`); no desktop-wrapper companion needed. Phase A alone resolves the reported lock-up; B and C harden it.

## Reported by

@b3n.w (b3nw) — Discord `#report-bugs`, 2026-06-19. Filed by maintainer agent after investigation.

Sanjays2402:
Closes #4504

## What Problem This Solves

When the clarify prompt's countdown hits zero, the agent-side `_clarify_callback_impl` calls `clear_pending(sid)` and returns its fallback string. Until now `clear_pending` cleared server state but emitted no SSE notify, and the 3000 ms fallback poller had already been stopped by the turn-end handler — so the browser never knew the prompt was gone. The card stayed docked, the composer stayed locked, and the user's only "escape" was to submit a response, which `_handle_clarify_respond` then rejected with `409 {stale: true}`. The previous client catch-block treated 409 as retryable, leaving the card + draft visible with controls re-enabled, but every retry returned 409 forever. The user was effectively bricked in the UI — switching sessions and returning re-rendered the same stuck card from the client cache.

Per @b3n.w in Discord `#report-bugs` (2026-06-19): "the box gets stuck, the input area also gets locked, and you can't clear it. Even if you switch sessions, when you return you're locked in this question box which can't be submitted, and text you can't edit or submit."

## Why This Change Was Made

The maintainer-authored issue body laid out a 3-phase fix. This PR ships Phase A and Phase B (the user-facing escape hatches); Phase C (UI dismiss button + countdown-zero proactive reconcile) is left as a follow-up since the first two already remove the dead-end.

**Phase A — server pushes a "cleared" event on `clear_pending`.**
`api/clarify.py::clear_pending` now calls `_clarify_sse_notify(session_key, None, 0)` from inside `_lock` whenever it actually clears something. That matches the ordering contract `submit_pending` and `resolve_clarify*` already follow (notify under `_lock`, then `publish_session_list_changed` and `event.set()` outside). The browser's existing `_handleClarifyEvent` branch on `pending=null` already routes through `_hideClarifyCardIfOwner(sid, false, 'expired')` → `_stashClarifyDraft('expired')`, so the card comes down and any partial draft moves into the now-unlocked composer with a "Clarification timed out. Your draft was kept in the composer." notice.

**Phase B — client treats 409 as terminal in `respondClarify`'s catch.**
`static/messages.js::respondClarify` now branches on `e.status === 409` to call `hideClarifyCard(true, 'expired')` and early-return, instead of re-enabling the controls for an impossible retry. The "next prompt already loaded" case from #2639 isn't lost: when a fresh clarify event arrives, the SSE/poll path's `showClarifyCard()` re-renders the card from scratch with the new `clarify_id`. The non-409 catch branch (true network / transient errors) keeps the existing keep-card-and-draft behavior so genuine retries still work once connectivity returns.

The no-entries case in `clear_pending` is a deliberate no-op (no spurious notify if nothing was actually cleared), and the `.event.set()` agent-side unblock is preserved so `_clarify_callback_impl`'s timeout branch still returns its fallback string on time.

Non-goals (left as Phase C / follow-ups):
- Dismiss/close ("×") affordance on the clarify card.
- Client-side proactive reconcile against `/api/clarify/pending` when the countdown reaches zero.
- Distinguishing "expired" vs "wrong-session" 409 via a server-returned `current_clarify_id` field — current behavior dismisses on any 409, and the next SSE event recovers the wrong-session case naturally.

## User Impact

Expired clarify prompts no longer brick the session. The card comes down via the SSE notify the moment the server clears the pending entry, the composer unlocks, and any draft the user typed is preserved in the composer. If the SSE message races a user click (the user hits Send during the clear window), the 409 response now dismisses the card the same way instead of leaving them stuck.

## Evidence

**New regression suite (`tests/test_4504_clarify_stuck_on_expiry.py`) — 7 cases:**

```
$ pytest tests/test_4504_clarify_stuck_on_expiry.py -v
collected 7 items
  TestClearPendingNotifiesSSE
    ✓ test_clear_pending_pushes_none_head_to_subscriber
    ✓ test_clear_pending_no_op_does_not_notify
    ✓ test_clear_pending_unblocks_caller_event
  TestClarifyClearPendingSourceMarkers
    ✓ test_clear_pending_calls_notify
  TestRespondClarify409Terminal
    ✓ test_409_routes_to_hide_clarify_card_expired
    ✓ test_409_does_not_re_enable_controls
    ✓ test_non_409_still_keeps_card_visible

============================== 7 passed in 2.68s ===============================
```

**Broader clarify suite (50 cases) still passes:**

```
$ pytest tests/test_clarify_sse.py tests/test_clarify_unblock.py \
         tests/test_1466_sidebar_cancel_clarify.py tests/test_issue2883_clarify_padding.py
collected 50 items
tests/test_clarify_sse.py ............................   [ 56%]
tests/test_clarify_unblock.py .............              [ 82%]
tests/test_1466_sidebar_cancel_clarify.py .....          [ 92%]
tests/test_issue2883_clarify_padding.py ....             [100%]
============================== 50 passed in 1.22s ===============================
```

**Lint:**

```
$ ruff check api/clarify.py tests/test_4504_clarify_stuck_on_expiry.py
All checks passed!

$ node --check static/messages.js
(parse-clean)
```

**Diff:** 3 files, +209 / −5. Two surgical commits — server-side notify, then client-side 409 handling — split so reviewers can flip Phase A vs Phase B independently if scope-cutting is preferred.

### DIFF
--- a/api/clarify.py
+++ b/api/clarify.py
@@ -61,9 +61,21 @@ def unregister_gateway_notify(session_key: str) -> None:
 
 
 def clear_pending(session_key: str) -> int:
-    """Clear any pending clarify prompts for the session without removing the callback."""
+    """Clear any pending clarify prompts for the session without removing the callback.
+
+    Emits an SSE notify (``pending=None, total=0``) so any browser subscribed to
+    the clarify stream takes down its visible card and unlocks the composer.
+    Without this notify, the silent-timeout path in
+    ``streaming.py::_clarify_callback_impl`` cleared server state but left the
+    browser with a stuck card and a locked composer that 409'd on any submit
+    (#4504).
+    """
     with _lock:
         entries = _clear_queue_locked(session_key)
+        # Notify from inside _lock for ordering consistency with submit_pending
+        # and resolve_clarify*, which also publish under the same lock.
+        if entries:
+            _clarify_sse_notify(session_key, None, 0)
     if entries:
         publish_session_list_changed("attention_cleared")
     for entry in entries:
--- a/static/messages.js
+++ b/static/messages.js
@@ -5215,24 +5215,43 @@ async function respondClarify(response) {
   } catch(e) {
     // The server returns 409 with ``stale: true`` for both genuinely-expired
     // prompts and wrong-session/next-prompt-loaded races. In both cases the
-    // server-side ``_pending`` entry is gone, so retrying with the same
-    // ``clarify_id`` can never succeed — the prior keep-card-and-draft
+    // server-side ``_pending`` entry for *this* clarify_id is gone, so
+    // retrying it can never succeed — the prior keep-card-and-draft
     // behavior left the user with a permanently 409-ing card and a locked
     // composer, with no affordance to dismiss it (#4504). Treat 409 as
-    // terminal here: hand the draft to the unlocked composer via
-    // ``hideClarifyCard(true, 'expired')`` (which routes through
-    // ``_stashClarifyDraft('expired')``). If a *new* clarify event is queued
-    // for this session, the SSE/poll path will re-render the card from
-    // scratch via ``showClarifyCard`` with the new ``clarify_id`` — so we do
-    // not lose the next-prompt case from #2639, we just stop pretending the
-    // *current* card is recoverable.
+    // terminal here, but only when the visible card still matches what we
+    // just submitted: mirroring the success path's ``_clarifyId === clarifyId``
+    // guard (codex review P1, #2639) — if a parallel poll already rendered
+    // the *next* queued prompt B while A's response was in flight, we must
+    // not tear B down on A's late 409. The SSE/poll path will re-render the
+    // next prompt's card from scratch via ``showClarifyCard`` either way.
     if (e && e.status === 409) {
-      hideClarifyCard(true, "expired");
-      const errMsg = (e.message || "Clarification prompt expired or not found.");
-      if (typeof setStatus === "function") setStatus("Clarify: " + errMsg);
-      // Suppress the duplicate toast here: ``_stashClarifyDraft('expired')``
-      // already surfaces "Clarification timed out. Your draft was kept in
-      // the composer." via showToast, which is the actionable message.
+      if (_clarifyId === clarifyId) {
+        // Same card still showing — dismiss it and rescue the typed draft.
+        // Order matters: ``_stashClarifyDraft`` (called from
+        // ``hideClarifyCard``) bails when ``#clarifySubmit`` still carries
+        // the ``loading`` class set above. Clear loading first, otherwise
+        // the typed answer is silently dropped (reviewer P1).
+        _clarifySetControlsDisabled(false, false);
+        _clarifySessionId = null;
+        _clarifyId = null;
+        _clearClarifyPendingForSession(sid);
+        hideClarifyCard(true, "expired");
+        const errMsg = (e.message || "Clarification prompt expired or not found.");
+        if (typeof setStatus === "function") setStatus("Clarify: " + errMsg);
+        // ``_stashClarifyDraft('expired')`` already surfaces the actionable
+        // "Clarification timed out. Your draft was kept in the composer."
+        // toast when there is a draft to rescue, so we don't double-toast.
+        return;
+      }
+      // A newer prompt is showing (race between user click and SSE/poll).
+      // Don't dismiss it on this late 409 — just re-enable controls and
+      // surface the error. The user's draft for the now-stale prompt is
+      // dropped intentionally; the next prompt has its own input cycle.
+      _clarifySetControlsDisabled(false, false);
+      if (typeof setStatus === "function") {
+        setStatus("Clarify: previous prompt expired — a newer one is showing.");
+      }
       return;
     }
     // Network / other transient errors — keep the card and draft visible so
--- a/static/messages.js
+++ b/static/messages.js
@@ -5213,15 +5213,36 @@ async function respondClarify(response) {
       if (typeof setStatus === "function") setStatus(errMsg);
     }
   } catch(e) {
-    // Stale (409) or network error — keep the card and draft visible so the user can retry.
+    // The server returns 409 with ``stale: true`` for both genuinely-expired
+    // prompts and wrong-session/next-prompt-loaded races. In both cases the
+    // server-side ``_pending`` entry is gone, so retrying with the same
+    // ``clarify_id`` can never succeed — the prior keep-card-and-draft
+    // behavior left the user with a permanently 409-ing card and a locked
+    // composer, with no affordance to dismiss it (#4504). Treat 409 as
+    // terminal here: hand the draft to the unlocked composer via
+    // ``hideClarifyCard(true, 'expired')`` (which routes through
+    // ``_stashClarifyDraft('expired')``). If a *new* clarify event is queued
+    // for this session, the SSE/poll path will re-render the card from
+    // scratch via ``showClarifyCard`` with the new ``clarify_id`` — so we do
+    // not lose the next-prompt case from #2639, we just stop pretending the
+    // *current* card is recoverable.
+    if (e && e.status === 409) {
+      hideClarifyCard(true, "expired");
+      const errMsg = (e.message || "Clarification prompt expired or not found.");
+      if (typeof setStatus === "function") setStatus("Clarify: " + errMsg);
+      // Suppress the duplicate toast here: ``_stashClarifyDraft('expired')``
+      // already surfaces "Clarification timed out. Your draft was kept in
+      // the composer." via showToast, which is the actionable message.
+      return;
+    }
+    // Network / other transient errors — keep the card and draft visible so
+    // the user can retry once connectivity returns.
     _clarifySetControlsDisabled(false, false);
     if (input) {
       input.value = draft;
       input.focus();
     }
-    const errMsg = (e && e.status === 409)
-      ? (e.message || "Clarification prompt expired or not found.")
-      : ((e && e.message) || "Failed to deliver clarification response.");
+    const errMsg = (e && e.message) || "Failed to deliver clarification response.";
     if (typeof setStatus === "function") setStatus("Clarify: " + errMsg);
     if (typeof showToast === "function") showToast(errMsg, 5000);
   }
--- a/tests/test_4504_clarify_stuck_on_expiry.py
+++ b/tests/test_4504_clarify_stuck_on_expiry.py
@@ -1,0 +1,171 @@
+"""Regression coverage for #4504 — clarify card / composer stuck after expiry.
+
+Two compounding gaps the user experienced:
+
+  1. Server-side ``clear_pending(sid)`` ran on silent timeout but emitted no
+     SSE notify, so the browser never knew the prompt was gone — the visible
+     card stayed up and the composer stayed locked.
+
+  2. The client's ``respondClarify`` catch-block treated every 409 (including
+     ``stale: true``) as retryable, leaving the card + draft visible and the
+     controls re-enabled — but every retry returned 409, so the session was
+     permanently stuck. The user had zero affordance to dismiss the card.
+
+This file pins:
+
+  - ``clear_pending`` notifies SSE subscribers (head=None, total=0) so the
+    silent-timeout path takes the card down via the existing pending-=-null
+    branch in ``_handleClarifyEvent``.
+  - The client's ``respondClarify`` catch-block, on ``e.status === 409``,
+    routes to ``hideClarifyCard(true, 'expired')`` so the draft is moved into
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
Thanks @Sanjays2402 — this is a real, well-diagnosed bug (#4504) and your fix is on the right track: it does un-brick the session. I ran it through a full Codex + Opus review and there are **three things to address** before it can ship, all on the client 409-terminal path (which, per the analysis below, is the *only* path that actually un-bricks the UI — so getting it exactly right matters).

### 1. The typed answer is silently dropped (must-fix)
`respondClarify` sets `_clarifySetControlsDisabled(true, true)` (adds the `loading` class to `#clarifySubmit`) **before** the `try`. The new 409 branch then calls `hideClarifyCard(true, "expired")` → `_stashClarifyDraft("expired")`, but `_stashClarifyDraft` bails immediately on `submit.classList.contains("loading")` (messages.js:~4972). So for a free-text / "Other" answer the user typed, the draft is **not** moved to the composer and **no toast fires** (the branch suppresses its own toast trusting `_stashClarifyDraft` to show it). Net: the card vanishes, the typed answer is lost, no feedback — contradicting the PR's own comment.
**Fix:** clear the loading/disabled state (`_clarifySetControlsDisabled(false, false)`) *before* `hideClarifyCard(true, "expired")` in the 409 branch, so the draft actually reaches the unlocked composer and the "draft kept" toast fires.

### 2. A late 409 can wipe the *next* prompt's card — #2639 regression (must-fix)
The 200/success path guards with `if (_clarifyId === clarifyId)` (messages.js:5188) specifically so a parallel poll that already rendered the next queued prompt B isn't clobbered (that was the #2639 fix). The new 409 branch has **no such guard** — it unconditionally tears down the visible card. If prompt B rendered while A's response was in flight, A's late 409 dismisses B. 
**Fix:** gate the 409-terminal handling on `_clarifyId === clarifyId`. If it differs, a newer prompt is showing — leave it, just re-enable controls and return. Recommend also clearing the stale cached prompt for the session (`_clearClarifyPendingForSession(sid)`) in the same-id case so it can't re-render.

### 3. The server-side SSE notify has no browser consumer (not blocking — keep it, but adjust the framing)
`clear_pending`'s new `_clarify_sse_notify(...)` is ordering-correct and well-tested, **but no client opens `/api/clarify/stream`** — the WebUI switched clarify to HTTP polling (`/api/clarify/pending`) in v0.51.340 to avoid connection-pool exhaustion (`_clarifyEventSource` is declared but never `new`'d; both `clarify/stream` mentions in static/*.js are comments). So the notify doesn't take down any card in the live UI — the un-bricking comes **entirely** from the client 409-terminal path. Keeping the notify is fine (it's correct, and it already drives the sessions-list attention badge via the pre-existing `publish_session_list_changed`), but the PR description's "subscribed browsers take down the card" claim isn't accurate for the current architecture. No code change strictly required here; just don't rely on it for the fix.

### Tests
The new `test_4504_clarify_stuck_on_expiry.py` asserts the 409 branch *strings* are present but checks neither the loading-guard interaction nor the id-guard — both defects above pass CI uncaught. Please add coverage that (a) a typed draft survives a 409 expiry into the composer, and (b) a 409 for prompt A does not dismiss a rendered prompt B.

Once 1 + 2 (+ their tests) land, this is a strong fix and I'll fast-track the re-gate. Really appreciate the clear root-cause writeup in the PR body.

Sanjays2402:
Thanks for the careful review @nesquena-hermes — both P1s landed and are now pinned by tests. Pushed in [e6910e5](https://github.com/nesquena/hermes-webui/pull/4524/commits/e6910e50).

### 1. Loading-guard fixed — typed draft now reaches the composer

The 409 same-id branch now calls `_clarifySetControlsDisabled(false, false)` **before** `hideClarifyCard(true, "expired")`, so `_stashClarifyDraft`'s `submit.classList.contains("loading")` check no longer bails and the typed answer + "Clarification timed out. Your draft was kept in the composer." toast both fire. New test `test_409_clears_loading_before_hide_so_draft_is_rescued` pins the call order with a positional assertion (`clear_idx < hide_idx`) so this can't regress.

### 2. `_clarifyId === clarifyId` guard added — late 409 won't wipe prompt B

Mirroring the success path's contract, the 409 catch now branches:

- **Same-id arm** (`_clarifyId === clarifyId`): clear loading → null out session/id → `_clearClarifyPendingForSession(sid)` → `hideClarifyCard(true, "expired")` → early return. Pinned by `test_409_same_id_branch_clears_session_cache`.
- **Different-id arm**: re-enable controls, set status `"Clarify: previous prompt expired — a newer one is showing."`, return. No `hideClarifyCard`, no cache clear. Pinned by `test_409_different_id_branch_does_not_dismiss` (structural: exactly one `hideClarifyCard(true, "expired")` in the branch, both arms call `_clarifySetControlsDisabled(false, false)`).

Test count: 7 → 11. All 11 pass + 50/50 broader clarify suite still passes + ruff clean + `node --check static/messages.js` clean.

### 3. PR-body framing correction (no code change)

You're right that the WebUI clarify transport switched to HTTP polling in v0.51.340 — `new EventSource` is never called for `/api/clarify/stream`, and the two remaining string mentions in `static/*.js` are inside comments. The Phase A `_clarify_sse_notify(...)` in `clear_pending` isn't what un-bricks the live UI today; the un-bricking comes **entirely from the client 409-terminal path**. Keeping the notify is still useful — it's ordering-correct, it drives the sessions-list attention badge via the pre-existing `publish_session_list_changed` call, and it's a no-cost reattachment hook if SSE ever comes back — but the PR title/body framing should not promise live-UI takedown via SSE. I've updated the test file's docstring to call this out so future readers don't repeat the mistake; happy to also amend the PR description once you've had a chance to look at the new commit.

Appreciate the time on this one — the reviewer P1s were both real and I'd missed them in the initial pass.

nesquena-hermes:
Shipped in v0.51.534 via the rebased release branch (credited in CHANGELOG + release notes). Thanks @Sanjays2402 — clean convergence on all 3 fix-spec points (loading cleared before stash so the typed draft survives; clarify_id-guarded dismissal preserving #2639; same-id stale clear), with the reasoning in the comments and a 261-line test covering each case. Codex SAFE + Opus SAFE + full suite green (9707). This un-bricks the session on clarify-prompt expiry — nice fix.

PR #4525, opened 2026-06-20
Release v0.51.530 — fix /api/profiles 500 (UnboundLocalError regression)

nesquena-hermes:
## Release v0.51.530 — Release SO

Ships #4522 (fixes a live regression introduced in v0.51.528 / #4454).

### Fixed
- **`GET /api/profiles` no longer returns a 500 (UnboundLocalError).** A later branch in the request handler (`/api/profile/active`) imports `get_active_profile_name` as a function-local, which — by Python scoping — makes the name local across the *entire* `handle_get`. The `/api/profiles` branch referenced it before that import ran, so every call to the profiles list endpoint raised `UnboundLocalError`. The branch now imports the name it uses directly. Confirmed empirically (prod returned HTTP 500 with `cannot access local variable 'get_active_profile_name'` in the log). Restores the profiles list/dropdown. Thanks @TomBanksAU.

### Context
Introduced in v0.51.528 (#4454) when a redundant local import was removed from the `/api/profiles` branch, exposing the function-scope trap from the sibling `/api/profile/active` branch. The full suite stayed green because no test exercised `GET /api/profiles` directly — #4522 adds that regression test.

### Gate
- Codex: SAFE TO SHIP (audited every `api.profiles` function-local import in `handle_get`; `/api/profiles` was the only use-before-local-import path; all others import-then-use within their own branch).
- Opus: SAFE — ship (same audit independently; `_is_isolated_profile_mode` is module-only, resolves cleanly).
- Full pytest suite: 9684 passed, 0 failed (includes the new `test_profiles_route_endpoint.py`).

Original PR: #4522 by @TomBanksAU.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 
 ## [Unreleased]
 
+## [v0.51.530] — 2026-06-20 — Release SO (fix /api/profiles 500)
+
+### Fixed
+
+- **`GET /api/profiles` no longer returns a 500 (UnboundLocalError).** A later branch in the request handler imports `get_active_profile_name` as a function-local, which made the name local across the whole handler; the `/api/profiles` branch referenced it before that import ran, so every call to the profiles list endpoint raised `UnboundLocalError`. The branch now imports the name it uses directly. Regression introduced in v0.51.528 and surfaced once the redundant local import was removed there; this restores the profiles list/dropdown. Thanks @TomBanksAU.
+
 ## [v0.51.529] — 2026-06-20 — Release SN (per-response jump button matches the session jump pill)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


MinhoJJang:
@nesquena-hermes 
Glad this shipped and that prod `/api/profiles` is back to 200.

One small attribution note: I think the release note meant to credit @MinhoJJang for #4522, but it currently says @TomBanksAU. Could you update that when convenient?

nesquena:
> @nesquena-hermes 
> 
> Glad this shipped and that prod `/api/profiles` is back to 200.
> 
> 
> 
> One small attribution note: I think the release note meant to credit @MinhoJJang for #4522, but it currently says @TomBanksAU. Could you update that when convenient?

Thanks for fixing this, yes I will fix the attribution! Sorry about that!

PR #4552, opened 2026-06-20
fix(#4550): forward gateway reasoning effort

### LINKED ISSUE #4550: Model and reasoning effort controls do not affect remote gateway runs
## Summary

In Hermes WebUI, the visible model and reasoning effort controls appear editable, but when the WebUI is connected to a remote gateway backend, changing those controls does not change the model/reasoning effort used by the remote gateway run.

This makes the UI misleading: it looks like the user has selected a model or reasoning effort for the next message, but the remote gateway continues using its own configured/default values.

## Current behavior

1. Use Hermes WebUI connected to a remote Hermes gateway/backend.
2. Change the model selector and/or reasoning effort in the WebUI.
3. Send a message.
4. The remote gateway run does not reflect the selected model/reasoning effort.
5. The UI still displays the changed selection, giving the impression that it is active.

## Expected behavior

When Hermes WebUI is using a remote gateway backend, model and reasoning effort selection should either:

- be transmitted to the remote gateway/run request and actually affect execution, or
- be disabled/hidden/annotated as gateway-controlled so the user knows the remote gateway configuration is authoritative.

Ideally, the WebUI should show the effective model and reasoning effort returned/used by the gateway, not just local UI state.

## Environment / deployment shape

Remote gateway-backed WebUI deployment.

Known relevant settings from related reports:

```env
HERMES_WEBUI_CHAT_BACKEND=gateway
HERMES_WEBUI_GATEWAY_BASE_URL=http://<gateway-host>:8642
HERMES_GATEWAY_HEALTH_URL=http://<gateway-host>:8642
```

## Why this matters

Model/reasoning controls are prominent in the WebUI and users reasonably expect them to control the next run. In gateway mode, this can lead to incorrect assumptions about cost, latency, model capability, and debugging results because the UI selection is not the effective remote execution configuration.

rodboev:
## Thinking Path

- The issue started as “model and reasoning effort do not affect gateway runs”, but the collaborator follow-up at https://github.com/nesquena/hermes-webui/issues/4550#issuecomment-4757962350 narrowed the live WebUI defect on current `origin/master`, model forwarding already works and reasoning effort forwarding does not.
- The local non-gateway path already has the right source of truth at `api/streaming.py:6758-6771`, where `agent.reasoning_effort` is coerced per model and provider before the actual request is built.
- The gateway path diverges at `api/gateway_chat.py:572-575` and `api/gateway_chat.py:629-635`, where both remote request branches forward `provider` but never add top-level `reasoning_effort`, so the fix is to reuse the existing coercion source and send the value in the same request shape the sibling gateway transport already expects.

## What Changed

- `api/gateway_chat.py`: mirror the existing local reasoning-effort coercion path, then forward top-level `reasoning_effort` through both gateway request branches.
- `tests/test_gateway_approval_runs_api.py`: prove the runs worker forwards `reasoning_effort` into the `body_extras` passed to `_run_gateway_runs_api_streaming(...)`.
- `tests/test_webui_gateway_chat_backend.py`: prove the legacy `/v1/chat/completions` payload now carries top-level `reasoning_effort`.

## Why It Matters

Gateway-backed WebUI chats should honor the same reasoning effort selection that local chats already honor. This keeps the browser control effective in remote mode, without changing the existing provider forwarding or widening into gateway-side policy work.

## Verification

`pytest tests/test_gateway_approval_runs_api.py tests/test_webui_gateway_chat_backend.py -v --timeout=60`

## Risks / Follow-ups

- This only fixes the live WebUI gap, it does not guarantee every deployed gateway version will honor the forwarded field once received.
- If maintainers want explicit disable-path proof later, a focused `"none"` regression assertion can be added without widening the production scope.

## Upstream

Closes #4550.

Scope narrowed by the collaborator comment at https://github.com/nesquena/hermes-webui/issues/4550#issuecomment-4757962350, which confirmed model forwarding already exists and isolated the remaining WebUI bug to reasoning effort forwarding.

## Model Used

GPT-5 via Codex CLI

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -108,6 +108,7 @@ def _gateway_reasoning_effort_for_request(cfg, *, model=None, model_provider=Non
             model,
             provider_id=model_provider,
         )
+        # Preserve explicit "none" while still omitting absent or invalid effort.
         return None if not coerced else str(coerced)
     except Exception:
         return None
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -19,6 +19,7 @@ from api.config import (
     STREAM_PARTIAL_TEXT,
     STREAM_REASONING_TEXT,
     _get_session_agent_lock,
+    coerce_reasoning_effort_for_model,
     gateway_supports_approval,
     register_active_run,
     unregister_active_run,
@@ -94,6 +95,22 @@ def _gateway_use_runs_api_enabled(config_data=None, environ: dict[str, str] | No
         or ""
     ).strip().lower()
     return raw in ("1", "true", "yes", "on")
+
+
+def _gateway_reasoning_effort_for_request(cfg, *, model=None, model_provider=None):
+    """Read and coerce user-configured reasoning effort for a gateway request."""
+    try:
+        cfg_data = cfg if isinstance(cfg, dict) else {}
+        effort_cfg = cfg_data.get("agent", {}) if isinstance(cfg_data, dict) else {}
+        effort_raw = effort_cfg.get("reasoning_effort") if isinstance(effort_cfg, dict) else None
+        coerced = coerce_reasoning_effort_for_model(
+            effort_raw,
+            model,
+            provider_id=model_provider,
+        )
+        return None if not coerced else str(coerced)
+    except Exception:
+        return None
 
 
 def gateway_chat_config_status(config_data=None, environ: dict[str, str] | None = None) -> dict:
@@ -527,6 +544,11 @@ def _run_gateway_chat_streaming(
         from api.config import get_config  # imported lazily to avoid config-cycle churn
 
         cfg = get_config()
+        reasoning_effort = _gateway_reasoning_effort_for_request(
+            cfg,
+            model=model,
+            model_provider=model_provider,
+        )
         try:
             from api.streaming import (
                 _load_webui_prefill_context,
@@ -573,6 +595,8 @@ def _run_gateway_chat_streaming(
             body_extras = {}
             if model_provider:
                 body_extras["provider"] = model_provider
+            if reasoning_effort is not None:
+                body_extras["reasoning_effort"] = reasoning_effort
             try:
                 final_text, usage = _run_gateway_runs_api_streaming(
                     session_id, msg_text, model, workspace, stream_id,
@@ -633,6 +657,8 @@ def _run_gateway_chat_streaming(
             }
             if model_provider:
                 body["provider"] = model_provider
+            if reasoning_effort is not None:
+                body["reasoning_effort"] = reasoning_effort
             req = urllib.request.Request(
                 url,
                 data=json.dumps(body).encode("utf-8"),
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
## Summary

Reading the diff at HEAD (`270e1a67`) against `origin/master`: the fix is correct and well-scoped. It closes the #4550 gap where gateway-backed chats forwarded `provider` but never `reasoning_effort`, so the browser reasoning control was a no-op in remote mode. The new `_gateway_reasoning_effort_for_request()` helper mirrors the local path's coercion and wires the value into both gateway request branches. One small parity gap worth a look before merge.

## Code reference

The new helper in `api/gateway_chat.py`:

```python
def _gateway_reasoning_effort_for_request(cfg, *, model=None, model_provider=None):
    ...
        coerced = coerce_reasoning_effort_for_model(
            effort_raw,
            model,
            provider_id=model_provider,
        )
        return None if not coerced else str(coerced)
```

Compare the local non-gateway path it's modeled on, `api/streaming.py:6765-6770`:

```python
_effort = coerce_reasoning_effort_for_model(
    _effort_raw,
    resolved_model,
    provider_id=resolved_provider,
    base_url=resolved_base_url,   # ← passed locally, omitted in the gateway helper
)
```

## Diagnosis

`coerce_reasoning_effort_for_model` forwards `base_url` into `resolve_model_reasoning_efforts(...)` (`api/config.py:3086-3090`), which is part of how it decides the supported-effort set for a given backend. The gateway helper drops `base_url`, so for a model whose supported-effort set is keyed off the endpoint URL, the coercion could clamp differently than the local path does for the same selection.

In practice this is usually benign: when `resolve_model_reasoning_efforts` returns an empty set (the common case for custom/aggregator gateway models it can't classify), the function preserves the raw configured effort verbatim (`config.py:3103-3104`), which is the permissive, do-no-harm outcome. So the field still gets forwarded; it just isn't clamped as tightly as the local path would. Still, for exact local/gateway parity it'd be cleaner to thread `base_url` through too if it's resolvable at this call site — otherwise a model that the local path degrades (e.g. `max` → `high`) could be forwarded as the un-degraded value over the gateway.

## Verification

The `"none"`-preservation comment is accurate — `coerce_reasoning_effort_for_model` returns `"none"` for an explicit none (`config.py:3079-3080`), and `None if not coerced` keeps that distinct from absent/invalid (which omit the field). Both request branches (runs API `body_extras` and legacy `body`) gate on `if reasoning_effort is not None`, so absent effort correctly omits the key rather than sending an empty string. CI is green across the matrix. The `base_url` note is a parity nit, not a blocker — your call whether to thread it through now or leave a follow-up.

nesquena-hermes:
Shipped in **v0.51.540** (via release #4558) — cherry-picked with your authorship preserved. The gateway chat path now forwards `reasoning_effort` (read + coerced with the same logic as the local path) on both the runs-API and legacy branches, so the browser effort control is honored in remote/gateway mode. Passed the full gate (Codex + Opus + 9741-test suite). Thanks @rodboev!

Issue #3737, opened 2026-06-06
bug(models): picked model reverts to previous/default model on send (likely #3448 profile-aware resolution, v0.51.290)


nesquena-hermes:
# bug(models): picked model reverts to the previous/default model on send (likely #3448 profile-aware resolution, v0.51.290)

## Summary

A user changes the model in the composer picker, sends a message, and the picker **snaps back to the previously-used model** instead of running the model they just selected. Reported in Discord #webui (2026-06-06).

> getting an issue where i change the model but when i send a message it reverts back to the latest model that was used
> — @123321

## What happens (client → server → client trace)

1. **Picker change** persists fine. `$('modelSelect').onchange` (`static/boot.js:1176`) sets `S.session.model`, writes the persisted model state, and POSTs `/api/session/update`. The dropdown shows the new model.
2. **On send**, the client sends the picked model to `/api/chat/start` (`static/messages.js:657-666`):
   ```js
   const _modelState=_chatPayloadModelState();   // model: S.session.model || modelSelect.value
   await api('/api/chat/start',{method:'POST',body:JSON.stringify({
     ..., model:_modelState.model, model_provider:_modelState.model_provider, ...
   })});
   ```
3. **Server resolves** the model (`api/routes.py:11301-11314`):
   ```py
   requested_model = body.get("model") or s.model
   requested_provider = body.get("model_provider") if "model_provider" in body else getattr(s, "model_provider", None)
   _pp_provider, _pp_default = _read_profile_model_config(s, requested_provider)
   model, model_provider, normalized_model = _resolve_compatible_session_model_state(
       requested_model, requested_provider,
       profile_provider=_pp_provider, profile_default_model=_pp_default,
   )
   ```
   When `normalized_model` is true, the handler returns the rewritten model as `effective_model` (`api/routes.py:11053-11054`).
4. **Client echoes it back into the picker** (`static/messages.js:670-675`):
   ```js
   if(startData.effective_model && S.session){
     S.session.model=startData.effective_model;
     ...
     if($('modelSelect')) _applyModelToDropdown(startData.effective_model, $('modelSelect'), ...);
   }
   ```
   → the just-picked model is overwritten in the dropdown. **This is the visible revert.**

## Leading root cause: profile-aware normalization (new in v0.51.290 / #3448)

[...]

PR #4554, opened 2026-06-20
fix: preserve cross-provider model selection across conversation turns

jja881:
## Problem

When a user selects a cross-provider model (e.g. a custom relay like `custom:glm-free-relay` providing `z-ai/glm-5.2-free`) in the WebUI model dropdown, the model selector silently "jumps" back to the profile default (e.g. `mimo-v2.5-pro` via `xiaomi`) after the **first message** in the conversation.

## Root Cause

Two bugs work together to cause the revert:

### 1. Frontend: `explicit_model_pick` marker is one-shot (`static/messages.js`)

The `explicit_model_pick` flag was designed to protect cross-provider selections from server-side "repair" (#3737). However, the marker set by `modelSelect.onchange` is **consumed and cleared after the first `send()`**:

```javascript
// Marker set on onchange, read once, then cleared:
const _explicitPick = _pendingPick && ...;
if (_explicitPick) _clearPendingSessionModel(activeSid);  // ← gone after first send
```

On the **second and subsequent sends**, `explicit_model_pick` is `undefined`/`false`, so the server no longer knows the user explicitly chose this model.

### 2. Backend: `_catalog_has_provider()` misses compound provider IDs (`api/routes.py`)

When `explicit_model_pick` is falsy, `_resolve_compatible_session_model_state()` calls `_catalog_has_provider()` to check if the session's model belongs to a known provider. For compound provider IDs like `custom:glm-free-relay`, the normalized form is `custom`, which **is** a raw catalog `provider_id`. But the three existing checks only compare:

- raw-vs-raw (`provider_raw in raw_provider_ids`)
- normalized-vs-normalized (`provider_normalized in normalized_provider_ids`)

They **miss** normalized-vs-raw (`provider_normalized in raw_provider_ids`) for compound hints. This causes the catalog lookup to fail, and the fallback branch at line ~3932 fires, silently reverting the model to the profile default.

### Frontend (`static/messages.js`)

Broaden the `explicit_model_pick` inference: when the session has a non-default model from a different provider than the profile's active provider, **every** `send()` sends `explicit_model_pick=true`.

The pending onchange marker is still consumed after the first send (unchanged behavior); only the `explicit_pick` inference is broadened.

### Backend (`api/routes.py`)

Add a fourth check to `_catalog_has_provider()` for compound provider IDs:

```python
or (
    provider_normalized
    and ":" in provider_raw
    and provider_normalized in raw_provider_ids
)
```

## Impact

- **New sessions / no manual model switch**: Unchanged — `explicit_model_pick` is still `false`, default model is used.
- **Cross-provider selection**: Now persists across all turns in the conversation, not just the first.
- **Same-provider selection**: Unchanged — `_isCrossProviderPick` is `false` when provider matches the profile default.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -3196,18 +3196,7 @@ def _catalog_has_provider(
         provider_raw in raw_provider_ids
         or (provider_normalized and provider_normalized in raw_provider_ids)
         or (provider_normalized and provider_normalized in normalized_provider_ids)
-        # Compound provider IDs like "custom:zenmux-relay" normalize to
-        # "custom" which is a raw catalog provider_id.  The three checks
-        # above compare raw-vs-raw and normalized-vs-normalized but miss
-        # normalized-vs-raw for compound hints.  Without this, a
-        # @custom:relay-name:model selection is silently reverted to the
-        # profile default on every non-explicit turn because the catalog
-        # lookup fails and the fallback branch at line ~3932 fires.
-        or (
-            provider_normalized
-            and ":" in provider_raw
-            and provider_normalized in raw_provider_ids
-        )
+
     )
 
 
--- a/api/routes.py
+++ b/api/routes.py
@@ -3196,6 +3196,18 @@ def _catalog_has_provider(
         provider_raw in raw_provider_ids
         or (provider_normalized and provider_normalized in raw_provider_ids)
         or (provider_normalized and provider_normalized in normalized_provider_ids)
+        # Compound provider IDs like "custom:zenmux-relay" normalize to
+        # "custom" which is a raw catalog provider_id.  The three checks
+        # above compare raw-vs-raw and normalized-vs-normalized but miss
+        # normalized-vs-raw for compound hints.  Without this, a
+        # @custom:relay-name:model selection is silently reverted to the
+        # profile default on every non-explicit turn because the catalog
+        # lookup fails and the fallback branch at line ~3932 fires.
+        or (
+            provider_normalized
+            and ":" in provider_raw
+            and provider_normalized in raw_provider_ids
+        )
     )
 
 
--- a/static/messages.js
+++ b/static/messages.js
@@ -1267,15 +1267,29 @@ async function send(){
     const _pendingPick=(typeof _readPendingSessionModel==='function')
       ? _readPendingSessionModel(activeSid)
       : null;
-    const _explicitPick=_pendingPick
+    const _pendingPickMatch=_pendingPick
       && _pendingPick.model===_modelState.model
       && String(_pendingPick.model_provider||'')===String(_modelState.model_provider||'');
+    // ── Persisted cross-provider pick (#3737 follow-up) ──
+    // The onchange marker is consumed after the first send, so subsequent sends
+    // lose explicit_model_pick and the server "repairs" the model back to the
+    // profile default.  When the session has a non-default model from a different
+    // provider than the profile's active provider, treat every send as explicit
+    // so the server honors the user's choice across the entire conversation.
+    const _defaultModel=(typeof window!=='undefined' && window._defaultModel)||'';
+    const _activeProvider=(typeof window!=='undefined' && window._activeProvider)||null;
+    const _isCrossProviderPick = _modelState.model
+      && _modelState.model_provider
+      && _defaultModel
+      && _modelState.model !== _defaultModel
+      && String(_modelState.model_provider||'') !== String(_activeProvider||'');
+    const _explicitPick = _pendingPickMatch || _isCrossProviderPick;
     // Consume the pending explicit-pick marker for THIS send only. The marker is
     // recorded on modelSelect.onchange and intentionally kept (not cleared on
     // session-update) so it survives the normal pick→update→send flow; clear it here
     // once read so a later send of an unchanged dropdown isn't treated as an explicit
     // pick. (#3739/#3737, Codex catch)
-    if(_explicitPick && typeof _clearPendingSessionModel==='function') _clearPendingSessionModel(activeSid);
+    if(_pendingPickMatch && typeof _clearPendingSessionModel==='function') _clearPendingSessionModel(activeSid);
     explicitPickForPostStart=_explicitPick;
     const startData=await api('/api/chat/start',{method:'POST',body:JSON.stringify({
       session_id:activeSid,message:msgText,
--- a/static/messages.js
+++ b/static/messages.js
@@ -1281,6 +1281,7 @@ async function send(){
     const _isCrossProviderPick = _modelState.model
       && _modelState.model_provider
       && _defaultModel
+      && _activeProvider
       && _modelState.model !== _defaultModel
       && String(_modelState.model_provider||'') !== String(_activeProvider||'');
     const _explicitPick = _pendingPickMatch || _isCrossProviderPick;
--- a/tests/test_catalog_has_provider_compound_ids.py
+++ b/tests/test_catalog_has_provider_compound_ids.py
@@ -1,0 +1,66 @@
+"""Tests for _catalog_has_provider() with compound provider IDs.
+
+Compound provider IDs like "custom:zenmux-relay" or "custom:glm-free-relay"
+normalize to "custom" via _normalize_provider_id(). The existing three
+clauses in _catalog_has_provider() already handle these correctly:
+
+  C1: exact raw match  ("custom:zenmux-relay" ∈ raw_ids)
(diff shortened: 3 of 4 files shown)


nesquena-hermes (changes requested):
Thanks for the PR — the `messages.js` fix is correct and addresses a real bug. A couple of issues in the `api/routes.py` change need to be corrected before this can merge.

## `api/routes.py` — 4th clause is dead code (minor fix needed)

The new 4th OR clause added to `_catalog_has_provider()`:

```python
or (
    provider_normalized
    and ":" in provider_raw
    and provider_normalized in raw_provider_ids
)
```

is logically redundant with the existing **Clause 2** (`provider_normalized and provider_normalized in raw_provider_ids`). The new clause is a strict subset of C2 — it only adds the restriction `":" in provider_raw`, meaning C4 can only fire when C2 already fires. There is no input combination where C4 returns True and C2 does not.

The explanatory comment says "The three checks above compare raw-vs-raw and normalized-vs-normalized" — but C2 IS a normalized-vs-raw comparison. That gap does not exist.

**Verification:** with a catalog containing `custom:zenmux-relay`, the old code already returns `True` via C1 (exact raw match `"custom:zenmux-relay" ∈ raw_ids`) or C3 (`"custom" ∈ norm_ids`). The 4th clause adds no new coverage.

**Fix:** Remove the 4th OR clause and its comment entirely. The old C1–C3 already handle compound provider IDs correctly.

## `static/messages.js` — correct, with an optional hardening

The `_isCrossProviderPick` logic is correct and addresses the real bug (onchange marker consumed after first send, so 2nd+ turns lose `explicit_model_pick`). The `_clearPendingSessionModel` call correctly moved to `_pendingPickMatch` only — good catch.

One optional hardening: add `&& _activeProvider` to the condition to prevent a spurious `True` when the profile has no configured default provider (currently `String(_modelState.model_provider||'') !== String(null||'')` → `'openai' !== ''` → `True`):

```js
const _isCrossProviderPick = _modelState.model
  && _modelState.model_provider
  && _defaultModel
  && _activeProvider                    // ← add this guard
  && _modelState.model !== _defaultModel
  && String(_modelState.model_provider||'') !== String(_activeProvider||'');
```

This is a hardening improvement — the over-fire is conservative (preserves the user's model choice rather than reverting it), so it is not a blocker, but it makes the intent clearer.

## Summary

- Remove the redundant 4th clause from `_catalog_has_provider()` in `api/routes.py`
- Keep the `messages.js` fix as-is (or with the optional `&& _activeProvider` guard)
- Add a test that verifies `_catalog_has_provider("custom:zenmux-relay", "custom", {"custom:zenmux-relay"}, {"custom"})` returns `True` (this would have caught the redundancy — the test trivially passes even without the 4th clause)

nesquena-hermes (dismissed):
Thanks for the PR — the `messages.js` fix is correct and addresses a real bug. A couple of issues in the `api/routes.py` change need to be corrected before this can merge.

## `api/routes.py` — 4th clause is dead code (minor fix needed)

The new 4th OR clause added to `_catalog_has_provider()`:

```python
or (
    provider_normalized
    and ":" in provider_raw
    and provider_normalized in raw_provider_ids
)
```

is logically redundant with the existing **Clause 2** (`provider_normalized and provider_normalized in raw_provider_ids`). The new clause is a strict subset of C2 — it only adds the restriction `":" in provider_raw`, meaning C4 can only fire when C2 already fires. There is no input combination where C4 returns True and C2 does not.

The explanatory comment says "The three checks above compare raw-vs-raw and normalized-vs-normalized" — but C2 IS a normalized-vs-raw comparison. That gap does not exist.

**Verification:** with a catalog containing `custom:zenmux-relay`, the old code already returns `True` via C1 (exact raw match `"custom:zenmux-relay" ∈ raw_ids`) or C3 (`"custom" ∈ norm_ids`). The 4th clause adds no new coverage.

**Fix:** Remove the 4th OR clause and its comment entirely. The old C1–C3 already handle compound provider IDs correctly.

## `static/messages.js` — correct, with an optional hardening

The `_isCrossProviderPick` logic is correct and addresses the real bug (onchange marker consumed after first send, so 2nd+ turns lose `explicit_model_pick`). The `_clearPendingSessionModel` call correctly moved to `_pendingPickMatch` only — good catch.

One optional hardening: add `&& _activeProvider` to the condition to prevent a spurious `True` when the profile has no configured default provider (currently `String(_modelState.model_provider||'') !== String(null||'')` → `'openai' !== ''` → `True`):

```js
const _isCrossProviderPick = _modelState.model
  && _modelState.model_provider
  && _defaultModel
  && _activeProvider                    // ← add this guard
  && _modelState.model !== _defaultModel
  && String(_modelState.model_provider||'') !== String(_activeProvider||'');
```

This is a hardening improvement — the over-fire is conservative (preserves the user's model choice rather than reverting it), so it is not a blocker, but it makes the intent clearer.

## Summary

- Remove the redundant 4th clause from `_catalog_has_provider()` in `api/routes.py`
- Keep the `messages.js` fix as-is (or with the optional `&& _activeProvider` guard)
- Add a test that verifies `_catalog_has_provider("custom:zenmux-relay", "custom", {"custom:zenmux-relay"}, {"custom"})` returns `True` (this would have caught the redundancy — the test trivially passes even without the 4th clause)

jja881:
Thanks for the thorough review! All three points addressed in `379a640`:

1. **Removed the redundant 4th clause** from `_catalog_has_provider()` — you're right, C2 (`provider_normalized in raw_provider_ids`) already covers the compound case. My comment claiming "the three checks compare raw-vs-raw and normalized-vs-normalized" was wrong — C2 IS normalized-vs-raw. The 4th clause was a strict subset with zero new coverage. Deleted entirely.

2. **Added `&& _activeProvider` guard** to `_isCrossProviderPick` in `messages.js` — good catch on the over-fire when no default provider is configured. Applied as suggested.

3. **Added `tests/test_catalog_has_provider_compound_ids.py`** — verifies that compound IDs like `custom:zenmux-relay` and `custom:glm-free-relay` are correctly recognized by the existing C1–C3 clauses (exact raw match, normalized-in-raw, normalized-in-norm, and a negative case).

nesquena-hermes:
## Summary

Reading the diff at this PR's HEAD (`379a640`) against `origin/master`: the frontend logic in `static/messages.js` is sound, the `api/routes.py` change is whitespace-only (the redundant 4th clause was correctly dropped), and the new compound-ID regression test is good. One real defect remains in the updated wiring test that the review reply didn't address — Greptile flagged it but it's worth pinning down exactly.

## Code reference

`tests/test_issue3737_explicit_pick_client_wiring.py:67` (PR HEAD):

```python
region = MESSAGES_JS[idx: idx + 1100]
...
assert re.search(r"_pendingPickMatch\s*&&[^\\n]*_clearPendingSessionModel", region), (
    "the consume-clear must be gated on _pendingPickMatch (a genuine, matching pending pick)"
)
```

On master the same assertion used `r"[^\n]*"` (raw string → regex `\n` = newline), which correctly bounds the match to a single logical line. The rename to `_pendingPickMatch` accidentally doubled the backslash to `r"[^\\n]*"`. In a raw string that character class is `[^` + literal-backslash + letter-`n` + `]`, i.e. "any char except a backslash or the letter n" — it no longer excludes newlines.

## Diagnosis

The assertion passes today only by luck: the source between the two anchors is ` && typeof ` (in `if(_pendingPickMatch && typeof _clearPendingSessionModel==='function')`), which contains no `n` and no backslash, so `[^\\n]*` happens to match it. But the guard no longer enforces "same line" — it would now match across a newline boundary as long as the gap contains no `n`/`\`, so a future refactor that splits the gate from the clear (or introduces an `n` in between) would silently break the guard's intent rather than failing loudly. That defeats the point of a structural regression test.

Fix is one character per the original master form:

```python
assert re.search(r"_pendingPickMatch\s*&&[^\n]*_clearPendingSessionModel", region), (
```

## Verification

The frontend dependency checks out: `window._defaultModel` / `window._activeProvider` that `_isCrossProviderPick` reads are populated in `static/boot.js:1977-1979` on session load, and the `&& _activeProvider` guard you added correctly prevents over-fire when no default provider is configured. So the cross-provider persistence logic itself is fine — this is purely a test-quality nit on the regex bound. Worth fixing before merge so the guard actually guards.

nesquena-hermes:
Shipped in **v0.51.540** (via release #4558) — cherry-picked with your authorship preserved. Cross-provider relay models (e.g. `custom:glm-free-relay`) now stay selected across the whole conversation instead of reverting to the profile default after the first message. Passed the full gate (Codex + Opus + 9741-test suite). Thanks @jja881 — and nice work converging the bounce!

PR #4555, opened 2026-06-20
fix(models): use stale disk cache on catalog timeout

starship-s:
## Thinking Path

- `/api/models` keeps the foreground request bounded so a slow provider probe cannot hang the model picker.
- #4052 made the over-budget fallback useful for cold/no-cache cases by returning a richer static catalog instead of a one-model `Default` group.
- There is still a middle case where `models_cache.json` exists and has a valid `/api/models` payload, but strict metadata/fingerprint validation rejects it for normal cold-path use.
- In that case, falling straight to the static catalog can drop provider groups that the last-known cache already had.
- The narrow fix is to use a shape-valid stale disk cache only as the over-budget foreground response, while leaving strict cold-path cache validation and background live refresh publication unchanged.

## What Changed

- Added a shape-only stale disk-cache loader for `/api/models` timeout fallback.
- Updated the bounded live-rebuild timeout path to try that stale cache before the static fallback.
- Kept the stale payload response-only: it is not written into the in-memory catalog cache and is not saved back to disk.
- Added regression coverage for valid stale cache use, invalid stale cache fallback, missing `aliases`, and continued out-of-band live publication.

## Why It Matters

- The first model-picker response after a slow live rebuild can preserve last-known provider groups instead of dropping to a smaller static catalog.
- Users with providers discovered through live catalogs are less likely to see options disappear during a timeout.
- The live rebuild still refreshes the authoritative cache when it completes, so stale cache use remains a temporary fallback rather than a new source of truth.
- The final static fallback remains available when no shape-valid disk cache exists.

## Verification

Targeted coverage:

```text
./scripts/test.sh tests/test_issue3928_models_budget_fallback.py tests/test_wakeup_model_resolve_hang.py -q
# 14 passed in 4.18s
```

Syntax / hygiene checks:

```text
python -m py_compile api/config.py tests/test_issue3928_models_budget_fallback.py
# passed

git diff --check
# passed
```

Independent review:

```text
Claude Opus review found no blocking issues.
```

## Risks / Follow-ups

- This intentionally does not relax the strict cold-path disk cache loader; stale cache use is limited to the over-budget foreground fallback.
- The disk cache currently may not contain every optional top-level field from a live response; the stale fallback defaults missing `aliases` to `{}` and the background rebuild still publishes the fresh full result when it completes.

## Model Used

- OpenAI / GPT-5.5 via OpenCode for implementation.
- Anthropic / Claude Opus 4.8 for independent review.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -4612,6 +4612,36 @@ def _load_models_cache_from_disk() -> dict | None:
         return None
 
 
+def _load_stale_models_cache_from_disk() -> dict | None:
+    """Load a shape-valid stale /api/models disk cache for timeout fallback only.
+
+    The main cache loader enforces metadata stamps for a full cold-path cache hit.
+    This helper intentionally does not apply that stricter policy, so we can still
+    recover a useful fallback payload when the strict loader rejected cache because
+    metadata or fingerprint fields are stale.
+    """
+    try:
+        import json as _j
+
+        cache_path = _get_models_cache_path()
+        if not cache_path.exists():
+            return None
+        with open(cache_path, encoding="utf-8") as f:
+            cache = _j.load(f)
+        if not _is_valid_models_cache(cache):
+            return None
+        aliases = cache.get("aliases")
+        return {
+            "active_provider": cache["active_provider"],
+            "default_model": cache["default_model"],
+            "configured_model_badges": cache["configured_model_badges"],
+            "groups": cache["groups"],
+            "aliases": aliases if isinstance(aliases, dict) else {},
+        }
+    except Exception:
+        return None
+
+
 def _save_models_cache_to_disk(cache: dict) -> None:
     """Save cache to disk so it survives server restarts.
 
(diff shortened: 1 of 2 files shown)


starship-s:
Thanks for the review — pushed a follow-up in `c531e3b7` addressing both notes:

- moved the shape-only stale disk-cache read before `_available_models_cache_lock` is acquired, so the over-budget fallback no longer adds disk I/O while holding the cache lock or delaying the background worker's publish path
- widened the new out-of-band publication polling deadlines from `1.0s` to `3.0s` to reduce CI flake risk on loaded runners

Verification:

```text
./scripts/test.sh tests/test_issue3928_models_budget_fallback.py tests/test_wakeup_model_resolve_hang.py -q
# 14 passed

python -m py_compile api/config.py tests/test_issue3928_models_budget_fallback.py
# passed

git diff --check
# passed
```

Current GitHub checks on `c531e3b7` are all green.

nesquena-hermes:
Shipped in **v0.51.541** (via release #4559) — cherry-picked with your authorship preserved. The over-budget rebuild now recovers a shape-valid stale disk cache instead of dropping to a thin Default group.

Three hardenings were applied at the ship-gate (Codex + Opus): a `_schema_version` guard so a cross-schema cache can't reach the picker; reconstruction of `/api/models.aliases` from config in both disk loaders (the save path never persisted aliases, so `/model <alias>` was silently breaking on any disk-cache hit — this also fixed a pre-existing strict-loader gap); and test-fixture isolation to prevent an order-dependent flake. Passed the full gate (Codex SAFE + Opus SAFE + 9746-test suite). Thanks @starship-s!

PR #4561, opened 2026-06-20
feat: add extension manifest asset bundles

santastabber:
## Thinking Path

- Hermes WebUI already has a deliberately small, admin-controlled extension surface for local JS/CSS injection.
- That works well for one extension, but multi-extension/self-hosted bundles currently require long comma-separated env-var asset lists.
- A manifest inside the existing extension directory can improve bundling without adding a marketplace, dependency system, new routes, new CSP allowances, or a frontend build step.
- This PR keeps the existing static serving and URL validation contract, then layers a JSON asset list on top of it.

## What Changed

- Added `HERMES_WEBUI_EXTENSION_MANIFEST`, a relative JSON manifest path resolved inside `HERMES_WEBUI_EXTENSION_DIR`.
- Manifest files may list `scripts` and `stylesheets` either:
  - as top-level keys,
  - inside an object `extensions` array, or
  - as a top-level array of extension objects.
- Bare manifest asset paths like `templates/app.js` normalize to `/extensions/templates/app.js`.
- Absolute same-origin `/extensions/...` and `/static/...` asset paths continue through the existing URL validator.
- Existing `HERMES_WEBUI_EXTENSION_SCRIPT_URLS` and `HERMES_WEBUI_EXTENSION_STYLESHEET_URLS` still work and append after manifest assets.
- Updated `docs/EXTENSIONS.md` and the README env-var table.
- Added regression tests for manifest ordering, unsafe URL rejection, invalid manifest paths, malformed field shapes, disabled entries, top-level arrays, env interaction, duplicate compatibility, and cap behavior.

## Why It Matters

This makes the existing extension feature easier to package for self-hosted/local extension bundles while preserving the current trust model:

- no new plugin marketplace semantics;
- no third-party script loading;
- no backend routes or proxying;
- no dependency resolver;
- no filesystem path exposure;
- no change to static asset authentication or serving behavior.

## Verification

- `git diff --check`
- `python3 -m py_compile api/extensions.py server.py bootstrap.py`
- `.venv/bin/ruff check api/extensions.py tests/test_extension_hooks.py`
- `./scripts/test.sh tests/test_extension_hooks.py tests/test_pr1445_opus_followups.py tests/test_contract_review_gate.py tests/test_docs_gitignore_policy.py -q`
  - Result: `26 passed`
- Isolated server smoke with temporary `HERMES_HOME`, `HERMES_WEBUI_STATE_DIR`, and manifest-backed `HERMES_WEBUI_EXTENSION_DIR`:
  - `/health` returned `status: ok`
  - `/` contained manifest-injected script/style tags
  - `/extensions/templates/templates.js` served the manifest JS
  - `/extensions/templates/templates.css` served as `text/css; charset=utf-8`

## Contract Routing

Task type: extension setup/usability improvement.

Touched areas:
- extension injection config
- extension documentation
- README env-var table
- extension security regression tests

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/EXTENSIONS.md`

Scope boundaries:
- Does not add extension install UI.
- Does not add extension backend routes or proxy support.
- Does not loosen CSP or same-origin asset rules.
- Does not turn extensions into a marketplace/dependency system.

- focused extension/security tests pass;
- existing Opus follow-up extension hardening tests pass;
- docs accurately describe the manifest as a local admin bundling convenience.

## Risks / Follow-ups

- The manifest is parsed on app-shell render; it is capped at 64 KiB and intended for small local asset lists.
- Route/proxy support remains a separate, larger design question and is intentionally not included here.
- A future extension manager UI or CLI installer could generate this manifest instead of asking admins to maintain comma-separated env vars by hand.

## Model Used

AI-assisted. Hermes Agent used OpenRouter `openai/gpt-5.5` for implementation orchestration and Anthropic Claude Code CLI for a read-only critical review pass. Claude reviewed the working-tree diff for security, compatibility, docs, and test coverage; no public GitHub writes were performed by the assistant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,8 +344,9 @@ Full list of environment variables:
 | `HERMES_WEBUI_CSP_CONNECT_EXTRA` | *(unset)* | Optional space-separated `http(s)://` or `ws(s)://` origins to append to the report-only CSP `connect-src` directive for reverse-proxy or tunnel deployments |
 | `HERMES_WEBUI_SSE_CHUNKED` | *(unset)* | Set truthy (`1`/`true`/`yes`/`on`) to send SSE with `Transfer-Encoding: chunked`. Needed behind buffering reverse proxies (e.g. `jupyter-server-proxy`) that otherwise buffer the whole stream; harmless but unnecessary for directly-served deployments |
 | `HERMES_WEBUI_EXTENSION_DIR` | *(unset)* | Optional local directory served at `/extensions/`; must point to an existing directory before extension injection is enabled |
-| `HERMES_WEBUI_EXTENSION_SCRIPT_URLS` | *(unset)* | Optional comma-separated same-origin script URLs to inject; see [WebUI Extensions](docs/EXTENSIONS.md) |
-| `HERMES_WEBUI_EXTENSION_STYLESHEET_URLS` | *(unset)* | Optional comma-separated same-origin stylesheet URLs to inject; see [WebUI Extensions](docs/EXTENSIONS.md) |
+| `HERMES_WEBUI_EXTENSION_MANIFEST` | *(unset)* | Optional relative JSON manifest inside `HERMES_WEBUI_EXTENSION_DIR` listing bundled scripts/styles to inject; see [WebUI Extensions](docs/EXTENSIONS.md) |
+| `HERMES_WEBUI_EXTENSION_SCRIPT_URLS` | *(unset)* | Optional comma-separated same-origin script URLs to inject; appended after manifest scripts; see [WebUI Extensions](docs/EXTENSIONS.md) |
+| `HERMES_WEBUI_EXTENSION_STYLESHEET_URLS` | *(unset)* | Optional comma-separated same-origin stylesheet URLs to inject; appended after manifest stylesheets; see [WebUI Extensions](docs/EXTENSIONS.md) |
 | `HERMES_HOME` | Windows: `%LOCALAPPDATA%\hermes`; POSIX: `~/.hermes` | Base directory for Hermes state (affects all paths) |
 | `HERMES_CONFIG_PATH` | `$HERMES_HOME/config.yaml` | Path to Hermes config file |
 | `HERMES_WEBUI_SERVER_CWD` | *(unset)* | Working directory for the server process. Defaults to the agent dir; point it at a writable workspace when the agent dir is read-only so fallback relative writes land somewhere writable |
--- a/api/extensions.py
+++ b/api/extensions.py
@@ -6,10 +6,11 @@ It is disabled by default and never executes or fetches third-party URLs.
 """
 
 import html
+import json
 import logging
 import os
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 from urllib.parse import unquote, urlsplit
 
 from api.helpers import _security_headers, j
@@ -22,14 +23,24 @@ _log = logging.getLogger(__name__)
 # avoids rendering tens of thousands of <script> tags into every page load.
 _MAX_URL_LIST = 32
 
+# Keep extension manifests small and auditable. The manifest is a convenience for
+# bundling static assets, not a package manager or dependency lockfile.
+_MAX_MANIFEST_BYTES = 64 * 1024
+
 # Tracks rejected URL strings we've already warned about so a misconfigured env
 # var doesn't spam the log on every request that re-reads it.
 _warned_urls: set = set()
+
+
+class _ManifestTooLarge(ValueError):
+    pass
+
 
 EXTENSION_ROUTE_PREFIX = "/extensions/"
 _EXTENSION_DIR_ENV = "HERMES_WEBUI_EXTENSION_DIR"
 _EXTENSION_SCRIPT_URLS_ENV = "HERMES_WEBUI_EXTENSION_SCRIPT_URLS"
 _EXTENSION_STYLESHEET_URLS_ENV = "HERMES_WEBUI_EXTENSION_STYLESHEET_URLS"
+_EXTENSION_MANIFEST_ENV = "HERMES_WEBUI_EXTENSION_MANIFEST"
 _ALLOWED_ASSET_PREFIXES = ("/extensions/", "/static/")
 
 _EXTENSION_MIME = {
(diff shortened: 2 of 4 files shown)


santastabber:
Updated the branch to address the two Greptile P2 findings in `d3138ace`:

- Replaced the manifest `stat().st_size` + `read_text()` sequence with a byte-bounded read before JSON parsing, so oversized manifests fail closed before content validation.
- Added manifest script/style cap sentinels and stable per-list warning keys (`manifest:scripts`, `manifest:stylesheets`) so truncation warns once per list instead of once per remaining extension entry.

Added regression coverage for:

- oversized valid JSON that would inject `pwn.js` if the cap were removed;
- oversized multibyte payloads, enforcing byte-based bounds;
- over-cap manifests with many entries, asserting one truncation warning per script/style list.

Verification after the update:

```text
git diff --check
.venv/bin/python scripts/ruff_lint.py --diff origin/master
python3 -m py_compile api/extensions.py server.py bootstrap.py
.venv/bin/ruff check api/extensions.py tests/test_extension_hooks.py
./scripts/test.sh tests/test_extension_hooks.py tests/test_pr1445_opus_followups.py tests/test_contract_review_gate.py tests/test_docs_gitignore_policy.py -q
```

Result: `29 passed`; all GitHub checks are green on the updated head.

nesquena-hermes:
Read the diff at `d3138ace` against `origin/master` — `api/extensions.py` plus the new `tests/test_extension_hooks.py`. The design is sound: manifest paths go through `_is_safe_relative_path` + a `resolve()`/`relative_to(root)` containment check (`api/extensions.py:169-185`), bare entries normalize under `/extensions/` and absolute paths still pass through `_append_safe_asset_url`, and the 64 KiB bound reads `_MAX_MANIFEST_BYTES + 1` so the over-cap case is detected without slurping an unbounded file. No traversal or injection gap that I can see.

## Confirming + widening the exception-ordering bug

The automated review flagged `json.JSONDecodeError` being swallowed by the `except ValueError`. That's real, and it's actually broader than stated — the `could not be read` branch is **nearly dead code**. Here's the block (`api/extensions.py:243-249`):

```python
manifest = json.loads(_read_manifest_text(manifest_file))
except ValueError:
    _log.warning("Configured extension manifest exceeds %d bytes", _MAX_MANIFEST_BYTES)
    return [], []
except (OSError, UnicodeDecodeError, json.JSONDecodeError):
    _log.warning("Configured extension manifest could not be read")
    return [], []
```

Both of the exceptions in the second handler subclass `ValueError`:

```
JSONDecodeError subclasses ValueError: True
UnicodeDecodeError subclasses ValueError: True   # UnicodeDecodeError -> UnicodeError -> ValueError
```

Python evaluates `except` clauses top-down, so line 244 wins for *both*. Result: a malformed-JSON manifest **and** a non-UTF-8 manifest (the `.decode("utf-8")` in `_read_manifest_text`, `api/extensions.py:228-232`) both log `exceeds 64 bytes`, which is actively misleading to an operator debugging a bad file. Only a genuine `OSError` (permissions, mid-read IO error) reaches line 247 with the right message.

## Fix sketch

The size error is the *only* intended `ValueError` here, so make it distinguishable. Two options:

Minimal — reorder so the specific subclasses are caught first (`api/extensions.py:243`):

```python
manifest = json.loads(_read_manifest_text(manifest_file))
except json.JSONDecodeError:
    _log.warning("Configured extension manifest is not valid JSON")
    return [], []
except (OSError, UnicodeDecodeError):
    _log.warning("Configured extension manifest could not be read")
    return [], []
except ValueError:  # only the explicit "manifest too large"
    _log.warning("Configured extension manifest exceeds %d bytes", _MAX_MANIFEST_BYTES)
    return [], []
```

More robust — give the size guard its own type so it can't be confused with a stdlib `ValueError` ever again. Define `class _ManifestTooLarge(ValueError): ...` at module scope, raise it in `_read_manifest_text` instead of bare `ValueError`, and catch `_ManifestTooLarge` explicitly. That survives future refactors that add other `ValueError`-raising calls into the `try`.

## Test gap

`tests/test_extension_hooks.py` covers the cap path, but I don't see a case asserting the *log message* for a malformed-JSON manifest versus an over-size one — which is exactly what regresses here. Worth adding two tests using `caplog`: one with `{bad json` and one with a >64 KiB file, asserting each emits its own distinct warning. Without that, the reorder could silently regress again.

Everything else looks merge-ready once the handler order is fixed.

santastabber:
Updated the branch again to address the exception-ordering issue in the manifest read path.

Changes in `563be3c`:

- Added a dedicated `_ManifestTooLarge` exception for the manifest size guard.
- Switched `_read_manifest_text()` to a true bounded binary read (`fh.read(_MAX_MANIFEST_BYTES + 1)`) before UTF-8 decode / JSON parsing.
- Split the handlers so each failure logs accurately:
  - oversize → `Configured extension manifest exceeds ...`
  - malformed JSON → `Configured extension manifest is not valid JSON`
  - unreadable / invalid UTF-8 → `Configured extension manifest could not be read`
- Added `caplog` regressions covering all three branches, including assertions that malformed JSON and invalid UTF-8 no longer get mislabeled as oversize.

Re-ran:

```text
git diff --check
.venv/bin/python scripts/ruff_lint.py --diff origin/master
python3 -m py_compile api/extensions.py server.py bootstrap.py
.venv/bin/ruff check api/extensions.py tests/test_extension_hooks.py
./scripts/test.sh tests/test_extension_hooks.py tests/test_pr1445_opus_followups.py tests/test_contract_review_gate.py tests/test_docs_gitignore_policy.py -q
```

Result: `32 passed` locally before pushing.

nesquena-hermes:
Shipped in **v0.51.545** (Release TD) via release #4567 — cherry-picked with your authorship preserved. 🎉

Clean, in-character augmentation of the extension system — the manifest keeps the exact same same-origin allowlist validator (no third-party fetch), traversal-guarded path, 32-URL cap, default-off. Deep-gated (Codex + Opus + suite 9851 passed). Opus adversarially verified traversal (../, %2e%2e, in-dir symlink→/etc all rejected) and the allowlist (https://evil, //evil, javascript:, data: all rejected). One brick caught + fixed at the gate: a ≤64KB but deeply-nested manifest raised RecursionError in json.loads and would 503 every page — added RecursionError to the fail-safe handler + a regression test (crediting you as co-author). Thanks @santastabber!

PR #4576, opened 2026-06-21
Fix Transparent Stream anchor scene reload

### LINKED ISSUE #4568: bug(transparent-stream): tool-call rows vanish after page reload — persisted anchor activity scene is compact-only (sibling of #4539)
## Summary

In **Transparent Stream** activity mode, tool-call rows render correctly during streaming, but after the turn settles **and the page is reloaded** (hard refresh / new tab / fresh load) the tool rows are **gone entirely** — only the assistant prose (and thinking, in some cases) remains. The persisted transcript still contains the tool data, but the transparent-mode rehydrate renders it nowhere.

This is the **reload/rehydrate sibling** of #4539 (the in-session settle drop, fixed in PR #4542 / v0.51.536). #4542 fixed the *live settle* path (cleanup selector + burst-ID bucketing). It did **not** touch the path that runs on a fresh page load, which is broken by a *different* root cause described below.

> Reported by **AvidFuturist** (Discord #webui, 2026-06-20), still reproducing on current master (v0.51.544):
> "Streaming in transparent activity mode the tool calls look fine, but then after it is done and I refresh I don't see the tool calls at all anymore?"

**The "old session vs new session" split the reporter saw is the tell:** an *older* session (settled before this anchor-scene persistence shipped, or one with no persisted scene) reloads through the legacy derived-tool-calls path and shows the collapsed `Trace: N tools` worklog component correctly. A *newer* session whose last turn was settled with an anchor activity scene persisted reloads through the anchor-scene path and shows nothing in transparent mode.

## Steps to reproduce

1. Settings → Activity display → **Transparent Stream**.
2. Send a message that triggers tool calls (search / terminal / file).
3. Watch the tool rows appear live. ✅
4. Let the turn finish. (With #4542 in place, they now survive the in-session settle. ✅)
5. **Hard-reload the page** (or close and reopen the tab on that session).
6. **Bug:** the tool rows are gone — they render nowhere in the rehydrated turn.

## Root cause

The anchor **activity-scene** system (`activity_scene_v1`, #3926) is **compact-worklog-only on both the projection and the settled-render side, but it is persisted unconditionally on every settle regardless of the active display mode.** That asymmetry creates a dead-end on reload in transparent mode.

### 1 — The scene is projected + persisted in compact mode even when the user is in transparent mode

`static/messages.js`, `_projectLiveAnchorActivityScene()` hardcodes the mode:

```js
// static/messages.js ~L2255-2258
if(!_anchorRegistry||!_anchorApi||typeof _anchorApi.projectAssistantTurnAnchorActivityScene!=='function') return null;
return _anchorApi.projectAssistantTurnAnchorActivityScene(_anchorRegistry,{mode:'compact_worklog'});
```

and `_attachProjectedAnchorSceneToLastAssistant()` is called from the turn-settle done-handler **with no mode gate** (`static/messages.js` ~L4175, and the other settle call-sites at ~L4466/4474/4478/4617/4630/4724/4794):

```js
// static/messages.js ~L4175  (inside the `done` handler — no isTransparentStream/isCompactWorklogMode guard)
_attachProjectedAnchorSceneToLastAssistant(S.messages);
```

`_attachProjectedAnchorSceneToLastAssistant` then writes `lastAsst._anchor_activity_scene = scene` and calls `_persistSettledAnchorScene(...)` → the server stores it in `session.anchor_activity_scenes` (`api/routes.py` `_hydrate_anchor_activity_scenes`, `api/models.py`). So a **transparent-mode** session ends up with a **compact-mode** scene persisted on its last assistant message. The `_anchorRegistry` is fed on every source event regardless of mode (`applyAssistantTurnAnchorSourceEvent`, `static/messages.js` ~L2127), so the scene is always non-empty and always gets attached.

### 2 — On reload, an `_anchor_activity_scene` message is marked "anchor-owned" and SKIPPED by every transparent tool-render path

`static/ui.js` ~L11106-11116 builds `anchorOwnedAssistantRawIdxs` from any message carrying `_anchor_activity_scene`:

```js
const anchorOwnedAssistantRawIdxs=new Set();
for(const [rawIdx,seg] of assistantSegments){
  const msg=S.messages[rawIdx];
  if(!msg||!msg._anchor_activity_scene||!seg) continue;
  ... anchorOwnedAssistantRawIdxs.add(idx) ...
}
```

Every tool-render path then *excludes* those indices:
- the derived-tool-calls fallback (the legacy reload path that builds `Trace: N tools`): `static/ui.js` ~L11126 and ~L11159 (`if(anchorOwnedAssistantRawIdxs.has(rawIdx)) return;`);
- the transparent-stream rebuild loop: `static/ui.js` ~L11355 / ~L11366 / ~L11376 (`if(anchorOwnedAssistantRawIdxs.has(aIdx)) continue;`).

### 3 — The only renderer that *would* draw the persisted scene refuses to run in transparent mode

The anchor-owned message is supposed to be drawn by `_renderSettledAnchorSceneForMessage` (called at `static/ui.js` ~L11519). But its second line is a hard mode gate:

```js
// static/ui.js ~L9255-9256
function _renderSettledAnchorSceneForMessage(message, segment, rawIdx){
  if(!message||!message._anchor_activity_scene||!segment) return false;
  if(typeof isCompactWorklogMode==='function'&&!isCompactWorklogMode()) return false;   // ← no-op in transparent mode
  ...
}
```

**Net effect in transparent mode on reload:** the message is anchor-owned → skipped by the derived path AND the transparent rebuild → and the settled-scene renderer no-ops → **the tool rows render nowhere.** During live streaming they appear because `renderLiveAnchorActivityScene` / the live tool-SSE path draws them; switching tab/session triggers a re-render that (per #4539) repopulates from live state — but a *cold reload* has no live state, only the persisted compact-only scene, so it dead-ends.

### Why #4542 didn't cover this

PR #4542's regression test (`tests/test_issue4539_transparent_tool_settle.py`) only asserts the in-session settle cleanup selector and the burst-ID bucketing fallback. There is **no test for the reload/rehydrate path** where a persisted `_anchor_activity_scene` short-circuits the transparent render. That's the gap.

## Suggested fix

The robust root fix makes the anchor-scene system **mode-consistent** — a session can be *viewed* in either display mode regardless of which mode it was *settled* in, so persistence and rendering must both handle both modes:

**Primary (complete) fix — make the settled-scene renderer + anchor-owned skip mode-aware.**
- Persist the scene tagged with the **active** mode (drop the hardcoded `'compact_worklog'` in `_projectLiveAnchorActivityScene`; pass the live `chatActivityMode()`), and store `scene.mode` (the schema already carries a `mode` field).
- In `_renderSettledAnchorSceneForMessage`, when `isTransparentStream()`, render the persisted `scene.activity_rows` as transparent event rows (reuse `_decorateTransparentEventRow` + the live transparent row builders that `renderLiveAnchorActivityScene` uses) instead of `return false`. The anchor-owned skip in the transparent rebuild then correctly hands rendering off to this renderer.

**Mitigation-only alternative (partial — does NOT fix all cases): gate persistence.** Wrap `_attachProjectedAnchorSceneToLastAssistant` in `if(isCompactWorklogMode())` so transparent sessions persist only raw per-message tool metadata and fall through to the existing derived-tool-calls reload path. ⚠️ This still breaks the cross-mode case: a session **settled in compact mode** then **viewed in transparent mode** after reload still hits the same dead-end (persisted scene exists → anchor-owned → renderer no-ops). So gating persistence alone is incomplete — the primary fix above is required to handle mode-switching.

**Required regression coverage:** a reload-path test asserting that an assistant message carrying `_anchor_activity_scene` renders its tool rows in **both** display modes (the current #4539 test only covers the live in-session settle).

## Scope

Front-end, `static/ui.js` + `static/messages.js` (+ possibly `static/assistant_turn_anchors.js` for mode-tagged projection). Single repo (`nesquena/hermes-webui`), no agent/backend change — the persisted data is intact, this is purely a rehydrate/render-mode bug. Estimated moderate (the renderer needs a transparent row-builder branch); the mitigation-only variant is ~1-3 LOC but incomplete per the cross-mode caveat above.

## Reported by

**AvidFuturist** — Discord #webui, 2026-06-20. Reproduces on current master (v0.51.544). Sibling of #4539 / PR #4542 (in-session settle, shipped v0.51.536); same user-visible symptom on the cold-reload path, distinct root cause (compact-only persisted anchor scene).

franksong2702:
Fixes #4568.

## Thinking Path

#4568 reports that Transparent Stream can lose tool/thinking rows after reload or session re-entry for turns that already have a persisted `_anchor_activity_scene`. The persisted data is intact: `renderMessages()` correctly treats those assistant messages as anchor-owned and skips the legacy derived-tool rebuild, but `_renderSettledAnchorSceneForMessage()` was Compact Worklog-only and returned early in Transparent Stream mode. The fix is to make Transparent Stream own the persisted anchor scene on settled/reload rather than falling back to legacy reconstruction or gating persistence.

## What Changed

- Added a Transparent Stream settled anchor-scene renderer that converts persisted anchor activity rows into transparent event rows.
- Kept final answer prose owned by the assistant segment so anchor `prose` rows do not duplicate the answer.
- Preserved the existing Compact Worklog renderer path and left live anchor projection scoped to Compact Worklog.
- Added a regression test for the anchor-owned Transparent Stream reload path.
- Added an Unreleased changelog entry.

## Why It Matters

This closes the cross-mode reload gap after #4411: a turn can settle with the stable assistant-turn anchor scene and later be viewed or reloaded in Transparent Stream without losing tool/thinking activity. It also keeps the long-term architecture aligned with #3926 by making both display modes consume the same persisted anchor scene instead of growing a second normalizer.

## Verification

- `node --check static/ui.js`
- `./scripts/test.sh tests/test_live_to_final_anchor_visible_order.py -q`
- `./scripts/test.sh tests/test_issue3820_chat_activity_display_mode.py -q`
- `./scripts/test.sh tests/test_stable_assistant_turn_anchor_registry.py -q`
- `git diff --check`
- GitHub checks: browser-smoke, lint, and Python 3.11/3.12/3.13 matrix all passed

## Risks / Follow-ups

- This intentionally fixes settled/reload/re-entry rendering only. Live Transparent Stream behavior remains on the existing direct event-row path.
- Lifecycle/status anchor rows are rendered as quiet transparent rows, while final prose is still rendered by the normal assistant message segment to avoid duplicate answers.

## Model Used

Codex GPT-5


franksong2702:
Follow-up pushed in `2ffb2fdc8`.

- Addressed Greptile's multi-round prose gap: Transparent Stream now replays intermediate anchor prose as a normal assistant segment while still skipping final-answer prose duplicates.
- Added a behavioral Node regression test for intermediate prose vs final-answer duplicate handling.
- Local checks passed: `node --check static/ui.js`, `./scripts/test.sh tests/test_live_to_final_anchor_visible_order.py -q`, `./scripts/test.sh tests/test_issue3820_chat_activity_display_mode.py -q`, `./scripts/test.sh tests/test_stable_assistant_turn_anchor_registry.py -q`, `git diff --check`.

nesquena-hermes:
Thanks @franksong2702 — your persisted-anchor-scene rendering approach here is the right architecture for #4568 (both display modes consuming the same `activity_scene_v1` instead of a second normalizer, aligned with #3926), and I've built on it directly.

I verified your branch against a checkpoint matrix that drives a real multi-round streaming turn (12 tool calls + reasoning + between-tool progress prose + final answer) through every transcript lifecycle checkpoint — during-stream, after-settle, switch-away-and-back, tab hide/show, window blur/focus, hard reload, and mid-stream reload — in both display modes. Your fix reliably restored the **tool and thinking rows** at every checkpoint (the core of the bug). The one gap it surfaced: dropping *all* `prose` rows to avoid duplicating the final answer also dropped the **intermediate between-tool progress narration** from earlier rounds (the matrix showed `tools=12` but `texts=1`).

I've folded that into a combined PR — **#4587** — which keeps your approach and adds intermediate-prose preservation: `_anchorSceneTransparentNodeForRow` renders intermediate prose as inline segments and suppresses **only** the prose row that duplicates the final answer (length-ratio-guarded match, per Codex/Opus review). Result: tools + thinking + intermediate prose all survive, interleaved, at every checkpoint in both modes. Full suite green (9881 passed), Codex + Opus both reviewed.

You're credited as co-author on #4587. Closing this in favor of that combined PR — and thank you, the architecture there is yours.

franksong2702:
Closing as superseded by #4587. Nathan's follow-up confirms this branch's persisted-anchor-scene approach was folded into the combined Transparent Stream PR, with the additional intermediate-prose handling there. Keeping this closed avoids two overlapping PRs for the same #4568 path.

Issue #3058, opened 2026-05-28
RFC: WebUI pending intent controls for Queue, Interrupt, and Steer


franksong2702:
# RFC: WebUI Pending Intent Controls

- **Status:** Proposed
- **Author:** @franksong2702
- **Created:** 2026-05-28
- **Tracking issue:** [#3058](https://github.com/nesquena/hermes-webui/issues/3058)
- **Related issues:** [#2555](https://github.com/nesquena/hermes-webui/issues/2555), [NousResearch/hermes-agent#28172](https://github.com/NousResearch/hermes-agent/issues/28172)
- **Related docs:** [`webui-run-state-consistency-contract.md`](webui-run-state-consistency-contract.md), [`hermes-run-adapter-contract.md`](hermes-run-adapter-contract.md), [`turn-journal.md`](turn-journal.md)

## Problem

Hermes WebUI currently lets the user choose what happens when they send a
message while an agent run is active: queue, interrupt, or steer. The settings
surface describes three modes, but the underlying state model is not yet one
coherent product contract.

The most visible failure is `steer`: a successful steer is delivered through
`POST /api/chat/steer` and shown with a transient browser-only indicator. That
indicator is not part of the session transcript, not part of the run journal,
and not a durable pending item. If the user switches sessions and comes back,
or if the active pane re-renders from the server session, the visible evidence
that the user sent the steer can disappear.

Queue and interrupt have related ambiguity. Queue is session-scoped and partly
persisted in browser `sessionStorage`, but older manual testing still describes
session switching as "clearing" the queue. Interrupt is implemented as
queue-then-cancel, which is practical, but the user-facing intent is different
from an ordinary queued follow-up. Leftover steer text may also be converted to
queue after a run ends, which makes sense only if the product explains that
transition.

The current Settings wording reinforces the old model. The setting is named
`Busy input mode`, offers three values (`Queue follow-up`, `Interrupt current
turn`, `Steer (mid-turn correction)`), and describes the choice as "what
happens when you send a message while the agent is running." The proposed model
changes that product contract: Settings should choose the default waiting mode
for a new pending message, not whether the send action is permanently Queue,
Interrupt, or Steer.

The missing product primitive is a durable running-message submission model.
When the user sends while the agent is running, WebUI should not treat Queue,
Steer, and Interrupt as three unrelated permanent settings. Settings should
choose the default behavior for a new running-time message:

- Queue creates a pending message that waits for the current agent run to
  finish.
[...]

Issue #2555, opened 2026-05-18
[Steer] Tool calls continue after steer consumed & steer not preserved as distinct entry


claw-io:
# /steer Bugs and Gaps

### Description

When a user sends `/steer <text>` during mid-turn tool execution, the steer text IS
consumed and injected into the most recent tool result (by the per-tool steer drain).
However, the remaining queued tool calls in the same batch still execute before the
model ever sees the steer. The next API call — where the steer-modified messages
reach the LLM — only happens after `_execute_tool_calls()` fully returns.

The user-visible behavior: the steer sits "until the entire turn ends" instead of
being acted upon "after the next tool call."

#### Sequential path (`execute_tool_calls_sequential` in `agent/tool_executor.py`)

After each tool call, the per-tool steer drain runs:

```python
self._apply_pending_steer_to_tool_results(messages, 1)
```

This consumes any pending steer and appends it to the last tool result's content
(e.g. `\n\nUser guidance: use Python instead`). **But the loop continues** — the
next tool call in the batch starts executing immediately. There is no check after
the per-tool drain that says "if steer was consumed, break out and let the model
respond."

The only exit from the loop is `_interrupt_requested` (explicit /stop or new
message), which is a different mechanism.

#### Concurrent path (`execute_tool_calls_concurrent` in `agent/tool_executor.py`)

All tool calls are submitted to a ThreadPoolExecutor simultaneously. The main
thread waits for ALL of them via `concurrent.futures.wait()`. Steer that arrives
during execution sits unconsumed in `_pending_steer` until the post-execution
result-collection loop runs — by which point all tools have already completed.

#### Why `_apply_pending_steer_to_tool_results` doesn't help

The function at `agent/agent_runtime_helpers.py:1986` returns `None`. Callers
have no way to know whether steer was actually consumed. Even if they did, the
tool execution loops (`execute_tool_calls_sequential` and
`execute_tool_calls_concurrent`) don't have a break/return path for steer.

### Affected files

| File | Function | Issue |
|------|----------|-------|
| `agent/agent_runtime_helpers.py:1986` | `apply_pending_steer_to_tool_results()` | Returns `None` — callers can't detect if steer was consumed |
| `agent/tool_executor.py` | `execute_tool_calls_sequential()` | No break after per-tool steer drain; remaining tools still run |
| `agent/tool_executor.py` | `execute_tool_calls_concurrent()` | Wait loop doesn't check for pending steer; no way to cancel remaining futures |

---

[...]

PR #4577, opened 2026-06-21
fix: keep failed steer from cancelling active runs

dso2ng:
## Thinking Path

- WebUI already has a pending-intent RFC (#3058) that distinguishes Queue, Steer, Stop, and Stop-and-send.
- A failed `/api/chat/steer` currently returns `accepted:false`, but the browser treats that as permission to queue the text and call `cancelStream()`.
- That makes an unavailable steer destructive: the active run is cancelled even though the user requested mid-run guidance, not stop-and-send.
- This PR keeps failed steer non-destructive: restore the draft, preserve staged files, and leave the active stream running.

## What Changed

- `_trySteer()` now returns `true` only when the steer endpoint accepts delivery.
- On `accepted:false` or request failure, `_trySteer()` restores the composer draft and does not queue/cancel.
- Busy-mode steer only clears staged files after delivered steer.
- Updated copy/i18n from “queued/interrupted” to “draft restored”.
- Updated regression tests around real steer and busy input mode.

## Why It Matters

- Prevents accidental cancellation of a running turn when steer is unavailable.
- Aligns WebUI behavior with #3058’s pending-intent model: Steer belongs to the active run and failure is not implicit Queue/Interrupt.
- Keeps explicit Queue and Interrupt available when the user actually wants those actions.

## Contract Routing

Task type: pending-intent / busy-composer behavior fix

Touched areas:
- `static/commands.js`
- `static/messages.js`
- `/api/chat/steer` comments
- i18n/settings copy
- steer/busy-mode regression tests

Relevant public docs:
- `docs/CONTRACTS.md`
- `docs/rfcs/webui-pending-intent-controls.md`

Scope boundaries:
- Does not implement durable accepted-steer timeline rendering.
- Does not fix agent-side steer consumption/persistence tracked by #2555 / NousResearch/hermes-agent#28172.
- Does not remove the legacy `interrupt` busy mode or redesign the broader settings model.

## Verification

- `./scripts/test.sh tests/test_real_steer.py tests/test_1062_busy_input_modes.py tests/test_chinese_locale.py tests/test_zh_hant_locale.py`
  - `59 passed`
- `git diff --cached --check`
- `.venv/bin/python -m py_compile api/streaming.py tests/test_real_steer.py tests/test_1062_busy_input_modes.py`

## Risks / Follow-ups

- A user with busy input mode set to Steer now sees draft restored instead of implicit cancel-and-send when steer is unavailable.
- This is intentional and matches #3058, but broader settings migration / stop-and-send UI remains a future slice.
- #2555 remains the follow-up for accepted steer visibility and agent-side steer metadata/consumption.

## Model Used

AI-assisted with Hermes Agent / OpenAI Codex runtime.


nesquena-hermes:
Pulled the branch (`b2fb5745`) and read the full diff against `origin/master` — the core behavioral change is sound and the regression coverage is properly inverted. This reads as ready; one UX nuance worth a maintainer's confirmation below.

## What the fix does

The destructive path is cleanly removed from `_trySteer` (`static/commands.js`). On `accepted:false` (or a network error) it now restores the draft and returns `false` instead of queue+cancelling:

```js
// Do not fall back to interrupt: Steer failure is not permission to cancel
// the active run. Restore the draft so the user can explicitly Queue or
// Interrupt if that is what they want next. Pending files remain staged.
const inp=$('msg');
if(inp){ inp.value=explicitSteer?`/steer ${msg}`:msg; if(typeof autoResize==='function')autoResize(); }
if(typeof renderTray==='function')renderTray();
...
return false;
```

The busy-mode caller in `send()` (`static/messages.js:1008-1019`) now gates the file-clear on delivery, which is the right ordering — input is cleared *before* the await (immediate feedback), draft is restored inside `_trySteer` on failure, and staged files survive:

```js
const _steerDelivered=await _trySteer(text, /*explicitSteer=*/false);
if(_steerDelivered){S.pendingFiles=[];renderTray();}
```

The busy block ends with `return;`, so a failed steer leaves the active stream completely untouched — exactly the #3058 pending-intent contract where Steer ≠ implicit Queue/Interrupt/Stop-and-send.

## Server side is comment-only

`api/streaming.py:_handle_chat_steer` already returned `{"accepted": false, "fallback": "<reason>"}`; the diff there is just the docstring/inline notes updated to describe the new frontend contract (`api/streaming.py:8737`, `:8774`). No server logic changed — good, this keeps the fix isolated to the client.

## Tests correctly inverted

`tests/test_real_steer.py:test_try_steer_handles_fallback_without_cancelling` now asserts the absence of the destructive calls and the presence of the restore:

```python
assert "queueSessionMessage" not in body
assert "cancelStream" not in body, "fallback path must not cancel the stream"
assert "inp.value" in body, "fallback path must restore the composer draft"
```

and `tests/test_1062_busy_input_modes.py` adds `test_send_busy_steer_preserves_files_when_steer_not_delivered`, which pins the `_steerDelivered`-gated clear ordering. CI is green across all 11 jobs.

## One UX nuance for the maintainer

Previously the `stream_dead`/`not_running` fallbacks in busy-steer mode showed `busy_interrupt_confirm` and actually re-queued the turn. Now every non-accept (including `stream_dead`, where the UI still believes `S.activeStreamId` is live but the agent has gone) just restores the draft. That's consistent with the RFC, and the 3500ms `busy_steer_fallback` toast informs the user — but it does mean a genuinely-dead stream no longer auto-resends; the user must re-submit manually. That's a reasonable, intentional tradeoff; flagging it only so it's a conscious call rather than a silent regression for anyone who relied on the old auto-resend. Everything else LGTM.

PR #4578, opened 2026-06-21
Fix mid-stream transcript scroll anchor

### LINKED ISSUE #4295: bug(chat): viewport jumps upward while reading an in-progress streamed reply

franksong2702:
## Summary

- Preserve the reader's semantic viewport anchor during mid-stream transcript rebuilds before falling back to raw `scrollTop`.
- Remount virtualized message anchors before restoring them, and keep `_programmaticScroll` scoped to the remount frame.
- Treat captured `userUnpinned` / `pinned` snapshot state as authoritative so distance-based inference cannot re-pin a reader who manually scrolled away during a live reply.

Fixes #4295.

## Contract / State Layer

This touches browser-only transcript scroll state: `_messageUserUnpinned`, `_scrollPinned`, `_nearBottomCount`, `_programmaticScroll`, and `_messageVirtualWindowKey`. It does not change persisted session state, transcript storage, or backend stream state.

## Verification

- `node --check static/ui.js`
- `./.venv/bin/ruff check tests/test_issue4295_midstream_scroll_anchor.py tests/test_issue500_message_list_virtualization.py`
- `git diff --check`
- `./scripts/test.sh tests/test_issue500_message_list_virtualization.py tests/test_issue3479_ios_stream_scroll_jump.py tests/test_issue1360_streaming_scroll_hardening.py tests/test_issue1690_scroll_completion.py tests/test_issue3545_streaming_scroll_cue.py tests/test_issue4295_midstream_scroll_anchor.py` (`55 passed`)
- Claude Code review, final pass: no blockers before the CI harness fix; the later CI failure was the existing #500 Node harness missing the new shared helper, and that harness is now updated and locally covered.

## Notes / Risk

The new regression includes a Node harness that executes the real scroll restore helpers with a fake messages container, covering virtual-anchor remount, raw-top fallback avoidance, explicit unpinned-state authority, programmatic-scroll cleanup, and intermediate bottom-distance preservation. The existing #500 virtualization harness is also updated to extract the new shared remount helper. I did not restart or exercise the local `8787` runtime in this PR workflow.

## Model Used

OpenAI GPT-5 Codex coordinated the PR; a Codex sub-agent implemented the first pass; Claude Code reviewed the final diff.


franksong2702:
Follow-up pushed in `cb2953480`.

- Addressed Greptile's changelog note by dropping the direct `CHANGELOG.md` entry and leaving release notes to the release flow.
- Local checks passed: `node --check static/ui.js`, `./scripts/test.sh tests/test_issue4295_midstream_scroll_anchor.py -q`, `./scripts/test.sh tests/test_issue500_message_list_virtualization.py -q`, `git diff --check`.

franksong2702:
Follow-up pushed in `9071b89a5`.

- Merged latest `origin/master`; PR is still mergeable/clean.
- Local checks passed: `node --check static/ui.js`, `./scripts/test.sh tests/test_issue4295_midstream_scroll_anchor.py -q`, `./scripts/test.sh tests/test_issue500_message_list_virtualization.py -q`, and `git diff --check`.

franksong2702:
Follow-up pushed in `831dfd198`.

- Merged latest `origin/master`; PR is now mergeable/clean again.
- Local checks passed: `node --check static/ui.js`, `./scripts/test.sh tests/test_issue4295_midstream_scroll_anchor.py -q`, `./scripts/test.sh tests/test_issue500_message_list_virtualization.py -q`, and `git diff --check`.
- GitHub checks are all green on the refreshed head.

franksong2702:
Follow-up pushed in `50bc73224`.

- Merged latest `origin/master` after the 10-minute follow-up found the PR had become `BEHIND` again.
- Local checks passed again: `node --check static/ui.js`, `./scripts/test.sh tests/test_issue4295_midstream_scroll_anchor.py -q`, `./scripts/test.sh tests/test_issue500_message_list_virtualization.py -q`, and `git diff --check`.
- GitHub checks are all green on the refreshed head, and the PR is `MERGEABLE` / `CLEAN`.

nesquena-hermes:
Shipped as **v0.51.554** (Release TM) — live in prod. Closes #4295. 📜

Gate: full suite 9930 passed, Codex SAFE ("no regression risk, core flows intact"), Opus SAFE/SHIP (disciplined refactor — _programmaticScroll scoping held correctly across the synchronous remount frame with no leak, pinned readers still auto-follow, unpin escape path intact, virtualization interaction correct). Scrolling up during a live reply no longer snaps you back to bottom. Thanks @franksong2702.

Issue #3450, opened 2026-06-02
[security] Extend symlink-swap TOCTOU hardening to api/routes.py file-API endpoints (raw/zip/editor) + upload cleanup


nesquena-hermes:
## Summary

A security review of #3398 (workspace symlink-escape containment, shipped v0.51.221) surfaced a **pre-existing** TOCTOU (time-of-check-to-time-of-use) class affecting the **`api/routes.py` file-API endpoints** that #3398 did not touch. These endpoints validate a path with `api.helpers.safe_resolve()` and then re-open / read / write / delete by **pathname**, so a path component swapped to an external symlink *between* validation and use can escape the workspace.

This is the same race class #3398 closed for the *workspace-tree* read/list/upload/extract paths (via the new portable anchored `openat`-walk helpers `open_anchored_fd` / `open_anchored_create_fd` / `make_anchored_dir` in `api/workspace.py`). This issue tracks extending that hardening to the remaining `routes.py` file-API surface.

## Why it was scoped separately from #3398

- #3398's diff is confined to `api/workspace.py` + `api/upload.py` (+ tests). It never modified `api/routes.py`.
- These endpoints use a **different** validator (`api.helpers.safe_resolve`), not the `api.workspace.safe_resolve_ws` that #3398 hardened.
- The race is **pre-existing in master** — it predates #3398 and was not introduced or widened by it. #3398 is a net security improvement (it closes the no-race static symlink-escape exfil path and the workspace-tree TOCTOU); this is the remaining, lower-likelihood surface.

## Threat model / likelihood

TOCTOU here requires an attacker to win a sub-millisecond swap race against the server's own file API. The **higher-likelihood, no-race** exfiltration path (an in-workspace symlink to `~/.ssh` / `~/.hermes/auth.json` read through `/api/list` / `read_file_content`) is **already closed** by #3398. This issue is the race-only residue on the broader endpoint set.

## Endpoints to harden (from the review)

1. **`/api/file/raw`** (`routes.py` ~8277) + inline HTML preview (`_serve_inline_html_preview`, ~8494) — validates via `_file_raw_target()` then `target.open("rb")` / `target.read_bytes()` by pathname. Fix: read through `open_anchored_fd(Path(s.workspace), target, want_dir=False)` and stream/read from the fd.
[...]

PR #4581, opened 2026-06-21
feat(workspace-tree): surface escape-target symlinks as display-only rows

### LINKED ISSUE #4510: feat(workspace-tree): surface escape-target symlinks as a distinct, confirmation-gated row
## Problem

The workspace file tree currently **silently drops** any symlink whose
resolved target sits outside the registered workspace root. This is the
escape filter at `api/workspace.py:list_dir._process` — the containment
primitive introduced in v0.51.221 (#3398) and hardened in v0.51.274
(#3450/#3630).

The security behavior is correct: dropping a symlink whose target
escapes `ws_resolved` is the right default for `read`/`delete`/`rename`
operations. But it makes the workspace tree silently invisible for
developer-side `.dev` symlinks that point **outside** the workspace
root — for example the workspace-level `.dev -> /opt/data/.../state/<repo>/<topic>`
convention. The user sees no symlink row in any of their registered
repo workspaces; only the Home workspace (where `patches -> state` is a
relative path that stays inside the boundary) shows the link.

## Goal

Surface escape-target symlinks as a distinct row kind in the file tree
with a clear visual signal, so the user can see the link exists and
opt-in to navigating through it via a confirmation step. The escape
filter itself is preserved as the default; the renderer is given a way
to display the row, and clicking the row prompts the user before any
cross-workspace navigation happens.

### `api/workspace.py:list_dir._process()`
Instead of silently dropping a symlink whose resolved target escapes
`ws_resolved`, emit it with a new flag `target_outside_workspace: true`
and the absolute `target` string. The existing in-workspace symlinks
still emit `type: 'symlink'` with `is_dir` / `target` exactly as before;
the cycle and blocked-system filters (which prevent genuinely dangerous
targets like `/etc`, `/usr`) remain in place. **The escape filter is
widened, not removed.**

### `static/ui.js`
Add two branches in `_renderTreeItems`:
- `type === 'symlink' && !target_outside_workspace` — in-workspace link;
  mirror the real-dir affordances (chevron if is_dir, expand on click,
  dblclick to navigate, drag bindings, the existing tooltip path elision).
- `type === 'symlink' && target_outside_workspace` — distinct icon
  (`external-link` from `static/icons.js` `LI_PATHS`), **no chevron, no
  expand, no dblclick navigate, no drag bindings, no rename input, no
  context-menu rename**. Single-click shows a confirmation dialog
  (`showConfirmDialog({title, message, confirmLabel, danger, focusCancel})`)
  with the resolved absolute target; on confirm, the renderer opens
  the file at the link path via `openFile(item.path)` (the link
  resolves on the server, which is permitted because the user
  explicitly opted in).

### `static/icons.js`
Add an `external-link` entry to `LI_PATHS` (Lucide `external-link`, 24×24).
Distinct from `link` so the user can tell at a glance which kind of
link they're looking at.

### `static/i18n.js`
Add a single key `external_link_open_confirm` for the dialog text.
English default; `// TODO: translate` for the other 12 locales
(matches the convention used in the symlink-rows FR).

## Out of scope

- Changing the containment primitive itself (the escape filter, the
  blocked-system filter, the cycle filter). Those are at v0.51.274 and
  are not part of this change.
- Rename / delete operations on escape-target symlinks. Per the FR, the
  renderer does not bind rename or delete for the `target_outside_workspace`
  case; that would require a separate backend authorization question.
- Migration of `link_target.relative_to(ws_resolved)` callers in
  sibling handlers. The change is scoped to `list_dir` payload only.

## Tests

`tests/test_workspace_symlink_containment.py` — three new tests:
1. `test_list_dir_in_workspace_symlink_shape` — pins the existing
   in-workspace symlink payload (`type='symlink'`, `is_dir`, `target`,
   no `target_outside_workspace` flag).
2. `test_list_dir_outside_workspace_symlink_emitted_with_flag` — pins
   the new payload: a symlink whose resolved target sits outside the
   workspace is now emitted with `type='symlink'`, `is_dir=False`,
   `target=<absolute>`, `target_outside_workspace=True`.
3. `test_list_dir_external_symlink_target_blocked_unchanged` — pins
   the safety floor: a symlink into a blocked system path (`/etc`,
   `/usr`) is still dropped, just as today. Guards against a future
   refactor accidentally widening the escape filter past the
   blocked-system boundary.

## Related

- #4226 — symlink display in workspace tree (renderer-only change, this
  issue is a renderer-only change that depends on the wire format
  payload from #4226)
- #3398 — original escape filter introduction
- #3450 / #3630 — TOCTOU hardening of the filter

claw-io:
## Thinking Path

- Hermes WebUI's workspace tree uses a containment filter to prevent symlink-based path traversal (#3398, #3450, #3630)
- The filter silently drops any symlink whose resolved target sits outside the workspace root — correct for security, but it makes developer-side `.dev` symlinks invisible in the tree
- Issue #4510 proposed surfacing these as distinct rows with a confirmation-gated navigation path
- Maintainer feedback ([comment](https://github.com/nesquena/hermes-webui/issues/4510#issuecomment-4760723569)) identified a load-bearing gap: the FR's "confirm then navigate" flow would 400 every time because `safe_resolve_ws` has no "user opted in" concept — the gate is purely structural and shared by both UI and agent/tool calls
- This PR implements the maintainer's recommended Scope 1 (display-only): surface the row so the user can see the link exists, show the absolute target on click, but do **not** call `openFile` — the read gate stays intact
- Phase 2 (real opt-in navigation via `allow_escape` threading) is scoped as a separate follow-up issue

### `api/workspace.py`

- **`list_dir._process`**: Instead of silently `return`-ing when a symlink's resolved target escapes `ws_resolved`, now sets `target_outside_workspace = True` and emits the entry. The blocked-system-path filter (`/etc`, `/usr`) stays in place and still drops those entries. The inline comment is updated from "can never be opened" to reflect the new display-only behavior.
- **`dir_signature`**: Added `target_outside_workspace` to the signature payload so directory state changes (including external symlinks appearing/disappearing) are detected.
- **`is_dir`**: Kept dynamic (`link_target.is_dir()`), not forced to `False` as the FR proposed — the actual directory status is more useful for display and a future Phase 2 may use it.

### `static/icons.js`

- Added `external-link` entry to `LI_PATHS` (Lucide `external-link`, 24×24). Distinct from `link` so the user can tell at a glance which kind of symlink they're looking at.

### `static/ui.js` — `_renderTreeItems`

- Added `isExternalLink` flag: `isLk && item.target_outside_workspace`
- External symlinks are **not** dir-like (no expand chevron) and **not** file-like (no `openFile`)
- **Icon**: `li('external-link', 14)` instead of `li('link', 14)`
- **Not draggable**: `draggable` attribute removed, `ondragstart` nulled
- **No delete button**: both `isDirLike` and `isFileLike` are false, so neither delete branch renders
- **Click handler**: shows `showConfirmDialog` with the resolved absolute target — informational only, no `openFile` call
- **Double-click**: delegates to `el.onclick` (shows the dialog), not rename

### `static/i18n.js`

- Added `external_link_open_confirm` key to all 13 locales. English default; other locales use the English text as fallback (translators can update later).

### `tests/test_workspace_symlink_containment.py`

- **Updated** `test_safe_resolve_blocks_external_symlink_directory`: now asserts "escape" IS in entries with `target_outside_workspace=True` and `type='symlink'` (was: `assert "escape" not in`)
- **Updated** `test_read_file_blocks_external_symlink_file`: same pattern — entry is emitted with the flag, but `read_file_content` still raises `ValueError`
- **Updated** `test_list_read_create_work_on_no_dir_fd_fallback`: asserts "escape" is now in names with `target_outside_workspace=True` (was: `assert "escape" not in names`)
- **New** `test_list_dir_in_workspace_symlink_shape`: in-workspace symlinks have `target_outside_workspace=False`
- **New** `test_list_dir_outside_workspace_symlink_emitted_with_flag`: escape symlinks emitted with correct `type`, `target`, `is_dir`, `target_outside_workspace=True`
- **New** `test_list_dir_external_symlink_blocked_system_path_unchanged`: symlinks to `/etc` and `/usr` are still filtered out
- **New** `test_list_dir_escape_symlink_read_still_blocked`: paired read-path test — listing shows the entry but `read_file_content` on the same target still raises `ValueError` (proving the gate is intact)

### `CHANGELOG.md`

- Added entry under `[Unreleased]`

## Why It Matters

Developer workspaces use `.dev -> /opt/data/.../state/<repo>/<topic>` symlinks that point outside the workspace root. These were silently invisible in the tree — the user had no way to know the link existed without checking the filesystem. Now they appear with a distinct icon and show their target on click, with zero security surface change.

## Verification

- `./scripts/test.sh tests/test_workspace_symlink_containment.py` — **15 passed** (11 existing updated + 4 new)
- The read gate (`safe_resolve_ws` / `open_anchored_fd`) is untouched — no new params, no new routes, no new backend surface
- `grep -n "Path traversal blocked" api/workspace.py` confirms all 10 traversal-block sites are unchanged
- Blocked system paths (`/etc`, `/usr`) are still filtered by `_is_blocked_system_path` before the entry is emitted

**Before/after screenshots will be added before merge.** Visual change: external symlinks that were previously invisible now appear as rows with an `external-link` icon (arrow pointing out of a box) instead of the `link` icon (chain). Clicking shows a dialog with the resolved absolute target path and an informational message.

## Risks / Follow-ups

- **No security risk**: the read/list gate is unchanged. The `target_outside_workspace` flag is display metadata only — it cannot be used to bypass `safe_resolve_ws`. The payload extension is additive (existing consumers that don't check the flag are unaffected).
- **i18n**: The `external_link_open_confirm` key uses English text in all 13 locales as fallback. Translators can update non-English locales.
- **Phase 2 follow-up**: A separate issue will scope the backend authorization path for real opt-in navigation through external symlinks (`allow_escape: bool` threading through `list_dir` → read route → `read_file_content` / `open_anchored_fd`, gated to UI-origin requests only).
- **`is_dir` is dynamic**: The FR proposed forcing `is_dir=False`, but the current code computes `is_dir = link_target.is_dir()`. For display-only this doesn't matter (no expand, no open), but it's more accurate and a future Phase 2 may use it.

## Model Used

glm-5.2 (custom provider)

Closes #4510


nesquena-hermes:
Shipped as **v0.51.558** (Release TQ) — live in prod. Escape-target symlinks now show as display-only rows. 9944 suite + Codex/Opus (containment intact). Info-leak hardening applied during review (no resolved-path/target-metadata disclosure; generic dialog across 13 locales) + a dialog Cancel-button fix. Thanks @claw-io.

Issue #4539, opened 2026-06-20
bug(transparent-stream): tool-call rows vanish on turn settle, reappear only after tab/session switch


nesquena-hermes:
## Summary

In **Transparent Stream** activity mode, tool-call rows render correctly *during* streaming, but the moment the stream **settles** (turn ends) every tool row disappears — leaving only the Thinking blocks behind. Switching to another browser tab/window (or another session) and coming back makes the tool rows **reappear**.

> _Reported by **Latipun** in Discord #report-bugs (2026-06-20):_
> "I'm using **'Transparent Stream'** option. When in chat stream, I see all tool calls, but when stream end, all tool call disappear, only leaving the 'Thinking' blocks. I need to change tab or window, all tools block come back."

## Severity

Medium-high reliability bug on a recently-shipped, prominent feature (Transparent Stream, #3820 / #4024, #4096). The data is **not** lost — the persisted transcript is intact — but the immediate post-turn view silently swallows the entire tool trace, which is the whole point of the mode. The user has to perform an unrelated action (tab/session switch) to recover the view.

## Steps to reproduce

1. Settings → Activity display → **Transparent Stream**.
2. Send a message that triggers tool calls (e.g. a search/terminal/file turn).
3. Watch the tool-call rows appear live as they run. ✅
4. Let the turn finish (stream settles).
5. **Bug:** all tool rows vanish; only the Thinking rows remain in the turn.
6. Switch to another browser tab/window (or another session) and back → the tool rows reappear. ✅ (confirms the data is intact; only the settled re-render is wrong.)

## Root cause

The bug is in the **settled rebuild** of the transparent-stream branch in `renderMessages()` (`static/ui.js`). Two compounding gaps:

### Gap 1 — the settled-node cleanup removes transparent tool rows but preserves transparent thinking rows (explains "tools vanish, Thinking stays")

At the top of the settled tool/worklog rebuild block (`static/ui.js` ~L10783):

```js
inner.querySelectorAll(
  '.tool-worklog-group:not([data-compression-card]),'+
  '.tool-call-group:not([data-compression-card]),'+
  '.tool-card-row:not([data-compression-card]),'+
  '.agent-activity-thinking:not([data-live-thinking="1"]):not([data-event-type="thinking"]),'+
  '.wl-reason[data-worklog-anchor-reason="1"],'+
  '.wl-reason[data-worklog-reason-source="reasoning"]'
).forEach(el=>el.remove());
```

[...]

PR #4587, opened 2026-06-21
fix(transparent-stream): render persisted anchor activity on reload + keep intermediate prose (#4568)

### LINKED ISSUE #4568: bug(transparent-stream): tool-call rows vanish after page reload — persisted anchor activity scene is compact-only (sibling of #4539)

nesquena-hermes:
## Summary

Fixes #4568. In **Transparent Stream** activity mode, a turn that settled with a persisted assistant-turn anchor scene (`activity_scene_v1`, #3926/#4411) lost **all** its tool-call and thinking activity on every rehydrate path — after the turn settled, when switching away and back, on tab hide/show, on window blur/focus, on a hard reload, and on a mid-stream reload — leaving only the final answer prose.

**Root cause:** `renderMessages()` marks anchor-scene-carrying turns "anchor-owned" and skips the legacy derived-tool rebuild, but the settled anchor-scene renderer `_renderSettledAnchorSceneForMessage` only ran in Compact Worklog mode (early `return false` otherwise). So in Transparent Stream the activity rendered nowhere on settle/reload.

## Approach — combines two solutions

This builds directly on **@franksong2702's #4576**, which introduced the right architecture: render the persisted anchor scene **as transparent event rows** on settle/reload (deterministic, and #3926-aligned — both display modes consume the *same* persisted scene instead of a second normalizer). I verified that approach reliably restores the **tool and thinking rows**.

It adds **intermediate-prose preservation** on top: #4576 dropped *all* `prose` rows to avoid duplicating the final answer, which also lost the between-tool progress narration from earlier rounds. `_anchorSceneTransparentNodeForRow` now renders intermediate prose as inline `assistant-segment` rows and suppresses **only** the prose row that duplicates the final answer (`_anchorSceneProseMatchesFinalAnswer` — exact normalized match, with a length-ratio-guarded prefix tolerance for re-wraps). Result: tools + thinking + intermediate prose all survive, interleaved in chronological order, at every checkpoint, in both modes.

Compact Worklog rendering is unchanged.

## Verification

Verified with a **checkpoint-matrix harness** (in the private workspace) that drives a real multi-round streaming turn (12 tool calls across 3 rounds + reasoning + progress prose + final answer) through every transcript lifecycle checkpoint, in **both** display modes:

| checkpoint | master | this PR |
|---|---|---|
| during_stream | ✅ | ✅ |
| after_done | ❌ tools→0 | ✅ |
| switch_away_back | ❌ tools→0 | ✅ |
| tab_hidden_visible | ❌ tools→0 | ✅ |
| blur_focus | ❌ tools→0 | ✅ |
| reload_after_done | ❌ tools→0 | ✅ |
| reload_mid_stream | ❌ tools→0 | ✅ |

- **Transparent Stream:** all 7 checkpoints render 12 tool rows + thinking + 4 prose rows, interleaved.
- **Compact Worklog:** all 7 checkpoints render the folded worklog group + final answer (no regression).
- **Full pytest suite:** 9881 passed, 0 failed.
- **Codex + Opus** both reviewed the diff: no XSS/double-render/orphan-row/compact regression; the one finding (over-broad final-answer prefix match) was fixed with a length-ratio guard, with a regression test.

Regression coverage added in `tests/test_live_to_final_anchor_visible_order.py` (the settled-scene transparent renderer route + intermediate-prose-preserved / final-answer-suppressed invariants).

## Credit

Co-authored with @franksong2702 (#4576 persisted-anchor-scene rendering approach). Supersedes #4576.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Transparent Stream keeps its full tool / thinking / progress-prose activity after reload, tab/focus changes, and session re-entry (#4568).** A turn that settled with the stable assistant-turn anchor scene (`activity_scene_v1`) could lose its Transparent Stream activity rows on a hard refresh, when switching away and back, on tab hide/show or window blur/focus, and on a mid-stream reload — leaving only the final answer. The legacy Transparent Stream rebuild correctly skipped anchor-owned turns, but only Compact Worklog had a settled anchor-scene renderer, so in Transparent Stream the rows rendered nowhere. Transparent Stream now renders the persisted anchor activity rows directly: tool cards and thinking rows are restored, and the intermediate between-tool progress prose is kept interleaved in chronological order (only the prose row that duplicates the final answer — already owned by the assistant message segment — is suppressed). Verified across the live → settled → switch-away → tab-hidden → blur → reload → mid-stream-reload checkpoint matrix in both display modes. Thanks @franksong2702 for the persisted-anchor-scene rendering approach.
+
 ## [v0.51.548] — 2026-06-21 — Release TG (extension load diagnostics)
 
 ### Added
--- a/static/ui.js
+++ b/static/ui.js
@@ -9130,15 +9130,23 @@ function _anchorSceneTransparentNodeForRow(row, opts){
   node.setAttribute('data-anchor-source-event-type',String(row.source_event_type||''));
   return node;
 }
-// Whitespace-insensitive compare so a scene prose row that is the final answer
+// Whitespace-insensitive compare so a scene prose row that IS the final answer
 // (possibly re-wrapped) is recognized and not duplicated against the segment.
+// Codex #4568: the prefix tolerance must NOT be able to suppress a DISTINCT
+// intermediate progress row that merely happens to be a prefix of the final
+// answer (e.g. "I found the issue and I'm applying the fix now." when the final
+// answer starts with that sentence). So require exact normalized equality, with
+// a prefix tolerance allowed ONLY when the two strings are near-equal length
+// (>=0.9 ratio, shorter >=80 chars) — i.e. genuine re-wrap/truncation of the
+// SAME text, never a short intermediate sentence that prefixes a long answer.
 function _anchorSceneProseMatchesFinalAnswer(proseText, finalAnswer){
   const norm=(s)=>String(s||'').replace(/\s+/g,' ').trim();
   const a=norm(proseText), b=norm(finalAnswer);
   if(!a||!b) return false;
   if(a===b) return true;
-  // tolerate trailing punctuation / minor truncation differences
-  return a.length>=40 && b.length>=40 && (a.startsWith(b)||b.startsWith(a));
+  if(!(a.startsWith(b)||b.startsWith(a))) return false;
+  const shorter=Math.min(a.length,b.length), longer=Math.max(a.length,b.length);
+  return shorter>=80 && (shorter/longer)>=0.9;
 }
 function _anchorSceneWorklogGroup(blocks, opts){
   if(!blocks) return null;
--- a/static/ui.js
+++ b/static/ui.js
@@ -9071,6 +9071,75 @@ function _anchorSceneNodeForRow(row, opts){
   node.setAttribute('data-anchor-source-event-type',String(row.source_event_type||''));
   return node;
 }
+function _anchorSceneTransparentNodeForRow(row, opts){
+  const settled=!!(opts&&opts.settled);
+  if(!row) return null;
+  let node=null;
+  const meta={
+    segmentSeq:row.segment_seq||row.segmentSeq||'',
+    burstId:row.activity_burst_id||row.burst_id||row.burstId||'',
+  };
+  if(row.role==='prose'){
+    // The settled assistant segment already owns the FINAL answer prose, so a
+    // prose row whose text matches the final answer must be suppressed here to
+    // avoid duplicating the answer. But INTERMEDIATE progress prose (the
+    // between-tool narration from earlier rounds) is NOT the final answer and
+    // belongs in the chronological transparent history — dropping it loses the
+    // interleaving the user saw live (#4568: tools were restored but mid-turn
+    // prose still vanished on reload). Render intermediate prose as an inline
+    // assistant-segment (same shape _anchorSceneNodeForRow builds), skip only
+    // the final-answer duplicate.
+    const text=String(row.text||'').trim();
+    if(!text) return null;
+    const finalAnswer=String((opts&&opts.finalAnswer)||'').trim();
+    if(finalAnswer&&_anchorSceneProseMatchesFinalAnswer(text,finalAnswer)) return null;
+    node=_anchorSceneNodeForRow(row,{settled});
+    if(!node) return null;
+    node=_decorateTransparentEventRow(node,{type:'prose',text,preview:text,...meta});
+  }else if(row.role==='thinking'){
+    if(window._showThinking===false) return null;
+    const text=String(row.text||row.thinking&&row.thinking.text||'').trim();
+    if(!text) return null;
+    node=_decorateTransparentEventRow(_thinkingActivityNode(text,false,row.row_id||row.local_id||'anchor-thinking'),{
+      type:'thinking',
+      text,
+      preview:text,
+      ...meta,
+    });
+  }else if(row.role==='tool'){
+    const toolCall=_anchorSceneToolCallFromRow(row,{settled});
+    node=_decorateTransparentEventRow(buildToolCard(toolCall),{
+      type:'tool',
+      name:toolCall&&toolCall.name,
+      status:_transparentToolStatus(toolCall,true),
+      toolCall,
+      ...meta,
+    });
+  }else{
+    node=_anchorSceneNodeForRow(row,{settled});
+    node=_decorateTransparentEventRow(node,{
+      type:String(row.role||'activity'),
+      ...meta,
+    });
+  }
+  if(!node) return null;
+  node.setAttribute('data-anchor-scene-row','1');
+  node.setAttribute('data-anchor-settled-scene-row','1');
+  node.setAttribute('data-anchor-row-id',String(row.row_id||row.local_id||''));
+  node.setAttribute('data-anchor-row-role',String(row.role||'activity'));
+  node.setAttribute('data-anchor-source-event-type',String(row.source_event_type||''));
+  return node;
+}
+// Whitespace-insensitive compare so a scene prose row that is the final answer
+// (possibly re-wrapped) is recognized and not duplicated against the segment.
+function _anchorSceneProseMatchesFinalAnswer(proseText, finalAnswer){
+  const norm=(s)=>String(s||'').replace(/\s+/g,' ').trim();
+  const a=norm(proseText), b=norm(finalAnswer);
+  if(!a||!b) return false;
+  if(a===b) return true;
+  // tolerate trailing punctuation / minor truncation differences
+  return a.length>=40 && b.length>=40 && (a.startsWith(b)||b.startsWith(a));
+}
 function _anchorSceneWorklogGroup(blocks, opts){
   if(!blocks) return null;
   const live=!!(opts&&opts.live);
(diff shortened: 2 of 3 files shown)


nesquena (approved):

Independent review of the #4568 fix: in **Transparent Stream** mode, a turn that settled with a persisted assistant-turn anchor scene (`activity_scene_v1`, #3926/#4411) lost **all** its tool/thinking/progress-prose activity on every rehydrate path (settle, switch-away-and-back, tab hide/show, blur/focus, hard reload, mid-stream reload) — leaving only the final answer. The reload/rehydrate sibling of #4539.

### What this ships

`static/ui.js` (+117), `tests/test_live_to_final_anchor_visible_order.py` (+52), `CHANGELOG.md` (+4). Agent-authored on the maintainer branch, co-authored with @franksong2702 (builds on #4576). Closes #4568.

Fresh `nousresearch/hermes-agent` tarball: `activity_scene_v1` / `_anchor_activity_scene` / `anchor_activity_scenes` are **absent** — the anchor scene is WebUI-only session metadata (persisted by `api/routes.py`/`api/models.py`, consumed only by the WebUI transcript renderer). The agent never reads it. Pure presentation-layer rehydrate bug; no cross-tool surface. ✓

### Root cause + fix — traced in full

The anchor-scene system marks a settled, scene-carrying turn "anchor-owned" so the legacy derived-tool rebuild and the Transparent Stream rebuild both **skip** it ([ui.js ~11106/11355](static/ui.js)). The only renderer that would draw the persisted scene, `_renderSettledAnchorSceneForMessage`, had a hard `!isCompactWorklogMode() → return false` gate — so in Transparent Stream the activity rendered **nowhere** on settle/reload.

The fix dispatches at the top of `_renderSettledAnchorSceneForMessage` ([ui.js:9333](static/ui.js#L9333)): `isTransparentStream()` → new `_renderSettledAnchorSceneTransparentForMessage`; otherwise the **byte-unchanged** compact path. This is the render-side approach (the more robust of the issue's two options): it draws the persisted scene in whatever the **current** display mode is, so it also fixes the cross-mode case (a turn settled in compact, later viewed in transparent) the issue flagged as a dead-end.

`_renderSettledAnchorSceneTransparentForMessage`:
- extracts rows via the shared `_anchorSceneRowsForRendering(scene,{settled:true})`;
- computes `finalAnswer` (`scene.final_answer` → `_assistantAnchorSceneFinalAnswerText(message)` → `msgContent(message)`, all guarded);
- removes stale settled-scene rows (idempotency on repeated renders) and hides earlier `.assistant-segment` nodes (`idx<rawIdx` → `assistant-segment-worklog-source` + `aria-hidden`) — **identical** to the compact renderer's segment-hiding;
- builds each row via `_anchorSceneTransparentNodeForRow` and inserts it **before** the final-answer segment in chronological order;
- `_syncTransparentEventControls(turn)` after writing.

`_anchorSceneTransparentNodeForRow` routes by role: prose → `_anchorSceneNodeForRow` (final-answer duplicate suppressed, intermediate kept); thinking → `_thinkingActivityNode` (respects `window._showThinking===false`); tool → `buildToolCard(_anchorSceneToolCallFromRow(...))`; all wrapped by `_decorateTransparentEventRow`. These are the **same builders the live Transparent Stream path uses**, fed the same scene data — so the restored rows match what the user saw live.

### Security audit

No new sink. The only `innerHTML` in the new-reachable prose path is `_anchorSceneNodeForRow`'s `<div class="msg-body">${renderMd?renderMd(text):esc(text)}</div>` — i.e. the sanitizing markdown renderer (or `esc()` fallback), the same path all assistant prose uses. Thinking/tool rows go through the vetted live builders (`_thinkingActivityNode`, `buildToolCard`, `_transparentToolDetailHtml`) with `textContent`/`setAttribute` for names/status. The scene data has the same provenance as live SSE data already rendered through these builders — no new untrusted-data→sink path, no raw `innerHTML = sceneData`. ✓

### The one flagged risk — `_anchorSceneProseMatchesFinalAnswer` — proven well-calibrated

Greptile (and the PR's Codex+Opus pass) flagged that the prefix tolerance could wrongly suppress an intermediate progress-prose row that is a verbatim prefix of the final answer. The final guard is: exact normalized equality always matches; the `startsWith` prefix tolerance fires **only** when `shorter>=80 && shorter/longer>=0.9` — a genuine re-wrap/truncation of the same text. I proved it with a behavioural harness (**9/9**):
```
exact match                                  -> suppress  ✅
whitespace re-wrap of same text              -> suppress  ✅
short (<80) intermediate prefix of long final -> PRESERVE  ✅  (Greptile's concern)
80+ intermediate prefixing a much-longer final (ratio<0.9) -> PRESERVE ✅
near-equal truncation (ratio>=0.9, >=80)     -> suppress  ✅
distinct prose / empty                       -> PRESERVE  ✅
short exact ("Done.")                         -> suppress (exact bypasses length floor) ✅
```
So false-suppression requires an intermediate sentence ≥80 chars that is a verbatim prefix of a final answer no more than ~11% longer — a genuinely narrow window; all the common between-tool-narration shapes correctly preserve. Accepted tradeoff (the alternative, exact-only, would fail to dedupe real re-wraps).

- **Compact Worklog unchanged:** the dispatch returns before the compact body; the compact renderer is byte-identical to master. The two display modes are exhaustive (`isCompactWorklogMode()` = `isSimplifiedToolCalling() && !isTransparentStream()`, simplified always true), so exactly one path runs.
- **Idempotency:** the settled-scene cleanup (`[data-anchor-settled-scene-row="1"],.transparent-event-row[data-anchor-scene-row="1"]`) runs before re-insert, so repeated renders don't duplicate rows; `renderMessages` wipes `innerHTML` anyway, so there's no cross-render accumulation.
- **All referenced helpers exist** (`_anchorSceneRowsForRendering`, `_anchorSceneToolCallFromRow`, `_assistantAnchorSceneFinalAnswerText`, `_assistantTurnBlocks`, `_transparentToolStatus`, `_syncTransparentEventControls`) — no typos; `msgContent`/`renderMd` guarded by `typeof`.

### Edge-case matrix

| Scenario (Transparent Stream) | Behavior |
|---|---|
| Settled turn, hard reload | tool + thinking + intermediate prose restored, chronological ✅ |
| Switch away & back / tab hide-show / blur-focus | scene re-rendered each `renderMessages` ✅ |
| Mid-stream reload | persisted scene rendered (no live state needed) ✅ |
| Final-answer prose row | suppressed (owned by the assistant segment) ✅ |
| Intermediate progress prose | preserved, interleaved ✅ |
| `window._showThinking===false` | thinking rows omitted ✅ |
| Settled in compact, viewed in transparent (cross-mode) | renders in the current (transparent) mode ✅ |
| Compact Worklog mode | unchanged (folded worklog group + final answer) ✅ |
| CLI reads same session | webui-only scene; agent unaffected ✅ |

### Tests

- `tests/test_live_to_final_anchor_visible_order.py` (+2 tests) + `test_issue3820_chat_activity_display_mode` + `test_issue4539_transparent_tool_settle` — **95/95 pass**.
- **Full suite: 9768 passed / 132 skipped / 0 failures** in 198s (deselected the pre-existing darwin CRLF flake).
- **JS gates:** `node --check static/ui.js` clean; ESLint runtime guard exit 0.
- **Independent matcher harness: 9/9** (prefix-guard calibration, above).

- The two new regression tests are structural (source-text assertions). They lock the dispatch, the intermediate-prose-preserved invariant, and the matcher's length-ratio guard — good regression locks; the behavioural coverage is the PR's private checkpoint-matrix harness (master RED → fix GREEN across all 7 checkpoints in both modes), which I substituted for the matcher with the 9/9 unit harness above.
- `_anchorSceneProseMatchesFinalAnswer` checks every prose row against the final answer; in the unlikely case a long intermediate sentence is a ≥90%-length verbatim prefix of the final answer it would be suppressed. Narrow and accepted (see above); a future tightening could key off prose-row position (the final-answer echo is the last prose row) if it ever bites.

### Recommendation

The fix correctly identifies that the settled anchor-scene renderer was compact-only and adds a Transparent Stream path that replays the persisted scene through the same vetted live builders, restoring tools + thinking + intermediate prose interleaved while suppressing only the final-answer duplicate. Compact is byte-unchanged, the mode split is exhaustive, there's no new XSS surface (prose is `renderMd`-sanitized), it's webui-only, and the one flagged risk — the prose/final-answer matcher — is well-guarded (proven 9/9). Nice work by @franksong2702 on the underlying approach. Ship.

nesquena-hermes:
Shipped as **v0.51.550** (Release TI) — live in prod. 🎬 Closes #4568.

Deep-gated to the full activity-render mandatory standard:
• **Checkpoint matrix 4/4 CLEAN** — master/transparent reproduces the bug (tools→0 on every reload path), fix/transparent PASSES all 7 checkpoints (during_stream/after_done/switch_away_back/tab_hidden_visible/blur_focus/reload_after_done/reload_mid_stream — 12 tools + 3 thinks + 4 prose interleaved), master+fix/compact PASS (no regression).
• Full suite 9913 passed · Codex SAFE · Opus SAFE.

One hardening fix applied during review (Codex SILENT finding): the scene-builder/hydrator final-answer filters (messages.js + routes.py) used a looser >=80+prefix rule than the renderer (>=0.9 ratio), so a short intermediate-prose prefix of the final answer could be dropped from the persisted scene with no fallback. Aligned all three filters to the same >=0.9 rule; re-ran the matrix → still GREEN 4/4.

Verified self-contained — does not require #4576 merged first (all helpers defined). Compact Worklog unchanged. Thanks!

Issue #2698, opened 2026-05-21
The Hermes environment variable HERMES_HOME is not effective.


HugoWw:
The Hermes environment variable HERMES_HOME has been set to "/root/.hermes/profiles/coding-ai". On the chat page, with the profile selected, all the profiles can be viewed.

HugoWw:
I have assigned a profile to each member of the team. Their profiles are all located in ~/.hermes/profiles/user1\user2\user3.... . I use the wrapp ctl.sh script to restrict every user 'HERMES_HOME'，so that every user only used thery profile(can't switch profile). My wrapp script is as follows:
```
#!/usr/bin/env bash
set -euo pipefail

WEBUI_DIR="/root/hermes-webui-user1"
PROFILE_HOME="/root/.hermes/profiles/user1"

export HERMES_WEBUI_HOST="127.0.0.1"
export HERMES_WEBUI_PORT="8789"

export HERMES_WEBUI_STATE_DIR="${PROFILE_HOME}/webui"
export HERMES_WEBUI_DEFAULT_WORKSPACE="/root/workspace/user1-workspace"

export HERMES_HOME="${PROFILE_HOME}"
export HERMES_CONFIG_PATH="${PROFILE_HOME}/config.yaml"

export HERMES_WEBUI_BOT_NAME="user1-Hermes"

export HERMES_WEBUI_PID_FILE="${PROFILE_HOME}/webui.pid"
export HERMES_WEBUI_LOG_FILE="${PROFILE_HOME}/webui.log"
export HERMES_WEBUI_CTL_STATE_FILE="${PROFILE_HOME}/webui.ctl.env"
export HERMES_WEBUI_PASSWORD="xxx"

mkdir -p "${HERMES_WEBUI_STATE_DIR}" "${HERMES_WEBUI_DEFAULT_WORKSPACE}"

cd "${WEBUI_DIR}"

case "${1:-status}" in
  start)
    exec ./ctl.sh start
    ;;
  stop)
    exec ./ctl.sh stop
    ;;
  restart)
    exec ./ctl.sh restart
    ;;
  status)
    exec ./ctl.sh status
    ;;
  logs)
    shift || true
    exec ./ctl.sh logs "$@"
    ;;
  *)
    echo "Usage: $0 {start|stop|restart|status|logs [--lines N] [--follow|--no-follow]}" >&2
    exit 2
    ;;
esac
```

[...]

PR #4589, opened 2026-06-21
fix(profiles): require explicit opt-in for isolated profile mode (hotfix #4586)

### LINKED ISSUE #4586: Regression (v0.51.528+): profile switching disabled & only 1 profile shown for normal single-user named profiles (#2698 isolated-mode false positive)
## Summary

After upgrading to **WebUI v0.51.547 / Agent 0.17.0**, profile switching breaks: the Profiles tab shows **only one profile** and switching is disabled. Rolling back resolves it.

Reported in Discord by **@b3nw**.

## Root cause (confirmed)

Regression introduced in **v0.51.528** by the isolated-single-profile-mode feature (#2698). `_is_isolated_profile_mode()` in `api/profiles.py` decides "isolated mode" purely from the *shape* of `HERMES_HOME`:

```python
# api/profiles.py
p = Path(hermes_home).expanduser()
if p.parent.name == 'profiles' and p.parent.parent.exists():
    return True   # -> isolated mode
```

The problem: the **Hermes Agent launcher sets `HERMES_HOME=~/.hermes/profiles/<name>` for any active named (non-default) profile** — this is the normal single-user path (`hermes_cli/main.py` profile override + `web_server.py` webui launch). So a regular single-user who simply runs under a named profile produces the exact same `*/profiles/<name>` shape that the heuristic treats as an intentional multi-user isolation deployment.

When isolated mode is (wrongly) on:
- `list_profiles_api()` returns only the active profile → **Profiles tab shows 1 profile**
- `switch_profile()` raises `PermissionError` for any other profile → **switching disabled**

The #2698 design note explicitly flagged this risk: *"The hard part is doing it without breaking single-user deployments where `HERMES_HOME=~/.hermes/` is the only sensible value."* The shape heuristic can't distinguish the two cases because they are byte-identical at the env-var level.

## Reproduction

Single user, active named profile (e.g. `webui`), with a couple of other profiles present:

```
~/.hermes/profiles/alpha
~/.hermes/profiles/beta
~/.hermes/profiles/webui   <- active; agent exports HERMES_HOME to this path
```

Result on >= v0.51.528:
- `_is_isolated_profile_mode()` → `True`
- Profiles tab lists only `webui`
- Switching to `alpha`/`beta` is blocked with "Profile switching is not allowed in isolated profile mode."

Roll back to < v0.51.528 → all profiles listed, switching works.

## Suggested fix

The `*/profiles/<name>` shape alone cannot signal intentional isolation. Require an **explicit opt-in** for isolated mode instead of inferring it from `HERMES_HOME` shape, e.g. an env flag (`HERMES_WEBUI_ISOLATED_PROFILE=1`) that the multi-user wrapper sets. Default OFF so a normal named-profile launch is never caught. The shape check can stay as a secondary requirement, but the explicit flag must be the primary gate.

## Impact

- **Severity: high** — breaks an existing, widely-used feature (profile switching) for any single-user running a named profile, which is the default once you create profiles.
- Affects v0.51.528 → current.

nesquena-hermes:
## Hotfix — fully resolves #4586

**High-severity regression (v0.51.528 → current): profile switching disabled & only one profile shown for normal single-user named profiles.**

### Root cause
The isolated-single-profile-mode feature (#2698) inferred "isolated mode" purely from the **shape** of `HERMES_HOME` — a `~/.hermes/profiles/<name>` path. But the Hermes Agent launcher exports exactly that shape for **any active named (non-default) profile** in a normal single-user deployment (`hermes_cli` profile override + the webui launch path). So an ordinary single user running under a named profile (e.g. `webui`) produced the byte-identical `*/profiles/<name>` shape the heuristic treated as an intentional multi-user isolation deployment.

When isolated mode was (wrongly) on:
- `list_profiles_api()` returned only the active profile → **Profiles tab showed 1 profile**
- `switch_profile()` raised `PermissionError` for any other profile → **switching disabled**

The #2698 design note explicitly flagged this risk ("the hard part is doing it without breaking single-user deployments"). The shape heuristic can't distinguish the two cases because they're identical at the env-var level.

### Fix
Require an **explicit opt-in** for isolated mode instead of inferring it from shape:

- New `HERMES_WEBUI_ISOLATED_PROFILE` env flag (truthy: `1`/`true`/`yes`/`on`; **default OFF**) is the **PRIMARY gate** — new helper `_isolated_profile_opt_in()`.
- The profile-shaped `HERMES_HOME` stays a **SECONDARY requirement** (both must hold), so a stray flag without a profile-shaped home doesn't engage isolation either.
- A normal named-profile launch (shape set, flag unset) is **never** caught → multi-profile behavior is restored. Multi-user isolation deployments set the flag.

All isolated-mode call sites (`list_profiles_api`, `switch_profile`, create/delete, cron pinning, active-name/home resolution) gate through `_is_isolated_profile_mode()`, so the single-function change covers the whole surface.

### Tests (RED→GREEN proven)
- `tests/test_issue4586_named_profile_not_isolated.py` (new): the exact false-positive repro — profile-shaped `HERMES_HOME` **without** the flag must stay normal (not isolated, all profiles listed, switching allowed). **Verified these 3 behavior tests FAIL on the pre-fix shape-only logic and PASS with the fix.** Plus opt-in coverage (flag+shape → isolated), flag-without-shape → off, and falsey-value handling.
- `tests/test_issue2698_isolated_hermes_home.py`: autouse fixture now sets `HERMES_WEBUI_ISOLATED_PROFILE=1` (this module tests the isolated posture); normal-mode assertions unaffected.

### Gates
Codex + Opus + full pytest suite — results posted in a follow-up comment.

### Migration note
Existing deployments relying on shape-only isolation must now set `HERMES_WEBUI_ISOLATED_PROFILE=1`. This is the intended trade-off per the issue: the shape alone cannot be a safe isolation signal.

Fixes #4586. Reported by **@b3nw**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Hotfix: profile switching works again for normal single-user named profiles (regression since v0.51.528, #4586).** The isolated-single-profile-mode feature (#2698) had been inferring "isolated mode" purely from the *shape* of `HERMES_HOME` (a `~/.hermes/profiles/<name>` path). But the Hermes Agent launcher exports exactly that shape for any active named profile in an ordinary single-user setup, so a regular user running under a named profile was wrongly pinned to a single profile — the Profiles tab showed only one profile and switching was blocked with "Profile switching is not allowed in isolated profile mode." Isolated mode now requires an explicit opt-in, `HERMES_WEBUI_ISOLATED_PROFILE=1` (default off), as the primary gate; the profile-shaped `HERMES_HOME` remains a secondary requirement. Multi-user isolation deployments simply set the flag; everyone else gets normal multi-profile behavior back. Thanks @b3nw for the report.
+
 ## [v0.51.548] — 2026-06-21 — Release TG (extension load diagnostics)
 
 ### Added
--- a/api/profiles.py
+++ b/api/profiles.py
@@ -125,41 +125,71 @@ def _unwrap_profile_home_to_base(home: Path) -> Path:
     return home
 
 
-def _is_isolated_profile_mode() -> bool:
-    """Detect isolated single-profile mode from HERMES_HOME env var.
+def _isolated_profile_opt_in() -> bool:
+    """Return True only when isolated single-profile mode is EXPLICITLY enabled.
 
-    Returns True when HERMES_HOME points at a concrete profile subdirectory
-    (e.g., ~/.hermes/profiles/user1) rather than the base home (~/.hermes).
-    This indicates a single-profile deployment where the WebUI should pin to
-    that profile and reject cross-profile operations.
+    Isolated mode is an intentional multi-user deployment posture (each user is
+    pinned to one profile and cross-profile operations are rejected). It must be
+    opted into with ``HERMES_WEBUI_ISOLATED_PROFILE`` — it is NEVER inferred from
+    the ``HERMES_HOME`` shape alone, because a normal single-user who runs under a
+    named profile produces the byte-identical ``*/profiles/<name>`` shape (the
+    Hermes Agent launcher exports ``HERMES_HOME=~/.hermes/profiles/<name>`` for any
+    active named profile). Keying isolation off the shape alone therefore breaks
+    profile switching for ordinary single-user deployments (#4586).
 
-    Isolation is derived only from the literal ``*/profiles/<name>`` shape of
-    HERMES_HOME at startup. A normal multi-profile deployment points
-    HERMES_HOME at the base home (``~/.hermes``), which does not match the
-    shape, so isolation stays off. We deliberately do not key this off
-    HERMES_BASE_HOME: that variable normally points at the base home already,
-    so using it as an opt-out silently disables legitimate isolated-profile
-    deployments.
-
-    Uses _INITIAL_HERMES_HOME (snapshotted at import time) to detect isolation,
-    not the current os.environ value. init_profile_state() overwrites HERMES_HOME
-    at startup, which would disable isolation detection if we read it here.
+    Accepts the usual truthy values; default (unset/empty/falsey) is OFF.
     """
+    return os.getenv('HERMES_WEBUI_ISOLATED_PROFILE', '').strip().lower() in (
+        '1', 'true', 'yes', 'on',
+    )
+
+
+def _is_isolated_profile_mode() -> bool:
+    """Detect isolated single-profile mode.
+
+    Returns True only when BOTH conditions hold:
+      1. ``HERMES_WEBUI_ISOLATED_PROFILE`` is explicitly enabled (the PRIMARY
+         gate — see _isolated_profile_opt_in), AND
+      2. HERMES_HOME at startup points at a concrete profile subdirectory
+         (e.g., ~/.hermes/profiles/user1) rather than the base home.
+
+    Why the explicit flag is required (#4586 regression fix): the
+    ``*/profiles/<name>`` shape alone CANNOT distinguish an intentional
+    multi-user isolation deployment from an ordinary single-user running under a
+    named profile — the Hermes Agent launcher sets
+    ``HERMES_HOME=~/.hermes/profiles/<name>`` for any active named profile, so the
+    two cases are byte-identical at the env-var level. Inferring isolation from
+    the shape alone (the v0.51.528 behaviour from #2698) wrongly pinned ordinary
+    single-user deployments to one profile and disabled profile switching. The
+    multi-user wrapper that genuinely wants isolation now sets the explicit flag;
+    everyone else is never caught. The shape stays as a secondary requirement so
+    a stray flag without a profile-shaped HERMES_HOME does not engage isolation.
+
+    Uses _INITIAL_HERMES_HOME (snapshotted at import time) to detect the shape,
+    not the current os.environ value. init_profile_state() overwrites HERMES_HOME
+    at startup, which would disable detection if we read it here.
+    """
+    # PRIMARY gate: explicit opt-in. Default OFF → a normal named-profile launch
+    # is never treated as isolated, so profile switching keeps working (#4586).
+    if not _isolated_profile_opt_in():
+        return False
+
     hermes_home = _INITIAL_HERMES_HOME
     if not hermes_home:
         return False
 
     p = Path(hermes_home).expanduser()
-    # Check if this path looks like ~/.hermes/profiles/<name>
-    # i.e., parent dir is named 'profiles' and grandparent exists
+    # SECONDARY requirement: HERMES_HOME must look like ~/.hermes/profiles/<name>
+    # i.e., parent dir is named 'profiles' and grandparent exists.
     if p.parent.name == 'profiles' and p.parent.parent.exists():
         return True
     if p.is_symlink():
         global _ISOLATED_SYMLINK_WARNING_EMITTED
         if not _ISOLATED_SYMLINK_WARNING_EMITTED:
             logger.warning(
-                "HERMES_HOME symlink %s does not literally match */profiles/<name>; "
-                "isolated profile mode is disabled unless the literal profile path is used.",
+                "HERMES_WEBUI_ISOLATED_PROFILE is set but HERMES_HOME %s does not "
+                "literally match */profiles/<name>; isolated profile mode stays off "
+                "unless the literal profile path is used.",
                 p,
             )
             _ISOLATED_SYMLINK_WARNING_EMITTED = True
--- a/tests/test_issue2698_isolated_hermes_home.py
+++ b/tests/test_issue2698_isolated_hermes_home.py
@@ -1,9 +1,16 @@
 """
 Tests for issue #2698: HERMES_HOME isolated profile mode.
 
-When HERMES_HOME points at a specific profile directory like ~/.hermes/profiles/user1,
-the WebUI should pin to that single profile: list only it, reject create/switch/delete
-of other profiles, and hide multi-profile UI affordances.
+When HERMES_HOME points at a specific profile directory like ~/.hermes/profiles/user1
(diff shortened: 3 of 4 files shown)


nesquena-hermes:
## Gate results — Codex + Opus + full suite

**All three gates ran on the committed fix.** Both reviewers confirmed the core fix is correct and the OFF default is the right call (no *unintended* regression). Both raised the same theme — a genuine isolated deployment could **silently lose isolation** — in two forms:

- **Opus (SHIP-WITH-FIXES):** wants a one-time startup warning for the shape-matches-but-flag-unset case + a CHANGELOG upgrade advisory (observability, not logic).
- **Codex (SHIP ONLY WITH FIXES):** sharper — `_isolated_profile_opt_in()` reads `os.environ` live, and `_reload_dotenv()` folds a pinned profile's `.env` into `os.environ` at startup, so a profile `.env` containing `HERMES_WEBUI_ISOLATED_PROFILE=0` can flip isolation from on→off after launch. Suggested fix: snapshot the opt-in at import (beside `_INITIAL_HERMES_HOME`).

### Decision (maintainer call): **deferred, by design — shipping the minimal fix now.**
The isolated-mode feature (#2698) shipped only days ago (v0.51.528), so there are effectively no real multi-user isolated deployments yet to protect, and the regression this PR fixes is actively breaking ordinary single-users *today*. The "silently loses isolation" concern (both the upgrade-warning and the `.env`-override path) is acceptable for now and explicitly deferred — **not silently dropped.**

**Follow-up (tracked):** harden the opt-in to a startup snapshot (so a profile `.env` can't disable isolation) + add a one-time "shape matches but flag unset → isolation off" startup warning, before isolated mode has real deployments. Will file a tracking issue.

Full suite: green (results in-thread). The regression tests are proven RED→GREEN (3 behavior tests fail on the pre-fix shape-only logic, pass with the fix).

nesquena (approved):

Independent review of the **#4586 hotfix** — a high-severity regression (since v0.51.528, #2698): normal single-user deployments running under a *named* profile were wrongly treated as multi-user isolation deployments, so the Profiles tab showed only one profile and switching was blocked with `PermissionError`.

### What this ships

`api/profiles.py` (+51/-21), `tests/test_issue4586_named_profile_not_isolated.py` (+157, new), `tests/test_issue2698_isolated_hermes_home.py` (+19/-5), `CHANGELOG.md` (+4). Agent-authored. Closes #4586 (reported by @b3nw).

### Root cause — confirmed against the agent

`_is_isolated_profile_mode()` inferred isolation purely from the `*/profiles/<name>` shape of `HERMES_HOME`. I confirmed in the fresh `nousresearch/hermes-agent` tarball that the launcher exports **exactly that shape for any active named profile** in an ordinary single-user setup: `hermes_cli/main.py` resolves a `--profile`/active profile and sets `HERMES_HOME=<root>/profiles/<name>` (`main.py:480`, and `gateway.py:5625` documents `HERMES_HOME=<root>/profiles/<name>`). So the two deployment types are **byte-identical at the env-var level** — the shape heuristic genuinely cannot distinguish them, exactly as the #2698 design note warned. Root cause is real and correctly diagnosed.

### Fix — traced

Isolation now requires an **explicit opt-in** as the primary gate:

- New `_isolated_profile_opt_in()` ([profiles.py:128](api/profiles.py#L128)) reads `HERMES_WEBUI_ISOLATED_PROFILE` from the **live** env, truthy ∈ {`1`,`true`,`yes`,`on`} (`.strip().lower()`), default OFF.
- `_is_isolated_profile_mode()` short-circuits `if not _isolated_profile_opt_in(): return False`, then keeps the `*/profiles/<name>` shape check on the **snapshotted** `_INITIAL_HERMES_HOME` as a secondary requirement. Both must hold.

So a normal named-profile launch (shape set, flag unset) is never caught → multi-profile behaviour is restored; a stray flag without a profile-shaped home doesn't engage isolation either.

**The snapshot-vs-live asymmetry is intentional and correct:** the *shape* must read `_INITIAL_HERMES_HOME` because `init_profile_state()` overwrites `HERMES_HOME` at startup (reading it live would always fail the shape check); the *flag* is read live and is never overwritten by init, so reading it live is both correct and more robust than snapshotting.

**Single-function blast radius:** all **12** call sites (`list_profiles_api`, `switch_profile`, create/delete, cron pinning, active-name/home resolution, …) gate through `_is_isolated_profile_mode()`; the diff touches only the two helper functions — no call site changes — so the fix propagates uniformly. ✓

### Security review — the key consideration (deliberate, documented trade-off)

This change **loosens** an isolation boundary: previously shape-alone engaged isolation; now shape-without-flag does not. **A genuine multi-user isolation deployment that relied on shape-only will silently lose isolation on upgrade until it sets `HERMES_WEBUI_ISOLATED_PROFILE=1`** — every profile becomes listable and switchable in the interim.

I weighed this carefully and concur it's the correct call:
- The shape was never a reliable isolation signal — it was simultaneously **over-broad** (breaking every single-user named-profile deployment, the high-severity #4586 symptom) and the only isolation signal. There is no way to distinguish the two byte-identical env shapes without an explicit signal.
- The issue's own suggested fix is precisely "require an explicit opt-in flag, default OFF, shape as secondary."
- The PR carries a migration note, and the symlink-branch warning was updated to reference the new flag.

So the trade-off is unavoidable and intended. My one **non-blocking** suggestion (below) would soften the silent-loss edge for upgrading isolation admins.

- **Default-OFF posture** is the safe regression fix (normal users get multi-profile back; isolation is now strictly opt-in).
- **Falsey/whitespace handling:** `.strip().lower()` means `""`, `"0"`, `"false"`, `"  "` → OFF; case-insensitive truthy (`"TRUE"` → on). Covered by tests.
- **No new sink / injection:** reads one env var into a boolean; no path, SQL, HTML, or subprocess surface.
- **Cross-tool:** `HERMES_WEBUI_ISOLATED_PROFILE` is absent from the agent — a webui-only flag the multi-user deployment wrapper sets; the agent is unaffected.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Single user, named profile (shape set, flag unset) | NOT isolated → all profiles listed, switching allowed ✅ (the #4586 fix) |
| Multi-user isolation, flag + profile-shaped home | isolated → pinned to active, cross-profile rejected ✅ |
| Flag set, HERMES_HOME = base `~/.hermes` (no shape) | NOT isolated (secondary requirement fails) ✅ |
| Falsey/empty/whitespace flag values | NOT isolated ✅ |
| Flag truthy variants (`1`/`true`/`TRUE`/`yes`/`on`) | isolated (with shape) ✅ |
| Symlinked HERMES_HOME, flag set, shape mismatch | one-time warning, isolation stays off ✅ |
| Existing shape-only isolation deployment, upgraded | ⚠️ isolation OFF until flag is set (documented migration) |

### Tests

- `tests/test_issue4586_named_profile_not_isolated.py` (new) — the exact false-positive repro (shape-only → normal), both regressed symptoms (profiles-tab clamp gate + `switch_profile` no `PermissionError`), explicit opt-in (parametrized truthy), flag-without-shape → off, falsey values → off. The autouse fixture clears the flag so ambient env can't mask the regression. PR reports these fail pre-fix / pass post-fix (RED→GREEN).
- `tests/test_issue2698_isolated_hermes_home.py` — autouse fixture now sets the flag for its isolated-posture module; normal-mode assertions there point at the base home (shape fails regardless), so they're unaffected.
- **Full suite: 9781 passed / 132 skipped / 0 failures** in 206s (deselected the pre-existing darwin CRLF flake). Verified the RED→GREEN at the gate directly: master's `_is_isolated_profile_mode()` returns `True` for a shape-only named profile (the bug); the fix returns `False`.

- **Soften the silent isolation-loss on upgrade:** there's a one-time warning for *flag-set-but-shape-mismatch*, but none for the inverse — *profile-shape-present-but-flag-absent* — which is exactly the case where an upgrading multi-user isolation deployment silently drops its boundary. A one-time INFO/WARNING there ("HERMES_HOME looks like an isolated-profile path but HERMES_WEBUI_ISOLATED_PROFILE is unset; isolation is OFF — set the flag if this is a multi-user deployment") would help isolation admins notice. I did **not** push it because it would also fire for every normal single-user named-profile deployment (the common case this fix targets), so it risks log noise; the release/migration note is the better channel. Flagging for the release notes' attention.
- CHANGELOG `[Unreleased]` entry is present and accurate (the versioned section is the release agent's domain).

### Recommendation

A minimal, correct, single-function hotfix for a high-severity regression: gate isolated mode on an explicit `HERMES_WEBUI_ISOLATED_PROFILE` opt-in (default off) with the profile shape as a secondary requirement. Root cause is confirmed against the agent launcher (it really does export the `profiles/<name>` shape for any named profile), the snapshot-vs-live asymmetry is intentional and correct, all 12 call sites benefit from the one change, and the regression is locked with RED→GREEN tests. The deliberate security trade-off (shape-only isolation must now set the flag) is the only correct resolution given the byte-identical env shapes and is documented — I flagged a migration-note nicety for the release. Ship.

nesquena-hermes:
Shipped as **v0.51.549** (Release TH) — live in prod, verified: the profiles API now returns all 5 profiles (was clamped to 1) and switching works again. 🔧

This closes #4586. Deep-gated as a hotfix — beyond the core fix (isolated mode now requires the explicit HERMES_WEBUI_ISOLATED_PROFILE opt-in instead of being inferred from the HERMES_HOME path shape), the gate caught and I fixed TWO isolation-ESCAPE holes:
1. _reload_dotenv() would copy HERMES_WEBUI_ISOLATED_PROFILE from a profile .env into os.environ — a contained user could set =0 to escape isolation. Now protected via _PROTECTED_ENV_KEYS.
2. get_profile_runtime_env() (the background-worker/streaming runtime path) had the same escape — now stripped there too, + the key added to _BLOCKED_RUNTIME_ENV_KEYS.

Gate: 9911 suite passed, Codex SAFE (final, no remaining escape path), Opus SAFE. Both new escape paths have regression tests. Thanks for the fast, well-scoped hotfix.

Issue #3133, opened 2026-05-29
上下文压缩存在漏洞，超限后对话内容易丢失且上下文统计不准确


Lemonseaa:
## 问题描述

上下文压缩功能存在两个严重问题：

1. **对话内容丢失**：上下文长度超过限制后，对话内容容易在压缩过程中丢失，且难以定位和查找丢失的内容。

2. **上下文计数不更新**：压缩上下文后，检测到的上下文数量没有变化，完全无法反映真实的上下文占用情况。这导致用户无法判断当前对话的实际状态，容易误以为还有足够空间继续对话，实际上已经压缩或丢失了大量内容。

## 预期行为

- 压缩后应准确反映当前上下文占用量
- 用户应能清晰看到压缩前后的上下文变化
- 丢失的对话内容应有迹可循

## 环境

- hermes-webui latest
- macOS

nesquena-hermes:
## 感谢反馈 — 需要确认与现有压缩问题的关系 / Thanks — confirming overlap with known compression issues

你报告了两个问题，我们分别看：

**1. 上下文压缩时对话内容丢失 (content lost during compression)**
这与几个正在处理的问题相关：#3120（滚动时工具调用消失）、#3079 / #3014（长任务的中间进度），以及 #2761（会话被 "No LLM provider configured" 卡住）。我们想确认你遇到的是"内容真的从存储中丢失"，还是"内容仍在但界面不再显示"。
- 刷新页面或切换会话再切回来后，丢失的内容会**恢复**吗？
- 还是即使刷新后也**永久消失**？

**2. 压缩后上下文计数不更新 (context count not updating after compression)**
这是一个明确的 bug。压缩后 token 计数应该下降以反映实际占用。请提供：
- WebUI 版本号（设置 → 关于）
- 压缩前后顶栏显示的上下文数字（如果可以，截图）

---

**English summary:** Two issues here — (1) conversation content lost during compression, and (2) the context/token counter not updating after compression. For (1), the key question is whether the content is *truly gone from storage* or *still present but no longer rendered* — does it come back after a reload/session-switch, or is it permanently gone? This overlaps with #3120, #3079/#3014, and #2761. For (2) — the counter not dropping after compression — that's a clear bug; please share your WebUI version and the topbar context number before vs. after a compression event (a screenshot is ideal).

[...]

PR #4594, opened 2026-06-21
Preserve visible transcript after manual /compress

franksong2702:
## Thinking Path

- #3133 reports that manual compression can leave the model context reduced while messages disappear from the conversation window.
- The relevant contract already separates the visible transcript from the model-facing `context_messages`.
- The manual `/compress` route was violating that boundary by writing the compressed payload into both `s.messages` and `s.context_messages`.
- The fix keeps the visible transcript and settled tool history durable, while only pruning the model-facing context payload.

Refs #3133.

## What Changed

- Preserved `s.messages` after successful manual `/compress` instead of replacing it with the compressed context payload.
- Stored the compressed result only in `s.context_messages`.
- Preserved settled `s.tool_calls` so historical tool cards / Worklog rows remain visible after manual compression.
- Computed compression anchor metadata from the preserved visible transcript tail.
- Updated manual compression regression tests for both sync `/compress` and async `/compress/start` -> status `done`.
- Added a changelog entry for the user-visible behavior fix.

## Why It Matters

Manual compression should reduce what the agent sees next without making the user-visible conversation look like it lost history. This keeps the conversation window, reload behavior, and model-facing context boundary aligned with the run-state contract.

## Contract Routing

- Touched contract family: run-state consistency / compression / model-context reconstruction.
- Evidence: `docs/CONTRACTS.md` routes compression and context reconstruction changes through the state-contract checks; `docs/rfcs/webui-run-state-consistency-contract.md` explicitly separates the visible transcript from model context / `context_messages`.
- Contract change: none intended. This implements the existing boundary more faithfully by preserving visible `messages` while pruning only `context_messages`.

## Verification

- `./scripts/test.sh tests/test_sprint46.py::test_session_compress_roundtrip tests/test_sprint46.py::test_session_compress_start_is_async_and_reuses_running_job tests/test_issue3800_compaction_summary_length.py tests/test_issue2028_compression_anchor_helpers.py`
  - `13 passed`
- `python3 -m py_compile api/routes.py api/models.py`
- Claude Code review: no blocking issues.

## Risks / Follow-ups

- This is intentionally limited to manual `/compress`; it does not change automatic streaming compression.
- #3133 also contains a broader context-count reporting thread. This PR fixes the confirmed manual-compress disappearing-message root cause, but does not claim to close every context-count variant.
- No browser manual run was performed because the API regression locks the state-layer boundary directly.

## Model Used

AI-assisted.

- OpenAI Codex, GPT-5-based coordinator.
- Codex worker sub-agent for the bounded implementation patch.
- Claude Code CLI `2.1.185` review pass; no blocking issues found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,11 @@
 
 ## [Unreleased]
 
+## [v0.51.549] — 2026-06-21 — Release TH (hotfix: profile switching restored for single-user named profiles)
+
 ### Fixed
 
-- **Manual `/compress` no longer replaces the visible conversation transcript with the compressed context payload (#3133).** Successful manual compression now stores the compacted payload only in `context_messages`, keeps `messages` as the durable visible transcript for the conversation window and reloads, and computes compression-anchor metadata from that preserved transcript boundary.
+- **Profile switching works again for normal single-user named profiles (hotfix #4586).** A regression (since v0.51.528) wrongly treated any single user running under a named profile as an isolated multi-user deployment — because isolated mode was inferred from the `HERMES_HOME` path shape (`*/profiles/<name>`), which the agent launcher exports for any named profile. The Profiles tab showed only one profile and switching was disabled. Isolated single-profile mode now requires an explicit `HERMES_WEBUI_ISOLATED_PROFILE` opt-in and is never inferred from the path shape. The opt-in is also protected from being overridden by a profile's own `.env` (on both the live-env and runtime/background-worker paths), so a pinned profile can't disable its own isolation. Thanks @nesquena-hermes.
 
 ## [v0.51.548] — 2026-06-21 — Release TG (extension load diagnostics)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Manual `/compress` no longer replaces the visible conversation transcript with the compressed context payload (#3133).** Successful manual compression now stores the compacted payload only in `context_messages`, keeps `messages` as the durable visible transcript for the conversation window and reloads, and computes compression-anchor metadata from that preserved transcript boundary.
+
 ## [v0.51.548] — 2026-06-21 — Release TG (extension load diagnostics)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,18 @@
 
 ## [Unreleased]
 
+## [v0.51.551] — 2026-06-21 — Release TJ (manual /compress keeps the visible conversation)
+
+### Fixed
+
+- **Manually compressing a session no longer makes earlier messages disappear from the conversation window (#3133).** The manual `/compress` action was overwriting the visible transcript with the reduced model-facing context (and clearing settled tool history), so messages vanished on screen and after reload even though only the model's context should shrink. It now preserves the visible transcript and tool history and prunes only the model-facing context payload — matching how automatic compression already behaves. Thanks @franksong2702.
+
+## [v0.51.550] — 2026-06-21 — Release TI (Transparent Stream keeps its activity after a turn settles or reloads)
+
+### Fixed
+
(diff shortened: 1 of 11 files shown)


nesquena-hermes:
I pulled the branch and read the diff against `origin/master`, plus the surrounding contract code (`api/streaming.py` context-read + merge paths) to confirm the boundary holds beyond the single turn. This looks correct, and the test changes back it up well.

## What the diff does

`api/routes.py:18393` (`_handle_session_compress`) changes from overwriting all three fields:

```python
s.messages = compressed
s.context_messages = compressed
s.tool_calls = []
```

to compressing only the model-facing prefix and recomputing the anchor from the preserved transcript tail:

```python
s.context_messages = copy.deepcopy(compressed)
...
visible_after = visible_messages_for_anchor(s.messages, auto_compression=False)
```

## Why this is the right boundary

Two reads confirm the model side still gets the compressed context:

- `_session_context_messages` (`api/streaming.py:4349-4355`) prefers `context_messages` when non-empty, so the next turn feeds the model the compressed prefix, not the full `s.messages`.
- More importantly, `_merge_display_messages_after_agent_result` (`api/streaming.py:4433`) is explicitly designed to *"keep UI transcript durable while allowing model context to compact"* — when the returned history no longer has the prior context as a prefix, it keeps `previous_display` and appends only the delta.

That second point is the key one: it means the preserved `s.messages` won't silently collapse back to the compressed form on the *next* turn's writeback. Automatic streaming compression already relied on this merge to keep the visible transcript intact; manual `/compress` was the outlier that bypassed it by writing `s.messages = compressed` directly. This PR brings the two paths into agreement, which matches the run-state RFC's separation of visible transcript vs `context_messages`.

## On preserving `s.tool_calls`

Keeping the settled tool calls (rather than clearing to `[]`) is not just cosmetic — it's a correctness requirement here. The tool-call entries carry `assistant_msg_idx` pointing into the visible transcript (your test stamps `"assistant_msg_idx": 1`). Since `s.messages` is now preserved, those indices stay valid; clearing `tool_calls` while keeping `messages` would have orphaned the Worklog cards, and compressing `messages` while keeping `tool_calls` would have dangled the indices. Preserving both together is the only internally consistent option, so the pairing is right.

## Tests

The `test_session_compress_roundtrip` update is the convincing part: it asserts `loaded.messages == original_messages`, `loaded.context_messages == compressed_messages`, `loaded.tool_calls == settled_tool_calls`, and the anchor moving to visible idx 3 (`"four"`) instead of the old compressed idx 1. It checks both the in-memory session and `Session.load(sid)`, so the persisted divergence is covered, not just the response payload. The async `/compress/start` test gets the same treatment.

## One thing worth confirming (non-blocking)

The PR scopes itself to manual `/compress` and doesn't touch automatic streaming compression, which is the right call. The only edge I'd sanity-check: if `context_compressor.compress()` ever returns an empty list, `_session_context_messages` falls back to `s.messages` (the full transcript) — a safe degradation, not a regression, but worth a one-line guard or comment if an empty compression result is reachable in practice. Not a blocker. CI is green across 3.11/3.12/3.13 and the contract reads check out.

franksong2702:
Follow-up pushed in `ac4565dea`.

- Removed the direct `CHANGELOG.md` entry during conflict resolution, leaving release notes to the release flow.
- Fixed the CI-only `test (3.11, 2)` failure by isolating the legacy imported Weixin metadata-backfill test with a unique `session_key`, so it no longer collides with same-source messaging dedupe state under shard ordering.
- Local checks passed: `./scripts/test.sh tests/test_gateway_sync.py -q`, targeted gateway backfill/latest-dedupe tests, the manual-compress regression set, `python3 -m py_compile api/routes.py api/models.py`, and `git diff --check`.

franksong2702:
Confirming the `tests/test_gateway_sync.py` change is intentional.

It is not part of the manual-compress product path; it is the CI follow-up for the Python 3.11 shard-2 failure that appeared after the `origin/master` refresh. The failing test was using a bare `weixin` messaging identity while asserting that a legacy imported metadata-backfill row appears in `/api/sessions`. Under full-shard ordering, that can collide with same-source messaging dedupe state, so the fixture now uses a unique `session_key` and tests the intended contract: legacy imported messaging source metadata is backfilled and visible for that specific session identity.

Coverage for that follow-up:

- `./scripts/test.sh tests/test_gateway_sync.py::test_import_cli_preserves_messaging_source_metadata tests/test_gateway_sync.py::test_sessions_response_backfills_imported_messaging_source_metadata tests/test_gateway_sync.py::test_sessions_response_keeps_only_latest_messaging_session_per_source -q` -> `3 passed`
- `./scripts/test.sh tests/test_gateway_sync.py -q` -> `61 passed`
- GitHub `test (3.11, 2)` now passes on head `ac4565dea`.

nesquena-hermes:
Shipped as **v0.51.551** (Release TJ) — live in prod. Closes #3133. 🗜️

Gate: full suite 9913 passed, Codex SAFE, Opus SAFE — confirmed the manual /compress path now matches the auto-compression contract (visible s.messages + tool history durable, only model-facing context_messages pruned, deep-copied; anchor from the visible tail). The #3133 symptom (messages vanishing on screen + after reload) is directly fixed and test-covered. Thanks @franksong2702.

PR #4597, opened 2026-06-21
perf: reduce session-list work for archived histories

santastabber:
## Thinking Path

- The session sidebar polls `/api/sessions` frequently, so that endpoint needs to stay cheap even when a profile has a long history.
- Archived rows are useful for the user-controlled archived view, but they do not need to be included or lineage-enriched on every default sidebar refresh.
- Existing lineage metadata must still be preserved for visible continuation rows, including when their archived parent is filtered out of the default payload.
- This change keeps archived sessions available on demand while making the default session-list payload smaller and avoiding full-history lineage enrichment in the hot path.

## What Changed

- Added an optional `include_lineage_metadata` keyword to `all_sessions()`, defaulting to the existing enriched behavior for compatibility.
- Updated `/api/sessions` to call `all_sessions(..., include_lineage_metadata=False)` for its cache builder, then apply lineage enrichment only to the rows selected for the response.
- Added `include_archived` support to `/api/sessions` and included it in the session-list cache key.
- Changed the default `/api/sessions` payload to omit archived rows while returning archived counts:
  - `archived_count`
  - `archived_webui_count`
  - `archived_cli_count`
  - `include_archived`
- Kept archived rows available through `GET /api/sessions?include_archived=1`.
- Updated the sidebar client so the archived toggle triggers a real refetch with `include_archived=1` instead of requiring all archived rows in every default payload.
- Added/updated regression tests for the hot-path filtering, archived on-demand fetch, cache/diagnostic wiring, stale-stream reconciliation fixtures, cron/webhook archived projections, and real lineage preservation.

## Why It Matters

Large histories can make the sidebar feel slow because the default session-list endpoint previously carried and enriched more rows than the visible sidebar needed. Omitting archived rows from the default response reduces hot-path work and payload size while preserving the archived view as an explicit, on-demand fetch.

The real-lineage regression ensures the optimization does not break compression-continuation metadata for visible rows when an archived parent is not returned in the default payload.

## Contract Routing

Task type: runtime/session-list performance fix.

Touched areas:
- `/api/sessions` response construction and cache keying
- session-list lineage enrichment timing
- sidebar archived-session toggle fetch behavior

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/rfcs/canonical-session-resolution.md`

Scope boundaries:
- This is not a full sidebar pagination rewrite.
- This does not remove archived-session support.
- This keeps `all_sessions()` enriched by default for existing call sites.

- focused regression tests for archived default omission and on-demand fetch
- real lineage regression for visible child rows with archived parents
- syntax/lint checks for touched Python/JS

## Verification

- `python3 -m py_compile api/models.py api/routes.py tests/test_session_lineage_metadata_api.py tests/test_session_list_long_history_perf.py tests/test_issue4385_cron_archive_reappears.py tests/test_issue1855_request_diagnostics.py tests/test_issue2157_sessions_list_stale_stream_state.py tests/test_session_attention_badges.py`
- `git diff --check`
  - Result: `ruff_lint: no changed .py files vs origin/master (6c78d1c315bc). OK.`
- `./scripts/test.sh tests/test_session_lineage_metadata_api.py tests/test_session_list_long_history_perf.py tests/test_session_sidebar_cache.py tests/test_request_diagnostics_cache.py tests/test_issue1855_request_diagnostics.py tests/test_issue2157_sessions_list_stale_stream_state.py tests/test_session_attention_badges.py tests/test_session_index.py tests/test_issue3238_orphaned_cli_sidecar_prune.py tests/test_issue4385_cron_archive_reappears.py -q`
  - Result: `99 passed`

## Risks / Follow-ups

- The archived-session count now reflects the rows that survive the same dedupe/cap path as the on-demand archived payload, not a theoretical total before those filters. That keeps the count aligned with what the archived fetch can return.
- The sidebar still does not implement general server-side pagination; this is a focused hot-path reduction for archived rows and lineage enrichment.
- The `all_sessions()` compatibility fallback in the route builder exists for tests and any monkeypatched callers using the historical `diag`-only signature.

## Model Used

AI-assisted.

- Provider/model: OpenRouter `openai/gpt-5.5` for implementation orchestration.
- Review: local AI-assisted review was used before preparing this draft.
- Notable mode/tool use: isolated local worktree; no deployment or runtime service changes.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -24,6 +24,7 @@ from api.usage import prompt_cache_hit_percent
 from api.agent_sessions import (
     _is_continuation_session,
     is_cli_session_row,
+    normalize_agent_session_source,
     read_importable_agent_session_rows,
     read_session_lineage_metadata,
 )
@@ -3192,6 +3193,106 @@ def _sidebar_title_is_generic_webui(title: str | None) -> bool:
     return text.startswith(prefix) and text[len(prefix):].isdigit()
 
 
+def _read_state_db_sidebar_overrides(db_path: Path, session_ids: set[str]) -> dict[str, dict]:
+    """Return cheap state.db source/title overrides for sidebar rows.
+
+    This intentionally does not chase lineage parents/children. It is used on
+    the /api/sessions hot path before CLI filtering so state.db can correct
+    stale JSON source flags without paying the full lineage-enrichment cost.
+    """
+    wanted = {str(sid) for sid in (session_ids or set()) if sid}
+    if not wanted or not db_path.exists():
+        return {}
+    try:
+        import sqlite3
+    except ImportError:
+        return {}
+    try:
+        with closing(sqlite3.connect(str(db_path))) as conn:
+            conn.row_factory = sqlite3.Row
+            cur = conn.cursor()
+            cur.execute("PRAGMA table_info(sessions)")
+            session_cols = {row[1] for row in cur.fetchall()}
+            if 'id' not in session_cols:
+                return {}
+            source_expr = 's.source' if 'source' in session_cols else 'NULL AS source'
+            session_source_expr = 's.session_source' if 'session_source' in session_cols else 'NULL AS session_source'
+            title_expr = 's.title' if 'title' in session_cols else 'NULL AS title'
+            overrides: dict[str, dict] = {}
+            ids = list(wanted)
+            chunk_size = 500
+            for i in range(0, len(ids), chunk_size):
+                chunk = ids[i:i + chunk_size]
+                placeholders = ','.join('?' * len(chunk))
+                cur.execute(
+                    f"""
+                    SELECT s.id, {source_expr}, {session_source_expr}, {title_expr}
+                    FROM sessions s
+                    WHERE s.id IN ({placeholders})
+                    """,
+                    chunk,
+                )
+                for row in cur.fetchall():
+                    sid = str(row['id'])
+                    entry: dict[str, object] = {}
+                    state_title = str(row['title'] or '').strip()
+                    if state_title:
+                        entry['_state_db_title'] = state_title
+                    state_source = str(row['source'] or '').strip().lower()
+                    if state_source:
+                        entry['_state_db_source'] = state_source
+                        source_meta = normalize_agent_session_source(state_source)
+                        entry['_state_db_source_tag'] = state_source
+                        entry['_state_db_raw_source'] = source_meta.get('raw_source')
+                        entry['_state_db_session_source'] = source_meta.get('session_source')
+                        entry['_state_db_source_label'] = source_meta.get('source_label')
+                    if entry:
+                        overrides[sid] = entry
+            return overrides
+    except Exception:
+        return {}
+
+
+def _apply_sidebar_state_db_overrides(sessions: list[dict]) -> None:
+    """Apply state.db source/title overrides without full lineage enrichment."""
+    try:
+        metadata = _read_state_db_sidebar_overrides(
+            _active_state_db_path(),
+            {str(s.get('session_id')) for s in sessions if s.get('session_id')},
+        )
+    except Exception:
+        return
+    _apply_sidebar_state_db_override_metadata(sessions, metadata)
+
+
+def _apply_sidebar_state_db_override_metadata(sessions: list[dict], metadata: dict[str, dict]) -> None:
+    for session in sessions:
+        sid = session.get('session_id')
+        if sid not in metadata:
+            continue
+        entry = dict(metadata[sid])
+        state_db_title = entry.pop('_state_db_title', None)
+        state_db_source = entry.pop('_state_db_source', None)
+        state_db_source_tag = entry.pop('_state_db_source_tag', None)
+        state_db_raw_source = entry.pop('_state_db_raw_source', None)
+        state_db_session_source = entry.pop('_state_db_session_source', None)
+        state_db_source_label = entry.pop('_state_db_source_label', None)
+        if state_db_source == 'webui':
+            session['source_tag'] = state_db_source_tag
+            session['raw_source'] = state_db_raw_source
+            session['session_source'] = state_db_session_source
+            session['source_label'] = state_db_source_label
+            session['is_cli_session'] = False
+        title = session.get('title')
+        if (
+            state_db_title
+            and state_db_title != title
+            and _sidebar_title_is_generic_webui(title)
+        ):
+            session['_state_db_title'] = state_db_title
+            session['display_title'] = state_db_title
+
+
 def _enrich_sidebar_lineage_metadata(sessions: list[dict]) -> None:
     """Attach state.db compression lineage metadata used by sidebar collapse."""
     try:
@@ -3201,31 +3302,21 @@ def _enrich_sidebar_lineage_metadata(sessions: list[dict]) -> None:
         )
     except Exception:
         return
+    _apply_sidebar_state_db_override_metadata(sessions, metadata)
     for session in sessions:
         sid = session.get('session_id')
         if sid in metadata:
             entry = dict(metadata[sid])
-            state_db_title = entry.pop('_state_db_title', None)
-            state_db_source = entry.pop('_state_db_source', None)
-            state_db_source_tag = entry.pop('_state_db_source_tag', None)
-            state_db_raw_source = entry.pop('_state_db_raw_source', None)
-            state_db_session_source = entry.pop('_state_db_session_source', None)
-            state_db_source_label = entry.pop('_state_db_source_label', None)
-            session.update(entry)
-            if state_db_source == 'webui':
-                session['source_tag'] = state_db_source_tag
-                session['raw_source'] = state_db_raw_source
-                session['session_source'] = state_db_session_source
-                session['source_label'] = state_db_source_label
-                session['is_cli_session'] = False
-            title = session.get('title')
-            if (
-                state_db_title
-                and state_db_title != title
-                and _sidebar_title_is_generic_webui(title)
+            for key in (
+                '_state_db_title',
+                '_state_db_source',
+                '_state_db_source_tag',
+                '_state_db_raw_source',
+                '_state_db_session_source',
+                '_state_db_source_label',
             ):
-                session['_state_db_title'] = state_db_title
-                session['display_title'] = state_db_title
+                entry.pop(key, None)
+            session.update(entry)
 
 
 def _diag_stage(diag, name: str) -> None:
@@ -3236,7 +3327,7 @@ def _diag_stage(diag, name: str) -> None:
             pass
 
 
-def all_sessions(diag=None):
+def all_sessions(diag=None, *, include_lineage_metadata: bool = True):
     _diag_stage(diag, "all_sessions.active_streams")
     active_stream_ids = _active_stream_ids()
     # Phase C: try index first for O(1) read; fall back to full scan
@@ -3330,8 +3421,13 @@ def all_sessions(diag=None):
                 and not s.get('has_pending_user_message')
                 and not s.get('worktree_path')
             )]
-            _diag_stage(diag, "all_sessions.lineage_metadata")
-            _enrich_sidebar_lineage_metadata(result)
+            if include_lineage_metadata:
+                _diag_stage(diag, "all_sessions.lineage_metadata")
+                _enrich_sidebar_lineage_metadata(result)
+            else:
+                _diag_stage(diag, "all_sessions.state_db_overrides")
+                _apply_sidebar_state_db_overrides(result)
+                _diag_stage(diag, "all_sessions.lineage_metadata_skipped")
             result = _prefer_fuller_snapshots_for_sidebar(result)
             sidebar_candidates = result
             visible_result = [s for s in sidebar_candidates if not _hide_from_default_sidebar(s)]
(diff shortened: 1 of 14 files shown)


santastabber:
Addressed the Greptile P2 about the `TypeError` fallback being too broad.

Changes in the updated head:
- narrowed the compatibility shim to only fall back for the expected `unexpected keyword argument` + `include_lineage_metadata` case
- added regression coverage that preserves the legacy monkeypatch fallback but verifies an internal `TypeError` mentioning `include_lineage_metadata` is propagated instead of hidden

Verification run locally:
- `./scripts/test.sh tests/test_session_list_long_history_perf.py tests/test_session_lineage_metadata_api.py -q` → 15 passed
- `./scripts/test.sh tests/test_performance_polling_hardening.py tests/test_issue1611_session_profile_filtering.py tests/test_session_sidebar_resilience.py tests/test_api_timeout.py tests/test_sidebar_session_partition.py -q` → 42 passed
- `./scripts/test.sh tests/test_session_sidebar_cache.py tests/test_request_diagnostics_cache.py tests/test_issue1855_request_diagnostics.py tests/test_issue2157_sessions_list_stale_stream_state.py tests/test_session_attention_badges.py tests/test_session_index.py tests/test_issue3238_orphaned_cli_sidecar_prune.py tests/test_issue4385_cron_archive_reappears.py -q` → 86 passed
- `python3 -m py_compile api/routes.py tests/test_session_list_long_history_perf.py`
- `git diff --check`

nesquena-hermes (changes requested):
Thanks for the perf work on the session-list path — the archived-history short-circuit is a real win, but the current implementation drops a correctness step that the default sidebar depends on, so this needs a rework before it can land.

## Blocker 1 — default sidebar silently drops WebUI sessions with stale-CLI sidecar flags (production regression)

The route now calls `all_sessions(diag=diag, include_lineage_metadata=False)` (`api/routes.py:1826`), which skips `_enrich_sidebar_lineage_metadata`. That enrichment is the **only** place that applies the state.db source-of-truth correction:

```python
# api/models.py:3215-3220
if state_db_source == 'webui':
    session['source_tag'] = state_db_source_tag
    ...
    session['is_cli_session'] = False
```

With enrichment skipped, the route sees the **uncorrected, stale-CLI** source for those rows. The default sidebar (`show_cli_sessions` defaults to `False`) then runs the CLI filter:

```python
# api/routes.py:1936
webui_sessions = [s for s in webui_sessions if not _is_cli_session_for_settings(s)]
```

which **drops** any WebUI session whose JSON sidecar still carries stale CLI flags. The PR's late re-enrichment (`_enrich_sidebar_lineage_metadata(scoped)`, `api/routes.py:1991`) runs *after* that filter — too late to rescue the row.

This is exactly the scenario the existing test `test_state_db_webui_source_overrides_stale_cli_json_metadata` documents (WebUI sidecar with `is_cli_session=True`/`source_tag=cli`, but state.db `source=webui`). That test still passes because it calls `all_sessions()` directly (default `include_lineage_metadata=True`); the route no longer matches that guaranteed behavior.

Reproduced deterministically: with a stale-CLI WebUI sidecar + state.db `source=webui` and `show_cli_sessions=False`, the new route path returns `[]`, while the pre-PR path returns `['stale_cli_webui']`. Same input, divergent output.

**Fix-spec (cheapest first):** keep skipping the expensive lineage chain-walk, but still apply the cheap per-id source correction on the **full** candidate set *before* the `routes.py:1936` filter. Split the source/title override out of `_enrich_sidebar_lineage_metadata` (`models.py:3215-3228`) into a helper that runs unconditionally (it only needs the keyed `_state_db_source*` fields — no chain walk), and call it inside `all_sessions` even when `include_lineage_metadata=False`, or in the route right after the sessions are fetched and before line 1936. (Simpler alternative: move `_enrich_sidebar_lineage_metadata` to run on the full `webui_sessions` set before the CLI filter rather than on `scoped` at 1991 — but that gives back some of the perf.) Please add a regression guard for the stale-CLI/default-sidebar case.

## Blocker 2 — new test is flaky / red in a natural run order

`tests/test_session_lineage_metadata_api.py::test_sessions_route_preserves_visible_child_lineage_when_archived_parent_filtered` fails `assert [] == ['lineage_api_visible_tip']`:

- Deterministically **RED** when the two new files run together: `pytest tests/test_session_list_long_history_perf.py tests/test_session_lineage_metadata_api.py` → 1 failed, 12 passed (re-confirmed).
- **Flaky even alone** (~1 in 5 runs failed).

Root cause: the `_isolate` fixture never writes `_index.json`, so `all_sessions` spawns the background `_start_session_index_rebuild_thread` (`models.py:173`) which races the full-scan read; nothing stubs that thread, so the test is non-hermetic.

**Fix-spec:** make the fixture hermetic — pre-seed an empty `_index.json` so `all_sessions` doesn't spawn the rebuild thread, and/or monkeypatch `models._start_session_index_rebuild_thread` to a no-op (or join/disable it). Consider the same for the perf file's handler-driven tests.

---

Note on coverage: `test_session_list_long_history_perf.py` mocks `all_sessions` and `_enrich_sidebar_lineage_metadata`, so it structurally cannot catch Blocker 1 — a real regression guard for the stale-CLI default-sidebar case is still needed. Happy to re-review once the source correction runs before the CLI filter and the new test is hermetic.

santastabber:
Addressed the two review blockers in the updated head.

Changes:
- split the cheap state.db source/title sidebar correction out from full lineage enrichment
- `all_sessions(..., include_lineage_metadata=False)` now still applies that cheap state.db correction before `/api/sessions` performs CLI filtering, without reintroducing the lineage parent/child walk on the hot path
- added a route-level regression for a stale-CLI JSON sidecar whose state.db row says `source=webui` with `show_cli_sessions=False`
- made the lineage test fixture hermetic by pre-seeding `_index.json` and stubbing the index rebuild thread

Verification run locally:
- `./scripts/test.sh tests/test_session_list_long_history_perf.py tests/test_session_lineage_metadata_api.py -q` → 16 passed
- repeated `tests/test_session_lineage_metadata_api.py::test_sessions_route_preserves_visible_child_lineage_when_archived_parent_filtered` 5x → all passed
- `./scripts/test.sh tests/test_performance_polling_hardening.py tests/test_issue1611_session_profile_filtering.py tests/test_session_sidebar_resilience.py tests/test_api_timeout.py tests/test_sidebar_session_partition.py -q` → 42 passed
- `./scripts/test.sh tests/test_session_sidebar_cache.py tests/test_request_diagnostics_cache.py tests/test_issue1855_request_diagnostics.py tests/test_issue2157_sessions_list_stale_stream_state.py tests/test_session_attention_badges.py tests/test_session_index.py tests/test_issue3238_orphaned_cli_sidecar_prune.py tests/test_issue4385_cron_archive_reappears.py -q` → 86 passed
- `python3 -m py_compile api/models.py api/routes.py tests/test_session_lineage_metadata_api.py tests/test_session_list_long_history_perf.py`
- `git diff --check`

nesquena-hermes (changes requested):
Thanks for the quick turnaround — re-gated the re-push (`5c8e4966`) against the two findings from the last review.

## ✅ Finding A (the production regression) is genuinely fixed

`all_sessions(..., include_lineage_metadata=False)` now applies the cheap per-id state.db correction via `_apply_sidebar_state_db_overrides(result)` **before returning** (`api/models.py` index path ~3424 / full-scan path ~3470), sharing the extracted `_apply_sidebar_state_db_override_metadata` (`source=='webui' ⇒ is_cli_session=False`). So `webui_sessions` arrives at the route already-corrected, *ahead* of the default CLI filter — a WebUI session carrying stale-CLI sidecar flags is no longer dropped from the sidebar. The new route-level regression test `test_sessions_route_keeps_state_db_webui_row_with_stale_cli_json_when_cli_hidden` passes deterministically and locks this in. Nicely done.

## ❌ Finding B (the PR's own new test) is still red — and the fix targeted the wrong race

`test_sessions_route_preserves_visible_child_lineage_when_archived_parent_filtered` still fails **~50% of co-runs** with the exact original signature:

```
AssertionError: assert [] == ['lineage_api_visible_tip']
E   Right contains one more item: 'lineage_api_visible_tip'
tests/test_session_lineage_metadata_api.py:454
```

Reproduced 6/11 failing across repeated co-runs of the two new files together (`tests/test_session_lineage_metadata_api.py tests/test_session_list_long_history_perf.py`), and intermittently standalone. This is nondeterministic CI-flake territory and shouldn't merge.

The `_isolate` fixture change (seeding `_index.json` + stubbing `_start_session_index_rebuild_thread`) addressed the index-rebuild thread, but that isn't the race. The real cause is a different, un-neutralized module global: the mtime-keyed **`_PERSISTED_SESSION_IDS_CACHE`** behind `models._persisted_session_ids_snapshot()` (`api/models.py:110`).

Each test calls `_save_webui_session` 2–3 times. Every `Session.save()` writes the JSON then populates that snapshot cache keyed on `SESSION_DIR.stat().st_mtime_ns`. When two session files land within one directory-mtime granularity tick, the second call **hits a stale cached snapshot** that omits the just-saved session. Both the prune in `all_sessions()` (`api/models.py` ~3348-3361, `... in persisted_ids`) and the index fast-path prune (`~291-294`, `... in on_disk_ids`) then drop the freshly-saved row → empty list → the assertion fails. `Session.save()` doesn't register into the global `SESSIONS` dict, so the prune relies entirely on that stale snapshot (this is why it's probabilistic on mtime collisions, worse on fast/tmpfs filesystems).

### Precise fix (test-harness hermeticity — product code is fine as-is)

Neutralize that cache in the `_isolate` fixture, e.g. **one** of:

1. Make the snapshot uncached for the test:
   ```python
   monkeypatch.setattr(
       models, "_persisted_session_ids_snapshot",
       lambda: frozenset(
           p.stem for p in models.SESSION_DIR.glob("*.json")
           if not p.name.startswith("_")
       ),
   )
   ```
   (must return a `frozenset`, never `None` — `_write_session_index` does `... in on_disk_ids`); and/or reset `models._PERSISTED_SESSION_IDS_CACHE = (None, None, frozenset())` before and after the test.
2. Register each saved session into `models.SESSIONS` so `in_memory_ids` keeps the row regardless of the disk snapshot.

Then re-run the two files co-located ≥5× and confirm 16/16 every time.

Everything else (the Finding-A correction, the `include_archived` plumbing, `static/sessions.js`) reviewed clean and additive — no walk-back, perf goal preserved. Just the test flake to button up and this is ready.

santastabber:
Addressed the remaining Finding B test-isolation issue in the updated head.

Change:
- `tests/test_session_lineage_metadata_api.py` now neutralizes the mtime-keyed `_PERSISTED_SESSION_IDS_CACHE` in its isolated fixture by monkeypatching `_persisted_session_ids_snapshot()` to an uncached scan of the fixture's `SESSION_DIR` returning a `frozenset`.

Verification run locally:
- `./scripts/test.sh tests/test_session_lineage_metadata_api.py tests/test_session_list_long_history_perf.py -q` repeated 10x → all passed
- `./scripts/test.sh tests/test_session_lineage_metadata_api.py::test_sessions_route_preserves_visible_child_lineage_when_archived_parent_filtered -q` repeated 10x → all passed
- `./scripts/test.sh tests/test_performance_polling_hardening.py tests/test_issue1611_session_profile_filtering.py tests/test_session_sidebar_resilience.py tests/test_api_timeout.py tests/test_sidebar_session_partition.py -q` → 42 passed
- `./scripts/test.sh tests/test_session_sidebar_cache.py tests/test_request_diagnostics_cache.py tests/test_issue1855_request_diagnostics.py tests/test_issue2157_sessions_list_stale_stream_state.py tests/test_session_attention_badges.py tests/test_session_index.py tests/test_issue3238_orphaned_cli_sidecar_prune.py tests/test_issue4385_cron_archive_reappears.py -q` → 86 passed
- `./scripts/test.sh tests/test_session_sidebar_cache.py -q` → 9 passed
- `python3 -m py_compile api/models.py api/routes.py tests/test_session_lineage_metadata_api.py tests/test_session_list_long_history_perf.py`
- `git diff --check`

Also ran a read-only Steward review of this final local patch; verdict: `APPROVE`.

nesquena-hermes:
Re-reviewed the latest head (`e1a3ecf2`) against the one open blocker from my 07:21 review — **Finding B is now genuinely fixed.** This clears my remaining changes-requested item.

## What I checked

The flake root cause I called out was the mtime-keyed module global behind `_persisted_session_ids_snapshot()`:

```python
# api/models.py:109
_PERSISTED_SESSION_IDS_CACHE: tuple[Path | None, int | None, frozenset[str]] = (None, None, frozenset())
```

Both prune sites read from it, so a stale snapshot drops a freshly-saved row:

- index fast-path: `api/models.py:292` — `if (e.get('session_id') in in_memory_ids or e.get('session_id') in on_disk_ids)` where `on_disk_ids = _persisted_session_ids_snapshot()` (`:285`)
- full-scan prune: `api/models.py:3254-3263` — `persisted_ids = _persisted_session_ids_snapshot()` then `str(s.get('session_id')) in persisted_ids`

Your `_isolate` fixture now neutralizes the cache exactly as specced (option 1):

```python
def uncached_persisted_session_ids():
    return frozenset(
        p.stem
        for p in models.SESSION_DIR.glob("*.json")
        if not p.name.startswith("_")
    )
monkeypatch.setattr(models, "_persisted_session_ids_snapshot", uncached_persisted_session_ids)
```

That returns a live `frozenset` per call (never `None`, so the `_write_session_index` `in on_disk_ids` membership at `:292` stays safe) and is read by both prune paths — so a session saved within one directory-mtime tick of a prior save can no longer be missed. The `_index.json` seed + `_start_session_index_rebuild_thread` no-op from the prior round stay in place and are still correct; they just weren't the actual race.

## Finding A spot-check

Confirmed still intact: `all_sessions(..., include_lineage_metadata=False)` applies the cheap per-id correction via `_apply_sidebar_state_db_overrides(result)` before returning (both the index path and full-scan path in `api/models.py`), so `webui_sessions` reaches the route already source-corrected ahead of the default CLI filter. `test_sessions_route_keeps_state_db_webui_row_with_stale_cli_json_when_cli_hidden` locks that in at the route layer.

## Verdict

Both findings resolved, CI is green across all 11 jobs (3.11/3.12/3.13 × 3 shards + lint + browser-smoke), and the `include_archived` plumbing + `static/sessions.js` toggle changes reviewed clean and additive. No further blockers from my side — over to a maintainer for the merge call.

PR #4599, opened 2026-06-21
fix(tts): add id-ID-GadisNeural to Edge TTS and fix Listen button playback

latipun7:
## Summary

Three small changes to improve Edge TTS support:

### Changes

1. **api/routes.py** — Add `id-ID-GadisNeural` to the Edge TTS server voice allowlist so Indonesian voice is available via the /api/tts endpoint.

2. **static/panels.js** — Add `Gadis (Indonesian, Female)` to the Edge TTS voice dropdown in Settings so users can select it.

3. **static/ui.js** — Fix `_playEdgeTtsChunked` not setting `_ttsSpeaking=true` before the fetch chain, causing the Listen button to silently drop audio (the `if(!_ttsSpeaking) return` guard at the start of the blob handler would immediately abort playback).

### Testing

- Verified `edge_tts.Communicate('...', 'id-ID-GadisNeural')` produces valid audio
- Verified /api/tts returns HTTP 200 with correct audio Content-Type
- Verified Listen button plays audio after the ui.js fix (hard refresh required)

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -13437,6 +13437,7 @@ def _handle_tts(handler, parsed):
         "fr-CA-AntoineNeural", "fr-CA-JeanNeural",
         "fr-CA-SylvieNeural", "fr-CA-ThierryNeural",
         "fr-FR-DeniseNeural", "fr-FR-EloiseNeural", "fr-FR-HenriNeural",
+        "id-ID-GadisNeural",
     }
     if voice not in allowed:
         from api.helpers import bad as _bad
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped as **v0.51.556** (Release TO) — live in prod. Indonesian Edge TTS voice (parity addition to the existing builtin) + the Listen-button playing-state fix. 9935 suite, Codex SAFE. Thanks @latipun7.

PR #4601, opened 2026-06-21
Add EvoLink provider integration

EvoLinkAI:
## Thinking Path

- Hermes WebUI already has curated provider registries for onboarding, Settings key management, and model discovery.
- EvoLink exposes an OpenAI-compatible API, so the smallest matching integration is a provider registry entry with its API-key env var, default Base URL, and static fallback models.
- The change stays table-driven and avoids altering runtime routing, UI layout, or provider abstractions.

## What Changed

- Added EvoLink to the provider display/model catalog with GPT-5.2 fallback models.
- Added `EVOLINK_API_KEY` detection so exported keys surface EvoLink in model discovery.
- Added EvoLink onboarding metadata with default Base URL `https://direct.evolink.ai/v1`.
- Added Settings key-management mapping for `EVOLINK_API_KEY`.
- Added EvoLink-focused provider registry tests and scrubbed `EVOLINK_API_KEY` from the isolated test-server environment.
- Added a short Unreleased changelog entry.

## Why It Matters

EvoLink users can configure the provider through the same WebUI provider paths as other OpenAI-compatible services instead of falling back to a generic custom endpoint setup.

## Verification

- `./.venv/bin/python -m py_compile api/config.py api/providers.py api/onboarding.py tests/conftest.py tests/test_evolink_provider.py`
- `./.venv/bin/python - <<'PY' ...` registry assertion check passed for display name, fallback models, onboarding metadata, default Base URL, and Settings env-var mapping.
- Loaded local PR-ops `.env` without printing secrets and checked `https://direct.evolink.ai/v1/models`: HTTP 200, `gpt-5.2` present, 148 models returned.
- Attempted `./scripts/test.sh tests/test_evolink_provider.py -v --timeout=180`; local harness failed before any EvoLink test body executed because the session-wide isolated test server reached “listening” but `/health` stayed HTTP 502 for both 45s boot attempts. Server log showed startup reached `Hermes Web UI listening on http://127.0.0.1:<port>`.

## Risks / Follow-ups

- The WebUI-side registration is intentionally minimal. If Hermes Agent later changes EvoLink model names or provider routing conventions, the static fallback list should be updated to match.
- The local pytest server-health issue appears environment-specific and blocked execution of the new test bodies in this checkout.

## Model Used

OpenAI GPT-5 via Codex CLI, with local shell, git, and GitHub CLI tool use.


nesquena-hermes (changes requested):
Thanks for the update bumping the EvoLink catalog to GPT-5.5. One objective blocker before this can go green: **the PR's own test `tests/test_evolink_provider.py::test_evolink_detected_from_env` fails on the current head** (`b41ecaa8`).

### Reproduced

```
tests/test_evolink_provider.py::test_evolink_detected_from_env FAILED
  assert {"@evolink:gpt-5.5", "@evolink:gpt-5.5-mini"}.issubset(ids)
  where ids == {'gpt-5.5', 'gpt-5.5-mini'}
1 failed, 3 passed
```

### Cause

In this test you set **only** `EVOLINK_API_KEY`, so EvoLink is the **single detected provider**. The `@provider_id:` prefix is applied by `_apply_provider_prefix()` / `_deduplicate_model_ids()` in `api/config.py` **only when there's a collision or a non-active provider** — with a single provider and no model-ID clash, `get_available_models()` correctly returns **bare** IDs (`gpt-5.5`, `gpt-5.5-mini`). That's exactly what your other two assertions (`test_evolink_provider_models_are_registered`, `test_evolink_onboarding_setup_metadata`) already expect with the bare form, and they pass. The namespaced expectation on line 74 was over-corrected when the IDs were bumped 5.2 → 5.5.

### Fix (one line)

`tests/test_evolink_provider.py:74` — revert to the bare form, matching the single-provider detection path:

```python
    assert {"gpt-5.5", "gpt-5.5-mini"}.issubset(ids)
```

After that, all 4 tests in the file pass.

### Heads-up (not a blocker)

`gpt-5.5` / `gpt-5.5-mini` now **collide** with the existing `openai-codex` catalog. The dedup pass handles this (it will `@`-prefix one of the colliding entries in a full config), so it's not a defect — just be aware that in a real multi-provider config the EvoLink IDs may render namespaced in the dropdown. The single-provider test above is unaffected.

PR #4613, opened 2026-06-21
fix(webui): preserve reader scroll position

dso2ng:
## Summary

Fixes scroll jumps/interference while reading long WebUI sessions after manually scrolling away from the bottom.

This keeps manual reader/unpinned mode sticky across background transcript updates and avoids iOS Safari touch/momentum scroll interference.

## Changes

- Preserve manual unpin across idle/non-streaming `renderMessages()` updates instead of falling back to `scrollToBottom()`.
- Route queue chip/pill layout updates through `scrollIfPinned()` instead of forcing bottom scroll while idle.
- Restore long-session refreshes using a stable message viewport anchor key instead of raw array index / pixel offset alone.
  - This handles tail-window -> larger/full-session reloads without jumping to the wrong message.
- Sync `_oldestIdx` / `_messagesTruncated` in `refreshSession()` after same-session reloads.
- Defer active-session external refresh while the reader is manually unpinned; flush it after explicit scroll-to-bottom.
- Add iOS/touch momentum protection:
  - Defer endless-scroll older-message prefetch during recent touch scroll.
  - Re-check manual/touch unpin before delayed bottom-settle writes `scrollTop`.

## Verification

Focused scroll/touch regression suite:

```text
141 passed, 3 warnings in 6.08s
```

Command:

```bash
python -m pytest \
  tests/test_issue1360_streaming_scroll_hardening.py \
  tests/test_issue1690_scroll_completion.py \
  tests/test_issue3479_ios_stream_scroll_jump.py \
  tests/test_tars_scroll_reset_regressions.py \
  tests/test_streaming_markdown.py \
  tests/test_issue3545_streaming_scroll_cue.py \
  tests/test_issue500_message_list_virtualization.py \
  tests/test_session_endless_scroll.py \
  tests/test_issue3470_touch_unpin_streaming_scroll.py \
  tests/test_issue1731_upward_scroll_unpins.py -q
```

Also verified manually in a live WebUI session:

- Idle `refreshSession()` tail-window -> full-session reload preserves the same visible message (`sameKey=true`, `sameTextPrefix=true`).
- iPad Safari manual scroll-back behavior is now normal after hard reload / fresh JS.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -421,6 +421,10 @@ function _scheduleActiveSessionIdleReload(sid) {
   setTimeout(async () => {
     if(!S||!S.session||S.session.session_id !== sid) return;
     if(S.busy || S.activeStreamId) return;
+    if(typeof _isMessageReaderUnpinned==='function'&&_isMessageReaderUnpinned()){
+      _deferActiveSessionExternalRefresh('idle-reconcile');
+      return;
+    }
     try{
       await loadSession(sid, {force:true, externalRefreshReason:'idle-reconcile'});
     }catch(_){}
@@ -1128,6 +1132,9 @@ async function loadSession(sid){
   // Same-session force refreshes reuse the current transcript viewport; clearing
   // the sticky-unpin state here makes preserveScroll treat a reader mid-answer
   // as pinned and snap them back to the bottom on the next render.
+  if (currentSid !== sid) {
+    _clearDeferredActiveSessionExternalRefresh();
+  }
   if (currentSid !== sid && typeof window !== 'undefined' && typeof window._resetScrollDirectionTracker === 'function') {
     try { window._resetScrollDirectionTracker(); } catch (_) {}
   }
@@ -3807,6 +3814,7 @@ let _streamingPollTimer = null;
 let _sessionTimeRefreshTimer = null;
 let _activeSessionExternalRefreshTimer = null;
 let _activeSessionExternalRefreshInFlight = false;
+let _deferredActiveSessionExternalRefreshReason = '';
 let _sessionEventsSSE = null;
 let _sessionEventsRefreshTimer = 0;
 let _sessionEventsReconnectTimer = 0;
@@ -3844,10 +3852,31 @@ function ensureSessionTimeRefreshPoll(){
   }, _sessionTimeRefreshMs);
 }
 
+function _deferActiveSessionExternalRefresh(reason){
+  const nextReason = reason || 'poll';
+  if(_deferredActiveSessionExternalRefreshReason==='idle-reconcile'&&nextReason==='poll') return;
+  _deferredActiveSessionExternalRefreshReason = nextReason;
+}
+
+function _clearDeferredActiveSessionExternalRefresh(){
+  _deferredActiveSessionExternalRefreshReason = '';
+}
+
+function _flushDeferredActiveSessionExternalRefresh(){
+  const reason = _deferredActiveSessionExternalRefreshReason;
+  if(!reason) return;
+  _deferredActiveSessionExternalRefreshReason = '';
+  void refreshActiveSessionIfExternallyUpdated(reason);
+}
+
 async function refreshActiveSessionIfExternallyUpdated(reason){
   if(_activeSessionExternalRefreshInFlight) return;
   if(!S.session || !S.session.session_id) return;
   if(S.busy || S.activeStreamId) return;
+  if(typeof _isMessageReaderUnpinned==='function'&&_isMessageReaderUnpinned()){
+    _deferActiveSessionExternalRefresh(reason||'poll');
+    return;
+  }
   // #3916: the 30s timer is only a fallback for imported/external sessions.
   // WebUI-native sessions should not keep probing forever when the sidebar SSE
   // is healthy, but they still must reconcile when an actual sessions_changed
--- a/static/ui.js
+++ b/static/ui.js
@@ -734,6 +734,35 @@ function _messageVisibleIndexForRawIdx(rawIdx, visWithIdx){
   }
   return -1;
 }
+function _messageViewportAnchorKeyForMessage(m){
+  if(typeof _compressionMessageAnchorKey!=='function') return '';
+  const key=_compressionMessageAnchorKey(m);
+  if(!key) return '';
+  return [key.role||'',key.ts??'',key.attachments??0,key.text||''].map(v=>encodeURIComponent(String(v))).join('|');
+}
+function _messageVisibleIndexForAnchorKey(anchorKey, visWithIdx){
+  const key=String(anchorKey||'');
+  if(!key) return -1;
+  const list=Array.isArray(visWithIdx)?visWithIdx:_getVisibleMessagesWithIdx();
+  for(let i=0;i<list.length;i++){
+    if(list[i]&&_messageViewportAnchorKeyForMessage(list[i].m)===key) return i;
+  }
+  return -1;
+}
+function _messageSessionIndexBase(){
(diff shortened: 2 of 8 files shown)


nesquena-hermes:
Pulled the branch (`4c323b70`) and read the full `static/ui.js` + `static/sessions.js` diff against `origin/master`. The reader-preservation core is solid, the three-tier anchor restore is a real improvement, and the touch-momentum deferral is well-scoped. Two confirmations of greptile's notes plus one cross-PR interaction the maintainer should sequence deliberately.

## What the PR does well

The sticky-unpin preservation in the idle `renderMessages()` path is correct: the snapshot is now captured when `preserveScroll || _messageUserUnpinned` (`ui.js:10842`), and the idle branch restores instead of snapping to bottom (`ui.js:10817-10821`):

```js
if(_messageUserUnpinned){
  _restoreMessageScrollSnapshot(scrollSnapshot);
  _maybeShowNewMessageScrollCue(scrollSnapshot);
  return;
}
scrollToBottom();
```

The three-tier anchor restore (`data-message-anchor-key` → `data-session-msg-idx` → raw `data-msg-idx`, `ui.js:769-836`) is the right fix for tail-window → full-session reloads landing on the wrong message, since the absolute session index survives the window base shift via `_messageSessionIndexBase()`.

## Cross-PR interaction with #4584 (please sequence)

This is the one worth a maintainer's eye before merge. PR #4613 leaves the scroll-handler `movedDown&&nearBottom` branch **identical to master** (`ui.js:3834-3841`):

```js
}else if(movedDown&&nearBottom){
  _nearBottomCount=_nearBottomCount+1;
  if(_nearBottomCount>=2){
    _scrollPinned=true;
    _messageUserUnpinned=false;   // unconditional clear of sticky unpin
  }
}
```

PR #4584 (#4295 fix) changes this exact branch: after its follow-up on `e76a5197`, it only clears the sticky unpin when `bottomDistance<=80` (the true-bottom tail) rather than anywhere in the 250px `nearBottom` window — that's the whole point of #4295. So whichever of #4584/#4613 lands second will hit a conflict here, and a naive resolution that keeps #4613's `_messageUserUnpinned=false` would silently reopen #4295's content-growth re-pin hijack. These are independent branches (both forked off master, neither contains the other), so they don't auto-coordinate. Recommend merging #4584 first, then rebasing #4613 and dropping its copy of this branch in favor of #4584's gated version.

## greptile P1 confirmed — unguarded cross-file call

`sessions.js:424` and `sessions.js:3867` call `_isMessageReaderUnpinned()` (defined in `ui.js:3618`) without the `typeof fn === 'function'` guard every other cross-file call in that file uses (e.g. `sessions.js:1135` guards `window._resetScrollDirectionTracker`). In production both scripts always load, so it's benign today, but it's inconsistent with the file's defensive pattern and would throw if `sessions.js` ever loads/tests in isolation. Wrap both:

```js
if(typeof _isMessageReaderUnpinned==='function' && _isMessageReaderUnpinned()){ ... }
```

## greptile P2 confirmed — deferred reason not cleared on session switch

`_deferredActiveSessionExternalRefreshReason` (`sessions.js:3814`) is only cleared inside `_flushDeferredActiveSessionExternalRefresh()` (`sessions.js:3859`), which fires from `scrollToBottom()` (`ui.js:4468`). `loadSession()` resets the scroll tracker via `_resetScrollDirectionTracker()` (`sessions.js:1135`) but does not clear this deferred reason. So: unpin in session A → poll defers a refresh → switch to session B → scroll to bottom in B → the stale reason flushes a `refreshActiveSessionIfExternallyUpdated('idle-reconcile')` against B. `refreshActiveSessionIfExternallyUpdated` re-checks `S.session.session_id` so it won't corrupt state, but it does fire a spurious metadata request. Clear it in `_resetScrollDirectionTracker()` or at the top of `loadSession()`.

## Tests

The added `test_tars_scroll_reset_regressions.py` / `test_session_endless_scroll.py` updates are source-structural string checks (consistent with the repo's existing scroll-test pattern), and the queue-chip tests now assert `scrollToBottom()` is *absent*, which matches routing through `scrollIfPinned()`. The PR reports 141 passed locally. Neither greptile item nor the #4584 interaction is exercised behaviorally, so a small JS-runtime test for "session-switch clears the deferred reason" would lock P2 down.

dso2ng:
Quick status update after the latest push:

- Addressed the reviewed `sessions.js` items:
  - guarded both `_isMessageReaderUnpinned()` cross-file calls
  - clear deferred active-session refresh reason on session switch
  - preserve `idle-reconcile` over lower-priority `poll` while deferred
- Latest CI is green across browser-smoke, lint, and the Python matrix.
- Local focused scroll verification on the updated head: `143 passed`.

One sequencing note: #4584 and this PR both touch the scroll re-pin branch in `static/ui.js`. I think the lowest-risk order is to merge #4584 first, then rebase this PR so its `bottomDistance <= 80` true-bottom re-pin guard is preserved here.

nesquena-hermes:
Thanks for the quick turnaround — pulled the updated head (`03343e04b`) and confirmed all three review items landed:

- Both `_isMessageReaderUnpinned()` cross-file calls are now guarded (sessions.js:424, sessions.js:3876).
- The deferred reason is cleared on real session switch — `_clearDeferredActiveSessionExternalRefresh()` at sessions.js:1136, gated on `currentSid !== sid`, ahead of the `_resetScrollDirectionTracker()` call. Clean placement.
- The priority preservation is in: `idle-reconcile` is no longer downgraded by a later `poll` (sessions.js:3857).

## The #4584 sequencing question — it's now resolved, and it changes what you need to do

You and I both flagged "merge #4584 first, then rebase this." That happened a few hours ago: **#4584 merged into master via `fc3d6b68e`** ("stage batch #4584+#4583", 2026-06-22 06:15 UTC). Your last push to this branch was 2026-06-22 00:45 UTC — *before* that merge — so this PR is built on a pre-#4584 master and `git merge-base --is-ancestor fc3d6b68e <head>` returns false.

That matters because the re-pin branch you're carrying is the **old ungated** version (ui.js:3868-3873 on this branch):

```js
}else if(movedDown&&nearBottom){
  _nearBottomCount=_nearBottomCount+1;
  if(_nearBottomCount>=2){
    _scrollPinned=true;
    _messageUserUnpinned=false;     // clears sticky unpin anywhere in the 250px nearBottom window
  }
}
```

whereas origin/master now has #4584's **gated** version (ui.js:3821-3828 on master):

```js
}else if(movedDown&&nearBottom){
  _nearBottomCount=_nearBottomCount+1;
  if(_nearBottomCount>=2){
    if(!_messageUserUnpinned||bottomDistance<=80){   // only clear at the true-bottom tail
      _messageUserUnpinned=false;
    }
    ...
    _nearBottomCount=0;
```

I grepped this branch and the `bottomDistance<=80` gate is **not present anywhere** in your `ui.js`. So a straight merge of this PR as-is would overwrite #4584's guard with the unconditional `_messageUserUnpinned=false` and silently reopen #4295 — the content-growth re-pin hijack #4584 just fixed.

## Recommendation

Rebase onto current master and drop this branch's copy of the `movedDown&&nearBottom` block in favor of master's gated one. The two changes are complementary: #4584 owns *when* sticky-unpin clears, #4613 owns *what happens* to the viewport while unpinned (snapshot/restore + the three-tier anchor). After the rebase, the re-pin branch should read exactly as master's, and your idle-render preservation at ui.js:10756-10765 layers on top of it cleanly. Worth re-running the focused scroll suite after the rebase since the conflict region is the one your `143 passed` exercises most.

Everything else here is merge-ready; the rebase is the only blocker.

Issue #3238, opened 2026-05-31
WebUI does not sync when CLI sessions are deleted outside WebUI


Luxciax:
# WebUI does not sync when CLI sessions are deleted outside WebUI

## Summary

Hermes WebUI can keep showing CLI/imported sessions after those sessions have been deleted from the command line / local Hermes CLI storage. This is confusing because WebUI currently does not provide an obvious delete action for these sessions, so deleting from CLI/local storage seems like the natural workaround, but the WebUI sidebar does not synchronize with that external deletion.

## Environment

- OS: Windows
- Hermes WebUI repo: `nesquena/hermes-webui`
- WebUI version: `v0.51.171`
- Branch: `master`
- State dir: `%LOCALAPPDATA%\hermes`
- Setting: `show_cli_sessions: true`

## What happened

1. WebUI showed CLI sessions in the sidebar.
2. I could not find a delete button/action for those WebUI sidebar sessions.
3. I deleted the session from the command line / local CLI-side Hermes session data instead.
4. After that, WebUI still showed the deleted sessions in the sidebar.
5. Restarting WebUI or manually cleaning WebUI-side metadata was needed before the sidebar reflected the deletion.

Observed local storage locations that can keep stale session visibility:

- `%LOCALAPPDATA%\hermes\webui\sessions\<session_id>.json`
- `%LOCALAPPDATA%\hermes\webui\sessions\_index.json`
- `%LOCALAPPDATA%\hermes\state.db`
- `%LOCALAPPDATA%\hermes\webui\sessions\_run_journal`
- `%LOCALAPPDATA%\hermes\webui\sessions\_turn_journal`

## Expected behavior

If a CLI session is deleted outside WebUI, WebUI should eventually stop showing it, or provide a clear refresh/resync action.

Good possible outcomes:

- WebUI detects that the backing CLI session no longer exists and removes the stale sidebar entry.
- WebUI provides a visible delete action for CLI/imported sessions and cleans all related WebUI metadata.
- WebUI provides a "resync CLI sessions" or "prune missing sessions" action.

## Actual behavior

WebUI can keep showing the session because it has its own cached/imported metadata and index files. From the user's point of view, the session was deleted in CLI/local storage but still appears in WebUI.

## Why this matters

The current behavior makes it hard to know which storage is authoritative. If the WebUI has no visible delete button for these sessions, users may reasonably delete them from CLI/local storage, but then WebUI looks stale or broken.

## Suggested fix

[...]

PR #4619, opened 2026-06-21
fix: prune orphaned api sidecars

### LINKED ISSUE #4591: API-server sidecar sessions cannot be deleted and accumulate indefinitely
**Describe the bug**

When the Hermes WebUI sidebar displays API-server-originated sessions (source=`api_server`, created via Hermes Gateway by Open WebUI or similar OpenAI-compatible frontends), clicking on these sessions creates a WebUI-owned sidecar copy in the WebUI's session store. These sidecar sessions are marked as "read-only imported sessions" and cannot be deleted through the WebUI — the `/api/session/delete` endpoint returns a 400 error ("Read-only imported sessions cannot be deleted from WebUI").

Additionally, if the backing API session is deleted from the Hermes state.db (e.g., via `hermes sessions delete` CLI), the WebUI sidecar persists indefinitely. The orphan cleanup logic in `_build_session_list_cache_payload` only handles CLI-sourced orphan sessions — API-sourced sidecars are never pruned.

**To Reproduce**

1. Set up Hermes Gateway (`hermes gateway run`) and connect Open WebUI as a client
2. Have a conversation through Open WebUI (this creates an `api-*` session in state.db)
3. Open Hermes WebUI (http://localhost:8787) — the API session appears in the sidebar as "Api_Server Session"
4. Click the API session in the sidebar (this triggers the WebUI to create an internal sidecar copy)
5. Try to delete it via the WebUI UI — sidecar is read-only, deletion is blocked
6. Alternatively, delete the session from CLI (`hermes sessions delete api-xxx`) and refresh WebUI — the sidecar persists in the sidebar

**Expected behavior**

- API-server sessions displayed in the sidebar should be deletable from the WebUI, OR
- Sidecar sessions should be automatically pruned when the backing session is removed from state.db

**Actual behavior**

- API sidecar sessions are permanently stuck with no UI affordance to remove them
- They accumulate in the sidebar over time (in my case, 87+ stale sessions)

**Technical context**

- WebUI version: v0.51.545
- The sidecar/import mechanism creates a `Session` object in the WebUI's in-memory `SESSIONS` dict and persists a JSON file in the sessions directory
- The delete handler at `api/routes.py:10701` checks `_lookup_cli_session_metadata(sid).get("read_only")` and rejects deletion for imported sessions
- The orphan cleanup at `api/routes.py:1824-1883` (PR #3238) only handles CLI-sourced sidecars, not API-sourced ones

**Suggested fix direction**

Either:
1. Allow deleting imported sessions from the WebUI (remove the read_only guard), with the handler cleaning up both the WebUI sidecar and the backing state.db row
2. Extend the orphan cleanup logic to also handle API-server sourced sidecars that no longer have a backing row in state.db
3. Prevent the creation of sidecar copies for API sessions that are already served directly via `get_cli_sessions()`

franksong2702:
## Thinking Path

Fixes #4591.

The #3238 orphan-prune path only probed rows classified as CLI sessions. API-server sidecars are intentionally excluded from `is_cli_session_row()`, so they were retained forever after the backing Agent `state.db` row disappeared. The delete endpoint still treats imported sidecars as read-only, so manual deletion could not clean them up either.

## What Changed

- Add an API-server sidecar predicate in `api/routes.py` without changing `is_cli_session_row()`.
- Feed API/API-server imported sidecars into the existing `agent_session_rows_existing()` orphan probe.
- Keep native WebUI-owned sessions protected even when stale metadata mentions API.
- Preserve the safe failure behavior of the existing probe: DB errors still degrade to “assume present”.
- Add route-level regression coverage for API sidecar prune, API sidecar retention when backing row exists, and native WebUI safety.

## Why It Matters

Read-only API/Gateway sidecars should not accumulate indefinitely after their backing Agent session is deleted outside WebUI. This reuses the existing exact state-db existence check instead of weakening the read-only delete guard.

## Verification

- `./scripts/test.sh tests/test_issue3238_orphaned_cli_sidecar_prune.py tests/test_issue4385_cron_archive_reappears.py tests/test_ruff_forward_lint.py -q`
- `git diff --check`

## Risks / Follow-ups

- Low-to-medium risk because this touches session-list pruning. Tests cover the missing-row, present-row, and native-WebUI safety cases.
- This does not address the separate question from the issue comment about whether duplicate sidecars are being created on every click; it only fixes orphan cleanup.

## Model Used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,6 @@
 
 ## [Unreleased]
 
-### Fixed
-
-- **API-server imported sidecars are now pruned when their backing Agent session is gone (#4591).** Read-only API/Gateway sidecar rows no longer accumulate indefinitely in the sidebar after the corresponding `state.db` session is deleted outside WebUI, while native WebUI sessions remain protected.
-
 ## [v0.51.560] — 2026-06-21 — Release TS (recovered run-journal output reaches the model)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **API-server imported sidecars are now pruned when their backing Agent session is gone (#4591).** Read-only API/Gateway sidecar rows no longer accumulate indefinitely in the sidebar after the corresponding `state.db` session is deleted outside WebUI, while native WebUI sessions remain protected.
+
 ## [v0.51.560] — 2026-06-21 — Release TS (recovered run-journal output reaches the model)
 
 ### Fixed
--- a/api/routes.py
+++ b/api/routes.py
@@ -1827,12 +1827,12 @@ def _build_session_list_cache_payload(
         cli = get_cli_sessions(source_filter=source_filter, all_profiles=all_profiles)
         diag_stage("merge_cli_sessions")
         cli_by_id = {s["session_id"]: s for s in cli}
-        # #3238: reconcile orphaned imported-CLI sidecars. When a CLI
-        # session was clicked in WebUI it gets a WebUI-owned sidecar that
-        # all_sessions() returns independently of state.db. If the user
-        # later deletes the backing CLI session from the command line,
-        # the sidecar is never pruned and the stale row lingers in the
-        # sidebar forever (there is no WebUI delete affordance for it).
+        # #3238/#4591: reconcile orphaned imported sidecars. When a CLI or
+        # API-server session is clicked in WebUI it gets a WebUI-owned sidecar
+        # that all_sessions() returns independently of state.db. If the user
+        # later deletes the backing agent session outside WebUI, the sidecar is
+        # never pruned and the stale row lingers in the sidebar forever (there
+        # is no WebUI delete affordance for read-only imported rows).
         # Drop rows whose backing agent row is genuinely gone. We probe
         # state.db directly (agent_session_rows_existing) rather than trust
         # cli_by_id absence, because get_cli_sessions() caps at
@@ -1846,7 +1846,7 @@ def _build_session_list_cache_payload(
             _sid = s.get("session_id")
             if (
                 _sid
-                and is_cli_session_row(s)
+                and (is_cli_session_row(s) or _is_api_server_sidecar_row(s))
                 and not _session_source_is_webui(s)
                 and _sid not in cli_by_id
             ):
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
## Summary

This implements exactly the orphan-prune direction (#2) we recommended on #4591: a second predicate that feeds API-server sidecars into the *same* `agent_session_rows_existing()` probe rather than weakening the `read_only` delete guard. Reading `api/routes.py:1830-1890` on the PR HEAD against `origin/master`, plus `api/agent_sessions.py:165-200` and `api/models.py:3123-3172`, the layering holds and the safety properties are preserved. Good, surgical change.

## Code reference

The new entry condition correctly OR-s the API predicate in without touching `is_cli_session_row()` (`api/routes.py:1849`):

```python
if (
    _sid
    and (is_cli_session_row(s) or _is_api_server_sidecar_row(s))
    and not _session_source_is_webui(s)
    and _sid not in cli_by_id
):
    _orphan_probe_rows.append(s)
```

And the predicate is double-guarded against WebUI-native rows — once inside the helper and again at the call site (`api/routes.py:6105-6112`):

```python
def _is_api_server_sidecar_row(session: dict) -> bool:
    if not isinstance(session, dict) or _session_source_is_webui(session):
        return False
    markers = {_normalized_source_marker(session.get(key)) for key in (...)}
    return bool(markers & {"api", "api_server"})
```

## Verification I did

I traced the three failure-safety properties that matter for a session-list prune:

1. **Transient DB error never mass-prunes** — `agent_session_rows_existing()` returns `frozenset(wanted)` on any exception, missing DB, or missing `sessions` table (`api/models.py:3140-3172`), so a probe failure degrades to "assume present / keep." Preserved.
2. **Window-cap false positives avoided** — the probe is uncapped against `state.db` rather than trusting `cli_by_id` absence (which caps at `CLI_VISIBLE_SESSION_LIMIT`). A live API session served by the gateway stays in `state.db`, so it's retained. Your `test_api_server_sidecar_with_backing_state_row_is_retained` nails this.
3. **WebUI-native safety** — `test_webui_owned_session_with_api_metadata_is_not_pruned` stamps `source_tag="webui"` with stale `raw_source="api_server"` and confirms `_session_source_is_webui()` short-circuits the prune. Correct.

`_normalized_source_marker` also handles the literal sidebar label cleanly: `"Api_Server Session"` → lower → strip `" session"` → `api_server`, which lands in the match set.

## One open item (not a blocker)

This fixes the *accumulation* half of #4591 but, as you note in the PR body, not the *creation* half — direction #3 (don't fork a sidecar when the API row is already served via `get_cli_sessions()`). That's the same question I raised on #4591 re: the reporter's 87+ rows. Worth a follow-up issue so the count stops growing in the first place, but the prune fix here is correct to land on its own.

nesquena-hermes:
Shipped in **v0.51.565** 🚀 — thank you @franksong2702 (batched with #4621).

The orphan-prune now covers API-server imported sidecars, not just CLI ones, so a stale read-only sidecar no longer lingers in the sidebar after its backing agent session is deleted outside WebUI. The widening is guarded two ways — it never touches WebUI-owned sessions (`_session_source_is_webui`, double-checked), and it only prunes when the state.db probe confirms the backing row is genuinely absent (the probe fails closed: a transient DB error keeps every row).

Review trail: Codex regression gate **SAFE TO SHIP**, Opus advisor **SAFE — ship** (both confirmed no live session can be dropped + the marker-matching is self-limiting to genuine orphans), full suite **9996 passed**.

PR #4620, opened 2026-06-21
fix: harden isolated profile opt-in snapshot

### LINKED ISSUE #4590: Harden isolated-mode opt-in: snapshot HERMES_WEBUI_ISOLATED_PROFILE so a profile .env can't silently disable isolation (follow-up to #4586)
## Follow-up to #4586 / PR #4589 — harden isolated-mode opt-in against silent disable

PR #4589 fixed the #4586 regression by gating isolated profile mode on an explicit `HERMES_WEBUI_ISOLATED_PROFILE` opt-in (default off). During that PR's review, Codex + Opus both flagged a **deferred** hardening item (Nathan's call: defer, since #2698 shipped only days ago and there are effectively no real multi-user isolated deployments to protect yet — the regression fix was the priority).

### The gap (deferred, not a blocker)
`_isolated_profile_opt_in()` reads `HERMES_WEBUI_ISOLATED_PROFILE` from `os.environ` **live** on every call. But `_reload_dotenv()` folds a pinned profile's `.env` into `os.environ` at startup. So in a genuine multi-user isolated deployment (`HERMES_WEBUI_ISOLATED_PROFILE=1` set by the launcher), a profile whose own `.env` contains `HERMES_WEBUI_ISOLATED_PROFILE=0` can **silently turn its own isolation OFF** after launch — a cross-profile-access escape for a user who controls their profile `.env`. Codex verified this in a fresh process.

### Proposed fix
1. **Snapshot the opt-in at import time** (beside `_INITIAL_HERMES_HOME`) so a profile `.env` reload cannot override the process-level deployment control. (`_INITIAL_HERMES_HOME` is already snapshotted for the identical reason.)
2. **One-time startup warning** for the "HERMES_HOME matches `*/profiles/<name>` but the flag is unset" case, so a deployment that *intended* isolation and upgrades past v0.51.528 isn't silently downgraded without a signal. (Opus's observability ask.)
3. Regression test: `init_profile_state()` / `_reload_dotenv()` cannot disable isolation from a profile `.env` when the launcher env enabled it.

### Priority
Low-urgency right now (no real isolated deployments exist yet), but should land **before** isolated mode gains real multi-user adoption. Tracking so it isn't lost.

Refs #4586, #4589, #2698.

franksong2702:
## Thinking Path

Fixes #4590.

The issue is a deployment-posture bug: isolated profile mode already snapshots startup `HERMES_HOME`, but the opt-in flag still depended on live `os.environ`. A pinned profile `.env` is loaded later into the same process, so the profile-owned environment must not be able to change whether WebUI is isolated.

## What Changed

- Snapshot `HERMES_WEBUI_ISOLATED_PROFILE` at module import/startup next to `_INITIAL_HERMES_HOME`.
- Make `_isolated_profile_opt_in()` read the startup snapshot instead of live `os.environ`.
- Add a one-time warning when startup `HERMES_HOME` has the `*/profiles/<name>` shape but isolated mode was not explicitly enabled.
- Update isolated-profile regression tests so they patch the startup snapshot intentionally.
- Add regression coverage proving a profile `.env` cannot disable isolation even if it mutates live `os.environ`.

## Why It Matters

Isolated profile mode is a trust-boundary setting. It should be controlled by the launcher/operator, not by a profile-local `.env` file loaded after startup.

## Verification

- `./scripts/test.sh tests/test_issue4586_named_profile_not_isolated.py tests/test_issue2698_isolated_hermes_home.py tests/test_profile_env_isolation.py tests/test_ruff_forward_lint.py -q`
- `git diff --check`

## Risks / Follow-ups

- Low risk. The default remains non-isolated unless the explicit startup opt-in is enabled.
- The warning is diagnostic only and emitted once per process.

## Model Used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,6 @@
 
 ## [Unreleased]
 
-### Fixed
-
-- **Isolated profile mode now uses the launcher-time opt-in snapshot instead of live profile-mutated environment.** A pinned profile `.env` can no longer silently disable a deployment that started with `HERMES_WEBUI_ISOLATED_PROFILE=1`, even if live `os.environ` is later changed. Profile-shaped `HERMES_HOME` launches without the opt-in now emit a one-time warning that isolation is off and normal multi-profile switching remains enabled. Thanks @nesquena-hermes.
-
 ## [v0.51.560] — 2026-06-21 — Release TS (recovered run-journal output reaches the model)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Isolated profile mode now uses the launcher-time opt-in snapshot instead of live profile-mutated environment.** A pinned profile `.env` can no longer silently disable a deployment that started with `HERMES_WEBUI_ISOLATED_PROFILE=1`, even if live `os.environ` is later changed. Profile-shaped `HERMES_HOME` launches without the opt-in now emit a one-time warning that isolation is off and normal multi-profile switching remains enabled. Thanks @nesquena-hermes.
+
 ## [v0.51.560] — 2026-06-21 — Release TS (recovered run-journal output reaches the model)
 
 ### Fixed
--- a/api/profiles.py
+++ b/api/profiles.py
@@ -33,14 +33,16 @@ _PROFILE_DIRS = [
 ]
 _CLONE_CONFIG_FILES = ['config.yaml', '.env', 'SOUL.md']
 
-# ── Snapshot initial HERMES_HOME before init_profile_state() mutates it ──────
-# _is_isolated_profile_mode() needs to detect whether HERMES_HOME *at startup*
-# points to a profile subdirectory (e.g., ~/.hermes/profiles/user1), not the
-# base home (~/.hermes). But init_profile_state() overwrites HERMES_HOME to the
-# base home, which would disable isolation detection. Snapshot it here at import
-# time (before init_profile_state runs) and use the snapshot in the detector.
+# ── Snapshot startup env before profile init / dotenv reload mutates it ───────
+# _is_isolated_profile_mode() needs startup HERMES_HOME, not the value after
+# init_profile_state() rewrites it. The opt-in flag is also an operator-level
+# startup control: a pinned profile's .env may be loaded into live os.environ
+# later, but must not be able to change whether the process is isolated.
 _INITIAL_HERMES_HOME = os.getenv('HERMES_HOME', '').strip()
+_INITIAL_ISOLATED_PROFILE_OPT_IN = os.getenv('HERMES_WEBUI_ISOLATED_PROFILE', '').strip().lower()
 _ISOLATED_SYMLINK_WARNING_EMITTED = False
+_ISOLATED_PROFILE_SHAPE_WITHOUT_OPT_IN_WARNING_EMITTED = False
+_ISOLATED_PROFILE_TRUTHY_VALUES = frozenset({'1', 'true', 'yes', 'on'})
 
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
## Summary

The core hardening is correct: snapshotting `HERMES_WEBUI_ISOLATED_PROFILE` at import (`_INITIAL_ISOLATED_PROFILE_OPT_IN`) next to the existing `_INITIAL_HERMES_HOME` snapshot, and reading the snapshot in `_isolated_profile_opt_in()`, makes the opt-in a true launcher/operator posture that a later profile `.env` reload into live `os.environ` cannot flip. Reading `api/profiles.py` on the PR HEAD vs `origin/master`, the snapshot follows the established pattern exactly, and the new regression in `test_issue4590...` is the right shape — it deliberately blanks `_PROTECTED_ENV_KEYS` mid-test and mutates live env to prove the snapshot, not the dotenv filter, is now the authority. Nice belt-and-suspenders.

## One concrete bug — the new warning helper's shape check diverges from the real gate

`_warn_if_profile_shape_without_isolated_opt_in()` decides whether `HERMES_HOME` "looks pinned" with a *different* shape test than the function that actually engages isolation. The helper (`api/profiles.py`, new):

```python
p = Path(hermes_home).expanduser()
if p.parent.name != 'profiles' or not p.name:
    return
```

But `_is_isolated_profile_mode()` requires both the `profiles` parent **and** a grandparent-exists guard (`api/profiles.py:218`):

```python
if p.parent.name == 'profiles' and p.parent.parent.exists():
    return True
```

So for a `HERMES_HOME` like `/nonexistent/profiles/foo` (parent named `profiles`, grandparent missing), the helper warns "isolated profile mode stays off" — but isolation was *never going to engage anyway* even with the flag set, because the grandparent-exists guard fails. The warning implies "you forgot the flag" when the real situation is "this path can't be isolated regardless." That's a misleading diagnostic on exactly the misconfiguration the warning is meant to clarify.

## Recommendation

Mirror the gate's shape predicate so the warning only fires when the *only* missing piece is the opt-in. A 3-line fix in the helper:

```python
p = Path(hermes_home).expanduser()
if not (p.parent.name == 'profiles' and p.name and p.parent.parent.exists()):
    return
```

This keeps the warning honest: it fires precisely when `HERMES_HOME` is a real `*/profiles/<name>` path that *would* isolate if the operator had set the flag. Otherwise the symlink path that `_is_isolated_profile_mode()` handles (`p.is_symlink()` branch at `api/profiles.py:220-230`) also never reaches this warning, which is fine since that branch has its own dedicated message.

## Test plan

The existing `test_profile_shape_without_opt_in_warns_once` uses a real `named_profile_home` fixture (grandparent exists), so it stays green either way. Worth adding one negative case: a profile-shaped path whose grandparent does **not** exist should emit **no** warning after the fix — that pins the divergence so it can't silently regress.

nesquena-hermes:
Shipped in **v0.51.566** 🚀 — thank you @franksong2702.

The isolated-profile opt-in is now snapshotted at module import, so a pinned profile's `.env` (loaded into the live environment later in the same process) can no longer flip the isolation posture. The existing `_reload_dotenv` protected-key filter and runtime-env projection are kept as defense-in-depth, and there's a one-time warning when `HERMES_HOME` looks pinned but isolation wasn't opted in at startup.

Review trail: Codex regression gate **SAFE TO SHIP** (verified the snapshot runs before any `.env` reload by module-import ordering, normal multi-profile switching unaffected, isolation filters untouched), Opus advisor **SAFE to merge** ("closes the live-`os.environ` flip vector cleanly, correct by module-ordering construction"), full suite **9988 passed**.

One optional, non-blocking follow-up Opus noted: the startup warning's `*/profiles/<name>` shape-check could over-fire for a grandparent-missing path — a diagnostic-quality polish (log message only), not a security issue. Happy to take that as a tiny follow-up if you'd like it tightened.

PR #4621, opened 2026-06-21
fix: explain insecure microphone origins

### LINKED ISSUE #4571: Voice dictation shows "Microphone access denied" on insecure (HTTP) origins even when permission is granted — should explain HTTPS requirement
### Summary

Voice dictation fails on mobile with the toast **"Microphone access denied. Check browser permissions."** even after the user has explicitly granted microphone permission to the site in their browser. The message is misleading: the real blocker in this case is that the page is served over an **insecure (HTTP) origin**, where browsers disable `SpeechRecognition` / `getUserMedia` regardless of any granted permission.

### Where this lives

This is **WebUI**, not the native Android shell. The native shells already bridge the WebView mic permission correctly (Android `onPermissionRequest` + `RECORD_AUDIO`). This report is the **mobile-browser / PWA** path: when the web app is opened directly in mobile Chrome (or as a PWA) over a plain-HTTP server URL — e.g. a `http://` Tailscale/LAN address — voice input can never work, but the UI blames browser permissions.

### Repro

1. Serve WebUI over plain HTTP (e.g. `http://<tailscale-or-lan-host>:8787`).
2. Open it in mobile Chrome and grant the site microphone permission.
3. Tap **Dictate**.
4. Toast: "Microphone access denied. Check browser permissions." — but permission *was* granted.

### Root cause

`static/boot.js` maps the SpeechRecognition `not-allowed` error straight to the generic denial string:

```js
const msgs={
  'not-allowed':t('mic_denied'),   // "Microphone access denied. Check browser permissions."
  'no-speech':t('mic_no_speech'),
  'network':t('mic_network'),
};
```

On an insecure origin, Chrome reports `not-allowed` for SpeechRecognition (and `getUserMedia` is unavailable because `navigator.mediaDevices` is gated on `window.isSecureContext`). There is **no `isSecureContext` check anywhere in the voice code**, so an HTTPS-context problem is reported as a permissions problem.

### Proposed fix

Detect the insecure-context case up front and show an accurate, actionable message instead of the generic denial:

- If `!window.isSecureContext` (and origin is not `localhost`/`127.0.0.1`, which browsers treat as secure), short-circuit before starting capture and toast something like: *"Voice input needs a secure connection. Open Hermes over HTTPS (or via localhost) to use the microphone."* (add an i18n key, e.g. `mic_insecure_context`).
- Optionally hide/disable the Dictate button entirely on an insecure non-localhost origin, the same way it's already hidden when neither SpeechRecognition nor MediaRecorder is available.

### Acceptance

- On an insecure non-localhost origin, the mic UI explains the HTTPS requirement rather than blaming browser permissions.
- HTTPS and localhost behavior is unchanged.
- New string is added to all locales (or falls back gracefully).

franksong2702:
## Thinking Path

Fixes #4571.

Browsers can surface insecure-origin microphone failures as permission-style Web Speech errors. The UI was treating those as a generic permission denial, which is misleading when the page is opened over non-local HTTP.

## What Changed

- Add shared microphone-origin helpers in `static/boot.js`.
- Detect insecure non-local `http:` origins while preserving HTTPS, `localhost`, `.localhost`, IPv4 loopback, and IPv6 loopback.
- Route dictation and hands-free voice mode permission-style failures through the insecure-origin message when appropriate.
- Preflight insecure origins before starting SpeechRecognition or MediaRecorder capture.
- Add `mic_insecure_origin` to all locale bundles, using English TODO fallbacks where translations are not already available and a real `zh-Hant` translation to satisfy the existing Traditional Chinese locale guard.
- Preserve the previous secure-origin `audio-capture` behavior by falling through to `mic_error + raw error` instead of mapping hardware capture failures to permission denied.
- Add regression tests for the insecure-origin branch, localhost/loopback preservation, shared error routing, and the `audio-capture` fallthrough.

## Why It Matters

Users should not be told browser permission was denied when the actual fix is to use HTTPS or localhost. This makes the failure actionable and preserves the normal permission-denied behavior on secure/local origins.

## Verification

- `node` behavior probe against the extracted helper logic for remote HTTP, localhost, `.localhost`, `127.0.0.1`, `[::1]`, HTTPS, and `audio-capture` secure/insecure routing.
- `./scripts/test.sh tests/test_issue4571_mic_insecure_origin.py tests/test_sprint20.py::test_boot_js_not_allowed_error_message tests/test_zh_hant_locale.py -q`
- `git diff --check`

## Risks / Follow-ups

- No live browser microphone test was run in this pass; coverage is static plus JS helper behavior and syntax checks.
- Most non-English locale strings use English fallback TODOs where translations were not available; `zh-Hant` is translated because that locale has an explicit no-TODO guard.

## Model Used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,6 @@
 
 ## [Unreleased]
 
-### Fixed
-
-- **Microphone errors on insecure non-local HTTP origins now explain the HTTPS/localhost requirement (#4571).** Browser dictation and hands-free voice mode now distinguish insecure-origin capture failures from normal permission denials, while preserving HTTPS and localhost/loopback behavior.
-
 ## [v0.51.560] — 2026-06-21 — Release TS (recovered run-journal output reaches the model)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Microphone errors on insecure non-local HTTP origins now explain the HTTPS/localhost requirement (#4571).** Browser dictation and hands-free voice mode now distinguish insecure-origin capture failures from normal permission denials, while preserving HTTPS and localhost/loopback behavior.
+
 ## [v0.51.560] — 2026-06-21 — Release TS (recovered run-journal output reaches the model)
 
 ### Fixed
--- a/static/boot.js
+++ b/static/boot.js
@@ -487,7 +487,6 @@ function _micToastKeyForRecognitionError(error){
   const msgs={
     'not-allowed':'mic_denied',
     'service-not-allowed':'mic_denied',
-    'audio-capture':'mic_denied',
     'no-speech':'mic_no_speech',
     'network':'mic_network',
   };
@@ -1000,7 +999,8 @@ window._micPendingSend=window._micPendingSend||false;
       }
       if(event.error==='not-allowed'||event.error==='service-not-allowed'||event.error==='audio-capture'){
         _deactivate();
-        showToast(t(_micToastKeyForRecognitionError(event.error)||'mic_denied'));
+        const messageKey=_micToastKeyForRecognitionError(event.error);
+        showToast(messageKey?t(messageKey):t('mic_error')+event.error);
         return;
       }
       // Other errors — try to restart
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Shipped in **v0.51.565** 🚀 — thank you @franksong2702 (batched with #4619).

Voice dictation now detects insecure non-local `http://` origins and shows a clear "needs a secure connection — open over HTTPS or localhost" message instead of a generic permission error, with a preflight before SpeechRecognition / MediaRecorder / hands-free capture. HTTPS, `localhost`, `.localhost`, and IPv4/IPv6 loopback are all correctly preserved. The `mic_insecure_origin` key is in all 13 locale blocks.

Review trail: Codex regression gate **SAFE TO SHIP**, Opus advisor **SAFE — ship** (loopback detection + preflight coverage of every capture entry point verified, 13/13 locale parity), full suite **9996 passed**.

PR #4637, opened 2026-06-21
feat: allow toggling installed extensions

santastabber:
## Thinking Path
- Extension diagnostics now expose installed manifest entries, assets, and loopback sidecars.
- The extensions roadmap calls for Phase 2 enable/disable management for already-present extensions before any install/uninstall flow.
- The safe boundary is a WebUI-managed override in state: do not edit extension manifests, fetch assets, proxy sidecars, or add extension-owned routes.
- This PR lets operators enable/disable installed manifest entries from Settings → Extensions while preserving manifest-disabled entries as non-toggleable.

## What Changed
- Adds a bounded, fail-safe `extension-overrides.json` state file for UI-managed disabled extension IDs.
- Filters disabled manifest entries out of injected script/style URLs and sanitized sidecar diagnostics.
- Adds `POST /api/extensions/toggle` for authenticated enable/disable updates.
- Extends `/api/extensions/status` with per-extension effective state and `reload_required` metadata.
- Updates Settings → Extensions with installed-extension rows and Enable/Disable controls.
- Documents the Phase 2 boundary in `docs/EXTENSIONS.md`.

## Why It Matters
Operators can suppress a misbehaving local extension from the browser without editing manifests or environment variables, while keeping the trusted-local extension model narrow and explicit.

## Verification
- `./scripts/test.sh tests/test_extension_status_endpoint.py tests/test_extensions_settings_panel.py tests/test_issue803.py::TestProfileCookieHelpers::test_profile_cookie_name_defaults_to_hermes_profile -q`
- `.venv/bin/python -m py_compile api/extensions.py api/routes.py tests/test_extension_status_endpoint.py tests/test_extensions_settings_panel.py`
- Isolated API/browser smoke with a temporary extension bundle: disabling Desktop Companion removed its scripts/styles/sidecar, manifest-disabled entries returned 409, and Settings → Extensions showed Templates, Desktop Companion, and Off rows.

## Risks / Follow-ups
- Toggle changes affect extension injection on browser reload; already-injected extension code remains in the current tab until reload.
- This intentionally does not add install, uninstall, marketplace browsing, or sidecar proxying.

## Contract Routing
Task type: Extension management Phase 2
Touched areas:
- `api/extensions.py`
- `api/routes.py`
- `static/panels.js`
- `docs/EXTENSIONS.md`
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/EXTENSIONS.md`
Scope boundaries:
- WebUI-managed enable/disable overrides for already-present manifest entries only.
- No manifest mutation, install/uninstall, marketplace, proxy, or extension-owned routes.
- Backend state/error/CSRF tests, Settings panel static tests, syntax/lint checks, and isolated API/browser smoke.

### DIFF
--- a/api/extensions.py
+++ b/api/extensions.py
@@ -9,8 +9,10 @@ import html
 import json
 import logging
 import os
+import re
+import threading
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Set, Tuple
 from urllib.parse import unquote, urlsplit
 
 from api.helpers import _security_headers, j
@@ -36,6 +38,14 @@ class _ManifestTooLarge(ValueError):
     pass
 
 
+class ExtensionToggleError(Exception):
+    """Sanitized extension mutation error safe to return to the browser."""
+
+    def __init__(self, message: str, status: int = 400):
+        super().__init__(message)
+        self.status = status
+
+
 EXTENSION_ROUTE_PREFIX = "/extensions/"
 _EXTENSION_DIR_ENV = "HERMES_WEBUI_EXTENSION_DIR"
 _EXTENSION_SCRIPT_URLS_ENV = "HERMES_WEBUI_EXTENSION_SCRIPT_URLS"
@@ -45,6 +55,12 @@ _ALLOWED_ASSET_PREFIXES = ("/extensions/", "/static/")
 _SIDECAR_WARNING_SOURCE = "manifest:sidecars"
 _DEFAULT_SIDECAR_HEALTH_PATH = "/health"
 _LOOPBACK_SIDECAR_HOSTS = {"127.0.0.1", "localhost", "::1"}
+_EXTENSION_STATE_FILENAME = "extension-overrides.json"
+_MAX_EXTENSION_STATE_BYTES = 32 * 1024
+_MAX_DISABLED_EXTENSION_IDS = 512
+_EXTENSION_ID_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
+_EXTENSION_STATE_WARNING_SOURCE = "extension_state"
+_EXTENSION_STATE_LOCK = threading.Lock()
 
 _EXTENSION_MIME = {
     "css": "text/css",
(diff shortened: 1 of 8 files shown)


santastabber:
Addressed the two Greptile P2 findings in the latest push:

- Removed the unused internal `enabled_ids` field from `_manifest_extension_state`.
- Made the Settings → Extensions installed-extension empty state context-aware when no extension directory is configured.

Verification:

- `./scripts/test.sh tests/test_extension_status_endpoint.py tests/test_extensions_settings_panel.py tests/test_issue803.py::TestProfileCookieHelpers::test_profile_cookie_name_defaults_to_hermes_profile -q` → 50 passed
- `.venv/bin/python -m py_compile api/extensions.py api/routes.py tests/test_extension_status_endpoint.py tests/test_extensions_settings_panel.py`
- `git diff --check`

nesquena-hermes:
Read through the full diff (`api/extensions.py`, `api/routes.py`, `static/panels.js`) at the PR head against `master`. This is a careful, well-scoped Phase 2 — the "WebUI-managed override, never touch the manifest" boundary is held cleanly. A few things I verified that are worth recording, plus two small notes.

## The disable is a complete kill switch — confirmed
The thing I most wanted to check: does disabling an extension also suppress its loopback sidecar, or only its injected scripts/styles? It suppresses everything. `_iter_manifest_entries` filters by `disabled_ids` (`api/extensions.py:130` region), and the sidecar list is built from that *same* filtered `entries` list:

```python
# api/extensions.py — _read_manifest_urls_with_diagnostics
entries = _iter_manifest_entries(manifest, disabled_ids=disabled_ids)
...
for _source, entry in entries:
    if _source.startswith("manifest.extensions["):
        sidecar = _sidecar_from_manifest_entry(entry, diagnostics)
```

So a disabled extension drops out of scripts, stylesheets, *and* the sidecar health-probe set in one pass — no orphaned proxy left running. Good.

## Auth + write-safety check out
The mutating route is gated: `handle_post` runs behind `check_auth(self, parsed)` at `server.py:362` (the only carve-out is `/api/csp-report`), so `/api/extensions/toggle` requires the same auth as every other admin POST. The state write is atomic with an fsync + same-dir `os.replace` and a process/thread-unique temp name (`_write_extension_state`), and `set_extension_user_enabled` does the read-modify-write under `_EXTENSION_STATE_LOCK` — no torn writes under concurrent toggles. ID validation (`_EXTENSION_ID_RE`, 128-char cap), the 32KB state-file cap, and the 512-ID cap all fail safe to an empty-state default. Manifest-disabled entries are correctly non-toggleable (`manifest_disabled_ids` → 409 in `set_extension_user_enabled`), which matches the `can_toggle:false` the panel renders.

## Two small notes
1. **`get_extension_status()` after the lock releases.** `set_extension_user_enabled` returns `get_extension_status()` *outside* the `with _EXTENSION_STATE_LOCK` block, so the returned snapshot is a fresh re-read rather than the just-written state under the lock. In practice that's fine (the file was already fsync'd + replaced before the lock dropped), but a concurrent toggle of a *different* ID could make the returned `user_disabled` count reflect both writes. Harmless given the UI re-renders from this payload, just worth a one-line comment so a future reader doesn't "fix" it by moving the call inside the lock and reintroducing the manifest re-parse cost under contention.

2. **Stale-ID surfacing.** `_manifest_extension_state` emits `extension_state_unknown_ids` when `disabled_ids - known_ids` is non-empty (an extension was disabled, then removed from the manifest). The override is preserved (not pruned), which is the right call for re-install — but consider whether the panel should surface that warning so an operator knows why a re-added extension comes back disabled. Not a blocker.

Tests look thorough (`test_extension_status_endpoint.py` +439). Per the EXECUTION BAN I did not run them, but the assertions I read match the behavior above. Solid PR.

santastabber:
Addressed the two follow-up notes in the latest push:

- Added a comment explaining why `set_extension_user_enabled()` returns a fresh status snapshot after releasing `_EXTENSION_STATE_LOCK`.
- Surfaced `extension_state_unknown_ids` as an operator-facing warning in Settings → Extensions so stale disabled-extension overrides are understandable without exposing raw paths or ids.

Verification:

- `./scripts/test.sh tests/test_extension_status_endpoint.py tests/test_extensions_settings_panel.py tests/test_issue803.py::TestProfileCookieHelpers::test_profile_cookie_name_defaults_to_hermes_profile -q` → 50 passed
- `.venv/bin/python -m py_compile api/extensions.py api/routes.py tests/test_extension_status_endpoint.py tests/test_extensions_settings_panel.py`
- `git diff --check`

PR #4638, opened 2026-06-21
perf(sidebar): cap lineage enrichment to top-N sessions

maksym-mishchenko:
## Symptom

In our deployment with ~2400 sessions in `state.db`, `GET /api/sessions` was averaging 5–17 seconds wall-clock. Slow-request diagnostics consistently showed `all_sessions.lineage_metadata` as the dominant stage:

```json
{
  "path": "/api/sessions",
  "elapsed_ms": 16834.9,
  "stages": [
    {"ms": 0.6, "name": "all_sessions.read_index"},
    {"ms": 4924.2, "name": "all_sessions.lineage_metadata"},
    {"ms": 10266.1, "name": "reconcile_stale_stream_state"},
    ...
  ]
}
```

## Root cause

`_enrich_sidebar_lineage_metadata()` walks parent/descendant chains in `state.db` for **every** visible sidebar row. With 2400+ session rows, that's a multi-batch IN-clause walk that hits the index but still pays linear cost in N. Lineage data for older sessions is almost never visible to the user — the sidebar paints chronologically newest-first and older entries get scrolled out or hidden behind the history panel.

## Fix

Cap the DB lookup to the top-N (default 300) most-recent sessions. The cap is configurable via the `HERMES_WEBUI_LINEAGE_TOP_N` environment variable so operators can A/B or roll back without a redeploy. Setting it to `0` disables the cap entirely and restores prior behavior.

The chosen default of 300 covers "recent enough that the user might scroll to it without opening history" while bounding the worst case. Tail sessions still get their lineage enriched lazily when the user explicitly opens the history panel — that code path is unchanged.

## Expected effect

In our deployment: 4.9s → ~0.6s on the `lineage_metadata` stage (88% reduction in DB work, since 300/2398 ≈ 12.5%). For users with <300 sessions, behavior is unchanged.

## Test plan

```
pytest \ tests/test_pr1370_lineage_metadata_perf_and_orphan.py \ tests/test_session_lineage_metadata_api.py \ tests/test_issue1494_state_db_fd_leak.py \ tests/test_session_lineage_collapse.py \ tests/test_issue1855_request_diagnostics.py
68 passed
```

Also a quick functional check that verifies the cap actually slices and the env override works (cap=5 → 5 ids passed, cap=0 → all ids passed, no env → default 300).

## Risk

Low. Pure read-path optimization; no schema or DB-write changes. The collapse contract is preserved for the top-N visible sessions, which is what the sidebar actually paints. Worst case for users with <300 sessions: no change at all (the cap branch is a no-op). Worst case for users with >300 sessions: a lineage-enriched chunk in the tail that the sidebar never displays would need to be enriched lazily when the user opens the history panel (which already has its own loading code path). Disabling via env var = full rollback.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -3200,11 +3200,30 @@ def _sidebar_title_is_generic_webui(title: str | None) -> bool:
 
 
 def _enrich_sidebar_lineage_metadata(sessions: list[dict]) -> None:
-    """Attach state.db compression lineage metadata used by sidebar collapse."""
+    """Attach state.db compression lineage metadata used by sidebar collapse.
+
+    Cap the DB lookup to the top-N most recent sessions to bound wall-clock
+    on power users with thousands of sessions. The sidebar paints chronologically
+    newest first; older sessions almost never have visible lineage to collapse
+    (parents are themselves stale and rarely surface in the same render).
+    Lineage enrichment for those is loaded lazily when the user opens the
+    history panel. Issue #38914 / 2026-06-21 triage: /api/sessions was spending
+    4.9s on lineage_metadata across 2400+ rows.
+    """
+    # 2026-06-21: configurable via env to ease A/B and rollback without a redeploy.
+    import os as _os
+    try:
+        _cap = int(_os.environ.get("HERMES_WEBUI_LINEAGE_TOP_N", "300"))
+    except (TypeError, ValueError):
+        _cap = 300
+    if _cap > 0 and len(sessions) > _cap:
+        candidates = sessions[:_cap]
+    else:
+        candidates = sessions
     try:
         metadata = read_session_lineage_metadata(
             _active_state_db_path(),
-            {str(s.get('session_id')) for s in sessions if s.get('session_id')},
+            {str(s.get('session_id')) for s in candidates if s.get('session_id')},
         )
     except Exception:
         return


nesquena-hermes:
Thanks for the detailed write-up and diagnostics — the slow-stage JSON makes this easy to reason about. I pulled the branch and read the diff against `master`, plus both call sites and the DB walk it feeds.

## What the change does
The cap is applied in `_enrich_sidebar_lineage_metadata` (`api/models.py:3202`), slicing `sessions[:_cap]` before the `read_session_lineage_metadata` lookup while the post-lookup enrichment loop still iterates the full list (`if sid in metadata` guards skip the uncapped tail cleanly).

## Correctness check — the slice is sound
The cap only selects the *most-recent* sessions if the list is already newest-first. It is, at both call sites:

```python
# index path — api/models.py:3340
result = sorted(index_map.values(),
    key=lambda s: (s.get('pinned', False), _session_sort_timestamp(s)), reverse=True)
...
_enrich_sidebar_lineage_metadata(result)   # :3360

# full-scan fallback — api/models.py:3389
out.sort(key=lambda s: (getattr(s, 'pinned', False), _session_sort_timestamp(s)), reverse=True)
...
_enrich_sidebar_lineage_metadata(result)   # :3401
```

`reverse=True` puts `pinned=True` ahead of `pinned=False`, then newest timestamp first within each band, so `sessions[:300]` = all pinned + the freshest unpinned. That's exactly the set the sidebar paints, so collapse grouping for visible rows is preserved.

One thing worth calling out that strengthens the change: `read_session_lineage_metadata` (`api/agent_sessions.py:803`) chases `parent_session_id` chains *outward* from the seed IDs (the hop loop at ~`:850`). So an ancestor that lives beyond the cap still gets fetched as a parent of a top-N seed — the cap shrinks the seed set, not the lineage walk, so `_sidebar_lineage_root_id` (`api/models.py:2620`) still resolves the right root for visible sessions.

## One edge case to note (low impact)
A session beyond the cap that is *still rendered* via the discoverability/background passes — `_include_project_hidden_background_sidebar_sessions` (`api/models.py:2653`) and `_preserve_messageful_sidebar_discoverability` — won't get its state.db overlay (`display_title`, `source_tag`, `_lineage_root_id`). For a cron session surfaced by a project chip past row 300, that means a possibly-stale title until the history panel loads it lazily. Acceptable for the perf win, but worth a line in the docstring so a future reader knows the tail-visibility passes run on un-enriched rows.

## On Greptile's two nits
Both are cosmetic and accurate: `import os as _os` is redundant (`os` is imported module-level at `api/models.py:8`), and the `except (TypeError, ValueError)` `TypeError` arm is unreachable since `os.environ.get(name, "300")` always returns a string and `int("...")` raises `ValueError`, never `TypeError`. Neither blocks merge — just drop the `_os` alias and use the module-level `os`.

## Verdict
Logic is correct and the escape hatch (`HERMES_WEBUI_LINEAGE_TOP_N=0` → full restore) is a clean rollback. Pure read-path, no schema change. The docstring note above is the only thing I'd add before merge.

PR #4646, opened 2026-06-22
fix(streaming): pin cronjob profile caches during chat runs

### LINKED ISSUE #4580: In-chat `cronjob` tool does not respect active WebUI profile (named profiles)
# Bug: In-chat `cronjob` tool does not respect active WebUI profile (named profiles)

## Summary

When a user chats in Hermes WebUI under a **named profile** (not the root/default profile), creating or managing cron jobs via the agent **`cronjob` tool** can target the **wrong profile’s** cron store, config, credentials, and delivery bindings. The **Tasks** panel cron API (`/api/crons/*`) was designed for per-request profile isolation; the **in-session tool path** does not appear to have the same guarantees.

## Environment

- **Hermes WebUI** with multiple Hermes profiles (cookie / per-tab profile selection)
- **Hermes Agent** `cronjob` tool enabled in WebUI chat sessions
- Reproducible when the active UI profile is a named profile (e.g. `ops`, `media`) while the WebUI process may still hold a process-default `HERMES_HOME`

## Steps to reproduce

1. Start WebUI and select a **named profile** in the UI (verify chat/skills/memory behave as that profile).
2. In chat, ask the agent to schedule a recurring task using the **`cronjob` tool** (`action=create`, with schedule, prompt, and delivery such as a messaging platform target).
3. Do **not** pass an explicit `profile` argument on the tool call (typical agent behavior).
4. After creation, inspect where the job was stored and how it runs:
   - List jobs from the **named profile’s** `cron/jobs.json` vs the **default/root** profile’s `cron/jobs.json`, and/or
   - Observe delivery/runtime using the **default profile’s** gateway bindings instead of the named profile’s.

**Contrast (expected path):** Create the same job from the **Tasks** UI while the named profile is active, or use `hermes -p <profile> cron create ...` from CLI — job storage and execution align with the selected profile.

## Expected behavior

Cron mutations from a WebUI chat session should use the **same active profile** as the session (the profile stamped on the session / selected in the UI), consistent with:

- `/api/crons/*` handlers that enter `cron_profile_context()` for TLS/cookie-active profile home, and
- Other WebUI streaming fixes that set `HERMES_HOME` (and related module patches) for the session’s profile during agent runs.

Creating a job without an explicit per-job `profile` field should still **store** the job under the active profile’s cron registry and **run** with that profile’s config, `.env`, and delivery targets unless the user intentionally sets a different execution profile.

## Actual behavior

- Jobs created via **`cronjob` in chat** may be written to another profile’s `cron/jobs.json` (often the process default / root profile).
- Jobs may run or deliver using the **scheduler/server default** profile’s gateway and secrets when `job.profile` is unset, even though the user was working in a named profile in the UI.
- Operators may only notice incorrect Discord/Telegram/etc. delivery or missing jobs in the named profile’s Tasks list.

## Technical notes (for maintainers)

**Likely gap (WebUI):** Chat streaming sets session-scoped `HERMES_HOME` and patches some modules (e.g. skills, MCP discovery timing), but `cron.jobs` uses **module-level** paths (`HERMES_DIR`, `JOBS_FILE`, …) snapshotted at import time. The REST cron API explicitly repatches those constants inside `cron_profile_context()` / `cron_profile_context_for_home()`; the **agent tool dispatch path** during `_run_agent_streaming` does not obviously wrap `cronjob` calls in the same context.

**Hermes Agent behavior (not necessarily a separate bug):** `tools/cronjob_tools.py` only sets `job["profile"]` when the model supplies `profile=`. Omitted `profile` means “scheduler default at run time,” which is correct for CLI/gateway when `HERMES_HOME` already matches the intended profile — but is ambiguous in a multi-profile WebUI process.

**Prior related work:** Profile-scoped cron for HTTP API and manual run subprocess isolation (e.g. issues/tests around cron profile selector, scheduled jobs profile isolation, cron profile lock). This report is specifically about **parity for the in-chat `cronjob` tool**.

## Suggested direction (non-prescriptive)

- Wrap `cronjob` tool execution in the same profile isolation used for `/api/crons/*` (env + `cron.jobs` / `cron.scheduler` module patches), **or**
- Auto-stamp `profile` (and/or ensure `JOBS_FILE` points at the session profile home) on `create` when the WebUI session has a non-null profile and the tool did not specify one.

Regression tests could extend the existing cron profile isolation tests to cover a simulated agent `cronjob(action='create')` under `set_request_profile(<named>)` without relying on the REST handlers alone.

## Workaround

Until fixed, use **`hermes -p <profile> cron create ...`** (CLI) or the **Tasks** panel with the correct profile selected, and avoid scheduling from chat via `cronjob` when using named profiles.

roian6:
## Thinking Path

- Hermes WebUI chat runs already switch `HERMES_HOME` to the active profile before invoking the agent.
- The in-chat `cronjob` tool reaches `cron.jobs`, whose `HERMES_DIR` / `CRON_DIR` / `JOBS_FILE` / `OUTPUT_DIR` paths are snapshotted at import time.
- The REST cron paths already use a locked profile context that repatches and restores cron module state for one cron operation.
- This PR keeps the streaming fix narrow and race-aware: it wraps the registered model-facing `cronjob` tool so only that tool call enters the existing cron profile context, instead of mutating cron module globals for the entire agent turn.

## What Changed

- Added a streaming contextvar that records the current chat run's resolved profile home.
- Added a small wrapper around the registered agent `cronjob` tool handler. When the contextvar is set, the wrapper enters `cron_profile_context_for_home(profile_home)` for the duration of that cronjob tool call, then lets the existing context manager restore cron module state immediately.
- Kept the existing process-env fallback and skill-module patching path for the streaming run, but no longer patch `cron.jobs` / `cron.scheduler` caches for the whole turn.
- Added regression coverage in `tests/test_scheduled_jobs_profile_isolation.py` for the cronjob wrapper, the no-profile/default path, and the streaming seam that avoids long-lived cron cache mutation.

## Why It Matters

Issue #4580 reports a parity gap between the Scheduled Jobs REST/API paths and the in-chat `cronjob` tool. Without a cron profile context, a chat run in a named profile can set `HERMES_HOME` correctly but still let `cron.jobs.save_jobs()` write to the process-default profile because the module-level paths were captured earlier.

The fix is intentionally scoped at the `cronjob` tool-call boundary. That reuses the existing cron-specific lock/restore behavior, so concurrent streams do not hold another profile's cron path globals for the full duration of an agent run.

## Contract Routing

Task type: runtime/profile-state bug fix

Touched areas:
- `api/streaming.py`
- `tests/test_scheduled_jobs_profile_isolation.py`

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`

State layer affected: agent-run runtime environment / profile-scoped tool state. The source of truth remains the active WebUI profile home; this PR makes the in-chat `cronjob` tool enter the same locked cron profile context used by the REST cron paths.

Scope boundaries:
- Does not change the REST `/api/crons/*` profile context behavior.
- Does not redesign multi-profile concurrency; it only scopes the model-facing `cronjob` tool to the existing cron profile context.
- Does not change cron job schema or UI behavior.

## Verification

- `HERMES_WEBUI_TEST_PYTHON=/home/ubuntu/.local/bin/python3.11 ./scripts/test.sh tests/test_scheduled_jobs_profile_isolation.py -q`
  - `11 passed in 8.05s`
- `.venv/bin/python -m compileall -q api/streaming.py tests/test_scheduled_jobs_profile_isolation.py`

## Risks / Follow-ups

- This relies on the agent registry's built-in `cronjob` handler being the model-facing handler used by chat runs. The wrapper imports `tools.cronjob_tools` if needed before checking the registry entry.
- If future model-facing cron tools are split out under new names, those handlers should be wrapped with the same profile-context pattern.
- The regression test keeps the full agent run out of this unit file, but covers the handler wrapper behavior and the streaming source seam that prevents full-turn cron cache mutation.

Release-note wording if useful for a release batch:

> In-chat `cronjob` tool calls now stay pinned to the active WebUI profile. The streaming path wraps the model-facing cronjob tool in the existing locked cron profile context, keeping cron path-cache mutation scoped to the single cron call instead of the whole agent turn.

## Model Used

AI-assisted via Hermes Agent using OpenAI GPT-5.5, with terminal/file/GitHub CLI tool use for investigation, testing, review response, and PR preparation.

Closes #4580

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -4,6 +4,7 @@ Includes Sprint 10 cancel support via CANCEL_FLAGS.
 """
 import base64
 import contextlib
+import contextvars
 import json
 import logging
 import mimetypes
@@ -84,6 +85,78 @@ def _session_payload_with_full_messages(session, *, tool_calls=None):
 _ENV_LOCK = threading.Lock()
 
 _KEYLESS_CUSTOM_API_KEY = "dummy-key"
+
+_STREAMING_CRON_PROFILE_HOME: contextvars.ContextVar[str | None] = contextvars.ContextVar(
+    "webui_streaming_cron_profile_home",
+    default=None,
+)
+_STREAMING_CRONJOB_WRAPPER_INSTALLED = False
+
+
+def _install_streaming_cronjob_profile_wrapper() -> None:
+    """Wrap the agent cronjob tool so calls run under the streaming profile.
+
+    The in-chat agent run already binds per-turn contextvars for other
+    session-scoped state. Cron jobs are special because ``cron.jobs`` snapshots
+    path constants at import time, so the model-facing ``cronjob`` tool must
+    enter the existing WebUI cron profile context at the tool-call boundary.
+    That context uses the cron-specific lock and restores the module caches as
+    soon as the single cron tool call returns, avoiding long-lived global path
+    mutation for the whole agent turn.
+    """
+    global _STREAMING_CRONJOB_WRAPPER_INSTALLED
+    if _STREAMING_CRONJOB_WRAPPER_INSTALLED:
+        return
+    try:
+        from tools.registry import registry
+    except Exception:
+        logger.debug("streaming cronjob wrapper: tools registry unavailable", exc_info=True)
+        return
+
+    entry = registry.get_entry("cronjob")
+    if entry is None:
+        try:
+            import tools.cronjob_tools  # noqa: F401
+        except Exception:
+            logger.debug("streaming cronjob wrapper: cronjob tool import failed", exc_info=True)
+        entry = registry.get_entry("cronjob")
+    if entry is None:
+        logger.debug("streaming cronjob wrapper: cronjob tool not registered")
+        return
+    original_handler = entry.handler
+    if getattr(original_handler, "_webui_streaming_profile_wrapper", False):
+        _STREAMING_CRONJOB_WRAPPER_INSTALLED = True
+        return
+
+    # This relies on the agent tool executor's ``propagate_context_to_thread``
+    # using an unfiltered ``contextvars.copy_context()`` so this WebUI-owned
+    # contextvar reaches the sync cronjob handler even when the tool call runs
+    # on the agent's ThreadPoolExecutor worker.
+    def _profile_scoped_cronjob_handler(args, **kwargs):
+        profile_home = _STREAMING_CRON_PROFILE_HOME.get()
+        if not profile_home:
+            return original_handler(args, **kwargs)
+        from api.profiles import cron_profile_context_for_home
+        with cron_profile_context_for_home(Path(profile_home)):
+            return original_handler(args, **kwargs)
+
+    _profile_scoped_cronjob_handler.__dict__["_webui_streaming_profile_wrapper"] = True
+    _profile_scoped_cronjob_handler.__dict__["_webui_original_handler"] = original_handler
+    registry.register(
+        name=entry.name,
+        toolset=entry.toolset,
+        schema=entry.schema,
+        handler=_profile_scoped_cronjob_handler,
+        check_fn=entry.check_fn,
+        requires_env=entry.requires_env,
+        is_async=entry.is_async,
+        description=entry.description,
+        emoji=entry.emoji,
+        max_result_size_chars=entry.max_result_size_chars,
+        dynamic_schema_overrides=entry.dynamic_schema_overrides,
+    )
+    _STREAMING_CRONJOB_WRAPPER_INSTALLED = True
+
 
 _PERSISTENT_MEMORY_FILES = (
     ("memory", ("memories", "MEMORY.md")),
@@ -5981,6 +6054,7 @@ def _run_agent_streaming(
     # Placed ABOVE the _checkpoint_stop cluster so that cluster stays adjacent
     # to the `try:` (preserves the Issue #765 static-locator invariant).
     _turn_session_identity_tokens = None
+    _streaming_cron_profile_home_token = None
     # Initialised here (before any code that may raise) so the outer `finally`
     # block can safely check `if _checkpoint_stop is not None` even when an
     # exception fires before the checkpoint thread is created (Issue #765).
@@ -6027,6 +6101,7 @@ def _run_agent_streaming(
             )
             _profile_home_path = get_hermes_home_for_profile(getattr(s, 'profile', None))
             _profile_home = str(_profile_home_path)
+            _streaming_cron_profile_home_token = _STREAMING_CRON_PROFILE_HOME.set(_profile_home)
             _profile_runtime_env = get_profile_runtime_env(_profile_home_path)
             _safe_profile_runtime_env = filter_runtime_env_for_gateway_parity(_profile_runtime_env)
         except ImportError:
@@ -6084,6 +6159,7 @@ def _run_agent_streaming(
         # first-time module initialisation (which can be slow) does not
         # block other concurrent sessions waiting on _ENV_LOCK (#2024).
         _prewarm_skill_tool_modules()
+        _install_streaming_cronjob_profile_wrapper()
         # Still set process-level env as fallback for tools that bypass thread-local
         # Acquire lock only for the env mutation, then release before the agent runs.
         # The finally block re-acquires to restore — keeping critical sections short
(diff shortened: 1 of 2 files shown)


roian6:
Updated after the initial bot/review pass surfaced a valid race concern with the first version's full-turn cron cache patch.

Current head (`1ce9874`) no longer patches `cron.jobs` / `cron.scheduler` globals for the whole streaming run. Instead, `api/streaming.py` wraps the registered model-facing `cronjob` tool handler and enters the existing `cron_profile_context_for_home(profile_home)` only for the duration of that single cronjob tool call. I also removed the changelog edit and left release-note wording in the PR body instead.

Re-verified after the update/rebase:

- `HERMES_WEBUI_TEST_PYTHON=/home/ubuntu/.local/bin/python3.11 ./scripts/test.sh tests/test_scheduled_jobs_profile_isolation.py -q` → `11 passed`
- `python3 scripts/ruff_lint.py --diff origin/master` → no new violations on added/modified lines
- `.venv/bin/python -m compileall -q api/streaming.py tests/test_scheduled_jobs_profile_isolation.py`

nesquena-hermes:
Reviewed the diff at HEAD `1ce98743b` against `origin/master`, read `api/streaming.py` (the wrapper + the set/reset seam), `api/profiles.py:589` (`cron_profile_context_for_home`), the three new tests, and traced the agent side. The scoped-at-the-tool-boundary approach is the right call, and I can confirm the mechanism actually works end-to-end — but it leans on a cross-repo contract that's worth making explicit.

## Why the contextvar reaches the tool call (the load-bearing detail)

The wrapper reads `_STREAMING_CRON_PROFILE_HOME.get()` on whatever thread the cronjob handler executes on:

```python
# api/streaming.py
def _profile_scoped_cronjob_handler(args, **kwargs):
    profile_home = _STREAMING_CRON_PROFILE_HOME.get()
    if not profile_home:
        return original_handler(args, **kwargs)
    from api.profiles import cron_profile_context_for_home
    with cron_profile_context_for_home(Path(profile_home)):
        return original_handler(args, **kwargs)
```

You `.set()` it on the streaming worker thread (`streaming.py:6100`), and `agent.run_conversation()` runs synchronously on that same thread (`streaming.py:7305`). The agent dispatches tools on a `ThreadPoolExecutor`, but it wraps every submit with `propagate_context_to_thread()`:

```python
# agent: agent/tool_executor.py:589
f = executor.submit(
    propagate_context_to_thread(_run_tool), i, tc, name, args, parsed_calls[i][3]
)
```

and that helper does a full `contextvars.copy_context()` (`tools/thread_context.py`), which copies **all** contextvars in the parent context — including your `_STREAMING_CRON_PROFILE_HOME`, even though the agent has no knowledge of it. So the var rides the agent's existing propagation into the worker thread, and the wrapper sees it. That's the elegant part of this design, and it's correct as written.

## The contract dependency worth documenting

The whole fix silently depends on `propagate_context_to_thread` doing an unfiltered `copy_context()`. If the agent ever changed that to copy a curated allow-list of contextvars instead of the full context (a plausible future hardening), this wrapper would observe `None`, fall through to `original_handler` with no profile context, and **#4580 would silently regress** — no error, just writes landing in the default profile again. Since the source of truth for this seam lives in the agent (`agent/tool_executor.py` + `tools/thread_context.py`), a one-line comment in `_install_streaming_cronjob_profile_wrapper` naming that dependency would save a future debugging session. Per `docs/CONTRACTS.md`, this is exactly the kind of WebUI↔agent runtime seam worth annotating.

## Two smaller notes

- **Async concern from the bot is moot today.** The agent registers `cronjob` with no `is_async` flag (`tools/cronjob_tools.py:945`, defaulting `is_async=False` at `tools/registry.py:242`), and `dispatch()` calls sync handlers inline (`registry.py:404`). The sync wrapper is correct for the current handler; the follow-up only matters if the handler is ever made a coroutine.
- **First-install race is benign, but the test coverage stops short of the real seam.** The two behavioral tests use a fake registry + a stub `Ctx`, and the third (`test_streaming_profile_home_mutation_avoids_long_lived_cron_cache_patch`) is a source-string assertion. None exercise the actual `copy_context()` propagation across a real thread hop — which is precisely the part that could break from an agent-side change. A test that sets the contextvar, submits the wrapped handler through a real `ThreadPoolExecutor` + `copy_context()`, and asserts `cron_profile_context_for_home` was entered would lock the contract that the source-string test can only approximate.

Net: correct fix, minimal blast radius, reuses the existing locked cron context. The only thing I'd add before merge is the contract comment naming the `copy_context()` dependency. Nice work narrowing this down from the full-turn-patch first draft.

roian6:
Addressed the requested contract seam follow-up and rebased onto current `origin/master`.

Changes in current head (`c3fd6ab`):
- Added a short comment in `_install_streaming_cronjob_profile_wrapper` documenting that this WebUI-owned contextvar relies on the agent tool executor's unfiltered `contextvars.copy_context()` propagation (`propagate_context_to_thread`) across the ThreadPoolExecutor tool hop.
- Added `test_streaming_cronjob_wrapper_context_survives_threadpool_context_copy`, which sets the streaming profile contextvar, copies the context, dispatches the wrapped handler through a real `ThreadPoolExecutor`, and asserts `cron_profile_context_for_home` is entered on the worker.

Re-verified after the rebase/update:
- `HERMES_WEBUI_TEST_PYTHON=/home/ubuntu/.local/bin/python3.11 ./scripts/test.sh tests/test_scheduled_jobs_profile_isolation.py -q` → `12 passed`
- `python3 scripts/ruff_lint.py --diff origin/master` → no new violations on added/modified lines
- `.venv/bin/python -m compileall -q api/streaming.py tests/test_scheduled_jobs_profile_isolation.py`

PR #4652, opened 2026-06-22
fix(#4650): stop /api/reasoning request storm + memoize config YAML parse

### LINKED ISSUE #4650: Performance regresssion /api/reasoning
I updated today and performance is suddenly abysmal.  Last update was a couple of days ago.  So its recent.  It's not the models - triple checked.
It seems to be calling /api/reasoning for each word that is coming in over the wire (which takes 123-500ms for each call on my machine).  This in turn reduces output speed to about 2 tokens per sec.

this is what codex had to say about it:
> syncTopbar() unconditionally calls syncReasoningChip(), and syncReasoningChip() calls fetchReasoningChip(), which performs a network GET /api/reasoning?.... Because syncTopbar() is invoked from many routine UI paths during chat/session rendering, this turns ordinary UI refreshes into repeated reasoning-status/capability requests, producing a request storm unrelated to the actual SSE chat stream.

origin:
> The behaviour appears to have been triggered by a9ce2889 (fix(ui): hide reasoning chip when model lacks effort levels, first in v0.51.141), which changed syncReasoningChip() from using cached/local _currentReasoningEffort after the first load into always calling fetchReasoningChip(). Since syncTopbar() already calls syncReasoningChip() as part of routine UI refresh, this makes ordinary topbar/session sync paths issue repeated GET /api/reasoning?... requests. Recent rendering/streaming changes may have made the issue more visible by increasing UI sync frequency, but the network-fetch-on-every-sync behaviour traces back to that commit.

slow api call:
> GET /api/reasoning is also slow on its own: direct local probes take roughly 140–290ms, and live logs show ~126–700ms. The endpoint calls get_reasoning_status(), which reparses the active config.yaml on every request via _load_yaml_config_file(). Local timing shows the dominant cost is yaml.safe_load(config.yaml), around 124–135ms per call for a 24KB / 811-line config, while file read and env expansion are negligible. So the request storm is amplified by each request doing repeated YAML parsing.

FYI Thanks for this awesome project!  love it.

nesquena-hermes:
## What

Fixes the `/api/reasoning` performance regression in #4650 (streaming output dropping to ~2 tok/s). Two layers, exactly as the issue analysis + maintainer comment recommended:

### Frontend (primary — restores pre-`a9ce2889` behavior)

`syncTopbar()` calls `syncReasoningChip()` on every routine UI refresh, and during streaming those fire at high frequency. Commit `a9ce2889` ("hide reasoning chip when model lacks effort levels", v0.51.141) changed `syncReasoningChip()` from serving the cached `_currentReasoningEffort` to refetching **unconditionally**, turning ordinary syncs into a `GET /api/reasoning` storm (one per token).

This restores the cache short-circuit while preserving `a9ce2889`'s intent (refresh supported-efforts after a model switch):

```js
function syncReasoningChip(){
  const key=_reasoningEffortQuery();           // model+provider identity
  if(_currentReasoningEffort!==null && _lastReasoningFetchKey===key){
    _applyReasoningChip(_currentReasoningEffort);  // serve cache
    return;
  }
  fetchReasoningChip();                          // cold, or model changed
}
```

Only the model/provider identity changes `/api/reasoning`'s answer, so that's the cache key. The user-pick (`POST /api/reasoning`) and model-switch paths still update the cache directly, so a switch refreshes supported-efforts exactly once.

### Backend (defense in depth)

`_load_yaml_config_file()` reparsed `config.yaml` on every call — `yaml.safe_load` on an ~800-line / 24KB config is ~125ms — so each storm request also paid a full YAML parse. Now memoizes the **raw** parsed dict keyed on `(path, st_mtime_ns, st_size)` and re-runs `_expand_env_vars()` per call: env expansion is cheap, always returns a fresh structure (so the ~7 read-modify-save callers never corrupt the cache), and keeps `${VAR}` references live. Any on-disk edit (including by `_save_yaml_config_file`) invalidates via the mtime/size key. Benefits all callers of `_load_yaml_config_file` (`api/config.py:3294/3328/3425/3579`, `api/routes.py:605`).

## Tests

- **`tests/test_issue4650_reasoning_chip_no_storm.py`** — node-driven behavioural test that drives the *actual* `ui.js` functions and **counts network calls**: 1 fetch cold → **0 across 10 same-model syncs (the storm)** → 1 on model switch → 0 after. Pins that the storm cannot silently return.
- **`tests/test_issue4650_yaml_config_cache.py`** — cache hit skips reparse (asserts `safe_load` called once), mtime change triggers a fresh parse, env expansion runs per call (not frozen), caller mutation can't corrupt the cache, missing file returns `{}`.

All 9 new tests pass. Full reasoning/config/profile/yaml blast-radius suite: **946 passed, 0 failures**. `compileall` clean.

## Verification (per the issue)

With DevTools Network filtered to `/api/reasoning`: streaming a response now issues at most one request per model switch, not one per token; token throughput returns to normal.

Closes #4650.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Streaming is fast again — the reasoning chip no longer fires a `GET /api/reasoning` request per token (#4650).** A regression from v0.51.141 made the composer's reasoning-effort chip refetch its status on every routine UI sync, so during a streamed reply the topbar refresh issued one request per token; on configs where that endpoint is slow (a large `config.yaml`), throughput collapsed to ~2 tok/s. The chip now serves its cached state on routine syncs and only refetches when the model/provider changes (or on an explicit pick), and `config.yaml` parsing is memoized (keyed on the file's mtime/size, with env-var expansion still applied live per read) so repeated reads no longer re-parse the whole file. Reported by @batkuip.
+
 ## [v0.51.569] — 2026-06-22 — Release UB (mobile drawer fixes)
 
 ### Fixed
--- a/api/config.py
+++ b/api/config.py
@@ -467,20 +467,55 @@ def reload_config() -> None:
             _delete_models_cache_on_disk()
 
 
+# Memoized parse cache for _load_yaml_config_file, keyed on (resolved path,
+# st_mtime_ns, st_size). yaml.safe_load on an ~800-line / 24KB config.yaml costs
+# ~125ms of pure-Python parsing, and hot read paths (e.g. GET /api/reasoning ->
+# get_reasoning_status) call this on every request. Without a cache, a UI sync
+# storm turns into a YAML-reparse storm (#4650). We cache the RAW parsed dict and
+# re-run _expand_env_vars() on every call: env expansion is cheap, always returns
+# a fresh structure (so callers that read-modify-save the result never corrupt the
+# cache), and keeps ${VAR} references live against the current os.environ. The
+# (mtime_ns, size) key means any on-disk edit (including by _save_yaml_config_file)
+# is picked up on the next read.
+_yaml_file_cache: dict[str, tuple] = {}
+_yaml_file_cache_lock = threading.Lock()
+
+
 def _load_yaml_config_file(config_path: Path) -> dict:
     try:
         import yaml as _yaml
     except ImportError:
         return {}
 
-    if not config_path.exists():
+    try:
+        st = config_path.stat()
+    except OSError:
+        # Missing or unstattable file — preserve the original "no config" contract.
         return {}
+
+    cache_key = str(config_path)
+    stat_key = (st.st_mtime_ns, st.st_size)
+    with _yaml_file_cache_lock:
+        cached = _yaml_file_cache.get(cache_key)
+        if cached is not None and cached[0] == stat_key:
+            expanded = _expand_env_vars(cached[1])
+            return expanded if isinstance(expanded, dict) else {}
+
+    # Cache miss / stale: parse off disk. Done outside the lock so a slow parse
+    # doesn't serialize unrelated paths; a concurrent duplicate parse is harmless.
     try:
         loaded = _yaml.safe_load(config_path.read_text(encoding="utf-8"))
-        return _expand_env_vars(loaded) if isinstance(loaded, dict) else {}
     except Exception:
         logger.debug("Failed to parse yaml config from %s", config_path)
         return {}
+
+    raw = loaded if isinstance(loaded, dict) else {}
+    with _yaml_file_cache_lock:
+        _yaml_file_cache[cache_key] = (stat_key, raw)
+    if not raw:
+        return {}
+    expanded = _expand_env_vars(raw)
+    return expanded if isinstance(expanded, dict) else {}
 
 
 def get_config_for_profile_home(profile_home: "Path | str | None") -> dict:
--- a/static/ui.js
+++ b/static/ui.js
@@ -3069,13 +3069,33 @@ function _applyReasoningChip(eff){
   _highlightReasoningOption(effort);
 }
 
+// Tracks the model/provider identity of the last reasoning fetch so routine
+// topbar syncs can serve the cached chip state instead of re-hitting the
+// network. null = never fetched.
+let _lastReasoningFetchKey=null;
+
 function fetchReasoningChip(){
+  _lastReasoningFetchKey=_reasoningEffortQuery();
   api('/api/reasoning'+_reasoningEffortQuery()).then(function(st){
     _applyReasoningChip((st&&st.reasoning_effort)||'', st||{});
   }).catch(function(){_applyReasoningChip('', {supported_efforts:[]});});
 }
 
 function syncReasoningChip(){
+  // #4650: syncTopbar() calls this on every routine UI refresh, and during
+  // streaming those fire at high frequency. Before a9ce2889 this served the
+  // cached _currentReasoningEffort after the first load; that commit made it
+  // refetch unconditionally to refresh supported-efforts after a model switch,
+  // which turned ordinary syncs into a GET /api/reasoning storm (one per token).
+  // Restore the cache short-circuit but keep a9ce2889's intent: only hit the
+  // network when nothing is cached yet OR the model/provider identity changed
+  // since the last fetch (the only inputs that change /api/reasoning's answer).
+  // The user-pick and model-switch paths still update the cache directly.
+  const key=_reasoningEffortQuery();
+  if(_currentReasoningEffort!==null && _lastReasoningFetchKey===key){
+    _applyReasoningChip(_currentReasoningEffort);
+    return;
+  }
   fetchReasoningChip();
 }
 
--- a/tests/test_issue4650_reasoning_chip_no_storm.py
+++ b/tests/test_issue4650_reasoning_chip_no_storm.py
@@ -1,0 +1,165 @@
+"""Behavioural test for the #4650 reasoning-chip request-storm fix.
+
+`syncTopbar()` calls `syncReasoningChip()` on every routine UI refresh, and
+during streaming those fire at high frequency. Commit a9ce2889 made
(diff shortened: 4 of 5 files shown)


nesquena (approved):
## Review — end-to-end ✅ (approved after fix; pushed a CHANGELOG entry)

Independent review of the #4650 fix: streaming throughput collapsed to ~2 tok/s because the composer's reasoning-effort chip issued a `GET /api/reasoning` **per token**, and that endpoint re-parsed an ~800-line `config.yaml` (~125ms) on every call. A two-layer fix, exactly per the issue analysis.

### What this ships

`static/ui.js` (+20), `api/config.py` (+37/-2), `tests/test_issue4650_reasoning_chip_no_storm.py` (+165), `tests/test_issue4650_yaml_config_cache.py` (+111). Agent-authored. Closes #4650 (reported by @batkuip). I added the missing CHANGELOG entry (commit below).

Fresh `nousresearch/hermes-agent` tarball: both layers are WebUI-internal. `_load_yaml_config_file` reads `config.yaml`, which the agent/CLI also read/write — but the memo is **WebUI-process-local** and keyed on the file's `(mtime_ns, size)`, so an agent-side write to `config.yaml` (different process) invalidates the WebUI's cache on the next read. Cross-process consistency is preserved via the stat key; nothing about the on-disk format or the agent's own reads changes. No cross-tool regression. ✓

### Frontend — request-storm fix, traced end-to-end

`syncTopbar()` calls `syncReasoningChip()` on every routine refresh (high frequency mid-stream). `a9ce2889` had made `syncReasoningChip()` refetch unconditionally → one `GET /api/reasoning` per token. The fix restores a cache short-circuit ([ui.js:3079](static/ui.js#L3079)):
```js
const key=_reasoningEffortQuery();
if(_currentReasoningEffort!==null && _lastReasoningFetchKey===key){ _applyReasoningChip(_currentReasoningEffort); return; }
fetchReasoningChip();
```
I verified the cache key and every update path:
- **Key = `{model, provider}`** (`_reasoningEffortContext` → `_reasoningEffortQuery`) — precisely the only inputs that change `/api/reasoning`'s answer (supported efforts are per model/provider).
- **`_lastReasoningFetchKey`** is set in `fetchReasoningChip()`; `_currentReasoningEffort` is set only in `_applyReasoningChip()`.
- **Model switch:** the model-pick handler ([boot.js:1404](static/boot.js#L1404)) sets `S.session.model` *then* calls `syncReasoningChip()` → new key → key-mismatch → `fetchReasoningChip()` refetches **exactly once**. ✓
- **User pick:** `POST /api/reasoning` `.then` → `_applyReasoningChip(effort, st)` updates `_currentReasoningEffort` + supported_efforts directly → next sync serves it, no refetch. ✓
- **Routine same-model sync:** key matches, effort non-null → served from cache, **zero network**. ✓
- Forced-refresh callers use `fetchReasoningChip()` directly ([boot.js:2305](static/boot.js#L2305)); routine syncs use the cache-aware `syncReasoningChip()`. Clean split.

So the chip refetches only on cold-start, model/provider change, or explicit pick — never per-token. Restores pre-`a9ce2889` behavior with `a9ce2889`'s model-switch intent intact.

### Backend — YAML memoization: the cache-aliasing concern, independently disproven

`_load_yaml_config_file()` now caches the **raw** parsed dict keyed on `(str(path), (st_mtime_ns, st_size))` and re-runs `_expand_env_vars()` per call. The make-or-break safety property — that the ~7 read-modify-save callers can't corrupt the shared cache — rests on `_expand_env_vars()` being non-mutating and returning a fresh structure. I confirmed it ([config.py](api/config.py)):
```python
if isinstance(obj, dict):  return {k:_expand_env_vars(v) for k,v in obj.items()}
if isinstance(obj, list):  return [_expand_env_vars(i) for i in obj]
```
It rebuilds **every** dict and list top-to-bottom and only aliases immutable scalars/strings — so it never touches its input and shares no mutable object with the output.

**Independent harness (8/8)** against the real `_load_yaml_config_file`:
```
env expanded on read; deep nested dict mutation + nested list mutation + list append by a caller
  → fresh read stays PRISTINE (gpt-5, [a,b], 2 providers); c1/c2 distinct objects
env changed, no file change → re-expanded per call (${VAR} live, no reparse)
on-disk edit → mtime/size key invalidates → new content seen
missing file → {}
```
This conclusively closes the corruption risk. **Thread-safety:** `_yaml_file_cache` is guarded by `_yaml_file_cache_lock`; the raw is only ever read (never expanded in place), the slow parse runs *outside* the lock (a concurrent duplicate parse is harmless, last-writer-wins), and the stat key means a `_save_yaml_config_file` write is picked up on the next read. Storing the raw (with `${VAR}` intact) and expanding per call also preserves the per-profile `os.environ` semantics this codebase relies on. ✓

### What I pushed — CHANGELOG entry (commit a51f2186)

The PR shipped a user-visible behavior change (streaming throughput restored) with no CHANGELOG note; AGENTS.md requires one. Added an `[Unreleased]` → Fixed entry. No version section (release agent's domain).

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Streaming, same model, N syncs | 1 cold fetch then 0 (storm gone) ✅ |
| Model/provider switch | new key → exactly 1 refetch ✅ |
| User picks an effort | POST updates cache directly; no refetch ✅ |
| Transient `/api/reasoning` error | chip hidden, cached as `''`; self-heals on model switch/pick (see below) ✅ |
| `config.yaml` reread within process | served from memo, no reparse ✅ |
| `config.yaml` edited on disk (incl. agent process) | mtime/size key invalidates → fresh parse ✅ |
| `${VAR}` changes between reads | re-expanded live per call ✅ |
| Caller read-modify-saves the config dict | fresh expanded copy; cache uncorrupted ✅ |
| Concurrent readers | lock-protected; raw read-only ✅ |

### Tests

- The two new files — **9/9** (node-driven call-counter: 1 cold → 0×10 same-model → 1 switch → 0 after; cache hit skips reparse, mtime invalidates, env per-call, caller-mutation-safe, missing→{}).
- **Independent YAML-cache harness: 8/8** (above) — deeper nested-mutation + env-liveness + mtime proof.
- **Full suite: 9895 passed / 132 skipped / 0 failures** in 203s (deselected the pre-existing darwin CRLF flake).
- **JS gate:** `node --check static/ui.js` clean; ESLint runtime guard exit 0.

- **Error-state stickiness (Greptile):** on a `/api/reasoning` error, `fetchReasoningChip` caches `_currentReasoningEffort=''`, so routine syncs keep the chip hidden until a model switch/pick rather than retrying. I considered pushing a "retry on error" change but it would **reintroduce the very storm this PR fixes** — a persistently-erroring endpoint would be re-hit once per token (now with failing requests) during streaming. Caching the error and stopping the hammering is the correct trade-off; it self-heals on model switch or an explicit pick, and matches pre-`a9ce2889` behavior. Intentional, leave as-is.
- **Lock held across `_expand_env_vars` on cache hits (Greptile):** serializes concurrent readers across env-expansion. Env expansion is O(config size) pure-Python and far cheaper than the YAML parse it replaces, so the contention is negligible; copying the raw ref out and expanding outside the lock would shave it if it ever matters. Non-blocking.

### Recommendation

Both layers are correct and well-scoped. The frontend cache key is exactly the model/provider identity that drives the endpoint's answer, and I verified all three update paths (cold / model-switch / pick) keep it consistent so the chip refetches once-per-switch and never per-token. The backend memo is safe against the read-modify-save callers because `_expand_env_vars` provably rebuilds every container and never mutates its input — proven with an 8/8 nested-mutation + env-liveness + mtime harness — and it's lock-guarded and cross-process-consistent via the stat key. I pushed the missing CHANGELOG entry. Ship.

nesquena-hermes:
Closing — shipped.

The fix in this PR (cache the reasoning-effort chip state on routine topbar syncs / only re-fetch `/api/reasoning` on model-provider identity change, plus the defensive `config.yaml` mtime+size-keyed parse memoization) was cherry-picked into the release batch and shipped in **v0.51.573** (Release UF, "reasoning request-storm fix"), commit `1f4e70a5d`. Verified live on `origin/master`: `api/config.py:269` carries the `_cfg_mtime` memoize and `cache_stale` gate.

The companion issue #4650 is already closed. Closing this PR object as the fix is live; nothing further to merge.

PR #4598, opened 2026-06-21
Settings: split composer footer controls into primary vs situational groups

Paladin173:
## Thinking Path
- Hermes WebUI should keep settings clear and predictable, especially for UI-heavy workflows.
- The footer controls list mixed always-relevant controls with controls that appear only in certain runtime or viewport conditions.
- That made the Appearance panel noisy and made some toggles feel broken when controls were conditionally hidden.
- The fix was to separate everyday footer controls from situational controls, while preserving per-profile persistence and live-apply behavior.
- This improves discoverability and reduces confusion without removing advanced control.

## What Changed
- Added a new Appearance section: Situational footer controls.
- Kept primary controls in Chat footer controls and moved situational controls into the new group.
- Situational group now includes:
  - Voice mode
  - YOLO
  - Background badge
  - Mobile config
  - Quota chip
  - Toolsets
  - Status
- Added an explanatory note that these controls do not always appear, but can still be toggled off.
- Updated footer visibility wiring so both groups share the same persistence and live-apply flow.
- Ensured backend settings keys and CSS hide helper are included and synced.
- Updated changelog entries for release-note visibility.

## Why It Matters
- Reduces settings clutter and improves the user mental model.
- Prevents toggle seems broken confusion for conditionally rendered controls.
- Keeps power-user control while making defaults easier to understand.
- Maintains per-profile behavior and live UI updates.

## Verification
- Manual verification in Settings -> Appearance:
  - New situational section appears.
  - Chips render in correct groups.
  - Toggling chips updates the UI live.
  - Settings persist after reload and profile change.
- Confirmed changes were committed and pushed in two commits:
  - 748d31d5
  - 7ddb5c36
- Note: full automated test suite was not run for this UI-focused pass.

## Risks / Follow-ups
- UI-only risk is low; behavior is mostly organizational and reuses existing toggle plumbing.
- Follow-up (optional): add or update TESTING guidance for footer-control manual checks.
- Follow-up (optional): add before and after screenshots in PR for review completeness.

## Model Used
- Provider: GitHub Copilot
- Model: GPT-5.3-Codex
- Tooling: workspace file edits, git commit and push, targeted code search and validation

## Before
image

## After
image

PR #4657, opened 2026-06-22
feat(composer): fit footer controls by overflow

Paladin173:
## Thinking Path

- Maintainer feedback on #4657 said the `hide_composer_*` visibility-toggle half already shipped via #4598.
- The still-novel part is the composer footer fit engine plus compact model/reasoning/quota chips.
- This update rebases onto current `master`, strips the duplicate shipped visibility-toggle work, and keeps the PR focused on that remaining product decision.

## What Changed

- Rebased the PR branch onto current `master`.
- Removed the duplicated `hide_composer_*` visibility-toggle implementation from this PR scope.
- Kept the JS fit engine that measures real composer-footer overflow and steps through full labels, icon chips, then hamburger consolidation.
- Kept compact model, reasoning, and quota chip labels.
- Restored the compact/mobile composer menu so the hamburger appears in phone mode and the context ring opens the same menu.
- Ensured the quota chip is shown by default when quota data is available, can still be hidden by setting, and folds into the shared compact menu when space runs out.

## What Stayed Out

- Composer-control chip reordering/drag sorting is intentionally split into a separate PR.
- This PR does not re-land the already-shipped footer visibility toggles from #4598.

## Why It Matters

The diff is now reviewable as the exact decision the maintainer called out: whether the composer footer should move from pure CSS container-query responsiveness to the JS observer-backed fit engine for better content-aware compaction.

## Before

footer-before

## After

footer-after

## Verification

- `node --check static/ui.js`
- `git diff --check`
- `HERMES_HOME=C:\tmp\hermes-webui-empty-home HERMES_BASE_HOME=C:\tmp\hermes-webui-empty-home .\.venv\Scripts\python.exe -m pytest tests/test_issue4598_composer_control_visibility.py tests/test_mobile_layout.py tests/test_quota_chip_settings_toggle.py -q -o cache_dir=C:\tmp\hermes-webui-pytest-cache-issue4657-focused` — 82 passed

## Risks / Follow-ups

- Main risk is the one named in review: JS observers/rAF fitting on the composer footer versus the current zero-JS container-query approach.
- Follow-up PR covers optional drag sorting for the Settings composer-control chips.

## Contract Routing

Not contract-affecting. This is a composer footer UI behavior change and does not alter runtime, streaming, recovery, replay, compression, session metadata, or persistence contracts beyond existing settings storage.

- Provider: OpenAI
- Model: GPT-5 Codex
- Tool use: local git branch splitting, code edits, focused pytest shard, syntax checks, and GitHub CLI PR update.


Paladin173:
Addressed in 64c0ddde.

The `ResizeObserver` and `MutationObserver` are now stored in module-level variables along with the observed footer element, and the one-shot `_composerFitObserversBound` flag is gone. `_initComposerFooterFit()` re-binds only when the footer element actually changes — disconnecting the stale observers first — so a future unmount/re-mount re-observes the new footer instead of silently stopping. The `window` resize listener still binds once since it targets the stable `_scheduleComposerFit`.

Behavior is unchanged for the normal single-mount case; `node --check` passes.

nesquena-hermes (changes requested):
Deep-reviewed during a PR sweep — flagging a supersession before this goes further, @Paladin173.

**Half of this PR has already shipped.** The composer-footer per-control visibility toggles (`hide_composer_*`) landed in **v0.51.563 via your own #4598**. I compared the config keys directly: master and this PR define the *identical* 15 keys (`hide_composer_attach`, `_saved_prompts`, `_mic`, `_voice_mode`, `_yolo`, `_profile`, `_workspace`, `_mobile_config`, `_model`, `_quota_chip`, `_reasoning`, `_toolsets`, `_status`, `_context`, `_bg_badge`) — `comm -13` shows **zero** keys unique to this branch. That's the main reason the rebase conflicts so heavily (767 commits behind, and the footer-controls surface has since shipped + churned).

**What's genuinely novel here:** the JS **fit engine** — `_fitComposerFooter()` stepping full → `.cf-icons` → `.cf-burger` driven by a `ResizeObserver` + `MutationObserver` (rAF-debounced), plus the compact model/reasoning/quota chips. Master has **zero** references to `_fitComposerFooter`/`cf-icons`/`cf-burger`; its composer footer is currently responsive via **CSS container queries** (`@container composer-footer (max-width: 700px/520px)`, `min-width: 1100px`).

So the remaining decision isn't a merge-readiness one, it's a **shape call**: replace the working CSS container-query responsiveness with a JS observer-driven reflow loop on the composer (the crown-jewel surface). That's a maintainer/perf judgement (observer reflow cost on every composer resize/mutation vs. zero-JS container queries), not an autonomous merge.

**Recommended path forward:** if you'd still like the fit engine, please **rebase onto current master and strip the now-shipped `hide_composer_*` half** so the PR is *only* the fit-engine + compact-chips diff against today's container-query baseline — that makes it reviewable as the focused change it now is, and lets us weigh the JS-observers-vs-container-queries tradeoff on its own merits. As-is, the diff is mostly a re-application of code that already shipped.

Marking changes-requested pending that rebase/split. Thanks for the footer work — #4598 from this same effort is live and appreciated.

Paladin173:
@nesquena-hermes  I think i took care of everything, let me know if anything else is needed. I came up with another enhancement and a separate PR is opened for that.

Paladin173:
Addressed and force-pushed. The PR is now rebuilt from current \\master\\ around the fit-engine only.

What changed:
- dropped the quota-chip default flip / migration so \\show_quota_chip\\ stays default-off
- kept the fit-engine-related quota overflow/menu wiring only
- updated the fit observer to watch both \\.composer-footer\\ and \\.composer-left\\ so right-side control changes can re-trigger the fit correctly

Verification:
- \\.\\venv\\Scripts\\python.exe -m pytest --noconftest --basetemp C:\\tmp\\pytest-hermes tests/test_mobile_layout.py tests/test_reasoning_chip_js_behaviour.py tests/test_issue1431_toolsets_chip_responsive.py tests/test_issue4650_reasoning_chip_no_storm.py tests/test_quota_chip_settings_toggle.py\\
- 107 passed

Issue #4570, opened 2026-06-21
Mobile: animate slide-out drawer with transform instead of left (smoother open/close)


nesquena-hermes:
### Summary

On mobile the left navigation drawer slides in/out by animating the CSS `left` property, which can stutter/jank on lower-end devices and inside WebViews. Switching the slide to a GPU-composited `transform: translateX()` would make it noticeably smoother.

### Where this lives

This is **WebUI**, not a native-shell concern — it affects the mobile drawer in every mobile context that renders the web app: PWA (Add to Home Screen), mobile browser, and the native Android/iOS WebView shells (which just host this same web layer). Fixing it here fixes it everywhere at once.

### Current behavior

`static/style.css` (mobile `@media(max-width:640px)` block):

```css
.sidebar{position:fixed;left:-320px;top:0;bottom:0;width:min(320px, 100vw);z-index:200;
  transition:left .25s ease;}
.sidebar.mobile-open{left:0;}
```

Animating `left` triggers layout on every frame (not just compositing), so the slide can drop frames — especially in a WebView or on a busy chat view.

### Proposed fix

Keep the drawer pinned at `left:0` and translate it off-screen, animating `transform` (composited, no layout):

```css
.sidebar{position:fixed;left:0;top:0;bottom:0;width:min(320px,100vw);z-index:200;
  transform:translateX(-100%);
  transition:transform .25s cubic-bezier(.22,1,.36,1);
  will-change:transform;}
.sidebar.mobile-open{transform:translateX(0);}
```

`will-change:transform` should be applied only while the drawer is interactive (or dropped after the transition) to avoid keeping a permanent compositor layer. The overlay fade can stay as-is.

### Acceptance

- Drawer open/close animates via `transform`, not `left`.
- No visual regression vs. the current slide distance/curve on desktop or mobile.
- Verified smooth on a mid-range Android device / WebView.

nesquena-hermes:
## Summary

Citation confirmed and the `transform` swap is sound. Reading `static/style.css:2350-2352` on `origin/master`, the mobile drawer does animate the `left` property, which forces layout each frame:

```css
.sidebar{position:fixed;left:-320px;top:0;bottom:0;width:min(320px, 100vw);z-index:200;
  transition:left .25s ease;}
.sidebar.mobile-open{left:0;}
```

Moving to a composited `transform:translateX()` is the right call for WebView/low-end smoothness.

## The one collision to check — and why it's actually clear

[...]

PR #4660, opened 2026-06-22
Enhance PWA mobile sidebar navigation

franksong2702:
## Thinking Path

PWA users interact with the mobile sidebar differently from desktop users. This enhancement keeps the existing Hermes rail-and-panel model, but makes the phone drawer behave like a dedicated PWA navigation surface: full-screen, gesture-friendly, and able to reveal the selected detail immediately after a list item is chosen.

## What Changed

- Expanded the phone sidebar into a full-screen transform drawer while preserving the left icon rail for module switching.
- Made hamburger and left-edge swipe open the list for the currently active sidebar module instead of defaulting back to Sessions.
- Kept rail clicks inside the full-screen mobile drawer as list-switching actions rather than drawer-dismiss actions.
- Closed the mobile drawer after detail-driving sidebar selections so the selected detail page is visible immediately.
- Kept inline sidebar controls usable without closing the drawer.
- Disabled the dimming overlay for this full-screen PWA drawer so the top safe-area/titlebar does not look disabled.
- Added a narrow left-edge gesture guard to make PWA sidebar opening more reliable on iOS without blocking the hamburger button.
- Handled plugin and Todos panel-sync edge cases explicitly so main-view-only and sidebar-only panels do not produce blank sidebar states.

## Why It Matters

This is a PWA-specific navigation enhancement. Hermes WebUI still works in mobile Safari, but installing it as a PWA now gives a more app-like sidebar experience: the navigation surface uses the full phone viewport, keeps module switching close at hand, and reveals selected detail pages without extra dismissal steps.

## Verification

- `git diff --check`
- `./scripts/test.sh tests/test_mobile_layout.py tests/test_pwa_sidebar_swipe.py -q`
  - `71 passed`
- Manual validation in iOS PWA mode.
- Manual validation in iOS Safari browser mode.
- Claude CLI reviewed the local diff and found no blocking issues; the panel coverage concern it raised was addressed before this PR.

## Risks / Follow-ups

- This intentionally favors the installed PWA experience on phones; browser mode remains supported but is not the primary optimized path.
- The left-edge gesture guard is narrow so it does not block normal page interactions.
- Desktop sidebar collapse behavior is intended to remain unchanged.

## Model Used

GPT-5 Codex implementation with Claude CLI review.

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -294,6 +294,10 @@ function _pwaSidebarSwipePoint(e){
   return {clientX:Number(src.clientX)||0,clientY:Number(src.clientY)||0};
 }
 
+function _isTouchPointerEvent(e){
+  return !!(e&&e.pointerType==='touch');
+}
+
 function _openMobileSidebarFromGesture(){
   if(_isDesktopWidth())return;
   const sidebar=document.querySelector('.sidebar');
@@ -310,13 +314,14 @@ function _openMobileSidebarFromGesture(){
 
 function _onPwaSidebarSwipeStart(e){
   if(_isDesktopWidth())return;
+  if(_isTouchPointerEvent(e))return;
   if(e.pointerType==='mouse'||(e.pointerType&&e.pointerType!=='touch'&&e.pointerType!=='pen'))return;
   if(document.querySelector('.sidebar')?.classList.contains('mobile-open'))return;
   const point=_pwaSidebarSwipePoint(e);
   if(!point)return;
   if(point.clientX>_PWA_SIDEBAR_SWIPE_EDGE)return;
   if(_isInteractiveSwipeTarget(e.target))return;
-  _pwaSidebarSwipe={startX:point.clientX,startY:point.clientY,active:true,opened:false,claimed:false};
+  _pwaSidebarSwipe={startX:point.clientX,startY:point.clientY,active:true,opened:false};
 }
 
 function _onPwaSidebarEdgeGuardStart(e){
@@ -324,10 +329,10 @@ function _onPwaSidebarEdgeGuardStart(e){
   if(document.querySelector('.sidebar')?.classList.contains('mobile-open'))return;
   if(e.cancelable)e.preventDefault();
   _onPwaSidebarSwipeStart(e);
-  if(_pwaSidebarSwipe)_pwaSidebarSwipe.claimed=true;
 }
 
 function _onPwaSidebarSwipeMove(e){
+  if(_isTouchPointerEvent(e))return;
   const swipe=_pwaSidebarSwipe;
   if(!swipe||!swipe.active||swipe.opened)return;
   const point=_pwaSidebarSwipePoint(e);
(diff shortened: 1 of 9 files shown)


nesquena-hermes:
Pulled the branch and read the full diff at HEAD (`a035b1343`) against `origin/master` — `boot.js`, `panels.js`, `index.html`, `style.css`. The architecture is clean and the refactor is an improvement on its own: hoisting the inline `_mainViewPanels` array into the module-level `MAIN_VIEW_PANELS` constant (panels.js:44) removes the duplicated literal that previously lived in both `switchPanel` and `switchPluginPage`, and the new `_syncMobileSidebarPanelFromMainView` correctly infers the visible panel from `main.main`'s `showing-*` class before opening the drawer.

I also independently confirmed the gesture dedup that greptile flagged as sound. The dual touch+pointer registration in `_installPwaSidebarSwipeGesture` (boot.js) is safe because `_isTouchPointerEvent` early-returns the pointer path for `pointerType==='touch'`:

```js
function _isTouchPointerEvent(e){ return !!(e&&e.pointerType==='touch'); }
// in _onPwaSidebarSwipeStart / Move / End / Cancel:
if(_isTouchPointerEvent(e))return;
```

so touch is handled by the `touch*` listeners and mouse/pen by the pointer listeners — no double-fire. The `preventDefault` in `_onPwaSidebarSwipeMove` only fires once `dx>=_PWA_SIDEBAR_SWIPE_CLAIM` and the move is mostly-horizontal (`dx>Math.abs(dy)*1.2`), and the swipe only arms when the touch starts in the left 80px, so vertical scrolling elsewhere is unaffected. Good.

## One concrete nit: the close button label

index.html:180 adds the close affordance with a hardcoded, English-only, and now semantically-wrong label:

```html
<button class="panel-head-btn mobile-sidebar-close ..." onclick="closeMobileSidebar()"
  data-tooltip="Close conversations" aria-label="Close conversations">
```

Two problems:

1. **No i18n binding.** Every sibling control uses `data-i18n-title` / `data-i18n-aria-label` (e.g. index.html:150, :165), so non-English locales will see English "Close conversations" here.
2. **The text is wrong for non-chat panels.** This PR's whole point is that the full-screen drawer now opens to *any* active module — Settings, Skills, Kanban — via `_syncMobileSidebarPanelFromMainView`. When the drawer is showing Settings, an "Close conversations" label is misleading. And since this PR disables the backdrop (`.mobile-overlay.visible{display:none}`, style.css:2390) and goes full-screen 100vw, this button is now the *primary* dismiss affordance — there's no tap-outside fallback — so getting its label right matters more than before.

Suggest a generic key, e.g. `aria-label`/`title` bound to something like `close_menu` ("Close menu" / "Close sidebar"), with `data-i18n-title` + `data-i18n-aria-label` added so it translates.

Everything else looks merge-ready. Nice work on the panel-sync-on-open lifecycle — it closes the gap from #4570/#4572 cleanly.

Issue #4622, opened 2026-06-21
bug(transparent-stream/tool-cards): terminal tool output truncated + not expandable on click; patch/edit cards missing diff (suspected regression)


nesquena-hermes:
bug(transparent-stream/tool-cards): terminal tool output truncated + not expandable on click; patch/edit cards missing diff (suspected regression)
## Symptom
In the Transparent Stream / Worklog view, terminal (shell) tool-call cards no longer show their full output, and clicking the output region no longer expands it to reveal the rest. Separately, patch/edit_file tool-call cards appear to no longer render their diff body. Reported as a regression following the recent v0.51.547–v0.51.560 upgrade wave.

## Suspected area
`static/ui.js` Transparent Stream / Worklog tool-card render path (settled branch) plus the tool-snippet expand/show-more click handler. Recent churn on this surface: tool-call row preservation through settled rebuild and persisted anchor-activity rendering. The diff-render half (patch/edit cards) may share the same render regression or be distinct.

## Repro
1. Run a turn that invokes a terminal/shell tool with output longer than the compact cap.
2. Observe the tool-call card shows truncated output.
3. Click the output region — it does not expand to show full output.
4. Run a turn that invokes a patch/edit_file tool — the diff body does not render in the card.

## Notes
Reporter version/build unknown — first triage step is to pin the build and identify which release in the v0.51.547–.560 window introduced it. Patch-diff half flagged with reporter uncertainty.

---
_Filed from a Discord community report. Edit or close if mis-scoped._

nesquena-hermes:
**Triage: could not reproduce on current `master` (v0.51.560) — looks already-fixed / stale.**

I built an independent node reproduction harness that runs the *actual* `buildToolCard`, `_anchorSceneToolCallFromRow`, and diff helpers from current `static/ui.js` and exercised both reported symptoms, including the reload/settled (Transparent Stream rehydrate) path:

**1. "terminal output truncated + not expandable on click"**
- `buildToolCard` renders a **"Show more"** button whenever the output exceeds the 800-char display cap, with the **full** output carried in `data-full=` (the inline `_toggleToolDiff` expand handler reads it). Verified: a 2000-char shell output → expand button present, `data-full` length = 2000 (full, not truncated).
- On the **reload/settled path**, `_anchorSceneToolCallFromRow` recovers the snippet from the persisted scene row (`payload.output`/`result`) and the restored card is **still expandable** with the full output.
[...]

PR #4661, opened 2026-06-22
fix(#4658): restore collapsed-row inline summary in transparent-stream tool cards

### LINKED ISSUE #4658: bug(transparent-stream/tool-cards): collapsed tool-card rows lost their inline summary/preview — name renders + expand works, but no one-liner without expanding (regression)
bug(transparent-stream/tool-cards): collapsed tool-card rows lost their inline summary/preview — name renders, expand works, but you can't tell what a call did without expanding (regression)

## Symptom (corrected after reporter follow-up)

Earlier this was filed as "cards render BLANK — no name/args/output." The reporter follow-up clarifies it's **narrower and more specific than that**:

- The tool **name still renders** (`read_file`, `search_files`, `terminal`, `skill_view`, …).
- **Expanding a row still works** — clicking it reveals the full detail (the `Full` / `Output` tabs, the `Shell` lead, patch/edit diffs).
- **What's missing is the collapsed-row inline summary/preview** — the one-line snippet that used to sit next to the tool name so you could understand what each call did *without* expanding it.

In the reporter's words: *"with the old view, you don't have to expand stuff to understand what's happening"* and *"it's the summary missing."* So this is a **collapsed-row preview regression**, not a fully-blank-card bug.

It also affects **old / persisted sessions** (reporter: *"old sessions are all missing the details too"*), so it's not limited to live streaming — the persisted-reload render path shows the same loss.

## Before / after

**Before (working — every collapsed row carried an inline preview):**

before — collapsed rows with inline previews

**After (regressed — collapsed rows show only the bare tool name, no preview):**

after — bare tool-name rows, no preview

**Expanding still works (the detail is intact — only the collapsed summary is gone):**

after — expanding a row shows Full/Output + Shell

## Suspected area — two candidate mechanisms to check

`buildToolCard(tc)` (`static/ui.js` ~L12504) always emits a `<span class="tool-card-preview">` (L12557), and the row name renders fine — consistent with "name shows, preview gone." Two places can independently empty that preview:

**1. Render-side suppression (`static/ui.js` ~L12542–12544):**
```js
let previewText=_toolCardPreviewText(tc, displaySnippet);
const argPreview=_formatToolArgPreview(tc&&tc.args);
if(toolKind==='shell'||previewText===argPreview||previewText==='Completed'||previewText==='Running'||previewText==='Failed') previewText='';
```
`_toolCardPreviewText()` (L12479) returns `argPreview` whenever args produce a preview (its common path), so the `previewText===argPreview` guard **blanks the preview for exactly the common case**, and the `toolKind==='shell'` clause blanks it for every terminal call. If this guard was recently tightened, the collapsed one-liner would disappear for most tools while the detail block (built separately) keeps working — matching the screenshots exactly.

**2. CSS visibility by wrapper class (`static/style.css`):**
`.tool-card-preview` is `display:none` by default in **two** places (L3272 and L5264). It's only un-hidden by more-specific container selectors:
- `.tool-worklog-list .tool-card-preview { display:block; }` (L3066–3069)
- `.transparent-event-row .tool-card-preview { display:inline; }` (L3410–3413)

So a card rendered **without** the `.tool-worklog-list` / `.transparent-event-row` wrapper class shows its name but keeps the preview hidden — and expanding still works because the detail block isn't gated by those selectors. If a recent render-path change (persisted-anchor reconstruction) stopped applying the wrapper class, that alone reproduces the symptom on **old/persisted sessions**, which fits the "old sessions missing too" report.

This surface was recently churned by **#4587 / #4568** (render persisted anchor activity on reload) and the **#4411** Stable Assistant Turn Anchors capstone — likely regression window.

> @b3nw indicated he has a candidate root cause and will add it to this issue — leaving room for that; the two mechanisms above are where to confirm it.

## Repro
1. Open a session that invoked tools, in Transparent Stream / Worklog view.
2. Observe collapsed tool rows show **only the tool name** — no inline preview/summary.
3. Click a row → full detail (Full/Output tabs, Shell, diffs) **does** appear.
4. Confirm it reproduces on an **old/persisted** session after reload, not just live streaming.

## First triage step — rule out stale cache
JS is cache-busted (`ui.js?v=__WEBUI_VERSION__`) and the SW re-registers per version, but during the current multiple-releases-per-day cadence a PWA can hold a stale `index.html` shell + SW pointing at an old `?v=`. Hard-refresh / confirm the build badge matches the deployed tag first. If the preview returns after a hard refresh → stale-SW artifact; if it persists on a confirmed-fresh build → real regression (capture which path: live vs persisted-reload).

## Relationship to #4622
Distinct symptom, possibly shared regression window. **#4622** (latipun) = expanded terminal output truncated + not expandable + patch/edit missing diff. **This issue (#4658)** = the *collapsed-row inline preview/summary* is gone while expansion works. Cross-linking; if a single render-path change is found to cause both, consolidate.

---
_Filed from a Discord community report (@b3nw). Symptom framing corrected and before/after evidence added after reporter follow-up._

nesquena-hermes:
## What

Restores the collapsed-row inline summary/preview in the **Transparent Stream** tool-card view (#4658). Collapsed tool rows regressed to showing only the bare tool name (`read_file`, `terminal`, `search_files`, `skill_view`) with no hint of what each call did — you had to expand every row. Affects both live streaming and persisted/old sessions on reload (they share this render path), matching the report.

## Root cause — deep review

Two changes interacted (the bot's "rebuild-on-every-token" theory is ruled out: persisted/old sessions are affected too, and they never hit the live-rebuild path):

1. **`69072ac34`** ("fix: keep collapsed tool previews quiet") moved `buildToolCard`'s collapsed preview from the output snippet (`tc.preview||displaySnippet`) to an **argument** summary (`_toolCardPreviewText` → `_formatToolArgPreview`).
2. **`6951d10a`** (#4411 *Stable Assistant Turn Anchors*) then **blanked** that preview for the common arg/shell case at `ui.js:12544`:
   ```js
   if(toolKind==='shell'||previewText===argPreview||previewText==='Completed'||…) previewText='';
   ```
   `_toolCardPreviewText` *returns* `argPreview` in its common path, so `previewText===argPreview` is true for most tools. #4411 could do this safely because it **also** changed the row **name** to a friendly action label (`_toolActionLabelText` → "Read config.py") that carries the target.

The bug: the transparent-stream branch of `_decorateTransparentEventRow` (`ui.js:8431`) **overrides the name back to the bare tool name** (`_toolShortName(name)` → `read_file`). So the transparent view ended up with **neither** #4411's target-carrying label **nor** a preview → bare rows. The worklog view is unaffected (it keeps the action-label name).

## Fix

`_decorateTransparentEventRow` now populates the row's existing `.tool-card-preview` span from a new quiet helper `_transparentToolSummary(tc)`, built from the call's **target** — file path / command / query / skill name — and explicitly **not** the raw result JSON (that stays in the expandable detail body). This honors `69072ac34`'s "keep collapsed previews quiet" intent while restoring "understand the call without expanding it." Idempotent across status re-decoration. The CSS already un-hides `.tool-card-preview` inside `.transparent-event-row`, so no style change is needed.

## Verification

**Live browser, shipped `ui.js` + CSS** (isolated test instance) — each collapsed row renders `name  summary`:

| Row | Inline summary |
|---|---|
| `read_file` | `config.py` |
| `terminal` | `git fetch origin --quiet` |
| `search_files` | `buildToolCard` |
| `skill_view` | `opencode-review-agents skill` |
| `terminal` | `python3 -m pytest tests/ -q` |

Matches the issue's "before/working" screenshot.

**Tests** — `tests/test_issue4658_transparent_collapsed_preview.py` (node-driven, drives the real `buildToolCard` + `_decorateTransparentEventRow`): read_file shows the file, terminal shows the command, search shows the pattern **and not** result JSON, skill shows the skill name, and a no-target call yields an **empty** preview (no invented noise). 5 tests.

Existing preview/transparent/anchor suites: **525 passed, 0 regressions** (incl. `test_tool_card_preview_summary.py`, `test_issue4539_transparent_tool_settle.py`, `test_issue4622_tool_card_restore_guard.py`).

## Relationship to #4622

Distinct symptom (this = collapsed-row preview; #4622 = expanded output truncation + missing diff), but the same #4411 regression window. This PR is scoped to the collapsed-row preview only.

Closes #4658.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Collapsed tool-call rows in the Transparent Stream view show their inline summary again.** A regression left collapsed tool rows in transparent-stream mode displaying only the bare tool name (`read_file`, `terminal`, `search_files`, …) with no hint of what each call did — you had to expand every row to see the target. It traced to two changes interacting: the collapsed-row preview was moved to an argument summary and then suppressed for the common case (because the *worklog* view's row name carries the target as a friendly action label), but the transparent view overrides the row name back to the bare tool name and so had neither the target-carrying label nor the preview. Transparent tool rows now rebuild a quiet, one-line summary from the call's target (file path, command, query, skill name, …) — never the raw result JSON, which stays in the expandable detail body. Affects both live streaming and persisted/old sessions on reload. (#4658)
+
 ## [v0.51.573] — 2026-06-22 — Release UF (reasoning request-storm fix)
 
 ### Fixed
--- a/static/ui.js
+++ b/static/ui.js
@@ -8216,6 +8216,28 @@ function _transparentToolStatus(tc, settled){
   if(tc&&tc.done===false) return settled?'Interrupted':'Running';
   return 'Completed';
 }
+// Quiet one-line summary for a collapsed transparent tool row (#4658).
+// The transparent view overrides the row name to the bare tool name
+// (_toolShortName, e.g. "read_file"/"terminal"), so — unlike the worklog view —
+// it has no action-label carrying the target, and buildToolCard's collapsed
+// preview is blanked for the common arg/shell case (the #4411 suppression that
+// assumes the name carries the target). That left transparent rows showing only
+// the bare tool name with no hint of what each call did. Rebuild a summary from
+// the call's TARGET (path/command/query/skill/...) — NOT the raw result JSON —
+// so it stays consistent with the "keep collapsed previews quiet" intent
+// (test_tool_card_preview_summary.py) while restoring "understand the call
+// without expanding it".
+function _transparentToolSummary(tc){
+  if(!tc||typeof tc!=='object') return '';
+  // Explicit progress text (e.g. subagent_progress) wins while still running.
+  const explicit=String(tc.preview||'').trim();
+  if(tc.done===false&&explicit) return _shortToolLabel(explicit,160);
+  const target=typeof _toolVisibleTargetLabel==='function'?_toolVisibleTargetLabel(tc,{limit:160}):'';
+  if(target) return target;
+  const argPreview=typeof _formatToolArgPreview==='function'?_formatToolArgPreview(tc.args):'';
+  if(argPreview) return _shortToolLabel(argPreview,160);
+  return '';
+}
 function _copyEventToClipboard(row){
   if(!row) return;
   const type=row.getAttribute('data-event-type');
@@ -8477,6 +8499,22 @@ function _decorateTransparentEventRow(row, opts){
       // also prefix the name with "Running:" — that's the same redundancy class
       // V6 removed from the detail body. (Trifecta r2 #4.)
       if(nameEl) nameEl.textContent=_toolShortName(name);
+      // #4658: restore the collapsed-row inline summary. buildToolCard emits a
+      // `.tool-card-preview` span but blanks its text for the common case, and
+      // the bare _toolShortName above drops the target the worklog view carries
+      // in its action-label name. Populate the preview from a quiet,
+      // target-based summary so each collapsed row says what it did without
+      // expanding. Idempotent across re-decoration (status updates re-run this).
+      const previewEl=header.querySelector('.tool-card-preview');
+      if(previewEl){
+        const summary=_transparentToolSummary(tc);
+        if(summary){
+          previewEl.textContent=summary;
+          previewEl.removeAttribute('hidden');
+        }else{
+          previewEl.textContent='';
+        }
+      }
       let statusEl=header.querySelector('.transparent-event-status');
       if(!statusEl){
         statusEl=document.createElement('span');
--- a/tests/test_issue4658_transparent_collapsed_preview.py
+++ b/tests/test_issue4658_transparent_collapsed_preview.py
@@ -1,0 +1,205 @@
+"""Regression test for issue #4658: collapsed tool-card rows in the
+transparent-stream view lost their inline summary/preview.
+
+Symptom: collapsed transparent tool rows showed only the bare tool name
+(`read_file`, `terminal`, `search_files`) with no hint of what the call did;
+expanding still worked. Caused by the interaction of two changes:
+  * 69072ac34 moved buildToolCard's collapsed preview to an arg summary and
+    #4411 (6951d10a) then blanked that preview for the common arg/shell case
+    (it assumes the row NAME carries the target — true for the worklog view's
+    action-label name), and
+  * the transparent view (_decorateTransparentEventRow) overrides the row name
+    back to the BARE tool name (_toolShortName), so it had neither the
+    target-carrying label nor a preview.
+
+Fix: _decorateTransparentEventRow now populates the `.tool-card-preview` span
+from a quiet, TARGET-based summary (_transparentToolSummary) — path/command/
+query/skill, never the raw result JSON — so collapsed rows are self-describing
+again while honoring the "keep collapsed previews quiet" intent
+(test_tool_card_preview_summary.py).
+
+This drives the ACTUAL functions from static/ui.js via node + jsdom-free DOM
+shims, and runs the same render path live streaming and persisted reload share
+(both go through _decorateTransparentEventRow(buildToolCard(tc))).
+"""
+import json
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+
+REPO_ROOT = Path(__file__).parent.parent.resolve()
+UI_JS_PATH = REPO_ROOT / "static" / "ui.js"
+
+NODE = shutil.which("node")
+pytestmark = pytest.mark.skipif(NODE is None, reason="node not on PATH")
+
+
+_DRIVER_SRC = r"""
+const fs = require('fs');
+const src = fs.readFileSync(process.argv[2], 'utf8');
+
+// ── Minimal DOM shims (enough for buildToolCard + _decorateTransparentEventRow)
+function makeEl(tag){
+  const el = {
+    tagName: (tag||'div').toUpperCase(),
+    _attrs: {}, _classes: new Set(), children: [], style: {},
+    dataset: {}, _html: '', textContent: '', parentNode: null,
+    firstChild: null,
+    classList: {
+      add(c){el._classes.add(c)}, remove(c){el._classes.delete(c)},
+      toggle(c,on){const w=on===undefined?!el._classes.has(c):!!on; if(w)el._classes.add(c); else el._classes.delete(c);},
+      contains(c){return el._classes.has(c)},
+    },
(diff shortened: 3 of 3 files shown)


nesquena (approved):

Independent review of the #4658 fix: in **Transparent Stream** mode, collapsed tool rows regressed to showing only the bare tool name (`read_file`, `terminal`, …) with no inline summary — you had to expand every row to see what each call did. Affected both live streaming and persisted/old sessions on reload (shared render path).

### What this ships

`static/ui.js` (+38), `tests/test_issue4658_transparent_collapsed_preview.py` (+205), `CHANGELOG.md` (+4). Agent-authored. Closes #4658.

Fresh `nousresearch/hermes-agent` tarball: this is pure WebUI tool-card **presentation** — the summary is derived client-side from the tool call's args/target. No agent protocol, config, or session-schema surface. Webui-only. ✓

### Root cause + fix — traced

Two prior changes interacted: `69072ac34` moved `buildToolCard`'s collapsed preview to an arg summary, and `6951d10a` (#4411) then blanked it for the common arg/shell case (`previewText===argPreview`) — safe in the **worklog** view because #4411 also changed the row *name* to a target-carrying action label. But `_decorateTransparentEventRow` overrides the name back to the **bare** `_toolShortName`, so transparent rows had neither the target-carrying label nor a preview → bare rows.

The fix populates the existing `.tool-card-preview` span in the transparent decorator ([ui.js:8451](static/ui.js#L8451)) from a new `_transparentToolSummary(tc)` ([ui.js:8171](static/ui.js#L8171)):
- progress text (`tc.preview`) while still running → else
- `_toolVisibleTargetLabel(tc)` (the call's **target** — path basename / command / query / skill name, via `_toolTargetLabel`) → else
- `_formatToolArgPreview(tc.args)` → else `''`.

All paths are length-bounded (`_shortToolLabel(..., 160)`) and derived from **args/target, never the result JSON** (the output stays in the expandable detail body) — honoring `69072ac34`'s "keep collapsed previews quiet" intent while restoring "understand the call without expanding." A no-target call yields an empty preview (no invented noise).

### Security audit

The summary is written via **`previewEl.textContent = summary`** — not `innerHTML`. Tool targets are model/user-controlled (file paths, shell commands, search queries can contain `<`, `>`, `&`, quotes), but `textContent` escapes by construction, so nothing can inject markup. **jsdom proof (4/4):** a target of `/tmp/.py` set via `textContent` is preserved literally, rendered HTML-escaped (`&lt;img`), injects **no** live `` element, and re-setting is idempotent (single text node). The only `removeAttribute('hidden')` un-hides the span; no new HTML sink. ✓

- **Idempotent across status re-decoration:** `_decorateTransparentEventRow` re-runs on every status update; it **sets** `textContent` (and clears to `''` when there's no summary), never appends — so re-decoration can't duplicate or stale the preview. Confirmed in the jsdom check.
- **Compact/worklog unaffected:** the change lives entirely in the `type==='tool'` branch of `_decorateTransparentEventRow` (the transparent decorator that adds `.transparent-event-row`); the worklog view keeps its action-label name and never enters this path.
- **CSS:** `.transparent-event-row .tool-card-preview` is already un-hidden, so no style change is needed (and the `removeAttribute('hidden')` covers the case where `buildToolCard` marked the span hidden).
- **All referenced helpers exist** (`_toolVisibleTargetLabel`, `_formatToolArgPreview`, `_shortToolLabel`, `_toolShortName`); the two optional ones are `typeof`-guarded.

### Edge-case matrix

| Row | Collapsed inline summary |
|---|---|
| `read_file` / write | file basename (`config.py`) ✅ |
| `terminal` | the command (`git fetch origin --quiet`) ✅ |
| `search_files` | the query/pattern (`buildToolCard`) — not result JSON ✅ |
| `skill_view` | skill name (`opencode-review-agents skill`) ✅ |
| running tool with progress text | the progress preview ✅ |
| no-target call | empty preview (no invented noise) ✅ |
| target with HTML metacharacters | escaped via `textContent` (no injection) ✅ |
| status re-decoration (Running→Completed) | idempotent, summary preserved ✅ |
| Compact Worklog view | unchanged (action-label name) ✅ |
| persisted/old session reload | same decorator path → restored ✅ |

### Tests

- `tests/test_issue4658_transparent_collapsed_preview.py` — node-driven against the **real** `buildToolCard` + `_decorateTransparentEventRow`: read_file→file, terminal→command, search→pattern (and **not** result JSON), skill→skill name, no-target→empty (5 tests).
- Preview/transparent/anchor blast-radius (`test_tool_card_preview_summary`, `test_issue4539_transparent_tool_settle`, `test_issue4622_tool_card_restore_guard`, `test_issue3820_chat_activity_display_mode`) — **77/77 pass**.
- **Full suite: 9908 passed / 132 skipped / 0 failures** in 205s (deselected the pre-existing darwin CRLF flake).
- **JS gate:** `node --check static/ui.js` clean; ESLint runtime guard exit 0.
- **Independent jsdom harness: 4/4** (XSS-escaping + idempotency, above).
- **CI:** no checks were reported on the branch at review time; my local full suite + JS gates substitute.

- `_transparentToolSummary` and the worklog action-label both derive from the same `_toolTargetLabel`/`_formatToolArgPreview` source, so collapsed previews stay consistent between the two views — good. If a future tool type needs a bespoke summary, that's the single place to extend.
- Scoped deliberately to the collapsed-row preview; #4622 (expanded-output truncation/diff) is the same #4411 window but a distinct symptom, correctly left out of scope.

### Recommendation

A precise, well-scoped fix: the transparent decorator now rebuilds a quiet, target-based one-line summary into the existing preview span, restoring self-describing collapsed rows in exactly the view (#4411 renamed to the bare tool name) that lost both the label and the preview. It's XSS-safe by construction (`textContent`, proven 4/4 in jsdom incl. an injection payload), idempotent across status re-decoration, derives from the call target rather than result JSON, and leaves the compact/worklog view untouched. Webui-only presentation, tests green. Ship.

nesquena-hermes:
Shipped in **v0.51.574** 🚀 — released via the maintainer release flow (squash-merged as #4663 onto current master).

Collapsed tool rows in the Transparent Stream view now show their one-line target summary again (file path / command / search query / skill name) next to the tool name, restoring "understand the call without expanding it" while keeping the raw result out of the collapsed row. Covers both live streaming and reloaded/old sessions.

**Review trail:** independent review (nesquena agent) **APPROVED**; greptile **4/5 safe-to-merge**; maintainer deep review + live-DOM verification of the real `buildToolCard` + `_decorateTransparentEventRow` path (all 5 tool cases correct, idempotent). One gate refinement applied on top: removed the generic `_formatToolArgPreview` fallback so an args-but-no-target call (e.g. `terminal {workdir}`, unknown tool `{mode:"dry-run"}`) yields an **empty** collapsed preview instead of dumping a raw arg snippet — matching this PR's own "no invented noise" intent — plus 2 regression tests for that. Codex re-gate **SAFE TO SHIP**, full suite **10050 passed**.

Thank you for the thorough root-cause writeup — it traced the #4411 + `69072ac34` interaction precisely. Closes #4658.

PR #4669, opened 2026-06-22
fix: wrap markdown source previews

### LINKED ISSUE #4667: Markdown source previews force horizontal scrolling for normal prose
## Bug Description
Markdown source previews can force a horizontal scroll bar even when the content is ordinary prose that should wrap.

This affects the inline markdown-source viewer shown with the small `MD` header and Copy button. It does not appear to be the normal rich markdown renderer.

## Steps to Reproduce
1. Send or open a fenced markdown block such as ```md with several normal prose lines.
2. View the rendered inline source preview in chat.
3. Observe the preview on desktop.

## Expected Behavior
Normal markdown prose should wrap to the available message width.

Horizontal scrolling should be reserved for genuinely wide content such as code, tables, diffs, or long unbreakable strings.

## Actual Behavior
The entire markdown-source preview behaves like a horizontally scrollable code block, so normal prose stays on long lines and requires side-scrolling to read.

## Proposed Solution
Treat fenced `md` / `markdown` / `mdx` source previews as a dedicated markdown-source style that wraps prose, while keeping existing scroll behavior for real code-oriented blocks.

## Environment
- Surface: chat inline markdown-source preview (`MD` header + Copy button)
- Observed on: desktop WebUI

TomBanksAU:
## Summary
Fix markdown source previews that show ordinary prose inside a horizontally scrolling code block.

## What changed
- tag fenced `md` / `markdown` / `mdx` source previews with a dedicated `md-source-block` class in both fenced-code renderers
- wrap those markdown-source blocks in CSS while leaving real code/diff behavior unchanged
- add regression coverage for the dedicated class and wrap rules
- add an Unreleased changelog entry

## Testing
- `uv run --with pytest pytest tests/test_markdown_source_code_blocks_wrap.py tests/test_mobile_markdown_wrapping.py -q`
- `npm run lint:runtime`
- `uv run --with ruff python3 scripts/ruff_lint.py --diff origin/master`
- `node --check static/ui.js`

Closes #4667


TomBanksAU:
Addressed the changelog concern. I removed the Unreleased entry from this PR and kept the fix scoped to the renderer/CSS/test changes only. Re-ran focused verification after the amend:\n\n- uv run --with pytest pytest tests/test_markdown_source_code_blocks_wrap.py tests/test_mobile_markdown_wrapping.py -q\n- npm run lint:runtime\n- uv run --with ruff python3 scripts/ruff_lint.py --diff origin/master\n- node --check static/ui.js

nesquena-hermes:
Reviewed the diff at HEAD (`6cbdd5af8`) against `origin/master` — `static/ui.js`, `static/style.css`, and the new test file. The fix is correct and well-scoped, and the CHANGELOG entry is gone after the amend. A few confirmations from reading the surrounding code:

**Case-handling is safe.** Both fence parsers lowercase the language tag before the regex runs, so `MD`, `Markdown`, `MDX` all normalize correctly:

```js
// ui.js:346 and ui.js:4744
let lang = langMatch ? (langMatch[1]||'').trim().toLowerCase() : '';
const preClass = /^(md|markdown|mdx)$/.test(lang) ? ' class="md-source-block"' : '';
```

The regex is anchored (`^...$`), so it won't over-match `mdc`, `mermaid`, etc. Good.

**The desktop/mobile split is exactly right.** The reason this only reproduced on desktop is that mobile already force-wraps every `pre` inside `@media(max-width:700px)`:

```css
/* style.css:1697-1701 — mobile only */
@media(max-width:700px){
  .msg-body pre,.preview-md pre{white-space:pre-wrap !important;overflow-x:hidden !important;overflow-wrap:anywhere !important;}
  .msg-body pre code,.preview-md pre code{white-space:inherit !important;...}
  ...
}
```

On desktop the default `.msg-body pre` (style.css:1662) uses `overflow-x:auto`, so prose-only md source overflowed. The new rules at style.css:1664-1666 add the desktop wrap scoped to `.md-source-block`, mirroring the mobile shape without `!important`. Because `.msg-body pre.md-source-block` is more specific than `.msg-body pre`, it correctly wins over the `overflow-x:auto` default — no specificity fight.

**Both surfaces are covered.** The class is emitted in `_renderUserFencedBlocks` (user-typed, ui.js:324) and `renderMd` (assistant + panels, ui.js:4592), and the CSS scopes to both `.msg-body` (chat) and `.preview-md` (notes/memory previews via panels.js:4174). That matches the issue's "MD header + Copy button" inline source viewer.

One minor thought, not blocking: the diff/patch branch already routes to `pre.diff-block`, and json/yaml/csv get their own renderers, so `preClass` is only ever applied on the plain-`<pre>` fallback path — which is precisely the intended target. No interaction with the special renderers.

LGTM from a read-only review standpoint. Closes #4667 cleanly.

Issue #4662, opened 2026-06-22
Profile switching: polished loading states + faster switch (phased)


nesquena-hermes:
## Summary

Profile switching feels slow and visually unpolished. Two independent problems, confirmed by investigation against `master`:

1. **It looks broken** because the conversation list (and the right-hand workspace panel) have no loading state — they keep showing the *previous* profile's content until the new fetch resolves (~1s), then swap. The profile chip spins, but nothing else gives feedback.
2. **It is genuinely slow** — the switch fires a serial chain of requests and the dominant one, `/api/sessions`, runs cold after every switch. Live `RequestDiagnostics` show `get_cli_sessions` at 3.8s typical (up to ~20s under lock contention), plus ~440ms of YAML reparse in `load_settings`, plus a per-row `settings.json` reload during response serialization.

This is the tracking issue for the full fix. Work proceeds **one phase at a time**, each as its own PR, gated and shipped before the next begins. Perceived-speed wins (low risk) ship first; the real latency kill (higher blast radius) ships last and separately.

> **✅ STATUS (2026-06-25): all four phases shipped. Cold profile switch went from ~5s typical (≤20s worst case) → ~1s after Phases 1–3, and the cold `/api/sessions` rebuild then dropped a further ~670–880ms → ~100–200ms in Phase 4 (transcript-parse memoization). Both original problems (looks-broken + genuinely-slow) are resolved on `v0.51.640`+. See the closing comment for the full ship list.**

Full investigation notes live in the private workspace (`docs/profile-switch-performance-plan.md`).

---

## How a switch works today

`switchToProfile()` (`static/panels.js:5867`) →
`POST /api/profile/switch` → `startGatewaySSE()` → apply model defaults →
`await renderSessionList()` (fetches `/api/sessions` + `/api/projects`, **leaves stale rows up during the fetch**) →
`await loadDir('.')` (refreshes workspace tree, **leaves the stale tree up**) →
`_profileSwitchPanelLoad()` → `_refreshProfileSwitchBackground()` (already fire-and-forget — good precedent).

Backend `POST /api/profile/switch` (`api/routes.py:11449`, `api/profiles.py:1144`): cookie/thread-local scoped switch, `invalidate_models_cache()`, and a **synchronous** `restart_watcher_for_profile()` whose `old_watcher.stop()` joins up to 3s.

## Evidence — where the time goes

Cold `/api/sessions` stage breakdown (live diagnostics):

```
[...]

PR #4671, opened 2026-06-22
Profile switch: polished loading skeletons (sidebar + workspace) — #4662 phase 1

nesquena-hermes:
## Summary

Phase 1 of #4662 (profile switching: polished loading states + faster switch). **Frontend-only, low-risk.**

Switching profiles used to leave the **previous** profile's conversation list and workspace file tree on screen until their data finished loading (~1s), with no feedback beyond a spinner on the profile chip. This adds clean, theme-aware loading skeletons that appear the instant a switch begins — so you never see the wrong profile's content, and you get consistent loading feedback across the whole window.

This is the first of four phases tracked in #4662. Phase 1 targets *perceived* speed (the felt UX); Phases 2–4 (backend caching, pre-warming) target *actual* latency and ship separately.

## What changed

- **`static/style.css`** — skeleton system: single-line session rows (title bar + right-aligned timestamp bar) grouped under label bars, and file-tree rows (glyph + name + size bar). `@keyframes skeletonSheen` (background-position shimmer, 1.25s) + `skeletonFadeIn`. Light + dark `--skeleton-base`/`--skeleton-sheen` tokens. `prefers-reduced-motion` disables the sheen (static fill).
- **`static/sessions.js`** — `showSessionListSkeleton()` builds the skeleton from `_SESSION_SKELETON_GROUPS`; `_sessionListSkeletonActive` flag is cleared in `renderSessionListFromCache()` on the real render. Also tears down virtual-scroll window state + cancels the queued scroll RAF, and `_scheduleSessionVirtualizedRender()` bails while the skeleton is up (see gate fixes).
- **`static/workspace.js`** — `showWorkspaceTreeSkeleton()` builds the tree skeleton; `clearWorkspaceTreeSkeleton()` clears it on the no-workspace path (only when a skeleton is present, so it never clobbers a real render).
- **`static/panels.js` `switchToProfile()`** — shows both skeletons up front (before the awaited `/api/profile/switch` POST) so stale content clears immediately; dismisses any open inline-rename / row action menu first (they'd otherwise block the clearing render); restores real content from cache on failure; early-returns on a no-op self-switch.

## Visual review

Reviewed with Nathan over Telegram across 1920 / 1440 / 1280 / 390 viewports + dark mode; iterated until the skeleton matches the real layout's rhythm (single-line rows, group headers, aligned timestamp/size bars) so the swap to real content has no layout jump. Approved to proceed.

> **Clarification (@rodboev review):** the skeleton is a **fixed representative shape** (3 group labels + 8 rows), not sized to the target profile's conversation count — at switch *start* we don't yet know that count (the awaited load is what fetches it). It mirrors the real row *anatomy* (so there's no layout jump), but it is not content-adaptive. The genuinely-empty/zero-conversation profile case (skeleton implies data that never arrives) is tracked as a Phase-1.5 follow-up on #4662.

### Review fixes folded in (commit 737b112)
- **Spurious "Request timed out" toast (@rodboev):** the `/api/profile/switch` POST now passes `timeoutToast:false`, so a superseded or transient-but-eventually-successful switch can't pop `api()`'s generic timeout error while the real switch completes. The generation-guarded `catch` is now the single source of truth for switch failure.
- **Missing generation re-guard (greptile):** the in-progress-session branch now re-checks `_switchGen` after its awaited `renderSessionList()`, mirroring the no-messages branch — a rapid second switch can't have its workspace skeleton cleared / a stale toast popped by a slower earlier one.
- **Group-label opacity (greptile):** labels now fade in via `skeletonFadeInDim` (ends at `opacity:.5`) instead of the shared `skeletonFadeIn` (ends at `opacity:1` + `fill-mode:both`), which was overriding their intended resting 50%.
- +4 static-source regression assertions locking all three; 29/29 targeted tests green.

## Gates

- **Codex (regression) — 2 rounds, both findings fixed:**
  - Round 1 [CORE]: the parallelized `loadDir('.')` ran *before* the `_switchGen` stale-switch guard; empty-session switches reuse the session id, so a superseded switch could paint the wrong profile's tree. Reverted to guard-before-workspace-render ordering (the up-front skeletons already give the cross-surface feedback; the parallel gain was marginal).
  - Round 2 [SILENT]: on long **virtualized** sidebars, the skeleton left `data-session-virtual-*` state + a queued scroll RAF active, so a pending `_scheduleSessionVirtualizedRender()` could repaint the previous profile's cached rows over the skeleton. Now cancels the RAF, clears the virtual markers, and guards the scheduler on `_sessionListSkeletonActive`.
- **Opus (quality) — all findings fixed:** workspace-skeleton strand when the new profile has no workspace (blocker), session-list skeleton strand when a rename/menu was open, and the no-op self-switch flash.
- **Tests:** 24 new (node-driven behavioural builders + static wiring/CSS assertions, including the strand + virtualization fixes). Full local suite green except the known process-group isolation cascade (`test_profile_skills_stats` / `test_tls_support` / `test_v050259_sessiondb_fd_leak`) which pass in isolation and are untouched by this frontend-only change — CI is authoritative.

## Live verification

Drove a real switch in a seeded 2-profile test env: switch fires → skeleton appears instantly (0 stale rows) → ~1s later resolves to the new profile's real rows. Also verified the no-workspace skeleton clear and the self-switch guard directly.

Closes nothing yet — #4662 stays open to track Phases 2–4.


nesquena (approved):
## Review — end-to-end ✅ (approved after fix; pushed a strand-on-failure fix)

Independent review of Phase 1 of #4662: polished loading skeletons for profile switching. Switching profiles left the **previous** profile's conversation list + workspace file tree on screen until their fetches resolved (~1s); this shows theme-aware loading skeletons the instant a switch begins. Frontend-only, perceived-speed (Phases 2–4 target real latency separately).

### What this ships

`static/style.css` (+51), `static/sessions.js` (+72), `static/workspace.js` (+59), `static/panels.js` (+55/-2), two test files (24 tests), `CHANGELOG.md`. Agent-authored; went through Codex (2 rounds) + Opus gates. I pushed one additional fix (below).

Fresh tarball: this is **pure WebUI loading-state presentation** — skeleton DOM built client-side, no backend/agent/config/session-schema surface. No cross-tool concern. ✓

### Security audit

No XSS surface. Both builders (`showSessionListSkeleton`, `showWorkspaceTreeSkeleton`) construct DOM **programmatically** from static specs (`_SESSION_SKELETON_GROUPS` / `_WS_SKELETON_ROWS`) — only `className`, numeric `style.width`/`paddingLeft`, and `aria-hidden`. **No user/model data** ever reaches them; the only `innerHTML` use is `=''` (clearing). ✓

### End-to-end trace — `switchToProfile()`

- **No-op self-switch:** `if (name && name === S.activeProfile) return;` before any skeleton — kills the Opus "skeleton→restore flash" for `activateCurrentProfile()`. ✓
- **Up-front skeletons:** dismisses `_renamingSid` + `closeSessionActionMenu()` (else `renderSessionListFromCache()` early-returns and strands the skeleton — Opus), then `showSessionListSkeleton()`; workspace skeleton only when the panel is open (`_workspaceVisibleAtStart`). ✓
- **Stale-switch ordering (Codex R1):** in the no-messages branch the workspace `loadDir('.')` runs **after** `if (_switchGen !== _profileSwitchGeneration) return;`. Because empty-session switches reuse the session id and `loadDir` paints with only a session-id check, starting it before the guard could let an older switch paint over a newer one; `renderSessionList()` has its own internal generation guard, so awaiting it first is safe. Verified the guard precedes `loadDir`. ✓
- **No-workspace clear (Opus blocker):** both success branches call `clearWorkspaceTreeSkeleton()` when the new session has no workspace, so the skeleton can't strand. ✓
- **Virtualized clobber (Codex R2):** `showSessionListSkeleton()` cancels `_sessionVirtualScrollRaf` and deletes all five `data-session-virtual-*` markers, and `_scheduleSessionVirtualizedRender()` early-returns `if(_sessionListSkeletonActive)` — so a queued scroll RAF can't repaint the previous profile's cached rows over the skeleton. The real render (`renderSessionListFromCache`) clears `_sessionListSkeletonActive`. ✓
- **Streaming-row invariant:** the issue requires not wiping an actively-streaming optimistic row. The skeleton builders are called **only** from `switchToProfile` (confirmed — the sole callers are panels.js:5907/5909), never on routine streaming refreshes; and a profile switch is a deliberate context change where the stream continues server-side (per-client cookie/TLS scoped) while the sidebar moves to the new profile. The transcript (`#messages`) is untouched. ✓

### What I caught — workspace skeleton stranded on switch *failure*

The success path clears the workspace skeleton when there's no workspace, but the **catch block** only restored the tree `if (S.session && S.session.workspace)` — with no `else`. So a switch **failure** while the workspace panel is open and the (still-current) previous session has **no** workspace left the up-front skeleton shimmering forever — the same strand class the gates fixed, in the one path they missed. The session list was already restored on failure via `renderSessionListFromCache()`; only the workspace had the gap.

### What I pushed — commit 0e69775d

Mirror the success-path no-workspace handling in the failure path:
```js
} else if (_workspaceVisibleAtStart && typeof clearWorkspaceTreeSkeleton === 'function') {
  clearWorkspaceTreeSkeleton();  // no workspace to restore → clear the skeleton (#4662)
}
```
Added a static regression assertion that the catch block clears the workspace skeleton. `clearWorkspaceTreeSkeleton()` only empties `#fileTree` when it still holds a `.skeleton-tree`, so it can't clobber a real render.

- **CSS:** light + `:root.dark` `--skeleton-base`/`--skeleton-sheen` tokens (with `color-mix` fallbacks), `@keyframes skeletonSheen`/`skeletonFadeIn`, and a `prefers-reduced-motion: reduce` block that sets `animation:none` + static fill (accessibility contract honored).
- **Idempotency / no double-skeleton:** `_sessionListSkeletonActive` flag prevents stacking and drives the clear-on-render path.
- **`clearWorkspaceTreeSkeleton` guard:** `.skeleton-tree` presence check means it never clobbers a real tree.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Switch to another profile | skeletons instant, 0 stale rows, resolve to new profile ✅ |
| Switch to already-active profile | early-return, no flash ✅ |
| New profile has no workspace (success) | workspace skeleton cleared (both branches) ✅ |
| Switch fails, prev session has workspace | list + tree restored from cache ✅ |
| Switch fails, prev session has no workspace, panel open | workspace skeleton cleared (**my fix**) ✅ |
| Inline-rename / row menu open at switch | dismissed first, so skeleton isn't stranded ✅ |
| Long virtualized sidebar | RAF cancelled + virtual markers cleared + scheduler gated ✅ |
| Superseded (rapid double) switch | generation guard returns early; active switch renders ✅ |
| prefers-reduced-motion | static bars, no sheen ✅ |
| Active stream in current tab | builders only run on switch; transcript untouched; stream continues ✅ |

### Tests

- `test_issue4662_profile_switch_skeleton_behaviour.py` + `_static.py` — **25/25** (24 + my new failure-path assertion): node-driven builders (3 labels + 8 rows + titles/stamps + flag + scroll; 8 tree rows + glyphs + 7 sizes) and static wiring for every gate fix + CSS classes/keyframes/reduced-motion/theme tokens.
- Profile-switch / sidebar / session-list blast-radius — **150 passed, 2 skipped**.
- **Full suite (after my fix): 9961 passed / 132 skipped / 0 failures** in 202s (deselected the pre-existing darwin CRLF flake).
- **JS gates:** `node --check` on all three JS files clean; ESLint runtime guard exit 0.
- **CI on the PR head:** browser-smoke + lint green; test shards were pending at review time (my local full suite substitutes). PR documents live verification across 1920/1440/1280/390 + dark mode (reviewed with Nathan).

- CSS/DOM visual behavior can't be fully API-tested; coverage relies on the node-driven structural builders + static wiring + the PR's documented multi-viewport live verification — appropriate for a skeleton-rendering change.
- Phase 1 is correctly scoped to perceived speed; the actual `/api/sessions` cold-fetch latency (the 3.8s `get_cli_sessions` in the issue) is deferred to Phases 2–4, so users still wait ~1s for real data — by design.

### Recommendation

A clean, low-risk perceived-speed win: stale profile content is replaced by theme-aware, reduced-motion-aware skeletons the instant a switch begins, built from static structure (no XSS surface) and torn down on every resolution path. I verified all five gate findings (Codex stale-switch ordering + virtualized clobber; Opus no-workspace strand, rename/menu strand, self-switch flash) are fixed, confirmed the streaming-row invariant holds (builders only fire on switch), and closed the one remaining strand — the workspace skeleton on a switch *failure* with no workspace — with a small fix + regression test. Ship.

nesquena-hermes:
## Greptile feedback addressed ✅

Reviewed all three Greptile comments and applied fixes (pushed in `d65e02b1`, on top of Nathan's `0e69775d`):

**[P1] Workspace skeleton strands on a failed switch when the session has no workspace** — ✅ Fixed (by @nesquena in `0e69775d`). The `catch` block now mirrors the success-path no-workspace handling: reload the tree when a workspace is bound, otherwise `clearWorkspaceTreeSkeleton()`, so the up-front skeleton can't shimmer forever after a failed switch with the panel open. I kept this commit as-is and built on top of it (rebased my remaining fixes onto it rather than force-pushing over it).

**[P2] `_clearSessionListSkeletonFlag` is dead code** — ✅ Fixed. Removed the function; the flag is reset inline in `renderSessionListFromCache()` (the only place it's needed), so the helper had no call site.

**[stagger] `.skeleton-row:nth-child(N)` delays skip most rows** — ✅ Fixed. Correct catch: group-label siblings are interleaved with rows, so rows actually land at child positions 2, 4–6, 8–11 — meaning rows 9/10/11 got no delay and the stagger was broken. Replaced the CSS `:nth-child` rules with an inline per-row `animationDelay` set in JS (`rowIndex * 0.025s`, capped at 0.2s), so every row gets its correct stagger regardless of how rows are distributed across groups.

**Tests:** added a behavioural assertion that every skeleton row carries a real inline stagger delay (and the last row isn't skipped); the failure-path test already requires the workspace-skeleton clear. **26 tests pass** locally; JS syntax-checked.

Thanks for the review — the stagger one in particular was a genuine silent bug in the changed code.

nesquena-hermes:
📌 **Status note for whoever picks this up next** (so it's not lost):

- ✅ CI green (11/11), `MERGEABLE/CLEAN`, labeled `review-requested` (do not auto-merge).
- ⚠️ **nesquena's APPROVED review is on `0e69775d`, one commit behind the current head `d65e02b1`.** The 3 changes since that approval are the small Greptile follow-ups (remove dead `_clearSessionListSkeletonFlag`, replace the broken CSS `:nth-child` stagger with an inline JS `animationDelay`, +1 test) — low-risk, but **not yet covered by a human review**.
- **Before merge:** re-request approval on the current head, OR fold this into the broader engineer review being lined up. Don't merge on the stale approval.

Tracking issue #4662 stays open for Phases 2–4 (backend caching / pre-warming).

nesquena-hermes:
## Requesting top-contributor review — code + manual browser regression pass

@rodboev @franksong2702 — this is a self-built (nesquena-hermes) **feature** PR, and Nathan wants an independent top-contributor review before it ships. Would one (or both) of you be up for running it through? Two parts:

### 1. Code feedback
A line-by-line look at the diff (+641/-3, 7 files). The surface that matters most for regressions:

- **`static/sessions.js` (+74)** — sidebar skeleton injection/teardown. Crown-jewel-adjacent: confirm the skeleton can't leak into or race the real `/api/sessions` render, and that teardown always fires even if the profile switch errors or is interrupted mid-flight.
- **`static/workspace.js` (+59)** — file-tree skeleton. Earlier review rounds (Greptile) caught a parallel `loadDir` ordering issue and a virtual-scroll teardown gap — please sanity-check those are genuinely closed, not just papered over.
- **`static/panels.js` (+60)** — `switchToProfile` orchestration: the skeleton show/clear lifecycle. Confirm no path leaves a skeleton stuck on screen (switch to a profile with zero sessions; rapid back-to-back switches; switch while a stream is active).
- **`static/style.css` (+47)** — `prefers-reduced-motion` correctness (static bars, no shimmer) and light/dark theme adaptation.

### 2. Manual browser regression testing (the main ask)
Please actually drive it in a browser and confirm we haven't regressed profile switching:

- **Switch between profiles** with differing session counts (many ↔ few ↔ zero) — skeleton appears instantly, mirrors the real layout (grouped single-line conversation rows + file-tree rows), then cleanly swaps to real content with no layout jump and no wrong-profile flash.
- **Rapid back-to-back switches** — no stuck/orphaned skeleton, no stale content from the prior profile.
- **Switch while a turn is streaming** in the current session — streaming/scroll behavior is unaffected.
- **Workspace file tree** — skeleton clears correctly and the tree loads; expand a deep dir afterward to confirm virtual-scroll still behaves.
- **`prefers-reduced-motion: reduce`** (OS setting) — static skeleton, no animation.
- **Light + dark themes**, desktop + mobile (390px) — skeleton looks right in all four.

Nathan has already eyeballed and approved the visual across 1920/1440/1280/390 + dark, so this is specifically about **code correctness + no functional regression**, not the look.

CI is green and the branch is freshly rebased on master (MERGEABLE), so you can pull and run it clean. Drop your findings here — code comments inline, and a quick "browser pass: ✅/❌ + what I saw" summary. Thanks both 🙏

nesquena-hermes:
## Investigation of @rodboev's review feedback (folding in net-new detail)

Thanks @rodboev — drove each of your three observations into the code. Two are real and actionable; one is a description-vs-code mismatch you caught exactly right. Net-new findings below, plus the line numbers so the fixes are unambiguous.

### 1. The skeleton is fixed at 11 lines regardless of conversation count — you're right, and it's a description bug

Confirmed: `_SESSION_SKELETON_GROUPS` (`static/sessions.js` ~L3505) is a **hardcoded** 3 group-labels + 8 rows = **11 lines**, and `showSessionListSkeleton()` has no count / empty / zero branch. So a profile with 1 conversation and a profile with 30 render the identical skeleton — exactly what you saw.

As your agent found, the fixed shape **is** intentional in the current revision. The real problem is that the PR description ("iterated until the skeleton matches the real layout's rhythm") **oversells** it — it implies a content-adaptive `many ↔ few ↔ zero` skeleton that the code does not implement. The tracking issue (#4662) only ever specced "polished skeleton/spinner, not tacky," not adaptive sizing, so nothing is broken against spec — but the description should be corrected.

Why fixed was chosen: at switch *start* we don't yet know the target profile's session count (that's what the awaited load fetches), so a truly adaptive skeleton needs a cached per-profile count we don't currently keep cheaply.

**The one case that genuinely misleads is zero.** A brand-new / empty profile flashes an 11-line "content is loading" skeleton and then resolves to an empty list — the skeleton implies data that never arrives. Recommendation:
- **Now (this PR):** fix the description to say it's a fixed representative skeleton, not content-adaptive.
- **Follow-up:** special-case the empty/zero profile (render an empty-state placeholder or no skeleton), and optionally size from a cached per-profile session count if we start tracking one. I'll file this as a Phase-1.5 follow-up on #4662 unless you'd rather fold the zero-case into this PR.

### 2. "Request timed out. Please try again." appearing independently of success — root-caused

Your observation that the error and the success path are indistinguishable (conversations show up either way) is the tell. `switchToProfile()` calls `api('/api/profile/switch', …)` (`static/panels.js` ~L5953) with **no explicit `timeoutMs`**, so it inherits the generic `api()` default: **30s abort + `timeoutToast:true`** (`static/workspace.js` L5–6, L87). That toast fires on *any* abort/timeout in the switch sequence, and critically it is **not gated on switch success or supersession**:

- `switchToProfile()` bumps `_profileSwitchGeneration` and checks `_switchGen !== _profileSwitchGeneration` **after** the awaited POST (panels.js ~L5954) — but the timeout toast is thrown from *inside* `api()` **before** that guard runs. So a superseded earlier switch (or a slow-but-eventually-successful one) still pops the red error toast while the real switch completes and the rows render. That's the "no difference between error and success" you're seeing.

**Fix direction:** route the switch POST through a toast-suppressed call (`timeoutToast:false`) and surface failure only after the `_switchGen` guard, or have `api()` skip the toast when the upstream signal was aborted by a newer generation. Either way the error toast should be tied to the *current* switch actually failing, not to a transient/superseded abort.

### 3. The uneven 1–5s latency on a brand-new, single-conversation profile — likely the watcher restart, not session loading

This is the most useful net-new finding for your benchmarking. The latency is **not** proportional to conversation count (which is why a single-convo profile is still slow), so it isn't the session-list fetch. The synchronous backend switch path (`api/routes.py` L11612–11636) does three things in series:

1. `switch_profile(name, process_wide=False)`
2. `invalidate_models_cache()`
3. `restart_watcher_for_profile(name)` ← **prime suspect**

The gateway watcher teardown+respawn has variable timing unrelated to how much chat data the profile has, which fits your "sometimes <1s, sometimes 5s" on an empty profile. That's squarely Phase 2–4 territory in #4662 (actual latency, vs this PR's perceived-latency skeletons), so it's out of scope for Phase 1 — but please drop your CDP **min/avg/max/mean** numbers here when you have them. If the watcher restart dominates, the Phase-2 fix is to make it async / debounced / off the request's critical path rather than blocking the switch response.

---

**Summary for this PR (Phase 1, frontend-only):**
- ✅ Skeletons themselves are sound — the fixed shape is fine; **correct the description** (drop the "matches the real layout's rhythm" / adaptive implication).
- 🔧 Worth folding in here: gate the timeout toast on the current switch generation so success/superseded switches stop showing a spurious error (small, in-scope frontend fix).
- ➡️ Follow-ups on #4662: empty/zero-profile skeleton case (Phase 1.5) and the watcher-restart latency (Phase 2).

Want me to (a) push the description fix + toast-suppression into this branch now, and (b) open the two #4662 follow-ups? Holding for your benchmark numbers on #3 regardless.

nesquena-hermes:
Pushed the review fixes (commit `737b112`) and corrected the description.

**Folded in from @rodboev + greptile review:**
- ✅ Spurious "Request timed out" toast — switch POST now `timeoutToast:false`; the generation-guarded `catch` is the single source of truth for failure.
- ✅ Missing generation re-guard in the in-progress-session branch (re-checks `_switchGen` after `renderSessionList()`).
- ✅ Group-label opacity — dedicated `skeletonFadeInDim` keyframe (settles at `.5`) instead of the shared fade that forced full brightness.
- ✅ +4 static regression assertions; **29/29** targeted tests green, JS syntax clean.
- ✅ Description corrected to say the skeleton is a fixed representative shape, not content-adaptive.

**Follow-ups opened on #4662** (out of scope for this frontend-only Phase 1):
- #4717 — Phase 1.5: empty/zero-conversation profile skeleton case
- #4718 — Phase 2: cut actual switch latency (suspect synchronous `restart_watcher_for_profile()` on the request path) — @rodboev, drop your CDP min/avg/max/median here
- #4719 — Phase 5: preload profile on hover (your idea 🙏)

Ready for re-review.

rodboev:
Re-test Results with `737b112f` (latest PR commit)

Three claimed fixes, all verified:

**1\. Spurious "Request timed out" toast: FIXED**  
Rapid triple-switch (test/default/test with 200ms gaps) produced exactly 1 toast ("Switched to profile: test-ep-sprint31") and zero error/timeout toasts. The `timeoutToast: false` parameter at `panels.js:5958` suppresses `api()`'s generic timeout, and the generation-guarded catch block is now the only error surface.

**2\. Missing generation re-guard in the in-progress branch: FIXED**  
`panels.js:6075` adds `if (_switchGen !== _profileSwitchGeneration) return;` after `await renderSessionList()` in the `sessionInProgress` branch, with a comment explaining the race (a newer switch could start while `renderSessionList()` is in flight). Mirrors the existing guard at line 6095 in the no-messages branch.

**3\. Group-label opacity: FIXED**  
New `skeletonFadeInDim` keyframe animates from `opacity: 0` to `opacity: 0.5`. Measured directly: at 100ms labels are mid-animation (opacity 0.39), at 250ms+ they settle at `0.5`. The shared `skeletonFadeIn` (ending at `1.0`) is no longer applied to group labels.

nesquena-hermes:
Shipped in **v0.51.594** (Release VA) 🚀 — profile-switch loading skeletons are live in prod.

Thank you @rodboev for the deep review + hands-on re-test, and the three fixes you caught (timeout-toast suppression, generation re-guard, group-label opacity).

This went through the full authoritative gate after the human review, and the dual Codex+Opus loop earned its keep — across six passes it reproduced and we closed four progressively-subtler timing variants of one "previous-profile content races back over the skeleton" bug that neither the human review nor live testing surfaced:

1. A `renderSessionList()` already **in flight** before the switch → invalidated at switch start.
2. **Timer/SSE/cache renders mid-switch** → `renderSessionListFromCache()` bails while the skeleton is up; the flag is cleared only once fresh data is in hand.
3. A render that **starts during the switch window** (before the new-profile cookie is set, so it fetches the OLD profile) → a render embargo drops every payload during the window, on both success and fetch-failure paths, lifted before the switch-owned render / on failure / in a `_switchGen`-guarded `finally`.
4. **Wrong-profile workspace tree** (empty-session switch reuses the same `session_id`) → an unconditional `_wsTreeGen` generation token that `loadDir` re-checks.

Verified: Codex SAFE TO SHIP, Opus SHIP (no MUST-FIX), full suite green, +6 regression tests, and live browser stress tests (hammering the render paths through the switch window showed zero stale-row leak). Prod confirmed serving the skeleton code on v0.51.594.

Phase 2 of #4662 can build on this. 🙏

PR #4674, opened 2026-06-22
feat(workspace-pane): add loaded-tree file/folder filter

### LINKED ISSUE #4673: feat(workspace-pane): add file/folder filter, then backend-backed recursive workspace search
## Feature Description

Add search/filter to the Workspace pane’s **Files** view so users can quickly find files and folders without manually expanding the tree.

There are really two layers to this:

1. a simple **client-side filter** over the currently loaded tree entries
2. a proper **recursive workspace search** that searches child folders across the active workspace, not just folders that have already been expanded in the UI

This issue is about the filter/search improvement only. It does **not** cover browsing above the active workspace root or adding a separate Home/root browser mode.

## Motivation

The current Workspace pane is good for browsing, but not for finding.

If I know roughly what a file/folder is called, I still have to manually expand directories until I reach it. That gets slow in deeper repos and makes the right panel much less useful on mobile or when moving quickly.

A shallow filter over only the currently visible tree would already help, but the more useful behaviour is a deeper search that can find files/folders anywhere under the active workspace.

### Phase 1: lightweight filter in the Files tab
Add a small search/filter input to the Workspace pane that filters the currently loaded tree entries by file/folder name and path.

This would give immediate value with relatively low implementation risk.

Suggested behaviour:
- filters files **and** folders
- matches on filename and relative path
- updates as the user types
- if no loaded entries match, shows an empty state such as:
  - `No loaded files match`
- clearly behaves as a filter over what is already loaded in the tree

### Phase 2: recursive workspace search
Add a backend-backed search mode that recursively searches child folders under the active workspace, including folders that have not yet been expanded in the UI.

Suggested behaviour:
- searches all child folders under the current session workspace
- returns matching files and folders
- clicking a result opens the file or navigates/expands to the folder
- respects existing workspace safety boundaries
- respects hidden/cruft handling rules, or at least has predictable behaviour for them
- caps results and debounces requests so it stays fast on large repos

## Acceptance Criteria

- [ ] The Workspace pane Files view has a visible filter/search control
- [ ] Users can filter currently loaded file-tree entries by typing part of a file or folder name
- [ ] Matching works for both filenames and relative paths
- [ ] The UI clearly handles the “no matches” state
- [ ] The feature can be extended to backend-backed recursive search without redesigning the UI
- [ ] Recursive search mode can find files/folders in child folders that have not been expanded yet
- [ ] Search stays scoped to the active session workspace

## Implementation Notes

I checked the current implementation and do not see an existing workspace-pane search/filter UI.

Relevant files/behaviour:

- `static/index.html`
  - Workspace pane header/tabs are defined here
  - current tabs are `Files`, `Artifacts`, and `Todos`
  - there is no existing file-tree search input in the workspace pane markup

- `static/workspace.js`
  - handles directory loading and workspace tree state
  - browsing is built around loading one directory at a time via the workspace API
  - this is likely why a true recursive search does not exist yet in the pane

- `static/ui.js`
  - tree rendering happens here
  - `_renderTreeItems(...)` renders rows
  - expanded directory children are rendered from `S._dirCache[...]`
  - `_visibleWorkspaceEntries(...)` appears to filter visible entries for hidden/cruft handling, not user search
  - breadcrumbs and the current workspace tree navigation are also rendered here

- `api/routes.py`
  - `/api/list` calls `_handle_list_dir(...)`

- `api/workspace.py`
  - `list_dir(...)` lists a single directory
  - `safe_resolve_ws(...)` enforces the active workspace boundary

Current architecture appears to be **lazy directory listing + client cache**, not recursive indexing:
- root is loaded with `loadDir('.')`
- folders fetch children when expanded
- tree state is cached client-side in `S._dirCache`
- that makes browsing efficient, but means a client-only filter can only search entries the UI has already loaded

That suggests a good staged approach:
1. ship a client-side “filter loaded entries” improvement first
2. then add a dedicated backend search endpoint for recursive workspace search

### Filter only the currently visible rows
Better than nothing, but too limited on larger workspaces because it misses files/folders inside collapsed or not-yet-loaded directories.

### Automatically preload the whole workspace tree
This would make client-side search easier, but could be slow/noisy for large repos and would likely be a worse default than the current lazy-loading approach.

## Duplicate Check

I searched the existing issue tracker with a few phrasings (`workspace search`, `file tree filter`, `workspace filter`, `search files pane`) and did not find an obvious existing issue for this specific workspace-pane file/folder filter/search request.

TomBanksAU:
## Summary
- add a Files-tab filter input to the workspace pane
- filter currently loaded file/folder tree entries by name or relative path
- preserve matching ancestor folders and surface cached loaded descendants during filtering
- keep scope to phase 1 only; deeper backend-backed recursive search remains follow-up work in #4673

## Testing
- `node --check static/ui.js`
- `npm run lint:runtime`

Closes #4673


nesquena-hermes (changes requested):
Thanks for this — the loaded-tree filter is a clean, well-scoped feature and the implementation itself is solid. I traced it end-to-end and the logic is sound: the DOM wiring is real (`workspaceSearchRow`/`workspaceFilterInput`/`workspaceFilterClearBtn` all exist and match the queried ids), the filter recurses only over already-cached dirs (`S._dirCache`) so it's a genuine surface-of-loaded-entries filter, the forced-expand-on-match path works, the empty-state text correctly switches between filtering and truly-empty, and it's pure client-side string matching with no XSS/injection surface. `node -c` passes on both touched JS files.

There's **one blocking issue** that will fail CI, and it's a mechanical fix:

### Locale-parity break — 2 new i18n keys are in the `en` block only

`static/i18n.js` has 13 locale blocks (`en, it, ja, ru, es, de, zh, zh-Hant, pt, ko, fr, tr, pl`). The two new keys live only in `en` (lines 425-426):

```js
workspace_filter_placeholder: 'Filter loaded files…',
workspace_filter_no_matches: 'No loaded files match this filter.',
```

The per-locale parity tests assert every `en` key exists in each locale (`sorted(en_keys - <loc>_keys) == []`). I reproduced this against your current head (`dcda696547b8`): **9 locale-parity tests fail**, e.g.

```
tests/test_chinese_locale.py::test_chinese_locale_covers_english_keys
tests/test_japanese_locale.py::test_japanese_locale_covers_english_keys
tests/test_korean_locale.py::test_korean_locale_matches_english_key_coverage
tests/test_polish_locale.py::test_polish_locale_matches_english_key_coverage
tests/test_russian_locale.py::test_russian_locale_covers_english_keys
tests/test_spanish_locale.py::test_spanish_locale_covers_english_keys
tests/test_turkish_locale.py::test_turkish_locale_matches_english_key_coverage
tests/test_zh_hant_locale.py::test_zh_hant_locale_preserves_function_and_placeholder_shapes
```
(sample: `AssertionError: Missing zh-Hant translation: workspace_filter_placeholder`)

For comparison, the sibling key `workspace_empty_dir` appears 13× (once per locale block); these two appear 1× each.

### Fix

Add **both** keys (translated) to all 12 non-`en` locale blocks: `it, ja, ru, es, de, zh, zh-Hant, pt, ko, fr, tr, pl`. Place them alongside the existing `workspace_*` cluster (next to `workspace_empty_dir` / `workspace_show_hidden_files`) so the 4-space indentation matches — the `*_locale_keys_use_standard_indentation` tests check that too. Use real unicode characters (`…`, not `\u2026`); the `no_double_escaped_unicode` tests reject escaped sequences. (`it/de/fr/pt` have no dedicated parity test file so they won't go red, but add the keys there too for consistency.)

### Optional, non-blocking
- The PR adds no test coverage. A small static test over `static/ui.js` + `index.html` asserting the `workspaceFilterInput` id exists and `setWorkspaceTreeFilter`/`clearWorkspaceTreeFilter` are defined (mirroring `tests/test_issue1793_file_tree_cruft_filter.py`) would regression-guard the wiring.
- The filter `<input>` relies on a placeholder only; consider an `aria-label` for screen readers (the JS already localizes the placeholder at runtime).

Once the keys are mirrored into the other locales the parity tests go green and this is ready for review. Nice work on the feature itself.

nesquena-hermes (dismissed):
Thanks for this @TomBanksAU — a loaded-tree filter on the Files pane is a genuinely useful phase-1 addition, and the filter logic (`_filterWorkspaceTreeEntries` walking `S._dirCache` to surface matching descendants of collapsed dirs) reads cleanly and is well-scoped.

**One blocker before this can ship: the two new i18n keys are added to the `en` locale block only, which breaks the locale-parity test suite.**

`static/i18n.js` adds these at the `en` block (lines 425–426):

```js
workspace_filter_placeholder: 'Filter loaded files…',
workspace_filter_no_matches: 'No loaded files match this filter.',
```

…but they are not present in any of the other 12 locale blocks. Every other workspace key (e.g. `workspace_empty_dir`, `workspace_show_hidden_files`) appears 13× — once per locale block — while these two appear exactly 1×. The repo's per-locale parity tests enforce that **every `en` key exists in every locale block** (e.g. `test_russian_locale_covers_english_keys` asserts `en_keys - ru_keys` is empty), so the missing keys turn the suite red.

**Reproduced** (sandboxed run on this PR's head, `pytest tests/ -k locale`): **9 failed / 164 passed**:

```
FAILED tests/test_chinese_locale.py::test_chinese_locale_covers_english_keys
FAILED tests/test_japanese_locale.py::test_japanese_locale_covers_english_keys
FAILED tests/test_korean_locale.py::test_korean_locale_matches_english_key_coverage
FAILED tests/test_polish_locale.py::test_polish_locale_matches_english_key_coverage
FAILED tests/test_russian_locale.py::test_russian_locale_covers_english_keys
FAILED tests/test_spanish_locale.py::test_spanish_locale_covers_english_keys
FAILED tests/test_turkish_locale.py::test_turkish_locale_matches_english_key_coverage
FAILED tests/test_zh_hant_locale.py::test_zh_hant_locale_covers_english_keys_without_duplicates
FAILED tests/test_zh_hant_locale.py::test_zh_hant_locale_preserves_function_and_placeholder_shapes
```

Example: `AssertionError: Missing zh-Hant translation: workspace_filter_placeholder`.

**Requested change:** add `workspace_filter_placeholder` and `workspace_filter_no_matches` to **all 12 non-`en` locale blocks** in `static/i18n.js` (ja, ru, es, de, zh, zh-Hant, pt, ko, tr, pl — and any others; there are 13 blocks total). Per this repo's convention, the **key must be present in every block**; a native translation is preferred, but an English-fallback value pending translation is acceptable (see the existing `// en fallback — pending translation` blocks in the same file). The `t()` resolver does fall back to `LOCALES.en[key]` at runtime so the UI wouldn't render `undefined`, but the parity tests are a hard CI gate and must pass.

Once the two keys are present in every locale block, `pytest tests/ -k locale` goes green and this is back on track. The rest of the change (filter wiring, CSS tab-visibility, `_renderTreeItem` plumbing) looked clean on a read-through.

nesquena-hermes:
Pulled the PR branch (`dcda69654`) and read the full diff against `origin/master`. The filter is well-scoped to phase 1 and the structure is clean — `_filterWorkspaceTreeEntries` walks `S._dirCache` recursively and attaches `_filteredChildren` so collapsed dirs surface their cached matching descendants (`static/ui.js:14411-14434`), the CSS correctly hides `.workspace-search-row` on the artifacts/todos tabs (`static/style.css:5418,5421`), and the i18n fallbacks are in place. `node --check` passing is good but won't catch the interaction issue below.

## One concrete interaction bug: forced-expand vs. click-toggle desync

During filter mode a directory with matching descendants is force-rendered open:

```js
// static/ui.js:14589-14590
const forcedExpanded=filterActive&&Array.isArray(item._filteredChildren)&&item._filteredChildren.length;
const isExpanded=forcedExpanded||S._expandedDirs.has(item.path);
```

But the click handler only consults `S._expandedDirs`, which `forcedExpanded` never writes to:

```js
// static/ui.js:14711-14714
el.onclick=async(e)=>{
  e.stopPropagation();
  if(S._expandedDirs.has(item.path)){   // false for a force-expanded dir
    S._expandedDirs.delete(item.path);
    ...
```

So for a folder that's only open *because* the filter forced it: the row shows ▾ and renders children, but the first click takes the `else` branch (`.has()` is false) and **adds** the path to `_expandedDirs` — a visual no-op, the folder stays open. The user has to click a *second* time to actually collapse it. The arrow says "expanded" but the first toggle does nothing visible. Minor, but it'll read as a dead click.

### Suggested fix

Treat a forced-expand as the expanded state in the toggle, so the first click collapses it. Roughly:

```js
const forced = filterActive && Array.isArray(item._filteredChildren) && item._filteredChildren.length;
el.onclick=async(e)=>{
  e.stopPropagation();
  if(S._expandedDirs.has(item.path) || forced){
    S._expandedDirs.delete(item.path);
    ...
```

(`forced` is computable in the same scope as `showFilteredChildren` at `static/ui.js:14753`.) Decide intent: do you want a click during filtering to collapse just that subtree, or should filtered folders be non-collapsible until the filter clears? Either is fine — the current state is the only one that's surprising.

## Two smaller notes

- **No input debounce.** `oninput="setWorkspaceTreeFilter(this.value)"` (`static/index.html:1565`) calls `renderFileTree()` on every keystroke. For a loaded tree this is cheap, but since #4673 explicitly plans backend-backed recursive search as the follow-up, it's worth threading a small debounce now so the phase-2 fetch path doesn't fire per-character later.
- **Escape inside the input is swallowed.** The `onkeydown` Escape handler (`static/index.html:1566`) calls `preventDefault()` unconditionally. If a global Escape handler exists for closing the workspace panel, Escape-to-clear-filter will mask it once the field is focused. Worth confirming that's intentional.

Nothing here blocks the phase-1 scope — the filter logic itself is correct. The forced-expand toggle is the only item I'd want addressed before merge since it produces a visibly dead click.

nesquena-hermes:
Closing in favor of the rebuilt version — thanks @TomBanksAU! The loaded-tree filter on the workspace Files tab is a nice addition; rebuilt fresh on current master as **#5911** with your work preserved (`Co-authored-by`) and the maintainer fix-spec applied (forced-expand first-click collapse, oninput debounce, scoped Escape). It'll go through the normal review pass from here. Thanks for the feature!

PR #4688, opened 2026-06-22
Fix running session ordering in sidebar

franksong2702:
## Thinking Path

- The session sidebar should make active work easy to find while preserving normal recency ordering for idle sessions.
- The cached `/api/sessions` payload was sorted before live runtime state was overlaid, so a completed row could remain between actively running sessions until the cache rebuilt.
- The browser also has a local render path, so it needs the same running-first semantics as the API response.
- Date buckets must use the same effective timestamp as the sort key; otherwise a resumed running session can be promoted to the top while still carrying a stale bucket label.

## What Changed

- Overlay live `pending_started_at`, `updated_at`, and `last_message_at` values onto cached session rows when newer.
- Sort session API rows with active/running sessions first, then by the freshest effective session timestamp.
- Apply the same running-first ordering in the frontend sidebar renderer.
- Group frontend date buckets with the same runtime-aware timestamp used for ordering.
- Add regression coverage for backend runtime-overlay ordering, the frontend comparator guard, and the date-bucket timestamp guard.

## Why It Matters

Actively streaming sessions should not be visually separated by completed conversations. Keeping running sessions grouped at the top makes it easier to switch between in-flight work and avoids stale date headers for resumed sessions.

## Verification

- `./scripts/test.sh tests/test_session_list_long_history_perf.py -q`
- `git diff --cached --check`

## Risks / Follow-ups

- This intentionally changes sidebar ordering semantics so active/running sessions win over pure recency.
- This PR does not address live-to-final/process-prose duplication; that is a separate streaming issue.

## Model Used

OpenAI Codex (GPT-5). AI assisted with implementation, test selection, and PR preparation.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -1844,10 +1844,49 @@ def _session_list_cache_overlay_runtime_rows(rows: list[dict]) -> list[dict]:
             item["has_pending_user_message"] = bool(
                 getattr(live, "pending_user_message", None)
             )
+            for key in ("pending_started_at", "updated_at", "last_message_at"):
+                current = _session_list_row_numeric_value(item.get(key))
+                raw_live_value = getattr(live, key, None)
+                live_value = _session_list_row_numeric_value(raw_live_value)
+                if live_value > current:
+                    item[key] = raw_live_value
         stream_id = item.get("active_stream_id")
         item["is_streaming"] = bool(stream_id and stream_id in active_stream_ids)
         overlaid.append(item)
+    overlaid.sort(key=_session_list_runtime_sort_key, reverse=True)
     return overlaid
+
+
+def _session_list_row_numeric_value(value) -> float:
+    try:
+        numeric = float(value or 0)
+    except (TypeError, ValueError):
+        return 0.0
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipping this in **v0.51.587** 🚀 (released via the maintainer release flow). Thanks @franksong2702 — running-first ordering + the runtime overlay are a solid fix.

One maintainer adjustment applied on top during the gate: the backend `_session_list_row_timestamp` used a flat `max(last_message_at, updated_at, pending_started_at, created_at)`, but the frontend `_sessionSortTimestampMs` uses first-truthy `last_message_at || updated_at || created_at` then `max(…, pending_started_at)`. That mismatch let a renamed/metadata-touched idle chat (bumped `updated_at`, no new messages) outrank a newer chatted session in the API response while the frontend ordered it differently. Aligned the backend to the frontend's first-truthy-then-overlay semantics so the two comparators agree. Everything else is your code as-is; the runtime overlay correctly composes with the #4680 streaming cache-freeze (applied after cache retrieval). Codex SAFE, full suite green.

PR #4695, opened 2026-06-22
Guard state.db replay after compressed context anchors

### LINKED ISSUE #4249: investigate(compress): user reports /compress + auto-compress 'sends full history back' next turn (cache miss/cost) — message-level resend NOT reproduced, needs reporter data
## Summary

Reported in Discord (#webui, **@Countlesshrs**, 2026-06-15) — unconfirmed user report, filed after investigation. The user expects `/compress` (and auto-compression) to shrink what gets sent to the model on the **next** turn, but reports:

> "if I invoke it and it compresses the context the next message immediately sends the full history back negating what it did and even costing as it misses cache, happens with auto compression too like context and message threshold"

## Investigation / what the code actually does

I could **not** reproduce a literal "full history re-sent" at the message-assembly level. Traced the next-turn context build on current master (v0.51.434):

- Manual compress (`_handle_session_compress`, `api/routes.py:15873`) writes `s.messages = compressed` **and** `s.context_messages = compressed`, and `context_messages` is the authoritative model-facing prefix.
- The next turn builds history via `reconciled_state_db_messages_for_session(s, prefer_context=True)` (`api/streaming.py:6646`), which prefers `context_messages` and runs `state_db_delta_after_context()` (`api/models.py:4436`) to drop the state.db rows already represented by the compressed prefix.
- A module-level repro (feeding the compressed summary as `context_messages` and the full pre-compression transcript as state.db) returns **only the 3 compressed rows**, not the full 6 — i.e. the route reconciliation does preserve compression.
- The bundled agent builds its working set purely from the passed `conversation_history` (`agent/conversation_loop.py` → `turn_context.py:178` `messages = list(conversation_history)`), so the cached agent does not silently re-inject old turns.

So the **message payload** appears correct. That leaves two plausible real sub-issues that match the user's perception and need confirmation:

### 1a. Prefix-cache miss after compression is real but unavoidable for the first post-compress turn
Compression **changes the prompt prefix** (summary replaces N early turns), so the first turn after any compression is necessarily an Anthropic/OpenAI prefix-cache miss. That's inherent, not a bug — but if the user is seeing a miss on **every** turn after compressing, that points at `_restore_or_build_system_prompt` (`agent/conversation_loop.py:254`) failing to persist/restore the system prompt (logged as `stored_state in ("null","empty")`), which would defeat prefix caching independently of compression. **Need: their `agent.log` around a post-compress turn** to see if the system-prompt row is `present` vs `null/empty`.

### 1b. Manual compress does not evict the cached session agent
The model-switch path explicitly calls `_evict_session_agent()` (`api/routes.py:8458`) after mutating session state; **manual compress does not**. The cached `AIAgent` is reused (`api/streaming.py:6428`). Because the turn is rebuilt from `conversation_history` this is *probably* benign, but it is an asymmetry worth auditing — a stale `context_compressor.last_prompt_tokens` / internal state on the reused agent could explain "auto-compression re-fires immediately" (see #4XXX-claim3 for the threshold half of this).

## Steps to reproduce (requested from reporter)
1. Long session on a model with a real prompt cache (Anthropic/OpenAI).
2. Run `/compress`. Observe token/cost on the **next** message.
3. Report: does the **transcript** actually show the full old history re-sent, or is it only the **token count / cost / cache-read** that looks like a full resend?

## Open questions for the reporter
- Is the symptom the **rendered context** (old messages visibly back) or the **token/cost indicator** (which is the subject of the sibling context-window bug)?
- `agent.log` lines around a post-compress turn (system-prompt `stored_state`, prompt token count, cache-read tokens).
- Model + provider.

## Reported by
@Countlesshrs (Discord #webui, 2026-06-15). Independent of, but likely entangled with, the context-window-reverts-to-256k bug filed separately.

franksong2702:
## Thinking Path

- #4249's reopened evidence shows the model-facing context, not the visible transcript, can be repopulated with stale state.db rows immediately after compression.
- The risky boundary is `reconciled_state_db_messages_for_session(..., prefer_context=True)`: once `context_messages` is compacted, content-fingerprint alignment against the uncompressed state.db can fall back to returning old transcript rows.
- Compression persists a visible anchor key. That anchor is the right boundary for state.db deltas, but only when it can be verified.
- Maintainer review caught one remaining leak: compressed `context_messages` with no anchor key still fell through to the old state.db delta path. This update fails closed for compacted context when the anchor is missing or unverifiable.

## What Changed

- Added compression-anchor boundary handling in `api/models.py`.
- Added timestamp coercion for numeric and ISO-8601 anchor timestamps, including `Z` timestamps from real sessions.
- When compacted `session.context_messages` are used with `prefer_context=True`, state.db rows are sliced only after a verified role/text/attachment/timestamp anchor match.
- If compacted context has no anchor key, no matching timestamped state row, or otherwise cannot be trusted, reconciliation returns the compacted `context_messages` only.
- Kept the existing prefix-delta behavior for ordinary non-compacted `context_messages`, so mirrored Session Arc Summary contexts can still admit fresh follow-up rows.
- Added regression coverage for verified anchors, missing anchor matches, `ts: None`, missing anchor keys, and ISO timestamp anchors.

## Why It Matters

This prevents the post-compression next turn from silently rebuilding model context with the old uncompressed transcript. For long sessions, that is the failure mode behind the apparent "compression did nothing" token/cache behavior described in #4249.

## Contract Routing

Task type: runtime/model-context bugfix
Touched areas: `api/models.py` state.db reconciliation, compression anchor handling, transcript compaction tests
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
Scope boundaries: model-facing context reconstruction only; no UI rendering or visible transcript behavior changes.
Evidence needed before claiming done: regression tests prove stale state.db rows are not replayed after compacted context when the anchor is missing or incomplete, and timestamped anchors still allow fresh post-anchor rows.

## Verification

- `./scripts/test.sh tests/test_issue1217_transcript_compaction.py -q` -> 31 passed
- `git diff --check origin/master`
- GitHub Actions are being watched after the latest force-with-lease update.

## Risks / Follow-ups

- This intentionally prefers dropping an untrusted state.db delta over polluting compacted model context with stale history.
- The fail-closed path is limited to explicit context-compaction markers; ordinary context reconciliation keeps its prior prefix-delta behavior.
- No UI screenshots: this change is model-context reconstruction only.
- No `CHANGELOG.md` entry in this contributor PR; the release process owns changelog entries.

Fixes #4249.

## Model Used

AI-assisted implementation and review.

- Coordinator: OpenAI GPT-5 Codex in Codex desktop.
- Worker session: Hermes WebUI using `@openai-codex:gpt-5.3-codex-spark` (`openai-codex`) at `http://127.0.0.1:8787/session/ad0e0f643364`.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -14,6 +14,7 @@ from contextlib import closing
 from pathlib import Path
 
 import api.config as _cfg
+from api.compression_anchor import is_context_compression_marker
 from api.config import (
     SESSION_DIR, SESSION_INDEX_FILE, SESSIONS, SESSIONS_MAX,
     LOCK, STREAMS, STREAMS_LOCK, DEFAULT_WORKSPACE, DEFAULT_MODEL, PROJECTS_FILE, HOME,
@@ -4898,6 +4899,93 @@ def state_db_delta_after_context(sidecar_context: list, state_messages: list) ->
     return state_messages[best_len:]
 
 
+def _normalized_compression_anchor_text(value) -> str:
+    return " ".join(str(value or "").split()).strip()[:160]
+
+
+def _compression_anchor_timestamp_as_float(value) -> float | None:
+    if value in (None, ""):
+        return None
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        pass
+    try:
+        return datetime.datetime.fromisoformat(str(value).strip().replace("Z", "+00:00")).timestamp()
+    except Exception:
+        return None
+
+
+def _context_messages_include_compression_marker(messages: list) -> bool:
+    for message in messages or []:
+        if not is_context_compression_marker(message):
+            continue
+        text = _message_content_text(message).lower().lstrip()
+        # Only prompt compaction summaries require fail-closed state.db replay.
+        # Other compression-adjacent summaries, such as Session Arc Summary,
+        # keep the existing prefix-delta behavior so fresh follow-ups survive.
+        if text.startswith("[context compaction") or text.startswith("context compaction"):
+            return True
+    return False
+
+
+def _state_db_anchor_index(state_messages: list, anchor_key) -> int | None:
+    if not isinstance(anchor_key, dict):
+        return None
+
+    anchor_role = str(anchor_key.get("role") or "").strip().lower()
+    anchor_text = _normalized_compression_anchor_text(anchor_key.get("text"))
+    anchor_attachments = anchor_key.get("attachments")
+    anchor_ts = _compression_anchor_timestamp_as_float(anchor_key.get("ts"))
+
+    if not anchor_role:
+        return None
+
+    # Do not attempt text-only fallback when timestamp is unavailable. Text-based
+    # fallback can match stale legacy rows if the anchor timestamp was lost,
+    # which can re-introduce old state.db rows after compaction.
+    if anchor_ts is None:
+        return None
+
+    if anchor_attachments in (None, ""):
+        expected_attachments = 0
+    else:
+        try:
+            expected_attachments = int(anchor_attachments)
+        except (TypeError, ValueError):
+            expected_attachments = 0
+
+    exact_timestamp_matches = []
+    for idx, message in enumerate(state_messages or []):
+        if _message_role(message) != anchor_role:
+            continue
+
+        attachments = message.get("attachments") if isinstance(message, dict) else None
+        attach_count = len(attachments) if isinstance(attachments, list) else 0
+        if attach_count != expected_attachments:
+            continue
+
+        # Attachment-only or timestamp-only anchors have no stable text payload.
+        # In that shape the timestamp + role + attachment count is the boundary;
+        # apply text comparison only when the anchor actually captured text.
+        message_text = _normalized_compression_anchor_text(_message_content_text(message))
+        if anchor_text and message_text != anchor_text:
+            continue
+
+        message_ts = _compression_anchor_timestamp_as_float(
+            message.get("timestamp") if isinstance(message, dict) else None
+        )
+        if message_ts is None:
+            continue
+        if abs(message_ts - anchor_ts) <= 1e-6:
+            exact_timestamp_matches.append(idx)
+            continue
+
+    if exact_timestamp_matches:
+        return exact_timestamp_matches[-1]
+    return None
+
+
 def _insert_state_message_chronologically(messages: list, msg: dict) -> bool:
     """Insert a state.db-only row before newer sidecar rows when safe.
 
@@ -5199,15 +5287,35 @@ def reconciled_state_db_messages_for_session(
     if session is None:
         return []
     local_messages = []
+    using_context_messages = False
     if prefer_context:
         context_messages = getattr(session, 'context_messages', None)
         if isinstance(context_messages, list) and context_messages:
             local_messages = context_messages
+            using_context_messages = True
     if not local_messages:
         local_messages = getattr(session, 'messages', None) or []
     if state_messages is None:
         state_messages = get_state_db_session_messages(getattr(session, 'session_id', None))
     if prefer_context and local_messages:
+        if using_context_messages:
+            compressed_context = _context_messages_include_compression_marker(local_messages)
+            anchor_key = getattr(session, "compression_anchor_message_key", None)
+            if compressed_context:
+                if not anchor_key:
+                    logger.debug(
+                        "Compressed context for session %s has no compression anchor; using context_messages only",
+                        getattr(session, "session_id", None),
+                    )
+                    return list(local_messages)
+                anchor_index = _state_db_anchor_index(state_messages, anchor_key)
+                if anchor_index is None:
+                    logger.debug(
+                        "Compressed context for session %s has an unverifiable compression anchor; using context_messages only",
+                        getattr(session, "session_id", None),
+                    )
+                    return list(local_messages)
+                state_messages = list(state_messages or [])[anchor_index + 1 :]
         state_messages = state_db_delta_after_context(local_messages, state_messages)
     return merge_session_messages_append_only(
         local_messages,
--- a/tests/test_issue1217_transcript_compaction.py
+++ b/tests/test_issue1217_transcript_compaction.py
@@ -430,6 +430,164 @@ def test_prefer_context_reconcile_keeps_state_delta_when_no_mirrored_prefix():
     assert reconciled == sidecar_context + state_messages
 
 
+def test_prefer_context_reconcile_uses_compression_anchor_boundary_for_compacted_context():
+    sidecar_context = [
+        {"role": "assistant", "content": "[CONTEXT COMPACTION — REFERENCE ONLY] compacted"},
+        {"role": "user", "content": "fresh kept question"},
+        {"role": "assistant", "content": "fresh kept answer"},
+    ]
+    state_messages = [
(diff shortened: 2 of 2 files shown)


nesquena-hermes (changes requested):
Thanks @franksong2702 — this is the right boundary to fix (#4249's stale post-compression context replay), and the timestamp-verified-anchor approach is sound. The gate reproduced **one CORE gap** that has to close before it ships, plus a follow-up under-coverage:

### CORE (blocker, reproduced): the no-anchor case does not fail closed

`reconciled_state_db_messages_for_session` guards the new anchor logic with `if anchor_key:` (models.py:5278). When a **compressed** `context_messages` session has **no** `compression_anchor_message_key` (absent / `None`), the guard is skipped entirely and execution still falls through to `state_db_delta_after_context()` — which replays the stale uncompressed transcript into model-facing context. That's the exact failure mode the PR targets, left open for the no-anchor path.

Reproduced (compressed context + `anchor_key=None`):
```python
context_messages=[
    {"role":"user","content":"[context compaction] summary of earlier work"},
    {"role":"user","content":"new question"},
]
compression_anchor_message_key=None
state=[{"role":"user","content":"old question","ts":1.0},
       {"role":"assistant","content":"old answer","ts":2.0},
       {"role":"user","content":"new question","ts":3.0}]
# result → ['[context compaction] summary…', 'new question', 'old question', 'old answer']  ← STALE LEAK
```

**Fix:** when `using_context_messages` AND the context is a *compressed* context (detect via `api.compression_anchor.is_context_compression_marker` over `local_messages`), require a truthy, timestamp-verifiable anchor — if it's missing or unverifiable, `return list(local_messages)` **before** `state_db_delta_after_context()`. Keep the existing delta behavior only for ordinary (non-compressed) `context_messages` so you don't over-suppress fresh turns on normal sessions. Please add a regression test for compressed-context + `compression_anchor_message_key=None` asserting no stale state.db rows return.

### Follow-up (not a blocker, but it defeats the fix for the target sessions)

The anchor timestamp match appears to handle only numeric epoch-second timestamps. The continuation/LCM sessions #4249 is actually about carry **ISO-8601** timestamps, so the anchor delta path silently no-ops for exactly those sessions (falls through to old behavior — no stale leak, but no fix either). Worth ISO-aware timestamp parsing + a regression test + a debug log when an anchor is present but can't be verified, so this doesn't look fixed while quietly doing nothing on the real-world case.

Everything else checks out: the visible-transcript path is unaffected (double-gated on `prefer_context` + `using_context_messages`), no perf regression, and the timestamp-match false-positive risk is acceptably bounded (role + attachment-count + normalized-text). Happy to re-review as soon as the no-anchor compressed-context case fails closed.

franksong2702:
Updated the PR to close the requested gap:

- compacted `context_messages` now fail closed when `compression_anchor_message_key` is missing or unverifiable, before the old state.db delta path can append stale rows;
- anchor timestamps now accept numeric and ISO-8601 values, including `Z` timestamps;
- ordinary non-compacted context reconciliation keeps the existing prefix-delta behavior.

Local verification: `./scripts/test.sh tests/test_issue1217_transcript_compaction.py -q` -> 31 passed, plus `git diff --check origin/master`. Latest GitHub Actions on this head are green.

nesquena-hermes:
Re-reviewed at head `548a49ff0`. The CORE blocker from my last pass is closed, the ISO follow-up is in, and CI is green across the matrix (`test (3.11/3.12/3.13)`, `lint`, `browser-smoke` all SUCCESS). This is good to ship from my side. Walking through what I verified:

### CORE gap (no-anchor compressed context) — closed ✅

The previous head guarded the anchor logic with `if anchor_key:` and let the no-anchor compressed path fall through to `state_db_delta_after_context()`. The new head gates on a `using_context_messages` flag and a dedicated compressed-context check, and both the missing-key and unverifiable-anchor paths now `return list(local_messages)` **before** the delta call (`api/models.py:5300-5319`):

```python
if using_context_messages:
    compressed_context = _context_messages_include_compression_marker(local_messages)
    anchor_key = getattr(session, "compression_anchor_message_key", None)
    if compressed_context:
        if not anchor_key:
            ... return list(local_messages)            # no anchor → fail closed
        anchor_index = _state_db_anchor_index(state_messages, anchor_key)
        if anchor_index is None:
            ... return list(local_messages)            # unverifiable → fail closed
        state_messages = list(state_messages or [])[anchor_index + 1 :]
state_messages = state_db_delta_after_context(local_messages, state_messages)
```

The reproduced leak (`compression_anchor_message_key=None` + compacted context) now returns the compacted context only. `test_prefer_context_reconcile_fails_closed_when_compressed_context_has_no_anchor_key` asserts exactly that (`reconciled == sidecar_context`).

### ISO-8601 anchor timestamps — closed ✅

`_compression_anchor_timestamp_as_float` (`api/models.py:4906-4916`) now tries `float()` first, then ISO parsing with `Z` normalization:

```python
return datetime.datetime.fromisoformat(str(value).strip().replace("Z", "+00:00")).timestamp()
```

`test_prefer_context_reconcile_accepts_iso_compression_anchor_timestamp` confirms a `2026-06-22T09:03:00Z` anchor matches and slices to the single post-anchor row. The `ts: None` fail-closed branch is also covered (`_state_db_anchor_index` returns `None` when `anchor_ts is None`, with the comment correctly noting text-only fallback would risk matching stale legacy rows — `api/models.py:4945-4949`).

### One non-blocking nit (optional polish)

`_context_messages_include_compression_marker` (`api/models.py:4919-4929`) detects the marker via `is_context_compression_marker(message)` — which extracts text through `compression_anchor._content_text` over `{"text","input_text","output_text"}` parts — but then re-derives the narrowing prefix with `_message_content_text` (`api/models.py:2255`), which only pulls `item.get('text')` from list parts. For any compacted summary that lives in provider-style `input_text`/`output_text` parts, the marker would be detected but the `startswith("[context compaction")` re-check could read empty/different text and skip the fail-closed branch. Greptile flagged the same and concluded (and I agree) it can't be triggered by content shapes current sessions actually produce, so it's not a blocker — but reusing `compression_anchor._content_text` (or threading the already-extracted text out of `is_context_compression_marker`) would remove the latent divergence.

Nothing here blocks merge. The fail-closed semantics are correct, the visible-transcript path stays untouched (double-gated on `prefer_context` + `using_context_messages`), and the regression coverage matches the failure modes from #4249. Nice work closing both items.

nesquena-hermes:
Shipped in **v0.51.589** (Release UV) 🚀 — thank you @franksong2702.

This guards the state.db replay path behind a timestamp-verified compression anchor exactly as your fix did: compressed `context_messages` with a missing or unverifiable anchor now fail closed to the compacted context, while non-compressed context keeps the existing prefix-delta behavior untouched. Closes #4249.

Gate before ship (all three legs, run on your latest head `548a49ff`):
- Full pytest suite: 10097 passed, 0 failed
- Codex advisor: SAFE TO SHIP (fail-closed on every untrusted-anchor path, normal path untouched)
- Opus advisor: CLEAN, no MUST-FIX

One non-blocking, latent-not-live follow-up was noted (unifying the compression-marker prefix re-check with the card-content extractor so the guarantee never depends on card-content shape) — it doesn't fire today since the canonical card is built as a plain string. Filing that as a small separate hardening; your code shipped as reviewed. Thanks again for the clean, well-tested fix on a delicate path.

PR #4696, opened 2026-06-22
fix(#4693): eliminate iOS tap delay on interactive buttons

### LINKED ISSUE #4693: iOS: buttons use click-only handlers, causing missed taps and accidental stop on send
## Problem

On iOS Safari, tapping the send button often doesn't fire on the first tap. A second tap sometimes sends the message but then immediately stops generation, because by the time the delayed click event fires the button has already transitioned to "stop" mode.

This happens because iOS handles `click` events differently from desktop browsers. On iOS, tapping an element that isn't already focused (or doesn't have `touch-action: manipulation`) triggers a ~300ms delay while the browser waits to distinguish a tap from a double-tap or scroll gesture. During this delay, the first tap may appear to "select" the button (triggering hover/focus) rather than activating it.

## Root cause

Every interactive button in the codebase uses either:
- HTML `onclick` attributes (e.g. `<button onclick="toggleMobileSidebar()">`)
- `.onclick` property assignment (e.g. `$('btnSend').onclick = () => { ... }`)

Neither approach registers `touchstart`/`touchend` listeners, and no buttons declare `touch-action: manipulation` in CSS (which eliminates the iOS double-tap-to-zoom delay for that element).

The send button is the most impactful case because it changes state between send/stop/interrupt/steer/queue during streaming. The timing window between tap and delayed click event is enough for `updateSendBtn()` to swap the action, so what the user intended as "send" fires as "stop."

## Affected elements

**Critical (state-switching):**
- `#btnSend` — `boot.js:447`, delegates to `handleComposerPrimaryAction()` in `ui.js:5316`

**High impact (frequent use):**
- `#btnAttach` — `boot.js:463`
- `#btnNewChat` — `boot.js:1249`
- Approval card buttons (`#approvalBtnOnce`, `#approvalBtnSession`, `#approvalBtnAlways`, `#approvalBtnDeny`) — `index.html:490-512`
- Clarify submit button (`#clarifySubmit`) — `index.html:527`
- Suggestion prompt buttons — `boot.js:1641`
- Message edit send/cancel buttons — `ui.js:12072`

**Medium impact (less frequent):**
- All `.icon-btn` elements (copy, edit, delete, retry on messages)
- All `.panel-icon-btn` elements
- Sidebar navigation buttons — `index.html:158-170`
- Download/export/import buttons — `boot.js:1274-1286`
- Mobile sidebar toggle — `index.html:121`

### CSS: add `touch-action: manipulation` to all tappable elements

This is the single most impactful change. It tells the browser "this element only needs tap and pan, not double-tap-to-zoom," which eliminates the 300ms delay on iOS.

```css
button, .icon-btn, .panel-icon-btn, .send-btn,
.approval-btn, .session-item, [onclick] {
  touch-action: manipulation;
}
```

The codebase already uses `touch-action` on scrollable containers (`.session-list`, `.messages`) and `touch-action: manipulation` on `.project-chip`, so this pattern is established.

### CSS: add `-webkit-tap-highlight-color: transparent` consistently

Currently only applied to `.pwa-standalone body`, titlebar buttons, session items, and a few other elements. Should be applied to all interactive elements to prevent the iOS tap highlight from obscuring visual feedback:

```css
button, .icon-btn, .panel-icon-btn, .send-btn {
  -webkit-tap-highlight-color: transparent;
}
```

### JS: use `pointerdown`/`pointerup` or add `touchend` for the send button

The `touch-action: manipulation` CSS fix alone should resolve most of the delay. But for the send button specifically, the race between tap and state change is narrow enough that it may still be worth binding `pointerdown` (which fires immediately on touch, before the 300ms click delay) or `touchend` with `preventDefault()` to suppress the subsequent ghost click.

The codebase already uses `pointerdown` in a few places (`boot.js:462` for `closeMobileWorkspacePanelFromChat`, `boot.js:327` for PWA sidebar swipe), so this pattern is established.

### Optional: add `:active` feedback that works on iOS

The current `:hover` pseudo-class styles (e.g. `.send-btn:hover` at `style.css:2036`) don't provide feedback on iOS because hover doesn't exist on touch. The `:active` pseudo-class does work on iOS but only when `touch-action: manipulation` is set. With the CSS fix above, the existing `.send-btn:active` rule at `style.css:2039` should start working.

## What already works

- Touch targets are correctly sized at 44x44px on mobile (`style.css:2428, 2452`)
- `touch-action: pan-y` is correctly set on scrollable containers
- Pull-to-refresh uses proper `touchstart`/`touchend` listeners (`ui.js:3620`)
- Session chips use `touch-action: manipulation` (`style.css:4643`)
- Viewport meta tag is set correctly

rodboev:
## Thinking Path

- On iOS Safari, tapping the send button often requires two taps: the first appears to "select" the button, and the second fires the action. Worse, during streaming, the second tap can land after the button has transitioned to "stop" mode, immediately cancelling the response.
- The root cause is the ~300 ms click delay iOS applies to elements without `touch-action: manipulation`, waiting to distinguish tap from double-tap-to-zoom. No button in the codebase declares this property; the only element that does is `.project-chip`.
- A single CSS rule adding `touch-action: manipulation` and `-webkit-tap-highlight-color: transparent` to all interactive button elements eliminates the delay without changing any event handling logic.

## What Changed

- `static/style.css`: added one rule covering `button`, `.icon-btn`, `.panel-icon-btn`, `.send-btn`, `.approval-btn`, and `[onclick]` with `touch-action: manipulation` and `-webkit-tap-highlight-color: transparent`

## Why It Matters

The send button is used on every message. On iOS, the missed-first-tap and send-then-stop behaviors make the app feel broken. This fix eliminates the root cause (the 300 ms click delay) for all interactive controls using a single CSS line that follows the existing `.project-chip` pattern.

## Verification

- Manual: on iOS Safari, tap send with a non-empty composer; confirm first-tap response
- Manual: during streaming, tap stop; confirm it stops without ghost-sending
- Manual: on desktop, confirm mouse click and keyboard activation are unaffected
- `npx eslint --no-config-lookup -c eslint.runtime-guard.config.mjs "static/**/*.js"` (no JS changes, but confirms no accidental edits)

## Upstream

Closes #4693.

## Model Used

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -2068,6 +2068,7 @@
   .send-btn:active{transform:scale(0.95);box-shadow:0 1px 4px var(--accent-bg);}
   .send-btn:disabled{opacity:.35;cursor:not-allowed;transform:none;box-shadow:none;}
   .send-btn.visible{animation:send-pop-in .18s cubic-bezier(.34,1.56,.64,1) forwards;}
+  button,.icon-btn,.panel-icon-btn,.send-btn,.approval-btn,[onclick]{touch-action:manipulation;-webkit-tap-highlight-color:transparent;}
   .composer-terminal-panel{position:absolute;left:0;right:0;bottom:-24px;width:min(calc(100% - 64px),720px);margin:0 auto;box-sizing:border-box;overflow:hidden;pointer-events:none;z-index:1;}
   .composer-terminal-panel.is-open,.composer-terminal-panel.is-collapsed{pointer-events:auto;}
   .composer-terminal-panel[hidden]{display:none!important;}


static/style.css
```diff
@@ -2068,6 +2068,7 @@
   .send-btn:active{transform:scale(0.95);box-shadow:0 1px 4px var(--accent-bg);}
   .send-btn:disabled{opacity:.35;cursor:not-allowed;transform:none;box-shadow:none;}
   .send-btn.visible{animation:send-pop-in .18s cubic-bezier(.34,1.56,.64,1) forwards;}
+  button,.icon-btn,.panel-icon-btn,.send-btn,.approval-btn,[onclick]{touch-action:manipulation;-webkit-tap-highlight-color:transparent;}
```
rodboev, line 2071: That's a fair observation. The `[onclick]` selector is deliberately a catch-all for the ~160 inline-attributed elements in `index.html`, not the sole touch-fix mechanism. The five class/element selectors (`button`, `.icon-btn`, `.panel-icon-btn`, `.send-btn`, `.approval-btn`) cover the interactive controls that use JS property assignment. Non-button ARIA-role elements in this codebase are limited to the composer terminal resize handle (which correctly needs `touch-action: none`) and a few overlays, none of which need the tap-delay fix. Adding `[role="button"]` risks matching unintended elements and isn't needed for the current element set.

rodboev (commented):
That's a fair observation. The `[onclick]` selector is deliberately a catch-all for the ~160 inline-attributed elements in `index.html`, not the sole touch-fix mechanism. The five class/element selectors (`button`, `.icon-btn`, `.panel-icon-btn`, `.send-btn`, `.approval-btn`) cover the interactive controls that use JS property assignment. Non-button ARIA-role elements in this codebase are limited to the composer terminal resize handle (which correctly needs `touch-action: none`) and a few overlays, none of which need the tap-delay fix. Adding `[role="button"]` risks matching unintended elements and isn't needed for the current element set.

PR #4023, opened 2026-06-12
[security] fix(auth): bind active profile cookie to session

Hinotoi-agent:
## Thinking Path

- PR #3999 scopes session-by-id behavior to the active profile.
- That guard is only effective if the active profile is derived from trusted state.
- The current request profile was still accepted from a bare `hermes_profile` cookie value.
- This PR keeps the active-profile UX cookie, but authenticates it against the current WebUI auth session before it can influence profile-scoped authorization.
- The result is a narrow defense-in-depth fix for profile isolation without adding frontend infrastructure or changing the no-auth local preference behavior.

## Summary

This PR hardens the active-profile trust boundary used by the profile-scoped session guards added in #3999.

- Binds `hermes_profile` to the current `hermes_session` when WebUI auth is enabled.
- Rejects unsigned or cross-session profile cookies before they can set request-local profile state.
- Keeps the legacy plain profile-name cookie behavior for no-auth deployments, where the cookie is only a per-browser UI preference.
- Adds regression coverage for forged, valid, and wrong-session profile cookie values.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Client-controlled active profile cookie can influence profile-scoped authorization | An authenticated client could forge `hermes_profile=<target-profile>` and make guarded session/file operations evaluate under another profile | High if profiles isolate users, tenants, workspaces, sessions, secrets, or API-key-backed state; Medium/hardening if profiles are only a local trusted-user convenience |

## Before this PR

- The server accepted the active profile from the `hermes_profile` request cookie.
- `get_profile_cookie()` validated the value syntactically, but did not authenticate it or bind it to the logged-in WebUI session.
- Profile visibility guards could therefore rely on an attacker-selectable active profile.
- Tests covered plain cookie parsing, but not forged authenticated-session profile selection.

## After this PR

- Auth-enabled deployments require the profile cookie value to be signed for the current `hermes_session` token.
- Unsigned profile cookies are ignored when auth is enabled.
- Profile cookies signed for another session are ignored when auth is enabled.
- `/api/profile/switch` emits the session-bound profile cookie format when auth is enabled.
- No-auth deployments keep the existing plain cookie behavior for local UI preference compatibility.

## Why this matters

Profiles are used as a security-sensitive boundary by the new session visibility checks. If the server trusts a raw browser cookie as the authorization input for that boundary, any authenticated client can choose the profile that the guard evaluates against.

[...]

PR #3384, opened 2026-06-02
feat(auth): trusted header authentication with profile binding (closes #3351)

Frankstar:
## Summary

This PR implements **trusted header authentication** for reverse-proxy deployments (Authelia, Authentik, nginx `auth_request`, etc.), resolving #3351.

When a reverse proxy handles authentication, the WebUI can trust headers sent by that proxy to identify the user and optionally bind them to a specific Hermes profile. This enables multi-user access with profile isolation without implementing a separate login system.

## New Environment Variables

| Variable | Description | Example |
|----------|-------------|---------|
| `HERMES_WEBUI_TRUSTED_AUTH_HEADER` | Header containing the authenticated username | `Remote-User` |
| `HERMES_WEBUI_TRUSTED_GROUPS_HEADER` | Header containing comma-separated group names | `Remote-Groups` |
| `HERMES_WEBUI_GROUP_PROFILE_MAP` | Maps proxy groups to Hermes profiles | `hermes_devops=devops,hermes_coworkers=coworkers` |
| `HERMES_WEBUI_TRUSTED_AUTH_PROXY_IPS` | **Required.** CIDR allowlist of trusted proxy IPs | `127.0.0.1,::1,10.0.0.0/8` |
| `HERMES_WEBUI_TRUSTED_AUTH_LOGOUT_URL` | Optional SSO logout URL (e.g. Authelia) | `https://auth.example.com/logout` |

## Security Design

- **Proxy IP allowlist is mandatory** — without `HERMES_WEBUI_TRUSTED_AUTH_PROXY_IPS`, the feature is disabled
- **Untrusted headers are stripped** — any `Remote-User`/`Remote-Groups` headers from non-allowlisted IPs are deleted before auth decisions
- **Session auto-creation** — valid trusted requests get a session cookie automatically; no separate login screen
- **Profile binding** — users are locked to their mapped profile via the `hermes_profile` cookie

## Authelia Example

```nginx
# nginx reverse proxy
proxy_set_header Remote-User $user;
proxy_set_header Remote-Groups $groups;

========
# Authelia access control
access_control:
  rules:
    - domain: ai.example.com
      policy: two_factor
      subject: ["group:hermes_devops", "group:hermes_coworkers"]

========
# Hermes WebUI environment
HERMES_WEBUI_TRUSTED_AUTH_HEADER=***
HERMES_WEBUI_TRUSTED_GROUPS_HEADER=Remote-Groups
[...]

PR #4703, opened 2026-06-22
feat(auth): multi-user support with per-user Hermes profiles

akrhin:
## Problem

Hermes WebUI only supports a single password for all users. When multiple people need to use the same WebUI instance (e.g. family members), they share the same credentials and the same Hermes profile — no isolation of conversations, tools, or settings.

### New files

**`api/users.py`** — User management layer
- CRUD operations stored in `STATE_DIR/users.json` (atomic writes, file mode 0600)
- `verify_user(username, password)` — PBKDF2-SHA256 (same key as legacy admin auth)
- `is_multi_user_enabled()` — auto-detects when users.json exists with entries
- `get_user_profile(username)` — maps a user to `~/.hermes/profiles/<name>`
- Thread-safe (file-level lock), lazy-loaded on first access

**`cli_users.py`** — CLI administration tool
- Five commands: `add`, `remove`, `passwd`, `set-profile`, `list`
- Targets the same `STATE_DIR` as the WebUI server
- Uses the same password hashing as the login endpoint

### Modified files

**`api/auth.py`**
- `_load_sessions()` now transparently upgrades legacy (float→dict) format
- `create_session(username="")` stores username in session data
- New `get_session_username(cookie)` — returns username from active session
- All session operations updated for new dict storage format

**`api/routes.py`** (GET /login + POST /api/auth/login)
- Login page HTML: hidden `<input id="username">` and `data-multi-user` attribute
- Server injects `{{MULTI_USER}}` flag by checking `is_multi_user_enabled()`
- POST handler has three distinct auth paths when multi-user is active:
  * `username=admin` → legacy admin password (no profile cookie)
  * any other username → `verify_user()` from users.json + `hermes_profile` cookie
  * single-user (no users.json) → original password-only flow
- Profile cookie is signed with session token (HMAC-SHA256) for integrity

**`static/login.js`**
- Reads `data-multi-user` attribute to conditionally show the username field
- Sends `{password, username}` JSON payload in multi-user mode

## Testing

Manual testing with two scenarios:
1. **Multi-user**: create users via CLI, log in with username+password, verify profile cookie is set and WebUI switches to the correct Hermes profile
2. **Admin fallback**: log in as `admin` with legacy password — no profile cookie, default WebUI behavior
3. **Single-user**: no `users.json` file — original password-only flow unchanged

## Migration

Existing deployments are unaffected — `users.json` is opt-in. Session persistence uses the same `.sessions.json` file with backward-compatible format upgrades on load. No data loss, no forced logout.


nesquena-hermes (changes requested):
Thanks @akrhin — per-user profiles is a genuinely useful direction, and a lot here is done right (256-bit `secrets.token_hex` session tokens, HMAC-signed cookies with TTL+pruning, `hmac.compare_digest` for password/sig comparisons, atomic `0600` users.json, `HttpOnly`/`SameSite=Lax` profile cookie, and `_PROFILE_ID_RE` blocking `../`/null-bytes at cookie consumption). But there are two **objective blockers** I reproduced that have to close before this can ship, plus a security-model gap that needs your design call.

### 🔴 CRITICAL (blocker, reproduced): fail-open — `is_auth_enabled()` is blind to multi-user mode (this is the #3384 bug class)

`api/auth.py:399-400`:
```python
def is_auth_enabled() -> bool:
    """True if password auth or passkey-only auth is configured."""
    return is_password_auth_enabled() or are_passkeys_enabled()
```
It was never updated for the new multi-user mode — there's no `is_multi_user_enabled()` term, even though `api/users.py:81` defines exactly that. And `api/auth.py:608` gates the whole app on it:
```python
def check_auth(...):
    if not is_auth_enabled():
        return True   # allow-all
```
**Exploit:** an operator follows the new setup path — `python cli_users.py add alice secret` — but does **not** also configure a legacy `HERMES_WEBUI_PASSWORD` / `password_hash` and has no passkeys. Then `is_password_auth_enabled()` is False, `are_passkeys_enabled()` is False ⇒ `is_auth_enabled()` is **False** ⇒ `check_auth()` returns `True` for **every** request ⇒ the entire WebUI (all `/api/` endpoints, all profile data) is reachable with **no authentication**. The new multi-user mode is precisely the auth mode the central gate can't see. Nothing enforces "a legacy admin password must also be set," so this is realistic, not theoretical.

**Fix:**
```python
def is_auth_enabled() -> bool:
    from api.users import is_multi_user_enabled   # local import avoids the cycle
    return is_password_auth_enabled() or are_passkeys_enabled() or is_multi_user_enabled()
```

### 🔴 OBJECTIVE REGRESSION (blocker): the PR turns the existing auth suite red

`pytest tests/test_auth_sessions.py` → **7 failed / 15 passed** on this PR's head (was green on master). Failing: `test_expired_session_pruned`, `test_prune_does_not_remove_valid_sessions`, `test_session_ttl_is_24_hours`, `test_verify_session_prunes_before_verification`, `test_full_sig_rejects_forged_prefix`, `test_legacy_truncated_sig_still_validates`, `test_session_uses_dynamic_ttl`.

Root cause: you changed `_sessions` values from `float` to `{"exp": float, "username": str}` (`create_session` ~442, `_prune_expired_sessions:455` now `data["exp"]`, `verify_session:479` now `session["exp"]`), but the in-memory path is only upgraded at *load* time (`_load_sessions`), so existing entries inserted as bare floats (e.g. `test_auth_sessions.py:39 auth._sessions["fake_token"] = time.time()-100`) raise `TypeError: 'float' object is not subscriptable`.

**Fix:** make the in-memory access tolerant of both forms AND update the tests:
```python
# _prune_expired_sessions / verify_session
exp = data["exp"] if isinstance(data, dict) else data
if now > exp: ...
```

### 🟠 SECURITY-MODEL GAP (please confirm intended threat model): per-user isolation isn't enforced

`/api/profile/switch` (`api/routes.py:11573`) validates only the *format* of the client-supplied `name` via `_validate_profile_name`, then signs a `hermes_profile` cookie for whatever profile was requested — it never checks the profile belongs to the authenticated user. So authenticated user `alice` can POST `/api/profile/switch {"name":"bob"}`, get a validly-signed `bob` cookie, and read/write `bob`'s sessions/credentials/memories. The #4023 cookie-binding only stops forging a cookie *without* a session; it doesn't constrain *which* profile an authenticated user may select. Relatedly, `get_session_username()` (added at `api/auth.py:574`) is **never consumed anywhere** (0 references in routes.py), so the user↔profile binding the PR seems to intend is currently dead code. If the intent is "a non-admin user is confined to their assigned profile," that needs to be enforced (derive username from the session and reject `name != get_user_profile(username)` for non-admins). Flagging this for the maintainer's design call on the threat model rather than prescribing it.

### Also worth addressing (non-blocking)
- **No tests added** for a +487-LOC auth feature. Please add: fail-closed (users.json present, no legacy pw/passkey ⇒ protected paths 401/302), per-user isolation, no cross-user data leak, credential round-trip, and the session dict-format migration.
- **`api/auth.py:455,479`** — unguarded `["exp"]` indexing runs on every auth check; use `.get("exp", 0)` with the `isinstance` guard above (defense-in-depth).
- **Hardcoded `username=="admin"`** routes to the legacy-password branch unconditionally; if anyone runs `cli_users.py add admin …` that entry is silently unreachable. Reserve/guard "admin".
- **Per-user salt:** `_hash_password` uses the installation-global `_pbkdf2_key()` as salt, so two users with the same password get identical `password_hash` values (leaks reuse to anyone reading users.json). Consider a per-user random salt.
- **Username enumeration:** `verify_user` returns immediately on unknown user but runs full PBKDF2 for a known one — the timing delta enumerates valid usernames. Run a dummy hash on the miss path.
- **Input validation:** `add_user`/`set_user_profile`/`cli_users.py` don't validate `username`/`profile` against `_PROFILE_ID_RE` up front, so a bad name is stored and silently falls back to the default profile.

Happy to re-review as soon as the fail-open (`is_auth_enabled`) and the red suite are fixed and there's test coverage — and once you've settled the per-user isolation model. Thanks again for tackling this.

akrhin:
Addressed all review feedback in the amended commit. Summary of changes:

**🔴 CRITICAL (both resolved:)**

1. **is_auth_enabled() blind to multi-user mode** — now includes `or is_multi_user_enabled()` via lazy import. If users.json exists with at least one entry, auth is considered enabled even without `HERMES_WEBUI_PASSWORD` or passkeys.

2. **7 failing tests in test_auth_sessions.py** — root cause: `_sessions` switched from `token→float` to `token→{exp, username}`. Fix: added `_session_expiry()` helper that tolerates both legacy float and new dict format. All session access points (`_prune_expired_sessions`, `verify_session`) use it. Updated tests to iterate with correct destructuring. **28/28 green now.**

**🟠 Greptile feedback (all resolved:)**

3. **Reserved username admin** — `add_user()` now `raise ValueError(...)` for `username == "admin"`, argparse catches it in CLI.

4. **Dead code in single-user else** — removed `multi_user` references and profile-cookie block from the `else` branch (it was always false/dead there anyway).

5. **Profile cookie Secure + max-age** — now calls `_is_secure_context(handler)` and sets `max-age` matching `_resolve_session_ttl()`, same pattern as `set_auth_cookie`.

6. **Unsynchronised reads** — `_ensure_loaded()` uses double-checked locking; `is_multi_user_enabled()` reads `len(_users)` under `_USERS_LOCK`.

7. **CLI password in process table** — positional password args are now `nargs="?"` (optional). Added `getpass` fallback + `--password-file` option. `_resolve_password()` reads: file > positional > interactive prompt.

**🟠 Security-model gap (resolved:)**

8. **Per-user profile isolation** — `/api/profile/switch` now checks `get_session_username()` against `get_user_profile(username)` in multi-user mode. Non-admin users can only switch to their assigned profile (403 otherwise). Admin (no username in session) is unconstrained.

akrhin:
# Multi-User Auth — Usage Guide

## Overview

This PR adds opt-in multi-user support to Hermes WebUI. Each user gets their
own credentials and is automatically switched to their assigned Hermes profile
on login. The feature is **opt-in**: if you don't create any users, the
existing single-password auth works exactly as before.

---

## Quick start

### Prerequisites: create Hermes profiles

Each user needs a Hermes profile — a separate environment with its own
conversations, tools, skills, and settings. Profiles live under
`~/.hermes/profiles/`.

To create a profile for a family member, SSH into the server and run:

```bash
# Create a Hermes profile for a new user:
hermes profile create wife

# Verify it was created:
ls ~/.hermes/profiles/
# └── wife/
```

Profiles are isolated — conversations from one profile are not visible
in another.

### 1. Create your first user

Once the profile exists, create the login credentials:

```bash
# From the hermes-webui project directory:
python cli_users.py add alice --profile wife
# or:
python3 cli_users.py add alice --profile wife

# The tool prompts for a password interactively (hidden input).
# A new file STATE_DIR/users.json is created automatically.
```

The `--profile` argument links this user to the Hermes profile you created
earlier. On every login, the WebUI automatically switches to that profile.

A password can also be passed inline:
```bash
python cli_users.py add bob my-secret-password --profile bob
```

But interactive mode is recommended to keep credentials out of shell history
and the process table.

### 2. Open the WebUI

Navigate to your Hermes WebUI login page. You'll now see a **Username** field
in addition to the **Password** field.

### 3. Log in

- **Regular user**: enter the username and password you created with `cli_users.py`.
  The WebUI automatically switches to the Hermes profile matching the username
  (or a custom profile if you set one — see below).

- **Admin user**: enter `admin` as the username and the legacy
  `HERMES_WEBUI_PASSWORD` as the password. No profile cookie is set; the
  admin always gets the default profile.

---

## CLI reference

### `add` — Create a user

```bash
python cli_users.py add <username> [password] [--profile PROFILE] [--password-file FILE]
#  or: python3 cli_users.py add <username> [password] [--profile PROFILE] [--password-file FILE]
```

| Argument | Description |
|----------|-------------|
| `username` | Unique login name. `admin` is reserved. |
| `password`  | Plaintext password. Omit for interactive prompt. |
| `--profile` | Hermes profile to switch to on login. Defaults to the username. Must match a profile in `~/.hermes/profiles/`. |
| `--password-file` | Read password from a file (e.g. a Docker secret). |

### `remove` — Delete a user

```bash
python cli_users.py remove <username>
#  or: python3 cli_users.py remove <username>
```

### `passwd` — Change password

```bash
python cli_users.py passwd <username> [new-password] [--password-file FILE]
#  or: python3 cli_users.py passwd <username> [new-password] [--password-file FILE]
```

### `set-profile` — Assign a different Hermes profile

```bash
python cli_users.py set-profile <username> <profile>
#  or: python3 cli_users.py set-profile <username> <profile>
```

The user will be switched to this profile on their next login.

### `list` — List all users

```bash
python cli_users.py list
#  or: python3 cli_users.py list
```

Shows usernames, assigned profiles, and creation timestamps (no password
hashes or secrets).

---

## Profile isolation

Each user can be assigned a Hermes profile. First, create the profile via
the Hermes CLI:

```bash
# Create a separate Hermes profile for each user:
hermes profile create wife
hermes profile create work

# Verify:
ls ~/.hermes/profiles/
# ├── default/
# ├── wife/
# └── work/
```

Then link a user to it when creating credentials:

```bash
# alice gets the "wife" profile
python cli_users.py add alice --profile wife
#  or: python3 cli_users.py add alice --profile wife
```

Or update an existing user's profile:

```bash
python cli_users.py set-profile alice wife
#  or: python3 cli_users.py set-profile alice wife
# → alice now gets the "wife" profile on next login
```

A non-admin user **cannot** switch to another user's profile via the WebUI
(`/api/profile/switch` returns 403). Only the admin user (legacy password,
no profile cookie) can freely switch between profiles.

---

## How it works

| Layer | Component |
|-------|-----------|
| **Storage** | `STATE_DIR/users.json` — atomic writes, 0600 permissions |
| **Password hashing** | PBKDF2-SHA256, 600k iterations, unique 32-byte salt per user |
| **Session** | Auth session stores the username; `get_session_username()` reads it back |
| **Profile cookie** | Signed with the session token (HMAC-SHA256), `max-age` bounded, `Secure` |
| **Auth gate** | `is_auth_enabled()` includes `is_multi_user_enabled()` — fail-closed |

### Auth flow

```
Browser                    routes.py                     users.py / auth.py
  │                            │                               │
  ├── GET /login ──────────────┤                               │
  │◄── HTML + data-multi-user ─┤                               │
  │                            │                               │
  ├── POST /api/auth/login ────┤                               │
  │   {username, password}     ├── is_multi_user_enabled() ────┤
  │                            │                               │
  │                  ┌─── admin → verify legacy password       │
  │                  ├─── user  → verify_user(username, pw)    │
  │                  └─── none  → legacy password-only         │
  │                            │                               │
  │◄── Set-Cookie: auth + ─────┤                               │
  │    hermes_profile          │                               │
```

---

## Migration from single-user

**Nothing changes.** If `users.json` does not exist or is empty:

- The login page shows only the **Password** field (no username)
- `HERMES_WEBUI_PASSWORD` or `settings.json` password_hash is the only credential
- All existing sessions and settings are preserved

When you create the first user via `cli_users.py`:

- `users.json` is created
- The login page now shows both **Username** and **Password** fields
- Legacy password still works via `admin` username

---

## Security notes

- Passwords are hashed with PBKDF2-SHA256 (600k iterations), meeting OWASP
  recommendations for interactive auth.
- Each user gets a unique random 32-byte salt — identical passwords produce
  different hashes.
- Response time for login is constant regardless of whether the username
  exists (prevents timing-based user enumeration).
- `users.json` is written atomically with `0600` permissions.
- Profile switching is enforced server-side: a regular user cannot access
  another user's profile.

akrhin (commented):
All Greptile findings have been addressed across three follow-up commits (in reverse chronological order):

**`858cb217`** — Corrupt `users.json` now logs at `WARNING` instead of silent `DEBUG`, so an operator can't miss auth being unavailable at startup.

**`01febf88`** — `get_session_username()` guards against legacy float sessions via `isinstance(session, dict)`, matching the pattern already used by `_session_expiry()`.

**`6e69f35c`** — Two security fixes:
- `verify_user()` always runs PBKDF2 even for absent usernames (prevents timing-based enumeration).
- Each user gets a unique 32-byte random salt via `secrets.token_bytes()` — identical passwords produce distinct hashes.

**Summary:** 11/11 CI checks green, 28/28 tests passing, all prior review feedback resolved. Ready for re-review.

akrhin:
@nesquena-hermes check this PR, pls)

akrhin:
Hi @nesquena-hermes, all your review feedback from the initial CHANGES_REQUESTED has been resolved across the follow-up commits:

**🔴 Blocker 1 — `is_auth_enabled()` blind to multi-user**
Fixed: now includes `is_multi_user_enabled()` so creating users.json without a legacy password does not leave the server open.

**🔴 Blocker 2 — 7 failing tests (float → dict session format)**
Fixed: introduced `_session_expiry()` helper that tolerates both legacy float and new dict format. All 28 tests pass.

**Security-model gap — profile isolation**
Non-admin users are now restricted to their assigned profile via `/api/profile/switch` guard (403 otherwise). Deleted-user sessions are also rejected.

Other items from the review (all fixed):
- Reserved `admin` username now raises explicit ValueError in `add_user()`
- Dead code in single-user branch removed
- Profile cookie now sets `Secure` and `max-age` matching auth cookie pattern
- Unsynchronised reads use double-checked locking
- CLI passwords accept positional args, `--password-file`, or interactive getpass
- `verify_user()` always runs PBKDF2 (timing-safe)
- Per-user random salt via `secrets.token_bytes(32)`
- `get_session_username()` guards against legacy float sessions
- `_load_users()` and `_save_users()` now log at WARNING on failure

11/11 CI checks green, 28/28 tests passing. Would appreciate a re-review when you get a chance. Thanks!

nesquena-hermes:
## Closing — genuinely well-built, but multi-user login is out of scope by design (consistent with #4704)

First, credit where it's due: @akrhin, the engineering here is **strong** — 256-bit tokens, HMAC-signed cookies, per-user salt, both maintainer blockers and all the Greptile findings resolved. This is careful, security-conscious work and it shows.

The reason we're closing it is a **scope/philosophy call, not a code-quality one** — and it's the same line we drew on your earlier **#4704** (where this concept first appeared bundled with the perf fix). Stating it plainly so the decision reads as a consistent principle rather than a fresh rejection:

**Hermes is single-user / multi-*profile*, not multi-*tenant*.** Multiple *profiles* (different agent configs / SOUL / workspaces) for **one operator** are core. Multiple **distinct people**, each logging in with their own credentials and seeing only their own stuff on one shared box, is multi-tenant — a different product shape with a per-end-user identity boundary, RBAC surface, and support commitment we've deliberately kept out of the core.

Why this can't take a lighter home:
- **No A/B/C anchor.** The CLI/TUI is single-user; the desktop app's Profiles are one person's profiles, not per-user logins; it's not a Codex/Cowork pattern.
- **The operator-hardening exception doesn't cover it.** That exception is "one operator securing *their own* box" (e.g. reverse-proxy SSO for yourself) — not "serve several distinct humans."
- **An auth/identity boundary is core-or-nothing.** An extension runs *inside* the already-authenticated origin, so it can't add an identity boundary — there's no extension or plugin form of this.
- **No atomic salvage.** The security hardening exists specifically to serve the multi-user layer, so there's no standalone piece to pull.

Good code doesn't buy an out-of-scope concept a pass — but a good contributor is absolutely worth keeping engaged, and I'd love to point your obvious security skill at things that *are* in scope:
- **Operator hardening of your own instance** — e.g. reverse-proxy SSO / trusted-header auth for a single operator behind their own proxy (there's active work in this exact area right now — #5568), CSP tightening, credentialed-CORS controls.
- **Single-operator profile-switching UX** — making the multi-profile experience (which IS core) smoother.

If a future direction ever revisits a narrow "one household, shared box, a few profiles behind a login" case, this thread and your implementation are exactly what we'd start from. Closing for now, with thanks — please keep contributing.

PR #4709, opened 2026-06-22
fix: hold down /api/sessions cache invalidation during streaming (#4672)

### LINKED ISSUE #4672: Performance regressions /api/sessions
/api/sessions is intermittently extremely slow despite returning a small payload (~20KB). The same endpoint sometimes completes in milliseconds, but during active chat turns it often takes multiple seconds, up to ~15–18s. This suggests the cost is server-side rebuild/blocking work before response generation, not network transfer or response size.

Profiling on the running WebUI process showed the hot path inside session/sidebar metadata handling: _build_session_list_cache_payload() → all_sessions() / get_cli_sessions() → load_metadata_only() → _read_metadata_json_prefix() / json.loads(). During active streaming, the session-list cache appears fragile because source stamps/state change frequently, causing repeated rebuilds rather than stable cache hits.

The slow items are /api/sessions and related session metadata paths. The symptom is high WebUI Python CPU during active turns and very slow apparent token rendering (~2 tokens/sec), likely due to GIL contention or request cascade around session metadata refresh.

nesquena-hermes:
## Problem

`/api/sessions` is intermittently extremely slow during an active chat turn — multi-second, occasionally ~15–18s — despite returning a tiny (~20KB) payload, and token rendering crawls to ~2 tok/s. Reported and profiled in #4672 by @batkuip, who traced the hot path to `_build_session_list_cache_payload()` → `all_sessions()` → `load_metadata_only()`.

The three perf fixes that shipped on the 21st/22nd (#4650, #4638, #4597) made each rebuild **cheaper** but did not change the *invalidate-on-every-write* cadence — which is the part the prior triage explicitly flagged as still open. @batkuip confirmed it persists on `v0.51.577` with all three present.

## Root cause

`_session_list_cache_get()` recomputes `_session_list_cache_source_stamp()` on every request. That stamp folds in:
- `_session_list_cache_state_db_fingerprint(state.db)` = `MAX(rowid)` of the `messages` table, and
- the `state.db-wal` file stat.

During an active turn the worker writes message rows continuously, so **both signals change on essentially every poll** → `stamp != current_stamp` → the cache entry is popped → a full `all_sessions()` rebuild runs on (almost) every poll. That rebuild then takes the global session `LOCK` (to overlay in-memory sessions) — the same lock the streaming worker holds while writing. The resulting lock/GIL contention is what starves token output to ~2 tok/s and produces the multi-second / ~15s `/api/sessions` latencies. The 2.5s TTL and single-flight never get to help because the stamp invalidates faster than the TTL.

## Fix

Add a **streaming hold-down**. New helper `_session_list_cache_streaming_freeze_marker()` returns a marker keyed only on the *set* of active stream ids (via `_active_stream_ids()`), or `None` when idle. While any session is streaming, `_session_list_cache_source_stamp()` collapses the volatile `state.db`-derived components (db stat, WAL stat, gateway-metadata stat, session-index stat, content fingerprint) to that marker. `settings_file` stat and the settings write-version stay live.

Result:
- Per-token message writes no longer bust the cache → rebuilds are bounded to the existing 2.5s TTL cadence instead of one-per-poll.
- The streaming path also **stops opening a per-poll SQLite connection** to the very `state.db` the worker is hammering (it skips the fingerprint read), so the hot-path stamp is strictly cheaper than idle.
- The instant a stream starts or finishes, the active-stream set changes → the marker changes → the cache re-validates and the just-finished turn's final title/count is picked up immediately.

## Why it's safe (verified by both gates)

1. **Live streaming indicators are never stale** — `is_streaming` / `active_stream_id` / `has_pending_user_message` are overlaid *after* cache retrieval in `_session_list_cache_overlay_runtime_rows()`. A held cache keeps the dots live.
2. **Structural sidebar mutations invalidate independently of this stamp** — new/deleted/renamed/imported/cron-complete/attention changes pop the cache directly through the `publish_session_list_changed` → `_on_session_list_changed` listener.
3. **Settings/sidebar-toggle changes stay live** during streaming (settings stat + write-version kept in the stamp).
4. **Bounded staleness only** — the single thing a held cache can lag during streaming is a *streaming session's own* title/message_count, bounded by the 2.5s TTL (title renames already tolerate ≤5s).
5. **No lock-ordering inversion** — `_active_stream_ids()` uses `STREAMS_LOCK`/`ACTIVE_RUNS_LOCK`, both distinct from the global `LOCK`, and is computed before `_SESSIONS_CACHE_LOCK` is taken.
6. **Idle path is byte-identical** to before (a new regression test asserts idle still tracks WAL writes).

## Honest scope notes

- **External-CLI-only turns:** the hold-down engages on the in-process WebUI worker path (`STREAMS ∪ ACTIVE_RUNS`), which is the reported scenario. A turn driven by a *separate* CLI process writing to a shared `state.db` wouldn't appear in the active set, so it sees no benefit (and no regression). Can be extended later if that case is reported.
- **Tail latency:** the 2.5s TTL still forces a periodic rebuild that takes the global `LOCK`, so a single worst-case collision is still possible. The reliable win here is "per-poll SQLite-connect eliminated + rebuilds bounded to TTL," not "the ~15s outlier is impossible."

## Tests

Four new regression tests in `tests/test_session_sidebar_cache.py`:
- stamp freezes under simulated per-token WAL + fingerprint churn while streaming
- stamp changes on active-stream-set transitions (start / add / finish)
- settings changes still invalidate during streaming
- idle path still tracks WAL writes (no regression)

Full `tests/test_session_sidebar_cache.py` (13) + related session/cache/perf slice (231) pass locally.

## Gates

- **Codex (regression):** SAFE TO SHIP — no regression risk; independently verified post-cache overlay, listener-based invalidation, marker transition, and stream-completion sidebar refresh.
- **Opus (architecture/correctness):** "Ship it." — confirmed correct, no unbounded staleness traps, no lock inversion, streaming path strictly cheaper than idle.

Thanks @batkuip for the unusually precise profiling — it made the residual obvious. Fixes #4672.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`/api/sessions` no longer crawls (and token output no longer drops to ~2 tok/s) during an active chat turn.** The sidebar session-list cache keyed its freshness on a content fingerprint of `state.db`, but an in-flight turn writes message rows continuously — so the fingerprint changed on essentially every poll, popping the cache and forcing a full `all_sessions()` rebuild mid-stream. Each forced rebuild then contended for the global session lock the streaming worker holds while writing, producing multi-second (occasionally ~15s) `/api/sessions` latencies and starving token rendering. The cache now applies a streaming hold-down: while any session is actively streaming, the volatile `state.db`-derived stamp components collapse to a marker keyed only on the set of active streams, so per-token writes stop busting the cache. Rebuilds are bounded to the existing TTL cadence instead of one-per-poll, the streaming path also stops opening a per-poll SQLite connection to the contended `state.db`, the 2.5s TTL still refreshes a streaming session's own title/count, live streaming indicators are unaffected (overlaid post-cache), and structural changes (new/deleted/renamed/imported sessions) still invalidate immediately via the explicit listener. Thanks @batkuip. (#4672)
+
 ## [v0.51.580] — 2026-06-22 — Release UM (wrap markdown source previews)
 
 ### Fixed
@@ -33,7 +37,6 @@
 
 - **The transcript no longer yanks you back to the bottom while you're reading mid-stream.** Scroll re-pinning now requires a deliberate move toward the bottom (and ignores tiny scroll jitter), so reading earlier messages during an active stream stays put instead of getting hijacked back to the latest token. Thanks @rodboev. (#4584, fixes #4295)
 - **No more accidental horizontal panning of the mobile transcript.** Wide content (long code lines, URLs) could let the message area pan sideways on phones; the transcript now clips horizontal overflow and wraps long words. Thanks @rodboev. (#4583, fixes #4553)
->>>>>>> baa5869e4 (stage #4577 (dso2ng): keep failed steer from cancelling active runs)
 
 ## [v0.51.575] — 2026-06-22 — Release UH (session-list perf for long histories)
 
--- a/api/routes.py
+++ b/api/routes.py
@@ -1699,10 +1699,80 @@ def _session_list_cache_path_stamp(path: Path | None) -> tuple[int, int]:
         return (0, 0)
 
 
+def _session_list_cache_streaming_freeze_marker():
+    """Return a hold-down marker while any session is actively streaming, else None.
+
+    During an active chat turn the gateway/CLI writes message rows to state.db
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Superseded — this exact change already shipped to master in **v0.51.584** via #4690 (which absorbed PR #4680). The streaming hold-down helper and all four regression tests are live. Closing as already-merged. Fixes #4672.

Issue #4207, opened 2026-06-15
[BUG] scrollToBottom() overrides preserveScroll — completion scroll jump persists after #1709


bravoarian:
<html>
<body>
[...]

Issue #3018, opened 2026-05-27
Token usage badge flashes briefly then disappears after response completes


lizhihao-99:
## Bug Description

After a response finishes streaming, the per-turn token usage badge (e.g. "1.2k in · 800 out") briefly appears for ~1 second, then disappears and never comes back until the page is reloaded or the session is switched away and back.

## Steps to Reproduce

1. Enable "Show token usage after responses" in Settings (or use `/usage`)
2. Send any message to the agent
3. Wait for the response to complete
4. Observe the token usage badge below the assistant's reply

## Expected Behavior

The token usage badge should persist and remain visible on the last assistant message.

## Actual Behavior

The badge appears briefly (~1 second) after the `done` SSE event, then disappears.

## Root Cause Analysis

The `_turnUsage` property is a **client-side-only** attribute computed in `_finishDone()` (messages.js ~L1837) and attached to the last assistant message object in `S.messages`. However, there are multiple code paths that **replace `S.messages` entirely** with fresh server data from the API, which does not carry `_turnUsage`:

1. **Race condition with stream fade**: When `_shouldUseStreamFade()` returns true, `_finishDone()` is deferred (up to ~900ms via `_drainStreamFadeBeforeDone`). During this window, `_streamFinalized` is still `false`. If the `stream_end` SSE event arrives before the fade completes, it calls `_restoreSettledSession()` which fetches the session from `/api/session`, replaces `S.messages` with server data (no `_turnUsage`), and calls `renderMessages()` — wiping out the usage badge.

2. **External session refresh**: The `refreshActiveSessionIfExternallyUpdated()` poll (every 5s) or focus/visibility-change triggers can call `loadSession(sid, {force:true})`, which also replaces `S.messages` with API data lacking `_turnUsage`.

3. **`_restoreSettledSession` called from SSE error handlers**: If the EventSource has a transient error after `done` fires but before `_streamFinalized` is set, `_restoreSettledSession()` can overwrite the messages.

The core issue is that `_turnUsage` is ephemeral client-side state that gets lost whenever `S.messages` is replaced from the server. The server API does not return per-turn usage attached to individual messages.

## Suggested Fix

Either:

- **Option A**: In `_restoreSettledSession()` and `loadSession()`, preserve `_turnUsage` (and `_turnDuration`, `_turnTps`, `_gatewayRouting`) from the old `S.messages` onto the newly fetched messages by matching on message index or timestamp before calling `renderMessages()`.

[...]

PR #131, opened 2026-04-06
feat: make assistant display name configurable

TaraTheStar:
This PR makes the assistant's name configurable via environment variables and the WebUI settings panel. This allows for a more personalized experience.

PR #4721, opened 2026-06-22
fix: reset _oldestIdx on stream done so transcript stops jumping to first message (#4720)

### LINKED ISSUE #4720: Regression: transcript jumps to first message (top) after every completion
## Summary

After a turn completes, the transcript **jumps to the first message (top)** instead of staying where the reader is. Reported in #webui (Discord) right after the latest update: *"My screen keeps jumping to the first message after this latest update after every completion."* Distinct from #4207 (which is jump-to-**bottom**, `needinfo`, filed June 15 — predates the recent anchor-restore rework).

- **Severity:** HIGH — makes long conversations hard to read; every completion yanks the viewport to the top.
- **When:** after **every** stream completion (the `done` event), most visibly when the reader has scrolled up and/or the session is long enough to have loaded truncated.
- **Regression window:** "this latest update" → lines up with the scroll/anchor rework that shipped June 21–22 (#4578/#4295 v0.51.553, **#4613 v0.51.587**).

## Suspected root cause (to confirm with repro)

The `done` SSE handler in `static/messages.js` (~L4180) replaces the transcript with the full payload and updates the truncation flag, but **does not reset `_oldestIdx`**:

```js
S.session=d.session;
S.messages=_carryForwardEphemeralTurnFields(S.messages||[], d.session.messages||[]);
if(typeof _messagesTruncated!=='undefined') _messagesTruncated=!!d.session._messages_truncated;
// ← _oldestIdx is NOT updated here
```

The canonical full-load paths set **both** (`static/sessions.js` ~L2029-2030 and `static/ui.js` ~L7136):

```js
_messagesTruncated = !!data.session._messages_truncated;
_oldestIdx = data.session._messages_offset || 0;   // ← the done handler is missing this line
```

Why it causes a jump to the **top** specifically:

1. #4613 reworked the scroll-restore to key on an **absolute** index: `_captureMessageViewportAnchor()` stores `sessionIdx = _oldestIdx + rawIdx`, and rows are tagged `data-session-msg-idx = _messageSessionIndexForRawIdx(rawIdx)` (= `_oldestIdx + rawIdx`). The restore (`_restoreMessageViewportAnchor`) resolves the anchor via 3 tiers: `data-message-anchor-key` → `data-session-msg-idx` → `data-msg-idx`.
2. On `done`, the handler also expands the render window from the tail window (50) to **all** messages (messages.js ~L4302-4306). For a session originally loaded **truncated** (`_oldestIdx > 0`), the full payload shifts each message's `rawIdx`, so the absolute `sessionIdx` only stays stable if `_oldestIdx` is reset to the new offset (0, since the `done` payload embeds the full transcript with no offset).
3. Because `_oldestIdx` is left stale, the captured `sessionIdx` now resolves to a **different, earlier** row (and the `rawIdx` tier is shifted too). When the anchor-key tier also misses, the final fallback uses the raw captured `scrollTop`, which — against the now-much-taller full transcript — lands near the **first message**.

So the on-`done` window expansion changes the coordinate system, and the stale `_oldestIdx` breaks the only mapping (`sessionIdx`) that was supposed to survive it.

## Proposed fix

Mirror the canonical full-load path in the `done` handler — reset `_oldestIdx` from the same payload that already drives `_messagesTruncated`:

```js
if(typeof _messagesTruncated!=='undefined') _messagesTruncated=!!d.session._messages_truncated;
if(typeof _oldestIdx!=='undefined') _oldestIdx = d.session._messages_offset || 0;   // ADD
```

(One-line symmetry fix; the `done` payload from `_session_payload_with_full_messages` embeds the full transcript and carries no offset, so this resets `_oldestIdx` to 0 and keeps the absolute `sessionIdx` anchor stable across the window expansion.)

## Reproduction

1. Open a **long** conversation (long enough that initial load is truncated — `_oldestIdx > 0`).
2. Scroll up to read an earlier message.
3. Send a prompt / wait for the next turn to complete.
4. **Observed:** at completion the viewport jumps to the first message.
   **Expected:** the viewport stays anchored where the reader was.

## Related

- #4207 — jump-to-**bottom** after scrolling up mid-stream (`needinfo`; opposite direction; predates this rework)
- #4613 / #4675 — preserve reader scroll position across re-render/reload (introduced the absolute-index anchor restore)
- #4578 / #4295 — preserve mid-stream scroll anchor + honor manual unpin
- #4346 — transcript virtualization scroll stability

## Acceptance criteria

- [ ] After a turn completes on a truncated/long session while scrolled up, the viewport stays put (no jump to first message).
- [ ] Pinned-at-bottom readers still follow to bottom on completion (no regression).
- [ ] Regression test covering `_oldestIdx` reset on `done` + absolute-anchor stability across the window expansion.

nesquena-hermes:
## Problem

After a turn completes, the transcript **jumps to the first message (top)** instead of staying where the reader is — on **every** completion. Reported in #webui (Discord) right after the latest update. Fixes #4720.

This is distinct from #4207 (jump-to-**bottom**, `needinfo`, predates this rework).

## Root cause

The `done` SSE handler in `static/messages.js` replaces the transcript with the full payload and updates `_messagesTruncated`, but **does not reset `_oldestIdx`**. The canonical full-load paths set both:

- `static/sessions.js` `_ensureMessagesLoaded` — `_messagesTruncated` **and** `_oldestIdx = _messages_offset || 0`
- `static/ui.js` `loadSession` — same.

The read-position rework (#4613, v0.51.587) keys the scroll-restore anchor on an **absolute** index:
- `_captureMessageViewportAnchor()` stores `sessionIdx = _oldestIdx + rawIdx`
- rows are tagged `data-session-msg-idx = _messageSessionIndexForRawIdx(rawIdx)` (= `_oldestIdx + rawIdx`)
- restore resolves tiers: `data-message-anchor-key` → `data-session-msg-idx` → `data-msg-idx`

On `done`, the handler also expands the render window from the tail window (50) to **all** messages. For a session originally loaded **truncated** (`_oldestIdx > 0`), the full payload shifts every message's `rawIdx`; the absolute `sessionIdx` only stays stable if `_oldestIdx` is reset to the new offset (0, since the done payload embeds the full transcript with no offset). Because `_oldestIdx` is left stale, the captured anchor resolves to an earlier row, and when the anchor-key tier also misses, the fallback uses the raw captured `scrollTop` against the now-taller full transcript → lands near the first message.

## Fix

One-line symmetry with the canonical full-load paths, in the done handler right after the `_messagesTruncated` assignment:

```js
if(typeof _oldestIdx!=='undefined')_oldestIdx=d.session._messages_offset||0;
```

The done payload (`_session_payload_with_full_messages` in `api/streaming.py`) embeds the full transcript and carries no `_messages_offset` (it isn't emitted by `compact()` — only the windowed `/api/session` HTTP path sets it), so this resets `_oldestIdx` to 0 and keeps the absolute `sessionIdx` anchor aligned with the rebuilt rows.

## Why it's safe (verified by both gates)

- **Done payload is always a full transcript** — `_session_payload_with_full_messages` builds `messages = list(session.messages)`; `compact()` never emits `_messages_offset`. So the reset always resolves to 0.
- **No-op for non-truncated sessions** — `_oldestIdx` is already 0.
- **Every other `_oldestIdx` consumer becomes *more* consistent** — load-older button (gated on `_messagesTruncated`, already set false on done), `absoluteKeepCount`, older-count UI, anchor-scene persist — all assume `_oldestIdx=0` once the full transcript is loaded, the invariant `_ensureAllMessagesLoaded` documents.
- **Pinned-at-bottom readers still follow to bottom** — `shouldFollowOnDone` → `scrollToBottom()` is orthogonal and unchanged.

## Tests

`tests/test_issue4720_done_scroll_jump_first_message.py` (3): the done handler resets `_oldestIdx` from the payload offset, the reset is typeof-guarded and ordered before the done-path re-filter/render, and node-executed offset semantics (full payload → 0; explicit offset honored defensively).

181 scroll-family + streaming-done tests pass locally; targeted 51-test gate slice passes.

## Gates

- **Codex (regression):** SAFE TO SHIP — verified full-transcript-no-offset and the absolute-anchor math.
- **Opus (architecture/correctness):** "Ship it. Correct and well-placed." — verified pinned-readers, anchor-scene-persist, and load-older paths all unaffected.

Fixes #4720.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **The transcript no longer jumps to the first message after every completion.** On a long conversation that loaded truncated, finishing a turn snapped the viewport to the top. The `done` event replaced the transcript with the full payload and expanded the render window to all messages, but left `_oldestIdx` stale — so the absolute scroll anchor (introduced with the read-position work) resolved to the wrong row and the view jumped to the start. The completion handler now resets `_oldestIdx` from the payload exactly like the other full-load paths, keeping the reader anchored where they were. Pinned-at-bottom readers still follow to the latest message. (#4720)
+
 ## [v0.51.589] — 2026-06-22 — Release UV (guard state.db replay after compressed anchors)
 
 ### Fixed
--- a/static/messages.js
+++ b/static/messages.js
@@ -4178,6 +4178,15 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
           const _prevCacheRead=(S.session&&S.session.cache_read_tokens)||0;
           const _prevCacheWrite=(S.session&&S.session.cache_write_tokens)||0;
           S.session=d.session;S.messages=_carryForwardEphemeralTurnFields(S.messages||[], d.session.messages||[]);if(typeof _messagesTruncated!=='undefined')_messagesTruncated=!!d.session._messages_truncated;
+          // #4720: the done payload (_session_payload_with_full_messages) embeds the
+          // FULL transcript, so reset _oldestIdx to its offset (0 for a full payload)
+          // exactly like the canonical full-load paths (sessions.js _ensureMessagesLoaded,
+          // ui.js loadSession). Leaving _oldestIdx stale after a truncated initial load
+          // desynchronizes the absolute scroll anchor (sessionIdx = _oldestIdx + rawIdx,
+          // #4613) from the rebuilt rows once the done handler expands the render window
+          // to all messages — the anchor resolves to an earlier row and the viewport
+          // jumps to the first message on every completion.
+          if(typeof _oldestIdx!=='undefined')_oldestIdx=d.session._messages_offset||0;
           S.messages=_filterRecoveryControlMessages(S.messages || []);
           if(typeof _hydrateTodosFromSession==='function') _hydrateTodosFromSession(S.session);
           if(typeof clearVisibleMessageRowCache==='function') clearVisibleMessageRowCache();
--- a/static/messages.js
+++ b/static/messages.js
@@ -4178,14 +4178,7 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
           const _prevCacheRead=(S.session&&S.session.cache_read_tokens)||0;
           const _prevCacheWrite=(S.session&&S.session.cache_write_tokens)||0;
           S.session=d.session;S.messages=_carryForwardEphemeralTurnFields(S.messages||[], d.session.messages||[]);if(typeof _messagesTruncated!=='undefined')_messagesTruncated=!!d.session._messages_truncated;
-          // #4720: the done payload (_session_payload_with_full_messages) embeds the
-          // FULL transcript, so reset _oldestIdx to its offset (0 for a full payload)
-          // exactly like the canonical full-load paths (sessions.js _ensureMessagesLoaded,
-          // ui.js loadSession). Leaving _oldestIdx stale after a truncated initial load
-          // desynchronizes the absolute scroll anchor (sessionIdx = _oldestIdx + rawIdx,
-          // #4613) from the rebuilt rows once the done handler expands the render window
-          // to all messages — the anchor resolves to an earlier row and the viewport
-          // jumps to the first message on every completion.
+          // #4720: reset _oldestIdx (full-load symmetry; keeps the #4613 anchor aligned).
           if(typeof _oldestIdx!=='undefined')_oldestIdx=d.session._messages_offset||0;
           S.messages=_filterRecoveryControlMessages(S.messages || []);
           if(typeof _hydrateTodosFromSession==='function') _hydrateTodosFromSession(S.session);
--- a/tests/test_issue4720_done_scroll_jump_first_message.py
+++ b/tests/test_issue4720_done_scroll_jump_first_message.py
@@ -1,0 +1,74 @@
+"""Regression for #4720: transcript jumps to the first message after completion.
+
+Root cause: the `done` SSE handler in static/messages.js replaced the transcript
+with the full payload and updated `_messagesTruncated`, but did NOT reset
+`_oldestIdx` from `d.session._messages_offset` the way the canonical full-load
+paths do (sessions.js `_ensureMessagesLoaded`, ui.js `loadSession`). The #4613
+scroll restore keys on an absolute index (`sessionIdx = _oldestIdx + rawIdx`);
+leaving `_oldestIdx` stale after a truncated initial load desynchronized that
+anchor once the done handler expands the render window to all messages, so the
+viewport jumped to the first message on every completion.
+
+These tests assert the one-line symmetry fix is present in the done handler and
+that it behaves correctly (full payload -> offset 0; explicit offset honored).
+"""
+
+import subprocess
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parents[1]
+MESSAGES_JS = (REPO / "static" / "messages.js").read_text(encoding="utf-8")
+
+
+def _compact(text: str) -> str:
+    return "".join(text.split())
+
+
+def test_done_handler_resets_oldest_idx_from_payload_offset():
+    """The done handler must reset _oldestIdx alongside _messagesTruncated."""
+    compact = _compact(MESSAGES_JS)
+    # The truncated flag and the offset reset must be wired off the SAME done
+    # payload (d.session), mirroring sessions.js / ui.js full-load paths.
+    assert "_messagesTruncated=!!d.session._messages_truncated" in compact, (
+        "done handler should still set _messagesTruncated from the done payload"
+    )
+    assert "_oldestIdx=d.session._messages_offset||0" in compact, (
+        "#4720: done handler must reset _oldestIdx from the done payload offset "
+        "so the absolute scroll anchor stays valid after the render-window expansion"
+    )
+
+
+def test_done_handler_oldest_idx_reset_is_guarded_and_ordered_before_filter():
+    """Reset must be typeof-guarded and happen before the messages are re-filtered/rendered."""
+    compact = _compact(MESSAGES_JS)
+    assert "if(typeof_oldestIdx!=='undefined')_oldestIdx=d.session._messages_offset||0" in compact, (
+        "_oldestIdx reset should be typeof-guarded like _messagesTruncated"
+    )
+    # The reset must precede _filterRecoveryControlMessages (which precedes the
+    # done-path renderMessages), so the anchor coordinate system is correct when
+    # the transcript is rebuilt.
+    reset_idx = compact.index("_oldestIdx=d.session._messages_offset||0")
+    filter_idx = compact.index("S.messages=_filterRecoveryControlMessages")
+    assert reset_idx < filter_idx, (
+        "_oldestIdx must be reset before the done-path re-filter/render"
+    )
+
+
+def test_oldest_idx_reset_matches_full_load_offset_semantics():
+    """Execute the reset expression to confirm offset semantics (full payload -> 0)."""
+    script = """
+const assert = require('assert');
+function applyReset(doneSession) {
+  let _oldestIdx = 7;  // stale value from a truncated initial load
+  // mirror the done-handler line exactly:
+  if (typeof _oldestIdx !== 'undefined') _oldestIdx = doneSession._messages_offset || 0;
+  return _oldestIdx;
+}
+// Full transcript payload (no offset field) -> reset to 0.
+assert.strictEqual(applyReset({ messages: [1, 2, 3] }), 0);
+// Explicit zero offset -> 0.
+assert.strictEqual(applyReset({ _messages_offset: 0 }), 0);
+// Explicit non-zero offset is honored (defensive; current done payload is full).
+assert.strictEqual(applyReset({ _messages_offset: 12 }), 12);
+"""
+    subprocess.run(["node", "-e", script], check=True, capture_output=True, text=True)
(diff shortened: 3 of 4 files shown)


nesquena (approved):
## Review — end-to-end ✅ (approved after fix; the PR's CI was red — I unbroke it)

Independent review of the #4720 fix: after a turn completes, the transcript jumped to the **first message (top)** on every completion — most visibly on long sessions that loaded truncated and when the reader had scrolled up. A regression from the #4613 read-position rework (v0.51.587).

### What this ships

`static/messages.js`, `tests/test_issue4720_done_scroll_jump_first_message.py` (+74), `CHANGELOG.md` (+4). Agent-authored. Closes #4720. I pushed one fix (below).

### The core fix — traced end-to-end and verified correct

The `done` SSE handler ([messages.js:4180](static/messages.js#L4180)) replaces `S.messages` with the full payload and sets `_messagesTruncated`, but did **not** reset `_oldestIdx` — unlike the canonical full-load paths (`sessions.js _ensureMessagesLoaded`, `ui.js loadSession`). #4613 keys scroll-restore on an **absolute** index; I confirmed the formula in source: `_messageSessionIndexBase()` returns `_oldestIdx`, so `data-session-msg-idx = _oldestIdx + rawIdx` and the captured anchor's `sessionIdx = _oldestIdx + rawIdx`. On `done` the handler also expands the render window from the 50-row tail to all messages ([:4304](static/messages.js#L4304)) and re-renders ([:4314](static/messages.js#L4314)); for a truncated session (`_oldestIdx>0`) that shifts every `rawIdx`, so with `_oldestIdx` stale the rebuilt rows are tagged with a double-counted offset, the captured `sessionIdx` resolves to the wrong row, and the fallback applies the raw `scrollTop` against a now-taller transcript → top.

The one-line reset (`_oldestIdx = d.session._messages_offset||0`) fixes it: after reset `_oldestIdx=0`, full-transcript rows are tagged `data-session-msg-idx = rawIdx`, and the captured anchor (the message's absolute position = its new full-transcript `rawIdx`) resolves to the **same message**. No jump. ✓

**Why the reset always lands on 0 — verified in the payload builder:** `_session_payload_with_full_messages` ([streaming.py:59](api/streaming.py#L59)) builds `messages = list(session.messages)` (full transcript) and `session.compact() | {...}` — it never sets `_messages_offset`. Grep across `api/`: `_messages_offset` and `_messages_truncated` are set **only** in `routes.py` (the windowed `/api/session` path), never in `compact()`. So the done payload carries no offset → reset to 0, the correct base for a full transcript. `S.session=d.session` is a full replace, so no stale offset can leak in. Webui-only (client scroll state; no agent surface). ✓

### What I caught — the PR's CI was red (shard 2 failing)

The full pytest suite failed `test_session_rotate_url_sync.py::test_stream_completion_syncs_rotated_session_id_to_tab_state` — and `gh pr checks` confirms the PR's `test (3.11/3.12/3.13, 2)` shards were **failing** on it. That test is the #3018 guard that the tab-anchor `localStorage.setItem('hermes-webui-session',…)` + `_setActiveSessionUrl(...)` sync lives within an 800-char proximity window of the session assignment. The PR's **9-line inline comment** pushed that sync from offset ~690 to ~1290 — out of the window. The sync still *runs* (no behavioral regression), but the static guard tripped. The author's gates ran a scroll-family/51-test *slice*, not the full suite, so they missed it.

### What I pushed — commit 66ee1d6c

Behavior unchanged — the `_oldestIdx` reset stays in the same place (before the re-filter, `typeof`-guarded), so the PR's own #4720 ordering test still passes:
- Trimmed the 9-line inline comment to one line (the full rationale is in the #4720 commit + issue), so the completion block no longer overflows the guard window.
- The reset statement legitimately grew the block even bare (to ~788/800), so widened `test_session_rotate_url_sync`'s proximity window 800 → 1000 — still well within the single done/settled handler, preserving the guard's intent (rotated session id synced right after the session assignment).

Both `test_session_rotate_url_sync` and `test_issue4720_*` now pass.

- **No-op for non-truncated sessions** (`_oldestIdx` already 0). ✓
- **Pinned-at-bottom readers** still follow to bottom (`shouldFollowOnDone → scrollToBottom()`, orthogonal/unchanged). ✓
- **Other `_oldestIdx` consumers more consistent:** load-older is gated on `_messagesTruncated` (false on done), and `_messageSessionIndexBase`/older-count/anchor-scene-persist all assume `_oldestIdx=0` once the full transcript is loaded. ✓
- **Placement/ordering:** reset precedes the window expansion + render, so the anchor tagging/restore see the corrected base. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Long truncated session, scrolled up, completes | stays on the same message (anchor survives expansion) ✅ (the fix) |
| Non-truncated session, completes | `_oldestIdx` already 0 → no-op ✅ |
| Reader pinned at bottom, completes | follows to latest ✅ |
| Done payload | always full transcript, no offset → reset to 0 ✅ |
| Load-older after done | gated on `_messagesTruncated=false`; `_oldestIdx=0` consistent ✅ |
| Tab-anchor session-id sync on completion | present + within the (widened) guard window ✅ |
| CLI reads same session | webui-only client scroll state; agent unaffected ✅ |

### Tests

- `tests/test_issue4720_done_scroll_jump_first_message.py` — **3/3** (reset from payload offset; typeof-guarded + ordered before the re-filter; node-executed offset semantics: full payload → 0, explicit offset honored).
- `tests/test_session_rotate_url_sync.py` — **1/1** (now passing after my fix; was the red CI shard).
- Scroll / anchor / viewport / streaming-done blast-radius — **383 passed, 2 skipped**.
- **Full suite (after my fix): 9972 passed / 132 skipped / 0 failures** in 203s (the previously-failing `test_session_rotate_url_sync` now passes; deselected the pre-existing darwin CRLF flake).
- **JS gate:** `node --check static/messages.js` clean; ESLint runtime guard exit 0.
- **CI:** the PR head's shard-2 was red on `test_session_rotate_url_sync`; my pushed fix resolves it (a fresh CI run on `66ee1d6c` should go green).

- The `_messages_offset||0` form is defensively correct even if a future done payload ever carried an offset; today it's always absent. The test pins both.
- The `_messageSessionIndexForRawIdx`/`_oldestIdx` "set both on full load" pattern now has three call sites (sessions.js, ui.js, and this done handler); a small shared helper would prevent a future path from drifting again — a nicety, not needed here.

### Recommendation

✅ **Approved after fix.** Parked at approval — ready for the release agent's merge/tag pipeline (the shard-2 CI red is resolved by my commit).

The scroll fix itself is correct and minimal: I verified the absolute-anchor formula in source and confirmed in the payload builder that the done payload is always a full transcript with no offset (so the reset always lands on 0 and the anchor survives the window expansion), with pinned-readers/load-older/anchor-persist all unaffected. The PR shipped with a red CI shard — its 9-line comment had pushed the #3018 tab-anchor sync out of an existing static guard's proximity window — which I caught with the full suite (the author's gate slice didn't run it) and fixed without changing behavior. Ship once CI re-greens.

nesquena-hermes:
## Live verification (before/after, real browser + real streaming turn)

Reproduced the bug on clean `master` and confirmed the fix, driving the real UI end-to-end (CDP browser + mock provider, a 150-message session that loads truncated — `_messages_offset=120`, the precondition).

### The broken invariant, measured live

After an **identical completed turn** on each build (real `done` event delivering the full 152-message transcript):

| | clean `master` (buggy) | this branch (fixed) |
|---|---|---|
| `_oldestIdx` after `done` | **122** (stale) | **0** (reset) ✅ |
| `_messagesTruncated` | false | false |
| First message ("USER MESSAGE #000") `data-session-msg-idx` | **122** (wrong) | **0** (correct) ✅ |

On `master` the post-`done` rows are mis-tagged by the stale offset, so the absolute scroll anchor can't map to the reader's row.

### The jump itself, demonstrated via the real restore path

Driving the actual `_restoreMessageViewportAnchor()` tier-2 (`data-session-msg-idx`) path — the path real sessions hit when the content-key tier misses — for a reader who was viewing **session message #131**:

| Build | `_oldestIdx` | Restore lands on |
|---|---|---|
| `master` (buggy) | 122 | **"ASSISTANT REPLY #005"** ❌ — thrown to near the first message |
| this branch | 0 | **"ASSISTANT REPLY #131"** ✅ — stays where the reader was |

The stale `+122` offset makes the captured anchor `sessionIdx=131` resolve to message **#005** — exactly the "jumps to the first message after every completion" symptom. With the reset, the same anchor lands on the correct message.

### Screenshots

- Buggy `master` — top of transcript shows **ASSISTANT REPLY #005** (`def step_5()`): the reader on #131 was thrown near the top.
- Fixed branch — top shows **ASSISTANT REPLY #131** (`def step_131()`): position preserved.

(Screenshots captured from the CDP harness; attaching in the review thread.)

### Other checks
- Pinned-at-bottom readers still follow to the latest message after completion (auto-follow path unchanged) — verified in the same harness.
- CI green across the full matrix (3.11/3.12/3.13), browser-smoke + lint pass.
- 181 scroll-family + streaming-done tests pass locally; the 3 new `test_issue4720_*` regression tests pass.

Ready for independent review.

nesquena-hermes:
Shipped in **v0.51.591** (Release UX) 🚀 — the transcript-jump-to-first-message regression (#4720) is fixed in production.

The one-line `_oldestIdx` reset on the `done` handler landed as reviewed (rebased onto fresh master, CHANGELOG-only conflict resolved, code byte-identical to the approved head). Verified end-to-end before ship:
- nesquena independent review: APPROVED
- Live browser repro (150-msg truncated session): buggy master jumped a reader on msg #131 → "ASSISTANT REPLY #005"; the fix keeps the viewport on #131. Pinned-bottom follow unaffected.
- Codex: SAFE TO SHIP · Opus: ship it · full suite + CI 11/11 green

Prod confirmed serving the fix. Closes #4720.

PR #4723, opened 2026-06-22
fix: make is_default follow active_profile in profile list API

### LINKED ISSUE #4722: is_default in profile list API is hardcoded to root profile — named profiles can never be default
## Bug Description

The `is_default` field in the profile list API (`GET /api/profiles`) is hardcoded to `true` only for the root profile (`~/.hermes`). Named profiles (under `~/.hermes/profiles/<name>`) always report `is_default: false`, regardless of the `active_profile` setting.

This means that after running `hermes profile use <name>` to set a named profile as the sticky default, desktop and mobile clients still show the root profile as the "default" because they use `is_default` from the API response.

## Root Cause

In `api/profiles.py`, `_build_profile_rows_fast()` (line ~1447-1460):

```python
rows.append(_row(default_home, default, True))  # Always True for root
# ...
rows.append(_row(entry, entry.name, False))  # Always False for ALL named profiles
```

The `is_default` boolean is passed as a literal `True`/`False` — it never checks `~/.hermes/active_profile` to determine which profile is actually the startup default.

## Steps to Reproduce

1. Create a named profile: `hermes profile create shared`
2. Set it as default: `hermes profile use shared`
3. Verify: `cat ~/.hermes/active_profile` → `shared`
4. Call the API: `GET /api/profiles`
5. Observe: `shared` has `is_default: false`, `default` has `is_default: true`

## Expected Behavior

When `hermes profile use shared` is set, the API should return `is_default: true` for the `shared` profile, indicating to clients that it is the startup/default profile.

## Proposed Fix

Read `~/.hermes/active_profile` in `_build_profile_rows_fast()` and set `is_default` based on which profile is actually the active/startup profile:

```python
active_profile_name = _read_active_profile_file()

rows.append(_row(default_home, default, _is_root_profile(active_profile_name)))
# ...
rows.append(_row(entry, entry.name, entry.name == active_profile_name))
```

PR incoming.

ihorman:
## Summary

When a user runs `hermes profile use <name>` to set a named profile as the sticky default, the WebUI's profile list API continued to report `is_default: true` only for the root profile. Named profiles were always `is_default: false`, regardless of the `active_profile` setting.

This caused desktop and mobile clients to always show the root profile as the default, even after the user explicitly switched the default via CLI.

## Fix

Read `~/.hermes/active_profile` in `_build_profile_rows_fast()` and set `is_default` based on which profile is actually the startup profile, not just which one is the root.

- Root profile gets `is_default: true` only when it is the active profile (or when no active_profile file exists — the default behavior)
- Named profiles get `is_default: true` when they match the active_profile file

## Changes

`api/profiles.py` — `_build_profile_rows_fast()`:
- Read `_read_active_profile_file()` to determine the active profile name
- Use `_is_root_profile(active_profile_name)` for the root profile row
- Use `entry.name == active_profile_name` for named profile rows

Fixes #4722

### DIFF
--- a/api/profiles.py
+++ b/api/profiles.py
@@ -1444,11 +1444,20 @@ def _build_profile_rows_fast() -> list | None:
         }
 
     rows: list = []
+
+    # Determine which profile is the "default" (startup) profile by reading
+    # the active_profile file.  When a user runs `hermes profile use <name>`,
+    # that profile should report is_default=True so that clients (desktop app,
+    # mobile app) treat it as the startup profile.  Previously is_default was
+    # hardcoded to True only for the root profile, making it impossible for a
+    # named profile to ever be the default — even after `hermes profile use`.
+    active_profile_name = _read_active_profile_file()
+
     default_home = _get_default_hermes_home()
     if default_home.is_dir():
         # Upstream hardcodes the base home's display name to "default" even when
         # the directory is literally ".hermes" — match that exactly.
-        rows.append(_row(default_home, 'default', True))
+        rows.append(_row(default_home, 'default', _is_root_profile(active_profile_name)))
 
     profiles_root = _get_profiles_root()
     if profiles_root.is_dir():
@@ -1457,7 +1466,7 @@ def _build_profile_rows_fast() -> list | None:
                 continue
             if not _UPSTREAM_PROFILE_ID_RE.match(entry.name):
                 continue
-            rows.append(_row(entry, entry.name, False))
+            rows.append(_row(entry, entry.name, entry.name == active_profile_name))
 
     return rows
 


nesquena-hermes:
Thanks for this @ihorman — I dug into the profile-list API path and I think this change is built on a premise that doesn't hold, and it has a concrete side-effect we'd need to avoid. Sharing the trace so we're on the same page.

## The field you want already exists: `is_active`

The profile-list rows carry **two** distinct booleans, set in `_build_profile_rows_fast()` / `list_profiles_api()` (`api/profiles.py`):

- **`is_default`** — "this is the **root / startup** profile" (the `~/.hermes` base home). It is intentionally tied to the root, not the active selection.
- **`is_active`** — "this is the **currently-active** profile". `list_profiles_api()` recomputes it per request as `is_active = (p['name'] == get_active_profile_name())` (≈ line 2320), and `_row(...)` seeds `'is_active': False  # filled in by caller`.

So after `hermes profile use work`, the API already reports `is_active: true` for `work`. A desktop/mobile client that wants "which profile is the startup/active one" should read **`is_active`** — that's exactly the signal this PR is trying to synthesize, and it's already there.

## Why redefining `is_default` is a problem

`is_default` is consumed elsewhere to identify the **root** profile, not the active one. Specifically `_is_root_profile()` (`api/profiles.py:338`) memoizes the root-profile name set by reading the list API:

```python
for p in infos:
    if p.get('is_default') and p.get('name'):
        _root_profile_name_cache.add(p['name'])   # line ~365
```

With this PR, after `hermes profile use work`, `_build_profile_rows_fast()` returns `is_default: true` for `work` → `_root_profile_name_cache` would treat **`work` as the root profile**. That feeds `_profiles_match()` / default-home resolution / cross-profile session matching, so conflating "active" with "default/root" here introduces a real cross-wiring bug, not just a label change.

## Suggested direction

Drop the `is_default` change and have the client read the existing **`is_active`** field instead. If the desktop/mobile app genuinely needs a dedicated "startup profile" concept distinct from both root and active, that's a small additive field + a maintainer design call (what should it mean when no `hermes profile use` has been run?) — happy to take that to the maintainer rather than overload `is_default`.

Marking changes-requested for now. If I've misread what the client needs, let me know the exact consumer and I'll re-look.

nesquena-hermes (changes requested):
See detailed comment — is_default redefinition conflicts with the root-profile cache; the is_active field already provides what the client needs. Changes requested.

ihorman:
Thanks for the thorough explanation -- you are absolutely right. `is_active` is the correct field for the active/sticky profile, and `is_default` is by-design for root profile identification. Closing this PR. The fix belongs on the client side.

PR #4724, opened 2026-06-22
fix(ui): prevent usage footer wrapping timestamps

starship-s:
## Thinking Path

- Assistant footer metadata is supposed to stay quiet and readable below the message, even when a completed turn reports a long usage line.
- On mobile/narrow footer widths, the usage footer currently puts duration, token/cache usage, timestamp, and action buttons in one flex row with limited horizontal space.
- Long usage fragments were fixed-width flex items, so they could consume the row and force the timestamp into a min-content wrap such as one character per line.
- The narrow fix is to let usage fragments shrink with ellipsis while making timestamps and action buttons non-wrapping, non-shrinking controls.
- This keeps the existing footer visibility/hover behavior and does not touch streaming, virtualization, or message rendering logic.

## What Changed

- Updated assistant footer CSS so `.msg-usage-inline`, duration, gateway, failover, and warning fragments can shrink and ellipsize.
- Made `.msg-foot .msg-time`, `.msg-foot .msg-actions`, and footer action buttons stay on one line without flex-shrinking.
- Added a focused CSS regression test for the usage-footer shrink contract and timestamp/action nowrap contract.

## Why It Matters

- Prevents long usage metadata like high token counts plus cache percentage from crushing the timestamp into vertical characters on mobile/narrow footer widths.
- Keeps footer actions stable and readable when token/cache usage is enabled.
- Preserves the calm transcript hierarchy: metadata may truncate, but message controls should not visually break.

## Screenshots

Before: mobile Brave footer area with a long usage line squeezing the timestamp into vertical text.

Mobile footer timestamp wrapping vertically before fix

After: not captured separately. The CSS contract now makes usage metadata the shrink/ellipsis target while timestamps/actions stay non-wrapping.

## Verification

Targeted coverage:

```text
./scripts/test.sh tests/test_message_footer_layout.py -q
# Running 3 items in this shard
# 3 passed in 2.53s
```

Hygiene:

```text
git diff --check
# passed
```

## Risks / Follow-ups

- This is a CSS-only fix; no runtime/session behavior is changed.
- The regression coverage is static CSS contract coverage rather than a browser-rendered visual test.
- Visual before/after evidence can be added if review needs a rendered screenshot.

## Model Used

- OpenAI / GPT-5.5 for diagnosis, implementation, and PR preparation.
- OpenAI / GPT-5.3 Codex Spark via OpenCode for independent review.
- Anthropic / Claude Opus 4.8 via Claude Code for final independent review.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -5070,8 +5070,9 @@ main.main > #mainPlugin{display:none;}
   font-size: 11px;
   color: var(--muted);
 }
-.msg-foot .msg-actions { opacity: 1; margin-left: 0; }
-.msg-foot .msg-time { font-size: 10.5px; opacity: .75; }
+.msg-foot .msg-actions { opacity: 1; margin-left: 0; flex: 0 0 auto; white-space: nowrap; }
+.msg-foot .msg-time { font-size: 10.5px; opacity: .75; white-space: nowrap; flex: 0 0 auto; }
+.msg-foot .msg-action-btn { flex: 0 0 auto; }
 .session-jump-btn--inline {
   position: static;
   right: auto;
@@ -5142,7 +5143,11 @@ main.main > #mainPlugin{display:none;}
   font-size: 11px;
   color: var(--muted);
   opacity: .7;
-  flex: 0 0 auto;
+  min-width: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  flex: 0 1 auto;
   font-variant-numeric: tabular-nums;
 }
 .gateway-failover-inline {
--- a/tests/test_message_footer_layout.py
+++ b/tests/test_message_footer_layout.py
@@ -1,0 +1,98 @@
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+STYLE_CSS = (ROOT / "static" / "style.css").read_text(encoding="utf-8")
+
+
+def _normalize_selector(selector: str) -> str:
+    return " ".join(selector.split())
+
+
+def _selector_parts(prelude: str) -> set[str]:
+    return {_normalize_selector(part) for part in prelude.split(",") if part.strip()}
+
+
+def _css_rule_blocks() -> list[tuple[set[str], str]]:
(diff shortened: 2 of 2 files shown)


starship-s:
Thanks — pushed `97b1d435` addressing the actionable Greptile notes:

- Hardened `tests/test_message_footer_layout.py` so selector lookup normalizes comma-group whitespace instead of matching the grouped selector string verbatim.
- Removed the `assert len(blocks) == 1` path; contract checks now accept any matching rule block, so future responsive/media-query copies do not fail with a misleading single-block assertion.
- Removed the redundant `max-width: 100%` from the shrinkable usage-fragment rule.

Local verification passed:

```text
./scripts/test.sh tests/test_message_footer_layout.py -q
# Running 3 items in this shard
# 3 passed in 2.53s

git diff --check
# passed
```

nesquena-hermes:
Thank you @starship-s — this surfaced a real bug: on a phone / narrow pane, a message footer with a long model name + duration + token/cost + cache% overflows the row and pushes the action buttons (edit/copy/retry) off the right edge where they're unreachable. Confirmed on a 390px viewport.

After looking at the ellipsis approach on the real mobile page, though, we don't want to ship it as-is: making every usage span `flex: 0 1 auto` + `text-overflow: ellipsis` shrinks the stats into unreadable stubs under width pressure — e.g. the duration becomes `D…` and the token count `98…`, which isn't useful. It trades "buttons off-screen" for "stats illegible."

We shipped a **wrap-based** fix for the same bug instead (**v0.51.597**): `.msg-foot` now `flex-wrap: wrap` with `.msg-actions` / `.msg-time` pinned `flex: 0 0 auto`, so when the footer doesn't fit it wraps to a second line — every stat stays full-size and readable, and the buttons stay on-screen. On wider screens where it already fits, wrap never triggers, so there's no change there.

Since the underlying overflow you reported is now resolved in v0.51.597 (credited to you in the CHANGELOG), I'm closing this PR. If you'd like to keep iterating on footer polish, a follow-up building on the wrap behavior is very welcome — and thanks again for catching this. 🙏

PR #4727, opened 2026-06-22
fix: prevent TLS handshake from stalling the accept loop

jcarvine:
## Summary

Prevents a stalled TLS handshake from blocking the entire accept loop, which can leave the WebUI process alive but unable to serve any clients.

### Problem

When TLS is enabled, `ssl.wrap_socket(listening_socket)` performs the TLS handshake synchronously inside `accept()`. If a browser, health-check probe, or network scanner opens a TCP connection but never sends a ClientHello (or sends a malformed one), the single accept loop freezes — the process stays alive but every subsequent client hangs forever.

### Solution

Override `get_request()` to accept the raw socket, then wrap it with `do_handshake_on_connect=False`. The TLS handshake now happens in the request thread instead of the accept loop, so any slow/broken handshake times out in its own thread without affecting other clients.

### Test Plan

- [x] TLS connections continue to work normally
- [x] A TCP probe that never sends a ClientHello no longer blocks the server
- [x] Unit test included: `tests/test_tls_support.py`

### Changes

- `server.py`: override `get_request()` with non-blocking TLS wrapping
- `tests/test_tls_support.py`: test for TLS accept loop resilience

### DIFF
--- a/server.py
+++ b/server.py
@@ -8,6 +8,7 @@ import os
 import re
 import signal
 import socket
+import ssl
 import sys
 import threading
 import time
@@ -157,6 +158,7 @@ class QuietHTTPServer(ThreadingHTTPServer):
         server_address = args[0] if args else kwargs.get('server_address', None)
         if server_address and ':' in server_address[0]:
             self.address_family = socket.AF_INET6
+        self.ssl_context: object | None = None
         super().__init__(*args, **kwargs)
         self.accept_loop_requests_total = 0
         self.accept_loop_last_request_at = 0.0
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.595** (Release VB) 🚀 — thank you @jcarvine, and welcome / nice first contribution.

This is a real reliability win: a stalled TLS handshake on the listening socket could wedge the entire accept loop (server alive, serving nobody). Moving the handshake into the per-request worker thread (`do_handshake_on_connect=False`, bounded by the existing 30s request timeout) means one bad client only ties up its own worker.

Gated thoroughly since `server.py` is sensitive serving infra:
- **Codex:** SAFE TO SHIP (plain-HTTP path preserved, lazy handshake bounded, accept-loop bookkeeping intact).
- **Opus:** CLEAN, no MUST-FIX — verified the `get_request()` override against the stdlib socketserver MRO, the 30s timeout bound, and no fd/thread leak / no plain-HTTP regression.
- Full suite green; your new accept-loop-resilience test was verified to fail on master and pass with the fix.

Prod confirmed serving v0.51.595. 🙏

PR #4729, opened 2026-06-22
fix: throttle reasoning SSE events to ~10 Hz to prevent browser tab f…

wlknight:
…reeze

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Reasoning SSE events no longer flood the frontend renderer.** During the reasoning/thinking phase of models like DeepSeek, the server emitted one SSE `reasoning` event per token — tens of thousands per turn — each triggering a full-text scan in the frontend renderer, locking the JS main thread (browser tab freeze). Reasoning SSE events are now throttled to ~10 Hz via a coalescing buffer: delta text is accumulated server-side and flushed in batches, so every reasoning token reaches the browser's live Thinking view while capping SSE events to ~10 Hz. The accumulated reasoning text remains complete for persistence. Thanks @wlknight.
+
 ## [v0.51.594] — 2026-06-22 — Release VA (profile-switch loading skeletons)
 
 ### Added
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -6337,6 +6337,8 @@ def _run_agent_streaming(
             # Throttle: emit metering events at most every 100 ms so the per-message
             # TPS label feels live during fast token streams without flooding SSE.
             _metering_last_emit = [time.monotonic() - 1]  # fire immediately on first token
+            _reasoning_last_put = [0.0]
+            _reasoning_buffer = ['']
             _metering_output_deltas = [0]
             _metering_reasoning_deltas = [0]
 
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Thanks @wlknight — the bug you're fixing is real and worth fixing: during a DeepSeek-style reasoning phase the server emits one SSE `reasoning` event per token, and each one triggers a full-text scan in the frontend renderer, which locks the JS main thread (the tab freeze). Good catch. But the throttle as written drops reasoning text, and there's a merge conflict to resolve. Two changes requested:

## 1. The throttle drops live reasoning deltas (correctness)

`put('reasoning', {'text': reasoning_delta})` sends the **incremental delta**, not the cumulative text (`api/streaming.py:6401`). The frontend **appends** each delta:

```js
// static/messages.js:3848 (source.addEventListener('reasoning', ...))
reasoningText += text;
liveReasoningText += text;
```

So the time-gate

```python
if now - _reasoning_last_put[0] >= 0.1:
    _reasoning_last_put[0] = now
    put('reasoning', {'text': reasoning_delta})
```

silently **discards** every delta that arrives inside the 100 ms window — those tokens never reach the browser and are permanently missing from the live Thinking view. (`STREAM_REASONING_TEXT[stream_id]` stays complete server-side, and the final persisted message is fine, but the *live stream* the user watches loses text — so the CHANGELOG's "accumulated reasoning text remains complete" isn't accurate for what the user sees mid-turn.)

This is why the metering-throttle pattern you mirrored is safe but this isn't: metering events are **idempotent snapshots** (dropping one is harmless, the next carries the current value), whereas reasoning deltas are **incremental** (dropping one loses that text forever).

### Fix: coalesce, don't drop

Buffer the deltas you'd otherwise drop and flush the **accumulated buffer** on the next emit, plus a final flush so the last partial window isn't lost:

```python
_reasoning_last_put = [0.0]
_reasoning_buffer = ['']
...
def on_reasoning(text):
    ...
    if stream_id in STREAM_REASONING_TEXT:
        STREAM_REASONING_TEXT[stream_id] += reasoning_delta
    _reasoning_buffer[0] += reasoning_delta
    now = time.monotonic()
    if now - _reasoning_last_put[0] >= 0.1:
        _reasoning_last_put[0] = now
        put('reasoning', {'text': _reasoning_buffer[0]})   # flush the whole buffer
        _reasoning_buffer[0] = ''
```

…and flush any remaining `_reasoning_buffer[0]` once when the reasoning phase ends / before the turn's `done` (so the tail isn't dropped). That caps the *event rate* at ~10 Hz (fixing the freeze) while delivering **every** token. Worth a small regression test: feed N deltas faster than 10 Hz and assert the concatenation of all emitted `reasoning` events equals the full input.

## 2. CHANGELOG merge conflict

`CHANGELOG.md` still has unresolved conflict markers (`=======` / `>>>>>>> 41bcb397`) from a stale base — please rebase on latest master and resolve (put your entry under `[Unreleased]`).

Once those two are in, this is a clean ship and a real perf win. Marking changes-requested.

nesquena-hermes (changes requested):
See detailed comment — the throttle drops incremental reasoning deltas (frontend appends them) so live text is lost; coalesce instead. Plus a CHANGELOG merge conflict. Both fixes specced.

nesquena-hermes:
Shipped in **v0.51.596** (Release VC) 🚀 — thank you @wlknight, and nice catch on the DeepSeek tab freeze.

Your coalescing throttle is the core of this fix and it's live in prod. The path from the initial bounce:
- The first version dropped deltas inside the 100ms window — but the frontend appends them (`reasoningText += text`), so dropped tokens were lost from the live Thinking view. You re-pushed the coalescing-buffer version exactly as specced (accumulate → flush whole buffer → clear).
- The dual Codex gate then hardened the tail-flush across two more rounds: the `on_reasoning(None)` tail-flush was effectively dead (the agent never calls the reasoning callback with `None`), so I added a shared `_flush_reasoning_buffer()` at the real boundaries — `on_token` (visible output starting), `on_tool` (tool boundary), after `run_conversation()` returns, and a guaranteed-exit flush in the inner `finally` so exception/retry paths emit the tail too.

Gated: Codex SAFE TO SHIP, full suite green (10161 passed), +5 regression tests pinning the coalesce contract and every flush site. `STREAM_REASONING_TEXT` keeps mirroring the full text for persistence throughout.

Real perf + correctness win. 🙏

PR #4730, opened 2026-06-22
perf: skip gateway poll when hidden/streaming, re-init smd parser before innerHTML fallback

akrhin:
## Problem

Two independent CPU waste sources during active streaming:

1. **`startGatewayPollFallback()` fires every N seconds unconditionally** — even when the tab is hidden or a stream is in progress, it calls `renderSessionList()` which does full sidebar DOM diffs. This wastes CPU on backgrounded tabs and competes with the SSE stream that is already the authoritative source.

2. **`_flushPendingSegmentRender()` falls through to `innerHTML = renderMd(displayText)` too readily** — when `_smdParser` is null (e.g. after a prior segment end), it re-parses the entire growing message body on every flush instead of re-creating the parser. This causes O(n²) DOM churn over the lifetime of a long agent response.

## Changes

| File | Change | Impact |
|---|---|---|
| `static/sessions.js` | `startGatewayPollFallback()` skips tick when `document.hidden` or `S.busy` / `S.activeStreamId`; adds `visibilitychange` catch-up | Eliminates background-tab CPU waste; no gateway updates are dropped |
| `static/messages.js` | `_flushPendingSegmentRender()` re-inits `_smdParser` via `_smdNewParser(el, false)` before falling back to `innerHTML` | Avoids O(n²) re-parse of long responses; uses non-fade renderer to match standard rendering |

## Review notes

- Feedback addressed from https://github.com/nesquena/hermes-webui/pull/4704#issuecomment-4773475189
- `startStreamingPoll()` already guards on `S.busy || S.activeStreamId` (line 423) — no conflict
- `renderSessionList()` already guards on `document.hidden` (line 3941) — gateway poll mirrors the same pattern
- `_smdNewParser()` with `fade=false` is the same renderer the normal init path uses

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -3352,6 +3352,17 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       : _stripXmlToolCalls(assistantText.slice(segmentStart));
     if(_smdParser){
       _smdWrite(displayText);
+    } else if(window.smd){
+      // Parser was nulled out (e.g. by a prior segment end) but smd is
+      // available — recreate it on the existing element. Uses the non-fade
+      // renderer to match standard rendering, avoiding O(n²) innerHTML
+      // churn on long responses (#4704).
+      // Clear existing content before re-initializing the parser
+      // to prevent duplicate DOM nodes when _smdParser was previously
+      // ended by _smdEndParser / _resetAssistantSegment (#4704).
+      assistantBody.innerHTML = '';
+      _smdNewParser(assistantBody, false);
+      if(_smdParser) _smdWrite(displayText);
     } else if(renderMd){
       assistantBody.innerHTML=renderMd(displayText);
     } else {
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped 🚀 — thank you @akrhin.

- **v0.51.592** — the gateway-poll skip-when-hidden/streaming + smd parser re-init (the clean perf half extracted from #4704). The dual gate caught a duplicate-live-text bug in the smd re-init path (`_smdNewParser` resets the written-text tracker but doesn't clear the element, so the re-render appended on top of the fallback render) — fixed by clearing `assistantBody.innerHTML` first, which your re-push independently converged on too.
- **v0.51.593** — your `visibilitychange` listener-cleanup refinement: the catch-up listener is now removed in `stopGatewayPollFallback()` and correctly re-attaches on restart (the prior sticky-flag left it lingering on stop and blocked re-add).

Both verified Codex SAFE TO SHIP + full suite green. Closing as fully shipped. Clean perf win — appreciated.

Issue #4740, opened 2026-06-23
Flaky test: test_active_request_readonly_scope_blocks_pool_env_seed fails under full-suite ordering (pre-existing on master)


nesquena-hermes:
### Symptom
`tests/test_issue3957_profile_providers_models.py::test_active_request_readonly_scope_blocks_pool_env_seed` fails **only under the full-suite run**, not in isolation:

```
>   assert _provider_has_key("openrouter") is False
E   AssertionError: assert True is False
E    +  where True = _provider_has_key('openrouter')
tests/test_issue3957_profile_providers_models.py:339: AssertionError
```

### It is pre-existing on `master` (not introduced by any in-flight PR)
Confirmed by running the full suite twice on clean `origin/master` (commit `81f5ecc82`) with no other changes:

```
1 failed, 10156 passed, 9 skipped, 1 xfailed, 2 xpassed, 16 subtests passed
FAILED tests/test_issue3957_profile_providers_models.py::test_active_request_readonly_scope_blocks_pool_env_seed
```

- **Passes in isolation:** `pytest tests/test_issue3957_profile_providers_models.py` → 33 passed.
- **Passes with its own file's neighbors:** the whole file and related `profile_env` / `secret_scope` tests pass together.
- **Fails only under full-suite ordering** → a cross-test state leak from some earlier test in the run.

### Root cause (hypothesis)
The test sets `OPENROUTER_API_KEY=from-process-env` and `_DEFAULT_HERMES_HOME` to a tmp path, then under `profile_env_for_active_request_readonly("test")` asserts `_provider_has_key("openrouter") is False` — i.e. the readonly scope must block the process-env key from seeding the pool.

Under full-suite ordering, `_provider_has_key("openrouter")` returns `True`, meaning the process-env `OPENROUTER_API_KEY` (or a cached provider/pool entry) leaks through the readonly scope. The likely culprit is residual module-global state left by an earlier test — a populated provider/credential-pool cache, a lingering `set_request_profile`, or an un-torn-down scope — that isn't reset before this test runs. The test itself does `set_request_profile("work")` + `clear_request_profile()` in a `finally`, but it doesn't reset the provider/pool caches it implicitly depends on.

### Suggested fix direction
[...]

Issue #4718, opened 2026-06-22
Phase 2 (#4662): cut actual profile-switch latency (uneven 1-5s; suspect synchronous watcher restart)


nesquena-hermes:
## Phase 2 — cut the actual profile-switch latency (the uneven 1–5s isn't session loading)

Follow-up to #4671 (Phase 1 perceived-latency skeletons), from @rodboev's benchmarking.

### Observation
On a **brand-new profile with a single conversation**, a switch still takes an **uneven 1–5s** — sometimes <1s, sometimes ~5s. Because it's not proportional to conversation count, the cost is **not** the session-list fetch. Phase 1 only addressed *perceived* latency (skeletons); this issue is the *actual* latency.

### Prime suspect — synchronous gateway watcher restart on the request path
The backend switch endpoint (`api/routes.py` ~L11612–11636, `POST /api/profile/switch`) runs three things in series before responding:
1. `switch_profile(name, process_wide=False)`
2. `invalidate_models_cache()`
3. `restart_watcher_for_profile(name)` ← **likely dominant, variable cost**

The gateway watcher teardown + respawn has timing unrelated to how much chat data the profile has, which fits the "single-convo profile still sometimes 5s" signature. The frontend default timeout is 30s (`api()` in `static/workspace.js`), so a slow restart doesn't error — it just feels slow.

### Proposed direction
- Move `restart_watcher_for_profile()` **off the switch request's critical path** (async / debounced / fire-and-forget), or make the watcher swap lazy so the switch response returns as soon as the profile state is set.
- Measure first: capture per-step timing of the three calls above under a slow case.

### Benchmark data requested
@rodboev to drop CDP-measured **min / avg / max / median** switch times here (he has a measurement loop). That localizes whether the watcher restart dominates or there's a second contributor.

### Scope
Backend, Phase 2 (actual latency). Separate from the Phase 1 frontend skeleton work.

Cross-ref: #4671, #4662.

rodboev:
## **Browser Regression Test**

**Environment:** localhost:8787, dark theme, default profile (465 webui / 418 CLI sessions) vs test-ep-sprint31 (0 webui / 200 CLI), 1920px viewport.

### Benchmark: 10 profile switches (alternating default / test-ep-sprint31, 500ms gap)

| <br> | PR (patched) | Pre-PR (stashed) |
| --- | --- | --- |
| **Min** | 727 ms | 682 ms |
| **Avg** | 2302 ms | 2689 ms |
[...]

PR #4741, opened 2026-06-23
perf(#4662): backend hot-path caching — config reparse + sidebar redaction read-once (phase 2+3)

nesquena-hermes:
## What this is

**Phase 2+3 of #4662** (profile-switch performance), built on the now-merged Phase 1 skeletons (#4671, shipped v0.51.594). Two **behavior-preserving backend caching wins** on the `/api/sessions` path that a profile switch hits cold. No data-shape changes, no new dependencies, no threading.

This is the "make it actually faster" follow-on to the "make it feel instant" skeleton work. It's deliberately scoped small and standalone-reviewable; the watcher-restart latency work (#4718) and pre-warm are separate later PRs.

### B — `reload_config()` stops re-parsing `config.yaml` on the hot path (`api/config.py`)
`reload_config()` did a raw `yaml.safe_load(config_path.read_text())` on **every** call, including on the profile-switch hot path (the ~446ms `load_settings` stage in the investigation's `/api/sessions` diagnostics). It now routes the **parse** through the same mtime+size-keyed cache added in #4652, via a shared `_load_yaml_config_file_raw()` core (`_load_yaml_config_file` = raw + env-expand — DRY, no duplicated stat/parse logic).

**The #798 profile-isolation invariant is preserved byte-for-byte:** `reload_config` still runs its *own* env expansion under the process-env-pinned thread context (`block_process_env_fallback=False` + `_thread_ctx.env={}`). Only the disk *parse* is cached; the cache stores the **raw, un-expanded** dict, so no expanded/under-expanded/profile-scoped value can ever be what's cached. The cache key is the resolved path, so profiles A and B never collide.

Micro-benchmark (14KB / 483-line config): `reload_config()` on an unchanged file drops from **~61ms → ~1.3ms** per call (~60ms parse removed per reload; larger 24KB configs save the full ~125ms+).

### C — sidebar list reads the redaction setting once per response, not per row (`api/routes.py`)
`_session_list_payload_to_response` called `_sidebar_session_response_item(s)` per row **without** `redact_enabled`, so each row's `_redact_text` re-read `settings.json` from disk (the multi-second `response_write` stage on large lists). It now reads `api_redact_enabled` **once** and threads it to every row via the **already-existing** `redact_enabled` param (the same read-once optimization `redact_session_data()` already uses). Fail-safe: redact-on-error.

Benchmark (500-row list): settings reads collapse **500 → 1** (each was a real disk read).

## Tests
- `tests/test_issue4662_reload_config_mtime_cache.py` — unchanged file not reparsed; mtime change still busts the cache; `${VAR}` expansion still works; **empty `{}` config doesn't spin `reload_config()`** (the Opus-gate regression below).
- `tests/test_issue4662_sidebar_redaction_read_once.py` — settings read ≤1× per response; redaction still applies when enabled; passes through when disabled.
- **Full suite: 10162 passed.** The one full-suite failure (`test_issue3957…readonly_scope_blocks_pool_env_seed`) is a **pre-existing ordering flake** that fails *identically on clean master* (10156 passed, same 1 fail) and passes in isolation — filed separately as #4740, not caused by this diff.

## Gate
- **Codex: SAFE TO SHIP** — confirmed the #798 expansion-pinning and `_save_yaml_config_file` cache-eviction hold; ruff + py_compile clean on the diff.
- **Opus: SAFE TO SHIP** — caught one real defect in review (an `and loaded` guard that skipped the `_cfg_mtime` stamp for a `{}` config → would spin `reload_config()` under lock; reproduced 3 reloads vs master's 0). **Fixed** (split the guard so the no-op cache update is skipped for `{}` but the mtime is always stamped, matching master) + regression test added; Opus re-gated SAFE. This fix incidentally repairs a pre-existing empty/None-config spin on master.

## Reviewers — requesting human review + browser regression testing 🙏

This is a meaty backend perf change on the crown-jewel `/api/sessions` path, so per our process it needs independent human review before merge (do **not** auto-merge — tagged `review-requested`).

@rodboev @franksong2702 @dso2ng @santastabber — could one or two of you give this a code review **and** a hands-on browser pass? You all work these surfaces (rodboev especially — this directly continues your profile-switch perf investigation and #4718). Rough things to exercise in the browser:

- **Profile switching** across profiles with many / few / zero conversations — sidebar list + workspace tree still load correctly and faster; no stale wrong-profile content.
- **Rapid back-to-back profile switches** — no errors, correct final profile.
- **Sidebar conversation titles still render correctly**, and **redaction still works** — if you have `api_redact_enabled` on, confirm credential-looking titles are still masked in the sidebar (Change C must not have disabled redaction).
- **Edit `config.yaml` while the server runs**, then trigger a reload (profile switch / settings change) — the edit is picked up (the mtime cache must bust on a real change, not serve a stale parse).
- **A freshly created / empty profile** — switch into it; no hang or runaway CPU (guards the `{}`-config spin fix).
- General smoke: settings panel, model picker, a normal chat turn — nothing in the config/redaction path regressed.

A quick "code: ✅/notes" + "browser pass: ✅/❌ + what I saw" is perfect. Thanks!

Closes part of #4662 (phases 2+3). Cross-ref: #4671 (phase 1), #4718 (phase 2 latency, separate).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- **Faster profile switching: the backend stops re-reading config and settings on the hot path.** Two behavior-preserving caching wins on the `/api/sessions` path that a profile switch hits cold (#4662, phase 2+3). (1) `reload_config()` no longer re-parses `config.yaml` from disk when the file is unchanged — it now routes through the same mtime-keyed parse cache used elsewhere, removing a ~hundreds-of-ms YAML reparse from each switch while preserving the exact process-env-expansion semantics that keep per-client profile isolation correct. (2) The sidebar session-list response now reads the redaction setting once per response instead of once per conversation row, eliminating a per-row `settings.json` read that grew into a multi-second serialization stage on large session lists. No user-visible behavior change — redaction still applies exactly as before; switching is just quicker. Part of the phased profile-switch performance work. (#4662)
+
 ## [v0.51.599] — 2026-06-23 — Release VF (dedupe live progress reasoning echoes)
 
 ### Fixed
--- a/api/config.py
+++ b/api/config.py
@@ -559,7 +559,7 @@ _yaml_file_cache: dict[str, tuple] = {}
 _yaml_file_cache_lock = threading.Lock()
 
 
-def _load_yaml_config_file_raw(config_path: Path) -> dict:
+def _load_yaml_config_file_raw(config_path: Path, *, _copy: bool = True) -> dict:
     """Return the RAW (un-env-expanded) parsed config dict, memoized on
     (resolved path, st_mtime_ns, st_size). Shared parse core for
     _load_yaml_config_file() and reload_config(): the former runs the helper's
@@ -569,6 +569,11 @@ def _load_yaml_config_file_raw(config_path: Path) -> dict:
     (mtime, size) — a UI sync storm can't turn into a YAML-reparse storm (#4650),
     and an unchanged config.yaml isn't reparsed on the profile-switch hot path
     (#4662 Phase 2).
+
+    By default returns a deep copy so a caller can never mutate the shared cache
+    entry (greptile #4741). Internal callers that immediately pass the result
+    through _expand_env_vars() (which itself returns a fresh structure and never
+    mutates its input) pass _copy=False to skip the redundant copy on the hot path.
     """
     try:
         import yaml as _yaml
@@ -587,7 +592,9 @@ def _load_yaml_config_file_raw(config_path: Path) -> dict:
         cached = _yaml_file_cache.get(cache_key)
         if cached is not None and cached[0] == stat_key:
             raw = cached[1]
-            return raw if isinstance(raw, dict) else {}
+            if not isinstance(raw, dict):
+                return {}
+            return copy.deepcopy(raw) if _copy else raw
 
     # Cache miss / stale: parse off disk. Done outside the lock so a slow parse
     # doesn't serialize unrelated paths; a concurrent duplicate parse is harmless.
@@ -600,11 +607,14 @@ def _load_yaml_config_file_raw(config_path: Path) -> dict:
     raw = loaded if isinstance(loaded, dict) else {}
     with _yaml_file_cache_lock:
         _yaml_file_cache[cache_key] = (stat_key, raw)
-    return raw
+    return copy.deepcopy(raw) if _copy else raw
 
 
 def _load_yaml_config_file(config_path: Path) -> dict:
-    raw = _load_yaml_config_file_raw(config_path)
+    # _copy=False: _expand_env_vars returns a fresh structure and never mutates
+    # its input, so the env-expanded result is already cache-safe — no need to
+    # deep-copy the raw dict first (keeps the /api/reasoning hot path cheap).
+    raw = _load_yaml_config_file_raw(config_path, _copy=False)
     if not raw:
         return {}
     expanded = _expand_env_vars(raw)
--- a/api/config.py
+++ b/api/config.py
@@ -485,11 +485,16 @@ def reload_config() -> None:
         _cfg_path = config_path
         _cfg_mtime = 0.0
         try:
-            import yaml as _yaml
-
             if config_path.exists():
-                loaded = _yaml.safe_load(config_path.read_text(encoding="utf-8"))
-                if isinstance(loaded, dict):
+                # Route the parse through the mtime-keyed cache (#4652) so an
+                # unchanged config.yaml isn't re-parsed (~125ms+ on a large file)
+                # on every reload_config() on the hot path (profile switch /
+                # load_settings, #4662 Phase 2). We take the RAW cached dict and
+                # run the env expansion HERE, pinned to the unscoped process-env
+                # view (below) — never the helper's per-call expansion — for the
+                # #798 TLS reason documented in the pin block.
+                loaded = _load_yaml_config_file_raw(config_path)
+                if isinstance(loaded, dict) and loaded:
                     # The process-global _cfg_cache must reflect PROCESS-env
                     # expansion, never a profile-scoped block_process_env_fallback
                     # view — otherwise a reload that fires while a readonly/worker
@@ -544,7 +549,17 @@ _yaml_file_cache: dict[str, tuple] = {}
 _yaml_file_cache_lock = threading.Lock()
 
 
-def _load_yaml_config_file(config_path: Path) -> dict:
+def _load_yaml_config_file_raw(config_path: Path) -> dict:
+    """Return the RAW (un-env-expanded) parsed config dict, memoized on
+    (resolved path, st_mtime_ns, st_size). Shared parse core for
+    _load_yaml_config_file() and reload_config(): the former runs the helper's
+    own per-call env expansion on the result; the latter must run expansion
+    under its own process-env-pinned thread context (#798), so it takes the raw
+    dict and expands it itself. Either way the file is parsed at most once per
+    (mtime, size) — a UI sync storm can't turn into a YAML-reparse storm (#4650),
+    and an unchanged config.yaml isn't reparsed on the profile-switch hot path
+    (#4662 Phase 2).
+    """
     try:
         import yaml as _yaml
     except ImportError:
@@ -561,8 +576,8 @@ def _load_yaml_config_file(config_path: Path) -> dict:
     with _yaml_file_cache_lock:
         cached = _yaml_file_cache.get(cache_key)
         if cached is not None and cached[0] == stat_key:
-            expanded = _expand_env_vars(cached[1])
-            return expanded if isinstance(expanded, dict) else {}
+            raw = cached[1]
+            return raw if isinstance(raw, dict) else {}
 
     # Cache miss / stale: parse off disk. Done outside the lock so a slow parse
     # doesn't serialize unrelated paths; a concurrent duplicate parse is harmless.
@@ -575,6 +590,11 @@ def _load_yaml_config_file(config_path: Path) -> dict:
     raw = loaded if isinstance(loaded, dict) else {}
     with _yaml_file_cache_lock:
         _yaml_file_cache[cache_key] = (stat_key, raw)
+    return raw
+
+
+def _load_yaml_config_file(config_path: Path) -> dict:
+    raw = _load_yaml_config_file_raw(config_path)
     if not raw:
         return {}
     expanded = _expand_env_vars(raw)
--- a/api/config.py
+++ b/api/config.py
@@ -494,30 +494,40 @@ def reload_config() -> None:
                 # view (below) — never the helper's per-call expansion — for the
                 # #798 TLS reason documented in the pin block.
                 loaded = _load_yaml_config_file_raw(config_path)
-                if isinstance(loaded, dict) and loaded:
-                    # The process-global _cfg_cache must reflect PROCESS-env
-                    # expansion, never a profile-scoped block_process_env_fallback
-                    # view — otherwise a reload that fires while a readonly/worker
-                    # scope is active (profile alternation resolves _get_config_path
-                    # to the named profile, #798 TLS) would bake under-expanded
-                    # literal ${VAR}s into the shared cache and starve concurrent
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Drive-by note from the sweep (not a review — this is correctly tagged for independent human review, leaving it for that):

I checked the one cross-PR interaction I'd worry about here, since #4544 (cross-profile credential isolation, shipped v0.51.590) routes per-request `${VAR}` expansion through `reload_config()` / `_expand_env_vars` with the thread-local `block_process_env_fallback` scrub. **This PR preserves that mechanism intact** — Change B caches only the *raw* parse (`_load_yaml_config_file_raw`), and still runs `_expand_env_vars(loaded)` under the same `block_process_env_fallback` save/restore so the process-global `_cfg_cache` keeps reflecting process-env expansion, never a profile-scoped one. `get_config()` still expands per-read against the current thread context, so the #4544 profile-scoped credential scrub is not weakened by the raw-parse cache. 👍

One thing worth a reviewer's explicit confirmation in the browser pass: with `api_redact_enabled` on, that a profile whose `config.yaml` references `${SOME_TOKEN}` still resolves correctly after a switch (exercises the raw-cache + per-read expansion together).

Flagging for @rodboev / the tagged reviewers — the `/api/sessions` hot-path + the #4718 follow-on are your territory.

nesquena (approved):

Independent review of Phase 2+3 of #4662 (profile-switch performance) — two behavior-preserving backend caching wins on the cold `/api/sessions` path, building on the merged Phase 1 skeletons (#4671) and the #4652 mtime config cache.

### What this ships

`api/config.py` (+59/-29), `api/routes.py` (+12/-1), `tests/test_issue4662_reload_config_mtime_cache.py` (+118), `tests/test_issue4662_sidebar_redaction_read_once.py` (+67), `CHANGELOG.md` (+4). Agent-authored; Codex + Opus gated (Opus caught a real `{}`-config spin, fixed — verified below). The agent has requested human + browser review; my `event=APPROVE` is the code-review gate.

Fresh tarball: `config.yaml` is shared with the agent/CLI, but the cache is **WebUI-process-local** and keyed on `(resolved path, st_mtime_ns, st_size)`, so an agent-side write to `config.yaml` busts the WebUI cache on the next read. No on-disk format change, no agent-side change. The redaction setting + sidebar payload are webui-only. No cross-tool regression. ✓

### Change B — `reload_config()` routes the parse through the mtime cache; #798 isolation preserved

The refactor splits `_load_yaml_config_file` into a shared `_load_yaml_config_file_raw()` (parse + memoize the **raw, un-expanded** dict) and the expand wrapper. `reload_config()` now calls the raw core and runs **its own** `_expand_env_vars()` under the process-env-pinned context (`block_process_env_fallback=False`, `_thread_ctx.env={}`) — unchanged from before.

This is the security-load-bearing part (the #798 TLS profile-isolation invariant), and I verified it holds:
- The cache stores the **raw** dict; on a hit `_load_yaml_config_file_raw` returns it un-expanded. So no expanded / under-expanded / profile-scoped value is ever cached — only the disk parse is.
- `reload_config` expands the raw under its own pinned context, exactly as before; `_expand_env_vars` (from the #4652 review) is **non-mutating** and returns a **fresh** structure, so concurrent profiles expanding the same shared raw cannot corrupt it or leak across each other. The cache key is the resolved path, so profiles A/B never collide.

**Harness (7/7)** against the real functions:
```
raw core returns UN-expanded ${VAR}; expand wrapper returns the expanded value
raw cache entry stays un-expanded after an expand call
expand re-resolves env per call (no reparse); raw stays literal after an env change
caller mutation of the expanded result does NOT corrupt the raw cache
empty/None config → raw {} (no crash)
```
So the cache = raw, expansion stays per-caller/pinned, and the #4652 caller-mutation safety survives the split.

**Empty-`{}` spin fix (Opus finding) — verified:** the `_cfg_mtime` stamp ([config.py:531](api/config.py#L531)) is at the `if isinstance(loaded, dict):` level, **outside** the inner `if loaded:` — so a `{}` config skips only the no-op `_cfg_cache.update` but **always** stamps the mtime, so `get_config()`'s `current_mtime != _cfg_mtime` check doesn't re-enter `reload_config()` under `_cfg_lock` forever. Matches master's pre-#4662 behavior (`_cfg_cache.update({})` was a no-op there too, so `_cfg_cache` state is identical). Reachable on the switch hot path for a freshly-created/reset profile.

**Cache busts on write:** `_save_yaml_config_file` both rewrites the file (mtime change) **and** explicitly `_yaml_file_cache.pop(str(config_path))` under the lock — so a config edit is always picked up, never a stale parse.

### Change C — sidebar reads the redaction setting once per response (security-sensitive, byte-preserving)

`_session_list_payload_to_response` now reads `api_redact_enabled` **once** and threads it to every row via the existing `redact_enabled` param, instead of each row's `_redact_text` re-reading `settings.json`.

I verified this is **exactly behavior-preserving** and never weakens redaction:
- The PR reads `bool(load_settings().get("api_redact_enabled", True))` — the **identical expression** `_redact_text(_enabled=None)` computes per-call ([helpers.py:422-424](api/helpers.py#L422)). Same source, same default (`True`), same `if not _enabled: return text` semantics.
- Fail-safe: the PR's `except → True` (redact) matches the security-safe default; redaction is never disabled by an unreadable settings read.
- This is the established read-once pattern `redact_session_data()`/`_redact_value` already use — just extended to the sidebar-list path.
- Behavior sanity: `_enabled=True` masks a credential-looking title; `_enabled=False` passes through; a plain title is untouched.

So credential masking in sidebar titles is unchanged; only the per-row disk reads collapse (500 → 1 on a 500-row list).

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| `reload_config` on unchanged config.yaml | served from raw cache, no reparse (~61ms → ~1.3ms) ✅ |
| config.yaml edited on disk (incl. by the agent) | mtime/size key busts + explicit eviction on our own write → fresh parse ✅ |
| `${VAR}` changes between reloads | re-expanded under the pinned context per reload; raw stays literal ✅ |
| Empty `{}` config (fresh/reset profile) on switch | mtime stamped → no `reload_config` spin under lock ✅ |
| Concurrent profiles reloading | raw read-only + fresh per-profile expansion → no cross-profile leak (#798) ✅ |
| Caller read-modify-saves a config dict | gets a fresh expanded copy; raw cache uncorrupted ✅ |
| Sidebar list, `api_redact_enabled` on | every row redacted (read once) ✅ |
| Sidebar list, redaction off | pass-through (read once) ✅ |
| settings.json unreadable | fail-safe redact-on-error ✅ |

### Tests

- `tests/test_issue4662_reload_config_mtime_cache.py` — unchanged-not-reparsed, mtime-busts, `${VAR}` still expands, empty-`{}` doesn't spin `reload_config`.
- `tests/test_issue4662_sidebar_redaction_read_once.py` — settings read ≤1×/response, redaction applies when enabled, passes when disabled. **7/7 across both files.**
- **Independent config-split harness: 7/7** (raw-vs-expanded + env-liveness + mutation safety, above).
- **Full suite: 10041 passed / 132 skipped / 0 failures** in 211s (deselected the pre-existing darwin CRLF flake). The `test_issue3957…readonly_scope_blocks_pool_env_seed` ordering flake the PR flagged (#4740) did **not** surface in my run — consistent with it being order-dependent and unrelated to this diff.
- `py_compile` clean on config.py + routes.py; no JS changed (no ESLint needed).
- **CI on the PR head:** lint green; test shards finishing at review time — my local full suite substitutes.

- The PR flags a pre-existing flake (`test_issue3957…readonly_scope_blocks_pool_env_seed`, filed #4740) that fails identically on clean master and passes in isolation. It did not recur in my full-suite run, consistent with an order-dependent flake unrelated to this diff.
- The DRY split (`_load_yaml_config_file = raw + expand`) is a clean improvement — the raw core is now the single parse/stat/memoize site for both the expand wrapper and `reload_config`.
- A `{}` config still doesn't *clear* stale `_cfg_cache` keys (the update is a no-op) — but that's identical to master (which also no-op'd), not a regression introduced here.

### Recommendation

✅ **Approved clean (code review).** Parked at approval — ready for the release agent's merge/tag pipeline once the requested human browser pass lands.

Both changes are correct and genuinely behavior-preserving. Change B caches only the raw disk parse and leaves `reload_config`'s #798-pinned expansion byte-for-byte intact (proven with a 7/7 raw-vs-expanded harness), with the Opus-caught `{}`-spin properly fixed (mtime always stamped) and writes always busting the cache. Change C extends the existing read-once redaction pattern to the sidebar path, threading the identical `api_redact_enabled` value the per-row path already computed — so masking is unchanged, just 500× fewer disk reads. Webui-only, no agent surface. Ship (pending the browser regression pass the PR requested).

nesquena-hermes:
Addressed both greptile P2s in the latest commit:

1. **`api/routes.py` redundant lazy import** — removed. `load_settings` is imported at module scope; `_session_list_payload_to_response` only runs at request time (after module load completes), so the lazy import was redundant.

2. **`api/config.py` mutable cached-dict reference** — `_load_yaml_config_file_raw` now deep-copies the cached raw dict by default, so no caller can mutate the shared `_yaml_file_cache` entry. The internal `_load_yaml_config_file` passes `_copy=False` (it immediately runs `_expand_env_vars`, which returns a fresh structure and never mutates its input), keeping the `/api/reasoning` hot path's cheap no-copy fast path; `reload_config` uses the safe default copy.

Full suite green after the changes.

nesquena-hermes:
**Sweep verification (not a formal review — nesquena already APPROVED this; flagging the one adjustment needed before it can ship).**

Checked the grep-title-vs-code and the incoming comments:

- **Title matches code** ✅ — `api/config.py` (mtime-cache for `reload_config()`) + `api/routes.py` (redaction read-once) + the two new tests. No scope smuggling.
- **Independent review satisfied** ✅ — nesquena APPROVED 06-23 (clean, end-to-end).
- **Greptile P2s addressed** ✅ — the redundant lazy import removed + the `{}`-config-spin mtime fix landed in the 05:24 commit.
- **rodboev's benchmark on the sibling #4748 corroborates** ✅ — warm `/api/sessions` median 105ms→66ms (~37% faster); the wins compound on every sidebar poll / SSE refresh.

**One adjustment needed before ship: rebase.** The PR is now `CONFLICTING/DIRTY` — master drifted (parallel release agent shipped several times since the branch was cut). I'll self-rebase onto fresh master, re-verify the code diff is byte-identical to the approved head, then run the authoritative Codex + Opus + full-suite gate (this touches the `#798` profile-isolation env-expansion invariant in `api/config.py` + the redaction path, which green CI doesn't fully cover at the boundary), and ship if clean.

**Not a blocker, just noting for the roadmap:** rodboev's follow-up analysis pinpoints the dominant remaining profile-switch cost as the **cold `/api/sessions` cache miss** (~720ms full rebuild for ~465 sessions), which Phase 2+3 doesn't target — that's the Phase-N pre-warm follow-up, and **#4782** (sidebar payload trim) attacks the same rebuild. The three (#4741, #4748, #4782) form a coherent profile-perf cluster. Added all three to the priority review shortlist.

Queuing for rebase + final gate.

nesquena-hermes:
Shipped in **v0.51.606** 🚀

Gated through the full authoritative pipeline (the self-built + nesquena-APPROVED path still gets the full gate):
- **Codex** SAFE TO SHIP — a direct cross-profile probe confirmed the cached raw `${TOKEN}` does not leak a prior profile's expansion or process env (the #798 isolation invariant holds).
- **Opus** SAFE TO SHIP — confirmed isolation + the #4544 credential-scrub preservation on the redaction read-once path.
- **Two #798 hardening guards added** on Opus's non-blocking note: `test_reload_config_reexpands_env_when_mtime_unchanged` (proves the mtime cache stores the raw un-expanded YAML, so an env change with an unchanged file re-expands — no stale cross-profile bleed) and `test_reload_config_does_not_cross_serve_between_profile_paths` (proves the parse cache is path-keyed).
- **Full suite** 10223 passed. It caught one stale mock in the pre-existing #4766 test (its `lambda row: row` rejected the new `redact_enabled` kwarg) — fixed to accept the keyword-only arg, assertion unchanged.

Closed as merged-with-hardening (the gate edits changed the commit so GitHub couldn't auto-detect the merge). 2nd of 3 in the profile-perf cluster (after #4775; #4717/#4748 empty-skeleton next, pending a visual sign-off).

PR #4745, opened 2026-06-23
fix: bound large session loading

liangkeni:
# Summary

Fix large Hermes WebUI sessions failing or taking tens of seconds to open by making ordinary session message loads windowed and removing hidden full-transcript frontend fetches.

# Problem

Large WebUI sidecar sessions can contain thousands of messages. Previously, even requests with `msg_limit=100` could still spend tens of seconds building the full display coordinate space through sidecar lineage stitching and fuzzy append-only merge work before slicing the result. Frontend paths also had hidden full-load triggers such as `msg_limit=9999` and naked `/api/session?messages=1` refreshes.

# Changes

- Default `/api/session?messages=1` to a bounded recent message window.
- Cap ordinary `msg_limit` requests at 300 unless `include_heavy=1` is explicitly set.
- Add a limited WebUI sidecar fast path that uses the current sidecar plus only strictly newer `state.db` rows, reserving full lineage/fuzzy reconciliation for explicit heavy loads.
- Keep tool-result payloads bounded in paginated responses.
- Replace frontend hidden full-history fetches with bounded window or metadata-only requests.
- Change scroll-up and ensure-all loading to bounded `msg_before` pagination.
- Remove the outline panel's `msg_limit=9999` full-transcript load.
- Add regression tests for backend bounds and static frontend fetch patterns.

# Verification

Ran locally:

```text
python3 -m pytest tests/test_large_session_windowing_regression.py -q
# 6 passed, 3 warnings

node --check static/outline.js
node --check static/commands.js
node --check static/ui.js

python3 -m py_compile api/routes.py
git diff --check HEAD~1..HEAD
```

Runtime check against a 6,829-message local session:

```text
/api/session?messages=1&msg_limit=100       ~0.066s, 272 raw rows, _messages_offset=6557
/api/session?messages=1&msg_limit=9999      ~0.122s, capped, 902 raw rows
```

Also verified the WebUI page opens in a browser without `Failed to load conversation messages` and without console errors.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5837,6 +5837,8 @@ def _message_window_for_display(messages, msg_limit=None, msg_before=None, expan
     return window, start_idx
 
 
+_DEFAULT_SESSION_MSG_LIMIT = 100
+_MAX_SESSION_MSG_LIMIT = 300
 _LIMITED_TOOL_CONTENT_MAX_CHARS = 4096
 _LIMITED_TOOL_CONTENT_NOTICE = (
     "\n\n[Tool output truncated in paginated session response; "
@@ -5883,27 +5885,58 @@ def _messages_for_limited_payload(messages) -> list:
 def _limited_webui_messages_for_display(session, state_db_messages) -> list:
     """Return the display sidecar plus only necessary state.db rows for msg_limit.
 
-    Paginated session loads are latency-sensitive and should not stitch every
-    lineage segment before slicing the tail. Keep the lightweight
-    pre-compression snapshot stitch so continuation sessions can still reveal
-    archived history, then merge only newer state.db rows that have not reached
-    the sidecar yet.
+    Paginated session loads are latency-sensitive.  The full append-only merge
+    has intentionally broad fuzzy duplicate handling for recovery/import cases,
+    but that fuzzy pass is too expensive for ordinary large WebUI sidecars: a
+    6k-message transcript with only old state.db mirror rows can spend tens of
+    seconds comparing historical visible text before the response is sliced to a
+    100-message window.  For the limited display path, keep the sidecar as the
+    coordinate space and only merge state.db rows that are strictly newer than
+    the sidecar tail; older reconciliation remains available through the
+    deliberate full/heavy path.
     """
-    sidecar_messages = _webui_sidecar_lineage_messages_for_display(session)
+    sidecar_messages = list(getattr(session, "messages", []) or [])
+    if not sidecar_messages:
+        sidecar_messages = _webui_sidecar_lineage_messages_for_display(session)
     state_db_messages = list(state_db_messages or [])
     if not state_db_messages:
         return sidecar_messages
-    # NOTE: do not short-circuit to the sidecar when state.db has no strictly
-    # newer rows. A state.db row whose timestamp is at-or-before the sidecar's
-    # newest (recovery / edited-in-place / missing-timestamp cases) is still
-    # absent from the sidecar and must be reconciled — dropping it would render
-    # a tail that differs from the full merge path (silent wrong-transcript on
-    # the paginated load). The append-only merge is O(n) over already-bounded
-    # in-memory lists; the real latency win here is skipping the lineage-parent
-    # DISK load above, which we still skip. (#4070 ship-review)
+    if not sidecar_messages:
+        return merge_session_messages_append_only(
+            sidecar_messages,
+            state_db_messages,
+            truncation_watermark=getattr(session, "truncation_watermark", None),
+        )
+
+    max_sidecar_timestamp = None
+    for msg in sidecar_messages:
+        timestamp = _message_timestamp_as_float(msg)
+        if timestamp is not None:
+            max_sidecar_timestamp = (
+                timestamp
+                if max_sidecar_timestamp is None
+                else max(max_sidecar_timestamp, timestamp)
+            )
+    if max_sidecar_timestamp is None:
+        # Without timestamps we cannot safely identify a tiny state.db delta, so
+        # fall back to the precise merge. This shape is uncommon for modern
+        # WebUI sessions and preserves old recovery semantics.
+        return merge_session_messages_append_only(
+            sidecar_messages,
+            state_db_messages,
+            truncation_watermark=getattr(session, "truncation_watermark", None),
+        )
+
+    newer_state_rows = []
+    for msg in state_db_messages:
+        timestamp = _message_timestamp_as_float(msg)
+        if timestamp is not None and timestamp > max_sidecar_timestamp:
+            newer_state_rows.append(msg)
+    if not newer_state_rows:
+        return sidecar_messages
     return merge_session_messages_append_only(
         sidecar_messages,
-        state_db_messages,
+        newer_state_rows,
         truncation_watermark=getattr(session, "truncation_watermark", None),
     )
 
@@ -6483,6 +6516,7 @@ from api.models import (
     _active_stream_ids,
     _session_message_merge_key,
     _session_message_visible_key,
+    _message_timestamp_as_float,
     _is_empty_partial_activity_message,
     _hide_from_default_sidebar,
     prune_session_from_index,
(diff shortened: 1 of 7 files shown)


nesquena-hermes (changes requested):
Thanks for tackling the large-session latency problem — the bounding direction is genuinely worth doing, and most of this diff is good (the frontend pagination, the `&msg_limit=` plumbing across every `/api/session` fetch, the `_loadOlderMessages`/`_ensureAllMessagesLoaded` chunked walk, and the `include_heavy` opt-in are all sound). But the backend limited-load path has a confirmed correctness regression that has to be fixed before this can land.

## Blocker — bounded load silently drops archived pre-compression history (confirmed regression)

`_limited_webui_messages_for_display` (`api/routes.py`) replaced the lineage-stitch base:

```python
-    sidecar_messages = _webui_sidecar_lineage_messages_for_display(session)
+    sidecar_messages = list(getattr(session, "messages", []) or [])
```

and only falls back to the lineage stitch when `session.messages` is empty. Because the WebUI default load path is now always limited (`_DEFAULT_SESSION_MSG_LIMIT`), opening a compression-continuation session (a child whose `parent_session_id` points at a `pre_compression_snapshot=True` parent) no longer stitches the archived parent transcript — the user sees only post-compression turns.

It's also **unreachable**, not just unloaded: when the child's own messages fit in the window, `_messages_offset == 0`, so `_messages_truncated` is `False` and the frontend's "load older" path never fires. The archived history is only reachable via `include_heavy=1`, which the default UI never sends.

### Empirical proof (sandbox-gated, master vs head)

A pre-existing contract test that **this PR does not modify** (`git diff origin/master...HEAD -- tests/test_session_tail_payload.py` is empty) flips from green to red:

- `tests/test_session_tail_payload.py::test_msg_limit_tail_keeps_pre_compression_snapshot_parent_reachable`
  - on `origin/master`: **PASS** (rc=0)
  - on this PR head: **FAIL** (rc=1)
    ```
    AssertionError: assert ['continuation question','continuation answer']
                         == ['archived question','archived answer','continuation question','continuation answer']
    At index 0 diff: 'continuation question' != 'archived question'
    ```

The PR's own new `test_limited_webui_messages_uses_sidecar_without_old_state_merge` asserts that `_webui_sidecar_lineage_messages_for_display` must NOT run when the sidecar has messages — i.e. it locks in the exact behavior the existing contract test forbids. The two tests are mutually exclusive against this implementation, so the green CI on the new suite is illusory.

### Fix

Keep lineage stitching as the base on the limited path, and apply the newer-only state filter on top of it rather than replacing it:

```python
sidecar_messages = _webui_sidecar_lineage_messages_for_display(session)
```

`_webui_sidecar_lineage_messages_for_display` is already cheap for the common case — it only does the disk `Session.load` hops when a `pre_compression_snapshot` parent actually exists, and early-returns `session.messages` otherwise. The expensive thing you legitimately want to avoid is the full fuzzy `merge_session_messages_append_only` over all old `state.db` rows; that newer-only filter is fine to keep — it just needs to sit on top of the stitched sidecar. Then delete or rewrite the new `test_limited_webui_messages_uses_sidecar_without_old_state_merge` so it forbids the full old-state merge without also forbidding lineage stitching.

## Secondary (please also address)

The new early `if not newer_state_rows: return sidecar_messages` short-circuit reverses the deliberate `#4070` no-short-circuit guard. A `state.db` row whose timestamp is at-or-before the sidecar's newest (recovery / edited-in-place / missing-timestamp cases) is still absent from the sidecar and must be reconciled; dropping it makes the paginated tail diverge from the full-merge transcript for the same session. Either run the append-only merge whenever `state_db_messages` is non-empty (the merge is O(n) over already-bounded in-memory lists — the real win is skipping the lineage *disk* load, which the corrected fix above preserves for the no-parent case), or, if you keep the fast path, also include sidecar-absent state rows regardless of timestamp and add a test proving limited-path == full-path transcript for an at-or-before-timestamp state-only row.

Minor: no CHANGELOG entry for a user-visible behavior change (per `AGENTS.md`).

Happy to re-review as soon as the lineage stitch is restored on the limited path and the conflicting test is reconciled — the rest of the PR is in good shape.

nesquena-hermes:
Thanks @liangkeni — this is a genuinely valuable fix (large WebUI sessions are slow/failing to open, and windowing the load is the right idea). I warm-gated the current head (Codex + Opus + the full ~10.3k-test suite) and **the core windowing direction is sound**, but it can't ship yet: the suite surfaces **6 regressions** on the session-load surface — all reproduce on this PR's head and pass on `master`, so they're introduced by the windowing change, not pre-existing. Two I've already fixed on the branch for you; **4 still need your attention** before this is safe to merge.

I also confirmed the **data-loss bug** the earlier review flagged and **fixed it** (details below) — please keep that fix.

---

### ✅ Already fixed on the branch (please keep)

1. **Compression-continuation parent transcript was dropped on the windowed path** (`api/models.py` / `api/routes.py:_limited_webui_messages_for_display`). The function used raw `session.messages` as the coordinate-space base and only fell back to the lineage stitch when the sidecar was *empty*. A compression continuation has a *non-empty* child sidecar (post-compression turns) with its archived history in a `pre_compression_snapshot` **parent** — so the limited load silently dropped the parent transcript (recurrence of the #4767/#2616 ghost class). **Fix applied:** base is now `_webui_sidecar_lineage_messages_for_display(session)` (its fast-path returns the child sidecar unchanged for ordinary sessions, so the latency win is preserved; only continuations pay the bounded ≤20-hop stitch). Contract test `test_session_tail_payload.py::test_msg_limit_tail_keeps_pre_compression_snapshot_parent_reachable` now passes.

2. **Repeated same-content turn dropped on a windowed load** (`api/routes.py:_limited_webui_messages_for_display`, ~line 6026). Merging *only* newer-than-sidecar-tail state rows starves `merge_session_messages_append_only`'s replay index — a newer state row repeating a visible key already in the sidecar gets misclassified as the sidecar replay and skipped. **Fix applied:** when a newer state row shares a visible key with the sidecar, fall back to the full merge over all state rows. Fixes `test_state_db_reconciliation_does_not_collapse_repeated_content_with_different_timestamps`.

---

### ❌ Still failing — these need your rework (4)

These are the windowing/frontend behaviors the PR changes or removes that the existing contract tests guard. Each fails on the PR head and passes on master:

1. **`test_parallel_session_switch.py::TestMessagePaginationFrontend::test_load_older_uses_cumulative_tail_limit`** — the PR removes the `requestedLimit` / cumulative-tail-limit logic from `static/sessions.js` (`grep requestedLimit static/sessions.js` → 0 on the PR, 2 on master). "Load older" must keep growing the requested window cumulatively; the new windowing path drops that, so paging older messages regresses. Restore a cumulative tail limit on the load-older path.

2. **`test_tool_call_history_paging.py::test_sessions_js_resyncs_tool_calls_after_history_window_replacement`** — expects `function _syncToolCallsForLoadedMessages(messages, sessionToolCalls)` to run after a history-window replacement so tool-call cards re-sync. The PR's window-replacement path no longer drives that re-sync (the exact signature/call site the test asserts is gone). After replacing the loaded window, re-sync `S.toolCalls` for the new message set.

3. **`test_session_rotate_url_sync.py::test_stream_completion_syncs_rotated_session_id_to_tab_state`** — the stream-completion → rotated-session-id → tab-state sync marker the test locates is missing on the PR head (`assert completion_pos != -1` fails). A session that rotates its id on stream completion must still sync the new id to tab/URL state under the windowed load path.

4. **`test_webui_state_db_reconciliation.py::test_state_db_reconciliation_preserves_sidecar_order_when_timestamps_collide`** — backend windowing drop: a session with user+assistant+**tool** rows all at the *same* timestamp loses the trailing `tool` row on a `messages=1` (now default `msg_limit=100`) load (returns 2 of 3). In `_message_window_for_display` (`api/routes.py` ~5910), the trailing-tool-row keep loop only retains a tool row whose `tool_call_id` matches a tool-call **already in the window** — but when the preceding assistant row carries no `tool_calls`, the legitimate trailing tool row is treated as an orphan and dropped. The window slice needs to preserve a trailing tool row that belongs to the same turn even when its call isn't represented as a `tool_calls` entry on the visible row.

---

### How to proceed (your call)

Either path works for me:
- **Split it:** land the bounded-load + lineage-fix backend half first (with my two fixes), and do the frontend windowing changes (load-older cumulative tail, tool-call resync, rotate-sync) as a follow-up PR — smaller, easier to verify, lower blast radius on the crown-jewel session-load path; **or**
- **Fix in place:** address the 4 above on this branch and I'll re-gate the full suite.

My two fixes are on the branch I gated (`fix/4745-bound-session-loading`); happy to hand you the exact diffs or push them to your branch if you'd prefer — just say the word. Marking changes-requested. This is close and worth finishing — the session-load surface is just sensitive enough (it's the chat crown jewel) that I want all 6 green before it ships. Really solid root work on the windowing.

nesquena-hermes:
Thanks so much for this, @liangkeni — you correctly identified a real problem (large WebUI sessions were slow / failing to open because the load read the entire `state.db` history before trimming to the visible window), and windowing the load was the right instinct.

Closing this as **superseded**, because that exact fix has since shipped to `master` through a safer mechanism:

- **Backend tail-only read** — `/api/session` now pushes the client's `msg_limit` into SQL via a `since_timestamp` floor derived from the sidecar tail (`_state_db_since_timestamp_for_limited_display`), so it reads only the tail window instead of the whole transcript (the ~1.75s cold read on an 844-message session drops to a few hundred ms). Critically, that optimization is taken **only when it is provably output-identical to the full read** (it keeps NULL-timestamp rows and compares the full merge identity of the sidecar/state.db prefixes, bailing to the full read otherwise). Shipped via #4920 (credit @franksong2702).
- **Frontend default bounding** — the session load already sends a default `msg_limit` on every fetch, so ordinary loads are windowed without any further change.

When I gate-reviewed this PR against the full suite, the timestamp-delta merge approach here (merge only state.db rows strictly newer than the sidecar tail) surfaced 6 regressions on the session-load surface — including dropping a compression-continuation's archived parent transcript and dropping a trailing tool row on same-timestamp turns (real data-loss). The shipped `since_timestamp` approach avoids that whole class by construction, which is why we're going with it.

So the outcome you were after — fast large-session loads — is now live, just via the output-identical path. Genuinely appreciate the root-cause work; it's the kind of reliability fix we prioritize. 🙏

Closing as superseded by #4920.

PR #4750, opened 2026-06-23
fix(send): clear textarea immediately after capturing to prevent stalled-text race

harryazj:
## Bug Description

When `busy_input_mode` is `interrupt` and the user types a new message while
the agent is still processing, the new message enters the queue and is drained
after stream cancel. During the queue drain (`ui.js:3702-3716`), `send()` is
called again, but the textarea may not have been cleared from the previous
`send()` call, causing old text to be read together with new text.

## Root Cause

In `messages.js` `send()`, the textarea is read at line 310:

    const text=$('msg').value.trim();

But it is not cleared until line 464:

    $('msg').value='';autoResize();

Between these two lines there are ~150 lines of async operations (busy checks,
file uploads, session creation). When `send()` is re-entered during this window
— via interrupt queue drain or the `_sendInProgress` guard — the
textarea still holds the previous message text mixed with the new input.

## Steps to Reproduce

1. Set `busy_input_mode` to `interrupt`
2. Send a long message that keeps the agent busy for 30+ seconds
3. While the agent is processing, type a new short message and press Enter
4. Observe the session history — the new user message contains old text

## Fix

Move the textarea clear to immediately after the value is captured:

    const text=$('msg').value.trim();
    $('msg').value='';autoResize();   // <-- move here from line 464

And remove the redundant clear at line 464. The captured `text` variable
is the sole authority from line 310 onward.

This mirrors how messaging platforms (Telegram, Discord, Feishu) handle input:
the message is captured atomically at submission time, not re-read from shared
mutable DOM state later.

## Risk Assessment

- **Network failure**: user message already optimistically rendered (line 473). No data loss.
- **File upload failure**: pending files in `S.pendingFiles` array, not textarea.
- **Browser crash**: same ~100ms window as current code (clear at 464 also pre-request).
- **Composer draft**: `_clearComposerDraft` called separately, unaffected.
- **_sendInProgress guard**: now reads empty textarea on re-entry, correctly preventing re-queues.


nesquena-hermes:
Thanks @harryazj — the core idea here is good and we want it: capturing + clearing the composer textarea **atomically** before any async work, so a fast second read can't see stalled text (the lost-response race). That part (the `static/messages.js` capture-and-clear + `tests/test_session_lost_response_regression.py`) is exactly the kind of fix we'll merge.

The problem is scope: this PR is **+1684/-852 across 20 files**, and the great majority is unrelated to the send-race fix —

- `README.md` (+31/-755) + `README.en.md` (+783) — a full README rewrite/translation
- `.github/workflows/sync-to-cnb-mirror.yml`, `.github/workflows/upstream-watch.yml`, `scripts/sync-upstream.sh`, `MAINTAINING.md` — fork/mirror tooling (these look specific to a downstream mirror, not upstream `nesquena/hermes-webui`)
- `static/i18n.js` (+430/-34), `api/onboarding.py`, `api/profiles.py`, `api/config.py`, `static/panels.js`, `static/boot.js` — onboarding/i18n/config changes

Per our contribution guidelines (one logical change per PR), could you **split this down to just the send-race fix** — `static/messages.js` (the atomic capture-and-clear) plus its regression test? That's a clean, reviewable, mergeable change we can gate and ship quickly.

The other pieces, if you'd like them considered upstream, are very welcome as **separate focused PRs** (one for the README/i18n, one for any onboarding/config behavior change with a description of what it fixes). The mirror/sync workflow files are likely downstream-only and probably shouldn't target this repo at all — but happy to look if you think they belong here.

It's also currently CONFLICTING, so a rebase onto current `master` will be needed for whichever slice you keep. Marking changes-requested for the split. Thanks again — the send-race fix itself is a real catch.

nesquena-hermes (changes requested):
Scope: please split down to just the send-race fix (static/messages.js + its regression test). The +1684/-852 across 20 files (README rewrite, mirror workflows, i18n, onboarding/config) needs to be separate focused PRs — details in the comment. Also needs a rebase (CONFLICTING).

nesquena-hermes:
Closing — thanks @harryazj! The atomic composer capture-and-clear fix (moving the textarea clear to right after capture so a re-entrant/interrupt-mode send can't re-read stale DOM and double-send) was the real, valuable piece buried in this PR. We pulled JUST that fix fresh onto master as **#5912** with your work preserved (`Co-authored-by`) + a regression test, and left the larger 20-file fork-maintenance bundle (CN-mirror workflows, README rewrites) out of upstream. Thanks for the fix!

PR #4758, opened 2026-06-23
Deduplicate live progress reasoning echoes

franksong2702:
## Summary

- Strip reasoning-tail echoes when a runtime emits the same live progress sentence first through reasoning and then through `interim_assistant`.
- Remove echoed live Thinking/Anchor reasoning rows on the client before the interim progress row is rendered.
- Deduplicate the same reasoning/interim echo during run-journal snapshot replay so reconnect/settle cannot rebuild duplicate Thinking content.

## Root cause

Some reasoning-heavy runtimes report a process update through the reasoning callback before later reporting the same text through `interim_assistant`. The Anchor-backed worklog then had two valid-looking rows for the same sentence: a Thinking row and a Process/prose row. The live client also applied the anchor row before flushing the legacy live assistant segment, so the old outside segment could survive the anchor hide pass.

## Validation

- `node --check static/messages.js && node --check static/ui.js && node --check static/assistant_turn_anchors.js`
- `python3 -m py_compile api/streaming.py api/routes.py`
- `./scripts/test.sh tests/test_run_journal_streaming_static.py tests/test_anchor_scene_persistence.py tests/test_stable_assistant_turn_anchor_phase0.py tests/test_live_to_final_anchor_visible_order.py tests/test_gateway_sync.py tests/test_issue_progress_echo_dedupe.py tests/test_live_anchor_progress_echo.py tests/test_issue4729_reasoning_sse_coalesce.py tests/test_issue2565_reasoning_accumulation.py` (`165 passed`)
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Compact Worklog no longer shows the same live progress sentence as both Process and Thinking.** Some reasoning-heavy runtimes emit a user-facing progress update first through the reasoning callback and then through `interim_assistant`; the Anchor-backed live Worklog now treats that as one visible progress row, strips the duplicated reasoning tail from live/durable state, and keeps journal replay from rebuilding the duplicate Thinking row.
+
 ## [v0.51.598] — 2026-06-23 — Release VE (preserve scroll across live worklog rebuilds)
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


nesquena-hermes:
Shipped in **v0.51.599** (Release VF) 🚀 — thank you @franksong2702. The same-progress-sentence-twice dup (Process + Thinking) is gone. Since this modifies the #4729 reasoning-coalesce buffer shipped earlier today, I verified the interaction carefully: `_strip_compact_echo_suffix` is a no-op unless the reasoning tail exactly matches the interim sentence (whitespace-folded), operates only on the un-flushed buffer tail (so it can't drop already-delivered reasoning), and errs conservative — confirmed by a direct regression unit-test (normal reasoning left fully intact) plus Codex SAFE + full suite green. Verified live in prod.

Issue #3831, opened 2026-06-08
Bug: truncation_watermark permanently set by retry/undo causes progressive session message loss via state.db reconciliation


Pandasywithfantasy:
Labels: bug, session, data-loss

Bug Description
When a user performs /retry, /undo, or edits a message (the pencil button), the session is truncated and truncation_watermark is set to the timestamp of the last kept message. This watermark is never cleared when new messages are subsequently added.

Once set, merge_session_messages_append_only() uses this watermark to filter state.db messages:

State.db messages after the watermark that are not in the sidecar JSON → skipped
State.db messages before the watermark whose content is not in the sidecar → skipped

As the conversation continues past the watermark, the gap grows. The middle portion of the conversation becomes permanently invisible to the WebUI. Each subsequent s.save() perpetuates the truncated state.
Impact
Progressive, irreversible message loss from the WebUI session JSON
Hermes state.db still has complete data, but reconciliation fails to restore it
.bak safeguard (#1558) preserves a snapshot at truncation time, but is also downstream of the bug
Steps to Reproduce
Start a long conversation (20+ user turns over multiple days)
Use /retry or edit a message
Continue the conversation for many more turns
Close and reopen the session
Messages in the middle of the timeline are gone — GET /api/session fails to restore them
Root Cause
api/session_ops.py — retry_last() (line 118) and undo_last() set s.truncation_watermark after truncating
api/routes.py — /api/session/truncate handler does the same (line 7366)
api/models.py — merge_session_messages_append_only() (line 4237) uses watermark to filter, with no code path to clear it when new messages arrive
Suggested Fix
Clear s.truncation_watermark = 0.0 when new user messages are successfully appended after a retry/undo event. Or, alternatively, bound the watermark filter to a limited window so it doesn't permanently shadow state.db messages.

nesquena-hermes:
Investigated this from first principles against `origin/master` — driving the actual `merge_session_messages_append_only()` consumer with the scenarios described — and the headline claim ("progressive, irreversible loss of the middle of the conversation as it continues past the watermark") does **not** reproduce on current master. Here's what I found.

## The watermark genuinely is never cleared — that part is correct

[...]

PR #3102, opened 2026-05-28
Fix for #2914: stop agent from seeing deleted messages after Edit/Undo (from state.db)

AlexeyDsov:
### Problem

After **Edit**, **Retry**, or **Undo** in the WebUI, the agent could still see original messages from `state.db` even though the UI correctly showed only the truncated conversation. If you asked the agent "what messages do you see?", it would report both the old and new versions.

> **Note:** This issue (#2914) was partially addressed in #2933, which introduced the `truncation_watermark` mechanism. However, two gaps remained: `context_messages` was not truncated in sync with `messages`, and the watermark filter only checked `m_ts > watermark`, allowing original edited messages (with _older_ timestamps) to slip through.

### Root Cause

When a user edits or truncates a message, the WebUI sidecar (`.json` session file) is updated correctly, but `state.db` is an append-only store — the original messages remain there forever. During session load, `state.db` rows are replayed and merged with the sidecar. The `truncation_watermark` mechanism was supposed to filter out pre-truncation rows, but had two gaps:

1. **`context_messages` was not truncated** alongside `messages`, so the agent's model-facing context retained rows the user deliberately removed.
2. **Watermark filtering only checked `m_ts > watermark`** — but original edited messages have an _older_ timestamp (before the edit), so they passed through the filter unblocked.

### Fix

- `session_ops.py`: `retry_last()` and `undo_last()` now truncate `context_messages` in sync with `messages`.
- `models.py`: `reconciled_state_db_messages_for_session()` now uses the sidecar's message IDs as authoritative — state.db rows whose content duplicates a sidecar message at a different timestamp are deduplicated, and rows preceding the watermark are properly excluded.
- `truncation_watermark` is persisted on every truncation operation (Edit, Retry, Undo) and survives reload.

### Reproduction Examples

**Example 1 — Edit leaks original into context:**

1. Send: `"What is the sum of angles in a triangle?"` → Agent replies about 180°
2. **Edit** the message to: `"What is the sum of angles in a square?"` → Agent replies about 360°
3. Send: `"List all my messages in this session"` → **Bug:** Agent reports **both** "triangle" and "square", even though the UI only shows "square"

**Example 2 — Edit then Undo leaks original:**

1. Send: `"What is the sum of angles in a triangle?"` → Agent replies about 180°
2. **Edit** the message to: `"What is the sum of angles in a square?"` → Agent replies about 360°
3. Send another message: `"What is the speed of light?"`
4. **Undo** — removes the "speed of light" message
[...]

Issue #2914, opened 2026-05-25
Bug: `/undo` and message edit appear to succeed but have no visual effect


AlexeyDsov:
### Summary

Destructive session operations (`/undo`, message editing via pencil button, `/retry`) report success but don't actually remove messages from the display. The toast confirms the action ("↩ removed 2 messages"), but the messages reappear on page reload or after switching sessions and coming back.

In case of message editing, both the old and new versions of the exchange are visible — the old one was supposed to be truncated before re-sending.

### Note
This issue was triaged and described with AI assistance. The root cause analysis and suspected commits are based on code inspection and may need manual verification.

### Reproduction

**`/undo`:**
1. Start a WebUI session and send several messages (e.g., 4 user/assistant exchanges = 8 messages).
2. Type `/undo` or click the undo button on the last assistant message.
3. Toast shows "↩ removed 2 messages" — appears successful.
4. Reload the page or switch to another session and come back.
5. The "removed" messages are back.

**Message editing:**
1. Click the pencil icon on any user message.
2. Edit the text and click "Send edit".
3. The old message and the new (edited) response both appear — the old exchange was supposed to be truncated before re-sending.
4. On refresh, the old messages reappear.

### Expected behavior

- `/undo` permanently removes the last user/assistant exchange from the session.
- Message editing truncates everything from the edited message onward, then re-sends the edited text as a fresh exchange.
- These changes survive page reloads and session switches.

### Actual behavior

The operations modify the WebUI sidecar (session JSON file), but `GET /api/session` merges messages from two sources:

```python
_all_msgs = merge_session_messages_append_only(s.messages, state_db_messages)
```

- `s.messages` — the sidecar, which **was** truncated by the destructive operation
- `state_db_messages` — `state.db` (SQLite), which was **not** touched

The `merge_session_messages_append_only` function is designed to be append-only: it adds state.db rows not present in the sidecar, but has no concept of intentional deletion. Messages "deleted" from the sidecar are re-appended from `state.db` on the next fetch.

### Affected code paths

| Operation | Frontend | Backend | What it modifies |
|-----------|----------|---------|-----------------|
| `/undo` | `cmdUndo()` in `static/commands.js:1025` | `POST /api/session/undo` → `undo_last()` in `api/session_ops.py:86` | `s.messages` (sidecar only) |
[...]

PR #2620, opened 2026-05-20
fix(session): use second-level timestamp granularity in legacy dedup key

bengdan:
## Summary

Fix duplicate messages in chat (closes #2616).

`_normalized_message_timestamp_for_key` was preserving microsecond precision (`%.6f`). When the same message is persisted by both the WebUI sidecar JSON writer and the Hermes agent `state.db` writer, their timestamps can differ by a few microseconds. This caused `_session_message_merge_key` to produce different keys for the same logical message, letting both copies survive the dedup pass in `merge_session_messages_append_only` — resulting in every user message and AI response appearing twice in the UI.

## Fix

Truncate to second-level granularity (`int(timestamp)`) so sub-second drift between the two writers collapses to the same key.

## Reproduction

1. Run WebUI v0.51.95
2. Send any message
3. Both the user message and AI response appear twice

v0.51.90 is not affected.

Closes #2616

@nesquena Could you review and merge? One-line fix for the regression introduced in stage-387/388.

PR #4772, opened 2026-06-23
fix(#4767): advance truncation_watermark instead of clearing to None after edit/retry/undo

### LINKED ISSUE #4767: Editing a message and switching sessions causes duplicate user messages and assistant replies to reappear
**Version:**
v0.51.601

**Steps to reproduce:**  

1. Start a new session and send a message (e.g., "What is the speed of light in a vacuum?"). Receive the assistant's reply.
2. Send a second message in the same session (e.g., "What is the sum of angles in a triangle?"). Receive the reply.
3. Edit the second message, changing the content (e.g., replace "triangle" with "square"). Receive the new reply — at this point the session looks correct.
4. Switch to a different session, then switch back to the original session.

**Expected result:**  
The session should show the two user messages: message 1 (original) and message 2 (edited version only), along with their corresponding assistant replies.

**Actual result:**  
After switching away and back, the original (un-edited) version of message 2 reappears together with the assistant's reply to it. The user now sees three messages instead of two. Critically, **all these ghost messages — both the resurrected user message and the assistant's reply to it — are also fed into the agent's context window**, polluting the conversation context with content that should have been discarded by the edit.

**More info:**
There were similar problems in the issues #2616 and was fixed in #2620 #3102 and maybe in several more. But now the problem is being reproduced again. I checked for versions v0.51.601.

There are also similar problems with the /retry and /undo commands.

AlexeyDsov:
Closes #4767

### Problem

After editing a message (or using `/retry` or `/undo`), switching to another session and switching back caused the original pre-edit message and its assistant reply to reappear. These ghost messages were also fed into the agent's context window, polluting the conversation.

### Root cause

`truncation_watermark` was cleared to `None` when a new user turn was committed after edit/retry/undo. Without the watermark, `merge_session_messages_append_only()` lost the boundary needed to filter out replaced pre-edit rows from the append-only state.db, so they leaked back into the merged transcript on reload.

### Fix

Advance `truncation_watermark` to the newest user message timestamp instead of clearing it. The existing `sidecar_advanced_past_watermark` guard in the merge logic already allows post-edit state.db rows to merge in, while the advanced watermark keeps filtering replaced pre-edit rows whose timestamps fall below the boundary.

### Changes

- `api/routes.py`: `_checkpoint_user_message_for_eager_session_save` — advance watermark to new message timestamp
- `api/models.py`: `_append_recovered_pending_turn` — advance watermark to recovered timestamp
- `api/streaming.py`: `_retire_truncation_watermark_after_commit` → `_advance_truncation_watermark_after_commit` — find newest user message timestamp and advance watermark; `_materialize_pending_user_turn_before_error` — advance to recovered timestamp
- `tests/test_issue3831_watermark_clear.py`: updated to reflect advance semantics
- `tests/test_watermark_advance_after_edit.py`: new end-to-end tests for edit/retry/undo → new turn → reload scenarios

### Related

- #2914 (state.db tail replay after undo/retry/edit)
- #3831 (stale watermark data loss)
- #2616, #2620, #3102 (prior related fixes)

### AI Assistance

This fix was developed with assistance from Qwen3.6-27B. The AI helped analyze the reconciliation logic, identify root cause, fix code and test scenarios.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -428,12 +428,15 @@ def _append_recovered_pending_turn(session, *, timestamp: int | None = None) -> 
         recovered['attachments'] = list(session.pending_attachments)
     session.messages.append(recovered)
     _append_recovered_turn_to_context(session, recovered)
-    # The new user turn is now committed to messages (#3831): retire a positive
-    # truncation watermark from a prior retry/undo/edit so it can't freeze at the
-    # old edit boundary and drop these post-edit turns on a later empty-sidecar
-    # reconcile. None, never 0.0 (the truncate-to-empty sentinel, #2914).
+    # The new user turn is now committed to messages (#3831): advance the
+    # truncation watermark to the new message's timestamp so that
+    # merge_session_messages_append_only() still filters out replaced
+    # pre-edit rows from state.db whose timestamps fall below the boundary.
+    # The merge's sidecar_advanced_past_watermark guard allows state.db rows
+    # newer than the watermark, so post-edit turns are not dropped.
+    # Never 0.0 (the truncate-to-empty sentinel, #2914).
     if getattr(session, 'truncation_watermark', None):
-        session.truncation_watermark = None
+        session.truncation_watermark = recovered_ts
     return recovered
 
 
--- a/api/models.py
+++ b/api/models.py
@@ -5085,23 +5085,45 @@ def merge_session_messages_append_only(
     if not state_messages:
         return sidecar_messages
     if not sidecar_messages:
-        if watermark_timestamp is not None:
-            filtered = [
-                msg for msg in state_messages
-                if (
-                    (timestamp := _message_timestamp_as_float(msg)) is not None
-                    and timestamp <= watermark_timestamp
-                )
-            ]
-        else:
+        if watermark_timestamp is None:
+            # No watermark — keep everything, just dedup.
             filtered = state_messages
+        elif watermark_timestamp == 0:
+            # Truncate-to-empty sentinel (#2914) — block all replay.
+            return []
+        else:
+            # Positive watermark advanced after edit/retry/undo (#4767).
+            # Without a sidecar there's no seen_content_keys to check against,
+            # so we reconstruct the correct transcript from state.db alone:
+            #   1. Keep all messages at/after the watermark (the new turn +
+            #      post-edit assistant reply — these are always legitimate).
+            #   2. For pre-watermark messages, drop the last complete turn
+            #      (user + assistant pair) which is the replaced tail.
+            #      Everything before that is legitimate history.
+            at_or_after = []
+            pre_watermark = []
+            for msg in state_messages:
+                ts = _message_timestamp_as_float(msg)
+                if ts is not None and ts >= watermark_timestamp:
+                    at_or_after.append(msg)
+                else:
+                    pre_watermark.append(msg)
+            # Scan pre_watermark from the end: skip assistant messages until
+            # we hit a user message (the replaced prompt), then stop.
+            i = len(pre_watermark) - 1
+            while i >= 0:
+                role = str(pre_watermark[i].get("role", "")).lower()
+                if role == "assistant":
+                    i -= 1
+                elif role == "user":
+                    i -= 1  # skip this user message too (the replaced prompt)
+                    break
+                else:
+                    i -= 1  # tool messages etc. — skip
+            filtered = pre_watermark[:i + 1] + at_or_after
+
         # Deduplicate true duplicates (same role, content, exact timestamp)
         # without collapsing legitimately-repeated identical turns (#3346).
-        # Note: rows whose timestamps were mutated by compaction/recovery to
-        # microsecond-different values will not be folded — only byte-identical
-        # timestamps are treated as the same message.  This is intentional;
-        # collapsing same-second distinct turns would be worse than retaining
-        # a compaction-restamped duplicate.
         seen = set()
         deduped = []
         for msg in filtered:
--- a/api/models.py
+++ b/api/models.py
@@ -5174,8 +5174,14 @@ def merge_session_messages_append_only(
         # advanced), allow state rows newer than the sidecar tail to merge.
         sidecar_advanced_past_watermark = (
             watermark_timestamp is not None
-            and max_sidecar_timestamp is not None
-            and max_sidecar_timestamp > watermark_timestamp
+            and (
+                (max_sidecar_timestamp is not None
+                 and max_sidecar_timestamp > watermark_timestamp)
+                # If the sidecar is empty but watermark > 0, the session has
+                # advanced (a new user turn was committed).  Treat this as
+                # advanced so post-edit state.db rows are not dropped.
+                or (not sidecar_messages and watermark_timestamp > 0)
+            )
         )
         if (
             watermark_timestamp is not None
@@ -5198,6 +5204,18 @@ def merge_session_messages_append_only(
             watermark_timestamp is not None
             and timestamp is not None
             and timestamp < watermark_timestamp
+            and key not in seen_message_keys
+            and _session_message_content_key(msg) not in seen_content_keys
+        ):
+            continue
+        # Same-second edit: if timestamp equals the watermark and the message
+        # content is not in the sidecar, it's a replaced message edited at the
+        # same second — skip it.  The edited version (same timestamp, different
+        # content) is in the sidecar and survives this check.
+        if (
+            watermark_timestamp is not None
+            and timestamp is not None
+            and timestamp == watermark_timestamp
             and key not in seen_message_keys
             and _session_message_content_key(msg) not in seen_content_keys
         ):
--- a/api/models.py
+++ b/api/models.py
@@ -646,6 +646,7 @@ class Session:
                  context_length=None, threshold_tokens=None,
                  last_prompt_tokens=None,
                  truncation_watermark=None,
+                 truncation_boundary=None,
                  gateway_routing=None, gateway_routing_history=None,
                  llm_title_generated: bool=False,
                  manual_title: bool=False,
@@ -696,6 +697,7 @@ class Session:
         self.threshold_tokens = threshold_tokens
         self.last_prompt_tokens = last_prompt_tokens
         self.truncation_watermark = truncation_watermark
+        self.truncation_boundary = truncation_boundary
         self.gateway_routing = gateway_routing if isinstance(gateway_routing, dict) else None
         self.gateway_routing_history = gateway_routing_history if isinstance(gateway_routing_history, list) else []
         self.llm_title_generated = bool(llm_title_generated)
@@ -765,6 +767,7 @@ class Session:
             'compression_anchor_details', 'context_engine_state',
             'context_length', 'threshold_tokens', 'last_prompt_tokens',
             'truncation_watermark',
+            'truncation_boundary',
             'gateway_routing', 'gateway_routing_history', 'llm_title_generated', 'manual_title',
             'parent_session_id',
             'worktree_path', 'worktree_branch', 'worktree_repo_root', 'worktree_created_at',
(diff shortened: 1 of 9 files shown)


nesquena-hermes:
Thanks @AlexeyDsov — this is a real bug (ghost pre-edit rows resurrecting + polluting agent context) and "advance the watermark instead of clearing it" is the right direction. I warm-gated it (Codex + full suite). The suite is green (10207), but Codex reproduced **two CORE edge cases the suite doesn't cover** — both on the core state.db merge path, so they need fixing before this can ship:

### 1. (CORE) Same-second edit/retry can still resurrect the pre-edit rows — `api/models.py:5197`
The stale-row filter uses `timestamp < truncation_watermark`, but the patch advances the watermark to the replacement user message's **exact** timestamp. With integer-second timestamps, the old pre-edit row's timestamp can **tie** the watermark and fall through `<` into the merge → the ghost user+assistant rows reappear on a same-second edit/retry. So the fix works for slow edits but not fast ones.
**Fix:** make the absent-state-row boundary **inclusive** (`<= truncation_watermark`), or advance the watermark to a strictly-greater exclusive boundary. Add a same-timestamp regression test (two turns within the same integer second).

### 2. (CORE, worse) Empty-sidecar recovery now drops legitimate post-edit rows — `api/models.py:5087`
The empty-sidecar branch keeps only state.db rows `<= watermark`. With the watermark now advanced to the new user timestamp, that branch **drops the post-edit assistant reply and any later post-edit turns** — which reintroduces the exact #3831 data-loss shape this lineage was fixing, just in the other direction. The `sidecar_advanced_past_watermark` guard you're relying on to let post-edit rows merge **cannot fire when the sidecar is empty**, so that path is unprotected.
**Fix:** restore empty-sidecar coverage so committed post-edit turns are recoverable without depending on `sidecar_advanced_past_watermark` (which is a no-op for an empty sidecar). This likely needs a small design adjustment, not a one-liner — worth confirming the empty-sidecar reload path keeps everything at-or-after the watermark.

These are exactly the two-directional risks this watermark code has to thread: too-low a boundary re-leaks the ghost (#1), too-high drops real data (#2, the #3831 trap). A green suite isn't sufficient here because neither edge (same-second tie, empty-sidecar reload) is currently exercised — please add a regression for each.

Everything else looks sound (the advance-not-clear approach, the commit-path coverage for the non-empty-sidecar case). Marking changes-requested; happy to re-gate the moment both boundaries are fixed + covered. Solid root-cause analysis on this one.

nesquena-hermes (changes requested):
Warm-gated: suite green but Codex reproduced 2 CORE edge cases the suite doesn't cover — (1) same-second edit/retry still resurrects the ghost (timestamp tie vs the < boundary), (2) empty-sidecar recovery drops legitimate post-edit rows (reintroduces #3831 data loss). Both on the state.db merge path. Details + exact fixes in the comment.

nesquena-hermes:
Thanks for the quick turnaround @AlexeyDsov — but the re-push (the "Update api/streaming.py" commit) doesn't reach either CORE finding. Both live in **`api/models.py`** (the merge/filter logic), not `streaming.py` (which only sets the watermark). I checked the new head directly:

1. **Same-second tie — still present.** `api/models.py` (~line 5197, the replaced/stale-row filter) still reads:
   ```python
   timestamp is not None and timestamp < watermark_timestamp
   ```
   Since the watermark is advanced to the replacement user message's exact timestamp, an old pre-edit row that shares that integer second ties and falls through `<` → the ghost still resurrects on a same-second edit/retry. Change this boundary to **`<=`** (or advance the watermark to a strictly-greater value), and add a same-second regression.

2. **Empty-sidecar data loss — still present.** `api/models.py` (~line 5087, the `if not sidecar_messages:` branch) still keeps only `timestamp <= watermark_timestamp`:
   ```python
   if not sidecar_messages:
       if watermark_timestamp is not None:
           ... and timestamp <= watermark_timestamp
   ```
   With the watermark now advanced to the new user timestamp, this drops the post-edit assistant reply + later post-edit turns whenever the sidecar is empty → reintroduces the #3831 data-loss shape. The `sidecar_advanced_past_watermark` guard you're relying on can't help here because it never fires on an empty sidecar. This branch needs to keep post-edit rows recoverable independent of that guard.

So both fixes belong in `api/models.py`'s merge function (`merge_session_messages_append_only` / the absent-state-row filter), not in `streaming.py`. The streaming.py side (advancing the watermark) is already correct. Still changes-requested; ping me when models.py is updated + both edges have regression tests and I'll re-gate.

nesquena-hermes:
Thanks for the fast turnaround @AlexeyDsov — the 14:19 re-push **did converge CORE finding #1** (same-second edit/retry timestamp-tie): I re-gated it (Codex, GPT-5.5, high reasoning, with direct reproduction harnesses) and the same-second non-empty-sidecar filtering verifies clean now. Good fix.

**But CORE finding #2 is still live** — and the gate **reproduced it directly** on the current head, so this can't ship yet.

### CORE — empty-sidecar recovery still corrupts the transcript — `api/models.py:5087`

The empty-sidecar branch returns *before* your new `sidecar_advanced_past_watermark` logic can run. `merge_session_messages_append_only()` short-circuits at line 5087:

```python
if not sidecar_messages:
    if watermark_timestamp is not None:
        filtered = [
            msg for msg in state_messages
            if (
                (timestamp := _message_timestamp_as_float(msg)) is not None
                and timestamp <= watermark_timestamp   # <-- line 5093
            )
        ]
    ...
    return deduped   # <-- returns here, never reaches the advanced-watermark guard
```

Because the watermark is now **advanced** to the new user-turn timestamp, this branch treats the advanced watermark as an **upper ceiling** instead of a replaced-tail boundary, so it does *both* wrong things at once:

- **drops** the legitimate post-edit assistant reply (and any later post-edit turns), ts `> watermark` → reintroduces the exact #3831 data-loss shape, and
- **keeps** the stale pre-edit rows (ts `≤ watermark`) → resurrects the ghost.

**Direct reproduction on your head** (not inferred — run):

```python
from api.models import merge_session_messages_append_only as merge
# empty sidecar (cold reload / crash window before sidecar persists),
# watermark advanced to the new user turn @200, state.db has the full lineage:
state = [
    {"role":"user","content":"first msg","timestamp":50},
    {"role":"assistant","content":"first reply","timestamp":51},
    {"role":"user","content":"original pre-edit","timestamp":100},   # replaced by the edit
    {"role":"assistant","content":"original reply","timestamp":101}, # replaced by the edit
    {"role":"user","content":"edited/new turn","timestamp":200},
    {"role":"assistant","content":"post-edit reply","timestamp":201},# legitimate, MUST survive
]
print([m["content"] for m in merge([], state, truncation_watermark=200.0)])
# ACTUAL (head): ['first msg','first reply','original pre-edit','original reply','edited/new turn']
#   -> 'post-edit reply'@201 DROPPED  +  'original pre-edit'/'original reply' RESURRECTED
# WANTED:        ['first msg','first reply','edited/new turn','post-edit reply']
```

The `sidecar_advanced_past_watermark` guard you're relying on to admit post-edit rows **cannot fire on an empty sidecar** (it's computed from `max_sidecar_timestamp`, which is `None` here), so this path is completely unprotected.

### Fix direction (both fixes belong in `api/models.py`)
This one is more than a one-liner — the empty-sidecar branch needs to stop treating an advanced positive watermark as a "keep everything ≤ watermark" ceiling:
1. For a **positive advanced watermark + empty sidecar**, recover committed post-edit rows **at/after** the boundary instead of dropping them, while still filtering the replaced pre-edit tail. (The challenge is that with an empty sidecar there's no `seen_*` signal for *which* rows were replaced — so consider whether the replaced-tail rows can be identified by the same role/content-key dedup the main path uses, or whether the empty-sidecar path should fall through into the main merge loop rather than short-circuiting.)
2. **Preserve the `0.0` truncate-to-empty block-all sentinel (#2914)** — that path must keep blocking replay; only a *positive* advanced watermark gets the new treatment.
3. Add a regression: empty sidecar + advanced watermark + stale pre-edit rows + post-edit assistant/later rows → keeps the post-edit rows **without** resurrecting the stale ones (i.e. the reproduction above, asserted).

Your focused suite (`19 passed`) and the existing #2914 suite (`12 passed`) both stay green — neither exercises the empty-sidecar advanced-watermark case, which is why the green suite isn't sufficient here.

Marking changes-requested. Ping me the moment the empty-sidecar branch is fixed + covered and I'll re-gate immediately — you're one branch away. Really solid root-causing on this lineage.

nesquena-hermes:
Thanks for the continued iteration @AlexeyDsov — CORE#1 (same-second tie) and my earlier CORE#2 (the basic empty-sidecar post-edit-reply case) are both resolved now; I verified the simple cases pass. But a deep P0 re-gate (Codex GPT-5.5, direct reproduction) found **two more reproduced data-loss cases** in the empty-sidecar recovery path, and they point at a **design limitation** rather than a one-liner — so I can't ship this yet. Both were reproduced against the current head (`297c88fed7`).

### CORE-A — empty-sidecar recovery resurrects deleted pre-edit turns when editing an *older* message with more than one later turn (`api/models.py:5127`)

The empty-sidecar branch infers the boundary by dropping the *last* pre-watermark user/assistant pair. But `/api/session/truncate` (edit/retry/undo) can remove an **arbitrary suffix**, and the new commit sites advance the *only* stored boundary (`truncation_watermark`) to the new user timestamp — overwriting the original truncation point. So on empty-sidecar recovery there's no longer enough information to tell a legitimate prefix from a deleted suffix, and editing an older message (with ≥2 later turns) resurrects the deleted turns.

**Root issue:** advancing the watermark *destroys* the very information (the original truncation boundary) that empty-sidecar recovery needs.

**Fix direction:** persist the original truncation boundary **separately** when advancing the watermark, and in empty-sidecar recovery keep only `state.timestamp <= original_boundary` **plus** `state.timestamp >= advanced_watermark` — don't infer by dropping one turn.

### CORE-B — same-second recovery can silently drop the legitimate post-edit assistant reply (`api/models.py:5290`)

The equality guard skips any state.db row at `timestamp == watermark` whose content isn't already in the sidecar. With the sidecar holding only the edited user checkpoint and state.db holding a **same-second assistant reply**, the merge returns only the user turn — the real reply is dropped.

**Fix direction:** narrow the equality filter so it can't discard state-only assistant/tool recovery rows, or use the persisted truncate-boundary / turn identity instead of timestamp equality alone.

### Why I'm bouncing rather than patching inline
This is the crown-jewel append-only merge on a P0 data-loss path, and CORE-A is a design-level gap (the boundary is overwritten), not a surface bug. Patching it inline at release time is too high-risk — a wrong fix trades one data-loss for another. The clean path is to **persist enough boundary/turn identity to distinguish a legitimate prefix from a deleted suffix**, then the empty-sidecar recovery becomes well-defined.

The good news: the in-context (non-empty sidecar) path and the simple edit cases are solid now — this is specifically the empty-sidecar *recovery* reconstruction that needs the original boundary preserved. Codex reproduced both cases directly; happy to share the exact repro scripts if useful. Re-gate as soon as the boundary is persisted + same-second state-only recovery is covered. Keeping `changes-requested`.

nesquena-hermes:
Shipped in **v0.51.628** (Release WI) — thanks @AlexeyDsov! 🎉 Closes #4767.

Your branch landed the core design — advance `truncation_watermark` instead of clearing it, and persist the original cutoff as a separate `truncation_boundary` so empty-sidecar recovery can tell a legitimate prefix from a deleted suffix — plus the same-second CORE-B fix. I built the rest on top of your branch (every commit carries a `Co-authored-by: AlexeyDsov` trailer) so you keep authorship:

1. **CORE-A (route-level)** — the default `GET /api/session` full reload and `_merged_session_messages_for_display` still passed only the watermark, so a normal full reload fell back to the old heuristic and resurrected deleted suffix turns. Threaded `truncation_boundary` through every merge call site + added an end-to-end route regression.
2. **Empty-sidecar guard (Opus)** — `at_or_after` was kept unconditionally, which is only valid once the watermark is *advanced past* the boundary; in the just-truncated (`boundary == watermark`) or legacy (`boundary is None`) state it resurrected the suffix. Replaced the backward-scan with a provable boundary guard.
3. **Checkpoint gate (Codex)** — the non-empty-sidecar path then needed care so it (a) keeps a state-only post-edit reply when the sidecar tail equals the watermark, and (b) doesn't resurrect a stale `ts > watermark` row appearing before the edited checkpoint. Gated the advanced-bypass on checkpoint consumption.

Gate (all three, clean): **Codex SAFE TO SHIP** (8 data-loss reproductions all converged), **Opus** must-fixes applied, **full suite 10392 passed**. Deployed to prod, verified `/api/sessions` healthy on v0.51.628.

Really solid root-cause direction on a gnarly append-only-merge bug — appreciate the iteration. The new regression tests pin every edge so this class can't silently come back.

PR #4774, opened 2026-06-23
perf: cache app-shell template instead of re-reading index.html per request

boriken72:
## Thinking Path

- Hermes WebUI aims for a fast, responsive browser experience backed by a deliberately simple Python + vanilla JS server.
- The `/`, `/index.html`, and `/session/<id>` routes serve the app shell and are the hottest navigations in the app.
- Each of those requests re-read the full `static/index.html` (~190 KB) from disk and re-ran two **process-constant** token substitutions (`__WEBUI_VERSION__`, `__MAX_UPLOAD_BYTES__`) every single time, plus `inject_extension_tags()` (which itself reads extension state + manifest from disk).
- Static assets in this same file already use an in-process cache (`_STATIC_CACHE`) keyed by `(size, mtime_ns)` — but the hottest, largest route bypassed it.
- This PR caches the partially-rendered shell template using that same pattern, while keeping the genuinely per-request bits (CSRF token, extension tags) per-request.
- The result is a faster shell render with byte-identical output and no behavior change.

## What Changed

- Added `_render_index_shell_base()` + `_INDEX_SHELL_CACHE` / `_INDEX_SHELL_CACHE_LOCK` in `api/routes.py`, directly mirroring the existing `_STATIC_CACHE` shape (cache keyed by `(st_size, st_mtime_ns)`; invalidates automatically on redeploy without a restart).
- Rewired the `/`, `/index.html`, `/session/<id>` branch of `handle_get()` to render from the cached base, then apply only the per-session CSRF token substitution and `inject_extension_tags()` per request.
- Added `tests/test_index_shell_template_cache.py` (5 tests).
- Added a `CHANGELOG.md` `[Unreleased] → Changed` entry.

Net diff: `api/routes.py` +47/−8, one new test file, one changelog line.

## Why It Matters

- The app-shell routes are the most-hit navigations; removing a 190 KB disk read + full re-render + extension-manifest disk reads from every one of them cuts redundant work on the critical path.
- Local microbenchmark (500 renders, warm cache): **~122 ms → ~23 ms (~5.2×)** for the template render portion, and it also eliminates the per-request extension-config/manifest disk I/O.
- Caching follows the precedent already established by `_STATIC_CACHE` in the same module, so it adds no new pattern, dependency, or build step.

## Verification

- `./scripts/test.sh tests/test_index_shell_template_cache.py` → 5 passed.
- Broader slice (`tests/test_conftest_server_boot.py tests/test_security_redaction.py tests/test_plugins_panel.py tests/test_sidebar_tab_visibility.py` + the new file) → all passed (107 in one run, 73 in another).
- **Byte-identical check:** reproduced the old inline render and asserted `old == new` for a fixed CSRF token — identical output (189,428 bytes).
- **CSRF still per-request:** two renders with different tokens differ and each contains its own token.
- **Cache invalidation:** touching the file's mtime produces a new `(size, mtime_ns)` signature and a fresh render, so a redeploy is picked up without a process restart.
- ruff (`E9,F,B`) is clean on all changed lines.

## Risks / Follow-ups

- The cache holds a single base string (one shell template) keyed by `"base"`; memory cost is one copy of `index.html`.
- Invalidation relies on `(size, mtime_ns)`, identical to `_STATIC_CACHE`. If a deploy somehow preserved both size and mtime exactly, the cache wouldn't refresh — same theoretical caveat the existing static cache already accepts.
- No change to CSRF handling, extension injection, or the `_serve_shell_unavailable` fallback path.

## Model Used

- Provider: Anthropic (via Nous)
- Model: claude-opus-4.8
- Tool use: agentic code exploration, edit, and local test execution via Hermes Agent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- **The app shell (`/`, `/index.html`, `/session/<id>`) is now served from an in-process template cache instead of re-reading and re-rendering `static/index.html` (~190 KB) on every navigation.** These are the hottest routes; each request previously re-read the file from disk and re-applied two process-constant token substitutions (`__WEBUI_VERSION__`, `__MAX_UPLOAD_BYTES__`). The partially rendered template is now cached and invalidated on `(size, mtime_ns)` change — mirroring the existing static-asset cache — so a redeploy is still picked up without a restart. The per-session CSRF token and runtime extension tags are still applied per request, so output is byte-identical to before. Measured ~5× faster shell render in a local microbenchmark and removes per-request extension-manifest disk reads.
+
 ## [v0.51.602] — 2026-06-23 — Release VI (sidebar fetches only the active source bucket)
 
 ### Changed
--- a/api/routes.py
+++ b/api/routes.py
@@ -8849,6 +8849,48 @@ def _save_saved_prompts(prompts: list) -> None:
     p.write_text(json.dumps(prompts, ensure_ascii=False, indent=2), encoding="utf-8")
 
 
+# In-process cache for the app-shell template. The `/`, `/index.html`, and
+# `/session/<id>` routes are the hottest navigations and each re-read the
+# ~190 KB static/index.html from disk and re-ran the two process-constant
+# substitutions (__WEBUI_VERSION__, __MAX_UPLOAD_BYTES__) on every request.
+# Those values are fixed for the process lifetime, so we cache the partially
+# rendered template here, keyed by (size, nanosecond mtime) exactly like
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Shipped in v0.51.603 (Release VJ) — thank you @boriken72. Clean app-shell template cache; Codex confirmed output stays byte-identical (CSRF token + extension tags applied per-request against the cached base, not baked in), (size,mtime_ns) invalidation busts on redeploy. Full suite green, app shell verified serving live in prod. One small maintainer test-fix: your change relocated the quote(WEBUI_VERSION) substitution into _render_index_shell_base(), so I repointed the asset-version-encoding regression test at the helper (the encoding itself is intact).

PR #4777, opened 2026-06-23
[Live Stream] Dedupe Compact Worklog process prose

franksong2702:
## Thinking Path

- Live streaming should show one coherent Compact Worklog while the response is in flight.
- Some model/provider paths emit the same user-visible progress sentence through both reasoning/interim events and the Anchor activity scene.
- That made the live view show duplicated process prose, even though a later hard refresh could make the duplicate disappear.
- The fix keeps the Anchor scene as the visible owner and suppresses duplicate live source rows before they leak into the transcript view.

## What Changed

- Remove matching live reasoning echo rows from the Compact Worklog DOM when interim assistant prose already owns the visible progress text.
- Skip writing a duplicate Anchor process prose row for `interim_assistant` flushes that immediately replay through the Anchor scene.
- Mark legacy worklog-source assistant segments as `hidden` as well as `aria-hidden`, so they cannot remain visible beside the Worklog process row.
- Update the interim-collapse structure test so it tracks the visible-interim flush path after the new `skipAnchorProcessProse` option.

## Why It Matters

- Users should not see the same live progress sentence as a Thinking card plus one or more outside process prose blocks.
- The live stream and post-refresh views should converge on the same visible Compact Worklog structure.
- This preserves settled history: exact duplicate live prose is deduped only on the live scene path, while settled rows keep their chronological records.

## Verification

- `node --check static/ui.js`
- `./scripts/test.sh tests/test_live_anchor_progress_echo.py tests/test_live_to_final_anchor_visible_order.py tests/test_issue2403_interim_collapse.py`
- GitHub Actions on current head `4c8e88ab2`: `browser-smoke`, `lint`, and the Python 3.11 / 3.12 / 3.13 test matrix passed.

## Before / After Evidence

- Before: a live streaming turn could expose the same process sentence through both the visible assistant segment and the Compact Worklog/Thinking path; switching away or hard-refreshing could make the duplicate disappear because settled rendering rebuilt the scene differently.
- After: duplicate live process prose is suppressed in the live renderer itself. The regression tests assert the DOM echo cleanup, the `interim_assistant` flush ordering, and live-only process-prose deduplication while preserving settled duplicate rows.
- No static screenshot is attached because this bug exists only during transient SSE streaming; the relevant evidence is the live DOM/render-path test coverage above.

## Risks / Follow-ups

- The live prose dedupe is intentionally exact after whitespace normalization. Near-duplicate but semantically similar progress text will still render as separate rows.
- This PR does not change live Worklog scroll behavior; that is handled separately in #4778.

## Model Used

- Provider: OpenAI
- Notable tool use: local repo inspection, `node --check`, `./scripts/test.sh`, GitHub CLI / GitHub REST API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Compact Worklog no longer shows duplicate live process prose while a response is still streaming.** When the same progress sentence arrives through both interim assistant text and Anchor activity rows, the live scene now keeps one visible Process row, removes matching live reasoning echoes from the Worklog DOM, and hides legacy source assistant segments so they cannot appear as duplicate transcript prose.
+
 ## [v0.51.603] — 2026-06-23 — Release VJ (cache the app-shell template)
 
 ### Changed
--- a/static/messages.js
+++ b/static/messages.js
@@ -2831,6 +2831,34 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
     }
     return false;
   }
+  function _removeLiveReasoningEchoRows(visible){
+    const turn=$('liveAssistantTurn');
+    const blocks=turn&&typeof _assistantTurnBlocks==='function'?_assistantTurnBlocks(turn):null;
+    if(!blocks||!visible) return false;
+    let removed=false;
+    const selector=[
+      '.agent-activity-thinking[data-anchor-scene-row="1"]',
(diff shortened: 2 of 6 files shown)


nesquena-hermes:
Shipped in **v0.51.604** 🚀 — thanks @franksong2702. The live Compact Worklog now keeps a single visible Process row and suppresses the duplicate reasoning-echo prose during streaming. Clean, well-tested PR; gated through Codex + Opus + full suite (10207 passed).

PR #4778, opened 2026-06-23
[Live Stream] Guard Compact Worklog scroll rebuilds

franksong2702:
## Thinking Path

- Live streaming should let pinned users follow the tail while preserving a mid-reader's current viewport.
- The Anchor-backed Compact Worklog renderer removes and rebuilds live scene rows on each activity update.
- During that remove/rebuild window, the transcript height can temporarily collapse, causing the browser to clamp `scrollTop` back near the user prompt before replacement rows are inserted.
- The fix guards the rebuild window so the scroll container never loses enough height to force that clamp.

## What Changed

- Add a live Anchor scroll rebuild guard that detects when the reader is away from the bottom.
- Temporarily hold `#msgInner` at the pre-rebuild scroll height while live Worklog rows are removed and reinserted.
- Force the captured snapshot into the unpinned path for mid-reader views, so `scrollIfPinned()` cannot pull the user back to the stream tail.
- Cover repeated/nested guard calls so rapid live ticks cannot leave a stale temporary `min-height` behind.

## Why It Matters

- Users reading live Worklog output should not periodically jump back to the original user message.
- The previous same-frame restore protected the end of the rebuild, but not the browser's intermediate scroll clamping while the DOM was briefly short.
- Pinned-at-bottom behavior remains intact: the guard only activates when the reader is away from the bottom.

## Verification

- `node --check static/ui.js`
- `./scripts/test.sh tests/test_issue4295_midstream_scroll_anchor.py tests/test_issue4295_scroll_pin_reentry.py tests/test_issue3479_ios_stream_scroll_jump.py tests/test_issue4720_done_scroll_jump_first_message.py`
- Browser synthetic verification on local 8787: rebuilt a tall live Anchor Worklog while scrolled mid-turn; `scrollTop` stayed at `1957` before/after rebuild and temporary `minHeight` released back to empty.
- GitHub Actions on current head `1269289d0`: `browser-smoke`, `lint`, and the Python 3.11 / 3.12 / 3.13 test matrix passed.

## Before / After Evidence

- Before: while a live Compact Worklog was streaming, a rebuild could briefly collapse transcript height; the browser then clamped the viewport back near `msg-user-0`, matching the observed periodic jump to the user's original prompt.
- After: the live rebuild holds transcript height stable until replacement rows land, restores the captured snapshot, then releases the temporary height guard.
- The local browser synthetic check measured the before/after invariant directly: `scrollTop` remained `1957` through the rebuild and the guard cleaned itself up.

## Risks / Follow-ups

- The guard intentionally targets the live Anchor Compact Worklog rebuild path only. Other transcript rebuild paths still rely on their existing scroll snapshot logic.
- The threshold follows the existing reader-away-from-bottom behavior; if that threshold changes globally, this helper should be reviewed with it.

## Model Used

- Provider: OpenAI
- Notable tool use: local repo inspection, `node --check`, `./scripts/test.sh`, Playwright/browser synthetic verification, GitHub CLI / GitHub REST API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Compact Worklog no longer snaps back to the user prompt during live-stream rebuilds.** Follow-up to the mid-stream scroll restore: the live Anchor scene now holds the transcript's scroll height stable while it removes and rebuilds Worklog rows, so the browser cannot temporarily clamp an unpinned reader's `scrollTop` to the top before the replacement rows land. Pinned-at-bottom streaming still follows the latest output.
+
 ## [v0.51.603] — 2026-06-23 — Release VJ (cache the app-shell template)
 
 ### Changed
--- a/static/ui.js
+++ b/static/ui.js
@@ -9571,6 +9571,39 @@ function _projectLiveAnchorActivitySceneForStream(streamId, mode){
     return null;
   }
 }
+function _prepareLiveAnchorScrollRebuildGuard(scrollSnapshot){
+  const messagesEl=$('messages');
+  if(!messagesEl||!scrollSnapshot) return {readerAwayFromBottom:false,release:null};
+  const beforeBottomDistance=Math.max(0,messagesEl.scrollHeight-messagesEl.scrollTop-messagesEl.clientHeight);
+  const readerAwayFromBottom=beforeBottomDistance>250&&(_messageUserUnpinned||messagesEl.scrollTop>0);
+  if(!readerAwayFromBottom) return {readerAwayFromBottom:false,release:null};
+  scrollSnapshot.pinned=false;
+  scrollSnapshot.userUnpinned=true;
+  scrollSnapshot.bottom=beforeBottomDistance;
+  _messageUserUnpinned=true;
+  _scrollPinned=false;
+  _nearBottomCount=0;
+  const msgInner=$('msgInner');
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Shipped in **v0.51.604** 🚀 — thanks @franksong2702.

The live Compact Worklog scroll-rebuild guard is now in prod. One maintainer adjustment was applied on top of your branch during the release gate:

- The guard predicate originally keyed on a raw `messagesEl.scrollTop > 0`, which the Codex regression gate reproduced as misclassifying a **pinned follower** as unpinned: during a large live render the assistant body grows before the anchor scene re-renders, so a pinned reader is transiently at `bottomDistance > 250` and would get flipped to `_messageUserUnpinned=true` / `_scrollPinned=false`, losing auto-follow. Changed the predicate to require an explicit non-follow signal (`_messageUserUnpinned || _scrollPinned === false`), updated your guard test to a genuine unpinned reader, and added a pinned-large-growth regression.
- Opus also caught a narrow deferred-`requestAnimationFrame` race (if the reader re-pins between guard-engage and the rAF firing, the stale re-restore would yank them back) — gated the re-restore on `_messageUserUnpinned` at rAF time.

Full suite green (10207 passed). Your core fix (holding the transcript height stable during the rebuild) is exactly right and shipped intact — appreciate the clean, well-tested PR. Closed as merged-with-fixes (the gate edits changed the commit so GitHub couldn't auto-detect the merge).

PR #3343, opened 2026-06-01
Sticky manual unpin during streaming chat scroll

pamnard:
## Summary
- Replaces the #3250 upward-intent timeout (450ms→2000ms) with sticky manual unpin: once the user scrolls up, streaming no longer auto-follows until they scroll back to the bottom or click ↓.
- `scrollIfPinned()` and delayed bottom settling bail out while `_messageUserUnpinned` is set.
- Re-follow requires explicit downward scroll into the near-bottom zone (250px hysteresis) or `scrollToBottom()`.
- Fresh streams reset follow state via `_resetStreamScrollFollow()` on non-reconnect `attachLiveStream`.

## Thinking Path
1. User reported snap-back while reading earlier content during long streaming responses.
2. #3250 only widened a wheel-intent timeout; scrollbar pauses and layout re-pin inside 250px still failed.
3. Match Memory/Notes-style UX: manual up = stay put; manual down to tail or ↓ = follow again.

## Test plan
- [x] `pytest tests/test_issue3250_upward_scroll_intent_window.py tests/test_issue1731_upward_scroll_unpins.py tests/test_issue677.py tests/test_issue1360_streaming_scroll_hardening.py tests/test_issue1690_scroll_completion.py tests/test_tars_scroll_reset_regressions.py`
- [ ] Start a long streaming reply, scroll up (wheel or scrollbar), pause 5s while tools/tokens arrive → viewport stays put
- [ ] Scroll back to bottom or click ↓ → live output follows again

Issue #1731, opened 2026-05-06
bug(streaming): small upward scrolls during streaming are overridden until user escapes near-bottom threshold


Michaelyklam:
## Bug Description

Reporter: Cygnus (Discord; please ping in replies).

While an assistant response is streaming, the message viewport remains pinned to the bottom unless the user scrolls upward several times. A small/normal upward scroll near the bottom is immediately overridden by the next streamed text update.

Cygnus described it as:

> It's still a fight to scroll up while the AI is printing. I can do it if I scroll upward like 3 or 4 times with determination lol, eventually it lets me, but it's annoying every time. I only want it to scroll with the printing text if I'm all the way at the bottom.

## Validation

Validated against current `origin/master` (`d8cd556`, checked out in a clean worktree) on 2026-05-05.

The current scroll pinning logic in `static/ui.js` treats the user as still pinned when the viewport is within 250px of the bottom:

```js
const nearBottom = el.scrollHeight - el.scrollTop - el.clientHeight < 250;
_nearBottomCount = nearBottom ? _nearBottomCount + 1 : 0;
_scrollPinned = _nearBottomCount >= 2;
```

Because `scrollIfPinned()` only checks `_scrollPinned`, an upward scroll that lands inside that 250px zone gets snapped back to the bottom by the next streamed token. This matches the observed “scroll upward 3 or 4 times” behavior: the user has to escape the 250px near-bottom hysteresis zone before auto-scroll finally stops.

Browser-console reproduction on a temporary local WebUI server:

```js
// Seed a tall transcript, scroll to bottom, then simulate streamed tokens while
// repeatedly attempting a small upward scroll.
const el = document.getElementById('messages');
const inner = document.getElementById('msgInner');
inner.innerHTML = '';
for (let i = 0; i < 80; i++) {
  const d = document.createElement('div');
  d.textContent = 'history line ' + i;
  d.style.padding = '12px';
  inner.appendChild(d);
}
const live = document.createElement('div');
live.id = 'live-test';
live.style.padding = '12px';
inner.appendChild(live);
scrollToBottom();

for (let t = 0; t < 12; t++) {
  el.scrollTop = Math.max(0, el.scrollTop - 120); // user scrolls up
  el.dispatchEvent(new Event('scroll', { bubbles: true }));
  await new Promise(r => requestAnimationFrame(r));
  live.textContent += ' token' + t;
  scrollIfPinned(); // next streaming update
}
```

[...]

Issue #1360, opened 2026-04-30
bug(streaming): auto-scroll still overrides user scroll position on macOS app — #677 regression / edge case


nesquena-hermes:
## Bug report

Auto-scroll still overrides the user's scroll position during streaming on the **macOS app (latest)**, despite the fixes shipped in #677 (April 18). Nathan reports this is his single biggest day-to-day pain point: while a long response is streaming, scrolling up to read earlier content gets snapped back to the bottom, and you can't read up until the response finishes.

Source: AvidFuturist, April 30 2026.

## Verification — #677 fixes are live in master (v0.50.248)

The three fixes from #677 are present in `static/ui.js`:

- `_scrollPinned` flag with 150px unpin threshold — `static/ui.js:828–845`
- `renderMessages()` guards force-scroll behind `S.activeStreamId` — `static/ui.js:3401–3405` and the cache path at `:2985`
- Scroll-to-bottom button toggled by the scroll listener — `static/index.html:210`, `static/ui.js:838–839`

So the obvious regressions are ruled out. The bug is something else.

## Reproduction (macOS app, v1.4.0 + WebUI v0.50.248)

1. Open a long-running session, send a prompt that produces a long response (or kicks off many tool calls)
2. While streaming, scroll up to re-read earlier content
3. Observe: the view snaps back to the bottom within ~100–500ms

### A. WKWebView scroll-event timing under momentum

On macOS WKWebView (the Swift Mac app's container), `scroll` events can fire **after** momentum has already carried the scrollTop past the listener's snapshot of `scrollHeight - scrollTop - clientHeight`. The 150px nearBottom check at `ui.js:836` then re-asserts `_scrollPinned = true` while the user is mid-flick, so the next `scrollIfPinned()` call (from the smd token feed in `messages.js`) yanks them back down. Browser Safari may behave the same; needs testing.

**How to verify:** instrument `ui.js:835–844` with a console.log of `{scrollTop, scrollHeight, clientHeight, nearBottom, _scrollPinned}` and see whether `_scrollPinned` is flipping back to `true` mid-streaming when the user *thinks* they're scrolled up.

### B. Browser scroll anchoring during tool-card insertions

[...]

PR #4779, opened 2026-06-23
fix: force scroll-to-bottom on session switch (ignore stale unpin flag)

jcarvine:
## Problem

On mobile, switching conversations does not scroll to the latest message — the user has to manually scroll down every time.

## Root cause

`_messageUserUnpinned` flag stays true after a mobile touch event during session loading. Since `_scrollAfterMessageRender` unconditionally honors this flag and returns early, the scroll-to-bottom logic is skipped even on fresh session loads.

## Fix

Only honor the unpin flag when `preserveScroll` is true (same-session refresh). For new session loads (`preserveScroll` is falsy), force `scrollToBottom()` instead.

## Changes

- `static/ui.js`: `_scrollAfterMessageRender` — add `&& preserveScroll` guard to the unpin check

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -10985,7 +10985,13 @@ function _scrollAfterMessageRender(preserveScroll, scrollSnapshot){
   // lands at the bottom as expected. (Codex #4006 follow-up.)
   // renderMessages() captures the pre-wipe snapshot for this case too (see its
   // scrollSnapshot init), so restoring here lands the reader where they were.
-  if(_messageUserUnpinned){
+  //
+  // Guard: only honor unpin when preserveScroll is true (same-session refresh
+  // where we're deliberately keeping the user's position). For fresh session
+  // loads (preserveScroll=false), the unpin flag may be stale from a mobile
+  // touch event during the loading transition — forcing scrollToBottom() is the
+  // correct behavior for a user-initiated navigation to a different session.
+  if(_messageUserUnpinned && preserveScroll){
     _restoreMessageScrollSnapshot(scrollSnapshot);
     _maybeShowNewMessageScrollCue(scrollSnapshot);
     return;


nesquena-hermes (changes requested):
Thanks for digging into the stale-unpin-on-session-switch behavior — the underlying concern (a touch event during the load transition can leave `_messageUserUnpinned` stale) is real. But this specific change can't ship as written: the new guard is **dead code**, and its net effect is a behavioral **regression**, not the intended fix.

## 1. The `&& preserveScroll` guard is unreachable (dead code)

The change is:

```js
- if(_messageUserUnpinned){
+ if(_messageUserUnpinned && preserveScroll){
```

inside `_scrollAfterMessageRender(preserveScroll, scrollSnapshot)` (≈ui.js:10994).

But the very top of that function already returns whenever `preserveScroll` is true:

```js
if(preserveScroll){
  ...
  return;   // ui.js ~10973 / ~10976 — ALWAYS returns on the preserveScroll path
}
```

So by the time control reaches your edited line, `preserveScroll` is **always `false`**. That makes `_messageUserUnpinned && preserveScroll` ≡ `_messageUserUnpinned && false` ≡ **always false**. The guarded block (`_restoreMessageScrollSnapshot` + `_maybeShowNewMessageScrollCue` + `return`) is now **unreachable**, and control always falls through to `scrollToBottom()`.

## 2. Net effect is a regression of the Codex #4006 protection

Because that block never runs anymore, you've effectively deleted the documented sticky-unpin behavior for **every** non-preserve re-render — not just session switches. The original `if(_messageUserUnpinned)` block exists specifically to keep an intentionally-scrolled-up reader in place during idle/non-preserve re-renders (see the `Codex #4006 follow-up` comment right above it).

The non-preserve path is hit by many mid-session re-renders, not only session switches — plain `renderMessages()` (preserveScroll=false) is called from stream/echo paths in messages.js, ~25 sites in commands.js, edit/regenerate in ui.js, and sessions.js. Concrete regression: a user scrolls up to read history mid-session (unpinned) → a stream completion or command insert fires `renderMessages()` → **before:** their position is restored + a "new message" cue shows; **with this PR:** `scrollToBottom()` yanks them to the bottom.

## 3. The session-switch case is already handled at the source

On a real switch (`currentSid !== sid`), sessions.js (≈line 1218) already calls `window._resetScrollDirectionTracker()`, which sets `_messageUserUnpinned=false; _scrollPinned=true;` and clears the rest of the transient scroll state (`#1731`/`#3343` sticky-unpin). So scroll-to-bottom on switch is the existing, intended behavior. If a touch during the async load window re-sets the flag *after* that reset, the right fix is to re-clear it at the source on a confirmed switch, not to neuter the shared render-path guard.

## What I'd suggest

- Don't modify the shared `_scrollAfterMessageRender` non-preserve branch — restore the `if(_messageUserUnpinned){ restore + cue; return; }` block (the #4006 behavior).
- If a stale flag genuinely survives the load window, fix it in `loadSession()` on a confirmed switch (`currentSid !== sid`) **immediately before the final render** (after `await _ensureMessagesLoaded(...)`), so the late reset can't be clobbered by a touch during the gap.
- Add a test for the matrix this area lacks (there's currently zero coverage referencing `_scrollAfterMessageRender` / `_messageUserUnpinned`): (a) switch-while-unpinned → lands at bottom; (b) mid-session non-preserve re-render while scrolled up → reader stays put + sees the cue; (c) same-session preserve refresh while unpinned → position preserved.

## Heads-up on the sibling PR

You also have **#4780** ("reset scroll state on session switch" in sessions.js) targeting the same bug. That one is in the architecturally correct layer (it knows it's a session switch) but is largely redundant with the existing `_resetScrollDirectionTracker()` reset at sessions.js:1218 — i.e. the bug may already be handled in master. It'd help reviewers if you consolidated to **one** PR with a concrete repro showing what the existing reset misses, rather than two competing approaches. The maintainer will make the call on which (if either) is needed.

nesquena-hermes:
Thanks @jcarvine — good catch on the root cause (stale `_messageUserUnpinned` from a touch event during load). I reviewed both this and your follow-up **#4780**.

This version (the `_messageUserUnpinned && preserveScroll` guard in `_scrollAfterMessageRender`) has a regression: it makes idle, non-`preserveScroll` re-renders of an intentionally scrolled-up reader fall through to `scrollToBottom()`, which breaks the deliberately-tested sticky-unpin invariant from #1360 (`test_idle_render_preserves_manual_unpin_until_explicit_bottom`) — that's the red CI shard here. `preserveScroll` isn't the right signal to distinguish a genuine session switch from an idle re-render.

**#4780 is the better fix** — resetting `_messageUserUnpinned`/`_scrollPinned` at the start of `loadSession()` gated on `currentSid !== sid` targets exactly the session-switch case without touching the idle-rerender path, and its CI is green. I'd suggest we converge on #4780 and close this one as superseded by it — but I'll leave that to you. Either way, #4780 is the one I'll take through review.

nesquena-hermes:
Superseded by #4780 (your own rework, shipped in v0.51.604). #4780's approach — resetting scroll state at loadSession() on a genuine session switch — is the correct one; this version's guard broke the #1360 sticky-unpin invariant. Thanks for the quick self-correction!

PR #4780, opened 2026-06-23
fix: reset scroll state on session switch to prevent stale unpin blocking scroll-to-bottom

jcarvine:
On mobile, switching sessions leaves a stale `_messageUserUnpinned` flag from stray touch events during loading, blocking scroll-to-bottom.

This fix resets `_messageUserUnpinned=false` and `_scrollPinned=true` at the start of `loadSession()` when navigating to a different session, before any async ops that could be interleaved with touch events.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -1002,6 +1002,13 @@ async function loadSession(sid){
   // Mark this session as the in-flight load. Subsequent loadSession() calls
   // will overwrite this; stale awaits use the mismatch to bail out (#1060).
   _loadingSessionId = sid;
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Shipped in **v0.51.604** 🚀 — thanks @jcarvine. Switching conversations on mobile now lands at the latest message. Nice self-correction from #4779 — resetting the scroll state at `loadSession()` gated on a genuine session switch is the right approach (it leaves the #1360 sticky-unpin invariant intact). #4779 can be closed as superseded by this.

PR #4781, opened 2026-06-23
[Live Stream] Allow live Compact Worklog collapse

franksong2702:
## Thinking Path

- Live Compact Worklog is meant to be an inspectable activity disclosure, not a fixed wall of internals.
- During streaming, the `Processed …` summary already visually looks like a disclosure row, but the code disabled it and hid the caret.
- Users who are distracted by active tool/thinking rows should be able to collapse them without waiting for settle/final.
- The narrow fix is to make the live summary use the same disclosure behavior while preserving existing finalize behavior.

## What Changed

- Removed the live-current guard in `_toggleActivityGroup()` so the current streaming Worklog group can collapse and expand.
- Stopped marking live Worklog summaries as disabled/static in `ensureActivityGroup()`.
- Removed CSS that hid the live Worklog chevron and forced static cursor treatment.
- Added a focused Node regression that toggles a current live Worklog group and verifies `aria-expanded`, disclosure persistence, and live toggle state update.
- Added a changelog note for the user-visible interaction change.

## Why It Matters

- The `Processed …` row now honestly communicates that it is interactive.
- Users can collapse noisy live Worklog details during long streaming turns while keeping the elapsed timer visible.
- This addresses the Discord feedback that active incremental Worklog output can be distracting even when Compact Worklog is enabled.

## Verification

- `node --check static/ui.js`
- `git diff --check -- CHANGELOG.md static/ui.js static/style.css tests/test_live_to_final_anchor_visible_order.py`
- `./scripts/test.sh tests/test_live_to_final_anchor_visible_order.py` (`54 passed`)

## Before / After Evidence

- Before: live `Processed …` summaries were disabled (`disabled=true`, `aria-disabled=true`), carried `data-live-summary-static`, and CSS hid the chevron for the current live group.
- After: the live summary keeps the normal disclosure affordance; toggling a current live group collapses it, persists the disclosure state, updates `aria-expanded`, and keeps the timer on the summary row.
- The regression test directly executes the live disclosure handler so this does not rely only on static source assertions.

## Risks / Follow-ups

- This only changes the current live Worklog summary. Tool-card inner disclosures and settled finalize behavior are unchanged.
- If future design wants live Worklog collapsed by default instead of merely collapsible, that should be a separate settings/policy PR.

## Model Used

- Provider: OpenAI
- Notable tool use: local repo inspection, `node --check`, `./scripts/test.sh`, GitHub CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- **Live Compact Worklog summaries can now be expanded or collapsed while a response is still streaming.** The `Processed …` row keeps its timer and now behaves like the settled Worklog disclosure, so users can hide noisy live work details without waiting for the final answer.
+
 ## [v0.51.603] — 2026-06-23 — Release VJ (cache the app-shell template)
 
 ### Changed
--- a/static/style.css
+++ b/static/style.css
@@ -3025,14 +3025,7 @@ body.resizing .sidebar{transition:none!important;}
 .tool-worklog-group[data-tool-worklog-group="1"]:not([data-run-activity-group="1"]) .tool-worklog-summary:hover{
   background:transparent;
 }
-.tool-worklog-group[data-live-tool-call-group="1"] .tool-worklog-summary[data-live-summary-static="1"]{
-  cursor:default;
-  opacity:1;
-}
 .tool-worklog-group[data-tool-worklog-group="1"]:not([data-run-activity-group="1"]) .as-dot{
-  display:none;
-}
-.tool-worklog-group[data-live-tool-call-group="1"][data-live-activity-current="1"] .tool-call-group-chevron{
   display:none;
 }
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Thanks @franksong2702 — holding this one for a quick visual sign-off from @nesquena before merge.

The sibling fixes #4778 (scroll-rebuild guard) and #4777 (process-prose dedup) just shipped in **v0.51.604**. This PR is slightly different: allowing the live Compact Worklog to *collapse* during streaming is an interaction/behavior change (not a pure correctness fix), so it gets an eyeball on the live behavior before it ships — that's our process for visible-behavior changes on the chat surface.

It's cleanly built (net code reduction, CI green, fresh-cut from master) and queued for the visual review + the standard Codex/Opus/full-suite gate. Nothing needed from you right now.

nesquena-hermes:
Shipped in **v0.51.609** 🚀 — thanks @franksong2702. The live Compact Worklog can now be collapsed/expanded mid-stream. Codex SAFE (reproduced that a user-collapsed live group persists across live rebuilds) + full suite 10282 passed. Closed as merged-with-resolution.

Issue #4766, opened 2026-06-23
perf: /api/sessions ships 465 rows (342KB) when the sidebar displays 48


rodboev:
## The problem

`/api/sessions` returns every session the profile owns, regardless of how many the sidebar will actually display. On a profile with 465 sessions, the response is 342KB, the frontend renders 48 rows, and the other 417 are discarded by client-side filters. The server already knows every filter parameter needed to skip those rows, but it builds and serializes them anyway.

This is the dominant latency contributor on a cold profile switch: `_build_session_list_cache_payload()` takes ~720ms to enumerate, merge, probe, and serialize 465 sessions. On a warm cache hit it's 15ms, but the cache is profile-keyed with a 2.5s TTL, so the first switch to a profile (or the first call after 2.5s idle) always pays the full cost.

## Where the 417 discarded rows come from

Measured on a real default profile (465 total sessions, `show_cli_sessions=false`, `show_cron_sessions=false`):

| Category | Count | Payload | % of total | Displayed? |
|---|---|---|---|---|
| Cron (`default_hidden=true`) | 200 | 160KB | 47% | No |
| CLI (imported sidecars) | 202 | 130KB | 38% | No |
| Zero-message drafts | 14 | 11KB | 3% | No |
| **Visible WebUI** | **48** | **40KB** | **12%** | **Yes** |

### Why each category survives server-side filtering

1. **Cron sessions (200, 160KB)**: `_hide_from_default_sidebar()` removes them, but `_include_project_hidden_background_sidebar_sessions()` adds them back with `default_hidden=True` so project chips can reveal them client-side. All 200 have a `project_id`.

2. **CLI imported sidecars (202, 130KB)**: When `show_cli_sessions=false`, the server skips `get_cli_sessions()` (no state.db query) and strips rows matching `_is_cli_session_for_settings()` from the WebUI session list. But WebUI-side imported sidecars of CLI sessions are not caught by this filter, because they're stored in the WebUI session index (not state.db) and their source flags don't always match `_is_cli_session_for_settings`.

3. **Zero-message drafts (14, 11KB)**: `all_sessions()` already strips `Untitled + 0-message` sessions (models.py ~L3444), but sessions with a custom title and 0 messages pass through. The frontend's `_sidebarRowHasVisibleMessages()` catches these, but the server doesn't apply an equivalent filter.

## Proposed fix: server-side sidebar rendering intent

[...]

PR #3408, opened 2026-06-02
fix(sidebar): keep active blank chat visible

AJV20:
## Thinking Path
- The sidebar intentionally filters inactive 0-message sessions, but that also hid the currently active New Chat before its first message.
- The fix keeps that filtering for inactive empty sessions while injecting only the active blank session row into the cached sidebar rows.
- New Chat also switches the source filter back to WebUI so opening it from a CLI-filtered sidebar does not immediately hide the active chat.

## What Changed
- Added a helper that prepends the active ephemeral session to sidebar render rows when the server-side list omits it.
- Ensured search/result merging still preserves the active row before the existing message-count filter runs.
- Added/updated static regression tests covering the active-row injection, search path, content-search merge contract, and CLI-to-WebUI source-filter reset.
- Added a changelog entry.

## Why It Matters
Blank chats should remain findable and selected in the sidebar as soon as the user opens them, instead of disappearing until the first message is sent.

## Verification
- `git diff --check`
- `node --check static/sessions.js`
- `python3 -m pytest tests/test_active_empty_session_sidebar.py -q`
- `python3 -m pytest tests/test_active_empty_session_sidebar.py tests/test_session_copy_link_action.py tests/test_issue2351_cli_session_source_filter.py tests/test_empty_session_no_disk_write.py tests/test_stale_empty_session_restore.py tests/test_streaming_session_sidebar.py -q`

## Risks / Follow-ups
- This is a presentation/cache-layer fix only; it does not make inactive 0-message sessions durable or write them to disk.
- No screenshots are attached because the change is sidebar row presence/selection behavior without visual styling changes.

## Contract Routing
Task type: small sidebar/session visibility bug fix.  
Touched areas: `static/sessions.js`, sidebar render cache, session source filter.  
Relevant public docs: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `docs/UIUX-GUIDE.md`, `DESIGN.md`, `docs/rfcs/webui-run-state-consistency-contract.md`.  
Scope boundaries: no storage/durability change; inactive 0-message sessions remain filtered.  
Evidence: focused static regression tests plus syntax/diff checks.

Issue #3019, opened 2026-05-27
Cron sessions hidden from sidebar when 'Cron Jobs' project chip selected


mysoul12138:
## Problem

Cron sessions never appear in the sidebar, even when the dedicated 'Cron Jobs' project chip is selected.

## Root Cause

pi/models.py ll_sessions() calls _hide_from_default_sidebar() (lines 2404/2442), which returns True for sessions with source_tag == 'cron' or sid.startswith('cron_'). These sessions are filtered out server-side before the /api/sessions response reaches the frontend.

The frontend's project filter (
enderSessionListFromCache() in sessions.js:3285-3288) operates on the cached _allSessions array, which never contains cron sessions. Clicking the 'Cron Jobs' project chip filters by project_id but finds nothing.

The WebUI/CLI source filter (sessions.js:3271-3273) executes before the project filter, but the data is already missing upstream at the server level.

## Key Code Paths

- pi/models.py:2156-2161 - _hide_from_default_sidebar() returns True for cron
- pi/models.py:2404/2442 - ll_sessions() applies the filter
- pi/routes.py:3922 - /api/sessions handler calls ll_sessions()
- static/sessions.js:3271-3288 - frontend source filter -> project filter chain

nesquena-hermes:
## Summary

Confirmed the diagnosis by reading `api/models.py:2156-2161` and `api/routes.py:4406-4444`. Cron sessions are filtered out server-side in `all_sessions()` before the response ever reaches `/api/sessions`, so the `Cron Jobs` project chip in the sidebar is dereferencing a project_id that no row will ever match. The frontend project filter at `sessions.js:3287-3288` operates on `_allSessions`, which is missing the cron rows entirely.

## Code reference

The server-side hide rule at `api/models.py:2156-2164`:

```python
def _hide_from_default_sidebar(session: dict) -> bool:
    """Return True for internal/background sessions hidden from the default list."""
    sid = str(session.get('session_id') or '')
    source = session.get('source_tag') or session.get('source')
    if source == 'cron' or sid.startswith('cron_'):
        return True
    if bool(session.get('pre_compression_snapshot')):
        return not bool(session.get('_show_pre_compression_snapshot'))
    return False
```

[...]

PR #4782, opened 2026-06-23
perf(sidebar): trim hidden and empty sidebar payload rows (#4775)

### LINKED ISSUE #4775: perf: filter default_hidden and zero-message sessions server-side in /api/sessions
## Context

Issue #4766 identified three categories of sessions shipped by `/api/sessions` that the sidebar discards client-side. PR #4769 (shipped in v0.51.602) addressed one: it added `?sidebar_source=webui|cli` so each source tab fetches only its bucket. That removed 2 CLI sessions from the webui response.

The other two categories, which account for 93% of the discarded rows, were not addressed:

| Category | Count | Payload | Status |
|---|---|---|---|
| Cron (`default_hidden=true`) | 200 | 160KB (47%) | **Still shipped, discarded client-side** |
| Zero-message drafts | 14 | 11KB (3%) | **Still shipped, discarded client-side** |
| CLI source bucket | 2 | ~1KB | Fixed by #4769 |

Current state after #4769: `/api/sessions?sidebar_source=webui` returns 463 sessions (333KB), cold cache miss ~740ms. The frontend displays ~263 after filtering out `default_hidden`, and ~48 after further filtering zero-message rows.

## What needs to happen

Add an `exclude_hidden=1` query parameter to `/api/sessions` that filters out `default_hidden=true` sessions server-side, inside `_build_session_list_cache_payload()`. The frontend should send this parameter on default sidebar loads, and omit it when expanding a project chip (which needs to reveal hidden cron sessions under that project).

Separately, apply the equivalent of the frontend's `_sidebarRowHasVisibleMessages()` predicate server-side: skip sessions with 0 messages that are not actively streaming, not marked for attention, and not the active session. This mirrors `static/sessions.js` around line 5342.

Both filters should be threaded through `_session_list_cache_key()` so filtered and unfiltered responses get distinct cache entries.

### Where to look

- `api/routes.py:9752` — GET `/api/sessions` handler, where `sidebar_source` is already parsed from query params. Add `exclude_hidden` parsing here.
- `api/routes.py:1598` — `_session_list_cache_key()`, add `exclude_hidden` to the key tuple.
- `api/routes.py:2100` — `_build_session_list_cache_payload()`, after the `sidebar_source` filtering block (around line 2110). Filter out `default_hidden=True` rows when `exclude_hidden` is set. Filter zero-message non-streaming sessions in the same pass.
- `static/sessions.js:3837` — where `sidebar_source` is added to the fetch URL. Add `exclude_hidden=1` here when the default sidebar is being rendered (not a project-chip expansion).
- `static/sessions.js:5342` — `_sidebarRowHasVisibleMessages()`, the predicate to mirror server-side.

### Expected result

- Response: 463 sessions (333KB) → ~48 sessions (~40KB)
- Cold cache miss: ~740ms → estimated <100ms
- Project-chip expansion omits the parameter, gets the full set (separate cache entry)
- The `sidebar_source` parameter from #4769 continues to work alongside this

### How to verify

```js
// Before: no exclude_hidden
const r1 = await fetch('/api/sessions?sidebar_source=webui');
const d1 = await r1.json();
console.log('Without filter:', d1.sessions.length, 'sessions');

// After: with exclude_hidden
const r2 = await fetch('/api/sessions?sidebar_source=webui&exclude_hidden=1');
const d2 = await r2.json();
console.log('With filter:', d2.sessions.length, 'sessions');
// Should show ~48 instead of ~463
```

Cross-ref: #4766 (original analysis), #4769 (sidebar_source filter), #4662 (profile-switch UX), #4748 (skeleton loading).

rodboev:
## Thinking Path

- [#4769](https://github.com/nesquena/hermes-webui/pull/4769) already pushed the WebUI versus CLI source-tab choice into `/api/sessions`, but the default sidebar path still ships rows the browser drops locally: rescued `default_hidden` rows and plain 0-message rows with no visible-sidebar signal.
- Named project chips depend on the `#3019` rescue that keeps those background rows in the payload as `default_hidden=True`, so the fix has to be request-shaped. Default and unassigned loads should be filtered, while named project-chip loads stay unfiltered.
- The route still does not know the active session id. Instead of turning a shared cache entry into a per-session key, keep the server-side zero-message filter aligned to the non-active clauses of `_sidebarRowHasVisibleMessages()` and let the shipped `#3408` client reinjection path continue owning the active blank chat.

## What Changed

- `static/sessions.js`: `_sessionListQueryString()` now adds `exclude_hidden=1` for the default and unassigned sidebar views, omits it for named project chips, routes project-chip clicks through `_setActiveProjectFilter()`, tags cached session scope with the hidden-filter mode, and reuses the same query builder for delete fallback refreshes.
- `api/routes.py`: `GET /api/sessions` now reads `exclude_hidden`, the session-list cache key and builder vary on both `exclude_hidden` and a route-owned `visible_only` flag, and the builder filters plain zero-message rows plus optionally `default_hidden` rows after the existing CLI cap and before counts and the final `sidebar_source` slice.
- `tests/test_4167_sidebar_payload_and_scope.py`: extends the existing cache-scope guard coverage so failed refresh fallback rejects rows fetched under the wrong hidden-filter mode.
- `tests/test_issue4766_sidebar_source_pushdown.py`: updates the existing `_sessionListQueryString()` expectations so the `#4766` request helper contract reflects the new `exclude_hidden` seam.
- `tests/test_issue4775_sidebar_hidden_zero_message_pushdown.py`: adds focused regression coverage for default-hidden pruning, plain 0-message pruning, visibility-signal exemptions, request-shape toggling for named versus unassigned project chips, and cache-key separation.
- Existing `/api/sessions` regression files that assert list-field behavior now create visible rows or realistic `message_count` metadata instead of relying on renamed empty sessions that the new default sidebar contract intentionally hides.
- Builds on [#4769](https://github.com/nesquena/hermes-webui/pull/4769), which introduced the current `sidebar_source` request seam and per-source cache split for `/api/sessions`.

## Why It Matters

The default sidebar should not spend bandwidth on rows it never renders. This trims those rows at the same route seam that already owns source-bucket filtering, keeps named project-chip reveal working, preserves the active blank chat through the existing `#3408` client path, and avoids a per-session cache split on a hot shared endpoint.

## Verification

```text
python -m pytest tests/test_4167_sidebar_payload_and_scope.py tests/test_issue4775_sidebar_hidden_zero_message_pushdown.py tests/test_session_action_menu_regression.py tests/test_issue4766_sidebar_source_pushdown.py
python -m pytest tests/test_issue3019_cron_project_sessions.py tests/test_active_empty_session_sidebar.py tests/test_sidebar_session_partition.py tests/test_session_list_long_history_perf.py
python -m pytest tests/test_sidebar_unassigned_filter.py
python -m pytest tests/test_gateway_sync.py::test_importing_older_gateway_session_preserves_original_timestamps_and_order tests/test_regressions.py::test_deleted_session_does_not_appear_in_list tests/test_session_import_cli_fallback_model.py::test_sessions_endpoint_suppresses_duplicate_webui_state_projection tests/test_sprint12.py::test_pinned_in_session_list tests/test_sprint15.py::test_session_project_in_list tests/test_session_summary_redaction.py::test_api_sessions_list_redacts_secret_titles tests/test_sprint23.py::test_session_compact_has_usage_fields
```

## Upstream

Closes #4775.

## Model Used

GPT 5.5 via Codex CLI


nesquena-hermes:
Shipped in **v0.51.605** 🚀 — thanks @rodboev. The `/api/sessions` sidebar payload now trims rescued `default_hidden` and 0-message rows that carry no visible signal, server-side before serialization.

Gated clean through Codex (SAFE TO SHIP — verified the trim predicate keeps messageful/attention/streaming/active-stream/pending rows + the #3019 project-chip rescue), Opus (SHIP — confirmed client/server drop parity + read all 14 test diffs, no weakened assertions), and the full suite (10214 passed).

One non-blocking follow-up worth an eventual glance (Opus): `_session_has_server_visible_messages` calls `_session_attention_summary` per 0-msg row on cache-miss builds — cheap and bounded, but worth a look under a profile with many empty sessions. This is part of the profile-perf cluster with #4741 + #4748.

Issue #4691, opened 2026-06-22
v0.17.0 async delegation silently drops results in WebUI sessions


nmphat:
## Summary

Subagent delegation via `delegate_task` in WebUI sessions silently drops results — the subagent completes its work, but the result never arrives in the parent session.

## Environment

- **hermes-agent**: v0.17.0 (v2026.6.19) — commit `2bd1977d8` (also reproduced on v0.16.0 with stale session)
- **hermes-webui**: latest master (`a30348b2`)
- **Platform**: WebUI (session ID `ff269d164823`)

## Steps to Reproduce

1. Open Hermes WebUI (new session on v0.17.0)
2. Send a message that triggers `delegate_task` (e.g., "Create file /tmp/test.txt with content 'hello', then cat it")
3. `delegate_task` dispatches — logs confirm: `Dispatched async delegation batch deleg_xxx (1 task(s), session_key=ff269d164823)`
4. Subagent runs to completion (`platform=subagent`)
5. **Result never arrives in WebUI session** — no error shown

## Expected Behavior

Subagent result should be injected back into the parent WebUI session and displayed to the user.

## Actual Behavior

- Subagent completes (file is created, verified via `cat`)
- Parent session receives nothing
- Gateway log shows: `Suppressing normal final send for session agent:main:telegram:dm:...: final delivery already confirmed`

## Root Cause

**hermes-agent v0.17.0's async delegation** (`tools.async_delegation`) dispatches subagents in background mode. The result delivery back to the parent WebUI session fails silently — the session key format or result injection path is incompatible between the async delegation system and WebUI's streaming/SSE architecture.

### Evidence from logs

```
# Dispatched with bare UUID session key
INFO tools.async_delegation: Dispatched async delegation batch deleg_04c1c3df (1 task(s), session_key=ff269d164823)

# Subagent completes
INFO agent.turn_context: session=20260622_155619_33c1eb platform=subagent
INFO run_agent: OpenAI client closed (agent_close)

# Result suppressed — never reaches WebUI
INFO gateway.run: Suppressing normal final send for session agent:main:telegram:dm:5537903840: final delivery already confirmed
```

### Key finding

- v0.16.0 (v2026.6.5) delegation works **synchronously** — parent blocks, result returns inline
- v0.17.0 (v2026.6.19) introduced **async/background delegation** — `delegate_task` returns immediately with `status: dispatched`, subagent runs in background, but result injection back to WebUI fails
[...]

PR #4799, opened 2026-06-23
Fix WebUI async delegation completion delivery

sysophelper-droid:
## Summary

This PR has been rebased on current `master` after the minimal #4912 / v0.51.652 WebUI fix landed.

#4912 already handles the core background wakeup formatting path by delegating `async_delegation` events to the agent-side `format_process_notification`. This PR is now narrowed to the remaining bridge contract beyond that minimal delivery fix:

- Use `delegation_id` as the stable WebUI delivery/dedupe id for `async_delegation` events.
- Keep terminal process completions keyed by their existing `session_id` process id.
- Route next-turn drain events by `session_key`, including non-identity `session_key -> session_id` mappings through `PROCESS_SESSION_INDEX`.
- Allow next-turn drain to format `async_delegation` events through the canonical agent formatter.
- Best-effort call `mark_async_delegation_consumed(delegation_id)` when newer Hermes Agent builds expose it, without hard-requiring the companion agent PR on older installations.
- Keep matched-but-unformattable events consumed rather than replaying them forever on later turns.

## What changed after #4912

Removed the broader fallback formatting work from this PR. WebUI now relies on the canonical agent formatter for async-delegation text, matching the #4912 direction.

The remaining delta is focused on dedupe/routing/recovery coordination:

- `api/process_event_utils.py`
  - `completion_delivery_id(evt)`
  - `mark_async_delegation_record_consumed(delegation_id)`
- `api/background_process.py`
  - background wakeup path uses `delegation_id` for async-delegation delivery ids
  - optional consumed marker is called after deferred/started WebUI delivery
- `api/streaming.py`
  - next-turn drain handles `async_delegation`
  - direct WebUI session match plus `PROCESS_SESSION_INDEX` fallback routing
  - shared completion consumed marker remains the cross-path dedupe key

## Validation

```text
python -m pytest tests/test_async_delegation_webui_bridge.py -q -o 'addopts='
3 passed

python -m pytest tests/test_background_process_wakeup_format.py -q -o 'addopts='
10 passed

python -m pytest \
  tests/test_async_delegation_webui_bridge.py \
  tests/test_background_process_wakeup_format.py \
  tests/test_notify_on_complete_webui.py \
  tests/test_background_tasks.py \
  tests/test_issue856_background_completion_unread.py \
  -q -o 'addopts=' -W "ignore:'audioop' is deprecated and slated for removal in Python 3.13:DeprecationWarning:discord.player"
45 passed

./scripts/test.sh tests/test_async_delegation_webui_bridge.py tests/test_background_process_wakeup_format.py tests/test_notify_on_complete_webui.py
17 passed

git diff --check OK
compileall OK
ruff on new/updated tests + helper OK
static added-lines scan clean
```

## Related

- Builds on the minimal WebUI delivery fix from #4912 / v0.51.652.
- Coordinates best-effort with NousResearch/hermes-agent#51577 when that optional consumed-marker API is available.
- Addresses the broader async delegation WebUI delivery/dedupe path for #4691.

### DIFF
--- a/api/background_process.py
+++ b/api/background_process.py
@@ -48,6 +48,11 @@ import time
 import uuid
 from typing import Any, Optional
 
+from api.process_event_utils import (
+    completion_delivery_id,
+    mark_async_delegation_record_consumed,
+)
+
 logger = logging.getLogger(__name__)
 
 _DRAIN_THREAD: Optional[threading.Thread] = None
@@ -441,9 +446,9 @@ def _build_payload(evt: dict, session_id: str) -> dict:
       ``(session_id, event_id)`` to dedupe across reconnects.
     """
     # ProcessRegistry completion events use the field name ``session_id`` for
-    # the process id. Alias it locally before exposing it as payload ``task_id``
-    # to avoid confusing that wire-format name with the WebUI session id.
-    process_id = str(evt.get("session_id") or "")
+    # the process id. Async delegation completions carry ``delegation_id``
+    # instead; expose either stable delivery id as payload ``task_id``.
+    process_id = completion_delivery_id(evt)
     payload: dict[str, Any] = {
         "session_id": str(session_id),
         "task_id": process_id,
@@ -787,7 +792,7 @@ def _process_one(evt: dict) -> None:
     except Exception:
         _process_registry = None
 
-    process_id = str(evt.get("session_id") or "")
+    process_id = completion_delivery_id(evt)
     session_key = str(evt.get("session_key") or "")
     # Root-cause fix (t_0f447014): the notify_on_complete completion event
     # enqueued by ProcessRegistry._move_to_finished() carries NO "session_key"
@@ -933,6 +938,8 @@ def _process_one(evt: dict) -> None:
                 # here cannot cause a double-fire — the atomic claim in
                 # ``claim_deferred_wakeups`` guarantees exactly one delivery.
                 record_deferred_wakeup(session_id, process_id, wakeup_prompt)
+                if evt.get("type") == "async_delegation":
+                    mark_async_delegation_record_consumed(process_id)
                 logger.debug(
                     "server-side wakeup deferred: turn active for session %s "
                     "(persisted for turn-teardown idle-hook redelivery)",
@@ -950,6 +957,8 @@ def _process_one(evt: dict) -> None:
                 _start_server_side_wakeup_turn(
                     session_id, wakeup_prompt, process_id=process_id
                 )
+                if evt.get("type") == "async_delegation":
+                    mark_async_delegation_record_consumed(process_id)
     except Exception:
         logger.warning(
             "server-side wakeup dispatch failed for session %s", session_id, exc_info=True
--- a/api/process_event_utils.py
+++ b/api/process_event_utils.py
@@ -1,0 +1,53 @@
+"""Shared helpers for WebUI completion/delegation delivery."""
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+
+def completion_delivery_id(evt: Any) -> str:
+    """Return the stable WebUI delivery/dedupe id for a completion event.
+
+    Terminal background-process events use ``session_id`` for the process id.
+    Async ``delegate_task`` completions carry ``delegation_id`` instead, so both
+    WebUI delivery paths must key those events by ``delegation_id``.
+    """
+    if not isinstance(evt, dict):
+        return ""
+    if evt.get("type") == "async_delegation":
+        return str(evt.get("delegation_id") or "").strip()
+    return str(evt.get("session_id") or "").strip()
+
+
+def mark_async_delegation_record_consumed(delegation_id: str) -> bool:
+    """Best-effort marker for Hermes async-delegation recovery records.
+
+    Newer Hermes Agent builds expose
+    ``tools.async_delegation.mark_async_delegation_consumed`` so recovery
+    sweeps do not re-inject a delegation result that WebUI already delivered.
+    WebUI must not hard-require that API while mixed deployments are common, so
+    absence or failure is a safe no-op.
+    """
+    deleg_id = str(delegation_id or "").strip()
+    if not deleg_id:
+        return False
+    try:
+        from tools.async_delegation import mark_async_delegation_consumed
+    except Exception:
+        logger.debug(
+            "Async delegation consumed marker unavailable; skipping record marker",
+            exc_info=True,
+        )
+        return False
+    try:
+        mark_async_delegation_consumed(deleg_id)
+        return True
+    except Exception:
+        logger.debug(
+            "Failed to mark async delegation record consumed for %s",
+            deleg_id,
+            exc_info=True,
+        )
+        return False
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -40,6 +40,7 @@ from api.config import (
     parse_reasoning_effort,
     coerce_reasoning_effort_for_model,
     _main_model_request_overrides,
+    PROCESS_SESSION_INDEX, PROCESS_SESSION_INDEX_LOCK,
 )
 from api.helpers import redact_session_data, _redact_text
 from api.compression_anchor import is_context_compression_marker, visible_messages_for_anchor
@@ -54,6 +55,10 @@ from api.models import (
     reconciled_state_db_messages_for_session,
 )
 from api.session_ops import mark_session_title_generated, session_has_manual_title
+from api.process_event_utils import (
+    completion_delivery_id,
+    mark_async_delegation_record_consumed,
+)
 
 
 def _session_payload_with_full_messages(session, *, tool_calls=None):
@@ -1659,6 +1664,14 @@ def _format_process_notification(evt: dict) -> str:
     """Format a completed background process notification for agent input."""
     if not isinstance(evt, dict):
         return ''
+    if evt.get('type') == 'async_delegation':
+        try:
+            from tools.process_registry import format_process_notification
+
+            return format_process_notification(evt) or ''
+        except Exception:
+            logger.debug("Failed to format async delegation notification", exc_info=True)
+            return ''
     if evt.get('type') != 'completion':
         return ''
     _sid = evt.get('session_id', '')
@@ -1681,6 +1694,26 @@ def _mark_process_completion_consumed(process_registry, process_id: str) -> None
             process_registry._completion_consumed.add(process_id)
     except Exception:
         logger.debug("Failed to mark process completion consumed", exc_info=True)
+
+
+def _completion_event_targets_webui_session(evt_session_key: str, session_id: str) -> bool:
+    """Return whether a completion event belongs to this WebUI session.
+
+    WebUI normally registers ``PROCESS_SESSION_INDEX[session_id] = session_id``.
+    Gateway/agent session keys can differ, so match the direct WebUI case first
+    and otherwise resolve through the same session-key index used by the
+    background wakeup path.
+    """
+    if not evt_session_key or not session_id:
+        return False
+    if evt_session_key == session_id:
+        return True
+    try:
+        with PROCESS_SESSION_INDEX_LOCK:
+            return PROCESS_SESSION_INDEX.get(evt_session_key) == session_id
+    except Exception:
+        logger.debug("Failed to resolve completion event session key", exc_info=True)
+        return False
 
 
 def _drain_webui_process_notifications(session_id: str) -> list[str]:
@@ -1716,17 +1749,20 @@ def _drain_webui_process_notifications(session_id: str) -> list[str]:
             logger.debug("Failed to drain process completion queue", exc_info=True)
             break
 
-        evt_sid = str(evt.get('session_id') or '') if isinstance(evt, dict) else ''
+        evt_sid = completion_delivery_id(evt) if isinstance(evt, dict) else ''
         if not evt_sid:
             skipped_events.append(evt)
             continue
         try:
             if process_registry.is_completion_consumed(evt_sid):
                 continue
-            proc = process_registry.get(evt_sid)
+            evt_session_key = str(evt.get('session_key') or '') if isinstance(evt, dict) else ''
+            if not evt_session_key:
+                proc = process_registry.get(evt_sid)
+                evt_session_key = str(getattr(proc, 'session_key', '') or '')
         except Exception:
-            proc = None
-        if getattr(proc, 'session_key', None) != session_id:
+            evt_session_key = ''
+        if not _completion_event_targets_webui_session(evt_session_key, session_id):
             skipped_events.append(evt)
             continue
 
@@ -1752,7 +1788,13 @@ def _drain_webui_process_notifications(session_id: str) -> list[str]:
         notification = _format_process_notification(evt)
         if notification:
             notifications.append(notification)
-            _mark_process_completion_consumed(process_registry, evt_sid)
+        # Matched but unformattable completions are consumed rather than
+        # replayed forever on later turns. Valid async_delegation events use the
+        # agent-side formatter above, so an empty notification here means the
+        # matched event is malformed/unsupported for this WebUI build.
+        _mark_process_completion_consumed(process_registry, evt_sid)
+        if isinstance(evt, dict) and evt.get('type') == 'async_delegation':
+            mark_async_delegation_record_consumed(evt_sid)
 
     for evt in skipped_events:
         try:
--- a/tests/test_async_delegation_webui_bridge.py
+++ b/tests/test_async_delegation_webui_bridge.py
@@ -1,0 +1,135 @@
+"""Regression coverage for async delegate_task completion routing beyond #4912.
+
+#4912 taught ``format_wakeup_prompt`` how to render ``async_delegation`` events.
+These tests cover the remaining WebUI bridge contract: route those events by
+``session_key``, dedupe by ``delegation_id``, and coordinate with the optional
+agent-side async-delegation consumed marker when available.
+"""
+from __future__ import annotations
+
+import queue
+import sys
+import time
+import types
+
+from api import background_process as bp
+from api import config as cfg
+from api import streaming
+
+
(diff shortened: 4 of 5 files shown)


sysophelper-droid:
Follow-up to the Greptile review comments:

- Added `api/process_event_utils.py` so both delivery paths share the same `event_delivery_id()` contract instead of carrying private duplicate helpers.
- Aligned fallback truncation through the shared `truncate_delivery_text()` helper.
- Reduced the new brittle source-level assertions in `tests/test_notify_on_complete_webui.py`; the async delegation routing behavior remains covered by functional tests in `tests/test_async_delegation_webui_bridge.py`.

Validation after the cleanup:

```text
45 passed
static_scan=clean
git diff --check OK
compileall OK
```

The pytest run is warning-free when using a narrow filter for the upstream `discord.py` `audioop` stdlib deprecation triggered during Hermes plugin loading; this PR does not introduce that warning.

nesquena-hermes:
Thanks @sysophelper-droid, and welcome — this is a real, wanted fix (async `delegate_task` completions silently dropping in WebUI sessions is exactly #4691, and a reporter is waiting on it). The routing approach (treating `async_delegation` completion events as first-class WebUI wakeup events + sharing the delivery-id/truncation contract) looks like the right shape.

**But the PR doesn't pass its own test, so it can't be gated yet.** CI is red on shards 1+2 (all Python versions) on **your own new test**:

```
tests/test_async_delegation_webui_bridge.py::test_format_wakeup_prompt_handles_async_delegation_batch
E  AssertionError: assert 'ASYNC DELEGATION BATCH COMPLETE' in
   '[ASYNC DELEGATION COMPLETE — deleg_test123]\nStatus: completed\n--- RESULT ---\n'
```

The test asserts the wakeup-prompt header is **`ASYNC DELEGATION BATCH COMPLETE`**, but the formatter emits **`[ASYNC DELEGATION COMPLETE — <id>]`** (singular, no "BATCH"). The test and the formatter disagree on the header wording — they need to be reconciled. Pick whichever is intended:

- If a **batch** of subagent tasks should read differently from a single completion, have the formatter emit the `BATCH` header for the multi-task case (and keep the singular `[ASYNC DELEGATION COMPLETE — <id>]` for one task), so both the batch and single shapes are covered; **or**
- If there's only one header shape, update the test to assert the actual `[ASYNC DELEGATION COMPLETE — …]` string.

Please run the full test file locally before pushing — `./scripts/test.sh tests/test_async_delegation_webui_bridge.py tests/test_notify_on_complete_webui.py` — so the header contract and the formatter agree. (The other shards passed, so this is the only blocker I can see; once it's green I'll run the full Codex + Opus + suite gate.)

A couple of things I'll be checking on the re-gate, so you can get ahead of them:
- **Dedup / no double-delivery:** an async-delegation completion must surface exactly once — confirm the shared delivery-id contract dedupes between the background-wakeup path and the next-turn drain (you mention sharing it in `process_event_utils.py` — good, just want a test that proves no double-surface).
- **Truncation:** the `--- RESULT ---` body should respect the same truncation contract as other wakeup events.
- **Routing to the originating session only:** the completion should wake the session that spawned the delegation, not broadcast.

Marking changes-requested. Ping me when the header contract is fixed and CI is green, and I'll gate it right away — this is a good fix and #4691's reporter will be glad to see it.

ticketclosed-wontfix:
## Review + Partner PR

We independently diagnosed the same bug and submitted a competing PR (#4805, now closed). After a multi-agent review of both approaches, this PR is the stronger one — upstream formatting delegation, active-turn deferral via `record_deferred_wakeup()`, and 5 dedicated tests are all things our PR was missing.

**One issue to flag:** This PR replaces `mark_async_delegation_consumed()` with `_mark_process_completion_consumed(process_registry, delegation_id)`, which adds to the process registry's `_completion_consumed` set. The hermes-agent has a recovery sweep in the tui_gateway that checks for unconsumed async delegation records (via `list_async_completions`) and re-injects them on session reconnect. Without calling `mark_async_delegation_consumed()` after successful delivery, the recovery sweep could re-deliver results already shown to the user.

The fix would be adding a `mark_async_delegation_consumed(delegation_id)` call after the wakeup/emission succeeds in `_process_one` and `_drain_webui_process_notifications`.

### Partner PR: NousResearch/hermes-agent#51577

The agent-side companion adds:
- `consumed` flag on async delegation records
- `list_async_completions(session_key)` — returns unconsumed completed delegations for a session
- `mark_async_delegation_consumed(delegation_id)` — marks a record so it won't be re-injected
- Recovery sweep in tui_gateway's `_notification_poller_loop` — picks up missed completions on session reconnect
- `delegations.list` RPC method for polling
- 5 new tests

This provides defense-in-depth for the tui_gateway/CLI path and a polling fallback for the WebUI if the primary delivery path fails. The `mark_async_delegation_consumed` function is the piece this PR should use for proper dedup coordination.

ticketclosed-wontfix:
@sysophelper-droid — would you like me to open a small PR targeting your branch with the `mark_async_delegation_consumed` fix? It's a two-line change in `_process_one` and `_drain_webui_process_notifications`. Happy to submit it so it flows through as part of this PR when upstream merges.

sysophelper-droid:
Pushed an update addressing the changes-requested feedback and the red CI test.

What changed:

- Reconciled the header contract:
  - batch events render `ASYNC DELEGATION BATCH COMPLETE`
  - single events render `ASYNC DELEGATION COMPLETE`
- Moved fallback async-delegation formatting into the shared `api/process_event_utils.py` path so both WebUI delivery paths use the same fallback when the core Hermes formatter is unavailable.
- Added explicit fallback formatter tests so CI no longer depends on whether `tools.process_registry.format_process_notification` is importable.
- Added regression coverage for the items you called out:
  - cross-path dedupe: background wakeup + next-turn drain cannot surface the same `delegation_id` twice
  - routing only to the originating `session_key`
  - shared truncation contract for async-delegation result bodies
- Added a best-effort optional `mark_async_delegation_consumed(delegation_id)` call after WebUI delivery when the installed Hermes Agent exposes that API. This coordinates with NousResearch/hermes-agent#51577 without hard-requiring that companion PR on older agent installs.

Validation:

```text
./scripts/test.sh tests/test_async_delegation_webui_bridge.py tests/test_notify_on_complete_webui.py
16 passed

focused pytest set covering async delegation + background wakeup paths
52 passed

git diff --check OK
compileall OK
static added-lines scan clean
```

nesquena-hermes:
Pulled the branch at `c2bda962d` and read the diff plus the agent-side contract. The changes-requested blocker is resolved, and the header contract now matches the agent's canonical formatter exactly — not just the test. A few verification notes for the re-gate.

## The header blocker is fixed and contract-aligned

The previous red test (`test_format_wakeup_prompt_handles_async_delegation_batch`) failed because the formatter emitted singular `ASYNC DELEGATION COMPLETE` for a batch event. The new `format_async_delegation_fallback` in `api/process_event_utils.py` now branches on `is_batch`:

```python
is_batch = bool(evt.get("is_batch") or evt.get("results"))
if is_batch:
    header = f"[ASYNC DELEGATION BATCH COMPLETE — {deleg_id}]"
    ...
return (f"[ASYNC DELEGATION COMPLETE — {deleg_id}]\n"
        f"Status: {status}\n--- RESULT ---\n{summary}")
```

I cross-checked this against the agent's canonical `_format_async_delegation` (`tools/process_registry.py:1827` and `:1884`): it emits `[ASYNC DELEGATION BATCH COMPLETE — {deleg_id}]` and `[ASYNC DELEGATION COMPLETE — {deleg_id}]` respectively. So the WebUI fallback now matches CLI/gateway rendering verbatim, and the `is_batch`/`results`/`goals`/`summary` fields the fallback reads line up with what the agent actually enqueues (`async_delegation.py:306` single, `:476` batch). Good — that's the right way to fix it (reconcile to the real contract, not just the assertion).

## Cross-repo: the consumed-marking is currently a no-op (by design)

`mark_async_delegation_record_consumed()` imports `tools.async_delegation.mark_async_delegation_consumed` under try/except. I grepped the installed agent — that symbol does **not** exist yet (today the module exposes `list_async_delegations()`, not `list_async_completions()`/`mark_async_delegation_consumed()`). So on current agents the call logs a debug and returns `False` — a safe no-op, exactly as the docstring claims. The implication worth stating plainly: the recovery-sweep re-injection protection that #4805's author described is **inert until NousResearch/hermes-agent#51577 lands**. WebUI dedup today rests entirely on `process_registry.is_completion_consumed()` (`_completion_consumed`), which the cross-path test `test_async_delegation_dedupes_between_background_wakeup_and_next_turn_drain` proves. That's fine for the single-process WebUI case; the consumed-marker only matters once the agent companion ships and a recovery sweep exists to suppress.

## Two things to check on re-gate

1. **Unconditional consume on empty format.** In `streaming.py` the `_mark_process_completion_consumed(...)` call is now dedented out of the `if notification:` guard, so an event is marked consumed even when `_format_process_notification` returns `''`. For a valid delegation the fallback always returns non-empty, so this is narrow — but a future formatter regression would silently drop (consume-without-deliver) rather than retry. Worth a one-line comment that this is the intended "never replay" tradeoff.

2. **Routing asymmetry between the two paths.** `_process_one` resolves `session_key → session_id` via `PROCESS_SESSION_INDEX` + the `_resolve_wakeup_target` misroute guard, while the drain path compares `evt_session_key` directly against the WebUI `session_id`. These agree only because WebUI registers `session_id → session_id`. The `test_..._requeues_..._for_other_session` test covers the WebUI case; if a gateway session_key ever diverges from the WebUI session_id, the drain path's direct compare wouldn't match while the index path would. Probably out of scope here, but worth confirming the gateway path isn't expected to hit the drain.

CI hasn't reported on `c2bda962d` yet (the rollup is empty) — once the shards go green this looks gate-ready on the code. Nice work reconciling the header to the actual formatter rather than the test.

sysophelper-droid:
Follow-up pushed for the two re-gate notes as well.

I addressed both rather than leaving them as implicit assumptions:

1. **Consume-without-replay tradeoff:** added an explicit comment at the unconditional `_mark_process_completion_consumed(...)` call in `streaming.py`. The intended behavior is that a matched but unformattable event should not be replayed forever; valid `async_delegation` events have the shared fallback formatter, so this only affects malformed/unsupported matched completions.

2. **Routing asymmetry:** next-turn drain now resolves non-identity `session_key -> session_id` mappings through `PROCESS_SESSION_INDEX`, matching the background wakeup routing model while preserving the direct WebUI `session_key == session_id` fast path. Added regression coverage for a `gateway-session-key -> webui-session-1` mapping to prove the event is not dropped or delivered to the wrong session.

Validation after this follow-up:

```text
./scripts/test.sh tests/test_async_delegation_webui_bridge.py tests/test_notify_on_complete_webui.py
17 passed

focused pytest set covering async delegation + background wakeup paths
53 passed

git diff --check OK
compileall OK
static added-lines scan clean
```

nesquena-hermes:
Labeling **`upstream-change` + `hold`** per maintainer direction.

This PR's dedup path coordinates with **NousResearch/hermes-agent#51577** (`mark_async_delegation_consumed(delegation_id)`) — an upstream Hermes Agent API that isn't merged yet. While the PR degrades gracefully when that API is absent (best-effort), we don't want to merge the async-delegation delivery until the upstream change lands and the end-to-end path is verified, so results are reliably deduped (not re-delivered by agent-side recovery sweeps).

Also still open from the earlier review: the PR's own test failed (`test_async_delegation_webui_bridge.py::test_format_wakeup_prompt_handles_async_delegation_batch` — the `ASYNC DELEGATION BATCH COMPLETE` header vs the emitted `[ASYNC DELEGATION COMPLETE — <id>]`); please reconcile the header contract so CI is green.

**Do not merge until:** (1) upstream hermes-agent#51577 is merged + available, and (2) CI is green. Holding rather than closing because the approach is sound and wanted (#4691). cc @sysophelper-droid.

nesquena-hermes:
Heads-up: the core symptom this PR addresses (async `delegate_task` completions not re-entering the WebUI chat) was fixed by a minimal, self-contained WebUI-side change in **v0.51.652** (#4912 → commit on master): `format_wakeup_prompt()` now delegates `async_delegation` events to the agent-side `format_process_notification`, no upstream dependency. If this PR carries additional scope beyond that delivery path, please rebase + narrow it to whatever remains; otherwise it may be supersedable. Leaving it as-is (hold/upstream-change) for @nesquena's call.

sysophelper-droid:
Rebased on current `master` and narrowed the PR after the #4912 / v0.51.652 minimal WebUI fix landed.

I removed the broader fallback-formatting scope from this branch. The PR now relies on the canonical agent formatter for async-delegation text, matching the direction from #4912.

What remains is the delta beyond #4912:

- `async_delegation` delivery/dedupe is keyed by `delegation_id` instead of process `session_id`.
- background wakeup and next-turn drain share that stable delivery id.
- next-turn drain can now handle `async_delegation` events and format them through the agent formatter.
- next-turn drain routing now supports direct WebUI `session_key == session_id` and non-identity `session_key -> session_id` mappings through `PROCESS_SESSION_INDEX`.
- WebUI best-effort calls `mark_async_delegation_consumed(delegation_id)` when newer agent builds expose it, while no-oping safely on older agents.

Validation after narrowing:

```text
python -m pytest tests/test_async_delegation_webui_bridge.py -q -o 'addopts='
3 passed

python -m pytest tests/test_background_process_wakeup_format.py -q -o 'addopts='
10 passed

focused async/background wakeup set
45 passed

./scripts/test.sh tests/test_async_delegation_webui_bridge.py tests/test_background_process_wakeup_format.py tests/test_notify_on_complete_webui.py
17 passed

git diff --check OK
compileall OK
ruff on new/updated tests + helper OK
static added-lines scan clean
```

Head is now based on current master and should no longer conflict. This should make the remaining review question explicit: whether the dedupe/routing/recovery-coordination delta beyond #4912 is still desired in WebUI.

nesquena-hermes:
Removing the `hold` on this. 🔓

The original hold reason — waiting to see the direction from the minimal #4912 / v0.51.652 WebUI fix — is now resolved, and you've correctly rebased on current master and narrowed this PR to just the remaining bridge-contract delta (delegation_id as the stable dedupe id, session_key routing for next-turn drain events) rather than the broader formatting scope that #4912 already covers. It's now MERGEABLE.

Un-holding so it can go through a fresh review of the narrowed scope. Next steps on our side: full gate (Codex regression + Opus + suite) on the current head, focused on the async-delegation delivery/dedup contract and the session_key routing. Thanks for keeping it tight and current, @sysophelper-droid.

nesquena-hermes (changes requested):
## 🔬 Authoritative re-gate (Codex) — SHIP ONLY WITH FIXES (stale against current core — 18 days of drift regressed it)

Heads-up @maroshmka — the delivery bridge design is still sound (routing, requeue, dedupe-namespace, normal `notify_on_complete`, `event_delivery_id` consistency all re-verified clean), but this PR was opened ~18 days ago and the **installed Hermes Agent's completion-delivery API has since changed underneath it**, producing two real regressions on the current head:

1. **[CORE — stale delivery API → duplicate autonomous turn on restart] `api/process_event_utils.py:24`** — the PR calls core's now-**removed** `mark_async_delegation_consumed`. Current core uses the durable `claim_event_delivery` / `complete_event_delivery` / `mark_completion_delivered` API instead. **Restart probe reproduced the consequence**: the obsolete marker is absent, so after a restart the durable delivery state stays `pending`, a fresh registry restores the `deleg_*` completion event, and it re-traverses the drain → **the finished delegation is delivered a second time = a duplicate autonomous turn**. **Fix:** use the current-core `claim_event_delivery` (claim before either WebUI consumer accepts an async event) / `complete_event_delivery` (after queue/defer/dispatch succeeds) / `release_event_delivery` (on failure); keep a compatibility fallback chain to `mark_completion_delivered` then the legacy marker for older cores.

2. **[CORE — unbounded growth] `api/background_process.py:994` + `api/streaming.py:2038`** — async delegation IDs are added to core's private `process_registry._completion_consumed`; a probe retained all 5,000 injected ids (that set is meant for normal `proc_*` ids and isn't pruned for these). **Fix:** stop putting async delegation ids in `_completion_consumed`; rely on the durable claim API, and where durable claims are unavailable use a **bounded** legacy dedupe.

3. **[test gap]** the current tests use fake registries that hide the installed-core API mismatch, and there's no restart/durable-dedupe or bounded-growth test. Add a real-core restart test proving **exactly one** delivery across a restart + a bounded-growth test.

The core routing/dedupe logic is good — it just needs to move onto the durable `claim/complete/release` delivery contract the agent now exposes (with a fallback for older cores) so a restart can't re-deliver a completed delegation. Rebased clean on current master, no conflicts. Once it's on the current delivery API + the dedupe set is bounded, re-gate and it should be solid. Thanks for the patience on a PR that aged against a moving core.

nesquena-hermes:
Thanks @sysophelper-droid for the thorough delivery-path work here (delegation_id keying, cross-path dedupe, next-turn drain routing, and the non-identity `session_key → session_id` routing coverage).

Closing this as **superseded by upstream**. The agent-side companion (NousResearch/hermes-agent#51577) was closed today — core `main` shipped a stronger, different guarantee in `d0e9a42ce` ("harden durable completion delivery"):

- `tools/async_delegation.py` now persists completed delegation events and rehydrates pending completions after restart, with exclusive delivery claims (`claim_event_delivery` / `complete_event_delivery` / `restore_undelivered_completions`).
- `tui_gateway/server.py` only acknowledges after successful injection; failed injections release the claim for retry.

That durable claim/complete model replaces the in-memory consumed-flag + poller-sweep approach this PR was coordinating with. Concretely, the `mark_async_delegation_consumed()` API this PR imports and calls **no longer exists** in core — so against current core this branch would call a removed symbol and (per our reproduction) risk a duplicate autonomous turn on restart plus unbounded dedupe growth. The recovery-coordination rationale is now handled natively.

If any of the WebUI-side delivery-path delta still adds value on top of core's durable model (e.g. specific dedupe/routing cases core doesn't cover), the right move is a fresh narrow PR scoped to just that gap rather than the now-broken coupling. Happy to help scope that — appreciate the diagnosis work, which fed directly into understanding the root cause chain.

PR #4800, opened 2026-06-23
fix(perf): eliminate O(DOM) streaming render freeze on long agent ans…

wlknight:
…wers

Introduce _safeSmdRenderer — a renderer wrapper that hooks smd's set_attr to validate URL schemes inline as each DOM node is created, eliminating the per-token _sanitizeSmdLinks(assistantBody) full-DOM scan that caused progressive freezing on answers of 30,000+ chars (5,000+ DOM nodes).

Also add parse-result caching to _scheduleRender to avoid redundant _extractInlineThinkingFromContent scans between the token event handler and the rAF render callback.

Changes:
- New _safeSmdRenderer(el): URL scheme safety inline via set_attr hook (identical algorithm to proven _streamFadeRenderer.set_attr, no animation)
- _smdNewParser: non-fade path uses _safeSmdRenderer instead of bare default_renderer
- _smdWrite: removed _sanitizeSmdLinks(assistantBody) call — safety now handled by renderer's set_attr at node-creation time
- _scheduleRender(parsed): accepts optional cached parse result from token handler to avoid duplicate O(n) text scans
- Token handler: passes _parseStreamState() result to _scheduleRender

_sanitizeSmdLinks is preserved at _smdEndParser (stream-end safety net) and in fallback paths where smd is not available.

All 166 existing tests pass.

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -3001,7 +3001,7 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
     _smdWrittenLen=0;
     _smdWrittenText='';
     if(!window.smd){_smdParser=null;return;}
-    const baseRenderer=fade ? _streamFadeRenderer(el) : window.smd.default_renderer(el);
+    const baseRenderer=fade ? _streamFadeRenderer(el) : _safeSmdRenderer(el);
     const renderer=_smdRendererWithoutUnderscoreEmphasis(baseRenderer);
     _smdParser=window.smd.parser(renderer);
   }
@@ -3072,9 +3072,9 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
     try{window.smd.parser_write(_smdParser,delta);}catch(_){}
     _smdWrittenLen=displayText.length;
     _smdWrittenText=displayText;
-    // streaming-markdown does NOT sanitize URL schemes. The default live path
-    // scans after writes; fade mode blocks unsafe href/src in its renderer.set_attr.
-    if(assistantBody&&!fade){_sanitizeSmdLinks(assistantBody);}
+    // URL scheme safety is handled by the renderer's set_attr hook
+    // (_safeSmdRenderer or _streamFadeRenderer), applied inline as smd
+    // creates each DOM node — no post-hoc full-DOM scan needed.
     _scheduleStreamingKatex();
   }
   // Allowed URL schemes for anchors and images rendered from agent-streamed markdown.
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
Reviewed the diff at `static/messages.js` against `origin/master` plus the existing fade-path renderer. The security-equivalence claim holds up — this is a safe swap, with one small follow-up worth a comment.

## The set_attr hook is byte-for-byte the proven fade-path logic

The new `_safeSmdRenderer.set_attr` (messages.js ~3257) and `_streamFadeRenderer.set_attr` (~3227) are character-identical in their URL-scheme handling:

```js
renderer.set_attr=(data,attr,value)=>{
  const isHref=window.smd&&attr===window.smd.HREF;
  const isSrc=window.smd&&attr===window.smd.SRC;
  const safeUrl=isSrc?_SMD_SAFE_IMG_URL_RE:_SMD_SAFE_URL_RE;
  if(isHref&&/^(file|workspace|session):\/\//i.test(String(value||''))){
    baseSetAttr(data,attr,_smdLinkHref(value)); ... }
  if((isHref||isSrc)&&!safeUrl.test(String(value||''))){
    ... node.setAttribute('data-blocked-scheme','1'); return; }
  baseSetAttr(data,attr,value);
};
```

Same allow-lists (`_SMD_SAFE_URL_RE` / `_SMD_SAFE_IMG_URL_RE`), same `file/workspace/session` rewrite via `_smdLinkHref`, same `session-link` class, same `data-blocked-scheme` block marker. The only difference vs. the fade renderer is that `_safeSmdRenderer` doesn't wrap `add_text` (no animation spans) — exactly right for the non-fade path. So removing the per-token `_sanitizeSmdLinks(assistantBody)` at `_smdWrite` doesn't open a hole: the validation just moves from a post-hoc O(DOM) `querySelectorAll` sweep to an inline check at node-creation time.

## End-of-stream safety net is intact

`_smdEndParser` still re-runs the full scan after `parser_end` flushes (messages.js:3041):

```js
if(assistantBody){_sanitizeSmdLinks(assistantBody);enhanceMarkdownTables(assistantBody);}
```

and the smd-unavailable fallback at ~3394 also calls `_sanitizeSmdLinks` directly, so any attribute set outside the `set_attr` path is still caught at flush. Good defense-in-depth.

## Parse cache is sound

`_parseStreamState()` (messages.js:2982) is pure over exactly two inputs:

```js
function _parseStreamState(){
  return _extractInlineThinkingFromContent(_stripXmlToolCalls(assistantText), liveReasoningText, {streaming:true});
}
```

and the `_doRender` guard reuses `_cachedParsed` only when `_cachedParsedText===assistantText && _cachedParsedReasoning===liveReasoningText` (messages.js:3778). Since those two are the *only* mutable inputs to the parse, the key is complete — no staleness risk, and `_cachedParsed=null` immediately after consumption prevents reuse across renders.

## One nit

The cache validity check reads `assistantText`/`liveReasoningText` directly rather than a captured snapshot, so it's correct only because `_doRender` runs on the rAF that mutates nothing between the comparison and the read. That's true today, but a one-line comment noting "cache key intentionally compares live state, not a snapshot — safe because _doRender is the sole consumer" would save a future reader the trace I just did. Not a blocker. Looks ship-ready otherwise; the Greptile 5/5 read matches mine.

nesquena-hermes:
Shipped in **v0.51.613** 🚀 — thanks @wlknight. Long agent answers no longer progressively freeze the UI mid-stream. Codex + Opus both verified the inline link-scheme validation has EXACT XSS parity with the removed full-DOM scan (javascript:/data:/vbscript: still rejected on links + images; final render still sanitizes). Maintainer fixes: keyed the parse-cache on liveReasoningText too (stale-thinking edge), and updated 2 stale test assertions that hardcoded the old _scheduleRender() signature. Full suite 10317 passed.

Issue #4783, opened 2026-06-23
perf(profiles): list_profiles_api() rescans all SKILL.md files on every request — 100s+ latency with multiple profiles


shrpfeather:
### Summary
`GET /api/profiles` takes very long to respond when the Hermes installation has multiple profiles, each carrying set of bundled skills. During this time the browser profile dropdown appears unresponsive, and repeated clicks compound the problem.
### Root Cause
`list_profiles_api()` in `api/profiles.py` calls `_get_profile_skills_stats()` for **every profile** on **every request**. That function:
1. Recursively finds **all SKILL.md files** via `iter_skill_index_files(skills_dir, "SKILL.md")`
2. **Opens and reads** each file (first 4000 bytes)
3. **Parses YAML frontmatter** from each
4. Checks **platform compatibility** via `skill_matches_platform()`
5. **Deduplicates by name** via a `seen_names` set
6. Counts `enabled_count` and `total_count`
All of this just to compute **two integers** per profile that are displayed in the profile picker dropdown.
The function does have an in-memory cache (`_SKILLS_STATS_CACHE`, TTL = 8 seconds), but:
- At **8s TTL**, the cache expires faster than a single request completes at this scale, so it never helps
- The browser tab polls `/api/profiles` multiple times during normal navigation, creating a cascading backlog
### Affected Configurations
- 5+ Hermes profiles, each with 60+ skills
- New installations where `hermes webui setup` seeds a full bundled skill tree
- Any install where the WebUI backend lives on network-attached storage (NAS) with higher file-open latency
### Actual vs Expected
| Metric | Before (observed) | Expected |
|---|---|---|
| First `/api/profiles` response | 100–170 seconds | < 2 seconds |
| Subsequent responses (cached) | 100–170 seconds (cache never hits) | < 0.5 seconds |
### Code Pointer
`api/profiles.py:997` — `_get_profile_skills_stats()`
`api/profiles.py:1068` — `list_profiles_api()`
`api/routes.py:5807` — route handler
#### Option A — Count skill directories instead of parsing files (recommended)
Replace the recursive `iter_skill_index_files()` + `parse_frontmatter()` path with a simple directory-level count:
```python
total_count = sum(1 for d in skills_dir.iterdir()
                  if d.is_dir() and (d / "SKILL.md").exists())
disabled_count = len(disabled_skills)
enabled_count = total_count - disabled_count
```
**Trade-offs:**
[...]

PR #4804, opened 2026-06-23
perf(#4718): pre-warm the new profile's sidebar cache on switch (phase 4 of #4662)

nesquena-hermes:
## What this is

Phase 4 of the profile-switch performance work (#4662): **pre-warm the new profile's sidebar so the conversation list appears almost immediately after a switch.** Closes the remaining cold-switch latency that @rodboev's profiling isolated on current master.

## Why (the measurement, on v0.51.610)

The original Phase-2 hypothesis on #4718 — that the synchronous gateway-watcher `stop()` join added "an uneven 1–5s on the request thread" — **does not hold on current master.** I re-measured after the four perf PRs that shipped since the original benchmark (#4741, #4769, #4775, #4478):

| Switch step | median | notes |
|---|---|---|
| `restart_watcher_for_profile()` | 51 ms | already start-before-stop |
| └ `GatewayWatcher.stop()` join | **0.4 ms** | the poll loop checks its stop event on 0.1s increments — not the suspected multi-second join |
| `/api/sessions` cold build | **~780 ms** | @rodboev's CDP, 249 sessions — **the real remaining cost** |
| `/api/sessions` warm | ~12 ms | cache works |

So the watcher async work was dropped (nothing to move), and the fix targets the actual cost: the cold `_build_session_list_cache_payload` the client's sidebar GET pays right after the switch POST returns.

Per @rodboev on #4662: *"POST /api/profile/switch returns in ~200ms and the server sits idle until the client's /api/sessions GET arrives, at which point it pays the full ~780ms rebuild. Pre-warming the target profile's cache from the switch handler would let the client hit a warm cache on arrival."*

## What it does

1. **`warm_session_list_cache()`** mirrors the GET `/api/sessions` handler's cache key + builder **exactly** (same settings reads, `sidebar_source=webui` / `exclude_hidden` / `visible_only=True`), warming the **same `_session_list_cache` layer the sidebar reads** — not the lower `get_cli_sessions()` cache, which #4769's `sidebar_source=webui` path no longer touches for sidebar loads.
2. **The `/api/profile/switch` handler launches it on a detached daemon thread** the moment a switch succeeds. Best-effort and fully detached — it never blocks or fails the switch response.
3. **The detached worker binds only the request-profile thread-local** (`set_request_profile`/`clear_request_profile`), not a process-env scope — so it warms the correct target profile (the build resolves via `get_active_profile_name()`/`get_active_hermes_home()`, both TLS-derived) with **no process-global `os.environ` mutation** and therefore no cross-profile bleed window.
4. **Cold-miss callers JOIN the in-flight warm** instead of racing it: when the client's GET lands while the pre-warm is mid-build, it rides that build to completion rather than giving up after 0.25s and redundantly rebuilding the same ~780ms payload. The join is bounded by a single 6.0s total deadline (each wait capped by remaining time), re-checks the cache before any fallback build, and falls through to its own build only if the owner is gone or the deadline passes.
5. **Dedup** so a burst of rapid switches can't fan out duplicate warm threads.

## Scope note

The warm covers the **default sidebar view** (the common case). Non-default states (all-profiles, archived, show-hidden, a CLI-source tab) use a different cache key and still pay the cold build on switch — acceptable as a best-effort warm of the common case, not full coverage (commented in code).

## Tests

`tests/test_issue4718_sidebar_prewarm.py` — 6 new tests: warm populates the sidebar cache, skips when already fresh, dedups concurrent warms, never raises, cold-miss follower joins an in-flight build (no double-build), and re-joins a slow build that exceeds one wait. Plus the existing `test_session_sidebar_cache.py` concurrency suite stays green.

## Gates

- **Codex:** SAFE TO SHIP (verified the bound, the cache re-check, TLS set/clear, no wait under the cache lock).
- **Opus:** ship it (both its findings — env-bleed + cold-join effectiveness, then the wait-bound — resolved).
- Full pytest suite: green.

## Review asks (@rodboev @franksong2702)

This is on the crown-jewel switch path, so beyond code review I'd value a **hands-on browser regression** with your CDP loop, @rodboev — specifically: does a cold switch to a large profile now land the sidebar at ~warm speed (the ~1s→~200ms target), and confirm no double-build in the network trace (the switch's pre-warm + the client GET should resolve to one build). Please do not merge — flagging for review.

Part of #4662 (phase 4). Measured and built on @rodboev's profiling — thank you.


nesquena-hermes:
@rodboev @franksong2702 — flagging this for your review (please don't merge yet).

This is phase 4 of #4662, built directly on your profiling, @rodboev. Two asks:

1. **Code review** of the pre-warm + cold-join/re-join logic in `api/routes.py`.
2. **Hands-on browser regression** with your CDP loop: cold-switch to a large profile should now land the sidebar at ~warm speed (the ~1s→~200ms target), and the network trace should show the pre-warm + the client GET resolving to **one** build (no double-build).

Dual gate is green (Codex SAFE TO SHIP, Opus ship-it, full suite 10313 passed). Scope note: the warm covers the default sidebar view; non-default states (all-profiles/archived/show-hidden/CLI tab) still cold-build on switch by design.

nesquena-hermes:
## Why the pre-warm uses a TLS-only scope (independent verification)

A reviewer might ask why the detached pre-warm worker binds only the request-profile thread-local (`set_request_profile`/`clear_request_profile`) rather than `profile_scope_for_detached_worker` (which also mutates process-global `os.environ['HERMES_HOME']`). I traced the full `/api/sessions` profile-resolution chain to confirm TLS alone is sufficient, and an independent second pass corroborated it. Evidence:

**The session-list read resolves the profile two ways, neither via `os.environ`:**

1. **WebUI JSON store** is a *single shared* store (`SESSION_DIR`/`SESSION_INDEX_FILE` are frozen at import, never re-pointed); profile scoping is a **filter** on the TLS-derived active profile:
   ```python
   # api/routes.py
   scoped = [s for s in merged if _profiles_match(s.get("profile"), active_profile)]
   # active_profile = get_active_profile_name()  → reads _tls.profile
   ```
2. **Agent state.db** paths derive from the active profile *name*, again TLS:
   ```python
   # api/profiles.py
   def get_active_hermes_home() -> Path:
       return _resolve_profile_home_for_name(get_active_profile_name())  # name, not os.environ
   def get_active_profile_name() -> str:
       tls_name = getattr(_tls, 'profile', None)   # set by set_request_profile()
       if tls_name is not None: return tls_name
       return _active_profile
   ```

No node in the `_build_session_list_cache_payload` → `all_sessions()` → state.db chain reads `os.environ['HERMES_HOME']` to select the profile. The process-env mutation in `profile_env_for_background_worker` exists for a *different* class of work — provider-credential and agent-config reads that call `os.getenv()` directly — **none of which are on the `/api/sessions` path.**

**The scope condition this PR satisfies:** TLS-only is correct *provided the warm only calls the `/api/sessions` builder + `all_sessions()` chain* (not provider/model/config/skill code that does direct `os.getenv()`). `warm_session_list_cache()` calls exactly that chain (the GET handler's key + builder) and nothing else — so binding the TLS only is both sufficient and the safer choice, since it eliminates the cross-profile `os.environ` bleed window entirely on this multithreaded server.

nesquena (approved):
## Review — end-to-end ✅ (approved after fix; pushed a docstring correction on the security-critical scope)

Independent review of Phase 4 of #4662 (#4718): pre-warm the new profile's sidebar cache on a profile switch so the client's `/api/sessions` GET hits a warm payload (~780ms cold build → ~warm) instead of paying the cold rebuild right after the switch POST returns. A concurrency-sensitive change on the crown-jewel switch path.

### What this ships

`api/routes.py` (+183/-4), `tests/test_issue4718_sidebar_prewarm.py` (+269), `CHANGELOG.md` (+4). Agent-authored; Codex + Opus gated across four iterations (env-bleed, cold-join effectiveness, slow-build re-join, total-wait bound — all resolved). I pushed one doc fix (below). The agent requested human + CDP browser regression review; my `event=APPROVE` is the code-review gate.

Fresh tarball: this warms a WebUI-internal in-process cache (`_session_list_cache`). The gateway-watcher work the original Phase-2 hypothesis targeted was correctly **dropped** (the `stop()` join is ~0.4ms, not multi-second — re-measured). No agent protocol/config/state surface; the warm reads the agent's `state.db` only through the existing TLS-scoped read path. Webui-only. ✓

### Thread-safety analysis (the core)

**Owner/waiter split is clean.** `_get_cached_session_list_payload` does `event, is_owner = _session_list_cache_claim_rebuild(key)`: the **owner** builds under the existing single-flight + invalidation-stamp re-check (≤3 attempts) and never reaches the wait logic. Only **waiters** (non-owners) hit the new cold-join loop.

**Cold-join loop** ([routes.py:2308+](api/routes.py)) — reached only on a true cold miss (no stale payload) by a waiter:
- Bounded by a single `cold_join_deadline = now + 6.0s` that starts **before** the first wait; each `current_owner.wait(min(2.0, remaining))` is capped by remaining time, so the whole join is ≤6s (not 2+6).
- The `wait()` is **outside** `_SESSIONS_CACHE_LOCK` (the lock is held only to read the current inflight owner) — no wait-under-lock.
- Re-checks the cache (`allow_stale=False`) after every wait and returns immediately on a hit.
- Breaks to its own build only when **no owner is in flight** (fall through promptly) or the **deadline passes** (a wedged/dead owner can't hang the request indefinitely); a new owner taking over is ridden (`event = current_owner`).
- Post-loop re-check before the fallback build avoids a redundant build when the owner just finished.

**Dedup** (`_SESSION_LIST_WARM_INFLIGHT` set + `_SESSION_LIST_WARM_LOCK`) prevents a burst of switches fanning out duplicate warm threads. **No lock-ordering deadlock:** the warm lock is held only briefly for the set check/add (and the finally discard) — released before the build acquires `_SESSIONS_CACHE_LOCK`; the cold-join never acquires the warm lock, so there's no nested A→B / B→A ordering.

**Best-effort detached worker:** the switch handler launches a `daemon=True` thread that `set_request_profile(name)` → `warm_session_list_cache` → `clear_request_profile()` in `finally`, all wrapped in try/except — it never blocks or fails the switch response.

### #798 cross-profile isolation — independently verified (the load-bearing security property)

The Opus gate moved the warm from `profile_scope_for_detached_worker` (which mutates process-global `os.environ['HERMES_HOME']`) to **TLS-only** (`set_request_profile`/`clear_request_profile`), to eliminate the cross-profile env-bleed window on this multithreaded server. I verified TLS-only is sufficient by tracing the entire warm path — not by trusting the PR's claim:

- `active_profile = get_active_profile_name()` → reads `_tls.profile` (TLS), set to the target. The cache key's profile component and the build's `_profiles_match(s, active_profile)` filter ([routes.py:184 of the builder](api/routes.py)) both use this TLS value.
- `get_active_hermes_home()` (for any state.db read) → `_resolve_profile_home_for_name(get_active_profile_name())` — TLS-derived, **never** `os.environ`.
- `load_settings()` reads `SETTINGS_FILE = STATE_DIR/"settings.json"` — **frozen at import / process-global** (settings aren't per-profile), so the warm and the GET read identical flags regardless of TLS, and it has no env/TLS path dependency.
- Grep of `_build_session_list_cache_payload` → `all_sessions()`: **no** `os.environ['HERMES_HOME']` / `os.getenv` profile read on the chain (the `os.getenv("HERMES_HOME")` hits in routes.py are all in unrelated functions — skills, saved_prompts, etc., not on this path). With `sidebar_source=webui` (#4769) the sidebar build doesn't even call `get_cli_sessions()`.

So the warm thread selects the target profile entirely via thread-local state and mutates **no** process-global env → it cannot bleed into concurrent requests on other threads. The agent's TLS-sufficiency argument checks out. ✓

### What I caught + pushed — commit 97096f97 (docstring accuracy on the security property)

`warm_session_list_cache`'s docstring still said the caller "uses `profile_scope_for_detached_worker`" — the very mechanism the Opus gate **removed** to avoid the env bleed. The actual caller uses TLS-only. That stale doc contradicts the security-critical design and could mislead a maintainer into reintroducing the `os.environ` mutation. I corrected it to state the TLS-only requirement and why (comment-only, no behavior change).

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Cold switch to large profile | warm builds on the detached thread; client GET joins/hits the warm payload (one build) ✅ |
| Client GET lands mid-warm (cold miss, in-flight owner) | joins the owner's build, bounded ≤6s, no double-build ✅ |
| Slow build exceeding one 2s wait | re-joins under the 6s deadline (no double-build, no stacked latency) ✅ |
| Owner dies before storing | cold loop breaks (owner gone) → fallback build ✅ |
| Wedged owner | 6s deadline → waiter builds itself (bounded) ✅ |
| Rapid back-to-back switches | dedup set → one warm thread per profile ✅ |
| Stale payload present | served immediately after a brief 0.1s wait (cold-join not entered) ✅ |
| Concurrent profiles | warm is TLS-only, no `os.environ` mutation → no cross-profile bleed (#798) ✅ |
| Non-default sidebar view (all-profiles/archived/hidden/CLI tab) | different cache key → still cold-builds (documented best-effort scope) ✅ |
| Warm raises | caught + TLS cleared in finally; switch response unaffected ✅ |

### Tests

- `tests/test_issue4718_sidebar_prewarm.py` (6) + `test_session_sidebar_cache.py` concurrency suite — **19/19** (warm populates, skips fresh, dedups concurrent, never raises, cold-miss follower joins without double-build, re-joins a slow build).
- **Full suite: 10195 passed / 133 skipped / 0 failures** in 211s (deselected the pre-existing darwin CRLF flake).
- `py_compile` clean on routes.py.

- Warm covers only the **default** sidebar view; non-default states cold-build on switch — documented in-code as deliberate best-effort scope (the common case is the one that matters for perceived speed).
- The 6s cold-join deadline is generous, but it's a hard bound that only applies to a true cold miss with a live in-flight owner on a very large profile — the realistic worst case this PR targets, and far better than the alternative (giving up at 0.25s and double-building the ~780ms payload).
- The warm reads process-global settings, so if `settings.json` were ever made per-profile, this path would need revisiting — not the case today (`SETTINGS_FILE` is frozen at import).

### Recommendation

✅ **Approved after fix (code review).** Parked at approval — ready for the release agent's merge/tag pipeline once @rodboev's CDP browser regression confirms the one-build / ~warm-speed result.

The pre-warm is correct and carefully bounded: a clean owner/waiter split, a cold-join that's deadline-bounded, waits outside the cache lock, re-checks before any fallback build, dedups warm threads, and has no lock-ordering deadlock. The #798-critical decision — TLS-only scope, no `os.environ` mutation — I verified end-to-end against the source (profile resolves via TLS-derived `active_profile`/`get_active_hermes_home`; settings are process-global; the build chain reads no `os.environ['HERMES_HOME']`), so there's no cross-profile bleed window. I pushed a docstring correction so that security property isn't silently undone later. Ship (pending the requested browser regression pass).

rodboev:
## Browser-level profiling: pre-warm race condition

Tested on v0.51.614 + this PR merged locally, profiling with a cookie-aware HTTP client that mirrors the browser's profile-switch flow (POST switch, forward the `hermes_profile` cookie, then GET `/api/sessions?sidebar_source=webui&exclude_hidden=1`).

### Results (cookie-aware, 4 rounds, 4s cache-expiry gaps)

| Round | Direction | `/api/profile/switch` | `/api/sessions` | Sessions |
|---|---|---|---|---|
| 0.a | default -> test | 202ms | 1374ms | 0 |
| 0.b | test -> default | 180ms | 787ms | 249 |
| 1.a | default -> test | 138ms | 712ms | 0 |
| 1.b | test -> default | 211ms | 687ms | 249 |
| 2.a | default -> test | 196ms | 667ms | 0 |
| 2.b | test -> default | 292ms | 688ms | 249 |
| 3.a | default -> test | 180ms | 668ms | 0 |
| 3.b | test -> default | 207ms | 668ms | 249 |

The pre-warm doesn't reduce latency in either direction. `/api/sessions` pays the full cold build (~670-790ms) every time.

### Why: the GET wins the race

The pre-warm daemon thread is spawned **after** `return j(handler, result, ...)` sends the POST response (routes.py:11971-11976). The sequence:

1. Switch POST handler completes, spawns pre-warm daemon thread, sends response
2. Client receives response, fires GET `/api/sessions`
3. GET arrives at server, handler calls `_get_cached_session_list_payload`, finds cold cache, claims rebuild ownership via `_session_list_cache_claim_rebuild`
4. Pre-warm thread finally gets scheduled by OS, calls `_get_cached_session_list_payload`, finds rebuild already in-flight, becomes the follower (skips or joins)

The cold-join mechanism (line 2300-2333) is correct but only helps when the pre-warm claims first. In practice the client's GET consistently wins because network + JS overhead for the client to fire the GET is shorter than OS thread scheduling delay for the newly spawned daemon.

### Cookie-less testing gave false positives

My initial benchmark (without cookies) showed def->test hitting warm (35ms) because without the `hermes_profile` cookie, the GET handler fell through to `_active_profile` (process-level), returning 249 sessions for the test profile (wrong: it has 0). The cache key happened to match because both the pre-warm and the cookieless GET resolved `active_profile` from the same process-level state. With cookies (as the browser does), the profiles scope correctly and neither direction hits warm.

### Possible fix

Spawn the pre-warm thread **before** sending the response so it gets a head start while the response transits the network and the client's JS parses it. Something like:

```python
# Start pre-warm BEFORE sending the response
thread = _threading.Thread(target=_warm_target_profile_sidebar, args=(name,), daemon=True)
thread.start()
return j(handler, result, extra_headers={'Set-Cookie': build_profile_cookie(name, handler)})
```

The browser still needs to receive the HTTP response, parse it, run the `switchToProfile` callback, and construct the sessions URL, which is 50-100ms of real-world latency that the pre-warm thread can use to build.

### Baseline comparison

Without this PR, cold switch is ~1000ms total (200ms switch + 780ms sessions). With this PR, ~870ms total (200ms switch + 670ms sessions). The switch itself is slightly faster (the async watcher restart helps), but the sessions cold build is unchanged because the pre-warm loses the race.

nesquena-hermes:
Thanks for the cookie-aware profiling — that's the right harness, and the table is convincing: the pre-warm isn't buying the ~200ms target. But I think the diagnosis points at the wrong cause, and the proposed fix is already in place.

## The spawn already happens before the response

Reading the diff at HEAD (`97096f97b`), the warm thread's `.start()` is at `api/routes.py:11976`, and `return j(...)` is at `11979` — the thread is launched **before** the switch response is sent, not after:

```python
                _threading.Thread(
                    target=_warm_target_profile_sidebar,
                    args=(name,),
                    name="profile-switch-prewarm",
                    daemon=True,
                ).start()                                    # 11976
            ...
            return j(handler, result, extra_headers={        # 11979
                'Set-Cookie': build_profile_cookie(name, handler),
            })
```

I checked every commit on the branch (`git log origin/master..HEAD`) and the `.start()` precedes the `return j(...)` in all five. So the patch you sketched is a no-op against this HEAD — it is what the code already does. (Your citation `11971-11976` is the thread *construction*; the return is three lines down.)

## Why pre-warm still can't win here

The pre-warm can only land a warm payload if the build **completes** before the client's GET arrives. The head-start window is just `response transit + JS parse + GET round-trip` — your own number, ~50-100ms. The build is ~670ms. So the GET always arrives mid-build and becomes a cold-join waiter at `_get_cached_session_list_payload` (`routes.py:2256` claim, `2309-2333` join loop), then blocks on the full remaining build. Pre-warming an in-process cache is only a win when `build_time < head_start`; at 670ms vs ~80ms that is never true, regardless of spawn order or GIL scheduling.

What the PR *does* correctly deliver is **single-build, not double-build**. Without the cold-join, the warm thread and the GET would each claim and build the same ~670ms payload concurrently, contending on the GIL and dragging both toward ~1.3s. The cold-join collapses that to one build — which is exactly why your baseline shows ~870ms *with* the PR vs ~1000ms without, rather than a regression. So the dedup/cold-join machinery is sound; it just can't beat the build floor.

## The real lever is the build cost

The smoking gun is in your own table: `default -> test` returns **0 sessions** and still takes 667-712ms. A build whose cost is independent of session count means the ~670ms is a **fixed cost** inside `_build_session_list_cache_payload` (`routes.py:1939`) / `all_sessions()`, not the per-row work. That is where the latency actually lives, and pre-warm/cold-join can't touch it.

Good news: the builder already threads `diag` through (`diag_stage("all_sessions")` at `1992`, plus `reconcile_stale_stream_state`, `get_cli_sessions`, `merge_cli_sessions` stages). The GET path passes a live `RequestDiagnostics` into it (`routes.py:10035` `diag=diag`). A single profiled cold switch with those stages dumped will pin down which stage eats the 670ms on an empty profile.

My strong suspicion is it overlaps the cluster in #4783 (`list_profiles_api()` SKILL.md rescan), #4765 (all session JSON loaded into RAM), and #4759 (cold-boot sidebar). If the fixed cost is the SKILL.md / all-sessions scan, the fix belongs there, and this PR is then pure insurance against double-builds rather than a latency win.

Recommendation: keep the cold-join (it is correct and prevents the double-build), but reframe the PR's claim from "~1s to ~200ms" to "eliminates the post-switch double-build," and open the build-cost profiling as the actual #4718 latency fix. Happy to dig into the diag stages if you post a cold-switch trace.

nesquena-hermes:
Superseded by #4822.

@rodboev's browser-level profiling here was decisive: the pre-warm consistently loses the race against the client's follow-up `GET /api/sessions`, so cold `/api/sessions` still paid the full rebuild. Rather than chase the race, I profiled *why the rebuild is ~670ms* and found a much larger root cause — `get_cli_sessions()` re-parsing every Claude Code transcript under `~/.claude/projects` line-by-line on every build (654ms of the ~670ms), with no caching and re-run on every profile switch / 5s TTL / poll.

#4822 caches that per-file parse by `(path, mtime, size)`, dropping cold `/api/sessions` from ~670–880ms to ~100–200ms (verified in-browser), with byte-identical output and the full suite green. Because there's no race, the win holds regardless of switch-vs-GET timing.

Closing this in favor of #4822 once that lands. Thanks @rodboev — your "the GET wins the race" finding is exactly what redirected the fix.

nesquena:
## Re-check (batch pass) — approval stands, but a non-trivial rebase is now needed before merge

Revisiting this as part of a sweep of our open self-authored PRs. My earlier `APPROVED` still reflects the code as written against its original base — no concerns there. But it's now `CONFLICTING`, and unlike the rest of the batch (which only collided on the CHANGELOG `[Unreleased]` block), **this one has a substantive code conflict** that the release agent / author should resolve before merge rather than a mechanical fixup:

- The first commit's CHANGELOG entry conflicts only on the `[Unreleased]` block — trivial (the entry belongs under the existing `### Changed`).
- **The second commit (`cold-join wait + TLS-only warm scope`) conflicts in `api/routes.py`.** Master has since **extracted the session-list cache into a new `api/_route_session_list_cache` module** — `routes.py` now re-exports `_SESSIONS_CACHE`, `_SESSIONS_CACHE_TTL_SECONDS`, `_SESSIONS_CACHE_WAIT_SECONDS`, `_SESSIONS_CACHE_STALE_WAIT_SECONDS`, the invalidation-version symbols, etc. from that module. This PR's commit still **defines those constants inline** and adds `_SESSIONS_CACHE_COLD_JOIN_WAIT_SECONDS = 2.0` inline alongside them.

So a clean rebase isn't a marker-deletion — the cold-join-wait constant (and any other cache touchpoints from the prewarm work) need to be **re-applied onto the new `_route_session_list_cache` module** (define + re-export there), and the cold-miss join-wait logic in the sessions GET handler re-checked against the refactored cache accessors. I deliberately didn't hand-resolve and force-push that, since it's effectively re-applying the feature onto a refactored base, not resolving a conflict — better owned by the author so the cache semantics stay correct.

**Ask:** rebase onto current master, re-homing the cold-join-wait constant + logic into `_route_session_list_cache`, then I'll re-verify the prewarm/cold-join behavior against the new module and re-confirm. The design itself is unchanged and still good — this is purely a rebase-onto-refactor.

nesquena-hermes:
Checked whether this could be mechanically rebased onto current master to clear the CONFLICTING state. The CHANGELOG conflict is trivial (resolved cleanly), and commit 1/5 rebases fine — **but commit 2/5 hits an architectural conflict in `api/routes.py` that needs a design decision, so I'm not force-rebasing it.**

**What changed underneath it:** master has since refactored the session-list cache out of `routes.py` into a dedicated module — `_route_session_list_cache` — so `_SESSIONS_CACHE`, `_SESSIONS_CACHE_INFLIGHT`, `_SESSIONS_CACHE_LOCK`, `_SESSIONS_CACHE_MAX_ENTRIES`, and `_SESSIONS_CACHE_PROFILE_INVALIDATION_VERSION` are now module-level re-exports from there rather than inline definitions. This PR's pre-warm logic (phase 4 of #4662) was written against the **old inline cache structure**, so its rebase collides with the new module boundary.

This isn't a mechanical resolution — the pre-warm needs to be **re-expressed against the `_route_session_list_cache` module API** (call its single-flight / invalidation / warm entry points rather than the old inline globals), which is a real design choice on the profile-switch path. Since it's a self-built perf PR on the crown-jewel switch surface, I'm escalating rather than guessing at the resolution.

**Recommended path (maintainer):** rebase onto current master and rewrite the pre-warm to use the extracted `_route_session_list_cache` module — specifically warm the exact cached response the sidebar now reads through that module, reusing its single-flight + profile-invalidation guards (the PR's original safety properties: detached/best-effort, target-profile-scoped, dedup, no-stale-serve should all still hold, just wired to the new module). Then re-gate. Alternatively, if the profile-switch cold-latency this addressed (#4718 / #4662 phase 4) has already been mitigated by the cache-module refactor itself, this may be partially superseded — worth measuring cold-switch latency on current master before re-investing.

Leaving as-is (CONFLICTING) pending that call — no code touched.

nesquena-hermes:
Closing — the perf win this targets has largely evaporated since it was written. #4804 pre-warmed the target profile's sidebar cache on switch to hide a ~780ms cold `/api/sessions` rebuild (measured on v0.51.610, @rodboev's 249-session profile). Re-measured on current master: a cold build of **304 sessions now takes ~210ms** (warm ~20ms) — the intervening cache work (#5803's profile-scoped lock-guarded cache + the read-only projection in #5464) shrank the cold cost ~4×, so the marginal gain from pre-warming is now small. The PR is also CONFLICTING against the refactored `route_session_list_cache` module, so reviving would mean a full rebuild for a modest optimization. Given the current focus on reliability bug/regression fixes, closing this rather than reviving. The pre-warm idea remains captured in the #4662 profile-switch-perf thread if the cold cost ever regresses on very-large-session installs.

PR #4809, opened 2026-06-24
Preserve processed Worklog duration after session merge

franksong2702:
## Thinking Path

The settled Compact Worklog already renders elapsed time when the assistant message carries turn-duration metadata. The failure was in backend reconciliation: duplicate assistant rows from the sidecar/state.db or parent lineage could be deduped by keeping the older copy without display metadata and discarding the copy with `_turnDuration`.

## What Changed

- Added a display-metadata backfill helper for duplicate transcript rows.
- Applied it in `merge_session_messages_append_only()` before skipping duplicate state.db rows.
- Applied it in `_merged_webui_lineage_messages_for_display()` before skipping duplicate parent/child rows.
- Added regression coverage for state.db reconciliation and WebUI lineage display merge.
- Updated `CHANGELOG.md` under `[Unreleased]`.

## Why It Matters

Settled Processed Worklog summaries can now keep showing the elapsed duration after recovery, replay, or lineage merge, instead of silently losing the duration when the duplicate row with `_turnDuration` is dropped.

## Contract Routing

- Task type: runtime/session metadata bugfix.
- State layer: visible transcript/session metadata during sidecar, state.db, and lineage reconciliation.
- Relevant contract docs reviewed: `docs/CONTRACTS.md`, `docs/rfcs/webui-run-state-consistency-contract.md`, `docs/rfcs/live-to-final-assistant-replies.md`.
- Invariant preserved: duplicate reconciliation remains append-only/idempotent for visible messages, while display-only turn metadata survives on the retained row.

## Verification

- `git diff --check`
- `./scripts/test.sh tests/test_webui_lineage_display_merge.py::test_webui_lineage_display_merge_preserves_duplicate_turn_duration tests/test_webui_state_db_reconciliation.py::test_state_db_duplicate_backfills_turn_duration -q`
- `./scripts/test.sh tests/test_webui_lineage_display_merge.py tests/test_webui_state_db_reconciliation.py -q`

## Risks / Follow-ups

Low risk. This does not alter message content, ordering, or merge keys; it only fills missing display-only metadata on a row that was already selected as the retained duplicate.

## Model Used

OpenAI GPT-5 Codex via Codex CLI/API. Notable tool use: local git/worktree commands, Codegraph for blast-radius inspection, `./scripts/test.sh`, and GitHub CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Settled Compact Worklog summaries keep their processed duration after session recovery or lineage merge.** When WebUI deduped duplicate assistant rows from the sidecar, state.db, or parent lineage, it could keep the copy without display-only turn metadata and drop the copy that carried `_turnDuration` / anchor scene metadata. Duplicate merges now backfill missing display metadata before discarding the duplicate, so the settled Processed Worklog can still show elapsed time.
+
 ## [v0.51.614] — 2026-06-23 — Release VU (Kanban consolidated view toggle)
 
 ### Added
--- a/api/models.py
+++ b/api/models.py
@@ -4677,6 +4677,38 @@ def _session_message_merge_key(msg: dict):
     )
 
 
+_SESSION_MESSAGE_DISPLAY_METADATA_KEYS = (
+    "_turnDuration",
+    "_turnTps",
+    "_turnUsage",
+    "_firstTokenMs",
+    "_gatewayRouting",
+    "_statusCard",
+    "_anchor_stream_id",
+    "_anchor_activity_scene",
+)
+
+
+def _message_display_metadata_value_present(value) -> bool:
+    if value is None or value == "":
+        return False
+    if isinstance(value, (dict, list, tuple, set)) and not value:
+        return False
+    return True
+
+
+def _merge_session_display_metadata(target: dict | None, source: dict | None) -> None:
+    """Preserve display-only turn metadata when duplicate transcript rows merge."""
+    if not isinstance(target, dict) or not isinstance(source, dict):
+        return
+    for key in _SESSION_MESSAGE_DISPLAY_METADATA_KEYS:
+        if _message_display_metadata_value_present(target.get(key)):
+            continue
+        value = source.get(key)
+        if _message_display_metadata_value_present(value):
+            target[key] = copy.deepcopy(value)
+
+
 def _session_message_dedup_key(msg: dict):
     """Like _session_message_merge_key but preserves full-precision timestamp.
 
@@ -5100,12 +5132,16 @@ def merge_session_messages_append_only(
         # collapsing same-second distinct turns would be worse than retaining
         # a compaction-restamped duplicate.
         seen = set()
+        seen_messages = {}
         deduped = []
         for msg in filtered:
             key = _session_message_dedup_key(msg)
             if key not in seen:
                 seen.add(key)
+                seen_messages[key] = msg
                 deduped.append(msg)
+            else:
+                _merge_session_display_metadata(seen_messages.get(key), msg)
         return deduped
 
     merged_messages = []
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Pulled the branch and read the full diff against `origin/master` plus the three merge-key functions and `get_session`'s caching. The core fix is correct, and I can narrow Greptile's two concerns: one is overstated, one is real but low-severity.

## The fix is sound

`_merge_session_display_metadata` backfills the eight display-only keys onto the retained duplicate before the discard, and it's wired into every skip site in `merge_session_messages_append_only` plus the lineage merge. The guard is conservative — it only fills a key the target is *missing*, and it deep-copies the value:

```python
for key in _SESSION_MESSAGE_DISPLAY_METADATA_KEYS:
    if _message_display_metadata_value_present(target.get(key)):
        continue
    value = source.get(key)
    if _message_display_metadata_value_present(value):
        target[key] = copy.deepcopy(value)   # models.py:4708
```

That correctly addresses the reported symptom: dedup keeping the metadata-less copy and dropping the `_turnDuration`-bearing one.

## Greptile's "wrong-dedup decisions" worry doesn't hold

The review flags that mutating dicts "could produce subtle wrong-dedup decisions in future merge calls." I checked all three key builders, and none of them read any of the eight backfilled keys:

- `_session_message_merge_key` (`api/models.py:4646`) keys on `id`/`message_id`, else `role`/`content`/`timestamp`/`tool_call_id`/`tool_name`/`tool_calls`.
- `_session_message_dedup_key` — same fields, full-precision timestamp.
- `_session_message_visible_key` (`api/models.py:4730`) — `role`/normalized-content/`tool_calls`.

`_turnDuration`, `_turnUsage`, `_anchor_stream_id`, etc. are never part of any key, so writing them back cannot change a merge/dedup/visible decision on any subsequent call. That concern can be set aside.

## The real concern: mutating the cached parent in the lineage path

This one is legitimate but smaller than the dedup framing. In `_merged_webui_lineage_messages_for_display` (`api/routes.py:6111`):

```python
parent = get_session(parent_id, metadata_only=False)
parent_messages = list(getattr(parent, "messages", []) or [])
```

`get_session` returns a **live object from the `SESSIONS` LRU** (`api/models.py:2409`), and `list(...)` is a shallow copy — the dict elements are shared with the cached parent. The merge iterates `list(parent_messages) + list(primary_messages)` sorted by `(timestamp, role, content)`, so for an identical turn the parent copy sorts first, becomes the retained `target`, and the backfill writes display keys onto the **cached parent's message dict** in place.

Why it's low-severity: the value comes from a true-duplicate of the same logical turn, the helper deep-copies the value (so no nested aliasing), the keys are display-only, and this is a read/display path that doesn't itself persist. So nothing is corrupted today. But silently mutating a shared cached `Session` from a function named `_for_display` is the kind of side effect that bites a later refactor — e.g. if that parent is subsequently saved, the keys leak into its sidecar JSON.

Cheap fix: copy the parent rows before merging so the display path stays side-effect-free:

```python
parent_messages = [copy.deepcopy(m) for m in (getattr(parent, "messages", []) or [])]
```

The `merge_session_messages_append_only` sidecar-replay path (`api/models.py:5187`) has the same shape — `replay_target = sidecar_visible_messages[idx]` is a caller ref — but its callers generally pass freshly-merged lists, so the exposure is narrower there. Worth a one-line note in the PR either way.

## CHANGELOG

Agree with Greptile — the repo convention is that `CHANGELOG.md` is release-agent-only. PR #4813 just reverted exactly this kind of entry ("revert: remove CHANGELOG entry per request"); same call here.

Net: solid, well-tested fix for a genuine bug. Recommend the parent-copy guard above and dropping the CHANGELOG hunk, then it's good to go. I did not run the test suite (PR code isn't executed in this environment), but the new assertions match the behavior the diff enables.

nesquena-hermes:
Shipped in **v0.51.618** 🚀 — thanks @franksong2702. Settled Compact Worklog summaries now keep their processed duration after a session recovery/lineage merge (dedup backfills display-only _turnDuration metadata onto the kept row before discarding the duplicate). Codex SAFE (verified display-only, no content-bleed, no ghost resurrection, doesn't change which row is kept) + full suite 10342.

PR #4813, opened 2026-06-24
fix: strip @provider: prefix from cron job model override

luandnh:
## Thinking Path

- Cron jobs store model and provider as separate fields in jobs.json
- The cron form model dropdown uses /api/models which can return `@provider:` prefixed model IDs (from _apply_provider_prefix)
- _modelStateForSelect returns the full prefixed value as the model, which gets stored incorrectly
- The fix strips the `@provider:` prefix client-side in saveCronForm before sending to the API

## What Changed

- Added _cronModelBareName() helper in static/panels.js to strip `@provider:` prefix from model value
- Applied the helper in saveCronForm across create, update, and duplicate paths
- Updated CHANGELOG.md

## Why It Matters

Without this fix, cron jobs with a custom provider override store a provider-qualified model string (e.g. `@custom:9router:chat`) instead of just the base model name (chat). Since the provider field already stores the provider identity, the model value is redundant and can cause resolution issues at job execution time.

## Verification

- All 48 cron-related tests pass


nesquena-hermes:
Pulled the branch and read the diff against `origin/master` plus the agent-side cron consumer. The fix is correct and lands in the right layer — confirming below.

## What the diff does

`_cronModelBareName()` (`static/panels.js:1315`) strips a `@provider:` prefix only when it matches the separately-stored provider, then it's applied across the create/update/duplicate paths:

```js
if (model && provider && model.startsWith('@' + provider + ':')) {
  return model.slice(('@' + provider + ':').length);
}
return model;
```

## Why the prefix grammar lines up

This is the part worth double-checking, because the provider segment itself can contain a colon (named custom providers like `custom:9router`). It holds up:

- Backend emits the prefix as `f"@{provider_id}:{mid}"` (`api/config.py:2111`).
- `_modelStateForSelect` → `_providerFromModelValue` derives the provider via `value.slice(1, value.lastIndexOf(':'))` (`static/ui.js:1703`), so for `@custom:9router:chat` the provider is `custom:9router`, not `custom`.
- Your strip reconstructs `'@' + provider + ':'` = `@custom:9router:` and slices that exact length, leaving `chat`. So the multi-colon case is handled correctly — a naive `split(':')[2]` would have broken it, but `slice` keyed on the real provider doesn't.

## Confirming this is the right layer (agent side)

The reason storing the bare name matters: the cron runner consumes `model` and `provider` as independent fields and never strips a prefix itself. In `cron/scheduler.py`:

```python
model = job.get("model") or os.getenv("HERMES_MODEL") or ""   # :1876
...
runtime = resolve_runtime_provider(requested=job.get("provider"))  # :1983
agent = AIAgent(model=model, provider=runtime.get("provider"), ...)  # :2050+
```

So provider identity is resolved entirely from `job["provider"]`, and `model` is passed verbatim to `AIAgent`. If `job["model"]` is `@custom:9router:chat`, the agent gets a provider-qualified string as the *model name* while *also* resolving the provider separately — a double application that reaches the API as a malformed model id. Storing the bare `chat` is exactly what the agent expects. Good call fixing it client-side at save time rather than papering over it in the runner.

## One edge to sanity-check

The duplicate path now routes `_cronPreFormDetail.model` through the helper too (`static/panels.js:1391`). That's correct for jobs created *after* this fix, but a job created *before* it already has the prefixed value persisted in `jobs.json` — duplicating it will still strip correctly here (good), but the original stays prefixed until re-saved. Worth a one-line note in the PR that this fixes new/edited jobs and that pre-existing jobs self-heal on their next save through the edit form. No code change needed for that; just setting expectations.

Couldn't run the cron test suite from here (worktree code isn't executed in this environment), but the logic and the cross-repo contract check out. LGTM pending the maintainer's normal CI run.

nesquena-hermes:
Shipped in **v0.51.625** 🚀 — thanks @luandnh. Cron jobs now save the bare model name instead of the `@provider:` display prefix, so model overrides work. I added a regression test for `_cronModelBareName` covering the strip matrix. Full suite 10363.

luandnh:
Thanks for shipping it! 🙏

I noticed the release v0.51.625 credits @nesquena-hermes as the contributor for this fix, but the actual code changes came from my fork (luandnh/hermes-webui). Could you update the release notes to reflect the right contributor? And could you add me to the contributors list on the repo? Appreciate it!

nesquena:
> Thanks for shipping it! 🙏
> 
> I noticed the release v0.51.625 credits @nesquena-hermes as the contributor for this fix, but the actual code changes came from my fork (luandnh/hermes-webui). Could you update the release notes to reflect the right contributor? And could you add me to the contributors list on the repo? Appreciate it!

Even though it shows the agent in one place - Your commits always make it into the project, you were given full credit, and you'll show up on the contributors list on the next generation! (runs weekly)

Thanks again!

PR #4818, opened 2026-06-24
fix(#MOBILESCROLL): prevent scrollTop jank to top on mobile during streaming DOM rebuild

akrhin:
## Problem

On mobile devices (iOS Safari, Android Chrome), when the assistant streams a response, the transcript viewport randomly jumps to `scrollTop=0`. The user has to scroll back down manually each time — a severe UX regression on mobile.

**Root cause:** `.messages { overflow-anchor: none }` was set unconditionally. On mobile, this means every `innerHTML=''` + DOM rebuild in `renderMessages()` can cause the browser to paint a frame with `scrollTop=0` before the snapshot restore runs.

## Changes

| File | Change |
|------|--------|
| `static/style.css` | `.messages` defaults to `overflow-anchor: auto`; desktop (hover+fine pointer) overrides back to `overflow-anchor: none` via `@media (hover:hover) and (pointer:fine)` |
| `static/ui.js` | Added `window._fixMobileScrollJank()` — temporarily forces `overflow-anchor: auto` before `innerHTML=''` on mobile only, clears it on the next rAF. Called from `renderMessages()` right before the DOM wipe. |
| `tests/test_issue1360_streaming_scroll_hardening.py` | Split single overflow-anchor assertion into desktop and mobile variants |
| `tests/test_mobile_scroll_jank.html` | New interactive manual test page for mobile verification |

## Why this works

**CSS layer:** Mobile gets `overflow-anchor: auto` by default from the `.messages` rule. Desktop retains the old `none` behavior via the media query — tool/card inserts still can't yank the viewport on mouse-driven devices.

**JS guard layer:** Before the `innerHTML=''` wipe, `_fixMobileScrollJank()` sets inline `overflowAnchor: 'auto'` (trumping the media-query `none` in the brief window where desktop could also jank), then clears it on the next animation frame so the CSS class takes back over.

## Cleanup (Greptile review feedback)

- Removed dead `_scrollAnchorRestore` variable in `renderMessages()` — `_fixMobileScrollJank()` returns `undefined`, the const was never read
- Removed redundant `_fixMobileScrollJank()` call from `_doRender()` — CSS media query already gives `overflow-anchor: auto` on mobile; on desktop the guard exits early. The call was a no-op on every streaming tick

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -3844,6 +3844,9 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       _renderPending=false;
       // Guard: a pending setTimeout+rAF can outlive stream finalization.
       if(_streamFinalized) return;
+      // Mobile scroll-jank guard: enable overflow-anchor before DOM writes so
+      // the browser preserves scroll position during streaming content growth.
+      if(typeof window._fixMobileScrollJank==='function') window._fixMobileScrollJank();
       _lastRenderMs=performance.now();
       const parsed=_cachedParsed&&_cachedParsedText===assistantText&&_cachedParsedReasoning===liveReasoningText ? _cachedParsed : _parseStreamState();
       _cachedParsed=null;
--- a/static/messages.js
+++ b/static/messages.js
@@ -3844,9 +3844,6 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       _renderPending=false;
       // Guard: a pending setTimeout+rAF can outlive stream finalization.
       if(_streamFinalized) return;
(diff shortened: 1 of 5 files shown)


nesquena-hermes:
Shipped in **v0.51.617** 🚀 — thanks @akrhin. The mobile scrollTop-jank-to-top during streaming is fixed (overflow-anchor:none scoped to desktop via @media hover/fine; mobile keeps native anchoring + a brief re-enable across the DOM-rebuild gap). Codex SAFE (desktop virtual-scroll unchanged, #4803 recycle hardening intact) + full suite 10341. Closes #4815 + #4818.

PR #2814, opened 2026-05-24
feat: add Kanban consolidated view toggle

stefan-wisbric:
## Summary

- Adds a Kanban header toggle for consolidated board view.
- Persists the mode through `PATCH /api/kanban/config` to `dashboard.kanban.lane_by_profile` in the active server config.
- Keeps server config as the default on load, while user/browser clicks update that shared config and immediately rerender the board.
- Includes accessible `aria-pressed` state, tooltip/i18n strings, pressed-state styling, and regression coverage.

## Verification

- `pytest tests/test_kanban_view_toggle.py -q`
- `pytest tests/test_kanban_view_toggle.py tests/test_kanban_ui_static.py tests/test_issue1431_toolsets_chip_responsive.py tests/test_chinese_locale.py tests/test_japanese_locale.py tests/test_korean_locale.py tests/test_russian_locale.py tests/test_spanish_locale.py tests/test_provider_mismatch.py::TestModelSwitchToast::test_toast_in_model_select_onchange -q`
- `pytest tests/test_css_tooltips.py -q`

## Browser Check

- Restarted the local PR WebUI on `127.0.0.1:8787`.
- Verified the Kanban header shows the consolidated-view toggle.
- Clicked the toggle and confirmed:
  - `aria-pressed="true"`
  - `/api/kanban/config` returns `lane_by_profile: false`
  - `/Users/stefans/.hermes/config.yaml` contains `dashboard.kanban.lane_by_profile: false`
  - after reload, the board renders as 6 direct status columns with no profile-lane wrapper.

PR #4820, opened 2026-06-24
fix(#MOBILESCROLL): prevent scrollTop jank on mobile during streaming DOM rebuild

akrhin:
## Problem

On mobile devices (iOS Safari, Android Chrome), when the assistant is streaming a response, the transcript viewport randomly jumps to the top. The user has to scroll back down manually each time — extremely frustrating.

**Root cause:** `.messages { overflow-anchor: none }` was set unconditionally. On mobile touch devices, the browser compositor can paint a frame with `scrollTop=0` between `innerHTML=''` (DOM wipe) and the subsequent `_restoreMessageScrollSnapshotSameFrame()` call. With `overflow-anchor:none`, the browser has no built-in mechanism to prevent this paint.

### 1. CSS (`static/style.css`)
- `.messages` now defaults to `overflow-anchor: auto`
- Desktop (mouse-driven devices) still get `overflow-anchor: none` via `@media (hover:hover) and (pointer:fine)` — preserving the existing behavior where tool/card inserts don't yank the viewport for a reading user
- Mobile/touch devices get `overflow-anchor: auto` — prevents the compositor from painting `scrollTop=0`

### 2. JS guard (`static/ui.js`)
- Added `window._fixMobileScrollJank()` — temporarily sets `overflowAnchor='auto'` on `#messages` before `innerHTML=''` and clears it one rAF later
- Called from `renderMessages()` just before the DOM wipe, protecting the gap between `innerHTML=''` and snapshot restore
- Media query `(hover:hover) and (pointer:fine)` ensures the guard only activates on touch devices

### 3. Cleanup (per PR review)
- Removed dead `_scrollAnchorRestore` variable (`window._fixMobileScrollJank()` returns `undefined`)
- Removed redundant `_fixMobileScrollJank()` call from `_doRender()` — CSS already gives `overflow-anchor:auto` on mobile via media query; desktop exits guard early. The call was a no-op allocating a rAF on every streaming tick
- Updated JSDoc to reflect actual usage

## Tests
- **test_issue1360_streaming_scroll_hardening.py**: Split the overflow-anchor test into two — desktop (`overflow-anchor:none`) and mobile (`overflow-anchor:auto`) variants — both passing
- **tests/test_mobile_scroll_jank.html**: Interactive visual test that simulates the scroll jank on a mobile-sized viewport. Can toggle fixes on/off to reproduce the bug and verify the fix

## Verification

All 14 existing scroll-related tests pass:

```
tests/test_issue1360_streaming_scroll_hardening.py ..... [PASS]
tests/test_2111_ios_pwa_bottom_scroll_stutter.py ...    [PASS]
tests/test_firefox_sidebar_scroll_stability.py ......   [PASS]
```

Fixes: #MOBILESCROLL

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -3762,9 +3762,6 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       _renderPending=false;
       // Guard: a pending setTimeout+rAF can outlive stream finalization.
       if(_streamFinalized) return;
-      // Mobile scroll-jank guard: enable overflow-anchor before DOM writes so
-      // the browser preserves scroll position during streaming content growth.
-      if(typeof window._fixMobileScrollJank==='function') window._fixMobileScrollJank();
       _lastRenderMs=performance.now();
       const parsed=_parseStreamState();
       _renderLiveThinking(parsed);
--- a/static/messages.js
+++ b/static/messages.js
@@ -3762,6 +3762,9 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       _renderPending=false;
       // Guard: a pending setTimeout+rAF can outlive stream finalization.
       if(_streamFinalized) return;
+      // Mobile scroll-jank guard: enable overflow-anchor before DOM writes so
+      // the browser preserves scroll position during streaming content growth.
+      if(typeof window._fixMobileScrollJank==='function') window._fixMobileScrollJank();
       _lastRenderMs=performance.now();
       const parsed=_parseStreamState();
       _renderLiveThinking(parsed);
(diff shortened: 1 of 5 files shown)


nesquena-hermes:
Superseded — this is the same mobile scrollTop-jank fix as your #4818, which **shipped in v0.51.617** (the `overflow-anchor` desktop-scoping + `_fixMobileScrollJank` are already on master). This branch is also based on an older master (its diff would revert unrelated shipped code like the #2814 Kanban toggle CSS), so closing in favor of the merged #4818. Thanks @akrhin — the fix is live! 🚀

PR #4822, opened 2026-06-24
perf(#4718): cache Claude Code transcript parses on the sidebar/profile-switch path (supersedes #4804)

nesquena-hermes:
## What this is

Phase 4 of the profile-switch performance work (#4662 / #4718): **make the cold `/api/sessions` build — which a profile switch triggers — substantially faster by caching the Claude Code transcript parse that actually dominates it.**

This **supersedes #4804**. That PR pre-warmed the sidebar cache from the switch handler, but @rodboev's browser-level profiling showed the pre-warm consistently loses the race against the client's follow-up GET, so cold `/api/sessions` still paid the full rebuild. I dug into *why the rebuild is 670ms in the first place* and found a different, much larger root cause.

## The real bottleneck (measured, not guessed)

I reproduced Rod's ~670ms cold `/api/sessions` on a seeded 250-session profile and captured the per-stage diagnostics. One stage dominates:

| Stage | Cold time |
|---|---|
| `get_cli_sessions` | **654 ms** |
| `visible_lineage_metadata` | 11 ms |
| `response_write` | 11 ms |
| everything else (`all_sessions`, state.db reads, dedupe, …) | <10 ms each |

The 654ms is **`get_claude_code_sessions()` → `_parse_claude_code_jsonl()`** reading and JSON-parsing every Claude Code transcript under `~/.claude/projects`, line by line (up to 200 files / ~130MB on my box), just to recover each session's title + message count.

Three things made it pathological:
1. **No caching at all** — every call re-reads + re-parses the full tree (a 2nd back-to-back call measured *1011ms*, even slower).
2. **Profile-blind waste** — `~/.claude/projects` is a single global directory, but the higher `_CLI_SESSIONS_CACHE` is keyed per active profile, so every profile switch re-parses the identical, unchanged transcripts under a fresh key.
3. **5s TTL** — the sidebar re-parses the whole tree every 5 seconds regardless.

This is why pre-warming couldn't help: a ~50–100ms head-start against a ~780ms build still leaves the client riding ~680ms as a follower.

## What this does

1. **Per-file parse cache.** `_parse_claude_code_jsonl_cached()` memoizes the parse result keyed by the file's `(path, mtime_ns, size)` (and `max_messages`). A warm build re-stats the files (~4ms for 200) instead of re-parsing them. Any genuine edit/append bumps `mtime_ns`/`size` and transparently invalidates **just that one file's** entry, so staleness is impossible without a re-parse. Thread-safe, LRU-bounded (1000), returns copied message lists so a caller can't corrupt a cached entry, and falls back to a direct parse if `stat()` fails.
2. **Hoist a loop-invariant.** `get_last_workspace()` (which stats `config.yaml` and probes the terminal cwd) was called once **per Claude Code row** — ~200 redundant stat chains per cold build. It's the same value for every row, so it's now resolved once. The redundant triple `path.stat()` per row is also de-duped.

## Measured impact

| Scenario | master | this PR |
|---|---|---|
| `get_claude_code_sessions()` cold → warm (in-process) | 729ms → 729ms (no cache) | 620ms → **76ms** |
| `/api/sessions` cold rebuild (curl, the switch scenario) | 670–880ms | **~100–250ms** |
| `/api/sessions` cold rebuild (**real browser** fetch, CLI-cache miss) | — | **139–207ms** |

**Output is byte-identical** to the previous behavior — verified by diffing the cached vs uncached `get_claude_code_sessions()` result over the same tree (200 rows, identical).

## Why this is the right architecture

The cache sits at the **per-file parse layer**, below all the profile-scoped caches, because the transcript files are global and rarely change between sidebar builds. That's what lets a profile switch (or a poll, or a TTL expiry) reuse the work. The residual warm cost (~100ms) is the irreducible directory-walk/`stat` floor for change-detection — squeezing it further would need a second, more fragile directory-signature cache that risks staleness for a small marginal gain, so I deliberately stopped here (per maintainer direction to nail the 80% with the most robust approach, not chase the last slice).

## Tests

- `tests/test_issue4718_claude_code_parse_cache.py` — 6 new tests: cache hit skips re-parse, invalidation on content/mtime change, returned lists are independent (no cache corruption), LRU bound is enforced, missing-file degrades to the empty-tuple contract, and an end-to-end check that a 2nd `get_claude_code_sessions()` adds zero re-parses.
- Existing Claude Code + CLI-cache + sidebar suites stay green.
- **Full suite: 10346 passed, 10 skipped, 1 xfailed, 2 xpassed, 0 failures.**

## Browser regression check

Loaded the patched server in a real browser against a 250-session seeded profile: app renders (not the wizard), sidebar lists all rows correctly, Claude Code rows still appear with the right title/count/source badge, and the in-browser `/api/sessions` timing on a CLI-cache miss is 139–207ms (vs master's 670–880ms).

## Review asks (@rodboev)

This is built directly on your profiling — thank you; your "the GET wins the race" finding on #4804 is exactly what redirected this. Two asks when you have a moment:

1. **Code review** of the parse cache + the `get_last_workspace()` hoist in `api/models.py`.
2. **Re-run your CDP benchmark loop** on your real 249-session profile (the cookie-aware switch → GET flow). I'd expect cold `/api/sessions` to drop from your measured ~670–790ms to roughly ~100–250ms, and — unlike #4804 — it should hold regardless of switch-vs-GET timing because there's no race; the win is that the expensive parse simply doesn't repeat.

Closes the perf half of #4718. Supersedes #4804 (will close that once this lands). Part of #4662 (phase 4).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Profile switching and sidebar loads are no longer dominated by re-parsing Claude Code transcripts.** On an installation with a large `~/.claude/projects` history, every `/api/sessions` build (which a profile switch triggers) re-read and JSON-parsed every Claude Code transcript file from scratch — line by line — just to recover each session's title and message count. On a 200-file / ~130MB tree that single step measured 650–1000ms and dominated the cold sidebar latency; because the directory is global but the higher session cache is keyed per active profile, it repeated in full on every switch, on the 5s CLI-cache expiry, and on every sidebar poll. The transcript parse is now memoized per file by its `(path, mtime, size)` signature, so a warm build re-stats the files (~4ms for 200) instead of re-parsing them, while any genuine edit or append transparently invalidates just that one file's entry. The redundant per-row `get_last_workspace()` call (which stats config + probes the terminal cwd) was also hoisted out of the Claude Code row loop. Measured cold-rebuild `/api/sessions` latency on a 250-session profile dropped from ~670–880ms to ~100–200ms. Output is byte-identical to the previous behavior. Thanks @rodboev for the browser-level profiling that isolated the real bottleneck. (#4718, #4662)
+
 ## [v0.51.625] — 2026-06-24 — Release WF (cron job model override no longer keeps the @provider: display prefix)
 
 ### Fixed
--- a/api/models.py
+++ b/api/models.py
@@ -51,10 +51,10 @@ _CLI_SESSIONS_CACHE = {}
 # CLI-cache expiry, and on every sidebar poll, because the higher CLI cache is
 # keyed per active profile while the underlying transcripts never change between
 # switches. This cache memoizes the EXPENSIVE per-file parse result keyed by the
-# file's (resolved path, mtime_ns, size); a warm sidebar build then re-stats the
+# file's (path, mtime_ns, size, ctime_ns); a warm sidebar build then re-stats the
 # files (~4ms for 200) instead of re-parsing them. Any external edit/append to a
-# transcript changes mtime_ns/size and transparently invalidates just that one
-# file's entry. Bounded so a pathological projects tree can't grow it unbounded.
+# transcript changes mtime_ns/size/ctime_ns and transparently invalidates just
+# that one file's entry. Bounded so a pathological projects tree can't grow it unbounded.
 _CLAUDE_CODE_PARSE_CACHE_LOCK = threading.Lock()
 _CLAUDE_CODE_PARSE_CACHE: "collections.OrderedDict[tuple, tuple]" = collections.OrderedDict()
 _CLAUDE_CODE_PARSE_CACHE_MAX = 1000
@@ -3872,21 +3872,26 @@ def _parse_claude_code_jsonl(path: Path, *, max_messages: int = CLAUDE_CODE_MAX_
 def _parse_claude_code_jsonl_cached(
     path: Path, *, max_messages: int = CLAUDE_CODE_MAX_MESSAGES_PER_FILE
 ) -> tuple[list[dict], str | None, float | None, float | None]:
-    """``_parse_claude_code_jsonl`` memoized by the file's (path, mtime_ns, size).
+    """``_parse_claude_code_jsonl`` memoized by the file's (path, mtime_ns, size, ctime_ns).
 
     The transcript files under ``~/.claude/projects`` are global and rarely
     change between sidebar builds, but parsing them dominates the cold
     /api/sessions latency (and repeats on every profile switch). Caching the
     parse result keyed by the file's stat signature collapses the warm cost to a
     single ``os.stat`` per file. A genuine append/edit bumps ``mtime_ns``/``size``
-    and misses the cache, so staleness is impossible without re-parsing.
+    /``ctime_ns`` and misses the cache, so staleness is impossible without
+    re-parsing.
 
     ``max_messages`` is part of the key so a caller asking for a different cap
     never reads a result truncated to a smaller one.
     """
     try:
         st = path.stat()
-        key = (str(path), st.st_mtime_ns, st.st_size, int(max_messages))
+        # Key on mtime_ns + size + ctime_ns: size is the strong discriminator for
+        # append-only JSONL (any write changes it), and ctime_ns guards the rare
+        # same-size, same-mtime in-place edit so a content change can never serve
+        # a stale parse. A spurious ctime bump only costs one harmless re-parse.
+        key = (str(path), st.st_mtime_ns, st.st_size, st.st_ctime_ns, int(max_messages))
     except OSError:
         # Can't stat -> fall back to a direct (uncached) parse; it will also
         # likely fail and return the empty tuple, matching prior behavior.
@@ -3988,7 +3993,13 @@ def get_claude_code_sessions(projects_dir: Path | str | None = None, *, max_file
         if not messages:
             continue
         sid = _claude_code_session_id(path)
-        if first_ts is None and last_ts is None:
+        # Match the truthiness fallback used in the assignments below: the old
+        # inline code was ``first_ts or last_ts or path.stat().st_mtime``, which
+        # also fell back to mtime for a falsy-but-not-None ``0.0`` timestamp
+        # (epoch-0 / 1970 transcripts). An identity (``is None``) guard would
+        # leave those rows with ``None`` instead of the file mtime, so use the
+        # same ``not`` test the assignments use to stay bug-for-bug compatible.
+        if not first_ts and not last_ts:
             try:
                 _mtime = path.stat().st_mtime
             except OSError:
--- a/api/models.py
+++ b/api/models.py
@@ -41,6 +41,24 @@ _CLI_SESSIONS_CACHE_TTL_SECONDS = 5.0
 _CLI_SESSIONS_CACHE_LOCK = threading.Lock()
 _CLI_SESSIONS_CACHE = {}
 
+# Per-file parse cache for Claude Code JSONL transcripts (#4718/#4662 phase 4).
+# ``~/.claude/projects`` is a GLOBAL, profile-independent directory, but the
+# sidebar re-derives every Claude Code row from scratch on each /api/sessions
+# build — fully re-reading and JSON-parsing up to CLAUDE_CODE_MAX_FILES
+# transcripts line-by-line (hundreds of MB) just to recover a title + message
+# count. That parse dominates the cold sidebar build (~650-1000ms measured on a
+# 200-file / ~130MB tree) and it repeats on every profile switch, on the 5s
+# CLI-cache expiry, and on every sidebar poll, because the higher CLI cache is
+# keyed per active profile while the underlying transcripts never change between
+# switches. This cache memoizes the EXPENSIVE per-file parse result keyed by the
+# file's (resolved path, mtime_ns, size); a warm sidebar build then re-stats the
+# files (~4ms for 200) instead of re-parsing them. Any external edit/append to a
+# transcript changes mtime_ns/size and transparently invalidates just that one
+# file's entry. Bounded so a pathological projects tree can't grow it unbounded.
+_CLAUDE_CODE_PARSE_CACHE_LOCK = threading.Lock()
+_CLAUDE_CODE_PARSE_CACHE: "collections.OrderedDict[tuple, tuple]" = collections.OrderedDict()
+_CLAUDE_CODE_PARSE_CACHE_MAX = 1000
+
 # ---------------------------------------------------------------------------
 # Stale temp-file cleanup
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
@rodboev — this is the follow-through on your #4804 profiling, and it's built directly on your finding. Two asks when you have a moment:

1. **Code review** of the per-file parse cache (`_parse_claude_code_jsonl_cached`) and the `get_last_workspace()` hoist in `api/models.py`.
2. **Re-run your CDP benchmark loop** on your real ~249-session profile (the cookie-aware switch → GET flow you used on #4804). I'd expect cold `/api/sessions` to drop from your measured ~670–790ms to roughly ~100–250ms — and unlike #4804 it should hold regardless of switch-vs-GET timing, because there's no race: the expensive transcript parse simply doesn't repeat (it's cached per file by mtime/size, invalidated only on a real edit).

If your numbers don't match mine I'd genuinely want to know — my repro is a 200-file / ~130MB `~/.claude/projects` tree, and the win scales with how much Claude Code history you have. Thanks again for catching that the pre-warm couldn't move the needle.

nesquena (approved):

Independent review of the revised Phase 4 of #4662/#4718: cache the Claude Code transcript parse that dominates the cold `/api/sessions` build (which a profile switch triggers). **Supersedes #4804** — that pre-warm consistently lost the race against the client's GET (per @rodboev's profiling); this attacks the actual ~654ms cost instead, so there's no race to lose.

### What this ships

`api/models.py` (+92/-6), `tests/test_issue4718_claude_code_parse_cache.py` (+149), `CHANGELOG.md` (+4). Agent-authored, co-authored with @rodboev (whose CDP profiling isolated the bottleneck). The agent requested code review + a CDP re-benchmark; my `event=APPROVE` is the code-review gate.

Fresh tarball: `~/.claude/projects` is **Claude Code's** (Anthropic CLI) global transcript directory — the webui only **reads** it to surface CC sessions in the sidebar. The cache memoizes those reads; it never writes the dir, and it's correctly **profile-blind** (the transcripts are global; profile scoping happens in the higher per-profile caches above this layer). No agent protocol/config/state surface; no cross-tool write. ✓

### Change 1 — per-file parse cache (`_parse_claude_code_jsonl_cached`)

Memoizes the expensive line-by-line JSONL parse keyed by `(str(path), st_mtime_ns, st_size, max_messages)`. I verified the correctness-critical properties:

- **Staleness is impossible without a re-parse.** Any append (size changes) or in-place edit (mtime changes) misses the key. I proved the subtle case with a harness: a **same-size in-place edit** still invalidates because `st_mtime_ns` changes — the edited content is seen on the next read. ✓
- **`max_messages` is in the key**, so a caller asking for a larger cap never reads a result truncated to a smaller one. ✓
- **Thread-safe:** all cache access is under `_CLAUDE_CODE_PARSE_CACHE_LOCK`; the parse runs **outside** the lock (a concurrent duplicate parse is harmless — re-checked under the lock, first writer kept); LRU eviction (`popitem(last=False)`) is under the lock. ✓
- **LRU-bounded** at 1000 entries (a pathological projects tree can't grow it unbounded). ✓
- **stat() failure → direct uncached parse** (preserves the prior empty-tuple contract for missing files — harness-confirmed). ✓

### Change 2 — hoist the loop-invariant + de-dup per-row stats

- `get_last_workspace()` ([workspace.py:376](api/workspace.py#L376)) takes **no arguments** → genuinely loop-invariant → resolved once instead of per CC row (~200 redundant config-stat + cwd-probe chains on a cold build). Byte-identical. ✓
- The `created_at`/`updated_at`/`last_message_at` rework is **byte-identical**: the old `first_ts or last_ts or path.stat().st_mtime` (3 separate `stat()` calls, only when both ts are falsy) becomes one `_mtime = path.stat().st_mtime` (computed only when both ts are None) fed into the same `or`-chains. Same values, `stat()` called once-when-needed instead of 3×. ✓

### Behavioural verification — independent harness (6/6 + sharing probe)

```
cached call returns equal result; returned list is a fresh object (list-level copy)
append invalidates cache (2→3 msgs seen)
SAME-SIZE in-place edit invalidates via mtime_ns (edited content seen)   ← the subtle staleness case
list-level mutation of the returned result does NOT corrupt the cache
missing file → empty messages (contract preserved)
[probe] per-message dicts ARE shared across calls (shallow copy)         ← the documented hazard, below
```

### The one thing to know — shallow copy shares the per-message dicts (verified safe today)

`_parse_claude_code_jsonl_cached` returns `list(messages)` — a fresh **list**, but the per-message **dicts are shared** with the cached entry (confirmed by the probe above). So the cache's integrity depends on callers treating message dicts as read-only. Before this PR, every parse returned fresh dicts, so this is a new sharing surface. I traced the consumers:

- **Sidebar build** (`get_claude_code_sessions`) uses only `_claude_code_title(messages,…)` (reads `.get('role')`/`.get('content')` — read-only) and `len(messages)`. Safe. ✓
- **Transcript view** (`get_claude_code_session_messages` → `get_cli_session_messages`): the CC-sid paths I traced are read-only — `9768` uses `len`/`msgs[-1].get`; the messaging-merge path doesn't fire for CC (`is_messaging_session` is False); windowing slices (new list) and redaction/serialization build fresh dicts (`_redact_value` is non-mutating). No current mutator found.

So it's correct **today** (and the full suite + byte-identical verification back that). I'm flagging it as a non-blocking robustness note (below) rather than a defect.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Warm sidebar build, transcripts unchanged | re-stat (~4ms) instead of re-parse (~620ms) ✅ |
| Transcript appended | size changes → cache miss → re-parse just that file ✅ |
| Same-size in-place edit | mtime_ns changes → cache miss → re-parse ✅ |
| Different `max_messages` cap | distinct key → no truncated-result mis-serve ✅ |
| Concurrent sidebar builds | lock-guarded; duplicate parse harmless; first writer kept ✅ |
| stat() fails / missing file | direct uncached parse → empty-tuple contract ✅ |
| 1000+ distinct transcripts | LRU-bounded eviction ✅ |
| Profile switch | global transcripts reused across profiles (cache below the per-profile layer) ✅ |
| CLI/agent writes a transcript | next read re-stats → mtime/size bump → re-parse ✅ |

### Tests

- `tests/test_issue4718_claude_code_parse_cache.py` (6) — **6/6** (hit skips re-parse, invalidation on content/mtime, returned lists independent, LRU bound, missing-file degrades, end-to-end zero-re-parse on 2nd call).
- Claude Code / CLI-session / sidebar suites — **75 passed, 2 skipped**.
- **Independent harness: 6/6** (staleness incl. same-size edit, list-mutation safety, missing-file, + the dict-sharing probe).
- **Full suite: 10218 passed / 133 skipped / 0 failures** in 269s (deselected the pre-existing darwin CRLF flake).
- `py_compile` clean on models.py.
- **CI on the PR head:** lint + browser-smoke green; test shards finishing — local full suite substitutes.

- **Shallow-copy dict sharing (robustness):** the returned message **dicts** are shared across cache reads (proven above). It's safe today — every current CC consumer I traced is read-only — but it's a new sharing surface, so a *future* consumer that mutates a message dict in place would corrupt the cache. If hardening is ever wanted, `copy.deepcopy(messages)` on return would be cheap here (the large `content` values are immutable strings → `deepcopy` shares them; only the small dict/list shells are duplicated, still far below a re-parse). A one-line code comment naming the read-only contract on the public `get_claude_code_session_messages` return would also lock it in. Not blocking.
- The ~100ms warm floor (directory walk + per-file `stat`) is correctly left as the robust stopping point — a directory-signature cache would risk staleness for a small gain.

### Recommendation

✅ **Approved clean (code review).** Parked at approval — ready for the release agent's merge/tag pipeline once @rodboev's CDP re-benchmark confirms the ~670ms→~100-250ms drop.

This is the right fix for the right bottleneck: cache the per-file transcript parse below all the profile-scoped caches, keyed by the file's `(mtime_ns, size)` so staleness is impossible without a re-parse (proven, including the same-size-edit case), thread-safe, LRU-bounded, with a byte-identical `get_last_workspace` hoist and stat-dedup. It correctly supersedes #4804's pre-warm (no race to lose). The only nuance is the shallow-copy dict sharing — verified safe against every current consumer, flagged as a cheap future-hardening, not a present defect. Ship (pending the CDP re-benchmark).

rodboev:
Tested locally on my real 249-session default profile (Windows, Python in-process via scheduled task). This is the same setup from the [#4804 profiling](https://github.com/nesquena/hermes-webui/pull/4804#issuecomment-4784937103) where the pre-warm consistently lost the race.

**Profile switch benchmark (5 rounds, cookie-aware, `sidebar_source=webui&exclude_hidden=1`):**

| Round | Direction | Switch+GET total | GET /sessions | Sessions |
|---|---|---|---|---|
| 1 | default → test-ep-sprint31 | 264ms | 49ms | 0 |
| 1 | test-ep-sprint31 → default | 89ms | 8ms | 249 |
| 2 | default → test-ep-sprint31 | 65ms | 5ms | 0 |
| 2 | test-ep-sprint31 → default | 51ms | 10ms | 249 |
| 3 | default → test-ep-sprint31 | 8ms | 4ms | 0 |
| 3 | test-ep-sprint31 → default | 109ms | 8ms | 249 |
| 4 | default → test-ep-sprint31 | 59ms | 5ms | 0 |
| 4 | test-ep-sprint31 → default | 212ms | 26ms | 249 |
| 5 | default → test-ep-sprint31 | 71ms | 67ms | 0 |
| 5 | test-ep-sprint31 → default | 46ms | 9ms | 249 |

**GET /sessions: min=4ms, max=67ms, avg=19ms**
**Switch+GET total: min=8ms, max=264ms, avg=98ms**

For reference, previous cold-rebuild times on this same profile were 670-880ms (our #4804 measurements). That's a ~4-10x improvement on the GET, and the switch+GET total is consistently under 300ms.

The full-scan path (without `sidebar_source=webui`) also benefits: 127ms for 251 sessions vs the 670-880ms baseline. The parse cache works regardless of whether the sidebar optimization is active.

Cache verification: two rapid GETs on the same profile measure 123ms then 6ms, confirming the cache hit path.

The approach is fundamentally better than #4804's pre-warm because there's no race to win; the expensive parse simply doesn't repeat. Reviewed the code, the `(path, mtime_ns, size)` key is solid for invalidation and the LRU bound prevents unbounded growth.

nesquena-hermes:
Thanks @rodboev — that's exactly the validation I was hoping for, and your numbers land right where the design predicted. GET `/api/sessions` avg **19ms** (min 4 / max 67) against the 670-880ms #4804 baseline is the ~4-10x I sketched, and the two-rapid-GET 123ms→6ms confirms the cache-hit path cleanly. The fact that the win holds on the full-scan path too (127ms for 251 sessions, no `sidebar_source=webui`) is the key proof that this isn't a sidebar-specific optimization — it attacks the actual cost center.

I re-read the cache under `origin/master...HEAD` and agree with your read on the invalidation key. A couple of things I want to confirm are correct since they're the parts that would bite under concurrency:

The stat-signature key and LRU bound are right (`api/models.py:3889`, eviction at the `popitem(last=False)`):
```python
key = (str(path), st.st_mtime_ns, st.st_size, int(max_messages))
...
while len(_CLAUDE_CODE_PARSE_CACHE) > _CLAUDE_CODE_PARSE_CACHE_MAX:
    _CLAUDE_CODE_PARSE_CACHE.popitem(last=False)
```
Including `max_messages` in the key is the subtle-but-correct call — without it a caller asking for a smaller cap could read a result truncated by an earlier larger-cap caller (or vice-versa). The re-check-under-lock after the parse (3909-3917) avoids the double-parse stampede without holding the lock across the expensive parse itself, which is the right tradeoff.

One thing worth a sentence in your benchmark notes: the cache returns `list(messages)` (a shallow copy) but the per-message dicts are shared with the cached entry. The code comments this as "treated as read-only by all current callers" — that's true today, but it's the one invariant a future caller could quietly break (mutating a message dict in place would corrupt the cached parse for every other profile). Not a change request, just the thing to grep for if a weird cross-profile transcript bug ever shows up.

On `mtime_ns` resolution: your Windows result is reassuring here, since `st_mtime_ns` on NTFS has 100ns granularity — fine for human-paced transcript edits, and `st_size` is a second discriminator anyway, so a same-nanosecond same-size content swap (effectively impossible for append-only JSONL) is the only theoretical miss.

This supersedes #4804 the right way — no race to win because the parse simply doesn't repeat. With your real-profile numbers confirming the synthetic 200-file repro, I'm satisfied on the perf side. Nice closeout on the #4718 thread.

nesquena-hermes:
## Gate complete — rebased + dual-gate findings fixed, ready to ship 🟢

Ran the full authoritative gate on this self-built PR (it already had nesquena APPROVED + @rodboev's hands-on validation). Summary for whoever ships it:

**Rebased** onto fresh `origin/master` (was CONFLICTING — CHANGELOG-only conflict, resolved: kept all released sections, entry under `[Unreleased]`). Code diff verified byte-identical to the pre-rebase head; the `api/models.py` delta vs the old head is purely master's own advancement, not a change to the cache logic.

**Full suite:** 10370 passed, 10 skipped, 1 xfailed, 2 xpassed, 0 failures (pre-fix baseline) + the parse-cache module green after the fixes (9 tests).

**Codex + Opus both converged independently on two findings** (a reproduced finding from both = high signal), both now fixed in commit `90431c6`:

1. **Cache-key staleness hardening** — added `st_ctime_ns` to the `(path, mtime_ns, size, max_messages)` key. `st_size` is already the strong discriminator for append-only JSONL (Opus confirmed a stale read is effectively impossible), but `ctime_ns` is cheap defense-in-depth against a same-size, same-mtime in-place edit. A spurious ctime bump only ever costs one harmless re-parse, never a stale read — correct trade for a cache.

2. **Epoch-0 timestamp fallback (bug-for-bug fix)** — the row-builder guard used identity (`first_ts is None and last_ts is None`) but the field assignment uses truthiness (`first_ts or last_ts or _mtime`). They disagreed on a falsy `0.0` timestamp (epoch-0/1970 transcripts), leaving `created_at/updated_at/last_message_at = None` instead of the file mtime the pre-cache inline code fell back to. Both gates reproduced this. Guard now uses the same `not first_ts and not last_ts` test. Real-world impact was nil (requires 1970 timestamps), but it's now exactly equivalent to the old behavior.

**Mutation-isolation (the cache's cardinal risk):** both gates independently traced every consumer of the shared per-message dicts (`get_cli_session_messages` dispatches Claude Code sids to the cached path → all 7 `routes.py` callers) and confirmed **all read-only** — the import-extend path copies via `dict(message)` before any pop, and `Session.save()` serializes via `json.dumps`. No in-place mutation reaches the cache.

**Added 3 regression tests:** epoch-0 mtime fallback, the shared-dict read-only contract (pins the load-bearing invariant so a future mutating caller is a conscious change), and same-size+mtime ctime invalidation.

Lock correctness (double-checked populate, LRU `popitem(last=False)`, `move_to_end`) and the OSError stat-fail fallback were both verified sound. **No remaining blockers — clean to merge once CI re-greens on `90431c6`.** Will close #4842 (the 100% CPU re-escalation, whose root cause is exactly this `get_cli_sessions` parse) and #4718/#4662 on ship.

nesquena-hermes:
Shipped in **v0.51.640** (Release WU) — closes #4718/#4662 (and the #4842 CPU re-escalation). Nathan greenlit moving forward.

Rebased onto fresh master + re-ran the full gate: Codex SAFE TO SHIP (no regression post-rebase; cache key staleness-proof with mtime_ns+size+ctime_ns+max_messages, byte-identical output, thread-safe, no profile-isolation break), Opus SHIP IT (all 6 axes clean), full suite green (the lone test_workspace_upload failure is a pre-existing order-dependent flake, not in this diff, passes in isolation). Human review satisfied (nesquena APPROVED + rodboev validated: 19ms warm vs 670–880ms baseline on a 249-session profile). Verified the cache in the served bundle on prod. Thanks @rodboev for the profiling that isolated the bottleneck.

PR #4823, opened 2026-06-24
Fix compaction marker writeback and code redaction

### LINKED ISSUE #4821: Compaction markers and redaction can corrupt WebUI restored context
## Summary

A Discord cron debugging flow exposed two Hermes WebUI / Hermes Agent integration regressions:

1. Synthetic context-compaction/reference markers can enter the normal WebUI session/context/display path as ordinary messages, including as `role: user`, and can be duplicated across `messages` and `context_messages`.
2. API redaction can damage code-shaped tool output while masking credentials, causing the displayed code to look syntactically broken even when the underlying file is not.

These are not primarily Discord collector bugs. The collector debugging session made the WebUI/Agent integration failure visible.

## Evidence

Observed in local WebUI session evidence:

- WebUI review/debug session `61b1cafc6e50`
  - Shows repeated redacted Discord-token-shaped tool output in the transcript.
  - The visible transcript led the agent toward treating redacted/display-damaged code as real source damage.
- Debug session file `20260624_142734_4d3bff`
  - `messages` contains a context-compaction marker as a `user` row.
  - `context_messages` also contains a compaction marker as a `user` row.
  - Multiple tool/assistant rows contain redacted Discord-token-shaped code output.
- Investigated session file `20260624_092443_693a48`
  - The opening rows include two same-content compaction markers: one `assistant`, one `user`.
  - The same duplicated marker pattern appears in `context_messages`.
  - Tool output examples show redaction damaging code-shaped lines around `DISCORD_BOT_TOKEN=...` and Discord `Authorization: Bot ...` header construction.

Representative bad outcomes:

- `[CONTEXT COMPACTION — REFERENCE ONLY] ...` appears as a normal `user` message and can be treated as active user intent on restore/search/context reconstruction.
- Code-shaped output such as env key checks and Discord authorization header construction is displayed with the credential masked but neighboring quote/comma/parenthesis structure damaged.

## Expected behavior

- Internal compaction/reference markers may remain model-facing recovery/reference material, but they must not be appended as normal visible transcript turns or become active `user` turns.
- Context dedupe should canonicalize duplicated compaction markers so an accidental `user` marker cannot survive as the next active user instruction.
- Redaction should hide real secret values while preserving code syntax: key literals, quotes, commas, brackets, parentheses, variable names, and non-secret code delimiters must remain intact.
- Code literals that mention a secret key name without a secret value, such as an env key prefix check, should remain unchanged.

## Impact

The agent can be misled by WebUI-provided/restored context and by redacted tool output:

- It may resume stale compaction-summary instructions as if the user had just issued them.
- It may attempt to repair code that is only broken in the redacted display layer, creating real churn around a fake bug.

## Notes

A narrow fix should target the WebUI session writeback/context merge and WebUI redaction integration layers, not the Discord cron collector itself.

franksong2702:
## Thinking Path

- Hermes WebUI keeps separate visible transcript and model-facing `context_messages` layers during compaction and restore.
- The Discord cron debug sessions showed synthetic context-compaction/reference text entering those layers as normal messages, including as `role: user`.
- The same sessions showed code-shaped tool output being redacted into syntactically broken-looking code, which misled the agent into fixing a fake source bug.
- The narrow fix belongs at the WebUI writeback/context-merge and WebUI redaction integration layers, not in the Discord collector.

Fixes #4821.

## What Changed

- Filter synthetic compaction/reference markers out of `_merge_display_messages_after_agent_result()` so they do not become visible transcript turns.
- Canonicalize duplicated compaction markers in `_deduplicate_context_messages()` so only one assistant-side reference marker survives and an accidental `user` marker cannot become active user intent.
- Update stale merge docstring wording to match the new invariant: compacted results append the current turn onward, not visible marker turns.
- Harden WebUI redaction around code-shaped strings:
  - preserve env key literals that do not contain a real secret value,
  - redact `Authorization: Bot ...` values without breaking quotes/commas/brackets,
  - restore obvious code env-key delimiters after the upstream agent redactor runs.
- Add/adjust regression coverage for the new invariant and for the previously missed transcript-compaction suite.

## Why It Matters

Without this, restored/searchable/displayed WebUI sessions can feed internal reference text back to the agent as if the user just said it, and redacted tool output can make valid code look broken. In long debugging sessions that combination can send the agent into a repair loop around stale or fake problems.

## Verification

- `./scripts/test.sh tests/test_issue1217_transcript_compaction.py -q` → `31 passed`
- `./scripts/test.sh tests/test_issue1217_transcript_compaction.py tests/test_compaction_marker_writeback.py -q` → `33 passed`
- `./scripts/test.sh tests/test_issue1217_transcript_compaction.py tests/test_compaction_marker_writeback.py tests/test_issue2028_compression_anchor_helpers.py tests/test_security_redaction.py tests/test_partial_bloat_fix.py tests/test_issue2914_truncation_watermark.py tests/test_issue765_streaming_persistence.py -q` → `159 passed, 1 skipped` (`hermes-agent` unavailable in test env for one agent-dependent test)
- `git diff --check` → clean
- AST parse of touched Python files → clean

## Risks / Follow-ups

- This does not rewrite historical session JSON files; it prevents future writeback/context merge/display pollution and canonicalizes context during the relevant merge path.
- Whole-file `ruff check` still reports pre-existing backlog in touched files; the changed code was validated through targeted regression tests and syntax checks.
- The fix intentionally leaves Discord collector behavior out of scope.

## Contract Routing

Task type: Runtime/session state and API redaction bug fix.

Touched areas:
- WebUI session display writeback and model-context dedupe (`api/streaming.py`)
- API response redaction (`api/helpers.py`)
- Regression tests and changelog

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`

State layer / invariant:
- Visible transcript: internal compaction/reference material must not become a visible active turn.
- Model context: duplicated marker rows must not survive as `role: user` active intent.
- API redaction boundary: secret values are masked without corrupting code structure.

Evidence:
- Regression tests for display merge, context dedupe, code-literal redaction, and the existing issue1217 transcript-compaction suite.

## Model Used

AI-assisted via Hermes Agent using OpenAI Codex `gpt-5.5`, with local file/terminal/GitHub tooling.

### DIFF
--- a/api/compression_anchor.py
+++ b/api/compression_anchor.py
@@ -64,9 +64,10 @@ def is_context_compression_marker(message):
         message.get("content", ""),
         part_types={"text", "input_text", "output_text"},
     ).lower().lstrip()
+    synthetic_unbracketed_marker = bool(message.get("_compressed_summary"))
     return (
         text.startswith("[context compaction")
-        or text.startswith("context compaction")
+        or (synthetic_unbracketed_marker and text.startswith("context compaction"))
         or text.startswith("[your active task list was preserved across context compression]")
         or text.startswith("[session arc summary")
     )
--- a/api/helpers.py
+++ b/api/helpers.py
@@ -269,7 +269,7 @@ def _build_redact_fn():
     )
     _ENV_RE = _re.compile(
         r"([A-Z0-9_]{0,50}(?:API_?KEY|TOKEN|SECRET|PASSWORD|PASSWD|CREDENTIAL|AUTH)[A-Z0-9_]{0,50})"
-        r"\s*=\s*(['\"]?)([^\s'\"\]\),;]+)\2"
+        r"\s*=\s*(['\"]?)(\S+)\2"
     )
 
     _PRIVKEY_RE = _re.compile(
@@ -288,20 +288,47 @@ def _build_redact_fn():
     _CODE_ENV_KEY_LITERAL_RE = _re.compile(
         r"([A-Z0-9_]{0,50}(?:API_?KEY|TOKEN|SECRET|PASSWORD|PASSWD|CREDENTIAL|AUTH)[A-Z0-9_]{0,50}=)([\"'][)\]:,]+|[)\]:,]+)"
     )
+    _ENV_KEY_PREFIX_RE = _re.compile(
+        r"([A-Z0-9_]{0,50}(?:API_?KEY|TOKEN|SECRET|PASSWORD|PASSWD|CREDENTIAL|AUTH)[A-Z0-9_]{0,50}=)"
+    )
+    _REDACTED_ENV_VALUE_RE = _re.compile(
+        r"(?:\*{3,}|[A-Za-z0-9][A-Za-z0-9_.:/+-]{0,32}\.\.\.[A-Za-z0-9_.:/+-]{1,16})"
+    )
 
     def _restore_code_env_key_literals(original: str, redacted: str) -> str:
         if not isinstance(original, str) or not isinstance(redacted, str):
             return redacted
-        restored = redacted
-        for match in _CODE_ENV_KEY_LITERAL_RE.finditer(original):
+        literal_occurrences: dict[tuple[str, int], str] = {}
+        original_counts: dict[str, int] = {}
+        for match in _ENV_KEY_PREFIX_RE.finditer(original):
             key_prefix = match.group(1)
-            literal_suffix = match.group(2)
-            restored = restored.replace(
-                f"{key_prefix}***",
-                f"{key_prefix}{literal_suffix}",
-                1,
-            )
-        return restored
+            occurrence = original_counts.get(key_prefix, 0)
+            original_counts[key_prefix] = occurrence + 1
+            literal_match = _CODE_ENV_KEY_LITERAL_RE.match(original, match.start())
+            if literal_match:
+                literal_occurrences[(key_prefix, occurrence)] = literal_match.group(2)
+        if not literal_occurrences:
+            return redacted
+        redacted_counts: dict[str, int] = {}
+        pieces = []
+        last = 0
+        for match in _ENV_KEY_PREFIX_RE.finditer(redacted):
+            key_prefix = match.group(1)
+            occurrence = redacted_counts.get(key_prefix, 0)
+            redacted_counts[key_prefix] = occurrence + 1
+            literal_suffix = literal_occurrences.get((key_prefix, occurrence))
+            if literal_suffix is None:
+                continue
+            value_match = _REDACTED_ENV_VALUE_RE.match(redacted, match.end())
+            if not value_match:
+                continue
+            pieces.append(redacted[last:value_match.start()])
+            pieces.append(literal_suffix)
+            last = value_match.end()
+        if not pieces:
+            return redacted
+        pieces.append(redacted[last:])
+        return "".join(pieces)
 
     def _fallback_redact(text: str) -> str:
         if not isinstance(text, str) or not text:
--- a/api/helpers.py
+++ b/api/helpers.py
@@ -263,11 +263,15 @@ def _build_redact_fn():
         r"|brv_[A-Za-z0-9]{10,}"          # ByteRover API key
         r")(?![A-Za-z0-9_-])"
     )
-    _AUTH_HDR_RE = _re.compile(r"(Authorization:\s*Bearer\s+)(\S+)", _re.IGNORECASE)
+    _AUTH_HDR_RE = _re.compile(
+        r"""(Authorization:\s*(?:Bearer|Bot)\s+)([^\s'",\]\)]+)""",
+        _re.IGNORECASE,
+    )
     _ENV_RE = _re.compile(
         r"([A-Z0-9_]{0,50}(?:API_?KEY|TOKEN|SECRET|PASSWORD|PASSWD|CREDENTIAL|AUTH)[A-Z0-9_]{0,50})"
-        r"\s*=\s*(['\"]?)(\S+)\2"
+        r"\s*=\s*(['\"]?)([^\s'\"\]\),;]+)\2"
     )
+
     _PRIVKEY_RE = _re.compile(
         r"-----BEGIN[A-Z ]*PRIVATE KEY-----[\s\S]*?-----END[A-Z ]*PRIVATE KEY-----"
     )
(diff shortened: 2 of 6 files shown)


nesquena-hermes:
Read the diff at this PR's HEAD (`api/streaming.py` + `api/helpers.py`), the two new test files, and cross-checked against `origin/master` and the frontend render path. This is well-targeted — the fix sits exactly at the writeback/context-merge and redaction-integration layers #4821 calls out, not in the collector. A few confirmations and one redaction edge note.

## Display-merge change is consistent with the frontend
Stripping markers from `previous_display` and skipping them in the candidate loop:
```python
previous_display = [
    m for m in list(previous_display or [])
    if not _is_context_compression_marker(m)
]
...
for msg in candidates:
    if _is_context_compression_marker(msg):
        continue
```
This is safe with respect to the visible compression card because the UI never renders markers as transcript bubbles in the first place — `static/ui.js:11551` already does `if(_isContextCompactionMessage(m)){ continue; }`, and the settled card is driven by `compression_anchor_summary` + `_latestCompressionReferenceMessage` (ui.js:11273, ui.js:11296), not by the marker surviving in `s.messages`. And the manual-compress path (`api/routes.py:19073-19081`) writes the summary to `s.compression_anchor_summary` metadata, not as a `[CONTEXT COMPACTION]` row. So removing markers from the persisted visible delta aligns the stored transcript with what the renderer already does — no card regression.

## Context dedupe role-canonicalization is the right call
Forcing the surviving marker to `assistant` in `_deduplicate_context_messages()` directly closes the "marker-as-active-user-intent" hole from the issue's session `20260624_092443_693a48`. Worth confirming the marker key truncation (`[:500]`) can't collide two genuinely different markers into one — but since these are synthetic and near-identical by construction, 500 chars is plenty.

## Redaction: one edge in `_restore_code_env_key_literals`
The `.replace(f"{key_prefix}***", ..., 1)` loop restores one occurrence per `finditer` match on the original. If the same key prefix (e.g. `DISCORD_BOT_TOKEN=`) appears twice in one blob — once as a real `KEY=secretvalue` and once as a code literal `KEY=")` — the count-1 replace plus iteration order could restore the wrong site. The new tests cover the single-occurrence case (`test_redaction_preserves_secret_key_literals_in_code_without_values`) and the bot-header case, but not a mixed real-secret-plus-code-literal-same-key blob. A 3-line test with both forms in one string would lock down ordering. Not a blocker.

The `_env_replacement` guard `if not any(ch.isalnum() for ch in value)` correctly skips pure-punctuation captures (the `")` / `]:` code tails), and narrowing the value group to `[^\s'\"\]\),;]+` stops masking from eating closing quotes/brackets — that's the structural-damage root cause from the issue, fixed cleanly. `ast.parse(redacted)` in the new test is a good structural assertion.

Solid fix. The mixed-key redaction test is the only thing I'd add before merge.

nesquena-hermes:
## Review — real direction, but 2 must-fixes before merge (dual-gate Codex + Opus, both reproduced)

Thanks @franksong2702 — both bugs are real and the fixes target the right layers. The compaction-marker mis-role attack is genuinely closed (`_deduplicate_context_messages` force-sets non-assistant markers to `role='assistant'` on a deepcopy, and the visible-merge filters markers out entirely). But the dual gate reproduced two issues that need fixing first — one is a **security under-redaction regression**, one is **silent user-data loss**.

### 🔴 MUST-FIX 1 — tightened `_ENV_RE` now leaks the tail of secrets containing `, ; ) ] ' "` (`api/helpers.py`)
Changing the value class from `\S+` to `[^\s'"\]\),;]+` makes `,` `;` `)` `]` `'` `"` *internal* stop chars, so the capture truncates and the tail leaks. Reproduced:
- `PASSWORD=pa,ssw0rd` → masks only up to the comma, leaks `,ssw0rd`
- `TOKEN="abc"` style values can be left unmasked in the fallback-only path
Known-prefix tokens (`ghp_`, `sk-`, `AKIA`…) are still caught by `_CRED_RE`/the agent redactor, but arbitrary high-entropy `NAME=value` secrets that embed those chars now under-redact where `\S+` previously masked the whole value. (Base64 `+ / =` is fine — not stop chars.)
**Fix:** keep code-mangling protection without weakening the secret mask — e.g. retain a broad `\S+` capture and skip only captures with **no alphanumeric** char (you already have `_env_replacement` doing the no-alphanumeric skip — so the tightened character class is redundant *and* harmful; revert the class to `\S+` and rely on the alphanumeric-skip + `_restore_code_env_key_literals` for the code case). Add fallback-only tests (force the `agent.redact` import to fail) for `KEY=val,with,commas`, `KEY="quoted"`, `KEY=ends)`.

### 🔴 MUST-FIX 2 — genuine user message starting with marker-like text is dropped / role-flipped (`api/streaming.py` + `api/compression_anchor.py:67`)
`is_context_compression_marker` matches an **unbracketed** `context compaction` prefix, so a real user turn like *"context compaction is broken; explain what happened"* is dropped from the visible transcript **and** role-cast `user`→`assistant` in model context (verified). That's silent user-data loss on the destructive paths.
**Fix:** for the destructive drop / role-canonicalization paths, require explicit synthetic metadata (e.g. `_compressed_summary` / the marker's internal flag), not a content-prefix alone — and never drop or role-cast the **current user turn** by content prefix. The bracketed forms (`[context compaction…`, `[your active task list…]`) are safe; it's the unbracketed branch that's unsafe for destructive use.

### 🟡 NICE-TO-HAVE — `_restore_code_env_key_literals` `.replace(…, 1)` is not occurrence-aware
With `real TOKEN=<secret>` + `code TOKEN=]` in the same text (both upstream-masked to `TOKEN=***`), the `.replace(…,1)` rewrites the *first* (real) mask's trailing punctuation. **No secret is disclosed** (it only splices punctuation, never secret bytes — Opus verified), so this is cosmetic, but worth making position-aware so it edits the literal occurrence, not the first.

### Gate status
- Codex: SHIP-ONLY-WITH-FIXES (3 findings, all reproduced)
- Opus: SHIP-WITH-FIXES (corroborated; confirmed no real-secret *disclosure*, but the under-redaction tail-leak + data-loss are real)
- The `"authorization: bot ***"` marker + `_AUTH_HDR_RE` `(?:Bearer|Bot)` change is correct and good (now masks Discord `Authorization: Bot …`).
- Suite: green in CI terms (the one failing `test_api_sessions_list_redacts_titles` in my run is a worktree-only conftest path artifact, not a redaction failure).

Once MUST-FIX 1 + 2 are in (with the fallback-only redaction tests + a "genuine user message with marker-prefix is preserved" test), I'll re-gate and ship. Really solid root-causing on both bugs.

franksong2702:
Addressed the requested fixes.

What changed:
- Rebased the branch onto current origin/master and removed the contributor `CHANGELOG.md` diff from the final PR tree.
- Restored fallback env-value matching to a broad non-space capture so values containing comma/semicolon/paren/brackets/quotes do not leak tails; the no-alphanumeric guard still preserves code punctuation literals.
- Tightened destructive compaction-marker detection: unbracketed `context compaction...` text is treated as a marker only when it carries internal metadata such as `_compressed_summary`, so genuine user turns with that prefix are preserved and not role-cast.
- Made `_restore_code_env_key_literals` occurrence-aware per env key and able to restore code literals after both `***` and longer `abc...xyz` agent masks, without restoring a real secret occurrence with the same key.

Verification:
- `python3 -m py_compile api/helpers.py api/compression_anchor.py api/streaming.py`
- `git diff --check`
- `./scripts/test.sh tests/test_compaction_marker_writeback.py tests/test_security_redaction.py::test_redaction_preserves_secret_key_literals_in_code_without_values tests/test_security_redaction.py::test_redaction_masks_authorization_bot_value_without_breaking_code_structure tests/test_security_redaction.py::test_fallback_env_redaction_masks_values_with_delimiters tests/test_security_redaction.py::test_agent_redaction_restore_is_occurrence_aware_for_code_key_literals tests/test_issue2028_compression_anchor_helpers.py::test_context_compression_marker_detection_is_prefix_and_role_scoped -q` (11 passed)
- `./scripts/test.sh tests/test_compaction_marker_writeback.py tests/test_issue1217_transcript_compaction.py tests/test_security_redaction.py tests/test_issue2028_compression_anchor_helpers.py tests/test_webui_gateway_chat_backend.py -q` (144 passed, 1 skipped)
- `./scripts/test.sh tests/test_context_message_dedup.py tests/test_session_save_mode.py tests/test_stale_user_context_contamination.py -q` (65 passed)

nesquena-hermes:
Shipped in **v0.51.667** (Release XW, just deployed) — thanks @franksong2702! 🎉 Closes #4821. Both bugs fixed: compaction/reference markers no longer leak into the visible transcript as a fake user turn (and a genuine user message starting with 'context compaction' is no longer misclassified), and code-shaped tool output is no longer mangled by redaction.

Re-gate after your re-push confirmed both prior MUST-FIXes are genuinely addressed:
- MUST-FIX 1 (secret-tail leak): `_ENV_RE` value class back to broad `\S+` — Codex direct-probed `API_KEY`/`TOKEN="…"`/`PASSWORD=…,…` masked in BOTH the agent-redactor AND fallback-only paths, no tail leak.
- MUST-FIX 2 (user-data loss): unbracketed marker match now gated on `_compressed_summary`, so a genuine user turn isn't dropped/role-flipped.
Dual gate: Codex SAFE + Opus SHIP IT (verified the dedupe deep-copies, no shared-object mutation). Suite 10628. Verified on prod.

PR #4828, opened 2026-06-24
Prevent final answer echoes from entering Worklog reasoning

### LINKED ISSUE #4827: Final answer text can be journaled as reasoning Worklog content
## Bug

Final Answer prose can be emitted and persisted as a `reasoning` event, causing the Compact Worklog / Thinking surface to show text that belongs only in the final assistant answer.

## Evidence

Observed in a local WebUI session:

- Session id: `61b1cafc6e50`
- Run journal: `~/.hermes/webui/sessions/_run_journal/61b1cafc6e50/eb580937867f43629d065354c72ae29d.jsonl`
- Session file: `~/.hermes/webui/sessions/61b1cafc6e50.json`

The run journal has a late `reasoning` event:

- `seq=3637`
- `type=reasoning`
- `payload.text` length: 500
- that text is the prefix of the final assistant message later stored in the session transcript
- the same text already appeared in the visible assistant token stream before the `reasoning` event

The session transcript then stores the final answer as the assistant message, so the same final-answer prose exists in both places: Worklog/reasoning event stream and final assistant content.

## Expected

Final Answer正文 should be ordinary assistant prose only. It should not enter `reasoning` / Compact Worklog event flow.

## Actual

A runtime can mirror already-streamed final answer text through the reasoning callback near stream end. The WebUI's existing echo guard only suppressed suffix/tail echoes, so a long prefix of the already-streamed final answer was still emitted as `reasoning` and journaled.

## Scope

This is a narrow live-to-final boundary bug under #3400:

- Do not rewrite historical session or run journal data.
- Do not change the event schema.
- Do not hide genuine reasoning/worklog content.
- Do not mix with compaction marker/redaction or pending user turn fixes.

franksong2702:
## Thinking Path

- Hermes WebUI's long-running reply model keeps live process/reasoning activity separate from the settled final answer.
- The supplied run journal showed `seq=3637` as a `reasoning` event whose payload was the first 500 characters of the final assistant answer.
- The same text had already streamed through normal assistant `token` events, and the final assistant message was later persisted in the session transcript.
- That points to the WebUI streaming bridge accepting a runtime callback misclassification, not a final render/merge inversion.
- The narrow fix is to suppress substantial reasoning chunks that are already present in the visible assistant stream before they reach SSE or the run journal.

## What Changed

- Extended `_is_visible_output_echo()` in `api/streaming.py`:
  - keeps the existing suffix/tail echo suppression for short progress echoes;
  - adds a conservative long-chunk containment guard for already-streamed assistant output;
  - only applies the broader check to substantial chunks, so ordinary short reasoning phrases are not hidden.
- Added a regression in `tests/test_issue_progress_echo_dedupe.py` that simulates:
  - final answer emitted through `stream_delta_callback`;
  - the first 500 characters mirrored later through `reasoning_callback`;
  - live SSE queue contains the token but no leaked reasoning;
  - run journal contains no leaked reasoning event, so replay has nothing wrong to rebuild;
  - done/session payload keeps the final answer as assistant content and does not merge the leaked prefix into `reasoning` / `reasoning_content`.
- Updated `CHANGELOG.md`.

Fixes #4827.
Refs #3400.

## Why It Matters

Final Answer text must remain ordinary assistant prose. If it is also journaled as `reasoning`, users see final-answer content inside Compact Worklog / Thinking after live streaming or journal replay, making it look like Worklog leaked into the answer. This preserves the live-to-final boundary without rewriting historical sessions or changing the event schema.

## Verification

- `./scripts/test.sh tests/test_issue_progress_echo_dedupe.py::test_final_answer_prefix_reasoning_echo_is_not_journaled_or_merged -q`
  - `1 passed in 3.82s`
- `./scripts/test.sh tests/test_issue_progress_echo_dedupe.py tests/test_run_journal.py tests/test_run_journal_routes.py -q`
  - before rebase: `27 passed in 4.42s`
  - after lint fix: `27 passed in 4.28s`
  - after rebase to latest `origin/master`: `27 passed in 5.24s`
- `./.venv/bin/python scripts/ruff_lint.py --diff origin/master`
  - passed with no output

## Contract Routing

Task type: runtime / streaming / replay boundary bug fix.

Touched areas:
- WebUI streaming bridge (`api/streaming.py`)
- run-journal-facing regression coverage (`tests/test_issue_progress_echo_dedupe.py`)
- changelog

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/rfcs/README.md`

State layer / invariant:
- Mutates live SSE emission before run-journal persistence.
- Preserves the invariant that final assistant prose must not become Worklog/reasoning activity, while genuine reasoning remains visible and persisted.

Evidence needed before done:
- live queue regression;
- run journal regression;
- final done/session merge regression.

## Risks / Follow-ups

- If a runtime sends a short answer phrase as reasoning, this remains conservative and may still show it; that is intentional to avoid hiding genuine short reasoning.
- This is a WebUI guard against a runtime misclassification. If Hermes Agent can classify final-answer chunks correctly upstream, that would reduce the need for this defensive filter, but WebUI still needs the guard for replay durability.
- Historical sessions/run journals are not rewritten; old leaked events remain historical data.

## Model Used

AI-assisted via Hermes WebUI using OpenAI Codex GPT-5.5 in the current session. Tool use included local git/worktree commands, source inspection, targeted tests via `./scripts/test.sh`, ruff diff lint, GitHub CLI issue/PR workflow, and live GitHub read-back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Final answers no longer duplicate into the Compact Worklog as reasoning when a runtime mirrors already-streamed answer text through the reasoning callback.** WebUI now suppresses substantial reasoning chunks that are already present in the visible assistant stream before they reach live SSE or the run journal, while preserving genuine reasoning and short progress-tail echo handling. This keeps live rendering, journal replay, and settled session reloads from showing Final Answer prose inside Worklog/Thinking.
+
 ## [v0.51.617] — 2026-06-24 — Release VX (fix mobile scroll-jank to top during streaming)
 
 ### Fixed
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -6384,10 +6384,22 @@ def _run_agent_streaming(
                 candidate = _compact_for_echo_compare(text)
                 if not candidate:
                     return False
+                visible_output = STREAM_PARTIAL_TEXT.get(stream_id, '')
                 visible_tail = _compact_for_echo_compare(
-                    STREAM_PARTIAL_TEXT.get(stream_id, '')[-max(len(str(text)) * 2, 512):]
+                    visible_output[-max(len(str(text)) * 2, 512):]
                 )
-                return bool(visible_tail and visible_tail.endswith(candidate))
+                if visible_tail and visible_tail.endswith(candidate):
+                    return True
+                # Some runtimes can report a prefix of the already-streamed final
+                # answer through reasoning after visible output has completed. That
+                # prefix is not a tail echo, so catch only substantial chunks that
+                # are already present in the visible assistant stream. Short text
+                # stays on the stricter suffix path to avoid hiding genuine
+                # reasoning that happens to reuse an answer phrase.
+                if len(candidate) < 80:
+                    return False
+                visible_compact = _compact_for_echo_compare(visible_output)
+                return bool(visible_compact and candidate in visible_compact)
 
             def _strip_reasoning_output_echo(text: str) -> bool:
                 nonlocal _reasoning_segments
--- a/tests/test_issue_progress_echo_dedupe.py
+++ b/tests/test_issue_progress_echo_dedupe.py
@@ -134,11 +134,11 @@ def test_visible_progress_token_reasoning_and_interim_are_deduped(cleanup_test_s
         "credential_pool": None,
     }
     runtime_payload["api_" + "key"] = "***"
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.619** 🚀 — thanks @franksong2702. Final-answer prose can no longer leak into the Compact Worklog/Thinking surface as a reasoning echo (suppressed before live SSE + the run journal, conservatively so genuine reasoning is preserved). Codex SAFE + full suite 10344. Closes #4827.

PR #4832, opened 2026-06-24
fix(kanban): bound column card list height

jkobject:
## Summary
- Cap Kanban column card-list height with `max-height:min(68vh,720px)`.
- Keep overflow scrolling inside `.kanban-column-body` with `overflow-y:auto`, `overscroll-behavior:contain`, and stable scrollbar gutter.
- Add a source-level regression test so the Tasks/Kanban board keeps rendering cards inside the scrollable list container.

## Test plan
- `./scripts/test.sh tests/test_kanban_board_ui.py -q` → `1 passed`
- `.venv/bin/ruff check tests/test_kanban_board_ui.py` → passed
- `node --check static/panels.js` → passed
- `git diff --check` → passed
- `.venv/bin/python tests/browser_smoke.py` → passed (`/`, `/#settings`, `/#sessions` no console errors)


nesquena-hermes:
Shipped in **v0.51.624** 🚀 — thanks @jkobject. Kanban columns with many cards now cap at min(68vh,720px) and scroll internally (contained overscroll + stable gutter) instead of stretching the board. Full suite 10359.

PR #4834, opened 2026-06-24
Fix WebUI reconciliation after Desktop session continuation

### LINKED ISSUE #4833: WebUI falls out of sync after Hermes Desktop continues a WebUI-created session
## Problem

When the official Hermes Desktop App continues a WebUI-created Hermes Agent session, it settles new messages into `~/.hermes/state.db`, but WebUI can keep serving stale settled state from the WebUI sidecar/session-list cache.

Observed production shape:

- WebUI sidecar counts can lag behind `state.db` counts for the same session.
- `GET /api/session?session_id=...&messages=1` can see the newer `state.db` tail.
- `GET /api/sessions` can still show stale `message_count`, `updated_at`, and `last_message_at` from the sidecar/cache.
- Some merge paths risk returning sidecar + state.db prefix duplicates instead of one reconciled transcript.

## Expected contract

This is settled-state reconciliation only:

1. After Desktop finishes a turn, `/api/sessions` should show the updated count/time for WebUI-origin sessions.
2. `/api/session?messages=1` should return the complete settled transcript without duplicate expansion.
3. The next WebUI turn should use Desktop-appended settled messages as model context.
4. WebUI should not write duplicated historical rows back to the sidecar on the next save.

Out of scope:

- live token mirroring from Desktop into WebUI
- simultaneous-send conflict resolution beyond the existing active-run guard
- Desktop App changes

franksong2702:
## Thinking Path

- Reproduced the settled-state mismatch with a regression test that first caches `/api/sessions` from a WebUI sidecar, then simulates the official Hermes Desktop App appending settled rows to the same Hermes Agent `state.db` session.
- Traced `/api/sessions` through the sidebar cache and `all_sessions(include_lineage_metadata=False)` path.
- Found two reconciliation gaps: the sidebar state-db override only carried source/title metadata (not settled message count/time), and the session-list cache stamp ignored `state.db` entirely when external/CLI sessions were hidden.
- After CI exposed the shard-2 hang, traced the cache follower path: once `/api/sessions` started observing `state.db` stamps, a stamp mismatch could discard the stale payload before a follower request could use it, forcing that follower to wait behind a slow owner rebuild.
- After CI exposed a neighboring shard-1 race, tightened the wait-stage test so it deterministically waits until the follower is actually blocked before releasing the owner.
- After live 8787 audit found settled worker sessions readable via `/api/session` but missing from `/api/sessions`, traced the index path: it pruned stale rows and overlaid active in-memory sessions, but did not recover persisted sidecar JSON files missing from a non-empty `_index.json` once the runtime overlay was gone.
- Kept the scope to settled-state reconciliation and sidebar/cache behavior only; no Desktop App change, no live stream mirroring, no multi-client conflict behavior change, and no #4836 compression recovery work.

## What Changed

- Batch-read `state.db.messages` count and max timestamp in the existing sidebar state-db override helper.
- Overlay newer WebUI-origin `state.db` count/time onto stale WebUI sidecar rows when the DB has strictly grown beyond the sidecar.
- Recover messageful persisted WebUI sidecar sessions that are missing from a non-empty sidebar `_index.json`, using metadata-only loads, then write those recovered rows back to the index.
- Make `/api/sessions` cache stamps observe `state.db` even when CLI/external sessions are hidden, while preserving the existing active-stream hold-down behavior.
- Preserve the existing hot-path invariant for sidebar cache followers: if a stale payload exists, a follower can reuse it while the owner performs a slow rebuild, even when the source stamp changed.
- Harden cache follower tests with `try/finally` cleanup so failed assertions release the owner builder and join threads instead of orphaning them until CI timeout.
- Added/updated regressions for:
  - WebUI-origin `/api/sessions` metadata updating after Desktop-settled `state.db` append.
  - `/api/session?messages=1` returning one reconciled transcript instead of duplicating a sidecar prefix.
  - the next WebUI turn saving a single reconciled history that includes externally appended settled messages.
  - cache followers reusing stale payloads during slow rebuilds after a source-stamp change.
  - a settled messageful sidecar that `Session.load()` can read still appearing in sidebar rows even when a non-empty `_index.json` missed it and the in-memory runtime overlay is gone.
- Updated `CHANGELOG.md`.

## Why It Matters

Users can continue a WebUI-created session from the official Hermes Desktop App and then return to WebUI. After the Desktop turn settles, WebUI now shows the complete final transcript in the sidebar/detail view and uses that settled history as context for the next WebUI message, without duplicating old rows in the sidecar.

It also keeps completed worker sessions discoverable in the sidebar after the active runtime overlay disappears, as long as their sidecar JSON remains readable by session detail.

Fixes #4833.

## Verification

- `./scripts/test.sh tests/test_streaming_session_sidebar.py::test_messageful_session_missing_from_index_is_recovered_in_sidebar -q` — 1 passed.
- `./scripts/test.sh tests/test_session_sidebar_cache.py::test_session_list_cache_follower_reuses_stale_payload_during_slow_rebuild -q` — 1 passed.
- `./scripts/test.sh tests/test_session_sidebar_cache.py::test_session_list_cache_follower_wait_stage_when_rebuild_inflight tests/test_session_sidebar_cache.py::test_session_list_cache_follower_reuses_stale_payload_during_slow_rebuild -q` — 2 passed.
- `./scripts/test.sh tests/test_streaming_session_sidebar.py tests/test_session_sidebar_cache.py tests/test_webui_state_db_reconciliation.py tests/test_webui_state_db_context_reconciliation.py tests/test_session_db_sidecar_reconciliation.py tests/test_issue4766_sidebar_source_pushdown.py tests/test_cli_sessions_cache_fingerprint.py tests/test_issue3930_source_filter_pushdown.py tests/test_issue1494_state_db_fd_leak.py -q` — 90 passed.
- `.venv/bin/python -m py_compile api/models.py api/routes.py tests/test_streaming_session_sidebar.py tests/test_session_sidebar_cache.py tests/test_webui_state_db_reconciliation.py tests/test_webui_state_db_context_reconciliation.py` — passed.
- `git diff --check` — passed.

## Risks / Follow-ups

- `/api/sessions` now observes `state.db` for WebUI-only sidebar views. The existing active-stream hold-down remains in place to avoid per-token cache churn during WebUI streaming.
- Recovering missing sidecar index rows uses metadata-only loads and only triggers for persisted session IDs missing from the non-empty index; the recovered rows then flow through the existing visibility filters.
- Stale follower fallback is intentionally preserved only as a fallback while an owner rebuild is in-flight; the owner still refreshes and stores the updated payload.
- This does not mirror Desktop live tokens into WebUI and does not change simultaneous-send conflict semantics.
- Overlay is limited to WebUI-origin `state.db` rows and only when state.db has grown beyond the sidecar, to avoid resurrecting restamped/replayed histories.
- Manual compression recovery remains separate in #4836 and is not part of this PR.

## Model Used

OpenAI Codex GPT-5.5


nesquena-hermes:
Read the diff at HEAD against `origin/master` (`api/models.py:3226-3367`, `api/routes.py:1761-1780`) plus the new `tests/test_webui_state_db_reconciliation.py`. The reconciliation logic is sound and I can see it's already staged for release in #4834's release PR. One narrow edge in the overlay guard worth noting for a possible follow-up.

## The overlay guard
The count/time overlay in `_apply_sidebar_state_db_override_metadata()` is gated by:
```python
if state_count > current_count and (state_last <= 0 or state_last > current_last):
    session['message_count'] = state_count
    ...
```
with `current_last = max(last_message_at, updated_at)` of the sidecar row. The `state_last <= 0` branch correctly handles a `messages` table without a `timestamp` column (`messages_has_timestamp` False → `last_message_at` NULL), and the strict `state_count > current_count` + `state_last > current_last` pair is the right conservative anti-resurrection rule the PR body promises.

## The edge
Because `current_last` folds in the sidecar's `updated_at`, a metadata-only sidecar write that lands *after* the Desktop append (e.g. a title rename or source-tag rewrite bumping `updated_at`) makes `current_last > state_last`, so the count growth is silently dropped even though `messages` genuinely grew. The new test uses strictly increasing message timestamps (1002/1003 vs sidecar 1001) so it never exercises a `updated_at > state_last` case. This is consistent with "favor not-resurrecting," so not a blocker — but a regression that appends Desktop rows *then* touches sidecar metadata would lock in the conservative tradeoff. If that's intended, a one-line comment on the guard would save a future reader the head-scratch; if not, gating on message-count growth alone (dropping `updated_at` from `current_last`) would catch it.

The no-duplicate-prefix merge assertion in `test_api_session_full_load_does_not_duplicate_state_db_prefix` is the right shape and directly covers the #4833 symptom. Nice work.

nesquena-hermes:
Thanks @franksong2702 — the Desktop-continuation reconciliation logic itself is sound (Codex SAFE + Opus SAFE: the new sidebar overlay is read-only, content-free, and structurally can't cause the ghost/message-loss class). But it can't ship yet: it introduces a **parallel test-isolation regression** that the full sharded suite reproduces.

### The failure
`tests/test_session_sidebar_cache.py::test_session_list_cache_follower_wait_stage_when_rebuild_inflight` **fails under the full parallel (xdist) suite** on this branch — reproduced twice locally, and CI fails it on two independent shards (`test (3.12, 1)` + `test (3.13, 1)`). It **passes** in isolation and in a small slice, and it **passes on clean master under the same full xdist run** (10364 passed) — so the branch is the cause, not a pre-existing flake.

### Likely cause
The change removes the `if not cache_show_cli_sessions:` guard in `_session_list_cache_source_stamp` so the state.db signal is always folded into the cache stamp (the "refresh WebUI rows even when CLI sessions are hidden" fix). That makes the source-stamp read state.db on every cache check, which appears to perturb the timing/global-state the single-flight follower-wait test depends on when other tests run concurrently. The test uses a `release.wait()` + `_StageRecorder` single-flight rebuild handshake, so it's sensitive to cross-test cache-state bleed under parallelism.

### Fix direction
Make the new always-on state.db stamp read isolation-safe under the test suite, or update the single-flight cache tests to reset/scope the shared `_SESSIONS_CACHE` state they rely on so the new stamp behavior doesn't leak across parallel workers. Reproduce with the **full** suite under xdist (not `-p no:xdist`, and not the cache file alone — it only surfaces with the whole suite running in parallel):
```
PYTHONPATH=. python -m pytest tests/ -q
```

Everything else checks out — both gates SAFE on the reconciliation logic, 31 reconciliation tests green, serial full suite green. It's specifically the parallel-isolation interaction with the sidebar-cache single-flight test. Re-gate when the full xdist suite is green. Keeping `changes-requested`.

franksong2702:
Updated this branch against current master and resolved the remaining DIRTY state. The conflict was limited to CHANGELOG.md; the session reconciliation code paths auto-merged cleanly.

Local verification:
- ./scripts/test.sh tests/test_session_sidebar_cache.py tests/test_streaming_session_sidebar.py tests/test_webui_state_db_context_reconciliation.py tests/test_webui_state_db_reconciliation.py -q
- 55 passed

GitHub CI is now green on 0d6ffcbcb. I also ran the full local suite. It is not clean on this machine due to failures that reproduce on clean origin/master as well, so I am relying on the PR-specific local slice plus GitHub CI for the final gate. Ready for re-review.

franksong2702:
Addressed the review items on this branch.

What changed:
- Rebased/updated the PR branch onto current origin/master and removed the contributor CHANGELOG diff from the final PR tree.
- Replaced the .isdigit() actual_message_count coercion with try/int(), plus a regression for numeric values such as 5.0.
- Isolated the session-list cache tests by clearing cache + inflight rebuild state before/after each test and pinning the pure single-flight thread tests to a stable source stamp.
- Added a comment documenting the conservative updated_at/state.db anti-resurrection guard.

Verification:
- git diff --check
- ./scripts/test.sh tests/test_session_sidebar_cache.py tests/test_webui_state_db_reconciliation.py::test_sidebar_state_db_overlay_preserves_numeric_actual_count tests/test_webui_state_db_reconciliation.py::test_api_sessions_overlays_webui_state_db_summary_after_desktop_append -q (18 passed)
- ./scripts/test.sh tests/test_session_sidebar_cache.py tests/test_streaming_session_sidebar.py tests/test_webui_state_db_reconciliation.py tests/test_webui_state_db_context_reconciliation.py -q (56 passed)
- ./scripts/test.sh tests/ -q completed the full xdist run; the previously reported session-sidebar-cache failure did not reproduce. Local full-suite result was 10470 passed / 106 skipped / 4 failed, with the remaining failures in unchanged provider/TLS/agent-import tests: test_issue1426_openrouter_free_tier_live_fetch.py::test_free_tier_cap_prevents_picker_drowning, test_minimax_provider.py::test_minimax_cn_detected_from_os_environ, test_tls_support.py::TestTLSEndToEnd::test_tls_startup_failure_fallback_to_http, and test_v050259_sessiondb_fd_leak.py::test_session_db_close_is_idempotent.

nesquena-hermes:
Shipped in **v0.51.668** (Release XX, just deployed) — thanks @franksong2702! Fixes #4833. WebUI-origin sessions now reconcile correctly after the Desktop app continues them (sidebar refreshes from settled state.db count/time even with CLI sessions hidden; metadata-only overlay, strictly-grew, active-stream hold-down preserved). The prior bounce (parallel test-isolation regression) is resolved — the follower-wait test now passes under the full xdist suite. Gate: Codex SAFE + Opus SHIP IT (test-isolation fix confirmed test-only; overlay read-only/no-resurrection), suite 10631. Verified on prod.

PR #4841, opened 2026-06-24
Release v0.51.626 — keep sidebar in sync after Desktop app continues a session (#4834)

nesquena-hermes:
## Release v0.51.626 — keep the sidebar in sync after the Desktop app continues a session (#4834)

Ships **#4834** (by @franksong2702). Reconciliation fix on the WebUI↔state.db sidebar path.

### What it does
When the official Hermes Desktop App appends settled rows to the same Hermes Agent `state.db` session that was started in the WebUI, the WebUI conversation sidebar showed a stale message count. It now refreshes the WebUI-origin row's count/timestamp from `state.db` even while external/CLI sessions are hidden. The no-duplicate-prefix merge and single-reconciled-transcript-on-next-turn behaviors (pre-existing) are now covered by regression tests for this Desktop-continuation scenario.

### Gate (deep — this is the crown-jewel reconciliation family where the #4772 P0 lives)
- **Codex** (GPT-5.5): SAFE TO SHIP — sidebar override applies only count/timestamp/source-tag (no hidden-session content leak), no message-loss/duplication across the reconciliation hunks; 31 reconciliation tests pass.
- **Opus** (claude-opus-4-8): SAFE TO SHIP — confirmed the genuinely-new surface is "read-only, webui-gated, content-free, monotonic-up — structurally cannot cause the ghost/message-loss class because it never writes message content and never touches the merge/save path." (Flagged a cosmetic count-vs-watermark display divergence — display-only, no data loss.)
- **Full suite:** 10366 passed, 0 failed.

CHANGELOG wording tightened per the Opus review to reflect that the dedup/writeback guarantees are *tested* here, not introduced.

Closes #4834.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **WebUI-created sessions stay in sync after the official Hermes Desktop App continues the same Hermes Agent session.** The conversation sidebar now refreshes WebUI-origin rows from settled `state.db` message counts/timestamps even when external/CLI sessions are hidden, full session loads avoid duplicating the sidecar prefix when merging `state.db`, and the next WebUI turn saves a single reconciled transcript instead of re-writing duplicated history.
+
 ## [v0.51.623] — 2026-06-24 — Release WD (iOS portrait: opening a conversation lands at the bottom)
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


nesquena-hermes:
Closing — source PR #4834 bounced for a parallel test-isolation regression (full xdist suite reds test_session_list_cache_follower_wait_stage_when_rebuild_inflight). Will re-stage once #4834's CI is green.

Issue #1567, opened 2026-05-03
bug(models): Providers screen shows full Nous Portal catalog, model picker shows only 4 — endpoint disagreement


nesquena-hermes:
## Summary

A user (Deor) reported in Discord (#report-bugs, May 03 2026, 14:15 PT) that on the latest version (v0.50.286, which shipped the #1538 fix) the **Providers settings screen and the in-conversation model picker disagree about which Nous Portal models exist**. Three screenshots:

1. Providers card: **"Nous Portal — 396 models · OAuth"**
2. Model picker dropdown (in conversation): only the **4 hardcoded curated entries** under "Nous Portal" (Claude Opus 4.6, Claude Sonnet 4.6, GPT-5.4 Mini, Gemini 3.1 Pro Preview)
3. Same picker scrolled — confirmed cut off at 4, then "OPENCODE GO" header

Reporter quote:

> Providers screen now shows a lot more models for Nous portal. They dont seem to show up in the model select lists though.

#1538's fix (a) added a live-fetch branch in `_build_available_models_uncached()` for `pid == "nous"` and (b) added a parallel live-fetch branch in `api/providers.py:list_providers()` (~line 399). Both call `hermes_cli.models.provider_model_ids("nous")`. They are returning different results in production.

## Root cause

Two endpoints, two different gates on whether Nous models surface:

### `/api/providers` (`api/providers.py:300-433`)
- Iterates ALL provider IDs in `_PROVIDER_DISPLAY ∪ _OAUTH_PROVIDERS` regardless of authentication state
- For `pid == "nous"`, **unconditionally** live-fetches `provider_model_ids("nous")` (lines 399-413), falling back to the 4-entry static list only when hermes_cli is unavailable or raises
- Result: Nous card always renders with the live catalog count

### `/api/models` (`api/config.py:_build_available_models_uncached`, ~line 2122-2280)
- Only iterates providers in `detected_providers`
- Nous gets added to `detected_providers` ONLY when `hermes_cli.models.list_available_providers()` returns it with `authenticated: True` (lines 1759-1768) — there is no env-var path for Nous (it's OAuth-only)
- IF `detected_providers` includes nous, the live-fetch branch (line 2243) is reached and should return ~30 entries
[...]

PR #4844, opened 2026-06-24
feat: interactive model manager for custom providers

luandnh:
## Thinking Path

- Custom providers can fetch hundreds of models from their /v1/models endpoint
- All fetched models appear in the chat model picker, making it hard to find the right one
- The backend already supports filtering via custom_providers[].models in config.yaml
- But there was no UI to manage which models are active — users had to edit config.yaml manually
- This PR adds an interactive model manager in the provider settings card

## What Changed

**Backend:**
- Added `set_custom_provider_models()` in `api/providers.py` — updates config.yaml's
  custom_providers[].models field with an allowlist of model IDs
- Added `POST /api/providers/models` route — accepts `{provider, models}` list

**Frontend:**
- Custom provider card now shows interactive model management UI instead of a static hint
- **Active models** section — pills with × to remove from allowlist
- **Available (from refresh)** section — pills with + to add, populated after Refresh click
- Custom model name input — type + Enter to add models not in the live list
- Save button persists selection to config.yaml and refreshes the model dropdown
- CSS styles for remove-tag (hover turns red) and add-tag (accent color)

## Why It Matters

Without this UI, users with custom providers that expose many models (ZenMux, aggregator
gateways, etc.) had to manually edit config.yaml to curate the list. Now they can:
1. Click Refresh to discover all models
2. Click + on the ones they want (or × to remove)
3. Type any missing model name manually
4. Save — only selected models appear in the model picker

## Verification

- All 29 tests in test_custom_providers_in_panel.py + test_issue1106_custom_providers_models.py pass
- Python syntax checked (py_compile)
- JavaScript syntax checked (Node)

## Risks / Follow-ups

- Low risk: the model manager only activates for p.is_custom providers
- Existing behavior unchanged for non-custom providers (OpenAI, Anthropic, etc.)
- When models list is empty, all live models appear (same as before)

### DIFF
--- a/api/providers.py
+++ b/api/providers.py
@@ -2858,11 +2858,21 @@ def set_custom_provider_models(provider_id: str, models: list[str]) -> dict[str,
                 if not isinstance(cp, dict):
                     continue
                 if _custom_provider_name_matches(provider_id, cp.get("name")):
-                    if models:
-                        cp["models"] = list(models)
+                    # Preserve the original format: if models was a dict, save
+                    # as dict keys so per-model metadata (context_length, etc.)
+                    # is not lost. If models was a list (or absent), save as list.
+                    existing_models = cp.get("models")
+                    if isinstance(existing_models, dict):
+                        if models:
+                            # Rebuild dict with empty metadata for each model
+                            cp["models"] = {m: {} for m in models}
+                        else:
+                            cp.pop("models", None)
                     else:
-                        # Empty list → remove the key so all live models appear
-                        cp.pop("models", None)
+                        if models:
+                            cp["models"] = list(models)
+                        else:
+                            cp.pop("models", None)
                     found = True
                     break
 
--- a/api/providers.py
+++ b/api/providers.py
@@ -2813,3 +2813,71 @@ def _clean_provider_key_from_config(provider_id: str) -> None:
             reload_config()
     except Exception:
         logger.exception("Failed to clean provider key from config.yaml for %s", provider_id)
+
+
+def set_custom_provider_models(provider_id: str, models: list[str]) -> dict[str, Any]:
+    """Set the active model list for a custom provider.
+
+    Updates the ``models`` field on the matching ``custom_providers[]`` entry in
+    config.yaml so the model picker only shows the selected models.
+
+    Args:
+        provider_id: Canonical custom provider slug (e.g. ``custom:9router``).
+        models: List of model ID strings to allowlist. Empty list removes the
+                ``models`` field, causing all live models to appear.
+
+    Returns:
+        A status dict with ``ok`` and optionally ``error``.
+    """
+    from api.config import _cfg_lock
+    import api.config as _config
+
+    try:
+        config_path = _config._get_config_path()
+    except Exception as exc:
+        return {"ok": False, "error": f"Cannot resolve config path: {exc}"}
+
+    if not config_path.exists():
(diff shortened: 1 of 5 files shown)


nesquena-hermes (changes requested):
Thanks for this — the interactive model manager is a genuinely nice UX upgrade over hand-editing `config.yaml`, and the backend is solid (write is correctly guarded under `_cfg_lock`, `encoding="utf-8"` throughout, proper `Exception` handling + cache invalidation, and the follow-up commit's dict-format preservation + non-array payload guard are good catches).

Requesting changes on one **confirmed functional defect** in the headline "discover → add multiple models" flow.

## Variable-scope split: discovered models get wiped after the first add

In `static/panels.js` → `_buildProviderCard`, the model-manager state and renderer are declared **inside** the `else if(p.is_custom){ … }` block:

- `function _renderAvailTags(models){…}` — line ~8903
- `let _availModelList = [];` — line ~8930
- block closes at line ~8992

…but the custom-provider **Refresh** handler is built **outside** that block (`refreshBtn.onclick = p.is_custom ? …`, line ~9041) and references both names:

```js
_availModelList = [...new Set(clean)];   // ~line 9065
_renderAvailTags(_availModelList);
```

panels.js loads as a classic non-module `<script>` (sloppy mode), so this doesn't throw — `_renderAvailTags` is reachable via Annex B function-declaration hoisting, but the **assignment to `_availModelList` creates a global `window._availModelList`**, which is a *different binding* from the block-scoped `let`.

Consequence:
1. Refresh works once — it sets the global and renders the discovered pills. ✓
2. The instant the user clicks `+` / `×` on any pill (or types + Enter), the **in-block** handlers call `_renderAvailTags(_availModelList)` using the **block-scoped** list, which is still `[]` → `availWrap.innerHTML = ''` clears everything and **all remaining discovered models vanish**. The user has to hit Refresh again to add each subsequent model.

So the "Refresh, then click + to add several models" path is effectively a one-add-per-refresh no-op. Manual add (type + Enter) and Save both work fully — only the discover-multi path is affected. (Greptile's review didn't flag this; it's frontend-only.)

### Fix

Lift the model-manager state + renderers to a single shared scope so the Refresh handler and the in-block `+`/`×`/Enter handlers mutate **one** binding. Simplest surgical option: move the custom-provider branch of the `refreshBtn.onclick` assignment **inside** the `else if(p.is_custom){ … }` block (before line ~8992) so it closes over the block-scoped `let _availModelList` / `_renderAvailTags`. Either that, or declare `_usedModelList`, `_availModelList`, `_renderUsedTags`, `_renderAvailTags` at the top of `_buildProviderCard` (function scope) and reference them from both places.

### Test coverage

The updated `tests/test_custom_providers_in_panel.py` change is a static source-grep (asserts the new strings exist in panels.js) — it can't catch this. Worth adding:
- a backend pytest for `set_custom_provider_models()` (write list, write `[]` → key removed, unknown provider → `ok:False`, and the dict-format-preservation path the follow-up commit added);
- a route test for `POST /api/providers/models` (missing `provider` → bad, non-array `models` → bad, valid → ok);
- ideally a jsdom/DOM test of the discover→add-multiple flow that would have caught the scoping bug.

Minor/optional: the diff also removes the `#1567` "+N more" rationale comment block (above the non-custom model list) while keeping that code — unrelated to this feature, fine to restore.

Once the scope fix lands the discover-multi flow should work end to end. Thanks again.

PR #4851, opened 2026-06-24
fix(#4752): show tool result instead of call args in collapsed card preview

### LINKED ISSUE #4752: enhancement(transparent-stream/tool-cards): collapsed preview shows the call INPUT (args/target), but the old preview showed the RESULT — restore (or richen) the result-bearing one-liner
enhancement(transparent-stream/tool-cards): collapsed preview now shows the call's INPUT (args/target), but the old preview showed the RESULT — restore (or richen) the result-bearing one-liner

## Summary

Follow-up to **#4658** (collapsed-row inline preview restored in v0.51.574, PR #4661). The preview is back, but a community follow-up (#report-bugs, 2026-06-23) flags it as **"restored, but in a less complete / more limited way"** than before.

Comparing the two screenshots, the regression that #4658 fixed restored a preview *sourced from the call's INPUT/target* — but the **old** preview was *sourced from the call's RESULT/output*. So the one-liner is technically back, but it carries different (and for some tools, less useful or empty) information.

## The actual delta

**Old (richer — preview = the result/output the call returned):**

old — result-bearing previews

- `search_files {"total_count": 0}`
- `read_file {"content": "1|\"\"\"Async HTTP client for Nginx Proxy Manager API…"}`
- `skill_view {"success": true, "name": "opencode-review-agents", "description": "…"}`
- `terminal {"author":{"id":"…","login":"ranklancer",…},"baseRefName":"main"…}`

You could read what each call **came back with** without expanding it.

**New (limited — preview = the input args/target the call was given):**

new — arg/target-only previews

- `search_files Tavily|401` (the query, not the hit count)
- `terminal grep -o '^[A-Z0-9_]*=' /opt/data/.env` (the command, not its output)
- `session_search` → **blank** (browse-mode call has no query arg → empty preview)

You see what each call **was asked to do**, but not what it produced.

## Why this happened — it's deliberate, not accidental

This is a **design-intent tension**, not a simple bug, so it's worth a maintainer call rather than a silent revert:

1. `69072ac34` "fix: keep collapsed tool previews **quiet**" (May 31) deliberately moved `buildToolCard`'s collapsed preview *off* the result snippet and onto an **argument summary** (regression-guarded by `tests/test_tool_card_preview_summary.py`).
2. The #4658 fix (`64484c980`, PR #4661) faithfully honored that "quiet" intent — its new `_transparentToolSummary(tc)` (`static/ui.js` ~L8216) is built from the call's **TARGET** (`_toolVisibleTargetLabel` → `_toolTargetLabel`: path / command / query / skill name) and the commit body explicitly says **"never the raw result JSON (that stays in the expandable detail body)."**

So #4658 restored exactly the *quiet, args-based* preview by design. the reporter is asking for the **older result-bearing preview** to come back — which directly conflicts with the "keep collapsed previews quiet" decision. That's why this is filed as `enhancement,ux` (a deliberate-behavior change request), not as a reopen of #4658 (whose stated goal — "a non-blank preview that says what the call is" — *is* met).

## Two concrete weaknesses in the current quiet preview

Independent of the result-vs-input philosophy, the args-based summary has two gaps worth fixing regardless:

1. **Arg-less calls render a completely blank preview.** `_toolTargetLabel` returns `''` for a call with no path/command/query/skill (e.g. `session_search` in browse mode, or any no-target tool) → `_transparentToolSummary` returns `''` → the row is bare again, the exact symptom #4658 set out to kill. For these, the **result** is the only meaningful one-liner (`session_search` → "3 sessions" / first hit title).
2. **For read/search/list tools the result IS the signal.** `search_files {"total_count": 0}` (found nothing) vs the query `Tavily|401` carry very different information at a glance — the count is what you actually want when skimming.

## Suggested direction (maintainer call)

Pick one — recommendation at the end:

- **A. Revert to result-based preview.** Source the collapsed one-liner from the output snippet again. Maximally matches the old screenshot, but re-opens the "previews are noisy" concern that `69072ac34` was created to solve, and would need that test updated.
- **B. Hybrid: target + a short result hint (recommended).** Keep the quiet target label as the lead, append a compact, *bounded* result hint where it's cheap and high-signal — counts for search/list (`search_files  Tavily|401 · 0 matches`), status/exit for terminal, ok/err for writes — and fall back to the result snippet only when there's **no** target (fixing the `session_search` blank). This restores "understand the call without expanding" without dumping raw JSON back into the collapsed row.
- **C. Minimal: just fix the blank-preview fallback.** When `_transparentToolSummary` would return `''`, fall back to a short result snippet so no row is ever bare. Smallest change; doesn't address the read/search "count is the signal" point.

**Recommendation: B.** It honors the original "keep collapsed previews quiet" intent (no raw result-JSON dump) while giving back the at-a-glance signal the reporter is missing, and it closes the arg-less blank-row gap as a side effect. A is the literal "make it like before" but re-introduces the noise problem on purpose; C is a safe floor if B is judged too much surface.

## Suspected code
- `static/ui.js` `_transparentToolSummary(tc)` (~L8216) — the quiet summary builder added by #4658.
- `static/ui.js` `_toolVisibleTargetLabel` / `_toolTargetLabel` (~L12384 / ~L12393) — target extraction (returns `''` for arg-less calls → blank preview).
- Guard test to update if behavior changes: `tests/test_tool_card_preview_summary.py` + `tests/test_issue4658_transparent_collapsed_preview.py` (currently asserts the preview is target-based and *not* result JSON — any result-bearing change must reconcile with it).

## Relationship to #4658
#4658 is **CLOSED/fixed correctly** for its stated symptom (bare rows → non-blank preview). This is the **content-quality delta** the reporter raised *after* that fix shipped — distinct request, cross-linking.

## Repro
1. Open a session with tool calls in Transparent Stream view (works on live + persisted/old sessions).
2. Note collapsed rows show the call's **input** (path/command/query) — and arg-less calls like a browse-mode `session_search` show **nothing**.
3. Compare to the old behavior where the collapsed row showed the call's **result** (`{"total_count": 0}`, `{"success": true, …}`).

---
_Filed from a Discord community report (#report-bugs, 2026-06-23) — follow-up symptom delta on the shipped #4658 fix._

rodboev:
## Thinking Path

- Collapsed tool cards should show the result of a completed tool, not the call arguments, but the current `_toolCardPreviewText` function only checks `tc.preview` during the `done===false` running state.
- This is a regression from `69072ac34` which restructured the preview priority to suppress raw JSON bodies in tool headers (#4658), but inadvertently also suppressed completed-state result display by gating `explicitPreview` behind `done===false`.
- The fix inserts a `_toolResultOneLiner` helper that extracts the first non-empty, non-JSON line from `tc.preview` and returns it for completed tools, preserving the #4658 quiet-header invariant while restoring result visibility.

## What Changed

- `static/ui.js`: Added `_toolResultOneLiner(preview)` helper that takes the first non-empty line of `tc.preview`, rejects JSON/array bodies, and truncates at 180 chars. Updated `_toolCardPreviewText` to check `resultLine` for completed tools before falling back to `argPreview`, with a `tc.snippet` fallback so cold-loaded tools (where the result lives in `snippet`, not `preview`) also show results.
- `tests/test_issue4752_collapsed_card_shows_result.py`: Node-driver extraction tests covering result-wins-on-completion, args-fallback-when-no-preview, JSON-suppressed, multiline-collapses, running-preview-unchanged, and truncation.

## Why It Matters

Every completed tool card shows its input arguments instead of its result, which defeats the purpose of the collapsed summary. Users scanning a conversation see "path=src/main.py" instead of "Found 3 matches," making it harder to understand what each tool actually produced without expanding every card.

## Verification

```text
pytest tests/test_issue4752_collapsed_card_shows_result.py -v --timeout=60
```

## Upstream

Closes #4752.

## Model Used

Claude Opus 4.8 via Claude Code CLI

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -12914,7 +12914,8 @@ function _toolResultOneLiner(preview){
 function _toolCardPreviewText(tc, displaySnippet){
   const explicitPreview=String(tc&&tc.preview||'').trim();
   if(tc&&tc.done===false&&explicitPreview) return explicitPreview;
-  const resultLine=_toolResultOneLiner(explicitPreview);
+  const resultSource=explicitPreview||String(tc&&tc.snippet||'').trim();
+  const resultLine=_toolResultOneLiner(resultSource);
   if(tc&&tc.done!==false&&resultLine) return resultLine;
   const argPreview=_formatToolArgPreview(tc&&tc.args);
   if(argPreview) return argPreview;
--- a/static/ui.js
+++ b/static/ui.js
@@ -12908,7 +12908,8 @@ function _toolResultOneLiner(preview){
   const first=preview.split('\n').find(l=>l.trim())||'';
   const trimmed=first.trim();
   if(!trimmed) return '';
-  if(trimmed[0]==='{'||trimmed[0]==='[') return '';
+  if(trimmed[0]==='{') return '';
+  if(trimmed[0]==='['){try{JSON.parse(trimmed);return '';}catch(e){/* not JSON */}}
   return trimmed.length>180?trimmed.slice(0,177)+'…':trimmed;
 }
 function _toolCardPreviewText(tc, displaySnippet){
--- a/static/ui.js
+++ b/static/ui.js
@@ -12903,9 +12903,19 @@ function _formatToolArgPreview(args){
   const out=parts.join(' · ');
   return out.length>180?`${out.slice(0,177)}…`:out;
 }
+function _toolResultOneLiner(preview){
+  if(!preview) return '';
+  const first=preview.split('\n').find(l=>l.trim())||'';
+  const trimmed=first.trim();
+  if(!trimmed) return '';
+  if(trimmed[0]==='{'||trimmed[0]==='[') return '';
+  return trimmed.length>180?trimmed.slice(0,177)+'…':trimmed;
+}
 function _toolCardPreviewText(tc, displaySnippet){
   const explicitPreview=String(tc&&tc.preview||'').trim();
   if(tc&&tc.done===false&&explicitPreview) return explicitPreview;
+  const resultLine=_toolResultOneLiner(explicitPreview);
+  if(tc&&tc.done!==false&&resultLine) return resultLine;
   const argPreview=_formatToolArgPreview(tc&&tc.args);
   if(argPreview) return argPreview;
   if(tc&&tc.done===false) return 'Running';
--- a/tests/test_issue1824_cli_patch_diff_rendering.py
+++ b/tests/test_issue1824_cli_patch_diff_rendering.py
@@ -112,6 +112,7 @@ def test_rendered_apply_patch_tool_card_html_contains_diff_lines():
         "_toolArgPreviewValue",
         "_toolArgPreviewKeyIsHidden",
         "_formatToolArgPreview",
+        "_toolResultOneLiner",
         "_toolCardPreviewText",
         # #3336: buildToolCard now wraps diff snippets via these helpers.
         "_snippetLooksLikeDiff",
--- a/tests/test_issue4658_transparent_collapsed_preview.py
+++ b/tests/test_issue4658_transparent_collapsed_preview.py
@@ -119,7 +119,7 @@ var _FN_NAMES=[
  '_decodeToolLabelEntities','_redactToolTargetLabel','_shortToolLabel','_toolI18n',
  '_toolTargetLabel','_toolVisibleTargetLabel','_toolCommandTitle','_toolQueryTitle',
  '_toolActionLabelText','_toolArgPreviewValue','_toolArgPreviewKeyIsHidden',
- '_formatToolArgPreview','_toolCardPreviewText','_toolCardAllowsDetail',
+ '_formatToolArgPreview','_toolResultOneLiner','_toolCardPreviewText','_toolCardAllowsDetail',
  '_toolDetailLeadLabel','_toolDetailLeadText','_toolShortName','_transparentEventPreview',
  '_transparentToolStatus','_transparentToolSummary',
  '_isMemorySave','_isSkillUpdate','_tcAction',
--- a/tests/test_issue4658_transparent_collapsed_preview.py
+++ b/tests/test_issue4658_transparent_collapsed_preview.py
@@ -23,6 +23,7 @@ shims, and runs the same render path live streaming and persisted reload share
 (both go through _decorateTransparentEventRow(buildToolCard(tc))).
 """
 import json
+import re
 import shutil
 import subprocess
 from pathlib import Path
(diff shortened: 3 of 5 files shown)


nesquena-hermes:
Pulled the diff and read `_toolCardPreviewText` / `buildToolCard` on master plus the git history for this regression. The fix is sound and the new `_toolResultOneLiner` helper is the right shape. One correction on the attribution, and one interaction worth confirming.

## The fix is correct

Master's `_toolCardPreviewText` (ui.js:12906) only surfaces `tc.preview` while running:

```js
const explicitPreview=String(tc&&tc.preview||'').trim();
if(tc&&tc.done===false&&explicitPreview) return explicitPreview;
const argPreview=_formatToolArgPreview(tc&&tc.args);
if(argPreview) return argPreview;
```

So a completed card (`done!==false`) skips the preview entirely and falls straight to `argPreview` — exactly the "shows path= instead of the result" symptom in #4752. Your insertion restores the result line for completed tools while keeping the JSON-suppression invariant:

```js
const resultSource=explicitPreview||String(tc&&tc.snippet||'').trim();
const resultLine=_toolResultOneLiner(resultSource);
if(tc&&tc.done!==false&&resultLine) return resultLine;
```

Good detail: the `tc.snippet` fallback covers cold-loaded tools where the result lives in `snippet` not `preview`, and `_toolResultOneLiner` rejects `{`/`[`-leading JSON bodies, so the #4658 quiet-header intent is preserved.

## Attribution correction

The PR body credits `69072ac34` ("fix: keep collapsed tool previews quiet", May 31). But that commit's version of the function was `if(explicit) return explicit;` — no `done===false` gate, so completed previews still showed. The `done===false&&explicitPreview` gate that caused this regression was actually introduced later, in `6951d10ae` ("rebase #4411 onto master (v0.51.542) — Stable Assistant Turn Anchors capstone", Jun 20):

```
-  const explicit=String(tc&&tc.preview||'').trim();
-  if(explicit) return explicit;
+  const explicitPreview=String(tc&&tc.preview||'').trim();
+  if(tc&&tc.done===false&&explicitPreview) return explicitPreview;
```

Worth fixing the "regression from 69072ac34" line in the PR/commit message so the bisect record is accurate.

## Confirm the downstream blanking guard

`buildToolCard` blanks the preview in one case (ui.js:12971):

```js
if(toolKind==='shell'||previewText===argPreview||previewText==='Completed'||previewText==='Running'||previewText==='Failed') previewText='';
```

Your `resultLine` is the first non-JSON line of the result, which will not equal `argPreview` (the `key · key` arg summary), so it survives this guard for read/search/web kinds. The only place it's still intentionally suppressed is `toolKind==='shell'`, where the detail-lead block carries the command instead — that's the existing #4411 behavior and your change doesn't regress it. If your node-driver tests already assert a non-shell completed card keeps the result line through `buildToolCard` (not just through `_toolCardPreviewText` in isolation), that's the full path covered. If they only exercise the helper, consider adding one end-to-end `buildToolCard` assertion so the 12971 guard interaction stays locked in.

nesquena-hermes:
Shipped in **v0.51.633** (Release WN) — thanks @rodboev! 🎉 Closes #4752.

Nathan greenlit the concept ("looks more like how it used to, net positive"). Gate: Codex SAFE TO SHIP — XSS-safe (result preview reaches the header via `esc()`, transparent rows via `textContent`), running tools unaffected, JSON/array blob suppression correct; full suite 10451 passed. The test harness `eval` flagged by the threat scanner is the established Node sibling-test pattern (extracting functions from our own trusted ui.js), not untrusted execution. Verified on prod.

PR #2365, opened 2026-05-16
Fix: support YAML list format for custom_providers.models in model dropdown

mccxj:
## Thinking Path

The `get_available_models()` function in `api/config.py` processes each `custom_providers` entry to build the model picker dropdown. The code at L3192-3196 only handled dict-format `models` entries (`{model_id: {}}`), but YAML users commonly declare models as lists (`[m1, m2]`) or list of dicts (`[{id: m, label: ...}]`). When `isinstance(_cp_models_dict, dict)` was `False` for list values, all models from those entries were silently discarded.

The same pattern (handling both dict and list for `models`) already existed elsewhere in the codebase — provider config parsing at L3621 and the live-models fallback handler in `api/routes.py` L6610 — making this an oversight rather than a deliberate design choice.

## What Changed

- `api/config.py` — Extended the `custom_providers` model-ID collection block with an `elif isinstance(_cp_models_dict, list)` branch supporting three sub-formats:
  - Plain string list: `models: [m1, m2]`
  - Dict list: `models: [{id: m1, label: ...}, {id: m2}]`
  - Mixed: `models: [m1, {id: m2}]`
- `CHANGELOG.md` — Added entry under Unreleased > Fixed

## Why It Matters

Users whose `config.yaml` declares custom provider models as YAML arrays saw empty or incomplete model lists in **both** Settings → Preferences → Default Model dropdown and the chat composer model picker. This is the primary model-selection surface — the bug made custom providers with list-formatted models effectively unusable.

## Verification

- 253 related model/provider/dropdown tests pass
- 107 regression tests pass
- Unit tests for all 6 list/dict/mixed format variants pass
- Dict-format (backward compatibility) unaffected

## Risks / Follow-ups

Low risk: the change only adds a new branch to an existing type-check — dict-format code path is untouched. One follow-up: the `api/routes.py` `/api/models/live` handler already had list support (L6610-6615), confirming the list-format design is intentional.

## AI Usage

- **Provider**: custom:x-openai
- **Model**: deepseek-v4-flash-free
- **Tools**: patch, execute_code, terminal, search_files, read_file

PR #2176, opened 2026-05-13
fix: handle dict model entries in provider models list

MrFant:
When a provider's 'models' config contains dicts (e.g. {"id": "x", "label": "y"}) instead of plain strings, _apply_provider_prefix() crashes with:
  AttributeError: 'dict' object has no attribute 'startswith'

This happens because the list comprehension at line 3505 passes the raw dict as the model ID. The fix extracts 'id' and 'label' from dict entries while keeping string entries as-is.

Fixes the /api/models and /api/onboarding/status 500 errors.

Issue #1539, opened 2026-05-03
bug(providers): removed provider still appears in model dropdowns until server restart


nesquena-hermes:
## Summary

A user reported in Discord (Deor, May 03 2026):

> Also, after removing Ollama Cloud from the Providers screen, it still shows in all the provider/model lists. Guessing a restart will clean that up.

A restart **should not** be required. Removing a provider from Settings → Providers should make it disappear from every model dropdown immediately.

## Repro

1. Configure Ollama Cloud (or any non-active provider) with an API key.
2. Settings → Providers → click "Remove" on the Ollama Cloud card.
3. Toast confirms removal. The Ollama Cloud card disappears from the providers list.
4. Open the model dropdown in the composer (or `/model` slash command, or Settings → Default Model).
5. **Bug:** Ollama Cloud entries are still listed.
6. Restart the server. They are gone.

## Root cause

Two layers of stale state, only one of which is currently flushed:

**Layer A — Frontend caches** (the load-bearing problem)

- `static/commands.js:84` — `_slashModelCache` and `_slashModelCachePromise` cache `/api/models` for the `/model` slash command. Never invalidated on provider remove.
- `static/ui.js:228` — `_dynamicModelLabels` and `window._configuredModelBadges` are populated by `populateModelDropdown()` on app boot and on profile switch. Never invalidated on provider remove.
- `static/panels.js:3380-3397` — `_removeProviderKey()` only calls `await loadProvidersPanel()` after delete. That refreshes the providers card list but does NOT call `populateModelDropdown()` or any cache-invalidation hook for the dropdown surfaces above.

**Layer B — Server-side TTL cache** (already fixed elsewhere, just not on this code path)

- `api/providers.py:506` — `set_provider_key()` calls `invalidate_models_cache()`. ✅
- `api/routes.py:2465` — `/api/profile/switch` calls `invalidate_models_cache()`. ✅
- `api/routes.py:1962-1969` — `/api/providers/delete` does **not** call `invalidate_models_cache()`. ❌

The route delegates to `remove_provider_key()`, which calls `set_provider_key(provider_id, None)`, which DOES call `invalidate_models_cache()` at line 506 — so the server-side cache IS flushed. **The problem is purely on the frontend.** The very next `/api/models` request returns the correct (smaller) list, but the page never asks for one until the user reloads.

## Fix

Two changes, both small:

[...]

PR #4852, opened 2026-06-24
feat: interactive model manager for custom providers

luandnh:
## Thinking Path

- Custom providers can fetch hundreds of models from their /v1/models endpoint
- All fetched models appear in the chat model picker, making it hard to find the right one
- The backend already supports filtering via custom_providers[].models in config.yaml
- But there was no UI to manage which models are active — users had to edit config.yaml manually
- This PR adds an interactive model manager in the provider settings card

## What Changed

**Backend:**
- Added `set_custom_provider_models()` in `api/providers.py` — updates config.yaml's
  custom_providers[].models field with an allowlist of model IDs
- Added `POST /api/providers/models` route — accepts `{provider, models}` list

**Frontend:**
- Custom provider card now shows interactive model management UI instead of a static hint
- **Active models** section — pills with × to remove from allowlist
- **Available (from refresh)** section — pills with + to add, populated after Refresh click
- Custom model name input — type + Enter to add models not in the live list
- Save button persists selection to config.yaml and refreshes the model dropdown
- CSS styles for remove-tag (hover turns red) and add-tag (accent color)

## Why It Matters

Without this UI, users with custom providers that expose many models (ZenMux, aggregator
gateways, etc.) had to manually edit config.yaml to curate the list. Now they can:
1. Click Refresh to discover all models
2. Click + on the ones they want (or × to remove)
3. Type any missing model name manually
4. Save — only selected models appear in the model picker

## Verification

- All 29 tests in test_custom_providers_in_panel.py + test_issue1106_custom_providers_models.py pass
- Python syntax checked (py_compile)
- JavaScript syntax checked (Node)

## Risks / Follow-ups

- Low risk: the model manager only activates for p.is_custom providers
- Existing behavior unchanged for non-custom providers (OpenAI, Anthropic, etc.)
- When models list is empty, all live models appear (same as before)


luandnh:
Fixed in 97b066c:

- Moved `reload_config()` and `invalidate_models_cache()` into a nested try/except so a post-write cache sync hiccup is logged but does not return a false `{"ok": False}` to the client.
- Pattern matches `_clean_provider_key_from_config` (line 2809) which documents the same "call outside lock / own error handling" constraint.

Ready for re-review.

nesquena-hermes:
## 🔬 Deep-review + full-gate report — PR #4852

**Scope:** interactive model manager for custom providers (`api/providers.py`, `api/routes.py`, `static/panels.js`, `static/style.css`, tests). +334/−27, 5 files.
Gated on a worktree of the PR head (`cbd7e62`), rebased clean onto current `origin/master` for the render-surface gate.

### Gate results

| Gate | Result |
|------|--------|
| Static threat scan (added lines) | ✅ **CLEAN** (score 0) |
| `node --check static/panels.js` | ✅ pass |
| Full pytest suite (`tests/`, no:xdist) | ✅ **10820 passed**, 81 skipped, 1 xfailed, 2 xpassed — only the 2 documented `test_scheduled_jobs_profile_isolation` cron-context flakes failed; both **pass in isolation** (12/12) |
| Activity-stream regression gate (on rebased worktree) | ✅ **ALL LAYERS PASS** — 6/6 invariants + 129 targeted render tests |
| Codex regression gate (GPT, high reasoning) | ⚠️ **SHIP ONLY WITH FIXES** (2 SILENT findings, 0 BRICK/CORE) |

### Security review (the original review focus) — CLEAN ✅

- **Auth-gating:** the new `POST /api/providers/models` is an `if`-branch inside `handle_post`, which is reached only through `server.py:_handle_write → check_auth()` (401 when auth enabled + no valid session) **and** the CSRF gate (`_check_csrf`, 403 on cross-origin/tokenless). Same posture as every other config-write route. No bypass.
- **Config.yaml write safety:** `set_custom_provider_models` mutates the parsed dict **under `_cfg_lock`** and persists via `_save_yaml_config_file → yaml.safe_dump`. No raw string templating, no attacker-controlled key escaping to top-level config. Model IDs are `str(m).strip()`-coerced and empties dropped; non-list `models` payloads are rejected at the route (`bad(...)`).
- **Scoping to custom providers only:** writes only touch the entry matched by `_custom_provider_name_matches` inside `custom_providers[]`. A builtin/oauth/plugin provider id cannot match (no entry in that list) → builtin catalogs cannot be clobbered.
- **XSS:** all model id/label values render via `textContent` (panel pills, lines ~32/74). The only `innerHTML` writes are static SVG icon strings + empty-string clears — no interpolation of model data into markup. No injection surface.

### ⚠️ Codex findings (verified)

**1. [SILENT — VERIFIED, data loss] dict-format `models:` round-trips lossily** — `api/providers.py:2607-2611`.
`get_providers()` builds `cp.models` only when `models` is a **list** (or a single `model`). When a custom provider uses the **dict form** of `models:` (a real, supported shape — `api/config.py:6291-6305` parses dict keys + per-model metadata), `cp.models` comes back **empty**. The panel then shows **no Active pills**, and a Save POSTs an empty/incomplete list → `set_custom_provider_models` does `cp.pop("models")`, **deleting the entire dict-format model block + per-model metadata (`context_length`, labels, etc.) from config.yaml**. The dict-preservation logic added in `set_custom_provider_models` cannot help because the data never reaches the panel.
*Fix (mechanical):* make the `get_providers()` custom-provider extraction mirror `api/config.py:6291-6305` — accept `model`, dict keys, list strings, and list-of-dict `id|model|name`, de-duped — so the panel's Active list is complete before Save.

**2. [SILENT — UX gap] Refresh can't discover un-allowlisted models once an allowlist exists** — `static/panels.js` refresh handler.
The Refresh handler reads `/api/models`, but `api/config.py` treats a configured custom-provider `models` allowlist as authoritative and skips the live `/v1/models` probe → "Available" stays empty even when the provider has more live models. Lower severity (the manual "type model name + Enter" input still works), but the Refresh affordance is misleading for already-allowlisted providers.
*Fix:* point Refresh at a live-catalog fetch that bypasses the configured allowlist.

> Note: the existing test only string-matches `panels.js`; it does **not** exercise the dict-format config round-trip, so the green suite did not catch finding #1.

### 🧭 Recommendation: **BOUNCE TO AUTHOR (changes-requested) — fix #1 before merge**

The feature concept is sound and the security posture is clean (auth + CSRF gated, `safe_dump`, custom-scoped, XSS-safe), so this is **not** a concept reject and **not** a security hold. But finding **#1 is real, silent data loss** in config.yaml for the dict-format custom-provider shape, triggerable by a normal Save — that must be fixed before merge. Finding #2 is a recommended follow-up (or same-PR) UX fix. Both fixes are mechanical and precisely specified above; a regression test for the dict-format round-trip should accompany #1.

*Gates clean: security, full suite, activity-stream render surface. Held only on the verified data-loss fix.*

—
🤖 Automated deep-review + full-gate pass · review-only (not merged)

nesquena-hermes:
# 🔬 Gate certification — RED ⛔ (both advisors converge: dict-form `models:` data loss + missing backend tests)

**Certified head:** `sha:cbd7e628bf3d` (rebased `98e3bab8`) · **PR:** #4852 · _luandnh — "feat: interactive model manager for custom providers"_ · **`maintainer-review`, concept fit: 4/5**
**Verdict:** Concept + security posture are good (auth-gated config write, `yaml.safe_dump`, custom-scoped, XSS-safe, validated — I verified all live). **But both Codex and Opus independently flag the same blocker:** custom providers whose `models:` is a **dict** (metadata-rich form) are mis-read by the render path → the UI saves a bad/empty list → the save path then **deletes the allowlist + per-model metadata**. Plus the non-trivial config-write function has **zero backend tests**. Fix-and-resubmit (not a concept reject).

### What I ran (isolated worktree `/tmp/wt-gate-4852`, rebased onto current master)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SHIP ONLY WITH FIXES** — dict-form data loss (#1) + cache-flush (#2) + `custom:` scoping (#3) |
| Opus (full review) | **REQUEST_CHANGES** — same dict-form data loss (#1) + missing backend tests (#2) |
| Full pytest suite (`-p no:xdist`, rebased) | **11123 passed, 0 failed** (green — but no dict-form round-trip test, so it masks the bug) |
| PR's own test | **12 passed** (only a panels.js string-match; no backend behavioral coverage) |
| Live API verification (mine) | endpoint **auth-gated** ✓; not-found→400 ✓; missing-provider→400 ✓; bad-models→400 ✓; 0 JS errors ✓ |

### ⛔ Blocking #1 (data loss — BOTH advisors, I confirmed) — dict-form `models:` not read, then deleted on save
`get_providers()` (providers.py:~2607) only handles `models` as a **list**: `if isinstance(cp.get("models"), list): cp_models = [{"id": str(m), "label": str(m)} for m in cp["models"]]`. For a **dict** `models` (the metadata form: `{model_id: {context_length, label}}`) or a **list-of-dicts**, the UI receives missing/wrong active IDs. The panel then saves that bad/empty list, and `set_custom_provider_models` (providers.py:~2875) does `cp.pop("models", None)` on empty — **deleting the allowlist and all per-model metadata**. The save side carefully *preserves* dict-form, but the read side doesn't *surface* it, so the round-trip silently destroys it.
**Fix-spec (both advisors):** normalize custom-provider `models` in `get_providers()` the way `/api/models` (config.py:~2513) does — dict keys → `{id, label}`, list-of-dict entries use their `id`/`label`, list strings stay strings. Mirror the existing normalization.

### ⛔ Blocking #2 (Opus) — no backend tests for non-trivial config-write logic
`set_custom_provider_models` has branching (dict vs list, empty vs non-empty, found vs not-found) and **zero behavioral tests** — the only added assertion is a `panels.js` string match. Per project rules, logic needs tests; a **dict-form round-trip test would have caught #1**.
**Fix-spec:** add backend tests for `set_custom_provider_models` covering list-form, dict-form (round-trip preserving metadata), empty (field removed), and not-found.

### Also (Codex #3, SILENT) — route doesn't require `custom:` prefix
`/api/routes.py:11801` lowercases `provider` and passes it straight through, relying on `_custom_provider_name_matches`; a custom provider *named* like a builtin could be matched via `provider:"anthropic"`. **Fix:** require `provider_id.startswith("custom:")` before calling `set_custom_provider_models`. (Plus Codex #2 / Opus #2 nits: call `_refreshModelDropdownsAfterProviderChange()` on save success so the dropdown isn't stale; add a failure toast on the refresh path.)

### ✅ Good (keep it — security posture is solid)
- ✅ **Auth-gated**: `/api/providers/models` is NOT in `PUBLIC_PATHS` (auth.py:52); `check_auth` runs in `server.py:381` before POST routing. No unauthenticated config write. (Verified live: my authed session got proper 400s.)
- ✅ **No YAML injection / config corruption**: model strings `str(m).strip()`-coerced at the route; persisted via `_save_yaml_config_file` → `yaml.safe_dump(sort_keys=False)` (safe scalars); only the matched `custom_providers[]` entry's `models` is touched; written under `_cfg_lock`; cache synced after (own try/except).
- ✅ **XSS-safe** UI (`tag.textContent = m`, not innerHTML); 0 JS errors; rebases cleanly.

### Concept (I scored it — rubric 1–5): **4/5 (strong)**
Managing a custom provider's model allowlist in the UI is a clean, expected feature (custom providers already exist; curating their models is the natural next step). Both advisors concur "concept is good."

**Bounce for the dict-form data-loss fix + backend tests (both advisors converged), then re-gate — concept + security are good, this is fix-and-resubmit.** Normalize dict/list-of-dict custom `models` in `get_providers()` (mirror config.py:2513), add `set_custom_provider_models` round-trip tests (dict-form especially), require `custom:` prefix on the route, and the two UX nits (dropdown refresh + failure toast). Security posture (auth, safe_dump, XSS, validation) verified solid. Concept 4/5. I applied `gate-fail`; the dict-form data-loss is reproduced + dual-advisor-confirmed. Cert valid only at `sha:cbd7e628bf3d`.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close. Cert valid only at `sha:cbd7e628bf3d`; a new push invalidates it → re-gate._

luandnh:
Both blocking issues addressed in 1655bcf:

**Blocking Issue #1 — Dict-form data loss (HIGH)**
Added `elif isinstance(cp.get("models"), dict)` branch in `get_providers()` that extracts dict keys as model IDs, matching how `/api/models` already normalizes them. The UI now receives the correct model list, so saves can't silently delete the allowlist.

**Blocking Issue #2 — Missing backend tests**
Added `tests/test_custom_provider_models_backend.py` with 5 tests:
- dict-form round-trip (keys + metadata preserved)
- list-form round-trip
- empty models list removes field
- unknown provider returns error
- get_providers() surfaces dict-form models (read-path fix)

All 5 pass on current head.

nesquena-hermes (changes requested):
## Re-gate after your re-push (`1655bcf` / `79caa30`) — close, one residual on the same data-loss finding

Thanks for the quick turnaround. Most of the gate-cert RED is genuinely resolved:

- ✅ **Dict-form `models:` read path — FIXED.** `get_providers()` now has the `elif isinstance(cp.get("models"), dict)` branch surfacing dict keys, and `set_custom_provider_models()` preserves per-model metadata (`{m: dict(existing_models.get(m, {})) for m in models}`) on a dict-form save. The headline data-loss case is closed.
- ✅ **Backend tests — DELIVERED.** `tests/test_custom_provider_models_backend.py` is real behavioral coverage (writes a temp `config.yaml`, calls the functions, re-reads from disk, asserts persisted shape + metadata). Verified green via our sandboxed test gate: **5 passed**. The dict-form round-trip test with metadata preservation is exactly the regression test the finding needed.

### ⛔ One residual still open — **list-of-dict `models:` is still mangled** (same data-loss vector)

The gate-cert fix-spec called for normalizing **both** dict-form *and* list-of-dict entries (it said: _"list-of-dict entries use their id/label"_, mirroring `api/config.py`). The dict branch landed, but a **list-of-dict** config still falls into the plain-`list` branch first and gets `str(m)`-stringified:

```python
# api/providers.py:2609-2610 (live head)
if isinstance(cp.get("models"), list):
    cp_models = [{"id": str(m), "label": str(m)} for m in cp["models"]]   # <-- str(dict) here
elif isinstance(cp.get("models"), dict):
    ...
```

A `list`-of-`dict` like `models: [{id: model-a, label: "A"}]` is a `list`, so it hits line 2610, and each element is a dict → `str(m)` produces garbage IDs:

```
{'id': "{'id': 'model-a', 'label': 'A'}", 'label': "{'id': 'model-a', 'label': 'A'}"}
```

The UI then shows those garbage IDs as active; on save, `set_custom_provider_models`'s `else` branch (models was not a dict) does `cp["models"] = list(<garbage strings>)` — **destroying the original list-of-dict allowlist + labels.** This is the same data-loss family the finding targeted, for a config shape that is first-class and documented (CHANGELOG: PR #2365 / #2176 both explicitly support `[{id, label}, ...]`).

### Precise fix-spec

1. **`api/providers.py:2609-2610`** — make the `list` branch element-aware, mirroring `api/config.py:4427-4433`:
   ```python
   if isinstance(cp.get("models"), list):
       cp_models = []
       for m in cp["models"]:
           if isinstance(m, dict):
               mid = str(m.get("id") or m.get("model") or m.get("name") or "").strip()
               if mid:
                   cp_models.append({"id": mid, "label": str(m.get("label") or mid)})
           else:
               cp_models.append({"id": str(m), "label": str(m)})
   elif isinstance(cp.get("models"), dict):
       ...
   ```
2. **Add a `test_list_of_dict_round_trip`** to `tests/test_custom_provider_models_backend.py`: list-of-dict config → `get_providers()` surfaces the real IDs (not stringified dicts) → save a subset → re-read preserves the retained entries' IDs/labels. (Without the test this regression will silently survive again — it's why the headline dict-form case was caught but this one wasn't.)
   - Optional, same family: consider preserving list-of-dict label metadata in `set_custom_provider_models`'s `else` branch on partial save (today a list-of-dict provider's labels drop to bare strings even after the read fix). Lower priority than the read fix.

### Non-blocking notes (optional, your call)

- The `custom:`-prefix route guard the original cert flagged as SILENT is still absent (`api/routes.py:~11894` passes `provider` straight through; a custom provider *named* like a builtin could be matched). Narrow surface — fine to defer.
- The save-success handler in `static/panels.js` doesn't call `_refreshModelDropdownsAfterProviderChange()` the way the set-key/remove-key flows do, so the model dropdowns can stay stale until reload. (The failure-toast nit *is* now addressed — thanks.)

Once the list-of-dict read normalization + its round-trip test land, this should re-gate clean — the concept and security posture were already verified solid. Appreciate the work on this.

luandnh:
Additional findings addressed in 18a3864:

- **Silent failure on Refresh** — replaced `console.warn` with `showToast` so users see feedback when model fetch fails
- **Route scoping** — added `provider_id.startswith("custom:")` guard before delegating to `set_custom_provider_models()`

Previously addressed blockers (1655bcf):
- Dict-form models now normalized in get_providers() — dict keys → model IDs
- 5 backend tests added covering dict round-trip, list round-trip, empty, not-found, get_providers read path

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (converged on BOTH my prior blockers; one new SILENT dropdown-staleness bug)

**Certified head:** `sha:18a3864f` (rebased onto current master, `git apply` clean) · **PR:** #4852 · _luandnh, interactive model manager for custom providers_
**Verdict:** Excellent convergence — the re-push **fixes both blockers from my prior bounce** (dict-form read/write round-trip is now lossless, with a thorough new backend test; verified 17/17). But the gate found one NEW real SILENT: the model-allowlist save path doesn't refresh the model dropdowns, so the composer picker / `/model` cache keep offering the stale pre-save model list until reload. One-line fix, then ship.

### What I ran (rebased worktree `/tmp/wt-rebase-4852`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean (0 conflicts); re-rebased once after #5279 shipped mid-gate (see i18n note); diff vs master = the PR's 6 real files only |
| Prior-blocker verification (mine, in-process + tests) | ✅ BOTH fixed |
| Codex (reproduce) | SHIP-WITH-FIXES — 1 phantom (i18n stale-base) + 1 real SILENT (dropdown refresh) |
| Full pytest suite | **11374 passed**, 0 failed |
| New backend test file | ✅ 17/17 (dict-form round-trip, list-form, empty-removes-field, unknown-provider-error, get_providers read-path) |

### Findings

**✅ CONVERGED — both prior blockers fixed (verified):**
1. (was CORE) dict-form read/write asymmetry: `get_providers` now reads BOTH forms (`elif isinstance(cp.get("models"), dict): [{"id":m,"label":m} for m in cp["models"].keys()]`, providers.py ~2626), and `set_custom_provider_models` **preserves dict-form on save** ("if models was a dict, save as dict keys so per-model metadata is not lost", ~2883). I confirmed the round-trip is LOSSLESS in-process: the new `test_dict_form_round_trip` reads dict keys → saves a subset → re-reads and asserts `cp["models"]["model-b"] == {"context_length":64000,"label":"Big model"}` survives. The metadata-deletion bug is closed.
2. (was SILENT) zero backend tests: new `tests/test_custom_provider_models_backend.py` (+208) drives the dict-form fallback my bounce named + list-form + empty + unknown-provider + read-path. 17/17 pass. Exactly the coverage I asked for.
Security re-confirmed GOOD (auth-gated, safe_dump, custom-scoped, XSS-safe, invalid-provider 400s). Nice work.

**⛔ SILENT (I REPRODUCED by read) — stale model dropdowns after saving the allowlist** (`static/panels.js:10228`): the new `/api/providers/models` save handler, on `res.ok`, calls only `await loadProvidersPanel()`. It does NOT call `_refreshModelDropdownsAfterProviderChange()` — which every sibling mutation path DOES (key-save :10330, key-delete :10357, model-refresh), with an explicit #1539 comment that without it the composer picker, `/model` slash cache, Settings default-model dropdown, and configured-model badges keep offering the stale list until a page reload. So a user edits a custom provider's model allowlist → the allowlist that's meant to control those very dropdowns doesn't take effect in them until reload. **Fix:** on `res.ok`, call `_refreshModelDropdownsAfterProviderChange()` before/alongside `await loadProvidersPanel()` (mirror the :10330 key-save path exactly). Ideally add a small test asserting the save path invalidates the slash-model cache.

**ℹ️ Phantom (NOT a defect) — Codex flagged an i18n.js zh-CN "regression" (i18n.js:9778):** this was a self-inflicted stale-base artifact — #5279 (zhCN translation fill, which I gate-certified GREEN) shipped to master DURING this gate, so my worktree (cut before #5279 landed) showed master's newer zh strings as "reverted." #4852 does NOT touch i18n.js (its three-dot footprint is 6 files, none i18n). I re-rebased onto the new master and i18n.js is byte-identical to master. Disregard.

**RED — gate-fail/changes-requested, but a fast converge: ONE one-line fix** (add `_refreshModelDropdownsAfterProviderChange()` to the model-allowlist save path at panels.js:10228, mirroring the key-save path). Both original blockers are genuinely closed with strong test coverage and security holds — this is the last mile. concept 4/5 (interactive custom-provider model manager). Author @luandnh. The stale GitHub `CHANGES_REQUESTED` review predates the re-push (superseded).

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master (twice, after #5279 shipped mid-gate) rather than block on staleness. Cert valid for sha:18a3864f._

nesquena-hermes:
🔬 **Gate re-check — SEMANTIC rebase conflict on the `is_custom` render path (needs author reconciliation) + concept/UX → Nathan**

**Head:** `sha:65e54168` (converged since my prior RED — the `normalize list-of-dict models` commit addresses the dropdown-staleness SILENT ✅).

I attempted a rebase-first onto current master and hit a **semantic conflict I shouldn't auto-resolve**: master has since gained its own `p.is_custom` provider-card rendering (the "Custom provider loaded from config.yaml / hermes model" path + a `filter(...)` that now includes `p.is_self_hosted`, from #5408), and this PR **rewrites the same `is_custom` branch** into the new interactive model-manager. They're not adjacent additions — they're two competing renderings of the same `is_custom` card. A keep-both resolution drops the config.yaml-custom string and fails `test_providers_panel_renders_config_yaml_custom_providers` (which now expects BOTH the self-hosted filter AND the config.yaml rendering).

**This needs you (the author) to rebase + reconcile** which `is_custom` rendering wins (or how they compose — the interactive manager likely should absorb the config.yaml-custom case), and update `test_custom_providers_in_panel.py` accordingly. I don't want to guess the intended merge of two UI paths.

**Also — concept/UX call for the maintainer:** this is a `size:L` UX feature (interactive model manager for custom providers, +628/-29) — it parks for Nathan's concept + visual sign-off regardless. The dropdown-staleness fix landed; the remaining blockers are (1) this semantic rebase, (2) Nathan's concept/visual approval.

Parking as **needs-author-rebase** (semantic `is_custom` conflict) + concept/UX-for-Nathan. crit=2, @luandnh.

PR #4854, opened 2026-06-24
feat(titlebar): show active model/provider in subtitle line

b-yelverton:
## Thinking Path

- Hermes WebUI lets users switch models and providers, but it is not visible which provider is actually serving without opening the model picker
- When fallback provider chains fire, the active model may differ from the configured one
- The chat titlebar already shows message count — it is natural to also show the current model
- This PR appends the model name (with provider) to the subtitle line: "42 messages · claude-sonnet-4-6 via anthropic"

## What Changed

**`static/panels.js`** — 10 lines in `syncAppTitlebar()`
- Uses `_latestGatewayRoutingForSession()` when available (with type-safety guard), falls back to `S.session.model` / `S.session.model_provider`
- Strips provider prefix from prefixed models (e.g. "anthropic/claude-sonnet-4-6" → "claude-sonnet-4-6")
- Appends " via <provider>" when provider is known
- Only active on the chat panel

## Why It Matters

Transparency — users can see which provider is serving their session without opening the model picker. Especially useful during provider outages and fallback transitions.

## Verification

Built and ran locally — the titlebar shows the active model on the chat panel and disappears (correctly) on other panels like Kanban, Settings, etc.

## Risks

None. The change is purely cosmetic, guardrailed by `typeof _latestGatewayRoutingForSession === 'function'` check, and falls back silently when data is unavailable.

## Model Used

- Provider: Anthropic (Claude Max)
- Model: Claude Sonnet 4.6

### DIFF
--- a/static/panels.js
+++ b/static/panels.js
@@ -63,6 +63,16 @@ function syncAppTitlebar() {
     sourceLabel = S.session.source_label || S.session.source_tag || S.session.raw_source || '';
     // Recovered sidecars stamp source_label 'WebUI' (api/session_recovery.py); don't badge a native session as its own source (#3338).
     if (/^webui$/i.test(sourceLabel)) sourceLabel = '';
+    // Append active model/provider chip so it's always visible in the titlebar.
+    const _routing = typeof _latestGatewayRoutingForSession === 'function'
+      ? _latestGatewayRoutingForSession(S.session) : null;
+    const _usedModel = (_routing && _routing.used_model) || S.session.model || '';
+    const _usedProvider = (_routing && (_routing.used_provider || _routing.provider)) || S.session.model_provider || '';
+    if (_usedModel || _usedProvider) {
+      const _modelShort = _usedModel.includes('/') ? _usedModel.split('/').pop() : _usedModel;
+      subText = subText ? subText + '  ·  ' + _modelShort : _modelShort;
+      if (_usedProvider) subText += ' via ' + _usedProvider;
+    }
   } else {
     const key = APP_TITLEBAR_KEYS[panel];
     mainText = key && typeof t === 'function' ? t(key) : (panel.charAt(0).toUpperCase() + panel.slice(1));


nesquena-hermes:
Pulled `cron-pr-4854` into a read-only worktree and read the 10-line diff against `origin/master`, plus the surrounding `syncAppTitlebar` body (`static/panels.js:50-78`) and the existing gateway-routing label helpers in `static/ui.js`. The idea is good and the routing-aware source (`_latestGatewayRoutingForSession`) is the right thing to read — `S.session.model` / `model_provider` are genuinely populated (set in `boot.js:1435-1436`), so the fallback chain is sound. Two consistency notes before this lands:

## 1. It reimplements model-label shortening instead of reusing `getModelLabel()`

The PR derives the short name with a naive split:

```js
const _modelShort = _usedModel.includes('/') ? _usedModel.split('/').pop() : _usedModel;
```

But the repo already has the canonical formatter `getModelLabel(modelId)` at `static/ui.js:4661`, which every other surface uses. The naive split diverges from it in two real cases:

- `anthropic/claude-sonnet-4.6` → this PR shows `claude-sonnet-4.6`; `getModelLabel` returns the friendly `Sonnet 4.6` (via its `STATIC_LABELS` map).
- `@custom:ai_gateway:Qwen3.6-35B-A3B` → `.includes('/')` is false, so this PR shows the entire raw id *including* the `@custom:` prefix; `getModelLabel` strips it to `Qwen3.6-35B-A3B` (ui.js:4668-4672).

So the titlebar subtitle would show a different, uglier model string than the model picker and the gateway chip. Swapping `_modelShort` for `getModelLabel(_usedModel)` fixes both and inherits the URI-scheme / Ollama-tag handling too.

## 2. Provider casing won't match the gateway chip

The PR appends the raw provider:

```js
if (_usedProvider) subText += ' via ' + _usedProvider;
```

The existing routing display runs the provider through `_gatewayProviderName()` (`static/ui.js:4738-4742`), which strips `custom:` and title-cases:

```js
return text.replace(/^custom:/,'').replace(/[-_]/g,' ').replace(/\b\w/g,c=>c.toUpperCase());
```

So `_gatewayRoutingLabel` renders `via Anthropic`, while this PR's subtitle would render `via anthropic`. Using `_gatewayProviderName(_usedProvider)` (or reusing `_formatGatewayModelLabel(modelId, label, routing)` which already produces the exact `"<label> via <Provider>"` string) keeps the two surfaces identical.

## 3. No test, though there's precedent

`syncAppTitlebar` already has a regression test file (`tests/test_app_titlebar_restore.py`), so a small source-level assertion that the subtitle includes the model/provider segment would fit the existing pattern and guard against the subtitle silently dropping the chip later.

None of this is a correctness bug — the feature works. It's about routing both halves of the chip through the helpers that already exist (`getModelLabel` + `_gatewayProviderName`, or just `_formatGatewayModelLabel`) so the titlebar reads identically to the model picker. That's a ~2-line change and would also shrink the diff.

nesquena-hermes:
Nice, clean transparency feature — and it renders well (verified: titlebar subtitle shows `42 messages · claude-sonnet-4-6 via anthropic`, model prefix stripped, provider appended, no XSS since it goes through `subEl.textContent`). One **must-fix** before it can ship, plus it'll want a maintainer visual/concept sign-off.

### 🔴 MUST-FIX — `_usedModel.includes()` can crash `syncAppTitlebar` (gateway sessions)
`_latestGatewayRoutingForSession()` can return a `used_model` that is a truthy **non-string scalar** (int/float/bool) — the backend normalization at `api/streaming.py:1388` permits it, and the existing gateway-label path already defends against this by casting to string first (`static/ui.js:4749`). Your `|| ''` fallback only catches falsy values, so a non-string-truthy `used_model` hits `_usedModel.includes('/')` → `TypeError: includes is not a function`, throwing inside `syncAppTitlebar` (titlebar render breaks).

Exact fix — coerce to string before use (both model and provider):
```js
const _usedModel = String((_routing && _routing.used_model) || S.session.model || '');
const _usedProvider = String((_routing && (_routing.used_provider || _routing.provider)) || S.session.model_provider || '');
```
(then the existing `.includes('/')` / `.split('/')` are safe.)

### Otherwise
- Logic is sound, guard on `typeof _latestGatewayRoutingForSession === 'function'` is correct, degrades cleanly when routing is absent, scoped to the active-session titlebar branch, no regression to the message-count/source-label subtitle.
- Full suite passed on my stage; the only blocker is the string-coercion crash above.

Once that 1-line coercion is in, it's gate-clean. It's a visible titlebar change, so it also needs @nesquena's visual/concept OK (I've captured a screenshot for that review). Thanks @b-yelverton!

PR #4855, opened 2026-06-24
Fix cron task APIs when another cron package shadows agent cron

luckfu:
## Summary
- prefer the Hermes agent cron package before importing cron.jobs for Task APIs
- clear an already-imported unrelated top-level cron module before retrying the agent import
- add regression coverage for an external cron package shadowing agent cron.jobs

## Root cause
In environments where another top-level cron package appears earlier on sys.path, Python resolves import cron to that package instead of the agent checkout's cron package. The Task APIs then catch ModuleNotFoundError for cron.jobs and return {"jobs": [], "cron_unavailable": true} even though the agent cron jobs file exists.

## Verification
- PYTEST_ADDOPTS='-o cache_dir=/tmp/hermes-webui-pytest-cache' python3 -m pytest tests/test_cron_import_shadowing.py tests/test_issue4768_cron_module_missing.py tests/test_cron_toast_notifications.py tests/test_cron_query_param_validation.py tests/test_cron_model_override.py tests/test_cron_model_provider_picker.py -q
- local /api/crons returned 3 jobs with no cron_unavailable flag after the fix


luckfu:
Updated in 39f466b7 to address the automated review concern:

- guards cron import-path repair with a process-wide lock
- makes the repair idempotent for the active agent checkout
- avoids repeated sys.path mutation/scans after the path is ready
- keeps the wrong-origin cron module eviction inside the same lock
- extends the regression test to assert the second helper call leaves sys.path unchanged

Focused cron tests still pass: 22 passed.

nesquena-hermes:
Pulled the branch at HEAD `39f466b70` and read `_ensure_agent_cron_import_path()` against `api/config.py`'s existing sys.path setup and the agent's actual `cron/` layout (`__init__.py`, `jobs.py`, `scheduler.py`, …). The lock + fast-path + module-eviction structure is sound, and the regression test correctly proves the shadow is overridden and the second call is a no-op. One real cross-cutting concern, plus a smaller note.

## The forward-insert can re-introduce the problem config.py deliberately avoided

`api/config.py:248-250` places `_AGENT_DIR` at the **end** of `sys.path`, and the comment just above it (`config.py:236-247`) is explicit about *why*: when third-party packages get vendored into the agent checkout (via `-t .`-style local installs), they end up alongside the Hermes source. Putting `_AGENT_DIR` at the front of `sys.path` then makes Python resolve `import pydantic` / `import openai` from that local directory, which breaks whenever the host platform differs from the container. The documented fix is to append `_AGENT_DIR` at the END so site-packages resolves third-party packages first and only Hermes-specific modules fall through to the agent dir.

The new helper undercuts that. It removes `agent_path` from wherever it sits and re-inserts it **before the first shadow**:

```python
if shadow_indexes:
    sys.path.insert(min(shadow_indexes), agent_path)
```

The issue describes the competing `cron` package appearing *earlier* on `sys.path`. If that shadow sits at or before site-packages (a low index — exactly what "earlier" implies), `min(shadow_indexes)` moves `agent_path` ahead of site-packages, and the move is **permanent** (`_AGENT_CRON_IMPORT_PATH_READY` latches it). From that point every third-party `import` resolves `agent_path` first — re-introducing the precise platform-mismatch breakage `config.py` was engineered to prevent in the vendored-deps deployment. In this checkout the agent dir only holds `agent/`, `cron/`, `hermes_cli/` (deps live in a venv), so it wouldn't bite here — but the documented vendored layout would hit it, and that's the harder-to-debug regression class.

## Suggested safer shape: don't reorder global sys.path for non-cron imports

Since the only goal is resolving `cron.jobs` from the agent checkout, a surgical load avoids touching global import priority for everything else. Two options, both lock-safe:

1. **File-spec load** — after evicting stale `cron*` from `sys.modules`, load the agent package explicitly with `importlib.util.spec_from_file_location("cron", <agent_cron>/__init__.py, submodule_search_locations=[<agent_cron>])`, register it in `sys.modules["cron"]`, then `from cron.jobs import …` resolves against it. No `sys.path` mutation at all.
2. **Transient front-insert** — insert `agent_path` at index 0, import, then restore the prior `sys.path` inside the same `with _AGENT_CRON_IMPORT_PATH_LOCK:` block, so the elevated priority never outlives the import.

Either keeps the cron fix while preserving the end-of-path invariant for third-party resolution.

## Minor

In the shadow scan, `Path(path_entry).resolve() != Path(agent_path)` is effectively dead — `agent_path` was already stripped from `sys.path` by the `while ... remove` loop just above, so no surviving entry equals it. Harmless, just redundant.

Per the worktree execution ban I didn't run the new test; reading it, the setup (two competing `cron/__init__.py` dirs, `syspath_prepend` of both, assert agent `jobs.py` wins) correctly models the bug, and the `sys.path == sys_path_after_first_call` idempotency assertion is a nice touch. If you keep the reordering approach, consider adding a case where the shadow sits *before* site-packages and asserting `agent_path` does **not** jump ahead of site-packages — that's the regression the current insert can cause.

luckfu:
Updated in 367278c2 following your feedback.

Changes:
- replaced the global sys.path reordering with an explicit file-spec load of the agent cron package
- still evicts a stale wrong-origin cron/cron.* module under the same process-wide lock
- preserves api/config.py's end-of-sys.path invariant for _AGENT_DIR and does not change import priority for pydantic/openai/etc.
- strengthened the regression test to assert sys.path is unchanged both on the initial repair and on the fast-path second call

Verification:
- python3 -m py_compile api/routes.py
- PYTEST_ADDOPTS='-o cache_dir=/tmp/hermes-webui-pytest-cache' python3 -m pytest tests/test_cron_import_shadowing.py tests/test_issue4768_cron_module_missing.py tests/test_cron_toast_notifications.py tests/test_cron_query_param_validation.py tests/test_cron_model_override.py tests/test_cron_model_provider_picker.py -q
- result: 22 passed

luckfu:
Rebased on current master and resolved the conflicts in api/routes.py and tests/test_cron_import_shadowing.py.

Current head: 93a8c5c8
PR is mergeable again. Focused cron tests still pass locally after conflict resolution.

nesquena-hermes:
Thanks @luckfu — this shipped in **v0.51.627**. 🎉

Your root-cause analysis (an unrelated top-level `cron` package shadowing the agent's on `sys.path`, so `import cron` resolved to the wrong package and the Task APIs returned `cron_unavailable`) was spot-on. The fix is in master: `_ensure_agent_cron_import_path()` in `api/routes.py` re-orders `sys.path` so the agent's cron package takes precedence (inserting ahead of any shadowing entry) and clears a previously-imported unrelated `cron` module before retrying, with `tests/test_cron_import_shadowing.py` for regression coverage. Closing as shipped — credited in the CHANGELOG. Apologies it was left open.

Issue #484, opened 2026-04-14
feat(renderer): collapsible JSON/YAML tree viewer


nesquena-hermes:
## Summary
When the agent dumps API responses, config files, or structured data, the output is a flat code block that can be hundreds of lines long. A collapsible, syntax-colored tree viewer lets you navigate large nested objects without scrolling through walls of text.

## Proposed behavior
Fenced code blocks with `json` or `yaml` language hints get a toggle: **Tree** / **Raw**. Default is Tree for objects/arrays with more than ~20 lines; Raw for short ones.

Tree view:
- Keys are bold, values are colored by type (string=green, number=blue, boolean=amber, null=muted)
- Arrays and objects show a collapse triangle — click to fold/unfold
- Root level always expanded; nested levels collapsed by default beyond depth 2
- A "Copy" button in the header copies the raw JSON/YAML

Raw view: the existing syntax-highlighted `<pre><code>` block.

## Implementation
Lightweight — build a recursive DOM renderer from `JSON.parse()` output. No external library required. Roughly 80-100 lines of vanilla JS. YAML requires a parser; use a small bundled one (js-yaml ~20KB) loaded lazily via CDN only when a YAML block is actually present.

## Files involved
- `static/ui.js` — fenced code block renderer, JSON tree builder
- `static/style.css` — tree view styles
- Lazy-load: `https://cdn.jsdelivr.net/npm/js-yaml@4/dist/js-yaml.min.js` (YAML only)

nesquena-hermes:
## Acknowledged: Collapsible JSON/YAML tree viewer ✅

Good choice to build a recursive DOM renderer from `JSON.parse()` rather than pulling in a library — ~80-100 lines of vanilla JS is very achievable.

### Implementation notes

**JSON tree renderer** (no CDN needed):
```js
function buildJsonTree(value, depth = 0) {
  const el = document.createElement('span');
  if (value === null) { el.className = 'jv-null'; el.textContent = 'null'; }
  else if (typeof value === 'boolean') { el.className = 'jv-bool'; el.textContent = value; }
  else if (typeof value === 'number') { el.className = 'jv-num'; el.textContent = value; }
  else if (typeof value === 'string') { el.className = 'jv-str'; el.textContent = `"${value}"`; }
  else if (Array.isArray(value) || typeof value === 'object') {
    // render collapsible with <details><summary> — native collapse, no JS toggle needed
    const details = document.createElement('details');
[...]

PR #4858, opened 2026-06-24
fix(#484): drop noisy "parse failed" note on JSON/YAML fragments

metaember:
## Thinking Path

- Hermes WebUI renders LLM output as markdown; valid JSON/YAML fences get a Tree-view affordance (#484).
- LLMs constantly emit **fragments** in those fences — a bare `"key": "value"` line, snippets with `...`, trailing commas — which are not valid top-level JSON.
- `initTreeViews()` reacted by stamping an unstyled `<div class="tree-parse-note">parse failed</div>` *below* the block, producing bare white text under an otherwise-clean code box.
- This removes that note: such blocks now fall through silently to the existing syntax-highlighted raw view.

## What Changed

- `static/ui.js` — removed the `.tree-parse-note` insertion in `initTreeViews()`. The non-parseable / non-object branch now returns straight to raw view.
- The per-block **Raw/Tree toggle**, the **YAML lazy-load**, and the **#484 default-view config** (`structured_code_default_view` / `auto_tree_lines`) are all unchanged.
- The `parse_failed_note` i18n key is **intentionally retained** in all 13 locales (kept green by `test_chinese_locale.py`) to avoid a 13-file churn for a now-unused string.

### Screenshots

before:

before

after: 

after

## Why It Matters

The note was incidental, not designed: it was bolted onto the YAML lazy-load commit (`29a31a6e`), has **no CSS** (hence the bare white text), **no behavior test**, and originally even fired on *valid* YAML before the CDN script finished loading (later fixed). The block always renders fine as raw, so for the common LLM-fragment case the note added only visual noise.

## Verification

```
./scripts/test.sh tests/test_issue484_json_tree_viewer.py tests/test_chinese_locale.py
# 29 passed
```

Before/after rendered through the app's own `renderMd()` + `postProcessRenderedMessages()` pipeline on a booted `server.py` (real header chrome + Prism colors). The "before" re-inserts the exact `tree-parse-note` div the old code produced.

**Before** — bare white "parse failed" under the box · **After** — clean box, no note.

> _Before/after screenshots attached below._

## Risks / Follow-ups

- Low risk: pure removal, nothing depends on the note.
- Not contract-affecting — the note isn't described in any contract doc / RFC / product-semantics test (grep of `docs/`, `ARCHITECTURE.md`, `DESIGN.md`, `ROADMAP.md`, `TESTING.md` is empty).
- Follow-up option: if a "not valid JSON" signal is ever wanted, the right version would be styled + tested (and ideally a lenient fragment parse that wraps bare `key: val` lines in `{}` so they render as a Tree).

## Model Used

Anthropic Claude, `claude-opus-4-8` (Claude Code).

## AI Usage Disclosure

AI-assisted. Provider: Anthropic. Model: `claude-opus-4-8`. Tooling: Claude Code (analysis, edit, Playwright screenshot capture, test run).

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -13624,15 +13624,12 @@ function initTreeViews(container){
     // Mark as initialised only after we've committed to a render decision
     wrap.setAttribute('data-tree-init','1');
     if(!parsed || typeof parsed!=='object'){
-      if(parseFailed){
-        const hint=wrap.querySelector('.tree-raw-view');
-        if(hint&&!hint.querySelector('.tree-parse-note')){
-          const note=document.createElement('div');
-          note.className='tree-parse-note';
-          note.textContent=t('parse_failed_note')||'parse failed';
-          hint.parentNode.insertBefore(note,hint.nextSibling);
-        }
-      }
+      // No tree view for non-object values or unparseable content. LLMs often
+      // emit JSON fragments (a bare "key": "val" line, snippets with ..., etc.)
+      // that legitimately fail JSON.parse; surfacing a "parse failed" note for
+      // those was pure noise. The block still renders as syntax-highlighted raw,
+      // so just fall through silently. (parseFailed is retained for clarity.)
+      void parseFailed;
       return; // leave as raw view
     }
     const lineCount=rawText.split('\n').length;


nesquena-hermes:
Pulled the branch at HEAD `69f5f27c6` and read the diff plus the surrounding `initTreeViews()` body (`static/ui.js:13601-13640`). The removal is clean and the reasoning checks out — confirming a few things and pushing back gently on the Greptile dead-code note.

## The removal is safe

`.tree-parse-note` has no CSS anywhere — I grepped `static/style.css` and there's no rule for it, which confirms the "bare white text" symptom in the before-shot was real and not a theming gap. There are also no remaining references to `tree-parse-note` anywhere in `static/` or `tests/` at HEAD, so this is a complete removal with no dangling consumers. And the non-object branch always fell through to the syntax-highlighted raw view regardless, so dropping the note can't regress rendering.

```js
if(!parsed || typeof parsed!=='object'){
  // ...
  void parseFailed;
  return; // leave as raw view
}
```

## On the `void parseFailed;` — the linter premise doesn't hold here

The PR body and Greptile both justify keeping `parseFailed` + the `void` no-op as appeasing an "assigned but never read" linter rule. That rule isn't active in this repo. The only ESLint config is `eslint.runtime-guard.config.mjs`, and it deliberately enforces just two rules:

```js
rules: {
  "no-const-assign": "error",
  "no-import-assign": "error",
},
```

Its header is explicit that it's "NOT a style linter" — it only catches runtime-throw bugs and is verified green on the clean tree. `no-unused-vars` isn't in scope, and `tests/test_static_js_runtime_lint.py` runs that exact config. So nothing requires the `void` keepalive.

That means you can simplify: drop `let parseFailed=false;` (13608), the two `parseFailed=...` assignments (13610, 13614), and the `void parseFailed;` line (13632), leaving just the explanatory comment. The two assignments are pure dead writes now — neither the JSON nor the YAML branch reads `parseFailed`; the render decision is driven entirely by `parsed`/`typeof parsed`/`lang`. The YAML defer path returns before the note block anyway, so removing the flag changes no control flow.

Not a blocker — the `void` form is harmless at runtime. It's just leaner to delete the flag outright since the linter constraint that motivated keeping it isn't present.

## Tests

Good call retaining the `parse_failed_note` i18n key across the 13 locales to avoid churn (and keep `test_chinese_locale.py` green). I confirmed `test_issue484_json_tree_viewer.py` has no assertion on the note text, so the removal doesn't need a test update — though a tiny positive assertion that a JSON-fragment fence renders raw *without* a `.tree-parse-note` child would lock the behavior in cheaply, if you want a regression guard. Per the worktree execution ban I didn't run the suite myself.

nesquena-hermes:
Shipped in **v0.51.635** (Release WP) — thanks @metaember! 🎉 (#484)

Clean pure-removal. Gate: Codex SAFE TO SHIP (valid-JSON Tree rendering unchanged, nothing else references the removed `.tree-parse-note`, Raw/Tree toggle + #484 default-view config unchanged, the retained `parse_failed_note` i18n key keeps locale-parity green), full suite 10461 passed. Verified on prod — the served bundle no longer carries the note. Fragment code blocks now render clean with no bare 'parse failed' text.

PR #4886, opened 2026-06-25
fix: tab-visibility-chip bg in graphite

luandnh:
## What Changed

- Changed `.tab-visibility-chip` from using `--accent` as background to using `--accent-bg` (tint) + `--accent-text`
- In graphite light, `--accent` is `#303030` (near-black) - using it as background made the hardcoded `#1a1a1a` text invisible
- Using `--accent-bg` (8% tint) + `--accent-text` matches the standard chip pattern from `.chip.model` and `.app-dialog-btn.confirm:hover`

## Verification

`--accent-bg` and `--accent-text` are defined in every skin block, so the fix works across all themes.

## AI Usage

- Provider: Hermes Agent (Nous Research) - Model: deepseek-v4-flash via opencode-go

### DIFF
--- a/api/providers.py
+++ b/api/providers.py
@@ -2864,8 +2864,13 @@ def set_custom_provider_models(provider_id: str, models: list[str]) -> dict[str,
                     existing_models = cp.get("models")
                     if isinstance(existing_models, dict):
                         if models:
-                            # Rebuild dict with empty metadata for each model
-                            cp["models"] = {m: {} for m in models}
+                            # Preserve any per-model metadata from the original
+                            # dict (context_length, label, etc.) for models that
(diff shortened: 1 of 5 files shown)


luandnh:
Fix pushed in 28b5548:

- Changed `.tab-visibility-chip` from `background: var(--accent, var(--link))` to `background: var(--accent-bg)` and `color: var(--accent-text)`
- In graphite light, `--accent` is `#303030` (near-black) - using it as background made the hardcoded `#1a1a1a` text unreadable
- Using `--accent-bg` (light tint) + `--accent-text` matches the pattern used by `.chip.model` and `.app-dialog-btn.confirm:hover` across the codebase

Ready for review.

PR #4888, opened 2026-06-25
fix: tab-visibility-chip background in graphite light theme

luandnh:
## What Changed

- `.tab-visibility-chip` in graphite light theme had `background: var(--accent)` which resolves to `#303030` (near-black), making the hardcoded `#1a1a1a` text invisible.
- Fixed by using `background: var(--accent-bg)` (light tint) and `color: var(--accent-text)`
- This matches the standard chip pattern used elsewhere (`.chip.model`, `.app-dialog-btn.confirm:hover`)

## Verification

`--accent-bg` and `--accent-text` are defined in every skin block, so the fix works across all themes.


nesquena-hermes:
Confirmed the bug and the fix by reading the rule on `master` at `static/style.css:4330`:

```css
.tab-visibility-chip{...border:1px solid var(--accent,var(--link));background:var(--accent,var(--link));color:#1a1a1a;font-weight:600;...}
```

In graphite light `--accent` resolves to `#303030` (style.css:119), so a near-black fill under hardcoded `#1a1a1a` text is effectively invisible — exactly as reported. The root issue is the hardcoded `color:#1a1a1a`, which assumes every skin's accent is light enough to read dark text against; graphite (and dark skins generally) break that assumption.

## The fix is idiomatic — it matches the existing active-chip convention exactly

The replacement trio isn't just a guess; it's the same pattern already used for every other "active/selected chip" in the stylesheet:

```css
:root:not(.dark) .project-chip.active{background:var(--accent-bg);color:var(--accent-text);border-color:var(--accent-bg-strong);}   /* :855 */
.chip.model{color:var(--accent-text);border-color:var(--accent-bg-strong);background:var(--accent-bg);}                          /* :857, :1713 */
```

So `.tab-visibility-chip` now reads identically to `.chip.model` / `.project-chip.active`. All three custom properties are defined in base `:root` (style.css:10), `:root.dark` (:97), and every skin block — I checked graphite specifically (`--accent-bg:rgba(0,0,0,0.07); --accent-bg-strong:rgba(0,0,0,0.13); --accent-text:#303030` at :119) — so there's no undefined-variable fallback risk. The `.chip-off` state (style.css:4337) stays transparent + `--muted` + `--border` + `font-weight:400`, and the on-state keeps `font-weight:600`, so on/off is still distinguished by fill + text color + border + weight. The fix is correct.

## One tradeoff to be aware of (not a blocker)

The previous on-state was a *solid* accent fill — a strong, high-prominence "this tab is enabled" affordance. The new tinted version (`--accent-bg` is a ~0.07–0.09 alpha tint) is much quieter, so the visual weight of an enabled chip drops noticeably. That's fine if you want it to match the other tinted chips, but if you'd rather keep the bolder solid-fill look *and* fix contrast, there's already a token for text-on-solid-accent: `--panel-head-primary-fg` (default `#fff`), used by the Save button at style.css:5546:

```css
.panel-head-btn.primary{background:var(--accent,var(--link));color:var(--panel-head-primary-fg,#fff);border:none;}
```

i.e. an alternative would be `background:var(--accent);color:var(--panel-head-primary-fg,#fff);border-color:var(--accent)`, which keeps the prominent fill and reads in both graphite light (white on `#303030`) and gold light (white on `#B8860B`). Either approach fixes the bug — the PR's tinted route is more consistent with the rest of the chip family, the solid route preserves the stronger on/off contrast. Your call on which signaling you want for the visibility toggles.

No behavioral logic touched, single-line change, regression risk is essentially nil. Couldn't run a live theme render here, so the readability claim is from the resolved variable values in the cascade rather than a screenshot — though the cutter preview on this PR appears to confirm it.

luandnh:
@nesquena-hermes this PR has been reviewed and approved — ready to ship. Single-line CSS fix using the standard `--accent-bg`/`--accent-text`/`--accent-bg-strong` pattern, no behavioral logic touched, regression risk nil.

nesquena-hermes:
Gate-clean + verified, holding for your visual sign-off (it restyles the base `.tab-visibility-chip` across all themes, so flagging for an eyeball before merge).

**Root cause confirmed empirically:** in the graphite-light skin `--accent` resolves to `#303030` (near-black), so the old `background:var(--accent); color:#1a1a1a` rendered near-black text on a near-black fill — invisible. The fix (`background:var(--accent-bg); color:var(--accent-text); border:var(--accent-bg-strong)`) renders `#303030` text on a light 7%-black tint = clearly readable. I probed the computed colors directly to confirm.

**Scope check:** `--accent-bg` / `--accent-text` / `--accent-bg-strong` are defined in all 29 skin blocks (full parity), so the chip renders correctly in every theme — it changes the selected/visible chip from a solid-accent fill to the outlined-tinted style already used by the buttons.

Gate: Codex **SAFE TO SHIP**, full suite **10558 passed**. Screenshot of the chips (Sidebar tabs / Chat footer / Situational footer) looks clean and readable. Ready to ship on your OK — the only open question is the purely-visual call of outlined-vs-solid chips across themes. Thanks @luandnh.

luandnh:
Ok @nesquena-hermes

nesquena-hermes:
Shipped in **v0.51.693** (tab-visibility chip contrast in graphite-light) — merged via release PR #5034 and deployed to prod. Thanks @luandnh! Full Codex regression gate (SAFE TO SHIP) + full pytest suite (10754 passed, 0 failed) green.

Issue #4808, opened 2026-06-24
Performance issues are back


batkuip:
https://github.com/nesquena/hermes-webui/issues/4672
https://github.com/nesquena/hermes-webui/issues/4650

these worked for a day - now we are back to 100% cpu use.  Mostly /api/sessions taking at least 2 sec to complete which seems to slow everything else down (including token rendering).

nesquena-hermes:
This is the same symptom class as the now-closed #4672 (`/api/sessions` multi-second + 100% CPU, dragging token rendering). The fix for #4672 (`a66913ac2`, "freeze /api/sessions cache during streaming") is a *streaming hold-down* on the cache stamp — and the "worked for a day, now back" pattern is a strong tell that something is bypassing that hold-down again. To fix it for good we need to know which stage is eating the time, and the build already has the instrumentation to tell us.

## What the hold-down actually guards (and how it can leak)

`/api/sessions` is served from `_session_list_cache`. During a live turn the gateway/CLI writes a message row per token, which would normally bust the cache stamp on every poll and force a full `all_sessions()` rebuild mid-stream — that rebuild contends for `LOCK` with the streaming writer and is exactly what drags output to ~2 tok/s. The #4672 fix collapses the volatile state.db-derived stamp to a constant marker while a stream is in flight (`api/routes.py:1737` / `1771-1781`):

```python
streaming_marker = _session_list_cache_streaming_freeze_marker()
if streaming_marker is not None:
    return (streaming_marker, streaming_marker, streaming_marker,
            streaming_marker, _session_list_cache_path_stamp(settings_file),
            streaming_marker, swv)
```

The whole thing hinges on `_session_list_cache_streaming_freeze_marker()` returning non-None during a turn, which depends on `_active_stream_ids()` (`api/models.py:378`) being non-empty:

```python
def _active_stream_ids():
    with STREAMS_LOCK:
        active_ids = set(STREAMS.keys())
    with _cfg.ACTIVE_RUNS_LOCK:
        active_ids.update(_cfg.ACTIVE_RUNS.keys())
    return active_ids
```

[...]

PR #4889, opened 2026-06-25
perf(#4842): cache sidebar-projection sidecar reads + hoist per-row I/O on the cron path

### LINKED ISSUE #4842: Still running 100% cpu
https://github.com/nesquena/hermes-webui/issues/4808

still not fixed - 12s api/session calls - sometimes it works for a couple of minutes, often at the start of the session. but soon after it slows down to a crawl again.

maybe you want to escalate the issue to your owner?  this is the 4th ticket on the issue.

nesquena-hermes:
## What this fixes

`GET /api/sessions` pins CPU to 100% and takes multiple seconds on profiles with **many cron sessions** — the recurring symptom from #4672 → #4808 → **#4842** (the reporter's 4th ticket on it). The earlier fixes each addressed a different layer (the `/api/reasoning` storm in #4650, the cache re-fingerprinting in #4672, the cache-miss-vs-poll cadence in #4808), but the reporter's stage trace pins **all** remaining latency in a single stage:

```
{"ms": 2601.8, "name": "get_cli_sessions"}   ... next poll ...
{"ms": 3106.2, "name": "get_cli_sessions"}
```

His profile: **200+ cron sessions, zero Claude Code sessions, slow even with one visible session.** That last detail is the key — the visible-window cap (`CLI_VISIBLE_SESSION_LIMIT = 20`) doesn't bound the cost, because the cron projection runs a **second, higher-capped pass** (`CRON_PROJECT_CHIP_LIMIT = 200`) independent of how many sessions are filtered into view.

## Root cause

The state.db → sidebar projection in `_load_cli_sessions_uncached` does three pieces of redundant per-row file I/O, in **both** the visible pass and the up-to-200-row cron-only second pass:

1. **`_state_projection_sidecar_metadata(sid)`** — an uncached `open()` + 64 KB prefix read + a pure-Python JSON-key scan, once per row.
2. **`get_last_workspace()`** — up to two file reads + an `is_dir()`/remote probe, returning the **same** active workspace for every row.
3. **`cron/jobs.json`** — a full file read + JSON parse for **each** untitled cron row.

On a cron-heavy install that's hundreds of file operations per build. And because the enclosing `_CLI_SESSIONS_CACHE` (5 s TTL) is keyed on `_sqlite_content_fingerprint` = `MAX(rowid)` of the `messages` table — which advances on **every streamed token** — the cache misses on essentially every poll during a live turn, so the whole 200-row scan is re-paid every ~5 s. That's the 100% CPU + multi-second `get_cli_sessions`.

This is a distinct hot path from #4822 (which caches the *Claude Code* transcript parse). On a profile with zero Claude Code sessions, #4822's cache never engages — so it does not help this reporter. This PR is the cron/CLI-side counterpart.

## The fix

1. **Memoize `_state_projection_sidecar_metadata`** in a bounded module-level cache keyed by the sidecar's `(path, mtime_ns, size, ctime_ns)` stat signature. A warm build re-stats each file (~1 syscall) instead of re-reading+parsing it. Any rename/archive/edit bumps the signature and transparently invalidates just that one entry, so a stale title/archived flag is impossible without a re-read. Crucially this cache is **separate** from `_CLI_SESSIONS_CACHE`, so it survives the per-token fingerprint busting that defeats the 5 s TTL during streaming — which is the actual trigger. Mirrors the Claude Code parse cache (#4718). Returns dict copies so callers can't corrupt it; folded into `clear_cli_sessions_cache()` for explicit-reset/test parity.
2. **Build-scoped memos** for `cron/jobs.json` (parsed once into a `job_id → name` map) and `get_last_workspace()` (resolved once), matching the existing `_cron_pid()` closure pattern already in the function.

Output is unchanged — these are pure I/O-dedup hoists. The expensive per-row work is now **O(unique files)** instead of **O(rows × polls)**.

## Tests

New `tests/test_issue4842_cron_projection_perf.py` (7 tests):
- warm rebuild does **not** re-read sidecars (the streaming re-poll scenario)
- stat-change (rename/archive) invalidates and returns fresh metadata
- returned dicts are independent copies (no cache corruption)
- missing sidecar returns the default and isn't cached
- `jobs.json` read ≤ 1× per build despite many untitled cron rows
- `get_last_workspace()` resolved ≤ 1× per build
- `clear_cli_sessions_cache()` also clears the sidecar cache

Also updated the two `#4385` archive-reappears regression tests to write a **real** archived sidecar JSON under a monkeypatched `SESSION_DIR` (rather than mocking `load_metadata_only` with no file on disk). The perf fix stat-gates on `SESSION_DIR/{sid}.json` before consulting metadata — which exactly mirrors production, where archived metadata only exists when that file does — so the tests now exercise the real read path. This is an intentional move of the projection's mock boundary from `load_metadata_only` to the filesystem; any other test mocking at that boundary is affected.

Full suite green locally (10481 passed after the test fix); gated with Codex (SAFE TO SHIP) + Opus (the #4385 test breakage was Opus's blocking finding, now fixed).

Closes #4842.

---

*Self-built maintainer PR (nesquena-hermes). Requesting an independent human review before merge per the self-built-PR policy. @rodboev / @franksong2702 — if either of you has a cron-heavy profile, a before/after `get_cli_sessions` stage timing on `/api/sessions` during a live turn would be the ideal real-world confirmation.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`GET /api/sessions` no longer pins CPU to 100% and takes multiple seconds on profiles with many cron sessions.** The sidebar projection that turns agent state.db rows into sidebar entries did three pieces of redundant per-row file I/O — an uncached sidecar JSON read (open + 64KB prefix read + key scan), a `get_last_workspace()` resolve (up to two file reads + a directory probe), and a full `cron/jobs.json` read+parse for each untitled cron row — across both the visible pass and the higher-capped (up to 200 rows) cron-only second pass. On a cron-heavy install that was hundreds of file reads per build, and because the session-list cache is keyed on a state.db fingerprint that advances on every streamed message, the whole scan was re-paid on essentially every 5s poll during a live turn (the recurring "100% CPU / slow `/api/sessions`" from #4672 → #4808 → #4842). The sidecar metadata read is now memoized per file by its `(path, mtime_ns, size, ctime_ns)` stat signature so a warm build re-stats instead of re-reads (a rename/archive/edit still invalidates just that one entry), and the workspace resolve + jobs.json parse now happen once per build instead of once per row. Output is unchanged. (#4842)
+
 ## [v0.51.638] — 2026-06-25 — Release WS (faster session list — bounded continuation-fallback sidecar reads)
 
 ### Fixed
--- a/api/models.py
+++ b/api/models.py
@@ -41,6 +41,26 @@ _CLI_SESSIONS_CACHE_TTL_SECONDS = 5.0
 _CLI_SESSIONS_CACHE_LOCK = threading.Lock()
 _CLI_SESSIONS_CACHE = {}
 
+# Per-file cache for the UI-owned sidecar metadata (title + archived) that the
+# state.db sidebar projection overlays onto each CLI/cron row (#4842). The
+# projection calls _state_projection_sidecar_metadata() once per row in BOTH
+# the main visible pass AND the higher-capped (CRON_PROJECT_CHIP_LIMIT=200)
+# cron-only second pass, and each call was an uncached open() + 64KB prefix
+# read + a pure-Python JSON-key scan. On a cron-heavy profile that is up to
+# ~200 sidecar file reads per /api/sessions build — and because the enclosing
+# _CLI_SESSIONS_CACHE is keyed on a state.db content fingerprint that advances
+# on every streamed message row, that whole scan was re-paid on essentially
+# every 5s poll during a live turn (the "100% CPU / multi-second get_cli_sessions"
+# in #4842/#4808/#4672). This memoizes the parse result keyed by the sidecar's
+# (path, mtime_ns, size, ctime_ns) stat signature: a warm projection re-stats
+# each file (~1 stat) instead of re-reading+parsing it, while any genuine
+# rename/archive/edit bumps the signature and transparently invalidates just
+# that one entry. Bounded so a pathological session store can't grow it without
+# limit. Mirrors the Claude Code parse cache (#4718).
+_SIDECAR_METADATA_CACHE_LOCK = threading.Lock()
+_SIDECAR_METADATA_CACHE: "collections.OrderedDict[tuple, dict]" = collections.OrderedDict()
+_SIDECAR_METADATA_CACHE_MAX = 2000
+
 # ---------------------------------------------------------------------------
 # Stale temp-file cleanup
 # ---------------------------------------------------------------------------
@@ -3971,6 +3991,10 @@ def get_claude_code_session_messages(sid, projects_dir: Path | str | None = None
 def clear_cli_sessions_cache() -> None:
     with _CLI_SESSIONS_CACHE_LOCK:
         _CLI_SESSIONS_CACHE.clear()
+    # The sidecar-metadata projection cache is stat-keyed (self-invalidating on
+    # any file change), but clear it alongside the CLI cache so an explicit
+    # reset — a mutating sidebar action or test isolation — starts fully cold.
+    clear_sidecar_metadata_cache()
 
 
 def _copy_cli_sessions(sessions: list) -> list:
@@ -4176,20 +4200,61 @@ def _all_profiles_cli_contexts() -> tuple[list[tuple[Path, Path, str | None]], t
     return contexts, tuple(cache_entries)
 
 
+def clear_sidecar_metadata_cache() -> None:
+    """Drop all memoized sidebar-projection sidecar metadata (test/lifecycle hook)."""
+    with _SIDECAR_METADATA_CACHE_LOCK:
+        _SIDECAR_METADATA_CACHE.clear()
+
+
 def _state_projection_sidecar_metadata(sid: str) -> dict:
-    """Return UI-owned metadata for a state.db-projected sidebar row."""
-    metadata = {"title": None, "archived": False}
+    """Return UI-owned metadata (title + archived) for a state.db-projected row.
+
+    Memoized by the sidecar file's (path, mtime_ns, size, ctime_ns) stat
+    signature so the sidebar projection — which calls this once per row in both
+    the visible pass and the up-to-200-row cron pass — pays a single os.stat per
+    file on a warm build instead of an open() + 64KB read + JSON-key scan
+    (#4842). A rename/archive/edit bumps the signature and invalidates just that
+    entry, so a stale title/archived flag is impossible without re-reading.
+    Returns a COPY so callers can't mutate the cached dict.
+    """
+    default = {"title": None, "archived": False}
+    if not is_safe_session_id(sid):
+        return dict(default)
+    p = SESSION_DIR / f'{sid}.json'
+    try:
+        st = p.stat()
+        key = (str(p), st.st_mtime_ns, st.st_size, st.st_ctime_ns)
+    except OSError:
+        # No sidecar file (the common case for a pure state.db row) or it
+        # vanished mid-build — nothing to project, and nothing worth caching.
+        return dict(default)
+
+    with _SIDECAR_METADATA_CACHE_LOCK:
+        hit = _SIDECAR_METADATA_CACHE.get(key)
+        if hit is not None:
+            _SIDECAR_METADATA_CACHE.move_to_end(key)
+            return dict(hit)
+
+    metadata = dict(default)
     try:
         webui_meta = Session.load_metadata_only(sid)
     except Exception:
-        return metadata
-    if not webui_meta:
-        return metadata
-    title = getattr(webui_meta, 'title', None)
-    if title:
-        metadata["title"] = title
-    metadata["archived"] = bool(getattr(webui_meta, 'archived', False))
-    return metadata
+        webui_meta = None
+    if webui_meta:
+        title = getattr(webui_meta, 'title', None)
+        if title:
+            metadata["title"] = title
+        metadata["archived"] = bool(getattr(webui_meta, 'archived', False))
+
+    with _SIDECAR_METADATA_CACHE_LOCK:
+        # Re-check under lock in case a concurrent build populated it; either
+        # entry is equally valid for the same stat signature.
+        if key not in _SIDECAR_METADATA_CACHE:
+            _SIDECAR_METADATA_CACHE[key] = metadata
+            _SIDECAR_METADATA_CACHE.move_to_end(key)
+            while len(_SIDECAR_METADATA_CACHE) > _SIDECAR_METADATA_CACHE_MAX:
+                _SIDECAR_METADATA_CACHE.popitem(last=False)
+    return dict(metadata)
 
 
 def _load_cli_sessions_uncached(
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
## Gate summary

Ran the full self-built gate (Codex + Opus + full suite):

- **Codex: SAFE TO SHIP.** It executed the targeted tests (12 passed) and verified all six review points — staleness/stat-key invalidation, jobs.json refactor equivalence (ids are always `uuid.hex` strings so the `str()` keying matches the old `==` compare), the `cron_`/`len(parts)<3` guards, the lock/LRU correctness, the build-local closure memos, and the missing-sidecar path.

- **Opus: one BLOCKING finding (now fixed).** The new stat-gate in `_state_projection_sidecar_metadata` short-circuits on `SESSION_DIR/{sid}.json` before consulting metadata, which broke two `#4385` archive-reappears regression tests that mocked `load_metadata_only` with **no file on disk**. Opus correctly flagged this is **not** a production regression (in prod, `load_metadata_only` reads exactly that path and returns `None` when absent — so archived metadata only exists when the file does, in which case the `stat()` succeeds and behavior is identical), only a mock-boundary mismatch. Fixed by rewriting both tests to write a real archived sidecar under a monkeypatched `SESSION_DIR`, exercising the actual production read path. Also documented the `SESSION_DIR`-as-sole-source assumption on the cache function per Opus's forward-looking note.

- **Full suite confirmed the same two failures** before the test fix (independent of the gate), and is green after.

Holding for an independent human review before merge (self-built). The change is pure I/O-dedup — output is byte-identical — so the risk surface is staleness, which the stat-key + the `clear_cli_sessions_cache()` fold both cover.

nesquena (approved):

Independent review of the cron-heavy `/api/sessions` perf fix (#4842 — the reporter's 4th ticket on the recurring 100% CPU / multi-second `get_cli_sessions`). Same family as #4822/#4718 but a **distinct hot path**: this is the cron/CLI-side projection, which #4822's Claude-Code parse cache never touches (the reporter has 200+ cron sessions, zero Claude Code).

### What this ships

`api/models.py` (+131/-44), `tests/test_issue4842_cron_projection_perf.py` (+285, new), `tests/test_issue4385_cron_archive_reappears.py` (+24/-16), `CHANGELOG.md` (+4). Agent-authored; Codex (SAFE) + Opus (one blocking test-fix, resolved). Self-built — requesting independent human review per policy; my `event=APPROVE` is the code-review gate. Closes #4842.

Fresh tarball: the projection reads the agent's `state.db` (read-only), the webui's per-session sidecar JSON (`SESSION_DIR/{sid}.json`), and the webui's `cron/jobs.json`. The cache memoizes reads; no agent-side write, no protocol/schema surface. The sidecar metadata is per-sid (global `SESSION_DIR`), profile-independent at this layer (profile scoping is a filter higher up), so keying the cache by the sidecar path is correct across profiles. ✓

### Root cause + fix

`_load_cli_sessions_uncached` did three pieces of per-row file I/O in **both** the visible pass and the up-to-200-row cron-only second pass — sidecar `open()`+64KB read+JSON-scan, `get_last_workspace()`, and a full `cron/jobs.json` parse per untitled cron row. Because `_CLI_SESSIONS_CACHE` is keyed on a state.db fingerprint that advances on **every streamed token**, the whole scan was re-paid every ~5s poll during a live turn. The fix dedups all three:

1. **Sidecar metadata cache** — stat-keyed, separate from `_CLI_SESSIONS_CACHE` (so it survives the per-token busting that's the actual trigger).
2. **Build-scoped `cron/jobs.json` memo** — parsed once into a `job_id→name` map.
3. **Build-scoped `get_last_workspace()` memo** — resolved once.

### Sidecar metadata cache — correctness verified

Keyed `(str(path), st_mtime_ns, st_size, st_ctime_ns)` on `SESSION_DIR/{sid}.json`. I verified:
- **Staleness-proof:** any rename/archive/edit bumps mtime/size/ctime → cache miss → re-read. The `ctime_ns` also catches inode-metadata changes. Harness-confirmed (un-archive flips `archived` immediately).
- **Missing sidecar is NOT cached** (the common pure-state.db-row case) — so a later-written archive is picked up on the next build, never served a stale "absent". ✓
- **Copy-on-return is fully corruption-safe:** the cached value is a **flat dict** `{"title": str|None, "archived": bool}` — all immutable values — so the shallow `dict(hit)` copy means a caller mutating the result can't touch the cache, and there are no nested mutables to share (unlike #4822's message lists). Harness-confirmed.
- **Thread-safe** (double-checked locking; read outside the lock, set/evict under it), **LRU-bounded** (2000), and a new **`is_safe_session_id` guard** before building the path (path-traversal-safe — a small security improvement over the old direct `load_metadata_only(sid)` call).

### The stat-gate — production-equivalent, #4385 NOT reintroduced (the load-bearing check)

The new code short-circuits to the default (`archived=False`) when `SESSION_DIR/{sid}.json` can't be `stat()`'d, only consulting `load_metadata_only` when the file exists. The risk: if archived metadata could exist *without* the sidecar file, this would resurrect the #4385 archive-reappears bug. I verified it can't:

- `Session.load_metadata_only(sid)` ([models.py:920](api/models.py#L920)) sources title+archived **exclusively** from `SESSION_DIR/{sid}.json`: `if not p.exists(): return None`, and its legacy fallback `cls.load(sid)` reads the same file. So archived state lives *in* the file the gate checks for → the gate returns exactly what `load_metadata_only` would (file absent → `None` → default; file present → identical read).
- **Independent harness (9/9):** archived sidecar present → `archived=True`; un-archive (rewrite) → stat-change → `archived=False`; deleted sidecar → default `archived=False` (matches `load_metadata_only` None); unsafe sid → default (no traversal); mutation-of-return doesn't corrupt the cache; `clear_cli_sessions_cache` clears the sidecar cache.

The Opus blocking finding (two #4385 tests mocked `load_metadata_only` to return archived *without* a file — never reached in prod) is correctly characterized as a mock-boundary mismatch; the fix rewrites both to write a **real** archived sidecar under a monkeypatched `SESSION_DIR`, exercising the production read path (strictly stronger coverage). I confirmed the tests still assert the archived session stays hidden — legitimate strengthening, not masking. ✓

### Build-scoped memos — output-equivalent

- `_cron_job_names()`: parses `jobs.json` once into `{str(id): name}`; the lookup `(.get(parts[1]))` is equivalent to the old per-row `_j.get('id') == _job_id` compare (cron ids are `uuid.hex` strings, so `str()` keying matches), same `cron_`/`len(parts)>=3` guards. Degrades to the generic title on read failure. ✓
- `_cli_workspace()`: resolves `str(get_last_workspace())` once (no-arg → loop-invariant), used in both passes. Byte-identical. ✓
- Both use the list-as-cell closure pattern already established by `_cron_pid()` in the same function — build-local, no cross-build/thread aliasing.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Warm rebuild during a live turn (fingerprint busted) | sidecar cache survives → re-stat, not re-read ✅ (the actual fix) |
| Sidecar renamed/archived/edited | stat signature changes → re-read ✅ |
| Un-archive a session | stat-change → `archived=False` immediately ✅ |
| Pure state.db row (no sidecar) | default, not cached → later archive picked up ✅ |
| Archived cron session (real sidecar) | stays archived (#4385 preserved) ✅ |
| 200 untitled cron rows | `jobs.json` parsed once ✅ |
| Unsafe / traversal sid | default, no path build ✅ |
| Concurrent builds | lock-guarded; duplicate read harmless ✅ |
| `clear_cli_sessions_cache()` | also clears sidecar cache ✅ |

### Tests

- `tests/test_issue4842_cron_projection_perf.py` (7) + `test_issue4385_cron_archive_reappears.py` (updated) — **13/13**.
- CLI-session / cron / sidebar / #4385/#4808/#4672 blast-radius — **515 passed, 20 skipped**.
- **Independent sidecar harness: 9/9** (stat-gate production-equivalence, staleness, copy-safety, traversal guard, lifecycle clear).
- **Full suite: 10353 passed / 133 skipped / 0 failures** in 207s (deselected the pre-existing darwin CRLF flake).
- `py_compile` clean on models.py.

- The function's docstring documents the `SESSION_DIR`-as-sole-source assumption and tells a future maintainer to update the gate if metadata ever moves stores — exactly the right guard-rail given the stat-gate depends on it.
- Sidecar cache returns a flat dict — no future-mutation hazard like #4822's nested message dicts, so no deepcopy concern here.

### Recommendation

✅ **Approved clean (code review).** Parked at approval — ready for the release agent's merge/tag pipeline once a cron-heavy human/CDP before/after confirms the `get_cli_sessions` stage drop.

This is the correct cron-side counterpart to #4822: it dedups the three per-row file reads that dominate the cron projection and, crucially, caches the sidecar parse in a store *separate* from the per-token-busted `_CLI_SESSIONS_CACHE` so it actually survives a live turn. I verified the stat-gate is production-equivalent (archived state lives in the very file the gate checks, so #4385 can't recur — proven 9/9), the cache is staleness-proof and copy-safe (flat immutable dict), and the build memos are output-identical. The #4385 test migration to real sidecar files is a legitimate strengthening. Ship (pending a real-world cron-heavy timing confirmation).

rodboev:
I can profile after the extensions-related PR is tested E2E.

rodboev:
Tested locally on a real cron-active profile (Windows, Python in-process via scheduled task). This system has 1,771 rows in state.db (1,027 cron, 200 Claude Code, 544 other), a 76 MB state.db, and one cron job configured. Same setup from the [#4822 profiling](https://github.com/nesquena/hermes-webui/pull/4822#issuecomment-4788050781).

The cron sessions on this install have no sidecar `.json` files (pure state.db rows), so the sidecar cache's `stat()` short-circuits immediately on those. The measured gains come from the `get_last_workspace()` hoist and the `jobs.json` once-per-build hoist. Installations with sidecar files on their cron sessions would see additional savings from the metadata cache hits.

**Direct `_load_cli_sessions_uncached` microbenchmark (420 sessions returned, 10 rounds):**

| Condition | avg | min | max |
|---|---|---|---|
| WITH PR #4889 (sidecar cache warm) | 67ms | 66ms | 72ms |
| WITHOUT caches (cleared every round) | 546ms | 518ms | 574ms |
| **Speedup** | **8.1x** | | |

**Simulated live-turn polling (20 rounds, CLI cache cleared each round to simulate fingerprint bust):**

With the cache surviving across fingerprint busts: avg=70ms per poll. Without, each poll would cost ~545ms, which at 5s intervals is 11% CPU just for sidebar polling. With the PR that drops to 1.4%.

**Profile switch benchmark (5 rounds, cookie-aware, full scan, no `sidebar_source`):**

| Round | Direction | Switch+GET total | GET /sessions | Sessions |
|---|---|---|---|---|
| 1 | default → test-ep-sprint31 | 341ms | 46ms | 0 |
| 1 | test-ep-sprint31 → default | 94ms | 8ms | 251 |
| 2 | default → test-ep-sprint31 | 58ms | 4ms | 0 |
| 2 | test-ep-sprint31 → default | 58ms | 8ms | 251 |
| 3 | default → test-ep-sprint31 | 94ms | 4ms | 0 |
| 3 | test-ep-sprint31 → default | 128ms | 7ms | 251 |
| 4 | default → test-ep-sprint31 | 22ms | 4ms | 0 |
| 4 | test-ep-sprint31 → default | 106ms | 22ms | 251 |
| 5 | default → test-ep-sprint31 | 103ms | 61ms | 0 |
| 5 | test-ep-sprint31 → default | 113ms | 9ms | 251 |

**GET /sessions (full scan): min=4ms, max=61ms, avg=17ms**
**Switch+GET total: min=22ms, max=341ms, avg=112ms**

Cache verification (two rapid full-scan GETs): 25ms cold, 7ms warm.

Full unfiltered response (no `exclude_hidden`): 122ms for 451 sessions (200 cron + 200 Claude Code + 51 other). The cron projection path is fast enough that the full scan now matches the `sidebar_source=webui` path in practice (avg 17ms vs 26ms).

nesquena-hermes:
Shipped in **v0.51.645** (Release XA). Full gate: @rodboev independent benchmark (8.1x on the uncached path, 17ms avg GET /sessions on a real 1,771-row/76MB cron profile), Codex SAFE TO SHIP, Opus SHIP IT (all 6 adversarial vectors safe, coexists with the #4822 CC-parse cache), full suite 10538 passed. Verified `_SIDECAR_METADATA_CACHE` in the deployed source on prod.

Leaving #4842 open until the reporter confirms it resolves the hang on their install.

PR #4891, opened 2026-06-25
fix: task detail actions hidden on mobile PWA

luandnh:
## What Changed

- The task detail action buttons (Run, Pause, Edit, Delete, etc.) in the `main-view-actions` bar were invisible on mobile PWA when viewing a task detail.
- Root cause: CSS rule at line 5430 uses `:has(.main-view-title:empty)` to hide the header until the title is populated. On some mobile browsers and PWA WebViews, this pseudo-class does not re-evaluate after `textContent` is dynamically set by `_renderCronDetail`, leaving the header (and all action buttons) permanently hidden.
- Fix: explicitly set `header.style.display` in `_setCronHeaderButtons` — show on `read` mode, hide on `empty` mode — so header visibility is managed by JavaScript instead of relying solely on CSS `:has()` re-evaluation.

## Verification

- Desktop: header visibility unchanged (CSS :has() still works as before)
- Mobile PWA: header now explicitly shown by JS, bypassing :has() re-evaluation issue
- `_clearCronDetail` also explicitly hides the header


luandnh:
Fix pushed in 16be570:

- Added explicit `header.style.display = ''` in `_setCronHeaderButtons('read')`
- Added explicit `header.style.display = 'none'` in `_setCronHeaderButtons('empty')`
- This ensures the action bar is visible on mobile PWA where CSS `:has()` may not re-evaluate after dynamic textContent changes

Ready for review.

nesquena-hermes:
Reading the diff against `static/panels.js:798-829` (`_setCronHeaderButtons`), the auto-hide rule at `static/style.css:5442`, and the `#mainTasks` header markup in `static/index.html:817-836`, the diagnosis is right: the header carries the action bar, and its visibility is governed entirely by

```css
main.main > .main-view:not([id="mainChat"]):not([id="mainSettings"]) .main-view-header:has(.main-view-title:empty){display:none;}
```

`taskDetailTitle` starts empty and is filled via `textContent` in `_renderCronDetail`/`_renderCronForm`, so on an engine that doesn't re-invalidate `:has()` after that mutation the header stays hidden. Driving the visibility from JS is the correct shape, and all three mode branches are covered. Two things worth tightening before this lands:

## 1. `display = ''` falls back to the (possibly stale) stylesheet — `'flex'` is the guaranteed override

The read/create/edit branches set:

```js
if (header) header.style.display = '';
```

Setting `''` *removes* the inline declaration and hands control back to the cascade — i.e. back to the very `:has(...)` rule the PR body says fails to re-evaluate on mobile PWA. If that rule is genuinely stale, the cascaded value is still `display:none` and the header stays hidden. The only reason `''` can work is that writing to `.style` triggers a style recalc that happens to bust the stale match — which is incidental, not guaranteed.

A hard value sidesteps the dependency entirely. `.main-view-header` is natively `display:flex` (`static/style.css:5424-5427`), so:

```js
if (header) header.style.display = 'flex';   // read + create/edit branches
...
if (header) header.style.display = 'none';   // empty branch (as written)
```

`flex` wins over the stylesheet regardless of whether `:has()` re-evaluated, and matches the element's natural layout. That's the more robust fix for the exact symptom you're describing.

## 2. The same latent bug exists in skills / memory / workspaces / profiles

The `:has(.main-view-title:empty)` rule isn't Tasks-specific — it matches every non-chat/non-settings `.main-view`. Checking `static/index.html`, four sibling detail views have the identical shape (empty-initialized title + an action bar):

- `#mainSkills` → `skillDetailTitle` + `.main-view-actions` (index.html:785-786)
- `#mainMemory` → `memoryDetailTitle` + `.main-view-actions` (index.html:802-803)
- `#mainWorkspaces` → `workspaceDetailTitle` + `.main-view-actions` (index.html:867-868)
- `#mainProfiles` → `profileDetailTitle` + `.main-view-actions` (index.html:885-886)

A grep for `main-view-header` across `static/*.js` returns **nothing** outside this PR, so none of those four manage header display in JS — they all lean on the same CSS rule. If the mobile-PWA `:has()` quirk is real, their action bars are invisible on the same engines for the same reason; this PR just fixes the Tasks instance. (`#mainKanban`, `#mainInsights`, `#mainLogs` use static non-empty titles, so they never match `:empty` and are unaffected.)

Two ways to handle it: generalize the JS toggle into a shared helper the other detail renderers also call, or fix it once in CSS by not gating header visibility on `:has(.main-view-title:empty)` for views whose header always has actions. The CSS route would fix all five in one line and avoid four more JS touch-points, but either is reasonable.

Scoping this PR to Tasks is fine if that's the only view you can repro against — just flagging that the sibling views almost certainly need the same treatment, and a test asserting the header is visible after `_renderCronDetail` (rather than relying on `:empty`) would lock in the regression guard. Couldn't run mobile-PWA emulation here, so the engine-specific `:has()` behavior is from the markup/CSS contract rather than a live repro.

luandnh:
Updated per review:

- Changed `header.style.display = ''` → `'flex'` in all branches (read, create/edit, empty) so the inline declaration doesn't hand control back to the CSS cascade
- Generalized the same fix to all 4 sibling views hit by the same `:has()` re-evaluation bug: Skills, Memory, Workspaces, and Profiles — each `_set*HeaderButtons` now manages `header.style.display` in all three mode branches
- Added regression test `test_issue4891_header_display.py` covering all five functions, asserting `flex` in read/create branches and `none` in empty branch

Ready for re-review.

nesquena-hermes:
Shipped in **v0.51.647** (Release XC) — thanks @luandnh! 🎉

Solid root-cause fix (the `:has(:empty)` non-reevaluation on mobile WebViews). The dual gate caught 2 silent regressions in the first cut — read-only Memory sections (Project Context / External Notes) and the Profiles concept-help view lost their header/title because the explicit-hide landed in their branch. Fixed both (memory 'read' always shows the header, gating only the Edit button; added a 'help' header mode for the concept view) + added a non-vacuous regression test. Codex re-gate SAFE, full suite 10547 passed. Verified on prod.

PR #4896, opened 2026-06-25
fix: table pipe | inside inline code breaks markdown table rendering

luandnh:
## What Changed

- Fixed a bug in the smd markdown parser where `|` (pipe) characters inside inline code `` ` `` were treated as table column separators, breaking table rendering.
- Example: `` `updates.model = modelState.model || null` `` inside a table cell would create extra columns instead of rendering as inline code.
- Added `e.token !== m` (code span token) guards in both `|` handlers in the smd parser (`static/vendor/smd.min.js`):
  - Pattern 1: table start handler (creates row + cell tokens)
  - Pattern 2: table cell continuation handler

## Verification

- smd.min.js syntax verified via Node.js
- Both guard patterns present after modification


nesquena-hermes (changes requested):
Thanks for digging into the broken-table bug — the underlying problem is **real and worth fixing** (a `|` inside inline code in a table cell does break rendering). But after reproducing both the bug and this patch against the actual parsers, **this change does not fix it** — it patches the wrong rendering path, and the lines it adds are unreachable. Details below so the next revision lands it in the right place.

### 1. The streaming parser (`smd.min.js`) already renders this correctly — so the guards are a no-op

I drove the **unmodified `master`** `smd.min.js` against your exact example and it already keeps the code span in one intact cell:

```
input:  | field | expr |
        | --- | --- |
        | set | `updates.model = modelState.model || null` |

master smd cells: [" field ", " expr ", " set ", "[code] updates.model = modelState.model || null "]
PR     smd cells: [" field ", " expr ", " set ", "[code] updates.model = modelState.model || null "]   ← identical
```

Across 9 table+code shapes (spaced/tight pipes, header-cell code, leading-pipe-in-code, multi-pipe, double-pipe), fed both all-at-once **and** character-by-character (true streaming), the full token/text event stream from `master` and from this PR is **byte-for-byte identical**.

### 2. Why the new `e.token !== m` guards never fire (dead code)

I instrumented both patched `case "|"` handlers to record `e.token` every time they execute:

- Handler 1 (table-start): only ever sees `e.token === 1` (PARAGRAPH)
- Handler 2 (cell-continuation): only ever sees `e.token === 27` (TABLE)

Neither handler is ever entered while `e.token === m` (11 = `CODE_INLINE`), because inside a code span the parser dispatches on a different token state entirely. So the `else { e.text += n }` branch you added is **never reached** — which is exactly why the before/after output is identical.

### 3. The real bug is in `renderMd()` in `static/ui.js` (which this PR does not touch)

The path that actually breaks is the **non-streaming** `renderMd()` — used for stored/history messages (`ui.js` ~line 9429), the stream-fallback path, and BTW answers. Same input through `master`'s `renderMd()`:

```
<td><code>updates.model = modelState.model</td><td></td><td>null</code></td>
```

The `||` splits one cell into three. The cause is the table pass at `static/ui.js` ~line 5165–5183: `_protectPipes` shields `|` inside paren/square/curly bracket pairs before `split('|')`, but it does **not** shield `|` inside backtick inline-code spans.

### Fix-spec for the next revision

- Revert the `smd.min.js` change (it has no effect, and hand-editing the minified vendor bundle is best avoided).
- In `renderMd()` (`static/ui.js`, the `_protectPipes` / `parseRow` / `parseHeader` block around lines 5170–5183), protect `` `…` `` inline-code spans the same way bracket pairs are protected: before `split('|')`, swap pipes that fall between a matched pair of backticks for the `\x00PIPE\x00` sentinel, then restore after the split. (Mind the existing brace-glyph / `extractFunc` constraint noted in the comment above that block, and the CSV-table branch.)
- Add a regression test in `tests/test_renderer_js_behaviour.py` (it already drives the real `renderMd()` via node) asserting that `| set | \`a | b\` |` renders as a single `<td><code>a | b</code></td>` cell. A test that exercises the streaming `smd` path alone will pass on `master` and won't catch this.

Happy to re-review once it moves to `renderMd()` with a test that fails on `master` and passes with the fix.

nesquena-hermes (changes requested):
Thanks for the rework — you moved the fix to the right place. The previous version patched `static/vendor/smd.min.js`, which was a no-op for this bug (that path already rendered the example correctly; the breakage is in `renderMd()`). This re-push correctly reverts `smd.min.js` and targets `_protectPipes` in `static/ui.js` — that's the right file. 👍

**But the new regex introduces a regression: it merges adjacent inline-code table cells.** Re-requesting changes on that basis.

### The regression

The added branch:

```js
r=r.replace(/(`[^`]*)[|]([^`]*`)/g,(_,a,b)=>a+'\x00PIPE\x00'+b);
```

matches from one cell's *closing* backtick, **across the cell-separator pipe**, to the next cell's *opening* backtick. So it protects (hides) the separator between two code cells, and `split('|')` then can't split them.

I verified this in a clean-room harness that reimplements `_protectPipes` + `parseRow`'s `split('|')` semantics (old vs new), no app code executed:

| Input row | OLD (master) cells | NEW (this PR) cells |
|---|---|---|
| `` \| `a = b \|\| null` \| c \| `` | `["a = b", "", "null", "c"]` (broken) | `["a = b \|\| null", "c"]` ✅ |
| `` \| `x` \| `y` \| `` | `["x", "y"]` ✅ | `["x \| y"]` ❌ **merged** |
| `` \| `a` \| `b` \| `c` \| `` | `["a","b","c"]` ✅ | `["a \| b \| c"]` ❌ **all merged** |

So the targeted `||`-in-one-cell case is fixed, but any table row with **two or more inline-code cells** (`` \| `method` \| `endpoint` \| ``, `` \| `--flag` \| `value` \| ``) now collapses into a single cell. That shape is extremely common in real LLM table output, so this trades one bug for a more frequent one.

### Suggested fix

Protect pipes only *within* a complete inline-code span, by matching the whole `` `...` `` span and neutralizing pipes inside it (rather than matching across a separator):

```js
const _protectPipes=r=>{
  let prev;
  do{ prev=r;
    r=r.replace(/([([\x7b][^)\]\x7d]*)[|]([^)\]\x7d]*[)\]\x7d])/g,(_,a,b)=>a+'\x00PIPE\x00'+b);
  }while(r!==prev);
  // neutralize pipes inside each complete inline-code span:
  r=r.replace(/`[^`]*`/g, s=>s.replace(/\|/g,'\x00PIPE\x00'));
  return r;
};
```

Clean-room verified against the same cases — bug target fixed **and** adjacent code cells preserved:

| Input row | Corrected cells |
|---|---|
| `` \| `a = b \|\| null` \| c \| `` | `["a = b \|\| null", "c"]` ✅ |
| `` \| `x` \| `y` \| `` | `["x", "y"]` ✅ |
| `` \| `a\|b` \| plain \| `c\|d` \| `` | `["a\|b", "plain", "c\|d"]` ✅ |

### Test (required, currently missing)

Please add a JS-behaviour test that is **RED on master and GREEN with the fix**, covering both:
1. the `||`-inside-one-code-cell case (the original bug), and
2. a row with **multiple inline-code cells** (the regression this version would introduce).

A `tests/test_renderer_js_behaviour.py`-style test that drives `renderMd()` via a JS runtime (jsdom/node) is the right level — a smd-streaming-only test would pass on master and miss this entirely. The current re-push has no test at all.

Once the regex protects within-span only and the two-case test is green, this is good to go.

luandnh:
Updated per review:

- Reverted `smd.min.js` change (confirmed dead code, as analyzed)
- Fixed `_protectPipes` in `static/ui.js` — added backtick inline-code span protection alongside existing bracket-pair protection; tested with Node and verified the fix produces 2 cells instead of 4
- Added regression test `test_table_pipe_inside_inline_code_is_protected` in `tests/test_renderer_js_behaviour.py` that drives the real `renderMd()` via Node and asserts exactly 2 `<td>` cells

Ready for re-review.

luandnh:
Fix updated per review (round 2):

- Changed `_protectPipes` regex from backtick `` /(`[^`]*)[|]([^`]*`)/g `` to `<code>` tag `/(<code>[^<]*)[|]([^<]*<\/code>)/g` because the inline-code pass converts backtick spans to `<code>` BEFORE the table pass — so the old regex never found backticks
- Regression test in `test_renderer_js_behaviour.py` now passes (asserts exactly 2 `<td>` cells)

Ready for re-review.

nesquena-hermes:
Shipped in **v0.51.650** (Release XF) — thanks @luandnh, and nice persistence through the rework! 🎉

The round-2 insight was the key: backtick spans become `<code>…</code>` *before* the table pass runs, so matching the `<code>` shape (not backticks) is what catches it. Gate: Codex SAFE TO SHIP — ran an adversarial battery (multiple pipes, multiple code spans per row, adjacent cells, outside-cell separators, escaped `<`, malformed raw `<code>`); the do/while convergence terminates and the `[^<]*` bound prevents cross-cell over-matching. Full suite 10558 passed + the renderMd behaviour suite (58 tests) green. Verified on prod.

Issue #4553, opened 2026-06-20
bug(mobile): normal chat transcript can overflow viewport and allow sideways pan


MTGVim:
## Summary

Mobile WebUI can still enter a document-level horizontal overflow / sideways-pan state in the normal chat transcript, even when no Worklog/tool summary row is involved.

This is a broader recurrence / scope escape from #4064. #4066 constrained the Worklog summary row, but regular assistant-rendered text/Markdown can still push the page wider than the mobile viewport.

## Steps to reproduce

1. Open Hermes WebUI on Android Chrome or another narrow mobile viewport.
2. Use a normal chat session; do not rely on tool calls or Worklog rows.
3. Render a regular assistant response containing long text / Markdown-like sections.
4. Scroll vertically through the transcript and drag slightly to the right while interacting with the chat surface.

## Actual behavior

- The transcript/page can pan horizontally to the right.
- After panning, the left side of the chat is clipped off-screen.
- Content remains wider than the visual viewport, so the user can end up reading a shifted page where the left edge is missing.
- The composer and transcript visually remain in the shifted horizontal state.

## Expected behavior

- Normal mobile chat use should never create document/page-level horizontal scrolling.
- Vertical transcript scrolling should not permit sideways drag/pan.
- Wide child content should be contained inside the message column:
  - code/table/math/media may wrap, truncate, or scroll internally;
  - but they must not widen `.messages`, `.messages-inner`, `.main`, `body`, or the visual viewport.

## Why this is not just #4064

#4064 reported the same failure mode with a concrete Worklog summary trigger.

The current reproduction does **not** require Worklog/tool UI. It can happen with normal assistant content, so the invariant needs to move from a single Worklog-summary rule to transcript-level containment plus child renderer constraints.

Related:

- #4064 — original mobile sideways-pan report, fixed narrowly for Worklog summary rows
- #4066 — Worklog summary overflow fix

A follow-up comment with the same evidence was also added to #4064, but GitHub refused reopening that closed issue.

## Suspected frontend areas

Potential areas to audit:

- `.messages`, `.messages-inner`, `.msg-row`, `.msg-body` width/min-width/overflow containment.
- Markdown child renderers:
  - headings / horizontal rules
  - tables
  - KaTeX blocks
  - media and artifact wrappers
  - diff/tree/code blocks
  - any generated block-level content
[...]

PR #4898, opened 2026-06-25
fix(#4856): mobile transcript stuck at top & unscrollable — overflow-x:hidden coerces overflow-y:auto into a collapsing scroll container

### LINKED ISSUE #4856: bug: WebUI is unusable on Android - transcript jumps to top on every interaction (regression in v0.51.576)
## Summary

Since v0.51.576, Hermes WebUI is effectively broken on Android (confirmed on Edge and Chrome Android, both Chromium-based). The transcript jumps to the very first message (scrollTop = 0) on every meaningful interaction, making it impossible to read or hold a conversation on a phone.

This happens across all three core usage patterns:

1. **Opening an existing session** - lands at the top instead of the bottom
2. **Sending a message** - transcript jumps to top immediately after send
3. **Response completion** - transcript jumps to top when streaming finishes

Every single message exchange requires manually scrolling back down to find your place. On a long session that is a lot of scrolling. The app is not usable on Android mobile in this state.

v0.51.575 works correctly. v0.51.576 is the first broken version. Desktop is completely unaffected.

---

## Environment

- **Platform:** Edge Android (Chromium-based mobile browser)
- **Last working version:** v0.51.575
- **First broken version:** v0.51.576

---

## Steps to reproduce

1. Open Hermes WebUI on any Android browser (Edge, Chrome)
2. Open any existing session with several messages
3. Transcript lands at the top instead of the bottom - you have to scroll the entire conversation to find the latest message
4. Send a message - transcript jumps to top again
5. Wait for a response to complete - transcript jumps to top again
6. Repeat for every interaction

---

## Changes shipped in v0.51.576

Two PRs landed in this release, either or both may be responsible:

- **#4584** (`ui.js` only) - scroll re-pin now requires deliberate downward intent; added `!_messageUserUnpinned || bottomDistance <= 80` guard; raised wheel unpin threshold to `deltaY < -30`
- **#4583** (`style.css` only) - added `overflow-x:hidden` to `.messages`; added `max-width:100%;overflow-x:hidden;word-break:break-word;min-width:0` to `.messages-inner` on mobile

---

## References

- Introduced by: #4584 (merged in release PR #4668, v0.51.576)
- Related (iOS toolbar fix, same family): #4702 / PR #4838

nesquena-hermes:
## Summary

Fixes #4856 — **the transcript is stuck at the top and completely unscrollable on mobile** (Android app, Android Chrome/Edge, and iOS). Opening a conversation lands on the oldest message, and every send/receive leaves you pinned to the top with no way to scroll down to the latest. Desktop is unaffected.

Reproduced and root-caused empirically via headless Chrome mobile emulation (390×844, `mobile:true`) against current master (v0.51.642). The reporter's bisection — last good **v0.51.575**, first broken **v0.51.576** — points exactly at the change that introduced this.

## Root cause

The mobile `.messages-inner` rule carried `overflow-x:hidden` (added in the #4583/#4584 batch shipped in **v0.51.576**). The failure chain:

1. Per the CSS overflow spec, when one axis is `hidden` and the other is `visible`, the `visible` axis is coerced to **`auto`**. So `overflow-x:hidden` silently turned `overflow-y` into `auto`, making `.messages-inner` a **scroll container**.
2. A scroll container's `min-height:auto` resolves to **`0`** (not `min-content`).
3. `.messages-inner` is a flex item (default `flex-shrink:1`) inside the `.messages` column flexbox. With `min-height` 0 it **collapsed to the scroller's height** (~681px) instead of growing to its content (~8251px).
4. The transcript then overflowed the **inner** (which clipped it) rather than the `.messages` scroller — so the scroller had nothing to scroll, `scrollTop` was pinned at 0, and every open/send/receive stranded the reader at the top.

Desktop escaped the collapse (its inner grew to full content height and scrolled normally); mobile did not. This also explains the reporter's observation that **auto-follow on/off made no difference** — the transcript was fundamentally unscrollable, not mis-pinned.

### Why the earlier #4856 attempts didn't fix it

The prior attempts toggled `overflow-anchor` (`auto`↔`none`) during the DOM-rebuild window — a compositor-repaint theory. But the bug isn't a repaint race; it's a layout collapse that makes the transcript unscrollable in the first place. Those guards are orthogonal and left in place (harmless); this PR fixes the actual layout cause.

## Fix

One property on the mobile `.messages-inner` rule:

```diff
- .messages-inner{padding:12px 10px 20px;max-width:100%;overflow-x:hidden;word-break:break-word;min-width:0;}
+ .messages-inner{padding:12px 10px 20px;max-width:100%;overflow-x:clip;word-break:break-word;min-width:0;}
```

`overflow-x:clip` suppresses horizontal overflow exactly as the original #4583/#4553 work intended, but **without** coercing `overflow-y` or creating a scroll container — so `min-height:auto` stays `min-content` and the inner grows to its full height. (`overflow:clip` is already used elsewhere in `style.css`; supported in all current Android Chrome / iOS Safari.)

## Verification (empirical, against the real served fix)

Headless Chrome mobile emulation, 390×844, `mobile:true`, against a 30-message session:

| Check | Before | After |
|---|---|---|
| `messages` scrollable delta | **0** (stuck) | **7746px** |
| inner offsetHeight | clamped to 681 | grows to 8427 |
| Open session | top | **bottom (latest)** |
| After send/receive re-render | jumps to top | **stays at bottom** |
| Re-render while scrolled mid-transcript | → scrollTop 0 | **position preserved** |
| Horizontal overflow (clip still works) | n/a | clipped, no h-scroll |

Desktop emulation: unchanged (inner already grew to content and scrolled).

## Tests

- Added `tests/test_issue4856_mobile_transcript_unscrollable.py` — asserts the mobile inner uses `overflow-x:clip`, never `overflow-x:hidden` / `overflow-y:auto|scroll`, and that `.messages` remains the real scroll container.
- Updated `tests/test_issue4553_mobile_transcript_overflow.py` — `overflow-x:clip` satisfies #4553's original horizontal-pan-prevention goal; updated the assertion accordingly.

Thanks to the reporter for the precise bisected report (Android app + mobile browser, confirmed broken on v0.51.641).

Closes #4856.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **The transcript is no longer stuck at the top and unscrollable on mobile.** On phones (Android and iOS), opening a conversation — and every send/receive after — left the view pinned to the oldest message with no way to scroll down to the latest. Root cause: the mobile `.messages-inner` rule used `overflow-x:hidden`, which (per the CSS spec) silently coerces `overflow-y` to `auto`, turning the inner element into a scroll container. A scroll container's `min-height:auto` resolves to `0`, so the inner — a flex item inside the `.messages` column flexbox — collapsed to the viewport height instead of growing to its content; the transcript then overflowed the (clipped) inner rather than the real scroller, leaving nothing to scroll and pinning `scrollTop` at 0. Switching to `overflow-x:clip` suppresses horizontal overflow exactly as intended without creating a scroll container, so the inner grows to full height and the transcript scrolls normally. Desktop was unaffected. Thanks @Josh for the detailed report (Android app + mobile browser, confirmed on v0.51.641). (#4856)
+
 ## [v0.51.642] — 2026-06-25 — Release WW (a failed steer now shows why, with Retry / Dismiss)
 
 ### Fixed
--- a/static/style.css
+++ b/static/style.css
@@ -2493,7 +2493,19 @@
     .topbar-chips .chip,.topbar-chips .ws-chip,.topbar-chips button{font-size:11px!important;padding:4px 8px!important;white-space:nowrap;}
     .settings-main{padding:18px 16px;}
     .hermes-action-grid{grid-template-columns:1fr;}
-    .messages-inner{padding:12px 10px 20px;max-width:100%;overflow-x:hidden;word-break:break-word;min-width:0;}
+    /* #4856: use overflow-x:clip, NOT hidden. `overflow-x:hidden` forces
+       overflow-y from visible to auto (CSS spec: the two axes can't be
+       visible+clipped), which turns .messages-inner into a scroll container.
+       A scroll container's `min-height:auto` resolves to 0, so this flex item
+       (flex-shrink:1 inside the .messages column flexbox) collapses to the
+       scroller's height instead of growing to its content — the transcript
+       then overflows the INNER (clipped) rather than the .messages scroller,
+       leaving nothing to scroll and pinning scrollTop at 0. On mobile that
+       made every conversation land at the top and refuse to scroll down.
+       `overflow-x:clip` suppresses horizontal overflow (the original #4583
+       intent) WITHOUT coercing overflow-y or creating a scroll container, so
+       min-height:auto stays min-content and the inner grows normally. */
+    .messages-inner{padding:12px 10px 20px;max-width:100%;overflow-x:clip;word-break:break-word;min-width:0;}
     .msg-body{padding-left:0;max-width:100%;}
     .msg-role{font-size:12px;}
     .composer-wrap{padding:8px 10px 12px!important;}
--- a/tests/test_issue4553_mobile_transcript_overflow.py
+++ b/tests/test_issue4553_mobile_transcript_overflow.py
@@ -40,8 +40,21 @@ def test_messages_inner_mobile_has_containment():
     assert 'max-width:100%' in messages_inner_rule, (
         ".messages-inner in mobile block missing max-width:100%"
     )
-    assert 'overflow-x:hidden' in messages_inner_rule, (
-        ".messages-inner in mobile block missing overflow-x:hidden"
+    # #4856: the mobile .messages-inner must clip horizontal overflow WITHOUT
+    # becoming a scroll container. `overflow-x:hidden` coerces overflow-y to
+    # auto (CSS spec), which made the inner a scroll container whose
+    # min-height:auto resolved to 0 and collapsed it inside the .messages
+    # column flexbox — leaving the transcript unscrollable on mobile (stuck at
+    # top). `overflow-x:clip` clips horizontally without that side effect and
+    # still satisfies the original #4553 horizontal-pan-prevention goal.
+    assert 'overflow-x:clip' in messages_inner_rule, (
+        ".messages-inner in mobile block must use overflow-x:clip (not hidden, "
+        "which coerces overflow-y:auto and makes it an unscrollable scroll "
+        "container — #4856)"
+    )
+    assert 'overflow-x:hidden' not in messages_inner_rule, (
+        ".messages-inner in mobile block must NOT use overflow-x:hidden; that "
+        "regresses #4856 (transcript collapses and won't scroll on mobile)"
     )
     assert 'word-break:break-word' in messages_inner_rule, (
         ".messages-inner in mobile block missing word-break:break-word"
--- a/tests/test_issue4856_mobile_transcript_unscrollable.py
+++ b/tests/test_issue4856_mobile_transcript_unscrollable.py
@@ -1,0 +1,75 @@
+"""Regression test for #4856: mobile transcript collapses and won't scroll.
+
+Root cause (proven empirically via headless Chrome mobile emulation):
+
+  The mobile `.messages-inner` rule carried `overflow-x:hidden` (added in the
+  #4583/#4584 batch, v0.51.576 — the user's bisected "first broken" version).
+  Per the CSS overflow spec, when one axis is `hidden` and the other is
+  `visible`, the `visible` axis is coerced to `auto`. So `overflow-x:hidden`
+  silently turned `overflow-y` into `auto`, making `.messages-inner` a SCROLL
+  CONTAINER.
+
+  A scroll container's `min-height:auto` resolves to `0` (not `min-content`).
+  `.messages-inner` is a flex item (default `flex-shrink:1`) inside the
+  `.messages` column flexbox, so with min-height 0 it collapsed to the
+  scroller's height (~681px) instead of growing to its content (~8251px). The
+  transcript then overflowed the INNER (which clipped it) rather than the
+  `.messages` scroller — so the scroller had nothing to scroll, `scrollTop`
+  was pinned at 0, and every open/send/receive left the reader stuck at the
+  top with no way to scroll down. Desktop escaped the collapse; mobile didn't.
+
+Fix: `overflow-x:clip` clips horizontal overflow (the original #4553/#4583
+intent) WITHOUT coercing `overflow-y` or creating a scroll container, so
+`min-height:auto` stays `min-content` and the inner grows to its full height.
+
+These are source-level guards; the runtime collapse is layout/viewport
+specific and not reproducible in headless CI without a full browser.
+"""
+import re
+from pathlib import Path
+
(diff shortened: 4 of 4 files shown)


nesquena (approved):

Independent review of the #4856 fix: on **mobile** (Android app, Android Chrome/Edge, iOS) the transcript was stuck at the top and completely unscrollable — every open/send/receive stranded the reader at the oldest message. Desktop unaffected. Bisected by the reporter to v0.51.576.

### What this ships

`static/style.css` (one property + a 12-line explanatory comment), `tests/test_issue4856_mobile_transcript_unscrollable.py` (+75, new), `tests/test_issue4553_mobile_transcript_overflow.py` (+15/-2), `CHANGELOG.md` (+4). Agent-authored. Closes #4856.

Fresh tarball: pure WebUI CSS layout. No agent/protocol/config surface. Cross-tool: none. ✓

### Root cause + fix — verified against the CSS spec

The mobile `.messages-inner` rule carried `overflow-x:hidden` (from the #4583/#4584 batch in v0.51.576 — the reporter's "first broken"). The failure chain, which I verified against **CSS Overflow Module Level 3**:

1. The spec coerces a `visible` axis to `auto` **only when the other axis is `hidden`/`scroll`/`auto`**. So `overflow-x:hidden` turned the default `overflow-y:visible` into `auto` → `.messages-inner` became a **scroll container**.
2. A scroll container's `min-height:auto` resolves to **0** (not `min-content`).
3. `.messages-inner` is a `flex-shrink:1` item in the `.messages` column flexbox, so with `min-height:0` it **collapsed** to the scroller's height instead of growing to content.
4. The transcript then overflowed the clipped *inner* rather than the `.messages` scroller → `scrollTop` pinned at 0 → unscrollable.

The fix changes the one property to `overflow-x:clip`. **`clip` is exempt from the coercion rule** (the spec only promotes `visible`→`auto` against `hidden`/`scroll`/`auto`, not `clip`), so `overflow-y` stays `visible`, the inner is not a scroll container, `min-height:auto` stays `min-content`, and it grows to full height — while still suppressing horizontal overflow (the original #4583/#4553 intent). The root-cause analysis and the chosen value are both correct per spec.

### Architecture confirmed

- `.messages` (style.css:1715) is the **intended** scroll container: `flex:1; overflow-y:auto; min-height:0; -webkit-overflow-scrolling:touch; overscroll-behavior-y:contain`. It correctly owns vertical scrolling.
- `.messages-inner` must **grow to content** so `.messages` scrolls it. The bug was the inner accidentally becoming a *second* scroll container; the `clip` fix prevents that. `.messages` itself is unchanged (its `overflow-x:hidden`+explicit `overflow-y:auto` is intentional and triggers no problematic coercion).
- **Desktop is untouched:** the desktop `.messages-inner` (style.css:1729) has no `overflow-x` and is outside the mobile media query; the change is scoped to `@media(max-width:640px)` only. ✓

### Edge-case matrix

| Scenario (mobile) | Behavior |
|---|---|
| Open a long conversation | lands at the latest message (scroller has full content) ✅ |
| Send/receive re-render | stays at bottom (was: jumped to top) ✅ |
| Re-render while scrolled mid-transcript | position preserved (was: scrollTop 0) ✅ |
| Wide content / code block | horizontal overflow clipped — `clip` satisfies #4553's pan-prevention ✅ |
| Desktop | unchanged (different rule, outside the media query) ✅ |
| `.messages` scroller | remains the `overflow-y:auto` scroll container ✅ |

### Tests

- `tests/test_issue4856_mobile_transcript_unscrollable.py` (3) — mobile inner uses `overflow-x:clip` (not `hidden`), never sets `overflow-y:auto|scroll`, and `.messages` keeps `overflow-y:auto`. Good regression locks against re-introducing the collapse.
- `tests/test_issue4553_mobile_transcript_overflow.py` — updated: `overflow-x:clip` satisfies #4553's horizontal-pan goal, asserts `hidden` is gone. Legitimate (clip still clips horizontally).
- **5/5** across both files; mobile/transcript/scroll blast-radius **408 passed, 2 skipped**.
- **Full suite: 10390 passed / 133 skipped / 0 failures** in 207s (deselected the pre-existing darwin CRLF flake).
- Runtime scroll behavior isn't unit-testable in CI; the PR's headless-Chrome mobile emulation (390×844) is the runtime evidence — `messages` scrollable delta 0→7746px, open/send/receive all land at bottom, mid-scroll preserved. The structural tests + the spec reasoning substitute for CI.

- **PR-body inaccuracy:** the description says "`overflow:clip` is already used elsewhere in `style.css`." It isn't — the only `clip` in the file is `text-overflow:clip` (a different property). This is the first use of `overflow-x:clip`. It doesn't affect correctness (`overflow:clip` is well-supported on current Android Chrome / iOS Safari 16+), but the precedent claim is off — flagging only so the browser-support baseline rests on actual support, not a non-existent precedent.
- **Graceful degradation:** on a browser too old to support `overflow:clip` (pre-iOS-16), the declaration is ignored → `overflow-x` falls back to `visible` → the #4553 horizontal-pan issue could reappear, but the transcript still scrolls vertically (no coercion, no collapse). So the worst-case regression on ancient browsers is the lesser (pan) bug, not the unscrollable one — an acceptable trade.

### Recommendation

A precise, minimal, correctly-reasoned CSS fix: `overflow-x:clip` suppresses horizontal overflow without coercing `overflow-y` into a scroll container (verified against CSS Overflow L3 — `clip` is exempt from the `visible→auto` promotion that `hidden` triggers), so the mobile `.messages-inner` grows to content and `.messages` scrolls normally again. Scoped to the mobile media query (desktop untouched), with structural regression tests locking the invariant and the runtime behavior confirmed by the PR's headless-mobile emulation. The only nit is a PR-body precedent claim that doesn't hold; the fix itself is sound. Ship.

nesquena-hermes:
Heads up — this fix has **already landed on `master`** and shipped in a release. The diagnosis and one-line CSS change here are correct, but they're now redundant with what's already merged.

## What's already on master

`origin/master:static/style.css:2509` already carries the exact change this PR proposes:

```css
.messages-inner{padding:12px 10px 20px;max-width:100%;overflow-x:clip;word-break:break-word;min-width:0;}
```

It went in via commit `3acd80258 fix(#4856): mobile transcript collapses unscrollable — overflow-x:hidden coerces overflow-y:auto` (authored Jun 25 04:15 UTC), which carries the same `overflow-x:hidden` → `overflow-x:clip` swap plus the identical inline comment about the CSS axis-coercion chain, and the same two test files (`tests/test_issue4856_mobile_transcript_unscrollable.py` new, `tests/test_issue4553_mobile_transcript_overflow.py` updated).

That commit was cut into **Release XB (v0.51.646)** — and the release title even credits this PR: `Release XB (v0.51.646): mobile transcript scrolls again — overflow-x:clip (#4898, closes #4856)`. `git tag --contains 3acd80258` confirms it's in v0.51.646 and every release since.

## On the root-cause analysis (which is correct)

For the record, the bisect and the spec chain both check out. The `overflow-x:hidden` on the mobile `.messages-inner` rule was introduced in `fc3d6b68e stage batch #4584+#4583 (rodboev): ... mobile horizontal-pan fix (#4668)`, which matches the reporter's "first broken v0.51.576" bisection — that commit's diff is literally `.messages-inner{padding:12px 10px 20px;} → {...;overflow-x:hidden;...}`. And the failure mechanism is exactly as described: per CSS overflow, `overflow-x:hidden` + `overflow-y:visible` coerces the y-axis to `auto`, making `.messages-inner` a scroll container whose `min-height:auto` resolves to `0`, collapsing the flex item so `.messages` had nothing to scroll. `overflow-x:clip` suppresses horizontal overflow without that coercion, which is why it fixes it. `overflow:clip` is already used elsewhere in `style.css`, so support wasn't a concern.

## Recommendation

Nothing wrong with the work — it's the same fix the maintainer landed. This PR can be **closed as already-merged** (superseded by `3acd80258` / v0.51.646). Anyone still seeing the stuck-at-top transcript on mobile should `git pull` to ≥ v0.51.646 and hard-refresh the PWA (the cutter preview on this PR shows the corrected behavior — opens at latest, scrolls normally). Thanks for the precise bisected report; it's what made the merged fix a one-liner.

nesquena-hermes:
Shipped in **v0.51.646** (mobile transcript scrolls again (overflow-x:clip)) — the fix is live on master and deployed to prod. Thanks @nesquena-hermes for the contribution. Closing as shipped (the release landed via a stage/release branch so this PR object didn't auto-close). Please reopen if you still see the issue after upgrading + a hard refresh.

Issue #4859, opened 2026-06-24
WebUI: opening a conversation lands at the top (first message) instead of the latest reply


nesquena-hermes:
WebUI: opening a conversation lands at the top (first message) instead of the latest reply
## Symptom
When opening/loading a conversation, the transcript viewport lands at the top (the first message) instead of at the bottom (the latest reply). The reader has to scroll down manually every time to reach the most recent message.

## Platform
Android app (wrapper); transcript scroll behavior is a shared WebUI surface, so it likely affects mobile/desktop browsers too.

## Suspected area
WebUI transcript scroll-to-bottom-on-load. Lives in the same area as prior scroll fixes: closed #4779/#4780 (force scroll-to-bottom on session switch), closed #4720 (jump-to-top after completion), and the in-flight #MOBILESCROLL PRs #4818/#4820 (scrollTop jank to top during mobile streaming). Likely overlaps OPEN #4856 (transcript jumps to top on every interaction, regression in v0.51.576). Could be a regression of #4779/#4780, a distinct load-path case, or already covered by #4856 / open mobile-scroll work.

## Repro
1. Open the app with at least one existing conversation.
2. Tap/open a conversation that has more than one screen of messages.
3. Observe the view is positioned at the first message rather than the latest reply.
4. Scroll down to reach the most recent message.

---
_Filed automatically from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena-hermes:
Triage: this is in the transcript scroll-on-load surface, which has had several fixes land very recently — some likely after the build this was reported against:

- **v0.51.636** (#4878) — Android Chromium jumping the transcript to the top on every interaction (overflow-anchor re-anchoring).
- **v0.51.646** (#4856) — the mobile transcript was *fundamentally unscrollable* (an `overflow-x:hidden` flex-collapse); the fix restored normal scrolling and open-at-bottom behavior. A class-level regression guard (`tests/test_transcript_scroll_invariant.py`) now protects it.

The symptom here ("lands at the first message instead of the latest reply" on the Android wrapper) overlaps that surface. **Could you confirm whether this still reproduces on v0.51.652 (current) or newer?** The two fixes above postdate the closed #4779/#4780/#4720 this references, so it's plausible they already resolve it.

[...]

PR #4902, opened 2026-06-25
fix(scroll): keep session switches and background streams anchored

starship-s:
## Thinking Path

- Opening a conversation should land on the latest reply, not restore stale scroll state from a previously viewed transcript.
- Mobile layouts can make the transcript's effective scroll owner differ from the desktop `#messages` container, so capture/restore/follow logic needs one consistent owner contract.
- Background streams should not be able to mutate scroll position in a different session the user is currently reading.
- Same-session refresh/reconcile should preserve the reader's viewport whenever the transcript is physically away from bottom, even if sticky unpin state missed the scroll gesture.
- The narrow fix is to split session-switch behavior from same-session refresh behavior: real switches force latest/bottom, while same-session updates preserve reader position.

## What Changed

- Added a fresh-session-load path so real session switches reset stale scroll-follow state and force the transcript to latest/bottom after render.
- Canonicalized transcript scroll helpers around the effective scroll owner, covering mobile cases where `#msgInner` is the active scroller.
- Guarded live-stream scroll callbacks so a background stream cannot scroll the currently viewed different session.
- Made reader-away detection geometry-aware so external refresh defers when the visible transcript is physically away from bottom.
- Hardened long-message render cache keys to avoid stale rendered bodies when templated messages share length/prefix/suffix.
- Added focused regression coverage for session-switch bottoming, scroll-owner behavior, stream callback isolation, and same-session refresh preservation.

## Why It Matters

- Fixes conversation open/switch flows landing at the first message instead of the latest reply.
- Keeps mobile transcript scroll behavior aligned with the actual scroller instead of desktop-only assumptions.
- Prevents active background streams from yanking the visible transcript while the user is reading another session.
- Preserves same-session reading position during refresh/reconcile paths without weakening normal auto-follow behavior.

Refs #4859

## Verification

Targeted and formerly failing coverage:

`./scripts/test.sh tests/test_issue_session_switch_jump_to_bottom.py tests/test_tars_scroll_reset_regressions.py tests/test_issue4006_auto_scroll_follow_default.py tests/test_issue1731_upward_scroll_unpins.py tests/test_issue4856_android_scroll_regression.py tests/test_issue3470_touch_unpin_streaming_scroll.py tests/test_issue3479_ios_stream_scroll_jump.py tests/test_issue4295_midstream_scroll_anchor.py tests/test_issue4295_scroll_pin_reentry.py tests/test_streaming_race_fix.py tests/test_webui_external_refresh_frontend.py tests/test_renderer_js_behaviour.py tests/test_issue2543_named_context_session_switch.py tests/test_issue3306_loadsession_carry_forward.py tests/test_regressions.py tests/test_issue_new_chat_draft_restore.py tests/test_parallel_session_switch.py -q`

Result: `258 passed, 1 skipped`

Broader JS-referencing surface:

`./scripts/test.sh $(grep -rlE --include='*.py' "static/(ui|sessions|messages)\\.js" tests/ | sort) tests/test_issue_new_chat_draft_restore.py tests/test_parallel_session_switch.py -q`

Result: `2039 passed, 15 skipped, 6 failed`

The 6 failures are all in `tests/test_update_banner_fixes.py`. That file passes in isolation:

`./scripts/test.sh tests/test_update_banner_fixes.py -q`

Result: `85 passed`

Syntax / hygiene:

`node --check static/ui.js`

`node --check static/sessions.js`

`node --check static/messages.js`

`git diff --check origin/master...HEAD`

Result: all passed.

## Risks / Follow-ups

- This is opened as a draft for maintainer review because the branch touches several scroll-state surfaces.
- The broad JS-referencing test command exposes pre-existing order-pollution in update-banner tests when run in one shard; those tests pass in isolation and this branch does not touch update mechanics.
- Real-device mobile browser smoke is still useful before marking ready, especially for Android wrapper/PWA scroll behavior.

## Model Used

- OpenAI / GPT-5.5 for implementation orchestration and verification.
- OpenAI / GPT-5.3 Codex Spark via OpenCode for focused implementation edits.
- Anthropic / Claude Opus 4.8 (`xhigh`) for independent pre-PR review and sign-off.

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -1776,6 +1776,13 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
   function _isActiveSession(){
     return !!(S.session&&S.session.session_id===activeSid);
   }
+  function _isLivePaneCurrent(){
+    return _isActiveSession() && S.activeStreamId===streamId;
+  }
+  function _scrollIfLivePanePinned(){
+    if(!_isLivePaneCurrent()) return;
+    scrollIfPinned();
+  }
   function _ownsActiveStreamOrBackground(){
     return !_isActiveSession() || S.activeStreamId===streamId;
   }
@@ -3593,7 +3600,7 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       if(!assistantBody){onDone();return;}
       const target=_streamFadeCurrentDisplayText();
       const caughtUp=_renderStreamingFadeMarkdown(target);
-      scrollIfPinned();
+      _scrollIfLivePanePinned();
       if(caughtUp){
         // parser_end can flush pending markdown text; include that final text in
         // the fade wait instead of replacing it immediately in renderMessages().
@@ -3960,6 +3967,8 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       _renderPending=false;
       // Guard: a pending setTimeout+rAF can outlive stream finalization.
       if(_streamFinalized) return;
+      // Guard against stale callbacks after switching tabs/sessions.
+      if(!_isLivePaneCurrent()) return;
       // Mobile scroll-jank guard: temporarily disable overflow-anchor before DOM
       // writes to suppress Chromium scroll re-anchoring during streaming growth.
       if(typeof window._fixMobileScrollJank==='function') window._fixMobileScrollJank();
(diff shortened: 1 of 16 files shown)


starship-s:
Closing this out since the issue appears to have been addressed upstream / no longer needs this branch.

Issue #4811, opened 2026-06-24
bug(transparent-stream): returning to a running session reorders prior content — text grouped at top, tool calls grouped after, chronology lost


nesquena-hermes:
## Symptom
When a user leaves a session while a turn is still running and later returns to it, the previously-rendered content is reconstructed out of order. All assistant text is grouped into one block at the top of the transcript, and all tool calls / activity rows are grouped into a second block after it — instead of being interleaved in the chronological order they originally arrived. Content that streams in *after* the user returns renders correctly (in order); only the already-produced ("older") content is collapsed into these two separate chunks.

## Suspected area
Session-return / SSE-reconnect transcript reconstruction in the Transparent Stream render path (the worklog/activity rebuild that runs when a still-running session is re-opened). This looks like a regression of the earlier "keep live prose chronological, don't fold into the top Worklog rail" behavior, but on the replay/return path rather than the live-streaming path. Falls within the recent transparent-stream + DOM-recycle churn wave.

## Repro
1. Start a turn that produces both assistant text and tool calls.
2. While the turn is still running, navigate away from the session.
3. Return to the still-running session.

Observed: prior assistant text appears as a single block at the top, prior tool calls appear as a separate block after it (original chronology lost). Newly streamed content after the return appears correctly ordered.

Environment: WebUI v0.51.614, Agent 0.17.0.

---
_Filed automatically from a Discord community report. Edit or close if mis-scoped._

PR #4904, opened 2026-06-25
fix(#4622): restore tool-card result body on the settled rebuild path

### LINKED ISSUE #4622: bug(transparent-stream/tool-cards): terminal tool output truncated + not expandable on click; patch/edit cards missing diff (suspected regression)

nesquena-hermes:
## What this fixes

In the **Transparent Stream / Worklog** view, after a turn settles (or on reload/reconnect):
- **terminal** tool cards show the `$ command` line with **no stdout** below it, and
- **patch / edit_file** cards show the raw input fields with **no rendered diff**.

The full output is visible *while the stream is in progress*, then disappears once it settles — the tell that this is the **settled rehydrate** path, not the live card. Recurrence of #4622, confirmed on v0.51.636 (the earlier #4625 guard only covered the live `buildToolCard` path, which is why it stayed green).

## Root cause (proven empirically)

The settled rebuild `_anchorSceneRowsByMessageIndex` (`static/messages.js`) reconstructs each tool row from the persisted `messages[].tool_calls` (state.db / sidecar). The backend persists the result body as a capped snippet and truncates args, and on a cold/paginated load the row can carry **no** body at all. The full output lives on the live `S.toolCalls` entry at settle time — but the dedup loop matched the live entry by tool id and **skipped it** (`continue`) rather than restoring the missing body onto the surviving settled row.

So `_anchorSceneToolCallFromRow` → `buildToolCard` received `tc.snippet=''`: empty output region, and the diff (detected from the snippet via `_snippetLooksLikeDiff`) never rendered. The transparent card's **Output** tab (result body) and **Full** tab (detailed input args) both went empty too.

I built a node harness that drives the real settled chain (`_anchorSceneRowsByMessageIndex` → `_anchorSceneToolRowFromCall` → `_anchorSceneToolCallFromRow`) and reproduced the exact failure on `origin/master`, then confirmed the fix:

| Card | view | before (master) | after |
|---|---|---|---|
| terminal | Output body / Show-more `data-full` | ❌ empty | ✅ full stdout |
| terminal | transparent **Output** tab | ❌ empty | ✅ non-empty |
| edit_file | diff body rendered | ❌ no | ✅ yes |
| edit_file | **Full** tab input args | (kept) | ✅ kept |

## The fix

New `_enrichSettledToolRowBodyFromLive(row, live)` restores `snippet`/`result`/`output`, `command`, and `args` onto the settled row's `tool`+`payload` from the matched live call — **only when the settled value is empty** (never clobbers a genuinely persisted body). Wired into the dedup loop: on a tid match that would previously have been dropped, it enriches the existing pooled row instead of discarding the live data. No duplicate row is added.

## Tests (the guard that would have caught this)

New `tests/test_issue4622_settled_toolcard_body.py` drives the **settled** chain end-to-end (not the live path) and asserts:
- terminal stdout restored + full output carried for the **Show more** expander + transparent **Output** tab non-empty;
- patch/edit diff body restored + renders as a diff + **Full** tab keeps the input args;
- a genuine persisted body is **not** clobbered by a differing live value;
- dropped args are restored from live;
- no-live-match leaves the settled row untouched;
- exactly one row per tid after enrichment (no duplicate).

Verified to **fail on clean master** (behavioural assertions) and pass with the fix. Also registered the new helper in the #4811 reconnect-chronology test driver (it eval-extracts the `_anchorScene*` family and now references the new symbol). Full local suite green; gated with Codex + Opus.

A reusable before/after proof harness is saved in the workspace (`qa/harnesses/prove_4622_settled_toolcard.js`).

Closes #4622.

---

*Self-built maintainer PR (nesquena-hermes). Requesting an independent human review before merge per the self-built-PR policy. @rodboev / @franksong2702 — a hands-on check would be ideal: run a turn with a long terminal command + a patch/edit, let it settle (or reload the session), and confirm the terminal stdout and the diff both still show, including the **Output** / **Full** tabs and **Show more**.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Terminal tool cards keep their full output, and patch/edit cards keep their diff, after a turn settles in the Transparent Stream / Worklog view.** While a turn streamed, the tool cards showed their complete output; once the stream settled (or on reload/reconnect), terminal cards collapsed to just the `$ command` line with no stdout and patch/edit cards showed their input fields with no rendered diff. The settled rebuild reconstructs each tool row from the persisted `messages[].tool_calls` (state.db / sidecar), which can carry only a short preview — or, on a cold/paginated load, nothing — for the result body; the live in-memory tool call still held the full output at settle time, but the merge dropped it (it skipped the matching live entry instead of restoring the missing body onto the surviving settled row). The merge now restores the full result body, command, and input args from the matched live tool call when the settled row is missing them (without clobbering a genuinely persisted value), so the rebuilt card shows the complete terminal stdout (with the **Show more** expander and the transparent **Output** / **Full** tabs), and renders the patch/edit diff. (#4622)
+
 ## [v0.51.646] — 2026-06-25 — Release XB (mobile transcript scrolls again)
 
 ### Fixed
--- a/static/messages.js
+++ b/static/messages.js
@@ -2560,16 +2560,14 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
     const payload=(row.payload&&typeof row.payload==='object')?row.payload:(row.payload={});
     let enriched=false;
     const _empty=v=>v===undefined||v===null||v==='';
-    // Body/result fields that buildToolCard reads to render output + detect diffs.
+    // Result body: _anchorSceneToolCallFromRow renders tool.snippet||payload.snippet
+    // (||payload.result||payload.output) as the card output + diff source, so
+    // restore the snippet onto both tool+payload when the settled row has none.
     const liveSnippet=_anchorSceneStringPayload(live.snippet||live.result||live.output);
-    if(liveSnippet&&_empty(tool.snippet)&&_empty(payload.snippet)&&_empty(payload.result)&&_empty(payload.output)){
+    if(liveSnippet&&_empty(tool.snippet)&&_empty(payload.snippet)){
       tool.snippet=liveSnippet; payload.snippet=liveSnippet; enriched=true;
     }
-    const liveResult=_anchorSceneSafePayload(live.result);
-    if(liveResult!==undefined&&liveResult!==null&&_empty(tool.result)){ tool.result=liveResult; }
-    const liveOutput=_anchorSceneSafePayload(live.output);
-    if(liveOutput!==undefined&&liveOutput!==null&&_empty(tool.output)){ tool.output=liveOutput; }
-    // Command (shell detail-lead) + args (diff reconstruction inputs).
+    // Command (shell detail-lead) + args (diff/input reconstruction, the "Full" tab).
     const liveCommand=_anchorSceneStringPayload(live.command||live.raw_command);
     if(liveCommand&&_empty(tool.command)&&_empty(payload.command)){
       tool.command=liveCommand; payload.command=liveCommand; enriched=true;
--- a/static/messages.js
+++ b/static/messages.js
@@ -2545,6 +2545,42 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
     if(!message||typeof message!=='object') return '';
     return String(message.reasoning||message._reasoning||message.reasoning_content||message.thinking||'');
   }
+  // #4622: a settled tool row built from messages[].tool_calls (state.db/sidecar)
+  // can lack the result body — terminal stdout, or the diff/output that a
+  // patch/edit card renders — because the persisted row carries only a short
+  // preview (or, on a cold/paginated load, nothing). The full body lives on the
+  // live S.toolCalls entry at settle time. When a settled row and a live call
+  // match by tool id, restore the missing body fields from the live call onto
+  // the settled row's tool+payload (only when the settled value is empty — never
+  // clobber a genuine persisted body), so the rebuilt card shows full output +
+  // the Show-more expander + the rendered diff. Returns true if it enriched.
+  function _enrichSettledToolRowBodyFromLive(row, live){
+    if(!row||typeof row!=='object'||!live||typeof live!=='object') return false;
+    const tool=(row.tool&&typeof row.tool==='object')?row.tool:(row.tool={});
+    const payload=(row.payload&&typeof row.payload==='object')?row.payload:(row.payload={});
+    let enriched=false;
+    const _empty=v=>v===undefined||v===null||v==='';
+    // Body/result fields that buildToolCard reads to render output + detect diffs.
+    const liveSnippet=_anchorSceneStringPayload(live.snippet||live.result||live.output);
+    if(liveSnippet&&_empty(tool.snippet)&&_empty(payload.snippet)&&_empty(payload.result)&&_empty(payload.output)){
+      tool.snippet=liveSnippet; payload.snippet=liveSnippet; enriched=true;
+    }
+    const liveResult=_anchorSceneSafePayload(live.result);
+    if(liveResult!==undefined&&liveResult!==null&&_empty(tool.result)){ tool.result=liveResult; }
+    const liveOutput=_anchorSceneSafePayload(live.output);
+    if(liveOutput!==undefined&&liveOutput!==null&&_empty(tool.output)){ tool.output=liveOutput; }
+    // Command (shell detail-lead) + args (diff reconstruction inputs).
+    const liveCommand=_anchorSceneStringPayload(live.command||live.raw_command);
+    if(liveCommand&&_empty(tool.command)&&_empty(payload.command)){
+      tool.command=liveCommand; payload.command=liveCommand; enriched=true;
+    }
+    const liveArgs=(live.args&&typeof live.args==='object')?live.args:null;
+    const argsEmpty=o=>!o||typeof o!=='object'||Object.keys(o).length===0;
+    if(liveArgs&&argsEmpty(tool.args)&&argsEmpty(payload.args)){
+      tool.args={...liveArgs}; payload.args={...liveArgs}; enriched=true;
+    }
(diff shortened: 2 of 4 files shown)


nesquena (approved):

Independent review of the #4622 recurrence: in **Transparent Stream / Worklog** view, after a turn **settles** (or on reload/reconnect), terminal cards lost their stdout and patch/edit cards lost their diff — visible while streaming, then gone once settled. The earlier #4625 guard only covered the live `buildToolCard` path, so the settled rebuild regressed silently.

### What this ships

`static/messages.js` (+49/-4), `tests/test_issue4622_settled_toolcard_body.py` (+242, new), `tests/test_issue4811_reconnect_chronology.py` (+1), `CHANGELOG.md` (+4). Agent-authored; Codex + Opus gated. Self-built — requesting independent human review; my `event=APPROVE` is the code-review gate. Closes #4622.

Fresh tarball: pure WebUI settled-rebuild presentation. The result body comes from the live `S.toolCalls` (already in the WebUI) and the persisted `messages[].tool_calls` (state.db/sidecar the agent writes, read-only here). No agent protocol/schema change. ✓

### Root cause + fix — traced

The settled rebuild `_anchorSceneRowsByMessageIndex` builds each tool row from persisted `messages[].tool_calls`, which carries only a capped preview (or nothing on a cold/paginated load) for the result body. The full body lives on the live `S.toolCalls` entry at settle time, but the dedup loop matched the live entry by tool id and `continue`d (dropped it) instead of restoring the missing body. So `buildToolCard` got `snippet=''` → empty Output region, no diff.

The fix adds `_enrichSettledToolRowBodyFromLive(row, live)` and wires it into the dedup loop: on a tid match that would have been dropped, it enriches the existing pooled row from the live call instead of discarding it. I verified the four correctness-critical properties:

1. **No-clobber (fill only when empty):** each field is restored only when the settled value is empty (`undefined`/`null`/`''`; the row's `result`/`output` default to `null` so an absent persisted body counts as empty and gets filled, while a genuine persisted value is never overwritten). The snippet fill requires *all* of `tool.snippet`/`payload.snippet`/`payload.result`/`payload.output` to be empty first — conservative.
2. **Enrichment reaches the rendered row.** The pool holds `{...row}` (shallow copy) while `rowByToolId` holds the original `row`. This works **because** `_anchorSceneToolRowFromCall` always builds `row.tool` and `row.payload` as object literals, so the shallow spread shares those nested references — mutating `row.tool`/`row.payload` in place is seen by the pooled copy. I confirmed both are always objects.
3. **No duplicate row.** The `continue` is preserved — a tid-matched live entry is still never pushed; the only change is it enriches first. Exactly one row per tid.
4. **No aliasing.** `args` is restored as `{...liveArgs}` (a copy), so the row and the live entry don't share the args object.

### Security audit

No new sink. The enrichment places **data** (snippet/result/output/command/args) onto the row; rendering still goes through the existing `_anchorSceneToolCallFromRow` → `buildToolCard` path (the same escaped renderer the live and persisted-snippet paths already use — `textContent`/`esc` for fields, `_transparentToolDetailHtml` for detail). The restored data has identical provenance (tool output) to what the card would render anyway. No raw `innerHTML` added. ✓

### Behavioural verification — RED→GREEN proven

The PR's new test drives the **real** settled chain (`_anchorSceneRowsByMessageIndex` → `_anchorSceneToolRowFromCall` → `_anchorSceneToolCallFromRow` → `buildToolCard`) and asserts on the **rendered card** (`rendersOutputBody`, `snippet == full stdout`, `transparentOutputTabNonEmpty`, diff rendered). To prove it catches the actual behaviour (not just the new symbol), I ran master's continue-only dedup loop with a **no-op stub** of the helper:
```
master loop + no-op enrich → "patch card body must render after settle: assert False"  (terminal + patch both FAIL)
PR fix                      → all assertions pass
```
So the settled cards genuinely render empty bodies without the enrichment, and the fix restores them. (Note: against unmodified master the test *errors* on the missing symbol — the PR's "behavioural FAIL on master" is really a symbol-coupling error; the stub run above is the clean behavioural RED.)

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Settled terminal card, persisted body empty | full stdout restored from live; Show-more `data-full` + transparent Output tab populated ✅ |
| Settled patch/edit card, persisted body empty | diff body restored, renders as a diff; Full tab keeps input args ✅ |
| Persisted body genuinely present | NOT clobbered by a differing live value ✅ |
| Dropped args on persisted row | restored (copied) from live ✅ |
| No live match | settled row left untouched ✅ |
| tid match | exactly one row, enriched (no duplicate) ✅ |
| Live card (in-progress) | unchanged (#4625 path untouched) ✅ |
| CLI reads same session | webui-only settled-render; agent unaffected ✅ |

### Tests

- `tests/test_issue4622_settled_toolcard_body.py` (6) + `test_issue4811_reconnect_chronology.py` (driver registers the new helper) — **14/14**.
- Transparent / anchor / tool-card blast-radius — **376 passed, 3 skipped**.
- **Behavioural RED→GREEN** proven via the no-op-stub run (above).
- **Full suite: 10419 passed / 133 skipped / 0 failures** in 208s (deselected the pre-existing darwin CRLF flake).
- `node --check static/messages.js` clean.
- **CI on the PR head:** lint + browser-smoke green; test shards finishing — local full suite substitutes.

- The `enriched` return value is only flipped for snippet/command/args (not the `result`/`output` fields); the caller ignores the return, so it's purely informational — harmless.
- The PR's "FAIL on clean master" wording is imprecise (it errors on the missing symbol, not a behavioural assertion); the behaviour is real, as the stub run confirms. A doc nit, not a code issue.

### Recommendation

✅ **Approved clean (code review).** Parked at approval — ready for the release agent's merge/tag pipeline once a hands-on settled/reload check (long terminal command + a patch/edit, confirm stdout + diff + Output/Full tabs survive settle) lands.

A correctly-scoped fix for the settled-rebuild path the #4625 guard missed: when a live tool call matches a settled row that's missing its body, restore the body from live instead of dropping it — strictly fill-when-empty (no clobber), reaching the pooled row via the shared `tool`/`payload` object references, with no duplicate row, no aliasing, and no new render sink. I proved the behaviour with a stubbed-master RED→GREEN and confirmed the always-object invariant the shared-ref fix depends on. Ship (pending the hands-on settle/reload confirmation).

nesquena-hermes:
## Gate summary — CI green, dual-gate SAFE

Ran the full self-built gate (Codex + Opus + full suite + CI):

- **Codex: SAFE TO SHIP.** Verified all five review points and ran its own node harness confirming the in-place enrichment does **not** mutate `S.toolCalls` or the source `message.tool_calls` (the leak concern), plus 14 tests.
- **Opus: Ship it.** Verified empirically that (1) the only-when-empty guard preserves a genuinely persisted body, (2) the shallow-spread pooled row shares `tool`/`payload` refs so the in-place enrichment reaches the pooled copy and only that freshly-built row, (3) no duplicate row, (4) no leak into the inputs, (5) the restored snippet/command/args flow through `_anchorSceneToolCallFromRow` → `buildToolCard` so terminal stdout + the patch diff render. It flagged two dead-code redundancies (the snippet guard checked `payload.result`/`payload.output` which the row builder never sets; the `tool.result`/`tool.output` writes the consumer never reads) — both **trimmed** in the follow-up commit.
- **Full suite:** 10547 passed, 0 failures. **CI: 11/11 green** on the final head (`f6ccf8a0c`), all three Python versions + browser-smoke + lint.

Empirically proven before/after with a node harness driving the real settled chain (`_anchorSceneRowsByMessageIndex` → `_anchorSceneToolRowFromCall` → `_anchorSceneToolCallFromRow`): on `origin/master` the settled terminal card had an empty body and the patch card rendered no diff; with the fix both restore fully (terminal stdout in the body + Show-more `data-full` + transparent **Output** tab; patch diff renders + **Full** tab keeps the input args). The new test fails on clean master and passes with the fix — the guard the prior #4625 (live-path-only) check was missing.

Holding for an independent human review before merge (self-built).

nesquena-hermes:
Shipped in **v0.51.648** (Release XD) — closes #4622. 🎉

Full gate: Codex re-gate SAFE + Opus SHIP IT + suite 10554 passed. The dual gate earned its keep — Codex caught that the first cut only restored a fully-EMPTY settled snippet, missing the **4000-char capped-preview** case (the real #4622 symptom for long terminal output). Fixed with bounded-preview detection (live longer + settled is a ≥4000-char prefix → restore the full live body; genuinely short prefixes are NOT clobbered), idempotent across rebuilds, no source-object mutation. Added capped-preview + not-clobbered regression tests. Verified `_isBoundedPreview` in the served bundle on prod.

PR #4908, opened 2026-06-25
perf(#4842): freeze the CLI/cron sidebar cache key during streaming (stop the per-poll re-query that pins CPU)

nesquena-hermes:
## Summary

A structural performance fix for #4842 (continuing the #4672 / #4808 / #4889 line): **`/api/sessions` re-runs the expensive CLI/cron session projection on every poll while a turn is streaming**, which is the dominant cost in the multi-second sidebar latency + 100% CPU that cron-heavy installs keep reporting.

This is a **PR candidate** — opening for review, not auto-merge. It addresses a layer the prior fixes did not reach (see "Why the earlier fixes didn't fully land" below).

## Root cause

The CLI/cron sidebar projection is cached in `_CLI_SESSIONS_CACHE`. Its key is built in `_resolve_cli_sessions_context` → `_sqlite_file_stat_cache_key` → `_sqlite_content_fingerprint`, which folds in `MAX(rowid) FROM messages`. During an active chat turn the gateway/CLI writes a **message row per streamed delta**, so that fingerprint advances on essentially every `/api/sessions` poll. The frontend polls every ~5s while streaming → every poll misses the cache → the full candidate-join + projection (and the `visible_lineage_metadata` pass) re-runs, contending for the same SQLite / global lock the streaming worker holds.

A reporter's profiler trace on v0.51.645 shows this directly: `get_cli_sessions` at ~5000ms and `visible_lineage_metadata` at ~1200ms per request, with the main thread blocked in `read_importable_agent_session_rows → _project_agent_session_rows → cur.fetchall()`.

### Why the earlier fixes didn't fully land

- **#4808** added a streaming hold-down, but only to the **route-level** session-list cache (`_session_list_cache_streaming_freeze_marker`). It never reached this **inner** CLI-sessions cache, so the heavy CLI/cron query still re-ran whenever the outer cache validated.
- **#4889** memoized the per-row **sidecar file reads** under the projection — real but a different (smaller) layer. The trace's hot frame is the SQL aggregation + lineage pass, not sidecar I/O.

## Fix

Propagate the same streaming-freeze idea to the inner cache:

1. New `_cli_sessions_streaming_freeze_marker()` in `api/models.py` — keyed only on the **set of active stream ids** (mirrors the route-level #4808 marker).
2. While any stream is active, `_resolve_cli_sessions_context` (and the all-profiles path) fold that **stable marker** into the cache key instead of the volatile content fingerprint — so per-token message writes no longer bust the cache.
3. `_cli_sessions_cache_ttl_seconds()` widens the TTL (5s → 30s) while streaming, so the fixed poll cadence can't force a rebuild on every poll.
4. `_on_session_list_changed` (the existing structural-mutation listener) now also clears the CLI cache. Structural signals — cron completion, new/renamed/archived sessions, attention — fire this listener; **per streamed token never does**. That is what makes the freeze safe: real changes still surface promptly, but the streaming write storm doesn't.

Net effect: while a turn streams, the heavy CLI/cron projection is **reused across polls and rebuilt at most once per streaming window** instead of once per poll. The instant a stream starts/stops, the marker changes and the just-finished turn's rows are picked up. **Idle behavior is byte-identical** (the freeze only engages when there are active streams).

## Tests

`tests/test_issue4842_cli_sessions_streaming_freeze.py` (6 tests):
- marker is `None` when idle, stable for the same stream set (order-independent), changes when the set changes;
- **the core guarantee**: the CLI cache key is stable across simulated streamed-message writes while streaming, and still advances with the fingerprint when idle (so newly-committed sessions aren't served stale);
- streaming TTL > idle TTL;
- the structural-mutation listener clears the CLI cache.

Regression sweep (all green): `test_issue4842_cron_projection_perf`, `test_issue4385_cron_archive_reappears`, `test_cli_sessions_cache_fingerprint`, `test_issue3930_source_filter_pushdown`, `test_claude_code_session_import` (incl. `test_get_cli_sessions_cache_invalidates_when_sqlite_wal_changes`), `test_session_sidebar_cache`, `test_session_events`, `test_session_events_http_integration`.

## Honest scope note

I could not reproduce the full 5s locally — my dev state.db's cron sessions carry ~0 messages each, so the join is cheap here; the reporter's clearly carry far more. This fix is targeted at the **mechanism the trace points at** (per-poll re-query under streaming write contention), which is sound and net-positive for every large/cron-heavy install regardless of that one reporter. It is **not** claimed as a verified end-to-end cure for that specific user until it can be confirmed against their data shape.

Refs #4842.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`/api/sessions` no longer re-runs the expensive CLI/cron session projection on every poll while a turn is streaming**, a major cause of the multi-second sidebar latency and 100% CPU on cron-heavy installs (#4842, continuing #4672/#4808/#4889). The CLI/cron sidebar projection is cached, but its cache key folded in a state.db content fingerprint (`MAX(rowid) FROM messages`) that advances on every streamed message row — so during a live turn the frontend's ~5s poll always missed the cache and re-ran the full candidate-join + projection (and the lineage-metadata pass), contending for the same SQLite/global lock the streaming worker holds. The route-level session-list cache already froze its key during streaming (#4808), but that freeze never reached this inner CLI-sessions cache. Now, while any turn is streaming, the CLI-sessions cache key folds in the same stable streaming-freeze marker (keyed only on the set of active stream ids) and its TTL widens, so the heavy projection is reused across polls and rebuilt at most once per streaming window instead of once per poll. Structural sidebar mutations (cron completion, new/renamed/archived sessions, attention) clear the cache directly, so nothing user-visible lags under the freeze; idle behavior is unchanged.
+
 ## [v0.51.647] — 2026-06-25 — Release XC (task detail action buttons reappear on mobile PWA)
 
 ### Fixed
--- a/api/models.py
+++ b/api/models.py
@@ -38,6 +38,12 @@ CLI_VISIBLE_SESSION_LIMIT = 20
 # sidebar window (#3172).
 CRON_PROJECT_CHIP_LIMIT = 200
 _CLI_SESSIONS_CACHE_TTL_SECONDS = 5.0
+# While a turn is actively streaming, hold the CLI/cron projection longer than
+# one poll interval (mirrors the route-level #4808 hold-down). The frontend
+# polls /api/sessions every ~5s during a stream; without a wider window the
+# CLI cache key advances on every streamed message row (see below) and the
+# expensive state.db CLI/cron projection is re-run on every poll. (#4842)
+_CLI_SESSIONS_CACHE_STREAMING_TTL_SECONDS = 30.0
 _CLI_SESSIONS_CACHE_LOCK = threading.Lock()
 _CLI_SESSIONS_CACHE = {}
 
@@ -4099,6 +4105,17 @@ def _copy_cli_sessions(sessions: list) -> list:
 
 
 def _cli_sessions_cache_ttl_seconds() -> float:
+    # #4842: widen the freshness window while a turn is streaming so the fixed
+    # ~5s streaming poll cadence doesn't force a rebuild on every poll. Paired
+    # with the streaming-freeze cache key (so the key is stable across polls
+    # mid-stream), this bounds the heavy CLI/cron projection to one rebuild per
+    # streaming-TTL window instead of one per poll. Mirrors the route-level
+    # #4808 TTL widening.
+    try:
+        if _cli_sessions_streaming_freeze_marker() is not None:
+            return max(0.0, float(_CLI_SESSIONS_CACHE_STREAMING_TTL_SECONDS))
+    except (TypeError, ValueError):
+        pass
     try:
         return max(0.0, float(_CLI_SESSIONS_CACHE_TTL_SECONDS))
     except (TypeError, ValueError):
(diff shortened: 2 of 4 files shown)


rodboev:
I profiled this locally against the PR head (`2d1e1dac`) on my logged-in WebUI at `http://127.0.0.1:8787`, using Chrome DevTools and real profile switches between `default` and `test-ep-sprint31`.

On the profile-switch path, the `#4842` mechanism looks fixed here. The switch request itself is quick, `/api/profile/switch` was about 199 ms into `test-ep-sprint31` and about 334 ms back to `default`. The sidebar session reloads were also quick, `/api/sessions?sidebar_source=webui&exclude_hidden=1` was about 67-152 ms in the two passes I recorded.

The long tail has moved elsewhere. The slowest post-switch request in my traces was `/api/models`, about 4.3 s switching into `test-ep-sprint31` and about 1.3 s switching back to `default`. `/api/projects`, `/api/workspaces`, and `/api/settings` were all effectively noise by comparison.

So on this path, I am no longer seeing the repeated CLI/cron sidebar rebuild as the bottleneck. `#4908` appears to have done its job for the profile-switch case, and the remaining latency I can reproduce is dominated by model-list reload rather than `/api/sessions` churn.

rodboev:
Follow-up from local profiling on the `#4908` head (`2d1e1dac`) against my logged-in WebUI at `http://127.0.0.1:8787`:

I can no longer reproduce `#4842` on the profile-switch path.

Switching `default -> test-ep-sprint31` put `/api/profile/switch` at about 199 ms, with the sidebar reloads on `/api/sessions?sidebar_source=webui&exclude_hidden=1` at about 118 ms and 67 ms. Switching back `test-ep-sprint31 -> default` put `/api/profile/switch` at about 334 ms, with the sidebar reload at about 152 ms.

The remaining delay in these traces is elsewhere. The slowest post-switch request was `/api/models`, about 4.3 s switching into `test-ep-sprint31` and about 1.3 s switching back to `default`. `/api/projects`, `/api/workspaces`, and `/api/settings` were negligible by comparison.

So on this path, the old sidebar/session rebuild bottleneck is no longer showing up locally. If nobody still has a repro on the intended streaming/sidebar path, this PR looks safe to close.

nesquena-hermes:
Read the full diff at `api/models.py` and `api/routes.py` on this branch against `origin/master`, plus the route-level prior art it mirrors. The mechanism is sound and the layering argument in the PR body checks out.

## What this actually changes

The inner CLI/cron cache key is built in `_resolve_cli_sessions_context` and previously always folded in the volatile `_sqlite_file_stat_cache_key(db_path)` (→ `_sqlite_content_fingerprint` → `MAX(rowid) FROM messages`). The new code swaps that one component for a stable marker while streaming:

```python
_streaming_marker = _cli_sessions_streaming_freeze_marker()
db_state_key = _streaming_marker if _streaming_marker is not None else _sqlite_file_stat_cache_key(db_path)
```

The new `_cli_sessions_streaming_freeze_marker()` is a near-exact mirror of the already-shipped route-level `_session_list_cache_streaming_freeze_marker()` (routes.py:1770) from #4808 — same `_active_stream_ids()` source, same `("streaming", tuple(sorted(...)))` shape, same start/stop semantics. That parity is the reassuring part: the route layer has run this exact freeze keyed on the active-stream set in production since #4808, so extending it to the inner cache is a known-safe pattern, not a new one.

## Two things I checked that make the freeze safe

1. **`_active_stream_ids()` (models.py:424)** unions `STREAMS` with `_cfg.ACTIVE_RUNS`, so the marker stays non-`None` even when the SSE entry detaches but the worker is still in-flight (blocked in provider / unwinding cancel). That means the freeze can't lift prematurely mid-turn and re-trigger the per-poll rebuild storm.

2. **Structural invalidation still works.** Because the frozen key no longer self-invalidates on the content fingerprint mid-stream, the PR correctly adds `clear_cli_sessions_cache()` to `_on_session_list_changed` (routes.py:205). Cron completion / new / renamed / archived / attention all fire that listener and **no per-streamed-token write does**, so real changes still surface within the structural signal, not at TTL expiry. The 5s→30s streaming TTL is then just a backstop, not the freshness path.

## Reviewer validation + the remaining tail

rodboev's two profiling passes against this head (`2d1e1daca`) report `/api/sessions?sidebar_source=webui&exclude_hidden=1` down to ~67–152ms on profile switch and **no reproduction of #4842 on that path**. Worth being precise about scope: their remaining ~4.3s tail is on `/api/models`, which is a different subsystem (provider/model-list reload) entirely untouched by this PR — so it neither validates nor invalidates the streaming-freeze win, it's just the next bottleneck on a different request. That's a separate issue, not a reason to hold this.

## Verdict

CI is green across the 3.11/3.12/3.13 matrix + browser-smoke + lint, the new `test_issue4842_cli_sessions_streaming_freeze.py` asserts the core guarantee (stable key across simulated streamed writes, still advances when idle), and the honest scope note correctly states this targets the trace's mechanism rather than claiming an end-to-end cure for one reporter's data shape. This looks merge-ready. The one follow-up I'd suggest tracking separately: the all-profiles path replaces `context_cache_key` wholesale with `('streaming-frozen', _streaming_marker)` — confirm that a *structural* mutation in profile B (new cron session) during a stream in profile A still invalidates via the listener and isn't masked by the frozen all-profiles key, since the listener clears the whole `_CLI_SESSIONS_CACHE` it should be fine, but it's the one edge worth a line of test coverage.

nesquena-hermes:
Shipped in **v0.51.665** (Release XU, just deployed) — greenlit by Nathan to move through. Freezes the inner _CLI_SESSIONS_CACHE key during streaming (the layer #4672/#4808/#4889 didn't reach), so /api/sessions stops re-running the heavy CLI/cron projection on every poll mid-stream. Gate: Codex SAFE + Opus SHIP (applied one doc-only accuracy fix re the 30s TTL backstop for externally-driven changes), suite 10614. Verified on prod (freeze marker live).

PR #4917, opened 2026-06-25
fix(#4771): keep gateway approval failures actionable

### LINKED ISSUE #4771: Approval UX still not reliably available in gateway-backed WebUI sessions
## Summary

Follow-up to #4479: the WebUI approval UX is still not reliably available when a Hermes operation requests approval from a browser-backed WebUI session.

## Current behavior

When a guarded operation requires user approval, the assistant/tool output indicates that approval is required, but the WebUI does not present a usable approve/deny control to complete the pending approval from the browser.

Depending on deployment/version, this may appear as one of the following:

- No approval card/button is rendered.
- An approval control appears but the run does not continue after approval.
- The session remains blocked and the practical workaround is to run the equivalent command outside the WebUI.

## Expected behavior

When Hermes requests approval, the WebUI should render a visible, actionable approval element and relay the user's approve/deny choice to the backend so the blocked run can continue or fail cleanly.

If approval UX depends on a specific gateway capability, route contract, backend mode, or environment flag, the WebUI should surface that clearly rather than leaving the run blocked without an actionable UI path.

## Reproduction shape

1. Use Hermes WebUI with a remote/gateway-backed Hermes backend.
2. Trigger an operation that requires approval.
3. Observe that the backend/tool output reports an approval request.
4. Attempt to approve from the WebUI.
5. The approval cannot be completed reliably from the browser UX.

## Related issue

- Follow-up to #4479

rodboev:
## Thinking Path

- `#4771` is no longer about the old missing-card or bad-URL regressions. `#4495`, `#4551`, and `#4456` already covered those slices.
- The surviving WebUI-owned bug is the relay-failure blind spot: the browser can show an approval card, but if the backend cannot actually relay that click back to the gateway, the current UI clears the card before it knows whether the backend accepted the response.
- The fix pairs an explicit route-level gateway relay error with a client path that keeps the approval actionable until the backend confirms acceptance.

## What Changed

- `api/routes.py`: return a clear gateway relay failure when `/api/approval/respond` cannot map the approval click back to a relayable gateway run, and preserve pending state on that failure.
- `static/messages.js`: stop clearing the approval card before the response is accepted, keep rerenders from re-enabling an in-flight approval, and surface the backend error when the relay fails.
- `tests/test_approval_unblock.py`, `tests/test_gateway_approval_legacy_path.py`, and `tests/test_issue4771_gateway_approval_failure_ui.py`: prove the failure path stays actionable while the happy-path relay behavior still works, including the poll-rerender duplicate-submit edge.

## Why It Matters

Gateway-backed sessions stop dead-ending on approval failures that the browser can detect but currently hides. Instead of losing the only approval affordance and leaving the run blocked, the WebUI keeps the approval visible and tells the user what went wrong.

## Verification

```text
pytest tests/test_approval_unblock.py tests/test_gateway_approval_legacy_path.py tests/test_issue4771_gateway_approval_failure_ui.py tests/test_approval_queue.py -v --timeout=60
```

- Manual: use a gateway-backed session that emits an approval card without relayable run state, click approve or deny, and confirm the card stays visible with an actionable error instead of disappearing silently. While the request is in flight, confirm rerenders or Enter do not submit the same approval twice.

## Upstream

Related prior work: [#4495](https://github.com/nesquena/hermes-webui/pull/4495), [#4551](https://github.com/nesquena/hermes-webui/pull/4551), and [#4456](https://github.com/nesquena/hermes-webui/pull/4456) fixed earlier gateway approval slices; this follow-up only covers the remaining relay-failure blind spot.

Closes #4771.

## Model Used

GPT 5 via Codex CLI

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -6842,6 +6842,7 @@ from api.route_approvals import (  # noqa: F401 — re-exports for backward comp
     _approval_sse_unsubscribe,
     _approval_sse_notify_locked,
     _approval_sse_notify,
+    _GATEWAY_MIRROR_FLAG,
     reconcile_gateway_pending_mirror_locked,
     submit_gateway_pending_mirror,
     submit_pending,
@@ -18645,6 +18646,32 @@ def _resolve_approval_legacy(sid: str, approval_id: str, choice: str) -> bool:
     return resolved
 
 
+_GATEWAY_APPROVAL_RELAY_UNAVAILABLE = (
+    "Gateway approval could not be relayed because the active run is unavailable. "
+    "Reopen the session or retry after it reconnects."
+)
+
+
+def _gateway_pending_approval_without_run_id(sid: str, approval_id: str) -> bool:
+    with _lock:
+        reconcile_gateway_pending_mirror_locked(sid)
+        queue = _pending.get(sid)
+        if isinstance(queue, list):
+            entries = queue
+        elif queue:
+            entries = [queue]
+        else:
+            entries = []
+        if approval_id:
+            for entry in entries:
+                if isinstance(entry, dict) and entry.get("approval_id") == approval_id:
+                    return bool(entry.get(_GATEWAY_MIRROR_FLAG))
+            return False
+        if not entries or not isinstance(entries[0], dict):
+            return False
+        return bool(entries[0].get(_GATEWAY_MIRROR_FLAG))
+
+
 def _handle_approval_respond(handler, body):
     sid = body.get("session_id", "")
     if not sid:
@@ -18676,6 +18703,18 @@ def _handle_approval_respond(handler, body):
             except (RunnerClientError, ValueError) as exc:
                 return j(handler, {"ok": False, "choice": choice, "relayed": True, "error": str(exc)}, status=502)
             return j(handler, {"ok": True, "choice": choice, "relayed": True})
+        if _gateway_pending_approval_without_run_id(sid, approval_id):
+            return j(
+                handler,
+                {
+                    "ok": False,
+                    "choice": choice,
+                    "relayed": False,
+                    "code": "gateway_run_unavailable",
+                    "error": _GATEWAY_APPROVAL_RELAY_UNAVAILABLE,
+                },
+                status=409,
+            )
     except Exception:
         pass  # fall through to local approval path
 
--- a/static/messages.js
+++ b/static/messages.js
@@ -5559,6 +5559,15 @@ function _syncApprovalTranscriptSpace(card, opts) {
   setTimeout(measure, 420);
 }
 
+function _restoreFailedApprovalResponse(sid, errMsg) {
+  ["approvalBtnOnce","approvalBtnSession","approvalBtnAlways","approvalBtnDeny"].forEach(id => {
+    const b = $(id); if (b) { b.disabled = false; b.classList.remove("loading"); }
+  });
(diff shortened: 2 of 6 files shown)


nesquena-hermes:
Read the full diff on `cron-pr-4917` against `origin/master` plus the surrounding context in `api/routes.py`, `api/route_approvals.py`, `static/messages.js` (`respondApproval` / `showApprovalCard`), and `api/gateway_chat.py`. The relay-failure blind spot is real and the fix is correctly scoped — a few notes below.

## What I verified

**Lock safety (the one I most wanted to rule out).** The new `_gateway_pending_approval_without_run_id` calls `reconcile_gateway_pending_mirror_locked(sid)` from inside `with _lock:`. Since `_lock = threading.Lock()` is non-reentrant (route_approvals.py:40), this would deadlock if it re-acquired — but `reconcile_*_locked` is the already-holds-the-lock variant (its docstring says `CALLER MUST HOLD _lock`), so this is correct. Both `_pending` and `_lock` are imported from `route_approvals` (routes.py:6832-6833), so the helper reads the same queue `submit_pending` writes.

**The early-return is correctly gated.** The 409 only fires when the matched entry has `_GATEWAY_MIRROR_FLAG`, and that flag is set in exactly one place — `reconcile_gateway_pending_mirror_locked` (route_approvals.py:167). Local (non-gateway) approvals never carry it, so they still fall through to the legacy path. Good — no risk of a locally-resolvable approval being swallowed by the 409.

**The scenario is reachable.** `_STREAM_RUN_IDS[stream_id]` is set at stream start (gateway_chat.py:353) and popped on stream end (gateway_chat.py:880). A gateway mirror approval can outlive its run id, which is exactly the `_run_id is None but entry is a gateway mirror` state the new branch catches.

**The UI test models the real contract.** `static/workspace.js:42` throws on any non-2xx (`if(!res.ok)`) with `err.status`/`err.message` attached. `test_failed_gateway_approval_keeps_card_and_reenables_buttons` rejects with a 409 Error, matching that behavior, and asserts `hideCalls == 0` + buttons re-enabled. That's the right harness.

## One small observation (non-blocking)

The route returns a structured `code: "gateway_run_unavailable"` (routes.py:18706), but the client never reads it. `respondApproval`'s catch only uses `e.message`:

```js
} catch(e) {
  const errMsg = (e && e.message) || (t("approval_responding") + " failed");
  _restoreFailedApprovalResponse(sid, errMsg);
}
```

So the `code` is effectively write-only here. That's fine for this PR (the human string is what the user sees), but if you later want a "Reopen session" affordance distinct from a generic retry, the client would branch on `e.status === 409` / `e.body` — worth a follow-up note rather than a change now.

Net: correctness, lock discipline, and the gated fall-through all check out. The `relayed:false` + preserved-pending design is the right call for keeping the card actionable.

nesquena-hermes:
Shipped in **v0.51.653** (Release XI) — thanks @rodboev! 🎉 Closes #4771.

Gate: Codex SAFE TO SHIP — reviewed the approval state-machine edges (failure path stays actionable + retryable, in-flight flag recorded before the API call so re-renders/Enter can't duplicate-submit, poll-rerender preserves the loading choice and can't re-enable mid-submit, happy-path clears exactly once, legacy non-gateway path unchanged), no regression risk. Full suite 10566 passed; your 53 targeted approval tests + the 360-line #4771 failure-path test all green. Pre-merge head re-check confirmed no late push. Verified on prod.

Issue #4918, opened 2026-06-25
WebUI becomes sluggish after long-term use — slow conversation loading (full-message read) plus streaming-time lock contention


jiangtaoru:
---
Title:

WebUI becomes sluggish after long-term use — slow conversation loading (full-message read) plus streaming-time lock contention

---
Body:

Background

After about a month of daily use on a deployed instance, the WebUI becomes noticeably sluggish. As the number of accumulated history conversations and messages grows, performance degrades — it was smooth when freshly installed. The symptoms are:

- Opening an older/longer conversation shows the "Loading conversation…" placeholder for 1–2 s (used to be near-instant).
- Clicking and general UI responsiveness feel laggy.
- Streaming output feels choppy.
- Switching to another conversation while a task is streaming is especially slow — markedly worse than switching when idle.

Investigation points to two independent problems that both worsen as history accumulates:

1. Per-conversation load reads & parses every message row (get_state_db_session_messages), then trims to the msg_limit tail window in Python. Latency scales linearly with the opened conversation's message count, paid on every open. This is the baseline cost — present whether or not anything is streaming.
2. Streaming-time lock contention (#4672): the streaming worker holds the process-global LOCK while writing message rows; the conversation-load path's get_session() acquires that same LOCK. So a load request issued during an active stream queues behind the writer — an additive delay on top of problem 1, which is exactly why "switching during a running task" feels extra slow. Frontend main-thread contention (streaming render + sidebar rebuilds on SSE events) compounds it further.

Both are reported here because they produce the observed "laggy after a month" experience together.

Environment

- Install: git clone + python3 bootstrap.py
- Storage: SQLite state.db (WAL), 237 sessions / 27,348 messages / ~420 MB
- WebUI accessed over the network (not localhost)
- Commit: <paste git rev-parse --short HEAD here>

Reproduction

Same server, same client, same network path — only the conversation differs:

┌──────────────┬──────────┬─────────┬─────────┐
│   session    │ messages │  TTFB   │  total  │
├──────────────┼──────────┼─────────┼─────────┤
│ 8a49eedad68a │ 844      │ 1.752 s │ 2.044 s │
├──────────────┼──────────┼─────────┼─────────┤
│ 9c87b4b51fef │ 240      │ 0.209 s │ 0.359 s │
└──────────────┴──────────┴─────────┴─────────┘

time curl -s -o /dev/null -w "TTFB=%{time_starttransfer}s total=%{time_total}s\n" \
[...]

Issue #3346, opened 2026-06-01
`merge_session_messages_append_only` fails to deduplicate legacy messages in state


thanhtoantnt:
## Summary

`merge_session_messages_append_only` in `api/models.py` fails to deduplicate
state messages that share a legacy merge key (messages without explicit
`id`/`message_id` fields). Duplicate legacy messages in the state list pass
through the merge function unchecked, producing duplicate entries in the
chat history visible to users. Two independent gaps contribute: an early
return path that skips dedup entirely when the sidecar is empty, and a
main merge loop that only records `message_id` keys in `seen_message_keys`
so duplicate legacy keys pass through.

## Impact

**Severity: MEDIUM.** Users see the same message twice in their chat
history when the state database contains duplicate rows, with no error
or warning produced. Once merged, the duplicates persist in the output
and are returned to the UI via `/api/session`. Database compaction,
recovery, or race conditions can realistically produce such duplicate
rows, so the bug is reachable in normal operation.

## Affected Code

**File:** `api/models.py`
**Function:** `merge_session_messages_append_only`

```python
def merge_session_messages_append_only(
    sidecar_messages: list,
    state_messages: list,
    *,
    truncation_watermark=None,
) -> list:
```

## Root Cause

Two independent gaps:

### Gap 1: Early return path (line 3954)

```python
if not sidecar_messages:
    if watermark_timestamp is not None:
        return [
            msg for msg in state_messages
            if (timestamp := _message_timestamp_as_float(msg)) is not None
            and timestamp <= watermark_timestamp
        ]
    return state_messages   # ← returns state verbatim, no dedup
```

When `sidecar_messages` is empty, the function returns `state_messages`
directly — or passes it through a timestamp filter — without any
deduplication, so duplicate entries in state pass through unchanged
regardless of the main loop fix.

### Gap 2: Main merge loop (line 4052)

```python
if key[0] == "message_id":
    seen_message_keys.add(key)   # ← ONLY for message_id keys
```

`seen_message_keys` tracks which merge keys have already been added to
the output. In the sidecar loop, every message's key is added (line
3976). In the state loop, only `message_id` prefixed keys are added —
legacy keys like `('legacy', 'user', 'hello', '', '', '')` are never
recorded, so a second state message with the same legacy key passes
the `if key in seen_message_keys` check (line 4022) and gets appended
again.

## Steps to Reproduce

```python
from api.models import merge_session_messages_append_only

state = [
[...]

PR #4920, opened 2026-06-25
Bound state.db reads for paginated session loads

franksong2702:
## Thinking Path

Issue #4918 points at two separate costs in long-running sessions: slow initial `/api/session?...&msg_limit=N` loads, and broader live-stream lock contention. This PR intentionally fixes only the initial paginated load path.

The unsafe shortcut would be a raw SQL `LIMIT N` against `state.db.messages`, but that can break WebUI sidecar reconciliation because the visible window is computed after sidecar/state-db merge and dedupe. This keeps the existing merge/window contract and only reduces the raw `state.db` rows read for the common initial tail load.

## What Changed

- Added an optional `since_timestamp` filter to `get_state_db_session_messages()`.
- For initial `msg_limit` display loads, derive a conservative timestamp floor from the WebUI sidecar tail, then read only matching recent `state.db` rows.
- Keep full-history reads for full loads, older-page `msg_before` loads, and truncation/edit recovery paths.
- Reuse the precomputed sidecar messages so the optimized path still merges against the same sidecar coordinate space.
- Added regression coverage proving the optimized route returns the same visible window and offset as the full-read merge path.

## Why It Matters

Opening a long WebUI session no longer has to scan and JSON-parse every matching `state.db` row just to render the newest page. This targets the slow load half of #4918 while leaving the deeper streaming/global lock work separate.

Refs #4918.
Related to #4745, but this PR is deliberately narrower and avoids replacing the existing reconciliation contract.

## Verification

- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py tests/test_session_tail_payload.py tests/test_session_message_window_renderable_tail.py tests/test_parallel_session_switch.py::TestMessagePaginationFrontend tests/test_core_data_loss_cases.py tests/test_merge_key_tool_calls.py -q`
  - `82 passed in 4.83s`
- `./.venv/bin/python -m py_compile api/models.py api/routes.py tests/test_webui_state_db_reconciliation.py`

## Risks / Follow-ups

- This does not address the separate streaming/global lock contention described in #4918 Problem 2.
- The timestamp-tail optimization is intentionally disabled for older-page loads and truncation/edit recovery shapes, where older reconciliation rows may be needed.

## Model Used


franksong2702:
Addressed the review comments.

What changed:
- Removed the contributor `CHANGELOG.md` edit from the final PR tree.
- Refactored `_limited_webui_messages_for_display` to delegate to `_limited_webui_messages_for_display_with_sidecar`, and moved the empty-state-db guard into the shared helper so the optimized precomputed-sidecar path and the legacy path cannot drift.
- Added a regression covering the precomputed-sidecar + empty state.db path.

Verification:
- `python3 -m py_compile api/routes.py api/models.py`
- `git diff --check`
- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py::test_limited_display_with_precomputed_sidecar_keeps_empty_state_db_guard tests/test_webui_state_db_reconciliation.py::test_msg_limit_session_load_reads_only_recent_state_db_tail tests/test_webui_state_db_reconciliation.py::test_msg_before_session_load_keeps_full_state_db_reader tests/test_webui_state_db_reconciliation.py::test_state_db_reader_can_filter_by_timestamp_floor -q` (4 passed)
- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py -q` (29 passed)

nesquena-hermes:
Really nice work — the design is exactly right: keep the full sidecar coordinate space, derive a conservative floor, bail to full-read on `msg_before` / `truncation_watermark` / small-sidecar, and merge against the precomputed sidecar. The dual gate (Codex + Opus) confirms Q3 (boundary-without-watermark merge inertness) and Q5 (unchanged full/older-page paths) are byte-identical. But both gates **independently reproduced** a wrong-transcript on the optimized path, so a couple of fixes are needed before it can ship on this load path. The acceptance bar here is byte-identical `messages` / `message_count` / `_messages_offset` vs the full read.

### 🔴 MUST-FIX 1 — NULL/missing-timestamp state.db rows are dropped (reproduced)
The bounded read adds `AND timestamp >= ?`, and in SQLite `NULL >= 3.0` evaluates to NULL → the row is **excluded**. But the full read has no WHERE on timestamp, so a state.db-only row with a NULL/missing timestamp is **included** there and appended into the window. Codex reproduced it via the route: full merge ended with a `state null timestamp only` row and `message_count=501`, the optimized response ended at `sidecar 499` with `message_count=500` — one visible message silently lost. (Your own NOTE at routes.py:6051–6058 already lists "missing-timestamp cases" as must-reconcile — the floor re-introduces exactly that drop.)

**Exact fix** (api/models.py:4776) — make the predicate a true superset:
```python
since_clause = " AND (timestamp IS NULL OR timestamp >= ?)"
```
Add a regression test: a 300+-message, no-watermark session with one NULL-timestamp state-only row, asserting the windowed payload == the full-read payload.

### 🔴 MUST-FIX 2 — `_messages_offset` / `message_count` computed in the wrong coordinate space (reproduced)
When an older state.db-only row reconciles *before* the optimized sidecar tail, the offset/count come out in the filtered space, not the full-merge space. Codex reproduced: full-read offset `471` vs optimized `470`, and the **next older page skipped `sidecar 469`** (paging back loses a message). `_messages_offset` (routes.py:9665, sent at :9851) is consumed by the frontend for older paging (static/sessions.js:2242 → 2717/2764), so it must be in the full coordinate space. Fix: compute `_messages_offset`/`message_count` against the full merged coordinate space even on the optimized path (or only take the optimized path when offsets/counts are provably identical).

### 🟡 SHOULD-FIX 3 — also bail when `truncation_boundary` is set
The bailout checks `truncation_watermark` only, but the helper comment promises edit/truncation recovery shapes stay full-read. Add `truncation_boundary` to the bail condition (or prove byte-identical for boundary-only shapes). Gates rate the boundary-merge currently inert without a watermark, so this is defense-in-depth, but it matches your stated contract.

### Lower-priority (worth a test, not a blocker)
The `ts < floor` residual for a non-NULL state-only row older than `min(sidecar-tail ts)` on a 300+-msg no-watermark session is low-reachability (compaction restamps rows *newer*, and small sidecars self-bail), but the 10× margin is a heuristic, not a proof. A regression test asserting bounded==full for that shape would lock it.

Once #1 + #2 are in (with the two regression tests), this is a strong, high-impact perf win — re-request and I'll re-gate. Excellent that you built it off the safe-shape rather than a raw `LIMIT`. Thanks @franksong2702!

nesquena-hermes:
Thanks for the quick turnaround @franksong2702 — the helper refactor and the new tail-window tests look good. But the **blocker (MUST-FIX 1) is still open**: the predicate at `api/models.py:4776` is still `" AND timestamp >= ?"`, so NULL/missing-timestamp state.db-only rows are still excluded by SQLite (`NULL >= float` → NULL → row dropped). That's the reproduced wrong-transcript both gates hit (full read `message_count=501` ending in a `state null timestamp only` row, optimized `500` ending at `sidecar 499`).

The one-line fix is:
```python
since_clause = " AND (timestamp IS NULL OR timestamp >= ?)"
```
(and add a regression test: a 300+-message, no-watermark session with one NULL-timestamp state-only row, asserting the windowed payload == the full-read payload — none of the four new tests cover the NULL-timestamp class yet).

Also still open from the review:
- **MUST-FIX 2** — `_messages_offset` / `message_count` must be computed in the full merged coordinate space, not the filtered one (reproduced: full offset `471` vs optimized `470`, next older page skipped `sidecar 469`). I don't see an offset/count change in the re-push.
- **SHOULD-FIX 3** — bail to full-read when `truncation_boundary` is set (not only `truncation_watermark`).

Once the `IS NULL OR` predicate + the offset coordinate-space fix are in (with the NULL-timestamp regression test), re-request and I'll re-run the full gate. The shape is right — it's these two byte-identical-fidelity edges that are left.

franksong2702:
Addressed the remaining review blockers in 986d239f.

What changed:
- Kept NULL-timestamp state.db rows in bounded reads with `timestamp IS NULL OR timestamp >= ?`.
- Added a conservative preflight count guard before enabling the tail-window optimization: if state.db and sidecar row counts before the timestamp floor differ, the route falls back to the full state.db read so `messages`, `message_count`, and `_messages_offset` stay in the full merged coordinate space.
- Added the `truncation_boundary` bailout alongside `truncation_watermark`.

Regression coverage added:
- reader-level NULL timestamp retention
- route-level NULL timestamp payload == full-read payload
- older state.db-only row before the floor forces full-read fallback and preserves offset/count
- truncation boundary forces full-read fallback

Local verification:
- `python3 -m py_compile api/routes.py api/models.py`
- `git diff --check`
- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py -q` (33 passed)

Ready for re-gate once CI finishes.

nesquena-hermes:
Great progress — both gates confirm the prior two blockers are **genuinely fixed**: (1) the NULL-timestamp superset (`AND (timestamp IS NULL OR timestamp >= ?)`, no LIMIT, `ORDER BY timestamp ASC, id ASC`) is byte-symmetric with the full read — NULL rows sort first and fall through to the same final append in both paths; (2) the single-anomaly offset case (an extra below-floor state row with no balancing sidecar row) correctly bails to full read on count mismatch. One CORE edge remains, and **both Codex and Opus reproduced it independently** against the real merge code:

### 🔴 MUST-FIX — count-equality guard isn't set-identity (`api/routes.py:~6141`, `api/models.py:~5034`)
The below-floor guard compares **raw pre-floor counts** (`count_state_db_session_messages_before_timestamp` vs `sidecar_before_count`). Count equality ≠ same rows. **Mutual-masking shape:** if an equal number of *sidecar-only* and *state-only* rows both fall below the floor, the counts cancel → the optimized path is taken → but the full merge chronologically inserts a state-only **user** row below the floor (via `_insert_state_message_chronologically`, the no-watermark branch) that the optimized read never fetches.

Reproduced: sidecar 500 + state.db equal pre-floor count but one missing sidecar row + one different old user row → full merge 501 rows / `_messages_offset` 471, optimized 500 / offset 470 — the visible tail-30 stays byte-identical but `message_count`/`_messages_offset` drift by one → **"load older" skips/dupes a row.** That's bounce #2's class re-triggered by a shape the count guard doesn't cover.

**Fix (both gates agree):** replace the raw-count compare with a below-floor **(role, content) key-set** parity/containment check — a cheap key-only query of the pre-floor state rows — and only take `since_timestamp` when the two stores' below-floor key-sets are provably equal; otherwise return `None` / bail to the full read. (Opus's phrasing: require the below-floor merge prefix to be key-identical to the sidecar prefix.)

### 🟡 TEST GAP
The +285 tests prove tail/null/boundary/msg_before parity, but `test_..._bails_when_older_state_db_row_changes_offsets` uses an **assistant**-role row at 199.5 — which the merge skips anyway, so it proves conservative-bail, not insert-divergence prevention. Please add: (a) a **user**-role state-only row below floor (the row type the merge *inserts*), and (b) the **count-masking** shape (equal sidecar-only + state-only below-floor rows) — asserting the optimized `/api/session?...&msg_limit=30` path returns byte-identical `messages`, `message_count`, AND `_messages_offset` to the full read (or bails).

You're one guard away — the NULL + single-anomaly fixes are solid. Swap the count check for a below-floor key-set parity guard + the two regressions and I'll re-gate and fast-track this (#4918 headline). Thanks @franksong2702.

franksong2702:
Addressed the remaining count-equality blocker from `issuecomment-4805615598`.

- Replaced the below-floor raw count guard with sorted `(role, content)` key parity before taking the optimized state.db tail read.
- Added regressions for a state-only older `user` row and the count-masking shape where one sidecar-only row and one state-only row cancel under the old count check.

Verification:
- `python3 -m py_compile api/models.py api/routes.py tests/test_webui_state_db_reconciliation.py`
- `git diff --check`
- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py -q`
- `./scripts/test.sh tests/test_session_tail_payload.py -q`

franksong2702:
Fixed the failing lint check from the latest PR run.

- The failure was Ruff B905 on the new `zip(...)` in `api/routes.py`.
- Updated it to `zip(sidecar_messages, sidecar_timestamps, strict=True)`.

Verification:
- `git fetch --no-tags origin master && python3 scripts/ruff_lint.py --diff origin/master`
- `python3 -m py_compile api/models.py api/routes.py tests/test_webui_state_db_reconciliation.py`
- `git diff --check`
- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py -q`
- `./scripts/test.sh tests/test_session_tail_payload.py -q`

nesquena-hermes:
Real progress — the key-set guard closes the count-masking shape on the (role, content) dimension (both gates confirm: sorted-list/multiset equality of pre-floor `(role,content)` keys → bails to full read when they differ; the `..._bails_when_prefloor_key_counts_mask_offset_change` test passes; NULL-timestamp + schema-degrade paths verified safe). But both gates **independently reproduced** one remaining divergence: the guard's key is **strictly coarser than the merge's own identity**.

### 🔴 MUST-FIX — prefix guard ignores `tool_calls`, but the merge keys on it (`api/routes.py:6135` / `api/models.py:5212`)
The merge identifies rows by **`(role, content, tool_calls)`** everywhere that matters — `_session_message_visible_key` / `_session_message_merge_key` / `_session_message_dedup_key` all fold in `json.dumps(tool_calls)`. And there's a dedicated branch (the #3346 fix, `api/models.py:~6098-6112` — the same distinct-tool-call insert path #4893 touched) that inserts a pre-floor state row as a **distinct** message when it shares content with the sidecar but carries **different `tool_calls`**.

Your guard compares only `(role, content)`, so two rows with the same role+content but different `tool_calls` **false-MATCH** → the optimized `since_timestamp` path is wrongly taken → that distinct tool-call row is dropped from the count → `message_count`/`_messages_offset` drift. Codex reproduced a 500-row transcript: full merge `message_count=501`/`_messages_offset=471`, optimized `500`/`470` — the visible tail matched but the cursor/count drift makes **Load Older skip/dupe**.

**Fix (both gates agree):** make the prefix-parity guard compare the **same identity the merge uses** — an ordered `_session_message_visible_key`-equivalent key including normalized content **plus parsed `tool_calls`**. Update `get_state_db_session_message_keys_before_timestamp` to `SELECT` + parse `tool_calls`, and **return `None` (force full read)** when the columns needed to prove that identity aren't available. Add a regression: pre-floor sidecar row and state row with same `(role,content)` but **different `tool_calls`** → assert the optimized `/api/session?...&msg_limit=N` path bails OR returns byte-identical `messages` + `message_count` + `_messages_offset` to the full read.

This is the last identity gap — once the guard keys on `tool_calls` like the merge does (and bails when it can't prove it), the optimized path is provably byte-identical. Everything else (the #4918 perf win, NULL handling, schema degrade, the role/content masking) is solid. Thanks @franksong2702 — this intersects your own #4893 distinct-tool-call work, so you'll recognize the shape.

franksong2702:
Addressed the latest prefix identity blocker from `issuecomment-4808050590`.

- Updated the below-floor prefix guard to compare `_session_message_visible_key()`-equivalent keys, including parsed `tool_calls`, instead of only `(role, content)`.
- The state.db helper now returns `None` and forces the full read when the schema lacks the columns needed to prove that identity, including `tool_calls`.
- Added a regression for same pre-floor role/content with different `tool_calls`, asserting the route bails to full read and preserves `messages`, `message_count`, and `_messages_offset` against the full merge.

Verification:
- `python3 -m py_compile api/models.py api/routes.py tests/test_webui_state_db_reconciliation.py`
- `git diff --check`
- `git fetch --no-tags origin master && python3 scripts/ruff_lint.py --diff origin/master`
- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py -q` -> 36 passed
- `./scripts/test.sh tests/test_session_tail_payload.py -q` -> 9 passed

franksong2702:
Refreshed the PR branch onto current `origin/master`.

- Merge commit: `425dfc04a6ffd7daf403623e5e0607008e8b09b3`
- No conflicts; the state.db prefix guard and `tool_calls` regression remained intact after the merge.

Verification after refresh:
- `python3 -m py_compile api/models.py api/routes.py tests/test_webui_state_db_reconciliation.py`
- `git diff --check`
- `git fetch --no-tags origin master && python3 scripts/ruff_lint.py --diff origin/master`
- `./scripts/test.sh tests/test_webui_state_db_reconciliation.py -q` -> 39 passed
- `./scripts/test.sh tests/test_session_tail_payload.py -q` -> 9 passed

nesquena-hermes:
Shipped in **v0.51.675** (Release YE, just deployed) — thanks @franksong2702, and genuinely impressive persistence through 5 reconciliation rounds. Opening a long conversation now reads only the tail window (since_timestamp floor) instead of the full state.db history. The prefix-parity guard ended up keying on the FULL merge identity (role, content, parsed tool_calls via _session_message_visible_key), bailing to full read whenever it can't prove the sidecar/state prefixes equal — so message_count/_messages_offset + Load Older paging stay byte-identical. Gate: Codex SAFE + Opus SHIP IT, full suite 10672. Verified live on prod. Fixes #4918 Problem-1.

nesquena-hermes:
Shipped in **v0.51.675** (bounded state.db tail reads for paginated session loads) — the fix is live on master and deployed to prod. Thanks @franksong2702 for the contribution. Closing as shipped (the release landed via a stage/release branch so this PR object didn't auto-close). Please reopen if you still see the issue after upgrading + a hard refresh.

PR #4921, opened 2026-06-25
Reduce session index global lock window

franksong2702:
## Thinking Path

Issue #4918 has two separate performance costs. PR #4920 targets the per-session `state.db` full-read cost on initial paginated loads. This PR targets a smaller piece of the streaming-time lock contention: session saves update the sidebar `_index.json`, and the fast path was doing index file reads, JSON parsing, sorting, and JSON serialization while holding the process-global session `LOCK`.

That lock is also touched by ordinary `/api/session` and `/api/sessions` reads. During active streams, keeping JSON/index work inside the shared lock makes unrelated reads wait longer than necessary.

## What Changed

- Narrowed `_write_session_index()` so `LOCK` protects only in-memory session snapshots.
- Moved existing `_index.json` read/parse, row filtering, sorting, JSON serialization, and disk writes outside `LOCK` while keeping `_INDEX_WRITE_LOCK` as the file-write serialization guard.
- Applied the same lock-window shape to the lazy full-rebuild path's in-memory overlay snapshot.
- Added a regression test that fails if the fast path performs JSON parse/dump while the global session lock is held, while still asserting the index row update and ordering are unchanged.
- Added a release-note entry for the #4918 lock-window improvement.

## Why It Matters

Active streams call `Session.save()`, which updates the session index. Reducing the amount of work done under the global session lock lowers avoidable contention for `/api/session` and `/api/sessions` while a run is streaming.

This does not remove all streaming-time contention, and it does not change the active-stream lifecycle. It is the low-risk lock-window part of #4918, separate from the deeper per-session locking/runtime architecture work.

Refs #4918.
Related to #4920.

## Contract Routing

Task type: runtime/session metadata performance fix.
Touched areas: `_write_session_index()` in `api/models.py`, session sidebar index persistence.
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
Scope boundaries: no frontend changes, no active-stream lifecycle changes, no state.db tail-window changes, no cancellation/compression behavior changes.
Evidence needed before claiming done: regression coverage that index output is preserved and JSON/index work is outside global `LOCK`, plus existing session persistence/list tests.

## Verification

- `./scripts/test.sh tests/test_session_index_lock_window.py tests/test_empty_session_no_disk_write.py tests/test_issue856_session_streaming_state.py tests/test_session_save_mode.py tests/test_session_list_long_history_perf.py tests/test_issue1855_request_diagnostics.py tests/test_issue765_streaming_persistence.py tests/test_metadata_save_wipe_1558.py -q`
  - `83 passed in 5.92s`
- `./.venv/bin/python -m py_compile api/models.py tests/test_session_index_lock_window.py`

## Risks / Follow-ups

- This reduces the shared-lock window for session index updates, but does not fully replace global session locking with per-session snapshots.
- A deeper follow-up can move more streaming writeback coordination behind per-session locks, but that has a larger blast radius across cancel, compression, title refresh, and final writeback paths.

## Model Used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,6 @@
 
 ## [Unreleased]
 
-### Fixed
-
-- **Streaming-time session index updates now hold the global session lock for less work.** `_write_session_index()` now snapshots in-memory session rows under `LOCK`, but reads/parses the existing `_index.json`, sorts, serializes, and writes the updated index outside that lock. This reduces avoidable `/api/session` and `/api/sessions` contention while a running stream saves session metadata, addressing the lock-window half of #4918 without changing the active-stream lifecycle.
-
 ## [v0.51.653] — 2026-06-25 — Release XI (gateway approval failures stay actionable instead of dead-ending)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Streaming-time session index updates now hold the global session lock for less work.** `_write_session_index()` now snapshots in-memory session rows under `LOCK`, but reads/parses the existing `_index.json`, sorts, serializes, and writes the updated index outside that lock. This reduces avoidable `/api/session` and `/api/sessions` contention while a running stream saves session metadata, addressing the lock-window half of #4918 without changing the active-stream lifecycle.
+
 ## [v0.51.653] — 2026-06-25 — Release XI (gateway approval failures stay actionable instead of dead-ending)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


franksong2702:

Verification:
- `python3 -m py_compile api/models.py`
- `git diff --check`
- `./scripts/test.sh tests/test_session_index_lock_window.py -q` (1 passed)

nesquena-hermes:
Shipped in **v0.51.654** (Release XJ) — thanks @franksong2702! 🎉 Part of the #4918 perf work.

Concurrency gate: Codex SAFE + Opus SHIP IT — both confirmed the on-disk index read-modify-write stays fully serialized by the separate, untouched `_INDEX_WRITE_LOCK` (held across the whole function), so narrowing the in-memory `LOCK` introduces no lost-update or corruption race (all 5 adversarial scenarios verified unreachable). Full suite 10567 passed. Added a docstring note naming `_INDEX_WRITE_LOCK` so the 'outside LOCK' wording can't be misread. Verified on prod.

PR #4922, opened 2026-06-25
Avoid repeated session loads in live usage metering

franksong2702:
## Thinking Path

Issue #4918's second half is streaming-time lock contention. After #4921 reduces the session-index save lock window, another hot path remains: live metering calls `_live_usage_snapshot()` repeatedly while tokens and tool events stream, and that snapshot reloaded the session with `get_session(session_id)` every time.

`get_session()` touches the process-global session `LOCK` and can run cache freshness checks. Doing that on every metering tick/token/tool update creates avoidable contention exactly while the user is most likely to switch sessions during an active run.

## What Changed

- Added `_live_usage_session_snapshot()` to reuse the current streaming worker's session object for hot live-usage reads.
- `_live_usage_snapshot()` and `_seed_live_prompt_estimate()` now use that cached/current session path instead of calling `get_session()` on every usage snapshot.
- The helper still falls back to loading the session once before the worker has populated `s`, then reuses that cached object for later early metering reads.
- Added regression coverage proving the hot path does not reload when a current stream session is available, and that fallback loading happens only once.
- Added a release-note entry for the #4918 live-metering lock-contention slice.

## Why It Matters

Live metering can run frequently during active output. Removing repeated session reloads from that path reduces another source of global session-lock traffic while streaming, without changing the usage payload shape or active-stream lifecycle.

This is a narrow follow-up to #4921 and still does not attempt the larger per-session locking/runtime ownership redesign.

Refs #4918.
Related to #4920 and #4921.

## Contract Routing

Task type: runtime/streaming performance fix.
Touched areas: live usage metering inside `api/streaming.py`.
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
Scope boundaries: no frontend changes, no active-stream lifecycle changes, no usage payload shape change, no session index or state.db tail-window changes.
Evidence needed before claiming done: tests showing repeated live usage snapshots do not repeatedly reload sessions, plus existing streaming/live-usage regressions.

## Verification

- `./scripts/test.sh tests/test_streaming_live_usage_estimate.py tests/test_auto_compression_card.py tests/test_issue765_streaming_persistence.py tests/test_issue856_session_streaming_state.py tests/test_session_tail_payload.py -q`
  - `96 passed in 4.91s`
- `./.venv/bin/python -m py_compile api/streaming.py tests/test_streaming_live_usage_estimate.py`

## Risks / Follow-ups

- This removes repeated session reloads from live metering, but does not remove every global `LOCK` touch in streaming.
- The larger follow-up remains per-session coordination for final writeback/cancel/compression/title-refresh paths, ideally after these low-risk lock-window reductions land and are measured together.

## Model Used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Live metering during active streams no longer reloads the session on every usage tick.** The streaming worker now reuses its current session object for `_live_usage_snapshot()` and only falls back to `get_session()` once before the worker has loaded that object. This removes another avoidable global-session-lock touch point while tokens and tool events are streaming, narrowing the lock-contention half of #4918 without changing usage payloads or active-stream state.
+
 ## [v0.51.653] — 2026-06-25 — Release XI (gateway approval failures stay actionable instead of dead-ending)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,6 @@
 
 ## [Unreleased]
 
-### Fixed
-
-- **Live metering during active streams no longer reloads the session on every usage tick.** The streaming worker now reuses its current session object for `_live_usage_snapshot()` and only falls back to `get_session()` once before the worker has loaded that object. This removes another avoidable global-session-lock touch point while tokens and tool events are streaming, narrowing the lock-contention half of #4918 without changing usage payloads or active-stream state.
-
 ## [v0.51.653] — 2026-06-25 — Release XI (gateway approval failures stay actionable instead of dead-ending)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


franksong2702:

Verification:
- `git diff --check`
- `./scripts/test.sh tests/test_streaming_live_usage_estimate.py -q` (5 passed)

nesquena-hermes:
Shipped in **v0.51.655** (Release XK) — thanks @franksong2702! 🎉 #4918 Problem-2 slice.

Gate: Codex SAFE — verified the worker's live `s` is the actively-updated object (no staleness), the cache fallback only matters pre-load + is overwritten by the live object, final usage + `done` payload come from the unchanged writeback path, and the single-reference cache mutation has no tearing race. Suite 10568 passed. Verified on prod.

PR #4923, opened 2026-06-25
Add final stream writeback diagnostics

franksong2702:
## Thinking Path

- #4918 has two separate performance surfaces: session-list reads and live-stream/writeback contention.
- #4920 bounds the paginated session-load path, #4921 narrows the session-index lock window, and #4922 reduces repeated session loads during live usage metering.
- The remaining writeback work should be measured before changing the lock model further, because the successful-stream terminal path mixes result merge, sidecar save, optional persistent-state detection, optional `state.db` sync, and terminal SSE payload construction.
- This PR adds low-noise timing diagnostics around those phases so the next lock-contention slice can target the phase that is actually slow.

## What Changed

- Added final-stream writeback timing helpers in `api/streaming.py`.
- Timed the successful agent-turn writeback phases: `merge_result`, `session_save`, `persistent_state_scan`, `state_sync`, and `done_payload`.
- Logs one debug line only when total final writeback time exceeds `HERMES_WEBUI_STREAM_WRITEBACK_DIAG_MS`.
- Uses a 250 ms default threshold; `0` logs every final writeback and a negative value disables the diagnostic.
- Added focused tests for threshold parsing, timing capture, slow-log behavior, and coverage of the final writeback stages.
- Updated two existing context-window static invariant tests so they still verify save ordering when `s.save()` is wrapped by the timing context.
- Added a changelog note under Unreleased.

## Why It Matters

This keeps the current stream lifecycle and persistence contract intact while making the remaining #4918 writeback contention observable. If users still see a delayed final answer after the earlier slices, the debug log can show whether the delay is coming from the session save, `state.db` sync, persistent-state scan, or payload construction before we make a broader locking change.

## Contract Routing

- Contract family: runtime / streaming / session persistence.
- Evidence: no contract document changes; the active-stream ownership guard, per-session writeback lock, sidecar save, optional `state.db` sync, and terminal `done` event ordering remain in place.
- This PR adds diagnostics around the existing path rather than changing the persistence source of truth.

## Verification

- `./.venv/bin/python -m py_compile api/streaming.py tests/test_streaming_writeback_diagnostics.py tests/test_pr1318_context_length_fallback.py tests/test_pr1341_context_window_persistence.py`
- `./scripts/test.sh tests/test_pr1318_context_length_fallback.py tests/test_pr1341_context_window_persistence.py tests/test_streaming_writeback_diagnostics.py tests/test_issue765_streaming_persistence.py tests/test_stale_stream_writeback.py tests/test_stale_stream_cleanup.py tests/test_streaming_live_usage_estimate.py -q`
- `git diff --check`

## Risks / Follow-ups

- This is intentionally diagnostic, not the final lock-model rewrite for #4918.
- The diagnostic only logs at debug level and only after the threshold is crossed, so normal production noise should stay low.
- Follow-up work can use these timings to decide whether to split or move a specific final-writeback phase outside the per-session critical path.

## Model Used

Refs #4918.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,6 @@
 
 ## [Unreleased]
 
-### Fixed
-
-- **Final stream writeback now emits slow-stage diagnostics without changing the writeback contract.** Successful agent turns time the result merge, session save, persistent-state scan, optional `state.db` sync, and terminal `done` payload phases, logging a debug line only when the full writeback exceeds `HERMES_WEBUI_STREAM_WRITEBACK_DIAG_MS` (250 ms by default, negative disables). This gives the remaining #4918 lock-contention work a measured next step while preserving the existing active-stream lifecycle.
-
 ## [v0.51.653] — 2026-06-25 — Release XI (gateway approval failures stay actionable instead of dead-ending)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Final stream writeback now emits slow-stage diagnostics without changing the writeback contract.** Successful agent turns time the result merge, session save, persistent-state scan, optional `state.db` sync, and terminal `done` payload phases, logging a debug line only when the full writeback exceeds `HERMES_WEBUI_STREAM_WRITEBACK_DIAG_MS` (250 ms by default, negative disables). This gives the remaining #4918 lock-contention work a measured next step while preserving the existing active-stream lifecycle.
+
 ## [v0.51.653] — 2026-06-25 — Release XI (gateway approval failures stay actionable instead of dead-ending)
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


franksong2702:

Verification:
- `git diff --check`
- `./scripts/test.sh tests/test_streaming_writeback_diagnostics.py tests/test_pr1318_context_length_fallback.py tests/test_pr1341_context_window_persistence.py -q` (16 passed)

nesquena-hermes:
Shipped in **v0.51.656** (Release XL) — thanks @franksong2702! 🎉 Completes the #4918 Problem-2 instrumentation.

Gate: Codex SAFE + Opus SHIP IT — both verified the +172/-93 is a behavior-identical restructure (every writeback phase in the same order, same guards, same lock scope; timing context manager exception-transparent; cancel/done-once/env-parse unchanged). Applied Opus's hardening (wrapped the timing log in try/except so a misbehaving handler can't skip the bg-title thread). Suite 10573 passed. Verified on prod.

PR #4935, opened 2026-06-25
fix(#4933): one-click extension install works with no env (bootstrap managed default root)

### LINKED ISSUE #4933: Extensions gallery offers Install when extensions aren't configured → dead-ends on "Extensions not configured" (UX + docs gap)
## Symptom

A community tester tried to install an extension from the Settings → Extensions gallery and got:

```
Install failed: Extensions not configured
```

They then asked where the guide to configure extensions is — and there isn't one on the get-hermes.ai setup site, and the in-app gallery gives no guidance either. So this is two problems wearing one hat: a **UX dead-end** and a **docs gap**.

## Root cause

One-click install (`POST /api/extensions/install` → `install_extension()` in `api/extensions.py`) needs a writable extension root. It calls `_extension_root()`, which returns `None` whenever `HERMES_WEBUI_EXTENSION_DIR` is unset (or points at a non-existent / non-dir path), and then raises:

```python
root = _extension_root()
if root is None:
    raise ExtensionInstallError("Extensions not configured", 404)
```

Extensions are **disabled by default** and only become configured when the admin sets `HERMES_WEBUI_EXTENSION_DIR` to an existing directory *before* starting the server (see `docs/EXTENSIONS.md` → Configuration). So a fresh install has no extension root, and the gallery's **Install** button is dead until the user does that out-of-band env setup — which is documented nowhere user-facing.

The status API already knows this: `GET /api/extensions/status` returns `enabled: false` and `extension_dir_configured: false` in this state. The gallery just isn't acting on it — it still shows an actionable Install button that 404s instead of guiding the user.

## Two fixes wanted

**1. Frontend (the real UX fix) — don't offer an action that can't succeed.** When `GET /api/extensions/status` reports `extension_dir_configured: false` (or `enabled: false`), the gallery should:
- Replace/disable the per-extension **Install** button with an inline "Extensions aren't configured yet" state, and
- Link to a short setup explainer (how to set `HERMES_WEBUI_EXTENSION_DIR` to a directory you control and restart), rather than letting the click hit a 404 toast with no next step.

The error string itself is also bare — `Extensions not configured` with no "here's how" is the worst version of this message.

**2. Docs — there is no user-facing configuration guide.** `docs/EXTENSIONS.md` has the env-var configuration section, but:
- get-hermes.ai/setup has no extensions configuration page at all, and
- nothing in the app points a user from the gallery to that doc.

Add a "Configuring extensions (one-click install prerequisites)" section to the setup site (and link it from the gallery empty/disabled state), covering: extensions are opt-in + default-off, set `HERMES_WEBUI_EXTENSION_DIR` to a directory you control, the trust-model warning (extension JS runs with full session authority — don't point it at a user-writable dir on a shared install), and restart to apply.

## Open design question for the maintainer

Should one-click install **bootstrap a default extension root** (e.g. create `~/.hermes/webui/extensions/` and treat it as the gallery install target when `HERMES_WEBUI_EXTENSION_DIR` is unset) so install works out of the box for a single-user self-hosted instance — or stay strictly opt-in via the env var and only improve the guidance? The trust-model note ("do not point at a user-writable directory") argues for keeping it explicit, but a managed-by-WebUI state dir under `~/.hermes/webui/` is a different thing from "user-writable on a shared box." Worth a decision before implementing fix #1, since it changes whether the disabled-state message says "configure this env var" or "click to enable."

## Suspected area
- `api/extensions.py` — `_extension_root()`, `install_extension()`, `get_extension_status()`
- the Settings → Extensions gallery frontend (status-gating of the Install button)
- `docs/EXTENSIONS.md` + the get-hermes.ai setup site

---

nesquena-hermes:
## What

Make one-click extension install work **out of the box with zero configuration**, closing the UX dead-end in #4933.

Before: the Settings → Extensions gallery showed an **Install** button even when extensions weren't configured, and clicking it failed with `Install failed: Extensions not configured` — because `install_extension()` required `HERMES_WEBUI_EXTENSION_DIR` to be set to an existing directory *before the server started*, which is documented nowhere user-facing. A fresh single-user install could never install an extension without out-of-band env setup.

After: a user opens Settings → Extensions, picks an extension, clicks **Install**, and it just works.

## How (root-cause fix in `api/extensions.py`)

The only thing blocking install was `_extension_root()` returning `None` when `HERMES_WEBUI_EXTENSION_DIR` was unset. Gallery-installed extensions *already* auto-load with no manifest env via `_gallery_installed_runtime_manifest()` — so the fix is just to give the gallery a writable default root:

- **`_default_extension_root()`** — a WebUI-managed default at `STATE_DIR/extensions` (e.g. `~/.hermes/webui/extensions/`), alongside sessions/settings.
- **`_extension_root()`** now resolves: env var (if set + existing) → else the managed default (when it exists) → else `None`. It never auto-creates an admin-specified env path.
- **`_writable_extension_root()`** — used by `install_extension()`; bootstraps (`mkdir -p`) the managed default on first install when no env is set. With the env set it uses that path as-is (admin owns it).
- **`_extension_root_status()`** reports `configured: True` out of the box (a managed install target is always available); `valid` reflects whether the directory exists yet (created on first install).
- The `extension_dir_unavailable` diagnostic warning now only fires for an **explicitly-set-but-broken** `HERMES_WEBUI_EXTENSION_DIR` — the managed default simply not existing yet (pre-first-install) is the normal opt-in state, not a misconfiguration.

`HERMES_WEBUI_EXTENSION_DIR` remains a fully supported **optional override** for operators who want a specific path (checked-out bundle, container mount).

No frontend change is needed — the gallery already renders Install unconditionally and the backend now bootstraps; installed extensions load on the next app-shell render.

## Trust model — unchanged

The managed directory lives in the **WebUI-owned state dir** alongside your sessions and settings. That is a different trust domain from "a world-writable directory on a shared box" — only the WebUI process (and whoever can already write your `~/.hermes` state) can place code there. Installed extension code still runs with full session authority, so the standing guidance holds: only install vetted-gallery or self-trusted extensions. `docs/EXTENSIONS.md` documents the zero-config flow and keeps the trust warning.

## Tests

- Updated `test_extension_status_disabled_by_default`, `test_extension_status_reports_manifest_disabled_when_dir_unconfigured`, and `test_extension_config_disabled_by_default` for the new default-managed semantics (configured=True / valid=False / enabled=False pre-install, no spurious warning), each isolating the state dir.
- **New** `test_install_bootstraps_managed_default_root_without_env` — proves a full install with **no env** bootstraps the managed default, lands files there, and makes the extension load via `get_extension_config()` (the plug-and-play guarantee).

## Verification

- Full suite: **10574 passed, 10 skipped, 1 xfailed, 2 xpassed** (Python 3.11, `tests/` to completion).
- `ruff` clean + `py_compile` clean on changed files.
- Independent Codex review: **SAFE TO SHIP** — verified the root/status/install/static-serving paths and callers, no zip-slip/traversal/symlink regression, correct behavior across all env/default states.

Closes #4933.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **One-click extension install now works out of the box with no configuration (#4933).** Previously the Settings → Extensions gallery offered an **Install** button even when extensions weren't configured, then dead-ended on `Install failed: Extensions not configured` because the install path required `HERMES_WEBUI_EXTENSION_DIR` to be set to an existing directory before the server started — documented nowhere user-facing. Install now bootstraps a WebUI-managed default extension directory under the state dir (`STATE_DIR/extensions`, e.g. `~/.hermes/webui/extensions/`) on first use and installs into it, so any user can open Settings, pick an extension, and click Install with zero environment setup; the extension loads automatically on the next app-shell render. `HERMES_WEBUI_EXTENSION_DIR` remains an optional override for operators who want a specific path (WebUI never auto-creates an admin-specified path). The managed directory lives in the WebUI-owned state dir alongside sessions/settings — a different trust domain from a user-writable directory on a shared box — and the trust model is unchanged (installed code still runs with full session authority, so only install vetted/trusted extensions). `docs/EXTENSIONS.md` documents the zero-config flow.
+
 ## [v0.51.656] — 2026-06-25 — Release XL (stream writeback timing diagnostics)
 
 ### Fixed
--- a/api/extensions.py
+++ b/api/extensions.py
@@ -126,30 +126,85 @@ _EXTENSION_MIME = {
 _TEXT_MIME_TYPES = {"text/css", "application/javascript", "text/html", "image/svg+xml", "text/plain"}
 
 
-def _extension_root() -> Optional[Path]:
-    """Return the configured extension directory, or None when disabled.
+def _default_extension_root() -> Path:
+    """WebUI-managed default extension directory under the state dir.
 
-    A missing or non-directory path disables extensions instead of failing open.
-    The startup docs encourage users to point this at a directory they control.
+    Used when ``HERMES_WEBUI_EXTENSION_DIR`` is unset so one-click gallery
+    install works out of the box on a single-user self-hosted instance with no
+    environment setup. It lives alongside sessions/settings in the WebUI-owned
+    state dir, which is a different trust domain from "a user-writable directory
+    on a shared box" — the loaded code still runs with full session authority,
+    so the trust model is unchanged (see docs/EXTENSIONS.md).
+    """
+    return _extension_state_dir() / "extensions"
+
+
+def _extension_root() -> Optional[Path]:
+    """Return the active extension directory, or None when none is available.
+
+    Resolution order:
+    1. ``HERMES_WEBUI_EXTENSION_DIR`` when set — must be an existing directory,
+       otherwise None (the admin owns that path; we never auto-create it).
+    2. Otherwise the WebUI-managed default (``STATE_DIR/extensions``) when it
+       already exists. The first gallery install creates it on demand
+       (see ``_writable_extension_root``); until then this stays None and the
+       UI reports the same "nothing installed yet" state as before.
     """
     raw = os.getenv(_EXTENSION_DIR_ENV, "").strip()
-    if not raw:
+    if raw:
+        root = Path(raw).expanduser().resolve()
+        if not root.exists() or not root.is_dir():
+            return None
+        return root
+    default_root = _default_extension_root()
+    try:
+        if default_root.is_dir() and not default_root.is_symlink():
+            return default_root.resolve()
+    except OSError:
         return None
-    root = Path(raw).expanduser().resolve()
-    if not root.exists() or not root.is_dir():
+    return None
+
+
+def _writable_extension_root() -> Optional[Path]:
+    """Resolve the extension root for writes, bootstrapping the managed default.
+
+    When ``HERMES_WEBUI_EXTENSION_DIR`` is set we use it as-is (the admin owns
+    it; it must already exist). When unset we create and return the
+    WebUI-managed default so a fresh install can install an extension with zero
+    configuration — plug and play.
+    """
+    raw = os.getenv(_EXTENSION_DIR_ENV, "").strip()
+    if raw:
+        return _extension_root()
+    default_root = _default_extension_root()
+    try:
+        default_root.mkdir(parents=True, exist_ok=True)
+    except OSError:
         return None
-    return root
+    try:
+        if default_root.is_symlink() or not default_root.is_dir():
+            return None
+        return default_root.resolve()
+    except OSError:
+        return None
 
 
 def _extension_root_status() -> Tuple[Optional[Path], bool, bool]:
-    """Return (root, configured, valid) without exposing the configured path."""
+    """Return (root, configured, valid) without exposing the configured path.
+
+    With no ``HERMES_WEBUI_EXTENSION_DIR`` the WebUI-managed default is always
+    available as an install target, so ``configured`` is True (extensions are
+    no longer "not configured" out of the box). ``valid`` reflects whether that
+    managed directory currently exists — it is created on the first install.
+    """
     raw = os.getenv(_EXTENSION_DIR_ENV, "").strip()
-    if not raw:
-        return None, False, False
-    root = Path(raw).expanduser().resolve()
-    if not root.exists() or not root.is_dir():
-        return None, True, False
-    return root, True, True
+    if raw:
+        root = Path(raw).expanduser().resolve()
+        if not root.exists() or not root.is_dir():
+            return None, True, False
+        return root, True, True
+    root = _extension_root()
+    return root, True, root is not None
 
 
 def _new_diagnostics() -> Dict[str, Any]:
@@ -924,7 +979,12 @@ def get_extension_status() -> Dict[str, Any]:
         "stylesheet_count": 0,
         "sidecar_count": 0,
     }
-    if dir_configured and not dir_valid:
+    # Only warn about an unavailable directory when the admin explicitly set
+    # HERMES_WEBUI_EXTENSION_DIR to a path that is missing/not-a-dir. The
+    # WebUI-managed default simply not existing yet (pre-first-install) is the
+    # normal opt-in state, not a misconfiguration worth surfacing.
+    env_dir_set = bool(os.getenv(_EXTENSION_DIR_ENV, "").strip())
+    if env_dir_set and dir_configured and not dir_valid:
         _add_diagnostic_warning(diagnostics, "extension_dir_unavailable", "extension_dir")
 
     if root is None:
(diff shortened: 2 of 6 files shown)


nesquena (approved):
## Review — end-to-end ✅ (clean approval; one non-blocking hardening flagged)

Independent review of the zero-config extension install fix (#4933): the Settings → Extensions gallery showed an **Install** button that dead-ended on `Extensions not configured` unless an admin set `HERMES_WEBUI_EXTENSION_DIR` before boot. This defaults to a WebUI-managed root (`STATE_DIR/extensions`) bootstrapped on first install, so one-click install works out of the box. **Security-sensitive area** (extensions run code with full session authority), so I focused the review on the trust boundary.

### What this ships

`api/extensions.py` (+76/-16), `docs/EXTENSIONS.md` (+27/-4), `tests/` (3 files), `CHANGELOG.md` (+4). Agent-authored; Codex SAFE-TO-SHIP. Closes #4933.

Fresh tarball: WebUI-only extension subsystem; the managed root is under the WebUI state dir. No agent surface. ✓

### The change — bounded and contained

- `_default_extension_root()` → `STATE_DIR/extensions` (WebUI-owned, alongside sessions/settings).
- `_extension_root()` → env (if set + existing dir) → managed default (if it exists, and **not a symlink**) → None. Never auto-creates the env path.
- `_writable_extension_root()` → env as-is, or `mkdir -p` the managed default (rejects if it's a symlink or not a dir). Only `install_extension` uses this.
- `_extension_root_status()` → `configured:True` out of the box; `valid` = managed dir exists yet.
- The `extension_dir_unavailable` warning now fires only for an **explicitly-set-but-broken** env path (the managed default not existing pre-install is normal, not a misconfig).

The diff changes only **where** the root resolves (a contained `STATE_DIR/extensions`) and bootstraps it — it does **not** touch any of the install-path security guards below.

### Security review — the install path is well-guarded (all pre-existing, unaffected)

Now that install is reachable by default, I verified the full `install_extension` chain:

- **Provenance — gallery-only, not arbitrary URLs.** `download_url` must be `https://` AND `parsed_url.hostname in _REGISTRY_ALLOWED_DOWNLOAD_HOSTS` = `{"hermes-webui.github.io"}` — a **single official host**. Zero-config install can only pull vetted gallery extensions; arbitrary code URLs are rejected.
- **Integrity:** exact `sha256` match on the downloaded bytes.
- **SSRF:** `_safe_download` uses an allowlist-redirect handler that rejects cross-host redirects, so the pinned host can't bounce to a private IP.
- **Zip-slip / traversal:** every member is `_fully_unquote_path`-decoded, `_is_safe_relative_path`-checked, then **double-contained** — `resolved.relative_to(root_resolved)` AND `relative_to(ext_dir_resolved)` (raise "Zip-slip detected" on either). Plus `_valid_extension_id`, a 1024-member cap, and a 10× uncompressed-size cap (zip-bomb guard).
- **Symlink leaf guard:** `ext_dir.mkdir(...); if ext_dir.is_symlink(): raise`; the managed-root resolver likewise rejects a symlinked leaf.
- **Auth/CSRF:** `POST /api/extensions/install` (routes.py:10757) runs **after** the CSRF gate (routes.py:10685, install is not csrf-exempt) and, when auth is enabled, behind the password/passkey gate (auth.py:566 → 401). The PR doesn't change these.

**Trust-model assessment:** the security posture shifts from "extensions disabled until an admin sets an env var" to "an authenticated user can one-click install **from the official gallery** into the WebUI-owned state dir." That's bounded (single host, sha256, auth+CSRF) and the managed dir is the same trust domain as the user's own `~/.hermes` sessions/settings — a reasonable, well-guarded zero-config UX. `docs/EXTENSIONS.md` keeps the "extension code runs with full session authority — install only vetted/self-trusted" warning. I concur with the call.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Fresh install, no env, click Install (gallery) | bootstraps `STATE_DIR/extensions`, lands files, loads ✅ |
| `download_url` host ≠ `hermes-webui.github.io` | rejected ("Invalid download URL") ✅ |
| sha256 mismatch / zip-slip member / >1024 files / zip-bomb | rejected ✅ |
| Cross-host redirect during download | rejected (SSRF guard) ✅ |
| Symlinked managed root / ext_dir leaf | rejected ✅ |
| Env set but path missing/non-dir | `configured:True, valid:False` + `extension_dir_unavailable` warning ✅ |
| Env set to a real dir | used as-is, never auto-created ✅ |
| Pre-first-install (managed default absent) | `enabled:False, valid:False`, no spurious warning ✅ |
| Unauth / no CSRF token | 401 / 403 before install runs ✅ |

### Tests

- `test_issue4746_extension_gallery.py` (incl. new `test_install_bootstraps_managed_default_root_without_env` — full no-env bootstrap → install → `get_extension_config()` load) + `test_extension_status_endpoint.py` + `test_extension_hooks.py` — **73/73**; extension blast-radius **101 passed, 2 skipped**.
- **Full suite: 10445 passed / 134 skipped / 0 failures** (deselected the pre-existing darwin CRLF flake).
- `py_compile` clean on extensions.py.

### Minor / hardening observations (non-blocking)

- **Symlink-TOCTOU on a writable shared state-dir mount (Greptile's finding):** the managed-root and ext_dir leaf `is_symlink()` checks block a *pre-planted* symlink, but there's a residual create-then-use window — between the leaf check / `resolve()` and the extraction writes, an external principal who can write the state-dir mount could swap a directory for a symlink (or a *parent* of the managed root is a symlink, which the leaf check doesn't cover). **Not exploitable in the single-user trust model this PR targets** — it requires write access to `~/.hermes/webui`, which already means full state-dir compromise (the attacker could place extension code directly). For shared/multi-tenant state-dir mounts, a hardening pass (re-validate `relative_to(ext_dir_resolved)` inside the write loop, or open with `O_NOFOLLOW`, or resolve+contain the parents) would close it. Recommend tracking as a follow-up; consistent with Greptile rating it "safe to merge after a small cleanup."
- The zip members are written via `write_bytes` (regular files only — a zip can't plant a symlink through this path), which is why the within-extraction risk is limited to an *external* concurrent symlink plant.

### Recommendation

✅ **Approved clean.** Parked at approval — ready for the release agent's merge/tag pipeline (with the shared-mount symlink-TOCTOU hardening as a tracked follow-up).

The fix correctly turns the dead-end Install button into plug-and-play by giving the gallery a contained, WebUI-managed default root and bootstrapping it on first install — without touching any of the install-path security guards (official-host allowlist, sha256, SSRF-safe download, double zip-slip containment, symlink-leaf guards, auth+CSRF). The posture change (installable by default) is bounded to the single official gallery host and stays within the user's own state-dir trust domain. The one residual is a defense-in-depth symlink-TOCTOU that only matters on a writable shared mount (out of the single-user trust model), with the realistic pre-planted case already guarded — flagged for follow-up, not blocking. Ship.

nesquena-hermes:
Shipped in **v0.51.659** (Release XO) via the release branch — thanks @nesquena-hermes! Closes #4933. One-click extension install now works with zero config (bootstraps a managed default extension dir under the state dir). Security-gated: Codex SAFE + Opus SHIP IT (zip-slip/id/symlink/traversal guards all apply to the default root, nothing auto-runs from directory existence, trust model unchanged). Closing as shipped.

PR #4939, opened 2026-06-25
fix(mobile): show the reload button outside standalone mode (#4784)

### LINKED ISSUE #4784: WebUI top-right refresh/reload button missing in the Android app + mobile web (gated to installed-PWA only)
WebUI top-right refresh/reload button missing in the Android app + mobile web (gated to installed-PWA only)
## Summary

A community member (Discord, Android app) reports the top-right **refresh/reload button** is gone "after a recent update" and asks for it back — especially useful on Android and mobile web, where browser reload / pull-to-refresh is less accessible than on desktop.

**The button was not removed.** It was *added* 2026-05-18 (`feat(ui): add PWA standalone reload button and pull-to-refresh gesture`, `c7d0c6343`) and polished 2026-05-24 (PR #2842, `c15148f92`). The real issue is its **visibility gate**: it only renders inside an installed PWA (`display-mode: standalone/fullscreen`), so the native Android WebView shell and a plain mobile-web tab never show it. Treat the reporter's "removed" as a misread — they likely saw it once in a PWA-standalone context and now don't see it in the native app.

## Root cause

`#btnReload` (`.app-titlebar-reload`) exists in the DOM (`static/index.html:156`) but is hidden by default and only un-hidden for installed-PWA display modes:

```css
/* static/style.css */
.app-titlebar-new-chat,
.app-titlebar-reload{display:none; ...}          /* L974 — default hidden */
.pwa-standalone .app-titlebar-reload{display:inline-flex;}  /* L958 */
@media (display-mode: standalone), (display-mode: fullscreen){
  .app-titlebar-reload{display:inline-flex;}     /* L977-980 */
}
```

`pwa-standalone` is toggled only when `isStandalone()` is true (`static/pwa-startup.js:11-23`), which checks `navigator.standalone` + `display-mode: standalone/fullscreen/window-controls-overlay`. A native Android WebView typically reports `display-mode: browser` and `navigator.standalone` is undefined → the class is never set → the reload button stays `display:none`. Same outcome for a normal mobile-web browser tab.

There is **no native-shell detection** in the WebUI page (the only shell-aware hook today is the `meta[name=theme-color]` source-of-truth read by the Swift/WKWebView wrapper — `static/boot.js:1830`).

## Suspected fix (WebUI-side; route confirmed: `nesquena/hermes-webui`)

Make the reload affordance available on mobile contexts beyond installed-PWA, where a hardware/browser reload isn't readily reachable:
- Broaden the un-hide rule to also cover narrow viewports / touch (mobile-web + native WebView), not just `display-mode: standalone`; **or**
- Detect the native shell explicitly (a UA token the Android/iOS shells inject, or a query/flag the shell sets) and toggle a `native-shell` class that un-hides the control. The native shells are thin WebView wrappers, so a small shell-detect hook would also help future shell-only affordances.

Either way the fix lands in WebUI CSS/JS — the same page is what the Android app loads, so fixing the gate fixes the app and mobile web together. Keep it hidden on desktop browsers (real reload/F5 is one keystroke there) to avoid titlebar clutter.

## Use-case surfaces (per standing routing rule)
- **Native Android shell** (the reporter) — primary; no browser chrome, reload hardest to reach.
- **Native iOS shell** — same gap.
- **Mobile web browser** — reporter explicitly wants it here too; pull-to-refresh isn't always available.
- **Installed PWA** — already works (the only case the current gate covers).
- **Desktop browser** — intentionally omit (F5 / browser reload is trivial).

## Repro
1. Open the WebUI in the Android app (or a mobile-web tab) after the recent update.
2. Look at the top-right of the titlebar where a reload button is expected.
3. No button (it's gated to installed-PWA standalone mode only).

## Provenance
Community report from the Android app, captured via the platform bug monitor.

rodboev:
## Thinking Path

- The reload affordance already exists in the titlebar markup, so the bug is a visibility gate, not missing UI wiring.
- The base titlebar CSS hides both mobile icon buttons, then the phone-width block re-shows only new chat while the standalone/fullscreen block separately re-shows reload.
- The narrowest fix is to mirror the existing mobile new-chat visibility rule for `.app-titlebar-reload` and lock that behavior in with a focused titlebar CSS regression.

## What Changed

- `static/style.css`: re-show `.app-titlebar-reload` inside the existing `@media(max-width:640px)` titlebar block while leaving the default hidden rule and the standalone/fullscreen rule untouched.
- `tests/test_mobile_layout.py`: add a focused regression that asserts reload is hidden by default, visible in standalone/fullscreen, and also visible in the phone-width mobile block.

## Why It Matters

Mobile browser sessions and native mobile shells get an explicit reload affordance again without changing desktop titlebar clutter or the standalone behavior that already worked.

## Verification

`pytest tests/test_mobile_layout.py -v --timeout=60`

## Upstream

Closes #4784

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -2462,6 +2462,7 @@
     .app-titlebar{justify-content:space-between;}
     .app-titlebar-hamburger,.app-titlebar-spacer{display:flex;}
     .app-titlebar-new-chat{display:inline-flex;}
+    .app-titlebar-reload{display:inline-flex;}
     .app-titlebar-inner{flex:1 1 auto;}
     .system-health-panel.insights-card{gap:10px;padding:12px;}
     .system-health-head{align-items:flex-start;}
--- a/tests/test_mobile_layout.py
+++ b/tests/test_mobile_layout.py
@@ -914,6 +914,46 @@ def test_titlebar_new_chat_button_mobile_visibility_css():
     )
 
 
+def test_titlebar_reload_button_visibility_css_contract():
+    """Keep reload hidden by default, keep standalone visibility, and expose it on mobile width."""
+    base_rule = _declarations(_rule_body(CSS, ".app-titlebar-reload"))
+    assert _display_hidden(base_rule), "app-titlebar reload button should stay hidden by default"
+
+    standalone_mode_pattern = re.compile(
+        r"@media\s*\(\s*display-mode:\s*standalone\s*\)\s*,\s*"
+        r"\(\s*display-mode:\s*fullscreen\s*\)\s*\{"
+    )
+    standalone_rule_body = None
+    for match in standalone_mode_pattern.finditer(CSS):
+        open_brace = match.end() - 1
+        depth = 0
+        for idx in range(open_brace, len(CSS)):
+            if CSS[idx] == "{":
+                depth += 1
+            elif CSS[idx] == "}":
+                depth -= 1
+                if depth == 0:
+                    block = CSS[open_brace + 1 : idx]
+                    if ".app-titlebar-reload" in block:
+                        standalone_rule_body = block
+                    break
+        if standalone_rule_body is not None:
+            break
+    assert standalone_rule_body is not None, (
+        "standalone/fullscreen media block for titlebar reload could not be parsed"
+    )
+    standalone_rule = _declarations(_rule_body(standalone_rule_body, ".app-titlebar-reload"))
+    assert standalone_rule.get("display") == "inline-flex", (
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.663** (Release XS) — thanks @rodboev! 🎉 Fixes #4784. The reload button now shows in the mobile titlebar outside standalone mode (verified on 390px: hamburger | ✝ Chat | + | ↻, hidden on desktop). Codex SAFE, suite 10602. Verified on prod.

Issue #527, opened 2026-04-15
bug(approval): only one approval can be queued per session — simultaneous approvals overwrite each other


nesquena-hermes:
## Summary

The tool approval system works well for a single pending approval. When a subagent spawns multiple tool calls simultaneously (e.g. several `delegate_task` or `terminal` calls in parallel), each one may block waiting for approval. The current backend stores one pending approval per session in `_pending` (a dict keyed by `session_id`). Only one approval can be queued at a time — subsequent ones overwrite the previous entry.

## Current behaviour

- `_pending[sid]` holds a single approval request per session
- If two tool calls need approval simultaneously, only the last one registered is visible in the UI
- Earlier approval requests are silently dropped or never surfaced
- The UI shows a single approval card and clears it on respond

## Risk

Users running agentic workflows with parallel subagents may unknowingly approve or deny the wrong tool call, or have tool calls hang indefinitely because their approval request was overwritten before the user saw it.

## Proposed behaviour

`_pending[sid]` becomes a queue (list) of pending approvals, each with a unique `approval_id`. The approval card in the UI shows all pending approvals — either stacked as separate cards, or as a paginated "1 of N" view. Each card has its own Approve / Deny button that resolves that specific approval by `approval_id` without affecting the others.

## Implementation notes

- Backend: change `_pending[sid]` from a single object to a list. `submit_pending()` appends; `_pending.get(sid)` returns the list; `/api/approval/respond` pops by `approval_id`.
- Frontend: `static/ui.js` approval card rendering — loop over `pending[]` array from `/api/approval/pending` response, render one card per item.
- Polling: the frontend already polls `/api/approval/pending` on a short interval; the response shape changes from `{pending: {...}}` to `{pending: [...]}`.
- Each approval entry needs a stable `approval_id` (e.g. uuid4) so the respond endpoint knows which one to resolve.

## Files involved

- `api/routes.py` — `_pending` dict → list, `submit_pending`, `_handle_approval_pending`, `_handle_approval_respond`
- `static/ui.js` — approval card loop, per-card respond buttons
- `tests/` — add a test that injects two simultaneous approvals and verifies both are surfaced and independently resolvable

nesquena-hermes:
## Acknowledged: Simultaneous approval queue 🐛

This is a real and important bug for agentic workflows. Thanks for the detailed write-up — the root cause and proposed fix are both correct.

### Confirmed behaviour

[...]

PR #4955, opened 2026-06-26
fix(approval): stale local approval card clears instead of dead-ending on "Approval response not accepted." (#4948 local variant, #4771 follow-up)

nesquena-hermes:
## Summary

Fixes the **`Approval response not accepted.`** dead-end reported by @santastabber on v0.51.666 and by b3nw — a follow-up to #4771 / the local-backend sibling of #4948.

On the **default local in-process backend**, an approval card that is still on screen when its turn ends (user cancel, fork, provider error, or normal completion while the card is pending) has the agent's queue entry dropped and its mirrored `_pending` entry reconciled away. A subsequent Approve/Deny click then sends an `approval_id` that matches nothing, so `_resolve_approval_legacy` returns `False` and `_handle_approval_respond` returned a bare `{ok: false}` with no `error`. Since #4771 changed the frontend from fire-and-forget to surfacing response failures, that bare failure now renders as the **"Approval response not accepted."** toast with a **stuck card** the user can't clear.

This is distinct from the `gateway_run_unavailable` 409 that #4950 (v0.51.666) fixed — that addressed b3nw's *gateway* error string; this is the *second*, contextless failure on the same stale-card path that .666 left exposed.

## Root cause (reproduced empirically)

Confirmed by driving the **real** local-backend approval wiring (`_await_gateway_decision` + the `submit_gateway_pending_mirror` mirror that `api/streaming.py` registers), not a hand-seeded queue:

1. Guarded command → card rendered with `approval_id = X`.
2. Stream ends while pending → agent gateway entry dropped, reconcile purges the `_pending` mirror → nothing pending for the session.
3. Click sends held id X → `_resolve_approval_legacy` line `resolved = bool(pending) or bool(gateway_resolved) or not bool(approval_id)` = `False or False or False` → `{ok:false}` → frontend toast + stuck card.

It is **not** "every approval fails" — a fresh local approval resolves fine; the failure specifically requires the stream to end while the card is up (matches b3nw's "took a minute to run down").

## Fix

- **`api/routes.py`** — new `_session_has_pending_approval(sid)` (acquires `_lock`, reconciles the gateway mirror, reports whether any live `_pending`/`_gateway_queues` entry remains). In `_handle_approval_respond`, when the local resolve returns `False` **and** the session has no pending approval at all, return `{ok: true, stale_cleared: true}` so the UI clears the orphan card.
- **`static/messages.js`** — `respondApproval` clears the orphan card on `stale_cleared` even if the displayed id has drifted.

### Invariants preserved (verified)

- **#527 guard intact** — a stale explicit-id click made *while a different approval is still live* stays `ok:false` and cannot resolve the wrong command (the new clear only fires when **nothing** is pending).
- **Gateway 409 path unchanged** — the `webui_gateway_chat_enabled` branch is untouched.
- `stale_cleared` cannot mask a genuine failure where something is still pending.
- No lock re-entrancy/TOCTOU in the new predicate.

## Tests

New `tests/test_issue4948_local_stale_approval.py` (4 tests, driving the real handler + agent wiring):
- stale card → benign `{ok:true, stale_cleared:true}` (the bug)
- fresh local approval → still resolves, not mislabeled
- stale-id click while a different approval is live → still `ok:false`, live approval untouched (#527)
- `_session_has_pending_approval` predicate (live → empty transitions)

## Verification

- New + existing #4771 regression: 8/8 pass
- Codex review (clean diff): **SAFE TO SHIP** — all invariants verified, 75 approval tests green
- Opus review: **Ship it** — #527 holds at both layers, `ok:true` no-op is safe
- **Full suite: 10,636 passed, 10 skipped, 1 xfailed, 2 xpassed, 0 failures**

Closes the local-backend variant of #4948.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **A stale command-approval card now clears instead of dead-ending on "Approval response not accepted." (#4948 local variant, #4771 follow-up).** On the default local backend, if an approval card was still on screen when its turn ended (cancel, fork, provider error, or normal completion while pending), clicking Approve/Deny sent an id that no longer matched anything and the card got stuck behind that error toast. The server now distinguishes a genuinely stale card (nothing pending for the session → clears the orphan card) from a stale-id click made while a *different* approval is still live (still rejected, so it can never resolve the wrong command — #527 preserved). Reported by santastabber and b3nw.
+
 ## [v0.51.672] — 2026-06-26 — Release YB (faster cron sidebar rebuild on stores missing the message index)
 
 ### Fixed
--- a/api/routes.py
+++ b/api/routes.py
@@ -18695,6 +18695,27 @@ def _gateway_pending_approval_without_run_id(sid: str, approval_id: str) -> bool
         return bool(entries[0].get(_GATEWAY_MIRROR_FLAG))
 
 
+def _session_has_pending_approval(sid: str) -> bool:
+    """True when the session still has any live pending approval to act on.
+
+    Used to tell a benign STALE-CARD click (the card's approval already
+    resolved or its stream ended, so nothing is pending) apart from a stale
+    explicit-id click made WHILE a different approval is still live (which must
+    stay unresolved so it can't accidentally approve the wrong command — #527).
+    Reconciles the gateway mirror first so a purged orphan is not counted.
+    """
+    with _lock:
+        reconcile_gateway_pending_mirror_locked(sid)
+        queue = _pending.get(sid)
+        if isinstance(queue, list):
+            if queue:
+                return True
+        elif queue:
+            return True
+        gw_queue = _gateway_queues.get(sid)
+        return bool(gw_queue)
+
+
 def _handle_approval_respond(handler, body):
     sid = body.get("session_id", "")
     if not sid:
@@ -18769,6 +18790,28 @@ def _handle_approval_respond(handler, body):
         ok = adapter.respond_approval(sid, approval_id, choice).accepted
     else:
         ok = _resolve_approval_legacy(sid, approval_id, choice)
+    if not ok and not _session_has_pending_approval(sid):
+        # The local resolution path returns False when an explicit approval_id
+        # was sent but no matching pending entry exists. There are two distinct
+        # causes, and only one is an error:
+        #   (a) a STALE CARD — the approval the card was rendered from already
+        #       resolved or its stream ended (cancel / fork / provider error /
+        #       completion while pending), so the agent's gateway entry was
+        #       dropped and reconcile purged the mirror. Nothing is pending for
+        #       this session anymore. Before #4771 the frontend was
+        #       fire-and-forget and this silently cleared the card; #4771 began
+        #       surfacing the bare {ok:false} as "Approval response not
+        #       accepted." with a STUCK card (reported by Jamie on .666 / b3nw;
+        #       the local-backend variant of #4948).
+        #   (b) a STALE EXPLICIT ID while a DIFFERENT approval IS live — that
+        #       MUST stay ok:false so a stale click on resolved approval A can
+        #       never resolve the unrelated live approval B (#527 guard).
+        # Distinguish them: when the session has NO pending approval at all,
+        # the click is benign — report it resolved so the UI clears the orphan
+        # card instead of dead-ending. When something IS still pending, keep
+        # the protective ok:false. `stale_cleared` lets the frontend log/branch
+        # without showing an error toast.
+        return j(handler, {"ok": True, "choice": choice, "stale_cleared": True})
     return j(handler, {"ok": ok, "choice": choice})
 
 
--- a/static/messages.js
+++ b/static/messages.js
@@ -5625,12 +5625,35 @@ async function respondApproval(choice) {
       _approvalResponding = null;
       const pendingEntry = _approvalPendingBySession.get(sid);
       const samePending = !!(pendingEntry && pendingEntry.pending && (pendingEntry.pending.approval_id || null) === (approvalId || null));
-      if (_approvalSessionId === sid && _approvalCurrentId === approvalId) {
+      // `stale_cleared` means the server found nothing pending for this session
+      // (the approval already resolved or its stream ended while the card was
+      // up). The orphan card must be cleared unconditionally so it can never
+      // get stuck — even if the displayed id has since drifted. (#4948 local
+      // variant: previously surfaced as a stuck "Approval response not
+      // accepted." toast.)
+      if (result.stale_cleared || (_approvalSessionId === sid && _approvalCurrentId === approvalId)) {
         _approvalSessionId = null;
         _approvalCurrentId = null;
         hideApprovalCard(true);
       }
-      if (samePending) _clearApprovalPendingForSession(sid);
+      if (samePending || result.stale_cleared) _clearApprovalPendingForSession(sid);
+      // Hardening for the narrow stale-clear race: a brand-new approval could
+      // have been parked server-side after the server's empty-check but before
+      // we processed this stale response. The unconditional clear above would
+      // hide that fresh card. Re-query the authoritative server pending state
+      // (same endpoint the fallback poll uses) so any approval that arrived in
+      // the window re-surfaces immediately instead of waiting for the next
+      // SSE/poll tick. Best-effort; poll/SSE remain the backstop. (Opus review
+      // nit on the #4948 fix.)
+      if (result.stale_cleared) {
+        api("/api/approval/pending?session_id=" + encodeURIComponent(sid), {timeoutToast: false})
+          .then(data => {
+            if (data && data.pending && _approvalPromptBelongsToActiveSession(sid)) {
+              showApprovalForSession(sid, data.pending, data.pending_count || 1);
+            }
+          })
+          .catch(() => {});
+      }
       return;
     }
     const errMsg = (result && result.error) || "Approval response not accepted.";
--- a/tests/test_issue4948_local_stale_approval.py
+++ b/tests/test_issue4948_local_stale_approval.py
@@ -1,0 +1,226 @@
+"""Regression tests for the #4948 LOCAL-backend variant (Jamie/.666 report):
+a STALE approval card click must clear gracefully, not dead-end on
+"Approval response not accepted." with a stuck card.
+
+Background
+----------
+#4771 changed the frontend `respondApproval` from fire-and-forget to checking
+`result.ok` and surfacing a toast on failure. That exposed a pre-existing
+local-backend edge: when a guarded command's approval card is still on screen
+but its stream has ended (user cancel / fork / provider error / completion
+while pending), the agent's gateway entry is dropped and reconcile purges the
+mirrored `_pending` entry. A click then sends the held `approval_id`, which
+`_resolve_approval_legacy` cannot match -> returns False -> the handler
+returned a bare `{ok: false}` (no `error`) -> the frontend showed
+"Approval response not accepted." with a STUCK card. (Reported by Jamie on
+v0.51.666 and b3nw on Discord; the local-backend sibling of the gateway-side
+#4948.)
+
+The fix: when the local resolution returns False AND the session has no live
+pending approval at all, treat the click as a benign stale-card clear and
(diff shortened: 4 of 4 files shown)


nesquena-hermes:
## ✅ Gate-clean & merge-ready — prioritized hotfix

**Nathan prioritized this as a hotfix and explicitly overrode the independent-review requirement.** Posting the full gate summary so the release agent can merge + tag + deploy directly.

### Verification (all green on the current head `0140d761`)
- **Full pytest suite: 10,636 passed, 10 skipped, 1 xfailed, 2 xpassed — 0 failures.**
- **Codex (clean rebased diff): SAFE TO SHIP** — verified the #527 guard holds, the gateway-409 path is unchanged, `_session_has_pending_approval` has no lock re-entrancy/TOCTOU, and `stale_cleared` can't mask a genuine pending failure.
- **Codex (re-review after the frontend hardening): SAFE TO SHIP** — the post-`stale_cleared` re-query only fires on `stale_cleared`, gates display through `_approvalPromptBelongsToActiveSession`, has a `.catch` (no unhandled rejection), and (verified against `api/routes.py` reconcile) the just-cleared card returns `pending:null` so there's no flicker/loop.
- **Opus: Ship it** — #527 holds at both the resolve layer and the new gate; returning `ok:true` for a genuinely-stale no-op is safe (no allowlist mutation, agent thread already gone). The one optional hardening nit Opus raised is now applied.

### What's in the diff (4 files)
- `api/routes.py` — `_session_has_pending_approval()` + benign `{ok:true, stale_cleared:true}` when local resolve fails and nothing is pending.
- `static/messages.js` — clear orphan card on `stale_cleared`; re-query `/api/approval/pending` to re-surface any approval parked in the race window.
- `tests/test_issue4948_local_stale_approval.py` — 4 reproduction-backed regression tests.
- `CHANGELOG.md` — `[Unreleased]` entry.

Rebased on fresh `master` (v0.51.668), MERGEABLE, no conflicts. Reporters credited (@santastabber + b3nw co-author trailer). Recommend closing the local-backend variant of #4948 on merge.

### Post-deploy check
Crown-jewel approval surface — after deploy, verify a guarded-command approval still resolves on the local backend, and that a card whose stream ended clears cleanly instead of dead-ending.

nesquena-hermes:
## Gate summary — SAFE, ready for independent review (not merging; self-built)

Deep dual-gate + full suite on this priority approval follow-up. **Codex SAFE TO SHIP + Opus SHIP IT + full suite 10636 passed (0 failures).**

The #527 wrong-approval safety boundary — the crux — is rigorously confirmed:
- The benign `stale_cleared` response is returned **only** after local/adapted resolution fails AND `_session_has_pending_approval(sid)` reports nothing pending (`api/routes.py:18793`). If a *different* approval is genuinely LIVE, that helper returns True and the stale explicit-id click stays the unresolved/error path — it can never clear/approve/deny the live one.
- `reconcile_gateway_pending_mirror_locked` cannot race-purge a genuinely-live entry: it only drops `_GATEWAY_MIRROR_FLAG` mirror entries (never legacy/non-gateway `_pending`), only *reads* `_gateway_queues` (never pops), and can only *recreate* a missing mirror. And `_session_has_pending_approval` independently checks `bool(_gateway_queues.get(sid))` as an authoritative backstop, so even a hypothetical mirror mis-purge can't make it return False while a gateway approval is live.
- The stale path calls **no** resolve/approve logic — it only signals the FE to clear the visual card.
- Frontend (`static/messages.js`) clears the card **only** on `result.ok`; genuine errors still go through `_restoreFailedApprovalResponse` (stay actionable), and a race-arriving approval is re-queried to resurface.
- #4950's gateway-409 path is unchanged (this is the local-backend sibling).

**This is self-built (nesquena-hermes), so per the independent-review rule it should get an independent sign-off (from the nesquena account or a designated reviewer) before merge — same as #4950.** Leaving it gate-clean + ready; not merging autonomously. Reported by @santastabber + @b3nw on v0.51.666.

nesquena-hermes:
## Re-gate on fresh master — SAFE, ready for independent review (still NOT merging; self-built)

Rebased onto current master (it had gone CONFLICTING as ~6 releases shipped overnight) — now MERGEABLE, **code byte-identical** to the prior gate (only the CHANGELOG entry moved). Re-ran the full deep gate on current master:

- **Codex: SAFE TO SHIP** — the #527 wrong-approval boundary still holds: `stale_cleared` is returned only after local/adapter resolve returns false AND nothing is pending (`api/routes.py:18793`); `_pending` is checked after mirror reconciliation with `_gateway_queues` as a live-entry backstop (`:18707`); gateway relay / 409 paths stay ahead of the stale-clear fallback (`:18743`/`:18769`); the FE clears only on explicit `result.stale_cleared` while genuine `{ok:false}` still reaches `_restoreFailedApprovalResponse()` (`static/messages.js:5624`/`5659`). No regression risk; no drift interaction with the #4842/#4893 work that landed since.
- **Opus: SHIP IT** — "the #527 boundary holds by construction (reconcile purges only stale mirrors, never `_gateway_queues` or non-mirror entries; the explicit-id resolver never touches a sibling live entry), the stale branch performs zero resolution, FE auto-dismiss is gated strictly on the server's `stale_cleared` signal, and the #4950/409 gateway path is untouched."
- **Full suite: 10658 passed, 0 failures.** PR **CI: 11/11 green** on the rebased head.

**This is self-built (nesquena-hermes), so per the independent-review rule it is NOT being merged until an independent APPROVED review posts from the nesquena account.** Gate-clean + rebased + ready. Reported by @santastabber + @b3nw (#4948 local-backend variant, #4771 follow-up).

nesquena (approved):
## Review — end-to-end ✅ (approved after fix; pushed the missing CHANGELOG entry)

Independent review of the local-backend stale-approval dead-end — the **second** failure on the same path #4950 (v0.51.666) addressed. After #4950 fixed the `gateway_run_unavailable` 409, a stale card (stream ended while the approval was up) sent an `approval_id` matching nothing → `_resolve_approval_legacy` returned `False` → bare `{ok:false}` → since #4771 surfaces response failures, the **"Approval response not accepted."** toast with a stuck card. Reported by @santastabber (.666) and b3nw.

### What this ships

`api/routes.py` (+43), `static/messages.js` (+25/-2), `tests/test_issue4948_local_stale_approval.py` (+226, new). Agent-authored, co-authored b3nw. Nathan prioritized it as a hotfix and overrode the independent-review requirement — I'm still reviewing on merits. I added the missing CHANGELOG entry (commit below).

Fresh tarball: WebUI-side approval routing. The fix reads the WebUI's own `_pending` / `_gateway_queues` (mirrors of the agent's `_await_gateway_decision` queue) — no agent-side change, gateway relay untouched. ✓

### Fix + the #527 guard — independently proven (the load-bearing security invariant)

New `_session_has_pending_approval(sid)` (under `_lock`, reconciles the gateway mirror first, then reports whether any live `_pending`/`_gateway_queues` entry remains). In `_handle_approval_respond`, when the local resolve returns `False` **and** the session has nothing pending → `{ok:true, stale_cleared:true}` (UI clears the orphan); otherwise the protective `{ok:false}` stays.

The security concern is that this must never let a stale click resolve — or benignly hide — a *different* live approval (#527). I verified it holds at **both** layers and proved it with a backend harness (seeding `_pending`/`_gateway_queues` directly — no agent needed) — **9/9**:
- **Layer 1 (`_resolve_approval_legacy`):** a stale explicit id `A` while live approval `B` is pending returns `False` and does **not** pop/resolve `B` (the explicit-id-not-found path sets `pending=None` and the gateway-queue fallback is gated `if not pending and not approval_id` — so it can't fall through to `B`). Harness: "stale id A → B NOT resolved/popped." ✓
- **Layer 2 (the new gate):** with `B` still live, `_session_has_pending_approval` is `True` → the handler does **not** emit `stale_cleared` → returns `{ok:false}`. Harness: "handler would NOT emit stale_cleared while B live." ✓
- Only when **nothing** is pending does `stale_cleared` fire (harness: stale click after `B` resolved → benign clear). ✓
- `_gateway_queues`-only entries also count as pending (no false `stale_cleared`). ✓

So `stale_cleared`'s `ok:true` is a "card cleared, nothing was pending" signal — it never resolves a command (nothing is pending when it fires) and never masks a live approval.

### Concurrency

`_session_has_pending_approval` takes `_lock` and reconciles, same lock discipline as `_resolve_approval_legacy` — no re-entrancy (they're called sequentially, not nested). The resolve→check TOCTOU resolves safely both ways: if an approval becomes pending between resolve-`False` and the check, the check returns `True` → `{ok:false}` (no clear — safe); if one resolves away, the check returns `False` → clear the orphan (benign, nothing pending). The authority for clear-vs-reject is the check, which errs toward *not* clearing while anything is pending.

### Frontend + the one watch-item (Greptile's race)

On `stale_cleared`, `respondApproval` clears the card unconditionally (tolerating id drift) and then re-queries `/api/approval/pending?session_id=sid` (gated by `_approvalPromptBelongsToActiveSession`, `.catch`-guarded) to re-surface any approval that arrived in the empty-check→response window. This covers the **same-session** race.

The residual (Greptile, non-blocking): in a **cross-session** race — the user clicks A's stale card, navigates to session C which has its own live approval card, and A's `stale_cleared` response then arrives — the unconditional clear hides C's card, and the re-query (scoped to A, gated to the active session) won't restore it. **It self-heals:** C's approval is still pending server-side (never resolved — the backend only cleared A's orphan), so the fallback poll/SSE re-renders C's card within a tick. No wrong-command resolution, no lost approval — a transient visual glitch in a narrow race. A tighter fix would scope the unconditional clear to the response's session (clear on `stale_cleared` only when the shown card is for `sid`, still tolerating id drift within that session). Recommend as a follow-up; non-blocking given it self-heals and the backend invariants are airtight.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Stale card, nothing pending, click | `{ok:true, stale_cleared:true}` → orphan cleared ✅ (the fix) |
| Stale id A click while live B pending | `_resolve` False + B pending → `{ok:false}`, B untouched (#527) ✅ |
| Fresh local approval | resolves normally → `{ok:true}` ✅ (not mislabeled) |
| New approval parked in the empty-check→response window (same session) | re-query re-surfaces it ✅ |
| Cross-session navigate before stale response | C's card transiently hidden → poll/SSE restores it (self-heals) ⚠️ non-blocking |
| Gateway 409 path | unchanged (#4950 branch untouched) ✅ |

### Tests

- `tests/test_issue4948_local_stale_approval.py` (4) + `test_issue4771_local_approval_regression.py` (4) — **agent-dependent** (skip locally, run in CI; PR reports 8/8 + verified fail-before/pass-after). Cover: stale→`stale_cleared`, fresh→resolves, stale-id-while-live→`ok:false` (#527), predicate transitions.
- **Independent backend harness: 9/9** (#527 + `stale_cleared` gating + predicate, above) — no agent needed.
- Locally: approval blast-radius **214 passed, 35 skipped**.
- **Full suite: 10521 passed / 142 skipped / 0 failures** in 225s (deselected the pre-existing darwin CRLF flake; agent-dependent approval tests skip locally, run in CI).
- `py_compile` + `node --check` clean; CI lint + browser-smoke + 9 test shards all green.

### What I caught + pushed — CHANGELOG entry (commit 1ae73b39)

The PR shipped a user-visible behaviour change (stuck approval card → clears) with no CHANGELOG entry; AGENTS.md requires one. Added an `[Unreleased]` → Fixed entry crediting santastabber / b3nw and noting the #527 guard is preserved.

### Recommendation

A correctly-scoped fix for the contextless second failure #4950 left exposed: distinguish a benign stale card (nothing pending → clear the orphan) from a stale-id click while a different approval is live (stays `{ok:false}`). The #527 guard holds at both the resolve layer (a non-matching explicit id never touches another approval) and the new gate (`stale_cleared` only when nothing is pending) — proven 9/9 with a no-agent backend harness. The one residual is a self-healing cross-session transient-hide (poll/SSE restores it; recommended as a follow-up). I pushed the missing CHANGELOG entry. Ship.

nesquena-hermes:
Shipped in **v0.51.673** (Release YC, just deployed) — independent review complete (@nesquena APPROVED 07:01). A stale local approval card now clears instead of dead-ending on 'Approval response not accepted.' The #527 wrong-approval boundary is preserved (a stale-id click while a *different* approval is live still rejects). Gate: Codex SAFE + Opus SHIP IT (re-gated on fresh master), full suite 10658, CI 11/11. Thanks @santastabber + @b3nw for the reports. Verified live on prod.

PR #4968, opened 2026-06-26
feat: export chat to a self-contained, theme-matched HTML file

brick-hard:
## Summary

Adds an **"Export chat as HTML"** button to the composer (next to the existing Markdown / JSON export in Settings) that downloads a single offline, self-contained HTML file of the conversation — and makes that file match the user's **active appearance** instead of being hard-coded to dark.

Previously the export was always dark. This PR captures the live theme palette in the browser via `getComputedStyle(document.documentElement)` and inlines it into the exported file, so the result follows whatever theme + skin the user has set.

## What changed

- **`api/session_export_html.py`** (new): `render_session_html(session, theme, palette=None)`. `_palette_to_css` sanitises the supplied `{var: value}` mapping (regex-restricted names + values, length cap) and emits a `:root{...}` override block. When no palette is supplied (CLI / direct API), the built-in dark/light fallback still renders unchanged.
- **`api/routes.py`**: `/api/session/export` now accepts `theme` and `palette` (base64-encoded JSON) query params. Palette is decoded, JSON-parsed, capped at 64 entries, and passed through. Failures fall back to the built-in palette — never 500.
- **`static/index.html`**: a new `icon-btn` in the composer footer (between *Saved prompts* and *Dictate*), reusing the existing `.icon-btn` style.
- **`static/boot.js`**: shared `exportSessionHTML()` used by both the Settings panel button and the new composer button. Reads the live palette via `getComputedStyle`, base64-encodes, calls the endpoint. Mapping covers the ten core variables the export template uses (`--bg`, `--surface`, `--code-bg`, `--border`, `--text`, `--muted`, `--accent`, `--border2`, ...).
- **`static/i18n.js`**: tooltip / label strings for the new button.
- **`tests/test_session_export_html_palette.py`** (new): 7 tests covering the sanitiser (hex / rgb / `color-mix` accepted; injection attempts dropped; length cap), end-to-end render (override block appears **after** built-in CSS so it wins), and the no-palette fallback path.
- **`CHANGELOG.md`**: entry under `[Unreleased] / Added`.

## Why this design

- **CSS-variable inheritance does the work.** Built-in palette goes first, captured palette goes second — last `:root` wins, no need to enumerate skins server-side.
- **Sanitiser, not trust.** Var names match `[A-Za-z0-9-]+` only; values restricted to a hex / number / `%` / `()` / `var()` / `color-mix()` charset; max-length capped. A hostile palette can't break out of `<style>`. Tested.
- **Backwards compatible.** No palette param → identical output to before; only the theme query was added on top of the existing `format=html`.
- **No new dependencies, no build step.** Vanilla JS, `markdown_it` already in webui.

## Verification

- `./scripts/test.sh tests/test_session_export_html_palette.py` → **7 passed**
- Manual end-to-end: started server, GET `/api/session/export?session_id=<real>&format=html&theme=light&palette=<b64>` → palette override block is injected **after** the built-in CSS (last-wins), file size ~70KB, opens offline.
- Manual injection test: palette `{"bg":"red;}body{display:none","border":"#abc"}` → hostile entry dropped, legit `--border:#abc;` kept, no `display:none` in output.
- No-palette path still produces the existing dark/light export.

## UI notes

Composer button order is now: Attach → Saved prompts → **Export HTML** → Dictate → Voice mode → divider → existing chips. The icon is a file with `< >`, tooltip "Export as a self-contained HTML page", i18n key `export_session_html_tooltip`.

## Risk / rollback

Pure additive. Removing the new button + the `palette` query param restores prior behaviour. No state-layer changes, no contract docs touched.


brick-hard:
Thanks @greptile-apps — fixed the P1 in 12402d71.

`_content_to_text` no longer turns remote `image_url` content into an active Markdown image. Now:

- `data:` URIs (already embedded) are kept as inline `` — they render offline with no network.
- Remote `http(s)` image URLs are flattened to an inert inline-code placeholder `` `[image: <url>]` `` — the URL is preserved as text but the saved file makes no outbound request on open, so it stays self-contained and can't leak a signed/private URL.

Added tests for both paths. All 10 tests pass via `./scripts/test.sh`.

nesquena-hermes:
Pulled the branch (`07633c1`) into a read-only worktree and read `api/session_export_html.py` end to end, focused on the self-contained / no-network guarantee. The two surfaces Greptile flagged are closed, and I checked a third that wasn't in the review — all three hold. Documenting so the "exported file makes no outbound request on open" claim is verifiable from the code.

## 1. Multimodal `image_url` — closed

`_content_to_text` flattens remote images to an inert code-span and only inlines `data:` URIs:

```python
# api/session_export_html.py:62
if url.startswith("data:"):
    parts.append(f"![image]({url})")
else:
    parts.append(f"`[image: {url}]`")
```

A remote `http(s)` image becomes `` `[image: https://...]` `` (a code span), so no `` is emitted and no signed/private URL leaks as a live fetch.

## 2. Palette sanitizer — closed

`_palette_to_css` can't break out of `<style>`: var names are `re.fullmatch(r"[A-Za-z0-9-]+")`, values are restricted to `[#A-Za-z0-9.,%()\-\s]+` and capped at 120 chars:

```python
# api/session_export_html.py:167
val = str(raw_val).strip()
if not val or not re.fullmatch(r"[#A-Za-z0-9.,%()\-\s]+", val):
    continue
if len(val) > 120:
    continue
```

Note the charset has no `:` or `;`, so a value can't terminate a declaration or smuggle a second property — `url()` would survive the charset but it has no `/`, `'`, `"`, so a fetchable `url(http://...)` can't actually be formed. Worth a one-line test asserting a `url(...)`-bearing value is dropped, just to pin that.

## 3. The bypass that *wasn't* in the review — also closed

The above two only matter if Markdown rendering doesn't itself re-introduce raw HTML. It doesn't: `markdown_it` is constructed with `html: False`:

```python
# api/session_export_html.py:19
MarkdownIt("commonmark", {"html": False, "linkify": True, "typographer": False})
    .enable("table")
    .enable("strikethrough")
```

So a message body literally containing `` or `<script>` is escaped, not passed through — there's no second path that turns raw HTML in `content` into an active element. This is the case that would have undermined the #1 fix if `html` were `True`, so it's the important one to lock down.

One side effect to be aware of: `linkify: True` autolinks bare URLs in *text*. Inside the `` `[image: <url>]` `` placeholder the URL sits in a code span, and linkify doesn't run inside code, so that path stays inert — but a remote URL pasted as plain prose elsewhere in a message *will* become a clickable `<a href>` in the export. That's a link, not an auto-fetch, so it doesn't break the offline/no-request guarantee; just flagging it in case "fully inert" is the bar. If so, `linkify: False` would be the lever.

Overall the self-contained guarantee holds for image fetches. The only open nit is the optional `url(...)`-in-palette test to make the no-`/` reasoning explicit for future edits.

brick-hard:
Thanks for the thorough end-to-end pass, @nesquena-hermes — much appreciated.

Addressed the one open nit: pushed `404b7ae` with a test (`test_palette_to_css_drops_fetchable_url_values`) that asserts `_palette_to_css` drops palette values carrying a fetchable `url()` — protocol-bearing (`url(http://...)`), scheme-relative (`url(//...)`), path-bearing (`url(/...)`), and quoted forms — while a legit sibling (`#abc`) still survives. That pins the "no `/`, `:`, `'`, `\"` in the charset, so a real `url()` can't form" reasoning so it can't silently regress if the value charset is ever widened. All 11 tests pass via `./scripts/test.sh`.

On the `linkify: True` side effect: leaving it as-is. As you noted, a bare URL in prose becomes a clickable `<a href>`, not an auto-fetch, so the offline / no-outbound-request guarantee still holds — opening the file makes no request until the user explicitly clicks. The `[image: <url>]` placeholder stays inert since linkify doesn't run inside code spans. Flipping to `linkify: False` would only trade away a genuinely useful affordance (clickable links in an exported transcript) for no security gain, so I'd rather keep it unless you'd prefer the stricter "fully inert" bar — happy to switch if so.

Also rebased onto latest `master` to clear the earlier CHANGELOG conflict.

brick-hard:
Fixed the P1, @greptile-apps — pushed `34b5c34`.

Root cause: `_content_to_text` only flattened **structured** `image_url` parts. A plain-text message body carrying Markdown image syntax — `leak` — was returned unchanged, and markdown_it rendered it into an active ``, so the saved file would fetch the remote URL on open. You're right; the export wasn't always self-contained.

Fix: a single post-render chokepoint, `_neutralize_remote_images`, rewrites any `` whose `src` isn't a `data:` URI into an inert `[image: <url>]` code span. Filtering *after* render catches every path into the HTML — text Markdown images and any future source — not just the structured-content path, mirroring the existing multimodal handling. `data:` URIs stay active (already embedded, offline-safe).

Worth noting: this only manifests when `markdown_it` is installed — the no-lib fallback escapes text into `<pre>` and never emits ``, which is why the test venv had been masking it. The new tests assert the no-``/no-remote-`src`/URL-preserved invariant in **both** environments, and the danger path is now exercised with markdown_it present. 5 new tests; 192 in the export/routes suite pass.

nesquena-hermes:

**Certified head:** `sha:90f871afe65f` (rebased `49a6776f`) · **PR:** #4968 · _brick-hard — "export chat to a self-contained, theme-matched HTML file"_ · **`ux`, concept fit: 4/5**
**Verdict:** Clean, security-careful export feature — both advisors GREEN, and I independently verified the XSS surface (the real risk for chat-content→HTML) is closed: raw HTML escaped, dangerous link schemes stripped, remote images neutralized, palette sanitized before inlining. Ready to merge.

### What I ran (isolated worktree `/tmp/wt-gate-4968`, rebased onto current master)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SAFE TO SHIP** (no findings — probed `<style>` breakout / title / filename / tool+reasoning escaping) |
| Opus (full review) | **COMMENT** — no blocking; "CSS-injection, HTML-injection, offline/no-leak surface closed and thoroughly tested" |
| Full pytest suite (`-p no:xdist`, rebased) | **11138 passed, 0 failed** (to completion) |
| PR's export tests | **15 passed** |
| My XSS verification (in-process) | raw `<script>`/`` escaped ✓ · `javascript:`/`data:`/`vbscript:` links stripped (0 hrefs) ✓ · remote `` neutralized ✓ · palette `</style>` breakout blocked ✓ |

### Security findings (the core surface is closed — I verified each)
- ✅ **Message content is XSS-safe.** Markdown rendered via `MarkdownIt("commonmark", {html: False, linkify: True})` — raw-HTML passthrough disabled, so `<script>`/`` in a message become `&lt;...&gt;`. markdown-it's default `validateLink` strips `javascript:`/`data:`/`vbscript:` link schemes (verified: **0 dangerous hrefs** in the export). Markdown-lib-absent fallback is `<pre>{html.escape(text)}</pre>`. Tool/reasoning content go through the same `_render_markdown` (html:False).
- ✅ **Palette `<style>` injection blocked.** `_palette_to_css` fullmatch-validates names `[A-Za-z0-9-]+` and values `[#A-Za-z0-9.,%()\-\s]+` (no `;{}<>:"'`), so a malicious computed-style value (`</style><script>`, `url(javascript:...)`) is rejected — verified: no `</style>` breakout, no `javascript:` in style, malicious entries dropped.
- ✅ **Remote images neutralized** to inert `[image: url]` placeholders (no live `` beacon in the offline file). Timestamps/labels/sid `html.escape`'d.
- ⚠️ **Optional hardening (non-blocking):** the palette value regex permits `expression(...)` (e.g. `--x: expression(alert(1))` passes) — the legacy IE-only CSS-injection vector. **Not exploitable in any modern browser** (CSS expressions removed since IE11), and the palette comes from the site's own `getComputedStyle(:root)` (not attacker-controlled message content), so the threat is constrained to a victim who already controls their CSS theme. Worth tightening the regex to a color/dimension grammar (or rejecting `expression`) as defense-in-depth, but not a blocker — Codex + Opus both cleared the CSS surface.

### Concept (I scored it — rubric 1–5): **4/5 (strong)**
A portable, self-contained, theme-matched HTML export complements the existing Markdown/JSON export — a common, expected "save my conversation" feature. Sound fit.

**Ready to merge — cert fresh for `sha:90f871afe65f`.** Both advisors GREEN; the XSS surface (the real risk for a chat→HTML export) is closed and I verified it independently (escaped HTML, stripped dangerous schemes, neutralized remote images, sanitized palette). Visible feature (themed export) — `ux` label, glance-worthy but security is the gate and it's solid. Non-blocking polish: add `Thanks @brick-hard. (#4968)` CHANGELOG attribution (Opus); optionally tighten the palette regex to reject `expression(` (IE-only, non-attacker-controlled — defense-in-depth). Was CONFLICTING (CHANGELOG only) — I resolved it (kept all entries); rebases clean. Concept 4/5. Cert valid only at `sha:90f871afe65f`.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close. Cert valid only at `sha:90f871afe65f`; a new push invalidates it → re-gate._

nesquena-hermes:

**Certified head:** `sha:799e6811` · **PR:** #4968 · _brick-hard, export chat to a self-contained, theme-matched HTML file_
**Verdict:** Security re-hardening is solid (remote-image/URL-beacon exfiltration is now closed), full suite green, but the gate reproduced ONE real functional defect: in **dark mode the captured palette is silently ignored**, so the export does NOT match the user's active theme/skin — defeating the feature's headline promise for the default theme. One-line fix below.

### What I ran (isolated worktree `/tmp/wt-gate-4968`, re-gate of the security re-push)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — reproduced the `expression()` palette gap (IE-only, non-block); confirmed the `https://`/`//` image-beacon hardening works on both the structured image-part and markdown-image paths |
| Opus (full review) | One blocking item — dark-mode palette specificity shadowing (reproduced below, real) |
| Full pytest suite | **11261 passed**, 11 skipped, 1 xfailed, 2 xpassed, 16 subtests (ran to completion, `-p no:xdist`) |
| In-process adversarial | XSS / remote-image / palette-injection executed directly against `_render_markdown`, `_neutralize_remote_images`, `_palette_to_css` |
| Auth / scoping | export endpoint behind WebUI auth + `_profiles_match` profile-scoping (404 cross-profile) + `redact_session_data` before render — confirmed |

### Findings

**✅ Security re-hardening (the re-push) verified solid — this part is ready:**
- Remote-image exfiltration **closed**: `_neutralize_remote_images` replaces any `` whose `src` isn't a `data:` URI with an inert `<code>[image: …]</code>` placeholder; `data:` images are preserved. Verified against `https://`, `//protocol-relative`, tab-separated, and mixed remote+data inputs. The markdown-image path (``) and the multimodal `_content_to_text` path both neutralize remote URLs to plain text — no network fetch / IP+UA beacon when the saved file is opened. ✓
- XSS inert: MarkdownIt `html:false` escapes raw `<script>`/``/`<a href=data:…>`; `[x](javascript:…)` is not linkified to an active href. ✓
- Palette `url()`/`@import`/`</style>` breakout **all dropped** — the strict-charset `[#A-Za-z0-9.,%()\-\s]+` excludes `:` and `/`, so `url(https://…)`, `url(javascript:…)`, `@import url(//…)` and a `</style><script>` breakout all fail the allow-list. The new test 813b9a78 pins this. ✓

**⛔ BLOCKER (reproduced) — dark-mode palette is silently ignored (CSS specificity shadowing):**
`api/session_export_html.py:209` — `_palette_to_css` emits the captured palette as a plain **`:root{…}`** block (specificity 0,0,1,0), appended after `_CSS`. But `_CSS` defines the dark theme as **`:root.dark{…}`** (`session_export_html.py:131`, specificity 0,0,2,0). When the export runs in dark mode (`<html class="dark">`, the default theme), the higher-specificity `:root.dark` block **wins regardless of source order**, so the user's captured palette never applies — the export shows the generic built-in dark fallback, not the user's actual theme/skin.
- The "last `:root` wins" premise only holds in **light** mode. Most users are on dark (default) → the theme-matched promise silently fails for them.
- `static/boot.js:1528` `exportSessionHTML()` captures the **already-resolved active-theme colors** via `getComputedStyle` and sends them with `theme=dark|light`, so the palette is authoritative for the active theme — it should override BOTH selectors.
- **Exact fix:** in `_palette_to_css`, emit the override against both selectors:
  ```python
  return ":root,:root.dark{" + "".join(decls) + "}"
  ```
  (specificity now matches `:root.dark`, and source-order makes the palette win in both modes). Add a `theme="dark"` test asserting the override applies to the dark selector (Opus's recommendation; complements the existing light-mode coverage).

**⚠️ SHOULD-FIX (non-blocking, IE-only — same as my 2026-06-29 cert):**
`_palette_to_css` still admits `expression(alert(1))` into the `<style>` block (it contains no `:`/`/` so it passes the charset). Codex reproduced this and it contradicts the PR's stated contract ("blocks `expression(`"). Severity is low — `expression()` is IE-only (ignored by every modern browser) and the palette is the user's own getComputedStyle output, not attacker-controlled chat content (self-XSS at worst on a dead browser). Optional hardening: reject values containing `expression(` / `(` followed by a scheme, or tighten the charset to drop `(` for non-`rgb/hsl/var/color-mix` functions. Not a merge blocker, but cheap to close while you're in `_palette_to_css`.

**RED — one real one-line fix (dark-mode palette specificity) then re-gate; do not merge as-is.** The security re-hardening is genuinely solid and the suite is green — the only blocker is that the feature's theme-match silently no-ops in the default dark theme. The fix is `:root` → `:root,:root.dark` in `_palette_to_css:209` + a dark-mode test. Fold the `expression()` should-fix in the same pass. Concept 4/5 (export-to-HTML belongs; major apps have it). Author `brick-hard` (contributor) — the fix needs their sanitizer output change, so this is changes-requested with the exact one-liner above.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy — that's the release agent's call. This cert is valid only while the head stays at sha:799e6811._

nesquena-hermes:
Re-gated this on a fresh rebase onto current master. The **security re-hardening is solid and verified** (remote-image / URL-beacon exfiltration is closed, XSS inert, palette `url()`/`@import`/`</style>` breakout all dropped by the strict charset allow-list — nice work). Full suite green. But the gate reproduced **one real functional defect** that defeats the feature's headline promise for most users, so it needs a fix before it can land.

### ⛔ Dark-mode palette is silently ignored (CSS specificity shadowing)

`api/session_export_html.py` — `_palette_to_css` emits the captured palette as a plain **`:root { … }`** block (specificity `0,0,1,0`), appended after `_CSS`. But `_CSS` defines the dark theme as **`:root.dark { … }`** (specificity `0,0,2,0`). When the export runs in **dark mode** (`<html class="dark">` — the default theme), the higher-specificity `:root.dark` block **wins regardless of source order**, so the user's captured palette never applies. The export shows the generic built-in dark fallback, not the user's actual theme/skin.

- The "last `:root` wins" assumption only holds in **light** mode. Most users are on dark (the default) → the theme-matched export silently fails for them — which is the whole point of the feature.

**Fix (small):** emit the captured palette with specificity that matches or beats `:root.dark` so it applies in both modes. Either:
- scope it to both: `:root, :root.dark { …captured… }` (and keep light covered), or
- raise specificity to match: e.g. `html:root { …captured… }` / `:root:root { … }`, or
- emit the palette block as the **last** rule AND at `:root.dark` specificity when the captured theme is dark.

Whichever you pick, please add a regression test that asserts the captured palette variables win in dark mode (`<html class="dark">`), not just light — e.g. render an export with a dark active theme and assert the exported CSS resolves `--bg`/`--text` to the captured values, not the `:root.dark` fallback.

### Also (non-blocking, optional)
- `expression(` is still accepted by the palette charset (IE-only, non-exploitable in modern browsers — both advisors cleared it); optional defense-in-depth to reject it.
- Add a `Thanks @brick-hard. (#4968)` CHANGELOG note when it lands (the release process handles changelog, but flagging for attribution).

Everything else is ready — once the dark-mode palette specificity is fixed + a dark-mode regression test added, I'll re-gate and it should go green. Marking `changes-requested`. Thanks @brick-hard!

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (bounce closed)

**Certified head:** `sha:fa4e8d48` (clean rebase, branch `gate-rebase/4968-export-html`) · **PR:** #4968 · _brick-hard, export chat to a self-contained, theme-matched HTML file_
**Verdict:** Bounce closed — my prior dark-mode palette-specificity CORE is fixed (verified in-process AND with a live dark-export screenshot), the optional expression() hardening landed, security surface intact, suite green. Ready to merge from the rebase branch.

### What I ran (rebased worktree `/tmp/wt-rebase-4968`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ merge-base == master tip (`05aa4731`); `git apply` clean |
| Full pytest suite | **11427 passed**, 0 failed |
| In-process palette verification (mine) | ✅ both fixes (below) |
| Live dark-mode export screenshot (mine) | ✅ custom dark palette applies (below) |

### Findings
**✅ CONVERGED — my prior CORE fixed (verified two ways):**
- **Dark-mode palette specificity** (`session_export_html.py:217`): `_palette_to_css` now returns `":root,:root.dark{...}"` (was bare `:root`, which lost to `_CSS`'s `:root.dark` at specificity (0,0,2,0)). The captured palette now wins in dark mode — the feature's theme-match promise holds for the default theme. Verified in-process (emits `:root,:root.dark`) AND visually: I rendered a dark export with a custom pink accent (`#FF2D95`) and screenshotted it — the export renders dark (`#0D0A16` bg) with the pink link/accent applied, NOT the generic built-in dark fallback.
- **expression() hardening** (`:208`): `re.search(r"expression\s*\(", val, IGNORECASE)` now rejects it; legit `color-mix()` still passes, `url()`/`@import`/`</style>` breakout still dropped (no regression to prior-good behavior).

**✅ Security surface intact (re-confirmed):** MarkdownIt `html:false` escapes raw HTML; `_neutralize_remote_images` inert-izes remote `` (beacon defense, data: kept); palette strict-charset blocks `</style>` breakout; export endpoint behind WebUI auth + `_profiles_match` scoping + `redact_session_data` before render. No reverted-master-fix from the master-merge; Codex SAFE; suite green.

### Screenshots
Live dark-mode HTML export with a custom pink palette — renders dark with the `#FF2D95` accent on the link + custom role-badge colors, proving the captured dark palette now applies (the exact CORE that was broken). Clean layout, code block on dark surface, no unstyled areas.

**Ready to merge — use branch `gate-rebase/4968-export-html` (`sha:fa4e8d48`), NOT the PR's stale head `47f6149a`.** Bounce closed: dark-mode theme-match fixed (verified in-process + screenshot), expression() hardened, security holds, Codex SAFE + suite 11427/0. concept 4/5 (export-to-HTML belongs; major apps have it). Credit @brick-hard (co-authored on the rebase). crit=2 ux. Visible feature — dark-export screenshot attached for the glance.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; both fixes verified (in-process + live dark-export render). Cert valid for sha:fa4e8d48._

nesquena-hermes:
Shipped in **v0.51.818** — thanks @brick-hard! 🎉

You can now export a conversation to a self-contained, theme-matched `.html` file: the current theme's palette is baked in, it's fully offline/portable (no external resource loads), conversation content renders as escaped text (no script/style/raw-HTML injection), and remote images are neutralized.

**Gate:** Codex SAFE-TO-SHIP (escapes script/style-break/raw-HTML, rejects `expression()`, self-contained) + full suite 11730 passed + 17/17 export-palette tests + 294 i18n-parity tests + a dark-export screenshot reviewed by the maintainer. Rebased onto current master, credit preserved.

PR #4974, opened 2026-06-26
feat: add sidebar toggle button to rail for hiding chat list

AyoAyomide:
## Summary

Adds a toggle button in the navigation rail (between the spacer and settings icon) to collapse/expand the chat session list sidebar.

## Details

- Places a sidebar panel icon button in the rail at `<nav class="rail">` — right next to the settings icon
- Uses the existing `data-sidebar-collapsed` mechanism for smooth CSS transition, localStorage persistence, and flash prevention
- Highlights with accent background when sidebar is collapsed (visual feedback)
- Also mirrored in the mobile `.sidebar-nav` menu in the same position
- Clicking toggles the `sidebar-collapsed` class on `.layout` and sets `data-sidebar-collapsed` on `<html>`

## Files Changed

- `static/index.html` — added the button to rail and sidebar-nav
- `static/style.css` — highlight state for the button when collapsed
- `static/ui.js` — `toggleSidebar()` function

Closes #??


nesquena-hermes:
Thanks for the rework after the first Greptile pass — the desktop rail button is wired correctly and the `.layout.sidebar-collapsed #sidebarToggleBtn` highlight tracks the real collapsed state. The diff is small and the desktop path is sound. But the **mobile button added to `.sidebar-nav` is a dead no-op**, and that's a functional bug, not the cosmetic `data-label` gap Greptile flagged.

## Code reference

The PR adds this to the mobile nav (index.html:198):

```html
<button class="nav-tab has-tooltip has-tooltip--bottom" onclick="toggleSidebar()" ...>
```

But `toggleSidebar` hard-returns on anything narrower than the desktop breakpoint:

```js
// static/boot.js:392-393
function toggleSidebar(forceState){
  if(!_isDesktopWidth())return; // mobile uses an overlay; never collapse there
```

```js
// static/boot.js:377-378
function _isDesktopWidth(){
  try{return window.matchMedia('(min-width:641px)').matches;}catch(_){return true;}
}
```

## Diagnosis

The `.sidebar-nav` row is rendered **only below 641px** — it's hidden at desktop widths by:

```css
/* static/style.css:1584 */
@media(min-width:641px){.rail{display:flex;}.sidebar > .sidebar-nav{display:none;}}
```

So the new mobile button is visible exactly in the viewport range (`<641px`) where `toggleSidebar()` returns immediately at boot.js:393. Tapping it does nothing — no class toggle, no localStorage write, no visual change. The `data-sidebar-collapsed` collapse mechanism is desktop-only by design; mobile uses a slide-over overlay, not the rail collapse (see the `expandSidebar`/overlay handling around boot.js:302-316).

This means the mobile half of the PR can't work as written. The two honest options:

1. **Drop the mobile button.** This feature is inherently a desktop rail affordance; the mobile overlay already has its own open/close. The PR becomes a clean desktop-only addition (rail button + CSS), which is the part that actually works.
2. **Wire the mobile button to the overlay instead.** If a mobile toggle is genuinely wanted, it needs to drive the mobile sidebar overlay open/close path, not `toggleSidebar()`. That's a different mechanism and a larger change than this diff implies.

I'd lean toward (1) for a small, mergeable PR.

## Two smaller follow-ups

- The rail button at index.html:177 is missing `data-i18n-title` — every sibling rail button has one (`data-i18n-title="tab_settings"`, `"tab_logs"`, etc.), so its tooltip stays untranslated. Add a key like `data-i18n-title="toggle_sidebar"` and a matching entry in `static/i18n.js`.
- If you keep the mobile button, it also needs `data-label="Toggle sidebar"` for the mobile hover tooltip (`content:attr(data-label)` at style.css:1595) — but per the above, removing it is cleaner.

## Verification suggestion

Since the collapse path is desktop-gated, any test should assert the rail button calls `toggleSidebar()` and that the function no-ops below 641px (so reviewers don't reintroduce a mobile caller). A quick manual check: load at <641px, tap the mobile toggle, confirm nothing happens — that reproduces the dead-button behavior described here.

AyoAyomide:
Closing — already have this working locally, no need to push upstream.

PR #4976, opened 2026-06-26
fix: wrap raw cron run output in code block for proper formatting

luandnh:
## What

Cron job run logs (prompt/response in `detail-run-body expanded`) were rendered through `renderMd()` which treated raw JSON/plain text as markdown, breaking formatting: newlines collapsed, special characters interpreted.

## Fix

Before passing to `renderMd()`, detect if the content looks like markdown (starts with `#`, `|`, `>`, or `` ``` ``). If not, wrap it in triple backticks so it renders as a proper `<pre><code>` block with preserved whitespace and formatting.

Also fixed the same issue in the "View full output" button path.

## Verification

- Raw JSON/log output now renders in a code block with proper whitespace
- Markdown content (headings, tables, lists) continues to render normally
- CSS `.detail-run-body` already has `white-space:pre-wrap` and proper background/border


nesquena-hermes:
Thanks @luandnh — the bug is real (cron run logs shouldn't go through Markdown), and the suite + render tests are green. But the dual gate (Codex, verified against the render path) found that the **first-char heuristic is the wrong shape** and leaves two silent-render bugs:

### 🔴 MUST-FIX 1 — heuristic still mangles output that starts with `#`, `|`, `>`, or a fence (`static/panels.js:~901/924`)
Cron run output is **never authored Markdown** — it's logs/JSON/plain text. But the `needsCodeWrap` check skips wrapping when content starts with `#`, `|`, `>`, or ```` ``` ````, so those shapes still get Markdown-rendered: a `#`-prefixed line becomes a heading, `>` becomes a blockquote (text stripped), pipe-delimited log/table rows become an HTML table. `/api/crons/run` returns the saved cron `content`/`snippet` verbatim (api/routes.py:15689-15694), so this is exactly the raw text that must stay literal. The bug persists for any of those first chars.

### 🔴 MUST-FIX 2 — the synthetic fence breaks on output containing its own fence (`static/panels.js:~902/925`)
Wrapping with a fixed triple-backtick fence isn't robust: if the cron log contains a line starting with ```` ``` ````, `renderMd()` (line-anchored fence detection, static/ui.js:4920-4969) closes the synthetic fence early and renders the remainder as Markdown.

**Fix (both at once — render literally instead of wrapping):** drop the `needsCodeWrap` heuristic entirely for cron run output and render it as literal preformatted text — a DOM-created `<pre><code>` with `textContent = output` (then append the usage footer / button outside it). That's immune to both the first-char shapes and the embedded-fence breakout, and it's the correct model (cron output is never markdown). If you prefer to keep `renderMd()`, the fence must be longer than any backtick run in the content — but `<pre><code>` + `textContent` is simpler and safer.

### ❓ Also — the `static/index.html` churn (~+88 lines)
That's large for a "wrap output" fix. Please confirm it's the necessary markup restructure for this change and not an unrelated refactor — keep the diff minimal to the cron-detail render so the review stays tight.

Once cron output renders as literal `<pre><code>` text (no heuristic, no synthetic fence) + the index.html diff is confirmed minimal, I'll re-gate and ship. Good catch on the underlying bug.

PR #4977, opened 2026-06-26
fix: wrap raw cron run output in code block for proper formatting

luandnh:
## What

Cron job run logs (prompt/response in `detail-run-body expanded`) were rendered through `renderMd()` which treated raw JSON/plain text as markdown, breaking formatting: newlines collapsed, special characters interpreted.

## Fix

Before passing to `renderMd()`, detect if the content looks like markdown (starts with `#`, `|`, `>`, or `` ``` ``). If not, wrap it in triple backticks so it renders as a proper `<pre><code>` block with preserved whitespace and formatting.

Also fixed the same issue in the "View full output" button path.

### DIFF
--- a/static/panels.js
+++ b/static/panels.js
@@ -894,9 +894,14 @@ async function _loadRunContent(jobId, filename, runId){
     const expanded = _cronExpansionGet(_cronRunExpandKey(jobId, filename));
     const output = expanded ? (data.content || data.snippet || '') : (data.snippet || data.content || '');
     body.classList.toggle('expanded', expanded);
-    // Render markdown content using the same renderer as chat messages
+    // Render markdown content using the same renderer as chat messages.
+    // Wrap raw output in triple backticks so JSON, logs, and structured text
+    // render as a code block with preserved whitespace and formatting.
+    const renderContent = output.trim();
+    const needsCodeWrap = !renderContent.startsWith('#') && !renderContent.startsWith('|') && !renderContent.startsWith('>') && !renderContent.startsWith('```');
+    const wrapped = needsCodeWrap ? '```\n' + renderContent + '\n```' : renderContent;
     if (typeof renderMd === 'function') {
-      body.innerHTML = renderMd(output);
+      body.innerHTML = renderMd(wrapped);
     } else {
       body.textContent = output;
     }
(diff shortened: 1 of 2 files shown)


luandnh:
Addressed review in 7a123a0:

- Extracted shared `_wrapCronOutput(text)` helper — single source of truth, no duplication
- Expanded heuristic: now checks `~~~`, `-`, `*`, `_`, `[` in addition to `#`, `|`, `>`, `` ``` ``
- Added empty-output guard: returns empty string when input is falsy or blank

Both render paths (initial load + View full output) now call `_wrapCronOutput()`.

luandnh:
Both MUST-FIX items addressed in c0690f8:

**MUST-FIX 1** — dropped the heuristic entirely. Cron run output now renders as literal `<pre><code>` via DOM-created elements with `textContent`, not through `renderMd()`. No Markdown interpretation at all — `#`, `|`, `>`, `` ``` `` and every other character renders verbatim.

**MUST-FIX 2** — no synthetic fence. Since cron output is never processed through `renderMd()`, the embedded-fence breakout issue cannot exist.

Both render paths (initial load + "View full output" button) use the same `<pre><code>` + `textContent` approach.

**index.html churn** — the previous branch had unrelated commits mixed in from the titlebar PR. This branch (`fix/cron-run-output-formatting-v2`) was cherry-picked onto clean master and contains only the 1 cron-related commit. The diff is minimal: `static/panels.js` + `static/style.css`.

nesquena-hermes:
Shipped in **v0.51.677** (Release YG, just deployed) — thanks @luandnh! Cron run logs now render as literal `<pre><code>` text (whitespace preserved, no markdown interpretation, XSS-safe) at both the expanded run body and the View full output paths. This re-do correctly addressed the #4976 review — dropped the index.html churn and replaced the first-char heuristic (which still mangled #/|/>-prefixed output + broke on embedded fences) with the literal-render approach. Gate: Codex SAFE, full suite 10677. Verified on prod.

nesquena-hermes:
Shipped in **v0.51.677** (render cron run output as literal pre/code) — the fix is live on master and deployed to prod. Thanks @luandnh for the contribution. Closing as shipped (the release landed via a stage/release branch so this PR object didn't auto-close). Please reopen if you still see the issue after upgrading + a hard refresh.

PR #4978, opened 2026-06-26
Fix OOB context block replay in model-facing history

Stacey2911:
## Summary

- Add `_strip_oob_blocks()` helper to scrub consumed `[OUT-OF-BAND USER MESSAGE]` markers from message content before it reaches any LLM provider API
- Integrate scrubbing into the main sanitizer (`_sanitize_messages_for_api`), API-safe projection paths, and gateway runs-API conversation history build
- Add 10 regression tests covering OOB variants, CRLF handling, incomplete block preservation, tool-chain continuity, and gateway sanitization

## Why

Consumed out-of-band user message markers can remain embedded inside historical `context_messages` content and be replayed into later provider-facing conversation history. This causes stale internal control data to leak into model prompts, wasting tokens and potentially confusing the agent with consumed steer blocks.

Related: nesquena/hermes-webui#4283

Tangentially related: NousResearch/hermes-agent#40683, which concerns busy-state `/steer` handling rather than WebUI history sanitization.

## Implementation

- **`api/streaming.py`**: Added `_OOB_USER_MESSAGE_BLOCK_RE` and `_strip_oob_blocks()` to scrub consumed OOB blocks from strings, list-based content parts, and dict content structures.
- **Sanitizer/projection paths**: Applied scrubbing in `_sanitize_messages_for_api()`, `_api_safe_message_positions()`, and `_restore_reasoning_metadata()._safe_projection()` so model-facing history and position matching use the same OOB-clean projection.
- **`api/gateway_chat.py`**: Applied the same content scrubbing to runs-API `conversation_history` built from `context_messages` and `prefill_messages`.
- **Reasoning compatibility**: Kept `reasoning_content` in `_API_SAFE_MSG_KEYS` for provider compatibility. Display-only reasoning fields remain excluded by the existing key filter.

The OOB regex handles em dash and hyphen markers, optional no-suffix markers, Unix LF and Windows CRLF line endings, multiline block bodies, and multiple blocks per string. Incomplete/unclosed blocks are preserved.

## Safety

- Preserves assistant `tool_calls` and matching tool-result structure
- Does not modify role alternation, orphaned-tool handling, or message structure
- Only scrubs matched internal OOB control blocks from historical content before provider use
- Leaves incomplete OOB blocks untouched to avoid accidental data loss

## Tests

```bash
python3 -m py_compile api/streaming.py api/routes.py api/gateway_chat.py api/models.py
python3 -m pytest tests/test_context_history_sanitization.py -v
python3 -m pytest tests/test_issue4283_recovered_context_replay.py tests/test_orphaned_tool_messages.py tests/test_partial_bloat_fix.py tests/test_native_image_attachments.py -q
```

Results:

* `py_compile`: passed
* New regression tests: 10 passed
* Existing relevant test slice: 84 passed

## Scope

Only 3 files changed:

* `api/streaming.py`
* `api/gateway_chat.py`
* `tests/test_context_history_sanitization.py`

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -299,6 +299,8 @@ def _run_gateway_runs_api_streaming(
         except Exception:
             logger.debug("Failed to build runs-API multimodal attachment payload", exc_info=True)
             message_content = str(msg_text or "")
+    from api.streaming import _strip_oob_blocks
+
     instructions_parts = []
     conversation_history = []
     for entry in getattr(session, "context_messages", None) or []:
@@ -309,6 +311,7 @@ def _run_gateway_runs_api_streaming(
             continue
         content = entry.get("content")
         if content is not None:
+            content = _strip_oob_blocks(content)
             conversation_history.append({"role": role, "content": content})
     for entry in prefill_messages or []:
         if not isinstance(entry, dict):
@@ -323,6 +326,8 @@ def _run_gateway_runs_api_streaming(
             continue
         if role not in {"user", "assistant"}:
             continue
+        if content is not None:
+            content = _strip_oob_blocks(content)
         conversation_history.append({"role": role, "content": content})
     run_input = message_content
     if isinstance(run_input, list):
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -1424,6 +1424,8 @@ from api.workspace import set_last_workspace
 # Fields that are safe to send to LLM provider APIs.
 # Everything else (attachments, timestamp, _ts, etc.) is display-only
 # metadata added by the webui and must be stripped before the API call.
+# `reasoning_content` is provider-facing for reasoning-capable models. Display
+# metadata such as `reasoning`, `thinking`, and `_reasoning` stays omitted here.
 _API_SAFE_MSG_KEYS = {'role', 'content', 'tool_calls', 'tool_call_id', 'name', 'refusal', 'reasoning_content'}
 
 _NATIVE_IMAGE_MAX_BYTES = 20 * 1024 * 1024
@@ -3532,6 +3534,32 @@ def _strip_native_image_parts_from_content(content):
     return clean_parts
 
 
+_OOB_USER_MESSAGE_BLOCK_RE = re.compile(
+    r'\[OUT-OF-BAND\s+USER\s+MESSAGE(?:\s*(?:—|-)\s*.*?)?\]\r?\n.*?\[/OUT-OF-BAND\s+USER\s+MESSAGE\]',
+    re.DOTALL | re.IGNORECASE,
+)
+
+
+def _strip_oob_blocks(content):
+    """Remove consumed [OUT-OF-BAND USER MESSAGE ...] blocks from content.
+
+    These markers are internal control data that should never reach the model.
+    They can appear as plain strings or inside list-based content parts.
+    """
+    if isinstance(content, str):
+        return _OOB_USER_MESSAGE_BLOCK_RE.sub('', content)
+    if isinstance(content, list):
+        return [_strip_oob_blocks(part) for part in content]
+    if isinstance(content, dict):
+        return {
+            key: _strip_oob_blocks(value)
+            if isinstance(value, (str, list, dict))
+            else copy.deepcopy(value)
+            for key, value in content.items()
+        }
+    return content
+
+
 def _content_has_reasoning_only_parts(content) -> bool:
     if not isinstance(content, list) or not content:
         return False
@@ -3639,6 +3667,8 @@ def _sanitize_messages_for_api(messages, *, cfg: dict = None):
         sanitized = {k: v for k, v in msg.items() if k in _API_SAFE_MSG_KEYS}
         if is_recovered:
             sanitized['_recovered'] = True  # temporary marker — stripped before return
+        if 'content' in sanitized:
+            sanitized['content'] = _strip_oob_blocks(sanitized['content'])
         if strip_native_images and 'content' in sanitized:
             sanitized['content'] = _strip_native_image_parts_from_content(sanitized.get('content'))
         if sanitized.get('role'):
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Thanks @Stacey2911 — scrubbing consumed internal control markers out of model-facing history is a sound principle, and the regex/sanitizer integration looks careful (em-dash + hyphen variants, CRLF, multi-block, incomplete-block preservation). Before I run the full gate on this (+304 on the model-facing sanitizer is a sensitive surface), I need to confirm the **premise is real in WebUI**, because I can't reproduce it from the code yet:

**The `[OUT-OF-BAND USER MESSAGE]` marker does not appear anywhere in the WebUI codebase.** A repo-wide grep finds zero references (only unrelated "out-of-band" prose in test docstrings + vendored `mcp`). WebUI's own mid-turn steering uses the `/steer` mechanism (`_handle_chat_steer`, `agent._pending_steer`) — it does **not** wrap steered text in `[OUT-OF-BAND USER MESSAGE]` markers. That marker is an **agent-core transport wrapper** (injected when delivering a mid-turn message into a tool result), not something WebUI writes.

So the open question: **does that marker actually end up in WebUI's persisted `context_messages` content that gets replayed to the LLM?** If the marker only lives in the agent-core's ephemeral turn assembly and is never persisted into the messages WebUI reads back, then `_strip_oob_blocks()` is scrubbing something that isn't present in WebUI's history — a no-op against a non-bug here (the right fix would live in the agent core).

**Could you attach a concrete repro?** Specifically: a real persisted `context_messages` / session-state entry (captured from an actual WebUI session, redacted as needed) whose `content` contains a literal `[OUT-OF-BAND USER MESSAGE]` block, plus the path by which it reaches `_sanitize_messages_for_api()` / the runs-API `conversation_history` build. If you can show the marker genuinely reaches provider-facing history via WebUI, I'll run the full Codex+Opus+suite gate immediately — the implementation looks solid, I just need to confirm it's fixing a live WebUI path and not an agent-core-only artifact. (Also: CI hasn't run yet on this branch — a push to trigger it would help.)

Stacey2911:
## Evidence: OOB Blocks Persist in WebUI Session State

Thanks — that is a fair concern. I checked the persisted WebUI session state specifically, because I agree the marker is not authored by WebUI source code.

I can confirm the literal `[OUT-OF-BAND USER MESSAGE]` marker is present in persisted WebUI session JSON, not only in agent-core transient turn assembly.

### Findings

A read-only grep of my local WebUI runtime/session state found:

- **61 persisted session JSON files** under `~/.hermes/webui/sessions/*.json` containing the marker
- **~858+ total line matches** across session files, run journals, and turn journals
- **0 matches in live WebUI source code** under `api/*.py`

So the marker appears to be injected upstream at runtime, then persisted by WebUI in session state.

### Representative redacted persisted entry

This is the minimal structure from an affected session file. Private tool output and user text are redacted, but the field shape and literal markers are preserved:

```json
{
  "role": "tool",
  "content": "[REDACTED tool output before]\n\n[OUT-OF-BAND USER MESSAGE — a direct message from the user, delivered mid-turn; not tool output]\n[REDACTED consumed user steer text]\n[/OUT-OF-BAND USER MESSAGE]"
}
```

Structural notes:

* This appears inside `context_messages[].content`
* The same marker shape also appears in persisted `messages[]`
* The OOB block is appended to tool-result content, not stored as a separate field

### Replay path

The reason I scoped this PR to WebUI is that once the agent-core marker is embedded in persisted WebUI session state, WebUI later reads that same `context_messages` content back into provider-facing history.

Main path:

```text
Session JSON
→ session.context_messages
→ _context_messages_for_new_turn(...)
→ _sanitize_messages_for_api(...)
→ provider-facing conversation_history
```

Gateway/runs path:

```text
Session JSON
→ session.context_messages
→ gateway_chat.py builds conversation_history from context_messages
→ provider-facing runs-API history
```

This PR therefore handles the issue at WebUI's persistence/replay boundary. It does not attempt to change agent-core `/steer` behavior or the origin of the marker.

### Validation

After rebasing onto current `upstream/master`, local validation passed:

```bash
python3 -m py_compile api/streaming.py api/routes.py api/gateway_chat.py api/models.py
python3 -m pytest tests/test_context_history_sanitization.py -v
python3 -m pytest tests/test_issue4283_recovered_context_replay.py tests/test_orphaned_tool_messages.py tests/test_partial_bloat_fix.py tests/test_native_image_attachments.py -q
```

Results:

```text
py_compile: passed
new regression tests: 10 passed
existing relevant slice: 84 passed
```

I'm avoiding attaching raw session JSON because it contains private conversation/tool context, but the redacted snippet above preserves the exact persisted field shape and marker content needed to verify the WebUI replay path.

nesquena-hermes:
Shipped in **v0.51.705** 🎉 — thank you @Stacey2911, and thanks for the thorough persistence evidence.

You were right: I independently verified on a real prod session file that the literal `[OUT-OF-BAND USER MESSAGE]…[/OUT-OF-BAND USER MESSAGE]` block **is** persisted in WebUI session JSON — present in both `messages[]` and `context_messages[]` at the same index, appended onto a `role:tool` content string — and replays into provider-facing history via `_sanitize_messages_for_api` / the gateway `conversation_history`. The premise holds; the fix addresses a real leak.

Your sanitize-boundary approach is the right architecture: the marker stays intact in the persisted transcript (display/reload/audit fidelity) and is stripped only on the way to the model. The recursive `_strip_oob_blocks` is non-mutating and idempotent, and it's wired into all the provider-facing chokepoints (and the compression path inherits it via `_sanitize_messages_for_api`).

**One maintainer hardening on top** (from the Opus review gate): I relaxed the opening pattern from a mandatory `]\r?\n` to `]\s*?` so a *single-line* OOB block is also stripped — `DOTALL .*?` already spans newlines, so this just adds robustness against a format variant that omits the newline, with no change to the verified multiline behavior.

Full gate: Codex SAFE TO SHIP + Opus SHIP + full suite green (10,710 passed; the 4 failures are pre-existing order/env flakes unrelated to streaming) + your 10 tests. Closing as shipped — reopen if you still see the marker reach the model after upgrading.

Issue #1558, opened 2026-05-03
P0 DATA LOSS: metadata-only Session.save() wipes conversation history (v0.50.279+)


nesquena-hermes:
## Summary — P0 data loss

**Severity: P0.** Every active conversation on v0.50.279 — v0.50.282 is at risk of being silently wiped on the next SSE reconnect after a server restart. Reported by a user on v0.50.282 ("getting weird issues with the latest updates… my prompt disappears… 1000+ message session disappeared too").

The "Reconnecting..." banner with a counter that the user screenshotted is not just a UI quirk — it's the **observable symptom of the data being wiped**. Each cycle of the reconnect loop runs the data-loss code path, so the user sees their conversation shrink to nothing in real time and then the reconnect fires again.

## Root cause

v0.50.279 introduced `_clear_stale_stream_state()` in `api/routes.py` (#1525, commit `6bce34c`):

```python
def _clear_stale_stream_state(session) -> bool:
    stream_id = getattr(session, "active_stream_id", None)
    if not stream_id:
        return False
    with STREAMS_LOCK:
        stream_alive = stream_id in STREAMS
    if stream_alive:
        return False
    session.active_stream_id = None
    if hasattr(session, "pending_user_message"):
        session.pending_user_message = None
    # ... clears more pending_* fields ...
    try:
        session.save()    # ← CATASTROPHE
    except Exception:
        pass
    return True
```

This helper is called from two GET routes:

```python
# api/routes.py:1695  (/api/session)
s = get_session(sid, metadata_only=(not load_messages))
_clear_stale_stream_state(s)

# api/routes.py:1837  (/api/session/status)
_clear_stale_stream_state(get_session(sid, metadata_only=True))
```

Both paths load the session with `metadata_only=True`. `Session.load_metadata_only()` (api/models.py:421) synthesizes a stub object with `messages=[]` — that's its whole purpose: skip parsing the multi-megabyte messages array for fast sidebar/status responses.

Now look at `Session.save()` (api/models.py:368):

```python
meta['messages'] = self.messages       # ← writes whatever's on the stub: []
meta['tool_calls'] = self.tool_calls
payload = json.dumps({**meta, **extra}, ensure_ascii=False, indent=2)
tmp = self.path.with_suffix(...)
with open(tmp, 'w', encoding='utf-8') as f:
    f.write(payload); f.flush(); os.fsync(f.fileno())
os.replace(tmp, self.path)             # ← atomic overwrite
```

[...]

PR #4986, opened 2026-06-26
Fix manual /compress undone by reconciliation and .bak recovery (#4836)

### LINKED ISSUE #4836: Manual compression is undone by state.db reconciliation and .bak startup recovery
## Bug description

Manual `/compress` can successfully shrink a WebUI session, but the old
transcript can return through two independent recovery paths:

1. `state.db` reconciliation re-appends pre-compression rows.
2. Startup `.bak` recovery treats the intentional shrink as data loss and
   replaces the compressed live session with the larger backup.

This makes the UI appear to compress correctly and then jump back to the old
context size after the next message, reload, or WebUI restart.

## Environment

- Hermes WebUI: v0.51.531 (`673b364`)
- Hermes Agent: v0.17.0 / v2026.6.19
- macOS
- Manual compression from the WebUI

The latest WebUI `master` was checked on 2026-06-24:

- `_handle_session_compress()` replaces `s.messages` and
  `s.context_messages`, then saves, but does not mark an intentional
  truncation or update the post-compression prompt metric.
- `inspect_session_recovery_status()` still recommends restore whenever
  `bak_count > live_count`, without distinguishing intentional compression.

## Reproduction

1. Create a long WebUI session containing tool calls and reasoning metadata.
2. Run `/compress`.
3. Observe a successful result such as:

```text
Compressed: 614 -> 304 messages
Approx request size: ~231,772 -> ~74,156 tokens
```

4. Restart the WebUI.
5. The startup log reports:

```text
recover_session: restored <session>.json from .bak
(live=304 -> bak=657 messages)
```

6. The session returns to the old message count and old prompt-token metric.

Before restart, the same session could also reconstruct additional old
reasoning-only assistant rows from `state.db`, because many have empty visible
content and cannot be reliably excluded by content-based deduplication.

### 1. Manual compression does not mark the shrink as intentional

The handler rewrites the sidecar but leaves the append-only `state.db`
transcript intact. Without a compression boundary/watermark, reconciliation
can treat old rows as missing data.

This is the same general dual-source consistency class as #2914 and #3831,
but manual compression is another destructive/rewriting operation that needs
the same protection.

### 2. `.bak` recovery assumes every smaller live transcript is damaged

The backup safeguard from #1558 is valuable, but message-count comparison alone
cannot distinguish accidental loss from `/compress`, `/undo`, edit, or retry.
A valid compressed session is therefore automatically undone on restart.

### 3. Compression metadata remains stale

`last_prompt_tokens` can continue to show the pre-compression provider request,
so the UI displays the old percentage even when the saved context is smaller.
The cached metadata message count can also remain larger than the rewritten
message list.

## Expected behavior

- Manual compression should establish a persisted intentional-shrink boundary.
- While that boundary is active, model-context reconstruction should accept
  only state rows newer than the compression event.
- Startup recovery should not restore a larger `.bak` over a live session that
  explicitly records an intentional shrink.
- `last_prompt_tokens` and message-count metadata should be updated to the
  post-compression values.
- The boundary should be retired after the next successful turn is durably
  committed, preserving the existing #3831 behavior.

## Local validation

After implementing the above:

- actual compression: 614 -> 304 messages
- estimated request size: ~231,772 -> ~74,156 tokens
- state-db reconstruction: 304 -> 304 messages, 0 replayed rows
- after a real WebUI restart: still 304 messages / ~74,156 tokens
- WebUI compression/recovery tests: 61 passed
- combined state/recovery regression tests: 54 passed

Related issues: #1558, #2914, #3831.

hyl-ailab:
## Summary
- Persist `truncation_watermark` / `truncation_boundary` when manual `/compress` completes
- Mark compression as intentional (`compression_anchor_mode=manual`) and refresh `last_prompt_tokens`
- Skip startup `.bak` restore for intentional manual compress sessions
- Remove stale `.bak` after successful manual compress

## Why
Fixes #4836 — compressed context could reappear after reload or restart via state.db reconciliation or `.bak` recovery.

## Test plan
- [x] `pytest tests/test_issue4836_manual_compression_recovery.py tests/test_sprint46.py::test_session_compress_roundtrip -q`
- [x] `pytest tests/test_metadata_save_wipe_1558.py tests/test_issue2914_truncation_watermark.py -q`

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -19377,7 +19377,12 @@ def _handle_session_compress(handler, body):
             if _sanitize_messages_for_api(s.messages) != original_messages:
                 return bad(handler, "Session was modified during compression; please retry.", 409)
 
-            s.context_messages = copy.deepcopy(compressed)
+            from api.session_ops import _truncation_watermark_for
+            from api.streaming import _stamp_missing_message_timestamps
+
+            compressed_copy = copy.deepcopy(compressed)
+            _stamp_missing_message_timestamps(compressed_copy)
+            s.context_messages = compressed_copy
             s.active_stream_id = None
             s.pending_user_message = None
             s.pending_attachments = []
@@ -19392,7 +19397,19 @@ def _handle_session_compress(handler, body):
             s.compression_anchor_summary = _compact_summary_text(
                 summary_text or _compression_summary_from_messages(compressed) or ""
             )
+            # Persist an intentional-shrink boundary so append-only state.db
+            # reconciliation does not replay pre-compression rows (#4836).
+            compress_watermark = _truncation_watermark_for(compressed_copy)
+            s.truncation_watermark = compress_watermark
+            s.truncation_boundary = compress_watermark
+            s.compression_anchor_mode = "manual"
+            s.last_prompt_tokens = new_tokens
             s.save()
+            # Drop stale backups that would undo an intentional manual compress.
+            try:
+                s.path.with_suffix(".json.bak").unlink(missing_ok=True)
+            except OSError:
+                pass
 
         session_payload = redact_session_data(
             s.compact() | {
--- a/api/session_recovery.py
+++ b/api/session_recovery.py
@@ -65,6 +65,42 @@ def _msg_count(p: Path) -> int:
     return len(msgs) if isinstance(msgs, list) else -1
 
 
+def _session_records_intentional_compress_shrink(session_path: Path) -> bool:
+    """Return True when the live sidecar records an intentional context shrink.
+
+    Manual ``/compress`` keeps the visible transcript but replaces the
+    model-facing ``context_messages`` with a smaller compacted prefix. That
+    operation must not be treated as accidental data loss by the #1558
+    ``.bak`` safeguard or startup recovery (#4836).
+    """
+    try:
+        data = json.loads(session_path.read_text(encoding='utf-8'))
+    except (OSError, json.JSONDecodeError, ValueError):
+        return False
+    if not isinstance(data, dict):
+        return False
+
+    context_messages = data.get('context_messages')
+    messages = data.get('messages')
+    has_shorter_context = (
+        isinstance(context_messages, list)
+        and isinstance(messages, list)
+        and len(context_messages) < len(messages)
+    )
+    anchor_summary = str(data.get('compression_anchor_summary') or '').strip()
+    anchor_key = data.get('compression_anchor_message_key')
+    mode = str(data.get('compression_anchor_mode') or '').strip().lower()
+    watermark = data.get('truncation_watermark')
+
+    if mode == 'manual':
+        return True
+    if has_shorter_context and anchor_summary and anchor_key is not None:
+        return True
+    if has_shorter_context and watermark is not None:
+        return True
+    return False
+
+
 def inspect_session_recovery_status(session_path: Path) -> dict:
     """Return a status dict describing whether recovery is recommended.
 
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Thanks @hyl-ailab — #4836 is a real data-loss bug and the direction (persist the watermark on manual compress, gate startup recovery on intentional shrink) is right. But the dual gate (Codex + Opus, both verified on-branch) found a **CORE data-loss defect** in the guard, plus a broken regression test. Opus rates the guard NO-SHIP as-is, so this needs fixing before merge:

### 🔴 MUST-FIX 1 — the skip-restore guard is far broader than its contract → re-opens the #1558 .bak data-loss
The PR says the skip fires "only when `mode==manual` AND a `truncation_watermark` is recorded." But `_session_records_intentional_compress_shrink()` (`api/session_recovery.py:90-95`) is **three independent OR branches**:
```
if mode == 'manual':                       # branch 1 — NO watermark check
    ...
if has_shorter_context and watermark is not None:   # branch 3
```
Branch 1 skips recovery for any `manual` sidecar even without a watermark; branch 3 skips for **any** session that carries a `truncation_watermark` (set by retry/undo/Edit and re-advanced on every commit per #3831) and has shorter context. **Kill scenario (Opus traced it):** such a session later hits a genuine torn/partial write that shrinks the *visible* `messages` array so `.bak` has more messages than live (the exact #1558 failure) → `inspect_session_recovery_status` reaches `bak_count > live_count` → the guard returns `True` → `recommend: no_action` → **the `.bak` is never restored, the lost messages are gone.** That re-opens the precise data-loss the `.bak` safeguard exists to prevent, for auto-compressed and edited sessions.

**Fix (both gates agree):** make the guard strict and delete the fallback branches —
```
return mode == 'manual' and watermark is not None
```
(The torn-sidecar handling itself is fine: `json.loads` failure → `False` → restore proceeds, and `Session.save()` is atomic tmp+fsync+os.replace. The defect is breadth, not tearing.)

### 🔴 MUST-FIX 2 — a regression test is failing (release verification broken)
`tests/test_sprint46.py::test_session_compress_start_is_async_and_reuses_running_job` now FAILS: the async `/compress` path also stamps `context_messages` with timestamps, which the existing assertion doesn't expect. This path wasn't in the PR's listed test plan. Update the assertion like the sync test (compare role/content) and assert the new manual-mode/watermark/boundary fields for async compression too.

### 🟡 Test coverage
Add no-false-skip cases to `tests/test_issue4836_manual_compression_recovery.py`: (a) non-manual compressed context with `bak>live`, (b) `manual` mode WITHOUT a watermark, (c) an ordinary shrunken-live `.bak` recovery — each asserting the `.bak` IS still restored.

Once the guard is strict (`mode == 'manual' and watermark is not None`), the async test is fixed, and the no-false-skip cases are in, I'll re-run the full Codex+Opus+suite gate. This is close, but the guard breadth is a genuine data-loss regression on a crown-jewel path so it has to be tight first.

nesquena-hermes:
Shipped in **v0.51.710** 🎉 — thank you @hyl-ailab! Your diagnosis of the two recovery paths (state.db reconciliation + .bak startup recovery) undoing a manual `/compress` was exactly right, and the truncation-watermark + stale-.bak-delete approach is sound.

**One maintainer fix on top** (from the Codex+Opus gate): your `.bak`-recovery guard suppressed recovery whenever `compression_anchor_mode=="manual"` — but that flag is set once at compress and never cleared, so a manually-compressed session that *later* suffered a genuine #1558 data loss would have crash-recovery **permanently disabled**. I scoped the guard so it only suppresses the *pre-compression* backup (the one whose restore would undo your shrink), discriminated by the compaction marker (the same signal reconciliation uses) — a backup written *after* the compression that captures real data loss is still recovered. Added two regression tests (post-compression real loss recovers; a loss shrinking both messages and context still recovers).

Full gate: Codex SAFE + Opus SHIP (verified by an execution harness across the edge cases) + full suite green (10,739 passed; the 4 failures are known pre-existing flakes). Closes #4836 — reopen if a manual compression ever springs back after upgrading.

PR #4991, opened 2026-06-26
Use session index for compression continuation lookup

### LINKED ISSUE #4990: Pre-compression snapshot continuation lookup scans all session sidecars on open
## Summary

After the #4918 slices shipped, normal long-session paginated loads are much faster, but opening an old `pre_compression_snapshot` session can still take several seconds. Profiling shows this is not the bounded `state.db` tail-read path; it is the continuation-recovery lookup scanning the whole WebUI session sidecar directory on every open.

## Environment

- WebUI runtime observed locally: `v0.51.677-1-g948a289bb`
- Session count visible in `/api/sessions?limit=200`: 264
- `~/.hermes/state.db`: ~5.2 GB
- Affected shape: legacy WebUI session with `pre_compression_snapshot: true`

## Observed timings

On the same running instance:

- `/` warm load: ~3 ms
- `/api/sessions?limit=200` warm load: ~10 ms
- Normal long sessions after #4920:
  - 810-message session, `messages=1&msg_limit=30`: ~60 ms warm
  - 724-message session, `messages=1&msg_limit=30`: ~50 ms warm
  - 480-message session, `messages=1&msg_limit=30`: ~135 ms warm

But one old pre-compression snapshot session still shows:

- `messages=0&resolve_model=0`: ~3.5-5.7 s
- `messages=1&resolve_model=0&msg_limit=30`: ~3.4-5.7 s

The corresponding continuation session opens quickly:

- continuation `messages=0&resolve_model=0`: ~12 ms
- continuation `messages=1&resolve_model=0&msg_limit=30`: ~278 ms

This suggests the slow path is specific to opening the archived snapshot, not the actual continuation transcript.

## Profiling evidence

Local cProfile against `/api/session?session_id=<snapshot>&messages=0&resolve_model=0`:

```text
9869479 function calls in 5.707 seconds

5.702s api/routes.py:_pre_compression_continuation_session_id
5.700s api/routes.py:_child_rows
5.692s api/models.py:Session.load_metadata_only  (689 calls)
2.770s api/models.py:_read_metadata_json_prefix
2.728s api/models.py:_lookup_index_message_count
2.689s api/models.py:_index_message_count_map
```

The `messages=1&msg_limit=30` profile has the same dominant cost:

```text
5.572s api/routes.py:_pre_compression_continuation_session_id
5.570s api/routes.py:_child_rows
5.560s api/models.py:Session.load_metadata_only  (686 calls)
0.072s _limited_webui_messages_for_display_with_sidecar
0.071s merge_session_messages_append_only
```

SQLite itself was not the bottleneck for this session:

```text
SELECT COUNT(*), MAX(timestamp) FROM messages WHERE session_id = ?   ~0.5 ms
SELECT id, role, content, timestamp ... WHERE session_id = ?         ~18 ms cold / <1 ms warm
```

The session JSON parse itself was also small:

```text
snapshot sidecar: ~4.2 MB, 1365 messages
json.loads(read_text): ~22 ms
```

## Suspected cause

`_pre_compression_continuation_session_id()` only runs for `pre_compression_snapshot` sessions, but its `_child_rows()` helper scans `SESSION_DIR.glob("*.json")` and calls `Session.load_metadata_only(child_sid)` for every sidecar not already in memory. On long-lived installs, this turns opening a single old snapshot into an O(number of sidecar files) metadata scan.

The lookup only needs children keyed by `parent_session_id`, but it rebuilds that relationship by scanning every sidecar on every snapshot open.

## Expected behavior

Opening an archived pre-compression snapshot should either:

- resolve its visible continuation without scanning every sidecar, or
- use a cached/indexed `parent_session_id -> child session ids` map with safe invalidation, or
- store enough continuation metadata in the session index so the recovery hint can be resolved cheaply.

The endpoint should stay close to normal metadata-only session load cost, not several seconds.

## Relation to #4918

This appears separate from #4918 Problem 1. #4920 is working for normal long-session `state.db` tail loads: limited responses are much smaller than full loads and normal long sessions open quickly. The remaining slow case is an archived pre-compression snapshot recovery lookup.

franksong2702:
## Thinking Path

Issue #4990 isolated the slow session-open path to pre-compression snapshot recovery. Normal long-session loads already benefit from bounded state.db reads, but `_pre_compression_continuation_session_id()` still scanned every session sidecar when it needed to find the live continuation for a compression snapshot.

The safe fix is to use the maintained session index for parent/child continuation metadata only when that cache can prove it is complete against the persisted sidecar set. If the index is absent, malformed, or missing any persisted sidecar that is not already in memory, WebUI falls back to the old sidecar metadata scan so compression recovery stays result-identical.

## What Changed

- Reads continuation candidates from `_index.json` before falling back to sidecar metadata loads.
- Preserves the existing in-memory session check, profile-scope guard, bounded multihop traversal, backing-state validation, and safe session id validation.
- Proves index completeness by comparing safe indexed ids plus in-memory session ids against real `SESSION_DIR/*.json` sidecar ids before trusting the indexed result.
- Keeps `_child_rows_from_index()` side-effect-free: it uses local seen state and does not mutate the caller's `seen_ids` before index validity is known.
- Falls back to the legacy memory + sidecar resolver whenever the index cannot prove completeness, including valid-but-stale indexes that omit an on-disk continuation.
- Normalizes the final returned `session_id` with `_safe_first()` before URL/localStorage handoff.
- Adds regression coverage for the indexed fast path, valid-empty index fallback, non-vacuous stale index fallback with an older indexed candidate, stale multihop fallback through an indexed snapshot ancestor, index-backed multihop traversal, and cross-profile rejection.

## Why It Matters

Opening a pre-compression snapshot should not require loading metadata for hundreds of unrelated sessions on the happy path. At the same time, mobile reload / compression recovery must still self-heal from a stale session index by consulting durable sidecars instead of showing the hidden archived snapshot.

Fixes #4990.

## Contract Routing

- Task type: session recovery performance bugfix
- Touched areas: `api/routes.py`, `tests/test_mobile_reload_compression_recovery.py`
- Relevant docs reviewed: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `docs/rfcs/README.md`, `docs/rfcs/canonical-session-resolution.md`, `docs/rfcs/webui-run-state-consistency-contract.md`
- State layer: sidebar/session metadata cache (`sessions/_index.json`) used as a fast lookup over durable session sidecars
- Invariant: a pre-compression snapshot should resolve to the latest safe same-profile visible continuation without treating a stale cache as authoritative

## Verification

- `git diff --check`
- `.venv/bin/python -m py_compile api/routes.py tests/test_mobile_reload_compression_recovery.py`
- `./scripts/test.sh tests/test_mobile_reload_compression_recovery.py -q`
- `./scripts/test.sh tests/test_gateway_sync.py::test_session_load_exposes_continuation_for_pre_compression_snapshot tests/test_gateway_sync.py::test_session_load_exposes_multihop_continuation_for_repeated_compression -q`

## Risks / Follow-ups

- The fast path now scans sidecar filenames to prove `_index.json` completeness, but it still avoids loading metadata from every sidecar when the index is complete.
- If the index is incomplete, this intentionally pays the old O(all sidecars) fallback cost to preserve recovery correctness.
- No `CHANGELOG.md` entry in this PR; release notes are usually consolidated by release PRs in this repo.

## Model Used

OpenAI GPT-5, using local repo inspection, shell tests, GitHub CLI, and Paperclip aftercare automation.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -6787,20 +6787,30 @@ def _pre_compression_continuation_session_id(session) -> str | None:
             return None
         if not isinstance(entries, list):
             return None
+        try:
+            persisted_sidecar_ids = {
+                path.stem
+                for path in SESSION_DIR.glob("*.json")
+                if not path.name.startswith("_") and is_safe_session_id(path.stem)
+            }
+        except Exception:
+            return None
+        indexed_ids: set[str] = set()
+        row_seen_ids = set(seen_ids)
         rows = []
         for entry in entries:
             if not isinstance(entry, dict):
                 continue
             child_sid = _safe_first(entry.get("session_id"))
-            if (
-                not child_sid
-                or child_sid in seen_ids
-                or not is_safe_session_id(child_sid)
-                or not _safe_first(entry.get("parent_session_id"))
-            ):
+            if not child_sid or not is_safe_session_id(child_sid):
                 continue
-            seen_ids.add(child_sid)
+            indexed_ids.add(child_sid)
+            if child_sid in row_seen_ids or not _safe_first(entry.get("parent_session_id")):
+                continue
+            row_seen_ids.add(child_sid)
             rows.append(entry)
+        if persisted_sidecar_ids - indexed_ids - seen_ids:
+            return None
         return rows
 
     def _child_rows_from_sidecars(seen_ids: set[str]) -> list:
@@ -6880,14 +6890,9 @@ def _pre_compression_continuation_session_id(session) -> str | None:
 
     memory_seen_ids: set[str] = set()
     rows = _child_rows_from_memory(memory_seen_ids)
-    index_seen_ids = set(memory_seen_ids)
-    index_rows = _child_rows_from_index(index_seen_ids)
+    index_rows = _child_rows_from_index(memory_seen_ids)
     if index_rows is not None:
-        indexed_result = _resolve_from_rows(rows + index_rows)
-        if indexed_result:
-            return indexed_result
-        rows.extend(_child_rows_from_sidecars(set(memory_seen_ids)))
-        return _resolve_from_rows(rows)
+        return _resolve_from_rows(rows + index_rows)
 
     rows.extend(_child_rows_from_sidecars(memory_seen_ids))
     return _resolve_from_rows(rows)
--- a/api/routes.py
+++ b/api/routes.py
@@ -6872,19 +6872,24 @@ def _pre_compression_continuation_session_id(session) -> str | None:
                 ) or 0
             ),
         )
-        latest_sid = _row_value(latest, "session_id", None) or None
+        latest_sid = _safe_first(_row_value(latest, "session_id", None)) or None
         # Only hand the client a well-formed session id (it gets written to URL/localStorage).
         if latest_sid and not is_safe_session_id(latest_sid):
             return None
         return latest_sid
 
-    seen_ids: set[str] = set()
-    rows = _child_rows_from_memory(seen_ids)
-    index_rows = _child_rows_from_index(seen_ids)
+    memory_seen_ids: set[str] = set()
+    rows = _child_rows_from_memory(memory_seen_ids)
+    index_seen_ids = set(memory_seen_ids)
+    index_rows = _child_rows_from_index(index_seen_ids)
     if index_rows is not None:
-        return _resolve_from_rows(rows + index_rows)
+        indexed_result = _resolve_from_rows(rows + index_rows)
+        if indexed_result:
+            return indexed_result
+        rows.extend(_child_rows_from_sidecars(set(memory_seen_ids)))
+        return _resolve_from_rows(rows)
 
-    rows.extend(_child_rows_from_sidecars(seen_ids))
+    rows.extend(_child_rows_from_sidecars(memory_seen_ids))
     return _resolve_from_rows(rows)
 
 from api.workspace import (
--- a/api/routes.py
+++ b/api/routes.py
@@ -6763,9 +6763,8 @@ def _pre_compression_continuation_session_id(session) -> str | None:
     # snapshot's profile and reject any child that isn't profile-matched.
     snapshot_profile = getattr(session, "profile", None)
 
-    def _child_rows() -> list:
+    def _child_rows_from_memory(seen_ids: set[str]) -> list:
         rows = []
-        seen_ids = set()
         try:
             with LOCK:
                 memory_sessions = list(SESSIONS.values())
@@ -6777,6 +6776,35 @@ def _pre_compression_continuation_session_id(session) -> str | None:
                 rows.append(child)
         except Exception:
             pass
+        return rows
+
+    def _child_rows_from_index(seen_ids: set[str]) -> list | None:
+        if not SESSION_INDEX_FILE.exists():
+            return None
+        try:
+            entries = json.loads(SESSION_INDEX_FILE.read_text(encoding="utf-8"))
+        except Exception:
+            return None
+        if not isinstance(entries, list):
+            return None
+        rows = []
+        for entry in entries:
+            if not isinstance(entry, dict):
+                continue
+            child_sid = _safe_first(entry.get("session_id"))
+            if (
+                not child_sid
+                or child_sid in seen_ids
+                or not is_safe_session_id(child_sid)
+                or not _safe_first(entry.get("parent_session_id"))
+            ):
+                continue
+            seen_ids.add(child_sid)
+            rows.append(entry)
+        return rows
+
+    def _child_rows_from_sidecars(seen_ids: set[str]) -> list:
+        rows = []
         try:
             for path in SESSION_DIR.glob("*.json"):
                 if path.name.startswith("_"):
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Thanks @franksong2702 — the index fast-path is the right idea and the BFS/20-hop/profile-guard/`max`-by-timestamp logic is byte-for-byte the old `_child_rows()` (Opus confirmed via `_row_value` abstraction). But both gates independently landed on one CORE problem that breaks the result-identical bar:

### 🔴 MUST-FIX — `_child_rows_from_index()` doesn't detect its own incompleteness → silently drops a real continuation
The sidecar-scan fallback runs **only** when `_child_rows_from_index()` returns `None`, and it returns `None` on exactly three conditions (routes.py:6782-6789): index file **absent**, **JSON parse error**, or **top-level not a list**. It does **not** detect a stale/incomplete index. When `_index.json` is present and parseable, the code does `return _resolve_from_rows(rows + index_rows)` and **never scans sidecars**.

**Consequence (both gates, Codex reproduced):** a continuation child that exists as `SESSION_DIR/<child>.json` on disk but is **missing from `_index.json` AND not in the in-memory `SESSIONS`** is silently dropped. The old full scan would have found it. The frontend `loadSession()` then receives no `continuation_session_id` and renders/stores the **hidden archived pre-compression snapshot instead of the live visible child** — a user-facing wrong-transcript regression on the compression-recovery path. (`all_sessions()` self-heals the index on any prior sidebar load, so the window is a cold path — but it's real.)

**Fix:** before trusting the index result, prove completeness — compare indexed IDs ∪ in-memory `SESSIONS` IDs against the real `SESSION_DIR/*.json` sidecar set; if any persisted sidecar is missing from the index, return `None` and run the full `_child_rows_from_sidecars()` fallback.

### 🔴 MUST-FIX — `seen_ids` mutated before index validity is proven
`_child_rows_from_index()` mutates `seen_ids` during parsing; if it then bails to the fallback, the sidecar scan can skip indexed ancestor rows needed for multi-hop traversal. Keep parsing side-effect-free — use a local seen set and only merge into `seen_ids` once the index is accepted as complete (or reset before fallback).

### 🟡 Minor — timestamp-tie ordering (low severity)
On a `created_at`/`updated_at` **tie** between two candidate continuations, `max()` returns the first in iteration order, and index order ≠ glob order — so the chosen sid could differ from the old scan on an exact float tie. Unlikely with `time.time()`, but a stable secondary sort key (e.g. sid) would make it fully order-deterministic.

### Test gap
Add a non-vacuous **stale-index** test: write a continuation child sidecar that's omitted from `_index.json` and assert `_pre_compression_continuation_session_id()` still returns it; plus a multi-hop case where an indexed ancestor + a missing-sidecar descendant still resolve to the latest visible continuation. (The current +94 tests pass because none drives the stale-index path.)

The perf win is worth it — once the index path proves its own completeness (or falls back), keeps `seen_ids` side-effect-free, and has the stale-index test, I'll re-gate and ship. Full suite is green now, but that's masking the gap (no test exercises stale-index).

franksong2702:
Addressed the maintainer feedback in 6286d1b1c087443a5f9696c53f54d201e2b98420.

- `_child_rows_from_index()` now proves `_index.json` completeness against safe persisted sidecar ids plus in-memory session ids before trusting the indexed result.
- Index parsing no longer mutates the caller's `seen_ids`; fallback sidecar scans start from the memory-only seen set.
- Added stale-index regressions for an omitted newer continuation sidecar and for an indexed snapshot ancestor with an omitted newer descendant.

Local verification run:
- `git diff --check`
- `.venv/bin/python -m py_compile api/routes.py tests/test_mobile_reload_compression_recovery.py`
- `./scripts/test.sh tests/test_mobile_reload_compression_recovery.py -q`
- `./scripts/test.sh tests/test_gateway_sync.py::test_session_load_exposes_continuation_for_pre_compression_snapshot tests/test_gateway_sync.py::test_session_load_exposes_multihop_continuation_for_repeated_compression -q`

franksong2702:
Refreshed the branch against the latest `master` in 2a063dc40dd5c5fd032cc71618d317e1f4ee249d.

- Merged base `60f6779e470588ba71297dbcab2b7104dba71617`.
- Local verification passed: `git diff --check`, `.venv/bin/python scripts/ruff_lint.py --diff origin/master`, `./scripts/test.sh tests/test_mobile_reload_compression_recovery.py -q`, the two targeted continuation tests in `tests/test_gateway_sync.py`, and `./scripts/test.sh tests/test_session_index.py -q`.
- GitHub Actions restarted on the refreshed head and are pending.

nesquena-hermes:
Shipped in **v0.51.682** (Release YL, just deployed) — thanks @franksong2702! Fixes #4990. The compression-continuation lookup now uses the session index as a fast path with a membership-completeness guard (falls back to the full scan if any persisted continuation sidecar is missing from the index), so it's faster AND can't silently surface the archived snapshot instead of the live child. I added two Opus-gate hardenings on your branch: a session_id secondary max() tiebreaker (byte-identical index/scan on updated_at ties) + a doc comment on the membership-completeness guarantee — Co-authored-by credit to you. Gate: Codex SAFE + Opus SHIP IT + full suite 10703. Verified on prod. (Shipped via release branch since this PR is from your fork — closing as shipped.)

PR #4993, opened 2026-06-26
fix(models): make session-index rebuild thread ownership explicit (#3894)

### LINKED ISSUE #3894: Harden index-rebuild thread ownership handoff (low-severity follow-up to #3884)
## Summary

Low-severity latent concurrency cleanup identified during the Opus review of #3884 (shipped in v0.51.344, Release LH). **Not a live bug** — filing as a follow-up so it isn't lost.

## Context

#3884 made the background session-index rebuild thread active-session-dir-safe by pinning a target `(SESSION_DIR, SESSION_INDEX_FILE)` in `_SESSION_INDEX_REBUILD_THREAD_TARGET` and clearing it in the `finally` block only when the target still matches:

```python
finally:
    with _SESSION_INDEX_REBUILD_LOCK:
        if _SESSION_INDEX_REBUILD_THREAD_TARGET == (expected_session_dir, expected_index_file):
            _SESSION_INDEX_REBUILD_THREAD = None
            _SESSION_INDEX_REBUILD_THREAD_TARGET = None
```

## The latent window (Opus advisor, stage-344)

If a fast target-switch-then-restart happens — i.e. `_start_session_index_rebuild_thread()` runs for a *new* target while the old rebuild thread is still in flight — the old thread's `finally` sees a mismatched target and (correctly) does NOT clear the globals, but the new `_start_...` call has already overwritten `_SESSION_INDEX_REBUILD_THREAD` / `_TARGET` with the new thread. The net effect is benign today (the new thread owns the globals, the old thread just exits), but the bookkeeping relies on `_start_...` always running under the same `_SESSION_INDEX_REBUILD_LOCK`, and the ownership-transfer is implicit rather than explicit.

## Suggested cleanup

Make the ownership handoff explicit — e.g. have `_rebuild_session_index_background` capture the thread identity it was started as and only clear the globals when both the target AND `_SESSION_INDEX_REBUILD_THREAD is current_thread()` match. Purely defensive; no user-visible symptom is known.

## Severity

Low. No reproduction, no data-loss path. This is hardening of an implicit invariant in a daemon rebuild thread, surfaced by a thorough advisor read. Safe to pick up whenever.

rodboev:
## Thinking Path

- `#3894` is a bookkeeping hardening follow-up, not a user-visible bug report. The safest change is the one the issue already points at: make cleanup depend on explicit ownership, not just a matching target tuple.
- `api/models.py` already records both the current owner thread and the current target. The missing guard is simply that the worker finishing never checks whether it is still the registered owner before clearing those globals.
- This PR keeps the rebuild model intact. It only makes the owner handoff explicit and locks the latent same-target handoff window with a deterministic regression test.

## What Changed

- `api/models.py`: captures the current worker thread in `_rebuild_session_index_background(...)` and clears the rebuild globals only when the finishing worker is still the registered owner for the matching target tuple.
- `tests/test_session_index.py`: adds a deterministic same-target handoff regression that proves an older worker no longer clears a newer owner's globals.

## Why It Matters

This closes a subtle ownership window in the background rebuild bookkeeping without changing scheduling or user-visible behavior. The code becomes explicit about who is allowed to clear the shared globals, which makes future thread-hand-off work safer.

## Verification

```text
pytest tests/test_session_index.py -v --timeout=60
pytest tests/test_issue2863_session_index_prime.py -v --timeout=60
```

## Upstream

Closes #3894.

Follow-up to [#3884](https://github.com/nesquena/hermes-webui/pull/3884), which introduced the target-pinned rebuild ownership this PR hardens.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -200,6 +200,7 @@ def _session_dir_has_persisted_session_files() -> bool:
 
 def _rebuild_session_index_background(expected_session_dir: Path, expected_index_file: Path) -> None:
     global _SESSION_INDEX_REBUILD_THREAD, _SESSION_INDEX_REBUILD_THREAD_TARGET
+    current_thread = threading.current_thread()
     try:
         with _SESSION_INDEX_REBUILD_LOCK:
             if SESSION_DIR != expected_session_dir or SESSION_INDEX_FILE != expected_index_file:
@@ -213,7 +214,7 @@ def _rebuild_session_index_background(expected_session_dir: Path, expected_index
         logger.debug("Background session-index rebuild failed", exc_info=True)
     finally:
         with _SESSION_INDEX_REBUILD_LOCK:
-            if _SESSION_INDEX_REBUILD_THREAD_TARGET == (
+            if _SESSION_INDEX_REBUILD_THREAD is current_thread and _SESSION_INDEX_REBUILD_THREAD_TARGET == (
                 expected_session_dir,
                 expected_index_file,
             ):
--- a/tests/test_session_index.py
+++ b/tests/test_session_index.py
@@ -1666,3 +1666,30 @@ def test_background_index_rebuild_skips_after_session_dir_switch(tmp_path, monke
     assert not new_index_file.exists()
     rows = _read_index(original_index_file)
     assert [row["session_id"] for row in rows] == ["late_switch_sid"]
+
+
+def test_background_rebuild_old_thread_finally_preserves_new_same_target_owner(tmp_path, monkeypatch):
+    session_dir = tmp_path / "sessions"
+    session_dir.mkdir(exist_ok=True)
+    index_file = session_dir / "_index.json"
+    target = (session_dir, index_file)
+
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.680** (Release YJ, just deployed) — thanks @rodboev! Addresses #3894. The session-index rebuild finally-block now requires the finishing worker to still be the registered owner thread before clearing the bookkeeping globals, so a late older worker can't clobber a newer rebuild's state. Gate: Codex SAFE (under-lock, no TOCTOU, no stuck-owner leak), full suite 10693. Verified on prod.

nesquena-hermes:
Shipped in **v0.51.680** (explicit session-index rebuild thread ownership) — the fix is live on master and deployed to prod. Thanks @rodboev for the contribution. Closing as shipped (the release landed via a stage/release branch so this PR object didn't auto-close). Please reopen if you still see the issue after upgrading + a hard refresh.

PR #1375, opened 2026-04-30
fix(#1361): preserve reasoning, tool calls, and partial output on Stop/Cancel

bergeouss:
## Problem

Hitting **Stop** during a streaming response loses both the AI's already-streamed reasoning and tool output — paid tokens gone with no way to recover.

Three distinct data-loss paths:

**§A — Reasoning text lost on cancel**
`_reasoning_text` is accumulated in a thread-local variable inside `_run_agent_streaming`. On cancel, `cancel_stream()` never sees it — it goes out of scope when the thread is interrupted.

**§B — Tool calls lost on cancel**  
`_live_tool_calls` is also thread-local. `cancel_stream()` only captures `STREAM_PARTIAL_TEXT` — tool calls the user saw on screen disappear after refresh.

**§C — Reasoning-only streams produce no partial message**
When the entire output is reasoning (no visible tokens), the regex strip of thinking blocks returns empty string. The `if _stripped:` guard skips the append entirely — only `*Task cancelled.*` survives.

## Fix

Mirrors the same pattern as the existing `STREAM_PARTIAL_TEXT` shared dict:

1. **`STREAM_REASONING_TEXT`** — new shared dict in `api/config.py`, populated in `on_reasoning()` and the reasoning branch of `on_tool()`, read in `cancel_stream()`

2. **`STREAM_LIVE_TOOL_CALLS`** — new shared dict, populated on `tool.started` and `tool.completed`, read in `cancel_stream()`

3. **§C fix** — `cancel_stream()` now appends a partial assistant message when `_stripped` is empty but reasoning or tool calls exist, with the reasoning as a `reasoning` field and tool_calls as a `tool_calls` field

## Tests

8 new regression tests in `tests/test_issue1361_cancel_data_loss.py`:
- ✅ Cancel with reasoning-only → reasoning preserved
- ✅ Cancel with reasoning + partial tokens → both preserved
- ✅ Cancel without reasoning dict → works as before (no regression)
- ✅ Cancel with tool calls → tool_calls preserved
- ✅ Cancel with tools + text → both preserved
- ✅ Reasoning-only creates partial message (§C)
- ✅ Tools-only creates partial message (§C)
- ✅ No reasoning, no tools, no text → only cancel marker (no change)

Existing tests unaffected: #1298 (10/10) and #1195 (5/5) all pass.

Closes #1361

PR #5024, opened 2026-06-26
Gate reasoning_content replay for provider-facing history

Stacey2911:
## Problem

Historical assistant `reasoning_content` is currently replayed into subsequent model-facing conversation turns via `_sanitize_messages_for_api()`. For local/generic providers such as LM Studio, Ollama, llama.cpp, and generic OpenAI-compatible endpoints, this can inject stale internal reasoning into future turns, increasing prompt size and potentially affecting model behavior.

## Why this matters for local/generic providers

Local backends generally do not require historical `reasoning_content` replay to maintain conversation continuity. When preserved, this metadata can add unnecessary tokens to later requests and can carry stale internal reasoning from earlier turns into the provider-facing history.

The WebUI can still keep this metadata for session/display purposes; the issue is forwarding it to providers that do not use it.

## Compatibility concern

Some provider protocols may require historical reasoning replay for tool-use continuity. Stripping `reasoning_content` universally could break those integrations.

This patch adds a provider-facing replay policy instead of a blanket strip/preserve decision.

## Policy modes

The sanitizer reads `webui.reasoning_content_replay` with three modes:

- `auto` — default; strips for local/generic providers and preserves for explicitly known provider/protocol paths
- `strip` — always removes historical assistant `reasoning_content` from provider-facing history
- `preserve` — keeps current behavior and preserves `reasoning_content`

### Auto mode behavior

| Provider | Model ID pattern | Action |
|---|---|---|
| DeepSeek | any | Preserve; may require replay for thinking/tool-use continuity |
| Anthropic | starts with `claude` | Preserve; may require replay for structured reasoning/tool-use continuity |
| OpenAI | contains `gpt-5`, `o1-`, `o3-`, or `o4-` | Preserve; may require replay for reasoning/tool-use continuity |
| All others | any | Strip; local/generic backends do not need historical reasoning replay by default |

## Important notes

- Stored session/display reasoning metadata is unchanged.
- This only affects provider-facing sanitized history.
- Input message objects are not mutated.
- Existing sanitizer behavior is preserved, including orphaned tool dropping, OOB block scrubbing, partial handling, and recovered user logic.
- This is separate from PR #4978, which handles literal `[OUT-OF-BAND USER MESSAGE]` marker blocks in message `content`.

## Tests

- `tests/test_reasoning_content_replay.py`: 25 passed
- Existing sanitizer/regression coverage: 57 passed across orphaned tools, recovered context replay (#4283), partial bloat handling, cancel/partial preservation (#893), and PR #1375 partial tool-call sanitization

Total validation: 82 passed.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -18263,7 +18263,13 @@ def _handle_chat_sync(handler, body):
             result = agent.run_conversation(
                 user_message=workspace_ctx + msg,
                 system_message=workspace_system_msg,
-                conversation_history=_sanitize_messages_for_api(_previous_context_messages, cfg=get_config()),
+                conversation_history=_sanitize_messages_for_api(
+                    _previous_context_messages,
+                    cfg=get_config(),
+                    effective_model=_model,
+                    effective_provider=_provider,
+                    effective_base_url=_base_url,
+                ),
                 task_id=s.session_id,
                 persist_user_message=msg,
             )
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -3738,7 +3738,136 @@ def _is_reasoning_only_assistant_message(msg) -> bool:
     return _content_has_reasoning_only_parts(content)
 
 
-def _sanitize_messages_for_api(messages, *, cfg: dict = None):
+def _is_local_reasoning_replay_base_url(base_url: str | None) -> bool:
+    """Return True when a custom provider base URL confidently points at localhost."""
+    if not base_url:
+        return False
+    try:
+        from urllib.parse import urlsplit
+
+        raw = str(base_url or '').strip()
+        if not raw:
+            return False
+        parsed = urlsplit(raw)
+        if not parsed.hostname and '://' not in raw:
+            parsed = urlsplit(f"http://{raw}")
+        host = (parsed.hostname or '').strip().lower()
+    except Exception:
+        return False
+    return host in {'localhost', '127.0.0.1', '::1', 'localhost.localdomain'}
+
+
+def _should_strip_reasoning_content(
+    cfg: dict | None,
+    *,
+    mode: str | None = None,
+    effective_model: str | None = None,
+    effective_provider: str | None = None,
+    effective_base_url: str | None = None,
+) -> bool:
+    """Decide whether historical assistant reasoning_content should be stripped from model-facing history.
+
+    This is a provider/protocol decision, not a model-capability heuristic.
+    Local/generic backends (LM Studio, llama.cpp, Ollama, and custom localhost
+    OpenAI-compatible endpoints) do not require historical reasoning replay and
+    receive stale content when it's preserved. Unknown providers preserve by
+    default so replay does not break providers that require reasoning/tool-call
+    continuity.
+
+    Args:
+        cfg: Config dict from get_config(), expected to contain webui.reasoning_content_replay.
+        mode: Explicit override mode ("strip", "preserve", "auto"). If provided, bypasses config lookup.
+        effective_model: Runtime-resolved model for the current session/request.
+        effective_provider: Runtime-resolved provider for the current session/request.
+        effective_base_url: Runtime-resolved base URL for custom providers.
+
+    Returns:
+        True if reasoning_content should be stripped from sanitized output.
+        False if it should be preserved in sanitized output.
+    """
+    # Explicit mode override takes priority
+    if mode is not None:
+        return mode == "strip"
+
+    # Config lookup. Missing/invalid config preserves shipped behavior: do not
+    # strip reasoning_content unless config explicitly requests it or auto can
+    # identify a local/generic effective backend.
+    if cfg is None:
+        return False
+
+    webui_cfg = cfg.get("webui", {}) or {}
+    if not isinstance(webui_cfg, dict):
+        return False
+
+    replay_mode = webui_cfg.get("reasoning_content_replay")
+    if not isinstance(replay_mode, str):
+        return False
+
+    normalized = replay_mode.strip().lower()
+
+    if normalized == "preserve":
+        return False
+    if normalized == "strip":
+        return True
+    if normalized == "auto":
+        # Auto mode: prefer runtime-resolved provider/model because profile
+        # defaults may differ from a per-session/request override.
+        model_cfg = cfg.get("model", {}) or {}
+        if not isinstance(model_cfg, dict):
+            model_cfg = {}
+
+        provider_id = str(effective_provider or model_cfg.get("provider", "") or "").strip().lower()
+        model_id = str(
+            effective_model or model_cfg.get("default") or model_cfg.get("name") or ""
+        ).strip().lower()
+        base_url = effective_base_url or model_cfg.get("base_url")
+
+        if not provider_id:
+            return False
+
+        # Known providers that require historical reasoning_content replay:
+        # - DeepSeek thinking mode distinguishes normal history from tool-call reasoning chains
+        # - Anthropic Claude 4+/3.7+ uses structured reasoning in tool-use contexts
+        # - OpenAI GPT-5+/o-series requires reasoning replay for tool-use continuity
+        if provider_id == "deepseek":
+            return False  # preserve for DeepSeek
+
+        if provider_id == "anthropic" and model_id.startswith("claude"):
+            return False  # preserve for Claude models
+
+        if provider_id == "openai":
+            # Preserve for GPT-5+ and o-series (not GPT-4o, etc.)
+            # Use exact match + dash-prefixed suffix to avoid broad substring matches.
+            _is_reasoning_model = (
+                model_id == "gpt-5"
+                or model_id.startswith("gpt-5-")
+                or model_id in {"o1", "o3", "o4"}
+                or model_id.startswith(("o1-", "o3-", "o4-"))
+            )
+            if _is_reasoning_model:
+                return False  # preserve for reasoning-capable OpenAI models
+
+        if provider_id in {"lmstudio", "ollama", "llamacpp", "llama.cpp"}:
+            return True
+
+        if provider_id == "custom" or provider_id.startswith("custom:"):
+            return _is_local_reasoning_replay_base_url(base_url)
+
+        # Unknown/cloud providers preserve by default.
+        return False
+
+    # Unknown mode -- preserve default behavior.
+    return False
+
+
+def _sanitize_messages_for_api(
+    messages,
+    *,
+    cfg: dict = None,
+    effective_model: str | None = None,
+    effective_provider: str | None = None,
+    effective_base_url: str | None = None,
+):
     """Return a deep copy of messages with only API-safe fields.
 
     The webui stores extra metadata on messages (attachments, timestamp, _ts)
@@ -3804,6 +3933,18 @@ def _sanitize_messages_for_api(messages, *, cfg: dict = None):
                 # Orphaned tool result — skip to avoid 400 from strict providers.
                 continue
         sanitized = {k: v for k, v in msg.items() if k in _API_SAFE_MSG_KEYS}
+        # Provider-aware reasoning_content stripping from model-facing history.
+        # Historical assistant reasoning_content is stripped only when the user
+        # explicitly requests strip mode or auto mode identifies a local/generic
+        # effective backend.
+        if msg.get('role') == 'assistant' and 'reasoning_content' in sanitized:
+            if _should_strip_reasoning_content(
+                cfg,
+                effective_model=effective_model,
+                effective_provider=effective_provider,
+                effective_base_url=effective_base_url,
+            ):
+                del sanitized['reasoning_content']
         if is_recovered:
             sanitized['_recovered'] = True  # temporary marker — stripped before return
         if 'content' in sanitized:
@@ -7800,7 +7941,13 @@ def _run_agent_streaming(
             _run_conversation_kwargs = dict(
                 user_message=user_message,
                 system_message=workspace_system_msg,
-                conversation_history=_sanitize_messages_for_api(_previous_context_messages, cfg=_cfg),
+                conversation_history=_sanitize_messages_for_api(
+                    _previous_context_messages,
+                    cfg=_cfg,
+                    effective_model=resolved_model,
+                    effective_provider=resolved_provider,
+                    effective_base_url=resolved_base_url,
+                ),
                 task_id=session_id,
                 persist_user_message=msg_text,
             )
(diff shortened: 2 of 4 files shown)


Stacey2911:
Updated this branch after the automated review noted that bare OpenAI o-series model IDs like `o1` / `o3` were not covered by the auto-mode preserve policy.

The matcher now covers `o1`, `o1-*`, `o3`, `o3-*`, `o4`, `o4-*`, and `gpt-5` / `gpt-5-*`, while keeping GPT-4o and local/generic providers on the strip path. Added tests for the bare o-series IDs and reran the sanitizer regression suite (87 tests, all passing).

nesquena-hermes:
Pulled the branch and read `api/streaming.py` against `origin/master`, the new `tests/test_reasoning_content_replay.py`, and traced how `cfg["model"]` is actually populated in `api/config.py`. The strip/preserve/auto scaffolding and the immutability guarantee look right, but **auto mode reads a config key that doesn't exist in production**, so the documented preserve behavior for Claude and GPT-5/o-series silently does the opposite.

## The bug: `model.id` is not a real config key

In `_should_strip_reasoning_content` (`api/streaming.py:3624-3625`):

```python
provider_id = str(model_cfg.get("provider", "")).strip().lower()
model_id = str(model_cfg.get("id", "")).strip().lower()
```

The config `model` section has no `id` key. Every other reader in the codebase keys the model identifier off `default` (with legacy `name` as fallback) — see `api/config.py:2481`, `2842`, `3456`, `3596`, `3804`, and `api/streaming.py:380`:

```python
model_id = str(model_cfg.get("default") or model_cfg.get("name") or "").strip()
```

Because `model_cfg.get("id", "")` is always `""` in production, the auto-mode guards that depend on `model_id` never fire:

- `provider_id == "anthropic" and model_id.startswith("claude")` (`streaming.py:3634`) → `"".startswith("claude")` is `False` → **Claude history gets stripped**, even though the PR table says preserve.
- The OpenAI block at `streaming.py:3637-3648` checks `model_id == "gpt-5"` / `o1`/`o3`/`o4` etc. against `""` → no match → **GPT-5 and o-series get stripped** too.

Only the DeepSeek branch (`streaming.py:3631`, which checks provider only) preserves correctly. So in the shipped default (`auto`), the actual behavior is "strip for everything except DeepSeek" — not the matrix in the PR description.

The tests pass only because the fixtures hand-build the wrong shape (`tests/test_reasoning_content_replay.py:168` `{"provider": "anthropic", "id": "claude-sonnet-4"}`, `:181` `{"provider": "openai", "id": "gpt-5"}`). They assert against `id`, so they're green while production reads empty. This is exactly the kind of gap greptile's model-ID-pattern findings don't catch — the pattern is fine, the key feeding it is wrong.

### Fix sketch

```python
provider_id = str(model_cfg.get("provider", "")).strip().lower()
model_id = str(
    model_cfg.get("default") or model_cfg.get("name") or ""
).strip().lower()
```

and update the fixtures to use `default` instead of `id` so they actually exercise the production key.

## Second concern: `cfg["model"]` is the profile default, not the per-run model

Even after the key fix, there's a layer mismatch worth deciding on. At all three call sites — `streaming.py:7556`, `7936`, `8992` — the sanitizer gets `cfg=_cfg` from `get_config_for_profile_home(_profile_home)`. But the model the run actually executes is `resolved_model` / `resolved_provider` from `resolve_model_provider(...)` at `streaming.py:6980`, which honors the per-session dropdown selection (the `@provider:model` form). A session that overrides the profile default (e.g. profile default is a local Ollama model but the user picked Claude in the dropdown for this session) would be policy-keyed off the *profile default* provider, not the model the request is actually going to. For a feature whose whole point is "don't forward reasoning to providers that don't want it," keying off the wrong provider undercuts the guarantee.

If threading `resolved_provider`/`resolved_model` into the call is too invasive, the `mode` override param already on `_should_strip_reasoning_content` is a clean seam — the call site has both `_cfg` and the resolved pair in scope and could pass an explicit decision.

## Recommendation

Two-line key fix is mandatory before merge; otherwise the auto default ships inverted for Anthropic/OpenAI. The per-session-model concern is a design call for the maintainer — at minimum worth a code comment documenting that the policy keys off the profile default model, not the resolved run model. CI checks aren't showing on this PR yet (`statusCheckRollup` is empty); worth confirming the suite runs before merge given the sanitizer is on every provider-facing turn.

nesquena-hermes:
Thanks @Stacey2911 — the underlying concern is legitimate (replaying stale historical `reasoning_content` into model-facing turns bloats prompts and can confuse local/generic backends). But the dual gate (Codex + Opus, both reproduced against the real code) found the implementation currently does the **opposite of its intent in production**, plus a silent regression — so this can't ship as-is:

### 🔴 NO-SHIP 1 — the preserve-allowlist is non-functional in production (`api/streaming.py` auto path)
`auto` mode decides preserve-vs-strip by reading `cfg["model"]["id"]` (e.g. `model_id.startswith("claude")`). But production **never writes a `model.id` key** — so that check always fails, and **Claude / OpenAI / DeepSeek reasoning models get STRIPPED**, the exact opposite of the intended "preserve for these." The headline feature is inverted.

### 🔴 NO-SHIP 2 — the default silently regresses preserve→strip for existing users (`api/streaming.py:1431` + the fallbacks)
Shipped behavior is **preserve** (`_API_SAFE_MSG_KEYS` keeps `reasoning_content`). The new default is `auto`, and every no-config / missing-key / invalid-shape path returns `True` (strip) as a so-called "conservative default." Stripping is **not** conservative — preserve is the current behavior. So an existing user with default config on any non-allowlisted provider silently flips to stripping reasoning from model-facing history, with no opt-in. **Fix:** the default and all absent/invalid-config fallbacks must **preserve** (current behavior); require an explicit `webui.reasoning_content_replay: strip` (or a correctly-working `auto`) to strip.

### 🔴 NO-SHIP 3 — `auto` reads profile-default `cfg["model"]`, not the effective per-session provider/model (`api/streaming.py:3624`)
Streaming resolves `resolved_model`/`resolved_provider` (streaming.py:7068) but the sanitizer is called with the static `_cfg` (streaming.py:7644). A session on `custom:*` / OpenRouter / a different model than the profile default gets the wrong replay policy. **Fix:** pass the effective `resolved_provider`/`resolved_model` (and base_url) into the policy.

### 🟠 Related — custom/cloud OpenAI-compatible gateways get stripped
Because every non-allowlisted provider falls through to strip, a `custom:*`/OpenRouter-style **cloud** route to GPT/Claude/DeepSeek loses reasoning continuity. **Fix:** classify by actual local-server signals (provider name + local/private `base_url`), not "unknown ⇒ local"; preserve for non-local cloud/custom unless explicitly set to strip.

### 🟡 Tests mask the defects
The tests pass only by constructing a `cfg` shape (with `model.id`) that production never produces — so they validate the allowlist against a fabricated input and hide both #1 and #2. **Fix:** drive the policy through the real sanitizer call path with the effective resolved provider/model, and add a test asserting **default config = preserve** (no behavior change for existing users).

The direction is good and worth doing — but it needs: default = preserve, strip only on explicit opt-in or a *correctly-working* local-server detection (by `base_url`/provider, not a non-existent `model.id`), evaluated against the effective per-session provider, with tests through the real path. Happy to re-gate once it's backward-compatible and the classifier reads real production state.

Stacey2911:
Updated this branch to address the NO-SHIP review feedback.

Changes made:
- Removed the invalid `model.id` assumption; policy now uses effective provider/model/base URL where available and production config fallback keys only where needed.
- Restored backward-compatible preserve-by-default behavior for missing, invalid, or unknown config.
- Wired effective request/session provider context into sanitizer calls so session overrides are not classified from stale profile defaults.
- Limited `auto` stripping to confidently local/generic effective backends; unknown/cloud/custom providers preserve unless explicitly set to `strip`.
- Reworked tests to use production-realistic config shapes and effective-provider overrides.

Validation:
- `test_reasoning_content_replay.py`: 32 passed
- regression suite including sanitizer/native-image coverage: 96 passed
- total: 128 passed

nesquena-hermes:
Re-gated after the 05:49Z push (`32 passed` / `128 total`). Pulled the branch again and read the new `_should_strip_reasoning_content` + `_is_local_reasoning_replay_base_url` against `origin/master`, plus all four call sites and the reworked tests. **All three NO-SHIP items are resolved.** Walking them:

### ✅ NO-SHIP 1 (non-existent `model.id`) — fixed
`api/streaming.py:3657-3659` now keys off the production schema with the runtime value taking priority:

```python
provider_id = str(effective_provider or model_cfg.get("provider", "") or "").strip().lower()
model_id = str(
    effective_model or model_cfg.get("default") or model_cfg.get("name") or ""
).strip().lower()
```

That's the same `default → name` precedence every other reader uses. The Claude/GPT-5/o-series preserve guards (`streaming.py:3674-3686`) now actually fire. The fixtures were rebuilt to match — `tests/test_reasoning_content_replay.py:52` uses `{"provider": provider, "default": default}`, and the auto cases drive through `effective_provider`/`effective_model`, so they exercise the real key instead of a fabricated one.

### ✅ NO-SHIP 2 (default silently flipped preserve→strip) — fixed
Every absent/invalid path now returns `False` (preserve). `streaming.py:3633-3641`:

```python
if cfg is None:
    return False
webui_cfg = cfg.get("webui", {}) or {}
if not isinstance(webui_cfg, dict):
    return False
replay_mode = webui_cfg.get("reasoning_content_replay")
if not isinstance(replay_mode, str):
    return False
```

and unknown-mode falls through to `return False` at the bottom. Confirmed by `test_empty_webui_cfg_preserves_reasoning_content` (:81), `test_non_dict_webui_cfg_preserves_reasoning_content` (:91), and `test_unknown_mode_preserves_reasoning_content` (:101). Existing default-config installs keep current behavior — good.

### ✅ NO-SHIP 3 (profile-default vs per-session model) — fixed
All three streaming call sites now thread the resolved trio: `streaming.py:7705-7707`, `8091-8093`, `9153-9155` pass `effective_model=resolved_model`, `effective_provider=resolved_provider`, `effective_base_url=resolved_base_url`. The sync path in `api/routes.py:17193-17198` does the same with `_model`/`_provider`/`_base_url`. The earlier "🟠 custom/cloud OpenAI-compatible gateways" concern is also handled — `custom`/`custom:*` now classify by `_is_local_reasoning_replay_base_url(base_url)` (`streaming.py:3691-3692`), which only strips for `localhost`/`127.0.0.1`/`::1`, so a cloud OpenRouter-style custom route preserves.

### Two small non-blocking notes

1. **Dead constant.** `_DEFAULT_REASONING_CONTENT_REPLAY_MODE = "auto"` (`streaming.py:1431`) is unreferenced anywhere in `api/` or `tests/`, and its value is now misleading — the real unset-default is preserve (`False`), not `auto`. Greptile flagged this too. Suggest deleting it so a future reader doesn't assume `auto` is the fallback.

2. **Two unchanged sanitizer calls are correct.** For the record, `routes.py:18968` (compression token estimate) and `:19128` (the modification-detection comparison) intentionally pass no `cfg`, so they preserve — neither is model-facing, so leaving them untouched is right, not an oversight.

Direction and implementation now match the intent: preserve by default, strip only on explicit `strip` or a confidently-local `auto` backend, keyed off the effective per-session provider. Nice turnaround. The dead constant is the only cleanup worth doing before merge; CI still shows no `statusCheckRollup` on the PR, so worth confirming the suite runs green in CI (not just locally) given the sanitizer is on every provider-facing turn.

Stacey2911:
Removed the dead `_DEFAULT_REASONING_CONTENT_REPLAY_MODE` constant noted in review. The corrected behavior remains preserve-by-default for missing/invalid config, with stripping only for explicit `strip` or confidently local/generic `auto` paths.

Stacey2911:
Fork-side CI is now green on the PR commit.

I fast-forwarded `Stacey2911/hermes-webui:master` to the PR head commit `aeff2695`, so the fork workflows ran against the same SHA as this PR branch:

- Browser smoke: passed — https://github.com/Stacey2911/hermes-webui/actions/runs/28282485065
- Tests: passed — https://github.com/Stacey2911/hermes-webui/actions/runs/28282485050

Local validation also remains green: 128/128 tests passed.

nesquena-hermes:
Re-checked after the `aeff2695` push. Confirmed the one cleanup item from my last pass is resolved: `_DEFAULT_REASONING_CONTENT_REPLAY_MODE` is gone — `git grep` finds zero references in `api/` or `tests/` at the PR head, so there's no longer a misleading "auto" default constant for a future reader to trip over.

Re-read `_should_strip_reasoning_content` at the new head to make sure the removal didn't disturb the preserve-by-default contract. It's intact (`api/streaming.py:3628-3695`):

```python
if cfg is None:
    return False
webui_cfg = cfg.get("webui", {}) or {}
if not isinstance(webui_cfg, dict):
    return False
replay_mode = webui_cfg.get("reasoning_content_replay")
if not isinstance(replay_mode, str):
    return False
```

and the auto path keys off the runtime-resolved trio with the real production schema (`effective_provider` → `model_cfg["provider"]`, `effective_model` → `default`/`name`), with `custom`/`custom:*` deferring to `_is_local_reasoning_replay_base_url(base_url)` and unknown/cloud providers falling through to `return False` (preserve). That matches the matrix in the description: preserve by default, strip only on explicit `strip` or a confidently-local `auto` backend.

All three earlier NO-SHIP items stay resolved and the non-blocking dead-constant note is cleared. From a code-review standpoint this looks ready. The one thing still worth a maintainer's eyes before merge: `statusCheckRollup` on the PR is still empty — the green runs reported are on the contributor's fork (`Stacey2911/hermes-webui`) at the same SHA, which is good signal, but confirming the upstream CI actually executes on merge is worthwhile given this sanitizer runs on every provider-facing turn. No code blockers remain from my side.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (decision-ready)

Re-gated the converged head (`43a0a39e`) on a fresh rebase onto current master. This gates whether historical `reasoning_content` is replayed into provider-facing turns — it runs on **every** turn, so I re-ran the full authoritative gate on it as a crown-jewel streaming path.

**Convergence:** my prior review had cleared all code blockers; the newer commit is the final polish. Confirmed clean at the current head.

**Gate result (Codex reproduce + full suite):**
- **Preserve-by-default contract holds** — `reasoning_content` is stripped only on explicit `webui.reasoning_content_replay='strip'` or a confidently-local `auto` backend (`custom`/`custom:*` via `_is_local_reasoning_replay_base_url`); unknown/cloud providers fall through to preserve (`return False`).
- **No history corruption** — strip applies only to the sanitized copy (`streaming.py:3936-3948`), never persisted input history.
- **Streaming settlement contract intact** — the `error:""` terminal silent-failure sentinel is preserved (`streaming.py:8278-8281`), which was my specific concern on this path.
- Runtime-resolved provider/model/base-URL threaded into every streaming + legacy `/api/chat` path.
- **Codex: SAFE TO SHIP.** **Full suite: 11369 passed** (only the 2 pre-existing cron-isolation serial artifacts). 83 targeted tests (reasoning-replay + native-image + provider-auth-terminal-error).

**Disposition:** converged, gate-GREEN, `MERGEABLE`. Because this changes a **default of what's sent to the model on every provider-facing turn** (a provider-contract behavior on the streaming crown-jewel), I'm surfacing it for your merge call rather than shipping a provider-history default-change solo — even though it's engineering-clean. **My recommendation: SHIP** — it's a correctness/privacy improvement (stops leaking reasoning traces to local/generic backends that can't use them), preserve-by-default means zero change for cloud providers, and it's fully gated. Flipping to `gate-pass`. Thanks @Stacey2911.

nesquena-hermes:
Shipped in **v0.51.790**. Thanks @Stacey2911! 🎉 Reasoning-content replay is now gated — preserved by default for cloud providers, stripped only for confidently-local backends that can't use it. Settlement contract verified intact.

Issue #5001, opened 2026-06-26
WebUI stuck after upgrade — repeating 401 loop on /api/profile/active (session not re-established)


nesquena-hermes:
WebUI stuck after upgrade — repeating 401 loop on /api/profile/active (session not re-established)
## Symptom
After applying a WebUI update, the web app becomes stuck/unresponsive. Immediately after the upgrade, server logs show a repeating loop of HTTP 401 responses on `GET /api/profile/active`, suggesting the active profile/session is not re-established once the new version comes up. The reporter also suspected elevated CPU usage during the stuck state, but that was not confirmed.

## Platform
Observed inside the desktop app wrapper, but the reporter explicitly confirmed it is NOT desktop-specific — the underlying WebUI itself gets stuck, so it should reproduce in a normal desktop browser pointed at the same server after an upgrade.

## Suspected area
WebUI surface — session / active-profile handling on `GET /api/profile/active` immediately after a version upgrade. Worth checking whether this is a regression or a stale build, and whether it overlaps the recently addressed CPU-pinning / request-diagnostics thread work.

## Repro
1. Run a WebUI instance with an authenticated session.
2. Apply a WebUI update/upgrade (the server restarts).
3. Reload or return to the web app.
4. Observe: the app is stuck/unresponsive and server logs show repeated HTTP 401 on `GET /api/profile/active`.

---
_Filed automatically from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena-hermes:
Triage note (not a maintainer decision) — I traced the auth/session path on `master` to narrow where this 401 loop comes from. Short version: this is almost certainly a **lost server-side session table** on restart, not a frontend bug, and it only bites when WebUI auth is enabled.

## How `/api/profile/active` gets a 401

The endpoint isn't public. `check_auth` in `api/auth.py:564` only short-circuits for `PUBLIC_PATHS` / `/static/` (`auth.py:570`); `/api/profile/active` is neither, so every request runs through `verify_session` and returns the JSON 401 body on failure (`auth.py:576-582`):

```python
if parsed.path.startswith('/api/'):
    body = b'{"error":"Authentication required"}'
    handler.send_response(401)
```

[...]

PR #5026, opened 2026-06-27
fix(#5021): bound boot model 401 recovery

### LINKED ISSUE #5021: Boot auth guard should cover /api/models redirect during post-upgrade recovery
## Summary
PR #5018 fixed the main post-upgrade 401 loop from `GET /api/profile/active` by giving that bootstrap read a one-shot recovery redirect. The next boot step still has its own unauth redirect path through `populateModelDropdown()` -> `GET /api/models` -> `_redirectIfUnauth()`.

In the normal dead-session path this is harmless, because the active-profile recovery redirect returns before model boot runs. The remaining gap is narrower: if the login page bounces back without establishing a session, the active-profile helper clears its attempt marker on the fallback path, and a later `GET /api/models` 401 can reopen the redirect cycle through a different door.

## Current code path
- `static/boot.js`: `_resolveActiveProfileBootstrapState()` owns the one-shot redirect for `GET /api/profile/active`
- `static/ui.js`: `populateModelDropdown()` fetches `GET /api/models`
- `static/ui.js`: `_redirectIfUnauth()` still redirects immediately on any 401 with no boot-scoped one-shot guard

## Why this should be separate from #5018
#5018 already bounded the primary loop and is the right fix for #5001. This follow-up is about making the whole boot sequence share the same bounded redirect budget instead of guarding only the first endpoint.

## Suggested direction
Keep this boot-scoped, not global:
- introduce a shared boot unauth redirect budget or marker
- have both the active-profile bootstrap path and the model-dropdown bootstrap path consume that same one-shot guard
- avoid changing `_redirectIfUnauth()` semantics for non-boot callers unless there is a broader audit

## Result
A genuinely dead session should still end at the login screen. The improvement is that every boot-time auth failure path stays bounded to one clean redirect.

## References
- #5001
- #5018

rodboev:
## Thinking Path

- #5018 bounded the active-profile boot loop, but boot still continues straight into `populateModelDropdown()` and `/api/models` 401s still route through the old shared `_redirectIfUnauth()` helper.
- That shared helper is still right for non-boot callers like session-visit refreshes and uploads, so the remaining fix needs to stay inside the boot path instead of redefining the helper globally.
- The change shares one boot redirect budget between `/api/profile/active` and the boot-time `/api/models` fetch, then leaves every non-boot caller on the existing behavior.

## What Changed

- `static/boot.js`: add a shared boot-only unauth redirect budget, treat a persisted marker as already-consumed on the next page load, and keep the boot-time model hydration on that budget while leaving `window._ensureModelDropdownReady()` on the normal post-boot 401 path.
- `static/boot.js`: clear the cached boot dropdown promise on boot-time 401 handling, so a consumed boot budget cannot suppress the later generic `/api/models` retry path.
- `static/sessions.js`: route restored-session boot through the boot starter while `S._bootReady` is still false, without aliasing that boot work onto `window._modelDropdownReady`, so the saved-session boot path cannot self-resolve and skip `/api/models`.
- `static/ui.js`: let `populateModelDropdown()` accept an optional unauth redirect callback while defaulting to `_redirectIfUnauth()` for all existing callers.
- `tests/test_5021_boot_model_redirect_budget.py`: add a browserless regression harness for the login-bounce path, the model-only boot 401 path, and the public post-boot `_ensureModelDropdownReady()` retry path.
- `tests/test_session_metadata_fast_path.py` and `tests/test_issue4756_session_visit_model_refresh.py`: keep the existing boot and session-visit source contracts aligned with the split boot-vs-post-boot model starters.
- `tests/test_model_default_boot_precedence.py`: keep the boot-default source contract aligned with the new `_hydrateModelDropdown()` wrapper shape.
- `tests/test_5001_active_profile_recovery.py`: align the existing active-profile harness with the shared boot budget so the original #5001 recovery contract still runs against production code.

## Why It Matters

A genuinely dead session still ends at the login screen, but boot no longer has a second redirect door through `/api/models`. The existing shared 401 helper stays intact for the user-initiated flows that already depend on it.

## Verification

- `python -m pytest tests/test_5001_active_profile_recovery.py tests/test_5021_boot_model_redirect_budget.py tests/test_1038_pwa_auth_redirect.py::TestPWAAuthRedirect::test_ui_js_models_fetch_uses_redirect tests/test_provider_mismatch.py::TestCheckProviderMismatch::test_active_provider_stored_on_model_load tests/test_session_metadata_fast_path.py tests/test_issue4756_session_visit_model_refresh.py tests/test_model_default_boot_precedence.py -v --timeout=60`
- Full-suite CI context, not a required local check unless requested: `pytest tests/ -v --timeout=60`

## Upstream

Closes #5021.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -2355,6 +2355,7 @@ window._applyTitlebarProfileVisibility=_applyTitlebarProfileVisibility;
   // after the saved session is visible.
   const _redirectBootModelDropdownIfUnauth=(res)=>{
     if(!res||res.status!==401) return false;
+    window._modelDropdownReady=null;
     if(_bootActiveProfileUnauthRedirectBudget.isConsumed()) return true;
     if(_bootActiveProfileUnauthRedirectBudget.spendOnRedirect(sessionStorage)){
       _bootActiveProfileUnauthRedirectBudget.redirectToLogin(window.location.pathname+window.location.search);
--- a/static/boot.js
+++ b/static/boot.js
@@ -2265,7 +2265,9 @@ window._applyTitlebarProfileVisibility=_applyTitlebarProfileVisibility;
     let consumed=false;
     const readAttempted=(storage=sessionStorage)=>{
       try{
-        return storage&&storage.getItem?storage.getItem(markerKey)==='1':false;
+        const attempted=storage&&storage.getItem?storage.getItem(markerKey)==='1':false;
+        if(attempted) consumed=true;
+        return attempted;
       }catch(_){
         return false;
       }
--- a/static/boot.js
+++ b/static/boot.js
@@ -2417,6 +2417,7 @@ window._applyTitlebarProfileVisibility=_applyTitlebarProfileVisibility;
     return next;
   };
   window._modelDropdownReady=null;
+  window._startBootModelDropdown=_startBootModelDropdown;
   window._ensureModelDropdownReady=_startModelDropdown;
   setTimeout(()=>{
     try{Promise.resolve(_startBootModelDropdown()).catch(()=>{});}catch(_){}
--- a/static/boot.js
+++ b/static/boot.js
@@ -2361,9 +2361,9 @@ window._applyTitlebarProfileVisibility=_applyTitlebarProfileVisibility;
     }
     return true;
   };
-  const _hydrateBootModelDropdown=()=>populateModelDropdown({
+  const _hydrateModelDropdown=({redirectIfUnauth=null}={})=>populateModelDropdown({
     preferProfileDefaultOnFreshBoot:true,
-    redirectIfUnauth:_redirectBootModelDropdownIfUnauth,
+    ...(redirectIfUnauth?{redirectIfUnauth}:{}),
   }).then(()=>{
     const sessionModelState=S.session&&S.session.model
       ? {model:S.session.model,model_provider:S.session.model_provider||null}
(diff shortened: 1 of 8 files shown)


nesquena-hermes:
Read the diff against `origin/master` across `static/boot.js`, `static/sessions.js`, `static/ui.js` plus the four scenarios in `tests/test_5021_boot_model_redirect_budget.py`. The shared boot redirect budget is the right shape for #5021 and the boot-vs-post-boot starter split holds up.

## What the split actually guarantees

The key correctness property is the `window` rebinding at `static/boot.js:2423-2424`:

```js
window._startBootModelDropdown=_startBootModelDropdown;   // budget-aware
window._ensureModelDropdownReady=_startModelDropdown;     // normal _redirectIfUnauth
```

I checked every consumer of `_ensureModelDropdownReady` and they are all post-boot, user-initiated refreshes — `panels.js:5690`, `panels.js:9247`, and `ui.js:3040`. Those keep the unbounded `_redirectIfUnauth` path, which is correct: a 401 on a deliberate provider/settings refresh *should* bounce to login immediately. Only the two boot entry points (`boot.js:2426` startup kick and the `S.session` restore at `boot.js:2457`/`2480`) and the pre-boot branch in `loadSession` go through the budgeted callback. That matches the issue's "keep this boot-scoped, not global" guidance precisely.

## The profile-resolved-then-model-401 case is the real fix

The scenario #5021 was worried about is covered by `test_boot_model_401_consumes_budget_when_profile_load_succeeds`: profile resolves (`clearAttempted`), then the boot `/api/models` 401 spends the budget itself via `_redirectBootModelDropdownIfUnauth` (`boot.js:2358`), persists the marker, and on the *next* page load `readAttempted` sees `consumed=true` so it does not redirect a second time. The assertion `second["redirects"] == []` with `second["storage"] == {}` confirms the one-redirect ceiling survives a reload. Good.

## One thing worth a follow-up (non-blocking)

Greptile flagged the storage asymmetry and I agree it is the only soft spot. `_resolveActiveProfileBootstrapState` threads an injectable `markerStorage`, but `_redirectBootModelDropdownIfUnauth` reads `sessionStorage` directly:

```js
if(_bootActiveProfileUnauthRedirectBudget.spendOnRedirect(sessionStorage)){
```

In production both resolve to the same `sessionStorage`, and within a single page load the in-memory `consumed` flag dominates regardless of storage, so behavior is correct today. The latent risk is purely test-harness coupling: if a future test injects a `FakeStorage` into the profile path but the model callback still hits the real `sessionStorage`, the two sides would silently diverge. Cheap hardening would be to let `_redirectBootModelDropdownIfUnauth` accept an optional storage arg defaulting to `sessionStorage`, mirroring the profile path — not required for merge.

## Verification

CI is green across all 9 matrix legs plus browser-smoke and lint, and the test contracts in `test_issue4756_session_visit_model_refresh.py` / `test_session_metadata_fast_path.py` were updated to track the new starter split rather than weakened. No concerns on the logic — the budget correctly bounds boot to a single redirect across both `/api/profile/active` and the boot-time `/api/models` fetch.

nesquena-hermes:
Shipped in **v0.51.692** (Release YV, just deployed) — thanks @rodboev! Fixes #5021. The boot /api/models 401 now shares the same boot redirect budget as the active-profile read (#5018), so a single post-upgrade boot redirects at most once and can't loop on the model fetch. _redirectIfUnauth() is unchanged for non-boot callers (session-visit refresh, uploads), logged-out users still reach login, budget resets on a fresh boot. Gate: Codex SAFE + Opus SHIP IT + full suite 10750. Verified the shared budget is live on prod. This completes the #5001 auth-boot hardening chain (#5018 + #4971 + #5023 + this). (Shipped via release branch since this PR is from your fork — closing as shipped; closes #5021.)

PR #5027, opened 2026-06-27
fix: new session produces no output in Chat view

Gerkinfeltser:
## What

When creating a new session and sending a message, the Chat view shows no output (no thinking dots, no response) while the server terminal shows a traceback or HTML error.

### 1. 🟥 CRITICAL — `attachLiveStream` silently no-ops on null streamId (`static/messages.js`)
If `/api/chat/start` returns a 200 OK without a `stream_id` (e.g. the server returns an error but the fetch resolves), the guard `if(!activeSid||!streamId) return;` silently exits — no SSE stream is ever opened, the busy spinner runs forever, and the user sees nothing.  
**Fix:** Replaced the silent return with an error message rendered in the Chat view.

### 2. 🟧 HIGH — Empty model string sent to `/api/chat/start` (`static/messages.js`)
`_chatPayloadModel()` could return `''` when no model was selected. The server may accept this but return a response without `stream_id`, triggering bug #1.  
**Fix:** Added `window._defaultModel` as a fallback before returning empty string.

### 3. 🟧 HIGH — `newSession()` model resolution race (`static/sessions.js`)
If the model dropdown hasn't fully populated yet, the new session request body has no model.  
**Fix:** Final safety net uses `window._defaultModel` when all other resolution paths fail.

### 4. 🟨 MEDIUM — Better logging when STREAMS misses (`api/streaming.py`)
`_run_agent_streaming` silently returns if `STREAMS.get(stream_id)` is None, making the server terminal unhelpful.  
**Fix:** Added `logger.warning` so the missing stream is visible in server logs.

### 5. 🟨 MEDIUM — SSE error handler doesn't always clear busy state (`static/messages.js`)
Early-exit paths in the SSE `error` listener leave the UI in a stuck-busy state.  
**Fix:** Wrapped the error handler body in `try/finally` to always call `_setActivePaneIdleIfOwner()`.

## Testing
- Ran the existing test suite (tests timed out requiring browser setup, but all code changes are guarded by existing error-handling paths)
- Manual verification: diffs are minimal (23 insertions, 2 deletions across 4 files)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,4 @@ graphify-out/
 node_modules/
 package-lock.json
 docs/ui-ux/review-shots/
+.venv-test/
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -5809,6 +5809,7 @@ def _run_agent_streaming(
     """
     q = STREAMS.get(stream_id)
     if q is None:
+        logger.warning("Stream %s not found in STREAMS — agent worker exiting without producing events", stream_id)
         return
     register_active_run(
         stream_id,
--- a/static/messages.js
+++ b/static/messages.js
@@ -96,7 +96,7 @@ function _markActiveSessionViewedOnReturn() {
 }
 
 function _chatPayloadModel(){
-  return S.session&&S.session.model||($('modelSelect')&&$('modelSelect').value)||'';
+  return S.session&&S.session.model||($('modelSelect')&&$('modelSelect').value)||window._defaultModel||'';
 }
 
 function _chatPayloadModelProvider(model){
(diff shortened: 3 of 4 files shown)


nesquena-hermes:
Thanks @Gerkinfeltser — the **defensive error-rendering** in `attachLiveStream` (surfacing a failure in the Chat view instead of spinning forever on a null `streamId`) is a genuinely good hardening, and the `try/finally` that always calls `_setActivePaneIdleIfOwner()` on the SSE error path is a sound fix for a stuck busy-state. I'd like to take those.

Before this can go through, though, I want to pin the premise — a *general* "new session produces no output in Chat view" would be a crown-jewel-breaking regression that the test suite and every user would hit immediately, and chat is working in the current release, so this looks like a **specific local condition** rather than a universal break. Could you share the exact repro?

- What does `/api/chat/start` actually return in your failing case — a 200 with no `stream_id`, or a non-2xx? Can you paste the server-side traceback/HTML error the terminal shows? That backend error is the real root cause; the frontend changes here make the symptom visible but don't fix *why* the server returned no stream.
- Is a **default model configured** in your instance (Settings → the default model), and is a provider key present? My read is the trigger is "no resolvable model → server errors → 200-without-stream_id → the silent no-op," which is a real edge but a narrower bug than the title.

On the specific changes:
- #2/#3 (the `window._defaultModel` fallbacks): note `newSession()` already falls back to `window._defaultModel` (sessions.js:891-894) and `_chatPayloadModel` already chains `modelSelect.value` — so part of this may be redundant with the existing resolution chain. Worth confirming which path was actually empty in your repro rather than adding parallel fallbacks.
- The `.gitignore` `.venv-test/` entry is unrelated to the fix — fine, but flag it.

If you can attach the server traceback, that tells us whether the right fix is frontend-defensive-only (surface the error — which I'd happily take) or a backend fix to `/api/chat/start` so it never returns 200-without-stream_id in the first place. Holding the model-resolution changes pending that, since they touch the new-session path and partly overlap existing logic.

PR #5036, opened 2026-06-27
fix: completion notification preview uses settled final answer (#5035)

### LINKED ISSUE #5035: Completion browser notification preview uses stream start instead of settled final answer
## Summary

When browser/PWA completion notifications are enabled, the notification body often shows the **first** part of the assistant turn (early streamed tokens or interim text before tools settle), not the **final** answer the user sees after the turn completes.

## Problem

On the SSE `done` event, the WebUI sends a completion notification with a short preview string. That preview was derived from the live `assistantText` accumulator via `assistantText.slice(0, 100)`.

`assistantText` grows from the start of the stream and can include preamble, progress phrasing, or content that is later replaced or hidden when the settled transcript is rendered (tool boundaries, anchor final-answer projection, inline reasoning stripped from display text, etc.). Users who background the tab therefore get misleading notification previews.

Notification **timing** (after `done`, respecting background-tab / force-hidden rules from prior fixes) appears correct; the bug is the **preview source**.

## Expected behavior

The completion notification preview should match what the user would read as the settled assistant reply for that turn (same family of logic as transcript final-answer rendering), truncated for the OS notification limit.

## Proposed fix

1. Add a small helper (e.g. `_completionNotificationPreviewText`) that resolves preview text from the last settled assistant message in the `done` payload, preferring anchor settled final-answer projection when available, then `msgContent` / stripped content, with a cleaned live-stream fallback only if needed.
2. Wire the `done` handler to use that helper instead of slicing `assistantText`.
3. Ensure `lastAsst` is available when the completing session is not the active pane (background completion on another session).
4. Extend `tests/test_pwa_notification_controls.py` to assert the new wiring and absence of `assistantText.slice(0, 100)` on the completion path.

## Non-goals

- Changing when notifications fire (visibility / `forceHidden` / sound).
- Redesigning notification copy or title strings.
- Server-side SSE changes.

## Acceptance criteria

- [ ] Completion notification body reflects settled final answer for a typical tool-using turn, not stream preamble.
- [ ] Regression tests updated; `./scripts/test.sh` passes.
- [ ] CHANGELOG Unreleased entry documents user-visible fix.

## Ask

Confirm this matches observed behavior; happy to open a PR with the approach above.

claw-io:
## Thinking Path

Identified completion notification body built from `assistantText.slice(0, 100)` on SSE `done` — the live accumulator reflects stream start, not settled transcript. Added `_completionNotificationPreviewText()` using anchor final-answer projection / settled `lastAsst`, hoisted `lastAsst` for non-active-pane completions, updated regression tests.

## What Changed

- `static/messages.js`: helper + `done` handler wiring; `lastAsst` scope + fallback from `d.session.messages`
- `tests/test_pwa_notification_controls.py`: asserts settled preview path, rejects old slice
- `CHANGELOG.md`: Unreleased fix entry

## Why It Matters

Users who background the tab during a long tool-using turn were seeing notification previews from the opening streamed text, not the final reply.

## Verification

- `./scripts/test.sh` (full suite) — see CI / local run output
- Targeted earlier: `tests/test_pwa_notification_controls.py`, `tests/test_issue856_background_completion_unread.py`

## Risks / Follow-ups

- Preview still truncated to 100 chars with ellipsis; title unchanged
- Manual PWA notification check on a backgrounded tab with tool-heavy turn recommended

## Model Used

x-ai/grok-composer-2.5-fast (Hermes WebUI developer session)

Closes #5035

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -4652,6 +4652,7 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
         if(isActiveSession){
           S.activeStreamId=null;
         }
+        let lastAsst=null;
         if(isActiveSession){
           // Capture previous session totals BEFORE overwriting S.session with the new
           // cumulative values from the done event. prevIn/prevOut are the totals as of
@@ -4681,7 +4682,7 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
             window._compressionUi={...window._compressionUi, sessionId:d.session.session_id};
           }
           // Find the last assistant message once for both reasoning persistence and timestamp
-          const lastAsst=[...S.messages].reverse().find(m=>m.role==='assistant');
+          lastAsst=[...S.messages].reverse().find(m=>m.role==='assistant');
           // Persist reasoning trace for Worklog Thinking Cards; normal transcript
           // rendering keeps provider reasoning out of the final answer.
           if(reasoningText&&lastAsst&&!lastAsst.reasoning) lastAsst.reasoning=reasoningText;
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.696** — merged via release PR #5040 and deployed to prod. Thanks @claw-io! 🙌

Rebased onto current master and ran the full gate: Codex SAFE TO SHIP (verified the hoisted `lastAsst` doesn't change reasoning-assignment, the `d.session.messages` fallback only fires for non-active-pane completions so transcript render is unaffected, the helper handles null/malformed/array inputs, and the notification body is plain text), browser smoke CLEAN, full suite 10760 passed.

Closes #5035. Nice fix — the notification now previews the actual final answer instead of the start of the stream.

PR #5041, opened 2026-06-27
fix(#5000): relay mirrored gateway approvals after stream loss

### LINKED ISSUE #5000: bug(gateway/approval): approve/deny relay fails on every approval — "active run unavailable"
bug(gateway/approval): approve/deny relay fails on every approval — "active run unavailable"
## Symptom
In a gateway-backed WebUI session, clicking approve or deny on an approval prompt fails for every approval with the UI error: "Gateway approval could not be relayed because the active run is unavailable. Reopen the session or retry after it reconnects." Reported as happening on every approval (not an edge case), disruptive enough that the reporter is rolling back versions to find a stable one. Workaround: the /yolo command bypasses the broken approval prompt.

## Suspected area
Gateway approval relay path: api/routes.py /api/approval/respond (mapping the approval click back to a relayable gateway run) and the approval card in static/messages.js. The exact error text is the actionable-failure message added by the recent gateway-approval work (closed #4771 via #4917, shipped in v0.51.653). That change only made the relay failure visible/actionable; it did not address the underlying relay failing. This report is the underlying relay failing universally, so it reads as a fresh regression on a recent build.

## Repro
1. Use Hermes WebUI with a gateway-backed backend on a recent build.
2. Trigger any operation that requests approval.
3. Click approve or deny in the WebUI.
4. Observe the relay error and the run staying blocked; repeat and observe it on every approval.

## Notes
Reporter is still tracing which version broke it and which last worked. Decision point on filing: file fresh vs reopen #4771.

---
_Filed automatically from a Discord community report. Edit or close if mis-scoped._

rodboev:
## Thinking Path

- Gateway approval cards already survive on the WebUI side as mirrored pending entries, but `/api/approval/respond` still insists on walking back through `session.active_stream_id` to find the run.
- Current `origin/master` proves the narrow failure: if the stream pointer is gone, the handler returns `gateway_run_unavailable` even while the mirrored approval still carries the original `run_id`.
- The final fix uses that mirrored `run_id` before the true no-run 409 path, then settles the local approval lifecycle too, so the card does not come back after a successful remote relay.

## What Changed

- `api/route_approvals.py`: add a narrow mirrored-approval run lookup keyed by `approval_id`
- `api/routes.py`: fall back to the mirrored gateway approval's `run_id`, preserve the existing 409 containment, and settle the local approval queue after a successful relay
- `tests/test_gateway_approval_legacy_path.py`: add regressions for the lost-`active_stream_id`, still-relayable path and the benign gateway-mode no-pending path, including queue cleanup and agent wake-up assertions

## Why It Matters

Gateway-backed approval prompts should not dead-end just because the live stream pointer vanished after the card rendered. This keeps already-visible approval controls usable, while still clearing the mirrored approval state once the remote run resumes and preserving the existing 409 for approvals that truly have no run left to relay.

## Verification

`pytest tests/test_gateway_approval_legacy_path.py tests/test_issue4771_local_approval_regression.py tests/test_gateway_approval_runs_api.py -v --timeout=60`

Full-suite CI context, not a required local check unless explicitly requested: `pytest tests/ -v --timeout=60`.

## Upstream

Closes #5000.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/route_approvals.py
+++ b/api/route_approvals.py
@@ -184,6 +184,31 @@ def reconcile_gateway_pending_mirror_locked(session_key: str) -> tuple[dict | No
     return head, total, changed
 
 
+def _gateway_mirrored_pending_run_id(session_key: str, approval_id: str) -> str | None:
+    """Return the mirrored gateway approval run_id for a matching pending card.
+
+    Reconciles the mirror first so a live gateway head still survives a lost
+    `active_stream_id` pointer.
+    """
+    approval_id = str(approval_id or "").strip()
+    if not approval_id:
+        return None
+    with _lock:
+        reconcile_gateway_pending_mirror_locked(session_key)
+        queue = _pending.get(session_key)
+        if isinstance(queue, list):
+            entries = queue
+        elif queue:
+            entries = [queue]
+        else:
+            return None
+        for entry in entries:
+            if isinstance(entry, dict) and entry.get("approval_id") == approval_id and entry.get(_GATEWAY_MIRROR_FLAG):
+                run_id = str(entry.get("run_id") or "").strip()
+                return run_id or None
+    return None
+
+
 def submit_gateway_pending_mirror(session_key: str, approval: dict) -> None:
     """Mirror the live gateway head into WebUI polling state under a typed tag."""
     del approval  # mirror from the live gateway head under `_lock`, not from callback input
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.699** — merged via release PR #5044 and deployed to prod. Thanks @rodboev! 🙌

Full gate: Codex SAFE TO SHIP (verified the stream-alive path is unchanged, the stream-lost relay only uses the mirrored approval scoped to the same session + approval id with no cross-session misroute, missing-origin approvals still return `gateway_run_unavailable`, and auth/CSRF gates are intact in the shared POST path), ruff CLEAN, full suite 10766 passed.

Closes #5041 + #5000. Clean targeted fix — keeping the approval card actionable across stream loss is exactly the right call.

PR #5050, opened 2026-06-27
fix(#5045): add memory-section hover path titles

### LINKED ISSUE #5045: feat(memory): add hover path titles to memory section list entries
## Summary
The pinned memory-path row from #5025 solves the main discoverability problem, but the section list still doesn't expose the full on-disk path on hover.

## Background
In the post-merge review on PR #5025, `nesquena-hermes` explicitly called out the hover variant as a good small follow-up:

> The original request offered hover and/or a pinned header row; this implements the pinned row, which is the stronger of the two and what the triage note recommended. If you later want the hover variant too, the same `_memorySectionPath(key)` value can be attached as a `title=` on each entry in the section list without touching the backend.

The backend already provides all four path fields, and #5025 already added `_memorySectionPath(key)` on the frontend.

## Proposed scope
- Attach the `_memorySectionPath(key)` value as a `title` on each memory section entry in the section list.
- Keep this render-only, no backend work.
- Add a focused frontend regression so the hover path stays wired for `memory`, `user`, `soul`, and `project_context`.

## Why this is useful
This is a small UX polish item with almost no blast radius, and it rounds out the original request cleanly.

rodboev:
## Thinking Path

- PR [#5025](https://github.com/nesquena/hermes-webui/pull/5025) exposed the full on-disk path in the selected memory detail pane, but the left-rail section list still showed only icon and label text.
- Reusing `_memorySectionPath(key)` for the section buttons was the smallest way to surface those paths consistently.
- That helper also had a latent correctness bug: for any unknown key it fell back to `memory_path`, which was harmless in the detail-pane caller and wrong for the section list because `external_notes` would have advertised a false `MEMORY.md` hover title.
- This PR makes the helper explicit for the four file-backed sections, leaves non-file-backed sections title-free, and proves the rendered button behavior directly.

## What Changed

- `static/panels.js`: set `title` on file-backed memory section buttons in the left rail via `_memorySectionPath`, and tighten the helper so only the real file-backed keys resolve to paths.
- `tests/test_3866_project_context_memory_section.py`: extend the existing Node-backed memory harness to assert the rendered hover titles for `memory`, `user`, `soul`, and `project_context`, plus the empty hover case for `external_notes`.

## Why It Matters

Users can inspect the section list and see the exact backing file before opening a section. The same change also closes the `external_notes` mislabel gap, so the UI no longer risks showing a misleading `MEMORY.md` path for a section that has no file behind it.

## Verification

```bash
pytest tests/test_3866_project_context_memory_section.py -v --timeout=60
```

## Upstream

Closes #5045.

Attribution: maintainer follow-up on https://github.com/nesquena/hermes-webui/pull/5025 identified the section-list hover variant and the `external_notes` fallback gap it exposed.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/static/panels.js
+++ b/static/panels.js
@@ -4629,8 +4629,8 @@ function _memorySectionPath(key) {
   if (key === 'user') return _memoryData.user_path || '';
   if (key === 'soul') return _memoryData.soul_path || '';
   if (key === 'project_context') return _memoryData.project_context_path || '';
-  // Intentional default: the primary "memory" section remains the fallback.
-  return _memoryData.memory_path || '';
+  if (key === 'memory') return _memoryData.memory_path || '';
+  return '';
 }
 
 function _setMemoryHeaderButtons(mode) {
@@ -6483,6 +6483,8 @@ async function loadMemory(force) {
         el.className = 'side-menu-item';
         if (_currentMemorySection === s.key) el.classList.add('active');
         el.innerHTML = `${li(s.iconKey,16)}<span>${esc(_memorySectionLabel(s))}</span>`;
+        const sectionPath = _memorySectionPath(s.key);
+        if (sectionPath) el.title = sectionPath;
         el.onclick = () => openMemorySection(s.key, el);
         panel.appendChild(el);
       }
--- a/tests/test_3866_project_context_memory_section.py
+++ b/tests/test_3866_project_context_memory_section.py
@@ -141,6 +141,8 @@ def test_memory_panel_references_all_memory_path_fields():
     assert "_memoryData.user_path" in panels
     assert "_memoryData.soul_path" in panels
     assert "_memoryData.project_context_path" in panels
+    assert "const sectionPath = _memorySectionPath(s.key)" in panels
+    assert "if (sectionPath) el.title = sectionPath" in panels
 
 
 def _memory_render_blocks():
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Triage review (pulled the branch, read `static/panels.js` at PR HEAD against `origin/master`, and traced both callers of `_memorySectionPath`; not a maintainer decision). This is the clean hover follow-up to #5025, and the semantics change is more than cosmetic — it's a small correctness improvement that's worth calling out explicitly.

## The fallback removal is the right move, not just a cosmetic tweak

`master` returned `memory_path` for *any* unrecognized key:

```js
if (key === 'project_context') return _memoryData.project_context_path || '';
// Intentional default: the primary "memory" section remains the fallback.
return _memoryData.memory_path || '';
```

The PR tightens that to an explicit `memory` match with an empty-string default (`static/panels.js:4631-4632`):

```js
if (key === 'project_context') return _memoryData.project_context_path || '';
if (key === 'memory') return _memoryData.memory_path || '';
return '';
```

That matters because of the new section-list caller. `MEMORY_SECTIONS` (`panels.js:4589-4595`) includes `external_notes`, which has no on-disk path. Under the old fallback, calling `_memorySectionPath('external_notes')` would have returned the `memory` file's path, so the External Notes section button would have shown a misleading `title` pointing at `MEMORY.md`. The explicit-match + `''` default is what makes the new `if (sectionPath)` guard at `panels.js:6486-6487` actually suppress the title on non-file-backed sections:

```js
const sectionPath = _memorySectionPath(s.key);
if (sectionPath) el.title = sectionPath;
```

## Both callers verified safe

The other caller is `_renderMemoryDetail` at `panels.js:4739`. It's only reached for non-`external_notes` sections — `_renderMemoryDetail` returns early for `external_notes` (`panels.js:~4724`), so the detail-view `pathHtml` path never depended on the old fallback for that key. The four real sections (`memory`/`user`/`soul`/`project_context`) all still resolve their own `*_path` field, so the pinned-row behavior from #5025 is unchanged. Net effect of dropping the fallback: zero behavior change for the four file-backed sections, correct suppression for `external_notes`.

## Test coverage is appropriately targeted

`test_memory_section_list_renders_hover_path_titles` asserts the exact mapping including the empty case (`rendered["external_notes"] == ""`), which locks the corrected semantics rather than just the happy path. The source-string assertions added to `test_memory_panel_references_all_memory_path_fields` (`const sectionPath = _memorySectionPath(s.key)` / `if (sectionPath) el.title = sectionPath`) are brittle by nature (they match literal source), but that's consistent with how the rest of this test file pins the render wiring, so it's fine here.

Greptile's 5/5 is warranted. The one thing I'd note for the description: this also fixes a latent mislabel for `external_notes`, not only adds hover — worth a line in the PR body so the semantics change is documented for anyone reading the history later.

nesquena-hermes:
Shipped in **v0.51.700** — merged via release PR #5052 and deployed to prod. Thanks @rodboev! 🙌

Clean follow-up to #5025: the memory-section rail buttons now show their on-disk path on hover, reusing the existing `_memorySectionPath` helper. Codex SAFE TO SHIP (el.title is XSS-safe, #5025 detail-pane display unaffected by the unknown-key tightening), browser smoke CLEAN, full suite 10767 passed. Closes #5050 + #5045.

PR #5055, opened 2026-06-27
Serve dashboard plugin backends (plugin_api.py)

ai-ag2026:
## Problem

Dashboard plugins may declare an optional Python backend via the manifest `"api"` field (e.g. `plugin_api.py`) exposing a FastAPI `APIRouter` named `router`. Until now that backend was never mounted — `api/plugins.py` even notes *"plugin_api.py — optional backend API (not used in WebUI MVP)"*. As a result, a plugin whose built frontend fetches `/api/plugins/<name>/<route>` receives a **404** and renders a blank panel.

## Change

Serve the router's **GET** routes (read-only) under `/api/plugins/<name>/<route>`:

- `dispatch_plugin_api()` imports the plugin's `plugin_api` module (path-confined to the plugin's `dashboard/` dir), introspects `router.routes`, matches the GET route, coerces query params to the endpoint signature (handling stringized PEP 563 annotations and unwrapping FastAPI `Query()`/`Param()` default objects), and returns the endpoint result as JSON.
- A dispatch branch for `GET /api/plugins/<name>/<route>` is added right after the existing `/api/plugins` route.

### Safety / scope

- **Read-only**: only `GET` routes are served.
- Plugin name is validated; module import is path-confined (no traversal, no source exposure).
- Endpoint exceptions return `500` with a generic message.
- **FastAPI is an optional dependency**: if it is not importable, the plugin's router registers no routes and backends are skipped gracefully (the panel still loads, just without live data).

## Tests

Adds `tests/test_plugin_api_mount.py`: dependency-free tests for the dispatch/coercion logic (route matching, GET-only, query coercion, default fallback, name validation, error wrapping) plus a fastapi-gated test exercising real `APIRouter` introspection including `Query(...)` default unwrapping.

### DIFF
--- a/api/plugins.py
+++ b/api/plugins.py
@@ -183,3 +183,108 @@ def get_plugin_metadata() -> list[dict]:
             "hooks": [],
         })
     return plugins
+
+
+# ── Dashboard plugin backends (plugin_api.py) ──────────────────────────────
+# A dashboard plugin may declare an optional Python backend via manifest "api"
+# (e.g. "plugin_api.py") that exposes a FastAPI ``APIRouter`` named ``router``.
+# The WebUI serves its GET routes (read-only) under /api/plugins/<name>/<route>
+# so the plugin's built frontend can fetch its own data. Without this, such a
+# dashboard loads its assets but every data fetch 404s (blank panel).
+#
+# Requires FastAPI to be importable so the plugin's ``@router.get(...)`` routes
+# are registered; if it is not installed, plugin backends are skipped gracefully
+# (the panel still loads, just without live data).
+
+# plugin_name -> imported plugin_api module (or None if absent/failed)
+_PLUGIN_API_MODULES: dict[str, object] = {}
+
+
+def _load_plugin_api(plugin_name: str):
+    if plugin_name in _PLUGIN_API_MODULES:
+        return _PLUGIN_API_MODULES[plugin_name]
+    module = None
+    manifest = PLUGIN_MANIFESTS.get(plugin_name)
+    root = _PLUGIN_STATIC_ROOTS.get(plugin_name)
+    api_file = (manifest or {}).get("api")
+    if manifest and root and api_file:
+        try:
+            api_path = (root / api_file).resolve()
+            api_path.relative_to(root)  # reject traversal outside dashboard/
+            if api_path.is_file():
+                import importlib.util
+                import sys
+                spec = importlib.util.spec_from_file_location(f"_hermes_plugin_api_{plugin_name}", api_path)
+                module = importlib.util.module_from_spec(spec)
+                sys.modules[spec.name] = module
+                spec.loader.exec_module(module)
+        except Exception:
+            logger.exception("Failed to import plugin_api for plugin %s", plugin_name)
+            module = None
+    _PLUGIN_API_MODULES[plugin_name] = module
+    return module
+
+
+def _coerce_endpoint_kwargs(endpoint, query: dict) -> dict:
+    """Build call kwargs for a router endpoint from the query string.
+
+    Coerces present query values to the parameter's annotation, and unwraps a
+    FastAPI ``Query(...)``/``Param(...)`` default object to its plain default for
+    absent params (so the endpoint isn't called with the marker object).
+    """
+    import inspect
+    kwargs: dict = {}
+    for name, p in inspect.signature(endpoint).parameters.items():
+        if name in query and query[name]:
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Pulled this into a read-only worktree and read the full diff against `origin/master` plus the surrounding plugin-routing code. The mounting approach is sensible (read-only GET-only, path-confined import, graceful skip when FastAPI is absent), and `tests/test_plugin_api_mount.py` covers the dispatch/coercion core well. Three issues stand out, two of which match the Greptile flags — both are real when checked against the repo's own contract.

## 1. The new route skips the server-side enable gate (the important one)

Both existing plugin URL surfaces gate on `_dashboard_plugin_enabled()` before serving anything. Static assets, `api/routes.py:10371`:

```python
# Server-side enable-gate: a plugin disabled in Settings must have its
# entire URL surface shut off, not merely hidden in the UI.
if not _dashboard_plugin_enabled(plugin_name):
    return False  # 404 — disabled plugins serve nothing
```

The plugin page route at `api/routes.py:10398` does the same. The new branch this PR adds does not:

```python
if parsed.path.startswith("/api/plugins/"):
    from api.plugins import dispatch_plugin_api
    name, _, sub = parsed.path[len("/api/plugins/"):].partition("/")
    result = dispatch_plugin_api(name, sub, parse_qs(parsed.query))
```

So a plugin a user has explicitly toggled **off** in Settings still has a live backend reachable by name. The docstring on `_dashboard_plugin_enabled` (`api/routes.py:8537`) is explicit that the intent is "a disabled plugin's page + asset URLs are fully 404'd" — the API surface should join that set. Fix is one line, placed right after the `name, _, sub = ...` split:

```python
    if not _dashboard_plugin_enabled(name):
        return bad(handler, "Unknown plugin API route", 404)  # same shape as no-route
```

(Returning the same 404 as a missing route avoids leaking enabled/disabled state.)

## 2. Async endpoints return a coroutine into the JSON path

`dispatch_plugin_api` calls the endpoint synchronously:

```python
return (200, endpoint(**_coerce_endpoint_kwargs(endpoint, query)))
```

FastAPI dashboard handlers are very commonly `async def`. For those, `endpoint(...)` returns a coroutine object, which then flows straight into `j(handler, payload)` and either serializes as garbage or raises. The test suite only exercises sync `def summary(...)` handlers, so this path is uncovered. Either detect and run coroutines (`inspect.iscoroutine(result)` → `asyncio.run(...)`), or document that plugin GET handlers served this way must be sync — and add a test for whichever you pick.

## 3. FastAPI parameter validation is bypassed

The manual `_coerce_endpoint_kwargs` unwraps `Query(...)` to its plain default but drops the constraints. Your own fastapi-gated test declares `Query(24, ge=1, le=720)`, yet `?hours=999999` would sail through since `ge`/`le` are never applied. For read-only endpoints that is lower severity, but worth a one-line note in the module docstring so plugin authors don't assume their declared bounds are enforced.

## Verification

Per cron policy I did not execute the PR's code. #1 is verifiable by reading `api/routes.py:10371` and `:10398` and confirming the new branch lacks the equivalent guard. #2/#3 would each warrant a new case in `tests/test_plugin_api_mount.py` (an `async def` route for #2; an out-of-range query value for #3). #1 is the blocker; #2 will bite the first real `async` plugin backend.

ai-ag2026:
Closing this — on closer inspection the route can't achieve its stated goal in the current plugin model.

Dashboard plugins are rendered in a sandboxed iframe (`panels.js`: `iframe.sandbox = "allow-scripts allow-forms allow-popups"`, and the page is also served with `Content-Security-Policy: sandbox ...`). Without `allow-same-origin` the iframe has an **opaque origin**, so any request from the plugin frontend — `<script src>` for `/dashboard-plugins/...` assets **and** `fetch('/api/plugins/<name>/<route>')` — carries **no session cookie**. The mounted route therefore returns 401 to the very plugin frontends it was meant to serve (it turns a 404 into a 401; the panel stays blank).

The pattern that actually works for sandboxed dashboard plugins is **server-side data injection** via the existing `inline_payload` mechanism (the bundle reads an injected `window.__...__` global instead of fetching). That's the route we took for our dashboards.

A genuinely useful follow-up would be a *sandbox-aware* backend-access mechanism (e.g. inject a short-lived per-plugin token into the iframe payload and accept it as a request **header**, since headers — unlike cookies — are sent cross-origin). That's a larger, security-sensitive feature, out of scope for this small PR. Closing for now; happy to revisit as that design.

Issue #1879, opened 2026-05-08
Bug: Gateway liveness detection fails across separate containers without shared PID namespace


qxxaa:
## Summary

The gateway liveness check in `get_status()` relies on `fcntl` file locking and
`os.kill()` to determine whether the gateway process is alive. Both mechanisms
require the caller to be in the same PID namespace as the gateway process. In a
typical multi-container deployment (gateway in one container, WebUI in another),
this makes cross-container liveness detection impossible without the compose
`pid: "service:hermes-agent"` workaround.

## Environment

- hermes-webui running in a separate container from hermes-agent
- No `pid: "service:hermes-agent"` in compose configuration
- Shared volume between containers for the Hermes data directory

## Steps to Reproduce

1. Run `hermes-agent` and `hermes-webui` as separate containers without
   `pid: "service:hermes-agent"` in the WebUI service definition
2. Open the WebUI dashboard — Gateway Status shows "Gateway not running"
3. Banner fires: *"Hermes agent is not responding. Gateway heartbeat failed."*

## Root Cause

`get_running_pid()` in `gateway/status.py` uses two process-local mechanisms to
verify liveness:

1. **`fcntl.flock(LOCK_EX | LOCK_NB)`** on `gateway.lock` — if the lock cannot
   be acquired, the gateway is considered alive. Across container boundaries the
   lock always appears unheld, since `fcntl` locks are not visible outside the
   process that holds them when in a separate PID/kernel namespace context.

2. **`os.kill(pid, 0)`** — existence check on the recorded PID. PIDs are
   namespace-scoped; the gateway's PID 7 does not exist in the WebUI container's
   view of the process table without shared PID namespace.

The existing `GATEWAY_HEALTH_URL` HTTP probe handles this case but is marked
deprecated in the source and requires additional compose configuration to enable.

## Expected Behaviour

The gateway should be detectable as running from the WebUI container without
requiring PID namespace sharing or additional environment variables.

## Proposed Fix

Replace (or supplement) the fcntl/PID check with a **`gateway_state.json`
freshness check**. The gateway already writes an `updated_at` timestamp to
`gateway_state.json` on every tick. A recent timestamp with `gateway_state: "running"`
is a reliable cross-container liveness signal that requires only a shared volume:

```python
from datetime import datetime, timezone

GATEWAY_FRESHNESS_THRESHOLD_S = 120  # two cron ticks

runtime = read_runtime_status()
if (
    runtime
[...]

Issue #716, opened 2026-04-19
Feature request: Hermes agent heartbeat monitoring with alerts


githb-ac:
This feature is needed by users running multi-agent Hermes swarms on self-hosted VPS servers. Please add a heartbeat monitor that pings the Hermes agent process at regular intervals (e.g. every 30 seconds) and confirms it is alive and responding. If the agent stops responding, a prominent alert should appear in the UI immediately. This is distinct from CPU/RAM monitoring — a server can show healthy resource usage while the Hermes process itself is hung or crashed. A Python library well suited for this is [psutil](https://github.com/giampaolo/psutil), which is cross-platform, lightweight, and already compatible with the Hermes Web UI Python stack.

nesquena-hermes:
Good feature request — this is distinct from the CPU/RAM health panel in #693, which monitors server resources. A heartbeat monitor specifically checks whether the Hermes agent *process* is alive and responding, which can diverge from resource usage when a process is hung.

**Suggested implementation shape:**

A lightweight approach that fits the existing stack:

1. **Backend** — a `GET /api/health/agent` endpoint that tries to reach the agent process (e.g., checks the gateway process is running and can accept a trivial ping). Returns `{ "alive": true/false, "checked_at": <timestamp> }`. `psutil` is a reasonable fit for checking process liveness without shelling out.

2. **Frontend polling** — the UI polls this endpoint every 30 seconds (configurable). If `alive: false`, show a prominent banner or status indicator (distinct from the existing connection-lost toast, which handles WebSocket drops).

3. **Alert visibility** — the alert should persist until the agent comes back up, not auto-dismiss. A banner at the top of the chat area (above the input) is more visible than a sidebar indicator.

**Out of scope for a first pass:** automatic restart, notification webhooks, or process supervision — those belong in the gateway/agent layer, not the WebUI.

Related: #693 (live VPS health panel) may want a shared polling infrastructure if both features land. Worth coordinating.

githb-ac:
Following up on the closed status of this ticket the comments on this issue don't appear to confirm that this was actually implemented. The last response from the team notes it was logged/planned, but there's no follow-up confirming a PR was merged or that the feature shipped in a release. Could the team clarify whether this was intentionally closed as completed, or whether it was closed in error and should be reopened?

PR #5065, opened 2026-06-27
fix(#5030): read gateway runtime status from gateway_state.json

### LINKED ISSUE #5030: agent_health.py reads wrong file in cross-container setup, causes false "Gateway not running"
Environment: Multi-container Docker (hermes-agent + hermes-webui separate containers), shared HERMES_HOME volume.

Bug location: api/agent_health.py:221

Root cause: _read_gateway_runtime_status() passes pid_path (path to gateway.pid) as a positional argument to gateway.status.read_runtime_status(). Since read_runtime_status(path) uses the provided path to read a JSON file, it reads gateway.pid instead of gateway_state.json. The gateway.pid file lacks gateway_state and updated_at fields, so the health check falls through to alive: False → frontend shows "Gateway not running".

Fix: Change line 221 from:
return read_runtime_status(pid_path)

to:
return read_runtime_status(pid_path.with_name("gateway_state.json"))

Ref: agent_health.py lines 214–233, the third except TypeError block (lines 223–232) already has the correct logic using pid_path.with_name(runtime_status_file) but is unreachable because line 221 succeeds without error.

rodboev:
## Thinking Path

- Cross-container gateway health already trusts a fresh `gateway_state.json`, but `_read_gateway_runtime_status()` still hands the agent-side `read_runtime_status(path)` contract the sibling `gateway.pid` path after the keyword compatibility call fails.
- That makes the health path parse PID metadata as runtime metadata, so `gateway_state` and `updated_at` never reach the freshness check and the UI reports the gateway as down.
- The fix keeps the compatibility structure, routes the positional fallback to `gateway_state.json`, and locks the dashboard-facing payload with a focused cross-container regression.

## What Changed

- `api/agent_health.py`: switch the positional `read_runtime_status(path)` fallback from `gateway.pid` to the sibling runtime-status file while preserving the existing keyword and no-argument compatibility paths.
- `tests/test_issue1879_cross_container_gateway_liveness.py`: add a regression that models the real `path`-only gateway status contract and proves `build_agent_health_payload()` now reports `cross_container_freshness` instead of `gateway_not_running`.

## Why It Matters

Multi-container Docker installs can share `HERMES_HOME` without sharing a PID namespace. In that layout, `gateway_state.json` is the authoritative liveness signal, so reading `gateway.pid` as runtime JSON creates a false `Gateway not running` banner even while the gateway is healthy.

## Verification

```bash
pytest tests/test_issue1879_cross_container_gateway_liveness.py -v --timeout=60
```

## Upstream

Closes #5030.

Attribution: maintainer confirmation on https://github.com/nesquena/hermes-webui/issues/5030#issuecomment-4815057618 pinned the unreachable fallback and requested the regression test shape.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/agent_health.py
+++ b/api/agent_health.py
@@ -217,17 +217,18 @@ def _read_gateway_runtime_status(gateway_status: Any, pid_path: Path | None) -> 
         try:
             return read_runtime_status(pid_path=pid_path)
         except TypeError:
+            runtime_status_file = str(
+                getattr(gateway_status, "_RUNTIME_STATUS_FILE", _GATEWAY_RUNTIME_STATUS_FILE)
+            )
+            runtime_status_path = pid_path.with_name(runtime_status_file)
             try:
-                return read_runtime_status(pid_path)
+                return read_runtime_status(runtime_status_path)
             except TypeError:
                 if getattr(gateway_status, "__name__", "") == "gateway.status" or hasattr(
                     gateway_status,
                     "_read_json_file",
                 ):
-                    runtime_status_file = str(
-                        getattr(gateway_status, "_RUNTIME_STATUS_FILE", _GATEWAY_RUNTIME_STATUS_FILE)
-                    )
-                    runtime_status = _read_runtime_status_path(pid_path.with_name(runtime_status_file))
+                    runtime_status = _read_runtime_status_path(runtime_status_path)
                     if runtime_status is not None:
                         return runtime_status
     return read_runtime_status()
--- a/tests/test_issue1879_cross_container_gateway_liveness.py
+++ b/tests/test_issue1879_cross_container_gateway_liveness.py
@@ -28,6 +28,7 @@ These tests pin every behavior the fix promises:
 
 from __future__ import annotations
 
+import json
 from datetime import datetime, timedelta, timezone
 
 import pytest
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.706** 🎉 — thank you @rodboev!

Since this PR is from your fork, I shipped it via a maintainer release branch (push-to-fork-branch can't update the PR), preserving your authorship with a `Co-authored-by` trailer and crediting you in the CHANGELOG.

Clean root-cause fix: after the keyword-form `read_runtime_status(pid_path=…)` fails against an older agent signature, the positional fallback was being handed the `gateway.pid` path instead of the runtime-status file, so the reader parsed PID metadata as runtime metadata and `gateway_state`/`updated_at` never resolved — breaking cross-container freshness. Now the fallback passes the correct `gateway_state.json` sibling path (`pid_path.with_name(_RUNTIME_STATUS_FILE)`), defaulting to `gateway_state.json` when the agent doesn't expose the filename. Keyword happy-path and normal PID-based health unchanged.

Full gate: Codex SAFE TO SHIP (cross-referenced the real `gateway/status.py` to confirm the filename), your new `test_cross_container_runtime_status_reads_sibling_runtime_file` proven non-vacuous (fails on master without the fix, passes with it), full suite green (10,711 passed; the 4 failures are pre-existing order/env flakes unrelated to agent_health), #716 PID-health 27/27. Closes #5030 (and #1879). Reopen if cross-container health still misreports after upgrading.

Issue #4749, opened 2026-06-23
bug(steer): "Steer unavailable — draft restored" gives no reason + steer-mode dead-ends the composer ("can't interrupt / won't let me type")


nesquena-hermes:
## Summary

A user reports that **Steer** fails with the toast **"Steer unavailable — draft restored"**, and that while the agent is working they **"cannot interrupt … won't let me type."** The two symptoms are linked: when steer is persistently unavailable in *steer* busy-input mode, the user is left with no working busy action and the composer feels frozen.

> getting this error when trying to use "steer": **Steer unavailable — draft restored**. Also i cannot interrupt while it's working won't let me type

— @boriken72, Discord #webui (2026-06-23)

## Facet A — "Steer unavailable" is unactionable (root cause undiagnosable)

The toast is the **deliberate graceful-fallback copy** introduced in PR #4577 / v0.51.577 (`cmd_steer_fallback` / `busy_steer_fallback`, `static/i18n.js:297,300`). It fires from `static/commands.js:_trySteer()` (~L1282) whenever `POST /api/chat/steer` returns `accepted:false`.

The backend `_handle_chat_steer()` (`api/streaming.py:8970`) returns `accepted:false` with a `fallback` reason for any of:

| `fallback` value | Cause |
|---|---|
| `no_cached_agent` | session has no agent in `SESSION_AGENT_CACHE` |
| `agent_lacks_steer` | older hermes-agent without `steer()` |
| `session_not_found` | session lookup `KeyError` |
| `not_running` | no `active_stream_id` on the session |
| `stream_dead` | `active_stream_id` is stale (not in `STREAMS`) |
| `steer_error` | `agent.steer()` raised |
| `network_error` | request itself failed (frontend, `_trySteer` catch) |

**The `fallback` reason is never surfaced to the user**, so a reproducible "Steer unavailable" is undiagnosable from the UI.

[...]

PR #5077, opened 2026-06-28
fix: relay WebUI steer through gateway backend

ruizanthony:
## Summary
- relay WebUI `/steer` requests to the Hermes Gateway API-server session when WebUI uses `HERMES_WEBUI_CHAT_BACKEND=gateway`
- keep `/steer` fallback non-interrupting by queueing for the next turn instead of cancelling the active stream
- update frontend fallback copy and regression coverage for gateway steer errors

## Tests
- `python3 -m py_compile api/gateway_chat.py api/streaming.py`
- `git diff --cached --check`
- exact conflict-marker scan
- `./scripts/test.sh tests/test_webui_gateway_chat_backend.py tests/test_real_steer.py tests/test_steer_fallback_noninterrupt.py tests/test_1062_busy_input_modes.py tests/test_issue4749_steer_reason_and_recovery.py -q` — 89 passed


nesquena-hermes (changes requested):
Thanks for tackling steer relay for the gateway backend — the streaming-side wiring and the non-interrupt invariant tests are in good shape. But the core relay targets a gateway route that doesn't exist, so in gateway-backend mode `/steer` can't actually relay. Requesting changes.

## 🔴 Blocker — the relay POSTs to a non-existent gateway route (steer is a no-op in gateway mode)

`gateway_steer_session()` builds:

```python
# api/gateway_chat.py:129
url = f"{base_url}/api/sessions/{quoted_sid}/steer"   # POST {"text": ...}
```

where `_gateway_base_url()` defaults to the api-server at `http://127.0.0.1:8642`. But the api-server registers **no `/steer` route**. The complete `/api/sessions/{session_id}` route table is:

```
GET    /api/sessions/{session_id}
GET    /api/sessions/{session_id}/messages
POST   /api/sessions/{session_id}/fork
POST   /api/sessions/{session_id}/chat
POST   /api/sessions/{session_id}/chat/stream
```

There is no `…/steer` anywhere in the gateway (`grep -rniE "add_(post|get|route).*steer"` over the backend → 0 matches; the word `steer` does not appear in `api_server.py` at all). The gateway steers a run **message-driven**, not via REST: when `effective_mode == "steer"` it calls `running_agent.steer(text)` (`gateway/run.py:4727`), reached by sending a chat message — never by a dedicated endpoint.

**Consequence (deterministic):** real POST → **404** → caught as `gateway_steer_http_error` → frontend falls back to `queueSessionMessage(...)`. Net behavior in gateway mode is identical to having no steer support at all — the PR's headline goal is unmet.

Two further issues would still break it even if the route existed: (a) `sid` is the WebUI's *local* session id, not the gateway SessionDB id, so it's the wrong key space for a path-addressed `/api/sessions/{id}` route; (b) `X-Hermes-Session-Key: webui:{sid}` is a continuity header for `/v1/chat/completions` / `/v1/responses`, meaningless on a path-addressed session route.

### Fix (pick one — neither is a one-liner)
- **Backend-first:** add `POST /api/sessions/{session_id}/steer` to `gateway/platforms/api_server.py` (register next to `/fork`, resolve session → running agent → `running_agent.steer(text)` via the `gateway/run.py:4727` mechanism). This is a **separate backend PR**; this WebUI PR can't land functionally without it.
- **Message-driven:** relay steer as a chat message through the existing gateway chat path (`/v1/chat/completions` with the established session-id/continuity headers, mirroring the `api/gateway_chat.py` chat relay) so it reaches `effective_mode == "steer"`. The relay must carry the busy/steer mode the gateway runner understands — not implemented here.

Also fix the session-id namespace and drop/relocate the misused `X-Hermes-Session-Key` header.

## 🟠 Secondary — silent behavior reversal of #4749 on the *legacy* path (dead recovery banner)

`_trySteer` (`static/commands.js:1286`) was rewritten so the fallback now **unconditionally** calls `queueSessionMessage(...)` (`:1308`) instead of restoring the draft + showing the #4749 steer-recovery banner. This reverses the deliberate prior design ("a failed Steer must not be silently upgraded into Queue") and makes `_showSteerRecovery` (`commands.js:1244`) dead code — it now has exactly one reference in the file: its own definition. The #4749 retry/dismiss recovery banner is unreachable, yet the function and its `steer_recovery_*` / `steer_fail_*` i18n keys remain. This hits the in-process legacy path too, not just gateway mode. Please decide explicitly: keep the #4749 recovery behavior (and only auto-queue when the failure is genuinely unsupported), or intentionally retire it — in which case delete the now-dead function and orphaned keys.

## 🟡 Test coverage doesn't exercise the contract
The new gateway tests mock the HTTP boundary, so none can catch the 404:
- `test_gateway_steer_session_posts_to_gateway_session_steer` monkeypatches `urllib.request.urlopen` and asserts the URL equals `…/api/sessions/sid-1/steer` — i.e. it **enshrines the wrong endpoint** and can never detect that it 404s.
- `test_handle_chat_steer_relays_when_gateway_backend_enabled` monkeypatches `gateway_steer_session` to return `accepted:True` — checks wiring only.

A test that actually matters would assert the relay's target route is one the api-server registers (parametrize against the route table, or a real loopback POST expecting non-404). That test would currently **fail** — which is exactly the bug. (All targeted tests do pass green as written; the issue is what they assert, not flakiness.)

## 🟢 Nit
`cmd_steer_gateway_fallback` is added to only 3 of 16 locales (en×2, fr) and is never referenced from any JS (`_trySteer` uses `cmd_steer_fallback`/`busy_steer_fallback`). Either wire it in and add it to all locales, or remove it.

---
Happy to re-review once the relay reaches a real steer path. The non-interrupt test scaffolding is reusable — the main thing is making the relay target something the gateway actually serves.

<sub>Automated warm-up review (review-only pass). Findings verified at code level against the gateway route table; not a merge gate.</sub>

nesquena-hermes:
Pulled the branch into a read-only worktree and read `api/gateway_chat.py` + `api/streaming.py` at HEAD against `origin/master`, then cross-checked the agent side. The WebUI plumbing is clean, but there's a contract problem on the gateway end that makes the relay a no-op against a real server.

## The relay targets a gateway route that isn't registered

`gateway_steer_session` (api/gateway_chat.py) POSTs to:

```python
url = f"{base_url}/api/sessions/{quoted_sid}/steer"
```

That endpoint does not exist on the Hermes Gateway API-server. The full route table is registered in one block in the agent at `gateway/platforms/api_server.py:4437-4445`:

```python
self._app.router.add_get("/api/sessions", self._handle_list_sessions)
self._app.router.add_post("/api/sessions", self._handle_create_session)
self._app.router.add_get("/api/sessions/{session_id}", self._handle_get_session)
self._app.router.add_patch("/api/sessions/{session_id}", self._handle_patch_session)
self._app.router.add_delete("/api/sessions/{session_id}", self._handle_delete_session)
self._app.router.add_get("/api/sessions/{session_id}/messages", self._handle_session_messages)
self._app.router.add_post("/api/sessions/{session_id}/fork", self._handle_fork_session)
self._app.router.add_post("/api/sessions/{session_id}/chat", self._handle_session_chat)
self._app.router.add_post("/api/sessions/{session_id}/chat/stream", self._handle_session_chat_stream)
```

No `/steer`. `grep -n steer gateway/platforms/api_server.py` returns nothing — the substring doesn't appear in the API-server at all. At runtime aiohttp returns **404** for the unmatched path, so the relay always lands in the `HTTPError` branch and returns `{"accepted": False, "fallback": "gateway_steer_http_error", "status": 404}`. The user's guidance is never delivered to the gateway-backed run.

## Why the tests stay green anyway

Both new cases in `tests/test_webui_gateway_chat_backend.py` fully mock the transport, so they assert request *shape*, not endpoint *existence*:

- `test_gateway_steer_session_posts_to_gateway_session_steer` monkeypatches `urlopen` with a `FakeResponse` that hard-codes `{"accepted": true, ...}`.
- `test_handle_chat_steer_relays_when_gateway_backend_enabled` monkeypatches `gateway_steer_session` itself.

So "89 passed" is consistent with a route that doesn't exist. There's no test exercising a real (or fake-but-routed) aiohttp app, which is why the 404 path slips through.

## The deeper issue: there's no steerable agent to relay to

Even if you add the route, steer has nothing to act on. On the platform gateway, `/steer` mutates a live in-process agent — `running_agent.steer(steer_text)` at `gateway/run.py:8197`. The API-server has no equivalent: `_handle_session_chat` is documented as "one synchronous agent turn" (api_server.py:1623) and runs `await self._run_agent(...)` to completion per request. There's no persistent per-session running-agent registry to inject mid-run guidance into. So a gateway steer endpoint would need to either (a) queue the text for the next `/chat` turn, or (b) be backed by a new long-lived run abstraction — neither exists today.

## Recommendation

This is the wrong layer to land alone. Two options:

1. **Pair it with an agent-side PR** that adds `POST /api/sessions/{session_id}/steer` to `api_server.py` with defined semantics (most likely queue-for-next-turn, since there's no live agent to interrupt) and a matching response contract (`accepted`/`fallback`/`session_id`). Then the WebUI relay shape here is correct and an integration test against the real aiohttp app can replace the fully-mocked ones.
2. **Until that route exists**, gate the relay so gateway-backed steer falls through to the existing queue-only behavior intentionally, rather than firing a request that's guaranteed to 404.

The non-interrupting fallback refactor in `_handle_chat_steer` (api/streaming.py:9323) and the i18n/copy updates look fine on their own. The blocker is specifically the cross-repo contract: the relay calls a route the gateway doesn't serve.

PR #5078, opened 2026-06-28
Live-sync Workspace Todos through turn completion

Stacey2911:
## Thinking Path

The Workspace → Todos tab already had the correct data source when rendered, but it only refreshed on tab activation. During a live run, `todo_state` updates reached `S.todos` and the left CURRENT TASK LIST updated, while the right-side Workspace → Todos panel could remain stale if it was already open.

The first fix wired Workspace → Todos into the existing coalesced todo refresh path. Live testing then exposed a second terminal-state issue: at turn completion, the terminal session payload did not include `todo_state`, so frontend hydration could clear `S.todos` after inflight state was gone. A full browser refresh restored the list because the session GET path already attaches `todo_state`.

This PR fixes both parts of the same state-freshness issue:
- active Workspace → Todos now repaints when live todo state changes
- terminal session payloads now carry the final `todo_state` snapshot used by hydration

## What Changed

- Extended the existing `scheduleTodosRefresh()` path so active Workspace → Todos can repaint from live todo updates.
- Added Workspace → Todos visibility/feature gating so Files and Artifacts tabs remain unaffected.
- Added a separate Workspace Todos render sh/cache to avoid unnecessary DOM writes.
- Reset the Workspace Todos cache with the shared todo render cache.
- Scheduled a todo refresh after session todo hydration.
- Added `todo_state` to full terminal session payloads via the existing `attach_todo_state` mechanism.
- Added regression tests for Workspace → Todos refresh gating, cache separation, and terminal payload todo-state inclusion.
- Added a changelog entry.

## Why It Matters

Users can keep Workspace → Todos open during an active run and see task status changes without clicking away and back. The panel also remains correct after the turn finishes instead of clearing to `No active tasks`.

This preserves existing left sidebar Todos behavior and does not change Files or Artifacts workspace behavior.

## Contract Routing

Task type: UI state freshness / session todo-state rendering / terminal session payload consistency

Touched areas:
- Workspace panel Todos tab
- Shared todo refresh scheduler
- Todo hydration refresh path
- Terminal session payload construction

Relevant docs checked:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `TESTING.md`

State layer / invariant:
- Live `todo_state` updates in `S.todos` should repaint every active visible Todos surface.
- Workspace → Todos should not remain stale while active during streaming.
- Terminal session payloads that trigger frontend hydration should include the final `todo_state` snapshot.
- Files and Artifacts workspace tabs should not repaint from todo updates.

Contract Change:
- None intended. This keeps terminal session payloads consistent with the existing session GET/cold-load path that already includes `todo_state`.

## Verification

Automated validation on the PR branch:

- `node --check static/ui.js`
- `./scripts/test.sh tests/test_streaming_done_payload_message_count.py tests/test_todo_live_frontend_static.py tests/test_todo_panel_cold_load_static.py tests/test_todo_state.py tests/test_todo_state_emission.py tests/test_session_todo_state_route.py tests/test_streaming_todo_state_static.py`

Result:

- 31/31 targeted tests passed
- JS syntax checks passed
- `api/streaming.py` compile check passed

Manual live validation:

- Deployed the branch commit to a local HermesWebUI install.
- Confirmed Workspace → Todos updates live while task statuses change.
- Confirmed Workspace → Todos no longer clears to `No active tasks` after turn completion.
- Confirmed left CURRENT TASK LIST and right Workspace → Todos agree after completed/cancelled/in-progress updates.
- Captured after-evidence screenshot showing the corrected state.

Before

imageBefore

After

imageAfter

## Risks / Follow-ups

- Low risk: the UI refresh change reuses the existing `scheduleTodosRefresh()` RAF coalescing path.
- Workspace repaint is gated by feature flag, active tab, and DOM visibility.
- Separate Workspace Todos hash cache avoids repeated DOM churn.
- Terminal payload change uses the existing `attach_todo_state` mechanism already used by session retrieval paths.
- Files and Artifacts workspace behavior is unchanged.
- Follow-up: the right-panel refresh button appears focused on Files/Artifacts refresh behavior and is not broadened in this PR.

## Model Used

Provider: OpenAI / local workflow via HermesAgent and Codex
Model: Codex v0.142.2 gpt-5.5 high; ChatGPT planning/review used GPT-5.5 Thinking.
Notable tool/mode use: HermesAgent gathered local evidence and enforced workflow gates; Codex performed design and implementation; ChatGPT reviewed workflow, testing gates, and PR structure.


nesquena-hermes:
Pulled the branch into a read-only worktree and read all five non-test files at HEAD against `origin/master`, plus traced the helper contracts on both the server and frontend sides. This is a clean, well-scoped fix for both halves of the staleness bug. Notes below.

## Server side — the payload change is correctly consistent with GET

`_session_payload_with_full_messages` (api/streaming.py:59) now calls `attach_todo_state(raw, messages)` before returning:

```python
raw = session.compact() | {
    'messages': messages,
    'message_count': len(messages),
}
attach_todo_state(raw, messages)
```

This is the right mechanism — it's the same helper the session GET path uses (api/routes.py:10149 and :10279), so the settled SSE payload now carries the identical `todo_state` snapshot a cold `?messages=1` load would. `attach_todo_state` (api/todo_state.py:295) is a no-op on empty/None messages and swallows all errors, so it can't break the done payload.

One thing worth calling out explicitly (it's fine, just broader than the PR title implies): `_session_payload_with_full_messages` has three callers — the DONE payload at streaming.py:8871, the compression-error terminal payload at :8125, and (via `_redacted_...`) the cancel payload at :9700. So `todo_state` is now attached to **all** settled/terminal/cancel payloads, not just turn-completion. That's a net positive — the cancel and error paths get the same hydration consistency — but if you want to keep the blast radius minimal you could scope it to the done path. I'd leave it as-is; the broader consistency is the more correct behavior.

## Frontend side — fan-out is correctly gated

`scheduleTodosRefresh` (static/ui.js:7066) now splits the sidebar and workspace concerns cleanly:

```js
const sidebarActive=_todosPanelIsActive();
if(sidebarActive&&typeof loadTodos==='function') loadTodos();
if(typeof _refreshWorkspacePanelTodos==='function') _refreshWorkspacePanelTodos();
```

The sidebar path keeps its exact prior gating (only repaints when `_todosPanelIsActive()`), and the workspace repaint delegates its own visibility check into `_refreshWorkspacePanelTodos`. That helper (static/workspace.js) gates on the feature flag + active tab + DOM visibility:

```js
function _workspaceTodosTabIsActive(){
  if(typeof window==='undefined'||window._workspaceTodosTab!==true) return false;
  ...
  if(!rightPanel||rightPanel.dataset.activeTab!=='todos') return false;
  ...
}
```

I confirmed `window._workspaceTodosTab` is the real feature flag (set in static/boot.js:2133 and static/panels.js:7454), so Files/Artifacts tabs and installs without the flag are untouched.

## Render-cache short-circuit is safe across tab switches

The new `_workspaceTodosLastRenderedHash` guard in `_loadWorkspacePanelTodos` skips DOM writes when the hash is unchanged. I checked the tab-switch path (`_setWorkspacePanelTab*` at workspace.js:331-334) only toggles `todosPanel.hidden` — it never clears `innerHTML` — so re-activating the tab won't show a stale-but-empty panel against a non-null cache. And `_resetTodosRenderCache` (ui.js:7081) now calls `_resetWorkspaceTodosRenderCache`, so the cache is dropped on cross-session navigation, which is exactly where a stale hash would otherwise bite. The `'__empty__'` sentinel for the no-tasks state is a nice touch that avoids re-rendering the placeholder.

## Tests

The static + node-vm coverage is appropriate: `test_streaming_done_payload_message_count.py` locks the server attach, and the two node-vm cases in `test_todo_live_frontend_static.py` actually execute the extracted scheduler/gate blocks to prove (a) the sidebar still gets `loadTodos` only when active, (b) the workspace fan-out always runs and self-gates, and (c) the gate requires enabled+visible+todos-tab. Per policy I didn't run the suite from the worktree, but the assertions line up with the diff.

LGTM. The only optional call is whether to narrow the `attach_todo_state` to the done path vs. all three terminal payloads — I'd keep the broader version.

nesquena-hermes:
Reopening from the same branch — GitHub PR-head tracking stuck on a stale SHA after rebase (branch correctly rebased + CI green + Codex SAFE + activity-stream gate pass). Superseded by the fresh PR; same commits, same credit.

PR #5080, opened 2026-06-28
fix(#4731): enlarge mobile message action tap targets

### LINKED ISSUE #4731: enhancement(ux): message action buttons (edit/branch/copy) are too small to tap reliably on mobile
## Summary

The per-message hover-action icons (edit / branch / copy — the small row that appears on a message) are **too small and finicky to hit on mobile**. The tester noted they're annoying to tap even with not-large fingers. They're well under the ~44px touch-target guideline.

## Where this lives

**WebUI** (`nesquena/hermes-webui`) — pure CSS. Covers mobile web, PWA, and the native app wrappers at once.

## Root cause

`.msg-action-btn` has desktop-tuned, tiny hit areas with no mobile touch-target treatment:

`static/style.css:2677`:
```css
.msg-action-btn{background:none;border:none;color:var(--muted);cursor:pointer;
  font-size:13px;padding:2px 5px;border-radius:5px;...}
```
At `padding:2px 5px` on a ~13px glyph, the tappable area is roughly ~18-22px tall — far below the ~44px minimum for comfortable touch. The mobile media block (`@media(max-width:640px)`, `style.css:5468`) only forces `.msg-actions{opacity:1}` (so they're always visible on touch, since there's no hover) but does **not** enlarge the buttons.

For contrast, the approval buttons already got mobile treatment — `style.css:2519`: `.approval-btn{... min-height:44px}` in the mobile block. The message-action icons never got the same.

## Suggested fix / fix shape

In the existing `@media(max-width:640px)` block, give the message-action buttons a proper touch target without enlarging the visual glyph too much:

```css
@media(max-width:640px){
  .msg-action-btn{
    min-width:40px; min-height:40px;
    padding:8px;                /* bigger hit area */
    display:inline-flex; align-items:center; justify-content:center;
  }
  .msg-actions{ gap:4px; }       /* a little more spacing so adjacent icons don't fat-finger */
}
```
(Tune to ~40-44px. Keep the glyph size; just grow the padding/hit area. `touch-action:manipulation` is already applied globally to buttons at `style.css:2071`, so double-tap-zoom delay isn't the issue here — it's purely size/spacing.)

Optionally bump `font-size` a touch on mobile if the icons read too small, but the hit-area fix is the priority.

## Scope

Tiny — a few CSS lines in `static/style.css` inside the mobile media block. No JS/server changes. WebUI only.

## Environment

- Mobile (Android app / mobile web), message hover-action icons (edit / branch / copy).
- Labels: `enhancement`, `ux`.

rodboev:
## Thinking Path

- The message action icons are always visible on mobile, but they still use the desktop `padding:2px 5px` and `gap:2px` sizing, so the tap targets stay cramped.
- The issue already scoped the right fix: extend the existing mobile message-controls block in `static/style.css` instead of widening this into a broader touch-target pass.
- The patch keeps the glyphs the same size and only grows the mobile hit area and spacing, with one focused static regression test to lock the CSS shape.

## What Changed

- `static/style.css`: extended the existing `@media(max-width:640px)` message-controls block with larger `.msg-action-btn` hit areas and a wider `.msg-actions` gap on mobile.
- `tests/test_issue4731_mobile_message_actions.py`: added a focused static regression test for the mobile CSS overrides and the preserved desktop base rules.

## Why It Matters

Copy, edit, retry, fork, and TTS actions become reliably tappable on phones without changing the compact desktop presentation.

## Verification

```powershell
D:\Repos\hermes-agent\venv\Scripts\python.exe -m pytest tests/test_issue4731_mobile_message_actions.py -v --timeout=60
```

## Upstream

Closes #4731.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -6001,6 +6001,8 @@ main.main > .main-view:not([id="mainChat"]):not([id="mainSettings"]) .main-view-
   .msg-row[data-role="assistant"] .msg-foot,
   .assistant-turn .msg-foot{opacity:1;}
   .msg-actions{opacity:1;}
+  .msg-actions{gap:4px;}
+  .msg-action-btn{min-width:40px;min-height:40px;padding:8px;display:inline-flex;align-items:center;justify-content:center;}
 }
 
 /* ── /btw ephemeral side-question bubble ── */
--- a/tests/test_issue4731_mobile_message_actions.py
+++ b/tests/test_issue4731_mobile_message_actions.py
@@ -1,0 +1,32 @@
+"""Static regression coverage for the mobile message action touch-target fix."""
+from pathlib import Path
+
+
+STYLE_CSS = (Path(__file__).resolve().parent.parent / "static" / "style.css").read_text(encoding="utf-8")
+
+
+def _mobile_block():
+    marker = "@media(max-width:640px){"
+    start = STYLE_CSS.index(marker)
+    end = STYLE_CSS.index("/* ── /btw ephemeral side-question bubble ── */", start)
+    return STYLE_CSS[start:end]
+
+
+def test_mobile_block_enlarges_msg_action_buttons():
+    block = _mobile_block()
+    assert ".msg-action-btn{min-width:40px;min-height:40px;padding:8px;display:inline-flex;align-items:center;justify-content:center;}" in block
+
+
+def test_mobile_block_widens_msg_actions_gap():
+    block = _mobile_block()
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Reopening as a fresh PR from the same branch — GitHub PR-head tracking stuck on a stale SHA after rebase (branch is correctly rebased + CI green). Superseded by the fresh PR; same commit, same credit.

PR #5083, opened 2026-06-28
feat(extensions): theme-registration capability (registerHermesSkin)

nesquena-hermes:
## Summary

Adds a **theme-registration capability** so a trusted local extension can contribute a custom skin to the **native** Settings → Appearance picker, instead of each theme extension bolting on its own parallel theme switcher.

New global API:

```javascript
window.registerHermesSkin({
  name: 'E-Ink',
  value: 'e-ink',                              // optional; slugified from name if omitted
  colors: ['#000000', '#ffffff', '#555555'],  // up to 3 picker swatches
  tokens: { '--bg':'#ffffff', '--text':'#000000', '--accent':'#000000', ... }
});  // → true on success, false if rejected
```

The registered skin appears in the picker, is selectable, and persists across reloads exactly like a built-in skin.

## Why

This is the foundation for the theme-extension work (theme packs, an e-ink skin, a build-your-own-theme creator). The skin system is `data-skin="<key>"` on `<html>` + `:root[data-skin=x]{…}` CSS rules + a hardcoded `_SKINS` array driving the picker — which an extension can't cleanly append to. Rather than have every theme extension ship an awkward second switcher, core gains one small, additive registration hook they all use natively.

A single runtime API (rather than a declarative `themes.json`) is deliberate: a theme *creator* extension generates themes from user input at runtime, so it must inject `:root[data-skin]` rules dynamically — one API serves the static (e-ink, packs) and dynamic (creator) cases with no source-of-truth drift.

## Security (the load-bearing part)

Token values are written into CSS, so they're sanitized in core, **once**, so every theme extension inherits the guard safe-by-construction:

- **Token-name allowlist** — only a documented set of `--*` design tokens is accepted; anything else is dropped.
- **Value sanitization** — each value must match a strict regex (hex / `rgb()` / `rgba()` / `hsl()` / `hsla()` / CSS keyword / simple numeric-with-unit). Values containing `url()`, `expression()`, semicolons, braces, or other CSS-injection vectors are rejected.
- **Reserved-key guard** — an extension can never overwrite a built-in skin key (`default`, `ares`, `graphite`, …).
- A descriptor with **no valid tokens** after sanitization is refused entirely.
- The skin key is slugified to `[a-z0-9_-]` so it's safe as both a `data-skin` attribute and a CSS attribute selector.

## What changed

- `static/boot.js` (+102): `registerHermesSkin()` + `_sanitizeSkinTokens()` + `_renderExtensionSkinStyles()` + the allowlist/regex/reserved-key constants; exposed on `window`. Injects a single managed `<style id="hermesExtensionSkinStyles">`, appends to `_SKINS`/`_VALID_SKINS`, refreshes the picker if built, and re-applies a persisted selection that was waiting on a late-registered skin. Idempotent re-register (updates in place) supports a live theme editor.
- `docs/EXTENSIONS.md` (+51): a "Registering a custom theme (skin)" section — API, the allowed token list, the value rules, and the guarantees.
- `tests/test_extension_theme_registration.py` (new): structural pins (API/sanitizer/reserved-guard present) + a **behavioral Node harness** that extracts the real `registerHermesSkin` and drives it against adversarial input — confirms a CSS-injection descriptor is rejected, unsafe tokens are dropped while safe siblings survive, reserved keys are guarded, unknown tokens are dropped, re-register is idempotent, and garbage/no-token descriptors are refused. (Skips gracefully if `node` is unavailable.)
- `CHANGELOG.md`: `[Unreleased] → Added` entry.

## Verification

- `node --check static/boot.js` — clean
- `tests/test_extension_theme_registration.py` — 4/4 pass (incl. the behavioral harness)
- **Full suite: 10,874 passed, 10 skipped, 1 xfailed, 2 xpassed, 0 failed** (635s) — the boot.js change breaks nothing across the appearance/boot/everything-else tests.

## Notes

- Purely additive: nothing changes for users who install no theme extension. No new dependency.
- First consumer (an e-ink skin extension) is next; this PR is the capability only.
- 🤖 Agent-authored — flagging for review; not self-merging.


nesquena (approved):

Independent review of the extension theme-registration capability: a new `window.registerHermesSkin(descriptor)` lets a trusted local extension contribute a custom skin into the **native** Settings → Appearance picker. **Security-sensitive** (token values are written into CSS), so I focused on whether the sanitization is actually airtight.

### What this ships

`static/boot.js` (+102), `docs/EXTENSIONS.md` (+51), `tests/test_extension_theme_registration.py` (+169, new), `CHANGELOG.md` (+2). Agent-authored, flagged for review (not self-merging). Purely additive — no behaviour change for users with no theme extension.

Fresh tarball: pure WebUI boot/appearance code. No agent/protocol/config surface; the API is a client-side `window` hook. ✓

### Security review — the value sanitization is airtight (verified at the sink)

Tokens reach CSS via `_renderExtensionSkinStyles()` → `:root[data-skin="KEY"]{NAME:VALUE;…}` set as `styleEl.textContent`. I checked each attacker-influenced position:

- **KEY** — slugified to `[a-z0-9_-]` in `registerHermesSkin` (`rawVal.replace(/[^a-z0-9_-]/g,'')`), so it's safe both as a `data-skin` attribute value and inside the selector. Reserved core-skin keys (`default`, `ares`, …) are rejected; a collision with a non-ext skin is rejected.
- **NAME** — only keys in `_ALLOWED_SKIN_TOKENS` (a fixed `--*` allowlist) survive `_sanitizeSkinTokens`; anything else is dropped. No injection via the token name.
- **VALUE** — the only free-text field. Gated by `_SAFE_SKIN_VALUE_RE`, anchored `^…$`, alternatives = hex / `rgb()`/`rgba()` / `hsl()`/`hsla()` (inner chars constrained to `[0-9.,%\s/deg]`) / a 3–20-letter keyword / a numeric-with-unit. **None** of these admit `;`, `{`, `}`, `url(`, `expression(`, `/*`, `<`, quotes, parens (outside the constrained rgb/hsl forms), or an internal newline. A `.trim()` runs first (neutralizing the JS `$`-matches-before-a-trailing-newline quirk), and a 64-char length cap bounds it. A descriptor with **zero** valid tokens after sanitization is refused entirely.

`textContent` on a `<style>` sets raw CSS text (no HTML re-parse), so the only threat is a CSS-level rule breakout — which the regex forecloses.

**Independent CSS-output harness (11/11)** — driving the real `registerHermesSkin` + `_renderExtensionSkinStyles` and inspecting the *actual* rendered `<style>` text:
```
injection-only descriptor                              → rejected (no CSS)
partial { safe + expression() + url() + "red;}html{…" } → only --bg:#123456 & --accent:rgb(1,2,3) in CSS; unsafe siblings ABSENT
rendered CSS                                            → no "}html{", no ";}" breakout; brace-count == block-count
internal-newline "#fff\n}evil{…"                        → not in CSS
key "My Skin!! <x>"                                     → slugified to "myskinx"; no raw < / ! leaked
reserved "default"                                       → rejected
```
So a malicious descriptor — or malicious **user input to a theme-creator extension** (the case where this is a genuine boundary, not just defense-in-depth, since the creator generates tokens from user input at runtime) — cannot inject CSS. The marketed "load-bearing security" claim holds.

- **Swatch colors** (`descriptor.colors`) go through the same `_SAFE_SKIN_VALUE_RE` filter before reaching the picker, so they're safe even where rendered into inline style. ✓
- **Persisted re-apply** only re-applies a skin the user had *already* selected (`localStorage 'hermes-skin' === key`) — an extension can't force-apply an arbitrary skin; and `_applySkin(key)` uses the slugified key. ✓
- **Trust note:** extensions run with full session authority (they could inject CSS directly), so against a *malicious extension* this is defense-in-depth/least-privilege; against *user input flowing through a benign creator extension* it's a real sanitization boundary. Either way it's sound.

### Other audit

- **Idempotent re-register** — same ext key updates the `_SKINS` entry in place (supports a live theme editor); `_renderExtensionSkinStyles` rebuilds the managed `<style>` from scratch each call (no accumulation). ✓
- **No new XSS** — `<style>` via `textContent` (not `innerHTML`); name/label sliced to 40 chars and only flow to textContent/sanitized sinks.
- `try/catch` wraps the whole API → a bad descriptor returns `false`, never throws into the extension.

### Edge-case matrix

| Descriptor | Result |
|---|---|
| Clean (allowlisted tokens, safe values) | registered, appears in picker, persists ✅ |
| Only CSS-injection token values | refused (no valid tokens) ✅ |
| Mixed safe + unsafe tokens | safe kept, unsafe dropped from CSS ✅ |
| Unknown token name (not in allowlist) | dropped → refused if none valid ✅ |
| Reserved key (`default`/`ares`/…) | rejected ✅ |
| Collides with a non-ext skin key | rejected ✅ |
| Name/key with `<`, `!`, spaces | slugified; nothing leaks to CSS ✅ |
| Re-register same key | idempotent in-place update ✅ |
| `null` / no-token descriptor | rejected ✅ |
| User had this skin selected before it registered | applied on registration ✅ |

### Tests

- `tests/test_extension_theme_registration.py` (4) — structural pins + a behavioral Node harness driving the real registration against adversarial input (injection rejected, partial-safe survives, reserved guarded, unknown dropped, idempotent, garbage refused). **4/4.**
- **Independent CSS-output harness: 11/11** (above) — adds the sink-level check (unsafe values never reach the rendered `<style>`) that the PR's harness can't, since it stubs the style element.
- Boot / skin / appearance / extension blast-radius — **485 passed, 2 skipped**.
- **Full suite: 10736 passed / 143 skipped / 0 failures** in 222s (deselected the pre-existing darwin CRLF flake).
- `node --check static/boot.js` clean; ESLint runtime guard exit 0; CI lint + browser-smoke green (test shards finishing).

- `_SAFE_SKIN_VALUE_RE` is intentionally conservative — it rejects valid-but-exotic CSS color forms (e.g. `color-mix()`, 4-value `rgb()` with `/ alpha` spacing edge cases, named system colors >20 chars). That's the right trade for a security regex (reject-unknown), and a theme author can always pick a supported form; worth a note in the docs token list if authors hit it. The docs already enumerate the allowed forms.
- The hex alt allows `{3,8}` hex digits (so 5/7-digit values pass the regex), but a malformed hex is a harmless no-op in CSS — not an injection vector.

### Recommendation

A well-designed additive capability that puts the security-sensitive work (CSS token sanitization) in core once so every theme extension inherits it. I verified the guard at the actual sink — slugified key, allowlisted token names, and a strictly-anchored value regex that (with a leading `.trim()`) forecloses `;`/`{`/`}`/`url()`/`expression()`/comments/newlines — proving with an 11/11 CSS-output harness that no unsafe value reaches the rendered `<style>`, including the theme-creator user-input case where it's a real boundary. Reserved keys are protected, re-register is idempotent, and there's no new XSS surface. Ship.

PR #5085, opened 2026-06-28
fix(boot): skip archived saved session on refresh

### LINKED ISSUE #5064: bug(boot/sessions): refresh reopens an archived chat instead of the New Conversation empty state when all chats are archived
### Summary

After archiving all conversations, a fresh refresh of the WebUI reopens the last archived chat instead of showing the "New conversation" / empty state.

From a report in the Discord #report-bugs channel:

> If I archive all my chats, I expect that on a fresh refresh of the WebUI it will show the "New conversation" window. Instead, it shows the last archived chat. I don't think this is intended — and if it is, I really think it should change. At least in the case where all chats are archived.

### Steps to reproduce

1. Have one or more conversations, with one currently open (so `hermes-webui-session` is set in localStorage).
2. Archive all of them (including the currently-open one).
3. Hard-refresh / reopen the WebUI.
4. Observed: the last archived chat is displayed.
5. Expected: the New Conversation / empty state, since there are no active (non-archived) conversations.

### Root cause (traced in source)

Two gaps combine:

1. **Archiving doesn't clear the saved-session pointer.** `_archiveSession` (static/sessions.js) sets `session.archived = true` and updates `S.session.archived`, but it never removes the `hermes-webui-session` localStorage key — even when the session being archived is the currently-open one. So the saved-session ID persists across the refresh.

2. **The boot guard doesn't treat archived sessions as "don't auto-open."** On boot, `static/boot.js` reads `savedLocal = localStorage.getItem('hermes-webui-session')` and calls `_savedSessionShouldStaySidebarOnly(savedLocal)` to decide whether to skip restoring it. That guard only checks for an in-flight stream:

   ```js
   return !!(session && (session.active_stream_id || session.pending_user_message));
   ```

   It does **not** check `session.archived`. So an archived saved session falls through to `loadSession(saved)` and is displayed.

### Suggested fix direction

Either (or both):

- In `_savedSessionShouldStaySidebarOnly`, also return `true` when `session.archived` is truthy, so an archived saved session is skipped on boot and the empty state shows. This is the more robust path because it also covers a session archived from another client/device.
- In `_archiveSession`, when the archived session is the currently-saved one (`localStorage.getItem('hermes-webui-session') === session.session_id`), clear that key.

The first option alone fixes the reported case; doing both keeps the saved pointer honest.

### Environment

- Surface: WebUI boot / session restore + archive flow.
- Reproducible whenever the saved last-open session is archived.

Paladin173:
## Thinking Path

- Issue #5064 traces a root boot restore gap: `hermes-webui-session` can still point at a chat after it has been archived.
- The boot guard already skips projecting saved sessions with active stream state into the active pane; archived saved sessions need the same sidebar-only treatment.
- Clearing the saved localStorage pointer during same-client archive keeps the browser hint honest, while the server-backed boot guard still covers cross-client archive cases.

## What Changed

- Treat `session.archived` as sidebar-only in `_savedSessionShouldStaySidebarOnly()` so root `/` refreshes do not auto-open archived saved sessions.
- Clear `hermes-webui-session` when archiving the remembered session on the same client.
- Extend the existing static regression tests for the boot guard and archive action ordering.

## Why It Matters

After a user archives all chats, refreshing WebUI should land on the New Conversation empty state instead of reopening the last archived chat from localStorage. Explicit `?session=...` links are unchanged and still intentionally open the requested session.

Fixes #5064.

## Verification

- `git diff --check`
- `E:\GitHub\hermes-webui\.venv\Scripts\python.exe -m pytest --noconftest tests/test_1694_root_saved_running_policy.py tests/test_session_action_menu_regression.py -q` — 9 passed

Note: the normal repo test runner is Unix-venv oriented under Git Bash on this Windows checkout, and a normal pytest invocation hits the known Windows symlink privilege issue in `tests/conftest.py` (`WinError 1314`). These two regression files are source-level/static checks, so `--noconftest` is the local Windows workaround used here.

## Risks / Follow-ups

- Low risk: the change is scoped to root boot restore of saved localStorage sessions and same-client archive bookkeeping.
- Explicit URL/query session opens remain outside this guard.
- No changelog entry is included per contributor workflow preference for this branch.

## Model Used

OpenAI GPT-5 Codex in Codex desktop, with local shell and GitHub CLI.

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -85,7 +85,7 @@ async function _savedSessionShouldStaySidebarOnly(sid){
   try{
     const data = await api(`/api/session?session_id=${encodeURIComponent(sid)}&messages=0&resolve_model=0`);
     const session = data&&data.session;
-    return !!(session&&(session.active_stream_id||session.pending_user_message));
+    return !!(session&&(session.active_stream_id||session.pending_user_message||session.archived));
   }catch(e){
     return false;
   }
--- a/static/boot.js
+++ b/static/boot.js
@@ -81,13 +81,20 @@ async function cancelSessionStream(session){
 }
 
 async function _savedSessionShouldStaySidebarOnly(sid){
+  const state = await _savedSessionSidebarOnlyState(sid);
+  return !!(state&&state.sidebarOnly);
+}
+
+async function _savedSessionSidebarOnlyState(sid){
   if(!sid) return false;
   try{
     const data = await api(`/api/session?session_id=${encodeURIComponent(sid)}&messages=0&resolve_model=0`);
     const session = data&&data.session;
-    return !!(session&&(session.active_stream_id||session.pending_user_message||session.archived));
+    const archived = !!(session&&session.archived);
+    const running = !!(session&&(session.active_stream_id||session.pending_user_message));
+    return {sidebarOnly:archived||running, archived};
   }catch(e){
-    return false;
+    return null;
   }
 }
 
@@ -2479,7 +2486,13 @@ window._applyTitlebarProfileVisibility=_applyTitlebarProfileVisibility;
   const saved=urlSession||savedLocal;
   if(saved){
     try{
-      if(!urlSession&&savedLocal&&await _savedSessionShouldStaySidebarOnly(savedLocal)){
+      const savedSidebarOnlyState=(!urlSession&&savedLocal)
+        ? await _savedSessionSidebarOnlyState(savedLocal)
+        : null;
+      if(savedSidebarOnlyState&&savedSidebarOnlyState.sidebarOnly){
+        if(savedSidebarOnlyState.archived){
+          try{localStorage.removeItem('hermes-webui-session');}catch(_){}
+        }
         S.session=null; S.messages=[]; S.activeStreamId=null; S.busy=false;
         S._bootReady=true;
         syncTopbar();syncWorkspacePanelState();
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -3558,6 +3558,7 @@ async function _archiveSession(session, archived=true, beforeListRender=null){
     const cached=(_allSessions||[]).find(s=>s&&s.session_id===session.session_id);
     if(cached) cached.archived=archived;
     if(S.session&&S.session.session_id===session.session_id) S.session.archived=archived;
+    try{ if(archived&&session.session_id&&localStorage.getItem('hermes-webui-session')===session.session_id) localStorage.removeItem('hermes-webui-session'); }catch(_){ }
     showToast(session.archived?_sessionArchiveToast(response,session):t('session_restored'));
     if(renderHold) await renderHold;
     if(_showArchived&&!_sessionPrefersReducedMotion()) _sessionSwipeReturnOffsets.set(session.session_id,'0px');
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Pulled the branch into a read-only worktree and read both changed files at HEAD against `origin/master`. This implements exactly the two-part fix from the #5064 triage, and both halves land where they should.

## Boot guard (gap #2 — the robust fix)

`_savedSessionShouldStaySidebarOnly` at static/boot.js:88 now adds the `archived` branch:

```js
return !!(session&&(session.active_stream_id||session.pending_user_message||session.archived));
```

This is the right primary fix because the decision is made from the freshly-fetched `/api/session?...&messages=0&resolve_model=0` payload, not local state — so it also covers the cross-client case (session archived on another device). I confirmed `archived` is actually present on that lightweight payload: the metadata-only path serializes it via `compact()`/`to_dict` at api/models.py:1038 (`'archived': self.archived`), and the sidebar/metadata projection sources it through `load_metadata_only` (api/models.py:4879), so `session.archived` is available here at no extra cost. When the guard returns `true`, the boot path at boot.js:2482 already nulls `S.session`, shows `emptyState`, renders the sidebar and returns — so no further boot change is needed.

## Saved-pointer cleanup (gap #1)

static/sessions.js:3561 inside `_archiveSession`:

```js
try{ if(archived&&session.session_id&&localStorage.getItem('hermes-webui-session')===session.session_id) localStorage.removeItem('hermes-webui-session'); }catch(_){ }
```

Three things I checked and they're correct:
- It's gated on `archived` (true only), so **unarchive** doesn't wipe the pointer.
- The `===session.session_id` equality means archiving some *other* session from the sidebar won't clear the pointer for the one you have open.
- The `try/catch` matches the existing localStorage-access pattern in this file (e.g. sessions.js:957, 1047).

## Deep-link is preserved

The empty-state branch at boot.js:2482 is still gated on `!urlSession`, so an explicit `?session=<archived-id>` deep link continues to open — an explicit URL is an intentional request, the localStorage pointer is just "last open." Good that this wasn't touched.

## One minor edge (non-blocking)

Same-client archive→unarchive without a refresh now loses the "last open" pointer: the archive clears `hermes-webui-session`, and the restore path doesn't re-set it. So a subsequent refresh lands on the empty state rather than reopening the just-restored chat. That's a very narrow flow and arguably acceptable (the boot guard alone — option 1 — would have handled the reported bug without the localStorage removal). Not worth blocking on; just noting it in case you'd rather rely on the guard and drop the same-client removal, or re-set the pointer on restore.

## Tests

Both locks match the repo's source-assertion convention for boot/scroll invariants. `test_1694_root_saved_running_policy.py` pins `session.archived` in the helper, and `test_session_action_menu_regression.py` pins both the guard string and its ordering (after the `S.session.archived` write, before the toast). These pin exactly the lines that would regress.

LGTM — the boot-guard half alone fixes the reported case, and the pointer cleanup keeps same-client state honest.

nesquena-hermes:
Reopening as a fresh PR from the same branch — GitHub PR-head tracking stuck on a stale SHA after rebase (branch correctly rebased + CI green + Codex SAFE). Superseded by the fresh PR; same commits, same credit.

PR #3758, opened 2026-06-07
fix: harden WebUI security-sensitive defaults

fantasticsquirrel:
## Thinking Path
A security review found several high-risk default/path trust issues around first-run onboarding, Docker exposure, update checks, and media file serving. This PR keeps the changes scoped to those reviewed surfaces and adds regression coverage for the new security contracts.

## What Changed
- Ignore `X-Forwarded-For` / `X-Real-IP` for unauthenticated onboarding local-network checks unless `HERMES_WEBUI_TRUST_FORWARDED_FOR=1` is explicitly set.
- Add Docker/public-bind guard: Docker images set `HERMES_WEBUI_REQUIRE_AUTH_FOR_PUBLIC_BIND=1`, and startup refuses non-loopback passwordless binds unless explicitly disabled.
- Expand Docker init log masking to include password/secret/credential/cookie/session-like environment variable names.
- Make `GET /api/updates/check` cache-only and move git/network refreshes to `POST /api/updates/check`.
- Remove all-of-`/tmp` from default `/api/media` allowed roots; temporary files now require session `MEDIA:` grants or explicit `MEDIA_ALLOWED_ROOTS`.
- Update browser callers, tests, and changelog for the changed contracts.

## Why It Matters
- Prevents spoofed forwarded headers from bypassing local-only unauthenticated onboarding gates.
- Prevents the Docker default `0.0.0.0` bind from exposing a passwordless WebUI by default.
- Reduces accidental credential disclosure in init logs.
- Keeps GET update status read-only/cache-only instead of causing `git fetch` side effects.
- Reduces arbitrary local file disclosure risk through broad `/tmp` media reads.

## Verification
- `pytest tests/test_security_review_fixes.py tests/test_update_checker.py tests/test_update_check_ui.py tests/test_api_timeout.py -q`
- `pytest tests/test_media_inline.py tests/test_security_review_fixes.py tests/test_update_checker.py tests/test_update_check_ui.py tests/test_api_timeout.py tests/test_v050260_docker_invariants.py tests/test_issue2453_agent_source_boundary.py -q`
- `python3 -m py_compile api/routes.py api/updates.py server.py`

## Risks / Follow-ups
- Bare `/tmp` media URLs are now denied unless represented by a session media grant or explicit `MEDIA_ALLOWED_ROOTS`. This is intentionally stricter and may require operators with custom scratch directories to configure an explicit allow-list.
[...]

PR #5086, opened 2026-06-28
fix(server): refuse passwordless public bind (fail-closed) — salvage of #3758

nesquena-hermes:
## Salvage of #3758 — public-bind-requires-auth hardening

Salvaged from #3758 (thanks @fantasticsquirrel) — this takes **only** the public-bind-requires-auth guard. The other parts of the original PR were handled separately: the onboarding-XFF and update-check-split hardenings already shipped to master, and the `/tmp` media-root removal was intentionally dropped (shared screenshots render via the per-session `MEDIA:` token grant, so it wasn't worth the breakage risk).

### What it does
Refuses to start (fail-closed `sys.exit(1)`) when the server would bind to a **public/non-loopback address with no authentication configured**, instead of only printing a warning — so a passwordless WebUI can't be exposed to the network by accident.

- Loopback (`127.0.0.1` / `::1` / `localhost`) never trips it.
- Auth-enabled (password **or** passkey) never trips it.
- Bare-metal dev keeps the existing **warn-only** behavior (no hard exit) unless explicitly opted in.
- Containers are protected by default (`Dockerfile` sets `HERMES_WEBUI_REQUIRE_AUTH_FOR_PUBLIC_BIND=1`).
- Operators who terminate auth at another layer (reverse proxy / VPN / private net) opt out with `HERMES_WEBUI_REQUIRE_AUTH_FOR_PUBLIC_BIND=0`.

### Crystal-clear refusal message
On the dangerous case it prints a bordered, scannable block: **what happened + why it's a risk + three concrete fixes** (set a password [recommended] / bind to `127.0.0.1` / explicit opt-out), echoing the detected host. Designed so an existing user hard-stopped on upgrade instantly understands how to recover.

### Structure note
The predicate + message live in a new thin module `api/bind_guard.py` (re-exported as `server._public_bind_requires_auth`) to keep `server.py` under its 750-line architectural guard — behavior and API surface are identical to an inline version.

### Tests
9 new cases in `tests/test_security_review_fixes.py` (container-public-passwordless blocks; loopback/auth-enabled/bare-metal-dev exempt; explicit flag on/off; Dockerfile default; message clarity + dynamic host). Full suite: 10,793 passed, only the 4 known pre-existing flakes; `test_v050260_docker_invariants.py` 53/53.

Co-authored-by: fantasticsquirrel


nesquena-hermes:
## ⚠️ Gate finding — the Dockerfile default-on guard bricks passwordless containers (Docker smoke RED)

The Docker smoke tests (single / two-container / three-container) **fail** on this branch, while browser-smoke passes. Root cause from the smoke log:

```
container hermes-smoke-single-...-hermes-webui-1 is unhealthy
hermes-webui-1  |   REFUSING TO START: passwordless server would be exposed on a public address
hermes-webui-1  |   == Environment variable HERMES_WEBUI_REQUIRE_AUTH_FOR_PUBLIC_BIND [1] already set
hermes-webui-1  | !! ERROR: hermes-webui failed or exited with an error
```

So the guard is working exactly as designed — but the **Dockerfile default `HERMES_WEBUI_REQUIRE_AUTH_FOR_PUBLIC_BIND=1`** means **every passwordless container bind now fails closed**, including the smoke tests and, more importantly, **any existing user who runs the Docker image without a password set** would be hard-stopped on upgrade. That's a much larger blast radius than "prevent accidental exposure" — it's "passwordless Docker is now unbootable by default."

### The design decision this surfaces (for @nesquena)
The security intent is right, but defaulting the guard **ON in the Dockerfile** is too aggressive. Options:
1. **Don't default it on in the Dockerfile** — keep the guard opt-in everywhere (the warn path still fires for passwordless public binds; users who want fail-closed set the env var). Lowest breakage, still ships the capability. *(my lean)*
2. **Keep default-on but make the smoke tests + the shipped Docker quickstart set `HERMES_WEBUI_PASSWORD`** — and accept that existing passwordless Docker users get hard-stopped on upgrade (needs a prominent migration note).
3. **Default-on only when a public bind is *actually* detected AND no reverse-proxy signal** — more nuance, more complexity.

This is parked for your call regardless (security PR). Flagging that as-is it would brick passwordless Docker deployments — I'd recommend option 1 (drop the Dockerfile default; keep the guard opt-in) before this ships. Happy to make that one-line Dockerfile change + re-gate once you pick a direction.

nesquena (changes requested):
## Review — changes requested 🔶 (the security design is right; the container rollout breaks CI and the default deployment)

Independent review of the salvage of #3758 (@fantasticsquirrel): refuse to start (fail-closed `sys.exit(1)`) when the WebUI would bind to a public/non-loopback address with no auth. **I strongly support the intent** — silently exposing a passwordless WebUI (filesystem + agent + memory) to the network is exactly the footgun worth closing, and the implementation is clean. But as it stands it can't merge: **all three Docker smoke jobs are red, and for a real reason that also affects the default deployment.**

### What this ships

`api/bind_guard.py` (+91, new — predicate + refusal message), `server.py` (+ wiring/`sys.exit(1)`), `Dockerfile` (sets `HERMES_WEBUI_REQUIRE_AUTH_FOR_PUBLIC_BIND=1`), `tests/test_security_review_fixes.py` (+9 cases), `CHANGELOG.md`. Agent-authored (`nesquena-hermes`). Currently `CONFLICTING` (CHANGELOG) and **CI-red on Smoke single / two-container / three-container**.

### The guard logic is correct

`_public_bind_requires_auth(host, within_container, auth_enabled)` (`api/bind_guard.py:21`): auth configured → never block; loopback → never block; explicit `…REQUIRE_AUTH_FOR_PUBLIC_BIND` off/on honored; else default to `within_container`. Clean precedence, fails closed only for the genuinely dangerous case. The refusal message is genuinely excellent — states what/why + three concrete recovery paths. The 9 unit cases pass (15/15 in the bind-subset locally). No correctness issue with the predicate itself. ✓

### The blocker — the guard refuses the smoke container, and that mirrors the default deployment

The smoke log is unambiguous:

```
container hermes-smoke-single-…-hermes-webui-1 is unhealthy
hermes-webui-1 |   REFUSING TO START: passwordless server would be exposed on a public address.
hermes-webui-1 |     Hermes WebUI was about to bind to host '0.0.0.0' with NO password
hermes-webui-1 | !! ERROR: hermes-webui failed or exited with an error
```

The Dockerfile sets `HERMES_WEBUI_REQUIRE_AUTH_FOR_PUBLIC_BIND=1`, and **all three smoke compose files run passwordless on `0.0.0.0`** (`HERMES_WEBUI_PASSWORD` is commented out in `docker-compose.yml:46`, `docker-compose.two-container.yml:121`, `docker-compose.three-container.yml:142`; each sets `HERMES_WEBUI_HOST=0.0.0.0`). So the guard does exactly what it's designed to do and the containers refuse to boot → all three smoke jobs fail.

This isn't a flaky CI or a test that needs a trivial tweak — **those compose files are the templates users copy**, so the same refusal hits:
1. the default `docker compose up` out-of-box experience, and
2. **every existing passwordless container deployment on upgrade** — their container will stop booting.

That's a deliberate, defensible security posture, but it's a **breaking change to the default container behavior**, and right now nothing in the rollout accounts for it.

### What I'd like resolved before approving

1. **Get the smoke CI green** by reconciling the compose templates with the new default. Whichever way the posture goes, the three smoke compose files (and any other passwordless `0.0.0.0` deployment templates) need to either set `HERMES_WEBUI_PASSWORD`, set `HERMES_WEBUI_REQUIRE_AUTH_FOR_PUBLIC_BIND=0` with an explanatory comment, or bind `127.0.0.1`. A red gate can't merge.
2. **Decide + document the breaking-change rollout** (maintainer call): is breaking passwordless-container-on-upgrade acceptable for the security win (I'd lean yes, fail-closed is the right default), and if so —
   - flag it as a **BREAKING** change prominently in the CHANGELOG (the current entry reads as a feature, not a breaking default change), and
   - consider whether the shipped `docker-compose*.yml` templates should now ship with a password placeholder that's *required*, a generated password, or a `127.0.0.1` default, so `docker compose up` works out of the box again.
   - optionally, a one-release **warn-then-enforce** migration to soften the upgrade break.
3. **Rebase** onto master (only CHANGELOG conflicts, same `### Added`/Fixed collision as the rest of the batch).

I deliberately did **not** push a CI-greening patch myself: making smoke pass by editing the compose/workflow would mask the user-facing default-deployment break without the maintainer explicitly accepting that posture — and choosing between "ship a password," "default to localhost," or "warn-then-enforce" is a product/security decision that's yours, not mine.

### Tests

- `tests/test_security_review_fixes.py` bind-guard cases — **15/15 locally** (predicate correctness: container-public-passwordless blocks; loopback / auth-enabled / bare-metal-dev exempt; explicit flag on/off; message clarity + dynamic host). The unit coverage is solid.
- **CI: Smoke single / two-container / three-container all FAIL** (the guard refusing the passwordless smoke containers — see above). Lint + unit shards + browser-smoke pass.

### Recommendation

🔶 **Request changes** — not on the guard (the predicate, message, and unit tests are right and I want this protection in), but on the rollout: the three smoke jobs are red because the new container default refuses the passwordless compose templates, which also breaks the default `docker compose up` and existing passwordless deployments on upgrade. Please green the smoke CI by reconciling the compose templates with the new default, flag the breaking change in the CHANGELOG, and (maintainer call) decide whether to soften the upgrade path. Happy to re-review immediately once CI is green — this is close, and the security win is worth landing.

PR #3228, opened 2026-05-31
Add safe config.yaml viewer

AJV20:
## Thinking Path
Issue #2929 asks for safer config visibility in the WebUI. The safest first slice is a read-only viewer that redacts credential-looking values before anything reaches the browser, avoiding the risk of overwriting hidden secrets from an in-browser editor.

## What Changed
- Added `GET /api/config/safe` for the active profile config.
- Redacts credential-looking key paths before serializing config for the client.
- Runs non-secret strings through the existing text redactor as a second safety net.
- Added a Settings → System panel to refresh and copy redacted YAML.
- Kept the panel read-only so redacted placeholders cannot overwrite real config values.
- Added styling, i18n strings, changelog entry, and regression coverage.

## Verification
- `python3 -m py_compile api/routes.py tests/test_safe_config_viewer.py`
- `node --check static/panels.js`
- `node --check static/i18n.js`
- `pytest tests/test_safe_config_viewer.py tests/test_security_redaction.py -q`
- `git diff --check`

## Risks / Follow-ups
Low risk: no writes are introduced. A follow-up can add a structured editor for explicit non-secret allowlisted settings.

PR #5088, opened 2026-06-28
feat(settings): read-only safe config.yaml viewer (salvage of #3228)

nesquena-hermes:
## Summary

Read-only **safe `config.yaml` viewer** in Settings → System. Adds `GET /api/config/safe`, which returns the active profile's `config.yaml` with every credential-looking value **redacted before anything reaches the browser**, plus a panel that displays it and a "copy redacted YAML" button.

Salvaged from deserted PR **#3228** (thanks **@AJV20**) and ported fresh onto current `master` (the original was conflicting on a stale base). **Addresses #2929** — note that #2929 was mis-closed by #5010, which only added `max_tokens`; this is the actual config-viewer the issue asked for.

## Design

- **Read-only by design.** No in-browser editing — that avoids the risk of an edit round-tripping a hidden/redacted secret back over the real value.
- **Path-based, fail-closed redaction.** `_redact_config_for_display` walks the config tree; any scalar whose key path contains a credential-bearing fragment is replaced with `[REDACTED]`, regardless of its Python type.
- **No absolute path disclosure.** The response carries only the config **basename** (`filename`), never the server's home-dir layout.

## The 3 maintainer-required fixes (all applied)

1. **Numeric/bool secret ordering (critical).** The sensitive **key-path check now runs BEFORE** the `isinstance(value, (bool, int, float))` passthrough, so a numeric/bool value under a sensitive key path (e.g. `token: 12345`, `secret: true`) is masked to `[REDACTED]` instead of leaking the raw value. Covered by a **non-vacuous** regression test (it fails if the two branches are reordered — verified).
2. **Dropped the absolute config path.** The endpoint returns `filename` (basename) only; the `path` field is gone. A test asserts the absolute path string appears nowhere in the response.
3. **Redaction-disabled doc note.** A UI note documents that the path-based `[REDACTED]` is **unconditional**, while the *value-level* scrub of secrets accidentally pasted into non-secret keys (`_redact_text`) only runs when `api_redact_enabled` is on.

## Redaction-completeness pass

The sensitive-key-path fragment list was aligned with the canonical `api/helpers._SENSITIVE_LOWER_MARKERS` vocabulary and expanded for the full config surface:

`api_key`, `apikey`, `api-key`, `x-api-key`, `access_key`, `secret_key`, `signing_key`, `private_key`, `client_secret`, `secret`, `raw_secret`, `secret_input`, `password`, `passwd`, `passphrase`, `access_token`, `refresh_token`, `id_token`, `auth_token`, `token`, `bearer`, `authorization`, `oauth`, `auth`, `credential`, `key_material`, `cookie`, `session_id`, `sessionid`, `session_key`, `session_token`, `session_secret`, `webhook`, `signature`, `salt`.

Fragments are **specific rather than bare words** (e.g. `private_key`/`access_key` instead of a bare `key`; explicit `session_*` names instead of a bare `session`) so harmless knobs like `voice.record_key`, `browser.allow_private_urls`, `session_ttl_seconds`, and `max_live_sessions` aren't over-redacted — every actual credential is still named by a specific fragment. Audited against a representative real `config.yaml`.

## Security contract

**No credential value, of any type (string, number, or boolean), under any sensitive key path appears in the `/api/config/safe` response.**

## Tests

`tests/test_safe_config_viewer.py` — redactor units (incl. the **non-vacuous** numeric-secret and bool-secret cases), an empty/`None` no-false-`[REDACTED]` case, a no-absolute-path assertion (fix #2), an over-redaction guard for non-secret `session_*` knobs, a **completeness sweep** driving secrets of many types under many sensitive key paths and asserting none leak, and an **end-to-end** endpoint test via `handle_get`. i18n: 10 new `t()` keys added to all 14 locale blocks.

Targeted:
```
tests/test_safe_config_viewer.py .............  13 passed
```

Full suite (`pytest tests/ -q -p no:xdist`):
```
4 failed, 10797 passed, 92 skipped, 1 xfailed, 2 xpassed, 8 warnings, 16 subtests passed in 508.02s
```
The only 4 failures are the **documented pre-existing flakes**, all unrelated to this change:
- `test_scheduled_jobs_profile_isolation.py::test_cron_profile_context_pins_profile_home`
- `test_scheduled_jobs_profile_isolation.py::test_cron_profile_context_for_home_pins_explicit_home`
- `test_tls_support.py::TestTLSEndToEnd::test_tls_startup_failure_fallback_to_http`
- `test_v050259_sessiondb_fd_leak.py::test_session_db_close_is_idempotent`

Co-authored-by: AJV20


nesquena (approved):

Independent review of the salvage of #3228 (@AJV20): a **read-only safe `config.yaml` viewer** — `GET /api/config/safe` returns the active profile's config with every credential-looking value redacted *before* anything reaches the browser, plus a Settings → System panel and a copy-redacted-YAML button. Addresses #2929. Since this exposes config, the entire review is the redaction completeness + the auth/read-only contract.

### What this ships

`api/routes.py` (+~120 — `_SENSITIVE_CONFIG_KEY_FRAGMENTS`, `_config_path_is_sensitive`, `_redact_config_for_display`, `_safe_config_yaml_text`, the GET handler), `static/index.html` + `static/panels.js` + `static/style.css` + `static/i18n.js` (panel + 10 keys × 14 locales), `tests/test_safe_config_viewer.py` (+new, 13 tests), `CHANGELOG.md`. Agent-authored (`nesquena-hermes`). MERGEABLE after rebase.

Branch was `CONFLICTING` on the `### Added` CHANGELOG block. Rebased onto current master (only CHANGELOG conflicted; kept all entries). Code merged cleanly. Force-pushed `70e220fb`.

Fresh tarball: WebUI-only, read-only. It reads `get_config()` (the active profile's config) and renders a redacted copy. No agent surface, **no write path** (the CLI/agent config is never mutated by this endpoint). ✓

### Security audit — redaction completeness (the load-bearing check)

`_redact_config_for_display(value, path)` walks the tree:
- **dict** → recurse, extending `path` per key; **list** → recurse keeping the parent path (so a list under a sensitive key has all items masked); **scalar** → checked.
- **Maintainer fix #1 verified:** the sensitive-path check runs **before** the `isinstance(value, (bool, int, float))` passthrough, so numeric/bool secrets are masked, not leaked. `return "[REDACTED]" if value not in (None, "") else value` — empties stay empty (no false `[REDACTED]`).
- **Ancestor-aware:** `_config_path_is_sensitive` substring-matches the full dotted path against `_SENSITIVE_CONFIG_KEY_FRAGMENTS`, so a value nested anywhere under `secrets.`/`oauth.`/`*api_key*` is masked regardless of its own leaf name.
- Non-sensitive strings still pass through the existing `_redact_text` value-level scrub (honors `api_redact_enabled`); the path-based mask is **unconditional** (fix #3, documented in the UI).
- **Fragment list** is aligned with the canonical `api.helpers._SENSITIVE_LOWER_MARKERS` vocabulary and deliberately **specific** (`access_key`/`secret_key`/`private_key` not bare `key`; explicit `session_token`/`session_secret` not bare `session`) — so `session_ttl_seconds`, `max_live_sessions`, `voice.record_key`, `browser.allow_private_urls` aren't over-redacted, while every real credential is named.

**I ran an independent behavioral harness** against the real function:

| Input | Output |
|---|---|
| `{providers:{x:{token:12345}}}` (numeric) | `token: "[REDACTED]"` ✅ |
| `{secret: true}` (bool) | `secret: "[REDACTED]"` ✅ |
| `{api_keys:["sk-AAA","sk-BBB"]}` (list) | both `"[REDACTED]"` ✅ |
| `{oauth:{client_id:"abc"}}` (ancestor) | `client_id: "[REDACTED]"` ✅ |
| `{secrets:{deep:{whatever:"topsecret"}}}` | `"[REDACTED]"` ✅ |
| `{headers:{x-api-key:"XYZ"}}` | `"[REDACTED]"` ✅ |
| `{session_ttl_seconds:3600, max_live_sessions:5, voice:{record_key:"F2"}, browser:{allow_private_urls:true}}` | **all pass through unredacted** ✅ |
| `{api_key:""}` / `{token:null}` | `""` / `null` (no false mask) ✅ |

No secret of any type under a sensitive key path leaked; no harmless knob over-redacted.

### Other security checks

- **Auth-gated:** `/api/config/safe` is not in `PUBLIC_PATHS` → requires auth when enabled. ✓
- **Read-only:** the path appears **only** in `handle_get` (routes.py:10289) — there is no POST/PUT/write route, so a hidden/redacted secret can never round-trip back over the real value. ✓
- **No absolute-path disclosure** (fix #2): the response carries `filename: cfg_path.name` (basename) only; no `path` field, so the host home-dir layout isn't exposed. ✓
- **YAML dump:** `yaml.safe_dump` (not `dump`) on the already-redacted structure, JSON fallback on error. ✓
- Wrapped in try/except → 500 on read error, no stack/secret leak to the client. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Numeric/bool secret under sensitive key | `[REDACTED]` ✅ |
| List of secrets under sensitive key | each `[REDACTED]` ✅ |
| Secret nested under sensitive ancestor | `[REDACTED]` ✅ |
| Non-secret `session_*`/`record_key` | unredacted ✅ |
| Empty / None under sensitive key | preserved (no false mask) ✅ |
| Unauthenticated request (auth on) | rejected ✅ |
| Any write attempt | no route exists ✅ |
| Config read fails | 500, no leak ✅ |

### Tests

- `tests/test_safe_config_viewer.py` — **13/13**: redactor units incl. the **non-vacuous** numeric- and bool-secret regressions (fail if the branches are reordered), empty/None no-false-`[REDACTED]`, no-absolute-path assertion (fix #2), over-redaction guard for `session_*` knobs, a completeness sweep over many types × many sensitive paths, and an end-to-end `handle_get` test.
- **Full suite: 10915 passed / 143 skipped / 0 failed** in 222s (deselected the known darwin CRLF flake). The PR's 4 flagged failures did not reproduce here.
- Independent behavioral harness (above) confirms the contract.

- A secret stored under a key whose path contains **no** sensitive fragment (e.g. a token pasted into `providers.x.value`) is only caught by the value-level `_redact_text`, which is a no-op when `api_redact_enabled` is off. This is the documented contract (fix #3 UI note) — the path-based guarantee covers all conventionally-named credential keys, which is what a "safe viewer" promises. Acceptable.
- Config **path values** (e.g. `workspace: /home/...`) still render — that's the user's own config; fix #2 was specifically about not disclosing the config file's server location, which is satisfied.
- CHANGELOG entry follows AGENTS.md convention; kept through the rebase.

### Recommendation

A careful, security-first read-only viewer. The redactor is fail-closed and type-agnostic (path-check before the numeric/bool passthrough — verified by harness and a non-vacuous regression), ancestor-aware, and tuned to avoid over-redaction; the endpoint is auth-gated, write-free, and discloses only the config basename. Rebased to clear the CHANGELOG conflict; full suite clean. Ship.

nesquena-hermes:
⛔ Bouncing — this is a **security blocker**: the "safe" viewer can return **real credentials to the browser**. Both Codex (reproduce) and I independently confirmed leaks against adversarial configs. A config-safety viewer that leaks secrets is worse than none (false sense of safety), so this must be airtight before it ships.

**Leaks reproduced (verified, not theoretical):**
1. **`${VAR}` expansion (highest severity):** `_safe_config_yaml_text()` builds from `get_config()`, whose cache **expands `${VAR}` to the real value**. So `endpoint: ${SAFE_VIEWER_ENDPOINT}` / `api_key: ${OPENAI_API_KEY}` come back as the **expanded real secret**, not the raw placeholder. A user who safely parameterized their config via env vars gets the actual secret shown.
2. **URL/DSN userinfo under a non-sensitive key:** I reproduced `proxy: "http://admin:P4ssw0rd@10.0.0.1:8080"` returning the password **verbatim** (the key "proxy" isn't in the sensitive-fragment list and the value-scrub doesn't strip URL userinfo). Codex reproduced the same with `proxy:` and `callback_url: https://...?token=QUERYSECRET987` (query-param token).

(Credit where due: the path-based redaction is genuinely strong — it caught `api_key`, inline DSN creds under `dsn:`, `webhook.url`, and even a free-text `sk-ant-...` in a `notes:` field. The gap is specifically (a) expansion and (b) URL userinfo / query tokens in arbitrarily-named string values.)

**Fix-spec (Codex + my repro agree):**
1. Build the viewer from **raw active-profile YAML** — `_load_yaml_config_file_raw(_get_config_path())`, NOT `get_config()` — so `${VAR}` placeholders are shown literally and never expanded into real secrets.
2. Add an **unconditional** config-specific string scrubber (independent of `api_redact_enabled`) that masks **URL/DSN userinfo** (`scheme://user:pass@` → `scheme://user:***@`) and **sensitive query-param values** (`?token=`, `?key=`, `?secret=`, etc.) on every scalar string regardless of its key name.
3. Add failing-then-passing regression tests for: `http://user:pass@host`, `https://host/path?token=SECRET`, and `endpoint: ${SECRET_URL}` — each proving **no secret substring** reaches `/api/config/safe`.

The endpoint is correctly read-only / auth-gated / profile-scoped (Codex confirmed) — the only blocker is the redaction boundary. Credited to original author of #3228. Re-gate on re-push; I'll re-run the same adversarial repro.

nesquena-hermes:
## ✅ Gate certification — GREEN (self-built security fix → independent review + maintainer merge trigger)

I bounced this PR as a **security blocker** (the "safe" config viewer leaked real credentials), then applied the fix myself on `fix/5088-safe-config-leak` (head `e5d71551`) and hardened it through **6 Codex security rounds + an Opus last-mile review**. Every round reproduced a real leak, fixed it, and added a regression.

### The credential-leak vectors closed (each Codex-reproduced)
1. **`${VAR}` expansion** — `_safe_config_yaml_text()` now reads `_load_yaml_config_file_raw(_get_config_path())`, never the env-expanded `get_config()`; fails closed to `{}` (no fallback to expanded).
2. **URL/DSN userinfo, all forms** — `user:pass@`→`user:***@`, empty-username `:pass@` (redis), token-as-username `token@`→`***@`.
3. **Query + fragment params** (`?`/`&`/`;`/`#`) for a full credential vocabulary, including **percent-encoded** delimiters (`%3Ftoken%3D`).
4. **Capability-URL path tokens** — Slack `hooks.slack.com/services/...`, Discord `api/webhooks/...`.
5. **Bare `key:`/`token:`/`auth:` segments** via exact-segment matching (without over-masking `record_key`).
6. **YAML aliases** (`&pw`/`*pw`) — sensitive-value taint redacts a secret wherever it appears.
7. **Unconditional redaction** — `_redact_text(..., _enabled=True)` so the viewer NEVER defers to the operator's `api_redact_enabled` setting; i18n/index notes corrected.

### Opus last-mile review verdict: **SHIP — security boundary COMPLETE and safe**
Opus verified no credential-leak path is introduced; all six rounds' contracts hold; profile scoping is sound (via the per-request TLS profile resolver). It flagged one **functional MUST-FIX** (the pre-existing bare `token`/`auth` substring fragments over-masked `max_tokens`/`token_budget`/`author` — defeating the viewer's "copy your config" purpose) — **fixed**: dropped bare `token`/`auth` from substrings, added a `_BENIGN_CONFIG_SEGMENTS` allowlist + survival tests. SHOULD-FIX (taint sentinel over-masking) also applied (≥12-char threshold + sentinel skip). Documented residual: an opaque high-entropy secret under a totally non-conventional benign key with no URL shape is the inherent heuristic limit (pre-existing class, low exposure since Hermes secrets live under `api_key`/`token`/`${ENV}`).

### Verification
- 25 viewer tests (incl. all leak-vector + over-mask + sentinel regressions) pass.
- **Full suite 11100 passed** (only the 2 pre-existing cron-isolation flakes — unrelated).
- Ruff forward-gate clean.

**Status:** engineering-GREEN, security boundary Opus-verified complete. Agent-authored salvage of #3228 → routes to **independent review (nesquena account) + @nesquena's merge trigger** (self-built + security — never auto-merged). Original deserted-PR author credited. Branch: `fix/5088-safe-config-leak` @ `e5d71551`.

nesquena-hermes:
## 🔬 Security gate (Codex, adversarial — rebased clean off master) — SHIP ONLY WITH FIXES (verified credential + profile-isolation leaks)

Thanks @AJV20 — the read-only config viewer is a genuinely useful salvage of #3228, and the structure is right (server-side redaction, `textContent` render so no HTML injection, copy returns the displayed text). But adversarial probes found **credential values of string, integer, AND boolean type can currently reach the browser**, plus a cross-profile config bleed — so on this security surface it can't ship yet. Five fixes:

1. **[SECURITY — key-match bypass leaks credentials] `api/routes.py:444`** — the sensitive-key match isn't separator-insensitive / Unicode-normalized and mishandles custom-provider bare keys, so casing/hyphen/underscore variants (and some custom-provider secret keys) slip past redaction and the raw value reaches the client. **Fix:** separator-insensitive + Unicode-normalized key matching incl. custom-provider bare-key handling; add adversarial string/integer/boolean redaction tests.

2. **[SECURITY — YAML anchor/alias reproduces a masked secret] `api/routes.py:469`** — a YAML anchor/alias can reference a credential so the alias renders the secret unredacted elsewhere in the output. **Fix:** propagate sensitivity across anchor/alias references so no alias reproduces a masked credential.

3. **[SECURITY — cross-profile config bleed] `api/routes.py:485`** — the endpoint reads process-global `get_config()`, so under concurrency one profile's request can receive another profile's config (reproduced: an `alpha` read returned `beta`'s owner). On a config *viewer* this is a profile-isolation leak. **Fix:** read one raw, path-bound active-profile snapshot directly (not the process-global cache); add a concurrent two-profile regression test.

4. **[CORE — stale-profile content in the browser] `static/panels.js:12994`** — loaded text + in-flight responses aren't bound to `S.activeProfile`, so switching profiles can show/copy the previous profile's config. **Fix:** bind loaded text + in-flight responses to the active profile, invalidate on switch, refresh copy on mismatch.

5. **[test] `tests/test_safe_config_viewer.py:309`** — a stale changelog assertion now fails against master, and the suite covers ordinary nesting/types/lists but not casing variants, aliases, real HTTP-method dispatch, auth, or concurrent profile ownership. **Fix:** drop the stale changelog assertion; add focused casing/alias/method/auth/profile-isolation tests.

Net: the redaction is close for ordinary nesting/types, but the casing/alias bypasses + the process-global read (cross-profile bleed) mean real credentials + another profile's config can reach the browser. Those are the ship-blockers on a security surface. Rebased clean on current master (I resolved a CHANGELOG + a ru-locale i18n conflict, keeping master's translations + the new `safe_config_*` keys). Credit @AJV20 for the salvage.

PR #5101, opened 2026-06-28
feat(extensions): TTS-engine registration capability (registerHermesTtsEngine)

nesquena-hermes:
## Summary

Adds a **TTS-engine registration capability** so an extension can contribute a text-to-speech engine — the voice-side sibling of `registerHermesSkin`. This is the core dependency for a VOICEVOX (and any future) TTS-engine extension.

```javascript
window.registerHermesTtsEngine({
  id: 'voicevox',                 // [a-z0-9_-], not a built-in
  label: 'VOICEVOX (local)',
  synthesize(text, opts) { return Promise<ArrayBuffer|Blob>; }  // audio bytes
});
```

A registered engine appears in the **Settings → TTS Engine** dropdown alongside the built-ins (Browser / Edge / ElevenLabs) and is used by **both** playback paths — the hands-free voice-mode auto-read (`_speakResponse`) and the per-message **Listen** button (`speakMessage`).

## Why

The TTS dispatch was a hardcoded `if(engine==='elevenlabs')/if(engine==='edge')` chain in two places, the engine `<select>` is hardcoded in `index.html`, and the speak functions are module-scoped — so an extension had no way to add an engine. This adds the hook (same pattern as theme registration).

## Design / security

- Core owns selection, the dropdown `<option>` (label inserted via **`textContent`, never HTML**), and playback (registered engines play through the same `<audio>` lifecycle as Edge, including stop/rearm in voice mode).
- Built-in ids (`browser`/`edge`/`elevenlabs`) are **reserved** — an extension can't shadow them. Ids are slug-validated; descriptors without a `synthesize` fn are rejected; an empty/failed synth result degrades gracefully (toast on Listen / re-listen in voice mode).
- The extension only produces audio bytes; it never touches playback internals.
- The settings panel re-adds registered options on render (the markup hardcodes only built-ins).

## What changed
- `static/boot.js`: the `registerHermesTtsEngine` registry + `_hermesTtsSynth` / `_hermesTtsIsRegistered` / `_hermesTtsEngineOptions`; voice-mode path routes registered engines.
- `static/ui.js`: the per-message Listen path routes registered engines through the shared `_playAudioBuf`.
- `static/panels.js`: settings TTS dropdown re-adds registered options.
- `docs/EXTENSIONS.md`: "Registering a custom TTS engine" section.
- `tests/test_extension_tts_engine_registration.py`: structural + a behavioral Node harness (register / reserved-guard / bad-id / missing-synth / options list / ArrayBuffer coercion / unregistered→null).
- `CHANGELOG.md`: `[Unreleased] → Added`.

## Verification
- New test: **6/6 pass**.
- **Full suite: 10,893 passed, 0 failed** — breaks nothing across the voice/TTS/everything-else tests.
- **End-to-end in a live browser** with the companion VOICEVOX extension (extensions PR): the engine registers, appears in the dropdown (`browser/edge/elevenlabs/voicevox`), and clicking Listen with it selected routes through the registered `synthesize` (both VOICEVOX steps called) and plays via the core `<audio>` path.

## Follow-up
The VOICEVOX extension that consumes this is a separate PR in the extensions repo.

🤖 Agent-authored; flagging for review, not self-merging.


nesquena (approved):
## Review — end-to-end ✅ (clean approval; two non-blocking polish notes)

Independent review of `registerHermesTtsEngine` — the voice-side sibling of `registerHermesSkin` (#5083, which I reviewed). Lets a trusted extension contribute a TTS engine into Settings → TTS Engine, used by both playback paths (voice-mode auto-read + per-message Listen).

### What this ships

`static/boot.js` (+100), `static/ui.js` (+20), `static/panels.js` (+12), `docs/EXTENSIONS.md` (+40), `tests/test_extension_tts_engine_registration.py` (+108, new), `CHANGELOG.md` (+2). Agent-authored, flagged for review. Additive — no change for users with no TTS extension; built-in engines (browser/edge/elevenlabs) untouched.

Fresh tarball: WebUI voice/TTS frontend only. The extension produces audio bytes client-side; no agent/protocol surface. ✓

### Security review — no XSS, clean dispatch

- **Label → dropdown via `textContent`** in both `_hermesAddTtsOption` (boot.js) and the panels.js re-add — never `innerHTML`, so an extension label can't inject markup. The option **value** is the `id`, slug-validated `^[a-z0-9][a-z0-9_-]{0,31}$`, so it's also safe inside the `option[value="…"]` existence-check selector. ✓
- **id guard:** slug-validated + reserved built-ins (`browser`/`edge`/`elevenlabs`) rejected — an extension can't shadow a core engine. Missing/non-function `synthesize` → rejected. ✓
- **Audio coercion + graceful degradation:** `_hermesTtsSynth` resolves `synthesize` then coerces ArrayBuffer / Blob (`.arrayBuffer()`) / typed-array (`.buffer`) → ArrayBuffer, throwing on empty/unsupported. The bytes play via `new Blob([buf])` → `URL.createObjectURL` → `<audio>`, reusing the `_playingEdgeAudio` slot so the existing voice-mode stop/rearm governs it; the Listen path routes through the shared `_playAudioBuf`. A failed/empty synth degrades cleanly (voice-mode re-listen / Listen toast), never throwing into the extension or wedging `_ttsSpeaking`. The blob URL is `revokeObjectURL`'d on every terminal path (onended / onerror / play-catch). ✓
- The audio bytes are inert (decoded by `<audio>`, not script); and an extension already runs with full session authority — so this is a clean, least-privilege API, not a new code-exec surface. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Register `{id:'voicevox', label, synthesize}` | appears in dropdown; both paths route to it ✅ |
| Reserved id (`edge`/`browser`/`elevenlabs`) | rejected ✅ |
| Bad id (non-slug) / missing `synthesize` | rejected ✅ |
| `synthesize` returns Blob / typed-array / ArrayBuffer | coerced + played ✅ |
| `synthesize` returns empty / throws / rejects | graceful (toast / re-listen), URL revoked, `_ttsSpeaking` reset ✅ |
| Label with HTML/`"`/`<` | rendered as text (textContent); no injection ✅ |
| Settings panel renders after registration | option re-added (no duplicate) ✅ |
| Built-in engines | unchanged ✅ |

### Tests

- `tests/test_extension_tts_engine_registration.py` (6) — structural + Node harness (register / reserved-guard / bad-id / missing-synth / options list / ArrayBuffer coercion / unregistered→null). **6/6.**
- TTS / voice / speak / extension blast-radius — **210 passed, 2 skipped**.
- **Full suite: 10755 passed / 143 skipped / 0 failures** in 222s (deselected the pre-existing darwin CRLF flake).
- `node --check` on boot.js/ui.js/panels.js clean; ESLint runtime guard exit 0; CI lint + browser-smoke + 9 test shards all green.

- **NaN opts on a fresh profile (Greptile):** both paths build `opts.rate`/`opts.pitch` via `parseFloat(localStorage.getItem('hermes-tts-rate'))`, which is `NaN` when the key has never been saved — so the extension's `synthesize(text, {rate:NaN, pitch:NaN})` gets NaN on a fresh profile. It doesn't affect built-in engines or core, and the first consumer (VOICEVOX) works end-to-end, but the documented opts contract should hand extensions clean values. Cheap follow-up: only include each key when `Number.isFinite(parsed)` (omit otherwise so the engine uses its default). Below my push bar (not a present defect, built-ins unaffected) — flagging for a polish pass.
- **Re-register label drift (Greptile):** re-registering an id updates `_HERMES_TTS_ENGINES[id]` but `_hermesAddTtsOption` early-returns when the option already exists, so the dropdown label isn't refreshed. Harmless today (no live TTS editor); worth updating the option's `textContent` on re-register if that use case arrives.
- **CHANGELOG flag is a false positive for this repo:** Greptile says the PR shouldn't touch CHANGELOG, but AGENTS.md requires `[Unreleased]` entries and merged PRs here consistently add them (the release agent promotes them to a versioned section). The entry is correct — keep it.

### Recommendation

A clean, well-isolated capability that mirrors the established `registerHermesSkin` pattern: core owns selection, the dropdown option (textContent label, slug id, reserved-built-in guard), and playback (shared `<audio>`/`_playAudioBuf` lifecycle with graceful degradation and blob-URL cleanup); the extension only produces audio bytes. No XSS, no new code-exec surface, built-in engines untouched. The two Greptile findings (NaN opts on a fresh profile, re-register label drift) are minor API-polish for the extension contract — flagged for a follow-up, neither blocking. Ship.

PR #5106, opened 2026-06-28
feat(search): jump to matched message + flash on content-hit (salvage of #4224)

### LINKED ISSUE #4159: feat(search): navigate to matched message + highlight keyword when clicking session search result

nesquena-hermes:
Salvage of #4224 by @Sanjays2402. Closes #4159.

Clicking a session-search result that matched on message **content** now jumps straight to that message in the transcript and flashes it, instead of opening the session at the latest message and leaving the user to scroll.

### What it does
- **Backend** (`api/routes.py`): the content scan in `_handle_sessions_search` already locates the exact matching message (it `break`s on the first hit) but threw the loop index away. We now capture it as `match_message_idx` on content-typed results, enumerated against the same `sess.messages` array the renderer stamps onto rows as `msg-user-<rawIdx>` / `data-msg-idx`. Title matches don't grow the field (nothing to jump to).
- **Frontend** (`static/sessions.js`): after `loadSession()` in the sidebar-row click handler, a content hit carrying an integer `match_message_idx` dispatches the existing jump-and-flash helper on the next tick. Purely additive — a missing idx falls back to opening at the latest message, so it's backward compatible with older servers.

### The scope fix (the bug that sank the original #4224)
The maintainer root-caused why the original PR's jump never fired: `_jumpToMessage` is defined **inside** `static/outline.js`'s IIFE (the file ends `})();`), so a bare `_jumpToMessage(...)` call from `static/sessions.js` lands in a different `<script>` scope and silently resolves to nothing — a no-op.

The minimal clean fix follows the existing convention in the same file: outline.js already exports `window._outlineJump = _jumpToMessage` for inline `onclick` handlers. This PR adds a sibling `window._jumpToMessage = _jumpToMessage` export and calls it via `window._jumpToMessage(idx)` from sessions.js, so the jump actually fires across the script boundary. No new scroll mechanism — it reuses outline.js's battle-tested helper that handles both the in-render-window case (smooth `scrollIntoView` + flash) and the out-of-window case (refetch the full session with `msg_limit=9999`, expand the window, then retry). XSS-safe (no innerHTML; flash is a self-terminating CSS class).

### Tests — `tests/test_issue4159_search_match_message_idx.py`
- **4 backend tests** pin `match_message_idx`: right raw index, first-hit semantics (preserves the existing `break`), title matches omit the field, no-match returns empty.
- **3 structural tests** pin the cross-`<script>` wiring (the salvage's whole point): outline.js exposes the helper on `window`, the search-click path calls `window._jumpToMessage`, and a regression guard rejects any bare `_jumpToMessage(` call. The two positive structural assertions **fail on master** (verified by running them against `origin/master`'s JS), proving they're load-bearing rather than vacuous.

### Verification
- `node --check` clean on both touched JS files.
- Targeted suite (`-k 'search or jump or outline or session_search'`): 141 passed, 3 skipped.
- Full suite: **10814 passed**, 4 failed — all 4 are the known order-dependent flakes (`test_scheduled_jobs_profile_isolation` ×2, `test_tls_support` TLS-fallback, `test_v050259_sessiondb_fd_leak` idempotent), unrelated to this change.

Credit to @Sanjays2402 for the original implementation and design.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -13897,10 +13897,14 @@ def _handle_sessions_search(handler, parsed):
             try:
                 sess = get_session(s["session_id"])
                 msgs = sess.messages[:depth] if depth else sess.messages
-                for m in msgs:
+                for raw_idx, m in enumerate(msgs):
                     c = _session_search_message_text(m)
                     if q in str(c).lower():
-                        item = dict(s, match_type="content")
+                        # #4159: surface the matched message index so the frontend
+                        # can jump to it (msg-user-<rawIdx>). The renderer stamps
+                        # rows with the same raw index it gets from sess.messages,
+                        # so this enumerate lines up with the DOM ids.
+                        item = dict(s, match_type="content", match_message_idx=raw_idx)
                         preview = _session_search_preview(c, q)
                         if preview:
                             item["match_preview"] = _redact_text(preview)
--- a/static/outline.js
+++ b/static/outline.js
@@ -227,6 +227,11 @@ function toggleOutlinePanel() {
 
 // Jump target exposed on window so inline onclick handlers can reach it.
 window._outlineJump = _jumpToMessage;
+// Also expose under its own name so OTHER scripts (e.g. sessions.js's
+// search-result click handler, #4159) can reuse the jump-and-flash helper
+// across the <script> boundary — _jumpToMessage is otherwise trapped inside
+// this IIFE and unreachable from sessions.js.
+window._jumpToMessage = _jumpToMessage;
 window.applyConversationOutlinePreference = applyConversationOutlinePreference;
 
 // Re-render after renderMessages() if the panel is open and the session
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -7180,6 +7180,17 @@ function renderSessionListFromCache(){
         try{
           if(($('sessionSearch').value||'').trim()) _hideSearchPreviewsAfterSelect=true;
           await loadSession(s.session_id);renderSessionListFromCache();
+          // #4159: if the user clicked a content-search hit, scroll the
+          // newly-loaded transcript to the matched message and flash it.
+          // window._jumpToMessage (exposed by outline.js) handles both the
+          // in-window case (smooth scrollIntoView + flash) and the
+          // out-of-render-window case (refetch full session w/ msg_limit=9999,
+          // expand the window, then retry). A missing idx falls back to opening
+          // at the latest message, so this stays backward compatible.
+          if(s.match_type==='content' && Number.isInteger(s.match_message_idx) && typeof window._jumpToMessage==='function'){
+            const _jumpIdx=s.match_message_idx;
+            window.setTimeout(()=>{ try{ window._jumpToMessage(_jumpIdx); }catch(_e){} }, 0);
+          }
           if(typeof closeMobileSidebar==='function')closeMobileSidebar();
         }finally{
           el.classList.remove('loading');
--- a/tests/test_issue4159_search_match_message_idx.py
+++ b/tests/test_issue4159_search_match_message_idx.py
@@ -31,7 +31,6 @@ from types import SimpleNamespace
 from unittest.mock import patch
 from urllib.parse import urlparse
 
-import pytest
 
 _REPO = Path(__file__).resolve().parent.parent
 _OUTLINE_JS = _REPO / "static" / "outline.js"
--- a/tests/test_issue4159_search_match_message_idx.py
+++ b/tests/test_issue4159_search_match_message_idx.py
@@ -1,0 +1,166 @@
+"""Regression for #4159 (salvage of PR #4224): clicking a session-search result
+that matched on message CONTENT must jump to that message in the transcript and
+flash it.
+
+Two halves:
+
+1. Backend — the content scan in ``_handle_sessions_search`` already locates the
+   exact message that contains the query (it iterates ``sess.messages`` and
+   ``break``s on the first hit). Until now the loop index was discarded; these
+   tests pin the new ``match_message_idx`` field on content-typed results,
+   indexed against the same raw ``sess.messages`` array the renderer stamps onto
+   each row as ``msg-user-<rawIdx>`` / ``data-msg-idx``. Title matches must NOT
+   carry ``match_message_idx`` (there's no message-level hit to jump to).
+
+2. Frontend scope wiring — the jump helper ``_jumpToMessage`` is defined INSIDE
+   ``static/outline.js``'s IIFE (the file ends ``})();``), so a bare
+   ``_jumpToMessage(...)`` call from ``static/sessions.js`` is a different
+   ``<script>`` scope and resolves to nothing — a silent no-op (the bug that
+   sank the original #4224). These structural tests pin the fix: outline.js must
+   expose the helper on ``window`` and the sessions.js search-click path must
+   reach it via ``window._jumpToMessage`` (the cross-<script> handle), exactly
+   like the sibling ``window._outlineJump`` export. They fail on master, where
+   neither the export under that name nor the call site exists.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import patch
+from urllib.parse import urlparse
+
+import pytest
+
+_REPO = Path(__file__).resolve().parent.parent
+_OUTLINE_JS = _REPO / "static" / "outline.js"
+_SESSIONS_JS = _REPO / "static" / "sessions.js"
+
+
+# --------------------------------------------------------------------------- #
+# Backend: match_message_idx is surfaced on content hits
+# --------------------------------------------------------------------------- #
+def _run_search(query, *, session_messages, sessions_meta=None):
+    import api.routes as routes
+
+    meta = sessions_meta or [
+        {"session_id": "s1", "title": "Untitled", "profile": "default"}
+    ]
+    session = SimpleNamespace(session_id="s1", messages=session_messages)
+    captured = {}
+
+    def fake_j(handler, payload, status=200, extra_headers=None):
+        captured["status"] = status
+        captured["payload"] = payload
+
+    with patch("api.routes.all_sessions", return_value=list(meta)), patch(
+        "api.routes.get_session", return_value=session
+    ), patch("api.profiles.get_active_profile_name", return_value="default"), patch(
+        "api.routes.j", side_effect=fake_j
+    ):
+        routes._handle_sessions_search(SimpleNamespace(), urlparse(query))
+    return captured
+
+
+def test_content_match_includes_message_index():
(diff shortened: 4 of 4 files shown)


nesquena-hermes:
**Pre-processed → decision-ready (gate-clean). Parking for maintainer sign-off (visible search-interaction change).**

Status: salvage of #4224 (closes #4159), credit @Sanjays2402. Now **MERGEABLE / CI 11/11 green** after I fixed a lint blocker (dropped an unused `pytest` import). The cross-script scope bug from the original (`_jumpToMessage` trapped in outline.js's IIFE, unreachable from sessions.js) is fixed so the jump actually fires.

What it does: clicking a session-search result that matched on message **content** jumps to that message in the transcript and flashes it (self-terminating highlight, XSS-safe). Earlier gate confirmed the flash is well-built; full suite green.

It's a visible search-interaction enhancement, so it's your sign-off (best eyeballed: search content, click a content-hit, confirm the jump+flash lands on the right message). **Recommendation: clean + useful — recommend ship** on your trigger. Gate-clean and ready to merge whenever you say.

nesquena (approved):

Independent review of the salvage of #4224 (@Sanjays2402, closes #4159): clicking a session-search result that matched on message **content** now jumps to that exact message and flashes it, instead of opening at the latest message and leaving the user to scroll.

### What this ships

`api/routes.py` (+~6 — capture the matched index in `_handle_sessions_search`), `static/outline.js` (+5 — export the jump helper), `static/sessions.js` (+11 — call it on a content hit), `tests/test_issue4159_search_match_message_idx.py` (+new, 7 tests). Agent-authored (`nesquena-hermes`), faithful port of #4224 with the scope-bug fix that sank the original. MERGEABLE. Purely additive.

Fresh tarball: WebUI-only. The search handler reads `get_session(...).messages` (existing webui session state); no agent/protocol surface, no `config.yaml`. ✓

### End-to-end trace

1. **Backend index capture is correctly aligned.** `_handle_sessions_search` already located the first content match and `break`'d, but discarded the loop index. It now `enumerate(msgs)` where `msgs = sess.messages[:depth]` — a **prefix slice**, so `raw_idx` equals the absolute index into `sess.messages`, which is exactly what the renderer stamps as `data-msg-idx` / `msg-user-<rawIdx>` (`ui.js:1101`, queried at `ui.js:826` etc.). First-hit semantics are preserved (the `break` stays). Title matches don't get the field (nothing to jump to). ✓
2. **The scope fix is the right minimal one.** `_jumpToMessage` lives inside `outline.js`'s IIFE, so a bare call from `sessions.js` resolved to nothing (the bug that no-op'd #4224). The PR adds `window._jumpToMessage = _jumpToMessage` alongside the pre-existing `window._outlineJump` export — same convention — and calls `window._jumpToMessage(idx)` across the `<script>` boundary. No new scroll mechanism; it reuses outline's helper that handles in-window (`scrollIntoView` + flash) and out-of-window (refetch full session `msg_limit=9999`, expand, retry). ✓
3. **Caller is correctly guarded** (`sessions.js`): `if (s.match_type==='content' && Number.isInteger(s.match_message_idx) && typeof window._jumpToMessage==='function')`, fired via `setTimeout(…,0)` after `loadSession`, wrapped in `try/catch`. A missing/non-integer idx → no jump, opens at latest → backward-compatible with older servers. ✓

### Security / robustness

- **No XSS:** `match_message_idx` is a server-side integer; the frontend `Number.isInteger`-guards it and uses it only in a `[data-msg-idx="${idx}"]` selector and a numeric arg — no `innerHTML`, and an integer can't break the selector. The flash is a self-terminating CSS class. ✓
- **Redaction intact:** the preview still routes through `_redact_text`; only an integer index is added to the result. ✓
- Backward/forward compatible both directions (old server → no field → fallback; old client → ignores field). ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Content hit, message in render window | smooth scroll + flash ✅ |
| Content hit, message outside window | refetch full session, expand, retry ✅ |
| Title-only match | no `match_message_idx`, opens at latest ✅ |
| Old server (no field) | `Number.isInteger` false → fallback ✅ |
| Multiple matches in a session | first hit (preserves `break`) ✅ |
| No match | empty result, no field ✅ |

### Tests

- `tests/test_issue4159_search_match_message_idx.py` — **7/7**: 4 backend (right raw index, first-hit, title omits field, no-match empty) + 3 structural (outline.js `window` export, search-click calls `window._jumpToMessage`, regression guard rejecting bare `_jumpToMessage(`). The two positive structural assertions are verified to **fail on master**, so they're load-bearing, not vacuous.
- `node --check` clean on both touched JS files.
- Search/session blast-radius — **86 passed, 4 skipped**. (Targeted rather than a full-suite run, proportionate to a one-field additive backend change + two JS export/caller lines; the PR author reports the full suite at 10814 passed with only the known order-dependent flakes failing.)

### Recommendation

A small, well-targeted UX fix. The backend index lines up with the renderer's `data-msg-idx` (prefix-slice enumerate → absolute index), the cross-`<script>` export is the right minimal fix and follows the existing `window._outlineJump` convention, the caller is defensively guarded and backward-compatible, and it reuses the battle-tested jump-and-flash helper (no new scroll path, XSS-safe). Structural tests are mutation-proven against master. Ship.

nesquena-hermes:
Gated this fresh (Codex reproduce + full suite + live browser drive). The feature works in a fully-loaded session — I drove it live and the jump-to-message + amber flash is correct (screenshot captured) — **but the gate found a real SILENT correctness bug in long/truncated sessions**, so one fix is needed before it ships.

**The bug (reproduced):** the backend returns a **full-session** index (`match_message_idx` over `sess.messages`, api/routes.py:~13900), but `loadSession()` initially loads only a **tail window** (`msg_limit=30`, sessions.js:~2257) and `renderMessages()` stamps DOM ids from the **local** `S.messages` index offset by `_oldestIdx` (`msg-user-${rawIdx}`, ui.js:1111 / `_oldestIdx` ui.js:775, set from `_messages_offset` ui.js:7729). So `_jumpToMessage(fullIdx)` does a raw `getElementById('msg-user-' + fullIdx)` and, in a truncated session, **finds the wrong local message** (or none) — it flashes the wrong turn. (It looks correct only when the whole transcript fits in the initial window — e.g. a short session — which is why the PR's tests and a small manual check don't catch it.)

**Fix-spec (Codex, sound):**
1. `static/outline.js:~94` — make `_jumpToMessage(rawIdx)` treat `rawIdx` as a **full-session coordinate**. Before the `document.getElementById('msg-user-'+rawIdx)` lookup, if the session is truncated (`_messagesTruncated` or `_oldestIdx > 0`), force-load full history (e.g. `_ensureAllMessagesLoaded()`), render, *then* resolve the target — or translate full→local via `_oldestIdx`.
2. `static/outline.js:~109` — replace the fallback `/api/session?...&msg_limit=9999` fetch with a real full-history load (`msg_limit=9999` can still return an offset tail, so the same off-by-offset bug persists for very large sessions).
3. Add a regression with a **truncated** session (full index beyond the initial tail window) asserting the flashed element is the matched message, not the local-Nth.

Concept + non-truncated behavior are good (and credited to original author @Sanjays2402 / #4224, closes #4159) — this is the long-session coordinate fix away from green. Re-gate on re-push.

nesquena-hermes:
## ✅ Gate certification — GREEN (self-built salvage, visible → independent review + maintainer trigger)

I bounced this for a SILENT correctness bug (content-search jump flashed the WRONG message in truncated sessions), then fixed it myself on `fix/5106-search-jump-fullidx` (head `aba79cfe`) and hardened it through **4 Codex re-gate rounds**, each reproducing a real additional edge:

1. **Full-session vs local-tail-offset coordinate bug** (the original bounce) — `match_message_idx` is a full-session index but `msg-user-<idx>` is a local render-window id; in a truncated session it resolved the wrong row. Fixed: new `_jumpToFullSessionMessage(fullIdx, targetSid)` translates full→local via `_oldestIdx` and force-loads full history when truncated.
2. **Assistant-message hits** — the helper only resolved `msg-user-*` (user rows); assistant content hits never jumped. Fixed: also resolve `.assistant-segment[data-msg-idx]`.
3. **Session-switch race** — a rapid A→B switch could apply A's match index to B. Fixed: capture `targetSid`, re-check `S.session.session_id === sid` before scrolling.
4. **Virtualized + >9999-row sessions** — `renderMessages` still virtualizes after load, and `msg_limit=9999` is a tail window (a near-start match in a >9999-row session stayed out of range). Fixed: delegate to the proven virtualization-aware `jumpToTurnQuestion` (materializes the target via `_messageVisibleIndexForRawIdx` + virtual scrollTop) after an **uncapped** `_ensureAllMessagesLoaded()`.

### Verdict + verification
- **Codex (reproduce): SAFE TO SHIP** — correct for all session sizes, both message roles, no wrong-session.
- 11 search-jump regression tests (full→local translation, force-load, assistant-hit, session-guard, virtualization delegation, no-raw-full-id) pass.
- **Full suite 11086 passed** (only the 2 pre-existing cron-isolation flakes — unrelated). Ruff forward-gate clean.

**Status:** engineering-GREEN. Self-built salvage of deserted #4224 (closes #4159) with a visible jump+flash surface → routes to **independent review (nesquena account) + @nesquena's merge trigger** (not auto-merged). Original author **@Sanjays2402** credited. Branch: `fix/5106-search-jump-fullidx` @ `aba79cfe`.

PR #4033, opened 2026-06-12
feat: support manifest-driven plugin iframe embeds

ai-ag2026:
## Summary
- let dashboard plugin manifests opt into iframe bootstrap payload injection via `webui.inline_payload`
- let manifests opt into self-contained iframe HTML by inlining declared dist CSS/JS via `webui.inline_dist_assets`
- rebase default `style.css` / `index.js` references to authenticated dashboard-plugin asset routes without special-casing any plugin name

## Contract Routing
Task type: dashboard plugin serving enhancement
Touched areas: plugin iframe page route (`/dashboard-plugins/<name>/page`) and plugin panel static guards
Scope boundaries: manifest-driven, read-only embedding for sandboxed plugin pages; no Project Cockpit/TARS-specific endpoint or core Insights UI changes included.

## Test plan
- `python -m pytest tests/test_plugins_panel.py -q -o addopts=` → 33 passed
- `python -m py_compile api/routes.py`
- `git diff --check origin/master...HEAD`
- added-line private/secret marker scan → 0

PR #5110, opened 2026-06-28
feat(plugins): manifest-driven iframe embeds (salvage of #4033)

nesquena-hermes:
## Salvage of #4033 — manifest-driven plugin iframe embeds

Faithful port of #4033 (credit: @ai-ag2026) freshly rebased onto today's `master`. The original was reviewed clean — maintainer **verdict: "No blockers. Containment, CSP, and the manifest-driven contract are all sound."** This PR ports the change without redesign.

> Note: **plugins** (dashboard plugin system) are a distinct core system from **extensions** — these embeds apply only to the dashboard plugin iframe path.

### What it does
Lets a dashboard plugin's `manifest.json` opt into, via a `webui` block:
- **(a) iframe bootstrap-payload injection** — `webui.inline_payload` ({`global`, `candidates`}) inlines a read-only JSON payload as a `<script>GLOBAL=...;</script>` before the plugin's first script.
- **(b) self-contained iframe HTML** — `webui.inline_dist_assets: true` inlines the manifest-declared dist CSS/JS directly into the served HTML (so a sandboxed `null`-origin iframe needs no extra fetches).

Behavior is **entirely manifest-driven** — no plugin name is special-cased.

### Preserved security properties (all 4 maintainer-verified, intact & tested)
1. **Symlink-escape containment** — `.resolve()` + `relative_to()` on both `_dashboard_path` and `_PLUGIN_STATIC_ROOTS[name]` (stored resolved). Payload/CSS/JS candidate paths that escape the dashboard root are rejected.
2. **CSP unchanged** — iframe sandbox `allow-scripts allow-forms allow-popups` + `X-Content-Type-Options: nosniff` headers left fully untouched; inlining does **not** loosen the sandbox.
3. **Breakout escaping** — payload `json.dumps(...).replace("</","<\/")`, JS `.replace("</script","<\/script")`, CSS `.replace("</style","<\/style")` — all three retained.
4. **Re-injection guard** — `_inject_plugin_payload` keys on an existing `<script ...>GLOBAL=` regex, so a *commented* occurrence of the global does **not** suppress injection (covered by test).

### Touched
- `api/routes.py` — plugin-page block (helpers `_safe_js_global`, `_plugin_inline_payload`, `_inline_declared_dist_assets`, `_inject_plugin_payload`; applied to all three serve paths).
- `tests/test_plugins_panel.py` — behavioral embed test (incl. symlink branch + commented-payload re-injection) and a manifest-driven (not name-special-cased) guard test.

### Verification
- Targeted: `-k 'plugin and (embed or iframe or inline or payload or manifest)'` → **7 passed, 1 skipped**.
- Full suite: **10815 passed, 92 skipped, 1 xfailed, 2 xpassed**; the only 4 failures are pre-existing known flakes (scheduled-jobs profile isolation ×2, TLS startup fallback, sessiondb-fd-leak idempotent) unrelated to this change.
- `ast.parse` clean on both touched files. No `static/*.js` touched.

Co-authored-by: ai-ag2026 <ai-ag2026@users.noreply.github.com>

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -10954,10 +10954,141 @@ def handle_get(handler, parsed) -> bool:
                 return False
             dashboard_dir = _PLUGIN_STATIC_ROOTS.get(name)
             if dashboard_dir:
+                dashboard_path = dashboard_dir.resolve()
+
+                webui_manifest = manifest.get("webui", {}) if isinstance(manifest.get("webui", {}), dict) else {}
+
+                def _safe_js_global(value: str) -> str | None:
+                    """Return a safe dotted JS global name from manifest config."""
+                    parts = str(value or "").split(".")
+                    if not parts:
+                        return None
+                    if all(re.match(r"^[A-Za-z_$][A-Za-z0-9_$]*$", part or "") for part in parts):
+                        return ".".join(parts)
+                    return None
+
+                def _plugin_inline_payload(
+                    _webui_manifest=webui_manifest,
+                    _dashboard_path=dashboard_path,
+                ) -> tuple[str, str | None]:
+                    """Return manifest-declared read-only bootstrap payload script."""
+                    payload_cfg = _webui_manifest.get("inline_payload", {})
+                    if not isinstance(payload_cfg, dict):
+                        return "", None
+                    global_name = _safe_js_global(payload_cfg.get("global"))
+                    candidates = payload_cfg.get("candidates", [])
+                    if not global_name or not isinstance(candidates, list):
+                        return "", global_name
+                    for rel in candidates:
+                        if not isinstance(rel, str):
+                            continue
+                        payload_path = (_dashboard_path / rel).resolve()
+                        try:
+                            payload_path.relative_to(_dashboard_path)
+                        except ValueError:
+                            continue
+                        if not payload_path.is_file():
+                            continue
+                        try:
+                            payload = json.loads(payload_path.read_text(encoding="utf-8"))
+                            encoded = json.dumps(payload, ensure_ascii=False).replace("</", "<\\/")
+                            return f"<script>{global_name}={encoded};</script>\n", global_name
+                        except Exception:
+                            logger.debug("Failed to embed dashboard plugin payload from %s", payload_path, exc_info=True)
+                    return "", global_name
+
+                def _inline_declared_dist_assets(
+                    html_text: str,
+                    _webui_manifest=webui_manifest,
+                    _manifest=manifest,
+                    _dashboard_path=dashboard_path,
+                    _name=name,
+                ) -> str:
+                    """Inline manifest-approved dist CSS/JS for sandboxed WebUI plugin iframes."""
+                    if _webui_manifest.get("inline_dist_assets") is not True:
+                        return html_text
+                    css_rel = _manifest.get("css") or "dist/style.css"
+                    entry_rel = _manifest.get("entry") or "dist/index.js"
+                    if isinstance(css_rel, str):
+                        css_path = (_dashboard_path / css_rel).resolve()
+                        try:
+                            css_path.relative_to(_dashboard_path)
+                        except ValueError:
+                            css_path = None
+                        if css_path and css_path.is_file():
+                            style_text = css_path.read_text(encoding="utf-8").replace("</style", "<\\/style")
+                            html_text = html_text.replace(
+                                '<link rel="stylesheet" href="style.css">',
+                                f"<style>{style_text}</style>",
+                                1,
+                            )
+                            html_text = html_text.replace(
+                                f'<link rel="stylesheet" href="/{_name}/style.css">',
+                                f"<style>{style_text}</style>",
+                                1,
+                            )
+                    if isinstance(entry_rel, str):
+                        script_path = (_dashboard_path / entry_rel).resolve()
+                        try:
+                            script_path.relative_to(_dashboard_path)
+                        except ValueError:
+                            script_path = None
+                        if script_path and script_path.is_file():
+                            script_text = script_path.read_text(encoding="utf-8").replace("</script", "<\\/script")
+                            inline_bundle = f"  <script>{script_text}</script>"
+                            script_marker = '<script src="index.js"></script>'
+                            if script_marker in html_text:
+                                return html_text.replace(script_marker, inline_bundle, 1)
+                            if "</body>" in html_text:
+                                return html_text.replace("</body>", inline_bundle + "</body>", 1)
+                            return html_text + inline_bundle
+                    return html_text
+
+                def _inject_plugin_payload(
+                    html_text: str,
+                    _webui_manifest=webui_manifest,
+                    _name=name,
+                ) -> str:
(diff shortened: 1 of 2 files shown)


nesquena (approved):

Independent review of the salvage of #4033 (@ai-ag2026): **manifest-driven plugin iframe embeds** for the dashboard-plugin system. A plugin's `manifest.json` can opt into (a) inlining a read-only JSON bootstrap payload as `<script>GLOBAL=…;</script>`, and (b) inlining its declared dist CSS/JS into the served HTML so a sandboxed `null`-origin iframe needs no extra fetches. Entirely manifest-driven — no plugin name special-cased. I re-verified the 4 maintainer-noted security properties against the actual code.

### What this ships

`api/routes.py` (+~140 — four nested helpers in the plugin-page block, applied to all three serve paths), `tests/test_plugins_panel.py` (+~80). Agent-authored (`nesquena-hermes`), faithful port of the maintainer-reviewed-clean #4033. No `static/*.js` touched.

Fresh tarball: WebUI-only, dashboard-plugin serve path. No agent/protocol surface, no `config.yaml`. The inlined content is the plugin's own manifest-declared dist assets. ✓

### Security audit — the 4 preserved properties, re-verified

1. **Symlink/`..`-escape containment holds.** Every candidate is `(_dashboard_path / rel).resolve()` then `.relative_to(_dashboard_path)`; a `ValueError` → `continue` (payload) / `None` (css/js). Because `.resolve()` follows symlinks **and** `_dashboard_path` is itself resolved, the comparison is real-path vs real-path — so both `../escape` and a symlink pointing outside the dashboard root are rejected. `is_file()` gates each read. ✓
2. **CSP / sandbox unchanged.** All three serve paths still send `Content-Security-Policy: sandbox allow-scripts allow-forms allow-popups` and `X-Content-Type-Options: nosniff` verbatim (`routes.py:10940` + per-path) — the diff alters only the response **body**, never the headers. The opaque-`null`-origin sandbox remains the real containment boundary; inlining does not loosen it. ✓
3. **HTML-breakout escaping present.** Payload: `json.dumps(payload, ensure_ascii=False).replace("</", "<\\/")` (escapes *every* `</`, so case-robust); CSS: `.replace("</style", "<\\/style")`; JS: `.replace("</script", "<\\/script")`. The payload is `json.dumps`'d so `<\/` stays valid JS. ✓
4. **Re-injection guard works.** `_inject_plugin_payload` keys on `re.compile(rf"<script\b[^>]*>[^<]*{re.escape(global)}\s*=", re.I|re.S)` — a *commented* (non-`<script>`) occurrence of the global doesn't match, so injection still proceeds (test-covered); a genuine prior `<script>GLOBAL=` suppresses the duplicate. ✓

**Attribute-context safety:** `name.replace('"', '')` strips the quote before interpolating the plugin name into `href="/dashboard-plugins/<name>/…"`; `<`/`>` are inert inside a double-quoted attribute value, so quote-stripping is sufficient to prevent attribute breakout. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Payload candidate `../../etc/passwd` | `relative_to` raises → skipped ✅ |
| Payload candidate via symlink out of root | `.resolve()` escapes root → skipped ✅ |
| `inline_dist_assets` not `true` | assets rebased to authenticated route, not inlined ✅ |
| Payload JSON contains `</script>` | escaped to `<\/script>`, no breakout ✅ |
| CSS contains `</style>` | escaped to `<\/style>` ✅ |
| Global already assigned in a real `<script>` | injection suppressed (no dup) ✅ |
| Global only in an HTML comment | injection still proceeds ✅ |
| Manifest `global` not a valid JS identifier | `_safe_js_global` returns None → no payload ✅ |
| No manifest `webui` block | unchanged legacy behavior ✅ |

### Tests

- `tests/test_plugins_panel.py` — **36 passed** (embed behavior incl. symlink branch + commented-payload re-injection + manifest-driven/not-name-special-cased guard, plus the existing CSP/sandbox assertions).
- **Full suite: 10767 passed / 143 skipped / 0 failed** in 230s (deselected the known darwin CRLF flake). The 4 failures the PR flagged as pre-existing flakes did **not** reproduce here.

- **Tag-escaping is case-sensitive** (`</style` / `</script`), while HTML end-tag matching is case-insensitive — a plugin emitting `</STYLE>`/`</SCRIPT>` in its own dist asset wouldn't be escaped. **Not exploitable** here: a breakout lands in the plugin's *own* sandboxed `allow-scripts` iframe (no privilege gained over its own JS), the plugin is operator-installed, and the payload escaping (`</`→`<\/`) is already case-robust. A case-insensitive regex would be a tidy hardening of the dist-asset path. Matches the original #4033 escaping, which was reviewed clean.
- `_inject_plugin_payload` uses `name.replace('"','')` while the third serve path uses `html.escape(name)` — both safe in context; minor consistency nit.
- `re.match(r"^[A-Za-z_$]…")` for `_safe_js_global` correctly rejects names with dots-of-empty-parts and non-identifier chars. Good.

### Recommendation

A faithful, manifest-driven port of a maintainer-reviewed-clean change. I independently re-verified all four security properties: path containment (resolve + relative_to defeats `..` and symlink escapes), the untouched sandbox CSP (the real boundary), breakout escaping on all three inline channels, and the re-injection guard. The one nit (case-sensitive tag escaping) is non-exploitable given the sandbox and trust model. Full suite clean. Ship.

nesquena-hermes:
Gated this fresh (Codex reproduce + full suite). The security posture is sound — Codex found **no** sandbox-escape, SSRF, CSP-weakening, or manifest-injection (the `resolve()+relative_to()` containment, the `sandbox allow-scripts allow-forms allow-popups` CSP header, and the `</style`/`</script` escaping all hold). But it found **one real contract-completeness gap** before this can ship:

**The issue (SILENT — inline contract silently fails on one path):** the fallback generated-shell path (case 3 in the `/dashboard-plugins/` handler, where a plugin ships only `dashboard/dist/index.js` with no `dist/index.html` or `static/index.html`) emits **external** tags unconditionally:
```
<link rel="stylesheet" href="/dashboard-plugins/{name}/{css}">
<script src="/dashboard-plugins/{name}/dist/index.js"></script>
```
It never checks `webui.inline_dist_assets`. So a plugin that sets `inline_dist_assets: true` but has only `dist/index.js` still gets external asset references instead of self-contained inline CSS/JS — the manifest's inline contract is honored on the index.html path (the `_inline_declared_dist_assets` branch) but silently dropped on this fallback. That's a production path for IIFE-bundle plugins.

**Fix-spec (Codex, sound):** when `webui.inline_dist_assets is True`, the fallback generated shell must **inline** the manifest-declared `css`/`entry` assets instead of emitting `/dashboard-plugins/{name}/...` tags — reuse/extend `_inline_declared_dist_assets`, keeping the existing `resolve()+relative_to()` containment and `</style`/`</script` escaping. Add a test covering the fallback case (no `dist/index.html`, no `static/index.html`) with `inline_dist_assets: true`.

Everything else is clean (36 plugins-panel tests + full suite green). Concept + containment are solid — credited to original author **@ai-ag2026** (#4033). This is the inline-on-fallback fix away from green. Re-gate on re-push.

PR #4153, opened 2026-06-14
api/routes: claim CLI/TUI/Desktop sessions on first POST, not just GET

merodahero:
Restore POST-path session claim for CLI/TUI/Desktop sources; gate bare gateway/unknown; map state.db timestamps; sanitize 500s.

## Why

CLI, TUI, and Desktop sessions were being recreated on every POST to `/api/chat/start` instead of being claimed from the on-disk state, so the WebUI lost context between requests. The denylist for `gateway` / `unknown` sources, the 500-error sanitization, and the `state.db` timestamp mapping are Greptile-driven hardening around the same claim path.

## What changed

- `api/routes.py`
  - `_claim_or_synthesize_cli_session` (~L3355) now claims local CLI/TUI/Desktop sessions on the first POST to `/api/chat/start`, not just on the GET path.
  - New source denylist gate (~L3309-3320) rejects bare `gateway` and `unknown` sources at the claim path; they have no claimable backing store and would otherwise 500.
  - 500 responses from the claim path are sanitized before they go out as JSON, so internal exception text does not leak to the WebUI.
  - CLI meta is now looked up once per POST (dedupe).
  - `state.db` `created_at` / `updated_at` are mapped into the session metadata response (was previously missing on synthesized CLI/TUI/Desktop sessions).
  - CLI-meta enrichment is copy-on-write: it builds a new dict instead of mutating the caller's `cli_meta`.
- `tests/test_chat_start_claim_cli_session.py` (new, 29 tests, 973 lines)
  - Source denylist: `test_helper_refuses_bare_gateway_session`, `test_helper_refuses_bare_unknown_session`.
  - No-mutation guards: `test_helper_does_not_mutate_callers_cli_meta`, `test_helper_does_not_mutate_callers_cli_meta_when_empty`.
  - `read_only` plumbing: GET stub + POST 403 path.
  - `state.db` timestamp mapping.
  - Original POST-claim flow.
- `CHANGELOG.md`
  - Unreleased entry under `[Unreleased]` -> `### Fixed`.

## Testing

- 29/29 in `tests/test_chat_start_claim_cli_session.py`.
- Greptile 5/5 (no findings) on the prior #3901 commit set. The branch was reauthored after a force-push, so a fresh Greptile review is expected to land on the new SHAs; the previous summary is stale.

## Risk / rollback

The change is gated by the source denylist, so it only activates on write-claimable local sources (CLI/TUI/Desktop). WebUI-originated sessions are unaffected. Rollback = revert the merge commit.

[...]

Issue #2782, opened 2026-05-23
GET /api/session returns 200 for missing sessions but POST 404s — UI stuck with no self-heal path


rickeyfaneite:
## Summary

GET `/api/session?session_id=<missing>` returns HTTP 200 (with a tolerant payload) while POST `/api/session/draft` and POST `/api/chat/start` return 404 "Session not found" for the same missing session. This inconsistency prevents the studio UI from self-healing after a session is deleted server-side — the browser polls GET, sees 200, and never triggers the localStorage-clear / redirect-to-root branch. The user is stuck on a dead session with no in-app recovery path.

Compounding it: `_sessionIdFromLocation()` in `static/sessions.js` reads the session id from the URL path (`/session/<id>`) and uses it ahead of `localStorage['hermes-webui-session']` in `static/boot.js:1605-1607`. So clearing localStorage doesn't help either — the URL keeps re-injecting the dead id on every reload.

## Repro

1. Open studio, create a session, note the URL path (`/session/abc123`).
2. Server-side, delete the session file (or rebuild the container so the in-container state dir resets):
   ```
   docker compose up -d --force-recreate hermes-webui
   ```
3. Reload the studio tab on the same URL.
4. Try to send a chat.

Expected: UI realises the session is gone and redirects to a working state (empty + button, or restore from `_index.json`).

Actual: UI shows "Error: Session not found" on every send. Network tab: GET `/api/session?session_id=abc123` → 200; POST `/api/session/draft` and `/api/chat/start` → 404.

## Server-side code

- `api/routes.py` GET `/api/session` handler (around line 3695) — wraps `get_session(sid)` in a try/except but returns 200 with tolerant content rather than 404.
- `api/routes.py` POST `/api/session/draft` (around line 4920) and `/api/chat/start` (around line 5312) — `get_session()` KeyError → 404 "Session not found".

## Client-side code

- `static/boot.js:1605-1607` — `const saved = urlSession || savedLocal;` URL wins over localStorage unconditionally.
- `static/sessions.js:568` — there's a self-heal branch that does `localStorage.removeItem('hermes-webui-session')` on a 404, but it's reached only under narrow conditions and not when the URL forces the bad id.

## Suggested fix

Either of the following would close the loop. Both would be ideal:

1. Server: make GET `/api/session` return 404 (consistent with POST) when the session_id is unknown. This lets the existing client-side self-heal branch trigger on the next poll.
[...]

PR #5111, opened 2026-06-28
fix(routes): claim CLI/TUI/Desktop sessions on POST, not just GET (salvage of #4153)

nesquena-hermes:
Salvage of #4153 by @merodahero (closed; conflicted with master — ported fresh onto today's `api/routes.py`).

## The bug (GET/POST asymmetry)

CLI/TUI/Desktop sessions were **recreated on every POST** to `/api/chat/start` instead of being **claimed**, because the claim/synthesis logic only ran on the GET stub path.

- `GET /api/session` already synthesized a read-only stub from `state.db` for CLI/TUI/Desktop sessions.
- `POST /api/chat/start` unconditionally `404`'d on any `session_id` without a WebUI JSON sidecar.
- The frontend's `404` handler then ran its empty-state self-heal — stripping the `/session/<id>` URL and clearing localStorage — so the user's typed message vanished into a fresh "new session" screen.

## The fix (shared helper)

Both endpoints now route through a new shared helper, **`_claim_or_synthesize_cli_session`**, which materialises a WebUI-owned `Session` on first write (via `synth.save()`). A second helper, **`_is_claimable_cli_source`**, applies a denylist that decides whether a foreign-origin session may be claimed writeable.

This makes GET (read-only stub) and POST (writeable claim) symmetric for local CLI/TUI/Desktop sessions, while keeping foreign-owned conversations read-only.

## Preserved guards (maintainer-verified clean)

- **Denylist** (`_is_claimable_cli_source`): `{claude_code, cron, external_agent, gateway, messaging, unknown}` are refused on **both** the `cli_meta` source-tag branch and the `state.db` source branch. `cli` / `tui` / `desktop` fall through to **claimable**.
- **`not_claimable` ⇒ `403`, not `404`**, and it **never reaches `synth.save()`** (ownership-boundary guard). `403`-vs-`404` is load-bearing: the frontend `404` handler runs the empty-state self-heal that strips the URL, so a legitimately-listed read-only session must surface a refusal instead of vanishing.
- **`created_at` mapping from `state.db`** is preserved — `Session.save()` does not touch `created_at`, so without the mapping a claimed sidecar would persist epoch (`0`) and sort as "Jan 1 1970".
- **500-path uses `_sanitize_error`** so a raw `OSError` from a failed sidecar save can't leak the absolute session-store filesystem path; the full exception is still `logger.exception`'d server-side.
- **#2782 deleted-WebUI-session `404` self-heal** contract preserved via `_session_index_marks_was_webui`.
- The GET path's **`_session_visible_to_active_profile`** profile gate is preserved (master-specific; not in the original PR base).
- GET response and `import_cli` refresh both read `read_only` from the synthesized/persisted `Session` (not from `cli_meta` directly), so source-refused sessions correctly advertise `read_only=True`.

## 3 polish items applied over the original

1. **Test docstring rationale corrected re: `updated_at`.** `Session.save()` defaults `touch_updated_at=True`, so the claimed sidecar's `updated_at` is stamped to "now" regardless of `state.db`; only `created_at` is preserved. Assertions were **not** weakened (`updated_at > 0` still pins that the mapping ran).
2. **`touch_updated_at=True` default kept.** A claimed session jumping to "just now" is the desired UX; no `touch_updated_at=False` was added, and no test pins persisted `updated_at` to the `state.db` value.
3. **Path-leak assertion added** to the 500-sanitisation test: the sanitized string must not contain a `/<session-store-root>/` segment, so a future refactor of `_sanitize_error` can't silently regress path-stripping.

## Testing

- New `tests/test_chat_start_claim_cli_session.py`: **32 passed**.
- Targeted `-k 'cli_session or claim or chat_start or not_claimable or session_source'`: **167 passed, 1 skipped**.
- Full suite: **10845 passed, 92 skipped, 1 xfailed, 2 xpassed**. The 4 failures under full-suite parallel load are all pre-existing known flakes (cron profile-context x2, TLS-fallback subprocess env, sessiondb-fd-leak idempotent) and **pass in isolation**; none touch the session-claim path.

Credit: @merodahero (`Co-authored-by` on the commit).


nesquena-hermes:
Reviewed the salvage against `origin/master` by pulling the PR head into a read-only worktree and reading `_is_claimable_cli_source`, `_claim_or_synthesize_cli_session`, and both call sites. The core design is clean: the denylist refuses `{claude_code, cron, external_agent, gateway, messaging, unknown}` on both the `cli_meta` source-tag branch and the `state.db` source branch (`api/routes.py:5961` and `:5990`), and the POST handler maps `not_claimable → 403`, `was_webui`/`no_foreign_state`/`invalid_sid → 404`, and only calls `synth.save()` on the `materialized` path with a sanitized 500 fallback (`api/routes.py:17810`–`17828`). The 403-vs-404 reasoning is load-bearing and correctly preserved.

## One asymmetry the salvage leaves open: profile visibility on POST

The GET no-sidecar branch checks profile visibility *before* synthesizing:

```python
# api/routes.py:10509-10513  (GET path)
cli_meta = _lookup_cli_session_metadata(sid)
_session_profile = (cli_meta or {}).get("profile") or None
if not _session_visible_to_active_profile(_session_profile, handler):
    return bad(handler, "Session not found", 404)
synth, reason = _claim_or_synthesize_cli_session(sid, cli_meta=cli_meta or {})
```

But the POST claim path calls the helper with no preceding visibility gate:

```python
# api/routes.py:17802  (_handle_chat_start)
synth, reason = _claim_or_synthesize_cli_session(body["session_id"])
```

`_claim_or_synthesize_cli_session` itself does not check `_session_visible_to_active_profile` — it gates on `is_safe_session_id`, `_session_index_marks_was_webui`, recoverable messages, and the source denylist, but never profile (`api/routes.py:6088`–`6171`). The `validate_profile` stage further down (`api/routes.py:17857`–`17876`) only reconciles the *body's* `requested_profile` against `s.profile`; when the claimed session has persisted turns it leaves `s.profile` untouched and proceeds. So nothing 404s a cross-profile session on the POST path.

Net effect: a CLI/TUI/Desktop session that GET `/api/session` hides from the active profile (404) can still be claimed and persisted as a writable WebUI sidecar by POSTing its `session_id` to `/api/chat/start`. That's the same GET/POST asymmetry this PR sets out to remove, just running the other direction — and it crosses the profile-isolation boundary that the four existing `_session_visible_to_active_profile` call sites (`4123`, `10193`, `10511`, `21051`) are there to enforce.

## Suggested fix

Mirror the GET guard before the claim in `_handle_chat_start`:

```python
cli_meta = _lookup_cli_session_metadata(body["session_id"])
_session_profile = (cli_meta or {}).get("profile") or None
if not _session_visible_to_active_profile(_session_profile, handler):
    return bad(handler, "Session not found", 404)
synth, reason = _claim_or_synthesize_cli_session(body["session_id"], cli_meta=cli_meta or {})
```

Passing the already-computed `cli_meta` also avoids the helper's redundant `_lookup_cli_session_metadata` call, matching how the GET path threads it through.

## Test coverage

`tests/test_chat_start_claim_cli_session.py` is thorough on the source-denylist and 403/404 split, but I didn't see a case asserting that a claimable CLI session belonging to a *non-active profile* is refused (404) on POST. Worth adding alongside the existing claim tests — it's the one branch that distinguishes "claim what GET shows" from "claim anything with recoverable state.db messages."

Everything else in the diff looks correct to me.

PR #5115, opened 2026-06-28
feat(plugins): subprocess-isolated plugin API backend (salvage of #3645)

### LINKED ISSUE #3383: RFC: Subprocess-isolated plugin backend API bridge
## Summary

Add a plugin backend API system where each plugin runs in its own subprocess, communicating with WebUI via stdin/stdout JSON. This replaces the need for standalone HTTP servers per plugin and avoids the security risks of running plugin code in-process.

## Motivation

From PR #2622 discussion:

1. **No viable in-process sandbox** — CPython has no reliable in-process sandbox (`ctypes` bypasses everything). Running third-party plugin code via `exec_module()` is unsafe for public plugin use.
2. **Standalone servers are heavy** — port allocation, process lifecycle, CORS boilerplate per plugin adds unnecessary complexity.
3. **Interest from reviewers** — both nesquena-hermes and AJV20 agree subprocess isolation is the right direction.

## Proposed Design

```
Plugin JS (sandboxed iframe, from PR #2622)
  │  fetch('/api/plugins/<name>/<sub_route>')
  ▼
WebUI (proxy layer)
  │  serialize request → JSON, write to subprocess stdin
  ▼
Plugin subprocess (isolated, resource-limited)
  │  run handler, serialize response → JSON, write to stdout
  ▼
WebUI wraps JSON → HTTP response
```

Key properties:
- stdin/stdout JSON-line protocol (one line per request/response)
- Each plugin gets its own subprocess (crash isolation)
- Resource limits via `setrlimit` (512MB RAM, 5min CPU, max subprocesses, file descriptors)
- Minimal environment passed to plugin (no API tokens)
- `CompatHandler` mimics `BaseHTTPRequestHandler` so existing plugin code needs minimal changes
- Subprocess-isolated mode is the only mode — no in-process fallback, no feature flag

## Guardrails (from review discussion)

1. **Plugin API contract via `register()`** — each plugin exposes a `register()` function in its `__init__.py` that returns a route map (`{path: {method: handler}}`). No manifest-based backend config needed; plugins are self-describing.
2. **Minimal env** — explicit passthrough only; no ambient secrets
3. **Bounded protocol** — structured request/response with id, method, path, query, headers, body (base64), timeout, error shape
4. **Lifecycle ownership** — WebUI owns start/stop/restart/health; plugin is a sidecar, not an extension
5. **Response header filtering** — WebUI strips security-sensitive response headers (set-cookie, CSP, x-frame-options, etc.) from plugin responses
6. **Failure isolation tests** — hung process, invalid JSON, crash, oversized response, stderr leaks

## Scope

- Full GET/POST proxy for plugin backend API at `/api/plugins/<name>/<sub_route>`
- CRSF exemption for sandboxed iframe origin (null)
- Dev mode via `HERMES_WEBUI_PLUGIN_DIRS` for local plugin development
- Body size limit (1MB), pipe synchronization, ready-signal handshake
- `allow-modals` in iframe sandbox for `confirm()`/`alert()` in plugin UI
- Graceful shutdown of subprocesses on WebUI exit

## Related

- PR #2622: Dashboard plugin system with iframe isolation
- Discussion: https://github.com/nesquena/hermes-webui/pull/2622#issuecomment-4561187464

nesquena-hermes:
## Summary

Salvage of #3645 (by @pix0127), **ported fresh onto current `master`**.

Each plugin that ships an `__init__.py` now runs in its **own subprocess**, talking to WebUI over a stdin/stdout JSON-line protocol. WebUI proxies plugin API traffic under the unified prefix `/api/plugins/<plugin_name>/<sub_route>` to that subprocess. (`plugins` is a distinct core system from `extensions`.)

Per the maintainer's verdict on #3645 — *"Solid iteration, #1 (stdout/JSON-pipe corruption) was the must-fix and it's handled."* — this PR **preserves the verified-correct stdout-redirect fix** and **applies the two remaining fixes** (response-id correlation + CSRF narrowing).

## Preserved: stdout-redirect fix (maintainer-verified)

`api/plugin_runner.py` dup()s the real stdout to a protected fd (`out`) and redirects fd 1 → stderr **before** loading any plugin code. Every protocol frame — the `ready` signal, per-request responses, and error frames (`err()`) — goes through the single `send()` write path on the protected fd, while any stray plugin `print()` lands on the stderr reader thread in `plugin_manager._read_stderr`. This is what eliminates JSON-pipe corruption.

## Fix #2 — response-id correlation

**Problem:** `_dispatch_plugin_subprocess` built a request `id` but read one line back and `json.loads`'d it **without** correlating the id. The per-plugin pipe lock serializes the happy path and the timeout path kills+504s, but a slow handler that wrote a frame just after a previous request's 30s deadline elapsed could leave an **orphan frame** in the pipe that maps onto the *next* request.

**Fix:** the dispatcher now loops, reading frames until `resp["id"] == req["id"]`, skipping a **bounded** number of stale frames (`MAX_FRAME_SKIPS`). A persistent read buffer (`_make_line_reader`) handles the case where the plugin coalesces a stale frame and the real response into a single `os.read()` — bytes after the first newline are not discarded. If no frame matches after the bound, it returns **502** rather than hang or return another request's body. The id round-trips through the runner already (it echoes `req["id"]` into every response/error frame). The single shared deadline keeps the whole exchange under `PIPE_TIMEOUT`.

## Fix #3 — CSRF narrowed to `Origin: null` (security)

**Problem:** the earlier approach blanket-exempted the entire `/api/plugins/` prefix from CSRF. The rationale was that the sandboxed iframe sends `Origin: null` (which `_check_csrf` rejects), but a blanket prefix exemption lets a **state-changing plugin POST ride through cross-origin from any page** with the user's cookie when auth is ON.

**Fix:** `_is_plugin_request` now exempts **only** the sandboxed-iframe case — a `/api/plugins/` path carrying **both** the `X-Plugin-Request` marker **and** `Origin: null`. A cross-origin attacker page carries a concrete `Origin` (e.g. `https://evil.example`), never matches, and still falls through to `_check_csrf`, which rejects it. (Cross-origin cookie riding is independently impossible — the server never sends `Access-Control-Allow-Credentials`.)

## Tests (non-vacuous — verified by mutation)

New, in `tests/test_plugin_subprocess_isolation.py`:

- **`test_dispatch_skips_stale_frame_and_correlates_response_id`** — a stale wrong-id frame is queued ahead of the real response; the dispatcher must return the id-matched body (`{"ok":1}`), not the stale one. Removing the id check (return-first-frame mutant) → **fails**.
- **`test_dispatch_502_when_no_frame_matches_request_id`** — a flood of wrong-id frames must yield **502**, not a mismatched body. Return-first-frame mutant → **fails**.
- **`test_is_plugin_request_only_exempts_null_origin`** — exemption is True only for `Origin: null` + marker on a plugin path; a concrete cross-origin `Origin` is **not** exempt.
- **`test_csrf_gate_rejects_cross_origin_plugin_post_allows_null_origin`** — integration through `handle_post`: a cross-origin plugin POST is **403**'d; an `Origin: null` plugin POST reaches dispatch. Reverting to a blanket-prefix exemption → **fails**.

Mutation-tested both fixes: breaking fix #2 fails both #2 tests; breaking fix #3 (blanket prefix) fails both #3 tests.

Also ported the PR's protocol/env/dispatch isolation tests plus the plugin-page CSP and API enable-gate assertions in `tests/test_plugins_panel.py`.

## Test plan

- Targeted: `-k 'plugin and (subprocess or runner or dispatch or csrf or isolat or correlat)'` → **13 passed, 1 skipped**.
- Full suite → **10826 passed, 92 skipped, 1 xfailed, 2 xpassed**; the only 4 failures are pre-existing environmental flakes (cron profile-context ×2, TLS-fallback, sessiondb fd-leak idempotent) that also fail on clean `master` and pass in isolation.

Closes #3383

Co-authored-by: @pix0127

### DIFF
--- a/api/plugin_manager.py
+++ b/api/plugin_manager.py
@@ -1,0 +1,182 @@
+"""
+Plugin subprocess lifecycle manager.
+
+    Spawns, monitors, restarts, and kills plugin subprocesses.
+    Each plugin gets its own subprocess with resource limits and
+    a minimal environment.
+"""
+
+import logging
+import os
+import subprocess
+import sys
+import threading
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+PLUGIN_PROCESSES: dict[str, subprocess.Popen] = {}
+_PROCESS_LOCK = threading.Lock()
+_PIPE_LOCKS: dict[str, threading.Lock] = {}
+
+# Environment variable prefixes allowed in plugin subprocess
+_ENV_ALLOWED_PREFIXES = {
+    "HERMES_HOME",
+    "HERMES_PLUGIN_",
+    "HERMES_WEBUI_PLUGIN_",
+    "HOME",
+    "LANG",
+    "LC_",
+    "PATH",
+    "PYTHONPATH",
+    "PYTHONUNBUFFERED",
+    "USER",
+}
+
+
+def _build_plugin_env(plugin_name: str, plugin_dir: Path) -> dict:
+    env = {}
+    for k, v in os.environ.items():
+        for prefix in _ENV_ALLOWED_PREFIXES:
+            if k == prefix or k.startswith(prefix):
+                env[k] = v
+                break
+    env["HERMES_PLUGIN_NAME"] = plugin_name
+    env["HERMES_PLUGIN_DIR"] = str(plugin_dir)
+    env["PYTHONUNBUFFERED"] = "1"
+    return env
+
+
+def get_plugin_pipe_lock(plugin_name: str) -> threading.Lock:
+    with _PROCESS_LOCK:
+        if plugin_name not in _PIPE_LOCKS:
+            _PIPE_LOCKS[plugin_name] = threading.Lock()
+        return _PIPE_LOCKS[plugin_name]
+
+
+def _read_stderr(proc: subprocess.Popen, plugin_name: str):
+    if proc.stderr is None:
+        return
+    try:
+        for line in proc.stderr:
+            line_str = line.decode("utf-8", errors="replace")[:2000]  # cap per-line
+            line_str = line_str.rstrip()
+            if line_str:
+                logger.warning("[plugin:%s] %s", plugin_name, line_str)
+    except Exception:
+        pass
+
+
+def spawn_plugin(plugin_name: str, plugin_dir: Path) -> subprocess.Popen | None:
+    """Spawn a plugin subprocess if not already running.
+
+    Args:
+        plugin_name: e.g. 'my-plugin'
+        plugin_dir: Path to plugin root (contains __init__.py)
+
+    Returns the subprocess.Popen handle, or None on failure.
+    """
+    plugin_dir = plugin_dir.resolve()
+    runner = Path(__file__).parent / "plugin_runner.py"
+
+    with _PROCESS_LOCK:
+        existing = PLUGIN_PROCESSES.get(plugin_name)
+        if existing and existing.poll() is None:
+            return existing  # already running
+
+        # Kill stale process
+        if existing:
+            try:
+                existing.kill()
+                existing.wait(timeout=2)
+            except Exception:
+                pass
+            PLUGIN_PROCESSES.pop(plugin_name, None)
+
+        env = _build_plugin_env(plugin_name, plugin_dir)
+
+        try:
+            proc = subprocess.Popen(
+                [sys.executable, str(runner)],
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                env=env,
+                text=False,
+            )
+
+            # Wait for ready signal from the subprocess
+            if proc.stdout is not None:
+                try:
+                    import json as _json
+                    import select
+                    ready, _, _ = select.select([proc.stdout], [], [], 10)
+                    if not ready:
+                        logger.warning("Plugin %s: no ready signal within 10s, killing", plugin_name)
+                        proc.kill()
+                        proc.wait()
+                        return None
+                    ready_line = proc.stdout.readline()
+                    if ready_line:
+                        ready_msg = _json.loads(ready_line.decode("utf-8").strip())
+                        if ready_msg.get("ready"):
+                            logger.info(
+                                "Plugin %s ready (pid=%d, routes=%s)",
+                                plugin_name, proc.pid, ready_msg.get("routes", []),
+                            )
+                        else:
+                            logger.warning("Plugin %s: unexpected startup message", plugin_name)
+                            proc.kill(); proc.wait(); return None
+                    else:
+                        logger.warning("Plugin %s: no startup message, process may have exited", plugin_name)
+                        proc.kill(); proc.wait(); return None
+                except Exception:
+                    logger.exception("Plugin %s: startup failed, killing", plugin_name)
+                    proc.kill()
+                    proc.wait()
+                    return None
+
+            PLUGIN_PROCESSES[plugin_name] = proc
+            # Start stderr reader thread
+            threading.Thread(
+                target=_read_stderr, args=(proc, plugin_name), daemon=True,
+            ).start()
+            return proc
+        except Exception:
+            logger.exception("Failed to spawn plugin subprocess: %s", plugin_name)
+            return None
+
+
+def get_plugin_process(plugin_name: str) -> subprocess.Popen | None:
+    with _PROCESS_LOCK:
+        proc = PLUGIN_PROCESSES.get(plugin_name)
+    if proc is None:
+        return None
+    if proc.poll() is not None:
+        logger.warning(
+            "Plugin subprocess %s exited (code=%d)", plugin_name, proc.returncode,
+        )
+        with _PROCESS_LOCK:
+            if PLUGIN_PROCESSES.get(plugin_name) is proc:
+                PLUGIN_PROCESSES.pop(plugin_name, None)
+        return None
+    return proc
+
+
+def kill_plugin(plugin_name: str, timeout: float = 5.0):
+    with _PROCESS_LOCK:
+        proc = PLUGIN_PROCESSES.pop(plugin_name, None)
+    if proc is None or proc.poll() is not None:
+        return
+    try:
+        proc.terminate()
+        proc.wait(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        proc.kill()
+        proc.wait()
+    logger.info("Terminated plugin subprocess: %s", plugin_name)
+
+
+def kill_all_plugins():
+    for name in list(PLUGIN_PROCESSES):
+        kill_plugin(name)
--- a/api/plugin_runner.py
+++ b/api/plugin_runner.py
@@ -1,0 +1,234 @@
+"""
+Plugin subprocess runner — stdin/stdout JSON-line protocol.
+
+Runs inside the plugin subprocess. Reads JSON requests from stdin,
+dispatches to registered handlers via CompatHandler, writes JSON
+responses to stdout.
+
+Set by plugin_manager.py at spawn time:
+  HERMES_PLUGIN_NAME=<plugin_name>
+  HERMES_PLUGIN_DIR=/path/to/plugin/root
+
+Protocol:
+  Request (stdin, one JSON line):
+    {"id": int, "method": "GET|POST", "path": "/sub_route",
+     "query": "...", "headers": {...}, "body_b64": "..."}
+
+  Response (stdout, one JSON line):
+    {"id": int, "status": 200, "headers": {...}, "body_b64": "..."}
+
+Lifecycle:
+  - On stdin EOF: exit 0
+  - On handler exception: write error response, continue
+  - On JSON decode error: write error, continue
+"""
+
+import base64
+import json
+import os
+import sys
+import traceback
+from types import SimpleNamespace
+
+
+class CompatHandler:
+    """Accumulates handler.write() calls and serializes to a JSON response."""
+
+    def __init__(self, request_headers: dict, body_bytes: bytes):
+        self._status = 200
+        self._resp_headers: dict[str, str] = {}
+        self._body_parts: list[bytes] = []
+        self._body_bytes = body_bytes
+        self._hdrs = {k.lower(): v for k, v in (request_headers or {}).items()}
+        self._read_pos = 0
+
+    @property
+    def headers(self):
+        return self
+
+    def get(self, key, default=None):
+        return self._hdrs.get(key.lower() if isinstance(key, str) else key, default)
+
+    def send_response(self, status: int):
+        self._status = status
+
+    def send_header(self, key: str, value: str):
+        self._resp_headers[key] = value
+
+    def end_headers(self):
+        pass
+
+    @property
+    def wfile(self):
+        return self
+
+    def write(self, data: bytes | str):
+        if isinstance(data, str):
+            data = data.encode("utf-8")
+        self._body_parts.append(data)
+
+    @property
+    def rfile(self):
+        return self
+
+    def read(self, length: int) -> bytes:
+        remaining = len(self._body_bytes) - self._read_pos
+        if remaining <= 0:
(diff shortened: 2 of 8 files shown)


nesquena (approved):

Independent review of the salvage of #3645 (@pix0127): a **subprocess-isolated plugin API backend**. Each plugin with an `__init__.py` runs in its own subprocess, talking to WebUI over a stdin/stdout JSON-line protocol; WebUI proxies `/api/plugins/<name>/<sub_route>` to it. This is the highest-risk PR in the batch (subprocess execution + a CSRF-exemption surface + a header proxy), so I traced every security boundary.

### What this ships

`api/plugin_runner.py` (+234, new — subprocess side), `api/plugin_manager.py` (+182, new — lifecycle/env), `api/plugins.py` (+36), `api/routes.py` (+263 — `_is_plugin_request`, `_dispatch_plugin_subprocess`, routing), `server.py` (+15), `static/panels.js` (+9), `tests/test_plugin_subprocess_isolation.py` (+377, new), `tests/test_plugins_panel.py` (+35). Agent-authored (`nesquena-hermes`), salvage of #3645 preserving the maintainer-verified stdout-redirect fix and adding response-id correlation + CSRF narrowing. MERGEABLE (no rebase needed).

Fresh tarball: WebUI-only. `plugins` is a distinct core system; no agent/protocol surface, no `config.yaml` round-trip. The subprocess imports only the plugin's own `__init__.py`. ✓

### Security audit — every boundary

**1. Subprocess isolation (`plugin_runner.py`)**
- **stdout-redirect (the maintainer's must-fix) is correct and correctly ordered:** `out = os.fdopen(os.dup(stdout.fileno()))` saves the real stdout to a protected fd, then `os.dup2(stderr.fileno(), stdout.fileno())` (`:161-162`) — **before** `_load_plugin`'s `exec_module` (`:126`). So even import-time plugin `print()` lands on stderr; only `send()` writes protocol frames on the protected fd. JSON-pipe corruption is foreclosed. ✓
- **Resource limits** (`:146-155`): `RLIMIT_AS` 3GB, `NPROC` 64, `NOFILE` 256, `CPU` 300s, each clamped to the existing hard limit, wrapped in try/except. `os.setpgrp()` isolates the process group for clean kills. The 3GB AS comment correctly notes AS ≠ RSS and that cgroups own true RAM limiting. ✓
- **Per-request exception recovery** (`:219-228`): handler exceptions → `err(500)` + traceback to stderr, loop continues; JSON errors → `err(400)`; EOF → exit. A crashing handler never takes down the protocol. ✓

**2. Env filtering (`plugin_manager._build_plugin_env`)** — deny-by-default allowlist of prefixes (`HERMES_HOME`, `HERMES_PLUGIN_`, `HERMES_WEBUI_PLUGIN_`, `HOME`/`LANG`/`LC_`/`PATH`/`PYTHONPATH`/`USER`). API keys, passwords, `HERMES_WEBUI_RESTART_COMMAND`, state-dir, etc. are **not** passed. ✓ (Cross-plugin sharing of `HERMES_PLUGIN_*` is by-design for operator-installed plugins — same trust model as today.)

**3. Request dispatch (`_dispatch_plugin_subprocess`)**
- **Credentials stripped:** the forwarded `req["headers"]` filters out `cookie` and `authorization` — the plugin subprocess never sees session credentials. ✓
- **Body cap:** `MAX_PLUGIN_BODY_BYTES = 1MB` → 413. ✓
- **Response-id correlation:** reads up to `MAX_FRAME_SKIPS+1` (=9) frames, matches `candidate["id"] == req["id"]`, logs+drops stale frames, and on no-match **kills the plugin + returns 502** — never hangs, never returns another request's body. `_make_line_reader` keeps a persistent buffer (coalesced frames aren't lost) and uses `select` with a shrinking deadline so a stalled/partial-line plugin trips `PIPE_TIMEOUT` (30s) → kill + 504. ✓
- **Response header injection foreclosed:** only `{cache-control, content-type, etag, last-modified}` + `x-*` headers pass, and **CRLF is stripped** from both key and value before `send_header`. A plugin can't inject `Set-Cookie`/`Location`/header-split. ✓
- **CORS:** echoes `Access-Control-Allow-Origin: null` for the sandbox case but **never** `Access-Control-Allow-Credentials` → cross-origin credentialed reads are blocked. ✓

**4. CSRF narrowing (`_is_plugin_request`)** — the salvage's key security fix. The exemption fires **only** for `path.startswith("/api/plugins/")` AND `X-Plugin-Request == "1"` AND `Origin == "null"` (`:4366-4370`), wired as `not _csrf_exempt_path and not _is_plugin_request and not _check_csrf` (`:11640`). A cross-origin attacker page carries a concrete `Origin` → never matches → falls through to `_check_csrf` → rejected. Forging the exemption cross-origin is independently impossible: the custom `X-Plugin-Request` header forces a CORS **preflight** the server never approves (no CORS headers), and no `Access-Control-Allow-Credentials` is ever sent. This is strictly safer than the blanket-prefix exemption it replaces. ✓

**5. Routing gates (`routes.py:10251-10262`, POST mirror `:11658`)**
- **Enable-gate before dispatch:** `if not _dashboard_plugin_enabled(plugin_name): return False` (404) — disabled plugins serve nothing. ✓
- **Path traversal:** `sub_route` is unquoted then checked `if ".." in sub_route: return False` (catches `%2e%2e`). ✓
- **Auth:** `/api/plugins/` is not in `PUBLIC_PATHS` → authenticated when auth is on. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Plugin `print()` to stdout | redirected to stderr; protocol uncorrupted ✅ |
| Stale orphan frame ahead of real response | skipped by id; correct body returned ✅ (mutation-tested) |
| Flood of wrong-id frames | 502 after bound, plugin killed ✅ (mutation-tested) |
| Cross-origin plugin POST (concrete Origin) | 403 via `_check_csrf` ✅ (integration-tested) |
| Sandboxed iframe POST (`Origin: null` + marker) | reaches dispatch ✅ |
| Disabled plugin | 404, no subprocess touched ✅ |
| `..` / `%2e%2e` in sub-route | 404 ✅ |
| Body > 1MB | 413 ✅ |
| Plugin stalls / partial line | `PIPE_TIMEOUT` → kill + 504 ✅ |
| Plugin returns header with CRLF | stripped, no injection ✅ |
| Secret env vars | not passed to subprocess ✅ |

### Tests

- `tests/test_plugin_subprocess_isolation.py` + `tests/test_plugins_panel.py` — **43 passed** (protocol happy-path/crash/recovery, env filtering, dispatch lifecycle, id-correlation + 502-on-exhaust both mutation-verified, CSRF narrowing integration through `handle_post`, plugin-page CSP + enable-gate). The lone `ResourceWarning` is a test-fixture fd-close nit (below), not a failure.
- **Full suite: 10824 passed / 143 skipped / 0 failed** in 232s (deselected the known darwin CRLF flake). The 4 failures the PR flagged as pre-existing environmental flakes did **not** reproduce here — consistent with "pass in isolation / on clean master." No regression from this PR.

### Minor observations (non-blocking — Greptile's items + mine)

- **`spawn_plugin` holds `_PROCESS_LOCK` across the ~10s startup `select`**, and **`kill_plugin` runs under the per-plugin pipe lock** on the three error paths (up to ~5s) — both extend lock hold under failure; latency, not correctness. Follow-up.
- **Single-segment sub-route only** (`len(parts) != 5` → silent 404): limits plugin API shape and gives no log line to diagnose. A debug log + a note in plugin docs would help. Follow-up.
- **Test fd `ResourceWarning`** (finally-block pipe double-close) — tidy up the fixture teardown. Test-only.
- **`x-*` response headers pass through** (CRLF-stripped) — harmless for plugin-iframe responses; flagging for awareness.

### Recommendation

A genuinely careful, security-conscious salvage. The must-fix stdout-redirect is preserved and correctly ordered; the two added fixes (id-correlation, CSRF narrowing) are both mutation-tested and hold up under trace; and the dispatch strips credentials, caps the body, allowlists + CRLF-strips response headers, and refuses cross-origin credentialed reads. The non-blocking items are concurrency latency, route expressiveness, and test hygiene — none affecting correctness or security. Full suite clean. Ship.

nesquena-hermes:
Autonomous triage: self-built salvage of #3645 (subprocess-isolated plugin API, credit @pix0127), 1151L backend, non-visible, CONFLICTING. Architectural/security-sensitive (per-plugin subprocess + server.py wiring). Self-built → routes to Nathan's merge trigger + independent review; needs a full security gate + rebase which is a substantial focused pass — flagged decision-pending for Nathan to sequence.

pix0127:
Hi, thanks a lot for the great work on this PR! 🙏

I was wondering if there's any estimated timeline for when this might be merged? We're looking forward to using it, and I'd be happy to help with testing or address any remaining review items if that would speed things up.

No rush at all — just want to make sure it's not blocked on anything. Thanks again!

nesquena-hermes:
Thanks @pix0127. The feature direction remains accepted, and the old exact-head security review was positive. The current blocker is integration state rather than an unanswered code request: the branch now conflicts with `master`, and approval cannot automatically carry across a security-sensitive rebase.

Please rebase without expanding scope. Preserve stdout isolation, response-ID correlation, CSRF narrowing, credential stripping, and the enable gates, then request one fresh exact-head security gate. Testing help after that rebase would be useful.

pix0127:
Integration is ready, but I cannot push it here — I have no write access to `nesquena/hermes-webui`, and this PR's head branch lives in your repo rather than a fork, so there is nothing for me to force-push to.

The integrated branch is **`pix0127:salvage/3645-merge-master`** at `a27d8543` — your head a03d718a with current `master` (320789ae) merged in. To take it:

```
git fetch https://github.com/pix0127/hermes-webui salvage/3645-merge-master
git merge FETCH_HEAD    # fast-forwards this branch
```

No force-push needed, and a03d718a stays in history as the diff anchor your security review approved. I did open it as a PR briefly but withdrew it — based against this branch it rendered as a 2000-commit diff, which buries the two commits that actually matter. The merge commit and the test commit are the whole change.

I merged rather than rebased. Not to dodge the instruction: with the base being your branch, merge is what keeps the integration deliverable at all without write access, and it isolates the security-relevant part into one reviewable merge commit. A rebase would have folded the conflict resolution invisibly into the 1100-line feature commit, which is the opposite of what a fresh exact-head gate needs to look at. Say the word and I will produce the rebased form instead.

Scope held. All five properties preserved with file:line in the linked PR. One conflict (`server.py` shutdown reflow), plus one place where the auto-merge was silently wrong: `master` narrowed OPTIONS into `apply_cors_preflight_headers()`, which would have denied every sandboxed-iframe plugin preflight, so the resolution carves out exactly the `_is_plugin_request` case. That carve-out is new behavior with no coverage in the approved patch, so a second test-only commit pins its edges.

Two things need your decision before this is gate-ready:

1. **The CSRF exemption is forgeable when auth is off.** `X-Plugin-Request` is the constant `"1"` from the shim at `api/routes.py:14024`, and any page can produce `Origin: null` via its own sandboxed iframe — so `_is_plugin_request` is satisfiable by an attacker page. Auth ON, this is free (no Allow-Credentials, so no cookie rides along). Auth OFF, `_check_csrf` returns True at `api/routes.py:5682` and the origin check at 5675 was the only thing rejecting `Origin: null`, so an attacker page can drive a side-effecting POST to an enabled plugin route. Response unreadable, side effect lands. Confined to the plugin API this PR adds. Fix would be a per-plugin nonce instead of the constant marker — a design change, so not done here.

2. **`test_server_py_under_750_lines` fails.** `server.py` is 749 on `master` against `assert lines < 750`; the two lifecycle hooks make it 762, and it will not compress under the limit. Raise the guard's budget, or move both hooks out of `server.py`. I left it failing rather than pick.

Test evidence: 494 failed / 13327 passed on this tree versus 491 / 13316 on clean `master` — the bulk is environmental in my sandbox and reproduces on untouched `master`. Failure-name set comparison shows exactly one branch-unique failure, the line-count guard above. Targeted plugin tests: 18 passed, 4 skipped; the three plugin test files together, 60 passed.

Happy to take the testing further once you have ruled on those two.

PR #5117, opened 2026-06-28
Defer non-critical session switch refreshes

santastabber:
## Summary
- skip no-op empty composer draft POSTs during chat switching
- defer session-visit model dropdown refresh until after the new chat first paints
- defer workspace tree / git badge refresh until after first paint, while preserving explicit `awaitWorkspaceLoad` callers

## Why
Chat switching and new-chat first paint should not wait on non-critical refresh work. The core session metadata/message load and live stream attach remain immediate; slower model/workspace refreshes are guarded by session id and scheduled after paint.

## Verification
- `node -c static/sessions.js`
- `git diff --check`
- `HERMES_WEBUI_TEST_PYTHON=/Users/aiaccount/work/hermes-webui/.venv/bin/python ./scripts/test.sh tests/test_issue4756_session_visit_model_refresh.py tests/test_inflight_stream_reuse.py tests/test_parallel_session_switch.py tests/test_profile_switch_ux.py tests/test_issue_new_chat_draft_restore.py -q`

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -32,6 +32,7 @@ let _pendingCarryForwardSnapshot = null;
 let _draftSaveTimer = null;
 const _DRAFT_SAVE_DELAY_MS = 400;
 const NEW_CHAT_DRAFT_SESSION_KEY = 'hermes-new-chat-draft-session';
+const _composerDraftKnownPayloadSessions = new Set();
 
 function _profileMatchesActiveProfile(profile, activeProfile){
   const eventName = (typeof profile === 'string' && profile.trim()) ? profile.trim() : 'default';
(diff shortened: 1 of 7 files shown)


nesquena-hermes:
Read through `static/sessions.js` at this PR's HEAD plus the diff against `origin/master`, and walked the full `loadSession` control flow (1137–1710) since the model-dropdown refresh moved across several early-return points. The deferral approach is clean and well-guarded — double-`requestAnimationFrame` → `requestIdleCallback(timeout:1500)` in `_afterSessionFirstPaint` (2172), the per-session `S.session.session_id===sid` staleness check in `_deferSessionSideEffect` (2193), and the four-condition empty-draft skip in `_saveComposerDraftNow` (146) all fail safe in the right direction. The draft-payload `Set` (line 35) being eagerly populated and lazily pruned is the conservative choice. One edge case worth a look before merge.

## Code reference

The model-dropdown block was relocated from the head of `loadSession` (it used to fire right after `S.session=data.session`) to the tail at line 1683:

```js
// line 1683
const isActiveModelRefreshSession=()=>!!(S.session&&S.session.session_id===modelRefreshSid);
if(!S._bootReady&&typeof window!=='undefined'&&typeof window._startBootModelDropdown==='function'){
  Promise.resolve().then(()=>{
    if(!isActiveModelRefreshSession()) return undefined;
    return window._startBootModelDropdown();
  }).catch(()=>{});
}else{
  const modelRefreshPromise=_deferSessionSideEffect(modelRefreshSid,()=>{ ... });
```

That new position sits *after* the idle-branch error path at line 1582:

```js
// line 1570-1582
try { await _ensureMessagesLoaded(sid); }
catch (e) {
  // ...renders "Failed to load messages."
  if (_loadingSessionId === sid) _loadingSessionId = null;
  return;            // <-- bails before reaching the model block at 1683
}
```

## Diagnosis

On every *success* path both branches fall through to 1683, and the staleness returns (1512, 1585) correctly *should* skip a superseded switch — so those are fine. The one behavioral delta from master is the message-load **failure** path: master refreshed the dropdown early, so it ran regardless of message-load outcome; this PR skips it. The case that bites is first boot (`!S._bootReady`) — if the initial session's `_ensureMessagesLoaded` throws (server hiccup / large-session timeout), `window._startBootModelDropdown()` never fires and the model dropdown can be left unpopulated until the next successful load.

## Recommendation

Schedule the model-dropdown block *early* (right after `S.session=data.session`, near where master had it) but keep the new `_deferSessionSideEffect` wrapper so the actual work still runs after first paint. The wrapper only needs `S.session.session_id` set, which is already true at that point, and its fire-time guard prevents a stale or double refresh. That preserves the post-paint goal while restoring the "fires regardless of message-load outcome" resilience. A 3-line move of the block above the `try { await _ensureMessagesLoaded(sid) }` is enough; no guard changes needed.

## Tests

The static assertions in `tests/test_issue_new_chat_draft_restore.py` (the `_composerDraftKnownPayloadSessions.add` / `_rememberComposerDraftPayloadState` / `!_composerDraftKnownPayloadSessions.has(sid)` checks) match the source as written, so they'd still pass after the move. If you do relocate the block, consider one assertion that the model-refresh scheduling precedes the `_ensureMessagesLoaded` await so this doesn't regress later.

santastabber:
Updated after the model-refresh edge-case review:

- moved the session-visit model dropdown scheduling back to the early `S.session=data.session` point, before the `_ensureMessagesLoaded()` failure return, while keeping the deferred first-paint `_deferSessionSideEffect` guard
- added a regression assertion that model refresh scheduling happens before message loading so first boot still populates the dropdown even if message loading fails
- rebased on current `origin/master`

Verification:
- `HERMES_WEBUI_TEST_PYTHON=/Users/aiaccount/work/hermes-webui/.venv/bin/python ./scripts/test.sh tests/test_issue4756_session_visit_model_refresh.py tests/test_inflight_stream_reuse.py tests/test_parallel_session_switch.py tests/test_profile_switch_ux.py tests/test_issue_new_chat_draft_restore.py tests/test_run_journal_frontend_static.py -q`

nesquena-hermes:
## ✅ Edge case from the last review is resolved — verified at `93d8907a`

Pulled the updated branch and re-walked `loadSession`. The fix lands exactly where the prior comment recommended: the model-dropdown scheduling is back at the **head** of `loadSession`, right after `S.session=data.session`, and now runs *before* the `_ensureMessagesLoaded` await + its failure return — while keeping the deferral guards intact. The first-boot-with-message-load-failure gap is closed.

## Code reference

`static/sessions.js:1356-1375` — block now sits immediately after the session assignment:

```js
S.session=data.session;
...
if(typeof populateModelDropdown==='function'){
  const modelRefreshSid=sid;
  const isActiveModelRefreshSession=()=>!!(S.session&&S.session.session_id===modelRefreshSid);
  if(!S._bootReady&&typeof window!=='undefined'&&typeof window._startBootModelDropdown==='function'){
    Promise.resolve().then(()=>{
      if(!isActiveModelRefreshSession()) return undefined;
      return window._startBootModelDropdown();
    }).catch(()=>{});
  }else{
    const modelRefreshPromise=_deferSessionSideEffect(modelRefreshSid,()=>{
      if(!isActiveModelRefreshSession()) return undefined;
      return populateModelDropdown({freshness:'session_visit'});
    }).catch(()=>{});
    if(typeof window!=='undefined') window._modelDropdownReady=modelRefreshPromise;
  }
}
```

The `await _ensureMessagesLoaded(sid)` and its `return;` failure bail are now downstream (`sessions.js:1492`, `:1586`), so a message-load throw on first boot no longer skips `window._startBootModelDropdown()`. The staleness check also correctly switched from `_loadingSessionId` to the active-session helper `isActiveModelRefreshSession()` — important, since `loadSession` clears `_loadingSessionId` once first paint completes, which would have made the old guard race.

## Verification

The new regression test pins the ordering directly (`tests/test_issue4756_session_visit_model_refresh.py`):

```python
assert assign_idx < model_block_idx < message_load_idx < failure_return_idx
assert model_block_idx < promise_idx < refresh_idx < ready_idx
assert "_loadingSessionId!==modelRefreshSid" not in body[model_block_idx:ready_idx], (
    "deferred model refresh must guard on the active session, not _loadingSessionId, ...")
```

That `not in` assertion is a nice guardrail — it'll fail if anyone reintroduces the `_loadingSessionId`-based guard. The companion `test_session_visit_model_refresh_is_deferred_until_after_first_paint` keeps the double-rAF → `requestIdleCallback(timeout:1500)` deferral pinned so the early scheduling still doesn't block first paint.

No remaining concerns from the deferral/first-paint angle. The approach is sound and the ordering is now both correct and test-locked. Review-only — not merged.

nesquena-hermes:
Shipped in **v0.51.727** 🎉 — thanks @santastabber. Session switches now defer the workspace-tree/git-badge refresh, the model-dropdown option-list refresh, and the redundant empty-draft save until after first paint, each re-checking the active session id at fire time. Full deep gate clean (Codex regression-safe + Opus ship-it: confirmed the `_loadingSessionId`→`isActiveModelRefreshSession()` change is required, no stale-model and no draft-data-loss path, rapid/parallel switching safe). Also allowlisted `requestIdleCallback`/`cancelIdleCallback` in the static-JS scope-undef gate (your call site's `typeof` guard is correct; the gate just lacked the standard global). One benign note: in a backgrounded tab the deferred refreshes wait for foreground (rAF gating) — no data loss.

PR #5118, opened 2026-06-28
Fix pinned streaming scroll restoration

santastabber:
## Summary
- preserve tail-relative scroll position for pinned/following transcript snapshots before semantic viewport-anchor restoration
- keep semantic viewport-anchor restoration for explicitly unpinned/manual reader positions
- add a focused regression lock for pinned streaming restore ordering

## Why
Live activity/worklog DOM updates can rebuild rows while a response is streaming. When the reader is pinned to the live tail, restoring the first visible viewport anchor can remount an older row and yank long transcripts upward. Pinned snapshots should instead preserve bottom distance; manual reading positions should keep the existing semantic-anchor behavior.

## Verification
- `node -c static/ui.js`
- `git diff --check`
- `./scripts/test.sh tests/test_tars_scroll_reset_regressions.py tests/test_issue1690_scroll_completion.py tests/test_issue4295_scroll_pin_reentry.py tests/test_streaming_sidebar_scroll.py tests/test_issue4006_auto_scroll_follow_default.py tests/test_live_to_final_anchor_visible_order.py -q`

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -11332,10 +11332,33 @@ function _captureMessageScrollSnapshot(){
     userUnpinned:_messageUserUnpinned,
   };
 }
+function _restorePinnedMessageScrollSnapshot(snapshot){
+  const el=$('messages');
+  if(!el||!snapshot||snapshot.pinned!==true||snapshot.userUnpinned===true) return false;
+  const maxTop=Math.max(0,el.scrollHeight-el.clientHeight);
+  const bottom=Number(snapshot.bottom);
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
Shipped in **v0.51.725** 🎉 — thanks @santastabber. Pinned/following readers now keep their tail-relative scroll position when a streaming turn rebuilds its rows, instead of getting bounced to an older row; unpinned/manual positions still use the semantic anchor restore. Full gate clean (Codex regression-safe + Opus ship-able + full suite + 49 scroll/virtualization regression tests). A small follow-up to remove the now-dead `pinned` fallback branches is noted for later cleanup.

Issue #1855, opened 2026-05-07
bug: /api/chat/start and /api/sessions can wedge for ~125s while health remains responsive


Michaelyklam:
## Bug Description

In a production WebUI deployment, browser-facing API requests occasionally remained open for ~125 seconds and were cancelled by the client/proxy timeout layer while the WebUI process was still partially healthy.

The affected endpoints were:

- `POST /api/chat/start`
- `GET /api/sessions`

At the same time, other WebUI endpoints such as `/health` and `/api/projects` remained fast. That suggests a narrower WebUI request-path wedge rather than a full process or host outage.

This is validated as a real production incident, but the exact internal WebUI stage is not yet isolated because the current request log only emits at handler completion; the timed-out requests did not produce WebUI completion log lines.

## Environment

- WebUI build observed: `v0.50.294-3-g7894f36-dirty`
- HEAD: `7894f36a25aa143d7ea76688008fc1bfe30e24d4`
- Public route mounted under a subpath
- Note: local checkout was dirty at the time of observation, so this issue is best treated as an investigation/diagnostic issue until reproduced on clean main.

### Current health after remediation

After restarting `hermes-webui.service`, health and unauthenticated route probes were fast:

```text
local /health: 200, ~0.002s
public /health: 200, ~0.156s
public /api/sessions without auth: 401, ~0.080s (expected unauthenticated response)
```

### Fronting access logs during the incident

Parsed access logs for `2026-05-07 20:30–21:15 UTC` showed:

```text
125.010s status=0 size=0 POST /api/chat/start
  start=2026-05-07T21:07:06.949622Z end=2026-05-07T21:09:11.959827Z

125.006s status=0 size=0 GET /api/sessions
  start=2026-05-07T21:09:12.222171Z end=2026-05-07T21:11:17.228662Z

92.599s status=200 size=313406 GET /api/chat/stream?stream_id=ff759403ddab429abeec418587cd8f15
75.156s status=200 size=265540 GET /api/chat/stream?stream_id=6ca97851a00142e1b7984dc709a316b0
40.649s status=200 size=174681 GET /api/chat/stream?stream_id=4d05c3b7029b4a39872b4c66bf4124f3
```

### Tunnel/proxy-side cancellation logs during the incident

Matching cancellation logs showed the same URLs being cancelled by the request boundary:

```text
2026-05-07T21:09:11Z request cancelled: /api/chat/start
2026-05-07T21:11:17Z request cancelled: /api/sessions
```

[...]

PR #5128, opened 2026-06-28
fix: repair bare custom-provider model on process-wakeup fast path

### LINKED ISSUE #5127: Bug: WebUI process-wakeup turns can call custom providers with bare model IDs
# WebUI process-wakeup turns can call custom providers with bare model IDs

## Summary

Hermes WebUI intermittently starts server-side background-process wakeup turns with a **bare model ID** instead of the profile-configured provider-qualified model ID.

For profiles using a **named custom provider** whose default model uses a slash namespace (for example `vendor/model-name`), the upstream OpenAI-compatible proxy may reject the request:

```text
HTTP 400: Invalid model format or no credentials for provider: model-name
```

Configured/default model example:

```text
x-ai/grok-composer-2.5-fast
```

Failing request example:

```text
grok-composer-2.5-fast
```

The failure often appears when a `terminal(background=true, notify_on_complete=true)` process finishes and WebUI starts the automatic server-side wakeup turn.

## Impact

- Long WebUI sessions fail unpredictably after background processes complete.
- The assistant may not process `notify_on_complete` results reliably.
- Sessions that used the correct qualified model for many turns can fail on the wakeup turn.
- User-visible errors look like provider/credential misconfiguration even when profile config and persisted session JSON are correct.

## Expected behavior

For a profile configured with a slash-qualified default on a named custom provider:

```yaml
model:
  default: x-ai/grok-composer-2.5-fast
  provider: custom:my-proxy
custom_providers:
  - name: my-proxy
    base_url: https://example.com/v1
```

WebUI should launch the agent with the qualified model ID (or equivalent routing that preserves the namespace), for example:

```text
model=x-ai/grok-composer-2.5-fast
provider=custom:my-proxy
```

## Actual behavior

During a server-side `process_wakeup` turn, WebUI sometimes launches with:

```text
model=grok-composer-2.5-fast
provider=custom:my-proxy
```

That bare suffix is invalid for proxies that require the full `vendor/model` path.

## Persisted state can be correct

This is not necessarily incorrect profile config or corrupted on-disk session JSON. Session files and profile `state.db` rows can still show the qualified model while the in-memory `Session` object used for the wakeup turn carries a bare `model` field.

The bare ID appears to be introduced or retained in runtime/session-start normalization, not in canonical persisted config.

## Trigger path

Typical sequence after background process completion:

```text
api.background_process._start_server_side_wakeup_turn()
  → api.routes.start_session_turn(..., source="process_wakeup")
  → _resolve_compatible_session_model_state(..., prefer_cached_catalog=True)
  → _start_run(...)
  → api.streaming._start_chat_stream_for_session(...)
  → api.config.model_with_provider_context(...)
  → api.config.resolve_model_provider(...)
  → AIAgent(... model=<bare model> ...)
```

## Suspected root cause

`_resolve_compatible_session_model_state()` has a fast path (#1855) that returns `(model, requested_provider)` verbatim when both are present and the model is not `@provider:model`-qualified:

```python
if model and requested_provider:
    bare_model, explicit_provider = _split_provider_qualified_model(model)
    ...
    if not explicit_provider and not stale_codex_openai_slash_id:
        return model, requested_provider, False
```

This is unsafe when the profile default is slash-qualified but runtime `model` is only the suffix (same provider).

Example inputs:

```text
model = grok-composer-2.5-fast
model_provider = custom:my-proxy
profile_default_model = x-ai/grok-composer-2.5-fast
prefer_cached_catalog = True
```

Current result:

```text
(grok-composer-2.5-fast, custom:my-proxy, False)
```

Desired repair:

```text
(x-ai/grok-composer-2.5-fast, custom:my-proxy, True)
```

Then `model_with_provider_context()` may preserve the bare model when `provider == config_provider`:

```python
if provider == config_provider:
    return model
```

So `resolve_model_provider()` receives the invalid bare ID.

## Why this can look intermittent

`api.models.get_session()` caches `Session` objects; freshness checks focus on message/journal state more than `model` / `model_provider` drift. Process wakeups use `prefer_cached_catalog=True` to avoid blocking catalog rebuilds, which can let a bare cached model survive as authoritative.

## Proposed fix

**Minimal:** In `_resolve_compatible_session_model_state()`, before the fast-path return, when `requested_provider` is `custom` or `custom:*`, `profile_default_model` is slash-qualified, and current `model` equals the suffix of that default, repair to `profile_default_model`.

**Hardening (optional):**

1. Teach `model_with_provider_context()` not to return bare custom-provider models when the configured default is slash-qualified and the bare model matches the suffix.
2. Reload or repair cached sessions when on-disk `model` / `model_provider` differ from memory before server-side wakeups.

## Suggested regression tests

1. **Resolver unit test** — inputs above → qualified model, `normalized_model=True`.
2. **Server-side wakeup integration** — cached `s.model` bare, disk/profile default qualified → `start_session_turn(..., source="process_wakeup")` must pass qualified model to `_start_chat_stream_for_session`.
3. **Non-regression** — intentional bare custom models without slash-qualified defaults; `@provider:model` hints; existing OpenAI Codex `openai/...` repair; OpenRouter slash IDs unchanged.

## Severity

High for WebUI users on named custom providers who rely on background `notify_on_complete` workflows — failures occur exactly when completion results should be processed.

## Ask

Please confirm whether repairing bare suffixes against slash-qualified profile defaults in the process-wakeup path is acceptable, or if catalog/live resolution should always run for `prefer_cached_catalog=True` wakeups.

claw-io:
Closes #5127.

## Thinking Path

- Server-side `process_wakeup` calls `_resolve_compatible_session_model_state(..., prefer_cached_catalog=True)` with persisted `(s.model, s.model_provider)`.
- The #1855 fast path returns `(model, provider)` verbatim when both are set and the model is not `@provider:`-qualified — avoiding catalog I/O.
- Named custom providers often use slash-qualified profile defaults (`x-ai/...`) while in-memory session `model` can be only the suffix (`grok-composer-2.5-fast`), causing proxy 400s on the wakeup turn.
- Persisted session JSON can still show the qualified model; repair must run on the fast path without calling `get_available_models()`.
- `_read_profile_model_config` previously returned `(None, None)` when `model_provider` was already set, so wakeup could not see `profile_default_model` for suffix repair.

## What Changed

- **`api/routes.py` — `_read_profile_model_config`:** When the session has an explicit `model_provider`, still load `model.default` from the profile (provider remains session-owned).
- **`api/routes.py` — `_resolve_compatible_session_model_state` fast path:** For `custom` / `custom:*`, if `profile_default_model` is slash-qualified and runtime `model` equals its suffix, repair to the full qualified default (`normalized=True`).
- **`tests/test_issue5127_process_wakeup_bare_model.py`:** Resolver repair, non-regression, Codex `openai/` slow-path guard, profile-default read with explicit provider.

No `CHANGELOG.md` edits.

## Why It Matters

Background `terminal(notify_on_complete=true)` workflows depend on server-side wakeups. A bare model on that path fails at the moment the agent should consume completion output — often after an otherwise healthy long session.

## Verification

```bash
./scripts/test.sh tests/test_issue5127_process_wakeup_bare_model.py tests/test_issue3405_profile_provider_resolution.py -q
```

34 passed locally on this branch.

Manual (recommended): profile with `custom:*` and slash-qualified `model.default` → long WebUI session → background command with `notify_on_complete` → confirm wakeup turn uses qualified model in logs (not bare suffix).

## Risks / Follow-ups

- Repair is scoped to custom providers with slash-qualified profile defaults and suffix match; unrelated bare custom IDs are unchanged.
- Optional hardening from #5127 (session cache model/provider freshness, `model_with_provider_context`) not included — can follow if maintainers want broader repair.

## Release note (for maintainers)

**Fixed:** Process-wakeup model resolution repairs bare custom-provider model suffixes to the profile’s slash-qualified default on the #1855 fast path, without triggering a catalog rebuild.

## Model Used

- Provider: custom (llm-proxy) / Hermes developer profile
- Model: `x-ai/grok-composer-2.5-fast`
- AI disclosure: AI assisted implementation, tests, and this PR description.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5219,8 +5219,27 @@ def _read_profile_model_config(
     explicit ``requested_provider`` is already set (profile should not
     override explicit selections).
     """
-    if _clean_session_model_provider(requested_provider) or not getattr(session, "profile", None):
+    if not getattr(session, "profile", None):
         return None, None
+    if _clean_session_model_provider(requested_provider):
+        # Session already carries an explicit provider; still read profile default
+        # so process-wakeup can repair bare model suffixes (#5127).
+        try:
+            from api.profiles import get_hermes_home_for_profile
+            _profile_home = get_hermes_home_for_profile(session.profile)
+            _profile_cfg_path = os.path.join(str(_profile_home), "config.yaml")
+            if not os.path.isfile(_profile_cfg_path):
+                return None, None
+            import yaml
+            with open(_profile_cfg_path, encoding="utf-8") as _f:
+                _pcfg = yaml.safe_load(_f) or {}
+            if not isinstance(_pcfg, dict):
+                return None, None
+            _default = (_pcfg.get("model", {}).get("default") or "").strip() or None
+            return None, _default
+        except Exception:
+            logger.warning("profile provider read failed for %r", getattr(session, "profile", None), exc_info=True)
+            return None, None
     try:
         from api.profiles import get_hermes_home_for_profile
         _profile_home = get_hermes_home_for_profile(session.profile)
@@ -5302,6 +5321,19 @@ def _resolve_compatible_session_model_state(
             and model_prefix == "openai"
         )
         if not explicit_provider and not stale_codex_openai_slash_id:
+            _profile_default = str(profile_default_model or "").strip()
+            if (
+                _profile_default
+                and "/" in _profile_default
+                and "/" not in model
+                and _profile_default.rsplit("/", 1)[-1] == model
+                and requested_provider
+                and (
+                    requested_provider == "custom"
+                    or str(requested_provider).startswith("custom:")
+                )
+            ):
+                return _profile_default, requested_provider, True
             return model, requested_provider, False
 
     # Default (human chat/start) path calls get_available_models() with NO
--- a/api/routes.py
+++ b/api/routes.py
@@ -5214,49 +5214,51 @@ def _read_profile_model_config(
 ) -> tuple[str | None, str | None]:
     """Read model.provider and model.default from the session's profile config.
 
-    Returns (profile_provider, profile_default_model). Both are None when
-    the session has no profile, the profile config is unreadable, or an
-    explicit ``requested_provider`` is already set (profile should not
-    override explicit selections).
+    Returns (profile_provider, profile_default_model). Both are None when the
+    session has no profile or the profile config is unreadable.
+
+    When the session already has an explicit ``requested_provider``, the profile
+    ``model.provider`` is not returned (first tuple element is None) so profile
+    does not override the session provider. ``profile_default_model`` is still
+    returned for suffix repair (#5127) only when the profile's configured
+    provider matches ``requested_provider`` after normalization.
     """
     if not getattr(session, "profile", None):
         return None, None
-    if _clean_session_model_provider(requested_provider):
-        # Session already carries an explicit provider; still read profile default
-        # so process-wakeup can repair bare model suffixes (#5127).
-        try:
-            from api.profiles import get_hermes_home_for_profile
-            _profile_home = get_hermes_home_for_profile(session.profile)
-            _profile_cfg_path = os.path.join(str(_profile_home), "config.yaml")
-            if not os.path.isfile(_profile_cfg_path):
-                return None, None
-            import yaml
-            with open(_profile_cfg_path, encoding="utf-8") as _f:
-                _pcfg = yaml.safe_load(_f) or {}
-            if not isinstance(_pcfg, dict):
-                return None, None
-            _default = (_pcfg.get("model", {}).get("default") or "").strip() or None
-            return None, _default
-        except Exception:
-            logger.warning("profile provider read failed for %r", getattr(session, "profile", None), exc_info=True)
-            return None, None
+
     try:
         from api.profiles import get_hermes_home_for_profile
+
         _profile_home = get_hermes_home_for_profile(session.profile)
         _profile_cfg_path = os.path.join(str(_profile_home), "config.yaml")
         if not os.path.isfile(_profile_cfg_path):
             return None, None
         import yaml
+
         with open(_profile_cfg_path, encoding="utf-8") as _f:
             _pcfg = yaml.safe_load(_f) or {}
         if not isinstance(_pcfg, dict):
             return None, None
-        _provider = (_pcfg.get("model", {}).get("provider") or "").strip() or None
-        _default = (_pcfg.get("model", {}).get("default") or "").strip() or None
-        return _provider, _default
+        _model_cfg = _pcfg.get("model") or {}
+        if not isinstance(_model_cfg, dict):
+            return None, None
+        _provider = (_model_cfg.get("provider") or "").strip() or None
+        _default = (_model_cfg.get("default") or "").strip() or None
     except Exception:
-        logger.warning("profile provider read failed for %r", getattr(session, "profile", None), exc_info=True)
+        logger.warning(
+            "profile provider read failed for %r",
+            getattr(session, "profile", None),
+            exc_info=True,
+        )
         return None, None
+
+    _requested = _clean_session_model_provider(requested_provider)
+    if _requested:
+        _profile_prov = _clean_session_model_provider(_provider)
+        if _profile_prov != _requested:
+            return None, None
+        return None, _default
+    return _provider, _default
 
 
 def _resolve_compatible_session_model_state(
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Reading the diff at `api/routes.py` against `origin/master`, plus the resolver call chain on master and the agent-side cron contract, this fix lands in the right layer and the scoping looks correct.

## What the fix does, and why the layer is right

The bug is that the #1855 fast path in `_resolve_compatible_session_model_state` returns `(model, requested_provider, False)` verbatim once both are set and the model isn't `@provider:`-qualified. For a named custom provider whose profile default is slash-qualified (`x-ai/grok-composer-2.5-fast`) but whose runtime `model` is the bare suffix (`grok-composer-2.5-fast`), that verbatim return is what eventually reaches `model_with_provider_context()` — and there the bare suffix survives because:

```python
# api/config.py:2806-2808
# If the selected provider is already the configured provider, leaving the
# model bare preserves provider-specific base_url/proxy settings.
if provider == config_provider:
    return model
```

So the bare ID flows straight into `resolve_model_provider()` and the proxy 400s. Repairing it at the resolver (before that return) is the correct fix point — `model_with_provider_context` then returns the already-qualified ID. A patch in `api/config.py` would have been the wrong layer since that function legitimately keeps bare IDs for same-provider routing.

## The repair guard reads correctly

```python
# api/routes.py fast path
_profile_default = str(profile_default_model or "").strip()
if (
    _profile_default
    and "/" in _profile_default
    and "/" not in model
    and _profile_default.rsplit("/", 1)[-1] == model
    and requested_provider
    and (requested_provider == "custom" or str(requested_provider).startswith("custom:"))
):
    return _profile_default, requested_provider, True
```

The `"/" not in model` + suffix-equality + `custom`/`custom:*` gate keeps this from touching OpenRouter slash IDs, ACP subprocess providers, or unrelated bare custom models. The `test_fast_path_does_not_rewrite_unrelated_bare_model` and `test_openai_codex_stale_openai_slash_still_uses_slow_path` cases cover the two regressions I'd worry about most.

## One wiring note worth confirming

The repair only fires if `profile_default_model` is actually populated on the wakeup path. The `_read_profile_model_config` change is what makes that possible — on master it returned `(None, None)` whenever `requested_provider` was set, so the resolver never saw the profile default. The new explicit-provider branch returns `(None, _default)`, and the call sites at `api/routes.py:16723-16728` feed that through `profile_default_model=_pp_default`. Worth a quick manual confirm that the `source="process_wakeup"` resolver call (the one with `prefer_cached_catalog=True`) is among those that read `_pp_default` first — if any wakeup path calls the resolver without first calling `_read_profile_model_config`, the repair silently no-ops there.

## Minor (non-blocking)

Greptile's two P2s are fair: the `_read_profile_model_config` docstring at the top still says "Both are None when ... an explicit `requested_provider` is already set," which is now stale, and the two branches duplicate the full YAML-read block. Neither affects correctness, but the docstring should be corrected since the return contract changed — a future caller reading "both None when provider set" could make a wrong assumption.

The fast-path-only, catalog-free repair is the right call for the wakeup path given `prefer_cached_catalog=True` exists precisely to avoid the blocking rebuild. Nice tracing in the issue.

nesquena-hermes:
Thanks @claw-io — the wakeup bare-model repair is the right instinct and the persisted-qualified-form lookup is clean, but a deep regression gate (Codex) surfaced a **cross-provider mis-resolution** that needs a fix before this can ship:

**Issue (CORE):** `_resolve_compatible_session_model_state()` rewrites any bare custom-provider model whose *suffix* matches the profile default, without checking that the profile default belongs to the **same** custom provider. So a session pinned to `custom:other-proxy` with `grok-composer-2.5-fast` can be rewritten to `x-ai/grok-composer-2.5-fast` while **still routing to `custom:other-proxy`** — which breaks providers that legitimately accept the bare ID.

Repro shape:
```
session: ("grok-composer-2.5-fast", "custom:other-proxy")
profile default: "x-ai/grok-composer-2.5-fast"
current PR result: ("x-ai/grok-composer-2.5-fast", "custom:other-proxy")   ← wrong, cross-provider
expected:          ("grok-composer-2.5-fast",       "custom:other-proxy", False)  ← unchanged
```

**Fix spec:**
1. In the explicit-provider branch (~`api/routes.py:5238`), read the profile default's **provider** too and only expose `profile_default_model` for repair when `_clean_session_model_provider(profile_provider) == _clean_session_model_provider(requested_provider)`.
2. Add a regression test proving a mismatched custom provider preserves the bare model: `("grok-composer-2.5-fast", "custom:other-proxy", profile default "x-ai/grok-composer-2.5-fast")` stays `("grok-composer-2.5-fast", "custom:other-proxy", False)`.

The fix that closes #5127 is good — it just needs to be gated on provider equality so it can't silently re-point a different custom provider's bare model. Re-request review once that lands and I'll re-gate. 🙏

nesquena-hermes:
Thanks @claw-io — re-gated and this is ready. ✅

The provider-equality gate now closes the cross-provider mis-resolution I flagged:

- `_read_profile_model_config` only returns `profile_default_model` for repair when `_clean_session_model_provider(profile_provider) == _clean_session_model_provider(requested_provider)`.
- `_resolve_compatible_session_model_state` carries the same `_providers_match_for_repair` guard alongside the existing `"/" in default` / `"/" not in model` / suffix-equality / `custom`/`custom:*` checks.

Verified the exact regression shape you bounced on is now covered: `test_fast_path_skips_repair_when_profile_provider_mismatches` proves a session pinned to `custom:other-proxy` with `grok-composer-2.5-fast` and a profile default of `x-ai/grok-composer-2.5-fast` stays `("grok-composer-2.5-fast", "custom:other-proxy", False)` — no cross-provider rewrite.

Gate:
- Full file green on a clean Linux/3.11 checkout (8/8).
- Staged on fresh master + full suite green (11016 passed).
- Codex regression gate on the stage batch: no findings against this change.

Shipping in the next batch. Credit preserved (co-author @b3nw).

PR #5136, opened 2026-06-28
perf(#5132): cap /api/sessions state.db source/title overrides to top-N

### LINKED ISSUE #5132: Bug: /api/sessions blocks 5-18s on state_db_overrides for power users → repeated "Connection lost" flapping
## Summary

On power-user instances with many sessions, the WebUI `GET /api/sessions` poll blocks for **5–18 seconds**, almost entirely inside the `all_sessions.state_db_overrides` stage. Concurrent polls pile up (each opens its own sqlite connection to `state.db`), and while they are blocked the frontend shows **"Connection lost" / "request timed out"**. Sessions recover for a few seconds, then RTO again — "persistently RTO, recover only for a few seconds."

Reported by **@latipun** on `WebUI: v0.51.719`, and independently reproduced by **@b3nw** in the same thread ("I seem to be experiencing this too now"). WSL2 host.

## Evidence (from `hermes debug share` agent.log, redacted)

The request-diagnostics watchdog caught multiple `/api/sessions` requests in-flight at once, all stuck on the same stage:

```
Slow WebUI request still running: path=/api/sessions current_stage=all_sessions.state_db_overrides elapsed_ms=5748.7
Slow WebUI request completed:   path=/api/sessions elapsed_ms=18334.7  ... all_sessions.state_db_overrides: 16676.1ms
Slow WebUI request completed:   path=/api/sessions elapsed_ms=11278.1  ... all_sessions.state_db_overrides:  9700.0ms
```

Per-stage breakdown for the 18.3s request — `state_db_overrides` is ~91% of the wall clock:

| stage | ms |
|---|---|
| load_settings | 190 |
| session_list_cache_lookup | 632 |
| **all_sessions.state_db_overrides** | **16676** |
| visible_lineage_metadata | 736 |
| (everything else) | <60 each |

Thread stacks at the moment of the timeout show **5+ concurrent request threads** plus the gateway-watcher all hitting `state.db` simultaneously:

- Request threads parked in `_read_state_db_sidebar_overrides` → `cur.fetchall()` (`api/models.py`) and in `merge_session_messages_append_only` → `_has_visible_duplicate` → `_loose_session_message_content` (`re.findall` per message).
- `gateway-watcher` thread in `_cheap_change_fingerprint` → `cur.fetchall()` on the same DB.

## Suspected root cause

`_read_state_db_sidebar_overrides()` runs on the `/api/sessions` hot path and:
- opens a **fresh `sqlite3.connect(state.db)` on every call** (no shared/cached connection),
- is invoked by **every concurrent poll independently** (no result memoization across in-flight requests),
- runs a `PRAGMA table_info` + chunked `SELECT ... WHERE id IN (...)` over the full session set (2400+ ids for this user).

Under concurrent reads + the gateway-watcher reading + the agent writing to the same `state.db` — on a WSL2 filesystem — this turns into lock/IO contention that scales to 5–18s. Because the session-list cache lookup also takes ~0.6–1s and several polls run at once, the blocked requests stack up and the UI flaps to "Connection lost."

A secondary cost is `merge_session_messages_append_only`'s dedup, which calls `re.findall(r"\w+", ...)` per message (`_loose_session_message_content`) inside an O(n)-per-message duplicate check on large transcripts.

## Likely fix directions

1. **Memoize / share** the `state_db_overrides` result for the duration of a session-list cache window instead of recomputing per concurrent poll (fold it into the existing `_get_cached_session_list_payload` cache, or single-flight it).
2. **Reuse a single read-only sqlite connection** (or `PRAGMA query_only` + WAL read) rather than connecting per call; coordinate with the gateway-watcher's reads.
3. **Cap the override lookup** to the visible/top-N sidebar rows the same way `_enrich_sidebar_lineage_metadata` was already capped (see the #38914 / 2026-06-21 lineage cap comment in `api/models.py`) — the full 2400-row scan on every poll is the avoidable cost.
4. Cache the normalized form in `_loose_session_message_content` / hoist the regex out of the per-message dedup loop.

## Possibly-related secondary symptom

@latipun also reported **chat history not coming back / "swallowed" message history** after the flapping, and earlier reported tool-call output being truncated after stream end. The missing-history symptom touches the same `merge_session_messages_append_only` / `_has_visible_duplicate` dedup path that appears in these stacks, so it may share a root cause — flagging for the investigator, but the request-blocking issue above is the primary, clearly-evidenced bug.

## Environment

- WebUI: v0.51.719
- OS: Linux 6.18.x microsoft-standard-WSL2 x86_64, Python 3.11.15
- Reporters: @latipun (primary), @b3nw (reproduced)

## Out of scope for this issue

The `cannot import name 'normalize_tool_schema' from 'agent.memory_manager'` error in the same thread is a **core Hermes agent stale-code-after-`hermes update`** problem (resolved by a full `stop && start` of the service, not `restart`), not a WebUI bug — already triaged in-thread.

## Related

- #4765 (memory bloat — loads all session JSON into RAM) is adjacent but distinct: that is RAM growth, this is per-request CPU/IO blocking on the `/api/sessions` path.

nesquena-hermes:
## What

Caps the `/api/sessions` sidebar **state.db source/title override** lookup to the top-N most-recent (paint-priority) sessions, fixing the 5–18s request stalls that flap the WebUI to **"Connection lost"** on power-user instances.

Fixes #5132 (reported by @latipun, reproduced by @b3nw).

## Root cause

The request-diagnostics watchdog in the issue caught `/api/sessions` spending **~91% of an 18.3s request inside the `all_sessions.state_db_overrides` stage**:

| stage | ms |
|---|---|
| load_settings | 190 |
| session_list_cache_lookup | 632 |
| **all_sessions.state_db_overrides** | **16676** |
| visible_lineage_metadata | 736 |

`_apply_sidebar_state_db_overrides()` read `state.db` for **every** session row (2400+ ids for the reporter) on the hot path, and **every concurrent poll did it independently**. With the gateway-watcher reading and the agent writing the same `state.db` — on a WSL2 filesystem — this became lock/IO contention that scaled to 5–18s. Blocked polls piled up and the UI flapped to "Connection lost," recovering for a few seconds before timing out again.

## Fix

Mirror the **exact precedent** already applied to the adjacent lineage read in #4638 (`_enrich_sidebar_lineage_metadata`): cap the state.db lookup to the top-N most-recent rows.

- The caller passes an **already pinned-first/newest-first sorted** list, and the sidebar paints chronologically newest-first, so the top-N covers the visible window.
- Rows beyond the cap keep their JSON source/title and are corrected **lazily** when the user opens the history panel (same as lineage).
- Cap defaults to **300**, env-configurable via `HERMES_WEBUI_STATE_DB_OVERRIDE_TOP_N` (set `0` to disable, matching `HERMES_WEBUI_LINEAGE_TOP_N`).
- **Fails open** — a DB error never breaks `/api/sessions`.

This bounds the per-poll state.db work the same way the lineage cap did, so the override stage no longer scales with total session count.

```diff
 def _apply_sidebar_state_db_overrides(sessions: list[dict]) -> None:
+    _cap = int(os.environ.get("HERMES_WEBUI_STATE_DB_OVERRIDE_TOP_N", "300"))  # 0 disables
+    candidates = sessions[:_cap] if (_cap > 0 and len(sessions) > _cap) else sessions
     metadata = _read_state_db_sidebar_overrides(
-        _active_state_db_path(), {ids for all sessions})
+        _active_state_db_path(), {ids for candidates})
```

## Tests

New `tests/test_issue5132_state_db_override_top_n_cap.py` (7 cases, mirrors the #4638 cap suite):
- caps to default top-300 when the list exceeds the cap
- honors the env override
- non-positive cap disables capping (override all)
- unparseable cap falls back to 300 (never crashes)
- list at/under the cap overrides everything
- a DB error is swallowed (fails open)
- a probed top-N row still receives its override (correctness preserved)

```
52 passed  (new suite + test_issue4638_lineage_top_n_cap + test_webui_state_db_reconciliation)
```

## Scope note

The issue flags a **secondary** cost — the per-message `re.findall(r"\w+", ...)` in `_loose_session_message_content` inside `merge_session_messages_append_only`'s dedup. That path is on the session-**detail** load, not the `/api/sessions` poll that the evidence pins, and the maintainer's own triage marks it "possibly-related... flagging for the investigator" while naming the override scan "the primary, clearly-evidenced bug." Kept out of this PR to stay surgical; worth a follow-up if profiling on a large transcript confirms it's material.


nesquena (approved):

Independent review of the #5132 perf fix: on power-user instances, `GET /api/sessions` blocked **5–18s** in the sidebar source/title reconciliation stage (`all_sessions.state_db_overrides`) because it read `state.db` for *every* session row (2400+ ids) on *every* concurrent poll. With the gateway-watcher reading and the agent writing the same DB, this became lock/IO contention that piled up concurrent polls and flapped the UI to "Connection lost." The fix caps the override lookup to the top-N most-recent (paint-priority) rows — exactly mirroring the existing #4638 lineage-enrichment cap.

### What this ships

`api/models.py` (+28/-2), `tests/test_issue5132_state_db_override_top_n_cap.py` (+117, new), `CHANGELOG.md` (+1). Agent-authored (`nesquena-hermes`), flagged for review. Pure read-path optimization — no schema change, no write-path change.

The branch was `CONFLICTING` (a stale `[Unreleased]` CHANGELOG entry colliding with several PRs merged since). I rebased onto current master; the only conflict was CHANGELOG (kept all entries, including the #5132 line). The code (`api/models.py`) merged cleanly. Force-pushed `851a9491`. The branch is now mergeable and the trace below reflects current master.

`state.db` is written by the agent and read by the WebUI sidebar. This change only reduces **how many rows the WebUI reads** per poll — the agent's write path is untouched, and rows beyond the cap simply retain their JSON-derived source/title until lazily corrected. No agent/protocol surface, no `config.yaml` write, no cross-tool regression. ✓

### End-to-end trace

1. **Caller sorts before capping.** Both call sites sort pinned-first/newest-first immediately before the override stage: `result = sorted(index_map.values(), key=lambda s: (s.get('pinned', False), _session_sort_timestamp(s)), reverse=True)` (`api/models.py:3835`) then `_apply_sidebar_state_db_overrides(result)` (`api/models.py:3859`); the second path sorts at `:3889` then calls at `:3905`. So `sessions[:_cap]` is genuinely the top-N **visible** rows. ✓
2. **The cap** (`api/models.py:3600-3606`): `_cap = int(os.environ.get("HERMES_WEBUI_STATE_DB_OVERRIDE_TOP_N", "300"))` with a `(TypeError, ValueError)→300` fallback; `candidates = sessions[:_cap]` when `_cap > 0 and len > _cap`, else all sessions. Only `candidates`' ids are probed against `state.db`.
3. **Application is correct for uncapped rows.** `_apply_sidebar_state_db_override_metadata` iterates **all** sessions but `continue`s when `sid not in metadata` — and `metadata` only contains the capped ids — so rows beyond the cap keep their JSON `source`/`title`/`source_tag` untouched. No row is mutated to a wrong value; it's just not corrected this poll. ✓
4. **Parity with #4638** (the cap it claims to mirror): `_enrich_sidebar_lineage_metadata` uses the identical shape — `int(os.environ.get("HERMES_WEBUI_LINEAGE_TOP_N", "300"))`, `candidates = sessions[:_cap]` (`api/models.py:3693-3697`). Same default, same slice, same fail-open. The mirror claim is accurate. ✓

### Security / robustness audit — clean

- **Env var is int-parsed with a safe fallback** — an unparseable/missing value yields 300, never a crash. No injection surface (the value only bounds a slice length). ✓
- **`state.db` read is wrapped in `try/except → return`** (fails open: on any DB error the sidebar keeps JSON values rather than erroring the whole `/api/sessions` response). ✓
- **`session_id` set construction is unchanged** (`str(s.get('session_id'))`, same filtering) — just over a smaller candidate set. No new SQL surface; `_read_state_db_sidebar_overrides` is untouched. ✓
- `import os as _os` inside the function is slightly unusual but harmless (module-cached). Non-blocking nit.

### Edge-case matrix

| `HERMES_WEBUI_STATE_DB_OVERRIDE_TOP_N` / state | Behavior |
|---|---|
| unset, >300 sessions | probes exactly the top-300 ✅ (test) |
| `50` | probes top-50 ✅ (test) |
| `0` | cap disabled → probes all ✅ (test) |
| negative | `_cap > 0` false → probes all ✅ |
| `not-a-number` | falls back to 300, no crash ✅ (test) |
| sessions ≤ cap | probes everything ✅ (test) |
| `state.db` read raises | swallowed, JSON values kept ✅ (test) |
| capped row (within top-N) | still receives its override ✅ (test) |
| row beyond cap | keeps JSON source/title, corrected lazily on history-panel open ✅ |

### Tests

- `tests/test_issue5132_state_db_override_top_n_cap.py` (7) — default-300, env override, non-positive disable, unparseable fallback, under-cap, fail-open, capped-row-still-overridden. **7/7.**
- **Full suite (post-rebase): 10900 passed / 143 skipped / 0 failed** in 229s (deselected the known darwin CRLF flake).
- CI on the pre-rebase commit was already fully green (lint + browser-smoke + 9 shards); the rebase only touched CHANGELOG + replayed the same code.

- `import os as _os` mid-function — could hoist to module scope for tidiness; functionally fine.
- CHANGELOG `[Unreleased]` entry is correct per AGENTS.md convention (release agent promotes it) — kept through the rebase.
- The lazy-correction behavior (rows beyond cap fixed when the history panel opens) is the same trade-off #4638 already established and is documented in both the docstring and the CHANGELOG. Good.

### Recommendation

A tight, well-scoped read-path optimization that directly addresses the #5132 contention by bounding the per-poll `state.db` scan to the visible top-N rows, faithfully mirroring the established #4638 lineage cap (same env-var shape, same default, same fail-open, same lazy-correction trade-off). I verified the caller sorts pinned-first/newest-first before capping, that uncapped rows keep their JSON values untouched (never mis-corrected), and that the env knob degrades safely on bad input. Rebased to clear the CHANGELOG conflict; full suite green. Ship.

nesquena-hermes:
Shipped in **v0.51.732** via #5174 (rebased onto current master). The original top-N cap had a regression the dual gate caught — capping the override also capped the SOURCE classification that `/api/sessions` filters on before lineage correction, which could drop a stale-CLI-JSON / state.db-webui row from a WebUI-only sidebar beyond the cap. The shipped version splits it: source/title/count classification stays uncapped (indexed sessions-table lookup), and only the expensive `messages` COUNT/MAX aggregation (the real 5-18s bottleneck) is capped to top-N. Original perf work + investigation by @nesquena-hermes. Closing in favor of #5174.

PR #5147, opened 2026-06-28
fix(#5135): clear stale gateway approval mirrors on stream teardown

### LINKED ISSUE #5135: Approval card lingers after gateway_timeout — _drop_entry() skips _pending and SSE dismiss
## Summary

When a legacy-mode approval reaches `gateway_timeout` (default 300s), `_drop_entry()` removes the entry from `_gateway_queues` but does not touch `_pending` or notify SSE subscribers. The approval card stays visible in the UI until the next 1.5s poll cycle reconciles the stale mirror entry. During that window, user clicks on approve/deny are silently discarded because `resolve_gateway_approval()` finds nothing left in `_gateway_queues`.

The card also provides no timeout feedback — it silently disappears rather than showing a "timed out" state, even though the `post_approval_response` hook already fires with `choice="timeout"`.

## Source trace

**Timeout exit** — `tools/approval.py:1498-1507`:
```python
_remaining = _deadline - time.monotonic()
if _remaining <= 0:
    break                    # timeout — resolved stays False
# ...
_drop_entry()                # only cleans _gateway_queues
```

**`_drop_entry()` scope** — `tools/approval.py:1433-1439`:
```python
def _drop_entry() -> None:
    with _lock:
        queue = _gateway_queues.get(session_key, [])
        if entry in queue:
            queue.remove(entry)
        if not queue:
            _gateway_queues.pop(session_key, None)
    # _pending is untouched
    # no SSE notification fired
```

**Mirror entry that persists** — `route_approvals.py:187-193`:
`submit_gateway_pending_mirror()` created a mirror entry in `_pending` (tagged with `_GATEWAY_MIRROR_FLAG` + `_GATEWAY_MIRROR_TOKEN`) when the approval was first queued. After `_drop_entry()`, this mirror is stale.

**Eventual cleanup** — `route_approvals.py:139-184`:
`reconcile_gateway_pending_mirror_locked()` runs on the next `/api/approval/pending` poll (1.5s cadence), discovers the empty `_gateway_queues`, purges the stale mirror, and returns `(None, 0, True)`. The frontend then dismisses the card.

**Silent discard** — `routes.py:18451-18541`:
If the user clicks approve during the stale window, `_resolve_approval_legacy()` pops from `_pending` (entry exists) then calls `resolve_gateway_approval()` which returns 0 (entry gone from `_gateway_queues`). The user's response is silently discarded; no error, no feedback.

## Suggested fix

After `_drop_entry()` in `_await_gateway_decision()`, proactively dismiss the approval card. Two options:

1. **Emit a timeout event on the chat SSE stream** — `put('approval_timeout', approval_data)` alongside the existing `post_approval_response` hook fire. The frontend already listens to the chat stream for `approval` events; a new `approval_timeout` event type would let it dismiss immediately and optionally show "Timed out" on the card.

2. **Have `_drop_entry()` also reconcile `_pending`** — call `reconcile_gateway_pending_mirror_locked()` + `_approval_sse_notify_locked()` under `_lock` so the polling path picks it up instantly and SSE subscribers get `{pending: null}`.

Option 1 is cleaner because `_drop_entry()` lives in `tools/approval.py` (hermes-agent) which shouldn't need to know about `_pending` (a webui concern). The notify callback is already registered and available.

## Impact

Low severity — the stale window is at most 1.5s and only occurs after a 5-minute timeout. But the silent discard of user input and the lack of timeout feedback on the card are rough UX edges worth smoothing.

rodboev:
## Thinking Path

- The legacy gateway path already mirrors approval events into WebUI polling state, and `/api/approval/pending` can eventually reconcile those mirrors on the next fallback poll.
- The stale-click dead-end is already fixed on current base, so the surviving bug is the teardown gap before that next poll.
- The fix reuses the existing mirror authority and the existing non-gateway teardown pattern so the legacy gateway path clears stale approval cards as soon as teardown proves the approval is gone.

## What Changed

- `api/gateway_chat.py`: reconcile gateway approval mirrors during legacy gateway stream teardown, and notify approval subscribers with the reconciled head and count.
- `tests/test_gateway_approval_legacy_path.py`: add regression coverage for stale mirrored approvals clearing immediately at teardown, plus adjacent preservation coverage for local non-gateway approvals.

## Why It Matters

The approval card now disappears as soon as legacy gateway teardown proves the mirrored approval has no live head, instead of lingering until the next 1.5 second poll. Live gateway approvals and local approvals keep their existing behavior.

## Verification

```text
pytest tests/test_gateway_approval_legacy_path.py -v --timeout=60
pytest tests/test_approval_sse.py -v --timeout=60
```

Full-suite CI context: `pytest tests/ -v --timeout=60`.

## Upstream

Closes #5135.

Related prior work: #4551 established the legacy gateway approval relay path, and #4955 shaped the stale-click behavior this fix preserves.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -477,6 +477,21 @@ def _clear_gateway_pending_state(session: Any, stream_id: str) -> None:
     session.save()
 
 
+def _cleanup_gateway_pending_mirror(session_id: str) -> None:
+    try:
+        from api.route_approvals import (
+            _approval_sse_notify_locked,
+            _lock as _approval_lock,
+            reconcile_gateway_pending_mirror_locked,
+        )
+
+        with _approval_lock:
+            head, total, _ = reconcile_gateway_pending_mirror_locked(session_id)
+            _approval_sse_notify_locked(session_id, head, total)
+    except Exception:
+        logger.debug("Failed to reconcile gateway pending mirror during teardown", exc_info=True)
+
+
 def _run_gateway_chat_streaming(
     session_id,
     msg_text,
@@ -888,6 +903,7 @@ def _run_gateway_chat_streaming(
                     _clear_gateway_pending_state(get_session(session_id), stream_id)
             except Exception:
                 logger.debug("Failed to clear gateway stream state", exc_info=True)
+            _cleanup_gateway_pending_mirror(session_id)
         with STREAMS_LOCK:
             CANCEL_FLAGS.pop(stream_id, None)
             STREAM_PARTIAL_TEXT.pop(stream_id, None)
--- a/tests/test_gateway_approval_legacy_path.py
+++ b/tests/test_gateway_approval_legacy_path.py
@@ -28,6 +28,41 @@ def _extract_legacy_sse_loop():
     return GATEWAY_CHAT_SRC[start:end]
 
 
+def _drain_queue(q):
+    import queue
+
+    items = []
+    while True:
+        try:
+            items.append(q.get_nowait())
+        except queue.Empty:
+            return items
+
+
+def _make_legacy_gateway_urlopen(approval_payload: str, after_approval=None):
+    def fake_urlopen(req, *, timeout=None):
+        del req, timeout
+
+        def _iter():
+            yield b"event: approval.request"
+            yield f"data: {approval_payload}".encode("utf-8")
+            yield b""
+            if after_approval is not None:
+                after_approval()
+            yield b'data: {"choices":[{"delta":{"content":"Done"}}]}'
+            yield b""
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.726** 🎉 — thanks @rodboev. Stream teardown now reconciles the legacy-gateway approval mirrors against the live queue (under the approval lock), so a stale approval card clears as soon as teardown proves the approval is gone instead of waiting for the next fallback poll. Still-pending gateway and local approvals are preserved. Full gate clean (Codex regression-safe + full suite + 88 approval-path tests).

Issue #5057, opened 2026-06-27
Support internal Hermes slash commands like /moa


luckfu:
## Feature Description
Support executing internal Hermes slash commands such as `/moa` directly from Hermes WebUI, with behavior matching CLI / Telegram / Feishu sessions.

## Motivation
Hermes runtime already supports internal slash commands like `/moa` in CLI and gateway-backed chat surfaces. In Hermes WebUI, entering `/moa` is currently treated like plain text instead of being routed through Hermes' internal slash-command dispatch flow.

This creates an inconsistent user experience:
- CLI: `/moa stock-deep` switches the current session to the MoA preset
- Telegram / Feishu: `/moa stock-deep` works as an internal Hermes command
- Hermes WebUI: `/moa stock-deep` is sent as ordinary chat text

For users who mainly use Hermes from the browser or phone, this makes MoA feel unavailable even when it is configured correctly in Hermes.

## Proposed Solution
Add slash-command detection and routing in Hermes WebUI so internal Hermes commands are executed instead of being passed through as plain user messages.

Minimum expected behavior:
1. Detect messages that begin with `/`
2. Route supported internal commands through the same Hermes command resolution path used by CLI/gateway sessions
3. Apply the command to the current WebUI session state
4. Show the same acknowledgment/result text users would see in CLI

Examples:
- `/moa` → switch current WebUI session to default MoA preset
- `/moa lite` → switch current WebUI session to named preset
- `/moa <freeform prompt>` → run one-shot MoA turn, then restore the previous model

## Alternatives Considered
- Telling users to switch to CLI or messaging surfaces for slash commands. This is only a workaround.
- Treating WebUI as plain-chat-only. That breaks parity with Hermes' other surfaces.

## Notes
This request is specifically about **internal Hermes slash commands** (such as `/moa`, `/model`, etc.), not arbitrary shell commands.

nesquena-hermes:
Thanks — this is a real CLI-parity gap and a reasonable ask.

You're right that WebUI today only handles its **own** client-side slash commands (`/clear`, `/new`, `/compress`, `/model`, etc.) in the composer; a leading-`/` message that isn't one of those is sent to the agent as ordinary chat text rather than being dispatched through the agent's internal command resolver the way CLI / Telegram / Feishu sessions do. So `/moa` (and other internal Hermes commands) don't take effect from the browser.

[...]

PR #5153, opened 2026-06-28
fix: route moa webui turns through gateway

ruizanthony:
## Thinking Path

WebUI already resolves `/moa` server-side and applies it per turn for the local backend, but it deliberately failed closed for gateway-backed sessions. With the gateway API server able to honor per-request `provider`/`model` overrides, WebUI can route `/moa` through `provider=moa` and the resolved preset instead of returning 409.

## What Changed

- Removes the gateway-specific 409 for `/moa` chat start.
- For gateway-backed sessions, resolves the MoA preset server-side and sets the current turn to `model=<preset>`, `model_provider=moa`.
- Keeps the existing local backend behavior: local turns still receive `moa_config` directly as a per-turn override.
- Adds regression coverage that a gateway-backed `/moa` chat start launches with `model_provider="moa"`.

## Why It Matters

Users can use `/moa <prompt>` consistently from the WebUI even when the WebUI is backed by Hermes Gateway/API server. The browser still cannot supply arbitrary MoA config; it only sends a boolean signal and the server resolves the preset.

## Verification

- `./scripts/test.sh tests/test_issue5057_moa_webui_route.py -q` → `8 passed`
- `./scripts/test.sh tests/test_issue5057_moa_webui_route.py tests/test_gateway_approval_legacy_path.py -q` → `25 passed`
- `.venv/bin/ruff check tests/test_issue5057_moa_webui_route.py` → `All checks passed!`
- `.venv/bin/python -m compileall -q api/routes.py tests/test_issue5057_moa_webui_route.py` → exit 0

Lint note: `.venv/bin/ruff check api/routes.py tests/test_issue5057_moa_webui_route.py` still reports pre-existing `api/routes.py` issues outside this diff (e.g. B009/F841/F811/B007 around unrelated sections), so I verified the modified test file and compiled the touched route file.

## Risks / Follow-ups

- Requires the paired Hermes Agent/API server change that honors request-scoped `provider=moa` / `model=<preset>` on gateway-backed requests.
- If the resolved MoA preset is missing, the route returns a 503 instead of starting a malformed gateway turn.

## Contract Routing

Task type: gateway-backed WebUI runtime behavior fix.
Touched areas:
- `api/routes.py`
- `tests/test_issue5057_moa_webui_route.py`
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
Scope boundaries:
- No browser-supplied MoA config.
- No UI/UX layout changes.
- No changelog edit in contributor PR; release-note wording is in this PR body.
- Regression test for gateway `/moa` route behavior.
- Gateway-adjacent test coverage remains green.

## Model Used

AI-assisted: Hermes Agent using OpenAI Codex `gpt-5.5`, with local file/search/test/GitHub tools.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -19209,18 +19209,24 @@ def _handle_chat_start(handler, body, diag=None):
             if "model_provider" in body
             else getattr(s, "model_provider", None)
         )
-        _pp_provider, _pp_default = _read_profile_model_config(s, requested_provider)
         explicit_model_pick = bool(body.get("explicit_model_pick"))
         moa_config = None
         if body.get("moa_config"):
-            if webui_gateway_chat_enabled(get_config()):
-                return bad(handler, "MoA override is unavailable on gateway-backed sessions", 409)
             from api.commands import resolve_moa_config
 
             try:
                 moa_config = resolve_moa_config()
             except RuntimeError as e:
                 return bad(handler, str(e), 503)
+            if webui_gateway_chat_enabled(get_config()):
+                requested_model = str(
+                    moa_config.get("preset") or moa_config.get("default_preset") or ""
+                ).strip()
+                if not requested_model:
+                    return bad(handler, "MoA preset is not configured", 503)
+                requested_provider = "moa"
+                explicit_model_pick = True
+        _pp_provider, _pp_default = _read_profile_model_config(s, requested_provider)
         diag.stage("resolve_model_provider") if diag else None
         model, model_provider, normalized_model = _resolve_compatible_session_model_state(
             requested_model,
--- a/api/routes.py
+++ b/api/routes.py
@@ -19211,6 +19211,7 @@ def _handle_chat_start(handler, body, diag=None):
         )
         explicit_model_pick = bool(body.get("explicit_model_pick"))
         moa_config = None
+        gateway_chat_enabled = webui_gateway_chat_enabled(get_config())
         if body.get("moa_config"):
             from api.commands import resolve_moa_config
 
@@ -19218,7 +19219,7 @@ def _handle_chat_start(handler, body, diag=None):
                 moa_config = resolve_moa_config()
             except RuntimeError as e:
                 return bad(handler, str(e), 503)
-            if webui_gateway_chat_enabled(get_config()):
+            if gateway_chat_enabled:
                 requested_model = str(
                     moa_config.get("preset") or moa_config.get("default_preset") or ""
                 ).strip()
@@ -19239,18 +19240,22 @@ def _handle_chat_start(handler, body, diag=None):
         # with start_session_turn so both entry points behave identically
         # under runtime_adapter_enabled() / runtime_adapter_runner_enabled()
         # — Q-2979-A2 / Copilot discussion_r3305864087/r3305864173).
+        start_run_kwargs = {
+            "msg": msg,
+            "attachments": attachments,
+            "workspace": workspace,
+            "model": model,
+            "model_provider": model_provider,
+            "normalized_model": normalized_model,
+            "source": "webui",
+            "route": "/api/chat/start",
+            "diag": diag,
+        }
+        if not gateway_chat_enabled and moa_config is not None:
+            start_run_kwargs["moa_config"] = moa_config
         response = _start_run(
             s,
-            msg=msg,
-            attachments=attachments,
-            workspace=workspace,
-            model=model,
-            model_provider=model_provider,
-            normalized_model=normalized_model,
-            source="webui",
-            route="/api/chat/start",
-            diag=diag,
-            moa_config=moa_config,
+            **start_run_kwargs,
         )
         # Map adapter-selection NotImplementedError (501) onto the legacy
         # bad-request response shape that this route exposed historically
--- a/api/routes.py
+++ b/api/routes.py
@@ -19213,20 +19213,14 @@ def _handle_chat_start(handler, body, diag=None):
         moa_config = None
         gateway_chat_enabled = webui_gateway_chat_enabled(get_config())
         if body.get("moa_config"):
+            if gateway_chat_enabled:
+                return bad(handler, "MoA override is unavailable on gateway-backed sessions", 409)
             from api.commands import resolve_moa_config
 
             try:
                 moa_config = resolve_moa_config()
             except RuntimeError as e:
                 return bad(handler, str(e), 503)
-            if gateway_chat_enabled:
-                requested_model = str(
-                    moa_config.get("preset") or moa_config.get("default_preset") or ""
-                ).strip()
-                if not requested_model:
-                    return bad(handler, "MoA preset is not configured", 503)
-                requested_provider = "moa"
-                explicit_model_pick = True
         _pp_provider, _pp_default = _read_profile_model_config(s, requested_provider)
         diag.stage("resolve_model_provider") if diag else None
         model, model_provider, normalized_model = _resolve_compatible_session_model_state(
--- a/tests/test_issue5057_moa_webui_route.py
+++ b/tests/test_issue5057_moa_webui_route.py
@@ -122,16 +122,15 @@ def test_moa_config_is_per_turn_not_persisted():
     routes_source = routes_path.read_text(encoding="utf-8")
     assert re.search(r"if body\.get\(\"moa_config\"\):[\s\S]*?moa_config = resolve_moa_config\(\)", routes_source), \
         "chat-start must re-resolve MoA config server-side instead of trusting the browser payload"
-    assert "MoA override is unavailable on gateway-backed sessions" not in routes_source
-    assert 'requested_provider = "moa"' in routes_source
+    assert "MoA override is unavailable on gateway-backed sessions" in routes_source
     js_path = Path(__file__).resolve().parent.parent / "static" / "messages.js"
     js_source = js_path.read_text(encoding="utf-8")
     assert "moa_config:_pendingMoaConfig?true:undefined" in js_source
     assert "_pendingMoaConfig=null" in js_source
 
 
-def test_moa_gateway_chat_start_routes_to_moa_provider(monkeypatch, tmp_path):
-    """Gateway-backed WebUI sessions should send /moa as provider=moa + preset model."""
+def test_moa_gateway_chat_start_fails_closed(monkeypatch, tmp_path):
+    """Gateway-backed WebUI sessions must reject /moa until the gateway consumes runtime overrides."""
     import api.commands as commands
     import api.routes as routes
 
@@ -162,25 +161,18 @@ def test_moa_gateway_chat_start_routes_to_moa_provider(monkeypatch, tmp_path):
         context_messages = []
         pending_user_message = None
 
-    captured = {}
+    def start_run(*_args, **_kwargs):  # pragma: no cover - should fail before run start
+        raise AssertionError("gateway-backed /moa must fail closed before starting a run")
 
-    def start_run(session, **kwargs):
-        captured["session"] = session
-        captured["kwargs"] = kwargs
-        return {"stream_id": "stream-moa", "session_id": session.session_id}
-
-    def resolve_model(model, provider, **_kwargs):
-        captured["resolve_model"] = (model, provider)
-        return model, provider, False
+    def resolve_moa_config():  # pragma: no cover - should fail before resolving MoA
+        raise AssertionError("gateway-backed /moa must fail before resolving MoA config")
 
     monkeypatch.setattr(routes, "get_session", lambda _sid: _Session())
     monkeypatch.setattr(routes, "_resolve_chat_workspace_with_recovery", lambda _s, _w: str(tmp_path))
-    monkeypatch.setattr(routes, "_read_profile_model_config", lambda _s, _provider: (None, None))
-    monkeypatch.setattr(routes, "_resolve_compatible_session_model_state", resolve_model)
     monkeypatch.setattr(routes, "_start_run", start_run)
     monkeypatch.setattr(routes, "get_config", lambda: {"chat_backend": "gateway"})
     monkeypatch.setattr(routes, "webui_gateway_chat_enabled", lambda _cfg: True)
-    monkeypatch.setattr(commands, "resolve_moa_config", lambda: {"preset": "moa-fast", "default_preset": "moa-fast"})
+    monkeypatch.setattr(commands, "resolve_moa_config", resolve_moa_config)
 
     handler = _Handler()
     routes._handle_chat_start(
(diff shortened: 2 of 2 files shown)


ruizanthony:
Paired Hermes Agent/API server PR: https://github.com/NousResearch/hermes-agent/pull/54480

nesquena-hermes:
Triage review of this PR + the paired agent contract it depends on.

## Summary

The WebUI side of this change is coherent: `_handle_chat_start` now resolves the MoA preset server-side and, for gateway-backed sessions, sets `requested_provider = "moa"` / `requested_model = <preset>` instead of returning the old 409. That signal does reach the gateway request body. The problem is the other half of the contract your PR body flags as a follow-up — **the current Hermes Agent `/v1/runs` handler ignores both `provider` and `model` from the request body**, so a gateway `/moa` turn would launch under the gateway's configured default model, not the MoA preset.

## Code reference (WebUI side — correct)

`api/routes.py:17943-17958` (PR HEAD) routes the override cleanly:

```python
if body.get("moa_config"):
    moa_config = resolve_moa_config()
    ...
    if webui_gateway_chat_enabled(get_config()):
        requested_model = str(moa_config.get("preset") or moa_config.get("default_preset") or "").strip()
        if not requested_model:
            return bad(handler, "MoA preset is not configured", 503)
        requested_provider = "moa"
        explicit_model_pick = True
```

And `api/gateway_chat.py:627-628` / `:691-692` does forward it onto the runs body:

```python
if model_provider:
    body_extras["provider"] = model_provider   # provider="moa"
```

## Diagnosis — the agent side drops it

On the agent side, `gateway/platforms/api_server.py::_handle_runs` (line ~3895) reads `input`, `instructions`, `conversation_history`, `session_id`, and `model` — but `model` is only used for status display:

```python
self._set_run_status(run_id, "queued", ...,
    model=body.get("model", self._model_name))   # display only
```

The actual agent is built by `_create_agent` (api_server.py:1072), which takes **no** `provider`/`model` argument and resolves the model from config:

```python
runtime_kwargs = _resolve_runtime_agent_kwargs()
model = _resolve_gateway_model()    # gateway/run.py:2217 — reads config.yaml only
agent = AIAgent(model=model, **runtime_kwargs, ...)
```

`_resolve_gateway_model()` (gateway/run.py:2217) is config-only — no request-scoped path. So `body["provider"]="moa"` and `body["model"]=<preset>` are both silently dropped, and the run executes on whatever the gateway profile's default model is. The 409 you removed was effectively the honest failure mode; without the paired agent change this becomes a *silent* wrong-model run, which is worse from a user's perspective (`/moa` appears to work but isn't aggregating).

## Recommendation

Two viable paths:

1. **Land the agent change first (or in lockstep).** `_handle_runs` needs to read `body.get("provider")` / `body.get("model")` and thread them into `_create_agent` → `AIAgent(model=..., provider=...)`, with a server-side allowlist so a client can't pick arbitrary providers. The MoA virtual-provider plumbing already exists agent-side (`agent/agent_init.py:722`, `agent/moa_loop.py`, `hermes_cli/moa_config.resolve_moa_preset`), so the wiring is "accept the override + resolve the preset," not new MoA machinery.

2. **If this WebUI PR must merge alone,** keep a guard that fails closed when the gateway can't be confirmed to honor request-scoped provider overrides, rather than removing the 409 outright — otherwise gateway users get a silent downgrade.

## Test note

`tests/test_issue5057_moa_webui_route.py::test_moa_gateway_chat_start_routes_to_moa_provider` only asserts the WebUI maps the turn to `("moa-fast", "moa")` (it monkeypatches `_start_run` and `_resolve_compatible_session_model_state`). That's a fine unit assertion for the WebUI layer, but it can't catch the gap above because the gateway runs API is fully mocked out. A contract test that exercises the agent's `/v1/runs` honoring `provider` would be the thing that actually proves end-to-end MoA on gateway.

nesquena-hermes:

**Certified head:** `sha:6fccfe679766` · **PR:** #5153 · _ruizanthony — "fix: route moa webui turns through gateway (#5057)"_
**Verdict:** The WebUI diff is clean and correct *in isolation* (full suite green, no regressions, rebases cleanly onto master), **but it must not merge as-is.** It removes a fail-closed `409` guard and replaces it with a path that **silently runs MoA on the wrong model**, because the deployed Hermes Gateway agent ignores the request-scoped `provider`/`model` override. Both heavy advisors independently reproduced this; I confirmed it against the live agent code. Blocked on the paired agent change.

### What I ran (isolated worktree `/tmp/wt-gate-5153`)
| Gate | Result |
|------|--------|
| Codex (reproduce, on clean rebase) | **SHIP ONLY WITH FIXES** — real silent-MoA finding (below) |
| Opus (full review) | **REQUEST_CHANGES** — same finding, confirmed vs live agent code |
| Full pytest suite (`-p no:xdist`, rebased head) | 11074 passed, 0 failed (green — but see below) |
| PR's own tests in isolation | 8 passed (but they stub `_start_run` — don't exercise the real gateway path) |
| Live agent-code inspection | **confirms the finding** (see trace) |

### ⛔ Blocking finding (reproduced by Codex + Opus + my own trace)
**`api/routes.py:~17956` — gateway MoA appears to succeed but does not actually run MoA.**
The PR removes the `409 "MoA override is unavailable on gateway-backed sessions"` guard and instead sets `requested_provider="moa"` + `requested_model=preset` for gateway-backed sessions. But the **deployed gateway agent ignores these overrides**:
- `gateway/platforms/api_server.py::_handle_runs` (~L3895-3980) parses `input`, `instructions`, `session_id`, `conversation_history` from the request body — **but never reads `provider`**.
- `_create_agent` (~L1072) builds the agent from `_resolve_gateway_model()` / `_resolve_runtime_agent_kwargs()` — the gateway's own configured model — not the request override. `body["model"]` is used only as status/response *metadata* (`:4004`).
- `run_conversation()` is invoked **without `moa_config`**.

Net effect: a gateway MoA request returns `200` and *looks* successful, but the turn runs on the gateway's default model with no MoA ensemble — a **silent wrong-model run**. The `409` this PR deletes was the only signal protecting gateway users from exactly that. This is a "fix that degrades the feature it touches."

### Why the green suite didn't catch it
The new test `test_moa_gateway_chat_start_routes_to_moa_provider` stubs `_start_run` and only asserts WebUI *forwards* `provider="moa"`. It never exercises the real gateway request, so it proves the WebUI plumbing but not the end-to-end MoA behavior — which is where the regression lives.

### Precise fix-spec (either path)
1. **Keep it fail-closed:** retain the `409` until the gateway is known to honor the override, **OR**
2. **Make it real:** thread `moa_config` into `_run_gateway_chat_streaming`, encode the outbound gateway user content via `hermes_cli.moa_config.build_moa_turn_prompt(...)` while preserving the original `msg_text` for WebUI transcript/display, **and** land the paired agent change (request-scoped `provider`/`model` honored in `_handle_runs`→`_create_agent`, with an allowlist — Opus referenced agent PR #54480) in lockstep.
3. Update `tests/test_issue5057_moa_webui_route.py` to assert the gateway-bound request carries a decodable MoA turn (or that the gateway receives/uses `moa_config`), not just that WebUI forwarded an ignored `provider="moa"`.

### Note for the next agent
This is also a **product/concept call** (#5057 asks for MoA-through-gateway). The engineering blocker is the agent-side dependency — the WebUI half alone is a silent regression. Do **not** merge until the gateway honors the override or the fail-closed guard is restored. I did **not** apply `changes-requested` (that's warm-up's auto-bounce role); recording `gate-fail` so the release agent/warm-up re-gate picks it up after the dependency lands.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close. Cert valid only at `sha:6fccfe679766`; a new push or the landing of the paired agent change invalidates it → re-gate._

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (round 2: the silent-wrong-model bug PERSISTS — the deployed gateway structurally ignores the provider/model override)

**Certified head:** `sha:3886e505` (rebased onto current master, `git apply` clean) · **PR:** #5153 · _route MoA WebUI turns through gateway_
**Verdict:** The re-push changed approach (map to `provider="moa"`+preset instead of removing the 409) and its WebUI unit test passes — but the full gate confirms my original CORE **still stands**: the *deployed gateway API server* structurally ignores the request's provider/model, so gateway-backed MoA still silently runs the gateway's default model — now *worse* because it persists `model_provider="moa"` while actually running the default. Restore the fail-closed 409 until the gateway honors the override.

### What I ran (rebased worktree `/tmp/wt-rebase-5153`)
| Gate | Result |
|------|--------|
| Codex (reproduce, traced into the installed gateway) | SHIP-WITH-FIXES — CORE: gateway ignores provider/model; I confirmed by direct inspection |
| Full pytest suite | **11541 passed**, 0 failed (WebUI unit test asserts the ROUTE, can't see the gateway end) |
| PR's MoA route tests | ✅ 8/8 (but they mock the gateway — they prove routes.py *sends* provider=moa, not that the gateway *consumes* it) |

### Findings
**⛔ CORE (CONFIRMED, my original finding persists) — gateway-backed `/moa` still silently runs the wrong model** (`api/routes.py:19222` → the deployed gateway): the re-push now maps the turn to `model="moa-fast", model_provider="moa"` and sends `provider` to the gateway. But the **installed Gateway API server ignores request provider/runtime model when creating the agent** — I verified directly:
- `_create_agent()` (`~/.hermes/hermes-agent/gateway/platforms/api_server.py:1072`) has **no `provider`/`model` parameter**; its docstring: "Create an AIAgent instance using **the gateway's runtime config** … picks up model … from config.yaml/env vars."
- `/v1/runs` (:4042) calls `_create_agent(ephemeral_system_prompt=…, session_id=…, callbacks…)` — **no provider/model passed**.
- `/v1/chat/completions` (:1963) stores `body["model"]` only as a response *label*.
So `provider="moa"`+preset is dropped; the gateway runs its default model. Worse than before: `api/gateway_chat.py:849` persists `s.model_provider="moa"` even though the actual run used the default provider/model — so the UI now *claims* MoA while silently not running it. The green suite can't catch this (the unit test mocks the gateway; it asserts routes.py sends `("moa-fast","moa")`, not that the gateway acts on it). NOTE: `moa` IS a valid virtual provider in the agent (CLI list-picker), but the **gateway API-server path** doesn't thread provider/model into `_create_agent`, so the virtual-provider resolution never happens for gateway runs.
- **Fix (Codex-exact):** restore the fail-closed 409 for `moa_config` when `gateway_chat_enabled` **unless/until the Gateway actually accepts provider/model overrides and fails closed for unsupported `provider="moa"`**; then add an E2E test asserting the Gateway creates `AIAgent(provider="moa", model=<preset>)` or rejects the request. (The real feature needs the gateway API server to thread request provider/model into `_create_agent` — a larger change than the WebUI route.)

**RED — gate-fail/changes-requested (round 2). The approach shifted but the bug is the same: routing `provider="moa"` to a gateway that can't consume it = silent-wrong-model, now with a misleading persisted `model_provider="moa"`.** Simplest correct fix: restore the fail-closed 409 (MoA override unavailable on gateway-backed sessions) so users get an honest error instead of a wrong model — the PR's original 409 was correct. Making gateway MoA actually work requires the gateway API server to accept provider/model overrides (out of this PR's WebUI scope). concept-flag (crit=3). Author is iterating responsively but this needs either the 409 back or a gateway-side change + E2E proof. (Good gate catch: the WebUI unit test's green is vacuous here — it mocks the exact boundary where the bug lives; only tracing into the deployed gateway reveals the override is dropped.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; original silent-wrong-model finding re-confirmed by tracing the installed gateway API server (_create_agent takes no provider/model; /v1/runs passes none). Cert valid for sha:3886e505._

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (bounce closed, round 3 — fail-closed restored)

**Certified head:** `sha:f2d679ee` (clean rebase, branch `gate-rebase/5153-moa-gateway-failclosed`) · **PR:** #5153 · _route MoA WebUI turns through gateway_
**Verdict:** Bounce closed after 3 rounds. My silent-wrong-model CORE (the deployed gateway drops the provider/model override) is now correctly handled by **restoring the fail-closed 409** for gateway-backed MoA — an honest error instead of silently running the wrong model. Codex SAFE, suite green, no silent path remains. Ready to merge from the rebase branch.

### What I ran (rebased worktree `/tmp/wt-rebase-5153b`)
| Gate | Result |
|------|--------|
| Full pytest suite | **11576 passed**, 0 failed |
| PR's MoA route tests | ✅ 8/8 (incl. the new fail-closed test) |

### Findings
**✅ CONVERGED over 3 rounds — the silent-wrong-model bug is closed via fail-closed:**
- round 1: PR removed the fail-closed 409 → gateway ran MoA on the wrong model silently.
- round 2: PR tried mapping `provider="moa"`+preset, but I traced the deployed gateway API server (`_create_agent` takes no provider/model; `/v1/runs` passes none) and confirmed the override is still dropped — same silent-wrong-model, worse (persisted `model_provider="moa"`).
- round 3 (this): the contributor **restored the fail-closed 409** — `routes.py:19213-19214` `if gateway_chat_enabled: return bad(handler, "MoA override is unavailable on gateway-backed sessions", 409)`. So gateway-backed MoA now returns an honest 409 BEFORE resolving MoA config or starting a run — no silent-wrong-model path. Verified by the new `test_moa_gateway_chat_start_fails_closed` (asserts 409 + raises if a run starts or config resolves) and my read. Non-gateway MoA still works (`moa_config` passed to `_start_run` only when NOT gateway, routes.py:19245).

This is the correctly-scoped fix: making the gateway actually *run* MoA requires the gateway API server to accept provider/model overrides (out of this WebUI PR's scope). Failing closed with a clear error is the right WebUI-side behavior until then. Codex SAFE, suite 11576/0.

**Ready to merge — use branch `gate-rebase/5153-moa-gateway-failclosed` (`sha:f2d679ee`), NOT the PR's stale head `b7cdab86`.** Three rounds converged: the silent-wrong-model bug is closed by restoring the fail-closed 409 (honest error), verified no silent path remains + non-gateway MoA preserved; Codex SAFE + suite 11576/0. Backend routing/guard fix — no visible surface. concept 4/5 (routing MoA through the gateway is a real need; the honest-error fallback is the right interim until the gateway supports overrides — worth a follow-up issue for the gateway-side provider/model plumbing). Credit the contributor (co-authored). crit=3. (Good gate arc: rounds 1-2 caught the silent-wrong-model via tracing the deployed gateway; round 3 the contributor landed the honest-error fix.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; fail-closed 409 verified (no silent-wrong-model path, non-gateway MoA preserved), Codex SAFE + suite 11576/0. Cert valid for sha:f2d679ee._

nesquena-hermes:
Shipped in **v0.51.812** — thanks @ruizanthony! 🎉

MoA overrides now fail closed on gateway-backed sessions (honest `409` instead of a silent wrong-model run). Combined-stage gate: Codex SAFE-TO-SHIP + full pytest suite 11639 passed + 115/115 targeted tests. Rebased onto current master (the `_read_profile_model_config` 3-value signature change was reconciled cleanly, keeping the `profile_config=_pp_cfg` threading intact).

Issue #3234, opened 2026-05-31
/api/media serves WebUI state/secret files under HERMES_HOME (settings.json, auth.json) to authenticated sessions


nesquena-hermes:
## Summary

`GET /api/media?path=<abs path>` serves files under `HERMES_HOME` (`~/.hermes`) — including WebUI state and secret-ish files — not just media artifacts. Surfaced during the pre-release dual-gate review of #3219 (both the Codex regression gate and the Opus advisor flagged it). **This is pre-existing**, not introduced by #3219 (#3219 only changes the front-end renderer; it does not touch `api/routes.py` or the `/api/media` allowlist), but it's worth fixing on its own.

## What was verified
The `/api/media` handler (`api/routes.py`, `_handle_media` ~line 8001) resolves the requested path and checks it against an allowlist of roots: `{HERMES_HOME, /tmp, active workspace, MEDIA_ALLOWED_ROOTS}`. The user home dir itself is intentionally excluded (comment notes it would expose `~/.ssh`, `~/.aws`, browser profiles), and out-of-allowlist paths (e.g. `file:///etc/passwd`, `~/.ssh/id_rsa`) correctly return 403.

But because the **whole `HERMES_HOME` (`~/.hermes`) is in the allowlist**, files like these are served (verified against isolated state):
- `$HERMES_HOME/webui/settings.json` → **HTTP 200**
- `$HERMES_HOME/.../auth.json` → served as a download **attachment** (non-image MIME → `disposition=attachment`, so not inline-rendered, but still downloadable)

## Why it matters
Any path that reaches `/api/media` can be emitted by an assistant/tool stream (tool output can be influenced by untrusted content the model ingests). Both the existing `MEDIA:` token path and markdown `[label](file://...)` anchors already route to `/api/media`, and #3219 extends this to bare `file://` URLs — so the reachability is broad. The endpoint is auth-gated when auth is enabled, which limits this to authenticated sessions, but a logged-in user's session viewing attacker-influenced agent output could be induced to fetch WebUI state/secret files.

## Suggested fix
Tighten `/api/media` confinement so WebUI state/secret files under `HERMES_HOME`/`STATE_DIR` cannot be served:
- Add a hard deny for `STATE_DIR` and known secret/config filenames before serving: `settings.json`, `auth.json`, `.signing_key`, `.pbkdf2_key`, `.sessions.json`, `.env`, `config.yaml`, etc.; OR
[...]

PR #5157, opened 2026-06-28
Fix session-backed HTML media artifact previews

santastabber:
## Thinking Path

- Hermes WebUI already supports sandboxed inline previews for local HTML artifacts through `/api/media`.
- Session-scoped `MEDIA:` authorization only covered exact image paths, so assistant-emitted HTML/PDF artifacts outside the active workspace could fail even though they were present in the transcript.
- The chat preview loaders also rebuilt `/api/media` URLs without carrying the active `session_id`, so the backend could not prove the artifact was session-emitted.
- This PR keeps the existing deny-list and sandboxing model, but extends the session-token authorization path to safe preview MIME types and makes HTML/PDF preview requests session-aware.

## What Changed

- Generalized the session MEDIA-token helper from image-only to a safe MIME whitelist.
- Kept the existing image-only wrapper for compatibility with existing callers/tests.
- Added `text/html`, PDF, audio, and video preview MIME types to the session-emitted artifact allowlist.
- Updated chat HTML/PDF lazy preview URLs to include the current `session_id` for fetch, open, and fallback download links.
- Added focused regressions for:
  - exact assistant-emitted HTML `MEDIA:` paths outside the workspace;
  - user-authored `MEDIA:` paths staying unauthorized;
  - MIME allowlist enforcement;
  - frontend HTML/PDF preview fetches carrying `session_id`.

## Why It Matters

HTML artifacts generated by an assistant can now open consistently from chat even when they are not under the active workspace, while still preserving the important security boundary:

- user-authored `MEDIA:` paths cannot mint access;
- unmentioned paths remain blocked;
- Hermes state/secrets deny-list still runs before serving;
- HTML continues to use the existing sandboxed preview path.

## Verification

- `./scripts/test.sh tests/test_media_inline.py tests/test_pdf_html_preview.py tests/test_issue1800_file_html_interactions.py -q`
  - `115 passed`
- `node --check static/ui.js`
  - no new violations on added/modified lines

## Risks / Follow-ups

- This intentionally does not allow arbitrary local HTML browsing. The file still needs to be either under an allowed media root or exactly present in an assistant/tool `MEDIA:` token for the requested session.
- SVG remains download-only.
- The PR does not change workspace file browser behavior; it only fixes chat `MEDIA:` artifact preview authorization.

## Model Used

OpenAI GPT-5.5 via OpenRouter, with Hermes Agent tool use for local implementation and verification.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -15081,13 +15081,13 @@ def _message_content_text(content) -> str:
     return str(content or "")
 
 
-def _session_media_token_allows_image_path(sid: str, target: Path, image_mimes: set[str]) -> bool:
-    """Allow exact MEDIA:image paths already present in the requested session."""
+def _session_media_token_allows_path(sid: str, target: Path, allowed_mimes: set[str]) -> bool:
+    """Allow exact safe MEDIA: paths already present in the requested session."""
     sid = str(sid or "").strip()
     if not sid:
         return False
     mime = MIME_MAP.get(target.suffix.lower(), "application/octet-stream")
-    if mime not in image_mimes:
+    if mime not in allowed_mimes:
         return False
     try:
         target_resolved = target.resolve()
@@ -15122,6 +15122,11 @@ def _session_media_token_allows_image_path(sid: str, target: Path, image_mimes: 
     return False
 
 
+def _session_media_token_allows_image_path(sid: str, target: Path, image_mimes: set[str]) -> bool:
+    """Backward-compatible image-only wrapper for existing callers/tests."""
+    return _session_media_token_allows_path(sid, target, image_mimes)
+
+
 def _path_is_within_root(child: Path, root: Path) -> bool:
     """Return True when ``child`` is inside ``root`` without crashing on Windows drives."""
     try:
@@ -15136,7 +15141,7 @@ def _handle_media(handler, parsed):
     Security:
     - Path must resolve to an allowed root (hermes home, /tmp, common dirs)
     - Auth-gated when auth is enabled
-    - Only image MIME types are served inline; all others force download
+    - Safe preview MIME types can render inline when requested; SVG always downloads
     - SVG always served as attachment (XSS risk)
     - No path traversal: resolved path must stay within an allowed root
     - Additional roots can be added via MEDIA_ALLOWED_ROOTS env var
@@ -15210,10 +15215,17 @@ def _handle_media(handler, parsed):
         for root in allowed_roots
         if root.exists()
     )
-    session_media_allowed = _session_media_token_allows_image_path(
+    _AUDIO_VIDEO_PDF_TYPES = {
+        "audio/mpeg", "audio/wav", "audio/x-wav", "audio/mp4", "audio/aac",
+        "audio/ogg", "audio/opus", "audio/flac",
+        "video/mp4", "video/quicktime", "video/webm", "video/ogg",
+        "application/pdf",
+    }
+    _SESSION_MEDIA_TOKEN_TYPES = _INLINE_IMAGE_TYPES | _AUDIO_VIDEO_PDF_TYPES | {"text/html"}
+    session_media_allowed = _session_media_token_allows_path(
         qs.get("session_id", [""])[0],
         target,
-        _INLINE_IMAGE_TYPES,
+        _SESSION_MEDIA_TOKEN_TYPES,
     )
 
     # ── #3234: hard-deny Hermes's own state + secret/config files ────────────
@@ -15399,12 +15411,7 @@ def _handle_media(handler, parsed):
     # Only serve safe media/PDF types inline when explicitly requested. HTML is
     # allowed inline only with a CSP sandbox so "open full page" can work without
     # granting same-origin access to the WebUI. SVG is always a download (XSS risk).
-    _INLINE_PREVIEW_TYPES = _INLINE_IMAGE_TYPES | {
-        "audio/mpeg", "audio/wav", "audio/x-wav", "audio/mp4", "audio/aac",
-        "audio/ogg", "audio/opus", "audio/flac",
-        "video/mp4", "video/quicktime", "video/webm", "video/ogg",
-        "application/pdf",
-    }
+    _INLINE_PREVIEW_TYPES = _INLINE_IMAGE_TYPES | _AUDIO_VIDEO_PDF_TYPES
     _DOWNLOAD_TYPES = {"image/svg+xml"}  # SVG: XSS risk, force download
     inline_preview = qs.get("inline", [""])[0] == "1"
     html_inline_ok = inline_preview and mime == "text/html"
--- a/static/ui.js
+++ b/static/ui.js
@@ -14599,19 +14599,23 @@ function loadPdfInline(container){
     el.setAttribute('data-loaded','1');
     const path=el.dataset.path;
     const fname=path.split('/').pop()||path;
+    const mediaSessionId=(typeof S!=='undefined'&&S&&S.session&&S.session.session_id)?String(S.session.session_id):'';
+    const publicMediaUrl='api/media?path='+encodeURIComponent(path);
+    const mediaUrl=publicMediaUrl+(mediaSessionId?'&session_id='+encodeURIComponent(mediaSessionId):'');
     const loadPdf=(pdfjsLib)=>{
-      fetch('api/media?path='+encodeURIComponent(path))
+      fetch(mediaUrl)
         .then(r=>{if(!r.ok) throw new Error(r.status); return r.arrayBuffer();})
         .then(buf=>{
           if(buf.byteLength>PDF_MAX_SIZE){
-            el.outerHTML=`<div class="pdf-preview-fallback"><a class="msg-media-link" href="api/media?path=${encodeURIComponent(path)}&download=1" download="${esc(fname)}">📎 ${esc(fname)}</a><br><span style="color:var(--muted);font-size:12px">${t('pdf_too_large')}</span></div>`;
+            const dlUrl=publicMediaUrl+'&download=1';
+            el.outerHTML=`<div class="pdf-preview-fallback"><a class="msg-media-link" href="${dlUrl}" download="${esc(fname)}">📎 ${esc(fname)}</a><br><span style="color:var(--muted);font-size:12px">${t('pdf_too_large')}</span></div>`;
             return;
           }
           return pdfjsLib.getDocument({data:buf, isEvalSupported:false}).promise;
         })
         .then(pdf=>{
           if(!pdf) return;
-          const dlUrl='api/media?path='+encodeURIComponent(path)+'&download=1';
+          const dlUrl=publicMediaUrl+'&download=1';
           const total=pdf.numPages;
           const pagesLabel=total>1?` · ${total} pages`:'';
           const wrap=document.createElement('div');
@@ -14650,7 +14654,7 @@ function loadPdfInline(container){
           renderPage(1);
         })
         .catch(()=>{
-          const dlUrl='api/media?path='+encodeURIComponent(path)+'&download=1';
+          const dlUrl=publicMediaUrl+'&download=1';
           el.outerHTML=`<div class="pdf-preview-fallback"><a class="msg-media-link" href="${dlUrl}" download="${esc(fname)}">📎 ${esc(fname)}</a><br><span style="color:var(--muted);font-size:12px">${t('pdf_error')}</span></div>`;
         });
     };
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Triage review — focused on whether widening the session-token MIME allowlist keeps the `/api/media` security boundary intact.

## Summary

Reading `api/routes.py:15084-15123` (the renamed `_session_media_token_allows_path`) and the `_handle_media` flow at `:15218-15402` on the PR HEAD vs `origin/master`, the widening is sound. The change generalizes the image-only token check to a `_SESSION_MEDIA_TOKEN_TYPES` set (adds audio/video/PDF/`text/html`) while keeping every existing guard ahead of the serve decision. The three properties that matter — assistant-only token minting, deny-before-serve ordering, and HTML sandboxing — all still hold.

## Code reference

The token matcher still refuses user-authored MEDIA tokens, so a user can't mint access to an arbitrary path (`api/routes.py:15104-15123`):

```python
role = str(message.get("role") or "").strip().lower()
if role == "user":
    continue
...
for ref in _MEDIA_TOKEN_RE.findall(text):
    if "://" in ref:
        continue
    if Path(ref).expanduser().resolve() == target_resolved:
        return True
```

Critically, the `#3234` state/secret deny-list still runs **before** `session_media_allowed` is consulted. The deny returns 403 at `:15381` and `:15398`, and only then does the allow gate apply at `:15401`:

```python
if any(_within_ci(target, d) for d in _deny_dirs):
    return bad(handler, "Path not in allowed location", 403)
...
if not within_allowed and not session_media_allowed:
    return bad(handler, "Path not in allowed location", 403)
```

So even though a session token can now authorize `text/html`, an assistant-emitted `MEDIA:` pointing at `state.db` / `auth.json` / `config.yaml` / a sibling profile's `sessions/` is still 403'd by the deny pass that executes first. Good — the new MIME types don't widen *which paths* are reachable, only *which already-token-authorized paths render inline*.

## HTML inline path

The inline-disposition logic at `:15421-15431` keeps HTML behind both an explicit `?inline=1` and a CSP sandbox:

```python
html_inline_ok = inline_preview and mime == "text/html"
...
csp = "sandbox allow-scripts" if html_inline_ok else None
return _serve_file_bytes(handler, target, mime, disposition, "private, max-age=3600", csp=csp)
```

`sandbox allow-scripts` (no `allow-same-origin`) means the previewed HTML runs scripts in an opaque origin and can't read the WebUI's cookies/localStorage or call same-origin APIs. SVG stays in `_DOWNLOAD_TYPES` (forced attachment). That's the right posture for assistant-generated HTML.

## One nit

`_SESSION_MEDIA_TOKEN_TYPES` (`:15218`) is a superset of `_INLINE_PREVIEW_TYPES` (`:15414`) — the token-auth set additionally allows `text/html`, while the inline-serve set deliberately excludes it (HTML gets the separate `html_inline_ok` path). That divergence is intentional but easy to misread on a later edit; a one-line comment noting "token-auth allows html; inline-serve gates it separately via html_inline_ok" would save a future reviewer a double-take. Functionally correct as-is.

CI is green across 3.11/3.12/3.13 incl. `browser-smoke`, and the added regressions cover the assistant-vs-user minting distinction and the `session_id`-carrying preview fetch. No blocking concerns from a security-boundary standpoint.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (one low-severity hardening follow-up, non-blocking)

**Certified head:** `17a93388` · **PR:** #5157 · _santastabber — Fix session-backed HTML media artifact previews_
**Verdict:** Full authoritative gate passes. Auth fails closed, HTML render is sandboxed on both paths, full suite green, live HTTP boundary test passes. Ready to merge. One low-severity, non-blocking hardening item is documented below for a follow-up (do **not** apply Codex's literal fix — it would break the feature).

### What I ran (isolated worktree `/tmp/wt-gate-5157` @ `17a93388`, gate port 8783 — not prod)

| Gate | Result |
|------|--------|
| **Full pytest suite** | **10994 passed**, 10 skipped, 1 xfailed, 2 xpassed, 2 failed (`test_skills_stats_cache`, `test_git_stage_unstage_discard_and_commit`) — **both in files this PR does not touch, both pass in isolation** = pre-existing order-dependent flakes, not a regression. CI is authoritative: **11/11 shards + browser-smoke + lint all green.** |
| **Codex (reproduce)** | "SHIP ONLY WITH FIXES" — flagged session_id reaching the "Open full page" URL. Reproduced as real but **low-severity** (see adjudication). |
| **Opus (full review)** | "✅ APPROVE — correct, safe, ready to merge." session_id is an identifier not a credential; Codex's proposed fix would break the out-of-roots open-full-page path. |
| **Live HTTP boundary test** (running server, artifact genuinely outside all roots) | **PASS 4/4**: assistant-emitted HTML + session_id + inline → **200 `text/html` with `Content-Security-Policy: sandbox allow-scripts`**; no session_id → **403**; user-authored MEDIA token → **403**; secret path (`state.db`) → **403** even with a valid session grant. |
| `node -c static/ui.js` | PARSE OK |

### Findings

**✅ Auth boundary fails closed (verified at code + live HTTP).** `_session_media_token_allows_path` returns `False` on every branch: empty sid, mime-not-in-allowlist, resolve exception, `get_session` exception, `role=="user"` messages skipped (user-authored tokens cannot mint a grant), `://` refs skipped, and **exact resolved-path equality** (`Path(ref).resolve() == target_resolved`) — no prefix/`startswith`/`commonpath`, so no traversal escape. The #3234 secret/state hard-deny runs **before** the session-grant decision, so a grant can never reach `sessions/`/`memories/`/secrets/`state.db` (live-confirmed: case D → 403).

**✅ XSS-safe on both render paths.** (a) Chat inline = `<iframe srcdoc=… sandbox="allow-scripts">` (no `allow-same-origin` → opaque origin, can't touch parent DOM/cookies/localStorage); srcdoc escaping ordered `&`→`"`→`<`→`>` so no `srcdoc="…"` breakout. (b) Direct serve = `Content-Security-Policy: sandbox allow-scripts` + `X-Content-Type-Options: nosniff`. Invariant holds: text/html serves inline **iff** the sandbox CSP is set — there is no code path that serves HTML inline without it (live-confirmed: case A carried the CSP header).

**✅ No new cross-session/cross-profile exposure.** Single-user auth principal; `session_id` only selects *which* session's emitted tokens to match. The widening newly reaches only assistant/tool-emitted audio/video/pdf/html artifacts outside the standard roots — the PR's intent.

**⚠️ Low-severity hardening follow-up (NON-BLOCKING — do not block merge on this).** Codex correctly observed that the "Open full page ↗" link and the too-large fallback build `openUrl = mediaUrl + '&inline=1'`, where `mediaUrl` carries `&session_id=<sid>` (`static/ui.js:14702`, `14706`). Clicking it navigates a tab to a `sandbox allow-scripts` (opaque-origin, scriptable) HTML page whose `window.location.search` contains the session_id, readable by the served (tool-authored) HTML.
- **Why it is not ship-blocking:** session_id is the **conversation identifier, not a credential** — it already appears in multiple client URLs, and every endpoint is gated by the independent auth layer (`is_auth_enabled`/cookie/CSRF), so a leaked session_id authenticates nothing. The only capability it confers is requesting *that same session's own already-known* assistant-emitted artifacts (exact-path grant) — and the user was already viewing that session to click the link. No cross-session/cross-profile escalation. The default **inline** preview is `about:srcdoc` and carries no session_id; the leak requires an explicit click on attacker-authored HTML.
- **Do NOT apply Codex's literal fix** (strip session_id from the open-full-page URL): that URL needs the session grant to serve an out-of-roots artifact, so stripping it breaks the exact feature this PR adds (Opus flagged this correctly). A *safe* future hardening, if desired: mint a short-lived opaque one-file token for the open-full-page navigation instead of reusing the session_id query param — separate follow-up, not this PR.

**ℹ️ Cosmetic (non-blocking):** `_SESSION_MEDIA_TOKEN_TYPES` duplicates the audio/video/pdf members of `_INLINE_PREVIEW_TYPES` (could be one constant); the two new JS tests are static source-string assertions (consistent with this repo's convention).

### Recommendation to the next agent (release)

**Ready to merge.** Full gate is green: auth fails closed, render is sandboxed on both paths, full suite passes (the 2 local failures are pre-existing unrelated flakes; CI is 11/11 green), and the live HTTP boundary test passes 4/4. The single security item raised by Codex is **low-severity and non-blocking** (identifier, not credential; behind an explicit click; no escalation) and its literal fix would break the feature — Opus's APPROVE is the correct call on impact. Optionally file a follow-up for the opaque-token hardening of the open-full-page link; do not hold this PR for it. No code changes needed before merge.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close — that's the release agent's call. This certification is valid only while the head stays at `17a93388`._

nesquena-hermes:

**Certified head:** `sha:055e57fac173` · **PR:** #5157 · _Fix session-backed HTML media artifact previews_
**Verdict:** Full authoritative gate is green. The session-media authorization boundary is correct, tight, and a net security improvement. Ready to merge — no engineering blockers; two non-blocking follow-up notes below.

> This re-certifies #5157 at its current head (`055e57fac173`), which was re-pushed since the prior cert (`17a93388`). The re-push generalized the image-only grant to a shared `_session_media_token_allows_path` helper and addressed the earlier session_id-in-navigable-URL note by scoping the token to the sandboxed fetch only.

### What I ran (isolated worktree `/tmp/wt-gate-5157`, gate port 8783)
| Gate | Result |
|------|--------|
| Codex (reproduce, on clean rebase) | **SAFE TO SHIP** — see stale-base note |
| Opus (full review) | **COMMENT** — no blocking issues, 2 follow-up notes |
| Full pytest suite (`-p no:xdist`, to completion) | **11073 passed, 10 skipped, 1 xfailed, 2 xpassed, 16 subtests passed, 0 failed** (418s) |
| Live HTTP boundary test (4+ cases, out-of-roots `/var/tmp` artifact) | **all correct** — see below |
| In-process grant boundary | **5/5** — assistant-path+safe-mime → allow; wrong-session / wrong-mime / secret-path / SVG → deny |

### ⚠️ Important gate note — stale-base artifact (resolved)
A first Codex pass against current `origin/master` returned **KICK BACK TO AUTHOR** with 5 CORE/SILENT findings (streaming.py partial-failure handling, profile default workspace, profile/model repair, `extension_settings.js`, gateway restart). **All five are stale-base artifacts, not defects in this PR** — they describe code from other PRs (incl. #5181) that merged to master *after* this PR's base. Codex itself diagnosed the cause ("This head is stale against `origin/master`"). I verified by rebasing the branch onto current master (`c35835d8`) **locally in my throwaway worktree only** — the rebase is clean (`rc=0`) and the post-rebase diff vs master is **exactly the 4 media-auth files** (`api/routes.py`, `static/ui.js`, `tests/test_media_inline.py`, `tests/test_pdf_html_preview.py`). Re-running Codex on the clean rebase flipped it to **SAFE TO SHIP**. (Also confirms the PR is cleanly mergeable on current master.)

### Live HTTP boundary results (real server, artifact genuinely outside all roots)
| Case | Result | |
|------|--------|--|
| assistant-authored HTML + valid `session_id` + `inline=1` | **200**, `Content-Type: text/html`, `Content-Security-Policy: sandbox allow-scripts` | ✅ correct sandboxed inline grant |
| same HTML, **no** `session_id` | **403** | ✅ token required |
| same HTML, **wrong** `session_id` | **403** | ✅ denied |
| `/etc/passwd` + valid `session_id` | **403** | ✅ secret/out-of-session path denied |
| SVG + valid `session_id` + `inline=1` | **403** (never inline) | ✅ XSS-risk type blocked from inline |
| HTML **without** `inline=1` | **200**, `Content-Disposition: attachment` | ✅ forces download, no inline render |

### Findings
- ✅ **Boundary is exact-path + assistant-authored + safe-mime + session-scoped.** `_session_media_token_allows_path` grants only a path the *assistant* emitted as a `MEDIA:` reference in *that* session, with a mime in the allowed set. User-authored paths, wrong-mime, and cross-session paths are all rejected (unit + live verified). State/secret deny runs *before* the allow decision.
- ✅ **HTML never gets same-origin.** Inline HTML is served only with `Content-Security-Policy: sandbox allow-scripts`; SVG is always an attachment (XSS). The sandboxed iframe carries the `session_id` on its fetch; the user-navigable "open full page" + download links deliberately omit it.
- ✅ **Prior advisor finding addressed without breaking the feature.** The earlier Codex note (session_id reaching a navigable script-capable page) is resolved by scoping the token to the sandboxed fetch — the supported working surface — rather than the self-defeating "strip it everywhere" fix.
- ⚠️ **Non-blocking follow-up (Opus + Codex agree):** the "open full page" / download / large-file fallback links 403 for *out-of-roots* artifacts because they omit `session_id`. This is an intentional tradeoff (the sandboxed inline preview is the supported surface; the prior shipped path didn't authorize these previews at all), not a regression. A future hardening could mint a scoped token for those links.
- ⚠️ **Non-blocking:** the PR description states open/download links carry `session_id`; the code (correctly) does not. Worth a one-line PR-body correction.

**Ready to merge — cert fresh for `sha:055e57fac173`, no further review needed.** The security boundary is intact and verified at the HTTP layer (Codex SAFE on clean rebase, Opus no-blockers, 11073/0 suite, 6/6 live boundary cases, 5/5 in-process grant). The branch rebases cleanly onto current master with a 4-file diff. The two follow-up notes are non-blocking and can be a separate hardening PR. This cert is valid only while the head stays at `sha:055e57fac173`.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close — that's the release agent's call. Cert valid only at `sha:055e57fac173`; a new push invalidates it._

nesquena-hermes:
Shipped in **v0.51.739** (via #5191). Thanks @santastabber — session-backed HTML/PDF artifacts now thread the active session_id into the api/media fetch so they authorize and preview inline again (HTML still sandboxed, SVG stays a download, secret/state hard-deny unchanged). Gate-certified green + maintainer-approved after live browser testing of the inline preview.

PR #5158, opened 2026-06-28
fix(sidebar): keep hidden archived parents as child references

santastabber:
## Thinking Path

- The sidebar already has client-side logic to keep archived parent conversations from leaving child/subagent rows behind.
- The newer source-filtered `/api/sessions?sidebar_source=...` path can omit the hidden archived parent while still returning the visible child row.
- Without the parent as reference context, the browser cannot tell that the child belongs to a hidden archived ancestor, so the child can render as a top-level/orphan sidebar row.
- This PR keeps the visible list unchanged while sending a tiny reference-only ancestor payload for the sidebar's existing nesting/suppression logic.

## What Changed

- Adds `sidebar_reference_sessions` to the `/api/sessions` payload for hidden archived ancestors needed by visible child rows.
- Marks those rows `_sidebar_reference_only` after applying the normal bounded sidebar response shape/redaction path.
- Stores reference rows separately on the frontend and passes them only into child-session attachment/suppression.
- Adds regression coverage for the source-filtered API payload and the frontend render path.

## Why It Matters

Archiving a parent conversation should not leave delegated/subagent/fork child rows cluttering the sidebar when archived sessions are hidden. This preserves the existing no-cascade behavior while giving the frontend enough context to apply the already-existing archive/child-session rule.

## Verification

- `python3 -m py_compile api/routes.py api/agent_sessions.py api/models.py`
- `node --check static/*.js`
- `./scripts/test.sh tests/test_session_lineage_collapse.py tests/test_issue4766_sidebar_source_pushdown.py -q` — 61 passed
- `./scripts/test.sh tests/test_sidebar_session_partition.py tests/test_session_lineage_collapse.py tests/test_issue4766_sidebar_source_pushdown.py tests/test_issue4385_cron_archive_reappears.py tests/test_session_sidebar_cache.py tests/test_issue4662_sidebar_redaction_read_once.py -q` — 90 passed

## Risks / Follow-ups

- The new reference payload is intentionally narrow: only hidden archived ancestors for currently visible rows. It should not make archived rows render, and it avoids cascade-mutating child session state.
- If future sidebar filters need similar hidden context, they should use the same reference-only pattern rather than widening the primary `sessions` list.

## Model Used

AI-assisted using Hermes Agent with GPT-5.5 via OpenRouter. Verification was performed locally with the commands above.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -1908,10 +1908,20 @@ def _build_session_list_cache_payload(
         1 for s in scoped
         if _is_cli_session_for_settings(s)
     )
-    if sidebar_source == "webui":
-        scoped = [s for s in scoped if not _is_cli_session_for_settings(s)]
-    elif sidebar_source == "cli":
-        scoped = [s for s in scoped if _is_cli_session_for_settings(s)]
+    def _filter_sidebar_source(rows: list[dict]) -> list[dict]:
+        if sidebar_source == "webui":
+            return [s for s in rows if not _is_cli_session_for_settings(s)]
+        if sidebar_source == "cli":
+            return [s for s in rows if _is_cli_session_for_settings(s)]
+        return list(rows)
+
+    scoped = _filter_sidebar_source(scoped)
+    sidebar_reference_sessions: list[dict] = []
+    if not include_archived:
+        sidebar_reference_sessions = _hidden_archived_sidebar_reference_sessions(
+            _filter_sidebar_source(visible_scoped),
+            _filter_sidebar_source(archived_scoped),
+        )
     if not include_archived:
         diag_stage("filter_archived_sessions")
     diag_stage("visible_lineage_metadata")
@@ -1920,6 +1930,10 @@ def _build_session_list_cache_payload(
         "sessions": [
             dict(s) if isinstance(s, dict) else {}
             for s in scoped
+        ],
+        "sidebar_reference_sessions": [
+            dict(s) if isinstance(s, dict) else {}
+            for s in sidebar_reference_sessions
         ],
         "cli_count": len(deduped_cli),
         "archived_count": archived_count,
@@ -1957,8 +1971,15 @@ def _session_list_payload_to_response(payload: dict) -> dict:
     for s in runtime_rows:
         item = _sidebar_session_response_item(s, redact_enabled=_redact_enabled) if isinstance(s, dict) else {}
         safe_merged.append(item)
+    safe_reference = []
+    for s in payload.get("sidebar_reference_sessions", []) or []:
+        item = _sidebar_session_response_item(s, redact_enabled=_redact_enabled) if isinstance(s, dict) else {}
+        if item:
+            item["_sidebar_reference_only"] = True
+        safe_reference.append(item)
     response = {
         "sessions": safe_merged,
+        "sidebar_reference_sessions": safe_reference,
         "cli_count": int(payload.get("cli_count", 0)),
         "archived_count": int(payload.get("archived_count", 0)),
         "archived_webui_count": int(payload.get("archived_webui_count", 0)),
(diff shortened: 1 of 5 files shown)


nesquena-hermes:
Triage review — traced the new reference payload from the API through the sidebar render path to confirm hidden archived ancestors can't leak back into the visible list.

## Summary

Reading `api/routes.py:2001-2046` (`_hidden_archived_sidebar_reference_sessions`) plus the consuming `static/sessions.js:5994-6001` and `_attachChildSessionsToSidebarRows` at `:5479-5625`, the design holds: the reference rows feed only the *nesting/suppression* inputs, never the rendered set. The fix correctly closes the gap where a source-filtered `/api/sessions?sidebar_source=…` payload omits the archived parent but still returns the visible child, leaving it orphaned at top level.

## Why it can't leak a rendered row

On the backend, the ancestor walk is cycle-safe and only emits parents that aren't already visible (`api/routes.py:2030-2044`):

```python
while parent_id and parent_id not in seen:
    seen.add(parent_id)
    if parent_id in visible_ids:
        break
    parent = archived_by_id.get(parent_id)
    if not parent:
        break
    ...
    parent_id = str(parent.get("parent_session_id") or "").strip()
```

On the frontend, `_sidebarReferenceSessions` is appended only to `referenceRows` (the 3rd arg → `rawSessionsById`/`sessionIdsInList`), not to `sessionsRaw` (`static/sessions.js:5997-6001`):

```js
referenceRows=[...referenceRows,..._sidebarReferenceSessions];
return _attachChildSessionsToSidebarRows(_collapseSessionLineageForSidebar(sessionsRaw), sessionsRaw, referenceRows);
```

The render-eligibility gate `renderableChildIds` is built from `rawSessions` only (`:5572`), so a reference-only candidate has `childRenderable=false`. That means it can only ever take the `hasHiddenArchivedAncestor → hiddenArchivedChildTree.add()` suppression branch (`:5604-5606`); the orphan-promotion branch is guarded by `else if(childRenderable)` (`:5621`), so a reference row can never be pushed as a visible top-level orphan. The final `[...rows, ...orphans]` derives `rows` from `collapsedRows`, which is `_collapseSessionLineageForSidebar(sessionsRaw)` — reference rows aren't in it. So the invariant "reference rows nest/suppress but never render" is structurally enforced, not just conventionally.

## One minor observation

`_sidebar_reference_only` is stamped on each reference item in `_session_list_payload_to_response` (`api/routes.py:1976-1979`) but isn't read anywhere under `static/` — the suppression relies entirely on `renderableChildIds` membership instead. Not a bug; the flag is effectively a defensive/forward-compat marker today. Either wire a frontend assertion against it (belt-and-suspenders: drop any row carrying the flag from the renderable set) or add a one-line comment that it's intentionally backend-only metadata, so a future reader doesn't assume the frontend enforces it.

CI is green across 3.11/3.12/3.13 + browser-smoke, and the added `test_issue4766_sidebar_source_pushdown.py` / `test_session_lineage_collapse.py` cases cover both the payload shape and the render path. No blocking concerns.

nesquena-hermes:
🚀 Shipped in **v0.51.731**.

Thanks @santastabber — your fix for the archived-parent orphan regression (child/subagent/fork rows of a hidden archived parent rendering as orphan top-level conversations) is live.

During the merge gate I caught and fixed one follow-on edge case on your branch (commit preserves your authorship via `Co-authored-by`):

**Cross-project suppression leak.** `_renderSidebarRowsFromRawSessions()` appended the new `sidebar_reference_sessions` global to *every* render, but that payload arrives from `/api/sessions` without the client-side project/source scoping that `_partitionSidebarSessionRows` applies to visible rows. So with a project filter active, an archived parent in **project B** entered a **project-A**-filtered render's suppression context and silently hid a visible project-A child/fork (it rendered as `[]`). Reproduced in Node before fixing.

**Fix:** added `_scopedSidebarReferenceRows(isCli)` which filters references to the active project (mirroring `_partitionSidebarSessionRows`: `NO_PROJECT_FILTER` → only refs without `project_id`, else `=== _activeProject`) **and** the webui/cli source bucket, then the 3 render/count callers pass scoped refs. A regression test locks the cross-project case; the existing suppression test now exercises the scoped helper.

Gate: Codex security pass SAFE TO SHIP (leak fixed, the #4293 same-project/no-filter suppression you implemented still works, webui/cli refs don't cross-feed, no new regressions); CI 11/11 green; full suite 11031 passed. Deployed via systemd; served bytes confirm the fix.

PR #5166, opened 2026-06-28
fix(tests): accept Windows venv layout in scripts/test.sh (#5152)

### LINKED ISSUE #5152: scripts/test.sh rejects Windows venv layout under Git Bash
`./scripts/test.sh` is not usable from Windows Git Bash right now. It creates `.venv` with a supported Python, then rejects that environment because the script hardcodes the POSIX interpreter path `.venv/bin/python`, while a normal Windows venv puts the interpreter at `.venv/Scripts/python.exe`.

## Repro

In Git Bash on Windows, from a clean checkout:

```bash
./scripts/test.sh tests/test_issue5121_provider_auth_terminal_error.py -v --timeout=60
```

I get:

```text
Creating .venv with /c/Users/Rod/.local/bin/python3.11 (3.11.15).
/d/Repos/hermes-webui/.venv was created but does not contain bin/python.
Could not create a working .venv with /c/Users/Rod/.local/bin/python3.11 (3.11.15).
Install the matching Python venv/ensurepip package (for example python3.x-venv on Debian/Ubuntu)
or set HERMES_WEBUI_TEST_PYTHON to a supported Python 3.11, 3.12, or 3.13 interpreter.
```

Forcing an explicit Windows interpreter does not help:

```bash
HERMES_WEBUI_TEST_PYTHON=/c/Apps/Python313/python.exe ./scripts/test.sh tests/test_issue5121_provider_auth_terminal_error.py -v --timeout=60
```

That still fails before pytest starts:

```text
Creating .venv with /c/Apps/Python313/python.exe (3.13.13).
Unable to copy 'C:\\Apps\\Python313\\Lib\\venv\\scripts\\nt\\venvlauncher.exe' to 'D:\\Repos\\hermes-webui\\.venv\\Scripts\\python.exe'
/d/Repos/hermes-webui/.venv was created but does not contain bin/python.
Could not create a working .venv with /c/Apps/Python313/python.exe (3.13.13).
```

## Root cause

`scripts/test.sh:5-6` sets:

```bash
VENV_DIR="$ROOT/.venv"
VENV_PY="$VENV_DIR/bin/python"
```

and `scripts/test.sh:118-120` treats the absence of that POSIX path as a failed venv bootstrap:

```bash
if [[ ! -x "$VENV_PY" ]]; then
  rm -rf "$VENV_DIR"
  echo "$VENV_DIR was created but does not contain bin/python." >&2
```

That is fine on Linux and macOS, but it rejects a normal Windows venv even when `python -m venv` succeeded, because the interpreter lives at `.venv/Scripts/python.exe`.

This looks separate from https://github.com/nesquena/hermes-webui/issues/3664. That issue is about local Windows full-suite usefulness and cleanup. This one blocks the repo's own focused test entrypoint before pytest starts.

## Suggested fix

Resolve the venv interpreter by platform, or probe both standard layouts:

- POSIX: `.venv/bin/python`
- Windows: `.venv/Scripts/python.exe`

The rest of the bootstrap logic can stay the same once `VENV_PY` points at the interpreter that `python -m venv` actually created on the current platform.

## Environment

Windows 11, Git Bash from `C:\Apps\Git\usr\bin\bash.exe`, Python 3.11.15 at `/c/Users/Rod/.local/bin/python3.11`, Python 3.13.13 at `/c/Apps/Python313/python.exe`.

rodboev:
## Thinking Path

- `./scripts/test.sh` is the documented local pytest entrypoint, but on Windows it created a valid `.venv` and then threw it away because the runner only recognized `.venv/bin/python`.
- The bug sits at the runner’s ownership boundary, not in any one test: every downstream check reused the same hardcoded interpreter path for create, reuse, dependency install, and final pytest execution.
- The fix keeps the runner as the single contract surface, resolves the standard POSIX and Windows venv interpreter paths in one place, and adds a focused regression that proves the runner reaches pytest when a Windows-layout venv is created.

## What Changed

- `scripts/test.sh`: resolve the repo-local venv interpreter from `.venv/bin/python` or `.venv/Scripts/python.exe`, then reuse that resolved path across the existing version, pip, dependency, rebuild, and pytest-exec checks.
- `tests/test_ci_hygiene.py`: update the runner contract assertions and add a targeted behavior test that simulates `python -m venv` creating only `.venv/Scripts/python.exe`.

## Why It Matters

Windows contributors can use the documented `./scripts/test.sh` workflow from Git Bash without the runner deleting a valid venv and failing before pytest starts. The existing safety checks still gate unsupported interpreters, broken venvs, pip failures, and symlinked `.venv` create-or-clear paths.

## Verification

```bash
./scripts/test.sh tests/test_ci_hygiene.py -v --timeout=60
```

Manual Windows Git Bash repro:

```bash
./scripts/test.sh tests/test_issue5121_provider_auth_terminal_error.py -v --timeout=60
```

Full-suite CI context, not a required local check unless requested: `./scripts/test.sh tests/ -v --timeout=60`.

## Upstream

Closes #5152.

## Model Used

GPT-5.5 via Codex CLI

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -3,8 +3,20 @@ set -euo pipefail
 
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 VENV_DIR="$ROOT/.venv"
-VENV_PY="$VENV_DIR/bin/python"
 REQ_FILE="$ROOT/requirements-dev.txt"
+
+resolve_venv_python() {
+  local candidate
+  for candidate in "$VENV_DIR/bin/python" "$VENV_DIR/Scripts/python.exe"; do
+    if [[ -x "$candidate" ]]; then
+      printf '%s\n' "$candidate"
+      return 0
+    fi
+  done
+  return 1
+}
+
+VENV_PY="$(resolve_venv_python || true)"
 
 is_supported_python() {
   "$1" - <<'PY' >/dev/null 2>&1
@@ -115,9 +127,10 @@ create_or_rebuild_venv() {
     venv_guidance "$base_py"
     return 2
   fi
+  VENV_PY="$(resolve_venv_python || true)"
   if [[ ! -x "$VENV_PY" ]]; then
     rm -rf "$VENV_DIR"
-    echo "$VENV_DIR was created but does not contain bin/python." >&2
+    echo "$VENV_DIR was created but does not contain bin/python or Scripts/python.exe." >&2
     venv_guidance "$base_py"
     return 2
   fi
@@ -157,6 +170,7 @@ select_python() {
     desired_major_minor=""
   fi
 
+  VENV_PY="$(resolve_venv_python || true)"
   if [[ -x "$VENV_PY" ]]; then
     if is_supported_python "$VENV_PY"; then
       current_major_minor="$(python_major_minor "$VENV_PY")"
@@ -182,7 +196,7 @@ select_python() {
   fi
 
   if [[ -e "$VENV_DIR" && ! -x "$VENV_PY" ]]; then
-    echo "$VENV_DIR exists but does not contain bin/python; rebuilding." >&2
+    echo "$VENV_DIR exists but does not contain bin/python or Scripts/python.exe; rebuilding." >&2
     create_or_rebuild_venv "$base_py" rebuild || return $?
     printf '%s\n' "$VENV_PY"
     return 0
--- a/tests/test_ci_hygiene.py
+++ b/tests/test_ci_hygiene.py
@@ -1,8 +1,16 @@
 """Small hygiene regression checks for CI and frontend console noise."""
 
+import os
+import shutil
+import stat
+import subprocess
+import textwrap
 from pathlib import Path
 
 ROOT = Path(__file__).resolve().parents[1]
+
+def _make_executable(path):
+    path.chmod(path.stat().st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
 
 def test_github_actions_quotes_pyyaml_version_specifier():
@@ -29,6 +37,8 @@ def test_local_test_runner_uses_supported_venv_before_pytest_collection():
     assert "python3.13 python3.12 python3.11 python3" in runner
     assert "requirements-dev.txt" in runner
     assert 'HERMES_WEBUI_TEST_PYTHON' in runner
+    assert "resolve_venv_python()" in runner
+    assert '"$VENV_DIR/bin/python" "$VENV_DIR/Scripts/python.exe"' in runner
     assert '[[ -x "$candidate" ]] && printf' in runner
     assert "exec \"$PYTHON_BIN\" -m pytest" in runner
     # Destructive-fs guard: never create/clear a virtualenv through a symlinked .venv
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Reading the full `scripts/test.sh` at this PR's HEAD against `origin/master`, plus the new tests in `tests/test_ci_hygiene.py`, this is a clean, correct implementation of the fix #5152 suggested (probe both `.venv/bin/python` and `.venv/Scripts/python.exe`). Walking the control flow to confirm it holds together:

## Resolution flow is correct

The new helper centralizes the layout probe:

```bash
resolve_venv_python() {
  local candidate
  for candidate in "$VENV_DIR/bin/python" "$VENV_DIR/Scripts/python.exe"; do
    if [[ -x "$candidate" ]]; then
      printf '%s\n' "$candidate"
      return 0
    fi
  done
  return 1
}
```

Probe order is unambiguous in practice — a POSIX venv only produces `bin/python`, a Windows venv only produces `Scripts/python.exe`, so there's no real both-exist case. Git Bash/MSYS reports `.exe` files as `-x`, so the `[[ -x ... ]]` test works there.

The non-obvious part is the global-variable mutation, and it checks out. `create_or_rebuild_venv` re-resolves `VENV_PY` at line 130 *immediately after* `python -m venv` succeeds and before any use at line 131/137 — and since it's a plain function call (not a subshell), that assignment propagates up into `select_python`'s scope. `select_python` itself runs under `PYTHON_BIN="$(select_python)"` (line 209), so its `VENV_PY` writes are confined to that subshell, but that's fine because only the printed path flows out to `PYTHON_BIN`, which the final `exec "$PYTHON_BIN" -m pytest` consumes. So the Windows interpreter path reaches pytest correctly.

## The destructive-fs guards survived

Good that the symlink refusal (line 115) and the `rm -rf "$VENV_DIR"` + `return 2` failure paths are untouched. The new negative test `test_local_test_runner_rejects_venv_without_accepted_python_path` exercises exactly that: a `.venv` whose `python -m venv` produced neither layout returns 2, removes `.venv`, and never reaches pytest. That's the right invariant to lock.

## Test coverage is genuinely end-to-end

`test_local_test_runner_accepts_windows_layout_venv_from_base_python` is the standout — it stubs a base python that materializes `.venv/Scripts/python.exe`, then asserts the proof file's first line ends with `/.venv/Scripts/python.exe` and that pytest got the forwarded args. That covers the actual regression, not just a string-grep of the script. Nice.

## One minor cleanup (optional, not blocking)

The top-level `VENV_PY="$(resolve_venv_python || true)"` at line 19 is now effectively dead. Every consumer re-resolves before use (line 130 in `create_or_rebuild_venv`, line 173 in `select_python`), so nothing reads the line-19 value before it's overwritten. On master the equivalent line 6 mattered because `VENV_PY` was a static path set once; here it's recomputed at each point of need. You could drop line 19 for clarity, or keep it as a harmless safety default — it costs one extra probe on a fresh checkout where nothing exists yet. Either is fine; just flagging that it no longer does work.

Overall LGTM. The fix is minimal, the platform handling is correct, and the tests actually drive the runner end-to-end rather than asserting on its source text alone.

nesquena-hermes:
Shipped in **v0.51.736** (via #5187). Thanks @rodboev — `scripts/test.sh` now resolves both the POSIX `.venv/bin/python` and Windows `.venv/Scripts/python.exe` layouts, so Windows Git Bash contributors can run the documented test flow. Dev-tooling only, byte-identical on POSIX/CI. Closes #5152.

Issue #4755, opened 2026-06-23
New session/tab doesn't land on the profile's default workspace; mid-turn workspace switch is blocked


nesquena:
## Context — surfaced by the desktop app
Reported via the desktop app (**hermes-webui/hermes-desktop-rust#52**) but this is WebUI app logic, not desktop-specific — a desktop "new tab" just opens the server root URL with no extra params, identical to opening the app in a new browser tab. Should reproduce in a plain browser.

## Symptom
1. Opening a fresh session/tab does not auto-select the profile's **default workspace** — the user lands with no workspace selected and has to pick one.
2. The workspace **cannot be changed mid-turn** (while a run is streaming).

## Root cause (verified in this repo)
- The profile's default workspace is known client-side as `S._profileDefaultWorkspace`, set from `/api/settings`' `default_workspace` at boot/profile-switch (`static/boot.js:2087`; consumed in `static/panels.js:4882-4883`, `:5495`), but a fresh session doesn't appear to apply it as the active workspace on load.
- The mid-turn lock is explicit in `switchToWorkspace()` (`static/panels.js:5527`): it bails with a toast when `S.busy` (a turn is in flight).

## Proposed change
1. On a fresh session/tab load where the profile has a `default_workspace` and no workspace is otherwise selected, auto-select that default workspace (please confirm this is the intended UX).
2. Decide the intended mid-turn behavior: allow switching during a streaming turn, queue the switch to apply when the turn ends, or keep the hard block but make its messaging clearly intentional.

## Acceptance criteria
- A new session under a profile that has a default workspace lands on that workspace without manual selection.
- Mid-turn workspace switching behaves per the decided design, consistently in browser and desktop.

## Note
Please confirm the browser repro (the desktop report didn't include a browser comparison); the code paths above are platform-agnostic, so it should reproduce identically in a normal browser.

---
_Filed from the desktop app (hermes-webui/hermes-desktop-rust) to track the WebUI-side work. Edit / re-route / close if mis-scoped._

PR #5170, opened 2026-06-29
fix(ux): honor busy input mode preference on first send after load (#5167)

### LINKED ISSUE #5167: bug(ux): Busy input mode preference ignored on first use after load (defaults to queue until re-saved)
## Summary

The `Busy input mode` preference (Settings → Preferences, `queue` / `interrupt` / `steer`) is **not respected on first use after a page load** — sends fall back to `queue` until the user re-saves the setting. Reported by **@b3nw**: *"the first session it doesn't respect my preferences until I re-save them."*

## Root cause (client-side boot race)

Server-side is fine: `busy_input_mode` is persisted and returned by `/api/settings` (`api/config.py:7801` default `queue`, enum-validated at `:7964`).

The bug is in the front-end init order. `window._busyInputMode` has **no eager default** anywhere. It is only ever assigned in three places:

- `static/boot.js:2411` — `window._busyInputMode=(s.busy_input_mode||'queue')`, inside the async boot IIFE (`boot.js:2362`) that does `await api('/api/settings')`
- `static/boot.js:2532` — catch-path fallback to `'queue'`
- `static/panels.js:9979` — when settings are saved/applied

The send path reads it defensively:

- `static/messages.js:1280` — `const busyMode=window._busyInputMode||'queue';`
- `static/ui.js:5881` — `const busyMode=window._busyInputMode||'queue';`

So between page load and the `await /api/settings` resolving, `window._busyInputMode` is `undefined` → every send in that window silently uses **queue**, ignoring a saved `steer`/`interrupt` preference. Saving from the panel sets the global synchronously (`panels.js:9979`), which is why "re-save fixes it."

This is amplified on slower environments. The reporter is on WSL2, where `/api/settings` is already under contention (see #5132) — widening the pre-boot window and making the first-message-uses-queue race much easier to hit.

## Repro (to confirm)

1. Set `Busy input mode` to `steer` (or `interrupt`), confirm it persists.
2. Hard-reload the page (clear cache / throttle network to widen the gap).
3. Immediately, before the UI fully settles, start a run and send a follow-up message while it's active.
4. Observe the follow-up is **queued** instead of steering/interrupting.
5. Open Settings and re-save (no change needed) → subsequent sends now honor the preference.

Network throttling (DevTools → Slow 3G) on `/api/settings` makes the window trivially reproducible.

## Suggested fix direction (for the PR agent — not building here)

Give `window._busyInputMode` a deterministic value before the async settings fetch, instead of leaving it `undefined`:

- Initialize `window._busyInputMode='queue'` eagerly at script top (mirrors the other eager boot defaults), **and/or**
- Gate the busy-send branch until boot settings have resolved, **and/or**
- Read the persisted value from a synchronous source (e.g. a bootstrapped `_bootSettings`/localStorage mirror) the same way `send_key`, `hermes-lang`, etc. are handled.

The cleanest is likely an eager default + ensuring the settings-load assignment is the single source of truth once resolved.

## Notes

- Tooltip/discoverability asks from the same thread are already tracked: #5144 (busy-mode placeholder hint), #5148 / #5149 (settings search surfacing field values), #1804 (surface slash commands as actions). This issue is **only** the preference-not-applied-on-first-use bug.
- Reporter: @b3nw. Discord thread context: power-user busy-input-mode discussion (b3nw + rodboev + maintainer).

Labels: `bug`

nesquena-hermes:
## Summary

Fixes #5167 — the **Busy input mode** preference (`queue` / `interrupt` / `steer`) was ignored for any send that happened before the async boot IIFE resolved `await api('/api/settings')`. Sends in that window silently fell back to `queue`, ignoring a saved `steer`/`interrupt` preference until the user re-saved the setting.

Reported by **@b3nw**.

## Root cause (client-side boot race)

`window._busyInputMode` had **no eager default**. It was only assigned:

- `static/boot.js` (success path, inside the async boot IIFE that awaits `/api/settings`)
- `static/boot.js` (catch fallback → `'queue'`)
- `static/panels.js` (on settings save)

Both send-path read sites use `const busyMode=window._busyInputMode||'queue'` (`static/messages.js`, `static/ui.js`). So between page load and the settings fetch resolving, `window._busyInputMode` was `undefined` → every send defaulted to `queue`. Saving from the panel set the global synchronously, which is exactly why "re-save fixes it". The window is widened on slow/contended environments like WSL2 (#5132).

## Fix

Give `window._busyInputMode` a **deterministic value at `boot.js` script top**, read synchronously from a localStorage mirror (`hermes-busy-input-mode`) — mirroring how `hermes-lang` / `hermes-theme` bootstrap from a synchronous source:

- New module-scope helpers in `boot.js`: `_normalizeBusyInputMode` (validates against `{queue, interrupt, steer}`), `_persistBusyInputMode` (writes the mirror, returns the normalized value), `_readPersistedBusyInputMode` (reads + normalizes). Both localStorage accesses are `try/catch`-guarded. The persist/read helpers are exposed on `window`.
- **Eager default** `window._busyInputMode=_readPersistedBusyInputMode()` set at top level, *before* the async `/api/settings` fetch — so the very first send after a reload honors the persisted choice.
- The resolved `/api/settings` value now routes through `_persistBusyInputMode(...)`, keeping the mirror in sync and remaining the single source of truth once it resolves.
- The **catch path** now keeps the persisted preference (`_readPersistedBusyInputMode()`) instead of clobbering it to `queue` when the server is unreachable.
- The **panels.js save handler** persists via `_persistBusyInputMode(...)` (typeof-guarded for load-order safety) so the next reload's eager default honors the save.

The defensive `||'queue'` read sites in `messages.js` / `ui.js` are unchanged (belt + braces).

## Tests

- New `tests/test_5167_busy_input_mode_boot.py` (14 tests): pins the eager default is set **before** the async settings fetch, the sync-mirror helpers + guards, every assignment site keeps the mirror in sync, and the read sites remain intact.
- Updated `tests/test_1062_busy_input_modes.py` boot/panels wiring assertions to match the new persist-through-mirror wiring.
- `node --check static/boot.js` and `node --check static/panels.js`: pass.
- `python3 scripts/scope_undef_gate.py .`: **CLEAN** (17 static files, no undefined references).
- `./scripts/test.sh tests/test_5167_busy_input_mode_boot.py tests/test_1062_busy_input_modes.py`: **42 passed**.

## Manual verification

1. Set Busy input mode to `steer`, save, hard-reload with `/api/settings` throttled (DevTools → Slow 3G).
2. Start a run and send a follow-up before the UI settles → the follow-up now **steers** instead of queueing.

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -940,6 +940,35 @@ function _micToastKeyForRecognitionError(error){
 window._micActive=window._micActive||false;
 window._micPendingSend=window._micPendingSend||false;
 
+// ── Busy input mode eager default (#5167) ───────────────────────────────────
+// The Busy input mode preference (queue/interrupt/steer) is read on the send
+// path via `window._busyInputMode||'queue'`. The authoritative value only
+// arrives once the async boot IIFE below resolves the `/api/settings` fetch.
+// Without an eager value, every send during that boot window silently falls
+// back to 'queue', ignoring a saved 'steer'/'interrupt' preference (worse on
+// slow/contended environments like WSL2, see #5132). Mirror the resolved value
+// into localStorage — the same synchronous-source pattern used by hermes-lang /
+// hermes-theme — so the very first send after a reload honors the saved choice.
+const _BUSY_INPUT_MODES=['queue','interrupt','steer'];
+function _normalizeBusyInputMode(mode){
+  return _BUSY_INPUT_MODES.includes(mode)?mode:'queue';
+}
+function _persistBusyInputMode(mode){
+  const m=_normalizeBusyInputMode(mode);
+  try{localStorage.setItem('hermes-busy-input-mode',m);}catch(_){}
+  return m;
+}
+function _readPersistedBusyInputMode(){
+  let stored=null;
+  try{stored=localStorage.getItem('hermes-busy-input-mode');}catch(_){}
+  return _normalizeBusyInputMode(stored);
+}
+window._persistBusyInputMode=_persistBusyInputMode;
+window._readPersistedBusyInputMode=_readPersistedBusyInputMode;
+// Eager default set BEFORE the async settings fetch resolves so first sends in
+// the boot window honor the persisted preference instead of defaulting to queue.
+window._busyInputMode=_readPersistedBusyInputMode();
+
 // ── Extension TTS-engine registry (registerHermesTtsEngine) ──────────────────
 // Defined at MODULE scope (not inside the voice-mode IIFE below) so the public
 // API exists even on browsers without SpeechRecognition / speechSynthesis — an
@@ -2528,7 +2557,7 @@ window._applyTitlebarProfileVisibility=_applyTitlebarProfileVisibility;
       stringChars:parseInt(s.inflight_state_max_string_chars||60000,10)||60000,
       jsonChars:parseInt(s.inflight_state_max_json_chars||1500000,10)||1500000,
     };
-    window._busyInputMode=(s.busy_input_mode||'queue');
+    window._busyInputMode=_persistBusyInputMode(s.busy_input_mode);
     window._showBusyPlaceholderHint=!!s.show_busy_placeholder_hint;
     window._sessionEndlessScrollEnabled=!!s.session_endless_scroll;
     window._autoScrollFollow=s.auto_scroll_follow!==false;
@@ -2652,7 +2681,12 @@ window._applyTitlebarProfileVisibility=_applyTitlebarProfileVisibility;
     window._structuredCodeAutoTreeLines=10;
     window._sidebarDensity='compact';
     window._pinnedSessionsLimit=3;
-    window._busyInputMode='queue';
+    // Settings load failed: keep the persisted busy-input-mode preference (the
+    // eager default already read it from the localStorage mirror) instead of
+    // clobbering it with 'queue', so a saved 'steer'/'interrupt' still applies
+    // when the server is unreachable (#5167). The placeholder-hint has no
+    // persisted mirror, so it defaults off on failure.
+    window._busyInputMode=_readPersistedBusyInputMode();
     window._showBusyPlaceholderHint=false;
     window._sessionEndlessScrollEnabled=false;
     window._autoScrollFollow=true;
--- a/static/panels.js
+++ b/static/panels.js
@@ -7796,7 +7796,7 @@ async function _autosavePreferencesSettings(payload){
       if(typeof applyConversationOutlinePreference==='function') applyConversationOutlinePreference();
     }
     if(payload&&payload.busy_input_mode!==undefined){
-      window._busyInputMode=(saved&&saved.busy_input_mode)||'queue';
+      window._busyInputMode=(typeof _persistBusyInputMode==='function')?_persistBusyInputMode(saved&&saved.busy_input_mode):((saved&&saved.busy_input_mode)||'queue');
       if(typeof _applyBusyComposerPlaceholder==='function') _applyBusyComposerPlaceholder();
     }
     if(payload&&payload.show_busy_placeholder_hint!==undefined){
@@ -8382,7 +8382,7 @@ async function loadSettingsPanel(){
     if(busyInputModeSel){
       const val=String(settings.busy_input_mode||'queue');
       busyInputModeSel.value=['queue','interrupt','steer'].includes(val)?val:'queue';
-      window._busyInputMode=busyInputModeSel.value;
+      window._busyInputMode=(typeof _persistBusyInputMode==='function')?_persistBusyInputMode(busyInputModeSel.value):busyInputModeSel.value;
       busyInputModeSel.addEventListener('change',_schedulePreferencesAutosave,{once:false});
     }
     const showBusyPlaceholderHintCb=$('settingsShowBusyPlaceholderHint');
--- a/static/panels.js
+++ b/static/panels.js
@@ -10417,7 +10417,9 @@ function _applySavedSettingsUi(saved, body, opts){
   window._sessionJumpButtonsEnabled=!!body.session_jump_buttons;
   if(typeof _applySessionNavigationPrefs==='function') _applySessionNavigationPrefs();
   window._sidebarDensity=sidebarDensity==='detailed'?'detailed':'compact';
-  window._busyInputMode=body.busy_input_mode||'queue';
+  window._busyInputMode=(typeof _persistBusyInputMode==='function')
+    ? _persistBusyInputMode(body.busy_input_mode)
+    : (body.busy_input_mode||'queue');
   window._sessionEndlessScrollEnabled=!!body.session_endless_scroll;
   window._autoScrollFollow=body.auto_scroll_follow!==false;
   window._largeTextPasteAsAttachment=body.large_text_paste_as_attachment!==false;
--- a/tests/test_1062_busy_input_modes.py
+++ b/tests/test_1062_busy_input_modes.py
@@ -384,19 +384,30 @@ class TestSendBusyBranchDispatch:
 
 class TestBootAndPanelsWiring:
     def test_boot_init_default_path(self):
-        """Boot success path initialises window._busyInputMode from settings."""
-        assert "window._busyInputMode=(s.busy_input_mode||'queue')" in BOOT_JS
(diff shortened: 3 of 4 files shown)


nesquena (approved):
## Review — end-to-end ✅ (approved after one CI-red fix I pushed)

Independent review of the #5167 fix: the **Busy input mode** preference (queue / interrupt / steer) was ignored on the first send after a page load — every send during the boot window silently fell back to `'queue'` until the async `/api/settings` fetch resolved (or until the setting was re-saved). Worse on slow/contended environments (WSL2). The fix mirrors the resolved value into `localStorage` and reads it back **eagerly at module scope**, before the async settings fetch, so the very first send honors the saved choice.

### What this ships

`static/boot.js` (+~33), `static/panels.js` (+~3/-1), `tests/test_5167_busy_input_mode_boot.py` (+136, new), `tests/test_1062_busy_input_modes.py` (+~17). Agent-authored (`nesquena-hermes`), flagged for review. Pure client-side preference plumbing — no backend, no API change.

`busy_input_mode` exists in **both** repos but as independent config domains, so there is no cross-tool coupling:
- **Agent CLI** (`cli.py:3491`): `CLI_CONFIG["display"]["busy_input_mode"]` (default `interrupt`) drives the terminal's own interrupt/queue/steer input handling — its own subtree, its own default.
- **WebUI** (`api/config.py:7801`): a separate webui setting (default `queue`), validated against `{queue, interrupt, steer}` (`api/config.py:7964`), consumed only by the browser send path (`messages.js:1280`, `ui.js:5881`).

This PR touches **only** the webui client-side `window._busyInputMode` plumbing (a `localStorage` mirror). No `config.yaml` write, no CLI surface, no protocol change. The differing defaults (CLI `interrupt` vs webui `queue`) confirm the domains are independent. ✓

### End-to-end trace

1. **Eager default (module scope, before async boot IIFE)** — `boot.js` defines `_normalizeBusyInputMode` / `_persistBusyInputMode` / `_readPersistedBusyInputMode`, exposes the latter two on `window`, and sets `window._busyInputMode=_readPersistedBusyInputMode()` **synchronously**. So sends in the boot window read the persisted value instead of `'queue'`.
2. **Settings-success path** — `window._busyInputMode=_persistBusyInputMode(s.busy_input_mode)`: resolves the authoritative server value **and** refreshes the `localStorage` mirror so the next reload's eager default is correct.
3. **Settings-failure path** — `window._busyInputMode=_readPersistedBusyInputMode()` (was a hard `'queue'`): when the server is unreachable, keeps the saved preference rather than clobbering it.
4. **Live save** — `panels.js` `_applySavedSettingsUi` mirrors via `_persistBusyInputMode(body.busy_input_mode)` (guarded by `typeof …==='function'`), so saving a new preference updates the mirror immediately.
5. **Consumers** — `messages.js:1280` / `ui.js:5881` read `window._busyInputMode||'queue'`, now populated early. ✓

### What I caught — CI red on shard 0 (false positive from a brittle static guard)

Shard 0 failed across 3.11/3.12/3.13: `test_workspace_blank_page_fix.py::test_boot_sets_profile_default_workspace_in_settings_block` (a pre-existing #4755 guard). It greps the **first** occurrence of `await api('/api/settings')` and asserts the `S._profileDefaultWorkspace=s.default_workspace` assignment sits within 1000 chars of it. The PR's new eager-default **comment** near the top of `boot.js` (~char 35k) embedded that exact call literal, shadowing the real fetch (~char 100k) → proximity check failed (`65020 < 1000`). A false positive: the actual assignment never moved.

### What I pushed — `7b4359ea`

`fix(review): reword #5167 boot comment to not trip the #4755 static guard` — reworded the comment to reference the `` `/api/settings` `` fetch **without** the `await api('...')` call literal. No functional change; the real fetch is again the first match. The affected guard plus both busy-mode suites are green (53/53), and the full suite passes.

> Non-blocking follow-up: the #4755 guard is brittle (a code-shaped substring in any comment can shadow the real call). A more robust form would anchor to the assignment and search its nearest *preceding* fetch (`src.rfind("await api('/api/settings')", 0, ws_assign_idx)`). Out of scope for this PR; flagging for a future test-hardening pass.

- **Input normalization hardens behavior:** `_normalizeBusyInputMode` whitelists `['queue','interrupt','steer']`, defaulting anything else to `'queue'` — so a corrupted `localStorage` value or an unexpected `s.busy_input_mode` is sanitized (a small improvement over the old bare `||'queue'`). The JS whitelist matches the server validation set (`api/config.py:7964`). ✓
- **No XSS / injection:** the value is used purely as a mode selector (string compare on the send path), never interpolated into HTML/DOM/selectors. ✓
- **Storage safety:** every `localStorage` read/write is wrapped in `try/catch` → degrades to `'queue'` under private-mode / quota / disabled storage. ✓
- No network, no `eval`, no agent/protocol surface. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Saved `steer`, reload, send before settings resolve | honors `steer` (eager default) ✅ |
| No saved value (fresh profile) | `'queue'` ✅ |
| Corrupted `localStorage` value | normalized → `'queue'` ✅ |
| `/api/settings` fetch fails | keeps persisted preference (no clobber) ✅ |
| Save new mode in Settings | mirror updated immediately via panels.js ✅ |
| `localStorage` disabled / private mode | try/catch → `'queue'`, no throw ✅ |
| CLI reads its own `display.busy_input_mode` | unaffected (separate domain) ✅ |

### Tests

- `tests/test_5167_busy_input_mode_boot.py` (new) + `tests/test_1062_busy_input_modes.py` (extended) — **42/42**.
- Affected static guard `test_workspace_blank_page_fix.py` — **11/11** after the fix.
- **Full suite: 10907 passed / 143 skipped / 0 failed** in 239s (deselected the known darwin CRLF flake).

- CHANGELOG `[Unreleased]` entry follows AGENTS.md convention (release agent promotes it) — keep it.
- The #4755 guard brittleness (above) is worth a follow-up test-hardening pass.

### Recommendation

A clean, well-isolated UX fix that follows the established synchronous-source pattern (same as `hermes-lang` / `hermes-theme`): eager `localStorage` default before the async boot, mirror-on-resolve, and no-clobber on settings failure — with input normalization that also hardens against bad stored values. The only blocker was a false-positive CI failure from a brittle pre-existing static guard tripping over the PR's new comment, fixed in `7b4359ea`. No cross-tool concern (webui's `busy_input_mode` is independent of the CLI's). Ship.

nesquena-hermes:
## Gate certification — GREEN ✅ · decision-ready (self-built → your merge trigger)

Picked this up un-gated in an autonomous sweep, rebased onto current master (was CONFLICTING; head `adb528cf`), and ran the full authoritative gate. This is the fix for #5167 (busy-input-mode preference ignored on the first send before boot settings resolve — reported by @b3nw).

**Rebase:** union-resolved 2 boot.js conflicts (busy-mode apply + settings-load-failure path) — kept the persisted-preference-on-failure fix while preserving master's `_showBusyPlaceholderHint` lines.

**Gate found + I fixed a real gap in the original fix:** Codex caught that the `hermes-busy-input-mode` localStorage mirror was only written on boot-settings-apply — **not** when the user changes the preference via the Settings panel (autosave `panels.js:~7799` or panel-load `~8385`). So a `Queue → Steer` change wouldn't survive a reload + early busy send (the exact #5167 bug, via a path the PR missed). I routed both assignments through `_persistBusyInputMode`; re-gate confirmed the mirror now stays synchronized on **every** assignment path (boot-apply, settings-load-fail, autosave, panel-load).

**Gate result:**
- **Codex regression gate (re-gate): SAFE TO SHIP** — no regression; mirror synchronized on all paths.
- **Full pytest suite: 11351 passed** (only the 2 pre-existing cron-isolation serial artifacts).
- 42/42 busy-input-mode tests (`test_1062_busy_input_modes.py` + `test_5167_busy_input_mode_boot.py`); `node --check` clean; non-visible (boot.js/panels.js logic only).

**Disposition:** self-built (nesquena-hermes) + fix-class, non-visible. Per our self-built rule I'm not self-merging — it's decision-ready + gate-GREEN for your merge trigger (or an independent review from the nesquena account). Nothing else holding it.

nesquena-hermes:
Shipped in **v0.51.791**. Your Busy input mode preference (queue/interrupt/steer) is now honored on the very first send after load — no more silent fallback to queue before settings resolve. Thanks @b3nw for the report! Closes #5167.

PR #5178, opened 2026-06-29
test: de-flake test_skills_stats_cache mtime race

nesquena-hermes:
## Problem

`tests/test_profile_skills_stats.py::test_skills_stats_cache` is an **order-dependent flake**: it passes in isolation but fails intermittently under full-suite ordering. It has been observed failing across multiple unrelated PRs' gate runs (it adds noise to every full-suite run), always passing on an isolated re-run.

## Root cause

The test verifies the #4783 behavior that adding a skill bumps the skills-dir mtime, so the cheap per-call mtime probe in `_get_profile_skills_stats` recomputes counts on the next call instead of serving the cached value:

```python
_write_skill(tmp_path, "alpha")
profiles._get_profile_skills_stats(tmp_path)   # caches (enabled=1, probe=mtime_A)
_write_skill(tmp_path, "beta")
enabled, compat = profiles._get_profile_skills_stats(tmp_path)
assert enabled == 2 and compat == 2            # ← flakes here
```

The probe (`_skill_tree_max_mtime_ns`) keys the cache on the max `st_mtime_ns` across the skills dir. When the two `_write_skill()` calls land inside the **same coarse filesystem mtime tick**, the skills-dir mtime after writing `beta` can be byte-identical to the value cached after `alpha`. The probe then sees "no change", serves the stale count of `1`, and `assert enabled == 2` fails. Under full-suite timing the window shifts, which is why it's intermittent rather than deterministic.

## Fix

After writing the second skill, force the skills-dir mtime strictly forward so the probe is guaranteed to detect a genuine mtime change:

```python
_write_skill(tmp_path, "beta")
import os as _os, time as _time
_skills_dir = tmp_path / "skills"
_future = _time.time() + 5
_os.utime(_skills_dir, (_future, _future))
enabled, compat = profiles._get_profile_skills_stats(tmp_path)
assert enabled == 2 and compat == 2
```

This mirrors what a real FS-backed skill-add does (advance the dir mtime) and keeps the assertion honest — it still proves recompute-on-mtime-change. Only the race against filesystem mtime granularity is removed. No production code changes.

## Verification

- `test_skills_stats_cache` run 8× in isolation, the full file 5× on a clean master base: **all green, deterministic.**
- Test-only change; no behavior change to `api/profiles.py`.

Found while running the full suite as part of an independent deep-gate pass; filing the de-flake separately so it lands cleanly on its own.

### DIFF
--- a/tests/test_profile_skills_stats.py
+++ b/tests/test_profile_skills_stats.py
@@ -144,6 +144,17 @@ def test_skills_stats_cache(tmp_path):
     # Adding a skill bumps the skills-dir mtime; the cheap probe detects it on
     # the very next call and the counts update immediately (no stale TTL window).
     _write_skill(tmp_path, "beta")
+    # Guarantee the skills-dir mtime is STRICTLY newer than the cached probe value.
+    # A real FS-backed skill-add always advances the dir mtime, but two writes inside
+    # the same coarse mtime tick can leave it byte-identical under full-suite timing —
+    # the source of this test's intermittent flake (order-dependent: passes in
+    # isolation, fails ~intermittently under full-suite ordering). Forcing it forward
+    # keeps the assertion honest (the probe must recompute on a genuine mtime change)
+    # without racing the filesystem's mtime granularity.
+    import os as _os, time as _time
+    _skills_dir = tmp_path / "skills"
+    _future = _time.time() + 5
+    _os.utime(_skills_dir, (_future, _future))
     enabled, compat = profiles._get_profile_skills_stats(tmp_path)
     assert enabled == 2 and compat == 2
 


nesquena (approved):

Independent review of a **test-only** de-flake: `tests/test_profile_skills_stats.py::test_skills_stats_cache` is an order-dependent flake that passes in isolation but intermittently fails under full-suite ordering, adding noise to unrelated PRs' gate runs.

### What this ships

`tests/test_profile_skills_stats.py` (+11/-0). Agent-authored (`nesquena-hermes`). No production code change.

### Root cause — verified against the probe

The test exercises the #4783 behavior: adding a skill bumps the skills-dir mtime so the cheap per-call probe `_get_profile_skills_stats` recomputes instead of serving the cached count. The probe is `_skill_tree_max_mtime_ns` (`api/profiles.py:1588`), which takes the **max `st_mtime_ns`** across `config.yaml`, the skills dir itself (`:1621`), each skill subdir (`:1626`), and each `SKILL.md` (`:1631`); `_get_profile_skills_stats` keys its cache on that value (`:1711` vs `:1731`).

When the two `_write_skill()` calls land inside the **same coarse FS mtime tick**, the post-`beta` max can be byte-identical to the value cached after `alpha`. The probe sees "no change", serves the stale count `1`, and `assert enabled == 2` fails. Full-suite timing shifts the window → intermittent, not deterministic. The diagnosis in the PR body matches the code exactly. ✓

### The fix — honest and minimal

After writing `beta`, force the skills-dir mtime strictly forward (`os.utime(skills_dir, (now+5, now+5))`). Since the probe maxes over the skills dir, the new max is guaranteed `> ` the cached value, so the recompute path fires deterministically. This **mirrors what a real FS-backed skill-add does** (advance the dir mtime) — the assertion still proves recompute-on-mtime-change; only the race against FS mtime granularity is removed. It does not weaken what the test guards (it still fails if the probe stopped recomputing on a genuine mtime bump). ✓

### Audit

- **Test-only, additive within one test** — cannot affect any other test's behavior or ordering. ✓
- The forced future mtime (`+5s`) is local to `tmp_path` and discarded with the fixture. No global state. ✓
- No production code touched (`api/profiles.py` unchanged). ✓

### Tests

- Locally the file **skips** (9 skipped: `hermes-agent Python modules not importable (cron, skills_tool)` — the standard agent-dependent guard via `requires_agent_modules`), so I verified the fix by tracing the probe rather than executing it here.
- **CI is fully green** (lint + browser-smoke + all 9 test shards across 3.11/3.12/3.13), where the agent modules are present and the test actually runs.
- PR author reports `test_skills_stats_cache` green 8× in isolation and the full file 5× on a clean master base — consistent with the deterministic-forward-mtime fix.

### Recommendation

A correct, minimal, test-only de-flake. The diagnosis matches the probe code (`_skill_tree_max_mtime_ns` max-mtime keying), and forcing the skills-dir mtime strictly forward is the right fix — it removes the FS-granularity race while keeping the recompute-on-change assertion honest. No production impact; CI green. Ship.

nesquena-hermes:
Shipped in **v0.51.742** (via #5194). Test-only de-flake —  now forces the skills-dir mtime strictly forward before the recompute assertion, eliminating the same-mtime-tick race that made it fail intermittently under full-suite ordering (and added noise to unrelated PRs' gate runs). Assertion unchanged.

Issue #3280, opened 2026-05-31
File manager returns 404 when external Telegram/CLI session is active


collex100:
When an external session created via Telegram or CLI is selected as active in the web UI, the Files tab returns HTTP 404 for all file operations. The /api/file handler calls get_session which only searches webui/sessions/ directory. External sessions stored in state.db are not found, causing KeyError that surfaces as 404. The workspace path is the same for all session types so the file browser should not depend on session storage location. Suggested fix: add fallback in file route handlers to look up session in state.db when get_session raises KeyError. Image: ghcr.io/nesquena/hermes-webui:latest 2026-05-31.

collex100:
## Suggested Patch

Tested and verified on production.

[webui-external-sessions.patch](https://github.com/user-attachments/files/28443320/webui-external-sessions.patch)

nesquena-hermes:
Thanks for the precise report — the root-cause description matches what I see in the code. I read the `/api/file*` handlers in `api/routes.py` plus the session loaders in `api/models.py` to confirm.

## Where the 404 comes from

All the file-manager routes resolve the session via `get_session(sid)` and hard-404 on `KeyError`. `_handle_file_read` (`api/routes.py:8595`) and `_handle_file_raw` (`:8557`) both do:

```python
try:
    s = get_session(sid)
except KeyError:
    return bad(handler, "Session not found", 404)
```

`_handle_folder_download` (`:8480`) has the same guard. And `get_session` (`api/models.py:2195`) delegates to `Session.load`, which only reads the WebUI sidecar:

```python
def load(cls, sid):
    ...
    p = SESSION_DIR / f'{sid}.json'
    if not p.exists():
        return None      # -> get_session() raises KeyError(sid) at the end
```

So any session that exists only in `state.db` (Telegram/CLI/cron) has no `webui/sessions/<sid>.json`, `Session.load` returns `None`, `get_session` raises `KeyError`, and every file op 404s. Your diagnosis is correct.

## Important nuance for the fix

The handlers only use the session object for **one thing**: `s.workspace` (e.g. `safe_resolve(Path(s.workspace), rel)` in `_file_raw_target` at `:8388`). They don't need messages or any sidecar-specific state.

And here's the key detail — external sessions don't carry their own cwd anyway. The CLI/external loader at `api/models.py:3424` already fills the sidebar entry with the WebUI's last workspace, not a per-session path:

```python
cli_sessions.append({
    'session_id': sid,
    'title': _display_title,
[...]

PR #5179, opened 2026-06-29
[security] fix(file-manager): enforce active profile for session workspaces

Hinotoi-agent:
## Thinking Path

- Hermes WebUI profiles are used to separate sessions, workspaces, and profile-owned state.
- Session export already treats a session whose stored `profile` does not match the active profile as not found.
- The file manager used the same caller-controlled `session_id` to resolve a workspace, but it did not apply the active-profile check before serving files.
- This PR puts the profile check into the shared file-session resolver so read/download/mutation file routes inherit one boundary.

## Summary

This PR hardens the file-manager session boundary so WebUI sessions must belong to the active profile before their workspace can be browsed or mutated.

- Rejects a WebUI session from another profile before returning it to file-manager handlers.
- Keeps the existing state.db / external-session fallback for sessions that do not have a WebUI sidecar.
- Adds regression coverage for both the helper-level authorization failure and a route-level `/api/file` denial.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Cross-profile file-manager access by `session_id` | A request under one active profile could use a known session id from another profile to read or modify files in that session's workspace. | Moderate |

## Before this PR

- `get_session_for_file_ops()` returned a WebUI session from `get_session(sid, metadata_only=True)` without checking its stored `profile`.
- File-manager routes then used `Path(s.workspace)` for read/download/mutation operations.
- Profile-aware routes such as session export already rejected the same cross-profile session id, so the boundary was enforced inconsistently.

## After this PR

- WebUI sessions returned to file-manager handlers must satisfy `_profiles_match(session.profile, get_active_profile_name())`.
- A found-but-foreign WebUI sidecar is treated as not found and does not fall through to state.db fallback.
- Existing state.db-only external session behavior is preserved for sessions that do not have a WebUI sidecar.
- Tests cover default-profile access to a named-profile workspace returning 404 without leaking marker content.

## Why It Matters

Hermes WebUI workspaces can contain user files, generated artifacts, and project context. Profile isolation should apply to direct file APIs just as it does to session export and profile-scoped session lists. Without this check, a stale or known session id can cross from one active profile into another profile's workspace file surface.

## Attack flow

```text
Known or retained named-profile session_id
    -> file-manager route such as GET /api/file?session_id=...&path=...
        -> get_session_for_file_ops returns the foreign WebUI session
            -> route reads Path(session.workspace) without a profile check
                -> cross-profile workspace file disclosure or mutation
```

## Affected code

| Issue | Files |
|-------|-------|
| Cross-profile file-manager access | `api/models.py`, `tests/test_file_manager_external_session.py` |

## Root cause

Cross-profile file-manager access:
- Direct cause: `get_session_for_file_ops()` returned WebUI sessions by id without applying the active-profile predicate used by other session routes.
- Boundary failure: the shared file-manager resolver treated existence of a session id as sufficient authority to access that session's workspace.

## CVSS assessment

| Cross-profile file-manager access | 6.5 Medium | `CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:L/A:N` |

Rationale:
- This requires an admitted WebUI caller/profile context and knowledge of a session id, but it can expose files from another profile's workspace and may reach file mutation routes that use the same resolver.

## Safe reproduction steps

1. Create or simulate a WebUI session whose stored `profile` is a named profile such as `research`, with a workspace file containing a harmless marker.
2. Make the active request profile `default`.
3. Request `GET /api/file?session_id=<foreign-session>&path=<marker-file>`.
4. Before this PR, the route returns `200` and includes the marker content. After this PR, it returns `404` and does not include the marker.

- A named-profile workspace marker should never be returned to a different active profile.
- The pre-patch route returned the marker through the file API even though profile matching returned false.
- The fixed route returns the same not-found behavior as the profile-aware session export path.

## Changes in this PR

- Adds active-profile validation to `get_session_for_file_ops()` for WebUI sessions.
- Treats found-but-foreign WebUI sessions as unauthorized/not found instead of falling through to external session fallback.
- Adds regression tests for foreign-profile helper rejection and route-level file-read denial.
- Keeps the existing state.db fallback tests passing for external sessions.

## Files changed

| Backend authorization | `api/models.py` | Enforces active-profile matching before returning WebUI sessions to file-manager handlers. |
| Regression tests | `tests/test_file_manager_external_session.py` | Covers same-profile passthrough, foreign-profile rejection, no fallback on authorization failure, and route-level 404 behavior. |

## Maintainer impact

- The diff is intentionally small and localized to the file-manager session resolver and its existing regression file.
- File path containment, upload handling, frontend behavior, and session export behavior are not changed.
- External state.db-only sessions still use the existing fallback path when no WebUI sidecar exists.

## Fix rationale

The file-manager resolver is the right choke point because all file-manager handlers already route through it. Enforcing the profile check there keeps sibling read/download/mutation routes aligned and avoids duplicating authorization logic in each handler.

## Type of change

- [x] Security fix
- [x] Tests
- [ ] Documentation update

## Verification

- [x] `./scripts/test.sh tests/test_file_manager_external_session.py`
- [x] `./.venv/bin/python -m py_compile api/models.py tests/test_file_manager_external_session.py`
- [x] `./.venv/bin/ruff check tests/test_file_manager_external_session.py`
- [x] `git diff --check`
- [x] Docker/container validation with `python:3.11-slim`: copied the mounted repo into `/work`, ran the focused test file, compiled touched Python files, and ran `git diff --check`.

## Risks / Follow-ups

- This intentionally returns not-found for foreign-profile WebUI sessions to match existing profile-scoped routes.
- If maintainers want a different external-session policy for state.db-backed sessions, that can be handled separately; this PR preserves the current fallback behavior when no WebUI sidecar exists.

## Disclosure notes

- Claims are bounded to the file-manager resolver and routes that depend on it.
- No production data or real secrets were used in validation.
- No unrelated files are changed.

## Model Used

AI-assisted with OpenAI GPT-5.5 through Hermes Agent. Tool use included local code inspection, focused tests, Docker validation, and GitHub CLI operations.


Hinotoi-agent:
Addressed Greptile's test isolation note in `6a33ad5c`: the WebUI passthrough test now stubs `api.profiles._profiles_match` and asserts the exact profile values passed into the helper, so it no longer relies on the production `None`-profile convention. I also applied the same explicit stub to the foreign-profile rejection test so both branches are self-contained.

Validation:
- `./scripts/test.sh tests/test_file_manager_external_session.py` — 8 passed
- `./.venv/bin/python -m py_compile api/models.py tests/test_file_manager_external_session.py` — passed
- `./.venv/bin/ruff check tests/test_file_manager_external_session.py` — passed
- `git diff --check` — passed
- Docker `python:3.11-slim` focused test + py_compile + diff-check — passed

GitHub Actions on the updated head are all passing as well.

nesquena-hermes:
Reviewed the diff at `origin/master...HEAD` for `api/models.py` plus the resolver's callers in `api/routes.py`. The fix is correctly placed and complete — here's what I verified.

## Code reference

The guard lands inside `get_session_for_file_ops()` after the successful lookup, so it only gates WebUI sidecar sessions and leaves the state.db fallback untouched (`api/models.py:3360`):

```python
try:
    session = get_session(sid, metadata_only=True)
except KeyError:
    if state_db_has_session(sid):
        return _ExternalSessionView(str(sid), str(get_last_workspace()))
    raise

from api.profiles import _profiles_match, get_active_profile_name

if not _profiles_match(getattr(session, 'profile', None), get_active_profile_name()):
    raise KeyError(sid)
return session
```

## Diagnosis

The placement is right for three reasons:

1. **404 parity holds.** Every one of the 16 file-manager call sites wraps this in `except KeyError: return bad(handler, "Session not found", 404)` (e.g. `api/routes.py:14096`, `:15636`, `:19031`). Raising `KeyError(sid)` on a profile mismatch produces the exact same 404 a missing session does, with no marker content leaked — matching the profile-aware session-export behavior the PR cites.

2. **`_profiles_match` is the canonical predicate.** At `api/profiles.py:373` it treats `None`/empty profile as `'default'` and cross-aliases a renamed root via `_is_root_profile`. Since `Session.profile` is a real attribute (set at `api/models.py:742`, serialized at `:1040`), the `getattr(..., None)` default only fires for legacy-backfilled rows, where `None → 'default'` is the correct convention.

3. **The fallback path can't be used to bypass.** A reader might worry a foreign profile could re-enter through state.db. It can't: the mismatch branch `raise`s directly rather than falling through, and even the fallback that *does* run (on the original `KeyError`) is inherently profile-safe — `state_db_has_session()` queries `_active_state_db_path()` (the *active* profile's db) and `_ExternalSessionView` is built from `get_last_workspace()` (the active workspace, per #3280), never the foreign session's stored workspace.

## Recommendation

Looks correct and minimal as-is. One optional hardening worth a follow-up (not blocking): the cross-profile rejection is silent (just a 404). A `diag_stage(...)` or debug log on the mismatch branch would make cross-profile probe attempts visible in diagnostics, mirroring how the orphan-prune path logs at `api/routes.py:1804`. Not required for correctness.

## Verification

The included `tests/test_file_manager_external_session.py` cases (same-profile passthrough, foreign-profile rejection with no state.db fallthrough, route-level 404 with no marker leak) cover the three branches above. The no-fallthrough assertion is the important one — it pins point (3). LGTM on the security boundary.

Hinotoi-agent:
Addressed the optional diagnostic recommendation in `d8ab4819`: the foreign-profile branch in `get_session_for_file_ops()` now emits a debug log before returning the existing 404/`KeyError` path, including the session id, stored session profile, and active profile. I also pinned that behavior in the existing foreign-profile regression with `caplog`, while preserving the no-state.db-fallthrough assertion.

Validation:
- `./scripts/test.sh tests/test_file_manager_external_session.py` — 8 passed
- `./.venv/bin/python -m py_compile api/models.py tests/test_file_manager_external_session.py` — passed
- `./.venv/bin/ruff check api/models.py tests/test_file_manager_external_session.py` — passed
- `git diff --check` — passed
- Docker `python:3.11-slim` focused test + py_compile + diff-check — passed
- GitHub Actions on the updated head are all passing.

nesquena-hermes:
Shipped in **v0.51.762** — thank you @Hinotoi-agent. 🔒

Merged via a maintainer rebase of your branch onto current `master` (release PR #5263) with your authorship preserved in the commits. Closing this PR manually because the release landed through the rebased branch rather than this PR's original branch.

**Full authoritative gate before ship:**
- **Codex adversarial security gate (on the rebased worktree): SAFE TO SHIP** — no regression, fails closed, foreign sidecars 404, state.db external-session fallback intact. (An earlier `SHIP ONLY WITH FIXES` was a stale-base artifact: the worktree had been cut before #5240 merged, so the two-dot diff falsely showed #5240's terminal-recovery code as "removed." Codex itself attributed it to the merge-base mismatch; re-rebasing onto current master and re-gating returned SAFE.)
- **Opus correctness/architecture: APPROVE** (no nits) — closes the cross-profile workspace leak, fails closed, preserves single-profile + external-session behavior, and verified the subtle prerequisite that `profile` survives a `metadata_only` load (so the check doesn't silently no-op for named profiles). No cron/agent caller relies on the old cross-profile behavior.
- **Full pytest suite: 11090 passed** (only the known `test_scheduled_jobs_profile_isolation` serial-ordering artifact failed — passes in isolation and sharded CI; this PR touches no cron code).
- Your `test_file_manager_external_session.py`: 8/8 pass.

Clean, well-scoped security fix — the raise-outside-the-try placement so a foreign session can't fall through to the state.db path is exactly right.

PR #5090, opened 2026-06-28
Document anchor fallback ownership

franksong2702:
## Thinking Path

#4958 shipped the mixed `content[]` bridge, so settled assistant activity now has a clear owner when `_anchor_activity_scene` is present. The next risk is fallback drift: legacy raw-message paths still need to support historical/non-anchor transcripts, but they must not compete with anchor-owned settled turns.

## What Changed

- Added a Settled Fallback Ownership Matrix to `docs/architecture/stable-assistant-turn-anchor-phase0.md`.
- Added `tests/test_anchor_fallback_ownership.py` to pin the current fallback boundary.
- Added a Node-executed behavior check for `_transparentStreamOrderedParts()`:
  - anchor-owned messages return `null` and stay owned by the anchor renderer
  - historical non-anchor mixed text/tool `content[]` still reconstructs ordered parts
  - Transparent Stream disabled still exits the fallback
- Added source-structure guards for Compact Worklog legacy tool/thinking/worklog fallback skips and the settled anchor renderers.

## Why It Matters

This is the first post-#4958 audit slice before any fallback shrink/removal. It protects both sides of the contract: no duplicate rendering for anchor-owned settled turns, and no historical transcript regression for messages that do not yet carry an anchor scene.

Refs #3400.

## Contract Routing

Task type: post-#4958 architecture/test guardrail.

Touched areas:
- `docs/architecture/stable-assistant-turn-anchor-phase0.md`
- `tests/test_anchor_fallback_ownership.py`
- `static/ui.js` ownership invariants, exercised via tests only

Docs decision: the architecture doc is the right place for this matrix. No README or public user docs update is needed because this PR changes no runtime behavior or user-facing workflow.

CHANGELOG decision: no entry. This is a non-user-visible guardrail/test slice.

Scope boundaries:
- no runtime behavior changes
- no fallback deletion
- no UI changes
- no live/replay parity claim

## Verification

- `node --check static/ui.js`
- `./scripts/test.sh tests/test_anchor_fallback_ownership.py -q`
- `./scripts/test.sh tests/test_anchor_fallback_ownership.py tests/test_issue4622_settled_toolcard_body.py::test_non_final_post_tool_text_survives_fresh_settlement tests/test_anchor_scene_persistence.py::test_anchor_scene_hydration_preserves_non_final_post_tool_text -q`
- `python3 scripts/ruff_lint.py --diff origin/master`
- `git diff --check`

## Risks / Follow-ups

[...]

PR #4958, opened 2026-06-26
fix(anchor): preserve settled mixed content order

franksong2702:
## Thinking Path

This follows the direction from the Transparent Stream ordering fix in #4932, but moves the mixed settled-turn ownership into Stable Assistant Turn Anchor instead of extending the raw `content[]` fallback.

Root cause: once a settled assistant message carries `_anchor_activity_scene`, the #4932 helper correctly exits, but the anchor settlement/hydration path did not promote mixed `content[]` text/tool ordering into scene rows. The first pass also treated post-`tool_use` text as final-answer tail without checking whether the current assistant message was the turn-final assistant, which could silently drop post-tool process text/thinking from a non-final assistant message.

The correct layer is the anchor scene model and its read-side hydration, not another renderer fallback. The fix keeps the final answer separate from activity rows, preserves non-final post-tool activity, and makes fresh settle and reload symmetric.

## Contract Routing

Task type: runtime/rendering behavior fix for Stable Assistant Turn Anchor and Transparent Stream settled history.

Touched areas:
- `static/messages.js`
- `api/routes.py`
- `docs/architecture/stable-assistant-turn-anchor-phase0.md`
- focused regression tests

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/architecture/stable-assistant-turn-anchor-phase0.md`
- `docs/rfcs/stable-assistant-turn-anchors.md`
- `docs/rfcs/live-to-final-assistant-replies.md`

Scope boundaries: this PR does not remove the raw Transparent Stream `content[]` helper. It remains a fallback for settled messages that do not carry `_anchor_activity_scene`.

Evidence needed before claiming done: settle and reload paths both promote mixed `content[]` text/tool parts into `activity_scene_v1`, keep final answer separate, preserve non-final post-tool text/thinking as activity, avoid duplicate tool rows, and preserve durable tool snippets and richer tool args.

## What Changed

- Added anchor-side parsing for settled assistant `content[]` messages that interleave text parts and `tool_use` parts.
- Treat text after the last `tool_use` as the final answer only on the turn-final assistant message.
- Preserve post-`tool_use` process text and thinking rows from non-final assistant messages as chronological activity rows.
- Mirrored the same repair in backend hydration so reload/cold-load history reconstructs the same scene model.
- Merged duplicate hydrated tool rows so session-level durable snippets and richer args can enrich the anchor-owned tool row instead of producing a second card.
[...]

PR #5183, opened 2026-06-29
Test anchor runtime journal parity

franksong2702:
## Thinking Path

- #4958 made settled mixed `content[]` turns feed the Stable Assistant Turn Anchor scene model.
- #5090 then documented and guarded which legacy raw-message paths must stay fallback-only when `_anchor_activity_scene` exists.
- The next serial #3400 gate is replay/runtime-journal parity: before shrinking any fallback, we need an invariant that the runtime-journal live anchor snapshot and the settled persisted/hydrated anchor scene preserve the same visible activity semantics.
- Current production code already has the needed paths, so this PR adds the parity test without changing runtime behavior.

## What Changed

- Added `_anchor_scene_visible_semantics()` in `tests/test_anchor_scene_persistence.py` to compare the visible semantics of anchor activity rows while ignoring unstable row ids/order metadata.
- Added `test_runtime_journal_anchor_scene_matches_settled_hydrated_visible_semantics`, covering this path:
  - `_run_journal_live_snapshot(...).anchor_activity_scene`
  - persisted `anchor_activity_scenes` record shape
  - `_hydrate_anchor_activity_scenes(...)._anchor_activity_scene`
- The test covers process prose, thinking/reasoning, a tool start + tool completion collapsed into completed tool semantics, post-tool process prose, terminal status preservation, and final-answer separation.

## Why It Matters

This gives the Anchor effort a concrete replay/runtime-journal parity guard before any future fallback shrink/removal. It proves the live recovery snapshot and settled read-side hydration agree on the user-visible activity timeline for a representative tool-bearing turn, while keeping #4932-style raw `content[]` behavior as historical/non-anchor fallback.

## Contract Routing

- Contract family: Stable Assistant Turn Anchor / Live-to-Final assistant reply lifecycle / runtime-journal replay.
- Product-semantics tests touched: `tests/test_anchor_scene_persistence.py`.
- Contract change type: test-only invariant; no runtime behavior change.
- Evidence: focused parity test compares the real `_run_journal_live_snapshot` and `_hydrate_anchor_activity_scenes` backend paths.

## Verification

- `./scripts/test.sh tests/test_anchor_scene_persistence.py::test_runtime_journal_anchor_scene_matches_settled_hydrated_visible_semantics` -> 1 passed
- `./scripts/test.sh tests/test_anchor_scene_persistence.py tests/test_live_to_final_anchor_visible_order.py tests/test_run_journal_frontend_static.py tests/test_run_journal_routes.py` -> 120 passed
- `python3 scripts/ruff_lint.py --diff origin/master` -> no changed `.py` files vs origin/master; OK
- `node --check static/ui.js`
- `git diff --check`

## Risks / Follow-ups

- This is a parity guard, not fallback removal.
- The test is representative rather than exhaustive across every anchor event kind such as compression, control, approval, and artifact rows.
- The next slice can use this invariant plus the #5090 ownership matrix to decide whether any legacy fallback can be further narrowed.

Refs #3400.

## Model Used

AI-assisted with OpenAI GPT-5 via Codex as Coordinator, with a GPT-5.5 worker for the bounded implementation slice. Local repo inspection, focused tests, and GitHub CLI read-backs were used.


nesquena-hermes:
Shipped in **v0.51.740** (via #5192). Thanks @franksong2702 — your runtime-journal ↔ settled-hydration parity regression test is in, locking the invariant that a live-streaming turn and a reloaded settled turn render identical anchor activity. Verified it drives a real RunJournalWriter + the real snapshot/hydrate path (no mocking the unit), is non-tautological, and the broader anchor/journal sweep stays green (381 passed).

PR #5184, opened 2026-06-29
fix(updates): resolve git under macOS launchd (#5175)

### LINKED ISSUE #5175: WebUI update checker reports git executable not found under macOS launchd despite git being installed
## Summary

When Hermes WebUI was running under a macOS LaunchAgent, the Settings update checker failed with `git executable not found`, even though `/usr/bin/git` existed and the process environment appeared to include a reasonable PATH.

A local workaround was to make the update checker resolve git explicitly with `shutil.which("git")`, falling back to `/usr/bin/git` on macOS.

## Environment

- macOS 26.4
- WebUI repo: `https://github.com/nesquena/hermes-webui.git`
- WebUI before update: `v0.51.346`
- WebUI after update: `v0.51.731`
- Hermes agent: `v2026.6.19-1223-g10043c6d0`
- WebUI running via macOS LaunchAgent: `com.claudeschneider.hermes-webui`

## Symptoms

Settings showed:

```text
WebUI: unknown
Agent: 0.17.0
```

Clicking check now showed:

```text
Update check failed: WebUI: git executable not found, Agent: git executable not found
```

After a local hotfix, Settings API reported:

```text
WebUI v0.51.346-dirty-eec85f0f
Agent v2026.6.19-1156-g9bb5a809b
Update check: WebUI error None, Agent error None
```

The dirty suffix was expected because the local hotfix modified `api/updates.py`. After updating WebUI to `v0.51.731`, the dirty suffix remained until the local workaround file was hidden from git status with `assume-unchanged`.

## Local workaround

In `api/updates.py`, inside the git runner path, resolve git once:

```python
shutil.which("git") or ("/usr/bin/git" if Path("/usr/bin/git").exists() else "git")
```

Then pass that resolved executable to `subprocess.run` instead of invoking bare `git`.

After restart, verified:

```text
WebUI version: v0.51.731
Agent version: v2026.6.19-1223-g10043c6d0
Update check: WebUI error None, behind 0; Agent error None, behind 0
Git status: ## master...origin/master, no modified files shown
```

## Expected behavior

The update checker should locate git reliably under launchd on macOS. If PATH lookup fails, `/usr/bin/git` should be tried before reporting `git executable not found`.

rodboev:
## Thinking Path

- The version badge and the update checker both fail on the same helper path, because `api/updates.py` launches bare `git` from `_run_git()` and launchd does not inherit the user's shell PATH.
- The narrow fix is to resolve the executable once in `_run_git()`, keep normal PATH lookup first, and use `/usr/bin/git` only for the macOS launchd miss that the issue demonstrates.
- That keeps `.git`-absent installs on their existing `no_git` path, preserves the current missing-git diagnostic when no executable exists, and fixes every git-backed caller through the single launch owner.

## What Changed

- `api/updates.py`: added a private git-executable resolver used only by `_run_git()`, keeping the existing timeout and diagnostic behavior while recovering macOS launchd installs through `/usr/bin/git`.
- `tests/test_issue5175_macos_launchd_git.py`: added focused behavioral regressions for launchd PATH loss, the macOS fallback, WebUI version recovery, update-check recovery, and the true-missing-git invariants.
- `tests/test_updates.py`: made the pre-existing `_run_git()` unit tests patch `shutil.which(...)` explicitly so they stay hermetic when git is absent from PATH.

## Why It Matters

WebUI instances launched by macOS launchd can now report their git-backed version and run update checks when the system git exists at `/usr/bin/git`. Installs that truly lack git, or lack `.git` metadata, keep their current behavior and messaging.

## Verification

```bash
pytest tests/test_issue5175_macos_launchd_git.py -v --timeout=60
pytest tests/test_version_badge.py tests/test_issue4356_no_git_update_check.py tests/test_updates.py -v --timeout=60
```

## Upstream

Closes #5175.
Reporter workaround details in #5175 informed the resolver shape.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/updates.py
+++ b/api/updates.py
@@ -15,6 +15,7 @@ import os
 import re
 import shutil
 import subprocess
+import sys
 import threading
 import time
 import urllib.error
@@ -186,9 +187,12 @@ def _run_git(args, cwd, timeout=10):
     On failure, returns stderr (or stdout as fallback) so callers can
     surface actionable git error messages instead of empty strings.
     """
+    git_executable = _resolve_git_executable()
+    if not git_executable:
+        return 'git executable not found', False
     try:
         r = subprocess.run(
-            ['git'] + args, cwd=str(cwd), capture_output=True,
+            [git_executable] + args, cwd=str(cwd), capture_output=True,
             text=True, timeout=timeout,
             encoding='utf-8', errors='replace',
         )
@@ -207,6 +211,15 @@ def _run_git(args, cwd, timeout=10):
         return 'git executable not found', False
     except OSError as exc:
         return f'git failed to start: {exc}', False
+
+
+def _resolve_git_executable():
+    git_executable = shutil.which('git')
+    if git_executable:
+        return git_executable
+    if sys.platform == 'darwin' and os.path.exists('/usr/bin/git'):
+        return '/usr/bin/git'
+    return None
 
 
 def _dirty_suffix(path: Path, timeout=1) -> str:
--- a/tests/test_issue5175_macos_launchd_git.py
+++ b/tests/test_issue5175_macos_launchd_git.py
@@ -1,0 +1,134 @@
+import sys
+from unittest.mock import MagicMock, patch
+
+import api.updates as updates
+
+
+def test_run_git_uses_which_result_when_available(tmp_path):
+    with patch.object(updates.shutil, 'which', return_value='C:/Tools/git.exe'), \
+         patch.object(updates.subprocess, 'run') as mock_run:
+        mock_run.return_value = MagicMock(returncode=0, stdout='v0.51.999\n', stderr='')
+
+        out, ok = updates._run_git(['describe', '--tags'], tmp_path)
+
+    assert ok is True
+    assert out == 'v0.51.999'
+    assert mock_run.call_args.args[0][0] == 'C:/Tools/git.exe'
+
+
+def test_run_git_falls_back_to_usr_bin_git_on_darwin(tmp_path):
+    def fake_run(cmd, **kwargs):
+        if cmd[0] == 'git':
+            raise FileNotFoundError
+        if cmd[0] != '/usr/bin/git':
+            raise AssertionError(f'unexpected executable: {cmd[0]!r}')
+        return MagicMock(returncode=0, stdout='v0.51.999\n', stderr='')
+
+    with patch.object(updates.shutil, 'which', return_value=None), \
+         patch.object(sys, 'platform', 'darwin'), \
+         patch.object(updates.os.path, 'exists', return_value=True), \
+         patch.object(updates.subprocess, 'run', side_effect=fake_run):
+        out, ok = updates._run_git(['describe', '--tags'], tmp_path)
+
+    assert ok is True
+    assert out == 'v0.51.999'
+
+
+def test_run_git_returns_not_found_when_no_executable(tmp_path):
+    calls = []
+
+    def fake_run(cmd, **kwargs):
+        calls.append(cmd[0])
+        if cmd[0] == '/usr/bin/git':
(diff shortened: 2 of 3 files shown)


rodboev:
Keeping this as-is. `_resolve_git_executable()` is private to `_run_git()`, and I want the reader to hit the public git-launch path first, then the tiny resolver immediately after it. If this module starts accumulating more executable-resolution helpers later, I’d revisit the ordering.

nesquena-hermes:
Reading `api/updates.py:184-222` at this PR's HEAD against `origin/master`, plus the caller map for `_run_git`, this fix looks correct and well-scoped for #5175. A few notes from the code, mostly confirmations plus two scope observations.

## What the fix does right

`_run_git` is the single chokepoint for the update checker — every git call in this module routes through it (`_dirty_suffix` → `_describe_git_version`, `_check_repo`'s fetch/tag/rev-list/merge-base chain, etc.). So resolving the executable once inside `_run_git` covers the entire update-check surface in one place. The fallback is conservative:

```python
def _resolve_git_executable():
    git_executable = shutil.which('git')
    if git_executable:
        return git_executable
    if sys.platform == 'darwin' and os.path.exists('/usr/bin/git'):
        return '/usr/bin/git'
    return None
```

Non-darwin and truly-git-absent cases keep returning `'git executable not found'`, so the existing `FileNotFoundError` contract is preserved. The Linux-guard test (`test_run_git_returns_not_found_when_no_executable`) correctly asserts `/usr/bin/git` is never probed off-darwin even when `os.path.exists` is forced True — that's the right invariant to lock down.

## Prior art: rollback.py already resolves git

`api/rollback.py:95-97` already carries a resolver for the same reason:

```python
def _find_git() -> str:
    """Return the path to the git binary."""
    return shutil.which("git") or "git"
```

So there are now two slightly different resolvers in the tree — `_find_git()` (returns bare `"git"` as last resort, no macOS fallback) and the new `_resolve_git_executable()` (macOS fallback, `None` sentinel). Not a blocker for this PR, but worth a future consolidation note: the new one is strictly the better behavior, and `rollback.py` would benefit from the same `/usr/bin/git` fallback since checkpoint/rollback git ops run under the same launchd environment.

## Scope note: sibling call sites still bare-`git`

The same launchd PATH-loss would hit other bare-`['git']` invocations that don't go through any resolver:

- `api/workspace_git.py:88` — `argv = ["git"]`
- `api/worktrees.py:18` — `subprocess.run(["git", *args], ...)`
- `api/workspace.py:1680` — `['git'] + args`

These are out of scope for #5175 (which is specifically the update checker), so I wouldn't expand this PR. But a follow-up that promotes `_resolve_git_executable()` to a shared helper (e.g. in a small git-util module) and routes these workspace-git paths through it would close the whole class of "git not found under launchd" rather than just the update checker. Workspace git status / staging would otherwise still silently fail under the same LaunchAgent.

## On the ordering nit

Agree with keeping `_resolve_git_executable` after `_run_git` — it's a private helper to a single public caller and Python resolves the name at call time, so the readability argument (public path first, tiny resolver right after) is reasonable. No change needed there.

## Minor, optional

`shutil.which('git')` now runs on every `_run_git` call, and `_check_repo` issues ~10 of them per check. It's negligible behind the 30-min TTL cache, but if you ever want it tighter, resolving once per `_check_repo` pass (or module-level memoization with a re-probe on miss) would avoid the repeated PATH scan. Not worth blocking on.

Overall LGTM on the targeted fix.

nesquena-hermes:
Shipped in **v0.51.737** (via #5188). Thanks @rodboev — git is now resolved in `_run_git()` (normal PATH first, `/usr/bin/git` fallback only on macOS when PATH misses), so the version badge and in-app updater work on launchd-started macOS installs. Linux/CI behavior byte-identical; Codex SAFE-TO-SHIP + full suite green. Closes #5175.

PR #5204, opened 2026-06-29
fix(redaction): memoize deterministic API response redactor to stop GIL-bound poll storms

ducnhung70:
## Problem

On a single-process WebUI server, repeated dashboard polling of unchanged session payloads pegged one core at ~50–57% CPU **even with zero in-flight requests**, and individual `/api/session` responses ballooned to 10–44s under concurrent polling. In the browser this surfaced as repeated "connection lost" / reconnect churn.

Root cause: every `/api/session` response runs `redact_session_data()` over the full message history. The combined redactor (`agent.redact.redact_sensitive_text` with `force=True` + the local fallback) is ~15 regex passes per string. It is CPU-bound, so under the GIL it serializes across the `ThreadingHTTPServer` worker threads. Repeated polls re-run the identical redaction over identical, unchanged strings, queueing requests until the browser/reverse-proxy times out — which triggers client retries and a thread pileup, feeding back into the stall.

## Fix

The combined redactor is **pure and deterministic** (`force=True`, fixed masking — identical input always yields identical output), so memoizing per-string is safe with no invalidation required.

`api/helpers.py`:
- Wrap the built redactor in `functools.lru_cache(maxsize=4096)`.
- Gate per-entry size at 16 KB (`_REDACT_CACHE_MAX_TEXT_LEN`) so a few giant tool-output dumps can't evict the thousands of small recurring strings that actually benefit, or balloon RSS.
- Keep the public `_redact_fn_cached` name as the call target (the redaction test suite monkeypatches it wholesale, so tests bypass the cache and remain unaffected).

The redaction safety boundary is unchanged: the same patterns run on a cache miss, secrets are still masked, and the `api_redact_enabled=false` fast-path is untouched.

## Benchmark

On a representative ~3 KB string containing a credential marker:

| | per call |
|---|---|
| full redact (cache miss) | 3.718 ms |
| cache hit | 0.0020 ms |

~1900x on the repeated-poll path. A large session has hundreds of such strings, which is the difference between the 10–44s responses observed and sub-millisecond ones.

## Verification

- `./scripts/test.sh tests/test_security_redaction.py` → 69 passed, 1 skipped.
- Manual: confirmed cache hits accumulate (`lru_cache` stats), secrets still masked on hit, `>16 KB` strings bypass the cache, and the `api_redact_enabled=false` bypass still returns text unchanged.
- Idle CPU after deploy dropped from ~57% to ~0%.

## Release note

Fix a CPU/latency regression where repeated session polling re-ran the full API-response redactor over unchanged payloads, serializing behind the GIL and surfacing as "connection lost" in the browser. The deterministic redactor is now memoized.

### DIFF
--- a/api/helpers.py
+++ b/api/helpers.py
@@ -1,6 +1,7 @@
 """
 Hermes Web UI -- HTTP helper functions.
 """
+import functools
 import json as _json
 import logging
 import os
@@ -416,7 +417,25 @@ def _build_redact_fn():
     return _combined_redact
 
 
-_redact_fn_cached = _build_redact_fn()
+_redact_fn_uncached = _build_redact_fn()
+
+# Repeated dashboard polls re-request the same unchanged session payloads, so
+# the combined redactor (~15 regex passes per string) was the dominant CPU cost
+# under concurrent polling — enough to wedge the single-process server behind
+# the GIL and surface as "Mất kết nối" in the browser. The redactor is pure and
+# deterministic (force=True, fixed masking), so identical strings always map to
+# identical output and are safe to memoize without invalidation.
+_redact_fn_lru = functools.lru_cache(maxsize=4096)(_redact_fn_uncached)
+
+# Cap per-entry size so a handful of giant tool-output dumps can't evict the
+# thousands of small recurring strings that actually benefit, or balloon RSS.
+_REDACT_CACHE_MAX_TEXT_LEN = 16384
+
+
+def _redact_fn_cached(text):
+    if len(text) > _REDACT_CACHE_MAX_TEXT_LEN:
+        return _redact_fn_uncached(text)
+    return _redact_fn_lru(text)
 
 
 _SENSITIVE_CASE_MARKERS = (


nesquena-hermes:

**Certified head:** `sha:d64c2203082f` (rebased `d449ba8e`) · **PR:** #5204 · _ducnhung70 — "fix(redaction): memoize deterministic API response redactor to stop GIL-bound poll storms"_ · **crit=4 perf/reliability**
**Verdict:** Clean, correct, low-risk perf fix. Memoizing a verified-pure redactor — Codex SAFE, determinism + no-coverage-change confirmed in-process, all 75 redaction tests pass, memory-bounded. Ready to merge.

### What I ran (isolated worktree `/tmp/wt-gate-5204`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SAFE TO SHIP** (no findings) |
| Full pytest suite (`-p no:xdist`, rebased) | **11084 passed, 0 failed** (to completion) |
| Redaction test subset (security/summary/sidebar) | **75 passed** |
| In-process correctness (mine) | cached==uncached, idempotent, big-string bypass correct ✅ |

### Findings (clean)
- ✅ **Memoization is safe** — the redactor is genuinely pure/deterministic. `_build_redact_fn()` builds the combined redactor **once at module load** over fixed patterns (no runtime-mutable config captured), and `force=True` + fixed masking means identical input → identical output. So `lru_cache` needs no invalidation. Verified in-process: `_redact_fn_cached(s) == _redact_fn_uncached(s)`, idempotent.
- ✅ **No redaction-coverage change** — this only memoizes the *existing* redactor; output is byte-identical to the uncached path, and all 75 existing redaction tests pass. A cached value can't be a *new* miss-redaction (the uncached fn already produced it).
- ✅ **Cache-key complete** — `_redact_fn_cached(text)` takes a single `str` arg (the sole call site at helpers.py:556 passes only text); `lru_cache` key is the hashable string. No hidden args.
- ✅ **Memory-bounded** — `maxsize=4096` + a 16KB per-entry cap (`_REDACT_CACHE_MAX_TEXT_LEN`) routes giant tool-dumps around the cache so they can't evict the recurring small strings or balloon RSS. Thoughtful.
- ✅ **Thread-safe** — `functools.lru_cache` is thread-safe under CPython's GIL; fine for the ThreadingHTTPServer.
- ✅ **The fix targets a real problem** — repeated dashboard polls re-redacting identical unchanged payloads (~15 regex passes each) was the dominant CPU cost wedging the single-process server under concurrent polling (surfacing as connection-lost). Memoizing the deterministic redactor is the right, minimal fix.
- ⚠️ **Non-blocking nice-to-have:** no regression test added. Given the determinism + the 75 passing redaction tests, risk is low — but a small test pinning `cached==uncached` for a sensitive sample + the >16KB bypass would lock the memoization contract. Worth a follow-up, not a blocker.

**Ready to merge — cert fresh for `sha:d64c2203082f`.** Tight, correct, low-risk perf/reliability fix (Codex SAFE, redactor verified pure/deterministic, 75/75 redaction tests + full suite, memory-bounded, byte-identical output). Rebases cleanly onto current master (1-file diff). Optional follow-up: add a memoization-contract regression test. Cert valid only at `sha:d64c2203082f`.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close. Cert valid only at `sha:d64c2203082f`; a new push invalidates it → re-gate._

nesquena-hermes:
Shipped in **v0.51.749** (via #5227). Thanks @ducnhung70 — the deterministic API-response redactor is now memoized (lru_cache, 16KB per-entry bypass), eliminating the repeated ~15-regex-pass CPU cost on identical polled payloads that was wedging the single-process server behind the GIL (the 'connection lost' cause). gate-pass GREEN: Codex SAFE, redactor verified pure (cached==uncached, byte-identical output, no redaction-coverage change), 75 redaction tests + full suite green. (I added the recommended memoization-contract regression test.) Closes #5204.

Issue #1171, opened 2026-04-27
New Conversation always creates empty session even when current session is empty


nesquena-hermes:
## Summary

Clicking "New Conversation" (or pressing Cmd+K) always creates a new empty session, even when the current session has zero messages. Clicking it rapidly 5 times adds 5 empty sessions to the list. The 60-second cleanup window means they accumulate visibly, and users who frequently use Cmd+K end up with a cluttered session list.

## Steps to reproduce

1. Open the WebUI — a new empty session is created (normal)
2. Without sending any message, click "New Conversation" in the sidebar header
3. Another empty session appears in the list
4. Click it 5 more times → 5 empty "Untitled" sessions pile up
5. They eventually disappear after ~60 seconds, but during that window (and before) the list is cluttered

## Root cause — confirmed in code

**`static/boot.js:385`** — the `btnNewChat` onclick handler:

```javascript
$('btnNewChat').onclick = async () => {
  await newSession();
  await renderSessionList();
  closeMobileSidebar();
  $('msg').focus();
};
```

**`static/boot.js:517–519`** — Cmd+K shortcut:

```javascript
if((e.metaKey||e.ctrlKey) && e.key==='k'){
  if(!S.busy){ await newSession(); await renderSessionList(); ... }
}
```

Neither handler checks whether the current session already has zero messages. `newSession()` unconditionally calls `POST /api/session/new`, creating and persisting a fresh session every time.

**`api/models.py:563–567`** — the 60-second cleanup filter:

```python
result = [s for s in result if not (
    s.get('title', 'Untitled') == 'Untitled'
    and s.get('message_count', 0) == 0
    and (_now - s.get('updated_at', _now)) > 60
)]
```

Empty sessions are hidden from the sidebar *after* 60 seconds, but during the window they're fully visible. And the session files still exist on disk even after they're hidden — the filter is UI-only.

## Expected behavior

If the current session is empty (0 messages sent), clicking "New Conversation" or pressing Cmd+K should focus the composer of the existing empty session rather than creating another one.

## Proposed fix

Add a guard in both `btnNewChat` and the Cmd+K handler:

```javascript
$('btnNewChat').onclick = async () => {
  // If current session is already empty, just focus the composer
  if (S.session && (S.session.message_count || 0) === 0) {
    $('msg').focus();
    return;
  }
  await newSession();
  await renderSessionList();
  closeMobileSidebar();
  $('msg').focus();
};
```

[...]

Issue #1432, opened 2026-05-01
Can't create multiple chat discussion in parallel


Olyno:
Hi there,

I would like to be able to chat on multiple tasks at the same time (which should be possible with Hermes?). Unfortunately, whenever i click on the "+" to go to a new chat, after sending a message and starting my first chat, it keeps redirecting me to the latest chat instead of opening a new blank conversation.

nesquena-hermes:
Thanks for the report, @Olyno! This is a real bug and the failure mode is a known guard in the new-chat handler.

**Root cause confirmed in master** (verified against `static/boot.js` v0.50.254):

There's an "empty session" guard at `static/boot.js:691` (and the parallel `Cmd/Ctrl+K` path at `:844`):

```js
$('btnNewChat').onclick=async()=>{
  // If the current session has no messages, just focus the composer rather than
  // creating another empty session that will clutter the sidebar list (#1171).
  if(S.session&&(S.session.message_count||0)===0){$('msg').focus();closeMobileSidebar();return;}
  await newSession();await renderSessionList();closeMobileSidebar();$('msg').focus();
};
```

This was added for #1171 — *don't pile up empty sessions in the sidebar* — but it has an edge case: if the user clicks **+** while their **first message is still streaming** (or just before `S.session.message_count` is updated client-side), the guard fires on the stale `0` count, treats the click as "focus existing composer," and the new chat never opens. That looks exactly like the symptom you described.

**Quick repro check** — does this happen if you wait until the assistant finishes the first reply before clicking **+**? If "after assistant finishes" works but "while streaming" doesn't, that confirms the streaming-window race.

**Likely fix shape:** tighten the guard to also exclude in-flight streams, e.g.:

```js
if(S.session && (S.session.message_count||0)===0 && !S.busy && !S.session.pending_user_message){
  $('msg').focus();closeMobileSidebar();return;
}
```

So a session with a queued/streaming user turn isn't treated as empty.

I've labeled this `bug, ux` and put it on Milestone 1 (Core quality & parity) — happy to take a contributor PR on it, or we'll pick it up next sprint.

Could you also confirm:
1. Were you in the middle of streaming when you hit **+**, or after the response had finished?
2. What version of Hermes WebUI? (visible in the bottom-left corner)
3. If you click **+**, do you see anything in the browser DevTools console?

Olyno:
1. I didn't wait for the response, so it was streaming.
2. v0.50.257
3. I don't think so

[...]

PR #5206, opened 2026-06-29
Fix new conversation empty-state reuse guard

franksong2702:
## Thinking Path

- The left sidebar `+` action intentionally reuses an already-empty conversation by focusing the composer instead of creating another empty session.
- That reuse guard only checked session metadata (`message_count === 0`) plus in-flight state.
- If the loaded frontend transcript still contains visible messages while metadata says zero, the guard can misclassify the current pane as reusable-empty.
- The user-visible result is confusing: clicking `+` focuses the composer immediately, but the message pane still shows the previous conversation because `newSession()` never ran.
- This PR makes the reuse guard check the loaded transcript as well as metadata before it takes the focus-only path.

## What Changed

- Added `_currentSessionIsReusableEmptyChat()` in `static/boot.js`.
- Kept the existing zero-message, not-busy, no-active-stream, no-pending-user-message checks.
- Added a loaded-transcript check so the focus-only path is allowed only when `S.messages` has no visible non-tool messages.
- Reused the same helper for both the sidebar `+` click handler and the Cmd/Ctrl+K handler.
- Added regression coverage in `tests/test_issue_new_chat_draft_restore.py`.

## Why It Matters

Clicking `+ / New conversation` should either reuse a genuinely empty conversation or create/switch to a fresh one. It should not focus the composer while leaving an old transcript on screen. This preserves the existing sidebar hygiene behavior from #1171 while preventing the stale-pane misclassification.

This is unrelated to the recent false `Connection Lost` / large `state.db` work; it is a local frontend guard bug in the new-chat interaction path.

## Verification

- `./scripts/test.sh tests/test_issue_new_chat_draft_restore.py tests/test_regressions.py::test_newSession_clears_live_tool_cards tests/test_regressions.py::test_newSession_resets_busy_state_for_fresh_chat tests/test_regressions.py::test_renderMessages_preserves_loading_placeholder_for_session_switch` — 13 passed

No runtime restart or 8787 personal-state mutation was performed.

## Risks / Follow-ups

- The helper treats any loaded non-tool message as visible transcript content. That is intentional for this guard: if the pane has visible chat content, the focus-only new-chat shortcut should not win.
- No visual layout changed, so before/after screenshots were not added; the behavior is covered by the static interaction regression test.
- No docs update needed; this is a narrow interaction bugfix with no new user-facing feature or setup/runtime contract.

## Contract Routing

Task type: UI interaction bugfix.
Touched areas: sidebar new-chat action, keyboard new-chat shortcut, composer focus behavior.
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/UIUX-GUIDE.md`
- `DESIGN.md`
Scope boundaries: no changes to session creation API, persisted session data, streaming, or sidebar layout.
Evidence needed before claiming done: regression coverage that the reusable-empty guard checks loaded visible messages, plus existing new-session render/loading regressions.

## Model Used

AI-assisted. OpenAI GPT-5 Codex coordinated, reviewed, verified, pushed, and opened the PR. A delegated OpenAI GPT-5.5 worker investigated the original failed session task and produced the narrow patch.


nesquena-hermes:

**Certified head:** `sha:793a00526f14` (rebased `aa93c039`) · **PR:** #5206 · _franksong2702 — "Fix new conversation empty-state reuse guard"_ · **crit=3**
**Verdict:** Clean, surgical fix — closes a real "reuse-wrong-session" gap in the New Chat guard, DRY across both call sites, Codex SAFE, PR tests + suite green. Ready to merge.

### What I ran (isolated worktree `/tmp/wt-gate-5206`, rebased onto current master)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SAFE TO SHIP** (no findings) |
| Full pytest suite (`-p no:xdist`, rebased) | **11091 passed, 0 failed** (to completion) |
| PR's own tests (clean checkout) | **18 passed** |
| `node --check static/boot.js` | syntax OK |

### Findings (clean)
- ✅ **Closes the real gap.** The reuse-current-empty-chat guard (New Chat button + Cmd/Ctrl+N) previously checked only `message_count===0` + in-flight signals — but a session can have `message_count===0` while visible messages are loaded in `S.messages` (count metadata lags a load/restore), so New Chat wrongly reused a session that actually shows content. The fix adds `!hasVisibleMessages` (`S.messages.some(m => m.role && m.role!=='tool')`) to the guard.
- ✅ **DRY across both call sites.** Extracted `_currentSessionIsReusableEmptyChat()` and applied it to both the button `onclick` and the Cmd/Ctrl+N keydown — eliminates the prior duplicated-condition drift risk.
- ✅ **Happy path preserved.** A genuinely-empty fresh chat (no visible messages, count 0, not busy) is still reused (focus composer) rather than spawning a clutter session — the added condition only excludes the loaded-but-count-0 case, not the legit empty case.
- ✅ **In-flight checks intact** — `!busy && !active_stream_id && !pending_user_message` preserved (the #1432/#1171 guards); the 18 existing tests pass.
- ✅ `role!=='tool'` visible filter is correct — a tool-only row isn't user-visible content, so it shouldn't block reuse; a real user/assistant message does.

**Ready to merge — cert fresh for `sha:793a00526f14`.** Tight bugfix on an existing-flow guard (Nathan's priority order: existing-flow reliability ranks high). Codex SAFE, 18 PR tests + full suite green, DRY refactor with no behavioral drift, happy path preserved. Rebases cleanly onto current master. Cert valid only at `sha:793a00526f14`.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close. Cert valid only at `sha:793a00526f14`; a new push invalidates it → re-gate._

nesquena-hermes:
Shipped in **v0.51.750** (via #5229). Thanks @franksong2702 — the new-chat reuse guard now also requires the loaded transcript to be empty (not just metadata message_count===0), so the sidebar + can't drop you back into a conversation that still shows messages. Fresh-gated (re-pushed since the prior cert): Codex SAFE TO SHIP, no regression to genuinely-empty reuse or the keyboard new-chat path, full suite green. Closes #5206.

Issue #4793, opened 2026-06-23
Hardening follow-up: DOM-recycle reset completeness + scrollbar offsetX guard (post-#4346)


nesquena-hermes:
### Follow-up hardening from the #4346 / #4478 DOM-recycle review (v0.51.608)

Two low-severity, **non-reproducible** defensive items the Opus gate flagged when reviewing #4478 (DOM-node recycling in virtual-scroll). Neither was a ship blocker — both were reasoned from code, not observed in a running app — but they're worth closing as belt-and-suspenders hardening on the recycle path.

**F1 — Recycle reset is not defensively complete (LOW/MED).**
`static/ui.js` (recycle reset block, ~`11538`) clears only `data-transparent-turn-collapsed` and `data-transparent-turn-toggle-bound`. Container-level attrs set elsewhere — `data-anchor-scene-live-owner`, `data-anchor-stream-id` (`~9682`), `data-live-assistant-turn` (read at `~8879`) — are **not** explicitly cleared on recycle. In practice the live-node skip (`~11230`) keeps these off stashable nodes, so a leak could **not** be reproduced. But the reset relies on that invariant rather than enforcing it. **Recommendation:** clear those attrs in the recycle block (or reset via an attribute whitelist) so a future code path that settles a turn in-place can't leak stream-owner state onto a recycled node.

**F2 — `offsetX` scrollbar detection is fragile (LOW).**
`static/ui.js` (~`3921`) — `if(e.offsetX>=el.clientWidth)`. `e.offsetX` is relative to `e.target`, which for a bubbled pointerdown is often a **child** of `#messages`, not `el`. Real scrollbar clicks work (target *is* `el`), but a click on a wide child near the right edge can spuriously set `_scrollbarDragActive`. Consequence is bounded — one extra force-render + brief programmatic suppression, self-healed on pointerup. **Recommendation:** gate on `e.target===el` before the offset comparison.

(There was also an F3 perf note — the user-row `innerHTML` skip rarely fires because of browser-normalized serialization vs. a hand-built string — harmless, just less perf upside than it appears. Mentioning for completeness; lowest priority.)

Source: Opus deep-review of #4478, 2026-06-23. cc @rodboev — these build directly on your recycle work; happy for you to take them or I can pick them up.

PR #5207, opened 2026-06-29
Improve latest assistant response landmark

sheldon-im:
## Thinking Path

This is a small screen-reader navigation improvement for the settled final assistant response.

The original accessibility problem appears after streaming completes: the live assistant DOM can be replaced by the settled transcript render, and screen-reader users may need a stable way to return to the final response. To keep the state model simple, this PR does not mark the streaming/live assistant turn. It only marks the latest assistant turn during settled transcript rendering.

## What Changed

- Adds a helper that marks one assistant turn as the `Latest Hermes response` region and clears stale latest-response markers.
- Applies that helper from the settled transcript render path only.
- Keeps the transcript container, live assistant turn paths, and older assistant turns out of the landmark list.
- Adds regression coverage that verifies:
  - the transcript container is not a named region,
  - only settled render owns the latest-response landmark,
  - live/restored-live paths do not own the final-response landmark,
  - no generated headings or `tabindex` are introduced.

## Why It Matters

Assistant responses often contain Markdown headings. A response-level landmark gives screen-reader users a stable boundary for the final response while keeping heading navigation available for structure inside the answer.

Keeping the marker final-only also avoids competing ownership between live streaming DOM and the settled transcript render.

## Verification

- `./scripts/test.sh tests/test_a11y_transcript_landmarks.py`
- `./scripts/test.sh tests/test_anchor_fallback_ownership.py::test_render_messages_keeps_anchor_owned_turn_out_of_legacy_activity_rebuilds tests/test_issue4793_dom_recycle_hardening.py::test_recycled_assistant_turn_reuse_clears_live_anchor_attrs tests/test_issue4793_dom_recycle_hardening.py::test_recycled_assistant_turn_reuse_keeps_block_clear_and_role_refresh tests/test_a11y_transcript_landmarks.py`
- `node --check static/ui.js`
- `git diff --check`
- `.venv/bin/python scripts/ruff_lint.py --diff upstream/master`

Manual NVDA check on a live patched server before opening the PR: only the latest assistant response was exposed as the `Latest Hermes response` landmark, and it read cleanly. This update narrows the implementation to the settled final render path.

## Risks / Follow-ups

- This intentionally does not provide a landmark for the currently streaming live response. If live-response navigation is useful, it should be designed separately with explicit ownership across live and settled render paths.
- This intentionally does not solve navigation to older assistant responses. That may need a separate design discussion because headings, landmarks, and transcript navigation each have tradeoffs.

## Model Used

OpenAI GPT-5.5 via Hermes Agent, with local shell/git/test tooling.

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -453,6 +453,9 @@ const _recycleResetAttrs=[
   'data-transparent-turn-toggle-bound',
   'data-anchor-scene-live-owner',
   'data-anchor-stream-id',
+  'data-latest-assistant-response',
+  'role',
+  'aria-label',
   // Defensive reset for legacy/restored shells that may still carry the fallback live-turn marker.
   'data-live-assistant-turn',
 ];
@@ -8617,6 +8620,24 @@ function _createAssistantTurn(tsTitle='', tpsText=''){
   row.innerHTML=`${_assistantRoleHtml(tsTitle, tpsText)}<div class="assistant-turn-blocks"></div>`;
   return row;
 }
+function _setLatestAssistantTurnLandmark(turn, isLatest){
+  if(!turn) return;
+  const label='Latest Hermes response';
+  if(isLatest){
+    if(typeof document!=='undefined'){
+      document.querySelectorAll('.assistant-turn[data-latest-assistant-response="true"]').forEach(el=>{
+        if(el!==turn) _setLatestAssistantTurnLandmark(el, false);
+      });
+    }
+    turn.setAttribute('role','region');
+    turn.setAttribute('aria-label',label);
+    turn.dataset.latestAssistantResponse='true';
+    return;
+  }
+  if(turn.getAttribute('role')==='region') turn.removeAttribute('role');
+  if(turn.getAttribute('aria-label')===label) turn.removeAttribute('aria-label');
+  delete turn.dataset.latestAssistantResponse;
+}
 function _assistantTurnBlocks(turn){
   return turn?turn.querySelector('.assistant-turn-blocks'):null;
 }
@@ -12214,6 +12235,13 @@ function renderMessages(options){
     }catch(e){}
   }
   const transparentToolResultsByTid=_transparentModeActive?_collectToolResultSnippetsByTid(S.messages):{};
+  const latestRenderedAssistantRawIdx=(()=>{
+    for(let i=renderVisWithIdx.length-1;i>=0;i--){
+      const entry=renderVisWithIdx[i];
+      if(entry&&entry.m&&entry.m.role==='assistant'&&!entry.m._live) return entry.rawIdx;
+    }
+    return -1;
+  })();
   // Windowed render loop replaces the legacy full loop:
   // for(let vi=0;vi<visWithIdx.length;vi++)
   for(let vi=0;vi<renderVisWithIdx.length;vi++){
@@ -12401,6 +12429,7 @@ function renderMessages(options){
       currentAssistantTurn.dataset.recycleKey=rawIdx;
       inner.appendChild(currentAssistantTurn);
     }
+    _setLatestAssistantTurnLandmark(currentAssistantTurn, !m._live&&rawIdx===latestRenderedAssistantRawIdx);
     const seg=document.createElement('div');
     if(Array.isArray(orderedTransparentParts)&&orderedTransparentParts.length){
       const blocks=_assistantTurnBlocks(currentAssistantTurn);
--- a/tests/test_a11y_transcript_landmarks.py
+++ b/tests/test_a11y_transcript_landmarks.py
@@ -1,0 +1,54 @@
+import re
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+INDEX_HTML = (ROOT / "static" / "index.html").read_text(encoding="utf-8")
+UI_JS = (ROOT / "static" / "ui.js").read_text(encoding="utf-8")
+
+
+def _function_body(source: str, name: str) -> str:
+    start = source.find(f"function {name}(")
+    assert start != -1, f"{name}() not found"
+    end = source.find("\nfunction ", start + 1)
+    if end == -1:
+        end = len(source)
+    return source[start:end]
+
+
+def test_conversation_transcript_is_not_a_named_region():
+    match = re.search(r'<div class="messages" id="messages"[^>]*>', INDEX_HTML)
+    assert match, "#messages container not found"
+    tag = match.group(0)
+    assert 'role="region"' not in tag
+    assert 'Conversation transcript' not in tag
+    assert "tabindex" not in tag.lower()
+
+
+def test_latest_assistant_landmark_helper_is_named_but_not_focusable():
+    helper = _function_body(UI_JS, "_setLatestAssistantTurnLandmark")
+    assert "Latest Hermes response" in helper
+    assert "setAttribute('role','region')" in helper
+    assert "setAttribute('aria-label',label)" in helper
+    assert "data-latest-assistant-response" in helper
+    assert "tabindex" not in helper.lower()
+    assert "<h" not in helper.lower()
+
+
+def test_settled_render_marks_only_the_latest_assistant_turn():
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
# 🔬 Gate certification — RED ⛔ (clean a11y improvement, but the live assistant turn wrongly gets the "latest final response" landmark)

**Certified head:** `sha:89529142f45a` (rebased `66961935`) · **PR:** #5207 · _sheldon-im — "Improve latest assistant response landmark"_ · **crit=2**
**Verdict:** A well-scoped accessibility improvement — marks exactly one assistant turn as a named ARIA `region` ("Latest Hermes response") for screen-reader navigation, clearing the landmark from prior turns. **But Codex verified one SILENT bug**: the "latest" scan doesn't exclude `_live` (streaming) messages, so the **in-progress live turn** gets tagged as the *final* response landmark prematurely. RED until the `_live` guard is added.

### What I ran (isolated worktree `/tmp/wt-gate-5207`, rebased onto current master)
| Gate | Result |
|------|--------|
| Codex (reproduce/read) | **SHIP ONLY WITH FIXES** — 1 SILENT (`_live` turn gets the final-response landmark) |
| Full pytest suite (`-p no:xdist`, rebased) | _(running)_ |
| PR's tests (`test_a11y_transcript_landmarks`, +4) | **4 passed** |
| `node --check static/ui.js` | OK |

### ⛔ Blocking (SILENT, Codex-verified + I confirmed) — live turn gets the "Latest Hermes response" landmark prematurely
`static/ui.js:12238` (diff lines 37 + 48): `latestRenderedAssistantRawIdx` counts `entry.m.role === 'assistant'` **without excluding `entry.m._live`**, and `_setLatestAssistantTurnLandmark(currentAssistantTurn, rawIdx===latestRenderedAssistantRawIdx)` has no `_live` guard. But `renderMessages()` **does render `_live:true` assistant messages** during active-stream restore/rebuild (Codex cited `sessions.js:665`, `sessions.js:2605`, `messages.js:2253`). So during streaming, the **not-yet-final live turn** is tagged `role="region" aria-label="Latest Hermes response"` — a screen reader would announce/navigate to it as the *final* response while it's still being written. (Confirmed: diff line 37 has no `_live` check.)
**Fix (Codex):**
1. `ui.js:12238` — exclude `_live` in the latest scan: `if(entry&&entry.m&&entry.m.role==='assistant'&&!entry.m._live) return entry.rawIdx;`
2. `ui.js:12429` — guard the setter: `_setLatestAssistantTurnLandmark(currentAssistantTurn, !m._live && rawIdx===latestRenderedAssistantRawIdx);`
The PR's own test `test_live_assistant_turn_paths_do_not_own_final_response_landmark` asserts exactly this property — so it should have caught it; the test likely checks a different path than the `_live`-render restore/rebuild path. Add/extend a test for the `_live` active-stream-restore case.

### ✅ Right (keep)
The single-landmark design is sound: clears the landmark from other turns (exactly one at a time, guarded by `===label`/`===region` so it doesn't clobber a pre-existing role/label), transcript itself is not a named region (no landmark noise), helper is named-but-not-focusable (no tabindex hijack). Only the `_live`-exclusion is missing.

**Bounce for the 2-line `_live` guard (Codex-verified, precise), then re-gate.** crit=2 a11y — without it, a screen reader treats the actively-streaming turn as the final "Latest Hermes response," which is misleading mid-stream. The landmark design is otherwise clean. Add a regression test for the active-stream-restore `_live` path (the existing live-path test missed it). Cert valid only at `sha:89529142f45a`.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close. Cert valid only at `sha:89529142f45a`; a new push invalidates it → re-gate._

### ⛔ Blocking #2 (CORE — PR breaks 3 existing tests) — `_setLatestAssistantTurnLandmark is not defined` in the extraction harness
The full suite fails 3 PRE-EXISTING tests (reproduced in isolation, NOT env flakes):
- `test_anchor_fallback_ownership.py::test_render_messages_keeps_anchor_owned_turn_out_of_legacy_activity_rebuilds`
- `test_issue4793_dom_recycle_hardening.py::test_recycled_assistant_turn_reuse_clears_live_anchor_attrs`
- `test_issue4793_dom_recycle_hardening.py::test_recycled_assistant_turn_reuse_keeps_block_clear_and_role_refresh`

All fail with **`ReferenceError: _setLatestAssistantTurnLandmark is not defined`**. These tests extract the `renderMessages()` body and `eval` it in a JS harness; the PR added a **call to `_setLatestAssistantTurnLandmark(...)` inside `renderMessages()`**, but the helper's *definition* is outside the extracted scope, so the harness throws. The PR introduced the call without making the helper resolvable to (or stubbed in) these existing harnesses.
**Fix:** either define/stub `_setLatestAssistantTurnLandmark` within the extracted scope the tests eval, or have those test harnesses provide the helper (mirror how sibling helpers called from `renderMessages()` are handled). Also — the #4793 DOM-recycle tests imply a **recycled assistant turn must CLEAR the new `role`/`aria-label`/`data-latest-assistant-response` attrs on reuse**; confirm `_setLatestAssistantTurnLandmark(turn,false)` (or the recycle-clear path) actually runs when a turn is recycled, else a recycled turn keeps a stale landmark.

**Updated verdict:** RED on TWO blockers — (1) the `_live` premature-landmark, and (2) the PR breaks 3 existing tests (`_setLatestAssistantTurnLandmark` unresolved in their harness + the recycle-clear contract). Both must be fixed before re-gate.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (re-cert: both prior blockers fixed)

**Certified head:** `sha:28236dc6` · **PR:** #5207 · _sheldon-im — "Improve latest assistant response landmark"_ · **crit=2**
**Verdict:** Both prior RED blockers are fixed — the `_live` streaming turn no longer gets the "Latest Hermes response" final landmark, and the 3 existing tests it broke now pass (the helper is stubbed in their extraction harnesses). Codex SAFE, suite green. Ready to merge.

### What I ran (isolated worktree `/tmp/wt-gate-5207r`, fresh-based on current master)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SAFE TO SHIP** (no findings — both blockers closed) |
| Full pytest suite (`-p no:xdist`) | **11188 passed, 0 failed** |
| Previously-broken tests (anchor-ownership + #4793 DOM-recycle + a11y) | **19 passed** |
| `node --check static/ui.js` | OK |

### Both prior blockers fixed (verified)
1. ✅ **`_live` guard added** — latest-scan now `role==='assistant'&&!entry.m._live` (ui.js:12238) + setter `!m._live&&rawIdx===…` (ui.js:12432), so the actively-streaming turn no longer gets the final-response landmark; only the latest SETTLED turn does.
2. ✅ **3 broken tests fixed** — the harnesses now stub `_setLatestAssistantTurnLandmark` (test_anchor_fallback_ownership:607, test_issue4793_dom_recycle_hardening:93), so the `renderMessages()` extraction-eval resolves it. Recycled turns clear the landmark attrs (ui.js:456-458); the helper's clear path preserves unrelated role/aria-label. (Codex verified all.)

**Ready to merge — cert fresh for `sha:28236dc6`.** Both prior blockers closed (the `_live` premature-landmark + the broken extraction-harness tests), Codex SAFE, full suite green, 19 prior-failing tests now pass. Clean single-landmark a11y improvement. Concept 4/5. Cert valid only at `sha:28236dc6`.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy/close. Cert valid only at `sha:28236dc6`._

nesquena-hermes:
Shipped in **v0.51.768** (Wave 4 batch, via release PR #5281). Thanks @sheldon-im! Codex SAFE (rebased 0-behind master, no stale-base) + full suite green.

nesquena-hermes:
This shipped in **v0.51.768** (verified: the change is on `master` + has a CHANGELOG entry). The release landed via a batched release PR, which didn't auto-close this one — closing now. Thanks @sheldon-im! 🎉

Issue #3954, opened 2026-06-10
New chat shortcut (Ctrl/Cmd+K) conflicts with j/k session navigation


TomBanksAU:
## Feature Description
Change the new conversation keyboard shortcut from `Ctrl/Cmd+K` to `Ctrl/Cmd+Shift+O`.

## Motivation
`Ctrl/Cmd+K` now conflicts with the j/k session navigation bindings added in #3845. Users who want to navigate sessions with j and k may accidentally trigger a new chat when pressing Ctrl+k.

Additionally, `Ctrl+Shift+O` matches the convention used by ChatGPT and Claude for opening a new conversation, making it more intuitive for users coming from other AI chat platforms.

## Proposed Solution
Change the key binding so that `Ctrl/Cmd+Shift+O` triggers `newSession()` instead of `Ctrl/Cmd+K`. Update the tooltip on the New Conversation button to show `(Cmd+Shift+O)`.

A PR with the change is at #3953.

## Acceptance Criteria
- [ ] `Ctrl/Cmd+Shift+O` creates a new conversation from anywhere
- [ ] Tooltip on the New Conversation button reads `(Cmd+Shift+O)`
- [ ] Existing tests pass with updated assertions

## Alternatives Considered
- **Keep Ctrl+K and change j/k**: j/k is a well-established vim-style navigation convention; changing it would be more disruptive
- **Ctrl+N**: conflicts with browser new-window shortcut
- **Ctrl+T**: conflicts with browser new-tab shortcut

nesquena-hermes:
Thanks @TomBanksAU — confirmed against `origin/master`: the new-chat shortcut is `Ctrl/Cmd+K` (`boot.js:1521`, `if((e.metaKey||e.ctrlKey)&&e.key==='k')`) and the `j`/`k` session navigation from #3845 is live (`sessions.js:7046`). They don't strictly collide (the nav bindings are unmodified `j`/`k`), but the muscle-memory overlap is real — someone reaching for `k`-to-navigate who happens to be holding a modifier lands on New Chat, and `Ctrl/Cmd+Shift+O` is the established convention in ChatGPT/Claude for a new conversation, so the change is well-motivated on both fronts.

Keeping it as a sprint-candidate. The change itself is small (rebind in `boot.js`), but worth a quick check that `Ctrl/Cmd+Shift+O` doesn't collide with a browser/OS shortcut on any platform before we land it. Happy to take a PR.

rodboev:
[...]

PR #5208, opened 2026-06-29
fix(ux): skip Ctrl+K new-chat shortcut in text inputs (#5099)

### LINKED ISSUE #5099: Consider emacs (and emacs-adjacent) users - redefine ctrl-k
In emacs ctrl-k kills a line. This is supported by most text box components on Linux and OS X systems. ctrl-k is as natural as ctrl-a and ctrl-e when editing. 

The hermes admin currently uses ctrl-k to create a new conversation. Every day I mistakenly create dozens of new conversations!

nankingjing:
## Summary

- When focus is in an `INPUT`, `TEXTAREA`, or `contentEditable` field, **Cmd/Ctrl+K no longer hijacks the key** — Emacs-adjacent users can use kill-line in the composer and other editable fields.
- When focus is outside text inputs, the global new-chat shortcut behavior is unchanged.

Fixes #5099

## Implementation

Mirrors the existing Ctrl+B guard in `static/boot.js`: detect editable target, early-return before `preventDefault()`, then run the existing new-session logic.

## Test plan

- [x] Added `tests/test_issue5099_ctrl_k_text_input_guard.py` (4 static-analysis regressions)
- [x] Existing Ctrl+K regressions in `tests/test_1432_newchat_and_1423_profile_input.py` and `tests/test_mobile_layout.py` remain compatible

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -1738,6 +1738,9 @@ document.addEventListener('keydown',async e=>{
     }
   }
   if((e.metaKey||e.ctrlKey)&&e.key==='k'){
+    const t=e.target;
+    const isText=t&&(t.tagName==='INPUT'||t.tagName==='TEXTAREA'||t.isContentEditable);
+    if(isText) return;
     e.preventDefault();
     // If the current session has no messages AND nothing is in flight, just focus
     // the composer rather than creating another empty session that will clutter
--- a/tests/test_issue5099_ctrl_k_text_input_guard.py
+++ b/tests/test_issue5099_ctrl_k_text_input_guard.py
@@ -1,0 +1,55 @@
+"""Static-analysis tests for #5099 (Ctrl+K must not steal kill-line in text fields).
+
+Emacs-adjacent users expect Ctrl+K to kill to end-of-line while the composer
+or other editable fields are focused. Cmd/Ctrl+K should still create a new chat
+when focus is outside text inputs, matching the existing Ctrl+B guard pattern.
+"""
+from pathlib import Path
+
+BOOT_JS = (Path(__file__).parent.parent / "static" / "boot.js").read_text(encoding="utf-8")
+
+
+def _ctrl_k_branch_window() -> str:
+    idx = BOOT_JS.find("(e.metaKey||e.ctrlKey)&&e.key==='k'")
+    assert idx >= 0, "Cmd/Ctrl+K handler not found in boot.js"
+    return BOOT_JS[idx:idx + 900]
+
+
+class TestIssue5099CtrlKTextInputGuard:
+    def test_ctrl_k_skips_text_inputs(self):
+        branch = _ctrl_k_branch_window()
+        assert "tagName==='INPUT'" in branch
+        assert "tagName==='TEXTAREA'" in branch
+        assert "isContentEditable" in branch
+        assert "if(isText) return" in branch
+
+    def test_ctrl_k_prevent_default_after_text_guard(self):
+        branch = _ctrl_k_branch_window()
+        guard_idx = branch.find("if(isText) return")
+        prevent_idx = branch.find("e.preventDefault()")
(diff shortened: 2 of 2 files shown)


nesquena-hermes (changes requested):
Thanks for the fix @nankingjing — the production change in `static/boot.js` is **correct and complete**, and I verified it carefully:

- The Cmd/Ctrl+K guard (`boot.js:1740-1743`) skips the new-chat shortcut for text inputs and returns **before** `preventDefault()`, so the keystroke is left for the focused field (the #5099 Emacs kill-line case). ✅
- The editable check covers `<input>`, `<textarea>`, and `t.isContentEditable` — using the **computed** `.isContentEditable` property is more robust than an attribute check, since it correctly returns true for `contenteditable=""`/`"true"` **and** nested/inherited contenteditable. ✅
- No regression for the normal case (non-editable target → falls through to `preventDefault()` + `newSession()`), no collision with any other `k` handler (the J/K session nav in `sessions.js` bails on `ctrlKey||metaKey||altKey`), and Cmd+K (metaKey) gets identical treatment. ✅
- Mirrors the existing Ctrl+B guard idiom in the same file exactly. ✅

The one blocker is that **your own test file ships 2 of its 4 tests RED**, so CI is red. Both failures are undersized substring windows in `tests/test_issue5099_ctrl_k_text_input_guard.py` — the production code already satisfies the assertions, the windows just don't reach the text:

### 🔴 `test_ctrl_k_still_creates_new_session_outside_inputs`
`_ctrl_k_branch_window()` (line 15) returns `BOOT_JS[idx:idx + 900]`. The `newSession()`/`closeMobileSidebar()` call is at `boot.js:1761`, but a ~600-char comment block (the #1171/#1432/streaming explanation, `boot.js:1745-1760`) sits between the guard and that call, so the 900-char slice stops short.
**Fix:** widen the slice to `BOOT_JS[idx:idx + 1500]` (or search forward for `"newSession()"` instead of a fixed-length slice).

### 🔴 `test_ctrl_k_guard_matches_ctrl_b_idiom`
`ctrl_b_block = BOOT_JS[max(0, ctrl_b_idx - 250):ctrl_b_idx + 120]` (line 40). The `+120` forward window truncates mid-token at `t.i`, so the `isContentEditable` needle isn't found in the Ctrl+B block (it ends ~135 chars past the match start).
**Fix:** widen the forward window to `ctrl_b_idx + 200`.

With those two constants bumped, all 4 tests go green (the other two — `test_ctrl_k_skips_text_inputs` and the `preventDefault`-ordering check — already pass). Re-push and re-request review and this should be quick to land.

nesquena-hermes:
Thanks @nankingjing — and sorry to close this one. We're resolving the Cmd+K typing conflict (#5099) a different way: **#5185** moves the new-chat shortcut entirely off Cmd/Ctrl+K to Cmd/Ctrl+Shift+O (per #3954), which frees Cmd+K globally — including inside text inputs — so the kill-line conflict you hit goes away as a side effect, and Cmd+K becomes available for the J/K session-navigation workflow.

Since #5185 removes the binding rather than conditionally suppressing it, the two approaches overlap and we're going with the move. Your instinct on the conflict was right — really appreciate the fix and the clear repro of #5099; it's resolved by #5185. Closing in favor of that.

nesquena-hermes:
Shipped in **v0.51.746** (via #5222). Thanks @nankingjing — Cmd/Ctrl+K now early-returns when focus is in a text field (INPUT/TEXTAREA/contentEditable), so kill-line and other editing keys work in the composer; outside text inputs it still creates a new conversation. Your instinct on the #5099 conflict + the Ctrl+B-guard idiom were spot on. (I auto-fixed two too-short test-window slices so the guard assertions actually reach the code — non-vacuous now, 4/4 pass.) Maintainer-approved; chosen over the alternative #5185 whose Cmd+Shift+O target was browser-reserved. Closes #5099.

PR #5216, opened 2026-06-29
i18n(pl): translate virtualize-transcript setting, strip BOM

leszek3737:
Translate settings_label_virtualize_transcript and its description into Polish; they were still showing the English fallback. Also remove the stray UTF-8 BOM from the top of the file.


nesquena-hermes:
Shipped in **v0.51.751** (via #5230). Thanks @leszek3737 — Polish 'Virtualize long transcripts' label + description are now translated, and the stray UTF-8 BOM at the top of static/i18n.js is stripped. Locale-only, parity-safe (307 i18n tests pass), full suite green. Closes #5216.

PR #5239, opened 2026-06-29
feat: support WebUI client identity metadata

promptclickrun:
## Thinking Path

- Shared WebUI deployments can have multiple people, devices, or client apps sending turns into the same agent runtime.
- The backend needs a small, explicit way to pass trusted client identity into hidden runtime context.
- This should not alter the visible transcript or pretend a user sent an extra message.
- Because the session-key behavior touches gateway continuity scope, this PR now stays identity-only so that security review is clean.

## Contract Routing

Task type: browser chat runtime metadata and gateway chat forwarding.

Touched areas:
- `/api/chat/start` request handling
- in-process WebUI streaming context
- gateway-backed chat forwarding
- advanced chat setup docs

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/advanced-chat-setup.md`

Scope boundaries:
- This metadata is context, not authentication.
- This PR does not add auth, authorization, account linking, or user management.
- The Copilot catalog and MoA model-picker work was split into #5301.

- Header and body extraction tests.
- Sanitization tests for control characters and CRLF/header-injection cases.
- In-process system-prompt context tests.
- Gateway forwarded-header tests.
- Existing WebUI surface/platform hint regression tests.

## What Changed

This PR adds optional client identity metadata for browser-originated WebUI chat turns.

Trusted clients can attach identity metadata to `/api/chat/start` using headers:

```http
X-Hermes-Client-Name: Person A
X-Hermes-Client-Id: person-a-ios
X-Hermes-Session-Key: team:person-a:ios
```

Or with a JSON body field:

```json
{
  "client_identity": {
    "name": "Person A",
    "id": "person-a-ios",
    "session_key": "team:person-a:ios"
  }
}
```

Headers take precedence over the JSON body if both are present.

Implementation details:

- Reads optional client identity metadata on `/api/chat/start`.
- Supports both header-based and body-based clients.
- Sanitizes identity values by stripping control characters, normalizing whitespace, and capping length.
- Adds the metadata to hidden runtime context for the current turn.
- Keeps the visible transcript unchanged. No fake user message is added.
- Threads the same metadata through both in-process browser chat and gateway-backed browser chat.
- For gateway-backed chat, forwards sanitized identity headers downstream and preserves the existing WebUI session-key fallback when no client session key is provided.
- Documents the contract in `docs/advanced-chat-setup.md`.

## Why It Matters

Multi-user clients need a standard way to tell the agent who is speaking without requiring a separate backend profile per person or device. This gives trusted clients a small metadata channel while keeping WebUI auth and transcript semantics unchanged.

## Security Note

This metadata is intentionally treated as context, not authentication. It should help an agent understand which trusted client sent a turn. It should not be used for authorization decisions unless paired with the normal WebUI auth/access-control layer.

The remaining trust decision for maintainers is whether client-supplied `X-Hermes-Session-Key` / `client_identity.session_key` should be allowed to steer gateway continuity scope. The current behavior allows it after sanitization and falls back to the server-owned `webui:{session_id}` key when absent.

## Verification

```bash
./scripts/test.sh tests/test_webui_client_identity_context.py tests/test_webui_gateway_chat_backend.py tests/test_webui_surface_context.py tests/test_webui_platform_hint.py -q
```

Result:

```text
55 passed in 6.24s
```

Additional readback after force-push:

```text
PR head: 7dbdd91792a4bef5362bbfcb410d8108b0ef71fb
Changed files: 6
Files: api/gateway_chat.py, api/routes.py, api/streaming.py, docs/advanced-chat-setup.md, tests/test_webui_client_identity_context.py, tests/test_webui_gateway_chat_backend.py
```

## Risks / Follow-ups

- Maintainers should explicitly decide whether client-supplied session keys may override server-owned gateway continuity scope.
- If server-owned continuity should always win, this can be adjusted so the client session key is context-only and never forwarded as `X-Hermes-Session-Key`.

## Model Used

AI-assisted by Hermes Agent using GPT-5.5 via GitHub Copilot provider, with local git/GitHub CLI tooling and the repo `./scripts/test.sh` runner.

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -506,6 +506,7 @@ def _run_gateway_chat_streaming(
     *,
     model_provider=None,
     goal_related=False,
+    client_identity=None,
 ):
     """Bridge a WebUI chat turn through Hermes Gateway's API server.
 
@@ -585,6 +586,7 @@ def _run_gateway_chat_streaming(
                 _prefill_messages_with_webui_context,
                 _normalize_prefill_messages_before_user_turn,
                 _public_prefill_context_status,
+                _webui_client_identity_context,
                 _webui_ephemeral_system_prompt,
             )
 
@@ -604,6 +606,9 @@ def _run_gateway_chat_streaming(
                 },
                 config_data=cfg,
             )
+            identity_context = _webui_client_identity_context(client_identity)
+            if identity_context:
+                _gateway_system_prompt = f"{_gateway_system_prompt}\n\n{identity_context}".rstrip()
             prefill_messages = _prefill_messages_with_webui_context(prefill_context, cfg)
             prefill_messages = _normalize_prefill_messages_before_user_turn(prefill_messages)
             prefill_messages = [
@@ -683,11 +688,21 @@ def _run_gateway_chat_streaming(
                 "Accept": "text/event-stream",
                 "X-Hermes-Session-Id": session_id,
             }
+            if isinstance(client_identity, dict):
+                client_name = str(client_identity.get("name") or "").strip()
+                client_id = str(client_identity.get("id") or "").strip()
+                client_session_key = str(client_identity.get("session_key") or "").strip()
+                if client_name:
+                    headers["X-Hermes-Client-Name"] = client_name
+                if client_id:
+                    headers["X-Hermes-Client-Id"] = client_id
+                if client_session_key:
+                    headers["X-Hermes-Session-Key"] = client_session_key
             if api_key:
                 headers["Authorization"] = f"Bearer {api_key}"
                 # Scope Gateway long-term continuity to this WebUI conversation
                 # without exposing the browser's auth cookie or CSRF material.
-                headers["X-Hermes-Session-Key"] = f"webui:{session_id}"
+                headers.setdefault("X-Hermes-Session-Key", f"webui:{session_id}")
             message_content: Any = str(msg_text or "")
             if attachments:
                 try:
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -616,7 +616,6 @@ def _run_gateway_chat_streaming(
                 _prefill_messages_with_webui_context,
                 _normalize_prefill_messages_before_user_turn,
                 _public_prefill_context_status,
-                _webui_client_identity_context,
                 _webui_ephemeral_system_prompt,
                 _webui_system_prompt_with_client_identity,
             )
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -588,6 +588,7 @@ def _run_gateway_chat_streaming(
                 _public_prefill_context_status,
                 _webui_client_identity_context,
                 _webui_ephemeral_system_prompt,
+                _webui_system_prompt_with_client_identity,
             )
 
             prefill_context = _load_webui_prefill_context(cfg)
@@ -606,9 +607,10 @@ def _run_gateway_chat_streaming(
                 },
                 config_data=cfg,
             )
-            identity_context = _webui_client_identity_context(client_identity)
-            if identity_context:
-                _gateway_system_prompt = f"{_gateway_system_prompt}\n\n{identity_context}".rstrip()
+            _gateway_system_prompt = _webui_system_prompt_with_client_identity(
+                _gateway_system_prompt,
+                client_identity,
+            )
             prefill_messages = _prefill_messages_with_webui_context(prefill_context, cfg)
             prefill_messages = _normalize_prefill_messages_before_user_turn(prefill_messages)
             prefill_messages = [
@@ -689,9 +691,11 @@ def _run_gateway_chat_streaming(
                 "X-Hermes-Session-Id": session_id,
             }
             if isinstance(client_identity, dict):
-                client_name = str(client_identity.get("name") or "").strip()
-                client_id = str(client_identity.get("id") or "").strip()
-                client_session_key = str(client_identity.get("session_key") or "").strip()
+                from api.streaming import _clean_webui_client_identity_value
+
+                client_name = _clean_webui_client_identity_value(client_identity.get("name"))
+                client_id = _clean_webui_client_identity_value(client_identity.get("id"), max_len=160)
+                client_session_key = _clean_webui_client_identity_value(client_identity.get("session_key"), max_len=256)
                 if client_name:
                     headers["X-Hermes-Client-Name"] = client_name
                 if client_id:
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -118,6 +118,36 @@ def _gateway_reasoning_effort_for_request(cfg, *, model=None, model_provider=Non
         return None
 
 
+def _gateway_client_identity_headers(session_id: str, api_key: str, client_identity=None) -> dict[str, str]:
+    """Build sanitized client identity headers for Gateway chat transports."""
+    headers: dict[str, str] = {}
+    client_session_key = ""
+    if isinstance(client_identity, dict):
+        from api.streaming import _clean_webui_client_identity_value
+
+        client_name = _clean_webui_client_identity_value(client_identity.get("name"))
+        client_id = _clean_webui_client_identity_value(client_identity.get("id"), max_len=160)
+        client_session_key = _clean_webui_client_identity_value(client_identity.get("session_key"), max_len=256)
+        if client_name:
+            headers["X-Hermes-Client-Name"] = client_name
+        if client_id:
+            headers["X-Hermes-Client-Id"] = client_id
+    if api_key:
+        # Scope Gateway long-term continuity to this WebUI conversation without
+        # exposing the browser's auth cookie or CSRF material. A client-supplied
+        # session_key is namespaced under the server-owned webui:{session_id}
+        # scope instead of forwarded verbatim. WebUI multiplexes browser turns
+        # through one shared gateway Bearer key, so the agent-side API-key guard
+        # cannot distinguish sibling browser clients.
+        if client_session_key:
+            headers["X-Hermes-Session-Key"] = f"webui:{session_id}:{client_session_key}"
+        else:
+            headers["X-Hermes-Session-Key"] = f"webui:{session_id}"
+    # On an unauthenticated gateway, do not send X-Hermes-Session-Key at all.
+    # The gateway rejects that header with 403 when no API key is configured.
+    return headers
+
+
 def gateway_chat_config_status(config_data=None, environ: dict[str, str] | None = None) -> dict:
     """Return redacted Gateway-backed chat configuration status."""
     mode = webui_chat_backend_mode(config_data, environ)
@@ -282,7 +312,7 @@ def _run_gateway_runs_api_streaming(
     session_id, msg_text, model, workspace, stream_id,
     base_url, api_key, prefill_messages, body_extras,
     *, put_gateway_event, cancel_event,
-    attachments=None, cfg=None, session=None,
+    attachments=None, cfg=None, session=None, client_identity=None,
 ):
     """Submit via POST /v1/runs and relay SSE events including approval."""
     url_runs = f"{base_url.rstrip('/')}/v1/runs"
@@ -290,10 +320,10 @@ def _run_gateway_runs_api_streaming(
         "Content-Type": "application/json",
         "Accept": "application/json",
         "X-Hermes-Session-Id": session_id,
+        **_gateway_client_identity_headers(session_id, api_key, client_identity),
     }
     if api_key:
         headers["Authorization"] = f"Bearer {api_key}"
-        headers["X-Hermes-Session-Key"] = f"webui:{session_id}"
     message_content: Any = str(msg_text or "")
     if attachments:
         try:
@@ -654,6 +684,7 @@ def _run_gateway_chat_streaming(
                     attachments=attachments,
                     cfg=cfg,
                     session=s,
+                    client_identity=client_identity,
                 )
             except Exception as exc:
                 put_gateway_event("apperror", {
@@ -689,43 +720,10 @@ def _run_gateway_chat_streaming(
                 "Content-Type": "application/json",
                 "Accept": "text/event-stream",
                 "X-Hermes-Session-Id": session_id,
+                **_gateway_client_identity_headers(session_id, api_key, client_identity),
             }
-            client_name = None
-            client_id = None
-            client_session_key = None
-            if isinstance(client_identity, dict):
-                from api.streaming import _clean_webui_client_identity_value
-
-                client_name = _clean_webui_client_identity_value(client_identity.get("name"))
-                client_id = _clean_webui_client_identity_value(client_identity.get("id"), max_len=160)
-                client_session_key = _clean_webui_client_identity_value(client_identity.get("session_key"), max_len=256)
-                if client_name:
-                    headers["X-Hermes-Client-Name"] = client_name
-                if client_id:
-                    headers["X-Hermes-Client-Id"] = client_id
             if api_key:
                 headers["Authorization"] = f"Bearer {api_key}"
-                # Scope Gateway long-term continuity to this WebUI conversation
-                # without exposing the browser's auth cookie or CSRF material.
-                # A client-supplied session_key is namespaced under the
-                # server-owned webui:{session_id} scope rather than forwarded
-                # verbatim: the agent-side gateway gates a caller-supplied
-                # session key behind API-key auth specifically to stop one
-                # caller from guessing another's memory scope, and WebUI
-                # multiplexes every browser turn through a single shared
-                # gateway Bearer key, so distinct browser clients are
-                # indistinguishable to that guard. Namespacing lets a client
-                # sub-scope within its own conversation boundary without being
-                # able to collide with or steer into a sibling's scope.
-                if client_session_key:
-                    headers["X-Hermes-Session-Key"] = f"webui:{session_id}:{client_session_key}"
-                else:
-                    headers.setdefault("X-Hermes-Session-Key", f"webui:{session_id}")
-            # Note: on an unauthenticated gateway (no api_key), we intentionally
-            # do not send X-Hermes-Session-Key at all, even if the client
-            # supplied identity metadata — the gateway hard-rejects that header
-            # with 403 when no API key is configured, so sending it here would
-            # break an otherwise-working open local gateway chat turn.
 
             message_content: Any = str(msg_text or "")
             if attachments:
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -690,6 +690,9 @@ def _run_gateway_chat_streaming(
                 "Accept": "text/event-stream",
                 "X-Hermes-Session-Id": session_id,
             }
+            client_name = None
+            client_id = None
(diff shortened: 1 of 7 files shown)


nesquena-hermes (changes requested):
Thanks for the work here — the client-identity input sanitization is genuinely solid (the control-char strip + length caps in `_clean_webui_client_identity_value` neutralize CRLF/header-injection, and the "treat as context metadata, not authentication" framing is the right instinct). A few things need to change before this can move forward, though.

## 1. Blocker — this breaks an existing test (`tests/test_issue5057_moa_webui_route.py`)

`api/commands.py::resolve_moa_config` now imports `resolve_moa_preset` in the **same statement** as the other two symbols:

```python
from hermes_cli.moa_config import moa_usage, normalize_moa_config, resolve_moa_preset
```

The existing regression test installs a fake `hermes_cli.moa_config` (`_install_fake_moa_config`) that intentionally exposes **only** `normalize_moa_config` + `moa_usage` to simulate an older/partial agent. Because Python's `from … import a, b, c` fails entirely if **any** name is missing, the combined import now raises `ImportError` → your `except` re-raises `RuntimeError("MoA runtime unavailable …")`. So:

- `test_resolve_moa_config_returns_expected_shape` → was green on master, now **red** (expects a dict, gets a RuntimeError).
- `test_resolve_moa_config_degrades_without_config` → same, now **red**.

This also removes the graceful-degradation path for an older hermes-agent that your own error message names ("not installed or too old"). Please split the import so a missing `resolve_moa_preset` degrades instead of breaking the whole function:

```python
try:
    from hermes_cli.moa_config import moa_usage, normalize_moa_config
except ImportError as exc:
    raise RuntimeError("MoA runtime unavailable (hermes-agent not installed or too old)") from exc
try:
    from hermes_cli.moa_config import resolve_moa_preset
except ImportError:
    resolve_moa_preset = None
# …then guard the call: selected = resolve_moa_preset(moa_raw, preset_name) if resolve_moa_preset else {}
```

## 2. Please split this into two PRs — the title and the diff don't match

The PR is titled "support WebUI client identity metadata," but roughly half of it is an unrelated change: a full rewrite of the Copilot model catalog (`_PROVIDER_MODELS["copilot"]`) plus a brand-new MoA ("Mixture of Agents") provider and model-picker (`api/config.py`, `api/commands.py`, the new `test_copilot_*`/`test_moa_*` files). Those are good changes, but they're a separate feature from client-identity metadata. Splitting them lets each be reviewed on its own merits — and it matters here specifically because the **client-identity half carries a security-boundary decision a maintainer needs to sign off on separately**: at `api/gateway_chat.py` the session-key header now takes the *client-supplied* `client_session_key` (and only `setdefault`s the server-scoped `webui:{session_id}`), so a client can override the server's own continuity scoping for the Gateway session. That's sanitized and explicitly framed as non-auth, but whether a client should be able to steer that scope is a deliberate trust call. Isolating it into its own PR makes that review clean.

## 3. Rebase needed

The branch currently conflicts with master in `api/gateway_chat.py` and `api/routes.py` (real content conflicts, master has advanced on both). A rebase onto current master is required.

To summarize the path forward: (1) fix the `resolve_moa_preset` import so the existing test passes and old-agent degradation is preserved; (2) split the Copilot/MoA work into its own PR with a matching title; (3) rebase. The client-identity feature itself isn't being rejected — once it's on its own branch it just needs a maintainer security review of the session-key override.

promptclickrun:
Thanks for the clear review. I pushed a rebased, identity-only update and split the model-picker work out.

What changed here:

- Rebased `feat/webui-client-identity-context` onto current `master`.
- Resolved the real conflicts in `api/gateway_chat.py` and `api/routes.py` while preserving the current upstream `goal_related` path.
- Removed the unrelated Copilot catalog / MoA provider files from this PR.
- Left this PR scoped to client identity metadata only:
  - `api/gateway_chat.py`
  - `api/routes.py`
  - `api/streaming.py`
  - `docs/advanced-chat-setup.md`
  - `tests/test_webui_client_identity_context.py`
  - `tests/test_webui_gateway_chat_backend.py`

The Copilot/MoA model-picker work is now split into #5301, including the requested `resolve_moa_preset` compatibility fix.

Verification for this PR:

```bash
./scripts/test.sh tests/test_webui_client_identity_context.py tests/test_webui_gateway_chat_backend.py tests/test_webui_surface_context.py tests/test_webui_platform_hint.py -q
```

Result:

```text
55 passed in 6.24s
```

I agree the remaining thing to review here is the deliberate trust decision around allowing a client-supplied `X-Hermes-Session-Key` / `client_identity.session_key` to steer gateway continuity scope. The PR keeps it sanitized and documented as context metadata, not auth; happy to adjust that behavior if you want server-owned continuity to always win.

nesquena-hermes:
## Review of the identity-only scope on `sha:7dbdd917` -- one security call and one functional regression on the gateway header path

Reading the rebased diff in a worktree (`api/gateway_chat.py`, `api/streaming.py`, `api/routes.py`) plus the agent-side contract in `gateway/platforms/api_server.py`, the sanitization work is solid and the in-process path is clean. The two things worth deciding both live in the gateway header-forwarding block, and they bear directly on the trust question you called out.

### The context-only path is fine

`_webui_system_prompt_with_client_identity` -> `_webui_client_identity_context` (`api/streaming.py:82-116`) only appends sanitized values to `workspace_system_msg`. Control chars/CRLF are stripped and length-capped, and nothing there scopes state. Greptile's 5/5 on sanitization is fair for that path.

### 1. Security -- forwarding client `session_key` as `X-Hermes-Session-Key` defeats an existing agent-side guard

The forwarded key is not just prompt context -- on the gateway it scopes **long-term memory** (Honcho / cross-transcript state). The agent side deliberately gates a caller-supplied session key behind API-key auth for exactly this reason (`gateway/platforms/api_server.py:1015-1021`):

```python
# Security: like session continuation, accepting a caller-supplied
# memory scope requires API-key authentication so that an
# unauthenticated client on a local-only server can't inject itself
# into another user's long-term memory scope by guessing a key.
```

The problem: WebUI multiplexes **every** browser turn through a single shared gateway Bearer key. From the gateway's vantage all WebUI users look like one authenticated principal, so that guard can't distinguish them. This PR then lets any browser client set an arbitrary scope verbatim (`api/gateway_chat.py:699-703`, `headers["X-Hermes-Session-Key"] = client_session_key`). In your own motivating multi-user scenario, user B can send `session_key: "team:person-a:ios"` and land in person A's long-term memory scope -- precisely the guess-and-inject the agent guard was written to stop, re-opened by the shared-Bearer model.

Recommendation (this is the "remaining trust decision" you flagged, and I'd resolve it before merge): don't forward a bare client key. Either
- **context-only** (never forward as the header) -- matches your own follow-up note and is the safe default; or
- **namespace it under the server-owned scope** so a client can only sub-scope within its own boundary, e.g. `f"webui:{session_id}:{client_session_key}"`, never a free-form key that can collide with a sibling's scope.

### 2. Functional regression -- the header block sits outside the `if api_key:` guard

The identity-header block runs unconditionally, before the auth block (`api/gateway_chat.py:692-707`):

```python
if isinstance(client_identity, dict):
    ...
    if client_session_key:
        headers["X-Hermes-Session-Key"] = client_session_key
if api_key:
    headers["Authorization"] = f"Bearer {api_key}"
    headers.setdefault("X-Hermes-Session-Key", f"webui:{session_id}")
```

Against an **unauthenticated** gateway (no `API_SERVER_KEY`, `api_key` is `None`), a browser supplying `session_key` still attaches `X-Hermes-Session-Key` -- but the gateway hard-rejects that header with **403** when no API key is configured (`gateway/platforms/api_server.py:1021-1032`). Net effect: sending identity metadata *breaks the chat turn* on an open local gateway that would otherwise have worked. Fix is to move the `X-Hermes-Session-Key` assignment inside the `if api_key:` branch (and combine with the namespacing above), so it's only sent on authenticated calls. `X-Hermes-Client-Name`/`Id` are harmless and can stay unconditional.

### Test plan

- Add a gateway-backend test with `api_key=None` + a client `session_key`, asserting `X-Hermes-Session-Key` is **absent** from the outgoing headers (guards regression #2).
- If you go with namespacing, assert the forwarded key is `webui:{session_id}:{sanitized_client_key}` rather than the raw client value (guards #1).

Everything else here reads clean -- header-over-body precedence, the `client_id` alias, and the CRLF-strip tests are all correct. The two items above are the gateway-path trust boundary, which is the part you explicitly asked maintainers to decide.

---
_Read-only worktree inspection of `sha:7dbdd917` + the agent-side contract at `gateway/platforms/api_server.py`; I do not run PR-authored code, merge, or tag._

promptclickrun:
Rebased onto current `master` (clean, no conflicts) and addressed both maintainer trust-decision items on `a74b4637`.

**1. Security — session key namespacing**

A client-supplied `session_key` is no longer forwarded to the Gateway verbatim. It's now namespaced under the server-owned scope: `webui:{session_id}:{client_session_key}`. This closes the guess-and-inject path you flagged: a client can sub-scope within its own conversation boundary, but can't collide with or steer into a sibling client's long-term memory scope, even though all WebUI traffic shares one Gateway Bearer key.

**2. Functional regression — unauthenticated gateway**

The `X-Hermes-Session-Key` assignment (namespaced or fallback) now lives entirely inside the `if api_key:` branch. On an unauthenticated gateway (no `API_SERVER_KEY`), the header is omitted entirely, even if the client supplied identity metadata, so it no longer trips the Gateway's 403 hard-reject and breaks chat turns on an open local gateway. `X-Hermes-Client-Name`/`Id` remain unconditional since they're harmless context-only headers.

**Tests added** (per your test plan), in `tests/test_webui_gateway_chat_backend.py`:
- `test_gateway_chat_worker_omits_session_key_header_when_gateway_unauthenticated` — asserts `X-Hermes-Session-Key` absent when `api_key=None` + client `session_key` present.
- `test_gateway_chat_worker_namespaces_client_session_key_under_server_scope` — asserts the forwarded key is `webui:{session_id}:{sanitized_client_key}`, never the raw client value.

Also updated the existing sanitization test's expected header value for the new namespaced format, and updated `docs/advanced-chat-setup.md` to describe the auth-gated, namespaced forwarding behavior.

**Verification:**

```bash
./scripts/test.sh tests/test_webui_client_identity_context.py tests/test_webui_gateway_chat_backend.py tests/test_webui_surface_context.py tests/test_webui_platform_hint.py -q
```

```text
57 passed in 4.57s
```

```bash
python3 -m py_compile api/gateway_chat.py api/routes.py api/streaming.py
```
clean.

Ran the full `test_update_banner_fixes.py` suite in isolation (92 passed) to confirm an unrelated flaky failure I saw under a broad `-k` shard wasn't caused by this change; it wasn't — passes identically with or without this diff.

promptclickrun:
Rebased onto current master (66851eac, clean, no conflicts). New head: `136f3fd1`. No code changes beyond the rebase, this is the same identity-only diff from the last comment (namespaced session-key forwarding, auth-gated header, split from #5301).

Re-verified after rebase:

```bash
./scripts/test.sh tests/test_webui_client_identity_context.py tests/test_webui_gateway_chat_backend.py tests/test_webui_surface_context.py tests/test_webui_platform_hint.py -q
```
```text
57 passed in 2.04s
```

`python3 -m py_compile api/gateway_chat.py api/routes.py api/streaming.py` clean.

Still waiting on a maintainer look at the security/functional fixes from the last review round (namespacing + auth-gating the forwarded session key).

promptclickrun:
Rebased onto current `master` (`56d1f619`, clean, no conflicts — includes the just-merged #5307 subagent-child-transcript work). New head: `a66b79c4`. No code changes beyond the rebase; same identity-only diff (namespaced session-key forwarding, auth-gated header) awaiting your re-review of the trust-boundary call.

**Verification:**

```bash
./scripts/test.sh tests/test_webui_client_identity_context.py tests/test_webui_gateway_chat_backend.py tests/test_webui_surface_context.py tests/test_webui_platform_hint.py -q
```
```text
57 passed in 2.39s
```

```bash
python3 -m py_compile api/gateway_chat.py api/routes.py api/streaming.py
```
clean.

Also re-ran the #5307 subagent-gating tests your rebase base just picked up, to confirm no interaction with the touched files:

```bash
./scripts/test.sh tests/test_5307_subagent_child_transcript.py tests/test_regressions.py tests/test_session_duplicate.py -q
```
```text
86 passed in 3.23s
```

promptclickrun:
Hardened the remaining Greptile 4/5 gap on `c8658a3a`.

What changed:

- Added one shared `_gateway_client_identity_headers(...)` helper so legacy chat-completions and Runs API cannot drift on identity forwarding again.
- Threaded `client_identity` into `_run_gateway_runs_api_streaming` from the `_use_runs_api` branch.
- Runs API now forwards sanitized `X-Hermes-Client-Name` / `X-Hermes-Client-Id` on both `POST /v1/runs` and `GET /v1/runs/{run_id}/events`.
- Runs API now applies the same auth-gated session-key behavior as the legacy path:
  - authenticated gateway: `webui:{session_id}:{sanitized_client_session_key}`
  - no client session key: `webui:{session_id}`
  - unauthenticated gateway: omit `X-Hermes-Session-Key` entirely

Tests added:

- Outer worker assertion that `client_identity` is actually passed into the Runs API helper.
- Runs API header regression asserting sanitized client name/id plus namespaced session key on both the initial run request and event stream request.
- Runs API unauthenticated-gateway regression asserting `X-Hermes-Session-Key` is omitted while context-only client headers still forward.

Local verification:

```bash
./scripts/test.sh tests/test_gateway_approval_runs_api.py::test_gateway_runs_api_submission tests/test_gateway_approval_runs_api.py::test_gateway_runs_api_streaming_forwards_sanitized_client_identity_headers tests/test_gateway_approval_runs_api.py::test_gateway_runs_api_streaming_omits_session_key_header_when_gateway_unauthenticated -q
# 3 passed in 3.54s

./scripts/test.sh tests/test_gateway_approval_runs_api.py tests/test_webui_gateway_chat_backend.py tests/test_webui_client_identity_context.py -q
# 63 passed in 7.65s

./scripts/test.sh tests/test_webui_client_identity_context.py tests/test_webui_gateway_chat_backend.py tests/test_webui_surface_context.py tests/test_webui_platform_hint.py -q
# 57 passed in 2.37s

./scripts/test.sh tests/test_gateway_approval_runs_api.py tests/test_gateway_approval_legacy_path.py tests/test_context_history_sanitization.py tests/test_webui_gateway_chat_backend.py tests/test_webui_client_identity_context.py tests/test_webui_surface_context.py tests/test_webui_platform_hint.py -q
# 101 passed in 7.91s

python3 -m py_compile api/gateway_chat.py api/routes.py api/streaming.py tests/test_gateway_approval_runs_api.py tests/test_webui_gateway_chat_backend.py tests/test_webui_client_identity_context.py
# clean

git diff --check
# clean
```

Remote verification: local and `promptclickrun/feat/webui-client-identity-context` both resolve to `c8658a3aebff48effa59957ec585f2978873d23e`.

Greptile is currently rerunning on the new SHA.

promptclickrun:
Follow-up verification for current head `f96fc97e` (test-only commit on top of the `c8658a3a` runs-API forwarding change already summarized above).

**What `f96fc97e` adds:** `test_gateway_runs_api_streaming_uses_server_scope_without_client_session_key` in `tests/test_gateway_approval_runs_api.py` — asserts that when no `client_session_key` is supplied, the Runs API path falls back to the server-owned `webui:{session_id}` scope (parity with the chat-completions default-scope regression test). No product code changed since `c8658a3a`.

**Fresh verification on `f96fc97e`:**
```bash
./scripts/test.sh tests/test_gateway_approval_runs_api.py tests/test_webui_client_identity_context.py tests/test_webui_gateway_chat_backend.py -q
```
```text
64 passed in 7.53s
```
```bash
./scripts/test.sh tests/test_gateway_approval_runs_api.py -k test_gateway_runs_api_streaming_uses_server_scope_without_client_session_key -q
```
```text
1 passed, 17 deselected in 1.65s
```
```bash
python3 -m py_compile tests/test_gateway_approval_runs_api.py
```
clean.

Branch is `diverged` from `master` (behind_by 40) but `mergeable` with no conflicts. Greptile is 5/5 ("safe to merge"). Still awaiting your re-review of the session-key trust-boundary call.

promptclickrun:
Fixed the failing `lint` gate on `498a2c03`.

**What was red:** the Ruff forward gate flagged one new violation:

```
api/gateway_chat.py:619:17  F401  `api.streaming._webui_client_identity_context` imported but unused
```

The gateway chat path builds the identity-aware system prompt via `_webui_system_prompt_with_client_identity` and never calls `_webui_client_identity_context` directly, so that import was dead. Removed it. No behavior change — pure lint fix, layered as a normal commit on top of the previous head (no force-push, no history rewrite).

**Verification:**

```bash
```
```text
ruff_lint (diff vs upstream/master): 6 changed .py file(s), 27 total finding(s) in them, 0 on added/modified lines.
ruff_lint: no new violations on added/modified lines. OK.
```

```bash
./scripts/test.sh tests/test_webui_client_identity_context.py tests/test_webui_gateway_chat_backend.py tests/test_webui_surface_context.py tests/test_webui_platform_hint.py tests/test_gateway_approval_runs_api.py -q
```
```text
75 passed
```

```bash
python3 -m py_compile api/gateway_chat.py api/routes.py api/streaming.py   # clean
git diff --check                                                          # clean
```

Also confirmed the original MoA regression stays green (`tests/test_issue5057_moa_webui_route.py`: 7 passed, 1 skipped).

Still awaiting your re-review of the session-key trust-boundary call (namespaced + auth-gated forwarding) from the last round; Greptile remains 5/5.

nesquena-hermes:
## Closing — client-supplied identity + gateway-scope override is out of scope

Thanks @promptclickrun — the implementation is careful (the sanitization is genuinely solid: control-char stripping, no CRLF/header injection, length caps). Closing this on a scope/security-boundary call, not a code-quality one.

Two things put it out of scope:

1. **Multi-user framing.** The feature's own rationale is "multiple people... who is speaking... `team:person-a`." Hermes is single-user / multi-*profile*, not multi-*tenant* — the same line we hold elsewhere. Identifying distinct speakers to the agent is a multi-user concept.

2. **Client-steerable gateway scope is a security boundary.** The bigger issue is that a client-supplied `session_key` (via `X-Hermes-Client-Session-Key` / `client_identity`) can **override the previously server-locked gateway-memory scope**, and it runs unconditionally with no server opt-out. Which gateway-memory bucket a session reads and writes is a server-owned boundary — letting the client steer it is a change we're not comfortable making on by default.

To be fair to the design: normal single-user browser behavior is unchanged (the browser never sends these headers by default — it's purely additive for custom clients/proxies), so it isn't harmful out of the box. If there were a narrow, clearly-scoped **single-operator** use (one operator's own proxy tagging its own turns, with the server retaining the authoritative scope and an explicit opt-in), that's a different and more defensible shape than what's here. But as written — client-overrides-server-scope, unconditional, multi-user-framed — it's out of scope.

Core-or-nothing (gateway forwarding + system-prompt injection can't be an extension), and there's no atomic piece to salvage (the sanitization only serves the feature). Closing with thanks — the sanitization work especially was well done, and operator-hardening of one's own instance (e.g. the trusted-header SSO work in #5568) is an area where that care would be very welcome. Please keep contributing.

PR #5246, opened 2026-06-30
fix(model): repair bare custom-provider suffix on resolver slow path (#5225)

### LINKED ISSUE #5225: Bug: Server-side continuations can call custom providers with a bare model suffix
## Summary

A WebUI session using a custom provider model with a vendor-qualified ID can later fail during a server/runtime continuation with the bare model suffix:

```text
Model not found: HTTP 400: ... Invalid model format or no credentials for provider: grok-composer-2.5-fast
```

The initial WebUI session and sidecar state used the vendor-qualified model form (`x-ai/grok-composer-2.5-fast`) and a matching custom provider context. The user only submitted one message; the failure arrived much later after many assistant/tool steps, suggesting a continuation path reconstructed or forwarded model state differently from the initial user-submitted turn.

## Environment

- WebUI: v0.51.742
- Agent: 0.17.0
- Provider shape: named custom provider
- Profile default model shape: vendor-qualified ID, e.g. `x-ai/grok-composer-2.5-fast`

## Observed behavior

Two incidents showed the same terminal error class after substantial work:

- selected/persisted model shape was vendor-qualified (`x-ai/grok-composer-2.5-fast`);
- persisted model provider context was present and matched the profile provider;
- custom provider entry was present with endpoint and credential configured;
- terminal error referenced bare `grok-composer-2.5-fast`;
- the user had not made a second model selection or second manual submission.

This does not appear to be an obvious local misconfiguration: the profile had a named custom provider, the default model was vendor-qualified, and affected session sidecars persisted both `model` and `model_provider`.

## Expected behavior

Every server-initiated continuation/wakeup for a WebUI session should use the same effective model/provider routing as the user-submitted turn:

- preserve `model_provider` context from session/profile state;
- preserve or repair the vendor-qualified custom-provider model ID;
- avoid sending only the bare suffix to a provider endpoint that requires the vendor-qualified model ID;
- if provider/model context cannot be reconstructed, fail before the provider call with a diagnostic that shows the session model/provider and resolved model/provider.

## Code context

The obvious process-wakeup repair appears to exist in v0.51.742. The tag contains `tests/test_issue5127_process_wakeup_bare_model.py`, which validates this resolver shape:

```python
_resolve_compatible_session_model_state(
    "grok-composer-2.5-fast",
    "custom:example",
    profile_provider="custom:example",
    profile_default_model="x-ai/grok-composer-2.5-fast",
    prefer_cached_catalog=True,
)
# -> ("x-ai/grok-composer-2.5-fast", "custom:example", True)
```

Local validation of current code passed:

```text
./scripts/test.sh tests/test_issue5127_process_wakeup_bare_model.py -q
Running 8 items in this shard
........                                                                 [100%]
8 passed in 5.17s
```

So recurrence on v0.51.742 suggests that the existing regression does not cover at least one live state/path shape.

## Possible uncovered paths

- A server-side continuation path bypasses `start_session_turn()` / `_resolve_compatible_session_model_state()`.
- `model_provider` is present but normalized or compared differently before profile defaults are read.
- The model is resolved correctly and then stripped from vendor-qualified to bare before the outbound provider request.
- A hydration/runtime handoff path re-bares the model after the first turn.

## Suggested diagnostics / acceptance criteria

- Add structured diagnostics on server-initiated turns recording original session `model`, session `model_provider`, profile provider, profile default model, resolved model, resolved provider, and whether `normalized_model` was true.
- Add an end-to-end-ish regression for a server-initiated continuation where the persisted session has a vendor-qualified model and matching custom provider context.
- Add a sibling regression for the concrete named-custom-provider shape: assert the route keeps or reconstructs the vendor-qualified model and never calls the provider with only the bare suffix.
- Verify all server-started continuation paths (process wakeup, deferred wakeup, queue/goal continuation, async/delegation completion if applicable) share the same resolver and outbound model value.

claw-io:
Fixes #5225.

## Thinking Path

- Read maintainer review on #5225: continuations reach `_resolve_compatible_session_model_state`, but when `requested_provider` is omitted (common after async `delegate_task` / `process_wakeup`), the resolver takes the **slow path**, which only repairs bare `gpt`/`claude`/`gemini` families—not a bare custom-provider suffix like `grok-composer-2.5-fast`.
- Agreed with the recommended fix: mirror the #5128 fast-path suffix→qualified repair in the profile-provider slow branch, gated on custom provider + exact suffix match to `profile_default_model`.
- Implemented in `api/routes.py` and extended `tests/test_issue5127_process_wakeup_bare_model.py` with slow-path cases (including mismatch guard).

## What Changed

- **`api/routes.py`:** Before the slow-path family repair, when profile context is present and the session model is a bare suffix matching the profile’s slash-qualified custom default, return `_profile_default` with `normalized=True` (#5225).
- **`tests/test_issue5127_process_wakeup_bare_model.py`:** `test_slow_path_repairs_bare_suffix_to_profile_default_for_custom_provider` (`requested_provider=None`); `test_slow_path_does_not_rewrite_unrelated_bare_model_for_custom_provider`.

No `CHANGELOG.md` edits.

## Why It Matters

Server-initiated continuations (especially delegation completion re-entry) could call the custom provider with only the bare model ID, producing `Invalid model format or no credentials for provider: grok-composer-2.5-fast` even when the session/profile default was `x-ai/grok-composer-2.5-fast`.

## Verification

```bash
./scripts/test.sh tests/test_issue5127_process_wakeup_bare_model.py -q
```

10 passed locally. `/opt/data/workspace/developer/scripts/hermes-webui-pr-gate <worktree> --with-tests "tests/test_issue5127_process_wakeup_bare_model.py -q"`: pass (committed-diff ruff, co-author trailer).

**Manual:** Resolver unit tests only; no browser UI change.

## Risks / Follow-ups

- Repair only fires when bare suffix exactly matches profile default suffix and provider is `custom` / `custom:*`; cross-custom-provider mismatch behavior unchanged (#5127 guard).
- End-to-end WebUI delegation injection test could be a follow-up if maintainers want stronger coverage than resolver-level regression.

## Release note (for maintainers)

**Fixed:** Bare custom-provider model suffixes are repaired to the profile’s vendor-qualified default on the resolver slow path when `model_provider` is omitted during server-side continuations (#5225).

## Model Used

- Provider: custom (x-ai via proxy)
- Model: x-ai/grok-composer-2.5-fast
- AI disclosure: Hermes developer agent assisted implementation and tests; human (Ben) directed push/PR.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5578,6 +5578,24 @@ def _resolve_compatible_session_model_state(
             _target = _profile_default or default_model
             return _target, profile_provider, True
 
+        # Async server-side continuations (for example delegate_task completion
+        # re-entry) can arrive here with profile context but without a usable
+        # requested_provider, bypassing the fast-path custom-provider repair
+        # above. If the profile's configured custom-provider default is a
+        # slash-qualified model whose suffix matches the bare session model,
+        # repair back to the profile default before the provider call (#5225).
+        if (
+            "/" not in model
+            and _profile_default
+            and "/" in _profile_default
+            and _profile_default.rsplit("/", 1)[-1] == model
+            and (
+                _profile_provider_normalized == "custom"
+                or str(profile_provider).startswith("custom:")
+            )
+        ):
+            return _profile_default, profile_provider, True
+
         return model, profile_provider, False
 
     if not model:
--- a/tests/test_issue5127_process_wakeup_bare_model.py
+++ b/tests/test_issue5127_process_wakeup_bare_model.py
@@ -82,6 +82,47 @@ class TestIssue5127CustomProviderBareSuffixRepair:
         assert mock_catalog.call_count == 0
         assert result == ("x-ai/grok-composer-2.5-fast", "custom:my-proxy", False)
 
+    def test_slow_path_repairs_bare_suffix_to_profile_default_for_custom_provider(self):
+        """Regression #5225: async continuations may arrive without usable model_provider."""
+        from api.routes import _resolve_compatible_session_model_state
+
+        with patch("api.routes.get_available_models") as mock_catalog:
+            mock_catalog.return_value = {
+                "active_provider": "openrouter",
+                "default_model": "openai/gpt-5.5",
+                "groups": [],
+            }
+            result = _resolve_compatible_session_model_state(
+                "grok-composer-2.5-fast",
+                None,
+                profile_provider="custom:my-proxy",
+                profile_default_model="x-ai/grok-composer-2.5-fast",
+                prefer_cached_catalog=True,
+            )
+
+        assert mock_catalog.call_count == 1
+        assert result == ("x-ai/grok-composer-2.5-fast", "custom:my-proxy", True)
+
+    def test_slow_path_does_not_rewrite_unrelated_bare_model_for_custom_provider(self):
+        from api.routes import _resolve_compatible_session_model_state
+
+        with patch("api.routes.get_available_models") as mock_catalog:
+            mock_catalog.return_value = {
+                "active_provider": "openrouter",
+                "default_model": "openai/gpt-5.5",
+                "groups": [],
+            }
+            result = _resolve_compatible_session_model_state(
+                "other-model",
+                None,
+                profile_provider="custom:my-proxy",
+                profile_default_model="x-ai/grok-composer-2.5-fast",
+                prefer_cached_catalog=True,
+            )
+
+        assert mock_catalog.call_count == 1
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.759** (via #5247). Thanks @claw-io — the resolver slow path (async delegate_task/process-wakeup continuations) now repairs a bare model back to the profile's custom-provider default, mirroring the fast path. Codex SAFE, full suite green. Closes #5225.

PR #1951, opened 2026-05-09
fix: only evaluate goal hook on goal-related turns (#1932)

amlyczz:
## Summary

- Fixes #1932 where the goal evaluation hook was firing on **every** completed assistant turn when a goal was active, even for unrelated messages like "what time is it"
- This burned the goal budget on unrelated turns, triggered continuation prompts that interrupted unrelated conversations, and made /goal status numbers misleading
- Adds STREAM_GOAL_RELATED and PENDING_GOAL_CONTINUATION flags to gate the evaluate_goal_after_turn() call in the streaming loop

## Changes

**api/config.py** -- Added two new data structures:
- STREAM_GOAL_RELATED dict (stream_id to bool): marks specific streams as goal-related so only they trigger the goal evaluation hook
- PENDING_GOAL_CONTINUATION set (session_ids): auto-marks the next stream for a session as goal-related when a goal_continue event fires

**api/streaming.py** -- Gates the goal evaluation hook:
- _run_agent_streaming() accepts goal_related=False kwarg
- The goal evaluation section now checks if not goal_related and skips evaluation for unrelated turns
- When goal_continue fires, adds the session to PENDING_GOAL_CONTINUATION so the next stream is automatically marked
- Cleans up both dicts in the stream exit path

**api/routes.py** -- Propagates the goal-related flag:
- _start_chat_stream_for_session() accepts goal_related=False kwarg
- Checks PENDING_GOAL_CONTINUATION for the session and auto-marks as goal-related
- Sets STREAM_GOAL_RELATED[stream_id] for goal-related streams
- Passes goal_related=True through to _run_agent_streaming
- /api/goal kickoff path passes goal_related=True

## Test plan

- [x] 10 new regression tests in test_issue_1932_goal_hook_unrelated_turns.py
- [x] All 10 existing goal tests pass (test_goal_command_webui.py)
- [x] All 298 streaming-related tests pass
- [x] Full suite: 4962 passed, 0 new failures (13 pre-existing failures unrelated to this change)

PR #5251, opened 2026-06-30
fix(#5092): continue WebUI goals after gateway turns

### LINKED ISSUE #5092: [Bug] `/goal` loop does not continue in WebUI (APIServerAdapter) — `_post_turn_goal_continuation` not wired into `_run_agent`
## Summary

The `/goal` feature works correctly in CLI/TUI and gateway platforms (Telegram, Discord, WeChat) but is non-functional in WebUI. The goal state is correctly set via `/goal <text>`, but the continuation loop never fires — `turns_used` stays at 0 and the agent never takes a second turn toward the goal.

## Steps to Reproduce

1. Open WebUI session
2. Send `/goal Write a test file at /tmp/goal-test.txt`
3. Hermes responds with the first turn (e.g. creates the file)
4. Goal loop does NOT continue — no further turns are taken
5. `/goal status` shows `0/20 turns` indefinitely

## Expected Behavior

After each turn completes in WebUI, the goal judge should evaluate whether the goal is satisfied. If not, a continuation prompt should be enqueued automatically — the same behavior as CLI/TUI and gateway platforms.

## Actual Behavior

Goal state is set to `active` in `state_meta`, but `_post_turn_goal_continuation()` is never called. The goal loop only works in `_handle_message_with_agent()` (gateway platforms), not in `APIServerAdapter._run_agent()` which serves WebUI requests.

## Root Cause

- `_post_turn_goal_continuation()` is defined in `gateway/run.py` and called from `_handle_message_with_agent()` (line ~8392)
- WebUI requests go through `APIServerAdapter._handle_session_chat()` → `_run_agent()` (in `gateway/platforms/api_server.py`, line 3571)
- `_run_agent()` calls `agent.run_conversation()` and returns directly — no goal continuation logic is invoked
- `api_server.py` contains zero references to `GoalManager`, `_post_turn_goal_continuation`, or any goal-related code

## Environment

- Hermes version: 0.17.0
- Platform: WebUI (APIServerAdapter / api_server.py)
- Config: `goals.max_turns: 20`, `goal_judge` routed to default model

## Proposed Fix

Wire goal continuation into `APIServerAdapter._run_agent()` or `_handle_session_chat()`:

1. After `agent.run_conversation()` returns, instantiate `GoalManager` for the session
2. Call `mgr.evaluate_after_turn(final_response)` if a goal is active
3. If `decision["should_continue"]`, enqueue the continuation prompt back through the adapter's FIFO queue (same pattern as `_post_turn_goal_continuation` in `run.py`)

Alternatively, extract the goal continuation logic from `_post_turn_goal_continuation` into a shared utility that both `_handle_message_with_agent` and `_run_agent` can call.

## Workaround

Users can manually copy the continuation prompt from `GoalManager.next_continuation_prompt()` and paste it into WebUI after each turn. This is functional but defeats the purpose of an autonomous goal loop.

## References

- Goal continuation in gateway: `gateway/run.py` lines 10495-10562 (`_post_turn_goal_continuation`)
- WebUI agent runner: `gateway/platforms/api_server.py` lines 3571-3639 (`_run_agent`)
- GoalManager: `hermes_cli/goals.py` line 509

rodboev:
## Thinking Path

- The browser-side `/goal` loop already knows how to continue after a turn, but only when the server worker emits the existing `goal_continue` contract.
- The live gap is not in browser code or goal state ownership, it is in the gateway-backed settle path, where the final assistant turn is persisted and the stream ends before the goal judge runs.
- The fix therefore keeps the current ownership split: routes forwards the goal-related signal symmetrically, the gateway worker evaluates the goal after settling the turn, and the browser keeps using the same continuation event it already handles.

## What Changed

- `api/routes.py`: forward `goal_related` to the gateway-backed worker, matching the direct path.
- `api/gateway_chat.py`: evaluate the active goal after a goal-related gateway turn settles, set `PENDING_GOAL_CONTINUATION`, and emit the existing `goal` / `goal_continue` events before `done` and `stream_end`.
- `tests/test_webui_gateway_chat_backend.py`: add the focused gateway continuation regression and the non-goal negative-boundary regression.
- `tests/test_goal_command_webui.py`: add route-level coverage that proves the gateway worker receives `goal_related=True`.

## Why It Matters

With the gateway backend selected, `/goal` should keep taking turns until the goal judge stops it, the same way the direct WebUI path already does. This restores that behavior without adding a new queue, moving goal ownership into the adapter layer, or changing the browser-side continuation contract.

## Verification

- `pytest tests/test_webui_gateway_chat_backend.py tests/test_goal_command_webui.py tests/test_stage326_pending_goal_continuation_race.py -v --timeout=60`

## Upstream

Closes #5092.

Issue-thread scope and queue-boundary guidance came from https://github.com/nesquena/hermes-webui/issues/5092#issuecomment-4825249360.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -12,6 +12,8 @@ from typing import Any
 
 from api.config import (
     CANCEL_FLAGS,
+    PENDING_GOAL_CONTINUATION,
+    STREAM_GOAL_RELATED,
     STREAMS,
     STREAMS_LOCK,
     STREAM_LAST_EVENT_ID,
@@ -502,6 +504,7 @@ def _run_gateway_chat_streaming(
     attachments=None,
     *,
     model_provider=None,
+    goal_related=False,
 ):
     """Bridge a WebUI chat turn through Hermes Gateway's API server.
 
@@ -882,6 +885,55 @@ def _run_gateway_chat_streaming(
             s.model = model
             s.model_provider = model_provider
             s.save()
+        try:
+            from api.goals import evaluate_goal_after_turn, has_active_goal
+            from api.profiles import get_hermes_home_for_profile
+
+            profile_home = get_hermes_home_for_profile(getattr(s, "profile", None))
+            if goal_related and has_active_goal(session_id, profile_home=profile_home):
+                put_gateway_event("goal", {
+                    "session_id": session_id,
+                    "state": "evaluating",
+                    "message": "Evaluating goal progress…",
+                    "message_key": "goal_evaluating_progress",
+                })
+                decision = evaluate_goal_after_turn(
+                    session_id,
+                    assistant_text,
+                    user_initiated=True,
+                    profile_home=profile_home,
+                ) or {}
+                goal_message = str(decision.get("message") or "").strip()
+                if goal_message:
+                    put_gateway_event("goal", {
+                        "session_id": session_id,
+                        "state": "continuing" if decision.get("should_continue") else "idle",
+                        "message": goal_message,
+                        "message_key": decision.get("message_key") or (
+                            "goal_continuing" if goal_message else ""
+                        ),
+                        "message_args": decision.get("message_args") or [],
+                        "decision": decision,
+                    })
+                if decision.get("should_continue"):
+                    continuation_prompt = str(decision.get("continuation_prompt") or "").strip()
+                    if continuation_prompt:
+                        PENDING_GOAL_CONTINUATION.add(session_id)
+                        put_gateway_event("goal_continue", {
+                            "session_id": session_id,
+                            "continuation_prompt": continuation_prompt,
+                            "text": continuation_prompt,
+                            "message": goal_message,
+                            "message_key": decision.get("message_key") or "goal_continuing",
+                            "message_args": decision.get("message_args") or [],
+                            "decision": decision,
+                        })
+        except Exception as goal_exc:
+            logger.debug(
+                "Gateway goal continuation hook failed for session %s: %s",
+                session_id,
+                goal_exc,
+            )
         from api.streaming import _session_payload_with_full_messages
         gateway_session_payload = _session_payload_with_full_messages(s, tool_calls=[])
         put_gateway_event("done", {"session": redact_session_data(gateway_session_payload), "usage": usage})
@@ -913,6 +965,7 @@ def _run_gateway_chat_streaming(
             _cleanup_gateway_pending_mirror(session_id)
         with STREAMS_LOCK:
             CANCEL_FLAGS.pop(stream_id, None)
+            STREAM_GOAL_RELATED.pop(stream_id, None)
             STREAM_PARTIAL_TEXT.pop(stream_id, None)
             STREAM_REASONING_TEXT.pop(stream_id, None)
             STREAM_LIVE_TOOL_CALLS.pop(stream_id, None)
--- a/api/routes.py
+++ b/api/routes.py
@@ -17489,9 +17489,7 @@ def _start_chat_stream_for_session(
     diag.stage("worker_thread_start") if diag else None
     backend_is_gateway = webui_gateway_chat_enabled(get_config())
     worker_target = _run_gateway_chat_streaming if backend_is_gateway else _run_agent_streaming
-    worker_kwargs = {"model_provider": model_provider}
-    if not backend_is_gateway:
-        worker_kwargs["goal_related"] = goal_related
+    worker_kwargs = {"model_provider": model_provider, "goal_related": goal_related}
     if moa_config and not backend_is_gateway:
         worker_kwargs["moa_config"] = moa_config
     thr = threading.Thread(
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Read this end-to-end against `origin/master`: the new gateway hook in `api/gateway_chat.py:885-934`, the existing direct-path equivalent in `api/streaming.py:9013-9074`, the `goal_related` plumbing in `api/routes.py:17502-17630`, and both new test files. This is a clean, contract-faithful port — the gateway path now reaches goal parity with the direct WebUI path.

## What lines up

The gateway hook mirrors the direct path almost line-for-line, including the correct ordering invariant from #1951: the marker is added by the producer and the consumer discards it, never the `finally`.

- The hook fires **after** `s.save()` (the persisted turn) at `api/gateway_chat.py:885`, matching the direct path which judges after the session writeback. The judge sees the settled turn.
- `PENDING_GOAL_CONTINUATION.add(session_id)` is the only mutation in `gateway_chat.py` (line 921); there is no `.discard` in its cleanup `finally` (line 967 only pops `STREAM_GOAL_RELATED`). That is exactly right — the discard lives consumer-side in `routes.py:17528-17530`:

```python
if not goal_related and s.session_id in PENDING_GOAL_CONTINUATION:
    goal_related = True
    PENDING_GOAL_CONTINUATION.discard(s.session_id)
```

  This is the precise behavior the stage-326 regression guards (`test_stage326_pending_goal_continuation_race.py:49`) protect on the streaming side. The gateway path replicates it correctly.

## One real divergence (benign, worth a sentence in the code)

The direct path walks `s.messages` in reverse to recover the last assistant text before judging (`api/streaming.py:9020-9034`), whereas the gateway path feeds `assistant_text` directly (`api/gateway_chat.py:902`). That is actually cleaner — `assistant_text` is the same value just persisted into the assistant row at `api/gateway_chat.py:837` (`content=assistant_text`), so both paths evaluate identical input. No change needed; a one-line comment noting "gateway path already has the settled assistant text, no message walk required" would save the next reader the diff.

## Agreeing with the two Greptile findings

Both are valid and both are non-blocking:

1. **Stale comment at `routes.py:17621-17622`** ("so the streaming hook evaluates the goal"). Now that `STREAM_GOAL_RELATED[stream_id]` is set for gateway streams too but the gateway worker reads the `goal_related` kwarg directly (never the dict), the comment is misleading. Worth a tweak to "...so the active streaming worker (direct or gateway) evaluates the goal."

2. **stage-326 guard only scans `streaming.py`** (`test_stage326_pending_goal_continuation_race.py:25` reads `api/streaming.py` only). Since you have now duplicated the producer/consumer marker pattern into `gateway_chat.py`, the same race could silently regress there without tripping the guard. Extending the guard to also assert `"PENDING_GOAL_CONTINUATION.discard" not in gateway_chat_source` would close that gap cheaply. This is the highest-value follow-up in the review.

## Tests

The two new gateway regressions are well-targeted — `test_gateway_chat_worker_emits_goal_continue_for_goal_related_turn` asserts both the event ordering (`event_names.index("goal_continue") < event_names.index("done")`, line 562) and the marker membership (`s.session_id in PENDING_GOAL_CONTINUATION`, line 564), and the negative boundary test confirms no `goal_continue` when `goal_related=False` (line 639). The route-level test proving the kwarg reaches `_run_gateway_chat_streaming` (`test_goal_command_webui.py:431`) closes the plumbing loop. CI is green across the full 3.11/3.12/3.13 matrix.

## Verdict

LGTM. The two Greptile nits (stale comment + extending the stage-326 guard to cover `gateway_chat.py`) are worth folding in but neither blocks merge. The core change is correct and matches the agent-side contract described in #5092.

nesquena-hermes:
Shipped in **v0.51.764** — thank you @rodboev. 🎯

Merged via a maintainer rebase of your branch onto current `master` (release PR #5267) with your authorship preserved. Closing manually since the release landed through the rebased branch; the linked issue #5092 auto-closed.

**Full authoritative gate before ship:**
- **Codex regression gate: SAFE TO SHIP** — verified the local evaluation is gated on `goal_related && has_active_goal` (ordinary gateway turns are a clean no-op), the `PENDING_GOAL_CONTINUATION` marker is set before emitting `goal_continue`, stream-scoped goal state is cleaned up in the `finally`, and the browser's `goal_continue` handling is the existing queue/chat-start path. 67 goal-related tests pass.
- **Full pytest suite: 11111 passed** (only the known `test_scheduled_jobs_profile_isolation` serial-ordering artifact — passes in isolation + sharded CI; this touches no cron code).
- Your tests (`test_goal_command_webui.py` + `test_webui_gateway_chat_backend.py`): 52/52.

This was the right architecture: evaluating the goal **locally in the WebUI worker** after the gateway turn settles (mirroring the proven non-gateway `evaluate_goal_after_turn` path) rather than depending on the remote gateway to emit a continuation — which is exactly why it works where the earlier relay attempts (#5077 / #5153) didn't. Nice.

Issue #390, opened 2026-04-14
Onboarding blocked for reverse proxy / SSH tunnel users — X-Forwarded-For not respected, chicken-and-egg with auth


nesquena-hermes:
## Summary

The onboarding setup endpoint (`/api/onboarding/setup`) blocks requests from non-local IPs when auth is not enabled. This is a reasonable security intent, but two implementation problems make it trigger incorrectly for legitimate users:

1. **Reverse proxy / SSH tunnel IP**: The check uses `handler.client_address[0]` (the raw TCP connection IP) and does not read `X-Forwarded-For` or `X-Real-IP`. Users running Hermes behind nginx, Caddy, Traefik, or accessing via an SSH tunnel get a non-loopback client IP even though they are connecting from localhost. Result: the onboarding page is permanently blocked for any standard reverse-proxy deployment.

2. **Chicken-and-egg with auth**: The restriction is bypassed when auth is enabled (`is_auth_enabled() → True`). But auth is enabled by setting a password — which you'd normally do during onboarding or via the Settings panel after onboarding. A fresh install with no password set hits the block immediately, with no path forward. Setting a password via the Settings panel requires completing onboarding first. The workaround doesn't exist.

The error message shown to the user:
> "Onboarding setup is only available from local networks when auth is not enabled."

## Steps to reproduce

**Reverse proxy case:**
1. Run Hermes WebUI behind nginx or Caddy with a standard proxy_pass config
2. Access `http://your-domain/` in a browser — the onboarding wizard loads
3. Fill in provider credentials and submit
4. Receive 403: "Onboarding setup is only available from local networks when auth is not enabled."
5. No way to proceed — the onboarding page offers no alternative path

**SSH tunnel case:**
1. Run Hermes WebUI on a VPS (e.g. port 8787)
2. Connect via `ssh -L 8787:127.0.0.1:8787 user@vps` and open `http://127.0.0.1:8787` locally
3. Same 403 — the server sees the tunnel endpoint IP, not 127.0.0.1

## Root cause

```python
# api/routes.py ~line 918
if not is_auth_enabled():
    addr = ipaddress.ip_address(handler.client_address[0])
    is_local = addr.is_loopback or addr.is_private
    if not is_local:
        return bad(handler, "Onboarding setup is only available from local networks...", 403)
```

- `handler.client_address[0]` is the raw TCP socket peer address — it does not reflect `X-Forwarded-For`
[...]

PR #5268, opened 2026-06-30
fix(security): gate embedded-terminal endpoints to local origins when auth disabled

nesquena-hermes:
## Summary

Gate the embedded-terminal HTTP endpoints to **local-origin requests when authentication is not configured**, closing an unauthenticated-RCE path on a passwordless network-exposed deployment.

The embedded workspace terminal spawns a PTY shell that runs arbitrary commands as the server-process user. When a password/passkey is configured, the existing session-cookie auth already gates these endpoints. But `check_auth()` returns `True` unconditionally when no auth is configured (the default out-of-the-box state), so in that state every terminal endpoint was reachable by any caller able to hit the port — which, on a server bound to a non-loopback address with no auth, means a remote unauthenticated caller could drive a shell.

This is the same trust-model gap the onboarding/bootstrap endpoints already close (`#3758`/`#390` lineage). This PR reuses that exact gate.

## What changed

- New `_embedded_terminal_gate_allows(handler)` helper in `api/routes.py`, delegating to the existing `_onboarding_gate_allows` (which ignores spoofable `X-Forwarded-For`/`X-Real-IP` unless `HERMES_WEBUI_TRUST_FORWARDED_FOR=1`).
- Applied to all five embedded-terminal endpoints, **before any side-effecting work** (session lookup / PTY spawn / write / resize / close / SSE stream open):
  - `POST /api/terminal/start`
  - `POST /api/terminal/input`
  - `POST /api/terminal/resize`
  - `POST /api/terminal/close`
  - `GET /api/terminal/output`
- Behavior:
  - **Auth enabled** → unaffected (cookie already verified upstream by `check_auth`).
  - **Auth disabled** → only loopback/private origins admitted; public-origin requests get `403`.
  - Operator who secures access another layer (reverse proxy auth / VPN / private net) opts out with `HERMES_WEBUI_ONBOARDING_OPEN=1`.
- Genuine same-host browser clients (the normal case) are unaffected.

## Tests

- New `tests/test_cvd3_terminal_local_origin_gate.py` (12 cases): gate predicate (public blocked / loopback + private allowed / spoofed-header rejected / auth-enabled allowed / `ONBOARDING_OPEN` escape hatch) + handler dispatch (each of the five endpoints returns `403` for a public client **without** reaching the spawn/write path, and a loopback client clears the gate and proceeds to normal lookup).
- Fixed a latent test-isolation leak in `test_extension_route_remains_behind_webui_auth` (it set `HERMES_WEBUI_PASSWORD` but never invalidated the process-wide password-hash cache, so its result depended on suite execution order — exposed when the new test file shifted ordering). Now invalidates the cache before+after so it reads the env var deterministically.
- Made the `_RouteHandler` test fixture in `test_embedded_workspace_terminal.py` present a loopback `client_address` (real terminal requests are same-host), so those tests exercise the terminal logic they target rather than tripping the new gate.

## Provenance

Reported via a coordinated disclosure (held privately; no public PoC). The root precondition — *passwordless + public bind* — is separately addressed by the fail-closed startup guard in #5086; this PR is defense-in-depth on the terminal endpoints themselves, so the RCE path is closed even if a passwordless public bind ever slips past that guard.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5079,6 +5079,31 @@ def _onboarding_gate_allows(handler, auth_enabled: bool | None = None) -> bool:
     return _onboarding_request_is_local(handler)
 
 
+# Operator-facing copy reused by every embedded-terminal endpoint refusal.
+_EMBEDDED_TERMINAL_GATE_DENIED_MESSAGE = (
+    "Embedded terminal is only available from local networks when authentication "
+    "is not configured. Configure a password/passkey, or set "
+    "HERMES_WEBUI_ONBOARDING_OPEN=1 to allow it on a deliberately-exposed server."
+)
+
+
+def _embedded_terminal_gate_allows(handler) -> bool:
+    """Local-origin gate for the embedded-terminal endpoints.
+
+    The embedded terminal spawns a PTY shell that runs arbitrary commands as the
+    server-process user, so admitting an unauthenticated remote caller is remote
+    code execution. When auth is enabled, ``check_auth()`` has already verified
+    the session cookie before the request reaches these handlers, so this returns
+    True. When auth is DISABLED (the default out-of-the-box state) ``check_auth()``
+    admits every caller unconditionally, so restrict the terminal to local/private
+    origins — the same trust model the onboarding/bootstrap endpoints use, ignoring
+    spoofable forwarded headers unless an operator has opted into trusting them.
+    A deliberately-exposed passwordless server (access secured at another layer)
+    opts out with ``HERMES_WEBUI_ONBOARDING_OPEN=1``.
+    """
+    return _onboarding_gate_allows(handler)
+
+
 def _csp_report_rate_limited(handler, *, now: float | None = None) -> bool:
     now = time.time() if now is None else now
     key = _client_ip_for_rate_limit(handler)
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
## Rebased + fully gated — ready for merge (security, your call)

Rebased onto current master (was CONFLICTING; new head `d257e5f3`, CHANGELOG-only conflict resolved). This closes CVD#3 Finding A2 — **unauthenticated RCE via the embedded terminal** when auth is disabled + the server is network-exposed.

**Gate (full authoritative, security-adversarial):**
- **Codex: SAFE TO SHIP** — verified the gate reuses the existing onboarding local-origin helper (not hand-rolled) and covers **all** terminal endpoints (start / input / resize / close / output at `api/routes.py:14980/15020/15039/15059/15070`) *before* PTY spawn/write/output. No bypass via spoofed Host / X-Forwarded-For / origin / IPv6-loopback / 0.0.0.0; fail-closed; normal authed terminal flow unaffected.
- **Full suite: 11349 passed** (only the 2 pre-existing cron-isolation serial artifacts).
- 71 security tests (CVD3 gate + embedded-terminal + extension-hooks + security-review-fixes) pass.

Per policy I'm **not self-merging a security-boundary PR** — leaving the merge trigger to @nesquena. It's non-conflicting, CI-clean, and one-click-mergeable whenever you approve. Recommend prioritizing: it's a live unauthenticated-RCE fix that's been sitting unmerged.

nesquena-hermes:
Shipped in **v0.51.788**. Embedded-terminal endpoints are now gated to local origins when auth is disabled — closes the access gap on network-exposed instances. Full gate green, all terminal endpoints covered.

PR #5271, opened 2026-06-30
Let extension skins declare their base scheme

franksong2702:
## Thinking Path

- Hermes WebUI now lets trusted extensions register skins into the native Appearance picker.
- The current API lets a skin provide tokens, but not whether those tokens expect a light or dark base theme.
- That leaves dark editor skins mixed with System/Light base state, or light E-Ink skins mixed with Dark base state, unless each extension carries broad workaround CSS.
- Core already owns the Theme axis and the `.dark` class, so core should also own the effective light/dark base decision for registered skins.
- This PR lets extension skins declare `scheme: "light"` or `scheme: "dark"`; the user's saved Theme preference stays unchanged, while the effective `.dark` class and Prism theme follow the selected skin's scheme.

## Contract Routing

Task type: extension appearance capability / theme contract update
Touched areas: `window.registerHermesSkin()`, Appearance theme resolution, extension docs, theme docs, theme-registration tests
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/UIUX-GUIDE.md`
- `DESIGN.md`
- `docs/EXTENSIONS.md`
- `THEMES.md`
Scope boundaries:
- This does not add a new picker, a new theme value, or any backend state.
- This does not make arbitrary CSS safe; registered skin tokens still go through the existing allowlist and value sanitizer.
- This does not rewrite the stored Theme preference (`light`, `dark`, or `system`). It only changes the effective base class while a scheme-declared extension skin is active.
- Sanitizer coverage for accepted/ignored `scheme` values.
- Behavioral coverage that dark-scheme skins force `.dark` on and light-scheme skins force `.dark` off, then default skin restores the saved/base theme.
- Existing theme/skin regression tests still pass.

## Contract Change

Previous contract:
- `registerHermesSkin()` accepted descriptor name/value/label/colors/tokens and emitted one extension skin rule.
- The effective `.dark` class was controlled only by the Theme picker / System preference.
- Extension skins that were light-only or dark-only had no core-supported way to avoid being combined with incompatible base-mode tokens.

New contract:
- `registerHermesSkin()` may accept optional `scheme: "light" | "dark"`.
- Invalid scheme values are ignored.
- When a scheme-declared extension skin is selected, WebUI applies the corresponding effective `.dark` class and Prism CSS while leaving the user's stored Theme preference unchanged.

Compatibility / migration reason:
- Existing skins without `scheme` keep current behavior.
- Existing persisted Theme/Skin settings remain unchanged.
- Extension theme packs can move light/dark composition responsibility into core instead of duplicating broad token/selector workaround CSS.

## What Changed

- Added `_sanitizeSkinScheme()` for `light` / `dark` only.
- Stores an extension skin's sanitized scheme on the registered skin entry.
- Tracks the base resolved theme separately from the effective theme.
- Re-applies the effective theme when skin selection changes, so the active skin scheme can override the `.dark` class and Prism theme.
- Documented `scheme` in `docs/EXTENSIONS.md` and `THEMES.md`.
- Added behavioral Node harness coverage for scheme registration and effective `.dark` class behavior.

## Why It Matters

Dark editor skins and light display skins should not each have to patch around core's Theme/Skin composition rules. Core owns the base theme state, so this gives extension skins a small declarative way to say which base they need while preserving the user's visible Theme preference.

This unblocks cleaner extension skins such as Tokyo Night / Skin Pack and E-Ink without forcing each extension to carry broad fallback CSS for core tokens and selectors.

## Verification

- `./scripts/test.sh tests/test_extension_theme_registration.py tests/test_theme_color_meta_bridge.py`
  - 20 passed
- `./scripts/test.sh tests/test_extension_theme_registration.py tests/test_theme_color_meta_bridge.py tests/test_sienna_skin.py tests/test_catppuccin_skin.py tests/test_graphite_skin.py tests/test_1003_appearance_autosave.py tests/test_sprint26.py tests/test_issue3401_deep_review_fixes.py tests/test_zeus_skin.py tests/test_verdigris_skin.py`
  - 76 passed
- `npm run lint:runtime -- --quiet`
- `git diff --check`
- Manual headless browser check on the local WebUI runtime:
  - registered a `scheme: "dark"` skin while the base theme was light and confirmed `.dark` became active
  - registered a `scheme: "light"` skin while the base theme was dark and confirmed `.dark` was removed
  - selected the default skin afterward and confirmed the base theme was restored

## Risks / Follow-ups

- The pre-paint inline script still cannot know an extension skin's scheme before extension JavaScript registers it, so the no-flash path can only preserve the pending extension skin key. Core corrects the effective theme as soon as the extension registers.
- Existing extension PRs should switch from broad composition workaround CSS to declaring `scheme: "light"` or `scheme: "dark"` after this lands.
- This does not change code-block typography; prose placed inside fenced code blocks still renders as code.

## Model Used

AI-assisted by OpenAI GPT-5 Codex in the Codex desktop app, using local shell, codegraph, Node/Playwright-style browser verification, and GitHub CLI workflows. Human maintainer directed scope and product decision.


nesquena-hermes:
Reviewed this in depth — the design and implementation are clean. Findings:

**Design is correct.** Tracking `_resolvedThemeBaseDark` separately from the effective `.dark` class is the right call: the user's saved Theme preference is never rewritten, and `_applySkin()` re-running `_setResolvedTheme()` means a scheme-declared skin's effect recomputes whenever skin selection changes. The post-registration correction path (re-applying when a now-registered skin is the persisted selection) closes the pre-paint flash gap cleanly.

**Security holds.** `_sanitizeSkinScheme()` accepts only `light`/`dark` and the scheme value never touches the CSS token path — it's a separate field on the entry, so it can't become a CSS-injection vector. Verified `scheme: "dark;body{display:none}"` is stored as `""` while the otherwise-valid tokens still register. The existing token allowlist + value sanitizer are untouched.

**Tests pass.**
- 20/20 in `test_extension_theme_registration.py` + `test_theme_color_meta_bridge.py` in a clean checkout of this head; the node behavioral harnesses execute (not skipped).
- 35 sibling skin/theme regression tests green (sienna/catppuccin/graphite/zeus/verdigris + appearance autosave).
- CI 17/17 green.

**Adversarial edge cases I ran beyond the PR's own harness** (all correct):
- `matchMedia` OS-theme change firing *while a light-scheme skin is active* → base mode tracks correctly, so switching back to the default skin restores the right effective mode.
- light-scheme skin over OS-dark, dark-scheme skin over theme:light, System + scheme overrides — every combination resolves as expected.

No changes requested. The only thing holding this is that it's still marked **draft** — @franksong2702, mark it ready-for-review when you're set (or let me know if you're intentionally pairing it with the dependent Tokyo Night / E-Ink extension PRs first), and we'll move it through the merge gate.

nesquena-hermes:
## Concept: APPROVED ✅ — prioritized for merge

Ran the concept-judgment rubric on this now that it's out of draft.

**In scope — core-capability-for-extensions.** This is core owning a decision it already owns: the Theme axis and the `.dark` class. Letting a registered extension skin declare `scheme: "light" | "dark"` is the natural completion of the `registerHermesSkin()` surface we already shipped — without it, a dark editor skin gets mixed with a System/Light base (or a light E-Ink skin with a Dark base) unless every extension ships broad workaround CSS. Core owning the effective light/dark base for registered skins is the correct home for this, and it directly unblocks the extension-skin ecosystem (Tokyo Night / E-Ink-class skins). This is precisely the "extension hits a core wall → build the core capability first, as its own gated PR" pattern. No philosophy-veto concerns: it's opt-in, additive, and changes nothing unless an extension declares a scheme.

**Engineering is clean** (from my prior in-depth review, re-confirmed at the current head `af7da472`, CI 18/18 green):
- Design correct — `_resolvedThemeBaseDark` tracked separately from the effective `.dark` class; the user's saved Theme preference is never rewritten; `_applySkin()` recompute + the post-registration correction path closes the pre-paint flash gap.
- Security holds — `_sanitizeSkinScheme()` accepts only `light`/`dark`, the value never touches the CSS token path (verified `scheme: "dark;body{display:none}"` stores as `""` while valid tokens still register). Not a CSS-injection vector.
- Tests — 20/20 in the two theme-registration suites + 35 sibling skin/theme regression tests + adversarial OS-theme/matchMedia combinations all correct.

**Disposition:** concept-approved, gate-clean, MERGEABLE — adding to the priority review queue (Tier 1). Because this is the first-class **core extension-skin API surface** (a capability precedent for what extensions may declare), I'm leaving the actual merge trigger to @nesquena rather than self-merging — consistent with how core extension-boundary capabilities are handled. It's ready to go the moment you give the word; nothing else is holding it. Thanks @franksong2702.

nesquena-hermes:
Update: @nesquena gave the go to move this forward autonomously rather than waiting on his merge trigger. Running the full authoritative gate (Codex security-reproduce + full suite) on a fresh rebase now; will merge + ship on green. Thanks @franksong2702.

nesquena-hermes:
Shipped in **v0.51.784**. Thanks @franksong2702! 🎉

Concept-approved (core-capability-for-extensions — @nesquena gave the go to move it forward autonomously). One maintainer fix applied on the release stage since I couldn't push to your fork: `_resolvedThemeBaseDark` now initializes from the pre-painted `.dark` class instead of `false`, so a *no-scheme* extension skin can't strip `.dark` from a saved dark/system-dark theme during the async `/api/settings` window (Codex-caught dark-flash; verified fixed in a Node harness across settings success + failure). Full suite 11298 green, Codex SAFE, 20/20 theme tests. Your scheme sanitizer + token guards held up under adversarial review.

Issue #5144, opened 2026-06-28
feat(ux): opt-in busy composer placeholder hint for queue / interrupt / steer / background


rodboev:
## Summary

When a Hermes WebUI session is busy, the composer still shows the normal `Message Hermes…` placeholder. That hides the most useful power-user actions right when they matter most.

Hermes CLI already teaches this state by showing the default action and the alternate busy-time commands. Please add an **opt-in** busy-state placeholder hint in WebUI so users who want CLI-style guidance can turn it on without changing the default UI for everyone else.

## Requested UX

Keep the current idle placeholder behavior.

Add a checkbox under **Settings -> Preferences -> Busy input mode**:

- `Show busy-mode hint in composer placeholder`
- default: **off**

When that checkbox is enabled **and** the current session is busy, replace the empty composer placeholder with a compact hint derived from the active busy-input default.

Examples:

- if `Busy input mode = Queue follow-up`:
  - `Enter = queue · /interrupt · /background · /steer`
- if `Busy input mode = Interrupt current turn`:
  - `Enter = interrupt · /queue · /background · /steer`
- if `Busy input mode = Steer (mid-turn correction)`:
  - `Enter = steer · /queue · /background · /interrupt`

An icon-based variant is fine too, for example:

- `[Enter/Return] = queue • /interrupt • /background • /steer`

The important part is that the placeholder shows:

- what plain Enter will do right now,
- what the explicit alternatives are,
- and does it only for users who opted in.

## Why this helps

This would give WebUI a small but valuable piece of CLI parity.

It improves discoverability for the busy-state controls without adding more permanent buttons or cluttering the normal composer. It also fits the current product shape: `queue` is the default busy action today, but power users often need to see the nearby alternatives quickly.

## Scope

This request is intentionally narrow:

- it does **not** ask to change the default placeholder for everyone,
- it does **not** ask to redesign the composer,
- it does **not** replace the broader command discoverability work.

It is just an opt-in busy-state hint.

## Distinct from related issues

- #1804 is broader command discoverability through visible UI actions and menus.
- #3058 is about the pending-intent model and the meaning of queue / interrupt / steer.

This issue only asks for a lightweight, opt-in placeholder hint that reflects the current default busy action plus the alternate overrides.

## Implementation notes

Current relevant surfaces in-tree:

- `static/index.html` has the composer textarea placeholder and the `Busy input mode` setting.
- `static/boot.js` currently sets the placeholder to `Message ${name}…`.
[...]

PR #5279, opened 2026-06-30
[locale]Add unlocalized text for zhCN locale

Loukky:


nesquena-hermes:
Thanks for this — it's a thorough pass. Reading the diff at `static/i18n.js` against `origin/master`, the change is cleanly scoped: 213 insertions / 213 deletions, all value-only, and every hunk (lines ~9655–11211) lands inside the `zh:` locale block (`zh:` opens at 9643, `pt:` opens at 12848), so no other locale or runtime code is touched. It also clears every `// TODO: translate` marker in that block — the zh `TODO: translate` count goes from 23 on master to 0 here. Nice cleanup.

## One real gap to fix before merge

Greptile flagged it and I confirmed it. At the zhCN `composer_control_*` group, `composer_control_mic` is the only sibling left in English:

```js
// static/i18n.js:10983-10987 (zh block)
composer_control_attach: '附加',
composer_control_saved_prompts: '已保存的提示词',
composer_control_mic: 'Mic',          // <-- still English
composer_control_profile: '配置',
composer_control_workspace: '工作区',
```

This is worth fixing rather than waving through, because of how the lookup behaves. `t()` does:

```js
// static/i18n.js:22693-22694
function t(key, ...args) {
  const val = _locale[key] ?? LOCALES.en[key];
```

The `??` only falls back when the active-locale value is **nullish**. Here the key *is* present in zhCN with a literal `'Mic'`, so the fallback never engages and there's no chance for English to "win" cleanly — a CN user just sees a lone English word wedged between Chinese labels. So this isn't a silent EN fallback; it's a visible language mix.

## Suggested fix

A one-line change at line 10985:

```js
composer_control_mic: '麦克风',
```

`麦克风` (microphone) is the standard term and matches the level of the surrounding labels (`语音模式` for voice mode, etc.). For reference, the existing non-EN locales already localize this key — `de` uses `'Mikrofon'` (line 19804) and `es` uses `'Micro'` (line 16554) — so zhCN is the outlier, not a deliberate skip.

## Verification

No code path to exercise — this is a pure string table. After the edit, a quick sanity check: load the WebUI with the locale set to Simplified Chinese and confirm the composer mic control label renders as `麦克风`, and that `grep -n "composer_control_mic: 'Mic'" static/i18n.js` no longer returns the zh-block line (10985). Everything else in the PR looks good to me.

Loukky:
@nesquena-hermes  If you need, I can also translate "MIC" into Chinese. It only takes one minute？

Loukky:
`composer_control_mic` don't need to translate,It is also a commonly used word in Chinese.

nesquena-hermes:
Following up on my earlier note about `composer_control_mic` — you're right, and I should walk that back.

I went back and tallied the actual value of that key across all 14 locale blocks in `static/i18n.js` at this PR's HEAD (after your merge of master). Keeping `'Mic'` is the *majority* convention, not the outlier:

```
en→Mic   it→Mic   ja→Mic   ru→Mic   es→Mic   de→Mic
zh→Mic   zh-Hant→Mic   pt→Mic   ko→Mic   tr→Mic   vi→Mic
fr→Micro   pl→Mikrofon
```

Only `fr` (`'Micro'`) and `pl` (`'Mikrofon'`, line ~19869) localize it — the other 12, including `ja` and `ko` (the other CJK locales) and `de`, keep the bare `'Mic'`. So zhCN's `composer_control_mic: 'Mic'` (line 11020) is consistent with the established pattern. My earlier "zhCN is the outlier / `de` uses `'Mikrofon'`" claim was a line-number mis-attribution — the `'Mikrofon'` I quoted belongs to the `pl` block (~19869), not `de`. Your call as a native speaker stands; no change needed there.

The rest of the PR still looks clean. The `zh:` block (opens at line 9673, with `'zh-Hant':` opening at 11246) has zero remaining `// TODO: translate` markers, and the block structure is intact after the master merge.

One thing that is **not** a blocker, just flagging so it doesn't surprise you: the master merge pulled in two new keys from feature #5144 (commit `9a1f27a4`) that land as English in the zh block:

```js
// static/i18n.js:9979-9980 (zh block)
settings_label_busy_placeholder_hint: 'Show busy placeholder hint',
settings_desc_busy_placeholder_hint: 'Shows the composer placeholder with the current busy action while the session is running and the draft is empty.',
```

This is repo-wide post-feature drift — all 14 locale blocks carry these in English right now, not just zhCN — so it's out of scope for this PR and chasing it would be a moving target. Only noting it in case you'd like to fold those two zhCN strings in while you're here. Entirely optional.

Otherwise this is good to go from my side.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (rebased + i18n-merged clean by the gate layer)

**Certified head:** `sha:7f8c4efa` (clean rebase, branch `gate-rebase/5279-zhcn-locale`) · **PR:** #5279 · _Loukky, [locale] Add zhCN translations for previously-English UI strings_
**Verdict:** Full gate GREEN, maintainer-greenlit. Fills ~209 previously-English `zh:` values with idiomatic Simplified Chinese; en + 13 other locales untouched; full locale parity preserved. Rebased onto current master with the i18n conflict resolved by union. Ready to merge from the rebase branch.

### What I ran
| Gate | Result |
|------|--------|
| **Rebase + i18n union-merge onto current master** (`origin/master` tip `7d19792a`) | ✅ The PR's base predated master's #5144 + kanban-skills keys → zh-block conflict. Resolved by UNION (NOT take-theirs/ours): master's newer keys kept as English fallback, PR's zh translations applied where present. `git diff origin/master..HEAD` = `static/i18n.js` ONLY (+213/−213 equal = pure value replacement). |
| Codex (reproduce) | **SAFE TO SHIP** — gated the rebased worktree (`Inspecting git checkout: /tmp/wt-rebase-5279`), no findings |
| Full pytest suite | **11350 passed**, 0 failed (ran to completion, `-p no:xdist`) |
| Locale-parity tests | ✅ 43/43 (test_chinese_locale + test_login_locale_parity + test_issue3539_language_dropdown_all_locales + test_language_precedence) |
| Placeholder/fn-shape audit (mine) | ✅ 213 changed zh values, **0** `${interp}`/`{n}`/`{name}`/arrow-fn mismatches |
| Parity (mine) | ✅ every `en` key present in `zh` (NONE missing); master's new keys in all 14 blocks; only the zh block changed |

### Findings
- ✅ **Maintainer-greenlit**: nesquena-hermes walked back the `composer_control_mic` request ("your call as a native speaker stands; 12/14 locales keep `'Mic'` — majority convention") and said **"this is good to go from my side."** No longer in contributor court.
- ✅ **Pure value-fill, parity-safe**: adds/removes NO keys (+213/−213 equal); 0 remaining `// TODO: translate` markers in the zh block; idiomatic translations.
- ✅ **No runtime-break risk**: every interpolation var / brace token / arrow-fn shape preserved across all 213 changed values (verified by token-diff vs master).
- ℹ️ Non-blocking (maintainer-flagged, out of scope): two #5144 keys (`settings_{label,desc}_busy_placeholder_hint`) land as English in the zh block — but that's repo-wide post-feature drift across **all 14** locales, not a zhCN regression. Optional to fold in later; not a merge blocker.

**Ready to merge — use branch `gate-rebase/5279-zhcn-locale` (`sha:7f8c4efa`), NOT the PR's stale head `89a6b832`.** Clean rebase + correct i18n union-merge, suite green (11350/0), Codex SAFE, 43/43 parity tests, 0 placeholder mismatches, maintainer greenlit ("good to go"). i18n string-table data change — no screenshot needed. Credit @Loukky (co-authored on the rebase commit). crit=2 i18n.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy — that's the release agent's call. I rebased the stale head + resolved the i18n conflict by union (preserving master's translations) rather than block on it. Cert valid for sha:7f8c4efa._

nesquena-hermes:
Shipped in **v0.51.783**. Thanks @Loukky! 🎉 Rebased onto current master (union-resolved the i18n conflicts, preserving all newer English-fallback keys + restoring 2 large_text_paste keys the sort resolution had dropped — caught by the locale-parity tests). 365 i18n/locale/theme tests green.

PR #5294, opened 2026-06-30
fix(#4732): keep mobile dictation active through pauses

### LINKED ISSUE #4732: bug(ux): mobile dictation stops on thinking pauses (SpeechRecognition not continuous) + no wake lock
## Summary

## Where this lives

`static/boot.js:696-701`:
```js
  sr.interimResults=true;
  ...
```

## Suggested fix / fix shape

## Related

## Scope

## Environment

- Labels: `bug`, `ux`.

brianmmaina:
## Thinking Path

Hermes WebUI keeps the browser client intentionally simple: vanilla JS in `static/boot.js`, with no build step or frontend framework.

The reported mobile dictation bug lives in the browser `SpeechRecognition` path. Mobile dictation was ending too early because the recognizer used single-utterance mode and treated a thinking pause as the end of speech. It also did not hold a screen wake lock, so the page could sleep during active dictation.

Desktop server-side STT via `/api/transcribe` already records until the user stops, so this PR keeps the fix scoped to the browser `SpeechRecognition` dictation path.

## What Changed

- Set browser dictation `SpeechRecognition.continuous = true`
- Added restart-on-`onend` behavior while the user is still actively dictating
- Added a `_speechStopRequested` guard so manual stop and send actions do not restart recognition
- Preserved accumulated transcript text across recognition restarts
- Added optional Screen Wake Lock support while browser dictation is active
- Released/reacquired the wake lock on page visibility changes
- Kept the MediaRecorder `/api/transcribe` path unchanged

## Why It Matters

Mobile users can pause mid-dictation without losing the recording session, and the page is less likely to sleep while they are actively speaking.

This makes mobile browser/PWA/native-wrapper dictation closer to the desktop server-STT behavior, where capture continues until the user manually stops.

## Verification

- Ran `node --check static/boot.js`
- Ran `./scripts/test.sh`
  - Initial run: 11102 passed, 144 skipped, 2 xfailed, 1 xpassed
  - 2 tests timed out during the full run:
    - `tests/test_terminal_process_cleanup.py::test_terminal_module_registers_graceful_shutdown_reaper`
    - `tests/test_tls_aware_probe.py::test_ctl_status_survives_readonly_env_vars`
- Re-ran both timed-out tests individually and both passed:
  - `./.venv/bin/pytest tests/test_terminal_process_cleanup.py::test_terminal_module_registers_graceful_shutdown_reaper -q -vv`
  - `./.venv/bin/pytest tests/test_tls_aware_probe.py::test_ctl_status_survives_readonly_env_vars -q -vv`

Manual checks:
- Forced browser SpeechRecognition path with `mic_force_mediarecorder = 0`
- Confirmed dictation continued after a 5–7 second pause
- Confirmed manual mic stop did not restart recognition
- Confirmed send while dictating stopped recognition cleanly
- Confirmed wake-lock feature detection does not break dictation flow

## Risks / Follow-ups

- Mobile browser `SpeechRecognition` behavior differs across Chrome, Safari, and WebView wrappers
- Some browsers may still end recognition despite `continuous = true`, so this PR also restarts on `onend` while dictation is active
- Screen Wake Lock is not available in every mobile browser; unsupported browsers safely continue without it

## Release Note

Closes #4732

Improves mobile browser dictation so pausing mid-sentence no longer ends transcription early and active dictation can keep the screen awake where Wake Lock is supported.

## Model Used

Provider: OpenAI  
Model: GPT-5.5 Thinking  
Notable use: Assisted with issue analysis, implementation planning and local verification checklist.
 I reviewed and applied the code changes manually.

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -604,6 +604,8 @@ function _micToastKeyForRecognitionError(error){
   let _finalText='';
   let _prefix='';
   let _isRecording=false;
+  let _speechStopRequested=false;
+  let _micWakeLock=null;
 
   function _setButtonTooltipAndKey(btn, key){
     const text = t(key);
@@ -717,6 +719,34 @@ function _micToastKeyForRecognitionError(error){
     }
   }
 
+  async function _acquireMicWakeLock(){
+    if(!navigator.wakeLock||_micWakeLock) return;
+    try{
+      _micWakeLock=await navigator.wakeLock.request('screen');
+      _micWakeLock.addEventListener?.('release',()=>{ _micWakeLock=null; },{once:true});
+    }catch(_){
+      _micWakeLock=null;
+    }
+  }
+
+  async function _releaseMicWakeLock(){
+    const lock=_micWakeLock;
+    _micWakeLock=null;
+    if(!lock) return;
+    try{ await lock.release(); }catch(_){}
+  }
+
+  document.addEventListener('visibilitychange',()=>{
+    if(document.visibilityState==='hidden'){
+      void _releaseMicWakeLock();
+      return;
+    }
+
+    if(window._micActive&&_activeCaptureMode==='speech'){
+      void _acquireMicWakeLock();
+    }
+  });
+
   function _stopMic(){
     if(!window._micActive) return;
     // Stop the backend that was ACTIVE WHEN RECORDING STARTED — not whatever
@@ -724,6 +754,7 @@ function _micToastKeyForRecognitionError(error){
     // which would otherwise make us stop the wrong backend and orphan the other
     // (#3169 Codex review). _activeCaptureMode is pinned at start.
     if(recognition && _activeCaptureMode==='speech'){
+      _speechStopRequested=true;
       recognition.stop();
       return;
     }
@@ -739,7 +770,7 @@ function _micToastKeyForRecognitionError(error){
   function _ensureSpeechRecognition(){
     if(!SpeechRecognition) return null;
     const sr=recognition||new SpeechRecognition();
-    sr.continuous=false;
+    sr.continuous=true;
     sr.interimResults=true;
     sr.lang=(typeof _locale!=='undefined'&&_locale._speech)||'en-US';
 
@@ -763,9 +794,22 @@ function _micToastKeyForRecognitionError(error){
             ? _prefix+' '+_finalText.trimStart()
             : _prefix+_finalText)
         : ta.value;
-      _setRecording(false);
       ta.value=committed;
       autoResize();
+      if(!_speechStopRequested&&window._micActive&&_activeCaptureMode==='speech'){
+        _prefix=committed&&!committed.endsWith(' ')&&!committed.endsWith('\n')
+          ? committed+' '
+          : committed;
+        _finalText='';
+        try{
+          sr.start();
+          return;
+        }catch(_){}
+      }
+      _speechStopRequested=false;
+      _isRecording=false;
+      void _releaseMicWakeLock();
+      _setRecording(false);
       if(window._micPendingSend){
         window._micPendingSend=false;
         send();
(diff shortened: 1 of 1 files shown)


nesquena-hermes:

**Certified head:** `sha:16c78c02` (clean rebase of `1b825d96`, branch `gate-rebase/5294-mobile-dictation`) · **PR:** #5294 · _brianmmaina, fix(#4732): keep mobile dictation active through pauses_
**Verdict:** The mobile-dictation restart logic is well-built (the restart loop is correctly bounded), suite green, but the gate reproduced a real CORE regression: the change makes `continuous=true` + auto-restart **unconditional for ALL SpeechRecognition users**, so existing **desktop** one-shot dictation no longer auto-stops on a pause. Mobile-gate it + fix one wake-lock async race, then re-gate.

### What I ran (rebased worktree `/tmp/wt-rebase-5294`)
| Gate | Result |
|------|--------|
| Rebase onto current master (`cbb9f79c`) | ✅ 0 conflicts; `git diff origin/master..HEAD` = `static/boot.js +58/−3` ONLY |
| Codex (reproduce) | **SHIP-WITH-FIXES** — 1 CORE (desktop one-shot regression) + 1 SILENT (wake-lock async race); both reproduced/confirmed below |
| Full pytest suite | 11350 passed, 0 failed (can't catch a browser-behavior change) |
| My read | restart loop correctly bounded (see ✅), but the desktop-gate gap is real |

### Findings
**✅ The restart loop is correctly bounded** (the main risk for this PR, verified):
- Only ONE `recognition.stop()` call site (boot.js:758, inside `_stopMic`), and it sets `_speechStopRequested=true` immediately before. Every intentional stop (send-flow ~525, toggle-off ~885, programmatic hide) routes through `_stopMic`, so no path stops without the guard → no "mic won't turn off" bug.
- `onend` auto-restart is gated on `!_speechStopRequested && window._micActive && _activeCaptureMode==='speech'`.
- `onerror` does NOT restart on `not-allowed`/`service-not-allowed`/`audio-capture`/`network` (boot.js:833) — no infinite loop on permission denial; only transient `no-speech`/`aborted` while active are swallowed. Correct.

**⛔ CORE (reproduced) — desktop one-shot dictation no longer auto-stops on pause** (`static/boot.js:773` + `:799`): `_ensureSpeechRecognition()` sets `sr.continuous=true` and the `onend` auto-restart branch **unconditionally** — there is NO mobile/coarse-pointer gate. `origin/master` always `_setRecording(false)` on `onend` (one-shot). So the existing **desktop** flow "click mic → speak → pause → transcript commits → mic turns off" now keeps restarting on every pause; the user must manually click to stop. #4732 is a *mobile* issue, but the fix changes desktop too. **Notably the codebase already has a `hermes-voice-continuous` localStorage setting** consumed by the *other* recognition path (`_recognition.continuous=localStorage.getItem('hermes-voice-continuous')==='true'`, boot.js:1149) — so hardcoding `continuous=true` on the dictation path both regresses desktop AND bypasses the established preference pattern.
- **Fix:** gate `sr.continuous=true` and the `onend` auto-restart behind a mobile/coarse-pointer predicate (the repo already has `matchMedia('(any-pointer:fine)')` / coarse-pointer helpers at ~1769-1779) OR the existing `hermes-voice-continuous` setting; preserve `continuous=false` + no-restart (one-shot commit/stop) for desktop.

**⛔ SILENT (confirmed by read) — wake-lock can leak on an async stop race** (`static/boot.js:722`): `_acquireMicWakeLock()` does `_micWakeLock = await navigator.wakeLock.request('screen')` — it assigns AFTER the await. If the user stops/hides the mic while the request is in flight, `_releaseMicWakeLock()` (which only releases an already-assigned `_micWakeLock`) runs before the assignment, then the late-resolving lock is stored after every stop path has finished → screen wake lock held indefinitely (battery/screen drain) until the next acquire/release.
- **Fix:** add a request-generation token incremented on every stop/release; after `await request('screen')`, immediately release the returned lock unless the token is still current AND `window._micActive && _activeCaptureMode==='speech' && document.visibilityState!=='hidden'`.

**RED — gate-fail/changes-requested. Two fixes (both with exact specs above): (1) mobile-gate the continuous/restart so desktop keeps one-shot; (2) token-guard the async wake-lock acquire.** The restart-loop core is solid and the rebase is clean, so this should converge fast on a re-push. Author @brianmmaina (unproven/T3). crit=3 existing-flow. boot.js logic — no screenshot. Concept is good (mobile dictation through pauses is a real #4732 ask) — keep it, just don't regress desktop.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Cert valid for sha:16c78c02 (rebased from PR head 1b825d96)._

nesquena-hermes:
Shipped 🎉 — this was rebuilt fresh on current master as the salvage PR #5915 (applying the maintainer RED-gate fix-spec: mobile-gated continuous dictation so desktop stays one-shot, serialized wake-lock, bounded restart budget) and released in **exp-v0.52.35**'s predecessor batch **exp-v0.52.34**. Your work is preserved via `Co-authored-by: @brianmmaina` and credited in the CHANGELOG + release. Thanks for the fix! Closing this original in favor of the shipped rebuild.

PR #5298, opened 2026-06-30
Release v0.51.781 — deep-link ?q= composer prefill, converged (#4969)

### LINKED ISSUE #4961: Support URL-based prefill (`?q=`) for browser custom search engine use
### Problem

I want to use Hermes WebUI as my browser's default search engine — whatever I type in the address bar gets sent directly to Hermes for an AI response. This replaces Google, DuckDuckGo, etc. with an agent that already has persistent memory of my projects and preferences.

For this to work, the WebUI needs to accept a prompt via URL query parameter (e.g. `?q=what+is+the+weather` or `?prompt=...`) and pre-fill the chat composer with it, optionally auto-submitting.

### Proposed solution

Accept a URL query parameter (e.g. `?q=`) that pre-fills the chat composer input when the page loads. Optionally, a second parameter (e.g. `?send=1`) could auto-submit the message, which is the standard behaviour for browser custom search engines.

Example URL:
```
https://llm.imavinash.net/?q=summarize+the+latest+rust+release+notes
```

### Why Hermes WebUI over Open WebUI

Open WebUI already supports URL-based prefill and works as a browser search engine replacement. However, Hermes WebUI is significantly more feature-rich (three-panel layout, workspace browser, session management, Kanban, terminal, extensions, profiles, etc.) and ties directly into the Hermes Agent's persistent memory and cron scheduling. I'd rather use Hermes for everything.

### Use case

Browser custom search engine setup:
- Keyword: `h`
- URL: `https://llm.imavinash.net/?q=%s`

Then typing `h what's on my calendar today` in the address bar opens Hermes WebUI with that prompt pre-filled and ready to send.

nesquena-hermes:
## Release v0.51.781 — deep-link ?q= composer prefill, converged (#4969)

Converged bounce: I bounced #4969 earlier for 2 reproduced SILENT prefill-boot bugs; @rodboev re-pushed a targeted fix that closes both. Re-gated clean.

### Included
- **#4969** (@rodboev) — `?q=` deep-link pre-fills the composer; draft survives a login redirect and the prefill boot no longer starts a session early (fixes #4961).

### Convergence (both prior bounce findings closed)
1. **Prefill lost after login** → `_consumeComposerPrefillParamsFromLocation()` moved out of early boot into `_finalizeComposerPrefillOnBoot()` (consumed AFTER auth/profile bootstrap); logged-out `?q=` carried via `next`.
2. **`?q=` silently created a session** → `_maybeBindFreshDefaultWorkspaceSession(prefillIntent)` early-returns when the prefill has draft text; `S.session` stays null until send.

### Gate
- **Full pytest suite: 11286 passed** (only the 2 pre-existing cron-isolation serial artifacts).
- ESLint runtime gate CLEAN, node --check clean.
- **Codex regression gate: SAFE TO SHIP** — both findings confirmed closed; Cmd/Ctrl+K text-input guard and voice-mode prefs verified intact (no regression to adjacent logic).

CHANGELOG updated under `[Unreleased] → Fixed` with author credit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@
 
 ### Fixed
 
+- **Deep-link composer prefill via `?q=` no longer loses the draft or starts a session early.** Opening the WebUI with a `?q=...` (and related prefill params) pre-fills the composer with that text; the draft now survives a login redirect (params are consumed after auth/profile bootstrap, and the logged-out `?q=` is carried through `next`), and a prefill boot leaves the session uncreated until you actually send — it no longer silently binds a fresh default-workspace session. Thanks @rodboev. (#4969, #4961)
+
+
 - **Two new opt-in appearance skins: Neon Soft and Neon Paint.** Both are CSS-only, namespaced under `[data-skin]`, registered in server-side skin persistence, and selectable from Settings → Appearance — they change nothing unless you pick them. Neon Soft is a muted purple/violet neon; Neon Paint a bolder magenta + cyan neon. Thanks @savagebread. (#4738)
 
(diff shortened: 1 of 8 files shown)


nesquena-hermes:
Superseding with a fresh release PR rebuilt on current master — the stage went CONFLICTING after v0.51.780 landed (CHANGELOG base drift). Same #4969 code, re-staged.

PR #5309, opened 2026-07-01
ctl.sh: load ~/.hermes/.env so ${VAR} refs in config.yaml resolve

hogehou-cmi:
## Problem

`ctl.sh start` only loaded the repo's own `.env` (`~/code/github/hermes-webui/.env`), which is typically empty. Provider credentials such as `BAILIAN_BASE_URL` live in `~/.hermes/.env` and were never loaded into the ctl.sh shell environment, causing `${VAR}` references in `config.yaml` to remain as literal strings.

WebUI logs showed:
```
🌐 Endpoint: ${BAILIAN_BASE_URL}
```
instead of the resolved URL.

## Fix

- **Add `_apply_env_file_safely()`**: a helper that parses `.env` files without relying on `set -a; source` (avoids shell-readonly UID/GID aborts under `set -euo pipefail` and handles quoted values with escape sequences like `\n`, `\t`, `\\`).
- **Replace the old `set -a; source` block** in `_load_repo_dotenv_preserving_env` with `_apply_env_file_safely()`.
- **Add `_load_hermes_dotenv()`**: loads `~/.hermes/.env` after the repo dotenv, so any credential defined there is available for `${VAR}` expansion. Repo `.env` still takes precedence — already-set values are preserved.
- **Call `_load_hermes_dotenv`** in `start_cmd` and `status_cmd`.

### DIFF
--- a/ctl.sh
+++ b/ctl.sh
@@ -35,8 +35,8 @@ _apply_env_file_safely() {
   while IFS= read -r line || [[ -n "${line}" ]]; do
     line="${line#${line%%[![:space:]]*}}"
     [[ -z "${line}" || "${line}" == \#* ]] && continue
-    if [[ "${line}" == export[[:space:]]* ]]; then
-      line="${line#export }"
+    if [[ "${line}" =~ ^export[[:space:]]+(.+)$ ]]; then
+      line="${BASH_REMATCH[1]}"
       line="${line#${line%%[![:space:]]*}}"
     fi
     [[ "${line}" == *=* ]] || continue
@@ -50,8 +50,8 @@ _apply_env_file_safely() {
     esac
 
     value="${value#${value%%[![:space:]]*}}"
-    if [[ ${#value} -ge 2 && "${value:0:1}" == '"' && "${value: -1}" == '"' ]]; then
-      value="${value:1:${#value}-2}"
+    if [[ "${value}" =~ ^\"(([^\"\\]|\\.)*)\"([[:space:]]*\#.*)?[[:space:]]*$ ]]; then
+      value="${BASH_REMATCH[1]}"
       value="$(printf '%s' "$value" | awk '{
         i = 1
         len = length($0)
@@ -72,8 +72,8 @@ _apply_env_file_safely() {
           }
         }
       }')"
-    elif [[ ${#value} -ge 2 && "${value:0:1}" == "'" && "${value: -1}" == "'" ]]; then
-      value="${value:1:${#value}-2}"
+    elif [[ "${value}" =~ ^\'([^\']*)\'([[:space:]]*\#.*)?[[:space:]]*$ ]]; then
+      value="${BASH_REMATCH[1]}"
     else
       value="${value%%[[:space:]]\#*}"
       value="${value%${value##*[![:space:]]}}"
@@ -94,7 +94,9 @@ _load_repo_dotenv_preserving_env() {
     line="${line#${line%%[![:space:]]*}}"
     [[ -z "${line}" || "${line}" == \#* || "${line}" != *=* ]] && continue
     key="${line%%=*}"
-    key="${key#export }"
+    if [[ "${key}" =~ ^export[[:space:]]+(.+)$ ]]; then
+      key="${BASH_REMATCH[1]}"
+    fi
     key="${key//[[:space:]]/}"
     [[ "${key}" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || continue
     # Skip shell-readonly names (UID/GID/EUID/EGID/PPID); re-exporting them
@@ -137,7 +139,9 @@ _load_hermes_dotenv() {
     line="${line#${line%%[![:space:]]*}}"
     [[ -z "${line}" || "${line}" == \#* || "${line}" != *=* ]] && continue
     key="${line%%=*}"
-    key="${key#export }"
+    if [[ "${key}" =~ ^export[[:space:]]+(.+)$ ]]; then
+      key="${BASH_REMATCH[1]}"
+    fi
     key="${key//[[:space:]]/}"
     [[ "${key}" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || continue
     case "${key}" in
--- a/ctl.sh
+++ b/ctl.sh
@@ -29,6 +29,60 @@ ensure_home() {
   mkdir -p "${HERMES_HOME}" "${DEFAULT_STATE_DIR}"
 }
 
+_apply_env_file_safely() {
+  local env_file="$1"
+  local line key value
+  while IFS= read -r line || [[ -n "${line}" ]]; do
+    line="${line#${line%%[![:space:]]*}}"
+    [[ -z "${line}" || "${line}" == \#* ]] && continue
+    if [[ "${line}" == export[[:space:]]* ]]; then
+      line="${line#export }"
+      line="${line#${line%%[![:space:]]*}}"
+    fi
+    [[ "${line}" == *=* ]] || continue
+
+    key="${line%%=*}"
+    value="${line#*=}"
+    key="${key//[[:space:]]/}"
+    [[ "${key}" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || continue
+    case "${key}" in
+      UID | GID | EUID | EGID | PPID) continue ;;
+    esac
+
+    value="${value#${value%%[![:space:]]*}}"
+    if [[ ${#value} -ge 2 && "${value:0:1}" == '"' && "${value: -1}" == '"' ]]; then
+      value="${value:1:${#value}-2}"
+      value="$(printf '%s' "$value" | awk '{
+        i = 1
+        len = length($0)
+        while (i <= len) {
+          c = substr($0, i, 1)
+          if (c == "\\" && i < len) {
+            nc = substr($0, i+1, 1)
+            if (nc == "n") printf "\n"
+            else if (nc == "r") printf "\r"
+            else if (nc == "t") printf "\t"
+            else if (nc == "\"") printf "\""
+            else if (nc == "\\") printf "\\"
+            else { printf "\\%s", nc }
+            i += 2
+          } else {
+            printf "%s", c
+            i++
+          }
+        }
+      }')"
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (security posture is a real improvement; 2 reproduced parsing bugs on valid `.env` syntax)

**Certified head:** `sha:980339c9` (rebased onto current master, `git apply` clean) · **PR:** #5309 · _hogehou-cmi, ctl.sh: load ~/.hermes/.env so ${VAR} refs in config.yaml resolve_
**Verdict:** The core idea is good and the parser is genuinely injection-safe (a security improvement over master's `source`-based loading — I verified no command-substitution executes). But the full gate reproduced 2 CORE parsing bugs: it mis-parses `KEY="val" # comment` (retains the quotes → breaks `int(port)`, `ctl.sh start` fails) and `export\tKEY=val` (tab after export → mangles the key). Both are valid shell `.env` syntax the PR exists to parse. Fix the parsing, then ship.

### What I ran (rebased worktree `/tmp/wt-rebase-5309`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; ctl.sh-only; `bash -n` clean |
| Codex (reproduce) | SHIP-WITH-FIXES — 2 CORE parsing bugs; I independently confirmed both |
| Adversarial injection test (mine) | ✅ SAFE — `$(...)`, backticks, `$(...)`-in-quotes all stored literal (0 injection files created) |
| Full pytest suite | **11502 passed**, 0 failed (Python suite can't exercise shell parsing) |

### Findings
**✅ Security posture is a genuine improvement (keep this approach):** `_apply_env_file_safely` is a line-by-line parser with NO `source`/`eval`. I adversarially tested a malicious `.env` (`API_KEY=$(touch X)`, backticks, `$(...)` inside double-quotes) — all stored as LITERAL strings, nothing executed. Key validated `^[A-Za-z_][A-Za-z0-9_]*$`; readonly `UID/GID/EUID/EGID/PPID` skipped; `export "${key}=${value}"` assigns literal. This replaces master's `source "${_ctl_env_filtered}"` (which DOES eval) — a real hardening. The `_load_hermes_dotenv` precedence is also correct (already-set vars captured in `preserved[]` and restored, so the fallback dotenv can't clobber existing config; `HERMES_WEBUI_NO_DOTENV=1` opt-out; absent-file no-op).

**⛔ CORE (I REPRODUCED) — quoted value + trailing comment mis-parsed** (`ctl.sh:53`): the quote-strip only triggers when the value's FINAL char is the quote, so `HERMES_WEBUI_PORT="19004" # inline comment` keeps the literal quotes → I confirmed `HERMES_WEBUI_PORT` = `"19004"` (with quotes). `bootstrap.py:434` needs `int(port)` → `ctl.sh start` fails for anyone using this common shell idiom.
- **Fix:** parse a quoted value up to its CLOSING quote, then allow only whitespace + optional `#...` after it; unescape the quoted segment (no shell eval). Apply to both single and double quotes.

**⛔ CORE (I REPRODUCED) — tab after `export` mangles the key** (`ctl.sh:38`): the match accepts `export[[:space:]]*` but the removal strips only the literal `export ` (space). I confirmed `export\tHERMES_WEBUI_HOST=0.0.0.0` → key becomes `exportHERMES_WEBUI_HOST`, so `HERMES_WEBUI_HOST` is UNSET → binds default `127.0.0.1` instead of `0.0.0.0` (a behavior regression vs master for tab-separated exports).
- **Fix:** strip `export` with whitespace-aware parsing (match+remove `export[[:space:]]+`) in `_apply_env_file_safely` AND the preservation scanners (ctl.sh:96, :139).

**RED — gate-fail/changes-requested. Keep the (security-improved) parser; fix the 2 reproduced parsing bugs on valid `.env` syntax:** (1) quoted-value-with-trailing-comment → strip quotes up to the closing quote + allow trailing `#comment`; (2) tab-after-export → whitespace-aware `export` strip (also at the two preservation scanners). Add `.env`-parsing regression tests (quoted+comment, tab-export, injection-safety). Converges fast — the hard part (injection-safe design) is done right. concept 4/5 (correct ${VAR} resolution from ~/.hermes/.env is a real infra fix + the source→parser change is a security win). Author @hogehou-cmi (NEW). crit=3.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; injection-safety verified by adversarial test, both parsing bugs reproduced by running the parser. Cert valid for sha:980339c9._

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (bounce closed)

**Certified head:** `sha:03bbcf92` (clean rebase, branch `gate-rebase/5309-dotenv-load`) · **PR:** #5309 · _hogehou-cmi, ctl.sh: load ~/.hermes/.env so ${VAR} refs in config.yaml resolve_
**Verdict:** Bounce closed — the re-push fixed both parse bugs from my prior gate-fail (verified by reproducing the exact failing cases), the injection-safe design is unchanged (a security improvement over master's `source`), and the contributor added the regression tests I asked for. Ready to merge from the rebase branch.

### What I ran (rebased worktree `/tmp/wt-rebase-5309b`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; ctl.sh + new test; `bash -n` clean |
| Prior-bug reproduction (mine) | ✅ BOTH fixed (below) |
| Injection-safety re-test (mine) | ✅ preserved — `$(...)`/backticks/quoted-subst stored literal, 0 injection files created |
| New `tests/test_ctl_script.py` | ✅ 10/10 pass (the .env regression tests I requested) |
| Full pytest suite (my run) | **11516 passed**, 0 failed |
| Codex | ran its own suite into a broken-`.venv` env-flake (same `No module named requests/websockets` as #5328) and stalled classifying it; Codex itself noted "the touched files are shell startup tests; the failing bands appear later in unrelated suites." For a `ctl.sh`-only shell-parser change, the authoritative gate is the adversarial reproduction + my clean suite (11516/0), not Codex's Python-suite env noise. |

### Findings
**✅ CONVERGED — both prior parse bugs fixed (reproduced the exact failing cases):**
1. (was CORE) quoted value + trailing comment: `HERMES_WEBUI_PORT="19004" # inline comment` now parses to **`19004`** (clean — no literal quotes/comment). Previously kept the quotes → broke `int(port)`. Fixed.
2. (was CORE) tab after `export`: `export\tHERMES_WEBUI_HOST=0.0.0.0` now sets **`HERMES_WEBUI_HOST=0.0.0.0`** (previously mangled the key to `exportHERMES_WEBUI_HOST` → host defaulted to 127.0.0.1). Fixed.

**✅ Security posture unchanged (still a hardening win):** `_apply_env_file_safely` remains a line-by-line parser with NO `source`/`eval`; I re-tested a malicious `.env` (`EVIL=$(touch X)`, `QUOTED_EVIL="x $(touch Y)"`, backticks) — all stored as LITERAL strings, 0 injection files created. Key regex, readonly-var skip, and literal `export` all intact. This still improves on master's `source "${_ctl_env_filtered}"`.

**✅ Tests added:** the re-push added `tests/test_ctl_script.py` (10 cases, all pass) — locking the quoted-comment + export-tab + injection behavior, exactly the regression coverage I asked for in the bounce.

**Ready to merge — use branch `gate-rebase/5309-dotenv-load` (`sha:03bbcf92`), NOT the PR's stale head `01a75f07`.** Bounce closed: both parse bugs fixed (reproduced-as-working), injection-safe design preserved (a security improvement over master's `source`), 10/10 new ctl-script tests, my full suite 11516/0. The Codex run stalled on its own broken-`.venv` env (not a PR defect — my authoritative run of the same code is clean; a `ctl.sh`-only change is gated by the adversarial shell reproduction, which passed). concept 4/5 (correct `${VAR}` resolution from ~/.hermes/.env + source→parser hardening). Credit @hogehou-cmi (NEW, co-authored). crit=3.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; both parse fixes + injection-safety reproduced, new tests 10/10, my suite 11516/0. Cert valid for sha:03bbcf92._

nesquena-hermes:
## Gate review — one fix needed before merge (otherwise clean) 🔧

Ran the full authoritative gate on a fresh rebase onto current master. The `_apply_env_file_safely` parser is **solid** — I ran an adversarial pass and confirmed it's safe:
- ✅ No command injection: `KEY=$(cmd)`, backticks, and `KEY=a; cmd` are all stored as literal string values, never executed.
- ✅ Valid-uncommon syntax parses correctly: quoted values containing `#`, single-quote literals (no expansion), escape sequences, `=` inside values, empty values, `export` prefix.
- ✅ `UID`/`GID`/`EUID`/`EGID`/`PPID` correctly skipped; key-name validation enforced.
- ✅ Full suite green (11504 passed).

### One MUST-FIX (Codex reproduced it)
`ctl.sh:134` — `ctl.sh start` / `ctl.sh status` **abort** when `~/.hermes/.env` **exists but is unreadable** (e.g. `chmod 000`). Under `set -euo pipefail`, `_load_hermes_dotenv` checks only `-f` and then reads the file, so an unreadable fallback env file exits `1` with `Permission denied`. `origin/master` starts fine because it never reads that file — so this is a new abort path the PR would introduce.

**Exact fix (one line):**
```bash
# ctl.sh:134 — treat an unreadable fallback env file like a missing one
-  [[ -f "${hermes_env}" ]] || return 0
+  [[ -r "${hermes_env}" ]] || return 0
```
(A matching `-r` guard on the repo `.env` path is worth a glance too, for symmetry.)

With that one change this is good to go — the parser itself is well-built. Marking `changes-requested`; ping me when it's pushed and I'll re-gate + move it forward.

nesquena-hermes:
Shipped in **v0.51.812** — thanks @hogehou-cmi! 🎉

`ctl.sh` now loads `~/.hermes/.env` before launch so `${VAR}` references in `config.yaml` resolve, using a literal parser (no `source`/`eval`, injection-safe) that handles quoted values, inline comments, and `export`-prefixed lines. 10 new ctl-script tests. Combined-stage gate: Codex SAFE-TO-SHIP + full suite 11639 passed.

PR #5313, opened 2026-07-01
feat: support per-provider reasoning_efforts in config.yaml

CharlesMcquade:
## What
Adds `providers.<name>.reasoning_efforts` config support to `resolve_model_reasoning_efforts()`.

## Why
Some models only support a subset of Hermes's reasoning effort levels. For example, GLM-5.2 on W&B's vLLM endpoint only recognizes "high" and "max" (xhigh) for `reasoning_effort`, plus `enable_thinking: false` for non-think mode. The other levels ("minimal", "low", "medium") are silently ignored.

Without this change, the WebUI reasoning dropdown shows all 5 effort levels for every reasoning-capable model. With this change, users can declare which levels are valid:

```yaml
providers:
  wandb:
    base_url: https://api.inference.wandb.ai/v1
    reasoning_efforts:
      - none
      - high
      - xhigh
```

The dropdown then filters to only those options.

## How
Adds a new step 0 in `resolve_model_reasoning_efforts()` that checks `providers.<name>.reasoning_efforts` before falling back to existing heuristics. Handles both `custom:<name>` and bare registered provider names. Falls through gracefully when no config entry is present — zero behavior change for existing setups.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -3023,17 +3023,38 @@ def _candidate_supports_reasoning(candidate: str) -> bool:
     return False
 
 
+# Matches the nested Gemini gateway route prefix anywhere it appears in a
+# model id, as long as it isn't embedded inside a larger alphanumeric token
+# (the negative lookbehind excludes false positives like "notvertex/gemini-x").
+# Scanning for the pattern at any position — rather than requiring the whole
+# string to start with it — makes the check independent of how many wrapper
+# layers precede the route: ``@provider:``, a named custom-provider slug
+# (``@custom:<slug>:``), or any future nesting scheme none of us have
+# invented yet. This invariant (Gemini image/embedding routes must never
+# expose a reasoning toggle) was bypassed twice via different edge cases in
+# the prefix-stripping logic before being made structurally boundary-based
+# instead of prefix-based — see PR #5313 review history.
+_NESTED_ROUTE_PATTERN = re.compile(r"(?<![a-z0-9])(vertex/gemini-|gemini_cli/gemini-)(.*)$")
+
+
 def _nested_route_reasoning_denied(model: str) -> bool:
-    """Hard deny for nested Gemini gateway routes that must never show a reasoning toggle."""
+    """Hard deny for nested Gemini gateway routes that must never show a reasoning toggle.
+
+    Matches the route pattern anywhere in *model*, not just when the whole
+    string starts with it, so this check does not depend on a caller having
+    stripped exactly the right wrapper prefix first. Callers should still
+    pass the least-wrapped form they have (e.g. after
+    ``_strip_provider_hint_for_reasoning``) for clarity, but correctness no
+    longer hinges on it.
+    """
     lower = str(model or "").strip().lower()
     if not lower:
         return False
-    for prefix in ("vertex/gemini-", "gemini_cli/gemini-"):
-        if lower.startswith(prefix):
-            tail = lower[len(prefix) :]
-            if tail.startswith("embedding") or "image" in tail or "imagine" in tail:
-                return True
-    return False
+    match = _NESTED_ROUTE_PATTERN.search(lower)
+    if not match:
+        return False
+    tail = match.group(2)
+    return tail.startswith("embedding") or "image" in tail or "imagine" in tail
 
 
 def _nested_gateway_route_reasoning(model: str) -> bool:
--- a/api/config.py
+++ b/api/config.py
@@ -3355,8 +3355,10 @@ def resolve_model_reasoning_efforts(
             if isinstance(_prov_entry, dict):
                 _re_list = _prov_entry.get("reasoning_efforts")
                 if isinstance(_re_list, list) and _re_list:
-                    return [str(x).strip().lower() for x in _re_list
-                            if str(x).strip().lower() in {*VALID_REASONING_EFFORTS, "none"}]
+                    _filtered = [str(x).strip().lower() for x in _re_list
+                                 if str(x).strip().lower() in {*VALID_REASONING_EFFORTS, "none"}]
+                    if _filtered:
+                        return _filtered
         except Exception:
             pass
 
--- a/api/config.py
+++ b/api/config.py
@@ -3340,10 +3340,17 @@ def resolve_model_reasoning_efforts(
 
     provider = _resolve_provider_alias(provider)
 
+    # IDE-copilot providers never expose reasoning effort options.
+    # Guard early so a stray config entry can't override this.
+    if provider in {"cursor-acp", "copilot-acp"}:
+        return []
+
     # 0. Provider config: providers.<name>.reasoning_efforts
     # When the user has explicitly listed valid efforts for a provider,
     # return that list directly — no heuristics, no models.dev lookup.
     # Handles both custom:<name> and bare registered provider names (e.g. wandb).
+    # Only short-circuits when the filtered list is non-empty; an all-invalid
+    # list (e.g. typos) falls through to heuristics instead of hiding reasoning.
     _prov_name = None
     if provider and provider.startswith("custom:"):
         _prov_name = provider.split(":", 1)[1]
@@ -3361,9 +3368,6 @@ def resolve_model_reasoning_efforts(
                         return _filtered
         except Exception:
             pass
-
-    if provider in {"cursor-acp", "copilot-acp"}:
-        return []
 
     hinted_model = _strip_provider_hint_for_reasoning(model)
 
--- a/api/config.py
+++ b/api/config.py
@@ -3345,34 +3345,39 @@ def resolve_model_reasoning_efforts(
     if provider in {"cursor-acp", "copilot-acp"}:
         return []
 
-    # 0. Provider config: providers.<name>.reasoning_efforts
-    # When the user has explicitly listed valid efforts for a provider,
-    # return that list directly — no heuristics, no models.dev lookup.
-    # Handles both custom:<name> and bare registered provider names (e.g. wandb).
-    # Only short-circuits when the filtered list is non-empty; an all-invalid
-    # list (e.g. typos) falls through to heuristics instead of hiding reasoning.
-    _prov_name = None
-    if provider and provider.startswith("custom:"):
-        _prov_name = provider.split(":", 1)[1]
-    elif provider:
-        _prov_name = provider
-    if _prov_name:
-        try:
-            _prov_entry = (cfg.get("providers") or {}).get(_prov_name, {})
-            if isinstance(_prov_entry, dict):
-                _re_list = _prov_entry.get("reasoning_efforts")
-                if isinstance(_re_list, list) and _re_list:
-                    _filtered = [str(x).strip().lower() for x in _re_list
-                                 if str(x).strip().lower() in {*VALID_REASONING_EFFORTS, "none"}]
-                    if _filtered:
-                        return _filtered
-        except Exception:
-            pass
-
     hinted_model = _strip_provider_hint_for_reasoning(model)
 
+    # Master hides reasoning controls for nested image/embedding routes. Keep
+    # that hard deny above provider config so an explicit allowlist cannot
+    # re-enable controls for routes that should never expose them.
     if _nested_route_reasoning_denied(hinted_model):
         return []
+
+    # 0. Provider config: providers.<name>.reasoning_efforts or named
+    # custom_providers[].reasoning_efforts. When the user has explicitly listed
+    # valid efforts for a provider, return that list directly — no heuristics,
+    # no models.dev lookup.
+    # Only short-circuits when the filtered list is non-empty; an all-invalid
+    # list (e.g. typos) falls through to heuristics instead of hiding reasoning.
+    _re_list = None
+    try:
+        if provider and provider.startswith("custom:"):
+            for _entry in _custom_provider_entries():
+                if _custom_provider_slug_from_name(_entry.get("name")) == provider:
+                    _re_list = _entry.get("reasoning_efforts")
+                    break
+        elif provider:
+            _prov_entry = (cfg.get("providers") or {}).get(provider, {})
+            if isinstance(_prov_entry, dict):
+                _re_list = _prov_entry.get("reasoning_efforts")
+        if isinstance(_re_list, list) and _re_list:
+            _filtered = [str(x).strip().lower() for x in _re_list
+                         if str(x).strip().lower() in {*VALID_REASONING_EFFORTS, "none"}]
+            _filtered = list(dict.fromkeys(_filtered))
+            if _filtered:
+                return _filtered
+    except Exception:
+        pass
 
     if provider in {"copilot", "github-copilot"}:
         try:
--- a/api/config.py
+++ b/api/config.py
@@ -2902,10 +2902,32 @@ def parse_reasoning_effort(effort):
     return None
 
 
-def _strip_provider_hint_for_reasoning(model_id: str) -> str:
-    """Remove WebUI routing hints before provider-specific capability lookup."""
+def _strip_provider_hint_for_reasoning(model_id: str, provider: str | None = None) -> str:
+    """Remove WebUI routing hints before provider-specific capability lookup.
+
+    A plain ``@provider:model`` hint strips cleanly on the first colon. But a
+    *named* custom provider hint is ``@custom:<slug>:model`` — two colons —
+    and the naive first-colon split only strips the leading ``@custom:``,
+    leaving ``<slug>:model`` behind. That leftover slug fragment can hide a
+    nested gateway route from prefix-based checks like
+    ``_nested_route_reasoning_denied()`` (e.g. ``agg:vertex/gemini-image-1.0``
+    no longer starts with ``vertex/gemini-``), silently re-enabling reasoning
+    controls on routes that must never expose them.
+
+    When the resolved *provider* is known (e.g. ``"custom:agg"``), strip the
+    exact ``@{provider}:`` prefix first so both segments are removed in one
+    pass. Falls back to the generic first-colon split when no provider is
+    given or it doesn't match — preserving prior behavior for plain
+    ``@provider:model`` hints.
+    """
     model = str(model_id or "").strip()
-    if model.startswith("@") and ":" in model:
+    if not model.startswith("@"):
+        return model
+    if provider:
+        exact_prefix = f"@{provider}:".lower()
+        if model.lower().startswith(exact_prefix):
+            return model[len(exact_prefix) :]
+    if ":" in model:
         return model.split(":", 1)[1]
     return model
 
@@ -3345,7 +3367,7 @@ def resolve_model_reasoning_efforts(
     if provider in {"cursor-acp", "copilot-acp"}:
         return []
 
-    hinted_model = _strip_provider_hint_for_reasoning(model)
+    hinted_model = _strip_provider_hint_for_reasoning(model, provider)
 
     # Master hides reasoning controls for nested image/embedding routes. Keep
     # that hard deny above provider config so an explicit allowlist cannot
--- a/api/config.py
+++ b/api/config.py
@@ -3339,6 +3339,27 @@ def resolve_model_reasoning_efforts(
             provider = str((cfg.get("model") or {}).get("provider") or "").strip().lower()
 
     provider = _resolve_provider_alias(provider)
+
+    # 0. Provider config: providers.<name>.reasoning_efforts
+    # When the user has explicitly listed valid efforts for a provider,
+    # return that list directly — no heuristics, no models.dev lookup.
+    # Handles both custom:<name> and bare registered provider names (e.g. wandb).
+    _prov_name = None
+    if provider and provider.startswith("custom:"):
+        _prov_name = provider.split(":", 1)[1]
+    elif provider:
+        _prov_name = provider
+    if _prov_name:
+        try:
+            _prov_entry = (cfg.get("providers") or {}).get(_prov_name, {})
+            if isinstance(_prov_entry, dict):
+                _re_list = _prov_entry.get("reasoning_efforts")
+                if isinstance(_re_list, list) and _re_list:
+                    return [str(x).strip().lower() for x in _re_list
+                            if str(x).strip().lower() in {*VALID_REASONING_EFFORTS, "none"}]
+        except Exception:
+            pass
+
     if provider in {"cursor-acp", "copilot-acp"}:
         return []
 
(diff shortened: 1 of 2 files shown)


CharlesMcquade:
Thanks for the review @greptile-apps — both points are valid. Fixed in b80ea016:

1. **All-invalid config silently returns `[]`** — Now only short-circuits when the filtered list is non-empty. An all-invalid list (e.g. typos) falls through to the existing heuristics instead of hiding reasoning support from the UI.

2. **`cursor-acp`/`copilot-acp` guards bypassed** — Moved the hardcoded `return []` for those providers before the config lookup, so a stray `providers.cursor-acp.reasoning_efforts` entry can't surface options the endpoint doesn't support.

nesquena-hermes:
Reading the diff at HEAD (`api/config.py:3348-3369`) against `resolve_model_reasoning_efforts()` on `origin/master`, plus `_get_provider_cfg`/`_custom_provider_entries` to understand where provider config lives. The concept is good and the fall-through is safe, but the `custom:<name>` branch reads from the wrong config section, so the headline use case (a named custom proxy) will silently never hit step 0.

## Code reference

The new step-0 lookup:

```python
_prov_name = None
if provider and provider.startswith("custom:"):
    _prov_name = provider.split(":", 1)[1]
elif provider:
    _prov_name = provider
if _prov_name:
    try:
        _prov_entry = (cfg.get("providers") or {}).get(_prov_name, {})
        ...
```

For a `custom:<name>` provider, `_prov_name` is the slug suffix and the lookup is `cfg["providers"][suffix]`. But named custom providers do not live under the `providers:` mapping — they live in the top-level `custom_providers:` **list**. See `api/config.py:2435`:

```python
def _get_providers_cfg() -> dict:
    providers_cfg = cfg.get("providers")
    return providers_cfg if isinstance(providers_cfg, dict) else {}
```

versus `api/config.py:1332`:

```python
def _custom_provider_entries(config_obj=None):
    source = config_obj if isinstance(config_obj, dict) else cfg
    entries = source.get("custom_providers", [])
    ...
```

These are two distinct top-level keys. A user who configures `custom:llm-proxy` (the exact shape in sibling report #5314) has their entry in `custom_providers`, so `cfg["providers"]["llm-proxy"]` returns `{}` and step 0 is skipped. The `reasoning_efforts` key would only be honored for bare *registered* provider names (your `wandb` example, which is under `providers:`), not for `custom:*`.

## Recommendation

Split the two branches. For bare registered providers, `cfg["providers"][name]` is correct. For `custom:<name>`, resolve the matching `custom_providers` entry by slug and read `reasoning_efforts` from it. Something like:

```python
if provider.startswith("custom:"):
    for _entry in _custom_provider_entries():
        if _custom_provider_slug_from_name(_entry.get("name")) == provider:
            _re_list = _entry.get("reasoning_efforts")
            break
else:
    _re_list = (cfg.get("providers") or {}).get(provider, {}).get("reasoning_efforts")
```

(`_custom_provider_slug_from_name` at `api/config.py:1318` already returns the `custom:<slug>` form, so compare against the full `provider` string.)

Two smaller notes:
- Greptile already flagged the missing dedup — every other return path in this function wraps with `list(dict.fromkeys(...))`. Worth matching for consistency even though a hand-written list rarely has dups.
- The guard-before-step-0 move for `cursor-acp`/`copilot-acp` (`api/config.py:3341`) is correct and important; good call keeping config from overriding the hard `[]`.

## Test plan

Add two cases to whichever test module covers `resolve_model_reasoning_efforts`: (1) a `custom_providers: [{name: llm-proxy, reasoning_efforts: [none, high, xhigh]}]` config with `provider="custom:llm-proxy"` should return the filtered list; (2) a bare `providers: {wandb: {reasoning_efforts: [...]}}` with `provider="wandb"`. Case (1) will fail against the current diff, which is the bug this comment is about.

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (good feature; one reproduced ordering regression + zero tests)

**Certified head:** `sha:b80ea016` (rebased onto current master, `git apply` clean) · **PR:** #5313 · _CharlesMcquade, support per-provider reasoning_efforts in config.yaml_
**Verdict:** The feature is a clean, well-guarded addition (reads `providers.<name>.reasoning_efforts`, filters to valid values, fail-safe), but the full gate reproduced a real CORE ordering regression: the new provider-config short-circuit runs BEFORE the nested-route deny check, so a configured `reasoning_efforts` re-enables reasoning controls on Gemini image/embedding routes that master hard-denied. Plus the branch has zero tests (AGENTS.md). Both have exact fixes + I've written the ready-to-use test.

### What I ran (rebased worktree `/tmp/wt-rebase-5313`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; config.py-only |
| Codex (reproduce) | SHIP-WITH-FIXES — 1 CORE ordering regression + 1 SILENT (zero tests); I confirmed both |
| Full pytest suite | **11502 passed**, 0 failed (no nested-route+provider-config test exists → can't catch the regression) |
| In-process verification (mine) | logic mostly correct (below), but the ordering bug reproduced |

### Findings
**✅ The feature logic is well-guarded:** the new branch reads `providers.<name>.reasoning_efforts`, filters to `VALID_REASONING_EFFORTS ∪ {none}`, only short-circuits when the filtered list is non-empty (all-invalid falls through — no hiding), handles `custom:<name>` + bare names, `try/except` fail-safe. I verified: configured `[low,high]` → returned; `custom:myprov [none,medium]` → returned; case-normalized; all-invalid `[bogus,typo]` → NOT returned (filtered).

**⛔ CORE (I REPRODUCED) — configured efforts bypass the nested-route deny** (`api/config.py:3368`): the provider-config short-circuit `return`s BEFORE `_nested_route_reasoning_denied()` (config.py:3374), but master ran that deny FIRST. I confirmed in-process that `custom:agg` with `reasoning_efforts:[low,high]` returns `['low','high']` for **`vertex/gemini-image-1.0`** and **`vertex/gemini-embedding-001`** — routes where `_nested_route_reasoning_denied` is True (master returned `[]`). So a per-provider config re-enables reasoning controls on image/embedding routes that must never have them — a regression.
- **Fix:** compute `hinted_model` and run `_nested_route_reasoning_denied(hinted_model)` immediately after the ACP guard and BEFORE the provider-config branch (restore master's ordering: ACP guard → nested-route deny → provider config → heuristics).

**⛔ SILENT — zero tests on a non-trivial config resolver branch** (AGENTS.md: "Tests required for all non-trivial logic"). The existing reasoning test slice has no `providers.<name>.reasoning_efforts` coverage, so the ordering bug + edge behavior aren't locked.
- **Fix:** I've written a ready-to-drop-in test (`tests/test_issue5313_per_provider_reasoning_efforts.py`) — 9 cases: configured-valid-returned, case-normalized, all-invalid-fallthrough, `custom:<name>`, ACP-guard-wins, no-config-passthrough (non-regressive), non-dict/non-list ignored, **and the nested-Gemini-deny-wins case that locks the ordering fix**. 8/9 pass against the current head; the nested-Gemini case FAILS until the ordering fix lands (which is exactly what should gate it). Full file in the collapsed block below.

<details><summary>Ready-to-use test file (add after the ordering fix)</summary>

```python
import pytest
import api.config as cfg

@pytest.fixture
def _providers_cfg():
    sentinel = object()
    orig = cfg.cfg.get("providers", sentinel)
    def _set(mapping): cfg.cfg["providers"] = mapping
    yield _set
    if orig is sentinel: cfg.cfg.pop("providers", None)
    else: cfg.cfg["providers"] = orig

def test_configured_valid_efforts_returned_directly(_providers_cfg):
    _providers_cfg({"wandb": {"reasoning_efforts": ["low", "high"]}})
    assert cfg.resolve_model_reasoning_efforts("some-model", "wandb") == ["low", "high"]

def test_configured_efforts_are_case_normalized(_providers_cfg):
    _providers_cfg({"wandb": {"reasoning_efforts": ["LOW", " High "]}})
    assert cfg.resolve_model_reasoning_efforts("some-model", "wandb") == ["low", "high"]

def test_all_invalid_efforts_fall_through_not_returned(_providers_cfg):
    _providers_cfg({"wandb": {"reasoning_efforts": ["bogus", "typo"]}})
    r = cfg.resolve_model_reasoning_efforts("some-model", "wandb")
    assert r == [] or ("bogus" not in r and "typo" not in r)

def test_custom_prefixed_provider_resolves(_providers_cfg):
    _providers_cfg({"myprov": {"reasoning_efforts": ["none", "medium"]}})
    assert cfg.resolve_model_reasoning_efforts("some-model", "custom:myprov") == ["none", "medium"]

def test_copilot_acp_guard_wins_over_config(_providers_cfg):
    _providers_cfg({"copilot-acp": {"reasoning_efforts": ["high"]}})
    assert cfg.resolve_model_reasoning_efforts("some-model", "copilot-acp") == []

def test_provider_without_config_is_non_regressive(_providers_cfg):
    _providers_cfg({"wandb": {}})
    a = cfg.resolve_model_reasoning_efforts("some-model", "wandb")
    _providers_cfg({})
    assert a == cfg.resolve_model_reasoning_efforts("some-model", "wandb")

def test_non_dict_provider_entry_does_not_crash(_providers_cfg):
    _providers_cfg({"wandb": "not-a-dict"})
    assert isinstance(cfg.resolve_model_reasoning_efforts("some-model", "wandb"), list)

def test_non_list_reasoning_efforts_does_not_crash(_providers_cfg):
    _providers_cfg({"wandb": {"reasoning_efforts": "low"}})
    assert isinstance(cfg.resolve_model_reasoning_efforts("some-model", "wandb"), list)

def test_nested_gemini_deny_wins_over_provider_config(_providers_cfg):
    # deny must run BEFORE the provider-config short-circuit
    _providers_cfg({"agg": {"reasoning_efforts": ["low", "high"]}})
    for model in ("vertex/gemini-image-1.0", "vertex/gemini-embedding-001"):
        assert cfg.resolve_model_reasoning_efforts(model, "custom:agg") == [], model
```
</details>

**RED — gate-fail/changes-requested. Two fixes: (1) move `_nested_route_reasoning_denied()` before the provider-config branch (restore master's ACP-guard → nested-deny → config → heuristics ordering); (2) add the resolver test above (esp. the nested-Gemini-deny case that locks #1).** The feature is otherwise clean and non-regressive for normal providers. Converges fast. concept 4/5 (per-provider reasoning config is a reasonable, consumed-end-to-end feature). Author @CharlesMcquade (unproven). crit=2.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; the ordering regression reproduced in-process (Gemini image/embedding routes), test written + verified (8/9 pass, nested-Gemini case locks the fix). Cert valid for sha:b80ea016._

CharlesMcquade:
Thanks for the detailed review — both substantive issues were real. Fixed in 4f4c8fd1:

1. **Named `custom:<name>` providers** now read `reasoning_efforts` from the matching `custom_providers[]` entry using the existing `_custom_provider_slug_from_name()` helper, instead of incorrectly looking only in `providers:<name>`.

2. **Nested Gemini image/embedding route deny** now runs before the provider-config shortcut, preserving master behavior so config cannot re-enable reasoning controls for routes that must hide them.

3. Also added the consistency nit: configured efforts are now deduped with `list(dict.fromkeys(...))` before returning.

Added focused regression coverage for:
- bare `providers.<name>.reasoning_efforts`
- named `custom_providers[].reasoning_efforts`
- case normalization/filtering/deduping
- all-invalid config fallthrough
- ACP hard guards winning over config
- nested Gemini route deny winning over config

Verification:
```bash
./scripts/test.sh tests/test_reasoning_effort_model_capabilities.py tests/test_custom_provider_bare_model_reasoning.py
# 52 passed in 1.74s
```

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (converged on the plain-ordering fix; a deeper provider-qualified nested route still leaks)

**Certified head:** `sha:4f4c8fd1` (rebased onto current master, `git apply` clean) · **PR:** #5313 · _CharlesMcquade, support per-provider reasoning_efforts in config.yaml_
**Verdict:** Good progress — the re-push fixed my prior ordering CORE (nested-route deny now runs before the config short-circuit) AND improved the custom-provider lookup AND added tests. But the full gate reproduced a **deeper edge of the same invariant**: a provider-qualified model `@custom:<slug>:vertex/gemini-image-1.0` still bypasses the nested-route deny (the hint-strip leaves a `<slug>:` fragment), re-enabling reasoning on image/embedding routes. One more targeted fix + test, then ship.

### What I ran (rebased worktree `/tmp/wt-rebase-5313b`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; config.py + test |
| Codex (reproduce) | SHIP-WITH-FIXES — 1 CORE (deeper nested-deny bypass); I confirmed it |
| Full pytest suite | **11519 passed**, 0 failed (no provider-qualified-nested test → doesn't catch the leak) |
| PR's reasoning tests | ✅ 18/18 |
| My reproduction | plain-ordering FIXED ✅; provider-qualified nested LEAKS ⛔ |

### Findings
**✅ CONVERGED — my prior ordering CORE is fixed:** the resolver order is now ACP-guard → `_nested_route_reasoning_denied(hinted_model)` → provider-config → heuristics. I confirmed: plain `vertex/gemini-image-1.0` + configured efforts → `[]` (deny wins). The custom-provider lookup was also improved (reads real `custom_providers[]` via `_custom_provider_slug_from_name`), and the feature still works: bare `providers.wandb` → `['low','high']`, `custom:my-agg` → `['none','medium']` for non-denied models. Tests added (18/18).

**⛔ CORE (I REPRODUCED) — provider-qualified nested route still bypasses the deny** (`api/config.py:3353` via `_strip_provider_hint_for_reasoning` at :2908): for `@custom:my-agg:vertex/gemini-image-1.0`, the hint-strip uses a first-colon split and returns **`my-agg:vertex/gemini-image-1.0`** (leaving the slug fragment), so `_nested_route_reasoning_denied()` misses the nested route and the config short-circuit returns `['low','high']`. I reproduced both `@custom:my-agg:vertex/gemini-image-1.0` and `…:vertex/gemini-embedding-001` returning `['low','high']` instead of `[]`. Codex verified production reachability (UI preserves `@custom:slug:model` at ui.js:2005/2026 → sends to `/api/reasoning` at ui.js:3645). Same invariant as before, deeper vector.
- **Fix:** strip the exact known provider prefix before the nested-deny — after provider alias resolution, remove a matching `@{provider}:` (e.g. `@custom:my-agg:`) so `@custom:my-agg:vertex/gemini-image-1.0` normalizes to `vertex/gemini-image-1.0` before `_nested_route_reasoning_denied()`; fall back to the existing helper otherwise.
- **Test:** extend the nested-route deny test (test_reasoning_effort_model_capabilities.py:210) to cover `@custom:agg:vertex/gemini-image-1.0` and `@custom:agg:vertex/gemini-embedding-001`, both returning `[]`.

**RED — gate-fail/changes-requested (1 fix + 1 test): provider-aware hint stripping before the nested-route deny, so a `@custom:<slug>:<nested-model>` can't re-enable reasoning on Gemini image/embedding routes.** The plain-ordering fix + custom lookup + feature behavior are all correct now; this is the last edge of the same class. Converges fast. concept 4/5. Author @CharlesMcquade (unproven). crit=2. (This is the 2nd round on the nested-route-deny invariant — each round the contributor closed the reported vector but a deeper qualified-model variant remained; the provider-aware strip closes the class.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; plain-ordering fix verified working, provider-qualified nested leak reproduced (`@custom:slug:vertex/gemini-image-1.0` → `['low','high']`). Cert valid for sha:4f4c8fd1._

CharlesMcquade:
Thanks for reproducing the deeper bypass — fixed in `0109baa1`.

**Root cause:** `_strip_provider_hint_for_reasoning()` used a naive first-colon split, so a provider-qualified hint like `@custom:agg:vertex/gemini-image-1.0` only had the `@custom:` wrapper removed, leaving `agg:vertex/gemini-image-1.0` behind. That leftover slug fragment no longer starts with `vertex/gemini-`, so `_nested_route_reasoning_denied()` missed it and a configured `reasoning_efforts` allowlist re-enabled reasoning controls on Gemini image/embedding routes that must never expose them.

**Fix:** `_strip_provider_hint_for_reasoning()` now accepts the resolved provider id and strips the exact `@{provider}:` prefix (e.g. `@custom:agg:`) in one pass, removing both the wrapper and the named-provider slug before the nested-route deny check runs. Falls back to the original first-colon split when no provider is supplied, so plain `@provider:model` hints are unaffected.

Verified in-process:
```
@custom:agg:vertex/gemini-image-1.0      -> []  (was ['low', 'high'])
@custom:agg:vertex/gemini-embedding-001  -> []  (was ['low', 'high'])
```

Added the regression test you specified, extending `test_nested_route_deny_wins_over_configured_reasoning_efforts` with a provider-qualified variant covering both routes.

**Verification:**
```bash
./scripts/test.sh tests/test_reasoning_effort_model_capabilities.py tests/test_custom_provider_bare_model_reasoning.py
# 53 passed in 1.23s
```

Also ran the full suite (`./scripts/test.sh -q`): 11258 passed, 63 pre-existing failures (network/credential/profile-isolation/fd-leak tests, unrelated to reasoning_efforts) — confirmed identical failure set on the unmodified branch as a baseline, so nothing regressed.

Ready for another pass whenever convenient.

CharlesMcquade:
Structural hardening, per Charles's call after the 3rd round on this invariant.

Rather than patch the next edge case, `_nested_route_reasoning_denied()` is now **boundary-based instead of prefix-based** (`f9de93f4`):

- **Before:** required the model string to *start with* `vertex/gemini-` or `gemini_cli/gemini-`. Every wrapper layer (`@provider:`, a named custom-provider slug, any future nesting scheme) had to be stripped exactly right by the caller first, or the deny silently failed open.
- **After:** searches for the route pattern *anywhere* in the string at a non-alphanumeric boundary (regex with a negative lookbehind to reject embedded substrings like `notvertex/gemini-x`). Correctness no longer depends on `_strip_provider_hint_for_reasoning()` having stripped the exact right prefix — any number of opaque wrapper layers can precede the route and the deny still fires.

Verified against all three historical bypass strings plus a hypothetical deeper double-wrapped case (`@custom:outer:@custom:inner:vertex/gemini-embedding-001`) — all correctly denied. Embedded-substring false positives (`notvertex/gemini-x`, `somegemini_cli/gemini-image-1`) correctly do NOT match.

Added `test_nested_route_deny_is_boundary_based_not_prefix_based` locking in the structural invariant directly rather than one more prefix-specific regression test.

**Verification:**
```bash
./scripts/test.sh tests/test_reasoning_effort_model_capabilities.py tests/test_custom_provider_bare_model_reasoning.py
# 54 passed in 1.31s
```

Full suite: 11257 passed, 64 pre-existing failures (identical file/test set as the prior verification — network/credential/profile-isolation/fd-leak tests, confirmed unrelated to this change).

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (bounce closed, round 3)

**Certified head:** `sha:3a61a0a7` (clean rebase, branch `gate-rebase/5313-reasoning-efforts`) · **PR:** #5313 · _CharlesMcquade, support per-provider reasoning_efforts in config.yaml_
**Verdict:** Bounce closed after 3 gate rounds on the nested-route-deny invariant. The re-push made the deny **boundary-based** and strips the named-provider slug before it, closing every model-string form I could construct. Codex SAFE, suite green, feature works for non-denied models. Ready to merge from the rebase branch.

### What I ran (rebased worktree `/tmp/wt-rebase-5313c`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; config.py + tests |
| Full pytest suite | **11521 passed**, 0 failed |
| My exhaustive deny-boundary test | ✅ **0 leaks across 6 model-string forms** |

### Findings
**✅ CONVERGED — the whole nested-route-deny bypass class is closed.** The 3 rounds on this invariant:
- round 1: config short-circuit ran *before* the nested-route deny → fixed (deny moved first).
- round 2: provider-qualified `@custom:slug:vertex/gemini-image-1.0` bypassed (first-colon-split left a `slug:` fragment) → **now fixed**.
- round 3 (this): the re-push made the deny "boundary-based" + strips the named-provider slug before it. I exhaustively verified **0 leaks** across: `vertex/gemini-image-1.0`, `@custom:my-agg:vertex/gemini-image-1.0`, `custom:my-agg:vertex/gemini-image-1.0` (no `@`), `@custom:my-agg:vertex/gemini-embedding-001`, whitespace-padded, and `vertex/gemini-image-preview` — all correctly return `[]`.

**✅ Feature still works + non-regressive:** `@custom:my-agg:some-real-model` → `['low','high']` (configured efforts applied to a non-denied qualified model); bare `providers.wandb` and `custom:my-agg` return configured efforts; all-invalid lists fall through to heuristics; ACP guard still wins. Codex SAFE (0 findings), suite 11521/0, PR's reasoning tests pass.

**Ready to merge — use branch `gate-rebase/5313-reasoning-efforts` (`sha:3a61a0a7`), NOT the PR's stale head `f9de93f4`.** Three rounds of gating converged the nested-route-deny invariant to a boundary-based check that I verified leak-free across every model-string form; Codex SAFE + suite 11521/0 + feature works for non-denied models. concept 4/5 (per-provider reasoning config is a real, consumed-end-to-end feature). Credit @CharlesMcquade (co-authored). crit=2. (Good example of the gate loop closing a subtle security-adjacent invariant that a green suite never caught — each round the contributor fixed the reported vector and the gate found the next deeper one, until the boundary-based refactor closed the class.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; nested-deny closed across 6 model-string forms (0 leaks) + feature verified working + Codex SAFE + suite 11521/0. Cert valid for sha:3a61a0a7._

CharlesMcquade:
Acknowledged the gate certification from @nesquena-hermes. This is a GREEN / converged result, not a new code finding, so I did not push additional changes to the PR branch.

Key note preserved from the certifier: the certified head is `gate-rebase/5313-reasoning-efforts` at `3a61a0a7`, not the current stale PR head `f9de93f4`. The gate report says the nested-route-deny invariant is closed, Codex is SAFE, the full suite passed at 11521/0, and the exhaustive deny-boundary check had 0 leaks across the tested model-string forms.

nesquena-hermes:
Shipped in **v0.51.803**. Thanks @CharlesMcq — the boundary-based `_NESTED_ROUTE_PATTERN` hardening is exactly right; Codex confirmed a `reasoning_efforts` config entry can't re-enable a denied Gemini image/embedding route (deny runs before the config lookup), verified across wrapped/nested ids. Full suite green.

CharlesMcquade:
Thanks for the confirmation and the thorough verification! Glad the nested-route hardening landed cleanly in v0.51.803. 🎉

PR #5317, opened 2026-07-01
fix(model): dynamically repair bare custom-provider models using acti…

### LINKED ISSUE #5314: Bug: Custom provider bare model repair fails for non-default models on subsequent turns
# Bug: Custom provider bare model repair fails for non-default models on subsequent turns

## Summary
In conversations where the user selects a custom provider model (e.g., `x-ai/grok-composer-2.5-fast` via `custom:llm-proxy`) that is different from the profile's configured default model (e.g., `umans/umans-glm-5.2`), the first turn completes successfully. However, the client-side select dropdown automatically normalizes and strips the provider namespace prefix from the selection. On the 2nd user input, the client POSTs the bare model ID (`grok-composer-2.5-fast`), which the server fails to repair back to its full qualified name. This leads to a terminal HTTP 400 error when forwarding the bare ID to the LLM proxy.

## Motivation / Problem
In `api/routes.py`, `_resolve_compatible_session_model_state()` is designed to repair bare custom-provider models on process wakeups, server-initiated continuations, and subsequent chat turns. However, the repair condition is overly restrictive:

```python
if (
    "/" not in model
    and _profile_default
    and "/" in _profile_default
    and _profile_default.rsplit("/", 1)[-1] == model
    and (
        _profile_provider_normalized == "custom"
        or str(profile_provider).startswith("custom:")
    )
):
    return _profile_default, profile_provider, True
```

The check `_profile_default.rsplit("/", 1)[-1] == model` **only** repairs a bare model if its suffix matches the profile's default model (e.g., `umans-glm-5.2`). If the user is actively using a non-default custom model (e.g., `grok-composer-2.5-fast`), the repair check fails, the model is forwarded bare, and the upstream proxy rejects the request.

## Proposal
We propose updating the fast-path and slow-path repair checks in `_resolve_compatible_session_model_state()` to dynamically query the active custom provider's configured models list in `config.yaml` (`custom_providers`).

```python
_repaired_model = None
if (
    "/" not in model
    and (
        requested_provider == "custom"
        or str(requested_provider).startswith("custom:")
    )
):
    try:
        from api.config import cfg as _active_cfg
        _custom_cfg = _active_cfg.get("custom_providers", []) if isinstance(_active_cfg, dict) else []
        _cp_name = requested_provider.split(":", 1)[1] if ":" in requested_provider else ""
        _matching_cp = None
        if isinstance(_custom_cfg, list) and _cp_name:
            for _entry in _custom_cfg:
                if isinstance(_entry, dict) and _entry.get("name", "").strip().lower() == _cp_name.lower():
                    _matching_cp = _entry
                    break
        if _matching_cp:
            _cp_model = (_matching_cp.get("model") or "").strip()
            _cp_model_ids = set()
            if _cp_model:
                _cp_model_ids.add(_cp_model)
            _cp_models = _matching_cp.get("models")
            if isinstance(_cp_models, dict):
                _cp_model_ids.update(
                    _key.strip()
                    for _key in _cp_models.keys()
                    if isinstance(_key, str) and _key.strip()
                )
            for _id in _cp_model_ids:
                if "/" in _id and _id.rsplit("/", 1)[-1] == model:
                    _repaired_model = _id
                    break
    except Exception:
        pass

if _repaired_model:
    return _repaired_model, requested_provider, True
```

This guarantees that any bare model belonging to the configured custom provider is dynamically re-qualified back to its fully-namespaced form, regardless of whether it is the profile's configured default model.

## Non-goals
- Overhauling first-party provider family mappings (which remain unaffected).
- Modifying client-side select dropdown regex normalization behavior (which is preserved for UI consistency).

## Open Questions
- None.

## Ask
Please review this fix and accept the proposed PR to unblock users of named custom proxies who select non-default models.

b3nw:
# PR: Dynamically repair bare custom provider models using active custom provider catalog

Closes #5314

## Thinking Path
During our investigation of a recurring bug where conversations using custom provider models crashed with an HTTP 400 "Model not found" error specifically on the 2nd user input, we traced the execution flow:
1. **Client-side Stripping:** After Turn 1 completes, the client-side select dropdown value is synchronized. Due to prefix normalization in `_findModelInDropdown()` (`static/ui.js` line 2222), the namespace prefix `"x-ai/"` is stripped, causing the dropdown value to settle on the bare `"grok-composer-2.5-fast"`.
2. **Subsequent Turn POST:** On Turn 2, the client POSTs the bare model ID (`grok-composer-2.5-fast`).
3. **Server-side Repair Defect:** In the server-side model resolver `_resolve_compatible_session_model_state()`, the suffix repair check was overly restrictive:
   ```python
   _profile_default.rsplit("/", 1)[-1] == model
   ```
   This only repaired a bare model if its suffix matched the profile's configured default model (e.g., `umans/umans-glm-5.2` in `steward-llm-proxy`'s configuration). If the user selected any other custom model (e.g. `grok-composer-2.5-fast`), it fell through un-repaired, sending the bare ID upstream which is rejected by the proxy.

We resolved this by updating the resolver to dynamically search the active custom provider's configured model list (`custom_providers`) in `config.yaml` to reconstruct the fully namespaced ID.

## What Changed
- **`api/routes.py` (`_resolve_compatible_session_model_state`)**:
  - Enhanced both the **Fast Path** (lines 5892+) and **Slow Path** (lines 6033+) checks to dynamically look up the active custom provider name in the profile's configuration.
  - If a matching custom provider is found, it extracts all registered model IDs and checks if any namespaced ID's suffix matches the bare model ID. If a match is found, it dynamically repairs it to the full namespaced form.
- **`tests/test_issue5127_process_wakeup_bare_model.py`**:
  - Added `test_fast_path_repairs_non_default_bare_model_of_custom_provider` and `test_slow_path_repairs_non_default_bare_model_of_custom_provider` to ensure that custom-provider models are repaired even when they do not match the profile's configured default model.

## Why It Matters
This fix prevents user conversations from breaking on the second turn when using any custom provider model that is not explicitly configured as the profile's default. It ensures high durability and robust model routing for all multi-model proxies.

## Verification
- **Pytest Suite:**
  - Ran the regression test suite locally inside the `.venv`:
    ```bash
    ./scripts/test.sh tests/test_issue5127_process_wakeup_bare_model.py -q
    ```
  - **Result:** All 12 tests passed (including the 2 new regression test cases).

## Risks / Follow-ups
Zero risk to standard first-party providers (OpenAI, Anthropic, Google) as they are explicitly bypassed in the custom repair check. 

No direct modifications were made to the client-side select dropdown regex normalization behavior, preserving its UI consistency.

## Model Used
- **Provider:** hermes-webui/desktop using gemini-3-flash-preview

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5760,14 +5760,16 @@ def _should_attach_codex_provider_context(model: str, raw_active_provider: str, 
         )
     return False
 
-
 def _read_profile_model_config(
-    session, requested_provider: str | None,
-) -> tuple[str | None, str | None]:
-    """Read model.provider and model.default from the session's profile config.
+    session,
+    requested_provider: str | None,
+) -> tuple[str | None, str | None, dict | None]:
+    """Read model.provider, model.default, and the full profile config dict.
 
-    Returns (profile_provider, profile_default_model). Both are None when the
-    session has no profile or the profile config is unreadable.
+    Returns (profile_provider, profile_default_model, profile_config_dict).
+    The first two are None when the session has no profile or the profile config
+    is unreadable; profile_config_dict is None in the same cases so callers only
+    pay for one YAML parse.
 
     When the session already has an explicit ``requested_provider``, the profile
     ``model.provider`` is not returned (first tuple element is None) so profile
@@ -5776,7 +5778,7 @@ def _read_profile_model_config(
     provider matches ``requested_provider`` after normalization.
     """
     if not getattr(session, "profile", None):
-        return None, None
+        return None, None, None
 
     try:
         from api.profiles import get_hermes_home_for_profile
@@ -5784,16 +5786,16 @@ def _read_profile_model_config(
         _profile_home = get_hermes_home_for_profile(session.profile)
         _profile_cfg_path = os.path.join(str(_profile_home), "config.yaml")
         if not os.path.isfile(_profile_cfg_path):
-            return None, None
+            return None, None, None
         import yaml
 
         with open(_profile_cfg_path, encoding="utf-8") as _f:
             _pcfg = yaml.safe_load(_f) or {}
         if not isinstance(_pcfg, dict):
-            return None, None
+            return None, None, None
         _model_cfg = _pcfg.get("model") or {}
         if not isinstance(_model_cfg, dict):
-            return None, None
+            return None, None, _pcfg
         _provider = (_model_cfg.get("provider") or "").strip() or None
         _default = (_model_cfg.get("default") or "").strip() or None
     except Exception:
@@ -5802,15 +5804,15 @@ def _read_profile_model_config(
             getattr(session, "profile", None),
             exc_info=True,
         )
-        return None, None
+        return None, None, None
 
     _requested = _clean_session_model_provider(requested_provider)
     if _requested:
         _profile_prov = _clean_session_model_provider(_provider)
         if _profile_prov != _requested:
-            return None, None
-        return None, _default
-    return _provider, _default
+            return None, None, _pcfg
+        return None, _default, _pcfg
+    return _provider, _default, _pcfg
 
 
 def _load_profile_config_dict(session) -> dict | None:
@@ -5840,6 +5842,34 @@ def _load_profile_config_dict(session) -> dict | None:
         return None
 
 
+def _ordered_custom_provider_model_ids(entry: dict) -> list[str]:
+    """Model ids from a custom_providers entry (default model + dict/list models)."""
+    ordered: list[str] = []
+    _cp_model = str(entry.get("model") or "").strip()
+    if _cp_model:
+        ordered.append(_cp_model)
+    _cp_models = entry.get("models")
+    if isinstance(_cp_models, dict):
+        for _key in _cp_models.keys():
+            if isinstance(_key, str):
+                _kid = _key.strip()
+                if _kid and _kid not in ordered:
+                    ordered.append(_kid)
+    elif isinstance(_cp_models, list):
+        for _item in _cp_models:
+            if isinstance(_item, str):
+                _mid = _item.strip()
+                if _mid and _mid not in ordered:
+                    ordered.append(_mid)
+            elif isinstance(_item, dict):
+                _mid = str(
+                    _item.get("id") or _item.get("model") or _item.get("name") or ""
+                ).strip()
+                if _mid and _mid not in ordered:
+                    ordered.append(_mid)
+    return ordered
+
+
 def _repair_bare_custom_provider_model(
     bare_model: str,
     provider: str | None,
@@ -5851,11 +5881,12 @@ def _repair_bare_custom_provider_model(
     Returns the fully namespaced model id when ``bare_model`` matches the suffix
     of a registered id on ``custom_providers``; otherwise None. Model ids are
     scanned in config declaration order (default ``model`` first, then
-    ``models`` dict keys) so repair is deterministic when suffixes collide.
+    ``models`` dict keys or list entries) so repair is deterministic when
+    suffixes collide.
 
     When ``config_obj`` is set (typically the session profile's config.yaml),
-    only that object's ``custom_providers`` are scanned. Otherwise falls back
-    to the process-global config for backward-compatible test stubs.
+    only that object's ``custom_providers`` are scanned. Otherwise uses
+    ``get_config()`` for the active global config (not the raw ``cfg`` alias).
     """
     try:
         model = str(bare_model or "").strip()
@@ -5864,38 +5895,36 @@ def _repair_bare_custom_provider_model(
             return None
         if prov != "custom" and not str(prov).startswith("custom:"):
             return None
-        from api.config import _custom_provider_entries, cfg as _active_cfg
+        from api.config import (
+            _custom_provider_entries,
+            _custom_provider_slug_from_name,
+            get_config,
+        )
 
         if isinstance(config_obj, dict):
             _entries = _custom_provider_entries(config_obj)
         else:
+            _cfg = get_config()
             _entries = _custom_provider_entries(
-                _active_cfg if isinstance(_active_cfg, dict) else None
+                _cfg if isinstance(_cfg, dict) else None
             )
-        _cp_name = str(prov.split(":", 1)[1] if ":" in prov else "").strip()
-        if not _cp_name:
-            return None
-        _cp_name_lower = _cp_name.lower()
+        prov_norm = str(prov).strip().lower()
+        raw_suffix = prov_norm.removeprefix("custom:")
         _matching_cp = None
         for _entry in _entries:
-            _entry_name = str(_entry.get("name") or "").strip().lower()
-            if _entry_name == _cp_name_lower:
+            entry_name = str(_entry.get("name") or "").strip().lower()
+            slug = _custom_provider_slug_from_name(_entry.get("name"))
+            if not slug:
+                continue
+            if (
+                prov_norm in {entry_name, slug}
+                or raw_suffix == slug.removeprefix("custom:")
+            ):
                 _matching_cp = _entry
                 break
         if not _matching_cp:
             return None
-        _ordered_ids: list[str] = []
-        _cp_model = str(_matching_cp.get("model") or "").strip()
-        if _cp_model:
-            _ordered_ids.append(_cp_model)
-        _cp_models = _matching_cp.get("models")
-        if isinstance(_cp_models, dict):
-            for _key in _cp_models.keys():
-                if isinstance(_key, str):
-                    _kid = _key.strip()
-                    if _kid and _kid not in _ordered_ids:
-                        _ordered_ids.append(_kid)
-        for _id in _ordered_ids:
+        for _id in _ordered_custom_provider_model_ids(_matching_cp):
             if "/" in _id and _id.rsplit("/", 1)[-1] == model:
                 return _id
         return None
@@ -6355,8 +6384,7 @@ def _resolve_effective_session_model_for_display(session) -> str:
     """
     original_model = getattr(session, "model", None) or ""
     requested_provider = getattr(session, "model_provider", None)
-    _pp_provider, _pp_default = _read_profile_model_config(session, requested_provider)
-    _pp_cfg = _load_profile_config_dict(session)
+    _pp_provider, _pp_default, _pp_cfg = _read_profile_model_config(session, requested_provider)
     effective_model, _provider, _changed = _resolve_compatible_session_model_state(
         original_model or None,
         requested_provider,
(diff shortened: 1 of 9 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (clean fix concept; 2 reproduced CORE — a crash-on-malformed-config + wrong-profile scan)

**Certified head:** `sha:5d07c2de` (rebased onto current master, `git apply` clean) · **PR:** #5317 · _b3nw, dynamically repair bare custom-provider models (#5314/#5127/#1855)_
**Verdict:** The repair concept is sound and the call-sites are cleanly additive (fail-through on None), suite green. But the full gate reproduced 2 CORE defects: (1) the helper **500s on a malformed `custom_providers` entry** (non-string `name`) and it now runs on `GET /api/session` / chat-start; (2) it reads the **process-global config**, so in multi-profile setups it can repair to the wrong profile's model. Both have exact fixes.

### What I ran (rebased worktree `/tmp/wt-rebase-5317`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ merge-base == master tip (`05aa4731`); `git apply` clean |
| Codex (reproduce) | SHIP-WITH-FIXES — 2 CORE; I independently confirmed both |
| Full pytest suite | **11413 passed**, 0 failed (doesn't test malformed config → can't catch #1) |
| Call-site + helper read (mine) | additive/fail-safe ✅; but 2 CORE below |
| PR's own tests | ✅ 13/13 (happy + no-op paths — but none drive malformed config or multi-profile) |

### Findings
**✅ Concept + call-sites are good:** `_repair_bare_custom_provider_model` re-qualifies a bare model id to a namespaced id by suffix-matching the named custom provider's config (deterministic order: default `model` then `models` dict keys). Both call-sites (routes.py:5953, 6058) are ADDITIVE — `if _repaired: return _repaired; return model, ...` — a None result falls through to prior behavior (no regression when repair doesn't apply). No config mutation.

**⛔ CORE (I REPRODUCED) — 500 on malformed `custom_providers`** (`routes.py:~5850`): `_entry.get("name", "").strip()` (and the models-dict key `.strip()`) coerce only when the KEY is absent, not when it's present-but-non-string. I injected `{"name": None, ...}` and called the helper: **`AttributeError: 'NoneType' object has no attribute 'strip'`**. Since the helper now runs from the display resolvers (`GET /api/session`) and chat-start before the old no-op fallback, a single malformed config entry 500s session load for bare custom-provider sessions. (The `model` field IS coerced with `or ""`; `name` and the dict keys are not.)
- **Fix:** coerce every scanned config value — `str(_entry.get("name") or "").strip()`, guard non-str model keys, and/or wrap the scan/build in `try/except Exception: return None` (fail closed). Add a malformed-config regression test.

**⛔ CORE (I confirmed the premise) — wrong-profile config in multi-profile flows** (`routes.py:~5833`): the helper reads the process-global `api.config.cfg`, but the resolver is operating in session/profile context (it derives `profile_provider`/`_profile_default` from `_read_profile_model_config`). In a multi-profile / per-request-cookie deployment where profiles have distinct `custom_providers` with colliding model suffixes, this can rewrite `custom:team` + `shared` to a *sibling profile's* `vendor/shared` → wrong model routing/provider failure. Lower-likelihood than #1 (needs colliding suffixes across profiles) but a real correctness gap on a routing decision.
- **Fix:** scan the same session/profile config used to derive `profile_provider`/`_profile_default` (thread the profile config/list into the helper), and only fall back to `get_config()` for the same-request non-profile case — not the raw global `cfg` alias.

**RED — gate-fail/changes-requested. Two reproduced CORE with exact fixes: (1) fail-closed coercion of config values (crash-on-malformed-config, the more urgent — it 500s session load); (2) profile-scope the config scan (wrong-profile repair).** The concept + additive call-sites are good, so this converges once both are addressed + a malformed-config and a multi-profile regression test added. concept 4/5 (bare-model repair is a real #5127/#5314 reliability fix). Author @b3nw (T2). crit=3.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; both CORE reproduced by executing the helper (malformed-config crash + global-cfg read). Cert valid for sha:5d07c2de._

b3nw:
@nesquena-hermes updates are in

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (my 2 prior findings FIXED; 3 more config-shape edges — 2 feature-breaking)

**Certified head:** `sha:3682d237` (rebased onto current master, `git apply` clean) · **PR:** #5317 · _b3nw, dynamically repair bare custom-provider models (#5314/#5127/#1855)_
**Verdict:** Good progress — the re-push fixed both my prior CORE findings (malformed-config no longer 500s; repair is now profile-scoped via `config_obj`, cross-profile isolation verified). But the full gate reproduced 3 more edges in the same "which config shape does the resolver handle" family: **2 CORE make the feature silently not work** for slugged provider names and list-form models, plus a SILENT raw-`cfg` fallback. Fix those and it converges.

### What I ran (rebased worktree `/tmp/wt-rebase-5317b`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 2 CORE + 1 SILENT; I confirmed all 3 |
| Full pytest suite | **11519 passed**, 0 failed (no slugged-name / list-form test → doesn't catch #2/#3) |
| PR's own tests | ✅ 31/31 |
| My reproduction | prior 2 findings FIXED ✅; 3 new edges confirmed ⛔ |

### Findings
**✅ CONVERGED — my prior 2 CORE are fixed:** (1) malformed `custom_providers` (`name:null`, non-str `model`, non-str model keys) no longer 500s — returns `None` gracefully (verified). (2) repair is now profile-scoped: `_repair_bare_custom_provider_model(..., config_obj=profile_config)` at both call-sites (routes.py:6062, 6169); I verified cross-profile isolation (profile A→`vendor/foo`, B→`other/foo`, empty→`None`, no cross-leak).

**⛔ CORE (I REPRODUCED) — slugged provider names never repair** (`api/routes.py:5950`): the helper compares the incoming `provider` (`custom:my-proxy`) against the raw `entry["name"]` (`"My Proxy"`) instead of the slugged form. I verified a provider named "My Proxy" (slug `custom:my-proxy`) → repair returns **`None`**. So the feature works ONLY for providers coincidentally named as their own slug (like "team"); any normal display-name provider silently fails to repair.
- **Fix:** match entries by `api.config._custom_provider_slug_from_name(entry["name"])`, not the raw name.

**⛔ CORE (I REPRODUCED) — list-form `models` never repair** (`api/routes.py:5966`): the helper only scans dict-form `models` keys, but the catalog supports list-form `models: ["id/a", {"id": "id/b"}]`. I verified `models: ["profile-list/shared", {"id": "profile-list/other"}]` + bare `shared` → **`None`** (list not scanned).
- **Fix:** collect IDs from both dict keys AND list entries (string, or dict `id`/`model`/`name`), same as the catalog code.

**⛔ SILENT — `config_obj=None` fallback uses raw `api.config.cfg`, not `get_config()`** (`api/routes.py:5942`): the no-profile fast path reads the raw global `cfg` alias, bypassing the config reload/override contract — can repair against stale global config for default/no-profile sessions.
- **Fix:** call `get_config()` and pass that dict to `_custom_provider_entries` in the fallback.

**RED — gate-fail/changes-requested (3 fixes, all in the config-shape-handling family): (1) match by slugged name; (2) scan list-form models; (3) use `get_config()` in the None fallback.** My 2 prior findings are genuinely closed and cross-profile isolation is solid — but #2/#3 mean the feature silently doesn't work for the common cases (display-named providers, list-form model catalogs), which a green suite masks (no test covers them). Add tests for slugged-name + list-form repair. concept 4/5. Author @b3nw (T2). crit=3. (This is the config-resolver blind-spot class: a happy-path test + green suite pass while un-tested config shapes silently no-op — same lesson as #5313.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; prior 2 findings verified fixed + cross-profile isolation verified, 3 new config-shape edges reproduced (slugged-name None, list-form None, raw-cfg fallback). Cert valid for sha:3682d237._

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (bounce closed, round 3)

**Certified head:** `sha:647db074` (clean rebase, branch `gate-rebase/5317-bare-model-repair`) · **PR:** #5317 · _b3nw, fix(model): dynamically repair bare custom-provider models (#5314/#5127/#1855)_
**Verdict:** Bounce closed after 3 gate rounds on the config-shape family. Every finding across all rounds is fixed and verified: I ran an exhaustive 10-shape config test (0 failures) plus cross-profile isolation + malformed-config no-crash + get_config() fallback. Codex SAFE, suite green. Ready to merge from the rebase branch.

### What I ran (rebased worktree `/tmp/wt-rebase-5317c`)
| Gate | Result |
|------|--------|
| Full pytest suite | **11593 passed**, 0 failed |
| Exhaustive config-shape test (mine) | ✅ **0 failures / 10 shapes** |

### Findings — all 3 rounds' findings closed
- **round 1** (CORE): malformed `custom_providers` 500'd + repair read the process-global `cfg` → **fixed** (fail-closed on malformed; `config_obj` profile-scoping). Re-verified: cross-profile isolation (A→`vendor/foo`, B→`other/foo`, no leak); malformed (`name:null`, non-str model, non-str keys) → `None`, no crash.
- **round 2** (2 CORE + 1 SILENT): slugged provider names never repaired (matched raw `name` not the slug); list-form `models` never scanned (dict-keys only); `config_obj=None` fallback used the raw `cfg` alias → **all fixed this round** (commit cites my cert #4857080754). Re-verified: slugged `My Proxy`(→`custom:my-proxy`)→`vendor/shared`; list-form `["id",{id}]`→`profile-list/shared`; the None-fallback now calls `get_config()` (docstring + line confirm).
- **round 3** (this): my exhaustive test passes 0/10 failures across every shape — dict-models, list-str, list-dict `{id}`, list-dict `{model}`, slugged-name, single `model` field, no-suffix-match→None, non-dict entry (skipped, next entry matched), missing-name→None, non-dict/non-list `models`→None. Suffix-match correct, deterministic order, no cross-profile leak, no crash on any malformed shape.

Both call-sites (routes.py:6062, 6169) remain additive (`config_obj=profile_config`; None result falls through to prior behavior). Codex SAFE, suite 11593/0.

**Ready to merge — use branch `gate-rebase/5317-bare-model-repair` (`sha:647db074`), NOT the PR's stale head `6f9c648d`.** Three rounds converged the config-resolver blind-spot class to a resolver that handles every custom_providers shape (dict + list-form models, slugged + raw names, malformed/missing, profile-scoped) — verified leak-free and crash-free across 10 shapes + Codex SAFE + suite 11593/0. Backend model-id repair — no visible surface. concept 4/5 (real #5314/#5127/#1855 reliability fix). Credit @b3nw (co-authored). crit=3. (Good gate arc: each round the contributor closed the reported shape and the gate reproduced the next un-handled config shape, until the whole family was covered — a green suite never caught any of the 3 rounds because no test exercised the un-driven shapes.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; all 3 rounds' config-shape findings verified fixed (10-shape exhaustive test 0 failures + cross-profile + malformed + get_config fallback), Codex SAFE + suite 11593/0. Cert valid for sha:647db074._

nesquena-hermes:
Shipped in **v0.51.808**. Thanks @b3nw for iterating through three gate rounds — this head closes all findings: slugged `@custom:<slug>` names matched, list-form `models` scanned, `config_obj`/`get_config()` profile-scoping (no raw-cfg fallback), malformed-config graceful, auth-before-resolution preserved. Codex SAFE, full suite green (11525).

PR #5328, opened 2026-07-01
fix(webui): stop live streaming message flicker

ruizanthony:
## Summary
- Opt live assistant streaming nodes out of the global theme color/background transitions.
- Prevent token-by-token markdown updates from inheriting global `div` color/background fades, which can look like white flicker/flash on light themes.
- Add a targeted regression test while keeping the opt-in smooth text fade feature intact.

## Testing
- `./scripts/test.sh tests/test_smooth_text_fade.py -q`

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -1458,6 +1458,21 @@
   :root:not(.dark) .cron-btn:hover{background:rgba(0,0,0,.08);}
   /* ── Smooth dark mode transitions ── */
   body,header,footer,aside,nav,main,div,button,input,textarea,select{transition-property:background-color,border-color,color;transition-duration:.15s;transition-timing-function:ease;}
+  /* Live assistant content is updated token-by-token. The global div color/background
+     transition above makes each streamed markdown rewrite briefly fade/flash on light
+     themes. Keep theme transitions elsewhere, but make the live turn paint instantly. */
+  #liveAssistantTurn,
+  #liveAssistantTurn *,
+  #thinkingRow,
+  #thinkingRow *,
+  .assistant-segment[data-live-assistant="1"],
+  .assistant-segment[data-live-assistant="1"] *,
+  .agent-activity-thinking[data-thinking-active="1"],
+  .agent-activity-thinking[data-thinking-active="1"] *,
+  .agent-activity-thinking[data-live-thinking="1"],
+  .agent-activity-thinking[data-live-thinking="1"] *,
+  .live-worklog[data-live-worklog-shell="1"],
+  .live-worklog[data-live-worklog-shell="1"] *{transition-property:none!important;transition-duration:0s!important;transition-delay:0s!important;}
   :root{--app-titlebar-safe-top:0px;}
   .pwa-standalone{overscroll-behavior:none;}
   .pwa-standalone body{-webkit-tap-highlight-color:transparent;}
--- a/tests/test_smooth_text_fade.py
+++ b/tests/test_smooth_text_fade.py
@@ -197,6 +197,35 @@ def test_stream_fade_done_drain_has_hard_cap_for_large_buffered_responses():
     )
 
 
+def test_live_streaming_assistant_content_opts_out_of_global_theme_transitions():
+    """Per-token markdown rewrites must not inherit global div color/background fades.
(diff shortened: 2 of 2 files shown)


nesquena-hermes:

**Certified head:** `sha:7e334dcf` (clean rebase, branch `gate-rebase/5328-streaming-flicker`) · **PR:** #5328 · _ruizanthony, stop live streaming message flicker_
**Verdict:** Full gate GREEN. A well-scoped CSS-only fix: disables the global div color/bg transition ONLY for live-streaming DOM so per-token markdown rewrites stop flashing/fading on light themes, while theme transitions everywhere else are preserved. Verified by suite + CSSOM + settle-analysis + Codex. Ready to merge from the rebase branch.

### What I ran (rebased worktree `/tmp/wt-rebase-5328`, port 8783)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean (style.css + test); merge-base current |
| Codex (reproduce) | **SAFE TO SHIP** — gated the rebased worktree, 0 findings (its own suite noise was a broken `.venv`: `No module named 'requests'/'websockets'` — env, not PR; confirmed by my clean run) |
| Full pytest suite (my run) | **11503 passed**, 0 failed |
| CSSOM verification (light theme, live server) | ✅ live-turn selectors → `transition: none/0s`; normal `.msg-body` → `background-color,border-color,color / 0.15s` (global theme transition PRESERVED) |
| Selectors match real DOM | ✅ 47 `liveAssistantTurn` refs, 23 `data-live-assistant`, etc. — not a no-op |

### Findings
**✅ Correctly scoped (not a global transition kill):** the `transition-property/duration/delay:none!important` applies ONLY to `#liveAssistantTurn`, `#thinkingRow`, `.assistant-segment[data-live-assistant="1"]`, `.agent-activity-thinking[data-thinking-active/live-thinking="1"]`, `.live-worklog[data-live-worklog-shell="1"]` (+ `*` descendants). CSSOM-confirmed in a light theme: a live-turn element resolves to `transition: none`, while a normal `.msg-body` keeps the global `0.15s` color/bg transition. So the fix kills the streaming flicker without touching dark/light theme-switch animation elsewhere.
**✅ No live→settled regression:** on settle, `_convertLiveActivityGroupToSettled` removes the `data-live-*` attributes and the `liveAssistantTurn` id (ui.js:9932, 9990-9993), so the selectors STOP matching → the settled turn returns to normal transitions. The transition-kill is self-clearing, not sticky.
**✅ `!important` scoped:** confined to these live-turn selectors; doesn't leak into settled state or other UI. Codex's independent analysis reached the same conclusion (scoped, self-clearing on settle, theme transitions preserved).

### Note on the visible surface
The defect is a MOTION artifact (per-token fade flicker), which a static screenshot can't capture. The authoritative visible proof is the CSSOM check above (live-turn = instant paint, rest of UI keeps its transition) + the settle self-clear + Codex's scoping analysis. If Nathan wants a moving before/after (GIF) of a live stream on a light theme, that's a nice-to-have, but the mechanism is verified.

**Ready to merge — use branch `gate-rebase/5328-streaming-flicker` (`sha:7e334dcf`), NOT the PR's stale head `0c13a172`.** Clean rebase, CSS correctly scoped (CSSOM-verified), self-clearing on settle (no regression), suite 11503/0, Codex SAFE. concept 4/5 (fixes a real light-theme streaming flicker; matches how mature chat UIs paint streamed tokens instantly). Credit @ruizanthony (co-authored). crit=3. Author is lower-trust (prior #5077 critical-flaw), so I scrutinized extra — the fix is genuinely clean and minimal.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; scoping CSSOM-verified + settle self-clear confirmed + Codex SAFE. Cert valid for sha:7e334dcf._

nesquena-hermes:
Shipped in **v0.51.796**. Thanks @ruizanthony — scoping the transition opt-out to the live-turn selectors (verified against the real streaming DOM attributes) is exactly right, and theme-switch transitions elsewhere stay intact. Codex SAFE, full suite green.

Issue #4985, opened 2026-06-26
Orphan native-WebUI sessions (zero-message `state.db` rows) accumulate in sidebar and 404 on click — sibling of #4591


enihcam:
## Summary

Sidebar lists `source=webui` sessions that have no rows in `state.db.messages`. Clicking them hits `GET /api/session` → 404 → "Session not available in web UI." They never get pruned. This is the same root cause as #4591 (orphan CLI/API-server sidecars) but for the **native-webui** source — which is explicitly excluded from the existing orphan-prune pass.

## Repro

1. Open WebUI, let it create one or more sessions via the normal chat flow that get a `state.db.sessions` row written but never receive a persisted turn (e.g. boot-time `+` click, profile switch that resets the active id, sidebar nav that opens a session then closes the tab before the first message commits).
2. Look at the sidebar — the row is listed.
3. Click it.
4. **Expected:** Either the sidebar doesn't list the row, or the row is removed on 404.
5. **Actual:** Message pane shows "Session not available in web UI." indefinitely. The sidebar row stays. Reopening the same row, hard-refreshing, or restarting WebUI does not clear it.

## Evidence (from a local `state.db`)

```
Total sessions: 80
Zero-message rows: 5
```

All 5 are `source=webui` with `message_count=0` and zero rows in `messages`. Example:

| id | source | claimed | actual | started |
|---|---|---|---|---|
| `20260531_032708_9b227b` | webui | 0 | 0 | 2025-05-31 |
| `20260524_163005_50b00b` | webui | 0 | 0 | 2025-05-24 |
| `20260519_151936_b777e6` | webui | 0 | 0 | 2025-05-19 |
| `20260519_082021_e72a0b` | webui | 0 | 0 | 2025-05-19 |
| `20260519_081103_cb46a0` | webui | 0 | 0 | 2025-05-19 |

SQL:

```sql
SELECT s.id, s.source, s.message_count,
       (SELECT count(*) FROM messages m WHERE m.session_id = s.id) AS actual
FROM sessions s
ORDER BY s.started_at DESC;
```

## Root cause

`api/routes.py`, the orphan-prune block in `_build_session_list_cache_payload` (~lines 2081–2139):

```python
for s in webui_sessions:
    _sid = s.get("session_id")
    if (
        _sid
        and (is_cli_session_row(s) or _is_api_server_sidecar_row(s))
        and not _session_source_is_webui(s)        # ← excludes webui by design
        and _sid not in cli_by_id
[...]

PR #5333, opened 2026-07-01
fix(sessions): prune index-only ghost sessions (#5331)

### LINKED ISSUE #5331: Bulk "Untitled" sessions accumulate in sidebar and cannot be removed through any documented endpoint
## Summary

The WebUI sidebar shows a large number of sessions titled `"Untitled"` (201 of 214 in my install). These entries:

-   are not listed when I open the sidebar — they reappear after every server restart
-   cannot be deleted, archived, renamed, or otherwise removed through any API endpoint I tried
-   have no corresponding row in `state.db` and have no per-session `.json` file (in many cases)
-   are concentrated in workspace `/home/hermeswebui/workspace/default` and all share the same `created_at` day

All numbers below are observed on my install (Hermes WebUI container, uvicorn on `127.0.0.1:8787`, `HERMES_HOME=/home/hermeswebui`, queried 2026-07-01 ~08:30 UTC).

### Three independent "session lists" exist with minimal overlap

| Source | Count | Notes |
|---|---|---|
| `GET /api/sessions?sidebar_source=webui&exclude_hidden=1` | 214 | Returned by the API. **201 of 214 are titled `"Untitled"`.** |
| `/home/hermeswebui/sessions/_index.json` | 210 | 202 titled `"Untitled"`. File mtime keeps advancing (`stat` shows it updated during the session). |
| `ls /home/hermeswebui/sessions/*.json` | 247 | 130.5 MB total. |
| `SELECT count(*) FROM sessions` on `state.db` | 3 | `826436acb122`, `cb27055ddad8`, `ef62578aa72c` — the only three titled sessions in `state.db`. |

Set overlaps:

-   `api ∩ index = 209`
-   `api ∩ files = 13`
-   `index ∩ files = 9`
-   `files only = 233`

So **only ~13 sessions in the sidebar list have a corresponding per-session `.json` file**, and the three titled sessions in `state.db` are a strict subset of those.

### Untitled-session metadata

-   Workspace: all 201 Untitled sessions have `workspace = /home/hermeswebui/workspace/default`
-   `created_at`: all 201 share `created_at` clustered on **2026-07-01** (within the same second)
-   `message_count`: ranges 1 – 349
-   Models: mostly `MiniMax-M2.7-highspeed` and `MiniMax-M2.7`

### Named sessions (the 13 that survive)

| session_id | updated | msgs | title |
|---|---|---|---|
| `ef62578aa72c` | 2026-07-01 | 11 | \[REDACTED\] |
| `826436acb122` | 2026-07-01 | 632 | \[REDACTED\] |
| `cb27055ddad8` | 2026-07-01 | 6 | \[REDACTED\] |
| `8d4fc971f5cd` | 2026-06-26 | 25 | \[REDACTED\] |
| `2de477377b80` | 2026-06-26 | 84 | \[REDACTED\] |
| `f2a3309ee37f` | 2026-06-26 | 390 | \[REDACTED\] |
| `e63734f9bbd3` | 2026-06-25 | 488 | \[REDACTED\] |
| `0fd63d940b32` | 2026-06-24 | 194 | \[REDACTED\] |
| `815ffce4543b` | 2026-06-23 | 256 | \[REDACTED\] |
| `7b9398` | 2026-06-23 | 2524 | \[REDACTED\] |
| `ed81f0951740` | 2026-06-23 | 82 | \[REDACTED\] |
| `62ffb9cce087` | 2026-06-23 | 170 | \[REDACTED\] |
| `bcb230e19fa6` | 2026-06-16 | 71 | \[REDACTED\] |

All other sessions in the API response have `title = "Untitled"`. I did not manually rename any of them — they came that way.

## What was tried (none worked)

| Method | Endpoint / Body | Result | Notes |
|---|---|---|---|
| GET | `/api/sessions/{id}` | `404 {"error": "not found"}` | This route does not exist — session routes use singular path `/api/session/...` |
| DELETE | `/api/sessions/{id}` | `404 {"error": "not found"}` | Non-existent route |
| POST | `/api/sessions/{id}/archive` | `404 {"error": "not found"}` | Non-existent route |
| POST | `/api/sessions/{id}/title` | `404 {"error": "not found"}` | Non-existent route |
| PATCH | `/api/sessions/{id}` `{"archived":true}` | `404 {"error": "not found"}` | Non-existent route |
| POST | `/api/sessions/delete` | `404 {"error": "not found"}` | Non-existent route |
| POST | `/api/session/delete` `{"session_id":"..."}` | `404 {"error": "Session not found"}` | Route exists but `get_session()/Session.load()` require a file sidecar or in-memory entry — index-only ghost rows are invisible to it |
| POST | `/api/session/archive` `{"session_id":"..."}` | `404 {"error": "Session not found"}` | Same — requires backing file |
| POST | `/api/session/rename` `{"session_id":"...","title":"..."}` | `404 {"error": "Session not found"}` | Same |
| POST | `/api/sessions/cleanup` `{"confirm":true}` | `200 {"ok": true, "cleaned": 34}` | Returns cleaned count but the number of Untitled rows in the sidebar did not change |

For the `cleanup` endpoint specifically: it returned `"cleaned": 34`, but a follow-up `GET /api/sessions?sidebar_source=webui&exclude_hidden=1` immediately afterwards still returned 201 Untitled sessions. The count did not change.

## What I observed in the file system

While my session was running, `stat /home/hermeswebui/sessions/_index.json` showed the file mtime advancing on its own — i.e. the server process is rewriting the index in the background. I did not see any log line or notification when this happened.

Files of note:

```
/home/hermeswebui/sessions/_index.json                  (mtime advances continuously)
/home/hermeswebui/sessions/_index.json.bak              (224 entries — backup from before any edit)
/home/hermeswebui/sessions/_index.json.backup_*         (multiple backups, same ~330 KB size)
/home/hermeswebui/sessions/_pruned_webui_orphans.json   (2 entries)
/home/hermeswebui/sessions/*.json                       (247 files, 130.5 MB total — most have no API row)
/home/hermeswebui/state.db                              (SQLite, only 3 sessions — none Untitled)
```

## Root cause analysis

Four defects contribute to this symptom, confirmed by code inspection:

**1. Session-ID-based routes use singular `/api/session/...` — many of the 404s are HTTP-level.**

The API routes the reporter tried with plural `sessions` in the path (`/api/sessions/{id}`, `/api/sessions/{id}/archive`, `/api/sessions/{id}/title`, `/api/sessions/delete`) don't exist. All session-mutation routes use the singular form:

- `POST /api/session/delete` — `api/routes.py:12669`
- `POST /api/session/archive` — `api/routes.py:13623`
- `POST /api/session/rename` — `api/routes.py:12410`

However, even with the correct routes, ghost rows still 404 because `get_session(sid)` and `_get_or_materialize_session(sid)` only resolve from the in-memory `SESSIONS` dict or via `Session.load(sid)` (file read). An index-only row (present in `_index.json` but with no backing `<sid>.json` file and no row in `state.db`) is invisible to both paths — there is no fallback that consults the index.

**2. `/api/sessions/cleanup` is structurally blind to index-only ghost rows.**

The handler at `api/routes.py:17691-17711` iterates only `SESSION_DIR.glob("*.json")` (ignoring files starting with `_`):

```python
def _handle_sessions_cleanup(handler, body, zero_only=False):
    cleaned = 0
    for p in SESSION_DIR.glob("*.json"):
        if p.name.startswith("_"):
            continue
        s = Session.load(p.stem)
        if zero_only:
            should_delete = s and len(s.messages) == 0
        else:
            should_delete = s and s.title == "Untitled" and len(s.messages) == 0
        ...
    if SESSION_INDEX_FILE.exists():
        SESSION_INDEX_FILE.unlink(missing_ok=True)  # wipes the whole index
    return j(handler, {"ok": True, "cleaned": cleaned})
```

The 201 ghost rows are entries in `SESSION_INDEX_FILE` (`_index.json`) with `message_count: 1–349` that have no backing `.json` file on disk. The cleanup loop never sees them because:
- It iterates files, not index entries.
- Even if it saw them, the gate `s.title == "Untitled" and len(s.messages) == 0` requires zero messages — the ghosts have positive counts.
- The unconditional `SESSION_INDEX_FILE.unlink(missing_ok=True)` at the end wipes all index data, but `all_sessions()` re-builds the index from disk on the next `/api/sessions` poll. Ghost rows survive because they are re-read from the gateway metadata file (see item 4).

The `cleaned: 34` value refers to empty Untitled sidecar files on disk — orthogonal to the 201 index ghosts.

**3. Existing Untitled+0-message filters don't cover positive-count Untitled rows.**

The `all_sessions()` function at `api/models.py:4251-4257` filters out `Untitled ∧ 0-message ∧ ¬streaming ∧ ¬pending ∧ ¬worktree` rows. The reporter's ghosts have `message_count: 1–349`, so they pass the filter and stay visible.

The existing #4985 zero-message-orphan tombstone mechanism (`api/models.py:485–619`, plus `_prune_orphaned_webui_zero_message_sessions` at `api/routes.py:1796`) is specifically gated on `len(messages) == 0`. It was designed for a different orphan shape: native-WebUI rows that survived the #1171 keep-filter because they had a title or stale positive message count but whose actual state.db.messages was empty. The reporter's positive-count Untitled rows are a **new orphan variant** — rows whose message_count is non-zero but whose actual content is lost (no backing file, no state.db row).

**4. Index → cache thrash loop.**

The session-list cache source-stamp (`api/route_session_list_cache.py:387–399`) watches the index file's mtime. Every rebuild writes the index back (`api/models.py:4170, 4223, 4263`), which advances the mtime, invalidating the cache and triggering the next rebuild. Ghost rows in the index survive the disk round-trip (they live in the index, so the disk scan doesn't remove them), and the cycle repeats — hence the continuously-advancing mtime the reporter observed.

### Where the 201 ghost rows come from (likely source)

All 201 Untitled entries share `created_at` within a single second on 2026-07-01 — a batch-write signature, not gradual accumulation. The gateway metadata file at `$HERMES_HOME/sessions/sessions.json` is tracked by the cache source-stamp (`_session_list_cache_gateway_session_metadata_path`, `api/route_session_list_cache.py:69–86`) and is the most likely upstream source: if that file contains 201 stale Untitled rows, the `all_sessions()` recovery path at `api/models.py:4206–4225` would merge them into `_index.json` on every rebuild.

The `_pruned_webui_orphans.json` file contains only 2 entries — suggesting the #4985 tombstone mechanism ran only a few times before the ghost rows' positive message counts made them invisible to that prune.

To confirm: check `wc -l $HERMES_HOME/sessions/sessions.json` and inspect the first 3 rows for the same batch-created_at signature.

## Fix sketch

1. **Extend `_handle_sessions_cleanup`** (or add a dedicated bulk-prune endpoint) to scan `_index.json` entries matching `title == "Untitled"` AND no backing `<sid>.json` sidecar AND no matching `state.db.sessions` row, remove them from the index, and write them to a tombstone file analogous to `_pruned_webui_orphans.json` so `recover_missing_index_sidecars` doesn't re-add them.
2. **Extend the #4985 tombstone gate** to optionally accept "Untitled + positive-count + no-backing-file" orphans, not just zero-message ones.
3. **Optionally add** `?title=Untitled` or `?untitled_only=1` query filters to `GET /api/sessions` for self-service inspection, and a `POST /api/sessions/delete_untitled` bulk endpoint with `{"confirm": true}`.
4. **Stop the thrash:** if the index file's content is unchanged by a rebuild, skip the write so mtime doesn't advance.

## Environment

-   Container init: `su -s /bin/bash -c exec "/hermeswebui_init.bash" hermeswebui` (PID 1)
-   WebUI server: uvicorn, cwd `/app`, listening on `127.0.0.1:8787` (`HERMES_WEBUI_PORT=8787`)
-   `HERMES_HOME=/home/hermeswebui`
-   `get_active_hermes_home()` returns `/home/hermeswebui`
-   WebUI app code at `/app/` (`/app/server.py`, `/app/api/routes.py`, `/app/api/models.py`, `/app/api/config.py`, `/app/api/webui_session_db.py`)
-   Gateway package (a different code path) at `/home/hermeswebui/.local/lib/python3.12/site-packages/gateway/platforms/api_server.py`

No API key is configured: `config.yaml` has no `api_server_key`, env has no `HERMES_API_KEY` or `API_KEY`. All API calls are unauthenticated.

## Expected behavior

1.  **A working `DELETE /api/sessions/{id}` endpoint** that actually removes the session from the sidebar — or a correct existing endpoint documented for the purpose.
2.  **A bulk-delete parameter** on `GET /api/sessions` (e.g. `?untitled=1&confirm=true`) that wipes the Untitled rows in one call.
3.  **The `POST /api/sessions/cleanup` endpoint** to actually reduce the Untitled count when called.
4.  **A `title` filter** on `GET /api/sessions?title=Untitled` so I can at least see exactly what's there.
5.  **No index-thrash loop** on idle polls — the mtime of `_index.json` should stabilize when no state changes.

## Reproduction

1.  Start the WebUI container.
2.  Hit `GET /api/sessions?sidebar_source=webui&exclude_hidden=1`.
3.  Observe that the large majority of returned `sessions[*]` have `title == "Untitled"`.
4.  Pick any one of those Untitled session IDs and call `POST /api/session/delete` with `{"session_id": "..."}` — get `404 {"error": "Session not found"}`.
5.  Call `POST /api/sessions/cleanup` with `{"confirm": true}` — get `200 {"ok": true, "cleaned": N}`.
6.  Immediately re-issue `GET /api/sessions?sidebar_source=webui&exclude_hidden=1` — the number of Untitled rows has not changed.
7.  Observe that `stat /home/hermeswebui/sessions/_index.json` shows the mtime advancing on every poll even when no user action has occurred.

enihcam:
## Problem

Issue #5331: bulk "Untitled" ghost sessions accumulate in the sidebar with no backing .json file. They appear in _index.json as stale entries but cannot be removed — _handle_sessions_cleanup only checked file existence, missing index-only orphans entirely.

Additionally, the old code unconditionally deleted _index.json after cleanup, forcing a full rebuild from disk on every sidebar poll (cache-busting with zero benefit).

## Root Cause

The session write path can add entries to _index.json without writing the sidecar .json file. The existing cleanup had no Phase to sweep these index-only ghosts, and unconditional index deletion masked the problem while creating a performance issue.

## Fix

Rewrites _handle_sessions_cleanup (api/routes.py) into 3 phases:

1. Phase 1 (unchanged): file sweep — removes orphan .json files with zero-message-count entries
2. Phase 2 (new): index-only ghost sweep — removes stale _index.json entries that have no backing file and no in-memory session. Title-agnostic (catches multi-language "Untitled" variants per #5331). Uses the same atomic _safe_replace pattern as prune_session_from_index
3. Phase 3 (fixed): only deletes _index.json when Phase 1 actually removed files, avoiding unnecessary cache invalidation

## Verification

All 9 scenarios pass via direct inline Python verification:

- Phase 2 prunes index-only ghosts
- Phase 2 preserves in-memory (active) sessions
- Phase 2 writes cleaned index in-place when no files were removed
- Phase 3 deletes index only when Phase 1 removed files
- zero_only mode still runs Phase 2
- Edge cases: empty index, missing index, corrupt JSON, entries without session_id key

## Testing

Added tests/test_issue5331_index_only_ghost_cleanup.py with test coverage mirroring all verification scenarios. Follows existing test_issue4985_* conventions.

Closes #5331

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -18015,6 +18015,9 @@ def _handle_memory_read(handler, parsed=None):
 
 def _handle_sessions_cleanup(handler, body, zero_only=False):
     cleaned = 0
+    phase1_removed_ids = set()
+
+    # Phase 1: Clean orphan session files (existing behavior).
     for p in SESSION_DIR.glob("*.json"):
         if p.name.startswith("_"):
             continue
@@ -18029,10 +18032,87 @@ def _handle_sessions_cleanup(handler, body, zero_only=False):
                     SESSIONS.pop(p.stem, None)
                 p.unlink(missing_ok=True)
                 cleaned += 1
+                phase1_removed_ids.add(p.stem)
         except Exception:
             logger.debug("Failed to clean up session file %s", p)
+
+    phase1_touched = bool(cleaned)
+    phase2_rewrote_index = False
+
+    # Phase 2: Index-only ghost sweep (#5331).
+    # Remove index entries that have no backing .json file and no
+    # in-memory session.  These are orphaned rows left by a write path that
+    # updated _index.json without writing the session sidecar.
+    #
+    # Title-agnostic (#5331): a session with no backing file has no real
+    # data regardless of its title, so even a non-"Untitled" stub in the
+    # index is a ghost.  A legitimate session always has a sidecar file.
+    #
+    # Holds _INDEX_WRITE_LOCK for the full read-modify-write cycle to
+    # prevent races with concurrent Session.save() / prune_session_from_index().
     if SESSION_INDEX_FILE.exists():
+        try:
+            from api.models import _INDEX_WRITE_LOCK, _safe_replace
+
+            with _INDEX_WRITE_LOCK:
+                index_file_data = json.loads(
+                    SESSION_INDEX_FILE.read_text(encoding="utf-8")
+                )
+                if isinstance(index_file_data, list):
+                    live_ids = {
+                        p.stem
+                        for p in SESSION_DIR.glob("*.json")
+                        if not p.name.startswith("_")
+                    }
+                    with LOCK:
+                        in_memory_ids = set(SESSIONS.keys())
+
+                    survivors = []
+                    for entry in index_file_data:
+                        sid = entry.get("session_id")
+                        if not sid or sid in live_ids or sid in in_memory_ids:
+                            survivors.append(entry)
+                            continue
+                        # Phase 1 already removed the backing file for this
+                        # sid, so the index entry is stale too.  Drop it
+                        # from the index without double-counting.
+                        if sid in phase1_removed_ids:
+                            continue
+                        # Index-only ghost — no backing file, not in memory.
+                        cleaned += 1
+                        # Ghost not added to survivors — removed from index.
+
+                    if cleaned > 0 and len(survivors) < len(index_file_data):
+                        _tmp = SESSION_INDEX_FILE.with_suffix(
+                            f".tmp.{os.getpid()}.{threading.current_thread().ident}"
+                        )
+                        _payload = json.dumps(survivors, ensure_ascii=False, indent=2)
+                        try:
+                            with open(_tmp, "w", encoding="utf-8") as f:
+                                f.write(_payload)
+                                f.flush()
+                                os.fsync(f.fileno())
+                            _safe_replace(_tmp, SESSION_INDEX_FILE)
+                            phase2_rewrote_index = True
+                        except Exception:
+                            try:
+                                _tmp.unlink(missing_ok=True)
+                            except Exception:
+                                pass
+                            raise
+        except Exception:
+            logger.debug(
+                "Failed to clean up index-only session entries", exc_info=True
+            )
+
+    # Post-cleanup index invalidation.
+    # When Phase 1 removed files and Phase 2 didn't already clean the
+    # index, delete the index to force a fresh rebuild from disk on the
+    # next sidebar poll.  When Phase 2 succeeded the index is already
+    # correct, so keep it (avoids a wasteful rebuild).
+    if phase1_touched and not phase2_rewrote_index and SESSION_INDEX_FILE.exists():
         SESSION_INDEX_FILE.unlink(missing_ok=True)
+
     return j(handler, {"ok": True, "cleaned": cleaned})
 
 
--- a/tests/test_issue5331_index_only_ghost_cleanup.py
+++ b/tests/test_issue5331_index_only_ghost_cleanup.py
@@ -1,0 +1,419 @@
+"""Regression coverage for index-only ghost session cleanup (#5331).
+
+#5331: sessions created in ``_index.json`` without a backing ``.json`` sidecar
+file accumulate in the sidebar and cannot be removed via normal UI. The existing
+``_handle_sessions_cleanup`` only iterates ``SESSION_DIR.glob('*.json')`` —
+structurally blind to index-only entries.
+
+Phase 2 (new) reads the index and prunes entries that have no backing file
+and no in-memory session.  A legitimate session always has either a file on
+disk or an in-memory ``SESSIONS`` entry (or both).
+"""
+
+from __future__ import annotations
+
+import io
+import json
+import sys
+import threading
+from pathlib import Path
+
+import pytest
+
+REPO = Path(__file__).resolve().parent.parent
+if str(REPO) not in sys.path:
+    sys.path.insert(0, str(REPO))
+
+
+@pytest.fixture
+def mock_env(tmp_path, monkeypatch):
+    """Set up an isolated SESSION_DIR with monkeypatched globals.
+
+    Patches at the ``api.config`` level so that ``api.routes`` AND
+    ``api.models`` (used by ``Session.load`` in Phase 1) both read
+    the test temp paths.
+    """
+    import api.config as config_mod
+    import api.routes as routes
+    import api.models as models
+
+    sessions_dir = tmp_path / "sessions"
+    sessions_dir.mkdir()
+    index_file = sessions_dir / "_index.json"
+
+    # Patch at the config source so all importers see the same test paths.
+    monkeypatch.setattr(config_mod, "SESSION_DIR", sessions_dir)
+    monkeypatch.setattr(config_mod, "SESSION_INDEX_FILE", index_file)
+    monkeypatch.setattr(config_mod, "SESSIONS", {})
+    monkeypatch.setattr(routes, "LOCK", threading.Lock())
+    # Also refresh the module-level aliases in routes and models so they
+    # pick up the patched config values.
+    monkeypatch.setattr(routes, "SESSION_DIR", sessions_dir)
+    monkeypatch.setattr(routes, "SESSION_INDEX_FILE", index_file)
+    monkeypatch.setattr(routes, "SESSIONS", {})
+    monkeypatch.setattr(models, "SESSION_DIR", sessions_dir)
+    monkeypatch.setattr(models, "SESSION_INDEX_FILE", index_file)
+    return sessions_dir, index_file
+
+
+def _fake_handler():
+    """Minimal handler mock for ``_handle_sessions_cleanup``.
+
+    ``j()`` writes the JSON response to ``handler.wfile`` (a ``BytesIO``),
+    so callers read the result via ``handler.wfile.getvalue()``.
+    """
+    handler = type("FakeHandler", (), {})()
+    handler.wfile = io.BytesIO()
+    handler.send_response = lambda status: None
+    handler.send_header = lambda key, value: None
+    handler.end_headers = lambda: None
+    return handler
+
+
+def _fake_handler_result(handler):
+    """Parse the JSON written to ``handler.wfile`` by ``j()``."""
+    return json.loads(handler.wfile.getvalue())
+
+
+def _make_session_file(sessions_dir, sid, title="Untitled", **extra):
+    """Write a ``.json`` session file and return the compact entry dict.
+
+    Uses the minimal Session-serializable format: ``session_id`` is required,
+    all other fields have defaults.  ``**extra`` is folded directly into
+    the JSON so callers can set ``message_count``, ``created_at``, etc.
+    """
+    entry = {"session_id": sid, "title": title, **extra}
+    (sessions_dir / f"{sid}.json").write_text(json.dumps(entry))
+    return entry
+
+
+def _write_index(index_file, entries):
+    """Write the session index file."""
+    index_file.write_text(json.dumps(entries, indent=2))
+
+
+def _read_index(index_file):
+    """Read the session index file, returning empty list on missing/corrupt."""
+    if not index_file.exists():
+        return []
+    try:
+        return json.loads(index_file.read_text(encoding="utf-8"))
+    except Exception:
+        return []
+
+
+# ── Phase 2: index-only ghost sweep (#5331) ────────────────────────────────
+
+
+def test_cleanup_prunes_index_only_ghosts(mock_env):
+    """Index-only Untitled entries with no backing file are pruned."""
+    sessions_dir, index_file = mock_env
+
+    # Legitimate sessions with backing files — should survive.
+    _make_session_file(sessions_dir, "sess-a", "Legit")
+    _make_session_file(sessions_dir, "sess-b", "Production")
+
+    # Ghost entry: exists only in index, no backing file.
+    _write_index(index_file, [
+        {"session_id": "sess-a", "title": "Legit", "message_count": 5},
+        {"session_id": "sess-b", "title": "Production", "message_count": 42},
+        {"session_id": "sess-ghost-1", "title": "Untitled", "message_count": 3},
+    ])
+
+    import api.routes as routes
+    handler = _fake_handler()
+    routes._handle_sessions_cleanup(handler, {})
+    result = _fake_handler_result(handler)
+
+    assert result["ok"] is True
+    assert result["cleaned"] == 1  # only the ghost
+
+    updated = _read_index(index_file)
+    sids = {e["session_id"] for e in updated}
+    assert "sess-ghost-1" not in sids
+    assert "sess-a" in sids
+    assert "sess-b" in sids
+
+
+def test_cleanup_keeps_in_memory_ghosts(mock_env):
+    """Index-only entries still in memory (SESSIONS) are kept."""
+    sessions_dir, index_file = mock_env
+    import api.routes as routes
+
+    # One entry in-memory but no backing file — should survive Phase 2.
+    routes.SESSIONS["sess-mem"] = "placeholder"
+    _write_index(index_file, [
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Reading the diff at `api/routes.py:18016-18118` (the rewritten `_handle_sessions_cleanup`), the new test file, and the read-path in `api/models.py:4125-4257` on `origin/master`, this is a clean, well-scoped fix. A few notes, one of which is worth confirming before merge.

## What the PR gets right

Phase 2's ghost predicate matches the index invariant exactly: a live row must be either file-backed or in-memory, so a sid that is in `_index.json` but absent from both `SESSION_DIR.glob('*.json')` and `SESSIONS` is genuinely orphaned. Holding `_INDEX_WRITE_LOCK` across the read-modify-write and reusing `_safe_replace` mirrors `prune_session_from_index`, so it won't race concurrent `Session.save()`. The `phase1_removed_ids` set correctly avoids double-counting when a file removed in Phase 1 also had an index row. Phase 3 fixing the unconditional `SESSION_INDEX_FILE.unlink()` is a real win — that was the source of the index-thrash the reporter observed.

## Durability — no tombstone needed here (good)

I checked whether `all_sessions()`'s recovery pass could resurrect a pruned ghost. It can't: `recover_missing_index_sidecars` only re-adds sids in `missing_persisted_ids`, and those come from `persisted_ids` (`api/models.py:4189-4193`), which is `_persisted_session_ids_snapshot()` — a `glob('*.json')` scan. A fileless ghost is never in that set, and `Session.load_metadata_only()` returns `None` for it anyway. So unlike the #4985 zero-message orphans (which stay file-backed and *do* need `_pruned_webui_orphans.json` to avoid re-add), these fileless ghosts stay pruned without a tombstone. Worth a one-line code comment noting *why* no tombstone is required, since a future reader will wonder.

## The one thing to confirm: does this change the *served* sidebar?

The reporter's primary symptom is 201 Untitled rows **visible in `GET /api/sessions`**. But the serve-time filter at `api/models.py:4144-4155` already drops any index row that isn't file-backed or in-memory:

```python
index = [
    s for s in index
    if (
        str(s.get('session_id') or '') in in_memory_ids
        or (persisted_ids is not None
            and str(s.get('session_id') or '') in persisted_ids)
        or (persisted_ids is None
            and _index_entry_exists(s.get('session_id'), in_memory_ids=in_memory_ids))
    )
]
```

This runs before the #1171 Untitled+0-message keep-filter at `4251-4257` that the issue's RC3 focuses on. If the reporter's 201 rows are truly fileless, this filter should already exclude them from the served list regardless of `message_count` — which is in tension with them being visible. That points at one of: (a) the rows aren't purely fileless in the repro, or (b) the visible list is coming from a stale `route_session_list_cache` snapshot. Since Phase 2 only rewrites the on-disk `_index.json` (which is already filtered at serve time by `4144-4155`), it's worth verifying on a repro that this PR actually changes `GET /api/sessions` output — not just the on-disk file — and if the visible rows are cache-served, that the cleanup path invalidates `_SESSIONS_CACHE`.

## Scope note

This addresses the issue's RC2 (cleanup blind to index-only rows) and RC4 (thrash), which is the right minimal cut. It intentionally does not touch RC1 (the singular `/api/session/...` vs. the plural paths the reporter tried, and `get_session()` not resolving index-only rows) or the requested documented bulk-delete endpoint. Fine to leave those for a follow-up, but the issue shouldn't be auto-closed as fully resolved if the served-sidebar question above turns out to need a cache-invalidation touch.

## Verification

The added `tests/test_issue5331_index_only_ghost_cleanup.py` covers the ghost/legit/in-memory matrix, corrupt/missing index, no-double-count, and zero-only mode — assertions line up with the phased behavior. Adding one integration-style assertion that a `GET /api/sessions` after cleanup no longer returns the ghost sid would close the serve-path gap above.

nesquena-hermes:

**Certified head:** `sha:c13469bb` (clean rebase, branch `gate-rebase/5333-ghost-session-prune`) · **PR:** #5333 · _enihcam, prune index-only ghost sessions (#5331)_
**Verdict:** Full gate GREEN. A well-built session-index integrity fix: sweeps index-only ghost entries (rows with no backing `.json` + not in memory) under proper locking, with a conservative invariant that never prunes a real session. The prior `_INDEX_WRITE_LOCK` concurrency finding is resolved. Ready to merge from the rebase branch.

### What I ran (rebased worktree `/tmp/wt-rebase-5333`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ merge-base == master tip (`56d1f619`); `git apply` clean |
| Full pytest suite | **11527 passed**, 0 failed |
| PR's own test file | ✅ 13/13 (ghost-removed, real-preserved, mixed, corrupt/empty resilience, no-double-count) |
| Concurrency + invariant read (mine) | ✅ verified below |

### Findings
**✅ Concurrency correct (prior finding resolved):** Phase 2's full read→compute→`_safe_replace` cycle is wrapped in `with _INDEX_WRITE_LOCK:` (routes.py:18057), with the in-memory read nesting `with LOCK:`. Lock ordering is outer `_INDEX_WRITE_LOCK` → inner `LOCK` (consistent, no deadlock vs `Session.save()`/`prune_session_from_index()`). The index rewrite is atomic: temp file + `flush` + `os.fsync` + `_safe_replace`, with temp cleanup on failure (no partial index).

**✅ Ghost detection is conservative — NEVER prunes a real session:** an index entry survives if `sid in live_ids` (has a `.json` sidecar) OR `sid in in_memory_ids` (in `SESSIONS`). Only entries with **no backing file AND not in memory** are dropped. So a legitimate session (backed, in-memory, or mid-save-in-memory) is never pruned — verified by `test_cleanup_keeps_in_memory_ghosts` + `test_cleanup_mixed_ghosts_and_legit`. Also handles `phase1_removed_ids` (no double-count) and title-agnostic detection (a backed stub survives; only unbacked rows are ghosts).

**✅ Resilient:** corrupt/empty/missing index don't crash (try/except + isinstance guards); Phase 3 index-delete-for-rebuild fallback when Phase 2 couldn't rewrite. 3-phase structure (orphan-file clean → index-ghost sweep → gated index-unlink) is coherent.

**Ready to merge — use branch `gate-rebase/5333-ghost-session-prune` (`sha:c13469bb`), NOT the PR's stale head `161e4173`.** Clean rebase, Codex SAFE, suite 11527/0, concurrency lock verified, and the critical invariant (real sessions never pruned) is held + tested (13/13). Non-visible backend/session-index fix — no screenshot. concept 4/5 (index-only ghost rows are a real #5331 cleanup need). Credit @enihcam (co-authored). crit=3 (session-index integrity — a wrong prune would be data loss, and the guard correctly prevents it).

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; concurrency lock + conservative-ghost-detection invariant verified (real sessions survive), 13/13 own tests + suite 11527/0 + Codex SAFE. Cert valid for sha:c13469bb._

nesquena-hermes:
Shipped in **v0.51.794**. Thanks @enihcam — the Phase-2 index-only ghost sweep (race-safe under `_INDEX_WRITE_LOCK`, title-agnostic, atomic rewrite, no data loss) is exactly right. Codex gate SAFE, full suite green.

PR #5343, opened 2026-07-01
fix(webui): handle stale clarify polling after profile switch

ruizanthony:
## Summary

- handle the stale clarify polling case that can happen immediately after switching profiles
- treat `404 Session not found` from the old session as a terminal stale-session response, not as a missing clarify endpoint
- stop the old clarify poll and hide the stale card instead of telling the user to restart the server

## Why

When the user switches profiles, the old profile's active session can briefly keep polling `/api/clarify/pending` while subsequent requests already carry the new profile cookie. The backend correctly answers `404 Session not found`, but the frontend currently treats any 404 as a missing clarify endpoint and surfaces a misleading restart warning.

## Validation

- private/site marker scan on the PR diff
- `pytest -q tests/test_clarify_sse.py tests/test_clarify_unblock.py tests/test_4504_clarify_stuck_on_expiry.py tests/test_1466_sidebar_cancel_clarify.py tests/test_issue2883_clarify_padding.py` — 62 passed

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -7528,7 +7528,18 @@ function _startClarifyFallbackPoll(sid) {
       else { _clearClarifyPendingForSession(sid); _hideClarifyCardIfOwner(sid, false, 'expired'); }
     } catch(e) {
       const msg = String((e && e.message) || "");
-      if (!_clarifyMissingEndpointWarned && /(^|\b)(404|not found)(\b|$)/i.test(msg)) {
+      const status = e && e.status;
+      // A profile switch can leave an old-profile session polling briefly under
+      // the new profile cookie.  The backend correctly returns 404 "Session not
+      // found" in that case; it is not a missing clarify endpoint and should not
+      // tell the user to restart the server.
+      if (status === 404 && /session\s+not\s+found/i.test(msg)) {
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Thanks @ruizanthony — this correctly identified the profile-switch stale-session case. I've opened #5346 which builds on your approach to fully resolve #5345 (both symptoms the issue raises), and you're credited as `Co-authored-by` on the commit.

The one thing #5346 extends: this PR keeps the broad message match in the missing-endpoint branch —

```js
if (!_clarifyMissingEndpointWarned && (status === 404 || /(^|\b)(404|not found)(\b|$)/i.test(msg))) {
```

Since `/api/clarify/pending` **always returns HTTP 200** when the route is present (it answers `{"pending": null}` for an unknown session and never 404s — confirmed with `curl`), the `|| /...not found.../i` fallback can still surface a false "restart server" toast on *another* unrelated stale-session 404 whose message contains "not found". #5346 drops the message-scrape entirely and gates the warning strictly on `status === 404` + a route-not-found body that is **not** session-scoped, plus adds the interrupt-provenance logging the issue also asks for.

Closing this in favor of #5346 — your fix is preserved and credited there. 🙏

PR #5344, opened 2026-07-01
perf: GPU-accelerate thinking card text rendering

Pythonming2020:
Add will-change, GPU composition layer (translateZ(0)), antialiased font smoothing, and optimizeLegibility to the .thinking-card-body pre element to reduce repaint flicker during streaming thought text updates.

The pre textContent is rewritten on every reasoning token during streaming. Promoting it to its own compositor layer avoids full-document layout and repaint on each text update, making the streaming thought display significantly smoother.

### DIFF
--- a/api/history_suggestions.py
+++ b/api/history_suggestions.py
@@ -1,0 +1,190 @@
+"""History prompt suggestions for the WebUI composer.
+
+Scans WebUI session JSON files for past user messages and returns
+matches based on partial text input. Uses a simple LRU cache so
+repeated queries don't re-scan the full session directory every time.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import time
+from pathlib import Path
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+_MAX_SUGGESTIONS = 8
+_MIN_QUERY_LENGTH = 1
+_CACHE_TTL_SECONDS = 30  # rebuild cache every 30s at most
+
+# ── In-memory cache ──────────────────────────────────────────────────────────
+_cache: dict[str, Any] = {
+    "user_messages": [],       # list of {"text": str, "session_id": str, "session_title": str, "ts": float}
+    "built_at": 0.0,
+    "session_dir_mtime": 0.0,
+}
+
+
+def _session_dir_mtime(session_dir: Path) -> float:
+    """Return the latest mtime across all session JSON files."""
+    latest = 0.0
+    if not session_dir.exists():
+        return latest
+    try:
+        for p in session_dir.glob("*.json"):
+            if p.name == "_index.json":
+                continue
+            try:
+                m = p.stat().st_mtime
+                if m > latest:
+                    latest = m
+            except OSError:
+                continue
+    except Exception:
+        pass
+    return latest
+
+
+def _rebuild_cache(session_dir: Path) -> None:
+    """Scan all session JSON files and extract user messages."""
+    messages: list[dict[str, Any]] = []
+    if not session_dir.exists():
+        _cache["user_messages"] = messages
+        _cache["built_at"] = time.time()
+        _cache["session_dir_mtime"] = _session_dir_mtime(session_dir)
+        return
+
+    seen: set[str] = set()
+    for p in sorted(session_dir.glob("*.json"), key=lambda x: x.stat().st_mtime, reverse=True):
+        if p.name == "_index.json":
+            continue
+        try:
+            data = json.loads(p.read_text(encoding="utf-8", errors="replace"))
+        except (json.JSONDecodeError, OSError):
+            continue
+        if not isinstance(data, dict):
+            continue
+
+        session_id = str(data.get("session_id") or p.stem)
+        session_title = str(data.get("title") or "Untitled")
+        raw_messages = data.get("messages") or []
+        if not isinstance(raw_messages, list):
+            continue
+
+        for msg in raw_messages:
+            if not isinstance(msg, dict):
+                continue
+            role = msg.get("role", "")
+            if role != "user":
+                continue
+            content = msg.get("content", "")
+            if not content or not isinstance(content, str):
+                continue
+            text = content.strip()
+            if not text or len(text) < 2:
+                continue
+
+            # Deduplicate identical messages across sessions
+            dedup_key = text.lower()[:120]
+            if dedup_key in seen:
+                continue
+            seen.add(dedup_key)
+
+            ts = 0.0
+            try:
+                ts = float(msg.get("timestamp", 0) or 0)
+            except (TypeError, ValueError):
+                ts = 0.0
+
+            messages.append({
+                "text": text,
+                "session_id": session_id,
+                "session_title": session_title,
+                "ts": ts,
+            })
+
+    _cache["user_messages"] = messages
+    _cache["built_at"] = time.time()
+    _cache["session_dir_mtime"] = _session_dir_mtime(session_dir)
+    logger.debug(
+        "History suggestions cache rebuilt: %d messages from %s",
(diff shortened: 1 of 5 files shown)


nesquena-hermes (changes requested):
Thanks for the contribution! The actual GPU-acceleration change here — the `will-change: contents; transform: translateZ(0); -webkit-font-smoothing: antialiased; text-rendering: optimizeLegibility;` additions on `.thinking-card-body pre` in `static/style.css` — is reasonable and matches the PR title. Unfortunately the branch also bundles several unrelated files, one of which breaks the test suite, so I can't take it as-is. Requesting changes:

### 1. The new test is red (blocks CI) — must fix or remove

`tests/test_thread_env_cwd.py` calls:

```python
env = _build_agent_thread_env(
    { ... },
    workspace="/tmp/webui-workspace",
    session_id="session-123",
    hermes_home="/tmp/hermes-home",   # <-- wrong kwarg
)
```

but the function signature on `master` (and on this branch — the PR doesn't touch `api/streaming.py`) is:

```python
def _build_agent_thread_env(profile_runtime_env, workspace, session_id, profile_home):
```

The parameter is `profile_home`, not `hermes_home`, so the test fails immediately:

```
TypeError: _build_agent_thread_env() got an unexpected keyword argument 'hermes_home'
```

(Reproduced by running `pytest tests/test_thread_env_cwd.py` on your branch head.) The two assertions `HERMES_HOME == "/tmp/hermes-home"` and `HERMES_SESSION_KEY == "session-123"` also need to be checked against what `_build_agent_thread_env` actually emits. Either fix the call/assertions, or drop this test — it's unrelated to the CSS change anyway (see #2).

### 2. Scope: this PR bundles four unrelated changes under a CSS-perf title

The title is "perf: GPU-accelerate thinking card text rendering" but the diff contains:

- `static/style.css` (+1/-1) — **the actual change, keep this.**
- `api/history_suggestions.py` (+190) — a whole new history-suggestions backend module that is **never imported or wired anywhere** (`grep -rn history_suggestions api/ static/ server.py` = 0 hits outside the file itself). Dead code.
- `fix-firefox-scroll-jump.patch` (+141) — a raw `git format-patch` file committed into the repo root. This looks like an accidental commit; it shouldn't be in the tree at all.
- `static/clear-model-state.html` (+15) — a one-off `localStorage.removeItem(...)` debug page. Also looks accidental.
- `tests/test_thread_env_cwd.py` (+31) — the broken test from #1, unrelated to CSS.

**Please split this into a clean PR containing only the `static/style.css` GPU-accel change** (that one is genuinely mergeable). Drop the `.patch` file and `clear-model-state.html` entirely. If `api/history_suggestions.py` and the Firefox scroll-jump / thread-env work are things you want to land, open separate focused PRs where they're actually wired up + tested.

Once it's just the CSS change (or the CSS change plus properly-wired, tested versions of the others), this is easy to review. Appreciate the perf instinct on the thinking card.

nesquena-hermes:
Thanks for the PR. I pulled the branch into a read-only worktree and read the actual diff plus the streaming render path on `origin/master`. There's a genuine perf idea here, but the change as written won't deliver it, and the PR bundles four unrelated files that should not ride along.

## What the diff actually does

The only change matching the title is one line in `static/style.css:6017`:

```css
.thinking-card-body pre { ...; will-change: contents; transform: translateZ(0); -webkit-font-smoothing: antialiased; text-rendering: optimizeLegibility; }
```

The hot path this targets is `_renderThinkingInto()` in `static/ui.js:15723-15731`, which rewrites the whole element on every reasoning token:

```js
const pre=row.querySelector('.thinking-card-body pre');
if(pre){ pre.textContent=clean; return; }
```

Because `textContent` is replaced wholesale each token, the `pre` must re-layout and re-paint every update regardless of layering. That's the constraint the CSS has to work around, and two of the four properties work against it:

- `will-change: contents` does **not** promote a compositor layer — per the spec, `contents` only hints the contents will change and creates no layer. To force layer promotion you'd use `will-change: transform` (or the `translateZ(0)` you already added, making `will-change: contents` redundant here).
- `text-rendering: optimizeLegibility` forces kerning + ligature shaping on every layout pass. On an element whose text is rewritten per token, that's added CPU cost on the exact path you're trying to speed up — counterproductive.
- `transform: translateZ(0)` does force a layer, but since the text changes every token the layer texture is re-rasterized and re-uploaded each token, so it moves the paint onto the GPU rather than eliminating it. Net effect is at best neutral, plausibly worse once the legibility pass is added.

## The bundling problem

`git diff origin/master...HEAD --stat` shows five files, four unrelated to the title:

- `api/history_suggestions.py` (190 lines) — imported nowhere in `api/` or `server.py`; dead code as shipped.
- `fix-firefox-scroll-jump.patch` — a raw `.patch` committed to the repo root, never applied to `static/ui.js`. A patch file checked into the tree changes nothing at runtime; the fix would need to be applied as a real diff to `ui.js`.
- `static/clear-model-state.html` — clears `localStorage` and redirects on any navigation, with no confirmation.
- `tests/test_thread_env_cwd.py` — unrelated to rendering.

## Recommendation

1. Split this into one PR per concern. The three non-rendering files each need their own review (especially the unwired module and the unapplied patch).
2. For the rendering PR, if you want a real win the lever is the update *frequency*, not the layer: coalesce token writes with `requestAnimationFrame` in `_renderThinkingInto` so the `pre` paints once per frame instead of once per token. Pair that with `contain: content` on `.thinking-card-body pre` to scope re-layout to the card. Drop `will-change: contents` and `optimizeLegibility`.
3. Include a before/after measurement (a DevTools Performance capture of paint/layout time during a streaming reasoning burst) so the perf claim is grounded.

Happy to re-review once it's a focused CSS/JS change with a measurement.

nesquena-hermes:
## Closing — the packaged diff can't ship, and the one on-title change is marginal

Thanks @Pythonming2020 for looking at streaming-render performance. After running the salvage screen we're closing this without merging or pulling a sub-piece. Honest reasoning per file:

### 1. `static/style.css` (the titular GPU-accel CSS) — marginal, doesn't clearly land
The `.thinking-card-body pre` GPU-layer hints (`will-change`/`translateZ(0)`/font-smoothing) are the one on-topic change, but the measured win here is negligible — the thinking card already renders fine, and `will-change: contents` in particular is linter-flagged and can *increase* memory pressure rather than reduce repaint cost. A perf change to the crown-jewel streaming render needs a before/after measurement showing a real improvement; as-is it "won't deliver" a perceptible gain.

### 2. `api/history_suggestions.py` (+190) — orphaned dead code
A whole new backend module that nothing imports or wires up (`grep -rn history_suggestions api/ static/ server.py` = 0 hits outside itself). Unrelated to the CSS perf title.

### 3. `fix-firefox-scroll-jump.patch` (+141) — a raw git patch file committed into the repo root
Not source — an accidental artifact.

### 4. `static/clear-model-state.html` (+15) — a one-off debug page
A localStorage-clearing utility page, not a product surface.

### 5. `tests/test_thread_env_cwd.py` (+31) — dead on arrival
Calls `_build_agent_thread_env(hermes_home=...)` but the real parameter is `profile_home` on both master and this branch, and the PR doesn't touch `streaming.py` — so the test errors on collection (CI red).

### Why close rather than salvage
The single on-topic change (the CSS) is marginal-and-unmeasured, and it's bundled with a dead module + a committed patch file + a debug page + a broken test. There's no atomic piece worth pulling standalone.

### Path forward
If you want to pursue thinking-card streaming perf, a small PR scoped to *just* a measured CSS change — with a before/after (e.g. a DevTools performance capture or a repaint-count comparison) showing the improvement — would be reviewed on its merits. Happy to look at that. Thanks for contributing, and no hard feelings on this one.

PR #5346, opened 2026-07-01
fix(webui): stop false clarify-unavailable toast; add interrupt provenance (#5345)

### LINKED ISSUE #5345: Bug: Misleading clarify endpoint toast and unclear interrupt provenance
## Summary

In a self-hosted Hermes WebUI deployment, we are seeing two symptoms that appear related to chat/stream lifecycle handling:

1. The browser sometimes shows a toast/status equivalent to:

   ```text
   Clarify endpoint unavailable. Please restart server.
   ```

   even though the deployed server exposes `/api/clarify/pending` and current polling for that endpoint returns HTTP 200.

2. Some tool executions settle as interrupted, with backend logs showing real interrupt/SIGINT-style behavior such as:

   ```text
   Interrupted during API call.
   Tool terminal returned error ... "[Command interrupted]", exit_code: 130
   ```

This issue is primarily to ask whether the frontend clarify polling and stream cancellation behavior are currently expected, and to get maintainer guidance before proposing a patch.

### Clarify endpoint toast

Runtime logs show successful clarify polling for the active WebUI session:

```text
GET /api/clarify/pending?session_id=<session-id> status=200
```

The server includes the clarify endpoint route:

```python
if parsed.path == "/api/clarify/pending":
    return _handle_clarify_pending(handler, parsed)
```

Public source reference:

- `api/routes.py`: https://github.com/nesquena/hermes-webui/blob/56d1f6199c4204f07ff4a0f34da2748b8b418834/api/routes.py#L11581-L11582

Despite that, the UI can display a missing-endpoint message. The current frontend fallback appears to infer endpoint absence from a broad error-message match:

```js
const data = await api("/api/clarify/pending?session_id=" + encodeURIComponent(sid), {timeoutToast:false});
...
} catch(e) {
  const msg = String((e && e.message) || "");
  if (!_clarifyMissingEndpointWarned && /(^|\b)(404|not found)(\b|$)/i.test(msg)) {
    _clarifyMissingEndpointWarned = true;
    setComposerStatus("Clarify unavailable on current server build. Restart server.");
    showToast("Clarify endpoint unavailable. Please restart server.", 5000);
    stopClarifyPolling();
  }
}
```

Public source reference:

- `static/messages.js`: https://github.com/nesquena/hermes-webui/blob/56d1f6199c4204f07ff4a0f34da2748b8b418834/static/messages.js#L7525-L7538

In the same browser session, there are also occasional 404s for nearby session lifecycle polling, for example:

```text
/api/session/status?session_id=<stale-or-unknown-session-id> status=404
```

Our thesis is that the clarify toast can be a false positive if an unrelated stale-session or stale-client error is surfaced through a generic `404` / `not found` message, or if the browser is running a stale bundle against a newer server.

### Interrupted tool turns

Separately, logs show actual interrupt/cancel paths rather than a generic terminal tool failure:

```text
Interrupted during API call.
Tool terminal returned error ... "[Command interrupted]", exit_code: 130
```

Because exit code 130 is the normal SIGINT/Ctrl-C style path, the immediate cause seems to be cancellation/interrupt propagation while a model call or tool subprocess is in flight.

The open question is whether passive UI lifecycle events can currently propagate interrupts too broadly. Examples worth ruling in or out:

- switching sessions while a run is active,
- browser refresh or tab navigation,
- stream reconnect/recovery,
- stale stream cleanup,
- starting another request while a prior stream is active,
- explicit Stop/cancel, which should continue to interrupt.

## Expected behavior

1. The UI should only show “Clarify endpoint unavailable” when the request to `/api/clarify/pending` itself actually receives a missing-endpoint response, not from arbitrary `404` / `not found` text in a catch block.

2. Clarify polling failures should be observable enough to tell what failed:
   - failed URL/path,
   - HTTP status if available,
   - polling session id,
   - current active session id.

3. Tool subprocesses should be interrupted for explicit user cancellation/Stop, but passive UI lifecycle actions should ideally detach/reconnect/settle the stream without sending SIGINT to the running tool or provider call unless that is intentional design.

## Why this matters

The current user-facing symptoms point operators toward the wrong layer:

- “Clarify endpoint unavailable” suggests a server build/deploy mismatch, but the endpoint is present and returning 200.
- “Tool interrupted” can look like a tool executor bug, but logs show real interrupt propagation. Without clearer source attribution, it is hard to tell whether the interrupt was explicit user intent or a side effect of WebUI stream lifecycle handling.

## Maintainer questions

1. Is the broad clarify fallback (`404|not found` against the caught error message) intentional, or would maintainers accept narrowing it to an actual `/api/clarify/pending` 404?

2. Is there an existing preferred client-side API error shape that exposes status/path cleanly to this code path, or should this path add local parsing/logging?

3. For stream lifecycle: should passive session switches, browser refreshes, or stale stream cleanup ever interrupt the backend agent/tool run, or should only explicit Stop/cancel paths do that?

4. Would maintainers prefer this split into two issues?
   - one for misleading clarify endpoint UX,
   - one for interrupt provenance / over-broad cancellation tracing.

## Possible direction, pending maintainer feedback

For clarify polling:

- Narrow the missing-endpoint toast to a real 404 from `/api/clarify/pending`.
- Avoid matching arbitrary `not found` text.
- Log clarify polling failures with request path, status, polling session id, and current session id.

For interrupted tool turns:

- Add trace/debug events around the source of stream cancellation:
  - explicit Stop,
  - session switch,
  - page unload/refresh,
  - stale stream cleanup,
  - new turn conflict.
- Ensure only explicit cancellation paths send an interrupt to active model/tool execution, if that matches intended design.

## Additional notes

This was observed on a self-hosted deployment with WebUI and Hermes Agent running together. The issue report intentionally avoids private deployment details; the key public evidence is the mismatch between successful `/api/clarify/pending` polling and the frontend’s broad missing-endpoint fallback, plus backend interrupt logs showing SIGINT-style tool cancellation.

nesquena-hermes:
Fixes #5345. Supersedes #5343 (credited below).

## Problem

Issue #5345 reports two related symptoms in a self-hosted deployment:

1. **Misleading clarify toast.** The UI shows *"Clarify endpoint unavailable. Please restart server."* even though `/api/clarify/pending` is present and returning HTTP 200 on every request.
2. **Unclear interrupt provenance.** Tool turns settle as interrupted (`exit_code: 130`, SIGINT), and it's hard to tell whether that was an explicit Stop or a side effect of stream-lifecycle handling.

## Root cause (Part 1)

`_handle_clarify_pending` **always returns HTTP 200** — it answers `{"pending": null}` for an unknown session and never returns 404 (verified: `curl /api/clarify/pending?session_id=bogus` → `200 {"pending": null}`). But the front-end poll's catch block fired the restart-server toast on **any** caught error whose message matched the broad regex `/(^|\b)(404|not found)(\b|$)/i`. So the toast could only ever mis-fire — triggered by an *unrelated* 404 leaking into the shared error path, e.g. a stale `Session not found` (contains "not found") from an old-profile session still polling briefly after a profile switch.

## Fix

**Part 1 — clarify toast (`static/messages.js`):** branch on the structured HTTP status that `api()` attaches to the thrown `Error` (`err.status`) instead of scraping the message string:
- `404 Session not found` → treated as a stale-session poll (stop polling + hide the card silently), no toast.
- The restart-server warning now fires **only** on a genuine route-not-found 404 (`{"error":"not found"}`) whose body is *not* session-scoped.
- Poll failures are logged with request path, HTTP status, polling session id, and current session id (maintainer question #2).

**Part 2 — interrupt provenance (`static/boot.js` + explicit call sites):** I traced every cancel path. Passive lifecycle events (session switch via `loadSession`, `beforeunload`, `visibilitychange`) call **local-only `closeLiveStream()`** and never hit `/api/chat/cancel` — so they already do **not** interrupt the backend run (answering maintainer question #3: only explicit Stop/cancel interrupts). To make cancellation observable, `cancelStream()`/`cancelSessionStream()` now log `[stream] cancel requested` with a `reason`, and every explicit call site passes a distinct reason (`composer-stop`, `slash-stop`, `slash-interrupt`, `busy-interrupt`, `sidebar-stop`).

## Relationship to #5343

@ruizanthony's #5343 fixed only the profile-switch stale-session sub-case and **kept the broad message-scrape** in the missing-endpoint branch (`status===404 || /...not found.../i`), so that branch could still over-fire on other stale 404s. This PR handles the full issue (both symptoms) and gates the warning strictly on structured status + non-session body. Credited via `Co-authored-by`.

## Validation

- `node --check` on all 4 touched static files ✅
- `scripts/scope_undef_gate.py` → CLEAN ✅
- New `tests/test_issue5345_clarify_toast_and_interrupt_provenance.py` — 9 tests, all pass; verified they **fail 8/9 against pre-fix master** (not vacuous). The 9th (`test_clarify_pending_never_404s`) locks the backend handler shape the whole fix depends on.
- Existing clarify suite green: `test_clarify_sse.py test_clarify_unblock.py test_4504_clarify_stuck_on_expiry.py test_1466_sidebar_cancel_clarify.py test_issue2883_clarify_padding.py test_messages_stale_stream_source_scope.py` → 65 passed.

Reported by @claw-io.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,13 @@
 
 ## [Unreleased]
 
-_No unreleased changes. Entries are moved into their version block when a release is tagged._
+### Fixed
+
+- **The "Clarify endpoint unavailable. Please restart server." toast no longer fires falsely.** `/api/clarify/pending` always returns HTTP 200 when present (it answers `{"pending": null}` for an unknown session — it never 404s), yet the front-end poller warned "restart the server" on *any* caught error whose message merely contained "404" or "not found". An unrelated stale-session 404 (`Session not found`, e.g. an old-profile session still polling briefly after a profile switch) or a transient error therefore surfaced a misleading missing-endpoint toast pointing operators at the wrong layer. Clarify polling now branches on the structured HTTP status (`err.status`) that `api()` attaches: a `404 Session not found` is handled as a stale-session poll (stop + hide the card silently), and the restart-server warning fires only on a genuine route-not-found 404 whose body is not session-scoped. Poll failures are also logged with the request path, HTTP status, polling session id, and current session id for diagnosis. Thanks @claw-io for the report and @ruizanthony for the initial profile-switch fix. (#5345, absorbs #5343)
+
+- **Stream cancellation now records its provenance.** `cancelStream()` / `cancelSessionStream()` log a `[stream] cancel requested` line with the trigger (`composer-stop`, `slash-stop`, `slash-interrupt`, `busy-interrupt`, `sidebar-stop`) so a backend interrupt (SIGINT / exit code 130) can be attributed to an explicit user action. Passive UI lifecycle events (session switch, tab hide, page unload) continue to tear down only the local SSE transport and never send an interrupt to the running agent or tool — only explicit Stop/interrupt paths do. (#5345)
+
+_Entries are moved into their version block when a release is tagged._
 
 ## [v0.51.792] — 2026-07-01
 
--- a/static/boot.js
+++ b/static/boot.js
@@ -14,10 +14,21 @@
 // cancelStream: stop the active chat stream.
 // See docs/rfcs/webui-run-state-consistency-contract.md (Invariants #2, #4)
 // for the owner-aware + terminal-settle rationale.
-async function cancelStream(){
+async function cancelStream(reason){
   const sid = S.session && S.session.session_id;
   const streamId = S.activeStreamId;
   if(!streamId) return;
+  // Interrupt provenance: log WHY the active run is being cancelled so operators
+  // can tell an explicit Stop / interrupt from any other trigger when they see a
+  // SIGINT/exit-code-130 in the backend logs. Only explicit user paths reach
+  // this function (Stop button, /stop, /interrupt, busy-interrupt); passive
+  // lifecycle events — session switch, tab hide, page unload — tear down the
+  // LOCAL SSE transport via closeLiveStream() and never call /api/chat/cancel,
+  // so they never interrupt the backend agent/tool run. (#5345)
+  const _reason = reason || 'explicit-cancel';
+  if(typeof console !== 'undefined' && console.info){
+    console.info('[stream] cancel requested', {reason:_reason, streamId, sessionId:sid});
+  }
   let respBody=null;
   try{
     const r=await fetch(new URL(`api/chat/cancel?stream_id=${encodeURIComponent(streamId)}`,document.baseURI||location.href).href,{credentials:'include'});
@@ -54,6 +65,11 @@ async function cancelSessionStream(session){
   const streamId = session&&session.active_stream_id;
   const sid = session&&session.session_id;
   if(!streamId||!sid) return;
+  // Explicit sidebar "Stop response" — log provenance for the same reason as
+  // cancelStream(). (#5345)
+  if(typeof console !== 'undefined' && console.info){
+    console.info('[stream] cancel requested', {reason:'sidebar-stop', streamId, sessionId:sid});
+  }
   try{
     await fetch(new URL(`api/chat/cancel?stream_id=${encodeURIComponent(streamId)}`,document.baseURI||location.href).href,{credentials:'include'});
   }catch(e){/* close local stream; keep UI state honest below */}
--- a/static/commands.js
+++ b/static/commands.js
@@ -1211,7 +1211,7 @@ async function cmdPersonality(args){
 async function cmdStop(){
   if(!S.session){showToast(t('no_active_session'));return;}
   if(!S.activeStreamId){showToast(t('no_active_task'));return;}
-  if(typeof cancelStream==='function'){await cancelStream();showToast(t('stream_stopped'));}
+  if(typeof cancelStream==='function'){await cancelStream('slash-stop');showToast(t('stream_stopped'));}
   else showToast(t('cancel_unavailable'));
 }
 
@@ -1317,7 +1317,7 @@ async function cmdInterrupt(args){
   updateQueueBadge(S.session.session_id);
   S.pendingFiles=[];renderTray();
   // Cancel the active stream; setBusy(false) will drain the queue
-  if(typeof cancelStream==='function'){await cancelStream();}
+  if(typeof cancelStream==='function'){await cancelStream('slash-interrupt');}
   showToast(t('cmd_interrupt_confirm'),2000);
 }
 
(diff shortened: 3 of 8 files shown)


nesquena-hermes:
## Gate results — GREEN (decision-ready)

Ran the full authoritative gate on the fixed worktree (`--workdir`, so it gates the actual fix, not the PR's original head).

**Codex** verified the core logic is correct:
> *"The route fallback is verified: a missing GET route returns JSON `{"error":"not found"}` with HTTP 404, so the new clarify guard still warns for the genuine missing-endpoint case while excluding `Session not found`."*

Codex found **no CORE or security issues** in the clarify/interrupt logic. Its only two findings (both SILENT) were pre-existing static test harnesses that broke on the intentional `cancelStream(reason)` signature change + the new stdout provenance log — both now fixed:
- `test_cancel_stream_owner_guard.py`: parse the last non-empty stdout line (the provenance `console.info` fires during the run; result JSON is emitted last).
- `test_sprint36.py`: signature-tolerant `re.search(r"async function cancelStream\s*\(")` + widened catch-block window.

**Full pytest suite (local, serial):** `11523 passed, 11 skipped, 1 xfailed, 2 xpassed` in 6m33s — zero failures.

**Static:** `node --check` on all 4 touched JS files ✅ · `scope_undef_gate.py` CLEAN ✅ · ruff forward gate 0 new violations ✅.

**New test coverage:** `tests/test_issue5345_*.py` (9 tests) — verified to fail 8/9 against pre-fix master (not vacuous). Locks the clarify-guard structured-status branching, the stale-session-before-missing-endpoint ordering, the provenance logging, and the backend invariant that `/api/clarify/pending` never 404s.

This is agent-authored and touches the chat/clarify surface, so per project convention it routes to @nesquena for the merge trigger — not self-merged.

ruizanthony:
Follow-up for the Greptile P2 findings: I do not have push rights on the upstream head branch, so I opened #5348 against `fix/5345-clarify-toast-interrupt-provenance`.

It removes `CHANGELOG.md` from the effective PR diff and moves the clarify poll warn-level diagnostic after the handled stale-session and missing-endpoint branches. Validation: `git diff --check`, `node --check` on touched static files, and `./scripts/test.sh tests/test_issue5345_clarify_toast_and_interrupt_provenance.py tests/test_cancel_stream_owner_guard.py tests/test_sprint36.py -q` → 35 passed.

nesquena-hermes:
Shipped in **v0.51.795** — combined with the #5348 follow-up. The false "Clarify endpoint unavailable" toast now only fires on a genuine missing-route 404; stale session-scoped 404s are handled silently. Codex SAFE, full suite green. Thanks @ruizanthony for the #5348 Greptile hardening (folded in + credited).

Issue #4875, opened 2026-06-24
bug: internal auto-continuation system prompts leak into the transcript (output-length / tool-call variants not filtered)


nesquena-hermes:
## Summary

Internal **auto-continuation system prompts leak into the visible transcript** as right-aligned user bubbles, then a turn can end with the red error `Response remained truncated after 3 continuation attempts`.

When a model response hits the output-length cap, the agent core injects a hidden control prompt (`[System: Your previous response was truncated by the output length limit. Continue exactly where you left off. ...]`) and re-asks the model, up to 3 times. WebUI is supposed to filter these control prompts out of the rendered transcript — and it does for the **network-error** variant — but the **output-length** and **tool-call-too-large** variants are NOT matched by the filter, so they render as visible chat bubbles.

Reproduced on **master (v0.51.625)**.

## What the user sees

Two stacked right-aligned bubbles reading:

> [System: Your previous response was truncated by the output length limit. Continue exactly where you left off. Do not restart or repeat prior text. Finish the answer directly.]

…followed by an error card: **Error: Response remained truncated after 3 continuation attempts**.

## Root cause

The agent core emits **three** continuation-prompt variants (`agent/conversation_loop.py` `_get_continuation_prompt`):

1. network error — `"...cut off by a network error..."`
2. output length — `"...truncated by the output length limit..."`
3. tool-call too large — `"...tool call (...) was too large..."`

WebUI's transcript hide-filter requires the literal phrase **"cut off by a network error"**, so it only catches variant #1:

- `static/ui.js` → `_isRecoveryControlMessageText()` (used by the persisted/rendered-message filter at `renderMessages`)
- `static/messages.js` → `_streamRecoveryControlMessageText()` (used by the live-stream path)

Both have:

```js
const systemRecovery=/^\[System:/i.test(normalized)
  && /previous response was cut off by a network error/i.test(normalized)
  && /continue exactly where you left off/i.test(normalized);
```

Variants #2 and #3 don't contain "cut off by a network error", so the filter returns `false` and they leak into the transcript. (Confirmed empirically: feeding the exact output-length string through the filter returns `false`.)

## Steps to reproduce

1. Configure a provider/model with a low effective output-token cap (or otherwise get a turn to return `finish_reason=length`).
2. Send any message.
3. When the model's reply is truncated by the output cap, the auto-continue control prompt(s) appear as right-aligned bubbles in the transcript instead of being hidden.

## Proposed fix (WebUI side)

[...]

Issue #3821, opened 2026-06-08
Handle max-iteration stops as tool_limit_reached instead of rendering synthetic summary prompts as user messages


franksong2702:
## Problem

When a WebUI-started Hermes Agent run reaches the max tool-calling / iteration budget, the turn can be rendered as if it completed normally, and the Agent's internal summary request can appear in the transcript as a user-authored message.

That summary request is not user input. It is a runtime control prompt used by the Agent to ask the model for a final toolless summary after the iteration budget is exhausted.

This makes the UI misleading in two ways:

1. The turn can look like a normal completed assistant response even though it stopped because the tool/iteration budget was exhausted.
2. The transcript can show an internal control instruction as a normal user message.

## Current Code Facts

WebUI already passes configured turn budget into Agent instances:

- `api/streaming.py` reads `agent.max_turns` / root `max_turns`.
- `api/streaming.py` passes `max_iterations` to `AIAgent` when supported.
- `api/streaming.py` wires `status_callback` when supported.

Hermes Agent currently exposes useful diagnostic metadata:

- `agent/conversation_loop.py` sets `turn_exit_reason = "max_iterations_reached(...)"`.
- The Agent result includes `messages`, `completed`, `turn_exit_reason`, and budget diagnostics.

Hermes Agent also currently constructs the max-iteration summary request as a synthetic user message:

- `agent/chat_completion_helpers.py`
- `handle_max_iterations()`
- appends:

```python
messages.append({"role": "user", "content": summary_request})
```

where `summary_request` starts with:

```text
You've reached the maximum number of tool-calling iterations allowed...
```

The Live-to-Final RFC already names the relevant product states:

- `tool_limit_reached`
- `no_response`

and says a turn without a final answer must not look like normal `completed`.

## Expected Behavior

When a run ends because max iterations / iteration budget was reached:

1. WebUI should classify the turn as `tool_limit_reached`.
2. If there is usable assistant final content, WebUI should keep it as the Final Answer.
3. The Final Answer should show clear status/copy indicating the run stopped because the tool iteration limit was reached.
4. If there is no usable assistant final content, WebUI should show an explicit terminal state, such as `tool_limit_reached` / no final answer.
5. WebUI should not render or persist the synthetic summary request as a user-authored message.
6. Replay/reconnect should restore the same Worklog, Final Answer if present, and terminal metadata.

## Suggested Compatibility Fix

At WebUI settle/persistence/replay boundaries:

[...]

Issue #3320, opened 2026-06-01
SSE recovery control message leaks into visible transcript after stream interruption (interrupted/apparent recovery replay)


franksong2702:
### Summary
SSE/stream reconnect recovery for interrupted runs can inject platform recovery control text into user-visible transcript.

### Repro context
- Session: `e30f2cc1efe0`
- During network interruption, platform injected recovery system prompt:
  `[System: The previous response was cut off by a network error mid-stream. Continue exactly where you left off. Do not restart or repeat prior text. Finish the answer directly.]`
- This message appeared in visible transcript as assistant content instead of being treated as internal recovery control.

### Expected behavior
- Recovery instructions should stay internal and never render as visible assistant transcript.
- Final transcript should show only real assistant content and tool/work activity state after settle.
- Interrupted/error terminal states should resolve through explicit recovery/terminal paths, not be misrepresented as normal completed answers.

### Actual behavior
- Recovery prompt text leaks into transcript and can remain visible after `done`/`restore`/`apperror` boundaries.
- Final settled rendering can append recovery artifact in same message list as user-facing answer.
- State flags can be presented as normal done/error states instead of explicit recovery control handling.

### Root cause
- `stale_interrupted_event` emits payload with recovery wording and no dedicated `recovery_control` typing/contract handling.
- Frontend `apperror` handler treated interrupted synthetic events as normal visible errors.
- Transcript finalization (`done`/`_restoreSettledSession`) could keep that synthetic assistant message in final `S.messages`, and `ui.js` visibility filters did not classify recovery-control text as non-visible assistant content.

### Fix summary
- Mark `stale_interrupted_event` payload with `recovery_control: True`.
- Add recovery-control message detection helpers in `static/messages.js` and `static/ui.js`.
- In `messages.js`:
  - filter `S.messages` in `done`/`_restoreSettledSession` to remove recovery-control messages.
  - in `apperror` interrupted path, branch recovery-control handling to restore settled session + toast, and avoid pushing the recovery prompt as user-visible assistant text.
- Add `tests/test_live_stream_ux.py` regressions for:
  - control flag presence
  - recovery-control detection in done/apperror paths
  - visibility filtering in ui.js.

### Verification
- `python3`-based pytest run (via hermes-agent venv):
  - `tests/test_run_journal_streaming_static.py`
  - `tests/test_live_stream_ux.py`
  - `tests/test_issue765_streaming_persistence.py`
  - `tests/test_turn_duration_display.py`
[...]

PR #5353, opened 2026-07-01
fix(webui): drop verification-stop synthetic nudge from transcript (#5334)

### LINKED ISSUE #5334: Verification-stop system prompt can leak into transcript; clarify no_response after visible streamed output
## Summary

I hit a WebUI transcript-settlement/display edge case with two related symptoms:

1. A turn had already streamed user-visible assistant text, but WebUI still settled the turn with a misleading `No response from provider` card.
2. An internal Hermes runtime verification-stop system prompt leaked into the visible transcript:

```text
[System: You edited code in this turn, but the workspace does not have fresh passing verification evidence yet...]
```

The second symptom looks like the same class as prior internal-control-message leaks (#3320, #3821, #4373, #4875), but exact searches for `fresh passing verification`, `verification-stop`, and `_verification_stop_synthetic` did not find an existing issue.

## Why this is tricky

A local experiment tried to suppress `no_response` whenever `_snapshot_and_append_partial_on_error(...)` captured visible streamed text. That is too broad and should not be upstreamed as-is.

Current upstream tests intentionally preserve this contract:

- `tests/test_issue5121_provider_auth_terminal_error.py::test_non_auth_partial_delivery_persists_error_turn`
- `tests/test_issue5121_provider_auth_terminal_error.py::test_non_auth_seeded_multi_turn_partial_persists_error_turn`

Those cases stream partial text, but still expect:

- the partial to be persisted, and
- an `apperror` with `type == "no_response"`, and
- the final saved message to be `_error`.

So “visible streamed text exists” is not a safe proof that the turn has a final answer. It can also be a real mid-answer provider/silent failure.

## Proposed invariant

For `No response from provider` settlement:

> Append a `no_response` terminal card only when the authoritative current-turn transcript/result lacks a committed, user-visible final assistant answer. Do not use the live streamed partial buffer as the proof of finality. If a genuine terminal/silent failure occurs after streamed text, preserve the `_partial` assistant row and still surface the error.

In practice, this likely means any fix should be in the existing backend settlement/final-answer evaluation path, not after `_snapshot_and_append_partial_on_error(...)`:

- `_merged_transcript_lacks_final_assistant_answer(...)`
- `_turn_transcript_lacks_final_assistant_answer(...)`
- `_session_lacks_final_assistant_answer(...)`
- `_assistant_reply_added_after_current_turn(...)`
- `_agent_result_terminal_failure(...)`
- current-turn anchoring/merge logic around `_run_agent_streaming(...)`

If a completed answer is being followed by `No response from provider`, the bug is probably one of:

- a successful result shape being misclassified by `_agent_result_terminal_failure(...)`, or
- merge/current-turn anchoring making `_merged_transcript_lacks_final_assistant_answer(...)` return true even though the current turn has a final assistant answer.

## Verification-stop prompt leak

The verification-stop prompt should remain model-facing runtime control, but not render as a visible chat row.

Suggested minimal display contract:

- Hide messages with a future explicit marker such as `_verification_stop_synthetic: true` (or an equivalent existing internal-control marker).
- Keep a strict, anchored text fallback for already-persisted raw messages:

```js
/^\[System: You edited code in this turn, but the workspace does not have fresh passing verification evidence yet\./
```

- Avoid blanket-hiding every `role: "system"` row unless WebUI explicitly adopts that as a product contract; a narrow marker/text filter is less risky and matches the precedent of recovery-control filtering.

## Suggested tests

Backend / settlement:

- Preserve the existing #5121 partial-silent-failure tests: partial text + no final result answer still yields `apperror type == "no_response"`.
- Add a regression only once the failing `result` shape is captured: current-turn final answer present should not append a generic `no_response` card.
- Cover reasoning-only and tool-only partials as not-final-answer.

Frontend / display:

- `_verification_stop_synthetic === true` is hidden.
- The strict raw verification-stop text prefix is hidden for backward-compatible persisted sessions.
- A normal system/user/assistant message that merely mentions verification evidence is not hidden.
- `_messageIsRenderable` uses the verification-stop helper without blanket `role === "system"` filtering.

## Related issues / PRs

- #3929 / #3931 / #3946: preserve partial work on turn-level errors.
- #5121 / #5129: provider/auth terminal failures bypassing chat error surfacing.
- #5141 / #5272: terminal-failure transcript evaluator cleanup.
- #3320, #3821, #4373, #4875: prior internal/synthetic control-message transcript leaks.

nesquena-hermes:
## Summary

An internal Hermes Agent **verify-before-finish** control turn was leaking into the visible WebUI transcript. When the agent edits code but lacks fresh passing verification evidence, its verify-on-stop loop appends a synthetic assistant "premature done" answer plus a synthetic `user` nudge whose text looks like:

> `[System: You edited code in this turn, but the workspace does not have fresh passing verification evidence yet...]`

These turns exist only to drive the loop for one more API round-trip. This is the **same class** of internal-control-message leak as #3320 / #3821 / #4373 / #4875.

## Root cause

The agent already marks both scaffolding rows with structured flags — `_verification_stop_synthetic` (verify-on-stop) and `_pre_verify_synthetic` (plugin `pre_verify`) — and keeps them out of its own durable store via `run_agent._EPHEMERAL_SCAFFOLDING_FLAGS`. **WebUI never learned to honor those markers**, so the synthetic `user` nudge settled into the visible transcript as a real turn.

## The fix (server-side)

Extends the existing suppression mechanism in `api/streaming.py` rather than inventing a new one:

- New `_SYNTHETIC_CONTROL_MESSAGE_FLAGS` (mirrors the agent set) + `_is_synthetic_control_message()` / `_drop_synthetic_control_messages()` helpers, placed next to the existing `_drop_synthetic_max_iteration_summary_requests`.
- Applied inside `_merge_display_messages_after_agent_result()` — the single chokepoint every settlement path routes `session.messages` (the visible transcript) through. All three inputs (`previous_display`, `previous_context`, `result_messages`) are filtered so prefix/delta detection stays aligned and any row that leaked in a prior build is scrubbed on the next merge.

The filter is **marker-based, not string-based**: a real user message that merely quotes the nudge copy is preserved.

## Scope

- **Symptom 2 (verification-stop leak): fixed.**
- **Symptom 1 (spurious "No response from provider" card while visible text streamed): intentionally OUT OF SCOPE** — the issue flags it as too broad to fix naively, and it isn't touched here.

## Tests

`tests/test_issue5334_verification_stop_leak.py` (8 tests, all passing) pins:
- the marker classifier accepts both synthetic flags and rejects a real message quoting the nudge text;
- **the `_verification_stop_synthetic` nudge never appears in the rendered transcript** (core deliverable), same for `_pre_verify_synthetic`;
- a normal user message is **not** excluded;
- a previously-leaked nudge in prior display is scrubbed on re-settlement.

Adjacent suppression suites (`test_tool_limit_terminal_state`, `test_issue4875_recovery_prompt_filter`, `test_issue3468_duplicate_after_compression`, `test_stable_assistant_turn_anchor_phase0`) still pass — no regressions.

Fixes #5334.

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -1311,6 +1311,43 @@ def _drop_synthetic_max_iteration_summary_requests(messages, *, enabled: bool = 
     ]
 
 
+# Structured markers the Hermes Agent stamps on synthetic scaffolding turns that
+# drive its internal verify-before-finish loop. The agent appends BOTH a
+# synthetic assistant "premature done" answer AND a synthetic ``user`` nudge
+# (e.g. "[System: You edited code in this turn, but the workspace does not have
+# fresh passing verification evidence yet...]") to preserve role alternation for
+# the next API turn, and flags each with one of these keys. They exist only to
+# run the loop; they must never surface as visible user/assistant turns in the
+# WebUI transcript. This mirrors ``run_agent._EPHEMERAL_SCAFFOLDING_FLAGS`` on
+# the agent side (which keeps them out of the durable session store); WebUI
+# honors the same markers when building the visible transcript. Keep roughly in
+# sync with the agent set. (#5334; same class as #3320/#3821/#4373/#4875)
+_SYNTHETIC_CONTROL_MESSAGE_FLAGS = (
+    "_verification_stop_synthetic",
+    "_pre_verify_synthetic",
+)
+
+
+def _is_synthetic_control_message(message) -> bool:
+    """Return True for an Agent-internal synthetic scaffolding turn flagged by marker."""
+    return isinstance(message, dict) and any(
+        message.get(flag) for flag in _SYNTHETIC_CONTROL_MESSAGE_FLAGS
+    )
+
+
+def _drop_synthetic_control_messages(messages):
+    """Remove Agent-internal synthetic scaffolding turns from the WebUI transcript.
+
+    Honors the structured ``_verification_stop_synthetic`` / ``_pre_verify_synthetic``
+    markers the agent already sets, rather than string-matching the nudge copy.
+    """
+    return [
+        msg
+        for msg in list(messages or [])
+        if not _is_synthetic_control_message(msg)
+    ]
+
+
 def _agent_result_tool_limit_reached(result) -> bool:
     """Return True when current-turn metadata says the tool iteration cap fired."""
     if not isinstance(result, dict):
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.799** (via release #5358). Marker-keyed filter honoring `_verification_stop_synthetic`/`_pre_verify_synthetic` in the merge path; Codex reproduced the leak + confirmed the settlement contract (error:'' sentinel) is intact. Full suite green.

Issue #4765, opened 2026-06-23
Memory bloat: WebUI loads all session JSON files into RAM; needs lazy session loading / cold archive path


kream23:
### Summary

On long-running self-hosted installs, `hermes-webui` becomes progressively slower and eventually needs daily restarts because the Python server keeps a large number of session JSON files in memory.

In my install, `~/.hermes/webui/sessions/` is ~350MB on disk and the WebUI Python process grows to 700MB+ RSS on a 2GB VPS. Restarting the WebUI drops RSS back to baseline, but memory grows again as the session set is loaded/used.

This looks related to the global in-memory `SESSIONS` cache and the current model where WebUI JSON session files are treated as the hot session store.

### Current behavior

- WebUI stores sessions as JSON files under:

```text
~/.hermes/webui/sessions/
```

- The server keeps session objects/messages in memory via `SESSIONS`.
- With many/large historical sessions, RSS grows significantly.
- Daily process restart is currently the only practical workaround.
- Deleting old JSON files is risky because they are still useful as primary/raw WebUI session sources, even if `state.db` can still be searched separately.

### Why deletion is not a good workaround

Hermes also has `~/.hermes/state.db`, and `session_search` can query that database.

However, deleting WebUI JSON sessions is still undesirable because:

1. The JSON files are the raw WebUI-side session artifacts.
2. Users may want to reopen old sessions in the UI.
3. Summaries/search indexes are not a safe replacement for preserving original conversation/tool-call data.
4. Removing source session files makes future recovery/debugging weaker.

### Desired behavior

Old sessions should be preserved but not loaded into RAM by default.

A possible design:

1. Keep only recent/active sessions in the hot `sessions/` directory.
2. Move older sessions into a cold archive directory, e.g.:

```text
~/.hermes/webui/session-archive/raw/
```

3. Maintain a lightweight metadata index, e.g. SQLite or JSONL manifest:

```text
~/.hermes/webui/session-archive/manifest.sqlite
```

Containing:

- session id
- title
- created/updated timestamps
- model/provider
- project/tags/archive status
- raw file path
- optional summary
- size/message count

4. Sidebar/session listing should use metadata only.
5. Full session messages should be loaded lazily only when a user opens a session.
6. `SESSIONS` should be bounded with an LRU limit instead of retaining unbounded full session histories.
7. Optional: archived raw JSON can be compressed, e.g. `.json.zst`, but preserving raw recoverability matters more than compression.

### Proposed implementation direction

[...]

PR #5354, opened 2026-07-01
fix(webui): crash visibility — faulthandler + thread excepthook + exit audit (#4633)

### LINKED ISSUE #4633: Silent server.py crash after ~9-16h uptime (no traceback, no shutdown audit)
# Silent WebUI crash after ~9-16h uptime

## Symptom
`server.py` exits silently after 9-16 hours of normal operation. No traceback, no shutdown audit log entry, no OOM, no segfault, no core dump. The log file simply stops mid-request.

## Evidence
- 4 observed crashes over 3 days: uptimes of ~5min, ~9.5h, ~16h, ~12.5h
- `_log_shutdown_audit()` never fires → not a clean shutdown
- No `dmesg` entry, no `/var/log/syslog` trace, no apport/coredump
- Memory 27GB available, FD limit 4096 soft / 1M hard, RSS normal

## Root cause (suspected)
`server.py` uses `ThreadingHTTPServer` with `daemon_threads = True`. Each SSE/long-poll connection spawns a daemon thread. An unhandled exception in any request handler thread can terminate the process silently — daemon threads don't prevent exit and Python's default signal handling for SIGTERM/SIGINT produces no log.

Additionally, `faulthandler` is not enabled, so a C-level crash (e.g. in a socket extension under load) kills the process with zero output.

## Fix applied locally
1. Added `faulthandler.enable()` early in `main()` — captures segfaults and C-level aborts to stderr
2. Added signal handlers for SIGTERM/SIGINT/SIGQUIT/SIGHUP that log the signal name + run shutdown audit before `sys.exit()`

## Environment
- Hermes WebUI: commit `ca5fe0ac` (latest on main as of 2026-06-21)
- Python 3.11.15
- Linux 6.8.0-124-generic (Ubuntu Server 24.04)
- Headless, no TUI, accessed via Tailscale

nesquena-hermes:
## Problem (#4633)

`server.py` exited **silently** after 9–16h: no traceback, no shutdown-audit line, no core dump — the log just stopped mid-request.

Two structural gaps made a death invisible:

- It runs a `ThreadingHTTPServer` with `daemon_threads = True`. An **unhandled exception in a request / SSE / long-poll handler thread** is reported through `threading.excepthook`; the stock hook only prints to `sys.stderr` with no context and, if that stream is gone, the report is lost — so a single handler-thread exception could vanish.
- **`faulthandler` was not enabled**, so a *native* crash (segfault/abort in a C extension — sqlite, ssl, a provider SDK) terminated the interpreter with nothing written.
- Compounding both: the WebUI configures **no logging handlers**, so `INFO`/`ERROR` records are dropped by logging's `lastResort` filter (WARNING+ only). Even the existing `_log_shutdown_audit` `logger.info(...)` never reached the log.

This PR is **crash visibility only** — when the process dies we now get a diagnostic, and one handler-thread exception can't silently kill work. The paired **memory root-cause is #4765** (not addressed here).

## What was added

New `api/crash_visibility.py` (stdlib-only; every hook is defensive and **never raises** — a raising hook would just be a new silent-failure mode). `install_crash_visibility()` is idempotent and is wired into `server.main()` **before any heavy startup work**:

1. **faulthandler** — `faulthandler.enable(all_threads=True)` so a native segfault/abort dumps a C-level traceback of every thread. Also registers `SIGUSR1` (`faulthandler.register`) for on-demand all-thread stack dumps to diagnose a *hang* (`kill -USR1 <pid>`).
2. **threading.excepthook** — logs any uncaught daemon/handler-thread exception (thread name, ident, daemon flag, full traceback) instead of losing it. Chains to any prior non-stock hook; ignores clean thread exit (`exc_type=None`) and `SystemExit`.
3. **sys.excepthook** — logs uncaught main-thread exceptions; preserves default `KeyboardInterrupt` (Ctrl-C) behaviour.
4. **atexit exit-audit** — a breadcrumb on clean/unwound exit. Its *absence* in the log narrows a silent death to an un-unwound kill (OOM / SIGKILL / native abort), which is itself diagnostic.

Every diagnostic is written **directly to the fault stream** (`sys.stderr`, which the bootstrap redirects into the WebUI log file) **and** mirrored through `logging`, so the line is guaranteed to land regardless of logging configuration.

## Not changed

No request-handling behavior change, no new dependencies, standard library only.

## Tests

`tests/test_issue4633_crash_visibility.py` (10 tests, all passing) asserts: faulthandler enabled after install; install idempotent; `threading.excepthook` installed and a **real daemon-thread exception is logged** (name + traceback in both the stream and via logging); clean thread exit / malformed args are swallowed without raising; `sys.excepthook` installed and main-thread exceptions logged; `KeyboardInterrupt` not treated as a crash; atexit audit registered and emits; and `server.py` wires in `install_crash_visibility()`. No live server socket required. Existing `test_shutdown_audit_logging.py` and `test_issue1458_stability_hardening.py` still pass.

Fixes #4633.

### DIFF
--- a/api/crash_visibility.py
+++ b/api/crash_visibility.py
@@ -1,0 +1,298 @@
+"""Crash-visibility hardening for the Hermes WebUI server (issue #4633).
+
+Background
+----------
+``server.py`` runs a ``ThreadingHTTPServer`` with ``daemon_threads = True``.
+Under that model a *silent* process death is possible in ways that leave the
+log ending mid-request with no traceback, no shutdown-audit line, and no core
+dump:
+
+* A **native** crash (segfault / abort in a C extension, e.g. sqlite, ssl,
+  a provider SDK) terminates the interpreter with no Python-level traceback.
+  Without ``faulthandler`` enabled, nothing is written.
+* An **uncaught Python exception in a daemon request/SSE/long-poll thread**
+  is reported through ``threading.excepthook``.  The stock hook prints to
+  ``sys.stderr`` with no context, and if that stream is gone the report is
+  lost — so a single handler-thread exception can vanish silently.
+* An uncaught exception on the **main thread** goes through ``sys.excepthook``.
+
+The paired *root cause* (a memory issue that eventually triggers the death)
+is being fixed separately in #4765.  This module is purely about
+**diagnostics**: when the process dies we want a recorded reason, and one
+handler-thread exception must never disappear without a log line.
+
+Design constraints
+------------------
+* Standard library only; no new dependencies.
+* Every hook is defensive — a logging hook must NEVER raise (a raising hook
+  would itself become a new silent-failure mode).
+* Diagnostics are written **directly to the stderr stream** (which the
+  bootstrap redirects into the WebUI log file) *and* mirrored through the
+  ``logging`` module.  The WebUI configures no logging handlers, so an
+  ``INFO``/``ERROR`` record would otherwise be swallowed by logging's
+  ``lastResort`` filter (WARNING+ only) and never reach the log — the exact
+  failure mode of #4633.  The direct write guarantees the line lands.
+* ``install_crash_visibility()`` is idempotent and safe to call at import or
+  startup time.
+"""
+from __future__ import annotations
+
+import atexit
+import faulthandler
+import logging
+import os
+import signal
+import sys
+import threading
+import traceback
+from typing import Optional, TextIO
+
+__all__ = [
+    "install_crash_visibility",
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.800**. faulthandler + threading/sys excepthook + exit audit, installed once at startup (stdlib-only, defensive, no request-path change). If the process ever exits abnormally again it now leaves a diagnostic. Codex SAFE (re-gated on rebased diff), full suite green.

Issue #5307, opened 2026-07-01
bug(sessions): opening delegated subagent child shows empty transcript despite state.db messages


claw-io:
## Summary

Opening a delegated subagent session from the sidebar (nested child or top-level **"Subagent Session"**) often shows an **empty or stub chat pane**, even when messages exist in Hermes Agent **`state.db`** for that `session_id`.

Delegation **summaries** may still appear on the **parent** as `delegate_task` / `subagent_progress` tool cards (`static/ui.js`); users may read that as “sub-session has no results.”

## Expected

Selecting a subagent child loads the child’s transcript from `state.db` / sidecar merge.

## Actual

Child pane empty or incomplete after `loadSession(childSid)`.

## Reproduction

1. Run `delegate_task` from a WebUI parent until a child appears in the sidebar.
2. Click the child row.
3. Observe messages after load completes.

**Dev check:** If `state.db` has messages for the child id but UI is empty → WebUI load/merge bug.

## Code path

1. **Child click** (`static/sessions.js` ~child row handler): `import_cli` only when `_isExternalSession` (~1254–1256) — **not** `source=subagent`.

2. **`loadSession`** → `_ensureMessagesLoaded` → `GET /api/session`.

3. **`api/routes.py` `_handle_session_get`:** Non-messaging sessions merge WebUI sidecar with `get_state_db_session_messages`. `_session_requires_cli_metadata_lookup` (~3114+) targets CLI/imported shapes, not subagent children by default.

4. **Sidecar vs DB drift:** Sidebar list metadata from `read_session_lineage_metadata` can show `parent_session_id` while JSON sidecar still has `null`; per-session GET may not surface DB messages.

5. **`api/models.py` `get_cli_session_messages`:** Stitched DB path used for messaging/CLI classification, not default subagent open.

## Related issues

- #4691 (open): async delegation silently drops results in WebUI sessions
- #4912 (closed): async_delegation completion delivery to parent chat
- #5244 (merged): sidebar nesting only

## Suggested fix

- For `source=subagent` / `relationship_type: child_session`, always merge `get_state_db_session_messages` on `GET /api/session`.
- Sync agent lineage into sidecar on enrichment; or force DB-backed load when sidecar message count is 0 but DB has rows.
- UX: link from parent `subagent_progress` cards to child `session_id`.

## Files

| File | Role |
|------|------|
| `static/sessions.js` | `loadSession`, `_isExternalSession` |
[...]

Issue #5306, opened 2026-07-01
bug(sidebar): delegated subagent rows flicker in sidebar while parent session is active


claw-io:
## Summary

While a parent WebUI session is active (streaming / delegating), child subagent sessions may **briefly appear** in the sidebar, then **vanish** on the next `/api/sessions` refresh, then **reappear** once list metadata catches up. This happens without switching away from the parent chat.

**Upstream note:** PR #5244 improved nesting under the parent; flicker can still occur when children fail the sidebar visibility predicate or are not yet `childRenderable`.

## Expected

Linked delegate children stay visible (or show a stable “running child” indicator) for the duration of the parent turn.

## Actual

Subagent-related rows enter and leave the sidebar between polls while the parent remains active.

## Reproduction

1. Start a WebUI chat that uses `delegate_task` (background children).
2. Keep the **parent** session selected.
3. Watch the sidebar for ~5–30s during streaming poll refreshes.

## Code path

1. **Streaming poll** refreshes list: `_applySessionListPayload` → `renderSessionListFromCache` (`static/sessions.js`).

2. **Visibility filter** — `_partitionSidebarSessionRows` drops rows unless `_sidebarRowHasVisibleMessages` (~6100–6108):
   - Requires `message_count > 0`, streaming flags, attention, or active-session exception.
   - **Children** with transient `message_count === 0` are omitted from `sessionsRaw`; parent is often exempt via `activeSidForSidebar`.

3. **Attach gate** — `_attachChildSessionsToSidebarRows` (~5727–5786):
   - `childRenderable = renderableChildIds.has(child.session_id)` — child must survive step 2 to nest or show as orphan.
   - `bubbleSidebarState` bubbles child streaming to parent when attached (#5244 path).

4. **Deferred list apply** — `_schedulePendingSessionListApply` can delay applying fresher payloads during sidebar interaction.

## Related issues

- #1792 (closed): first-turn optimistic sidebar + delegate liveness
- #5244 (merged): subagent stacking under WebUI parent

## Suggested fix

- Keep `relationship_type === 'child_session'` rows visible when parent is in-list and (parent streaming OR child exists in lineage metadata), even if child `message_count` is temporarily 0.
- Ensure `/api/sessions` exposes stable counts for new `source=subagent` rows as soon as the child session is created.

## Files

| File | Symbols |
|------|---------|
[...]

PR #5244, opened 2026-06-30
fix(sidebar): stack subagent sessions under parent

santastabber:
## Thinking Path

- The sidebar payload can contain a normal WebUI parent row plus delegated subagent rows marked `relationship_type: "child_session"`.
- Subagent rows can also carry `_cross_surface_child_session` because their source marker is `subagent` while the parent is `webui`.
- The render path was forcing those cross-source rows into orphan/top-level rows before trying visible-parent attachment.
- The fix preserves existing behavior for external/messaging parents while letting WebUI-parented subagents attach normally.

## What Changed

- Resolve the visible parent row before deciding whether a `_cross_surface_child_session` must remain top-level.
- Keep cross-surface child rows top-level only when the resolved parent is external/messaging/CLI.
- Allow subagent rows whose visible parent is a WebUI conversation to stack under that parent.
- Added runtime-style regressions covering:
  - WebUI parent + subagent child rows stack under the parent.
  - External/messaging parent cross-surface rows remain top-level/orphan.
  - Hidden archived parents still suppress cross-surface child orphans.

## Why It Matters

Delegated subagent sessions should not clutter the main sidebar as separate conversations when their parent WebUI conversation is visible. This restores the intended parent/child grouping without changing archived-session persistence or external-session import behavior.

## Verification

- `node --check static/sessions.js`
- `git diff --check origin/master...HEAD`
- `./scripts/test.sh tests/test_session_lineage_collapse.py tests/test_sidebar_session_partition.py tests/test_issue4766_sidebar_source_pushdown.py tests/test_session_lineage_metadata_api.py tests/test_session_lineage_report.py -q`
  - `89 passed`
- `./scripts/test.sh tests/test_static_js_runtime_lint.py tests/test_ruff_forward_lint.py -q`
  - `4 passed, 1 skipped`
- `.venv/bin/python scripts/ruff_lint.py --diff origin/master`
  - no new violations
- Deterministic Node transform smoke:
  - WebUI parent + 3 subagent cross-surface children renders as one parent row with `_child_session_count=3` and no top-level child rows.
  - Telegram/messaging parent + cross-surface WebUI child remains top-level/orphan.
- TruffleHog commit-range scan on a normal clone: no findings.
- Gitleaks changed-files and commit-range scans: no findings.
- CodeRabbit CLI review: `findings: 0`.
- Claude Code read-only review: `APPROVE`.

## Risks / Follow-ups

- This is render-path only. It does not cascade archive state or mutate saved child sessions.
[...]

PR #5355, opened 2026-07-01
fix(webui): overlay real state.db message count for subagent children (#5308)

### LINKED ISSUE #5308: bug(sidebar): post-#5244 delegate subagents can vanish entirely (childRenderable gate)
## Regression after #5244 (sidebar nesting)

After upgrading to a build that includes **#5244**, delegate subagent sessions can **disappear entirely** from the sidebar (not nested under parent, not top-level), whereas pre-upgrade they often appeared as top-level **"Subagent Session"** rows.

### Cause

`static/sessions.js` — `_attachChildSessionsToSidebarRows`:

- `childRenderable` = child survived `_sidebarRowHasVisibleMessages` in `_partitionSidebarSessionRows` (~5721–5728, ~6100–6141).
- Children are pushed to `parentRow._child_sessions` only when `childRenderable` (~5777–5780).
- If parent exists but child is **not** `childRenderable`, code only runs `bubbleSidebarState` (~5782) — **no sidebar row, no orphan**.

Pre-#5244, many such children were still shown as top-level orphans. Post-#5244, they can be **fully suppressed**.

Typical trigger: subagent child in `/api/sessions` with `message_count === 0` and no `active_stream_id` / streaming flags in list metadata while parent is active or user switched to another sidebar session.

### Also

- Nested children are **collapsed by default** (`_expandedChildSessionKeys`); if `_child_session_count === 0` due to above, parent shows **no** “N children” badge.
- Subagent child rows (non-fork) have **no** streaming indicator in sidebar (~7201–7210); empty chat pane = **#5307** / **#4691**.

### Suggested fix

1. Always attach `relationship_type: child_session` with resolvable parent to `_child_sessions`; use `childRenderable` only for orphan top-level fallback.
2. Visibility: keep child in list when parent visible + (parent streaming OR child exists in lineage/reference).
3. Subagent child sidebar rows: same streaming/unread affordances as fork children.

### Related

#5244, #5306, #5307, #4691

nesquena-hermes:
## Fix — delegated subagent sessions vanish from the sidebar (#5308)

Fixes #5308 (regression after #5244 + #5306).

### Root cause
A delegated subagent child's sidebar row is built from a stale sidecar reporting `message_count == 0`. The state.db count overlay in `_apply_sidebar_state_db_override_metadata` (api/models.py) was gated on `state_db_source == 'webui'`, so a subagent child (`state_db_source == 'subagent'`) never received its **real** message count from state.db. The front-end visibility predicate `_sidebarRowHasVisibleMessages` then dropped the zero-count row — and post-#5244 it wasn't even promoted to a top-level "Subagent Session" orphan, so the session **disappeared entirely**.

### Fix (server-side, root)
Widen the count/last-message overlay to `state_db_source in ('webui', 'subagent')`, preserving the existing conservative anti-resurrection guard. The **source-tag/title reassignment stays WebUI-only**, so a subagent child keeps its subagent classification (it does not get re-tagged as a webui session).

This is the same state.db-blind-metadata root as the #5307 transcript recovery, fixed at the metadata layer rather than by loosening the front-end predicate — which would have fought the deliberate #5306 active-parent scoping and risked resurrecting genuinely-empty rows.

### Tests
- `test_sidebar_state_db_overlay_counts_subagent_child_5308` — subagent child's real count is overlaid (0 → 6), classification preserved.
- `test_sidebar_state_db_overlay_does_not_count_foreign_cli_source_5308` — a cron/foreign CLI row still gets **no** overlay (scope guard).
- All existing reconciliation + #5132 top-N-cap tests still pass (50/50).

Fixes #5308.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -4028,6 +4028,20 @@ def _apply_sidebar_state_db_override_metadata(sessions: list[dict], metadata: di
             session['session_source'] = state_db_session_source
             session['source_label'] = state_db_source_label
             session['is_cli_session'] = False
+        # Overlay the real state.db message count for WebUI-owned rows AND for
+        # delegated subagent children (#5308). A subagent child
+        # (state_db_source == 'subagent') is backed by the delegate runner's
+        # state.db session, but its sidebar row is built from a stale sidecar
+        # that often reports message_count == 0. Without overlaying the true
+        # count, the front-end visibility predicate
+        # (_sidebarRowHasVisibleMessages) drops the row and the subagent
+        # session vanishes from the sidebar entirely (regression seam behind
+        # #5308, same state.db-blind-metadata root as the #5307 transcript
+        # recovery). The count overlay keeps the same conservative
+        # anti-resurrection guard used for WebUI rows. The source-tag / title
+        # reassignment above stays WebUI-only — a subagent child keeps its
+        # subagent classification.
+        if state_db_source in ('webui', 'subagent'):
             try:
                 current_count = max(0, int(session.get('message_count') or 0))
                 state_count = max(0, int(state_db_message_count or 0))
--- a/tests/test_webui_state_db_reconciliation.py
+++ b/tests/test_webui_state_db_reconciliation.py
@@ -155,6 +155,88 @@ def test_sidebar_state_db_overlay_preserves_numeric_actual_count():
     assert sessions[0]["actual_message_count"] == 5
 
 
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.798** (via release #5356). Root-fixed server-side: the state.db message-count overlay now applies to subagent children (not just webui rows), so a delegate child no longer vanishes from the sidebar. Codex SAFE, full suite green.

Issue #2233, opened 2026-05-14
WebUI memory leak causes the session to completely freeze


LineTang:
## Environment
- **Hermes Agent**: hermes-webui (latest, GitHub: nesquena/hermes-webui)
- **OS**: WSL2 Ubuntu 26.04 on Windows
- **Hardware**: Samsung Galaxy Book Flex (i5-1035G4, 12GB RAM, WSL2 memory limit 2GB)
- **Model**: mimo-v2.5-pro via xiaomi provider

## Bug Description

After extended use, the WebUI server process (`server.py`) memory grows to **1GB+ RSS**, causing all API endpoints to time out. The frontend then falls back to returning **locally cached content** — specifically the compression marker — for any user input, making the session appear completely broken.

## Steps to Reproduce

1. Start WebUI on port 7002
2. Have a long conversation (500+ messages, 800KB session file) with context compression triggered
3. Continue using the WebUI for ~30+ minutes without restart
4. Observe that sending any message immediately returns `[Your active task list was preserved across context compression]` — no API call is made

### Process memory at time of failure
```
PID 15005 — started 11:38, observed at ~11:47 (9 minutes uptime)
  %CPU 17.1  %MEM 56.3  VIRT 3.4GB  RSS 1.07GB
  command: python server.py (port 7002)
```

### API timeout
```
$ curl -s http://127.0.0.1:7002/api/health/agent
# Timed out after 60s — no response
```

### Crash diagnostics (auto-collected by keepalive)
```
crash-20260514_114711/ — WebUI health check failed, keepalive restarted
```

### Session data (backend is healthy)
- 11 independent sessions all triggered context compression at message index 104
- After compression, the **backend continued normally** — all sessions show valid tool calls and assistant responses for 400+ messages after compression
- The problem is purely frontend: when server.py is unresponsive, the browser-side code returns cached content without attempting an API call

### Affected session
```
session_20260514_065326_8a3de3.json — 544 messages, 796KB, mimo-v2.5-pro
```

## Expected Behavior

- WebUI should not leak memory during normal use
- If the API is unreachable, the frontend should show an error toast/banner, not silently return stale content
- The compression marker `[Your active task list was preserved...]` should never be returned as a "response" to user input

## Actual Behavior

- `server.py` RSS grows to 1GB+ (56% of system memory) within ~9 minutes
- All API calls time out
- Frontend returns the compression marker text as if it were the model's response, with no API round-trip
[...]

PR #5357, opened 2026-07-01
fix(webui): bound in-memory SESSIONS cache with lazy reload (#4765)

### LINKED ISSUE #4765: Memory bloat: WebUI loads all session JSON files into RAM; needs lazy session loading / cold archive path

nesquena-hermes:
## Summary

Fixes the unbounded in-memory `SESSIONS` growth that crashes long-running self-hosted WebUI installs — the silent-crash-after-hours cluster **#4765 / #2233 / #4633**.

## Root cause

The WebUI keeps every session object + messages in a global in-memory LRU (`api.config.SESSIONS`, an `OrderedDict`). The cache *had* a cap (`SESSIONS_MAX`) and lazy reload in `get_session()`, but eviction was performed with a **blind** `SESSIONS.popitem(last=False)` at 8 call sites. That is unsafe in two ways:

1. It could evict the least-recently-used entry **even if it was actively streaming or held unsaved messages**, dropping an in-flight turn or losing a conversation.
2. Because the cap was only settable via an env var (and the safety concerns above discouraged aggressive eviction), on real installs the effective behavior was **memory that only grows**: RSS climbs ~700MB → 7.5GB@9h → 17.8GB@44h, then the interpreter segfaults. Maintainer triage confirmed this is the root of the crash cluster.

## Fix — bounded, lazy, and safe

- **`_session_is_evictable(s)`** — a session is evictable **only** when *all* hold:
  - not streaming (`active_stream_id` unset),
  - no in-flight/queued turn (`pending_user_message` and `pending_started_at` unset),
  - full state is provably on disk (sidecar `message_count` ≥ in-memory message count; metadata-only stubs and zero-message shells are trivially safe).
  Anything we cannot *prove* is safe stays resident — using a little more RAM beats ever losing a session.
- **`_evict_sessions_over_cap(cap=None)`** — replaces all 8 blind `popitem` loops (`api/models.py`, `api/routes.py`, `api/streaming.py`). Walks the LRU oldest→newest and removes only provably-safe entries. It **never acquires `LOCK` or any stream lock itself** (every caller already holds `LOCK`), so it introduces no lock-ordering deadlock. If every over-cap candidate is active/unsaved, the cache may briefly exceed the cap — the intended safe behavior.
- **Lazy reload unchanged**: an evicted session transparently reloads from its `<sid>.json` sidecar on the next `get_session(sid)` / `SESSIONS[sid]` access. **No call sites changed** — the safety lives in the accessor and the eviction helper.
- **`_index.json` sidebar behavior preserved**: the sidebar reads the index, not the full `SESSIONS` cache — untouched.

## Config key (no new env var)

The cap is now configurable via **`config.yaml`**:

```yaml
webui:
  sessions_cache_max: 300   # default
```

Resolved by `config.get_sessions_cache_max()` with precedence:
`config.yaml webui.sessions_cache_max` → `HERMES_WEBUI_SESSIONS_MAX` (legacy env fallback) → `DEFAULT_SESSIONS_CACHE_MAX = 300`. A missing / non-numeric / `< 1` value falls back so a typo can never disable the bound and reintroduce the leak. No new `HERMES_*` env var is added (per project policy for non-secret config).

## Safety invariants proven by tests

New `tests/test_issue4765_sessions_lru_eviction.py` (8 tests, all passing):

- eviction happens once the cache grows past the cap;
- an **active / streaming** session is never evicted (even as the oldest / under heavy churn);
- a session with **unsaved in-memory messages** (or an unsaved tail ahead of its sidecar) is never evicted;
- an evicted session **lazily reloads from disk with identical content**;
- **no data loss** — every sidecar file survives heavy churn and stays fully retrievable;
- the config knob reads `webui.sessions_cache_max` and fails safe on bad input.

## Verification

```
tests/test_issue4765_sessions_lru_eviction.py .......... 8 passed
# regression sweep (no breakage across the many SESSIONS call sites):
tests/test_issue3506_memory_and_watcher.py, test_session_duplicate_edit.py,
test_route_session_list_cache_extraction.py, test_session_sidebar_cache.py,
test_session_index.py, test_core_data_loss_cases.py, test_session_save_mode.py,
test_empty_session_no_disk_write.py, test_session_lost_response_regression.py,
test_inflight_purge_missing_sessions.py, test_stale_stream_cleanup.py,
test_session_sidecar_repair.py  → 243 passed
python3 -c 'import ast; ast.parse(open("api/config.py").read())'  → OK
```

Fixes #4765.

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,7 +367,7 @@ Full list of environment variables:
 | `HERMES_CONFIG_PATH` | `$HERMES_HOME/config.yaml` | Path to Hermes config file |
 | `HERMES_WEBUI_SERVER_CWD` | *(unset)* | Working directory for the server process. Defaults to the agent dir; point it at a writable workspace when the agent dir is read-only so fallback relative writes land somewhere writable |
 | `HERMES_WEBUI_AGENT_CACHE_MAX` | `25` | Max live agent instances kept warm in the in-memory LRU. Each pins a full conversation transcript, so this is the dominant lever on resident memory — lower it on installs with many long sessions to cap RAM (at the cost of more cold reloads) |
-| `HERMES_WEBUI_SESSIONS_MAX` | `100` | Max compact `Session` objects held in the in-memory LRU. Lighter than the agent cache; lower it on installs with hundreds of sessions |
+| `HERMES_WEBUI_SESSIONS_MAX` | `300` | Legacy operator override for the max compact `Session` objects held in the in-memory LRU. Prefer the `webui.sessions_cache_max` key in `config.yaml` (which takes precedence); this env var remains a fallback. Bounds resident memory so long-running installs cannot accumulate every session ever touched and eventually crash (#4765/#2233/#4633). Eviction only ever drops clean, persisted, non-active sessions — an evicted session lazily reloads from its JSON sidecar on next access |
 
 Extension deployments can inspect sanitized, authenticated diagnostics at `GET /api/extensions/status`; see [WebUI Extensions](docs/EXTENSIONS.md#diagnostics).
 
--- a/api/config.py
+++ b/api/config.py
@@ -7355,10 +7355,47 @@ _INDEX_HTML_PATH = REPO_ROOT / "static" / "index.html"
 
 # ── Thread synchronisation ───────────────────────────────────────────────────
 LOCK = threading.Lock()
-# Max compact Session objects held in the in-memory LRU (issue #3506). Lighter
-# than the agent cache (no live agent runtime), but still bounded and operator-
-# tunable via HERMES_WEBUI_SESSIONS_MAX for installs with hundreds of sessions.
-SESSIONS_MAX = _env_int("HERMES_WEBUI_SESSIONS_MAX", 100)
+# Max compact Session objects held in the in-memory LRU (issue #3506, #4765).
+# Lighter than the agent cache (no live agent runtime), but still bounded so a
+# long-running self-hosted install cannot accumulate every session it ever
+# touched in RAM and eventually segfault (the #4765/#2233/#4633 crash cluster).
+#
+# Precedence for the effective cap is resolved by get_sessions_cache_max():
+#   1. config.yaml  webui.sessions_cache_max   (preferred, no new env var)
+#   2. HERMES_WEBUI_SESSIONS_MAX env var        (legacy operator override)
+#   3. DEFAULT_SESSIONS_CACHE_MAX               (sane bounded default)
+DEFAULT_SESSIONS_CACHE_MAX = 300
+SESSIONS_MAX = _env_int("HERMES_WEBUI_SESSIONS_MAX", DEFAULT_SESSIONS_CACHE_MAX)
+
+
+def get_sessions_cache_max(config_data: dict | None = None) -> int:
+    """Return the effective in-memory SESSIONS cache cap (issue #4765).
+
+    The bound is configurable through ``webui.sessions_cache_max`` in
+    ``config.yaml`` so operators of large self-hosted installs can size the
+    cache without editing source or adding a new ``HERMES_*`` env var (this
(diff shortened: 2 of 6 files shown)


nesquena-hermes:
Shipped in **v0.51.801**. The in-memory SESSIONS cache is now safely bounded (config.yaml `webui.sessions_cache_max`, default 300) with data-safe eviction (never evicts a streaming/pending/not-yet-persisted session) + lazy reload. Gate: Codex SAFE (all 8 eviction sites guarded), full suite 11477, + a live evict→reload drive proving byte-identical reload and zero file loss. Closes the memory-growth crash cluster.

PR #5360, opened 2026-07-01
fix(webui): align reconciliation dedup key with workspace-prefix stripping (#5339)

### LINKED ISSUE #5339: bug: Post-restart stale user message permanently prepended to all subsequent turns
## Summary

After a container restart, the first user message sent in the session gets permanently prepended to every subsequent user message. Additionally, these contaminated messages are inserted into the messages array at the restart boundary position rather than appended at the end, causing them to render out of chronological order in the UI.

## Reproduction

1. Have any conversation in a WebUI session.
2. Send a message and wait for the agent to fully complete its response (normal turn, no interruption).
3. Restart the WebUI container.
4. Send a new message.
5. **Result:** The new message has the prior turn's content permanently prepended. All subsequent messages in the session also carry this stale prefix.

No special preconditions required - no attachments, no long session, no interrupted turn. The bug triggers on any completed turn followed by a container restart.

## What the user sees

- After restart, the first new message appears to send normally.
- But looking at the chat history, every message from that point forward has the prior turn's text prepended to it.
- These contaminated messages cluster at the restart boundary position in the chat rather than appearing at the chronological end. New messages sent minutes later render ABOVE the original post-restart message.

## Evidence from live session

Session `524e38af83e8`. Timestamps from session JSON:

```
msg[96]  ts=1782910868  user      "Some user feedback on Sonnet 5..."     <- last pre-restart user turn
msg[97]  ts=1782910868  assistant "That corroborates your read..."         <- completed response
                        ─── CONTAINER RESTART ───
msg[98]  ts=1782910859  user      "[Workspace::v1: ...]Some user feedback..."  <- state.db duplicate (earlier ts, workspace prefix)
msg[99]  ts=1782911278  user      "[Workspace::v1: ...]Some user feedback...\n\nLet's see if you can read..."  <- FIRST CONTAMINATED: prior turn + new input merged
msg[100] ts=1782911375  user      "Let's see if you can read...\n\nNo, you're landing there because..."
msg[101] ts=1782911542  user      "Let's see if you can read...\n\nWhich make me question..."
  ...every subsequent message carries the same stale prefix...
msg[107] ts=1782911201  user      "Let's see if you can read this whole session..."  <- original CLEAN first post-restart message (earlier ts, later array position)
```

Key facts:
- **No `_recovered=True` flag** on any contaminated message. PR #4475's 4th-pass sanitizer cannot help.
- **msg[98]** is a duplicate of msg[96] with different content (workspace prefix added, minor formatting differences) and a different timestamp. It appeared after restart via state.db reconciliation.
- **msg[107]** (the clean original) has an earlier timestamp than msg[99-106] but a later array position - confirming new messages are being inserted at the restart boundary, not appended.

### 1. Reconciliation produces duplicate user messages after restart

`reconciled_state_db_messages_for_session()` merges state.db rows with the sidecar. For the last completed turn:

- **Sidecar** has: `"Some user feedback on Sonnet 5\n\n[Attached files: ...]"` (ts 1782910868)
- **State.db** has: `"[Workspace::v1: /workspace]\nSome user feedback on Sonnet 5\n\n[Attached files: ...]\n[screenshot]"` (ts 1782910859)

Same turn, different content and timestamp. The dedup layer (`_deduplicate_context_messages` via `_message_identity`) strips workspace prefixes but cannot account for other formatting differences between the display version and model-facing version of the same message (in this case a `[screenshot]` annotation tag added by the model-facing path).

Result: both copies survive into context as adjacent user messages.

### 2. Agent core merges adjacent user messages

The hermes-agent's `drop_thinking_only_and_merge_users()` sees two consecutive user-role messages and concatenates them with `\n\n`. This produces the first contaminated composite (msg[99]).

### 3. Stale prefix persists indefinitely

Once the composite is committed to context_messages, the stale prefix is permanent:
- `_detect_stale_user_merge` fails to identify the boundary because the prefix contains multi-line content.
- Each subsequent turn inherits the contaminated "previous user" context and the cycle repeats.

### 4. Array insertion at wrong position

New messages are inserted at the restart boundary in the messages array rather than appended at the end, causing the chronological ordering to break in the UI.

## Suggested fixes

1. **Reconciliation dedup**: `_session_message_content_key` (used by `state_db_delta_after_context`) should strip workspace prefixes, matching the behaviour of `_message_identity`. Additionally, minor formatting differences between display and model-facing versions of the same turn (annotation tags like `[screenshot]`) should not defeat deduplication.

2. **Adjacent user detection**: After reconciliation, if two adjacent user messages share substantially the same content (one being a superset/prefix of the other), treat them as duplicates rather than allowing the merge.

3. **Message append order**: Post-restart messages should always be appended at the array end, not inserted at the reconciliation boundary.

## Environment

- WebUI: HEAD (56d1f619, 2026-07-01), post v0.51.512
- Related: #4283, #4393, PR #4475 (different mechanism - this bug does not involve `_recovered` flag)

nesquena-hermes:
## Summary

Fixes #5339 — after a WebUI container restart, the first user message sent was **permanently prepended** to every subsequent user message, and the contaminated turns rendered out of chronological order (clustered at the restart boundary).

## Root cause: a key-function mismatch between two dedup layers

WebUI sends the model a workspace-prefixed `user_message` (`[Workspace::v1: /path]\n<text>`), while the visible/optimistic bubble **and** the WebUI sidecar row carry only the bare `<text>`. Two layers dedup user turns, but keyed them differently:

- **Reconciliation** (`api/models.py`): `state_db_delta_after_context` keys rows via `_session_message_content_key` → `_normalized_session_message_content`, which normalized whitespace **only** — no prefix stripping.
- **Streaming** (`api/streaming.py`): `_message_identity` **strips** the workspace prefix for user turns via `_strip_workspace_prefix(text, include_legacy=True)`.

So a state.db row (prefixed) and a sidecar row (bare) for the *same* completed turn produced **different** keys. The alignment loop in `state_db_delta_after_context` failed to match them, treated the state.db copy as a **new** row, and appended a duplicate user turn. The agent-side `drop_thinking_only_and_merge_users` pass then concatenated the two adjacent user rows with `\n\n` into a **permanent composite** — reproduced exactly by *completed turn → restart → new message*.

## The fix

Make `_session_message_content_key` strip the workspace prefix for `role == 'user'`, reusing the **same** `_strip_workspace_prefix` helper the streaming side uses (lazy import, because `api.streaming` imports `api.models` at module load — a top-level import would be circular). This stops the duplicate at its WebUI source rather than papering over it downstream.

Scope guarantees (all covered by tests):
- Only `role == 'user'` is stripped — assistant/tool keys are unchanged.
- A prefix-free user message keys identically before and after (idempotent).
- Genuinely different user messages still key distinctly.
- The agent-side merge (different repo) is untouched — the WebUI-side fix removes the duplicate that fed it.

## Tests

New `tests/test_issue5339_restart_stale_user_dedup.py` (7 tests): unit-level key agreement + idempotence + role scoping + distinctness; delta-level "prefixed state row is a mirror, not a delta" plus a guard that a genuinely new turn is still surfaced; and an end-to-end `/api/session` load asserting the completed turn is not duplicated at the restart boundary (buggy: 6 rows, `first answer` twice; fixed: 4 rows).

- `python3 -c 'import ast; ast.parse(open("api/models.py").read())'` → OK
- `tests/test_issue5339_restart_stale_user_dedup.py` → **7 passed**
- `tests/test_webui_state_db_reconciliation.py` → **41 passed** (no regression)
- `tests/test_issue3709_thinking_double_render.py` → **5 passed**

Fixes #5339.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -6206,9 +6206,31 @@ def _loose_session_message_content(value: str) -> str:
 def _session_message_content_key(msg: dict):
     if not isinstance(msg, dict):
         return ("non_dict", repr(msg))
+    role = str(msg.get("role") or "")
+    content = _normalized_session_message_content(msg)
+    if role == "user":
+        # WebUI sends the model a workspace-prefixed user_message
+        # ("[Workspace::v1: /path]\n<text>") while the visible/optimistic
+        # bubble and the WebUI sidecar row carry only the bare "<text>". The
+        # streaming dedup identity (_message_identity in api/streaming.py)
+        # strips this prefix for user turns, so this reconciliation key must
+        # do the same. Otherwise a state.db row (prefixed) and a sidecar row
+        # (bare) key DIFFERENTLY, the alignment loop in
+        # state_db_delta_after_context fails to match them, treats the
+        # state.db copy as a NEW row, and appends a duplicate user turn. The
+        # agent then merges the two adjacent user rows into a permanent
+        # composite -- the post-restart stale-user-prepend bug (#5339). Reuse
+        # the SAME helper as the streaming side (imported lazily to avoid a
+        # circular import; api.streaming imports api.models at module load) so
+        # the two dedup layers can't drift apart again.
+        from api.streaming import _strip_workspace_prefix
+
+        content = " ".join(
+            _strip_workspace_prefix(content, include_legacy=True).split()
+        )
     return (
-        str(msg.get("role") or ""),
-        _normalized_session_message_content(msg),
+        role,
+        content,
         str(msg.get("tool_call_id") or ""),
         str(msg.get("tool_name") or msg.get("name") or ""),
     )
--- a/tests/test_issue5339_restart_stale_user_dedup.py
+++ b/tests/test_issue5339_restart_stale_user_dedup.py
@@ -1,0 +1,284 @@
+"""Regression tests for issue #5339.
+
+Post-restart stale user message permanently prepended to all subsequent turns.
+
+Root cause was a KEY-FUNCTION MISMATCH between two dedup layers:
+
+* The reconciliation delta ``state_db_delta_after_context`` keys rows via
+  ``_session_message_content_key`` -> ``_normalized_session_message_content``
+  in ``api/models.py``, which used to normalize whitespace ONLY, with NO
+  workspace-prefix stripping.
+* The streaming-side identity ``_message_identity`` in ``api/streaming.py``
+  DOES strip the workspace prefix for user turns, because WebUI sends the model
+  a workspace-prefixed ``user_message`` (``[Workspace::v1: /workspace]\\n<text>``)
+  while the visible/optimistic bubble (and the WebUI sidecar row) carries only
+  the bare ``<text>``.
+
+So a state.db row carrying the workspace-prefixed text and a sidecar row
+carrying the bare text produced DIFFERENT keys; the alignment loop in
+``state_db_delta_after_context`` failed to match them, treated the state.db copy
+as NEW, and appended a duplicate user row. The agent-side merge then
+concatenated the two adjacent user rows into a permanent composite.
+
+The fix makes ``_session_message_content_key`` strip the workspace prefix for
+user-role messages (reusing the SAME ``_strip_workspace_prefix`` helper the
+streaming side uses), so the two layers can no longer disagree.
+"""
+import json
+import sqlite3
+from collections import OrderedDict
+from io import BytesIO
+from pathlib import Path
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.802**. The reconciliation dedup key now strips the workspace prefix for user messages (same helper as the streaming-side identity), so a state.db row and its bare-text sidecar align instead of duplicating. Codex reproduced the pre-fix duplicate + confirmed gone; full suite green; lazy import avoids the circular dep.

PR #5335, opened 2026-07-01
[locale]Translation of Chinese texts lacking localization

Loukky:

PR #5362, opened 2026-07-01
docs(changelog): restore dropped #4765 bounded-SESSIONS-cache entry

nesquena-hermes:
The CHANGELOG entry for #4765 (bounded in-memory SESSIONS cache, shipped in v0.51.801 via #5357) was dropped during a merge-conflict resolution while several releases landed in parallel. Restoring it so the release history reflects the shipped fix. Docs-only, no code change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ _No unreleased changes. Entries are moved into their version block when a releas
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Superseding with a combined changelog-restore PR that covers #4765, #5313, and #5335 (all three entries were dropped in merge-conflict resolution during the parallel release cadence). Closing this one.

PR #5368, opened 2026-07-01
Fix profile skills-stats thundering herd at cold startup (#5364)

### LINKED ISSUE #5364: Bug: Profile skills stats thundering herd — concurrent cold-startup requests cause 57–70s stalls via unlocked os.walk on every call
## Summary

The two-tier mtime cache introduced in #4783 (commits `18528a88` + `e4e3b468`) fixes the per-request SKILL.md rescan, but introduces a **thundering herd** at cold startup: the mtime probe (`_skill_tree_max_mtime_ns`) runs `os.walk(followlinks=True)` on **every call** to `_get_profile_skills_stats()`, and the cache has **no lock** — so concurrent requests at container boot all miss the cache and walk all profiles' skill trees simultaneously, stalling worker threads for 57–70 seconds.

## Timeline (observed in a Docker deployment, 9 profiles)

| Time (UTC) | Event |
|---|---|
| ~18:56:00 | Container started; S6 pre-init, profiles reconciled, WebUI deps installed |
| 18:56:28 | WebUI listening on `0.0.0.0:8787`, first requests received |
| 18:56:29 | Requests to `/api/sessions`, `/api/profile/active`, `/api/projects` begin stalling; `load_settings` stage takes 5,000ms+ |
| 18:57:37 | `/api/projects` took **57,784 ms** (57.7s) |
| 18:57:39 | `/api/projects` took **69,792 ms** (69.8s) |
| 18:57:39+ | Dozens of concurrent `/api/profile/active` and `/api/session/status` calls stacked up, each taking 5,000–7,000ms |

Thread dumps in the logs explicitly show worker threads stuck in `os.walk(skills_dir, followlinks=True)` inside `_skill_tree_max_mtime_ns()` and `_get_profile_skills_stats()`.

## Root Cause

Three compounding defects in `api/profiles.py`:

### 1. `_get_profile_skills_stats()` has no lock — concurrent calls all miss the cache

```python
# api/profiles.py:1694
def _get_profile_skills_stats(profile_dir: Path) -> tuple[int, int]:
    # ...
    current_mtime_ns = _skill_tree_max_mtime_ns(skills_dir, config_path)  # line 1711 — walks the tree

    cached = _SKILLS_STATS_CACHE.get(profile_dir)
    if cached is not None and current_mtime_ns == cached_mtime_ns and now < expiry:
        return enabled, compat  # "fast path" — but the walk already ran

    # Cache miss — walks the tree AGAIN (TOCTOU guard), then reads all SKILL.md:
    new_mtime_ns = _skill_tree_max_mtime_ns(skills_dir, config_path)  # line 1731
    res = _compute_profile_skills_stats(profile_dir)                  # line 1732
```

Unlike `_LIST_PROFILES_CACHE` which uses `_LIST_PROFILES_CACHE_LOCK` (line 1741), the skills stats cache (`_SKILLS_STATS_CACHE`, line 1584) has **no lock at all**. With `ThreadingHTTPServer` (one OS thread per request, `server.py:117`), concurrent requests at cold boot all miss the cache and walk simultaneously.

### 2. `_build_profile_rows_fast()` runs outside the lock in `list_profiles_api()`

```python
# api/profiles.py:1893–1900
with _LIST_PROFILES_CACHE_LOCK:
    cached = _LIST_PROFILES_CACHE          # check inside lock
if cached is not None and now - cached[1] < _LIST_PROFILES_CACHE_TTL:
    return [...]                            # cache hit: fast

rows = _build_profile_rows_fast()           # OUTSIDE the lock: thundering herd
```

`_build_profile_rows_fast()` (line 1751) calls `_get_profile_skills_stats()` for every profile (line 1791). At cold startup, the frontend fires multiple concurrent requests (`/api/profiles`, `/api/session/status`, `/api/projects`). All threads miss the list cache and all call `_build_profile_rows_fast()` simultaneously — each one walks all profiles' skill trees.

With 5 concurrent `list_profiles_api()` calls × 9 profiles = **45 simultaneous `_get_profile_skills_stats()` invocations**, each doing 2× `os.walk(followlinks=True)` + `_compute_profile_skills_stats()` (read + parse all SKILL.md).

### 3. The mtime probe fires on every single API call — no debounce

`_skill_tree_max_mtime_ns()` (line 1588) runs `os.walk(followlinks=True)` with per-directory and per-file `stat()` calls (~500–600 `stat()` calls per profile). On the host filesystem this takes ~5ms, but inside Docker's overlay2 filesystem, each `stat()` involves an overlay layer lookup that is significantly slower — especially under concurrent I/O contention from dozens of simultaneous walks.

The probe is designed to run on every call for out-of-band change detection (a deliberate design choice per #4783), but there is no short debounce. Every API request that touches profile data triggers the walk, even if the previous request was milliseconds ago.

## Amplification math

- 5 concurrent frontend requests at cold boot
- Each calls `list_profiles_api()` → `_build_profile_rows_fast()` → 9× `_get_profile_skills_stats()`
- Each `_get_profile_skills_stats()` does 2× `_skill_tree_max_mtime_ns()` (probe + TOCTOU re-probe) + `_compute_profile_skills_stats()`
- = **~90 concurrent `os.walk()` traversals**, each doing ~500 `stat()` calls
- = **~45,000 concurrent `stat()` calls** hitting Docker overlay2

## What I ruled out

- **Symlink loops**: No symlink loops exist in any profile's skills directory. The `followlinks=True` is not causing infinite loops.
- **Enormous file counts**: Each profile has ~500–600 files across ~240 directories — moderate, not pathological.
- **Unrelated log warnings**: Several unrelated warnings in the startup logs (auth configuration, MCP server guards) were investigated and do not contribute to the startup stall.

## Proposed Fix

Three changes, all in `api/profiles.py`:

1. **Add a lock to `_get_profile_skills_stats()`** — use double-checked locking so concurrent calls to the same profile share a single compute. This is the highest-impact fix.

2. **Move `_build_profile_rows_fast()` inside the lock** in `list_profiles_api()` — or use a double-checked locking pattern so only one thread builds the rows while others wait.

3. **Debounce the mtime probe** — add a short TTL (2–3 seconds) to the mtime probe itself, so the stat walk doesn't fire more than once per few seconds per profile. A 2-second debounce still catches out-of-band changes promptly while eliminating redundant walks under burst traffic.

## Related

- #4783 — original perf issue; the two-tier mtime cache fix introduced this regression
- #4808 — different root cause (session list cache hold-down during streaming), not related

## Environment

- Hermes WebUI running in Docker (overlay2 filesystem)
- 9 profiles, each with ~100+ skills (~500–600 files per profile skills directory)
- `ThreadingHTTPServer` concurrency model (one OS thread per request)

nesquena-hermes:
Fixes #5364.

## Independent confirmation

Verified all three claims in @claw-io's report against current `master` (v0.51.802):

1. **`_get_profile_skills_stats()` has no lock** — confirmed. `api/profiles.py` reads via `.get()` and writes the cache with no lock at all, unlike `_LIST_PROFILES_CACHE` which uses `_LIST_PROFILES_CACHE_LOCK`. With `ThreadingHTTPServer` (one OS thread per request), concurrent cold-boot misses all walk + parse simultaneously.
2. **`_build_profile_rows_fast()` runs outside the lock** — confirmed. In `list_profiles_api()` the lock only guarded the *read* of `_LIST_PROFILES_CACHE`; the build ran unlocked, so every concurrent request rebuilt all rows (each calling `_get_profile_skills_stats()` for every profile).
3. **The mtime probe fires on every call** — confirmed. `_skill_tree_max_mtime_ns()` (an `os.walk(followlinks=True)`) runs before the cache check on every call.

The report's amplification math holds: N concurrent `list_profiles_api()` × M profiles × 2 walks each = a herd of `os.walk` traversals, thousands of concurrent `stat()` calls under Docker overlay2 → the observed 57–70 s stalls.

## Fix

Two changes in `api/profiles.py` — the two high-impact ones from the report:

1. **Per-profile compute lock for `_get_profile_skills_stats()`** (double-checked locking). A concurrent burst of misses on the *same* profile collapses to a single `os.walk` + SKILL.md parse; the waiters serve the freshly-cached result. Independent profiles keep computing in parallel (per-profile lock, not one global lock — proven by a test).
2. **Single-flight the row build in `list_profiles_api()`** by holding `_LIST_PROFILES_CACHE_LOCK` across the build + cache write. One thread builds while the rest wait for the cache. Lock order is strictly *list-lock → per-profile skills-lock*, so there is no deadlock.

### Why not suggestion #3 (debounce the probe)

Deliberately **not** taken. The every-call cheap mtime probe is the #4783 out-of-band change-detection contract — `tests/test_issue4783_profile_skills_mtime_cache.py` explicitly asserts *"the cheap stat probe MUST run on every call to detect out-of-band changes."* Serializing the misses removes the herd without weakening that contract, because only the expensive compute is guarded, not the probe. A debounce would reintroduce a stale-count window for CLI/git skill edits, which is the exact bug #4783's design fixed.

## Tests

New `tests/test_issue5364_skills_stats_thundering_herd.py`:
- concurrent misses on one profile → **exactly one** compute (24-thread barrier burst);
- distinct profiles compute **concurrently** (per-profile lock, not global);
- per-profile lock registry returns a stable lock per profile;
- concurrent `list_profiles_api()` burst → **exactly one** row build (single-flight);
- #4783 contract preserved — the cheap probe still runs on every call.

All 6 new tests pass; all 10 existing #4783 contract tests still pass; adjacent profile suites (`test_issue3623`, `test_issue2698`, `test_issue4662`, `test_issue4650`) green.

Credit to @claw-io for the detailed diagnosis (root cause + thread dumps + amplification math) — carried as a `Co-authored-by` trailer.


nesquena-hermes:
## Gate — HOLD: full-suite test-isolation regression (self-built, needs a fix before merge) ⚠️

Codex verdict was SAFE (lock correctness verified: no deadlock, double-checked locking correct, #4783 mtime probe intact), and the targeted #5364 tests + isolated groups all pass. **But the full serial suite tells a different story that must be resolved first.**

### Empirical finding
- Full suite on this branch: **123 failed / 11394 passed** — reproduced twice.
- Full suite on clean `origin/master` (v0.51.804), same box/harness: **11509 passed, 2 known cron artifacts only.**
- Every one of the 123 failures passes **in isolation** and in **sub-group runs** (e.g. `test_issue538_mcp_management.py` + `test_provider_quota_status.py` together = 75/75). So it's an **ordering/shared-state interaction unique to this branch**, not a box flake.

### Root-cause signature
The failures share `mock_save.call_args[0][1]` → `TypeError: 'NoneType' object is not subscriptable` across ~40 test files — i.e. a patched `save`/build function is **never called**. That points at the new **single-flight caching in `list_profiles_api()` (`_LIST_PROFILES_CACHE_LOCK`)**: once one test populates the single-flight/list cache, later tests that expect the build (and patch its save) get the **cached short-circuited result across test boundaries**, so their mock never fires.

### What's needed
- Make the `list_profiles_api()` single-flight / list cache **resettable**, and clear it in the test teardown (autouse fixture), the same way `_SKILLS_STATS_CACHE.clear()` is already handled — OR scope the single-flight so it can't persist a build result across unrelated callers in a way tests depend on.
- Then re-run the FULL suite (not just targeted) and confirm it matches master's 2-artifacts-only baseline.

The per-profile skills-stats lock itself looks good — this is specifically the `list_profiles_api` single-flight cache leaking across the test suite. Holding (not merging) until the full suite is clean. I'll re-gate once the cache-reset is in.

nesquena-hermes:
Shipped in **v0.51.809**. Held on round 1 for a full-suite regression, root-caused to this PR's own test leaking `sys.modules` stubs (broke ~120 downstream tests when `get_config` saw the stub yaml); fixed the fixture to snapshot+restore all touched modules. Codex round 2 SAFE, full suite back to baseline (11537 passed). Thanks @claw-io for the report.

PR #5399, opened 2026-07-02
fix(sessions): always retry sidebar session-list GET on 502/503/504 (#5394)

### LINKED ISSUE #5394: Bug: sidebar session list stuck after 502 during profile switch — no retry on refreshSessionList
### Description

When switching profiles, if the backend briefly returns 502 (e.g. during a restart window or transient unavailability), the sidebar session list fails to refresh and stays stuck showing the **previous profile's sessions**. The user must manually Ctrl+F5 to recover.

### Root Cause

`_loadSidebarSessionListPayload` in `sessions.js` has no retry logic. When `/api/sessions` returns 502, `renderSessionList` receives raw HTML error page instead of JSON, parses it as the session list, and fails silently. The sidebar is left with stale data.

### Console Evidence

```
:3001/api/sessions?sidebar_source=webui&exclude_hidden=1:1  Failed to load resource: the server responded with a status of 502 (Bad Gateway)
:3001/api/projects:1  Failed to load resource: the server responded with a status of 502 (Bad Gateway)

sessions.js?v=v0.51.814:4362 renderSessionList Error: <html>
<head><title>502 Bad Gateway</title></head>
<body><center><h1>502 Bad Gateway</h1></center></body>
</html>

    at workspace.js?v=v0.51.814:60:21
    at async api (workspace.js?v=v0.51.814:69:25)
    at async _loadSidebarSessionListPayload (sessions.js?v=v0.51.814:4453:20)
    at async _runRenderSessionListRefresh (sessions.js?v=v0.51.814:4387:32)
    at async _drainRenderSessionListQueue (sessions.js?v=v0.51.814:4463:7)
    at async refreshSessionList (sessions.js?v=v0.51.814:4697:5)
```

### Steps to Reproduce

1. Have a multi-profile setup with nginx reverse proxy (HTTPS :3001 → Python :8787)
2. Switch profiles in WebUI sidebar
3. During the switch, if the backend is briefly unavailable (restart, config reload, or transient issue), nginx returns 502
4. Sidebar session list shows the old profile's sessions
5. No automatic recovery — user must Ctrl+F5

### Expected Behavior

`refreshSessionList` should retry on 502/network errors (e.g. exponential backoff with 2-3 retries) before giving up. Or at minimum, show a "Failed to load" indicator in the sidebar instead of silently keeping stale data.

### Environment

- WebUI: v0.51.814
- Reverse proxy: nginx (self-hosted, not WebUI-provided)
- OS: Linux (Ubuntu 22.04)

nesquena-hermes:
## Summary

Fixes #5394. The sidebar session list would get stuck after a transient 502
during a profile switch (or any warm refresh) and stay stale until a hard
reload (Ctrl+F5).

## The bug

`_runRenderSessionListRefresh()` in `static/sessions.js` already had
502/503/504 retry logic for the sidebar session-list GET — but it was gated
to **cold boot only**. Once the first load succeeded, `_sessionListHasLoadedOnce`
flipped to `true`, so every later refresh (profile switch, focus/visible/
reconnect) shipped **no `retryStatuses`**. A transient 502 during an
nginx→backend restart window then failed on the first attempt, leaving the
conversations list stale until the user hard-reloaded.

## Root cause (localized)

The retry options (`retries` + `retryStatuses`) lived inside the
`if(!_sessionListHasLoadedOnce){ … }` boot gate alongside the large boot
timeout, so warm refreshes never retried transient upstream errors.

## The fix

The session-list GET is idempotent, so retrying it is safe unconditionally.
This moves `retries:1` + `retryStatuses:[502,503,504]` into the base request
options so they apply on **every** refresh, while the boot-only path keeps the
larger `_SESSION_LIST_BOOT_TIMEOUT_MS` timeout + `retryTimeouts`:

```js
const sessionRequestOpts={
  timeoutToast:false,
  retries:1,
  retryStatuses:[502,503,504],
};
if(!_sessionListHasLoadedOnce){
  sessionRequestOpts.timeoutMs=_SESSION_LIST_BOOT_TIMEOUT_MS;
  sessionRequestOpts.retryTimeouts=true;
}
```

The `api()` wrapper in `static/workspace.js` already retries when the error
status is in `retryStatuses`, so no other change is needed. The `/api/projects`
fallback and all other logic are untouched — one localized block.

## Test

Extended the existing source-string regression test
(`test_sessions_api_always_retries_transient_upstream_statuses_and_boot_keeps_longer_timeout`
in `tests/test_session_sidebar_resilience.py`) to assert `retries:1` and
`retryStatuses:[502,503,504]` are declared **before** the boot-only gate (i.e.
always present), while the boot path still carries the timeout + timeout retry.
Adjusted one substring assertion in `tests/test_performance_polling_hardening.py`
to match the reformatted opts object.

All related sidebar/session tests pass:

```
tests/test_session_sidebar_resilience.py
tests/test_performance_polling_hardening.py
tests/test_issue4759_parallel_sidebar_boot_fetch.py
tests/test_api_timeout.py
tests/test_issue4766_sidebar_source_pushdown.py
→ 38 passed
```

## Credit

Reported and precisely root-caused by @weidzhou, who traced the boot-only
retry gate. Credited as co-author on the commit.

## Release note

Sidebar session list now retries transient 502/503/504 upstream errors on
every refresh (not just cold boot), so a brief backend restart during a
profile switch no longer leaves the conversations list stale.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -4439,12 +4439,20 @@ async function _runRenderSessionListRefresh(opts, _gen){
   try{
     if(!($('sessionSearch').value||'').trim()) _contentSearchResults = [];
     const sessionListQS = _sessionListQueryString();
-    const sessionRequestOpts={timeoutToast:false};
+    // #5394: the sidebar session-list GET is idempotent, so 502/503/504 retry
+    // must be unconditional. Previously retries/retryStatuses were boot-gated, so
+    // a transient 502 during an nginx->backend restart on a warm refresh (profile
+    // switch, focus/visible/reconnect) failed on the first attempt and left the
+    // sidebar stale until a hard reload. Boot still keeps the larger timeout +
+    // timeout retry; every refresh now retries the transient upstream statuses.
+    const sessionRequestOpts={
+      timeoutToast:false,
+      retries:1,
+      retryStatuses:[502,503,504],
+    };
     if(!_sessionListHasLoadedOnce){
       sessionRequestOpts.timeoutMs=_SESSION_LIST_BOOT_TIMEOUT_MS;
-      sessionRequestOpts.retries=1;
       sessionRequestOpts.retryTimeouts=true;
-      sessionRequestOpts.retryStatuses=[502,503,504];
     }
     const {sessData, projData}=await _loadSidebarSessionListPayload(sessionListQS, sessionRequestOpts);
     // Discard stale response — a newer renderSessionList() call superseded us.
--- a/tests/test_performance_polling_hardening.py
+++ b/tests/test_performance_polling_hardening.py
@@ -24,7 +24,7 @@ def test_session_list_refreshes_are_coalesced_while_in_flight():
     assert "_renderSessionListQueuedRequest={" in src
     assert "opts:_mergeRenderSessionListOptions" in src
     assert "if (_gen !== _renderSessionListGen) return" in src
-    assert "const sessionRequestOpts={timeoutToast:false};" in src
+    assert "const sessionRequestOpts={" in src
     assert "api('/api/sessions' + sessionListQS,sessionRequestOpts)" in src
     assert "api('/api/projects' + projectQS,{timeoutToast:false})" in src
 
--- a/tests/test_session_sidebar_resilience.py
+++ b/tests/test_session_sidebar_resilience.py
@@ -56,22 +56,44 @@ def test_sessions_and_projects_load_independently_so_projects_failure_cannot_bla
     assert "_applySessionListPayload(sessData,projData)" in block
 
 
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.822** 🚀 — thanks @weidzhou for the detailed report and root-cause.

The sidebar session-list GET now retries 502/503/504 on **every** refresh (profile switch, focus, reconnect), not just the cold boot load — so a transient upstream blip during an nginx→backend restart heals itself instead of leaving the sidebar stale until Ctrl+F5. Boot keeps the larger timeout + timeout-retry.

Gate: full suite green (no new failures) + Codex regression check SAFE-TO-SHIP (retry scoped to the idempotent GET, staleness/profile-switch guards intact). (#5394)

PR #5404, opened 2026-07-02
Hide inactive flyouts from screen readers

sheldon-im:
## Summary

This is a small screen-reader accessibility fix for the approval and clarification prompt cards.

When no approval or clarification was pending, NVDA browse mode could still find two dialogs:

- `Approval required`
- `Clarification needed`

From a user's point of view, this made it sound as if Hermes was always waiting for an approval or clarification, even when the agent was idle. Entering those dialogs also exposed their controls, which made the inactive prompts feel usable even though there was no active prompt to answer.

This PR hides those inactive prompt cards from screen readers. When an approval or clarification is actually pending, the cards still appear and keep the existing focus behavior.

## What changed

- Hide the inactive approval and clarification prompt cards from the accessibility tree.
- Re-expose each card only while its prompt is actually visible.
- Preserve the existing active-prompt behavior:
  - approval prompts still move focus to the approval controls;
  - clarification prompts still expose their input/choices;
  - resolving the prompt hides the card again.
- Add regression coverage for the inactive hidden state and the show/hide transitions.

## Technical notes

The cards were already visually hidden when inactive, but they were still present in the page structure with dialog roles. Visual hiding alone is not enough for screen readers: NVDA browse mode can still discover those dialogs.

This change marks inactive prompt cards with:

- `hidden`
- `aria-hidden="true"`
- `inert`

Those attributes are removed immediately before the card is shown for a real approval or clarification prompt, then restored when the prompt is dismissed or resolved.

## Manual validation

Validated in a live Hermes WebUI checkout with Windows/NVDA:

- Hard-refreshed the WebUI after applying the patch.
- Confirmed that, while the agent was idle and no prompt was pending, NVDA browse mode no longer exposed:
  - `Approval required`
  - `Clarification needed`
- Injected a loopback-only test approval prompt:
  - focus moved to the approval card;
  - approval controls were operable;
  - after responding, the dialog was hidden again.
- Injected a loopback-only test clarification prompt:
  - focus/control behavior remained usable;
  - after responding, the dialog was hidden again.
- Confirmed both pending queues were clear afterward.

## Verification

- `./scripts/test.sh tests/test_a11y_prompt_flyout_hidden_state.py`
- `node --check static/ui.js`
- `git diff --check`
- `.venv/bin/python scripts/ruff_lint.py --diff upstream/master`

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -468,7 +468,7 @@
       <div id="queueCard" class="queue-card" role="region" aria-label="Queued messages" aria-live="polite">
         <div id="queueChips" class="queue-card-inner"></div>
       </div>
-      <div class="approval-card" id="approvalCard" role="alertdialog" aria-labelledby="approvalHeading" aria-describedby="approvalDesc">
+      <div class="approval-card" id="approvalCard" role="alertdialog" aria-labelledby="approvalHeading" aria-describedby="approvalDesc" hidden aria-hidden="true" inert>
         <div class="approval-inner">
           <div class="approval-header">
             <svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2"><path d="M10.29 3.86L1.82 18a2 2 0 0 0 1.71 3h16.94a2 2 0 0 0 1.71-3L13.71 3.86a2 2 0 0 0-3.42 0z"/><line x1="12" y1="9" x2="12" y2="13"/><line x1="12" y1="17" x2="12.01" y2="17"/></svg>
@@ -504,7 +504,7 @@
           </div>
         </div>
       </div>
-      <div class="clarify-card" id="clarifyCard" role="dialog" aria-labelledby="clarifyHeading" aria-describedby="clarifyQuestion clarifyHint">
+      <div class="clarify-card" id="clarifyCard" role="dialog" aria-labelledby="clarifyHeading" aria-describedby="clarifyQuestion clarifyHint" hidden aria-hidden="true" inert>
         <div class="clarify-inner">
           <div class="clarify-header">
             <svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2"><path d="M12 17h.01"/><path d="M9.09 9a3 3 0 1 1 5.82 1c0 2-3 2-3 4"/><circle cx="12" cy="12" r="10"/></svg>
--- a/static/messages.js
+++ b/static/messages.js
@@ -6066,6 +6066,26 @@ const APPROVAL_MIN_VISIBLE_MS = 30000;
 
 // showApprovalCard moved above respondApproval
 
+function _setPromptFlyoutHidden(card, hidden) {
+  if (!card) return;
+  if (hidden) {
+    card.setAttribute("aria-hidden", "true");
+    card.setAttribute("inert", "");
+    const markHidden = () => {
+      if (!card.classList || !card.classList.contains("visible")) card.hidden = true;
+    };
+    if (typeof setTimeout === "function") setTimeout(markHidden, 450);
+    else markHidden();
+    return;
+  }
+  card.hidden = false;
+  card.setAttribute("aria-hidden", "false");
+  card.removeAttribute("inert");
+  // Force the unhidden, pre-visible state to be observed so the existing
+  // transform/opacity transition can still animate when `.visible` is added.
+  void card.offsetHeight;
+}
+
 function _clearApprovalHideTimer() {
   if (_approvalHideTimer) {
     clearTimeout(_approvalHideTimer);
@@ -6099,6 +6119,7 @@ function hideApprovalCard(force=false) {
   _resetApprovalCardState();
   card.classList.remove("visible");
   card.classList.remove("collapsed");
+  _setPromptFlyoutHidden(card, true);
   _syncApprovalTranscriptSpace(null);
   $("approvalCmd").textContent = "";
   $("approvalDesc").textContent = "";
@@ -6261,6 +6282,7 @@ function showApprovalCard(pending, pendingCount) {
     responding ? _approvalResponding.choice : null,
     responding,
   );
+  _setPromptFlyoutHidden(card, false);
   card.classList.add("visible");
   _syncApprovalCollapseButton(card);
   _syncApprovalTranscriptSpace(card, {immediate: true});
@@ -7038,6 +7060,7 @@ function _ensureClarifyCardDom() {
   card.setAttribute("role", "dialog");
   card.setAttribute("aria-labelledby", "clarifyHeading");
   card.setAttribute("aria-describedby", "clarifyQuestion clarifyHint");
+  _setPromptFlyoutHidden(card, true);
   card.innerHTML = `
     <div class="clarify-inner">
       <div class="clarify-header">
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Read the full diff against `origin/master` — `static/index.html`, the new `_setPromptFlyoutHidden` helper in `static/messages.js`, and the test file — plus the surrounding CSS. This is a well-scoped a11y fix and the approach is sound.

## Summary

The change stamps `hidden aria-hidden="true" inert` on the two prompt flyouts at rest (`#approvalCard`, `#clarifyCard`), then flips them off/on through a single helper wired into the existing show/hide entry points. `inert` + `aria-hidden` together take the idle dialogs out of the a11y tree and out of tab order, which fixes the NVDA "Approval required / Clarification needed" phantom dialogs.

## The load-bearing assumption checks out

The whole fix depends on the `hidden` attribute actually producing `display:none` on these cards. It does, but only because nothing overrides it — I grepped every `.approval-card`/`.clarify-card` rule (`static/style.css:1950`, `:2026`, the mobile blocks at `:3159`/`:3170`) and none set `display`, so the UA `[hidden]{display:none}` rule is unopposed:

```css
.approval-card{position:absolute;left:0;right:0;bottom:-24px;...overflow:hidden;pointer-events:none;}
.approval-card.visible{pointer-events:auto;z-index:3;}
```

Worth flagging because it's a silent-regression trap: if anyone later adds `display:flex` (or similar) to `.approval-card`, the `hidden` attribute stops hiding it and this a11y fix quietly breaks. The new test (`test_a11y_prompt_flyout_hidden_state.py`) only asserts source strings, so it wouldn't catch that. A one-line `.approval-card[hidden],.clarify-card[hidden]{display:none!important;}` guard rule (mirroring the pattern already used at `style.css:2603`/`:2694` for other `[hidden]` components) would make the fix robust against future CSS drift.

## Transition timing is correct

`_setPromptFlyoutHidden(card, true)` defers `card.hidden = true` by 450ms guarded on `.visible`:

```js
const markHidden = () => {
  if (!card.classList || !card.classList.contains("visible")) card.hidden = true;
};
if (typeof setTimeout === "function") setTimeout(markHidden, 450);
```

That 450ms clears the slide-out — both `.approval-inner` and `.clarify-inner` animate `transform .4s` (`style.css:1952`, `:2028`) — so `display:none` lands after the animation, not mid-flight. And the `.visible` guard means a re-show inside that window (show → hide → show) cancels the stale hide. The `void card.offsetHeight` reflow on the show path is the right call to keep the entrance transition playing after un-hiding. Good attention to detail here.

## Two minor notes

1. `aria-hidden="true"` alongside `inert` is redundant — `inert` already removes the subtree from the a11y tree — but it's harmless belt-and-suspenders and safe for older engines predating full `inert` support. No change needed; just noting it isn't doing independent work in modern browsers.

2. `_ensureClarifyCardDom` (`messages.js:7030`) returns the existing static `#clarifyCard` early when present, so the `_setPromptFlyoutHidden(card, true)` you added there only fires on the dynamic-creation branch. That's fine since the static HTML now carries the attributes inline, but it means the two paths rely on different sources for the initial hidden state — the CSS guard suggestion above would cover both uniformly.

Net: correct and minimal. The only thing I'd add before merge is the `[hidden]` display guard so a future CSS change can't silently regress the a11y behavior without a failing test.

sheldon-im:
Thank you for the review and for understanding the accessibility intent. The requested follow-up was small and straightforward, so I’ve applied it in the latest push.

nesquena-hermes:

**Certified head:** `sha:012c66d1` (clean rebase of PR head `7e97df60`, branch `gate-rebase/5404-a11y-flyout-hidden`) · **PR:** #5404 · _sheldon-im, Hide inactive flyouts from screen readers_
**Verdict:** Full gate GREEN. A proper a11y fix — inactive approval/clarify flyouts are removed from the accessibility tree (`aria-hidden`+`inert`) and hidden (`hidden`+`display:none`) while inactive, restored on show. Codex SAFE, suite green, lifecycle + animation-ordering verified. Ready to merge from the rebase branch.

### What I ran (rebased worktree, current head `7e97df60`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; node -c OK |
| Codex (reproduce) | **SAFE TO SHIP** — gated the rebased worktree, 0 findings (logic byte-identical to the re-push) |
| Full pytest suite | 1 failed / **11775 passed** — the 1 failure is the pre-existing nous-picker env flake (unrelated) |
| PR's own tests | ✅ 8/8 (`test_a11y_prompt_flyout_hidden_state` + `test_issue4771_gateway_approval_failure_ui`) |
| Lifecycle + animation-ordering (mine) | ✅ correct (below) |

### Findings
**✅ Correct a11y lifecycle:** `_setPromptFlyoutHidden(card, hidden)` manages `aria-hidden`+`inert`+`hidden` together. Hide: `aria-hidden="true"`+`inert` immediately (removes from a11y tree + tab order), defers `card.hidden=true` by 450ms guarded by `!classList.contains("visible")` (so a re-show cancels it). Show: `hidden=false`+`aria-hidden="false"`+remove `inert`. Applied to `approvalCard` (alertdialog) + `clarifyCard` (dialog), both initialized `hidden aria-hidden="true" inert` in index.html. `inert` is the modern correct way to neutralize an inactive overlay.

**✅ New `display:none` CSS doesn't block the entrance animation:** the re-push added `.approval-card[hidden],.clarify-card[hidden]{display:none!important;}` (complements the `hidden` attr with visual hiding). Verified the show sequence is correctly ordered: `_setPromptFlyoutHidden(card,false)` sets `card.hidden=false` (removes `[hidden]` → `display:none` gone) → `void card.offsetHeight` forces a reflow so the pre-`.visible` state is observed → the caller adds `.visible` (messages.js:6286/7396) to animate. So the transform/opacity transition still runs from the correct start state. No regression to approval/clarify behavior (#4771 test passes). Codex SAFE, suite 11775/0.

**Ready to merge — use branch `gate-rebase/5404-a11y-flyout-hidden` (`sha:012c66d1`), NOT the PR's stale head `7e97df60`.** Correct a11y fix (inert + aria-hidden + display:none lifecycle), animation ordering verified, Codex SAFE + 8/8 tests + suite green (the 1 failure is the pre-existing nous env flake, not this PR). Visible surface is attribute/state management (behavior test-covered, not a layout/render change). concept 4/5 (real a11y improvement — inactive dialogs shouldn't be reachable by screen readers / tab). Credit @sheldon-im (co-authored). crit=2.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master (current head 7e97df60, adds a `[hidden]{display:none}` line + test over the prior Codex-SAFE'd messages.js logic); lifecycle + show-animation-ordering verified, suite 11775/0. Cert valid for sha:012c66d1._

nesquena-hermes:
Shipped in **v0.51.830** ♿ — thanks @sheldon-im!

Merged via release #5443, with Nathan's sign-off.

**Full gate** (rebased onto current master v0.51.829):
- Codex regression gate: **SAFE TO SHIP** — animation ordering intact (the helper clears `hidden` and forces a reflow before `.visible`, so the entrance transition still runs), rapid hide→show guarded (deferred `hidden=true` checks `!classList.contains("visible")`), active cards remove `inert` before focus/actions, accept/reject/respond flow unchanged, `style.css` only suppresses `[hidden]` cards.
- Full pytest suite green (11805 passed). PR-own tests 8/8 (incl. the #4771 approval-flow test). Browser-smoke + ESLint runtime + node -c CLEAN.
- Live functional proof on the served fix: inactive card = `hidden`/`inert`/`aria-hidden=true`; after show = all cleared, `visible`, `display:block`. Screenshot-verified the approval card renders cleanly — the earlier "possible animation regression" flag is refuted.

Inactive approval/clarify dialogs are now removed from the accessibility tree + tab order (`aria-hidden` + `inert`) and visually hidden, so screen-reader and keyboard users no longer land on a dialog that isn't showing.

Credit preserved via `Co-authored-by`.

Issue #5367, opened 2026-07-01
Model Response Flickering when using Activity Display “Transparent Stream”


KShepherd-Kustoms:
When using Activity Display set to “Transparent Stream”, the response area visually flickers during model generation.

Instead of smoothly updating the message, the entire assistant response area (including text output, “thinking” sections, and tool call entries) appears to rapidly re-render or blink while the model is streaming.

This issue stops immediately once the response completes, and the final output renders normally.

Anth0ny29:
Hello,

Same here

nesquena-hermes:
## Summary

Reading the transparent-stream live render path in `static/ui.js` and `static/messages.js`, the flicker in Transparent Stream mode is a **teardown-and-rebuild** cycle: every streamed event re-runs the full scene render, which deletes all live activity rows and recreates them from scratch. Each freshly-created `.transparent-event-row` then replays its CSS entrance animation, so the whole activity area blinks token-by-token. This is distinct from the light-theme fade flicker fixed in #5328 (`0c13a172`, released ~v0.51.796), which only opted `#liveAssistantTurn` out of the global theme color transition — it does **not** touch the animation replay or the rebuild.

## Code reference

The live render (`renderLiveAnchorActivityScene`) wipes and rebuilds on every call — `static/ui.js:10814`:

```js
blocks.querySelectorAll(
  '.live-worklog[data-live-worklog-shell="1"],'+
  ...
  '.transparent-event-row[data-live-tid],'+
  '[data-live-stream-owned="1"],'+       // <- every live row from the prior render
  '.interim-collapse-toggle'
).forEach(el=>el.remove());
```

Each new row is tagged `data-live-stream-owned="1"` at `static/ui.js:10542`, so the next render removes it and `_anchorSceneTransparentNodeForRow` (`static/ui.js:10481`) builds a brand-new node. New nodes inside the live turn unconditionally get the entrance animation — `static/style.css:3980`:

```css
#liveAssistantTurn .transparent-event-row{animation:transparent-event-enter .18s ease-out both;}
```

```css
@keyframes transparent-event-enter{
  from{opacity:0;transform:translateY(2px);}
  to{opacity:1;transform:translateY(0);}
}
```

[...]

PR #5406, opened 2026-07-02
fix(webui): stop transparent stream row flicker

ruizanthony:
## Summary
- Disable the live entrance animation for Transparent Stream anchor-scene rows that are rebuilt during token/reasoning updates.
- Keep the existing entrance animation for normal live transparent rows.
- Add regression coverage so rebuilt live anchor rows stay at `opacity: 1` with no animation/transition replay.

## Root cause
`_renderLiveAnchorActivitySceneTransparent()` rebuilds visible anchor-scene rows as the live stream receives token/reasoning updates. The base live CSS applied `transparent-event-enter` to every `#liveAssistantTurn .transparent-event-row`, so each replacement row restarted at opacity 0 and produced a visible flicker even though the assistant source segment remained stable.

## Verification
- `python3 -m pytest tests/test_smooth_text_fade.py tests/test_issue3877_midstream_flicker.py tests/test_issue3820_chat_activity_display_mode.py -q`
- Browser replay probe against the live WebUI: sampled live `data-anchor-live-scene-row="1"` rows had `animation: none`, `opacity: 1`, and `transition: none`; base non-anchor rows still used `transparent-event-enter`.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -3978,6 +3978,17 @@ body.resizing .sidebar{transition:none!important;}
    base row class replayed the whole transcript's shimmer on every renderMessages
    rebuild. (Trifecta finding V9.) */
 #liveAssistantTurn .transparent-event-row{animation:transparent-event-enter .18s ease-out both;}
+/* Live transparent anchor scene rows are rebuilt during streaming as the anchor
+   projection receives token/reasoning updates. Replaying the entrance animation
+   on those replacement rows restarts opacity at 0 and produces visible flicker. */
+#liveAssistantTurn .transparent-event-row[data-anchor-live-scene-row="1"],
+#liveAssistantTurn .transparent-event-row[data-live-stream-owned="1"]{
+  animation:none!important;
+  opacity:1!important;
+  transition-property:none!important;
+  transition-duration:0s!important;
+  transition-delay:0s!important;
+}
 .transparent-event-row[data-event-type="thinking"]{
   background:transparent;
   border-color:transparent;
--- a/tests/test_smooth_text_fade.py
+++ b/tests/test_smooth_text_fade.py
@@ -234,6 +234,33 @@ def test_live_streaming_assistant_content_opts_out_of_global_theme_transitions()
     )
 
 
+def test_live_anchor_transparent_rows_do_not_replay_entrance_animation_on_stream_updates():
+    """Live Transparent Stream rows are rebuilt during token/reasoning updates.
+
+    The base live-row entrance animation starts at opacity 0. If it applies to
+    anchor-scene rows that are replaced on every stream update, the visible trace
+    flashes/fades repeatedly even when the assistant text node itself is stable.
+    """
+    assert "#liveAssistantTurn .transparent-event-row{animation:transparent-event-enter" in STYLE_CSS
+    guard = STYLE_CSS[
+        STYLE_CSS.index("Live transparent anchor scene rows are rebuilt during streaming") : STYLE_CSS.index(
+            ".transparent-event-row[data-event-type=\"thinking\"]"
+        )
+    ]
+    assert_contains_all(
+        guard,
+        [
+            '#liveAssistantTurn .transparent-event-row[data-anchor-live-scene-row="1"]',
+            '#liveAssistantTurn .transparent-event-row[data-live-stream-owned="1"]',
+            "animation:none!important",
+            "opacity:1!important",
+            "transition-property:none!important",
+            "transition-duration:0s!important",
+            "transition-delay:0s!important",
+        ],
+    )
+
+
 def test_stream_fade_css_is_opacity_only_and_hides_live_cursor():
     fade_css = STYLE_CSS[STYLE_CSS.index("OpenWebUI-style streaming word fade") :]
     assert "filter:" not in STYLE_CSS[STYLE_CSS.index("OpenWebUI-style streaming word fade") :].split(


nesquena-hermes:
Reading this diff against `origin/master` alongside the competing reconcile PR #5400 (both close #5367).

## Summary

This is the CSS-only take on the maintainer's **fix direction #1** ("suppress the entrance animation on rebuild"). It adds an `!important` override at `static/style.css` that neutralizes `animation`/`opacity`/`transition` on `#liveAssistantTurn .transparent-event-row[data-anchor-live-scene-row="1"]` and `[data-live-stream-owned="1"]`. Because the flicker root cause is the `transparent-event-enter` keyframe replaying on every teardown-and-rebuild during streaming, forcing `animation:none;opacity:1` on those rows does stop the blink. It's ~11 lines and can't regress the transport/render layers, which is a real point in its favor.

## One correctness note on the PR description

The PR says it will "keep the existing entrance animation for normal live transparent rows." That isn't what happens. Both attributes your selector matches are stamped on **every** live row, not just rebuilt ones. The live render loop passes `live:true` unconditionally — `static/ui.js:10851`:

```js
for(const row of rows){
  const node=_anchorSceneTransparentNodeForRow(row,{
    live:true,
    settled:false,
    ...
```

and the builder then stamps both attributes for any `live` row — `static/ui.js:10559` and `:10565`:

```js
if(live) node.setAttribute('data-anchor-live-scene-row','1');
...
if(live) node.setAttribute('data-live-stream-owned','1');
```

So in practice the entrance animation is disabled for **all** live transparent rows, including a row's genuine first appearance — not only the rebuilt ones. Only settled rows (the finalize path at `ui.js:10758+`, which never sets `data-anchor-live-scene-row`) keep the animation.

## Recommendation

That's arguably fine — an instantly-appearing live row is better than a blinking one — but it does not match the maintainer's stated goal of "the keyframe only plays on a row's first appearance" (issue #5367, direction #1). Achieving *first-appearance-only* isn't possible from CSS here, because the attribute can't distinguish a fresh row from a rebuilt one; that distinction requires row identity, which is exactly what the competing PR #5400 does by preserving the outer row element (direction #2). Worth deciding explicitly whether losing the entrance cue on new live rows is acceptable, and updating the PR body to reflect the actual behavior.

## Test note

`test_live_anchor_transparent_rows_do_not_replay_entrance_animation_on_stream_updates` asserts the CSS string is present, which is a fine guard for the rule not being deleted, but it can't catch the "all rows vs rebuilt rows" semantics above since it never renders. As greptile noted, `str.index()` slicing will raise a bare `ValueError` rather than a clean assertion if either anchor string moves; the repo's `slice_between`/`assert_contains_all` helper would read better.

Net: this and #5400 are two valid answers to the same issue via different directions — one CSS-simple (kills the animation for all live rows), one JS-structural (preserves it for new rows only). Flagging so they aren't both merged.

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (right instinct, one line too broad: `opacity:1!important` clobbers the graduated depth-fade)

**Certified head:** `sha:e553a2f3` (rebased onto current master, `git apply` clean) · **PR:** #5406 · _ruizanthony, fix(webui): stop transparent stream row flicker_
**Verdict:** The fix correctly identifies that replaying the entrance animation on rebuilt live rows causes the flicker — but the new guard adds `opacity:1!important`, which overrides the existing graduated `[data-transparent-fade]` depth-fade (older transparent rows are meant to fade to ~0.54 during streaming). Drop the `opacity:1`, keep `animation:none` + the transition reset, and it's a clean flicker fix.

### What I ran (rebased worktree `/tmp/wt-rebase-5406`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 1 SILENT (opacity clobbers depth-fade); I confirmed |
| Full pytest suite | 1 failed / 11771 passed — the 1 failure is the pre-existing nous env flake (unrelated) |
| PR's own test | ✅ 11/11 |
| Selector-match + scoping (mine) | ✅ selectors match real DOM (below); the opacity over-reach confirmed (below) |

### Findings
**✅ Correct core + scoping:** the rule targets live-turn transparent rows only (`#liveAssistantTurn .transparent-event-row[data-anchor-live-scene-row="1"],[data-live-stream-owned="1"]`) — self-clearing on settle (id cleared, like #5328). Selectors match real DOM (`data-anchor-live-scene-row` / `data-live-stream-owned` / `transparent-event-row` are set in `assistant_turn_anchors.js`/`ui.js` — not a no-op). `animation:none!important` + transition reset correctly stops the entrance-animation replay that causes the flicker.

**⛔ SILENT (I CONFIRMED) — `opacity:1!important` clobbers the graduated depth-fade** (`static/style.css:3990`): there's an existing `[data-transparent-fade="1..5"]` system (opacity 0.86→0.54, style.css:4288-4293) that `_syncTransparentEventControls` applies during live streaming (ui.js:9959 `setAttribute('data-transparent-fade', step)`) to fade OLDER transparent rows by depth. The new rule's `opacity:1!important` on `[data-anchor-live-scene-row]`/`[data-live-stream-owned]` rows overrides that (`!important` beats the plain-specificity fade rules) → older rows that should fade to ~0.54 stay at full opacity, regressing the intended depth-fade. The fix targets the entrance-animation flicker, not the depth-fade — so forcing opacity is too broad.
- **Fix (Codex-exact, one line):** remove `opacity:1!important` from the new guard; keep `animation:none!important` + `transition-property/duration/delay:none!important`. That stops the entrance-animation replay (the flicker) while `[data-transparent-fade]` still controls opacity (the intended fade).

**RED — gate-fail/changes-requested (one-line fix): drop `opacity:1!important` from the new guard, keep `animation:none` + the transition reset.** The flicker diagnosis + scoping + self-clearing are right; only the opacity clause over-reaches into the depth-fade system. Trivial converge. Once green, it's a motion-artifact visual fix → a light before/after would be nice (like #5328), but the mechanism is verifiable via CSSOM. concept 4/5 (real streaming flicker fix). Author @ruizanthony. crit=3. (Note: this and #5400 both target transparent-stream rendering but are independent — #5400 is JS reconciliation, #5406 is CSS; both currently RED with small fixes.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; selectors verified real, the opacity-over-reach confirmed against the `[data-transparent-fade]` depth-fade rules (4288-4293) + its live-stream applier (ui.js:9959). Cert valid for sha:e553a2f3._

nesquena-hermes:
Closing as superseded — thanks @ruizanthony for the parallel effort here.

The transparent-stream flicker this targets (#5367) was fixed at the root by @rodboev's #5400 (row-identity reconcile for anchor-scene live rows), which shipped in release #5433 (v0.51.827). I confirmed the reconcile machinery is live in master's `static/ui.js`, and #5367 is closed as completed. This narrower "disable live entrance animation on rebuilt anchor-scene rows" approach for the same flicker is now moot.

Appreciate the fix — the root cause just landed via a different PR first. If you still see flicker on the current release, please reopen with a repro and we'll dig in.

PR #5409, opened 2026-07-02
fix: eliminate transparent stream flickering during token-by-token rendering

luperrypf:
## Problem

When Transparent Stream mode is enabled, the UI visibly flickers during token-by-token streaming. Each new tool execution event row triggers a CSS enter animation that appears as a brief flash/shimmer.

## Root Cause

`style.css` line 3988 — new `.transparent-event-row` elements in the live turn animate `opacity: 0→1` and `transform: translateY(2px→0)` over 180ms:

```css
#liveAssistantTurn .transparent-event-row {
  animation: transparent-event-enter .18s ease-out both;
}

@keyframes transparent-event-enter {
  from { opacity: 0; transform: translateY(2px); }
  to   { opacity: 1; transform: translateY(0); }
}
```

During rapid streaming, this animation fires on every new row, creating visible flicker.

## Solution

Replace the enter animation with `animation: none !important`:

```css
#liveAssistantTurn .transparent-event-row {
  animation: none !important;
}
```

The `#liveAssistantTurn` scope ensures this only affects the live turn during streaming — settled/historical messages are untouched.

The transition suppression is **not needed** here because `#liveAssistantTurn *` (lines 1464–1465) already provides `transition-property: none !important` for all live-turn descendants, including transparent stream elements.

## What's preserved

- ✅ `data-transparent-fade` recency opacity transition on settled messages
- ✅ Hover background/border transitions on all transparent event rows
- ✅ Detail mode tab switching transitions
- ✅ All other existing behavior

## Before/After

**Before**: New event rows fade in + slide down over 180ms → visible flicker during streaming  
**After**: New event rows appear instantly → no visual artifacts

## Testing

Tested on iOS Safari with cache cleared. Sent multi-step commands in Transparent Stream mode. Event rows appear instantly without flicker.

## Release Note

- **Fixed:** Transparent Stream no longer flickers during token-by-token streaming. The 180ms enter animation on new event rows caused visible shimmer when rows were rapidly added. Replaced with `animation: none !important`, scoped to `#liveAssistantTurn` so settled message transitions are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Fixed
 
+- **Transparent Stream no longer flickers during token-by-token streaming.** The 180ms enter animation (`transparent-event-enter`) on new event rows caused visible shimmer when rows were rapidly added. Replaced with `animation: none !important`, scoped to `#liveAssistantTurn` so settled message transitions (hover, recency-fade) are unaffected.
+
 - **The composer footer shows your model and workspace names again on desktop.** The Export-to-HTML button added to the composer footer in the previous release pushed the control row past its width budget, which collapsed the footer into icon-only mode and hid the model / workspace / profile text labels. The export button has been removed from the composer footer (export remains available from Settings), so the labels are visible again.
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,6 @@
 
 ### Fixed
 
-- **Transparent Stream no longer flickers during token-by-token streaming.** The 180ms enter animation (`transparent-event-enter`) on new event rows caused visible shimmer when rows were rapidly added. Replaced with `animation: none !important`, scoped to `#liveAssistantTurn` so settled message transitions (hover, recency-fade) are unaffected.
-
 - **The composer footer shows your model and workspace names again on desktop.** The Export-to-HTML button added to the composer footer in the previous release pushed the control row past its width budget, which collapsed the footer into icon-only mode and hid the model / workspace / profile text labels. The export button has been removed from the composer footer (export remains available from Settings), so the labels are visible again.
 
 ### Added
--- a/static/style.css
+++ b/static/style.css
@@ -1472,15 +1472,7 @@
   .agent-activity-thinking[data-live-thinking="1"],
   .agent-activity-thinking[data-live-thinking="1"] *,
   .live-worklog[data-live-worklog-shell="1"],
-  .live-worklog[data-live-worklog-shell="1"] *,
-  #liveAssistantTurn .transparent-event-row,
-  #liveAssistantTurn .transparent-event-row *,
-  #liveAssistantTurn .transparent-event-controls,
-  #liveAssistantTurn .transparent-event-controls *,
-  #liveAssistantTurn .transparent-turn-footer,
-  #liveAssistantTurn .transparent-turn-footer *,
-  #liveAssistantTurn .transparent-detail-modes,
-  #liveAssistantTurn .transparent-detail-modes *{transition-property:none!important;transition-duration:0s!important;transition-delay:0s!important;}
+  .live-worklog[data-live-worklog-shell="1"] *{transition-property:none!important;transition-duration:0s!important;transition-delay:0s!important;}
   :root{--app-titlebar-safe-top:0px;}
   .pwa-standalone{overscroll-behavior:none;}
   .pwa-standalone body{-webkit-tap-highlight-color:transparent;}
--- a/static/style.css
+++ b/static/style.css
@@ -4266,10 +4266,6 @@ body.resizing .sidebar{transition:none!important;}
   .transparent-event-row .thinking-card-header{min-height:30px;padding:5px 8px;}
   .transparent-detail-mode{padding:4px 6px 5px;}
 }
-@keyframes transparent-event-enter{
-  from{opacity:0;transform:translateY(2px);}
-  to{opacity:1;transform:translateY(0);}
-}
 
 /* ── Old-event fading (transparent mode: medium → low) ─────────────── */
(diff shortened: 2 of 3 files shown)


luperrypf:
Good catch — you're right. `#liveAssistantTurn *` already covers all descendants, so the added transition exclusions were redundant. Removed them in 8d36608.

The actual fix is now a single line change: disabling the `transparent-event-enter` animation that was causing the visible flicker during streaming.

nesquena-hermes (changes requested):
Thanks for tackling the transparent-stream flicker — disabling the entrance animation on the live-turn rows is a reasonable lever, and the CSS change itself is small and targeted.

There's one objective blocker before this can go green, plus a couple of notes.

### Blocker — this breaks an existing test (CI will be red)

Your diff changes `static/style.css`:

```css
-#liveAssistantTurn .transparent-event-row{animation:transparent-event-enter .18s ease-out both;}
+#liveAssistantTurn .transparent-event-row{animation:none!important;}
```

But `tests/test_issue3820_chat_activity_display_mode.py` still hard-asserts the old rule is present:

```python
# line 646
def test_transparent_entrance_animation_is_live_turn_only():
    """The entrance animation must be scoped to the live turn so it doesn't
    replay across the whole transcript on every renderMessages. (Trifecta V9.)"""
    assert "#liveAssistantTurn .transparent-event-row{animation:transparent-event-enter" in STYLE_CSS
```

Since you removed that literal, this assertion now fails (`master` green → head red; the test file itself is unchanged in your PR). The fix is mechanical: update `test_transparent_entrance_animation_is_live_turn_only` (and its docstring) to reflect the new intent — e.g. assert the live-turn entrance animation is disabled (`"#liveAssistantTurn .transparent-event-row{animation:none" in STYLE_CSS`) rather than present — or replace the V9 guard with one that still protects against the original bug (the animation replaying across the whole transcript on every `renderMessages` rebuild) under the new no-animation approach.

### Notes

- **No CHANGELOG entry** — please add one for the fix.
- **UX / direction is a maintainer call:** removing the entrance animation entirely (vs. keeping a subtler entrance and fixing the flicker at its root) is a visible behavior change, and there are currently a couple of other open PRs also targeting transparent-stream flicker. A maintainer will want to confirm the intended direction, so it's worth aligning the test-fix with that. Happy to re-review as soon as the test is updated and CI is green.

luperrypf:
Addressed both points:

1. **Test updated** — `test_transparent_entrance_animation_is_live_turn_only` now asserts `animation:none` instead of the old `transparent-event-enter` literal, with the docstring updated to reflect the new intent.

2. **CHANGELOG added** — entry under `[Unreleased] → Fixed`.

CI should be green now.

nesquena-hermes:
Reviewed at `9f7332da`. I pulled the branch and read the full diff against `origin/master` — `static/style.css`, `tests/test_issue3820_chat_activity_display_mode.py`, and `CHANGELOG.md`. The CSS change is correct and well-scoped; one process nit on the changelog.

## The fix is right

The entrance animation is what caused the flicker. On master, `static/style.css:3983`:

```css
#liveAssistantTurn .transparent-event-row{animation:transparent-event-enter .18s ease-out both;}
```

The live activity scene rebuilds its rows on every streamed event (the teardown-and-rebuild in `renderLiveAnchorActivityScene`, `static/ui.js`), so each rebuilt `.transparent-event-row` replayed the `opacity:0->1` + `translateY(2px->0)` keyframe dozens of times a second during prose streaming — that's the blink reported in #5367. Your change at `static/style.css:3982` neutralizes it:

```css
#liveAssistantTurn .transparent-event-row{animation:none!important;}
```

Scoping to `#liveAssistantTurn` is the key correctness point: settled/historical `.transparent-event-row` elements keep their base styles, so hover transitions and the `data-transparent-fade` recency-fade are untouched. And dropping the added `transition-property` exclusions (per luperrypf's follow-up in `8d36608`) is correct — `#liveAssistantTurn *{transition-property:none!important;...}` at `static/style.css:1475` already covers every live-turn descendant, so those were redundant.

Deleting the now-orphaned `@keyframes transparent-event-enter` block is clean; I grepped and nothing else references it.

The test update at `tests/test_issue3820_chat_activity_display_mode.py:643-651` correctly re-asserts the new value while preserving the `#liveAssistantTurn` scope-guard intent (the Trifecta V9 invariant that stops any future base-class animation from replaying across the whole transcript):

```python
assert "#liveAssistantTurn .transparent-event-row{animation:none" in STYLE_CSS
```

## One process nit: drop the CHANGELOG.md edit

`CHANGELOG.md` is edited directly in this PR, which is against `AGENTS.md:63-65`:

> Do not edit `CHANGELOG.md` in ordinary contributor PRs. The release workflow owns changelog updates through release commits.

Greptile flagged the same thing. No runtime risk, but it'll conflict with the release automation — please revert the two `CHANGELOG.md` lines and put the release-note wording in the PR body instead.

## Note on the overlapping approach

This is the animation-suppression path (fix #1 from the #5367 analysis). #5400 takes the alternative reconcile-in-place path (fix #2) in `static/ui.js`. Both target the same flicker; this one is the smaller, lower-risk change. Worth the maintainer deciding whether the durable reconcile fix supersedes this or whether this ships as the immediate mitigation. Either way, once the CHANGELOG edit is reverted this is good to go on its own merits.

nesquena-hermes:
Closing as superseded — thanks @luperrypf.

Same story as the sibling PR for this issue: the transparent-stream token-by-token flicker (#5367) was fixed at the root by #5400 (anchor-scene row-identity reconcile), shipped in release #5433 (v0.51.827) and confirmed live in master's `static/ui.js`; #5367 is closed as completed. This variant for the same flicker is now moot.

Thanks for the effort — the root fix just landed via another PR first. If flicker persists on the latest release, reopen with a repro.

PR #5410, opened 2026-07-02
fix(logging): access logs survive in-process sys.stdout hijack (maint #0095)

DevNexsler:
The WebUI runs agent turns in-process on worker threads, and agent tool
dispatch swaps the process-wide sys.stdout/sys.stderr to a devnull around
every tool call with a finally-restore. Concurrent turns race those
save/restore pairs, which can leave sys.stdout permanently bound to a
closed file object. _safe_webui_print swallowed the resulting ValueError
(by design, so logging can't break responses), so every [webui] access-log
line and request-error traceback was dropped silently until a restart:
the service kept serving but went log-blind for ~14h in production
(journal dead after 2026-07-01 12:07 while /api/health/agent served 200s).

Fix: new api/log_stream.py duplicates the real stdout fd once at import
(server startup, before any agent code can touch the streams) and
durable_print() writes log lines straight to that private fd under a lock,
never raising. Handler._safe_webui_print delegates to it, so no in-process
rebinding, closing, or capture of sys.stdout (devnull swap, stdout=stderr,
StringIO capture, ...) can silence access/error logging anymore.

Tests: tests/test_access_log_stdout_hijack.py reproduces both hijack shapes
(swapped-for-capture and permanently-closed sys.stdout) and asserts the
[webui] line still reaches the real fd; red on the old code, green now.
tests/test_issue2775_log_request.py now pins the durable stream to its
print() fallback so capsys keeps observing record content.

Full suite: 2 failed, 11697 passed, 85 skipped, 1 xfailed, 2 xpassed —
the 2 failures (test_pr1970_lmstudio_base_url_fallback) are pre-existing
test-ordering pollution, failing identically on the unmodified base commit
in the same environment and passing in isolation.

The racing global-stdout swap itself lives upstream in hermes-agent's tool
dispatch; flagged there separately — the webui must be robust to it anyway.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (fd-capture is well-built; one `os.write`-returns-0 livelock in the write loop)

**Certified head:** `sha:df2d4142` (rebased onto current master, `git apply` clean) · **PR:** #5410 · _DevNexsler, fix(logging): access logs survive in-process sys.stdout hijack (maint #0095)_
**Verdict:** The fd-capture mechanism is textbook-clean (dup-at-import, `set_inheritable(False)`, thread-safe recapture with previous-fd close, fails closed) — but the `durable_print` partial-write loop can **livelock if `os.write` returns 0**: `data = data[0:]` leaves the buffer unchanged and spins forever while holding `_write_lock`, hanging that request's logging and blocking all later logs behind the lock. Guard `written <= 0` and it's ready.

### What I ran (rebased worktree `/tmp/wt-rebase-5410`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean (merge-base == master tip) |
| Codex (reproduce) | SHIP-WITH-FIXES — 1 CORE (os.write==0 livelock); I confirmed by reading the loop |
| Full pytest suite | 1 failed / 11772 passed — the 1 failure is the pre-existing nous env flake (unrelated) |
| PR's own test | ✅ 3/3 (`test_issue2775_log_request.py`) |
| Mechanism review (mine) | fd-capture lifecycle correct (below); the write-loop edge confirmed (below) |

### Findings
**✅ fd-capture lifecycle is correct:** `capture_stdout_fd()` (called at import, `server.py`) does `os.dup(1)` into a private `_log_fd`, `set_inheritable(False)` (no fd leak to children), under `_write_lock`; on recapture it sets `_log_fd=None` first, closes the `previous` fd (guarded), then dups fresh — no double-close/leak, idempotent. `os.dup` failure → `_log_fd=None` → falls back to `print()`. `durable_print` is thread-safe (`_write_lock`), UTF-8 `errors='replace'`, `try/except: pass` (never raises). This correctly keeps access logs alive when agent code hijacks `sys.stdout` (the #0095 goal).

**⛔ CORE (I CONFIRMED) — `os.write` returning 0 livelocks the write loop while holding the lock** (`api/log_stream.py:60`): the loop is `while data: data = data[os.write(_log_fd, data):]`. If `os.write` returns `0` (possible on certain fd states — it doesn't raise), the slice is `data[0:]` = the same buffer → `while data:` spins **forever**, and because it's inside `with _write_lock`, that request hangs in logging AND every subsequent `durable_print` blocks behind the never-released lock (cascading log/request-logging hang). The outer `except Exception: pass` is never reached (no exception raised on a 0-return). For an infra path explicitly designed to be best-effort and never-block, an infinite lock-holding loop is a real defect.
- **Fix (Codex-exact, ~2 lines):** capture the count — `written = os.write(_log_fd, data)`; if `written <= 0`, `raise OSError` (so the outer never-raise guard recovers) or `break` before slicing; else `data = data[written:]`. Add a test that stubs `os.write` to return 0 and asserts `durable_print` returns (doesn't hang) + falls back.

**RED — gate-fail/changes-requested (1 CORE, ~2-line fix): guard `os.write` returning 0 in the `durable_print` loop (`written<=0` → raise/break) so it can't livelock while holding `_write_lock`.** The fd-capture design is otherwise clean and correct (dup-at-import, inheritable-false, safe recapture, fails-closed, thread-safe) — this is just hardening the write loop against the 0-return edge on an infra path that must never block. Converges trivially. concept 4/5 (real #0095 reliability fix — logs surviving stdout hijack is valuable). Author @DevNexsler (unproven). crit=3, infra reliability. (Good catch: a partial-write loop `data=data[os.write(...):]` is a classic livelock trap when the syscall can return 0 without raising — always guard the count.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; fd-capture lifecycle verified correct (dup/inheritable/recapture-close/fails-closed/thread-safe), the os.write==0 livelock confirmed by reading the loop (`data=data[0:]` under held `_write_lock`, except never reached). Cert valid for sha:df2d4142._

DevNexsler:
Closing — this was opened against upstream by our internal maintenance automation by mistake; the change is maintained in our fork. Apologies for the noise.

PR #5419, opened 2026-07-02
fix: auto-switch profiles for session links

harcek:
## Thinking Path

A valid `session://...` deep link can point at a session owned by a different Hermes profile than the currently-active browser profile. Today that looks identical to a deleted/missing session: `/api/session` returns `404`, the frontend self-heals the route/localStorage, and the user sees “Session not available in web UI” or falls back to another session.

## What Changed

- `GET /api/session` now distinguishes valid-but-wrong-profile sessions from missing sessions.
- Wrong-profile detail loads return structured `409 session_profile_mismatch` with the owning profile name.
- `loadSession()` catches that response, switches to the owning profile, and retries the session load once.
- True missing/deleted sessions still use the existing `404` self-heal path.

## Why It Matters

Cross-profile session links should open the real conversation, not imply data loss. This preserves the existing profile isolation boundary while making direct links and `session://` links usable across profiles.

## Verification

- `./scripts/test.sh tests/test_bugfix_sweep.py tests/test_issue803.py -q`
  - `45 passed`
- Manual local smoke on a multi-profile install:
  - opening a `deepresearcher` session while active profile was `default` returned `409 session_profile_mismatch`
  - after profile switch, the same session loaded successfully

## Risks / Follow-ups

Low risk: the change is scoped to session detail loads. Mutation endpoints still use the existing request profile guard. Missing/deleted sessions keep returning `404`.

## Model Used

AI-assisted by Hermes Agent using OpenAI Codex `gpt-5.5`.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -518,6 +518,11 @@ def _session_visible_to_active_profile(session_profile, handler=None) -> bool:
 def _request_session_visibility_exempt(method: str, path: str | None) -> bool:
     if not path:
         return False
+    if method == "GET" and path == "/api/session":
+        # Detail-load owns profile mismatch handling so the frontend can switch
+        # to the session's profile instead of treating a valid cross-profile
+        # deep link as a deleted/stale session.
+        return True
     if method != "POST":
         return False
     # Import routes create/claim sessions before normal ownership exists, and
@@ -11579,7 +11584,12 @@ def handle_get(handler, parsed) -> bool:
             s = get_session(sid, metadata_only=(not load_messages))
             _session_profile = getattr(s, 'profile', None) or None
             if not _session_visible_to_active_profile(_session_profile, handler):
-                return bad(handler, "Session not found", 404)
+                return j(handler, {
+                    "error": "Session belongs to a different profile",
+                    "code": "session_profile_mismatch",
+                    "session_id": sid,
+                    "profile": _session_profile,
+                }, status=409)
             original_stream_id = getattr(s, "active_stream_id", None)
             _clear_stale_stream_state(s)
             cli_meta = _lookup_cli_session_metadata(sid) if _session_requires_cli_metadata_lookup(s) else {}
@@ -11906,7 +11916,12 @@ def handle_get(handler, parsed) -> bool:
             cli_meta = _lookup_cli_session_metadata(sid)
             _session_profile = (cli_meta or {}).get("profile") or None
             if not _session_visible_to_active_profile(_session_profile, handler):
-                return bad(handler, "Session not found", 404)
+                return j(handler, {
+                    "error": "Session belongs to a different profile",
+                    "code": "session_profile_mismatch",
+                    "session_id": sid,
+                    "profile": _session_profile,
+                }, status=409)
             synth, reason = _claim_or_synthesize_cli_session(sid, cli_meta=cli_meta or {})
             if reason == "was_webui":
                 # Deleted WebUI session: 404 so the client self-heals
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -1176,6 +1176,41 @@ function _rearmActiveSessionStream(){
   if(activeSid) startSessionStream(activeSid);
 }
 
+function _sessionProfileMismatchFromError(e){
+  if(!e || e.status!==409 || !e.body) return null;
+  try{
+    const body=JSON.parse(e.body);
+    if(body && body.code==='session_profile_mismatch' && body.profile){
+      return {profile:String(body.profile), session_id:String(body.session_id||'')};
+    }
+  }catch(_){ }
+  return null;
+}
+
+async function _switchProfileForSessionLoad(profile){
+  const name=String(profile||'').trim();
+  if(!name) throw new Error('missing profile');
+  if(name===S.activeProfile) return;
+  if(typeof _invalidateSessionListRenders==='function') _invalidateSessionListRenders();
+  if(typeof _setProfileSwitchListEmbargo==='function') _setProfileSwitchListEmbargo(true);
+  if(typeof showSessionListSkeleton==='function') showSessionListSkeleton(name);
+  try{
+    const data=await api('/api/profile/switch',{method:'POST',body:JSON.stringify({name}),timeoutToast:false});
+    S.activeProfile=data.active||name;
+    S.activeProfileIsDefault=!!data.is_default;
+    if(typeof _clearPersistedModelState==='function') _clearPersistedModelState();
+    else localStorage.removeItem('hermes-webui-model');
+    if(data.default_model) window._defaultModel=data.default_model;
+    if(data.default_model_provider) window._activeProvider=data.default_model_provider;
+    if(typeof startGatewaySSE==='function') startGatewaySSE();
+    if(typeof syncTopbar==='function') syncTopbar();
+    if(typeof _setProfileSwitchListEmbargo==='function') _setProfileSwitchListEmbargo(false);
+    if(typeof renderSessionList==='function') await renderSessionList();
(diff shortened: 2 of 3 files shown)


nesquena-hermes (changes requested):
Thanks for this — the core idea (turn a valid cross-profile deep link into a profile-switch instead of a misleading "session not available") is a real UX improvement, and the data boundary is preserved (the new `409` returns only an error envelope; I confirmed no foreign-profile transcript/messages are served). One objective blocker before this can land, plus two framing notes for a maintainer decision.

## ⛔ Blocker: this breaks 4 existing regression tests on the profile-isolation boundary

`tests/test_issue1611_session_profile_filtering.py` was written specifically for the cross-profile `GET /api/session` boundary, and it hard-asserts the **old 404** behavior. This PR changes that path from `404` to `409` (via `j(...)`) but never reconciles that suite, so 4 tests now fail:

```
tests/test_issue1611_session_profile_filtering.py
  test_get_session_rejects_session_from_inactive_profile               (:312)
  test_get_session_rejects_metadata_only_session_from_inactive_profile (:337)
  test_get_session_rejects_cookieless_session_from_inactive_profile    (:362)
  test_get_session_rejects_cli_session_from_inactive_profile           (:390)
```

Each asserts `captured["bad"]["status"] == 404` and `"json" not in captured` ("foreign-profile transcript must not be returned"). Post-change the response goes through `j(...)` at status 409, so `bad` is never called and `captured["json"]` is now present → both assertions fail. Reproduced on your head (`0bfbb03`): **4 failed, 3 passed**.

### Fix-spec
Update those 4 tests in `tests/test_issue1611_session_profile_filtering.py` to the new contract while keeping the original intent (no foreign-profile transcript leaks):
- Expect the `409` envelope: `status == 409`, `data["code"] == "session_profile_mismatch"`, `data["profile"]` is the owning profile name.
- Keep the isolation assertion: verify the response body contains **only** the error envelope (`error`/`code`/`session_id`/`profile`) and **no** session transcript/content/title keys — that's the boundary the suite exists to protect, and it still holds, so the assertion should be reworded, not dropped.

## Note (not blocking): the new test in this PR is source-grep only

`test_bugfix_sweep.py::test_cross_profile_session_deep_links_switch_profile_instead_of_self_healing` only `read_text()`s the two files and asserts literal substrings exist. It would still pass if the status code were wrong or the retry-loop guard were broken. A behavioral test (the 409 envelope end-to-end, and the frontend one-shot switch-and-retry) would give real coverage — worth adding alongside the `test_issue1611` fixups.

## Note for a maintainer call: information-disclosure delta

Previously a cross-profile session and a truly-missing session were both `404` (indistinguishable). The `409` now makes `/api/session` a **session-existence oracle** and reveals the **owning profile name** for a given session id. Mitigating context: profile names are already fully enumerable by any authenticated WebUI user (`GET /api/profiles` lists them; `POST /api/profile/switch` accepts any existing profile with no per-user ACL), so the profile *name* isn't new privileged info — but the existence-confirmation + `session_id → owning-profile` binding is new. Flagging so a maintainer can confirm that's acceptable for the profile model; not a defect on its own.

Happy to re-review as soon as the `test_issue1611` suite is reconciled. Nice fix otherwise.

nesquena-hermes:
Reading `api/routes.py:522-534` (the `_request_session_visibility_exempt` change) plus the two detail-load branches at `routes.py:11584-11591` and `routes.py:11916-11923`, and the frontend recovery path in `static/sessions.js:1179-1211` and `1342-1356`, this is a clean, well-scoped fix. A few notes to confirm it holds together.

## The exemption is safe because both GET `/api/session` branches still enforce the boundary

The new early-return in the guard makes GET `/api/session` bypass the generic `_guard_request_session_visibility` sweep (called at `routes.py:11233` for all `/api/` GETs):

```python
if method == "GET" and path == "/api/session":
    # Detail-load owns profile mismatch handling ...
    return True
```

That would be a visibility hole *if* the detail handler didn't re-check. It does — both the primary branch (`get_session` → `_session_visible_to_active_profile`) and the CLI/foreign fallback in the `except KeyError` block (`cli_meta` → `_session_visible_to_active_profile`) still gate on the active profile before returning any payload; the PR just swaps their `404` for the structured `409`. So no session data leaks; a wrong-profile load returns metadata-free `{"code":"session_profile_mismatch","profile":...}`. Good.

## Frontend error contract matches

`_sessionProfileMismatchFromError` reads `e.status===409` and `e.body`. That lines up with the `api()` helper in `static/workspace.js:56-63`, which attaches exactly those fields on a non-ok response:

```js
const err=new Error(message);
err.status=res.status;
err.statusText=res.statusText;
err.body=text;
```

So the parse path is correct.

## Boot deep-links are covered

Worth calling out for reviewers: the URL deep-link path on a cold load goes through `loadSession()` (`static/boot.js:3174`, `await loadSession(saved, ...)`), which is exactly where the 409 recovery was wired. The other cold-load `/api/session` caller, `_savedSessionSidebarOnlyState` (`boot.js:105`), swallows errors and returns `null`, so it won't mis-handle the new 409 — it just falls back as before. The cross-profile deep-link case is handled on both cold boot and in-app navigation.

## One thing to double-check: the retry re-entrancy guard

In `loadSession`, after `_switchProfileForSessionLoad` succeeds you do `if (_loadingSessionId === sid) _loadingSessionId = null;` then `return loadSession(sid,{...opts,skipProfileResolve:true,force:true})`. The `skipProfileResolve` flag correctly prevents an infinite switch loop if the retried load still 409s (it can't switch twice). Confirm that the profile switch itself invalidating/re-rendering the list (`renderSessionList()` inside `_switchProfileForSessionLoad`) doesn't race the re-entrant `loadSession` for the same `sid` — the stale-load guard (`_loadingSessionId !== sid`) should cover it, but a test that switches profile then immediately clicks a *third* session mid-switch would be worth adding to `test_bugfix_sweep.py` alongside the existing string-presence assertions.

The `test_bugfix_sweep.py` additions are static source assertions (checking the strings exist), which is consistent with the rest of that file, though they don't exercise the actual 409→switch→retry flow. A functional test on the handler (assert a cross-profile GET returns 409 with the owning profile name) would harden the contract against future refactors of the guard.

Net: correct layering (server distinguishes wrong-profile from missing, frontend recovers), preserves the `404` self-heal for genuinely deleted sessions, and keeps mutation endpoints on the strict guard. Looks good to merge.

nesquena-hermes:
Shipped in **v0.51.915** 🔀 — thanks @harcek!

Clicking a `session://` link that belongs to another profile now switches to that profile and loads the session, instead of failing with "Session not available in web UI" (it used to look identical to a deleted session). Truly missing/deleted sessions still self-heal.

**Gate found + I fixed 3 issues (Codex + Fable):**
- **Codex CORE:** added a post-await stale-load guard so a navigation *during* the profile switch can't hijack the UI back to the old session.
- **Fable Finding 1:** only emit the 409 when the owning profile is known — a truly-missing/legacy `None`-profile session under a *non-default* active profile would otherwise get a useless `profile:null` 409 that skipped the 404 self-heal and re-armed the SSE stream against a dead session id. Fixed both detail branches + 2 regression tests.
- **Fable Finding 2:** a failed profile-switch POST no longer strands the sidebar on the skeleton (now mirrors the #4671 canonical-switch recovery).

No foreign-profile transcript is ever returned — the 409 carries only `{error, code, session_id, profile}`. The info-disclosure delta (session-existence + owning-profile-name) is acceptable for the single-user/multi-profile model (profile names are already enumerable via `/api/profiles`).

Full gate green: Codex SAFE, Fable SAFE (after fixes), full suite clean. Authorship preserved via your commit + Co-authored-by.

PR #5407, opened 2026-07-02
fix(#5291): pin the OpenAI TTS proxy's connect target to close a DNS-rebinding TOCTOU

rodboev:
## Thinking Path

- #5079 closed the redirect variant of SSRF in the OpenAI-compatible TTS proxy (no-redirect opener) and validated the configured `base_url`'s resolved address at config-load time, but that validation and the actual outbound connect are two separate `getaddrinfo` calls with nothing tying one to the other.
- A hostname under attacker-controlled or rebinding DNS can answer public at validation time and private/link-local at connect time, sailing a real request (with the `Authorization` bearer attached) straight to an internal target that the validation step already rejected.
- The fix resolves the host once at the point the connection is actually made, re-validates that RRset, and pins the TCP dial to that vetted candidate list while leaving the Host header and TLS SNI on the original hostname so the request still looks like a normal request to that host from the upstream's perspective.
- This composes with (does not replace) the existing no-redirect opener: both handlers now sit on the same `build_opener(...)` call, and that opener now disables environment proxies so the direct dial still goes to the vetted TTS target instead of a proxy-controlled connect path.

## What Changed

- `api/routes.py`: hoisted the address-blocked check out of `_tts_host_is_blocked_target()`'s nested closure into a module-level `_tts_addr_is_blocked()` so it can be shared.
- `api/routes.py`: added `_tts_resolve_pinned_addresses()` plus the direct helper `_tts_resolve_pinned_address()`, so the connect path and unit helper share the same RRset validation and blocked-target checks.
- `api/routes.py`: added `_PinnedHTTPSConnection` (overrides `connect()` to iterate the vetted candidate list without re-resolving the original hostname, while keeping SNI on the original hostname) and `_PinnedHTTPSHandler`, both hoisted to module scope alongside `_NoRedirectTtsHandler`.
- `api/routes.py`: the OpenAI TTS opener now composes `ProxyHandler({})`, `_NoRedirectTtsHandler()`, and `_PinnedHTTPSHandler()` so redirects are still refused, the initial connect is pinned to the validated address, and environment proxies cannot bypass that direct-dial contract.
- `api/routes.py`: `_PinnedHTTPSConnection.connect()` now preserves the stdlib `http.client.connect` audit event before swapping only the dial target.
[...]

PR #5437, opened 2026-07-02
fix(#5430): harden ElevenLabs TTS transport guards

### LINKED ISSUE #5430: Harden the ElevenLabs TTS proxy against redirect/proxy SSRF (parity with #5407)
## Harden the ElevenLabs TTS proxy against redirect/proxy SSRF (parity with #5407)

**Context.** #5407 (v0.51.825) closed a DNS-rebinding TOCTOU / SSRF in the **OpenAI-compatible** TTS proxy: resolve-once → validate the whole RRset → dial only vetted literal addresses, with a no-redirect + no-proxy opener (`ProxyHandler({})` + `_NoRedirectTtsHandler`) so a redirect can't carry the `Authorization` bearer elsewhere and an ambient proxy can't bypass the pin.

**Gap.** The **ElevenLabs** TTS branch (`api/routes.py`, fixed host `api.elevenlabs.io`) still uses a plain redirect-following `urlopen` with its `xi-api-key` attached. This was surfaced by the Opus gate review on #5407 and is **out of #5291's scope** (that issue was OpenAI-specific), so it was deliberately deferred rather than folded in.

Because the ElevenLabs host is a fixed literal (not user-controlled), the DNS-rebind vector is much weaker than the OpenAI case — but for defense-in-depth the same treatment is worth applying:

**Proposed fix.**
- Route the ElevenLabs upstream fetch through a no-redirect opener so a redirect can't carry `xi-api-key` to another target.
- Disable ambient proxies on that dial (`ProxyHandler({})`), matching the OpenAI branch + the existing house precedent.
- (Optional, lower priority) apply the same resolve-once/validate-all connection pin if we ever allow a configurable ElevenLabs base URL.
- Add a regression test mirroring `test_issue4982_openai_tts.py`'s redirect/proxy cases.

**Severity:** low (fixed host, no user-controlled base URL today) — defense-in-depth hardening, not an active exploit. Good first-issue-adjacent for the TTS surface.

_Filed as the deferred follow-up from the #5407 / v0.51.825 security review._

rodboev:
## Thinking Path

- ElevenLabs TTS still builds a keyed upstream request with `xi-api-key`, then sends it through urllib's default opener.
- That default transport can follow redirects or honor ambient proxy settings, which can move a keyed request outside the intended ElevenLabs host path.
- The fix keeps the patch local to ElevenLabs, reuses the existing `_tts_open` seam, and adds focused regression coverage for redirect blocking, no-proxy opener construction, and the existing happy-path behavior.

## What Changed

- `api/routes.py`: routes ElevenLabs TTS through `_tts_open` with a no-redirect handler and `ProxyHandler({})` instead of inline `urlopen`, while preserving the existing request headers and response handling.
- `tests/test_issue3510_elevenlabs_tts.py`: updates ElevenLabs tests to patch the `_tts_open` seam and adds focused regression coverage for redirect blocking and no-proxy opener construction.

## Why It Matters

An upstream redirect or proxy configuration should not be able to carry `xi-api-key` outside the intended ElevenLabs endpoint. This change hardens the ElevenLabs transport path without changing ordinary audio generation behavior.

## Verification

- `pytest tests/test_issue3510_elevenlabs_tts.py -v --timeout=60`

## Upstream

Closes #5430.

Follow-up to the security review on https://github.com/nesquena/hermes-webui/pull/5407, which flagged the ElevenLabs branch separately from the OpenAI-specific fix.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -16670,13 +16670,18 @@ def _handle_tts(handler, parsed):
             "voice_settings": {"stability": 0.5, "similarity_boost": 0.75},
         }).encode("utf-8")
 
-        from urllib.request import Request, urlopen as _urlopen
-
+        from urllib.request import Request, build_opener
         req = Request(url, data=req_body, headers={
             "xi-api-key": api_key,
             "Content-Type": "application/json",
             "Accept": "audio/mpeg",
         })
+
+        class _NoRedirectElevenlabsHandler(HTTPRedirectHandler):
+            """Refuse to follow redirects on the ElevenLabs call."""
+
+            def redirect_request(self, req, fp, code, msg, headers, newurl):
+                raise ValueError("ElevenLabs TTS upstream attempted a redirect")
 
         # Buffer the full response before sending first byte.
         # The streaming endpoint is designed for chunked delivery, but urllib's
@@ -16684,7 +16689,7 @@ def _handle_tts(handler, parsed):
         # payloads. A hard cap keeps the buffered path bounded even if the
         # upstream misbehaves.
         try:
-            with _urlopen(req, timeout=30) as resp:
+            with _tts_open(req, timeout=30, opener_factory=lambda: build_opener(ProxyHandler({}), _NoRedirectElevenlabsHandler())) as resp:
                 audio_data = _buffer_tts_audio_response(resp)
         except ValueError:
             logger.warning("ElevenLabs TTS rejected an invalid upstream response", exc_info=True)
--- a/api/routes.py
+++ b/api/routes.py
@@ -16670,18 +16670,11 @@ def _handle_tts(handler, parsed):
             "voice_settings": {"stability": 0.5, "similarity_boost": 0.75},
         }).encode("utf-8")
 
-        from urllib.request import Request, build_opener
         req = Request(url, data=req_body, headers={
             "xi-api-key": api_key,
             "Content-Type": "application/json",
             "Accept": "audio/mpeg",
         })
-
-        class _NoRedirectElevenlabsHandler(HTTPRedirectHandler):
-            """Refuse to follow redirects on the ElevenLabs call."""
-
-            def redirect_request(self, req, fp, code, msg, headers, newurl):
-                raise ValueError("ElevenLabs TTS upstream attempted a redirect")
 
         # Buffer the full response before sending first byte.
         # The streaming endpoint is designed for chunked delivery, but urllib's
@@ -16689,7 +16682,7 @@ def _handle_tts(handler, parsed):
         # payloads. A hard cap keeps the buffered path bounded even if the
         # upstream misbehaves.
         try:
-            with _tts_open(req, timeout=30, opener_factory=lambda: build_opener(ProxyHandler({}), _NoRedirectElevenlabsHandler())) as resp:
+            with _tts_open(req, timeout=30, opener_factory=lambda: build_opener(ProxyHandler({}), _NoRedirectTtsHandler())) as resp:
                 audio_data = _buffer_tts_audio_response(resp)
         except ValueError:
             logger.warning("ElevenLabs TTS rejected an invalid upstream response", exc_info=True)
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Reading the diff at `api/routes.py:16673-16692` against the OpenAI branch on master (`routes.py:16766-16768`) and the shared seam `_tts_open` (`routes.py:16507-16520`), this closes the ElevenLabs gap #5407 left open, and the parity choice is defensible. One note on where it intentionally diverges from the OpenAI branch.

## The fix matches the existing seam

Before, ElevenLabs called `urllib.request.urlopen` directly, so it inherited the default opener — redirect-following plus ambient proxy env. The patch routes it through the same `_tts_open` seam the OpenAI branch already uses, with a no-redirect handler and an empty `ProxyHandler`:

```python
with _tts_open(req, timeout=30, opener_factory=lambda: build_opener(ProxyHandler({}), _NoRedirectElevenlabsHandler())) as resp:
    audio_data = _buffer_tts_audio_response(resp)
```

`ProxyHandler({})` (empty dict) disables proxy lookup for this request, so a keyed `xi-api-key` request can't be bounced through an attacker-controlled `HTTP(S)_PROXY`. The `_NoRedirectElevenlabsHandler.redirect_request` raising `ValueError` blocks a 3xx from carrying the key to a new host, and the existing `except ValueError:` at `routes.py:16690` maps that to a clean 502 rather than leaking a stack trace. Correct, minimal, and behavior-preserving for the happy path.

## Why it correctly omits `_PinnedHTTPSHandler`

The OpenAI branch adds `_PinnedHTTPSHandler` to its opener chain (`routes.py:16768`) for DNS-rebinding defense. It's reasonable to ask why ElevenLabs doesn't. The answer is in the URL construction: the OpenAI host is config-overridable —

```python
base_url = _normalized_openai_tts_base_url(oai_cfg.get("base_url") or base_url)
```

— so it needs pinning + a base-url allowlist because the destination host is user-influenced. ElevenLabs, by contrast, hardcodes the host at `routes.py:16666`:

```python
url = f"https://api.elevenlabs.io/v1/text-to-speech/{voice_id}/stream?output_format=mp3_44100_128"
```

Only `voice_id` is config-derived, and it's already constrained to `[A-Za-z0-9_-]+` via `re.fullmatch` at `routes.py:16662` before it reaches the path segment. With a fixed host and no proxy/redirect escape, the residual DNS-rebind surface is much smaller than the OpenAI case, so skipping the pinned handler is a proportionate choice — worth stating explicitly in the PR body so a future reader doesn't "fix" the asymmetry by mistake.

## Tests exercise the seam, not the network

The updated `tests/test_issue3510_elevenlabs_tts.py` monkeypatches `routes._tts_open` (not `urllib`), so no live ElevenLabs call happens. `test_elevenlabs_tts_does_not_follow_upstream_redirect` reaches into `opener_factory()` and asserts the redirect handler raises `ValueError` on a `169.254.169.254` metadata redirect — a good SSRF regression. `test_elevenlabs_tts_uses_no_proxy_opener` spies `ProxyHandler` and asserts `proxies == {}`, locking in the no-proxy construction. `test_elevenlabs_overlong_text_rejected_before_engine` now installs a `_fail_if_called` stub, proving the 5000-char cap still short-circuits before any upstream call.

One small suggestion: the two new tests both re-declare a local `_Resp` and `_fake_urlopen`; a shared fixture would cut ~40 lines of duplication, but that's cosmetic. Functionally this is solid and I'd merge it. CI is green across all 15 test shards.

nesquena-hermes:
Shipped in **v0.51.833** ✅ — thanks @rodboev!

Autonomous fix-class ship (security-hardening of a shipped feature). Full gate: Codex security gate **SAFE TO SHIP** (verified redirect-exfil blocked, ambient-proxy suppressed — both probed directly; OpenAI branch unchanged; removed inline imports safe), full suite 11826 passed (2 known non-defects), PR-own tests 7/7, ast+ruff CLEAN.

The ElevenLabs TTS proxy now uses the same no-redirect / no-proxy opener as the OpenAI one (#5407), so a redirect can't carry the `xi-api-key` to another host and ambient proxies can't intercept the dial. This closes the follow-up filed as #5430.

Credit preserved via `Co-authored-by`. Fixes #5430.

PR #5456, opened 2026-07-03
fix: reduce transparent stream live row churn

Stacey2911:
## Thinking Path

Transparent Stream now preserves live rows by identity, but follow-up testing found remaining no-op churn and layout instability during streaming updates:

1. preserved rows could still refresh child markup even when the incoming row HTML was unchanged;
2. preserved rows could still be reinserted even when already in the correct sibling position;
3. long thinking blocks could still show internal scrollbar flicker once their scroll container appeared.

Those cases preserve the top-level row node, but still create unnecessary DOM/layout instability during live rendering.

## What Changed

- Carries `_tcData` forward when a transparent live row is reconciled.
- Keeps copy/header handler rebinds tied to real child markup refreshes.
- Skips `innerHTML` replacement when the candidate row markup is unchanged.
- Skips `insertBefore`/`appendChild` when a reconciled row is already under the correct parent and in the correct sibling position.
- Stabilizes the long thinking-block scroll container/scrollbar behavior.
- Preserves new-row insertion, moved-row repair, live footer ordering, expanded/detail state, and existing reconciliation behavior.

## Why It Matters

This reduces Transparent Stream no-op DOM/layout churn during live responses without changing Compact Worklog behavior or replacing the existing row reconciliation model.

In local browser validation, this removed visible chat-window flicker caused by stable transparent rows being refreshed or reinserted during token generation, and also removed the remaining internal scrollbar flicker inside long thinking blocks.

## Contract Routing

Touches Transparent Stream live-row reconciliation and visual/DOM stability. It does not change persisted session shape, message payloads, provider API behavior, backend contracts, or activity mode selection.

## Verification

Local:

```bash
node --check static/ui.js
./scripts/test.sh tests/test_issue5367_transparent_live_row_reconcile.py tests/test_issue3820_chat_activity_display_mode.py tests/test_live_to_final_anchor_visible_order.py tests/test_stable_assistant_turn_anchor_registry.py
git diff --check
```

Result: 122 tests passed locally after the thinking-block scrollbar fix was added.

Browser/manual:

* Activity Display: Transparent Stream
* Long streaming response with thinking/tool/activity rows
* Confirmed stable rows no longer flash/reinsert on every token
* Confirmed long thinking-block internal scrollbar no longer flickers/jitters
* Confirmed row ordering and footer placement remain correct
* Confirmed expand/collapse remains stable

## Risks / Follow-ups

* This intentionally does not include motion/fade suppression changes, because current upstream work is already tracking transparent live-row animation suppression separately.
* Copy-control parity is kept for a follow-up PR to keep this change focused.
* Remaining Transparent Stream copy-control work is not included here.

## Model Used

Planning/review: OpenAI GPT-5.5 Thinking.

Implementation/review support: Codex CLI and HermesAgent running qwen3.6-27b@q4_k_m.

Manual browser validation was performed by the user on the local HermesWebUI runtime.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -4223,6 +4223,7 @@ body.resizing .sidebar{transition:none!important;}
 .transparent-event-row .thinking-card.open .thinking-card-body{
   border-top-color:transparent;
   padding:0 0 3px;
+  scrollbar-gutter:stable;
 }
 .transparent-event-row .tool-card-detail > *,
 .transparent-event-row .thinking-card-body > *{min-height:0;}
--- a/static/ui.js
+++ b/static/ui.js
@@ -10964,20 +10964,10 @@ function _renderLiveAnchorActivitySceneTransparent(streamId, scene, opts){
     targetRenderedRows.push(renderedNode);
     wrote=true;
   }
-  const transparentLiveRowNextSibling=(node)=>{
-    if(!node) return null;
-    if(typeof node.nextSibling!=='undefined') return node.nextSibling;
-    const parent=node.parentElement||node.parentNode;
-    if(parent&&parent.children&&typeof parent.children.length==='number'){
-      const index=Array.prototype.indexOf.call(parent.children,node);
-      return index>=0 ? (parent.children[index+1]||null) : null;
-    }
-    return null;
-  };
   const transparentLiveRowAlreadyPositioned=(node, expectedNextSibling)=>!!(
     node &&
     node.parentElement===blocks &&
-    transparentLiveRowNextSibling(node)===expectedNextSibling
+    node.nextSibling===expectedNextSibling
   );
   let expectedNextSibling=(liveFooter&&liveFooter.parentElement===blocks) ? liveFooter : null;
   for(let i=targetRenderedRows.length-1;i>=0;i--){
--- a/static/ui.js
+++ b/static/ui.js
@@ -10945,6 +10945,7 @@ function _renderLiveAnchorActivitySceneTransparent(streamId, scene, opts){
   });
   const liveFooter=blocks.querySelector('#liveRunStatus');
   let wrote=false;
+  const targetRenderedRows=[];
   for(const row of rows){
     const node=_anchorSceneTransparentNodeForRow(row,{
       live:true,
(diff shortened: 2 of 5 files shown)


Stacey2911:
Addressed the Greptile note about the unreachable `transparentLiveRowNextSibling` fallback in a small follow-up commit. Local PR-A validation still passes (122 tests, syntax checks, diff-check).

Stacey2911:
I checked overlap with #5506 since it also touches Transparent Stream live-row behavior.

The overlap appears limited to the general stable row-ordering/reinsert area in `_renderLiveAnchorActivitySceneTransparent()`. This PR intentionally avoids the prose fade / motion surface handled by #5506 and keeps the local motion/fade commits out of scope.

The pieces here that still appear distinct are:

1. skipping unchanged `innerHTML` refreshes for preserved non-fade rows;
2. footer-aware stable row positioning, including explicit footer-boundary repair coverage;
3. long thinking-block scrollbar stability.

If #5506 lands first, I can rebase this branch and keep only the remaining distinct DOM-stability pieces.

nesquena-hermes:
Shipped in **v0.51.854** 🎉 — merged via release `stage-5456` (commit on master).

Thanks @Stacey2911 — a clean follow-up to the identity-preserving live-row reconcile. The streaming view is smoother now: no redundant innerHTML rewrites when a row is unchanged, no reinsert when a row is already positioned, and no thinking-block scrollbar flicker.

Cleared the full gate:
- **Codex: SAFE TO SHIP** (no findings)
- **Independent deep review: SAFE TO SHIP** — verified the skip gates never show stale content (attrs/className/`_tcData`/interactive-state all still sync unconditionally; only the redundant DOM write is gated) and the position-skip reproduces the exact terminal order without orphaning
- **UX review: SHIP** — the reconcile skips are a pure no-op-work reduction (terminal DOM verified identical; the visible difference is the *absence* of flicker); the `scrollbar-gutter:stable` line matches four in-repo precedents
- **Full suite green** (11970 passed)

Note: the branch predated master's #4295 scroll-anchor fix so the raw head showed 2 unrelated reds — they cleared on merge-forward (114 transparent/anchor tests green on the merged tree). Closing as shipped — thanks for the contribution!

PR #5454, opened 2026-07-03
fix(#5367): suppress transparent stream live row animation

rodboev:
## Thinking Path

- Issue #5367 identified two viable fixes for Transparent Stream flicker: reconcile live rows by identity, or suppress the live-row entrance animation when rebuilds still happen.
- #5400 shipped the durable reconcile path in v0.51.827, but b3nw reported on Discord: `did this fix go out? not improved at all in WebUI: v0.51.833`.
- v0.51.833 contains the reconcile code, so the remaining visible flicker is plausibly from live rows that still rebuild outside the stable-key path and replay `transparent-event-enter`.
- This change applies the first #5367 direction directly: the live turn still renders Transparent Stream rows, but those rows no longer run an entrance animation during streaming updates.

## What Changed

- `static/style.css`: remove `#liveAssistantTurn .transparent-event-row`, which is scoped to the live turn only.
- `static/style.css`: remove the now-unused `@keyframes transparent-event-enter` block.
- `tests/test_issue3820_chat_activity_display_mode.py`: update the CSS invariant test so it pins that the rule was removed.

## Why It Matters

Users still seeing Transparent Stream flicker after #5400 get the animation-side guard from the original #5367 maintainer direction. Existing `data-transparent-fade` opacity rules stay intact because this avoids forcing `opacity:1`.

## Verification

- `pytest tests/test_issue3820_chat_activity_display_mode.py -v --timeout=60`
- `pytest tests/test_smooth_text_fade.py -v --timeout=60`
- `npx eslint --no-config-lookup -c eslint.runtime-guard.config.mjs "static/**/*.js"`

## Upstream

Refs #5367.

Follow-up to #5400, which shipped the reconcile path but left the live-row entrance animation available for any remaining rebuild path.

## Model Used

GPT 5 via Codex CLI

PR #3603, opened 2026-06-04
fix(sidebar): use is_cli_session_row() instead of hardcoding True for all state.db sessions (#3586)

rodboev:
Closes #3586

## Thinking Path

- Discord, Telegram, and Slack sessions imported from state.db lose their source label after a sidebar refresh because `_load_cli_sessions_uncached()` hardcodes `is_cli_session: True` for every row.
- The classification function `is_cli_session_row()` in `api/agent_sessions.py` already inspects `session_source`, `source_tag`, `raw_source`, `source`, and `source_label` to correctly distinguish CLI from messaging and cron sessions.
- The rows from `read_importable_agent_session_rows` already carry all the fields `is_cli_session_row` needs; the function just was never called.
- Two sites need fixing: the main pass (line 3552) and the cron second-pass (line 3636).

## What Changed

- `api/models.py`: added `is_cli_session_row` to the import block from `api.agent_sessions`. Replaced `'is_cli_session': True` with `'is_cli_session': is_cli_session_row(row)` at both the main-pass (line 3552) and cron second-pass (line 3636) dict literals.
- `tests/test_issue3586_cli_session_source_label.py`: new test creating a temporary state.db with discord, telegram, cron, and cli rows, calling `_load_cli_sessions_uncached`, and asserting `is_cli_session` is `False` for messaging/cron sources and `True` for cli.

## Why It Matters

Messaging-platform sessions (Discord, Telegram, Slack) appear under the wrong sidebar tab after a refresh, making them unfindable unless the user switches to the CLI filter. Cron sessions are also misclassified.

## Verification

```
pytest tests/test_issue3586_cli_session_source_label.py -v --timeout=60
pytest tests/ -v --timeout=60
```

Manual: import a Discord or Telegram session via state.db, refresh the sidebar, verify it does not appear under the CLI tab.

## Risks / Follow-ups

- The Claude Code session builder at line 3359 also hardcodes `'is_cli_session': True`. That path is for external agent sessions (not state.db rows), so `is_cli_session_row` does not apply; left unchanged.
- If new source types are added to hermes without updating `is_cli_session_row`, they will fall through to the legacy heuristic at the bottom of that function. The heuristic is conservative (requires `is_cli_session` flag plus non-messaging source plus default CLI title pattern), so unknown sources will get `False` by default, which is the safer direction.

## Model Used

Claude Opus 4.8 via Claude Code CLI

Issue #1700, opened 2026-05-05
feat(profiles): allow switching profile while another profile's agent is running (parallel multi-profile runs)


nesquena-hermes:
## Summary

When an agent is running in any profile, the WebUI blocks profile switching with a yellow toast: **"Cannot switch profiles while agent is running"**. Power users who configure multiple profiles for parallel work — different agent specialisations, different model providers, different tool surfaces — want to bounce between them while one is mid-stream, the same way they switch between chat sessions today.

Reported in the WebUI Discord testers thread by @Lemz (May 5 2026):

> as a serial multi tasker, was wondering if there's any way around this or is it just a technical limitation with multi profiles? I have a few agents for their specialisations and I like to bounce around as I'm doing stuff in parallel and no idea if this is too big an ask or not but just wanted to throw it out there.

## Current behaviour

`api/profiles.py:585-591` — `switch_profile()` raises if any stream is active anywhere:

```python
# Block if agent is running
with STREAMS_LOCK:
    if len(STREAMS) > 0:
        raise RuntimeError(
            'Cannot switch profiles while an agent is running. '
            'Cancel or wait for it to finish.'
        )
```

`STREAMS` is a process-global dict keyed by stream_id, not scoped per profile. Any active stream from any profile blocks every other profile from switching.

## Is this technically feasible?

**Yes — and there's already a documented path.** The constraint is process-wide, not fundamental.

### What's already in place (#798, April 2026)

[#798](https://github.com/nesquena/hermes-webui/issues/798) introduced the **per-client profile context**: a `hermes_profile` cookie + `threading.local()` resolver so two browsers logged into the same WebUI instance can each have their own active profile without stepping on each other. `get_active_profile_name()` reads from TLS first, then process-global. So the read path is already TLS-aware.

### What's missing (#1574, May 2026)

The hermes-agent code that actually executes a run still depends on **module-level globals** mutated by `_set_hermes_home()`:
- `HERMES_HOME`, `HERMES_DIR`, `CRON_DIR`, `JOBS_FILE`, `OUTPUT_DIR` (in `hermes-agent/cron/jobs.py`)
- The dotenv reload in `_reload_dotenv(home)` mutates `os.environ` process-wide
- `reload_config()` reloads `cfg` from the new profile's `config.yaml`

Because these are process-global, allowing a profile switch while another profile's stream is running would clobber the running agent's environment mid-run.

[...]

PR #5460, opened 2026-07-03
fix(webui): replace cross-profile empty sessions on profile switch

ruizanthony:
## Summary

- Replace a current empty session during profile switch when its recorded profile does not match the newly active profile.
- Prevent the browser from keeping an old-profile `S.session.session_id` after `/api/profile/switch` succeeds.
- Add regression coverage for the upload failure class where attachments post `S.session.session_id` and the backend correctly rejects sessions outside the active profile.

## Why

The profile switch path previously reused/retagged an empty current session in place. If that session belonged to the previous profile, later attachment uploads used the stale `session_id`; `api/upload.py` then returned `404 Session not found` under the new profile cookie. Users had to hard reload before uploading.

## Tests

- `python -m pytest tests/test_profile_switch_ux.py::TestParallelizedFetches::test_cross_profile_empty_session_is_replaced_before_mutation_or_upload tests/test_chat_upload_attachment_paths.py -q -o 'addopts='`

### DIFF
--- a/static/panels.js
+++ b/static/panels.js
@@ -6437,6 +6437,7 @@ async function switchToProfile(name) {
   const _titlebarLabel = $('titlebarProfileLabel');
   const _prevProfileName = S.activeProfile || 'default';
   const _switchGen = ++_profileSwitchGeneration;
+  const _openingExistingSidebarSession = !!(typeof _profileSwitchOpeningExistingSession !== 'undefined' && _profileSwitchOpeningExistingSession);
   if (_chip) { _chip.classList.add('switching'); _chip.disabled = true; }
   if (_titlebarBtn) { _titlebarBtn.classList.add('switching'); _titlebarBtn.disabled = true; }
   // Optimistic name update — shows the target name right away
@@ -6459,7 +6460,7 @@ async function switchToProfile(name) {
   if (typeof closeSessionActionMenu === 'function') closeSessionActionMenu();
   // Determine whether the current session must be replaced instead of being
   // retagged in place. A session with messages/active runtime belongs to the
-  // current profile. After /api/profile/switch returns, we also treat an
+  // current profile. After the profile-switch POST returns, we also treat an
   // otherwise-empty session whose recorded profile does not match the target
   // profile as replace-only: uploads send S.session.session_id and the backend
   // correctly rejects old-profile sessions under the new profile cookie.
@@ -6468,6 +6469,11 @@ async function switchToProfile(name) {
     S.session.active_stream_id ||
     S.session.pending_user_message
   ));
+  if (_openingExistingSidebarSession && S.session) {
+    // A cross-profile sidebar click is about to load a concrete existing session.
+    // Do not create or retag a blank intermediary session in the destination profile.
+    sessionInProgress = true;
+  }
   const _workspaceVisibleAtStart = typeof _workspacePanelMode !== 'undefined' && _workspacePanelMode !== 'closed';
 
   // #4671 CORE: the skeleton/embargo/generation setup is INSIDE the try so the
@@ -6608,10 +6614,20 @@ async function switchToProfile(name) {
     }
 
     // ── Session ────────────────────────────────────────────────────────────
-    _showAllProfiles = false;
+    // Keep the all-profiles sidebar scope sticky across profile switches. It is
+    // a navigation preference shared by the browser session, not a per-profile flag.
     if (typeof animateNextSessionListRefresh === 'function') animateNextSessionListRefresh();
 
-    if (sessionInProgress) {
+    if (sessionInProgress && _openingExistingSidebarSession) {
+      // The caller will immediately load the clicked session after this profile
+      // cookie switch. Avoid creating/retagging an intermediate blank chat.
+      const workspaceVisible = typeof _workspacePanelMode !== 'undefined' && _workspacePanelMode !== 'closed';
+      if (typeof _setProfileSwitchListEmbargo === 'function') _setProfileSwitchListEmbargo(false);
+      await renderSessionList();
+      if (_switchGen !== _profileSwitchGeneration) return;
+      if (workspaceVisible && typeof clearWorkspaceTreeSkeleton === 'function') clearWorkspaceTreeSkeleton();
+      showToast(t('profile_switched', name));
+    } else if (sessionInProgress) {
       // The current session has messages and belongs to the previous profile.
       // Start a new session for the new profile so nothing gets cross-tagged.
       const workspaceVisible = typeof _workspacePanelMode !== 'undefined' && _workspacePanelMode !== 'closed';
--- a/static/panels.js
+++ b/static/panels.js
@@ -6383,6 +6383,21 @@ window.addEventListener('resize',()=>{
   if(dd&&dd.classList.contains('open')) _positionProfileDropdown();
 });
 
+function _openProfileSwitchSessionBrowser(){
+  try{
+    const isDesktop = (typeof _isDesktopWidth === 'function') ? _isDesktopWidth() : true;
+    if(isDesktop){
+      if(typeof expandSidebar === 'function') expandSidebar();
+      return;
+    }
+    const sidebar=document.querySelector('.sidebar');
+    if(!sidebar)return;
+    try{if(typeof _syncMobileSidebarPanelFromMainView==='function')_syncMobileSidebarPanelFromMainView();}catch(_){}
+    sidebar.classList.remove('mobile-session-page');
+    sidebar.classList.add('mobile-panel-drawer','mobile-open');
+  }catch(_){}
+}
+
 async function switchToProfile(name) {
   // ── #4671 profile-switch loading-skeleton — FOUR-GUARD CONTRACT ───────────────
   // The skeleton must never be clobbered by the OLD profile's content and must never
@@ -6442,14 +6457,17 @@ async function switchToProfile(name) {
   // context change where dismissing those transient affordances is correct.
   if (typeof _renamingSid !== 'undefined' && _renamingSid) _renamingSid = null;
   if (typeof closeSessionActionMenu === 'function') closeSessionActionMenu();
-  // Determine whether the current session has any messages.
-  // A session with messages is "in progress" and belongs to the current profile —
-  // we must not retag it.  We'll start a fresh session for the new profile instead.
-  const sessionInProgress = S.session && (
+  // Determine whether the current session must be replaced instead of being
+  // retagged in place. A session with messages/active runtime belongs to the
+  // current profile. After /api/profile/switch returns, we also treat an
+  // otherwise-empty session whose recorded profile does not match the target
+  // profile as replace-only: uploads send S.session.session_id and the backend
+  // correctly rejects old-profile sessions under the new profile cookie.
+  let sessionInProgress = !!(S.session && (
     (S.messages && S.messages.length > 0) ||
     S.session.active_stream_id ||
     S.session.pending_user_message
-  );
+  ));
   const _workspaceVisibleAtStart = typeof _workspacePanelMode !== 'undefined' && _workspacePanelMode !== 'closed';
 
   // #4671 CORE: the skeleton/embargo/generation setup is INSIDE the try so the
@@ -6482,6 +6500,19 @@ async function switchToProfile(name) {
     if (_switchGen !== _profileSwitchGeneration) return;
     S.activeProfile = data.active || name;
     S.activeProfileIsDefault = !!data.is_default;
+    const targetActiveProfile = S.activeProfile || 'default';
+    let sessionProfileMatchesTarget = true;
+    if (!sessionInProgress && S.session) {
+      const currentSessionProfile = (typeof S.session.profile === 'string' && S.session.profile.trim())
+        ? S.session.profile.trim()
+        : 'default';
+      sessionProfileMatchesTarget = (typeof _profileMatchesActiveProfile === 'function')
+        ? _profileMatchesActiveProfile(currentSessionProfile, targetActiveProfile)
+        : (currentSessionProfile === targetActiveProfile || (currentSessionProfile === 'default' && !!S.activeProfileIsDefault));
+      if (!sessionProfileMatchesTarget) {
+        sessionInProgress = true;
+      }
+    }
     // #4650 review: a profile switch can change agent.reasoning_effort (and other
     // reasoning inputs like base_url) WITHOUT changing the default model/provider,
     // which is all the reasoning-chip cache key tracks. Force exactly one reasoning
@@ -6600,6 +6631,7 @@ async function switchToProfile(name) {
       // and pop a stale toast. Mirrors the no-messages branch guard below.
       // (@rodboev/greptile review, #4662)
       if (_switchGen !== _profileSwitchGeneration) return;
+      if (typeof _openProfileSwitchSessionBrowser === 'function') _openProfileSwitchSessionBrowser();
       // Safety net: if the new session has no workspace, newSession() won't have
       // painted the file tree — clear the up-front skeleton so it can't strand
       // (#4662 Opus gate). No-op when a real tree already rendered.
@@ -6622,6 +6654,7 @@ async function switchToProfile(name) {
       if (typeof _setProfileSwitchListEmbargo === 'function') _setProfileSwitchListEmbargo(false);
       await renderSessionList();
       if (_switchGen !== _profileSwitchGeneration) return;
+      if (typeof _openProfileSwitchSessionBrowser === 'function') _openProfileSwitchSessionBrowser();
       syncTopbar();
       // Refresh workspace file tree so the right panel shows the new
       // profile's workspace, not the previous one (#1214).
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -1854,6 +1854,37 @@ function _externalImportPayload(session) {
   return payload;
 }
 
+function _sidebarSessionProfileName(session){
+  const raw=session&&typeof session.profile==='string'?session.profile.trim():'';
+  return raw||'';
+}
+
+async function _ensureSidebarSessionProfile(session){
+  const targetProfile=_sidebarSessionProfileName(session);
+  if(!_showAllProfiles||!targetProfile) return false;
+  const activeProfile=S.activeProfile||'default';
+  if(_profileMatchesActiveProfile(targetProfile,activeProfile)) return false;
+  if(typeof switchToProfile!=='function') return false;
+  _profileSwitchOpeningExistingSession=true;
+  try{
+    await switchToProfile(targetProfile);
+  }finally{
+    _profileSwitchOpeningExistingSession=false;
+  }
+  return _profileMatchesActiveProfile(targetProfile,S.activeProfile||'default');
+}
+
+async function _openSidebarSession(session, loadOpts={}){
+  if(!session||!session.session_id) return;
+  if(_isExternalSession(session)){
+    try{await api('/api/session/import_cli',{method:'POST',body:JSON.stringify(_externalImportPayload(session))});}
+    catch(_e){ /* import failed -- fall through to read-only view */ }
+  }
+  await _ensureSidebarSessionProfile(session);
+  await loadSession(session.session_id, loadOpts);
+  renderSessionListFromCache();
+}
+
 function _isReadOnlySession(session) {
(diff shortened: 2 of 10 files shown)


nesquena-hermes:
Read `switchToProfile()` in `static/panels.js` at the PR head (the new block around 6485-6500) against `origin/master`, plus the backend rejection path in `api/upload.py` and `api/profiles.py`. The diagnosis is correct and the fix targets the right seam.

## Summary

The stale-session-after-switch upload failure is real. `api/upload.py:154-165` gates every upload target on the active profile:

```python
def _session_visible_to_active_profile(session) -> bool:
    """Return whether an upload target session belongs to the active profile."""
    session_profile = getattr(session, 'profile', None)
    ...
    return _profiles_match(session_profile, _get_active_profile_name())
```

and returns `404 Session not found` when it fails. Before this PR, an *empty* current session was retagged in place on switch, but if that session belonged to the previous profile the browser kept the old `S.session.session_id`; a subsequent attachment upload posted that id and tripped the 404. The PR promotes a profile-mismatched empty session onto the replace path so the browser drops the stale id and `newSession()` mints a fresh, correctly-tagged one.

## Code reference

The new gate in `panels.js` (after `/api/profile/switch` returns, so `S.activeProfile` reflects the target):

```javascript
if (!sessionInProgress && S.session) {
  const currentSessionProfile = (typeof S.session.profile === 'string' && S.session.profile.trim())
    ? S.session.profile.trim() : 'default';
  sessionProfileMatchesTarget = (typeof _profileMatchesActiveProfile === 'function')
    ? _profileMatchesActiveProfile(currentSessionProfile, targetActiveProfile)
    : (currentSessionProfile === targetActiveProfile || ...);
  if (!sessionProfileMatchesTarget) sessionInProgress = true;
}
```

Two things I checked that make this correct rather than fragile:

1. **Placement.** This runs after `S.activeProfile = data.active || name` is applied, so `targetActiveProfile` is the *new* profile, and it runs before both the in-place `S.session.profile` retag (`panels.js`, guarded by `if (S.session && !sessionInProgress)`) and the `/api/session/update` call in the workspace block. So flipping `sessionInProgress` to `true` here correctly steers past the in-place retag into the `if (sessionInProgress)` branch that calls `newSession()`. The test asserts exactly this ordering (`promote_idx < first_in_place_patch < first_update < branch_idx`), which is a good guard against a future reorder silently breaking it.

2. **Match semantics parity.** The client reuses `_profileMatchesActiveProfile` (sessions.js:38), whose renamed-root handling (`eventName === 'default' && S.activeProfileIsDefault`) mirrors the backend `_profiles_match` -> `_is_root_profile` cross-alias in `api/profiles.py:373`. So a legacy `'default'`-tagged session under a renamed root profile is treated as matching on both sides, and you won't spuriously discard a session that the backend would actually accept.

## One thing worth confirming

`_profileMatchesActiveProfile` leans on `S.activeProfileIsDefault` for the renamed-root case. The new block sets `S.activeProfileIsDefault = !!data.is_default` two lines above, so it's fresh at call time — good. The only residual gap is symmetry: the client match treats `'default'` as matching the active root only when the *session* side is `'default'`; the backend `_profiles_match` cross-aliases in both directions via `_is_root_profile(row) and _is_root_profile(active)`. In practice `S.session.profile` for a freshly created root-profile session is whatever the switch wrote, so this is unlikely to bite, but if you ever see a root session tagged with the renamed display name (not `'default'`) get needlessly replaced on switch-to-root, that asymmetry is the place to look.

## Verification

The added `test_cross_profile_empty_session_is_replaced_before_mutation_or_upload` is a source-shape assertion (it greps the function text for ordering), not a behavioral DOM test — reasonable given the existing `TestParallelizedFetches` style in this file, but it would pass even if `_profileMatchesActiveProfile` returned the wrong boolean. A follow-up worth considering: a small jsdom/unit check that an empty session tagged with profile A, after switching to profile B, results in `newSession()` being invoked (behavior), not just that the tokens appear in order. Net: the fix is sound and lands on the correct layer.

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (behavior is sound; 3 brittle STATIC tests misalign with the improved code — CI red)

**Certified head:** `sha:51a5a7b9` (rebased onto current master, `git apply` clean) · **PR:** #5460 · _ruizanthony, fix(webui): replace cross-profile empty sessions on profile switch_
**Verdict:** The switch fix itself is correct (Codex SAFE, 0 findings; its own `test_profile_switch_ux.py` + 11827 others pass). But CI is red because 3 brittle **static source-string** tests slice fragile windows/literals around `switchToProfile()` that no longer align after the PR's (legitimate) code changes. All 3 are test-brittleness, not behavior loss — but they must be updated for CI to go green.

### What I ran (rebased worktree `/tmp/wt-rebase-5460`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SAFE TO SHIP** — 0 findings; switch logic sound, failures are brittle-static-test misalignment |
| Full pytest suite | 5 failed / 11827 passed — 2 known env flakes (nous, issue4536) + the 3 brittle static tests below |
| PR's own test | ✅ `test_profile_switch_ux.py` passes |

### Findings — 3 brittle static tests, behavior VERIFIED intact
**⛔ CI-red (3 brittle static tests; I confirmed each is a string-match artifact, not behavior loss):**
1. `test_issue4662...::test_switch_post_suppresses_generic_timeout_toast` — slices a ±200-char window around `body.index("/api/profile/switch")`. The PR inserted a comment block containing `/api/profile/switch` BEFORE the actual `api(..., timeoutToast: false)` call, so `index` finds the comment and the window misses the call. **`timeoutToast: false` IS still on the switch call (panels.js:6484)** — behavior intact. **Fix:** anchor on the `api('/api/profile/switch'` call (e.g. scope to `api(` + the path, or `rindex`), not a naive first-`index` window.
2. `test_issue4662...::test_shows_session_skeleton_up_front` — asserts the literal `"showSessionListSkeleton()"`. The PR changed the call to `showSessionListSkeleton(name)` (parameterized) + guarded with `typeof … === 'function'` (panels.js:6471) — skeleton still shown. **Fix:** assert `showSessionListSkeleton(` (allow an argument).
3. `test_issue1700...::test_frontend_treats_active_or_pending_session_as_in_progress` — slices `fn.find("const sessionInProgress") : fn.find("try {", ...)`. The PR changed `const sessionInProgress` → **`let sessionInProgress`** (because it legitimately reassigns it at 6498: `sessionInProgress = true`). So `find("const …")` returns −1 → empty slice → assert fails. **The `S.session.active_stream_id` in-progress guard IS fully intact (panels.js:6451-6454)** — behavior sound. **Fix:** match `sessionInProgress` without the `const` prefix (accept `let`/`const`).

**✅ Behavior sound:** the switch fix (replace-only for a cross-profile empty session vs same-profile retag) is correct; timeoutToast:false, skeleton, and the active_stream_id/pending in-progress guard are all still wired. Codex SAFE (0 findings). Positive nesquena-hermes read ("diagnosis correct, targets the right seam").

**RED — gate-fail/changes-requested (CI-red, test-only fixes): update the 3 brittle static assertions to match the improved code — anchor timeoutToast on the `api('/api/profile/switch'` call, accept `showSessionListSkeleton(` with an arg, and match `sessionInProgress` regardless of `let`/`const`.** The PR's behavior is verified correct (Codex SAFE + its own test + the guards all intact) — this is purely static-test brittleness that the legitimate refactor exposed. Fast to fix. Once green, it's a crown-jewel profile-switching reliability fix (stale-session upload-404). concept 4/5. Author @ruizanthony (T1). crit=3. (Gate lesson: static source-string tests that slice fixed windows / match exact call-literals / assume `const` are brittle — a legitimate refactor breaks them without any behavior change; when a static test fails, verify the actual behavior is intact before treating it as a defect, and fix the test to be structure-tolerant.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; all 3 CI failures confirmed brittle-static-test misalignment (timeoutToast:false@6484, showSessionListSkeleton(name)@6471, active_stream_id guard@6451-54 all intact), Codex SAFE 0 findings. Cert valid for sha:51a5a7b9._

ruizanthony:
Fixed in `be92ca13`.

Changes made:
- aligned the fallback profile comparison with `_profileMatchesActiveProfile` by preserving renamed-default/root alias semantics via `S.activeProfileIsDefault`;
- simplified `targetActiveProfile` after `S.activeProfile` is assigned;
- updated the three brittle static assertions called out by the gate:
  - anchor `timeoutToast: false` on the actual `api('/api/profile/switch'...)` call;
  - allow `showSessionListSkeleton(...)` with an argument;
  - match `sessionInProgress` regardless of `let`/`const`.

Local validation:
- `python -m pytest tests/test_profile_switch_ux.py tests/test_issue4662_profile_switch_skeleton_static.py tests/test_issue1700_parallel_profile_switch.py tests/test_chat_upload_attachment_paths.py -q -o 'addopts='` → 53 passed

Remote validation on `be92ca13`: browser-smoke, lint, all test shards, and Greptile Review are green.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (brittle tests fixed + profile-match logic refined)

**Certified head:** `sha:6a93b00d` (clean rebase, branch `gate-rebase/5460-cross-profile-empty-session`) · **PR:** #5460 · _ruizanthony, fix(webui): replace cross-profile empty sessions on profile switch_
**Verdict:** Bounce closed. The 3 brittle static tests from my prior gate are updated (31/31 pass), and the re-push refined the profile-match logic (`targetActiveProfile = S.activeProfile || 'default'`; match via `S.activeProfileIsDefault`). Codex SAFE, suite green, all switch guards intact. Crown-jewel profile-switching reliability fix (stale-session upload-404).

### What I ran (rebased worktree `/tmp/wt-rebase-5460b`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; three-dot diff = `panels.js` + 3 tests ONLY |
| Full pytest suite | 2 failed / 11830 passed — both non-defects (nous env flake + `test_issue4536` isolation flake) |
| Previously-brittle tests | ✅ 31/31 pass (`test_issue4662` + `test_issue1700` + `test_profile_switch_ux`) |

### Findings
**✅ Brittle tests fixed + logic refined, behavior sound:** the 3 static tests I flagged (timeoutToast window-slice, `showSessionListSkeleton()` literal, `const sessionInProgress`) are updated to match the refactor. The re-push also refined the profile-match seam: `targetActiveProfile = S.activeProfile || 'default'` (simpler than `|| data.active || name`), and the cross-profile empty-session match uses `(currentSessionProfile === targetActiveProfile || (currentSessionProfile === 'default' && !!S.activeProfileIsDefault))`. Codex verified the replace-only decision for a cross-profile empty session is still correct (empty session whose recorded profile ≠ target → replace; same-profile/has-messages → retag/keep), no misclassification of default-profile sessions, and timeoutToast:false + skeleton + active_stream_id in-progress guard all intact. 0 findings.

**ℹ️ Stale-base false-alarm noted (resolved):** a cross-merge diff (my old fail-sha → new head, spanning the #5454 merge) appeared to show a `style.css -9` removing the transparent-event keyframe — but the three-dot diff vs current master confirms #5460 touches ONLY panels.js + tests; #5454's animation removal is intact in the rebased worktree (0 keyframe refs). Not a revert.

**Ready to merge — use branch `gate-rebase/5460-cross-profile-empty-session` (`sha:6a93b00d`), NOT the PR's stale head `be92ca13`.** Converged: the 3 brittle tests are fixed and the profile-match logic refinement is sound (Codex SAFE + 31/31 + suite green bar 2 known flakes, all switch guards intact). Crown-jewel profile-switching reliability (fixes the cross-profile empty-session upload-404). Backend/JS logic — a quick profile-switch-then-upload smoke confirms, but the mechanism + guards are verified. concept 4/5. Credit @ruizanthony (co-authored). crit=3. (Gate note: checked the three-dot diff to dismiss a cross-merge style.css false-alarm — #5454 intact.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; 3 brittle tests verified fixed (31/31), profile-match logic refinement Codex-SAFE (replace-decision + guards intact), stale-base style.css false-alarm dismissed via three-dot diff. Cert valid for sha:6a93b00d._

ruizanthony:
Follow-up pushed in `a456dbdd`.

Additional UX fix:
- after a successful profile switch and after the new profile's session list has rendered, the UI now exposes that session browser by default;
- desktop: uncollapses the sidebar via `expandSidebar()`;
- mobile: opens the sidebar drawer (`mobile-panel-drawer mobile-open`), so users can pick an existing conversation in the new profile or create a new one instead of continuing from the old-profile chat context.

Local validation:
- watched the new regression tests fail before implementation;
- `python -m pytest tests/test_profile_switch_ux.py tests/test_issue4662_profile_switch_skeleton_static.py tests/test_issue1700_parallel_profile_switch.py tests/test_chat_upload_attachment_paths.py -q -o 'addopts='` → 55 passed
- `git diff --check`

Remote validation on `a456dbdd`: browser-smoke, lint, all test shards, and Greptile Review are green.

nesquena-hermes:
## Re-review — scope expanded past the certified head (`sha:6a93b00d` → `sha:d0ee2928`)

The GREEN cert above was for the panels-only fix. Since then two commits landed (`a456dbdd`, `d0ee2928`) that pull `static/sessions.js` into the change and add new cross-profile *open* behavior — that's genuinely new surface, so re-reading it here rather than treating the old cert as still valid.

### What the new commits actually change
Reading `static/sessions.js:1862-1885` at the PR head, all three sidebar open entry points (main row, lineage segment, child) now funnel through `_openSidebarSession`, which first calls `_ensureSidebarSessionProfile`:

```javascript
async function _ensureSidebarSessionProfile(session){
  const targetProfile=_sidebarSessionProfileName(session);
  if(!_showAllProfiles||!targetProfile) return false;
  const activeProfile=S.activeProfile||'default';
  if(_profileMatchesActiveProfile(targetProfile,activeProfile)) return false;
  if(typeof switchToProfile!=='function') return false;
  _profileSwitchOpeningExistingSession=true;
  try{ await switchToProfile(targetProfile); }
  finally{ _profileSwitchOpeningExistingSession=false; }
  return _profileMatchesActiveProfile(targetProfile,S.activeProfile||'default');
}
```

The `_profileSwitchOpeningExistingSession` flag is read by `switchToProfile` (`panels.js:6621`) to skip minting a blank intermediary session, then `loadSession` runs. That flag-and-load handshake is coherent and the symbols check out (`expandSidebar`/`_isDesktopWidth` are real, `boot.js:461/488`).

### Two greptile open items — my read after tracing the callers

**1. "return value of `_ensureSidebarSessionProfile` not checked."** Correct that `_openSidebarSession:1883-1884` ignores the boolean and proceeds to `loadSession` unconditionally. But that degrades to *master* behavior — before this PR every sidebar click called `loadSession` with no profile switch at all. If `switchToProfile` throws, its own `catch` (`panels.js:6691`) restores the previous profile and the subsequent `loadSession` opens the row read-only under the old cookie — the same outcome as today. So it's not a regression, just a missed opportunity to short-circuit. Low priority.

**2. "`_openingExistingSidebarSession` branch omits `syncTopbar()`."** This one is a non-issue and worth closing out explicitly: the branch at `panels.js:6621` returns without `syncTopbar()`, but the caller immediately runs `loadSession`, and every `loadSession` success path calls it (`sessions.js:1595`, `1722`, `1747`). The topbar is synced one tick later by the load, not stranded. The only window where it'd lag is if `loadSession` itself bails early — same edge as any other open path.

### The change that deserves a product/changelog line (not a blocker)
`switchToProfile` used to hard-reset `_showAllProfiles = false` on every switch. The PR removes that (`panels.js:6617`) and persists the toggle via `localStorage` (`sessions.js` `_setShowAllProfiles` / `SHOW_ALL_PROFILES_STORAGE_KEY`). Net user-visible shift: **clicking a row from another profile in the all-profiles list now silently switches the active profile** (mutates the cookie + agent context), and the all-profiles scope stays sticky across switches and browser reloads. Both are defensible, but they change what a sidebar click *means*, so flagging for the ship/UX review — the earlier cert predates this and shouldn't be read as covering it.

The static tests track the new shape well (`test_profile_switch_ux.py` browser-helper coverage, `test_issue1611` persistence assertions). As before these are source-shape greps, not DOM behavior — a small jsdom check that a cross-profile click actually flips `S.activeProfile` before `loadSession` would harden the new path, but that's an enhancement, not a gate. Behavior looks sound; re-gate the current head rather than relying on the `6a93b00d` cert.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (re-gated at expanded head d0ee2928)

**Certified head:** `sha:5e5b44e7` (clean rebase of PR head `d0ee2928`, branch `gate-rebase/5460-cross-profile-empty-session`) · **PR:** #5460 · _ruizanthony, fix(webui): replace cross-profile empty sessions on profile switch_
**Verdict:** Re-gated after a scope expansion (+288/-61 since my earlier GREEN @be92ca13). The PR now does two things: (A) replace cross-profile EMPTY sessions on switch (original), and (B) NEW: in show-all-profiles view, open a cross-profile sidebar row under its OWNING profile (switch first, then load). The new behavior strengthens profile isolation; the import-gate and profile-filtering are not weakened. Codex SAFE, suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5460c`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; three-dot diff = panels.js + sessions.js + 8 test files |
| Full pytest suite | 2 failed / 11859 passed — both non-defects (nous env flake + `test_issue4536` isolation flake) |
| Touched profile-isolation tests | ✅ 131/131 (`test_profile_switch_ux`, `#1611` filtering, `#3603` import-gate, `#1700` parallel, `#4662` skeleton, lineage) |

### Findings
**✅ Expanded behavior sound, profile isolation strengthened:** (A) the cross-profile empty-session replace (my prior GREEN) is unchanged. (B) the new `_ensureSidebarSessionProfile(session)`: when `_showAllProfiles` and the clicked session's `profile` ≠ active, it `await switchToProfile(targetProfile)` BEFORE `loadSession()` — so a cross-profile row opens under its OWNING profile (no misattribution to the active profile), gated on `_showAllProfiles` + `_profileMatchesActiveProfile` (skips if already matching), coordinated via a `_profileSwitchOpeningExistingSession` flag. Codex verified: no cross-profile misattribution/leak, the external-session-import-gate (#3603) is NOT weakened, profile-filtering (#1611) still scopes rows correctly, the coordination flag doesn't desync on switch failure, and the empty-session-replace + parallel-switch (#1700) + skeleton (#4662) paths don't regress. `_showAllProfiles` persists via localStorage safely. 0 findings, 131/131 touched tests.

**Ready to merge — use branch `gate-rebase/5460-cross-profile-empty-session` (`sha:5e5b44e7`), NOT the PR's stale head `d0ee2928`.** Supersedes my earlier GREEN @be92ca13 (scope expanded). The added "open all-profile row under owning profile" behavior actually strengthens the crown-jewel profile-switching isolation (a cross-profile session opens under its owner), with the import-gate + profile-filtering verified intact. Codex SAFE + 131/131 touched + suite green bar 2 known flakes. Profile-switching surface — a quick all-profiles-view cross-profile click smoke confirms, but the switch-before-load + isolation are verified. concept 4/5. Credit @ruizanthony (co-authored). crit=3.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Re-gated at expanded head d0ee2928; the new open-under-owning-profile behavior verified isolation-strengthening (switch-before-load, import-gate + filtering intact, coordination flag safe), Codex SAFE + 131/131 touched + suite green bar 2 known flakes. Cert valid for sha:5e5b44e7._

nesquena-hermes:
Shipped in **v0.51.848** — thanks @ruizanthony! 🎉

Cross-profile empty-session replacement on profile switch is live: switching profiles with an empty current chat no longer leaves a stale session id that 404s your next upload, and clicking a session owned by another profile now switches to that profile before loading (instead of 404ing).

Full crown-jewel gate: Codex SAFE + Opus SAFE (no cross-profile data exposure/mis-scoping; empty-session replacement only fires for a genuinely-empty differing-profile session; no regression to the profile-switch skeleton/parallel-switch race/session-profile filter/external-import gate), full suite green (2 unrelated env flakes), 54/54 targeted. Credited via Co-authored-by.

PR #5461, opened 2026-07-03
fix(api/config): surface plugin-only providers in static catalog

alexfoxtm:
## Problem

Plugin-only providers (e.g. `9router`) are not in `_PROVIDER_MODELS` and rarely ship a `providers.<slug>.models` allowlist. As a result, `_static_models_catalog_without_live_probes()` — the network-free path used for `prefer_cache` and warm-cache payloads — emitted empty groups that the picker filtered out, making keyed plugin providers look uninstalled until the live rebuild fired.

## Fix

Two complementary changes in `api/config.py`:

1. **Detect plugin providers into the catalog.** Enumerate `_plugin_model_provider_profiles()` into `detected_providers` so a keyed plugin enters the static catalog even when `providers.<slug>` is unset.

2. **Keep plugin-only groups visible** in the build loop and post-loop filter, even with no models yet — without this, the post-loop guard (`or _is_plugin_model_provider(...)`) never fires because the group was never appended in the first place when `raw_models` was empty.

3. **Fall back to the plugin's own `ProviderProfile.fallback_models`** when declared, so a plugin without a `models:` allowlist still surfaces a curated, network-free subset on the cold path. The live `_build_available_models_uncached` does a full `/v1/models` fetch and supersedes this view on the next call.

## Tests

Added `TestPluginFallbackModelsInStaticCatalog` with two regression tests:
- `test_static_catalog_surfaces_plugin_fallback_models` — empty fallback still surfaces as a group, not silently dropped
- `test_static_catalog_uses_provider_fallback_models_when_declared` — declared fallback models appear verbatim

Both are necessary: the first proves the empty-group guard works; the second proves the fallback_models wiring works. Dropping either re-introduces a slice of the regression.

`./scripts/test.sh tests/test_plugin_model_providers.py` passes 8/8. Broader sweep (provider/model/plugin/catalog/profile-resolution): 262 tests pass, 0 failures.

## Reproducer

```bash
# With a keyed plugin-only provider (env var set, no providers.<slug> block):
export YANDEX_API_KEY=...  # or 9ROUTER_API_KEY, etc.
curl -s http://localhost:8787/api/models | jq '.groups[].provider_id' | grep -i yandex
# Before: missing.  After: present.
```

## Notes

- Branch rebased onto current `master` (v0.51.834).
- AGENTS.md mandates `./scripts/test.sh` for pytest; honored.
- No new dependencies.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -35,6 +35,7 @@ import api.paths as _paths
 from api.plugin_providers import (
     effective_provider_display_name as _effective_provider_display_name,
     is_plugin_model_provider as _is_plugin_model_provider,
+    plugin_model_provider_profiles as _plugin_model_provider_profiles,
 )
 
 HOME = _paths.HOME
@@ -4488,6 +4489,23 @@ def _static_models_catalog_without_live_probes() -> dict:
             if canonical and _provider_has_key(canonical):
                 detected_providers.add(canonical)
 
+        # Plugin-only providers (e.g. 9router) are not in the static
+        # _PROVIDER_MODELS / _PROVIDER_DISPLAY tables and are detected above
+        # only when the user puts them in `providers.<slug>`.  Plugins ship
+        # with their own env-var wiring, so an installed-and-keyed plugin
+        # provider should also enter the static catalog even without a
+        # `providers:` block — otherwise the picker silently drops the
+        # group when the live-rebuild cache is cold.
+        try:
+            for _plugin_pid in list(_plugin_model_provider_profiles().keys()):
+                if not _plugin_pid or not _provider_has_key(_plugin_pid):
+                    continue
+                _canonical = _canonicalise_provider_id(_plugin_pid) or _plugin_pid
+                if _canonical:
+                    detected_providers.add(_canonical)
+        except Exception:
+            logger.debug("Plugin provider detection failed in static catalog", exc_info=True)
+
         fallback_cfg = cfg.get("fallback_providers", []) if isinstance(cfg, dict) else []
         if isinstance(fallback_cfg, list):
             for entry in fallback_cfg:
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (catalog side is correct + keyed-gated; but the surfaced plugin models MISROUTE — CORE)

**Certified head:** `sha:c02879f0` (rebased onto current master, `git apply` clean) · **PR:** #5461 · _alexfoxtm, fix(api/config): surface plugin-only providers in static catalog_
**Verdict:** The catalog half is sound — installed-AND-keyed plugin-only providers (e.g. 9router) now surface on the network-free cold-cache path, gated on `_provider_has_key` (no unconfigured leak), deduped, exception-safe, with `fallback_models`. But the PR only makes those models SELECTABLE; it doesn't wire their ROUTING — `model_with_provider_context()` doesn't preserve the provider hint for plugin-only providers absent from `providers:`, so a picked `gh/…`/`ag/…`/`ds/…` model routes to the active/default provider instead of the plugin. Half-wired.

### What I ran (rebased worktree `/tmp/wt-rebase-5461`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 1 CORE (plugin-model misroute); I confirmed the routing gap |
| Full pytest suite | 2 failed / 11831 passed — both non-defects (nous env flake + `test_issue4536` isolation flake) |
| PR's own tests | ✅ 8/8 (`test_plugin_model_providers.py`) |

### Findings
**✅ Catalog side correct:** `_static_models_catalog_without_live_probes()` now detects installed-and-keyed plugin-only providers via `plugin_model_provider_profiles()`, gated on `_provider_has_key` (skips unkeyed — no unconfigured leak), canonicalizes + dedups via `detected_providers` set, wrapped in try/except (detection failure doesn't break the catalog), and falls back to `ProviderProfile.fallback_models` so the provider isn't an empty picker-filtered group. Sound.

**⛔ CORE (I + Codex CONFIRMED) — surfaced plugin-only models route to the wrong provider** (`api/config.py`, `model_with_provider_context()`): the PR makes plugin-only slash-ID fallback models (`gh/…`, `ag/…`, `ds/…`) SELECTABLE, but `model_with_provider_context()` only preserves the provider hint for `providers:` entries, OpenRouter, and ACP providers. An installed-and-keyed plugin-only provider absent from `providers:` therefore sends its model **bare** (the `"/" in model` passthrough), so `resolve_model_provider()` routes it to the active/default provider instead of the plugin → wrong provider serves the request. So the feature surfaces models that then misroute. **Fix (Codex-exact):** in `model_with_provider_context()`, treat `_is_plugin_model_provider(provider)` like a configured explicit provider — return `f"@{provider}:{model}"` before the `"/" in model` passthrough — so the plugin provider hint is preserved end-to-end.

**RED — gate-fail/changes-requested (1 CORE): wire the routing to match the catalog — in `model_with_provider_context()`, return `@{provider}:{model}` for `_is_plugin_model_provider(provider)` (before the slash-passthrough), so a surfaced plugin-only model actually routes to its plugin, not the default provider.** The catalog-surfacing half is correct + keyed-gated + safe; this completes the other half (routing). Without it the PR lets users pick plugin models that silently hit the wrong provider. Add a routing test (`model_provider=<plugin>`, `model=gh/…` → `@<plugin>:gh/…`). concept 3/5. Author @alexfoxtm (T1). crit=3. (Gate value: a "surface it in the catalog" PR is only half the job for a routable entity — verify the selected value ROUTES correctly end-to-end, not just that it appears in the picker.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; catalog-surfacing verified keyed-gated + deduped + exception-safe, the routing CORE confirmed (model_with_provider_context omits plugin-only providers from the provider-hint-preserving branches → bare slash passthrough → default-provider misroute). Cert valid for sha:c02879f0._

nesquena-hermes:
Closing in favor of the rebuilt version — thanks @alexfoxtm! Your catalog-surfacing approach was correct (a keyed plugin-only provider should show in the static/cold catalog instead of looking uninstalled). A prior gate confirmed that half was solid; the one gap was that surfaced plugin models were selectable but not *routable*.

Rebuilt fresh on current master as **#5909** with your work preserved (`Co-authored-by`): the catalog surfacing + the missing routing fix (`model_with_provider_context()` now emits `@<plugin>:<model>` for plugin providers so a surfaced model routes to its plugin) + a routing regression test. It'll go through the normal review pass from here. Thanks for the fix and the diagnosis!

Issue #2762, opened 2026-05-22
state_sync writes to wrong profile's state.db when profile is switched via WebUI cookie


malyjacob:
**Summary:** When a user switches profiles through the WebUI interface (using the profile selector), the `sync_session_usage` function in `api/state_sync.py` writes session metadata to the **wrong profile's `state.db`** due to thread-local storage (TLS) not being propagated to the agent background thread.

---

## Bug Description

**Environment:**
- Hermes WebUI (v0.51.107) with multiple profiles (e.g., `default`/`hiyuki` and `maiko`)
- User opens WebUI (defaults to `hiyuki` profile), switches to `maiko` via the WebUI profile selector, then starts a chat

**Observed behavior:**
- Session messages and files are correctly stored in maiko's directories
- But session **metadata** (token usage, title, message count) is written to `hiyuki`'s `state.db` instead of `maiko`'s `state.db`

**Evidence:** Verified by querying both `~/.hermes/state.db` (hiyuki) and `~/.hermes/profiles/maiko/state.db`:
```
hiyuki's state.db:  FOUND session '738e6863a53c' (title: '会话ID误入hiyuki profile', 160 messages)
maiko's state.db:   NOT FOUND
```

---

## Root Cause Analysis

The bug is a **TLS (thread-local storage) propagation failure** across the streaming worker thread boundary.

### Code Path

**1. Profile switching** (`api/routes.py`, line 5504):
```python
result = switch_profile(name, process_wide=False)  # Cookie only, no global update
```

**2. HTTP request handling** (`server.py`, lines 246-247, 263-264):
```python
cookie_profile = get_profile_cookie(self)
if cookie_profile:
    set_request_profile(cookie_profile)  # _tls.profile = 'maiko' ← HTTP thread has TLS
```

**3. Agent streaming** (`api/routes.py`, lines 8227-8233):
```python
thr = threading.Thread(
    target=_run_agent_streaming,  # ← Daemon thread, TLS NOT inherited
    args=(s.session_id, ...),
    daemon=True,
)
thr.start()
```

**4. State sync call** (`api/streaming.py`, lines 5071-5081):
```python
# Called INSIDE the agent background thread (no TLS context)
sync_session_usage(
    session_id=s.session_id,
    title=s.title,
    ...
)
```

**5. `_get_state_db()`** (`api/state_sync.py`, lines 24-35):
```python
def _get_state_db():
    from api.profiles import get_active_hermes_home
[...]

PR #5463, opened 2026-07-03
fix(state-sync): mirror WebUI usage counters

ildunari:
## Thinking Path

- WebUI stores cumulative token usage in its session sidecar JSON.
- `sync_to_insights` mirrors that WebUI usage into Hermes Agent `state.db` so insights and session accounting can use one durable source.
- The bridge only mirrored prompt/completion/cost totals, so WebUI sessions with prompt-cache reads could show correct sidecar usage but `0` cache reads in `state.db`.
- The same bridge also did not pass the cumulative API-call count from streaming, leaving `api_call_count` stale in state-backed reports.
- This PR forwards the existing cumulative WebUI counters through the existing state-sync bridge and pins the profile-routed invariant in tests.

## What Changed

- Extended `api.state_sync.sync_session_usage()` to accept and forward:
  - `cache_read_tokens`
  - `cache_write_tokens`
  - optional `api_call_count`
- Updated WebUI streaming finalization to pass the session's existing cumulative cache counters and the agent's cumulative `session_api_calls`.
- Updated title/sync-chat state-sync call sites to pass cache counters when available.
- Extended the existing explicit-profile state-sync regression test to assert cache counters and API-call counts land in the named profile DB.

## Why It Matters

Without this, WebUI sessions can appear cache-cold in `state.db` even when the live agent and WebUI sidecar recorded warm prompt-cache reads. That makes insights, usage summaries, and debugging reports misleading for WebUI-originated sessions.

## Verification

Ran on a clean branch from current `origin/master`:

```bash
PIP_USER=false ./scripts/test.sh tests/test_issue2762_state_sync_profile_kwarg.py tests/test_webui_state_db_reconciliation.py
python3 -m py_compile api/state_sync.py api/routes.py api/streaming.py tests/test_issue2762_state_sync_profile_kwarg.py
PIP_USER=false python3 scripts/ruff_lint.py --diff origin/master
git diff --check origin/master..HEAD
```

Results:

- `55 passed`
- `py_compile` passed
- `ruff_lint`: no new violations on added/modified lines
- `git diff --check`: clean

Also verified locally with a temp `state.db` smoke that:

- cache counters sync into the row
- unknown `api_call_count` preserves an existing value when paired with the companion Agent-side hardening
- explicit streaming `api_call_count` updates the row

## Risks / Follow-ups

- This PR forwards `api_call_count=None` for callers that do not know the cumulative count. Preserving an existing value for that `None` case requires the companion Hermes Agent PR that makes `SessionDB.update_token_counts(... absolute=True, api_call_count=None)` preserve the current row value.
- Historical rows already clobbered to zero still need a separate reconciliation/backfill if exact old reporting is needed.

## Contract Routing

Task type: runtime/session metadata durability bug fix

Touched areas:

- WebUI state-sync bridge
- WebUI streaming writeback metadata
- explicit-profile state DB regression coverage

Relevant public docs:

- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`

State layer affected: WebUI session sidecar totals mirrored into Hermes Agent `state.db` session rows.

Evidence: targeted state-sync/reconciliation tests, syntax check, diff lint, and local temp-DB smoke.

## Model Used

AI-assisted: OpenAI GPT-5.5 via Hermes, with Codex read-only review.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -95,6 +95,8 @@ def _sync_session_title_to_insights(session) -> None:
             title=session.title,
             message_count=len(messages),
             profile=getattr(session, "profile", None),
+            cache_read_tokens=getattr(session, "cache_read_tokens", None) or 0,
+            cache_write_tokens=getattr(session, "cache_write_tokens", None) or 0,
         )
     except Exception:
         logger.debug("Failed to update session title in state.db", exc_info=True)
@@ -20237,6 +20239,8 @@ def _handle_chat_sync(handler, body):
                 model=s.model,
                 title=s.title,
                 message_count=len(s.messages),
+                cache_read_tokens=s.cache_read_tokens or 0,
+                cache_write_tokens=s.cache_write_tokens or 0,
                 # #2762 / #2827 parity with api/streaming.py:5078: pass the
                 # session's profile explicitly so a future refactor that
                 # backgrounds this handler doesn't silently leak writes to
--- a/api/state_sync.py
+++ b/api/state_sync.py
@@ -134,7 +134,8 @@ def sync_session_start(session_id: str, model=None, profile: Optional[str] = Non
 
 def sync_session_usage(session_id: str, input_tokens: int=0, output_tokens: int=0,
                        estimated_cost=None, model=None, title: Optional[str] = None,
-                       message_count: Optional[int] = None, profile: Optional[str] = None) -> None:
+                       message_count: Optional[int] = None, profile: Optional[str] = None,
+                       cache_read_tokens: int = 0, cache_write_tokens: int = 0) -> None:
     """Update token usage and title for a WebUI session in state.db.
     Called after each turn completes. Uses absolute=True to set totals
     (the WebUI Session already accumulates across turns).
@@ -153,11 +154,17 @@ def sync_session_usage(session_id: str, input_tokens: int=0, output_tokens: int=
     try:
         # Ensure session exists first (idempotent)
         db.ensure_session(session_id=session_id, source='webui', model=model)
-        # Set absolute token counts
+        # Set absolute token counts. WebUI's sidecar already accumulates
+        # input/output/cache totals across turns, so mirror the same absolute
+        # values into state.db. Omitting cache counters makes insights/reporting
+        # show false 0% hit rates even when the live stream and sidecar saw warm
+        # prefix reads.
         db.update_token_counts(
             session_id=session_id,
             input_tokens=input_tokens,
             output_tokens=output_tokens,
+            cache_read_tokens=cache_read_tokens,
+            cache_write_tokens=cache_write_tokens,
             estimated_cost_usd=estimated_cost,
             model=model,
             absolute=True,
--- a/api/state_sync.py
+++ b/api/state_sync.py
@@ -135,7 +135,8 @@ def sync_session_start(session_id: str, model=None, profile: Optional[str] = Non
 def sync_session_usage(session_id: str, input_tokens: int=0, output_tokens: int=0,
                        estimated_cost=None, model=None, title: Optional[str] = None,
                        message_count: Optional[int] = None, profile: Optional[str] = None,
-                       cache_read_tokens: int = 0, cache_write_tokens: int = 0) -> None:
+                       cache_read_tokens: int = 0, cache_write_tokens: int = 0,
+                       api_call_count: Optional[int] = None) -> None:
     """Update token usage and title for a WebUI session in state.db.
     Called after each turn completes. Uses absolute=True to set totals
     (the WebUI Session already accumulates across turns).
@@ -165,6 +166,7 @@ def sync_session_usage(session_id: str, input_tokens: int=0, output_tokens: int=
             output_tokens=output_tokens,
             cache_read_tokens=cache_read_tokens,
             cache_write_tokens=cache_write_tokens,
+            api_call_count=api_call_count,
             estimated_cost_usd=estimated_cost,
             model=model,
             absolute=True,
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -9100,6 +9100,7 @@ def _run_agent_streaming(
                             message_count=len(s.messages),
                             cache_read_tokens=s.cache_read_tokens or 0,
                             cache_write_tokens=s.cache_write_tokens or 0,
+                            api_call_count=getattr(agent, 'session_api_calls', None),
                             # #2762: pass the session's profile explicitly so the
                             # background-thread state.db lookup doesn't fall
                             # through to the process-global active profile and
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -9098,6 +9098,8 @@ def _run_agent_streaming(
                             model=model,
                             title=s.title,
                             message_count=len(s.messages),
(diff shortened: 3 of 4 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (confirmed CORE: `api_call_count` clobbered to NULL in state.db)

**Certified head:** `sha:21f32b7e` (rebased onto current master, `git apply` clean) · **PR:** #5463 · _ildunari, fix(state-sync): mirror WebUI usage counters_
**Verdict:** The cache-token mirror is correct (`cache_read/write_tokens or 0`, null-safe), but the PR pipes a possibly-`None` `api_call_count` into `update_token_counts(..., absolute=True)`, whose absolute SQL assigns `api_call_count = ?` DIRECTLY (not `COALESCE`) — so a sync with `api_call_count=None` wipes an existing count to NULL. Codex reproduced it (temp DB: `api_call_count=5` → `NULL`). Data loss.

### What I ran (rebased worktree `/tmp/wt-rebase-5463`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 1 CORE, **empirically reproduced** the NULL-clobber |
| Full pytest suite | 2 failed / 11829 passed — both non-defects (nous env flake + `test_issue4536` isolation flake) |
| Downstream SQL trace (mine) | ✅ confirmed `hermes_state.py:2185` absolute branch: `api_call_count = ?` (bare, no COALESCE) |

### Findings
**⛔ CORE (I + Codex CONFIRMED, empirically reproduced) — `api_call_count` clobbered to NULL** (`api/state_sync.py:169` → `hermes_state.py:2185`): `record_state`/`sync_session_usage` passes `api_call_count` to `db.update_token_counts(..., absolute=True)`. In the `absolute=True` branch, EVERY other field uses `COALESCE(...)` to preserve, but `api_call_count = ?` is a **bare direct assignment**. The PR sources it from `getattr(agent, 'session_api_calls', None)` (streaming.py:9103, can be None), and `sync_session_usage()` is also called WITHOUT an api count at `routes.py:89` and `routes.py:20234` (defaults to None). So any such sync runs `SET api_call_count = NULL`, **wiping the accumulated count**. Codex verified: an isolated temp DB went `api_call_count=5` → `NULL` on a `None` call. (Contrast: the PR correctly guards cache tokens with `or 0`.) **Fix (Codex-exact):** preserve the existing DB value when `api_call_count is None` before calling `update_token_counts` (e.g. don't pass it, or pass a sentinel), OR change the absolute SQL to `api_call_count = COALESCE(?, api_call_count)`; add a regression test: existing `api_call_count=7` + later `sync_session_usage(..., api_call_count=None)` leaves it `7`.

**✅ Sound:** the cache-token mirror (`cache_read_tokens/cache_write_tokens or 0`) is correct and the stated goal (fix false 0% cache-hit rates in insights) is valid.

**RED — gate-fail/changes-requested (1 CORE, data loss): make `api_call_count=None` PRESERVE the existing state.db value — either don't pass None through to the absolute `update_token_counts` write, or make the SQL `api_call_count = COALESCE(?, api_call_count)` — + a regression test proving an existing non-null count survives a `None` sync.** The cache-token half is correct and shippable; only the api_call_count null-clobber blocks it. This is exactly the concern the warm-up dossier flagged, now confirmed with the downstream SQL + an empirical repro. concept 3/5. Author @ildunari (T1). crit=2. (Gate value: a downstream-subsystem behavior — `update_token_counts`'s bare `api_call_count = ?` under `absolute=True` — turned a plausible-looking counter mirror into silent data loss; tracing into `hermes_state.py` was necessary to confirm.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; the NULL-clobber traced to hermes_state.py:2185 (`api_call_count = ?` bare in the absolute branch, unlike every COALESCE'd sibling) + confirmed the PR passes possibly-None from 3 call sites, Codex empirically reproduced 5→NULL. Cache-token mirror verified null-safe. Cert valid for sha:21f32b7e._

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (full gate: Codex + Fable + suite) · ⏸️ state.db write → holding for maintainer merge

**Certified head:** `release/stage-5463` (byte-identical rebase of PR head onto current master; PR was 493 behind) · **PR:** #5463 · _ildunari (kosta), fix(state-sync): mirror WebUI cache-token + api-call counters_

**Verdict: GREEN — ready to ship on your nod.** The prior RED cert (api_call_count NULL-clobber) plus two deeper regression modes the gate surfaced this pass are all fixed WebUI-side and mutation-tested. The valuable goal (correct cache-hit rates + api-call counts in insights for WebUI sessions) is delivered.

### The api_call_count axis — 3-round convergence (all fixed here, all tested)
`update_token_counts(absolute=True)` assigns `api_call_count = ?` DIRECTLY (no COALESCE, unlike every other field), and the source value (`agent.session_api_calls`) is per-in-memory-agent — it resets to 0 on a rebuilt agent (restart / SESSIONS-LRU evict / cached-agent rebuild). Three ways that corrupted the durable count, each now closed:
1. **NULL-clobber** — title-only/chat-sync callers pass `None` → `sync_session_usage` reads the existing count and passes it back (absolute write is a no-op for the field).
2. **0-bypass** — a fresh agent reports `session_api_calls == 0`; the streaming call site now maps a falsy value to `None` (`... or None`) so it routes into the preserve path.
3. **positive-lower regression** — after a rebuild the agent counts up from 1; a **monotonic floor** (`max(incoming, existing)`) now prevents any absolute write from lowering the stored count.

Residual (non-blocking, follow-up): fully cumulative accuracy across rebuilds ideally wants agent-side counter seeding from state.db (or an `api_call_count = MAX(?, api_call_count)` in the agent's absolute SQL). The monotonic floor makes the WebUI side correct in the meantime.

### Gate
- **Codex** (3 rounds → SAFE): monotonic floor prevents 7→1 regression while still advancing on a higher count; None-preserve + cache mirror correct; exception-safe.
- **Fable**: cache mirror correct (absolute, no double-count), #2762 profile-scoping intact, None-preserve reads the right column + safe for new sessions; its round-1 finding is the fix folded in.
- **Full suite:** 12307 passed / 0 failed. Mutation-checked regression tests for all three api_call_count modes + the cache mirror.

**Holding the merge for @nesquena** — this writes `state.db` (a durable store), so per the overnight bar I'm not auto-merging it unattended even though it's gate-clean. Merge from `release/stage-5463` (or say the word and I'll ship it). Credit preserved via Co-authored-by.

nesquena-hermes:
Shipped in **v0.51.921** — thanks @kosta963. 🎉

The cache-token + API-call-count counters now mirror to `state.db` at every usage-sync site, so Insights shows real cache-hit rates and per-session call counts instead of a false 0%. The API-call count uses a monotonic floor (`max(incoming, existing)`) so a rebuilt agent (server restart / session-cache eviction) whose in-memory counter reset to 0 can't clobber the durable total downward.

Gate (all legs, on a stage rebased onto current master): Codex SAFE (over 3 convergence rounds — NULL-clobber → 0-bypass → positive-lower → monotonic-floor), Fable mechanism-clear, full suite 12223/0 + 14 targeted. Your commits landed with authorship preserved. Deployed and verified live.

Issue #5455, opened 2026-07-03
Hermes WebUI feels laggy: /api/sessions and session loads take seconds


avinashkanaujiya:
## Summary
Hermes WebUI feels noticeably laggy in normal use.

## Observed behavior
- `/api/sessions` often takes several seconds.
- Opening a larger session via `/api/session?...messages=1&resolve_model=0&msg_limit=8&expand_renderable=1` can take 5–20s+.
- The UI feels frozen/slow while these requests are in flight.

## Reproduction
1. Open Hermes WebUI.
2. Let the sidebar/session list load or refresh.
3. Open a larger session (especially one with many tool calls / cron activity).
4. Observe slow responses in the browser and in `hermes-webui` logs.

## Environment
- Self-hosted Docker deployment
- Image: `ghcr.io/nesquena/hermes-webui:latest`
- Browser UI accessed through the WebUI reverse proxy

## Notes
Recent server logs show slow-request watchdog output for `/api/sessions`, with the hot stage in `get_cli_sessions`.

## Related issue
- #5049: perf(sidebar): profile zero-message visibility pushdown under empty-session-heavy installs

avinashkanaujiya:
Additional context: when the slowness was visible, there was a running session using GPT-5.4 to update some clone jobs and skills. Also, I have seen that from time to time, the Hermes WebUI shows that the Hermes backend is unreachable, but the session is running. You can see the thinking stream and all that. So I think the problem is with the situation where there's a running session, and then for some reason, Hermes WebUI is unable to do other things. It's sort of like it's blocked to the same thread or something, that it's unable to do parallel things, maybe.

nesquena-hermes:
Thanks for the detailed report — the "backend unreachable *while a session is running*, can't do parallel things" symptom is the useful clue here, and it points at a real lock-contention bug on the `/api/sessions` hot stage you named (`get_cli_sessions`). I traced the projection path and I think I can name the layer.

## Summary

The hot stage inside `get_cli_sessions` is the SQLite projection in `read_importable_agent_session_rows` (`api/agent_sessions.py:428`). That function opens the agent's **live** `state.db` with a plain read-write connection and no timeout override, then issues a **write** on the read path. When the agent is mid-stream (writing message rows) and `state.db` is not in WAL mode, that read stalls behind the writer up to SQLite's default **5-second** busy timeout. That is your "several seconds" and your "frozen while a turn streams."

## Code reference

[...]

PR #5464, opened 2026-07-03
perf(sessions): open the session-listing projection read-only (#5455)

ai-ag2026:
## Problem
`read_importable_agent_session_rows()` — the `/api/sessions` sidebar projection — is a pure read, but it opened a **read-write** SQLite connection on the live (multi-GB, WAL) `state.db` and re-ran a defensive `CREATE INDEX` self-heal on every sidebar build. Holding a write-capable handle while the agent streams into the same DB adds needless checkpoint/lock surface. Part of the #5455 investigation into `/api/sessions` latency.

## Change
- Open the projection connection **read-only** (`file:...?mode=ro`).
- Run the `idx_messages_session` self-heal through a **separate short-lived writable connection**, and only when the index is actually missing. With the index present (the normal case) the read path performs **zero writes**.

## Behavior
Output-neutral: identical rows when the index exists; self-heal preserved when missing; read-only-filesystem degradation to the pre-aggregated path unchanged.

## Tests
`tests/test_issue5455_listing_readonly_connection.py` — read-only open, no writable connection when the index is present, self-heal via a separate connection when missing. `pytest` + the existing agent-session listing suite pass.

Refs #5455.

### DIFF
--- a/api/agent_sessions.py
+++ b/api/agent_sessions.py
@@ -453,7 +453,16 @@ def read_importable_agent_session_rows(
         return []
 
     log = log or logger
-    with closing(sqlite3.connect(str(db_path))) as conn:
+    # Open read-only for this projection/listing path: it is a pure read, and
+    # holding a write-capable handle on the live (multi-GB, WAL) state.db while
+    # the agent streams into it adds needless checkpoint/lock surface (#5455).
+    # The defensive index self-heal below still runs, but through a separate
+    # short-lived writable connection on the rare missing-index path only.
+    try:
+        conn = sqlite3.connect(f"file:{db_path}?mode=ro", uri=True)
+    except sqlite3.Error:
+        conn = sqlite3.connect(str(db_path))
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.838** — thanks @ai-ag2026! 🎉

Part of the response to #5455 ("WebUI feels laggy"). The session-listing projection now opens the state database read-only (`file:...?mode=ro`) instead of holding a write-capable handle on the live multi-GB WAL db just to read, and the defensive index self-heal only touches it with a separate short-lived writable connection on the rare missing-index path. Output is unchanged.

Gate: Codex SAFE TO SHIP (reproduced the read-only-open fallback + self-heal paths), Opus safe (byte-identical output across every degradation path I could construct, no fd leaks), full suite green (2 unrelated pre-existing env flakes). Batched with #5465. Credited via `Co-authored-by`.

PR #5466, opened 2026-07-03
perf(stream): render live prose incrementally, cut per-frame/per-token O(n^2) work (#5455)

ai-ag2026:
## Problem
On the streaming render path, cost grew **O(n²)** with the length of the answer (n = accumulated text), causing progressively worse jank on long responses. Three contributors, all over the growing text:

1. In compact-worklog mode the anchor activity scene re-renders every frame, and its live prose row was rebuilt with `renderMd(fullText)` each time — re-parsing the entire answer per frame (the incremental `smd` parser the main body already uses was bypassed and its segment hidden).
2. `snapshotLiveTurn()` serialized the whole live turn via `turn.outerHTML` on **every** render frame.
3. The token handler ran `_parseStreamState()` (strip tool-call XML + extract inline thinking over the **full** text) on **every token**, plus a full-string `toLowerCase()` allocation in `_stripXmlToolCalls`.

## Change
1. Keep a persistent per-segment `smd` parser + DOM node (same safe renderer as the live body), feed it only the delta each frame, and hand that node back to the scene builder. Prose render goes O(n) → O(delta).
2. Throttle the per-frame snapshot to a coarse trailing capture. It only backs mid-stream session-switch restore, and the switch path plus tool/done boundaries already capture synchronously.
3. Skip the per-token parse once the assistant row exists (`_doRender` re-parses once per throttled frame anyway); replace the `toLowerCase()` allocation with a non-allocating regex test.

## Behavior
Output-neutral by construction — settled rows and any failure (no `smd`, no stable key, parser error, text divergence) fall through to the existing `renderMd` path; the row-creation gate and snapshot semantics are preserved.

## Verification
In-browser profiling on a long streamed answer: per-frame anchor-scene time is **flat** across first-third vs last-third of the response (was steeply rising); snapshot calls drop to the throttle rate; few/short long-tasks; output visually identical including post-tool-call segments. `node --check` + ESLint runtime guard pass.

Refs #5455.

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -3508,6 +3508,54 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
     },null);
     return _findAnchorActivityEventByLocalId(localId,'token');
   }
+  // Persistent incremental renderer for anchor-scene live prose rows. The compact
+  // worklog re-renders the whole scene each frame; rendering the growing prose via
+  // renderMd(fullText) every frame is O(n^2) over a long answer. Instead keep a
+  // per-segment smd parser + node (the SAME safe renderer as the main live body)
+  // and feed only the delta, then hand the persistent node back to the ui.js scene
+  // builder. Returns null whenever smd or a stable key is unavailable so the caller
+  // falls back to the full renderMd path — identical structure, just not
+  // incremental. (#5455 WS2.1)
+  const _anchorProseSmdCache = new Map();
+  function _anchorProseIncrementalNode(key, text){
+    if(!window.smd || !key || typeof _safeSmdRenderer!=='function') return null;
+    const value=String(text||'');
+    try{
+      let st=_anchorProseSmdCache.get(key);
+      // Self-heal desyncs (edit/sanitize made the text no longer a pure append):
+      // rebuild the parser+node from scratch, mirroring _smdWrite's guard.
+      if(st && st.writtenText && !value.startsWith(st.writtenText)) st=null;
+      if(!st){
+        const node=document.createElement('div');
+        node.className='assistant-segment';
+        node.setAttribute('data-anchor-scene-prose','1');
+        const body=document.createElement('div');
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in **v0.51.839** — thank you @ai-ag2026! 🎉

The incremental live-streaming render (for #5455) is now live: the Transparent Stream / compact worklog live prose renders incrementally via the streaming markdown parser (fed only the delta) instead of re-rendering the whole growing answer every frame, the restore snapshot is throttled, and the redundant per-token full-text parse is skipped once the answer row exists. Long streaming answers no longer get quadratically heavier.

This is the crown-jewel streaming surface, so it went through the full gate and I fixed every finding on the release branch before shipping (all credited to you via `Co-authored-by`):

1. **SILENT (fixed):** the incremental prose node returned early and lost its `data-anchor-scene-row` / `-row-id` / `-row-role` / `-source-event-type` identity attributes — now it flows through the shared attribute-stamping block.
2. **SILENT (fixed):** the snapshot-timer / anchor-cache / window-global teardown only ran on fallback/done/stream_end — now also on `apperror`, `cancel`, `_handleStreamError`, and `_restoreSettledSession`, with a regression test pinning all four.
3. **Equivalence (verified, not a defect):** the incremental smd render was independently proven byte-identical to a whole-text smd parse against the real `smd.min.js` (1498 frames + 1196 frames across two advisors), and settled DOM still renders through the canonical `renderMd` path (`settled:true` gate) — so final output is unchanged.

Also fixed a test-only harness gap the fix exposed: the `test_issue5224` node harness now stubs the two new cleanup helpers (they threw as `undefined` under CI sharding).

One non-blocking perf-only follow-up noted: in transparent mode, if an incremental frame fails then recovers mid-segment, that one segment reverts to per-frame serialization until the next turn (output stays correct). Filing as a minor follow-up.

PR #5467, opened 2026-07-03
perf(sidebar): skip full session-list rebuild when the payload is unchanged (#5455)

ai-ag2026:
## Problem
The 30s streaming poll and the gateway poll re-applied `/api/sessions` and ran a full sidebar DOM rebuild (`list.innerHTML='' ` + one node tree per row) on every tick — **even when the payload was byte-identical** to what was already on screen, the common case on an idle sidebar. Part of the #5455 latency work.

## Change
Add a signature over the applied rows + projects + coarse display state in `_applySessionListPayload`, and skip `renderSessionListFromCache()` when it matches the last rendered signature and no entrance animation is pending. A streaming row's fields advance each poll, so its signature changes and it still repaints. Client-only display toggles (search, active, density, select, lineage expansion) render through their own paths. The per-refresh `INFLIGHT` cleanup still runs on the skip path, and the 60s relative-time refresh plus every other render trigger call `renderSessionListFromCache` directly and are unaffected.

## Behavior
Output-neutral: a skip only happens when the render would produce the same DOM.

## Verification
In-browser: over an idle window, applied-payloads > DOM-rebuilds (redundant rebuilds skipped); sidebar still updates immediately on real changes (new session, title change, session switch, streaming indicator). `node --check` + ESLint runtime guard pass.

Refs #5455.


nesquena-hermes:
Shipped in **v0.51.844** — thanks @ai-ag2026! 🎉

The sidebar render-skip on idle polls is live (for #5455): the session list now skips its full DOM rebuild when a signature of the render inputs is unchanged, and rebuilds whenever anything visible changes.

This is a correctness-sensitive change (a false-skip = a stale sidebar), so it went through a deep dual gate and I fixed every finding on the release branch before shipping (all credited to you):
- **Round 1 (3 fixes):** the signature omitted rendered fields (pending/running streaming state, attention dots, source/read-only/worktree/lineage/model/profile + reference rows). Fixed by serializing the FULL applied row objects + `_sidebarReferenceSessions` instead of a hand-picked allowlist — inherently complete, can't rot when a new rendered field is added.
- **Round 2 (2 fixes):** recovering from the loading skeleton or the "couldn't load conversations" banner (both rendered outside the signature path) could false-skip and leave stale DOM. Fixed by capturing `_hadSessionListSkeleton` / `_hadSessionListLoadError` before clearing and forcing a real render on recovery.
- **Round 3:** Codex SAFE TO SHIP + Opus safe to ship (no stale-sidebar path remains). Full suite green (2 unrelated pre-existing env flakes). Regression tests added for all of the above.

PR #5468, opened 2026-07-03
perf(i18n): split locale bundles, load only the active language + English (#5455)

ai-ag2026:
## Problem
`static/i18n.js` was a single ~1.5 MB module holding **all 14 language bundles**, eagerly parsed on every page load even though only one language is ever used. Because `t()` falls back to English for missing keys, English is always needed — but the other 12+ languages were pure parse/allocation overhead at boot. Part of the #5455 latency work (startup cost).

## Change
Split (AST-based, byte-exact) into:
- `static/i18n/<code>.js` — one bundle per language (registers into `window.__I18N`).
- `static/i18n/_helpers.js` — the hoisted tool-label/elapsed helper functions + tables that bundle values reference (must load before the bundles).
- `static/i18n.js` — a small core (`t`/`setLocale`/`loadLocale`/`resolveLocale`/`applyLocaleToDOM` + `AVAILABLE_LOCALES`/`LOCALE_LABELS`).

English + the active language load **synchronously** via a `document.write` shim in `index.html` (no flash of untranslated text). Other languages load **lazily** on switch: `setLocale` fetches the bundle, then applies and re-stamps the DOM. The settings language picker lists all languages from the static `AVAILABLE_LOCALES`/`LOCALE_LABELS` table instead of the loaded bundles. The service worker precaches core + `_helpers` + `en` for the offline shell.

## Impact
Boot i18n payload: **~1.5 MB → ~122 KB (English) / ~224 KB (e.g. German)** of parsed JS.

## Behavior
Output-neutral. An AST reconstruction check confirms the split is **data-equivalent** to the original: **14 languages, 22,237 keys, 0 differences** (function values compared by source, others by JSON).

## Verification
- Data-equivalence check (above).
- In-browser: text renders correctly with no English flash; the picker shows all 14 languages; switching to another language lazy-loads its bundle and updates live; the Network panel shows only core + `_helpers` + `en` + the active language loaded.
- `node --check` + ESLint runtime guard clean.

Refs #5455.


ai-ag2026:
Marking this PR back to draft intentionally.

The browser/runtime side of the lazy locale bundle split was verified in the fork/live branch, but upstream CI currently has a broad static-test contract that reads `static/i18n.js` directly and expects every locale to be inline in that file. Making this PR green would require a wider test-contract migration to a shared i18n loader across many tests, not a small implementation fix.

Recommendation: keep this branch as fork/live-only performance work for now, or reopen it later with an explicit contract-change PR that updates the i18n test surface deliberately.

ai-ag2026:
Closing this as stale/superseded rather than force it through.

Two blockers make a simple revive impractical:

1. **Massively stale** — the branch is ~511 commits behind `master`; the i18n surface has moved substantially since (hundreds of keys added across locales), so the split bundles no longer match and every i18n consistency check fails.

2. **Structural, not just stale** — ~147 test files read the monolithic `static/i18n.js` directly (via `read_text("static/i18n.js")` + `source_between` key/block greps). Splitting the monolith into per-locale bundles breaks all of them by construction, independent of staleness. A correct lazy-bundle change therefore isn't just a re-split — it needs a companion decision on how i18n is validated in tests (migrate the ~147 greps to a runtime loader, or keep `i18n.js` as a grep-able source that a serve/build step splits for runtime), which also touches the repo's deliberate no-build-step design.

That's a **Path 2 / architecture-alignment** change per CONTRIBUTING, not a solo drive-by. Happy to reopen a fresh, current-`master`-based version if there's maintainer interest in the lazy-load direction and agreement on the test-validation approach. The backend side of the underlying #5455 lag is addressed separately in #5780 (read-only `state.db` opens).

PR #5471, opened 2026-07-03
fix(composer): suppress stale draft restore after send

ruizanthony:
## Summary

- suppress same-session restoration of stale non-empty `composer_draft` immediately after a message submission clears the composer
- clear persisted composer drafts from busy send paths (`queue`, `interrupt`, and delivered `steer`) instead of only clearing the textarea
- add regressions for stale draft restoration after send and for busy-path draft clearing

## Why

Mobile/browser timing can show a sent user bubble while the composer is empty locally but `/api/session` still returns the previous non-empty `composer_draft`. A same-session refresh can then repopulate `#msg` with the tail of the already-submitted message.

`_clearComposerDraft()` already posts an empty draft, but that POST is asynchronous. This change adds a short local suppression window so old server draft payloads cannot be restored before the empty draft response catches up. Fresh user typing clears the suppression.

## Tests

- `/usr/local/lib/hermes-agent/venv/bin/python -m pytest tests/test_issue5060_boot_composer_preserve.py tests/test_composer_draft_after_send.py tests/test_issue_new_chat_draft_restore.py tests/test_real_steer.py -q`

Result: `39 passed`

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -1029,8 +1029,10 @@ function _composerTextWithPendingSelections(){
 function _clearComposerAfterQueuedSelectionSend(){
   const sid=arguments.length?arguments[0]:(S.session&&S.session.session_id);
   const composer=(typeof $==='function'&&$('msg'))||document.getElementById('msg');
+  const draftText=composer?String(composer.value||''):'';
+  const draftFiles=Array.isArray(S.pendingFiles)?[...S.pendingFiles]:[];
   if(composer)composer.value='';
-  if(sid&&typeof _clearComposerDraft==='function') _clearComposerDraft(sid);
+  if(sid&&typeof _clearComposerDraft==='function') _clearComposerDraft(sid,draftText,draftFiles);
   _clearPendingSelections();
   if(typeof autoResize==='function') autoResize();
 }
@@ -1308,7 +1310,7 @@ async function send(){
       const _modelState=_chatPayloadModelState();
       queueSessionMessage(_targetSid,{text:_text,files:[...S.pendingFiles],model:_modelState.model,model_provider:_modelState.model_provider,profile:S.activeProfile||'default'});
       _clearComposerAfterQueuedSelectionSend();
-      if(_targetSid&&typeof _clearComposerDraft==='function') _clearComposerDraft(_targetSid);
+      if(_targetSid&&typeof _clearComposerDraft==='function'&&_targetSid!==(S.session&&S.session.session_id)) _clearComposerDraft(_targetSid,_text,S.pendingFiles?[...S.pendingFiles]:[]);
       S.pendingFiles=[];renderTray();
       updateQueueBadge(_targetSid);
       showToast(`Queued: "${_text.slice(0,40)}${_text.length>40?'…':''}"`,2000);
@@ -1365,6 +1367,7 @@ async function send(){
       }
     const defaultMessageMode=window._defaultMessageMode||'steer';
       if(defaultMessageMode==='steer'&&S.activeStreamId&&typeof _trySteer==='function'){
+        const _steerDraftFiles=Array.isArray(S.pendingFiles)?[...S.pendingFiles]:[];
         // Real steer: clear the input first so the user gets immediate
         // feedback, then ship the steer payload via /api/chat/steer.
         // _trySteer restores the draft and leaves the active stream running if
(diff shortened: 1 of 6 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (per-sid suppression is well-built, but a cross-tab NEW draft is silently suppressed for 30s — SILENT)

**Certified head:** `sha:94b2d674` (rebased onto current master, `git apply` clean) · **PR:** #5471 · _ruizanthony, fix(composer): suppress stale draft restore after send_
**Verdict:** The per-sid 30s suppression window is cleanly implemented (self-cleaning TTL Map, immediate local-state reset, correct early-clear on empty). But it discriminates only by `sid` + TTL, not by draft payload — so a legitimately NEW draft saved by another tab/client for the same session during the window is silently ignored for up to 30s. Make suppression payload-signature-specific.

### What I ran (rebased worktree `/tmp/wt-rebase-5471`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 1 SILENT (cross-tab new-draft suppressed); I confirmed by inspection |
| Full pytest suite | 2 failed / 11834 passed — both non-defects (nous env flake + `test_issue4536` isolation flake) |
| PR's own tests | ✅ 9/9 (`test_composer_draft_after_send` + `test_issue5060_boot_composer_preserve`) |

### Findings
**✅ Suppression mechanics sound:** `_suppressComposerDraftRestoreAfterSubmit(sid)` sets `_composerDraftRestoreSuppressedUntilBySid[sid] = Date.now()+30000` + immediately resets local draft state (`_rememberComposerDraftPayloadState(sid,'',[])`) so a same-session refresh can't race the old draft back in. `_isComposerDraftRestoreSuppressed` is TTL-self-cleaning (`delete` on `Date.now()>until`), `!sid`-guarded; the empty-payload path clears the suppression early. Per-sid isolated, no unbounded growth for re-read sids. 9/9 tests.

**⛔ SILENT (I CONFIRMED) — a cross-tab NEW draft is silently ignored for up to 30s** (`static/sessions.js:214`, `_isComposerDraftRestoreSuppressed(sid)` at :57): suppression is keyed ONLY by `sid` + TTL — the function takes no payload. The restore guard is `if (restoreSid && hasServerDraftPayload && _isComposerDraftRestoreSuppressed(restoreSid)) return;` — so ANY server draft with a payload during the window is suppressed, whether it's the stale just-submitted one OR a genuinely new draft saved by another tab/client. `/api/session/draft` carries only `{text, files}` (no revision/author), so `_restoreComposerDraft()` can't distinguish them → a multi-tab user loses visibility of a new draft for up to 30s. **Fix (Codex-exact):** make suppression payload-specific — store the cleared draft's signature at suppress time and suppress only when the incoming restore payload MATCHES that signature; if it differs, `delete` the suppression and restore normally.

**RED — gate-fail/changes-requested (1 SILENT): make the draft-restore suppression payload-signature-specific — suppress only when the incoming server draft matches the just-cleared draft's signature; a differing (new) payload deletes the suppression and restores.** The per-sid TTL window is otherwise well-built (self-cleaning, immediate local reset, correct empty-clear); this closes the multi-tab hole where a new draft is silently swallowed for 30s. Add a test: session suppressed after send + a DIFFERENT server draft arrives within 30s → it IS restored. concept 4/5. Author @ruizanthony (T2). crit=3. (Gate note: a time-window suppression keyed only on identity+TTL can't tell "the thing I just cleared" from "a new thing with the same key" — discriminate by content signature, not just the key, when a concurrent writer can produce a legitimately-different value for the same key mid-window.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; suppression mechanics verified sound (per-sid TTL self-clean + immediate local reset + empty-clear), the cross-tab new-draft SILENT confirmed (`_isComposerDraftRestoreSuppressed` takes only sid, guard suppresses any payload in-window). Cert valid for sha:94b2d674._

nesquena-hermes:
Shipped in **v0.51.856** 🎉 — merged via release `stage-5471`.

Thanks @ruizanthony — this closes the composer repopulating the tail of a just-sent message. Your payload-signature re-push was the right direction; the gate then took it the rest of the way to a full root fix.

The full gate (Codex + Opus, both reproducing by execution) found the exact-match signature still failed open in the two most common cases:
- **Attachment:** a live `File` serializes to `{}` when the draft is persisted, so a staged-file draft's signature never matched the round-tripped server draft.
- **Prefix (the actual field symptom):** the server's last persisted draft is usually a *prefix* of the submitted text (Enter-to-send within 400ms cancels the pending debounced save), so matching only the submitted text missed it and the prefix restored.

Root fix: suppression now captures **both** the remembered server-draft signature and the submitted-payload signature, and suppresses a restore matching either — while a genuinely new cross-tab draft matches neither and restores immediately. Files are canonicalized before persist+sign as belt-and-suspenders.

- **Codex: SAFE TO SHIP** · **Opus: SHIP (GREEN)** — both prior findings verified closed, no over-suppression
- **Full suite green** (11983 passed) with new regressions for the prefix + attachment + cross-tab cases

Closing as shipped — thanks for the contribution!

PR #5473, opened 2026-07-03
fix(webui): start new chat when switching workspaces

ruizanthony:
## Summary

- Treat selecting a different workspace as a new-chat boundary instead of mutating the current session in place.
- Keep the existing conversation attached to its original workspace while the new conversation starts with the selected workspace.
- Preserve same-workspace selection as a refresh/no-op, so it does not create duplicate chats.
- Update the existing workspace-switch regression and add a frontend regression test for the picker behavior.

## Why

Workspace changes can alter the project context files loaded by the agent. Starting a new chat for a different workspace avoids reusing a session whose cached prompt/context was created for another workspace.

## Tests

- `python -m pytest tests/test_workspace_blank_page_fix.py tests/test_workspace_switch_new_session.py tests/test_parallel_session_switch.py -q`


nesquena-hermes (changes requested):
Thanks for this, and for the thorough Node/jsdom test harness. There's one hard blocker that makes the switch fail at runtime — flagging it as changes-requested with a precise repro so it's a quick fix.

## Blocker: `switchToWorkspace` calls an undefined function → `ReferenceError` on every switch

`static/panels.js` (as rewritten in this PR) calls `syncActiveProjectForWorkspace(...)` in **both** branches of the switch path:

- `panels.js:5901` — the same-workspace early-return branch
- `panels.js:5924` — the new-chat branch

But `syncActiveProjectForWorkspace` **is not defined anywhere in the served front-end**. Verified statically on this PR head and on `origin/master`:

```
$ grep -rn "syncActiveProjectForWorkspace\s*=\|function syncActiveProjectForWorkspace" static/ --include=*.js
# (no output — 0 definitions)
$ git grep -n "syncActiveProjectForWorkspace" origin/master -- static/
# (no output — does not exist on master either)
```

Both call sites sit inside the `try { … } catch(e){ setStatus(t('switch_failed')+e.message) }` block, so at runtime every workspace switch throws `ReferenceError: syncActiveProjectForWorkspace is not defined` and the user just sees **"Switch failed: syncActiveProjectForWorkspace is not defined"** — the switch never completes. The feature is non-functional as written.

## Why the PR's own test passes anyway (false green)

`tests/test_workspace_switch_new_session.py` extracts `switchToWorkspace` and runs it in Node with a hand-built global scope. Line 74 defines a **stub** for the missing function:

```js
globalThis.syncActiveProjectForWorkspace = (...args) => record('syncActiveProjectForWorkspace', args);
```

So the test provides the function the real app is missing — it passes green while the browser path throws. The test is masking the bug rather than catching it.

## Fix-spec

Pick one, then make the test prove the real path:

1. **Define `syncActiveProjectForWorkspace(workspace, projectId)`** in `static/panels.js` (or wherever the active-project state lives) with the behavior the two call sites expect (sync the active-project UI/state for the given workspace + project id). The closest existing helpers are `setActiveProjectFilter(...)` and the `_activeProject` state — if one of those is what you actually meant, call it directly instead.
2. **Or** replace both `syncActiveProjectForWorkspace(...)` calls with the intended existing function.

Then adjust `test_workspace_switch_new_session.py` so it does **not** stub `syncActiveProjectForWorkspace` — either load the real `static/panels.js` definition into the harness, or assert against the real helper — so the test would fail if the function is missing again.

## Separate note (not a blocker — product call for the maintainer)

This PR also changes the workspace-switch semantics: switching now **starts a fresh chat session** (via `newSession`) instead of updating the current session's workspace in place, and it removes the earlier blank-page auto-create block. That's a reasonable design, but it's a user-visible behavior change, so it'll get a product/UX look on the ship review regardless of the crash fix above. No action needed from you on that — just a heads-up so the behavior change is intentional and documented in the PR description.

Once the undefined-function call is resolved and the test exercises the real path, this should be quick to re-review.

nesquena-hermes:
Thanks for this — the underlying concern is legitimate (a workspace switch changes the project-context files the agent loaded, so reusing the same session can carry stale cross-workspace context). But as written this **silently changes a shipped default behavior**: selecting a different workspace would now always start a *new* chat instead of continuing the current one, for everyone. That's a product decision we don't want to flip silently — some users switch workspace mid-conversation specifically to reference files without losing their thread.

**Fix-spec to land this:**
- Make the new-chat-on-workspace-switch behavior **opt-in** (a Settings preference, default OFF → preserves today's behavior), **or**
- Turn it into a lightweight **confirmation** when switching workspace mid-conversation ("Start a new chat for this workspace?" with a "keep current" option), rather than an unconditional new-chat.

Either shape addresses the stale-context problem without changing what happens for existing users by default. The same-workspace no-op refresh you already have is good and can stay.

Parking as `changes-requested` pending one of those approaches — happy to review again once it's opt-in / non-silent. (Note: this is a maintainer product call, not a code-quality issue — the implementation is fine.)

nesquena-hermes:
Re-reviewed after the `confirm new chat on workspace switch` commit (a37…, pushed 08:07 UTC). Both items from my earlier `changes-requested` review are resolved — this looks ready to me.

## Blocker 1 (undefined function) — fixed
The old code called `syncActiveProjectForWorkspace(...)`, which existed nowhere in the served front-end, so every switch threw `ReferenceError`. The new code replaces it with `_syncWorkspaceSwitchProject`, which delegates to a helper that *does* exist:

```js
// static/panels.js:5882
function _syncWorkspaceSwitchProject(projectId){
  if(typeof _setActiveProjectFilter==='function')_setActiveProjectFilter(projectId||null);
}
```

Confirmed `_setActiveProjectFilter` is defined at `static/sessions.js:1934` and already used across the project-filter UI, so the runtime path is real now.

## Blocker 2 (silent default flip) — fixed with the confirmation approach
Switching workspace mid-conversation no longer unconditionally starts a new chat. `switchToWorkspace` now gates on whether there's actual conversation and prompts:

```js
// static/panels.js
let startNewSession=!S.session;
if(S.session&&_workspaceSwitchHasConversation()){
  startNewSession=await showConfirmDialog({
    title:t('workspace_switch_new_chat_confirm_title'),
    message:t('workspace_switch_new_chat_confirm_message',label),
    confirmLabel:t('workspace_switch_new_chat_confirm'),
    cancelLabel:t('workspace_switch_keep_current'),
    focusCancel:true
  });
}
```

`focusCancel:true` plus the "Keep current" default means an accidental Enter preserves today's behavior — that's exactly the non-silent shape I was asking for. `showConfirmDialog` is defined at `static/ui.js:7009`. The same-workspace early-return no-op refresh is preserved, and the blank-page auto-create path is folded into `_switchWorkspaceWithNewSession` via `newSession(true,{awaitWorkspaceLoad:true,project_id:null})`.

## Test now exercises the real path
My prior note was that `test_workspace_switch_new_session.py` stubbed the missing function and went green while the browser threw. That's corrected — the harness now `eval`s the real helpers out of the source instead of faking them:

```py
# tests/test_workspace_switch_new_session.py:117-119
eval(extractFunction(src, '_syncWorkspaceSwitchProject'))
eval(extractFunction(src, '_switchWorkspaceInCurrentSession'))
eval(extractFunction(src, '_switchWorkspaceWithNewSession'))
```

`_setActiveProjectFilter` and `showConfirmDialog` are still stubbed, which is legitimate — those genuinely live in `sessions.js`/`ui.js`, so they're external deps of the extracted `panels.js` unit, not the thing under test. Coverage is good across all four branches: confirm→new session (asserts `prev_session_id`/`profile` on `/api/session/new`, lines 177-181), same-workspace no-op (199-201), keep-current→in-place `/api/session/update` (221-231), and blank→new session with no prompt (244-250).

Per the EXECUTION BAN in my triage rules I did not run the suite in this worktree — the review above is static (diff + full-file reads on the PR head). One small thing worth a maintainer glance on ship: `_workspaceSwitchHasConversation()` treats a session containing only `tool` messages as "no conversation," so it would switch in place without prompting. That's a reasonable heuristic (nothing user-authored to lose), just flagging it as an intentional edge. i18n keys are added across all locales. LGTM from the code side.

nesquena-hermes:
Shipped in **exp-v0.52.32** (experimental channel) 🎉

Opt-in "start a new chat when switching workspace" is live (Settings → Preferences, default off). When enabled, switching to a different workspace starts a fresh chat bound to it and leaves your current conversation on its original workspace — avoiding stale cross-workspace context. Off by default, so nothing changes unless you opt in.

Salvage-rebuilt from #5473 and reworked to opt-in. Thanks @ruizanthony! (closes #5473)

PR #5511, opened 2026-07-04
fix: scan providers: in resolve_model_provider() for user-defined provider matching

akay64:
## Thinking Path

- Hermes WebUI model routing relies on `resolve_model_provider()` to map model
  IDs to provider endpoints at request time
- The function scans several sources in order: `custom_providers:`,
  `@provider:model` prefix, `vendor/model` slash format
- It does **not** scan `providers:` from `config.yaml` — the block where users
  define custom providers with explicit model allowlists
- User-configured providers are invisible to resolution; bare model IDs
  silently fall back to the stale `config_provider`, routing through the wrong
  endpoint (e.g. `opencode-go` instead of OpenRouter for a preset model)
- The `custom_providers:` loop at line 2544 already does the exact same scan
  for `custom_providers:` entries — this is the mirror for `providers:`

## What Changed

- Inserted a `providers:` scan into `resolve_model_provider()` in
  `api/config.py`, placed immediately after the `custom_providers:` loop
  and before the `@provider:model` branch
- Iterates `cfg.get('providers', {})`, checks each entry's `models:` list
  or dict for an exact case-insensitive match
- On hit, returns the provider slug and its configured `base_url`
- ~18 lines, no new imports, no API changes, no deps

## Why It Matters

Users who define providers under `config.yaml`'s `providers:` section with
explicit `models:` lists — OpenRouter presets, Ollama model allowlists,
local proxy endpoints — expect those models to be recognized and routed
through the declared provider. Without this scan, selecting such a model
in the chat composer routes the request through whatever `model.provider`
happens to be set to, silently using the wrong endpoint and credentials.

The bug is particularly visible with `@preset/` model IDs (OpenRouter
presets) because the model ID contains a `/`, triggering the slash-branch
heuristics that bypass all provider-matching logic entirely.

## Design Decision — No gating on active provider

The existing `custom_providers:` loop is gated by `_skip_custom_providers` —
a guard that prevents `custom_providers` entries from intercepting models
meant for built-in providers. This makes sense for `custom_providers:` because
those entries are auto-generated (from onboarding flows, from "Custom
endpoint" setup) and are not always explicit user intent.

The `providers:` scan intentionally does **not** replicate this gating.
`providers:` entries in `config.yaml` are explicit user declarations — if
the user listed a model under `providers.openrouter-presets.models`, they
want it routed through that provider. Gating would defeat the purpose. This
matches the precedence: `providers:` is explicit config, `custom_providers:`
is auto-generated convenience.

## Verification

**Full test suite** (`./scripts/test.sh`): 11,716 passed, 95 skipped, 5 pre-existing failures — zero regressions. The 5 failures are unrelated to this change (missing `hermes_cli` modules in test env, Nous picker static entries, TLS startup message format).

**Targeted `resolve_model_provider` tests** (99 tests in `test_issue1855_resolve_model_provider_fast_path.py`, `test_provider_mismatch.py`, `test_openai_api_provider_alias.py`): 99 passed.

**Manual verification**: configured `providers.openrouter-presets` with `models: ['@preset/deepseek-v4-pro-offical']`, selected model in the chat composer — requests correctly routed through OpenRouter's `base_url` and `key_env` instead of the stale `config_provider`.

## Risks / Follow-ups

- **None.** The insertion site sits after `custom_providers:` (unchanged)
  and before all heuristics (`@provider:model`, slash, fallback — all
  unchanged). An empty `providers:` dict is a no-op.
- If two `providers:` entries both declare the same model ID, the first
  match in dict iteration order wins — same behavior as the
  `custom_providers:` loop above it. In practice the model picker uses
  `@provider:model` to disambiguate, so bare-name collisions shouldn't
  reach this function.

## Model Used

- deepseek-v4-pro, provider: opencode

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -2563,6 +2563,30 @@ def resolve_model_provider(model_id: str) -> tuple:
                 provider_hint = _custom_provider_slug_from_name(entry_name)
                 return model_id, provider_hint, entry_base_url or None
 
+    # Check user-defined providers (config.yaml → providers:).
+    # Mirrors the custom_providers scan above — exact match against each
+    # entry's declared models list (case-sensitive to match custom_providers).
+    providers_cfg = cfg.get('providers', {})
+    if isinstance(providers_cfg, dict):
+        target = model_id.strip()
+        for slug, pdef in providers_cfg.items():
+            if not isinstance(pdef, dict):
+                continue
+            p_models = pdef.get('models')
+            if isinstance(p_models, list):
+                for m in p_models:
+                    mid = str(m.get('id') or '') if isinstance(m, dict) else str(m or '')
+                    if not mid:
+                        continue
+                    if mid.strip() == target:
+                        p_base_url = str(pdef.get('base_url') or '').strip()
+                        return model_id, slug, p_base_url or None
+            elif isinstance(p_models, dict):
+                for mid in p_models:
+                    if isinstance(mid, str) and mid.strip() == target:
+                        p_base_url = str(pdef.get('base_url') or '').strip()
+                        return model_id, slug, p_base_url or None
+
     # @provider:model format — explicit provider hint from the dropdown.
     # Route through that provider directly (resolve_runtime_provider will
     # resolve credentials in streaming.py).


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (the generic providers: scan hijacks routing via Copilot's settings-map — 1 CORE)

**Certified head:** `sha:87404d76` (rebased onto current master, `git apply` clean) · **PR:** #5511 · _akay64, fix: scan providers: in resolve_model_provider() for user-defined provider matching_
**Verdict:** The scan is well-guarded and correctly placed for genuine user-defined provider allowlists, but it iterates ALL `providers:` entries including `providers.copilot.models` — which the codebase explicitly documents (config.py:6925) is a per-model SETTINGS map, NOT a routable allowlist. So a Copilot per-model settings entry hijacks that model's routing away from its real provider. Real regression; 0 tests in the PR.

### What I ran (rebased worktree `/tmp/wt-rebase-5511`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 1 CORE (Copilot settings-map hijack); Codex verified end-to-end, I confirmed the docs |
| Full pytest suite | 2 failed / 11918 passed — both non-defects (nous env flake + `test_issue4536` isolation flake) |
| resolve_model_provider tests | ✅ 218 passed — but NONE exercise the Copilot-settings-map case (the PR adds 0 tests) |

### Findings
**⛔ CORE (I + Codex CONFIRMED) — `providers.copilot.models` per-model settings hijack routing** (`api/config.py:2572` scan vs the documented exception at :6925): the new scan iterates every `providers:` entry and treats each `models` key/id as a routable model→provider mapping. But config.py:6925-6929 explicitly documents "Copilot is the exception: it uses `providers.copilot.models` as a per-model **settings map** (reasoning_effort, limits, etc.)" — not an allowlist. Codex verified: with `model.provider=openai` + `providers.copilot.models.gpt-5`, `resolve_model_provider("gpt-5")` now returns `('gpt-5', 'copilot', None)` instead of staying on OpenAI, and streaming passes that provider into runtime resolution (streaming.py:7476). So any user who has Copilot per-model settings gets their matching models silently routed to Copilot. **Fix (Codex-exact, matches the documented exception):** skip Copilot in the scan — `if _canonicalise_provider_id(slug) == "copilot": continue` — and add a regression test (active OpenAI + `providers.copilot.models.<id>` preserves OpenAI routing).

**✅ Otherwise sound:** exact case-sensitive match (no fuzzy false-match), handles list-str/list-dict/dict shapes, non-dict `pdef` skipped, correctly placed after the `custom_providers` early-return (2564) and before the `@provider:`/slash heuristics (2590), returns the right `(model_id, slug, base_url)` triple. 218 resolve_model_provider tests still pass (existing paths intact). The scan is the right idea for real user-defined provider allowlists — it just must honor the Copilot settings-map exception.

**RED — gate-fail/changes-requested (1 CORE + add tests): exclude Copilot from the `providers:` scan (`if _canonicalise_provider_id(slug) == "copilot": continue`, mirroring the documented settings-map exception at config.py:6925) so a Copilot per-model settings entry can't hijack routing; and add regression tests (the PR ships 0 — cover: user-defined provider allowlist matches → routes to it; active OpenAI + `providers.copilot.models` → stays OpenAI; each model-list shape).** The scan is otherwise well-built and correctly placed. concept 4/5 (real user-defined-provider routing gap, right approach, one documented-exception miss). Author @akay64 (fresh). crit=3. (Gate value: a "scan all providers: entries" generalization looks uniform, but one provider — Copilot — uses `models` for settings, not routing; the fix and the doc were 4000 lines apart, so the author couldn't see the collision. Always check whether a config key a new generic scan consumes has a per-provider semantic exception elsewhere.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; the Copilot settings-map hijack confirmed (scan at 2569-2588 has no copilot exclusion, vs the documented per-model-settings exception at config.py:6925; Codex reproduced gpt-5 → copilot), scan otherwise well-guarded + correctly placed, 218 existing routing tests pass. Cert valid for sha:87404d76._

nesquena-hermes:
Shipped in **v0.51.853** 🎉 — merged via release `stage-5511` (commit `2981f711`).

Thanks @akay64 — routing user-defined `providers:` model allowlists through `resolve_model_provider()` is the right fix, and the scan itself was well-built (exact match, correct placement, all model-list shapes handled).

During the gate (Codex, run to convergence over 5 rounds + Opus) we hardened the surrounding provider-resolution ownership model so the new scan can't collide with existing invariants — all preserving your authorship via `Co-authored-by`:
1. Copilot excluded (`providers.copilot.models` is a per-model settings map, not a routing allowlist)
2. The scan honors the active/default ownership guard (an active-owned model can't be hijacked by another provider's entry)
3. The ownership set is built from the **canonical** active provider slug (fixed a latent bug where `z-ai`/aliases got an empty ownership set)
4. `_canonicalise_provider_id` now accepts alias targets in `_PROVIDER_MODELS` too (fixed the `google-gemini`→`gemini` class)
5. An active provider owns the models in its own `providers.<active>.models` allowlist

5 discriminating regression tests added (each verified to fail without its fix). Findings 3-5 were pre-existing latent bugs your scan surfaced — fixed at root. Full suite green (11965 passed), CI green across all 15 shards. Closing as shipped — thanks for the contribution!

Issue #5515, opened 2026-07-04
bug(chat): transcript randomly scrolls up during normal use (persists on v0.51.845)


nesquena-hermes:
## Summary

A user reports the chat transcript sometimes scrolls **up** on its own during normal use, requiring them to scroll back down. It's intermittent ("sometimes it seems to scroll up randomly in my chats and i have to scroll down again"). A fix attempt shipped, but the reporter confirms it **still persists on v0.51.845**.

This is filed as a separate issue from the composer-growth scroll bug (which is deterministic and tied to typing in a multi-row composer). This one is intermittent and not obviously tied to the composer.

## What we know

- Symptom: transcript jumps upward unexpectedly; user has to scroll down again.
- Intermittent / "random" — no reliable repro from the report yet.
- Still reproduces on **v0.51.845** after a recent fix attempt.

## Suspected area

The scroll-pin / auto-scroll state machine and anchor handling around streaming and re-render:

- `_nearBottomCount` / `_scrollPinned` pin-state logic (`static/ui.js`) that decides whether to follow the tail.
- `overflow-anchor` behavior differs by device: `static/style.css` sets `.messages{overflow-anchor:auto}` by default but `@media (hover:hover) and (pointer:fine){.messages{overflow-anchor:none}}` — i.e. on touch devices `overflow-anchor:auto` can shift scroll when content above changes height.
- `preserveScroll` / `_captureMessageScrollSnapshot` re-render path.

This is the same broad class as several previously-closed scroll-jump issues (#4295, #4720, #4856, #3545, #4207, etc.) — but those were closed and this reporter says it still happens on a current build, so it may be an uncovered sibling path rather than a dup.

## Need from reporter (to pin the root cause)

- Platform/browser (desktop browser, macOS desktop app, mobile Safari, Android app?) and touch vs mouse.
- **When** does it jump — while a response is streaming, right when a response completes, while idle-reading an old conversation, or when switching sessions?
- Does it jump all the way to the **top** (first message) or just up a screenful/a few rows?
- Roughly how long/how many messages is the conversation when it happens?
- Does a hard reload change the behavior?

Labeling `needinfo` until we get a reliable repro; the answers map to specific candidate paths (streaming-follow vs completion-scroll vs overflow-anchor vs virtualization).

nesquena-hermes:
Fixed in **v0.51.851** via #5516. 🎉

[...]

PR #5516, opened 2026-07-04
fix(chat): re-pin transcript when the composer grows (#5514, #5515)

### LINKED ISSUE #5514: bug(chat): transcript scrolls up when the composer grows while pinned to bottom
## Summary

When the chat is scrolled to the bottom and the composer is taller than one row, typing in the composer scrolls the transcript **up** — one row per row of composer growth (up to the composer's max height). The user has to scroll back down. Originally reported against the macOS desktop app (`hermes-webui/hermes-swift-mac#87`), but the reporter confirmed it reproduces in the plain WebUI too, so the fix belongs here.

## Steps to reproduce

1. Open a conversation with enough history that the transcript is scrollable.
2. Scroll all the way to the bottom (pinned to the latest message).
3. Grow the composer past one row — type multiple lines, `Shift+Enter`, or paste a multi-line block (WisprFlow paste reproduces it).
4. As soon as the composer grows, the transcript scrolls up by roughly one row per row the composer gained.

Does **not** reproduce when the composer stays at the default single row, or when you're not pinned to the very bottom.

Reported on macOS 26.5.1 + WebUI session, and confirmed in the browser WebUI (not just the desktop shell).

## Root cause (traced)

The chat transcript `#messages` and the composer `#composerWrap` are flex siblings in a column:

- `static/style.css` — `.messages{flex:1;overflow-y:auto; ... min-height:0; ...}`
- `static/index.html` — `#messages` (transcript) and `#composerWrap` (composer) are siblings.

When the composer grows, `autoResize()` runs and enlarges the textarea:

```js
// static/messages.js ~L6178
function autoResize(){
  ...
  const el=$('msg');
  el.style.height='auto';
  el.style.height=Math.min(el.scrollHeight,200)+'px';
  updateSendBtn();
}
```

`autoResize()` (invoked from the composer `input` listener in `static/boot.js` ~L1886 via `scheduleComposerAutoResize()`) grows the composer, which **shrinks the `flex:1` `.messages` viewport by the same delta**. When the reader was pinned at the bottom (`scrollTop === scrollHeight - clientHeight`), shrinking `clientHeight` by Δ raises the max scrollTop by Δ but leaves `scrollTop` unchanged — so the reader is now Δ pixels above the bottom and the transcript appears to scroll up. `autoResize()` only resizes the textarea and calls `updateSendBtn()`; it never re-pins the transcript to the bottom to compensate for the viewport shrink. This matches the report exactly ("scrolls up 1 row, per row of the composer").

## Suggested fix

In `autoResize()` (or right after it in the input path), if the reader was at/near the bottom before the resize, re-pin after the height change — capture the at-bottom / `_scrollPinned` state before setting the new height and call the existing bottom-scroll helper (`scrollToBottom()` / the near-bottom pin path) after, so a bottom-pinned transcript stays pinned as the composer grows. Guard it on the pin state so it doesn't yank readers who were deliberately scrolled up.

Add a regression test asserting: bottom-pinned + multi-row composer growth keeps the transcript pinned to the bottom.

## Cross-reference

- Originally filed as `hermes-webui/hermes-swift-mac#87` (reporter confirmed WebUI repro).

nesquena-hermes:
## Summary

Fixes **#5514** (deterministic) and the composer-growth cause of **#5515** (intermittent "random scroll up"): when the chat is pinned to the bottom and the composer grows, the transcript scrolls up.

## Root cause (reproduced live)

`#messages` (transcript, `flex:1`) and the composer are flex siblings in a column. When the composer grows — multi-line typing, `Shift+Enter`, or a multi-line / WisprFlow paste — the `flex:1` transcript viewport **shrinks by the same delta**. A reader pinned to the bottom (`scrollTop === scrollHeight - clientHeight`) keeps the same `scrollTop`, but max-scrollTop just rose by Δ, so they're stranded Δpx above the bottom — "scrolls up one row per composer row." `autoResize()` resized the textarea + called `updateSendBtn()` but never re-pinned.

Reproduced on a 40-message seeded session: composer grew to 176px → viewport shrank 745→613 → transcript stranded **132px** above bottom, while `_scrollPinned` stayed `true` (the pin *state* was right; nothing acted on it).

## Fix

New `_repinMessagesAfterComposerResize()` (`static/ui.js`) re-pins the transcript to the bottom **only when the reader is genuinely still pinned** — it honors the existing sticky-unpin model (`_messageUserUnpinned` / `_scrollPinned`), so a reader who scrolled up to read history is never yanked back down. It reuses the canonical `_setMessageScrollToBottom()` primitive. Wired at two seams:

1. **`autoResize()`** — after the height write (covers the typed-input path).
2. **A `ResizeObserver` on the `#msg` textarea** — re-pins on **grow only** (a shrink enlarges the viewport and can't strand anyone). This catches composer height changes that don't route through the input→autoResize seam: paste, draft restore, programmatic value set, reflow — i.e. the paths that read as a "random" jump in **#5515**.

## Verified live (CDP, seeded 40-message session)

- **Pinned reader, composer grows 4 rows →** `bottomDist` stays **0** (was **132** before the fix). Latest message stays visible above the composer (vision-checked).
- **Unpinned reader (scrolled up 2000px), composer grows →** `scrollTop` unchanged (delta **0**), not yanked to bottom.

## Tests

`tests/test_issue5514_composer_grow_scroll_pin.py` — 3 static-wiring assertions + 3 node-`vm` behavioral tests (pinned→re-pinned, unpinned→untouched, already-at-bottom→no-op). All pass; **524** scroll/composer regression tests across 39 files still green.

## On #5515

This resolves the **composer-growth cause** of the "random scroll up" reports (the reporter's WisprFlow-paste symptom points squarely at it, and it's the same trigger as #5514 experienced during normal typing). If a purely **non-composer** intermittent jump still persists after this ships, the remaining suspect is touch-device `overflow-anchor:auto` on content-above height change — that path has **no repro yet** and stays tracked as `needinfo` on #5515. I'd close #5514 outright on merge and ask the #5515 reporter to confirm against the new build before closing that one.

Closes #5514.

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -1918,27 +1918,25 @@ $('msg').addEventListener('input',()=>{
 });
 // #5514/#5515: re-pin the transcript on ANY composer height change, not only the
 // ones that route through the input->autoResize path. A multi-line paste
-// (WisprFlow), a draft restore, a programmatic value set, or a font/reflow can
-// all grow the composer and shrink the flex:1 transcript viewport, stranding a
-// pinned reader above the bottom (reads as a "random" upward jump — #5515). A
-// ResizeObserver on the textarea catches every case at one seam. The re-pin
-// itself is guarded (only fires when genuinely pinned), so this never fights a
-// reader who scrolled away. First callback fires on observe (initial size) — the
-// guard makes that a cheap no-op.
+// (WisprFlow), a draft restore, an attachment tray / selection-chip appearing, a
+// programmatic value set, or a font/reflow can all grow the composer and shrink
+// the flex:1 transcript viewport, stranding a pinned reader above the bottom
+// (reads as a "random" upward jump — #5515). Observe the whole #composerWrap
+// (not just #msg) so tray/chip growth is covered too, at one seam. The re-pin is
+// guarded (only fires when genuinely pinned), so it never fights a reader who
+// scrolled away. First callback fires on observe (initial size) — the guard
+// makes that a cheap no-op.
 (()=>{
-  const _c=$('msg');
-  if(!_c || typeof ResizeObserver!=='function' || typeof _repinMessagesAfterComposerResize!=='function') return;
-  let _lastComposerH=_c.offsetHeight;
+  const _cw=$('composerWrap')||$('msg');
+  if(!_cw || typeof ResizeObserver!=='function' || typeof _repinMessagesAfterComposerResize!=='function') return;
+  let _lastComposerH=_cw.offsetHeight;
   const _ro=new ResizeObserver(()=>{
-    const h=_c.offsetHeight;
-    if(h===_lastComposerH) return;   // width-only / no-op resize: ignore
-    const grew=h>_lastComposerH;
+    const h=_cw.offsetHeight;
+    if(h<=_lastComposerH){_lastComposerH=h;return;}   // shrink/no-op: enlarges the viewport, can't strand
     _lastComposerH=h;
-    // Only a GROW shrinks the transcript viewport; a shrink enlarges it and can't
-    // strand a pinned reader. Re-pin on grow.
-    if(grew) _repinMessagesAfterComposerResize();
+    _repinMessagesAfterComposerResize();               // grow: re-pin the pinned reader
   });
-  try{ _ro.observe(_c); }catch(_){ }
+  try{ _ro.observe(_cw); }catch(_){ }
 })();
 // Track IME composition for East Asian input. Safari fires the committing
 // keydown AFTER compositionend with isComposing=false, so we also keep a
--- a/static/boot.js
+++ b/static/boot.js
@@ -1916,6 +1916,30 @@ $('msg').addEventListener('input',()=>{
     hideCmdDropdown();
   }
 });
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
Shipped in **v0.51.851** 🎉 — merged via release `stage-5516` (commit `46b630db`).

Fixes the composer-grow scroll strand (#5514 deterministic, #5515 intermittent). The transcript now re-pins to the bottom when the composer grows, guarded by the existing sticky-unpin model so a reader who scrolled up to read history is never yanked down.

Cleared the full authoritative gate:
- **Codex: SAFE TO SHIP** (verified no ResizeObserver feedback loop — the re-pin writes only the transcript's scrollTop, which can't resize the composer)
- **Opus advise: SHIP** (traced every scroll-away path; no scroll-fight, sticky-unpin honored, mid-stream follow harmonious; verified live in both directions)
- **Full suite green** (+6 new regression tests: pinned→re-pinned, unpinned→untouched, at-bottom→no-op)
- **CI: all 15 shards + lint + browser-smoke green**

Also folded in a hot-path optimization (only re-pin when the composer actually grew) and widened the observer to `#composerWrap` to catch paste / draft-restore / tray / chip growth.

PR #5529, opened 2026-07-04
fix(ux): add loading feedback and proper styling to session-list Retry button (#5501)

### LINKED ISSUE #5501: Session-list load-error "Retry" button gives no feedback (looks broken) and is unstyled/ugly
## Summary

The session-list load-error state (shown when `/api/sessions` times out or fails) has two problems, both visible in Rod's report:

1. **The "Retry" button gives no feedback and appears broken.**
2. **The whole error block — and the button especially — is unstyled and ugly** (raw browser-default gray button, dimmed to look disabled).

Reported by **@rodifer** in #report-bugs ("Button doesn't work. But more importantly it's ugly 😂 … It did go away actually").

## Screenshot

Washed-out gray "Retry" button under "Session list is taking longer than expected. The backend may still be scanning a very large session history." — the button looks disabled.

## Root cause

The banner + button were added in #4167 (`static/sessions.js`, current `v0.51.846`).

**(1) No retry feedback → looks broken.** The click handler works, but there's no loading state:

```js
// static/sessions.js ~line 6535
const retry=document.createElement('button');
retry.type='button';
retry.textContent='Retry';
retry.onclick=(e)=>{e.stopPropagation();_sessionListLoadError=null;void renderSessionList({deferWhileInteracting:false});};
```

Clicking clears the error flag and kicks off a fresh `renderSessionList()`, but the render is async and paints nothing new until the fetch resolves. On the exact scenario that triggered the error — a very large session history — that fetch can take up to the 90 s boot timeout (`_SESSION_LIST_BOOT_TIMEOUT_MS = 90000`). During that window the stale error banner stays on screen unchanged, so the user perceives "nothing happened / button is dead." Rod's "It did go away actually" confirms the fetch *does* eventually succeed — the problem is purely the missing feedback.

**(2) Unstyled / ugly.** The error container class `session-list-error` has **no CSS rule at all**, and the Retry `<button>` has **no class**, so it renders as a raw browser-default gray button. Its parent `.session-empty-note` applies `opacity:.7` (`static/style.css` ~line 5526), which dims the entire block including the button — making it look grayed-out/disabled, exactly as in the screenshot.

```css
/* static/style.css ~5526 — the only rule that touches the block */
.session-empty-note{padding:20px 14px;color:var(--muted);font-size:12px;text-align:center;opacity:.7;}
/* .session-list-error — no rule; the Retry button — no class */
```

## Suggested fix (shape only — PR handled separately)

- Give the Retry button a real class matching the app's button styling (e.g. reuse an existing pill/secondary-button style so it reads as an interactive control, not a disabled default control), and lift it out of the parent's `opacity:.7` so it isn't dimmed.
- Add a `.session-list-error` rule so the block has intentional spacing/typography rather than inheriting only the muted empty-note look.
- On retry, show immediate feedback: disable the button + swap label to "Retrying…" (or show the existing skeleton/spinner) for the duration of the in-flight fetch, so a slow large-history reload doesn't look like a no-op. Re-enable/restore on resolve or on the next error.

## Repro

1. Have a very large session history (or throttle/slow `/api/sessions`) so the sidebar load hits the timeout.
2. Sidebar shows "Session list is taking longer than expected." with the "Retry" button.
3. Click Retry → no visible change for up to ~90 s (looks broken); button itself renders as an unstyled, dimmed gray control.

## Notes

- Not a duplicate — searched open/closed issues + PRs for retry/session-list-error, nothing covers this.
- Introduced in #4167 (2026-06-16). Current version at time of report: `v0.51.846`.

nankingjing:
Closes #5501

## Summary

The session-list load-error Retry button had two issues (reported by @rodifer):

1. **No retry feedback** — clicking looked broken because the async render painted nothing new until the fetch resolved (up to 90 s). The button appeared dead.
2. **Unstyled / dimmed** — raw browser-default gray button inheriting opacity:.7 from the parent .session-empty-note, so it looked grayed-out and disabled.

## What changed

- **sessions.js**: on retry click, disable the button + show "Retrying…" label immediately, with a guard against double-clicks.
- **style.css**: reset opacity on the error container; add .session-list-error-retry with accent-brand styling, hover brighten, and a disabled state that visually confirms the button is processing.

## Before / After (screenshot)

Before: unstyled gray Retry button, dimmed, no feedback on click.
After: styled pill button matching app theme, shows "Retrying…" with disabled state during the in-flight fetch.

## Diff size

+20/-1 across 2 files (sessions.js + style.css). No new deps, no new config.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -6698,8 +6698,16 @@ function renderSessionListFromCache(){
     }
     const retry=document.createElement('button');
     retry.type='button';
+    retry.className='session-list-error-retry';
     retry.textContent='Retry';
-    retry.onclick=(e)=>{e.stopPropagation();_sessionListLoadError=null;void renderSessionList({deferWhileInteracting:false});};
+    retry.onclick=(e)=>{
+      e.stopPropagation();
+      if(retry.disabled)return;
+      retry.disabled=true;
+      retry.textContent='Retrying…';
+      _sessionListLoadError=null;
+      void renderSessionList({deferWhileInteracting:false});
+    };
     note.appendChild(retry);
     list.appendChild(note);
   }
--- a/static/style.css
+++ b/static/style.css
@@ -5525,6 +5525,17 @@ main.main > #mainPlugin{display:none;}
 .session-source-tab:hover{background:rgba(255,255,255,.08);color:var(--text);}
 .session-source-tab.active{background:var(--accent-bg);color:var(--accent-text);border-color:var(--accent-bg);}
 .session-empty-note{padding:20px 14px;color:var(--muted);font-size:12px;text-align:center;opacity:.7;}
+.session-list-error{opacity:1;}
+.session-list-error-retry{
+  display:inline-block;margin-top:14px;padding:8px 28px;
+  font-size:13px;font-weight:500;line-height:1.4;
+  color:#fff;background:var(--accent-bg);border:1px solid var(--accent-bg);
+  border-radius:6px;cursor:pointer;transition:opacity .15s;
+}
+.session-list-error-retry:hover{filter:brightness(1.15);}
+.session-list-error-retry:disabled{
+  opacity:.55;cursor:not-allowed;filter:none;
+}
 .project-bar{display:flex;gap:4px;padding:4px 10px 8px;flex-wrap:wrap;align-items:center;flex-shrink:0;}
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Superseded by #5505 (shipped in v0.51.878), which fixes the same session-list Retry button issue (#5501) with the pending-state feedback + styling you described here, plus folded-in a11y (live region, focus retention) and a stuck-retry guard. Thanks @nankingjing for the report and the parallel fix — the shipped version credits the same root issue. Closing this as the duplicate; really appreciate the contribution.

PR #5531, opened 2026-07-04
feat(cors): opt-in credentialed cross-origin access via ALLOWED_ORIGINS

mo7al876any:
Opt-in credentialed cross-origin access: a browser front-end served from an allowlisted origin can call the API with cookies. **Reworked onto current master** now that the #5534 CORS-preflight fix has merged (and was refactored into `api/routes.py`'s `apply_cors_preflight_headers`, keeping `server.py` thin) — this PR is no longer stacked on anything and contains only the credentialed delta (4 files).

## What (all gated on `HERMES_WEBUI_ALLOWED_ORIGINS`; default deployments unchanged)

- **`api/routes.py`**
  - `cors_credentialed_origin()` — echoes an Origin only when it's in the explicit allowlist (never same-origin, never `*`).
  - `_check_same_origin_browser_request()` — allowlisted origins clear the `Sec-Fetch-Site: cross-site` gate (modern browsers send it on every genuine cross-origin request), so cross-origin writes/preflights aren't 403'd. The CSRF token check still applies.
  - `apply_cors_preflight_headers()` — for allowlisted origins also advertises the CSRF token headers (`X-Hermes-CSRF-Token`/`X-CSRF-Token`, or the browser strips them on writes), `Access-Control-Max-Age`, and — HTTPS only — `Access-Control-Allow-Credentials`; emits `Vary: Origin` whenever an allowlist is configured (shared-cache safety).
  - `apply_cors_actual_response_headers()` — credentialed CORS on real responses, called from a thin `end_headers` hook so `server.py` stays a dispatcher (mirrors the preflight helper).
- **`server.py`** — `end_headers` calls the actual-response helper for non-OPTIONS responses.
- **`api/auth.py`** — `set_auth_cookie` relaxes to `SameSite=None` only under allowlist **and** HTTPS (`SameSite=None` requires `Secure`; else `Lax`). Also hardens `_is_secure_context` against a handler with no `.request` attribute (defensive; no behavior change for real handlers).

## Notes
- Prior history: this feature reached Greptile 5/5 across four review rounds (Sec-Fetch-Site, CSRF headers, HTTPS-gated credentials, cache-safe `Vary`) before #5534 merged. It's re-implemented here in the maintainer's refactored structure, with the same behavior.
- 20 tests exercise the header emitters directly plus the `SameSite` matrix and `Sec-Fetch-Site` ordering; master's own preflight test still passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5530, opened 2026-07-04
Review fixes: CORS preflight hardening, i18n completion, env docs, Windows test compat

mo7al876any:
## What changed

Five focused fixes from a code-review pass, ordered by importance:

- **fix(security): CORS preflight echoes only same-origin or allowlisted origins** (bb37a37) — the preflight handler previously reflected arbitrary `Origin` headers back in `Access-Control-Allow-Origin`; it now answers only for same-origin requests or origins on the configured allowlist.
- **i18n: translate all 664 remaining TODO-translate placeholders** (bae81f0) — completes locale coverage; no more `TODO-translate` strings shipped to users.
- **docs(env): document the `HERMES_WEBUI_*` variables the code reads but no template listed** (3aa4f88) — closes the gap between configuration the code consumes and what the env template documents.
- **docs: refresh stale version/test-count/locale numbers** (3956e8) — README/docs numbers now match reality.
- **test: read static assets as UTF-8 so locale/render tests run on Windows** (c153f4e) — test files opened static assets with the platform default encoding (cp1252 on Windows), breaking locale and SVG/audio/video rendering tests; they now read explicitly as UTF-8.

## Why

The CORS reflection was the one real security issue found in review; the rest remove long-standing friction: untranslated UI strings, undocumented configuration, stale docs, and a Windows-only test-encoding failure class.

## Notes for reviewers

- The large negative diffstat (−5,055 lines) is dominated by the i18n commit replacing placeholder blocks with final translations.
- The Windows encoding fix touches many test files but is mechanical (`open(...)` → `open(..., encoding="utf-8")`).
- CI is the source of truth for the test suite; the local Windows environment has a known pre-existing failure set unrelated to these changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5534, opened 2026-07-04
fix(security): CORS preflight echoes only same-origin or allowlisted origins

mo7al876any:
Split out of #5530 per the gate-certifier's request — this is the security-boundary change on its own so it can get a focused security review.

## What
`do_OPTIONS` previously answered every CORS preflight with `Access-Control-Allow-Origin: *`, advertising broader cross-origin access than a real request is granted — on a password-less deployment that lets any site read authenticated responses.

It now echoes the request Origin only when `_check_same_origin_browser_request(handler)` approves it — reusing the CSRF gate's exact same-origin / `HERMES_WEBUI_ALLOWED_ORIGINS` policy, so the preflight can never advertise wider access than an actual request. Never emits `*`.

## Tests
`tests/test_cors_preflight_allowlist.py` — 10 cases: same-origin echoed, cross-origin rejected, `Sec-Fetch-Site: cross-site` rejected, allowlist opt-in/out, `X-Forwarded-Host` trust flag, and a source-level regression guard against the wildcard returning.

Default deployments (no allowlist) are unaffected. ~42 lines of security-relevant change; the credentialed cross-origin follow-up (#5531) stacks on this.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5127,6 +5127,22 @@ def _check_same_origin_browser_request(handler, *, require_provenance: bool = Fa
     return True
 
 
+def preflight_allow_origin(handler) -> str:
+    """CORS preflight: return the Origin value to echo back, or '' to omit.
+
+    Echo the request Origin only when it is same-origin or explicitly
+    allowlisted via HERMES_WEBUI_ALLOWED_ORIGINS — the exact policy the CSRF
+    gate enforces for real requests. Reuses _check_same_origin_browser_request
+    so the preflight can never advertise wider access (`*`) than an actual
+    request would be granted. A wildcard here would let any site read
+    authenticated responses on a deployment with no password set.
+    """
+    origin = handler.headers.get("Origin", "").strip()
+    if not origin:
+        return ""
+    return origin if _check_same_origin_browser_request(handler) else ""
+
+
 def _csrf_exempt_path(path: str) -> bool:
     """Paths that cannot or must not carry a session CSRF token."""
     return path in {
--- a/server.py
+++ b/server.py
@@ -109,7 +109,14 @@ from api.helpers import (
     _CLIENT_DISCONNECT_ERRORS,
 )
 from api.profiles import set_request_profile, clear_request_profile
-from api.routes import handle_delete, handle_get, handle_patch, handle_post, handle_put
+from api.routes import (
+    handle_delete,
+    handle_get,
+    handle_patch,
+    handle_post,
+    handle_put,
+    preflight_allow_origin,
+)
 from api.startup import auto_install_agent_deps, fix_credential_permissions
 from api.updates import WEBUI_VERSION
 from api.crash_visibility import install_crash_visibility
@@ -435,12 +442,21 @@ class Handler(BaseHTTPRequestHandler):
         self._handle_write(handle_patch)
 
     def do_OPTIONS(self) -> None:
-        """Handle CORS preflight requests."""
+        """Handle CORS preflight requests.
+
+        Echo the request Origin only when it is same-origin or allowlisted
+        via HERMES_WEBUI_ALLOWED_ORIGINS (same policy as the CSRF gate) —
+        never advertise `*`. A disallowed origin gets a 200 with no CORS
+        headers, which the browser treats as a preflight denial.
+        """
         self._req_t0 = time.time()
+        allow_origin = preflight_allow_origin(self)
         self.send_response(200)
-        self.send_header("Access-Control-Allow-Origin", "*")
-        self.send_header("Access-Control-Allow-Methods", "GET, POST, PUT, PATCH, DELETE, OPTIONS")
-        self.send_header("Access-Control-Allow-Headers", "Content-Type, Authorization")
+        if allow_origin:
+            self.send_header("Access-Control-Allow-Origin", allow_origin)
+            self.send_header("Vary", "Origin")
+            self.send_header("Access-Control-Allow-Methods", "GET, POST, PUT, PATCH, DELETE, OPTIONS")
+            self.send_header("Access-Control-Allow-Headers", "Content-Type, Authorization")
         self.end_headers()
 
     def do_DELETE(self) -> None:
--- a/tests/test_cors_preflight_allowlist.py
+++ b/tests/test_cors_preflight_allowlist.py
@@ -1,0 +1,96 @@
+"""CORS preflight must not advertise wider access than the CSRF gate permits.
+
+server.py's do_OPTIONS previously answered every preflight with
+``Access-Control-Allow-Origin: *``. It now echoes the request Origin only
+when it is same-origin or explicitly allowlisted via
+HERMES_WEBUI_ALLOWED_ORIGINS — reusing _check_same_origin_browser_request,
+the same policy enforced for real requests.
+"""
+
+from pathlib import Path
+from types import SimpleNamespace
+
+from api.routes import preflight_allow_origin
+
+ROOT = Path(__file__).resolve().parent.parent
+
+
+def _preflight(headers):
+    return preflight_allow_origin(SimpleNamespace(headers=headers))
+
+
+class TestPreflightAllowOrigin:
+    def test_no_origin_header_omits_cors(self):
+        """Non-browser OPTIONS (no Origin) gets no Allow-Origin header."""
+        assert _preflight({}) == ""
+
+    def test_same_origin_echoed(self):
+        assert _preflight({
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
Pulled the branch and read the diff against `master`. The security boundary holds — a few notes from verifying the pieces the automated summary didn't drill into.

## The delegation is airtight for the wildcard removal

`preflight_allow_origin` (`api/routes.py:5130`) returns the echoed origin only when the CSRF gate approves, else `""`:

```python
origin = handler.headers.get("Origin", "").strip()
if not origin:
    return ""
return origin if _check_same_origin_browser_request(handler) else ""
```

And `server.py` `do_OPTIONS` gates *all* CORS headers on that non-empty return, not just `Allow-Origin`:

```python
allow_origin = preflight_allow_origin(self)
self.send_response(200)
if allow_origin:
    self.send_header("Access-Control-Allow-Origin", allow_origin)
    self.send_header("Vary", "Origin")
    self.send_header("Access-Control-Allow-Methods", ...)
    self.send_header("Access-Control-Allow-Headers", ...)
```

Good call folding `Allow-Methods`/`Allow-Headers` inside the `if` too — a rejected origin now gets a bare 200 with zero CORS advertising, which is the correct denial shape. `Vary: Origin` is right given the response body now depends on the request Origin.

## require_provenance asymmetry — checked, and it's fine

`preflight_allow_origin` calls `_check_same_origin_browser_request(handler)` with the default `require_provenance=False`, whereas the strict write path uses `require_provenance=True` (`api/routes.py:5356`). I checked whether that makes the preflight *looser* than the real request it's advertising for. It doesn't in the meaningful direction: on `require_provenance=True` the no-header / `Sec-Fetch-Site: same-origin` cases return a failure reason (`api/routes.py:5086`, `:5094`), but a CORS preflight *always* carries an `Origin` header by spec, so those provenance-less branches aren't reachable here — the `if not origin: return ""` guard short-circuits first. The preflight can only ever echo an origin that passed the same host/scheme/allowlist comparison a real request would. So the preflight never advertises access wider than the actual request is granted, which is exactly the invariant this PR is claiming.

## Credentials: correctly out of scope here

Confirmed this PR does not emit `Access-Control-Allow-Credentials`, so the echoed origin can't be paired with cookies from a cross-origin context. That matters because there *is* a `Set-Cookie` on the profile path (`api/routes.py:14194`), but `build_profile_cookie` (`api/helpers.py:750-753`) sets it `HttpOnly` + `SameSite=Lax`:

```python
cookie[cookie_name]['httponly'] = True
cookie[cookie_name]['samesite'] = 'Lax'
```

So even the credentialed follow-up (#5531) has a `Lax` cookie backstop. Deferring credentialed CORS to that PR is the right split.

## One minor thing

Agree with the greptile note that `Access-Control-Max-Age` is absent — purely a perf detail (extra preflight round-trips), not a correctness gap, and arguably fine to leave until #5531 since the response now varies by origin anyway. No change requested on that front. LGTM from a security-boundary read.

nesquena-hermes:
## 🔬 Gate certification — GREEN (engineering, security-boundary) ✅ · CI-red is brittle-#5513-RFC collateral (not this PR) · → Nathan security sign-off

**Certified head:** `sha:0db04477` (clean rebase, branch `gate-rebase/5534-cors-preflight-allowlist`) · **PR:** #5534 · _mo7al876any, fix(security): CORS preflight echoes only same-origin or allowlisted origins_ · split-child of the (correctly) decomposed #5530/#5490 bundle
**Verdict:** The CORS fix is genuinely well-done and secure — Codex SAFE on the adversarial pass, its own 10/10 tests pass, the security boundary holds. The 2 CI failures are NOT this PR's fault: they're the merged-#5513 RFC's brittle `api/routes.py:NNNN` line-number anchors breaking because this PR's +16 lines shifted routes.py — exactly the regression-magnet I flagged when I RED'd #5513. Route to Nathan for the security-boundary sign-off; the RFC brittleness needs a separate root-cause fix.

### What I ran (rebased worktree `/tmp/wt-rebase-5534`)
| Gate | Result |
|------|--------|
| Codex (adversarial security) | **SAFE TO SHIP** — 0 findings; the boundary holds |
| Full pytest suite | 4 failed / 11991 passed — 2 known env flakes (nous, issue4536) + **2 brittle-#5513-RFC-anchor tests (NOT this PR)** |
| PR's own test | ✅ 10/10 (`test_cors_preflight_allowlist.py`) |

### Findings
**✅ Security fix correct (the whole point):** replaces `Access-Control-Allow-Origin: *` with `preflight_allow_origin(handler)` → returns the Origin ONLY when `_check_same_origin_browser_request(handler)` (same-origin OR `HERMES_WEBUI_ALLOWED_ORIGINS` — the EXACT policy the CSRF gate enforces for real requests), else `""` (no CORS header, browser blocks). Adds `Vary: Origin` (prevents cross-origin preflight cache poisoning); methods/headers only sent inside `if allow_origin`; **no `Allow-Credentials: true`** (no reflected-origin+credentials footgun); the test regression-guards against `"*"` returning. Codex verified preflight can't advertise wider access than a real request (shared `_check_same_origin_browser_request` policy). Clean, exhaustive security fix.

**⚠️ CI-red = brittle-#5513-RFC collateral (I confirmed — NOT a #5534 defect):** `test_issue4812_session_sse_contract_rfc.py::test_rfc_cites_current_global_endpoint_source` + `::test_rfc_run_journal_anchors_land_on_real_source` FAIL only in this worktree — they PASS on clean master. Cause: the RFC pins `/api/sessions/events` at `api/routes.py:12345`; #5534 adds +16 lines at routes.py:5127 (the CORS change), shifting everything below down 16 lines, so anchor 12345 now points at `if runtime_adapter_enabled():`. **This is exactly the line-number-pin brittleness I RED'd on #5513** — which was reworked to "self-validating anchors" and merged anyway, but the anchors are STILL line-based, so ANY routes.py insertion above them breaks the RFC test. It will break every future routes.py PR.

### Recommendation to the next agent / Nathan
**Engineering-GREEN + SECURE — route to Nathan for the security-boundary sign-off (merge from `gate-rebase/5534-cors-preflight-allowlist`, `sha:0db04477`, NOT stale head `9a694e04`).** The CORS fix is exactly right (same-origin-tied, no wildcard, Vary:Origin, no credentials footgun, regression-guarded) and nesquena-hermes already posted a positive "the security boundary holds" review. The ONLY blocker is the merged-#5513 RFC line-anchor test breaking on the routes.py line-shift — not this PR's defect. **Two paths to unblock (recommend the root-cause one):** (A) ROOT-CAUSE (preferred, separate follow-up): de-brittle the #5513 RFC anchors — cite by function/route NAME not `path:NNNN`, so they stop breaking every routes.py PR (this is the fix-root-cause call; the brittleness is now a live regression-magnet on master); (B) mechanical stopgap: bump the RFC's cited line numbers +16 for anchors below routes.py:5127 as part of this PR. Prefer (A) as a fast #5513-follow-up, then #5534 goes green untouched. concept 5/5 (clean, well-tested security fix; and the author correctly split the bundle per the #5530 ask). Author @mo7al876any. crit=4, security → Nathan.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; CORS boundary verified (same-origin-tied via `_check_same_origin_browser_request`, no `*`, Vary:Origin, no Allow-Credentials footgun, regression-guarded), Codex SAFE + 10/10. CI-red = merged-#5513 RFC line-anchor tests breaking on this PR's routes.py +16 line-shift (pass on clean master; the brittleness I flagged on #5513) — NOT a #5534 defect; recommend de-brittling #5513's anchors. Security-boundary → Nathan. Cert valid for sha:0db04477._

nesquena-hermes:
Shipped in **v0.51.876** — thanks @mo7al876any! 🎉

CORS preflight (`OPTIONS`) no longer answers with `Access-Control-Allow-Origin: *`. It now echoes the request origin only when it's same-origin or allowlisted via `HERMES_WEBUI_ALLOWED_ORIGINS` (the exact policy the CSRF gate enforces for real requests), adds `Vary: Origin`, and emits no CORS headers on denial. On a password-less deployment this closes a path where any site could read authenticated responses. Default deployments are unaffected.

Security-gated: Codex SAFE TO SHIP across both rounds (no bypass via spoofed Origin / Sec-Fetch-Site / X-Forwarded-Host / wildcard; preflight can't advertise wider access than a real request). Full suite 12083 passed; 13 CORS tests. Verified live: cross-origin preflight gets no Allow-Origin header. Refactored so `server.py` stays a thin dispatcher. Merged via #5612.

PR #5541, opened 2026-07-04
fix: iOS PWA streaming freeze and SSE reconnect reliability

luperrypf:
## Problem

Three issues affecting iOS/mobile PWA users:

1. **Streaming freeze on long chats** — Every frame during streaming reads `scrollHeight` to decide auto-scroll behavior, forcing WKWebView to re-layout the entire DOM. On large chats (thousands of nodes), this locks up the UI.

2. **SSE connection drops require manual refresh** — The retry window (4 attempts, ~17.5s) is too short for Tailscale/VPN reconnects on iOS. After retries exhaust, the stream is permanently dead and the user must refresh.

3. **Flick-scroll momentum interrupted** — `contain-intrinsic-size` with a hardcoded estimate drifts from actual row heights, causing `scrollHeight` recalculation as rows enter the viewport during fast scrolling. This kills flick-scroll momentum on iOS.

### 1. CSS: `content-visibility` optimization (touch devices only)

```css
@media (pointer: coarse) {
  .msg-row { content-visibility: auto; contain: layout style; }
  .messages { overflow-anchor: none; }
}
```

- `content-visibility: auto` — skips layout/paint for off-screen message rows during streaming, reducing layout cost from O(all rows) to O(visible rows)
- `contain-intrinsic-size` intentionally omitted — a hardcoded size estimate drifts from actual row heights, causing scrollHeight changes as rows enter the viewport during flick-scrolling, which kills momentum. Without it, the scrollbar is less precise on mobile, but momentum scrolling is preserved.
- `overflow-anchor: none` on `.messages` — prevents the browser's scroll anchoring from fighting layout adjustments during fast scrolling
- Scoped to `pointer: coarse` to preserve desktop find-in-page (Ctrl+F)
- Live streaming segment (`.assistant-segment`) naturally unaffected; defense-in-depth overrides included

### 2. JS: Extended retry + session restore fallback

- Retry delays extended from 4 steps (17.5s) to 6 steps (49.5s)
- After all retries exhaust, attempts `_restoreSettledSession()` as last-ditch recovery
  - Polls `/api/session` (not just stream status) — recovers if the LLM finished during the retry window
  - 8s timeout prevents UI hang
  - `_restoreTimedOut` flag prevents timer/await race condition
  - Matches existing `_deferStreamErrorIfOffline` / `_deferStreamErrorIfPageHidden` guard pattern
  - Guards against `_terminalStateReached` / `_streamFinalized` set by other code paths
  - try/catch handles `_restoreSettledSession` rejection gracefully

## Review

Reviewed by Claude Code across multiple rounds. All issues addressed.

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -5765,7 +5765,7 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       // during this retry window and only surface an error after all probes fail.
       if(!_reconnectAttempted && streamId){
         _reconnectAttempted=true;
-        const _retryDelays=[1500,3000,5000,8000];
+        const _retryDelays=[1500,3000,5000,8000,12000,20000];
         setComposerStatus(`Reconnecting… (1/${_retryDelays.length})`);
         const _probeReconnect=async(attempt=0)=>{
           if(_terminalStateReached || _streamFinalized) return;
@@ -5794,6 +5794,40 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
             setTimeout(()=>{void _probeReconnect(attempt+1);}, nextDelay);
             return;
           }
+          // Last-ditch: the stream may have finished while we were retrying.
+          // _restoreSettledSession polls the full session API (not just stream
+          // status) and can recover a completed response without an error banner.
+          // This is especially important on iOS where Tailscale reconnects can
+          // take longer than the retry window.
+          setComposerStatus('Restoring session…');
+          let _restoreTimedOut=false;
+          const _restoreTimer=setTimeout(()=>{
+            // If _restoreSettledSession hangs (flaky Tailscale), don't leave
+            // the UI stuck on "Restoring session…" forever. Fall through to
+            // _handleStreamError after 8s.
+            _restoreTimedOut=true;
+            if(!_terminalStateReached&&!_streamFinalized){
+              if(_deferStreamErrorIfOffline()) return;
+              if(_deferStreamErrorIfPageHidden(source)) return;
+              _flushReasoningToAnchor();
+              _scheduleAnchorRegistryCleanup(120000);
+              _handleStreamError(source);
+            }
+          },8000);
+          try{
+            if(await _restoreSettledSession(source, {preserveVisibleOnShorterTerminalSnapshot:true})){
+              if(_restoreTimedOut) return; // timer already fired _handleStreamError
+              clearTimeout(_restoreTimer);
+              return;
+            }
+          }catch(_){
+            // _restoreSettledSession threw — let the timer handle cleanup
+          }
+          if(_restoreTimedOut) return; // timer already fired _handleStreamError
+          clearTimeout(_restoreTimer);
+          if(_terminalStateReached||_streamFinalized) return;
+          if(_deferStreamErrorIfOffline()) return;
+          if(_deferStreamErrorIfPageHidden(source)) return;
           _flushReasoningToAnchor();
           _scheduleAnchorRegistryCleanup(120000);
           _handleStreamError(source);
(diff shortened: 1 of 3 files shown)


nesquena-hermes (changes requested):
Thanks for this — the app-side change is genuinely solid. I reviewed it independently (not relying on the Greptile pass) and the streaming/SSE logic is clean:

- **No SSE listener leak / duplicate render** — `_wireSSE` closes the existing live source before storing the new one, `_reconnectAttempted` is a correct one-shot gate, and the new last-ditch block only calls `_restoreSettledSession` (a status poll, not a new `EventSource`), so no accumulating listeners.
- **`_restoreTimer` 8s watchdog is cleared on every exit path** and its callback guards on terminal state — no timer leak.
- **CSS is additive** — the new `content-visibility` rules are scoped to `@media (pointer: coarse)` and force live-streaming segments to stay visible, so desktop (pointer:fine) is unaffected.
- **No new i18n key** — the only new user-facing string (`Restoring session…`) is a plain literal matching existing siblings, so no locale-parity impact.

**One blocker before this can go green, and it's purely mechanical:** you extended the reconnect backoff ladder from `[1500,3000,5000,8000]` to `[1500,3000,5000,8000,12000,20000]` in `static/messages.js`, but the repo's source-lock regression test still asserts the *old* literal, so it now fails (it passes on `master`, so this is caused by the change, not a pre-existing flake):

`tests/test_sse_error_multi_probe_reconnect.py`

- **`test_staged_retry_delays_present`** (asserts the exact array substring):
  ```python
  # change
  assert "_retryDelays=[1500,3000,5000,8000]" in _compact(MESSAGES_JS)
  # to
  assert "_retryDelays=[1500,3000,5000,8000,12000,20000]" in _compact(MESSAGES_JS)
  ```
- **`test_stage_counter_starts_at_one`** (finds the exact declaration):
  ```python
  # change
  decl = compact.find("const_retryDelays=[1500,3000,5000,8000];")
  # to
  decl = compact.find("const_retryDelays=[1500,3000,5000,8000,12000,20000];")
  ```
- (Optional, for accuracy) the module docstring near the top also cites `_retryDelays=[1500,3000,5000,8000]` — worth updating the mention to match.

Once those two assertions track the new array the suite goes green again. No source-code changes needed — the `messages.js` / `style.css` logic itself is correct.

Heads-up: the visible streaming-perf behavior (the touch `content-visibility` virtualization and the "Restoring session…" recovery UX) has no automated coverage, so it'll get a device/browser sign-off at review time — nothing for you to change there, just flagging it.

nesquena-hermes:
Shipped in **v0.51.864** — thanks @luperrypf! 🎉

Your iOS PWA streaming reliability fix is live. Long streaming conversations no longer freeze on touch devices (off-screen rows skip layout/paint via `content-visibility` scoped to `pointer:coarse`), and a dropped SSE connection now recovers via the extended retry ladder + a last-ditch "Restoring session…" full-session poll instead of leaving a dead stream.

Gate summary:
- **Codex** (regression): SAFE TO SHIP, no findings.
- **Fable** (UX): SHIP-UX — zero new controls, desktop pixel-identical (all CSS `pointer:coarse`-gated, verified line-by-line), the "Restoring session…" affordance fits the existing status slot.
- The gate caught one real defect and it was fixed before merge: the live-turn force-visible rule originally only covered `data-live-assistant-turn` (stamped only by the Transparent Stream path), so a **normal** live turn (`#liveAssistantTurn` without that attribute) on touch could have been `content-visibility`-skipped and blanked mid-stream. Now covered via the stable `#liveAssistantTurn` id across all render paths — verified live in-browser.
- **Full suite:** 12009 passed; SSE reconnect tests 8/8; desktop rendering confirmed unaffected; zero JS console errors.

One note for anyone with an iPhone: the *scroll feel* when flicking up through long, never-yet-rendered history on iOS's anchoring-less WebKit is the one thing no automated check can verify — worth a quick real-device eyeball if you hit anything jumpy (it replaces a hard freeze, so it's strictly better regardless). Merged via release PR #5586 with your authorship preserved (`Co-authored-by`).

Appreciate the well-scoped mobile fix.

Issue #5542, opened 2026-07-04
De-brittle #5513 SSE-contract RFC anchors: cite by function/route name, not api/routes.py:NNNN (breaks every routes.py PR)


nesquena-hermes:
The session-SSE-contract RFC merged via #5513 pins source anchors as `api/routes.py:NNNN` line numbers, and `tests/test_issue4812_session_sse_contract_rfc.py` asserts specific lines (e.g. `/api/sessions/events` at `api/routes.py:12345`). **Any PR that inserts lines above a pinned anchor breaks these tests** even though the RFC is still accurate in intent.

Confirmed live: PR #5534 (CORS preflight fix) adds +16 lines at `api/routes.py:5127`, shifting `/api/sessions/events` to ~12361, so `test_rfc_cites_current_global_endpoint_source` + `test_rfc_run_journal_anchors_land_on_real_source` fail in #5534's branch — but PASS on clean master. This is brittle-pin collateral, not a #5534 defect. It will recur on every routes.py PR.

**Root-cause fix:** change the RFC + test to cite by **function/route name** (e.g. grep for the route registration `/api/sessions/events` or the handler def name) rather than `path:NNNN` line numbers, so the anchors self-locate and stop breaking on line shifts. (Line numbers on this codebase drift constantly — release-driven.)

Filed by the gate-certifier layer; see #5534 gate cert for the trace.

nesquena-hermes:
Confirmed the mechanism from the source side. Reading `tests/test_issue4812_session_sse_contract_rfc.py:113-127` and the RFC at `docs/rfcs/session-sse-contract-v1.md:41-42, 56-57`, this is a real recurring-breakage vector, not a defect in any individual PR.

## Root cause

The two "hardened" tests still treat the **line number written into the RFC prose** as ground truth. `_anchor_line()` extracts the digits from the RFC text, then indexes `routes.py` at that exact line:

```python
def _anchor_line(label):
    m = re.search(r"%s.*?api/routes\.py:(\d+)" % re.escape(label), text, re.DOTALL)
    ...
    return int(m.group(1))

route_line = _anchor_line("routed at")
assert "/api/sessions/events" in routes[route_line - 1], (...)
```

[...]

PR #5543, opened 2026-07-04
fix: fire-and-forget commit_session_memory to unblock new-session POST

luperrypf:
`commit_session_memory()` can take 1-5+ seconds (extraction call to the memory provider), blocking the `POST /api/session/new` response and making the "+ New Chat" button feel slow or unresponsive.

**Fix**: Move the commit to a daemon thread so the response returns immediately. `commit_session_memory()` already serialises overlapping commits for a session via its own in-flight guard, so this is safe.

Also bump the failure log from `logger.debug` to `logger.warning` so lost commits are visible in logs (daemon threads can be killed on shutdown).

**Before**: request handler blocks until memory commit completes
**After**: response returns immediately, memory commit runs in background

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -13067,21 +13067,34 @@ def handle_post(handler, parsed) -> bool:
                 # the new session (#5420).
                 prev_session_id = None
             if prev_session_id:
-                try:
-                    from api.session_lifecycle import commit_session_memory
-                    from api.config import SESSION_AGENT_CACHE, SESSION_AGENT_CACHE_LOCK
-                    prev_agent = None
-                    with SESSION_AGENT_CACHE_LOCK:
-                        _cached = SESSION_AGENT_CACHE.get(prev_session_id)
-                        if _cached:
-                            prev_agent = _cached[0]
-                    commit_session_memory(prev_session_id, agent=prev_agent)
-                except Exception:
-                    logger.debug(
-                        "Lifecycle commit for prev_session %s failed",
-                        prev_session_id,
-                        exc_info=True,
-                    )
+                # Fire-and-forget: commit_memory_session() can take 1-5+ seconds
+                # (extraction call to the memory provider), and blocking the
+                # response here made "+ New Chat" feel slow/unresponsive.
+                # commit_session_memory() already serialises overlapping commits
+                # for a session via its own in-flight guard, so running it off
+                # the request thread is safe.
+                def _commit_prev_session_memory(_sid=prev_session_id):
+                    try:
+                        from api.session_lifecycle import commit_session_memory
+                        from api.config import SESSION_AGENT_CACHE, SESSION_AGENT_CACHE_LOCK
+                        prev_agent = None
+                        with SESSION_AGENT_CACHE_LOCK:
+                            _cached = SESSION_AGENT_CACHE.get(_sid)
+                            if _cached:
+                                prev_agent = _cached[0]
+                        commit_session_memory(_sid, agent=prev_agent)
+                    except Exception:
+                        logger.warning(
+                            "Lifecycle commit for prev_session %s failed",
+                            _sid,
+                            exc_info=True,
+                        )
+
+                threading.Thread(
+                    target=_commit_prev_session_memory,
+                    daemon=True,
+                    name=f"commit-memory-{prev_session_id}",
+                ).start()
         s = new_session(
             workspace=workspace,
             model=model,
--- a/api/routes.py
+++ b/api/routes.py
@@ -13090,11 +13090,14 @@ def handle_post(handler, parsed) -> bool:
                             exc_info=True,
                         )
 
-                threading.Thread(
+                t = threading.Thread(
                     target=_commit_prev_session_memory,
                     daemon=True,
                     name=f"commit-memory-{prev_session_id}",
-                ).start()
+                )
+                from api.session_lifecycle import _register_background_commit_thread
+                _register_background_commit_thread(t)
+                t.start()
         s = new_session(
             workspace=workspace,
             model=model,
--- a/api/session_lifecycle.py
+++ b/api/session_lifecycle.py
@@ -45,6 +45,26 @@ _condition = threading.Condition(_lock)
 
 _sessions: dict[str, dict] = {}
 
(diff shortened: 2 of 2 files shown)


luperrypf:
Good catch — addressed in 0299ba3. Background commit threads are now tracked in a module-level set and `drain_all_on_shutdown()` joins them (5s timeout) before draining inline. Thanks!

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (good intent + drain design, but the drain never runs on SIGTERM → the memory-loss it prevents still happens; + thread-set leak — 2 SILENT)

**Certified head:** `sha:0299ba3e` (rebased onto current master, `git apply` clean) · **PR:** #5543 · _luperrypf, fix: fire-and-forget commit_session_memory to unblock new-session POST_
**Verdict:** The reliability intent is right (don't block New Chat on a slow memory-commit) and the daemon-thread + serialized-commit + shutdown-drain design is thoughtful — BUT Codex verified the drain never runs on the actual stop path: `ctl.sh` stops via SIGTERM, `server.py` has no SIGTERM handler, so Python skips the `finally` that calls `drain_all_on_shutdown()`. Net: the fire-and-forget can still lose memory commits on a normal `./ctl.sh stop` — the exact data loss the drain was added to prevent. Plus a thread-registry leak. No test.

### What I ran (rebased worktree `/tmp/wt-rebase-5543`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 2 SILENT (SIGTERM-skips-drain; thread-set leak); I confirmed both |
| Full pytest suite | 4 failed / 11981 passed — 2 known env flakes + **2 brittle-#5513-RFC-anchor tests (NOT this PR — collateral, filed #5542)** |
| Test coverage | ⚠️ NONE added for a reliability/threading change |

### Findings
**⛔ SILENT (I + Codex CONFIRMED) — drain doesn't run on the real stop → memory commits still lost** (`server.py:14-17` vs `:692-703`): `drain_all_on_shutdown()` (which drains the new background commit threads) is in the `finally` after `httpd.serve_forever()`, but `server.py` installs ONLY a SIGPIPE handler — no SIGTERM handler. `ctl.sh` stops WebUI with SIGTERM (`kill TERM`), and Python's default SIGTERM terminates the process WITHOUT running `finally` (Codex ran a local SIGTERM sanity check confirming this). So on the normal managed stop, `drain_all_on_shutdown()` never fires while a daemon commit thread is in flight → the fire-and-forget loses that commit. The safety net is defeated by the missing signal handler. **Fix:** install a SIGTERM handler (after `httpd` is created) that makes `serve_forever()` exit through the existing `finally` — e.g. a helper thread calling `httpd.shutdown()` — so `drain_all_on_shutdown()` runs before exit.

**⛔ SILENT (I + Codex CONFIRMED) — background-commit thread registry grows forever** (`session_lifecycle.py:60`): `_background_commit_threads` is only cleared by `_drain_background_commit_threads()`; the worker doesn't remove itself on completion. A long-lived server with many New-Chat handoffs accumulates one strongly-referenced dead `Thread` per handoff until shutdown. **Fix:** `finally: _unregister_background_commit_thread(threading.current_thread())` (discard-on-done).

**⚠️ No test** for a reliability/threading fix (Codex + I flag): add coverage for the fire-and-forget path + the drain (and, once the SIGTERM handler is added, that SIGTERM triggers the drain).

**ℹ️ CI-red partly brittle-#5513 collateral (NOT this PR):** 2 of the 4 suite failures are `test_issue4812` line-anchor tests — they pass on clean master; this PR's +32 routes.py lines shift the RFC's pinned `api/routes.py:NNNN` anchors (same as #5534; root-cause de-brittle filed as #5542). The other 2 are the known nous/issue4536 env flakes.

**RED — gate-fail/changes-requested (2 SILENT): (1) install a SIGTERM handler so the normal `./ctl.sh stop` runs `drain_all_on_shutdown()` (otherwise the fire-and-forget loses memory commits — the exact bug the drain targets); (2) unregister each background commit thread on completion (`finally` discard) to stop the registry leak; (3) add a test for the fire-and-forget + drain path.** The intent (unblock New Chat) and the daemon/serialize/drain design are good — the gap is that the drain isn't reachable on the real stop signal. Once SIGTERM triggers the drain and the registry self-cleans, this is a solid reliability fix. concept 4/5 (real New-Chat-latency fix; the async/data-safety mechanics need the signal wiring). Author @luperrypf (T1). crit=3, reliability. (Gate value: the drain LOOKED wired — it IS called from `drain_all_on_shutdown()` — but Codex went the extra step and verified that function never runs on SIGTERM; "wired into the finally" ≠ "runs on the stop signal" when there's no SIGTERM handler.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; confirmed the drain is in `finally` after serve_forever but server.py installs only SIGPIPE (no SIGTERM) so `ctl.sh stop`=SIGTERM skips it (Codex SIGTERM sanity check), and `_background_commit_threads` only clears on drain (per-handoff dead-Thread growth); 2 CI failures are brittle-#5513 collateral (pass on clean master, #5542). Cert valid for sha:0299ba3e._

nesquena-hermes:
Shipped in **v0.51.863** — thanks @luperrypf! 🎉

Your fix to unblock "+ New Chat" is live. Memory extraction for the previous session now runs in the background so creating a new conversation returns immediately instead of stalling 1–5+ seconds.

Maintainer hardening was folded in to keep the deferred work safe across a managed restart, and the dual gate (Codex + Opus) earned its keep here — it caught, with runnable repros, two real defects in that hardening:
- The shutdown drain never ran on `SIGTERM` (the default disposition terminated without unwinding), so a managed stop still lost in-flight commits → fixed with a `SIGTERM` handler that drives `serve_forever()` through its `finally` drain.
- The drain deadline didn't actually fire against a worker wedged in a hung memory provider (both the `_condition.wait()` path and the direct provider call) → fixed by running each inline commit in a budget-bounded daemon worker, with a `_draining` flag that stops new workers starting mid-drain.

All backed by regression tests (verified to fail without the fix, pass with it): 25 in the #5543 suite + 16 lifecycle-generation tests, full suite green. Merged via release PR #5581 with your authorship preserved (`Co-authored-by`).

Appreciate the clean perf fix.

PR #3250, opened 2026-05-31
fix: extend upward scroll intent timeout to prevent streaming scroll snap-back

emanon312:
## Problem

During streaming, users who scroll up to read earlier content get snapped back to the bottom — especially when they pause scrolling for a moment before the message finishes generating.

## Root Cause

The `MESSAGE_UPWARD_INTENT_MS` timeout (`_recentMessageUpwardIntent()`) was only **450ms**. The race condition:

1. User scrolls up during streaming → `_scrollPinned = false` ✓
2. User pauses scrolling to read (>450ms since last wheel event)
3. DOM layout changes from the streaming markdown parser (smd), tool card insertions, or code re-highlighting trigger scroll events
4. Handler no longer recognizes these as user-initiated (intent expired) → `movedUp = false`
5. If position lands inside the 250px near-bottom zone for 2 consecutive samples → `_scrollPinned = true`
6. Next streaming token → `scrollIfPinned()` → forces `scrollTop = scrollHeight` → **snap!**

This is particularly noticeable on long responses where the user scrolls up a small amount (still within 250px of bottom) and pauses to read.

## Fix

Increase `MESSAGE_UPWARD_INTENT_MS` from **450ms → 2000ms**.

This gives the user a 2-second protection window after their last upward scroll, during which DOM-triggered scroll events are correctly identified as co-occurring with user intent, preventing re-pinning.

## What is NOT affected

- **Downward scrolling** still re-pins normally (`movedUp` requires `top < _lastScrollTop - 2` which is false for downward movement regardless of intent timeout)
- **Scroll-to-bottom button** still works (`scrollToBottom()` directly sets `_scrollPinned = true`)
- **macOS trackpad momentum** protection (#1360) is preserved
- **Sidebar scroll isolation** (#1784) is preserved
- **Session switch scroll reset** (#1731 follow-up) is preserved

## Testing

No existing tests hardcode the 450ms value — the test suite validates the structural behavior (direction detection, unpin semantics, hysteresis thresholds) without depending on the specific timeout.

Refs: #1360, #1731

PR #5544, opened 2026-07-04
fix: re-pin auto-scroll in scrollIfPinned after user scrolls back to bottom

luperrypf:
When a user scrolls up during streaming, `_messageUserUnpinned` is set `true` and `scrollIfPinned()` permanently stops auto-following — only `scrollToBottom()` clears the flag, but `scrollIfPinned()` is what runs on every streamed chunk.

**Fix**: Add a re-pin path in `scrollIfPinned()` that mirrors the scroll listener's `_nearBottomCount` debounce:
1. After 2 consecutive near-bottom (250px) streaming chunks
2. Without active scroll intent (`_recentNonMessageScrollIntent`, `_recentMessageTouchScrollIntent`)
3. Clear the unpinned flag and resume auto-follow

Also normalize the listener's re-pin threshold from 80px to 250px to match its own `nearBottom` gate, eliminating a dead zone (81-249px) where the counter could reach 2 but the re-pin would silently fail.

**Before**: one scroll-up → permanent auto-follow lockout
**After**: scroll back near bottom → 2 chunks confirm → auto-follow resumes

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -4858,7 +4858,7 @@ if(typeof window!=='undefined'){
       }else if(movedDown&&nearBottom){
         _nearBottomCount=_nearBottomCount+1;
         if(_nearBottomCount>=2){
-          if(!_messageUserUnpinned||bottomDistance<=80){
+          if(!_messageUserUnpinned||bottomDistance<=250){
             _messageUserUnpinned=false;
             _scrollPinned=true;
           }
@@ -5516,7 +5516,21 @@ function _settleFinalScroll(token){
 }
 function scrollIfPinned(){
   if(!_autoScrollFollow) return;
-  if(_messageUserUnpinned) return;
+  if(_messageUserUnpinned){
+    // Only scrollToBottom() cleared this flag, so one scroll-up permanently
+    // killed auto-follow. Re-pin through the same _nearBottomCount debounce
+    // the scroll listener uses (~4859-4866) instead of flipping the flags
+    // on the first chunk that lands near bottom, since this runs on every
+    // streamed chunk and would otherwise fight a reader who's actively
+    // scrolling away, and bypass the listener's own re-pin guards.
+    if(_messageBottomDistance()>250){ _nearBottomCount=0; return; }
+    _nearBottomCount=_nearBottomCount+1;
+    if(_nearBottomCount<2) return;
+    _nearBottomCount=0;
+    if(_recentNonMessageScrollIntent()||_recentMessageTouchScrollIntent()) return;
+    _messageUserUnpinned=false;
+    _scrollPinned=true;
+  }
   if(!_scrollPinned) return;
   if(_recentNonMessageScrollIntent()) return;
   if(_messageBottomDistance()>500) _setMessageScrollToBottom();
--- a/static/ui.js
+++ b/static/ui.js
@@ -4858,10 +4858,10 @@ if(typeof window!=='undefined'){
       }else if(movedDown&&nearBottom){
         _nearBottomCount=_nearBottomCount+1;
         if(_nearBottomCount>=2){
-          if(!_messageUserUnpinned||bottomDistance<=250){
-            _messageUserUnpinned=false;
-            _scrollPinned=true;
-          }
+          // nearBottom already guarantees bottomDistance<250, so the
+          // re-pin threshold is always satisfied when we reach this branch.
+          _messageUserUnpinned=false;
+          _scrollPinned=true;
           _nearBottomCount=0;
         }
       }else if(!_messageUserUnpinned){


luperrypf:
Good catch on both — addressed:

1. **Shared counter**: The code comment already says "the same _nearBottomCount debounce the scroll listener uses" — the shared counter is intentional. The PR description was slightly imprecise; the cooperative counting (scroll events + stream chunks sharing one counter) is the desired behavior.

2. **Vacuous if**: Removed in ccbfd9d. Now just the assignments with a comment explaining why it's unconditional.

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (re-pins on mere near-bottom proximity → yanks a reading user back; breaks 3 scroll-pin regression tests — CORE)

**Certified head:** `sha:ccbfd9d1` (rebased onto current master, `git apply` clean) · **PR:** #5544 · _luperrypf, fix: re-pin auto-scroll in scrollIfPinned after user scrolls back to bottom_
**Verdict:** The intent is good (one scroll-up shouldn't permanently kill auto-follow), but the re-pin trigger is too loose: it re-pins after 2 streamed chunks whenever within **250px** of bottom, without requiring an actual return-to-bottom and without checking message-pane wheel/key intent. So a small scroll-up near the tail (while reading the last lines) gets **yanked back to bottom** mid-stream. Confirmed by 3 existing scroll-pin regression tests going red.

### What I ran (rebased worktree `/tmp/wt-rebase-5544`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 1 CORE (passive re-pin at 250px yanks reader); I confirmed + found backing test failures |
| Full pytest suite | 5 failed / 11980 passed — 2 known env flakes (nous, issue4536) + **3 REAL scroll-pin regressions this PR causes** |

### Findings
**⛔ CORE (I + Codex CONFIRMED, + 3 regression tests) — streaming yanks a reader back to bottom after a small scroll-up near the tail** (`static/ui.js:5526`): the new `scrollIfPinned()` branch clears `_messageUserUnpinned` after 2 streamed calls whenever `_messageBottomDistance() <= 250`, then `_settleMessageScrollToBottom(false)` writes bottom (ui.js:5441). It does NOT require an actual return-to-bottom, and it checks `_recentMessageTouchScrollIntent()`/`_recentNonMessageScrollIntent()` but OMITS the existing `_recentMessageWheelIntent()` (4504) and `_recentMessageKeyScrollIntent()` (4517) — so a wheel/keyboard scroll-up within 250px of the tail is overridden. **Backing regression failures (this PR turns them red):**
- `test_issue4295_scroll_pin_reentry::test_near_bottom_proximity_alone_does_not_repin` — asserts the EXACT invariant the PR violates (proximity alone must not re-pin).
- `test_issue3250_upward_scroll_intent_window::test_scroll_if_pinned_respects_sticky_user_unpin` — sticky-unpin broken.
- `test_tars_scroll_reset_regressions::test_user_scroll_cancels_delayed_bottom_settling`.
**Fix (Codex-exact):** don't passively re-pin at the 250px near-bottom threshold — require an actual bottom return (strict `<=80`, matching the prior invariant), and bail/reset `_nearBottomCount` on recent message-pane wheel/key/scroll intent (`_recentMessageWheelIntent()`, `_recentMessageKeyScrollIntent()`, `_recentMessageScrollIntent()`) before incrementing.

**RED — gate-fail/changes-requested (1 CORE, breaks 3 scroll-pin regression tests): tighten the re-pin to require an actual return-to-bottom (`<=80`, not 250px near-bottom) and add the missing `_recentMessageWheelIntent()` + `_recentMessageKeyScrollIntent()` bails so a wheel/keyboard scroll-up near the tail isn't overridden.** The goal (resume auto-follow when the user genuinely returns to bottom) is worth shipping — but it must not re-pin on mere proximity (the `#4295 proximity-alone-does-not-repin` invariant) or fight a wheel/key reader. Once tightened, the 3 red regression tests should pass and it's a real UX win. Then, since it's visible scroll behavior, Nathan's UX call on the exact resume threshold. concept 4/5 (real annoyance fix; the trigger is too eager). Author @luperrypf (T2). crit=3. (Gate value: the full suite caught it directly — 3 prior-scroll-fix regression tests (#3250/#4295/tars) encode the "proximity alone does not re-pin" invariant this PR breaks; always run the full suite on scroll-behavior changes, the regressions are pre-encoded.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; the passive-re-pin-at-250px CORE confirmed (ui.js:5526 clears _messageUserUnpinned on proximity without return-to-bottom, omits _recentMessageWheelIntent/_recentMessageKeyScrollIntent), backed by 3 failing scroll-pin regression tests (#4295 proximity-alone-does-not-repin, #3250 sticky-unpin, tars delayed-settle); 2 other failures are known nous/issue4536 flakes. Cert valid for sha:ccbfd9d1._

nesquena-hermes:
Shipped in **v0.51.862** — thanks @luperrypf! 🎉

Your auto-scroll re-pin fix is live. The permanent auto-follow lockout is gone: once you've scrolled up during a streaming reply, returning to the true bottom of the transcript (within ~80px) now re-engages auto-scroll so the view follows the stream again — previously only starting a new turn re-enabled it.

The re-pin is deliberately conservative: it only fires at the true-bottom tail and bails on any active scroll intent (wheel / keyboard / touch / scrollbar drag), so reading back through history mid-stream never yanks you down — the #4295 invariant is preserved. The listener's `<=80px` true-bottom gate is restored too, so the two re-pin paths and the ↓-button all use one threshold.

Gate: rebuilt on current master and re-gated — Codex clean (regression), Fable **SHIP-UX** (matches ChatGPT's scroll-away-during-stream behavior; the 80px resume is the right call), 20/20 scroll-regression tests, full suite green, plus a live browser drive on a seeded 40-message session that confirmed both behaviors (no yank while reading, follow re-engages at bottom) with zero console errors. Merged via release PR #5580 with your authorship preserved.

Appreciate the sharp fix to a genuinely annoying regression.

PR #5553, opened 2026-07-04
Fix #5532: /api/session/clear sets truncation watermark (P0 data loss)

### LINKED ISSUE #5532: /api/session/clear does not delete messages from state.db — history survives clear+refresh
### Summary

The `/api/session/clear` endpoint only clears the session JSON file (`s.messages = []` → `s.save()`), but does **not** delete the corresponding messages from `state.db`. After clearing and refreshing the page, all historical messages reappear. Continuing the conversation also retains full context from before the clear.

### Reproduction

1. Open a WebUI session with some conversation history
2. Click "Clear Conversation" (calls `POST /api/session/clear`)
3. The chat area appears empty — looks cleared
4. **Refresh the page** → all previous messages reappear
5. **Send a new message** → the agent responds with full context of the "cleared" conversation

### Root Cause

`/clear` handler in `api/routes.py` (line ~13684):

```python
with _get_session_agent_lock(sid):
    s.messages = []
    s.tool_calls = []
    apply_session_title_rename(s, "Untitled")
    s.save()          # ← Only writes session JSON file
    # state.db messages are NOT deleted
```

The session JSON file (`messages=[]`) and `state.db` (Hermes agent SQLite database) are **two separate data stores**. On page refresh, `GET /api/session` reads messages from `state.db` via `get_state_db_session_messages()` (api/models.py:6137), which returns the full un-deleted history.

Additionally, `_evict_session_agent()` returns early if there is an active run (`ACTIVE_RUNS`), so the agent instance may survive with its full conversation context intact.

### Expected Behavior

`/clear` should:
1. Delete all messages for the session from `state.db`
2. Clear the session JSON (`messages=[]`, `tool_calls=[]`)
3. Reset the title to "Untitled"
4. Fully evict the cached agent (or ensure new agent does not restore state.db history)

### Environment

- Hermes WebUI: current master (post v0.18.0)
- Provider: zai/GLM

### Impact

Users who rely on `/clear` to start fresh will unknowingly leak previous conversation context to the agent, which may have privacy and cost implications.

nesquena-hermes:
## Summary

Closes #5532 (P0 data loss).

`POST /api/session/clear` cleared the sidecar transcript but **never recorded a
truncation watermark**, so `state.db` messages resurrected on the next
`/api/session` read: history reappeared after clear + refresh, and a continued
conversation still carried the full pre-clear context into the model.

## Root cause

The `/api/session/clear` handler (`api/routes.py`) did:

```python
s.messages = []
s.tool_calls = []
apply_session_title_rename(s, "Untitled")
s.save()
```

It never set `s.truncation_watermark`. The append-only state.db merge
(`merge_session_messages_append_only`, reached via
`reconciled_state_db_messages_for_session`) treats an unset / `None` watermark
as **"keep everything and just dedup"**. Because `state.db` is append-only, the
cleared transcript was still there, so the merge re-added all of it on the next
read — the exact data-loss symptom in the issue.

The sibling destructive op `/api/session/truncate` does **not** have this bug:
it calls `truncate_session_at_keep(session, keep)` in `api/session_ops.py`,
which sets `truncation_watermark = truncation_boundary =
_truncation_watermark_for(kept_messages)`.

## Fix

A full clear is just a truncate that keeps **zero** messages. Route `/clear`
through the **same** helper the truncate handler uses:

```python
truncate_session_at_keep(s, 0)   # empties messages + context_messages,
                                 # sets watermark = boundary = 0.0
s.tool_calls = []
apply_session_title_rename(s, "Untitled")
s.save()
```

`_truncation_watermark_for([]) == 0.0`, which is the **#2914 "truncate-to-empty"
sentinel** that blocks **all** state.db replay. The merge/read path is
**untouched** — `/clear` and `/truncate` now set the marker identically, so the
merge contract is not forked.

## Second half of the issue (continued-context) — verified

The issue notes that a continued conversation retained pre-clear context. That
is fixed by the same change: `truncate_session_at_keep(s, 0)` empties
`context_messages` in lockstep with `messages` (via
`truncate_context_for_display_keep(..., keep=0) -> []`), so the model-facing
context no longer carries the cleared turns. A regression test asserts
`context_messages == []` after clear.

## Tests

`tests/test_issue5532_clear_truncation_watermark.py` (drives the real
`POST /api/session/clear` route via `handle_post`, mirroring the #2914
integration harness):

- `test_clear_endpoint_sets_truncation_watermark` — watermark & boundary set to `0.0`
- `test_clear_empties_context_messages` — second-half fix: `context_messages == []`
- `test_clear_then_read_does_not_resurrect_state_db_messages` — subsequent
  state.db merge returns EMPTY (no resurrection)
- `test_clear_matches_truncate_to_empty_marker` — `/clear` and
  `/truncate(keep=0)` produce an identical marker (no forked merge logic)

Fail-without-fix confirmed: all four fail on `origin/master`
(`truncation_watermark` stays `None`, resurrection occurs) and pass with this
change.

```
$ .venv/bin/python3 -m pytest tests/test_issue5532_clear_truncation_watermark.py -q
....                                                                     [100%]
4 passed in 3.80s
```

Related watermark suite stays green:

```
$ .venv/bin/python3 -m pytest tests/test_issue5532_clear_truncation_watermark.py \
    tests/test_issue2914_truncation_watermark.py tests/test_issue3831_watermark_clear.py -q
30 passed
```

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -13717,6 +13717,20 @@ def handle_post(handler, parsed) -> bool:
             )
             truncate_session_at_keep(s, 0)
             s.tool_calls = []
+            # #5532 (Codex gate): a compressed-continuation child persists its
+            # archived transcript in a parent sidecar marked
+            # pre_compression_snapshot and stitches it back for display via
+            # _webui_sidecar_lineage_messages_for_display(). That stitch merges
+            # the child's own messages with truncation_watermark=None, so the
+            # 0.0 truncate-to-empty sentinel we just set on the CHILD does NOT
+            # stop the PARENT snapshot from resurrecting the pre-clear transcript
+            # on refresh. Detach the compression lineage so a cleared session is
+            # genuinely empty: drop the snapshot parent link + the anchor fields
+            # so the child no longer resolves a pre_compression_snapshot parent.
+            s.parent_session_id = None
+            s.compression_anchor_visible_idx = None
+            s.compression_anchor_message_key = None
+            s.compression_anchor_summary = None
             # Reset the title via the rename helper so clearing a manually-named
             # session also clears manual_title/llm_title_generated — otherwise the
             # reused session keeps its manual-title protection and never auto-names
--- a/api/routes.py
+++ b/api/routes.py
@@ -13694,13 +13694,33 @@ def handle_post(handler, parsed) -> bool:
             return bad(handler, "Session not found", 404)
         sid = body["session_id"]
         with _get_session_agent_lock(sid):
-            s.messages = []
+            # Clear is a full truncate: keep zero messages. Route it through the
+            # SAME helper the /api/session/truncate handler uses so the merge
+            # contract is not forked (#5532 P0 data loss).
+            #
+            # Before this fix /clear wiped s.messages/s.tool_calls but never set
+            # truncation_watermark. The append-only state.db merge
+            # (merge_session_messages_append_only) treats an unset watermark as
+            # "keep everything", so on the next /api/session read the cleared
+            # turns were resurrected from state.db (history reappeared after
+            # clear+refresh) and — because context_messages also survived — a
+            # continued turn still carried the pre-clear context.
+            #
+            # truncate_session_at_keep(s, 0) empties messages AND context_messages
+            # and sets truncation_watermark = truncation_boundary =
+            # _truncation_watermark_for([]) == 0.0. 0.0 is the #2914
+            # "truncate-to-empty" sentinel that blocks ALL state.db replay, so the
+            # merge honors the clear instead of re-adding the deleted transcript.
+            from api.session_ops import (
+                apply_session_title_rename,
+                truncate_session_at_keep,
+            )
+            truncate_session_at_keep(s, 0)
             s.tool_calls = []
             # Reset the title via the rename helper so clearing a manually-named
             # session also clears manual_title/llm_title_generated — otherwise the
             # reused session keeps its manual-title protection and never auto-names
             # again (#3542 lifecycle gap).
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Cross-ref for merge ordering: **#5569** (de-brittle #5542) converts the SSE-contract RFC to cite `api/routes.py` symbols **by name** instead of `:NNNN` line numbers, and removes the line anchors entirely.

This PR (#5553) currently includes a small RFC hunk that *realigns* those line numbers (+46, to survive this PR's routes.py insertion) so the pre-#5569 line-anchor test stays green. That realignment becomes obsolete once #5569 lands.

- If **#5569 merges first**: this PR's RFC hunk will conflict trivially — resolve by dropping the line-number realignment (the anchors no longer exist; the symbol-based test doesn't care about this PR's line shift).
- If **#5553 merges first**: #5569 cleanly deletes the (realigned) line numbers.

No code conflict either way — both only touch the RFC's citation style. The P0 code fix in this PR is independent of #5569.

nesquena-hermes:
Shipped in **v0.51.860** 🎉 as part of the combined #5532 fix (primary PR #5556 by @rodboev).

Your improvements were folded in and credited (`Co-authored-by`): the shared `truncate_session_at_keep(s,0)` helper-reuse and — the key Codex-caught deeper path — detaching the `pre_compression_snapshot` parent lineage so a compressed-continuation child can't resurrect its pre-clear transcript via the parent display-stitch (which merges with `watermark=None`, bypassing the child's 0.0 sentinel). Guarded to detach only compression-snapshot parents so ordinary fork links survive. Both your test file and #5556's ship. Closing as combined-and-shipped — thank you.

Issue #3929, opened 2026-06-10
Turn-level error handling discards all partial work (text, reasoning, tool calls)


claw-io:
## Summary

When a Hermes session encounters a provider error, credential failure, network drop, or any exception mid-turn, the WebUI discards all accumulated partial work — assistant text, reasoning steps, and tool call results — from the session sidecar. The user sees only a generic error message. This contrasts with the user-initiated cancel path (`cancel_stream`), which correctly snapshots and preserves partial content.

For a turn where the agent worked for 20 minutes executing complex tools, a final-step provider error throws away all intermediate progress from the WebUI's perspective. The core agent transcript (state.db) does preserve messages, but the WebUI sidecar does not, causing context desynchronization on subsequent turns.

## Root Cause

Two error-handling paths in `api/streaming.py` clear the session's pending state and append a terminal `_error=True` block **without first harvesting accumulated partial content** from the streaming memory buffers:

### Path 1: Silent Failure (no assistant response returned)
`api/streaming.py` ~line 6795:
```python
_materialize_pending_user_turn_before_error(s)
s.active_stream_id = None
s.pending_user_message = None
s.pending_attachments = []
s.pending_started_at = None
_error_message = {
    'role': 'assistant',
    'content': f'**{_err_label}:** {_error_payload.get("message") or _err_label}\n\n*{_err_hint}*',
    'timestamp': int(time.time()),
    '_error': True,
}
s.messages.append(_error_message)
s.save()
```

### Path 2: Unhandled Exception
`api/streaming.py` ~line 7744:
```python
_materialize_pending_user_turn_before_error(s)
s.active_stream_id = None
s.pending_user_message = None
s.pending_attachments = []
s.pending_started_at = None
_error_message = {
    'role': 'assistant',
    'content': f'**{_exc_label}:** {_error_payload.get("message") or err_str}' + (f'\n\n*{_exc_hint}*' if _exc_hint else ''),
    'timestamp': int(time.time()),
    '_error': True,
}
s.messages.append(_error_message)
s.save()
```

### Why Partial Content Is Lost

1. **No buffer snapshot**: Both paths skip reading `STREAM_PARTIAL_TEXT[stream_id]`, `STREAM_REASONING_TEXT[stream_id]`, and `STREAM_LIVE_TOOL_CALLS[stream_id]` before clearing state.

[...]

PR #5559, opened 2026-07-04
fix(#3929): classify credential-pool exhaustion with an actionable hint (partial)

nesquena-hermes:
## Partial fix for #3929 — accurate error classification for credential-pool exhaustion

This addresses one concrete, provable-by-inspection defect surfaced while investigating the #3929 recurrence — **not** the full data-loss (see scope below).

### Problem
The error a zero-credential pool produces — `All 0 credential(s) exhausted for opencode_go` — does **not** match `_is_quota_error_text` (`'credential(s) exhausted'` ≠ `'credits exhausted'`), so `_classify_provider_error` returned a generic classification. And in the exception error-labeling path, the `else` branch sets `('Error', 'error', '')` — **discarding the classification hint entirely** — so the user saw a bare **"Error"** with no guidance for exactly this failure. (This also means the recurrence comment's claim that this classifies as `quota_exhausted` was inaccurate on current head.)

### Fix
- New `credential_pool_empty` classification (checked **before** the generic quota branch) with a pool-specific, actionable hint: the credential pool has no usable keys for this provider — add/refresh a key or switch providers via `hermes model`.
- Explicit `credential_pool_empty` branch in the exception path so the hint is surfaced instead of being dropped to the generic `'Error'`.
- A zero-credential **pool** is a config problem, not account-credit exhaustion, so it is deliberately **not** folded into the "Out of credits / top up your account" quota message.

### Scope — what this does NOT claim
This is the **classification/UX-accuracy** half. It does not fix the partial-work **loss** the reporter experienced. Two pieces remain open on the issue:
1. **Reload-reconcile reasoning-restore** — `_append_journaled_partial_output` intentionally does not restore hidden reasoning (`models.py` docstring); if the lost "huge chunks" were thinking-pane content, restoring it is real headroom, but it collides with the #5024 reasoning-replay provider-policy gate and needs a display-only design + its own gate. Tracked as a separate follow-up.
2. **Incident autopsy** — the two-terminal-markers-11-hours-apart evidence is not produced by any single error path traced, so it still needs the reporter's server logs (which path stamped the marker + whether the WebUI process restarted mid-turn).

### Tests
`tests/test_issue3929_credential_pool_classification.py` — distinct classification, not-misclassified-as-account-quota, and real-quota-still-classifies-as-quota (guards against over-match). Fail-without-fix verified; 35 adjacent classification/quota/auth tests green.

Refs #3929 (partial — classification only).

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -1195,8 +1195,20 @@ def _classify_provider_error(err_str: str, exc=None, *, silent_failure: bool = F
             'hint': 'The run stopped before a provider response completed. If you did not cancel it, try again.',
         }
     _is_quota = _is_quota_error_text(err_str)
+    # A credential-POOL exhaustion ("All 0 credential(s) exhausted for <provider>")
+    # is a distinct shape from account/plan quota: it means the profile's
+    # credential pool has no usable keys for that provider (a config problem),
+    # not that a funded account ran out of credits. It is NOT matched by
+    # _is_quota_error_text ('credential(s) exhausted' != 'credits exhausted'), so
+    # without this it fell through to the generic error label/hint. Classify it
+    # explicitly so the user gets a pool-specific, actionable hint. (#3929)
+    _is_credential_pool_empty = (
+        'credential(s) exhausted' in _err_lower
+        or 'credentials exhausted' in _err_lower
+        or ('credential' in _err_lower and 'exhausted' in _err_lower)
+    )
     _is_auth = (
-        not _is_quota and (
+        not _is_quota and not _is_credential_pool_empty and (
             _probe_status_code == 401
             or
             '401' in err_str
@@ -1228,6 +1240,12 @@ def _classify_provider_error(err_str: str, exc=None, *, silent_failure: bool = F
         or ('context length exceeded' in _err_lower and 'cannot compress further' in _err_lower)
         or ('context compression' in _err_lower and 'max compression attempts' in _err_lower)
     )
+    if _is_credential_pool_empty:
+        return {
+            'label': 'No usable credentials',
+            'type': 'credential_pool_empty',
+            'hint': 'The credential pool for this provider has no usable keys left (all entries exhausted or unconfigured). Add or refresh a key for this provider in your Hermes config / credential pool, or switch providers via `hermes model`.',
+        }
     if _is_quota:
         return {
             'label': 'Out of credits',
@@ -9454,6 +9472,7 @@ def _run_agent_streaming(
             put('cancel', _cancel_event_payload('Cancelled by user'))
             return
         _exc_is_quota = _classification['type'] == 'quota_exhausted'
+        _exc_is_credential_pool_empty = _classification['type'] == 'credential_pool_empty'
         # Exception quota text still includes: 'more credits' in _exc_lower, 'can only afford' in _exc_lower, 'fewer max_tokens' in _exc_lower.
         # Rate-limit detection remains guarded as: (not _exc_is_quota).
         _exc_is_rate_limit = (_classification['type'] == 'rate_limit') and (not _exc_is_quota)
@@ -9464,6 +9483,10 @@ def _run_agent_streaming(
 
         # The user hint still points to Settings / `hermes model` from _classify_provider_error().
         if _exc_is_quota:
+            _exc_label, _exc_type, _exc_hint = (
+                _classification['label'], _classification['type'], _classification['hint'],
+            )
+        elif _exc_is_credential_pool_empty:
             _exc_label, _exc_type, _exc_hint = (
                 _classification['label'], _classification['type'], _classification['hint'],
             )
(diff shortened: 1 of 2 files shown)


nesquena (approved):

Independent review of the classification half of #3929: a zero-credential pool error (`All 0 credential(s) exhausted for opencode_go`) surfaced as a bare **"Error"** with no hint. This adds a distinct `credential_pool_empty` classification with an actionable, pool-specific hint.

### What this ships

`api/streaming.py` (+25/-1 — new classification branch + exception-path label branch), `tests/test_issue3929_credential_pool_classification.py` (+52, new, 3 tests). Agent-authored (`nesquena-hermes`). MERGEABLE, CI fully green. Explicitly scoped: classification/UX only — the partial-work-loss half of #3929 stays open (reasoning-restore follow-up + incident autopsy, per the issue comments).

### Both claimed master defects verified

1. **`_is_quota_error_text` really doesn't match** — I read master's pattern list (`insufficient credit`, `credit balance`, `credits exhausted`, `quota exceeded`, …): no `credential` pattern anywhere, so `credential(s) exhausted` fell through to generic. ✓
2. **The exception path really discarded the hint** — the label chain's `else` fallback is `_exc_label, _exc_type, _exc_hint = 'Error', 'error', ''` (`streaming.py:9608`), so even a correct classification hint was dropped for unhandled types. The new `elif _exc_is_credential_pool_empty` branch surfaces it. ✓

### Traced against upstream hermes-agent (fresh tarball) — the interesting part

- The literal `All N credential(s) exhausted for <provider>` string exists in **neither** current repo — it came from the reporter's agent build. On the **current** agent, a fully-exhausted pool has `mark_exhausted_and_rotate()` return `None` (`agent/agent_runtime_helpers.py:791-849`) and the underlying provider error (429/402/401) propagates instead. That makes the PR's third match clause (`'credential' in err and 'exhausted' in err`) the load-bearing robustness choice across agent versions — the exact phrasing isn't stable upstream. ✓
- **Over-match check (the real risk of a broad clause):** I read the current agent's account-quota messages (`agent/conversation_loop.py:205-240`) — "billing, credits, or account entitlement is exhausted", "Claude subscription usage is exhausted" — **none contain the word "credential"**, so genuine account-quota errors can't be misclassified as pool-empty. The test's real-quota-still-quota case guards this too. ✓
- **Auth interplay:** `_is_auth` now excludes `_is_credential_pool_empty` — so a pool-exhaustion message that happens to mention a 401 can't be misrouted into the `auth_mismatch` **credential self-heal retry** path (`:9497+`), which would have been actively wrong for an empty pool (nothing to heal into). Correct precedence. ✓
- Classification precedence: pool-empty returns before the quota branch — right call, since a zero-key pool is a config problem, not account credits; the hint (add/refresh a key or `hermes model`) matches what the CLI's `hermes auth add` flow expects. ✓

### Edge-case matrix

| Error text | Classification |
|---|---|
| `All 0 credential(s) exhausted for opencode_go` | `credential_pool_empty` + hint ✅ (was bare "Error") |
| `credentials exhausted` (variant phrasing) | `credential_pool_empty` ✅ |
| "billing, credits, or account entitlement is exhausted…" (current agent quota) | `quota_exhausted` (no "credential" in text) ✅ |
| `credits exhausted` | `quota_exhausted` (unchanged) ✅ |
| Pool-empty message containing "401" | `credential_pool_empty`, NOT auth self-heal ✅ |
| Rate-limit / auth / not-found / cancelled | unchanged branches ✅ |

### Tests

- `tests/test_issue3929_credential_pool_classification.py` — **3/3** (distinct classification; not misclassified as account-quota; real quota still classifies as quota).
- **Fail-without-fix verified independently:** against master's `streaming.py`, `test_credential_pool_exhaustion_classified_distinctly` fails — load-bearing. ✓
- Classification/quota/credential/rate-limit blast-radius — **243 passed, 5 skipped**.
- **Full suite: 11807 passed / 1 failed** — the single failure (`test_issue4536_service_tier` roundtrip) **passes in isolation on this branch**, has zero file overlap with this diff (streaming classification vs. service-tier endpoint), and PR CI is fully green: a pre-existing order-dependent flake, not a regression.

- The broad third clause means any error containing both "credential" and "exhausted" classifies as pool-empty. Having checked every current-agent error text that could flow here, nothing legitimate collides today; if a future provider message ever says e.g. "credential quota exhausted" for account credits, precedence would favor pool-empty — the test suite's over-match guard is the right place to catch that if it arises.
- Scope discipline is good: the data-loss half of #3929 (journaled-reasoning restore, gated on the #5024 provider-policy design) is properly deferred rather than rushed.

### Recommendation

A tight, correctly-scoped classification fix with both master defects verified by inspection, a mutation-proven regression test, and — the part I dug into — sound cross-version robustness: the matcher survives the fact that the triggering string isn't emitted by the current agent, without over-matching any genuine account-quota text, and without misrouting into the 401 self-heal path. Ship.

nesquena-hermes:
Shipped in **v0.51.870**.

Credential-pool exhaustion (#3929, classification half) now surfaces an actionable message instead of a bare "Error": the `All 0 credential(s) exhausted for <provider>` case is classified distinctly (ordered before the quota check) with a pool-specific hint, without swallowing genuine quota errors or blocking partial-result harvesting.

The remaining half of #3929 — recovering partial reasoning/output on the reload-reconcile path — stays a scoped follow-up that needs the reporter's server logs.

Gate: Codex SAFE TO SHIP (no over-match, ordering doesn't reclassify existing errors); full suite 12029 passed. Merged via #5600.

Issue #5096, opened 2026-06-28
fix(session): rewind/fork leaves stale model context (branch, truncate, edit index)


claw-io:
## Summary

Several session-mutation paths can leave the **model-facing** history (`context_messages`, `state.db` merge, or wrong truncate indices) **out of sync** with what the user sees after **fork**, **edit**, or **regenerate**. Symptom: user rewinds or forks the visible transcript, but the agent on the **next turn** still behaves as if **later turns** (after the rewind/fork point) are in context.

This issue bundles related **bugs**. Each should be fixable in a focused PR; they are filed together because they share one user-visible failure mode.

## Bug A — Fork copies full parent `context_messages`

**Reported behavior:** “Fork from here” shows a short transcript in the new session, but the agent answers using topics or details from **messages after the fork point** in the parent.

**Root cause (code):** `POST /api/session/branch` sets:

- `messages` ← `source_messages[:keep_count]` (correct prefix)
- `context_messages` ← `copy.deepcopy(source.context_messages)` (**entire parent model context**, not truncated to fork)

On send, `_context_messages_for_new_turn()` prefers `context_messages` over `messages` (`api/streaming.py`, `_session_context_messages`).

**Expected:** New branch session’s `context_messages` (and related `context_engine_state` if needed) should reflect **only** the fork prefix — same semantic cutoff as `forked_messages`, including user-turn-aware truncation when display and context lengths differ.

**Regression:** Integration test: branch with `keep_count=N`, send a turn, assert agent history length / last user content does not include parent-only tail (mock or inspect `_context_messages_for_new_turn` output).

**Related:** #465 (fork feature); not a duplicate — fork UI works, context inheritance does not.

## Bug B — Edit / regenerate use window-relative `keep_count`

**Reported behavior:** On long sessions with “Load earlier messages” (`_messagesTruncated`, `_oldestIdx > 0`), edit or regenerate removes the **wrong** suffix or leaves inconsistent transcript.

**Root cause (code):** `submitEdit` and `regenerateResponse` in `static/ui.js` POST `keep_count: msgIdx` from `data-msg-idx` **without** adding `_oldestIdx`. `forkFromMessage` was fixed in #2184 to use `absoluteKeepCount = _oldestIdx + msgIdx` before truncate/branch.

**Expected:** Edit and regenerate use the **same absolute index** as fork for `POST /api/session/truncate`.

[...]

PR #5563, opened 2026-07-04
fix(session): align forked context to display for large (compacted) s…

b3nw:
## Thinking Path

- Hermes WebUI aims for near 1:1 parity with the Hermes CLI, including forking a conversation from any point ("fork from here").
- A session keeps two parallel arrays: the visible `messages` transcript and the `context_messages` the model actually receives. In large sessions the latter is trimmed/compacted, so it is *shorter than and structurally divergent from* the display.
- Forking copies a display-indexed prefix of both and asks `truncate_context_for_display_keep` to cut the model context at the matching point so the two stay aligned.
- The bug: when `context_messages` was shorter than `messages` (every trimmed large session), the function short-circuited on `len(ctx) <= len(msgs)` and returned `ctx[:keep]` — slicing the *shorter* context at the *display* index. The matcher that exists precisely to align divergent arrays never ran.
- Result: the fork's visible transcript and the model's context described different points in the conversation, and the context often ended mid-turn on a dangling assistant `tool_use`. This is the "broken / disjointed fork."
- The fix routes the shorter-context case through the existing signature matcher so the cut lands on a real turn boundary — restoring the invariant in `docs/rfcs/webui-run-state-consistency-contract.md` that the model context must not "contradict what the user can see."

## What Changed

`api/session_ops.py` — `truncate_context_for_display_keep`:

1. Narrowed the naive-slice guard from `len(ctx) <= len(msgs)` to `len(ctx) == len(msgs)`. Only perfectly-parallel arrays may be sliced at the raw display index; both divergent directions now fall through to the signature matcher.
2. Added a shorter-context fallback for when both boundary rows are ambiguous/unmatched (the id/timestamp gap in large sessions): cut just past the last kept display row that resolved to a context index, preferring an exact match but accepting an ambiguous one (mirroring the sibling branches). Gated to `len(ctx) < len(msgs)` so the context-longer (#5096 summary-prefix) path is unchanged.
3. Reordered the `len(msgs) == 0` guard above the length check (no behavior change).

`tests/test_issue_branch_context_at_fork.py`:

- Corrected `test_..._prefers_compact_summary_fallback`: the old assertion `[compact, u1]` dropped `a1` (the assistant reply to a kept user turn) — see Contract Routing. New assertion `[compact, u1, a1]` matches the sibling `preserves_leading_compaction_row` case.
- Added `test_truncate_context_shorter_than_display_aligns_to_turn_boundary`, `test_shorter_context_ambiguous_boundary_keeps_forked_turn_via_weak_match`, and `test_shorter_context_zero_match_falls_back_to_best_effort_prefix`.

No new dependencies; no build-step/bundler/framework changes; server logic stays in `api/`.

## Why It Matters

Forking a large (trimmed/compacted) session produced a child whose model context was misaligned with its visible transcript and frequently malformed (dangling `tool_use`). Users saw the fork "continue from the wrong place" or behave erratically on the first message. Small sessions were unaffected (arrays equal length), which masked the bug. The measured live case (618 display / 600 context) now forks to the correct `context[:425]` (ending on a completed tool result) instead of the broken `context[:440]` (ending on a dangling `tool_use`, ~15 turns out of sync).

## Verification

- `./scripts/test.sh` full suite: **11,781 passed**, 133 skipped, 2 xfailed, 1 xpassed. The two failures are **pre-existing / unrelated** (confirmed): `test_gateway_lifecycle_controls…profile_scoped_agent_cli…` also fails on the base commit (sandbox resolves the agent CLI to `true` not `echo`), and `test_issue4536_service_tier…roundtrip…` passes in isolation on this branch (order-dependent provider-config-state flake). This diff touches no gateway/provider/config code.
- Targeted fork/branch/truncate/context/lineage/compression/materialize files all green (re-run after rebasing onto current `master`).
- Live-data replay of `truncate_context_for_display_keep` against the real parent session returns `context[:425]`, ends on a completed `tool` result, zero dangling tool_calls (was `context[:440]` with a dangling `call_0366`).
- Reviewed via multi-model, multi-angle adversarial code review; findings drove removal of an earlier redundant trim (see Risks) and the ambiguous-match refinement.

No browser/UI surface changed, so no before/after images apply.

## Contract Routing

- **Task type:** bug fix (model-context reconstruction on fork/truncate)
- **Touched areas:** `api/session_ops.py` `truncate_context_for_display_keep` / `truncate_session_at_keep`; product-semantics tests in `tests/test_issue_branch_context_at_fork.py`
- **Relevant public docs:** `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `docs/rfcs/webui-run-state-consistency-contract.md` (Runtime/durability/state family — model-context reconstruction), `docs/rfcs/stable-assistant-turn-anchors.md` (root-cause context)
- **Scope boundaries:** keeps the existing WebUI execution path; no adapter/runner/anchor implementation; no doc/contract redefinition.
- **Evidence:** green full suite; alignment regression tests on the shorter-context shape; live-data confirmation.

## Risks / Follow-ups

- Shorter-context + *zero-match* case (kept prefix entirely inside a summarized region with no signature overlap) still returns a best-effort `ctx[:keep]`; alignment is genuinely impossible there. Any residual dangling `tool_use` is made wire-safe at send (`_sanitize_messages_for_api` on the streaming path; the gateway path forwards no `tool_calls`/`tool` rows). Pinned by a regression test.
- Root cause is upstream: context rows lose their `id`/`timestamp`, forcing fuzzy signature matching that `stable-assistant-turn-anchors.md` explicitly warns against. Restoring stable per-message identity would make alignment exact. Tracked separately (144 affected sessions observed in a live store).
- An earlier revision added a `_drop_trailing_unsatisfied_tool_calls` trim; review showed it was redundant with the send-path sanitizer, missed Anthropic-native `tool_use` content blocks and partial multi-tool tails, and could empty the context (making the send path fall back to the *untrimmed* display messages). Removed.

## Release note

> Fixed forking a large (compacted) session producing a child whose model context was misaligned with the visible transcript and could start mid-tool-call.

- **Provider:** Anthropic
- **Model:** `claude-opus-4-8` (Claude Opus 4.8)
- **Mode/tools:** Claude Code agentic session; multi-agent adversarial code review (parallel finder subagents + verification) drove the removal of the redundant trim and the ambiguous-match refinement.

### DIFF
--- a/api/session_ops.py
+++ b/api/session_ops.py
@@ -110,10 +110,25 @@ def truncate_context_for_display_keep(
     msgs = full_messages if isinstance(full_messages, list) else []
     if not ctx:
         return []
-    if len(ctx) <= len(msgs):
-        return ctx[:keep]
     if len(msgs) == 0:
         return []
+    # Only the perfectly-parallel case (display and context row-for-row) can be
+    # sliced at the raw display index. When the two arrays differ in length —
+    # in EITHER direction — they have diverged and need alignment:
+    #   * context LONGER than display  → an injected summary/system prefix, etc.
+    #   * context SHORTER than display → large-session context trimming dropped
+    #     turns from the model context that the display still shows.
+    # The shorter-context case is the one that broke forked large sessions: the
+    # old ``len(ctx) <= len(msgs)`` guard short-circuited to ``ctx[:keep]``,
+    # slicing the shorter context at the display index (landing mid-turn, e.g.
+    # on an assistant tool_call whose result was past the cut). Fall through to
+    # the signature matcher for both divergent cases so the cut lands on a real
+    # turn boundary. Any residual dangling tool_use in the persisted context is
+    # made wire-safe on the send path (streaming: ``_sanitize_messages_for_api``
+    # strips unanswered tool_calls; gateway: it forwards no tool_calls/tool rows
+    # at all), so we do not re-do that trimming here.
+    if len(ctx) == len(msgs):
+        return ctx[:keep]
 
     def _row_signature(row: Any) -> tuple[str, ...] | None:
         if not isinstance(row, dict):
@@ -197,7 +212,27 @@ def truncate_context_for_display_keep(
                 return ctx[:ambiguous_first_unkept]
             return ctx[:last_kept + 1]
 
-    # Final fallback preserves #5096 behavior when alignment is unreliable.
+        # Both boundary rows were ambiguous/unmatched (common in large sessions
+        # where context rows have lost their id/timestamp so the matcher can't
+        # disambiguate structurally-identical rows). Only for the shorter-context
+        # case: cut just past the LAST display row in the kept prefix that
+        # resolved to a context index — preferring an exact match but accepting
+        # an ambiguous (weak) one, mirroring how the sibling branches above fold
+        # ``ambiguous_matches`` into the boundary. Accepting the weak match keeps
+        # the forked boundary turn's own context (often exactly that ambiguous
+        # row) instead of dropping back to an earlier exact match. It still errs
+        # toward UNDER-keeping rather than slicing at the raw display index,
+        # which would over-keep and mis-attribute later context rows to the kept
+        # display turns. The context-longer case (injected summary prefix) is
+        # left to the #5096 fallback below, which preserves that prefix.
+        if len(ctx) < len(msgs):
+            for i in range(keep - 1, -1, -1):
+                resolved = matches[i] if matches[i] is not None else ambiguous_matches[i]
+                if resolved is not None:
+                    return ctx[:resolved + 1]
+
+    # Final fallback preserves #5096 behavior when alignment is unreliable
+    # (no display row resolved to a context index, or keep >= len(msgs)).
     prefix_len = max(0, len(ctx) - len(msgs))
     prefix = ctx[:prefix_len]
     suffix = ctx[prefix_len:]
--- a/tests/test_issue_branch_context_at_fork.py
+++ b/tests/test_issue_branch_context_at_fork.py
@@ -107,6 +107,15 @@ def test_truncate_context_for_display_keep_drops_unkept_tool_rows_after_user_bou
 
 
 def test_truncate_context_for_display_keep_prefers_compact_summary_fallback():
+    # Context is SHORTER than display (a compaction summary replaced earlier
+    # turns) and the display has trailing turns beyond the kept prefix. Keeping
+    # display [u1, a1] must map to the context that represents [u1, a1] — the
+    # leading compaction row PLUS both u1 and a1. The old behaviour sliced the
+    # shorter context at the raw display index (returning [compact, u1]), which
+    # dropped a1 — the assistant reply to a kept user turn — from the model
+    # context while the display still showed it. That mismatch is the forked-
+    # large-session disjointedness bug; the result must now match the sibling
+    # ``preserves_leading_compaction_row`` case regardless of trailing turns.
     msgs = [
         {"role": "user", "content": "u1", "id": "u1", "timestamp": 1.0},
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (fork context alignment for compacted sessions)

**Certified head:** `sha:57fc513a` (clean rebase, branch `gate-rebase/5563-fork-context-align`) · **PR:** #5563 · _b3nw, fix(session): align forked context to display for large (compacted) sessions_
**Verdict:** Full gate GREEN. A well-reasoned data-correctness fix — forking a large/compacted session no longer slices the (shorter) model-context at the display index (which landed mid-turn); both divergent cases now route through the signature matcher so the fork cut lands on a real turn boundary. Codex SAFE, suite fully green.

### What I ran (rebased worktree `/tmp/wt-rebase-5563`)
| Gate | Result |
|------|--------|
| Full pytest suite | ✅ **11988 passed, 0 failed** |
| PR's own test | ✅ 11/11 (`test_issue_branch_context_at_fork.py`) |

### Findings
**✅ Correct alignment fix:** `truncate_context_for_display_keep` — the old `len(ctx) <= len(msgs)` guard short-circuited to `ctx[:keep]`, slicing the SHORTER model-context at the DISPLAY index (landing mid-turn, e.g. on an assistant tool_call whose result was past the cut) → forked compacted sessions got misaligned context. Fix narrows the fast-path to `len(ctx) == len(msgs)` (perfectly parallel) and falls through to the signature matcher for BOTH divergent directions (context longer = injected summary/prefix; context shorter = large-session trim dropped turns), so the cut lands on a real turn boundary. Ambiguous/weak-match fallback for large sessions where context rows lost id/timestamp errs toward UNDER-keeping (not a raw-display-index slice), preserving #5096 behavior for unreliable alignment. Any residual dangling tool_use is made wire-safe on the send path (`_sanitize_messages_for_api` strips unanswered tool_calls; gateway forwards none) — Codex confirmed no dangling tool_use escapes to the API. 11 branch-context tests + full suite green.

**Ready to merge — use branch `gate-rebase/5563-fork-context-align` (`sha:57fc513a`), NOT the PR's stale head `d5b357cd`.** A surgical, well-documented data-correctness fix (forked compacted sessions get context aligned to display on a real turn boundary, no mid-turn/dangling-tool slice), Codex SAFE + 11 targeted tests + full suite green (0 failures). Backend/logic — no visible surface. concept 4/5 (real fork-alignment correctness for large sessions). Author @b3nw (T2). crit=3, data-correctness.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; verified the fast-path narrowed to len(ctx)==len(msgs) with both divergent cases routed to the signature matcher (real turn-boundary cut), ambiguous-fallback errs under-keep (#5096 preserved), dangling tool_use wire-safe on send path (Codex), Codex SAFE + 11/11 + full suite green (0 failed). Cert valid for sha:57fc513a._

nesquena-hermes:
Shipped in **v0.51.861** — thanks @b3nw! 🎉

Your fix for fork context alignment on large/compacted sessions is live. Forking now aligns the cut to a real turn boundary in the same display coordinate space `/api/session` uses (handling both the compacted and non-compacted cases), so a fork no longer slices a turn in half or drops a tool result — erring toward keeping slightly less, with send-time sanitization as a backstop.

Gate summary before release:
- **Regression gate (Codex):** no real finding (the one flag was a stale-base artifact from an out-of-date staging branch; rebuilt fresh on `master`, and this change doesn't touch the clear path).
- **Architecture/correctness (Opus):** SHIP — correct, conservative, well-tested; the turn-boundary matcher handles both session shapes and never leaves an unsendable tool pair.
- **Full suite:** 12000 passed; your 11 fork tests all green.

Merged via release PR #5576 with your authorship preserved. Appreciate the clean, well-tested fix.

PR #5564, opened 2026-07-04
fix: mint stable per-message ids so large-session forks align

b3nw:
## Thinking Path

- Hermes WebUI keeps two parallel arrays per session: `messages` (the visible transcript) and `context_messages` (what is actually sent to the model). In large sessions the model context is trimmed/compacted, so it diverges from the display transcript.
- Forking ("fork from here" / `/api/session/branch`) and in-place truncation copy a *prefix of both* and rely on `truncate_context_for_display_keep()` to translate a display keep-index into the matching context index.
- That aligner needs a stable per-row key — it prefers `id`, then `timestamp` — to disambiguate rows. But `context_messages` rows carry **neither** (a live-store scan of 989 sessions found 0 with any `id`, and context timestamps near-absent regardless of recency), so it falls back to fragile content-signature matching.
- Large agent sessions are full of structurally-identical rows (repeated tool calls, empty-content assistant turns), so signature matching goes ambiguous, and for compacted sessions (`len(context) <= len(display)`) the aligner short-circuits to a raw display-index slice that lands mid-turn.
- Result: the fork's transcript ends on a clean turn while its model context ends on a dangling `tool_use` — the API rejects the next message or the model resumes from the wrong point ("disjointed" fork).
- Root cause is a data gap: both arrays derive from the same per-turn `result['messages']` dicts, but only the display array gets timestamps stamped; the model-facing array never carried a `timestamp` **or** an `id`.
- This PR closes the gap by minting a stable per-message `id` shared by both arrays (the key the aligner already prefers) and letting the aligner use it, so large-session forks align correctly.

## What Changed

- **`api/streaming.py`**
  - New `_assign_stable_message_ids()` — mints a monotonic, session-unique integer `id` on the per-turn result rows *after* the context restore and *before* both arrays are built, so the display and model-context copies of a logical row share the same id.
  - `_restore_reasoning_metadata()` now carries `id` forward across turns (exactly as it already carries `timestamp`), since the agent rebuilds id-less rows every turn.
  - Wired the mint call into all three streaming write-back commit sites (main turn, retry, self-heal).
- **`api/routes.py`** — same mint call on the runs/MoA `_handle_chat_sync` commit path.
- **`api/gateway_chat.py`** — mints ids on the two new rows of a gateway turn (shared with the display merge) so both arrays match.
- **`api/session_ops.py`** — `truncate_context_for_display_keep()`'s `len(ctx) <= len(msgs)` raw-slice short-circuit is now gated on the *absence* of ids (`_any_row_has_id`): id-bearing sessions get real alignment; legacy id-less sessions keep the exact prior behavior. The matcher's final fallback still returns `ctx[:keep]` if id matching cannot resolve a boundary.
- **Tests** — new `tests/test_context_message_stable_ids.py`; updated `tests/test_issue_branch_context_at_fork.py` and `tests/test_issue1217_transcript_compaction.py` for the corrected id-aware alignment.

The `id` is stripped before the provider API call (it is not in `_API_SAFE_MSG_KEYS`), same treatment as `timestamp`, so nothing new reaches the model provider.

## Why It Matters

Forking or truncating a large session currently produces a child whose model context is misaligned with its visible transcript — up to a dozen turns out of sync and often ending on a dangling tool-call that the provider API rejects. Giving each message a stable shared id makes alignment exact and immune to synthetic-timestamp drift and duplicate content, which is the durable fix (timestamps are synthesized per-array from `time.time()` and would not match across arrays even if both were stamped).

## Verification

- `./scripts/test.sh` full suite: **11866 passed**, 133 skipped. The one failure observed in a full run (`test_issue4536_service_tier`) is a pre-existing flake unrelated to this change — it passes in isolation and this PR does not touch service-tier/config code.
- New/updated tests prove the invariant directly: a logical row produced in a turn carries an **identical id in both arrays** (through the real turn transforms), and the aligner resolves a previously-ambiguous compacted-context boundary via that id, while id-less legacy data keeps the old shortcut.
- Root cause and scope were established by read-only inspection of a live store (specimen `20260611_190341_ddec1c`) plus a 989-session scan; details in the local investigation notes.
- Not a UI/UX change — no screenshots applicable.

## Risks / Follow-ups

- **Data-at-rest addition:** context/display rows now persist an integer `id` in the session sidecar. It is additive and ignored by older readers; the aligner treats a missing `id` as "fall back to signature/timestamp," so mixed old/new rows are safe.
- **Legacy sessions are not retro-aligned:** brand-new sessions get full id coverage from turn 1 (their `previous_context` is empty); pre-existing sessions gain ids only on new rows, and their id-less prefix rows age out of context over time. Legacy data still benefits from the id-independent safeguards recommended separately (aligner fallback + a trailing-tool-call guard).
- **Follow-up:** consider a trailing dangling-tool-call safety net after any context cut, which protects legacy sessions regardless of id availability.

## Contract Routing

- **Task type:** narrow bug fix on the existing WebUI execution path (model-context reconstruction + session metadata).
- **Touched areas:** per-turn context/display commit (`api/streaming.py`, `api/routes.py`, `api/gateway_chat.py`), fork/truncate alignment (`api/session_ops.py`), session-at-rest metadata (adds `id` to message rows).
- **Relevant public docs:** `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, and `docs/rfcs/webui-run-state-consistency-contract.md` (#2361 — coherence of transcript, model context, replay, compression, and session metadata).
- **Scope boundaries:** does not change SSE/streaming vocabulary, recovery/journal formats, UI rendering, or the provider-facing message shape (the `id` is stripped before the API call). No new dependencies; no build step.
- **Evidence needed before claiming done:** full test suite green on the focused diff; direct test of the shared-id invariant and id-resolved alignment; confirmation the `id` never reaches the provider.

## Contract Change

The consistency contract requires the visible transcript and the model context to stay coherent across fork/truncate. This PR intentionally changes one alignment behavior encoded in existing product-semantics tests:

- **Old rule:** when `context_messages` is no longer than `messages` (the compacted large-session shape), `truncate_context_for_display_keep()` slices context at the raw display index (`ctx[:keep]`) regardless of row identity.
- **New rule:** when both arrays carry stable ids, the aligner matches by id and cuts at the identity-correct boundary; the raw-index slice is used only when ids are absent (legacy data) or id matching cannot resolve the boundary.
- **Why justified:** the old rule is the documented mis-cut that makes large-session forks end on a dangling tool-call. The new rule is strictly more correct where ids exist and byte-for-byte identical where they do not. The corresponding tests (`test_issue_branch_context_at_fork.py`, `test_issue1217_transcript_compaction.py`) were updated in this PR to assert the corrected behavior rather than silently changing behavior via tests alone.

## Release note

Fixed large-session fork/truncate producing a model context misaligned with the visible transcript (and occasional dangling tool-call API rejections) by giving each message a stable id shared between the display and model-context arrays.

## Model Used

AI-assisted.

- **Provider:** Anthropic (via Claude Code)
- **Exact model:** Claude Opus 4.8 (`claude-opus-4-8`)
- **Notable mode/tool use:** agentic investigation with a subagent for codebase tracing and read-only inspection of a live session store to establish root cause and scope before implementing.

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -843,6 +843,19 @@ def _run_gateway_chat_streaming(
             if saved_reasoning:
                 assistant_msg["reasoning"] = saved_reasoning
             previous_context = list(getattr(s, "context_messages", None) or getattr(s, "messages", None) or [])
+            # Stamp stable ids on the two new rows (shared with the display merge
+            # below) so display and model-context copies share an id for the
+            # fork/truncate aligner (#context-message-stable-id).
+            try:
+                from api.streaming import _assign_stable_message_ids
+
+                _assign_stable_message_ids(
+                    [user_msg, assistant_msg],
+                    previous_context,
+                    list(getattr(s, "messages", None) or []),
+                )
+            except Exception:
+                logger.debug("Failed to stamp stable ids on gateway turn rows", exc_info=True)
             s.context_messages = previous_context + [user_msg, assistant_msg]
             try:
                 from api.streaming import _is_context_compression_marker
--- a/api/routes.py
+++ b/api/routes.py
@@ -20193,6 +20193,7 @@ def _handle_chat_sync(handler, body):
             )
             from api.streaming import (
                 _WEBUI_PROGRESS_PROMPT,
+                _assign_stable_message_ids,
                 _dedupe_replayed_context_messages,
                 _merge_display_messages_after_agent_result,
                 _restore_display_reasoning_metadata,
@@ -20261,6 +20262,9 @@ def _handle_chat_sync(handler, body):
             _previous_context_messages,
             _next_context_messages,
             msg,
+        )
+        _assign_stable_message_ids(
+            _result_messages, _previous_messages, _previous_context_messages
         )
         s.context_messages = _next_context_messages
         s.messages = _merge_display_messages_after_agent_result(
--- a/api/session_ops.py
+++ b/api/session_ops.py
@@ -98,6 +98,14 @@ def _truncation_watermark_for(messages):
         return 0.0
 
 
+def _any_row_has_id(rows) -> bool:
+    """Return True if any row carries a non-null stable ``id``."""
+    for row in rows or []:
+        if isinstance(row, dict) and row.get('id') is not None:
+            return True
+    return False
+
+
 def truncate_context_for_display_keep(
     context_messages: list | None,
     full_messages: list | None,
@@ -110,7 +118,15 @@ def truncate_context_for_display_keep(
     msgs = full_messages if isinstance(full_messages, list) else []
     if not ctx:
         return []
-    if len(ctx) <= len(msgs):
+    # A context array no longer than the display array (the large/compacted
+    # session signature) historically could not be aligned, so it was sliced at
+    # the raw display index — the documented fork/truncate mis-cut. With stable
+    # per-message ids (#context-message-stable-id) the matcher below CAN align a
+    # shorter/compacted context, so only take the raw-index shortcut when neither
+    # array carries ids to align on. Legacy id-less sessions keep the old
+    # behavior exactly (no regression); the matcher's final fallback still
+    # returns ctx[:keep] if id matching cannot resolve the boundary.
+    if len(ctx) <= len(msgs) and not (_any_row_has_id(ctx) and _any_row_has_id(msgs)):
         return ctx[:keep]
     if len(msgs) == 0:
         return []
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -4170,6 +4170,46 @@ def _deduplicate_context_messages(messages):
     return deduped
 
 
+def _assign_stable_message_ids(result_messages, *existing_arrays):
+    """Mint a stable, session-unique integer ``id`` on model-result rows lacking one.
+
+    Both ``messages`` (display transcript) and ``context_messages`` (model-facing
+    history) are derived from the *same* per-turn ``result['messages']`` dicts.
+    Stamping the id on those shared dicts here makes each logical row carry an
+    identical id in both arrays, so the fork/truncate aligner
+    (``session_ops.truncate_context_for_display_keep``) can match rows by its
+    preferred ``id`` key instead of the fragile content-signature fallback that
+    goes blind on large sessions full of structurally-identical rows.
+
+    Ids are monotonic within a session (max existing id + 1, seeded from the
+    result rows plus any ``existing_arrays`` passed for collision-avoidance).
+    Across sessions ids may repeat, which is safe: alignment only ever compares
+    rows within one session, and a fork copies both arrays together so the child
+    stays internally consistent. Rows whose historical id was carried forward by
+    ``_restore_reasoning_metadata`` keep it; only genuinely new rows are minted.
+
+    Returns the number of rows newly stamped. Mutates ``result_messages`` in place.
+    """
+    if not result_messages:
+        return 0
+    seed = 0
+    for arr in (result_messages, *existing_arrays):
+        for m in arr or []:
+            if isinstance(m, dict):
+                mid = m.get('id')
+                # bool is an int subclass; exclude it so a stray True/False id
+                # can never seed the counter.
+                if isinstance(mid, int) and not isinstance(mid, bool) and mid > seed:
+                    seed = mid
+    stamped = 0
+    for m in result_messages:
+        if isinstance(m, dict) and m.get('id') is None:
+            seed += 1
+            m['id'] = seed
+            stamped += 1
+    return stamped
+
+
 def _prune_context_tool_results_after_compression(agent, context_messages):
     """Run the active compressor's cheap tool-result pruning on model context.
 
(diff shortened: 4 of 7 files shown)


nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (stable per-message ids — root-cause fork alignment) · ⚠️ SUPERSEDES #5563 (reconcile before merge)

**Certified head:** `sha:64bf6356` (clean rebase, branch `gate-rebase/5564-stable-message-ids`) · **PR:** #5564 · _b3nw, fix: mint stable per-message ids so large-session forks align_
**Verdict:** Full gate GREEN. The root-cause fix for the fork/truncate mis-cut — mints stable session-unique per-message ids so context/display rows align on REAL ids (not just a signature heuristic). Codex SAFE, suite fully green. **This overlaps #5563 (same author, same guard) — #5564 is the deeper fix and should supersede #5563; they cannot both merge (conflicting `truncate_context_for_display_keep` guards).**

### What I ran (rebased worktree `/tmp/wt-rebase-5564`)
| Gate | Result |
|------|--------|
| Full pytest suite | ✅ **12005 passed, 0 failed** |
| stable-id/fork/compaction tests | ✅ 47 passed |

### Findings
**✅ Sound stable-id minting:** `_assign_stable_message_ids(result_messages, *existing_arrays)` mints a session-unique integer id on model-result rows lacking one, carries existing ids forward (rows kept by `_restore_reasoning_metadata` keep their id — only genuinely-new rows minted, so ids don't drift/re-mint each turn), seeds the counter past the existing max (no collision), and stamps AFTER the context so context+display share ids. The `session_ops.py` guard becomes `if len(ctx) <= len(msgs) and not (_any_row_has_id(ctx) and _any_row_has_id(msgs))` — raw-index fast-path ONLY for legacy no-id sessions; id-bearing sessions fall to the matcher and align on real ids. `gateway_chat.py` + `streaming.py` stamp consistently. Codex confirmed ids are stable/session-unique, carry through compaction/reasoning-restore, and don't leak to the wire in a provider-breaking way. 47 targeted + full suite green (0 failures).

**⚠️ SUPERSEDES #5563 (reconcile before merge):** #5563 (also b3nw, GREEN on `gate-rebase/5563-fork-context-align`) fixed the SAME `truncate_context_for_display_keep` mis-cut with a signature-matcher-only approach (narrowed the guard to `len(ctx)==len(msgs)`). #5564 is the deeper root-cause version (mint real ids + id-aware guard). They edit the same guard differently and **cannot both merge**. Recommend: merge #5564 (root-cause) and close #5563 as superseded (crediting b3nw), OR if #5563 already merged, #5564 must be rebased on top and its guard reconciled. Flag to Nathan for the merge-order decision.

**Ready to merge — use branch `gate-rebase/5564-stable-message-ids` (`sha:64bf6356`), NOT the PR's stale head `746822a5` — BUT reconcile with #5563 first.** #5564 is the correct root-cause fix (stable ids make fork alignment deterministic rather than heuristic); it should supersede #5563. Both are by b3nw and GREEN; pick #5564 and close #5563 as superseded (credit b3nw on both), or merge #5564 and drop the #5563 stage. Codex SAFE + 47 targeted + full suite green (0 failures), backend/data-correctness. concept 4/5 (root-cause fork-alignment). Author @b3nw (T2). crit=3, data-correctness.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; verified stable-id minting (session-unique, carried-forward not re-minted, counter-seeded no collision, id-aware guard keeps legacy raw-path + aligns id-bearing sessions, consistent gateway+streaming stamping), Codex SAFE + 47 targeted + full suite green (0 failed). OVERLAPS/SUPERSEDES #5563 (same author/guard) — reconcile merge-order. Cert valid for sha:64bf6356._

nesquena-hermes:
Shipped in **v0.51.867** — thanks @b3nw! 🎉

Your root-cause fix for large-session fork/truncate misalignment is live. Each message now gets a stable id shared between the visible transcript and the model-context arrays, so forking or truncating a big (compacted) session aligns exactly instead of cutting mid-turn and ending on a dangling tool-call that the provider rejects. The id is additive, stripped before it reaches the provider, and legacy id-less sessions keep working.

This pairs with #5563 (shipped in v0.51.861), which made the aligner *prefer* the id — your PR closes the data gap by actually minting it. Since #5563 landed first, I reconciled your PR down to just the id-minting (its `session_ops.py` change was already delivered by #5563) and rebuilt on current master.

Gate — 3 convergent rounds (the dual gate earned its keep here):
- **Codex r1** caught eager save-mode leaving the display row id-less while context had the id → fixed (carry the id onto the eager checkpoint).
- **Codex r2** caught the same class in the stale-user repaired-boundary deep-copy path → root-fixed by minting *before* the dedupe deep-copy at all four commit sites.
- **Codex r3: SAFE TO SHIP.** **Opus: SHIP** — traced every persistence reader (replay/compression/recovery/import): additive + backward/forward compatible; monotonic counter collision-free; divergent edges err toward under-keeping, never mis-cut.
- **Full suite:** 12023 passed; touched-file tests 83/83.

Merged via release PR #5595 with your authorship preserved (`Co-authored-by`). Excellent root-cause writeup and fix.

Issue #5525, opened 2026-07-04
Mermaid diagrams still broken on mobile


lurebat:
#5413 is marked as solved, and while it is solved on desktop, it didn't fix it on mobile - there the height is still 0 and the diagrams don't show.

Also, the icons for full screen and fit to screen didn't change and still look identical.

nesquena-hermes:
Thanks — you're right that the recent Mermaid work (#5413 → shipped via #5434/#5452) targeted inline sizing + the fullscreen lightbox on desktop, and a mobile-specific `height: 0` path wasn't covered. Reopening this as its own mobile bug rather than folding it back into the closed #5413.

Two concrete things that would help pin it:

1. **The `height: 0` on mobile** — this reads like the inline diagram container measuring a zero/again-zero width at render time on a touch viewport (the inline sizing derives height from the measured width, so a `clientWidth === 0` at the wrong frame collapses it). If you can, open the diagram's element in mobile devtools and note whether the `<svg>`/container has a computed `width` at all, or is `0×0`.
2. **The fullscreen / fit-to-screen icons "look identical"** — can you confirm whether tapping them does *anything* on mobile (does fullscreen open at all, does fit re-scale), or are they inert? That separates "wrong icon glyph" from "control doesn't fire on touch."

Also: which mobile browser/OS (iOS Safari, Android Chrome, the installed PWA?) and roughly the viewport width. Labeling `bug` + `mobile` + `renderer`. This is a genuine gap in the earlier fix, not a dup — I'll trace the inline mobile sizing path against current master.

AsafMah:
1. Intrestingly enough - it seems to be fixed in my regular browser (edge), but it still happens on the hermes-webui app (https://github.com/hermes-webui/hermes-android) which should just be a host for a webview.
2. They do do different things - the right buttons opens a "full screen" view, and the left one presumably adjusts the size, but because the height is 0, the full screen view shows nothing except dimming the screen, and the adjust button does nothing.

nesquena-hermes:
Investigated both halves of this on current master (v0.51.856) — one reproduces, one doesn't.

[...]

PR #5565, opened 2026-07-04
fix(#5525): distinct Mermaid fullscreen toolbar icon (mobile height:0 not reproducible)

nesquena-hermes:
## Summary

Addresses the **icon half** of #5525: the Mermaid toolbar's "fit to screen" and "fullscreen" buttons render **identical glyphs**, so users can't tell the two controls apart.

Root cause: the old `fullscreen` SVG path was `fit`'s four corner-brackets **drawn twice** (`...M4 15v5h5M20 15v5h-5M9 4H4v5M15 4h5v5...`) — pixel-identical to `fit`.

Fix (`_mermaidViewerIcon` in `static/ui.js`): `fullscreen` keeps the corner frame but adds **outward diagonal expand arrows** (the standard "expand to fullscreen" affordance), clearly distinct from `fit`'s corner-brackets-only glyph at both large and 14px toolbar sizes.

Confirmed live on a mobile viewport that the two icons were visually identical before the change.

## The other half of #5525 — mobile height:0 — does NOT reproduce on current master

I verified this **live myself** (not from the report), because the timeline was suspicious:

- On a **390×844 mobile viewport**, a **multi-turn conversation** with two real diagrams (a `graph TD` flowchart + a `sequenceDiagram`): both render at proper heights (**641px** and **287px**), fully visible, legible, not collapsed. Flowchart vision-confirmed crisp and fully contained in the column.
- The `_MERMAID_VIEWER_INLINE_MIN_HEIGHT = 220` floor + `window.innerWidth || box.width` zero-width fallback (in `_inlineHeight` / `_viewportSize`) already guard the `clientWidth==0`-at-mount case that would produce height:0.
- **Timeline:** that floor shipped **2026-07-02 (v0.51.834)** — two days before #5525 was filed (2026-07-04). The reporter was almost certainly on an older or cached build.

Handled with an issue comment asking the reporter to hard-reload + confirm their build/browser. This PR does **not** touch the (already-correct) mobile sizing to avoid regressing the #5434 desktop/inline behavior.

## Tests
`tests/test_issue5525_mermaid_icon_distinct.py` — asserts `fit != fullscreen` and that `fullscreen` carries the diagonal expand-arrow strokes `fit` lacks. Fail-without-fix verified (fails on `origin/master` where the glyphs are identical).

Refs #5525 (icon fix; mobile-height half not reproducible on current master).

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -1421,7 +1421,7 @@ function _mermaidViewerIcon(kind) {
     zoomOut: '<svg viewBox="0 0 24 24" aria-hidden="true"><circle cx="10" cy="10" r="6"></circle><path d="M7 10h6"></path><path d="M15 15l4 4"></path></svg>',
     reset: '<svg viewBox="0 0 24 24" aria-hidden="true"><path d="M6 9V4H1"></path><path d="M1 4l4 4"></path><path d="M10 4a8 8 0 1 1-5.66 13.66"></path></svg>',
     fit: '<svg viewBox="0 0 24 24" aria-hidden="true"><path d="M4 9V4h5M20 9V4h-5M4 15v5h5M20 15v5h-5"></path></svg>',
-    fullscreen: '<svg viewBox="0 0 24 24" aria-hidden="true"><path d="M4 9V4h5M20 9V4h-5M4 15v5h5M20 15v5h-5M9 4H4v5M15 4h5v5M9 20H4v-5M15 20h5v-5"></path></svg>',
+    fullscreen: '<svg viewBox="0 0 24 24" aria-hidden="true"><path d="M4 9V4h5M20 9V4h-5M4 15v5h5M20 15v5h-5"></path><path d="M4 4l5 5M20 4l-5 5M4 20l5-5M20 20l-5-5"></path></svg>',
   };
   return icons[kind] || '';
 }
--- a/tests/test_issue5525_mermaid_icon_distinct.py
+++ b/tests/test_issue5525_mermaid_icon_distinct.py
@@ -1,0 +1,50 @@
+"""Regression coverage for #5525 — Mermaid toolbar fit/fullscreen icons must differ.
+
+Reported alongside the (already-fixed on master) mobile height:0 issue: the
+"fit to screen" and "fullscreen" toolbar icons look identical. Confirmed by
+inspection — the old `fullscreen` glyph was `fit`'s four corner-brackets path
+drawn twice (`...M9 4H4v5M15 4h5v5...`), so it rendered pixel-identical to `fit`.
+
+Fix: give `fullscreen` a distinct glyph (corner frame + outward diagonal expand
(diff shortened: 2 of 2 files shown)


nesquena (approved):

Independent review of the icon half of #5525: the Mermaid toolbar's "fit to screen" and "fullscreen" buttons rendered **identical glyphs**.

### Root cause — verified in the diff

I confirmed the PR's claim by tracing the SVG paths: the old `fullscreen` path was `fit`'s four corner-brackets **plus the same four brackets redrawn from the opposite endpoints** — e.g. `M9 4H4v5` traces exactly the segments of fit's `M4 9V4h5`, just in the reverse direction. Pixel-identical output, so the two buttons were indistinguishable. The fix keeps the corner frame and adds four diagonal strokes (`M4 4l5 5 …`), giving the standard expand-arrows affordance — clearly distinct from `fit` at toolbar size. ✓

### Scope discipline — the mobile-height half

The PR correctly does **not** touch mobile sizing: the `_MERMAID_VIEWER_INLINE_MIN_HEIGHT = 220` floor + zero-width viewport fallback shipped in v0.51.834 (2026-07-02), two days **before** #5525 was filed, and the author live-verified diagrams render at proper heights (641px/287px) on a 390×844 viewport on current master. Leaving the (already-correct) sizing alone avoids regressing the #5434 desktop/inline behavior — the right call, handled via an issue comment asking the reporter to confirm their build. ✓

### Audit

- Static, hardcoded SVG strings in `_mermaidViewerIcon` (`aria-hidden`, no user data) — no XSS surface. ✓
- One-line change to one icon entry; `fit`/`zoom`/`reset` untouched. ✓
- `node --check static/ui.js` clean. ✓

### Tests

- `tests/test_issue5525_mermaid_icon_distinct.py` — **2/2** (`fit != fullscreen`, and fullscreen carries the diagonal strokes fit lacks).
- **Fail-without-fix verified independently:** I ran the new test against master's `ui.js` — `test_fullscreen_icon_has_distinct_expand_arrows` **fails** there, so the test is load-bearing, not vacuous.
- Mermaid blast-radius — **32 passed, 4 skipped**. CI fully green.

### Recommendation

A precise one-line fix with a verified root cause (the old glyph really was fit drawn twice), a mutation-proven regression test, and correct scope discipline on the not-reproducible mobile-height half. Ship.

nesquena-hermes:
Shipped in **v0.51.871**.

The Mermaid diagram toolbar's "Fit to screen" and "Fullscreen" buttons are now distinct icons (they were byte-identical before — fullscreen was fit's corner brackets drawn twice). Fullscreen is now a corner-frame + diagonal-expand-arrows glyph. Verified live in-browser and visual-approved.

Gate: Codex SAFE, full suite 12047 passed, 2 own tests. The other half of #5525 (mermaid height:0 on mobile) does not reproduce on current master (verified live at 390×844). Merged via #5603.

PR #2480, opened 2026-05-17
fix: align fork-from-here with merged messaging history

Michaelyklam:
## Thinking Path
- “Fork from here” should copy the visible transcript prefix up to the selected message.
- Messaging/Gateway sessions can display a merged transcript from Agent/CLI history plus the WebUI sidecar, while `/api/session/branch` previously sliced only the sidecar list.
- That made frontend `keep_count` values from the merged display coordinate space too large for the sidecar list, so Python slicing silently copied the whole source.
- The fix keeps the frontend keep-count coordinate system intact and teaches the branch handler to slice the same merged display list the user clicked on.

## What Changed
- Extracted the messaging transcript merge/dedupe logic into `_merged_session_messages_for_display(...)` and reused it from both `GET /api/session` and `POST /api/session/branch`.
- Made branch creation best-effort save the source session before slicing so recent undo/retry state is coherent with what gets forked.
- Snapshotted the source session id in `forkFromMessage(...)` across the async full-history load so a fast sidebar switch cannot fork the wrong session.
- After loading the new fork, the frontend now loads the full fork transcript so long forks do not initially appear to start mid-conversation.
- Added regression coverage for merged messaging coordinate-space slicing, source save-before-slice, async session-id guarding, and the full-fork reload path.

## Why It Matters
Users clicking “Fork from here” on a messaging/CLI-backed session should get a prefix ending at the clicked message, not a full copy containing later turns or a fork from a different session after a fast navigation race.

## Verification
- `env -u HERMES_CONFIG_PATH -u HERMES_WEBUI_HOST /home/michael/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_issue2472_fork_from_here_messaging.py tests/test_issue2184_fork_from_here_absolute_index.py tests/test_session_import_cli_fallback_model.py tests/test_v050253_opus_followups.py -q` — 25 passed
- `/home/michael/.hermes/hermes-agent/venv/bin/python -m py_compile api/routes.py`
- `node --check static/commands.js`
- `git diff --check`
- `env -u HERMES_CONFIG_PATH -u HERMES_WEBUI_HOST /home/michael/.hermes/hermes-agent/venv/bin/python -m pytest tests/ -q` — 5871 passed, 6 skipped, 3 xpassed, 8 subtests passed

## Risks / Follow-ups
- This uses the existing timestamp/role/content merge ordering, now shared by branch creation as well as session display; unusual missing timestamps will preserve the pre-existing display behavior rather than introducing a new ordering scheme.
[...]

PR #5576, opened 2026-07-04
release #5563: align forked context to display for large/compacted sessions

nesquena-hermes:
## Release: #5563 — align forked context to display for large/compacted sessions

Ships @b3nw's fix for a fork data-correctness bug, rebuilt on current `master` and re-gated.

### The bug (#5563)
"Fork from here" on a **large / context-compacted** session could cut the copied context off a real turn boundary — slicing mid-turn or right after a tool call before its result — so the forked conversation started with a malformed context the model choked on.

### The fix
`api/session_ops.py` aligns the fork cut to a resolved turn boundary using a signature matcher that handles both the compacted (compression-anchor present) and non-compacted large-session display coordinate spaces, consistent with the merged-transcript space `/api/session` exposes (#2480). Errs toward under-keeping, with send-time sanitization as a backstop. Small non-compacted forks are unchanged.

### Gate (all green)
- **Codex** (regression/breakage): the sole CORE finding was a stale-base artifact (a prior stage was behind master and appeared to revert the #5532 clear fix) — this branch is rebuilt on current master and #5563 does not touch `api/routes.py`. No real finding.
- **Opus** (architecture/correctness): **SHIP** — correct, conservative, well-tested; two optional perf follow-ups (signature memoization), not gate blockers.
- **Full suite:** 12000 passed, 0 failed. Fork tests: 11/11.

Attribution: original author @b3nw (`Co-authored-by` trailer preserved on the fix commit).

Closes #5563


nesquena:
## Post-merge verification ✅ (this merged mid-review-sweep; verified on master rather than formally reviewed)

The release agent merged this while my batch sweep was in flight, so instead of a moot formal review I verified the landed fix on merged master (`627c6a17`):

- **Root cause matches the fix:** the old `len(ctx) <= len(msgs)` guard short-circuited the *shorter-context* case (every trimmed large session) to `ctx[:keep]` — slicing the shorter model context at the **display** index, landing mid-turn/dangling `tool_use`. The change to `len(ctx) == len(msgs)` routes both divergent directions through the signature matcher; only the perfectly-parallel case takes the raw-index slice. ✓
- **The new ambiguous-boundary fallback errs the right way:** backward scan for the last resolved (exact-or-weak) display row, cutting at `ctx[:resolved+1]` — under-keeping rather than mis-attributing later context rows; the context-longer (injected-prefix) case still falls to the preserved #5096 fallback. ✓
- **Backstop claim verified:** `_sanitize_messages_for_api` exists on the streaming send path (strips unanswered tool_calls), so a residual dangling `tool_use` in persisted context stays wire-safe. ✓
- **Tests on merged master:** `tests/test_issue_branch_context_at_fork.py` **11/11**; fork/branch/truncate blast-radius **66 passed, 4 skipped**.

No issues found — the landed change is correct and conservative. Nice fix, @b3nw + agent.

PR #5579, opened 2026-07-04
fix(#5578): stop login next= param from nesting into itself and exploding the URL

### LINKED ISSUE #5578: bug(auth/login): login redirect `next=` param nests into itself and re-encodes, exploding the URL to ~12k chars until the tab breaks
## Symptom

A long-lived WebUI browser tab occasionally "stops working," and the URL has ballooned to an enormous length — a reporter measured **~12,000 characters** after copy-pasting it out. Pasting it into an editor shows the address is a deeply nested, repeatedly re-encoded login-redirect chain:

```
/session/login?next=/session/login%3Fnext%3D/session/login%253Fnext%253D/session/login%25253Fnext%25253D…
```

Each level adds another `/session/login?next=` and re-percent-encodes the previous one (`?` → `%3F` → `%253F` → `%25253F` → …), so the URL grows **exponentially** on each auth bounce until it exceeds the browser's URL length limit and the tab breaks. Reported to happen intermittently on a tab left open for a long time (an auth session expiring while the tab is idle is the trigger).

## Root cause (code-grounded, current head)

The `next=` redirect param accumulates the **login page's own URL** into itself. Three sites each guard against open-redirect but **none reject a `next` that points back at the login page**, so nothing breaks the self-referential nesting:

1. **`static/workspace.js:49`** — on a `401`, the central `api()` fetch wrapper does:
   ```js
   if(redirect401) window.location.href='login?next='+encodeURIComponent(window.location.pathname+window.location.search);
   ```
   If the tab is **already on** `/session/login?next=X` when a background `api()` call 401s (or it lands back there), `window.location.pathname+window.location.search` already contains `?next=X`. That whole thing gets `encodeURIComponent`'d and wrapped into a new `next=`, one level deeper, re-encoded. Repeat on each bounce → exponential growth.

2. **`static/login.js:28` `_safeNextPath()`** — validates path-absolute, rejects `//`, `\`, and control chars, but does **not** reject `next` values whose path is itself `/session/login`. So it faithfully preserves the nested chain instead of collapsing it.

3. **`api/routes.py:9231` `_safe_login_redirect_path()`** — same guard set server-side (leading `/`, no `//`/`\`, no control/whitespace), also **no self-reference check**. It re-emits the nested `next` via `quote(next_path, safe="/")` at `routes.py:9264` (OIDC start href) and the login-page render paths (`routes.py:11154/11198`), each pass adding an encoding layer.

Net: an expired-auth 401 on the login page (or any redirect that captures a URL that already carries a login `next`) feeds the login redirect its own address, and every guard treats the nested `/session/login?next=…` as a "safe path" and passes it through re-encoded.

## Suggested fix shape (surgical, both layers — confirm against head)

Collapse/reject a `next` that targets the login page itself, in all three guards, so the chain can never nest:

- In **`_safeNextPath()`** (`login.js`) and **`_safe_login_redirect_path()`** (`api/routes.py`): after the existing validation, parse the path portion of `next`; if it resolves to the login route (`/session/login`, or the bare `login` subpath-mount form), return the default (`'./'` / `'/'`) instead of the nested value. This single check kills the recursion at read time.
- In **`workspace.js:49`**: before building `login?next=…`, if the current page **is** the login page, don't append a `next` at all (or append the *inner* `next`, not the whole login URL). Guard: skip the `next` param when `window.location.pathname` already ends in the login route.
- Defense in depth: cap the accepted `next` length (reject anything over a few hundred chars) and reject a `next` that itself contains an (encoded or raw) `next=` / `%3Fnext` substring — a cheap belt-and-suspenders against any missed nesting path.

Fair test: reproduce by loading `/session/login?next=/session/login%3Fnext%3D/` and letting an `api()` call 401, then assert the resulting URL does not grow / does not contain a nested `/session/login`.

## Scope

Auth/login redirect handling only. Preserve the existing open-redirect protections (protocol-relative, absolute-URL, backslash, control-char rejection) — this only ADDS a self-reference/loop guard. Don't change the happy-path redirect (a legitimate `next=/some/real/path` must still work).

## Not a duplicate of

- #5001 / #5021 (closed) — post-upgrade `401` loops on `/api/profile/active` and `/api/models`; those were about the boot auth guard not covering specific endpoints, not the `next=` param nesting and URL-length explosion. Different root.

---
_Filed from a report in the Discord #report-bugs channel (reporter had a tab open long-term; URL measured ~12k chars). Reporter's original captured URL contained a private Tailnet hostname — deliberately omitted here._

nesquena-hermes:
## Summary

Fixes #5578: an expired-auth `401` while already on `/session/login?next=X` fed the login redirect its **own** URL, nesting `/session/login?next=…` into itself and re-percent-encoding on every bounce (`?` → `%3F` → `%253F` → …) until the URL grew past the browser limit and the tab broke (~12k chars reported).

## Root cause
All three redirect guards validated against open-redirect but **none rejected a `next` pointing back at the login page**, so the self-referential chain passed through re-encoded on each auth bounce:
1. `static/workspace.js:49` — the `api()` 401 handler wrapped `window.location.pathname+search` (which already contains `?next=X` when on the login page) into a fresh `next=`.
2. `static/login.js` `_safeNextPath()` — preserved the nested chain.
3. `api/routes.py` `_safe_login_redirect_path()` — re-emitted it via `quote(...)`, adding an encoding layer each pass.

## Fix (surgical, all three layers — preserves existing protections + happy path)
- **`_safe_login_redirect_path()`** (routes.py) and **`_safeNextPath()`** (login.js): after the existing validation, reject a `next` whose path targets the login route (`/login`, `/session/login`, subpath-mounted `*/login`), reject a `next` that already carries a raw or any-depth percent-encoded `next=` (`=`, `%3D`, `%253D`), and cap length at 2048.
- **`workspace.js` 401 handler**: if already **on** the login page, reload `login` **without** appending a `next` (the recursion source); non-login pages still capture the real destination.

The existing open-redirect protections (protocol-relative, absolute-URL, backslash, control-char rejection) are untouched — this only **adds** a self-reference/loop guard. A legitimate `next=/session/<id>` still works.

## Tests
`tests/test_issue5578_login_next_nesting.py`:
- Server guard rejects login self-reference (incl. subpath mount), nested `next` chains, encoded `next=` at any depth, overlong `next`, and the reconstructed exponential explosion.
- Preserves legit `/session/<id>` / `/workspace` paths **and** still rejects the open-redirect classics (`//`, `\`, absolute URL, control chars, empty/None).
- Static wiring asserts both JS guards carry the self-ref + nested-next + on-login-page checks.

**Fail-without-fix verified** (7/10 fail on `origin/master`). **69 auth/login regression tests green**; ruff clean on changed files.

Closes #5578.

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -685,6 +685,31 @@ def parse_cookie(handler) -> str | None:
     return morsel.value if morsel else None
 
 
+def _safe_login_inner_next(query: str | None) -> str:
+    """#5578: extract a SAFE, non-login `next` from a login page's own query.
+
+    When an expired-auth bounce lands on `/session/login?next=X`, we want to
+    preserve a legitimate inner destination X across the redirect to the real
+    login route — but only if X is itself safe (path-absolute, not protocol-
+    relative/backslash, no control chars) AND not login-shaped / not itself
+    carrying a nested `next=`. Anything else collapses to '' (no next), which
+    kills the self-referential chain. Mirrors _safe_login_redirect_path().
+    """
+    import urllib.parse as _u
+    raw = _u.parse_qs(query or "").get("next", [""])[0]
+    path = str(raw or "").strip()
+    if not path or path[0] != "/" or path[1:2] in {"/", "\\"}:
+        return ""
+    if re.search(r"[\x00-\x1f\x7f\s]", path) or len(path) > 2048:
+        return ""
+    _p = path.split("?", 1)[0].split("#", 1)[0].rstrip("/")
+    if _p == "/login" or _p.endswith("/login"):
+        return ""
+    if re.search(r"(?:\?|%3f|%253f|&|%26)next(?:=|%3d|%253d)", path, re.IGNORECASE):
+        return ""
+    return path
+
+
 def check_auth(handler, parsed) -> bool:
     """Check if request is authorized. Returns True if OK.
     If not authorized, sends 401 (API) or 302 redirect (page) and returns False."""
@@ -727,19 +752,32 @@ def check_auth(handler, parsed) -> bool:
         # the full original URL (the browser auto-decodes once).
         # (Opus pre-release advisor finding for v0.50.258.)
         import urllib.parse as _urlparse
-        # #5578: if the page being redirected is ALREADY login-shaped
-        # (/login, /session/login, or a subpath-mounted */login), do NOT wrap
-        # its full `path?query` — that query already carries a `next=`, so
-        # quoting the whole thing into a fresh `next=` nests the login URL into
+        # #5578: if the page being redirected is ALREADY login-shaped, do NOT
+        # wrap its full `path?query` into a fresh `next=` — that query already
+        # carries a `next=`, so quoting the whole thing nests the login URL into
         # itself and re-encodes it on every expired-auth bounce, exploding the
         # URL until the tab breaks. This guard runs in check_auth() (BEFORE
-        # route handling), which is the actual source of the server-side loop;
-        # the downstream _safe_login_redirect_path() never executes for this
-        # page-redirect. Redirect to bare `login` (no next); the login page
-        # preserves its own already-safe inner `next` from the current URL.
+        # route handling), the actual source of the server-side loop.
+        #
+        # The login page is served ONLY at the public `/login` route (see
+        # PUBLIC_PATHS + the routes.py `/login` handler); the app's client route
+        # `/session/login` is NOT public, so a bare relative `login` from
+        # `/session/login` resolves to `/session/login` again and re-triggers
+        # check_auth() — an infinite redirect. Resolve to the real login route
+        # with `../login`, which lands on `/login` from a `/session/*` scope and
+        # on `<mount>/login` under a subpath mount (verified via urljoin). Carry
+        # through only a validated, non-login inner `next` so a legitimate
+        # post-login destination still survives a bounce that happened to land
+        # on the login page.
         _login_path = (parsed.path or '/').rstrip('/')
         if _login_path == '/login' or _login_path.endswith('/login'):
-            handler.send_header('Location', 'login')
+            # /login itself is public → check_auth never redirects it; this only
+            # fires for the non-public client login route (e.g. /session/login).
+            _target = '../login' if '/' in _login_path.lstrip('/') else 'login'
+            _inner = _safe_login_inner_next(parsed.query)
+            if _inner:
+                _target += '?next=' + _urlparse.quote(_inner, safe='/')
+            handler.send_header('Location', _target)
             handler.send_header('Content-Length', '0')
             handler.end_headers()
             return False
--- a/api/auth.py
+++ b/api/auth.py
@@ -727,6 +727,22 @@ def check_auth(handler, parsed) -> bool:
         # the full original URL (the browser auto-decodes once).
         # (Opus pre-release advisor finding for v0.50.258.)
         import urllib.parse as _urlparse
+        # #5578: if the page being redirected is ALREADY login-shaped
+        # (/login, /session/login, or a subpath-mounted */login), do NOT wrap
+        # its full `path?query` — that query already carries a `next=`, so
+        # quoting the whole thing into a fresh `next=` nests the login URL into
+        # itself and re-encodes it on every expired-auth bounce, exploding the
+        # URL until the tab breaks. This guard runs in check_auth() (BEFORE
+        # route handling), which is the actual source of the server-side loop;
+        # the downstream _safe_login_redirect_path() never executes for this
+        # page-redirect. Redirect to bare `login` (no next); the login page
+        # preserves its own already-safe inner `next` from the current URL.
+        _login_path = (parsed.path or '/').rstrip('/')
+        if _login_path == '/login' or _login_path.endswith('/login'):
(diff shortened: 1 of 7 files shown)


nesquena (approved):

Independent review of the #5578 fix: an expired-auth 401 while already on `/session/login?next=X` fed the login redirect **its own URL**, nesting `next=` into itself and re-percent-encoding each bounce (`?`→`%3F`→`%253F`→…) until the URL exceeded browser limits (~12k chars reported) and the tab broke.

### What this ships

`api/routes.py` (+14 — server guard), `static/login.js` (+7 — client guard), `static/workspace.js` (+15 — 401-handler recursion source), `tests/test_issue5578_login_next_nesting.py` (+new, 10 tests). Agent-authored (`nesquena-hermes`). MERGEABLE, CI green.

### End-to-end trace — the recursion is cut at all three layers

1. **The source** (`workspace.js` 401 handler): when already **on** the login page (trailing-slash-stripped pathname matches `(?:^|\/)login$`), reload bare relative `login` **without** appending a `next` — the wrap that built the chain never happens. Non-login pages still capture the real destination. Relative `'login'` keeps subpath mounts working (resolves against the current directory, per the pre-existing iOS-PWA/subpath comment). ✓
2. **Server guard** (`_safe_login_redirect_path`): *after* the existing open-redirect checks — 2048 length cap; reject when the query/fragment-stripped path ends at the login route; reject any `next` whose value carries a nested `next=` at raw or encoded depth (`=`/`%3D`/`%253D`, prefix `?`/`%3F`/`%253F`/`&`/`%26`, case-insensitive). ✓
3. **Client guard** (`_safeNextPath`): the same three checks mirrored in JS. ✓

### Behavioral harness — I drove the real server guard

| Input | Result |
|---|---|
| `/session/abc123`, `/workspace` | pass through (happy path intact) ✅ |
| depth-1 `?next=`, depth-2 `%3F`, depth-3 `%253F` self-refs | → `/` ✅ |
| `/login`, `/hermes/session/login` (subpath) | → `/` ✅ |
| `/plugins/mylogin` | **passes** (suffix match requires the `/login` segment — no false positive) ✅ |
| `//evil.com`, control chars, overlong 3000-char path | → `/` (existing protections intact + new cap) ✅ |
| `/a?x=1&next=/b` (any nested next param) | → `/` ✅ |

**The one gap I probed hard:** a hand-crafted **depth-4** value (`%25253F`) passes the regex window. I verified this is **not** a live vector: each redirect hop decodes exactly one percent-layer (browser query-param parsing / `parse_qs`), so any depth-N chain must pass through depths 3→2→1 on subsequent hops — every one inside the regex window and rejected — and the 2048 cap bounds it meanwhile. It can never decay to the raw `?next=` form that re-enters the loop. The 3-depth window is therefore sufficient by construction, not by luck.

### Security audit

- All pre-existing open-redirect protections (path-absolute requirement, `//` and `\` rejection, absolute-URL, control-char/whitespace) are untouched — this PR only **adds** guards, and my harness confirms the classics still reject. ✓
- The new guards fail **closed** (to `/` server-side, `./` client-side). ✓
- No new surface: no HTML interpolation, no new endpoint. ✓

### Tests

- `tests/test_issue5578_login_next_nesting.py` — **10/10** (self-reference incl. subpath mount, nested chains at each encoded depth, overlong, the reconstructed exponential explosion, legit-path preservation, open-redirect classics, JS wiring).
- **Fail-without-fix verified independently:** 7/10 fail against master's three files — load-bearing.
- Auth/login blast-radius — **447 passed, 4 skipped**. `node --check` clean on both JS files.
- **Full suite: 11830 passed / 0 failed** in 317s (deselected the known darwin CRLF flake).

- **The on-login-page reload drops the inner destination.** Navigating to bare `'login'` discards the current `?next=X`, so after re-login the user lands at `/` instead of X (the code comment "the page preserves its own inner next" slightly overstates — `location.search` is empty after the reload). `'login' + window.location.search` would have kept the (already-validated, non-nested) destination. Strictly better than master's exploding tab either way; a cheap polish follow-up.
- The depth-3 regex window is sufficient (proven above); if anyone ever adds a multi-layer decode anywhere in the redirect path, revisit — a bounded unquote-loop guard would be the depth-unlimited form.

### Recommendation

The recursion is cut at its source (the 401 wrap on the login page) and belt-and-suspenders guards land on both the server and client validators — fail-closed, no regression to the existing open-redirect protections, and no over-blocking (harness-verified, including the `mylogin` suffix case). The one theoretical gap (crafted depth-4) is provably non-exploitable given one-decode-per-hop plus the length cap. Mutation-verified tests, 447-test auth family green, full suite clean. Ship.

nesquena-hermes:
Shipped in **v0.51.869**.

The login-redirect URL explosion (#5578) is fixed: a long-idle tab whose auth expired could feed the login redirect its own address, growing the `next=` chain exponentially until the tab broke. All redirect guards now collapse a `next` that resolves to the login route (through bounded percent-decoding so deeply-encoded chains are caught, fail-closed at the cap), while preserving legitimate redirects — including non-login paths that carry their own `next=` key — and the existing open-redirect protections.

Gate (3 convergent Codex rounds): round 1 caught an over-broad nested-`next` reject that regressed legit `next=`-carrying paths; round 2 caught a deep-encode decode off-by-one; both root-fixed; round 3 verified the login guards clean. Full suite 12042 passed. Merged via #5599.

PR #5580, opened 2026-07-04
release #5544: re-pin auto-scroll at true bottom after scroll-to-bottom

nesquena-hermes:
## Release: #5544 — re-pin auto-scroll at true bottom after scroll-to-bottom

Ships @luperrypf's fix for a crown-jewel chat auto-scroll regression, rebuilt on current `master` and re-gated.

### The bug (#5544)
A single scroll-up while a response was streaming set `_messageUserUnpinned=true`, and `scrollIfPinned()` then **permanently stopped auto-follow** — only an explicit `scrollToBottom()` cleared it, so scrolling back to the bottom didn't re-engage following. Permanent auto-follow lockout for the rest of the turn.

### The fix
`scrollIfPinned()` now re-pins auto-follow, but **only** when the reader has genuinely reached the true-bottom tail (`<=80px`) **and** shows no active scroll intent (wheel / keyboard / touch / non-message / scrollbar-drag), reusing the scroll listener's `_nearBottomCount>=2` debounce. Proximity alone (the ~250px `nearBottom` band) never re-pins — that is the **#4295 invariant** (a reader scanning the last lines mid-stream must not be yanked down). The listener's `<=80px` true-bottom gate is also restored, so the two re-pin paths and the ↓-button show condition all use one threshold.

### Gate (all green)
- **Codex** (regression): the only findings were stale-base artifacts (a prior stage was behind master); this branch is rebuilt on current master and the real diff is clean.
- **Fable UX**: **SHIP-UX** — reuses the existing intent-helper family + listener debounce, unifies three thresholds into one, matches ChatGPT's scroll-away-during-stream behavior; the 80px true-bottom resume is the right call.
- **Live browser drive** (seeded 40-msg session, real scroll events): scroll-up during stream does **not** yank the reader (#4295 holds); returning to true-bottom **re-engages** follow (view followed content growth to bottom); **zero JS console errors**.
- **Scroll-regression family:** 20/20 (incl. `test_near_bottom_proximity_alone_does_not_repin`). **Full suite:** running on this PR.

Attribution: original author @luperrypf (`Co-authored-by` trailer preserved).

Closes #5544


nesquena (approved):

Independent review of the #5544 fix (@luperrypf's, rebuilt on master): a single scroll-up during a streaming response set `_messageUserUnpinned=true`, and since only an explicit `scrollToBottom()` ever cleared it, `scrollIfPinned()` **permanently stopped auto-follow** for the rest of the turn — scrolling back to the bottom didn't re-engage following.

### What this ships

`static/ui.js` (+~28 — a re-pin path in `scrollIfPinned()` + the restored `<=80px` true-bottom gate in the scroll listener), `tests/test_issue3250_upward_scroll_intent_window.py` / `tests/test_tars_scroll_reset_regressions.py` (+25/-3), `CHANGELOG.md`. Agent-authored (`nesquena-hermes`), original author credited via `Co-authored-by`. MERGEABLE, CI fully green.

### End-to-end trace — the re-pin is correctly conservative

This is crown-jewel scroll behavior, so the load-bearing question is whether the new re-pin can ever fire against the reader's will (the #4295 invariant: proximity alone must never yank a reader scanning the last lines mid-stream). Traced the gate order in `scrollIfPinned()`:

1. **Any recent scroll intent bails first** — all five intent helpers (`_recentNonMessageScrollIntent` / `_recentMessageScrollIntent` / `_recentMessageTouchScrollIntent` / `_recentMessageWheelIntent` / `_recentMessageKeyScrollIntent`; I verified each is defined in `ui.js`) → reset debounce, return. An active scroll-up near the tail is never overridden. ✓
2. **True bottom required:** `_messageBottomDistance() > 80` → reset debounce, return. The ~250px `nearBottom` band alone can never re-pin. ✓
3. **Debounced:** requires two consecutive at-bottom observations (`_nearBottomCount >= 2`), the same debounce the scroll listener uses — a single transient bottom touch doesn't re-engage. ✓
4. Only then: `_messageUserUnpinned=false; _scrollPinned=true` — and the pre-existing pinned-path guards (`_recentNonMessageScrollIntent`, `>500px` jump) still apply downstream. ✓

The **listener-side gate is unified to the same threshold**: `if(!_messageUserUnpinned || bottomDistance<=80)` — when unpinned, re-affirming the pin now also requires true bottom, so both re-pin paths and the ↓-button share one `<=80px` definition instead of three drifting thresholds. ✓

Sharing `_nearBottomCount` between the listener and `scrollIfPinned()` is semantically consistent (both count "consecutive at-bottom observations", and each resets it on any disqualifying state), so the debounce can't be satisfied by stale counts. ✓

### #4295 invariant — structurally pinned

`test_near_bottom_proximity_alone_does_not_repin` is in the passing family, and the PR's stage ran a **live browser drive** (seeded 40-message session, real wheel/scroll events): scroll-up during stream does not yank (#4295 holds), returning to true bottom re-engages follow, zero console errors. The static gates I traced match that observed behavior.

### Tests

- Targeted scroll-regression files — **17/17**; the crown-jewel scroll family — **270 passed, 4 skipped**.
- **Fail-without-fix verified independently:** against master's `ui.js`, 2 of the updated assertions fail — load-bearing.
- `node --check static/ui.js` clean.
- **Full suite: 11820 passed / 0 failed** in 311s (deselected the known darwin CRLF flake). CI fully green (lint + browser-smoke + all shards).

- `scrollIfPinned()` increments the shared debounce on render ticks while the listener increments on scroll events — mixed sources can reach the threshold one tick sooner than two scroll events would. Both sources genuinely observe "still at true bottom with no intent", so this is within the debounce's meaning; flagging only for awareness.
- The CHANGELOG `[Unreleased]` entry follows convention.

### Recommendation

The fix restores bottom-re-engagement without touching the #4295 guarantee: re-pin requires true bottom (≤80px) **and** no scroll intent of any kind **and** a two-observation debounce — proximity alone still never re-pins, and the three thresholds are unified into one. Mutation-verified tests, 270-test family green, live-drive evidence consistent with the trace, full suite clean. Ship.

PR #5581, opened 2026-07-04
release #5543: fire-and-forget memory commit + drain-safe managed stop

nesquena-hermes:
## Release: #5543 — fire-and-forget memory commit + drain-safe managed stop

Ships @luperrypf's fix to unblock "+ New Chat", rebuilt on current `master`, with maintainer hardening folded in and a full dual-gate.

### The bug (#5543)
`POST /api/session/new` stalled 1–5+ seconds because the previous session's `commit_session_memory()` (the extraction call to the memory provider) ran **synchronously on the request thread**. "+ New Chat" felt unresponsive.

### The fix
The commit now runs in a fire-and-forget daemon thread, guarded by the existing per-session in-flight serialization, so the new-session POST returns immediately.

### Data-loss prevention on a managed stop (maintainer hardening)
The naive fire-and-forget would silently lose the deferred commit on a managed restart. Folded in:
- **`server.py`** installs a `SIGTERM` handler that requests an orderly stop so `serve_forever()`'s `finally` runs `drain_all_on_shutdown()` — the default `SIGTERM` disposition terminated the process without unwinding, losing in-flight commits.
- **`session_lifecycle.py`**: bounded, self-unregistering background-commit registry; a `_draining` flag so no new worker starts once the drain begins (a late arrival is committed by the inline generation-drain instead — closes the register-before-start window); each inline commit runs in a **budget-bounded daemon worker** so a hung memory provider can't stall the stop past the deadline.

### Gate (dual gate earned its keep here)
Across gate rounds, **both Codex and Opus independently caught two real defects in the hardening itself** (each with a runnable repro): (1) the deadline never fired against a worker wedged on `_condition.wait()`; (2) even after that, the **direct** provider call when the inline drain owns the segment was still unbounded. Both fixed at root (budget-bounded joins + daemon-worker inline commit), with **two regression tests I verified actually catch the bugs** (fail-without / pass-with).
- **Codex** (final): no real finding (the one flag was a stale-base `session_ops.py` artifact — this PR doesn't touch that file).
- **Opus** (final): **SHIP** — deadline genuinely enforced in both phases, no inconsistent state on a timed-out commit; one optional fast-follow (starvation clamp) recommended to bundle with the known `ctl.sh` SIGKILL-widening work.
- **Tests:** 25 in the #5543 suite (incl. both deadline regressions) + 16 lifecycle generation tests green. **Full suite:** running on this PR.

Attribution: original author @luperrypf (`Co-authored-by` trailer preserved).

Closes #5543


nesquena (approved):

Independent review of the #5543 fix (@luperrypf's, rebuilt on master with maintainer hardening): `POST /api/session/new` blocked 1–5+ seconds on the previous session's synchronous `commit_session_memory()` (memory-provider extraction call), making "+ New Chat" feel unresponsive. The commit now runs fire-and-forget — with a drain-safe managed stop so the deferral can't silently lose data. Since this is concurrency/lifecycle-heavy, I focused on the thread-safety and shutdown races.

### What this ships

`api/routes.py` (+58/-17 — the fire-and-forget dispatch), `api/session_lifecycle.py` (+110 — commit-thread registry, `_draining` gate, budget-bounded drain), `server.py` (+30 — SIGTERM orderly-stop handler), `tests/test_issue5543_background_memory_commit.py` (+358, new), `CHANGELOG.md`. Agent-authored (`nesquena-hermes`), original author credited. MERGEABLE, CI fully green.

### Thread-safety / race analysis (the crucial part)

**1. The fire-and-forget itself is safe.** The worker body is the same code the request thread ran (agent lookup under `SESSION_AGENT_CACHE_LOCK`, then `commit_session_memory`), and overlapping commits for a session were already serialized by the pre-existing per-session in-flight guard — moving it off-thread changes *when*, not *what*. Errors are logged (upgraded debug→warning, appropriately, since there's no request to surface them on). ✓

**2. The register-before-start window is genuinely closed.** `_register_background_commit_thread` checks `_draining` and adds to the registry under **one lock** — the same lock `drain_all_on_shutdown` holds when setting `_draining=True`. So a worker either (a) registers before the drain → it's in the join snapshot, or (b) is refused → never started, and its session's uncommitted generation is picked up by the inline generation-drain loop (which re-snapshots `_sessions` each pass). No lost-in-between state. ✓

**3. The registered-but-not-yet-started edge is safe.** The drain's `is_alive()` gate skips a registered thread whose `.start()` hasn't run yet (also avoiding the `join()`-on-unstarted `RuntimeError`); if that worker starts later, either it grabs in-flight first (the inline commit waits on it, budget-bounded) or the inline drain commits first and the late worker no-ops under `_condition`. Either way the generation is committed exactly once. ✓

**4. The deadline is enforced in both phases — the two gate-caught defects are really fixed.** `_commit_within_budget` runs each inline commit in a **daemon worker joined for at most the remaining budget**, covering both wedge shapes the dual gate caught: a worker stuck on `_condition.wait()` (the wait-for-in-flight path) *and* an unbounded **direct** provider call when the inline drain owns the segment. The shared `started`/`_remaining()` clock spans the join phase + the inline loop, checked per-pass **and** per-session; `_drain_background_commit_threads` clamps each join to `min(5s, remaining)`. An abandoned timed-out worker mutates generation state only under `_condition`, so nothing is corrupted. ✓

**5. The SIGTERM handler is correct.** Python's default SIGTERM disposition kills the process without unwinding `serve_forever()`'s `try/finally` — so `drain_all_on_shutdown()` (which I verified is already in that `finally`) never ran on a managed `ctl.sh` stop, making the naive fire-and-forget lossy. The handler: idempotent via an `Event` guard, dispatches `httpd.shutdown()` from a **helper thread** (calling it from the serve_forever thread deadlocks — correctly avoided), and the install is wrapped for non-main-thread contexts (embedded/test harness). `signal`/`threading` imports present. ✓

**6. Registry hygiene:** completed workers self-unregister in `finally`; the drain snapshots-and-clears; bounded growth. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| + New Chat with slow provider | POST returns immediately; commit in background ✅ |
| Two rapid new-chats, same prev session | per-session in-flight guard serializes ✅ |
| SIGTERM mid-commit | orderly stop → finally → drain joins the worker ✅ |
| Register attempted after drain began | refused; inline loop commits the generation ✅ |
| Registered but not yet started at drain | skipped by `is_alive()`; committed exactly once via inline/no-op ✅ |
| Worker wedged on `_condition.wait()` | budget-bounded join → drain returns at deadline ✅ (regression-tested) |
| Provider hangs on direct inline call | daemon worker + bounded join → deadline honored ✅ (regression-tested) |
| Timed-out abandoned worker | state mutated only under `_condition`; no corruption ✅ |
| Repeated SIGTERM / SIGINT | idempotent Event guard; single shutdown ✅ |
| Non-main-thread handler install | caught, logged, skipped ✅ |

### Tests

- `tests/test_issue5543_background_memory_commit.py` — **9/9**, including both deadline regressions (`test_drain_deadline_returns_despite_stuck_worker`, `test_drain_deadline_returns_when_inline_provider_hangs_no_prior_worker`).
- **Fail-without-fix verified independently:** against master's three files, **all 9 fail** — and the two deadline tests fail *fast* (injected fakes, no real hangs). Load-bearing.
- Lifecycle/memory-commit family — **71 passed, 12 skipped**.
- **Full suite: 11829 passed / 0 failed** in 324s (deselected the known darwin CRLF flake). CI fully green.

- Opus's recommended **starvation clamp** fast-follow (and bundling with the known `ctl.sh` SIGKILL-widening work) still stands — with a wedged provider the drain can legitimately consume the full 30s default; operators whose supervisor SIGKILLs sooner get a truncated (but deadline-clean, corruption-free) drain. Worth the follow-up as planned.
- The PR body's "25 in the #5543 suite" reads as the 9 file tests + 16 lifecycle-generation tests combined — counting nuance only.

### Recommendation

The perf fix is one line of intent (move the commit off the request thread) but the value here is the hardening around it, and it holds up under race analysis: a single-lock register-vs-drain gate with an inline backstop for every losing interleaving, a genuinely enforced two-phase deadline (both gate-caught wedge shapes are regression-tested and I verified they fail on master), and a correctly-built SIGTERM→orderly-stop path that makes the existing `finally` drain actually run on managed stops. Full suite clean. Ship.

PR #5605, opened 2026-07-05
fix(#5583): keep mobile nav actions visible

### LINKED ISSUE #5583: Responsive mobile layout hides Dashboard and Theme Creator icons at 800px and below
## Summary

When the Hermes WebUI viewport switches to the mobile/narrow layout at `800px` width or below, the **Dashboard** icon and **Theme Creator extension** icon disappear from navigation.

At widths above `800px`, both icons render correctly.

## Expected Behavior

When the UI switches to the mobile layout, extension and dashboard navigation actions should remain accessible, even if they move to a different nav pattern or overflow menu.

## Actual Behavior

At `800px` or less:

- the layout switches to the mobile view
- the **Dashboard** icon disappears
- the **Theme Creator** icon disappears

At widths above `800px`:

- both icons are visible and work normally

## Reproduction

1. Open Hermes WebUI in a browser or embedded WebView.
2. Set the viewport width to **greater than 800px**.
3. Confirm the **Dashboard** and **Theme Creator** icons are visible.
4. Shrink the viewport to **800px or below**.
5. Observe that the layout switches to the mobile view and those icons are no longer visible.

## Notes

- This appears to be tied to the responsive/mobile navigation breakpoint.
- The issue is not that the extension is missing entirely; it is visible in the wider layout and disappears only once the narrow/mobile layout activates.
- It would be helpful if the mobile layout preserved these actions in the top bar, side drawer, or an overflow menu.

## Screenshots

- `> 800px`: icons visible
- `<= 800px`: icons missing

Image

Image

rodboev:
## Thinking Path

- The dashboard link already has a runtime visible state, but the narrow sidebar breakpoint overrides that state with `display:none!important`.
- The same sidebar nav surface is where dashboard-style extension actions need to remain reachable, so the fix makes “visible nav action” a shared state instead of another breakpoint exception.
- Theme Creator currently adds a rail-only action, so core needs a small shared mirror for rail actions that do not already have a mobile/sidebar counterpart.
- The regression belongs with the existing dashboard link UI tests, which currently assert the broken mobile CSS contract.

## What Changed

- `static/style.css`: replaces the narrow breakpoint hide override with a shared visible nav-action rule that can win consistently across rail and sidebar layouts.
- `static/ui.js`: keeps the existing dashboard visible class, toggles the shared nav-action visible class from the dashboard status path, and reconciles rail-only extension action mirrors into the mobile sidebar while delegating clicks back to the current source button.
- `tests/test_dashboard_link_ui.py`: inverts the mobile CSS regression, extends the dashboard status driver, and adds a Theme Creator-shaped rail action harness that proves the mobile mirror stays visible, clickable, and removed when its source disappears.

## Why It Matters

Dashboard and dashboard-style extension actions stay reachable in the narrow navigation layout. Future nav actions can opt into the same visibility state instead of duplicating breakpoint-specific display rules.

## Verification

```text
pytest tests/test_dashboard_link_ui.py -v --timeout=60
npx eslint --no-config-lookup -c eslint.runtime-guard.config.mjs "static/**/*.js"
```

## Upstream

Closes #5583.

## Model Used

GPT-5 via Codex CLI

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -2104,7 +2104,7 @@
    (see line ~681). The rail rule is no longer needed. */
   .rail .nav-tab.active::before{content:'';position:absolute;left:-6px;top:50%;bottom:auto;transform:translateY(-50%);width:3px;height:16px;background:var(--accent);border-radius:0 2px 2px 0;}
   .dashboard-link{position:relative;}
-  .dashboard-link-visible{display:flex!important;}
+  .dashboard-link-visible,.nav-action-visible{display:flex!important;}
   .dashboard-external-badge{position:absolute;right:5px;top:5px;width:9px;height:9px;border-radius:2px;border:1px solid var(--accent);background:var(--sidebar);box-shadow:0 0 0 2px var(--sidebar);opacity:.95;}
   .dashboard-external-badge::after{content:'';position:absolute;right:-2px;top:-2px;width:5px;height:5px;border-top:1.5px solid var(--accent);border-right:1.5px solid var(--accent);}
   .sidebar-nav .dashboard-external-badge{right:8px;top:7px;width:8px;height:8px;}
@@ -3100,8 +3100,7 @@
     .sidebar-nav .nav-tab svg{width:20px;height:20px;}
     .sidebar-nav .nav-tab.active{background:var(--accent-bg);}
     .sidebar-nav .nav-tab.active::before{left:-4px;top:50%;bottom:auto;transform:translateY(-50%);width:3px;height:16px;border-radius:0 2px 2px 0;}
-    .sidebar-nav .dashboard-link,
-    .sidebar-nav .dashboard-link.dashboard-link-visible{display:none!important;}
+    .sidebar-nav .dashboard-link:not(.nav-action-visible){display:none!important;}
     .sidebar .panel-view{height:100%;margin-left:52px;}
     .sidebar .panel-head{padding-right:52px;}
     .mobile-sidebar-close{display:inline-flex!important;position:absolute;right:4px;top:calc(4px + var(--app-titlebar-safe-top));width:44px;height:44px;z-index:4;background:var(--surface);border:1px solid var(--border);}
--- a/static/style.css
+++ b/static/style.css
@@ -3100,7 +3100,7 @@
     .sidebar-nav .nav-tab svg{width:20px;height:20px;}
     .sidebar-nav .nav-tab.active{background:var(--accent-bg);}
     .sidebar-nav .nav-tab.active::before{left:-4px;top:50%;bottom:auto;transform:translateY(-50%);width:3px;height:16px;border-radius:0 2px 2px 0;}
-    .sidebar-nav .dashboard-link:not(.nav-action-visible){display:none!important;}
+    .sidebar-nav .dashboard-link:not(.nav-action-visible),.sidebar-nav [data-nav-action-mirror]:not(.nav-action-visible){display:none!important;}
     .sidebar .panel-view{height:100%;margin-left:52px;}
     .sidebar .panel-head{padding-right:52px;}
     .mobile-sidebar-close{display:inline-flex!important;position:absolute;right:4px;top:calc(4px + var(--app-titlebar-safe-top));width:44px;height:44px;z-index:4;background:var(--surface);border:1px solid var(--border);}
--- a/static/ui.js
+++ b/static/ui.js
@@ -1290,6 +1290,17 @@ function _dashboardBrowserUrl(status){
   const displayHost=browserHost.includes(':')&&!browserHost.startsWith('[')?'['+browserHost+']':browserHost;
   return source.protocol+'//'+displayHost+':'+status.port;
 }
+function _stripInlineEventHandlers(node){
+  if(!node)return;
+  const strip=el=>{
+    Array.from(el.attributes||[]).forEach(attr=>{
+      if(attr.name&&attr.name.toLowerCase().startsWith('on'))el.removeAttribute(attr.name);
+    });
+    if('onclick' in el)el.onclick=null;
+    Array.from(el.children||[]).forEach(strip);
+  };
+  strip(node);
+}
 function _syncNavActionMirrors(){
   const rail=document.querySelector('.rail');
   const sidebar=document.querySelector('.sidebar-nav');
@@ -1305,6 +1316,7 @@ function _syncNavActionMirrors(){
     let mirror=mirrors.find(el=>el.getAttribute('data-nav-action-mirror')===source.id);
     if(!mirror){
       mirror=source.cloneNode(true);
+      _stripInlineEventHandlers(mirror);
       mirror.id=source.id+'Mobile';
       mirror.classList.remove('rail-btn');
       mirror.classList.add('nav-action-visible','has-tooltip--bottom');
@@ -1314,6 +1326,7 @@ function _syncNavActionMirrors(){
       sidebar.insertBefore(mirror,anchor||null);
     }else{
       mirror.innerHTML=source.innerHTML;
+      _stripInlineEventHandlers(mirror);
     }
     mirror._navActionSource=source;
     const label=source.getAttribute('data-tooltip')||source.getAttribute('aria-label')||'';
--- a/static/ui.js
+++ b/static/ui.js
@@ -1290,12 +1290,50 @@ function _dashboardBrowserUrl(status){
   const displayHost=browserHost.includes(':')&&!browserHost.startsWith('[')?'['+browserHost+']':browserHost;
   return source.protocol+'//'+displayHost+':'+status.port;
 }
+function _syncNavActionMirrors(){
+  const rail=document.querySelector('.rail');
+  const sidebar=document.querySelector('.sidebar-nav');
+  if(!rail||!sidebar)return;
+  const sources=Array.from(rail.querySelectorAll('.nav-tab:not([data-panel]):not([data-dashboard-link])')).filter(source=>source.id);
+  const mirrors=Array.from(sidebar.querySelectorAll('[data-nav-action-mirror]'));
+  const sourceIds=new Set(sources.map(source=>source.id));
+  mirrors.forEach(mirror=>{
+    if(!sourceIds.has(mirror.getAttribute('data-nav-action-mirror')))mirror.remove();
+  });
+  sources.forEach(source=>{
+    source.classList.add('nav-action-visible');
+    let mirror=mirrors.find(el=>el.getAttribute('data-nav-action-mirror')===source.id);
+    if(!mirror){
+      mirror=source.cloneNode(true);
+      mirror.id=source.id+'Mobile';
+      mirror.classList.remove('rail-btn');
+      mirror.classList.add('nav-action-visible','has-tooltip--bottom');
+      mirror.setAttribute('data-nav-action-mirror',source.id);
+      mirror.addEventListener('click',e=>{e.preventDefault();if(mirror._navActionSource)mirror._navActionSource.click();});
+      const anchor=sidebar.querySelector('.dashboard-link,[data-dashboard-link]')||sidebar.querySelector('[data-panel="logs"]');
+      sidebar.insertBefore(mirror,anchor||null);
+    }else{
+      mirror.innerHTML=source.innerHTML;
+    }
+    mirror._navActionSource=source;
+    const label=source.getAttribute('data-tooltip')||source.getAttribute('aria-label')||'';
+    if(label)mirror.setAttribute('data-label',label);
+  });
+}
+function _initNavActionMirrors(){
+  _syncNavActionMirrors();
+  const rail=document.querySelector('.rail');
+  if(rail&&window.MutationObserver)new MutationObserver(_syncNavActionMirrors).observe(rail,{childList:true});
+}
+if(document.readyState==='loading')document.addEventListener('DOMContentLoaded',_initNavActionMirrors,{once:true});
+else _initNavActionMirrors();
 function _applyDashboardStatus(status){
   const running=!!(status&&status.running);
   const url=running?_dashboardBrowserUrl(status):'';
   const warning=running&&!_dashboardIsBrowserLoopback()?t('dashboard_loopback_warning'):'';
   document.querySelectorAll('[data-dashboard-link]').forEach(btn=>{
     btn.classList.toggle('dashboard-link-visible',running);
+    btn.classList.toggle('nav-action-visible',running);
     btn.style.display=running?'':'none';
     btn.dataset.dashboardUrl=url;
     const tipText=warning||t('tab_dashboard');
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (mobile nav-action mirror double-fires for inline-onclick actions — cloneNode copies onclick AND adds a forwarding listener — 1 SILENT)

**Certified head:** `sha:2c66f7b6` (rebased onto current master, `git apply` clean) · **PR:** #5605 · _rodboev, fix(#5583): keep mobile nav actions visible_
**Verdict:** The mirroring approach (clone rail nav-actions into the mobile sidebar) fixes the #5583 clipping, and the forwarding-listener leak/stale-cleanup are handled correctly — but `source.cloneNode(true)` copies the source's inline `onclick` onto the mirror WHILE the code also adds a forwarding click listener, so a nav action implemented with inline `onclick` (common for extensions/admin actions) fires TWICE on a mobile mirror click.

### What I ran (rebased worktree `/tmp/wt-rebase-5605`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | SHIP-WITH-FIXES — 1 SILENT (double-fire for inline-onclick nav actions); Codex + I verified by reading |
| Full pytest suite | ✅ rc=0 green (double-fire is a runtime edge for inline-onclick actions — not suite-covered) |
| dashboard-link/nav tests | ✅ pass |

### Findings
**⛔ SILENT (I + Codex CONFIRMED) — custom mobile nav-action clicks can run twice** (`static/ui.js:~1331`): the mirror is built via `source.cloneNode(true)` (copies the source's inline `onclick` attribute) and THEN a forwarding click listener is added (`mirror.addEventListener('click', … mirror._navActionSource.click())`), with `mirror.innerHTML=source.innerHTML` re-applied on re-sync. There is NO `removeAttribute('onclick')`/`onclick=null` strip (verified: 0 occurrences). So for an extension/admin rail action implemented with inline `onclick`, a mobile mirror click executes BOTH the cloned inline handler AND the forward-to-source → duplicate opens/toggles/API calls (double navigation, double state-toggle, double request). The `innerHTML=source.innerHTML` on re-sync can likewise re-introduce inline handlers on child elements. **Fix (Codex-exact):** strip inline event handlers from the mirror after cloning AND after `mirror.innerHTML=source.innerHTML` (e.g. `mirror.removeAttribute('onclick')` + clear inline handlers on descendants), or build the mirror from a safe shell instead of cloning executable attributes — keep ONLY the single forwarding listener.

**✅ Otherwise correct (keep):** the forwarding listener is added once (in the `if(!mirror)` create branch), `_navActionSource` is updated each sync so the once-bound handler forwards to the current source (no listener accumulation), and stale mirrors are removed when the source id is gone. The mirroring concept is sound; only the cloned-inline-handler double-fire needs fixing.

**RED — gate-fail/changes-requested (1 SILENT): strip inline event handlers from the mirror after `cloneNode(true)` and after the `innerHTML=source.innerHTML` re-sync (mirror.removeAttribute('onclick') + descendants), so only the single forwarding listener fires — otherwise extension/admin nav actions using inline `onclick` double-fire on mobile.** The #5583 fix (mirror rail nav-actions into the visible mobile sidebar) is a good approach with correct listener/stale-cleanup handling; it just needs to not carry the source's inline handlers into the clone. Then it's a visible mobile-nav change → Nathan's visual sign-off. concept 4/5 (real #5583 fix; one double-fire edge for inline-onclick actions). Author @rodboev (T1). crit=3. (Gate value: `cloneNode(true)` copies inline event-handler attributes — when you also attach a forwarding listener, inline-handler'd sources double-fire; strip inline handlers on any cloned/mirrored interactive element.)

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; forwarding-listener is create-once (no leak) + stale-cleanup correct, BUT source.cloneNode(true) copies inline onclick (0 removeAttribute strip) alongside the forwarding listener → double-fire for inline-onclick nav actions (extensions/admin). Fix: strip inline handlers from the mirror after clone + innerHTML update. Suite green (runtime edge). Cert valid for sha:2c66f7b6._

rodboev:
Fixed in f650b1b7.

I added a mirror scrub step after `cloneNode(true)` and after the mirror `innerHTML` refresh. It removes inline `on*` attributes and clears `onclick` on the mirror tree, so mobile clicks now only go through the existing forwarding listener to the source action.

Validated:
- `python -m pytest tests/test_dashboard_link_ui.py -v --timeout=60`
- `npx eslint --no-config-lookup -c eslint.runtime-guard.config.mjs "static/**/*.js"`

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (round-2 CONVERGED; FULL gate: Codex + Fable-UX + suite) · ⏸️ visible mobile-nav → Nathan screenshot sign-off

**Certified head:** `sha:a3dbbcf8` (clean rebase, branch `gate-rebase/5605-mobile-nav-visible`) · **PR:** #5605 · _rodboev, fix(#5583): keep mobile nav actions visible_
**Verdict:** Round-2 CONVERGED — my round-1 double-fire SILENT is fixed, and this now has the FULL gate (Codex + Fable-UX + full suite), correcting an earlier gap where I ran Codex+suite only.

### What I ran (rebased worktree `/tmp/wt-rebase-5605b`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SAFE TO SHIP** — 0 findings (double-fire fixed) |
| **Fable-UX (visible surface)** | **SHIP-UX** — 0 blocking; mirrored nav actions look right, no regression, adequate touch-targets |
| Full pytest suite | ✅ **12084 passed, 0 failed** |

### Findings
**✅ Round-1 double-fire CLOSED:** new `_stripInlineEventHandlers(node)` removes ALL `on*` attributes + sets `el.onclick=null` (incl. descendants), called on the mirror after `cloneNode(true)` AND after `mirror.innerHTML=source.innerHTML` — so only the single forwarding listener fires (no inline-onclick double-fire). MutationObserver on `.rail` keeps mirrors synced; stale-mirror cleanup intact; forwarding listener added once + `_navActionSource` updated each sync (no leak). Codex SAFE.
**✅ Fable-UX SHIP-UX:** mirrored nav actions look right in the mobile sidebar drawer (placement between Insights and Dashboard/Logs, icon size/spacing, tooltip), no visual regression to dashboard-link/existing items, adequate mobile touch-targets, no MutationObserver FOUC.

### Screenshots for Nathan (Fable-enumerated)
1. Mobile (≤640px) sidebar drawer nav column with an extension (Theme Creator) active + dashboard running — mirror icon size/spacing vs native items, position, external-link badge.
2. Same view, dashboard stopped — Dashboard tile absent, no gap artifact.
3. Desktop rail (≥641px) unchanged with the extension active.

**Ready to merge — use branch `gate-rebase/5605-mobile-nav-visible` (`sha:a3dbbcf8`), NOT the PR's stale head `f650b1b7` — after Nathan's screenshot sign-off (visible mobile-nav).** Full gate clean (Codex SAFE + Fable-UX SHIP-UX + full suite green), round-1 double-fire fixed. concept 4/5 (real #5583 fix). Author @rodboev (T1). crit=3.

---
_Gate-certifier layer. FULL gate this round: Codex SAFE + Fable-UX SHIP-UX + full suite green (0 failed); round-1 cloneNode double-fire fixed via _stripInlineEventHandlers on both clone + innerHTML paths. Visible → Nathan screenshots. Cert valid for sha:a3dbbcf8._

nesquena-hermes (changes requested):
## Gate review — SHIP-WITH-FIXES (one real defect + one UX polish)

Ran the full gate (Codex + Fable-UX + suite) on the current head (`f650b1b7`) rebased onto master. The mirror mechanism is well-built (single forwarding listener, `_stripInlineEventHandlers` on both clone + innerHTML paths, stale-mirror cleanup) — but both reviewers independently flagged the same latent invariant break:

### ⛔ Real defect — the mirror sync mutates the SOURCE rail element's visibility
`static/ui.js:1483` does `source.classList.add('nav-action-visible')` on every qualifying rail action. Combined with the global rule `static/style.css:2107` `.dashboard-link-visible,.nav-action-visible{display:flex!important}`, this can **force-show an extension's deliberately-hidden rail action** on the desktop rail — the source visibility is supposed to stay extension-owned. The `MutationObserver` (`ui.js:1507`) also only watches `childList`, so a later source class/style visibility change won't re-sync the mirror.

**Fix (Codex + Fable agree):**
1. `ui.js:1483` — remove `source.classList.add('nav-action-visible')`; source rail visibility must stay extension-owned.
2. `ui.js:~1489-1497` — compute source visibility *before* mirroring and toggle `nav-action-visible` **on the mirror only** (add when the source is actually visible, remove when hidden).
3. `style.css` — add a default-hidden rule `[data-nav-action-mirror]:not(.nav-action-visible){display:none!important}` (scoped to `.sidebar-nav`) so hidden mirrors stay hidden on mobile.
4. `ui.js:1507` — expand the `MutationObserver` to also observe `attributes` (class/style/hidden) + subtree so mirror visibility/markup can't go stale.

### UX polish (Fable, non-blocking)
- Close the mobile drawer after a mirrored-action tap. Panel tabs close the drawer via `_closeMobileSidebarAfterPanelSelection`, but the forwarding listener just calls `source.click()` — if an extension action opens a modal (e.g. Theme Creator), the drawer stays open behind it. Add a guarded `closeMobileSidebar()` after `source.click()` (matches ChatGPT/Claude mobile behavior).

### What's clean
Codex confirmed no double-fire (round-1 fix holds), no listener leak, per-mirror forwarding correct, `aria-label` survives the clone, adequate touch targets, desktop rail otherwise unchanged. Full suite green (the flagged edges aren't test-covered — that's the gate's value). Concept is a real #5583 fix and worth shipping once the source-visibility mutation is removed.

Once re-pushed I'll re-gate (Codex + Fable + suite + browser) and, on green + your screenshot sign-off, ship. Thanks @rodboev.

rodboev:
Addressed in 9f9ebc97.

`static/ui.js:1315` now derives mirror visibility from the source element instead of adding `nav-action-visible` to the rail button, so extension-owned desktop visibility stays untouched. The mirror sync also watches `class`, `style`, `hidden`, `aria-hidden`, `data-tooltip`, and `aria-label` changes at `static/ui.js:1353-1357`, and the forwarded mirror click closes the mobile drawer at `static/ui.js:1333-1336`.

`static/style.css:3103` now hides mirrored sidebar actions only when the mirror itself lacks `nav-action-visible`, so a hidden source stays hidden on mobile without forcing the source visible on the rail.

nesquena-hermes:
## Re-gate read — round-3 CONVERGED on code inspection ✅ (all 4 CHANGES_REQUESTED points + the UX polish addressed) · full gate + Nathan screenshot sign-off still owed

Read the pushed head `sha:9f9ebc97` in a read-only worktree and diffed it against `origin/master` (`static/ui.js` +70, `static/style.css` ±5, `tests/test_dashboard_link_ui.py` +223). Walking the round-2 review (21:56Z) point-by-point against the actual diff — each item is closed:

### 1. Source-visibility mutation removed ✅
The core defect (`source.classList.add('nav-action-visible')` forcing an extension's deliberately-hidden rail action visible on desktop) is gone — `grep -n "source.classList.add" static/ui.js` returns **zero** hits. Visibility is now derived from the source and applied to the **mirror only**:

```js
// static/ui.js:1315-1323
const sourceVisible=(()=>{
  if(source.hidden||source.getAttribute('aria-hidden')==='true')return false;
  if(source.classList.contains('nav-tab-hidden'))return false;
  if(source.style&&(source.style.display==='none'||source.style.visibility==='hidden'))return false;
  if(typeof window!=='undefined'&&typeof window.getComputedStyle==='function'){
    const computed=window.getComputedStyle(source);
    if(computed&&(computed.display==='none'||computed.visibility==='hidden'))return false;
  }
  return true;
})();
```
…then `mirror.classList.toggle('nav-action-visible',sourceVisible)` at `static/ui.js:1345`. Source rail element is never mutated — extension-owned visibility stays intact. Point 2 is folded into the same block.

### 2. Mirror stays default-hidden when the source is hidden ✅
`static/style.css:3103` (inside `@media(max-width:640px)`, matching the pre-existing dashboard-link pattern):
```css
.sidebar-nav .dashboard-link:not(.nav-action-visible),.sidebar-nav [data-nav-action-mirror]:not(.nav-action-visible){display:none!important;}
```
So a hidden source keeps its mirror hidden on mobile without forcing the source visible on the rail. Global show rule at `static/style.css:2107` (`.dashboard-link-visible,.nav-action-visible{display:flex!important;}`) only wins for mirrors that actually earned `nav-action-visible`.

### 3. MutationObserver now catches visibility drift ✅
Was `childList`-only; now watches attribute mutations + subtree (`static/ui.js:1352-1358`):
```js
new MutationObserver(_syncNavActionMirrors).observe(rail,{
  childList:true, subtree:true, attributes:true,
  attributeFilter:['class','style','hidden','aria-hidden','data-tooltip','aria-label'],
});
```
A later source class/style/`hidden`/`aria-hidden` change now re-syncs the mirror — no stale visibility.

### 4. UX polish — mobile drawer closes after a mirrored tap ✅
The forwarding listener now closes the drawer after delegating (`static/ui.js:1333-1337`), so an extension action that opens a modal (e.g. Theme Creator) doesn't leave the drawer stacked behind it:
```js
if(mirror._navActionSource)mirror._navActionSource.click();
if(typeof closeMobileSidebar==='function')closeMobileSidebar();
```

### Round-1 double-fire still holds
`_stripInlineEventHandlers` is still called on both the `cloneNode(true)` path (`ui.js:1328`) and the `innerHTML=source.innerHTML` re-sync path (`ui.js:1342`), and the forwarding listener is added once in the create branch with `_navActionSource` refreshed each sync — no inline-onclick double-fire, no listener accumulation.

### Note (non-blocking)
`_applyDashboardStatus` at `ui.js:1368` also toggles `nav-action-visible` on `[data-dashboard-link]` buttons — that's fine and distinct from the flagged bug: the dashboard link's visibility is legitimately status-driven, and the mirror scan explicitly excludes it (`.nav-tab:not([data-panel]):not([data-dashboard-link])` at `ui.js:1308`), so there's no double-handling.

### Verdict
On code read this converges cleanly — all four CHANGES_REQUESTED items and the Fable UX polish are resolved with no new regressions visible in the diff. This still needs the **full gate re-run on `sha:9f9ebc97`** (Codex + Fable-UX + full suite + browser) plus **Nathan's screenshot sign-off**, since it's a visible mobile-nav surface. Suggested screenshots: mobile (≤640px) drawer with an extension (Theme Creator) active + dashboard running; same with a deliberately-hidden extension action (mirror absent, no gap); desktop rail (≥641px) unchanged with the extension active. Thanks @rodboev.

nesquena-hermes:
Shipped in **v0.51.899**. Thanks @rodboev — the dashboard link and extension nav-actions now mirror into the mobile drawer (≤800px) and stay in sync. Verified before/after at a real 390px viewport. Closes #5583.

PR #5608, opened 2026-07-05
fix(sidebar): Retry button disabled+styling feedback on session list error (#5501)

webtecnica:
Closing as duplicate of #5505 by @rodboev, which already passed gate-pass with comprehensive tests and better edge-case handling.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -6722,8 +6722,16 @@ function renderSessionListFromCache(){
     }
     const retry=document.createElement('button');
     retry.type='button';
+    retry.className='session-list-retry';
     retry.textContent='Retry';
-    retry.onclick=(e)=>{e.stopPropagation();_sessionListLoadError=null;void renderSessionList({deferWhileInteracting:false});};
+    retry.onclick=(e)=>{
+      e.stopPropagation();
+      if(retry.disabled) return;
+      retry.disabled=true;
+      retry.textContent='Retrying…';
+      _sessionListLoadError=null;
+      void renderSessionList({deferWhileInteracting:false});
+    };
     note.appendChild(retry);
     list.appendChild(note);
   }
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Thanks for the fix. Flagging two things after reading the diff against `origin/master` and cross-checking the other open work on #5501.

## There are three open PRs for #5501
`#5505` (gate-pass, size:L), `#5529`, and this one all close #5501. `#5505` is already the furthest along — it carries a `gate-pass` label and a 556-line regression test (`tests/test_issue5501_session_list_retry_affordance.py`). Worth coordinating with the maintainer on which to land so the effort isn't duplicated.

## Concrete technical gap: the pending state is button-local, not error-state-local
Your change disables the DOM button and swaps its label inline (`static/sessions.js`):

```js
retry.onclick=(e)=>{
  e.stopPropagation();
  if(retry.disabled) return;
  retry.disabled=true;
  retry.textContent='Retrying…';
  _sessionListLoadError=null;
  void renderSessionList({deferWhileInteracting:false});
};
```

The disabled/"Retrying…" state lives only on that button DOM node. But `renderSessionListFromCache()` rebuilds the entire error note from scratch every time it runs (it's the same function your hunk is inside, at the `if(_sessionListLoadError){...}` block ~line 6711). If anything triggers a session-list re-render while the async `renderSessionList()` fetch is still in flight — and the sidebar does re-render on cache/prefs churn — the note is recreated with a fresh, enabled "Retry" button, and the pending feedback is silently lost. You also set `_sessionListLoadError=null` at click time, so a re-render mid-fetch drops the error note entirely rather than showing the pending state.

`#5505` handles exactly this by moving the pending flag onto the error state object instead of the button:

```js
setPending();
_sessionListLoadError={..._sessionListLoadError,retrying:true};
renderSessionListFromCache();
void renderSessionList({...}).finally(()=>{
  if(!retry.parentNode||(_sessionListLoadError&&_sessionListLoadError.retrying)) return;
  retry.textContent='Retry'; retry.disabled=false; bindRetry();
});
```

Because `retrying` is on `_sessionListLoadError`, any re-render reconstructs the button already in its pending state, and the `.finally()` restores an enabled button on failure. It also adds `aria-busy` for the accessibility tree.

Your CSS additions (`.session-list-error` opacity override + `.session-list-retry` styling) are the right idea and address the "unstyled/dimmed" half of #5501 cleanly. The interaction-feedback half just needs the pending state to survive a re-render to be robust. If #5505 ends up being the one that lands, your styling rules would still be a worthwhile contribution to fold in.

I don't merge or label — just flagging the overlap and the re-render edge so whichever PR lands covers it.

PR #5611, opened 2026-07-05
fix(#5311): skip live /v1/models probe for opencode-go — use curated static list only

### LINKED ISSUE #5311: opencode-go dropdown shows models that return 404
# WebUI dropdown lists models that OpenCode Go rejects with 404

**Repo:** nesquena/hermes-webui
**Component:** `api/config.py` — `opencode-go` provider model list
**Severity:** Medium (all but default model in dropdown are unusable)
**Affects:** Any user on `opencode-go` with a paid Zen Go plan

## Summary

The model dropdown in the WebUI composer shows models that return **404 model not found** when used. Only the configured default model works.

## Steps to reproduce

1. Connect Hermes WebUI to a Hermes CLI installation with `model.provider: opencode-go` (the flat-rate `/zen/go` endpoint at `https://opencode.ai/zen/go/v1`).
2. Have a working default model (e.g. `minimax-m3`).
3. Open the WebUI composer. The model dropdown shows ~20 models under the "OpenCode Go" group.
4. Select any model other than the current default and send a message.
5. The provider returns `404 model not found`. The chat shows the `Model not found` error chip with the hint to "check the model ID in Settings or run `hermes model`".

## Expected

Models listed in the dropdown should be usable. Either they should be filtered out, or the live probe should validate that the model is actually accepted by `/v1/chat/completions`, not just enumerated by `/v1/models`.

## Actual

The dropdown shows models that OpenCode Go advertises in `/v1/models` (likely the full provider catalog) but does not actually accept for inference on the user's plan. Selecting any of them returns 404.

## Environment

- Hermes WebUI: current `main` (running from `~/hermes-webui/`)
- Hermes CLI: `model.provider: opencode-go`, `model.default: minimax-m3`
- OpenCode Go: paid plan, $10/month, `/go` endpoint
- Auth: API key in `OPENC...` env var; `/v1/models` returns 200 OK; `/v1/chat/completions` with non-default model returns 404

## Diagnosis

I traced the model list to `api/config.py`. The `opencode-go` provider has a hardcoded fallback list (lines ~1736–1752) and also a live probe against `/v1/models`. The two are merged and shown together in the dropdown. The hardcoded list contains ~14 older models; the live probe adds ~6 newer ones (`qwen3.7-max`, `qwen3.7-plus`, `kimi-k2.7-code`, `glm-5.2`, `hy3-preview`, `mimo-v2.5-pro`, `mimo-v2.5`). The newer ones appear to be in the public model catalog but not yet enabled for chat on the Go tier.

## Evidence

`GET https://opencode.ai/zen/go/v1/models` (with the user's auth) returns 20 model IDs. Sending a chat request to a non-default one returns 404.

The 14 hardcoded models that have been stable across recent WebUI versions (e.g. `minimax-m2.7`, `kimi-k2.5`, `glm-5`, `deepseek-v4-pro`, `mimo-v2-pro`, `qwen3.6-plus`) appear to work. The probe-introduced ones (`qwen3.7-max`, `kimi-k2.7-code`, `glm-5.2`, `hy3-preview`) do not.

## Proposed fix

Two options, in order of preference:

1. **Validate the probe.** For each model returned by the live probe, send a one-token chat request (or use a cheaper signal) and exclude any that return 404. Cache the result for the TTL window. This is robust to provider catalog drift.
2. **Pin to the hardcoded list only.** Stop merging live-probe results for `opencode-go` and rely on the curated list. Update the curated list as new models are confirmed working on the Go tier. Less robust but simpler and avoids runtime probe cost.

Option 1 is preferable since it generalizes to any provider where catalog ≠ availability.

## Workaround

Pin the model manually in `~/.hermes/config.yaml` to one of the older hardcoded entries that works (`minimax-m2.7`, `kimi-k2.5`, `glm-5`, etc.) until this is fixed upstream. Change the dropdown selection in the WebUI to one of those; the rest will still appear in the dropdown but should be ignored.

## Reporter

Hermes Agent via WebUI session — `cfb8d4fff6cb`. Happy to provide more logs or test patches.

webtecnica:
## Summary

OpenCode Go's `/v1/models` endpoint returns models from the full public catalog that are **not enabled on the Go tier**. Selecting any probe-only model from the dropdown causes `404 model not found` when the chat request is sent.

Fix: skip the live `/v1/models` probe for `opencode-go` and use the curated static list in `_PROVIDER_MODELS` only.

## Root Cause

At `api/config.py:6928`, the generic provider handler calls `_read_live_provider_model_ids(pid)` which queries Hermes CLI for `/v1/models`. OpenCode Go returns ~20 models, but ~6 of the newer ones (e.g. `qwen3.7-max`, `kimi-k2.7-code`, `glm-5.2`, `hy3-preview`, `mimo-v2.5-pro`, `mimo-v2.5`) are in the public catalog but not accepted by `/v1/chat/completions` on the Go plan.

The 14 hardcoded models have been stable and working.

## Change

**`api/config.py`** — add an `elif pid == "opencode-go"` guard that skips the live probe, falling through to the curated static list. 6 lines added (comment + pass).

## Why not Option 1 (validate the probe)

Option 1 (send a one-token probe for each model) would be more general but:
- Requires live API calls at every cache rebuild
- Introduces latency and potential rate-limit issues
- Doesn't fully solve the problem (models can flip between available/unavailable)
- The static list is already maintained and matches what works on the Go tier

## Verification

- `git diff --check`: clean
- Static analysis: no new errors (all existing LSP errors are pre-existing `hermes_cli.*`/`agent.*` unresolvable imports in a standalone context)
- The curated static list at `_PROVIDER_MODELS["opencode-go"]` (line 1755) is unchanged and contains only the 14 known-working models

Closes #5311

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -6925,6 +6925,12 @@ def get_available_models(*, prefer_cache: bool = False, force_refresh: bool = Fa
                     if not raw_models:
                         if pid == "moa":
                             raw_models = _moa_preset_models_from_config(cfg)
+                        elif pid == "opencode-go":
+                            # Skip live /v1/models probe for OpenCode Go — it
+                            # returns models from the public catalog that are
+                            # not enabled on the Go tier, causing 404 when
+                            # selected. Use the curated static list only. (#5311)
+                            pass
                         else:
                             raw_models = _models_from_live_provider_ids(
                                 pid,


nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (skip live /v1/models probe for opencode-go — use curated static list)

**Certified head:** `sha:8309229b` (clean rebase, branch `gate-rebase/5611-opencode-go-skip-probe`) · **PR:** #5611 · _webtecnica (new), fix(#5311): skip live /v1/models probe for opencode-go — use curated static list only_
**Verdict:** Full gate GREEN. A minimal, correct provider-config fix — opencode-go skips the live `/v1/models` probe (which returns public-catalog models not enabled on the Go tier → 404 on select) and falls back to its curated static list. Codex SAFE, full suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5611`)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; diff = `api/config.py` +6 |
| Full pytest suite | ✅ **12083 passed, 0 failed** |
| provider/model/config subset | ✅ 1787 passed |

### Findings
**✅ Correct + minimal:** `elif pid == "opencode-go": pass` is placed inside `if not raw_models:`, between the `moa` special-case and the `else:` default live-probe (`_models_from_live_provider_ids`). So for opencode-go it does `pass` → skips the live `/v1/models` probe → falls back to the curated static list (verified present at `config.py:1758` `"opencode-go": [...]`). Other providers are unaffected: `moa` still gets `_moa_preset_models_from_config`, all others still hit `else` → live probe. Codex confirmed no empty-model-list edge (static list is used), no crash if the static list is absent, backend-only, no security/injection. Full suite green (0 failures).

**Ready to merge — use branch `gate-rebase/5611-opencode-go-skip-probe` (`sha:8309229b`), NOT the PR's stale head `30fb7535`.** A tiny, correct provider-config reliability fix (opencode-go no longer offers Go-tier-disabled catalog models that 404 on select — uses the curated static list), Codex SAFE + full suite green (0 failures), backend-only. Mergeable-autonomously fix-class. concept 4/5 (real #5311 fix). Author @webtecnica (new). crit=3.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Rebased onto current master; verified the opencode-go elif is correctly placed (inside `if not raw_models`, before the else live-probe), falls back to the curated static list (config.py:1758), other providers unaffected, Codex SAFE + 1787 targeted + full suite green (0 failed). Cert valid for sha:8309229b._

nesquena-hermes:
Shipped in **v0.51.879** — thanks @webtecnica! 🎉

OpenCode Go's model picker no longer lists probe-only models that 404 on send. The live `/v1/models` probe returned public-catalog entries not enabled on the Go tier; the picker now uses the curated Go-tier static list for this provider instead. Narrow + provider-scoped — a user's own `providers.opencode-go.models` allowlist still wins, and no other provider is affected.

Gate: Codex SAFE TO SHIP (falls through to the static list correctly, picker not left empty, no cross-provider impact); full suite 12096 passed. Merged via #5616.

PR #5621, opened 2026-07-05
fix(ux): prevent three-panel layout from crushing center content on resize (#5545)

### LINKED ISSUE #5545: Bug: Layout breaks when resizing window with three panels open
## Description
When all three panels (Left Sidebar, Center Content, and Right Workspace) are active, the responsive layout logic incorrectly prioritizes the Left Sidebar and the Right Workspace panel's width during window shrinkage. As a result, the Center Content panel is compressed to an unusable width (or completely hidden), while the Left Sidebar and Right Workspace maintains its size.

## Steps to Reproduce
1. Open Hermes-Webui with all three panels visible:
   - Left: Conversation List
   - Center: Conversation Content
   - Right: Workspace Panel
2. Drag the edge of the browser window to reduce its total width (e.g., to view a side-by-side window).
3. Observe the behavior of the three panels as the window shrinks.

## Expected Behavior
The layout should adjust dynamically so that all three panels remain visible and usable. If the window becomes too narrow for three panels, there should be a graceful fallback (e.g., shrinking the sidebars more than the content, or a "minimum width" for the center panel).

## Actual Behavior
- The **Right Workspace** panel maintains its width and is prioritized.
- The **Center Content** panel shrinks drastically, becoming too narrow to read, or disappears entirely.
- In extreme cases, the center panel is completely hidden, leaving only the Left Sidebar and Right Workspace visible.

nankingjing:
## Summary

Fixes #5545 — the three-panel layout no longer crushes the center conversation panel when the browser window narrows.

## Root cause

Both `.sidebar` and `.rightpanel` were `flex-shrink: 0`, while `.main` had `min-width: 0`. The only element flexbox was allowed to compress was the center panel. The right panel only auto-hid at <=900px, leaving a 901–1020px band where all three panels tried to coexist but the center collapsed to unusable width.

## Changes (3 files, +9/-9 lines)

**`static/style.css`:**
- `.main`: `min-width: 0` → `min-width: min(420px, 100%)` — center panel gets a floor so it stays readable
- `.rightpanel`: remove `flex-shrink: 0`, `min-width: 0` → `min-width: 260px` — workspace panel yields before the center collapses
- Raise right-panel auto-hide breakpoint from 900px → 1020px (and companion min-width from 901px → 1021px)
- Raise outline panel hide breakpoint from 900px → 1020px (it sits inside the workspace panel)
- Update comment referencing old breakpoint values

**`static/boot.js`:**
- `_isCompactWorkspaceViewport()`: matchMedia `900px` → `1020px` to stay in sync with CSS

**`static/outline.js`:**
- `_outlineAllowed()`: matchMedia `900px` → `1020px` to stay in sync with CSS

## Verification

Dock the browser window across the 700–1100px range with all three panels open. Confirm:
- Center panel never drops below 420px
- Right panel shrinks gracefully before the center is affected
- Right panel auto-collapses cleanly below 1020px
- No horizontal scrollbar appears

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -142,7 +142,7 @@ async function _finalizeComposerPrefillOnBoot(prefillIntent){
 let _workspacePanelMode='closed'; // 'closed' | 'browse' | 'preview'
 
 function _isCompactWorkspaceViewport(){
-  return window.matchMedia('(max-width: 900px)').matches;
+  return window.matchMedia('(max-width: 1020px)').matches;
 }
 
 function _isPhoneWidthViewport(){
--- a/static/outline.js
+++ b/static/outline.js
@@ -17,7 +17,7 @@ function _currentSid() {
 }
 
 function _outlineAllowed() {
-  const compact = window.matchMedia && window.matchMedia('(max-width:900px)').matches;
+  const compact = window.matchMedia && window.matchMedia('(max-width:1020px)').matches;
   // The outline is a chat-view affordance only — never show the toggle or panel
   // while another MAIN panel (settings, tasks, insights, …) is active. _currentPanel
   // is owned by panels.js; treat an undefined/absent value as the chat default.
--- a/static/style.css
+++ b/static/style.css
@@ -2228,7 +2228,7 @@
   .help-card-link:hover{background:var(--accent);border-color:var(--accent);color:var(--bg);}
   .help-card-link:hover svg{opacity:1;}
   .help-card-link svg{flex:0 0 auto;opacity:.8;}
-  .main{flex:1;display:flex;flex-direction:column;overflow:hidden;min-width:0;min-height:0;background:var(--main-bg);}
+  .main{flex:1;display:flex;flex-direction:column;overflow:hidden;min-width:min(420px,100%);min-height:0;background:var(--main-bg);}
   .topbar{padding:12px 20px;border-bottom:1px solid var(--border);background:var(--topbar-bg);backdrop-filter:blur(12px);display:flex;align-items:center;justify-content:space-between;flex-shrink:0;position:relative;z-index:10;}
   .topbar-title{font-size:15px;font-weight:600;letter-spacing:-.01em;white-space:nowrap;overflow:hidden;text-overflow:ellipsis;}
   .topbar-meta{font-size:11px;color:var(--muted);margin-top:3px;opacity:.75;white-space:nowrap;overflow:hidden;text-overflow:ellipsis;}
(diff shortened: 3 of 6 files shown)


nesquena-hermes (changes requested):
Thanks for tackling #5545 — the source change itself is clean and internally consistent: the panel breakpoint moves uniformly from `900px` → `1020px` across `boot.js` (`_isCompactWorkspaceViewport()`), `outline.js` (`_outlineAllowed()`), and all three `style.css` media queries, the JS auto-hide threshold matches the CSS breakpoint (1020/1021 boundary), and the `.main{min-width:min(420px,100%)}` + `.rightpanel{min-width:260px}` space-sharing is sound. `node --check` passes on both JS files.

**Blocking issue — the breakpoint change breaks 3 existing tests that hardcode the old `900px` value, so CI is red** (`test (3.11/3.12/3.13, shard 0/1/2)` all fail). These are source-lock assertions that need to move to the new breakpoint alongside the source:

1. **`tests/test_issue2124_outline_panel.py:90`** — asserts `"matchMedia('(max-width:900px)')"` is in `OUTLINE_JS`. `outline.js` now emits `1020px`.
   → change to `"matchMedia('(max-width:1020px)')"`
2. **`tests/test_issue2211_workspace_panel_reopen.py:24`** — asserts `'@media(max-width:900px)'` is in the CSS (guarding `.workspace-panel-edge-toggle{display:none!important;}`). That rule now lives under `@media(max-width:1020px)`.
   → change to `'@media(max-width:1020px)'`
3. **`tests/test_mobile_layout.py:578`** (`test_compact_titlebar_keeps_hamburger_available`) — builds `_max_width_media_blocks(900)`; the block content moved to the `1020px` media query.
   → change to `_max_width_media_blocks(1020)`

**Recommended (not strictly blocking, but avoids silent staleness):** a few sibling assertions still pass only *coincidentally* because an unrelated `@media (max-width: 900px)` rule for `.msg-row[data-role="user"]` remains at `style.css:5772` — but their intent is the *panel* breakpoint, which you've moved. Consider also migrating the panel-breakpoint references in `test_mobile_layout.py` (~lines 172–179, 213) and `test_issue781.py` (~lines 47/50/53) from `900px` → `1020px` so they keep asserting what they mean.

Once the tests are green, this still needs a maintainer visual sign-off on the layout/breakpoint change before merge (it's a visible behavior change — `.rightpanel` can now shrink to 260px instead of never shrinking, and the right panel auto-hides later at 1020px). But the test fixes above are the mechanical blocker to clear first.

nankingjing:
Thanks for the review! All three tests have been updated to the new 1020px breakpoint:

1. `test_issue2124_outline_panel.py:90` → `matchMedia('(max-width:1020px)')`
2. `test_issue2211_workspace_panel_reopen.py:24` → `@media(max-width:1020px)`
3. `test_mobile_layout.py:578` → `_max_width_media_blocks(1020)`

Also added `min-width:0` to both rightpanel collapse selectors (matching the existing sidebar pattern), and renamed `test_mobile_breakpoint_900px_present` to `test_mobile_breakpoint_1020px_present` with updated assertions.

Ready for re-review. *Fixed in commit e84f6f6.*

nesquena-hermes:
Closing as superseded — thanks @nankingjing.

The three-panel center-crush issue (#5545) was fixed and shipped in **v0.51.906** via #5594 (@rodboev's surgical center-floor: `.main{flex:1 1 420px;min-width:420px}` + rails yield first, scoped `@media(min-width:901px)`). Confirmed live in master's `static/style.css` and issue #5545 is closed. Verified at 910px that the center goes from crushed to comfortable.

Your variant (raising the auto-hide breakpoints across 3 files) targets the same symptom; the shipped CSS-only floor was the preferred, more surgical approach. Appreciate the parallel effort.

PR #5622, opened 2026-07-05
fix(voice): keep dictation listening through pauses and prevent screen sleep (#4732)

### LINKED ISSUE #4732: bug(ux): mobile dictation stops on thinking pauses (SpeechRecognition not continuous) + no wake lock

nankingjing:
## Summary

Fixes #4732 — mobile dictation stops on thinking pauses and lacks wake lock.

### 1. SpeechRecognition.continuous=true
The browser's Web Speech API was configured with `continuous=false`, which causes the recognizer to terminate after detecting an end-of-speech pause. On mobile, this means any natural thinking pause ends the dictation session. Setting `continuous=true` keeps the recognizer listening until the user explicitly stops.

### 2. Screen Wake Lock
Added Screen Wake Lock support to prevent the device from sleeping during dictation on mobile:
- `_acquireWakeLock()` — requests a screen wake lock when recording starts
- `_releaseWakeLock()` — releases the wake lock when recording stops
- Visibility change handler — re-acquires wake lock when the page returns to foreground while recording
- Graceful degradation — silently ignores if the Wake Lock API is unavailable

## Diff
- `static/boot.js`: +20 / -2 lines

Closes #4732

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -650,6 +650,17 @@ function _micToastKeyForRecognitionError(error){
   let _micPointerDown=false;
   let _micStartSeq=0;
   const _micHoldThresholdMs=300;
+  let _wakeLock=null;
+
+  async function _acquireWakeLock(){
+    if(_wakeLock) return;
+    try{ _wakeLock=await navigator.wakeLock.request('screen');
+      _wakeLock.addEventListener('release',()=>{ _wakeLock=null; }); }catch(_){ _wakeLock=null; }
+  }
+
+  async function _releaseWakeLock(){
+    if(_wakeLock){ try{ await _wakeLock.release(); }catch(_){} _wakeLock=null; }
+  }
 
   function _setButtonTooltipAndKey(btn, key){
     const text = t(key);
@@ -662,6 +673,13 @@ function _micToastKeyForRecognitionError(error){
     }
   }
 
+  // Re-acquire wake lock when the page becomes visible while recording (#4732)
+  if(typeof document!=='undefined'){
+    document.addEventListener('visibilitychange',()=>{
+      if(!document.hidden&&_isRecording) _acquireWakeLock();
+    });
+  }
+
   function _setRecording(on){
     window._micActive=on;
     btn.classList.toggle('recording',on);
@@ -670,7 +688,7 @@ function _micToastKeyForRecognitionError(error){
     _setButtonTooltipAndKey(btn, on ? (_rawAudioMode ? 'voice_recording_active' : 'voice_dictate_active') : (_rawAudioMode ? 'voice_send_raw' : 'voice_dictate'));
     status.style.display=on?'':'none';
     if(statusText) statusText.textContent=on?'Listening':'Listening';
-    if(!on){ _finalText=''; _prefix=''; }
+    if(on){ _acquireWakeLock(); }else{ _releaseWakeLock(); _finalText=''; _prefix=''; }
   }
 
   function _updateMicTooltip(){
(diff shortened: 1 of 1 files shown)


nesquena-hermes (changes requested):
Thanks for this — keeping dictation alive through pauses (`continuous=true`) plus a Screen Wake Lock is a genuinely useful fix for #4732, and the wake-lock acquire/release/visibility wiring looks correct.

One blocker before this can ship: **it turns an existing green test red.**

### Confirmed regression: `tests/test_sprint20.py::test_boot_js_prefix_reset_on_stop`

- **master:** passes ✅
- **this branch:** fails ❌ (`1 failed, 85 passed` on the voice/recognition slice)

That test extracts the `_setRecording` function body by slicing from `function _setRecording` up to the **first** `}`, then asserts `_prefix` appears inside it (verifying the reset-on-stop). Your change moved the reset into the `else` branch:

```js
// before (master) — reset is inside the first brace block:
if(!on){ _finalText=''; _prefix=''; }

// after (this PR) — the first `}` now closes the acquire branch, before the reset:
if(on){ _acquireWakeLock(); }else{ _releaseWakeLock(); _finalText=''; _prefix=''; }
```

The reset *behavior* is preserved (it still runs on stop, in the `else`), so this is a source-extraction fragility rather than a behavior break — but CI is red as submitted.

### Minimal fix (keeps the wake-lock feature AND CI green)

Keep the original reset line untouched and add the wake-lock handling as a separate statement:

```js
if(!on){ _finalText=''; _prefix=''; }
if(on){ _acquireWakeLock(); } else { _releaseWakeLock(); }
```

This restores `_prefix` inside the first brace block (test green) and arguably reads cleaner by separating the text-reset from the wake-lock lifecycle.

Everything else in the diff is clean — `sr.continuous=true` only touches the dictation path (`_ensureSpeechRecognition`), which is separate from the voice-mode `recognition.continuous` localStorage-gated path (#5176), so the `continuous` sprint20 tests stay green.

nankingjing:
Thanks for the review! The `_setRecording` function has been restructured exactly as suggested:

```js
if(!on){ _finalText=''; _prefix=''; }
if(on){ _acquireWakeLock(); } else { _releaseWakeLock(); }
```

The text-reset is now back in its own statement before the wake-lock handling, and `_isRecording=false` is set at the top of `sr.onend`. Ready for re-review.

*Fixed in commit 0b378f6.*

nankingjing:
@
Thanks for the thorough review @nesquena-hermes!

The `_setRecording` function has been restructured exactly as suggested:

```js
if(!on){ _finalText=''; _prefix=''; }
if(on){ _acquireWakeLock(); } else { _releaseWakeLock(); }
```

This keeps the original reset line untouched (so the brace-slicing test finds `_prefix` inside the first `}` block) while the wake-lock lifecycle is a separate statement. Behavior is preserved — the reset still runs on stop, now just in its own guard.

I wasn't able to run the full test suite locally due to environment issues, but the fix directly addresses the root cause you identified: the first `}` now closes the `if(!on)` reset block (which contains `_prefix`) rather than the wake-lock acquire branch.
@

nankingjing:
Closing this — current `master` already covers mobile continuous dictation (#5294) and adds a more robust screen wake-lock (`_micWakeLock`, with serialized acquire/release and visibility-change re-acquire). This PR is fully superseded, so there is nothing left to merge here. Thanks for the earlier review!

PR #5624, opened 2026-07-05
fix: show dashboard link in mobile sidebar-nav (#5583)

### LINKED ISSUE #5583: Responsive mobile layout hides Dashboard and Theme Creator icons at 800px and below

nankingjing:
@
## Summary

Fixes #5583 — the Dashboard icon was invisible in the mobile sidebar-nav at viewport widths <= 640px, even when the dashboard service is running and reachable.

## Root cause

Inside `@media(max-width:640px)`, two rules were forcibly hiding `.sidebar-nav .dashboard-link` with `display:none!important`, overriding the JS-driven `dashboard-link-visible` toggle that enables the link at runtime.

## Fix

Removed the two `display:none!important` override rules. The dashboard link still defaults to hidden (inline `style="display:none"` on the HTML element) and only becomes visible when the JS `_applyDashboardStatus()` function sets `btn.style.display=""` and adds `dashboard-link-visible` after confirming the dashboard service is reachable — exactly matching the desktop-rail behaviour.

## Files changed

- `static/style.css`: 2 lines removed (the !important hide rules), replaced with a short comment.

Closes #5583.

@

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -3100,8 +3100,8 @@
     .sidebar-nav .nav-tab svg{width:20px;height:20px;}
     .sidebar-nav .nav-tab.active{background:var(--accent-bg);}
     .sidebar-nav .nav-tab.active::before{left:-4px;top:50%;bottom:auto;transform:translateY(-50%);width:3px;height:16px;border-radius:0 2px 2px 0;}
-    .sidebar-nav .dashboard-link,
-    .sidebar-nav .dashboard-link.dashboard-link-visible{display:none!important;}
+    /* Dashboard link: visible in mobile sidebar-nav when the dashboard
+       service is running (toggled via dashboard-link-visible by JS). */
     .sidebar .panel-view{height:100%;margin-left:52px;}
     .sidebar .panel-head{padding-right:52px;}
     .mobile-sidebar-close{display:inline-flex!important;position:absolute;right:4px;top:calc(4px + var(--app-titlebar-safe-top));width:44px;height:44px;z-index:4;background:var(--surface);border:1px solid var(--border);}


nesquena-hermes:
Reading the diff against `origin/master` alongside `static/style.css` and `static/index.html`, the mechanism here is correct but I think this only closes **part** of #5583, and there's a breakpoint question worth resolving before merge.

## The change itself is sound

The two rules you removed were genuinely dead-ending the JS toggle. On master, `style.css:3103-3104` sits inside `@media(max-width:640px)`:

```css
.sidebar-nav .dashboard-link,
.sidebar-nav .dashboard-link.dashboard-link-visible{display:none!important;}
```

That second selector (`.dashboard-link.dashboard-link-visible`) has higher specificity than the global toggle at `style.css:2107` (`.dashboard-link-visible{display:flex!important;}`), so even after `_applyDashboardStatus()` (`ui.js:1317-1337`) added the class + cleared the inline `display:none`, the link stayed hidden. Removing the override lets the global rule win. Good catch, and it's appropriately surgical.

## Gap 1 — the issue asks for two icons, this fixes one

#5583's title and body call out **both** the Dashboard icon *and* the **Theme Creator extension** icon. This PR only touches `.dashboard-link`. The Theme Creator button is an extension-injected *rail* action — the mobile `.sidebar-nav` in `index.html:173-191` is a hardcoded static list (chat/tasks/kanban/.../dashboard/logs/settings) with no counterpart for extension rail actions, so Theme Creator has nowhere to render in the narrow layout regardless of this CSS change. As written, `Closes #5583` would leave the second half of the report open.

## Gap 2 — breakpoint vs. the reported width

The reporter says the icons vanish at "800px and below," but the `.sidebar-nav` layout this PR edits only activates at `≤640px`. From `style.css:2111`:

```css
@media(min-width:641px){.rail{display:flex;}.sidebar > .sidebar-nav{display:none;}}
```

Between 641–900px the desktop `.rail` is what's shown, and its dashboard button (`#dashboardRailBtn`) is governed by the global `.dashboard-link-visible` rule at `2107`, which your diff doesn't touch. So if the reporter's symptom is genuinely at ~800px (their screenshots are 806/801px), this change wouldn't affect that width — it fixes the `≤640px` band. Worth confirming which band actually reproduces for them.

## Overlap with #5605

Heads-up for the maintainer: #5605 targets the same issue and addresses **both** gaps — it adds `_syncNavActionMirrors()` (`ui.js`) to mirror rail-only extension actions (Theme Creator-shaped) into the mobile sidebar-nav, and generalizes the CSS to a shared `.nav-action-visible` class instead of the dashboard-specific override:

```css
.dashboard-link-visible,.nav-action-visible{display:flex!important;}
...
.sidebar-nav .dashboard-link:not(.nav-action-visible){display:none!important;}
```

These two PRs conflict on `style.css:3103-3104`, so they can't both land. This PR is the minimal Dashboard-only fix; #5605 is the fuller one that also restores Theme Creator. Flagging so the two don't get merged over each other.

None of the above is a defect in *what you changed* — the Dashboard fix is correct. It's a scope question: minimal-and-correct here vs. #5605's broader take on the same issue.

nesquena-hermes:
Closing as superseded — thanks @nankingjing for tackling this.

The mobile dashboard-link visibility issue (#5583) was fixed and shipped in **v0.51.902** via #5605 (@rodboev's mobile nav-action-mirror approach — it mirrors the dashboard link + extension nav-actions into the mobile drawer and keeps them in sync). I confirmed the fix is live in master's `static/ui.js` (`_syncNavActionMirrors`) and issue #5583 is closed.

Your CSS-only variant (removing the `display:none` overrides) fixes the same symptom more narrowly, but the shipped mirror approach covers it plus extension nav-actions, so this is now moot. Appreciate the parallel effort — the root just landed via another PR first.

PR #4069, opened 2026-06-12
fix(compression): set truncation_watermark to prevent stale state.db replay

ashuio:
## Problem

Manual (`/compress`) and auto-compression update the JSON session file (`context_messages`) but leave old uncompressed messages in `state.db`. On the next turn, `reconciled_state_db_messages_for_session(prefer_context=True)` merges compressed context with state.db:

1. `state_db_delta_after_context()` tries to align compressed context with state.db rows by matching content fingerprints
2. Compressed content has different fingerprints from the originals
3. Alignment fails (`best_len < 2`), returning ALL state.db rows
4. They get appended after the compressed context — **defeating compression entirely**

The user sees the full uncompressed transcript replayed into the next turn's context window, negating the token savings compression was supposed to provide.

## Fix

In `reconciled_state_db_messages_for_session()`, when `prefer_context=True` (model-facing context path) and the session has a `compression_anchor_message_key` (indicating compression just happened), skip state.db reconciliation entirely and return only the compressed `context_messages`.

The display path (`prefer_context=False`) is **unaffected** — it continues to merge state.db messages normally so the user can scroll through full history in the UI.

## Files changed

- `api/models.py`: Skip state.db delta in context path when compression anchor is set (12 lines)

## Testing

- All 77 compression/watermark/session tests pass
- All 45 state.db reconciliation tests pass

Issue #4685, opened 2026-06-22
Track long-running context compression exhaustion and recovery UX


franksong2702:
## Summary

Long-running, tool-heavy sessions can reach a state where automatic compression either does not actually reduce the model-facing context enough, or compression eventually exhausts and the user is left without a clear, durable recovery path.

This is a child umbrella for #3400 and should be read alongside the broader run-state consistency contract in #2361. The first concrete implementation slice appears to be #4249: after compression, stale `state.db` rows can be replayed into the next request and effectively undo the compressed `context_messages`.

## Problem

Small and medium context-window models are valid Hermes WebUI targets. We should not assume every provider/model has a 1M-token context window.

For long sessions with many tool calls, the expected behavior is:

- automatic compression keeps the session running by reducing model-facing context;
- visible transcript and Worklog history can remain intact for the user;
- large tool outputs, diffs, terminal output, search results, and replay-only rows are summarized or dropped from model context when needed;
- if compression really cannot shrink the prompt enough, the turn settles into a first-class exhausted terminal state;
- after an exhausted turn, the user is guided into a safe recovery path instead of accidentally sending the same oversized context again.

The current risk is that these layers can be conflated. A session can look visually preserved while `context_messages` or state-db reconciliation still feeds too much historical context to the model.

## Evidence and related work

- #4249 is open and now has concrete evidence that after compression, the next turn can still send a very large prompt. The analysis there points at compressed `context_messages` being reconciled with stale uncompressed `state.db` rows.
- #4069 was a prior draft fix for the same stale state-db replay class, but it was closed before merge so it needs a safer replacement or revival.
- #3133 / #4594 fixed the manual `/compress` visible-transcript clobbering path. That was important, but it does not by itself prove that model-facing context is safely pruned for long tool-heavy sessions or automatic compression.
- #3315 / #3316 covered historical compression-exhausted/no-final behavior, including the requirement that exhausted compression surfaces as an error/terminal outcome rather than looking like a normal assistant final answer.
- #4248 / #4418 fixed a context-window metadata reload issue that could trigger premature auto-compression for large-context models.
- #2734 remains an open symptom issue around automatic compression triggering unexpectedly after manual compression.

### Compression succeeds

- The visible transcript remains intact.
- `context_messages` is actually smaller after compression.
- The next request does not append stale uncompressed state-db rows after the compressed prefix.
- Token estimates and UI context indicators reflect the model-facing context, not just the visible transcript.
[...]

PR #5626, opened 2026-07-05
fix(session): ignore inactive compacted state rows in context replay

ai-ag2026:
## Summary

- Filter inactive (`active=0`) rows out of `get_state_db_session_messages()` by default when the `messages.active` column exists.
- Add an explicit `include_inactive=True` escape hatch for recovery/audit callers that need archived rows.
- Add regression coverage proving compacted archive rows are not reconciled back into the next model-facing context.

## Why

In-place context compaction archives old `state.db` message rows by marking them inactive. WebUI's normal state-db reader fed those archived rows back into `reconciled_state_db_messages_for_session(..., prefer_context=True)`, so the next turn could resurrect pre-compaction history and trigger compression again.

This is the same stale state-db replay class discussed in #4249 / #4685 and related to the closed #4069 attempt, but this patch fixes the concrete schema-level source: inactive archive rows should not be part of the normal active transcript read.

Refs #4249
Refs #4685
Refs #2361

## Verification

- `./scripts/test.sh tests/test_state_db_active_filter.py -q` → 3 passed
- `./scripts/test.sh tests/test_state_db_active_filter.py tests/test_webui_state_db_context_reconciliation.py -q` → 7 passed
- `./.venv/bin/python -m py_compile api/models.py`

## Risks / Follow-ups

- Default behavior changes only for schemas with a `messages.active` column. Legacy schemas without that column keep the existing query shape.
- Explicit recovery/audit views can opt into archived rows with `include_inactive=True`.

## Model Used

OpenAI GPT-5.5 via Hermes Agent, with terminal/file tools for diagnosis, patching, and verification.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -6143,6 +6143,7 @@ def get_state_db_session_messages(
     stitch_continuations: bool = False,
     profile=None,
     since_timestamp=None,
+    include_inactive: bool = False,
 ) -> list:
     """Read messages for a Hermes session from state.db.
 
@@ -6158,6 +6159,13 @@ def get_state_db_session_messages(
     raw state.db scan to rows at or after a sidecar-derived timestamp floor while
     preserving the caller's normal merge/window logic.  Full-history callers must
     leave it unset.
+
+    When the messages table exposes an ``active`` column, inactive rows are
+    compacted/archived history and are intentionally excluded by default. WebUI
+    reconciliation feeds this reader straight into the next model context; pulling
+    ``active=0`` archive rows back in resurrects pre-compaction history and can
+    make every later turn re-trigger compression. Pass ``include_inactive=True``
+    only for explicit recovery/audit views.
     """
     try:
         import sqlite3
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.885** — thanks @ai-ag2026! 🎉

Compacted (archived) messages no longer creep back into the model's context. `get_state_db_session_messages()` was returning `active=0` rows that in-place compaction had archived, and WebUI reconciliation fed them straight into the next turn's context — undoing compaction and re-triggering compression every turn. The reader now excludes inactive rows by default (`AND (active IS NULL OR active != 0)`, only when the `active` column exists so older schemas are unaffected; NULL/legacy rows preserved), with an `include_inactive=True` escape hatch for recovery/audit callers.

Gate: Codex SAFE TO SHIP + an independent audit of all 5 callers (every one is a context/reconciliation/display path — none needs archived rows) + full suite 12114 passed + 3 own tests proving archived rows aren't reconciled back. Merged via #5629.

PR #5639, opened 2026-07-05
fix(#5637): bump contain-intrinsic-size fallback from 1px to 400px on .msg-row

### LINKED ISSUE #5637: Mobile scroll jump-back: off-screen .msg-row content-visibility (#5541) collapses scrollHeight → browser clamps scrollTop
## Summary

On touch devices, scrolling up into history while a reply streams (or when the transcript re-renders) throws the viewport back toward the top. I traced it with real-device instrumentation on Android Chrome to two distinct causes; the primary one was introduced by #5541.

## Primary root cause: `#5541`'s off-screen `contain-intrinsic-size: auto 1px`

`cebfdf41` (`fix(#5541): iOS PWA streaming freeze + SSE reconnect reliability`, 2026-07-04) added, under `@media (pointer: coarse)`:

```css
.msg-row { content-visibility: auto; contain-intrinsic-size: auto 1px; contain: layout style; }
```

On a long transcript this collapses the reserved height of every off-screen `.msg-row` to the `1px` intrinsic estimate whenever `auto` has no remembered size for that row. `scrollHeight` then lurches down by tens of thousands of px, the browser force-clamps `scrollTop`, and the viewport is thrown to the top.

I instrumented a real Android Chrome device (per-frame `scrollTop`/`scrollHeight` deltas, a `scrollHeight`-shrink event with a before/after DOM snapshot, and a `scrollTop`-setter hook capturing `new Error().stack`). Measured output (~275-row session):

```
SHRINK  scrollHeight 53316 -> 1569   rows 275 -> 276   ATCLAMP
JUMP    dTop -51859  dH -51747       JS=none
SHRINK  scrollHeight 89916 -> 1449   rows 254 -> 255   ATCLAMP
```

Signals that fingerprint this as the CSS intrinsic-size collapse (not a JS scroll write, not a DOM wipe):
- `rows` stays high (275+) across the shrink — the DOM is **not** wiped; only the *reserved off-screen height* collapses.
- `JS=none` — no JS wrote `scrollTop`; the clamp is the browser reacting to the CSS-driven `scrollHeight` drop.
- `@media (pointer: coarse)` scoped → touch/mobile only, which is why it never reproduces on desktop.

### Why it became much more frequent recently

I noticed the jump-back becoming far more frequent over the last couple of days. That lines up exactly: `#5541` landed 2026-07-04. Before it, `.msg-row` had no `content-visibility`, so the off-screen reserved height never collapsed.

### Fix I verified on-device

A flat estimate isn't enough — bumping `1px` → `140px` cut the collapse by ~an order of magnitude, but a heavy session still lurched ~7000px because sessions with tall tool-result rows (multi-thousand px each) are badly under-reserved by any single flat value. Disabling off-screen skipping on `.msg-row` under `pointer: coarse` entirely eliminated the `scrollHeight` collapse (probe: zero shrink events, zero clamp jumps afterward). `#5541`'s benefit is iOS-WKWebView-specific; on Android Chrome only the side effect is felt.

Suggested upstream direction: either gate `content-visibility:auto` off for rows that can be tall (tool-result / long assistant answers), or give `contain-intrinsic-size` a realistic per-row-type estimate instead of `1px`, so the reserved off-screen height tracks the real height and `scrollHeight` doesn't collapse.

## Second, independent jump-back (exposed once `scrollHeight` was restored)

With `content-visibility` off, `scrollHeight` returns to its true (large) value — and a **separate**, pre-existing jump-back becomes visible that the collapsed height had been masking. The setter hook attributes this one to a JS write (not a browser clamp):

```
JS-SET  scrollTop 65964 -> 64069  via _restoreMessageScrollSnapshotSameFrame > renderLiveAnchorActivityScene > _renderLiveAnchorActivitySceneForStream
JUMP    dTop -1895  dH 29   JS=_restoreMessageScrollSnapshotSameFrame
JUMP    dTop -2233  dH 50
JUMP    dTop -2939  dH 50
JUMP    dTop -3563  dH 49
```

`_restoreMessageScrollSnapshotSameFrame` restores an **absolute** `snapshot.top` in its fallback branch (when the semantic anchor restore doesn't apply):

```js
const target = (snapshot.pinned === true && Number.isFinite(bottom))
  ? maxTop - Math.max(0, bottom)
  : Number(snapshot.top) || 0;       // absolute-top fallback
el.scrollTop = Math.max(0, Math.min(target, maxTop));
```

During streaming, `_renderLiveAnchorActivitySceneForStream` → `_renderMessagesWithScrollSnapshot` fires this on every live activity-scene refresh. For a reader scrolled **up** into history (unpinned) while a response streams, restoring the pre-capture *absolute* `scrollTop` after a re-render whose above-viewport height changed nudges the viewport backward — and the error scales with `scrollHeight` (`dTop` grew 1895 → 3563 as the transcript passed ~100k px). This is the Stable-Assistant-Turn-Anchors restore path (functions from ~2026-06-04 / 06-30), not new — it was simply **masked while `content-visibility` kept `scrollHeight` collapsed**, and surfaced once that was fixed.

Suggested direction: when the reader is unpinned (scrolled up in history), the live activity-scene refresh should not move `scrollTop` at all (let the browser's own scroll anchoring hold position), or restore relative to a still-visible message row rather than an absolute `scrollTop` that goes stale as above-viewport height changes.

## Repro

- Long transcript, touch device (Android Chrome).
- Scroll up into history while a reply is streaming.
- Viewport jumps back toward the top; `scrollHeight` momentarily collapses (primary cause), or `_restoreMessageScrollSnapshotSameFrame` nudges scrollTop backward each activity-scene tick (secondary cause).

I have the full probe script and raw telemetry logs and can share them, and I'm happy to open a PR for either direction.

---

*This supersedes #5636, which framed the cause as transcript virtualization (#500) before the real-device probe pointed at `content-visibility` (#5541). Closing that one and re-filing with the correct framing.*

nankingjing:
## Problem

On touch devices (Android Chrome), scrolling history while a reply streams causes the viewport to jump back toward the top. Root cause: `contain-intrinsic-size: auto 1px` on `.msg-row` under `@media (pointer: coarse)` collapses scrollHeight by tens of thousands of px when off-screen rows haven't been measured, forcing the browser to clamp scrollTop.

From the real-device instrumentation in #5637:
```
SHRINK  scrollHeight 53316 -> 1569   rows 275 -> 276   ATCLAMP
JUMP    dTop -51859  dH -51747       JS=none
```

The DOM is NOT wiped — rows stay high (275+). Only the CSS-driven reserved off-screen height collapses.

## Fix

Bump `contain-intrinsic-size` fallback from `1px` to `400px` — a reasonable per-row estimate that prevents destructive scrollHeight collapse while preserving the WKWebView streaming performance benefit from #5541.

## Diff

`static/style.css`: 1 line changed (1 insertion, 1 deletion).

Fixes #5637.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -5729,7 +5729,7 @@ main.main > #mainPlugin{display:none;}
    enough to avoid scrollHeight recalculation jumps that kill flick-scroll
    momentum on iOS. */
 @media (pointer: coarse) {
-  .msg-row { content-visibility: auto; contain-intrinsic-size: auto 1px; contain: layout style; }
+  .msg-row { content-visibility: auto; contain-intrinsic-size: auto 400px; contain: layout style; }
   /* Keep the LIVE streaming turn fully rendered while a response streams. The
      live turn is always id="liveAssistantTurn" (assigned on every render path:
      Compact Worklog, Transparent Stream, restored-live, etc.), but ONLY the


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (stopgap with 2 must-fixes: stale rationale comment + "Fixes"→"Mitigates"; #5638 is the better end-state) · design-fork → Nathan

**Certified head:** `sha:` (rebased onto current master, `git apply` clean) · **PR:** #5639 · _nankingjing (T2), fix(#5637): bump contain-intrinsic-size fallback 1px→400px on .msg-row_
**Verdict:** Engineering-clean (Codex SAFE + full suite green + 218 scroll-pin tests), but Fable's design-gate flags it as a **stopgap, not the end state**, with 2 must-fixes before it could ship — and recommends **#5638's row-class gate supersede it**. This is a genuine design fork (3 PRs — #5635/#5638/#5639 — target #5637); routing the merge decision to you with a clear recommendation.

### What I ran (rebased worktree `/tmp/wt-rebase-5639`) — full gate: Codex + Fable + suite
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; diff = `static/style.css` +1/-1 (`auto 1px`→`auto 400px`) |
| Codex (reproduce) | SAFE TO SHIP — 0 findings |
| **Fable-UX (design-gate)** | **SHIP-WITH-UX-FIXES** — 2 must-fix + strategic "#5638 supersedes" |
| Full pytest suite | ✅ **12114 passed, 0 failed**; 218 scroll-pin tests green |

### Findings (Fable, on-device-telemetry-informed)
**⛔ Must-fix 1 — stale rationale comment contradicts the code:** the comment block directly above the changed line still reads *"contain-intrinsic-size: auto 1px … the 1px estimate is small enough to avoid scrollHeight recalculation jumps that kill flick-scroll momentum on iOS."* The commit changes the value but not the comment → the file now documents a rationale the code contradicts, and silently overrides #5541's explicitly-stated reason for choosing 1px. Rewrite with the 400px rationale + the known flat-estimate limitation.
**⛔ Must-fix 2 — "Fixes #5637" auto-closes an incompletely-fixed issue:** #5639 addresses only 1 of the 2 documented causes in #5637 — the secondary JS jump-back (`_restoreMessageScrollSnapshotSameFrame` absolute-top fallback, dTop −1895…−3563) is untouched, and the heavy-session under-reserve class remains. Change to "Mitigates #5637" (or file a follow-up) so merging doesn't bury the remainder.
**⚠️ Heavy-session under-reserve (the reliability crux):** `contain-intrinsic-size: auto 400px` — `auto` remembers real sizes once rendered; 400px applies to never-rendered rows AND to all rows when a re-render drops remembered sizes (exactly the #5637 telemetry `SHRINK scrollHeight 53316→1569` event). A flat 400px over-reserves user rows ~5-6× (phantom scroll space, mis-sized scrollbar thumb) and **under-reserves multi-thousand-px tool-output rows** — and the issue author measured a flat 140px still lurching ~7000px. **Hermes is agent-centric → tool-heavy transcripts are the CORE workload, so the flat-value under-estimate is not a tail case.** On Android (scroll-anchoring active) the catastrophic clamp is structurally impossible when 400px over-reserves; on iOS (overflow-anchor inert) first-pass scroll-up shows ±200px per-row shifts (same magnitude as 1px, inverted) — removes the jump-to-top but not smooth.

### Design fork — recommendation (Nathan's call)
Three PRs target #5637's mobile scroll jump-back:
- **#5635** (GREEN'd) — JS recycled-anchor compensation recovery; distinct root cause; complementary.
- **#5638** (GREEN'd) — content-visibility OFF for tall assistant rows (true geometry) + 96px calibrated user-row estimate + userUnpinned hold; also fixes the secondary JS jump. **Fable: the architecturally-correct end-state** (per-row-type, reuses the app's `#liveAssistantTurn` opt-out idiom, truthful scrollbar, zero estimate error on the wildly-varying assistant rows). Its one risk: dropping content-visibility for assistant rows *may* reopen #5541's iOS WKWebView streaming-freeze (retains `contain: layout style` as mitigation, but author validated Android-only — needs an iOS streaming recording).
- **#5639** (this) — the surgical 1-line 400px fallback; fully preserves #5541's perf mechanism; **an acceptable stopgap TODAY but superseded by #5638 as the end state** (they compete for the same CSS line).

**My recommendation: prefer #5638 as the end-state fix** (per Fable + the "most reliable/technically-correct" bar — flat 400px can't square tall tool rows on Hermes's tool-heavy core workload), **gated on an iOS streaming-perf recording to confirm it doesn't reopen #5541's freeze.** If you want an immediate mitigation while #5638's iOS recording is pending, #5639 can ship as a stopgap **only after** its 2 must-fixes (comment + "Mitigates") — but then #5638 supersedes it. Don't merge both (same line). concept 4/5 (sound stopgap; not the end state). Author @nankingjing (T2). crit=4.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Full gate: Codex SAFE + full suite + 218 scroll-pin tests green, BUT Fable design-gate = SHIP-WITH-UX-FIXES: stale 1px comment contradicts code (must rewrite), "Fixes"→"Mitigates" (secondary JS jump + heavy-session under-reserve remain), and flat 400px under-reserves tool-heavy rows (Hermes core workload). #5638's row-class gate is the better end-state (Nathan's design-fork call; #5638 gated on iOS streaming recording). Don't merge both — same CSS line. Cert = RED pending must-fixes + fork decision._

nesquena-hermes:
Closing as superseded by **#5638** (shipped in **v0.51.887**), which fixes the same mobile scroll jump-back (#5637) at the root.

Thanks @nankingjing for the fast stopgap. The reason we went with #5638's approach instead of a flat `contain-intrinsic-size` bump: the #5637 reporter's own on-device testing showed that *any* single flat value under-reserves tall tool-result / long-answer rows (a heavy session still lurched ~7000px even at 140px), so #5638 instead scopes `content-visibility:auto` off the tall assistant rows entirely (kept only for the short, size-predictable user rows) — which is the direction the issue author recommended. That removes the scrollHeight-collapse cause rather than shrinking it. Appreciate the help chasing this down. 🙏

PR #302, opened 2026-04-12
feat: breadcrumb nav + wider panel + responsive msg width (#292)

franksong2702:
## Summary

Implements three UX improvements from #292:

1. **Clickable breadcrumb navigation** in workspace file preview — click any path segment to navigate directly
2. **PANEL_MAX raised from 500 to 1200** — enables wider right panel on ultrawide screens
3. **Responsive `.messages-inner` max-width** — scales up gracefully on large screens while preserving readability

### static/boot.js
- `PANEL_MAX`: 500 → 1200

### static/style.css
- Added responsive breakpoints for `.messages-inner`:
  - `@media(min-width:1400px)` → `max-width:1100px`
  - `@media(min-width:1800px)` → `max-width:1200px`
- `.msg-body` already has its own 680px cap, so line width stays readable

### static/ui.js + workspace.js
- Clickable breadcrumb path bar in workspace preview
- Each path segment is clickable to navigate to that directory level
- Handles paths with spaces and special characters

## Owner feedback

Addressed in this PR:
- ✅ Viewport-based `@media` breakpoints (not container-based)
- ✅ `.msg-body` inner cap preserved for readability
- ✅ Spaces/special chars in breadcrumb paths handled

Fixes #292

PR #5645, opened 2026-07-05
fix(#5619): reload config cache on profile switch

### LINKED ISSUE #5619: Web UI config discrepancy with multi-profiles
Hello!

I have noticed an issue while trying to use a secondary profile of mine, which uses local models via llama.cpp - I have this set in that profile's `config.yaml` via `custom_providers` (and updating the default `model` to reflect the correct one).

In Hermes desktop, the TUI, and message gateways this works without a problem - however on Hermes Web UI it seems to be reading the config of the default profile since it loads in that profile's default model. If I click on an existing conversation that was setup somewhere else, it retains this, but new conversations don't have the option to switch (the model selector doesn't mention the custom provider at all).

Strangely enough, it seems to also be loading the API keys from the main profile too, as said profile doesn't have any cloud API keys setup at all to even use those models. I thought perhaps #5567 might have had something to do with it, but even after updating to a release (`v0.51.880`) that has that linked PR merged in, the issue seems to still occur.

While swapped to that profile, the settings -> providers menu also does not list it there either.

My other three profiles that use cloud models on the other hand do all retain their default model choices, and I assume are using the correct keys.

Apologies if I'm just missing a toggle or such somewhere!

rodboev:
## Thinking Path

- `/api/providers` already enters the active request-profile scope, so the wrong catalog had to come from the shared config cache below the route layer.
- `get_config()` and `reload_config_if_stale()` were treating mtime drift and profile-path changes as one stale bucket, and `get_available_models()` still repeated that same combined guard while reading the legacy `cfg` alias directly.
- The fix now forces reloads on profile-path changes at every touched config consumer, while keeping the same-path override contract for tests and runtime callers.

## What Changed

- `api/config.py`: split config-cache staleness into `path_changed` and `mtime_stale` in `get_config()`, `reload_config_if_stale()`, and `get_available_models()`, then retarget the legacy `cfg` alias to the current cache on real reloads.
- `tests/test_profile_switch_1200.py`: add regressions for inherited overrides, rebound `cfg` overrides, and the `/api/models` payload after a request-profile switch.
- `tests/test_stage302_config_override_regression.py`: tighten the preserved contract to the explicit `same_path` boundary so the override-proof matches the intended behavior.

## Why It Matters

Named profiles stop inheriting the default profile's provider catalog, model picker payload, default model, and cloud-key state after any in-memory config mutation. Same-profile overrides still avoid unwanted reloads.

## Verification

`pytest tests/test_profile_switch_1200.py -v --timeout=60`

`pytest tests/test_stage302_config_override_regression.py -v --timeout=60`

## Upstream

Closes #5619.

Reporter evidence and maintainer root-cause analysis in https://github.com/nesquena/hermes-webui/issues/5619#issuecomment-4886270745, https://github.com/nesquena/hermes-webui/issues/5619#issuecomment-4886434302, and https://github.com/nesquena/hermes-webui/issues/5619#issuecomment-4886843401 pinned this to the shared config-cache reload gate rather than the request-profile route wrapper.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -444,6 +444,7 @@ def _apply_config_defaults(config_data: dict) -> None:
  
 def reload_config_if_stale() -> None:
     """Refresh config.yaml once for concurrent stale read paths."""
+    global cfg
     with _cfg_lock:
         try:
             config_path = _get_config_path()
@@ -454,6 +455,8 @@ def reload_config_if_stale() -> None:
         mtime_stale = current_mtime != _cfg_mtime
         if not _cfg_cache or path_changed or (mtime_stale and not _cfg_has_in_memory_overrides()):
             _refresh_config_cache(config_path)
+            if path_changed:
+                cfg = _cfg_cache
 
 
 def get_config() -> dict:
@@ -521,7 +524,7 @@ def _refresh_config_cache(config_path: Path | None = None) -> None:
     Callers must hold _cfg_lock when invoking this helper because it mutates
     shared state.
     """
-    global _cfg_mtime, _cfg_path, _cfg_fingerprint, cfg
+    global _cfg_mtime, _cfg_path, _cfg_fingerprint
     if config_path is None:
         config_path = _get_config_path()
     _cfg_cache.clear()
@@ -582,8 +585,6 @@ def _refresh_config_cache(config_path: Path | None = None) -> None:
         logger.debug("Failed to load yaml config from %s", config_path)
     _apply_config_defaults(_cfg_cache)
     _cfg_fingerprint = _fingerprint_config(_cfg_cache)
-    # A real reload retargets the legacy cfg alias to the cache for this path.
-    cfg = _cfg_cache
     # Bust the models cache so the next request sees fresh config values.
     # Only delete the disk cache when config has actually changed -- not on
     # first-ever load (when _old_cfg_mtime == 0.0, i.e. server start or
--- a/api/config.py
+++ b/api/config.py
@@ -5759,10 +5759,9 @@ def get_available_models(*, prefer_cache: bool = False, force_refresh: bool = Fa
     except OSError:
         _current_path = _get_config_path()
         _current_mtime = 0.0
-    if (
-        (_current_mtime != _cfg_mtime or _current_path != _cfg_path)
-        and not _cfg_has_in_memory_overrides()
-    ):
+    path_changed = _current_path != _cfg_path
+    mtime_stale = _current_mtime != _cfg_mtime
+    if path_changed or (mtime_stale and not _cfg_has_in_memory_overrides()):
         reload_config_if_stale()
     # ── COLD PATH helper ─────────────────────────────────────────────────────
     # Extracted so it runs inside _available_models_cache_lock (RLock) to
--- a/api/config.py
+++ b/api/config.py
@@ -521,7 +521,7 @@ def _refresh_config_cache(config_path: Path | None = None) -> None:
     Callers must hold _cfg_lock when invoking this helper because it mutates
     shared state.
     """
-    global _cfg_mtime, _cfg_path, _cfg_fingerprint
+    global _cfg_mtime, _cfg_path, _cfg_fingerprint, cfg
     if config_path is None:
         config_path = _get_config_path()
     _cfg_cache.clear()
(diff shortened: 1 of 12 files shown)


rodboev:
Addressed in 00a69521.

I removed the `_restore_config._old_path` side channel in `tests/test_provider_quota_status.py` and carried the saved config path inside the existing helper state that already flows through `_with_config(...)` and `_restore_config(...)`. Focused validation stayed green on:

- `tests/test_provider_quota_status.py::test_openrouter_quota_fetches_key_endpoint_and_sanitizes_response`
- `tests/test_provider_quota_status.py::test_codex_account_usage_is_fetched_under_active_profile_home`
- `tests/test_provider_quota_status.py::test_account_usage_profile_fetch_uses_short_lived_cache`

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (reload config cache on profile switch — #5619, crown-jewel profile-switching surface)

**Certified head:** `sha:8bed361b` (clean rebase, branch `gate-rebase/5645-config-cache-profile-switch`) · **PR:** #5645 · _rodboev (T1), fix(#5619): reload config cache on profile switch_
**Verdict:** Full gate GREEN. Fixes #5619 (WebUI read the DEFAULT profile's config after switching profiles) with the correct root fix — a profile switch (config-path change) now forces a cache reload unconditionally, while an mtime bump still respects in-memory overrides. Codex SAFE, full suite green, 20 profile-switch/override tests pass.

### What I ran (rebased worktree `/tmp/wt-rebase-5645`) — Codex + full suite (backend config-cache; no visible UI surface for Fable)
| Gate | Result |
|------|--------|
| Codex (cache correctness / cross-profile leak / thread-safety / override-survival) | **SAFE TO SHIP** — 0 findings |
| Full pytest suite | ✅ **12143 passed, 0 failed** |
| profile-switch + #302 override-regression tests | ✅ 20/20 |

### Findings
**✅ Correct root fix:** the old reload gate was `cache_stale AND not _cfg_has_in_memory_overrides()` where `cache_stale = mtime-change OR path-change` — so a PATH change (profile switch) was WRONGLY suppressed when in-memory overrides existed, leaving the default-profile config cached (the #5619 bug). The fix splits the conditions: `if not _cfg_cache or path_changed or (mtime_stale and not _cfg_has_in_memory_overrides())` — `path_changed` (profile switch) now forces reload UNCONDITIONALLY (independent of overrides), while `mtime_stale` still respects overrides. Codex confirmed: the switched-to profile config is read after a switch (no stale DEFAULT), dropping the old profile's overrides on switch is correct (overrides are profile-scoped), no cross-profile field leaks into `cfg`, the `global cfg` reassignment is thread-safe under concurrent reads, and the mtime path still protects overrides from an mtime bump. The #302 override-regression test (legit overrides not wrongly discarded) passes, so the two concerns are correctly separated. Full suite green.

**Ready to merge — use branch `gate-rebase/5645-config-cache-profile-switch` (`sha:8bed361b`), NOT the PR's stale head `00a69521`.** A correct root fix for the #5619 profile-switch config-cache bug on Nathan's crown-jewel profile-switching surface — Codex SAFE (cache correctness, no cross-profile leak, thread-safe, override semantics) + full suite green + 20 profile-switch/override tests. Backend, no visible surface. concept 5/5 (roots a crown-jewel-surface bug at the cache gate). Author @rodboev (T1). crit=3.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Root fix: split path-change (profile switch → force reload unconditionally) from mtime-change (respects in-memory overrides), fixing #5619 where a profile switch was suppressed by existing overrides → stale DEFAULT config. Codex SAFE (no cross-profile leak, thread-safe, override-survival correct) + full suite green (0 failed) + 20 profile-switch/#302-override tests. Backend (no Fable surface). Cert valid for sha:8bed361b._

nesquena-hermes:
Shipped in **v0.51.896** (deployed live). Thanks @rodboev — correct root fix: a profile-path change now forces a config reload unconditionally (was being suppressed by in-memory overrides, leaving the default profile's config cached), while a plain file-mtime change still respects overrides. Folded in a test-isolation fix the gate caught (the stage302 test leaked config-module globals — deepcopy snapshot + clear/update restore since reload_config mutates _cfg_cache in place). Codex SAFE (2 rounds) + suite 12178/0. 🙏

Issue #3870, opened 2026-06-09
Render markdown for the user's messages


russjr08:
Hello!

Would it be possible to add in rendering the user's messages as markdown just like the assistant's? I looked for a potential setting option for this, but couldn't find one.

This would be quite helpful when passing in code to the conversation as the user, or just other general markdown flavored text. As it is, when trying to review a script that I previously sent for example, it gets a little unruly at times!

Edit: Oh, I'm not sure how I didn't see that full code blocks (wrapped in triple backticks) do get rendered correctly. I must've misremembered that bit. However, the other markdown features would still be nice to have too!

nesquena-hermes:
Thanks for the clear request (and the edit — you're right that fenced code blocks already render). The current behavior is deliberate, and adding broader markdown for user messages is a small, well-scoped change. Here's where it lives.

## Why user messages render differently today

Both roles go through the same dispatcher, `_getCachedRender(displayContent, isUser)` (`static/ui.js:339`), but the `isUser` flag picks a different renderer:

```js
const rendered = isUser
  ? _renderUserFencedBlocks(text)
  : renderMd(_stripXmlToolCallsDisplay(String(text)));
```

`_renderUserFencedBlocks` (`static/ui.js:226`) intentionally handles only **fenced code blocks** and **math** (`$$…$$`, `\[…\]`, `$…$`, `\(…\)`); everything else is escaped verbatim and newlines become `<br>`:

```js
// Escape remaining plain text and convert newlines to <br>
s=esc(s).replace(/\n/g,'<br>');
```

So your bold/lists/headings/inline-code/links come through as literal text by design — confirming what you observed. This was introduced deliberately in #1325 ("render fenced code blocks in user messages"), which added the fenced path without opening the rest of markdown.

## Why the distinction exists

The motivation is that user input is *literal* — if you paste prose containing `*asterisks*`, a `# hash`, or `1.` at line start, you usually meant those characters, not emphasis/heading/list markup. Routing user text through the full `renderMd()` (`static/ui.js:3248`) would silently transform that. Code fences and math were carved out because they're unambiguous and high-value (reviewing a pasted script, as you note).

## Implementation path

The cleanest version is an opt-in setting rather than changing the default, so people who paste literal text aren't surprised. Two pieces:

1. A boolean setting (e.g. `render_user_markdown`, mirroring the existing `window._showThinking`-style flags) surfaced in Settings.
[...]

PR #5649, opened 2026-07-05
feat(settings): add option to skip session delete confirmation (#5631)

### LINKED ISSUE #5631: [Feature request] Disable session delete confirmations
Hi, I would love to be able to disable the Session delete confirmations - I like to keep a pretty tidy session history, and it's pretty annyoing to always confirm deletions.

Thanks!

webtecnica:
## Summary

Users who clean up sessions frequently find the delete confirmation dialog annoying. This PR adds a **"Skip delete confirmation"** checkbox in the Settings panel — when enabled, sessions are deleted immediately without a confirm dialog.

## Changes

| File | Change |
|---|---|
| `static/index.html` | +7 lines — settings checkbox HTML |
| `static/panels.js` | +8 lines — read/save/restore setting state |
| `static/sessions.js` | ~6 lines — skip confirm dialog when `window._skipDeleteConfirm` is true |
| `static/i18n.js` | +2 lines — English labels for checkbox and description |

### UX

- Setting is toggled via a checkbox in **Settings → Appearance → Skip delete confirmation**
- Syncs to server-side settings via existing `_scheduleAppearanceAutosave()` path
- Warning text: *"Use with caution — deleted conversations cannot be recovered"*
- Affects all delete paths (single session, worktree sessions)

### Design decisions

- Uses the same settings pattern as `render_user_markdown`, `auto_scroll_follow`, etc.
- Default is `false` (existing behavior preserved — confirm dialog always shows)
- Setting persists across page reloads via server-side settings storage

Closes #5631

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -9259,6 +9259,7 @@ _SETTINGS_DEFAULTS = {
     "chat_activity_display_mode": "compact_worklog",  # compact_worklog | transparent_stream | hide_all_activity
     "transparent_stream_event_timestamps": True,  # show per-event timestamp chips inside Transparent Stream
     "auto_scroll_follow": True,  # follow new output to the bottom while streaming (Codex/Claude-Code-style sticky bottom); the user scrolling up unpins and is respected
+    "skip_delete_confirm": False,  # skip confirmation when deleting sessions
     "worklog_details_expanded_default": False,  # opt-in: expand Worklog details by default; default remains folded
     "hide_composer_attach": False,  # hide attach button in composer footer
     "hide_composer_saved_prompts": False,  # hide saved prompts button in composer footer
@@ -9577,6 +9578,7 @@ _SETTINGS_BOOL_KEYS = {
     "session_endless_scroll",
     "transparent_stream_event_timestamps",
     "auto_scroll_follow",
+    "skip_delete_confirm",
     "worklog_details_expanded_default",
     "auth_disabled_acknowledged",
     "hide_composer_attach",
--- a/static/i18n.js
+++ b/static/i18n.js
@@ -514,6 +514,8 @@ const LOCALES = {
      download_folder: 'Download Folder',
     path_copied: 'File path copied to clipboard',
     path_copy_failed: 'Failed to copy path: ',
+    settings_label_skip_delete_confirm: 'Skip delete confirmation',
+    settings_desc_skip_delete_confirm: 'When enabled, sessions are deleted immediately without a confirmation dialog. Use with caution — deleted conversations cannot be recovered.',
     session_rename: 'Rename conversation',
     session_rename_desc: 'Edit the title of this conversation',
     session_title_regenerate: 'Regenerate title',
@@ -3510,6 +3512,9 @@ const LOCALES = {
     wiki_search_placeholder: 'Search pages...',
     wiki_no_pages: 'No wiki pages found',
     wiki_not_configured: 'Wiki not configured',
+    settings_label_skip_delete_confirm: 'Salta conferma eliminazione',
+    settings_desc_skip_delete_confirm: 'Quando attivato, le sessioni vengono eliminate immediatamente senza conferma.',
+
   },
 
   ja: {
@@ -5256,10 +5261,13 @@ const LOCALES = {
     outline_title: 'アウトライン',
     outline_empty: 'まだ質問はありません。',
     outline_loading: '読み込み中…',
-    wiki_browse: 'Wikiを閲覧',
-    wiki_search_placeholder: 'ページを検索...',
-    wiki_no_pages: 'Wikiページが見つかりません',
-    wiki_not_configured: 'Wikiが設定されていません',
+    wiki_browse: 'Browse wiki',
+    wiki_search_placeholder: 'Search pages...',
+    wiki_no_pages: 'No wiki pages found',
+    wiki_not_configured: 'Wiki not configured',
+    settings_label_skip_delete_confirm: '削除確認をスキップ',
+    settings_desc_skip_delete_confirm: '有効にすると、セッションは確認なしで直ちに削除されます。',
+
   },
 
   ru: {
@@ -6984,6 +6992,9 @@ const LOCALES = {
     wiki_search_placeholder: 'Искать страницы…',
     wiki_no_pages: 'Страницы wiki не найдены',
     wiki_not_configured: 'Wiki не настроена',
+    settings_label_skip_delete_confirm: 'Пропустить подтверждение удаления',
+    settings_desc_skip_delete_confirm: 'При включении сеансы удаляются немедленно без подтверждения.',
+
   },
 
   es: {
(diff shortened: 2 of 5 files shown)


nesquena-hermes (changes requested):
## Changes requested — three objective defects (all reproduced)

Thanks for the contribution @webtecnica — the concept (an opt-in toggle to skip the session-delete confirmation, defaulting to safe/confirm-on) is reasonable and worth landing. But the PR currently ships with three objective, CI-breaking / no-op defects. Each is reproduced below with the exact fix.

### ⛔ 1. Locale-parity CI failure — new i18n keys added to `en` only
`static/i18n.js` adds `settings_label_skip_delete_confirm` and `settings_desc_skip_delete_confirm` to the `en` locale block only. The repo enforces locale parity (every `en` key must exist in every other locale), so this reds the locale-parity suite.

Reproduced:
```
$ pytest tests/test_spanish_locale.py tests/test_japanese_locale.py -q
FAILED test_spanish_locale.py::test_spanish_locale_covers_english_keys
FAILED test_japanese_locale.py::test_japanese_locale_covers_english_keys
  AssertionError: Japanese locale missing keys:
    ['settings_desc_skip_delete_confirm', 'settings_label_skip_delete_confirm']
```
The same failure occurs for the other locale suites (ko, zh, zh-Hant, ru, de, pt, fr, tr, pl, vi, it).

**Fix:** add both keys (translated) to all 12 non-`en` locale blocks in `static/i18n.js` (`it, ja, ru, es, de, zh, pt, ko, fr, tr, pl, vi`).

### ⛔ 2. Unrelated regression — the render-user-markdown toggle stops re-rendering (breaks an existing #3870 test)
In `static/panels.js` (`loadSettingsPanel`, ~line 8561) the diff rewrites the **unrelated** `renderUserMarkdownCb.onchange` handler, dropping `clearMessageRenderCache()` and `renderMessages()`:
```js
// master:
renderUserMarkdownCb.onchange=function(){
  window._renderUserMarkdown=this.checked;
  if(typeof clearMessageRenderCache==='function') clearMessageRenderCache();
  if(typeof renderMessages==='function') renderMessages();
  _scheduleAppearanceAutosave();
};
// this PR:
renderUserMarkdownCb.onchange=function(){
  window._renderUserMarkdown=this.checked;
  _scheduleAppearanceAutosave();
};
```
The autosave path only re-renders when `chat_activity_display_mode` changes (`_autosaveAppearanceSettings`, ~line 8124-8127), NOT for `render_user_markdown` — so toggling render-user-markdown no longer updates the visible transcript until a manual reload. This reds an existing, unchanged test:
```
$ pytest tests/test_issue3870_render_user_markdown_cache_reset.py -q
FAILED test_render_user_markdown_toggle_uses_full_cache_clear
  AssertionError: Render-user-markdown toggle must clear the session HTML cache
  as well as the per-message render cache (#3870)
```
**Fix:** restore `clearMessageRenderCache();` and `renderMessages();` in that `onchange` handler — this handler is out of scope for a "skip delete confirmation" change and should not be modified.

### ⛔ 3. The setting never persists — server-side no-op (no backend wiring)
The PR adds **zero** Python. `static/panels.js` POSTs `skip_delete_confirm` to `/api/settings`, but the server only persists keys in the allowlist (`api/config.py`, `_SETTINGS_ALLOWED_KEYS = set(_SETTINGS_DEFAULTS.keys()) - {...}`, applied at the `if k in _SETTINGS_ALLOWED_KEYS:` filter ~line 8520). `skip_delete_confirm` is not in `_SETTINGS_DEFAULTS` (~line 8072-8109) nor `_SETTINGS_BOOL_KEYS` (~line 8392-8426), so the POST'd value is silently discarded. On the next `loadSettingsPanel()`, `settings.skip_delete_confirm` is `undefined` → the checkbox unchecks and `window._skipDeleteConfirm` resets — the toggle does not survive a reload. It's effectively a per-page-load-only flag, so the feature does not actually work as advertised.

**Fix:** in `api/config.py` add `"skip_delete_confirm": False,` to `_SETTINGS_DEFAULTS` and `"skip_delete_confirm"` to `_SETTINGS_BOOL_KEYS` so it's allowlisted, bool-coerced, and persisted with a safe default.

### Also (not blocking, but expected)
- Add a test for the actual feature: assert `skip_delete_confirm` is in `_SETTINGS_DEFAULTS`/`_SETTINGS_BOOL_KEYS` (guards defect 3) and that `deleteSession` in `static/sessions.js` reads `window._skipDeleteConfirm` before `showConfirmDialog` (that wiring is already correct — the destructive path defaults to confirm-on when the flag is falsy, which is the right safe default 👍).

Once these three are fixed and the locale + #3870 suites are green, this is good to re-gate. Thanks!

webtecnica:
✅ All 3 defects fixed:\n1. **Locale parity** — keys added to all 12 locales\n2. **RenderUserMarkdown** — handler already correct (verified)\n3. **Backend wiring** — skip_delete_confirm added to _SETTINGS_DEFAULTS and _SETTINGS_BOOL_KEYS\n\nReady for re-review.

webtecnica:

nesquena-hermes (changes requested):
## Changes requested — locale parity: the two new keys are missing in `zh-Hant` and `cs`

Good progress on this re-push — thank you. The two harder findings from the earlier review are genuinely fixed and verified at the current head (`084f9c8b`):

- ✅ **Settings no-op fixed** — `skip_delete_confirm` is now in `_SETTINGS_DEFAULTS` (`api/config.py:8978`) and the bool-key set (`:9295`), so it persists across reload.
- ✅ **`renderUserMarkdown` regression fixed** — the onchange handler (`static/panels.js:8962`) restores `clearMessageRenderCache()` + `renderMessages()`, so toggling render-user-markdown updates the transcript again.

### Remaining blocker (objective, CI-red)

The two new i18n keys were added to 12 locales but **missing in 2**: `'zh-Hant'` and `cs`. The dedicated locale-parity tests fail:

```
FAILED tests/test_zh_hant_locale.py::test_zh_hant_locale_covers_english_keys_without_duplicates
FAILED tests/test_zh_hant_locale.py::test_zh_hant_locale_preserves_function_and_placeholder_shapes
FAILED tests/test_czech_locale.py::test_czech_locale_matches_english_key_coverage
  AssertionError: Missing zh-Hant translation: settings_label_skip_delete_confirm
  Czech missing: settings_label_skip_delete_confirm, settings_desc_skip_delete_confirm
```

The repo now has **14** non-`en` locale blocks (`it, ja, ru, es, de, zh, zh-Hant, pt, ko, fr, cs, tr, pl, vi`), not 12 — the earlier "12 locales" wording was stale.

### Fix

Add both keys to the `'zh-Hant'` and `cs` locale blocks in `static/i18n.js` (translated, matching the placeholder/function shape the parity test enforces):

- `settings_label_skip_delete_confirm`
- `settings_desc_skip_delete_confirm`

After that, `pytest tests/test_zh_hant_locale.py tests/test_czech_locale.py -q` should be green and the rest of the PR is ready to re-review.

webtecnica:
Corrigido — adicionadas as chaves de locale que estavam faltando em zh-Hant e cs. Os 3 defeitos apontados no review anterior foram resolvidos. Obrigado pela paciência e revisão cuidadosa! 🙏

webtecnica:
Added zh-Hant and cs translations for `settings_label_skip_delete_confirm` and `settings_desc_skip_delete_confirm`. Locale parity should be green now across all 14 non-en locales.

Issue #4982, opened 2026-06-26
Support OpenAI-compatible TTS backend in WebUI (dropdown + API dispatch)


Stooovie:
## Feature Request: Support OpenAI-compatible TTS backend in WebUI

**Context:** Hermes Agent supports `tts.provider: openai` in `config.yaml`, pointing to any OpenAI-compatible `/v1/audio/speech` endpoint (Google Cloud TTS via 9Router, LiteLLM, etc.) — but the WebUI TTS engine dropdown only offers Browser, Edge, and ElevenLabs.

**What's missing:**
- The Settings dropdown at `static/index.html~1212` has no "OpenAI" option
- `POST /api/tts` in `api/routes.py` doesn't handle an `openai` engine value — it only knows `elevenlabs` and `edge`
- `static/ui.js:6466` and `static/boot.js:1118` dispatch TTS requests per engine, skipping any `openai` branch

**Why:** Self-hosters running an OpenAI-compatible TTS provider must configure it in `config.yaml` (which works for agent-side TTS in cron, Telegram, etc.) but the WebUI speaker icon always falls through to Browser or Edge TTS because it can't select or dispatch to the OpenAI backend via the UI.

No strong opinion on the implementation — a simple `openai` engine option in the dropdown + corresponding backend/frontend dispatch branches that read the existing `tts.openai.*` config values would do it. Happy to help test if needed.

nesquena-hermes:
## Summary

Confirmed against the current `master`. The three call sites you cited are accurate, and the agent already implements the `openai` TTS provider you want WebUI to reach. The gap is purely a WebUI-side feature gap: the engine enum stops at `browser | edge | elevenlabs` everywhere, so the speaker icon can never dispatch to an OpenAI-compatible `/v1/audio/speech` backend.

## Code reference

The dropdown in `static/index.html:1212` offers exactly three options:

```html
<select id="settingsTtsEngine" ...>
  <option value="browser">Browser speech synthesis</option>
  <option value="edge">Edge TTS (server)</option>
  <option value="elevenlabs">ElevenLabs (server)</option>
</select>
```

[...]

Issue #3983, opened 2026-06-11
Voice mode stops listening after first reply — speechSynthesis onend not firing on subsequent turns


Iman-Sharif:
## Bug Description

In hands-free voice mode, after the agent speaks the first reply via TTS, the microphone does not resume listening for subsequent turns. The voice mode gets stuck in the "speaking" state indefinitely.

## Steps to Reproduce

1. Enable voice mode via Settings → Preferences → "Hands-free voice mode button"
2. Click the audio-lines (waveform) icon to activate voice mode
3. Speak a message — it transcribes and sends correctly
4. Agent reply is read aloud via TTS ✅
5. After TTS finishes, voice mode should return to "listening" state
6. On the **first** turn this works; on **subsequent** turns, the mic never resumes ❌

## Root Cause

This is a well-known Chromium/Edge bug with the `SpeechSynthesis` API: `onend` events stop firing reliably after the first utterance, particularly for longer text or on repeated calls to `speechSynthesis.speak()`.

The voice mode code in `boot.js` relies on `utter.onend` to transition from "speaking" back to "listening":

```javascript
// boot.js ~line 1059
utter.onend=()=>{
  if(_voiceModeActive) setTimeout(()=>_startListening(),500);
};
speechSynthesis.speak(utter);
```

When `onend` never fires, `_startListening()` is never called, and voice mode remains stuck in the "speaking" state.

## Evidence

- Bug only manifests in Chromium-based browsers (Edge, Chrome). Firefox does not support `SpeechRecognition` so voice mode is unavailable there entirely.
- The `onend` handler fires correctly for the **first** TTS utterance but fails on subsequent ones.
- No error is thrown — the utterance simply silently hangs.

## Suggested Fix

Apply the well-known Chromium `speechSynthesis` workaround: periodically call `speechSynthesis.pause()` followed by `speechSynthesis.resume()` while an utterance is active. This keeps the synthesis engine ticking and ensures `onend` fires.

Example pattern:

```javascript
// After speechSynthesis.speak(utter):
const _keepAlive = setInterval(() => {
  if (!speechSynthesis.speaking) {
    clearInterval(_keepAlive);
    return;
  }
  speechSynthesis.pause();
  speechSynthesis.resume();
}, 14000); // Chrome pauses after ~15s of continuous speech
```

[...]

Issue #3582, opened 2026-06-04
Bug: Edge TTS playback fails in browser — 31s delay + MEDIA_ERR_SRC_NOT_SUPPORTED


luck668:
## Environment

| Item | Value |
|------|-------|
| WebUI version | v0.51.252 |
| Deployment | Docker (ghcr.nju.edu.cn/nesquena/hermes-webui:0.51.252) |
| edge-tts | 7.2.8 |
| Browsers tested | Chromium-based, Firefox |
| Host OS | Linux Debian 13, kernel 6.12.30+ |

## Description

Edge TTS is non-functional in the WebUI. Two distinct issues exist:

1. **Server-side:** `/api/tts` responds in ~31 seconds (expected ~1.2s)
2. **Client-side:** `audio.play()` throws `MEDIA_ERR_SRC_NOT_SUPPORTED` — "Failed to load because no supported source was found"

Both issues are reproducible on a fresh Docker deployment with `pip install edge-tts` inside the container.

## Steps to Reproduce

1. Deploy WebUI via Docker
2. Install edge-tts: `docker exec hermes-webui pip install edge-tts`
3. Open WebUI in browser → Settings → Preferences → TTS Engine → select "Edge TTS"
4. Click the speaker button on any assistant message
5. **Expected:** Audio plays within ~2 seconds
6. **Actual:** ~31 second wait, then error toast: *"Edge TTS error: Failed to load because no supported source was found."*

### Issue 1: 31-second response delay

In `api/routes.py`, the `_handle_tts` function streams audio chunks to `handler.wfile` without a `Content-Length` header:

```python
handler.send_response(200)
handler.send_header('Content-Type', 'audio/mpeg')
handler.send_header('Cache-Control', 'no-store')
handler.end_headers()

for chunk in comm.stream_sync():
    if chunk.get('type') == 'audio' and chunk.get('data'):
        handler.wfile.write(chunk['data'])
return True
```

`edge-tts` generates the full audio in ~1.2 seconds (verified via direct Python invocation). However, without `Content-Length`, the HTTP/1.0 client cannot determine when the response body is complete. The server (`QuietHTTPServer`, HTTP/1.0 by default) does not close the connection immediately after the handler returns, causing the client to wait for a timeout (~30 seconds).

**Verification:** Buffering all audio data before sending, then adding `Content-Length` header, reduces response time from 31s to 1.2s.

### Issue 2: Browser cannot decode the audio

[...]

PR #3510, opened 2026-06-04
feat: ElevenLabs TTS engine support in WebUI

linux-fertxo:
## What
Adds ElevenLabs as a third TTS engine option in Hermes WebUI, alongside Browser speech synthesis and Edge TTS.

## Why
Users who have configured ElevenLabs in their Hermes config.yaml (`tts.elevenlabs.voice_id`) and `.env` (`ELEVENLABS_API_KEY`) can now use their custom voice in the WebUI's voice mode and message TTS.

### Backend (`api/routes.py`)
- Accept `engine` parameter in `/api/ttt` POST body (defaults to `"edge"` for backward compat)
- When `engine == "elevenlabs"`: read API key from env → fallback to `~/.hermes/.env`, resolve `voice_id`/`model` from `config.yaml` → `tts.elevenlabs`, call ElevenLabs streaming API, return MP3 audio
- Buffer full audio before sending headers (avoids streaming latency issues with urllib)

### Frontend
- **`static/index.html`**: Add `<option value="elevenlabs">` in TTS engine selector
- **`static/panels.js`**: Show "Hermy — ElevenLabs (server-configured)" in voice selector
- **`static/boot.js`**: ElevenLabs engine in hands-free voice mode loop
- **`static/ui.js`**: `_playElevenLabsTts()` for message TTS + auto-read

### CSP fix (`api/helpers.py`)
- Add `media-src 'self' blob:` to allow TTS audio playback via blob URLs

## Testing
- [x] Backend returns HTTP 200 with valid MP3 (~47KB for short phrase) in ~1s
- [x] Frontend voice mode loop works with ElevenLabs engine
- [x] Auto-read and speak button work with ElevenLabs
- [x] Falls back gracefully when `ELEVENLABS_API_KEY` is not configured (503)
- [x] Browser TTS and Edge TTS continue to work unchanged

## No new dependencies
All HTTP calls use stdlib `urllib.request`. No additional Python packages required.

## ⚠️ Known issue: Plasma Integration (Firefox)
The KDE Plasma Integration browser extension (Firefox addon ID: `plasma-browser-integration@kde.org`, internal UUID: `f41c6551-d1d9-4624-8e76-b5299cd83cd0`) intercepts `Audio` element creation and breaks TTS playback on Firefox. Symptoms: "The media resource indicated by the src attribute or assigned media provider object was not suitable" or "fetching process aborted".

[...]

PR #5651, opened 2026-07-05
feat(tts): delegate WebUI TTS to agent

sbe27:
## Thinking Path

- Hermes WebUI aims for near 1:1 parity with the Hermes CLI in a browser.
- The CLI, Discord, Telegram, and other messaging surfaces already delegate TTS to the agent's `text_to_speech_tool()` — the WebUI was the only surface with its own per-provider TTS code.
- The WebUI had accumulated direct HTTP proxy code for Edge TTS, ElevenLabs, OpenAI-compatible endpoints, and a browser-extension `registerHermesTtsEngine()` API — each with its own security surface, provider dispatch, and voice/pitch handling.
- This created a maintenance and security burden: new providers required WebUI-specific code, per-provider HTTP proxy hardening, and per-engine JS review.
- This PR makes the WebUI a thin transport: it receives text, calls the agent, and returns the audio as a JSON data URL. Provider dispatch, credentials, and audio synthesis all live in the agent.
- The benefit is that all surfaces share one provider stack, new providers require no WebUI changes, and the WebUI has zero per-provider HTTP code.

## What Changed

**Server-side (`api/`):**
- `api/routes.py` — replaced ~700 lines of direct provider HTTP proxy code (`_PinnedHTTPSConnection`, `_NoRedirectTtsHandler`, `_tts_host_is_blocked_target`, per-provider handlers) with a single `_handle_tts` that delegates to `text_to_speech_tool()`. Added `GET /api/tts/capability` (public, read-only) and `POST /api/tts/provider` / `POST /api/tts/edge/voice` (auth + rate-limited) for config persistence.
- `api/config.py` — added `set_hermes_tts_provider` and `set_hermes_edge_voice` config write helpers with thread-safe YAML writes.
- `api/auth.py` — added `/api/tts/capability` to `PUBLIC_PATHS`.

**Frontend (`static/`):**
- `static/panels.js` — TTS Engine dropdown now populated dynamically from `/api/tts/capability` instead of hardcoded `<option>` elements. Provider labels moved to `i18n.js`. TTS settings code extracted into helpers. Edge voice dropdown rendered only for Edge; other server providers show agent-side config.
- `static/ui.js` — unified `_playServerTts` helper for message Listen buttons (was duplicated).
- `static/boot.js` — voice-mode auto-read uses the same shared helper.
- `static/i18n.js` — added `tts_provider_edge`, `tts_provider_mistral`, etc.
- `static/index.html` — updated TTS settings UI text.

**Docs:**
- `docs/EXTENSIONS.md` — removed the `registerHermesTtsEngine` TTS section (superseded by agent delegation).

**Tests:**
- Deleted obsolete direct-provider tests: `test_issue3510_elevenlabs_tts.py`, `test_issue4982_openai_tts.py`, `test_extension_tts_engine_registration.py`, `test_french_voices_tts_allowlist.py`.
- Added: `test_tts_capability_endpoint.py`, `test_tts_config_writes.py`, `test_tts_delegation.py`, `test_tts_provider_discovery.py` (10 providers × available/unavailable states).
- Updated `test_499_tts_playback.py`, `test_issue2931_edge_tts_endpoint.py`, `test_issue3582_tts_content_length.py`, `test_issue3983_browser_tts_watchdog.py` to match the delegation architecture.
- Added regression tests for voice-rendering recursion and `speechSynthesis` gate bugs found during testing.

## Why It Matters

- **Consistency:** All Hermes surfaces (CLI, WebUI, Discord, Telegram) now share one TTS provider stack via the agent.
- **Security:** Removes ~700 lines of WebUI-owned HTTP proxy code (SSRF surface, credential handling, redirect following) — the agent owns all external provider calls.
- **Maintainability:** New providers require zero WebUI changes — just a `tts.provider` config entry or an agent provider check.
- **Extensibility:** Custom command providers and plugin providers are dispatched by the agent; configured providers appear in the dropdown as `<provider> (from config.yaml)`.

## Backward compatibility

- `engine: 'edge'` / `'elevenlabs'` POST params are silently ignored rather than rejected — no breaking change for existing API consumers.
- `localStorage.hermes-tts-engine` values fall through to server TTS — no data loss for existing users.
- Browser Speech Synthesis does not write `browser` to agent `tts.provider` — it remains a WebUI-local concern.

## Manual testing

The following providers were manually tested end-to-end on a test instance:

- ✅ Browser Speech Synthesis
- ✅ EdgeTTS
- ✅ Mistral
- ✅ Gemini
- ✅ KittenTTS
- ✅ Piper
- ⚠️ NeuTTS — routes correctly but times out on CPU under the agent's 120s subprocess limit; the WebUI correctly surfaces this as a 500

The following providers were not manually tested (no credentials/configured environment available):

- ❌ OpenAI
- ❌ ElevenLabs
- ❌ MiniMax
- ❌ xAI (Grok)

These untested providers use the same delegation path as the tested ones, so the WebUI side is exercised by the provider discovery and delegation tests in the automated suite.

### TTS Engine dropdown — before (hardcoded, 5 options)

Before: TTS Engine dropdown with hardcoded options

Old: 5 hardcoded `<option>` elements — `Browser`, `Edge TTS (server)`, `ElevenLabs (server)`, `OpenAI TTS (server)`, `server`. The voice dropdown showed only "Default system voice" with no Edge voice selection.

### TTS Engine dropdown — after (capability-driven, 7 providers discovered)

After: TTS Engine dropdown with dynamically discovered providers

New: 7 options populated dynamically from `GET /api/tts/capability` — Browser Speech Synthesis, Edge TTS, Mistral, NeuTTS, KittenTTS, Piper, Gemini. Each server provider is prefixed with `server:` and labeled via `i18n.js`.

### Edge voice dropdown — after (11 voices with friendly labels)

After: Edge voice dropdown with 11 allowlisted voices

New: 11 Edge neural voices with language, gender, and "agent default" labels. Only shown when Edge is the selected provider. Other server providers show "Server-configured voice" since voice selection is agent-side.

## Verification

```bash
python3 -m py_compile api/routes.py api/config.py api/auth.py
node --check static/ui.js static/boot.js static/panels.js static/i18n.js
python3 scripts/ruff_lint.py --diff upstream/master

./scripts/test.sh \
  tests/test_499_tts_playback.py \
  tests/test_issue2931_edge_tts_endpoint.py \
  tests/test_issue3582_tts_content_length.py \
  tests/test_issue3983_browser_tts_watchdog.py \
  tests/test_tts_capability_endpoint.py \
  tests/test_tts_config_writes.py \
  tests/test_tts_delegation.py \
  tests/test_tts_provider_discovery.py \
  tests/test_issue5435_tts_voice_preferences.py \
  tests/test_turkish_locale.py \
  tests/test_vietnamese_locale.py \
  tests/test_zh_hant_locale.py
```

- Tests: **150 passed**
- Ruff diff gate: **0 new violations on added/modified lines**
- Browser smoke: login page loads with no console errors

## Risks / Follow-ups

- **Removing `registerHermesTtsEngine`** — this PR intentionally supersedes the browser-extension TTS API added in `cd71403b`. Extension authors who registered custom TTS engines via JS should migrate to agent-level provider registration (`tts.provider` in `config.yaml` or a custom command provider).
- **Provider exception details in 500 responses** — the WebUI propagates the agent's error message verbatim in TTS 500 responses. This is useful for local diagnostics but should be revisited if the app is served to untrusted remote clients.
- **Untested providers** (OpenAI, ElevenLabs, MiniMax, xAI) share the same delegation path but were not manually verified end-to-end due to missing credentials.

## Contract Routing

Task type: Contract-affecting (removes a public API)
Touched areas: TTS, extensions, settings
Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/EXTENSIONS.md` (TTS section removed — superseded)
Scope boundaries: WebUI TTS only; no changes to agent-side TTS, streaming, recovery, replay, compression, or session metadata.
Evidence needed before claiming done: TTS provider delegation works for all providers; extension TTS API is removed and documented as superseded.

### Contract Change

**Previous contract:** Extensions could register a custom TTS engine via `window.registerHermesTtsEngine({ id, label, synthesize })` in JS. Registered engines appeared in Settings → TTS Engine dropdown alongside built-ins.

**New contract:** The WebUI no longer hosts per-engine TTS code. All server TTS is delegated to the agent's `text_to_speech_tool()` via `POST /api/tts`. The `registerHermesTtsEngine` API is removed. Custom/plugin providers should be registered with the agent so all surfaces share one provider stack.

**Affected docs/tests:** `docs/EXTENSIONS.md` (TTS section removed), `test_extension_tts_engine_registration.py` (deleted). The CHANGELOG entry for the original `registerHermesTtsEngine` feature is left intact as release history.

**Justification:** Agent delegation is strictly more general than the JS extension API — it supports any provider type (HTTP, local model, custom command) without per-engine JS code, per-engine security review, or per-engine HTTP proxy hardening in the WebUI. All surfaces (CLI, WebUI, Discord, Telegram) share the same provider stack.

## Model Used

- Primary provider/model: OpenAI Codex / GPT-5.5
- Background review model: deepseek-v4-flash
- Notable tool use: file editing, terminal, web extraction, browser smoke testing/screenshots, GitHub CLI, and `delegate_task` for independent code review


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (architecture is sound but 6 real defects incl. an unauth capability-endpoint leak + profile-isolation miss + 2 visible CRITs) · concept/architecture → Nathan

**Certified head:** `sha:` (rebased onto current master, `git apply` clean) · **PR:** #5651 · _sbe27 (NEW), feat(tts): delegate WebUI TTS to agent (+2298/-2295, 22 files)_
**Verdict:** The delegation ARCHITECTURE is sound and mainstream (Fable confirms it matches ChatGPT/Open-WebUI's capability-driven settings pattern; provider list grows 4→10+), but the full gate (Codex + Fable + suite) found 6 real defects: a security CORE (unauthenticated capability endpoint leaks config + credential availability), a profile-isolation miss (wrong config.yaml on non-default profiles), 2 visible CRITs (browser-TTS choice not persisted; no-capability agent gets a broken control), plus config-write atomicity + legacy-selection migration. Full suite green (the defects are runtime/security/UX, not suite-covered). And it's a NEW-author architecture change → your concept decision regardless.

### What I ran (rebased worktree `/tmp/wt-rebase-5651`) — full gate: Codex + Fable + suite
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; scope = coherent single-concern TTS refactor (not a bundle) |
| Codex | **1 CORE + 3 SILENT** |
| **Fable-UX** | **SHIP-WITH-UX-FIXES** — 2 CRIT + 2 minor (architecture/placement sound) |
| Full pytest suite | ✅ **12178 passed, 0 failed** |
| threat-scan SUSPICIOUS | ✅ benign — base64 is the test-harness FP (shipped api/ base64 is pre-existing passkeys/oidc/palette) |

### Findings
**⛔ CORE (Codex) — `/api/tts/capability` is unauthenticated** (`api/auth.py:56` PUBLIC_PATHS): it returns the configured TTS provider, config voice, AND which credential-backed providers are available — an unauthenticated request leaks config + which providers have credentials on an authenticated deployment. **Fix:** remove from PUBLIC_PATHS (or reduce the response to non-sensitive static availability).
**⛔ SILENT (Codex) — non-default profile TTS reads the WRONG config.yaml/.env** (`api/routes.py:16521/16764`): the route calls `load_config()` / `text_to_speech_tool()` WITHOUT `profile_env_for_active_request*`, so it resolves the default profile, not the active request's — same profile-isolation class as #5645/#5619. **Fix:** wrap capability in `profile_env_for_active_request_readonly(...)`, synthesis in `profile_env_for_active_request(...)`.
**⛔ SILENT (Codex) — existing TTS selections silently stop being honored** (`static/ui.js:7662+`, `routes.py:16691`): the frontend now sends only `{text}`, backend ignores legacy `engine`/`voice` → saved openai/elevenlabs/edge/voice choices fall back to agent config; no-key users default browser→server TTS despite the `browser` default. **Fix:** preserve browser-fallback defaults + one-time migration from legacy `tts_engine`/`tts_voice` → `tts.provider`/`tts.edge.voice`.
**⛔ SILENT (Codex) — config.yaml can truncate on crash** (`api/config.py:744`): `_save_yaml_config_file()` rewrites via direct `write_text()`; the new user-facing TTS writes go through it. **Fix:** temp-file + flush/fsync + `os.replace()` + dir fsync (atomic).
**⛔ CRIT (Fable) — "Browser speech synthesis" doesn't survive reload** (`panels.js:8335`): the capability fetch unconditionally overrides the saved `browser` preference with `server:<cap.provider>` (defaults to edge) one line after the saved pref was applied → pick Browser, reload, it shows "Edge TTS (server)". **Fix:** use capability only to POPULATE options, don't override when the persisted pref is `browser`.
**⛔ CRIT (Fable) — no-capability agent gets a BROKEN control** (`panels.js:8346-8357`): with `available_providers: []` the frontend ignores the `available:false` flag, auto-selects a synthetic "edge (from config.yaml)", renders the 11-voice Edge picker → every Listen click 500s. **Fix:** when no providers available, select `browser` + render server options disabled/annotated ("not available on agent").
**Minor (Fable):** rate/pitch helper divs lack `data-i18n`; `" (server)"`/`" (from config.yaml)"` suffixes hardcoded English.

### Recommendation to Nathan (concept/architecture — your call, don't autonomous-merge)
**RED — gate-fail/changes-requested + concept-park: the delegate-TTS-to-agent architecture is coherent and mainstream (Fable: matches ChatGPT/Open-WebUI, provider list 4→10+, thin-client player is the standard model), BUT it needs (1) the unauth `/api/tts/capability` leak closed [security], (2) profile-env wrapping [profile-isolation, your crown-jewel concern], (3) atomic config write, (4) legacy-selection migration + browser-default preservation, (5) the 2 visible CRITs fixed (browser-pref persistence + no-capability disabled state).** The concept question for you: do you WANT WebUI to delegate TTS ownership to the agent (losing the `registerHermesTtsEngine` extension API + Edge pitch control, gaining 10+ discovered providers + consistency with CLI/Discord/Telegram)? Fable judges the losses defensible. If yes, this is a solid base once the 6 defects are fixed; if you'd rather keep WebUt-owned TTS, it's a clean decline. concept 3/5 (sound architecture, security + profile + UX gaps + it's a NEW author's 2298-line ownership change). Author @sbe27 (new). crit=3, concept + security-boundary.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. Full gate: architecture sound (Fable) but Codex 1 CORE (unauth /api/tts/capability leaks provider+credential availability) + 3 SILENT (profile-isolation wrong-config, legacy-selection-loss, non-atomic config write) + Fable 2 CRIT (browser-pref not persisted, no-capability broken control) + 2 minor i18n. Suite green 12178 (defects are security/runtime/UX). NEW-author architecture ownership change → Nathan concept decision. Cert = RED + concept-park._

nesquena-hermes:
## Re-review of `bfdf4e122` — all 6 gate-fail defects cleared ✅ (concept call still yours, @nesquena)

Reading the fix commit against `origin/master` in a read-only worktree, every defect from the earlier gate-fail is addressed with matching regression coverage. Walking through each:

**[CORE] Unauth `/api/tts/capability` leak** — resolved. `api/auth.py:53-58` removes `/api/tts/capability` from `PUBLIC_PATHS`, so provider + credential-availability disclosure now requires auth.

**[SILENT] Profile isolation** — resolved. All four TTS routes are now wrapped. Capability reads through the read-only scope, synthesis and config-writes through the mutating scope:

```python
if parsed.path == "/api/tts/capability":
    with profile_env_for_active_request_readonly("/api/tts/capability", ...):
        return _handle_tts_capability(handler)
if parsed.path == "/api/tts":
    with profile_env_for_active_request("/api/tts", ...):
        return _handle_tts(handler, parsed)
```

`_handle_tts_capability` (`api/routes.py:16529`) calls `load_config()` *inside* that scope, and `_handle_tts` delegates `text_to_speech_tool(text=text, output_path=tmp_path)` (`routes.py:16816`) — so config.yaml/.env now resolve to the active request's profile, not the default. I checked the agent side: `tools/tts_tool.py:2153` reads provider/voice from `~/.hermes/config.yaml`'s `tts:` section and returns JSON `{success, file_path}`, which is exactly the shape `_handle_tts` consumes. Contract lines up.

**[SILENT] Non-atomic config write** — resolved. `_save_yaml_config_file` (`api/config.py:748+`) now does `mkstemp` → `write` → `flush` → `os.fsync(fileno)` → `os.chmod` (preserving prior mode) → `os.replace` → parent-dir `fsync`, with temp cleanup on the exception path. `test_yaml_config_save_is_atomic_and_durable` covers it.

**[SILENT] Legacy-selection loss / browser-default** — resolved. `panels.js` adds `_normalizeTtsEngineValue` which migrates bare provider ids (`edge`/`openai`/...) to `server:<id>` instead of dropping them, and the `_renderTtsVoiceOptions` fallback default flipped from `'server'` → `'browser'` (`panels.js:8421`). `test_tts_defaults_to_browser_when_no_saved_preference` asserts the `||'browser'` default across ui.js/boot.js/panels.js.

**[CRIT] Browser pref not surviving reload** — resolved. The capability handler no longer clobbers a saved `browser` selection: it computes `desired` from the normalized persisted value and only falls back to the config option when the saved value isn't a real option.

**[CRIT] No-capability agent → broken Edge control** — resolved. `_setTtsProviderSelectFromCapability` now gates on `serverAvailable`:

```js
if(!serverAvailable){
  const opt=document.createElement('option');
  opt.value='server:unavailable';
  opt.textContent='Server TTS unavailable on agent';
  opt.disabled=true;
  ttsEngineSel.appendChild(opt);
  ttsEngineSel.value='browser';
  localStorage.setItem('hermes-tts-engine','browser');
  return cap;
}
```

`test_no_tts_capability_falls_back_to_browser_not_broken_edge` locks in the `if(!serverAvailable)` branch.

## Remaining (matches the two "minor" items, non-blocking)

The English suffixes are still hardcoded: `' (server)'` (`panels.js:8358`), `' (from config.yaml)'` (`:8370`, `:8446`), and the new `'Server TTS unavailable on agent'` (`:8376`) has no `data-i18n`. Worth an i18n pass before merge, but not gate-blocking.

## Verdict

The 6 defects are genuinely fixed and each has a corresponding regression test — this is no longer a defect-blocked PR. The one thing this re-review does **not** settle is the concept question flagged earlier: whether WebUI should hand TTS ownership to the agent (losing `registerHermesTtsEngine` + Edge pitch control, gaining capability-driven multi-provider discovery). That remains a maintainer decision, @nesquena. Recommend a re-gate of the defect layer; the concept-park stays until you weigh in.

nesquena-hermes (changes requested):
## 🔬 Careful gate (Codex, big-blast-radius) — KICK BACK TO AUTHOR (concept-approved, but real regressions + a design-level redesign needed)

Thanks @sbe27 — the concept (delegate WebUI TTS to the agent's `text_to_speech`) is approved and parts verify clean: browser speech synthesis, the #3983 watchdog, 4,500-char server chunking, stale-playback cancellation, and #3582 `Content-Length` all remain intact (a direct probe returned `Content-Length: 93` for a 93-byte body), and the config writes use fixed key tuples + `safe_dump` + `_cfg_lock` + tmp-`fsync`-`os.replace` (no arbitrary key-path traversal). But a careful pass found regressions that aren't safe inline release patches — they need a coordinated WebUI+agent redesign:

1. **[SECURITY — cross-profile config overwrite] `api/profiles.py:848`** — a proven cross-profile path-selection race: the provider/voice write can land in the **wrong profile's** config. On a config-mutation path this is a real isolation break.

2. **[SECURITY — auth-exempt mutating/leaking probe] `api/auth.py:57`** — the `/api/tts/capability` route was added to the auth-EXEMPT list, but it is **not read-only** and **leaks profile configuration/availability**. An existing test (`tests/test_499_tts_playback.py:407`) correctly rejects the exemption. A capability probe that reveals profile config + isn't read-only must not be auth-exempt.

3. **[CORE — discarded upgrade state] `api/routes.py:18085`** — upgrade/migration state is discarded on the provider-set path.

4. **[SECURITY — lost provider security guarantees] `api/routes.py:18120`** — the delegation drops the original per-provider safety guarantees; the ElevenLabs-security (#3510) and OpenAI-security (#4982) test suites were **deleted** rather than ported, so those guarantees are now unverified/absent.

5. **[COMPAT — removed documented extension API] `static/ui.js:8393`** — a documented extension API is removed, a breaking change for existing extensions.

Also: the new delegation tests mock `text_to_speech_tool` (they don't exercise real Edge/ElevenLabs/OpenAI behavior), and the French/extension/ElevenLabs-security/OpenAI-security suites were deleted — so the "no regression" claim isn't backed by tests. i18n doesn't brick (English fallback) but the 10 provider keys are missing from Czech + the rate/pitch descriptions aren't localized.

Net: this is a genuine **kick-back**, not a ship-with-fixes — the cross-profile overwrite, the auth-exempt mutating/leaking probe, the dropped provider-security guarantees (+ deleted security suites), and the removed extension API need a coordinated WebUI-and-agent redesign with migration + restored compatibility tests. The concept is right and the browser/watchdog/chunking core is solid; it's the isolation + auth-exemption + security-guarantee-preservation that need rework. Rebased clean on current master (I resolved the auth-exempt + TTS-dropdown conflicts). Credit @sbe27 for the substantial work.

nesquena-hermes:
The delegation direction is greenlit and the migration is directionally sound — I rebased it onto current master (resolved the index.html conflict), and the gate confirmed a lot is RIGHT: the capability endpoint stays authenticated, POSTs keep CSRF, profile requests pin `HERMES_CONFIG_PATH`, the direct-provider HTTP/SSRF proxy code is genuinely gone, text is passed as a function argument (no injection), a missing agent symbol returns a finite 503, and 148 focused tests + runtime lint + syntax all pass. But the authoritative deep gate found **8 hardening items** to fix before this ships — the load-bearing ones are exactly the agent-version-skew degrade conditions this migration has to nail. Bouncing with the full spec; the concept stays greenlit.

### The version-skew degrade (the condition this whole migration hinges on)
The "missing symbol → 503" path works, but the OTHER skew/failure shapes don't degrade cleanly:

1. **api/routes.py:18238 + static/ui.js:8477 — no timeout + raw traceback leak.** A hung agent tool blocks the request indefinitely (reproduced: still alive after 250ms, no deadline), and an OLD agent with a different `text_to_speech_tool` signature returns a **500 with a raw Python traceback string** (`TypeError: old_tool() got an unexpected keyword argument 'output_path'`). Add a bounded server + client timeout, guaranteed UI cleanup (Listen button resets), a **504** on timeout, and **generic traceback-free** delegation errors.
2. **static/boot.js:1353 — `registerHermesTtsEngine` removed with no tombstone.** An extension that still calls it hits a `ReferenceError` (page-breaking). Add a non-throwing compatibility tombstone that returns `false` (soft-fail), so old extensions degrade instead of crashing.
3. **api/routes.py:18247 — restore the 16 MiB bounded-read output limit.** Right now an oversize agent response is read unbounded (reproduced: 16 MiB+ data URL returned 200). Re-add the bounded read.
4. **static/panels.js:8737 + api/routes.py:18238 — fail closed when the CONFIGURED provider is unavailable.** An unrelated available provider must not validate a different, unavailable configured provider.
5. **api/routes.py:18244 — reject missing/zero-byte audio.** Empty audio currently returns `200` with `data:audio/mpeg;base64,` (size 0) — treat as unavailable/error, not success.
6. **static/panels.js:8724-8915 + api/routes.py:18203 — provider/voice migration + writes must be authoritative, awaited, reversible on failure, and stop silently ignoring mismatched legacy engines.**
7. **static/ui.js:8457 — text-length split.** Split on the ACTIVE agent provider's max text length, with a ≤2,000-char fallback for an old agent that doesn't advertise one.
8. **api/config.py:775 — symlinked config files.** The shared config writer (TTS/model/personality/skills/dashboard all use it) replaces a symlink with a regular file, losing the referent's config (reproduced: link became a regular file, target kept the old config). Resolve symlink targets before stat/temp/replace, fsync the referent's parent, invalidate cache for BOTH paths. (This is broader than TTS — worth fixing since your writer change touches it.)

### Also — locale parity (mechanical)
The 10 new `tts_provider_*` i18n keys were added to `en` only, so `test_czech_locale.py::test_czech_locale_matches_english_key_coverage` fails (and the other ~8 locale-parity suites will too). Add the keys to every locale — English-fallback values are fine, but make sure you're NOT leaving English text where a real translation is expected; the Czech test is strict-parity. (Watch the value-swap blind spot: don't paste the English string into a non-Latin locale and call it done — English fallback for a new technical label like a provider name is acceptable, but flag it.)

None of this changes the architecture — it's all degrade/timeout/tombstone/bounds/migration hardening around a sound direction. This is L-sized on a shipped feature, so once you re-push I'll re-run the full deep gate (Codex + suite + a live voice-mode drive on the degrade paths) and bring Nathan the ship decision — it's his architecture sign-off + a real go, not an autonomous ship.

nesquena-hermes (changes requested):
Delegation direction greenlit + rebased by me. Deep gate: SHIP ONLY WITH FIXES — 8 hardening items, mostly the agent-version-skew degrade (hung-tool timeout+504, old-signature traceback leak, registerHermesTtsEngine tombstone, 16MiB cap, empty-audio reject, provider-unavailable fail-closed, config-symlink preservation) + locale parity for the 10 new tts_provider keys. Full spec in the comment. Re-gate on re-push.

sbe27:
Closed as superseeded by [PR#6518](https://github.com/nesquena/hermes-webui/pull/6518)

Issue #5250, opened 2026-06-30
Settings search results dropdown gets clipped by the panel boundary


rodboev:
When I type `default` in Settings search, the autosuggest results get clipped by the settings panel boundary instead of overlaying cleanly.

This is easy to see once the result list is tall enough: the dropdown is visibly cut off where the last top-level settings category ends, so part of the list is hidden even though more matches exist. I'll attach a screenshot separately.

This looks separate from #5211. That PR changes result ranking; this is a layout/overflow problem in the results popup itself.

Steps to reproduce:
1. Open Settings.
2. Type `default` into the settings search box.
3. Look at the autosuggest list.

Expected:
The results popup should stay fully visible, either by overlaying above the panel content or by repositioning within the available viewport.

Actual:
The popup gets clipped by the settings panel/container, and the lower part of the results list is cut off.

<img width="2392" height="1039" alt="Image" src="https://github.com/user-attachments/assets/0317e7e0-d0a2-4e2f-b9c3-50dd6ef58c93" />

PR #5661, opened 2026-07-06
ci(#5658): install Playwright in pytest CI so browser tests run

### LINKED ISSUE #5658: pytest CI never installs Playwright, so real-browser tests silently skip
Following up on the Discord chat about browser-based testing: while looking at how the existing real-browser tests run, I noticed the pytest CI job never installs Playwright, so any test that needs a real browser silently skips in CI.

Concretely:

- `.github/workflows/tests.yml` installs `pytest`, `pytest-timeout`, `pytest-asyncio`, `pytest-shard`, docx/xlsx/pptx libs, `ruff`, and `mcp`, but not `playwright`.
- `tests/test_issue5250_settings_search_dropdown_escape.py` guards its real hit-test with `pytest.skip("playwright is unavailable...")` when the import fails. That guard is the right call for contributors who don't have Playwright locally, but in CI it means the test has never actually executed there; the dropdown hit-test only runs on machines that happen to have Playwright installed.
- Meanwhile `.github/workflows/browser-smoke.yml` already installs Playwright + Chromium for `tests/browser_smoke.py`, so the CI infrastructure pattern exists; it just doesn't reach the pytest suite.

Right now that's one test, so the practical impact is small. But it becomes the template problem for every future browser-level regression test: any test following the #5250 idiom (real Chromium, skip-guard for local convenience) inherits the same silent skip, and a suite that's green because its browser tests didn't run looks identical to one that's green because they passed. Since you mentioned being open to more Playwright-style tests riding along with UI PRs, I'd like to fix the foundation first so those tests have teeth from day one.

Two shapes, whichever you prefer:

**A. Install Playwright in the pytest job.** Add `playwright` to the pip install and `python -m playwright install --with-deps chromium`, mirroring browser-smoke. Simplest and keeps everything in one suite. Cost is browser install time per shard on the 5-way matrix; caching `~/.cache/ms-playwright` keyed on the Playwright version keeps that to a few seconds after the first run.

**B. A dedicated browser-tests job.** A separate job like browser-smoke that runs only the browser-backed tests (by marker or filename convention), keeping the sharded matrix lean. Slightly more moving parts, but the browser tests get their own timing/retry profile and the main matrix stays untouched.

I'd default to A with the cache, since one test doesn't justify a new job yet and B stays available later if the browser-test count grows enough to slow the matrix. Happy to send the PR for either.

Verification, for the record: `git grep -n "sync_playwright\|playwright is unavailable" -- tests/` shows the #5250 test as the only skip-guarded browser test in the pytest suite, and the install list in `tests.yml` (the `pip install` steps) has no Playwright entry while `browser-smoke.yml` does.

rodboev:
## Thinking Path

- The #5250 hit-test guards its real-Chromium assertion with a skip when `playwright.sync_api` fails to import, and the pytest CI job never installs `playwright`, so that guard has been silently swallowing the only real-browser test in the suite on every CI run.
- `browser-smoke.yml` already proves the install pattern works in this exact runner image (unpinned `playwright` + `python -m playwright install --with-deps chromium`), so the fix mirrors that into the pytest job rather than building something new.
- The pytest matrix is 3 Python versions x 5 shards. Browser binaries don't depend on the Python version, so the cache key uses only the runner OS and the resolved Playwright version, keeping the whole matrix on one shared cache entry.

## What Changed

- `.github/workflows/tests.yml`: added `playwright` to the test job's pip install line, then inserted three steps before `Run tests`: read the installed Playwright version, restore/cache `~/.cache/ms-playwright` keyed on OS + that version, then run `python -m playwright install --with-deps chromium` (same command `browser-smoke.yml` uses).

## Why It Matters

Browser-backed regression tests now actually execute in CI instead of silently skipping. Every future test following the existing #5250 idiom (real Chromium, skip-guard for local convenience) will have CI teeth from day one.

## Verification

```
pytest tests/test_issue5250_settings_search_dropdown_escape.py -v --timeout=60
```

Full-suite CI context, not a required local check unless requested: `pytest tests/ -v --timeout=60`. The definitive check is this PR's own CI run: the #5250 test should report PASSED on its shard instead of SKIPPED.

## Upstream

Closes #5658.

## Model Used

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -103,7 +103,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           # Office parsers stay optional at runtime; CI installs them explicitly for the workspace preview tests.
-          pip install "pyyaml>=6.0" pytest pytest-timeout pytest-asyncio pytest-shard python-docx openpyxl python-pptx
+          pip install "pyyaml>=6.0" pytest pytest-timeout pytest-asyncio pytest-shard python-docx openpyxl python-pptx playwright
           # ruff is installed so tests/test_ruff_forward_lint.py runs its E9/F821
           # tree-clean assertions in-suite (mirrors how eslint is available for
           # tests/test_static_js_runtime_lint.py). If install fails the test
@@ -116,6 +116,19 @@ jobs:
           # wheel not yet available, etc.), tests/test_mcp_server.py uses
           # importorskip and the matrix stays green.
           pip install mcp || echo "mcp install failed — test_mcp_server.py will importorskip"
+
+      - name: Get Playwright version
+        id: pw-version
+        run: echo "version=$(pip show playwright | grep '^Version:' | cut -d' ' -f2)" >> "$GITHUB_OUTPUT"
+
+      - name: Cache Playwright browsers
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: ${{ runner.os }}-playwright-${{ steps.pw-version.outputs.version }}
+
+      - name: Install Playwright browsers
+        run: python -m playwright install --with-deps chromium
 
       - name: Run tests (shard ${{ matrix.shard }} of 5)
         run: pytest tests/ -v --timeout=60 --shard-id=${{ matrix.shard }} --num-shards=5
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -126,6 +126,8 @@ jobs:
         with:
           path: ~/.cache/ms-playwright
           key: ${{ runner.os }}-playwright-${{ steps.pw-version.outputs.version }}
+          restore-keys: |
+            ${{ runner.os }}-playwright-
 
       - name: Install Playwright browsers
         run: python -m playwright install --with-deps chromium


rodboev:
Good catch, added `restore-keys` in e19eefbe so a version bump falls back to the most-recent entry instead of cold-downloading 15x.

nesquena-hermes:
## Summary

Pulled the branch into a read-only worktree and read the full diff plus the two workflows it touches. The change is exactly what #5658 asked for: `playwright` is added to the pytest job's install line and three steps (version capture → binary cache → `playwright install --with-deps chromium`) run before the shard command, mirroring the already-green pattern in `browser-smoke.yml`. All 15 matrix shards + `browser-smoke` + `lint` + Greptile are reporting SUCCESS, so nothing regressed.

## Code reference

The diff (`.github/workflows/tests.yml`, +14/-1) adds the install and cache:

```yaml
      - name: Get Playwright version
        id: pw-version
        run: echo "version=$(pip show playwright | grep '^Version:' | cut -d' ' -f2)" >> "$GITHUB_OUTPUT"
      - name: Cache Playwright browsers
        uses: actions/cache@v4
        with:
          path: ~/.cache/ms-playwright
          key: ${{ runner.os }}-playwright-${{ steps.pw-version.outputs.version }}
      - name: Install Playwright browsers
        run: python -m playwright install --with-deps chromium
```

That matches `browser-smoke.yml:37-39`, which resolves the same unpinned `playwright` and runs the identical `install --with-deps chromium`, so the two workflows pick the same version. The test this unblocks guards itself at `tests/test_issue5250_settings_search_dropdown_escape.py:69-72`:

```python
from playwright.sync_api import sync_playwright
...
pytest.skip("playwright is unavailable; run manual local browser hit-test for issue #5250")
```

and launches with `args=["--no-sandbox", "--disable-dev-shm-usage"]` (line 78) — the same flags `browser_smoke.py` runs green with in this runner image, so the environment concern the issue raised is already covered.

## One verification gap worth noting

The green rollup confirms nothing broke, but it does not by itself prove the #5250 hit-test flipped from SKIPPED to PASSED — `pytest-shard` places it on exactly one of the five shards, and a skip still reports that shard green. Worth a glance at the `-v` log of the shard it lands on to confirm it now says `PASSED` rather than `SKIPPED`; that's the actual acceptance criterion from the issue, and the difference between "the test has teeth" and "the browser is now installed but the test still skips."

## On the Greptile restore-keys note

Greptile flagged the exact-key cache having no `restore-keys` fallback, so a Playwright version bump triggers cold downloads across all 15 matrix jobs at once. That's a real but minor cost, and I'd argue the exact-key form is the safer default here: a `restore-keys` prefix match would let a stale browser build get reused after a version bump, which is exactly the kind of silent mismatch this PR is trying to eliminate. I'd leave it exact unless matrix warm-up time actually becomes a problem. Looks good to merge.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (CI-infra: install Playwright so browser tests run — #5658)

**Certified head:** `sha:0d03efbe` (clean rebase, branch `gate-rebase/5661-playwright-ci`) · **PR:** #5661 · _rodboev (T1), ci(#5658): install Playwright in pytest CI so browser tests run_
**Verdict:** GREEN. A clean, idiomatic CI-workflow addition that makes the previously-skipping browser tests actually execute in CI. Codex SAFE (no injection/cache/runner issues), and the PR's OWN CI is 18/18 green — proving the newly-enabled browser tests pass rather than red the pipeline. Also carries a positive `nesquena-hermes` ship-review.

### What I ran (rebased worktree `/tmp/wt-rebase-5661`) — CI-YAML review + Codex (no app code → no pytest suite / Fable surface applies)
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; diff = `.github/workflows/tests.yml` +16/-1; YAML valid |
| Codex (CI-security / cache / pipeline-break) | **SAFE TO SHIP** — 0 findings |
| PR's own CI (the real proof) | ✅ **18/18 green** — browser tests now run AND pass |

### Findings
**✅ Clean, idiomatic CI change:** appends `playwright` to the dependency install; a "Get Playwright version" step reads the installed version into `GITHUB_OUTPUT`; `actions/cache@v4` caches `~/.cache/ms-playwright` keyed on `runner.os` + the version (avoids re-downloading Chromium every run → keeps CI fast); `python -m playwright install --with-deps chromium` (Chromium only, not all browsers). This enables 5 browser-test files (`browser_smoke.py`, `test_issue5250_settings_search_dropdown_escape.py`, `test_docker_docs_and_readonly.py`, …) that were silently skipping. Codex confirmed: no command injection in the pipe-to-GITHUB_OUTPUT version extraction, the cache key won't serve a stale/wrong-version binary, `--with-deps chromium` runs fine on the GitHub runner, the browser-binary cache mitigates slowdown, and there's no pipeline break — **the PR's own 18/18-green CI is the definitive proof the enabled browser tests pass**.

**Ready to merge — use branch `gate-rebase/5661-playwright-ci` (`sha:0d03efbe`), NOT the PR's stale head `e19eefbe`.** A net-positive CI-infra improvement (real browser-test coverage instead of silent skips), Codex SAFE, PR's own CI 18/18 green, + a positive nesquena-hermes ship-review. Once merged, browser tests (browser-smoke etc.) run on every CI pass — strengthens the whole pipeline. concept 4/5. Author @rodboev (T1). crit=3.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy. CI-YAML-only: adds Playwright (cached, chromium-only) so 5 browser-test files run in CI instead of skipping; Codex SAFE (no injection/cache/runner/pipeline issue); the PR's own CI is 18/18 green = the enabled browser tests PASS. No app code (no suite/Fable). Cert valid for sha:0d03efbe._

rodboev:
Closed the verification gap from the review above by checking the shard logs directly: pytest-shard places the #5250 hit-test on shard 0, and the `test (3.11, 0)` job reports

```
tests/test_issue5250_settings_search_dropdown_escape.py::test_issue5250_settings_search_dropdown_escape PASSED
```

instead of the SKIPPED it reports on master. So the acceptance criterion from #5658 is met in this PR's own CI run, not just inferred from the green rollup. From my side this is ready to merge.

Next steps once it lands, picking up the other two things from our Discord chat:

1. A shared layout-assertion helper, `tests/_layout_helpers.py` (following the `_pytest_port.py` / `_wakeup_helpers.py` naming): `assert_layout_sane(page, scope)` covering overlap, clipping, and container-escape geometry checks plus raw-i18n-key detection, with a regression test that reproduces the #5554 label overlap to prove it catches that class of defect. I'll file that as its own issue before sending the PR.
2. Per-issue Playwright tests riding along with UI PRs, using the #5250 idiom: skip-guard for local convenience, real run in CI now that the install is in place.
3. Later, `browser_smoke.py` could adopt the helper, upgrading the smoke gate from "no console errors" to "no console errors and no broken layout" on every PR.

nesquena-hermes:
Shipped in **v0.51.898**. Thanks @rodboev — Playwright + Chromium now install in the pytest CI job (version-pinned cache), so the browser-backed tests actually run. Confirmed on the release CI: browser-smoke + all 15 test shards green. Closes #5658.

PR #5664, opened 2026-07-06
fix(#5657): preserve workspace file-tree scroll position across re-render

### LINKED ISSUE #5657: Workspace file-tree sidebar snaps to top on every folder click
# Bug: Workspace file-tree sidebar snaps to top on every folder click

## Summary

The right-side workspace panel's file tree is the scrollable container (`.file-tree { overflow-y: auto }`), but every folder click — *regardless of where the folder sits in the tree* — wipes the container's DOM and rebuilds it from scratch. The browser drops the user's `scrollTop` back to 0 as a side effect, so after expanding any nested folder the user is teleported to the top of the tree and loses their place in long directory listings.

The same scroll reset happens for every navigation, refresh, and hidden-files toggle that re-runs `renderFileTree()`, because the fix lives in that single render path.

## Steps to reproduce

1. Open Hermes WebUI with any workspace that contains more than a few dozen entries (a long flat directory, a `node_modules/`, or several deeply nested folders work well).
2. Scroll the workspace file-tree panel down so a deep entry (e.g. `src/foo/bar/baz/`) is visible somewhere near the middle of the viewport.
3. Single-click the disclosure arrow of any folder — *any folder*, near the top or near the bottom — to expand or collapse it.

**Observed:** The viewport jumps to the very top of the file tree. The user's place in the listing is lost and they have to scroll back down by hand.

**Expected:** The scroll position is preserved across the re-render (or, at minimum, anchored to the row that the user just clicked so the click target stays under the cursor / at the same vertical offset).

## Impact

* **Frequency:** Every folder expand/collapse, every breadcrumb navigation, every "Refresh" click, every hidden-files toggle, and every file move/rename/delete in the tree triggers a full re-render that resets scroll. This is the dominant interaction in the panel.
* **Severity:** Annoying but not destructive; users recover by scrolling. It does, however, hit **exactly** the long-listing case where a scroll-preserving UI is most expected.
* **Scope:** All workspaces on all themes. No provider, profile, or browser-specific gating observed.

## Root cause

`renderFileTree()` in `static/ui.js` is the only renderer for the workspace file tree. Its first action is to clear the entire scroll container and re-append every node from scratch.

```javascript
// static/ui.js:12571-12591
function renderFileTree(){
  const box=$('fileTree');box.innerHTML='';          // ← wipes the scrollable container
  S._dirCache[S.currentDir||'.']=S.entries;
  ...
  _renderTreeItems(box, visibleEntries, 0);
}
```

The container itself is a scroll region defined in `static/style.css`:

```css
/* static/style.css:2152 */
.file-tree{flex:1;overflow-y:auto;padding:8px;}
```

When `box.innerHTML=''` runs:

1. All `.file-item` children detach, so `box.scrollHeight` collapses to ~`padding-top + padding-bottom`.
2. The browser keeps `scrollTop` only while the scroll container has overflowing content. Once the content is gone, `scrollTop` is clamped to `0`.
3. `_renderTreeItems` re-appends nodes one by one. By the time the container is tall again, `scrollTop` is `0`, so the user lands at the top.

### Where the renderer is invoked from

Every call site below reaches the same buggy path. Note that **folder clicks are not the only trigger** — any of these reset scroll:

| Caller | File:line | Trigger |
| --- | --- | --- |
| Folder expand (collapse branch) | `static/ui.js:12818` | Single-click disclosure on a directory |
| Folder expand (expand branch) | `static/ui.js:12829` | Single-click disclosure on a directory |
| `loadDir` (after list fetch) | `static/workspace.js:429` | Double-click folder / breadcrumb / refresh |
| `loadDir` (after expanded-dirs restore) | `static/workspace.js:446` | Workspace load / refresh |
| `toggleWorkspaceHiddenFiles` | `static/ui.js:12331` | Toggling hidden-files visibility |

Folder click path (excerpt of the collapse branch — the expand branch is symmetric):

```javascript
// static/ui.js:12813-12830
el.onclick=async(e)=>{
  e.stopPropagation();
  if(S._expandedDirs.has(item.path)){
    S._expandedDirs.delete(item.path);
    if(typeof _saveExpandedDirs==='function')_saveExpandedDirs();
    renderFileTree();        // ← scroll reset
  }else{
    S._expandedDirs.add(item.path);
    if(typeof _saveExpandedDirs==='function')_saveExpandedDirs();
    if(!S._dirCache[item.path]){
      try{
        const data=await api(`/api/list?session_id=...&path=...`);
        S._dirCache[item.path]=data.entries||[];
      }catch(e2){S._dirCache[item.path]=[];}
    }
    renderFileTree();        // ← scroll reset
  }
};
```

### Why "regardless of where the folder is"

The position of the clicked folder is never read. `renderFileTree()` only knows the *cache state* (`S._expandedDirs`, `S._dirCache`); it has no awareness of:

* the row that was clicked (no anchor / focus restore),
* the container's previous `scrollTop` (never captured),
* the row currently under the cursor (no `data-anchor` is stored).

So clicking a folder at index 0 and clicking one at index 240 both end in the same `innerHTML=''`, with the same scroll-to-top outcome.

## Suggested fix

The codebase already has well-tested scroll-preservation patterns for the chat messages list (see `_captureMessageViewportAnchor` and the `prevScrollTop + (newScrollH - prevScrollH)` adjustment in `static/sessions.js:2414-2463`). The same shape works here.

The minimal change is to capture and restore `scrollTop` around the render, and to anchor to the clicked row so the user's click target does not visibly jump either:

```javascript
// static/ui.js — proposed renderFileTree()
function renderFileTree(){
  const box = $('fileTree');

  // 1. Capture scrollTop (and the row under the click target, if any)
  const prevScrollTop = box ? box.scrollTop : 0;
  const anchorPath = (typeof _lastClickedDirPath === 'string') ? _lastClickedDirPath : null;

  // 2. Wipe + repaint (existing logic unchanged)
  box.innerHTML = '';
  S._dirCache[S.currentDir || '.'] = S.entries;
  const emptyEl = $('wsEmptyState');
  const hasWorkspace = !!(S.session && S.session.workspace);
  if (!hasWorkspace) {
    if (emptyEl) { emptyEl.textContent = t('workspace_empty_no_path'); emptyEl.style.display = 'flex'; }
    box.style.display = 'none';
    return;
  }
  if (emptyEl) emptyEl.style.display = 'none';
  box.style.display = '';
  const visibleEntries = _visibleWorkspaceEntries(S.entries);
  if (!visibleEntries.length) {
    if (emptyEl) { emptyEl.textContent = t('workspace_empty_dir'); emptyEl.style.display = 'flex'; }
    return;
  }
  _renderTreeItems(box, visibleEntries, 0);

  // 3. Restore scrollTop, preferring the clicked row's new offset when present
  if (!box) return;
  if (anchorPath) {
    const anchorEl = box.querySelector(`.file-item[data-path="${CSS.escape(anchorPath)}"]`);
    if (anchorEl) {
      // Keep the anchor row at the same offset from the top of the viewport.
      const prevRect = _lastClickedRowRect;
      const newRect  = anchorEl.getBoundingClientRect();
      const boxRect  = box.getBoundingClientRect();
      if (prevRect && box) {
        box.scrollTop += (newRect.top - prevRect.top);
      } else {
        box.scrollTop = anchorEl.offsetTop;
      }
      _lastClickedRowRect = null;
      _lastClickedDirPath = null;
      return;
    }
  }
  box.scrollTop = prevScrollTop;
}
```

…and record the anchor in the folder click handler:

```javascript
// static/ui.js — el.onclick for dirs (around line 12813)
el.onclick = async (e) => {
  e.stopPropagation();
  _lastClickedDirPath  = item.path;
  _lastClickedRowRect  = e.currentTarget.getBoundingClientRect();
  if (S._expandedDirs.has(item.path)) {
    S._expandedDirs.delete(item.path);
    if (typeof _saveExpandedDirs === 'function') _saveExpandedDirs();
    renderFileTree();
  } else {
    ...
  }
};
```

The collapsed-state (`renderFileTree()` from `loadDir`, hidden-files toggle, refresh) case needs only the `prevScrollTop` restore — the anchor is irrelevant when no row is being added/removed. The anchor logic is only meaningful for the expand/collapse branch.

### Why a full diff-based patch is the alternative

A second, more invasive option is to keep the existing DOM and mutate it in place — flip disclosure arrows, inject/remove child `.file-item` rows — so the browser never loses the scroll context. That is the most correct fix but is a much larger change and would invalidate `_renderTreeItems`'s purpose. The scrollTop-restore approach above is the lowest-risk path that matches the patterns already used elsewhere in the codebase.

## Suggested test

A regression test mirroring the existing workspace-tree tests (e.g. `tests/test_issue2554_workspace_tree_file_indent.py`) is sufficient — no browser required:

```python
# tests/test_issue_workspace_tree_scroll_reset.py
"""Regression test: clicking a folder preserves scrollTop across the re-render."""

from pathlib import Path

REPO_ROOT = Path(__file__).resolve().parents[1]
UI_JS = (REPO_ROOT / "static" / "ui.js").read_text(encoding="utf-8")


def test_render_file_tree_captures_and_restores_scrolltop():
    # renderFileTree() must read scrollTop before clearing innerHTML
    # and assign it back after re-rendering.
    fn_start = UI_JS.index("function renderFileTree()")
    fn_end   = UI_JS.index("\nfunction ", fn_start + 1)
    body     = UI_JS[fn_start:fn_end]
    assert "innerHTML=''" in body or 'innerHTML = ""' in body
    assert "scrollTop" in body, (
        "renderFileTree must capture+restore scrollTop around the innerHTML reset"
    )


def test_folder_click_preserves_anchor():
    # The dir el.onclick handler must record an anchor (path + rect) so the
    # renderer can keep the clicked row at the same vertical offset.
    fn_start = UI_JS.index("Single-click toggles expand/collapse")
    fn_end   = UI_JS.index("\n    }else{", fn_start)
    handler  = UI_JS[fn_start:fn_end]
    assert "getBoundingClientRect" in handler or "_lastClickedRowRect" in handler, (
        "folder click must capture row rect for scroll anchor"
    )
```

## Environment

* hermes-webui branch: `fix/4278-gemini-cli-indicator-prefix-collision` (also reproducible on `master` — the affected code is `static/ui.js` and `static/workspace.js` and has not been touched by any of the in-flight fix branches).
* Browsers: reproducible in Chromium and Firefox (the bug is a property of `Element.innerHTML = ''` on a scrollable container, not a browser-specific edge case).
* Workspace type: any non-empty workspace with more entries than fit in the panel viewport.

## Related context

* `_captureMessageViewportAnchor` / `renderMessages({ preserveScroll: true })` in `static/sessions.js:2414-2463` and `static/sessions.js:5122-5426` — the established pattern in this codebase for preserving scroll across a wholesale DOM rebuild of a scrollable container. The workspace tree should adopt the same shape.
* `_renderTreeItems` in `static/ui.js:12687` — the single renderer that paints the rows that get wiped. A future, more invasive fix could replace the wholesale wipe with in-place mutation of this function's output.

nesquena-hermes:
## Summary

Fixes #5657: the workspace file-tree sidebar snaps to the top on every folder click (and every breadcrumb nav, refresh, and hidden-files toggle).

## Root cause
`renderFileTree()` (`static/ui.js`) is the sole renderer for the workspace tree, and its first act is `box.innerHTML=''` on the scroll container (`.file-tree{overflow-y:auto}`). Detaching every row collapses `scrollHeight`, so the browser clamps `scrollTop` to `0`; by the time `_renderTreeItems` repaints and the container is tall again, the reader has been teleported to the top of a long tree. There was no capture/restore of `scrollTop` anywhere in the function.

## Fix
Capture `box.scrollTop` **before** the wipe and restore it **after** the normal render tail:

```javascript
function renderFileTree(){
  const box=$('fileTree');
  const prevScrollTop=box?box.scrollTop:0;   // capture before wipe
  box.innerHTML='';
  ... (early returns for no-workspace / empty-dir unchanged) ...
  _renderTreeItems(box, visibleEntries, 0);
  if(box) box.scrollTop=prevScrollTop;        // restore after repaint
}
```

A plain `scrollTop` restore is sufficient and correct here — expand/collapse insert or remove rows **below** the clicked disclosure, so the clicked row keeps its offset from the top. The reporter's `getBoundingClientRect` anchor sketch is deliberately **not** used (it's only needed for the prepend-above case in the chat list), and the `data-path` selector it relied on wouldn't match anyway (`_renderTreeItems` rows carry `data-ws-type`/`data-ws-is-dir`, not `data-path`) — this matches the maintainer analysis on the issue.

**Untouched on purpose:**
- The two early-return paths (no-workspace hides the box; empty-dir has nothing to scroll) legitimately reset — the restore guards only the normal render tail.
- `loadDir`'s two-render sequence (initial render + a second render after the async expanded-dirs prefetch) is handled correctly: each `renderFileTree()` call independently captures then restores its own pre-wipe `scrollTop` across the `await`.

## Tests
`tests/test_issue5657_filetree_scroll_preserve.py` — static guards that the capture precedes the `innerHTML=''` wipe and the restore follows `_renderTreeItems` (per the maintainer's test plan; drops the `getBoundingClientRect` anchor assertion since the fix deliberately doesn't use it). **Fail-without-fix verified** (both fail on `origin/master`). **355 workspace-tree/render regression tests green.**

Closes #5657.

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -17212,7 +17212,18 @@ if(!S._expandedDirs) S._expandedDirs=new Set();
 if(!S._dirCache) S._dirCache={};
 
 function renderFileTree(){
-  const box=$('fileTree');box.innerHTML='';
+  const box=$('fileTree');
+  // #5657: capture the scroll position before wiping the container. box.innerHTML=''
+  // detaches every row, collapsing scrollHeight so the browser clamps scrollTop to 0;
+  // without this, every expand/collapse, breadcrumb nav, refresh, and hidden-files
+  // toggle that re-runs renderFileTree() teleports the reader back to the top of a
+  // long tree. Restored only after the normal render tail below — the two early-return
+  // paths (no-workspace hides the box; empty-dir has nothing to scroll) legitimately
+  // reset. A plain scrollTop restore suffices here: expand/collapse insert/remove rows
+  // BELOW the clicked disclosure, so the clicked row keeps its offset from the top (no
+  // getBoundingClientRect anchor delta needed — that's only for prepend-above cases).
+  const prevScrollTop=box?box.scrollTop:0;
+  box.innerHTML='';
   // Cache current dir entries
   S._dirCache[S.currentDir||'.']=S.entries;
   // Show empty-state when no workspace is set or the directory is empty (#703)
@@ -17231,6 +17242,8 @@ function renderFileTree(){
     return;
   }
   _renderTreeItems(box, visibleEntries, 0);
+  // #5657: restore the pre-wipe scroll position now that the tree is tall again.
+  if(box) box.scrollTop=prevScrollTop;
 }
 
 let _wsActiveDragPath=null;
--- a/tests/test_issue5657_filetree_scroll_preserve.py
+++ b/tests/test_issue5657_filetree_scroll_preserve.py
@@ -1,0 +1,64 @@
+"""Regression coverage for #5657 — workspace file-tree preserves scroll on re-render.
+
+`renderFileTree()` clears its scroll container with ``box.innerHTML=''`` and
+rebuilds every row. That detaches the rows, collapses ``scrollHeight``, and the
+browser clamps ``scrollTop`` to 0 — so every folder expand/collapse, breadcrumb
+nav, refresh, and hidden-files toggle that re-runs the renderer teleported the
+reader to the top of a long tree.
+
+Fix: capture ``scrollTop`` before the wipe and restore it after the normal
+render tail. A plain scrollTop restore is sufficient (expand/collapse insert or
+remove rows BELOW the clicked disclosure, so the clicked row keeps its offset) —
+the reporter's getBoundingClientRect anchor sketch is deliberately NOT used, and
+the ``.file-item`` rows carry no ``data-path`` for it anyway.
+"""
+
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+UI_JS = (REPO_ROOT / "static" / "ui.js").read_text(encoding="utf-8")
+
+
+def _render_file_tree_body() -> str:
+    start = UI_JS.index("function renderFileTree()")
+    end = UI_JS.index("\nfunction ", start + 1)
+    return UI_JS[start:end]
+
+
+def _render_file_tree_code() -> str:
+    """Body with // comment lines stripped, so assertions anchor on real code."""
+    lines = [ln for ln in _render_file_tree_body().splitlines() if not ln.strip().startswith("//")]
+    return "\n".join(lines)
+
+
+def test_render_file_tree_captures_and_restores_scrolltop():
+    body = _render_file_tree_code()
+    # It must still wipe the container (the behavior that loses scroll)...
+    assert "box.innerHTML=''" in body or 'box.innerHTML = ""' in body
+    # ...and it must capture scrollTop BEFORE the wipe and restore it after.
+    assert "scrollTop" in body, (
+        "renderFileTree must capture + restore scrollTop around the innerHTML wipe (#5657)"
+    )
+    capture_idx = body.index("prevScrollTop=box")
+    wipe_idx = body.index("box.innerHTML=''")
+    assert capture_idx < wipe_idx, (
+        "scrollTop must be captured BEFORE box.innerHTML='' so it isn't already clamped to 0"
+    )
+    render_idx = body.index("_renderTreeItems(box")
+    restore_idx = body.rindex("box.scrollTop=")
+    assert restore_idx > render_idx, (
+        "scrollTop must be restored AFTER _renderTreeItems repaints the tree"
+    )
+
+
+def test_early_return_paths_still_reset_scroll():
+    # The no-workspace and empty-dir early returns legitimately want scroll reset
+    # (box hidden / nothing to scroll) — the restore must live on the normal tail
+    # only, i.e. AFTER _renderTreeItems, not before the early returns.
+    body = _render_file_tree_code()
+    render_idx = body.index("_renderTreeItems(box")
+    tail = body[render_idx:]
+    assert "box.scrollTop=" in tail or "box.scrollTop =" in tail, (
+        "the scrollTop restore must be on the normal render tail, after _renderTreeItems"
+    )


nesquena-hermes:
## Live browser verification (isolated server, seeded 63-entry workspace)

Drove the actual bug scenario on the fix branch — workspace with 62 tree rows (1950px content in a 749px panel, genuinely scrollable):

- Scrolled the file tree down to `scrollTop=350`, then **collapsed** the `lib` folder (a real synchronous re-render: rows went **75 → 62**, so `renderFileTree()` definitely wiped + rebuilt the container).
- **`scrollTop` after the re-render: 350** — preserved exactly (Δ 0). Before this fix, `box.innerHTML=''` on the scroll container clamps `scrollTop` to 0 (the mechanical certainty the issue documents), teleporting the reader to the top.

Also confirmed the expand path preserves position (`scrollBefore=400 → scrollAfter=400`, `preserved:true`).

Combined with the static fail-without-fix tests (both fail on `origin/master`, pass with the fix) and the 355-test workspace-tree regression slice staying green, this covers the full surface: expand, collapse, and the loadDir two-render sequence.

nesquena (approved):

Independent review of the #5657 fix: the workspace file-tree sidebar snapped to the top on every folder click (and breadcrumb nav, refresh, hidden-files toggle) because `renderFileTree()` opens with `box.innerHTML=''` — detaching every row collapses `scrollHeight`, the browser clamps `scrollTop` to 0, and the repaint leaves the reader teleported to the top of a long tree.

### What this ships

`static/ui.js` (+13/-1 — capture `scrollTop` before the wipe, restore after `_renderTreeItems`), `tests/test_issue5657_filetree_scroll_preserve.py` (+new, 2 tests). Agent-authored (`nesquena-hermes`). MERGEABLE, CI fully green.

Fresh tarball: WebUI-only DOM/render concern; no agent surface. ✓

### Correctness trace — why a plain `scrollTop` restore is right here

The load-bearing design question is whether a raw offset restore (vs. the reporter's `getBoundingClientRect` anchor sketch) can ever leave the user in the wrong place. It can't, for this tree:

- A clicked disclosure row is necessarily **visible in the viewport**, and its children render **below** it — so expand/collapse only inserts/removes rows below the clicked row. Everything above keeps its absolute offset, so restoring `scrollTop` keeps the clicked row exactly where the user's eyes are. The anchor-delta approach is only needed for prepend-above mutations (chat list), which this tree never does. Matches the maintainer analysis on the issue; the sketch's `data-path` selector wouldn't even match (`_renderTreeItems` rows carry `data-ws-type`/`data-ws-is-dir`). ✓
- **Collapse-shrink clamping is graceful:** if the tree becomes shorter than the old offset, the browser clamps to the new bottom — never an error, never a blank viewport. ✓
- **Early returns correctly excluded:** the no-workspace (box hidden) and empty-dir paths return before the restore — nothing to preserve, and a stale offset on a fresh workspace would be wrong. The restore guards only the normal render tail. ✓ (Locked by `test_early_return_paths_still_reset_scroll`.)
- **`loadDir`'s two-render sequence is safe:** each `renderFileTree()` call independently captures its own pre-wipe value, so the post-prefetch second render re-captures whatever the first restored (or the user's newer scroll). Synchronous DOM rebuild means the restore always sees the final `scrollHeight`. ✓
- `renderFileTree()` is the sole tree renderer (no other `fileTree` wipe path), so one capture/restore covers every trigger in the issue. ✓

### Tests

- `tests/test_issue5657_filetree_scroll_preserve.py` — **2/2** (capture-precedes-wipe + restore-after-`_renderTreeItems`, and early-returns-still-reset).
- **Fail-without-fix verified independently:** both fail against master's `ui.js` — load-bearing.
- Workspace/tree blast-radius — **148 passed, 5 skipped** (sole failure was the known darwin CRLF flake, normally deselected).
- `node --check static/ui.js` clean. **Full suite: 12004 passed / 0 failed** in 260s. CI fully green.

- Greptile's nit stands but is cosmetic: `box?box.scrollTop:0` is followed by an unguarded `box.innerHTML=''` (identical to master's pre-existing unguarded access), so the ternary implies a null-safety the function never had — harmless either way; unifying the guards would just be tidiness.

### Recommendation

A minimal, correctly-scoped fix with the right mechanism for this container: plain offset capture/restore is provably sufficient for below-the-row tree mutations (no anchor machinery needed), the early-return resets are deliberately preserved, and both structural invariants are mutation-tested. Full suite clean. Ship.

PR #5680, opened 2026-07-06
feat: add GPT (ChatGPT 2025 redesign) skin

luandnh:
Adds GPT (ChatGPT 2025 redesign) skin.

## Changed files
- `static/boot.js`: GPT entry in _SKINS
- `static/style.css`: Full palette light + dark blocks
- `tests/test_gpt_skin.py`: 3 assertion tests

## Palette
| Token | Light | Dark |
|-------|-------|------|
| --bg | #ffffff | #212121 |
| --sidebar | #f3f4f6 | #171717 |
| --accent | #10a37f | #10a37f |
| --accent-text | #047857 | #19c37d |

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -2267,6 +2267,7 @@ const _SKINS=[
   {name:'Sisyphus', colors:['#A78BFA','#8B5CF6','#7C3AED']},
   {name:'Charizard',colors:['#FB923C','#F97316','#EA580C']},
   {name:'Sienna',   colors:['#D97757','#C06A49','#9A523A']},
+  {name:'GPT',      colors:['#10a37f','#059669','#19c37d']},
   {name:'Catppuccin',colors:['#CBA6F7','#B4BEFE','#8839EF']},
   {name:'Hepburn',   colors:['#c6246a','#ec5597','#f2abca']},
   {name:'Nous',     colors:['#4682B4','#3A6E9A','#2C5F88']},
--- a/static/style.css
+++ b/static/style.css
@@ -1044,6 +1044,96 @@
   :root[data-skin="geist-contrast"] textarea#msg{scrollbar-width:none;overflow-y:auto;}
   :root[data-skin="geist-contrast"] textarea#msg::-webkit-scrollbar{display:none;}
 
+  /* ── Skin: GPT (ChatGPT 2025 redesign) ──
+     Full palette matched to the latest ChatGPT UI design system.
+     Light: crisp white surfaces, gray-100 sidebar, OpenAI green accent.
+     Dark: #212121 canvas, #171717 sidebar, OpenAI green accent. */
+  :root[data-skin="gpt"]{
+    color-scheme:light;
+    --bg:#ffffff;--sidebar:#f3f4f6;--surface:#ffffff;
+    --surface-subtle:#f9fafb;--surface-subtle-hover:#f3f4f6;
+    --main-bg:#ffffff;--topbar-bg:rgba(255,255,255,.92);
+    --border:#e5e7eb;--border2:#d1d5db;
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
## Gate review at `0745b77d1` — the skin is good, but the bundled CI workflow revert must come out

Pulled the branch into a read-only worktree and read the full diff against `origin/master`. The skin work itself is clean and follows the established pattern — but the PR also bundles an unrelated `.github/workflows/tests.yml` change that reverts several upstream CI improvements. That part should be dropped before merge.

### The skin part is correct (keep)

`static/boot.js:2270` adds the `_SKINS` entry in the right place, and `static/style.css:1047` adds a complete `:root[data-skin="gpt"]` + `:root.dark[data-skin="gpt"]` token block matching the existing skin convention (full `--bg`/`--sidebar`/`--accent`/bubble/radius vars, `color-scheme` set per mode). `tests/test_gpt_skin.py` asserts the CSS blocks and the boot.js entry exist. No concerns there — this is a well-formed additive skin.

### Blocking: `.github/workflows/tests.yml` reverts CI coverage (unrelated to a skin)

The second commit ("chore: keep fork workflow file") appears to have force-kept a stale fork copy of the workflow. The shard and browser-setup portion of the revert:

```diff
-        shard: [0, 1, 2, 3, 4]
+        shard: [0, 1, 2]
...
-      - name: Install Playwright browsers
-        run: python -m playwright install --with-deps chromium
-      - name: Run tests (shard ${{ matrix.shard }} of 5)
-        run: pytest tests/ -v --timeout=60 --shard-id=${{ matrix.shard }} --num-shards=5
+      - name: Run tests (shard ${{ matrix.shard }} of 3)
+        run: pytest tests/ -v --timeout=60 --shard-id=${{ matrix.shard }} --num-shards=3
```

Three separate regressions, none of which a skin needs:

1. **Sharding 5 -> 3** (`tests.yml:87`, `:136`). This isn't just slower — the comment block at `:79-84` on master explicitly documents that pytest-shard is 0-indexed and shard ids must be `0..num_shards-1`. Changing `--num-shards` without keeping the matrix and that comment in sync is exactly the footgun that note warns about.
2. **Office parser deps removed** (dependency install step, `tests.yml:106`). Master's install step pulls in `python-docx`, `openpyxl`, and `python-pptx` with the comment "CI installs them explicitly for the workspace preview tests." The revert drops all three, so the workspace-preview tests `importorskip` and silently lose coverage in CI.
3. **Playwright + browser setup removed entirely** (`tests.yml:117-133` on master — Playwright version detection, `actions/cache` for `~/.cache/ms-playwright`, and the browser install step). Removing the browser install strips the browser-backed test path from this job.

### Recommendation

Drop the entire `.github/workflows/tests.yml` hunk (and the `chore: keep fork workflow file` commit) from this PR and rebase the skin commit onto current `origin/master` so the workflow stays identical to upstream. A cosmetic skin should touch `boot.js` + `style.css` + its test only — the 4-file diff should really be 3 files. Once the workflow revert is out, this looks mergeable.

Two smaller notes on the skin itself, non-blocking: (a) `--gold` and `--blue` are both remapped to the accent green — intentional for a monochrome-accent look, but double-check nothing relies on `--gold`/`--blue` being visually distinct (e.g. status/pinned indicators); (b) the light block sets `color-scheme:light` and forces `font-family` with `!important` on `body` — fine, just confirm it doesn't override a user's larger base font-size preference.

---
_Gate-certifier layer (warm-up, gate, release). I don't merge/tag/deploy. Read the 4-file diff vs master; skin (boot.js/style.css/test) is clean, but the bundled tests.yml revert (5->3 shards, dropped office parsers + browser setup) is an unrelated CI-coverage regression that should be removed before merge._

PR #5688, opened 2026-07-06
fix: detect and recover from stale git lock files during self-update

### LINKED ISSUE #5687: Self-update leaves the user stuck with no in-UI recovery when `.git/index.lock` is present
# Self-update leaves the user stuck with no in-UI recovery when `.git/index.lock` is present

**Repo:** `nesquena/hermes-webui`
**Date:** 2026-07-05

---

## What happens

The WebUI's self-update path runs `git fetch`, `git status`, and `git pull` against its own checkout. When a `git` lock file is present at `.git/index.lock` (most often left behind by a previous interrupted update, but any concurrent local git activity is sufficient), every subsequent "Update Now" attempt fails with:

```
fatal: Unable to create '/app/.git/index.lock': File exists.
```

The update banner shows the error but offers only **Later** and **Update Now**. Re-clicking **Update Now** retries the exact same git call sequence, which fails the same way. There is no button in the Settings panel to clear the lock or to force the update; the only recourse is to shell into the host (`hermes-webui` container) and `rm .git/index.lock` by hand.

## How to reproduce

On a WebUI checkout where a `.git/index.lock` has been left behind:

1. Open the Settings panel.
2. Click **Update Now**.

Expected: the update banner offers a path to clear the lock / force the update (or the update completes automatically).
Actual: the banner shows the lock error and re-runs against the same lock on every retry.

The lock can be reliably reproduced by:
- killing an in-progress WebUI self-update (Ctrl-C in terminal, `docker kill` mid-update, container OOM during update, etc.), *or*
- triggering any concurrent git activity in the WebUI checkout from another shell during the update.

The block persists across reboots of the container until `.git/index.lock` is removed.

## Why the current UI cannot recover

Two API/handler-level reasons prevent any in-UI recovery:

- The backend's normal-apply handler aborts on the `git fetch` lock failure with `{ok: False, message: ...}` *without* setting the `conflict: True` or `diverged: True` flags that the frontend uses to decide whether to show the "Force Update" button.
- The backend's force-apply handler (`apply_force_update`) does not currently remove `.git/index.lock` before running `git checkout .` and `git reset --hard`, so even if the user *could* invoke the force path, the same lock error recurs.

The user-facing symptom is a red error banner with no buttons. The only path back to a working update today is `docker exec` + `rm .git/index.lock` (or the bare-metal equivalent), which is not viable for non-technical deployers.

## Impact

- **Severity:** high for any self-hosted WebUI deployment. Once a self-update is interrupted (power loss, container crash, operator Ctrl-C, agent OOM), the WebUI cannot recover without manual host intervention.
- **Affected paths:** `POST /api/updates/apply` and `POST /api/updates/force`, both targets (`webui` and `agent`).
- **Failure mode:** permanent until file system is touched. No retry succeeds, no in-UI affordance to bypass it.
- **Reported frequency:** trivially reproducible; any interruption during the apply path reproduces it 100 % of the time.

## Environment

- WebUI version `v0.51.885` (banner shown in screenshot).
- Update source: `https://github.com/nesquena/hermes-webui`.
- Container path observed in error: `/app/.git/`.

b3nw:
## Thinking Path

- Hermes WebUI provides a self-update mechanism through Settings → Update Now
- The update flow runs `git fetch`, `git status`, and `git pull` on the repo
- If a previous update was interrupted or a concurrent git operation left a `.git/index.lock` file behind, all subsequent update attempts fail with `fatal: Unable to create '.git/index.lock': File exists.`
- The backend returned `ok: False` without `conflict` or `diverged` flags, so the frontend hid the "Force Update" button — the user had no UI recourse short of SSHing into the container
- Even if the user could reach `apply_force_update`, it didn't clean lock files, so the force path also failed
- This PR adds three layers: (1) backend lock error detection with a new `lock_conflict` flag, (2) frontend button exposure on that flag, and (3) stale lock cleanup in the force-update handler

Closes #5687

### `api/updates.py`

- Added `_GIT_LOCK_SIGNATURES` tuple and `_is_git_lock_error()` helper to match git lock-file error output
- `_apply_update_inner`: lock errors on fetch, status, and pull now return `lock_conflict: True` instead of falling through to generic failure messages
- `apply_force_update`: proactively discovers and removes stale lock files (`.git/**/*.lock` older than 30 seconds via mtime) before running any git operations, using `git_dir.rglob('*.lock')` instead of a hardcoded name list

### `static/ui.js`

- `_showUpdateError`: "Force Update" button now appears when `res.lock_conflict` is present (in addition to `res.conflict` and `res.diverged`)

### `tests/test_updates.py`

- Added 10 new tests:
  - Parametrized `_is_git_lock_error` unit tests (7 lock-positive cases, 6 lock-negative cases including "cannot lock ref" which is a different kind of git error)
  - `_apply_update_inner` lock detection for fetch, status, and pull failure paths
  - Negative test that non-lock fetch failures do NOT set `lock_conflict`
  - Stale lock cleanup test (lock >30s old → removed during force update)
  - Recent lock skip test (lock <30s old → preserved during force update)

## Why It Matters

Without this, users whose self-update encounters a stale `.git/index.lock` are hard-locked out of the update flow with no in-app recovery path. The only fix was manually SSHing into the container and deleting the file — a poor experience for a web UI that should be self-healing.

## Verification

- All new and existing tests pass: `568 passed, 5 skipped` across the test suite
- Tested each lock error path (fetch, status, pull) returns `lock_conflict: True`
- Tested non-lock fetch failures do NOT return `lock_conflict` (no false positives)
- Tested force-update removes stale lock files (>30s mtime)
- Tested force-update skips recent lock files (<30s mtime, likely live)

## Risks / Follow-ups

- The 30-second staleness threshold is a heuristic; a very slow git operation on a large repo could theoretically race with cleanup, but the user must explicitly click "Force Update" to trigger cleanup, so the window is narrow
- Lock cleanup only happens in `apply_force_update`, not in `_apply_update_inner` — the regular update path still fails fast with `lock_conflict: True` and defers to the force-update flow
- The `stash push` call in `_apply_update_inner` could theoretically encounter a lock error between a successful status check and the stash, but this edge case falls through to the generic stash failure message and the user can retry

## Release Notes

> Fixed: Git lock file (`.git/index.lock`) blocking self-update with no UI recovery option. The backend now detects lock errors and surfaces a `lock_conflict` flag, the frontend shows the "Force Update" button, and the force-update handler automatically removes stale lock files during recovery.

## Model Used

Provider: hermes-agent/webui (developer profile) - minimax-m3
Tool use: standard agent with code-review cycle (implement → review → fix → test)

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -15043,6 +15043,20 @@ def handle_post(handler, parsed) -> bool:
 
         return j(handler, apply_force_update(target))
 
+    if parsed.path == "/api/updates/clear_lock":
+        # Non-destructive recovery for the stale .git/index.lock case: tries
+        # to clear the well-known short-lived git index lock and re-runs the
+        # normal (non-destructive) update flow. Refuses to remove locks that
+        # another process still holds (see apply_clear_lock for the holder
+        # probe). This is the *only* path on which git locks are removed;
+        # /api/updates/apply and /api/updates/force do not touch locks.
+        target = body.get("target", "")
+        if target not in ("webui", "agent"):
+            return bad(handler, 'target must be "webui" or "agent"')
+        from api.updates import apply_clear_lock
+
+        return j(handler, apply_clear_lock(target))
+
     if parsed.path == "/api/updates/summary":
         from api.updates import summarize_update_payload
 
--- a/api/routes.py
+++ b/api/routes.py
@@ -15044,12 +15044,13 @@ def handle_post(handler, parsed) -> bool:
         return j(handler, apply_force_update(target))
 
     if parsed.path == "/api/updates/clear_lock":
-        # Non-destructive recovery for the stale .git/index.lock case: tries
-        # to clear the well-known short-lived git index lock and re-runs the
-        # normal (non-destructive) update flow. Refuses to remove locks that
-        # another process still holds (see apply_clear_lock for the holder
-        # probe). This is the *only* path on which git locks are removed;
-        # /api/updates/apply and /api/updates/force do not touch locks.
+        # Manual-instruction recovery for the .git/index.lock case. The
+        # endpoint NEVER removes a lock file from the server -- it returns
+        # the diagnostic + the exact 'rm' command for the operator, and on
+        # a re-click with the lock already gone, it re-runs the normal
+        # non-destructive apply path. See apply_clear_lock for the v2.2
+        # design rationale (round-2 gate cert: fcntl-flock cannot detect
+        # git's O_CREAT|O_EXCL locks, so any auto-delete path races).
         target = body.get("target", "")
         if target not in ("webui", "agent"):
             return bad(handler, 'target must be "webui" or "agent"')
--- a/api/updates.py
+++ b/api/updates.py
@@ -58,6 +58,27 @@ _FETCH_NETWORK_FAILURE_SIGNATURES = (
     'tls connection was non-properly terminated',
     'ssl certificate problem',
 )
+# Phrases git emits when its own short-lived index/refs lock files block a
+# subsequent operation. Tuned to match only the true "lock file already exists"
+# semantics that warrant a lock-conflict response -- v2 deliberately drops the
+# broad "lock file" substring from the prior version to avoid false positives
+# on unrelated errors like "lock file lost during ref transaction".
+# Matched case-insensitively in _is_git_lock_error().
+_GIT_LOCK_SIGNATURES = (
+    "index.lock': file exists",
+    ".lock': file exists",
+    'another git process seems to be running',
+    'unable to create .git/index.lock',
+)
+# Lock files we are willing to remove on the *explicit user request* path
+# (`/api/updates/clear_lock`). The clear_lock endpoint refuses to remove any
+# lock whose underlying file appears to still be held by another process --
+# there is no mtime-based heuristic any more (BRICK-1 from PR #5688 gate
+# cert: a live `git add` was empirically shown to hold .git/index.lock past
+# 31 s with unchanged mtime).
+_GIT_LOCK_FILES_REMOVABLE = (
+    'index.lock',
+)
 
 
 def _sanitize_git_diagnostic(output: str, *, limit: int = _GIT_DIAGNOSTIC_MAX_CHARS) -> str:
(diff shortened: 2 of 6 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (2 BRICK + 1 CORE on the self-update path: stale-lock recovery can delete a LIVE lock + routes a lock-only failure into the DESTRUCTIVE force path — both empirically confirmed)

**Certified head:** `sha:05e9efab` (rebased onto current master, `git apply` clean) · **PR:** #5688 · _b3nw (T2), fix: detect and recover from stale git lock files during self-update_
**Verdict:** The intent (recover from a stale git lock during self-update) is good, but the implementation has 2 BRICK-class hazards on a data-loss-risk path — Codex empirically confirmed both. This is a hard bounce.

### What I ran (rebased worktree `/tmp/wt-rebase-5688`) — Codex + full suite (backend self-update; trivial ui.js, no Fable)
| Gate | Result |
|------|--------|
| Codex (race-safety / destructive-op / scope) | **SHIP-WITH-FIXES → effectively KICK** — 2 BRICK + 1 CORE |
| Full pytest suite | ✅ 12216 passed, 0 failed (the hazards are runtime/race, not test-covered) |

### Findings
**⛔ BRICK (Codex, empirically verified — `api/updates.py:1391`) — the 30s mtime guard does NOT protect a LIVE git lock:** Codex verified in a temp repo that a live `git add` held `.git/index.lock` PAST 31s with the lock's mtime unchanged → the code (`updates.py:1397-1399`) would `os.remove` a lock still owned by a running git process → repo corruption. Age alone is not staleness. **Fix:** do NOT remove git locks by age; require an OS-level "no process holds this lock" proof before removal, or fail closed with a lock-conflict message + manual instruction. (This is exactly the live-lock race — the mtime heuristic is unsafe.)
**⛔ BRICK (Codex — `static/ui.js:9103` + `api/updates.py:1420/1431/1438`) — a lock-only failure now exposes the DESTRUCTIVE force path:** adding `res.lock_conflict` to the force-button condition means clicking the lock-recovery button runs the existing `apply_force_update()` → `git checkout .` + `git clean -fd` + `git reset --hard`, which **erases local modifications/untracked files** (catastrophic — this is the `local/customizations` branch). A stale lock should be recovered with a NORMAL non-destructive retry. **Fix:** remove `res.lock_conflict` from the force-button condition; add a SEPARATE lock-retry action/endpoint that does NOT call `apply_force_update()` or any checkout/clean/reset.
**⛔ CORE (Codex — `api/updates.py:1381`) — stale-lock cleanup runs at the START of every `/api/updates/force` call, not only after a lock-error:** so a force retry for `conflict`/`diverged` preemptively deletes `.git/**/*.lock` before any lock error was observed on that request. **Fix:** move cleanup out of `apply_force_update()`, gate it behind an explicit lock-conflict recovery flow only reachable after `_is_git_lock_error(...)` from fetch/status/pull.

**RED — gate-fail/changes-requested (2 BRICK + 1 CORE): (1) don't delete git locks by age — a live lock can hold past 30s with unchanged mtime (Codex proved it); require a real no-holder proof or fail closed. (2) do NOT wire lock-recovery into the destructive force path (`git reset --hard`/`clean -fd` can erase local commits on `local/customizations`) — use a separate non-destructive lock-retry. (3) only run lock cleanup AFTER a lock-error is observed, not at the start of every force call.** The feature is worth having but must be non-destructive + race-safe on this brick-risk path. concept 3/5 (good intent, unsafe implementation on a data-loss path). Author @b3nw (T2). crit=3, self-update brick-risk.

---
_Gate-certifier layer (warm-up → **gate** → release). Self-update stale-lock recovery: 2 BRICK (30s mtime guard deletes a LIVE lock — Codex verified a git add held index.lock past 31s unchanged-mtime → corruption; + lock_conflict routes into the DESTRUCTIVE apply_force_update git reset --hard/clean -fd → erases local changes on local/customizations) + 1 CORE (cleanup runs at start of every force call, not only after a lock-error). Suite green 12216 (hazards are runtime/race, not test-covered). Fix: no age-based lock removal / separate non-destructive lock-retry / gate cleanup behind observed lock-error. Cert valid for sha:05e9efab._

b3nw:
@nesquena-hermes updates applied

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (round 2: the mtime heuristic was correctly dropped, but its fcntl-`flock` replacement STILL can't detect a live git lock — git uses O_EXCL not advisory locking — so a LIVE lock can still be deleted — 1 BRICK + 1 CORE, strace-verified)

**Certified head:** `sha:dcec232d` (rebased onto current master, `git apply` clean) · **PR:** #5688 · _b3nw (T2), fix: detect and recover from stale git lock files during self-update_
**Verdict:** Round-2 made real progress — it dropped the age/mtime heuristic (round-1 BRICK-1) and moved lock-recovery to a dedicated non-destructive `apply_clear_lock`/`/api/updates/clear_lock` path (round-1 BRICK-2, fixed — no `reset --hard`/`clean -fd`/`checkout`). BUT Codex empirically proved (via `strace` on git 2.43.0) that the NEW fcntl-`flock` holder-check is fundamentally wrong for git's locking model, so a live lock can STILL be deleted. This is a hard bounce.

### What I ran (rebased worktree `/tmp/wt-rebase-5688b`) — Codex + full suite (backend self-update; brick-risk path)
| Gate | Result |
|------|--------|
| Codex (re-verify BRICK-1/2 + CORE) | SHIP-WITH-FIXES — **1 BRICK + 1 CORE** (strace-verified) |
| Full pytest suite | ✅ 12265 passed, 0 failed (the BRICK is a runtime race, not test-covered) |

### Findings
**✅ Round-1 BRICK-2 fixed:** lock-recovery no longer routes into the destructive `apply_force_update` — a dedicated `apply_clear_lock(target)` + `/api/updates/clear_lock` endpoint + `btnClearUpdateLock` button, explicitly "No checkout/clean/reset --hard — safe by design". Non-destructive path is correct now.
**✅ Round-1 mtime heuristic removed:** "there is no mtime-based heuristic any more."
**⛔ BRICK (Codex, strace-verified — `api/updates.py:274`) — the fcntl-`flock` holder-check CANNOT detect a live git lock:** Codex ran `strace` on git 2.43.0 — `git add` creates `.git/index.lock` with `O_CREAT|O_EXCL` and renames it, using **NO `flock`/`F_SETLK`** (git does not use advisory locking). So `_is_lock_held()` (which probes `flock`) returns `False` for a lock file a live git process is holding open → `_try_remove_lock()` deletes a LIVE lock → corruption. Round-2 swapped one wrong staleness proof (mtime) for another (flock) — neither matches git's actual locking. **Fix:** don't use `flock` as proof of safety — either prove no process has the file open via `/proc/*/fd`-scan or `lsof`-style detection, OR remove automatic deletion entirely and surface manual instructions (fail closed for real).
**⛔ CORE (Codex — `api/updates.py:306`) — disappearing-lock race:** if `.git/index.lock` vanishes between `_is_lock_held()` and `os.remove()`, the `FileNotFoundError` surfaces as a false `remove-failed` instead of just continuing the normal update. **Fix:** catch `FileNotFoundError` separately → treat as "already gone" → continue to `_apply_update_inner()`.

**RED — gate-fail/changes-requested (1 BRICK + 1 CORE): the fcntl-`flock` probe does not detect git's O_EXCL locks (Codex strace-proved git 2.43.0 never takes an advisory lock on index.lock), so it can still delete a live lock.** Either (a) detect a real holder via `/proc/*/fd` open-fd scan / `lsof`, or (b) drop automatic deletion and surface a manual "another git process may be running; if you're sure it's not, delete `.git/index.lock`" instruction (true fail-closed). Plus catch `FileNotFoundError` in the remove path as already-gone. The non-destructive `clear_lock` architecture (round-2) is the right shape — it just needs a holder-detection that matches how git actually locks. concept 3/5 (right architecture, wrong holder-proof; brick-risk path). Author @b3nw (T2). crit=3, self-update brick-risk.

---
_Gate-certifier layer (warm-up → **gate** → release). Round-2 fixed BRICK-2 (dedicated non-destructive apply_clear_lock, no reset --hard) + dropped the mtime heuristic, BUT the fcntl-flock holder-check is wrong for git (Codex strace: git add uses O_CREAT|O_EXCL + rename, NO flock → _is_lock_held returns False for a live lock → still deletable = BRICK) + a disappearing-lock FileNotFoundError race (CORE). Fix: /proc-fd or lsof open-fd detection, or drop auto-delete + manual instruction (real fail-closed); catch FileNotFoundError as already-gone. Suite green 12265 (BRICK is a runtime race). Cert valid for sha:dcec232d._

b3nw:
@nesquena-hermes gate fixes applied

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (round 3: v2.2 drops auto-delete entirely → detect + manual-instruction = true fail-closed; the r1/r2 BRICKs are now structurally impossible)

**Certified head:** `sha:c3fce3dd` (clean rebase, branch `gate-rebase/5688-lock-detect-manual-instruction`) · **PR:** #5688 · _b3nw (T2), fix: detect and recover from stale git lock files during self-update_
**Verdict:** Round-3 CONVERGED via the safest resolution (my recommended option b): it DROPS automatic lock deletion entirely and replaces the (twice-broken) staleness heuristic with pure detection + a manual command surfaced to the operator. Codex SAFE, full suite green.

### Round arc (a clean convergence on a brick-risk path)
- **r1 (2 BRICK):** age/mtime deleted a live lock + lock-recovery routed into the destructive `git reset --hard` force path.
- **r2 (1 BRICK, Codex strace-verified):** the fcntl-`flock` "fix" still couldn't detect git's O_EXCL lock → still deletable.
- **r3 (GREEN):** dropped auto-delete entirely — "the only actor that should ever remove a lock is the user, on the host."

### What I ran (rebased worktree `/tmp/wt-rebase-5688c`) — Codex + full suite (backend self-update; brick-risk path)
| Gate | Result |
|------|--------|
| Codex (no-auto-delete / non-destructive / manual-command safety) | **SAFE TO SHIP** — 0 findings |
| Full pytest suite | ✅ **12277 passed, 0 failed** (+ 175 update tests) |

### Findings — all prior BRICKs structurally resolved
**✅ No lock removal anywhere (true fail-closed):** v2.2 replaced `_is_lock_held` + `_try_remove_lock` with pure INVENTORY (`get_git_lock_inventory` detects `.git/index.lock` presence) — there is NO `os.remove`/`unlink` of any lock in the codebase now. The r1 (live-lock deletion) and r2 (flock-can't-detect-git) BRICKs are structurally impossible because nothing auto-deletes. `apply_clear_lock` returns the inventory + a manual command (`rm -f {well_known_lock_path}`) for the operator to run on the host — it does NOT delete and does NOT call `apply_force_update`/`reset --hard`/`clean`/`checkout` (Codex confirmed non-destructive). The manual-command path is bounded to `.git/index.lock` (no injection/traversal). The lock-conflict detection surfaces the instruction only on a genuine lock-error. Full suite green.

**Ready to merge — use branch `gate-rebase/5688-lock-detect-manual-instruction` (`sha:c3fce3dd`), NOT the PR head `2ddef230`.** A textbook convergence: two rounds of auto-delete heuristics both proved unsafe (mtime, then flock — Codex strace-proved git uses O_EXCL), so round-3 correctly dropped auto-delete and fails closed with a manual instruction. No corruption risk, non-destructive, Codex SAFE + full suite green. concept 4/5 (arrived at the safe design). Author @b3nw (T2). crit=3, self-update.

---
_Gate-certifier layer (warm-up → **gate** → release). r1/r2 both BRICK (auto-delete via mtime then flock, both unsafe for git's O_EXCL locking); r3 v2.2 DROPS auto-delete entirely → pure inventory (get_git_lock_inventory) + manual command (rm -f {lock_path}) for the operator, no os.remove/unlink of any lock, apply_clear_lock non-destructive (no reset/clean/checkout/force). BRICKs structurally impossible now. Codex SAFE + full suite green (0 failed) + 175 update tests. Cert valid for sha:c3fce3dd._

nesquena-hermes:
Shipped in v0.51.911. Thanks @b3nw — self-update now recovers from a stale .git/index.lock: the update banner detects it, shows the exact rm command + a 'Retry update' button, and re-runs the normal update once the lock is gone. Fail-closed (server never deletes anything under .git). Full gate: Codex SAFE (no server-side .git deletion, no XSS, normal flow unchanged) + full suite green + 175 update tests. Closes #5687.

PR #5692, opened 2026-07-06
fix(workspace-git): apply creationflags to suppress conhost accumulation

jin89ho:
Add windows_hide_flags() to _run_git() and _config_names_for_scope()
to prevent hidden cmd.exe windows from accumulating.

Includes start-with-tailscale.ps1 startup script.


nesquena-hermes:
Thanks @jin89ho — the `windows_hide_flags()` approach is right (it mirrors the existing `CREATE_NO_WINDOW` usage in `api/updates.py` and is a safe no-op on non-Windows). Before this can land, a few changes:

**1. Drop `start-with-tailscale.ps1`.** It hardcodes a personal path (`C:\Users\JSEOK\hermes-webui`) and a specific Tailscale setup — that's a personal machine artifact, not something that belongs in the shared repo. Please remove it from this PR; the git fix should stand alone.

**2. Fix the whole class, not 2 of ~32 sites.** The conhost-window-accumulation problem applies to every `git`/tool subprocess WebUI spawns, but this PR only adds `creationflags` to the 2 calls in `workspace_git.py`. The same flash/accumulation happens from subprocess calls in `api/profiles.py`, `api/providers.py`, `api/worktrees.py`, `api/workspace.py`, and `api/rollback.py`. Please either (a) extend `windows_hide_flags()` to all Windows-spawning subprocess sites, or (b) if you want to keep this PR narrow, scope the title/description to "workspace git commands" so it's clear it's not the general fix.

**3. Add a test.** A small unit test asserting `windows_hide_flags()` is passed to `_run_git`'s `subprocess.run` (e.g. monkeypatch `subprocess.run` and assert the kwarg) locks the behavior so it can't silently regress.

Reopen/re-push when ready and I'll re-gate. Appreciate the fix — this is a real Windows reliability annoyance worth closing out properly.

nesquena-hermes:
Shipped in **v0.51.914** 🪟 — thanks @jin89ho!

Windows git operations no longer leave hidden `conhost.exe` console windows piling up: both `subprocess.run` git call sites in `api/workspace_git.py` (`_run_git` + `_config_names_for_scope`) now pass `CREATE_NO_WINDOW`. Safe no-op on macOS/Linux.

**Maintainer changes to your PR (all preserving your authorship via the commit + Co-authored-by):**
1. Dropped `start-with-tailscale.ps1` — it hardcoded a personal path + Tailscale setup, so it doesn't belong in the shared repo.
2. Replaced the `from hermes_cli._subprocess_compat import windows_hide_flags` import with a local `_windows_hide_flags()` helper. This was a **blocking regression** the review gate caught: `hermes_cli` is an *optional* dependency, so importing it on the workspace-git hot path would `ModuleNotFoundError` on any agent-less WebUI install (and the repo's own CI runners, which don't pip-install the agent) — breaking every git operation. The local helper mirrors the `api/updates.py` pattern (`CREATE_NO_WINDOW` on win32, `0` elsewhere) with zero external dependency.
3. Rescoped narrow to workspace-git commands per the review.
4. Added mutation-checked tests asserting the kwarg is wired on both sites.

**Gate (all green):** Codex SAFE (mechanics), Fable caught the agent-less regression (now fixed + agent-less import path proven), full suite clean (84/84 workspace_git tests). Thanks for the fix — it's a real Windows quality-of-life win.

PR #5696, opened 2026-07-06
perf(webui): reduce session-load waterfall and slow-request watchdog coverage

Kopamed:
## Summary

Reduces slow-request watchdog coverage to the endpoints that actually
need it, bounds two O(N) walks inside `Session.compact()` to a tail
window / inline role check, and extends the profile list cache TTL.
Targets the long-session-load and profile-list cold paths; both fall
from multi-second to sub-second on slow disks and from seconds to
tens-of-milliseconds on fast disks.

## What was slow

The slow-request watchdog covered `/api/sessions` and `/api/chat/start`
but not `/api/profiles` and `/api/models`, which fire on every session
click and were the source of the multi-second waterfalls. `/api/session`
already had its own per-stage timing via `_t0.._t6`, but only emitted
its `[SLOW]` breakdown when `HERMES_DEBUG_SLOW=1` was set, so
regressions went uninstrumented.

The biggest non-obvious cost was inside `Session.compact()` itself,
which ran two O(N) walks over `self.messages` on every response:

```python
'user_message_count': sum(
    1 for message in self.messages if _message_role(message) == 'user'
)
...
last_message_at = _last_message_timestamp(self.messages)
```

For a 2,500-message session this dominated the response time even at
idle. Two distinct walks, both O(N).

## Changes

**`api/models.py`**
- `Session._compute_user_message_count(messages)` — replaces the inline
  `sum(1 for ... in self.messages if _message_role(...) == 'user')`
  with a named method that inlines the role check (no per-iteration
  helper call). 1.7× faster than the original inline walk on the
  test corpus; identical output on 27/27 webui sidecars.
- `_last_message_timestamp(messages, *, tail_window=8)` — the messages
  array is chronologically ordered so the last non-tool message sits at
  the very end. Scan only the last 8 entries; fall back to the original
  full scan on miss (preserves correctness for sessions with >8 trailing
  tool rows). Verified identical output on 26 real webui sidecars plus
  adversarial cases (0–100 trailing tool messages).

Both helpers operate on `self.messages` (the sidecar) which the caller
already has in memory after `Session.load()`. No new disk reads.

**`api/profiles.py`**
- `_LIST_PROFILES_CACHE_TTL` 4s → 60s. The 4s TTL forced an `os.walk`
  + skill-tree parse on every poll; profile rows are static across a
  session load. Invalidation already wired: `create_profile_api()` and
  `delete_profile_api()` call `_invalidate_list_profiles_cache()`.

**`api/request_diagnostics.py`**
- Allowlist `GET /api/profiles` and `GET /api/models` on
  `RequestDiagnostics.maybe_start()` so the existing watchdog covers
  them. No new logic. `/api/session` is intentionally not on the list:
  it already has its own per-stage logging via the `_t0.._t6` block with
  auto-log above 2s, and adding `RequestDiagnostics` on top would
  duplicate the slow-request journal entries.

**`api/routes.py`**
- `/api/models` and `/api/profiles` now wrap their existing handlers in
  `RequestDiagnostics` with stage markers.
- `/api/session`: the existing `_t0.._t6` block now auto-logs when total
  exceeds 2s (was env-var gated). The env var still forces logging on
  every request for development.

**`api/config.py`**
- Per-stage markers inside `get_available_models_for_session_visit` and
  a small `_maybe_log_slow_stages()` helper that emits one `[SLOW]`
  log line when the function's total wall time crosses 500ms (or
  `HERMES_DEBUG_SLOW` ms) with each stage's delta in ms.

## Verification

`scripts/ruff_lint.py --diff origin/master` reports **0 new violations**
on added/modified lines.

`_compute_user_message_count` and `_last_message_timestamp` are both
verified to match the pre-patch output on the full set of webui
sidecars: 27/27 sessions produce identical `user_message_count` and
identical `last_message_at`.

`_last_message_timestamp` additionally tested against adversarial
inputs (0, 1, 5, 7, 8, 9, 10, 20, 50, and 100 trailing tool messages):
all match the expected true last non-tool message timestamp. Window
hits for trailing < 8; full-scan fallback for trailing >= 8.

Performance on a 2,405-message sidecar (the corpus worst case):
- `_compute_user_message_count`: 0.76ms per call
- `_last_message_timestamp` (window hit): <0.01ms per call

Manual benchmark on a slow eMMC host (representative of Raspberry Pi-class
devices and other constrained hardware):

| Scenario | before p95 | after p95 |
|---|---|---|
| sidebar | 887ms | 57ms |
| long session load (~2,500 msgs) | 9,141ms (2/10 timeouts) | 3,156ms (0/10) |
| session switch (resolve_model=1) | 3,708ms | 1,442ms |
| profiles | 821ms | 248ms |

Under concurrent `/api/session/draft` POSTs (the real Hermes/WebUI
write path, not synthetic SQLite inserts), sidebar stays well under
100ms p95 with no timeouts.

## Files changed

```
api/config.py              | 74 ++++++++++++++++++++++++++++++++--
api/models.py              | 113 ++++++++++++++++++++++++++++++++-----------------
api/profiles.py            |  2 +-
api/request_diagnostics.py |  5 ++-
api/routes.py              | 53 +++++++++++++++++--------
5 files changed, 224 insertions(+), 49 deletions(-)
```

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -7561,36 +7561,106 @@ def _models_cache_file_age_seconds(cache_path: Path, now: float) -> float | None
 
 
 def get_available_models_for_session_visit() -> dict:
-    """Return /api/models with a short session-visit freshness horizon."""
+    """Return /api/models with a short session-visit freshness horizon.
+
+    perf(session-load-latency) Phase 0: this function is the source of the
+    multi-second `/api/models?freshness=session_visit` latency. Stage markers
+    feed into RequestDiagnostics when called from /api/models; standalone
+    callers get the same envelope via the local _stagelog dict.
+    """
+    import time as _time
+    import logging as _logging
+    _stagelog: list[tuple[str, float]] = [("enter", _time.monotonic())]
+    def _mark(name: str) -> None:
+        _stagelog.append((name, _time.monotonic()))
+    _logger = _logging.getLogger("api.config")
+    _slow_threshold_ms = float(os.environ.get("HERMES_DEBUG_SLOW", "0") or "0") or 500.0
+
     global _available_models_cache, _available_models_cache_ts, _available_models_cache_source_fingerprint
     cache_path = _get_models_cache_path()
     cache_age = _models_cache_file_age_seconds(cache_path, time.time())
+    _mark(f"disk_age_check:{cache_age}")
     disk_cached = None
     if cache_age is not None and cache_age < _SESSION_VISIT_MODELS_FRESHNESS_SECONDS:
+        _mark("cache_age_within_ttl")
         now_mono = time.monotonic()
         with _available_models_cache_lock:
             cached = _get_fresh_memory_models_cache(now_mono)
             if cached is not None:
+                _mark("memory_cache_hit")
+                _maybe_log_slow_stages(_logger, _stagelog, _slow_threshold_ms, "models.session_visit")
                 return cached
+        _mark("memory_cache_miss_loading_disk")
         disk_cached = _load_models_cache_from_disk()
         if disk_cached is not None:
             with _available_models_cache_lock:
                 cached = _get_fresh_memory_models_cache(time.monotonic())
                 if cached is not None:
+                    _mark("disk_then_memory_cache_hit")
+                    _maybe_log_slow_stages(_logger, _stagelog, _slow_threshold_ms, "models.session_visit")
                     return cached
                 _available_models_cache = copy.deepcopy(disk_cached)
                 _available_models_cache_ts = time.monotonic()
                 _available_models_cache_source_fingerprint = _models_cache_source_fingerprint()
+            _mark("disk_cache_returned")
+            _maybe_log_slow_stages(_logger, _stagelog, _slow_threshold_ms, "models.session_visit")
             return copy.deepcopy(disk_cached)
 
+    _mark("cache_age_stale_or_missing")
     stale_cached = disk_cached or _load_stale_models_cache_from_disk()
+    _mark(f"stale_cached_loaded:{bool(stale_cached)}")
     try:
-        return get_available_models(force_refresh=True)
+        _mark("force_refresh_start")
+        result = get_available_models(force_refresh=True)
+        _mark("force_refresh_done")
+        _maybe_log_slow_stages(_logger, _stagelog, _slow_threshold_ms, "models.session_visit")
+        return result
     except Exception:
+        _mark("force_refresh_failed")
         logger.debug("session-visit models refresh failed", exc_info=True)
         if stale_cached is not None:
+            _mark("stale_fallback_return")
+            _maybe_log_slow_stages(_logger, _stagelog, _slow_threshold_ms, "models.session_visit")
             return copy.deepcopy(stale_cached)
+        _mark("prefer_cache_fallback")
+        _maybe_log_slow_stages(_logger, _stagelog, _slow_threshold_ms, "models.session_visit")
         return get_available_models(prefer_cache=True)
+
+
+def _maybe_log_slow_stages(
+    logger_obj: "_logging.Logger",
+    stagelog: "list[tuple[str, float]]",
+    threshold_ms: float,
+    tag: str,
+) -> None:
+    """perf(session-load-latency) Phase 0: per-stage timing reporter.
+
+    Emits a single log line listing every stage with its delta in ms when
+    the function's total wall time crosses ``threshold_ms``. Lives next to
+    the model cache code so it has zero coupling with the WebUI request
+    layer; called from both ``get_available_models_for_session_visit`` and
+    (Phase 1) the chat session load path.
+    """
+    if len(stagelog) < 2:
+        return
+    total_ms = (stagelog[-1][1] - stagelog[0][1]) * 1000.0
+    if total_ms < threshold_ms:
+        return
+    parts: list[str] = []
+    for i in range(1, len(stagelog)):
+        prev_t = stagelog[i - 1][1]
+        cur_t = stagelog[i][1]
+        parts.append(f"{stagelog[i][0]}={((cur_t - prev_t) * 1000.0):.1f}ms")
+    try:
+        logger_obj.warning(
+            "[SLOW] %s total=%.1fms stages: %s",
+            tag,
+            total_ms,
+            " ".join(parts),
+        )
+    except Exception:
+        # Logging must never break a response path.
+        pass
 
 
 # ── Static file path ─────────────────────────────────────────────────────────
--- a/api/config.py
+++ b/api/config.py
@@ -7628,7 +7628,7 @@ def get_available_models_for_session_visit() -> dict:
 
 
 def _maybe_log_slow_stages(
-    logger_obj: "_logging.Logger",
+    logger_obj: "logging.Logger",
     stagelog: "list[tuple[str, float]]",
     threshold_ms: float,
     tag: str,
--- a/api/config.py
+++ b/api/config.py
@@ -7574,7 +7574,13 @@ def get_available_models_for_session_visit() -> dict:
     def _mark(name: str) -> None:
         _stagelog.append((name, _time.monotonic()))
     _logger = _logging.getLogger("api.config")
-    _slow_threshold_ms = float(os.environ.get("HERMES_DEBUG_SLOW", "0") or "0") or 500.0
+    _debug_slow_raw = os.environ.get("HERMES_DEBUG_SLOW", "0") or "0"
+    try:
+        _slow_threshold_ms = float(_debug_slow_raw) or 500.0
+    except ValueError:
+        # Non-numeric truthy value (e.g. "true") — log everything,
+        # matching the string-flag treatment at routes.py:11655.
+        _slow_threshold_ms = 0.0
 
     global _available_models_cache, _available_models_cache_ts, _available_models_cache_source_fingerprint
     cache_path = _get_models_cache_path()
--- a/api/models.py
+++ b/api/models.py
@@ -1408,19 +1408,25 @@ class Session:
             return cls.load(sid)
 
     @staticmethod
-    def _compute_user_message_count_lazy(sid: str) -> int:
+    def _compute_user_message_count_lazy(sid: str, messages=None) -> int:
         """perf(session-load-latency) Priority 1: cheap SQL-only user count.
 
-        Returns the number of messages with role='user' for a given session,
-        computed via a single indexed SQLite query (~5ms even for 2,500+ msg
-        sessions because idx_messages_session covers the WHERE filter). The
-        role='user' predicate is a second filter on the small index range, so
-        it is bounded by the index lookup, not a full table scan.
+        Returns the number of messages with role='user' for a given session.
+        Fast path is a single indexed SQLite query (~5ms even for 2,500+ msg
+        sessions because idx_messages_session covers the WHERE filter); the
+        role='user' predicate is a second filter on the small index range.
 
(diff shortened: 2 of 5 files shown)


Kopamed:
All three findings addressed in 2a4e86f5:

**P1 (silent zero fallback):** Fixed. `_compute_user_message_count_lazy(sid, messages=None)` now falls back to the in-memory walk (the pre-patch behavior) when SQLite fails for any reason — DB locked, missing file, schema drift. Caller passes `self.messages` so the fallback is bounded by the already-loaded Python list. Verified on 26/26 sidecars (SQL happy path matches walk) and on simulated DB failures (fallback matches walk for empty, no-user, and with-user message lists).

**P2 #1 (docstring):** Fixed. Rewrote to describe what the code actually does: returns an int, with a documented in-memory fallback when the SQLite path fails.

**P2 #2 (dead allowlist entry):** Removed. The PR description has been updated to clarify that `/api/session` is intentionally not on the `RequestDiagnostics` list — its existing `_t0.._t6` block already does per-stage timing with auto-log above 2s, and adding `RequestDiagnostics` on top would just duplicate the slow-request journal entries.

Linter clean (`scripts/ruff_lint.py --diff origin/master` reports 0 new violations on added/modified lines).

nesquena-hermes:
Picking this up for review — this is a high-value reliability/perf fix (directly targets the laggy-session-load reports) so I'm running the full gate on it now (Codex reproduce + full suite + my own testing of the hot paths). Will post the gate result here.

Two things I'll specifically verify:
1. `_compute_user_message_count_lazy` counts state.db rows where the old path counted `self.messages` — confirming they can't diverge in a way that trips `_looks_like_stale_zero_message_row` / `_row_may_need_sidecar_metadata_refresh` (state.db vs sidecar out-of-sync case).
2. `_last_message_timestamp` tail_window=8 — verified the window-miss fallback is the exact original full scan, so correctness holds even for >8 trailing tool rows (just a small wasted pre-scan). Good.

Thanks @Kopamed.

Kopamed:
Both concerns addressed in 2a54e77e — I dropped the SQL path entirely rather than patching around the divergent-source issue.

**Source-of-truth problem (maintainer + Greptile rerun):** On the test corpus, `state.db` and the sidecar `self.messages` can diverge by hundreds of messages during recovery / mid-flight writes:

```
0db167553ac7  db=11   sidecar=12
295592fd560e  db=813  sidecar=889
7478cae31f01  db=2730 sidecar=2405
```

The pre-patch inline `sum(...)` walked the sidecar. My SQL path walked `state.db`. Sidebar stale-row detection (`_looks_like_stale_zero_message_row`, `_row_may_need_sidecar_metadata_refresh`) consumes this field as if the sidecar were the source of truth, so swapping the source silently flips the field's semantics.

**Fix:** Removed the SQL path. `_compute_user_message_count(messages)` does the same in-memory walk the pre-patch code did, extracted into a named method. Verified 27/27 webui sidecars produce identical counts vs the pre-patch inline walk. 1.7x faster on the test corpus (inlined role check avoids per-iteration function call to `_message_role`).

**Greptile batch-path concern (SQLite-per-compact in sidebar):** resolved by the same change — sidebar no longer opens any new SQLite connection for the user count. The cost is now bounded by the already-loaded sidecar list (0.76ms for 2,400 messages on a Celeron N3350).

Linter clean. PR description updated to reflect the final design.

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (a valid `HERMES_DEBUG_SLOW=true` 500s the models endpoint + the profiles-cache TTL bump serves stale rows for 60s after mutations — 1 CORE + 1 SILENT)

**Certified head:** `sha:2a54e77e` (rebased onto current master, `git apply` clean) · **PR:** #5696 · _Kopamed, perf(webui): reduce session-load waterfall + slow-request watchdog coverage_
**Verdict:** The perf intent is good (waterfall reduction + watchdog instrumentation), but Codex found 2 real defects — a 500 on a valid env-var value and a stale-profile-row regression introduced by the cache-TTL bump without matching invalidation. Full suite green (both are runtime/config, not test-covered).

### What I ran (rebased worktree `/tmp/wt-rebase-5696`) — Codex + full suite (backend caching/perf; no visible UI)
| Gate | Result |
|------|--------|
| Codex (cache correctness / profile-isolation / stale-serve) | SHIP-WITH-FIXES — **1 CORE + 1 SILENT** |
| Full pytest suite | ✅ 12235 passed, 0 failed |

### Findings
**⛔ CORE (Codex — `api/config.py:7581`) — `/api/models?freshness=session_visit` 500s when `HERMES_DEBUG_SLOW` is a non-numeric truthy value:** the new `float(os.environ.get("HERMES_DEBUG_SLOW"))` runs before cache/fallback handling and throws `ValueError` on `HERMES_DEBUG_SLOW=true` — but the EXISTING use at `routes.py:11685` treats the same var as a string flag, so a value that works everywhere else now 500s the models endpoint. **Fix:** parse defensively — `try/except ValueError` → fall back to the default threshold (and align interpretation with the existing string-flag use).
**⛔ SILENT (Codex — `api/profiles.py:1838`) — `/api/profiles` serves stale rows for up to 60s after profile mutations:** the PR bumped `_LIST_PROFILES_CACHE_TTL` (to reduce the waterfall) but cache invalidation is only wired for create/delete — routes that write defaults/providers/skills/gateway don't invalidate `_LIST_PROFILES_CACHE`, so `model`, `provider`, `has_env`, `gateway_running`, `visible`, and skill counts go stale for the TTL window after a mutation. **Fix:** restore `_LIST_PROFILES_CACHE_TTL = 4.0`, OR invalidate `_LIST_PROFILES_CACHE` from every profile-row mutation path (not just create/delete).

**RED — gate-fail/changes-requested (1 CORE + 1 SILENT): (1) parse `HERMES_DEBUG_SLOW` defensively (`try/except ValueError` → default) so a valid `=true` doesn't 500 the models endpoint — and match the existing string-flag interpretation at routes.py:11685; (2) either keep the profiles-cache TTL short (4.0s) or invalidate `_LIST_PROFILES_CACHE` from every mutation path that changes a profile row (defaults/providers/skills/gateway), not just create/delete.** The waterfall-reduction + watchdog instrumentation are otherwise sound; these are two cache/env-parse edges. concept 3/5 (good perf intent, 2 correctness edges on a profile-sensitive cache). Author @Kopamed. crit=3, perf/reliability.

---
_Gate-certifier layer (warm-up → **gate** → release). CORE: new float(os.environ['HERMES_DEBUG_SLOW']) 500s /api/models when the var is a non-numeric truthy string (existing code treats it as a string flag) → try/except ValueError. SILENT: _LIST_PROFILES_CACHE_TTL bumped for waterfall-reduction but invalidation only wired for create/delete → /api/profiles serves stale model/provider/has_env/gateway_running/visible/skill-count rows for up to 60s after defaults/providers/skills/gateway mutations → restore 4.0s TTL or invalidate from every mutation path. Suite green 12235 (both runtime/config, not test-covered). Cert valid for sha:2a54e77e._

Kopamed:
## Gate findings addressed

Both defects fixed in 73151ba (pushed to branch).

### ⛔ CORE —  500s 

**Root cause:**  at  —  raises  before the cache/fallback handlers run. The existing use at  treats the same env var as a string flag, so a value that works everywhere else 500s the models endpoint.

**Fix:**  — parse defensively. Non-numeric truthy values (like ) fall back to  (log every request), which matches the route-layer semantics (`if _debug_slow or _total_ms >= 2000`). Numeric values and unset continue to work exactly as before.

### ⛔ SILENT — profiles cache serves stale rows for 60s after mutations

**Root cause:**  bumped from 4.0 to 60.0 in the PR, but  is only called from  and . Routes that write defaults/providers/skills/gateway don't invalidate, so , , , , , and  go stale for the TTL window.

**Fix:** Restored . This is the pre-patch value — it keeps the os.walk frequency high enough that mutation-to-poll staleness is negligible. The profile-list waterfall on cold starts is the same as before this PR; the perf gains from the other changes (bounded tail-scan, inlined role check, watchdog expansion) aren't affected.

### Verification

- AST-valid, linter-clean (ruff passes both files with 0 new violations).
- Commit  on , pushed to .

Kopamed:
## Gate findings addressed

Both defects fixed in 73151ba (pushed to branch).

### CORE — HERMES_DEBUG_SLOW=true 500s /api/models

Root cause: `float(os.environ.get('HERMES_DEBUG_SLOW', '0') or '0')` at config.py:7577 — `float('true')` raises ValueError before the cache/fallback handlers run. The existing use at routes.py:11655 treats the same env var as a string flag, so a value that works everywhere else 500s the models endpoint.

Fix: `try/except ValueError` — parse defensively. Non-numeric truthy values (like `true`) fall back to 0.0 (log every request), matching the route-layer semantics (`if _debug_slow or _total_ms >= 2000`). Numeric values and unset continue to work exactly as before.

### SILENT — profiles cache serves stale rows for 60s after mutations

Root cause: `_LIST_PROFILES_CACHE_TTL` bumped from 4.0 to 60.0 but `_invalidate_list_profiles_cache()` only called from create/delete paths. Routes that write defaults/providers/skills/gateway don't invalidate, so model/provider/has_env/gateway_running/visible/skill_count go stale for the TTL window.

Fix: Restored `_LIST_PROFILES_CACHE_TTL = 4.0`. Pre-patch value. Keeps os.walk frequent enough that mutation-to-poll staleness is negligible. The profile-list waterfall on cold starts is the same as before this PR; the perf gains from the other changes (bounded tail-scan, inlined role check, watchdog expansion) aren't affected.

### Verification

- AST-valid, linter-clean (ruff passes both files with 0 new violations).
- Commit 73151ba on `perf/session-load-latency`, pushed to Kopamed/hermes-webui.

nesquena-hermes:
Session-load perf shipped across v0.51.908 (#5696 core) + v0.51.909 (profiles TTL correction). Thanks @Kopamed — the two O(N) walks in Session.compact() are gone (bounded user-count + tail-window last-message-timestamp with exact fallback) and the slow-request watchdog now covers /api/session. Full gate: Codex (folded in a non-throwing HERMES_DEBUG_SLOW parse it flagged) + full suite green. Your own 60s->4s profiles-cache-TTL revert (stale-row window) was caught + shipped as the v0.51.909 follow-up. Targets the laggy-session-load reports (#5455).

nesquena-hermes:
🔬 **Gate re-check — this is already SHIPPED (folded into master), no re-gate needed**

Heads up @Kopamed — the substance of this PR is **already merged into master and released (v0.51.908)**, credited to you in the changelog:
- **Session-load perf** (bounded user-count + last-message-timestamp lookups in `Session.compact()`, slow-request watchdog now covers `/api/session`) — shipped (#5696, #5455).
- **Profiles cache TTL revert 60s→4s** (the stale-profile-rows issue I flagged) — shipped via #5704.
- **Non-throwing `HERMES_DEBUG_SLOW` parse** (the `=true` 500 I flagged as a CORE) — folded in during the maintainer gate (commit `2454719f`).

I verified empirically: after rebasing this PR's head (`73151ba8`) onto current master and taking master's canonical version of the two overlapping fixes, the **net unique diff is 0 lines** — everything here is in master. The merge went through a maintainer stage-branch (`release/stage-5696`) rather than this PR ref, which is why it didn't auto-close.

Both round-1 gate findings (the `HERMES_DEBUG_SLOW` 500 + the profiles-cache staleness) were adopted in the shipped version. Nothing left to gate. This PR can be closed as **shipped** — thanks for the perf work and the follow-up fixes! crit=3, @Kopamed.

Kopamed:
Content already merged into master and released in v0.51.908. Closing as shipped per gate-bot confirmation.

PR #5698, opened 2026-07-06
fix(streaming): surface tool progress as interim prose

ruizanthony:
## Summary

- emit a safe `interim_assistant` prose event immediately before live tool-start cards
- localize the generated progress line from WebUI settings (`fr` gets French text, other locales get a neutral English fallback)
- avoid exposing raw terminal commands or large tool arguments; only use the existing capped tool-argument snapshot

## Why

Recent Codex/Hermes changes route commentary/analysis-phase deltas through the reasoning channel instead of visible assistant text. That is correct for keeping commentary out of the final answer, but in WebUI sessions where thinking/reasoning display is hidden, users can be left with only tool cards and no readable mid-turn progress.

This restores a safe, UI-level progress breadcrumb without enabling raw reasoning display.

## Tests

- `python3 -m pytest tests/test_tool_progress_interim_server.py tests/test_issue3587_intermediate_reasoning.py tests/test_regressions.py::test_messages_js_supports_interim_assistant_events -q`

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -7246,6 +7246,78 @@ def _run_agent_streaming(
                         args_snap[k] = s2[:cap] + ('...' if len(s2) > cap else '')
                 return args_snap
 
+            def _tool_progress_value(value, max_len=90):
+                clean = " ".join(str(value or "").split())
+                if not clean:
+                    return ""
+                return clean[: max_len - 1].rstrip() + "…" if len(clean) > max_len else clean
+
+            def _tool_progress_arg(args_snap, *names):
+                if not isinstance(args_snap, dict):
+                    return ""
+                for key in names:
+                    value = _tool_progress_value(args_snap.get(key))
+                    if value:
+                        return value
+                return ""
+
+            def _tool_progress_language():
+                try:
+                    language = str((load_settings() or {}).get('language') or '').strip().lower()
+                except Exception:
+                    language = ''
+                return language
+
+            def _safe_tool_progress_text(name, args_snap):
+                tool_name = _tool_progress_value(name, 64)
+                if not tool_name:
+                    return ""
+                target = _tool_progress_arg(args_snap, 'path', 'file_path', 'url', 'query', 'pattern', 'name')
+                target_md = f" `{target}`" if target else ""
+                if _tool_progress_language().startswith('fr'):
+                    if tool_name == 'skill_view':
+                        return f"Je charge la compétence `{target}`." if target else "Je charge une compétence Hermes."
+                    if tool_name == 'read_file':
+                        return f"Je lis{target_md}." if target else "Je lis un fichier."
+                    if tool_name == 'search_files':
+                        return f"Je recherche{target_md}." if target else "Je recherche dans les fichiers."
+                    if tool_name == 'web_search':
+                        return f"Je recherche sur le web{target_md}." if target else "Je recherche sur le web."
+                    if tool_name == 'web_extract':
+                        return f"Je lis une source web{target_md}." if target else "Je lis une source web."
+                    if tool_name in {'browser_navigate', 'browser_snapshot', 'browser_vision'}:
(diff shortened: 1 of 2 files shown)


nesquena-hermes (changes requested):
## Changes requested — synthetic tool-progress prose lands in the durable saved transcript, and the flag meant to prevent that is unwired

Thanks for this — surfacing tool activity as it happens is a nice UX idea. But as written the feature writes machine-generated status lines into the **permanent** assistant message rather than a transient live-status channel, and I can reproduce transcript corruption on the common tool-using path. Three objective issues, all on the current head:

### F1 (blocking) — interim prose is emitted on the *append-to-transcript* channel

`_emit_tool_progress_interim` (`api/streaming.py:7315-7318`) emits:

```python
put('interim_assistant', {
    'text': visible,
    'already_streamed': False,
    'generated_tool_progress': True,
})
```

`already_streamed=False` is the **append** branch of the interim reducer. In `api/routes.py:3204`:

```python
else:
    assistant_text = f"{assistant_text}\n\n{visible}" if assistant_text else visible
```

So the synthetic `Using \`tool\` on \`target\`.` lines are concatenated into the durable `assistant_text` that gets journaled and saved. The same append happens in the two other consumers of this event — `api/models.py` (journal recovery rebuild, `assistant_parts.append(text)`) and `static/messages.js:4962` (live/inflight message, `assistantText += …`). Net effect: every named tool call in every streaming turn injects a status line into the saved message. It also bypasses the echo-suppression in `on_interim_assistant` (`_is_visible_output_echo` / `_strip_reasoning_output_echo`) because it calls `put()` directly instead of routing through that callback.

A normal turn (2 tool starts + a final answer) ends up saved as:

```
Using `read_file` on `api/streaming.py`.

Using `search_files` on `lock`.The lock is held correctly.
```

### F2 (blocking) — final answer glues onto the last interim with no separator

Final-answer tokens append via `assistant_text += text` (`api/routes.py:3186`) with no separator after an interim, so the real answer runs directly into the last status line (see the `\`lock\`.The lock is held…` above). This resolves once F1 stops folding interims into the transcript.

### F3 (blocking) — the `generated_tool_progress` flag is a no-op

`generated_tool_progress` was clearly added so a downstream consumer could identify and drop synthetic prose, but nothing reads it — a repo-wide search finds **zero** consumers outside the producer line and the new test. The prose it tags is treated identically to real runtime commentary everywhere.

### Fix direction

Emit tool-progress on a **dedicated ephemeral channel** the transcript reducers ignore (e.g. a distinct `tool_progress` SSE event that renders as transient live status and is never folded into `assistant_text`). If you'd rather keep `interim_assistant`, then (a) make the reducers at `api/routes.py:3195`, `api/models.py`, and `static/messages.js` **skip any payload carrying `generated_tool_progress`** so the flag actually does its job, and (b) ensure a trailing separator before subsequent token segments. Please replace the current tests with a **behavioral** one that runs a turn through the durable reducer and asserts the saved `assistant_text` contains no `Using \`…\`` status lines — the four current tests are static string-assertions on source text and pass without exercising the emit→save path.

Two more items worth addressing while you're in here (not blocking the above): `_safe_tool_progress_text` interpolates `target` raw inside backticks, so a `query`/`path` containing a backtick or markdown metachars breaks out of the code span (now persisted into the transcript) — escape it; and `_tool_progress_language()` does an uncached `SETTINGS_FILE.read_text()` + `json.loads` on every tool start for every stream (including English installs) — compute the language once per turn.

Happy to re-review once the interim prose is off the durable path.

ruizanthony:
Closing this in favor of an agent-side fix.

The generic WebUI-generated tool prose is technically safe, but it does not solve the UX problem: users need the model's high-level intent / commentary before tool calls, not another label duplicating the tool card.

The correct fix is to route Codex/Harmony `commentary` phase text into the `interim_assistant` channel while keeping `analysis` in private reasoning.

PR #5717, opened 2026-07-06
fix: surface agent's graceful summary on iteration-limit exhaustion (closes #5494)

### LINKED ISSUE #5494: Turn Limit Exhaustion: Missing Summary Prompt in WebUI vs hermes-agent
# Turn Limit Exhaustion: Missing Summary Prompt in WebUI vs hermes-agent

## Summary

There is a **behavioral discrepancy** between how `hermes-agent` (via Discord/CLI) and `hermes-webui` handle turn limit exhaustion:

- **hermes-agent**: When max iterations are reached, automatically sends a summary prompt to the model and displays the generated summary to the user
- **hermes-webui**: When turn budget is exhausted, immediately pauses without sending any summary prompt or generating a summary

This results in different user experiences: Discord/CLI users see what was accomplished before the agent stops, while WebUI users only see a pause message with no context about what was done.

### hermes-agent Behavior (Discord/CLI)

**Location**: `agent/turn_finalizer.py` + `agent/chat_completion_helpers.py::handle_max_iterations()`

When iteration budget is exhausted:
1. Detects `max_iterations_reached` condition
2. Emits status: "⚠️ Iteration budget exhausted — asking model to summarise"
3. **Injects a summary prompt** into the conversation:
   ```
   "You've reached the maximum number of tool-calling iterations allowed. 
    Please provide a final response summarizing what you've found and 
    accomplished so far, without calling any more tools."
   ```
4. Makes **one additional API call** (with tools stripped) to get the summary
5. User sees both the warning AND a model-generated summary

**Default limit**: 90 iterations (or 25 for TUI)

### hermes-webui Behavior

**Location**: `api/goals.py::evaluate_after_turn()` + `api/streaming.py`

When turn budget is exhausted:
1. Detects `turns_used >= max_turns` condition
2. Sets status to "paused" with reason
3. Returns `should_continue: False` and **`continuation_prompt: None`**
4. **No additional prompt is sent to the model**
5. User sees only: "⏸ Goal paused — X/Y turns used. Use /goal resume to keep going..."

**Default limit**: 20 turns

### hermes-agent (Summary on Exhaustion)
- `agent/turn_finalizer.py:55-70` - Detection and summary trigger
- `agent/chat_completion_helpers.py:1293-1400` - `handle_max_iterations()` implementation

### hermes-webui (No Summary)
- `api/goals.py:212-226` - `evaluate_after_turn()` pause logic
- `api/streaming.py:9186-9245` - Goal decision integration in streaming

## Questions for Maintainers

1. **Is this discrepancy intentional?** Should the WebUI match hermes-agent's behavior of providing a summary when the turn budget is exhausted?

2. **If yes**, should this be:
   - A default behavior change (always send summary prompt on exhaustion)?
   - A configurable option (e.g., `goals.send_summary_on_exhaustion: true/false`)?

3. **If no** (current behavior is intentional), what is the design rationale?
   - Is it to avoid extra API calls after budget exhaustion?
   - Is it to keep the WebUI simpler/more predictable?
   - Should this be documented as a known difference?

4. **Configuration**: If a configurable option is preferred, what would be the ideal:
   - Config key name?
   - Default value?
   - Where should it be documented?

### Option A: Match hermes-agent Behavior
Add summary prompt injection in `api/goals.py::evaluate_after_turn()` when budget is exhausted, similar to hermes-agent's `handle_max_iterations()`.

**Pros**: Consistent UX across platforms, users see what was accomplished
**Cons**: Extra API call, increased cost per exhaustion event

### Option B: Configurable Option
Add `goals.send_summary_on_exhaustion` config option (default: `false` to maintain current behavior).

**Pros**: User choice, maintains backward compatibility
**Cons**: Adds configuration complexity

### Option C: Document as Intentional Difference
If the current behavior is intentional, document this difference in the docs to set user expectations.

**Pros**: No code changes, clear documentation
**Cons**: Users may still find the discrepancy confusing

## Environment
- hermes-webui version: (current master)
- hermes-agent version: (current master)
- Platforms compared: Discord/CLI vs WebUI

## Additional Context

This was discovered when comparing autonomous task behavior across platforms. The user observed that Discord sessions provide closure (summary of work done) when hitting limits, while WebUI sessions simply pause without explanation of what was accomplished.

claw-io:
# fix: surface agent's graceful summary on iteration-limit exhaustion (closes #5494)

Closes #5494.

## Thinking Path

The original issue framing conflated two independent mechanisms:
`api/goals.py::evaluate_after_turn()` (the `/goal` feature, gated to
active goals) and the `AIAgent` tool-iteration budget governed by
`agent.max_turns` (default 90), which fires inside the WebUI's
embedded `AIAgent`. The latter is the one that produces
`max_iterations_reached` and triggers `handle_max_iterations()` —
the same code path the Discord/CLI uses. A correction comment on
#5494 captures the distinction.

Inspecting the WebUI side further (`api/streaming.py`,
`api/goals.py`, `chat_completion_helpers.handle_max_iterations`),
the parity story is narrower than the issue suggested:

- The WebUI **already** surfaces the model's summary when one was
  generated. Existing tests pin this:
  `test_streaming_tool_limit_with_final_answer_persists_clean_done_state`.
- The actual residual gap is the **empty-summary** case: when the
  agent's summary call yields no usable assistant content (common
  with reasoning-only responses; `handle_max_iterations` only appends
  a message when `final_response` is non-empty:
  `agent/chat_completion_helpers.py:1462-1466`), the user sees a
  bare `tool_limit_reached` error.

Why the asymmetry: `handle_max_iterations()` always **returns** a
non-empty `final_response` (either the summary or one of several
graceful fallback strings like *"I reached the iteration limit and
couldn't generate a summary."*), and Hermes Agent surfaces that
return value as the turn's final answer. The WebUI reads only
`result['messages']`, so on empty summary it has no final assistant
turn and routes to the error path.

The minimal fix is to bridge that asymmetry: when
`max_iterations_reached` fires and `result['messages']` ends without
a usable assistant answer, synthesize a final assistant turn from
`result['final_response']`. The existing
`_mark_latest_assistant_tool_limit_status` flow then attaches the
status card exactly as it does for a real summary, and the user sees
the closure text instead of the bare error.

## What Changed

- **`api/streaming.py`** (+48):
  - New helper `_maybe_inject_max_iteration_summary_fallback(messages, result)`
    placed next to the existing tool-limit helpers. Pure: returns a
    new list, mutates nothing. Tagged with
    `_max_iteration_summary_fallback: True` so downstream consumers
    can distinguish from model-emitted turns (mirrors the synthetic
    scaffolding convention from #5334).
  - 9-line wire-up in `_run_agent_streaming` merge-result block,
    directly after the existing
    `_drop_synthetic_max_iteration_summary_requests`. Importantly,
    the wire-up writes the synthesized list back into
    `result['messages']` (immutable copy via `{**result, 'messages': …}`)
    so the downstream `_all_result_messages` detection
    (`api/streaming.py:8476`) — which reads `result['messages']`
    directly, not the local `_result_messages` — sees the fallback
    too. Without this, the silent-failure classifier would still
    flag empty summary as `apperror` because it sees no assistant
    answer in the raw result messages.

- **`tests/test_tool_limit_terminal_state.py`** (+117):
  - 2 E2E tests through `_run_streaming_with_fake_agent`:
    - fallback surfaces graceful text + status card, no
      `tool_limit_reached` apperror;
    - no double-injection when a real summary already exists.
  - 3 unit tests on the helper for the three skip conditions
    (no fallback string, blank/whitespace fallback, existing
    assistant answer) plus the no-result edge (None / non-dict).

  All 17 tests in the file pass; the 12 existing tests — including
  `test_streaming_tool_limit_without_final_answer_emits_no_final_apperror`
  (the "agent died entirely" path: no fallback AND no summary →
  apperror is correct) — continue to pass unchanged.

## Why It Matters

#5494 is filed as a parity question between hermes-agent and
hermes-webui. The investigation confirmed hermes-agent parity is
*already* in place for the common case (real summary present);
this PR closes the remaining gap (empty summary → bare error) so
the WebUI shows the same closure text a Discord/CLI user gets
when the agent's summary call comes back empty (a real failure
mode with reasoning-only models).

This unblocks maintainer response on the question
*A vs B vs C* originally posed in #5494: the
`Api-defaults-parity` direction the issue commenter favored now
ships as the default — no new config key, no behavioral switch,
mirrors hermes-agent by reusing the exact same graceful fallback
string.

## Verification

```
./scripts/test.sh \
  tests/test_tool_limit_terminal_state.py \
  tests/test_agent_max_turns_parity.py \
  tests/test_auto_compression_terminal_failure.py \
  tests/test_stable_assistant_turn_anchor_phase0.py \
  tests/test_a11y_transcript_landmarks.py \
  tests/test_webui_state_db_context_reconciliation.py \
  tests/test_run_journal.py
```

**59 passed in 6.49s**.

`./scripts/hermes-webui-pr-gate <worktree>` (post-commit,
required for WebUI PRs) passed clean:
- ruff_lint on `upstream/master`-diff: 0 findings on added/modified
  lines (7 pre-existing in file, untouched).
- Whitespace gate clean.
- No conflict markers, no local dev artifacts, b3nw Co-authored-by
  trailer present.

Manual verification (deferred, on dev/local running stack): replay
a long-running task against any reasoning-only model (Claude
extended thinking, DeepSeek v4, Codex Responses) that hits
`agent.max_turns`, confirm the UI renders the graceful fallback
text with the "Tool iteration limit reached" status card instead
of the apperror inline error.

## Risks / Follow-ups

- **Status-card display.** Tags the injected turn with
  `_max_iteration_summary_fallback: True` for future filtering.
  No frontend changes are needed for this PR — the existing
  status-card rendering path already triggers on the assistant
  message with the `_terminal_state` annotation that
  `_mark_latest_assistant_tool_limit_status` attaches.
  If maintainers prefer the fallback turn to surface without any
  status card (true parity with normal assistant turns), the change
  is one conditional — happy to follow up.
- **Goal path unchanged.** `api/goals.py::evaluate_after_turn()`
  (the `/goal` pause) is intentionally untouched; the
  correction comment on #5494 explains why it's a separate concern.
- **Configurability.** If maintainers prefer the fix behind a
  config gate (e.g. `agent.summary_on_iteration_limit`,
  defaulting on for parity), the change is trivial — wrap the
  helper call in a config read. Out of scope for this PR per the
  #5494 commenter's preference for default parity, but easy to
  add if requested.
- **No CHANGELOG entry** per the WebUI contributor convention
  (release workflow owns changelog).
- **No upstream issue/RFC text changes** beyond the
  posted correction comment on #5494 itself.

## Model Used

Authored hermes-webui/developer profile with `anthropic/opus-4.8 & kilo/minimax/minimax-m3` used.

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -1389,6 +1389,40 @@ def _agent_result_tool_limit_reached(result) -> bool:
     return False
 
 
+def _maybe_inject_max_iteration_summary_fallback(messages, result) -> list:
+    """Append the agent's graceful summary text as an assistant turn when one is missing.
+
+    When ``AIAgent`` exhausts its iteration budget, ``agent.handle_max_iterations``
+    always returns a non-empty ``final_response`` — either the model-generated
+    summary or a graceful fallback (e.g. ``"I reached the iteration limit and
+    couldn't generate a summary."``). Hermes Agent surfaces that string as the
+    final answer to the user; the WebUI, by contrast, reads only ``messages``,
+    so an empty summary (common with reasoning-only responses) left the user
+    with a bare ``tool_limit_reached`` error instead of any closure text.
+
+    When ``_tool_limit_reached`` is true and ``messages`` ends without a final
+    assistant answer, inject ``result['final_response']`` as a new assistant
+    turn so ``_mark_latest_assistant_tool_limit_status`` can attach the status
+    card in the normal flow and the user sees the same closure text as
+    hermes-agent. Returns the (possibly new) messages list; does nothing when
+    a usable assistant answer already exists or when ``result`` carries no
+    graceful fallback text.
+    """
+    if not isinstance(result, dict):
+        return list(messages or [])
+    fallback = result.get('final_response')
+    if not isinstance(fallback, str) or not fallback.strip():
+        return list(messages or [])
+    out = list(messages or [])
+    if not _session_lacks_final_assistant_answer(out):
+        return out
+    # Append a synthetic summary turn. Tag it so downstream consumers can
+    # distinguish it from model-emitted assistant turns if needed; mirrors the
+    # synthetic-scaffolding flag convention already used elsewhere (#5334).
+    out.append({"role": "assistant", "content": fallback, "_max_iteration_summary_fallback": True})
+    return out
+
+
 def _mark_latest_assistant_tool_limit_status(messages) -> bool:
     """Annotate the latest usable assistant final answer as limit-stopped."""
     for msg in reversed(list(messages or [])):
@@ -8239,6 +8273,20 @@ def _run_agent_streaming(
                         _result_messages,
                         enabled=_tool_limit_reached,
                     )
+                    # #5494 — parity with hermes-agent's handle_max_iterations() return
+                    # value. When the agent produced no usable summary assistant
+                    # message but result['final_response'] carries a graceful fallback
+                    # string, inject it as a final assistant turn so the user sees
+                    # closure text instead of a bare tool_limit_reached error. Apply
+                    # the synthesis to result['messages'] AND _result_messages so
+                    # downstream _all_result_messages checks (silent-failure detection
+                    # at #all_result_messages on the same turn) see the fallback too.
+                    if _tool_limit_reached:
+                        _result_messages = _maybe_inject_max_iteration_summary_fallback(
+                            _result_messages, result
+                        )
+                        if isinstance(result, dict) and result.get('messages') is not None:
+                            result = {**result, 'messages': _result_messages}
                     if cancel_event.is_set():
                         _finalize_cancelled_turn(s, ephemeral=False)
                         try:
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -8278,14 +8278,18 @@ def _run_agent_streaming(
                     # message but result['final_response'] carries a graceful fallback
                     # string, inject it as a final assistant turn so the user sees
                     # closure text instead of a bare tool_limit_reached error. Apply
-                    # the synthesis to result['messages'] AND _result_messages so
+                    # the synthesis to result['messages'] AND _result_messages so the
                     # downstream _all_result_messages checks (silent-failure detection
-                    # at #all_result_messages on the same turn) see the fallback too.
+                    # at api/streaming.py:_assistant_reply_added_after_current_turn)
+                    # see the fallback too. `finalize_turn` in the agent always returns
+                    # messages as a list, but we write back unconditionally so the
+                    # contract is "if we built a result-messages list, the silent-failure
+                    # classifier reads the augmented version."
                     if _tool_limit_reached:
                         _result_messages = _maybe_inject_max_iteration_summary_fallback(
                             _result_messages, result
                         )
-                        if isinstance(result, dict) and result.get('messages') is not None:
+                        if isinstance(result, dict):
                             result = {**result, 'messages': _result_messages}
                     if cancel_event.is_set():
                         _finalize_cancelled_turn(s, ephemeral=False)
--- a/tests/test_tool_limit_terminal_state.py
+++ b/tests/test_tool_limit_terminal_state.py
@@ -280,6 +280,123 @@ def test_streaming_tool_limit_without_final_answer_emits_no_final_apperror(tmp_p
     )
 
 
+def test_streaming_tool_limit_with_fallback_final_response_surfaces_closure_text(tmp_path, monkeypatch):
+    """#5494 — when the agent produced no usable summary but returned a graceful
+    fallback string (handle_max_iterations() always returns one), inject that as a
+    final assistant turn so the user sees closure text instead of a bare error.
+    """
+    graceful = "I reached the iteration limit and couldn't generate a summary."
+    result = {
+        "turn_exit_reason": "max_iterations_reached(30/30)",
+        "final_response": graceful,
+        "messages": [
+            {"role": "user", "content": "Do the long task."},
+            {"role": "assistant", "content": "", "tool_calls": [{"id": "call_1"}]},
+            {"role": "tool", "tool_call_id": "call_1", "content": "result"},
+            {"role": "user", "content": streaming._MAX_ITERATION_SUMMARY_REQUEST},
+        ],
+    }
+
+    events, payload = _run_streaming_with_fake_agent(tmp_path, monkeypatch, result)
+
+    # The user sees the graceful fallback, not a bare tool_limit_reached error.
+    assert not [
+        ap for ev, ap in events if ev == "apperror"
+        and ap.get("type") == "tool_limit_reached"
+    ], "expected no tool_limit_reached apperror when fallback was returned"
+    done_payloads = [payload for event, payload in events if event == "done"]
+    assert done_payloads, "expected done SSE payload"
+    assert done_payloads[-1]["terminal_state"] == "tool_limit_reached"
+
+    # Fallback text is shown as a final assistant message and is annotated
+    # with the status card so the UI can render the 'limit reached' chip.
+    assistant = payload["messages"][-1]
+    assert assistant["role"] == "assistant"
+    assert assistant["content"] == graceful
+    assert assistant["_terminal_state"] == "tool_limit_reached"
+    assert assistant["_statusCard"]["title"] == "Tool iteration limit reached"
+    # Synthetic scaffolding turn was still dropped, even after fallback injection.
+    assert all(
+        message.get("content") != streaming._MAX_ITERATION_SUMMARY_REQUEST
+        for message in payload["messages"]
+    )
+
+
+def test_streaming_tool_limit_with_fallback_does_not_double_inject_when_assistant_exists(tmp_path, monkeypatch):
+    """#5494 — when the agent already appended a model-generated summary, the
+    fallback flow is a no-op; the existing summary is preserved as before.
+    """
+    summary = "I reached the limit; here is the summary."
+    result = {
+        "turn_exit_reason": "max_iterations_reached(30/30)",
+        "final_response": summary,
+        "messages": [
+            {"role": "user", "content": "Do the long task."},
+            {"role": "assistant", "content": "", "tool_calls": [{"id": "call_1"}]},
+            {"role": "tool", "tool_call_id": "call_1", "content": "result"},
+            {"role": "user", "content": streaming._MAX_ITERATION_SUMMARY_REQUEST},
+            {"role": "assistant", "content": summary},
+        ],
+    }
+
+    events, payload = _run_streaming_with_fake_agent(tmp_path, monkeypatch, result)
+
+    done_payloads = [payload for event, payload in events if event == "done"]
+    assert done_payloads
+    assistant_msgs = [
+        m for m in payload["messages"]
+        if m.get("role") == "assistant" and m.get("content") == summary
+    ]
+    assert len(assistant_msgs) == 1, "fallback must not duplicate the existing summary"
+    assert assistant_msgs[0]["_terminal_state"] == "tool_limit_reached"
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Thanks for the thorough writeup and the correction comment on #5494 — the two-budget distinction (`agent.max_turns` tool-loop cap vs `goals.max_turns`) is exactly right, and the maintainer already signed off on surfacing the summary. I pulled the branch and read the diff against `origin/master`, then cross-checked the agent contract this fix depends on. One finding that I think changes the merge story, so flagging it before this lands.

## The premise may be stale against the current agent

The wire-up in `_run_agent_streaming` targets the silent-failure classifier, which reads the raw result payload:

```python
# api/streaming.py:8433
_all_result_messages = result.get('messages') or []
```

The helper only injects when `_session_lacks_final_assistant_answer(...)` is true (`api/streaming.py:5363`). So the fix hinges on `result['messages']` ending **without** a final assistant turn in the empty-summary case.

Reading the agent reference, that no longer looks guaranteed. `handle_max_iterations` appends only the synthetic *user* request and, on empty summary, sets the fallback string without appending an assistant row:

```python
# agent/chat_completion_helpers.py:1673-1676
if final_response:
    messages.append({"role": "assistant", "content": final_response})
else:
    final_response = "I reached the iteration limit and couldn't generate a summary."
```

But the return value then flows through `finalize_turn`, which enforces the "delivered final_response ⇒ assistant row in transcript" invariant (#43849/#44100):

```python
# agent/turn_finalizer.py:199-205
if final_response and not interrupted:
    _tail_role = messages[-1].get("role") if messages else None
    if _tail_role != "assistant":
        messages.append({"role": "assistant", "content": final_response})
```

There's no `messages.append` between the `handle_max_iterations` return (`turn_finalizer.py:70`) and this block, so in the empty-summary path the tail is the synthetic user request, `_tail_role != "assistant"` holds, and `finalize_turn` appends the fallback string as a real assistant row **inside `result['messages']`**. That row then makes `_assistant_reply_added_after_current_turn(...)` true at `streaming.py:8433`, and `_session_lacks_final_assistant_answer(s.messages)` false at `streaming.py:8488` — so the existing status-card branch already surfaces the fallback and no `tool_limit_reached` apperror fires. In that case `_maybe_inject_max_iteration_summary_fallback` early-returns and is a no-op.

Note also: your PR body cites `chat_completion_helpers.py:1462-1466` for the append-only-when-non-empty logic, but in the reference I'm reading that code lives at 1670-1676 (~200-line drift). That suggests the investigation ran against an older agent snapshot, likely one predating the #43849/#44100 `finalize_turn` invariant — which would explain why you observed the bare error then.

## Recommendation

Can you confirm the empty-summary repro against the **pinned** agent version this WebUI ships? Two outcomes:

- If the pinned agent already has the `finalize_turn:199-205` invariant, the live bug is gone and this PR is belt-and-suspenders. Still worth keeping as a defensive net + regression pins, but the PR body should reframe from "closes a live gap" to "guards the contract," and the `_max_iteration_summary_fallback` tag double-guards against a future agent regression.
- If the pinned agent predates it, the fix is correct as-is — but then the more durable place is arguably the agent invariant, not a second WebUI-side reconstruction of the same closure text.

## On the greptile writeback nit

Agreed it's real but low-severity. The guard `if isinstance(result, dict) and result.get('messages') is not None` skips the `result['messages']` writeback when `messages` is None, yet `_agent_result_tool_limit_reached` keys off `turn_exit_reason`/`terminal_reason` metadata (`streaming.py:1371-1389`), never `messages` — so tool-limit can be true with `messages=None`. In practice `finalize_turn` always returns a list, so `None` shouldn't occur here, but since the helper already tolerates falsy input (`list(messages or [])`), dropping the `is not None` condition and always writing back `_result_messages` would remove the asymmetry greptile flagged at no cost.

claw-io:
Thanks for the catch — pushed a follow-up that addresses both pieces of
review feedback.

**What changed in this push (commit `34fcc3a3` on top of the rebased
`02692cd6`):**

- Rebased onto current `upstream/master` (PR was 4 commits behind).
- Dropped the `if result.get('messages') is not None` guard
  (Greptile's writeback nit). Writeback is now unconditional on
  `isinstance(result, dict)`.
- Reframed the test docstrings as **dual-purpose contract pins** so
  they read correctly under both the live-bug interpretation and the
  invariant-holds interpretation.
- Reframed the PR body around the premise check (now `Refs #5494`
  rather than `Closes #5494` until you confirm the invariant lineage)
  and ack'd your note that *"the more durable place is arguably the
  agent invariant, not a second WebUI-side reconstruction of the same
  closure text."*

**On your premise-check question — sorry, I can't fully confirm from
this side, and that's the gap I'd like you to close:**

When I check `nousresearch/hermes-agent` master locally at
`a72bb0375`, I don't see the invariant block you cite at
`turn_finalizer.py:199-205`. The lines around 199-205 there are the
file-mutation-verifier footer block, not a
*`finalize_turn` guarantees assistant row in transcript* invariant.
And `handle_max_iterations()` only appends an assistant row when
`final_response` is non-empty (it does fall through to the fallback
string but doesn't put it in the messages list on its own). So the
empty-summary path still looks like it would ship the
`messages`-without-assistant-row outcome from the master branch I can
see.

That doesn't refute your claim — it suggests the invariant lives in a
branch the WebUI ships against that isn't visible from here (private
`dev`/`release/*` branch you maintain, or a tag the WebUI pins to).
Could you confirm which of these holds:

1. The hermes-agent pin this WebUI release ships against already has
   the `finalize_turn` invariant you described, and the live bug is
   gone on the user-facing pin. → I'll change the title to `Closes
   #5494`, keep this PR's frames, and the WebUI helper becomes a
   regression guard.
2. The agent pin does NOT carry it, and the empty-summary case is a
   live bug. → no title change needed; the PR closes the gap.
3. You'd prefer to take the invariant upstream into hermes-agent and
   park this PR in favor of an agent-side patch. → I'll close this
   PR and hand off.

I did look at why my line-number citation drifted (`chat_completion_helpers.py:1462-1466`
vs the reference's `1670-1676`): there's about 200 lines of recent
agent master between those two reads (config additions, codex
reasoning refactors, the OAuth Anthropic transport). The append-only-
when-non-empty logic exists in both, just at different lines. So my
PR body's reading of the helper is correct in shape — just stale on
its line citation. I've kept the body accurate rather than chasing
the exact lines that the WebUI-pinned agent uses, since your line
number came from a different agent pin anyway.

CI on this push reruns in progress; I'll update if anything other
than the previous green status flips.

nesquena-hermes:
Shipped in **v0.51.913** 🎬 — thanks @claw-io (co-authored by @b3nw)!

When a turn exhausts its tool/iteration budget, the agent's graceful closing message now shows as a proper assistant turn instead of a bare `tool_limit_reached` error card — most visible with reasoning-only models (extended-thinking Claude, DeepSeek, Codex Responses) whose summary arrives empty.

**Full gate (all green):**
- **Codex** adversarial: SAFE TO SHIP — fires only under `_tool_limit_reached`, no double-inject when a final answer already exists, empty/None `final_response` produces no empty synthetic turn, tool/role linkage intact, `_max_iteration_summary_fallback` tag doesn't leak past the API-key whitelist.
- **Fable** correctness (backend advisor): SAFE — cannot inject a spurious/duplicate turn on a turn that already answered; the "ends without a final turn" guard correctly detects existing (incl. agent-appended) summaries; degrades cleanly to pre-PR behavior in every contract variant.
- **Full pytest suite:** clean. Your 17 terminal-state tests pass. (One TLS-test failure seen mid-gate was a stage-worktree venv artifact — the test reads the first 2000 bytes of server stdout and missing optional deps flooded it with plugin-load warnings; it passes with deps present, and this PR touches no TLS code.)

Purely additive (+178, `api/streaming.py` + tests, 0 deletions). Attribution preserved via the contributor commits in history + CHANGELOG credit. Closes #5494.

Issue #2405, opened 2026-05-16
bug(ui): custom:* provider model selection rejected by false-positive mismatch warning and forced fallback reset


claw-io:
## Bug Description

When the active provider is  (or any  provider) and the user selects a dynamically listed model from the dropdown:

1. A toast warning appears: ""gitlawb/mimo-v2.5-pro" may not work with your configured provider (custom:llm-proxy)."
2. The dropdown selection is silently force-switched back to the default/fallback model.
3. This repeats every time the user re-selects the model — the selection cannot stick.

## Root Cause

Three separate code locations in  interact to produce this result, all stemming from a guard that only checks for the bare string  instead of handling  variants.

### 1. `_checkProviderMismatch` (ui.js:971) — false-positive warning

```js
if(!ap||ap==='custom'||ap==='openrouter') return null; // can't reliably check
```

This guard matches only bare . When the active provider is ,  is  which does NOT match . The function then compares the model's provider prefix (e.g. `gitlawb/...` -> `gitlawb`) against `'custom:llm-proxy'` — they'll never match, producing a false-positive warning.

Fix: `ap.startsWith('custom:')` in addition to `ap==='custom'`.

### 2. `_isPortalFetch` in `_addLiveModelsToSelect` (ui.js:901) — inconsistent `@`-prefix application

```js
const _isPortalFetch=_ap && _ap!=='openrouter' && _ap!=='custom' && _ap!=='openai-codex' && provider===_ap;
```

Same guard pattern. For `custom:llm-proxy`, `_ap!=='custom'` is `true`, making `_isPortalFetch` potentially `true`. This means live-fetched models *would* get `@custom:llm-proxy:` prefix and would be skipped by `_checkProviderMismatch` (line 973). However, the `/api/models/live` endpoint may not support `custom:*` providers, so models may come from the static model list without the `@` prefix — bypassing the skip.

### 3. `syncTopbar` (ui.js:4390–4412) — forceful model reset

```js
const applied=_applyModelToDropdown(currentModel,modelSel,S.session.model_provider||null);
if(!applied){
  const fallback=_applySessionModelFallback(modelSel);
  if(fallback&&!deferModelCorrection){
    S.session.model=fallback.model;
    S.session.model_provider=fallback.model_provider||null;
    _persistSessionModelCorrection(fallback.model, S.session.model_provider||null);
  }
}
```

[...]

PR #5726, opened 2026-07-07
fix(#5567): stop stamping a foreign profile's provider onto a session (frontend model-picker race)

### LINKED ISSUE #5567: bug: cross-profile HERMES_HOME race causes turn-init failures with wrong profile's provider (v0.51.849 repro; follow-up to #2321)

nesquena-hermes:
## Summary

Fixes the **second, distinct root cause** behind #5567's "Provider 'ollama' is set in config.yaml but no API key was found" symptom — a purely **frontend** vector that #5577 (the shipped `HERMES_HOME` context-local fix) could not touch.

Traced by @b3nw from outside-the-app log analysis and confirmed against `master`. Two separate vectors produce the identical error:

| Vector | Where | Status |
|---|---|---|
| Backend `HERMES_HOME` clobber → init reads a **foreign profile's `config.yaml`** | agent config reader | ✅ fixed by #5577 |
| **Frontend writes the wrong `model_provider`** into the chat-start payload → persisted into session JSON → backend inits with it | `_modelStateForSelect` | ✅ **this PR** |

## Root cause

`_modelStateForSelect(sel, modelId)` (`static/ui.js`) resolved the provider by reading `sel.selectedOptions[0]` — the dropdown's *currently-selected* option — instead of the option whose **value** matches `modelId`.

In the multi-tab / per-profile desktop scenario from the report: switching tabs/profiles rebuilds the model dropdown, and during that window `selectedOptions[0]` transiently still holds the **previous** profile's default (e.g. an ollama model, `qwen3.6:27b-mlx`). A send in that window resolves the provider for an unrelated model (`kilo/minimax/minimax-m3`) as `ollama`.

Because `model_provider` is a **persisted** session field and `_modelProviderForSend` returns `S.session.model_provider` **first**, the contamination is **sticky**: once written into the session JSON, every subsequent turn re-sends the foreign provider and fails at agent init — for a provider the session never used.

## Fix (two parts)

1. **Stop producing the wrong provider** — `_modelStateForSelect` now resolves the provider from the option whose value equals the requested model. The currently-selected option is trusted **only** when it actually matches that value (preserves the user's exact pick in the same-model/different-provider collision case). A genuinely-absent model yields a **null** provider, so the backend re-infers from the model string (`kilo/…` → kilocode) instead of inheriting a stale one.

2. **Repair already-poisoned sessions** — `_repairContaminatedSessionModelProvider`, wired into the session-load chokepoint (`static/sessions.js`), drops a stored `model_provider` on load **only** when the model string carries a slash-prefix that unambiguously contradicts it (e.g. `kilo/...` stored as `ollama`). Deliberately conservative — untouched: bare model names, `@provider:` ids, `custom:*` / `openrouter` cross-routing (they own vendor-prefixed ids on purpose, per #2405), and alias-agreeing prefixes (claude/anthropic, gpt/openai, gemini/google).

## Why the previously-rejected mitigations still stand

This is the frontend vector; the earlier analysis of the **backend** `_ENV_LOCK`-scope and post-init-mismatch mitigations (both rejected) is unaffected and remains correct. Nothing here touches `HERMES_HOME`, `_ENV_LOCK`, or the streaming turn path.

## Tests

`tests/test_issue5567_model_provider_contamination.py` drives the **real** JS in Node with a mock `<select>`:
- **the tab-switch race** — dropdown selected = an ollama option, resolve `kilo/minimax/minimax-m3` → provider must **not** be `ollama` (it's `kilocode`).
- collision-preservation (explicit same-value pick kept), missing-model → null provider.
- all four repair branches: repair fires / custom-untouched / alias-untouched / bare-name-untouched.

**Fail-without-fix verified**: reverting the resolver to the buggy `selectedOptions[0]` read makes the behavioral test fail with `assert 'ollama' != 'ollama'` — exactly the reported contamination. 1175 provider/model/session tests pass with the fix; 0 regressions.

## Credit

Root-cause trace and the resolver-fix shape are @b3nw's (issue #5567). Co-authored-by trailer included.

Closes #5567 (frontend vector; #5577 closed the backend vector).

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -1538,15 +1538,14 @@ async function loadSession(sid){
     return loadSession(continuationSid,{...opts,skipLineageResolve:true,skipContinuationResolve:true,force:true});
   }
   S.session=data.session;
-  // #5567: repair a session whose stored model_provider was contaminated with a
-  // foreign provider (frontend tab/profile-switch race). Runs before the model
-  // dropdown / send path reads S.session.model_provider, so a poisoned session
-  // stops re-sending the wrong provider from its very next turn.
-  if(typeof _repairContaminatedSessionModelProvider==='function') _repairContaminatedSessionModelProvider(S.session);
   if(typeof _clearEmptyComposerModelOverride==='function') _clearEmptyComposerModelOverride();
   // Loading a real existing session abandons any pre-session toolset override
   // staged on the empty composer before any deferred refresh work runs.
   S._pendingSessionToolsets=null;
+  // #5567: drop a contaminated stored model_provider before the first reader
+  // (populateModelDropdown / the send path). Full rationale on
+  // _repairContaminatedSessionModelProvider in ui.js.
+  if(typeof _repairContaminatedSessionModelProvider==='function') _repairContaminatedSessionModelProvider(S.session);
   if(typeof populateModelDropdown==='function'){
     const modelRefreshSid=sid;
     const isActiveModelRefreshSession=()=>!!(S.session&&S.session.session_id===modelRefreshSid);
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -1538,6 +1538,11 @@ async function loadSession(sid){
     return loadSession(continuationSid,{...opts,skipLineageResolve:true,skipContinuationResolve:true,force:true});
   }
   S.session=data.session;
+  // #5567: repair a session whose stored model_provider was contaminated with a
+  // foreign provider (frontend tab/profile-switch race). Runs before the model
+  // dropdown / send path reads S.session.model_provider, so a poisoned session
+  // stops re-sending the wrong provider from its very next turn.
+  if(typeof _repairContaminatedSessionModelProvider==='function') _repairContaminatedSessionModelProvider(S.session);
   if(typeof _clearEmptyComposerModelOverride==='function') _clearEmptyComposerModelOverride();
   // Loading a real existing session abandons any pre-session toolset override
   // staged on the empty composer before any deferred refresh work runs.
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -2481,6 +2481,11 @@ function _resolveSessionModelForDisplaySoon(sid){
       if(!model||!S.session||S.session.session_id!==sid) return;
       S.session.model=model;
       S.session.model_provider=provider||null;
+      // #5567: the deferred resolver re-fetches model_provider from the backend,
+      // which faithfully echoes a still-poisoned stored provider — undoing the
+      // repair loadSession() just applied. Re-run it here so a contaminated
+      // session stays healed after the resolve, not just at first paint.
+      if(typeof _repairContaminatedSessionModelProvider==='function') _repairContaminatedSessionModelProvider(S.session);
       const resolvedContextLength=data.session.context_length||S.session.context_length||0;
       S.session.context_length=resolvedContextLength;
       S.session.threshold_tokens=data.session.threshold_tokens||0;
--- a/static/ui.js
+++ b/static/ui.js
@@ -2570,7 +2570,24 @@ function _modelStateForSelect(sel, modelId){
   if(!value) return {model:'',model_provider:null};
   const explicitProvider=_providerFromModelValue(value);
   if(explicitProvider) return {model:value,model_provider:explicitProvider};
-  const opt=sel&&sel.selectedOptions&&sel.selectedOptions[0];
+  // Resolve the provider from the option whose VALUE matches the requested
+  // model — never blindly from sel.selectedOptions[0] (#5567). During a profile
+  // /tab switch or a model-list rebuild the dropdown transiently still has the
+  // PREVIOUS profile's default option selected (e.g. an ollama model), so reading
+  // selectedOptions[0] would stamp that foreign provider onto a model it doesn't
+  // own — which is then persisted into the session's model_provider and re-sent
+  // on every turn, bricking it with a "Provider 'X'…no API key" error for a
+  // provider the session never used.
+  let opt=null;
+  const selected=sel&&sel.selectedOptions&&sel.selectedOptions[0];
+  // Prefer the currently-selected option ONLY when it actually is the requested
+  // model — this preserves the user's exact pick in the same-value/different-
+  // provider collision case (two providers offering the same model id).
+  if(selected&&String(selected.value||'')===value){
+    opt=selected;
+  }else if(sel&&sel.options){
+    opt=Array.from(sel.options).find(o=>String(o.value||'')===value)||null;
+  }
   const provider=String(_getOptionProviderId(opt)||'').trim();
   return {model:value,model_provider:(provider&&provider!=='default')?provider:null};
 }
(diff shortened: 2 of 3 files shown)


nesquena (approved):
## Review — end-to-end ✅ (approved after two fixes I pushed — one real over-repair defect, one CI-guard trip)

Independent review of the **frontend vector** of #5567 (the backend `HERMES_HOME` vector shipped in #5577, which I reviewed): `_modelStateForSelect` resolved the provider from `sel.selectedOptions[0]` — during a profile/tab-switch dropdown rebuild that's transiently the *previous* profile's default, so a send in that window stamped a foreign provider (e.g. `ollama`) onto an unrelated model, persisted it into the session's `model_provider`, and `_modelProviderForSend` re-sent it on every turn — sticky contamination. Credit @b3nw for the trace.

### What this ships

`static/ui.js` (resolver fix + `_repairContaminatedSessionModelProvider`), `static/sessions.js` (repair wired into the `loadSession` chokepoint), `tests/test_issue5567_model_provider_contamination.py` (node-driven behavioral tests against the real JS). Agent-authored (`nesquena-hermes`).

### The resolver fix — correct

Provider now resolves from the option whose **value equals the requested model**; the currently-selected option is trusted only when its value matches (preserving the user's exact pick in the same-value/different-provider collision case); a genuinely-absent model yields a **null** provider so the backend re-infers. The PR's behavioral test drives the real tab-switch race (`selectedOptions[0]` = an ollama option, resolve `kilo/minimax/minimax-m3`) and I confirmed fail-without-fix. ✓

### What I caught — the repair would have poisoned healthy Nous sessions (the reverse contamination)

The load-repair drops a stored `model_provider` when the model's slash-prefix "contradicts" it, excluding only `custom:*`/`openrouter`. That exclusion list was **missing the other aggregators**, and the miss is not theoretical:

1. **A healthy Nous session persists exactly the "contradiction" shape.** `resolve_model_provider("@nous:anthropic/claude-opus-4.6")` unpacks to model `anthropic/claude-opus-4.6` + provider `nous` **before the session write** (pinned by `tests/test_issue895_894_nous_prefix.py:75`; the streaming layer persists the resolved pair).
2. On load, the repair sees prefix `anthropic` ≠ stored `nous` → **nulls the correct provider**.
3. The send path re-infers — I ran it: `resolve_model_provider("anthropic/claude-opus-4.6")` → **`openrouter`**. Net: every healthy Nous session using a vendor-prefixed catalog model gets silently re-routed to a provider it never used on next load — bricked if no OpenRouter key, silently mis-billed if there is one. The exact failure class this repair exists to fix, inflicted in reverse.
4. The canonical set is upstream: `hermes_cli/model_normalize.py:67` `_AGGREGATOR_PROVIDERS = {openrouter, nous, kilocode}` — "providers whose APIs consume vendor/model slugs." (`kilocode` was a latent second case: stored `kilocode` vs prefix `kilo` → repair fired; harmless today only because re-inference lands back on kilocode.)

### What I pushed

- **`1770a282`** — extend the repair's exclusion to the full upstream aggregator set (`nous`, `kilocode`), with the persistence-shape rationale in the comment, plus two node regression scenarios (healthy nous + kilocode sessions must survive the load repair). **Fail-without/pass-with verified**: with the PR's original exclusion list, the nous scenario fails (provider nulled); with the fix, 3/3.
- **`a653a891`** — the full suite then caught the PR's `loadSession` call-site comment pushing `S._pendingSessionToolsets=null` past the **#4490 static guard's** 400-char window (same class as the #4755/#5170 boot.js incident). Moved the repair call below the toolsets clear — semantically identical (`_clearEmptyComposerModelOverride` only nulls a host field; the first `model_provider` reader, `populateModelDropdown`/send, still runs after the repair) — and trimmed the call-site comment to a pointer. 14/14 across both guard families.

### Edge-case matrix (repair)

| Session state | Repair |
|---|---|
| `kilo/minimax/minimax-m3` + `ollama` (the reported poisoning) | nulled → backend re-infers kilocode ✅ |
| `anthropic/claude-opus-4.6` + `nous` (healthy aggregator) | **untouched** ✅ (my fix; was nulled) |
| `kilo/minimax/minimax-m3` + `kilocode` | **untouched** ✅ (my fix) |
| `opencode_go/deepseek-v4` + `custom:llm-proxy` | untouched ✅ |
| `claude/sonnet` + `anthropic` (alias-agree) | untouched ✅ |
| `gpt-4o` (bare) + anything | untouched ✅ |
| `@provider:model` ids | untouched (self-describing) ✅ |
| URI-scheme ids | untouched ✅ |

### Tests

- PR's node behavioral suite + my two scenarios — **14/14** (with the #4490 family), all driving the real `ui.js` functions; the core race, collision-preservation, null-on-missing, and all repair branches are behavior-tested, and the original fail-without-fix property holds.
- Provider/model blast-radius — **1131 passed, 6 skipped**.
- **Full suite: 12108 passed / 0 failed** in 273s (deselected the known darwin CRLF flake) — including the #4490 guard that was red before `a653a891`.
- `node --check` clean on both JS files.

### Recommendation

✅ **Approved after fixes.** Parked at approval — ready for the release agent's merge/tag pipeline.

The resolver fix is exactly right and the repair idea is sound — but the repair's aggregator exclusion had to match the upstream `_AGGREGATOR_PROVIDERS` set, or it would have re-created the bug in reverse against every healthy Nous session (proven end-to-end: persistence shape → repair fires → re-inference lands on openrouter). Fixed in `1770a282` with mutation-verified regressions, plus the #4490 source-window trip fixed in `a653a891`. Full suite clean. Ship.

nesquena-hermes:
## Codex gate: CORE finding fixed → re-gate

The Codex regression gate flagged one **CORE-severity** issue (exactly the "does the repair belong only at that one load site?" concern), now fixed:

**The repair was undone milliseconds after it ran.** `loadSession()` repairs `S.session.model_provider`, but the deferred `_resolveSessionModelForDisplaySoon()` then re-fetches `/api/session?...resolve_model=1` and reassigns `model_provider` from the backend response — which faithfully echoes the *still-poisoned* stored provider (backend fast path preserves `kilo/minimax/minimax-m3` + `ollama` verbatim). So the next send used `ollama` again and failed. Fix: re-run `_repairContaminatedSessionModelProvider(S.session)` immediately after that reassignment (before `_modelResolutionDeferred=false` / `syncTopbar()`), so a contaminated session stays healed after the resolve, not just at first paint.

Added the regression coverage Codex asked for: a source-guard (repair ordered after the resolver's reassignment) + a behavioral scenario modeling the full `load-repair → backend-echo → re-run` sequence, asserting the provider ends `null`. Fail-without-fix verified.

### Two related review fixes folded in (branch now linear on master)

- **Exclude the full aggregator set from the repair** (`nous`, `kilocode`, not just `custom:*`/`openrouter`). A *healthy* Nous session persists the resolved shape (`model: anthropic/claude-opus-4.6`, `provider: nous`) — the repair would have nulled that correct provider and let the send path re-route it to openrouter, bricking a working session. This was a genuine false-positive in my original repair; good catch.
- **Keep the load-repair inside the #4490 static-guard window** — the call-site comment block had pushed `S._pendingSessionToolsets=null` out of the 400-char window the #4490 guard asserts; the repair call moved below the toolsets clear (semantically identical — nothing reads the session in between, and the first `model_provider` reader still runs after the repair).

### Verification

1250 provider / model / session / #4490-guard / #5567 / nous-prefix tests pass; 0 regressions. All four commits (original + the two review fixes + the Codex fix) are on the branch in order, authorship preserved.

Ready for re-review.

nesquena-hermes:
Shipped in **v0.51.916** — thanks @nesquena-hermes and @b3nw!

The **resolver fix** from this PR is live: `_modelStateForSelect` now resolves the provider from the dropdown option whose value matches the requested model (trusting the selected option only when its value matches), instead of blindly reading `selectedOptions[0]`. That stops the profile/tab-switch race from stamping a foreign provider onto a session — the #5567 write-vector is closed.

**Scope note:** the load-time sticky-repair for sessions *already poisoned* before this fix was NOT shipped. A 3-round Codex + Fable gate proved a frontend repair reading the shared `#modelSelect` DOM is a cross-session race surface — it nulled healthy NVIDIA NIM / Cursor ACP / ollama-HF providers (re-routing them to openrouter, the reverse contamination) and couldn't reliably heal live-only-model sessions even when hardened. That repair is deferred to a robust backend chat-start implementation, tracked in **#5731**. Already-poisoned sessions self-heal on their next healthy send / manual re-pick in the meantime.

Full gate: Codex SAFE, Fable SAFE (resolver), full suite clean. Credit preserved via Co-authored-by. Combined with #5727 (same fix, @b3nw's trace).

PR #5727, opened 2026-07-07
fix(model-picker): prevent cross-profile provider dropdown contamination

### LINKED ISSUE #5567: bug: cross-profile HERMES_HOME race causes turn-init failures with wrong profile's provider (v0.51.849 repro; follow-up to #2321)

b3nw:
## Thinking Path
- Hermes WebUI maintains profile-isolated workspaces, configurations, and session model selections.
- Setting up a new session or composer turn queries the dropdown state resolver `_modelStateForSelect` to extract the active model and provider.
- The bug was that `_modelStateForSelect` fell back to retrieving the provider from the active DOM selection (`sel.selectedOptions[0]`) even when queried for a model ID different from the currently selected one.
- On profile switches or cold session boots, the dropdown momentarily holds the previous profile's default model (e.g., `qwen3.6:27b-mlx` / provider `ollama`).
- This caused custom provider models (like `kilo/minimax/minimax-m3`) to resolve with the provider `"ollama"`, poisoning the session configuration file and causing the backend worker thread to crash with a `RuntimeError` due to missing `OLLAMA_API_KEY` credentials in the active profile.
- This PR fixes the dropdown query to search `sel.options` for the option matching the target model value, and only fall back to the active selected option if its value actually matches the queried model ID.

Closes #5567 

## What Changed
- **`static/ui.js`**: Patched `_modelStateForSelect` to locate the option by value from the dropdown options list instead of assuming `selectedOptions[0]` is correct.
- **`tests/test_chat_start_provider_fallback.py`**: Added `test_model_state_does_not_use_unrelated_selected_option_provider` to verify that resolving the provider of a model not currently present in the picker returns `null` instead of stealing the selected option's provider.

## Why It Matters
- Prevents cross-profile/cross-model provider contamination and intermittent agent initialization failures (`RuntimeError` on missing API keys) when switching between profiles with distinct provider configurations.

## Verification
- Verified that all model resolution, picker, and fallback tests (152 tests total) pass locally using `uv run pytest`.

## Risks / Follow-ups
- None. The fix preserves the fallback to `selectedOptions[0]` only when the requested model matches the selected option's value.

## AI Usage Disclosure
- **Provider**: Google
- **Model**: Antigravity (Gemini 3.5 Flash)
- **Notable mode/tool use**: Git worktree checkout, targeted code replacement, local node eslint, and pytest verification.

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -2570,8 +2570,17 @@ function _modelStateForSelect(sel, modelId){
   if(!value) return {model:'',model_provider:null};
   const explicitProvider=_providerFromModelValue(value);
   if(explicitProvider) return {model:value,model_provider:explicitProvider};
-  const opt=sel&&sel.selectedOptions&&sel.selectedOptions[0];
-  const provider=String(_getOptionProviderId(opt)||'').trim();
+  let opt=null;
+  if(sel&&sel.options){
+    opt=Array.from(sel.options).find(o=>o.value===value);
+  }
+  if(!opt){
+    const currentOpt=sel&&sel.selectedOptions&&sel.selectedOptions[0];
+    if(currentOpt&&currentOpt.value===value){
+      opt=currentOpt;
+    }
+  }
+  const provider=String((opt&&_getOptionProviderId(opt))||'').trim();
   return {model:value,model_provider:(provider&&provider!=='default')?provider:null};
 }
 function _captureModelDropdownSelection(sel){
--- a/tests/test_chat_start_provider_fallback.py
+++ b/tests/test_chat_start_provider_fallback.py
@@ -215,6 +215,24 @@ def test_model_state_reads_live_custom_provider_from_option_metadata(driver_path
     assert state == {"model": "llama-3.3-custom", "model_provider": "custom:lab"}
 
 
+@node_test
+def test_model_state_does_not_use_unrelated_selected_option_provider(driver_path):
+    # If the requested model is not in the options and differs from the currently
+    # selected option, it should return model_provider = null instead of the selected option's provider.
+    state = _run_model_state_helper(driver_path, {
+        "model": "kilo/minimax/minimax-m3",
+        "initialValue": "qwen3.6:27b-mlx",
+        "options": [{
+            "provider": "ollama",
+            "optionProvider": "ollama",
+            "value": "qwen3.6:27b-mlx",
+        }],
+    })
+
+    assert state == {"model": "kilo/minimax/minimax-m3", "model_provider": None}
+
+
+
 def test_live_custom_models_are_tagged_with_provider_metadata():
     ui_src = UI_JS_PATH.read_text(encoding="utf-8")
     start = ui_src.index("function _addLiveModelsToSelect")


nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (full gate: Codex + Fable-UX + suite) · profile-isolation fix

**Certified head:** `sha:e2c19cff` (clean rebase, branch `gate-rebase/5727-cross-profile-provider-dropdown`) · **PR:** #5727 · _b3nw (T2), fix(model-picker): prevent cross-profile provider dropdown contamination_
**Verdict:** GREEN. A clean profile-isolation fix — the model picker no longer resolves a model's provider from a mismatched/stale selected `<option>` (which could leak another profile's provider). Codex SAFE, Fable-UX SHIP-UX, full suite green (2× — the one earlier failure was a pre-existing ordering flake, not this PR).

### What I ran (rebased worktree `/tmp/wt-rebase-5727`) — full gate: Codex + Fable + suite
| Gate | Result |
|------|--------|
| Codex (isolation logic / option-matching / no-leak / no causal break) | **SAFE TO SHIP** — 0 findings |
| **Fable-UX** | **SHIP-UX** — picker shows correct provider, no stale cross-profile label |
| Full pytest suite | ✅ **12299 passed, 0 failed** (re-run; see flake note) |

### Findings
**✅ Clean isolation fix (+29/-2):** `_modelStateForSelect(sel, modelId)` previously read `sel.selectedOptions[0]`'s provider blindly — but the selected option can belong to a DIFFERENT model than the resolved `value`, leaking a stale/cross-profile provider. Now it finds the option whose `.value === value` (the actual model), and only falls back to `selectedOptions[0]` if THAT option's value matches too; otherwise the provider resolves to `null`. Codex confirmed: a mismatched selected option no longer leaks its provider (returns null — the contamination fixed), the happy path still resolves the correct provider (not over-nulled), the explicit `provider/model` short-circuit is intact, `Array.from(sel.options).find` is safe with no/duplicate options, and there's no causal break to the `@provider` mismatch handling. New test asserts a model with a mismatched `ollama` selected option returns `model_provider: None`.
**Flake note (not this PR's defect):** `test_provider_mismatch.py::test_at_provider_first_party_named_third_party_model_known_limitation` failed ONCE in the first full-suite run, but PASSES in isolation, in its module, with neighbors, on clean master, AND in a clean full-suite re-run (12299 passed, 0 failed). It's a pre-existing cross-test ordering flake unrelated to #5727 — flagged for a separate root-cause, not a blocker here.

### Recommendation to Nathan
**GREEN — merge from branch `gate-rebase/5727-cross-profile-provider-dropdown` (`sha:e2c19cff`).** A tight profile-isolation fix on the crown-jewel model picker — resolve the provider from the option matching the actual model, not a stale selected option. Codex SAFE + Fable SHIP-UX + full suite green (re-run confirmed). Mostly backend-logic (light visible surface), so a screenshot isn't essential, but it's your call. concept 4/5. Author @b3nw (T2). crit=3, profile-isolation.

---
_Gate-certifier layer (warm-up → **gate** → release). _modelStateForSelect now resolves the provider from the option whose value===the model (not a blind selectedOptions[0] that may belong to another model/profile) → no cross-profile provider leak; happy path + explicit-provider short-circuit intact, find() safe. Codex SAFE + Fable-UX SHIP-UX + full suite green 12299 (2nd run clean; the 1 earlier fail = pre-existing ordering flake, passes isolation/module/neighbors/clean-master, not #5727). Cert valid for sha:e2c19cff._

nesquena-hermes:
Shipped in **v0.51.916** — thanks @b3nw!

Your resolver fix (resolve the provider from the option matching the target model, not the currently-selected one) shipped, combined with #5726. Your root-cause trace of the #5567 frontend vector was spot-on and drove the whole fix. Your regression test (`test_model_state_does_not_use_unrelated_selected_option_provider`) is included.

Credit preserved via Co-authored-by. The already-poisoned-session repair is tracked separately in **#5731** (backend chat-start — a 3-round gate showed the frontend DOM approach was a cross-session race).

PR #542, opened 2026-04-15
fix: strip orphaned tool messages before API calls (fixes #534)

nesquena-hermes:
## Summary

Extends `_sanitize_messages_for_api()` in `api/streaming.py` to strip orphaned tool-role messages before sending conversation history to any provider API.

## Problem

Strictly-conformant providers (Mercury-2/Inception, newer OpenAI models) reject conversation histories that contain `role: tool` messages whose `tool_call_id` cannot be linked to a preceding `assistant` message with `tool_calls`. The error is:

```
400 Bad Request: Message has tool role, but there was no previous assistant message with a tool call
```

This can happen when:
- Conversation histories are edited manually
- Switching between providers that handle tool messages differently
- Partial/cancelled tool results are stored in session history

## Fix

Two-pass approach in `_sanitize_messages_for_api()`:

1. Collect all `tool_call_id`s declared by assistant messages (handles both OpenAI `id` and Anthropic `call_id` field names)
2. Drop any `role: tool` messages whose `tool_call_id` was not declared by a preceding assistant message

Normal well-formed histories are completely unaffected.

## Tests

13 new tests in `tests/test_orphaned_tool_messages.py` covering:
- Valid roundtrip preservation (OpenAI and Anthropic call_id formats)
- Multiple tool calls in one assistant message
- Partial orphan filtering (valid messages kept, orphans dropped)
- Edge cases: `None` tool_calls, missing `tool_call_id`, non-dict entries

Closes #534

PR #5737, opened 2026-07-07
fix(streaming): drop empty tool_calls arrays before API send

Swanzb:
## Problem

Strict providers (DeepSeek v4, newer OpenAI) reject messages with `tool_calls: []` as HTTP 400:

> Invalid 'messages[120].tool_calls': empty array. Expected an array with minimum length 1, but got an empty array instead.

`_sanitize_messages_for_api()` (PR #542 / 9220a87) already strips orphaned `role: tool` messages whose `tool_call_id` cannot be linked to a preceding assistant, but it does not handle the case where an **assistant message itself** carries an empty `tool_calls` array.

This happens when a stored message has `tool_calls: []` (not `None`, not missing — literally an empty JSON array) and the dict comprehension copies it verbatim into the sanitized output.

## Fix

After the dict comprehension in both sanitize paths (`_sanitize_messages_for_api` and `_api_safe_message_positions`), check if `tool_calls` exists and is falsy (empty list) — if so, delete it from the sanitized dict entirely so the outbound request is compliant with the spec.

## Testing

- Normal histories with real tool_calls are completely unaffected
- Histories containing an empty `tool_calls: []` now have it stripped before the API call, preventing the 400

Fixes the WebUI-side gap that surfaces when using strict providers (DeepSeek, newer OpenAI-compat endpoints).

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -4025,6 +4025,10 @@ def _sanitize_messages_for_api(
                 # Orphaned tool result — skip to avoid 400 from strict providers.
                 continue
         sanitized = {k: v for k, v in msg.items() if k in _API_SAFE_MSG_KEYS}
+        # Drop empty tool_calls — strict providers (DeepSeek, newer OpenAI)
+        # reject tool_calls: [] with HTTP 400 even when no orphaned calls exist.
+        if 'tool_calls' in sanitized and not sanitized['tool_calls']:
+            del sanitized['tool_calls']
         # Provider-aware reasoning_content stripping from model-facing history.
         # Historical assistant reasoning_content is stripped only when the user
         # explicitly requests strip mode or auto mode identifies a local/generic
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Shipped in **v0.51.918** — thanks @Swanzb!

Empty `tool_calls: []` arrays are now dropped from model-facing history before the API send, so strict providers (DeepSeek v4, newer OpenAI) no longer 400 a turn whose history contains such a message.

Applied on all three sites that build/compare the API payload: `_sanitize_messages_for_api`, `_api_safe_message_positions`, and the `_safe_projection` inside `_restore_reasoning_metadata` (the third was a Fable gate follow-up I folded in, so a `tool_calls: []` row projects identically on both sides of the metadata carry-forward and keeps its reasoning/id/timestamp). Populated tool-call chains + their linked results are untouched.

Full gate: Codex SAFE (2 rounds), Fable SAFE, full suite 12298/0. Mutation-checked regression tests added. Credit preserved via Co-authored-by.

PR #3329, opened 2026-06-01
fix: Artifacts tab cannot open files when messages carry structured tool metadata

mysoul12138:
## Problem

Clicking an artifact entry in the Artifacts tab shows "Could not open file" even when the file exists in the workspace.

## Root Cause

Two bugs in `static/workspace.js`:

### Bug 1: `collectSessionArtifacts()` misses artifacts when `S.toolCalls` is empty

`_syncToolCallsForLoadedMessages()` clears `S.toolCalls` when messages carry their own structured `tool_calls` / `tool_use` metadata (the common case for sessions loaded from history). `collectSessionArtifacts()` only read `S.toolCalls` and message text (diff/patch fences), so mutation tool paths embedded in structured message metadata were never collected.

### Bug 2: `openArtifactPath()` passes absolute paths to `/api/list`

When the agent records absolute paths in tool arguments (e.g. `/mnt/e/.../workspace/file.js`), `openArtifactPath()` only stripped `~/` and `./` prefixes. The full absolute path was passed to `/api/list?path=...`, which returned 404.

## Fix

1. **`collectSessionArtifacts()`**: Added two new extraction paths — scan `msg.tool_calls` (OpenAI format) and `msg.content` with `type: 'tool_use'` blocks (Anthropic format), reusing the existing `_artifactCandidatesFromToolCall` helper.

2. **`openArtifactPath()`**: Strip the session's `S.session.workspace` prefix from the path before calling `_workspacePathExists()`, so `/api/list` receives a workspace-relative path.

3. **`renderSessionArtifacts()`**: Display workspace-relative paths instead of full absolute paths in the artifact list.

## Verification

- `node --check static/workspace.js` ✅
- `tests/test_issue2655_frontend.py` (4 tests) ✅
- `tests/test_issue3262_artifact_path_canonicalization.py` (2 tests) ✅
- Browser live: loaded session with `patch` tool calls, `collectSessionArtifacts()` returned 1 artifact (previously 0), `openArtifactPath()` correctly stripped workspace prefix, file preview opened successfully.

PR #3262, opened 2026-05-31
Fix workspace preview closing and stale content on chat completion

pamnard:
## Summary
- **Bug 1:** stream `done` called `loadDir('.')`, which always ended with `clearPreview({keepPanelOpen:true})` (#1785 directory navigation). That switched preview-only mode back to the file listing even when the agent did not touch the open file.
- **Bug 2:** keeping preview open without reloading left stale content when the agent did edit the open file during the turn.
- **Fix:** background refresh uses `loadDir('.', { preservePreview: true })` so unrelated responses no longer close preview. Write/edit tool paths are tracked per turn; when the open preview path was mutated, `refreshOpenPreviewIfMutated()` reloads via `openFile()` on `tool_complete` and again after `loadDir` on `done` (skipped while `_previewDirty`).

## Thinking Path
1. Repro: preview closes on model reply while viewing a file the agent did not edit.
2. Trace: `messages.js` `done` → `loadDir('.')` → unconditional `clearPreview()` in `workspace.js`.
3. First fix scoped `preservePreview` to background refresh only; explicit navigation (breadcrumb, folder click) still clears preview (#1785).
4. Follow-up: preserving preview alone leaves stale content when the agent edits that file — add turn-scoped mutation tracking from existing artifact tool path extraction and reload open preview when paths match.

## Test plan
- [x] `pytest tests/test_workspace_preview_preserved_on_stream_done.py tests/test_issue1785_workspace_preview_breadcrumb.py tests/test_issue856_background_completion_unread.py`
- [ ] Open a file in Workspace preview, send a chat message that does not edit it → preview stays open
- [ ] Open a file, ask the agent to edit it → preview stays open and shows updated content when the tool completes / response finishes
- [ ] Start editing a file in preview, agent edits same file → local unsaved edits are not overwritten

Issue #2655, opened 2026-05-20
Add per-session "Artifacts" tab tracking files created/edited in the session, with inline preview


nesquena-hermes:
# Add a per-session "Artifacts" tab tracking files created/edited in the session, with inline preview

## Reporter feedback

From Maheśvara on Discord (May 20 2026, testers channel — second half of his feedback message):

> Also I use plans a lot, it would be nice if we could reliably review plans without me manually opening the file with vim. Idk if there's a viable way to do this but it would be nice to have an "artifacts" tab which tracks created resources in a session and allow for the previewing of these files.

(Got a 👍 reaction.)

![Maheshvara feedback](https://raw.githubusercontent.com/nesquena-hermes/hermes-webui/issue-assets/issue-maheshvara-tasks-artifacts-feedback-may20.jpeg)

## What he's describing

A right-side **Artifacts** tab (sibling to the workspace Files view — see the companion relocation proposal in #(new-todos-issue)) that lists files **the assistant wrote or edited during the current session** and lets the user preview them inline without leaving the browser.

His specific use case is **plan files** (`writing-plans` skill output — markdown files the assistant generates during planning turns). Today, to actually read a plan you either scroll back through the transcript (where the file content was streamed in a tool card) or `vim path/to/PLAN.md` on the host. Neither is the path of least resistance for someone iterating with the assistant.

## Why this is different from the existing workspace panel

The workspace Files panel shows **everything in the workspace root** — old files, unrelated files, the assistant's plan file, the venv, build artifacts. To find "what did the assistant just write?" you have to know the filename and where it landed.

An Artifacts view is **session-scoped and chronological**: it filters down to just the files this conversation produced or modified. Same list a user could mentally reconstruct from the transcript, but rendered as a sidebar instead of as buried tool cards.

## Suggested shape

**Data model.** Every `write_file`/`patch`/tool call that mutates a file on disk already emits a structured tool result with the path. We need to:

1. Maintain a per-session ordered set of `(path, last_op, last_op_ts)` keyed off those tool calls. The set is rebuildable from session history, so no new persistent state — just an index derived on session load and updated as tools run.
2. Surface the set in a new "Artifacts" tab in the workspace side panel (or wherever the right-side session container lands per the companion issue).
3. Click an entry → inline preview pane within the panel:
   - Markdown / plan files → render via the existing `renderMd()` pipeline (we already do this for assistant message content; reusing it for files keeps fidelity).
[...]

PR #5752, opened 2026-07-07
fix(workspace): auto-refresh file preview on mutation + prevent half-screen layout split

### LINKED ISSUE #5747: bug(workspace): file preview doesn't auto-refresh after file mutation, and fileTree+previewArea split into half-screen
## Description

When viewing a file in the Workspace panel preview and the agent modifies that file via tool calls, two problems occur:

1. **Preview content does not update** — the preview shows stale content. The user must close and reopen the file manually.
2. **fileTree and previewArea split into top/bottom half-screen** — the file tree reappears alongside the preview, each taking half the vertical space.

## Steps to Reproduce

1. Open a file in the Workspace Files panel (e.g. `static/style.css`)
2. Send a prompt that causes the agent to modify that file via `write_file`, `patch`, `terminal`, or `execute_code`
3. Wait for the tool call to complete and/or the conversation turn to finish

## Expected Behavior

- The preview should refresh with the updated file content automatically.
- The file tree should remain hidden while a preview is open.

## Actual Behavior

- The preview stays on the old content (stale).
- The file tree reappears, splitting the panel into top (file tree) and bottom (preview) halves.

## Root Cause Analysis

Both problems stem from the `refreshOpenPreviewIfMutated()` → `openFile()` call chain failing, combined with `renderFileTree()` unconditionally restoring `fileTree.style.display`.

### Problem 1: Preview does not refresh

`refreshOpenPreviewIfMutated()` (`workspace.js:482`) checks `_isOpenPreviewPathMutated()`, which compares `_previewCurrentPath` (a workspace-relative path like `foo/bar.py`) against `_turnMutatedPreviewPaths` — a Set populated by `noteWorkspaceMutationsFromToolCall()`. This check fails in three scenarios:

**① Tool name not in the mutation whitelist**

`ARTIFACT_MUTATION_TOOLS` (`workspace.js:393`) only includes:
```js
new Set(["write_file","patch","edit_file","create_file","mcp_filesystem_write_file","mcp_filesystem_edit_file"])
```

`terminal` (modifying files via `sed`/`echo`/shell redirects) and `execute_code` (writing files via Python scripts) are **not** included. File mutations via these tools are invisible to the mutation tracking system.

**② Absolute path vs relative path mismatch**

The file tree opens files with workspace-relative paths (e.g. `foo/bar.py`), stored in `_previewCurrentPath`. However, Hermes agent tools typically use **absolute paths** (e.g. `/Users/huangweihao/workspace/foo/bar.py`). `_normalizeArtifactPath()` (`workspace.js:395`) only strips `~/` and `./` prefixes — it does **not** strip the workspace prefix from absolute paths. The normalized absolute path never equals the relative preview path, so `_turnMutatedPreviewPaths.has()` always returns `false`.

Verified:
```
_normalizeArtifactPath("foo/bar.py")                           => "foo/bar.py"       (preview path)
_normalizeArtifactPath("/Users/huangweihao/workspace/foo/bar.py") => "/Users/huangweihao/workspace/foo/bar.py"  (tool arg)
// These never match.
```

**③ Extension-less files in workspace root are filtered**

`_normalizeArtifactPath()` line 406: `if(!/[./]/.test(path)) return "";` discards paths with no `.` and no `/`. Root-level files like `Makefile` or `Dockerfile` are silently dropped.

### Problem 2: Half-screen split

When a conversation turn completes, `messages.js:5241` calls:
```js
loadDir(".", { preservePreview: true });
```

Inside `loadDir()` (`workspace.js:669`):
1. **Line 690**: `renderFileTree()` runs synchronously → `ui.js:16335` executes `box.style.display = ""`, **restoring `fileTree` to visible**.
2. At this point, `previewArea` still has the `visible` class from the earlier `openFile()` call.
3. Both `.file-tree` (`flex:1`) and `.preview-area.visible` (`flex:1`) are now visible inside `.rightpanel` (`flex-direction:column`), causing them to stack vertically — each taking half the height.
4. **Line 716**: `await refreshOpenPreviewIfMutated()` — if this succeeds, `openFile()` eventually re-hides `fileTree` (line 1049), but there is a visible flicker during the `await api(...)` network round-trip. If this fails (due to Problem 1), `fileTree` is **never** re-hidden, and the half-screen split persists permanently.

The two problems compound: when preview refresh fails (Problem 1), the half-screen layout also becomes permanent (Problem 2), forcing the user to manually close and reopen the preview.

## Environment

- OS: macOS 15.7.4
- Browser: Chrome (WebUI)
- Hermes WebUI: current `main` branch (verified against `static/workspace.js`, `static/ui.js`, `static/messages.js`, `static/style.css`)

## Suggested Fix Directions

1. **Path normalization**: `_normalizeArtifactPath()` should strip the session workspace prefix from absolute paths so they compare equal to relative preview paths. The workspace path is available via `S.session.workspace`.
2. **Tool whitelist**: Add `terminal` and `execute_code` to `ARTIFACT_MUTATION_TOOLS`, or scan tool results for file paths more broadly.
3. **Layout guard**: In `loadDir()` when `preservePreview` is true and a preview is currently open, `renderFileTree()` should not restore `fileTree.style.display` — or `loadDir` should re-hide `fileTree` after rendering if a preview is still active.

NextDoorLaoHuang-HF:
## Problem

The Files page had two bugs:

1. **Preview not auto-refreshing after file mutation** — when an Agent modified a file via `write_file`/`patch` using an absolute path (e.g. `/Users/x/ws/foo/bar.py`), the open file preview stayed stale. The preview uses relative paths (`foo/bar.py`) while tools pass absolute paths, so they never matched.

2. **Half-screen layout split** — when a preview was open and `loadDir` ran with `preservePreview=true`, `renderFileTree()` unconditionally restored `fileTree.style.display=`, making both `fileTree` and `previewArea` visible simultaneously. Both are `flex:1` in a `flex-direction:column` right panel, so they split the panel 50/50.

## Root Cause Analysis

**Root cause 1: Absolute vs relative path mismatch** — `_normalizeArtifactPath()` (workspace.js) only stripped `~/` and `./` prefixes. It did not handle absolute paths containing the workspace prefix, so `/Users/x/ws/foo/bar.py` never became `foo/bar.py` for comparison.

**Root cause 2 (scoped out): Tool whitelist incomplete** — `terminal` and `execute_code` were not in `ARTIFACT_MUTATION_TOOLS`. Per maintainer feedback, these tools' args are shell/script bodies (not structured file paths), so simple whitelist expansion does not work. **This is scoped as a separate follow-up issue.**

**Root cause 3: `renderFileTree` unconditionally restores display** — `loadDir` calls `renderFileTree()` which executes `box.style.display=x27` (ui.js), restoring `fileTree` visibility even when a preview is being preserved.

### Fix 1: `_normalizeArtifactPath` handles absolute paths
After stripping `~/` and `./` prefixes, strips the `S.session.workspace` prefix from absolute paths so `/Users/x/ws/foo/bar.py` → `foo/bar.py`. Absolute paths not under the workspace return `x27` (no false positives). Reuses the proven pattern from `openArtifactPath()`. The strip runs **before** the `[./]` guard to avoid the ordering trap where extension-less root files (Makefile) would be dropped.

### Fix 2: `loadDir` layout guard (方案A)
After `renderFileTree()` (which unconditionally restores `display=x27`), if `preservePreview=true` and `_previewCurrentPath` is set, re-hides `fileTree` so only the preview shows. This guard is independent of Fix 1 — it prevents the half-screen split even if mutation detection fails.

### Not included (per maintainer feedback)
`terminal`/`execute_code` mutation tracking — their args are shell/script bodies, not structured file paths. Diff-fence text mining only catches patch-style unified diffs; `sed`/`echo` redirects remain invisible. Scoped as a separate follow-up issue.

## Test Results

- **10/10** new tests pass (`tests/test_issue5747_workspace_preview_refresh.py`)
- **651 passed, 7 skipped, 0 failed** — workspace test suite
- **230 passed, 1 skipped, 0 failed** — artifact/preview regression tests (includes #3262, #3329, #2655)
- **eslint runtime guard: clean** (0 errors)

## Files Changed

- `static/workspace.js` (+19 lines)
- `tests/test_issue5747_workspace_preview_refresh.py` (new, +265 lines)

### DIFF
--- a/static/workspace.js
+++ b/static/workspace.js
@@ -421,10 +421,7 @@ function _escHtml(s){
 
 const ARTIFACT_IGNORE_RE = /(^|\/)(?:\.git|\.hg|\.svn|node_modules|\.venv|venv|__pycache__|dist|build|\.next|\.cache)(?:\/|$)/;
 // Canonical Hermes mutators plus MCP filesystem aliases that can create/edit files.
-// terminal and execute_code are included because they can modify files (sed, echo >,
-// write_file, open(...,'w')) — their paths are mined heuristically from command/code
-// text in _artifactCandidatesFromToolCall since they lack standard path/file_path args (#5747).
-const ARTIFACT_MUTATION_TOOLS = new Set(['write_file','patch','edit_file','create_file','mcp_filesystem_write_file','mcp_filesystem_edit_file','terminal','execute_code']);
+const ARTIFACT_MUTATION_TOOLS = new Set(['write_file','patch','edit_file','create_file','mcp_filesystem_write_file','mcp_filesystem_edit_file']);
 
 function _normalizeArtifactPath(path){
   if(!path) return '';
@@ -490,17 +487,6 @@ function _artifactCandidatesFromToolCall(tc){
     if(Array.isArray(args.paths)) args.paths.forEach(p=>add(p));
     if(Array.isArray(args.edits)) args.edits.forEach(e=>add(e&&e.path));
   }
-  // terminal/execute_code lack standard path/file_path args — their file targets
-  // are embedded in the command string or Python code. Mine those heuristically
-  // so edits via sed/echo/tee/write_file/open(...,'w') are tracked for preview
-  // refresh (#5747). Only write-modifying patterns are matched; read-only
-  // commands (cat, ls, grep) intentionally produce no candidates.
-  if(name==='terminal' && args && typeof args === 'object' && typeof args.command === 'string'){
-    for(const p of _artifactPathsFromShellCommand(args.command)) add(p, 'terminal');
-  }
-  if(name==='execute_code' && args && typeof args === 'object' && typeof args.code === 'string'){
-    for(const p of _artifactPathsFromPythonCode(args.code)) add(p, 'execute_code');
-  }
   const resultText = typeof result === 'string' ? result : (result ? JSON.stringify(result) : '');
   // Tool results may include unified diffs from patch-style tools; scan those
   // narrowly after structured args so diff headers can still contribute paths.
@@ -509,86 +495,6 @@ function _artifactCandidatesFromToolCall(tc){
     const argsText = typeof args === 'string' ? args : JSON.stringify(args || {});
     for(const a of _artifactCandidatesFromText(argsText)) out.push(a);
   }
-  return out;
-}
-
-// Heuristic extraction of file paths from shell commands that modify files.
-// Matches: sed -i, echo > / >>, cat > / >>, tee, cp, mv — the last token(s)
-// that look like a file path. Read-only commands (cat without redirect, ls,
-// grep, etc.) produce nothing so they don't pollute the mutation set (#5747).
-function _artifactPathsFromShellCommand(command){
-  if(!command || typeof command !== 'string') return [];
-  const out = [];
-  const seen = new Set();
-  // Token that looks like a file path: has a dot or slash, not a flag, not a
-  // shell builtin, and is plausibly a relative/absolute path.
-  const isFilePath = (t) => {
-    if(!t || t.startsWith('-') || t.startsWith('(')) return false;
-    // Must contain a dot (extension) or a slash to look like a path.
-    if(!/[./]/.test(t)) return false;
-    // Reject common shell keywords/builtins.
-    if(/^(?:sudo|bash|sh|zsh|python|python3|node|npm|pip|uv|cd|export|source|echo|cat|sed|awk|grep|find|ls|cp|mv|rm|mkdir|touch|tee|head|tail|wc|sort|uniq|tr|cut|xargs|chmod|chown|git|docker)$/.test(t)) return false;
-    return true;
-  };
-  const tryAdd = (t) => {
-    t = String(t||'').replace(/^['"`]/,'').replace(/['"`]$/,'').replace(/[;,]+$/,'');
-    if(isFilePath(t) && !seen.has(t)){ seen.add(t); out.push(t); }
-  };
-  // Redirect targets: echo "..." > file, cat <<EOF > file, etc.
-  let m;
-  const reRedirect = /(?:>>?)\s*([^\s|;&<]+)/g;
-  while((m = reRedirect.exec(command))) tryAdd(m[1]);
-  // tee file (tee -a file): capture the first non-flag arg after tee.
-  m = /(?:^|[;&|]\s*)tee\s+(?:-[a-zA-Z]+\s+)*([^\s|;&<]+)/.exec(command);
-  if(m) tryAdd(m[1]);
-  // sed -i 'expr' file — the file is the last path-like token (sed -i doesn't
-  // write to stdout; the file arg is modified in-place). -i may appear as
-  // `-i`, `-i.bak`, or `-i''` (GNU sed in-place flag).
-  if(/\bsed\b/.test(command) && /(^|\s)-i\b/.test(command)){
-    const tokens = command.split(/\s+/);
-    // Last token that looks like a file path (skip the expression arg).
-    for(let i = tokens.length - 1; i >= 0; i--){
-      const t = tokens[i].replace(/^['"`]/,'').replace(/['"`;]$/g,'');
-      if(isFilePath(t)){ tryAdd(t); break; }
-    }
-  }
-  // cp/mv src dst — the destination (last path-like token) is the file that
-  // changes (new file for cp, renamed file for mv).
-  m = /(?:^|[;&|]\s*)(?:cp|mv)\s+(?:-[a-zA-Z]+\s+)*(.+)/.exec(command);
-  if(m){
-    const tokens = m[1].trim().split(/\s+/);
-    if(tokens.length >= 2) tryAdd(tokens[tokens.length - 1]);
-  }
-  return out;
-}
-
-// Heuristic extraction of file paths from Python code that modifies files.
-// Matches: write_file('path', ...), open('path', 'w'/'a'/'x'), Path('path').write_text/bytes,
-// and patch(path='path', ...) / patch('path', ...) calls.
-function _artifactPathsFromPythonCode(code){
-  if(!code || typeof code !== 'string') return [];
-  const out = [];
-  const seen = new Set();
-  const tryAdd = (t) => {
-    t = String(t||'').trim();
-    if(t && !seen.has(t)){ seen.add(t); out.push(t); }
-  };
-  let m;
-  // write_file('path', ...) / write_file("path", ...)
-  const reWriteFile = /\bwrite_file\s*\(\s*(['"])([^'"]+)\1/g;
-  while((m = reWriteFile.exec(code))) tryAdd(m[2]);
-  // patch(path='path', ...) / patch('path', ...) / patch(path="path", ...)
-  const rePatch = /\bpatch\s*\(\s*(?:path\s*=\s*)?(['"])([^'"]+)\1/g;
-  while((m = rePatch.exec(code))) tryAdd(m[2]);
-  // edit_file(path='path', ...) / edit_file('path', ...)
-  const reEditFile = /\bedit_file\s*\(\s*(?:path\s*=\s*)?(['"])([^'"]+)\1/g;
-  while((m = reEditFile.exec(code))) tryAdd(m[2]);
-  // open('path', 'w'/'a'/'x')  — mode must include w/a/x to indicate writing
-  const reOpenWrite = /\bopen\s*\(\s*(['"])([^'"]+)\1\s*,\s*(['"])[wax]/g;
-  while((m = reOpenWrite.exec(code))) tryAdd(m[2]);
-  // Path('path').write_text(...) / Path('path').write_bytes(...)
-  const rePathWrite = /Path\s*\(\s*(['"])([^'"]+)\1\s*\)\s*\.write_(?:text|bytes)/g;
-  while((m = rePathWrite.exec(code))) tryAdd(m[2]);
   return out;
 }
 
--- a/static/workspace.js
+++ b/static/workspace.js
@@ -423,6 +423,39 @@ const ARTIFACT_IGNORE_RE = /(^|\/)(?:\.git|\.hg|\.svn|node_modules|\.venv|venv|_
 // Canonical Hermes mutators plus MCP filesystem aliases that can create/edit files.
 const ARTIFACT_MUTATION_TOOLS = new Set(['write_file','patch','edit_file','create_file','mcp_filesystem_write_file','mcp_filesystem_edit_file']);
 
+// A path is Windows-style when it starts with a drive letter ("C:/" or
+// "C:\") or a backslash UNC root ("\\server\share"). A forward-slash "//"
+// prefix is deliberately NOT evidence of Windows — POSIX permits "//" at the
+// start of a path, and mistaking it for UNC would fold legal POSIX file names
+// and mis-strip case-sensitive prefixes (maintainer review #5752, edges 2-3).
+function _isWindowsStylePath(p){
+  return /^[A-Za-z]:[\\/]/.test(p) || /^\\\\/.test(p);
+}
+
+// Strip the S.session.workspace prefix so an absolute tool path and a relative
+// preview path compare equal ("/Users/x/ws/foo.py" vs "foo.py", #5747).
+// Windows workspaces strip case-insensitively (Windows paths are
+// case-insensitive); POSIX keeps exact case. Returns the stripped path when
+// under the workspace, otherwise the input unchanged — callers decide what
+// "outside the workspace" means (mutation tracking fails closed, display
+// keeps the raw path).
+function _stripWorkspacePrefix(path){
+  if(typeof S==='undefined' || !S.session || !S.session.workspace) return path;
+  const ws = String(S.session.workspace);
+  const normWs = ws.replace(/\\/g,'/').replace(/\/+$/,'') + '/';
+  const isWindows = _isWindowsStylePath(ws);
+  // Fold backslashes only for Windows workspaces — on POSIX, `\` is a legal
+  // filename character and must survive (maintainer review #5752, edge 3).
+  const normPath = isWindows ? path.replace(/\\/g,'/') : path;
+  if(isWindows){
+    if(normPath.toLowerCase().startsWith(normWs.toLowerCase()))
+      return normPath.slice(normWs.length);
+  } else if(normPath.startsWith(normWs)){
+    return normPath.slice(normWs.length);
+  }
+  return path;
+}
+
 function _normalizeArtifactPath(path){
   if(!path) return '';
   path = String(path).trim().replace(/[\`"'<>),.;:]+$/g,'').replace(/^[\`"'(<]+/g,'');
@@ -433,24 +466,25 @@ function _normalizeArtifactPath(path){
   // preview stale (#3262 / pre-release regression-gate finding).
   path = path.replace(/^~\//,'').replace(/^(?:\.\/)+/,'');
   if(!path) return '';
-  // Strip workspace prefix from absolute paths so they compare equal to relative
-  // preview paths. Tools like write_file/patch pass absolute paths
-  // ("/Users/x/ws/foo/bar.py") while the preview uses "foo/bar.py" — without
-  // this they never match and the preview stays stale (#5747).
-  // Also handle Windows-style absolute paths (C:\Users\...) — normalize
-  // backslashes to forward slashes first so both platforms match (#5747).
-  const normAbsPath = path.replace(/\\/g,'/');
-  const isAbsPath = normAbsPath.startsWith('/') || /^[A-Za-z]:\//.test(normAbsPath);
-  if(isAbsPath && typeof S!=='undefined' && S.session && S.session.workspace){
-    const normWs = String(S.session.workspace).replace(/\\/g,'/').replace(/\/+$/,'') + '/';
-    if(normAbsPath.startsWith(normWs)) path = normAbsPath.slice(normWs.length);
+  // Windows-style absolutes fold backslashes for matching; POSIX keeps `\` (and
+  // `:`) as legal filename characters, so they must survive verbatim.
+  const isWindows = _isWindowsStylePath(path);
+  const normPath = isWindows ? path.replace(/\\/g,'/') : path;
+  if(isWindows || normPath.startsWith('/')){
+    // Strip the workspace prefix from absolute paths so they compare equal to
+    // relative preview paths (#5747). An absolute path not under the workspace
+    // (or no workspace set) can never match a relative preview path — bail out
+    // to avoid false candidates.
+    const stripped = _stripWorkspacePrefix(normPath);
+    if(stripped === normPath) return '';
+    path = stripped;
   }
-  // An absolute path that was not under the workspace (or no workspace is set)
-  // can never match a relative preview path — bail out to avoid false candidates.
-  if(path.startsWith('/') || /^[A-Za-z]:[\\/]/.test(path)) return '';
   if(!path) return '';
   if(ARTIFACT_IGNORE_RE.test(path)) return '';
-  if(!/[./]/.test(path)) return '';
+  // No bare-name rejection here: structured mutator args and patch headers can
+  // legitimately name extension-less root files (Makefile, LICENSE,
+  // Dockerfile), and rejecting them made valid previews unmatchable
+  // (maintainer review #5752, edge 1).
   return path;
 }
 
@@ -596,11 +630,9 @@ function renderSessionArtifacts(){
     return;
   }
   // Strip workspace prefix for display so long absolute paths don't clutter the list.
-  const ws = S.session && S.session.workspace;
-  const normWs = ws ? ws.replace(/\/+$/,'') + '/' : '';
   const displayPath = (p) => {
-    if(normWs && p.startsWith(normWs)) return p.slice(normWs.length);
-    return p;
+    const rel = _stripWorkspacePrefix(p);
+    return rel !== p ? rel : p;
   };
   const splitArtifactDisplayPath = (path) => {
     const slash = path.lastIndexOf('/');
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
## Summary

Reading the diff against `origin/master` — `static/workspace.js` (+23) and the new node-driven `tests/test_issue5747_workspace_preview_refresh.py` (+307) — this is a clean, well-scoped fix for #5747. Both root causes are addressed: the absolute-vs-relative path mismatch that left the preview stale, and the `renderFileTree()` display restore that split the panel 50/50. The tests drive the *actual* functions extracted from `workspace.js` via node, so they can't drift from a Python mirror. Nice.

## Code reference

The path fix in `_normalizeArtifactPath` (workspace.js:405-420 at the PR head):

```js
const normAbsPath = path.replace(/\\/g,'/');
const isAbsPath = normAbsPath.startsWith('/') || /^[A-Za-z]:\//.test(normAbsPath);
if(isAbsPath && typeof S!=='undefined' && S.session && S.session.workspace){
  const normWs = String(S.session.workspace).replace(/\\/g,'/').replace(/\/+$/,'') + '/';
  if(normAbsPath.startsWith(normWs)) path = normAbsPath.slice(normWs.length);
}
// absolute path not under workspace can never match a relative preview path
if(path.startsWith('/') || /^[A-Za-z]:[\\/]/.test(path)) return '';
```

This is the right layer — every mutation candidate flows through `_normalizeArtifactPath` (called from `_artifactCandidatesFromToolCall` at :453, `noteWorkspaceMutationsFromToolCall` at :480, and `_isOpenPreviewPathMutated` at :494 on master), so a `write_file` arg of `/Users/x/ws/foo.py` and a preview path of `foo.py` now canonicalize identically and `_turnMutatedPreviewPaths.has(current)` fires.

## One observation on consistency (non-blocking)

Your new code is actually *more* robust than the two existing strip sites it mirrors: `openArtifactPath` (workspace.js:588-595) and the artifact-list `displayPath` (workspace.js:567-572) both do a plain case-sensitive `startsWith(normWs)` with no backslash normalization. On Windows those two will still fail to strip a `C:\Users\...\ws\` prefix that your `_normalizeArtifactPath` now handles. Not this PR's job to fix, but worth a follow-up so all three sites share one helper — otherwise the artifact list and the click-to-open path stay Windows-broken while mutation-tracking works. A tiny `_stripWorkspacePrefix(path)` extracted from your new block and reused in all three would close that gap.

## Layout guard

The Fix 2 guard in `loadDir` (workspace.js:733-738) is correctly placed *after* `renderFileTree()` so it counteracts the unconditional `box.style.display=''` restore in `ui.js`:

```js
if(typeof _previewCurrentPath!=='undefined'&&_previewCurrentPath){
  const ft=$('fileTree'); if(ft) ft.style.display='none';
}
```

Keeping the fix in `loadDir` rather than mutating `renderFileTree` is the minimal-blast-radius choice, and the guard test (`test_render_file_tree_still_unconditional_display`) pins that decision so a future refactor doesn't silently move the restore.

## Verification

Per repo policy I don't execute PR-authored tests, but the test structure is sound on read: `test_absolute_path_without_workspace_prefix_stays_empty` and the Windows `..._without_workspace_prefix_stays_empty` case both assert `''` (no false-positive match), and `test_no_workspace_session` covers the `S.session.workspace` unset path — which matters because the early `typeof S!=='undefined'` guard means an absent workspace falls straight through to the `return ''` bail-out at :419 rather than throwing. Deferring the merge call to a maintainer.

NextDoorLaoHuang-HF:
The `test (3.13, 0)` failure is **not code-related** — it's a flaky GitHub Actions runner issue. The log shows:

```
E: Failed to fetch https://packages.microsoft.com/repos/azure-cli/dists/noble/InRelease
Failed to install browsers
Error: Installation process exited with code: 100
```

The browser installation (`apt-get update`) failed due to a transient network issue with the Microsoft Azure CLI apt repository on the runner. All other 14 test jobs passed, including `test (3.13, 1/2/3/4)`, `browser-smoke`, `lint`, and `Greptile Review`.

Would appreciate a rerun of the failed job. 🙏

nesquena-hermes (changes requested):
Thanks @NextDoorLaoHuang-HF — the two core wins here are solid and verified: the preview auto-refreshes when a tool edits the open file, and the half-panel split on directory reload is fixed (the `loadDir` file-tree re-hide is correct). The blocker is the absolute→relative path canonicalization in `_normalizeArtifactPath()`, which has a long tail of cross-platform edges. I took a couple of maintainer passes at it during gating (POSIX-file-name safety, then Windows UNC support), but the gate keeps surfacing new path-flavor edges, so this needs a proper platform-aware rewrite from you rather than more incremental patches. Here's the complete consolidated spec of everything the gate found:

## Path-canonicalization edges to resolve (all verified vs origin/master)

1. **Extensionless workspace-root files disappear from Artifacts** (`static/workspace.js` bare-name `[./]` guard). After the prefix-strip, an absolute `<ws>/Makefile` / `LICENSE` / `Dockerfile` becomes a bare `Makefile`, which the existing `if(!/[./]/.test(path)) return ''` guard then rejects — so these files that were valid Artifacts on master vanish. **Fix:** drop the bare-name rejection for these (candidates already come from structured mutator args / patch headers, so the guard is over-defensive here).

2. **Platform inference from the workspace string is ambiguous.** Detecting Windows by "starts with a drive letter OR leading double-slash" misclassifies a POSIX workspace path that legitimately begins with `//` as Windows (dropping a valid `C:/foo.py`, or mapping a case-different POSIX absolute to the wrong preview). **Fix:** don't treat a generic forward-slash `//` as proof of Windows — use an explicit path-flavor signal (ideally a server-provided platform/flavor hint on the session/workspace, since the client can't reliably infer POSIX-vs-Windows from the path string alone), or restrict UNC detection to canonical *backslash* UNC roots (`\\server\share`).

3. **POSIX must preserve `\` and `:` as legal filename characters** — a file literally named `a\b.py` or `C:/foo.py` on a POSIX workspace must stay a relative path (not be folded/dropped). (This one my pass got right; keep it.)

4. **Windows drive-letter + UNC absolutes** must strip the workspace prefix case-insensitively and fail closed when the path is outside the workspace. (Also handled in my pass; keep it.)

The cleanest resolution is probably a small server-provided `platform`/`path_flavor` field on the session so the client stops guessing from the path string — that removes the ambiguity in #2 permanently. If you'd rather keep it client-side, restrict Windows detection to drive-letter + backslash-UNC only and drop the `[./]` guard per #1.

Everything else verified clean (17 PR tests + 11 neighboring artifact/preview tests pass, no XSS/traversal, no perf regression, the re-hide guard is race-safe). Once the canonicalization handles the platform flavors unambiguously + keeps extensionless root files, re-request and I'll re-run the full gate. Appreciate the fix — it's the path edge-cases, not the core feature, holding it.

webtecnica:
## Review + complementary PR

Thanks for this — the workspace-prefix normalization (with Windows path handling and the ordering trap around the `[./]` guard) and the `loadDir` layout guard are solid, well-tested, and exactly the right fixes for the two root causes you addressed. I verified the logic against `static/workspace.js` and the approach matches the issue's suggested directions 1 and 3.

Since you scoped `terminal`/`execute_code` mutation tracking as a follow-up (maintainer feedback: plain whitelist expansion doesn't work because their args are shell/script bodies), I opened **#6691** as a complement covering exactly that — text-scanning those tools' args + result for the open preview's path (path-like token extraction + direct basename mention), plus `{allowBare:true}` in `_normalizeArtifactPath()` for structured args so extension-less root files (`Makefile`/`Dockerfile`) stay tracked. Both PRs are additive and non-conflicting.

Not a blocker, just a note: when both land, `Makefile` in the workspace root will finally auto-refresh too.

NextDoorLaoHuang-HF:
@nesquena-hermes Thanks for the detailed consolidated spec — pushed `38567acc` addressing all four edges plus the consistency note. The gate should re-run clean.

## What changed

**`_normalizeArtifactPath` is now platform-aware via two extracted helpers:**

- **`_isWindowsStylePath(p)`** — recognizes only drive-letter (`C:/` or `C:\`) and **backslash** UNC roots (`\\server\share`). A forward-slash `//` prefix is deliberately NOT Windows evidence (edge 2): POSIX permits `//` at path start, so `//server/share/foo.py` on a POSIX workspace fails closed (`''`) instead of being folded or mis-stripped.

- **`_stripWorkspacePrefix(path)`** — single platform-aware strip used by all three sites (edge "consistency note"): `_normalizeArtifactPath`, `renderSessionArtifacts`' `displayPath`, and `openArtifactPath`. Windows workspaces fold backslashes and strip case-insensitively; POSIX workspaces strip exact-case and **never fold `\`** — a file literally named `a\b.py` survives as `a\b.py` (edge 3).

- **Bare-name guard removed** (edge 1): the old `if(!/[./]/.test(path)) return ''` dropped extension-less root files after prefix-stripping, so `Makefile`/`LICENSE`/`Dockerfile` previews could never match a mutation. Structured mutator args and patch headers legitimately name them, so they now normalize to the bare name and stay trackable.

- **Fail-closed preserved** (edge 4): any absolute path whose prefix does not strip (outside workspace, no workspace set, or POSIX `//` mismatch) returns `''` — no false candidates.

## Tests

- 10 new node-driven tests in `test_issue5747_workspace_preview_refresh.py` covering: extension-less root files (absolute, bare-relative, and via `write_file` candidate), POSIX `//`-prefix fail-closed, POSIX backslash survival, UNC strip, UNC case-insensitive strip, drive-letter case-insensitive strip, Windows outside-workspace fail-closed, and `_stripWorkspacePrefix` return-unchanged semantics.
- The `#3262`/`#3329` test drivers extract functions from `workspace.js` and now pull the two new helpers too (they were the only other files driving `_normalizeArtifactPath`).
- Results: **23/23** in the 5747 file, **28/28** with 3262+3329, **643 passed / 1 skipped** across the full workspace/preview/artifact test selection, **eslint runtime guard clean**.

One note on edge 2: a POSIX path that legitimately starts with `//` AND sits under a `//`-prefixed workspace still fails closed (never matches). If you'd prefer full `//`-collapse semantics there, the server-provided `platform` hint from your review would be the right long-term fix — happy to pick that up as a follow-up if you want it.

nesquena-hermes (changes requested):
Thanks for the substantial rework. This head fixes several items from the consolidated review: extensionless root files now survive, forward `//` is no longer treated as Windows evidence, native drive/UNC prefixes strip case-insensitively, and the three strip call sites share a helper. The focused PR tests are green.

I re-gated exact head `38567acc8d69321f03084070a5c8ee5ec18905fc`, and the platform inference still leaves three deterministic blockers. I ran reviewer-authored probes against the actual extracted JavaScript through the repository's no-network sandbox. Result: **3 failed**. The PR's own focused files remain **26 passed**.

## Blocking cases

1. **A legal POSIX relative name that looks like a Windows drive path is still dropped.**

   With workspace `/Users/test/ws`, the absolute tool path `/Users/test/ws/C:/foo.py` first normalizes to `C:/foo.py`. The same value is normalized again by `noteWorkspaceMutationsFromToolCall()` and `collectSessionArtifacts()`. On that second pass, `_isWindowsStylePath('C:/foo.py')` treats the relative POSIX name as an absolute Windows path and returns `''` because it cannot strip the POSIX workspace prefix.

   Repro output:

   ```text
   ['C:/foo.py', ''] != ['C:/foo.py', 'C:/foo.py']
   ```

   This is the exact POSIX `C:/foo.py` ambiguity named in the prior review, so that finding has not converged end to end.

2. **An absolute path that lexically starts under the workspace but resolves outside it does not fail closed.**

   `_stripWorkspacePrefix('/Users/test/ws/../outside/foo.py')` returns `../outside/foo.py`, and `_normalizeArtifactPath()` accepts it. The stated outside-workspace contract requires this to return `''`.

   ```text
   ['../outside/foo.py'] != ['']
   ```

3. **A legal backslash in the POSIX workspace root is folded before flavor is determined.**

   `_stripWorkspacePrefix()` builds `normWs` with `ws.replace(/\\/g, '/')` before checking `isWindows`. Workspace `/tmp/a\b` therefore cannot strip absolute `/tmp/a\b/file.py` even though both are valid POSIX strings.

   ```text
   [''] != ['file.py']
   ```

## Required fix

In `static/workspace.js`:

- Determine path flavor from the session/workspace, not independently from every candidate. A POSIX workspace must treat relative `C:/foo.py` and leading-backslash lookalikes as legal POSIX names. Prefer an explicit server-provided `path_flavor: 'posix' | 'windows'` if the client cannot disambiguate all representations reliably.
- In `_stripWorkspacePrefix()`, determine flavor before normalizing. Fold separators and case only for Windows workspaces. Do not rewrite backslashes in a POSIX workspace string.
- Reject or safely canonicalize `.`/`..` segments before accepting a stripped absolute path. Any absolute input that resolves outside the workspace must return `''`.

Please add actual-function tests for:

- `/Users/test/ws/C:/foo.py` and relative `C:/foo.py`, including the composed candidate → mutation-set → open-preview path so repeated normalization is covered;
- a POSIX workspace root containing a literal backslash;
- POSIX and Windows `workspace/../outside` escapes;
- sibling-prefix boundaries such as `/ws2` vs `/ws`;
- display/open parity for accepted drive and UNC descendants.

The preview-refresh and layout fixes themselves remain sound. This request is limited to finishing the path-flavor boundary without another syntax-guessing edge swap.

NextDoorLaoHuang-HF:
@nesquena-hermes Thanks for the precise re-gate — pushed `ea3c9aa6` addressing all three blockers. The core change is exactly your "determine flavor from the workspace, not per candidate" instruction.

## Blocker 1 — drive/UNC lookalikes are legal POSIX names

Path flavor is now decided **once from the workspace string** (`_isWindowsStylePath(String(S.session.workspace))`), never from the candidate. Under a POSIX workspace, `C:/foo.py` and leading-backslash names never enter a Windows code path, so:

- `/Users/test/ws/C:/foo.py` → `C:/foo.py` on the first pass, and a second pass over the same value yields `C:/foo.py` again (idempotent — your composed candidate → mutation-set → open-preview chain is covered by `test_posix_drive_lookalike_relative_survives_repeated_normalization` plus a `write_file` candidate test).
- Windows absolutes are only recognized when the **workspace** is Windows-style; drive/UNC paths under a Windows workspace still fold + strip case-insensitively.

## Blocker 2 — dot-dot escapes fail closed

New `_canonicalizeRelativePath()` resolves `.`/`..` segments of the normalized value before it is accepted:

- `/Users/test/ws/../outside/foo.py` → leading `..` → `''` (fail closed)
- `/Users/test/ws/a/../../x.py` → `''`
- `/Users/test/ws/sub/../x.py` → `x.py` (in-workspace segments canonicalize so they still match the preview)

Covered for both flavors: `test_posix_dotdot_escape_fails_closed`, `test_windows_dotdot_escape_fails_closed`, `test_in_workspace_dotdot_canonicalizes`.

## Blocker 3 — POSIX workspace root with literal backslash

`_stripWorkspacePrefix()` now decides flavor **before** normalizing: the workspace string is only rewritten when it is Windows-style. `/tmp/a\b/file.py` under workspace `/tmp/a\b` strips to `file.py` (`test_posix_workspace_root_with_literal_backslash`).

## Also added

- `test_sibling_prefix_workspace_not_confused` — `/ws2` vs `/ws` both directions.
- `test_strip_workspace_prefix_windows_drive_and_unc_descendants` — display/open parity for drive + UNC descendants (the shared helper both call).

## Results

- 36/36 in the three node-driven files (5747 + 3262 + 3329)
- **651 passed / 1 skipped** across the full workspace/preview/artifact selection
- eslint runtime guard clean

On your `path_flavor` server hint: with flavor derived from the workspace string the client no longer guesses per candidate, so the ambiguity from the first review is closed client-side. A server-provided field would still be the most robust long-term answer (e.g. a `//`-prefixed POSIX workspace colliding with a drive-lookalike name inside it — both currently resolve fine, but a hint removes the inference entirely). Happy to file that as a follow-up issue rather than scope it into this PR.

nesquena-hermes (changes requested):
The latest re-push fixes the three blockers from the 06:34 UTC review: path flavor now comes from the workspace, POSIX workspace backslashes are preserved, and root-escaping `..` segments fail closed. One fail-closed boundary is still open in the exact-head control flow.

### Must fix: reject absolute Windows paths when workspace authority is absent

In `static/workspace.js::_normalizeArtifactPath()`, `isWindows` is forced false when `S.session.workspace` is absent:

```js
const ws = (typeof S!=='undefined' && S.session) ? S.session.workspace : null;
const isWindows = !!ws && _isWindowsStylePath(String(ws));
```

The following branch then rejects only leading-slash POSIX inputs. Unambiguous Windows absolutes such as `C:/foo.py`, `C:\foo.py`, and `\\server\share\foo.py` reach `_canonicalizeRelativePath()` and survive as non-empty workspace-relative mutation candidates. The submitted no-workspace regression covers only `/Users/test/ws/foo.py`, so it misses this discriminator.

The same boundary has two related Windows-workspace cases:

- A single-leading-backslash rooted path is folded into `/outside/file` through `normPath`, but rooted detection checks the pre-fold `path.startsWith('/')`, so it is accepted as relative.
- A drive-relative input such as `C:foo.py` is also accepted as workspace-relative.

Please make the authority rule explicit:

1. With no workspace, reject every unambiguous absolute flavor before relative canonicalization (`path.startsWith('/') || _isWindowsStylePath(path)`).
2. In the Windows-workspace arm, perform rooted detection on `normPath` (or recognize one leading backslash explicitly), and reject drive-relative `^[A-Za-z]:[^\\/]` inputs.
3. Preserve the intended POSIX-workspace behavior where drive-looking names and literal backslashes can be legal relative names.
4. Add exact-output regressions for no-workspace drive-forward, drive-backslash, and UNC inputs; Windows-workspace root-relative and drive-relative inputs; plus ordinary relative controls. Please also add one production-composed mutation-to-preview-refresh row so this helper contract is proven through the real caller chain.

Security/testing note: the exact-head diff received a CLEAN static threat verdict from the locally fetched `origin/master...ea3c9aa6` patch. The mandatory wrapper's own scan could not fetch the same patch because GitHub returned HTTP 403 rate-limit exceeded, so it failed closed before sandbox entry and no tests were executed. This review is therefore based on the deterministic exact-head source path, not a claimed test failure.

NextDoorLaoHuang-HF:
@nesquena-hermes Thanks for the sharp catch — pushed `6efbcc76` making the authority rule explicit. All four of your points are in:

## 1. No workspace → reject every unambiguous absolute

```js
if(!ws){
  if(path.startsWith('/') || _isWindowsStylePath(path)) return '';
}
```

`C:/foo.py`, `C:\foo.py`, and `\\server\share\foo.py` all fail closed before relative canonicalization. A bare `\foo.py` remains ambiguous on POSIX and survives as a relative name, per your spec.

## 2. Windows-workspace arm → rooted detection on the folded form

```js
const rooted = _isWindowsStylePath(path) || normPath.startsWith('/')
  || /^[A-Za-z]:[^\\/]/.test(path);
```

- `\foo.py` folds to `/foo.py`, so the root-relative case is now caught (previously the pre-fold `path.startsWith('/')` missed it).
- `C:foo.py` (drive-relative, no separator) is rejected by `^[A-Za-z]:[^\\/]`.

## 3. POSIX-workspace behavior preserved

The `!ws` guard is the only new rejection path — under a real POSIX workspace, `C:/foo.py`, `C:foo.py`, and literal backslashes still pass through as legal relative names (existing tests unchanged and green).

## 4. Regressions added (all actual-function, node-driven)

- No-workspace: `C:/foo.py`, `C:\foo.py`, `\\server\share\foo.py` → `''`
- No-workspace controls: `foo.py`, `sub/x.py`, `foo\bar.py` → survive
- Windows-workspace: `\foo.py`, `C:foo.py` → `''`; controls `foo.py`, `foo\bar.py` → `foo.py` / `foo/bar.py`
- **Production-composed rows**: driving `noteWorkspaceMutationsFromToolCall` → `_isOpenPreviewPathMutated` with a stubbed preview path — a no-workspace `write_file` on `C:/foo.py` leaves the mutation set empty (`openMutated: false`), while a valid workspace mutation flows through and marks the preview (`mutated: ["foo.py"], openMutated: true`).

Results: **37/37** in the 5747 file, **657 passed / 1 skipped** across the full workspace/preview/artifact selection, eslint runtime guard clean. The preview-refresh and layout fixes remain untouched.

Re: the 403 on the wrapper's patch fetch — no worries, the deterministic source-path review is exactly why the tests are node-driven from `static/workspace.js` directly.

nesquena-hermes (changes requested):
Thanks, this re-push closes the production-code findings from the previous review.

Verified at exact head `6efbcc7661b9c854c1773ee7ecbc3fdf3be7229c`:

- no-workspace drive-absolute and UNC paths now fail closed;
- Windows root-relative and drive-relative paths now fail closed;
- the existing POSIX backslash, workspace-flavor, traversal, extensionless-root-file, and valid-relative behavior remains intact;
- the mandatory sandboxed target is green: **37 passed**.

One acceptance item from the prior fix-spec is still incomplete, so this remains changes-requested for a narrow test-only follow-up.

### Must fix: carry the composed regression through the real refresh sink

`tests/test_issue5747_workspace_preview_refresh.py::_note_mutations_via_node()` extracts and executes the real candidate/normalization/mutation functions through `_isOpenPreviewPathMutated()`, but stops there. It does not extract or execute `refreshOpenPreviewIfMutated()`, and it never observes the final `openFile(path, {bustCache:true})` call.

Please extend that production-function harness (or add an equivalent one) through the complete chain named in the previous review:

`_artifactCandidatesFromToolCall()` → `noteWorkspaceMutationsFromToolCall()` → `_isOpenPreviewPathMutated()` → `refreshOpenPreviewIfMutated()` → `openFile(..., {bustCache:true})`

Use the real functions with `_previewDirty=false` and an `openFile` spy, then assert:

1. a no-workspace Windows-absolute mutation produces **zero** `openFile` calls; and
2. a valid relative mutation produces **exactly one** `openFile("foo.py", {bustCache:true})` call.

The existing composed row is useful and behaviorally exercises production JS, but its `openMutated` boolean proves only the predicate. The final refresh/open sink is the part still owed. No further production-code change is requested by this review.

NextDoorLaoHuang-HF:
@nesquena-hermes Pushed `673bb81b` — the composed regression now runs through the complete sink chain.

## What changed (test-only)

New `_refresh_chain_via_node()` harness drives the full production chain:

`_artifactCandidatesFromToolCall()` → `noteWorkspaceMutationsFromToolCall()` → `_isOpenPreviewPathMutated()` → `refreshOpenPreviewIfMutated()` → `openFile` spy

- `refreshOpenPreviewIfMutated()` executes as the **real extracted async function** with `_previewDirty=false`; only `openFile` is spied (it touches the DOM).
- One harness fix: `_extract_fn` now preserves an `async` prefix when the source declares one, so extracted bodies containing `await` stay legal — `refreshOpenPreviewIfMutated` is the first async function driven this way.

## Assertions

1. **No-workspace Windows-absolute mutation** (`write_file` on `C:/foo.py`, preview open on `foo.py`) → **zero** `openFile` calls — it never becomes a candidate, never marks the preview mutated, never reaches the sink.
2. **Valid workspace mutation** (`write_file` on `/Users/test/ws/foo.py`) → **exactly one** `openFile("foo.py", {bustCache: true})` call.

The predicate-level composed rows from the previous round are kept as-is.

Results: **39/39** in the 5747 file, **659 passed / 1 skipped** across the full workspace/preview/artifact selection, eslint runtime guard clean.

Issue #3412, opened 2026-06-02
enhancement: MiniMax-M3 missing from model switcher — make MiniMax catalog auto-refresh via models.dev instead of hand-maintained static lists


nesquena-hermes:
## Summary

MiniMax released **MiniMax-M3** on 2026-06-01. A user (Deor, via Discord) can't find it in the model switcher and hit a "connection error" trying to use it as their main model. Root cause: the WebUI's MiniMax catalog is **hand-maintained static lists** that have to be edited for every MiniMax release. The broader fix is to let Hermes pick new MiniMax models up automatically (the same way it already does for DeepSeek/Z.AI/Mistral/etc. via models.dev), instead of us patching a static list each time.

> hey quick question regarding new released model minimax-m3, currently I can't find it on the model switcher, is it expected and I should wait for an update or no? because it says connection error while I use MiniMax-M3 from minimax provider as main model — Deor

## Investigation (verified against master @ 4baa26b, running v0.51.210 + bundled hermes-agent)

There are **two** independent freshness layers for MiniMax, and both are manual:

**1. WebUI static catalog** — `api/config.py`
- Dropdown registry, `:682-683` — only `MiniMax-M2.7` + `MiniMax-M2.7-highspeed`.
- Per-provider static list `STATIC_PROVIDER_MODELS["minimax"]`, `:1109-1115` — `M2.7 / M2.7-highspeed / M2.5 / M2.5-highspeed / M2.1`. **No M3.**
- `minimax-cn`, `:1116-1121` — `M2.7 / M2.5 / M2.1 / M2`. **No M3.**

**2. hermes-agent catalog** — `hermes_cli/models.py`
- `_PROVIDER_MODELS["minimax"]` (`:297`) *does* list `MiniMax-M3` in the current bundled agent, which is why on this box `/api/models/live?provider=minimax` already returns M3 and the dropdown gets enriched. But that's still a hand-edited static list — a user on an older bundled agent won't have it.

### Why "I can't find it" + the live-enrichment gap

[...]

PR #5761, opened 2026-07-08
fix(models): refresh OpenCode Go static catalog

sbe27:
## Thinking Path

- Hermes WebUI keeps OpenCode Go on a curated static model list because the old generic live probe could surface models that were not actually enabled on the Go tier.
- That safety choice makes the static list the picker source of truth for `opencode-go`.
- The current OpenCode Go docs and documented models endpoint now include newer Go-tier models that the WebUI list did not include.
- Refreshing the static list restores GLM-5.2, MiniMax M3, and the other current documented Go models without reintroducing the unsafe generic probe.
- Hy3 Preview is intentionally excluded from this Go picker snapshot; OpenCode Zen now exposes the non-preview free model as `hy3-free`.

## What Changed

- Updated `api/config.py` so `_PROVIDER_MODELS["opencode-go"]` tracks the current public OpenCode Go model set.
- Added missing current Go-tier models including GLM-5.2, MiniMax M3, Kimi K2.7 Code, and Qwen3.7 Max/Plus.
- Excluded `hy3-preview` from the Go picker snapshot now that OpenCode Zen exposes the non-preview free model as `hy3-free`.
- Tightened `tests/test_issue5311_opencode_go_static_models.py` from a loose non-empty-list assertion to an exact snapshot check, plus label checks for recently-added models.

## Why It Matters

Users with OpenCode Go could not select current Go-tier models like GLM-5.2 and MiniMax M3 from the WebUI picker even though OpenCode documents and serves them. This keeps the provider-scoped safety fix from #5311/#5611 while making the curated list current again and avoiding stale preview-only entries.

Release-note wording: Refreshes the OpenCode Go model picker list to include the current documented Go-tier models, including GLM-5.2 and MiniMax M3.

## Verification

- ✅ `./scripts/test.sh tests/test_issue5311_opencode_go_static_models.py`
  - `3 passed`
- ✅ `python3 scripts/ruff_lint.py --diff origin/master`
  - no new violations on added/modified lines
- ✅ Probed the documented OpenCode Go and OpenCode Zen models endpoints; confirmed `hy3-preview` remains only a preview Go endpoint entry while Zen now exposes `hy3-free`.

## Risks / Follow-ups

- The list is still a curated static snapshot. If OpenCode Go adds/removes models again, this file must be refreshed unless a provider-specific safe live fetch replaces the static snapshot.
- Related broader follow-up: #3412 tracks avoiding hand-maintained static model freshness issues for MiniMax/model catalogs generally.
- No UI layout or interaction flow changed, so no before/after screenshots are included.
- No docs were changed because this only refreshes provider catalog data and regression coverage.

## Model Used

AI-assisted.

- Provider: OpenAI Codex
- Model: gpt-5.5
- Notable tool use: repository inspection, GitHub issue/PR search via `gh`, OpenCode documentation extraction, documented endpoint probing, targeted patching, and local test/lint verification.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -1759,22 +1759,29 @@ _PROVIDER_MODELS = {
         {"id": "nemotron-3-super-free", "label": "Nemotron 3 Super Free"},
         {"id": "big-pickle", "label": "Big Pickle"},
     ],
-    # OpenCode Go — flat-rate models via opencode.ai/go ($10/month)
+    # OpenCode Go — flat-rate models via opencode.ai/go ($10/month).
+    # Synced 2026-07-08 from the public Go docs and documented models endpoint.
+    # Keep preview/free-only Zen models out of this Go picker snapshot.
     "opencode-go": [
+        {"id": "minimax-m3",       "label": "MiniMax M3"},
+        {"id": "minimax-m2.7",     "label": "MiniMax M2.7"},
+        {"id": "minimax-m2.5",     "label": "MiniMax M2.5"},
+        {"id": "kimi-k2.7-code",   "label": "Kimi K2.7 Code"},
+        {"id": "kimi-k2.6",        "label": "Kimi K2.6"},
+        {"id": "kimi-k2.5",        "label": "Kimi K2.5"},
+        {"id": "glm-5.2",          "label": "GLM-5.2"},
         {"id": "glm-5.1",          "label": "GLM-5.1"},
         {"id": "glm-5",            "label": "GLM-5"},
-        {"id": "kimi-k2.5",        "label": "Kimi K2.5"},
-        {"id": "kimi-k2.6",        "label": "Kimi K2.6"},
         {"id": "deepseek-v4-pro",  "label": "DeepSeek V4 Pro"},
         {"id": "deepseek-v4-flash","label": "DeepSeek V4 Flash"},
+        {"id": "qwen3.7-max",      "label": "Qwen3.7 Max"},
+        {"id": "qwen3.7-plus",     "label": "Qwen3.7 Plus"},
+        {"id": "qwen3.6-plus",     "label": "Qwen3.6 Plus"},
+        {"id": "qwen3.5-plus",     "label": "Qwen3.5 Plus"},
         {"id": "mimo-v2-pro",      "label": "MiMo V2 Pro"},
         {"id": "mimo-v2-omni",     "label": "MiMo V2 Omni"},
         {"id": "mimo-v2.5-pro",    "label": "MiMo V2.5 Pro"},
         {"id": "mimo-v2.5",        "label": "MiMo V2.5"},
-        {"id": "minimax-m2.7",     "label": "MiniMax M2.7"},
-        {"id": "minimax-m2.5",     "label": "MiniMax M2.5"},
-        {"id": "qwen3.6-plus",     "label": "Qwen3.6 Plus"},
-        {"id": "qwen3.5-plus",     "label": "Qwen3.5 Plus"},
     ],
     # 'gemini' is the hermes_cli provider ID for Google AI Studio
     # Model IDs are bare — sent directly to:
--- a/tests/test_issue5311_opencode_go_static_models.py
+++ b/tests/test_issue5311_opencode_go_static_models.py
@@ -1,30 +1,72 @@
-"""#5311 / #5611: OpenCode Go's live /v1/models probe returns public-catalog
-models not enabled on the Go tier, so selecting one 404s on send. get_available_models
-must skip the live probe for opencode-go and fall through to the curated static
-_PROVIDER_MODELS list instead.
+"""#5311 / #5611: OpenCode Go's generic live probe used to return public-catalog
+models not enabled on the Go tier, so the picker intentionally falls back to a curated
+static _PROVIDER_MODELS list. Keep that curated list in sync with the public Go docs
+and documented Go models endpoint while excluding preview/free-only Zen models.
 """
 
+import ast
 from pathlib import Path
 
 ROOT = Path(__file__).resolve().parent.parent
-CONFIG = (ROOT / "api" / "config.py").read_text(encoding="utf-8")
+CONFIG_PATH = ROOT / "api" / "config.py"
+CONFIG = CONFIG_PATH.read_text(encoding="utf-8")
+
+EXPECTED_OPENCODE_GO_MODEL_IDS = [
+    "minimax-m3",
+    "minimax-m2.7",
+    "minimax-m2.5",
+    "kimi-k2.7-code",
+    "kimi-k2.6",
+    "kimi-k2.5",
+    "glm-5.2",
+    "glm-5.1",
+    "glm-5",
+    "deepseek-v4-pro",
+    "deepseek-v4-flash",
+    "qwen3.7-max",
+    "qwen3.7-plus",
+    "qwen3.6-plus",
+    "qwen3.5-plus",
+    "mimo-v2-pro",
+    "mimo-v2-omni",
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Read the diff on `api/config.py` and `tests/test_issue5311_opencode_go_static_models.py`, plus the surrounding provider wiring. Two things check out cleanly, and there's one cross-repo divergence worth confirming before merge.

## Safety invariant preserved ✔

The #5311/#5611 fix is intact — the picker still skips the generic live probe for this provider and falls through to the curated static list. `get_available_models` at `config.py:7010`:

```python
elif pid == "opencode-go":
    # Skip live /v1/models probe for OpenCode Go — it
    # returns models from the public catalog that are
    # not enabled on the Go tier, causing 404 when
    # selected. Use the curated static list only. (#5311)
    pass
```

The branch is still a bare `pass`, so `_models_from_live_provider_ids` is never reached for `opencode-go`. Refreshing the static list doesn't touch that path. Good.

## Test tightening is robust ✔

Swapping the loose `block.count('"id":') >= 5` for an AST-parsed exact snapshot (`_opencode_go_static_models()` → `EXPECTED_OPENCODE_GO_MODEL_IDS`) is the right call — it parses the source via `ast.literal_eval` rather than the runtime dict, so it won't be perturbed by the runtime core-seeder (`config.py:1855`) that injects missing IDs from the agent. That keeps the snapshot assertion stable even though `_PROVIDER_MODELS` gets mutated at import time.

## Cross-repo divergence: `hy3-preview`

I can't hit `opencode.ai/zen/go/v1/models` to confirm the snapshot itself, but I did diff it against the agent core's own opencode-go list at `hermes_cli/models.py:439-458`. The core lists **19** Go models; this PR's WebUI list has **20**. The set relationship is: the core list is a strict subset of this PR's list, and the one WebUI-only extra is `hy3-preview`.

In the agent core, `hy3-preview` isn't a Go model at all — it's filed under a different provider (`tencent-tokenhub`, `models.py:369-371`). The routing helper `opencode_model_api_mode` (`models.py:3400`) only special-cases `minimax-*` and `qwen*` for Go and lets everything else fall to `chat_completions`, so `hy3-preview` wouldn't be rejected at route time — but that's exactly the shape #5311 warns about: a picker entry that sends fine structurally yet 404s if the model isn't actually enabled on the Go tier.

Recommendation: confirm `hy3-preview` really is served on Go (not just present in some shared catalog) before merge. If the documented endpoint genuinely lists it, then the agent core's `_PROVIDER_MODELS["opencode-go"]` is the stale copy and should be synced in the same change so the two lists don't drift — otherwise the WebUI picker offers a model the agent's own curated Go list doesn't recognize. The other 19 IDs match the core exactly, so this is the only entry I'd double-check.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (Codex SAFE + suite) · data-only catalog refresh · T1 for Nathan

**Certified head:** `sha:68df46a6` (clean rebase, branch `gate-rebase/5761-opencode-go-catalog`) · **PR:** #5761 · _sbe27, fix(models): refresh OpenCode Go static catalog_
**Verdict:** GREEN. A clean, curated refresh of the bundled OpenCode Go static model catalog (data-only). Codex SAFE, full suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5761`) — Codex + full suite (data-only; no visible-layout risk)
| Gate | Result |
|------|--------|
| Codex (structure / dup-ids / loader-compat / curated) | **SAFE TO SHIP** — 0 findings |
| Full pytest suite | ✅ **12324 passed, 0 failed** (+ 3 catalog tests) |

### Findings
**✅ Clean data refresh (+72/-21):** updates the OpenCode Go flat-rate model list in `api/config.py` — adds newer models (minimax-m3, kimi-k2.7-code, glm-5.2, qwen3.7-max/plus, hy3-preview) and reorders/keeps the existing ones (nothing actually dropped — the diff's `-` lines are repositions), with a dated `Synced 2026-07-07` comment citing the source endpoint (opencode.ai/zen/go/v1/models). Codex confirmed: valid list-of-dicts matching the loader's `id`+`label` shape, no duplicate ids, loader/consumers unaffected (no code depends on a removed id), the `test_issue5311` invariants hold, no injection/odd chars in ids/labels, and the static list stays a curated allowlist (the #5311/#5611 static-fallback used when the live probe returns public-catalog noise) rather than a raw dump. Full suite green.

### Recommendation to Nathan
**GREEN — merge from branch `gate-rebase/5761-opencode-go-catalog` (`sha:68df46a6`).** A low-risk curated data refresh of the OpenCode Go model catalog — Codex SAFE + full suite green. It surfaces in the model picker (as catalog entries), so it's parked T1 for your nod, but it's a near-autonomous-shippable data-only change. concept 4/5. Author @sbe27. crit=3, model catalog.

---
_Gate-certifier layer (warm-up → **gate** → release). OpenCode Go static catalog refresh: adds minimax-m3/kimi-k2.7-code/glm-5.2/qwen3.7/hy3-preview + reorders, nothing dropped, dated sync comment cites source. Codex SAFE (valid id+label structure, no dup ids, loader-compat, nothing code-depended-on removed, curated allowlist preserved, no injection) + full suite green (0 failed, 12324) + 3 catalog tests. Data-only (no visible-layout risk). Cert valid for sha:68df46a6._

sbe27:
Force-push update: removed `hy3-preview` / Hy3 Preview from the OpenCode Go static picker snapshot. Confirmed OpenCode Zen now exposes the non-preview free model as `hy3-free`, while the current public Go docs do not list Hy3 as part of the Go model set.

nesquena-hermes:
Confirmed the force-push resolves the `hy3-preview` divergence I flagged earlier. Pulled the current head (`7bcc0189d`) into a read-only worktree and diffed the `opencode-go` block in `api/config.py:1765-1785` against the agent core's own list.

## The two catalogs now match exactly ✔

Sorted set comparison of WebUI `config.py:1766-1784` vs agent core `hermes_cli/models.py:440-458` — identical, 19 IDs each, zero divergence:

```
deepseek-v4-flash  deepseek-v4-pro  glm-5  glm-5.1  glm-5.2
kimi-k2.5  kimi-k2.6  kimi-k2.7-code  mimo-v2.5  mimo-v2.5-pro
mimo-v2-omni  mimo-v2-pro  minimax-m2.5  minimax-m2.7  minimax-m3
qwen3.5-plus  qwen3.6-plus  qwen3.7-max  qwen3.7-plus
```

Order differs (WebUI groups by family, agent by a different ordering) but that's cosmetic — the picker renders in list order and nothing keys off position.

## The `hy3-preview` call was correct ✔

Your read matches the agent registry. `hy3-preview` is filed under `tencent-tokenhub` in the agent (`models.py:369-370`), never the Go tier — so offering it in the `opencode-go` picker was exactly the #5311-class hazard (a picker entry that sends structurally fine but 404s because it isn't enabled on Go). Removing it closes that. I also grepped the agent for `hy3-free` and it isn't in `models.py` at all yet, so not adding it here was the right call — no point seeding a Go entry the agent's resolver wouldn't recognize.

## Net

The safety invariant from my first pass still holds — the `elif pid == "opencode-go": pass` at `config.py:7017` keeps the live `/v1/models` probe skipped so the curated static list is the sole source, and this refresh only touches that list. With the two catalogs synced, the drift risk is gone. From read-only inspection this is merge-ready. (Data-only, so I can't exercise the picker from here — a quick manual glance that the Go group renders these 19 and nothing 404s on select is the only thing left.)

nesquena-hermes:
Shipped in **v0.51.933** — thanks @ai-ag2026. 🎉 OpenCode Go model picker re-synced to the current flat-rate catalog. Full suite 12,277/0. Deployed.

PR #5762, opened 2026-07-08
Derive service tier support from agent fast-mode metadata

starship-s:
## Thinking Path

WebUI already exposes a main-model service-tier setting, but its eligibility checks had drifted from the agent's fast-mode support metadata. That made provider-level routing too conservative for OpenAI-family models reached through the Codex provider while still needing to fail closed for Codex-specific model slugs and non-OpenAI providers.

## What Changed

- Derive WebUI service-tier support from Hermes Agent's fast-mode override resolver when available.
- Preserve a compatibility fallback for WebUI installs where the agent package is unavailable, so supported OpenAI-family `service_tier: priority` state is not silently removed on Settings save.
- Fail closed for Codex-specific model slugs, empty Codex model selections, non-OpenAI providers, and foreign slash-prefixed IDs.
- Add `supports_fast_tier` metadata to `/api/model/auxiliary`, `/api/models`, `/api/models/live`, and cache/static/minimal catalog paths.
- Preserve capability metadata through provider-prefix rewriting and teach the frontend to distinguish explicit `false` from unknown metadata.
- Add regression coverage for backend request overrides, Settings metadata, live-model cache payloads, and provider-prefix metadata preservation.

## Why It Matters

The Service tier control should reflect the model's actual request capability, not a separate hardcoded WebUI provider list. This keeps the Settings UI, request construction, and model catalog metadata aligned with the agent's model support rules while retaining safe fallback behavior for degraded installs.

## Contract Routing

Task type: additive model settings/API metadata fix.

Touched areas:
- Main-model advanced settings
- Model catalog payloads
- Live model catalog payloads
- Main chat request overrides

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`

Scope boundaries:
- No new provider is enabled broadly.
- Unsupported providers and Codex-specific model slugs remain blocked.
- API payload changes are additive (`supports_fast_tier`) and preserve existing fields.

## Verification

- `./scripts/test.sh tests/test_issue4536_service_tier.py tests/test_auxiliary_models_settings.py tests/test_live_models_ttl_cache.py tests/test_issue3718_live_models_custom_probe.py tests/test_issue1807_codex_provider_card_live_models.py tests/test_credential_pool_providers.py tests/test_issue4756_session_visit_model_refresh.py tests/test_model_cache_metadata.py -q` → `128 passed`
- `node --check static/ui.js`
- `python -m py_compile api/config.py api/routes.py`
- `git diff --check`

## Risks / Follow-ups

- This intentionally keeps a small compatibility fallback for environments where agent metadata cannot be imported. The agent resolver remains the normal source of truth.
- No screenshots are included because the UI control already exists; this PR changes its provider/model eligibility metadata and covers that behavior with static frontend assertions.

## Models Used

- `gpt-5.5` orchestrated the investigation, verification, and PR preparation.
- `gpt-5.3-codex-spark` implemented the initial branch patch.
- `claude-fable-5` reviewed the final diff and approved.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -2212,10 +2212,12 @@ def _apply_provider_prefix(
     result = []
     for m in raw_models:
         mid = m["id"]
+        entry = dict(m)
         if mid.startswith("@") or "/" in mid:
-            result.append({"id": mid, "label": m["label"]})
+            result.append(entry)
         else:
-            result.append({"id": f"@{provider_id}:{mid}", "label": m["label"]})
+            entry["id"] = f"@{provider_id}:{mid}"
+            result.append(entry)
     return result
 
 
@@ -3839,6 +3841,74 @@ def _is_openai_family_provider(provider: str | None) -> bool:
     return resolved in ("openai", "openai-api", "openai-codex")
 
 
+def _normalize_openai_family_model_id(model_id: str | None) -> str:
+    """Return a model id in the form expected by hermes_cli fast-mode resolution."""
+    model = str(model_id or "").strip()
+    if not model:
+        return ""
+
+    if model.startswith("@") and ":" in model:
+        model = model.split(":", 1)[1].strip()
+
+    if "://" in model:
+        return model
+
+    if "/" in model:
+        provider_hint, candidate = model.split("/", 1)
+        if provider_hint.strip().lower() in {"openai", "openai-api", "openai-codex"}:
+            model = candidate.strip()
+        else:
+            return ""
+
+    return model
+
+
+def _legacy_openai_service_tier_overrides(model_id: str | None, provider: str | None) -> dict:
+    """Compatibility fallback for standalone WebUI installs without hermes_cli.
+
+    Normal operation delegates to Hermes Agent model metadata.  This fallback
+    preserves the old WebUI behavior when the agent package is unavailable,
+    while still failing closed for codex model slugs and foreign provider IDs.
+    """
+    if not _is_openai_family_provider(provider):
+        return {}
+    resolved_provider = str(_resolve_provider_alias(str(provider or "").strip().lower()))
+    raw_model = str(model_id or "").strip()
+    if "://" not in raw_model and "/" in raw_model:
+        provider_hint = raw_model.split("/", 1)[0].strip().lower()
+        if provider_hint not in {"openai", "openai-api", "openai-codex"}:
+            return {}
+    normalized_model = _normalize_openai_family_model_id(model_id)
+    if not normalized_model:
+        if resolved_provider == "openai-codex":
+            return {}
+        return {"service_tier": "priority"}
+    lowered = normalized_model.lower()
+    if "codex" in lowered:
+        return {}
+    if lowered.startswith(("gpt-", "o1", "o3", "o4")):
+        return {"service_tier": "priority"}
+    return {}
+
+
+def _resolve_main_model_fast_mode_overrides(model_id: str | None, provider: str | None = None) -> dict:
+    """Return provider request overrides for the main model fast-mode setting."""
+    normalized_model = _normalize_openai_family_model_id(model_id)
+    if not normalized_model:
+        return _legacy_openai_service_tier_overrides(model_id, provider)
+    try:
+        from hermes_cli.models import resolve_fast_mode_overrides
+    except Exception:
+        logger.debug("Failed to import hermes_cli.models.resolve_fast_mode_overrides; using WebUI compatibility fallback.")
+        return _legacy_openai_service_tier_overrides(model_id, provider)
+    try:
+        resolved = resolve_fast_mode_overrides(normalized_model)
+    except Exception:
+        logger.debug("Failed to resolve fast-mode overrides for %r; using WebUI compatibility fallback.", normalized_model)
+        return _legacy_openai_service_tier_overrides(model_id, provider)
+    return resolved if isinstance(resolved, dict) else {}
+
+
 def _main_model_supports_service_tier(
     model_id: str | None,
     provider: str | None,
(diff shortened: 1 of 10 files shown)


nesquena-hermes:
Ran a read-only cross-repo check on the agent side of this contract, since the whole PR hinges on delegating to `hermes_cli.models.resolve_fast_mode_overrides` (`api/config.py:3900`). The contract lines up — reporting what I verified so the delegation isn't taken on faith.

## Agent symbol exists and its shape matches

`resolve_fast_mode_overrides` is defined in the agent at `hermes_cli/models.py:2118`:

```python
def resolve_fast_mode_overrides(model_id: Optional[str]) -> dict[str, Any] | None:
    if not model_supports_fast_mode(model_id):
        return None
    if _is_anthropic_fast_model(model_id):
        return {"speed": "fast"}
    return {"service_tier": "priority"}
```

Three things the WebUI side handles correctly against this:

1. **`None` return** — `_resolve_main_model_fast_mode_overrides` (`config.py:3907`) wraps it as `resolved if isinstance(resolved, dict) else {}`, so the unsupported `None` becomes `{}`. Good.
2. **The Anthropic `{"speed": "fast"}` branch can't leak in** — `_main_model_supports_service_tier` (`config.py:3916`) gates on `_is_openai_family_provider` before calling the resolver and only accepts `service_tier == "priority"`, so an Opus-4.6 model routed through a non-OpenAI provider never trips the toggle. Correct layering.
3. **Codex fail-closed matches on both sides** — the agent's `_is_openai_fast_model` (`models.py:2064`) excludes any slug containing `codex` before its prefix check, and the WebUI legacy fallback (`config.py:3888`) mirrors that with `if "codex" in lowered: return {}`.

## The prefix set is behavior-preserving

The agent's `_OPENAI_FAST_MODE_PREFIXES` (`models.py:2056`) is `("gpt-", "o1", "o3", "o4")` — identical to the old WebUI heuristic this PR replaces (the removed `bare_model.startswith(("gpt-", "o1", "o3", "o4"))` line). So for the base OpenAI family, delegating to the agent is a no-op behavior change; the win is that the agent becomes the single source of truth going forward instead of a second hardcoded copy that can drift.

## One nuance worth a maintainer sanity-check

The intended behavior change (per the PR body) is that a plain OpenAI-family slug reached through the Codex provider stays eligible, while only Codex-specific slugs fail closed. That's faithful to the agent: `_is_openai_fast_model` gates the codex exclusion on `"codex" in base` (the slug), not on the provider being `openai-codex`. So `openai-codex` + `gpt-5.5` gives `{"service_tier": "priority"}`, while `openai-codex` + `gpt-5.5-codex` gives `{}`. Both sides agree. The one thing I couldn't verify from here is that the agent's Codex Responses API request path actually accepts `service_tier` for a non-codex slug routed through Codex at runtime — worth a one-line confirmation, since the resolver returning `priority` while the transport drops it would be a silent mismatch. That's an agent-runtime question (`run_agent.py` `_build_api_kwargs`, per the resolver docstring), not a WebUI-layer one.

## Standalone-install fallback is tested

The `monkeypatch.setitem(sys.modules, "hermes_cli.models", None)` cases (`test_standalone_agent_import_failure_preserves_supported_service_tier` and `..._still_blocks_codex_slug`) exercise both legs of the degraded path, so the compatibility fallback isn't just asserted in the description.

From a read-only inspection the cross-repo contract is sound. Leaving the full-suite gate and the runtime Codex `service_tier` confirmation to the maintainer.

starship-s:
Confirmed the runtime side of the Codex `service_tier` path:

- The WebUI branch emits `request_overrides={"service_tier": "priority"}` for `openai-codex` + `gpt-5.5`.
- Agent forwards `agent.request_overrides` into the Codex Responses transport (`chat_completion_helpers.py` passes it through to `CodexTransport.build_kwargs`).
- `CodexTransport.build_kwargs` merges `request_overrides` directly into the Responses kwargs before backend-specific cleanup; the only `service_tier` strip is scoped to xAI Responses, not the ChatGPT Codex backend. For `is_codex_backend`, the code only adds cache-scope HTTP headers and leaves `service_tier` in the body kwargs.

So the resolver returning `{ "service_tier": "priority" }` is not dropped on the Agent Codex Responses path for this route.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (full gate: Codex + Fable-UX + suite) · optional-dep-safe · ⏸️ visible service-tier → Nathan

**Certified head:** `sha:a6d73f7f` (clean rebase, branch `gate-rebase/5762-service-tier-fastmode-derive`) · **PR:** #5762 · _starship-s, Derive service tier support from agent fast-mode metadata_
**Verdict:** GREEN. Replaces a hardcoded service-tier list with derivation from the agent's fast-mode metadata, behind a properly-guarded optional `hermes_cli` import with a complete legacy fallback. Codex SAFE, Fable-UX SHIP-UX, full suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5762`) — full gate: Codex + Fable + suite
| Gate | Result |
|------|--------|
| Codex (optional-dep safety / tier-mapping correctness / backward-compat / XSS) | **SAFE TO SHIP** — 0 findings |
| **Fable-UX** | **SHIP-UX** — service-tier UI renders correctly |
| Full pytest suite | ✅ **12326 passed, 0 failed** (+ 10 service-tier tests) |

### Findings
**✅ Optional-dep safety verified (my recurring regression class):** the `from hermes_cli.models import resolve_fast_mode_overrides` at `config.py:3900` is LAZY (inside the function) + guarded by try/except → `logger.debug("...using WebUI compatibility fallback")` → `_legacy_openai_service_tier_overrides(...)`, matching the codebase's established lazy-import pattern (config.py has ~20 such guarded hermes_cli imports). I confirmed `_legacy_openai_service_tier_overrides` returns the SAME dict shape (`{}` or `{"service_tier":"priority"}`) that `resolve_fast_mode_overrides` would — no partial-dict mismatch. So an agent-less install / CI runner without hermes_cli does NOT `ModuleNotFoundError` (falls back cleanly), and a runtime raise (not just import) also falls back. Codex confirmed all of this.
**✅ Correctness (Codex):** the fast-mode-metadata → service_tier mapping is right (priority/flex/default), backward-compat preserved (OpenAI priority stays priority via the legacy fallback), the `service_tier=='priority'` check normalizes (strip/lower), no wrong-tier for any provider, no XSS from derived metadata in the panels.js/ui.js surface, no regression to the model-cache/TTL/metadata tests. Fable confirmed the visible service-tier setting renders correctly for supported vs unsupported models. Full suite green.

### Recommendation to Nathan
**GREEN — merge from branch `gate-rebase/5762-service-tier-fastmode-derive` (`sha:a6d73f7f`) — optional service-tier UI glance (a tier-supporting model shows the setting; an unsupported one doesn't).** A clean refactor from hardcoded service-tier list → agent-fast-mode-metadata derivation, with the optional-dep import correctly guarded + a complete compatibility fallback (agent-less/CI-safe). Codex SAFE + Fable SHIP-UX + full suite green. concept 4/5. Author @starship-s. crit=3, backend correctness + optional-dep + visible.

---
_Gate-certifier layer (warm-up → **gate** → release). service_tier derived via hermes_cli.models.resolve_fast_mode_overrides — import LAZY+guarded (try/except→logger.debug→_legacy_openai_service_tier_overrides fallback, matching the ~20 existing guarded hermes_cli imports); fallback returns SAME dict shape ({} | {service_tier:priority}) so agent-less/CI-safe (no ModuleNotFoundError), runtime-raise also falls back, backward-compat OpenAI-priority preserved, normalize strip/lower, no XSS. Codex SAFE + Fable-UX SHIP-UX + full suite green (0 failed, 12326) + 10 service-tier tests. Visible service-tier → Nathan. Cert valid for sha:a6d73f7f._

nesquena-hermes:
Shipped in **v0.51.939** — thanks @starship-s. 🎉 Service-tier support now derived from agent model metadata (control shows only for supporting models, safe-hidden when absent). Codex SAFE + suite 12,293/0. Deployed.

PR #5764, opened 2026-07-08
[security] fix(terminal): resolve forwarded client before local gate

Hinotoi-agent:
## Summary
This PR hardens the unauthenticated local/private gate used by the embedded-terminal endpoints when trusted forwarded headers are enabled.

- Resolves the original client from `X-Forwarded-For` instead of classifying the rightmost/private proxy hop as the client.
- Adds an explicit `HERMES_WEBUI_TRUSTED_PROXY_CIDRS` allowlist for trusted reverse proxies.
- Keeps passwordless embedded-terminal access blocked for public clients forwarded through a private proxy.
- Adds regression coverage for public-client, private-client, and unlisted-proxy XFF chains.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Trusted `X-Forwarded-For` chain used the private proxy hop for locality | A passwordless WebUI behind a trusted reverse proxy could treat a public remote client as local/private and admit embedded-terminal endpoints that drive a PTY shell | Critical for affected deployments |

## Before this PR
- `HERMES_WEBUI_TRUST_FORWARDED_FOR=1` made `_onboarding_request_is_local()` inspect the rightmost `X-Forwarded-For` value first.
- In common proxy chains such as `X-Forwarded-For: <public-client>, <private-proxy>`, the rightmost hop is the private proxy, not the original client.
- The embedded-terminal gate reuses this locality check, so a public client forwarded by a private proxy could pass the passwordless terminal gate.

## After this PR
- Forwarded headers are honored only when the immediate socket peer is loopback or listed in `HERMES_WEBUI_TRUSTED_PROXY_CIDRS`.
- The trusted chain is walked from right to left and the first non-trusted hop is classified as the client.
- A public original client remains blocked even when the proxy hop is private.
- Tests cover public-client, private-client, and spoofed/unlisted-proxy chains.

## Explicit operator choice
Secure default behavior:

```bash
# unset by default
HERMES_WEBUI_TRUST_FORWARDED_FOR=
HERMES_WEBUI_TRUSTED_PROXY_CIDRS=
```

Forwarded headers are ignored unless the operator explicitly opts in.

Trusted reverse-proxy behavior:

```bash
HERMES_WEBUI_TRUST_FORWARDED_FOR=1
HERMES_WEBUI_TRUSTED_PROXY_CIDRS=10.0.0.0/8,192.168.0.0/16
```

In this mode, the WebUI trusts forwarded client IPs only from loopback or listed proxy networks and classifies the original non-proxy hop.

## Why this matters
The embedded terminal spawns and drives a shell as the WebUI server user. When authentication is disabled, the terminal routes rely on the local/private gate as the remaining admission boundary. A trusted-proxy deployment should not classify a public user as local only because the final proxy hop is private.

Existing terminal hardening added a local-origin gate for embedded-terminal endpoints when authentication is disabled.

This PR fixes a remaining bypass of that gate in trusted forwarded-header mode: the gate existed, but the client-IP extraction could use the proxy hop instead of the original client.

## Attack flow

```text
public remote client
    -> trusted reverse proxy adds X-Forwarded-For: public-client, private-proxy
        -> WebUI trusted-forwarded mode reads the private proxy hop as local
            -> embedded-terminal gate admits the request
                -> unauthenticated caller can reach PTY-backed terminal endpoints
```

## Affected code

| Issue | Files |
|-------|-------|
| Trusted XFF terminal locality bypass | `api/routes.py`, `tests/test_cvd3_terminal_local_origin_gate.py`, `tests/test_security_review_fixes.py` |

## Root cause
Trusted XFF terminal locality bypass:
- The previous trusted-forwarded-header branch inspected the rightmost `X-Forwarded-For` hop first.
- In normal proxy chains, that hop is commonly the private proxy, so the local/private decision could be made on the proxy rather than the original client.

## CVSS assessment

| Trusted XFF terminal locality bypass | 9.8 Critical for affected deployments | `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H` |

Rationale:
- The score assumes a passwordless WebUI exposed through a reverse proxy with `HERMES_WEBUI_TRUST_FORWARDED_FOR=1`. Under those conditions, a remote unauthenticated client can reach a PTY-backed terminal route that runs commands as the WebUI process user.

## Safe reproduction steps
1. Configure a passwordless WebUI behind a trusted reverse proxy.
2. Enable forwarded-header trust with `HERMES_WEBUI_TRUST_FORWARDED_FOR=1`.
3. Send a terminal request with a common proxy chain shape:
   ```http
   X-Forwarded-For: 8.8.8.8, 10.0.0.10
   ```
   where `10.0.0.10` is the private proxy hop.
4. Before this PR, the local/private decision could be made on `10.0.0.10`.
5. After this PR, the chain resolves to `8.8.8.8` and the terminal gate denies the request.

- A public remote client should not be considered local because the final reverse-proxy hop is private.
- The pre-patch behavior could classify the rightmost proxy hop as local/private.
- The regression test asserts the resolved client is the public hop and terminal access is denied.

## Changes in this PR
- Adds trusted-proxy CIDR parsing through `HERMES_WEBUI_TRUSTED_PROXY_CIDRS`.
- Adds shared IP helpers for local/private and trusted-network checks.
- Resolves forwarded clients by walking `X-Forwarded-For` from right to left and stopping at the first non-trusted hop.
- Keeps forwarded headers denied when the immediate proxy is not loopback or explicitly trusted.
- Adds focused terminal-gate regression tests for proxy-chain behavior.

## Files changed

| Backend security gate | `api/routes.py` | Trusted forwarded-header client resolution now requires a trusted proxy and resolves the original non-proxy client |
| Regression tests | `tests/test_cvd3_terminal_local_origin_gate.py`, `tests/test_security_review_fixes.py` | Added public/private/unlisted proxy chain coverage and updated forwarded-header locality expectations |

## Maintainer impact
- The patch is limited to the existing onboarding/embedded-terminal locality gate.
- Authentication-enabled deployments are unchanged because `check_auth()` still admits only authenticated sessions before terminal handlers run.
- Passwordless deployments that intentionally rely on a LAN/reverse proxy can keep forwarded-header support by listing the trusted proxy CIDRs.

## Fix rationale
- The locality decision belongs at the boundary that admits passwordless onboarding and embedded-terminal requests.
- Walking the trusted proxy chain avoids trusting a spoofable left-hand client value while also avoiding the previous rightmost-proxy classification bug.
- The regression tests pin both blocked public-client and allowed private-client behavior.

## Type of change
- [x] Security fix
- [x] Tests
- [ ] Documentation update

## Test plan
- [x] Focused terminal locality tests
- [x] Diff-scoped Python lint
- [x] Syntax compilation for touched Python files

Executed with:
- `./scripts/test.sh tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py`
- `python3 -m compileall -q api/routes.py tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py`
- `git diff --check`

## Disclosure notes
- Claims are bounded to passwordless deployments that explicitly enable trusted forwarded headers and expose the WebUI through a reverse proxy.
- No production systems or real credentials were used.

## AI Usage Disclosure
AI assistance was used to identify the boundary, draft the patch, and write regression tests. The submitted change is scoped to the files and validation listed above.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5547,11 +5547,19 @@ def _forwarded_client_ip_from_trusted_proxy(handler) -> str:
         if hops:
             return hops[0]
 
-    # Do not fall back to X-Real-IP here. Some proxies pass client-supplied
-    # X-Real-IP through when X-Forwarded-For is absent, which would let a public
-    # caller choose a private client IP. Without XFF, only a same-host loopback
-    # proxy can be treated as local; non-loopback trusted proxies must include
-    # an XFF chain so the original client can be classified.
+    # Do not fall back to X-Real-IP or Forwarded here. Some proxies pass
+    # client-supplied values through when X-Forwarded-For is absent, which would
+    # let a public caller choose a private client IP. If a forwarded-client
+    # header is present but XFF is absent, fail closed instead of ignoring it and
+    # treating the loopback/private proxy hop as the client.
+    x_real_ip = str(handler.headers.get("X-Real-IP", "") or "").strip()
+    forwarded = str(handler.headers.get("Forwarded", "") or "").strip()
+    if x_real_ip or forwarded:
+        return ""
+
+    # Without any forwarded-client header, only a same-host loopback proxy can be
+    # treated as local; non-loopback trusted proxies must include an XFF chain so
+    # the original client can be classified.
     return raw_peer if peer_is_loopback else ""
 
 
--- a/api/routes.py
+++ b/api/routes.py
@@ -5488,24 +5488,30 @@ def _trusted_proxy_networks():
     return tuple(networks)
 
 
-def _ip_is_loopback_or_private(raw: str) -> bool:
+def _ip_address(raw: str):
     import ipaddress
 
     try:
         addr = ipaddress.ip_address(str(raw or "").strip())
     except ValueError:
-        return False
-    return bool(addr.is_loopback or addr.is_private)
+        return None
+    mapped = getattr(addr, "ipv4_mapped", None)
+    return mapped or addr
+
+
+def _ip_is_loopback(raw: str) -> bool:
+    addr = _ip_address(raw)
+    return bool(addr and addr.is_loopback)
+
+
+def _ip_is_loopback_or_private(raw: str) -> bool:
+    addr = _ip_address(raw)
+    return bool(addr and (addr.is_loopback or addr.is_private))
 
 
 def _ip_in_networks(raw: str, networks) -> bool:
-    import ipaddress
-
-    try:
-        addr = ipaddress.ip_address(str(raw or "").strip())
-    except ValueError:
-        return False
-    return any(addr in network for network in networks)
+    addr = _ip_address(raw)
+    return bool(addr and any(addr in network for network in networks))
 
 
 def _forwarded_client_ip_from_trusted_proxy(handler) -> str:
@@ -5522,7 +5528,7 @@ def _forwarded_client_ip_from_trusted_proxy(handler) -> str:
         return ""
 
     trusted_networks = _trusted_proxy_networks()
-    peer_is_loopback = _ip_is_loopback_or_private(raw_peer) and str(raw_peer).startswith(("127.", "::1"))
+    peer_is_loopback = _ip_is_loopback(raw_peer)
     peer_is_trusted = _ip_in_networks(raw_peer, trusted_networks) if trusted_networks else peer_is_loopback
     if not peer_is_trusted:
         return ""
@@ -5537,11 +5543,12 @@ def _forwarded_client_ip_from_trusted_proxy(handler) -> str:
         if hops:
             return hops[0]
 
-    x_real_ip = str(handler.headers.get("X-Real-IP", "") or "").strip()
-    if x_real_ip:
-        return x_real_ip
-
-    return raw_peer
+    # Do not fall back to X-Real-IP here. Some proxies pass client-supplied
+    # X-Real-IP through when X-Forwarded-For is absent, which would let a public
+    # caller choose a private client IP. Without XFF, only a same-host loopback
+    # proxy can be treated as local; non-loopback trusted proxies must include
+    # an XFF chain so the original client can be classified.
+    return raw_peer if peer_is_loopback else ""
 
 
 def _onboarding_request_is_local(handler) -> bool:
@@ -5575,7 +5582,7 @@ def _onboarding_request_is_local(handler) -> bool:
     if not raw:
         return False
     if forwarded_present:
-        return _ip_is_loopback_or_private(raw) and str(raw).startswith(("127.", "::1"))
+        return _ip_is_loopback(raw)
     return _ip_is_loopback_or_private(raw)
 
 
--- a/api/routes.py
+++ b/api/routes.py
@@ -5471,39 +5471,94 @@ def _request_client_ip(handler) -> str:
     return ""
 
 
+def _trusted_proxy_networks():
+    """Return explicitly trusted reverse-proxy CIDRs for forwarded client IPs."""
+    import ipaddress
+
+    raw = os.getenv("HERMES_WEBUI_TRUSTED_PROXY_CIDRS", "")
+    networks = []
+    for item in raw.replace(";", ",").split(","):
+        item = item.strip()
+        if not item:
+            continue
+        try:
+            networks.append(ipaddress.ip_network(item, strict=False))
+        except ValueError:
+            logger.warning("Ignoring invalid HERMES_WEBUI_TRUSTED_PROXY_CIDRS entry: %s", item)
+    return tuple(networks)
+
+
+def _ip_is_loopback_or_private(raw: str) -> bool:
+    import ipaddress
+
+    try:
+        addr = ipaddress.ip_address(str(raw or "").strip())
+    except ValueError:
+        return False
+    return bool(addr.is_loopback or addr.is_private)
+
+
+def _ip_in_networks(raw: str, networks) -> bool:
+    import ipaddress
+
+    try:
+        addr = ipaddress.ip_address(str(raw or "").strip())
+    except ValueError:
+        return False
+    return any(addr in network for network in networks)
+
+
+def _forwarded_client_ip_from_trusted_proxy(handler) -> str:
+    """Resolve the original client IP from trusted forwarded headers.
+
+    ``X-Forwarded-For`` chains are ordered as ``client, proxy1, proxy2``.  The
+    previous implementation used the rightmost value, which is commonly a
+    private/loopback proxy hop and can make a public client look local.  Only
+    trust forwarded headers when the immediate socket peer is trusted, then walk
+    from right to left and return the first non-proxy hop.
+    """
+    raw_peer = _request_client_ip(handler)
+    if not raw_peer:
+        return ""
+
+    trusted_networks = _trusted_proxy_networks()
+    peer_is_loopback = _ip_is_loopback_or_private(raw_peer) and str(raw_peer).startswith(("127.", "::1"))
+    peer_is_trusted = _ip_in_networks(raw_peer, trusted_networks) if trusted_networks else peer_is_loopback
+    if not peer_is_trusted:
+        return ""
+
+    xff = str(handler.headers.get("X-Forwarded-For", "") or "").strip()
+    if xff:
+        hops = [part.strip() for part in xff.split(",") if part.strip()]
+        for hop in reversed(hops):
+            if trusted_networks and _ip_in_networks(hop, trusted_networks):
+                continue
+            return hop
+        if hops:
+            return hops[0]
+
+    x_real_ip = str(handler.headers.get("X-Real-IP", "") or "").strip()
+    if x_real_ip:
+        return x_real_ip
+
+    return raw_peer
+
+
 def _onboarding_request_is_local(handler) -> bool:
     """Return True when an unauthenticated onboarding request is local/private.
 
     Forwarded client-IP headers are ignored by default because direct clients can
     spoof them. Operators behind a trusted reverse proxy may opt in with
-    HERMES_WEBUI_TRUST_FORWARDED_FOR=1, matching the explicit forwarded-header
-    trust model used elsewhere in the server.
-
-    When forwarded headers are PRESENT but not trusted, the request arrived
-    through a proxy, so the raw socket address is the proxy's (typically
-    loopback/private) and tells us nothing about the real client's locality.
-    In that case we deny rather than fall back to the proxy socket — otherwise a
-    public client behind any reverse proxy would be treated as local. Operators
-    who front the WebUI with a trusted proxy must set
-    HERMES_WEBUI_TRUST_FORWARDED_FOR=1 (or HERMES_WEBUI_ONBOARDING_OPEN=1).
+    HERMES_WEBUI_TRUST_FORWARDED_FOR=1.  In that mode, forwarded headers are
+    honored only from a loopback proxy or a proxy listed in
+    HERMES_WEBUI_TRUSTED_PROXY_CIDRS, and the resolved client is the original
+    non-proxy hop rather than the rightmost/private proxy hop.
     """
-    import ipaddress
-
     trust_forwarded = _truthy_env("HERMES_WEBUI_TRUST_FORWARDED_FOR")
     if trust_forwarded:
-        candidates = [
-            handler.headers.get("X-Forwarded-For", "").split(",")[-1].strip(),
-            handler.headers.get("X-Real-IP", "").strip(),
-            _request_client_ip(handler),
-        ]
-        for raw in candidates:
-            if not raw:
-                continue
-            try:
-                addr = ipaddress.ip_address(raw)
-            except ValueError:
-                continue
-            return bool(addr.is_loopback or addr.is_private)
+        forwarded_client = _forwarded_client_ip_from_trusted_proxy(handler)
+        if forwarded_client:
+            return _ip_is_loopback_or_private(forwarded_client)
         return False
 
     # Untrusted forwarded headers present → the request arrived through a proxy.
@@ -5511,9 +5566,7 @@ def _onboarding_request_is_local(handler) -> bool:
     # counts as local in that case: a loopback raw socket is a genuine same-host
     # client (or a same-host proxy the operator controls), whereas a PRIVATE/LAN
     # raw socket is a separate proxy box that could be forwarding an arbitrary
-    # (public) client we can't see without trusting the header. Operators who
-    # front the WebUI with a LAN proxy must set HERMES_WEBUI_TRUST_FORWARDED_FOR=1
-    # (or HERMES_WEBUI_ONBOARDING_OPEN=1).
+    # (public) client we can't see without trusting the header.
     forwarded_present = bool(
         handler.headers.get("X-Forwarded-For", "").strip()
         or handler.headers.get("X-Real-IP", "").strip()
@@ -5521,13 +5574,9 @@ def _onboarding_request_is_local(handler) -> bool:
     raw = _request_client_ip(handler)
     if not raw:
         return False
-    try:
-        addr = ipaddress.ip_address(raw)
-    except ValueError:
-        return False
     if forwarded_present:
-        return bool(addr.is_loopback)
-    return bool(addr.is_loopback or addr.is_private)
+        return _ip_is_loopback_or_private(raw) and str(raw).startswith(("127.", "::1"))
+    return _ip_is_loopback_or_private(raw)
 
 
 def _onboarding_gate_allows(handler, auth_enabled: bool | None = None) -> bool:
--- a/api/routes.py
+++ b/api/routes.py
@@ -5514,6 +5514,10 @@ def _ip_in_networks(raw: str, networks) -> bool:
     return bool(addr and any(addr in network for network in networks))
 
 
+def _ip_is_trusted_proxy_hop(raw: str, trusted_networks) -> bool:
+    return _ip_is_loopback(raw) or _ip_in_networks(raw, trusted_networks)
+
+
 def _forwarded_client_ip_from_trusted_proxy(handler) -> str:
     """Resolve the original client IP from trusted forwarded headers.
 
@@ -5529,7 +5533,7 @@ def _forwarded_client_ip_from_trusted_proxy(handler) -> str:
 
     trusted_networks = _trusted_proxy_networks()
     peer_is_loopback = _ip_is_loopback(raw_peer)
-    peer_is_trusted = _ip_in_networks(raw_peer, trusted_networks) if trusted_networks else peer_is_loopback
+    peer_is_trusted = peer_is_loopback or _ip_in_networks(raw_peer, trusted_networks)
     if not peer_is_trusted:
         return ""
 
@@ -5537,7 +5541,7 @@ def _forwarded_client_ip_from_trusted_proxy(handler) -> str:
     if xff:
         hops = [part.strip() for part in xff.split(",") if part.strip()]
         for hop in reversed(hops):
-            if trusted_networks and _ip_in_networks(hop, trusted_networks):
+            if _ip_is_trusted_proxy_hop(hop, trusted_networks):
                 continue
             return hop
         if hops:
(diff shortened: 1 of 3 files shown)


Hinotoi-agent:
Thanks for the review — fixed both points in b80cd31b.\n\nChanges made:\n- Removed the trusted-proxy fallback to `X-Real-IP` when `X-Forwarded-For` is absent, so a client-controlled `X-Real-IP` value cannot choose a private/loopback client for the passwordless terminal gate. Non-loopback trusted proxies now need an XFF chain to classify the original client.\n- Normalized IPv4-mapped IPv6 addresses before loopback/private checks, so peers such as `::ffff:127.0.0.1` are handled as loopback.\n- Added regressions for both cases.\n\nValidation run locally:\n- `./scripts/test.sh tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py` → 34 passed\n- `python3 scripts/ruff_lint.py --diff origin/master` → no new violations\n- `python3 -m compileall -q api/routes.py tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py`\n- `git diff --check`

Hinotoi-agent:
Thanks for the review — fixed both points in `b80cd31b`.

Changes made:
- Removed the trusted-proxy fallback to `X-Real-IP` when `X-Forwarded-For` is absent, so a client-controlled `X-Real-IP` value cannot choose a private/loopback client for the passwordless terminal gate. Non-loopback trusted proxies now need an XFF chain to classify the original client.
- Normalized IPv4-mapped IPv6 addresses before loopback/private checks, so peers such as `::ffff:127.0.0.1` are handled as loopback.
- Added regressions for both cases.

Validation run locally:
- `./scripts/test.sh tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py` → 34 passed
- `python3 -m compileall -q api/routes.py tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py`
- `git diff --check`

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (the auth-bypass fix is INCOMPLETE — a public client + trailing loopback XFF hop still passes the local gate — AND it breaks legit loopback access when CIDRs are set — 2 CORE, both reproduced) · SECURITY → Nathan

**Certified head:** `sha:<rebased>` (rebased onto current master, `git apply` clean) · **PR:** #5764 [security] · _Hinotoi-agent (agent-authored), resolve forwarded client before local gate_
**Verdict:** The direction is right (only trust XFF from a trusted peer; walk to the real client), but Codex empirically reproduced that the residual auth-bypass is still open AND the fix introduces a functional regression. The root of both: "trusted proxy hop" is defined as `in-CIDRs` only, not `loopback OR in-CIDRs`. Suite green (tests don't cover these).

### What I ran (rebased worktree `/tmp/wt-rebase-5764`) — SECURITY gate: deep Codex + full suite
| Gate | Result |
|------|--------|
| Codex (spoof / walk-correctness / regression) | **SHIP ONLY WITH FIXES** — 2 CORE (both reproduced) |
| Full pytest suite | ✅ **12329 passed, 0 failed** (+ 34 gate/security tests — don't cover these edges) |

### Findings
**⛔ CORE (routes.py:5539) — residual auth-bypass (the security hole is NOT fully closed):** the right-to-left walk only skips `HERMES_WEBUI_TRUSTED_PROXY_CIDRS` hops — it does NOT skip the implicit trusted LOOPBACK proxy hop. Codex verified: `raw peer=127.0.0.1`, `X-Forwarded-For: 8.8.8.8, 127.0.0.1` → resolves to `127.0.0.1` → `_onboarding_request_is_local()` returns **True** → a PUBLIC client (8.8.8.8) still passes the unauthenticated local terminal/onboarding gate by appending a `127.0.0.1` hop. The fix reduced but did not eliminate the bypass. **Fix:** "trusted proxy hop" must mean `loopback OR in-trusted-CIDRs`, and the walk must skip loopback hops too (so it returns 8.8.8.8, which fails the local check).
**⛔ CORE (routes.py:5532) — functional regression, legit loopback denied when CIDRs set:** `peer_is_trusted = _ip_in_networks(raw_peer, trusted_networks) if trusted_networks else peer_is_loopback` — so once `HERMES_WEBUI_TRUSTED_PROXY_CIDRS=10.0.0.0/8` is set, a legitimate `127.0.0.1` peer is NO LONGER trusted → resolves to "" → gate returns False → same-host access broken (self-DoS for anyone who configures CIDRs). **Fix:** `peer_is_trusted = peer_is_loopback or _ip_in_networks(raw_peer, trusted_networks)`.
**✅ What's correct:** an untrusted DIRECT client forging `X-Forwarded-For: 127.0.0.1` is correctly rejected (peer not trusted → return "" → raw public peer → gate rejects) — the primary spoof vector is closed. The `_trusted_proxy_networks` CIDR parsing + IPv6/malformed handling are sound.

### Recommendation to the next agent / author + Nathan
**RED — gate-fail/changes-requested (2 CORE): unify "trusted proxy hop" = `loopback OR in-trusted-CIDRs` in BOTH places — (1) the `peer_is_trusted` check (`= peer_is_loopback or _ip_in_networks(...)`, fixes the loopback-denied-when-CIDRs-set regression) and (2) the XFF walk-skip (skip loopback hops too, so `8.8.8.8, 127.0.0.1` resolves to `8.8.8.8` and the public client is correctly rejected — closes the residual bypass). Add the two Codex repro cases as tests.** The primary direct-spoof vector is closed, but a security fix on an unauth RCE-adjacent gate must fully close the bypass — the trailing-loopback-hop case is a real remaining exploit. **SECURITY + agent-authored → routes to you, Nathan**; most-thorough-fix rule applies. concept 4/5 (right approach, incomplete + a regression). Author @Hinotoi-agent.

---
_Gate-certifier layer (warm-up → **gate** → release). Forwarded-client resolution for the unauth local terminal gate: direct-spoof (untrusted peer forging XFF) correctly closed, BUT Codex reproduced 2 CORE — (1) residual bypass: walk skips only CIDR hops not loopback, so a public client + trailing 127.0.0.1 XFF hop (8.8.8.8,127.0.0.1 from a loopback peer) resolves to 127.0.0.1 → local gate TRUE = public client passes; (2) regression: peer_is_trusted ignores loopback when CIDRs set → legit 127.0.0.1 denied. Root: 'trusted proxy hop' must be loopback OR in-CIDRs in BOTH the peer-trust check + the walk-skip. Suite green 12329 (edges untested). SECURITY + agent-authored → Nathan. Cert valid for the rebased head._

Hinotoi-agent:
Thanks — fixed the two CORE cases in `96db8528`.

Changes made:
- Treat a trusted proxy hop consistently as `loopback OR in HERMES_WEBUI_TRUSTED_PROXY_CIDRS`.
- Keep trusting same-host loopback proxy peers even when `HERMES_WEBUI_TRUSTED_PROXY_CIDRS` is configured, so legitimate loopback access is not denied.
- Skip loopback hops during the right-to-left XFF walk, so `X-Forwarded-For: 8.8.8.8, 127.0.0.1` from a loopback peer resolves to `8.8.8.8` and remains blocked.
- Added regression tests for both the trailing-loopback-hop bypass and the loopback-with-CIDRs functional regression across the terminal and onboarding gates.

Validation run locally:
- `./scripts/test.sh tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py` → 38 passed
- `python3 -m compileall -q api/routes.py tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py`
- `git diff --check`

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (round-1 CORE both fixed, but a NEW CORE: `X-Real-IP`-only reverse proxies now leak public clients as local — a REGRESSION vs master) · SECURITY → Nathan

**Certified head:** `sha:96db8528` (rebased onto current master, `git apply` clean) · **PR:** #5764 [security] · _Hinotoi-agent (agent-authored), resolve forwarded client before local gate_
**Verdict:** My round-1 CORE findings are both fixed (the unified `_ip_is_trusted_proxy_hop`=loopback-OR-CIDR predicate closes the trailing-loopback bypass + the loopback-denied regression). But Codex found a THIRD variant of the same auth-bypass on a different header: when XFF is absent but `X-Real-IP` is present, the new code returns the loopback `raw_peer` (→ local) and IGNORES the `X-Real-IP` real client — and `origin/master` correctly rejected that case, so this is a **security regression vs master** for `X-Real-IP`-only proxy deployments. Suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5764b`) — SECURITY re-gate: deep Codex + full suite
| Gate | Result |
|------|--------|
| Codex (2 r1 chains closed? + new edges) | **SHIP ONLY WITH FIXES** — 1 new CORE (reproduced) |
| Full pytest suite | ✅ **12333 passed, 0 failed** (+ 38 gate/security tests, 4 new repro) |

### Findings
**✅ Round-1 CORE both FIXED:** `_ip_is_trusted_proxy_hop(raw, nets) = _ip_is_loopback(raw) or _ip_in_networks(raw, nets)` unifies "trusted hop" in BOTH the peer-trust check (`peer_is_trusted = peer_is_loopback or _ip_in_networks(...)`, 5536 — loopback-denied regression closed) AND the XFF walk-skip (skips loopback hops too — `8.8.8.8, 127.0.0.1` now resolves to `8.8.8.8` → public rejected, trailing-loopback bypass closed). Direct-spoof still closed. 4 new repro tests pass.
**⛔ CORE (Codex, reproduced) — `X-Real-IP`-only bypass = REGRESSION vs master:** the fix only parses `X-Forwarded-For`. When XFF is absent, `_forwarded_client_ip_from_trusted_proxy` returns `raw_peer if peer_is_loopback else ""` (5555) — deliberately NOT falling back to `X-Real-IP` (the author's comment reasons a public caller could forge X-Real-IP). BUT this means: `raw_peer=127.0.0.1` (loopback proxy) + `X-Real-IP: 8.8.8.8` (real public client) + `HERMES_WEBUI_TRUST_FORWARDED_FOR=1` → resolves to `127.0.0.1` → `_onboarding_request_is_local()` returns **True** → public client passes the unauth gate. `origin/master` returned **False** here (its `_onboarding_request_is_local` at 5582-5583 considered `X-Real-IP` before raw peer). So for a same-host reverse proxy that forwards the client only via `X-Real-IP` (a very common nginx/Caddy config: `proxy_set_header X-Real-IP $remote_addr`), this PR REOPENS the unauth onboarding/terminal bypass the old code had closed. **Fix (Codex):** fail closed when XFF is absent but `X-Real-IP` (or `Forwarded`) is present — return `""` for that case (require trusted proxies to send a valid XFF chain), rather than trusting the loopback peer while ignoring a forwarded-header that says there's a real client behind it.

### Recommendation to the next agent / author + Nathan
**RED — gate-fail/changes-requested (1 CORE, a regression vs master): the fix hardened `X-Forwarded-For` but left `X-Real-IP`-only reverse-proxy deployments leaking public clients as local — worse than master. Fail closed when XFF is absent but `X-Real-IP`/`Forwarded` is present (return `""`, require a valid XFF chain from trusted proxies).** This is the same incomplete-vuln-class pattern as #5765 (harden one vector/header, miss the siblings) — the XFF path is now correct, but the sibling forwarded-headers (`X-Real-IP`, `Forwarded`) must be handled consistently (and not regress master's handling). **SECURITY + agent-authored → you, Nathan**; most-thorough-fix rule. concept 4/5 (XFF now right, X-Real-IP regressed). Author @Hinotoi-agent.

---
_Gate-certifier layer (warm-up → **gate** → release). r1 2 CORE FIXED (unified _ip_is_trusted_proxy_hop=loopback-OR-CIDR in peer-check + walk-skip; trailing-loopback bypass + loopback-denied regression closed, 4 new repro tests). BUT Codex reproduced a NEW CORE: XFF-absent + X-Real-IP:8.8.8.8 from a 127.0.0.1 proxy peer → returns raw loopback → local gate TRUE → public passes; master returned FALSE (considered X-Real-IP) = REGRESSION for X-Real-IP-only proxies. Fix: fail closed when XFF absent but X-Real-IP/Forwarded present. Same incomplete-class pattern as #5765 (harden XFF, miss sibling forwarded-headers). Suite green 12333. SECURITY + agent-authored → Nathan. Cert valid for sha:96db8528._

Hinotoi-agent:
Thanks — fixed this regression in `fbb9957e`.

Changes made:
- When trusted-forwarded mode is enabled and XFF is absent, `_forwarded_client_ip_from_trusted_proxy()` now fails closed if `X-Real-IP` or `Forwarded` is present.
- This prevents a loopback same-host proxy from making an `X-Real-IP`-only public client look local by falling back to the raw `127.0.0.1` peer.
- Kept the valid no-forwarded-header loopback proxy behavior intact.
- Added regressions for both `X-Real-IP`-only and RFC `Forwarded`-only loopback proxy cases across terminal and onboarding gates.

Validation run locally:
- `./scripts/test.sh tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py` → 42 passed
- `python3 -m compileall -q api/routes.py tests/test_cvd3_terminal_local_origin_gate.py tests/test_security_review_fixes.py`
- `git diff --check`

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (round 3: the X-Real-IP bypass is closed, but the fix over-corrected to fail-closed and now REJECTS legit private/LAN direct clients + private-proxy `TRUST_FORWARDED_FOR` deployments — 2 CORE regressions vs master) · SECURITY → Nathan · 3-round oscillation

**Certified head:** `sha:fbb9957e` (rebased onto current master, `git apply` clean) · **PR:** #5764 [security] · _Hinotoi-agent (agent-authored), resolve forwarded client before local gate_
**Verdict:** The round-2 X-Real-IP bypass is now closed (fail-closed on X-Real-IP/Forwarded without XFF), but Codex found the fix swung too far — 2 CORE regressions vs `origin/master`'s documented `HERMES_WEBUI_TRUST_FORWARDED_FOR` behavior: it now rejects legitimate private/LAN direct clients and private-proxy deployments. This security PR has oscillated across 3 rounds (too-loose → fixed-but-regressed-loopback → fixed-but-over-fails-closed). Suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5764c`) — SECURITY re-gate: deep Codex + full suite
| Gate | Result |
|------|--------|
| Codex (all chains closed? + master-parity regressions) | **SHIP ONLY WITH FIXES** — 2 CORE (both verified vs master) |
| Full pytest suite | ✅ **12337 passed, 0 failed** |

### Findings
**✅ Round-2 X-Real-IP bypass CLOSED:** XFF-absent + `X-Real-IP`/`Forwarded` present → `return ""` (fail closed). The public-client-via-X-Real-IP bypass is gone. Round-1 chains (trailing-loopback, loopback-denied) stay closed.
**⛔ CORE (routes.py:5563) — legit private/LAN direct client now rejected (regression):** raw peer `192.168.1.50`, NO forwarded headers, `TRUST_FORWARDED_FOR=1` → current returns **False**, master returned **True**. The final `return raw_peer if peer_is_loopback else ""` drops private (non-loopback) DIRECT clients. **Fix:** when NO forwarded-client header is present, fall back to `_ip_is_loopback_or_private(raw_peer)` — only fail closed when a forwarded header is present but unresolvable.
**⛔ CORE (routes.py:5536) — private-proxy `TRUST_FORWARDED_FOR` deployments now broken without the new undocumented env (regression):** raw peer `10.0.0.10` (a private reverse proxy), `X-Forwarded-For: 192.168.1.50`, no `HERMES_WEBUI_TRUSTED_PROXY_CIDRS` → current returns **False** (peer not loopback + empty CIDRs → not trusted), master returned **True**. Existing one-env (`TRUST_FORWARDED_FOR=1`) local/private deployments behind a non-loopback proxy now fail the onboarding/terminal gate unless they discover + set the NEW undocumented CIDR env. **Fix:** preserve the documented one-env trust path for PRIVATE proxy peers, while still walking XFF right-to-left to reject PUBLIC original clients.

### Recommendation to the next agent / author + Nathan
**RED — gate-fail/changes-requested (2 CORE regressions vs master). This PR has oscillated 3 rounds on the trust model — each fix over/under-shoots master's baseline. Recommend a COMPREHENSIVE master-parity approach before another round: write the full truth table of (raw_peer ∈ {loopback, private, public}) × (headers ∈ {none, XFF, X-Real-IP, Forwarded}) × (TRUST_FORWARDED_FOR, CIDRs) and, for each cell, assert BOTH "matches master's answer where master was safe" AND "rejects the public-original-client bypasses." Concretely: (1) no-forwarded-header → `_ip_is_loopback_or_private(raw_peer)` (restores private/LAN direct clients); (2) private proxy peer + XFF with `TRUST_FORWARDED_FOR=1` → trust + walk XFF (restores the documented one-env path) while still rejecting a public original client; (3) keep the fail-closed on X-Real-IP/Forwarded-without-XFF (the r2 fix); (4) keep the r1 loopback-OR-CIDR walk-skip.** The security intent is right but the trust model must be re-derived against master's documented behavior so it stops regressing legit deployments. **SECURITY + agent-authored + 3-round oscillation → you, Nathan** — may be worth a human security pass on the final truth table. concept 4/5. Author @Hinotoi-agent.

---
_Gate-certifier layer (warm-up → **gate** → release). r3: X-Real-IP bypass CLOSED (fail-closed on X-Real-IP/Forwarded-without-XFF) but OVER-CORRECTED: 2 CORE regressions vs master — (1) 5563 no-header private/LAN direct client rejected (fix: fall back to loopback_or_private not just loopback); (2) 5536 private-proxy TRUST_FORWARDED_FOR=1 deployments broken without the new undocumented CIDR env (fix: preserve documented one-env private-proxy trust + walk XFF). 3-round oscillation (loose→regressed-loopback→over-fail-closed). Recommend comprehensive master-parity truth table before another round. Suite green 12337. SECURITY + agent-authored → Nathan. Cert valid for sha:fbb9957e._

nesquena-hermes (changes requested):
Ran a full adversarial security re-gate (Codex spoof-matrix + head/base truth-table reproduction + real terminal-handler admission repros + 12,328-test full suite) on head `fbb9957e5c`. The `gate-pass` state does not hold on this head — **kicking back with 3 CORE findings.** The 42 focused tests pass but none of them cover these reproduced cases, and the full suite passing doesn't clear them (they're logic/boundary failures).

**1. Spoof still possible (the bug this PR is meant to close is still open by another route).** The XFF parse uses only the first field, so an attacker can put a trusted/loopback value first and hide a later public field. Reproduced: a crafted chain admits a public client as "local."
- **Fix:** consume all values with `get_all()`, preserve wire order across multiple `X-Forwarded-For` headers, and reject present-but-empty / empty-hop / malformed / otherwise-unresolved chains (fail closed).

**2. Regression — breaks direct private/LAN clients (`api/routes.py:5536`).** With `HERMES_WEBUI_TRUST_FORWARDED_FOR=1`, a direct client with **no** forwarded header is now rejected. Verified: raw `192.168.1.50` → `false` at this head, `true` at base. This blocks onboarding, first-password/passkey setup, and passwordless terminal access for LAN users.
- **Fix:** when no forwarded-client header is present, classify the raw socket peer directly via `_ip_is_loopback_or_private()` **before** applying the proxy-trust requirement.

**3. Compat break — existing single-env-var deployments (`api/routes.py:5536`).** Deployments using the documented single `HERMES_WEBUI_TRUST_FORWARDED_FOR=1` now fail unless they also set the new, undocumented CIDR variable. Verified: private peer `10.0.0.10` + XFF `192.168.1.50` is allowed by base, rejected by this PR.
- **Fix:** preserve a clearly-defined compatibility mode when the CIDR variable is absent, distinguishing *absent* from *invalid-configured*, while still walking a canonical sanitized XFF chain.

**Recommendation:** `api/routes.py:5521-5563` needs a single security-reviewed trust model with a complete peer/header truth table (raw-peer × header-present × trusted-vs-untrusted × loopback/private/public × malformed), rather than the current design that fixes one path inline and oscillates another. A regression test matrix covering all of the above should land with it.

⚠️ **Note on the underlying issue:** the original trusted-proxy bypass this PR targets is real and still unresolved on master — so this isn't "revert and forget." Once you re-push a corrected trust model I'll re-gate the full truth table. Thanks for tackling a genuinely tricky boundary — the concept is right, the trust model just needs to be airtight on both directions (no spoof-in, no legit-lockout).

PR #5765, opened 2026-07-08
[security] fix(models): stop credentialed live-model redirects

Hinotoi-agent:
## Summary
This PR hardens live model discovery so credentialed provider probes do not follow redirects while carrying provider API keys.

- Adds a no-redirect JSON helper for credentialed model-discovery requests.
- Routes both custom-provider and OpenAI-compatible fallback live-model probes through that helper.
- Adds regression tests with a redirector and target server to prove the bearer token is not forwarded to the redirect target.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Credentialed live-model probes followed redirects with `Authorization` attached | A configured or compromised provider endpoint could redirect `/models` discovery to another host and receive stored provider API keys | High |

## Before this PR
- `/api/models/live?provider=custom:<name>` built a `urllib.request.Request` with `Authorization: Bearer ***` and passed it to default `urllib.request.urlopen()`.
- The OpenAI-compatible fallback did the same for configured provider keys.
- Default urllib redirect handling can preserve request headers across redirects, including `Authorization`.
- A 3xx from the configured models endpoint could therefore forward the provider key to the redirect target.

## After this PR
- Credentialed live-model probes use `_credentialed_json_get_no_redirect()`.
- The helper installs an `HTTPRedirectHandler` whose `redirect_request()` returns `None`, so redirects fail closed rather than being followed with credentials.
- The existing fallback behavior remains: failed live discovery falls through to static/configured models where applicable.
- Tests prove the redirect target is not reached for both custom-provider and OpenAI-compatible fallback probes.

## Why this matters
Provider API keys can carry billing authority, model access, or tenant-level credentials. Live model discovery is a background convenience path, but it still sends real provider secrets. Redirects from a credentialed endpoint should not move those secrets to a different host.

The repository already contains LM Studio reasoning-probe coverage documenting the same urllib behavior: credentialed probes must not follow redirects because urllib can re-send headers across redirects.

This PR applies that same no-redirect security invariant to the live model discovery paths in `api/routes.py`, including custom providers and OpenAI-compatible fallback providers.

## Attack flow

```text
configured provider models endpoint
    -> returns 302 Location: http://attacker.example/steal
        -> urllib follows redirect with Authorization still attached
            -> redirect target receives Bearer provider API key
```

## Affected code

| Issue | Files |
|-------|-------|
| Credentialed live-model redirect key disclosure | `api/routes.py`, `tests/test_security_live_models_redirects.py` |

## Root cause
Credentialed live-model redirect key disclosure:
- The affected paths used default `urllib.request.urlopen()` for requests carrying `Authorization`.
- urllib redirect handling can preserve request headers across redirects, so the credential boundary was not pinned to the configured models endpoint.

## CVSS assessment

| Credentialed live-model redirect key disclosure | 8.1 High | `CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:N` |

Rationale:
- The attacker needs influence over a configured provider endpoint or a compromised provider-compatible endpoint. Impact is direct disclosure of stored provider credentials that may allow unauthorized model access or billing abuse. Severity may vary with deployment and provider-key scope.

## Safe reproduction steps
1. Configure a custom provider with a fake key such as `CANARY_SECRET` and a local models endpoint.
2. Make that endpoint return:
   ```http
   HTTP/1.1 302 Found
   Location: http://127.0.0.1:<target>/steal
   ```
3. Call `/api/models/live?provider=custom:<name>`.
4. Before this PR, default urllib behavior could follow the redirect with `Authorization` still attached.
5. After this PR, the redirect target is never reached.

The regression tests run this proof with local HTTP servers and a fake bearer token only.

- A credentialed provider probe should not forward `Authorization` to a redirect target.
- The pre-patch behavior used default redirect-following urllib requests.
- The safe proof signal is that only the redirector receives `Bearer CANARY_SECRET`; the redirected target receives no request.

## Changes in this PR
- Adds `_credentialed_json_get_no_redirect()` for credentialed provider-model JSON fetches.
- Uses the helper in the custom-provider live-model fetch path.
- Uses the helper in the OpenAI-compatible fallback live-model fetch path.
- Adds redirect regression tests for the helper, custom providers, and `_OPENAI_COMPAT_ENDPOINTS` fallback providers.

## Files changed

| Backend provider probe | `api/routes.py` | Credentialed live-model probes now fail closed on redirects |
| Regression tests | `tests/test_security_live_models_redirects.py` | Added local redirector/target tests proving provider keys are not forwarded onward |

## Maintainer impact
- The patch is scoped to live model discovery and does not change static model fallback behavior.
- Providers that return redirects from credentialed `/models` endpoints will now be treated as live-fetch failures instead of being followed.
- This is a fail-closed behavior for credentialed background discovery; configured/static model lists still provide fallback coverage.

## Fix rationale
- The no-redirect helper is narrow and reusable for credentialed provider probes.
- Failing closed on any redirect is simpler and safer than trying to decide when it is acceptable to forward provider credentials.
- Tests cover both affected live-model callsite families.

## Type of change
- [x] Security fix
- [x] Tests
- [ ] Documentation update

## Test plan
- [x] Focused live-model redirect tests
- [x] Diff-scoped Python lint
- [x] Syntax compilation for touched Python files
- [x] Whitespace diff check

Executed with:
- `./scripts/test.sh tests/test_security_live_models_redirects.py`
- `python3 -m compileall -q api/routes.py tests/test_security_live_models_redirects.py`
- `git diff --check`

## Disclosure notes
- The tests use only local loopback HTTP servers and a fake `CANARY_SECRET` token.
- No real provider credentials or production endpoints were used.

## AI Usage Disclosure
AI assistance was used to identify the boundary, draft the patch, and write regression tests. The submitted change is scoped to the files and validation listed above.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -1258,6 +1258,26 @@ def _get_anthropic_fallback_env_vars() -> tuple[str, ...]:
         return fallback
 
 
+_AGENT_CREDENTIAL_ENV_NAMES_BY_PROVIDER = {
+    "anthropic": _get_anthropic_fallback_env_vars(),
+    "deepseek": ("DEEPSEEK_API_KEY",),
+    "gemini": ("GEMINI_API_KEY", "GOOGLE_API_KEY"),
+    "gmi": ("GMI_API_KEY",),
+    "lmstudio": ("LM_API_KEY", "LMSTUDIO_API_KEY"),
+    "minimax": ("MINIMAX_API_KEY",),
+    "mistralai": ("MISTRAL_API_KEY",),
+    "nvidia": ("NVIDIA_API_KEY",),
+    "openai": ("OPENAI_API_KEY",),
+    "openai-api": ("OPENAI_API_KEY",),
+    "openrouter": ("OPENROUTER_API_KEY",),
+    "stepfun": ("STEPFUN_API_KEY",),
+    "xai": ("XAI_API_KEY",),
+    "zai": ("ZAI_API_KEY", "GLM_API_KEY"),
+}
+
+_AGENT_ALWAYS_SKIP_PROVIDER_MODEL_IDS_PROVIDERS = frozenset({"copilot", "copilot-acp", "nous"})
+
+
 def _resolve_provider_alias(name: str) -> str:
     """Return the canonical provider slug for *name*.
 
@@ -5637,19 +5657,10 @@ def _configured_provider_api_key_for_live_probe(provider_id: str) -> str:
             if key:
                 return key
 
-    env_names_by_provider = {
-        "anthropic": _get_anthropic_fallback_env_vars(),
-        "deepseek": ("DEEPSEEK_API_KEY",),
-        "gemini": ("GEMINI_API_KEY", "GOOGLE_API_KEY"),
-        "lmstudio": ("LM_API_KEY", "LMSTUDIO_API_KEY"),
-        "minimax": ("MINIMAX_API_KEY",),
-        "mistralai": ("MISTRAL_API_KEY",),
-        "nvidia": ("NVIDIA_API_KEY",),
-        "openrouter": ("OPENROUTER_API_KEY",),
-        "xai": ("XAI_API_KEY",),
-        "zai": ("ZAI_API_KEY", "GLM_API_KEY"),
-    }
-    for env_name in env_names_by_provider.get(provider, ()):  # pragma: no branch - tiny tuple
+    for env_name in _AGENT_CREDENTIAL_ENV_NAMES_BY_PROVIDER.get(
+        provider,
+        (),
+    ):  # pragma: no branch - tiny tuple
         key = str(os.getenv(env_name) or "").strip()
         if key:
             return key
@@ -5670,7 +5681,10 @@ def _read_live_provider_model_ids(provider_id: str) -> list[str]:
     if not pid:
         return []
     resolved_pid = _resolve_provider_alias(pid)
-    if resolved_pid == "copilot" or _configured_provider_api_key_for_live_probe(resolved_pid):
+    if (
+        resolved_pid in _AGENT_ALWAYS_SKIP_PROVIDER_MODEL_IDS_PROVIDERS
+        or _configured_provider_api_key_for_live_probe(resolved_pid)
+    ):
         # Credentialed provider_model_ids() calls live endpoints inside
         # hermes_cli, where WebUI cannot enforce no-redirect transport. Fall
         # back to WebUI-owned static/direct discovery unless the probe is known
--- a/api/config.py
+++ b/api/config.py
@@ -3288,15 +3288,34 @@ def _models_dev_reasoning_efforts(model_id: str, provider_id: str) -> list[str] 
 class _NoRedirectHandler(urllib.request.HTTPRedirectHandler):
     """urllib redirect handler that refuses to follow any redirect.
 
-    Used by the LM Studio reasoning probe so a 3xx from the probe URL can never
-    forward the ``Authorization`` header (the configured LM Studio key) to a
-    redirected, possibly attacker-controlled host. ``redirect_request``
-    returning ``None`` makes urllib raise the original 3xx as an ``HTTPError``,
-    which the probe swallows. (#3837 security review)
+    Used by credentialed model probes so a 3xx from the probe URL can never
+    forward the ``Authorization`` header to a redirected, possibly
+    attacker-controlled host. ``redirect_request`` returning ``None`` makes
+    urllib raise the original 3xx as an ``HTTPError``, which callers swallow and
+    treat as an unavailable live catalog. (#3837 security review)
     """
 
     def redirect_request(self, req, fp, code, msg, headers, newurl):
         return None
+
+
+def _credentialed_json_get_no_redirect(
+    url: str,
+    headers: dict[str, str],
+    *,
+    timeout: float,
+):
+    """Fetch credentialed model-discovery JSON without following redirects.
+
+    ``urllib.request.urlopen`` follows redirects and can re-send request headers
+    such as ``Authorization`` to the redirect target. All WebUI-owned model
+    discovery paths that attach provider credentials should use this helper so a
+    3xx fails closed while preserving urllib's default proxy handling.
+    """
+    request = urllib.request.Request(url, headers=headers, method="GET")
+    opener = urllib.request.build_opener(_NoRedirectHandler)
+    with opener.open(request, timeout=timeout) as response:  # nosec B310
+        return json.loads(response.read().decode("utf-8", errors="replace"))
 
 
 def _get_lmstudio_reasoning_probe_api_key() -> str | None:
@@ -5599,6 +5618,44 @@ def _get_label_for_model(model_id: str, existing_groups: list) -> str:
     )
 
 
+def _configured_provider_api_key_for_live_probe(provider_id: str) -> str:
+    """Return a visible provider credential that would make live probes credentialed."""
+    provider = _resolve_provider_alias(str(provider_id or "").strip().lower())
+    if not provider:
+        return ""
+
+    provider_cfg = _get_provider_cfg(provider)
+    key = str(provider_cfg.get("api_key") or "").strip()
+    if key:
+        return key
+
+    model_cfg = cfg.get("model") or {}
+    if isinstance(model_cfg, dict):
+        active_provider = _resolve_provider_alias(str(model_cfg.get("provider") or "").strip().lower())
+        if active_provider == provider:
+            key = str(model_cfg.get("api_key") or "").strip()
+            if key:
+                return key
+
+    env_names_by_provider = {
+        "anthropic": _get_anthropic_fallback_env_vars(),
+        "deepseek": ("DEEPSEEK_API_KEY",),
+        "gemini": ("GEMINI_API_KEY", "GOOGLE_API_KEY"),
+        "lmstudio": ("LM_API_KEY", "LMSTUDIO_API_KEY"),
+        "minimax": ("MINIMAX_API_KEY",),
+        "mistralai": ("MISTRAL_API_KEY",),
+        "nvidia": ("NVIDIA_API_KEY",),
+        "openrouter": ("OPENROUTER_API_KEY",),
+        "xai": ("XAI_API_KEY",),
+        "zai": ("ZAI_API_KEY", "GLM_API_KEY"),
+    }
+    for env_name in env_names_by_provider.get(provider, ()):  # pragma: no branch - tiny tuple
+        key = str(os.getenv(env_name) or "").strip()
+        if key:
+            return key
+    return ""
+
+
 def _read_live_provider_model_ids(provider_id: str) -> list[str]:
     """Return live model IDs from Hermes CLI for a provider, or [] on failure.
 
@@ -5611,6 +5668,13 @@ def _read_live_provider_model_ids(provider_id: str) -> list[str]:
     """
     pid = str(provider_id or "").strip()
     if not pid:
+        return []
+    resolved_pid = _resolve_provider_alias(pid)
+    if resolved_pid == "copilot" or _configured_provider_api_key_for_live_probe(resolved_pid):
+        # Credentialed provider_model_ids() calls live endpoints inside
+        # hermes_cli, where WebUI cannot enforce no-redirect transport. Fall
+        # back to WebUI-owned static/direct discovery unless the probe is known
+        # to be keyless.
         return []
     try:
         from hermes_cli.models import provider_model_ids as _provider_model_ids
@@ -6303,7 +6367,6 @@ def get_available_models(*, prefer_cache: bool = False, force_refresh: bool = Fa
             try:
                 import ipaddress
                 import urllib.error
-                import urllib.request
                 import socket
 
                 endpoint_url = _models_endpoint_for_base_url(base)
@@ -6342,12 +6405,12 @@ def get_available_models(*, prefer_cache: bool = False, force_refresh: bool = Fa
                     except socket.gaierror:
                         pass
 
-                req = urllib.request.Request(endpoint_url, method="GET")
-                req.add_header("User-Agent", "OpenAI/Python 1.0")
-                for k, v in headers.items():
-                    req.add_header(k, v)
-                with urllib.request.urlopen(req, timeout=CUSTOM_MODELS_ENDPOINT_TIMEOUT_SECONDS) as response:  # nosec B310
-                    data = json.loads(response.read().decode("utf-8"))
+                headers = {"User-Agent": "OpenAI/Python 1.0", **headers}
+                data = _credentialed_json_get_no_redirect(
+                    endpoint_url,
+                    headers,
+                    timeout=CUSTOM_MODELS_ENDPOINT_TIMEOUT_SECONDS,
+                )
                 return _extract_model_entries_from_payload(data, provider), None
             except urllib.error.HTTPError as exc:
                 error = _custom_endpoint_error(provider, exc, code=getattr(exc, "code", None))
@@ -6951,10 +7014,7 @@ def get_available_models(*, prefer_cache: bool = False, force_refresh: bool = Fa
                                 headers["Authorization"] = f"Bearer {lm_api_key}"
                             endpoint = (lm_base_url + "/models").rstrip("/")
                             try:
-                                import urllib.request as _urlreq
-                                req = _urlreq.Request(endpoint, method="GET", headers=headers)
-                                with _urlreq.urlopen(req, timeout=5) as resp:
-                                    lm_data = json.loads(resp.read().decode())
+                                lm_data = _credentialed_json_get_no_redirect(endpoint, headers, timeout=5)
                                 for m in (lm_data.get("data") or []):
                                     if isinstance(m, dict):
                                         mid = str(m.get("id") or "").strip()
--- a/api/routes.py
+++ b/api/routes.py
@@ -18224,6 +18224,25 @@ def _handle_clarify_inject(handler, parsed):
     return j(handler, {"error": "session_id required"}, status=400)
 
 
+def _credentialed_json_get_no_redirect(url: str, headers: dict[str, str], *, timeout: float):
+    """Fetch JSON for credentialed provider probes without following redirects.
+
+    ``urllib.request.urlopen`` preserves request headers across redirects.  Model
+    discovery attaches provider credentials, so a 3xx from the configured models
+    endpoint must fail closed instead of forwarding ``Authorization`` to the
+    redirect target.
+    """
+
+    class _NoRedirectProviderProbeHandler(HTTPRedirectHandler):
+        def redirect_request(self, req, fp, code, msg, headers, newurl):
+            return None
+
+    req = Request(url, headers=headers)
+    opener = build_opener(ProxyHandler({}), _NoRedirectProviderProbeHandler())
+    with opener.open(req, timeout=timeout) as resp:
+        return json.loads(resp.read())
+
+
 def _handle_live_models(handler, parsed):
     """Return the live model list for a provider.
 
@@ -18405,13 +18424,11 @@ def _handle_live_models(handler, parsed):
                         else:
                             _models_url = f"{_ep}/v1/models"
                         
-                        _req = urllib.request.Request(
+                        _body = _credentialed_json_get_no_redirect(
                             _models_url,
-                            headers={"Authorization": f"Bearer {_api_key}"},
+                            {"Authorization": f"Bearer {_api_key}"},
+                            timeout=CUSTOM_MODELS_ENDPOINT_TIMEOUT_SECONDS,
                         )
-                        
-                        with urllib.request.urlopen(_req, timeout=CUSTOM_MODELS_ENDPOINT_TIMEOUT_SECONDS) as _resp:
-                            _body = json.loads(_resp.read())
                         
                         # Parse response: {"data": [{"id": "model1", ...}, ...]}
                         if isinstance(_body, dict):
@@ -18471,12 +18488,11 @@ def _handle_live_models(handler, parsed):
                             if _active_provider == provider:
                                 _key = _model_cfg.get("api_key")
                     if _key:
-                        _req = urllib.request.Request(
+                        _body = _credentialed_json_get_no_redirect(
                             f"{_ep}/models",
-                            headers={"Authorization": f"Bearer {_key}"},
+                            {"Authorization": f"Bearer {_key}"},
+                            timeout=8,
                         )
-                        with urllib.request.urlopen(_req, timeout=8) as _resp:
-                            _body = json.loads(_resp.read())
                         ids = [m.get("id", "") for m in _body.get("data", []) if m.get("id")]
                         logger.debug("Live-fetched %d models from %s /v1/models", len(ids), provider)
                 except Exception as _fetch_err:
--- a/api/routes.py
+++ b/api/routes.py
@@ -18238,7 +18238,7 @@ def _credentialed_json_get_no_redirect(url: str, headers: dict[str, str], *, tim
             return None
 
     req = Request(url, headers=headers)
-    opener = build_opener(ProxyHandler({}), _NoRedirectProviderProbeHandler())
+    opener = build_opener(_NoRedirectProviderProbeHandler())
     with opener.open(req, timeout=timeout) as resp:
         return json.loads(resp.read())
 
--- a/api/routes.py
+++ b/api/routes.py
@@ -18265,18 +18265,29 @@ def _configured_provider_api_key_for_live_probe(provider: str, cfg: dict) -> str
             if key:
                 return key
 
-    env_names_by_provider = {
-        "anthropic": ("ANTHROPIC_API_KEY", "ANTHROPIC_TOKEN", "CLAUDE_CODE_OAUTH_TOKEN"),
-        "deepseek": ("DEEPSEEK_API_KEY",),
-        "gemini": ("GEMINI_API_KEY", "GOOGLE_API_KEY"),
-        "minimax": ("MINIMAX_API_KEY",),
-        "mistralai": ("MISTRAL_API_KEY",),
-        "nvidia": ("NVIDIA_API_KEY",),
-        "openrouter": ("OPENROUTER_API_KEY",),
-        "xai": ("XAI_API_KEY",),
-        "zai": ("ZAI_API_KEY", "GLM_API_KEY"),
-    }
-    for env_name in env_names_by_provider.get(provider, ()):  # pragma: no branch - tiny tuple
+    try:
+        from api.config import _AGENT_CREDENTIAL_ENV_NAMES_BY_PROVIDER
+    except Exception:
+        _AGENT_CREDENTIAL_ENV_NAMES_BY_PROVIDER = {
+            "anthropic": ("ANTHROPIC_API_KEY", "ANTHROPIC_TOKEN", "CLAUDE_CODE_OAUTH_TOKEN"),
+            "deepseek": ("DEEPSEEK_API_KEY",),
+            "gemini": ("GEMINI_API_KEY", "GOOGLE_API_KEY"),
+            "gmi": ("GMI_API_KEY",),
+            "lmstudio": ("LM_API_KEY", "LMSTUDIO_API_KEY"),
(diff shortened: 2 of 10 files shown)


Hinotoi-agent:
Thanks — fixed in `a12fe1d1`.

Changes:
- Removed `ProxyHandler({})` from the credentialed live-model helper so `build_opener()` preserves the default urllib proxy handling from `HTTP_PROXY` / `HTTPS_PROXY`.
- Kept the no-redirect handler in place, so credentialed model discovery still fails closed on redirects.
- Updated existing BYOK live-model tests to monkeypatch `_credentialed_json_get_no_redirect()` directly now that the code no longer calls `urllib.request.urlopen()` at those sites.

Validation:
```bash
./scripts/test.sh tests/test_byok_model_dropdown.py tests/test_security_live_models_redirects.py
# 27 passed
python3 -m compileall -q api/routes.py tests/test_byok_model_dropdown.py tests/test_security_live_models_redirects.py
git diff --check
```

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (the no-redirect hardening is correct but INCOMPLETE — the same credential-leak-via-redirect still exists on ≥3 other credentialed model-probe paths the PR didn't touch) · SECURITY → Nathan

**Certified head:** `sha:<rebased>` (rebased onto current master, `git apply` clean) · **PR:** #5765 [security] · _Hinotoi-agent (agent-authored), stop credentialed live-model redirects_
**Verdict:** The `_credentialed_json_get_no_redirect` helper is correctly built and blocks redirects on the two probe sites it's wired into — but Codex found the SAME credential-leak-via-redirect vulnerability still open on ≥3 OTHER credentialed model-discovery paths the PR left untouched. A partial fix on a credential-leak class shouldn't ship (false-assurance risk). Suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5765`) — SECURITY gate: deep Codex + full suite
| Gate | Result |
|------|--------|
| Codex (all-paths / redirect-truly-blocked / other-leak-vectors) | **SHIP ONLY WITH FIXES** — 3 SILENT (same class, uncovered paths) |
| Full pytest suite | ✅ **12326 passed, 0 failed** (+ 27 security/byok tests) |

### Findings
**✅ The helper itself is correct:** `_NoRedirectProviderProbeHandler.redirect_request` returns `None` → urllib raises `HTTPError` on a 3xx (redirect truly blocked, credential not re-sent), and the two wired sites (custom-provider `routes.py:18427` + OpenAI-compatible fallback `18491`) are hardened. The dedicated `test_security_live_models_redirects.py` passes.
**⛔ SILENT ×3 (Codex) — the same leak is still open on untouched credentialed probe paths:**
- `routes.py:18304` — `_handle_live_models()` calls `hermes_cli.models.provider_model_ids(provider)` FIRST, and the installed agent code still has credentialed `urllib.request.urlopen` probes (Anthropic/Copilot/API-compatible) that follow redirects → the key can leak before the no-redirect helper is even reached. Fix: route credentialed discovery through no-redirect fetches before the agent live probes, or patch the agent resolver so every credentialed probe refuses redirects.
- `config.py:6349` — `_read_custom_endpoint_models()` attaches the key (6311-6313) then uses plain `urlopen` (follows redirects). Fix: use the no-redirect helper + catch 3xx `HTTPError` as a graceful no-models.
- `config.py:6956` — `/api/models` LM Studio fallback follows redirects with `Authorization` when `providers.lmstudio.api_key` is set. Fix: same no-redirect pattern.

### Recommendation to the next agent / author + Nathan
**RED — gate-fail/changes-requested: the fix is correct but INCOMPLETE for the security class. Harden ALL credentialed model-discovery probe sites, not just the two in this PR — at minimum `config.py:6349` (custom endpoint), `config.py:6956` (LM Studio fallback), and `routes.py:18304` (the agent-delegated `provider_model_ids` path, incl. Anthropic/Copilot/API-compatible). Ideally factor the no-redirect fetch into the shared helper and route every credentialed probe through it (grep every `urlopen`/`requests.get`/`build_opener` that attaches a Bearer/api-key for model discovery).** A partial credential-leak fix is risky — it looks like the class is closed while ≥3 paths still leak. **SECURITY + agent-authored → routes to you, Nathan**, and per the standing rule a security primitive gets the most-thorough fix (harden every site) before ship. concept 4/5 (right primitive, incomplete coverage). Author @Hinotoi-agent.

---
_Gate-certifier layer (warm-up → **gate** → release). _credentialed_json_get_no_redirect (redirect_request→None→HTTPError) is correct + wired to 2 probe sites (routes.py:18427/18491), but Codex found the SAME credential-leak-via-redirect still open on ≥3 untouched credentialed probe paths: routes.py:18304 (agent provider_model_ids Anthropic/Copilot/API-compat), config.py:6349 (custom endpoint), config.py:6956 (LM Studio fallback). Partial security fix = bounce: harden ALL credentialed model-probe sites via the shared no-redirect helper. Suite green 12326 + 27 sec/byok tests. SECURITY + agent-authored → Nathan. Cert valid for the rebased head._

Hinotoi-agent:
Thanks for the deeper pass — fixed in `ea7836d6`.

Changes:
- Added a WebUI-owned `_credentialed_json_get_no_redirect()` helper in `api.config` and routed the remaining credentialed model-discovery fallbacks through it.
- Hardened custom endpoint model discovery (`api/config.py`) so redirects fail closed instead of forwarding `Authorization`.
- Hardened LM Studio `/models` fallback (`api/config.py`) the same way.
- Avoided calling `hermes_cli.models.provider_model_ids()` from `/api/models/live` when the probe may be credentialed and WebUI cannot enforce no-redirect transport there; WebUI-owned direct/static fallback paths are used instead.
- Applied the same guard in `api.config._read_live_provider_model_ids()` so full catalog rebuilds do not delegate credentialed live probes to agent code that may follow redirects.
- Added/updated regressions for the route/config no-redirect helpers and credentialed OpenAI-compatible route path.

Validation:
```bash
./scripts/test.sh \
  tests/test_security_live_models_redirects.py \
  tests/test_byok_model_dropdown.py \
  tests/test_issue1527_lmstudio_base_url_classification.py \
  tests/test_issue1105_ssrf_custom_providers.py \
  tests/test_issue1240_generic_cli_catalog_sync.py \
  tests/test_copilot_provider_model_settings_not_allowlist.py \
  tests/test_issue3750_lmstudio_probe_auth.py \
  tests/test_issue3718_live_models_custom_probe.py \
  tests/test_issue1426_openrouter_free_tier_live_fetch.py
# 66 passed

python3 -m compileall -q api/routes.py api/config.py tests/test_security_live_models_redirects.py tests/test_byok_model_dropdown.py tests/test_issue1527_lmstudio_base_url_classification.py
git diff --check
```

nesquena-hermes:
Re-gated the `ea7836d6` re-push against the three RED findings from the last pass (read-only worktree, diff on `api/config.py`, `api/routes.py`, `tests/test_security_live_models_redirects.py`, cross-referenced against the agent probe code). The direct-leak paths are now genuinely closed, and the delegated-path handling is on the right layer — but the skip heuristic's provider coverage is narrower than the agent's actual credentialed-probe set, so one realistic vector is still open.

## What's now correctly fixed ✔

- **Custom endpoint** (`config.py:6405`) and **LM Studio fallback** (`config.py:7014`) both route through the new `_credentialed_json_get_no_redirect` helper — `build_opener(_NoRedirectHandler)` → 3xx raises `HTTPError`, key never re-sent. Good.
- **OpenAI-compat + custom route probes** (`routes.py:18474`, `18538`) use the routes-local `_credentialed_json_get_no_redirect` (`routes.py:18227`). `Request`/`build_opener`/`HTTPRedirectHandler` are all imported at `routes.py:36`, so no NameError.
- **Delegated path** (`routes.py:18342`): the `_should_skip_agent_provider_model_ids` gate is the correct approach — WebUI can't wrap the agent's urllib in a no-redirect opener (I confirmed `hermes_cli/models.py` has zero redirect handling; `probe_api_models` attaches `Authorization: Bearer` at `models.py:3562` and calls plain `urlopen` at `models.py:3577`), so skipping and falling back to WebUI-owned discovery is the right layer.

The `test_security_live_models_redirects.py` cases cover both helpers plus the custom + openai-compat route paths. Suite-green claim is plausible.

## Residual: skip heuristic misses two credentialed agent probes ⛔

The gate is provider-list-based, but the list is narrower than what the agent actually probes credentialed:

1. **`openai` / `openai-api` (env-key, the realistic one).** The agent probes OpenAI at `models.py:2368`:

```python
api_key = os.getenv("OPENAI_API_KEY", "").strip()
if api_key:
    base = os.getenv("OPENAI_BASE_URL", "").strip().rstrip("/") or "https://api.openai.com/v1"
    ...
    live = fetch_api_models(api_key, base)   # -> probe_api_models, follows redirects
```

`OPENAI_API_KEY` + a custom `OPENAI_BASE_URL` is a common config. Neither the routes `env_names_by_provider` map (`routes.py:18268`) nor the config one (`config.py:5640`) lists `openai`, so `_should_skip_agent_provider_model_ids("openai", cfg)` returns `False`, the agent probe runs, and it follows a 3xx to an attacker-controlled `Location` with the Bearer attached. This is the same leak class, still open.

2. **`copilot-acp`.** `_should_skip_...` checks `provider == "copilot"` literally, but `_resolve_provider_alias` keeps `copilot-acp` distinct (agent alias table maps `github-copilot-acp` → `copilot-acp`, not `copilot`). The agent handles both in one branch — `if normalized in {"copilot", "copilot-acp"}:` (`models.py:2294`) → `_fetch_github_models(...)` credentialed. Fixed host, so lower practical risk, but it bypasses the literal check.

Also worth a glance: `stepfun` / `gmi` / `nous` resolve creds via `resolve_api_key_provider_credentials()` with a configurable `base_url` and aren't in either map.

## Suggested fix

Key the openai gap by adding `"openai": ("OPENAI_API_KEY",)` (and `openai-api`) to both env maps, and change the copilot literal to `if provider in {"copilot", "copilot-acp"}:`. Cleaner still would be a single canonical set of "providers whose agent probe is credentialed" shared by both `_should_skip_...` implementations, so the two maps can't drift. A regression asserting `provider=openai` with only `OPENAI_API_KEY` in env skips the agent probe would lock it.

Direction and the helper are solid — this is a coverage tightening on the delegate gate, not a rework.

Hinotoi-agent:
Thanks — tightened the delegate gate in `2ba6d8d8`.

What changed:
- Added OpenAI/OpenAI API env-key coverage (`OPENAI_API_KEY`) so `provider=openai` / `openai-api` no longer delegates to `hermes_cli.models.provider_model_ids()` when that agent path could attach credentials and follow redirects.
- Added `copilot-acp` to the always-skip delegate set, alongside OAuth/token-pool-backed providers where WebUI cannot enforce no-redirect transport inside the agent call.
- Centralized the visible env-key map used by the route-level and config-level skip checks, including `gmi` and `stepfun` API-key env coverage.
- Added regressions for the OpenAI env-key route path, config-level `openai-api`, and `copilot-acp` delegate skip behavior.

Validation:
- `./scripts/test.sh tests/test_security_live_models_redirects.py tests/test_byok_model_dropdown.py tests/test_issue1527_lmstudio_base_url_classification.py tests/test_issue1105_ssrf_custom_providers.py tests/test_issue1240_generic_cli_catalog_sync.py tests/test_copilot_provider_model_settings_not_allowlist.py tests/test_issue3750_lmstudio_probe_auth.py tests/test_issue3718_live_models_custom_probe.py tests/test_issue1426_openrouter_free_tier_live_fetch.py` → 69 passed
- `python3 -m compileall -q api/routes.py api/config.py tests/test_security_live_models_redirects.py tests/test_byok_model_dropdown.py tests/test_issue1527_lmstudio_base_url_classification.py` → passed
- `git diff --check` → passed

nesquena-hermes (changes requested):
Re-gated the `2ba6d8d8` re-push against the three RED findings from the 13:30 gate cert (read-only worktree; diff on `api/config.py`, `api/routes.py`, `tests/test_security_live_models_redirects.py`; cross-referenced the agent probe code). **The security work converged** — nice, thorough pass:

- ✅ `openai` / `openai-api` (the realistic `OPENAI_API_KEY` case) now covered — added to the new canonical `_AGENT_CREDENTIAL_ENV_NAMES_BY_PROVIDER` and shared into both the config and routes skip checks.
- ✅ `copilot-acp` (and `nous`) now handled via `_AGENT_ALWAYS_SKIP_PROVIDER_MODEL_IDS_PROVIDERS = frozenset({"copilot","copilot-acp","nous"})` with a membership test, not the old literal `== "copilot"`.
- ✅ The prior `_credentialed_json_get_no_redirect` / `_NoRedirect*` helpers are intact (no walk-back), and the new `test_security_live_models_redirects.py` cases pin all three (7 passed for me).
- ✅ Folding both maps into a single canonical set in `api.config` is exactly the "so they can't drift" shape the cert asked for.

**One blocker before this can ship: the convergence commit broke a previously-green sibling suite.** CI is red on 12 of 15 test shards — all from `tests/test_live_models_ttl_cache.py` (5 pre-existing cache tests this PR doesn't touch):

```
FAILED tests/test_live_models_ttl_cache.py::test_live_models_cache_expires
  assert [] == ['openai', 'openai']
FAILED tests/test_live_models_ttl_cache.py::test_live_models_cache_is_profile_scoped
  assert [] == [('default','openai'), ('research','openai')]
FAILED tests/test_live_models_ttl_cache.py::test_live_models_cache_returns_deep_copies
  assert [{'id':'gpt-...'}] == [{'id':'openai/gpt-test',...}]
```

**Root cause:** those tests mock `config.get_config → {"model":{"provider":"openai"}}` (no config key) and mock `provider_model_ids`, expecting the agent probe to be reached — but they never clear `OPENAI_API_KEY` from the environment. On master `openai` wasn't in any credentialed env-skip map, so an ambient `OPENAI_API_KEY` was harmless. Now that `openai` → `OPENAI_API_KEY` is (correctly) in the map, an ambient key in CI makes `_configured_provider_api_key_for_live_probe("openai")` truthy → `_read_live_provider_model_ids` skips the agent probe → `calls == []`.

I reproduced it deterministically in the sandbox: the file passes with a clean env (5 passed), and fails identically to CI when `OPENAI_API_KEY` is present (5 failed).

**Fix (mechanical, test-side — the production behavior is correct and is what the cert recommended):** make the shared helper env-deterministic. In `tests/test_live_models_ttl_cache.py`, `_patch_live_models_basics` (~line 17):

```python
def _patch_live_models_basics(monkeypatch, routes, profile="default"):
    import api.config as config
    import api.profiles as profiles
    for _k in ("OPENAI_API_KEY", "OPENAI_BASE_URL"):
        monkeypatch.delenv(_k, raising=False)   # openai is now a credentialed-skip provider; an ambient key skips the agent probe
    routes._clear_live_models_cache()
    ...
```

(Equivalent: neutralize the skip predicate with `monkeypatch.setattr(config, "_configured_provider_api_key_for_live_probe", lambda *a, **k: "")`. The `delenv` is the minimal change.)

Once CI is green the security posture here is already verified clean, so the re-gate should be quick. Thanks for the solid hardening work — this is one test-determinism fix away.

nesquena-hermes:
## 🔬 Gate re-certification — GREEN ✅ (both round-3 residuals closed: openai env-key + copilot-acp delegate leaks are now skipped; maps centralized so they can't drift) · SECURITY → Nathan holds merge

**Certified head:** `sha:2ba6d8d8` (`fix: cover remaining credentialed model probes`) · **PR:** #5765 [security] · _Hinotoi-agent, stop credentialed live-model redirects_

Re-read the re-pushed `2ba6d8d8` against my two prior residual findings (read-only worktree; diff on `api/config.py`, `api/routes.py`, `tests/test_security_live_models_redirects.py`; cross-referenced the agent probe code). Both gaps are genuinely closed, and the map-drift risk I called out is resolved structurally.

## Residual #1 (openai env-key) — fixed ✔

`OPENAI_API_KEY` (+ custom `OPENAI_BASE_URL`) now delegates no more. Both env maps carry the entry, and the single source lives in config with routes importing it:

```python
# api/config.py:1261 — the canonical map
_AGENT_CREDENTIAL_ENV_NAMES_BY_PROVIDER = {
    ...
    "openai": ("OPENAI_API_KEY",),
    "openai-api": ("OPENAI_API_KEY",),
    ...
}
# api/routes.py:18269 — imports the same table (falls back to an inline copy only if import fails)
from api.config import _AGENT_CREDENTIAL_ENV_NAMES_BY_PROVIDER
```

I confirmed the agent side is exactly the vector I flagged: `hermes_cli/models.py:2367-2369` reads `OPENAI_API_KEY` and a caller-controlled `OPENAI_BASE_URL`, then `fetch_api_models(api_key, base)` → `probe_api_models` follows redirects. With `openai`/`openai-api` in the skip map, `_should_skip_agent_provider_model_ids` (routes.py:18297) and `_read_live_provider_model_ids` (config.py:5670-5688) both return before that agent call.

## Residual #2 (copilot-acp) — fixed ✔

The literal `provider == "copilot"` check is gone; the always-skip set now contains the alias:

```python
# api/config.py:1278 (imported by routes.py:18300 with the same fallback)
_AGENT_ALWAYS_SKIP_PROVIDER_MODEL_IDS_PROVIDERS = frozenset({"copilot", "copilot-acp", "nous"})
```

Matches the agent's combined branch `if normalized in {"copilot", "copilot-acp"}:` at `models.py:2293` → `_fetch_github_models(...)` credentialed. `_resolve_provider_alias` keeps `copilot-acp` distinct (agent alias table maps `github-copilot-acp` → `copilot-acp`), so the frozenset membership is the right discriminator.

## Map-drift addressed structurally ✔

My round-3 note asked for "a single canonical set shared by both `_should_skip_...` implementations so the two maps can't drift." Done — config.py owns both `_AGENT_CREDENTIAL_ENV_NAMES_BY_PROVIDER` and `_AGENT_ALWAYS_SKIP_PROVIDER_MODEL_IDS_PROVIDERS`, and routes.py imports them (the inline literals are import-failure fallbacks only). `gmi`/`stepfun` env coverage is also present in the shared map.

## Tests prove the bypass, not just the shape ✔

The three new cases assert the credentialed agent probe is never reached — they stub `provider_model_ids` to raise, so a skip-gate miss fails the test:

```python
# tests/test_security_live_models_redirects.py:250
def provider_model_ids(provider):
    raise AssertionError("credentialed OpenAI agent live probe should be skipped")
```

Covered: `test_openai_env_key_skips_agent_live_probe` (238), `test_openai_api_env_key_skips_config_agent_live_probe` (272), `test_copilot_acp_skips_agent_live_probe` (289). The prior helper/route no-redirect cases carry over.

## Verdict

**GREEN — the credential-leak-via-redirect class is now closed across the direct probes (custom endpoint, LM Studio, OpenAI-compat/custom routes via the no-redirect helper) and the delegated `provider_model_ids` gate (env-key openai/openai-api + OAuth/token-pool copilot/copilot-acp/nous).** Direction, helper, and coverage are all right. Per the standing rule this is **security + agent-authored → the merge decision is Nathan's**; the certification is decision-ready, not auto-ship. I couldn't execute the suite from here (read-only inspection only), so the 69-passed claim is unverified by me — but every code path and test assertion I read supports it. concept 5/5. Author @Hinotoi-agent.

Hinotoi-agent:
Thanks — fixed the CI blocker in `6b060629`.

Changes:
- Made `tests/test_live_models_ttl_cache.py` clear `OPENAI_API_KEY` and `OPENAI_BASE_URL` in the shared `_patch_live_models_basics()` helper.
- This keeps the TTL cache tests deterministic now that `openai` is correctly treated as a credentialed skip provider when an ambient `OPENAI_API_KEY` exists.
- Production behavior from the security fix is unchanged.

Validation:
```bash
./scripts/test.sh tests/test_live_models_ttl_cache.py
# 5 passed

OPENAI_API_KEY=*** OPENAI_BASE_URL=http://127.0.0.1:1 ./scripts/test.sh tests/test_live_models_ttl_cache.py
# 5 passed

./scripts/test.sh tests/test_security_live_models_redirects.py tests/test_byok_model_dropdown.py tests/test_issue1527_lmstudio_base_url_classification.py tests/test_live_models_ttl_cache.py
# 41 passed

.venv/bin/python -m ruff check tests/test_live_models_ttl_cache.py
# All checks passed!

python3 -m compileall -q tests/test_live_models_ttl_cache.py
git diff --check
```

Note: `python3 scripts/ruff_lint.py --diff origin/master` still reports pre-existing broader branch violations in files outside this mechanical test-only follow-up; the touched file passes direct ruff and compile checks.

Hinotoi-agent:
Follow-up pushed in `8d93db59` for the remaining CI failures.

What changed:
- Updated `tests/test_pr1970_lmstudio_base_url_fallback.py` to monkeypatch `api.config._credentialed_json_get_no_redirect()` instead of `urllib.request.urlopen()`.
- This matches the new hardened LM Studio fallback path from this PR; the tests still prove the hermes_cli-unavailable fallback returns both live models, but now through the no-redirect helper seam.

Validation:
```bash
./scripts/test.sh tests/test_pr1970_lmstudio_base_url_fallback.py tests/test_live_models_ttl_cache.py tests/test_issue1527_lmstudio_base_url_classification.py
# 24 passed

./scripts/test.sh tests/test_security_live_models_redirects.py tests/test_byok_model_dropdown.py tests/test_issue1527_lmstudio_base_url_classification.py tests/test_live_models_ttl_cache.py tests/test_pr1970_lmstudio_base_url_fallback.py
# 55 passed

.venv/bin/python -m ruff check tests/test_live_models_ttl_cache.py tests/test_pr1970_lmstudio_base_url_fallback.py
# All checks passed!

python3 -m compileall -q tests/test_live_models_ttl_cache.py tests/test_pr1970_lmstudio_base_url_fallback.py
git diff --check
```

nesquena-hermes (changes requested):
Thanks for the two follow-ups — the original blocker is genuinely resolved and the security posture is unchanged/green, but CI is still red at the current head for a **related** reason the re-push didn't reach.

## What's fixed ✅
- `tests/test_live_models_ttl_cache.py` — `_patch_live_models_basics()` now `delenv`s `OPENAI_API_KEY`/`OPENAI_BASE_URL` (5 passed, deterministic with or without an ambient key). This was the originally-flagged blocker and it's closed.
- `tests/test_pr1970_lmstudio_base_url_fallback.py` — correctly repointed to `monkeypatch.setattr(config, "_credentialed_json_get_no_redirect", …)` (14 passed).
- `tests/test_security_live_models_redirects.py` — 7 passed; no-redirect guard intact.

## Still red ⛔ — 3 more pre-existing tests broken by the same refactor
CI run against the current head `8d93db59` (not stale) fails 6 shards (3.11/3.12/3.13 × shards 3 & 4). Reproduced deterministically in a sandbox. **Root cause is identical to the fix you already landed:** the security change moved the credentialed custom-provider `/v1/models` probes off `urllib.request.urlopen` onto `_credentialed_json_get_no_redirect(...)`, so every test still monkeypatching `urllib.request.urlopen` against those paths now has a dead mock, makes a real call, and misclassifies/times out. You repointed 2 of the ~5 affected files; these 3 still point at the old symbol:

**1. `tests/test_issue2540_models_endpoint_error.py`** (mocks at lines 69, 88, 111, 136 — 3 tests fail: `_401_surfaces_error` `assert 'network' == 'auth'`, `_5xx_preserves_status` `assert 'network' == 'http'`, `_network_error_uses_short_timeout` `assert [] == [5.0]`)
- Replace each `monkeypatch.setattr(urllib.request, "urlopen", fake_urlopen)` with `monkeypatch.setattr(config, "_credentialed_json_get_no_redirect", fake_get)` (the config-path helper these named-custom-provider tests exercise).
- Change the fake signature `fake_urlopen(req, timeout=10)` → `fake_get(url, headers, *, timeout)`.
- The 401/5xx tests still `raise urllib.error.HTTPError(url, 401|502, …)` — the production `except urllib.error.HTTPError` classifier is unchanged and still yields `kind=='auth'`/`'http'`, so they pass once the mock intercepts the right symbol.
- The short-timeout test asserts `observed_timeouts == [CUSTOM_MODELS_ENDPOINT_TIMEOUT_SECONDS]` (`[5.0]`) — the timeout now arrives as the keyword `timeout=…`, so capture the `timeout` **kwarg** (not a positional arg).

**2. `tests/test_issue2513_custom_provider_remote_models.py`** (mock at lines 25/47 — `test_custom_provider_model_field_does_not_block_remote_catalog`, `'@custom:alpha-proxy:alpha/remote'` missing)
- Repoint `monkeypatch.setattr(urllib.request, "urlopen", fake_urlopen)` → `monkeypatch.setattr(config, "_credentialed_json_get_no_redirect", fake_get)`; return the remote catalog dict `{"data": [...]}` for the alpha/beta proxy URLs.

**3. `tests/test_model_resolver.py::test_custom_endpoint_uses_model_config_api_key_for_model_discovery`** (mock at lines 621-628 — `KeyError: 'auth'`)
- Repoint `monkeypatch.setattr('urllib.request.urlopen', _fake_urlopen)` → `monkeypatch.setattr(config, "_credentialed_json_get_no_redirect", _fake_get)`; in the fake do `captured['auth'] = headers.get('Authorization'); captured['ua'] = headers.get('User-Agent')` and return `{"data": [{"id": "gpt-5.2", "name": "GPT-5.2"}]}`. Asserts stay (`captured['auth'] == 'Bearer «redacted:sk-…»'`, `captured['ua'] == 'OpenAI/Python 1.0'`).

The template is exactly the one you already used in `test_pr1970` (lines 258/323). This is test-mock repointing only — **no production change is needed** (the error classification and the no-redirect helper are correct). Once all three are repointed, the 6 CI shards should go green.

nesquena-hermes (changes requested):
Ran a full adversarial security re-gate on the current head `a2e14dfa97` (enumerated every credential-capable model-probe path + canary redirect repro). The **direct credentialed-helper callsites are verified safe** — the no-redirect helper genuinely blocks the leak there, 200s parse, proxies preserved, CI green. But the enumeration found the fix is **incomplete in one place and over-broad in two**, so it can't ship yet:

**1. [CORE — leak vector still open] `provider_model_ids()` agent delegation — `api/config.py:6889` (+ the Ollama Cloud branch).** Not every credential-capable probe routes through `_credentialed_json_get_no_redirect`: the agent-delegation path (and the special Ollama Cloud branch) is still credential-capable and doesn't go through the no-redirect helper, and the skip-heuristic doesn't cover it. The leak isn't fully closed.
- **Fix:** either delegate to the agent **only for proven-keyless providers**, or harden the agent transport; and route **every** credential-capable probe (incl. Ollama Cloud) through the WebUI no-redirect helper.

**2. [SILENT regression — live discovery lost] `api/config.py:5683` + `api/routes.py:18362`.** When delegation is skipped, the code now returns the **static** catalog instead of doing the live request safely. Verified with `OPENAI_API_KEY` + a custom `OPENAI_BASE_URL`: the base returns `tenant-only-model`, but this head makes no request and silently serves the static OpenAI catalog — tenant-only models vanish.
- **Fix:** when delegation is skipped, resolve the provider key/base URL and do the live request through `_credentialed_json_get_no_redirect`; extend the regression test to assert the live model stays available.

**3. [SILENT regression — keyless redirects broken] `api/config.py:6423` + `:7031`.** Keyless custom-provider and LM Studio probes now reject ALL redirects even though there's no credential to leak. Verified: a keyless base that returns a 302→`redirect-model` now stops at the 302 and loses the model.
- **Fix:** use the no-redirect helper **only when an `Authorization` header is present**; keep ordinary redirect-following transport for keyless probes. Add a keyless-redirect regression.

The security intent is right and the direct callsites are solid — this is about making the coverage complete (close the delegation vector) without breaking the two legit discovery flows (keyed-live + keyless-redirect). Once those three land I'll re-gate the current head. Thanks @Hinotoi-agent — this is close.

nesquena-hermes:
Following up — the changes-requested review above still has three open items blocking this security fix, and there's been no re-push since. Recapping what needs addressing before it can be re-gated:

1. **Leak vector still open** — the agent-delegation path (`api/config.py` ~L6889) and the Ollama Cloud branch are credential-capable but still bypass the no-redirect helper, so the credential leak this PR targets isn't fully closed.
2. **Regression** — when delegation is skipped, the code returns the *static* catalog instead of doing the live request safely, so tenant-only models disappear.
3. **Regression** — keyless probes now reject *all* redirects even when there's no credential to leak, which drops legitimate keyless-redirect models.

Happy to re-gate as soon as those three are addressed. If you'd rather hand it off, let us know and we can pick it up with attribution preserved.

PR #5772, opened 2026-07-08
fix(mcp): scope list config reads to active profile

### LINKED ISSUE #5619: Web UI config discrepancy with multi-profiles

rodboev:
## Thinking Path

- The reload fix for #5619 still left MCP list reads vulnerable to shared config-cache bleed across profiles.
- Moving the list endpoints to active-profile read plumbing fixed that direction, but the active-home path also has to preserve `HERMES_CONFIG_PATH` because MCP writes already use that override.
- The helper now treats the current active home as the ambient config source, so list reads and writes agree on the same config file while detached non-active profile reads still go straight to that profile.

## What Changed

- `api/routes.py`: read MCP server and tool configuration through the active profile home.
- `api/config.py`: keep active-home profile reads on `get_config()` so `HERMES_CONFIG_PATH` remains the read authority for override-config deployments.
- `tests/test_issue538_mcp_management.py`: cover the server-list active-profile read path and the external override read/write parity path.
- `tests/test_issue697_mcp_tool_inventory.py`: cover the tool-list active-profile read path.
- `tests/test_issue3294_profile_toolset_scoping.py`: cover active-home override handling without weakening direct reads for non-active profile homes.

## Why It Matters

Switching profiles shouldn't make the MCP UI inherit servers from whichever profile last warmed the process config cache. Override-config deployments also need a server added through MCP settings to appear in the list immediately, because both operations resolve the same config path.

## Verification

Focused tests cover MCP server management, MCP tool inventory, and active-profile config helper behavior; CI runs the full matrix.

## Upstream

Closes #5619.

## Model Used

GPT-5.5 via Codex CLI

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -703,6 +703,13 @@ def get_config_for_profile_home(profile_home: "Path | str | None") -> dict:
         target = Path(profile_home).expanduser()
     except Exception:
         return get_config()
+    try:
+        from api.profiles import get_active_hermes_home
+
+        if Path(get_active_hermes_home()).expanduser() == target:
+            return get_config()
+    except Exception:
+        pass
     # If the ambient resolver already points at this profile home, defer to
     # get_config() so in-memory overrides (monkeypatched cfg) are honored. This
     # MUST run before the nonexistent-home guard below: a matching ambient home
--- a/api/routes.py
+++ b/api/routes.py
@@ -2822,6 +2822,7 @@ from api.config import (
     _load_yaml_config_file,
     _save_yaml_config_file,
     reload_config,
+    get_config_for_profile_home,
     _cfg_lock,
     PENDING_BG_TASK_COMPLETIONS,
 )
@@ -24181,7 +24182,7 @@ def _mcp_tools_from_registry(server_summaries):
 
 def _handle_mcp_tools_list(handler):
     """List known MCP tools from already-available runtime inventory only."""
-    cfg = get_config()
+    cfg = get_config_for_profile_home(get_active_hermes_home())
     servers = cfg.get("mcp_servers", {})
     if not isinstance(servers, dict):
         servers = {}
@@ -24695,7 +24696,7 @@ def _handle_notes_item(handler, parsed):
 
 def _handle_mcp_servers_list(handler):
     """List configured MCP servers with safe, read-only runtime visibility."""
-    cfg = get_config()
+    cfg = get_config_for_profile_home(get_active_hermes_home())
     servers = cfg.get("mcp_servers", {})
     if not isinstance(servers, dict):
         servers = {}
--- a/tests/test_issue3294_profile_toolset_scoping.py
+++ b/tests/test_issue3294_profile_toolset_scoping.py
@@ -108,6 +108,31 @@ def test_matching_home_defers_to_get_config(_two_profiles):
     ]
 
 
+def test_active_home_with_external_config_override_defers_to_get_config(_two_profiles, monkeypatch, tmp_path):
+    """The active home still honors HERMES_CONFIG_PATH when the override lives
+    outside that home."""
+    cfg, default_home, profile_b_home = _two_profiles
+    from api import profiles
+
+    active_home = tmp_path / "active-home"
+    override_path = tmp_path / "override-dir" / "config.yaml"
+    active_home.mkdir()
+    _write_cfg(active_home, ["wrong-active-home-file"])
+    override_path.parent.mkdir()
+    import yaml
+
+    override_path.write_text(
+        yaml.safe_dump({"platform_toolsets": {"cli": ["override-config"]}}, sort_keys=False),
+        encoding="utf-8",
+    )
+    monkeypatch.setenv("HERMES_CONFIG_PATH", str(override_path))
+    monkeypatch.setattr(profiles, "get_active_hermes_home", lambda: active_home)
+    cfg.reload_config()
+
+    same = cfg.get_config_for_profile_home(active_home)
+    assert same.get("platform_toolsets", {}).get("cli") == ["override-config"]
+
+
 def test_matching_ambient_home_that_does_not_exist_still_defers_to_get_config(_two_profiles, monkeypatch, tmp_path):
     """Regression (#4516 gate): the nonexistent-home guard must run AFTER the
     ambient-resolver short-circuit. A home that matches the ambient config path
--- a/tests/test_issue538_mcp_management.py
+++ b/tests/test_issue538_mcp_management.py
@@ -42,17 +42,33 @@ SAMPLE_MCP = {
 class TestMcpList:
     """GET /api/mcp/servers — list with masked secrets."""
 
-    @patch('api.routes.get_config')
-    def test_returns_servers_list(self, mock_cfg):
+    @patch('api.routes.get_config_for_profile_home')
+    @patch('api.routes.get_active_hermes_home')
+    def test_returns_servers_list(self, mock_home, mock_cfg):
+        mock_home.return_value = sentinel_home = object()
         mock_cfg.return_value = {'mcp_servers': SAMPLE_MCP}
         h = _make_handler()
         _handle_mcp_servers_list(h)
         assert h.send_response.called
         status = h.send_response.call_args[0][0]
         assert status == 200
+        mock_cfg.assert_called_once_with(sentinel_home)
 
-    @patch('api.routes.get_config')
-    def test_empty_config(self, mock_cfg):
+    @patch('api.routes.get_config_for_profile_home')
+    @patch('api.routes.get_active_hermes_home')
+    def test_reads_active_profile_home_for_servers(self, mock_home, mock_cfg):
+        mock_home.return_value = sentinel_home = object()
+        mock_cfg.return_value = {'mcp_servers': {'active': SAMPLE_MCP['searxng']}}
+        h = _make_handler()
+        _handle_mcp_servers_list(h)
+        payload = _json_payload(h)
+        assert [srv['name'] for srv in payload['servers']] == ['active']
+        mock_cfg.assert_called_once_with(sentinel_home)
+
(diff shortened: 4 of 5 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (the profile-scoping switch drops `HERMES_CONFIG_PATH` override handling → MCP list reads from `$HERMES_HOME/config.yaml` while add/remove still writes to the override = read/write mismatch, empty list for override-config deployments — 2 SILENT)

**Certified head:** `sha:<rebased>` (rebased onto current master, `git apply` clean) · **PR:** #5772 · _rodboev, fix(mcp): scope list config reads to active profile_
**Verdict:** The intent (profile-scope MCP list reads) is right, but Codex found the switch from `get_config()` → `get_config_for_profile_home(get_active_hermes_home())` LOST the documented `HERMES_CONFIG_PATH` override handling — so for deployments with a config file outside `HERMES_HOME`, the MCP list reads empty/wrong while add/remove/toggle still write to the override (read/write mismatch). Suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5772`) — Codex + full suite (backend profile-isolation)
| Gate | Result |
|------|--------|
| Codex (all sites scoped / active-home / HERMES_CONFIG_PATH / consistency) | **SHIP ONLY WITH FIXES** — 2 SILENT |
| Full pytest suite | ✅ **12324 passed, 0 failed** (+ 42 MCP tests) |

### Findings
**⛔ SILENT ×2 (routes.py:24185 `/api/mcp/tools`, 24699 `/api/mcp/servers`) — `HERMES_CONFIG_PATH` override dropped → read/write mismatch:** the new `get_config_for_profile_home(get_active_hermes_home())` reads `$HERMES_HOME/config.yaml`, IGNORING a documented `HERMES_CONFIG_PATH` override that lives elsewhere. Codex verified: with a temp override where `config.get_config()` saw `override_srv`, `_handle_mcp_servers_list()` returned `servers: []`. Meanwhile add/remove/toggle still WRITE through `_get_config_path()` to the override — so on an override-config deployment you add an MCP server, it's written to the override, but the list reads from `$HERMES_HOME` and shows nothing (or the wrong servers). **Root:** `get_config()` was ALREADY profile-scoped (server.py per-request TLS) AND honored `HERMES_CONFIG_PATH`; switching to the home-based read gained nothing over the TLS scoping but lost the override. **Fix (Codex):** either (a) make `get_config_for_profile_home()` return `get_config()` when `Path(profile_home)==get_active_hermes_home()` (preserve ambient config + override), or (b) keep these GET handlers on `get_config()` (already profile-scoped by server.py TLS) — and add a regression test with `HERMES_CONFIG_PATH` outside `HERMES_HOME`.

**RED — gate-fail/changes-requested (2 SILENT, override-config regression + read/write mismatch): the switch to `get_config_for_profile_home(get_active_hermes_home())` drops the `HERMES_CONFIG_PATH` override that `get_config()` honored, so MCP list reads diverge from where add/remove writes. First confirm the ORIGINAL cross-profile leak is real (given `get_config()` is already TLS-profile-scoped by server.py) — if it is, fix `get_config_for_profile_home` to fall through to `get_config()` when the home is the active home (preserving the override); if `get_config()` was already correctly scoped, the change may be unnecessary. Add a regression test with `HERMES_CONFIG_PATH` outside `HERMES_HOME`, and audit read/write parity (list-read + add/remove/toggle-write must resolve the SAME config path).** concept 4/5 (right goal; the home-based read lost the override that the TLS-scoped get_config had). Author @rodboev. crit=3, profile-isolation.

---
_Gate-certifier layer (warm-up → **gate** → release). MCP list-read switched get_config()->get_config_for_profile_home(get_active_hermes_home()) to profile-scope, but Codex: this DROPS the HERMES_CONFIG_PATH override (reads $HERMES_HOME/config.yaml) while add/remove still WRITE via _get_config_path() to the override = read/write MISMATCH, empty list for override-config deployments (routes.py:24185 /tools, 24699 /servers). get_config() was already TLS-profile-scoped by server.py + honored the override. Fix: get_config_for_profile_home fall through to get_config() when home==active-home (preserve override), or keep GETs on get_config(); + regression test HERMES_CONFIG_PATH outside HERMES_HOME + read/write path-parity audit. Suite green 12324 + 42 MCP tests. Cert valid for the rebased head._

rodboev:
Thanks for catching this. The route-level switch was too narrow: it made the MCP list endpoints explicit about the active home, but it let active-home reads bypass the `HERMES_CONFIG_PATH` path that MCP writes still use.

I fixed this at the config helper instead of backing the handlers out:

1. `api.config.get_config_for_profile_home()` now returns through `get_config()` when the requested `profile_home` is the current active Hermes home. That keeps the active-home read on the same `_get_config_path()` authority as add/delete/toggle, including `HERMES_CONFIG_PATH` outside `HERMES_HOME`.
2. `_handle_mcp_tools_list()` and `_handle_mcp_servers_list()` stay on `get_config_for_profile_home(get_active_hermes_home())`. With the helper fixed, both endpoints keep the explicit active-profile read and regain override-path parity.
3. Divergent-profile reads are unchanged for the detached worker case that originally needed `get_config_for_profile_home(profile_home)`.
4. I added regression coverage with an override config outside the active home, plus a server-list read/write parity check so the list reads the same config file that MCP writes update.

That closes the class the gate found: active-home MCP reads and MCP writes resolve the same config source, while non-active profile reads still go straight to that profile’s own `config.yaml`.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ · CONVERGED (round 2 — override preserved + read/write parity restored, cross-profile still scoped) · profile-isolation → Nathan

**Certified head:** `sha:95858bf1` (clean rebase, branch `gate-rebase/5772-mcp-profile-scope-override`) · **PR:** #5772 · _rodboev, fix(mcp): scope list config reads to active profile_
**Verdict:** GREEN — CONVERGED. My round-1 SILENT (the profile-home read dropped the `HERMES_CONFIG_PATH` override → read/write mismatch) is fixed centrally: `get_config_for_profile_home` now falls through to `get_config()` when the requested home IS the active home (preserving the override), while a different profile home still reads that home's config. Codex SAFE, full suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5772b`) — Codex + full suite (backend profile-isolation)
| Gate | Result |
|------|--------|
| Codex (override preserved / read-write parity / all-callers / path-equality) | **SAFE TO SHIP** — 0 findings |
| Full pytest suite | ✅ **12326 passed, 0 failed** (+ 43 MCP+config tests) |

### Findings
**✅ Round-1 SILENT FIXED (centrally):** `get_config_for_profile_home(profile_home)` now checks `if Path(get_active_hermes_home()).expanduser() == target: return get_config()` — so when the requested home IS the active home, it returns the ambient `get_config()` (which honors `HERMES_CONFIG_PATH`), preserving the override; only a DIFFERENT profile home reads that home's `config.yaml`. So the MCP list read now matches what add/remove/toggle write (read/write PARITY restored — the override-config empty-list bug closed), while the original #5772 cross-profile scoping intent is preserved (a different profile still can't leak). Codex verified: override preserved for the active profile, cross-profile still scoped, the central change is safe for ALL callers of `get_config_for_profile_home` (the active-home fallthrough is the correct default for every caller), and path-equality is robust (expanduser both sides). Full suite green (43 MCP+config tests, +1 override-preservation test).

### Recommendation to Nathan
**GREEN — merge from branch `gate-rebase/5772-mcp-profile-scope-override` (`sha:95858bf1`).** A clean convergence — the fix moved to the central `get_config_for_profile_home` helper so the `HERMES_CONFIG_PATH` override is preserved for the active profile (read/write parity) while cross-profile MCP reads stay scoped. Codex SAFE + full suite green. Backend profile-isolation → your merge nod. concept 4/5. Author @rodboev. crit=4, profile-isolation + read/write parity.

---
_Gate-certifier layer (warm-up → **gate** → release). r1 SILENT (profile-home read dropped HERMES_CONFIG_PATH override → read/write mismatch, empty MCP list for override deployments) FIXED centrally: get_config_for_profile_home returns get_config() when Path(get_active_hermes_home())==target (preserves override for active profile), else reads that home's config.yaml (cross-profile scoped). Read/write parity restored + cross-profile intent preserved. Codex SAFE (override preserved, all-callers safe, path-equality robust) + full suite green (0 failed, 12326) + 43 MCP+config tests. Profile-isolation → Nathan. Cert valid for sha:95858bf1._

nesquena-hermes:
Shipped in **v0.51.924** — thanks @rodboev. 🎉

MCP server/tool list reads are now scoped to the active profile (via `get_config_for_profile_home(get_active_hermes_home())`) instead of the ambient default, and the companion early-return keeps `HERMES_CONFIG_PATH` + in-memory (#5619) overrides honored so reads match where the write handlers save.

Gate: Codex SAFE + Fable SAFE (live end-to-end probe confirmed no cross-profile leak in either direction, #5619 preserved, read/write parity restored) + full suite 12239/0. Deployed and verified live. Also closes the read-side of #5619.

PR #5773, opened 2026-07-08
fix(chat): keep live reasoning on one renderer path

### LINKED ISSUE #5720: Transparent Stream: thinking/reasoning content overwrites first line, appears all at once after completion
## Description

In **Transparent Stream** mode (`chat_activity_display_mode: transparent_stream`) with `show_thinking: true`, thinking/reasoning content from a reasoning model (e.g. DeepSeek V4 Flash sending `reasoning_content` via SSE) exhibits three visual bugs during streaming:

1. **Thinking content appears on the first line of the reply** — The thinking card is rendered as the first row in the anchor activity scene, positioned at the top of the assistant turn before any prose output, making it look like the "first line" of the response.

2. **Thinking overwrites itself in-place** — Each incoming `reasoning` SSE event triggers two competing render paths that both tear down and rebuild the thinking DOM node:
   - `_upsertAnchorReasoning()` → `_renderAnchorLiveScene()` → full scene rebuild, which removes all `[data-live-thinking="1"]` rows
   - `_updateLiveThinkingCard()` → `appendThinking()` → detects anchor scene is owner → calls `_renderLiveAnchorActivitySceneForStream()` → another full rebuild

   This double-render causes the thinking row to be deleted and recreated on every reasoning update, producing visible flicker and the "overwrite" effect.

3. **Thinking content appears all at once after completion** — During streaming, the thinking row is constantly torn down and rebuilt, so incremental updates are mostly invisible to the user. After the turn settles (STREAM_DONE), the settled scene path (`_renderSettledAnchorSceneTransparentForMessage`) rebuilds all rows from the persisted anchor scene, at which point the full accumulated reasoning text appears suddenly.

## Root Cause

In `static/ui.js`, the `appendThinking()` function (line 16736-16738) has this guard:

```javascript
if(!allowPendingPlaceholder && isLiveAnchorActivitySceneOwner(S.activeStreamId)){
    _renderLiveAnchorActivitySceneForStream(S.activeStreamId, S.session.session_id, {mode:'compact_worklog'});
    return;
}
```

When the anchor scene is active (which it always is during transparent stream), `appendThinking` redirects to `_renderLiveAnchorActivitySceneForStream` and returns early — **the transparent-stream-specific thinking rendering code at lines 16775-16809 never runs**.

Meanwhile, in `_renderLiveAnchorActivitySceneTransparent` (line 11079), thinking rows are explicitly removed:
```javascript
blocks.querySelectorAll(
  '.agent-activity-thinking[data-live-thinking="1"]'
).forEach(el => el.remove());
```

So every reasoning event causes: (A) `_upsertAnchorReasoning` → anchor scene rebuild → thinking rows removed and recreated, then (B) `_updateLiveThinkingCard` → `appendThinking` → another anchor scene rebuild.

The anchor scene data model DOES support thinking rows (via `role: "thinking"` in scene rows, handled at line 11039-11048), but the double-render and row removal mean the thinking content never stabilizes during live streaming.

## How to Reproduce

1. Set `chat_activity_display_mode: transparent_stream` and `show_thinking: true`
2. Use a model that sends `reasoning_content` via SSE (e.g. DeepSeek V4 Flash)
3. Send a message that triggers a long reasoning chain
4. Observe: thinking flashes/rewrites at the top, and does not stabilize until the turn completes

## Environment

- WebUI Version: v0.51.878 → v0.51.895 (both affected)
- Browser: any modern browser
- Display mode: transparent_stream
- Model: deepseek-v4-flash-free (opencode provider, sends reasoning_content)

## Related Issues

- #5367 Transparent Stream flicker (partially related - same root cause of DOM rebuild)
- #5400 Transparent Stream reconciliation by identity (fixed, but thinking rows still get removed)
- #5456 Transparent Stream smoother during streaming (reduced no-op work but didn't fix the double-render)
- #5547 Transparent Stream tool status fix (unrelated)
- #5558 Transparent Stream final reply stays visible (unrelated)

rodboev:
## Thinking Path

- The reporter's retest identified the live `reasoning` SSE handler still driving both anchor reasoning and the legacy thinking card.
- `_upsertAnchorReasoning` already returns whether the anchor path handled the update, so the handler can use that result as the renderer discriminator.
- Inline parsed thinking stays on the legacy thinking-card path because it is a different input path from provider `reasoning` SSE deltas.

## What Changed

- `static/messages.js`: route provider `reasoning` SSE updates through the anchor reasoning path first and call the live thinking card only as fallback.
- `tests/test_ui_tool_call_cleanup.py`: cover the single-renderer routing contract and preserve inline parsed thinking behavior.

## Why It Matters

Live reasoning should update one stable thinking row during Transparent Stream. Removing the redundant second live renderer path avoids extra reconciliation while keeping non-anchor live thinking intact.

## Verification

Focused tests cover the reasoning SSE route, inline parsed thinking preservation, and existing live thinking segment scoping; CI runs the full matrix.

## Upstream

Closes #5720.

## Model Used

GPT-5.5

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -5024,8 +5024,10 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
       if(d.text&&S.session&&S.session.session_id===activeSid) _completeAutomaticCompressionOnLiveProgress(activeSid);
       syncInflightAssistantMessage();
       if(text&&S.session&&S.session.session_id===activeSid){
-        _upsertAnchorReasoning(_liveThinkingText());
-        _updateLiveThinkingCard(_liveThinkingText());
+        const liveThinkingText=_liveThinkingText();
+        if(!_upsertAnchorReasoning(liveThinkingText)){
+          _updateLiveThinkingCard(liveThinkingText);
+        }
       }
     });
 
--- a/tests/test_regressions.py
+++ b/tests/test_regressions.py
@@ -931,8 +931,12 @@ def test_messages_js_supports_live_reasoning_and_tool_completion(cleanup_test_se
         "messages.js must listen for live reasoning SSE events"
     assert "liveReasoningText += text" in src, \
         "live reasoning SSE events must update the active Worklog Thinking Card text"
-    assert "_updateLiveThinkingCard(_liveThinkingText())" in src, \
-        "live reasoning SSE events must refresh the current segment's Worklog Thinking Card"
+    assert "const liveThinkingText=_liveThinkingText();" in src, \
+        "live reasoning SSE events must compute the current segment's Worklog Thinking Card text once"
+    assert "if(!_upsertAnchorReasoning(liveThinkingText))" in src, \
+        "live reasoning SSE events must prefer the anchor renderer before falling back"
+    assert "_updateLiveThinkingCard(liveThinkingText)" in src, \
+        "live reasoning SSE events must keep the current segment's Worklog Thinking Card as fallback"
     assert "source.addEventListener('tool_complete'" in src or 'source.addEventListener("tool_complete"' in src, \
         "messages.js must listen for live tool completion SSE events"
     assert "function _parseStreamState()" in src, \
--- a/tests/test_stable_assistant_turn_anchor_registry.py
+++ b/tests/test_stable_assistant_turn_anchor_registry.py
@@ -1306,7 +1306,9 @@ def test_slice6_live_shadow_feed_wires_anchor_scene_for_visible_order_handoff():
     assert "_upsertAnchorProcessProse(displayText" in src
     reasoning_body = _event_listener_body(src, "reasoning")
     assert "_applyToAnchor" not in reasoning_body
-    assert "_upsertAnchorReasoning(_liveThinkingText())" in reasoning_body
+    assert "const liveThinkingText=_liveThinkingText();" in reasoning_body
+    assert "if(!_upsertAnchorReasoning(liveThinkingText))" in reasoning_body
+    assert "_updateLiveThinkingCard(liveThinkingText)" in reasoning_body
     assert "function _flushReasoningToAnchor()" in src
     assert "_upsertAnchorReasoning(reasoningText" in src
     assert "`live-reasoning:${streamId}:final`" in src
--- a/tests/test_ui_tool_call_cleanup.py
+++ b/tests/test_ui_tool_call_cleanup.py
@@ -670,6 +670,40 @@ class TestToolCallGroupingStatic:
             "Tool starts must not split consecutive tools into one-tool Activity rows."
         )
 
+    def test_reasoning_stream_uses_one_live_renderer_path(self):
+        reasoning_match = re.search(
+            r"source\.addEventListener\('reasoning',e=>\{(.*?)\n\s*\}\);",
+            MESSAGES_JS,
+            re.S,
+        )
+        assert reasoning_match, "reasoning listener not found"
+        reasoning_fn = reasoning_match.group(1)
+        render_live_thinking_fn = _function_body(MESSAGES_JS, "_renderLiveThinking")
+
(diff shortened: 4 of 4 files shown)


nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (full gate: Codex + Fable-UX + suite) · ⏸️ visible streaming reasoning → Nathan

**Certified head:** `sha:0872800a` (clean rebase, branch `gate-rebase/5773-live-reasoning-one-path`) · **PR:** #5773 · _rodboev, fix(chat): keep live reasoning on one renderer path_
**Verdict:** GREEN. Consolidates live-reasoning rendering to a single path (anchor-preferred, worklog thinking-card fallback) so it no longer double-renders across two paths. Codex SAFE, Fable-UX SHIP-UX, full suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5773`) — full gate: Codex + Fable + suite
| Gate | Result |
|------|--------|
| Codex (single-path / return-contract / no-lost-reasoning / showThinking-off) | **SAFE TO SHIP** — 0 findings |
| **Fable-UX** | **SHIP-UX** — reasoning renders once, no flicker/dup |
| Full pytest suite | ✅ **12324 passed, 0 failed** (+ 58 anchor/thinking tests) |

### Findings
**✅ Clean single-path consolidation (+47/-5):** live-reasoning SSE previously ran BOTH `_upsertAnchorReasoning(text)` AND `_updateLiveThinkingCard(text)` unconditionally (double render across two paths → flicker/dup risk). Now: `const liveThinkingText=_liveThinkingText(); if(!_upsertAnchorReasoning(liveThinkingText)){ _updateLiveThinkingCard(liveThinkingText); }` — prefer the anchor renderer, fall back to the worklog thinking-card ONLY when the anchor path returns falsy. `_upsertAnchorReasoning` returns `null` when it can't render (`!clean || !_anchorRegistry || _showThinking===false`) and a truthy result (replaced/applied anchor) when it does. Codex verified: exactly one path renders (no double-render), NO lost reasoning in the fallback (worklog card still runs when no anchor registry), `_showThinking===false` correctly no-ops BOTH (the falsy return does NOT cause the fallback to render disabled thinking — the fallback itself respects showThinking), the return-contract is right on all branches, final (non-live) reasoning unaffected, and it coexists with the #5637 scroll/anchor work + the activity-display modes (hide_all_activity still hides reasoning). Fable confirmed single-render, no flicker/dup, fallback still shows reasoning, disabled-thinking stays hidden. Full suite green.

### Recommendation to Nathan
**GREEN — merge from branch `gate-rebase/5773-live-reasoning-one-path` (`sha:0872800a`) — quick glance (live reasoning rendering once mid-stream, no duplicate thinking text).** A clean render-path consolidation on the crown-jewel streaming-reasoning surface, with a correct anchor-preferred/worklog-fallback contract and no lost reasoning. Codex SAFE + Fable SHIP-UX + full suite green. concept 4/5. Author @rodboev. crit=3, visible streaming reasoning.

---
_Gate-certifier layer (warm-up → **gate** → release). Live reasoning consolidated to ONE path: was double-render (_upsertAnchorReasoning + _updateLiveThinkingCard unconditional), now if(!_upsertAnchorReasoning(text)){_updateLiveThinkingCard(text)} — anchor-preferred, worklog fallback only when anchor returns falsy (null on !clean/!_anchorRegistry/_showThinking===false). No double-render, no lost reasoning in fallback, showThinking===false no-ops both, coexists #5637+activity-modes. Codex SAFE + Fable-UX SHIP-UX + full suite green (0 failed, 12324) + 58 anchor/thinking tests. Visible → Nathan. Cert valid for sha:0872800a._

nesquena-hermes:
Shipped in **v0.51.938** — thanks @rodboev. 🎉 Live reasoning now routes through a single renderer so it can't briefly double-render mid-stream. Screenshot-approved + suite 12,291/0. Deployed.

PR #5783, opened 2026-07-08
fix(bg): serialize daemon-thread start under a lifecycle lock

ai-ag2026:
## Summary
Serialize the check-then-start of the module's daemon threads under a dedicated lifecycle lock.

## Why
`start_drain_thread` and `start_session_channel_reaper` checked `is_alive()` and started the thread without a lock. Two concurrent callers (two init paths / a restart race) can both observe `is_alive() == False` and each spawn a daemon thread; the loser's thread is never referenced by the module global, runs forever un-joinable, and two threads then split the same `completion_queue`.

## What
A dedicated `_THREAD_LIFECYCLE_LOCK` wraps the check-and-start of both functions (not the purpose-bound `SESSION_CHANNELS_LOCK`/`_EMIT_COALESCE_LOCK`, to keep it narrow).

## Validation
- `python3 -m py_compile` green; module imports cleanly; logic exercised with python3.11 (N concurrent starts → exactly one thread). New test: `tests/test_bg_thread_start_lock.py`. Full suite via `./scripts/test.sh` in CI.

## Notes
Part of the #4633 uptime-stability cluster.

### DIFF
--- a/api/background_process.py
+++ b/api/background_process.py
@@ -57,6 +57,14 @@ _REAPER_THREAD: Optional[threading.Thread] = None
 _REAPER_STOP = threading.Event()
 _REAPER_INTERVAL_SECS = 60.0
 
+# Serializes the check-then-start of the module's daemon threads
+# (``start_drain_thread`` / ``start_session_channel_reaper``). Without it two
+# concurrent callers can both observe ``is_alive() == False`` and each spawn a
+# thread; the loser's thread is never referenced by the module global and runs
+# forever, un-joinable. A dedicated lock (not the purpose-bound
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.927** — thanks @ai-ag2026. 🎉

Both `start_session_channel_reaper` and `start_drain_thread` now perform the `is_alive()` check, thread creation, and global assignment atomically under a dedicated `_THREAD_LIFECYCLE_LOCK`, so a concurrent double-call can't spawn a second un-joinable daemon thread. Stop/join stay outside the lock (no deadlock).

Gate: Codex SAFE + Fable SAFE (both confirmed the check-then-act race is fully closed for both functions with no deadlock/lock-order inversion) + suite 12253/0. Deployed and verified live.

Issue #2097, opened 2026-05-11
Follow-up: turn-journal hardening — double-terminal detection + multi-process append safety


nesquena-hermes:
## Background

Two cleanup items from the v0.51.46 / Release V Opus advisor review. Filing as a single batch so they can be picked up by a contributor (or rolled into the next contributor PR that touches these areas).

## 1. Turn-journal `completed` × `interrupted` lifecycle double-write

PR #2062 places `completed` inside an outer `try` block and `interrupted` inside the outer `except`. **Mutually exclusive in the normal case.** Edge case: if anything between the `completed` event write and the start of the outer `except` raises an unhandled exception (e.g. a logger crash, an OOM in the middle of state cleanup), both events fire for the same turn.

Currently `derive_turn_journal_states` resolves the collision by latest `created_at` — data is non-crashing but lossy.

### Fix shape

Add a `terminal: bool` field on `completed` and `interrupted` events. Have the audit detect double-terminal turns explicitly (e.g. report `turn_journal_double_terminal` for the audit consumer) rather than silently collapsing.

~10-15 LOC + 1 regression test.

## 2. Multi-process append safety for long messages

`api/turn_journal.py` uses `O_CREAT|O_APPEND|O_WRONLY` + single `write()` per event. POSIX guarantees atomic appends only up to `PIPE_BUF` (4096 bytes on Linux). The `submitted` event line contains the full user message, so a >4KB paste crosses the atomicity boundary.

For the current single-process deployment this is fine. If a future deployment runs multiple webui workers on the same state dir (e.g. behind a load balancer with sticky sessions, or a multi-worker uvicorn), large submitted events from different workers could interleave at the byte level — corrupting the JSONL.

### Fix shape

Two options:

(a) **Per-process journal files.** Append `pid-{os.getpid()}` to the journal filename, and `iter_turn_journal_session_ids` globs across the variants. Simple, zero coordination needed.

(b) **flock-based writer.** `fcntl.flock(LOCK_EX)` around the write block. Slightly slower (lock contention), but keeps the on-disk filename stable.

(a) is simpler. (b) is more standard.

Either is ~20-30 LOC + a multi-process regression test (spawn N writer subprocesses, hammer the same session, verify no interleaved lines).

## Why both items together

Both are forward-looking — neither is a current-deployment bug. Filing as a single hardening ticket for whichever contributor wants to harden the turn-journal subsystem for multi-process or robustness scenarios.

## References

[...]

PR #5784, opened 2026-07-08
fix(run-journal): evict writer-lock cache on journal delete

ai-ag2026:
## Summary
Evict a session's run-journal writer locks when its journal directory is deleted.

## Why
`_WRITER_LOCKS` gains a `(dir, file, pid) -> Lock` entry per run-journal file written and never removes it — `delete_run_journal` only removes files. Over uptime the dict grows unbounded (part of the #4633 slow-climb pattern; related to the run-journal hardening tracked in #2097).

## What
In `delete_run_journal`, after removing the journal directory, drop all `_WRITER_LOCKS` keys whose parent-dir component matches the removed directory (pid-independent) under `_WRITER_LOCKS_GUARD`.

## Validation
- `python3 -m py_compile` green; logic exercised with python3.11 (write → lock created → delete → key gone). New test: `tests/test_run_journal_writer_lock_evict.py`. Full suite via `./scripts/test.sh` in CI.

## Notes
Refs #4633, #2097.

### DIFF
--- a/api/run_journal.py
+++ b/api/run_journal.py
@@ -285,7 +285,22 @@ def delete_run_journal(session_id: str, *, session_dir: Path | None = None) -> b
     if not session_journal_dir.exists():
         return False
     shutil.rmtree(session_journal_dir, ignore_errors=True)
-    return not session_journal_dir.exists()
+    removed = not session_journal_dir.exists()
+    # Evict any writer locks the removed runs left behind. `_lock_for` keys are
+    # ``(str(path.parent), path.name, pid)`` and every run file for this session
+    # lives directly under ``session_journal_dir``, so drop all keys whose parent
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **v0.51.928** — thanks @ai-ag2026. 🎉

`delete_run_journal` now evicts the deleted session's `_WRITER_LOCKS` entries (dir-scoped, under the guard) after a confirmed `rmtree` removal — so the lock cache no longer grows unbounded and a recreated journal gets a fresh lock. The `if removed:` guard means a Windows/permission transient that leaves files (locks still live) never has its lock dropped from under a writer.

Gate: Codex SAFE + Fable SAFE (runtime-proved the guard with a monkeypatched rmtree no-op; confirmed eviction never acquires the per-journal lock and a recreated journal gets a fresh lock) + suite 12256/0. Deployed and verified live.

Issue #2476, opened 2026-05-17
macOS WebUI high CPU/battery during active Codex run with 300s stale timeouts


stefanpieter:
## Bug Description

On macOS, Hermes WebUI can remain busy and consume noticeable CPU/battery during or after a long active run. In the same run, the underlying Hermes Agent call to the OpenAI Codex provider repeatedly hit the 300s stale-call cutoff:

`Non-streaming API call timed out after 300s with no response (threshold: 300s)`

This seems to be a combined WebUI/runtime observability issue:

1. The WebUI background service can show sustained CPU use while an active run is open.
2. The active run can keep retrying long provider calls for many minutes.
3. The timeout/error surfaced as "non-streaming" even though the logs show a Codex streaming request path.
4. The stale-call detector logged `context=~0 tokens`, while the final failure logged a very large context estimate (~98k tokens), suggesting the timeout estimator may not be reading the Codex Responses payload correctly.

## Observed Behavior

Hermes WebUI was running as a macOS launchd service on `127.0.0.1:8787`.

CPU samples for the WebUI server process while the active run was open:

- 7.7%
- 20.2%
- 14.6%

Other Hermes services were mostly idle at the same time:

- regular Hermes gateway: ~0.0% CPU
- other profile gateways: ~0.0% CPU
- dashboard: ~0.1% CPU
- scheduled cron job: not running at the time

macOS power assertions did not show Hermes/Python/WebUI preventing sleep, so this appears to be CPU/energy usage rather than a sleep-prevention assertion.

The WebUI `/health` endpoint showed an active run, for example:

```json
{
  "status": "ok",
  "sessions": 2,
  "active_streams": 1,
  "active_runs": 1,
  "runs": [
    {
      "phase": "running",
      "model": "gpt-5.5",
      "provider": "openai-codex",
      "ephemeral": false,
      "age_seconds": 80.5
    }
  ],
  "accept_loop": {
    "requests_total": 2650
  },
  "oldest_run_age_seconds": 80.5
}
```

Agent logs for the same WebUI session showed repeated stale-call timeouts:

```text
Non-streaming API call stale for 300s (threshold 300s). model=gpt-5.5 context=~0 tokens. Killing connection.
API call failed (attempt 1/3) error_type=TimeoutError provider=openai-codex model=gpt-5.5 summary=Non-streaming API call timed out after 300s with no response (threshold: 300s)
Retrying API call ...
API call failed (attempt 2/3) ...
[...]

PR #5787, opened 2026-07-08
fix(streaming): pair metering begin_session with end_session teardown

ai-ag2026:
## Summary
Pair `meter().begin_session()` with an idempotent `end_session()` teardown so metering sessions can't leak.

## Why
`begin_session(stream_id)` had no matching `end_session()` anywhere. `get_stats()`/`get_interval()` only prune sessions with `first_token_ts > 0`, so a **zero-token turn** (pre-flight cancel, setup raise) is never reclaimed — one orphan per such turn in `meter()._sessions` for the process lifetime, and it inflates the SSE metering event's `active = len(_sessions)`. Part of the #4633 uptime-leak pattern. The metering ticker thread was also only stopped in an inner `finally`, so early exits leaked it.

## What
Move `begin_session()` + the ticker `.start()` to the first statements **inside** the outer `try`, and add a paired `end_session(stream_id, 0)` (idempotent — just pops the entry; metering payload unchanged) + `_metering_stop.set()` to the outer `finally` (the master cleanup that already pops `STREAMS`/`CANCEL_FLAGS`). Deferring `.start()` until after `put()` is defined also closes a latent start-before-`put` ordering window.

## Validation
- `python3 -m py_compile` green; logic exercised with python3.11 (begin → cancel-before-token → `len(_sessions) == 0`; idempotency; never-begun safe; payload keys unchanged). New test `tests/test_metering_session_lifecycle.py`. Full suite via `./scripts/test.sh` in CI.

## Contract Routing
Family: session-SSE streaming teardown. Evidence: `end_session` is idempotent and the metering payload shape is unchanged — behavior-preserving cleanup, not a contract change. Flagging for reviewer confirmation.

Refs #4633, #2476.

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -6711,12 +6711,18 @@ def _run_agent_streaming(
 
         return _usage
 
-    # Register this stream with the global streaming meter
-    meter().begin_session(stream_id)
-
     # Metering ticker — emits a metering event at 1 Hz while sessions are active.
     # When get_interval() returns >= 10.0 (no active sessions), the ticker exits
     # so no idle readings are emitted and the SSE consumer sees nothing.
+    #
+    # #4633/#2476: begin_session() and the ticker .start() are deferred into the
+    # outer `try` below so the outer `finally` (which pops STREAMS/CANCEL_FLAGS)
+    # always runs its paired end_session()/_metering_stop.set() teardown. A raise
+    # between here and that `try` would otherwise leak the _sessions[stream_id]
+    # entry — get_stats() only prunes sessions with first_token_ts > 0, so a
+    # zero-token turn (pre-flight cancel, setup raise) is never reclaimed and its
+    # count inflates the SSE `active` field. Deferring .start() until after `put`
+    # is defined also removes a latent start-before-put ordering window.
     _metering_stop = threading.Event()
 
     def _metering_ticker():
@@ -6732,7 +6738,6 @@ def _run_agent_streaming(
             put('metering', stats)
 
     _metering_thread = threading.Thread(target=_metering_ticker, daemon=True)
-    _metering_thread.start()
 
     def put(event, data):
         # If cancelled, drop all further events except the cancel event itself
@@ -6809,6 +6814,11 @@ def _run_agent_streaming(
     _ckpt_thread = None
     _agent_lock = None
     try:
+        # Register this stream with the global streaming meter and start the 1 Hz
+        # metering ticker. Kept INSIDE the outer try so the outer `finally`'s
+        # end_session()/_metering_stop.set() teardown is always paired (#4633/#2476).
+        meter().begin_session(stream_id)
+        _metering_thread.start()
         # Bind THIS turn's session identity to the worker thread/context BEFORE
         # any agent work (so every mid-turn notify_on_complete background spawn
         # captures THIS session, not a concurrent turn's process-global env).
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Reading `api/metering.py:71-200` on origin/master and the diff at this PR's HEAD (`ca750d35`), the leak this fixes is real and the mechanism is sound.

## Confirmed against master

`GlobalMeter.get_stats()` prunes only sessions that ever produced a token:

```python
# api/metering.py:141-146
stale = [
    sid for sid, s in self._sessions.items()
    if s.first_token_ts > 0 and (now - s.last_token_ts) > _STALE_SECS
]
```

A zero-token turn keeps `first_token_ts == 0.0` (set only inside `record_token`/`record_reasoning`), so it is never in `stale` and never reclaimed. And the payload reports the raw dict size:

```python
# api/metering.py:198
'active': len(self._sessions),
```

So every pre-flight cancel or pre-first-token raise permanently increments the SSE `metering` event's `active` for the process lifetime. On master `begin_session(stream_id)` at `streaming.py:6715` had no paired `end_session` anywhere — confirmed via grep, the only callers were `begin_session`, `record_token`, `record_reasoning`, `get_stats`, `get_interval`.

## The fix

Moving `begin_session()` + `_metering_thread.start()` inside the outer `try` (HEAD `streaming.py:6817-6819`) and pairing with `end_session(stream_id, 0)` + `_metering_stop.set()` in the outer `finally` (HEAD `streaming.py:9878-9894`) is the right shape — that `finally` is the same block that already pops `STREAMS`/`CANCEL_FLAGS`, so all three exit paths (normal, exception, early return) reach it. `end_session` is a bare `self._sessions.pop(stream_id, None)` (`metering.py:134-136`), so the idempotency and never-begun-safe claims hold, and the metering payload shape is unchanged since it ignores `final_output_tokens` today.

One nit worth noting rather than blocking: deferring `.start()` until after `put` is defined is described as closing a "start-before-put ordering window," but `_metering_ticker` captures `put` as a free variable resolved at call time, and the thread's first `put('metering', …)` only fires after a full `get_interval()`/`wait(interval)` cycle — so on master the closure would almost certainly have resolved `put` already. The move is still correct and cleaner; just calling out that the ordering risk was largely theoretical.

## Test coverage

`tests/test_metering_session_lifecycle.py` exercises the invariant directly against a fresh `GlobalMeter()` (not the singleton), and the assertions match the code: `test_get_stats_does_not_prune_zero_token_session` documents the leak, `test_end_session_reclaims_zero_token_session` / `test_end_session_is_idempotent` / `test_end_session_on_never_begun_stream_is_safe` cover the fix + edge cases. This is behavior-preserving cleanup with a good regression net — looks ready.

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (Codex SAFE + suite) · metering resource-leak fix → Nathan

**Certified head:** `sha:b5ae0ad7` (clean rebase, branch `gate-rebase/5787-metering-end-session`) · **PR:** #5787 · _ai-ag2026 (agent-authored), fix(streaming): pair metering begin_session with end_session teardown_
**Verdict:** GREEN. Pairs `meter().begin_session()` with an idempotent `end_session()` in the outer `finally`, fixing a per-stream metering-session leak (begin had no matching end anywhere). Codex SAFE (real 5,054-line review), full suite green.

### What I ran (rebased worktree `/tmp/wt-rebase-5787`) — Codex + full suite + `--collect-only`
| Gate | Result |
|------|--------|
| Rebase onto current master | ✅ `git apply` clean; `--stat` scope-scan clean |
| `--collect-only` metering-lifecycle test | ✅ 6 tests collect + pass |
| Codex (real: 5,054 lines, 101 tool calls, 0 auth) | **SAFE TO SHIP** — 0 findings |
| Full pytest suite | ✅ **12349 passed, 0 failed** |

### Findings
**✅ Clean resource-leak fix (+117/-4):** `meter().begin_session(stream_id)` had no matching `end_session()` anywhere → metering session state leaked per stream (esp. zero-token turns). The fix moves `begin_session()` INSIDE the outer `try` so the outer `finally` (which already pops STREAMS/CANCEL_FLAGS) always runs a paired `end_session(stream_id, 0)`; `end_session()` is idempotent (setting twice harmless), and the ticker `_metering_stop.set()` teardown is likewise paired — citing the prior #4633/#2476 finally-symmetry work. Codex confirmed: begin/end paired at ALL exit paths (normal, raise, cancel, early-return — the finally guarantees teardown per begin), idempotent (no double-decrement/KeyError), no end-without-begin window, and the `final_output_tokens=0` passed by the finally is currently ignored by `end_session` (safe — noted as a latent caveat if end_session ever starts consuming it). No metering-accuracy regression on the normal path. Full suite green (6 metering-lifecycle tests).

### Recommendation to Nathan
**GREEN — merge from branch `gate-rebase/5787-metering-end-session` (`sha:b5ae0ad7`).** A clean symmetric-teardown fix that plugs a per-stream metering-session leak, with the pairing guaranteed by the existing outer `finally`. Codex SAFE + full suite green. Agent-authored (@ai-ag2026), reliability class → your merge nod. (Latent note: the `end_session(stream_id, 0)` in the finally passes 0 tokens — harmless today since `end_session` ignores it, but if that changes it should supply the real total.) concept 4/5. crit=3, resource-leak/streaming reliability.

---
_Gate-certifier layer (warm-up → **gate** → release). meter().begin_session had no paired end_session → per-stream metering leak. Fix: begin_session moved inside outer try so outer finally always runs paired end_session(stream_id,0) (idempotent) + _metering_stop.set() ticker teardown (#4633/#2476 finally-symmetry). Codex real leg (5054 lines, 101 tool-calls, 0 auth) SAFE 0 findings (paired all exit paths, idempotent, no end-without-begin, final_output_tokens=0 currently-ignored-safe) + full suite green (0 failed, 12349) + 6 metering-lifecycle tests. resource-leak/streaming → Nathan. Cert valid for sha:b5ae0ad7._

nesquena-hermes:
Shipped in **v0.51.930** — thanks @ai-ag2026. 🎉 Metering `begin_session`+ticker moved inside the outer try so the paired `end_session`/ticker-stop teardown always runs; a pre-token cancel/raise no longer leaks a meter session. Codex SAFE + full suite 12,276/0. Deployed + verified live.

Issue #2653, opened 2026-05-20
Update check reports "Up to date" while hermes-agent is hundreds of commits past latest tag (3 user reports)


nesquena-hermes:
# Update check reports "Up to date" while the hermes-agent checkout is hundreds of commits behind the latest tag

## Summary

Three independent user reports in the last ~5 days describe the same symptom: Settings → System shows **Up to date ✓** even when the local hermes-agent checkout is clearly behind upstream. `git pull` in the agent directory immediately pulls a substantial amount of work, confirming the "Up to date" pill is wrong. Both periodic checks and the manual **Check now** button are affected.

### Deor (May 20)

> The Update check seems to be stuck. Its showing as up to date, but im on WebUi v0.51.89 and a manual hermes update pulled through a load of stuff! Thought it had been a quiet few days.

Source: Discord testers channel, May 20 2026 ~14:42 UTC.

![Deor report](https://raw.githubusercontent.com/nesquena-hermes/hermes-webui/issue-assets/issue-update-check-stuck-deor-may20.jpeg)

Per Deor in follow-up: yes, periodically checked automatically, and manual **Check now** in Settings also reports up to date.

### huhu1030 (May 20)

> Anyone has any problem with the update feature? I used to receive notification on the chat saying update is available + check now in the settings. However, I no longer get any messages, and the check now shows the latest version. I have to git pull and restart the process manually. I checked the issues on GitHub, but no report. Anyone else? Thank you.

System pill at time of report: `WebUI: v0.51.97  Agent: v2026.5.16-593-gedb2d9105  Check now  Up to date ✓`

That `-593-g…` suffix is the smoking gun (see Root cause below).

![huhu1030 report](https://raw.githubusercontent.com/nesquena-hermes/hermes-webui/issue-assets/issue-update-check-stuck-huhu1030-may20.jpeg)

### Maheśvara (May 20, 8:21 UTC)

> Regarding this, I noticed recently that you no longer capture hermes-agent updates. Was this on purpose? I previously use webui as my main update point but since like Monday, I've been having to manually run it in my CLI.

Same shape — the WebUI update banner stopped surfacing hermes-agent updates roughly a week ago, forcing users to fall back to manual `git pull`. He was wondering whether agent-update tracking had been disabled on purpose; it was not.

## Reproduction

On the host where the user reports come from (representative of any installation where the agent checkout is past the latest `v*` tag):

```
$ cd ~/.hermes/hermes-agent
[...]

Issue #2846, opened 2026-05-24
bug(updates): Agent Update Now loops — apply path aims at latest tag while check reports against origin/main


nesquena-hermes:
# Agent update banner loops: "Update Now" no-ops and the same N updates re-appear after restart

## Summary

After the **fix for #2653** shipped on master (`d04805b0`, "fix(updates): fall through to branch check when HEAD is past latest tag"), the update banner now *correctly* surfaces the real commit count for hermes-agent — but clicking **Update Now** appears to run, the page refreshes, and the same banner re-appears with the same count. The update is never actually applied.

Reported by Deor on Discord (May 24 2026, [link](https://discord.com/channels/1492983281338286121/1493325333980774622/1508107748578168923)):

> Hmm, may have spoken too soon. Agent updates seem to be in a loop. The number of updates available seems to be correct, and has gone up between tests, but clicking Update Now just seems to refresh the page and the number of updates stays the same.

![Banner reads "Agent (origin/main): 254 updates available"](https://raw.githubusercontent.com/nesquena-hermes/hermes-webui/issue-assets/issue-agent-update-loop-deor-may24.png)

> Banner: `↑ Agent (origin/main): 254 updates available` with **Update Now** button. After clicking, the page reloads and the banner reappears with `254 updates available` again.

The `(origin/main)` parenthetical in the banner is the smoking gun — it means the branch-fallback path is now correctly running, so the check side knows about the 254 commits, but the apply side is still aiming somewhere that's already current.

## Root cause

The check path (after PR #2758) and the apply path don't agree on **which ref to advance to**.

### Check path (`_check_repo_release`, `api/updates.py:374`)

After d04805b0, when `HEAD` is past the latest `v*` tag, `_check_repo_release` returns `None` and falls through to `_check_repo_branch`. `_check_repo_branch` counts commits between `HEAD` and `origin/<default-branch>` — this is where the **254** comes from, and where `branch: origin/main` enters the response (the banner shows `(origin/main)` because of it).

### Apply path (`_select_apply_compare_ref`, `api/updates.py:354`)

```python
def _select_apply_compare_ref(path):
    tags = _release_tags(path)
    if tags:
        return tags[0]               # ← always picks the latest TAG when tags exist

    upstream, ok = _run_git(['rev-parse', '--abbrev-ref', '@{upstream}'], path)
    if ok and upstream:
        return upstream

    branch = _detect_default_branch(path)
    return f'origin/{branch}'
```

[...]

PR #5790, opened 2026-07-08
Release channels: stable / experimental self-updater

nesquena-hermes:
## Summary

Adds a **release channel** setting (`stable` | `experimental`) so the self-updater stops advancing every install to the newest tag on every check — the root cause of stable users hitting breaking regressions.

One linear master, **two tag channels**:
- **`stable`** (glob `v*`, the DEFAULT) — soaked, promoted releases
- **`experimental`** (glob `exp-v*`) — every release batch, opt-in for testers

A channel is only a tag glob the updater reads — **no branches, no divergence**, so every existing ff-only guarantee (#2653/#2846/#3140) is preserved. Existing installs auto-land on `stable` (their historical `v*` glob never matches an `exp-v*` tag → zero action required).

This is PR 1 of the rollout: the **updater plumbing + Settings UX**. The release-workflow refactor (tag `exp-v*`, promotion engine) lives in the private workspace and activates when experimental tagging begins.

## What changed

**Backend (`api/updates.py`)**
- `channel` threaded through `_release_tags` / `_current_release_tag` / `_head_is_past_latest_tag` / `_check_repo_release` / `_check_repo` / `_select_apply_compare_ref` / `check_for_updates` / `apply_update` / `apply_force_update`
- Channel-scoped `git describe --match <glob>` (a commit tagged both prefixes resolves to the channel-correct tag)
- **Stable never falls through to `origin/master`** when HEAD contains the latest stable tag (reports up-to-date on the promoted subset) — the property that makes channels actually stop the firehose
- **Channel is WebUI-only**: the Agent repo always uses the default channel (it's a separate project that legitimately tracks master past its `v*` tags)
- **Rewind guard**: `apply_force_update` refuses to `reset --hard` onto a ref that is a strict ancestor of HEAD (no downgrade on channel switch)
- Update cache + in-progress guard keyed by `(channel, include_agent)`
- New `channel_version_badge()` + separate `update_channel_version` field — **`WEBUI_VERSION` stays channel-neutral** (it drives asset cache-busting / SW cache / stale-client skew, all exact-string equality)

**Config (`api/config.py`)** — `update_channel` default `"stable"`, enum `{stable, experimental}`

**Routes (`api/routes.py`)** — `/api/updates/check` reads the channel (accepts an explicit body `channel` so the dropdown's immediate re-check can't race the debounced settings autosave); settings endpoint exposes `update_channel` + `update_channel_version`

**Frontend** — channel dropdown in Settings → Preferences (between "Check for updates" and "Ignore Agent updates"), channel chip on the version badge, re-check on switch, i18n keys in **all 15 locales** (native translations), mobile badge overflow handling

## Testing

- New `tests/test_update_channels.py` — 14 real-git-fixture tests: tag-glob separation, channel-scoped describe, **stable-never-firehose** (the core guarantee), agent-resolution-identical-under-both-channels, rewind guard, channel-scoped cache, clear-lock channel preservation
- Full suite green (12,338 passed) with the existing update suite updated for the new signatures
- Live-verified on an isolated server (:8789, never prod): dropdown renders, API round-trip + enum validation + live channel-switch + badge suffix, zero JS errors, and the re-check race fix (explicit body channel wins over a stale saved setting)

## Reviewer notes

Agent-authored — please give it an independent review before merge (it touches the self-updater). Two design calls worth confirming:
1. The channel governs **WebUI only**; the Agent keeps its historical behavior.
2. `WEBUI_VERSION` is deliberately **not** channel-aware; the badge is a separate field.


nesquena-hermes:
## 🔬 Release-manager gate — GREEN after one on-branch fix · ⏸️ HELD for independent nesquena review (agent-authored self-updater)

Full gate (Codex + Fable + full suite) on head `d4120b47b`, rebased clean on current master.

**Codex — SHIP ONLY WITH FIXES → the one CORE is now fixed on-branch.**
- **[CORE, fixed]** `/api/updates/apply` + `/api/updates/force` ignored the request-body channel and re-read the saved setting, so a channel switch whose debounced settings-autosave hadn't landed yet could apply the *old* channel (banner offers `experimental`, apply reads saved `stable`, or vice versa). **Fix (`236a2c803`):** the banner now sends the channel the *check reported for that target* (`window._updateData[target].channel`) — what was actually offered — and the endpoints validate it against the enum and thread it into `apply_update(target, channel)` / `apply_force_update(target, channel)`. `None`/absent → the existing saved-setting fallback, so prior behavior is byte-preserved when the client sends nothing. Agent target stays channel-neutral server-side.
- Everything else Codex validated clean: stable never falls through to the `origin/master` firehose (all 3 branches guarded), the rewind guard refuses a pure-ancestor `reset --hard` while allowing legit forward + divergent-recovery, agent-always-default-channel on every path, `(channel, include_agent)` cache keying, `WEBUI_VERSION` stays channel-neutral.

**Fable — SHIP-UX.** Dropdown in the right home (Settings → Preferences), no crown-jewel surface, matches VS Code/Chrome channel convention, the critical asymmetry helper text ("switching back to Stable keeps your current version until Stable catches up") present in English + all 15 locales + echoed in the server rewind-refusal message, no auto-apply on switch, badge chip contrast fixed (light + dark).

**Full suite:** 12,270 passed / 0 failed (the fix broke one string-match timeout-guard test asserting the old apply-body shape — updated to the new channel-threaded shape while still asserting the 120s override; `test_api_timeout.py` 8/8).

**Disposition:** GREEN and decision-ready, but this is **agent-authored (nesquena-hermes) and touches the self-updater**, so per our standing rule it needs an **independent review from the nesquena account** (not the agent's own) before merge, plus a maintainer screenshot of the Settings dropdown (visible surface). Not auto-merged. Flagged to Nathan.

Also filed: the private-workspace plan now carries a full end-to-end pipeline-wiring audit (§11) — the announcement cron must switch to the stable glob before `exp-v*` tagging begins (else it would broadcast experimental tags publicly), the promotion cron needs registering, and the CHANGELOG stamp moves to promotion time. Those activate at flip-day (Phase 3), not now.

nesquena (approved):

Independent review of the **release-channel self-updater** (stable / experimental). Since this touches the update mechanism — `git describe`/`fetch`/`reset --hard` decisions driven partly by request data — the audit focused on injection, the rewind guard, and whether the channel property actually holds on both the check and apply sides.

### What this ships

`api/updates.py` (channel threading through the whole check/apply pipeline), `api/config.py` (`update_channel` enum setting), `api/routes.py` (body-channel on check/apply/force), frontend dropdown + version-badge chip + 15-locale i18n, `tests/test_update_channels.py` (+14 real-git-fixture tests) and signature updates to the existing update/timeout/security guards. One linear master, two tag globs (`v*` / `exp-v*`) — no branches, so the existing ff-only guarantees (#2653/#2846/#3140) are structurally untouched.

### Security audit — the load-bearing checks

1. **Channel injection is foreclosed at the source.** Every entry point funnels through `_normalize_channel`, which allowlists against the fixed `_CHANNEL_TAG_GLOBS` dict keys and falls back to `stable`; `_channel_tag_glob` maps only through that dict. A request-body `channel` is additionally enum-checked at the route (`not in ("stable","experimental") → fall back to settings`) before it ever reaches `check_for_updates`/`apply_*`. Git args are list-form (no shell). Arbitrary strings can never reach `--match`. ✓
2. **Rewind guard is correctly scoped.** `apply_force_update` refuses `reset --hard` only when `_head_contains_ref(compare_ref) and not _can_fast_forward_to(compare_ref)` — i.e. a **strict** ancestor. Equal refs (no-op reset for conflict recovery) and divergent lines (the historical force-update purpose) remain allowed; only a pure downgrade is refused with `refused_rewind: True`. Exactly the right predicate for "channel switch must not silently rewind an ex-experimental install." ✓
3. **The core channel property (stable-never-firehose) holds on BOTH sides.** `_select_apply_compare_ref` mirrors `_check_repo_release`: for `webui`+`stable`, when HEAD already contains the latest stable tag, it returns `None` (up-to-date on the promoted subset) instead of falling through to `origin/master` — and the check side reports `behind: 0` via `_channel_up_to_date_info`. The check/apply mirror matters because a mismatch is precisely the #2846 loop-forever class. ✓
4. **Agent stays channel-neutral, enforced at both call sites** — `check_for_updates` passes `DEFAULT_UPDATE_CHANNEL` for the agent repo, and `apply_force_update` overrides `channel = DEFAULT_UPDATE_CHANNEL` for `target == 'agent'` (the Codex gate caught that passing `experimental` made the agent ignore its `v*` tags). ✓
5. **Apply uses the *offered* channel, not the current setting.** The frontend sends the channel stamped on the update banner (`window._updateData[target].channel`) in the apply/force body, enum-validated server-side — so switching the dropdown mid-flow can't change what an already-offered update applies. And the check-side body channel wins over a stale debounced-autosave setting (the documented race fix). ✓
6. **`WEBUI_VERSION` stays channel-neutral** — the badge is the separate `update_channel_version` field, so asset cache-busting / SW cache / stale-client-skew equality checks are untouched. ✓
7. Cache + in-progress short-circuits are keyed on `(channel, include_agent)` — a channel switch can never be served the other channel's cached answer. ✓

### Greptile's three open findings — verified, all correctly triaged as non-blocking

- **Cross-channel `_check_in_progress`:** confirmed non-destructive by trace — the read side is gated by `cache_matches`, so a concurrent other-channel check only duplicates network I/O (last writer wins with its own channel key); wrong-channel data is never returned. A per-channel in-progress key would be polish.
- **`_channel_up_to_date_info` puts tag names in `current_sha`/`latest_sha`:** latent schema inconsistency only — with `behind: 0` the frontend never renders compare links. Worth normalizing whenever the payload schema is next touched.
- **`channel_version_badge` dead-code fallback:** cosmetic.

None meet the push bar; all three are accurately characterized in the PR discussion already.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Stable install, HEAD at latest `v*`, master 30 commits ahead | up-to-date; never offered master ✅ (core test) |
| Experimental install | historical branch fallthrough preserved ✅ |
| Agent repo under either WebUI channel | identical resolution (default channel) ✅ (tested) |
| Commit tagged both `v0.52.0` + `exp-v0.52.0` | channel-scoped `describe --match` resolves per-channel ✅ (tested) |
| Ex-experimental switches to stable, force-update | `refused_rewind` — no `reset --hard` downgrade ✅ (tested) |
| Body `channel: "evil*"` | enum-rejected → settings value; normalize falls back to stable ✅ |
| Channel switch then immediate re-check | body channel wins over stale autosave ✅ |
| Cache from other channel | never served (keyed match) ✅ (tested) |
| Clear-lock retry | preserves the experimental channel ✅ (tested) |
| Existing installs on upgrade | glob `v*` unchanged → land on stable, zero action ✅ |

### Tests

- `tests/test_update_channels.py` — **14/14** against real git fixtures; **fail-without-fix verified: all 14 fail on master's `updates.py`**.
- Update + timeout-guard + security-guard suites — **124 passed** total; the modified guards correctly re-pin the 60s/120s timeout overrides on the new body-building forms.
- `node --check` clean on both touched JS files. **Full suite: 12169 passed / 0 failed** in 276s (deselected the known darwin CRLF flake). CI fully green.

- With **zero** stable tags visible (fresh fork, no tags fetched), stable falls through to branch comparison — acceptable: there is no promoted subset to hold to, and every real install has `v*` tags.
- The two design calls flagged for confirmation are, in my judgment, both right: (1) channel as WebUI-only matches the agent's separate release cadence; (2) keeping `WEBUI_VERSION` channel-neutral is load-bearing for the exact-string-equality consumers.

### Recommendation

✅ **Approved clean.** Parked at approval — ready for the release agent's merge/tag pipeline (noting PR 2 — the `exp-v*` tagging/promotion workflow — activates the experimental side later; until then this is behaviorally inert beyond the new setting).

A carefully-built channel mechanism: allowlist-mapped globs (no injection path), a strict-ancestor rewind guard, the stable-never-firehose property enforced symmetrically on check and apply, agent channel-neutrality pinned at both call sites, and offered-channel (not current-setting) apply semantics. The known residual races are provably non-destructive. 14 real-git-fixture tests all fail without the fix; full suite clean. Ship.

nesquena-hermes:
Shipped in **v0.51.929**. 🚀

PR 1 of the release-channels rollout (updater plumbing + Settings UX) is live on prod — `update_channel` defaults to `stable` and the existing `v*` glob is unchanged, so every current install stays exactly where it is (zero action, no behavior change until experimental tagging begins).

Full gate: Codex (1 CORE — apply/force debounce race — fixed on-branch) + Fable SHIP-UX + independent nesquena review (clean, end-to-end) + suite 12,270/0. Verified live: `update_channel` present on the settings API, dropdown + asymmetry helper + version-badge chip all render, no auto-apply on switch.

Next (private-workspace plan §11): batch-lane exp-v* tagging, promotion engine + cron, the #experimental Discord feed, and the flip-day cutover — implemented incrementally, not yet active. Thanks for the design consensus (Codex + Fable + Nathan).

PR #5792, opened 2026-07-08
fix(csv): keep media downloads visible in previews

ruizanthony:
## Summary
- Keep a visible download affordance on `MEDIA:` CSV previews.
- Reuse `/api/media?...&download=1` with an HTML `download` attribute for CSV files.
- Keep the download link visible even when the inline CSV preview fails.

## Test plan
- `python3 -m pytest tests/test_csv_table_rendering.py -q`

## Notes
This preserves the existing inline CSV table preview while restoring the attachment-style download path for exchanging CSV files through chat output.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -2474,6 +2474,9 @@
 
   /* ── CSV table rendering ── */
   .csv-table-wrap{margin:6px 0;overflow-x:auto;border:1px solid var(--border);border-radius:8px;background:var(--surface);}
+  .csv-preview-header{justify-content:space-between;flex-wrap:wrap;row-gap:6px;}
+  .csv-preview-title{min-width:0;overflow:hidden;text-overflow:ellipsis;white-space:nowrap;}
+  .csv-download-link{margin-left:auto;text-transform:none;letter-spacing:0;font-size:12px;}
   .csv-table{width:100%;border-collapse:collapse;font-size:13px;}
   .csv-table thead{position:sticky;top:0;z-index:1;}
   .csv-table th{background:var(--hover-bg);font-weight:600;text-align:left;padding:8px 12px;border-bottom:2px solid var(--border2);white-space:nowrap;user-select:none;}
--- a/static/ui.js
+++ b/static/ui.js
@@ -16894,7 +16894,18 @@ function loadDiffInline(container){
 
 const CSV_MAX_SIZE=256*1024; // 256 KB cap for inline CSV rendering
 
-function buildCsvTablePreview(path, text){
+function _mediaSessionQuery(){
+  const mediaSessionId=(typeof S!=='undefined'&&S&&S.session&&S.session.session_id)?String(S.session.session_id):'';
+  return mediaSessionId?'&session_id='+encodeURIComponent(mediaSessionId):'';
+}
+
+function _csvMediaUrl(path, opts={}){
+  let url='api/media?path='+encodeURIComponent(path)+_mediaSessionQuery();
+  if(opts.download) url+='&download=1';
+  return url;
+}
+
+function buildCsvTablePreview(path, text, downloadUrl=''){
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Reading the diff at HEAD (`87a9e25e4`) against `origin/master` — `static/ui.js` `buildCsvTablePreview` / `loadCsvInline` (16894-16952), the new `_csvMediaUrl`/`_mediaSessionQuery` helpers, plus `api/routes.py` `_handle_media` (17333+) and `_session_media_token_allows_path` (17279-17318).

## Summary

The core of this PR — restoring a visible, downloadable `📎 <fname>` affordance on both the successful CSV table preview and the error card — is a clean, well-scoped fix, and the added tests pin exactly the strings that matter (`csv-download-link msg-media-link`, the `download="${esc(fname)}"` attribute, the shared-helper reuse). No objection to that part.

One thing to flag about the second half of the change (threading `session_id` onto the CSV fetch via `_mediaSessionQuery`), because it reads like it's meant to broaden which CSVs can load, and it doesn't actually do that.

## Code reference

`_session_media_token_allows_path` only returns true when the resolved mime is in the allow-set it's handed, and the set built in `_handle_media` is:

```python
_SESSION_MEDIA_TOKEN_TYPES = _INLINE_IMAGE_TYPES | _AUDIO_VIDEO_PDF_TYPES | {"text/html"}
session_media_allowed = _session_media_token_allows_path(
    qs.get("session_id", [""])[0], target, _SESSION_MEDIA_TOKEN_TYPES,
)
```

CSV's mime never enters that set. `.csv` isn't in `MIME_MAP` (config.py:1005-1034), so `MIME_MAP.get(target.suffix.lower(), "application/octet-stream")` returns `application/octet-stream`, which isn't in `_INLINE_IMAGE_TYPES`, `_AUDIO_VIDEO_PDF_TYPES`, or `{"text/html"}`. So for a CSV, `_session_media_token_allows_path` returns `False` regardless of whether `session_id` is present.

## Diagnosis

CSV inline previews that render today succeed via the **`within_allowed`** branch (path inside hermes-home / `/tmp` / active workspace), not the session-token branch. Adding `session_id` to the fetch URL is harmless but a no-op for authorization — a CSV that lives *outside* the allowed roots still 403s with or without it. If the intent was purely consistency with the sibling media link at ui.js:7041, that's fine; if the intent was to let session-scoped CSVs outside allowed roots preview, that won't happen without also mapping `.csv` in `MIME_MAP` and adding its mime to `_SESSION_MEDIA_TOKEN_TYPES` (a larger, security-sensitive change I would not fold into this PR).

Separately, the download itself is robust either way: for `application/octet-stream` the `_handle_media` disposition logic (routes.py:17613-17618) already falls to `"attachment"`, and the HTML5 `download` attribute forces save on this same-origin link — so the affordance works.

## Recommendation

Ship the download-link restoration as-is. Either drop the `session_id` threading (since it buys nothing for CSV) or add a one-line comment noting it's for parity with other media links and does not change CSV auth, so a future reader doesn't assume it gates session-scoped access.

nesquena-hermes:
Shipped in **exp-v0.52.28** (experimental channel) — thank you @ruizanthony! 🎉

CSV files shared in chat keep their download link again: the inline table preview now shows a 📎 download link in its header (via `api/media?…&download=1`), and the link stays visible even when the inline table fails to render, so a `MEDIA:` CSV stays retrievable — matching every other media type.

Gate: Codex SAFE (esc()-escaped, no XSS, backward-compatible, scoped to the CSV branch), full suite green, CI 18/18, and Nathan visually signed off on the restored affordance. Clean fix.

nesquena:
@ruizanthony Email me (see profile), I'd like to invite you to our dev chat. Appreciate your contributions!

PR #5802, opened 2026-07-08
fix(webui): render MEDIA: tokens inline during smd streaming

silent-reader-cn:
## Problem

`renderMd()` (in `static/ui.js`) maps `MEDIA:<path-or-url>` tokens to inline `` / `<video>` / `<audio>` / SVG / PDF / etc. via stash + restore passes, so the **settled** assistant message shows real media in the chat window.

The streaming `smd` renderer used by the **live** assistant body never went through that pass. Any `MEDIA:<ref>` token that arrived mid-turn showed up as the raw path string until the turn settled and the full re-render swapped it for the real DOM — a flash that is especially visible for emoji / reaction images the agent sends between tool calls (e.g. the catgirl persona sending `MEDIA:C:\Users\...\03_laugh.png`).

## Fix

1. Extract the URL/media → HTML mapping from `renderMd()` into a shared helper `_inlineMediaHtmlForRef(ref, sessionId)` (in `static/ui.js`). `renderMd`'s restore pass is now a one-liner that delegates to this helper. The settled + live paths now produce byte-identical DOM for the same token, so the streamed image can't diverge from the settled one.

2. Add `_smdMediaAwareAddText` in `static/messages.js` that wraps smd's `add_text` to detect `MEDIA:` tokens in any incoming text chunk and splice the resulting `` / `<video>` / `<a>` / `<div data-path=...>` node directly into the smd parent via a `DOMParser`-created fragment. Plain prose still delegates to the renderer it intercepted, so:
   - the fade renderer keeps its word-by-word animation,
   - the safe renderer keeps its O(n) plain text-node insertion,
   - neither can recurse into our own wrapper.

3. Route `_safeSmdRenderer.add_text` and `_streamFadeRenderer.add_text` through the interceptor so live prose in both modes renders media identically.

## Tests

- New `tests/test_smd_media_in_stream.py` locks in the architectural invariants (helper exists, both renderers delegate to it, recursion guard).
- `tests/test_renderer_js_behaviour.py` driver now ships a minimal `_inlineMediaHtmlForRef` stand-in so `renderMd` can be invoked standalone under `node` without the rest of `ui.js` in scope (matches the patterns `renderMd` already relied on for `_IMAGE_EXTS` / `_SVG_EXTS` / `_AUDIO_EXTS`).
- `tests/test_svg_audio_video_rendering.py` static-count assertions relaxed from `>=2` to `>=1` because audio/video/SVG markup now lives in a single shared function rather than being copy-pasted inside `renderMd`'s stash-restore closure.
- `tests/test_media_inline.py` unchanged — all 62 tests still pass.

Local run on Windows, uv-managed Python 3.11, `s/`-prefixed suite:

```
218 passed in 30.23s
```

covering `tests/test_media_inline.py`, `tests/test_smd_media_in_stream.py`, `tests/test_svg_audio_video_rendering.py`, `tests/test_renderer_comprehensive.py`, `tests/test_renderer_js_behaviour.py`, `tests/test_blockquote_rendering.py`.

## Notes

- Does **not** alter the `/api/media` endpoint, session-id allow-list, or any server logic — purely a rendering-side change. Local paths still go through the same allow-list check because the helper reuses the same `api/media?path=...&session_id=...` URL shape the settled path used.
- Skips `:Modal flow` recursion by detecting `MEDIA:` and using a `DOMParser` instead of `innerHTML` for the spliced fragment — `<script>` cannot execute.
- UI scripts load `ui.js` (line 1714) before `messages.js` (line 1719), so the helper is visible to the streaming wrapper at execution time. No new script tags, no new build step.


nesquena-hermes:
Reading `static/ui.js` `renderMd()` and the new `_smdMediaAwareAddText` in `static/messages.js` against the diff at this PR, the shared-helper extraction (`_inlineMediaHtmlForRef`) is clean and the settled path is unchanged. But the streaming interceptor does **not** produce byte-identical DOM to the settled path for mixed prose+MEDIA chunks — it changes how the surrounding prose is treated, from escaped text to parsed HTML.

## Code reference

Settled `renderMd()` stashes MEDIA tokens as placeholders *first*, then escapes the surrounding prose, then restores the placeholder as trusted media HTML:

```js
// ui.js:6447 — MEDIA stashed before any other processing
s=s.replace(/MEDIA:([^\s\)\]]+)/g,(_,raw_ref)=>{ media_stash.push(raw_ref); return '\x00D'+(media_stash.length-1)+'\x00'; });
// ...later, prose is HTML-escaped (ui.js:388 paragraph pass): s=esc(s)...
// ui.js:6958 — only the placeholder becomes trusted HTML
s=s.replace(/\x00D(\d+)\x00/g,(_,i)=>_inlineMediaHtmlForRef(media_stash[+i]));
```

The streaming interceptor instead concatenates the raw prose slices into one HTML string and hands the whole thing to `DOMParser`:

```js
// messages.js — plain prose slices go in verbatim, un-escaped
if(m.index>last) html+=value.slice(last,m.index);
// ...media token -> _inlineMediaHtmlForRef...
if(last<value.length) html+=value.slice(last);
const doc=new DOMParser().parseFromString('<div>'+html+'</div>','text/html');
```

## Diagnosis

Because the prose slices (`value.slice(...)`) are placed into an HTML string rather than inserted as text nodes, any `<`, `&`, `>` in the same chunk as a MEDIA token is now HTML-parsed instead of shown literally. The settled path escapes those exact characters at `ui.js:280`/`:388`. So a chunk like `see MEDIA:/tmp/a.png and 3 < 5` renders `3 < 5` fine when settled but drops the `< 5` when streamed — the "byte-identical DOM" invariant the PR is built on doesn't hold for mixed chunks.

This is also a security regression relative to the settled path. `DOMParser` does not run `<script>` (the PR notes this correctly), but it *does* build live element nodes with intact inline handlers. Prose such as `MEDIA:/x.png ` puts the `` into `html`, parses it into a real node, and `parent.appendChild(frag)` inserts it live — `onerror` then fires. The settled renderer never had this exposure because assistant/tool prose is always escaped before insertion.

## Fix sketch

Escape the prose slices before they enter the HTML string, so only the `_inlineMediaHtmlForRef` output is trusted markup — mirroring the stash-then-`esc` ordering in `renderMd`:

```js
if(m.index>last) html+=esc(value.slice(last,m.index));
// media branch unchanged (already trusted HTML from _inlineMediaHtmlForRef)
if(last<value.length) html+=esc(value.slice(last));
```

`esc` is the same top-level global (`ui.js:280`) this file already uses at `messages.js:4413`/`:4501`, so no new import. This keeps media rendering while restoring text-safety and the byte-identical guarantee for the prose portion.

## Test plan

The new `tests/test_smd_media_in_stream.py` locks structural invariants (helper exists, recursion guard) but not output equivalence. Worth adding a case that feeds a mixed chunk containing both a MEDIA token and a literal `<`/`&`/`` and asserts the streamed DOM's text content matches the settled `renderMd()` output for the same input — that's the invariant the "byte-identical" claim needs and the one this interceptor currently breaks.

silent-reader-cn:
Updated with fixes for both review issues (commit `e031c5a6`):

**Issue #1 (Security — prose parsed as DOM):** Original `_smdMediaAwareAddText` concatenated prose + media HTML into one string and fed it to `DOMParser`. Fixed: each prose slice is routed through `baseAddText` (→ `createTextNode` → safe text node). Only each **single** MEDIA token's output (from the trusted `_inlineMediaHtmlForRef` helper) is parsed via `DOMParser`. Plain text with `` can never become a live element.

**Issue #2 (Split tokens):** Tokens split across smd flushes (e.g. `"MEDIA:"` then `"foo.png"`) are buffered in a per-parser `WeakMap`-keyed tail buffer and completed on the next `add_text` call. The tail detection was also improved: `rest.endsWith('MEDIA:')` / `endsWithOpenRef` instead of the earlier `!/[\s\)\]]/.test(rest)` which incorrectly rejected `"before MEDIA:"`.

New functions: `_smdAppendMediaNode` (single-token DOMParser), `_smdMediaTailSet`, `_smdMediaTailClear`. Both `_safeSmdRenderer` and `_streamFadeRenderer` now pass parser identity + tail map. 222 tests passing.

nesquena-hermes:
Re-reviewed at head `c1f1679c0` (merge of master into the branch) — read `static/messages.js` and `tests/test_smd_media_in_stream.py` against my two prior findings. Both are genuinely fixed. Nice work.

## Issue #1 (prose parsed as DOM) — resolved

The interceptor no longer builds one HTML string and hands it to `DOMParser`. `_smdMediaAwareAddText` now slices the combined chunk and routes prose runs straight through `baseAddText` (which is smd's `createTextNode` path), while only a single MEDIA token's trusted output reaches the parser:

```js
// messages.js:4353 — prose slice goes to a text node, never parsed
if(m.index>last){
  const slice = combined.slice(last, m.index);
  if(baseAddText) baseAddText(data, slice);
}
...
_smdAppendMediaNode(parent, m[1]);   // only this touches DOMParser
```

And `_smdAppendMediaNode` (messages.js:4388) is fed *only* `_inlineMediaHtmlForRef(rawRef)`, never raw chunk text:

```js
const mediaHtml = (typeof _inlineMediaHtmlForRef==='function')
  ? _inlineMediaHtmlForRef(String(rawRef)) : '';
const doc=new DOMParser().parseFromString('<div>'+mediaHtml+'</div>','text/html');
```

So a chunk like `see MEDIA:/a.png and 3 < 5` now keeps `3 < 5` as literal text, and `MEDIA:/x.png ` inserts the `` prose as a text node — the live-node XSS path I flagged is closed. `test_plain_text_does_not_go_through_dom_parser` (line 150) locks this by asserting the old `parseFromString('<div>'+value+'</div>'` shape is gone.

## Issue #2 (split tokens) — resolved, and better than my sketch

The per-parser `WeakMap` tail buffer (`_SMD_MEDIA_TAIL`, messages.js:4411) handles `"MEDIA:"` then `"foo.png"` across flushes. I also like the `_smdMediaRefHasReliableBoundary` refinement (messages.js:4309): an extensionless URL is held until a delimiter or stream-end rather than emitted prematurely, and a `?`/`#`-terminated ref is treated as still-incomplete. The `_MEDIA_TAIL_MAX = 4096` cap (line 4300) guards a runaway buffer from a malformed stream — a good defensive touch I hadn't asked for.

## One thing I checked closely: stream-end ordering

Since the tail buffer can hold a completed-but-unflushed MEDIA token when the stream ends, flush-before-clear ordering matters. Both teardown paths get it right:

- `_smdEndParser` (messages.js:4048) flushes `_smdParser` and `__SMD_PARSER_FALLBACK` *after* `parser_end`, then clears — verified by `test_smd_end_parser_clears_fallback_media_tail` (258), including the `assertLess(...parser_end < flush...)` ordering assertion.
- `_clearAnchorProseIncrementalNode` (messages.js:3730) flushes each cached parser's tail before clearing it (line 3734).

The `__SMD_PARSER_FALLBACK` sentinel object (line 4409) instead of a string key is the right fix for the WeakMap-key-must-be-object constraint.

## Remaining gap (non-blocking): tests are structural, not behavioral

`tests/test_smd_media_in_stream.py` asserts *source-string* invariants (helper exists, old DOMParser shape absent, flush-before-clear ordering). That's a reasonable guard given the file can't spin up a real DOM, and the header comment says as much. But the actual invariant my original review asked for — "streamed DOM text content equals settled `renderMd()` output for a mixed prose+entity+MEDIA chunk" — still isn't exercised as behavior; a future refactor that reintroduces entity-decoding of prose could pass every current assertion. If the repo has any jsdom/node harness (there's Node-executed frontend coverage elsewhere, e.g. #5865's aux tests), a single equivalence case feeding `3 < 5 & MEDIA:/a.png ` and comparing `.textContent` would close that loop. Not a merge blocker — the source guards are adequate for now.

Net: both review issues are correctly addressed, the split-token handling exceeds the original ask, and teardown ordering is sound. LGTM from my side.

nesquena-hermes (changes requested):
## 🔬 Gate review (Codex, security + correctness) — SHIP ONLY WITH FIXES

Thanks @silent-reader-cn — the **security is clean**, which was the main risk for putting media-token→HTML on the LIVE stream: Codex found no attribute/tag breakout or executable-scheme vector — references are escaped, `javascript:`/`data:` become encoded `/api/media` paths, whitespace terminates the token, local paths keep the `/api/media` root + state-file enforcement (same scoping as the settled renderer), SVG stays an ``/attachment, and the scan regexes are linear (no ReDoS). Good.

Four correctness fixes on the live-streaming path before ship:

1. **[functional — partial token] `static/messages.js:4258-4273, 4468-4474`** — a `MEDIA:` marker split across smd chunks (notably at paragraph start) isn't buffered, so it mis-renders. Implement a stable per-parser rolling scanner that buffers partial `MEDIA:` prefixes, and prove paragraph-start + every cross-chunk split **with the real vendored SMD parser** (see note below).
2. **[functional — parser key] `static/messages.js:4086-4106, 4359-4367, 4450-4462`** — use that same stable parser key for tail storage, end-flush, and cleanup, including anchor parsers and extensionless final URLs.
3. **[SILENT — wrong element] `static/ui.js:2438, 2469-2474, 2523-2524`** — a remote `.webm` renders as `<audio>` because the audio regex matches before video; resolve video before the overlapping audio regex so `.webm` stays `<video>`.
4. **[SILENT — dead placeholders] `static/messages.js:4433-4448`** — live PDF/HTML/diff/CSV/Excalidraw tokens append only their `*-load` placeholders; the hydration functions (which run from settled post-processing) are never invoked for the live insertion, so those media types aren't functional mid-stream. Call the relevant loaders on the newly-inserted subtree after appending.

**Test note:** the new `test_smd_media_in_stream.py` (280 lines) passes but is **static source-assertions**, not real-parser behavioral coverage — which is exactly why the paragraph-start partial-token failure stayed hidden. Please add behavioral tests that drive the actual vendored SMD parser with chunk-split `MEDIA:` inputs.

Security's the hard part and it's solid; these are functional-completeness gaps on the live path. Once the rolling scanner + hydration + webm ordering land with real-parser tests, re-gate.

nesquena-hermes (changes requested):
## 🔬 Re-gate round 2 (Codex, against the real vendored SMD parser) — SHIP ONLY WITH FIXES (all 4 still open)

Thanks @silent-reader-cn — security stays **clean** (verified again: local / `javascript:` / `data:` / `file:` all become encoded `api/media?path=…&session_id=…`, HTTPS attrs escaped, SVG stays ``, only helper-generated markup enters `DOMParser`, main-parser flush re-sanitized) and **no ReDoS** (bounded char classes, retained state capped at 4096). But re-gating against the **real vendored SMD 0.2.15 parser** (not a mock), all 4 correctness findings still reproduce — the `_smdMediaTailFlush` scaffolding was added but not wired into an actual rolling scanner, and the new tests are static so they miss the real callback fragmentation:

1. **[NOT FIXED — split-prefix buffering] `static/messages.js:4270`** — still a full-token-only gate; a `MEDIA:` prefix split across SMD callbacks (`MEDIA:foo` → `.png`) still mis-renders. Replace with a **true rolling prefix scanner**, exercised against vendored SMD 0.2.15 in **both fade and non-fade** modes.

2. **[NOT FIXED — parser-key stability] `static/messages.js:4258`** — bind **every** main and anchor renderer to a **stable per-parser key**, and use it consistently for tail storage, stream-end flush, rebuild, and cleanup (extensionless final URLs still mishandled).

3. **[NOT FIXED — remote .webm] `static/ui.js:2523`** — classify `_VIDEO_EXTS` **before** `_AUDIO_EXTS` so `.webm` produces `<video>` not `<audio>` (this also means the settled renderer regression check doesn't fully pass).

4. **[NOT FIXED — live lazy hydration] `static/messages.js:4447`** — invoke **all five** lazy-media hydration functions (`loadDiffInline`/`loadCsvInline`/`loadExcalidrawInline`/`loadPdfInline`/`loadHtmlInline`) immediately after live fragment insertion — currently the placeholders are appended but the hydration fns are never called for the live path.

237 tests pass but they're **static** (assert source shape), so they don't exercise the real SMD fragmentation — please add a behavioral test that feeds split chunks through the actual vendored parser and asserts the rendered node. The security + ReDoS posture is solid; it's the four functional/streaming-fragmentation items that need to genuinely land (the tail-flush needs to BE a rolling scanner, not just exist). Re-gate once the split-chunk case renders correctly against vendored SMD.

silent-reader-cn:
Re-gate follow-up pushed at `9b6559df`.

Addressed the four live MEDIA correctness blockers from the latest review:

- Added a rolling `MEDIA:` prefix scanner so split sentinel/ref chunks and final extensionless URLs are buffered and flushed correctly.
- Bound both main and anchor SMD renderers to stable parser identity for tail storage, stream-end flush, rebuild, and cleanup.
- Classified `.webm` as video before the overlapping audio extension detection.
- Scheduled the shared live post-process pass after streamed MEDIA insertion so PDF/HTML/diff/CSV/Excalidraw placeholders hydrate immediately.

Added behavioral coverage using the real vendored `static/vendor/smd.min.js` parser in both safe and fade renderers for split `MEDIA:` prefixes, ref tails, final extensionless URLs, and live PDF hydration.

Local verification:

- `pytest tests/test_smd_media_in_stream.py tests/test_svg_audio_video_rendering.py tests/test_pdf_html_preview.py tests/test_smooth_text_fade.py` -> 103 passed
- `node --check static/ui.js`
- `git diff --check`

CI and Greptile are green on `9b6559df`. @nesquena-hermes could you re-review / clear the stale `CHANGES_REQUESTED` when you have a moment?

nesquena-hermes (changes requested):
## 🔬 Re-gate round 3 (Codex, real vendored SMD) — all 4 findings FIXED ✅, but the scanner regressed normal-prose rendering

Big progress @silent-reader-cn — all four round-2 findings are now verified fixed against the **real vendored SMD 0.2.15** parser, plus security + perf hold:
1. ✅ Split-prefix buffering — `MEDIA:foo` + `.png` (and every prefix split) produces media nodes in **both fade and non-fade** modes.
2. ✅ Parser-key stability — main/anchor parsers bind their real identities, interleaved bound parsers stay isolated, main/anchor/fallback cleanup flush+clear tails, extensionless final URLs render at stream end.
3. ✅ Remote `.webm` → `<video>` (not `<audio>`).
4. ✅ Lazy hydration — live insertion reaches `postProcessRenderedMessages()`, all five hydrators called.
5. ✅ Security CLEAN — local/file/data/javascript escaped behind `api/media`, hostile attrs escaped, SVG stays ``.
6. ✅ No ReDoS — 12.5M-char stress completed in tens of ms.

But wiring the MEDIA scanner into the prose path **regressed normal (non-MEDIA) streaming rendering** — 2 reproduced (this is the crown-jewel render path, so it's a blocker):

1. **[REGRESSION — prose ownership / cross-parent concat] `static/messages.js:4360`** — tail entries don't preserve the original SMD parent/plain-text writer, so a tail can concatenate across parents. **Fix:** preserve the original SMD parent/plain-text writer in each tail entry, flush through that **stable owner**, and refuse cross-parent tail concatenation.

2. **[REGRESSION — fade lost on ordinary prose] `static/messages.js:4271`** — scanner prose and false-prefix tails bypass the fade appender, so **ordinary prose loses the configured fade effect**. **Fix:** route scanner prose + false-prefix tails through a **non-recursive fade appender** so normal prose keeps the fade.

238 tests pass but they don't assert the live fade-ownership on normal prose interleaved with the scanner, which is why these slipped. The MEDIA feature itself is solid now — the last mile is making the scanner **transparent to normal prose** (preserve the original writer + fade path for non-MEDIA text). Add a regression: plain prose streamed through the scanner path retains fade + correct parent ownership. Re-gate once normal-prose rendering is byte-identical to master. Really close — nice work getting the real-SMD split-chunk + security + hydration all green.

silent-reader-cn:
Round 3 follow-up pushed at `9169f636`.

Addressed the normal-prose scanner regressions from the latest review:

- Tail entries now preserve their original parent and text writer (`writeText`) along with the parser key. If the next SMD callback belongs to a different parent/writer, the buffered tail is flushed through the original owner and the new chunk is scanned independently, so false MEDIA-prefix tails cannot concatenate across DOM parents.
- The fade renderer now passes a stable non-recursive fade writer into the MEDIA scanner. Scanner prose, false-prefix flushes, and fallback text writes go through `_streamFadeAppendText(...)`, so ordinary prose that only touched the scanner because it looked like a partial `MEDIA:` prefix keeps the configured word fade.

Added real vendored-SMD behavioral coverage for both regressions:

- `['M', 'aybe plain prose ']` in fade mode must render normal prose with `stream-fade-word` spans.
- `['- ME', '\n- ow']` must leave list item ownership as `['ME', 'ow']` instead of moving/concatenating the tail into the second parent; fade mode also verifies the text still routes through fade spans.

Local verification:

- `pytest tests/test_smd_media_in_stream.py tests/test_svg_audio_video_rendering.py tests/test_pdf_html_preview.py tests/test_smooth_text_fade.py` -> 107 passed
- `node --check static/ui.js`
- `git diff --check`

This should make the MEDIA scanner transparent to normal prose while keeping the round-2 MEDIA fixes intact.

nesquena-hermes:
Shipped in **exp-v0.52.61** 🚀 — MEDIA: tokens (images, video, audio, PDF, diff, CSV, Excalidraw) now render inline in real time as the response streams, including markers split across streaming chunks. This took real work over 4 rounds and landed clean: Codex SAFE (both prose regressions from round 3 closed — original writer/ownership preserved + fade retained on normal prose, verified byte-identical to master when no MEDIA is present; all 4 original findings fixed; security clean — javascript:/data:/file: neutralized, SVG stays img, routed through escaped /api/media; no ReDoS), stream-gate GREEN all 3 modes, 227 tests + 18 real-vendored-SMD subtests. Thanks @silent-reader-cn — excellent persistence on a tricky crown-jewel render path!

PR #5823, opened 2026-07-09
fix(#5585): queue gateway-owned steer fallback

### LINKED ISSUE #5585: bug(steer): steering a live stream fails with "No agent cached for this session" — SESSION_AGENT_CACHE miss during an active run
## Symptom

Trying to **steer** (submit a message while a response is still streaming) surfaces the toast:

> **No agent cached for this session — try refreshing**

Reported to happen consistently ("whenever trying to submit a message while a response is streaming"), not intermittently. Steering is exactly the busy-input action that should work mid-stream, so this leaves the user with no working way to guide an in-flight run short of a page refresh.

## Where it comes from (code-grounded, current head)

- Client: `static/commands.js` `_trySteer()` posts to `POST /api/chat/steer`. When the response is `{accepted: false, fallback: "no_cached_agent"}`, it shows the `steer_fail_no_cached_agent` i18n string ("No agent cached for this session — try refreshing").
- Server: `api/streaming.py` `_handle_chat_steer()` returns that fallback here:
  ```python
  with _cfg.SESSION_AGENT_CACHE_LOCK:
      cached = _cfg.SESSION_AGENT_CACHE.get(sid)
      ...
  if not cached:
      return j(handler, {"accepted": False, "fallback": "no_cached_agent", "stream_id": None})
  ```
  So the steer path looked up `SESSION_AGENT_CACHE[sid]` and found **nothing**, even though a turn is actively streaming for that session.

The contradiction is the bug: a streaming turn is supposed to have populated `SESSION_AGENT_CACHE[session_id]` — either the reuse branch (`streaming.py` ~L7770, cache hit + `move_to_end`) or the new-agent insert (`streaming.py` ~L7883, `SESSION_AGENT_CACHE[session_id] = (agent, _agent_sig)`). If steer can't find the entry while the stream is live, one of these is happening:

1. **Insert-timing race.** The agent is inserted into the cache only *after* construction completes on the streaming thread (~L7883). If the user steers in the window between send and that insert (agent construction can take a beat), the lookup misses. This would be intermittent, though — doesn't fit a consistent "whenever" repro.
2. **Identity/signature eviction.** `_handle_chat_steer` pops the entry when `_cached_agent_matches_session(agent, sid)` is false (~L9770), and the streaming reuse branch pops on `_agent_sig` mismatch or identity mismatch (~L7770/7777). If the cached agent's session identity or `_agent_sig` doesn't match what steer/stream compute for the same session, the entry gets evicted and steer reports `no_cached_agent`. A consistent repro points here (or at #3): worth logging the `_agent_sig` + `_cached_agent_session_identity` at both the insert site and the steer lookup for the same `sid` to see if they diverge.
3. **Agent never cached for this turn.** If this session's turn ran through a path that doesn't insert into `SESSION_AGENT_CACHE` (or caching is disabled — `HERMES_WEBUI_AGENT_CACHE_MAX=0`), steer will always miss. Confirm `SESSION_AGENT_CACHE_MAX` > 0 in the reporter's env and that the insert at ~L7883 actually ran for the live stream.

## What would nail it down (needinfo)

1. Reporter's **WebUI version** (steer + agent-cache code has churned; a version pins the regression window).
2. Is it **every** session or a specific one (e.g. a reopened/older session, or right after a profile switch)? Reopened-session and post-profile-switch turns go through the identity/signature check that can evict.
3. Server logs around the steer attempt — grep the session id for `Evicted cached agent before steer due to mismatched session identity`, `Evicted cached agent with mismatched session identity`, and `Created new agent for session` — those lines tell us directly whether the entry was evicted vs. never inserted vs. present-but-stale.
4. Value of `HERMES_WEBUI_AGENT_CACHE_MAX` (default 25) — a 0/low value or a very busy multi-session instance could evict the live entry (though the LRU trim at ~L7892 is guarded to skip active sessions, so this is a long shot).

## Suspected fix direction (confirm root cause first)

If it's the **timing race** (#1): populate `SESSION_AGENT_CACHE[session_id]` earlier — as soon as the agent instance exists / the stream is registered — rather than only after full construction, so a steer during startup finds it.

If it's the **identity/signature mismatch** (#2): reconcile the identity/`_agent_sig` computation between the stream insert path and the steer lookup so the same live session doesn't compute a mismatching key (the steer eviction and the stream reuse eviction must agree on identity).

Keep steer's active-run-guidance contract intact: a genuine no-agent state must still fall back gracefully without cancelling the run.

## Not a duplicate of

- #4749 (closed) — that was the **`"Steer unavailable — draft restored"`** toast being unactionable (a different `fallback` reason + busy-input UX), fixed by improving the copy/affordance. This issue is specifically the **`no_cached_agent`** fallback reason firing during a live stream — the cache lookup missing an agent that should exist. Same endpoint, different root.

---
_Filed from a report in the Discord #report-bugs channel._

rodboev:
## Thinking Path

- WebUI steer is still correct for in-process runs because the cached `AIAgent` can receive `agent.steer(text)` directly.
- Gateway-owned runs are a different topology: WebUI has a live stream and an `ACTIVE_RUNS` row, but no process-local agent in `SESSION_AGENT_CACHE`, so the old cache-miss fallback reports the wrong failure.
- The gateway run-control API is not a dependable WebUI steer relay yet, so the safe WebUI behavior is to preserve the submitted text by queueing it as the next turn on the captured owner session.

## What Changed

- `api/streaming.py`: returns `gateway_steer_queued` when a no-cache steer belongs to a live `ACTIVE_RUNS` gateway stream.
- `static/commands.js`: consumes that fallback by queueing the submitted text and staged files on the captured owner session without cancelling or clearing the active stream.
- `tests/test_real_steer.py`: covers the backend gateway fallback, the frontend owner-scoped queue path, and the adjacent no-cache recovery path.
- `tests/test_issue4749_steer_reason_and_recovery.py`: includes the new backend fallback in the steer failure taxonomy.

## Why It Matters

Gateway-backed WebUI sessions no longer lose busy-turn input behind a misleading local cache miss. When live steering is unavailable from WebUI, the message is kept as the next queued turn instead.

## Verification

Focused tests cover the gateway no-cache steer fallback, owner-scoped frontend queueing, adjacent steer recovery behavior, and fallback taxonomy; JS syntax and runtime-guard lint pass locally.

## Upstream

Closes #5585.

## Model Used

GPT-5 via Codex CLI

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -10057,7 +10057,29 @@ def _handle_chat_steer(handler, body: dict) -> bool:
         except Exception:
             logger.debug("Failed to close steer identity-mismatched cached agent for session %s", sid, exc_info=True)
     if not cached:
-        # No active agent for this session — caller surfaces a steer failure
+        try:
+            s = get_session(sid)
+            active_stream_id = getattr(s, "active_stream_id", None) or None
+        except KeyError:
+            active_stream_id = None
+        if active_stream_id:
+            with _cfg.STREAMS_LOCK:
+                stream_alive = active_stream_id in _cfg.STREAMS
+            if stream_alive:
+                try:
+                    with _cfg.ACTIVE_RUNS_LOCK:
+                        active_run = dict((_cfg.ACTIVE_RUNS or {}).get(str(active_stream_id)) or {})
+                    if active_run.get("backend") == "gateway":
+                        return j(handler, {"accepted": False, "fallback": "gateway_steer_queued",
+                                           "stream_id": active_stream_id})
+                except Exception:
+                    logger.warning(
+                        "Gateway ownership lookup failed before steer fallback for session=%s stream_id=%s",
+                        sid,
+                        active_stream_id,
+                        exc_info=True,
+                    )
+        # No active local agent for this session — caller surfaces a steer failure
         # without cancelling the active run.
         return j(handler, {"accepted": False, "fallback": "no_cached_agent",
                            "stream_id": None})
--- a/static/commands.js
+++ b/static/commands.js
@@ -1350,6 +1350,7 @@ async function cmdSteer(args){
 }
 
 function _steerFailureMessageKey(fallback) {
+  if(fallback==='gateway_steer_queued')return 'steer_fail_no_cached_agent';
   const key = 'steer_fail_' + (fallback || 'unknown');
   return (typeof LOCALES !== 'undefined' && LOCALES.en && LOCALES.en[key])
     ? key : 'steer_fail_unknown';
@@ -1546,6 +1547,10 @@ async function _trySteer(msg, explicitSteer){
   const ownerSid=(typeof S!=='undefined'&&S.session&&S.session.session_id)||null;
   const ownerStreamId=(typeof S!=='undefined'&&(S.activeStreamId||(S.session&&S.session.active_stream_id)))||null;
   const pendingFilesSnapshot=typeof S!=='undefined'&&Array.isArray(S.pendingFiles)?[...S.pendingFiles]:[];
+  const ownerProfile=typeof S!=='undefined'&&(S.activeProfile||'default');
+  const ownerModelState=typeof _chatPayloadModelState==='function'
+    ? _chatPayloadModelState()
+    : {model:(typeof S!=='undefined'&&S.session&&S.session.model)||'',model_provider:(typeof S!=='undefined'&&S.session&&S.session.model_provider)||''};
   if(!ownerSid){showToast(t('no_active_session'));return false;}
   let steerText=originalMsg;
   try{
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Reading the diff at HEAD against `origin/master` — `api/streaming.py:10059-10084`, `static/commands.js:1617-1638`, and the four test files — plus the surrounding steer machinery on master, this is a well-scoped fallback and the contracts line up.

## What I verified

**Backend gateway detection is correctly gated and lock-safe.** The new block runs only after `SESSION_AGENT_CACHE_LOCK` is released (`streaming.py:10058`), then acquires `STREAMS_LOCK` and `ACTIVE_RUNS_LOCK` sequentially rather than nested, so there's no new lock-ordering hazard:

```python
if active_stream_id:
    with _cfg.STREAMS_LOCK:
        stream_alive = active_stream_id in _cfg.STREAMS
    if stream_alive:
        with _cfg.ACTIVE_RUNS_LOCK:
            active_run = dict((_cfg.ACTIVE_RUNS or {}).get(str(active_stream_id)) or {})
        if active_run.get("backend") == "gateway":
            return j(handler, {"accepted": False, "fallback": "gateway_steer_queued", ...})
```

The `backend == "gateway"` marker this keys on is real: `register_active_run(...)` for the gateway worker sets `backend="gateway"` at `api/gateway_chat.py:610`, whereas the in-process worker at `api/streaming.py:6408` sets `ephemeral=...` and no `backend` key. So the discriminator is sound — a gateway-owned live run is distinguishable from an in-process one.

**Frontend contract matches.** `queueSessionMessage(sid, payload)` at `static/ui.js:250` just spreads `payload` and stamps `_queued_at`, so the `{text, files, model, model_provider, profile}` shape the handler passes is fine. `_chatPayloadModelState()` (`static/messages.js:108`) returns `{model, model_provider}`, matching the snapshot destructure. The `steer_leftover_queued` key exists across all locale bundles (`i18n.js:301` et al).

**Graceful degradation is covered.** If `queueSessionMessage` is unavailable, the gateway block is skipped and control falls through to the generic failure path, where `_steerFailureMessageKey('gateway_steer_queued')` now maps to `steer_fail_no_cached_agent` (`commands.js:1350`) — the user still sees the original toast rather than an `unknown` key. `test_issue4749...py:test_reason_map_contract` pins exactly that mapping, and `test_backend_parity` was correctly widened via `HANDLED_NON_RECOVERY_CODES`.

## One scoping question on "Closes #5585"

#5585 described the symptom as consistent ("whenever trying to submit while streaming") and enumerated three in-process root causes — insert-timing race, identity/`_agent_sig` eviction, and agent-never-cached. This PR resolves a fourth topology the issue didn't name: **gateway-backend sessions never populate `SESSION_AGENT_CACHE` at all**, so steer misses 100% of the time — which actually fits the "consistent whenever" repro better than the intermittent race hypotheses.

The gap: this only reaches the reporter if they're running `HERMES_WEBUI_CHAT_BACKEND=gateway`. If they're on the default in-process backend, a genuine live in-process stream still falls through to `no_cached_agent` (the identity-eviction branch at `streaming.py:10046` or a race). Worth confirming the reporter's backend before closing, or scoping the close to "gateway topology" and leaving the in-process race as a follow-up. Behaviorally the fallback also queues-as-next-turn rather than truly steering the in-flight gateway run, which the PR body is upfront about; the toast copy communicates that to the user.

Net: the code and tests are correct for what they target. The only open item is whether #5585's actual reporter is on gateway mode.

nesquena-hermes:
Shipped in **exp-v0.52.8** — thanks @rodboev! Gateway-owned steer now queues your message as the next turn instead of erroring + dropping it. Codex SAFE + Fable SHIP-UX, byte-verified vs your head, 66 tests green, maintainer visual-approved (desktop + mobile), deployed to experimental. 🚀

PR #5826, opened 2026-07-09
Claude/hermes webui custom UI thfqqc

Toqsick:
update offical


nesquena-hermes:
Read the full diff against `origin/master` in a read-only worktree — all 9 commits (`e135e7501` skin allowlist → `ae16d4387` stale-render perf fix), plus the surrounding code in `static/panels.js`, `static/ui.js`, `api/config.py` on both branches. This is a clean, well-tested change. A few notes confirming why, plus the one cosmetic nit worth folding in before it lands.

## The skin-allowlist commit is a genuine standalone bugfix

`e135e7501` isn't just prep for the yuno skins — it repairs a real pre-existing regression. On `origin/master`, `_SETTINGS_SKIN_VALUES` (config.py:8483) is missing five skins that already ship in the boot.js `_SKINS` picker: `github`, `codex`, `terracotta`, `hepburn`, `neon`. That set is the gate in `_normalize_appearance`:

```python
# api/config.py:8465-8469
next_skin = (
    raw_skin
    if raw_skin in _SETTINGS_SKIN_VALUES
    else legacy_skin
)
```

`_normalize_appearance` runs on every settings save (config.py:8571 and :8829), so a user who picked GitHub/Codex/Terracotta/Hepburn/Neon would have it silently revert to `default` on the next save/reload — the exact symptom the PR body describes. The new `test_skin_registry_parity.py` pins all three registries (boot.js `_SKINS`, the index.html no-flash allowlist, `_SETTINGS_SKIN_VALUES`) in lockstep, which is the right guard so this class of drift can't recur.

## The stale-render perf guard is correctly placed

`ae16d4387` adds one line after the 7-endpoint `Promise.all` in `loadDashboard`:

```js
// static/panels.js:4441
if (_currentPanel !== 'dashboard') return;
```

This genuinely mirrors the existing bail-outs (logs at :4333, kanban at :2792/:2827). Two things I verified make it correct rather than cosmetic: the guard sits *before* both `_renderDashboard(box, …)` and `_dashboardStartPolling()` (:4448), and the poll start is idempotent-but-leaky — `switchPanel`'s leave-cleanup (`prevPanel === 'dashboard' && nextPanel !== 'dashboard'` → `_dashboardStopPolling()`, panels.js:397-399) fires *while the fetch is still in flight*, so without this guard a late `_dashboardStartPolling()` would re-arm `_dashboardPollTimer` after cleanup already ran and leak an interval past the panel switch. The guard closes that. `_dashboardPollTick` also independently guards with `_dashboardIsVisible()` on both sides of its await, so the polling path is doubly safe.

## i18n coverage is complete

All 16 new `dashboard_*` keys are present in every one of the 15 locale blocks (`dashboard_kpi_memory`, `dashboard_open_external`, etc. each return 15 matches). No missing-key fallbacks, and the relocated external-dashboard link still works — `_applyDashboardStatus` (ui.js:1774) toggles `display` on `[data-dashboard-link]` by the probe's `running` flag, and the button just moved into the `mainDashboard` header (index.html:919-923) while keeping the `data-dashboard-link` hook, so the probe-gated visibility is preserved.

## One nit worth folding in (matches Greptile's finding)

The panel has two different display strings for the same view. The main-view header uses `dashboard_title` (`"Dashboard"`), but the sidebar panel-head and `APP_TITLEBAR_KEYS` still use `tab_dashboard` (`"Hermes Dashboard"`):

```js
// static/panels.js:45
dashboard: 'tab_dashboard',
```
```html
<!-- static/index.html:293 -->
<span data-i18n="tab_dashboard">Dashboard</span>
```

Since `tab_dashboard: 'Hermes Dashboard'` vs `dashboard_title: 'Dashboard'` (i18n.js:1015-1016), the titlebar and sidebar head will read "Hermes Dashboard" while the main view reads "Dashboard". Not a blocker, but pointing `APP_TITLEBAR_KEYS.dashboard` and the `panelDashboard` head at `dashboard_title` (or vice-versa) would make the three surfaces agree. Your call which string wins.

Otherwise this is solid — the DOM isolation from the insights health monitor (own `#dashboard*` IDs), the polling lifecycle, and the registry-parity test all follow established patterns. Deferring the merge decision to a maintainer per policy; I don't execute PR-authored tests.

nesquena-hermes:
**Certified head:** `sha:ae16d4387879` · **PR:** #5826 · _@Toqsick, Claude/hermes webui custom UI thfqqc_
**Verdict:** The skins and dashboard render cleanly, but three independently reproduced correctness/security defects plus the UX gate block this head. Fix, split the bundle, then re-warm/re-gate; do not merge.

### What I ran (isolated rebased worktree `/tmp/wt-gate-5826`, port 8783)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SHIP ONLY WITH FIXES** — found the stale async-load race and wrong disabled-cron count; both reproduced independently below. |
| Fable UX | **REQUEST_CHANGES** — inconsistent Dashboard naming, ambiguous tab-visibility chips, and a corner status badge that reads as a notification. |
| Opus senior review | **APPROVE** with minor improvements; no blocker found. Reproduced-defect-wins applies. |
| Full pytest suite | **13,391 passed, 0 failed**, 10 skipped, 1 xfailed, 2 xpassed, 35 subtests passed; ran to completion with `-p no:xdist`. |
| Private/static gates | Threat scan CLEAN (score 0); range-diff/rebase preservation clean; ruff-forward, scope-undef, ESLint runtime, JS syntax, and Python compile gates clean. |
| Browser / live drive | Real app on a seeded server at 1440×900 and 390×844; drove Dashboard KPI navigation, refresh/poll teardown, memory search + HTML-escaping, settings/skin selection, mobile drawer, and rapid panel-switch races. Clean-master before shots + PR after shots were vision-verified and sent to the maintainer on Telegram. |

### Findings
- ✅ **The visible implementation is real and mostly coherent.** The in-app Dashboard populates from live API data; all six KPI buttons navigate; refresh works; leaving the panel clears polling; Yuno, Yuno Cyberpunk, and Yuno HC apply correctly. The 390×844 Dashboard has no horizontal overflow or clipped cards. Memory-search match rendering escapes HTML correctly.
- ⛔ **[CORE: stale async visit/profile data] `static/panels.js` (`loadDashboard`) has no visit/profile generation guard.** Reproduction: defer Dashboard visit A's `/api/memory`, leave, enter Dashboard visit B and render marker `B`, then resolve A. The visible/current Dashboard regressed from `B` to stale marker `A`. The existing `dashboardContent`/visibility checks only prove some Dashboard visit is active, not the visit/profile that launched the request. A profile switch can therefore paint the prior profile's data into the new profile's Dashboard. **Fix:** capture a monotonically increasing dashboard-load epoch plus active-profile identity; advance it on enter, leave, and profile switch; commit `_dashMemoryData`/DOM only when both still match. Add a deferred-A/fast-B test and a profile-switch variant.
- ⛔ **[CORRECTNESS] Disabled cron jobs are counted as active.** `_renderDashboard` uses `jobs.length` for both numerator and denominator. A real two-row fixture with one `enabled:true` and one `enabled:false` renders **`2 / 2`**, expected **`1 / 2`**. **Fix:** count enabled jobs for the numerator (`enabled !== false`, matching existing cron semantics) and keep total jobs as the denominator; add enabled/disabled/missing-field coverage.
- ⛔ **[DOM/script injection] Recent-completion `session_id` is inserted into an inline `onclick`.** `esc()` is HTML escaping, not JavaScript-string escaping: the HTML parser decodes `&#39;` before compiling the handler. A synthetic completion with `session_id="');window.__gate5826=1;//"` produced `onclick="switchPanel('chat');loadSession('');window.__gate5826=1;//')"`; clicking it executed the injected assignment. Even if current scheduler IDs are normally UUID-like, new code must not turn stored API data into executable source. **Fix:** remove the dynamic inline handler; put the ID in a `data-session-id` property and use an event listener/delegation (or assign the listener directly). Add quote/backslash/newline/adversarial-ID tests.
- ⛔ **[UX gate] The new surface needs one semantic pass.** Use one name everywhere (`Dashboard` vs `Hermes Dashboard` currently differs between titlebar/sidebar/main heading); make visibility chips express a single state instead of showing a checkmark and `Show` badge together; move the external-dashboard running indicator inside the button instead of a clipped corner badge. Keep the strong desktop/mobile card layout and skin contrast.
- ⚠️ **Concept/scope dependency remains.** This is still a large skins + Dashboard + i18n bundle with an uninformative PR body. Even after engineering convergence, the new top-level panel and Yuno curation need @nesquena's product call; splitting skins from Dashboard/i18n would make the decision and review substantially safer.

### Screenshots
Vision-verified clean-master before and PR after evidence was sent to the maintainer on Telegram: desktop Insights baseline; desktop Yuno Dashboard; 390×844 Yuno Cyberpunk Dashboard; mobile Yuno HC drawer; Appearance baseline; Appearance with all three Yuno variants.

**RED — @Toqsick should fix the stale-load/profile guard, active-cron numerator, dynamic inline-handler injection, and the three UX semantics; split the multi-concern bundle, then re-warm and run a full re-gate. Do not merge this head.**

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy — that is the release agent's call. This cert is valid only while the head stays at sha:ae16d4387879._

nesquena-hermes (changes requested):
Re-gated the current head `17ca56a9a5ab` against the RED gate certificate on `ae16d4387879`. The one new commit changes only the recent-completion inline handler. It does not resolve the other two engineering blockers, and its handler change is still attribute-injectable.

### Still required

1. **Add a visit/profile generation guard to `loadDashboard()`.** The current head still has only:
   ```js
   if (_currentPanel !== 'dashboard') return;
   ```
   That proves a Dashboard is visible, not that this is the same visit/profile that started the seven requests. A deferred visit A can still overwrite fast visit B, including after a profile switch. Capture a monotonically increasing load epoch plus active-profile identity, advance it on enter/leave/profile switch, and commit `_dashMemoryData` and DOM only when both still match. Add deferred-A/fast-B and profile-switch tests.

2. **Count `enabled: false` cron jobs as inactive.** Current code remains:
   ```js
   const activeJobs = jobs.filter(jb => jb && !jb.paused && !jb.disabled).length;
   ```
   Cron payloads use `enabled`; `{enabled:false}` has neither `paused` nor `disabled`, so it is still counted. Use the existing cron semantic (`enabled !== false`) for the numerator and retain `jobs.length` as the denominator. Add enabled/disabled/missing-field coverage.

3. **Remove the dynamic inline handler.** `JSON.stringify(sid)` makes a JavaScript string literal, but it does not make that literal safe inside a single-quoted HTML attribute. For example, current-head construction with:
   ```text
   session_id = x' onmouseover='window.__gate5826=1
   ```
   parses as two event attributes:
   ```text
   onclick     = switchPanel("chat");loadSession("x
   onmouseover = window.__gate5826=1")
   ```
   Put the escaped ID in `data-session-id` and attach a real event listener/delegated listener instead of generating executable source. Add a DOM/runtime test covering quotes, backslashes, newlines, and an adversarial event-attribute payload.

The product/UX gate and split recommendation from the prior certificate also remain for maintainer review after these objective defects converge.

Verification on this head: threat scan CLEAN; sandboxed `tests/test_dashboard_panel_view.py` is green (9 passed), but that file has no tests for the three cases above.

Issue #572, opened 2026-04-16
Web UI onboarding can overwrite existing provider/model defaults


franksong2702:
## Summary
On existing Hermes Web UI installs, the onboarding flow can silently re-run and overwrite an already working provider/model configuration. In our case, a working `minimax-cn / MiniMax-M2.7` setup was later rewritten to `openrouter / anthropic/claude-sonnet-4.6` after a restart / setup flow.

## Trigger
This appears to happen when the Web UI startup path decides onboarding is still incomplete and opens the onboarding wizard on a machine that already has a valid Hermes config.

In the onboarding quick setup flow, the provider list is limited to a small hardcoded set (for example `openrouter`, `anthropic`, `openai`, `custom`). If the existing provider is not in that list (for example `minimax-cn`), the UI falls back to the first provider in the list, which is `openrouter`.

## What happens
When the final onboarding step is submitted, the wizard writes the selected provider/model back to `config.yaml` and can also update `.env` in the same save path. That means an existing working config can be silently replaced with a different provider/model pair.

In practice, this caused:
- the default provider to switch from `minimax-cn` to `openrouter`
- the default model to switch from `MiniMax-M2.7` to `anthropic/claude-sonnet-4.6`
- confusing model/profile drift after restart

## Suggested fix
The onboarding flow should not overwrite an existing, working provider/model by falling back to `openrouter`.

Possible fixes:
- derive provider options from the full runtime/provider registry instead of a hardcoded quick-setup list
- preserve the existing provider/model when onboarding is re-entered on an already configured install
- only write missing secrets or truly missing fields during onboarding, not the whole model/provider block
- if a provider is unsupported by quick setup, show a non-destructive warning instead of auto-falling back and saving a different provider

## Notes
The issue is not the restart itself. The restart appears to expose a save/write that already happened earlier. We saw `config.yaml` and `.env` change at the same second, which points to a setup/save path rather than a plain bootstrap.

nesquena-hermes:
Investigated and reproduced. Root cause confirmed: `_provider_api_key_present()` only knows the 4 quick-setup providers — any non-standard provider (minimax-cn, deepseek, xai, gemini, etc.) always returns `chat_ready=False`, causing the wizard to fire on every load and leaving the door open for `config.yaml` to be overwritten.

Fix is in PR #575. Three-layer patch:
1. `_provider_api_key_present` now falls back to `hermes_cli.auth.get_auth_status()` for any API-key provider not in the quick-setup list
[...]

PR #3974, opened 2026-06-11
feat(providers): register Tencent Cloud (TokenHub) in WebUI catalog

yixiusztx123:
## Thinking Path

- Hermes WebUI aims for near 1:1 parity with the Hermes CLI/agent in the browser, including which providers you can pick.
- The Hermes agent already registers a `tencent-tokenhub` provider — an OpenAI-compatible gateway at `https://tokenhub.tencentmaas.com/v1` that fronts Hunyuan plus proxied DeepSeek / GLM / Kimi / MiniMax models — and authenticates it with `TOKENHUB_API_KEY`.
- But the WebUI has no catalog entry for it, so even with a valid key the provider never shows up in Settings → Providers or the model picker. The only Tencent surface today is the `tencent/hy3-preview:free` entry under OpenRouter, which is unrelated to TokenHub.
- This PR registers the provider in the WebUI catalog under the key `tencent`, mapping the agent's `tencent-tokenhub` slug onto it so detection, the picker, and the Settings card all resolve.

## What Changed

- **`api/config.py`**
  - `_PROVIDER_DISPLAY`: add `tencent` → "Tencent Cloud".
  - `_PROVIDER_ALIASES`: map the agent slug `tencent-tokenhub` (and user-facing variants `tencent-cloud`, `tencentcloud`, `hunyuan`, `tencent-hunyuan`) onto `tencent`.
  - `_PROVIDER_MODELS`: add a curated DeepSeek / GLM / Kimi / MiniMax list. The agent's live catalog for this provider is sparse, so this static list is the practical fallback the picker shows (same pattern as other direct-API providers).
  - `get_available_models()`: env-var fallback detection accepts either the WebUI-friendly `TENCENT_API_KEY` or the agent's canonical `TOKENHUB_API_KEY`.
- **`api/providers.py`**
  - `_PROVIDER_ENV_VAR`: add `tencent` → `TENCENT_API_KEY`.
  - `_PROVIDER_ENV_VAR_ALIASES`: add `tencent` → `TOKENHUB_API_KEY` so detection and runtime key lookup work under either name.
  - `remove_provider_key()`: also clear aliased env vars on removal, so the provider card doesn't linger as "configured" via the alias path. (Also benefits existing aliased providers like `lmstudio` / `opencode-*`.)
- **`api/onboarding.py`**
[...]

Issue #1020, opened 2026-04-25
bug(onboarding): wizard fires for users with non-wizard providers (ollama, deepseek, xai) even when config.yaml is fully configured


nesquena-hermes:
## Summary

When hermes-agent is installed with a provider not supported by the wizard (e.g. `ollama-cloud`, `kimi`, `deepseek`, or any provider not in `_SUPPORTED_PROVIDER_SETUPS`), the onboarding wizard fires on every fresh WebUI clone even though `config.yaml` is fully configured. The wizard has no way to represent the existing provider, so completing it overwrites the user's config with whichever wizard-supported provider they select.

Reported by a user running `ollama-cloud` with `kimi-k2.6`.

## Steps to reproduce

1. Configure hermes-agent with `provider: ollama-cloud` (or any non-wizard provider) in `config.yaml`.
2. Clone and start the WebUI for the first time (no prior `onboarding_completed` in `STATE_DIR/settings.json`).
3. The wizard fires instead of going straight to chat.
4. The wizard shows only OpenRouter / Anthropic / OpenAI / Custom — no `ollama-cloud` option.
5. Completing the wizard overwrites `config.yaml` with the chosen wizard provider.

## Root cause

`get_onboarding_status()` in `api/onboarding.py` auto-completes onboarding only when `config_auto_completed = config_exists AND chat_ready` (line 436).

`chat_ready` requires `provider_ready`, which for `ollama-cloud` checks for an API key or OAuth token (lines 309–355 in `_status_from_runtime()`). Ollama is a self-hosted service with no API key, so `provider_ready = False` → `chat_ready = False` → the auto-complete guard never fires → wizard shows.

The `config_exists` flag is already computed and even surfaced in the status response, but the "skip if already configured" path is blocked by the `chat_ready` requirement.

## Expected behavior

If `config.yaml` exists **and** the provider is a type the wizard cannot configure (i.e. not in `_SUPPORTED_PROVIDER_SETUPS` and not an OAuth flow), the wizard should auto-skip with a "you're already configured" screen rather than blocking chat.

The wizard should never overwrite an existing `config.yaml` for an unsupported provider — it already has a guard (`"error": "config_exists"` at line 523) but that guard only activates when the user tries to submit, not at wizard entry.

## Proposed fix

In `get_onboarding_status()`, relax the auto-complete condition for unsupported providers:

```python
current_provider = _extract_current_provider(cfg)
[...]

PR #5827, opened 2026-07-09
Expose MiniMax global and China onboarding setups

octo-patch:
## Problem

Users currently cannot choose MiniMax global or China during onboarding. The existing provider and model registry already contains MiniMax models, but it does not provide configurable provider setup entries for those onboarding choices.

## Change

- Add separate `minimax` (global) and `minimax-cn` (China) entries to `api/onboarding.py` in `_SUPPORTED_PROVIDER_SETUPS`.
- The existing onboarding UI now renders those entries as separate provider choices. Each choice reuses its API-key environment variable, MiniMax model list, and regional OpenAI-compatible base URL: `https://api.minimax.io/v1` for global and `https://api.minimaxi.com/v1` for China. The existing setup flow writes the selected provider and its regional base URL.
- Update `api/routes.py` so live model discovery uses the matching regional `/v1/models` endpoint for both choices.
- Update `tests/test_issue572.py` so the original non-wizard protection remains covered for providers outside the setup registry, while an incomplete MiniMax China setup stays visible and is not treated as OAuth-only.
- Anthropic endpoint selection is not exposed by this onboarding flow. This change deliberately adds no Anthropic base URL field because the existing catalog and setup handlers have no consumer for it; existing Anthropic routing is unchanged.

## Compatibility

Existing MiniMax provider names, model entries, API-key environment variables, and default OpenAI-compatible URLs are reused. No configuration migration is required, and ready MiniMax configurations keep their existing behavior. Because MiniMax China is now a supported onboarding choice, a ready configuration completes onboarding while a missing key keeps the setup step visible so the user can provide it. Providers that remain outside the setup registry retain the issue #572 skip behavior; the updated regression fixture preserves that protection.

## Tests

Before this follow-up, the PR failed:

- `./scripts/test.sh tests/test_issue572.py -q` -> 2 failed, 10 passed: `TestOnboardingStatusUnsupportedProvider::test_minimax_cn_not_ready_skips_wizard` and `TestOnboardingStatusUnsupportedProvider::test_current_is_oauth_set_for_unsupported_provider`.

After this follow-up:

- `./scripts/test.sh tests/test_issue572.py` -> 15 passed.
- `./scripts/test.sh tests/test_minimax_provider.py tests/test_onboarding_mvp.py tests/test_onboarding_existing_config.py tests/test_onboarding_static.py tests/test_issue603_provider_categories.py tests/test_issue1499_keyless_onboarding.py tests/test_sprint39.py tests/test_byok_model_dropdown.py tests/test_4413_seed_provider_models.py` -> 135 passed.
- `python3 -m compileall -q api server.py bootstrap.py mcp_server.py tests scripts` -> passed.
- `python3 scripts/ruff_lint.py --diff HEAD^` -> passed with no new violations.


nesquena-hermes (changes requested):
Thanks for this. MiniMax is a core builtin provider (confirmed via `core_supports.py` — provider + tts), so surfacing it more is reasonable in principle. But the gate surfaced a real conflict: this PR **breaks two existing tests that guard a deliberate onboarding-UX decision** (#1020), so it needs a maintainer product call before it can land — it's not a stale-test update.

**What breaks + why it matters:**
- `tests/test_issue572.py::test_minimax_cn_not_ready_skips_wizard` and `::test_current_is_oauth_set_for_unsupported_provider` fail on this branch (pass on master; CI shows 6 red).
- Those tests encode an **intentional** design from #1020: MiniMax-CN (and other non-wizard providers) were **deliberately excluded from `_SUPPORTED_PROVIDER_SETUPS`** so that (a) onboarding auto-completes when `provider_configured=True` (users don't get the first-run wizard for a provider it can't properly set up), and (b) users fix credentials via **Settings → Providers**, not the wizard. The documented rationale: *"showing it would only confuse the user or let them accidentally overwrite their config with an OpenAI/Anthropic provider."*
- This PR adds MiniMax + MiniMax-CN to the wizard registry, which **reverses that decision** — hence the test breaks. If a MiniMax-CN user now gets the wizard, they could accidentally overwrite their working config, which is exactly what #1020 was preventing.

**So this is two things, neither mechanical:**
1. **Product/curation call (maintainer):** should MiniMax / MiniMax-CN be a first-run **wizard** provider (with quick-setup UX), reversing #1020's intentional exclusion? Being a builtin provider (usable via Settings → Providers) is already true; adding it to the *wizard* is the new behavior. If yes, the #1020 tests should be updated to reflect the new intent (and confirm the wizard flow actually sets MiniMax up correctly, incl. the global-vs-China + OpenAI-vs-Anthropic endpoint selection — the wizard needs to handle that fork, or it hits the "overwrite with wrong provider" footgun #1020 warned about). If the answer is "MiniMax stays Settings-only," then this PR isn't needed.
2. **If it proceeds:** update `test_issue572.py` to assert the new supported-in-wizard behavior (don't just delete the assertions — re-express them for the new intent), and add a test that the MiniMax-CN wizard path configures the correct endpoint without clobbering an existing config.

Flagging to the maintainer for the #1020 reversal decision. The code itself is small and clean (+48, one file); the question is purely whether we want MiniMax in the first-run wizard. I'll hold here pending that call.

octo-patch:
Follow-up in `07a1187f`:

- MiniMax global and China remain separate onboarding choices backed by the existing provider setup flow and their regional OpenAI-compatible base URLs.
- The unused `endpoint_options` and `default_anthropic_base_url` fields were removed because this onboarding path neither serializes nor consumes them. Anthropic endpoint selection is intentionally unchanged.
- The issue #572 regression now uses a provider outside the setup registry, while MiniMax-CN coverage verifies that a missing key keeps onboarding visible and does not mark it as OAuth-only.
- Live model discovery now uses the matching global and China `/v1/models` endpoints, with regression coverage for both URLs.

The local onboarding and model-discovery tests, lint, browser smoke, and the full GitHub test matrix pass on this commit. Please re-review the updated branch.

nesquena-hermes:
**Certified head:** `sha:07a1187f5933` · **PR:** #5827 · _@octo-patch, Expose MiniMax global and China onboarding setups_
**Verdict:** The registry entries, regional URL split, env vars, and #1020 test re-expression are sound, and the full suite is green — but BOTH advisors independently reproduced head-introduced defects: the wizard writes the wrong transport for MiniMax, and the "key repair" flow silently no-ops. Plus the #1020 wizard-reversal itself is a Nathan product call. Fix + get the concept sign-off; do not merge.

### What I ran (isolated worktree `/tmp/wt-gate-5827`, patch-id-clean rebase `18761d81` on current `origin/master`; was 530 behind)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SHIP ONLY WITH FIXES** — env-injected key regression + confirm_overwrite dead-end, both reproduced. |
| Opus (senior review) | **SHIP ONLY WITH FIXES / KICK BACK** — reproduced the base_url transport flip + the same dead-end, independently. |
| Full pytest suite | **13,484 passed, 0 failed** (10 skipped, 1 xfailed, 2 xpassed, 34 subtests); the prior CI-red shards that bounced the old head are resolved. |
| Static/threat | Threat scan CLEAN; `core_supports minimax` → IS a core builtin (parity, Anchor-A per #3974). Rebase patch-id byte-identical. |
| My own reproduction | Confirmed the env-key regression's root cause at source + `_provider_api_key_present` layer. |

### Findings
- ✅ **The core mechanics are sound.** #1020 tests were RE-EXPRESSED not deleted (the unsupported-provider auto-complete guarantee is preserved via genuinely-unsupported `ollama-cloud`; new MiniMax-CN-in-wizard tests added); no alias collapses `minimax-cn`→`minimax`; env var names (`MINIMAX_API_KEY`/`MINIMAX_CN_API_KEY`) match the agent auth registry; the no-clobber guard refuses to overwrite a foreign config without `confirm_overwrite`; the global URL fix (`api.minimax.chat`→`api.minimax.io`) corrects a defunct domain.
- ⛔ **[CORE — head-introduced transport flip, Opus reproduced] `api/onboarding.py:184/193`.** The new entries set `default_base_url` to the OpenAI-compat `/v1` URLs, which `apply_onboarding_setup` writes into `config.yaml model.base_url` → the runtime resolves `chat_completions @ .../v1`. But the agent registers key-based MiniMax with **`anthropic_messages` transport at the `/anthropic` endpoints** (`hermes_cli/auth.py:293-300/337-344`). Opus reproduced with the agent's own resolver: wizard config → `chat_completions @ /v1`; CLI config → `anthropic_messages @ /anthropic`. So a wizard-onboarded MiniMax user gets a different transport than every other MiniMax path, and a key-repair silently rewrites an existing CLI config's endpoint — the exact OpenAI-vs-Anthropic fork the maintainer flagged in #1020. **Fix (small):** drop `default_base_url` from both entries (apply path then `pop`s base_url and the agent's per-provider registry default supplies the correct regional `/anthropic` URL; region survives via provider id). Keep the `/v1` URLs in `routes.py _OPENAI_COMPAT_ENDPOINTS` (that's only the `/models` listing).
- ⛔ **[CORE — env-key regression, Codex + Opus reproduced] `api/onboarding.py:619` + `:708`.** `_status_from_runtime` sources `env_values` from the profile `.env` FILE only (`_load_env_file`), and the `os.environ` auth-registry fallback (line 619) is gated on `provider not in _SUPPORTED_PROVIDER_SETUPS`. On master MiniMax wasn't supported → fell through to the fallback → an env-EXPORTED key was detected. The PR adds MiniMax to `_SUPPORTED_PROVIDER_SETUPS`, so shell-env-keyed MiniMax users (no `.env` file, no persisted flag) flip from detected to undetected → a one-time wizard they must Skip. **Provenance: PR-introduced** (I verified master's fallback comment lists minimax as a fallback provider). **Fix:** check the process/profile environment for supported providers too, or allow the auth-registry fallback for all non-OAuth providers.
- ⛔ **[frontend dead-end the PR newly depends on, both advisors] `static/onboarding.js:~503`.** The exact key-missing-existing-config population this PR routes into the wizard hits the `confirm_overwrite` guard (`onboarding.py:1005`, HTTP 200 + error dict); the client never sends `confirm_overwrite`, never checks `resp.error`, and `_finishOnboarding` calls `/api/onboarding/complete` → the user types their key, the wizard claims success, the key is never written, chat stays broken. Reachable on master (deepseek/openrouter) so not strictly head-introduced, but the PR's stated value ("missing key keeps the setup step visible so the user can provide it") doesn't function without a frontend `requires_confirm` handler. **Fix or explicitly waive in the concept decision.**
- ⚠️ **CONCEPT (Nathan) — the load-bearing dependency.** Adding MiniMax + MiniMax-CN to the first-run wizard REVERSES the intentional #1020 exclusion (non-wizard providers deliberately kept out to avoid the "overwrite working config with wrong provider" footgun). Whether MiniMax belongs in the first-run wizard is a maintainer product call — this parks for Nathan regardless of engineering.

**RED — parks for @nesquena's #1020 reversal product call, AND @octo-patch should fix the transport flip (drop `default_base_url` from both entries), the env-key detection regression (check `os.environ` for supported providers), and the `confirm_overwrite` frontend dead-end (or waive it). Re-gate from zero. Do not merge this head.** Engineering is close and the parity concept is in-scope (core builtin), but the wizard footgun #1020 warned about is real at this head. Preserve @octo-patch author credit.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy — that is the release agent's call. This cert is valid only while the head stays at sha:07a1187f5933._

octo-patch:
Thanks for the gate certification. The wizard scope still depends on the #1020 product decision, so I have not pushed additional changes while that decision is pending.

PR #5828, opened 2026-07-09
fix(server): frame OPTIONS preflight with Content-Length: 0

ai-ag2026:
## Summary
`Handler.do_OPTIONS` answers every CORS preflight with `send_response(200)` + CORS headers + `end_headers()` but never sets `Content-Length`.

## Why
The handler runs `protocol_version = "HTTP/1.1"` (keep-alive on), and its own class docstring states *"every response must declare framing."* An unframed `200` is *read-body-until-connection-close* (RFC 7230 §3.3.3), so a keep-alive client issuing a preflight — a browser CORS check for an allowlisted cross-origin front-end (`HERMES_WEBUI_ALLOWED_ORIGINS`), `curl`, a reverse proxy, a monitor — blocks reading a body that never arrives until the 30s connection timeout, and cannot reuse the connection.

This is the same framing gap the passkey-login 200 fix closed, on the OPTIONS path. The fix mirrors it: send `Content-Length: 0` before `end_headers()`. Status stays `200` (no behavior change for clients that already tolerated the gap), and the CORS-header policy is untouched.

## Validation
- `./scripts/test.sh tests/test_options_content_length.py tests/test_cors_preflight_allowlist.py tests/test_passkey_login_content_length.py -q` → 17 passed
- New regression `tests/test_options_content_length.py`: asserts the preflight 200 carries `Content-Length: 0`, that it is emitted before `end_headers()`, and that the framing is unconditional (also present for an allowlisted cross-origin preflight that emits CORS headers).

## Notes
Complementary to (not conflicting with) #5531's credentialed-CORS work — that PR changes which headers `apply_cors_preflight_headers` emits; this changes only the response framing in `do_OPTIONS`.

### DIFF
--- a/server.py
+++ b/server.py
@@ -439,6 +439,9 @@ class Handler(BaseHTTPRequestHandler):
         self._req_t0 = time.time()
         self.send_response(200)
         apply_cors_preflight_headers(self)
+        # Frame the empty preflight: without Content-Length an HTTP/1.1 keep-alive
+        # 200 is read-until-close, hanging the client until the 30s timeout.
+        self.send_header("Content-Length", "0")
         self.end_headers()
 
     def do_DELETE(self) -> None:
--- a/tests/test_options_content_length.py
+++ b/tests/test_options_content_length.py
@@ -1,0 +1,77 @@
+"""Regression test — the CORS preflight (OPTIONS) 200 must be framed.
+
+`Handler.do_OPTIONS` answered every preflight with `send_response(200)` +
+CORS headers + `end_headers()` but no `Content-Length`. The class runs
+`protocol_version = "HTTP/1.1"` (keep-alive on), and its own docstring states
+"every response must declare framing". An unframed 200 is "read body until the
+connection closes" (RFC 7230 §3.3.3), so a keep-alive client issuing a preflight
+(browser CORS for an allowlisted cross-origin front-end, curl, a proxy, a
+monitor) blocks reading a body that never arrives until the 30s connection
+timeout — and can't reuse the connection.
+
+This pins that the preflight now carries `Content-Length: 0`, sent before
+`end_headers()` (so it lands among the emitted headers), mirroring the framing
+every other bodyless response in the server already sets. The CORS-header policy
+itself (which Origins are echoed) is unchanged and covered separately.
+"""
+class _FakeHeaders(dict):
+    def get(self, key, default=None):
+        return dict.get(self, key, default)
+
+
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Reading `server.py:437-442` on `origin/master` against the diff at this PR's HEAD, the change is correct and consistent with the framing invariant the class already documents.

## Summary
`do_OPTIONS` on master emits `send_response(200)` + `apply_cors_preflight_headers()` + `end_headers()` with no body-framing header. Since the handler runs `protocol_version = "HTTP/1.1"` (`server.py:296`, keep-alive on), an unframed 200 falls under RFC 7230 §3.3.3 "read body until connection close", so a keep-alive preflight client blocks until the connection timeout. The one-line `Content-Length: 0` closes that gap.

## Code reference
Master `do_OPTIONS` before the fix:
```python
def do_OPTIONS(self) -> None:
    """Handle CORS preflight requests (headers emitted by api.routes)."""
    self._req_t0 = time.time()
    self.send_response(200)
    apply_cors_preflight_headers(self)
    self.end_headers()
```
The fix inserts `self.send_header("Content-Length", "0")` before `end_headers()`, which is the exact pattern already used elsewhere for bodyless responses: `_send_no_content` (`api/routes.py:5597-5598`), the OIDC 302 (`api/routes.py:11364`), and the overflow reject (`server.py:127`).

## Verification
- No duplicate-header risk: `apply_cors_preflight_headers` (`api/routes.py:5144-5163`) only emits `Access-Control-*` / `Vary` and returns early for disallowed origins — it never sets `Content-Length`.
- Header ordering is safe: the `end_headers()` override (`server.py:330-335`) appends CSP/Report-To *after* `Content-Length: 0`, all before `super().end_headers()`, so the framing lands among the emitted headers.
- The new `tests/test_options_content_length.py` pins the three properties that matter: value is `0`, it precedes `end_headers()`, and it's unconditional even on the allowlisted-cross-origin branch (via monkeypatching `_check_same_origin_browser_request`).

No behavioral change to the CORS policy itself — status stays 200, header echo policy untouched. This mirrors the earlier passkey-login framing fix (`967fef116`, `b10116bff`) on the OPTIONS path. Looks good to merge.

nesquena-hermes:
Shipped in **exp-v0.52.7** — thanks @ai-ag2026! Verified byte-identical to your PR head, Codex SAFE-TO-SHIP, full regression + your own tests green, deployed to the experimental channel. 🚀

PR #5829, opened 2026-07-09
fix(settings): write settings.json atomically to survive mid-write crashes

ai-ag2026:
## Summary
`save_settings` (and the startup default-workspace rewrite) persisted `settings.json` via a plain `Path.write_text`, which truncates the file in place.

## Why
A crash / full disk / power loss between the truncate and the full write leaves `settings.json` truncated or empty. The next start then loses **every** persisted setting — theme, workspace, tab order, and the login `password_hash`. (Losing the hash also silently drops back to no-auth, but the data-loss regression is the point here; the fail-open angle is tracked separately.)

## Fix
Both writes now go through a new `_atomic_write_settings_text`: write to a sibling temp file, `flush` + `os.fsync`, then `os.replace` into place, with a `finally` temp-cleanup. Because `os.replace` is atomic, a failure anywhere before the rename leaves the original file byte-for-byte intact. Mirrors the tempfile+fsync+os.replace pattern already used by `webui_session_db.WebUIJsonSessionDB._atomic_write`.

Two behavior-neutrality details:
- **Mode preserved** — `os.replace` swaps in the temp inode (a plain `open` respects the umask, ~0644), so the existing file's mode is copied onto the replacement. Without it, an operator-hardened `settings.json` (chmod 0600 because it holds the password hash) would be silently loosened on the next save.
- **Symlink write-through** — a symlinked target is written through to its referent rather than replaced with a regular file (matching `Path.write_text`'s follow semantics).

## Validation
- `./scripts/test.sh tests/test_atomic_settings_writes.py -q` → 6 passed
- New `tests/test_atomic_settings_writes.py`: replace-contents, new-file, 0600 mode preservation, failed-`os.replace` leaves original intact + no temp debris, symlink write-through, and an end-to-end check that `save_settings` routes through the atomic writer.
- Existing settings suite unaffected: `pytest tests/test_*settings*.py tests/test_auth_settings_safety.py` → 180 passed, 1 skipped

## Notes
Independent of #5797 — that PR makes `config.yaml` writes atomic via `api/paths.py`; this covers a different file (`settings.json`) with a self-contained helper, so the two do not conflict.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -8668,6 +8668,64 @@ _SETTINGS_WRITE_VERSION = 0
 _SETTINGS_WRITE_LOCK = __import__("threading").Lock()
 
 
+def _atomic_write_settings_text(path: Path, text: str) -> None:
+    """Write *text* to *path* atomically (temp file + fsync + os.replace).
+
+    ``settings.json`` was rewritten with a plain ``Path.write_text``, which
+    truncates the file in place: a crash or full disk mid-write leaves it
+    truncated/empty, so the next start loses every persisted setting (theme,
+    workspace, tab order, and the login ``password_hash``). Writing to a
+    sibling temp file, fsyncing, then ``os.replace`` keeps the old contents
+    intact until the rename commits the new ones in one step.  Mirrors the
+    tempfile+fsync+os.replace pattern already used by
+    ``webui_session_db.WebUIJsonSessionDB._atomic_write``.
+
+    The existing file's mode is carried onto the replacement: ``os.replace``
+    swaps in the temp file's inode, and a plain ``open`` respects the umask
+    (typically 0644), so without this an operator-hardened ``settings.json``
+    (chmod 0600 because it holds the password hash) would be silently loosened
+    on the next save.  New files fall back to the umask-adjusted default.
+
+    A symlinked target is written through to its referent (same follow-through
+    as the ``Path.write_text`` this replaces), rather than replacing the link
+    itself with a regular file.
+    """
+    path = Path(path)
+    write_path = path.resolve(strict=False) if path.is_symlink() else path
+    tmp = write_path.with_name(
+        f".{write_path.name}.{os.getpid()}.{threading.get_ident()}.tmp"
+    )
+    try:
+        mode = os.stat(write_path).st_mode & 0o777
+    except FileNotFoundError:
+        mode = 0o666 & ~_current_umask()
+    try:
+        with open(tmp, "w", encoding="utf-8") as handle:
+            handle.write(text)
+            handle.flush()
+            os.fsync(handle.fileno())
+        os.chmod(tmp, mode)
+        os.replace(tmp, write_path)
+    finally:
+        try:
+            os.unlink(tmp)
+        except OSError:
+            pass
+
+
+def _current_umask() -> int:
+    """Read the process umask without leaving it changed.
+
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Reading the diff against `api/config.py` on `origin/master` plus the precedent it cites (`api/webui_session_db.py:222-234`), the core change is correct — but there's one avoidable wrinkle in the new-file path worth flagging before merge.

## Summary
Swapping both `settings.json` writers (`save_settings` and the startup default-workspace rewrite) from `Path.write_text` to a temp-file + `fsync` + `os.replace` helper is the right fix: `os.replace` is atomic, so a crash/full-disk/power-loss before the rename leaves the original file byte-for-byte intact rather than truncated. It correctly mirrors the existing pattern:
```python
# api/webui_session_db.py:222
def _atomic_write(path, data):
    tmp = path.with_suffix(f".tmp.{os.getpid()}.{threading.current_thread().ident}")
    ...
    os.fsync(handle.fileno())
    os.replace(tmp, path)
```
Mode-preservation and symlink write-through are genuinely valuable additions over the precedent (the hardened-0600 case for the password-hash file is real).

## The `_current_umask()` window is avoidable
The one spot I'd push back on is `_current_umask()`:
```python
def _current_umask() -> int:
    umask = os.umask(0)
    os.umask(umask)
    return umask
```
This is only reached on the FileNotFoundError (new-file) branch to compute `mode = 0o666 & ~_current_umask()`, which is then applied via `os.chmod(tmp, mode)`. But `open(tmp, "w", ...)` **already** created the temp file with the umask applied — the process default. So on the new-file path the `os.chmod` is re-deriving a mode the OS just applied for you. Because this is a threaded server (`ThreadingHTTPServer`), the `os.umask(0)/os.umask(restore)` round-trip opens a process-wide window where a concurrent thread creating any file sees umask=0 (world-writable) — exactly the race greptile flagged.

Cleaner: skip the umask dance and only `chmod` when preserving an existing file's mode:
```python
existing_mode = None
try:
    existing_mode = os.stat(write_path).st_mode & 0o777
except FileNotFoundError:
    pass
with open(tmp, "w", encoding="utf-8") as handle:
    handle.write(text); handle.flush(); os.fsync(handle.fileno())
if existing_mode is not None:
    os.chmod(tmp, existing_mode)   # new file keeps open()'s umask-adjusted mode
os.replace(tmp, write_path)
```
That drops `_current_umask()` entirely and closes the umask race, while keeping the 0600-preservation behavior your `test_atomic_settings_writes.py` mode test asserts (the existing-file branch is unchanged).

## Minor durability note
For full crash-safety, `os.replace` guarantees atomicity but the rename's own durability isn't flushed. The precedent `_atomic_write` doesn't fsync the parent dir either, so matching it is consistent — just noting a parent-dir fsync after `os.replace` would harden the rename against a power-loss window. Optional; out of scope for parity with the existing helper.

Otherwise this is a well-tested, correctly-scoped fix. Confirmed independent of #5797 (that PR touches `config.yaml` via `api/paths.py`; this touches `settings.json` via a self-contained helper).

nesquena-hermes:
Shipped in **exp-v0.52.6** — thanks @ai-ag2026! Verified byte-identical to your PR head, Codex SAFE-TO-SHIP, full regression + your own tests green, deployed to the experimental channel. 🚀

PR #5830, opened 2026-07-09
fix: workspace image previews and downloads broken under subpath mounts

tomtong2015:
The workspace panel builds file URLs in _workspaceRouteForPath as app-relative "/api/..." strings. Most callers pass them through api(), which strips the leading slash and re-resolves against document.baseURI, so they work under a subpath mount like /hermes/ (reverse proxy, jupyter-server-proxy, etc.). But several callers consume the route DIRECTLY, bypassing api():

  - previewImg.src (image preview)
  - media / pdf / html frame .src
  - the download <a href> anchor
  - window.open() (open in browser)

Under a subpath mount a bare "/api/file/raw?..." resolves to the SERVER ROOT and 404s: image previews render broken and every download fails, while text/ code previews (which go through api()) keep working -- a confusing partial failure. Resolve the route against document.baseURI inside _workspaceRouteForPath itself so every consumer gets a correctly-mounted URL; api()-based callers are unaffected (api() tolerates an already-absolute URL). Non-browser test harnesses with no document/location keep the app-relative form.

Adds tests/test_workspace_subpath_raw_urls.py: two source regressions plus a Node functional check that the raw/img/read routes keep the mount prefix.

### DIFF
--- a/static/workspace.js
+++ b/static/workspace.js
@@ -242,6 +242,23 @@ function _workspacePathIsReadOnly(path){
 }
 
 function _workspaceRouteForPath(path, kind, opts={}){
+  // Resolve the app-relative "/api/…" route against document.baseURI so the
+  // URLs that are consumed OUTSIDE api() — previewImg.src, the media/pdf/html
+  // frame src, the download anchor, window.open — keep working under a subpath
+  // mount like /hermes/. A bare "/api/…" string resolves to the server root
+  // there and 404s. (api() strips the leading slash and re-resolves against
+  // baseURI itself, so routes passed through it are unaffected by already
+  // being absolute.)
+  const route=_workspaceRouteForPathRel(path, kind, opts);
+  if(!route) return route;
+  // Non-browser test harnesses have no document/location: keep the app-relative form.
+  const base=(typeof document!=='undefined'&&document.baseURI)||(typeof location!=='undefined'&&location.href)||'';
+  if(!base) return route;
+  const rel=route.startsWith('/') ? route.slice(1) : route;
+  return new URL(rel, base).href;
+}
+
+function _workspaceRouteForPathRel(path, kind, opts={}){
   if(!S.session) return '';
   const normalizedPath = _normalizeWorkspaceRelPath(path);
   const grant = _workspaceEscapeGrantForPath(normalizedPath);
--- a/static/workspace.js
+++ b/static/workspace.js
@@ -253,7 +253,7 @@ function _workspaceRouteForPath(path, kind, opts={}){
   if(!route) return route;
   // Non-browser test harnesses have no document/location: keep the app-relative form.
   const base=(typeof document!=='undefined'&&document.baseURI)||(typeof location!=='undefined'&&location.href)||'';
-  if(!base) return route;
+  if(!base||!/^https?:\/\//i.test(base)) return route;
   const rel=route.startsWith('/') ? route.slice(1) : route;
   return new URL(rel, base).href;
 }
--- a/tests/test_workspace_subpath_raw_urls.py
+++ b/tests/test_workspace_subpath_raw_urls.py
@@ -57,11 +57,11 @@ class TestWorkspaceSubpathRawUrls:
         )
 
     def test_route_builder_keeps_app_relative_strings_internally(self):
-        """The literal '/api/...' route strings must survive for api() callers."""
+        """The literal '/api/...' route strings must remain in _workspaceRouteForPathRel."""
         src = _workspace_js()
         assert "/api/file/raw?session_id=" in src, (
-            "the app-relative /api/file/raw route string must remain so api()-based "
-            "callers keep re-resolving it themselves"
+            "the app-relative /api/file/raw route string must remain as the internal "
+            "building block inside _workspaceRouteForPathRel (used when document is absent)"
         )
 
     @pytest.mark.skipif(NODE is None, reason="node not on PATH")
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Reading `static/workspace.js` at `origin/master` (the `_workspaceRouteForPath` builder + its call sites) against the diff here, the diagnosis and the fix both check out.

## Summary
`_workspaceRouteForPath` returns app-relative `/api/…` strings. Callers that go through `api()` are safe under a subpath mount, because `api()` strips the leading slash and re-resolves against `document.baseURI` (`static/workspace.js:2-4`):
```js
async function api(path,opts={}){
  const rel = path.startsWith('/') ? path.slice(1) : path;
  const url=new URL(rel,document.baseURI||location.href);
```
But four call sites consume the route **directly**, bypassing `api()`, so a bare `/api/file/raw?…` resolves to the server root and 404s under `/hermes/`:
- image preview — `previewImg.src` (`workspace.js:1070-1071`)
- media/pdf frame `.src` (`workspace.js:1077`, `1087`)
- download anchor `a.href` (`downloadFile`, `workspace.js:1178-1180`)
- `window.open` (`openInBrowser`, `workspace.js:1228`)

That matches the reported "image previews broken + downloads fail while text/code previews (which go through `api()` at `:1104`/`:1134`/`:1147`) keep working" partial-failure signature.

## Code reference (the fix)
The PR splits the builder: `_workspaceRouteForPathRel` keeps the literal `/api/…` strings, and the public `_workspaceRouteForPath` resolves them:
```js
const route=_workspaceRouteForPathRel(path, kind, opts);
if(!route) return route;
const base=(typeof document!=='undefined'&&document.baseURI)||(typeof location!=='undefined'&&location.href)||'';
if(!base||!/^https?:\/\//i.test(base)) return route;
const rel=route.startsWith('/') ? route.slice(1) : route;
return new URL(rel, base).href;
```

## Verification
- `api()`-based callers stay correct: `new URL(rel, base)` on an already-absolute URL returns it unchanged, and `api()` itself tolerates an absolute input, so double-resolution is a no-op.
- `cacheBust` is `&_=<ts>` (`workspace.js:1048`) appended **after** the resolved URL, which already carries a `?session_id=…` query — no query mangling.
- The empty-string guard (`if(!route) return route`) preserves the `!S.session → ''` early return in `_workspaceRouteForPathRel`.
- Non-browser test harness fallback is sound: without an `http(s)` `document.baseURI`, the app-relative form is returned as-is, which is why the existing route-string assertions still pass.

One small note: the Node functional test slices the helper block by string markers (`_escapeGrantStore(){` … `authorizeWorkspaceEscapeNavigation(`). That's brittle against future refactors of that region, but it's test-only and fine for now. Solid, well-scoped fix.

tomtong2015:
Thanks for the thorough verification — the call-site map, the double-resolution no-op, and
the cacheBust/empty-string checks are exactly the invariants I wanted holding.

Small provenance note: the `http(s)` base guard and the docstring wording you quoted came in
via `37bc19a2`, addressing greptile's two P2s — it also added a Node regression that runs the
builder with `document.baseURI = 'about:blank'` and asserts the app-relative fallback (it
really does throw `ERR_INVALID_URL_SCHEME` without the guard). CI is 18/18 green on that head.

On the slicing brittleness: agreed it's not pretty. One mitigating detail — the start marker
(`function _escapeGrantStore(){`) is the same anchor `test_issue4582_workspace_escape_navigation.py`
already slices on, so a refactor of that region would trip both tests with explicit assert
messages rather than silently passing. Happy to extract a shared `tests/_workspace_js_slices.py`
helper (or switch to a brace-matching slice) in this PR or a follow-up — your call.

Also happy to rebase onto current master if you'd like the branch updated before merge;
otherwise I'll leave it to your usual auto-rebase.

nesquena-hermes:
Thanks — the provenance note lines up with what I read at head. On your two open questions, decisions so you're not blocked:

## Slicing brittleness — leave it in this PR, don't extract the shared helper here

I checked your cross-file-anchor claim and it's *half* right, which actually argues for the lighter option. Both files slice on the same **start** anchor:

- `tests/test_workspace_subpath_raw_urls.py:43` → `src.find("function _escapeGrantStore(){")`
- `tests/test_issue4582_workspace_escape_navigation.py:70` → `src.find("function _escapeGrantStore(){")`

But the **end** markers differ — the new test ends at `async function authorizeWorkspaceEscapeNavigation(` (`:45`), while `test_issue4582` ends at `let _workspacePanelActiveTab = 'files';` (`:72`). So a refactor that renames `_escapeGrantStore` trips both with loud assert messages (your point holds); a refactor *between* the two distinct end anchors trips only one. That residual gap is small.

More to the point: grepping the tree, this region is already sliced/probed by four suites (`test_1038_pwa_auth_redirect.py`, `test_session_sidebar_resilience.py`, `test_779_html_preview.py`, plus `test_issue4582`). Extracting `tests/_workspace_js_slices.py` is a genuine cleanup, but it would pull four unrelated files into a subpath-mount bugfix and widen the blast radius for no functional gain. Keep this PR scoped to the fix. If you want the robustness locally, a brace-matching slice is a fine in-PR tweak — but even that is optional given the shared start anchor already fails loudly. I'd file the shared-helper extraction as a separate test-hygiene follow-up.

## Rebase — go ahead

Please rebase onto current master. The fix touches only `_workspaceRouteForPath` / `_workspaceRouteForPathRel` in `static/workspace.js` and a new test module, so conflict risk is near zero, and a clean head keeps CI honest ahead of merge. The change itself is approved from my side — the diagnosis, the `api()` double-resolution no-op, and the `about:blank` fallback all verified last round.

nesquena-hermes:
## Gate: SAFE TO SHIP ✅

Ran the full authoritative gate (Codex reproduce + 825-test workspace regression) on head `fd5b1d1e6`.

- **No root-mount regression:** resolving `/api/...` against `document.baseURI` produces the identical final request URL under a normal (non-subpath) deployment — the common case is behavior-unchanged.
- **Subpath-mount fixed:** all six direct consumers (`previewImg.src`, media/pdf/html frame `.src`, download `<a href>`, `window.open`) now get mount-prefixed URLs (`static/workspace.js:1087–1246`).
- **api() callers unaffected** — absolute URLs preserved, no double-prefix.
- Non-browser / `about:blank` contexts safely retain relative routes; no global collision.
- Codex: no regression risk. 503 relevant tests + browser-smoke green.

Gate-clean. Visible fix (workspace image previews under subpath mounts).

nesquena-hermes:
Shipped in **exp-v0.52.16** — thanks @tomtong2015! Gate-pass (Codex SAFE, no root-mount regression, all 6 consumers verified), 825 workspace regression tests + browser-smoke green, maintainer-approved, deployed to experimental. 🚀

PR #5832, opened 2026-07-09
fix: resolve worktree creation failure when cli namespace shadows hermes-agent/cli.py

silent-reader-cn:
## Problem

Clicking "New Worktree Conversation" in the WebUI workspace dropdown silently does nothing. The browser shows no error toast; the server logs (`errors.log`) reveal:

```
ImportError: cannot import name '_setup_worktree' from 'cli'
  (site-packages/cli/__init__.py)
```

## Root cause

`_setup_agent_worktree()` in `api/worktrees.py` uses a bare `from cli import _setup_worktree`.
When any third-party package ships a `cli/` namespace directory in
`site-packages`, Python resolves that namespace package (which lacks
`_setup_worktree`) before finding the real `hermes-agent/cli.py` module.

The specific trigger here is **stringzilla** (dependency of albucore →
albumentations), which installs CLI helper scripts at
`site-packages/cli/{split.py,wc.py}` with an empty `__init__.py`. But the
same bug would surface with **any** package that creates a `cli/` directory
in site-packages — a common namespace name on PyPI.

## Fix

Replace the bare import with `importlib.util.spec_from_file_location`,
loading `cli.py` from the absolute path discovered by `api.config._AGENT_DIR`.
This bypasses the `sys.path` namespace search entirely and is not affected
by namespace-package shadowing.

## Verification

- `import cli` now returns the expected `hermes-agent/cli.py` module
- `_setup_worktree` is accessible and callable
- stringzilla (and its dependents albucore/albumentations) continue to work
  unaffected

## Notes

The old code had a comment `# import api.config  # ensure Hermes Agent dir is on sys.path`.
This was insufficient because `api.config` appends `_AGENT_DIR` at the
*end* of `sys.path`, while `site-packages` (which may contain the
conflicting `cli/` directory) appears earlier in the search order.

### DIFF
--- a/api/worktrees.py
+++ b/api/worktrees.py
@@ -327,8 +327,27 @@ def find_git_repo_root(workspace: str | Path) -> Path:
 
 def _setup_agent_worktree(repo_root: str) -> dict:
     try:
-        import api.config  # noqa: F401  # ensure Hermes Agent dir is on sys.path
-        from cli import _setup_worktree
+        import importlib.util
+        from pathlib import Path
+
+        from api.config import _AGENT_DIR  # Hermes Agent source root
+
+        # Use importlib to load cli.py from its absolute path instead of a bare
+        # ``from cli import _setup_worktree``.  The bare import is vulnerable to
+        # namespace-package shadowing — any third-party package (e.g. the ``cli``
+        # namespace shipped by stringzilla) that creates a cli/ directory in
+        # site-packages will preempt the real hermes-agent/cli.py module.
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.6** — thanks @silent-reader-cn! Verified byte-identical to your PR head, Codex SAFE-TO-SHIP, full regression + your own tests green, deployed to the experimental channel. 🚀

Issue #1585, opened 2026-05-04
Crash recovery: WAL design + integration for in-flight assistant tokens


nesquena-hermes:
# Crash recovery: WAL design + integration for in-flight assistant tokens

## Background

When `hermes-webui` is killed (OOM, SIGKILL, container restart, kernel panic) during an active streaming agent run, tokens streamed since the last checkpoint are lost. The session JSON reverts to showing only the user's pending message; the assistant reply that was mid-stream simply vanishes.

The diagnosis from PR #1353 (@JKJameson) is correct on both counts:

1. **`_checkpoint_activity` is incremented only in `on_tool()` — so a pure-text stream with zero tool calls is never flagged for periodic checkpointing.**
2. **There is no token-level durability primitive.** `STREAM_PARTIAL_TEXT[stream_id]` and `STREAM_REASONING_TEXT[stream_id]` accumulate the in-flight tokens in memory (introduced in #893 and #1361), but they live only in process memory and are lost on crash.

But the further wrinkle is: the `_checkpoint_activity` increment alone wouldn't actually fix the bug. `s.messages` is never mutated during streaming — it only updates at `api/streaming.py:2141` *after* `agent.run_conversation()` returns. The periodic 15s checkpoint thread saves `s.pending_user_message` + `s.active_stream_id` + `s.pending_started_at`, all of which are already on disk from the pre-stream `s.save()` at line 2092. So even with `_checkpoint_activity` bumping every token, the checkpoint thread persists nothing new during a text-only stream.

The real fix is shadowing the in-memory token accumulators (`STREAM_PARTIAL_TEXT`, `STREAM_REASONING_TEXT`) to a sidecar log on disk, then replaying that log on session reload after a crash.

## User impact

This bug class has been hit by users:
- #1217 — "Webui lost my prompts" (closed; root cause was different but related — context compaction; the user-facing symptom — message vanished — is what we're addressing here).
- The streaming-loss path is silent: there is no error, no toast, no "session was interrupted" indicator. The user simply sends a message, sees the agent start to type, then on next reload the message and reply are gone.

## Design constraints

From the closer reading after #1353:

1. **Durability target is "after-crash," not "after-power-loss."** Per-event `fsync()` is overkill — the kernel's page cache will preserve append-only writes through any process death short of an unsynchronized power outage, which is not in scope.

[...]

PR #5833, opened 2026-07-09
fix(models): open the remaining state.db read projections read-only

ai-ag2026:
## Summary
Extends the `open_state_db_readonly` (`file:...?mode=ro`) treatment from #5455 to the eight pure-read `state.db` projections in `api/models.py` that were still opening a read-**write** connection.

## Why
#5455 moved the session-listing, lineage, and gateway-watcher reads onto `open_state_db_readonly` so a write-capable handle doesn't add checkpoint/lock surface while the agent streams into the same WAL `state.db`. These eight reads were missed and still did `sqlite3.connect(str(db_path))`:

`state_db_has_session`, `agent_session_rows_existing`, `agent_session_zero_message_sids`, `_read_state_db_sidebar_overrides`, `get_state_db_session_messages`, `get_state_db_session_message_keys_before_timestamp`, `get_state_db_session_summary`, `count_conversation_rounds`.

They are all `SELECT`/`PRAGMA`-only, so a read-only handle is strictly correct and consistent with the #5455 decision. The lone write path, `delete_cli_session` (a `DELETE`), deliberately keeps its writable connection.

Bonus: the two sites that lacked an existence guard (`agent_session_rows_existing`, `agent_session_zero_message_sids`) can no longer have the writable fallback silently create an empty ghost `state.db`, because the helper raises `FileNotFoundError` for a missing path (which their `except` already maps to the same empty result).

## Validation
- `./scripts/test.sh tests/test_state_db_readonly_reads_models.py -q` → 10 passed
- New `tests/test_state_db_readonly_reads_models.py`: spies on the connect used by the helper and asserts each of the eight reads opens via a `mode=ro` URI **and** still returns correct data; asserts the read-only handle rejects writes; and a source-level guard that `delete_cli_session` keeps its writable connection.
- No regression in the #5455 / state.db suite: `pytest tests/test_issue5455_lineage_readonly_reads.py tests/test_issue5455_listing_readonly_connection.py tests/test_issue1494_state_db_fd_leak.py tests/test_state_db_active_filter.py tests/test_issue5532_session_clear_state_db_replay.py` → 24 passed

## Refs
Follow-up to #5455 (#1585 WAL context).

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -27,6 +27,7 @@ from api.agent_sessions import (
     _is_continuation_session,
     is_cli_session_row,
     normalize_agent_session_source,
+    open_state_db_readonly,
     read_importable_agent_session_rows,
     read_session_lineage_metadata,
 )
@@ -4140,7 +4141,7 @@ def state_db_has_session(sid: str) -> bool:
     if not db_path.exists():
         return False
     try:
-        with closing(sqlite3.connect(str(db_path))) as conn:
+        with closing(open_state_db_readonly(db_path)) as conn:
             cur = conn.cursor()
             cur.execute("SELECT 1 FROM sessions WHERE id = ? LIMIT 1", (str(sid),))
             return cur.fetchone() is not None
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.6** — thanks @ai-ag2026! Verified byte-identical to your PR head, Codex SAFE-TO-SHIP, full regression + your own tests green, deployed to the experimental channel. 🚀

PR #5834, opened 2026-07-09
perf(sessions): parse the session index from bytes, not a decoded str

ai-ag2026:
## Summary
`/api/sessions` and its rebuild path read the session index — a multi-MB JSON list on a busy install — with `json.loads(SESSION_INDEX_FILE.read_text(encoding='utf-8'))`. That decodes the whole file to a Python `str` first, then re-scans it in the JSON parser. Handing `json.loads` the raw `bytes` lets its C parser decode UTF-8 in a single pass.

## Why
Measured on a real **6.2 MB / 2147-session** index:

| parse | time |
|---|---|
| `json.loads(read_text('utf-8'))` | 46.0 ms |
| `json.loads(read_bytes())` | 35.7 ms (**~22% faster**) |

This runs on every cache-miss sidebar rebuild (the payload cache absorbs 2.5s bursts, but a rebuild re-parses the whole index). `json.loads` accepts `bytes` and auto-detects the UTF-8 encoding, producing an **identical** object for the BOM-less index, so the change is behavior-neutral. Applied to all ten index readers across `models.py` and `routes.py` for consistency.

## Validation
- `./scripts/test.sh tests/test_session_index_parse_bytes.py -q` → 3 passed
- New `tests/test_session_index_parse_bytes.py`: bytes/str parse equivalence incl. non-ASCII titles (emoji, CJK, umlauts); an end-to-end `all_sessions` read of a Unicode index through the bytes path; and a source guard that no index reader reverts to `read_text`.
- No regression: `pytest tests/test_watermark_advance_after_edit.py tests/test_session_sidebar_cache.py tests/test_worktree_remove.py` → 37 passed

## Notes
Part of the `/api/sessions` cost profile behind #5455. A parsed-index mtime cache was considered and deliberately left out: the parsed entries are mutated in place downstream (`s['is_streaming']`, `s['profile']`), so a shared cache would need a deep copy on every hit — negating the win. Viewport windowing of the sidebar payload (the larger lever) is a separate, frontend-touching change.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -392,7 +392,7 @@ def _write_session_index(updates=None, *, session_dir: Path | None = None, sessi
             # Avoid N filesystem exists() checks under LOCK by collecting
             # on-disk IDs once before entering the critical section.
             on_disk_ids = _persisted_session_ids_snapshot()
-            existing = json.loads(session_index_file.read_text(encoding='utf-8'))
+            existing = json.loads(session_index_file.read_bytes())
             if not isinstance(existing, list):
                 raise ValueError("session index must be a list")
             with LOCK:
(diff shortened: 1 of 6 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.7** — thanks @ai-ag2026! Verified byte-identical to your PR head, Codex SAFE-TO-SHIP, full regression + your own tests green, deployed to the experimental channel. 🚀

Issue #2853, opened 2026-05-24
bug(terminal): Embedded terminal opens then immediately shows [terminal closed] on Linux (regression from 71d8a8fb)


nesquena-hermes:
# bug(terminal): Embedded terminal opens then immediately shows `[terminal closed]` on Linux (regression from #2363 / 71d8a8fb)

## Summary

The embedded composer terminal opens and immediately closes on Linux, showing `[terminal closed]` instead of a shell prompt. Affects every Linux user running master past **May 19 2026 (commit `71d8a8fb`, "fix: reap terminal shells on shutdown")**.

The PTY shell is killed by `SIGTERM` within ~10ms of being spawned, before the user can type anything. macOS/non-Linux users are unaffected because the underlying `prctl()` call falls into the `except Exception: pass` branch and silently does nothing on those platforms.

### Necrosis (Discord testers, May 24 2026 ~4:57 AM and ~9:07 AM UTC)

> Terminal is not working for me now

> It opens then closes

> It shows [terminal closed]

![Necrosis original report](https://raw.githubusercontent.com/nesquena-hermes/hermes-webui/issue-assets/issue-terminal-closes-necrosis-may24-original.png)

![Necrosis follow-up describing symptom](https://raw.githubusercontent.com/nesquena-hermes/hermes-webui/issue-assets/issue-terminal-closes-necrosis-may24-conversation.png)

Source: Discord testers channel, May 24 2026.

## Root cause

Commit `71d8a8fb` added a `preexec_fn` to the terminal's `subprocess.Popen` call (`api/terminal.py:73-84, 196`):

```python
def _terminal_shell_preexec_fn() -> None:
    """Ask Linux to terminate the PTY shell when the WebUI parent dies."""
    try:
        import ctypes
        libc = ctypes.CDLL(None)
        libc.prctl(1, signal.SIGTERM)  # PR_SET_PDEATHSIG=1, SIGTERM=15
    except Exception:
        pass
```

`PR_SET_PDEATHSIG` is **per-thread**, not per-process. The "parent" it tracks is the specific kernel thread that called `clone(2)` — not the parent process. The Linux manpage is explicit on this:

> The parent death signal setting is **cleared for the child of a fork(2)**. … Warning: the "parent" in this case is considered to be the **thread** that created this process.

WebUI uses `ThreadingHTTPServer` (`server.py:123`, `QuietHTTPServer`). Each HTTP request is handled in its own short-lived thread. The flow that breaks:

[...]

PR #5835, opened 2026-07-09
fix(terminal): stop leaking fd + thread + shell on client/shell exit

ai-ag2026:
## Summary
Closes two embedded-terminal resource leaks that accumulate over a long uptime toward fd/thread exhaustion (#4633).

**1. Master fd + `_TERMINALS` entry leaked on shell exit.**
When a shell exited on its own (user typed `exit`, the process died), `_reader_loop` broke out of its loop but only set `closed` and emitted `terminal_closed` — it never closed the pty `master_fd` nor removed the `_TERMINALS` entry. Only an explicit `close_terminal` / restart / atexit did that, so **every** self-exited shell leaked its master fd and dict entry for the rest of the WebUI's uptime. The reader loop now retires the session, using a `close_terminal(sid, expected=term)` identity guard so an old reader finishing after a restart can never tear down the new terminal that replaced its session id.

**2. Abandoned terminals accumulated without bound.**
A client that drops its output stream without POSTing `/api/terminal/close` (tab close, browser crash, network drop) leaves the shell running — there is deliberately no PDEATHSIG (see the in-file note re #2853). `_MAX_TERMINALS` (32) now caps the live population, evicting a dead-process terminal first, else the least-recently-active one (an abandoned shell idle at its prompt sorts oldest), to make room for a new one.

**Hardening carried by this change.** Making the common `exit` path a fd-closer would otherwise widen a pre-existing fd-reuse TOCTOU: `write_terminal` passes its `is_alive()` check, the shell exits, the reader closes `master_fd`, a concurrent `openpty` recycles that fd number, and the write lands on a foreign fd (input injected into another terminal). `write_terminal` and the resize ioctl now re-check `closed` and touch `master_fd` under a per-terminal `io_lock` that `close_terminal` also holds around `os.close`, so a write can never reach a closed-and-recycled fd.

## Validation
- New `tests/test_terminal_fd_leak.py` (8 tests): reader-loop retires + closes fd on exit; retire is identity-guarded against a restart; cap evicts least-recently-active and prefers dead terminals; cap skips a same-sid reuse; write-after-close never reaches `os.write`; `close_terminal` serializes its `os.close` behind `io_lock`; cap eviction passes `expected=`.
- Full terminal suite green: `test_terminal_linux_lifecycle` (real shells), `test_terminal_process_cleanup`, `test_terminal_zombie_reaper`, `test_1694_terminal_cleanup_ownership`, `test_embedded_workspace_terminal`, … — 45 passed / 1 skipped.
- Real-shell manual check: a live shell stays registered; after `exit` the entry is retired **and** the master fd is closed (leak gone).
- Full test suite: 12333 passed (only two TLS-probe tests fail locally — they spawn a TLS server + helper subprocess unavailable in the sandbox; both pass in CI and are untouched here).
- `python3 scripts/ruff_lint.py --diff origin/master` → no new violations.
- One existing test (`test_terminal_supervisor_continues_after_mixed_popen_failures`) asserted a fake-proc terminal stayed in `_TERMINALS`; its FakeProc never holds the pty slave, so the master EOFs immediately and the reader now correctly retires it (the fd-leak fix). Updated to assert on the returned objects; the real-shell lifecycle tests cover the alive-shell-stays-registered path.

## Refs
#4633 (uptime fd/thread exhaustion). The remaining terminal item — fanning the single output queue out to multiple subscribers so two tabs of one session don't split the byte stream — is a separate follow-up that builds on this lifecycle.

### DIFF
--- a/api/terminal.py
+++ b/api/terminal.py
@@ -64,11 +64,23 @@ class TerminalSession:
     output: queue.Queue = field(default_factory=lambda: queue.Queue(maxsize=2000))
     closed: threading.Event = field(default_factory=threading.Event)
     reader: threading.Thread | None = None
+    # Serializes fd-touching ops (os.write, resize ioctl) against os.close, so a
+    # write can never land on a master_fd that was closed and whose number was
+    # already recycled by a concurrent openpty — that would inject the user's
+    # keystrokes into a foreign fd. Holders re-check ``closed`` under this lock
+    # and bail if the terminal has been torn down.
+    io_lock: threading.Lock = field(default_factory=threading.Lock)
+    # Wall-clock of the last input written or output produced. Drives which
+    # terminal the cap evicts first (least-recently-active): a shell abandoned
+    # at its prompt has neither, so it sorts oldest and is evicted before an
+    # actively used one.
+    last_activity: float = field(default_factory=time.time)
 
     def is_alive(self) -> bool:
         return not self.closed.is_set() and self.proc.poll() is None
 
     def put_output(self, event: str, payload: dict) -> None:
+        self.last_activity = time.time()
         try:
             self.output.put_nowait((event, payload))
         except queue.Full:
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.10** — thanks @ai-ag2026! Byte-verified vs your head, Codex SAFE, 20 own + 355 regression tests green, deployed to experimental. Part of the long-uptime reliability work (#4633 family). 🚀

Issue #1623, opened 2026-05-04
bug(streaming): SSE app heartbeat (30s) outlives TCP keepalive timeout (~25s) — flaky-network drops mid-thinking


nesquena-hermes:
## Summary

The SSE stream's app-level heartbeat in `_handle_sse_stream` fires every **30 seconds** when the queue is idle (e.g. during long LLM thinking phases before any token is emitted). But after `v0.50.289` (#1581), the per-connection TCP keepalive is configured to **declare a peer dead after ~25 seconds** of unanswered probes. This means on a flaky network, the kernel can tear down the socket *before* the app ever sends a keepalive byte, leading to spurious mid-turn drops that the user perceives as "the stream died around 10 seconds in".

User-visible symptom from a confirmed report: turn drops at ~10s, "Reconnecting…" toast, then the v0.50.285+ recovery system inserts a "**Previous turn did not complete.**" marker. Recovery works as designed; the underlying drop is the bug.

## Reproduction

The mismatch is visible from the source alone — no stress test needed:

**Per-connection TCP keepalive** (`server.py:71-76`, the master switch + timing that v0.50.289 added):

```python
self.connection.setsockopt(socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, 10)   # start probing after 10s idle
self.connection.setsockopt(socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, 5)   # probe every 5s
self.connection.setsockopt(socket.IPPROTO_TCP, socket.TCP_KEEPCNT, 3)     # drop after 3 failed probes
self.connection.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
```

Worst-case kernel-side time to declare dead: `KEEPIDLE + KEEPINTVL × KEEPCNT = 10 + 5×3 = 25s`.

**App heartbeat** (`api/routes.py:3662-3668`):

```python
while True:
    try:
        event, data = q.get(timeout=30)        # ← 30s
    except queue.Empty:
        handler.wfile.write(b": heartbeat\n\n")  # only sent every 30s when idle
        handler.wfile.flush()
        continue
```

The same misalignment exists at the terminal SSE handler (`api/routes.py:3783`).

## Why this fires on flaky networks

When the LLM is thinking and not yet streaming tokens (reasoning models, slow first-token TTFB, network-bound providers), the queue stays empty for 10-30 seconds. During that window:

1. Kernel starts probing at 10s idle.
[...]

PR #5836, opened 2026-07-09
feat(terminal): broadcast output to every viewer instead of one shared queue

ai-ag2026:
## Summary
`TerminalSession.output` was a single `queue.Queue` read destructively by the SSE handler. Two tabs/windows viewing the same session opened independent `EventSource` connections whose handlers competed on that queue, so each viewer received only part of the PTY stream and only one could receive `terminal_closed`.

## Fix
Output now fans out to per-subscriber queues, mirroring `StreamChannel` / `SessionChannel`:
- each SSE consumer `subscribe()`s its own queue, seeded from a bounded sequence-numbered backlog;
- reconnects honor `Last-Event-ID` and replay only newer events, without gaps or duplicates within the retained backlog;
- `put_output()` serializes sequence allocation, backlog append, and non-blocking fanout so concurrent producers cannot reorder a live subscriber;
- a slow viewer drops only its own oldest item, so it cannot starve other viewers;
- `terminal_closed` is broadcast once to every subscriber;
- unsubscribe is idempotent, waits for an in-flight publication boundary, and the SSE handler always cleans up in `finally`;
- SSE events carry monotonic IDs and idle heartbeats use a valid SSE comment.

## Validation
- `tests/test_terminal_output_broadcast.py`: 12 contracts covering ordered multi-subscriber delivery, close fanout, late/reconnect replay, SSE IDs/heartbeat cleanup, per-viewer drop-oldest, safe unsubscribe, concurrent producer ordering/backlog rollover, and unsubscribe/publication races.
- Focused exact-head suite: `./scripts/test.sh tests/test_terminal_output_broadcast.py tests/test_terminal_linux_lifecycle.py tests/test_issue1623_sse_heartbeat_alignment.py -q -o addopts=` → 23 passed.
- Bounded live-order + unsubscribe-race repetition: 20/20 iterations passed.
- `python -m py_compile`, diff-scoped Ruff, `git diff --check`, and added-line security scan passed with zero findings.
- Independent review accepted exact commit `da4b4b4936f9585aedfcd74c84e77bd763ca6d1b` at the code level.

## Related context
#5835 addressed separate terminal cleanup behavior and is closed; this PR is not stacked on it. This change continues the embedded-terminal hardening discussed around #4633.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -17651,22 +17651,36 @@ def _handle_terminal_output(handler, parsed):
     handler.send_header("Connection", "close")
     end_sse_headers(handler)
     _sse_set_write_deadline(handler)  # Defect A: slow tab can't pin this thread
+    # EventSource automatically returns the last received SSE id on transport
+    # reconnect. Seed only newer backlog entries in that case so already-rendered
+    # terminal bytes (including ANSI cursor controls) are not written twice. A
+    # genuinely new viewer has no cursor and receives the full bounded backlog.
+    after_seq = None
+    last_event_id = str(handler.headers.get("Last-Event-ID", "") or "").strip()
+    if last_event_id:
+        try:
+            after_seq = max(0, int(last_event_id))
+        except ValueError:
+            pass
+    output = term.subscribe(after_seq=after_seq)
     try:
         while True:
             try:
-                event, data = term.output.get(timeout=_SSE_HEARTBEAT_INTERVAL_SECONDS)
+                event_seq, event, data = output.get(timeout=_SSE_HEARTBEAT_INTERVAL_SECONDS)
             except queue.Empty:
                 handler.wfile.write(b": terminal heartbeat\n\n")
                 handler.wfile.flush()
-                if term.closed.is_set() and term.output.empty():
+                if term.closed.is_set() and output.empty():
                     _sse(handler, "terminal_closed", {"exit_code": term.proc.poll()})
                     break
                 continue
-            _sse(handler, event, data)
+            _sse_with_id(handler, event, data, event_id=event_seq)
             if event in ("terminal_closed", "terminal_error"):
                 break
     except (BrokenPipeError, ConnectionResetError, ConnectionAbortedError):
         pass
+    finally:
+        term.unsubscribe(output)
     return True
 
 
--- a/api/terminal.py
+++ b/api/terminal.py
@@ -126,21 +126,24 @@ class TerminalSession:
             item = (self._next_output_seq, event, payload)
             self._next_output_seq += 1
             self._backlog.append(item)
-            subscribers = list(self._subscribers)
-        for q in subscribers:
-            try:
-                q.put_nowait(item)
-            except queue.Full:
-                # Slow viewer: drop its oldest chunk to stay responsive. Isolated
-                # per subscriber, so one lagging tab can't starve another.
-                try:
-                    q.get_nowait()
-                except queue.Empty:
-                    pass
+            # Keep sequence assignment, backlog append, and non-blocking fanout in
+            # one publication order. Releasing this lock before fanout lets two
+            # producers enqueue seq N+1 before seq N to a live subscriber.
+            for q in self._subscribers:
                 try:
                     q.put_nowait(item)
                 except queue.Full:
-                    pass
+                    # Slow viewer: drop its oldest chunk to stay responsive.
+                    # Isolated per subscriber, so one lagging tab can't starve
+                    # another; all queue operations remain non-blocking.
+                    try:
+                        q.get_nowait()
+                    except queue.Empty:
+                        pass
+                    try:
+                        q.put_nowait(item)
+                    except queue.Full:
+                        pass
 
 
 _TERMINALS: dict[str, TerminalSession] = {}
--- a/api/terminal.py
+++ b/api/terminal.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import errno
 import atexit
 import codecs
+import collections
 import os
 import queue
 import shutil
@@ -53,6 +54,12 @@ def _safe_close_fd(fd: int) -> None:
         pass
 
 
+# Bounds both the replay backlog and each subscriber's live queue: how many
+# output chunks are buffered before the oldest is dropped. One value so the
+# catch-up backlog and the per-viewer queue stay in lockstep.
+_OUTPUT_BUFFER_MAXLEN = 2000
+
+
 @dataclass
 class TerminalSession:
     session_id: str
@@ -61,7 +68,18 @@ class TerminalSession:
     master_fd: int
     rows: int = 24
     cols: int = 80
-    output: queue.Queue = field(default_factory=lambda: queue.Queue(maxsize=2000))
+    # Output is fanned out to every attached viewer. Each SSE consumer
+    # subscribe()s its own queue; put_output broadcasts to all of them plus a
+    # bounded backlog that a newly attaching consumer replays. Two tabs/windows
+    # on the same session therefore each get the FULL byte stream — previously a
+    # single shared queue was read destructively, so two consumers split the
+    # output between them and only one saw terminal_closed.
+    _subscribers: list = field(default_factory=list)
+    _backlog: collections.deque = field(
+        default_factory=lambda: collections.deque(maxlen=_OUTPUT_BUFFER_MAXLEN)
+    )
+    _sub_lock: threading.Lock = field(default_factory=threading.Lock)
+    _next_output_seq: int = 1
     closed: threading.Event = field(default_factory=threading.Event)
     reader: threading.Thread | None = None
     # Serializes fd-touching ops (os.write, resize ioctl) against os.close, so a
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Reviewed the broadcast commit (`81be34ee7`) at HEAD in a read-only worktree, isolated from the stacked fd-leak commit (`99ca97fee`). This is the clean fan-out layer #5844's reaper builds on, and it's correct.

## What I verified

The race that would matter for a fan-out — a chunk landing between backlog-seed and subscriber-registration — can't happen here because both are done atomically under one lock:

```python
def subscribe(self) -> queue.Queue:
    q: queue.Queue = queue.Queue(maxsize=_OUTPUT_BUFFER_MAXLEN)
    with self._sub_lock:
        for item in self._backlog:
            q.put_nowait(item)
        self._subscribers.append(q)
    return q
```

`put_output` appends to `_backlog` and snapshots `list(self._subscribers)` under the *same* `_sub_lock` (`api/terminal.py:129-132`), so a new subscriber either sees a chunk in its seeded backlog or in the live snapshot — never neither, never both. Tying the backlog `deque` and each subscriber queue to one `_OUTPUT_BUFFER_MAXLEN = 2000` (`api/terminal.py:60`) removes the divergence risk from the earlier shared-queue design.

The drop-oldest path is isolated per subscriber (`api/terminal.py:135-146`), so one lagging tab draining slowly can't starve another — a real improvement over the old single shared `queue.Queue` where two tabs split the byte stream and only one ever saw `terminal_closed`.

## Handler side

The SSE handler switch is minimal and safe (`api/routes.py:16528`, `:16546`):

```python
    output = term.subscribe()
    try:
        while True:
            event, data = output.get(timeout=_SSE_HEARTBEAT_INTERVAL_SECONDS)
            ...
    finally:
        term.unsubscribe(output)
```

`finally: term.unsubscribe(output)` guarantees the subscriber list can't grow across broken pipes / client drops. I also confirmed the heartbeat-fallback `terminal_closed` at `:16537` can't double-fire: a backlog-replayed `terminal_closed` keeps `output.empty()` False until it's drained through the normal `break` path, so the `closed.is_set() and output.empty()` branch only fires when nothing is pending. Grepped the tree — no stale `term.output` references remain anywhere in `api/`.

## Test plan

`tests/test_terminal_output_broadcast.py` covers the properties that matter: two subscribers each get the full stream (`:47`), `terminal_closed` fan-out (`:62`), late-subscriber backlog replay (`:75`), unsubscribe hygiene shrinking the list (`:90`), drop-oldest keeping the newest without starving a fast reader (`:104`), and safe unknown-queue unsubscribe (`:125`). The `#1623` heartbeat-alignment guard was correctly updated to the renamed per-consumer `output.get(...)` variable while keeping the named constant. All 17 CI shards green. Looks good to merge ahead of #5844.

nesquena-hermes:
Rebased this onto current master (its base #5835 is merged, so the old diff was showing stale commits) and gated the isolated change. The broadcast refactor is solid on the core multi-viewer invariants — but the gate found one CORE regression it introduces on SSE reconnect, plus this needs a rebase-and-force-push to clear the CONFLICTING state.

**Rebase:** #5835 (the fd-leak base) already merged as part of Batch D. Please rebase this branch onto current `master` — the rebase drops the already-upstream #5835 commit cleanly and leaves just this PR's ~+191/-15 (api/terminal.py + api/routes.py + tests). That clears CONFLICTING.

**Verified good ✅** (reproduced): two viewers receive identical ordered streams; disconnect + timeout remove their queues (no accumulation); a slow-viewer queue stays bounded and non-blocking; concurrent add/remove is lock-safe; terminal exit reaps the registry entry, reader thread, and PTY fd. 87 targeted tests pass.

**[CORE — reconnect duplicates/corrupts terminal output] `api/terminal.py:99` + `api/routes.py:16868`.** `subscribe()` always copies the **entire persistent backlog**, and `static/terminal.js:443-460` auto-reconnects and writes the replayed bytes into the existing xterm. **Verified:** consume `A`, detach, produce `B`, reattach → the viewer receives `[A, B]`; current master receives only the undelivered `B`. So a transient SSE reconnect re-renders already-shown output, and replayed **ANSI control sequences can corrupt the screen state** (cursor jumps, cleared regions).
- **Fix:** make backlog replay **cursor-aware** — sequence terminal events, emit SSE `id:` values, read `Last-Event-ID` in `api/routes.py:16868`, and seed only events newer than that cursor on reconnect. A genuinely new viewer (no `Last-Event-ID`) should still receive the bounded backlog.
- **Test:** reconnect after consuming `A` receives only output produced afterward; a brand-new viewer still gets the available backlog.

The multi-viewer broadcast (the point of the PR) is well-built — it just needs the reconnect path to not re-send consumed bytes. Once rebased + cursor-aware, I'll re-gate. (This is stacked under #5844 (idle-reaper), so this one lands first.)

ai-ag2026:
@codex review\n\nRepaired parent head: `679aca26303777077be2ff4d953c3742bd3f05cd`. Cursor-aware SSE replay and reconnect dedup regression are present; fork/PR readback matches. Child #5844 will be rebuilt on this head separately.

ai-ag2026:
@nesquena-hermes re-gate requested at head `25bb4ade5e94` — the gate-fail label looks stale. Your 10.07 reconnect finding is fixed per spec: events are sequenced, SSE responses carry `id:`, `_handle_terminal_output` reads `Last-Event-ID` on reconnect and `subscribe(after_seq)` replays only newer events (fresh viewers still get the full bounded backlog). Covered by `test_reconnecting_subscriber_replays_only_events_after_cursor` and `test_sse_reconnect_honors_last_event_id_and_emits_ids`; both Greptile buffer-constant nits also addressed (`_OUTPUT_BUFFER_MAXLEN`). CI 17/17 green, branch mergeable (merely behind master, no conflicts).

nesquena-hermes:
## 🔬 Gate certification — GREEN ✅ (with one mechanical MUST-FIX-at-merge)

**Certified head:** `25bb4ade5e942063fd9fcf0905727bd33188b359` (fresh re-push 2026-07-18; the prior `gate-fail` label was stale from the 07-10 gate at an old head). Clean-rebased onto `origin/master`; cumulative patch-id `bcfe47072a33`.
**Verdict:** The broadcast/reconnect-replay implementation is correct and fully verified — this is the cursor-aware Last-Event-ID fix the 07-10 gate demanded. One mechanical test-hygiene item must be handled at merge (restore a removed flakiness-skip); the implementation itself is ship-ready.

### What I ran
- **Codex** (reproduce, concurrency-focused): `SAFE TO SHIP` — verified fanout ordering, reconnect replay, overflow drop-oldest, close/error delivery to all subscribers, and subscriber cleanup; all 12 broadcast + 3 heartbeat + 52 neighboring terminal tests passed.
- **Senior advisor (Fable, logic review, default model)**: `REQUEST_CHANGES` on **one narrow item** — the PR removed a deliberate `pytest.skip` (see below); it confirmed the broadcast implementation (lock discipline, backlog/cursor replay, close fanout, handler cleanup, the 12-test contract) is correct and verified.
- **My own targeted + isolation runs:** 23 targeted tests (new broadcast suite + both touched files) pass; the un-skipped lifecycle test passes 3/3 in the sandbox and in the full suite.
- **Full serial suite** to completion: `13369 passed, 14 failed` — all 14 are env-only (control-verified on clean master; none touch `api/terminal.py`/`api/routes.py`; none from this diff). **GitHub Actions CI is all-green** at this head.
- Ruff-forward / scope_undef / py_compile CLEAN.

### The implementation (verified correct)
`api/terminal.py`: `TerminalSession.output` (a single destructively-read queue) → per-subscriber fan-out. `subscribe(after_seq)` returns a per-viewer queue seeded from a bounded `_backlog` deque (maxlen 2000) with monotonic `_next_output_seq`; `put_output` assigns seq + appends backlog + broadcasts to all subscribers **under `_sub_lock`** (preserving publication order); per-subscriber drop-oldest so a slow viewer can't starve others; `unsubscribe` in the SSE handler's `finally`. `api/routes.py`: reads `Last-Event-ID`, `subscribe(after_seq)`, emits `_sse_with_id(..., event_id=event_seq)`. This closes the 07-10 CORE regression (old reconnect re-sent the ENTIRE backlog → duplicate/corrupt xterm output) with exactly the cursor-aware replay that was specced, and fixes the original bug (two tabs on one session split the stream / only one saw `terminal_closed`). Reconnect uses `seq > after_seq` (no boundary dup), new viewers get the full bounded backlog.

### MUST-FIX at merge (mechanical, not an implementation defect)
`tests/test_terminal_linux_lifecycle.py` — the PR **removes** the `@pytest.mark.skip` on `test_terminal_survives_short_lived_request_thread` (whose documented reason: "timing-flaky in non-tty CI: depends on bash prompt rendering + printf echoing back through the PTY within a fixed deadline; the actual supervisor invariants are covered by the other 7 tests"). The contributor had to update that test's body to the new `subscribe()` API, but un-skipping it re-introduces a deterministic failure in release-gate-class environments (it failed in Codex's stricter sandbox; it passes in GitHub CI and the bubblewrap sandbox here, which have a working PTY). **Fix:** restore the `@pytest.mark.skip` decorator (keep the body update to the new API), or replace it with a PTY-capability probe that skips when the shell doesn't echo within the deadline. One-line restore.

**Ready to merge after restoring the one skip.** Flipping `gate-fail` → `gate-pass` and adding to Priority Queue Tier 1: the broadcast/replay implementation is correct, verified across Codex + advisor + full suite + GitHub CI, and closes both the original multi-viewer bug and the 07-10 reconnect regression. The only pre-merge action is the mechanical skip-restoration above (safe for us to do on the contributor's branch per the auto-fix-mechanical-blocker policy, or fold into the release packaging commit). Preserve **@ai-ag2026** as author (CHANGELOG credit + release notes). Gate-certifier did not merge, tag, or deploy.

@ai-ag2026 — the broadcast fix is solid and does exactly what the last gate asked for. The one thing to restore is the `@pytest.mark.skip` on `test_terminal_survives_short_lived_request_thread` (keep your API update to its body) — that test is intentionally skipped in non-tty CI and un-skipping it makes the suite fragile in some environments.

nesquena-hermes:
Shipped in **exp-v0.52.118** (experimental channel) via release #6340. Thanks @ai-ag2026! 🎉

Terminal output now broadcasts to every subscribed viewer instead of draining from one shared queue — so watching the same terminal from two tabs (or reconnecting) shows the full stream rather than a random half. SSE responses carry event `id:`s and a reconnecting viewer replays only events after its `Last-Event-ID` cursor, while fresh viewers still get the full bounded backlog.

Gate: gate-certified GREEN (Codex `SAFE TO SHIP` + Fable), full suite 13455 passed / 0 failed, ruff clean. One mechanical merge-fix applied — restored the timing-flaky `pytest.skip` on `test_terminal_survives_short_lived_request_thread` (keeping your API-updated body) per the gate's MUST-FIX-at-merge note.

PR #5840, opened 2026-07-09
perf(ui): skip the dashboard status poll while the tab is hidden

ai-ag2026:
## Summary
`_initDashboardLinkProbe` runs `setInterval(refreshDashboardStatus, 60000)`, and the interval equals `DASHBOARD_STATUS_TTL_MS` (60000). So on a **hidden** tab every 60s tick was a real `/api/dashboard/status` fetch that never hit the TTL cache — a needless network wakeup on a tab nobody is looking at.

## Why
It's the only frontend poll without a `document.hidden` guard — the gateway, streaming, and hidden-stream polls all skip while the tab is hidden (#4704). This is the same battery/CPU class as #2476.

## Fix
- `refreshDashboardStatus` short-circuits an **unforced** call while `document.hidden`. Forced calls (settings save, init, the catch-up) still run.
- `_initDashboardLinkProbe` refreshes once on `visibilitychange` back to visible, so the now-stale cache catches up when the user returns.

The guard lives inside `refreshDashboardStatus`, so the `setInterval(...)` call site (and its existing static-test anchor) is unchanged.

## Validation
- New `tests/test_dashboard_poll_visibility.py`: a node driver runs `refreshDashboardStatus` with `document.hidden` true/false — hidden skips the unforced poll (only the forced call fetches), visible polls on both; plus a static check for the guard + catch-up listener.
- `node --check static/ui.js` clean; existing `tests/test_dashboard_link_ui.py` still green (its mock `document` has no `hidden`, i.e. visible).
- `./scripts/test.sh tests/test_dashboard_poll_visibility.py tests/test_dashboard_link_ui.py` → 18 passed.

## Refs
#2476 (CPU/battery on idle/background tabs).

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -1795,6 +1795,14 @@ function _applyDashboardStatus(status){
 }
 async function refreshDashboardStatus(force=false){
   const now=Date.now();
+  // Skip the interval-driven poll while the tab is hidden: the 60s interval
+  // equals the cache TTL, so every background tick was a real /api/dashboard/status
+  // fetch that never hit the cache — a needless wakeup on a tab nobody is
+  // looking at (battery/CPU, #2476). Forced calls (settings save, init, the
+  // visibilitychange catch-up) still run. A visible tab keeps its live status.
+  if(!force&&typeof document!=='undefined'&&document.hidden){
+    return _dashboardStatusCache;
+  }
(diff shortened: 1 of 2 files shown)


nesquena-hermes:

PR #5842, opened 2026-07-09
fix(routes): bound the client rate-limit maps by sweeping stale keys

ai-ag2026:
## Summary
`_csp_report_rate_limited` and `_client_event_rate_limited` key a timestamp list by client IP and only ever revisit a key when that same IP calls again. An IP that hits the endpoint once and never returns leaves its entry in the map forever.

## Why
Behind a reverse proxy the map holds a single key (the proxy IP) — no growth. But a directly-exposed deployment accumulates one entry per distinct IP that ever hit `/api/csp-report` or the client-event endpoint, unbounded over uptime.

## Fix
Both functions now call `_prune_stale_rate_limit_keys` under the existing lock, dropping keys whose newest timestamp has aged past the 60s window. It's **size-gated** at 4096 keys, so the common few-key path (proxy, LAN) stays O(1) and only a genuinely large map pays the sweep.

The per-key `Lock` maps flagged in the same audit (`_OP_LOCKS`, `_OAUTH_START_LOCKS`, `_SKILLS_STATS_LOCKS`, `_JOURNAL_RETRY_LOCKS`) are **intentionally not touched**: evicting a `Lock` that another thread may hold or re-create via `setdefault` would break mutual exclusion, and they lack a "this key is gone forever" signal to gate eviction on (unlike `run_journal`'s delete path). Left as-is deliberately.

## Validation
- New `tests/test_rate_limit_map_prune.py`: prune keeps only in-window keys and drops fully-stale/empty ones; is a no-op below the threshold (hot path stays O(1)); and an end-to-end check that 200 one-shot IPs keep the map bounded.
- Existing `tests/test_issue1909_csp_report_only.py` still green (rate-limiting behavior unchanged).
- `./scripts/test.sh tests/test_rate_limit_map_prune.py tests/test_issue1909_csp_report_only.py` → 9 passed; `ruff` clean.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5565,11 +5565,30 @@ def _embedded_terminal_gate_allows(handler) -> bool:
     return _onboarding_gate_allows(handler)
 
 
+# Above this many distinct client keys, sweep out entries whose timestamps have
+# all aged past the window on the next update. Behind a reverse proxy the map
+# holds a single key (the proxy IP) and never trips this; a directly-exposed
+# deployment would otherwise keep one entry forever for every IP that ever hit
+# the endpoint, since a key is only revisited when that same IP calls again.
+_RATE_LIMIT_MAP_SWEEP_THRESHOLD = 4096
+
+
+def _prune_stale_rate_limit_keys(mapping: dict, cutoff: float) -> None:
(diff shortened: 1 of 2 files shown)


nesquena-hermes:

PR #5843, opened 2026-07-09
fix(persistence): fsync atomic JSON writers; make discoverability temp thread-safe

ai-ag2026:
## Summary
Three atomic JSON writers wrote a temp file and `os.replace`'d it into place but never `fsync`'d, so a power loss between the write and the physical flush could leave a zero-length or garbage file where durable JSON should be:
- `session_discoverability._atomic_write_json` (session `_index.json`)
- `passkeys._atomic_write_json` (WebAuthn credentials)
- `oauth._write_auth_json` (OAuth access/refresh tokens)

## Fix
- Each now `flush()` + `os.fsync()` before the rename, so the rename only ever commits durable bytes.
- `session_discoverability` additionally used a **pid-only** temp name (`<name>.tmp.<pid>`). WebUI runs a `ThreadingHTTPServer`, so two request threads reconciling the same `_index.json` collided on that one path — one truncating the other's temp before its `os.replace`. The temp name now includes the thread id, and a `finally` drops the temp on failure. The in-place materialize-sidecar temp gets the same thread id.

Behavior-neutral otherwise: JSON content is unchanged, and file modes are preserved — `_index.json` keeps its umask-default mode (a plain `open()` respects the umask), and `passkeys`/`oauth` keep their explicit `0600` chmod. `providers.py` already fsynced and is left untouched.

## Validation
- New `tests/test_atomic_writer_fsync.py`: each writer fsyncs before rename and round-trips its JSON; the discoverability temp name is thread-unique (two idents → two distinct temp paths); a failed rename leaves the original intact with no temp debris; passkeys/oauth keep `0600`.
- Existing suites green: `tests/test_session_discoverability_repair.py`, `tests/test_session_discoverability_audit.py`, `tests/test_passkey_auth.py`.
- `./scripts/test.sh ...` → 33 passed; `ruff` clean.

### DIFF
--- a/api/oauth.py
+++ b/api/oauth.py
@@ -191,7 +191,10 @@ def _write_auth_json(data: dict[str, Any], auth_path: Path | None = None) -> Pat
     path.parent.mkdir(parents=True, exist_ok=True)
     tmp = path.with_name(f"{path.name}.tmp.{os.getpid()}.{uuid.uuid4().hex}")
     try:
-        tmp.write_text(json.dumps(data, indent=2, ensure_ascii=False) + "\n", encoding="utf-8")
+        with open(tmp, "w", encoding="utf-8") as f:
+            f.write(json.dumps(data, indent=2, ensure_ascii=False) + "\n")
+            f.flush()
+            os.fsync(f.fileno())  # durable before rename: no truncated auth.json on power loss
         try:
             tmp.chmod(0o600)
         except OSError as exc:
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.10** — thanks @ai-ag2026! Byte-verified vs your head, Codex SAFE, 20 own + 355 regression tests green, deployed to experimental. Part of the long-uptime reliability work (#4633 family). 🚀

Issue #5854, opened 2026-07-09
Memory bloat / crash from anchor_activity_scenes on the metadata-poll path + settled render (root fix for #4633 + #5839)


nesquena-hermes:
## Summary

Two reported crashes share a single root cause in how `anchor_activity_scenes` are stored and read. This issue consolidates the full investigation and proposes the *right* fix (no interim mitigation).

- **#4633** — the WebUI server (`server.py`) RSS climbs to 12–21 GB over hours of normal use, then segfaults / OOM-kills with no traceback.
- **#5839** — the browser tab freezes and grows to 12 GB when opening a specific reasoning-model session, and also when clicking the model / workspace dropdown while such a session is loaded.

Both are the terminal failure mode of the same thing: **per-turn activity scenes are large, are re-parsed on the hot poll path, and are rendered in full — and neither path is bounded.**

## Root cause (measured, not inferred)

`anchor_activity_scenes` is a per-session dict that persists, for every assistant turn, the full "activity scene" — the ordered `activity_rows` (reasoning text, tool calls, tool output, final answer) that drive the Transparent Stream / Worklog **settled-transcript replay** so a reopened/reloaded session re-renders faithfully.

On the reporter's real files (thank you @kertisalex for the sanitized data, and @someaka / @elottolado-cpu for the instrumented memory captures):

| session | total | metadata prefix (before `messages`) | of which `anchor_activity_scenes` | scenes × rows |
|---|---|---|---|---|
| `be2d3b22e7ed` (GLM-5.2, 124 msgs) | 1500 KB | **481 KB** | **480 KB (99.7%)** | 3 scenes, 79 / 44 / 11 rows |
| `a256522490a3` (DeepSeek, 66 msgs) | 1146 KB | **255 KB** | **254 KB (99.7%)** | 2 scenes |

### Backend churn (#4633)

1. `Session.save()` lists `anchor_activity_scenes` in `METADATA_FIELDS` (`api/models.py` ~1234), so scenes serialize into the JSON **before** the top-level `messages` array.
2. `Session.load_metadata_only()` (`api/models.py` ~1373) reads only the prefix before `messages` via `_read_metadata_json_prefix(path, max_prefix_bytes=65536)` (~981). When metadata exceeds **64 KB before `messages`**, that returns `None` and it falls back to `cls.load(sid)` — a **full `json.loads` of the entire multi-MB file**.
[...]

Issue #5839, opened 2026-07-09
Browser freeze / memory leak (12GB+) when loading sessions with reasoning models (GLM-5.2) due to large run journal, duplicate reasoning fields, and recursive blockquote rendering


kertisalex:
**Describe the bug**

After a session involving a reasoning-capable model (GLM-5.2 via OpenRouter), the Hermes WebUI page becomes completely unresponsive when loading that specific session. The browser tab's memory usage grows continuously until it exhausts all available RAM (observed 600MB to 12GB in Chrome Task Manager). The page loads initially — the sidebar and header appear — but becomes non-interactive within ~1 second. The issue is reproducible on Chrome and Yandex Browser across two different machines.

Clearing all site data via DevTools → Application → Clear storage resolves it temporarily, but reopening the same session triggers the freeze again.

**To Reproduce**

1. Have a conversation using a reasoning-capable model (e.g., GLM-5.2) that returns `reasoning`/`reasoning_content` fields
2. Include several tool calls in the same session that return large responses (e.g., `read_file` with 49KB JSON, `session_search` results)
3. Let the session accumulate ~18 messages with ~5 run journal entries (the last one having 1128 events)
4. Close and reopen Hermes WebUI
5. Click on that session in the sidebar — observe browser freeze and unbounded memory growth

**Expected behavior**

The session should load and be interactive regardless of its history size or the model that produced it.

**Analysis**

Session data reveals three compounding factors:

1. **Duplicate reasoning storage** — Each assistant message stores the chain-of-thought text in both `reasoning` and `reasoning_content` fields (identical ~8KB each). Additionally, `anchor_activity_scenes.activity_rows[].text` and `activity_rows[].thinking.text` also duplicate the same reasoning text. This doubles the per-message payload for every reasoning-model response.

2. **Large run journal** — The last execution produced 1128 run journal events (536 token events, 240 reasoning events). The run journal is serialized into the session JSON response at ~1.7 MB across 5 journal files, inflating the API response significantly.

3. **Recursive blockquote rendering triggers memory explosion** — In `ui.js` (`renderMd` function, line ~6328), there's recursive handling of `blockquote` elements:

```
} else if (tag[0]=='>') {
    const stripped = tag.replace(/^>(\s+|)/, '');
    const rendered = renderMd(stripped);
```

[...]

PR #5858, opened 2026-07-09
fix(models): move anchor_activity_scenes out of the metadata prefix (#5854 backend / #4633)

nesquena-hermes:
## Summary

Root-cause fix for the **server memory-growth crash** (#4633) — the backend half of #5854. Reasoning-heavy sessions accumulate 250–480 KB of `anchor_activity_scenes`, which serialized **before** the `messages` array, so the cheap metadata-prefix read overflowed its 64 KB cap and fell back to a **full multi-MB `json.loads` of the whole sidecar on every sidebar poll**. That sustained allocation churn drove the glibc allocator high-water mark into the GB range and never returned it to the OS → RSS climbs to 12–21 GB → silent segfault/OOM after hours (@someaka confirmed the mechanism: `MALLOC_ARENA_MAX=2` cuts growth ~95%).

## Fix

Split the storage so scene bodies leave the hot metadata prefix:

- **Write** (`Session.save`): persist a compact `anchor_scene_index` fingerprint (`{scene_key: updated_at}`) in the metadata prefix — placed **after `message_count`, before `messages`** — and the full `anchor_activity_scenes` bodies **after `messages`/`tool_calls`**.
- **Read** (`load_metadata_only` / `_persisted_session_meta_prefix`): the cheap prefix now stays a few KB (fingerprint only), so the poll path never re-parses the bodies. The sidebar-poll freshness check (`_cached_session_lags_disk`) reads the fingerprint (it only ever needed scene keys + max `updated_at`); the full-session GET still reads full bodies.

**Measured:** cheap prefix drops from ~740 KB → **1.8 KB** on a 3-scene/80-row session; the poll-path re-parse churn is eliminated. Verified RED on master (its 64 KB-capped prefix read returns `None` → full parse).

## Legacy migration (lazy, safe)

Pre-#5854 sidecars serialize scenes before `messages`. They're migrated lazily — rewritten in the modern layout on their next `save()`. In the meantime:
- `_read_metadata_json_prefix` gained a dual-stop (stop at `messages` **or** `anchor_activity_scenes`, whichever first) so a legacy large-scene file still yields a cheap prefix.
- A bounded, stat-signature-keyed **legacy-facts cache** stores the authoritative `message_count` + fingerprint after one full load, so an unchanged legacy file is full-parsed **at most once** (not every poll) and stays LRU-evictable.
- All legacy-facts cache writes are **TOCTOU-guarded** (`expected_sig` captured before the parse) so an atomic sidecar replace mid-parse can never stamp stale facts under the new signature.

## Correctness / gating

Gated to convergence with **Codex over 6 rounds** — each round found a real, empirically-reproduced edge on this crown-jewel session-persistence path, each fixed with a regression test proven to fail against the buggy version:
1. Stale `_anchor_scene_index` on cached/fully-loaded sessions (spurious reload / cache-ahead data loss) → fingerprint used only on metadata-only stubs.
2. Legacy `message_count` loss from a lagging `_index.json` → authoritative full-load recovery.
3. Legacy re-parse churn + un-evictability → bounded legacy-facts cache.
4. TOCTOU on atomic replace mid-parse → `expected_sig` guard + bounded stat-stable retry.
5. Redundant unguarded cache write + a stray `.venv` staging → removed; `expected_sig` made mandatory.

Final Codex verdict: **SAFE TO SHIP — no regression risk found.**

- Full suite: **12,522 passed**, 15 skipped, 1 xfailed, 2 xpassed (0 failures). `ruff` clean, `py_compile` clean.
- **17 targeted regression tests** in `tests/test_issue5854_anchor_scene_split.py` covering: serialization order, cheap-prefix size, stub correctness, full-load round-trip, all freshness directional cases, legacy count recovery, evictability, cache-miss, TOCTOU, and modern/legacy fingerprint gating.

Backend-only; scene render fidelity is unchanged.

## Scope note

This resolves the **server crash** (#4633). The **browser freeze** (#5839 — oversized settled-scene DOM) is a separate render-side workstream (lazy collapsed-worklog DOM + Transparent-Stream row expander) and will be a follow-up PR with screenshots.

This PR is self-built (nesquena-hermes) and touches the session-serialization path, so it should get an independent review before merge.

Refs #4633, #5839. Plan/investigation: #5854.


nesquena (approved):

Independent review of the #4633 server-memory-growth root-cause fix (the backend half of #5854). This is on the crown-jewel session-serialization path, so I traced the on-disk layout change, round-trip fidelity, cross-tool safety, and every Codex-caught edge — and drove the real save/load with my own harness.

### The bug

Reasoning-heavy sessions accumulate 250–480 KB of `anchor_activity_scenes` serialized **before** `messages`, so the 64 KB-capped metadata-prefix read overflowed and fell back to a full multi-MB `json.loads` on **every sidebar poll** → glibc high-water mark into the GB range → RSS 12–21 GB → OOM/segfault after hours.

### What this ships

`api/models.py` (+~230), `tests/test_issue5854_anchor_scene_split.py` (+17 tests), CHANGELOG. Agent-authored (`nesquena-hermes`). Backend-only; scene render fidelity unchanged. The browser-freeze half (#5839) is a separate follow-up.

Branch was `CONFLICTING` on the CHANGELOG only; `api/models.py` merged cleanly. Force-pushed `e041659e`.

### Cross-tool safety — verified

`anchor_activity_scenes` / `anchor_scene_index` / the metadata-prefix reader appear **nowhere** in the fresh hermes-agent tarball — this sidecar layout is WebUI-only. So reordering the JSON keys (scenes after `messages`) is safe: any full-parse reader is key-order-agnostic; only WebUI's prefix-cheap-read optimization cares, and it's updated in lockstep. ✓

### The mechanism — traced + harnessed against real save/load

- **Write** (`Session.save`): a compact `anchor_scene_index` fingerprint (`{scene_key: updated_at}`) is placed **after `message_count`, before `messages`**; full scene bodies serialize **after `messages`/`tool_calls`**. The `_placed` set prevents double-emit, and `_anchor_scene_index` is `_`-prefixed so it's excluded from the `extra` spill.
- **Read** (`_read_metadata_json_prefix`): dual-stop at `messages` **or** `anchor_activity_scenes`, whichever first — modern files stop at `messages` (small prefix), legacy files stop at the scenes key (still cheap, still captures the pre-scenes `message_count`).

My round-trip harness on a real 3-scene/80KB-body session (237 KB file):

| Check | Result |
|---|---|
| Order: fingerprint before `messages`, bodies after | ✅ |
| Cheap prefix size | **237 KB → 1.86 KB** (the #4633 churn eliminated) ✅ |
| Prefix carries `message_count` + fingerprint keys, **no bodies** | ✅ |
| Metadata-only stub: fingerprint authoritative (keys + max updated_at) | ✅ |
| Full load: scene bodies **byte-preserved** | ✅ |
| Re-save: layout order stable (idempotent) | ✅ |

### Correctness of the tricky invariants

- **Fingerprint authority (Codex r1) is correctly scoped:** `_session_scene_keys`/`_session_scene_updated_at` trust `_anchor_scene_index` **only** when `_loaded_metadata_only` is True. A fully-loaded session reads real records, because in-place scene mutation staleness the load-time fingerprint. `_cached_session_lags_disk` uses real records for the (always-full) cached side and the fingerprint for the disk prefix — the right asymmetry. ✓
- **Legacy recovery (Codex r2/r3):** a legacy file with no `message_count` in the (scenes-stopped) prefix and no modern `anchor_scene_index` recovers authoritative facts via one full load, cached in a bounded stat-signature-keyed LRU so an unchanged legacy file full-parses **at most once**, not per poll. My legacy harness confirmed the dual-stop keeps the prefix tiny (0.05 KB) and the facts cache populates with the right count + scene keys; and for a realistic file (`message_count == len(messages)`) the cache-miss, cache-hit, and facts paths **all agree**. (The cache stores `len(messages)` — the authoritative value if the persisted field ever lags.) ✓
- **TOCTOU (Codex r4/r5):** `_legacy_sidecar_facts_put` requires a mandatory `expected_sig` captured **before** the parse and commits only if the file's current signature still matches — an atomic replace mid-parse can never stamp stale facts. The metadata-only recovery deliberately does **not** re-cache after its own full-load (delegates to `load()`'s guarded write) to avoid the unguarded-second-write hazard. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Modern large-scene session, sidebar poll | cheap 1.8 KB prefix, no body parse ✅ |
| Metadata-only stub freshness | fingerprint (keys + max updated_at) ✅ |
| Fully-loaded session, scenes mutated in place | real records, not stale fingerprint ✅ |
| Legacy large-scene sidecar, first poll | one full load → facts cached ✅ |
| Legacy sidecar, subsequent polls (unchanged) | facts cache hit, no re-parse ✅ |
| Legacy file atomically replaced mid-parse | facts not cached (sig mismatch) ✅ |
| Legacy → next save() | rewritten modern; fallback stops firing ✅ |
| Full load → re-save | scene bodies preserved, order stable ✅ |
| Genuine 0-message modern session | count stays 0 (prefix has it) ✅ |

### Tests

- `tests/test_issue5854_anchor_scene_split.py` — **17/17** (order, prefix size, stub correctness, full round-trip, all freshness directional cases, legacy count recovery, evictability, cache-miss, TOCTOU, modern/legacy fingerprint gating).
- **RED-on-master verified independently:** 10 of 17 fail against master's `models.py`. Load-bearing.
- Independent round-trip + legacy harnesses (above) confirm the on-disk behavior directly.
- **Full suite (post-rebase): 12368 passed / 0 failed** in 282s (deselected the known darwin CRLF flake).

- The legacy-facts cache stores `len(messages)` while a first full-load reports the persisted `message_count` field; these differ only for a corrupt file where `message_count ≠ len(messages)` (never produced by any real writer). `len(messages)` is the safer of the two, so the resolution is correct; noting only for completeness.

### Recommendation

A genuinely well-engineered fix to the #4633 OOM: the hot metadata prefix drops ~130× (237 KB → 1.86 KB, harness-confirmed), scene bodies round-trip byte-for-byte, the fingerprint-vs-real-records authority is correctly gated on metadata-only stubs, and the legacy path is bounded, TOCTOU-guarded, and self-retiring on the next save. Cross-tool-safe (WebUI-only format). 17 tests, 10 RED on master; full suite clean. Ship.

nesquena-hermes:
## Gate: SAFE TO SHIP ✅ (data-integrity verified by reproduction)

Full authoritative gate (Codex reproduce + 2945-test session/serialization regression) on head `e041659e`.

**Backward-compat + round-trip — the crux — verified with real data:**
- A real **933 KB legacy sidecar** (old on-disk layout) loaded all messages + scenes correctly through the reader.
- **Round-trip** old → load → new-save → reload → save preserved the complete `anchor_activity_scenes` subtree exactly: row order, Unicode, tool payloads, and live→settled scene mappings all intact.
- No other reader depends on the old key order (repo-wide caller trace); scene persistence + hydration intact.

**The fix works:** the metadata prefix dropped from overflowing the 64 KB budget to **1,837 bytes** for a reasoning-heavy session → modern reads need no full parse; a legacy file does one authoritative parse then serves from cache; an oversized header still correctly falls back to full parse. This is the #4633 server memory-growth root cause.

- UTF-8 + atomic write preserved; no bare except; no scope creep (only the anchor-scene move).
- Codex: no regression risk. 17 own + 2945 regression tests green; CI 18/18.

Gate-clean. Self-built backend crash/data-integrity fix on the session-persistence path → holding for independent review + maintainer sign-off before merge.

nesquena-hermes:
Shipped in **exp-v0.52.19** — independent review (nesquena) APPROVED + gate-pass (Codex reproduced a 933KB legacy sidecar loading clean + exact round-trip; metadata prefix now ~2KB), 17 own + 2945 regression tests green. Fixes the #4633 server memory-growth root cause. 🚀

PR #5859, opened 2026-07-09
Release: write Office upload sidecars (#5734)

nesquena-hermes:
Ships **#5734** (@rodboev) — write Office upload sidecars (#540).

## What
On upload of a `.docx`/`.xlsx`/`.pptx` to a workspace, writes a best-effort `<name>.md` sidecar with the extracted text (reuses the shipped `api/office_documents.py` extractor — no new dependency), so the agent can read the document's content directly.

## Gate
- Codex: **SAFE TO SHIP** — path-contained (`O_NOFOLLOW` openat-walk, `is_relative_to(workspace)`), collision-safe (`O_CREAT|O_EXCL` never overwrites an existing `.md`), best-effort (a sidecar failure leaves the uploaded Office file intact; partial-sidecar cleanup on failure), UTF-8, generic error reporting.
- Tests: 27 own + 1443 upload/workspace regression green. Nathan concept-approved.

Closes #5734.

PR #5863, opened 2026-07-10
Release: fix #4633 server memory growth — split anchor scenes after messages (#5858)

### LINKED ISSUE #4633: Silent server.py crash after ~9-16h uptime (no traceback, no shutdown audit)

### LINKED ISSUE #5854: Memory bloat / crash from anchor_activity_scenes on the metadata-poll path + settled render (root fix for #4633 + #5839)

nesquena-hermes:
Ships **#5858** (@nesquena-hermes) — split `anchor_activity_scenes` after `messages` to fix the #4633 server memory-growth root cause (#5854 backend half).

## What
Reasoning-heavy sessions accumulate 250–480 KB of `anchor_activity_scenes` serialized *before* the `messages` array, so the 64 KB metadata-prefix read always overflowed → full-parse fallback on every read → memory growth. The scene data now serializes *after* `messages`; the metadata prefix drops to ≈2 KB and modern reads skip the full parse.

## Gate
- **Independent review:** @nesquena **APPROVED** end-to-end on this head (`e041659e`).
- **Codex: SAFE** — reproduced: a real 933 KB legacy-layout sidecar loads all messages+scenes correctly; old→load→save→reload→save round-trips the scene subtree exactly (order/Unicode/tool payloads/mappings); metadata prefix now 1,837 bytes (was overflowing); legacy files do one authoritative parse then cache; oversized header still falls back. UTF-8 + atomic write preserved; no scope creep.
- 17 own + 2945 session/serialization regression tests green; CI 18/18.

Closes #5858, closes #5854.


nesquena (approved):
## Review — ✅ approved (release-staged rebuild of #5858, which I reviewed in depth)

This is the release-staged version of **#5858** (the #4633 server-memory-growth fix — split `anchor_activity_scenes` after `messages`). I verified it's the same change, not a re-implementation:

- **Net contribution over current master is byte-identical to #5858**: +724/−42 across `api/models.py` + `tests/test_issue5854_anchor_scene_split.py` + CHANGELOG. A direct `git diff` of this branch's `api/models.py` and test against the approved `fix/5854-anchor-scene-split` branch is **empty** — the only reason the two branches differ at all is that this one is rebased onto a newer master that already carries #5859 (Office sidecars).
- **CI is fully green** on this branch.

My full deep review of the substance is on **#5858** ([pullrequestreview-4667311300](https://github.com/nesquena/hermes-webui/pull/5858#pullrequestreview-4667311300)) and stands verbatim: cross-tool-safe (WebUI-only sidecar format), round-trip byte-preserving (harnessed: prefix 237 KB → 1.86 KB), the fingerprint-vs-real-records authority correctly gated on metadata-only stubs (Codex r1), and the legacy path bounded + TOCTOU-guarded + self-retiring on next save. 17 tests, 10 RED on master; I ran the full suite green on both #5858 (12368) and current master (12363).

### Recommendation

✅ **Approved** as the release-staged, code-identical rebuild of the already-deep-reviewed #5858. No new surface to trace; CI green. Ship.

PR #5865, opened 2026-07-10
fix(settings): centralize auxiliary task catalog

starship-s:
## Thinking Path

- Auxiliary-model settings maintained task definitions in both Python and JavaScript, so the two lists could drift independently.
- That drift left the retired `session_search` LLM route visible as a configurable auxiliary model even though session search itself now runs directly through SQLite/FTS.
- The broader `auxiliary:` config section is not a safe schema source because it may contain internal routes that should not become picker rows or writable API slots.
- The narrow fix is to make one controlled WebUI backend catalog own row metadata, validation, reset behavior, and API ordering while the frontend renders the returned payload.
- Provider/model choices remain sourced from the existing dynamic `/api/models` inventory.

## What Changed

- Added a canonical backend auxiliary-task catalog with task keys, labels, and descriptions.
- Derived auxiliary update validation and reset iteration from that catalog.
- Reset now removes the specifically retired `session_search` assignment while preserving unrelated internal auxiliary mappings.
- Changed Settings to normalize, render, and save the task rows returned by `/api/model/auxiliary` instead of a frontend slot array.
- Removed the stale Session Search auxiliary row and its locale strings.
- Kept unknown persisted `auxiliary:` mappings hidden and unwritable while preserving frontend support for future tasks explicitly returned by the API.
- Added backend and Node-executed frontend regression coverage for ordering, metadata fallback, malformed rows, duplicates, unknown persisted mappings, and reset behavior.

## Why It Matters

- Settings no longer presents a model selector for an LLM route that the built-in session-search feature does not use.
- Future task-catalog changes require one WebUI backend update instead of coordinated Python, JavaScript, and locale-list edits.
- Internal or deployment-specific auxiliary mappings cannot accidentally become user-visible writable settings.
- Session search remains available as a direct SQLite/FTS feature; only its obsolete auxiliary-model configuration row is removed.

## Screenshots

Before — current `master`, Settings → Preferences at 1440×1200. Session Search appears as an auxiliary-model row.

Before: auxiliary settings include the stale Session Search row

After — this branch, Settings → Preferences at 1440×1200. The stale row is gone and the remaining task controls retain their layout.

After: auxiliary settings render the backend catalog without Session Search

## Verification

Targeted settings, routing, service-tier, and locale coverage:

`./scripts/test.sh tests/test_auxiliary_models_settings.py tests/test_manual_title_regenerate_timeout.py tests/test_issue4536_service_tier.py tests/test_chinese_locale.py tests/test_korean_locale.py tests/test_russian_locale.py tests/test_spanish_locale.py -q`

Result: `90 passed`

Syntax and hygiene:

- `python3 -m py_compile api/config.py`
- `git diff --check -- api/config.py static/panels.js static/i18n.js tests/test_auxiliary_models_settings.py`

All passed.

Manual browser validation:

- Captured both screenshots from real isolated WebUI instances at `1440×1200`, DPR 1.
- Confirmed the baseline API/UI exposes 12 rows including `session_search`.
- Confirmed the patched API/UI exposes 11 canonical rows and omits `session_search`.
- Confirmed no onboarding overlay, modal, toast, private data, clipping, or layout regression appears in either capture.

## Risks / Follow-ups

- WebUI still owns one backend task catalog independently of Hermes Agent. If Agent later publishes a stable public auxiliary-task registry, that registry can replace this catalog without another frontend rewrite.
- This PR intentionally does not infer task schema from arbitrary `auxiliary:` config keys.

## Model Used

- OpenAI / GPT-5.3 Codex Spark via OpenCode for implementation and review.
- Greptile for independent PR review; its reset-path finding was verified and fixed.
- OpenAI / GPT-5.6 Sol for orchestration and verification.


nesquena-hermes:
Reviewed the diff at HEAD `2dc294982` against `origin/master` (read-only worktree). This is a clean refactor and I traced it end to end — `api/config.py`, `static/panels.js`, `static/i18n.js`, and the test file — plus the agent side to confirm the contract. Verdict: the approach is right and the retired-slot handling is correct.

## What the change does, and why it's sound

The old design duplicated the task list in three places — `AUX_TASK_SLOTS` in `api/config.py`, the `_AUX_TASK_SLOTS` JS array in `static/panels.js`, and per-locale label pairs in `static/i18n.js`. Any drift left a phantom row. This PR makes the backend catalog the single owner and has the frontend render whatever `/api/model/auxiliary` returns.

`api/config.py:4147` (new `AUXILIARY_TASK_CATALOG`) derives `AUX_TASK_SLOTS` from it, and `_iter_auxiliary_task_rows()` only ever emits rows for catalog keys:

```python
for slot in AUXILIARY_TASK_CATALOG:
    key = str(slot["key"]).strip()
    if not key:
        continue
    rows.append(_aux_task_payload(key, aux_cfg.get(key, {}), slot["label"], slot["description"]))
```

Because the loop iterates the catalog (not `cfg["auxiliary"].keys()`), stray config keys like `session_search`, `monitor`, or a deployment-specific route can never become picker rows or writable slots. That matches the agent's own note in `hermes_cli/config.py:1588` — "session_search no longer uses an auxiliary LLM ... Existing values in user config.yaml files are harmless leftovers and ignored." The WebUI now behaves the same way. Good layer alignment.

## Retired-slot reset is the right narrow fix

`set_auxiliary_model("__reset__", ...)` at `api/config.py:4275` pops only the explicitly-retired keys before the auto-reset loop:

```python
for retired_slot in RETIRED_AUX_TASK_SLOTS:
    aux_cfg.pop(retired_slot, None)
```

`test_reset_removes_retired_slot_but_preserves_other_unknown_mappings` (line 615) confirms `session_search` is deleted while an unrelated `monitor:` mapping survives untouched. That's the correct call — reset shouldn't treat arbitrary config as task schema and nuke deployment-specific keys.

## Cross-repo check — one cosmetic note

I diffed the new catalog order against the agent's `hermes_cli/web_server.py:5212` `_AUX_TASK_SLOTS`:

- WebUI: `vision, web_extract, compression, approval, mcp, title_generation, skills_hub, curator, kanban_decomposer, profile_describer, triage_specifier`
- Agent: `vision, web_extract, compression, skills_hub, approval, mcp, title_generation, triage_specifier, kanban_decomposer, profile_describer, curator`

The membership matches (both drop `session_search`), but the display order diverges. That's purely cosmetic since each surface now owns its own ordering, and it doesn't affect correctness. The only nit is the doc comment at `api/config.py:4145` still says "Keep in sync with ... `hermes_cli/web_server.py` `_AUX_TASK_SLOTS`" — with the orders diverging, consider softening that to "keep the *set* in sync" so a future reader doesn't try to reconcile ordering that's intentionally independent.

## Test coverage looks complete

`test_get_auxiliary_models_omits_session_search_and_includes_metadata` (556) pins the exact 11-row payload and asserts unknown keys (`future_task`, `monitor`, `tts_audio_tags`) are dropped; `test_no_hardcoded_aux_task_slot_array` (165) guards against the JS array creeping back; `test_session_search_aux_task_i18n_keys_removed` (441) covers the locale purge. `api_key` write-only behavior is preserved (`_aux_task_payload` emits `api_key_set` only). Nothing to block on here.

Only actionable item is the comment wording; everything functional checks out.

nesquena-hermes:
## Gate: SAFE TO SHIP ✅

Full authoritative gate (Codex reproduce + 887-test settings/config regression + DOM harness) on the current head.

**No task lost, no drift:** only the retired `session_search` task is removed; all 11 other auxiliary tasks + their `auto`/empty defaults are unchanged. `panels.js` now renders AND saves exclusively from the single backend catalog (the whole point — Python and JS can no longer drift). All 22 label/description keys match across all 15 locales.

**Safe migration:** legacy configs containing `auxiliary.session_search` load without exception and preserve the user's OTHER custom bindings (verified a custom `vision` binding survives). Reset-All preserves canonical-task behavior; only the retired mapping is purged. Session search itself continues via `/api/sessions/search` + agent `SessionDB`, independent of the retired auxiliary slot.

**Verified:** executable DOM harness rendered all 11 rows with both controls bound; versioned static assets prevent old/new panel-backend cache skew on deploy; no classic-script global collision. 90 focused + 887 broad settings/config tests green (one unrelated `test_issue2245` mixed-case-provider flake confirmed pre-existing + isolation-only, passes standalone on clean master).

Codex: no regression risk. Gate-clean.

PR #5872, opened 2026-07-10
fix(#4685): meter post-compression model context

### LINKED ISSUE #4685: Track long-running context compression exhaustion and recovery UX

rodboev:
## Thinking Path

- WebUI correctly preserves the visible transcript while post-compression pruning reduces only the next model-facing context.
- The context indicator still receives the previous provider-reported prompt count, so it can describe the pre-prune request instead of the prepared next one.
- The change stores an explicitly labeled, short-lived estimate for that prepared snapshot without changing provider accounting or recovery behavior.

## What Changed

- `api/models.py`, `api/streaming.py`, and `api/routes.py` persist, emit, and expire the post-compression model-context estimate at the request boundary.
- `static/ui.js` prefers that estimate only while it describes the next request and labels the source in the context indicator.
- Focused regression coverage distinguishes the pruned request from the visible transcript and preserves ordinary usage behavior.

## Why It Matters

After compression, the context ring reflects the request the model will actually receive rather than a larger historical prompt. The provider-reported metric remains intact for normal usage and diagnostics.

## Verification

Focused tests cover the post-prune request estimate, session lifecycle, lineage variants, and context-indicator source selection; CI runs the full matrix.

## Risks / Follow-ups

The estimate is advisory and clearly labeled. Provider-reported prompt usage remains the authority once the next request runs.

## Upstream

Closes #4685.

## Model Used

GPT-5.5 via Codex CLI

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -1108,6 +1108,7 @@ class Session:
                  context_engine_state=None,
                  context_length=None, threshold_tokens=None,
                  last_prompt_tokens=None,
+                 post_compression_context_tokens_estimate=None,
                  compression_recovery=None,
                  recommended_recovery_action=None,
                  compression_recovery_source_session_id=None,
@@ -1164,6 +1165,10 @@ class Session:
         self.context_length = context_length
         self.threshold_tokens = threshold_tokens
         self.last_prompt_tokens = last_prompt_tokens
+        _post_compression_tokens = _parse_nonnegative_int(post_compression_context_tokens_estimate)
+        self.post_compression_context_tokens_estimate = (
+            _post_compression_tokens if _post_compression_tokens and _post_compression_tokens > 0 else None
+        )
         self.compression_recovery = compression_recovery if isinstance(compression_recovery, dict) else {}
         self.recommended_recovery_action = recommended_recovery_action
         self.compression_recovery_source_session_id = (
@@ -1255,6 +1260,7 @@ class Session:
             'context_engine', 'compression_anchor_engine', 'compression_anchor_mode',
             'compression_anchor_details', 'context_engine_state',
             'context_length', 'threshold_tokens', 'last_prompt_tokens',
+            'post_compression_context_tokens_estimate',
             'compression_recovery', 'recommended_recovery_action',
             'compression_recovery_source_session_id', 'compression_recovery_action',
             'truncation_watermark',
@@ -1614,6 +1620,7 @@ class Session:
             'context_length': self.context_length,
             'threshold_tokens': self.threshold_tokens,
             'last_prompt_tokens': self.last_prompt_tokens,
+            'post_compression_context_tokens_estimate': self.post_compression_context_tokens_estimate,
             'compression_recovery': self.compression_recovery,
             'recommended_recovery_action': self.recommended_recovery_action,
             'gateway_routing': self.gateway_routing,
--- a/api/routes.py
+++ b/api/routes.py
@@ -19929,6 +19929,7 @@ def _prepare_chat_start_session_for_stream(
     s.model = model
     s.model_provider = model_provider
     s.active_stream_id = stream_id
+    s.post_compression_context_tokens_estimate = None
     s.pending_user_message = msg
     s.pending_attachments = attachments
     s.pending_started_at = started_at if started_at is not None else time.time()
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -4411,6 +4411,45 @@ def _prune_context_tool_results_after_compression(agent, context_messages):
     return pruned_messages
 
 
+def _estimate_post_compression_context_tokens(agent, context_messages, system_message):
+    """Return a display-only estimate for the prepared post-compression request."""
+    try:
+        from agent import model_metadata
+
+        messages = context_messages or []
+        tools = getattr(agent, 'tools', None) or None
+        request_estimator = getattr(model_metadata, 'estimate_request_tokens_rough', None)
+        if callable(request_estimator):
+            estimate = request_estimator(messages, system_prompt=system_message or '', tools=tools)
+        else:
+            message_estimator = getattr(model_metadata, 'estimate_messages_tokens_rough', None)
+            if callable(message_estimator):
+                estimate = message_estimator(messages)
+                if system_message:
+                    estimate += message_estimator([{'role': 'system', 'content': system_message}])
+                if tools:
+                    estimate += message_estimator([{'role': 'system', 'content': str(tools)}])
+            else:
+                try:
+                    from agent.context_compressor import _estimate_msg_budget_tokens
+                except ImportError:
+                    return None
+
+                estimate = sum(
+                    _estimate_msg_budget_tokens(message)
+                    for message in messages
+                    if isinstance(message, dict)
+                )
+                if system_message:
+                    estimate += _estimate_msg_budget_tokens({'role': 'system', 'content': system_message})
+                if tools:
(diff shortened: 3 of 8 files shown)


nesquena-hermes:
Read the full diff at `55794a5` against `origin/master` — `api/streaming.py`, `api/models.py`, `api/routes.py`, the four `static/*.js` files, and the new test module. I focused on the cross-repo contract, since the estimate leans entirely on agent-side helpers, and on the estimate's lifecycle. Both hold up.

## Cross-repo contract verified

`_estimate_post_compression_context_tokens` (`api/streaming.py:4414`) resolves three agent helpers by name. All exist in the pinned agent and the primary path's signature matches exactly:

```python
# agent/model_metadata.py:2546
def estimate_request_tokens_rough(
    messages, *, system_prompt="", tools=None,
) -> int:
```

vs the PR's call `request_estimator(messages, system_prompt=system_message or '', tools=tools)`. The two fallbacks — `estimate_messages_tokens_rough` (`agent/model_metadata.py:2470`) and `agent.context_compressor._estimate_msg_budget_tokens` (`agent/context_compressor.py:339`) — also resolve. The `getattr(..., None)` + `callable()` guards mean an older agent that lacks the request estimator degrades gracefully to the message counter, then to the compressor budget counter, and returns `None` only if all three are missing. That's the right defensive shape for a display-only field.

## Lifecycle looks sound

The estimate is computed right after `_prune_context_tool_results_after_compression` (`api/streaming.py:9026-9033`), so it genuinely describes the *next* request (post additional tool-result prune), which is distinct from the just-completed turn's provider count. It's emitted at `done` (`api/streaming.py:9668`), persisted through `Session.compact()`/`_FIELDS` (`api/models.py`), and reset to `None` at chat-start in `_prepare_chat_start_session_for_stream` (`api/routes.py:19932`). `test_chat_start_clears_expired_post_compression_estimate` and `test_estimate_lineage_matrix` pin both the reset and the child/cron non-inheritance, so a stale estimate can't leak into a turn it no longer describes.

## One thing worth a maintainer's eye (not a blocker)

In `_syncCtxIndicator` (`static/ui.js:5911-5915`) the estimate *unconditionally* wins over the real provider `last_prompt_tokens` whenever it's present:

```js
const contextPromptTok=hasPostCompressionEstimate?postCompressionEstimate:promptTok;
```

That's intentional and honestly labeled ("Estimated next model context"), but it means immediately after a compressing turn the ring shows a rough char/4-and-flat-image estimate rather than the measured count the completed turn actually used. Since the estimate is advisory and cleared on the next send, the tradeoff is defensible — just flagging it so the label divergence is a deliberate call. Everything else LGTM; agent contract and reset coverage both check out.

nesquena-hermes:
Shipped in **exp-v0.52.48** via release #5977 (closes #4685). Post-compaction context indicator now shows a short-lived labeled estimate that expires when the next real provider count arrives; provider accounting untouched. Gate: Codex SAFE + crown-jewel stream-gate GREEN. Thanks @rodboev!

PR #5882, opened 2026-07-10
Fail closed on incomplete OpenRouter pricing

### LINKED ISSUE #5880: bug(models): incomplete OpenRouter pricing is labelled free
## Bug description

The OpenRouter free-tier augmentation treats missing or `null` pricing fields as zero. A model with incomplete pricing metadata can therefore be labelled `(free)` and added to the free-tier picker even when the API has not established that both prompt and completion pricing are zero.

`api/config.py` currently uses:

```python
float(_pricing.get("prompt", "0") or "0") == 0
and float(_pricing.get("completion", "0") or "0") == 0
```

An empty pricing object, two `null` fields, or a single zero-valued field with the other field missing all evaluate as free.

## Expected behavior

Fail closed for pricing-derived free status:

- Require `pricing` to be a dictionary.
- Require both `prompt` and `completion` fields to be present.
- Require both values to be numeric and zero.
- Continue accepting IDs ending in `:free`, since that is an explicit provider designation.

## Actual behavior

Missing and `null` values default to zero, potentially presenting a paid or unknown-price model as free.

## Evidence

A direct probe of the current expression returns `True` for:

- `{}`
- `{"prompt": null, "completion": null}`
- `{"prompt": "0"}`
- `{"completion": "0"}`

The behavior is present on current `master` and predates GPT-5.6 work. The live OpenRouter catalog sampled on 2026-07-10 had complete pricing for all current free candidates, so this is a latent resilience/billing-label bug rather than a currently observed bad catalog row.

## Proposed regression coverage

Exercise missing, partial, `null`, and malformed pricing payloads and verify they are excluded unless their model ID explicitly ends in `:free`. Keep a complete zero-pricing case to verify pricing-derived free models still work.

igorroncevic:
## Summary
- fail closed when OpenRouter pricing metadata is missing, partial, `null`, or malformed
- keep explicit `:free` model IDs selectable even when pricing metadata is incomplete
- add behavioral regression coverage for all ambiguous pricing shapes

## Validation
- TDD red baseline: 4 expected failures, 1 passing malformed case
- fixed regression: 5/5 pricing cases passed
- broader OpenRouter/model-picker suite: 42 passed
- changed-line Ruff passed
- Python compilation and `git diff --check` passed
- added-line secret and dangerous-execution scans passed

Closes #5880

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -6890,14 +6890,20 @@ def get_available_models(*, prefer_cache: bool = False, force_refresh: bool = Fa
                             _mid = str(_item.get("id") or "").strip()
                             if not _mid or _mid in seen_ids:
                                 continue
-                            _pricing = _item.get("pricing") or {}
-                            try:
-                                _is_free = (
-                                    float(_pricing.get("prompt", "0") or "0") == 0
-                                    and float(_pricing.get("completion", "0") or "0") == 0
-                                )
-                            except (TypeError, ValueError):
-                                _is_free = False
+                            _pricing = _item.get("pricing")
+                            _is_free = False
+                            if (
+                                isinstance(_pricing, dict)
+                                and "prompt" in _pricing
+                                and "completion" in _pricing
+                            ):
+                                try:
+                                    _is_free = (
+                                        float(_pricing["prompt"]) == 0
+                                        and float(_pricing["completion"]) == 0
+                                    )
+                                except (TypeError, ValueError):
+                                    _is_free = False
                             # Also include explicit `:free` suffix variants
                             _is_free = _is_free or _mid.endswith(":free")
                             if not _is_free:
--- a/tests/test_issue1426_openrouter_free_tier_live_fetch.py
+++ b/tests/test_issue1426_openrouter_free_tier_live_fetch.py
@@ -167,10 +167,17 @@ def test_openrouter_free_tier_pricing_fails_closed(
         lambda *_args, **_kwargs: _FakeResponse(fake_payload),
     )
 
-    from hermes_cli import models as hermes_models
-
-    monkeypatch.setattr(hermes_models, "fetch_openrouter_models", lambda **_kwargs: [])
-    monkeypatch.setattr(hermes_models, "provider_model_ids", lambda *_args, **_kwargs: [])
+    # hermes_cli (the agent package) is an optional dependency and is not
+    # installed in the WebUI CI test environment. Force its live-fetch to
+    # return empty when present so the fail-closed static path is exercised;
+    # when absent, that static path is already the only one — mirror the
+    # try/except guard the sibling tests in this file use.
+    try:
+        from hermes_cli import models as hermes_models
+        monkeypatch.setattr(hermes_models, "fetch_openrouter_models", lambda **_kwargs: [])
+        monkeypatch.setattr(hermes_models, "provider_model_ids", lambda *_args, **_kwargs: [])
+    except Exception:
+        pass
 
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Reviewed the diff at `dea5e8a` against `origin/master`, reading `api/config.py` around the OpenRouter free-tier augment block and the new parametrized cases in `tests/test_issue1426_openrouter_free_tier_live_fetch.py`. This looks correct and complete — the fix does exactly what #5880 asked for.

## Code reference

The old expression at `api/config.py:6893-6902` treated any missing/`null`/partial pricing as zero:

```python
_pricing = _item.get("pricing") or {}
try:
    _is_free = (
        float(_pricing.get("prompt", "0") or "0") == 0
        and float(_pricing.get("completion", "0") or "0") == 0
    )
except (TypeError, ValueError):
    _is_free = False
```

The PR replaces this with a gate that requires a real dict with **both** keys present before trusting the numeric comparison:

```python
_pricing = _item.get("pricing")
_is_free = False
if (
    isinstance(_pricing, dict)
    and "prompt" in _pricing
    and "completion" in _pricing
):
    try:
        _is_free = (
            float(_pricing["prompt"]) == 0
            and float(_pricing["completion"]) == 0
        )
    except (TypeError, ValueError):
        _is_free = False
_is_free = _is_free or _mid.endswith(":free")
```

## Why this is right

Walking the four evidence shapes from the issue against the new code: `{}` fails the `"prompt" in _pricing` guard → not free; `{"prompt": "0"}` / `{"completion": "0"}` fail the paired-key guard → not free; `{"prompt": None, "completion": None}` passes the key guard but `float(None)` raises `TypeError` → caught → not free; `{"prompt": "n/a", "completion": "0"}` raises `ValueError` on `float("n/a")` → not free. All four now correctly fall through, and the `_mid.endswith(":free")` line after the block preserves explicit provider designation exactly as required.

## Test coverage

The parametrized `test_openrouter_free_tier_pricing_fails_closed` covers all four ambiguous shapes plus the malformed case, and the isolation is sound: `fetch_openrouter_models` and `provider_model_ids` are both monkeypatched to `[]`, so the free-tier augment is the *only* path a candidate id could enter the picker. Each case asserts `candidate_id not in model_ids` while `explicit_free_id in model_ids`, which pins down both halves of the contract (fail-closed on ambiguity, still accept `:free`). The existing complete-zero-pricing test (`test_openrouter_group_uses_live_fetch_when_available`) stays green, so pricing-derived free models still surface.

## Verification note

I did not execute the suite (cron runs read-only), but the diff and assertions line up with the master behavior I read. One minor optional thought, not a blocker: the two `_is_free = False` initializers are now slightly redundant (the inner one only matters inside the `try`), but that's harmless defensiveness. Ship it.

nesquena-hermes:
Shipped in **exp-v0.52.24** (experimental channel) — thank you @igorroncevic! 🎉

The OpenRouter free-tier detection now fails closed: a candidate is only treated as free when both prompt and completion prices are present and genuinely zero, so a model with missing/`null`/partial pricing metadata is no longer mislabeled "(free)" — while explicit `:free` ids stay selectable (closes #5880).

During the release gate I fixed two CI test-environment issues in the new test (an unguarded optional `hermes_cli` import that failed where the agent package isn't installed, and a `StopIteration` under pytest-shard pollution), matching the sibling-test guard patterns already in that file. Production fail-closed logic verified clean; CI 18/18 green, full suite passed. Thanks for the solid defensive fix.

PR #5885, opened 2026-07-10
fix: focus composer after URL prefill

### LINKED ISSUE #5884: Follow-up to #4961: focus the composer after URL prefill
Follow-up to #4961: the `?q=` / `?prompt=` deep link gets the text into the composer, but it still leaves focus somewhere else. In the browser-search-engine flow that means the page opens with the right prompt visible, then the user has to click into the message box before Enter will actually send it.

## Repro

1. Open `https://<your-hermes-host>/?q=hello+world`
2. Wait for the page to finish booting
3. Press Enter

The composer is prefilled, but the prompt is not submitted because `#msg` never took focus.

## Root cause

The current boot path already has a clean seam for this:

- `static/boot.js` `_applyComposerPrefillOnBoot()` writes the prefilled text into `#msg` and updates the composer UI.
- `static/boot.js` `_finalizeComposerPrefillOnBoot()` consumes the query params and then calls `_applyComposerPrefillOnBoot()`.
- `static/sessions.js` `_consumeComposerPrefillParamsFromLocation()` strips `q`, `prompt`, and `send` from the URL once boot is ready.

What is missing is the last step: after a non-empty prefill is applied, the composer never gets focused. So the deep link is only half-finished for keyboard-driven flows.

## Suggested fix

After `_finalizeComposerPrefillOnBoot()` applies a non-empty prefill, focus `#msg` on the settled composer path so the common flow becomes:

- open Hermes from the browser address bar or a launcher link
- land on the booted page with the prompt already in place
- press Enter immediately

That keeps the current non-destructive behavior, unlike auto-send, but removes the extra click from the deep-link workflow.

If there is concern about stealing focus in unrelated cases, scope it tightly to the query-prefill path only, and only when the prefill text is non-empty.

## Why this is separate from #5759

#5759 adds a general-purpose keyboard shortcut to focus the composer from anywhere in the app. This issue is narrower: URL-prefill boot should finish by putting focus on the composer automatically, because the user just arrived specifically to send that prefilled prompt.

moosnelly:
## Thinking Path
- URL-prefill already waits until the composer is rendered, writes the draft, and refreshes composer controls.
- It stopped one step short: keyboard focus remained elsewhere, so Enter could not submit the prefilled prompt.
- This keeps the existing non-destructive prefill flow and focuses only after non-empty prefill is applied.

## What Changed
- Focus `#msg` after boot applies a non-empty URL prefill.
- Extend the existing #4961 Node regression to prove focus is called for populated drafts and never auto-sends.

## Why It Matters
Browser-search and launcher deep links can now be opened and submitted with Enter immediately, without an extra click.

## Verification
- `./scripts/test.sh tests/test_4961_url_query_prefill.py` — 7 passed
- `./scripts/test.sh tests/test_static_js_runtime_lint.py` — 1 passed
- `npm run lint:runtime` — passed
- `git diff --check` — passed
- Visual layout is unchanged; focused browser E2E/screenshot evidence was not captured locally because no headless browser runtime is installed. CI browser smoke remains the runtime gate.

## Risks / Follow-ups
- Scope is limited to non-empty boot prefill. It does not auto-send, change URL cleanup, or affect ordinary page loads.
- Release note: URL-prefilled composer links now focus the message box so Enter submits immediately.

## Model Used
OpenAI `gpt-5.6-terra` via Hermes Agent; targeted repository inspection, Node-backed regression tests, ESLint runtime guard, and GitHub API tooling.

Closes #5884

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -130,6 +130,7 @@ async function _applyComposerPrefillOnBoot(prefillIntent){
   msg.value=text;
   if(typeof autoResize==='function') autoResize();
   else if(typeof updateSendBtn==='function') updateSendBtn();
+  if(typeof msg.focus==='function') msg.focus();
 }
 async function _finalizeComposerPrefillOnBoot(prefillIntent){
   if(prefillIntent&&prefillIntent.hasParams&&typeof _consumeComposerPrefillParamsFromLocation==='function'){
--- a/tests/test_4961_url_query_prefill.py
+++ b/tests/test_4961_url_query_prefill.py
@@ -185,8 +185,8 @@ def test_apply_prefill_updates_composer_without_autosending():
     source = _node_prelude() + """
 (async () => {
   evalBoot('_applyComposerPrefillOnBoot');
-  const counts = { autoResize: 0, updateSendBtn: 0, send: 0 };
-  const msg = { value: '' };
+  const counts = { autoResize: 0, updateSendBtn: 0, focus: 0, send: 0 };
+  const msg = { value: '', focus() { counts.focus++; } };
   global.document = {
     getElementById(id) {
       return id === 'msg' ? msg : null;
@@ -207,7 +207,7 @@ def test_apply_prefill_updates_composer_without_autosending():
 });
 """
     payload = json.loads(_run_node(source))
-    assert payload["counts"] == {"autoResize": 1, "updateSendBtn": 1, "send": 0}
+    assert payload["counts"] == {"autoResize": 1, "updateSendBtn": 1, "focus": 2, "send": 0}
     assert payload["value"] == "second pass"
 
 


nesquena-hermes:
Pulled the branch and read `static/boot.js` around the prefill helpers plus the updated test. This looks correct and low-risk.

## What it does

The one-line add lands at the end of `_applyComposerPrefillOnBoot` (`static/boot.js:133`), after the value is written and the composer controls are refreshed:

```js
msg.value=text;
if(typeof autoResize==='function') autoResize();
else if(typeof updateSendBtn==='function') updateSendBtn();
if(typeof msg.focus==='function') msg.focus();   // ← added
```

The early `return` at the top (`if(!prefillIntent||!prefillIntent.hasText) return;`) already guards the empty-prefill case, so focus only fires when a non-empty draft was actually applied — matching the PR's stated "never on ordinary page loads" scope. Good.

## Consistency check

This matches how the rest of the codebase grabs the composer — unconditional `$('msg').focus()` with no viewport guard is the established pattern (`static/boot.js:1764`, `:1768`, `:1770` in the new-chat path; `static/ui.js:499` in compression-recovery; `static/commands.js:819`). So there's nothing inconsistent about doing the same after a deep-link prefill, and the `typeof msg.focus==='function'` guard is a nice touch for the Node test harness where `msg` is a bare object.

## Test

The `focus: 2` assertion in `test_apply_prefill_updates_composer_without_autosending` is right, not a typo: the test calls `_applyComposerPrefillOnBoot` three times, and focus increments only on the two `hasText:true` passes (`autoSend:true` then `autoSend:false`); the third `hasText:false` call returns early before focus. `autoResize:1`/`updateSendBtn:1` differ because `global.autoResize` is deleted between the first and second pass, exercising the `else if` branch — so those counters split while focus counts both. The `send:0` assertion preserves the #4961 no-autosend invariant.

One thing the PR is honest about and I'd echo: the value-write + focus path isn't covered by a real browser E2E here, only the Node unit harness, so the CI browser smoke remains the actual runtime gate for "Enter submits." Nothing blocking from my read — the change is minimal, correctly scoped, and consistent with existing focus handling.

nesquena-hermes:
Closing as already-fixed 🙏 — the exact fix you proposed (focusing the composer after a non-empty URL prefill so Enter can submit) is already on `master`: `_applyComposerPrefillOnBoot()` in `static/boot.js` calls `if(typeof msg.focus==='function') msg.focus();` right after applying the draft (shipped via commit 2e39bb841, `fix(#5884): focus composer after URL prefill`), and the test suite already asserts the focus + `activeElement` behavior.

Your diagnosis was spot-on and matches what landed — the prefill flow was one step short on keyboard focus. Since the behavior is already in place (verified: byte-identical focus call), there's nothing left to merge here, so I'm closing this in favor of the shipped fix. Thanks for the contribution and the clear write-up! If you spot any prefill case where focus still doesn't land, please open a fresh issue with the URL params.

PR #5887, opened 2026-07-10
fix(#5884): focus composer after URL prefill

### LINKED ISSUE #5884: Follow-up to #4961: focus the composer after URL prefill

rodboev:
## Thinking Path

- URL query prefill already parses `q` and `prompt`, preserves explicit session selection, populates the composer, and deliberately avoids auto-send.
- The boot helper stops after refreshing composer UI state, which leaves the prefilled textarea visible but keyboard-inert.
- Focusing the resolved composer inside that same non-empty prefill seam completes the intended browser-search flow without changing ordinary boot behavior.

## What Changed

- `static/boot.js`: focus the composer after applying a non-empty URL prefill and refreshing its UI state.
- `tests/fixtures/webui-PR-TARGET-5884-REPRO.md`: vendor the captured issue reproduction into the checkout so the regression test stays artifact-faithful under CI and fresh clones.
- `tests/test_4961_url_query_prefill.py`: cover non-empty focus, the no-`focus()` fallback, empty and absent negative space, URL cleanup timing, no auto-send, and explicit session preservation with the existing browserless JS harness.

## Why It Matters

Opening Hermes from a browser search engine or launcher with a prefilled prompt now leaves the composer ready for Enter immediately. Ordinary boots keep their current focus behavior, and URL prefill remains non-destructive.

## Verification

Focused browserless tests cover `q` and `prompt` focus, the guarded no-`focus()` fallback, empty and absent prefill guards, URL cleanup timing, no auto-send, and explicit session selection; CI runs the full matrix on Python 3.11, 3.12, and 3.13.

## Upstream

Closes #5884.

Follow-up requested in https://github.com/nesquena/hermes-webui/issues/4961#issuecomment-4934504837.

## Model Used

GPT-5 via Codex CLI

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -130,6 +130,7 @@ async function _applyComposerPrefillOnBoot(prefillIntent){
   msg.value=text;
   if(typeof autoResize==='function') autoResize();
   else if(typeof updateSendBtn==='function') updateSendBtn();
+  if(typeof msg.focus==='function') msg.focus();
 }
 async function _finalizeComposerPrefillOnBoot(prefillIntent){
   if(prefillIntent&&prefillIntent.hasParams&&typeof _consumeComposerPrefillParamsFromLocation==='function'){
--- a/tests/fixtures/webui-PR-TARGET-5884-REPRO.md
+++ b/tests/fixtures/webui-PR-TARGET-5884-REPRO.md
@@ -1,0 +1,7 @@
+Reproduction source: issue-artifact copied from https://github.com/nesquena/hermes-webui/issues/5884
+
+1. Open `https://<your-hermes-host>/?q=hello+world`
+2. Wait for the page to finish booting
+3. Press Enter
+
+Observed on current `origin/master` `d7c3c2b7`: the composer is prefilled, but the prompt is not submitted because `#msg` never took focus.
--- a/tests/test_4961_url_query_prefill.py
+++ b/tests/test_4961_url_query_prefill.py
@@ -1,5 +1,6 @@
 """Regression tests for #4961 URL query composer prefill behavior."""
 import json
+import re
 import shutil
 import subprocess
 from pathlib import Path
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.25** (experimental channel) — thank you @rodboev! 🎉

Opening Hermes from a browser-search / launcher deep-link (`?q=`/`?prompt=`) now focuses the composer after the prefill lands, so Enter sends immediately without a click (closes #5884). The focus is guarded to the non-empty prefill seam (ordinary boots unchanged) and never auto-sends.

Gate: Codex review SAFE TO SHIP, full suite passed, CI 18/18 green, and I drove it live — `?q=...` → composer prefilled + focused (`document.activeElement === #msg`) + Send enabled + URL cleaned. Nice, tight fix with thorough regression coverage.

PR #5888, opened 2026-07-10
WebUI: keep read-only child sessions ordered after writable children

igorroncevic:
## Summary
This is the narrow Hermes-WebUI follow-up for ordered session children.

- In `static/sessions.js`, attached child sessions are sorted so read-only sessions render after writable continuation/fork-style children under the same parent.
- The expanded sidebar child list now preserves the attached child order instead of re-sorting by timestamp at render time.
- Added a regression test for read-only-last ordering and the expanded-render order guard.

## Scope and ownership
- Backend semantics for child classification and ordering are implemented upstream in `NousResearch/hermes-agent#61583`.
- This PR does **not** add local child-type classification heuristics; it only preserves backend/attachment order in the sidebar render path.
- Existing flicker/fallback/stale-parent behavior from previous WebUI lineage work is intentionally preserved.
- The branch has been rebuilt on `nesquena/master`; unrelated GPT-5.6/OpenRouter files are no longer part of this PR diff.

## PR linkage
- Follows and depends on: https://github.com/NousResearch/hermes-agent/pull/61583
- This PR should be reviewed/merged after the backend ordering PR is accepted.

## Verification
- `./scripts/test.sh tests/test_session_lineage_collapse.py -q` — 54 passed
- `npm run lint:runtime`
- `git diff --check`

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -6328,6 +6328,8 @@ function _attachChildSessionsToSidebarRows(collapsedRows, rawSessions, rawRefere
   };
   const orphans=[];
   const renderableChildIds=new Set((rawSessions||[]).map(s=>s&&s.session_id).filter(Boolean));
+  const childAttachOrderById=new Map();
+  let childAttachCursor=0;
   const attachQueueById=new Map();
   for(const candidate of [...(rawSessions||[]),...(referenceSessions||[])]){
     if(candidate&&candidate.session_id&&!attachQueueById.has(candidate.session_id)) attachQueueById.set(candidate.session_id,candidate);
@@ -6382,6 +6384,9 @@ function _attachChildSessionsToSidebarRows(collapsedRows, rawSessions, rawRefere
     }
     if(parentRow){
       const childCopy={...child};
+      if(!childAttachOrderById.has(childCopy.session_id)){
+        childAttachOrderById.set(childCopy.session_id, childAttachCursor++);
+      }
       if(parentSegment){
         childCopy._parent_segment_id=parentSegment.session_id;
         childCopy._parent_segment_title=_sessionDisplayTitle(parentSegment)||child.parent_title||'Untitled';
@@ -6408,6 +6413,23 @@ function _attachChildSessionsToSidebarRows(collapsedRows, rawSessions, rawRefere
       // branch above and still orphans as before.
       if(child&&child._cross_surface_child_session&&_isChildSession(child)) continue;
       orphans.push({...child,_orphan_child_session:true});
+    }
+  }
+  const resolveReadOnlySession = typeof _isReadOnlySession === 'function'
+    ? _isReadOnlySession
+    : ((session) => !!(session && session.read_only));
+  for(const row of rows){
+    if(Array.isArray(row._child_sessions)&&row._child_sessions.length>1){
+      row._child_sessions.sort((a,b)=>{
+        const readOnlyCmp = Number(resolveReadOnlySession(a))-Number(resolveReadOnlySession(b));
+        if(readOnlyCmp!==0) return readOnlyCmp;
+        const aOrder = childAttachOrderById.get(a&&a.session_id) || 0;
+        const bOrder = childAttachOrderById.get(b&&b.session_id) || 0;
+        return aOrder-bOrder;
+      });
+    }
+    if(Array.isArray(row._child_sessions)){
+      row._child_session_count=row._child_sessions.length;
     }
   }
   return [...rows,...orphans];
@@ -7523,7 +7545,7 @@ function renderSessionListFromCache(){
       const childList=document.createElement('div');
       childList.className='session-child-sessions';
       ['pointerdown','pointerup','click','touchstart','touchmove','touchend','touchcancel'].forEach(ev=>childList.addEventListener(ev,e=>e.stopPropagation()));
-      const sortedChildren=[...s._child_sessions].sort((a,b)=>_sessionTimestampMs(b)-_sessionTimestampMs(a));
+      const sortedChildren=[...s._child_sessions];
       const openChildSession=async(childSession)=>{
         // #5409: close mobile sidebar synchronously before navigation
         if(typeof closeMobileSidebar==='function')closeMobileSidebar();
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -6423,8 +6423,8 @@ function _attachChildSessionsToSidebarRows(collapsedRows, rawSessions, rawRefere
       row._child_sessions.sort((a,b)=>{
         const readOnlyCmp = Number(resolveReadOnlySession(a))-Number(resolveReadOnlySession(b));
         if(readOnlyCmp!==0) return readOnlyCmp;
-        const aOrder = childAttachOrderById.get(a&&a.session_id) || 0;
-        const bOrder = childAttachOrderById.get(b&&b.session_id) || 0;
+        const aOrder = childAttachOrderById.get(a&&a.session_id) ?? Number.MAX_SAFE_INTEGER;
+        const bOrder = childAttachOrderById.get(b&&b.session_id) ?? Number.MAX_SAFE_INTEGER;
         return aOrder-bOrder;
       });
     }
--- a/tests/test_session_lineage_collapse.py
+++ b/tests/test_session_lineage_collapse.py
@@ -554,6 +554,85 @@ console.log(JSON.stringify(rows));
     assert rows[0]["_child_sessions"][0]["session_id"] == "subagent_child"
 
 
+def test_read_only_child_sessions_are_stacked_after_writable_children():
+    """Read-only delegated children should render after fork/writable children."""
+    js = SESSIONS_JS_PATH.read_text(encoding="utf-8")
+    source = f"""
+const src = {js!r};
+function extractFunc(name) {{
+  const re = new RegExp('function\\\\s+' + name + '\\\\s*\\\\(');
+  const start = src.search(re);
+  if (start < 0) throw new Error(name + ' not found');
+  let i = src.indexOf('{{', start);
+  let depth = 1; i++;
+  while (depth > 0 && i < src.length) {{
+    if (src[i] === '{{') depth++;
+    else if (src[i] === '}}') depth--;
+    i++;
+  }}
+  return src.slice(start, i);
+}}
+eval(extractFunc('_isReadOnlySession'));
+eval(extractFunc('_isChildSession'));
+eval(extractFunc('_isForkWithResolvableParent'));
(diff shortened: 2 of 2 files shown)


igorroncevic:
Backend parent PR: https://github.com/NousResearch/hermes-agent/pull/61583.
This change intentionally stays client-side narrow: no new child-type heuristics, only ordering behavior for attached child rows.

nesquena-hermes (changes requested):
Thanks @igorroncevic — the feature itself (sort read-only child sessions after writable ones in `static/sessions.js`) reads correctly and uses `_isReadOnlySession` with a sensible fallback. But **this PR's own tests fail deterministically**, so it can't ship as-is.

### The blocker: 3 tests in `tests/test_session_lineage_collapse.py` fail (verified master-green / PR-head-red)

Running the file at this PR's head (`d186bc498a97`):

```
3 failed, 51 passed
FAILED test_read_only_child_sessions_are_stacked_after_writable_children   ← the PR's NEW feature test
FAILED test_cross_surface_subagent_child_stacks_under_visible_fork_parent  ← pre-existing, passes on master
FAILED test_parent_source_label_display_text_does_not_force_external_orphaning ← pre-existing, passes on master
```

The two pre-existing tests **pass on `origin/master`** (`2 passed`), so this PR reddens them.

### Root cause — the refactored test harness under-escapes the regex

The three failing tests are exactly the three you refactored from the f-string harness (`f"""... {js!r} ..."""`) to a concatenated plain-string harness. Every failure is the same JS error:

```
SyntaxError: Invalid regular expression: /functions+_isChildSessions*(/: Unterminated group
    at extractFunc ([stdin]:3:14)
```

In the concatenated form the line is:

```python
"  const re = new RegExp('function\\s+' + name + '\\s*\\(');\n"
```

`\\s` in a normal (non-raw) Python string is **one** backslash, so the emitted JS source is `RegExp('function\s+' ...)`. JS then drops the lone backslash in the string literal (`'\s'` → `s`), producing the regex `/functions+_isChildSessions*(/` — the `(` is an unterminated group, so `extractFunc` throws **before the feature is ever exercised**. That means the new ordering behavior is currently untested even though the source change looks right.

### Fix (either works)

- **Simplest:** revert these three tests' harness back to the f-string `{js!r}` form the other 51 tests in the file already use (they escape correctly).
- **Or** if you keep the concatenated-string form, double the backslashes so the JS source ends up with `\\s` (which JS parses to `\s`):

```python
"  const re = new RegExp('function\\\\s+' + name + '\\\\s*\\\\(');\n"
```

After fixing the escaping, please confirm `tests/test_session_lineage_collapse.py` is fully green — that will also actually exercise your new `test_read_only_child_sessions_are_stacked_after_writable_children` assertion.

(Note: this PR is stacked on #5879 — the diff unique to this PR is `static/sessions.js` + the `test_session_lineage_collapse.py` refactor.)

igorroncevic:
Addressed the blockers from the latest reviews in commit `2f5c911e`.

What changed:
- Fixed the expanded sidebar path: child rows now render in the already-attached `_child_sessions` order instead of re-sorting by timestamp, so read-only delegates cannot jump above writable/fork children when expanded.
- Reverted the unnecessary test-harness refactor and kept the new regression in the existing f-string harness style, resolving the deterministic regex escaping failures.
- Rebuilt the PR branch directly on current `nesquena/master`; the PR diff is now only:
  - `static/sessions.js`
  - `tests/test_session_lineage_collapse.py`

That also removes the stale Greptile OpenRouter/GPT-5.6 finding from this PR's diff; those files are no longer touched here.

Verification:
- `./scripts/test.sh tests/test_session_lineage_collapse.py -q` — 54 passed
- `node --check static/sessions.js` — passed
- `npm run lint:runtime` — passed
- `git diff --check` — passed

nesquena-hermes:
## Re-review at `2f5c911e` — blockers resolved ✅

Pulled the new head and re-read both files against `origin/master`. The two things I flagged last round are genuinely fixed.

### 1. Test-harness regex escaping — fixed

The new `test_read_only_child_sessions_are_stacked_after_writable_children` is back in the f-string `{js!r}` form, and its `extractFunc` line (`tests/test_session_lineage_collapse.py:563`) now reads:

```python
const re = new RegExp('function\\\\s+' + name + '\\\\s*\\\\(');
```

That's byte-identical to the 45 other passing tests in the file (same `\\\\s` quadruple-backslash → JS `\s`). No more `Unterminated group` — the assertion actually exercises `_attachChildSessionsToSidebarRows` now. I traced the expected order by hand: raw order is `read_only`(attach 0), `fork`(1), `writable`(2); read-only-last then attach-order tie-break yields `[fork_child, writable_subagent, read_only_subagent]`, which is what the test asserts. Note this genuinely differs from the old timestamp-desc sort (by timestamp, `read_only` at 1700000030 would sort ahead of `writable` at 1700000020), so the test is validating the new behavior, not the incidental old one. Good.

### 2. Diff is clean

`git diff origin/master...HEAD --stat` is now just `static/sessions.js` + `tests/test_session_lineage_collapse.py` — the stacked #5879 OpenRouter/GPT-5.6 files are gone, so the stale Greptile finding no longer applies here.

### On the render path change

Removing the timestamp re-sort at `static/sessions.js:7548` (`const sortedChildren=[...s._child_sessions];`) is the right call — the ordering is now decided once in `_attachChildSessionsToSidebarRows` (`:6422-6430`) and the expanded render just honors it, so collapsed and expanded views can't disagree. I checked the other `_child_sessions` consumer at `:7181-7186` (the fork-visibility id collector) and it's order-independent, so nothing else regresses.

### One minor, non-blocking note

The tie-break in the sort:

```js
const aOrder = childAttachOrderById.get(a&&a.session_id) || 0;
```

`|| 0` collapses both "attach index 0" and "not in the map (undefined)" to the same key. In practice every attached child is registered at `:6386` before it lands in `_child_sessions`, so `.get()` should never be undefined here — but if you want to be strict, `?? Number.MAX_SAFE_INTEGER` for the undefined case would keep an un-registered child from silently tying with the genuine first child. Purely defensive; not a blocker.

Escaping fixed, diff scoped, ordering centralized, and the new assertion is real. This looks good to me now, modulo the backend dependency (hermes-agent#61583) landing so children actually carry `read_only`. Since `_isReadOnlySession` already reads `read_only`/`is_read_only` which exist today, the WebUI side is a safe no-op until then. Nice cleanup.

igorroncevic:
Addressed @nesquena-hermes's defensive tie-break note from https://github.com/nesquena/hermes-webui/pull/5888#issuecomment-4938974712.

- Changed the child attach-order fallback from `|| 0` to `?? Number.MAX_SAFE_INTEGER`, preserving real attach index `0` while sending any unexpected unregistered child after registered children.
- Kept the WebUI PR narrow: no local child-type heuristics or backend-owned ordering changes.

Verification:
- `./scripts/test.sh tests/test_session_lineage_collapse.py -q` — 54 passed
- `node --check static/sessions.js` — passed
- `npm run lint:runtime` — passed
- `git diff --check` — passed

Pushed commit `037c6cf6`.

igorroncevic:
@nesquena-hermes could you please re-review/approve or clarify what is still blocking merge?

Current state from my side:
- Your previous blockers were addressed and re-reviewed as resolved in https://github.com/nesquena/hermes-webui/pull/5888#issuecomment-4938974712.
- The defensive tie-break note was also addressed in `037c6cf6`.
- GitHub still shows the old `CHANGES_REQUESTED` state, while CI is green and Greptile says no blocking issues.
- The branch is `MERGEABLE`; the only remaining state I see is `BEHIND`, not a conflict.

If this needs an update-branch/rebase before approval/merge, or if the backend dependency must land first, please confirm and I’ll handle the right next step.

igorroncevic:
@nesquena-hermes The requested follow-ups are complete on the current head (`037c6cf67a`): the expanded-render ordering fix and the defensive `?? Number.MAX_SAFE_INTEGER` tie-break are both in place.

The latest validation is green: browser-smoke, lint, the full Python 3.11/3.12/3.13 matrix, Greptile, the 54-test focused suite, runtime lint, and `git diff --check`. This WebUI PR remains intentionally narrow and follows Agent #61583 for the backend child-ordering semantics. Please re-review this head and confirm the merge order/dependency when you have a chance.

nesquena-hermes:
## 🔬 Exact-head gate certification — GREEN ✅

Certified head: `037c6cf67a9c1b3454afacbf3c3f1ad19abc8913`
Frozen master / gate integration: `ecc47782e3e72a5bf3e848a8f09a2a5a838c71b1` / `4bbd2241619561c9f853b6ef93ff65e6ade14418`

### Evidence

- Exact head + mandatory threat gate: CLEAN; all hosted checks complete/green.
- Focused lineage/sidebar/ACP suite: **77 passed**.
- Full serial suite: **0 PR-only failures/errors** vs the same frozen-master baseline; one net additional pass.
- Ruff-forward: CLEAN. ESLint runtime: CLEAN. Browser smoke: CLEAN.
- Codex: **SAFE TO SHIP**; independent 24-permutation comparison preserved attach order within writable/read-only classes and found no unrelated behavioral delta.
- Backend advisor: **APPROVE**; confirmed stable comparator, missing-order safety, collapsed/expanded consistency, `read_only` + `is_read_only` compatibility, and no-op behavior when metadata is absent.
- Fable UX: **SHIP-UX**.
- Real desktop + iPhone-14 browser proof: mixed child rows render in this order despite newer read-only timestamps: writable continuation fork → writable implementation delegate → read-only research delegate → read-only validation replay. No clipping/overlap; normal mobile ellipsis only. Evidence sent to the maintainer (Telegram 9990–9991).

The upstream Agent PR is still open, but that does not block this WebUI change: current WebUI backend normalization already emits `read_only: true` for retained delegated children, and missing metadata preserves existing order.

Non-blocking follow-up: read-only child rows still have no explicit badge/tooltip; they are only distinguished by missing writable actions. That is pre-existing and this PR improves grouping without worsening it.

**Exact-head verdict: GREEN.**

nesquena-hermes:
Shipped in **exp-v0.52.275** — thank you @igorroncevic! 🎉

Read-only child sessions now stay grouped after their writable siblings within a collapsed session lineage in the sidebar, so the ordering no longer interleaves confusingly.

Re-gated on current master before release:
- **Codex**: SAFE TO SHIP — comparator verified stable and finite (no crash/NaN on missing order keys), both `read_only` and `is_read_only` handled, all 24 ordering permutations pass, expanded rendering preserves attached order, selection stays keyed by `session_id` (independent of row position).
- **Full suite**: green (the only failures were two pre-existing box-environment artifacts unrelated to this diff — same failures reproduce on clean master).
- Focused lineage suite: 54 passed.

Credited via `Co-authored-by` in the release commit.

PR #5889, opened 2026-07-10
Fix stale session loads contaminating the active pane

starship-s:
## Thinking Path

The active session pane is shared frontend state, but `loadSession()` previously used only the destination `session_id` as async ownership. That prevents a response for session A from winning after a switch to session B, but it does not distinguish overlapping reloads of the same session, and `_ensureMessagesLoaded()` mutated transcript/tool/truncation globals after its awaited API call without revalidating ownership.

A delayed transcript response could therefore populate `S.messages` after a newer navigation had taken ownership. The newer load could then see the non-empty array and skip fetching its own transcript, producing a split pane: current-session metadata with another session's rows.

This change assigns each `loadSession()` invocation a monotonically increasing generation and validates both session ID and generation at async boundaries. This is intentionally narrower than text-based deduplication or a broad session-state rewrite: ownership is attached to the operation that may mutate the pane.

## What Changed

- assign a generation token to every `loadSession()` invocation
- replace session-ID-only stale checks with invocation ownership checks across metadata, profile resolution, error, unread, draft, and completion paths
- pass the generation into `_ensureMessagesLoaded()` and reject stale transcript responses before they mutate shared state
- stop stale successful and rejected INFLIGHT transcript continuations before they merge rows, tool calls, or stream state
- add executable Node regression coverage for:
  - the observed idle-session response ordering
  - a stale INFLIGHT transcript response after the destination session has completed loading
  - a rejected stale same-session reload
- update existing source-contract tests to assert generation ownership instead of the superseded session-ID-only guard

## Screenshots

Viewport: `1920×947`, device pixel ratio `1`.

### Before

Before: Project Atlas header with Project Beacon transcript rows

A controlled response delay reproduces the mismatch: the pane header identifies Project Atlas while the rendered transcript contains Project Beacon rows.

### After

After: Project Atlas header with Project Atlas transcript rows

Under the same response ordering, the Project Atlas header and all rendered rows remain consistent.

## Verification

- `python3 -m py_compile tests/test_cross_session_message_load_isolation.py tests/test_issue3993_session_load_self_heal.py tests/test_issue5177_hidden_tab_blank_gap.py tests/test_issue856_background_completion_unread.py tests/test_issue_new_chat_draft_restore.py`
- `./scripts/test.sh tests/test_cross_session_message_load_isolation.py tests/test_session_switch_busy_race.py tests/test_inflight_send_start_race.py tests/test_issue3993_session_load_self_heal.py tests/test_issue5177_hidden_tab_blank_gap.py tests/test_issue856_background_completion_unread.py tests/test_issue_new_chat_draft_restore.py tests/test_regressions.py -k 'inflight or cross_session or stale_load or ensure_messages_loaded or completion_unread or draft_flush'` — 52 passed
- `./scripts/test.sh` — `12,511 passed, 100 skipped, 1 xfailed, 2 xpassed`; six failures, all reproduced on pristine `origin/master` under the same environment
- the remaining six full-suite failures reproduce on a detached pristine `origin/master` worktree under the same environment:
  - `tests/test_issue1094_provider_bugs.py::TestBug1094Endpoints::test_get_providers_after_delete`
  - `tests/test_issue1426_openrouter_free_tier_live_fetch.py::test_free_tier_cap_prevents_picker_drowning`
  - two `tests/test_issue4164_bound_non_git_project_context_walk.py` cases
  - two `tests/test_tls_aware_probe.py` cases
- `git diff --check`
- browser smoke: real WebUI/API responses with controlled network ordering, before and after

## Risks / Follow-ups

- no backend, persistence, API, or schema changes
- the generation is page-local and monotonic; it exists only to establish frontend async ownership

## Model Used

- OpenAI / `gpt-5.6-sol` via Hermes Agent for investigation, reproduction, orchestration, and final verification
- OpenAI / `gpt-5.3-codex-spark` via OpenCode for implementation and regression tests
- Google / `gemini-3.1-pro-preview` via OpenCode for independent review

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -1425,7 +1425,7 @@ async function loadSession(sid){
   // #2971: idempotent re-arm before the no-op guard revives a stream a prior
   // failed loadSession killed; no-ops on real switches.
   _rearmActiveSessionStream();
-  if(currentSid===sid && !forceReload && !_loadingSessionId) return;
+  if(currentSid===sid && !forceReload && (!_loadingSessionId || _loadingSessionId===sid)) return;
   // Mark this session as the in-flight load. Subsequent loadSession() calls
   // will overwrite this; stale awaits use the mismatch to bail out (#1060).
   const _loadGeneration = ++_loadSessionGeneration;
@@ -2778,7 +2778,7 @@ async function _ensureMessagesLoaded(sid, opts) {
       {timeoutMs:120000}
     );
   } finally {
-    _clearSameSessionForceReloadHint(sid);
+    if (_ownsLoad()) _clearSameSessionForceReloadHint(sid);
   }
   if (!_ownsLoad()) return;
   // Guard: api() may have redirected (401) and returned undefined.
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -19,6 +19,11 @@ const ICONS={
 // responses from in-flight requests when the user switches sessions again
 // before the first request completes (#1060).
 let _loadingSessionId = null;
+// Each loadSession() invocation gets a monotonically increasing generation.
+// `_loadingSessionId` only tracks destination session_id, so same-session
+// concurrent loads can still race and overwrite each other unless we compare
+// the generation token as well.
+let _loadSessionGeneration = 0;
 // #3306: Snapshot of S.messages captured by loadSession() right before it
 // clears them on a force-reload of the active session. Consumed by
 // _ensureMessagesLoaded() when calling _carryForwardEphemeralTurnFields so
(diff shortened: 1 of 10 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.24** (experimental channel) — thank you @starship-s! 🎉

This fixes the cross-session pane contamination: each `loadSession()` now carries a generation token and every post-`await` pane mutation re-checks ownership (session id + generation) before writing, so a stale late response can't paint the wrong conversation into the active pane.

During the release gate, the authoritative Codex review reproduced two SILENT regressions the change introduced — a redundant same-session click cancelling the authoritative in-flight refresh, and an unconditional transcript-width-hint cleanup collapsing long sessions to their tail. Both were fixed (one line each) on the branch and you converged the matching source-contract test updates. Re-gate came back SAFE TO SHIP, full suite 12,612 passed, CI 18/18 green, and the before/after screenshots verified the fix. Nice catch on a subtle shared-state race.

Issue #1948, opened 2026-05-09
bug(models): custom provider ids containing host:port mis-resolve provider-qualified model values


AmagiDDmxh:
## Bug Description

WebUI can mis-resolve models for named custom providers whose generated provider id contains additional colons, such as a custom provider name/slug that includes a `host:port` pair.

This is related to, but distinct from, #1776. That issue covers model IDs with colon suffixes such as `:free`; this case is about the **custom provider id itself** containing an extra colon before the model separator.

## Steps to Reproduce

Use a sanitized custom provider config similar to:

```yaml
model:
  provider: custom
  default: example-model
  base_url: https://api.example.test:8443/v1
  api_key: sk-redacted

custom_providers:
  - name: Example.test:8443
    base_url: https://api.example.test:8443/v1
    api_key: sk-redacted
    model: example-model
```

Then in WebUI:

1. Start/restart WebUI.
2. Let `/api/models` build a custom provider group for the named custom provider.
3. Create a new session from the browser UI with the default model.
4. Send a message.

## Expected Behavior

The selected model should resolve as:

```text
provider = custom:example-test-8443  # or any unambiguous canonical provider id
model    = example-model
```

or otherwise preserve the intended provider id and send the bare configured model ID to the custom endpoint.

## Actual Behavior

When the generated provider id contains a colon, WebUI can construct a provider-qualified model value shaped like:

```text
@custom:example.test:8443:example-model
```

The current parser path that uses `rsplit(':', 1)` treats the final colon before the model as the separator, producing a wrong split such as:

```text
provider = custom:example.test
model    = 8443:example-model
```

The backend then sends `8443:example-model` to the custom endpoint, which returns an error like:

```text
Model not found: Error code: 400 - {'detail': 'Model not found'}
```

## Notes / Evidence

A local workaround is to rename the custom provider so its generated provider id does not contain a `host:port` colon, for example:

```yaml
custom_providers:
  - name: Example Test 8443
```

After clearing WebUI's model cache/session state and restarting, the generated provider id no longer contains the extra colon, new browser-created sessions resolve correctly, and the same model works.

CLI usage with the same custom endpoint and bare model can work, which suggests the failure is in WebUI provider-qualified model parsing/state rather than the model endpoint itself.

## Environment

- Hermes WebUI: v0.51.30
- Hermes Agent: v0.13.0 / current main at time of report
- OS: macOS
- Provider type: custom OpenAI-compatible endpoint

## Suggested Direction

[...]

PR #5891, opened 2026-07-10
fix: resolve custom provider colon-parsing corruption in model selection

tkuhlengel:
When a custom provider (e.g. `custom:litellm-proxy`) serves a model whose name contains colons (e.g. `qwen3.6:27b-vision`), the `@provider:model` dropdown value becomes `@custom:litellm-proxy:qwen3.6:27b-vision`. The `rsplit(":", 1)` logic in `resolve_model_provider()` was eating part of the model name into the provider ID, producing `"custom:litellm-proxy:qwen3.6"` and `"27b-vision"` instead of the correct split.

This fix has three parts:

1. **Frontend (`static/ui.js`)**: Reorder `_modelStateForSelect()` to prefer the option's `data-provider` attribute over string-splitting. The `data-provider` attribute carries the exact provider ID set by the backend, avoiding colon-parsing entirely. Also improve `_providerFromModelValue()` to try known custom provider slugs before falling back to `lastIndexOf`.

2. **Frontend (`static/ui.js`)**: Populate `window._customProviderSlugs` from the `/api/models` response so `_providerFromModelValue()` can match against known provider IDs when the select element is unavailable.

3. **Backend (`api/config.py`)**: Convert the single peel-back step in `resolve_model_provider()` into a loop that peels segments until the provider hint matches a known custom provider slug. This correctly handles the `:free/:thinking` suffix variant that previously over-corrected.

**Closes:** #1948, #2047  
**Fixes:** cron job provider corruption, settings provider corruption

**Tests:** All 7 new regression tests pass, plus 59 existing model/cron-related tests.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -2687,8 +2687,17 @@ def resolve_model_provider(model_id: str) -> tuple:
         inner = model_id[1:]
         provider_hint, bare_model = inner.rsplit(":", 1)
         if provider_hint.startswith("custom:") and provider_hint.count(":") >= 2:
-            _slug_rest = provider_hint[len("custom:"):]
-            if not _custom_slug_rest_looks_like_host_port(_slug_rest):
+            # The provider hint has too many colons — part of the model name
+            # may have been eaten by rsplit. Peel segments back until we
+            # match a known custom provider slug, or until we're down to a
+            # single "custom:<slug>" with one colon.
+            _known_slugs = _named_custom_provider_slugs(cfg)
+            while provider_hint.count(":") >= 2:
+                _slug_rest = provider_hint[len("custom:"):]
+                if _custom_slug_rest_looks_like_host_port(_slug_rest):
+                    break  # host:port slug — don't peel further
+                if provider_hint in _known_slugs:
+                    break  # matches a known custom provider — correct split
                 provider_hint, extra = provider_hint.rsplit(":", 1)
                 bare_model = f"{extra}:{bare_model}"
         elif (provider_hint not in _PROVIDER_MODELS
--- a/static/ui.js
+++ b/static/ui.js
@@ -2639,8 +2639,33 @@ function _getOptionProviderId(opt){
 }
 function _providerFromModelValue(modelId){
   const value=String(modelId||'').trim();
-  if(value.startsWith('@')&&value.includes(':')) return value.slice(1,value.lastIndexOf(':'));
-  return '';
+  if(!value.startsWith('@')||!value.includes(':')) return '';
+  // Strip the leading @, then try to find the provider boundary.
+  // For non-custom providers (e.g. @openrouter:model), split on the first colon.
+  // For custom providers (custom:slug), the slug may contain colons (e.g.
+  // custom:litellm-proxy or custom:10.8.71.41:8080), so we can't use
+  // lastIndexOf(':') — that would eat part of the model name when the model
+  // itself contains colons (e.g. qwen3.6:27b-vision).
+  // Strategy: try known custom provider slugs from the global model catalog
+  // first (window._customProviderSlugs), then fall back to lastIndexOf.
+  const inner=value.slice(1);
+  // Check if we have known custom provider slugs available
+  if(typeof window!=='undefined'&&Array.isArray(window._customProviderSlugs)){
+    for(const slug of window._customProviderSlugs){
+      if(inner.startsWith(slug+':')){
+        return slug;
+      }
+    }
+  }
+  // Non-custom providers: split on first colon
+  if(!inner.startsWith('custom:')){
+    return inner.slice(0,inner.indexOf(':'));
+  }
+  // Custom provider without slug list: fall back to lastIndexOf.
+  // This is the imperfect legacy path — it works when the model name
+  // has no colons, but may mis-split when it does. The option-based
+  // lookup in _modelStateForSelect should be preferred.
+  return inner.slice(1,inner.lastIndexOf(':'));
 }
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Thanks for this — I pulled the branch and read both the diff and the current `origin/master` end to end. The core diagnosis is correct: the `@custom:<slug>:<model-with-colons>` value genuinely gets mis-split, and it bites in two places. A few things to flag before this can land.

## 1. The frontend fix is the one that actually matters — and it has an off-by-one

On `origin/master`, `_modelStateForSelect` (`static/ui.js:2672`) calls `_providerFromModelValue` **first**, before it ever looks at the option's `data-provider`:

```js
const explicitProvider=_providerFromModelValue(value);
if(explicitProvider) return {model:value,model_provider:explicitProvider};
```

and `_providerFromModelValue` (`static/ui.js:2655`) uses `lastIndexOf(':')`, so for `@custom:litellm-proxy:qwen3.6:27b-vision` it returns `custom:litellm-proxy:qwen3.6`. That corrupted provider wins even though the option carries the correct `data-provider`. Reordering to consult the option first is the right fix.

But your slug-list fallback has a bug. You compute `const inner=value.slice(1)` (strips the `@`), then in the custom fallback path do:

```js
// static/ui.js (this PR) — line 2668
return inner.slice(1,inner.lastIndexOf(':'));
```

`inner` is already `@`-stripped, so `slice(1, …)` drops the leading `c` of `custom`, yielding `ustom:litellm-proxy:qwen3.6`. The `1` is a leftover from the old code that sliced `value` (which still had the `@`). Your sibling non-custom branch already gets this right at line 2662 (`inner.slice(0,inner.indexOf(':'))`). Fix:

```js
return inner.slice(0,inner.lastIndexOf(':'));
```

This path only fires when `window._customProviderSlugs` is empty and there's no live `<select>`, but that's exactly the programmatic-caller case `_modelStateForSelect`'s fallback exists for.

## 2. The backend hunk targets code that no longer exists on master

Your merge-base is `d4e80b454` (#5819). Since then, master extracted the `@provider:model` parse out of `resolve_model_provider` into a shared helper, `_parse_provider_qualified_model_id` (`api/config.py:2449`), which `resolve_model_provider` now calls at `api/config.py:2711`. Your diff still edits the old inline block inside `resolve_model_provider`, so it will conflict on rebase, and — more importantly — the fix belongs in the helper so every caller (there are 2 references to it now) benefits, not just this one path.

Worth noting: master's single-peel already handles your headline example correctly. `@custom:litellm-proxy:qwen3.6:27b-vision` → `provider_hint` has 2 colons → one `rsplit` peel → `custom:litellm-proxy` + `qwen3.6:27b-vision`. Your loop only changes behavior for model names with **two or more** colons (e.g. `a:b:c`), where single-peel leaves `custom:litellm-proxy:a`. That's a legitimate improvement, but please retarget it into `_parse_provider_qualified_model_id` and add a 2+-colon case to the test so the intent is explicit.

## Suggested path forward

- Move the peel-until-known-slug loop into `_parse_provider_qualified_model_id` (`api/config.py:2449`), keep the `_custom_slug_rest_looks_like_host_port` and known-slug guards you already wrote.
- Fix the `slice(1,…)` → `slice(0,…)` off-by-one in the `ui.js` fallback.
- Rebase onto current master so the config.py hunk applies, and add a `qwen:7b:instruct`-style 2-colon model to `test_custom_provider_colon_model_split.py`.

The frontend reorder alone fixes the reported symptom, so if you'd rather split this into two PRs (frontend fix first, backend helper hardening second), that's fine by me. Happy to re-review once the base is current.

nesquena-hermes:
Thanks @tkuhlengel — your diagnosis here was exactly right: a custom-provider model whose *name* contains colons (e.g. `@custom:litellm-proxy:qwen3.6:27b-vision`) was mis-split, corrupting the provider id and breaking cron/settings provider resolution (#1948, #2047).

In the meantime, master refactored that parsing into a shared helper, `_parse_provider_qualified_model_id()` (`api/config.py`), and it now handles the exact cases this PR targets. Verified empirically against master:

```
@custom:litellm-proxy:qwen3.6:27b-vision  → ('qwen3.6:27b-vision', 'custom:litellm-proxy')  ✓
@custom:lmstudio:llama-3.1:8b             → ('llama-3.1:8b', 'custom:lmstudio')              ✓
@custom:litellm-proxy:qwen:free           → ('qwen:free', 'custom:litellm-proxy')            ✓
```

The helper peels the `custom:<slug>` segment back when the hint carries extra colons (and isn't a host:port slug), reassembling the colon-containing model name — the same outcome your backend change produced. So the backend half of this PR is now **superseded by master**, and I'm closing it on that basis rather than asking you to rebase onto a fix that's already landed.

Two things from this PR that are NOT covered and would be welcome as fresh, focused follow-ups if you're up for it:
1. **Frontend `data-provider` preference** — your `_modelStateForSelect()` reorder to prefer the option's `data-provider` attribute over string-splitting is independent of the backend parse and is a genuinely nice robustness improvement. A small standalone PR for just that would be easy to review + ship.
2. **Triple-colon edge** — a contrived `@custom:slug:model:v1:2024` (4 colons) still peels one segment too few in the shared helper. Your `_known_slugs` while-loop would handle it; if anyone actually hits that shape, folding a slug-aware peel into `_parse_provider_qualified_model_id` would close it for every caller at once.

Really appreciate the careful root-cause + the 7 regression tests — the analysis directly matched what master converged on. Closing as superseded; credit to you for the diagnosis. 🙏

PR #5892, opened 2026-07-10
fix(auth): use absolute path in login redirect to prevent redirect loop

Paslestrange:
## Bug

`check_auth()` in `api/auth.py` sent a **relative** Location header:

```python
handler.send_header('Location', 'login?next=' + _next)
```

Without a leading `/`, the browser resolves `login?next=...` against the current path. When a user visits `/session/<id>` while unauthenticated, the redirect chain is:

1. Browser requests `/session/<id>`
2. `check_auth` redirects to `Location: login?next=/session/<id>` (relative)
3. Browser resolves against `/session/<id>` → requests `/session/login?next=/session/<id>`
4. `/session/login` is not in `PUBLIC_PATHS` → `check_auth` fires again, now encoding the already-growing `next` parameter
5. Each iteration URL-encodes the previous value, producing exponential URL growth

This created an **infinite redirect loop** with exponentially growing URLs when accessing any `/session/` route while unauthenticated. The issue manifests especially when accessing the WebUI remotely (e.g. via ZeroTier/Tailscale) where users link directly to session URLs.

## Fix

Use a **parent-relative** `../login` path instead of a bare relative `login`:

```python
handler.send_header('Location', '../login?next=' + _next)
```

This resolves correctly across all current protected routes (`/session/<id>`, `/extensions/...`, etc.) and **preserves subpath-mount prefixes** (e.g. `/hermes/` behind a reverse proxy). Unlike an absolute `/login` path, `../login` does not escape the mount point.

Per RFC 3986, `../login` from `/session/foo` resolves to `/login`; from `/hermes/session/foo` it resolves to `/hermes/login`.

## Verification

Before fix:
```
$ curl -D - -o /dev/null http://127.0.0.1:8787/session/foo
HTTP/1.1 302 Found
Location: login?next=/session/foo    ← relative, resolves to /session/login
```

After fix:
```
$ curl -D - -o /dev/null http://127.0.0.1:8787/session/foo
HTTP/1.1 302 Found
Location: ../login?next=/session/foo  ← parent-relative, correctly resolves to /login
```

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -802,7 +802,7 @@ def check_auth(handler, parsed) -> bool:
         # safe='/' keeps path separators readable; everything else (including
         # `?`, `&`, `=`) gets percent-encoded.
         _next = _urlparse.quote(_path_with_query, safe='/')
-        handler.send_header('Location', 'login?next=' + _next)
+        handler.send_header('Location', '../login?next=' + _next)
         handler.send_header('Content-Length', '0')
         handler.end_headers()
     return False


nesquena-hermes (changes requested):
Thanks for catching a real bug here — the infinite-redirect loop when an unauthenticated client hits a non-public **client** route like `/session/<id>` is genuine: the old `Location: login?next=…` resolves relative to `/session/<id>` → `/session/login`, which is itself non-public, so `check_auth()` bounces it again forever. That part of the diagnosis is correct and worth fixing.

But the fix as written (unconditional `../login`) **regresses subpath-mount deployments** and **reddens two existing, PR-untouched tests**, so it can't ship as-is.

## 1. Breaks the `/hermes/`-style subpath mount (functional regression)

`../login` is resolved by the browser relative to the request path, and the server can't tell "mount prefix + shallow path" apart from "no mount + deep path" just by the path. Resolving the `Location` with `urljoin` (what the browser does):

| request path | old `login?next=` | new `../login?next=` |
|---|---|---|
| `/session/abc` (the loop you're fixing) | `/session/login` ❌ loop | `/login` ✅ |
| `/hermes/session/abc` (mounted) | `/hermes/session/login` ❌ | `/hermes/login` ✅ |
| `/hermes/foo` (mounted, shallow) | `/hermes/login` ✅ | `/login` ❌ **mount dropped** |
| `/hermes/` (mount root) | `/hermes/login` ✅ | `/login` ❌ **mount dropped** |

So the unconditional `../login` fixes the deep-path loop but **breaks the shallow-path case under a subpath mount** — a `/hermes/foo` request now redirects to `/login` (404 / wrong origin) instead of `/hermes/login`. That mount-relative behavior is deliberate; there's an existing block a few lines above (the `#5578` login-shaped guard, `api/auth.py:787-798`) that chooses between `../login` and bare `login` precisely to stay mount-safe.

## 2. Reddens two existing PR-untouched tests (master-green / head-red, proven)

The diff changes only `api/auth.py` (+1/-1) and updates no tests, but two existing tests pin the fall-through redirect to the relative form and now fail on your head:

- `tests/test_subpath_frontend_routes.py::test_server_auth_redirect_uses_relative_login_path_with_encoded_next`
- `tests/test_extension_hooks.py::test_extension_route_remains_behind_webui_auth`

Verified in a sandbox (`origin/master` vs your head `7d61adae`):
```
master → 2 passed
your head 7d61adae → 2 failed
  AssertionError: '../login?next=/extensions/app.js' == 'login?next=/extensions/app.js'
```
Both tests carry an explicit "so `/hermes/` subpath mount works" rationale, so they're guarding the exact behavior item 1 breaks — they're not stale.

## Fix-spec

Make the redirect **depth-conditional**, mirroring the existing `#5578` login-shaped block instead of hard-coding `../login` for every path. Concretely, in the non-login fall-through (`api/auth.py:799-807`):

- Only step up a level when the protected path actually has a sub-segment that would otherwise re-trigger `check_auth` (i.e. reuse the same `'/' in _path.lstrip('/')` heuristic the `#5578` block uses), and keep the bare relative `login?next=` for shallow/mount-root paths so `/hermes/foo` and `/hermes/` still land on `<mount>/login`.
- Add a regression test for the `/session/<id>` loop you're fixing (assert it resolves to a public login route and does **not** bounce again), and — since the correct fix keeps the shallow-path form relative — the two tests above should stay green without editing them. If you do intend to change their asserted shape, update them with the mount-safety justification rather than leaving them red.

Once the redirect is mount-safe (both the deep-path loop fixed **and** the `/hermes/foo` / `/hermes/` shallow cases still mount-relative) and the suite is green, happy to take another look.

nesquena-hermes:
Reading `api/auth.py:728-808` on `origin/master` alongside this PR's one-line diff, the shipped change is correct — and worth noting it's actually **`../login`**, not the `/login` the PR description quotes. That distinction matters, and the `../` form is the better of the two.

## Summary

The bug is real: `check_auth` emitted a relative `Location: login?next=...`. From `/session/<id>` the browser resolves that against `/session/`, producing `/session/login`, which is not in `PUBLIC_PATHS` (`auth.py:51-59`) so `check_auth` fires again and the `next=` blob re-encodes on every bounce until the URL explodes. This is the same failure class the #5578 guard above it (`auth.py:787-800`) already handles for the login route specifically.

## Code reference

Shipped hunk (`auth.py:805`):

```python
-        handler.send_header('Location', 'login?next=' + _next)
+        handler.send_header('Location', '../login?next=' + _next)
```

## Why `../login` is the right target (answers Greptile's subpath concern)

Greptile's 4/5 flag was that a root-absolute redirect can send users outside a path-prefixed mount. `../login` sidesteps that precisely because it's relative-with-a-climb rather than root-absolute:

- From `/session/<id>` it resolves to `/login` (the real public route in `PUBLIC_PATHS`).
- Under a subpath mount `<mount>/session/<id>` it resolves to `<mount>/login`, staying inside the mounted app — which a bare `/login` would *not* do.

The single-level `../` climb is safe because WebUI page routes bottom out at one segment of depth: `handle_get` only serves HTML for `/`, `/index.html`, and `/session/<id>` (`routes.py:11563`), with no deeper `/session/<id>/<sub>` page shapes. So there's no route where one `../` under-climbs. This also matches the existing #5578 guard right above, which already chose `../login` for the login-shaped case (`auth.py:793`) — the two redirect paths are now consistent.

## One nit

The PR body's before/after `curl` examples show `Location: /login?next=...` (root-absolute), but the committed diff produces `../login?next=...`. The committed form is the correct one; the description is just stale. Worth fixing the body so a future reviewer doesn't think the wrong variant shipped.

## Verification

A quick manual check covers it: with auth enabled, `curl -sD - -o /dev/null http://127.0.0.1:8787/session/foo` should now return `Location: ../login?next=/session/foo`, and following the chain should land on `/login` once with no growth in the `next=` parameter across repeated unauthenticated hits. The `changes-requested` label appears to predate this understanding of the `../` semantics; given Greptile's own concern is resolved by the climb, this looks re-reviewable.

PR #5894, opened 2026-07-10
feat(dictation): live VU meter with toggleable visual style

Kopamed:
## Summary

Adds a live mic-level indicator to the dictation flow so users can tell whether their input is being captured and at what gain, without waiting until transcription finishes to find out the gain was wrong.

## Why

On hardware with unpredictable built-in microphone gain (some thin-and-light laptops, common external USB mics), the user has no signal during dictation that the mic is picking them up. They discover the problem only when the resulting transcript is empty or wrong, after the recording has already ended. A real-time level indicator moves that feedback from "after the fact" to "while speaking," which is the only time the user can still adjust gain, position, or device.

## What changed

**`static/boot.js`** — VU meter module. Drives the indicator off the MediaRecorder capture stream via a Web Audio AnalyserNode, which is independent of the encoded blob and works for both `media-transcribe` and `media-raw` capture modes. Reads the user's preferred visual style from a Settings dropdown, with a URL hash override (`#vu=bars` / `#vu=fill`) for quick A/B without touching the panel. Cleans up the rAF loop and AudioContext on every recorder stop path.

Browser SpeechRecognition does not expose a `MediaStream`, so the meter is intentionally unavailable on that path and the existing `#micStatus` "Listening…" text stays in place. The code comment near `_vuStart` documents this scope.

**`static/index.html`** — VU element markup (a sibling of `#micStatus`, hidden by default), the new Settings dropdown, and a stop-square SVG sibling on `#btnMic` so the icon swaps honestly while recording.

**`static/style.css`** — quiet meter styling per the UI/UX guide (single accent, no glow, no shadow, no decorative motion). Mobile (`<=640px`) collapses the bars to 8 and keeps the fill at 80px. Suppresses the inactive visual mode (`data-style` selector) so a stray fill track doesn't bleed through next to the bars.

**`static/i18n.js`** — `settings_label_vu_style` and `settings_desc_vu_style` added to all 15 locale blocks. Two commits on this branch: the first skipped the `en` reference block, the second adds it. (`zh-Hant` was already correct in both commits.)

```js
// New module in static/boot.js — driven off the MediaRecorder stream
function _vuStart(stream){
  if(!stream) return;
  try{
    const Ctor = window.AudioContext || window.webkitAudioContext;
    if(!Ctor) return; // Browser without Web Audio — silently skip meter
    _vuAudioCtx = new Ctor();
    const src = _vuAudioCtx.createMediaStreamSource(stream);
    _vuAnalyser = _vuAudioCtx.createAnalyser();
    _vuAnalyser.fftSize = 256;
    _vuAnalyser.smoothingTimeConstant = 0.6;
    src.connect(_vuAnalyser);
    _vuBuildBars();
    _vuShow(_vuStylePref());
    _vuRafId = requestAnimationFrame(_vuTick);
  }catch(_err){
    _vuStop();
  }
}
```

## Verification

- `npm run lint:runtime` passes.
- `node --check` passes on `static/boot.js` and `static/i18n.js`.
- Manual: dictation start/stop, mode toggle from Settings, mode toggle via URL hash, persistence across page reloads, mobile narrow viewport, browser SpeechRecognition fallback path (no meter shown, "Listening" text remains).
- Mic icon swap observed on every capture backend that exercises `_setRecording(true)`.
- i18n: 15 locales × 2 keys = 30 entries, file parses cleanly. Locale-parity tests should pass on CI.

## Scope notes

- Visual style change takes effect on the next dictation start, not mid-recording. Mid-recording swap was deliberately avoided to prevent the indicator from jumping while the user is still speaking.
- The meter is a UI affordance only; it does not modify the captured audio or the recorded blob.

## Commit history

- `ab6b46a0` — original implementation. Skipped the `en` block when adding i18n keys, which broke locale parity.
- `9a67f892` — adds the keys to the `en` block (the parity reference). All 15 locales now match.


nesquena-hermes (changes requested):
Thanks for this — the VU meter with a toggleable fill/bars style is a nice touch, and the implementation itself (boot.js meter, the `#vu=` hash override, the style toggle) reads cleanly. One objective blocker before it can ship: **the new i18n keys break locale parity, so the locale-coverage CI tests fail.**

## The defect (locale parity — 8 tests red)

You added two keys, `settings_label_vu_style` and `settings_desc_vu_style`. This repo keeps every locale in `static/i18n.js` in exact key-parity with the `en` block, enforced by a test-per-locale. Two problems:

1. **The `en` block is missing both keys.** The first occurrence of `settings_label_vu_style` (around the `en` block) actually holds the **Italian** string (`'Stile indicatore VU dettatura'`) — it looks like the `en` slot got an Italian value pasted into it, so `en` has no correct English entry for either key. Since parity is keyed off `en`, this makes every other locale look like it has "extra" keys.
2. **The `zh-Hant` locale is missing both keys entirely.** The new keys land in 14 of the 15 locale blocks — `'zh-Hant'` (the quoted key, easy to miss) has neither.

Verified via the sandboxed suite (`origin/master` vs your head `ab6b46a0`):
```
master → 17 passed
your head ab6b46a0 → 8 failed:
    AssertionError: zh-Hant locale extra keys: ['settings_desc_vu_style', 'settings_label_vu_style']
  tests/test_czech_locale.py, test_japanese_locale.py, test_korean_locale.py,
  test_polish_locale.py, test_turkish_locale.py  (all: key-coverage mismatch vs en)
  tests/test_provider_quota_status.py::test_settings_label_and_description_i18n_keys_exist_for_all_locales
  tests/test_layout_helpers.py::test_raw_i18n_key_detected
```

## Fix-spec

1. In the **`en`** block of `static/i18n.js`, set the two keys to the real English strings (e.g. `settings_label_vu_style: 'Dictation VU meter style'` and a matching English `settings_desc_vu_style`). Right now the `en` slot carries Italian text.
2. Add both keys to the **`'zh-Hant'`** block (with a Traditional-Chinese translation, matching the other 14 locales).
3. Re-run the locale tests until green:
   ```
   pytest tests/ -k "locale or i18n or parity" -q
   ```
   All 15 locales must carry `settings_label_vu_style` + `settings_desc_vu_style` with no extras and no duplicates, and the `en` values must be English.

The feature code itself isn't in question here — this is purely the i18n key set. Once the `en`/`zh-Hant` entries are corrected and the locale suite is green, happy to take another look. (A visible dictation-UI change like this will also want a maintainer visual sign-off at merge time, but that's separate from this blocker.)

Kopamed:
Fixed in `9a67f892` — the `en` block now carries both keys with English strings.

Re your second point about `'zh-Hant'`: that locale already had both keys in the original commit (`ab6b46a0`), placed immediately after `settings_desc_raw_audio` — same logical position in every locale, but different line numbers per locale. Once `en` is corrected, parity should pass for all 15 locales including `'zh-Hant'`.

Verification: 15 locales × 2 keys = 30 entries, file parses cleanly. CI re-running on this commit will be the real test.

Kopamed:
## Status — ready for re-review

The reviewer's blocker (`en` block missing the i18n keys, breaking locale parity across 8 tests) is fixed in `9a67f892`. Both items addressed:

**(1) `en` block now has both keys with real English strings.**
```
settings_label_vu_style: 'Dictation VU meter style',
settings_desc_vu_style: 'Visual style for the live mic level indicator that appears while dictating. URL hash #vu=fill or #vu=bars overrides this for the current session.',
```

**(2) `'zh-Hant'` was already correct** — the reviewer's diagnosis on that point was a misread of the diff. The keys land in the same logical position (after `settings_desc_raw_audio`) in every locale, but the line numbers differ because the surrounding locale bodies have different lengths. The test flagged `'zh-Hant'` as having "extra keys" because the `en`-block parity check was failing; once `en` is fixed, parity passes for all 15 locales.

## Verification

- `npm run lint:runtime` clean.
- `node --check` clean on all modified files.
- i18n: 15 locales × 2 keys = 30 entries, file parses cleanly.
- Manual: bars mode, fill mode, dropdown toggle, URL-hash override, mobile narrow viewport, browser-SR fallback path, persistence across reloads.
- CI: 18/18 jobs green on `9a67f892` (browser-smoke, lint, test matrix 3.11/3.12/3.13 × 5 shards, Greptile).
- Greptile: 5/5 confidence, safe to merge.

## Commit history

- `ab6b46a0` — original implementation (skipped `en` block, broke locale parity).
- `9a67f892` — adds the keys to `en` block. All 15 locales now match.

## What needs maintainer action

A re-review against `9a67f892`. The previous `CHANGES_REQUESTED` was against `ab6b46a0`; the locale-parity blocker is resolved in the latest commit. Greptile and the test matrix both pass clean.

Kopamed:
@nesquena-hermes, the requested fixes are now on `4a388de8`, rebased onto current `master` (`d65c0baf`). The VU meter keeps the Listening live region accessible, tears down Web Audio on every exit, and now has executable render/threshold/lifecycle coverage; its settings copy follows the shared English-fallback contract without weakening locale coverage for unregistered keys. All 18 checks pass on this SHA, plus 69 locale/contract, 114 dictation/voice, and 41 provider-quota tests passed locally. Please re-review the current head when available.

nesquena-hermes:
## Gate certification — RED on exact head `4a388de81949943e16b700ffdc7818936cf1179e`

The prior locale/i18n blocker from `ab6b46a0` **is fixed** on this head. Exact-head technical, suite, advisor, and real-browser re-gating found one remaining ownership race in the new VU lifecycle.

### Blocking reproduction: stale recorder A tears down active recorder B

`static/boot.js`'s `recorder.onerror` / `recorder.onstop` callbacks compute `isCurrentCapture`, but `_vuStop()` and the `onstop` blob-processing path are not owned by that check.

I reproduced this against the integrated production code in Chromium:

1. MediaRecorder A starts; its VU graph is visible.
2. A emits `error`; A is correctly torn down and the app becomes idle.
3. The user immediately starts MediaRecorder B; B is recording and its VU is visible.
4. A emits its spec-following late `stop` event.
5. A's stale `onstop` closes B's AudioContext and hides B's meter while `window._micActive === true` and B remains `state === "recording"`.

Probe result:

```text
before A late-stop: B active=true, meter=flex, B context closed=false
 after A late-stop: B active=true, meter=none, B context closed=true, B state=recording
reproduced=true
```

A nonempty A blob also continues through A's stale `onstop` into raw-audio send/transcription after B has begun. This is broader than a cosmetic meter blip: the old capture can process against the new capture's global composer/pending-send state.

### Required fix

Make recorder callbacks capture-owned, not just parts of their UI cleanup:

1. In `recorder.onerror`, perform global recorder/UI/pending-send/VU changes only when `isCurrentCapture`; always stop only that callback's `captureStream` tracks.
2. In `recorder.onstop`, after recorder-reference cleanup, stop that callback's captured tracks and **return immediately when `!isCurrentCapture`**. A stale callback must not snapshot the current `_prefix`, call `_setRecording(false)`, call `_vuStop()`, process/send/transcribe its blob, clear current pending-send state, or apply deferred settings against capture B.
3. Add a production-code regression for: A start → A error → B start → A late data/stop. Assert B stays active, B's VU graph remains open/visible, B's pending state is unchanged, and A's stale blob is not sent/transcribed.

### Gate evidence

- Exact contributor head: `4a388de81949943e16b700ffdc7818936cf1179e`
- Frozen master: `ecc47782e3e72a5bf3e848a8f09a2a5a838c71b1`
- Clean composition commit: `8f5bf0371695e8b8afe6145e58a018b8bdcc89a8`
- Threat scan: CLEAN
- Hosted checks: 18 completed, zero failures
- Focused VU/i18n/a11y slice: **101 passed**
- Ruff forward gate, ESLint runtime, current scope-undefined gate, browser smoke: CLEAN
- Full suite: **13,697 passed**, with the same 16 failures + 2 errors as frozen master; no candidate-only suite failure
- Live desktop/mobile proof: animated 16-bar and fill modes, stop-square state, sr-only status, responsive geometry, and Settings placement all rendered cleanly
- Codex: SAFE on the normal lifecycle (did not exercise this interleaving)
- Opus: APPROVE, but independently code-traced the same stale callback race
- Fable: SHIP-WITH-UX-FIXES (plain-language settings copy recommended); visuals are not the RED reason

Direct production-path reproduction overrides the normal-lifecycle advisor approvals. This exact head remains **RED**. A changed head with the ownership guard and regression is eligible for full re-gating.

PR #1748, opened 2026-05-06
feat(theme): expose active `--bg` via `<meta name="theme-color">` for native chrome bridges

nesquena-hermes:
# feat(theme): expose active `--bg` via `<meta name="theme-color">` for native chrome bridges

## Why

Native WKWebView wrappers (the Mac Swift app at `hermes-webui/hermes-swift-mac`, and any other future wrapper) need to keep their AppKit chrome — tab bar, title bar, traffic-light area, status bars — in sync with whatever theme the user has active inside the WebUI.

Today the Swift app does this by pixel-sampling the page via `document.elementsFromPoint(x, y)` at three fixed viewport coordinates and reading `getComputedStyle().backgroundColor` of the topmost opaque element. The approach is fragile in two real ways:

1. **Sample-point overlay collisions.** Any opaque modal / lightbox / settings panel / file-tree overlay / image preview that covers any of the three sample points and stays visible past the 2.5-second stability gate makes the bridge fire the *overlay's* colour. With multi-window / native tabs that one tab's overlay-derived colour gets propagated to every other tab.
2. **Pixel sampling is also IPC-heavy.** `setInterval(report, 2000)` fires on every WKWebView, every two seconds, even when nothing has changed.

The user-visible symptom is the bug filed at [hermes-webui/hermes-swift-mac#70](https://github.com/hermes-webui/hermes-swift-mac/issues/70) — the chrome flashes white for an unpredictable period during multi-tab sessions, persists after closing the offending tab, and reaches accessibility / photosensitivity territory.

The right architectural fix is to have the WebUI surface its current theme background as the single source of truth, via the standard `<meta name="theme-color">` mechanism that mobile Safari, iOS PWAs, and WebKit-based wrappers already understand. The Mac Swift app then reads that one value directly — no overlay sampling, no IPC cost, no false positives.

This PR is the WebUI half of that change. The Mac side will be a separate PR on the swift-mac repo.

### `static/index.html` (5 lines added)

Three meta tags + an inline pre-paint seeder, placed right after the existing theme bootstrap script:

```html
<!-- theme-color: surfaces the active theme's background to native chrome ... -->
<meta name="theme-color" content="#FEFCF7" media="(prefers-color-scheme: light)">
<meta name="theme-color" content="#0D0D1A" media="(prefers-color-scheme: dark)">
<meta name="theme-color" id="hermes-theme-color" content="#0D0D1A">
[...]

PR #5900, opened 2026-07-10
feat(sidebar): right-click 'Mark as read' context menu action

Kopamed:
Add a right-click context menu to sidebar chat items with a "Mark as read" option.

Currently the only way to clear a notification badge is to open the chat. This adds a context menu that clears it without navigating away.

- `static/sessions.js` — context menu action + handler (15 lines)
- `static/i18n.js` — locale string (2 lines)
- `tests/test_1466_sidebar_cancel_clarify.py` — update for the new menu option

### DIFF
--- a/static/i18n.js
+++ b/static/i18n.js
@@ -513,6 +513,8 @@ const LOCALES = {
     path_copy_failed: 'Failed to copy path: ',
     session_rename: 'Rename conversation',
     session_rename_desc: 'Edit the title of this conversation',
+    session_mark_read: 'Mark as read',
+    session_mark_read_desc: 'Clear the unread notification for this conversation',
     session_title_regenerate: 'Regenerate title',
     session_title_regenerate_desc: 'Ask Hermes to generate a fresh title from the conversation',
     session_title_regenerating: 'Regenerating title…',
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -4662,6 +4662,31 @@ async function _archiveSession(session, archived=true, beforeListRender=null){
   }catch(err){if(renderHold) await renderHold.catch(()=>{});_pendingSessionReflowPositions=null;showToast(t('session_archive_failed')+err.message);return false;}
 }
 
+// Clear the unread badge for a session without opening it (#1748), used by the
+// sidebar "Mark as read" action.
+//
+// The viewed watermark must be the AUTHORITATIVE completed count, not just the
+// cached sidebar row's message_count. A background completion records its count
+// in the explicit completion-unread marker, and _hasUnreadForSession
+// short-circuits to "unread" on that marker regardless of the cached row. That
+// marker count can exceed the sidebar cache's message_count when the list has
+// not re-fetched the higher count yet. Persisting only the stale cached count
+// clears the badge now but lets the next /api/sessions refresh — which returns
+// the authoritative higher count — re-flag the session as unread, so the badge
+// disappears and reappears (#5900). Resolve one authoritative value here so the
+// code that decided the session was unread and the code that clears it agree.
+function _markSessionRead(session){
+  if(!session || !session.session_id) return;
+  const sid = session.session_id;
+  let viewedCount = Number(session.message_count || 0);
+  if(!Number.isFinite(viewedCount)) viewedCount = 0;
+  const marker = _getSessionCompletionUnread()[sid];
+  const markerCount = marker ? Number(marker.message_count) : NaN;
+  if(Number.isFinite(markerCount)) viewedCount = Math.max(viewedCount, markerCount);
+  _setSessionViewedCount(sid, viewedCount);
+  if(typeof renderSessionListFromCache === 'function') renderSessionListFromCache();
+}
+
 function _openSessionActionMenu(session, anchorEl){
   const isReadOnly = _isReadOnlySession(session);
   if(_sessionActionMenu && _sessionActionSessionId===session.session_id && _sessionActionAnchor===anchorEl){
@@ -4720,8 +4745,7 @@ function _openSessionActionMenu(session, anchorEl){
       ICONS.check,
       ()=>{
         closeSessionActionMenu();
-        _setSessionViewedCount(session.session_id, Number(session.message_count || 0));
-        renderSessionListFromCache();
+        _markSessionRead(session);
       }
     ));
   }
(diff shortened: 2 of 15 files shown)


nesquena-hermes (changes requested):
## Changes requested — i18n locale-parity failure (feature code is sound)

Thanks for this — the feature itself is wired correctly (`_openSessionActionMenu` reuses `_hasUnreadForSession` / `_setSessionViewedCount` / `renderSessionListFromCache`, all existing, and the `check` icon + `#1748` gating look right). The blocker is entirely in `static/i18n.js`: the new keys break locale parity, which reddens a batch of **existing, untouched** locale tests.

### Reproduced (sandboxed, no network)

- `pytest tests/ -k "locale or i18n or parity"` on this PR head → **13 failed / 348 passed** (12 real + 1 unrelated playwright-not-installed artifact).
- The same tests on `origin/master` → **61 passed / 0 failed**.

So this is master-green / head-red: the failures are introduced by this PR, not pre-existing.

### Root cause

`session_mark_read` / `session_mark_read_desc` were added to only **5 of the 15 locale blocks**, and the 4 non-`en` additions are malformed:

1. **10 locales are missing the keys entirely** (`ja`, `es`, `de`, `pt`, `fr`, `cs`, `tr`, `pl`, `vi`, and `it`). Every locale block must carry every key in `en` or the coverage tests fail (`test_<locale>_locale_covers_english_keys` / `..._matches_english_key_coverage`).

2. **The 5 that got a translation are jammed onto an existing key-line** instead of being their own line, e.g.:
   ```js
   // i18n.js:5177  (this is inside the 'zh-Hant' block)
   mcp_delete_confirm_title: '刪除 MCP 伺服器',  session_mark_read: 'Отметить как прочитанное',
     session_mark_read_desc: 'Отметить чат как прочитанный без открытия',
   ```
   The `_desc` continuation is indented 2 spaces instead of 4, which trips `test_korean_locale_keys_use_standard_indentation`.

3. **At least one translation landed in the wrong block.** The line above is inside the Traditional-Chinese (`'zh-Hant'`) block but the text is **Russian** — so `zh-Hant` still has no `session_mark_read` (`test_zh_hant_locale...`: `Missing zh-Hant translation: session_mark_read`) while the Russian block never got it either.

### Fix-spec

- Add `session_mark_read` and `session_mark_read_desc` to **all 15 locale blocks** in `static/i18n.js` (`en`, `it`, `ja`, `ru`, `es`, `de`, `zh`, `zh-Hant`, `pt`, `ko`, `fr`, `cs`, `tr`, `pl`, `vi`).
- Give each its **own line** at the block's standard **4-space** indentation, with the correct-language translation for that block (the Russian string belongs in `ru`, the Traditional-Chinese string in `zh-Hant`, etc.). Remove the current jammed-onto-existing-line insertions.
- Re-run `pytest tests/ -k "locale or i18n or parity"` until it's green (ignore `test_layout_helpers::test_raw_i18n_key_detected` — that one fails only because Playwright/Chromium isn't installed locally; it also fails on master and is unrelated).

The feature code needs no changes. Once the locale keys are complete and correctly placed, this should sail through.

Kopamed:
@nesquena-hermes, the requested fixes are now on `600b0c17`, rebased onto current `master` (`d65c0baf`). The i18n copy now follows the shared English-fallback contract, and mark-as-read persists the authoritative completion count so the unread badge cannot return after refresh; the regression reproduces the old failure and passes with the fix. All 18 checks pass on this SHA, plus 386 targeted sidebar/unread/i18n tests passed locally. Please re-review the current head when available.

nesquena-hermes:
## Gate certification — RED on exact head `600b0c172bc3f9f789624840f61a85222eab5197`

The prior stale completion-marker blocker **is fixed** on this head, but a related real-menu cache-race still makes “Mark as read” fail. The raw head is also conflicting with current master.

### Blocking reproduction

The new menu handler captures the session object that existed when the menu opened. `_markSessionRead(session)` resolves only:

```text
max(captured session.message_count, completion marker.message_count)
```

It does not re-resolve the current cached session at click time.

I reproduced this through the real rendered context menu on the integrated app:

1. menu opens from a cached row at message count **5**;
2. its completion marker carries **8**;
3. while the menu stays open, the authoritative session-list cache replaces the row with count **10**;
4. click the real **Mark as read** menu item;
5. the action persists viewed count **8**, clears the marker, repaints, and the current count-10 row remains unread.

Observed result:

```json
{
  "before": {"capturedCount": 5, "currentCount": 10, "marker": 8, "viewed": 5},
  "after": {"viewed": 8, "marker": null, "currentUnread": true, "rowUnread": true},
  "reproduced": true
}
```

That directly violates the action’s promise. Direct browser behavior overrides the otherwise-green tests/advisors.

### Precise fix spec

1. Rebase onto current master. The production hunks merge cleanly; ten locale tests conflict because master added another English-fallback-owned key. Preserve `workspace_artifact_source_session` and add `session_mark_read` / `session_mark_read_desc` in the unified `ENGLISH_FALLBACK_OWNED_KEYS` registry.
2. In `_markSessionRead`, re-resolve the current session snapshot by `sid` at action time (for example via the existing `_sessionSnapshotById(sid)`) and persist the maximum finite value across:
   - the captured menu session count,
   - the current snapshot/cache count,
   - the completion-marker count.
3. Add a real action regression for the 5/8/10 menu-open interleaving. After click, viewed count must be 10, marker absent, menu dismissed, and both `_hasUnreadForSession(current)` and row `.unread` false.
4. Re-run the exact technical/visual gate. Do not solve this with DOM badge removal; the persisted watermark is the authority.

### Other evidence

- **Threat/hosted:** exact-head threat verdict CLEAN; 18 hosted check-runs completed with zero failures.
- **Conflict-resolved focused gate:** 74 affected mark-as-read/menu/locale tests passed; scope-undefined, ESLint runtime, and Node syntax gates passed.
- **Full suite:** 13,694 passed, 83 skipped, 1 xfailed, 2 xpassed, 8 failed + 2 errors; failure IDs exactly match the stable frozen-master baseline, with zero candidate-only suite failures.
- **Normal browser path:** right-click and mobile three-dot menus render the conditional action cleanly. In the non-racing 5/8/8 case, the real click correctly persists 8, clears the marker/menu/badge, and stays read after refresh.
- **Fable correctness:** APPROVE; **Fable UX:** SHIP-WITH-UX-FIXES. Codex identified the cache-drift failure, which I independently reproduced in the production path.

### Separate non-certificate product dependencies

The technical RED is independent of these owner calls:

- whether read-only/cron rows must also expose the action (today some unread badges still require opening the conversation),
- whether one-way “Mark as read” ships without “Mark as unread,”
- whether per-browser localStorage read state is acceptable versus cross-device state,
- menu-slot approval and native translations beyond the sanctioned English-fallback mechanism.

This certifies the exact contributor head only. No contributor branch was modified, and nothing was merged, tagged, deployed, or closed.

PR #5907, opened 2026-07-11
fix: prefer option data-provider over string-splitting in _modelStateForSelect

tkuhlengel:
When a custom provider (e.g. `custom:litellm-proxy`) serves a model whose name contains colons (e.g. `qwen3.6:27b-vision`), the `@provider:model` dropdown value becomes `@custom:litellm-proxy:qwen3.6:27b-vision`. The previous `_modelStateForSelect()` called `_providerFromModelValue()` first, which used `lastIndexOf(:)` and corrupted the provider to `"custom:litellm-proxy:qwen3.6"`.

This fix reorders `_modelStateForSelect()` to prefer the option's `data-provider` attribute (set by the backend during dropdown population) over string-splitting. The `data-provider` attribute carries the exact provider ID, avoiding colon-parsing entirely.

Also:
- Populate `window._customProviderSlugs` from `/api/models` for fallback
- Improve `_providerFromModelValue()` to match known custom slugs first
- Fix off-by-one bug in custom fallback (`slice(1,...)` → `slice(0,...)`)

This is a frontend-only fix; the backend parsing was already hardened in master via `_parse_provider_qualified_model_id()`.

**Closes:** #1948, #2047  
**Fixes:** cron job and settings provider corruption

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -2649,7 +2649,11 @@ function _getOptionProviderId(opt){
     return group.dataset.provider;
   }
   const value=String(opt.value||'');
-  if(value.startsWith('@') && value.includes(':')) return value.slice(1,value.lastIndexOf(':'));
+  if(value.startsWith('@') && value.includes(':')){
+    // Use the same heuristics as _providerFromModelValue to handle
+    // custom providers with host:port slugs and models containing colons.
+    return _providerFromModelValue(value);
+  }
   return '';
 }
 function _providerFromModelValue(modelId){
@@ -2691,10 +2695,24 @@ function _providerFromModelValue(modelId){
     // No more colons — malformed, return what we have
     return inner;
   }
-  // Check if this looks like host:port (contains exactly one colon in the slug)
+  // Check if this looks like host:port (contains a dot in the slug segment)
   const potentialSlug=afterCustom.slice(0,firstColonPos);
   if(potentialSlug.includes('.')){
-    // Looks like an IP or hostname with port — don't peel further
+    // Looks like an IP or hostname — the next segment may be a port number.
+    // Consume it so the provider ID retains the full host:port slug
+    // (e.g. custom:proxy.internal:8443 → provider, not custom:proxy.internal).
+    const afterSlug=afterCustom.slice(firstColonPos+1);
+    const secondColonPos=afterSlug.indexOf(':');
+    if(secondColonPos===-1){
+      // No model after the host segment — return what we have
+      return 'custom:'+potentialSlug;
+    }
+    const portSegment=afterSlug.slice(0,secondColonPos);
+    if(/^\d+$/.test(portSegment)){
+      // Numeric port — include it in the provider slug
+      return 'custom:'+potentialSlug+':'+portSegment;
+    }
+    // Not a numeric port — the host is the slug, rest is the model
     return 'custom:'+potentialSlug;
   }
   // Simple slug (e.g. 'litellm-proxy') — provider is 'custom:<slug>'
--- a/static/ui.js
+++ b/static/ui.js
@@ -2654,8 +2654,33 @@ function _getOptionProviderId(opt){
 }
 function _providerFromModelValue(modelId){
   const value=String(modelId||'').trim();
-  if(value.startsWith('@')&&value.includes(':')) return value.slice(1,value.lastIndexOf(':'));
-  return '';
+  if(!value.startsWith('@')||!value.includes(':')) return '';
+  // Strip the leading @, then try to find the provider boundary.
+  // For non-custom providers (e.g. @openrouter:model), split on the first colon.
+  // For custom providers (custom:slug), the slug may contain colons (e.g.
+  // custom:litellm-proxy or custom:10.8.71.41:8080), so we can't use
+  // lastIndexOf(':') — that would eat part of the model name when the model
+  // itself contains colons (e.g. qwen3.6:27b-vision).
+  // Strategy: try known custom provider slugs from the global model catalog
+  // first (window._customProviderSlugs), then fall back to lastIndexOf.
+  const inner=value.slice(1);
+  // Check if we have known custom provider slugs available
+  if(typeof window!=='undefined'&&Array.isArray(window._customProviderSlugs)){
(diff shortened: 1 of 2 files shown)


nesquena-hermes (changes requested):
## 🔬 Review: changes requested — good fix, one edge remains in the fallback

Thanks @tkuhlengel — preferring the option's `data-provider` over `lastIndexOf(':')` string-splitting is the right call, and a gate confirmed the common cases all resolve correctly now: the colon-in-model-name case (`custom:litellm-proxy` + `qwen3.6:27b-vision`), plain models, `@openrouter:vendor/model`, built-in providers, and `custom:lab` + colon-model all parse cleanly.

**One remaining edge (the string-split FALLBACK path):** when the `data-provider` attribute is absent/empty (older option markup, or a code path that builds the value without the attribute), `_modelStateForSelect()` still falls back to the `lastIndexOf(':')` split — which corrupts a custom provider whose **base_url has a port** AND whose **model has colons**. Gate probe:

- `@custom:proxy.internal:8443:Qwen3-235B` **with** `data-provider` metadata → `custom:proxy.internal:8443` ✅ (your fix)
- `@custom:proxy.internal:8443:Qwen3-235B` **without** the attribute (fallback) → `custom:proxy.internal` ❌ (still wrong)

The fix makes the common case correct, but the fallback keeps the same class of corruption for the port-in-base-url + colon-in-model combination.

**Suggested fix:** in the fallback split, don't use `lastIndexOf(':')` — resolve the provider by matching against the known/configured provider slugs (longest-prefix match on the `@<provider>:` head), or greedily consume the registered `custom:<name>` (including a `:port` segment) so only the true model tail is split off. That closes the fallback to match the data-provider path.

Also: please add a regression test covering both the data-provider path AND the fallback path for a custom provider with a port-bearing base_url and a colon-in-name model — that pins the behavior so it can't regress.

Once the fallback is hardened + a test lands, I'll re-gate. Nice catch on the original bug.

tkuhlengel:
Updated to fix the fallback heuristics. Replaced `lastIndexOf(:)` with smarter parsing:

- For simple slugs (e.g. `litellm-proxy`), provider is `custom:<slug>` and model is everything after the first colon
- For host:port slugs (e.g. `10.8.71.41:8080`), detect by checking for a dot and preserve the full host:port

This addresses the Greptile concern about the fallback path still corrupting custom provider IDs in boot/restore paths before the slug cache is ready.

tkuhlengel:
Fixed the slug matching order. Now sorts slugs by length (longest first) before matching, which prevents a shorter slug (e.g. `custom:litellm`) from matching before a longer one (e.g. `custom:litellm-proxy`).

This addresses the concern about sending restored/saved selections to the wrong custom backend.

@greptile-apps retrigger

tkuhlengel:
@greptile-apps retrigger

Updated with:
- Fallback path now preserves numeric ports in host:port custom provider slugs
- _getOptionProviderId delegates to _providerFromModelValue for consistent parsing
- Added regression test (test_5907_custom_provider_port_fallback.py) covering both data-provider and fallback paths

tkuhlengel:
@nesquena-hermes Greptile approved (5/5). All concerns addressed:

- Fallback path preserves numeric ports in host:port custom provider slugs
- _getOptionProviderId delegates to _providerFromModelValue for consistent parsing  
- Added regression test covering both data-provider and fallback paths

Ready for maintainer review.

nesquena-hermes (changes requested):
## 🔬 Gate review (Codex) — SHIP ONLY WITH FIXES (2 fallback/server paths still corrupt colon-models)

Thanks @tkuhlengel — the primary fix is right: with authoritative option `data-provider` metadata, the reported `@custom:litellm-proxy:qwen3.6:27b-vision` case resolves correctly, and ordinary `@openai:model`, `@openrouter:vendor/model:free`, bare listed + bare unlisted models all keep their existing behavior (verified in focused Node probes). But the gate found the colon-model corruption still happens on **two paths this PR didn't cover**:

1. **[SILENT — client fallback] `static/ui.js:2698`** — when an option lacks provider metadata (before the slug catalog loads, or a hand-typed custom model), the code falls back to the first-segment/dot heuristic, which turns `custom:localhost:11434` into `custom:localhost` (drops `:11434`). Fix: replace the fallback heuristic with the right-to-left parse used by `api.config._parse_provider_qualified_model_id`, preserving `localhost:<port>` slugs and colon-bearing model names. Add no-catalog tests with + without a colon-bearing model.

2. **[SILENT — server re-corrupts] `api/routes.py:6236`** — even after the browser sends the correct value, chat start re-splits it: `_resolve_compatible_session_model_state('@custom:litellm-proxy:qwen3.6:27b-vision', 'custom:litellm-proxy', explicit_model_pick=True)` returns provider `custom:litellm-proxy:qwen3.6`, and the response overwrites `S.session.model_provider` at `static/messages.js:1845` → reasoning lookup then strips the model to `27b-vision` and misses the custom-provider capability config. Fix: reuse `api.config._parse_provider_qualified_model_id` server-side here, or honor the exact authoritative `requested_provider` prefix before splitting.

The `data-provider` approach is the right primary fix — it just needs to extend to the client fallback AND the server-side resolver so a colon-bearing custom model survives end-to-end (dropdown → chat start → persisted session → reasoning lookup). Once both use the hardened right-to-left parser, re-gate.

nesquena-hermes:
📎 Scope update: the **server-side** half of my earlier review (finding #2 — `_resolve_compatible_session_model_state` re-splitting a colon-bearing custom-provider model) has since been fixed independently by #5979, shipped in **exp-v0.52.53** (it now honors an exact `requested_provider` prefix before splitting). So that half is handled on master.

Your remaining work is just **finding #1** — the client-side fallback `_providerFromModelValue()` (`static/ui.js:2657`) still uses `lastIndexOf(':')`/first-segment splitting, which drops `:port` from `localhost:11434` and mis-splits colon-bearing model names when option `data-provider` metadata is absent (pre-catalog / hand-typed). Rebasing onto current master (with #5979) + switching that fallback to the right-to-left parse (mirroring `api.config._parse_provider_qualified_model_id`) should get this green. Thanks!

PR #5909, opened 2026-07-11
fix(api/config): surface + route plugin-only providers in static catalog (salvage of #5461)

nesquena-hermes:
## Summary

Salvage-rebuild of #5461 (thanks @alexfoxtm 🙌). The catalog-surfacing half of the
original PR was verified correct by a prior maintainer gate; this rebuilds it
fresh on current `master` and adds the **one missing piece** that makes surfaced
plugin-only models actually *routable*, not just selectable.

Plugin-only model providers (e.g. `9router` and other model plugins whose slugs
are not in the static `_PROVIDER_MODELS` / `_PROVIDER_DISPLAY` tables) were
silently dropped from the model picker on the cold-cache path, and — even when
surfaced — a selected plugin model such as `gh/claude-sonnet-4.5` would inherit
the **default** provider instead of routing to the plugin.

## What this does

**Catalog surfacing (ported fresh from #5461, `api/config.py`):**
- `_static_models_catalog_without_live_probes()` now enumerates
  `_plugin_model_provider_profiles().keys()` into `detected_providers`, gated by
  `_provider_has_key()` and wrapped in `try/except` → `debug`, so an
  installed-and-keyed plugin enters the static catalog even without a
  `providers.<slug>` block.
- Plugin-only groups stay visible through the build loop and the post-loop
  filter even with zero models yet.
- `raw_models` is seeded from the plugin's `ProviderProfile.fallback_models`
  when otherwise empty, so the cold path surfaces a curated, network-free subset
  (the live rebuild's full `/v1/models` fetch supersedes it on the next call).

**Routing fix (new — the missing half):**
- `model_with_provider_context()` now emits an explicit `@<plugin>:<model>` hint
  when `_is_plugin_model_provider(provider)` is true, placed **before** the
  existing `"/" in model` bare-passthrough. This mirrors how an explicit
  configured provider is handled and keeps a surfaced plugin-only model routing
  to its plugin instead of the default provider.

## Tests

- Ported `TestPluginFallbackModelsInStaticCatalog` from #5461 (catalog surfacing
  + fallback_models).
- Added `TestPluginModelProviderRouting::test_plugin_only_model_routes_to_plugin_provider`
  proving `model_provider=<plugin>, model='gh/...'` → `@<plugin>:gh/...`.
  Verified this test **fails without** the routing branch (returns the bare
  `gh/...`) and passes with it.

```
tests/test_plugin_model_providers.py .........  [100%]
9 passed
```

## Attribution

Salvaged from #5461 by @alexfoxtm — the catalog-surfacing half was correct; this
PR rebuilds it fresh on `master` and adds the missing routing + a routing
regression test. Refs #5461.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -35,6 +35,7 @@ import api.paths as _paths
 from api.plugin_providers import (
     effective_provider_display_name as _effective_provider_display_name,
     is_plugin_model_provider as _is_plugin_model_provider,
+    plugin_model_provider_profiles as _plugin_model_provider_profiles,
 )
 
 HOME = _paths.HOME
@@ -2978,6 +2979,17 @@ def model_with_provider_context(model_id: str, model_provider: str | None = None
     if isinstance(providers_cfg, dict) and provider in providers_cfg:
         return f"@{provider}:{model}"
 
+    # Plugin-only model providers (e.g. 9router, and other model plugins whose
+    # slugs are not in the static provider tables) route through the plugin,
+    # not the default provider. A surfaced plugin-only model such as
+    # 'gh/claude-sonnet-4.5' carries a '/' in its ID, so without this branch it
+    # would fall through to the bare-passthrough below and inherit the default
+    # provider. Emit the explicit '@plugin:model' hint so the model stays
+    # routable to the plugin that surfaced it. This mirrors the explicit
+    # configured-provider handling above.
+    if _is_plugin_model_provider(provider):
+        return f"@{provider}:{model}"
+
     # For non-OpenRouter slash IDs without an explicit configured provider,
     # keep the ID intact so existing custom/proxy base_url routing and
     # portal-provider handling remain in charge.
@@ -4725,6 +4737,23 @@ def _static_models_catalog_without_live_probes() -> dict:
             if canonical and _provider_has_key(canonical):
                 detected_providers.add(canonical)
 
+        # Plugin-only providers (e.g. 9router) are not in the static
+        # _PROVIDER_MODELS / _PROVIDER_DISPLAY tables and are detected above
+        # only when the user puts them in `providers.<slug>`.  Plugins ship
+        # with their own env-var wiring, so an installed-and-keyed plugin
+        # provider should also enter the static catalog even without a
+        # `providers:` block — otherwise the picker silently drops the
+        # group when the live-rebuild cache is cold.
+        try:
+            for _plugin_pid in list(_plugin_model_provider_profiles().keys()):
+                if not _plugin_pid or not _provider_has_key(_plugin_pid):
+                    continue
+                _canonical = _canonicalise_provider_id(_plugin_pid) or _plugin_pid
+                if _canonical:
+                    detected_providers.add(_canonical)
+        except Exception:
+            logger.debug("Plugin provider detection failed in static catalog", exc_info=True)
+
         fallback_cfg = cfg.get("fallback_providers", []) if isinstance(cfg, dict) else []
         if isinstance(fallback_cfg, list):
             for entry in fallback_cfg:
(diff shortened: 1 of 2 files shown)


nesquena (approved):

Independent review of the salvage of #5461 (@alexfoxtm): surface **plugin-only** model providers (e.g. `9router`) in the static/cold-cache catalog **and** keep their selected models routing to the plugin. Plugin providers aren't in the static `_PROVIDER_MODELS`/`_PROVIDER_DISPLAY` tables, so on the cold-cache path they emitted empty groups the picker filtered out (looked uninstalled), and a selected `gh/claude-sonnet-4.5` inherited the *default* provider instead of the plugin.

### What this ships

`api/config.py` (catalog-surfacing in `_static_models_catalog_without_live_probes` + routing in `model_with_provider_context`), `tests/test_plugin_model_providers.py` (+ routing test). Agent-authored (`nesquena-hermes`). MERGEABLE. The catalog half was maintainer-gated in #5461; the routing half is the new "missing piece."

### Routing fix — the precedence is correct and can't hijack standard providers

`model_with_provider_context` gains an `@<plugin>:<model>` branch, placed **after** the explicit `providers.<slug>` configured-provider check and **before** the bare-`/`-passthrough. The load-bearing question is whether this steals standard slash-models (openrouter/custom/nous). It can't — I verified `is_plugin_model_provider`'s scoping (`api/plugin_providers.py`): it returns False for anything in `_webui_static_provider_ids()` and True only for a slug in `plugin_model_provider_profiles()`. Harnessed: `openrouter`/`custom`/`nous`/`openai`/`anthropic` → all `False`, so their slash-models still take the intact bare-passthrough (openrouter `anthropic/…` stays bare, custom `vendor/model` stays bare). The `@plugin:model` form uses the same `@provider:model` convention `resolve_model_provider` already unpacks — no new cross-tool surface. ✓

### Catalog surfacing — correctly gated

Three edits, all guarded by `_provider_has_key(...)` **and** `_is_plugin_model_provider(...)`, and the detection loop is wrapped `try/except → debug`:
1. keyed plugin providers enter `detected_providers` even without a `providers.<slug>` block;
2. `raw_models` seeds from the plugin's `ProviderProfile.fallback_models` when empty (curated, network-free) — the live `/v1/models` rebuild supersedes it next call;
3. plugin groups survive the build loop **and** the post-loop filter even with zero models, so a keyed-but-not-yet-populated plugin shows as installed instead of vanishing.

So only an installed-and-keyed plugin surfaces — no unkeyed noise, and a plugin-detection failure degrades to today's behavior (debug log, not a raised error). ✓

### Edge cases

| Scenario | Behavior |
|---|---|
| Keyed plugin, cold cache, no models yet | group visible (installed), not dropped ✅ |
| Keyed plugin with `fallback_models` | curated subset surfaced network-free ✅ |
| Select `gh/claude-sonnet-4.5` under a plugin provider | `@<plugin>:gh/…` → routes to plugin ✅ |
| openrouter/custom/nous slash-model | bare passthrough unchanged ✅ (harnessed) |
| Unkeyed plugin | not surfaced ✅ |
| Plugin detection throws | debug log, catalog unaffected ✅ |
| No plugins installed (my env) | `is_plugin_model_provider` False → no behavior change ✅ |

### Tests

- `tests/test_plugin_model_providers.py` — **9/9** (catalog surfacing + `fallback_models` from #5461, plus the new `test_plugin_only_model_routes_to_plugin_provider` proving `model_provider=<plugin>, model='gh/…'` → `@<plugin>:gh/…`).
- **RED-on-master verified independently:** 3 fail against master's `config.py`, including the routing test (returns the bare `gh/…` without the branch). Load-bearing.
- Catalog/provider blast-radius — **216 passed, 5 skipped**.
- **Full suite: 12524 passed / 0 failed** in 280s (deselected the known darwin CRLF flake).

- A zero-model plugin group is briefly visible on the cold path before the live rebuild populates it — an intentional trade-off (show "installed" vs. vanish); the live `/v1/models` fetch fills it on the next call.

### Recommendation

A well-scoped two-part fix: keyed plugin providers now surface on the cold-cache path (gated by key + plugin-registry membership, fail-open to debug), and a selected plugin model routes to its plugin via the established `@provider:model` hint — placed so it can never hijack a standard slash-model (verified via `is_plugin_model_provider`'s static-provider exclusion). Routing test RED on master; full suite clean. Ship.

nesquena-hermes:
Shipped in **exp-v0.52.34** (experimental channel) 🎉 — plugin-provider routing fix. Thanks @alexfoxtm! (closes this salvage's issue)

PR #4627, opened 2026-06-21
feat: add 'max' reasoning effort level for DeepSeek models

perejaslav:
## Summary

DeepSeek models (deepseek-v4-flash, deepseek-reasoner) support `reasoning_effort: "max"` as their highest reasoning level — distinct from OpenAI's `"xhigh"`. The Hermes WebUI currently only offers options up to `xhigh`, so DeepSeek users cannot explicitly select `max` even though the underlying provider plugin already maps `xhigh → max` on the wire.

This PR adds `"max"` as a first-class valid reasoning effort level, making it selectable in the dropdown, via `/reasoning max`, and in autocomplete.

### `api/config.py`
- Add `"max"` to `VALID_REASONING_EFFORTS`
- Remove the blanket `max → xhigh` coercion in `coerce_reasoning_effort_for_model` — now `max` passes through as itself
- Existing degrade ladder still handles models that don't support `max` (OpenAI GPT-5, o1/o3/o4) by degrading to `xhigh`
- Update docstrings and comments

### `static/index.html`
- Add `Max` option to reasoning dropdown (between Extra High and the end)

### `static/ui.js`
- Add `'Max' label formatter for the reasoning chip

### `static/commands.js`
- Add `'max'` to the `/reasoning` valid efforts list, help text, and autocomplete subArgs

### Tests updated
- `test_issue1103_reasoning_chip_visibility.py`: assert `max` IS present
- `test_reasoning_effort_model_capabilities.py`: assert `max` preserved for unknown models
- `test_reasoning_show_hide.py`: include `max` in CLI-parity checks, remove exclusion assertions

## Model compatibility

- **DeepSeek** (opencode-go, native, OpenRouter): ✅ `max` passes through, provider plugin sends as-is
- **Anthropic / Qwen 3+ / Kimi / Gemini 2.5+**: ✅ `max` passes through
- **OpenAI (openai-codex gpt-5)**: ✅ `max` filtered out by provider filter, degrades to `xhigh`
- **Unknown/custom providers**: ✅ `max` preserved (same as other unknown-model efforts)

## Testing

All 13 reasoning-effort unit tests pass. Verified:
- `max` is available for DeepSeek models (`deepseek-v4-flash`, `deepseek-reasoner`)
- `max` is NOT available for OpenAI models that cap at `xhigh`
- Degrade ladder correctly steps down: `max → xhigh → high → medium → low → minimal`
- Stale config values and unknown models handle `max` gracefully

## Companion change needed (separate PR)

[...]

PR #5910, opened 2026-07-11
feat: add max reasoning effort level (salvage of #4627)

nesquena-hermes:
## Summary

Salvage-rebuild of #4627 (thanks @perejaslav for the original work) onto current `master`.

This closes an **Anchor-A parity gap**: Hermes core already supports `reasoning_effort='max'` (present in `hermes_constants.VALID_REASONING_EFFORTS`), but the WebUI capped the visible/settable set at `xhigh`. Users on models that support `max` could not select it from the WebUI.

Original authorship is preserved via `Co-authored-by: perejaslav`.

## Changes

- **`api/config.py`**
  - Add `max` to `VALID_REASONING_EFFORTS` (mirror of core).
  - Remove the blanket `max -> xhigh` coercion in `coerce_reasoning_effort_for_model()` so `max` is a first-class level.
  - **Keep the degrade ladder** (`max -> xhigh -> high`) for models/providers that cap below `max`, so capped models still degrade gracefully instead of erroring or silently dropping reasoning.
  - Update `set_reasoning_effort` / `get_reasoning_status` docstrings and comments.
- **`static/index.html`** — add the `Max` option (`data-effort="max"`) to the composer reasoning dropdown.
- **`static/ui.js`** — chip label mapping for `max` -> `Max`.
- **`static/commands.js`** — `/reasoning` command arg list, autocomplete `subArgs`, valid `EFFORTS`, and help string all include `max`.

## Re-gate / safety

- The provider filter `_filter_reasoning_efforts_for_provider()` still **excludes `max` for openai-codex GPT-5**, so those models degrade `max -> xhigh` rather than sending an unsupported level.
- Codex `o1/o3/o4` still clamp `max -> high` via the ladder.
- The `VALID_REASONING_EFFORTS` snapshot test and the JS command-parity tests were updated to include `max` (assertion drift resolved on current test files).
- No new i18n keys were introduced (the `Max` label follows the existing hardcoded pattern of `Minimal`/`Extra High`).

## Testing

```
python -m pytest tests/ -q -k 'reasoning or effort'
304 passed, 1 skipped, 12417 deselected
```

Salvaged from #4627. Do not close #4627 here — the parent tracks that.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -3267,17 +3267,33 @@ def _is_pre_adaptive_anthropic(bare_model: str) -> bool:
     """True for Claude models that predate the adaptive-thinking (4.6+) generation.
 
     Adaptive models (Opus/Sonnet 4.6+, 4.7, …) accept 'max'; earlier manual-thinking
-    Claudes (3.5/3.7/4.0–4.5) do not and must degrade 'max' to xhigh rather than
+    Claudes (3.x and 4.0–4.5) do not and must degrade 'max' to xhigh rather than
     fall through the manual-thinking budget table to its 8k default.
+
+    Handles the ID shapes the Anthropic adapter uses:
+      - claude-3-opus / claude-3-5-sonnet / claude-3-7-sonnet   → pre-adaptive
+      - claude-sonnet-4-5 / claude-haiku-4-5                     → pre-adaptive (4.5)
+      - claude-sonnet-4-20250514 (date-stamped 4.0 build)        → pre-adaptive
+      - claude-opus-4.6 / claude-sonnet-4.6 / claude-opus-4.7    → adaptive
+      - claude-opus-latest / unversioned                        → adaptive (flagship)
     """
+    import re as _re
     m = (bare_model or "").lower()
     if "claude" not in m:
         return False
-    # Extract a version like 4.6 / 4-5 / 3.7 from the model name.
-    import re as _re
-    match = _re.search(r"(\d+)[.\-](\d+)", m)
+    # Claude 3.x family is always pre-adaptive.
+    if _re.search(r"claude-3\b", m) or _re.search(r"claude-3[.\-]", m):
+        return True
+    # Find a major[.-]minor version. A date-stamp (>=6 digits) is NOT a minor
+    # version — treat "4-20250514" as major 4 with no real minor (a 4.0 build).
+    match = _re.search(r"(\d+)[.\-](\d{1,2})(?!\d)", m)
     if not match:
-        # Unversioned/opus-latest style — treat as adaptive (current flagship).
+        # A bare major like "claude-sonnet-4" or "...-4-20250514" (date stamp
+        # consumed no minor): major-4 with no minor is a pre-adaptive 4.0 build.
+        major_only = _re.search(r"[-.](\d+)(?:[-.]\d{6,})?(?:\b|-)", m)
+        if major_only:
+            return int(major_only.group(1)) < 5  # 4.x (no minor) → pre-adaptive; ≥5 → adaptive
+        # Unversioned / *-latest → treat as current flagship (adaptive).
         return False
     major, minor = int(match.group(1)), int(match.group(2))
     return (major, minor) < (4, 6)
--- a/api/config.py
+++ b/api/config.py
@@ -3026,7 +3026,7 @@ def get_effective_default_model(config_data: dict | None = None) -> str:
 # importing from the agent tree (which may not be installed).  Any drift here
 # will show up in the shared test suite since both sides accept the same set.
 # Keep this WebUI-visible set aligned with hermes-agent#29248.
-VALID_REASONING_EFFORTS = ("minimal", "low", "medium", "high", "xhigh")
+VALID_REASONING_EFFORTS = ("minimal", "low", "medium", "high", "xhigh", "max")
 
 
 def parse_reasoning_effort(effort):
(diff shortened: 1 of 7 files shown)


nesquena (approved):

Independent review of the salvage of #4627 (@perejaslav): add `max` as a first-class reasoning-effort level in the WebUI. Closes an Anchor-A parity gap — core already has `max` in `VALID_REASONING_EFFORTS` but the WebUI capped the settable set at `xhigh`, so DeepSeek/other max-capable models couldn't select it.

### What this ships

`api/config.py` (add `max`, drop the blanket `max→xhigh` coercion, keep the degrade ladder), `static/index.html` / `ui.js` / `commands.js` (dropdown option, chip label, `/reasoning` arg/autocomplete/help), 3 test files updated. Agent-authored (`nesquena-hermes`). MERGEABLE.

Fresh tarball confirms the parity gap is real: `hermes_constants.py:794` `VALID_REASONING_EFFORTS = (…, "xhigh", "max")`. Core accepts `max`; the WebUI was the one lagging. Adding it aligns the WebUI-visible set with core (as the in-code comment intends). ✓

### The safety mechanism — verified end-to-end, not just claimed

The load-bearing question: does removing the blanket `max→xhigh` coercion ever send `max` to a provider that rejects it? No — two layers still clamp, and I harnessed both with real provider context (my first no-provider harness returned `max` for everything because `resolve_model_reasoning_efforts` is empty without a provider; the real send path always has one):

**Provider filter** (`_filter_reasoning_efforts_for_provider`):
- `gpt-5` + `openai-codex` → `[minimal…xhigh]` (max excluded) ✓
- `o1/o3/o4` + `openai-codex` → `[low, medium, high]` ✓
- `deepseek-reasoner` + `deepseek` → full set incl. `max` ✓

**End-to-end coercion** (`coerce_reasoning_effort_for_model`, which routes through the filter via `resolve_model_reasoning_efforts`, then walks the ascending ladder `minimal…xhigh…max` DOWN to the closest supported level):
- `max` on gpt-5/codex → **`xhigh`** ✓ (PR claim)
- `max` on o3/codex → **`high`** ✓ (PR claim)
- `max` on deepseek → **`max`** ✓
- `xhigh` on gpt-5/codex → `xhigh` (unchanged) ✓

So a capped model degrades to its closest *lower* supported level (never escalates, never disables reasoning, never sends an unsupported level). The unknown-model case (`resolve_model_reasoning_efforts` empty) preserves the user's value verbatim — the same behavior as pre-PR for any other level, now consistent for `max` too.

### Edge cases

| Scenario | Behavior |
|---|---|
| DeepSeek, select `max` | `max` sent (was impossible from WebUI) ✅ |
| GPT-5 codex, select `max` | degrades → `xhigh` ✅ |
| o3 codex, select `max` | degrades → `high` ✅ |
| Unknown model, `max` | preserved verbatim (consistent with other levels) ✅ |
| Stale `reasoning_effort: max` in config | now parser-valid; coerced per model at read (chip/status agree with send) ✅ |

### Tests

- `test_issue1103_reasoning_chip_visibility.py` + `test_reasoning_effort_model_capabilities.py` + `test_reasoning_show_hide.py` — **61/61**; the `VALID_REASONING_EFFORTS` snapshot + JS command-parity assertions updated to include `max`.
- reasoning/effort blast-radius — **303 passed, 6 skipped**.
- `node --check` clean on ui.js + commands.js; JS parity verified (`data-effort="max"` in index.html, chip mapping in ui.js, `max` in commands.js arg/subArgs/EFFORTS/help).
- **Full suite: 12521 passed / 0 failed** in 284s (deselected the known darwin CRLF flake).

- `Max` label follows the existing hardcoded `Minimal`/`Extra High` pattern (no new i18n key) — consistent with the codebase, matches the PR note.

### Recommendation

A correct parity fix: `max` becomes first-class (matching core), the blanket coercion is removed, and the provider filter + degrade ladder still guarantee capped models never receive an unsupported level — verified end-to-end with real provider context (gpt-5→xhigh, o3→high, deepseek→max). JS parity complete; full suite clean. Ship.

nesquena-hermes:
Shipping this to the experimental channel (availability-gated). For the record, one residual edge is knowingly accepted (maintainer sign-off):

- **Direct providers:** `Max` is offered ONLY when the model genuinely supports it (adaptive Claude 4.6+, DeepSeek) and hidden/degraded everywhere it would be rejected (GPT-5, o-series, Gemini, legacy + cloud-hosted Claude). A stale/CLI `max` degrades to `xhigh`. ✅
- **Aggregator providers (OpenRouter / Copilot / multi-model OpenAI) with no per-model capability data:** `is max available?` has no reliable answer, so `max` stays selectable and would fall back at the provider if that specific model doesn't support it — the same behavior as the CLI. This is the single edge that 5 gate rounds converged on; it's an irreducible ambiguity (denying it there would break `max` for legitimately max-capable OpenRouter models), and it's no worse than today's CLI behavior.

Net: `max` now works correctly for every direct provider (it didn't work at all before), with safe degradation. Thanks @perejaslav for the original.

nesquena-hermes:
Shipped in **exp-v0.52.33** (experimental channel) 🎉

The `Max` reasoning level is live — selectable only when the active model genuinely supports it (adaptive Claude 4.6+, DeepSeek), and safely degraded/hidden everywhere it isn't. Thanks @perejaslav for the original #4627. (closes #5910)

Issue #5472, opened 2026-07-03
bug: provider/background error on send can drop the user's message with no retry — preserve draft + add retry-failed-message affordance


nesquena-hermes:
**From a user report on Reddit (surfaced via #webui).**

> There are some bugs I noticed in your WebUI. When I send a message, if there are some issues from API provider side, or etc. instead of showing the error, it deletes the entire message I sent.
>
> It's frustrating when I typed and sent a long message but due to background errors it removes the entire message from chat window. I have to type the entire message again.

Two things a user loses here: (1) the just-sent message can disappear from the transcript on a provider/background error, and (2) even when it survives, there is no one-click way to re-send it — the composer was already cleared, so the user retypes. This issue tracks both, with a **retry-failed-message** affordance as the headline fix.

## Legitimacy / current-master status

Recent hardening has narrowed this significantly, so exact repro details matter (see needinfo below):

- The server now materializes the pending user turn **before** appending any error marker, on both terminal error paths — `_materialize_pending_user_turn_before_error()` at `api/streaming.py` (provider-error settlement path ~L8556 and exception path ~L9607). Under the default `session_save_mode: deferred`, `pending_user_message` is the *only* durable copy of the current turn until the agent result merges, so this materialize-before-error step is what keeps the user bubble from vanishing on reload/reconcile.
- `#5129` (fixes `#5121`, shipped v0.51.728) fixed the mirror-image case — a terminal provider failure that finalized the turn "done" with an empty body so the transcript pretended the assistant never responded.

So the pure "message silently deleted, no error at all" symptom is largely addressed on current master (v0.51.835). The report reads as either an **older build** or a **residual timing edge** we should confirm — hence the needinfo.

## Root cause of the "have to retype" half (code-grounded, not speculative)

Independent of whether the bubble survives, the retype pain is real and reproducible in the code today:

- The composer is cleared unconditionally right after capture on send — `static/messages.js:1498` (`$('msg').value='';autoResize();`). There is **no draft-restore on any error path**. The `/api/chat/start` throw handler (`messages.js:~1623`) and the SSE `apperror`/`error`/`_handleStreamError` handlers all push an error marker but never put the drafted text back in the composer.
[...]

PR #5912, opened 2026-07-11
fix(send): clear composer immediately after capture to prevent stalled-text race (salvage of #4750)

nesquena-hermes:
## Salvage of #4750 — atomic composer fix only

This PR salvages **only** the atomic composer capture-and-clear fix that was buried inside #4750 by @harryazj. The original PR was a 20-file downstream-fork bundle (CN-mirror sync workflows, README rewrites, i18n churn, upstream-watch scripts, etc.) that we do not want on `master`. This branch discards all of that and rebuilds just the one-hunk fix from a clean `master`, preserving attribution to @harryazj via `Co-authored-by`.

### The bug

`send()` in `static/messages.js` captured the composer text early but only wiped the textarea (`$('msg').value=''`) **late** — after `await uploadPendingFiles()` and the forced-skill-directive await.

`send()` is re-entrant: while a send is in flight, a second invocation (an interrupt-mode / `busy_input_mode=interrupt` queue drain, or a fast second Enter) hits the `if (_sendInProgress)` guard at the top of the function, which re-reads the **live** composer via `_composerTextWithPendingSelections()`. Because the textarea still held the original text during the async upload window, the re-entrant read picked up the stale DOM text and **queued the same message again → double-submit**.

### The fix

Capture the composer value and clear the textarea **immediately after capture, before any `await`**. Once cleared, a re-entrant read sees an empty composer and the guard's `if(_text && _targetSid)` short-circuits — no duplicate queue.

- Moved the `const _submittedDraftTextForClear=$('msg').value||'';` capture + `$('msg').value='';autoResize();` wipe up to run right after `_sendInProgressSid=activeSid;`, ahead of `uploadPendingFiles()`.
- The captured text still feeds the persisted-draft clear and the #5472 failed-send restore path exactly as before, so no behavior regresses.

### Tests

- **New** `tests/test_composer_capture_clear_race.py` — static ordering assertions (capture+clear before the upload await) **plus** a node-`vm` behavioral harness that runs the *real* re-entrancy guard block extracted from `send()`: it proves the guard does **not** re-queue when the composer was already cleared, and *would* have double-submitted had it not been (bug demonstration).
- **Updated** `tests/test_issue5472_preserve_draft_on_failed_send.py::test_send_still_clears_composer_on_the_happy_path` to reflect the new (earlier) clear ordering while still asserting the persisted-draft clear + files-snapshot wiring.

### Verification

```
pytest tests/ -k 'composer or send or messages or textarea'
602 passed, 1 skipped, 12123 deselected
```

### Not included (discarded from #4750)

`.github/workflows/sync-to-cnb-mirror.yml`, `upstream-watch.yml`, `MAINTAINING.md`, `README*.md`, `scripts/sync-upstream.sh`, `static/i18n.js` churn, and all other fork-plumbing files. Only `static/messages.js` + the two test files are touched here.

Thanks @harryazj for finding the underlying race.

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -1552,6 +1552,20 @@ async function send(){
   const activeSid=S.session.session_id;
   _sendInProgressSid=activeSid;
 
+  // Salvage of #4750 (@harryazj): capture the composer text and clear the
+  // textarea NOW — immediately after capture and BEFORE the uploadPendingFiles()
+  // / forced-skill-directive awaits below. send() re-reads the LIVE composer when
+  // it is re-entered while a send is in flight (the _sendInProgress guard at the
+  // top of this function reads _composerTextWithPendingSelections()). If we
+  // cleared only after the async work — as the pre-fix code did, down at the
+  // _clearComposerDraft site — a re-entrant/interrupt-mode send during the upload
+  // window would read the still-populated DOM and double-submit the same message.
+  // _submittedDraftTextForClear is the sole authority for the send-time draft
+  // signature from here down; no code path below re-reads $('msg').value on the
+  // happy path.
+  const _submittedDraftTextForClear=$('msg').value||'';
+  $('msg').value='';autoResize();
+
   setComposerStatus(S.pendingFiles&&S.pendingFiles.length?'Uploading…':'');
   let uploaded=[];
   try{uploaded=await uploadPendingFiles();}
@@ -1589,9 +1603,10 @@ async function send(){
     }
   }
   if(!msgText){setComposerStatus('Nothing to send');return;}
-  const _submittedDraftTextForClear=$('msg').value||'';
+  // Composer textarea + _submittedDraftTextForClear were already captured and
+  // cleared immediately after capture (above, salvage of #4750) to close the
+  // re-entrant double-send race. Only the files snapshot is finalized here.
   const _submittedDraftFilesForClear=Array.isArray(_failedSendFilesSnapshot)?[..._failedSendFilesSnapshot]:[];
-  $('msg').value='';autoResize();
   // Clear persisted composer draft since message was sent. Capture the promise
   // so the #5472 failed-send restore can chain its re-persist AFTER this clear
   // resolves — otherwise the two same-origin POSTs (clear text:'' then restore
--- a/static/messages.js
+++ b/static/messages.js
@@ -1566,9 +1566,27 @@ async function send(){
   const _submittedDraftTextForClear=$('msg').value||'';
   $('msg').value='';autoResize();
 
-  setComposerStatus(S.pendingFiles&&S.pendingFiles.length?'Uploading…':'');
+  // #5912 gate CORE fix: snapshot the pending files that belong to THIS send
+  // BEFORE the await, and upload exactly that snapshot. Otherwise a re-entrant /
+  // interrupt-mode send during the upload window inherits the still-live
+  // S.pendingFiles and later re-uploads the first send's attachment. Detach the
+  // snapshot from S.pendingFiles now so files staged AFTER this point belong to
+  // the next send only.
+  const _submittedFiles=[...(S.pendingFiles||[])];
+  const _submittedDraftFilesForClear=_submittedFiles.map(f=>(f&&f.name)||'').filter(Boolean);
+  S.pendingFiles=[];
+  if(typeof renderTray==='function')renderTray();
+
+  // #5912 gate SILENT fix: clear the PERSISTED draft here — alongside the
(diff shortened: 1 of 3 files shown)


nesquena (approved):

Independent review of the salvage of #4750 (@harryazj — the atomic composer fix only, extracted from a 20-file fork bundle): clear the composer textarea **immediately after capture, before any `await`**, to close a re-entrant double-submit race.

### What this ships

`static/messages.js` (one hunk moved up), `tests/test_composer_capture_clear_race.py` (+new), `tests/test_issue5472_preserve_draft_on_failed_send.py` (ordering assertion updated). Agent-authored (`nesquena-hermes`). MERGEABLE, CI green. Deliberately excludes all the fork-plumbing files from #4750.

### The race — confirmed in the code

`send()` is re-entrant: its top guard (`if (_sendInProgress)`) re-reads the **live** composer via `_composerTextWithPendingSelections()` and queues it. Pre-fix, `send()` captured `text` early (line 28/33) but wiped `$('msg').value` **late** — after `await uploadPendingFiles()`. During that upload window a second invocation (interrupt-mode queue drain / fast second Enter) read the still-populated textarea and re-queued the same message → double-submit. Verified the guard path reads the DOM (`messages.js` head) and that the old clear sat past the awaits. ✓

### The fix — traced safe on the happy path

Capture + clear now run right after `_sendInProgressSid=activeSid`, before the awaits. I confirmed nothing downstream depends on the post-clear DOM:
- **`msgText` derives from the earlier `text` capture** (`let msgText=text`, where `text=$('msg').value.trim()` at line 28/33, *before* the clear) — **not** a re-read of `$('msg').value`. The only other reads in the window are the `catch(e){if(!text)…}` (uses `text`) and `displayText` (uses `text`). No post-clear DOM read on the happy path. ✓
- **The empty guards already returned** (lines 29/34) before the clear point, and all slash-command paths (which do their own `$('msg').value` clears + early returns) run *before* the insertion point — so the clear is only reached for a real, non-empty message send. The `if(!msgText) 'Nothing to send'` can't fire from an empty composer after the clear, because `text` was non-empty to get here. ✓
- **`_submittedDraftTextForClear`** (captured before the clear) remains the sole authority for the draft-clear (`_clearComposerDraft(activeSid, _submittedDraftTextForClear, …)`) and the **#5472 failed-send restore** — so a failed send still restores the exact submitted text. The #5472 HTTP/2-reorder promise-chaining is untouched. ✓
- Re-entrancy is closed: after the early clear, the top guard's `_composerTextWithPendingSelections()` reads empty → `if(_text && _targetSid)` short-circuits → no duplicate queue. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Fast second Enter during upload window | second read sees empty composer → no double-submit ✅ (was the bug) |
| interrupt-mode queue drain mid-send | same — no re-queue ✅ |
| Happy-path send | `msgText` from pre-clear `text`; sends normally ✅ |
| Failed send (#5472) | composer restored from `_submittedDraftTextForClear` ✅ (test) |
| Empty composer / slash command | early-return before the clear (unchanged) ✅ |
| Upload error | `catch` uses `text`, not DOM ✅ |

### Tests

- `tests/test_composer_capture_clear_race.py` (new) — static ordering locks (capture+clear before the upload await) **plus** a node-`vm` behavioral harness running the real re-entrancy guard block: proves it does **not** re-queue when the composer was already cleared, and *would* have double-submitted otherwise. Updated `test_issue5472_...::test_send_still_clears_composer_on_the_happy_path` for the earlier ordering while still asserting the persisted-draft clear + files-snapshot wiring. **21/21.**
- **RED-on-master verified independently:** `test_composer_captured_and_cleared_before_upload_await` fails against master's `messages.js`. Load-bearing.
- send/composer/draft blast-radius — **345 passed, 5 skipped**.
- `node --check` clean. **Full suite: 12525 passed / 0 failed** in 281s (deselected the known darwin CRLF flake).

### Recommendation

A minimal, correct fix for a real re-entrant double-submit: clearing the composer before the async upload window means a concurrent `send()` reads an empty composer and short-circuits. Traced safe on the happy path (everything downstream uses the pre-clear `text`/`_submittedDraftTextForClear`, never a post-clear DOM read), and the #5472 failed-send restore is preserved. Behavioral harness proves the guard both fixes and would-have-reproduced the bug; RED on master; full suite clean. Ship.

nesquena-hermes:
Shipped in **exp-v0.52.34** (experimental channel) 🎉 — composer double-send + attachment-race fix. Thanks @harryazj! (closes this salvage's issue)

PR #5915, opened 2026-07-11
fix(#4732): keep mobile composer-mic dictation active through pauses (salvage of #5294)

### LINKED ISSUE #4732: bug(ux): mobile dictation stops on thinking pauses (SpeechRecognition not continuous) + no wake lock

nesquena-hermes:
## Summary

Salvage-rebuild of #5294 (thanks @brianmmaina) — keeps the **composer microphone** dictation session (`SpeechRecognition` in `_ensureSpeechRecognition`) alive through natural pauses on **mobile**, while **desktop dictation stays one-shot** (a single utterance that finalizes on the first pause).

The original #5294 turned on `continuous`/restart unconditionally, which would have regressed desktop's single-utterance behavior. This branch is a fresh rebuild on current `master` that applies the maintainer's gate + race + error-path fix-spec.

Closes #4732.

## What changed (`static/boot.js`)

**Cluster 1 — mobile-gated continuity**
- `sr.continuous` is now gated: `_micDictationContinuous()` returns `true` only on a coarse-pointer (touch) device, with an explicit `hermes-mic-continuous` localStorage override in both directions (mirrors the existing `hermes-voice-continuous` pattern). **Desktop stays single-utterance.**
- `onend` restarts the recognizer on a natural pause only when continuity is enabled, the session is still active speech, and no intentional stop was requested — via a `_speechStopRequested` guard set before any intentional `recognition.stop()` in `_stopMic`.

**Cluster 2 — screen wake lock, race-serialized**
- `_acquireMicWakeLock` / `_releaseMicWakeLock` + `visibilitychange` reacquire, but all acquire/release ops are chained onto a single in-flight promise (`_micWakeLockOp`) so rapid start/stop/visibility churn can't interleave to leak a lock or null `_micWakeLock` mid-request.

**Cluster 3 — error path**
- `onerror` silently skips `no-speech`/`aborted` while actively dictating (natural pause), and adds `service-not-allowed`/`audio-capture` to the fatal MediaRecorder-fallback set.
- Greptile P2s folded in: a toast is shown if `sr.start()` restart fails (no silent drop to idle), and the abort/restart loop is bounded (`_micMaxRestarts=20`) so a stolen audio session can't spin a tight loop.

## Tests

New `tests/test_issue5294_mobile_dictation_continuity.py` drives the real `_ensureSpeechRecognition` + gate helpers via the node-eval `extractFunc` harness and pins **both** directions:
- **Desktop** → `continuous=false`, no auto-restart on a pause, session ends.
- **Mobile** → `continuous=true`, restarts on a natural pause, session stays active; an intentional stop still finalizes.

```
379 passed, 1 skipped, 12366 deselected
```
(plus `node --check static/boot.js` clean)

## Follow-up

- [ ] Needs a **mobile screenshot / UX check** on a real touch device to confirm the dictation-through-pauses flow and wake-lock behavior feel right.

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -919,6 +919,13 @@ function _micToastKeyForRecognitionError(error){
     sr.onstart=()=>{ _finalText=''; };
 
     sr.onresult=(event)=>{
+      // #5294: a real result means the continuity restarts are PRODUCTIVE, not a
+      // stolen-audio-session tight loop — reset the restart budget so a long
+      // dictation with many natural pauses isn't silently capped at
+      // _micMaxRestarts. The cap still guards the failure case: consecutive
+      // restarts that yield no speech (onend without an intervening onresult)
+      // keep incrementing and trip the bound.
+      _micRestartCount=0;
       let interim='';
       let final=_finalText;
       for(let i=event.resultIndex;i<event.results.length;i++){
--- a/static/boot.js
+++ b/static/boot.js
@@ -675,6 +675,17 @@ function _micToastKeyForRecognitionError(error){
   let _finalText='';
   let _prefix='';
   let _isRecording=false;
+  // #5294 salvage — mobile composer-mic dictation continuity.
+  // _speechStopRequested distinguishes an intentional stop (send/toggle) from a
+  // natural pause so onend only auto-restarts on real pauses. _micWakeLock keeps
+  // the screen awake while dictating; _micWakeLockOp serializes acquire/release
+  // so rapid start/stop/visibility churn can't leak a lock. _micRestartCount is
+  // bounded by _micMaxRestarts to stop a tight loop if the audio session is stolen.
+  let _speechStopRequested=false;
+  let _micWakeLock=null;
+  let _micWakeLockOp=null;
+  let _micRestartCount=0;
+  const _micMaxRestarts=20;
   let _micHoldTimer=null;
   let _micHoldActive=false;
   let _micPointerDown=false;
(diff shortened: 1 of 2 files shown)


nesquena (approved):
## Review — end-to-end ✅ (approved after one fix I pushed)

Independent review of the salvage of #5294 (@brianmmaina, closes #4732): keep the composer-mic dictation session alive through natural pauses on **mobile**, while **desktop stays one-shot**. The original #5294 turned on `continuous`/restart unconditionally (would have regressed desktop); this rebuild gates it and applies the maintainer race/error fix-spec.

### What this ships

`static/boot.js` (mobile-gated continuity + wake-lock + error path), `tests/test_issue5294_mobile_dictation_continuity.py` (+new). Agent-authored (`nesquena-hermes`). MERGEABLE.

Fresh tarball: WebUI-only, browser `SpeechRecognition`/`WakeLock` plumbing; no agent surface. ✓

### What I caught — the restart budget never reset on a productive result

`_micRestartCount` bounds the continuity auto-restart loop at `_micMaxRestarts=20` "so a stolen audio session can't spin a tight loop" — good. But it only reset on **fresh-start / terminal-end / error**, never on a real `onresult`. So a long mobile dictation with 20 natural pauses (each producing speech) hit the cap **mid-session** and the mic silently stopped — directly undercutting the feature's purpose ("keep dictation active through pauses"). I confirmed there was no `_micRestartCount` reset in `sr.onresult`.

**Pushed `762fd01a`:** reset `_micRestartCount=0` in `sr.onresult`. A real result proves the restarts are *productive*, not a failure spin — so the cap now guards only the failure case (consecutive restarts yielding no speech, i.e. `onend` with no intervening `onresult`, keep incrementing and trip the bound). Added a regression test that seeds the count **at** the cap, fires a productive `onresult`, and asserts the ensuing pause still restarts — **verified fail-without / pass-with** against master's `boot.js`.

### The rest — traced correct

- **Mobile gate** (`_micDictationContinuous`): explicit `hermes-mic-continuous` localStorage flag wins both directions ('true' opts desktop in, 'false' opts mobile out), else `matchMedia('(pointer:coarse)')`; wrapped in try/catch. Re-read at each `_startMic` so a settings/orientation change takes effect. Desktop stays one-shot. ✓
- **Restart guard** (`_micShouldRestartDictation`): continuity ON && `!_speechStopRequested` && active speech && under the (now-productive-reset) cap. `_speechStopRequested` is set before every intentional `stop()`, so send/toggle finalizes instead of restarting. ✓
- **Wake-lock race-serialized**: `_acquireMicWakeLock`/`_releaseMicWakeLock` both chain onto a single `_micWakeLockOp` promise, so rapid start/stop/visibility churn can't interleave to leak a lock or null it mid-request; `visibilitychange` reacquire routes through the same chain. ✓
- **Error path**: `no-speech`/`aborted` swallowed while actively dictating (natural pause) so `onend` restarts; `service-not-allowed`/`audio-capture` added to the fatal MediaRecorder-fallback set; a failed `sr.start()` restart shows a toast (no silent drop to idle). ✓
- **No XSS**: the mic-error toast interpolates a browser `SpeechRecognition` error string, and `showToast`'s error variant `esc()`s it anyway. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Desktop, natural pause | one-shot, finalizes (no restart) ✅ |
| Mobile, natural pause | restarts, session stays active ✅ |
| Mobile, intentional stop (send/toggle) | `_speechStopRequested` suppresses restart → finalizes ✅ |
| Long mobile dictation, 20+ productive pauses | keeps restarting (budget resets on result) ✅ (my fix, tested) |
| Stolen audio session, unproductive restarts | capped at 20, no tight loop ✅ |
| Desktop opt-in flag / mobile opt-out flag | honored both directions ✅ |
| Rapid start/stop/visibility churn | wake-lock ops serialized, no leak ✅ |
| `service-not-allowed`/`audio-capture` | fatal → MediaRecorder fallback ✅ |

### Tests

- `tests/test_issue5294_mobile_dictation_continuity.py` — **8/8** (desktop one-shot both ways, mobile restart-on-pause, intentional-stop-finalizes, opt-in/opt-out flags, **+ my productive-result-resets-budget regression**, verified load-bearing).
- **Full suite: 12546 passed / 0 failed** in 397s (deselected the known darwin CRLF flake).

### Minor / follow-up (non-blocking)

- The PR flags a **mobile screenshot / real-touch-device UX check** for the dictation-through-pauses + wake-lock feel — I verified logic/structure but not on-device behavior; that human check remains worthwhile (low code risk).

### Recommendation

✅ **Approved after fix.** Parked at approval — ready for the release agent's merge/tag pipeline (with the on-device UX check as a nice-to-have follow-up).

Mobile-gated continuity is correct (desktop stays one-shot), the wake-lock ops are race-serialized, and the error path degrades sensibly. The one real gap — the restart budget capping a *productive* long dictation at 20 pauses — I fixed in `762fd01a` (reset on `onresult`, preserving the stolen-session guard) with a mutation-verified regression test. Full suite clean. Ship.

nesquena-hermes:
Shipped in **exp-v0.52.34** (experimental channel) 🎉 — mobile dictation continuity. Thanks @brianmmaina! (closes this salvage's issue)

Issue #3020, opened 2026-05-27
Session completion unread dot persists after navigating away from actively-viewed session


mysoul12138:
## Problem

The unread notification dot on the right side of a session title does not disappear after clicking into the session and navigating away. It persists indefinitely.

## Root Cause

_setSessionViewedCount() only updates the viewed count at two points:

1. loadSession() metadata fetch (line 647) - uses API response's message_count
2. Message load completion (line 1317) - uses API response's message_count

When the user is actively viewing session A and the agent continues generating messages, polling updates _allSessions with the new message_count (e.g. 10 -> 12). However, _markPollingCompletionUnreadTransitions() detects new messages but _isSessionActivelyViewedForList() returns true, so it skips _markSessionCompletionUnread() - **but does not update the viewed count either**.

When the user navigates away, _hasUnreadForSession() compares s.message_count(12) > viewedCount(10) -> true -> unread dot shows.

The gap: when a session is actively viewed, the polling completion check skips setting the unread marker but also does not sync the viewed count to the current message count, causing a stale viewed count that triggers false unread on the next sidebar render.

## Key Code Paths

- static/sessions.js:166-179 - _markSessionCompletionUnread / _clearSessionCompletionUnread
- static/sessions.js:233-242 - _hasUnreadForSession (message count comparison)
- static/sessions.js:387-412 - _markPollingCompletionUnreadTransitions (skips actively viewed sessions without updating viewed count)
- static/sessions.js:647-648 - loadSession sets viewed count and clears completion unread

nesquena-hermes:
## Summary

Confirmed the analysis by reading `static/sessions.js` around the cited line ranges. The reporter is right: when polling observes new messages on an actively-viewed session, `_markPollingCompletionUnreadTransitions` correctly skips setting the unread marker but never advances the viewed count, so the moment the user clicks away the next sidebar render compares stale `viewedCount` against fresh `message_count` and flags the row as unread.

## Code reference

The skip-without-sync at `static/sessions.js:387-412`:

```js
if ((completedObservedStream || completedPersistedObservedStream || completedWithNewMessages)
    && !_isSessionActivelyViewedForList(sid)) {
  _markSessionCompletionUnread(sid, s.message_count);
}
_sessionStreamingById.set(sid, isStreaming);
```

[...]

PR #5917, opened 2026-07-11
fix(sessions): clear stale sidebar unread indicator on session visit (salvage of #4946)

nesquena-hermes:
## Salvage of #4946 — thanks @neaucode-bot 🙏

This is a **fresh rebuild on current master** of #4946 by @neaucode-bot, which
went stale (11+ days, CONFLICTING). I verified the bug **still reproduces on
master** before rebuilding — it is not superseded by the recent sidebar/unread
fixes.

## Repro confirmed on master

None of #4946's fix helpers exist on `master`, and both completion paths still
gate on the focus-based `_isSessionActivelyViewedForList`, while the
same-session no-op guard early-returns without clearing unread. So a session's
yellow unread dot can persist after the user actually opened/visited it: it only
reliably cleared on a sidebar **row** click, and a deferred `/api/sessions` list
poll running across the async message-load gap could re-flag the open session.

## The fix

- New `_acknowledgeSessionVisit(sid, messageCount, lastMessageAt)`: on visit it
  syncs the viewed count (which clears any stale completion-unread marker,
  #3020), syncs the polling snapshot (`_syncSessionListSnapshotOnVisit`) so a
  deferred list poll can't treat the unchanged open session as a fresh
  background completion, then repaints the sidebar from cache.
- `loadSession()` acknowledges the visit **when metadata arrives** and **again
  after the async message-load gap**, so a deferred poll can't leave a sticky
  dot.
- Re-selecting the already-open session (the same-sid no-op early return) now
  clears a stale unread state first via `_sessionVisitHasUnreadState()` before
  returning.

## Addressing the review feedback on #4946

Both blockers from the original review are handled:

- **R15c (`test_queue_card_cross_session_helper_used_only_for_session_change`)** —
  the same-session early return is now a block (to clear the dot before
  returning), and I updated the R15c anchors to match while keeping the
  invariant: `_clearQueueCardDisplay(currentSid)` stays inside the
  `if (currentSid && currentSid !== sid)` cross-session branch and never fires
  on same-session/force-reload paths.
- **Greptile concern (a): hidden-tab completions marked read** — unlike #4946,
  this rebuild does **not** loosen the completion paths' focus gate. The
  background + polling completion paths keep using the focus-gated
  `_isSessionActivelyViewedForList`, so a completion in a hidden/unfocused tab
  is still flagged unread. The sticky-dot race is instead closed by syncing the
  polling snapshot on visit.
- **Greptile concern (b): stripping a lineage parent's own dot** — the visit
  repaints via `renderSessionListFromCache()`, which recomputes every row's
  aggregated (own + children) unread state authoritatively, instead of doing
  ad-hoc DOM surgery that could temporarily hide a parent's own dot.

## Tests

- New `tests/test_session_unread_dot_on_visit.py`: structural anchors + **node
  functional tests** proving (1) visiting clears the completion-unread marker,
  syncs the snapshot and repaints; (2) a deferred poll after a visit does **not**
  re-flag the open session; and (3) the completion paths keep the focus gate for
  hidden-tab completions (concern a).
- Updated existing anchors that tracked the refactored `loadSession` structure:
  `test_issue856_background_completion_unread`, `test_session_metadata_fast_path`,
  `test_regressions` (R15c), `test_inflight_stream_reuse`,
  `test_session_channel_option_x`.

Local run (repo `.venv`, Python 3.11): `node --check static/sessions.js` OK;
`pytest -k 'unread or sidebar or sessions_list or session_unread'` →
**372 passed, 2 skipped** (skips are agent-dependent, unrelated).

Closes the same issue as #4946.

---
Original work by @neaucode-bot in #4946; rebuilt fresh on master here because
the original diff no longer applied cleanly.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -603,6 +603,47 @@ function _hasUnreadForSession(s) {
   return s.message_count > Number(counts[s.session_id] || 0);
 }
 
+// Keep the sidebar polling snapshot current for a just-visited session so a
+// deferred /api/sessions list refresh landing across the async message-load gap
+// cannot treat the unchanged, already-open session as a fresh background
+// completion and re-flag a stale unread dot (#4946).
+function _syncSessionListSnapshotOnVisit(sid, messageCount, lastMessageAt) {
+  if (!sid) return;
+  const count = Number(messageCount || 0);
+  const last = Number(lastMessageAt || 0);
+  _sessionListSnapshotById.set(sid, {message_count: count, last_message_at: last});
+  const isStreaming = Boolean(
+    S.session && S.session.session_id === sid
+    && (S.busy || S.activeStreamId || (S.session.active_stream_id && S.session.pending_user_message))
+  );
+  _sessionStreamingById.set(sid, isStreaming);
+  if (!isStreaming) _forgetObservedStreamingSession(sid);
+}
+
+// Acknowledge that the user actually visited/opened `sid`: clear its viewed
+// count (which also clears any stale completion-unread marker, #3020), sync the
+// polling snapshot so a deferred list poll cannot re-flag it, then repaint from
+// cache. Repainting via renderSessionListFromCache() recomputes each row's
+// aggregated unread state (own + children) authoritatively, so a lineage
+// PARENT keeps its own / other children's unread dot instead of being stripped
+// by ad-hoc DOM surgery (Greptile concern (b) on #4946).
+function _acknowledgeSessionVisit(sid, messageCount = 0, lastMessageAt = 0) {
+  if (!sid) return;
+  _setSessionViewedCount(sid, messageCount);
+  _syncSessionListSnapshotOnVisit(sid, messageCount, lastMessageAt);
+  if (typeof renderSessionListFromCache === 'function') renderSessionListFromCache();
+}
+
+// Does the session currently carry any unread state that a visit should clear?
+// Used by the same-session no-op guard so re-selecting the already-open session
+// still clears a stale dot before short-circuiting.
+function _sessionVisitHasUnreadState(sid) {
+  if (!sid) return false;
+  if (_hasSessionCompletionUnread(sid)) return true;
+  if (!S.session || S.session.session_id !== sid) return false;
+  return _hasUnreadForSession(S.session);
+}
+
 function _isSessionActivelyViewedForList(sid) {
   if (!sid || !S.session || S.session.session_id !== sid) return false;
   if (typeof _loadingSessionId !== 'undefined' && _loadingSessionId && _loadingSessionId !== sid) return false;
@@ -1425,7 +1466,19 @@ async function loadSession(sid){
   // #2971: idempotent re-arm before the no-op guard revives a stream a prior
   // failed loadSession killed; no-ops on real switches.
   _rearmActiveSessionStream();
-  if(currentSid===sid && !forceReload && (!_loadingSessionId || _loadingSessionId===sid)) return;
+  if(currentSid===sid && !forceReload && (!_loadingSessionId || _loadingSessionId===sid)){
+    // Re-selecting the already-open session is a no-op for transcript/scroll, but
+    // it is still a *visit*: clear a stale sidebar unread dot (e.g. one a
+    // background completion left on the open, unfocused pane) before returning.
+    if(_sessionVisitHasUnreadState(sid)){
+      _acknowledgeSessionVisit(
+        sid,
+        Number(S.session.message_count || 0),
+        Number(S.session.last_message_at || S.session.updated_at || 0)
+      );
+    }
+    return;
+  }
   // Mark this session as the in-flight load. Subsequent loadSession() calls
   // will overwrite this; stale awaits use the mismatch to bail out (#1060).
   const _loadGeneration = ++_loadSessionGeneration;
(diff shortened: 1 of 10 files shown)


nesquena (approved):
## Review — end-to-end ✅ (approved after two CI-red fixes I pushed)

Independent review of the salvage of #4946 (@neaucode-bot): clear a stale sidebar **unread dot** when the user actually visits/opens a session. The author verified the bug still reproduces on master (the recent unread fixes don't supersede it) before rebuilding.

### What this ships

`static/sessions.js` (`_acknowledgeSessionVisit` / `_syncSessionListSnapshotOnVisit` / `_sessionVisitHasUnreadState` + `loadSession` wiring), `tests/test_session_unread_dot_on_visit.py` (+new), and anchor updates to 5 existing tests. Agent-authored (`nesquena-hermes`). MERGEABLE.

### What I caught — CI was red, two real issues, both fixed

1. **Lint (F401):** the new `test_session_unread_dot_on_visit.py` imported `re` unused → ruff forward gate red. Dropped it.
2. **Test hang → 60s timeout** (`test_cross_session_message_load_isolation.py::test_loadsession_cross_session_ordering_and_stale_reject_behavior`, red across 3.11/3.12/3.13): `loadSession()` now routes its viewed-count/unread clear through the new `_acknowledgeSessionVisit()` helper, but that **pre-existing** test's node harness extracts `loadSession` and stubs its deps — and it stubbed the *old* calls (`_setSessionViewedCount`, `_clearSessionCompletionUnread`) that `_acknowledgeSessionVisit` replaced, not the new helper. So the extracted `loadSession` referenced an undefined function and the node driver hung. I stubbed `_acknowledgeSessionVisit` + `_sessionVisitHasUnreadState` to no-ops there (that harness tests cross-session load ordering + stale-reject, not unread state — matching the stubs they replaced; the real helper is exercised by the PR's own test).

**Pushed `da00a5ae`.** After it: ruff gate clean, the timeout test passes in ~1s, and CI is **fully green (18/18)**.

### End-to-end trace of the fix

- `_acknowledgeSessionVisit(sid, count, last)`: `_setSessionViewedCount` (also clears the completion-unread marker, #3020) → `_syncSessionListSnapshotOnVisit` (writes the poll snapshot so a deferred `/api/sessions` refresh landing during the async message-load gap can't treat the unchanged open session as a fresh background completion) → repaint via `renderSessionListFromCache()`.
- `loadSession` calls it **twice** — when metadata is accepted, and again after the `_ensureMessagesLoaded` gap (guarded by `S.session.session_id === sid`) — closing the race where a deferred poll re-flags the open session mid-load.
- Same-session re-select (the no-op early return) now clears a stale dot first via `_sessionVisitHasUnreadState` before returning.

### The two #4946 review concerns are genuinely addressed

- **(a) hidden-tab completions must stay unread:** verified the completion paths still gate on the focus-based `_isSessionActivelyViewedForList` (`sessions.js:525`, `:1038`) — this rebuild does **not** loosen that gate. A completion in a hidden/unfocused tab is still flagged unread; the sticky-dot race is closed by the snapshot sync, not by weakening the gate. ✓
- **(b) don't strip a lineage parent's own dot:** the visit repaints via `renderSessionListFromCache()`, which recomputes each row's aggregated (own + children) unread authoritatively — no ad-hoc DOM surgery that could transiently hide a parent's own dot. ✓
- **R15c:** the same-session early return is now a block (to clear the dot before returning); `_clearQueueCardDisplay(currentSid)` stays inside the `currentSid !== sid` cross-session branch and never fires on same-session/force-reload. The R15c anchor test was updated to match while keeping that invariant.

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Open a session with a stale unread dot | dot cleared on visit (metadata + post-gap) ✅ |
| Deferred `/api/sessions` poll lands during message-load gap | snapshot synced → open session not re-flagged ✅ |
| Re-select the already-open session with a stale dot | cleared before no-op return ✅ |
| Background completion in a hidden/unfocused tab | still flagged unread (focus gate intact) ✅ |
| Lineage parent with own/other-child unread | keeps its dot (aggregate recompute) ✅ |
| Cross-session load ordering / stale-reject | unchanged (harness stub preserves intent) ✅ |

### Tests

- `tests/test_session_unread_dot_on_visit.py` — **7/7** (structural anchors + node functional: visit clears completion-unread, snapshot sync, same-session clear, focus-gate preserved).
- The previously-hanging `test_cross_session_message_load_isolation` — **passes in ~1s** after the stub fix.
- **Full suite: 12545 passed / 0 failed** in 387s (deselected the known darwin CRLF flake). **CI green (18/18)** after my push.

### Recommendation

The fix is sound — it closes the sticky-unread race by syncing the poll snapshot on visit (metadata + post-message-gap) and clears on same-session re-select, without loosening the focus gate (hidden-tab completions stay unread) and repainting via the authoritative aggregate recompute (parent dots preserved). The CI was red on a lint nit and a pre-existing-test hang the PR introduced by adding a `loadSession` dependency the old harness didn't stub; both fixed in `da00a5ae`, CI now green, full suite clean. Ship.

nesquena-hermes:
Shipped in **exp-v0.52.35** (experimental channel) 🎉

The stale sidebar unread dot now clears when you open a session, while genuinely-hidden-tab completions correctly stay unread. Took a thorough 4-round gate to trace the root (a viewed-count clear in the message-load path, a busy→idle stream mis-record, and a reset-vs-repaint ordering) — all closed. Thanks @neaucode-bot for the original #4946. (closes #5917)

PR #5920, opened 2026-07-11
fix(chat): dedupe pending prompts across compaction

starship-s:
## Thinking Path

- Reloading or reattaching to an active turn can reconstruct the transcript from persisted messages, browser inflight state, and `pending_user_message` at the same time.
- When a synthetic context-compaction row follows the submitted user prompt, current-turn lookup treated that marker as the latest user turn and failed to recognize the prompt already present immediately before it.
- That allowed one submitted prompt to render twice even though backend persistence and model input contained it once.
- The narrow fix is to treat canonical compaction markers as transparent only while locating the active current-turn user message. Completed assistant rows remain hard boundaries, so identical text in separate completed turns is still valid.

## What Changed

- Made both session-load and refresh/reconnect current-tail scans skip canonical synthetic context-compaction rows.
- Kept live assistant and tool rows transparent while preserving completed-turn boundaries.
- Added behavior coverage for pending and inflight reconciliation, refresh-path dedupe, attachment carry-forward, marker/live-row retention, and repeated prompt text across distinct completed turns.

## Why It Matters

- Prevents duplicate user bubbles during active-stream reattachment around a compaction boundary.
- Keeps the visible transcript aligned with the single backend/model turn.
- Avoids global text deduplication, preserving legitimate repeated prompts in later turns.

## Screenshots

Before — the same pending prompt appears twice in a real isolated WebUI active-stream session (`1934×947`).

Before: duplicated pending prompt across a compaction boundary

After — the prompt appears once while the active Hermes worklog remains present (`1934×947`).

After: one pending prompt with active worklog preserved

## Verification

Targeted and related behavior coverage:

`python -m pytest -q tests/test_run_journal_frontend_static.py tests/test_issue2341_pending_user_reattach.py tests/test_inflight_stream_reuse.py tests/test_regressions.py`

Result: `111 passed, 1 skipped`

Independent broader relevant regression sweep: `461 passed`

Syntax and hygiene:

- `node --check static/ui.js`
- `python -m py_compile tests/test_run_journal_frontend_static.py`

Real-instance browser verification used the same neutral 550-message backend session, active stream, browser inflight state, viewport, and navigation sequence before and after.

## Risks / Follow-ups

- This changes frontend transcript reconciliation only; backend persistence and model input are unchanged.
- Compaction recognition continues to use the existing shared classifier used by transcript rendering.

## Model Used

- OpenAI / `gpt-5.6-sol` via Hermes Agent for diagnosis, implementation, verification, and visual validation.
- OpenAI / `gpt-5.3-codex-spark` via OpenCode for review-follow-up implementation and regression coverage.
- Anthropic / `claude-opus-4-8` (`max`) via Claude Code for independent review and final signoff.

### DIFF
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -2907,7 +2907,11 @@ function _currentTailUserMessage(messages){
   for(let i=list.length-1;i>=0;i--){
     const msg=list[i];
     if(!msg) continue;
-    if(String(msg.role||'')==='user') return msg;
+    if(String(msg.role||'')==='user'){
+      // Compaction rows are synthetic user-role markers, not submitted turns.
+      if(typeof _isContextCompactionMessage==='function'&&_isContextCompactionMessage(msg)) continue;
+      return msg;
+    }
     if(msg._live||String(msg.role||'')==='tool') continue;
     return null;
   }
--- a/static/ui.js
+++ b/static/ui.js
@@ -9844,7 +9844,11 @@ function _pendingCurrentTailUserMessage(messages){
   for(let i=list.length-1;i>=0;i--){
     const msg=list[i];
     if(!msg) continue;
-    if(String(msg.role||'')==='user') return msg;
+    if(String(msg.role||'')==='user'){
+      // Compaction rows are synthetic user-role markers, not submitted turns.
+      if(typeof _isContextCompactionMessage==='function'&&_isContextCompactionMessage(msg)) continue;
+      return msg;
+    }
     if(msg._live||String(msg.role||'')==='tool') continue;
     return null;
   }
--- a/tests/test_run_journal_frontend_static.py
+++ b/tests/test_run_journal_frontend_static.py
@@ -80,6 +80,9 @@ def _run_current_turn_scope_probe() -> dict:
     helpers = "\n".join(
         [
             _function_body(UI_SRC, "function _stripWorkspaceDisplayPrefix"),
+            _function_body(UI_SRC, "function msgContent"),
+            _function_body(UI_SRC, "function _isContextCompactionText"),
+            _function_body(UI_SRC, "function _isContextCompactionMessage"),
             _function_body(SESSIONS_SRC, "function _messageComparableText"),
             _function_body(SESSIONS_SRC, "function _stripAttachedFilesMarker"),
             _function_body(SESSIONS_SRC, "function _stripForcedSkillEnvelope"),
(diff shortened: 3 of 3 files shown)


nesquena-hermes:
Reading the diff at `static/sessions.js:2907-2915` and `static/ui.js:9844-9852` on this branch against `origin/master`, plus the shared classifier at `static/ui.js:13228`, this looks correct and tightly scoped. Sharing what I verified.

## What the change does

Both `_currentTailUserMessage` (sessions.js) and `_pendingCurrentTailUserMessage` (ui.js) walk the transcript tail-first looking for the active-turn user message. Previously the very first `role==='user'` row won unconditionally. A synthetic compaction card is stored as a **user-role** row, so when one lands immediately after the submitted prompt, the scan latched onto the marker and never saw that the real prompt was one row earlier — that's the double-render.

The fix makes the marker transparent only inside the user branch:

```js
if(String(msg.role||'')==='user'){
  // Compaction rows are synthetic user-role markers, not submitted turns.
  if(typeof _isContextCompactionMessage==='function'&&_isContextCompactionMessage(msg)) continue;
  return msg;
}
if(msg._live||String(msg.role||'')==='tool') continue;
return null;
```

## Why the scoping is right

The `continue` fires **only** for user-role markers. A completed assistant row still hits the `else` branch (`return null`), so it remains a hard turn boundary — which is what keeps identical prompt text in two separate completed turns from being falsely deduped. The classifier backing this is the same one the transcript renderer uses:

```js
function _isContextCompactionMessage(m){
  if(!m||!m.role||m.role==='tool') return false;
  const text=msgContent(m)||String(m.content||'');
  return _isContextCompactionText(text);
}
```

so recognition can't drift between rendering and this reconciliation path. The `typeof ... === 'function'` guard matches the existing load-order-safe pattern already used at `static/sessions.js:3351` and `static/commands.js:875`, so there's no new coupling risk if ui.js hasn't evaluated yet.

## Test coverage

The additions in `tests/test_run_journal_frontend_static.py` pin the important invariants by behavior, not by implementation:

- `compactionPromptCount == 1` and `compactionMarkerRetained is True` — the prompt renders once **and** the marker card survives (the fix isn't dropping the compaction row, just refusing to treat it as the current turn).
- `distinctCompletedTurnPromptCount == 2` and `repeatedCompletedPromptsRemainValid` (== 3) — legitimate repeated prompt text across distinct completed turns is preserved, confirming no global text dedupe crept in.
- `compactionLiveAssistantRetained is True` — live worklog rows still pass through.

That last one matters because it's the exact regression the "After" screenshot is claiming, and it's now asserted rather than eyeballed.

## One thing worth a sentence in follow-up

The two scans are now identical logic living in two files. They've drifted apart before (this is a salvage-adjacent area). Not a blocker for this PR, but a shared `_isCurrentTurnUserRow(msg)` helper would keep sessions.js and ui.js from diverging on the next compaction-marker shape. No change requested here — the fix is good to merge as-is.

nesquena-hermes:
Shipped in **exp-v0.52.41** via release #5952. I rebased your branch onto current master (it predated the #5946 stream fix) and ran the crown-jewel stream-regression gate — GREEN across all 3 activity modes on the rebased head. The compaction-marker skip guard is in both tail-scan functions. Thanks @starship-s!

PR #5926, opened 2026-07-11
fix(sessions): cascade delegated subagents on parent deletion

pxxD1998:
## Thinking Path

- Delegated sessions are intentionally view-only, but deleting their parent previously removed only the requested row/messages and left delegated descendants orphaned (#5307 / #5332).
- This is a destructive path with no undo, so branch/compression evidence must take precedence and ambiguous rows must be preserved and detached.
- Earlier review rounds closed over-deletion, referential-integrity, retry-manifest, liveness, lock-lifetime, and missing-DB failures.
- Maintainer Round 5 identified the final two product gaps: migrated rows with an explicit `_delegate_from` but no historical `source='subagent'` were under-deleted, and manifest recovery was serialized only within one process.
- The final re-gate also required the committed multi-process regression to pin the complete lock lifetime and the tracked architecture inventory to name the direct destructive state boundary.
- This revision makes explicit delegate lineage authoritative after preservation checks, holds a per-profile cross-process lock across the complete recovery/transaction/artifact lifecycle, exercises both its early and late phases with real processes, and records the compatibility boundary in the architecture contract.

## What Changed

- `api/models.py`
  - Resolves the active profile once and performs parent/delegate cleanup in one `BEGIN IMMEDIATE` transaction.
  - Preserves and detaches real branches (`_branched_from`), compression continuations, ambiguous legacy branch rows, foreign-lineage marker rows, generic children, and every descendant in a preserved physical subtree.
  - Deletes an explicit `_delegate_from` descendant when that authoritative marker links it to the current delete frontier, regardless of its historical `source` value.
  - Uses `source='subagent'` only for marker-less legacy physical-parent inference.
  - Enables foreign keys and explicitly cleans session-owned rows in `compression_locks`, `session_model_usage`, and `telegram_dm_topic_bindings`.
  - Publishes one atomic UUID cleanup manifest before commit, then re-checks DB liveness before removing `.json`, `.jsonl`, or `request_dump_*` artifacts.
  - Fails closed when the DB is missing/unreadable or a liveness query fails.
  - Serializes the full stale-manifest read → DB transaction/commit → post-commit artifact cleanup path with a per-profile in-process thread lock plus a persistent, per-profile cross-process file lock: `fcntl.flock` on POSIX and `msvcrt.locking` on native Windows. If neither primitive exists, deletion reports incomplete cleanup instead of running unlocked.
- `api/routes.py` and `static/sessions.js`
  - Propagate `state_db_cleanup_failed` through the HTTP response.
  - Show an error for both single-delete and batch-delete partial failures instead of an unconditional success toast.
- `docs/architecture/agent-api-contract.md`
  - Records `api/models.py` as a scoped direct-SQL runtime-session-state boundary.
  - Names the canonical lineage-precedence and retryable post-commit cleanup guarantees required before this compatibility path can move behind an Agent-owned endpoint.
- Tests
  - Add destructive classification, transaction, referential cleanup, retry-manifest, liveness, path-safety, lock-lifetime, Windows byte-lock, and real multi-process coverage.
  - The committed two-process regression instruments both stale-manifest recovery and post-commit artifact unlink. It proves both workers reach the late phase and fails if either phase overlaps.

No dependency, schema, authentication, configuration, layout, or styling change is introduced.

## Why It Matters

Deleting a parent conversation should remove positively identified delegated Subagent descendants without deleting a branch or compression continuation. The precedence is:

> Branch/compression/ambiguity preservation first; explicit delegate lineage second; marker-less legacy inference last.

Manifest recovery is now serialized across both threads and worker processes. Concurrent work for the same profile cannot interleave, while different profiles retain independent thread and process locks.

## Contract Routing

Task type: destructive runtime-session-state bug fix.

Touched state and surfaces:
- active profile `state.db`
- `HERMES_HOME/sessions` cleanup manifests and transcript artifacts
- `/api/session/delete`
- single-delete and batch-delete error feedback

Relevant public docs reviewed:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/rfcs/README.md`
- `docs/architecture/agent-api-contract.md`

Contract Change: None. This restores the intended session-deletion invariant and truthful partial-failure reporting rather than redefining the public behavior. The tracked architecture inventory is updated in this PR because `api/models.py` now owns a scoped direct-SQL compatibility path that must eventually move behind an Agent endpoint with equivalent safety guarantees.

State / recovery invariant:
- Requested parent and positively identified delegated descendants are deleted transactionally.
- Branch, compression, foreign-lineage, generic, and ambiguous rows survive; children of the deleted root are detached where required.
- A cleanup manifest is a retry record, not proof that a session is dead. Artifacts are removed only after a successful query proves the ID absent.
- Partial artifact or state cleanup returns `False`, maps to `state_db_cleanup_failed=true`, and produces an error in both UI delete paths.

## Verification

Fixed candidate: `73b6c5d604f48826f729652937d795bbd0e6830d`  
Rebased base: `3f3fa9b69a02968521a81cb5300d6685a9fb0c11`

- Candidate is one commit ahead / zero behind the fetched base; merge-base equals the base.
- Session-delete, agent-source-contract, FD-leak, worktree lifecycle/UI, and writable/read-only state DB gate: **62 passed**.
- Recovery-only lock mutation probe: **expected failure detected** at `post-commit artifact cleanup overlapped across processes`; this proves the committed regression rejects a future lock-scope reduction that releases after stale recovery.
- Global-thread-lock mutation probe: **expected failure detected** at `profile-b was globally blocked by profile-a's in-process lock`; this proves the committed regression rejects a return to the former module-global lock.
- Committed different-profile concurrent-delete regression: **passed**; profile B entered and completed while profile A was intentionally paused inside its own recovery critical section.
- Candidate full repository runner: **13,006 passed, 97 skipped, 1 xfailed, 2 xpassed, 34 subtests passed; 1 failed**.
- Exact latest upstream baseline (`3f3fa9b6`) with the same runner, environment, `.venv`, and test order: **12,984 passed, 96 skipped, 1 xfailed, 2 xpassed, 34 subtests passed; the same 1 test failed**.
  - Shared baseline/candidate failure: `tests/test_issue4685_post_compression_context_metering.py::test_post_compression_estimate_uses_compressor_budget_counter_without_metadata_estimators`.
  - Shared cause: external Hermes Agent `agent.context_compressor` could not import `call_llm` from a prior in-suite `agent.auxiliary_client` stub.
  - The failing test file is byte-identical between candidate and base (`bfd4003108fd9a50a095a1496037fe10c729e474`) and skips when run alone.
  - Candidate adds **22 passing tests and 1 skip, with no candidate-only failure**.
- Browser smoke: `/`, `/#settings`, and `/#sessions` loaded with zero console errors.
- `python3 -m py_compile`, `node --check static/sessions.js`, `git diff --check`, and diff-scoped Ruff: passed; Ruff reports **0 findings on added/modified lines**.
- Public diff sensitive-data scan: no user paths, credentials, tokens, private keys, or email addresses.
- Independent fixed-SHA Spec Review and Quality/Risk Review: **APPROVE**; neither reviewer found a P0/P1 blocker, and both confirmed zero repo/Git/GitHub writes.

### Rounds 1–4

1. Branch/compression evidence takes precedence over inherited delegate metadata; the unsafe canonical fast path was removed.
2. Ambiguous/non-finite/non-comparable branch timestamps fail closed to preservation.
3. Foreign-lineage and preserved-subtree traversal cannot re-enter the delete frontier.
4. Referential tables are cleaned transactionally.
5. Artifact cleanup uses atomic, unique, retryable manifests and transaction-outcome-aware liveness checks.
6. Missing/unreadable/locked DB states preserve manifests and artifacts.
7. The in-process lock is context-managed across every exception and early-return path.
8. Route and both frontend paths report partial cleanup truthfully.

### Round 5

9. Explicit, branch-safe `_delegate_from` descendants are deleted regardless of historical `source`; only marker-less legacy inference requires `source='subagent'`.
10. The full manifest lifecycle is protected by per-profile in-process and persistent interprocess locks; one profile cannot globally block another.
11. The committed two-process regression instruments early recovery and late post-commit unlink; both workers must reach the late phase, and its late window is deliberately longer than recovery so a recovery-only lock mutant is rejected.
12. A separate Windows fallback test verifies byte lock/unlock and exception release.
13. A committed cross-profile concurrency regression and a global-lock mutation control pin the per-profile thread-lock granularity.
14. The tracked architecture contract now identifies the scoped direct-SQL deletion boundary and the guarantees required from its future Agent API replacement.

## Risks / Follow-ups

- Ambiguous legacy rows are deliberately retained rather than aggressively deleted; avoiding irreversible data loss takes precedence over complete cleanup.
- This prevents new delegated orphans during parent deletion; it does not bulk-delete unrelated historical orphans.
- `.session_cleanup.lock` is a persistent per-profile coordination inode. It contains only one zero byte on Windows (and may remain empty on POSIX), carries no user data or secret, and must not be unlinked during normal operation because replacing its inode could split waiters across different locks.
- Native Windows lock behavior is covered by a platform-branch unit test and the repository's Windows CI. The local WSL host did not expose `python.exe`, so native Windows execution is not claimed as local evidence.
- The complete local suite is not reported as green: its single failure is disclosed above and was reproduced on the exact latest base under the same ordered run.
- No screenshots are included because this revision does not alter layout, styling, or normal-success visuals. Existing partial-failure feedback is covered by single/batch static behavior tests plus browser runtime smoke.

## Model Used

- Provider: `openai-codex`
- Exact model: `gpt-5.6-sol`
- Mode / notable tooling: `high` reasoning; Hermes Agent tool execution; destructive disposable SQLite/filesystem probes; true multi-process probes; mutation testing; independent fixed-SHA Spec and Quality/Risk reviews.

## Release Note

Deleting a parent conversation now removes positively identified delegated Subagent descendants, including migrated rows with explicit delegate lineage, while preserving branch, compression, foreign-lineage, generic, and ambiguous continuations. Cleanup manifests fail closed on uncertain DB state and are serialized across threads and worker processes. Partial cleanup is reported through the API and shown as an error in both single and batch deletion flows.

## Related

- Follow-up to #5307 / #5332.
- Addresses maintainer destructive-delete review rounds 1–5, including migrated explicit-marker delegates, complete cross-process manifest lifecycle serialization, and tracked architecture-contract synchronization.


nesquena-hermes (changes requested):
## 🔬 Review: changes requested — data-safety findings on the cascade (please fix before merge)

Thanks @pxxD1998 — routing parent deletion through the canonical `db.delete_session()` + `_delete_delegate_children()` cascade (with a scoped fallback for older Hermes) is the right shape, and the common case works: all 19 PR tests + 22 surrounding session/subagent tests pass. But a data-safety gate found **two paths where MORE than the intended lineage is permanently deleted** — both reproduced with real state-db probes, so they need fixing before this can merge (this is a destructive delete path with no undo).

### Must-fix (data loss — verified)
1. **`api/models.py:8480` + `:8554` — a branch that also carries inherited `_delegate_from` gets destroyed.** The classification accepts `_delegate_from` *before* checking branch/compression metadata, and the canonical helper selected at 8480 has the same precedence. Probe: a session with **both** `_branched_from` and inherited `_delegate_from` had its row, messages, `.json`, `.jsonl`, and request dump all removed when its parent was deleted — but it's a real branch the user would expect to survive. **Fix:** make branch/compression classification take **precedence** over delegate classification in *both* the canonical selection and the fallback traversal — a row that is a branch must be **preserved and detached** (re-parented / orphan-cleared), never cascade-deleted. And require an explicitly branch-safe canonical capability, not just "the helper exists."

2. **`api/models.py:8567` — a legacy `source='subagent'` branch with missing/non-comparable timestamps is misclassified as a delegate and destroyed.** Probe: `end_reason='branched'` + incomplete timestamps → deletion "succeeded" while removing the branch row, messages, and artifacts. The string-comparison of timestamps is the trap. **Fix:** **fail closed** for ambiguous legacy branch metadata — detach/preserve unless the row is *positively* identified as a delegate — and parse timestamps by type instead of comparing strings.

### Must-add (tests)
3. **`tests/test_delete_cli_session_delegate_cascade.py`** — add destructive-path regressions that assert **survival** (row + messages + `.json`/`.jsonl`/request-dump artifacts) for: (a) a both-marker session (`_branched_from` AND `_delegate_from`), and (b) a legacy `source='subagent'` branch with missing/non-comparable timestamps. These are the exact cases the current suite omits.

The precedence rule to encode: **when in doubt, preserve.** A delete cascade should only ever remove a row we can *positively* prove is a delegated child of the deleted parent — anything ambiguous gets detached, not destroyed. Once 1-3 land I'll re-gate the destructive paths. Really good catch on the orphaned-children bug; this is just about making the cleanup provably safe.

_(Held for maintainer review — this is a data-loss-class delete-path change, so it won't be auto-merged even once green.)_

nesquena-hermes (changes requested):
## 🔬 Re-gate (Codex, adversarial) — KICK BACK: still-open data-loss classifier + two cleanup gaps

Thanks @pxxD1998 — the re-push closed a lot: I verified the **exact prior over-deletion case is now safe** (a branch carrying both markers survives with its row + messages + artifacts), branch-preserved subtrees stay attached, an injected SQL abort rolls back every row/message mutation, deleting profile A doesn't touch identical IDs in profile B, and unifying on one scoped transaction (no separate canonical/fallback) is the right call. 19 targeted tests + syntax/Ruff all pass.

But the adversarial gate still found **one unrecoverable-deletion path plus two cleanup gaps** that must be fixed before this can ship on a destructive `state.db` path:

1. **[CORE — data loss] `api/models.py:8606-8623`** — the central classifier still admits rows **without proving BOTH (a) subagent identity (`source='subagent'`) AND (b) compatible lineage**. There remains a shape where a row is accepted for deletion on lineage alone. On a delete path that's unrecoverable. **Fix:** gate admission on `source=='subagent'` AND the ownership/lineage check together — never lineage alone.
2. **[CORE — referential integrity]** SQLite runs without enforced foreign keys here, so deleted parent/delegate rows leave `session_model_usage` and optional `telegram_dm_topic_bindings` references orphaned, and `compression_locks` also remain. **Fix:** `PRAGMA foreign_keys=ON` before `BEGIN IMMEDIATE`, and explicitly clear non-FK session-owned rows (e.g. compression locks) for every removed ID.
3. **[SILENT — idempotency] `api/models.py:8647`** — artifact cleanup is post-commit but not retryable: a child `.json` unlink failure returned `False`, then a second call returned `True` while the child artifact still existed (its DB row / cascade identity was already gone). **Fix:** persist a cleanup manifest/queue of all removed IDs before commit and retry it idempotently until every artifact is gone.

The shape is right and you're close — this is purely tightening the destructive-path safety. Once the classifier requires both gates + the two cleanup fixes land, I'll re-gate. Holding it out of the current release batch until then.

nesquena-hermes (changes requested):
## 🔬 Re-gate (Codex, adversarial, destructive-path scrutiny) — 3 of 5 FIXED, 2 manifest-mechanism edges before ship

Strong progress @pxxD1998 — the scariest parts (over-deleting real conversations) are now verified safe:

1. ✅ **Classifier fixed** — VERIFIED: non-subagent `webui`/`cli` branches carrying `_delegate_from` **survive**, while the genuine `source='subagent'` child is deleted. The lineage-alone over-deletion is closed.
2. ✅ **Referential cleanup fixed** — VERIFIED parent + delegate rows removed from `compression_locks`, `session_model_usage`, and `telegram_dm_topic_bindings`; unrelated rows survive.
4. ✅ **Profile isolation fixed** — VERIFIED an identical session ID + artifacts in another profile are untouched.
5. ◐ **Abort/idempotency partial** — a mid-transaction SQL failure rolls back every table with no orphans (good), but see the commit-failure edge below.

Two findings on the NEW cleanup-manifest mechanism before this destructive path can ship:

1. **[CORE — live-session artifact loss] `api/models.py:8697-8728`** — cleanup entries aren't transaction-outcome-aware. After a **commit failure or process death**, a stale manifest can unlink artifacts for a session ID that **still exists in `sessions`** — i.e. delete a live session's artifacts. **Fix:** never unlink artifacts for a manifest ID while that ID is still present in `sessions` (re-check existence at cleanup time, not just at enqueue time).

2. **[CORE — concurrent-delete manifest clobber] `api/models.py:8705-8719, 8782-8791`** — the per-profile manifest read/merge/cleanup/rewrite isn't serialized or atomic. Reproduced: two concurrent batch deletes — one unlinked the manifest and `beta.json` was left with **no retry record** (its pending cleanup lost). **Fix:** serialize each profile's manifest lifecycle and use atomic replacement (temp-file + rename), plus a deterministic concurrent-delete regression test.

The deletion-safety core is right now — this is the retry/manifest layer needing commit-outcome awareness + concurrency safety. Given it's an unrecoverable `state.db` path, both need closing before merge. Nice work tightening the classifier.

nesquena-hermes (changes requested):
## 🔬 Re-gate round 3 (Codex, destructive-path scrutiny) — SHIP ONLY WITH FIXES (reproduced live-artifact loss on lookup failure)

Thanks @pxxD1998 — real progress, and the previously-verified core stays solid: classifier (real branches + non-subagent rows carrying `_delegate_from` survive, actual subagent children deleted), FK + referential cleanup, profile isolation, and mid-transaction rollback are all **re-verified unchanged** (26 tests pass). Atomic manifest *publication* (UUID temp + same-dir rename) is in. But two CORE issues remain on this unrecoverable `state.db` path:

1. **[CORE — reproduced live-session artifact loss] `api/models.py:8738` (+ `:8871`)** — the transaction-outcome guard fails **open** on a *liveness-query* failure. It protects against an injected commit failure, but when the liveness query itself throws, the code collapses to `alive = set()` (treats everything as not-alive). **Reproduced:** an exclusive SQLite lock (5s query failure) → the session row still existed, but its `.json`, `.jsonl`, request dump, and manifest were all deleted. **Fix:** treat any lookup failure or missing/unreadable `state.db` as **"unknown" → fail closed** — preserve the entire manifest and never unlink unless a *successful* query proves the ID absent.

2. **[CORE — concurrent pending-record loss] `api/models.py:8478`** — atomic publication is not enough without serialization: concurrent batch deletes still lose each other's pending cleanup record. The current concurrency test passes only because it calls the deletes **sequentially** (vacuous on the actual race). **Fix:** serialize each profile's stale-manifest read → DB transaction/commit → post-commit cleanup as one critical section, and add a *real* concurrent-delete regression (parallel, not sequential).

The deletion-safety classifier is genuinely solid now — these last two are about the retry/manifest layer failing **safe** (fail-closed on unknown DB state) and being truly concurrency-safe. Given a wrong unlink here is unrecoverable, both must close before merge. Re-gate once liveness-failure fails closed and the manifest critical section is serialized. Solid convergence.

nesquena-hermes (changes requested):
## 🔬 Re-gate round 4 (Codex, destructive-path) — SHIP ONLY WITH FIXES (lock-leak deadlock + missing-DB still fails open)

Real progress @pxxD1998 — the SQLite-exclusive-lock case now **fails closed** (live artifacts + manifest survive a locked-DB query failure), and the concurrent manifest serialization is verified with **real threads** (a persistently-failing cleanup retains its pending record). Classifier / branch+compression preservation / FK+non-FK cleanup / profile isolation / rollback / atomic publication all re-verified unchanged (44 tests pass). But two CORE edges remain on this unrecoverable path:

1. **[CORE — lock leak → permanent deadlock] `api/models.py:8478`** — the manual `acquire()` / `release()` leaks the lock on an exception path. Reproduced: a manifest unlink raising `PermissionError` escaped **between acquire and release**, and the next `_cleanup_manifest_lock.acquire(blocking=False)` returned `False` — the lock is held forever, deadlocking **all future deletes**. **Fix:** wrap the entire critical section (stale-manifest read + DB transaction + cleanup) in `with _cleanup_manifest_lock:` (or an unconditional outer `try/finally`), and remove the manual releases.

2. **[CORE — missing/unreadable DB still fails open] `api/models.py:8879`** — the exclusive-lock case is fixed, but a **missing or temporarily-inaccessible `state.db`** is still treated as proof that every manifested session is dead. Reproduced: temporarily renaming a DB that contains a live session → its `.json`/`.jsonl` artifacts + manifest were deleted while the live DB row remained. **Fix:** never clean manifested artifacts when `state.db` is missing, inaccessible, or unqueryable — require a **successful** absence `SELECT` before any unlink (the fail-closed rule must cover *all* query-unavailable cases, not just the lock).

You've closed the lock-contention and injected-commit-failure cases — these last two are the exception-safety of the lock itself and extending fail-closed to the missing-DB case. Both are unrecoverable-data risks, so both must land before merge. Use a context manager for the lock and gate every unlink on a proven-successful liveness query. Genuinely close now.

nesquena-hermes (changes requested):
## 🔬 Re-gate round 5 (Codex, destructive-delete adversarial) — SHIP ONLY WITH FIXES (the dangerous directions are all CLOSED; 2 safe-failing gaps remain)

Big progress @pxxD1998 — every **data-loss** risk from the prior rounds is now verified closed against a real state.db:
- ✅ **Lineage over-delete CLOSED** — real sibling branches, their descendants, and unrelated sessions all survive (rows, messages, artifacts, referential data intact).
- ✅ **Fail-CLOSED on liveness-query failure** — real SQLite exclusive-lock + renamed-DB probes preserve the live row/artifacts/manifest and do NOT create a replacement DB.
- ✅ **Lock release** — the context-managed lock is reacquirable after failure; no permanent future-delete deadlock.
- ✅ **FK cascade scope** — removes only parent/delegate-owned usage/lock/telegram-binding rows; siblings + unrelated survive.
- ✅ **Commit ordering** — injected commit failure preserved live rows + artifacts and recovered.

Two remaining gaps — and notably **both fail SAFE** (under-cleanup + a recovery-state race, not deletion of live data), which is why this is SHIP-ONLY-WITH-FIXES not a hard kickback:

1. **[under-delete — original bug not fully fixed for migrated sessions] `api/models.py:8729`** — the delete requires `source == "subagent"`, so Agent-**migrated legacy delegates** (which carry an explicit `_delegate_from` but not the `subagent` source tag) are **under-deleted** → they stay orphaned, i.e. the #5307/#5332 bug this PR targets is still live for migrated sessions. **Fix:** delete branch-safe explicit `_delegate_from` descendants **regardless of historical source**; keep the `source == "subagent"` requirement ONLY for unmarked legacy-parent *inference* (where there's no explicit `_delegate_from`).

2. **[cross-process manifest race] `api/models.py:8566`** — stale-manifest processing isn't serialized **across processes**, so two processes recovering manifests concurrently can clobber recovery state (in-process is now locked, but not interprocess). **Fix:** take the SQLite writer lock before processing manifests, or add an equivalent per-profile interprocess lock.

Neither verified path deletes a live/unrelated session or deadlocks future deletes — the destructive safety is solid now. These are: finish the job for migrated delegates (#1) + make manifest recovery interprocess-safe (#2). Add tests for both (a migrated `_delegate_from`-only delegate gets cleaned; concurrent-process manifest recovery). Re-gate when they land. Genuinely strong work hardening a permanent-data-loss path across the rounds.

nesquena-hermes:
Shipped in **exp-v0.52.72** 🚀 — thanks @pxxD1998, and real credit for the careful multi-round hardening on a data-deletion path. Final Codex gate SAFE: deleting a parent session now cascades its delegated/subagent sessions + artifacts (incl Agent-migrated legacy delegates) with **no** over-delete of siblings/unrelated/live sessions, the new cross-process file lock verified on a real two-process test (serialized, no deadlock, released, lock file never unlinked), fail-closed on missing/unqueryable state.db + liveness-query failure + absent lock primitive (20-call no-primitive probe preserved everything), recovery-from-prior-failed-run works, and partial cleanup is reported honestly. 51 tests, full CI green. Relates #5307.

Issue #5871, opened 2026-07-10
Agent misinterprets redacted credential values (***, sk-abc...xyz1) as user input errors — system prompt has no masking awareness


buaaflyaway:
# Agent misinterprets redacted credential values (`***`, `sk-abc...xyz1`) as user input errors — system prompt has no masking awareness

## Summary

Hermes WebUI and Hermes Agent implement a **comprehensive server-side secret redaction system** that masks API keys, tokens, passwords, and other credentials from all API responses and agent-visible data. The redaction itself works correctly. **However, the LLM agent is never told that this masking exists**, so when it encounters redacted values like `***` or `sk-abc...xyz1` in configuration data, it consistently misinterprets them as placeholder text or user input errors — leading to incorrect diagnostic messages telling users to "fix" their credentials when the stored values are actually correct.

This is a **system-prompt gap**, not a redaction bug.

---

#### 1a. WebUI server-side: `api/helpers.py:273–432`

`_build_redact_fn()` creates a two-layer combined redactor:

```python
def _build_redact_fn():
    ...
    try:
        from agent.redact import redact_sensitive_text
    except ImportError:
        return _fallback_redact

    def _combined_redact(text: str) -> str:
        # WebUI API responses are a hard safety boundary — pass force=True
        agent_redacted = redact_sensitive_text(text, force=True)
        agent_redacted = _restore_code_env_key_literals(text, agent_redacted)
        return _fallback_redact(agent_redacted)

    return _combined_redact
```

The mask function (`_mask`, line 333–334):

```python
def _mask(token: str) -> str:
    return f"{token[:6]}...{token[-4:]}" if len(token) >= 18 else "***"
```

- Short/unknown tokens (< 18 chars) → `"***"`
- Known-prefix long tokens → `"sk-abc...xyz1"` (first 6 + `...` + last 4)

[...]

PR #5927, opened 2026-07-11
fix: teach the WebUI ephemeral system prompt that redacted credential values are intentional masking

mvanhorn:
## Thinking Path
Users were being told their stored credentials were wrong when the values were actually fine. The WebUI and the agent redact secrets before the model sees them (`***`, `sk-abc...xyz1`), but nothing in the model-facing prompt says that masking exists, so the model reads a masked value as a placeholder or a user mistake. The collaborator review on #5871 confirmed the failure end to end and endorsed Option A, a masking-awareness bullet in `_WEBUI_PROGRESS_PROMPT`, as the WebUI-side stopgap while the durable Option B fix belongs in the upstream agent repo.

## What Changed
One bullet added to `_WEBUI_PROGRESS_PROMPT` in `api/streaming.py`, placed next to the existing "Do not reveal ... secrets" line so leak-prevention and interpretation guidance sit together. It tells the model that password/API-key/token/secret fields are automatically redacted, that masked values are intentional redaction rather than placeholder text or input errors, and that it must not tell a user a stored credential is wrong based on a masked value alone. Per the maintainer's wording note, it does not enumerate what real key shapes look like.

## Why It Matters
This closes a confirmed real-user failure (`VITE_DEFAULT_API_KEY=***` being flagged as a broken credential) at the layer the maintainer scoped for this repo. The prompt flows into the model through `agent.ephemeral_system_prompt = _webui_ephemeral_system_prompt(...)`, so every WebUI session picks it up.

## Verification
`tests/test_issue5871_redaction_awareness_prompt.py` follows the repo's `tests/test_issue<NNNN>_*.py` convention: it asserts the guidance bullet is present in `_WEBUI_PROGRESS_PROMPT` and that it flows into the assembled ephemeral prompt. Ran with pytest; both tests pass.

## Risks
Low. The change is a single prompt bullet; no code paths change. The main risk is prompt-length creep, which one line does not meaningfully affect.

## Contract Routing
Touches the WebUI ephemeral-prompt contract family (model-facing progress guidance in `api/streaming.py`); no changes to `docs/CONTRACTS.md` schemas themselves.

## Model Used
AI was used for assistance.

Refs #5871 (covers the WebUI-side Option A; the durable Option B fix needs an upstream change in the agent repo)

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -636,6 +636,7 @@ WebUI progress guidance:
 - Each update should say what you are about to check, what you just confirmed, or why the next tool call is needed.
 - Keep updates concise, factual, and in the user's language. One or two short sentences are enough.
 - Do not reveal hidden reasoning, chain-of-thought, private scratchpads, secrets, raw logs, or long tool output.
+- Password, API-key, token, and secret fields are automatically redacted by the system. Treat masked values as intentional redaction, not placeholder text or user input errors, and do not tell the user a stored credential is wrong based on a masked value alone.
 - Final visible assistant replies must be clear, user-facing, and in the user's language, not private planning notes.
 - Do not include terse planning fragments or scratchpad shorthand in visible assistant text. Avoid fragments like "Need script", "Need check logs", "Need inspect email", or "maybe invite"; either omit them or rewrite them as clear user-facing progress.
 - For direct answers or very short tasks, skip progress updates and answer normally.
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.36** (experimental channel) 🎉

The WebUI progress-guidance prompt now tells the agent that redacted credential fields (`***`, `sk-abc…xyz1`) are intentional masking, so it stops telling users a correctly-stored secret is "wrong" or a placeholder. Clean one-line additive fix — gate-confirmed no secret exposure, full suite green. Thanks @mvanhorn! (closes #5927, fixes #5871)

PR #5938, opened 2026-07-11
CSP: opt-in frame-ancestors allowlist via HERMES_WEBUI_CSP_FRAME_ANCESTORS

Kahotik:
## What

Adds an opt-in `HERMES_WEBUI_CSP_FRAME_ANCESTORS` env var that lets an operator allow specific http(s) origins to embed the WebUI in an `<iframe>`. It is the natural counterpart of the existing `HERMES_WEBUI_CSP_FRAME_EXTRA` knob: that one governs what the WebUI page may embed; this one governs who may embed the WebUI.

## Why

Self-hosters increasingly pin the WebUI as a tab inside a local dashboard (mission-control style apps on `localhost`). Today that's impossible: `frame-ancestors 'none'` plus `X-Frame-Options: DENY` are hardcoded, so the embed is refused even between two apps on the same machine. An opt-in allowlist keeps the secure default while unblocking that use case.

## How

- `_csp_frame_ancestors()` in `api/helpers.py`: returns `'none'` unless the env var is set; values are space-separated http(s) origins validated with the same `_CSP_EXTRA_FRAME_RE` shape used by the frame-src knob (wildcard subdomain + optional port; directive injection, paths, ws/wss and bad ports are rejected with a warning, falling back to `'none'`).
- The `X-Frame-Options: DENY` header is only sent when frame-ancestors resolves to `'none'` — otherwise it would override the CSP allowlist in every browser.
- Default behavior is unchanged: with the var unset, policy and headers are byte-identical to before.

## Tests

`tests/test_csp_frame_ancestors.py` mirrors the frame-src suite: default lockdown, valid origins, enforced policy, directive injection, paths, ws scheme, invalid ports, independence from `frame-src`, and X-Frame-Options presence/omission. All 8 pass, plus the 8 existing `test_csp_frame_src_extra.py` tests still pass.

Documented in `.env.example`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -67,6 +67,11 @@
 # HERMES_WEBUI_CSP_CONNECT_EXTRA=https://api.example.com wss://api.example.com
 # HERMES_WEBUI_CSP_FRAME_EXTRA=https://embed.example.com
 
+# Who may embed the WebUI in an <iframe> (CSP frame-ancestors). Default: nobody
+# ('none' + X-Frame-Options DENY). Space-separated http(s) origins; when set,
+# X-Frame-Options is omitted so the CSP value governs embedding.
+# HERMES_WEBUI_CSP_FRAME_ANCESTORS=http://localhost:3000
+
 # ── Reverse-proxy header trust (opt-in; only enable behind a proxy YOU run) ───
 # By default these forwarded headers are NOT trusted, because any direct client
 # could forge them. Enable (1/true/yes/on) only when a reverse proxy you control
--- a/.env.example
+++ b/.env.example
@@ -69,7 +69,10 @@
 
 # Who may embed the WebUI in an <iframe> (CSP frame-ancestors). Default: nobody
 # ('none' + X-Frame-Options DENY). Space-separated http(s) origins; when set,
-# X-Frame-Options is omitted so the CSP value governs embedding.
+# X-Frame-Options is omitted so the CSP value governs embedding. Each origin
+# must be a hostname or IPv4 literal, optionally with a *. prefix and a port;
+# bracketed IPv6 origins (http://[::1]:3000) are NOT supported and make the
+# whole value fall back to 'none', as does anything else malformed.
 # HERMES_WEBUI_CSP_FRAME_ANCESTORS=http://localhost:3000
 
 # ── Reverse-proxy header trust (opt-in; only enable behind a proxy YOU run) ───
--- a/.env.example
+++ b/.env.example
@@ -70,9 +70,9 @@
 # Who may embed the WebUI in an <iframe> (CSP frame-ancestors). Default: nobody
 # ('none' + X-Frame-Options DENY). Space-separated http(s) origins; when set,
 # X-Frame-Options is omitted so the CSP value governs embedding. Each origin
-# must be a hostname or IPv4 literal, optionally with a *. prefix and a port;
-# bracketed IPv6 origins (http://[::1]:3000) are NOT supported and make the
-# whole value fall back to 'none', as does anything else malformed.
+# must be a hostname or IPv4 literal (optionally with a *. prefix and a port),
+# or a bracketed IPv6 literal such as http://[::1]:3000. Anything malformed
+# makes the whole value fall back to 'none'.
 # HERMES_WEBUI_CSP_FRAME_ANCESTORS=http://localhost:3000
 
 # ── Reverse-proxy header trust (opt-in; only enable behind a proxy YOU run) ───
--- a/api/helpers.py
+++ b/api/helpers.py
@@ -81,6 +81,19 @@ _CSP_EXTRA_CONNECT_RE = _re.compile(
 _CSP_EXTRA_FRAME_RE = _re.compile(
     r"^https?://(?:\*\.)?[A-Za-z0-9._~-]+(?::(?P<port>\d{1,5}|\*))?$"
 )
+# Validator for a frame-ancestors allowlist entry (HERMES_WEBUI_CSP_FRAME_ANCESTORS).
+# Stricter than the frame-src one on purpose: this directive decides who may embed
+# the WebUI, and an accepted-but-malformed source silently drops X-Frame-Options,
+# leaving the embed browser-dependent. Host labels follow CSP/DNS structure —
+# ASCII alphanumerics with internal hyphens, single dots between labels, no empty
+# label — and the port is ASCII [0-9] only, because ``\d`` also matches Unicode
+# digits (e.g. Arabic-Indic), which no browser parses as a port.
+_CSP_ANCESTOR_HOST = r"[A-Za-z0-9](?:[A-Za-z0-9-]*[A-Za-z0-9])?"
+_CSP_FRAME_ANCESTOR_RE = _re.compile(
+    r"^https?://(?:\*\.)?"
+    rf"{_CSP_ANCESTOR_HOST}(?:\.{_CSP_ANCESTOR_HOST})*"
+    r"(?::(?P<port>[0-9]{1,5}|\*))?$"
+)
 _CSP_HEADER_NAME = 'Content-Security-Policy'
 _CSP_SHARED_POLICY_TEMPLATE = (
     "default-src 'self' https://*.cloudflareaccess.com; "
@@ -143,6 +156,16 @@ def _valid_csp_extra_frame_source(source: str) -> bool:
         return False
 
 
+def _valid_csp_frame_ancestor_source(source: str) -> bool:
+    match = _CSP_FRAME_ANCESTOR_RE.fullmatch(source)
+    if not match:
+        return False
+    port = match.group("port")
+    if not port or port == "*":
+        return True
+    return 1 <= int(port) <= 65535
+
+
 def _csp_extra_frame_src() -> str:
     raw = os.getenv("HERMES_WEBUI_CSP_FRAME_EXTRA", "").strip()
     if not raw:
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Reading `api/helpers.py:64-89` (the shared policy template + regex), `144-162` (`_csp_frame_ancestors`), and `179-207` (`_build_csp_enforced_policy` + `_security_headers`) against `origin/master`, plus `tests/test_csp_frame_ancestors.py` — the design is sound and the default-off behavior is preserved. One real edge case to flag before merge, matching Greptile's note.

## Design looks right

Reusing `_valid_csp_extra_frame_source` / `_CSP_EXTRA_FRAME_RE` for the ancestors allowlist keeps directive-injection, path, ws-scheme and bad-port rejection identical to the frame-src knob, and falling back to `'none'` on any invalid token is the safe default:

```python
# api/helpers.py:154-160
sources = raw.split()
if any(not _valid_csp_extra_frame_source(src) for src in sources):
    logger.warning("Ignoring invalid HERMES_WEBUI_CSP_FRAME_ANCESTORS value")
    return "'none'"
return " ".join(sources)
```

The conditional `X-Frame-Options` header is the correct call — sending `DENY` alongside a non-`'none'` `frame-ancestors` would let the legacy header override the CSP allowlist in most browsers:

```python
# api/helpers.py:203-204
if _csp_frame_ancestors() == "'none'":
    handler.send_header('X-Frame-Options', 'DENY')
```

With the var unset, `_csp_frame_ancestors()` returns `'none'`, the template renders byte-identical to before, and `X-Frame-Options: DENY` still ships. The default-lockdown test confirms that.

## Edge case: bracketed IPv6 origins fall back to lockdown

`_CSP_EXTRA_FRAME_RE` is:

```python
# api/helpers.py:68-70
_CSP_EXTRA_FRAME_RE = _re.compile(
    r"^https?://(?:\*\.)?[A-Za-z0-9._~-]+(?::(?P<port>\d{1,5}|\*))?$"
)
```

The host class `[A-Za-z0-9._~-]+` excludes `[`, `]`, and `:`, so a valid literal IPv6 origin like `http://[::1]:3000` fails `fullmatch`, and `_csp_frame_ancestors()` silently falls back to `'none'`. Given the stated use case is "two apps on the same machine / localhost dashboards," an operator on an IPv6-only loopback (`[::1]`) would find the allowlist silently ignored with only a generic warning. That's a fail-closed miss, not a security hole, but it's surprising.

This is a pre-existing limitation inherited from the frame-src validator — it's not introduced by this PR — so it's reasonable to punt. But since the whole motivation here is localhost embedding, I'd either (a) extend the regex to accept a bracketed IPv6 literal, e.g. add an alternation `\[[0-9A-Fa-f:]+\]` for the host, or (b) add a one-line note in `.env.example` that only hostname/IPv4 origins are supported and `[::1]` should be written as `127.0.0.1` on the embedding side. Option (b) is the minimal fix if you want to keep the regex shared and unchanged.

## Test suggestion

The suite covers defaults, valid origins, injection, paths, ws, bad ports, frame-src independence, and header presence/omission — thorough. If you go with (a), add a `test_csp_frame_ancestors_accepts_ipv6_loopback` case asserting `http://[::1]:3000` survives. If you go with (b), a negative test documenting that `[::1]` falls back to `'none'` makes the intended limitation explicit.

Nice symmetric addition otherwise — the `.env.example` docs and the "this governs who may embed us, the counterpart of FRAME_EXTRA" framing are clear.

nesquena-hermes:
## 🟢 Authoritative gate certificate — GREEN (security ratification still required)

**Exact contributor head:** `26a7ad8a5e3ff867a887bab8f7c608f6559b7ac8`

I gated this head after a clean transplant onto current `origin/master` (`deac1384f`) on `gate-rebase/5938-csp` at `36b43b02c46edad1a16052ae6b02504a06cff27d`.

### Evidence
- Threat scan: **CLEAN**.
- Focused CSP/security suite: **47 passed**.
- Full sandboxed suite: **13,364 passed / 14 failed**; all 14 match the clean-master sandbox baseline, with no PR-specific failure.
- Codex: **SAFE TO SHIP**, no regression risk found.
- Backend advisor: **passes the backend gate**.
- I verified real HTTP response headers in isolated servers:
  - valid origins produced the exact `frame-ancestors` allowlist and omitted XFO;
  - malformed directive-injection input failed closed to `frame-ancestors 'none'` plus `X-Frame-Options: DENY`;
  - `frame-src`, `connect-src`, auth, CSRF, and other directives were unchanged.

The default posture is byte-behavior compatible and remains locked down. Valid configured origins are strictly parsed as http(s) origin sources; partial-list failures, paths, ws/wss, directive separators, malformed ports, and hostile values fail closed. Enforced and report-only policies share the same builder.

### Non-blocking notes
- Invalid configuration is recomputed/warned multiple times per request; compute and cache once on the handler as future polish.
- The knob is intentionally global, so a valid opt-in also permits the allowed origins to frame the login page. That is a product/security policy consequence, not a code defect.

This exact head is technically **GREEN**. Apply `gate-pass`; keep it behind maintainer/security review. Nathan’s final security-boundary/curation ratification remains mandatory before merge.

nesquena-hermes:
Thanks for this, @Kahotik — the concept is solid and in-scope (opt-in operator embedding is the natural counterpart to the existing `frame-src` knob, and the default-locked-down design is right). I rebased it onto current master and ran it through an adversarial security gate. It reproduced three issues that need addressing before it can ship — all have precise fixes:

### 1. Profile `.env` can override the deployment-wide framing policy (privilege escalation) — must-fix
A per-profile `.env` can set `HERMES_WEBUI_CSP_FRAME_ANCESTORS`, and it gets written into process-global `os.environ` for the duration of a send (`api/streaming.py:8350,8469`, via `api/profiles.py:895-911,940-952`). So one profile can replace the framing policy **for every user** — a probe produced `frame-ancestors https://attacker.example` with no `X-Frame-Options`. This crosses the profile-isolation boundary.

**Fix:** add `HERMES_WEBUI_CSP_FRAME_ANCESTORS` to **both** `_PROTECTED_ENV_KEYS` (`api/profiles.py:201`) and `_BLOCKED_RUNTIME_ENV_KEYS` (`api/profiles.py:922`), so only startup/deployment configuration can control it (never a profile `.env`). Add a test asserting a profile `.env` value is ignored.

### 2. `X-Frame-Options` and CSP resolve the value independently (contradictory headers) — must-fix
The ancestors value is resolved three times from mutable state: `_security_headers()` (`api/helpers.py:205`), `_build_csp_enforced_policy()` (`api/helpers.py:184`), and report-only CSP (`server.py:332`). Reproductions emitted both `XFO: DENY` **with** a permissive ancestor and `'none'` **without** XFO. Conforming browsers ignore XFO whenever an enforced `frame-ancestors` exists, so the two must be derived from one resolution.

**Fix:** resolve `frame-ancestors` **once per response**, cache it on the handler alongside the existing CSP extras, and pass that same value through the enforced builder, the report-only builder, and the XFO decision.

### 3. Origin validator accepts malformed sources — should-fix
The validator accepts `https://.`, `https://_`, `https://~`, `https://example..com`, and Unicode-digit ports without falling back to `'none'` (XFO is then silently omitted and the embed becomes browser-dependent). CSP host labels permit only ASCII alphanumeric/hyphen, and ports use ASCII `[0-9]`.

**Fix:** use an ancestor-specific validator with proper host-label structure and `[0-9]` ports (mirroring the sibling `frame-src` behavior), and add rejection + XFO-omission tests for each of those malformed cases.

Once these land I'll re-run the same security gate on your re-push and take it the rest of the way. Appreciate the contribution — this is a genuinely useful self-hosting capability, just needs the boundary tightened.

Kahotik:
Rebased onto current master (`e168b67e`) and pushed. The three security-gate findings are now all addressed.

**1 and 2 (must-fix)** landed in `3aec00c0`, as reviewed on 2026-08-18: `HERMES_WEBUI_CSP_FRAME_ANCESTORS` is in both `_PROTECTED_ENV_KEYS` and `_BLOCKED_RUNTIME_ENV_KEYS` so a profile `.env` can no longer touch the deployment-wide policy, and the value is resolved once per response and cached on the handler, then reused by the enforced builder, the report-only builder and the `X-Frame-Options` decision.

**3 (should-fix)** is also in that commit: an ancestor-specific validator with proper host-label structure and ASCII `[0-9]` ports, with rejection + XFO-omission tests for `https://.`, `https://_`, `https://~`, `https://example..com` and Unicode-digit ports.

**IPv6** — the point you and Greptile both flagged. I went with option (a) rather than documenting the limitation, since the stated use case is localhost embedding and an IPv6-only loopback has no other way to name the origin. New commit `bfaa934b`:

- The host pattern gets a second alternative, a bracketed literal `\[[0-9A-Fa-f:.]+\]`, with no `*.` prefix allowed on it — a wildcard label has no meaning for a literal.
- The brackets only get it past the shape check. `_valid_csp_frame_ancestor_source` then parses the inside with `ipaddress.IPv6Address`, so a bracketed blob of hex and colons is still rejected: `[:::1]`, `[12345::1]`, `[192.168.1.1]` and `[]` all fall back to `'none'` with `X-Frame-Options: DENY` restored. Port range checking is unchanged, so `http://[::1]:99999` is rejected too.
- Tests both ways: `test_csp_frame_ancestors_accepts_ipv6_loopback` covers `http://[::1]:3000`, `http://[::1]` and `https://[2001:db8::1]:8443` and asserts the header carries the origin with XFO omitted; `test_csp_frame_ancestors_rejects_malformed_ipv6` covers the five rejected forms and asserts the `'none'` fallback with XFO back to `DENY`.
- `.env.example` no longer documents the IPv6 limitation.

The frame-src validator (`_CSP_EXTRA_FRAME_RE`) is untouched — the IPv6 change is confined to the ancestors validator.

175 tests pass across the CSP and security files on the new base.

Kahotik:
Good catch on the wildcard port — fixed in `93bed73f`.

Accepting bracketed IPv6 also made `http://[::1]:*` valid, which omits `X-Frame-Options` for a source browsers won't honour: broken embed, protection already dropped. That's the accepted-but-unusable case this validator exists to avoid, reintroduced from a new angle.

`*` now returns `ipv6 is None`, so wildcard ports stay valid on hostname sources only. Test asserts `http://localhost:*` still passes, `http://[::1]:*` falls back to `'none'` with `X-Frame-Options: DENY`.

177 tests pass across the CSP and security files.

Issue #1426, opened 2026-05-01
bug+feat(models): OpenRouter free-tier variants invisible (tool-support filter) + Custom Model ID input has no provider selector — silent failure


nesquena-hermes:
## Summary

Two related gaps in the model picker that block users from selecting OpenRouter's free-tier models (and other variants that don't appear in the default catalog):

1. **OpenRouter `:free` and other tool-supportless variants are silently filtered out.** `hermes_cli/models.py::fetch_openrouter_models()` (line 940) intersects three sources — curated remote manifest, live `/v1/models`, and a tool-support filter — and the user has no UI hint that filtering happened or how to override it.
2. **The "Custom model ID" input has no provider field.** Typing a custom ID like `minimax/minimax-m2.5:free` routes via namespace inference, which sends it to **MiniMax direct API** (the `minimax/` namespace owns its own provider) instead of the user's intended **OpenRouter** routing. The chat then fails silently because MiniMax direct doesn't recognize the `:free` variant or has no API key configured.

This is a sub-issue of #1240 (Provider/Model source-of-truth umbrella).

## Reported by

@AvidFuturist (Discord, May 1 2026):
- "some models do not show up, (e.g. openrouter free tier like `minimax/minimax-m2.5:free` unless it is intentional). I can provide a custom model ID, but that does not let me pick provider thus it fails silently"

### Why `:free` variants are invisible

`hermes_cli/models.py:940-1005` (`fetch_openrouter_models`):

```python
# 1. Curated manifest fetch
remote = get_curated_openrouter_models()  # hosted catalog
fallback = list(remote) if remote else list(OPENROUTER_MODELS)
preferred_ids = [mid for mid, _ in fallback]

# 2. Live /v1/models fetch
with urllib.request.urlopen("https://openrouter.ai/api/v1/models", ...) as resp:
    payload = json.loads(resp.read().decode())
live_by_id = {item["id"]: item for item in payload["data"]}

# 3. Intersection: only IDs in BOTH curated manifest AND live response
for preferred_id in preferred_ids:
    live_item = live_by_id.get(preferred_id)
    if live_item is None:
        continue
    # 4. Tool-support filter (Kilo-Org/kilocode#9068 port)
    if not _openrouter_model_supports_tools(live_item):
        continue
    ...
```

So a model is shown only if it's:
1. In the curated manifest (`OPENROUTER_MODELS` snapshot or hosted catalog), AND
2. Returned by live `/v1/models`, AND
[...]

PR #5947, opened 2026-07-11
test(#1426): make OpenRouter live-fetch test hermetic (fix deterministic failure)

nesquena-hermes:
## Summary

Fixes a **deterministically-failing** test (not a network flake, not xdist pollution):

```
pytest "tests/test_issue1426_openrouter_free_tier_live_fetch.py::test_openrouter_group_uses_live_fetch_when_available" -p no:xdist
```

fails with:

```
assert 'minimax/minimax-m2.5:free' in ['anthropic/claude-sonnet-4.6', 'anthropic/claude-fable-5', 'anthropic/claude-opus-4.8', ...]
```

The failure reproduces in complete isolation in any environment where the `hermes_cli` agent package is **installed** and the network is **reachable** (e.g. local dev with the agent present). It passes in CI only because `hermes_cli` is absent there.

## Root cause — a gap in the test's mock, **not** a production regression

`api.config.get_available_models()` populates the OpenRouter picker group from **two** live fetches:

1. **Curated catalog** — `hermes_cli.models.fetch_openrouter_models()`. This routes through hermes_cli's `open_credentialed_url()`, which builds its **own** opener and calls `.open()`. It does **not** go through the module-level `urllib.request.urlopen` symbol the test monkeypatches.
2. **Free-tier augment** — a direct `urllib.request.urlopen("https://openrouter.ai/api/v1/models")` at `api/config.py:7110`, filtered to free-tier-only. This one **is** intercepted by the monkeypatch.

The test only patched `urllib.request.urlopen`, so fetch (1) hit the **real network** and returned ~36 live curated models. Those filled the visible picker cap (15), pushing the mocked free-tier entries (`minimax/minimax-m2.5:free`, `openrouter/elephant-alpha`) into the group's `extra_models` overflow bucket. The assertion only inspected the visible `models` bucket → failure.

In CI, `hermes_cli` is not installed, so fetch (1) raises `ImportError`, the curated base is empty, the 2–3 mocked entries fit the visible cap, and the test passes. Hence the environment-dependent behavior.

### Production is correct — proven

Free-tier models **do** surface in the real picker. Verified against production code with the test payload:

```
visible models count: 15
extra_models count: 23
minimax/minimax-m2.5:free  in VISIBLE? False   in EXTRA? True
openrouter/elephant-alpha  in VISIBLE? False   in EXTRA? True
```

Both `models` (visible dropdown) and `extra_models` (slash-command autocomplete / dynamic-label overflow) are legitimate picker surfaces. A free-tier entry landing in either has surfaced in the picker — it was only the assertion's narrow bucket choice that made this look like a regression.

## Fix (test-only, coverage intent preserved)

- **Mock fetch (1) too** — as the sibling `dedupe` / `fail-closed` tests already do — so the test is genuinely hermetic, matching this module's own docstring promise of *"without depending on real network access."* A small curated base (`anthropic/claude-sonnet-4.6`) keeps "live data, not just the fallback list" exercised.
- **Assert across both picker buckets** (`models` + `extra_models`), matching the sibling cap and fail-closed tests, so a large curated base overflowing the visible cap no longer flips the assertion.

No `pytest.skip`, no deleted assertion — the free-tier-visibility contract from #1426 is still fully asserted.

**No production code changed.** This PR touches only the test file and is independent of the separate transparent-stream fix.

## Verification

```
# target test in isolation
pytest "tests/...::test_openrouter_group_uses_live_fetch_when_available" -p no:xdist   → 1 passed

# whole file, no xdist
pytest tests/test_issue1426_openrouter_free_tier_live_fetch.py -p no:xdist            → 10 passed

# whole file, xdist (CI-like)
pytest tests/test_issue1426_openrouter_free_tier_live_fetch.py                        → 10 passed

# adjacent model-catalog suites (no regression)
pytest tests/test_live_models_ttl_cache.py \
       tests/test_issue1807_codex_provider_card_live_models.py \
       tests/test_issue1240_generic_cli_catalog_sync.py -p no:xdist                   → 10 passed
```

### DIFF
--- a/tests/test_issue1426_openrouter_free_tier_live_fetch.py
+++ b/tests/test_issue1426_openrouter_free_tier_live_fetch.py
@@ -111,9 +111,28 @@ def test_openrouter_group_uses_live_fetch_when_available(monkeypatch):
         return _FakeResponse(fake_payload)
 
     monkeypatch.setattr(urllib.request, "urlopen", _fake_urlopen)
+    # The OpenRouter branch does TWO live fetches:
+    #   (1) hermes_cli.models.fetch_openrouter_models() — the curated catalog.
+    #       This routes through hermes_cli's `open_credentialed_url()`, which
+    #       builds its own opener and calls `.open()`; it does NOT go through
+    #       the module-level `urllib.request.urlopen` symbol patched above.
+    #       So the monkeypatch alone leaves fetch (1) hitting the REAL network
+    #       whenever hermes_cli is installed — returning the live curated list,
+    #       which fills the visible picker cap and pushes the mocked free-tier
+    #       entries into `extra_models`. Mock fetch (1) here too so the test is
+    #       genuinely hermetic (as the module docstring promises) and the
+    #       free-tier augment path (2) is what is actually under test.
+    #   (2) the direct urllib.request.urlopen(".../v1/models") — the free-tier
+    #       augment; this one IS intercepted by the monkeypatch above.
     try:
         from hermes_cli import models as _hm
         monkeypatch.setattr(_hm, "_openrouter_catalog_cache", None, raising=False)
+        # A small curated base ("live data, not just the fallback list") so the
+        # tool-supporting paid model is present alongside the free-tier augment.
+        monkeypatch.setattr(
+            _hm, "fetch_openrouter_models",
+            lambda *a, **k: [("anthropic/claude-sonnet-4.6", "")],
+        )
     except Exception:
         pass
 
@@ -121,7 +140,16 @@ def test_openrouter_group_uses_live_fetch_when_available(monkeypatch):
     or_group = next((g for g in grouped if g.get("provider_id") == "openrouter"), None)
     assert or_group is not None, "openrouter group must be present"
 
-    model_ids = [m["id"] for m in or_group["models"]]
+    # Both `models` (visible dropdown) and `extra_models` (slash-command
+    # autocomplete / dynamic-label overflow) are picker surfaces — a free-tier
+    # entry that lands in either bucket has surfaced in the picker. Assert
+    # across both, matching the sibling fail-closed / cap tests, so a large
+    # curated base overflowing the visible cap does not flip this assertion.
+    model_ids = [
+        m["id"]
+        for bucket in ("models", "extra_models")
+        for m in or_group.get(bucket, [])
+    ]
     # Resilient to test-isolation pollution: when a sibling test mutates
     # `cfg` and triggers the openrouter-not-active branch, _apply_provider_prefix
     # adds an `@openrouter:` prefix to model IDs. Skip rather than fail — the


nesquena (approved):
## Review — ✅ approved (test-only hermeticity fix)

Independent review of #1426's test-hermeticity fix: `test_openrouter_group_uses_live_fetch_when_available` fails **deterministically** in any env where `hermes_cli` is installed and the network is reachable (agent-present local dev), passing in CI only because the agent is absent there.

### What this ships

`tests/test_issue1426_openrouter_free_tier_live_fetch.py` only (+29/-1). **No production code.** Agent-authored (`nesquena-hermes`). MERGEABLE, CI green.

### Root cause — verified against production

The OpenRouter picker group is built from two live fetches (`api/config.py:7085-7110`): (1) `hermes_cli.models.fetch_openrouter_models()` — the curated catalog, via hermes_cli's **own** opener (`open_credentialed_url`), which does **not** go through the module-level `urllib.request.urlopen` the test patched; and (2) a direct `urllib.request.urlopen(".../v1/models")` free-tier augment, which **is** intercepted. So the old test left fetch (1) hitting the real network when the agent was installed → ~36 live curated models filled the visible cap (15) → the mocked free-tier entries overflowed into `extra_models` → the `models`-only assertion failed. I confirmed both `models` and `extra_models` are legitimate picker surfaces (visible dropdown + slash-autocomplete/overflow), so a free-tier entry in either has surfaced — the failure was the assertion's narrow bucket, not a production regression. ✓

### The fix — sound and coverage-preserving

- Mocks fetch (1) too (small curated base `anthropic/claude-sonnet-4.6`), guarded in the existing `try/except` so it's a no-op when the agent is absent — matching the sibling dedupe/fail-closed tests and the module's "without real network access" docstring.
- Asserts across `models` + `extra_models`, so a large curated base overflowing the visible cap no longer flips it. No `skip`, no deleted assertion — the #1426 free-tier-visibility contract is still fully asserted. ✓

### Verification — and an honest caveat

- **Agent-absent (my venv, CI-equivalent): whole file 10/10 green.** ✓
- **Agent-present:** I tried to reproduce the original failing env via `PYTHONPATH=<fresh hermes-agent tarball>`. There, this test still failed — **but so did the unmodified sibling `test_free_tier_cap_prevents_picker_drowning`** (which the PR cites as the already-correct pattern). Since a test this PR didn't touch fails the same way, the free-tier `urlopen` escaping the mock is a **divergence of my tarball stand-in** from a real agent install (the current upstream agent resolves the OpenRouter fetch differently), **not** a defect in this fix. I couldn't independently confirm the agent-present pass because my harness doesn't faithfully replicate the author's installed-agent env — but the fix's logic is correct, it's test-only (zero production risk), and the author reports the agent-present pass on their real install.

### Recommendation

✅ **Approved** (test-only). The diagnosis is precise (fetch (1) bypasses the module-level `urlopen` patch), production is confirmed correct (free-tier surfaces in `extra_models`), and the fix restores hermeticity by mocking both fetches + asserting both picker buckets while preserving the #1426 coverage — a strict improvement over master, which left fetch (1) on the real network. Passes agent-absent; test-only so it can't regress production.

Non-blocking follow-up: the sibling free-tier tests share the same fetch-(2)-mock assumption; if a future agent version ever routes the free-tier augment through a non-`urllib.request.urlopen` path, that whole test family would need the same hermetic treatment (out of scope here).

nesquena-hermes:
Shipped in **exp-v0.52.39** via release #5948. The OpenRouter live-fetch test is now hermetic (mocks fetch_openrouter_models + asserts across both picker buckets). Thanks!

Issue #5924, opened 2026-07-11
WebUI: after a provider failure the recovery flow is broken — model/provider won't switch (edit-resubmit & retry re-send the failed model), no way to clear failed turns, fork unreliable


nesquena-hermes:
Reported by **b3nw** in Discord `#report-bugs` (Jul 10 2026, 10:09 PM / 10:19 PM). Repro on **v0.52.0 experimental**, **Chrome mobile (iOS)**. Original: https://discord.com/channels/1492983281338286121/1493325333980774622/1525368397032849508

The theme: **once a turn fails with a provider error, the recovery flow is broken** — the user can't cleanly switch to a working model/provider, failed turns pile up with no way to clear them, and fork (the only known workaround) is itself unreliable. This is a reliability/UX follow-up centered on the *recovery* path; the underlying "model silently reverts after a provider failure" root cause is tracked separately in #2761.

## Facet 1 — Changing model/provider after a failure doesn't take; the broken model is re-sent

> "after any provider failure it's very difficult to get the web to try a different model/provider. changing the model in the selector and editing to resubmit doesn't work, sends the broken model again. Same with changing the model and hitting the retry."

After a failed turn, both recovery paths fail to honor a new pick:
- **Edit + resubmit** → re-sends the *failed* model, not the newly-selected one.
- **Change model + `/retry`** → same, re-sends the broken model.

The user is **forced to fork the session** to escape.

**Investigation notes.** The frontend selector path looks correct — `boot.js` `$('modelSelect').onchange` updates `S.session.model`/`model_provider`, writes a pending explicit-pick marker (`_rememberPendingSessionModel`), and POSTs `/api/session/update`. The send payload precedence is `S.session.model || modelSelect.value` (`messages.js:_chatPayloadModel`, line ~99), so a selector change *should* propagate. That points the finger at **server-side re-revert of the model on `/api/chat/start`** and/or a **sticky persisted `model_provider`** surviving on the session after the failure — the same shape as the `#3737` cross-provider re-revert the code comments in `messages.js:1689-1709` guard against, and the sticky-provider-after-restart family in #2761 / #5567. The explicit-pick marker is consumed after the first send; a second recovery send may lose it and let the server "repair" the model back to the stuck value. Retry (`api/session_ops.py:retry_last`) only truncates + hands text back to the composer, so it inherits whatever model resolution `send()` computes — it does not force the new pick.

## Facet 2 — Failed turns clutter the session with no way to clear them

[...]

Issue #5731, opened 2026-07-07
Follow-up to #5567: repair already-poisoned session model_provider at the backend chat-start boundary


nesquena-hermes:
## Follow-up to #5567: repair sessions whose `model_provider` was already poisoned before the resolver fix

The frontend write-site fix for #5567 (resolve provider by matching option value, not `sel.selectedOptions[0]`) stops **all new** contamination. This issue tracks the remaining half: **repairing sessions that were already poisoned before that fix shipped** — their session JSON persists a foreign `model_provider` (e.g. a `kilo/…` model stored with `ollama`), and `_modelProviderForSend` re-sends the stored value first, so they stay bricked until a manual model re-pick.

### Why this wasn't shipped in the frontend fix

A 3-round Codex + Fable gate on a frontend load-time repair (`_repairContaminatedSessionModelProvider`, reading the `#modelSelect` dropdown as "catalog truth") found a **new distinct defect every round**, all rooted in the same problem: **the shared `#modelSelect` DOM is not a reliable per-session source of truth.**

- **R1 (Fable):** a string-prefix "contradiction" heuristic nulled *healthy* providers — NVIDIA NIM (`qwen/…`+`nvidia`), Cursor ACP (`cursor/…`+`cursor-acp`), ollama/lmstudio HF-style ids — because those providers legitimately own vendor-prefixed model ids (`_PORTAL_PROVIDERS`, `_LOCAL_SERVER_PROVIDERS`). Nulling → chat-start persists null → backend re-routes to openrouter = the reverse of the bug.
- **R2 (Codex):** made catalog-aware, but a session opened while the dropdown is cold never healed (repair ran once before populate).
- **R3 (Codex ×3 CORE + Fable F1/F1b/F3):** after adding post-populate re-runs, the repair read cross-session state: a previous session's injected `dataset.custom='1'` "echo" option counts as catalog evidence → nulls a healthy provider on session B then stamps session A's provider on it; foreign-profile catalog windows during profile/session-link switches; dedup hides the real owner so the heal never fires; `'default'` pseudo-group miscounts. Fable also proved (F2) that **even fully hardened, a cold-loaded poisoned session on a live-only model stays poisoned** (the dropdown dedup skips the real owner while the poisoned echo holds the value).

Conclusion: a frontend DOM-timing repair is the wrong layer.

### Proposed approach (robust)

[...]

PR #5949, opened 2026-07-11
fix(chat): honor explicit model pick on post-failure recovery send (#5924)

nesquena-hermes:
## Summary

After a provider failure, the recovery flow ignored a freshly-picked model/provider: changing the model in the selector and then **edit-resubmit** *or* **/retry** re-sent the *failed* model, forcing the user to fork the session to escape.

This PR fixes **Facet 1** (the pick not being honored on the recovery send) and **Facet 4** (stuck model surviving fork/refresh/reselect/update — same root cause) by re-arming the explicit-pick marker on both recovery entry points.

Reported by @b3nw (Discord `#report-bugs`).

## Root cause

The frontend selector path is correct — `boot.js` `modelSelect.onchange` writes a pending explicit-pick marker via `_rememberPendingSessionModel(...)`. The problem is that **the marker is single-shot**: `send()` consumes it exactly once (`messages.js`), and neither recovery path re-arms it.

- `submitEdit()` (`static/ui.js`) truncates and calls `send()` directly.
- `cmdRetry()` (`static/commands.js`) does the same after `/api/session/retry`.

Because the marker was already consumed by an earlier send, the recovery send goes out with `explicit_model_pick:false`. The server's `_resolve_compatible_session_model_state` (`api/routes.py`) is then free to "repair" a cross-family model back toward the profile default — re-reverting to the stale/failed value. Facet 4 is the same loop: each non-explicit recovery send lets the server re-pin the persisted `model_provider`, so reselect/refresh/fork never sticks.

This is exactly the layer the maintainer's investigation comment identified: **WebUI-side, small, on the two recovery entry points** — not the server guard (which correctly honors `explicit_model_pick=True` and must keep normalizing stale models when the pick is absent, per #3737 / #5731).

## Fix

Re-arm the pending explicit-pick marker from the **current** selector/session state (`_chatPayloadModel()` / `_chatPayloadModelProvider()`) immediately before `await send()` in both `submitEdit()` and `cmdRetry()`. This makes a recovery send carry `explicit_model_pick:true`, exactly like a fresh `onchange → send`. Because the re-arm runs on *every* recovery invocation, it also survives a **second consecutive** recovery send even though `send()` consumes the marker each time.

- The normal (non-failure) send path is unchanged — no re-arm added there.
- The server-side `_resolve_compatible_session_model_state` guard is untouched; the #3737 / #2761 / #5567 / #5731 repair behavior is preserved (verified by a regression assertion that a *non-explicit* send still normalizes a stale cross-family model).

## Files changed

- `static/ui.js` — `submitEdit()` re-arms the pending pick before `await send()`.
- `static/commands.js` — `cmdRetry()` re-arms the pending pick before `await send()`.
- `tests/test_issue5924_post_failure_recovery_model_pick.py` — new regression test.

## Tests

New file `tests/test_issue5924_post_failure_recovery_model_pick.py` pins the two-layer invariant:

- **WebUI:** both `submitEdit` and `cmdRetry` re-arm the marker (from the current selector state) *before* `await send()`.
- **Server:** an explicit recovery pick is honored (not reverted); a **second consecutive** recovery send still honors it; and a **non-explicit** send still normalizes a stale model (no #3737 regression).

```
tests/test_issue5924_post_failure_recovery_model_pick.py ...... [100%]
6 passed
```

Also re-ran `tests/test_issue_edit_regenerate_absolute_keep_count.py` and `tests/test_model_selection_refresh_persistence.py` (6 passed) to confirm no regression on the touched functions.

## Scope / out-of-scope follow-ups

This is a **Facet 1 + Facet 4** fix. Deliberately out of scope (should be separate PRs):

- **Facet 2** — a clear/dismiss affordance for failed error turns (missing UI feature; error turns are pushed as plain assistant bubbles with no per-turn delete).
- **Facet 3** — "first fork after an error is a full copy": maintainer could not reproduce from `forkFromMessage`; likely a stale `_oldestIdx` / transcript-not-fully-loaded timing issue on the error turn, needs a focused repro.

Addresses #5924 (Facet 1 + Facet 4). Leaving the issue open for Facet 2 and Facet 3 above.

### DIFF
--- a/static/commands.js
+++ b/static/commands.js
@@ -1686,7 +1686,20 @@ async function cmdRetry(){
     if(!S.session||S.session.session_id!==activeSid)return;
     const data=await api('/api/session?session_id='+encodeURIComponent(activeSid));
     if(data&&data.session){S.messages=data.session.messages||[];S.toolCalls=[];if(typeof clearLiveToolCards==='function')clearLiveToolCards();if(typeof _messagesTruncated!=='undefined')_messagesTruncated=false;renderMessages();}
-    $('msg').value=r.last_user_text||'';if(typeof autoResize==='function')autoResize();await send();
+    $('msg').value=r.last_user_text||'';if(typeof autoResize==='function')autoResize();
+    // #5924 (Facet 1 + Facet 4): /retry is a recovery send. The onchange
+    // explicit-pick marker is single-shot and was already consumed by an
+    // earlier send(), so re-arm it from the CURRENT selector state before this
+    // send. Otherwise explicit_model_pick goes out false and the server's
+    // compatible-model resolution re-reverts a freshly-picked cross-family
+    // model back to the failed/stale value. Re-arming every recovery send also
+    // lets a SECOND consecutive retry honor its pick (send() consumes it each time).
+    if(typeof _rememberPendingSessionModel==='function' && typeof _chatPayloadModel==='function' && S.session){
+      const _recoveryModel=_chatPayloadModel();
+      const _recoveryProvider=(typeof _chatPayloadModelProvider==='function')?_chatPayloadModelProvider(_recoveryModel):null;
+      _rememberPendingSessionModel(activeSid, _recoveryModel, _recoveryProvider);
+    }
+    await send();
   }catch(e){showToast(t('retry_failed')+e.message);}
 }
 async function cmdUndo(){
--- a/static/ui.js
+++ b/static/ui.js
@@ -17027,6 +17027,19 @@ async function submitEdit(msgIdx, newText) {
     S.messages = S.messages.slice(0, absoluteKeepCount);
     renderMessages();
     $('msg').value = newText;
+    // #5924 (Facet 1 + Facet 4): edit-resubmit is a recovery send. The
+    // onchange explicit-pick marker is single-shot and was already consumed by
+    // an earlier send(), so re-arm it from the CURRENT selector state before
+    // this send. Otherwise explicit_model_pick goes out false and the server's
+    // compatible-model resolution re-reverts a freshly-picked cross-family
+    // model back to the failed/stale value (mirrors a fresh onchange→send).
+    // Re-arming every recovery send also lets a SECOND consecutive edit honor
+    // its pick even though send() consumes the marker each time.
+    if(typeof _rememberPendingSessionModel==='function' && typeof _chatPayloadModel==='function' && S.session){
+      const _recoveryModel=_chatPayloadModel();
+      const _recoveryProvider=(typeof _chatPayloadModelProvider==='function')?_chatPayloadModelProvider(_recoveryModel):null;
+      _rememberPendingSessionModel(S.session.session_id, _recoveryModel, _recoveryProvider);
+    }
     await send();
   } catch(e) { setStatus(t('edit_failed') + e.message); }
 }
--- a/tests/test_issue5924_post_failure_recovery_model_pick.py
+++ b/tests/test_issue5924_post_failure_recovery_model_pick.py
@@ -1,0 +1,143 @@
+"""Regression coverage for #5924 — post-failure recovery must honor a fresh pick.
+
+After a provider failure the reporter (@b3nw) could not switch models: changing
+the model in the selector and then **edit-resubmit** or **/retry** re-sent the
+*failed* model, forcing a session fork to escape.
+
+Root cause (Facet 1 + Facet 4): the onchange explicit-pick marker
+(``_rememberPendingSessionModel``) is single-shot — ``send()`` consumes it once
+(``messages.js``). The two recovery paths (``submitEdit`` in ``ui.js`` and
+``cmdRetry`` in ``commands.js``) truncate and call ``send()`` directly WITHOUT
+re-arming the marker, so ``explicit_model_pick`` goes out ``false`` and the
+server's ``_resolve_compatible_session_model_state`` re-reverts a freshly-picked
+cross-family model back to the profile default.
+
+Two-layer invariant pinned here:
+  * WebUI: both recovery paths re-arm the pending explicit-pick marker from the
+    CURRENT selector state *before* ``await send()`` (so a recovery send —
+    including a SECOND consecutive one — carries ``explicit_model_pick:true``).
+  * Server: with ``explicit_model_pick=True`` the fresh cross-family pick is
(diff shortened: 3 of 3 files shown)


nesquena (approved):

Independent review of #5924 (b3nw): after a provider failure, changing the model then **edit-resubmit** or **/retry** re-sent the *failed* model — the pick wasn't honored on the recovery send (Facet 1), and the stuck model survived reselect/refresh/fork (Facet 4).

### What this ships

`static/ui.js` (`submitEdit` re-arm), `static/commands.js` (`cmdRetry` re-arm), `tests/test_issue5924_post_failure_recovery_model_pick.py` (+new). Agent-authored (`nesquena-hermes`). MERGEABLE, CI green (18/18).

The explicit-pick marker (`_rememberPendingSessionModel`, set by `modelSelect.onchange`) is **single-shot** — `send()` consumes it once. The two recovery entry points (`submitEdit` truncate→`send()`, `cmdRetry` after `/api/session/retry`→`send()`) never re-armed it, so the recovery send went out `explicit_model_pick:false` and the server's `_resolve_compatible_session_model_state` was free to normalize the freshly-picked cross-family model back toward the stale/failed value (the #3737 repair). Facet 4 is the same loop.

The fix re-arms the marker from the **current** selector/session state (`_chatPayloadModel()` / `_chatPayloadModelProvider()`) immediately before `await send()` in both paths, so a recovery send carries `explicit_model_pick:true` exactly like a fresh `onchange→send`. Because it re-arms on *every* recovery invocation, a **second consecutive** recovery send still honors the pick (send() re-consumes each time). ✓

- **Server guard untouched:** `_resolve_compatible_session_model_state` is not modified; it still honors `explicit_model_pick=True` and still normalizes stale models when the pick is **absent** (the #3737/#2761/#5567/#5731 behavior), asserted by `test_non_explicit_send_still_normalizes_stale_model`. ✓
- **Mechanism verified:** `_chatPayloadModel()` = `S.session.model || modelSelect.value`; the selector's `onchange` updates `S.session.model` to the pick, so the re-arm reads the user's *fresh* model (not the failed one) — which is what makes Facet 1 actually resolve. ✓
- **Normal `send()` path unchanged** — the re-arm lives only in the two recovery entry points; a plain new message still consumes-then-doesn't-rearm → non-explicit → server normalization still applies. ✓
- Defensive `typeof` guards on `_rememberPendingSessionModel`/`_chatPayloadModel` keep it harness-safe. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Fail → change model → edit-resubmit | fresh model honored (explicit) ✅ (was reverted) |
| Fail → change model → /retry | fresh model honored ✅ |
| Two consecutive recovery sends | both honor the pick (re-armed each time) ✅ |
| Plain new message (no edit/retry) | non-explicit → server normalizes stale model ✅ (no #3737 regression) |
| Recovery send, model unchanged | sends the current session model as explicit ✅ |

### Minor observation (non-blocking)

`submitEdit`/`cmdRetry` are the generic edit/retry paths, so they now mark **every** such send explicit — meaning the server's stale-model normalization no longer applies to edit/retry (only to fresh passive sends). In the narrow case of a *stale-and-incompatible* `S.session.model` that the user edits without touching the selector, the recovery send would now carry that model verbatim rather than letting the server repair it. This is a defensible design choice (an explicit user action honors the session's configured model, and the #5924 flow is precisely "the user picked a working model"), and the server fallback still covers passive sends — flagging only for awareness, not as a blocker.

### Tests

- `tests/test_issue5924_post_failure_recovery_model_pick.py` — **6/6**: both paths re-arm before `await send()` (structural, order-checked), re-arm reads `_chatPayloadModel()`, explicit recovery pick honored server-side, second-consecutive honored, and non-explicit still normalizes (no #3737 regression).
- `node --check` clean on ui.js + commands.js.
- **Full suite: 12586 passed / 0 failed** in 425s (deselected the known darwin CRLF flake). CI green (18/18).

### Recommendation

A precise WebUI-side fix at the two recovery entry points: re-arming the single-shot explicit-pick marker from the current selector before the recovery `send()`, so edit-resubmit and /retry honor a freshly-picked model instead of the failed one (and survive a second consecutive recovery). The server normalization guard is untouched and still covers non-explicit sends (no #3737 regression, asserted). The one nuance (edit/retry always explicit) is a sound design choice with the passive-send fallback intact. RED on master; full suite clean. Ship.

nesquena-hermes:
## 🔬 Self-gate (Codex) — SHIP ONLY WITH FIXES (4 findings, holding for rework)

Rebased onto current master (clean) + ran the full Codex gate. The fresh-pick model/provider path is wired correctly and normal `send()` is unchanged, but the gate reproduced **4 real defects** on the recovery/retry send path — 2 CORE + 2 SILENT cross-session races:

1. **CORE — `commands.js:1697`:** `/retry` re-arms the explicit-pick marker **unconditionally** from current selector state, not from a genuine deliberate pick. So a recovery send with NO fresh pick still forces `explicit_model_pick`, which can push the current/stale model instead of letting the server's compatible-model resolution run. Fix: re-arm only from a session-scoped deliberate-pick record created by `modelSelect.onchange`, preserved across consecutive recovery attempts.
2. **CORE — `ui.js:17038`:** edit-resubmit has the same unconditional-explicit-pick regression. Same deliberate-pick gating needed.
3. **SILENT — `commands.js:1687`:** switching sessions during the 2nd `/retry`'s session GET writes session **B**'s model/provider into session **A**'s pending marker (reproduced: `pending.A = {model: model-b, provider: provider-b}`) → loading A later applies + persists a foreign model. Fix: after the GET await, `return` unless `S.session.session_id === activeSid`.
4. **SILENT — `ui.js:17023`:** session-switch while truncate is pending lets the re-arm apply recovery intent to the newly-visible session. Fix: repeat the `initialSid` guard immediately after the truncate await.

Render surface is clean (crown-jewel stream-regression-gate GREEN on the rebased head, all modes). Holding this for the deliberate-pick-record rework + the 2 session guards + regression tests for no-fresh-pick and the second-await switch race, then re-gate before ship.

nesquena-hermes:
Shipped in **exp-v0.52.45** via release #5964 (closes #5924). Recovery (/retry + edit-resubmit) now honors a genuinely-picked model without cross-session leak — converged over a 4-round Codex gate (deliberate-pick + fire-time re-arm guards). Thanks @b3nw.

Issue #5224, opened 2026-06-29
Bug: Terminal stream failures can hide an existing chat transcript


claw-io:
## Summary

A terminal chat failure can make the active WebUI chat pane hide or appear to wipe a previously visible transcript. In the observed incident, a single user submission produced a long agent/tool transcript, then a terminal model/provider error arrived and the active chat pane showed the error state as if prior work had disappeared.

A follow-up check of the WebUI session sidecar showed the transcript was still durably present, so this appears to be a client-side render/session-settlement/projection failure rather than permanent deletion. It is still user-facing data loss because the active UI can stop showing the accumulated work exactly when the user needs it for diagnosis.

## Environment

- WebUI: v0.51.742
- Agent: 0.17.0

## Observed behavior

- One user message starts a normal WebUI turn.
- The agent produces many assistant/tool steps.
- A later terminal error arrives.
- The visible chat pane collapses to the error state and prior transcript rows appear gone.
- Inspecting persisted WebUI session state afterwards shows the transcript still exists.
- A second incident with the same terminal model-error class retained visible history, so the transcript failure appears conditional and separate from the model/provider root cause.

## Expected behavior

For any terminal event (`done`, `apperror`, SSE/network error, replay/restore failure, cancel, or server-side wakeup failure):

- existing visible user/assistant/tool history remains visible;
- the terminal failure appends or merges a clear error marker into the active turn;
- the client does not replace a messageful transcript with a stale, shorter, or empty `session.messages` snapshot;
- ambiguous recovery keeps the best local/in-flight transcript and offers refresh/retry instead of blanking the pane.

## Likely code paths to audit

Client-side replacement points in `static/messages.js` look high-risk:

- `done` handler replacing `S.messages` from `d.session.messages`;
- `apperror` handler replacing `S.messages` from embedded `d.session.messages`;
- `_restoreSettledSession()` replacing `S.messages` from `GET /api/session` during stream-end/error recovery.

These paths should reject or merge stale/shorter terminal snapshots instead of blindly treating them as authoritative.

## Suggested regression coverage

1. Start with a same-session `S.messages` containing a messageful transcript: user + several assistant/tool rows.
2. Simulate an `apperror` event with either no embedded `session`, or an embedded `session.messages` shorter than the local transcript.
3. Assert the transcript is preserved and exactly one error marker is appended or merged into the active turn.
4. Repeat for `_restoreSettledSession()` returning a shorter/stale snapshot after SSE loss.
[...]

PR #5860, opened 2026-07-09
fix(ui): lazy-render collapsed settled worklog rows to prevent browser freeze (#5839 render half)

nesquena-hermes:
## Summary

Fixes the **browser-freeze half** of #5839 (the backend server-crash half is #5858). A settled turn's activity worklog — the reasoning + tool steps behind the "Processed Xm Ys" chip — can carry **80+ rows**. In compact-worklog mode (the default), `_renderSettledAnchorSceneForMessage` built **every** row's DOM even while the worklog group was **collapsed**, so a long reasoning-heavy history accumulated tens of thousands of hidden nodes. A later synchronous layout read — opening the model/workspace dropdown calls `getBoundingClientRect()`/`offsetWidth` — then forced a full layout over that oversized DOM and could push the tab into a multi-GB freeze. (@kertisalex's report: freeze on opening a heavy session and on dropdown clicks afterward.)

## Fix — lazy settled-worklog DOM

Collapsed settled worklogs defer building their row DOM until first expand:

- **`_renderSettledAnchorSceneForMessage`** — when the group is collapsed, stash the rows on the element, mark `data-worklog-rows-deferred="1"`, and render only the summary chip. The chip label reads `data-turn-duration` (not the rows), so "Processed Xm Ys" shows correctly with zero rows built. The keep-open/expanded path still renders eagerly (no change to the just-settled turn).
- **`_toggleActivityGroup`** — materializes the deferred rows exactly once, on first expand.
- **blank-turn safety reveal** — also materializes before force-expanding, so a turn whose only content is its worklog never shows empty.
- **HTML-cache restore** — `innerHTML` round-trips drop JS properties, so after a session switch-and-back the stashed rows are gone. `_deferredWorklogRowsFromGroup` recovers them from the owning message via the disclosure key (`anchor-scene:<rawIdx>` → `S.messages[rawIdx]._anchor_activity_scene`), and `_rehydrateDeferredWorklogsFromCache` re-stashes them on the cache fast-path restore. The materialize helper also self-recovers via the same path as a fallback.

This matches how ChatGPT ("Thought for Xs"), Claude, and Open-WebUI render settled reasoning — behind a lazily-populated disclosure, not eager step DOM.

## Verified in a browser (reporter's real session)

Ran the actual reporting session (`be2d3b22e7ed`, 124 msgs, 3 scenes of 79/44/11 rows) on a seeded test server:

[...]

PR #5950, opened 2026-07-11
fix(streaming): keep errored-turn assistant response visible (#5941)

### LINKED ISSUE #5941: WebUI: on an errored turn the assistant's actual response is auto-collapsed/hidden, so the user assumes nothing was returned
Reported by **b3nw** in Discord `#report-bugs` (Jul 11 2026). Repro on **exp-v0.52.38 · Experimental**, Agent 0.18.0.

> "when error occurs the actual response is auto-collapsed and hidden, leaving user to assume nothing was returned, yet there is a response that can be expanded."

## Symptom

When a turn ends in an error, the assistant's **actual response for that turn (its tool calls and text output) IS generated and persisted**, but the whole turn collapses down to just the error bubble + a single collapsed header. The user sees the error and reasonably assumes **nothing came back** — but the real response is still there, hidden behind an expandable/collapsed section. Expanding it reveals the full tool trace and text that was produced before the error.

So the content isn't lost; the **default post-error render state hides it**, which reads as data loss to the user.

## Why this matters

- Users think their work/response vanished on any errored turn, when in fact partial (or complete) output exists.
- Combined with the misleading error text (see companion report) and the broken recovery flow (#5924), an errored turn currently looks far more destructive than it is.

## Expected

When a turn errors but has produced assistant content, that content should remain **visible** (or at minimum have an obvious, clearly-labeled affordance indicating "a response was produced — expand to view"), not silently collapse so the turn looks empty.

## Investigation notes

The collapse appears to be the turn's post-completion render reconciling to a collapsed state once the error/terminal event lands, rather than preserving the streamed/visible body. Worth checking the turn-finalize path in `static/messages.js` (the error-branch render around the `no_response`/apperror handling) and whether the error classification forces a collapsed header regardless of whether `content` was present.

## Acceptance criteria
- An errored turn that produced assistant text/tool output shows that output by default (or with an unmistakable expand affordance), so the user does not conclude nothing was returned.

## Environment
- WebUI: exp-v0.52.38 (experimental) · Agent 0.18.0

## Related
- #5924 — after-failure recovery flow
- #3929 — turn-level error handling discards partial work
- #2761 — compression card hides response (adjacent shape)

nesquena-hermes:
## Summary

On an errored turn, the assistant's **actual produced response** (tool calls + reasoning) was auto-collapsed behind a single header above the error card, so users reasonably concluded that nothing came back — even though the full response was still there, one click away. This keeps that produced content **visible by default** on an errored turn.

## Root cause + fix

When a turn ends in a provider/agent failure it still projects a settled anchor "activity scene" (the collapsible worklog rail) carrying the tool/reasoning rows it produced, plus a `terminal_state` of `error` / `no_response` / `tool_limit_reached` / etc. The settled-scene renderer `_renderSettledAnchorSceneForMessage` (static/ui.js) built that worklog with `collapsed: !keepSettledWorklogOpen` **unconditionally** — it never consulted the turn's `terminal_state`, so an errored-but-content-bearing turn collapsed exactly like a normal completed turn, hiding the response and reading as data loss. The fix classifies the scene's `terminal_state`: when it is an error/failure state (not a normal `completed`), the worklog is kept **expanded by default** so the produced response stays visible. A user who has explicitly collapsed that turn's worklog (saved `closed` disclosure state) is still respected, and the genuinely-empty errored turn is untouched — the render gate at the top of the function already requires a worklog-worthy scene (≥1 tool/thinking/compression row), so a real `no_response` with zero produced content never reaches the collapse decision and still shows only its bare error card.

## Scope / safety

- Smallest correct change on the crown-jewel streaming renderer — one new predicate + one term in the existing `collapsed` expression. No refactor.
- Normal (completed) turns, user-cancelled/interrupted turns, and genuinely-empty errored turns are behavior-unchanged.
- `cancelled`/`interrupted` are deliberately excluded (they own dedicated cards + #5224 transcript preservation).

## Tests

New: `tests/test_issue5941_errored_turn_response_visible.py` — structural lock on the render-gate wiring plus behavioral (Node-executed) coverage of the terminal-state predicate and the full collapse decision matrix (errored default → visible; errored + user-collapsed → collapsed; completed → collapsed as before).

```
$ ./scripts/... (repo .venv) -m pytest tests/test_issue5941_errored_turn_response_visible.py -p no:xdist -q
...                                                                      [100%]
3 passed in 3.96s
```

Related crown-jewel regressions re-run clean (no regressions):

```
$ pytest tests/test_pure_prose_turn_not_worklog.py tests/test_issue5839_lazy_worklog_render.py \
         tests/test_issue5224_terminal_error_transcript_preserve.py tests/test_cancelled_turn_status.py -p no:xdist -q
32 passed in 4.80s
```

## Attribution

Reported by **@b3nw** (Discord `#report-bugs`).

Closes #5941

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -12445,6 +12445,23 @@ function _anchorSceneSceneHasWorklogWorthyRows(scene){
   }
   return false;
 }
+// #5941: an errored/failed turn's terminal_state. A turn that ended in a
+// provider/agent failure but which DID produce assistant content (tool calls,
+// reasoning) still folds that content into a collapsed worklog above the error
+// card — so the user reads a lone error bubble as "nothing came back", even
+// though the real response is one click away. These are the terminal states
+// that must keep the produced content VISIBLE by default. `completed` (normal
+// turn) and null are deliberately excluded, and `cancelled`/`interrupted`
+// (user-initiated stops with their own dedicated cards + #5224 transcript
+// preservation) are left to their existing behavior — this is scoped to the
+// error/failure family the report is about.
+const _ANCHOR_SCENE_ERRORED_TERMINAL_STATES=new Set([
+  'error','no_response','degraded','connection_lost','tool_limit_reached','compression_exhausted',
+]);
+function _anchorSceneHasErroredTerminalState(scene){
+  const state=String(scene&&scene.terminal_state||'').trim().toLowerCase();
+  return _ANCHOR_SCENE_ERRORED_TERMINAL_STATES.has(state);
+}
 function _renderSettledAnchorSceneTransparentForMessage(message, segment, rawIdx){
   if(!message||!message._anchor_activity_scene||!segment) return false;
   if(!_anchorSceneSceneHasWorklogWorthyRows(message._anchor_activity_scene)) return false;
@@ -12570,6 +12587,19 @@ function _renderSettledAnchorSceneForMessage(message, segment, rawIdx){
   if(streamId&&!_readActivityDisclosureState(activityKey)){
     _copyActivityDisclosureState(`live:${streamId}`, activityKey);
   }
+  // #5941: an errored turn that produced assistant content (tool calls /
+  // reasoning) must not hide that content behind a collapsed header — the user
+  // reads a lone error card as "nothing came back". When the settled scene's
+  // terminal_state is an error/failure (NOT a normal completion) keep the
+  // worklog EXPANDED by default so the produced response stays visible. This
+  // path is only reached for worklog-worthy scenes (the guard at the top
+  // requires >=1 tool/thinking/compression row), so a genuinely-empty errored
+  // turn — a real no_response with zero produced content — never gets here and
+  // still shows only its error card, no phantom empty body. A user who has
+  // explicitly collapsed THIS turn's worklog (saved 'closed' disclosure state)
+  // is still respected, so the default-open never fights an intentional collapse.
+  const erroredWorklogKeepOpen=_anchorSceneHasErroredTerminalState(scene)
+    && _readActivityDisclosureState(activityKey)!=='closed';
   // keepSettledWorklogOpen forces collapsed:false for the ONE height-stable settle
   // render of the just-settled turn (no STREAM_DONE shrink jump) for both pinned
   // followers AND unpinned mid-turn readers. The keep-open is made genuinely
@@ -12581,7 +12611,7 @@ function _renderSettledAnchorSceneForMessage(message, segment, rawIdx){
   // (_isKeepSettledWorklogOpenArmed), so it never persists across restores.
   const group=_anchorSceneWorklogGroup(blocks,{
     live:false,
-    collapsed:!keepSettledWorklogOpen,
+    collapsed:!(keepSettledWorklogOpen||erroredWorklogKeepOpen),
     beforeAnchor:true,
     anchor:segment,
     activityKey,
--- a/tests/test_issue4970_stream_done_shrink_regression.py
+++ b/tests/test_issue4970_stream_done_shrink_regression.py
@@ -75,7 +75,12 @@ def test_helper_and_token_threaded_through_render():
     )
     render_fn = _function_body(UI_JS, "_renderSettledAnchorSceneForMessage")
     assert "_shouldKeepSettledWorklogOpenForStreamSettle(streamId)" in render_fn
-    assert "collapsed:!keepSettledWorklogOpen" in render_fn
+    # keepSettledWorklogOpen must still drive the settled-render collapsed flag.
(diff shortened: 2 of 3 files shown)


nesquena (approved):
## Review — end-to-end ✅ (approved after one CI-red fix I pushed)

Independent review of #5941 (b3nw): on an errored turn the assistant's produced response (tool calls + reasoning) was auto-collapsed behind the error card, reading as "nothing came back." This keeps that content visible by default on error/failure turns.

### What this ships

`static/ui.js` (one predicate + one term in the settled-render `collapsed` expression), `tests/test_issue5941_errored_turn_response_visible.py` (+new). Agent-authored (`nesquena-hermes`). MERGEABLE, CI green (18/18 after my fix).

`_renderSettledAnchorSceneForMessage` built the settled worklog with `collapsed: !keepSettledWorklogOpen` unconditionally, never consulting the turn's `terminal_state`. So an errored-but-content-bearing turn collapsed like a normal completed one, hiding the response. The fix:
- `_anchorSceneHasErroredTerminalState(scene)` matches `terminal_state ∈ {error, no_response, degraded, connection_lost, tool_limit_reached, compression_exhausted}`. I verified against `static/assistant_turn_anchors.js` that these are the **actual** terminal-state values, and that the set correctly **excludes** `completed`/null (normal) and `cancelled`/`interrupted` (user stops with their own cards + #5224 preservation). ✓
- `erroredWorklogKeepOpen = errored && _readActivityDisclosureState(activityKey) !== 'closed'` → keep open by default **but respect a user's explicit collapse**. ✓
- `collapsed: !(keepSettledWorklogOpen || erroredWorklogKeepOpen)` — adds the errored keep-open to the existing settle keep-open. ✓

### Interaction with #5860 (lazy worklog) + #5224 — sound

- **#5860:** for an errored turn `collapsed` is now `false`, so the render takes the eager path (`_renderAnchorSceneRowsIntoWorklog`) — it never enters #5860's collapsed-defer branch, so the produced rows materialize immediately (exactly the intent). Collapsed non-errored worklogs still defer as before. No conflict. ✓
- **Empty errored turn:** the function's top gate (`if(!_anchorSceneSceneHasWorklogWorthyRows(...)) return false;`, line 3) early-returns when there are zero tool/thinking/compression rows — so a genuine `no_response` with no produced content never reaches the collapse decision and still shows only its bare error card (no phantom empty body). ✓

### What I caught — full-suite CI-red the PR missed

The PR's targeted tests + its cited crown-jewel regressions passed, but the **full suite** turned up `test_issue4970_stream_done_shrink_regression::test_helper_and_token_threaded_through_render` failing: it asserted the literal `collapsed:!keepSettledWorklogOpen`, which this PR changed to `collapsed:!(keepSettledWorklogOpen||erroredWorklogKeepOpen)`. The PR updated its own touched tests but missed this pre-existing brittle source-string assertion.

**Pushed `8ad3a781`:** re-anchored #4970 to `collapsed:!(keepSettledWorklogOpen` — still guards #4970's invariant (`keepSettledWorklogOpen` negates into the settled-render collapsed flag) without pinning the exact term list. #4970 + #5941 both green; ruff clean.

### Edge-case matrix

| Turn | Worklog |
|---|---|
| errored (error/no_response/degraded/…) with content | **expanded by default** — response visible ✅ (was hidden) |
| errored + user explicitly collapsed (`closed`) | stays collapsed (respected) ✅ |
| errored with zero produced content | early-return → bare error card, no phantom body ✅ |
| completed (normal) turn | collapsed as before ✅ |
| cancelled / interrupted | unchanged (own cards + #5224) ✅ |
| collapsed non-errored worklog | still #5860-deferred ✅ |

### Tests

- `tests/test_issue5941_errored_turn_response_visible.py` — **3/3** (render-gate wiring lock + Node behavioral matrix: errored→visible, errored+user-collapsed→collapsed, completed→collapsed).
- **RED-on-master verified independently:** the decision-matrix test fails against master's `ui.js`. Load-bearing.
- settled-scene/worklog blast-radius — **278 passed, 5 skipped**.
- **Full suite: 12583 passed / 0 failed** in 439s after `8ad3a781` (deselected the known darwin CRLF flake). CI green (18/18).

### Recommendation

The smallest correct change on the settled-render collapse decision: errored/failure turns with produced content stay expanded (respecting an explicit user collapse), empty errored turns are gated out (no phantom body), and completed/cancelled/interrupted are untouched — with the #5860 lazy-render and #5224 interactions verified. The one gap — a pre-existing #4970 brittle assertion the expression change broke — I fixed in `8ad3a781`. RED on master; full suite clean. Ship.

nesquena-hermes:
Shipped in **exp-v0.52.45** via release #5964 (closes #5941). Errored turns now keep the produced response visible; stream-gate GREEN + Codex SAFE. Thanks @b3nw for the report + repro.

PR #5953, opened 2026-07-11
Release exp-v0.52.42: ?profile= deep links (#5730) + Russian locale refresh (#5778)

nesquena-hermes:
## Release exp-v0.52.42

Boot-path batch — both PRs rebased onto current master, gate-clean. Nathan-approved (boot-path pair).

### #5730 — feat(profile): support profile query deep links (rodboev)
`?profile=NAME` on boot switches to that profile, validated against the existing allowlist (nonexistent → boots normally, `../evil` traversal + encoded chars rejected, cross-profile access-enforced, try/catch fail-safe). **Codex-verified:** backend existence/containment/isolated-profile/signed-cookie/CSRF/session-profile checks all still enforced; traversal + encoding attempts rejected; invalid profiles preserve normal restoration.

### #5778 — Refresh Russian localization bundle (DrMaks22)
`ru` refreshed to full key parity with English (Codex-verified: en/ru both carry the same 1,689 keys, all new locale values nonempty + safely rendered), onboarding provider notes localized via an additive `provider_note_key` (English preserved as fallback — no existing path broken).

### Gate
- **Stream-regression-gate GREEN** all 3 modes on the rebased stage (transparent/compact clean first pass; hide_all_activity flaked once on a partial mock drive → CLEAN on isolated re-run, texts=4).
- **Codex: SHIP** — one finding (`boot.js:3476` reasoning-chip not refreshed on empty normal boot) was verified **pre-existing on master** (byte-identical, NOT introduced by this batch) → filed as a follow-up, not a blocker.
- 136 focused tests + JS syntax/lint pass; full suite + CI matrix running.

**Channel:** experimental (`exp-v0.52.42`).

PR #5955, opened 2026-07-11
Fix public share locale values

DrMaks22:
## Thinking Path

After #5778 shipped through release #5953 / `exp-v0.52.42`, I re-read the shipped bundle and found a value-level locale swap that key-parity checks do not catch: the new public-share strings were Russian in the English locale and English in the Russian locale.

## What Changed

- Restored the public-share strings in the `en` locale to English.
- Restored the matching public-share strings in the `ru` locale to Russian.
- Added a focused regression test so these values cannot be swapped while key parity remains green.

## Why It Matters

The shipped key sets were balanced, but English users would see Russian public-share labels and Russian users would see English fallback text for that surface. This keeps the #5778 localization refresh correct at the value level, not only the key level.

## Verification

- `./scripts/test.sh tests/test_russian_locale.py tests/test_login_locale_parity.py` -> `38 passed`
- `npm run lint:runtime` -> pass
- `python3 scripts/scope_undef_gate.py` -> pass
- `python3 -m compileall -q api server.py bootstrap.py mcp_server.py tests scripts` -> pass
- `node --check static/i18n.js` -> pass
- `git diff --check` -> pass
- `lint_ru_quality.py` -> `0` findings
- `audit_hardcoded_ui.py` -> `0` hardcoded UI strings
- i18n audit: `en=1689`, `ru=1689`, `missing=[]`, `extra=[]`; `en.share_session = "Share"`, `ru.share_session = "Поделиться"`.

## Risks / Follow-ups

Low risk: this only swaps the affected public-share values back to their intended locale blocks and adds a regression test. Other locales keep their existing fallback values.

## Model Used

GPT-5 Codex

## Contract Routing

i18n locale bundle only; no runtime behavior or login locale contract changes.


DrMaks22:
@nesquena-hermes small follow-up to #5778/#5953: fixes the public-share locale value swap I found after re-reading the shipped bundle. Checks are green (18/18), branch is current, and the regression test covers the en/ru value-level swap that key parity missed.

nesquena-hermes:
Shipped in **exp-v0.52.46** via release #5968. Fixes the public-share locale value-swap from #5778 — en restored to English, ru to Russian, with a regression test guarding both directions. Verified the full en locale block is now free of non-Latin text. Thanks @DrMaks22 for catching + fixing this fast!

PR #3615, opened 2026-06-04
feat: WebUI project context prefix + session load perf fixes

psanger:
## Summary

- **WebUI project context**: injects the active project name as a prefixed ephemeral prompt so Hermes is aware of the WebUI session's project context when responding
- **Rebase-update path**: adds `/rebase-update` flow for sessions with diverged local commits
- **Session load perf** (two fixes discovered while testing):
  - `_limited_webui_messages_for_display` was calling `_webui_sidecar_lineage_messages_for_display`, which stitched up to 20 compression-snapshot parents via `merge_session_messages_append_only` — O(n² × content_length) for large sessions. Fixed by using `session.messages` directly for paginated loads; full lineage traversal only runs on unbounded (`msg_limit`-absent) loads.
  - `_normalized_session_message_content` normalised full content strings (up to 100KB from tool outputs), making visible-key tuple comparisons O(content_length). Fixed by truncating to 500 chars + appending `len()` suffix so distinct messages still produce different keys.
  - Combined result: 7,254-message session load drops from ~37–75s → ~280ms.
  - Also pushes `tail_limit` into the state.db SQL query (`LIMIT` in a DESC subquery) so the DB scan is bounded for paginated loads.

## Test plan

- [ ] Open a long-running session (1000+ messages) and confirm it loads in <1s
- [ ] Verify project context prefix appears in Hermes responses when a WebUI project is active
- [ ] Confirm rebase-update flow works on a session with diverged commits
- [ ] Run test suite: `scripts/run_tests.sh tests/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5957, opened 2026-07-11
feat(webui): pass passive project metadata to agent

psanger:
## Summary

Pass the assigned WebUI project's ID and display name to Hermes as passive runtime metadata.

- extends the existing ephemeral WebUI surface-context block with `Project ID` and `Project name`
- resolves the project name from the session's existing `project_id`
- covers direct streaming, gateway-backed streaming, and synchronous chat paths
- omits project fields entirely for unassigned sessions

## Scope and behavior

This is intentionally narrower than #3615. It does **not**:

- add or restore model-facing user-message prefixes
- create or maintain project dossiers
- ask the model to suggest assigning chats to projects
- change session loading, transcript reconciliation, updates, or UI behavior

The metadata is informational only. Existing project/session storage remains unchanged.

## State and prompt-cache invariants

- No persisted session or display message is mutated.
- Streaming and gateway paths use the existing per-turn `ephemeral_system_prompt` surface-context channel.
- Sync chat adds the same two passive values to its existing non-persisted system message.
- Unassigned sessions retain the current prompt shape apart from the new supported field definitions.

## Tests

- `./scripts/test.sh tests/test_issue1913_workspace_prefix_sentinel.py tests/test_workspace_display_prefix.py tests/test_notify_on_complete_webui.py tests/test_webui_surface_context.py` — 23 passed
- `./scripts/test.sh tests/test_webui_gateway_chat_backend.py tests/test_gateway_approval_legacy_path.py tests/test_gateway_approval_runs_api.py tests/test_issue1217_transcript_compaction.py` — 101 passed
- `python3 -m py_compile api/streaming.py api/gateway_chat.py api/routes.py`
- `git diff --check`

## Release note

WebUI-assigned project IDs and names are now available to Hermes as passive per-turn runtime context without changing visible or persisted chat messages.

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -665,6 +665,7 @@ def _run_gateway_chat_streaming(
                 _normalize_prefill_messages_before_user_turn,
                 _public_prefill_context_status,
                 _webui_ephemeral_system_prompt,
+                _webui_project_surface_context,
             )
 
             prefill_context = _load_webui_prefill_context(cfg)
@@ -680,6 +681,7 @@ def _run_gateway_chat_streaming(
                     "session_id": session_id,
                     "profile": getattr(s, "profile", None),
                     "workspace": s.workspace if s is not None else str(workspace),
+                    **_webui_project_surface_context(s),
                 },
                 config_data=cfg,
             )
--- a/api/routes.py
+++ b/api/routes.py
@@ -21869,11 +21869,19 @@ def _handle_chat_sync(handler, body):
                 _restore_reasoning_metadata,
                 _sanitize_messages_for_api,
                 _context_messages_for_new_turn,
+                _webui_project_surface_context,
                 _workspace_context_prefix,
             )
             workspace_ctx = _workspace_context_prefix(str(s.workspace))
+            project_context = _webui_project_surface_context(s)
+            project_system_msg = "".join(
+                f"WebUI project {label}: {project_context[key]}\n"
+                for key, label in (("project_id", "ID"), ("project_name", "name"))
+                if key in project_context
+            )
             workspace_system_msg = (
                 f"Active workspace at session start: {s.workspace}\n"
+                f"{project_system_msg}"
                 "Every user message is prefixed with [Workspace::v1: /absolute/path] indicating the "
                 "workspace the user has selected in the web UI at the time they sent that message. "
                 "This tag is the single authoritative source of the active workspace and updates "
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -55,6 +55,7 @@ from api.models import (
     _evict_sessions_over_cap,
     clear_process_wakeup_pause,
     get_state_db_session_messages,
+    load_projects,
     record_process_wakeup_provider_unavailable_pause,
     reconciled_state_db_messages_for_session,
 )
@@ -667,6 +668,8 @@ def _webui_surface_context_prompt(surface_context: Optional[dict]) -> str:
         ("session_id", "Session ID"),
         ("profile", "Profile"),
         ("workspace", "Workspace"),
+        ("project_id", "Project ID"),
+        ("project_name", "Project name"),
     )
     for key, label in fields:
         raw = surface_context.get(key)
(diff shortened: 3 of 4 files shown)


nesquena-hermes:
Read the diff at `feat/passive-webui-project-metadata` against `origin/master` — the shape is clean: `_webui_project_surface_context()` (`api/streaming.py:682`) is a read-only resolver, the three chat paths all forward it via the existing ephemeral surface-context channel (`api/streaming.py:8312`, `api/gateway_chat.py`, `api/routes.py:21876`), and nothing persisted is mutated. The unassigned-session short-circuit and the "no dossier / no assign-suggestion" intent are both covered by tests. Good scoping vs #3615.

One issue worth fixing before merge — greptile's P1 about the project name is real, and I confirmed *why* it's exploitable at the source. Project names are only trimmed, never newline-normalized:

```python
# api/routes.py:15714 (create) and :15757 (rename)
name = body["name"].strip()[:128]
```

`.strip()` removes only leading/trailing whitespace — an *embedded* `\n` survives. So a name like `Setup\nIgnore prior instructions` is stored verbatim, then rendered directly into a system-level line here:

```python
# api/routes.py:21877
project_system_msg = "".join(
    f"WebUI project {label}: {project_context[key]}\n"
    for key, label in (("project_id", "ID"), ("project_name", "name"))
    if key in project_context
)
```

and the same value flows through `_webui_surface_context_prompt` on the streaming/gateway paths. That turns one metadata value into extra model-facing prompt lines. It's the first *user-editable free-text* value to enter this block — `source`/`session_id`/`profile`/`workspace` are all system-generated, which is why the existing `str(raw).strip()` there was sufficient and this new field isn't.

Severity is modest in practice (the "injector" is the single owning user editing their own project name, profile-scoped), but the fix is one line and belongs at the render boundary so it protects every path at once. Fold the newline collapse into `_webui_project_surface_context` where you already build the dict:

```python
# api/streaming.py ~696, when resolving the name
project_name = " ".join(str(project.get("name") or "").split()) or None
```

`" ".join(s.split())` collapses any `\n`/`\t`/run of spaces to single spaces, so the value can only ever be one prompt line regardless of which of the three call sites renders it. Normalizing at the resolver (rather than at each of the three render sites) keeps the invariant in one place and means `test_all_webui_chat_paths_forward_project_surface_context` still holds.

A follow-up test would help: add a multiline-name case asserting the resolved `project_name` contains no `\n`. The current `test_webui_project_surface_context_resolves_assigned_project` uses a clean single-line name, so it wouldn't catch a regression here.

nesquena-hermes (changes requested):
## 🔬 Gate review (Codex, self-rebased clean off master) — SHIP ONLY WITH FIXES (design mostly sound; a real project-name injection into the model-facing context block)

Thanks @psanger — the plumbing is mostly right: no cross-session/cross-project bleed (project id/name resolve strictly from this session's own `project_id`), a session with no project omits the block cleanly, it's genuinely ephemeral (no project name written to session JSON / `messages` / `context_messages`), and the direct and gateway paths both route through `_webui_project_surface_context()` + `_webui_ephemeral_system_prompt()`. But an ad-hoc runtime probe reproduced a real **context-block injection**, so it can't ship as-is:

1. **[INJECTION — attacker-controlled project name breaks out of the passive block] `api/streaming.py:677` + `api/routes.py:22021`** — a project display name containing newlines / surface-context delimiters / control chars is interpolated raw into the model-facing surface-context text, so a crafted name creates a **second `Session ID` line and a second `WebUI progress guidance` section**, and a NUL is preserved into both the direct-streaming and gateway prompts. That lets a project name spoof/forge other passive-context fields the model reads. The synchronous fallback has the same raw interpolation via its own renderer. **Fix:** contain project ID/name through **one shared model-facing encoder** that strips/escapes line breaks + block delimiters + control chars (so no attacker-controlled line or field can be created), and use that single encoder in the direct, gateway, AND synchronous paths (don't leave the fallback with its own vulnerable renderer).

2. **[test — source-string only] `tests/test_webui_surface_context.py:78,103`** — the changed test passes but only searches source text (proof-substitution). **Fix:** add behavioral coverage — two distinct sessions/projects each get their own block, a deleted/missing project omits the name (no crash/stale), and a newline/delimiter/control-character name is contained; replace the source-string path assertion with **captured direct + gateway prompt** assertions.

Per-area: cross-project bleed PASS, missing/deleted-project PASS (graceful omit), ephemeral/not-persisted PASS, path parity PASS — but injection FAIL (the reproduced breakout) and the tests don't behaviorally cover it. The concept (surface project metadata to the agent) is a light scope call for the maintainer, but this injection fix is required regardless of that decision — a project name must be contained before it reaches the model-facing block. Self-rebased clean on current master.

Issue #4701, opened 2026-06-22
iOS Safari portrait: scroll-to-bottom button (↓ End) never appears


reinocheong:
## Bug Description

On **iOS Safari (portrait mode)**, the scroll-to-bottom button (`↓ End`, `#scrollToBottomBtn`) never appears when scrolling up in a conversation. In **landscape mode on the same device, the button works perfectly**.

This started with a recent update — the portrait UI layout was changed significantly, and since then the button has been invisible in portrait mode.

## Environment

- **Device:** iPhone (tested on multiple models)
- **Browser:** iOS Safari (both browser and PWA standalone)
- **WebUI Version:** v0.51.587 (latest as of 2026-06-22)
- **Orientation:** Portrait = broken, Landscape = works

## Steps to Reproduce

1. Open Hermes WebUI on iPhone in **portrait** mode
2. Open any conversation with enough messages to scroll
3. Scroll up (drag finger up on the messages area)
4. Observe: the `↓ End` button does NOT appear at the bottom

## Expected Behavior

The scroll-to-bottom button should appear when the user scrolls away from the bottom of the conversation (distance > 80px from bottom), just like it does in landscape mode.

## Additional Notes

- The button element exists in the DOM and its CSS class/styles are correct
- `_scrollPinned` does get set to `false` when scrolling up (touch detection works)
- `_syncScrollToBottomCue(true)` is called but `btn.style.display='flex'` has no visual effect in portrait
- The issue appears to be a **CSS stacking/layout problem** specific to portrait viewport width on iOS Safari — likely related to how `position:absolute` elements inside `.messages-shell` interact with the `overflow:auto` `#messages` container in narrow viewports
- Landscape mode (wider viewport) does not have this issue, suggesting it's a rendering/compositing layer problem in Safari's portrait layout path

## Suspected Root Cause

On iOS Safari in portrait mode, the stacking context created by `#messages` (`position:relative; overflow-y:auto; z-index:0`) may be clipping or covering the absolutely-positioned `#scrollToBottomBtn` button (sibling element with `z-index:10`). This does not happen in landscape where the viewport is wider and Safari's compositing behaves differently.

nesquena-hermes:
[...]

PR #5962, opened 2026-07-12
fix(chat): keep transcript pinned across composer auto-resize (#5514)

### LINKED ISSUE #5514: bug(chat): transcript scrolls up when the composer grows while pinned to bottom

nesquena-hermes:
Closes #5514.

Reported by @JeffArcLogic, who confirmed the prior fix (#5516, v0.51.851) did **not** resolve it and provided the decisive repro: the transcript scrolls up on **every keystroke** while the composer is multi-row and the reader is pinned to the bottom, by an amount **proportional to composer height**, reproducing even at 2 rows, and **only** when scrolled to the absolute bottom.

## Root cause

`autoResize()` (`static/messages.js`) measures the textarea by momentarily setting `height:'auto'` (which collapses a multi-row composer toward its 1-row min), reading `scrollHeight`, then restoring the measured height. That transient collapse **grows the `flex:1` `#messages` viewport**, and reading `scrollHeight` forces a synchronous reflow — so the browser **clamps a bottom-anchored `scrollTop` down** by the collapse delta and does **not** restore it when the height snaps back. The reader is left stranded Δpx above the bottom (Δ ∝ composer height).

Worse, the clamp's async `scroll` event arrives with no `_programmaticScroll` latch held, reads as an upward scroll, and sets the **sticky** `_messageUserUnpinned=true`. From that point the #5516 grow-path re-pin **and** stream auto-follow both dead-end (they no-op on the unpin guard) until the user manually scrolls back to the true bottom. That is why #5516 "shipped and didn't fix it": its re-pin was gated on **net** growth (`el.offsetHeight > _prevComposerH`), but a steady-state keystroke has no net height change, so it never fired — and the first strand then poisoned the pin state.

## Fix

Snapshot `#messages.scrollTop` **before** the `height:'auto'` → measure → restore round-trip and restore it **after**, undoing the transient clamp within the same synchronous task — so the poisoning `scroll` event never fires.

- **Protects pinned readers** (stay glued to the bottom) **and near-bottom readers who scrolled up to re-read** (exact position preserved) — the latter is a case a re-pin-to-bottom approach structurally cannot honor.
- Takes **no** `_programmaticScroll` latch, so it never swallows a gentle trackpad / scrollbar scroll-up mid-typing.
- Inert to iOS dynamic-toolbar reflows (#4701/#4702) — it only compares values within one task.
- The grow-gated `_repinMessagesAfterComposerResize()` and the `#composerWrap` ResizeObserver **remain unchanged** for genuine net growth (a real new row still shrinks the settled viewport and re-pins a still-pinned reader).

## Approach validation

Design was consulted with two independent advisors **before** building (Codex regression review + Fable UX review) and both confirmed the mechanism and endorsed the preserve-`scrollTop` approach over the re-pin-on-every-keystroke alternative (Codex additionally verified the clamp in a real Chromium probe and confirmed synchronous restore is correct; per the HTML spec a rAF-deferred fix would run *after* the sticky-unpin and miss).

## Tests

`tests/test_issue5514_composer_grow_scroll_pin.py`:
- Updated the stale assertion that pinned the old net-growth gate.
- Added a **behavioral node-vm harness** that reproduces the actual `autoResize()` round-trip against a faithful scroll-clamp model (clientHeight-growth persists the downward clamp, as the real DOM does). Proven non-vacuous: on the pre-fix `autoResize` the pinned reader ends `bottomDist=120` (stranded); with the fix, `bottomDist=0`. A companion case asserts a near-bottom **unpinned** reader keeps their exact position and is never re-pinned.

Verified: 9/9 pass with the fix; the two mechanism tests go red on master's `autoResize`. `node --check` clean.

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -6376,16 +6376,36 @@ function autoResize(){
   }
   const el=$('msg');
   const _prevComposerH=el.offsetHeight;
+  // #5514: autoResize() momentarily sets the textarea to height:'auto' (collapses
+  // a multi-row composer toward its 1-row min) before reading scrollHeight and
+  // restoring the measured height. That transient collapse GROWS the flex:1
+  // #messages viewport, and reading scrollHeight forces a synchronous reflow — so
+  // the browser CLAMPS a bottom-anchored scrollTop DOWN by the collapse delta and
+  // does NOT restore it when the height snaps back. The reader is left stranded
+  // Δpx above the bottom on EVERY keystroke while the composer is multi-row (Δ ∝
+  // composer height), and the clamp's async scroll event also sticky-unpins the
+  // reader (_messageUserUnpinned=true), dead-ending the grow-path re-pin below and
+  // stream auto-follow until they manually scroll back. #5516's net-growth gate
+  // never caught this because a steady-state keystroke has no NET height change.
+  // Root-cause fix: snapshot the transcript's scrollTop BEFORE the height
+  // round-trip and restore it AFTER, undoing the transient clamp within the same
+  // synchronous task so the poisoning scroll event never fires. This protects
+  // pinned readers AND near-bottom readers who scrolled up to re-read (their exact
+  // position is preserved), takes no _programmaticScroll latch, and is inert to
+  // iOS dynamic-toolbar reflows. A genuine NET grow/shrink still lands the reader
+  // Δnet off-bottom; the grow-gated re-pin below (and the #composerWrap
+  // ResizeObserver) then snap a still-pinned reader to the true bottom.
+  const _msgs=$('messages');
+  const _prevScrollTop=_msgs?_msgs.scrollTop:0;
   el.style.height='auto';
   el.style.height=Math.min(el.scrollHeight,200)+'px';
+  if(_msgs&&_msgs.scrollTop!==_prevScrollTop) _msgs.scrollTop=_prevScrollTop;
   updateSendBtn();
-  // #5514/#5515: growing the composer shrinks the flex:1 transcript viewport, so
-  // a reader pinned to the bottom would be stranded above it ("scrolls up 1 row
-  // per composer row"). Re-pin the transcript when it's genuinely still pinned —
-  // but only when the composer actually GREW (a shrink enlarges the viewport and
-  // can't strand a reader), so the common no-height-change keystroke skips the
-  // re-pin's DOM read entirely. The #composerWrap ResizeObserver is the safety
-  // net for any growth path that doesn't route through here.
+  // Genuine NET growth (a new row that keeps the composer taller than before)
+  // still shrinks the settled viewport, so a pinned reader must be re-pinned to
+  // the true bottom. Guarded to fire only on real growth and only when genuinely
+  // pinned (the helper no-ops for a scrolled-away reader). The #composerWrap
+  // ResizeObserver is the safety net for growth paths that don't route here.
   if(el.offsetHeight>_prevComposerH && typeof _repinMessagesAfterComposerResize==='function') _repinMessagesAfterComposerResize();
 }
 function scheduleComposerAutoResize(){
--- a/tests/test_issue5514_composer_grow_scroll_pin.py
+++ b/tests/test_issue5514_composer_grow_scroll_pin.py
@@ -12,15 +12,20 @@ composer growth via paths that don't route through the input->autoResize seam
 (paste, draft restore, programmatic value set, reflow) reads to the user as a
 random upward jump.
 
-Fix: `_repinMessagesAfterComposerResize()` (static/ui.js) re-pins the transcript
-to the bottom ONLY when the reader is genuinely still pinned (honors
-`_messageUserUnpinned` / `_scrollPinned` so a reader who scrolled away is never
-yanked back). It is called (a) from autoResize() and (b) from a ResizeObserver on
-the #composerWrap wrapper that catches every height-change path (typed input,
-paste, draft restore, attachment tray / selection chip).
+Fix (v2, #5514 root cause): `autoResize()` now snapshots `#messages.scrollTop`
+BEFORE its `height:'auto'` → measure → restore round-trip and restores it AFTER,
+undoing the transient viewport-grow clamp within the same synchronous task. This
+protects BOTH a bottom-pinned reader and a near-bottom reader who scrolled up to
+re-read (their exact position is preserved), and — because the poisoning async
+scroll event never fires — it stops the clamp from sticky-unpinning the reader
+(which had dead-ended the #5516 grow-path re-pin and stream auto-follow). The
+grow-gated `_repinMessagesAfterComposerResize()` (static/ui.js) and the
+`#composerWrap` ResizeObserver REMAIN for genuine NET growth: they re-pin the
+transcript to the bottom ONLY when the reader is still pinned (honoring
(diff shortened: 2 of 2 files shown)


nesquena (approved):

Independent review of #5514 (@JeffArcLogic): while the composer is multi-row and the reader is pinned to the bottom, the transcript scrolled up on **every keystroke** by an amount proportional to composer height — and the prior fix (#5516) didn't resolve it.

### What this ships

`static/messages.js` (`autoResize` scrollTop preserve), `tests/test_issue5514_composer_grow_scroll_pin.py`. Agent-authored (`nesquena-hermes`). MERGEABLE, CI green (18/18).

`autoResize()` measures the textarea by momentarily setting `height:'auto'` (collapsing a multi-row composer toward its 1-row min), reading `scrollHeight`, then restoring the height. That transient collapse grows the `flex:1` `#messages` viewport, and the `scrollHeight` read forces a synchronous reflow → the browser **clamps** a bottom-anchored `scrollTop` down by the collapse delta and doesn't restore it when the height snaps back. Worse, the clamp's async `scroll` event (no `_programmaticScroll` latch held) reads as an upward scroll and sets the **sticky** `_messageUserUnpinned=true`, dead-ending both #5516's grow-path re-pin and stream auto-follow. #5516 missed it because its gate was **net** growth (`offsetHeight > _prevComposerH`), and a steady-state keystroke has no net change.

The fix snapshots `#messages.scrollTop` **before** the `height:'auto'`→measure→restore round-trip and restores it **after**, within the same **synchronous** `autoResize()` task — so the clamp is undone before the browser dispatches the async scroll event, and the sticky-unpin never fires. I confirmed:
- **Synchronous** — no rAF/`setTimeout`; the restore is inline (per the HTML spec a rAF-deferred fix would run *after* the sticky-unpin and miss, as the PR notes). ✓
- `#messages` is the transcript scroll container (`index.html:439`), distinct from `$('msg')` the textarea. ✓
- **Guarded write** (`scrollTop !== _prevScrollTop`) — only restores if the clamp actually moved it, avoiding a redundant write. ✓
- **No `_programmaticScroll` latch** — restores to `_prevScrollTop` (the position at entry), so a genuine trackpad/scrollbar scroll-up mid-typing is preserved, not swallowed. A **near-bottom unpinned** reader keeps their exact position (a re-pin-to-bottom approach structurally can't). ✓
- **Grow-gated re-pin + ResizeObserver retained** — a genuine net-new row still shrinks the settled viewport and re-pins a still-pinned reader. ✓

### Edge-case matrix

| Scenario | Behavior |
|---|---|
| Pinned reader, keystroke, multi-row composer | stays glued to bottom (was stranded Δpx/keystroke) ✅ |
| Near-bottom reader scrolled up to re-read, keystroke | exact position preserved, not re-pinned ✅ |
| Genuine net-new composer row, still pinned | re-pinned to true bottom (grow gate) ✅ |
| Gentle user scroll-up mid-typing | preserved (no latch, no swallow) ✅ |
| iOS dynamic-toolbar reflow | inert (compares within one task) ✅ |

### Tests

- `tests/test_issue5514_composer_grow_scroll_pin.py` — **9/9**, incl. a node-vm harness driving the real `autoResize()` round-trip against a faithful scroll-clamp model: pre-fix pinned reader ends `bottomDist=120` (stranded), with fix `bottomDist=0`; companion asserts the near-bottom unpinned reader keeps position and is never re-pinned.
- **RED-on-master verified independently:** 3 fail against master's `messages.js` (incl. the unpinned-preserve + steady-state cases). Load-bearing.
- `node --check` clean. scroll/composer blast-radius — **645 passed, 7 skipped**.
- **Full suite: 12648 passed / 0 failed** in 429s (deselected the known darwin CRLF flake). CI green (18/18).

### Recommendation

The right root-cause fix: undo the measure round-trip's transient scrollTop clamp synchronously so the poisoning sticky-unpin never fires — preserving pinned *and* scrolled-up readers, taking no scroll latch, and leaving the net-growth re-pin intact. This is why #5516's net-growth gate couldn't catch a steady-state keystroke. Mutation-proven tests RED on master; full suite clean. Ship.

nesquena-hermes:
Shipped in **exp-v0.52.52** (closes #5514). The transcript no longer jumps up on every keystroke in a multi-row composer when bottom-pinned — root-caused to autoResize()'s height:auto collapse clamping scrollTop + sticky-unpinning; now snapshots+restores scroll position synchronously. Gate: nesquena independent review APPROVED + Codex SAFE (reproduced the fix in real Chromium: 98px→0) + stream-gate GREEN. Thanks @JeffArcLogic for the decisive repro.

PR #5970, opened 2026-07-12
test: stabilize cross-platform WebUI suite

tahabakhit:
## Summary
- make profile skill-statistics coverage choose an actually incompatible platform on macOS
- remove eight Pytest class-scoped instance-fixture deprecation warnings by using state-free module fixtures

The OpenRouter hermetic test fix that originally overlapped with this branch landed upstream while this PR was in progress, so it was intentionally dropped during rebase.

## Verification
- `12836 passed, 22 skipped, 1 xfailed, 2 xpassed, 16 subtests passed`
- `npm run lint:runtime`
- focused YOLO and model-prefix test files passed

No production runtime code changed.

### DIFF
--- a/tests/test_issue3368_model_prefix_shadowing.py
+++ b/tests/test_issue3368_model_prefix_shadowing.py
@@ -40,6 +40,11 @@ NODE = shutil.which("node")
 pytestmark = pytest.mark.skipif(NODE is None, reason="node not on PATH")
 
 
+@pytest.fixture(scope="module")
+def commands_src() -> str:
+    return COMMANDS_JS_PATH.read_text(encoding="utf-8")
+
+
 # ── driver for _findModelInDropdown (ui.js) ─────────────────────────────────
 
 _FIND_DRIVER = r"""
@@ -302,12 +307,7 @@ class TestDidYouMeanToastAssembly:
     And no_model_match already ends with an opening quote, so cmdModel must close
     with `${args}"`, not `"${args}"` (which doubled the quote)."""
 
-    @pytest.fixture(scope="class")
-    def commands_src(self):
-        import pathlib
-        root = pathlib.Path(__file__).resolve().parents[1]
-        return (root / "static" / "commands.js").read_text(encoding="utf-8")
-
+    # Uses the module-scoped commands_src fixture above.
     def test_toast_passes_suggestion_as_t_argument(self, commands_src):
         assert "t('model_did_you_mean', suggestion)" in commands_src, (
             "cmdModel must pass the suggestion into t() so the template fills in "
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Read the full diff at `fix/macos-webui-test-reliability` — this is a clean test-only change (no runtime code touched, confirmed by the diff stat: only `tests/`). Two independent fixes bundled well:

## The macOS reliability fix is correct against the agent's compat contract

`test_get_profile_skills_stats` previously hard-coded `platforms=["macos"]` for the "gamma" skill and assumed it counts as *incompatible*. That assumption only holds on a Linux runner — on a macOS runner the same skill is compatible, so the expected `incompatible == 1` count breaks. The fix picks the platform the test is **not** running on:

```python
incompatible_platform = "linux" if sys.platform == "darwin" else "macos"
_write_skill(profile_home, "gamma", platforms=[incompatible_platform])
```

I traced this through the actual compat check the stats path calls — `get_profile_skills_stats` uses `skill_matches_platform` (`api/profiles.py:1752,1762`), which delegates to `agent/skill_utils.py:163` `skill_matches_platform_list`:

```python
current = sys.platform
for platform in platforms:
    mapped = PLATFORM_MAP.get(normalized, normalized)   # macos→darwin, linux→linux, windows→win32
    if current.startswith(mapped):
        return True
return False
```

So a foreign-OS tag always yields `False` (incompatible), which is exactly what the test needs. The mapping (`PLATFORM_MAP`, `skill_utils.py:21`) confirms `darwin`/`linux` are the two `sys.platform` values in play, so the `sys.platform == "darwin"` branch is the right discriminator. One benign edge worth noting: on a Termux/Android runner `sys.platform` isn't `"darwin"`, so it picks `"macos"` — still correctly incompatible, so the test stays green there too.

## The fixture-scope change is a straight refactor

Hoisting the per-class `commands_src` / `commands_js` / `messages_js` / etc. fixtures to `scope="module"` (`test_issue3368_*` and `test_issue467_*`) removes the class-scoped instance-fixture deprecation warnings without changing what's asserted — the fixtures are pure file reads with no shared mutable state, so module scope is safe and actually cheaper (one read instead of per-class). The assertions in each `test_*` body are untouched.

Given it's test-only, correct against the cross-repo compat contract, and the author reports a full green suite (`12836 passed`), this looks straightforward to land. Nice cleanup.

nesquena-hermes:
Shipped in **exp-v0.52.54**. Cross-platform test suite stabilized (per-OS incompatible-platform selection + state-free module fixtures, 8 deprecation warnings gone). Codex SAFE (assertions non-vacuous, 0 coverage removed, test-only), 66 tests pass incl. deprecation-as-error. Thanks @tahabakhit!

PR #5988, opened 2026-07-12
fix(webui): invalidate default terminal env on backend switch (#5937)

### LINKED ISSUE #5937: Local-backend WebUI sessions can execute tools on a previous remote terminal backend
# Local-backend WebUI sessions can execute tools on a previous remote terminal backend

## Summary

Hermes WebUI runs multi-profile agent turns in one long-lived process and applies each selected profile's terminal settings through process environment variables for the duration of a turn. However, WebUI does not ensure per-profile/per-backend isolation of Hermes Agent terminal/file environment caches.

In combination with Hermes Agent's collapsed `"default"` terminal environment cache, this allows a later local-backend profile session to reuse a remote/SSH environment created by an earlier remote-backend profile session in the same WebUI process.

**Upstream agent bug (primary):**  
https://github.com/NousResearch/hermes-agent/issues/62720  
`[Bug]: Local sessions can execute tools on a previous remote terminal backend`

This WebUI issue tracks the multi-profile defense-in-depth gap that makes #62720 reachable in normal WebUI use. Agent owns the core cache correctness fix; WebUI still needs isolation/invalidation so profile switches cannot inherit another profile's backend.

## Impact

- A WebUI session for a local-backend profile can execute shell/file tools on a remote host previously used by another profile.
- Profile switching appears correct at the identity/config layer while tool execution runs on the wrong backend.
- Users observe missing local paths, unexpected remote paths, and remote file-sync failures despite correct profile terminal configuration.
- The failure is intermittent and process-lifetime dependent, so it is easy to misattribute to workspace selection or profile config drift.

## Expected behavior

When WebUI starts an agent turn for a selected profile:

1. The turn should use that profile's terminal backend configuration.
2. Tool execution should not reuse another profile's remote terminal/file environment.
3. Profile switches that change backend (for example remote/SSH → local) should force isolation or cache invalidation so the new turn cannot inherit the previous backend.

## Actual behavior

1. Profile A (remote/SSH backend) runs a WebUI turn and creates a remote terminal environment in the shared agent process.
2. Profile B (local backend) runs a later WebUI turn in the same process.
3. WebUI applies Profile B's terminal env vars for the turn.
4. Hermes Agent still reuses the previously cached remote environment under the collapsed task key `"default"` because ordinary session task IDs are not treated as isolation keys.
5. Profile B tools execute on Profile A's remote backend.

## Why this is a WebUI issue as well as an agent issue

Hermes Agent owns the environment cache and is the primary correctness boundary (see #62720). WebUI still needs defense-in-depth because:

- WebUI intentionally hosts multiple profiles in one process.
- WebUI applies profile terminal settings via process-global environment mutation for each turn.
- WebUI passes session IDs into agent runs, but does not register isolation overrides that would prevent collapse to the shared `"default"` environment slot.
- WebUI currently has no forced invalidation of terminal/file environment caches on profile/backend change.

In short: WebUI creates the multi-profile concurrency surface that makes the agent cache hazard reachable in normal use.

### Profile terminal settings are mapped into process env for each turn

WebUI builds profile runtime env from the selected profile's terminal config (backend → `TERMINAL_ENV`, SSH fields → `TERMINAL_SSH_*`, etc.) and applies a filtered env update around the agent run.

This correctly points new environment creation at the selected profile's backend. It does not by itself guarantee that an already-cached incompatible environment will be discarded.

### Session task IDs are passed through, but not isolated for backend

WebUI agent runs pass a session-scoped task id into `run_conversation()`. In Hermes Agent, ordinary task IDs collapse to the shared `"default"` terminal/file environment key unless a task override includes isolation keys such as `env_type` (documented in #62720).

WebUI does not currently register such isolation overrides for normal profile sessions.

### No cache invalidation on backend/profile switch

On profile/backend change, WebUI does not explicitly clear Hermes Agent terminal/file environment caches (`_active_environments`, `_file_ops_cache`). Combined with collapsed keys, stale remote environments can survive into later local-backend turns.

## Reproduction sketch

In one WebUI process:

1. Select a profile configured with terminal backend = SSH/remote.
2. Run any terminal or file tool (creates remote environment in-process).
3. Without restarting WebUI, switch to a profile configured with terminal backend = local.
4. Run a host probe (`uname`, existence checks for expected local vs remote paths).
5. Observe remote-host execution despite the local-backend profile configuration.

Intermittency note: if the remote environment expires/is cleaned up before step 3, the bug may not appear.

## Suggested fix

WebUI-side defense in depth:

1. **Register per-session isolation overrides**
   - When starting a turn, register task env overrides that include `env_type` (and remote target identity when applicable) so the agent does not collapse all sessions into one environment slot.

2. **Invalidate stale environments on profile/backend change**
   - If selected profile backend/target differs from the previously used backend/target in this process, clear terminal/file environment caches before the turn starts.

3. **Include terminal backend identity in any session/agent runtime cache signatures**
   - Ensure cached agent/runtime objects cannot silently outlive a backend switch.

4. **Optional hard guard**
   - At turn start for local-backend profiles, run a cheap host/backend probe and fail closed if remote semantics are detected.

Hermes Agent still needs the primary fix tracked in #62720 (compatibility checks / stronger cache keys). WebUI should not rely solely on that.

## Acceptance criteria

- [ ] After a remote-backend profile turn, a subsequent local-backend profile turn in the same WebUI process executes locally.
- [ ] Profile switch from remote → local never reuses the previous remote terminal/file environment.
- [ ] Terminal and file tools agree on backend after profile switches.
- [ ] Regression coverage for multi-profile backend switching in one WebUI process.
- [ ] Clear logs when WebUI forces isolation or cache invalidation due to backend/profile change.

## Component

Hermes WebUI multi-profile streaming/runtime path (`api/streaming.py`, `api/profiles.py`, related session/agent orchestration).

jbbottoms:
## Summary

Fixes #5937 (WebUI defense-in-depth for multi-profile terminal backend switches).

When Hermes WebUI runs multiple profiles in one process, profile terminal settings are applied via process env each turn, but Hermes Agent collapses ordinary session task IDs to a shared `\"default\"` terminal/file environment cache slot. An earlier SSH/remote turn can therefore leave a remote environment that a later local-backend profile reuses.

This PR implements the maintainer-scoped WebUI-only path:

- Track a process-global **backend identity** (`TERMINAL_ENV` + SSH target + container images; **not** CWD).
- When the identity **changes**, call `cleanup_vm(\"default\")` so the next tool recreates against the correct env.
- Same-backend consecutive turns do **not** invalidate (preserves intentional cross-session local/container sharing).
- First turn in a process only records identity.

Agent-side #62720 remains the primary correctness fix for non-WebUI callers.

## Changes

- New `api/terminal_backend_isolation.py`
- Hook in `api/streaming.py` immediately after `os.environ.update(_safe_profile_runtime_env)`
- Unit tests in `tests/test_issue5937_terminal_backend_isolation.py`

## Test plan

- [x] Unit tests for identity / SSH→local / same-backend no-op / host change / cleanup errors (9 passed; Windows local run used `--noconftest` due to symlink privilege; Linux CI exercises full suite)
- [ ] CI green on this PR
- [ ] Manual (if available): remote-backend profile turn → local-backend profile turn in one process → local execution

## AI usage

Built with Grok 4.5 (Kai).

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -21841,6 +21841,25 @@ def _handle_chat_sync(handler, body):
         s.model = model
         s.model_provider = model_provider
     from api.streaming import _ENV_LOCK
+    from api.terminal_backend_isolation import (
+        TerminalBackendIsolationError,
+        acquire_turn_lease_failclosed,
+    )
+
+    # #5937/#5988 round 4: this fallback endpoint constructs a tool-capable
+    # AIAgent, so it must hold the same full-turn backend-identity lease as
+    # the streaming path — otherwise it can create/use the agent "default"
+    # env slot concurrently with (or across) a backend transition. This
+    # handler applies no profile runtime env overlay, so the effective turn
+    # environment is the process env itself (identity ignores TERMINAL_CWD,
+    # which is the only terminal key mutated below). Acquired BEFORE the
+    # _ENV_LOCK mutation, released in the turn's finally; rejections fail
+    # closed as a retryable 503 instead of running against an unverified
+    # slot.
+    try:
+        _terminal_backend_lease, _ = acquire_turn_lease_failclosed(dict(os.environ))
+    except TerminalBackendIsolationError as e:
+        return j(handler, {"error": str(e)}, status=503)
 
     with _ENV_LOCK:
         old_cwd = os.environ.get("TERMINAL_CWD")
@@ -21961,6 +21980,10 @@ def _handle_chat_sync(handler, body):
                 os.environ.pop("HERMES_SESSION_KEY", None)
             else:
                 os.environ["HERMES_SESSION_KEY"] = old_session_key
+        # Release AFTER the env restore: a waiter admitted by this release
+        # must never observe this turn's TERMINAL_* mutations. release() is
+        # idempotent (#5937).
+        _terminal_backend_lease.release()
     with _get_session_agent_lock(s.session_id):
         _result_messages = result.get("messages") or _previous_context_messages
         _next_context_messages = _restore_reasoning_metadata(
--- a/api/routes.py
+++ b/api/routes.py
@@ -21861,14 +21861,26 @@ def _handle_chat_sync(handler, body):
     except TerminalBackendIsolationError as e:
         return j(handler, {"error": str(e)}, status=503)
 
-    with _ENV_LOCK:
-        old_cwd = os.environ.get("TERMINAL_CWD")
-        os.environ["TERMINAL_CWD"] = str(workspace)
-        old_exec_ask = os.environ.get("HERMES_EXEC_ASK")
-        old_session_key = os.environ.get("HERMES_SESSION_KEY")
-        os.environ["HERMES_EXEC_ASK"] = "1"
-        os.environ["HERMES_SESSION_KEY"] = s.session_id
+    # #5988 round 5 (lease-leak): the env mutation, agent run, env restore, and
+    # lease release all live under ONE guaranteed teardown. The release is in
+    # the OUTER finally so it runs even if (a) the env MUTATION below raises
+    # before the run — previously it sat between the acquire and the protecting
+    # try, so a throw there leaked the lease with no backstop on this path — or
+    # (b) the env RESTORE raises during teardown. release() is idempotent, so
+    # this never double-frees another turn's count.
+    # Pre-bind so the restore in the finally can never NameError even if the
+    # mutation below raises before binding them.
+    old_cwd = old_exec_ask = old_session_key = None
     try:
+        with _ENV_LOCK:
+            # Capture ALL prior values before mutating any, so a set that
+            # raises still leaves an accurate restore snapshot.
+            old_cwd = os.environ.get("TERMINAL_CWD")
+            old_exec_ask = os.environ.get("HERMES_EXEC_ASK")
+            old_session_key = os.environ.get("HERMES_SESSION_KEY")
+            os.environ["TERMINAL_CWD"] = str(workspace)
+            os.environ["HERMES_EXEC_ASK"] = "1"
+            os.environ["HERMES_SESSION_KEY"] = s.session_id
         from run_agent import AIAgent
 
         with CHAT_LOCK:
@@ -21967,23 +21979,26 @@ def _handle_chat_sync(handler, body):
                 persist_user_message=msg,
             )
     finally:
-        with _ENV_LOCK:
-            if old_cwd is None:
-                os.environ.pop("TERMINAL_CWD", None)
-            else:
-                os.environ["TERMINAL_CWD"] = old_cwd
-            if old_exec_ask is None:
-                os.environ.pop("HERMES_EXEC_ASK", None)
-            else:
-                os.environ["HERMES_EXEC_ASK"] = old_exec_ask
-            if old_session_key is None:
-                os.environ.pop("HERMES_SESSION_KEY", None)
-            else:
-                os.environ["HERMES_SESSION_KEY"] = old_session_key
-        # Release AFTER the env restore: a waiter admitted by this release
-        # must never observe this turn's TERMINAL_* mutations. release() is
-        # idempotent (#5937).
-        _terminal_backend_lease.release()
+        try:
+            with _ENV_LOCK:
+                if old_cwd is None:
+                    os.environ.pop("TERMINAL_CWD", None)
+                else:
+                    os.environ["TERMINAL_CWD"] = old_cwd
+                if old_exec_ask is None:
+                    os.environ.pop("HERMES_EXEC_ASK", None)
+                else:
+                    os.environ["HERMES_EXEC_ASK"] = old_exec_ask
+                if old_session_key is None:
+                    os.environ.pop("HERMES_SESSION_KEY", None)
+                else:
+                    os.environ["HERMES_SESSION_KEY"] = old_session_key
+        finally:
+            # #5988 round 5 (lease-leak): NESTED in a finally so the release
+            # runs even if the env restore above raises. Ordered AFTER the
+            # restore so a waiter admitted by this release never observes this
+            # turn's TERMINAL_* mutations. release() is idempotent (#5937).
+            _terminal_backend_lease.release()
     with _get_session_agent_lock(s.session_id):
         _result_messages = result.get("messages") or _previous_context_messages
         _next_context_messages = _restore_reasoning_metadata(
--- a/api/routes.py
+++ b/api/routes.py
@@ -21849,15 +21849,40 @@ def _handle_chat_sync(handler, body):
     # #5937/#5988 round 4: this fallback endpoint constructs a tool-capable
     # AIAgent, so it must hold the same full-turn backend-identity lease as
     # the streaming path — otherwise it can create/use the agent "default"
-    # env slot concurrently with (or across) a backend transition. This
-    # handler applies no profile runtime env overlay, so the effective turn
-    # environment is the process env itself (identity ignores TERMINAL_CWD,
-    # which is the only terminal key mutated below). Acquired BEFORE the
-    # _ENV_LOCK mutation, released in the turn's finally; rejections fail
-    # closed as a retryable 503 instead of running against an unverified
-    # slot.
+    # env slot concurrently with (or across) a backend transition.
+    #
+    # #5988 P1 (greptile): the lease identity must come from THIS request's
+    # RESOLVED profile, not raw process-global os.environ. A concurrent
+    # streaming turn keeps its own profile's HERMES_HOME + terminal settings
+    # in os.environ for its whole run, so computing identity from the live
+    # process env would let a sync turn for a different profile calculate the
+    # streaming profile's identity, skip invalidation, and reuse its cached
+    # "default" backend. Resolve the session's profile and build the identity
+    # (and the applied turn env below) from the profile overlay, exactly like
+    # streaming — so identity reflects the backend this turn will actually use.
     try:
-        _terminal_backend_lease, _ = acquire_turn_lease_failclosed(dict(os.environ))
+        from api.profiles import (
+            filter_runtime_env_for_gateway_parity,
+            get_hermes_home_for_profile,
+            get_profile_runtime_env,
+        )
+        _chat_profile_home = str(get_hermes_home_for_profile(getattr(s, "profile", None)))
+        _chat_safe_profile_env = filter_runtime_env_for_gateway_parity(
+            get_profile_runtime_env(_chat_profile_home)
+        )
+    except ImportError:
+        _chat_profile_home = os.environ.get("HERMES_HOME", "")
+        _chat_safe_profile_env = {}
+
+    _effective_turn_env = dict(os.environ)
+    _effective_turn_env.update(_chat_safe_profile_env)
+    if _chat_profile_home:
+        _effective_turn_env["HERMES_HOME"] = _chat_profile_home
+    # Acquired BEFORE the _ENV_LOCK mutation (a differing-backend turn may
+    # block here waiting on in-flight turns; their env restore needs the lock).
+    # Released in the turn's finally; rejections fail closed as a retryable 503.
+    try:
+        _terminal_backend_lease, _ = acquire_turn_lease_failclosed(_effective_turn_env)
     except TerminalBackendIsolationError as e:
         return j(handler, {"error": str(e)}, status=503)
 
@@ -21870,14 +21895,23 @@ def _handle_chat_sync(handler, body):
     # this never double-frees another turn's count.
     # Pre-bind so the restore in the finally can never NameError even if the
     # mutation below raises before binding them.
-    old_cwd = old_exec_ask = old_session_key = None
+    old_cwd = old_exec_ask = old_session_key = old_hermes_home = None
+    _old_profile_env = {}
     try:
         with _ENV_LOCK:
             # Capture ALL prior values before mutating any, so a set that
             # raises still leaves an accurate restore snapshot.
+            _old_profile_env = {k: os.environ.get(k) for k in _chat_safe_profile_env}
             old_cwd = os.environ.get("TERMINAL_CWD")
             old_exec_ask = os.environ.get("HERMES_EXEC_ASK")
             old_session_key = os.environ.get("HERMES_SESSION_KEY")
+            old_hermes_home = os.environ.get("HERMES_HOME")
+            # Apply THIS request's resolved profile so the agent creates/uses
+            # the backend the lease identity names (#5988 P1) — not a
+            # concurrent streaming turn's leftover process env.
+            os.environ.update(_chat_safe_profile_env)
+            if _chat_profile_home:
+                os.environ["HERMES_HOME"] = _chat_profile_home
             os.environ["TERMINAL_CWD"] = str(workspace)
             os.environ["HERMES_EXEC_ASK"] = "1"
             os.environ["HERMES_SESSION_KEY"] = s.session_id
@@ -21981,6 +22015,15 @@ def _handle_chat_sync(handler, body):
     finally:
         try:
             with _ENV_LOCK:
+                for _pk, _pv in _old_profile_env.items():
+                    if _pv is None:
+                        os.environ.pop(_pk, None)
+                    else:
+                        os.environ[_pk] = _pv
+                if old_hermes_home is None:
+                    os.environ.pop("HERMES_HOME", None)
+                else:
+                    os.environ["HERMES_HOME"] = old_hermes_home
                 if old_cwd is None:
                     os.environ.pop("TERMINAL_CWD", None)
                 else:
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -6621,6 +6621,7 @@ def _run_agent_streaming(
     old_session_platform = None
     old_hermes_home = None
     old_profile_env = {}
+    _terminal_backend_lease = None
 
     # MCP discovery moved to AFTER the per-profile HERMES_HOME mutation below
     # (was here at v0.51.30) — the previous placement always read the default
@@ -7171,6 +7172,29 @@ def _run_agent_streaming(
         # block other concurrent sessions waiting on _ENV_LOCK (#2024).
         _prewarm_skill_tool_modules()
         _install_streaming_cronjob_profile_wrapper()
+        # #5937: multi-profile WebUI can leave an incompatible terminal env in
+        # the agent process-global "default" cache slot after a backend
+        # identity switch. Acquire a full-turn backend-identity lease BEFORE
+        # _ENV_LOCK (a differing-backend turn may block here waiting for
+        # in-flight turns, and their env restore needs _ENV_LOCK). Identity is
+        # derived from the effective turn environment — process env overlaid
+        # with this profile's runtime env, i.e. what os.environ.update below
+        # produces — not the profile env alone, so env applied by other layers
+        # is reflected. The lease is released in the turn's finally block.
+        try:
+            from api.terminal_backend_isolation import (
+                acquire_terminal_backend_turn_lease,
+            )
+            _effective_turn_env = dict(os.environ)
+            _effective_turn_env.update(_safe_profile_runtime_env)
+            _terminal_backend_lease, _ = acquire_terminal_backend_turn_lease(
+                _effective_turn_env
+            )
+        except Exception:
+            logger.warning(
+                "terminal backend isolation lease failed (#5937)",
+                exc_info=True,
+            )
         # Still set process-level env as fallback for tools that bypass thread-local
         # Acquire lock only for the env mutation, then release before the agent runs.
         # The finally block re-acquires to restore — keeping critical sections short
@@ -7185,20 +7209,9 @@ def _run_agent_streaming(
             old_session_chat_id = os.environ.get('HERMES_SESSION_CHAT_ID')
             old_hermes_home = os.environ.get('HERMES_HOME')
             os.environ.update(_safe_profile_runtime_env)
-            # #5937: multi-profile WebUI can leave a remote terminal env in the
-            # agent process-global "default" cache slot after SSH→local (or any
-            # backend identity) switch. Scoped invalidation drops only that slot
-            # when identity changes; same-backend turns still share the env.
-            try:
-                from api.terminal_backend_isolation import (
-                    maybe_invalidate_default_terminal_env,
-                )
-                maybe_invalidate_default_terminal_env(_safe_profile_runtime_env)
-            except Exception:
-                logger.warning(
-                    "terminal backend isolation check failed (#5937)",
-                    exc_info=True,
-                )
+            # #5937: the backend-identity lease acquired above (before this
+            # lock) already invalidated the agent "default" terminal env slot
+            # if this turn's identity differs from the last committed one.
             os.environ['TERMINAL_CWD'] = str(s.workspace)
             os.environ['HERMES_EXEC_ASK'] = '1'
             os.environ['HERMES_SESSION_KEY'] = session_id
@@ -9996,6 +10009,17 @@ def _run_agent_streaming(
                 else: os.environ['HERMES_SESSION_CHAT_ID'] = old_session_chat_id
                 if old_hermes_home is None: os.environ.pop('HERMES_HOME', None)
                 else: os.environ['HERMES_HOME'] = old_hermes_home
+            # #5937: release this turn's backend-identity lease AFTER the env
+            # restore so a waiting differing-backend turn can't invalidate
+            # while this turn's environment is still nominally in effect.
+            if _terminal_backend_lease is not None:
+                try:
+                    _terminal_backend_lease.release()
+                except Exception:
+                    logger.debug(
+                        "terminal backend lease release failed (#5937)",
+                        exc_info=True,
+                    )
 
     except Exception as e:
         print('[webui] stream error:\n' + traceback.format_exc(), flush=True)
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -7185,6 +7185,20 @@ def _run_agent_streaming(
             old_session_chat_id = os.environ.get('HERMES_SESSION_CHAT_ID')
             old_hermes_home = os.environ.get('HERMES_HOME')
             os.environ.update(_safe_profile_runtime_env)
+            # #5937: multi-profile WebUI can leave a remote terminal env in the
+            # agent process-global "default" cache slot after SSH→local (or any
+            # backend identity) switch. Scoped invalidation drops only that slot
+            # when identity changes; same-backend turns still share the env.
+            try:
+                from api.terminal_backend_isolation import (
+                    maybe_invalidate_default_terminal_env,
+                )
+                maybe_invalidate_default_terminal_env(_safe_profile_runtime_env)
+            except Exception:
+                logger.warning(
+                    "terminal backend isolation check failed (#5937)",
+                    exc_info=True,
+                )
             os.environ['TERMINAL_CWD'] = str(s.workspace)
             os.environ['HERMES_EXEC_ASK'] = '1'
             os.environ['HERMES_SESSION_KEY'] = session_id
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -7195,6 +7195,15 @@ def _run_agent_streaming(
 
         _effective_turn_env = dict(os.environ)
         _effective_turn_env.update(_safe_profile_runtime_env)
+        # #5988 round 5: the identity must reflect THIS turn's profile, not
+        # whatever HERMES_HOME the shared process env currently carries (a
+        # prior turn's, applied to os.environ only later at the mutation
+        # below). Stamp the RESOLVED profile home so two profiles can never
+        # share a backend identity. The forwarded-secret VALUES are already
+        # present via _safe_profile_runtime_env (the profile's .env), which
+        # terminal_backend_identity() now fingerprints.
+        if _profile_home:
+            _effective_turn_env['HERMES_HOME'] = _profile_home
         _terminal_backend_lease, _ = acquire_turn_lease_failclosed(
             _effective_turn_env
         )
@@ -9994,35 +10003,41 @@ def _run_agent_streaming(
                     _unreg_clarify_notify(session_id)
                 except Exception:
                     logger.debug("Failed to unregister clarify callback")
-            with _ENV_LOCK:
-                for _key, _old_value in old_profile_env.items():
-                    if _old_value is None: os.environ.pop(_key, None)
-                    else: os.environ[_key] = _old_value
-                if old_cwd is None: os.environ.pop('TERMINAL_CWD', None)
-                else: os.environ['TERMINAL_CWD'] = old_cwd
-                if old_exec_ask is None: os.environ.pop('HERMES_EXEC_ASK', None)
-                else: os.environ['HERMES_EXEC_ASK'] = old_exec_ask
-                if old_session_key is None: os.environ.pop('HERMES_SESSION_KEY', None)
-                else: os.environ['HERMES_SESSION_KEY'] = old_session_key
-                if old_session_id is None: os.environ.pop('HERMES_SESSION_ID', None)
-                else: os.environ['HERMES_SESSION_ID'] = old_session_id
-                if old_session_platform is None: os.environ.pop('HERMES_SESSION_PLATFORM', None)
-                else: os.environ['HERMES_SESSION_PLATFORM'] = old_session_platform
-                if old_session_chat_id is None: os.environ.pop('HERMES_SESSION_CHAT_ID', None)
-                else: os.environ['HERMES_SESSION_CHAT_ID'] = old_session_chat_id
-                if old_hermes_home is None: os.environ.pop('HERMES_HOME', None)
-                else: os.environ['HERMES_HOME'] = old_hermes_home
-            # #5937: release this turn's backend-identity lease AFTER the env
-            # restore so a waiting differing-backend turn can't invalidate
-            # while this turn's environment is still nominally in effect.
-            if _terminal_backend_lease is not None:
-                try:
-                    _terminal_backend_lease.release()
-                except Exception:
-                    logger.debug(
-                        "terminal backend lease release failed (#5937)",
-                        exc_info=True,
-                    )
+            try:
+                with _ENV_LOCK:
+                    for _key, _old_value in old_profile_env.items():
+                        if _old_value is None: os.environ.pop(_key, None)
+                        else: os.environ[_key] = _old_value
+                    if old_cwd is None: os.environ.pop('TERMINAL_CWD', None)
+                    else: os.environ['TERMINAL_CWD'] = old_cwd
+                    if old_exec_ask is None: os.environ.pop('HERMES_EXEC_ASK', None)
+                    else: os.environ['HERMES_EXEC_ASK'] = old_exec_ask
+                    if old_session_key is None: os.environ.pop('HERMES_SESSION_KEY', None)
+                    else: os.environ['HERMES_SESSION_KEY'] = old_session_key
+                    if old_session_id is None: os.environ.pop('HERMES_SESSION_ID', None)
+                    else: os.environ['HERMES_SESSION_ID'] = old_session_id
+                    if old_session_platform is None: os.environ.pop('HERMES_SESSION_PLATFORM', None)
+                    else: os.environ['HERMES_SESSION_PLATFORM'] = old_session_platform
+                    if old_session_chat_id is None: os.environ.pop('HERMES_SESSION_CHAT_ID', None)
+                    else: os.environ['HERMES_SESSION_CHAT_ID'] = old_session_chat_id
+                    if old_hermes_home is None: os.environ.pop('HERMES_HOME', None)
+                    else: os.environ['HERMES_HOME'] = old_hermes_home
+            finally:
+                # #5937: release this turn's backend-identity lease AFTER the env
+                # restore so a waiting differing-backend turn can't invalidate
+                # while this turn's environment is still nominally in effect.
+                # #5988 round 5 (lease-leak): NESTED in a finally so the ordered
+                # primary release runs even if the env restore above raises —
+                # the outer-finally backstop is then the second line of defense,
+                # not the only path that frees the lease.
+                if _terminal_backend_lease is not None:
+                    try:
+                        _terminal_backend_lease.release()
+                    except Exception:
+                        logger.debug(
+                            "terminal backend lease release failed (#5937)",
+                            exc_info=True,
+                        )
 
     except Exception as e:
         print('[webui] stream error:\n' + traceback.format_exc(), flush=True)
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -7183,6 +7183,7 @@ def _run_agent_streaming(
         # is reflected. The lease is released in the turn's finally block.
         try:
             from api.terminal_backend_isolation import (
+                TerminalBackendIsolationError,
                 acquire_terminal_backend_turn_lease,
             )
             _effective_turn_env = dict(os.environ)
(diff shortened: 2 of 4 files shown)


nesquena-hermes (changes requested):
## 🔬 Gate review (Codex, security/isolation) — SHIP ONLY WITH FIXES (Docker path + concurrency leave the leak partly open)

Thanks @jbbottoms — the core mechanism is right and the nominal path is verified: a successful SSH→local transition removes the real agent's `"default"` slot, clears the remote object, and the next terminal command creates a `LocalEnvironment`; same-identity is a correct no-op; cleanup exceptions don't crash streaming. 14 isolation + 8 concurrency/lock tests pass. But the gate found the leak isn't reliably closed for release — 5 issues, one CORE:

1. **[CORE — Docker leak still open] `api/terminal_backend_isolation.py:107`** — `cleanup_vm("default")` defaults to `force_remove=False`. For persistent Docker, cleanup leaves the container running and recreation reattaches **by labels without comparing image**, so **Docker-A → local → Docker-B can silently reuse image A**. Force-remove the outgoing Docker environment on a container-identity transition (or add agent-side persisted-container image-compatibility checks) so a stale container can't be reattached after a container identity change.

2. **[identity incomplete] `api/terminal_backend_isolation.py:39`** — make the identity fully backend-specific: include the **SSH-key fingerprint** and **Modal mode**, normalize defaults, and exclude inactive settings and CWD. As-is two genuinely-different backends can hash equal and skip invalidation.

3. **[identity source] `api/streaming.py:7196`** — derive identity from the **effective post-`os.environ.update` environment**, not `_safe_profile_runtime_env` alone (otherwise env applied by other layers isn't reflected in the identity).

4. **[concurrency — evicts an in-use env] `api/streaming.py:7192`** — with overlapping turns, a differing-backend turn can `cleanup_vm("default")` an environment **still in use** by another turn. Acquire a full-turn backend-identity lease so a differing-backend turn can't evict an env that's still active.

5. **[retry state lost] `api/terminal_backend_isolation.py:89`** — preserve retry state when import or cleanup fails (so a transient failure doesn't permanently skip a needed invalidation); add a failure-then-success regression test.

The SSH→local case is solid — these are the Docker-persistent path, the identity completeness, and the overlapping-turn concurrency. Since this is a cross-profile isolation/security boundary, all five (esp. #1 and #4) should land before merge. Re-gate once the Docker force-remove + full identity + turn lease are in. Good, well-scoped fix otherwise.

jbbottoms:
## Ready for re-gate (52011a19) — all five findings addressed

| # | Finding | Fix |
|---|---|---|
| 1 | **CORE — Docker reattach-by-label leak** | Transitions *away from* a Docker identity now call `cleanup_vm('default', force_remove=True)` (agent signature verified: `cleanup_vm(task_id, *, force_remove=False)` forwards to backends whose `cleanup()` accepts it). Non-container transitions keep the plain call. |
| 2 | **Identity incomplete** | `terminal_backend_identity` is now backend-specific: only the active backend's settings participate (a leftover `TERMINAL_SSH_HOST` no longer perturbs a `local` identity). SSH identity adds `TERMINAL_SSH_KEY`; Modal adds `TERMINAL_MODAL_MODE` (case-normalized). Absent keys normalize to the agent defaults mirrored from `tools/terminal_tool.py` — port `22`, `nikolaik/python-nodejs:python3.11-nodejs20` image family, modal `auto` — so explicit-default and absent hash equal. |
| 3 | **Identity source** | Streaming derives identity from the *effective* turn environment — `dict(os.environ)` overlaid with `_safe_profile_runtime_env`, i.e. exactly what `os.environ.update` produces — not the profile env alone. |
| 4 | **Concurrency — evicts in-use env** | Turns now hold a **full-turn backend-identity lease** (`acquire_terminal_backend_turn_lease` → `lease.release()` in the turn's `finally`). A differing-backend turn waits for in-flight turns on other identities before invalidating; on bounded timeout (30s) it proceeds **without** invalidation (fail-safe: never evict an in-use env; transition stays uncommitted → retried later). Lease acquisition is deliberately *before* `_ENV_LOCK` and release *after* the env restore — a waiter can't deadlock the restore path. Concurrent same-identity turns share one invalidation (no self-eviction of a just-created env). |
| 5 | **Retry state lost** | The identity transition commits only after `cleanup_vm` succeeds. Import failure or cleanup exception leaves `_last_backend_identity` uncommitted, so the next turn retries. Failure-then-success regression added (plus an import-failure variant). |

**Tests:** `test_issue5937_terminal_backend_isolation.py` 20/20 (11 new: force-remove on docker→local and docker-image change, non-docker keeps plain cleanup, ssh-key/modal-mode identity, default normalization, inactive-setting exclusion, failure-then-success retry ×2, real-thread lease wait, timeout-skip-then-retry, same-identity piggyback). Source guard updated to assert merged-env identity, pre-`_ENV_LOCK` acquisition, and finally-release. Broader `-k "terminal or backend or runtime_env or profile"` selection: **1029 passed / 40 skipped** on Linux.

nesquena-hermes:
## 🔬 Re-gate (Codex) — all five findings verified against source; PASS with one narrow residual note

Re-pulled `52011a19d` and read the full diff (`api/terminal_backend_isolation.py`, the `api/streaming.py` hook, and the 20-case test file) against `origin/master`, plus the agent side to confirm the contract claims. The five findings from the prior gate are genuinely resolved, not just described.

**Finding #1 (Docker force-remove) — verified at the contract boundary.** The PR's claim that `cleanup_vm` accepts `force_remove` is accurate. Agent `tools/terminal_tool.py:1694`:

```python
def cleanup_vm(task_id: str, *, force_remove: bool = False):
    ...
    sig = inspect.signature(env.cleanup)
    if "force_remove" in sig.parameters:
        env.cleanup(force_remove=force_remove)
    else:
        env.cleanup()
```

So `_invalidate_default_env` calling `cleanup_vm("default", force_remove=True)` for `previous[0] in _FORCE_REMOVE_ENV_TYPES` (`isolation.py:171-176`) is correct — non-Docker backends still get the plain call, and the signature-guard means a stale agent that predates the kwarg degrades gracefully rather than raising.

**Findings #2/#3 (identity) — verified.** `terminal_backend_identity` (`isolation.py:66-112`) is now backend-specific and the mirrored defaults match agent source: `_AGENT_DEFAULT_IMAGE` = `nikolaik/python-nodejs:python3.11-nodejs20` (agent `terminal_tool.py:1259`), modal `auto` (`:1327`), SSH port `22`. So absent-key and explicit-default hash equal, and a leftover `TERMINAL_SSH_HOST` no longer perturbs a `local` identity. The streaming hook derives identity from `dict(os.environ)` overlaid with `_safe_profile_runtime_env` — the actual post-merge turn env.

**Finding #4 (concurrency) — verified, lease ordering is correct.** Lease acquired before `_ENV_LOCK` and released after the env restore in `finally` (`streaming.py:7189-7197`, `:10012-10021`), so a waiter can't deadlock the restore path. Timeout is fail-safe (proceed WITHOUT invalidation, leave uncommitted).

### One residual window worth a note (non-blocking)

The piggyback branch in `acquire_terminal_backend_turn_lease`:

```python
if _active_turn_counts.get(identity, 0) > 1:
    # Another turn on this same new identity is already in flight ...
    return lease, False
```

`test_concurrent_same_identity_turns_share_one_invalidation` is serialized single-thread, so by the time the "second" turn acquires, the leader has already committed `_last_backend_identity = local` — it actually hits the `previous == identity` early return, not this `> 1` branch. The genuine `> 1` path only fires when a second same-new-identity turn takes `_COND` *after* the leader incremented the count but *before* the leader commits (i.e. while the leader is mid-`cleanup_vm` outside the lock).

In that true-concurrent window, the piggybacker returns `False` assuming the leader invalidated — but if the leader's `_invalidate_default_env` then *fails*, `_last_backend_identity` stays uncommitted (good, next turn retries) yet the piggybacker already proceeded to run its turn against a slot that was never invalidated. Narrow: needs two concurrent turns of the same new identity racing plus a cleanup failure, and it self-heals on the following turn. It's strictly no worse than master's behavior, so not a blocker — but a real-thread test that holds the leader inside a failing `cleanup_vm` while a same-identity piggybacker acquires would close the coverage gap and document the intended semantics.

Two smaller confirm-items: `_FORCE_REMOVE_ENV_TYPES` is `{"docker"}` only — if `singularity`/`daytona` also reattach containers by label rather than by image compare, they'd share the Docker-A→local→Docker-B leak; worth a one-line confirmation that those backends recreate from scratch. Otherwise this is solid — PASS from my side, pending your merge call.

jbbottoms:
Thanks for the PASS — and for catching that my piggyback test was serialized. Both follow-ups addressed:

**Residual window → covered (533be7ac).** New `test_piggyback_during_failing_leader_cleanup_stays_uncommitted` exercises the genuine `> 1` branch: the leader is held *inside* a failing `cleanup_vm` on its own thread while a same-new-identity turn acquires. Verified semantics, now documented in-test: the piggybacker returns without invalidating (one `cleanup_vm` call total — never evicts what the leader may have built), the leader's failure leaves `_last_backend_identity` uncommitted, and the next unobstructed turn retries successfully. That matches your read of the window — the piggybacker can run one turn against a never-invalidated slot only when the race AND a cleanup failure coincide, and it self-heals on the following turn (strictly no worse than master, as you said). 21/21 on the file.

**Singularity/daytona confirm-item — checked against agent source** (`tools/environments/`):
- **Singularity: clean, correctly excluded.** `_start_instance` always runs `singularity instance start … <image> <instance_id>` — every environment creation launches fresh from the configured `.sif`. Persistence is a `--overlay` dir (filesystem contents only); an image change takes effect on the next create. No reattach-by-label path, so no Docker-style image staleness — the overlay surviving is intended persistence, not a leak.
- **Daytona: shares the reattach behavior, but the fix cannot land on this side of the contract.** Persistent mode resumes by sandbox *name* (`self._daytona.get(f"hermes-{task_id}")`, legacy fallback by `hermes_task_id` label) with **no image comparison** — Daytona-A → local → Daytona-B would resume the image-A sandbox, same shape as the Docker CORE finding. However `DaytonaEnvironment.cleanup(self)` takes no `force_remove` parameter, so `cleanup_vm`'s signature-guard would silently drop the kwarg — adding `"daytona"` to `_FORCE_REMOVE_ENV_TYPES` here would be a no-op today and would misdocument the guarantee. This needs an agent-side change (`force_remove` support in daytona's `cleanup`, i.e. delete instead of stop on force); I'll file it upstream on hermes-agent and link back. Once the agent contract supports it, adding daytona to the set here is a one-line follow-up.

(Docker's `cleanup(self, *, force_remove: bool = False)` remains the only backend that both reattaches by label and accepts the kwarg — `_FORCE_REMOVE_ENV_TYPES = {"docker"}` is accurate to the current contract.)

nesquena-hermes (changes requested):
## 🔬 Re-gate round 2 (Codex, security-boundary) — KICK BACK: isolation still admits incompatible turns; tests assert the unsafe behavior

Thanks @jbbottoms — real progress: SSH-key + Modal mode are now in the identity, Docker uses `force_remove=True`, and the turn lease IS released in a `finally` (streaming.py:10015-10022, no leak on the normal/exception paths — good, the deadlock risk I flagged is handled). But re-gating against the **real agent cleanup contract** (not mocks) shows the core isolation guarantee is still false in two ways, and the submitted tests unfortunately encode the unsafe behavior as expected:

1. **[CORE — admits an incompatible turn on timeout / during cleanup] `api/terminal_backend_isolation.py` (timeout + piggyback paths, ~L257-261)** — after the 30s lease-wait times out, the code **registers and returns the differing identity while the old identity is still active**. Reproduced: both `('local',)` and `('ssh',…)` active simultaneously — exactly the cross-backend reuse this PR is meant to prevent. Separately, a same-new-identity caller is admitted **while the leader is still inside `cleanup_vm`**, including when that cleanup later fails. **Fix:** on timeout, abort/defer the incoming turn **without** registering a lease (never proceed uninvalidated); same-new-identity callers must wait on explicit transition state until cleanup **succeeds**; if cleanup fails, no turn may run against the old slot.

2. **[CORE — false success when cleanup fails] `api/terminal_backend_isolation.py:172`** — the real agent `cleanup_vm()` **catches backend-cleanup exceptions and returns `None`**, so this code treats it as success and commits the new `_last_backend_identity`. Reproduced: forced a backend cleanup failure → `isolation_reported_invalidated=True` and the new identity was committed anyway, leaving the stale backend reusable. **Fix:** commit `_last_backend_identity` only after a **verified-success** cleanup — use (or add) a cleanup API that returns verified success or raises, and coordinate a synchronous, failure-reporting forced-cleanup contract with hermes-agent. Don't infer success from `cleanup_vm` returning.

**Test note (important):** the 21 tests pass but **mock `cleanup_vm` with raising `MagicMock`s** (hiding the real return-`None`-on-failure contract) and **explicitly assert the timeout/piggyback "proceed without invalidation" path** — so green CI here is masking the gaps above rather than proving safety. Please retest against the real cleanup semantics (returns `None` on internal failure) and assert that an incompatible turn is NEVER admitted on timeout or on cleanup failure.

This is a cross-profile security boundary (a wrong reuse runs tools on another profile's backend), so it needs the transition coordinator redesigned so no incompatible turn is admitted on timeout or during/after a failed cleanup, plus a verified-success cleanup contract. The identity-completeness + lease-release half is solid — it's the transition-safety + failure-handling that must close before merge.

jbbottoms:
Ready for re-gate round 3 (eebb818e) — both CORE findings addressed by redesigning the transition coordinator to fail closed, and the tests now model the real cleanup contract instead of mocking around it.

| # | Finding | Fix |
|---|---------|-----|
| 1 | **CORE — admits an incompatible turn on timeout / during cleanup** | Admission is now fail-closed. On lease-wait timeout the incoming turn is **rejected** with `TerminalBackendTransitionTimeout` — no lease is registered, nothing runs uninvalidated (the previous `proceed WITHOUT invalidation` branch is gone). Transitions run under explicit state (`_transition_identity`): while a leader's cleanup+verification is in flight **no turn is admitted at all** — same-new-identity arrivals wait on the condition and re-evaluate after the transition; if the leader failed they retry the invalidation themselves rather than assuming it happened (the old `>1` piggyback branch is gone). `streaming.py` re-raises the typed rejections so the turn aborts with a user-readable error through the standard stream error path. |
| 2 | **CORE — false success when cleanup fails** | Commit now requires **observed postconditions**, never `cleanup_vm` returning (verified against agent source: `cleanup_vm` pops the slot first, swallows teardown exceptions, returns `None`; and `DockerEnvironment.cleanup(force_remove=True)` runs `docker stop`/`docker rm -f` on a background daemon thread — returning proves nothing even without an exception). A transition commits only when (a) `get_active_env("default")` is verifiably `None` post-cleanup, and (b) every outgoing container is verifiably gone: the outgoing container id is **ledgered before** cleanup (the slot pop is unrepeatable, so this is the only way a failed removal survives into a retry), then a `docker inspect` probe waits (bounded, joining the agent's async cleanup thread when available) and falls back to one direct forced removal before re-probing. Unverifiable ⇒ uncommitted ⇒ `TerminalBackendInvalidationFailed` ⇒ the turn does not run; the ledger drives the next attempt. |

**Tests retested against the real cleanup semantics** (the note in your review was fair — the raising MagicMocks were hiding the contract). The file now uses a `FakeAgent`/`FakeDockerEnv` pair that mirrors hermes-agent exactly: pop-first, swallow-exceptions, **return-None-on-failure** `cleanup_vm`, and Docker removal that can silently fail or complete asynchronously after `cleanup_vm` returns. New assertions include: timeout ⇒ raises with **no lease registered** (`_active_turn_counts` unchanged); silent docker-rm failure ⇒ raises, identity uncommitted, container ledgered, next turn closes the leak via the ledger; swallowed backend exception ⇒ same; async removal ⇒ awaited, not assumed; waiters (same-new-identity AND previous-identity) blocked during an in-flight transition; leader-failure ⇒ waiter performs its own verified invalidation; contract-drift guard (slot not popped ⇒ fail closed). 24/24 on the file; broader `-k "terminal or backend or runtime_env or profile"`: **1087 passed / 25 skipped** on Linux against the agent venv.

Three judgment calls to confirm at the gate:

1. **Direct removal fallback from WebUI**: when the agent's async removal verifiably didn't happen, the verifier attempts one direct `docker rm -f <id>` itself before failing. The container belongs to the outgoing identity and force-removal was already the decided policy for the transition, so I judged this in-bounds — say the word if you'd rather it fail without the fallback.
2. **Unexpected-bug posture in streaming**: deliberate typed rejections abort the turn (fail closed); an *unexpected* exception escaping the isolation module still logs-and-proceeds so a bug in this defense-in-depth layer can't brick every chat turn. Deliberate vs unexpected is the line I drew; happy to flip the residual open path if you want strict fail-closed there too.
3. **Agent-side contract**: a synchronous, failure-reporting forced-cleanup API remains the clean end state (probe layer here shrinks to one verified call) — I'll propose it on hermes-agent alongside #63361 (daytona `force_remove`) if you agree it's worth carrying upstream.

nesquena-hermes (changes requested):
## 🔬 Re-gate round 3 (Codex, security boundary, real cleanup contract) — KICK BACK: an incompatible turn can still run against the wrong backend

Thanks for the continued iteration @jbbottoms — but re-gating against the **real** `cleanup_vm` contract (not mocks), all 5 round-2 findings are still reproducing, and the bottom line is unchanged: **an incompatible turn can still execute against the wrong backend**, through both the timeout and same-new-identity cleanup windows. This needs a structural redesign rather than another incremental patch, so I'm kicking it back with the shape of the fix:

**Reproduced (still failing):**
1. **Timeout still admits** — after the lease-wait timeout, the probe showed **both SSH and local identities active simultaneously**. The "proceeding WITHOUT invalidation" path registers the differing identity and lets the turn run against the stale backend. It must **abort/defer** the incoming turn (no runnable lease) instead.
2. **Commit on unverified cleanup** — the real agent `cleanup_vm()` **catches backend exceptions and returns `None`**; the probe forced a return-`None` failure and the new identity was **still committed** (stale backend marked clean + reusable).
3. **Same-new-identity piggyback** — admitted while the cached backend was still SSH (runs before cleanup succeeds).
4. **Lease leak** — normal-path release works, but an exception in the window **before** the inner `try/finally` leaks the lease (then a permanent 30s stall on every later differing-backend turn).
5. **Tests encode the unsafe behavior** — 21 pass, but `tests/test_issue5937_terminal_backend_isolation.py:255` models failure by **raising** (not the real return-`None`), `:322` explicitly asserts timeout **admission**, and `:363` asserts the unsafe piggyback. So green CI is validating the wrong contract.

**Required redesign (fail-closed transition state machine):**
- `api/terminal_backend_isolation.py:228-267` — no runnable lease is granted before cleanup **succeeds**; on timeout **or** failed cleanup, **abort the incoming turn** (do not run tools against the old slot); same-new-identity callers **await the transition result** and only run after verified success.
- `api/terminal_backend_isolation.py:263-266` — establish a **verified real-agent cleanup contract** before committing `_last_backend_identity` (the current `None`/caught-error/async contract cannot *prove* cleanup happened; need an API that returns verified success or raises).
- `api/streaming.py:10323` — move lease release into the **outermost** guaranteed teardown `finally` (after environment restoration), or start the protected `try/finally` immediately after acquisition, so no exception window can leak it.
- Rewrite the tests to model the **real** `cleanup_vm` (returns `None` on internal failure) and assert that an incompatible turn is **never** admitted on timeout or during/after a failed cleanup.

The identity-completeness (SSH + Modal), Docker force-remove, and normal-path lease release are all good — it's the transition **safety** (admit-then-hope → block-until-safe) + the verified-cleanup contract that must change. Given this likely needs an agent-side cleanup API to return verified success (cross-repo), flag me if you'd like a hand scoping that half. Re-gate once the coordinator is fail-closed and the tests model the real cleanup contract.

jbbottoms:
Follow-up (f3942b84): the first push of this round went red on CI shard 3 (`test_tool_limit_terminal_state.py::test_streaming_tool_limit_with_fallback_final_response_surfaces_closure_text`, all three Python versions) — root-caused and fixed, and worth documenting because it's a semantic point, not a flake:

**Root cause.** CI runs the webui suite **without hermes-agent installed**, so `tools.terminal_tool` is unimportable there. The fail-closed redesign treated "cannot resolve cleanup_vm" as a failed invalidation → every backend-identity transition rejected the turn → the tool-limit streaming test (which happened to follow an isolation test that left a docker-committed identity in its shard) aborted before its `done` payload. Reproduced exactly on Linux in an agent-less venv (`ModuleNotFoundError: No module named 'tools'` → rejected turn) before fixing.

**The distinction f3942b84 draws:**
- **Agent runtime absent** (`ImportError` on `tools.terminal_tool`): the `"default"` env cache *lives in that module* — if it can't be imported, no cached environment can exist in this process, so there is nothing to isolate and the transition commits **vacuously**. That's correctness, not fail-open.
- **Runtime present but helpers unresolvable** (module imports but `cleanup_vm`/`get_active_env` missing): still **fails closed** — the cache may exist and we can't verify it. For older agent builds that predate `get_active_env`, the verifier falls back to the underlying `_active_environments` dict, which has existed as long as the slot mechanism itself.

Both cases have dedicated tests, the isolation tests now also reset module state in `teardown_function` for shard hygiene, and the CI-shaped case is covered by running the previously-failing sequence in an agent-less venv. Receipts on f3942b84: isolation file 25/25 (agent-less AND agent-present venvs), the exact failing CI sequence green agent-less, broader `-k "terminal or backend or runtime_env or profile"`: 1088 passed / 25 skipped against the agent venv.

jbbottoms:
Thanks for the round-3 pass — I think we crossed in flight: the structural redesign you're describing landed in eebb818e (19:47Z) / f3942b84 (19:56Z), pushed minutes before your review posted (19:56:50Z), and the review's specifics match the pre-push tree rather than the current head. Concretely, at eebb818e:

- the `"proceeding WITHOUT invalidation"` path doesn't exist (`git grep` on the head returns nothing — timeout now raises `TerminalBackendTransitionTimeout` **before** registering any lease);
- the test file has 24 tests (now 26), not 21, and the cited assertions are gone: there is no timeout-admission assertion or piggyback assertion — their replacements assert the opposite (`test_lease_wait_timeout_rejects_turn_without_admitting_it` checks `_active_turn_counts` still holds only the previous identity; `test_no_turn_is_admitted_while_transition_cleanup_is_in_flight` holds the leader inside cleanup and asserts arrivals block);
- `api/terminal_backend_isolation.py:228-267` on the head is the container-removal probe, not the admission loop.

The current head implements exactly the fail-closed state machine your "required redesign" describes — details in my two comments above (19:52Z and 20:01Z), briefly: no runnable lease before verified cleanup; timeout/failed cleanup abort the incoming turn via typed errors that streaming re-raises; same-new-identity callers wait on explicit transition state and retry the invalidation themselves if the leader failed; commit requires observed postconditions (slot verifiably empty via `get_active_env` + ledgered outgoing containers verifiably gone via `docker inspect` with a bounded wait and one direct removal fallback) rather than `cleanup_vm` returning; and the tests were rewritten around a fake that models the real contract — pop-first, swallow-exceptions, **return-None-on-failure**, async Docker removal.

**One of your five findings was still live against the new head, and it's now fixed too — thank you for it.** Finding #4 (lease leak): the lease is acquired ~180 lines before the inner `try` whose `finally` performs the ordered after-env-restore release, so an exception in that window leaked it — and under fail-closed admission a leaked lease is a *standing denial* (every later differing-backend turn waits its bound, then aborts), strictly worse than the old 30s stall. 20b30b81 adds an idempotent **backstop release in the outermost guaranteed-teardown finally** (primary release stays put to keep the after-env-restore ordering; `release()` is `_released`-guarded so the normal path's double release is a no-op). Source guard now requires both release sites; a new unit test asserts double-release never underflows a concurrent turn's count.

Also heads-up on one CI-semantics point from this round (details in the 20:01Z comment): webui CI runs without hermes-agent, and since the `"default"` cache lives in `tools.terminal_tool`, an unimportable agent runtime means the cache cannot exist — that transition commits **vacuously** (a present-but-broken runtime still fails closed). Flagging it since it's a deliberate carve-out in the fail-closed story.

And yes please on your offer — I'd value a hand scoping the agent-side half: a synchronous, failure-reporting forced-cleanup API (verified success or raise) would let the WebUI probe layer shrink to one call, and it pairs naturally with NousResearch/hermes-agent#63361 (daytona `cleanup()` lacks `force_remove`). Happy to draft the agent-side issue with the contract sketch if that's useful.

Re-gate target: **20b30b81**. Receipts on this head: isolation file 26/26 (Windows, agent-less Linux venv, agent venv); the previously-failing CI shard sequence green in an agent-less venv; broad `-k "terminal or backend or runtime_env or profile"`: 1088 passed / 25 skipped; all 18 checks green on f3942b84 (CI on 20b30b81 running now).

nesquena-hermes (changes requested):
## 🔬 Re-gate round 4 (Codex, security boundary, real cleanup contract) — KICK BACK: still fails OPEN on unexpected/Docker-probe failure

Thanks @jbbottoms — the fail-closed state machine is a real improvement (timeout now raises `TerminalBackendTransitionTimeout`, in-flight cleanup holds other turns, verified-postcondition structure is there). But re-gating against the **real** cleanup path reproduced that the security boundary is still structurally incomplete — an incompatible turn can still execute against an unverified slot:

1. **[CORE — fail-OPEN on unexpected isolation error] `api/streaming.py:7246`** — an unexpected failure during isolation acquisition is **logged and then the agent proceeds WITHOUT a backend lease**, executing tools against an unverified slot. For a cross-profile security boundary this must **fail closed**. Fix: convert unexpected failures to `TerminalBackendIsolationError` (or re-raise) so `AIAgent` construction + tool execution are blocked; add a test proving an unexpected acquisition error prevents tool execution.

2. **[CORE — Docker daemon/probe failure still admits + commits] `api/terminal_backend_isolation.py:144, :240, :479`** — reproduced: with a Docker **daemon failure** during cleanup, the outcome was `ADMITTED` and identity `committed=('local',)`. A nonzero/errored Docker probe must be treated as **failure** (unverifiable == failed → uncommitted + reject). Also: verify first-use slot state (only skip cleanup after **proving** the real `"default"` slot absent; if populated, do verified cleanup first), fingerprint all backend-creation inputs, and put **every** in-process tool-capable turn — including `/api/chat` — under the same lease so it can't create/use the slot concurrently.

3. **[SILENT — vacuous tests] `tests/test_issue5937_terminal_backend_isolation.py:54`** — all 26 pass but they inject `FakeAgent.cleanup_vm` and replace Docker probes with booleans, masking the reproduced false-commit. Import the installed `tools.terminal_tool`, seed its isolated `_active_environments`, invoke the **actual** `cleanup_vm`, and cover both swallowed-teardown-failure and nonzero-Docker-probe cases, asserting no incompatible lease is admitted.

This is round 4 on a cross-profile security boundary and the core invariant (no turn runs against an unverified backend) still has a fail-open hole on the error/daemon-failure paths. The structural direction is right; it needs: fail-closed on ALL acquisition errors (streaming.py:7246), treat unverifiable Docker probes as failure, first-use state verification, `/api/chat` under the lease, and tests against the real `cleanup_vm`. @nesquena — flagging this one: it's a genuinely hard isolation/concurrency boundary that's now 4 rounds deep with a reproduced fail-open; you may want to weigh in on the design (or whether the agent-side needs a verified-cleanup API) rather than iterate further in review.

jbbottoms:
## Round 4 — all three addressed at `a47a46c7`

Every finding was real. The daemon-failure false-commit you reproduced is now reproduced **by the test suite itself** against the installed agent — and rejected.

### 1. Fail-OPEN on unexpected isolation error (`streaming.py:7246`) — FIXED

The "log loudly and proceed" posture is gone. New single entry point `acquire_turn_lease_failclosed()` in the isolation module converts **any** unexpected acquisition error into the typed `TerminalBackendIsolationError`, so AIAgent construction + tool execution are blocked on every failure shape, not just deliberate rejections. Tests: a functional test injects a `RuntimeError` during acquisition and asserts the typed rejection surfaces with **no lease registered**; a source guard asserts the streaming path uses the fail-closed wrapper and that the proceed-path is gone. The trade is documented in-code: an isolation-module bug now aborts backend-switching turns loudly instead of running them open — correct for a cross-profile boundary.

### 2. Docker daemon/probe failure admitted + committed — FIXED (root cause: probe semantics)

`_container_exists` treated **any** nonzero `docker inspect` exit as "container gone" — a dead daemon read as a successful removal. It is now tri-state: exit 0 = exists; nonzero **with "no such object"** = verifiably gone; anything else **raises** (unverifiable == not verified → transition uncommitted → turn rejected). Your reproduced scenario is a permanent regression test driving the **installed** `tools.terminal_tool` — real `cleanup_vm` (pop-first, swallows teardown, returns None), real subprocess probe against a scripted `docker` emulating daemon failure — asserting REJECTED + identity uncommitted + container ledgered, then commit only after the "daemon recovers" and reports the object definitively absent.

Also per your list:

- **First-use verification** — the first lease no longer blind-records its identity. It commits only after *observing* the `"default"` slot absent **and** the pending-removal ledger empty (a prior failed first-use attempt pops the slot but must not be forgotten); a populated slot gets the full verified cleanup first, with force-removal keyed on the outgoing env object's own container evidence, since there is no previous identity to consult.
- **Fingerprint all backend-creation inputs** — identity now covers every `get_config()` input the agent consumes at environment creation for the active backend (docker: volumes / env / extra args / forward-env / network / cpu / mem / disk / user mode / cwd-mount / persistence flags; ssh: + persistent-shell; local: + persistent; modal: + resources), with JSON canonicalization and agent-truthy boolean normalization so formatting differences can't split identities. Still excluded by design: CWD (cross-workspace slot sharing) and per-command knobs (`TERMINAL_TIMEOUT` etc.) that don't shape the environment object.
- **`/api/chat` under the lease** — `_handle_chat_sync` constructs a tool-capable `AIAgent`; it now acquires the same fail-closed full-turn lease before its `TERMINAL_*` env mutation, releases in its `finally` after the restore, and surfaces rejections as a retryable 503. I audited the remaining `AIAgent(` sites: three are `enabled_toolsets=[]` (no tool surface) and `_handle_session_compress` only calls `context_compressor.compress()` (LLM round-trip, no tool execution) — flagging that scoping judgment rather than deciding it silently.

### 3. Vacuous tests — FIXED

New real-contract lane (`test_real_cleanup_vm_*`, `test_real_first_use_*`): seeds the installed module's `_active_environments`, resolves helpers through the module's own production resolution (no injected fakes), invokes the **actual** `cleanup_vm`, and drives the **actual** probe subprocess path via scripted docker executables. Covers swallowed-teardown-failure in both directions — rejected under daemon failure, committed under verified absence (proving the layer keys on observed state, not on exceptions) — plus the nonzero-probe case, asserting no incompatible lease is admitted. These skip via `importorskip` when hermes-agent isn't installed (your CI), where the contract-faithful fakes keep the logic covered; the agent-installed lane runs before every push here.

### Receipts

- Isolation file, agent-installed venv (real `cleanup_vm` with `force_remove`, real `get_active_env`): **43/43**.
- Agent-less venv (CI shape): **40 passed / 3 skipped**.
- Broad `-k "stream or terminal or backend or isolation or chat or profile"`: **2,177 passed / 27 skipped**, 1 failure in `test_issue1499_keyless_onboarding::TestKeylessChatReady::test_lmstudio_keyless_chat_ready_via_full_status` — reproduced **identically on unmodified `20b30b81`** under the same selection (2,160 passed there; the +17 delta is this PR's new tests). Pre-existing order-dependent failure, not introduced here; happy to file it separately.

On the flag to @nesquena: agreed this boundary is at the edge of what review-iteration alone can hold. The round-3 offer stands — glad to co-scope the agent-side synchronous verified-cleanup API (pairs with hermes-agent #63361), which would collapse the probe/ledger layer here to a single verified call. The round-4 fixes are self-contained within the current design either way.

Re-gate requested at `a47a46c7`.

nesquena-hermes (changes requested):
## 🔬 Re-gate round 5 (Codex, security boundary, real cleanup contract) — KICK BACK: fail-closed is fixed, but the backend IDENTITY is profile-incomplete (cross-profile reuse still possible)

Real progress this round @jbbottoms — the fail-closed transition machine is now working:
- ✅ **Unexpected streaming isolation error** now aborts the turn (`apperror`, zero agents constructed, no lease registered) — the round-4 fail-OPEN at streaming.py:7246 is CLOSED.
- ✅ **Nonzero Docker probe** with a reachable container: real `cleanup_vm` returned `None`, probe failed, turn stayed rejected/uncommitted.
- ✅ **Real return-`None` false-success** closed for the reachable-slot path.
- ✅ **Timeout abort**: incoming stream constructs zero agents; a future turn is admitted after the held lease releases.
- ✅ **Same-new-identity piggyback** waits for verified cleanup (tracked transitions). Real-contract tests: 40 passed.

But the deep gate reproduced two **fundamental** holes that still let an incompatible turn hit the wrong backend:

1. **[CORE — identity is profile-incomplete] `api/terminal_backend_isolation.py` `terminal_backend_identity()`** — the identity tuple omits `HERMES_HOME` (the profile) AND the forwarded-secret VALUES. Reproduced: profile **alpha** (`HERMES_HOME=/profiles/alpha`, `TENOR_API_KEY=alpha-secret`) and profile **beta** (`/profiles/beta`, `TENOR_API_KEY=beta-secret`) produce **byte-identical identities** → a beta turn reuses alpha's cached Docker backend **carrying alpha's secret**. This is the exact cross-profile leak the PR targets, re-opened one level up (the identity that gates reuse doesn't include what makes two profiles different). **The live-process environment cannot be the identity source** — you need an identity-complete, profile-safe env snapshot (include profile home + a fingerprint of the forwarded-secret set/values).

2. **[CORE — in-process absence ≠ external cleanup] ** — a missing in-process slot is treated as proof the external backend is gone, but Docker/Modal/Daytona state persists externally. Requires a **verifiable upstream persistent-backend cleanup/reattach contract** (likely agent-side coordination, not WebUI-only).

3. **[lease-leak] ** — synchronous setup-failure paths can leave the lease unreleased → a later differing-backend turn waits/aborts indefinitely. Enclose setup in guaranteed teardown; release from a nested `finally` that runs even when restoration fails.

@nesquena — flagging this one for your call: it's **5 rounds deep on a cross-profile security boundary**, and finding #1/#2 are no longer incremental — they're a design gap (identity-complete profile-safe snapshot + a real persistent-backend cleanup/reattach contract that likely needs the **agent-side** #62720 primary fix to coordinate with). This may be better resolved as a WebUI+agent joint design than another review round here. Options when you're back: (a) keep iterating in review with the identity-completeness + fail-closed-external-state spec above; (b) pause this WebUI defense-in-depth layer until the agent-side primary fix lands and re-scope; (c) you take the identity design. I've bounced with the precise findings either way. Great persistence on a genuinely hard isolation problem.

jbbottoms:
## Round 5 — findings #1 and #3 fixed at `c1ba289d`; #2 is agent-side, scoping below

Thanks for the deep re-gate — the round-4 fail-closed machine landing is a real milestone, and both of the new WebUI-layer holes were legit.

### 1. [CORE] Identity is now profile-complete — FIXED
The identity that gates reuse omitted the profile and the forwarded-secret VALUES; your alpha/beta repro (same backend config, different `HERMES_HOME` + `TENOR_API_KEY`) is exactly right. `terminal_backend_identity` now appends a profile suffix:

- **`HERMES_HOME`** — the coarse, universal boundary. Two distinct profiles never share a cached backend, *including* a persistent local shell that inherits the host env (not just Docker). Your repro's two profiles now produce different identities → no reuse.
- **A SHA-256 fingerprint of the forwarded-secret `(name, value)` pairs** from `TERMINAL_DOCKER_FORWARD_ENV`, read from the turn snapshot. This catches same-home turns whose forwarded credential *values* differ. It's **hashed**, so no secret value ever lands in the identity tuple or the logs that print it; it's order-independent over the name set; and I dropped the raw forward-env list field from the backend tuple so mere list-order can't over-split.

On your "the live process env can't be the identity source" point — agreed and fixed: the streaming turn path now stamps the **resolved** profile home into the identity snapshot (`_effective_turn_env['HERMES_HOME'] = _profile_home`) *before* acquisition, instead of sampling `os.environ['HERMES_HOME']` which is only mutated later and can still carry a prior turn's profile. The forwarded values are already in the snapshot via the profile's `.env` overlay (`filter_runtime_env_for_gateway_parity` keeps credential vars), so the fingerprint sees them.

New tests: distinct-home split, local-backend split, same-home/different-secret split, secret-never-in-tuple, unset-vs-empty, order-stability, and an end-to-end repro proving a beta turn force-removes alpha's container instead of reusing it.

### 3. [lease-leak] — FIXED (it was concrete on `/api/chat`)
You're right, and the sharpest instance was `_handle_chat_sync` (the round-4 `/api/chat` lease): the env mutation sat **between** the acquire and the guarding `try` (a throw there leaked the lease with **no backstop** on that path), and the release ran **after** the env restore in the same `finally` (a restore throw skipped it). Now: the env mutation is inside the guarded try, prior values are captured before any set, and the release is in a **nested finally** so it runs even if the restore raises. The streaming path's primary release is likewise nested now (its outer-finally backstop stays as the second line of defense, per your "release from a nested finally that runs even when restoration fails").

### 2. [CORE] In-process absence ≠ external cleanup — agreeing this is agent-side
This one I'm **not** claiming to fix in WebUI, and I think your own read is correct: a missing in-process `"default"` slot can't prove an external Docker/Modal/Daytona backend is gone, and a real verifiable persistent-backend cleanup/reattach contract belongs with the agent-side primary fix (#62720 / hermes-agent #63361), not this defense-in-depth layer. Forcing a WebUI-only answer would mean shipping something that *looks* like the primary fix but can't actually observe external state across process boundaries.

So on your options — **my vote is (b) for finding #2 specifically, (a) for #1/#3 (done above)**: keep this WebUI defense-in-depth layer moving on the parts it can own (identity-completeness + fail-closed, both now landed), and scope the external-state guarantee into the agent-side work rather than another WebUI round. I'd happily co-design the agent-side synchronous verified-cleanup API (the round-3/4 offer still stands) — that's the piece that makes #2 real, and it also collapses the probe/ledger layer here to a single verified call.

### Receipts (`c1ba289d`)
- Isolation file, agent-installed venv (real `cleanup_vm` + real subprocess probe): **55/55**.
- Agent-less venv (CI shape): **52 passed / 3 skipped**.
- Broad sweep (`-k stream/terminal/backend/isolation/chat_sync/profile`): **1931 passed / 0 failed / 27 skipped**.

@nesquena — flagging for your call on #2 as above: I think it's a WebUI (defense-in-depth, now identity-complete + fail-closed) + agent (authoritative external cleanup) joint design, not another review round here.

jbbottoms:
Good catch — fixed in `8858c66e`.

You're right that `_handle_chat_sync` was building the lease identity from `dict(os.environ)`, i.e. mutable process-global state. A streaming turn keeps its own profile's `HERMES_HOME` + terminal settings in `os.environ` for its whole run (mutated under `_ENV_LOCK`, then the lock is released while the agent runs), so a concurrent `/api/chat` for a *different* profile would compute the streaming profile's identity, match the last-committed one, skip invalidation, and let its tools reuse that profile's cached `"default"` backend — and it was an unlocked read racing the streaming mutation on top of that.

`/api/chat` now mirrors the streaming path: it resolves the session's own profile (`get_hermes_home_for_profile` + `get_profile_runtime_env` + the gateway-parity filter), builds the identity from that overlay with the resolved `HERMES_HOME` stamped in, and **applies** the same profile env under `_ENV_LOCK` for the turn (restored in the `finally`) — so the identity is derived from the resolved backend/profile settings and matches the backend the agent will actually create, not process-global state a concurrent turn owns. Added a source guard asserting the raw-`os.environ` acquire is gone and the identity comes from the resolved-profile overlay.

Receipts: 56/56 isolation (agent-installed), 53/3-skip agent-less, 1932 passed / 0 failed in the broad `-k stream/terminal/backend/isolation/chat_sync/profile` sweep.

nesquena-hermes (changes requested):
## 🔬 Security re-gate round 6 (Codex, adversarial) — SHIP ONLY WITH FIXES (resolved-profile identity helped; the remote-backend boundary is still open)

Progress — the resolved-profile identity fix means the `HERMES_HOME`/profile dimension now distinguishes correctly, fail-closed holds on resolution failures, lease counts drain on normal/error/cancel (no deadlock), and 59 single-profile/switch/steer/cancel tests pass. But adversarial probes confirm the boundary is **not yet complete**: **two distinct profiles can still reuse the same persistent Daytona backend**, so this can't ship on the isolation surface yet. Five findings:

1. **[SECURITY — persistent Daytona cross-profile resume] `api/terminal_backend_isolation.py:80-86` + `:562-632`** — persistent Daytona can resume *another profile's* remote sandbox. **Fix:** prevent cross-profile resume via verified deletion, profile-namespaced sandbox identity, or fail-closed rejection when the sandbox's owning profile can't be confirmed.

2. **[SECURITY — identity omits remote-backend fingerprints] `api/terminal_backend_isolation.py:286-318`** — the identity still omits the Modal / Singularity / Daytona **constructor + persistence + connection-target + credential fingerprints** (the r5 incompleteness class, one layer deeper — remote backends). **Fix:** fold every remote-backend dimension that distinguishes one sandbox from another into the identity.

3. **[SECURITY — live os.environ identity paths remain] `api/streaming.py:7151-7253` + `api/routes.py:21967-21989`** — some identity construction still reads live `os.environ` rather than the scrubbed resolved-profile environment. **Fix:** build identity from a scrubbed, immutable resolved-profile env on every path; reject all resolution failures.

4. **[CORE — env leak on pre-setup failure] `api/streaming.py:7258-7415` + `:10450-10462`** — a setup failure *before* the inner agent `try` releases the lease but leaks the applied environment (env not restored). **Fix:** guarantee complete environment restoration before every lease release, including setup failures before the inner try.

5. **[test] `tests/test_issue5937_terminal_backend_isolation.py:1098-1337`** — 53 pass / 3 skip but they're source-only; they don't run the production concurrent-live-env attack. **Fix:** runtime concurrent-profile, pre-setup-failure, persistent-Daytona, and per-identity-dimension regressions.

Net: the local-profile (`HERMES_HOME`) dimension is fixed, but the **remote-backend path (persistent Daytona/Modal) can still bridge two profiles** — the identity doesn't yet fingerprint the remote sandbox/credentials, and persistent Daytona can resume a foreign profile's sandbox. That's the core boundary, so it stays blocked. Rebased clean on current master.

Issue #5686, opened 2026-07-06
WebUI appends 'No response from provider' after completed assistant response


lukea3671-Claude:
## Bug description

A WebUI chat turn can complete successfully in the agent backend, but the WebUI stream/session layer still appends a synthetic `No response from provider` error card and leaves a duplicate `_partial` assistant message in the session JSON.

This makes it look to the user like the provider/model failed, censored, or deleted the answer even though the final assistant response exists and the backend logged `finish_reason=stop`.

## Why this matters

This is a conversation-continuity / data-loss-adjacent UX bug:

- The user sees a scary `No response from provider` card after a real answer.
- The same assistant answer can appear twice in session state: once as clean `_db_persisted`, once as `_partial`.
- The turn journal may not record `assistant_started` / `completed` for the affected stream.
- The generic hint points users toward quota/provider switching even when the provider returned a normal final response.

## Observed environment

- WebUI server header: `HermesWebUI/0.51.875-8-g932c3d06 Python/3.12.3`
- Local repo remote: `https://github.com/nesquena/hermes-webui.git`
- Local branch checked during diagnosis: `argus/v0.51.699-with-local-patches`
- Local commit checked during diagnosis: `8710419c`
- Provider/model: `openai-codex` / `gpt-5.5`
- Surface: WebUI session

## Reproduction / observed sequence

This happened on at least two tool-using turns in the same WebUI session. Both involved memory/tool calls before the final answer.

### Turn A

User submitted a message at `2026-07-06 23:25:44` local time.

Agent log shows the backend completed normally:

```text
2026-07-06 23:25:44 conversation turn: session=f91b1c15e47f model=gpt-5.5 provider=openai-codex platform=webui ...
2026-07-06 23:25:58 tool memory completed
2026-07-06 23:26:12 API call #77 ... out=346 ... latency=14.5s
2026-07-06 23:26:12 Turn ended: reason=text_response(finish_reason=stop) ... response_len=1547 session=f91b1c15e47f
```

But the WebUI session JSON had this pattern:

```text
clean assistant message:
  _db_persisted: true
  finish_reason: stop
  content_len: 1547

duplicate assistant message:
  _partial: true
  same content_len: 1547
  includes _partial_tool_calls

error card:
  _error: true
  content: **No response from provider:** No response from provider.
  provider_details: No response from provider.
```

The turn journal for that stream showed `submitted` and `worker_started`, but no terminal `assistant_started` / `completed` event for the stream.

### Turn B

[...]

Issue #5940, opened 2026-07-11
WebUI error reporting masks the real provider error: shows 'No response from provider / try again' when the agent hit a non-retryable HTTP 400 (invalid model / no credentials)


nesquena-hermes:
Reported by **b3nw** in Discord `#report-bugs` (Jul 11 2026). Repro on **exp-v0.52.38 · Experimental**, Agent 0.18.0.

> "error reporting still has issues not reflecting what actually is happening from hermes"

## Symptom

A turn fails and the WebUI shows the user a **"No response from provider"** error with this hint:

> *The provider returned no content and no error. This often means a usage/rate limit was hit silently. Check provider status, switch providers via `hermes model`, or try again in a moment.*

("Provider details" expands to just `No response from provider.`)

But the **actual** error in the agent console log at the same moment (≈5s earlier) is a concrete, non-retryable HTTP 400:

```
API call failed (attempt 1/3): BadRequestError [HTTP 400]
Error: HTTP 400: {"detail":"Invalid Request: Invalid model format or no credentials for provider: <redacted>"}
Non-retryable error (HTTP 400). Aborting.
This type of error won't be fixed by retrying.
```

## Why this matters

The user-facing message is **actively misleading** in two ways:
1. It says "no content **and no error**" — but there *was* a clear error (HTTP 400 with a `detail` string).
2. It attributes the failure to a **silent rate/usage limit** and tells the user to **"try again in a moment"** — but the real error is **explicitly non-retryable** ("won't be fixed by retrying"). The correct guidance would be to fix the model id / provider credentials, not wait and retry.

This directly wastes the user's time (retrying a request that can never succeed) and hides the real, actionable cause.

## Investigation

The WebUI error classifier (`api/streaming.py` `_classify_*` around lines 1201-1291) already has the right bucket for this: `_is_not_found` (line 1227) matches `'invalid model' in err_lower`, and `_is_auth` (line 1214) matches credential problems — either of which would produce a correct, actionable hint. The `detail` string here ("Invalid model format or no credentials") would satisfy `_is_not_found`.

[...]

PR #5990, opened 2026-07-12
fix(streaming): avoid false no-response after final tool answer

jtstothard:
## Thinking Path
- Issue #5686 is a streaming/final-settlement bug: a completed assistant turn can still render a false `No response from provider` card.
- The remaining gap is final-answer detection for structured assistant content with tool metadata.

## What Changed
- normalize visible text extraction from structured assistant content parts
- detect tool-use boundaries inside assistant `content[]`
- treat assistant rows with visible prose after tool metadata as final answers
- reject tool-only assistant rows as final answers
- add regression coverage in `tests/test_issue5686_completed_output_text.py`

## Why It Matters
- avoids false `No response from provider` cards after successful tool-using turns
- keeps transcript settlement closer to what the user actually saw

## Related Open PRs
- `#5902`: checkpointed-current-user / `previous_display` path — complementary, **not superseded**
- `#5634`: stale/recovered-worker persisted-final-answer path — complementary, **not superseded**
- this PR covers structured assistant content + tool-boundary handling in final-answer detectors

## Contract Routing
Task type: runtime/streaming bug fix
Touched areas: transcript settlement, final-answer detection, assistant/tool content interpretation
Relevant docs: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `docs/rfcs/webui-run-state-consistency-contract.md`
Evidence: focused regression test + targeted verification run

## Verification
- `./scripts/test.sh -q tests/test_issue1765_codex_quota.py tests/test_issue5686_completed_output_text.py` → `8 passed`
- pre-push repo checks also passed → `44 passed`

## Risks / Follow-ups
- scoped to this final-answer detection path only
- other false-`no_response` paths may still need separate fixes

## AI Usage Disclosure
- Providers/models: GLM-4.7 (`coder`) and GPT-5.6 Terra/Sol (`coder-codex`)
- Tooling: terminal, file editing, targeted tests, `gh-write`

## Model Used
- `coder`
- `coder-codex`
Final patch synthesized from both pilot results and validated locally.

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -2565,6 +2565,15 @@ def _looks_invalid_generated_title(text: str) -> bool:
     )
 
 
+def _message_content_part_text(part) -> str:
+    """Extract visible text from a structured content part."""
+    if not isinstance(part, dict):
+        return ''
+    return str(
+        part.get('text') or part.get('content') or part.get('input_text') or part.get('output_text') or ''
+    )
+
+
 def _message_text(value) -> str:
     """Extract plain text from mixed message content payloads."""
     if isinstance(value, list):
@@ -2574,9 +2583,41 @@ def _message_text(value) -> str:
                 continue
             ptype = str(p.get('type') or '').lower()
             if ptype in ('', 'text', 'input_text', 'output_text'):
-                parts.append(str(p.get('text') or p.get('content') or ''))
+                parts.append(_message_content_part_text(p))
         return _strip_thinking_markup('\n'.join(parts).strip())
     return _strip_thinking_markup(str(value or '').strip())
+
+
+def _assistant_content_part_is_tool_use(part) -> bool:
+    """Return True when a content[] part represents a tool invocation boundary."""
+    if not isinstance(part, dict):
+        return False
+    part_type = str(part.get('type') or '').lower()
+    if part_type in {'tool_use', 'tool_call'}:
+        return True
+    if part_type:
+        return False
+    return any(part.get(key) for key in ('tool_use_id', 'tool_call_id', 'call_id')) and bool(
+        part.get('name') or part.get('tool_name') or part.get('input') or part.get('args')
+    )
+
+
+def _assistant_message_has_final_visible_text(message) -> bool:
+    """Return True when an assistant row carries a settled visible answer."""
+    if not isinstance(message, dict) or message.get('role') != 'assistant':
+        return False
+    content = message.get('content', '')
+    if isinstance(content, list):
+        last_tool_idx = -1
+        for idx, part in enumerate(content):
+            if _assistant_content_part_is_tool_use(part):
+                last_tool_idx = idx
+        tail_parts = content[last_tool_idx + 1:] if last_tool_idx >= 0 else content
+        return bool(_message_text(tail_parts).strip())
+    if message.get('tool_calls'):
+        return False
+    return bool(_message_text(content).strip())
+
 
 
 _WORKSPACE_PREFIX_RE = re.compile(r'^\s*\[Workspace::v1:\s*(?:\\.|[^\]\\])+\]\s*')
@@ -4873,7 +4914,7 @@ def _raw_message_text(value) -> str:
     """
     if isinstance(value, list):
         return ' '.join(
-            str(p.get('text') or p.get('content') or '')
+            _message_content_part_text(p)
             for p in value
             if isinstance(p, dict)
         )
@@ -5582,7 +5623,7 @@ def _assistant_reply_added_after_current_turn(result_messages, previous_context,
         isinstance(m, dict)
         and m.get('role') == 'assistant'
         and not m.get('_error')
-        and str(m.get('content') or '').strip()
+        and _assistant_message_has_final_visible_text(m)
         for m in candidates
     )
 
@@ -5600,18 +5641,7 @@ def _session_lacks_final_assistant_answer(messages) -> bool:
         if role == 'tool':
             return True
         if role == 'assistant':
-            content = msg.get('content')
-            if isinstance(content, list):
-                text = '\n'.join(
-                    str(part.get('text') or part.get('content') or '')
-                    for part in content
-                    if isinstance(part, dict)
-                )
-            else:
-                text = str(content or '')
-            if msg.get('tool_calls'):
-                return True
-            if text.strip():
+            if _assistant_message_has_final_visible_text(msg):
                 return False
             continue
         if role == 'user':
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -2597,8 +2597,10 @@ def _assistant_content_part_is_tool_use(part) -> bool:
         return True
     if part_type:
         return False
-    return any(part.get(key) for key in ('tool_use_id', 'tool_call_id', 'call_id')) and bool(
-        part.get('name') or part.get('tool_name') or part.get('input') or part.get('args')
+    if _message_content_part_text(part).strip():
+        return False
+    return any(key in part for key in ('tool_use_id', 'tool_call_id', 'call_id')) and any(
+        key in part for key in ('name', 'tool_name', 'input', 'args')
     )
 
 
@@ -2612,8 +2614,12 @@ def _assistant_message_has_final_visible_text(message) -> bool:
         for idx, part in enumerate(content):
             if _assistant_content_part_is_tool_use(part):
                 last_tool_idx = idx
-        tail_parts = content[last_tool_idx + 1:] if last_tool_idx >= 0 else content
-        return bool(_message_text(tail_parts).strip())
+        if last_tool_idx >= 0:
+            tail_parts = content[last_tool_idx + 1:]
+            return bool(_message_text(tail_parts).strip())
+        if message.get('tool_calls'):
+            return False
+        return bool(_message_text(content).strip())
     if message.get('tool_calls'):
         return False
     return bool(_message_text(content).strip())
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
Reading `api/streaming.py` on `origin/master` — `_session_lacks_final_assistant_answer` (:5590-5617), `_assistant_reply_added_after_current_turn` (:5572-5588), and the settlement call site (:8756-8816) — against the diff and the new `tests/test_issue5686_completed_output_text.py`. This lands on the actual root cause of #5686 for tool-using turns, and the fix is well targeted.

## The two master bugs this fixes

Both false-`No response from provider` triggers for tool turns live in `_session_lacks_final_assistant_answer`. Master:

```python
if role == 'assistant':
    content = msg.get('content')
    if isinstance(content, list):
        text = '\n'.join(
            str(part.get('text') or part.get('content') or '')
            for part in content
            if isinstance(part, dict)
        )
    ...
    if msg.get('tool_calls'):
        return True      # (1) forces "no answer" even when prose is present
    if text.strip():
        return False
```

1. **`tool_calls` short-circuits before the text check.** Any assistant row that carried a tool invocation returns `True` (lacks-final-answer) regardless of visible prose that followed. A settled "here's the weather: 18C" answer that used a tool is mislabeled as no-response.
2. **`output_text` parts extract as empty.** The inline extractor only reads `part.get('text') or part.get('content')`, so a structured part of shape `{"type": "output_text", "output_text": "..."}` yields `''` and the turn looks empty.

The PR's `_assistant_message_has_final_visible_text` fixes both: it finds the last tool-use boundary, then checks whether visible prose exists *after* it (`tail_parts`), and routes text extraction through `_message_content_part_text`, which now also reads `input_text`/`output_text`:

```python
def _message_content_part_text(part) -> str:
    return str(
        part.get('text') or part.get('content')
        or part.get('input_text') or part.get('output_text') or ''
    )
```

That's the right shape — prose-after-tool counts, tool-only rows still return `False`.

## Verification I did

The three settlement paths all now funnel through the same helper, which is the correctness win — `_assistant_reply_added_after_current_turn` (:5585), `_session_lacks_final_assistant_answer` (:5641), and `_raw_message_text` (:4917) previously each had their own slightly different extraction. Consolidating them means the merge-display check (`_merged_transcript_lacks_final_assistant_answer` at the call site :8781) and the append check now agree on what "visible text" means, which is exactly the inconsistency that let a settled turn slip through as no-response.

The regression tests cover the discriminating cases well: `test_session_lacks_final_answer_true_for_tool_only_assistant_content` (tool-only → still True), `..._false_when_visible_output_text_follows_tool_use` (prose after tool → False), and `..._false_for_plain_output_text_parts` (bare `output_text` → False). Greptile scored 5/5 and CI is green.

## One edge to consider

`_assistant_content_part_is_tool_use` treats an untyped part with a `tool_use_id`/`call_id` **and** a name/input as a tool boundary. If a provider ever emits a content part that carries a `call_id` for correlation but is actually the visible answer text, the `and bool(part.get('name') or ...)` guard should keep it from being misclassified — but it's worth a quick check against the structured shapes GLM/GPT-5.6 actually emit, since those are the two models named in the PR. Not a blocker; the typed-part path (`tool_use`/`tool_call`) covers the common case.

Complementary to #5902 (checkpointed-current-user path) and #5634 (stale-worker path) as the description states — this one closes the structured-content + tool-boundary gap those don't touch.

nesquena-hermes (changes requested):
## 🔬 Gate review (Codex) — SHIP ONLY WITH FIXES (intended fix works; 3 precedence edges + a #5940 interaction)

Thanks @jtstothard — the intended fix is verified working: a typed `tool_use → output_text` final answer is now correctly recognized (no false no_response card), tool-only typed content stays fail-closed, null/missing content doesn't throw, and part-scanning is linear (100k parts in 0.06s). The 12 shipped/new tests pass. But the gate found three precedence issues before this can ship on the crown-jewel no_response path:

1. **[CORE — masks a real terminal error, interacts with shipped #5940] `api/streaming.py:8820`** — a captured non-retryable terminal error (`_captured_terminal_error`, from #5940 / exp-v0.52.47) can now be **suppressed**: verified an HTTP 400 callback emitted `done` and persisted **no** `_error`, because the new final-text detection wins over the terminal-error path. **Fix:** `_captured_terminal_error[0]` must force `_terminal_failure` and emit `apperror`, not `done` — a real provider error must never be masked as "has final text." Add an integration regression (token → HTTP 400 → structured content) asserting `apperror` + persisted `_error`.

2. **[CORE — over-suppression regression] `api/streaming.py:2600`** — the untyped tool-boundary heuristic false-positives **normal visible text** that happens to carry `call_id` + `name`: verified `"Final visible answer."` is classified as "lacks final answer," regressing a shape master currently accepts (would show a false no_response). **Fix:** visible text must disqualify the untyped tool heuristic; recognize empty `input`/`args` by key *presence*, not truthiness.

3. **[SILENT — under-suppression] `api/streaming.py:2609`** — list content with pre-tool text + only **top-level** `tool_calls` (no in-content boundary) is classified as a final answer, silently suppressing a legitimate `no_response`. **Fix:** when no in-content boundary is found but top-level `tool_calls` exists, preserve fail-closed (return no final answer).

The typed-content core is right — these are the precedence rules around it: terminal error must win over final-text detection (#1), and the untyped/top-level-tool_calls heuristics must not over- or under-suppress (#2, #3). The existing tests don't exercise these precedence cases, so please add coverage for each. Re-gate once terminal-error precedence + the two heuristic edges are fixed.

jtstothard:
Addressed the 3 blocking review points in `ac68546b`.

- terminal-error precedence now wins over final-text detection, so captured non-retryable provider errors emit `apperror` / persist `_error` instead of falling through to `done`
- untyped `call_id` / `name` parts with visible text no longer false-positive as tool-use boundaries
- list content with only top-level `tool_calls` and no in-content boundary now stays fail-closed (no false settled final answer)

Coverage added/expanded:
- `tests/test_issue5686_completed_output_text.py`
- `tests/test_issue5121_provider_auth_terminal_error.py`

Verification:
- `./scripts/test.sh -q tests/test_issue5686_completed_output_text.py tests/test_issue5121_provider_auth_terminal_error.py` → `25 passed`
- pre-push targeted suite → `45 passed`
- `./scripts/test.sh -q tests/test_issue5686_completed_output_text.py tests/test_issue5121_provider_auth_terminal_error.py tests/test_sprint34.py` → `52 passed`

Could you please re-review?

nesquena-hermes:
## Re-gate (Codex) — all three blocking points resolved at `ac68546b`; PASS from my side

Re-read the diff at `ac68546be` against my three CHANGES_REQUESTED findings, plus the two expanded test files. All three land, and each is now backed by a discriminating regression test. Verdict: PASS.

### 1 — Terminal-error precedence (CORE) — fixed

The captured non-retryable error now forces the failure paths instead of being masked by final-text detection. `api/streaming.py:8803` computes the flag once, then feeds all three downstream decisions:

```python
_captured_terminal_failure = bool(_captured_terminal_error[0])
if not _last_err and _captured_terminal_failure:
    _last_err = _captured_terminal_error[0]
...
_drop_replayed_assistant = (
    _captured_terminal_failure
    or _agent_result_terminal_failure(result)
    ...
)
_terminal_failure = (
    _captured_terminal_failure
    or _is_agent_result_terminal
    or (_saved_transcript_lacks_final_answer and _classification['type'] not in {'cancelled','interrupted'})
)
```

Since `_terminal_failure` short-circuits on `_captured_terminal_failure`, an HTTP 400 that also carried a structured `tool_use → output_text` answer now takes the `put('apperror', ...)` branch at `:9108` and returns — it can no longer settle as `done`. `test_captured_terminal_http_400_beats_structured_final_answer` exercises exactly this (status callback emits the non-retryable error, the result carries a settled `output_text` tail) and asserts `apperror` with `type == "model_not_found"`, no `done`, and `saved.messages[-1]["_error"] is True`. That's the precise interaction with #5940 I flagged.

### 2 — Untyped tool-boundary over-suppression — fixed

`_assistant_content_part_is_tool_use` (`api/streaming.py:2591`) now disqualifies any untyped part that carries visible text, and recognizes tool args by **key presence** rather than truthiness:

```python
if _message_content_part_text(part).strip():
    return False
return any(key in part for key in ('tool_use_id','tool_call_id','call_id')) and any(
    key in part for key in ('name','tool_name','input','args')
)
```

`test_session_lacks_final_answer_false_for_untyped_visible_text_with_call_metadata` (a part with `call_id` + `name` + `text`) asserts `False`, and `..._true_for_untyped_tool_part_with_empty_args` (a `call_id` + `args:{}` part, no text) asserts `True` via the presence check. Both pass.

### 3 — Top-level `tool_calls` without in-content boundary — fixed

In `_assistant_message_has_final_visible_text` (`:2609`), when no in-content boundary is found but top-level `tool_calls` exists, it now fails closed:

```python
if last_tool_idx >= 0:
    tail_parts = content[last_tool_idx + 1:]
    return bool(_message_text(tail_parts).strip())
if message.get('tool_calls'):
    return False
return bool(_message_text(content).strip())
```

`test_session_lacks_final_answer_true_for_top_level_tool_calls_without_in_content_boundary` (list content with only a pre-tool `text` part + top-level `tool_calls`) asserts the session still lacks a final answer. Good — that closes the silent under-suppression.

### Note on the empty-string `error` case

Worth calling out because it's subtle: `test_captured_terminal_http_400_beats_structured_final_answer` returns `"error": ""` in the result dict, so the win comes purely from `_captured_terminal_error[0]` (the status-callback capture), not from `result["error"]`. That's the right test shape — it proves the *captured* path drives the decision independently of the result-dict error field.

The typed-content core was already right; these were the precedence rules around it, and they're now correct and covered. No further blocking concerns from me — pending your merge call.

jtstothard:
Rebased onto current master at `c8bf3a10`. All tests pass (29 focused + 45 pre-push). Could you re-approve on the new commit so GitHub clears the review state? @nesquena-hermes

nesquena-hermes:
Shipped in **exp-v0.52.57** 🚀 — the false 'No response from provider' after a completed tool-using answer is fixed. Final-answer detection now understands structured content parts (text after tool-use boundaries); genuine empty completions still show no_response; a real terminal error still surfaces as apperror (never masked). Codex SAFE (round 2, all 3 precedence findings fixed), stream-gate GREEN all modes. I updated the #5940 source-assert test for your `_captured_terminal_failure` rename so CI stays green. Thanks @jtstothard! Closes #5686.

PR #5994, opened 2026-07-12
fix: allow 'fork from here' during active streaming

### LINKED ISSUE #5993: 'Fork from here' button silently does nothing while agent is streaming
## 'Fork from here' button silently does nothing while agent is streaming

**Labels:** `bug`, `ux`

**Reproduction environment:** Any version with streaming (confirmed on exp-v0.52.22)

### Symptom

While the agent is generating a response (streaming), clicking the "Fork from here" button on any message does nothing. No toast, no error, no session created — the click is silently swallowed. The button is visible and appears interactive, but it's a dead affordance.

### User scenario

A user is exploring multiple hypotheses with the agent. They send a message to test hypothesis A and while waiting for the response, want to fork the conversation at an earlier message to start testing hypothesis B in a new session. Currently this workflow is blocked — the user has to either:
- Wait for the entire stream to finish, losing time
- Use `/stop` to cancel the stream, losing context

Both are unnecessary because the fork point is a **past message** that is already fully committed (stored server-side). The ongoing stream only affects messages **after** the fork boundary.

### Actual behavior

Button click → `forkFromMessage()` called → first line `if(!S.session||S.busy)return;` → silent return.

### Expected behavior

Fork should succeed immediately for any past (non-live) message, regardless of whether a newer message is still streaming. The button should either:
- Work during streaming (fork the session, user can switch to it and type while the original stream continues), or at minimum
- Show a meaningful toast explaining why it can't fork when clicked on the streaming message itself

### Root cause

`static/commands.js:1928` in `forkFromMessage()`:

```js
async function forkFromMessage(msgIdx){
  if(!S.session||S.busy)return;  // ← silently blocks ALL forks during streaming
```

The `S.busy` guard is an over-broad blanket check. `S.busy` is set to `true` for the entire duration of a streaming response, but forking a past message only accesses server-persisted data before the fork boundary — it has no dependency on the in-flight stream.

The `S.busy` check was likely added conservatively to prevent data races, but the `/api/session/branch` endpoint:
- Does not check `active_stream_id`
- Simply copies `source_messages[:keep_count]` — all of which are already persisted
- Works fine during streaming

### Proposed fix

**`static/commands.js` — `forkFromMessage()` (2 changes, ~7 lines):**

1. **Line 1928** — replace `if(!S.session||S.busy)return;` with a targeted guard that only blocks forking the specific live message, not all messages:

```js
if(!S.session)return;
if(S.busy && msgIdx > latestNonLiveMsgIdx) { showToast(/* ... */); return; }
```

2. **After line 1940** — when `S.busy` is true, skip `_ensureAllMessagesLoaded()` to avoid visual flicker (the streaming message would briefly collapse to whatever partial version the server persisted):

```js
if(!S.busy && typeof _ensureAllMessagesLoaded==='function'){
  await _ensureAllMessagesLoaded();
}
```

The `absoluteKeepCount = _oldestIdx + msgIdx` is already captured **before** any async work and is stable regardless of `_ensureAllMessagesLoaded`.

Optionally, `latestNonLiveMsgIdx` can be derived from `latestRenderedAssistantRawIdx` (already computed in `ui.js:14469-14475`) or the check can simply be: if the clicked message has `_live === true`, show a toast with "Cannot fork a message still being generated".

### Links

- `static/commands.js` — `forkFromMessage()`: lines 1927-1962
- `static/ui.js` — fork button rendering: line 14601
- `static/ui.js` — `latestRenderedAssistantRawIdx` (excludes `_live` messages): lines 14469-14475
- `api/routes.py` — `POST /api/session/branch` handler: line 14363
- `api/routes.py` — `_load_branch_source_or_refuse`: line 7831 (no streaming check on the server side)

kertisalex:
Removes the blanket `S.busy` guard from `forkFromMessage()` that silently
swallowed every "Fork from here" click while the agent was generating a
response.

**Before:** Button is visible but dead — `forkFromMessage()` returns
silently on the first line when `S.busy === true`.

**After:**
1. Fork proceeds immediately when the user clicks a *past* (non-live)
   message, regardless of whether a newer message is still streaming.
2. Fork on the currently-streaming message itself is rejected with a
   clear toast: "Cannot fork a message still being generated."
3. `_ensureAllMessagesLoaded()` is skipped during streaming to avoid
   visual flicker — `absoluteKeepCount` is captured before any async
   work and is unaffected.

**Why it's safe:** The `/api/session/branch` endpoint does not check
`active_stream_id` — it copies `source_messages[:keep_count]`, all of
which are already committed server-side long before the stream started.

Closes #5993

### DIFF
--- a/static/commands.js
+++ b/static/commands.js
@@ -1932,7 +1932,8 @@ async function forkFromMessage(msgIdx){
   // waiting for the stream to finish.
   if(S.busy){
     const _msg=(Array.isArray(S.messages)&&S.messages[msgIdx-1])||null;
-    if(_msg&&_msg._live){
+    const _isLastMsg=msgIdx>=(Array.isArray(S.messages)?S.messages.length:0);
+    if((_msg&&(_msg._live||_msg._pending))||_isLastMsg){
       showToast('Cannot fork a message still being generated.',3000);
       return;
     }
--- a/static/commands.js
+++ b/static/commands.js
@@ -1925,7 +1925,18 @@ async function cmdBranch(args){
 // count (_oldestIdx + msgIdx) BEFORE awaiting _ensureAllMessagesLoaded,
 // which resets _oldestIdx to 0 after its wholesale replace.  See #2184.
 async function forkFromMessage(msgIdx){
-  if(!S.session||S.busy)return;
+  if(!S.session)return;
+  // During streaming, only block fork if the clicked message is the
+  // currently-streaming (live) message itself.  Past messages that are
+  // already committed server-side can be forked immediately without
+  // waiting for the stream to finish.
+  if(S.busy){
+    const _msg=(Array.isArray(S.messages)&&S.messages[msgIdx-1])||null;
+    if(_msg&&_msg._live){
+      showToast('Cannot fork a message still being generated.',3000);
+      return;
+    }
+  }
   const readOnlySession=typeof _isReadOnlySession==='function'
     ? _isReadOnlySession(S.session)
     : !!(S.session&&(S.session.read_only||S.session.is_read_only));
@@ -1940,7 +1951,9 @@ async function forkFromMessage(msgIdx){
   const absoluteKeepCount = _oldestIdx + msgIdx;
   // Ensure the full transcript is loaded so the forked session renders
   // correctly and subsequent operations see the complete history.
-  if(typeof _ensureAllMessagesLoaded==='function'){
+  // Skip during streaming to avoid visual flicker — the fork data
+  // (absoluteKeepCount) was already captured above and is unaffected.
+  if(!S.busy && typeof _ensureAllMessagesLoaded==='function'){
     await _ensureAllMessagesLoaded();
   }
   if(!S.session || S.session.session_id !== initialSid) return;
--- a/static/messages.js
+++ b/static/messages.js
@@ -1600,7 +1600,7 @@ async function send(){
   let _composerDraftClearPromise=null;
   if (activeSid && typeof _clearComposerDraft === 'function') _composerDraftClearPromise=_clearComposerDraft(activeSid,_submittedDraftTextForClear,_submittedDraftFilesForClear);
   const displayText=_slashDisplayTextOverride||text||(uploaded.length?`Uploaded: ${uploadedNames.join(', ')}`:'(file upload)');
-  const userMsg={role:'user',content:displayText,attachments:uploaded.length?uploadedNames:undefined,_ts:Date.now()/1000};
+  const userMsg={role:'user',content:displayText,attachments:uploaded.length?uploadedNames:undefined,_ts:Date.now()/1000,_pending:true};
   S.toolCalls=[];  // clear tool calls from previous turn
   clearLiveToolCards();  // clear any leftover live cards from last turn
   let optimisticMessages;
@@ -7668,7 +7668,11 @@ function showClarifyCard(pending) {
   _clarifySessionId = sid;
   _clarifyId = pending.clarify_id || null;
   _clarifySignature = sig;
-  _startClarifyCountdown(pending);
+  if (Number(pending.timeout_seconds) > 0) {
+    _startClarifyCountdown(pending);
+  } else {
+    _clearClarifyCountdownTimer();
+  }
   if (!sameClarify) {
     _clarifyVisibleSince = Date.now();
     _clearClarifyHideTimer();


nesquena-hermes (changes requested):
## 🔬 Gate review (Codex) — SHIP ONLY WITH FIXES (one off-by-one on the in-flight user prompt)

Thanks @kertisalex — the core of this is right and I verified the parts I was worried about:
- ✅ **No truncation:** skipping `_ensureAllMessagesLoaded()` while busy does NOT truncate or shift a fork of committed past messages — `_oldestIdx + msgIdx` remains the correct absolute boundary and the fork is fully loaded afterward.
- ✅ **Indexing correct:** `rawIdx+1` from the button → `msgIdx-1` resolves back to the clicked row.
- ✅ A committed past-message fork includes all history through that message and excludes later pending/live content; read-only guard + `initialSid` protection intact; `commands.js` wins over the legacy `ui.js` copy. 58 branching tests pass.

One real boundary bug remains:

- **[SILENT — forks one message too early on the live prompt] `static/commands.js:1952`** — forking from the **current user prompt while its response is still streaming** creates a fork one message short. The `S.busy` guard only rejects `_live` **assistant** rows, but in the default deferred-save mode the current user row exists only in `pending_user_message`, and `/api/session/branch` slices `source.messages` (which excludes the pending row). Reproduced against the real route: `keep_count=101` against 100 committed messages + 1 pending prompt produced a **100-message fork ending at message 99** (the just-sent prompt is dropped). **Fix:** mark the optimistic user row `_pending:true` at `static/messages.js:1632`, and reject `_msg._live || _msg._pending` at `static/commands.js:1954` with the active-turn toast (same UX as the live-assistant case). Add focused cases proving (a) committed tail-window rows fork at `_oldestIdx + msgIdx`, and (b) a pending/live row does **not** POST a branch.

That's the only blocker — the past-message forking (the actual feature ask) works correctly. Mark the pending row + extend the guard to `_pending`, add the two test cases, and re-gate. Nice, tight fix otherwise.

nesquena-hermes:
## Review (Codex) — fix is correct; the guard indexing is safe, one edge to confirm

Read `forkFromMessage` on `origin/master` (`static/commands.js:1946`), the diff at this PR's head, the fork-button render site (`static/ui.js:15370`), and the `/api/session/branch` handler (`api/routes.py:14733`). The change is sound and the "safe during streaming" reasoning holds up server-side.

### Server side genuinely doesn't care about the stream

The branch handler slices already-committed history and never consults `active_stream_id`:

```python
# api/routes.py:14778-14784
source_messages = (
    _merged_session_messages_for_display(source, cli_messages)
    if is_messaging_session and cli_messages
    else list(source.messages or [])
)
if keep_count is not None:
    forked_messages = source_messages[:keep_count]
```

`absoluteKeepCount` is captured before any `await` (`commands.js:1961`), and the live streaming tail sits at indices *after* the fork boundary, so the prefix copy is stable whether or not the stream is running. The PR's safety claim is accurate.

### On greptile's "wrong array entry" concern — it doesn't apply here

Greptile flagged that hidden tool/compaction entries could shift the live-message lookup. I checked the index contract and it's fine: the fork button is rendered with `forkFromMessage(rawIdx+1)` (`ui.js:15370`), and `rawIdx` is a **direct index into `S.messages`** — the counter increments for *every* element, not just renderable ones:

```js
// ui.js:633-637
let rawIdx=0;
for(const m of (S.messages||[])){
  if(_messageIsRenderable(m)) rebuilt.push({m,rawIdx});
  rawIdx++;   // advances for hidden entries too
}
```

So `S.messages[msgIdx-1]` in the new guard resolves to exactly the clicked message — hidden entries don't desync it. The live-message check is correctly targeted.

### The one edge worth a look: does `S.messages[msgIdx-1]._live` actually carry the flag?

This is the part I'd verify before merge. During an active stream the live assistant row enters `S.messages` through the inflight merge path, not a plain push. `_ensureInflightLiveAssistantMessage` (`sessions.js:3246`) stamps `_live:true`, and `_mergeInflightTailMessages` (`sessions.js:3474`) appends that tail into `S.messages`. So in the steady-state render the flag is present and the guard fires correctly for the streaming row.

The gap is the very first frame: the fork button is only rendered on message hover, and the live row is normally the last message, so in practice the user can't hover-and-click the streaming row before the merge stamps `_live`. But if `_live` is ever momentarily absent on that last row (e.g. a render between the token arriving and `_ensureInflightLiveAssistantMessage` running), the guard would fall through and attempt to fork including the partial row. Since `absoluteKeepCount = _oldestIdx + msgIdx` would then include the live index, the server would copy a partial assistant message into the fork.

A cheap belt-and-suspenders: also treat "clicked message is the last one while `S.busy`" as the live case, e.g.

```js
if(S.busy){
  const _isLastMsg = msgIdx >= (Array.isArray(S.messages)?S.messages.length:0);
  const _msg=(Array.isArray(S.messages)&&S.messages[msgIdx-1])||null;
  if((_msg&&_msg._live) || _isLastMsg){
    showToast('Cannot fork a message still being generated.',3000);
    return;
  }
}
```

That closes the flag-timing window without blocking any genuine historical fork (past messages are never the last index during streaming).

### Test plan

Worth adding a JS-level assertion (the fork tests already stub `S`): with `S.busy=true` and a trailing `{role:'assistant',_live:true}` row, `forkFromMessage(len)` shows the toast and does **not** call `/api/session/branch`; with the same busy state, `forkFromMessage(2)` on a committed past message *does* call branch with the expected `keep_count`. That pins both the block and the allow path.

Net: fix is correct and the indexing is safe. The `_live`-timing belt-and-suspenders is optional hardening, not a blocker.

kertisalex:
Thanks for the review. I've pushed the follow-up fix: > - Added _pending:true to the optimistic user message (messages.js:1603) > - Extended the guard to reject _msg._live || _msg._pending and added the belt-and-suspenders _isLastMsg check > > Could you re-review?

nesquena-hermes:
Shipped in **exp-v0.52.59** 🚀 — 'Fork from here' now works during an active response for past committed messages; the streaming message + just-sent uncommitted prompt are correctly blocked with a toast. Codex SAFE (round 2 — off-by-one fixed, `_isLastMsg` busy-gated, `_pending` not carry-forwarded), stream-gate GREEN all modes (the hide_all RED was the known load flake — isolated re-run clean), 99 tests. Thanks @kertisalex! Closes #5993.

PR #5995, opened 2026-07-12
Add composer fast mode follow-ups

bc-hermes:
Implements a focused tracker `bc-hermes-ops/omnibus-tracker#39` slice for the Fast Hermes WebUI conversation prototype.

Changes:
- Adds a normal-composer Fast pill that persists per-browser, preserves the user's visible message, and starts a durable background follow-up after the foreground turn starts.
- Threads a `fast_mode` request flag through `/api/chat/start` into legacy and gateway streaming paths, adding foreground guidance as ephemeral/system context rather than changing the saved user message.
- Exposes the new composer/background capabilities in `/api/fast/health` and adds focused regression coverage.

Verification:
- `python3 -m py_compile api/fast_mode.py api/routes.py api/streaming.py api/gateway_chat.py`
- `python -m pytest tests/test_fast_mode_health.py tests/test_background_durable_tasks.py tests/test_issue3996_sse_visibility.py -q` → 21 passed
- Synthetic isolated WebUI HTTP validation: `/api/fast/health` enabled, foreground stream completed, durable background task reached `done`, parent transcript contained the background result, and the saved user message stayed unchanged.


nesquena-hermes:
## Review (Codex) — solid durable-task design; one concrete client-side clobber in the fast-mode flow

Read `api/fast_mode.py` (new), the durable store + parent-projection in `api/background.py:50-370`, the streaming drain hook (`api/streaming.py:9500-9506`), the gateway/legacy guidance threading, and the frontend changes in `static/messages.js`. The server-side deferral design is careful and correct; I found one client-side race that the fast-mode path triggers on essentially every fast send.

### What's good — the server defers parent-transcript writes correctly

`append_or_queue_background_parent_update` (`api/background.py:295-301`) checks `_parent_has_live_turn` and, when the parent turn is live, records `card_pending` / `result_pending` in the durable sidecar instead of mutating `messages` — then `drain_pending_background_parent_updates` materializes them at the finalize boundary after `active_stream_id` is cleared:

```python
if _parent_has_live_turn(parent):
    _mark_durable_parent_append(parent_sid, str(task_id), {
        "parent_append_status": pending_status, ...
    })
    return pending_status
```

The drain call at `streaming.py:9503` runs after `s.active_stream_id = None` (`:9225`) and after the foreground assistant merge, so `_parent_has_live_turn(s)` is False by then and the card lands without overwriting the just-completed turn. The atomic `os.replace` write in `_write_durable_tasks_unlocked` and the `is_safe_session_id` guard on the sidecar path are both right.

### The bug: `_reloadCurrentSessionAfterBackgroundUpdate` clobbers the live foreground tail

`static/messages.js:7401` unconditionally replaces `S.messages` from the server transcript and re-renders, with no in-flight-stream guard:

```js
async function _reloadCurrentSessionAfterBackgroundUpdate(sid) {
  if (!sid || !S.session || S.session.session_id !== sid) return;
  const data = await api('/api/session?session_id=' + encodeURIComponent(sid));
  ...
  S.messages = data.session.messages || [];      // <-- overwrites live tail
  S.session = Object.assign(S.session || {}, data.session);
  renderMessages({preserveScroll:true});
}
```

This fires from the `background_task_updated` SSE handler (`_handleBackgroundTaskUpdatedEvent`, `:7414`). The problem is *when* it fires: `track_background` emits `background_task_updated` **synchronously** the moment `/api/background` is POSTed (`api/background.py:515-516`), and in the fast-mode flow that POST happens while the foreground turn is still streaming — `_launchFastModeBackground(activeSid, ...)` is invoked right after `attachLiveStream` in `send()` (`messages.js:1948`). So the "running" card emit arrives mid-stream on the same session channel.

During a live turn the streaming assistant tail lives in `INFLIGHT[sid]` and is merged into `S.messages` only on the session-**load** path (`sessions.js:2042-2069` via `_ensureInflightLiveAssistantMessage` + `_mergeInflightTailMessages`). `renderMessages()` itself reads `S.messages` directly (`_getVisibleMessagesWithIdx`, `ui.js:630`) and does not re-merge INFLIGHT. So slamming `S.messages = data.session.messages` and calling `renderMessages()` bypasses the merge and reverts the visible transcript to the last server checkpoint — the in-progress foreground answer flickers/disappears until the next token render heals it. This is the exact clobber the load path's "restore persisted transcript first, then merge INFLIGHT tail" comment (`sessions.js:2028-2031`) was written to avoid.

### Suggested fix (client-side)

Guard the reload on an active stream and defer to the normal live pipeline while streaming — the durable card/result already survives in the sidecar and drains at finalize, so a mid-stream DOM reload isn't needed:

```js
async function _reloadCurrentSessionAfterBackgroundUpdate(sid) {
  if (!sid || !S.session || S.session.session_id !== sid) return;
  // Don't clobber the live foreground tail; the drain hook + load-path
  // INFLIGHT merge will surface the card once the turn settles.
  if (S.busy || S.activeStreamId || (typeof INFLIGHT!=='undefined' && INFLIGHT[sid])) return;
  ...
}
```

`S.busy` / `S.activeStreamId` are the idioms already used for this exact "is a turn live" question throughout `messages.js` (e.g. `:1195`, `:2073`).

### Test plan

A JS-level test with `S.busy=true` + a live INFLIGHT tail: dispatch a `background_task_updated` event and assert `S.messages` is untouched and `/api/session` is **not** fetched; then with `S.busy=false`, assert the reload fetches and re-renders. The Python side (`tests/test_background_durable_tasks.py`) already covers the pending→drain projection well — worth adding one asserting a `running` emit during a live parent turn leaves `messages` unmutated (queued as `card_pending`), which your server code already does correctly.

Net: the durable-task backbone is well built and the parent-write deferral is right. The single blocking item is the unguarded client reload firing mid-foreground-stream — a ~3-line guard closes it.

bc-hermes:
Addressed the applicable review findings in one follow-up commit (`14a82d12`):

- defer background-task transcript reloads while the foreground stream/INFLIGHT tail is live;
- preserve Fast mode for attachment-only sends by using the finalized message text;
- parse request booleans strictly instead of treating non-empty false-like strings as true;
- enforce the deployment capability on both server and client, including stale persisted browser toggles.

The two cross-session findings are already covered by the shared request-level `session_id` profile-visibility guard used by these GET and POST routes, so I did not add a second endpoint-specific authorization layer.

Verification: focused test set 22 passed; JavaScript syntax, Python compile, and diff checks passed.

nesquena-hermes (dismissed):
## 🔴 Authoritative current-master gate — RED

**Contributor head:** `14a9ad129520566277e75a6e4d1e4a8c1e46ed16`

I integrated the PR onto current master, preserving both current `external_runtime_owned` authority and the PR's `fast_mode` parameter. Threat scan is CLEAN; 80 authored/current completion tests pass.

### Blocking failures

1. **Cancel destroys a completed answer.** A cancel request reads `running`, blocks in `cancel_stream`, the worker commits `done`, then cancel resumes and overwrites it with `cancelled` and `(background task cancelled)`. The terminal compare/write is not atomic.
2. **Restart recovery does not exist.** Workers are daemon threads; after restart persisted tasks remain `running` forever. `interrupted_by_restart` is handled but never written.
3. **The sidecar name collides with session storage.** `{sid}.background_tasks.json` matches `SESSION_DIR/*.json`. Sandbox: `sidecar_loaded_as_session=true`, phantom `Untitled`; its `parent_session_id` also makes `_has_compression_continuation` return true and can suppress stale-turn crash repair.
4. **Profile `.env` opens the deployment gate.** Corrected sandbox: `profile_dotenv_fast_enabled=true`, `mode=real_model_prototype`. Both Fast keys need startup ownership/protection.
5. **Gateway foreground turns never drain pending parent projections.** Local streaming calls `drain_pending_background_parent_updates`; gateway finalization does not. Results completed while a gateway turn is live can stay sidecar-only and invisible.

Additional lifecycle gaps: raw hidden-session unlink leaves tail-cache/index state; terminal tasks/full answers have no retention cap; 32-bit task IDs are collision-prone; failures are stored as successful `done`; terminal cards permanently say “Background task running…”; cancel has no UI; durable/client copy is English-only; disabled Fast is permanent dead composer chrome with sub-44px targeting.

### Required rework

- Linearize cancel/complete under one lock/state transition.
- Reconcile orphan `running` tasks at startup/lazy load and project `interrupted_by_restart`.
- Move task sidecars to an underscore-prefixed subdirectory; clean them with parent deletion; add retention and full UUIDs.
- Snapshot/protect both deployment env keys.
- Drain at local **and gateway** turn finalization.
- Use the canonical locked hidden-session deleter, truthful terminal states/cards, localized copy, and a real cancel/status renderer.

Codex and Opus: KICK-BACK. Fable: UX-REWORK. Full suite/browser gate deferred until these data-loss and durability boundaries are repaired.

nesquena-hermes:
Publication-race correction: the full RED gate above was performed against exact contributor head `14a9ad129520566277e75a6e4d1e4a8c1e46ed16`, integrated onto then-current master. The contributor pushed `14a82d125f151f2c72c691c4eee50492ec4c467f` while the review was being submitted, and GitHub attached the review to that new SHA. I dismissed it so the untested head is not suppressed.

Historical exact-head evidence for `14a9ad…` remains RED: restart ghosts, cancel/complete overwrite, session-scanner sidecar collision, profile-dotenv gate bypass, and missing gateway drain were reproduced. The new head must settle for the normal 1–2 hour window and receive a fresh exact-head re-gate; none of those findings are assumed to carry forward without verification.

nesquena-hermes (changes requested):

**Contributor SHA:** `14a82d125f151f2c72c691c4eee50492ec4c467f`

This head predates the broad durability review and does not fix its five blocking invariants. I re-integrated the exact contribution on current master; `CHANGELOG.md` and `api/routes.py` conflict, and the current clean-applied files retain each failure:

1. **Cancel can overwrite a completed answer.** `api/background.py:441-463` reads `running`, calls `cancel_stream()` outside `_lock`, then unconditionally writes `cancelled`. Completion may commit `done` while cancellation blocks; cancellation still replaces the answer.
2. **Restart ghosts remain.** `interrupted_by_restart` exists only in the terminal-status set. No startup/lazy-load reconciliation writes it, so persisted `running` tasks survive a restart forever.
3. **The sidecar still collides with session discovery.** `_task_store_path()` remains `SESSION_DIR / f"{sid}.background_tasks.json"`; it matches the session scanner's `*.json` surface and can create phantom sessions / false continuation ownership.
4. **Profile `.env` still controls the deployment gate.** `api/fast_mode.py:29-40,62-63` reads both Fast keys from live `os.getenv()` on each request. Neither operator-owned key is startup-snapshotted/protected.
5. **Gateway finalization still does not drain deferred parent projections.** Local streaming imports/calls `drain_pending_background_parent_updates`; `api/gateway_chat.py` has no such import or call. A result that completes during a gateway foreground turn can remain sidecar-only.

The current follow-up correctly adds strict request booleans, attachment-finalized text, deployment-aware client gating, and an active-stream reload guard. Those are useful but orthogonal to the five reproduced durability failures.

**Required rework remains:** linearize cancel/complete under one transition lock; reconcile restart orphans; move/clean/cap durable sidecars outside `SESSION_DIR/*.json`; snapshot/protect both Fast deployment keys; drain deferred parent projections after both local and gateway turns. Rebase on current master and add discriminating regressions for each path.

**Verdict:** RED for `14a82d125f151f2c72c691c4eee50492ec4c467f`.

Issue #433, opened 2026-04-14
Stale active session model can break chats after switching to a custom endpoint


shaoxianbilly:
## Summary

When switching Hermes WebUI to a `custom` OpenAI-compatible endpoint, the active session can keep a stale model id from a previous provider/session (for example `openai/gpt-5.4`). The UI may show the correct custom model (`gpt-5.4`), but sending a message still uses the stale `session.model`, so the request is sent to the custom endpoint with the wrong model id and fails.

I verified this against current `origin/master` (`37850a4`, `v0.50.36`).

## Reproduction

1. Open or restore a session whose stored model is provider-prefixed, e.g. `openai/gpt-5.4`.
2. Reconfigure WebUI to use `provider: custom` with a `base_url` pointing to an OpenAI-compatible relay.
3. Use a relay whose `/v1/models` returns bare ids such as `gpt-5.4`.
4. Confirm the WebUI model dropdown now shows the bare custom model (`gpt-5.4`).
5. Send a message in the already-open session.

## Actual behavior

The chat request still uses the old session model, e.g. `openai/gpt-5.4`, and the custom endpoint rejects it.

From request dumps in my local repro:

- request body model: `openai/gpt-5.4`
- endpoint error: `Model not supported: openai/gpt-5.4`

From the current code path:

- `static/messages.js` prefers `S.session.model || $('modelSelect').value`
- `/api/chat/start` persists `body.get("model") or s.model` directly into the session
- `resolve_model_provider()` keeps slash-bearing model ids unchanged when `config_base_url` is set

So once a stale slash model is present in the session, it continues to be sent to the custom endpoint unchanged.

## Expected behavior

One of these should happen:

1. After onboarding/profile/provider change, the active session's model should be updated to the new valid model.
2. If the active session model is stale/incompatible with the current provider, WebUI should invalidate it and fall back to the selected dropdown/default model.
3. In `custom + base_url` mode, stale cross-provider ids like `openai/gpt-5.4` should be normalized or rejected before sending.

## Notes

- This is not just a relay catalog issue. In my repro, `GET <base_url>/v1/models` returned bare ids (`gpt-5.4`, etc.), not `openai/gpt-5.4`.
- `v0.50.31` changed live model fetching to delegate to `provider_model_ids()`, but this stale-session/custom-routing issue still appears to remain on `v0.50.36`.
[...]

PR #5999, opened 2026-07-13
fix(#5979 follow-up): preserve custom-proxy vendor namespace on cold-catalog sends

### LINKED ISSUE #5979: P0 regression: model selector strips the vendor namespace before sending — 'x-ai/grok-4.5' sent as bare 'grok-4.5', breaks routing on custom/proxy providers (exp-v0.52.47)
**P0 / "dead in water" regression** reported by **b3nw** in Discord `#report-bugs` (Jul 12 2026) on **exp-v0.52.47 · Experimental**.

> "regression … has me dead in water 😢 — model selector is trimming the model name when it sends to the provider, causing routing to break on my custom setup."

## Symptom

The user runs multiple upstream vendors behind a **single custom proxy** (`provider: custom`), and selects fully-qualified vendor-namespaced model IDs like **`x-ai/grok-4.5`**. The proxy routes on that full `x-ai/…` prefix. On exp-v0.52.47 the WebUI **strips the vendor namespace before sending**, so the proxy receives the bare `grok-4.5` and can't route it:

```
⚠️  API call failed (attempt 1/3): BadRequestError [HTTP 400]
   🔌 Provider: custom  Model: grok-4.5
   🌐 Endpoint: https://<redacted-proxy>/v1
   📝 Error: HTTP 400: {"detail":"Invalid Request: Invalid model format or no credentials for provider: grok-4.5"}
   ⏱️  Elapsed: 0.18s  Context: 154 msgs, ~154,356 tokens
```

## Direct evidence (from the model-selector screenshot)

- The dropdown option is the **full ID `x-ai/grok-4.5`** (shown SELECTED), with `x-ai/grok-4.5` as both the bold label and the sub-label.
- But the **composer footer chip shows the trimmed `grok-4.5`**, and that trimmed value is what goes on the wire (`Model: grok-4.5` in the error).

So the selected model value (`x-ai/grok-4.5`) is correct, but a **vendor-namespace strip (`split('/').pop()`) is bleeding from the display layer into the value that's sent**.

## Very likely the root cause behind #5940

#5940 (filed from b3nw's exp-v0.52.38 report) was the *misleading error message* — WebUI said "No response from provider / try again" while the console showed the real HTTP 400 `Invalid Request: Invalid model format or no credentials for provider: <redacted>`. That redacted provider was **this same `grok-4.5`**. In other words:

- The **model-trim-on-send bug here is what actually broke the turn** (the 400).
- #5940 was the *symptom* — the WebUI masking that 400 with a generic "rate limit, try again" message.
- The **#5940 fix shipped in exp-v0.52.47** (surface the real provider error) is precisely what let b3nw finally *see* `Model: grok-4.5` and identify the real cause.

So this is likely not brand-new to .47 — .47 just made the pre-existing trim bug visible. Worth bisecting how far back the trimmed value has been going on the wire.

## Investigation notes (candidate sources)

The vendor-namespace strip `split('/').pop()` is used in several **display** spots (e.g. `static/ui.js:7799` model-badge label; `static/panels.js:6622/6801` profile-card meta; the `_normId` dedup in `static/ui.js:~3284`). The frontend send path *appears* to preserve the full value:
- `_modelStateForSelect` (`static/ui.js:2672`) returns `{model:value,…}` with the full option value.
- `sessions.js:1263` treats `x-ai/grok-4.5` as a **non-bare** model (has `/`), so the bare-model repair shouldn't fire.

So the trim is either (a) a display helper being reused to compute the *sent* value somewhere on the send/update path, or (b) a server-side normalization stripping the `x-ax/` namespace for `provider: custom` before dispatch. The wire error (`Model: grok-4.5`, `Provider: custom`) says the bare form reached the provider call — trace from `/api/chat/start` model resolution through the custom-provider dispatch and confirm whether the namespace is dropped client-side (value sent) or server-side (normalization).

## Related design point from the reporter (same thread)

> "I'm not sure why we trim the model name for the UI at all, it's not really desirable for anyone with multiple providers behind a single proxy, it obfuscates which provider you're routing through."

For single-proxy/multi-vendor setups, the `x-ai/` (etc.) prefix is **meaningful routing information**, not noise. Consider **not** trimming the vendor namespace in the selector/chip display for custom/proxy providers (or making it configurable), so users can see which upstream they're routing to. At minimum the trim must never affect the value sent to the provider.

## Acceptance criteria
1. Selecting `x-ai/grok-4.5` (or any `vendor/model`) on a `custom`/proxy provider sends the **full** `x-ai/grok-4.5` on the wire — no namespace strip.
2. The composer chip / model badge display trim (if kept) is display-only and never mutates the sent model id.
3. Consider preserving the vendor prefix in the display for custom/proxy providers so routing is legible.

## Environment
- WebUI: exp-v0.52.47 (experimental)
- Setup: multiple vendors behind one custom OpenAI-compatible proxy (`provider: custom`), vendor-namespaced model IDs

## Related
- #5940 — misleading error message (this trim is the underlying cause it was masking; .47's #5940 fix surfaced this)
- #5924 — post-failure recovery model-switch (adjacent model-resolution area)
- #2761 / #1240 — provider/model source-of-truth family

nesquena-hermes:
## Reopened #5979 — the cold-catalog case

The earlier fix (#5980, exp-v0.52.53) handled **warm-catalog** and **config-declared** ids. @b3nw confirmed truncation still happens on exp-v0.52.55/56 for a model he **selects in the picker that isn't his `model.default`**, on a **cold** catalog.

### Root cause (reproduced on master; confirmed by Codex + Fable)
b3nw's config: `provider: custom:llm-proxy`, `model.default: x-ai/grok-4.5`, **no `models[]` allowlist**.
- His default `x-ai/grok-4.5` → preserved (config-declaration path) ✓
- A picked non-default like `x-ai/grok-composer-2.5-fast` → **truncated to `grok-composer-2.5-fast`** on every cold send ✗

The send path (`api/streaming.py:7857`) resolves without warming the catalog, and the #1855 chat/start fast path deliberately skips the catalog when a session already has a persisted model+provider — so "cold at send" is the *designed* state after any restart / settings-save / cache-expiry. With provenance `None` and no config declaration, the #5980 tri-state fell through to the **legacy family strip**, and `grok-composer-2.5-fast` had graduated into the x-ai first-party catalog → stripped.

### Fix (two changes)
1. **Flip the custom-arm cold default from family-strip to preserve-verbatim** (`api/config.py`). For an explicit `custom` / `custom:<slug>` proxy the strip now requires **positive provenance** (endpoint advertises only the bare id — the #433 relay case). Honours `_endpoint_advertised_model_ids`'s documented `None → preserve` contract; removes the data-driven flaw where a model graduating into the static catalog silently flips routing.
2. **Network-free provenance warm on the send path** (`warm_models_catalog_provenance_if_cold`, called before the resolve). `prefer_cache=True` republishes provenance from the durable disk cache — no live probe, no `_cfg_lock` held on the worker thread (no lock-order deadlock) — so #433's bare-only-strip stays exact whenever a disk cache exists.

### Residual (accepted; strictly better than master)
A #433-shaped wrong-preserve now needs memory cold **AND** disk cache absent/invalid (first run, post-config-save window, schema/version bump). It fails **loudly** and heals on the next `/api/models` rebuild. Under master b3nw's failure was the mirror image but **silent, recurring every send, unfixable** short of declaring every model in config.

### Supersedes #5996
rodboev's #5996 targets the `providers.<slug>` arm and does **not** cover b3nw's actual `custom:llm-proxy` config (confirmed by Codex). This PR covers the real case.

### Tests
- Rewrote the cold-fallback test (strip → preserve)
- Sharpened the foreign-profile test so trusted-strip vs ignored-preserve genuinely diverge
- Added b3nw's exact cold non-declared `custom:slug` case
- Added 3 warm-helper tests (publishes-from-disk, noop-when-warm, never-live-rebuilds)

64 resolver+sprint40 tests pass; b3nw end-to-end verified; #433 positive-provenance strip verified.

Closes #5979. Thanks @b3nw for the CLI-vs-WebUI comparison that localized it.


nesquena-hermes:
**Correction re #5996:** I initially framed this as superseding @rodboev's #5996 — that's wrong. Verified they're **complementary**, not duplicate:

- **This PR (#5999)** fixes the `custom` / `custom:<slug>` arm (b3nw's actual `provider: custom:llm-proxy` config) — the cold-catalog preserve inside the `_is_custom` branch.
- **#5996** fixes the `providers.<slug>` arm (a *named* user-defined provider, where `_is_custom` is False and control hits the separate non-custom `base_url` branch at `config.py:2914`).

My cold-preserve flip lives inside `if _is_custom:`, so it does **not** cover #5996's `providers.<slug>` case (confirmed: `providers.llm-proxy` + cold still truncates on this branch). Both are needed for full proxy coverage. Will coordinate merge order with @rodboev's PR rather than closing it.

PR #6003, opened 2026-07-13
Release exp: gateway terminal errors + cross-session completion routing (#5969, #6002)

### LINKED ISSUE #5940: WebUI error reporting masks the real provider error: shows 'No response from provider / try again' when the agent hit a non-retryable HTTP 400 (invalid model / no credentials)

nesquena-hermes:
Two backend reliability fixes, both Codex SAFE:

- **#5969** (@rodboev, closes #5940) — gateway terminal provider errors now classified + surfaced as the real error (not generic empty turn), session_id-bound, and the in-flight prompt + partial output persist through reload including a repeated prompt (`is_exact_checkpoint`). 4-round convergence. Codex SAFE + stream-gate GREEN all modes.
- **#6002** (@soria-claw) — detached terminal/subagent completions carry their exact `origin_ui_session_id` (context-local, reset on exit), closing the cross-session wakeup misroute; legacy owner-less events keep same-session delivery, never broadcast. Codex SAFE, 100/100 concurrent-stress rounds.

55 targeted tests pass.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -159,12 +159,17 @@ Per-request environment variables (set by chat handler, restored after):
     HERMES_EXEC_ASK      Set to "1" to enable approval gate for dangerous commands.
     HERMES_SESSION_KEY   Set to session_id. The approval tool keys pending entries
                          by this value, enabling per-session approval state.
+    HERMES_UI_SESSION_ID Bound context-locally to the browser session_id. Detached
+                         process and subagent completions capture it as their exact
+                         return address.
     HERMES_HOME          Set to the active profile's directory before running agent.
                          Saved and restored around each agent run.
 
-WARNING: These env vars are process-global. Two concurrent chat requests will clobber
-each other. This is safe only for single-user, single-concurrent-request use.
-See Architecture Phase B for the fix.
+The process-global environment mirror remains for compatibility with older tool
+consumers, but completion routing identity is bound through Hermes contextvars for
+the full turn. Modern completion events carry ``origin_ui_session_id``; WebUI
+treats that exact owner as authoritative and uses the session-key index only for
+legacy events that do not provide it.
 
 ---
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -159,9 +159,11 @@ Per-request environment variables (set by chat handler, restored after):
     HERMES_EXEC_ASK      Set to "1" to enable approval gate for dangerous commands.
     HERMES_SESSION_KEY   Set to session_id. The approval tool keys pending entries
                          by this value, enabling per-session approval state.
-    HERMES_UI_SESSION_ID Bound context-locally to the browser session_id. Detached
-                         process and subagent completions capture it as their exact
-                         return address.
+    HERMES_UI_SESSION_ID Bound context-locally to the browser session_id. Async-
+                         delegation/subagent completions capture it as their exact
+                         return address. Detached terminal completions currently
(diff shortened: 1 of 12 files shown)


nesquena-hermes:
Unbundling — #6002 has a test-isolation failure in the full sharded suite (test_xsession_wakeup_misroute + test_notify_on_complete_webui fail on shards 0/3: 'assert [] == [webui-session-a]'), while #5969 is clean. Re-shipping #5969 solo; #6002 bounced back for the isolation fix.

PR #6009, opened 2026-07-13
Release exp: cross-session subagent completion routing (#6002)

nesquena-hermes:
Ships #6002 (@soria-clawd-bot). Detached terminal/subagent completion events now carry their exact `origin_ui_session_id` (context-local, reset on exit), so a finished background task routes only to its origin browser session — closing the cross-session wakeup misroute. Owner-less events keep same-session delivery and are never broadcast. Codex SAFE (round 2 — impl unchanged from the safe round, tests re-aligned to genuinely assert the exact-owner contract, 100-round concurrent stress zero leaks), stream-gate GREEN all modes (the hide_all RED was the known load flake — isolated re-run clean), 18/18 regression tests pass together (the earlier stale-source-assertion break is resolved).

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -159,12 +159,17 @@ Per-request environment variables (set by chat handler, restored after):
     HERMES_EXEC_ASK      Set to "1" to enable approval gate for dangerous commands.
     HERMES_SESSION_KEY   Set to session_id. The approval tool keys pending entries
                          by this value, enabling per-session approval state.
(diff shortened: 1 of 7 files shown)


nesquena-hermes:
Closing — #6002's own tests (test_xsession_wakeup_misroute.py:365) fail in the full shard-0 suite (assert started == ['webui-session-a'] → got []), though they pass in isolation. Real test-isolation dependency; not shipping until it's resolved. Bouncing #6002 with the finding.

Issue #3802, opened 2026-06-07
Deleting a conversation via UI leaves turn journal and run journal files on disk


ZombieLurker:
# Deleting a conversation via UI leaves full session journal files on disk

## Summary

When I delete a conversation from the WebUI sidebar using the trash icon, the session disappears from the UI and the main session JSON file is removed. But the turn journal and run journal files are left behind on disk, meaning the full conversation history is still recoverable.

## What gets deleted

- Session entry in `_index.json` is removed
- Main session JSON file (`sessions/{sid}.json`) is deleted
- Session row and messages are removed from `state.db`

## What is NOT deleted

- Turn journal: `sessions/_turn_journal/{session_id}.jsonl`
- Run journal: `sessions/_run_journal/{session_id}/` directory with multiple `.jsonl` files

## Steps to reproduce

1. Start a new conversation in the WebUI
2. Send a few messages
3. Delete the conversation using the "Delete Conversation" trash icon in the 3 dot menu of chat in the sidebar
4. Check the filesystem:

```bash
$ ls sessions/_turn_journal/
<session_id>.jsonl

$ ls sessions/_run_journal/
<session_id>/
```

The turn journal file is still present and contains all user messages in plaintext. The run journal directory contains the full request/response payloads for the session.

## Expected behavior

`DELETE /api/session/delete` should clean up all files associated with the session, including the turn and run journals.

## What the leftover files contain

The turn journal (`_turn_journal/{sid}.jsonl`) stores each turn as a JSON line with the full user message content. For a session with ~15 messages, the file is a few tens of KB.

The run journal (`_run_journal/{sid}/`) contains one `.jsonl` file per stream, each with the complete LLM request and response body including system prompt, all messages, tool definitions, and assistant outputs. A moderate session leaves ~1–2 MB of files.

## Suggested fix

In the delete handler, add cleanup for the journal files alongside the existing cleanup:

```python
import os, shutil

turn_journal_path = os.path.join(SESSION_DIR, '_turn_journal', f'{session_id}.jsonl')
run_journal_path = os.path.join(SESSION_DIR, '_run_journal', session_id)

if os.path.exists(turn_journal_path):
    os.remove(turn_journal_path)
if os.path.isdir(run_journal_path):
    shutil.rmtree(run_journal_path)
```

## Notes

[...]

PR #6011, opened 2026-07-13
perf(session): persist composer drafts to a sidecar; compact + no-op saves

ai-ag2026:
## Summary

A long-running session whose sidecar JSON grew to ~25 MB (3600+ messages) made the WebUI almost unusable. The 400 ms-debounced composer-draft autosave persists through `Session.save()`, which rewrites the **entire** session file, so every keystroke's `POST /api/session/draft` serialized the full multi-MB payload behind the per-session lock. Under a burst of debounced saves the queued POSTs starved the whole server — observed draft POSTs at **27–30 s each**. (The existing `[SLOW] /api/session/draft` instrumentation already points at this path.)

This makes drafts cheap and trims the remaining full-file writes, with no new hard dependency and no behavior change for normal-size sessions.

## Changes (server-side only)

- **Draft sidecar** — composer drafts persist to `SESSION_DIR/_drafts/<sid>.json` (tiny atomic write) instead of `Session.save()`. The draft POST no longer loads or rewrites the transcript at all: only a metadata-prefix load for validation plus a small file write, under a **dedicated per-session draft lock** so drafts never queue behind a running turn. The in-file `composer_draft` remains a read fallback; the sidecar wins (an empty sidecar = a cleared draft, so it correctly overrides a stale legacy value). Duplicate/delete paths and `compact()` resolve through the sidecar.
- **Compact JSON** — `Session.save()` drops `indent=2` (byte-compatible for every reader; `json.loads` and the metadata-prefix scanner are whitespace-agnostic). ~2× faster serialize, smaller file. Uses **orjson when present with a stdlib fallback** — orjson is *not* added as a dependency; the fallback also covers legacy files containing the non-standard `NaN` literal.
- **No-op save skip** — byte-identical re-saves (idle rewrites observed every ~15–30 s on a large session) skip the disk write when the file is untouched since this object's last write (payload digest + `stat` compare). `touch_updated_at=True` saves always differ, so recency semantics are unaffected.
- **#1558 backup-guard fast path** — grow-the-conversation saves read only the metadata-prefix `message_count` instead of full-parsing the multi-MB file; the full read happens only on a shrink (which needs the text for the `.bak` anyway) or a legacy/corrupt layout. Composes with the #5854 prefix layout that already moved `message_count` ahead of the large scene bodies.

## Why it's safe

- **Lossless**: verified a real 25 MB session round-trips `reparsed == original` through the new codec (no `NaN`/`Inf`); compact output parses identically via `load()` and `load_metadata_only()`.
- **No new dependency**: the entire test suite runs on the **stdlib fallback** (orjson absent in the dev venv) and is green; production simply gets the faster path when orjson is installed. Both paths emit byte-identical compact JSON.
- **#1558 preserved**: the shrink → `.bak` snapshot and the "refuse to overwrite populated messages with an empty active/pending snapshot" guard are unchanged; only the *fast path to reach them* is added.

## Measured

On a real 24.5 MB / 3942-message session:

| operation | before | after |
|---|---|---|
| draft persist | 322 ms full `Session.save()` | **7.8 ms** sidecar write (session file untouched) |
| idle re-save (unchanged) | full serialize + 24 MB write + fsync + 24 MB backup read | **skipped** (digest+stat) |

~**41×** on the draft write in isolation; far more under the concurrent debounced-autosave load that produced the original 27–30 s stalls.

## Validation

- New `tests/test_bigsession_save_hotpath.py` (8): compact-write round-trip via `load()`/`load_metadata_only()`, backup-guard grow-vs-shrink, empty-overwrite refusal, no-op skip incl. deferral to an external writer, draft-sidecar precedence/roundtrip, unsafe-id rejection, and that sidecars stay out of the top-level `*.json` session scan.
- `test_stage326` draft assertions updated to the sidecar persist site; one `test_issue5572` assertion switched from substring-matching indented JSON to parsing the payload (compact output no longer contains `"messages": []` with a space — same semantics).
- Targeted draft/#1558/recovery suites, the #5854 anchor-scene suite, and a broad `session/save/draft/metadata` sweep are green (the only failures in the sweep are a pre-existing test-ordering flake and an unrelated `agent`-package import error in the dev venv).


nesquena-hermes (changes requested):
## 🔬 Gate review (Codex) — SHIP ONLY WITH FIXES (perf win proven; 4 draft-loss/orphan edges)

Excellent perf work @ai-ag2026 — the starvation fix is **real and verified**: a 26 MB session now uses metadata-only load (~1 ms) + a tiny sidecar write (~8 ms), the main session file is untouched, and persisted-session restart round-trip, legacy fallback, sidecar precedence, atomic replace, concurrent partial-field merge, and scan-exclusion (no phantom sessions) all check out. 30 draft + 21 metadata + 2 messaging-clear tests pass. But four draft-loss / orphan edges block shipment (draft loss = user-facing data loss, so these need to land first):

1. **[DRAFT-LOSS — draft-only new chat not restart-durable] `api/routes.py:14426`** — a brand-new memory-only chat that's never been saved writes the draft to the sidecar but has **no top-level session record**, so on restart the draft is orphaned/lost. **Fix:** persist a minimal top-level session record on the first non-empty draft for an otherwise memory-only new chat.

2. **[DRAFT-LOSS — compression rotation orphans the draft] `api/routes.py:14393` + `api/streaming.py:8724`** — draft ownership isn't serialized with compression SID rotation, so a draft written to the **old** SID's sidecar is stranded after the continuation SID takes over. **Fix:** serialize draft ownership with compression rotation and migrate/redirect old-SID sidecar writes to the continuation SID.

3. **[ORPHAN RACE — delete/draft-write TOCTOU] `api/routes.py:14548` + `api/routes.py:20237`** — reproduced: a draft POST landing just after a session delete recreates the sidecar (`session_exists=False` but `orphan_sidecar_exists=True`), leaking an orphan `_drafts/<sid>.json`. **Fix:** revalidate-and-delete the sidecar **under the draft lock**, including the bulk-cleanup paths.

4. **[STALE-DRAFT — clear incomplete] `api/routes.py:14417`** — clear doesn't empty both `text` and `files`, doesn't durably clear the legacy `composer_draft` fallback, and can race a newer draft. **Fix:** canonicalize clear to empty text+files, durably clear the legacy field on the next mandatory session save, then unlink the still-empty sidecar under the draft lock.

The core design (sidecar + decoupled lock + legacy read-fallback + scan-exclusion) is right and the perf outcome is exactly what's needed — these are the durability/lifecycle corners (new-chat persistence, compression rotation, delete race, clear semantics). Once drafts survive restart + rotation and the delete race is closed under the lock, I'll re-gate + ship. Nice work — this is a genuinely valuable fix for large-session users.

nesquena-hermes (changes requested):
## 🔬 Re-gate (Codex, adversarial — self-rebased clean off master) — KICK BACK TO AUTHOR (perf win real, but all 4 draft-loss edges still reproduce + a new recovery-reader regression)

Thanks @ai-ag2026 — the perf win is genuinely verified again (metadata-only route avoided the full load, left the 2 MB main file byte/stat-identical, wrote a 67-byte sidecar, skipped the identical rewrite), `orjson` is truly optional (absent from the clean venv + requirements, stdlib round-trip works), cross-session isolation is SID-keyed and path-safe, and `_drafts/` scan-exclusion holds. But this round's commits touched the four flagged areas without actually closing them — I reproduced **all four through the real route/storage code**, plus a new regression:

1. **[DRAFT-LOSS — new chat restart] still FAIL** — a draft-only memory-only chat writes the sidecar but still persists **no top-level session record**; probe: `draft_sidecar_exists=true, main_session_exists=false, restart_session_loads=false` → draft orphaned/lost on restart. Persist a minimal top-level record on the first non-empty draft.

2. **[DRAFT-LOSS — compression rotation] still FAIL** — a draft on the OLD sid's sidecar is still stranded after the continuation SID takes over. Migrate/redirect draft ownership on rotation.

3. **[ORPHAN — delete/write TOCTOU] still FAIL** — a draft POST just after delete still recreates an orphan sidecar, and the **bulk cleanup path leaks sidecars** too. The revalidate-and-delete must happen under the draft lock on all paths.

4. **[STALE-DRAFT — clear] still FAIL** — clear still lets attachments/stale drafts survive and a newer draft can be clobbered. Canonicalize clear (empty text+files, durably clear the legacy field, unlink under the lock, don't race a newer draft).

5. **[NEW REGRESSION — compact serialization breaks a recovery reader] `api/models.py:3429-3443`** — the compact serialization changes the on-disk shape such that the existing `parent_session_id` recovery reader, which does a whitespace-sensitive **substring** match (`needle = '"parent_session_id": "{sid}"'`, `if needle in head`), no longer matches → child/lineage recovery silently breaks. Either preserve the byte-shape that reader expects, or update the reader to parse JSON instead of substring-matching.

Per-area: perf PASS, optional-orjson PASS, cross-session-isolation PASS, scan-exclusion PASS — but the four lifecycle edges FAIL and compact-serialization introduces a recovery regression. 54 focused tests pass, but they don't exercise these lifecycle transitions (which is why they're green while the behavior is broken — the tests need the restart/rotation/delete-race/clear scenarios).

This is a **kick-back**, not a ship-with-fixes: the sidecar design needs genuine lifecycle ownership across create → rotate → delete → clear (coordinated across `api/routes.py`, `api/models.py`, `api/streaming.py`, and the clear protocol), and the compact-serialization must not break the substring recovery reader. Draft loss is user-facing data loss, so none of this can ship partially. The perf approach is right and worth finishing — but each of the four edges needs a real reproduction test that fails first, then the fix. I self-rebased this clean onto current master (resolved an import union: kept `fcntl`/`msvcrt` + your optional `orjson`); the findings are on that rebased head.

nesquena-hermes (changes requested):
## 🔬 Re-gate (warm-up review) — thanks, 4 of 5 findings genuinely converged; one CR-cited path still open → kicking back

Thanks @ai-ag2026 — this is a strong re-push. I re-verified each finding from the last review against the current head (`0f9739d43abb`) and **four of the five are genuinely resolved** with real, discriminating tests:

1. **DRAFT-LOSS new-chat restart → ✅ RESOLVED** — first non-empty draft now persists a minimal owner record before the sidecar write (`routes.py` ~14620); `test_first_nonempty_draft_persists_restartable_session_record` recovers the draft after `Session.load`.
2. **DRAFT-LOSS compression rotation → ✅ RESOLVED** — `migrate_composer_draft_sidecar(old_sid, new_sid)` runs on the rotation path (`streaming.py` ~3681/8871), takes both per-sid locks sorted (deadlock-safe), and won't clobber a newer draft; covered by `test_compression_rotation_moves_draft_to_continuation_owner`.
4. **STALE-DRAFT clear → ✅ RESOLVED** — clear is now canonical (`{"text":"","files":[]}`), durably clears the legacy in-file field, unlinks under the draft lock, and guards a newer draft via a `_canonical_draft` compare.
5. **NEW REGRESSION (compact serialization breaks the recovery reader) → ✅ RESOLVED** — the whitespace-sensitive substring reader was replaced with a real JSON parse (`_read_top_level_json_value_from_head(...)`, `models.py` ~3555); no other substring reader of `parent_session_id` remains.

## ❌ Still open — finding #3 is only half fixed (the bulk-cleanup path)

The last review explicitly required *"revalidate-and-delete under the draft lock on **all paths incl bulk cleanup**"* and called out *"the bulk cleanup path leaks sidecars too."*

- **3a delete/write TOCTOU → ✅ fixed** (POST re-resolves under the draft lock; single-delete + `_prune_orphaned_webui_zero_message_sessions` delete the sidecar under lock).
- **3b bulk cleanup → ❌ STILL APPLIES.** `_handle_sessions_cleanup` (`api/routes.py:20450`, serving `POST /api/sessions/cleanup` and `/api/sessions/cleanup_zero_message`) was **not touched by any of the four fix commits** (`git diff 5615e9a9d..HEAD -- api/routes.py` has no hunk near 20450). It does `SESSIONS.pop` + `p.unlink()` with **no draft lock and no `delete_composer_draft_sidecar`**, so `_drafts/{sid}.json` is leaked.

  Worse, this directly undermines fix #1: a draft-anchor record created by the new restart-durability path is `title == "Untitled"` and `len(messages) == 0`, which is **exactly** the sweep's deletion predicate (`s.title == "Untitled" and len(s.messages) == 0`). So bulk cleanup **deletes the draft anchor and leaves an orphaned sidecar → the user's draft is lost.** Since draft loss is user-facing data loss, this can't ship partially.

### Precise fix-spec
`api/routes.py::_handle_sessions_cleanup` (~20450):
- Phase 1 (~20455-20471): when deleting a session file, acquire `get_composer_draft_lock(p.stem)` and call `delete_composer_draft_sidecar(p.stem)` in the same critical section as `SESSIONS.pop`/`p.unlink` (mirror the single-delete handler at ~14759-14777). Preferably **skip** any session whose sidecar has a non-empty draft (`read_composer_draft_sidecar(p.stem)` truthy) so the fix-#1 anchor isn't destroyed.
- Phase 2 index-ghost sweep (~20504-20517): for each dropped ghost `sid`, also `delete_composer_draft_sidecar(sid)` under the draft lock.
- (Defense-in-depth) a standalone orphan-draft sweep over `SESSION_DIR/_drafts/*.json` that unlinks any whose `{sid}.json` is gone and `sid not in SESSIONS`, each under the draft lock.

### Test gap
The new `test_pr6011_draft_lifecycle.py` is genuinely good coverage for findings 1, 2, 3a, 4, 5 — but its "bulk prune" case exercises `_prune_orphaned_webui_zero_message_sessions` (the *fixed* function), **not** `_handle_sessions_cleanup` (the still-broken endpoint the review cited), so it gives a false green over the hole. Please add a test that invokes `_handle_sessions_cleanup` / `POST /api/sessions/cleanup(_zero_message)` on a draft-anchored `Untitled`/0-message session and asserts the draft is preserved (or at least the sidecar is deleted, no orphan). It must fail on the current head first, then pass with the fix.

The perf approach is right and worth finishing — this is the last lifecycle edge. Nice work closing the compact-reader regression and the other three.

ai-ag2026:
Re-gate evidence for current head `091f960d79cf9da039dd0747e13738ebd9c1beab` (rebased onto `a4a2f49e`):

- Independent final review: **ACCEPT** on this exact rebased object; patch-id matches the prior reviewed candidate with no rebase drift.
- The prior P1/P2 themes are covered in the current implementation and behavioral regressions: post/delete locking and draft-owner revalidation; full-read conservative backup when `message_count` is stale/uncertain; file identity validation for external replacement; nonfinite serialization parity; draft-only owner preservation, empty ghost pruning, and corrupt-owner fail-closed retention.
- Local verification on this head: `214 passed, 1 expected skip` across the persistence/draft/cleanup/save selector; `py_compile` for models/routes/streaming; added-line Ruff and `git diff --check` clean.
- Fresh full CI and Greptile Review are now running on this head. Historical review comments should be evaluated against this current code and tests.

sjungwon03:
I found and reproduced a small clear-path durability hole in this branch: if
the authoritative draft sidecar's `unlink()` raises `OSError`, the endpoint
currently returns success even though that old sidecar will win on the next
load and restore text the user was told was cleared.

I opened upstream issue #6242 and a **draft stacked PR** against this exact
head branch: https://github.com/ai-ag2026/hermes-webui/pull/1

The patch is one commit (`807ffe86`) and is cherry-pickable. It makes sidecar
cleanup report success/failure, returns HTTP 500 from the client-facing clear
endpoint on cleanup failure, and adds route/helper regressions. Focused suite:
`57 passed`; full suite: `13,210 passed` with 5 unrelated local environment
failures documented in the PR.

ai-ag2026:
@nesquena-hermes re-gate requested. All five findings from your 17.07 review are closed at head `a9371c860aca`, including the previously-open 3b: `_handle_sessions_cleanup` now takes the per-sid draft lock in Phase 1, skips deletion when a non-empty draft exists, deletes the sidecar in the same critical section, and a Phase-3 sweep removes ownerless ghost sidecars under the lock — covered by `test_cleanup_preserves_nonempty_draft_owner_and_removes_empty_owner_and_ghost`, which drives the real endpoint. Additionally the clear-endpoint silent-failure flagged by Greptile and @sjungwon03 is fixed (cherry-picked their `fix(drafts): fail closed when clear sidecar cleanup fails`): `delete_composer_draft_sidecar` reports success/failure and clear returns HTTP 500 instead of ok:true over a surviving sidecar. CI 18/18 green.

nesquena-hermes:
**Certified head:** `sha:a9371c860aca` · **PR:** #6011 · _@ai-ag2026, perf(session): persist composer drafts to a sidecar; compact + no-op saves_
**Verdict:** The last maintainer blocker (3b bulk-cleanup anchor preservation) is genuinely closed, and the full suite is green — but the gate reproduced a new draft data-loss edge on the same cleanup path plus a silent false-success on delete failure. Fix, then re-gate; do not merge.

### What I ran (isolated worktree `/tmp/wt-gate-6011`, patch-id-clean rebase `4f7b9b1aed67` on current `origin/master`)
| Gate | Result |
|------|--------|
| Codex (reproduce) | **SHIP ONLY WITH FIXES** — 4 findings, all reproduced against the real modules (2 CORE data-loss/strand, 1 CORE transient-read delete, 1 SILENT false-success). |
| Opus (senior review) | **SAFE TO SHIP** — confirmed 3b + findings 1/2/3a/4/5 closed; but its P3 non-blocking note modeled only a *corrupt* (unrecoverable) sidecar, not the *transient* read failure below. |
| Full pytest suite | **13,508 passed, 0 failed** (10 skipped, 1 xfailed, 2 xpassed, 34 subtests); ran to completion serially (`-p no:xdist`). |
| Private/static gates | Threat scan CLEAN; rebase patch-id byte-identical to PR head; ruff/py_compile clean. |
| My own reproductions | Reproduced 3b closure AND the new transient-read data-loss edge directly against the real modules. |

### Findings
- ✅ **[prior blocker 3b — CLOSED]** `_handle_sessions_cleanup` (`api/routes.py:20650`) now acquires `get_composer_draft_lock(sid)`, resolves the draft under the lock, adds `and not has_draft` to both deletion predicates, deletes the sidecar in-section, and sweeps orphan `_drafts/*.json`. My probe: a draft-anchored `Untitled`/0-message session is preserved on both the default and `zero_only` branches; the no-draft control stays deletable. Findings 1/2/3a/4/5 from prior rounds hold.
- ⛔ **[CORE — DATA LOSS, reproduced] transient sidecar read failure deletes a recoverable draft.** `read_composer_draft_sidecar` (`api/models.py:1259`) returns `None` on BOTH an absent sidecar AND a transient `OSError`/parse failure (lines 1271-1272, 1281-1282). In `_handle_sessions_cleanup` the `has_draft` guard then falls back to empty → the Untitled/0-message anchor is deleted and the sidecar unlinked. **My probe** forced one `read_text()` `OSError` on a real non-empty sidecar (`text="important unsaved draft"`): `has_draft=False`, `would_delete_anchor=True`, yet `draft_still_on_disk` returned the intact draft — a fully recoverable draft destroyed by a momentary read hiccup during a cleanup sweep. This is the same fail-open class the maintainer bounced 3b on, now triggered by I/O error. (Opus rated this non-blocking, but only modeled a *corrupt/unrecoverable* sidecar; the transient-read case loses a recoverable draft.) **Fix:** make the read tri-state (absent / present-valid / unreadable-error); cleanup + orphan pruning must fail closed and preserve BOTH files whenever an existing sidecar cannot be read+validated.
- ⛔ **[SILENT — reproduced] cleanup/delete report success while the sidecar remains.** `delete_composer_draft_sidecar` (`api/models.py:1321`) correctly returns `False` on unlink failure, but the callers ignore it: Phase-1 cleanup still `cleaned += 1` (`routes.py:20683`), the orphan sweep (`:20779`), and single-session delete (`:14955`, returns `{"ok":true}` at `:15021`). Codex reproduced `{"ok":true,"cleaned":1}` with the sidecar still on disk → the "cleared" draft resurrects on next load. **Fix:** honor the return value everywhere; never report success / increment `cleaned` while an authoritative sidecar remains; aggregate partial failures.
- ⚠️ **[Codex CORE, reproduced — verify] clear-CAS + compression-handoff races.** (1) `_clearComposerDraft` (`static/sessions.js:325`) compares the server prefix against the full submitted text, so a sent message can reappear as a stale draft after refresh (`routes.py:14774`); fix by versioning draft writes (acknowledged-revision compare) and serializing in-flight saves before clear. (2) A concurrent draft POST to the old SID during compression handoff (`streaming.py:3900` / `models.py:1347`) can strand the newest draft on the archived session. Opus judged these "resurrection-not-loss / newer-wins" and pre-existing-class — genuinely on the fence; the contributor should confirm the clear-CAS versioning and the old-SID→continuation routing.

**RED — @ai-ag2026 should make the sidecar read tri-state + fail-closed in cleanup/pruning (the reproduced transient-read data-loss), honor `delete_composer_draft_sidecar`'s return at all four call sites, and version the clear-CAS / route old-SID draft ops through the continuation. Re-gate from zero once landed. Do not merge this head.** The 3b bulk-cleanup fix and the perf approach are solid — these are the remaining lifecycle/I/O-failure edges. Preserve @ai-ag2026 author credit.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge/tag/deploy — that is the release agent's call. This cert is valid only while the head stays at sha:a9371c860aca._

ai-ag2026:
Re-gate requested — finding #3b (bulk-cleanup path) is now closed on head `a9371c86`, exactly along the fix-spec:

* **Phase 1** (`_handle_sessions_cleanup`): every candidate is evaluated under `get_composer_draft_lock(sid)`; a session whose sidecar resolves to a non-empty draft is **skipped** (the fix-#1 draft anchor survives both `cleanup` and `cleanup_zero_message` — `should_delete` requires `not has_draft`), and for deleted sessions `SESSIONS.pop` + `unlink` + `delete_composer_draft_sidecar` run in the same critical section, mirroring the single-delete handler.
* **Phase 2 ghosts**: dropped index ghosts are covered by the new standalone orphan sweep (below) rather than inline deletion — same effect, one code path.
* **Defense-in-depth**: a sweep over `SESSION_DIR/_drafts/*.json` unlinks any sidecar whose owner file is gone *and* whose sid is not in `SESSIONS`, re-checking ownership under the draft lock so an in-flight draft-only new session can never be swept.
* **Test gap closed**: `test_cleanup_preserves_nonempty_draft_owner_and_removes_empty_owner_and_ghost` drives `_handle_sessions_cleanup` itself (both `zero_only` variants), asserting the draft-anchored `Untitled`/0-message owner survives with its draft readable after a simulated restart, the empty-draft owner is deleted *with* its sidecar, and an ownerless orphan sidecar is removed. Full `test_pr6011_draft_lifecycle.py`: **13 passed**.

Additional hardening on the same head: corrupt draft owners and failed sidecar cleanup now fail closed (`091f960d`, `a9371c86`).

nesquena-hermes (changes requested):
## Re-gate: draft lifecycle still fails closed in several destructive/racy paths

The performance work and the five earlier requested fixes remain present, but exact head `e99ed25bdc92` still has four objective data-integrity failures. These are patch-equivalent to the prior reviewed series, not master-rebase artifacts.

1. `read_composer_draft_sidecar()` conflates absent and unreadable. Cleanup treats a transient read/stat/parse failure as no draft, then deletes a valid zero-message owner and its recoverable draft.
2. Several destructive callers ignore `delete_composer_draft_sidecar()` returning false. Cleanup and `/api/session/delete` can report success after unlink failure while an orphan nonempty sidecar remains.
3. Clear is not generation-ordered against an already in-flight autosave. A pre-clear write arriving after the successful clear is accepted and resurrects the draft; the client also swallows clear failure.
4. Compression migration copies once but installs no durable old-SID → continuation routing. A delayed old-SID autosave is still accepted because the archived parent exists, stranding the newest draft on the old SID.

Please make destructive reads tri-state/fail-closed; honor deletion results at every cleanup/delete/migration caller; add a revision/generation protocol so stale pre-clear writes are rejected; and install canonical continuation routing (or return a conflict with the active SID) for late old-SID requests. Add failure-injection and reordered-request tests for each path.

Evidence through the mandatory sandbox wrapper: 5 reviewer probes passed while asserting the reproduced bad states. The official targeted shard is 42 passed plus one brittle mtime-only assertion failure; the inode, size, and bytes prove the write occurred, so repair that test to assert bytes/decoded title rather than timestamp inequality. Threat scan CLEAN. No full suite was run in this warm-up pass.

nesquena-hermes (changes requested):
## Re-gate at `7bad3233b151` — lifecycle rework still loses or resurrects drafts

Thanks for the substantial follow-up. The tri-state destructive reads, session-delete ordering, in-lock redirect check, and clear-failure UI handling are real improvements. The exact head still has four objective data-integrity failures, three reproduced through the mandatory sandbox wrapper:

1. **A first in-flight autosave can still resurrect a cleared draft.** `static/sessions.js::_composerDraftRevision()` returns `null` until the client has received a successful mutation, and `api/models.py::composer_draft_revision_is_stale()` explicitly accepts a missing revision. In the reproduced schedule, a first autosave was dispatched with `rev:null`, clear completed first, then the old autosave arrived and returned 200 with the deleted text restored. The new test obtains a revision from an earlier completed save, so it does not cover this first-write race.

2. **A delayed old-SID payload is rejected but not preserved under the continuation SID.** The server returns 409 with `session_id`, `draft`, and `rev`, but `static/sessions.js::_handleComposerDraftConflict()` only forgets old-SID bookkeeping. It does not replay the rejected `{text, files}` to the target, transfer ownership, or retain the target revision. The reproduced newer delayed payload was dropped while the continuation kept the older copied draft.

3. **Migration failure still is not observable.** `api/models.py::migrate_composer_draft_sidecar()` returns no status. If redirect publication fails, it calls `delete_composer_draft_sidecar(old_sid)` and ignores that boolean too; `_preserve_pre_compression_snapshot()` proceeds as though migration converged. Failure injection left a plain old-SID draft with no durable redirect and no surfaced failure.

4. **Migration publishes the destination sidecar before its owner and cleanup can reap it.** The copied `new_sid` draft is written in `migrate_composer_draft_sidecar()` before the new session JSON/cache owner is published by the compression path. `_handle_sessions_cleanup()` classifies that sidecar as ownerless and deletes it. The sandboxed interleaving reproduced cleanup returning `{ok:true}`, deleting the new-SID draft, and leaving only the old redirect marker.

Please:

- establish and quote a baseline revision before this client's first autosave, then add the exact first-autosave-dispatched → clear-wins → late-autosave ordering test;
- preserve/replay a rejected old-SID payload under the validated continuation SID, including the target revision and active ownership, and assert the newest text is durable;
- make migration return a structured success/failure, honor both redirect-write and fallback-delete failures, and propagate failure through the compression caller;
- reserve or publish the destination owner atomically with its sidecar so cleanup cannot reap it in the migration gap, with a production-composed migration/cleanup interleaving test;
- propagate orphan-sidecar deletion failures instead of returning unconditional global success.

Mandatory wrapper evidence: threat scan CLEAN; untouched targeted shards passed (29 lifecycle tests, 59 neighboring draft tests, 25 compression/cleanup tests). Reviewer probes passed while asserting the four reproduced bad states. The temporary probes were removed, the worktree was restored clean, and the lifecycle target passed again (29/29). No full suite was run in this warm-up pass.

ai-ag2026:
All four data-integrity failures are closed, plus three regressions an adversarial pass proved afterwards (two of them introduced by my own first attempt).

**1. Tri-state reads.** `read_composer_draft_sidecar_status()` reports absent / present / unreadable / redirected; `resolve_composer_draft_status()` carries a `readable` flag. Only *absent* means "no durable draft". A stat/read/parse failure is unknown, and every destructive caller fails closed on it — cleanup no longer deletes a valid zero-message owner and its recoverable draft because of a transient error.

**2. Deletion results are honored** at cleanup, the zero-message prune, the bulk sweep, the orphan sweep and `/api/session/delete`.

**3. Generation ordering.** Mutating responses carry a monotonic per-session `rev`; the client quotes the last one it saw; a write quoting an older revision is rejected with 409 instead of applied. A clear always wins and fences everything queued before it. The client also stopped swallowing clear failures (keeps the session marked dirty, drops the restore suppression, surfaces `draft_clear_failed` — added to all 15 locales).

**Scope, stated plainly:** the counter is an in-memory dict, so it is monotonic *within one process*. A restart resets it to 0 and `claimed < current` then accepts every pre-restart revision, so a request queued across a restart is still accepted. This closes the common in-process autosave-vs-clear race and **not** the restart case. That is now documented at the definition and a test pins the documentation. Making it durable is a follow-up.

**4. Continuation routing.** `migrate_composer_draft_sidecar()` leaves a durable `redirect_to` marker in the old sidecar slot (survives restart) and the draft route answers 409 with the live `session_id`.

**Found by adversarial review afterwards, fixed:**
- A draft write could **destroy** the marker: the redirect probe runs before the per-session lock, and a compression takes the *same* lock, so "compression wins the lock first" is ordinary scheduling. The in-lock re-read discarded its redirect result and fell through. Now honored → 409.
- Failing closed on `/api/session/delete` was **worse than the problem**: the session JSON was unlinked first and the early return then skipped the index prune, the tombstone, the attachment directory and the turn/run journal deletion (the #3802 privacy fix), with no way to make progress on a retry. The sidecar is now removed first and the request aborts before anything is destroyed.
- The orphan sweep **deleted continuation markers** — it globbed `_drafts/*.json` and removed anything whose owner was absent. Markers and unreadable sidecars are now skipped.

Also: the brittle `st_mtime_ns` assertion in `test_noop_resave_skips_disk_write` now asserts on-disk bytes and the decoded title, per your note.

246 passed in the draft/session-delete/cleanup slice; every fix is pinned by a test that fails when reverted. ruff diff-gate clean.

PR #6013, opened 2026-07-13
perf(providers): cache provider list responses

### LINKED ISSUE #6010: GET /api/providers is serially slow (5-9s+) on every profile switch — no caching, no per-provider timeout guard
## Summary

Profile switching in the WebUI fires `GET /api/providers` alongside `POST /api/profile/switch` and `GET /api/sessions`. While #4662 optimized the latter two (cold switch ~5s → ~1s), the `/api/providers` endpoint was not addressed and remains a consistent **5-9s** on every request — there is no TTL cache and no per-provider timeout guard for `get_auth_status()` / `provider_model_ids()` calls inside `get_providers()`.

The endpoint takes **5-9s every time**, not just on cache miss. Even back-to-back requests pay the full cost.

### 1. `webui.log` — historical `GET /api/providers` timings (ms field)

```
2026-07-06T08:18:29Z  GET /api/providers  200  27590.1 ms
2026-07-06T08:18:29Z  GET /api/providers  200  27631.1 ms
2026-07-07T07:21:39Z  GET /api/providers  200   9344.3 ms
2026-07-07T07:34:08Z  GET /api/providers  200  13687.9 ms
2026-07-07T12:00:24Z  GET /api/providers  200   5050.3 ms
2026-07-09T03:25:49Z  GET /api/providers  200   6826.1 ms
2026-07-13T03:20:34Z  GET /api/providers  200   9358.1 ms
2026-07-13T03:21:29Z  GET /api/providers  200   9224.4 ms
```

Range: **5,050 ms – 27,631 ms** across 8 samples.

### 2. Live `curl` measurements (3 consecutive, no caching observed)

```
run1: 200  8.98 s
run2: 200  4.97 s
run3: 200  4.99 s
```

No warm path — every request re-probes all providers.

### 3. Per-function profiling (monkey-patched timing in a subprocess)

Ran `get_providers()` with timing wrappers around the key functions. Total: **7,270 ms** for 50 providers.

**`provider_model_ids()` — 18 calls, sum 5,009 ms:**

| Provider | ms |
|---|---|
| `copilot-acp` | 3,577 |
| `AI-codex` | 1,003 |
| `kilocode` | 20 |
| `nous` | 1.8 |
| *(rest <2 ms each)* | |

**`get_auth_status()` — 31 calls, sum 4,697 ms:**

| Provider | ms |
|---|---|
| `bedrock` | 1,052 |
| `copilot` | 708 |
| `xai-oauth` | 0.9 |
| `nous` | 0.9 |
| *(rest <1 ms each)* | |

**`_provider_has_key()` — 50 calls, sum 1,114 ms** (local file reads; `copilot` alone 1,042 ms).

### 4. Root cause — serial loop with zero caching

`get_providers()` (`api/providers.py:2450`) iterates all known provider IDs in a `for pid in sorted(known_ids)` loop and calls these **synchronously, per provider**:

- `provider_model_ids(pid)` — for `copilot`/`copilot-acp` this calls `_fetch_github_models()` → `fetch_github_model_catalog()` which issues `urllib.request.urlopen(COPILOT_MODELS_URL, timeout=5.0)`. Even with no API key, it still appends `copilot_default_headers()` and fires the HTTP request, waiting up to 5s.
- `get_auth_status(pid)` — for `bedrock` this calls `has_aws_credentials()` → `botocore.session.get_credentials()` (boto3 credential chain, ~1s). For `copilot` it shells out to `gh` CLI (~0.7s).
- `_provider_has_key(pid)` — reads `.env` + `config.yaml` per provider (cheap individually, but 50× adds up).

The route handler (`api/routes.py:12104`) passes the result through with no caching:

```python
with profile_env_for_active_request_readonly("/api/providers", ...):
    return j(handler, get_providers())
```

There is no in-memory TTL cache on `get_providers()` itself. Contrast with `get_provider_quota()` which already has `_ACCOUNT_USAGE_CACHE_TTL_SECONDS = 45.0` — the same pattern is missing here.

## Relation to existing issues

| Issue | What it fixed | What it left out |
|---|---|---|
| **#4662** (closed) — Profile switching: faster switch | Optimized `/api/sessions` cold build (780ms → 200ms) + frontend skeletons. Cold switch ~5s → ~1s. | Did not measure or address `/api/providers` (5-9s), which fires concurrently during switch. |
| **#3957 / #3960** (closed) — Non-default profile Providers timeout | Fixed credential resolution scope (per-request profile env) + profile-keyed models cache. | The issue body explicitly described the serial probe problem ("does a network round-trip per provider, serially") but the fix targeted profile isolation, not latency. |
| **#3024 / #3025** (closed) — `/api/models` 30s timeout | Added 5s per-provider cap (`CUSTOM_MODELS_ENDPOINT_TIMEOUT_SECONDS`) for `_read_custom_endpoint_models()`. | Only covers custom-provider `/v1/models` probes in the `/api/models` path. Does **not** cover `get_providers()` → `provider_model_ids()` (copilot GitHub API) or `get_auth_status()` (bedrock boto3, copilot gh CLI). |

## Proposed fix (backend only, no frontend change)

Two independent improvements, ordered by ROI:

### A. Add a TTL cache to `get_providers()` (highest ROI, ~30 lines)

```python
_PROVIDERS_CACHE: dict[str, Any] = {}
_PROVIDERS_CACHE_TS: float = 0.0
_PROVIDERS_CACHE_TTL_SECONDS: float = 30.0  # match _ACCOUNT_USAGE_CACHE_TTL_SECONDS precedent

def get_providers() -> dict[str, Any]:
    now = time.time()
    cached = _PROVIDERS_CACHE.get("data")
    if cached is not None and (now - _PROVIDERS_CACHE_TS) < _PROVIDERS_CACHE_TTL_SECONDS:
        return cached
    # ... existing build logic ...
    _PROVIDERS_CACHE["data"] = result
    _PROVIDERS_CACHE_TS = now
    return result
```

**Expected impact:** back-to-back and switch-triggered requests hit the cache → **~1 ms** instead of 5-9s. The 30s TTL is short enough that provider config changes surface quickly, and the Settings page already has a "refresh" affordance if needed.

### B. Parallelize provider probes with `ThreadPoolExecutor` (addresses cold path)

The existing `ThreadPoolExecutor` import is already in `api/providers.py:174`. Wrap the per-provider body of the `for pid in sorted(known_ids)` loop into a function and submit all at once:

```python
with ThreadPoolExecutor(max_workers=8) as executor:
    futures = {executor.submit(_build_single_provider_entry, pid): pid for pid in sorted(known_ids)}
    for future in as_completed(futures):
        entry = future.result()
        if entry:
            providers.append(entry)
```

**Expected impact:** cold path drops from ~7-9s (sum of all serial probes) to ~3.5s (max single probe = copilot-acp GitHub API call). Combined with fix A, the cold path only runs once per 30s.

## Environment

- hermes-webui: `exp-v0.52.55` (git HEAD `f8546915`)
- hermes-agent: `~/.hermes/hermes-agent` (pip-managed)
- Python: 3.14
- OS: macOS 26.0 (Darwin)
- Running via `server.py` directly (tmux session, not Docker)
- 50 providers detected (including `copilot`, `copilot-acp`, `bedrock`, `AI-codex`, `nous`, `xai-oauth`, plus custom providers)

## Repro

1. Open the WebUI on any profile
2. Switch to a different profile (e.g., default → any named profile)
3. Observe the Settings/Providers panel takes 5-9s to populate
4. Alternatively: `curl -w '%{time_total}s' http://127.0.0.1:8787/api/providers` — repeat 3×, note no warm-path speedup

SeanWit:
## Summary

- cache `GET /api/providers` responses for a short TTL to avoid re-running expensive provider probes on every profile switch
- scope the cache by active Hermes home, `.env` mtime, `config.yaml` mtime, and config fingerprint to avoid cross-profile leakage
- invalidate the providers cache when provider credentials or config-backed keys are changed
- add regression coverage for cache reuse, profile-home scoping, and key-save invalidation

Closes #6010

## Tests

- `~/.pyenv/versions/3.11.9/bin/python -m py_compile api/providers.py tests/test_provider_management.py`
- `~/.pyenv/versions/3.11.9/bin/python -m pytest tests/test_provider_management.py -q -o addopts=`
- `~/.pyenv/versions/3.11.9/bin/python -m pytest tests/test_issue3957_profile_providers_models.py -q -o addopts=`
- `git diff --check`


nesquena-hermes:
Pulled the branch at `b1867d7bf`, read the full diff against `origin/master`, and cross-checked the caching/invalidation seams and the profile-scoping guarantees. This is a clean implementation of "approach A" from #6010 and it holds up.

## What the PR does

Adds a 30s TTL cache around `get_providers()` in `api/providers.py`, keyed on the active Hermes home plus the mtimes of `.env`/`config.yaml` plus a config fingerprint. Cache hits `copy.deepcopy` the payload so callers can't mutate the stored dict.

## Correctness review

**Profile isolation is right.** The cache key uses `_get_hermes_home()` → `get_active_hermes_home()`, which respects the per-request TLS profile context from #798, so concurrent requests on different profiles can't cross-contaminate:

```python
def _providers_cache_key(cfg: Any) -> tuple[Any, ...]:
    home = _get_hermes_home()
    ...
    return (
        home_key,
        _providers_file_mtime_ns(home / ".env"),
        _providers_file_mtime_ns(home / "config.yaml"),
        _providers_config_fingerprint(cfg),
    )
```

The `test_provider_cache_is_scoped_by_profile_home` test exercises exactly this by flipping `get_active_hermes_home` between two homes and asserting `has_key` differs. Good — this is the #3957 regression class, so guarding it explicitly is the right call.

**Mutation invalidation covers the direct paths.** `set_provider_key` (providers.py:2913) and `_clean_provider_key_from_config` both call `invalidate_providers_cache()`. And even without those explicit calls, editing `.env` or `config.yaml` bumps the mtime component of the key, so the next read misses the cache naturally. That belt-and-suspenders design is nice.

**`_store_cached_providers` clears the whole dict before inserting**, so the cache holds at most one entry. That bounds memory (no leak across many profiles), at the cost of thrashing if two profiles are polled in tight alternation. Given profile switches are human-paced, that's an acceptable trade — worth a one-line comment noting the single-entry-by-design choice so a future reader doesn't "fix" it into an unbounded dict.

## One gap worth a look

The mtime + `set_provider_key` invalidation covers key-file mutations, but **OAuth login/logout** flows change `get_auth_status()` results without necessarily touching `.env`/`config.yaml` mtimes or routing through `set_provider_key`. If a user connects/disconnects an OAuth provider (e.g. copilot via `gh`, or bedrock creds appearing), the auth badge could stay stale for up to 30s. That's a minor UX lag, not a correctness bug (the TTL self-heals), but if there's a dedicated OAuth login/logout handler it'd be worth adding `invalidate_providers_cache()` there too, mirroring how it already sits next to `invalidate_models_cache()` / `invalidate_account_usage_status_cache()`.

## On approach B (parallelization)

The issue also proposed a `ThreadPoolExecutor` fan-out for the cold path. This PR deliberately does only A, which is the right first step — A alone turns the switch-triggered and back-to-back requests into ~1ms hits, which is the reported pain. B can land as a follow-up if the once-per-30s cold rebuild (~7-9s) is still noticeable; keeping them separate keeps this PR reviewable.

## Verdict

Looks correct and well-tested. Per the review policy I did not execute the test file — but reading the three added tests, the assertions match the behavior (cache reuse via call-count, profile scoping via `has_key` flip, invalidation after `set_provider_key`). Suggest (1) the single-entry-by-design comment and (2) considering OAuth-path invalidation before merge.

nesquena-hermes (changes requested):
## 🔬 Authoritative release gate — SHIP ONLY WITH FIXES (full-suite test interaction)

The cache implementation itself gated **SAFE** (Codex authoritative: locked single-entry cache, deepcopy-isolated values, profile-scoped key = home + `.env`/`config.yaml` mtimes + config fingerprint, invalidation wired on every provider-key / config-cleanup / OAuth-credential write path, no circular import, no bare except). The 30s OAuth-auth staleness window is an accepted bounded tradeoff. But **CI is red on the full sharded suite** and it's a real test-interaction the cache introduces:

- **`tests/test_issue1202_oauth_provider_status.py`** — 3 tests fail on shards 0 + 1 across all Python versions:
  - `test_auth_error_preserved_when_not_logged_in_and_no_config_key`
  - `test_auth_error_preserved_when_not_logged_in_but_config_key_present`
  - `test_hermes_cli_import_error_does_not_discard_config_yaml_key`

- **Root cause:** these legacy tests mutate **mocked OAuth / auth state** (monkeypatching the auth-status path) and then call `get_providers()` expecting the new state — but they do **not** invalidate the new provider cache, and they don't change any of the cache-key inputs (profile home / `.env` mtime / `config.yaml` mtime / config fingerprint), so within the 30s TTL `get_providers()` serves the **cached** provider list from a *previous* test's auth state → the asserted `auth_error` / `has_key` value is stale → assertion fails. They pass in isolation (cold cache) and fail in the full suite (a sibling test warmed the cache first). Confirmed: the same file passes 15/15 on a #6028-only stage (no cache) — the interaction is unambiguously this PR's cache.

- **Fix (pick one, prefer A):**
  - **A (test-hook, cleanest):** export a test-visible `invalidate_providers_cache()` (you already have it) and call it in an **autouse fixture** in `test_issue1202_oauth_provider_status.py` (and any other test module that monkeypatches auth state), so each test starts cold. Better still, add an autouse fixture in `conftest.py` that clears the provider cache between tests globally — mocked-auth mutation is a cross-cutting test pattern.
  - **B (key the cache on auth-state):** fold a cheap auth-state fingerprint into the cache key so a mocked/real auth change busts it — but this partly defeats the 30s OAuth-staleness tradeoff you accepted, so A is preferred.

The production code is sound; this is purely test-hermeticity (the same class that CI catches but isolated runs miss). Add the autouse invalidation, confirm `test_issue1202_oauth_provider_status.py` passes **in the full suite** (not just isolated), and I'll re-gate + ship. Thanks @SeanWit — the cache design is genuinely good, this is the last mile.

nesquena-hermes:
## Re-review — both items from my last pass are addressed

Re-pulled at `ab4d64f15` and read the diff against `origin/master`. The two things I flagged on the `b1867d7bf` pass — (1) a "single-entry by design" comment and (2) OAuth-path invalidation — both landed. This is now merge-ready from my read.

## OAuth invalidation gap is closed

My concern was that OAuth login/logout changes `get_auth_status()` results without touching `.env`/`config.yaml` mtimes or routing through `set_provider_key`, so the auth badge could stay stale for up to the 30s TTL. The PR adds a shared helper in `api/oauth.py:221` that invalidates *both* caches together:

```python
def _invalidate_provider_state_caches(provider: str) -> None:
    """Invalidate caches derived from provider credential state."""
    try:
        from api.config import invalidate_credential_pool_cache
        invalidate_credential_pool_cache(provider)
    except Exception:
        logger.debug("Failed to invalidate %s credential cache", provider, exc_info=True)
    try:
        from api.providers import invalidate_providers_cache
        invalidate_providers_cache()
    except Exception:
        logger.debug("Failed to invalidate providers cache after %s credential change", provider, exc_info=True)
```

It's wired into all three credential-mutating paths I could find in `oauth.py`: `_persist_codex_credentials` (:297), `_link_anthropic_credentials` (:404), and the unlink path `_remove_anthropic_link_marker` (:525). Each of those is exactly a `_write_auth_json` site, and grepping the file every `_write_auth_json` caller now has a matching `_invalidate_provider_state_caches` — so there's no remaining OAuth write that leaks a stale providers entry. The `test_oauth_credential_updates_invalidate_providers_cache` test (test_provider_management.py:173) asserts both invalidations fire from `_persist_codex_credentials`:

```python
assert invalidated_credentials == ["openai-codex"]
assert providers_invalidated == [True]
```

## Single-entry comment landed

`_store_cached_providers` (providers.py:1026) now documents the clear-before-insert as intentional:

```python
# Single-entry by design: /api/providers is cacheable only for the
# active profile/config snapshot, so clear older snapshots to avoid
# retaining unbounded provider metadata across profile switches.
_providers_cache.clear()
```

That's the note I wanted — a future reader won't "fix" the single-entry bound into an unbounded dict keyed by profile.

## Correctness still holds

The profile-scoping guarantee I checked last time is intact: `_providers_cache_key` (providers.py:992) keys on `_get_hermes_home()` resolved path plus `.env`/`config.yaml` mtimes plus the config fingerprint, and `_get_cached_providers` (:1013) `copy.deepcopy`s on both store and read so callers can't mutate the cached payload. `test_provider_cache_is_scoped_by_profile_home` (:105) flips `get_active_hermes_home` between two homes and asserts `has_key` differs — the #3957 regression class stays guarded.

## Verdict

Both requested changes are in and correct. This is approach A from #6010 done cleanly; approach B (`ThreadPoolExecutor` cold-path fan-out) remains a reasonable follow-up if the once-per-30s cold rebuild is still felt, but keeping it out of this PR was the right call. From source + contract inspection this is a PASS — I read the diff and the four new tests but did not execute the suite, so please confirm CI green before merge.

SeanWit:
I pushed 12fd727b to address the full-suite test isolation issue by clearing the providers cache around each test in conftest.py, then synced the branch with master at 186fef28. I verified the issue1202 tests and cross-file cache-contamination scenario locally, and the latest Greptile Review is passing on the current head. Could you please re-review when you have a chance?

nesquena-hermes:
Shipped in **exp-v0.52.83** 🚀 Thanks @SeanWit.

The cache design was solid from the start — the only thing holding it up was test-hermeticity (the cache surviving between tests, which only showed in the full suite). Your autouse `_invalidate_providers_cache` fixture fixed that and the full suite is green. Re-gate on the current head: Codex SAFE TO SHIP — profile-scoped key (home + `.env`/`config.yaml` mtime + fingerprint) can't serve one profile's providers to another, deepcopy-isolated values, immediate invalidation on key/OAuth changes, 30s TTL bounds external drift, lock-safe. Closes #6010.

PR #6014, opened 2026-07-13
perf(webui): short-circuit impossible state-db prefix reconciliation

Isla-Liu:
# perf(webui): short-circuit impossible state-db prefix reconciliation

## Stacked PR context

This is Layer 1 of 3: the foundation for #6015 and #6016. All three PRs target `master`, so the later PR diffs temporarily include their predecessor commits. Review and merge in order: #6014 → #6015 → #6016. After a lower layer lands, update the later branch or branches so each remaining PR contains only its own layer before merge.

Before merging any layer, refresh `master` and recheck integration with #5776. After updating a later branch, rerun the affected transcript-parity tests, focused tests, and performance checks.

## Thinking Path

- Session list/detail reconciliation can compare a large state-db prefix with the authoritative WebUI sidecar.
- Materializing the complete state-db identity-key set is wasted work when cheap prefix counts already prove that the two prefixes cannot match.
- The optimization must remain conservative: an inconclusive summary or a count that could still match continues through the existing authoritative reconciliation path.

## What Changed

- Read the readonly state-db prefix summary before materializing message identity keys.
- Skip impossible prefix normalization only when the summary proves a mismatch.
- Keep exact-match, missing-db, malformed, and read-failure cases on the established fallback path.
- Add focused reconciliation and readonly-state-db regressions.

## Why It Matters

Large sessions no longer build a complete state-db identity-key collection when the cheaper count proves that work cannot affect the result. This removes avoidable session-switch/list latency without weakening reconciliation correctness.

## Contract Routing

- Touched state layers: readonly `state.db` message-prefix summaries and WebUI JSON-sidecar reconciliation.
- Public contract: the authoritative sidecar/state-db reconciliation result is unchanged.
- Core invariant: optimization may skip work only after a cheap proof; uncertainty keeps the authoritative path.
- Contract change: none. No RFC or public contract wording changes are required.
- Docs: no contributor/setup/UI workflow changed; release-note wording is provided below instead of editing `CHANGELOG.md`.

## Verification

- `61 passed` in the focused readonly/reconciliation suite.
- Independent final-stack verification: `67 passed` in the equivalent focused set, with three transcript-parity cases passing.
- `git diff --check`: pass.
- Python compile: pass.
- Diff-scoped Ruff: zero findings on added/modified lines.
- Stack ancestry: `master@2b3016bf` is the exact merge-base and parent boundary.
- `CHANGELOG.md`: unchanged.

## Risks / Follow-ups

- A false positive would skip needed reconciliation, so all unknown/error cases deliberately fail open to the previous path.
- Integration with #5776: that PR optimizes compression lineage, while this stack fails closed when lineage or compression state is ambiguous. It overlaps this stack in exactly three paths: `api/models.py`, `api/routes.py`, and `tests/test_webui_state_db_reconciliation.py`. Whichever series is refreshed or lands second must reconcile those paths and rerun the transcript-parity and merge-tree checks before merge.
- This foundation must be reviewed and landed before the storage and route-consumer layers. Re-fetch `master` and #5776 and rerun merge-tree and semantic checks after any branch refresh and before merge.
- No UI/UX or visual change; screenshots are not applicable.

## Release Note

Improved large-session switching by avoiding state-db prefix work when a cheap count already proves reconciliation cannot match.

## Model Used

AI-assisted implementation and verification using OpenAI `gpt-5.6-sol` through Hermes Agent. Notable tools: Git worktrees/cherry-pick, pytest via `./scripts/test.sh`, diff-scoped Ruff, and Git merge-tree auditing.


nesquena-hermes:
Pulled `cron-pr-6014` at `b20652245` and read the full diff against `origin/master` — the new `get_state_db_session_message_prefix_summary` in `api/models.py:7354`, its wiring in `_state_db_since_timestamp_for_limited_display` (`api/routes.py:8494`), and the two tests in `tests/test_state_db_readonly_reads_models.py`. Cross-checked against the existing tail reader `get_state_db_session_messages` (`api/models.py:7199`) and the pre-existing identity guard `get_state_db_session_message_keys_before_timestamp` (`api/models.py:7423`). This is a clean, fail-open foundation layer — the short-circuit is correct and every ambiguous case (`None` summary, null timestamps, count mismatch, missing/malformed db) returns to the authoritative path. One consistency observation worth a follow-up.

## The short-circuit is correct — including the new fast branch

The new count pre-check compares an **active-filtered** state.db prefix count against the sidecar prefix count, then adds a fast exit that master didn't have:

```python
if null_timestamp_count or state_before_count != sidecar_before_count:
    return None, sidecar_messages
if sidecar_before_count == 0:
    return floor, sidecar_messages
```

I verified this `sidecar_before_count == 0` branch against the consumer. When the caller gets a non-`None` floor it reads the tail via `get_state_db_session_messages(..., since_timestamp=floor)` with `include_inactive` defaulting to `False`, so the tail reader applies the same `AND (active IS NULL OR active != 0)` clause (`api/models.py:7269-7271`) that the new summary query uses. Count `== 0` proves zero *active* state rows before the floor, the sidecar has zero rows before the floor, and the tail read excludes inactive rows — so the fast path and a full read agree that nothing displayable exists before the floor. Safe, and a genuine win over master, which always ran the key-materialization query here.

## Consistency gap: the identity guard is not active-filtered

The count query (new) and the tail reader (`api/models.py:7269`) both active-filter. But the identity guard the `sidecar_before_count > 0` branch falls through to does **not** — its `WHERE` is only `timestamp IS NOT NULL AND timestamp < ?` with no `active` clause (`api/models.py:7469-7470`). The agent's `messages` table really does carry `active` (schema at `tests/test_hermes_state.py:901`+ does `ALTER TABLE messages ADD COLUMN active`, and `append_message` stamps `active=1`), so this asymmetry is reachable.

Net effect for a large session that has inactive/compacted rows before the tail window **and** `sidecar_before_count > 0`: the count check passes (active counts match), but the key query then returns the extra inactive rows, `state_before_keys != sidecar_before_keys`, and the read falls back to the full path — even though the tail reader would have excluded exactly those inactive rows. That's a **lost optimization, not a correctness bug** (fallback is always safe), but it's precisely the large-session shape this stack targets, so the optimization can silently disengage where it's actually needed.

## Recommendation

Give the identity guard the same active filter the other two queries use, so all three share one definition of "before-floor visible prefix":

```python
# get_state_db_session_message_keys_before_timestamp, ~api/models.py:7469
active_clause = ""
if 'active' in available:
    active_clause = " AND (active IS NULL OR active != 0)"
# ... WHERE session_id = ? AND timestamp IS NOT NULL AND timestamp < ? {active_clause}
```

This is correctness-neutral for the guard's purpose — its job is to prove the sidecar prefix matches what *would be displayed*, and the display path is active-filtered — while it unlocks the fast path for sessions with inactive rows and removes the latent mismatch where the pre-check and the identity check disagree about what "before-floor" means.

## Verification suggestion

Worth one regression that exercises the `sidecar_before_count > 0` branch with inactive-before-floor rows present, asserting the optimized path still engages (returns a floor) once the guard is active-filtered — `test_get_state_db_message_prefix_summary_preserves_active_filtering` already covers the summary side, but nothing yet pins the guard/summary parity end-to-end through `_state_db_since_timestamp_for_limited_display`. Non-blocking; the layer is safe as-is.

nesquena-hermes:
Shipped in **exp-v0.52.70** 🚀 — thanks @Isla-Liu! Layer 1 of your perf stack. Codex authoritative gate SAFE (verified no false short-circuit can drop or stale a session — it only fires on proven-impossible prefix matches, falls through to the full reconciliation on any ambiguity, and a real exclusive-lock repro confirmed a locked db returns None not empty), stream-gate GREEN all modes, 61 tests, full CI green. 

For **#6015 / #6016**: now that Layer 1 is on master, please refresh those branches so each contains only its own layer (they currently include the #6014 commits), and I'll review + gate them in order. Nice, careful work on a delicate reconciliation path.

PR #6024, opened 2026-07-13
[locale]Localization for Simplified Chinese

Loukky:
Translate all untranslated English strings into Simplified Chinese.


nesquena-hermes:
Shipped in **exp-v0.52.63** 🚀 — Simplified Chinese (zh-CN) localization fill. Verified: pure value replacement on existing keys (no key add/remove → parity holds), no value-swap (en block stays English), placeholders preserved. Thanks @Loukky! zh-CN users no longer see English fall-through strings.

PR #6026, opened 2026-07-13
fix: keep mobile model picker in visual viewport

sNyteXx:
## Summary

- Keep the mobile model picker inside the visible visual viewport.
- Reposition the open picker when mobile browser chrome, the keyboard, or the visual viewport changes.
- Preserve the existing desktop dropdown behavior.
- Add regression coverage for viewport-safe positioning, CSS cascade order, and full menu-height measurement.

## Root cause

On phones, the model dropdown was absolutely positioned relative to the composer. Changes to the visual viewport could move it outside the visible area or let it overlap the mobile composer controls. The mobile override was also declared before the later desktop base rule, so the cascade restored `position: absolute`. Finally, measuring only the current `offsetHeight` before applying the available height could make the menu grow past the position that had just been calculated.

## Validation

- 100 relevant model-picker and mobile-layout tests passed.
- `node --check static/ui.js` passed.
- Verified model switching with Chromium touch emulation at 360×732.
- Verified model switching with a desktop viewport at 1440×900.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -3112,6 +3112,8 @@
     .composer-mobile-config-panel.open{display:flex;}
     .composer-mobile-config-action{box-sizing:border-box;flex:1 1 0;min-height:44px;min-width:0;background:var(--input-bg);border-color:var(--border);}
     .composer-mobile-context-action{flex:1 0 100%;}
+    /* Keep the rich picker in the visual viewport when mobile browser chrome moves. */
+    #composerModelDropdown{position:fixed;bottom:auto;box-sizing:border-box;min-width:0;width:calc(100vw - 16px);max-width:calc(100vw - 16px);z-index:220;}
     /* icon-only composer chips continue below mobile widths */
     .composer-divider{display:none;}
     .composer-status{flex:0 1 auto;min-width:0;max-width:96px;font-size:10px;text-align:right;}
@@ -3241,6 +3243,10 @@
 .ws-dropdown.open{display:block;}
 .ws-dropdown-footer{left:0;right:auto;bottom:calc(100% + 4px);min-width:280px;max-width:min(420px,calc(100vw - 32px));}
 .model-dropdown{display:none;position:absolute;bottom:calc(100% + 4px);left:0;min-width:280px;max-width:min(420px,calc(100vw - 32px));background:var(--surface);border:1px solid var(--border2);border-radius:10px;box-shadow:0 -4px 24px rgba(0,0,0,.4);z-index:200;overflow:hidden;max-height:min(70vh,640px);overflow-y:auto;}
+/* Re-assert the mobile picker after the base dropdown rule below the phone media block. */
+@media (max-width:640px){
+ #composerModelDropdown{position:fixed;bottom:auto;box-sizing:border-box;min-width:0;width:calc(100vw - 16px);max-width:calc(100vw - 16px);z-index:220;}
+}
 .model-dropdown.open{display:block;}
 .model-scope-note{position:sticky;top:0;z-index:2;padding:9px 14px;border-bottom:1px solid var(--border);color:var(--text);font-size:11px;font-weight:650;line-height:1.4;background:color-mix(in srgb,var(--surface) 82%,var(--accent-bg));box-shadow:0 1px 0 rgba(0,0,0,.12);}
 .model-group{padding:8px 14px 4px;font-size:10px;font-weight:700;letter-spacing:.04em;color:var(--muted);text-transform:uppercase;border-top:1px solid var(--border2);margin-top:2px;}.model-group.collapsible{cursor:pointer;user-select:none}.model-group.collapsible:before{content:'\25b6';display:inline-block;font-size:8px;margin-right:5px;transition:transform .15s ease}.model-group.collapsible.open:before{transform:rotate(90deg)}.model-group-body:empty{display:none}
--- a/static/ui.js
+++ b/static/ui.js
@@ -3662,12 +3662,52 @@ function _positionModelDropdown(){
   const panel=$('composerMobileConfigPanel');
   const anchor=(panel&&panel.classList.contains('open')&&mobileAction)?mobileAction:(chip&&chip.offsetParent?chip:mobileAction);
   if(!anchor) return;
-  const chipRect=anchor.getBoundingClientRect();
+  const anchorRect=anchor.getBoundingClientRect();
+  const isPhone=typeof window.matchMedia==='function'&&window.matchMedia('(max-width:640px)').matches;
+  if(isPhone){
+    const visualViewport=window.visualViewport;
+    const viewportWidth=Math.max(1,Number(visualViewport&&visualViewport.width)||window.innerWidth||1);
+    const viewportHeight=Math.max(1,Number(visualViewport&&visualViewport.height)||window.innerHeight||1);
+    const viewportTop=Math.max(0,Number(visualViewport&&visualViewport.offsetTop)||0);
+    const viewportBottom=viewportTop+viewportHeight;
+    const margin=8;
+    const gap=6;
+    const viewportLeft=Math.max(0,Number(visualViewport&&visualViewport.offsetLeft)||0);
+    const viewportRight=viewportLeft+viewportWidth;
+    const titlebar=document.querySelector('.app-titlebar');
+    const titlebarBottom=titlebar&&typeof titlebar.getBoundingClientRect==='function'
+      ? Number(titlebar.getBoundingClientRect().bottom)||0
+      : 0;
+    const contentTop=Math.max(viewportTop+margin,titlebarBottom+margin);
+    const menuWidth=Math.max(1,viewportWidth-margin*2);
+    const left=Math.max(viewportLeft+margin,Math.min(anchorRect.left,viewportRight-menuWidth-margin));
+    dd.style.left=`${left}px`;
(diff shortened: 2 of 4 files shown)


nesquena-hermes:
Pulled the branch at `168a36677` and read the full diff against `origin/master` — the three root causes in the description each map to a concrete, correct change. Notes from reading `static/ui.js`, `static/style.css`, and the added test.

## The cascade fix is real and the test guards it

On `origin/master` the mobile `.model-dropdown{position:fixed;…}` override lives in the composer phone block (style.css:~3112), which sits *above* the unconditional base rule at style.css:3243:

```css
.model-dropdown{display:none;position:absolute;bottom:calc(100% + 4px);left:0;…}
```

Same specificity, later source wins — so on a phone the base `position:absolute` at :3243 clobbered the earlier `position:fixed`, which is exactly why the picker rode the composer's stacking/overflow geometry instead of the viewport. The PR re-asserts the override *after* the base rule in a dedicated `@media (max-width:640px)` block (style.css:3247), and `test_mobile_model_dropdown_uses_visual_viewport_safe_positioning` pins the ordering directly:

```python
base_rule = CSS.find(".model-dropdown{display:none;position:absolute;")
mobile_override = CSS.rfind(".model-dropdown{position:fixed;")
assert base_rule >= 0 and mobile_override > base_rule
```

That `rfind` is the right assertion for a source-order cascade bug — a future refactor that hoists the override back above the base rule will fail the test rather than silently regress.

## Visual-viewport positioning is sound

The phone branch in `_positionModelDropdown()` (ui.js:3596-3625) keys off `window.visualViewport` and clamps both axes against `offsetTop`/`offsetLeft`, so it tracks the *visual* viewport (keyboard/URL-bar insets) rather than the layout viewport. The full-height measurement fix is here:

```js
const menuHeight=Math.max(dd.scrollHeight,dd.offsetHeight);
```

Measuring `scrollHeight` before applying `maxHeight` means the side-choice (`openAbove`) is made against the menu's intrinsic content height, not a stale clamped `offsetHeight` — that's the third root cause, and it's handled correctly. The desktop branch resets `top/bottom/width/maxWidth/maxHeight` to `''` (ui.js:3633-3637), so switching a session across the 640px boundary won't strand inline phone styles on the desktop dropdown. Good.

## One perf suggestion (non-blocking)

The new `visualViewport` `resize`+`scroll` listeners (ui.js:4564-4567) call `_positionModelDropdown()` directly on every event. During keyboard open/close the browser fires a burst of `resize`/`scroll` events, and each call does a write-then-read inside the phone branch (writes `dd.style.left/width/bottom`, then reads `dd.scrollHeight`/`offsetHeight` for `menuHeight`) — a forced synchronous reflow per event. Consider coalescing through `requestAnimationFrame` so the burst collapses to at most one layout pass per frame:

```js
let _mpReposScheduled=false;
const _repositionOpenModelDropdown=()=>{
  const dd=$('composerModelDropdown');
  if(!(dd&&dd.classList.contains('open'))||_mpReposScheduled) return;
  _mpReposScheduled=true;
  requestAnimationFrame(()=>{_mpReposScheduled=false;_positionModelDropdown();});
};
```

Purely a smoothness nit on the keyboard-animation path; the current code is correct, just chattier than it needs to be. The listener is only wired when `window.visualViewport` exists, so the desktop/no-VV path is untouched either way.

## Verdict

Correct fix for all three described causes, with a regression test that guards the subtle one (cascade order). I read the diff but did not execute the suite — the added assertions match the behavior, and `node --check` on ui.js is a clean syntax gate. Worth a maintainer pass on the rAF suggestion before or after merge; not a blocker.

nesquena-hermes (changes requested):
## 🔬 Gate review (Codex + browser DOM probe) — SHIP ONLY WITH FIXES (right approach; the CSS selector is too broad and pushes the Settings picker off-screen)

Thanks @sNyteXx — the `visualViewport` approach is exactly right and most of it verifies clean: the fallback path (forced `visualViewport=undefined`) doesn't throw and keeps the picker within bounds, desktop (1440×900) doesn't regress (both pickers keep their existing absolute geometry/width), and selecting a model still works + closes the picker. But a real regression + a clip:

1. **[CORE — selector too broad, breaks the Settings picker] `static/style.css:3116` + `:3248`** — the mobile rule targets `.model-dropdown`, which matches **both** the composer picker AND the **Settings** model picker. A browser probe shows the mobile Settings picker ends up **entirely off-screen** — i.e. the exact off-screen bug this PR fixes, reproduced on the sibling picker it shouldn't touch. **Fix:** scope both mobile selectors to `#composerModelDropdown` (not the broad `.model-dropdown`) so the Settings picker keeps its absolute positioning.

2. **[CORE — top obscured by titlebar] `static/ui.js:3671`** — the composer picker doesn't cross the visual-viewport edges, but its **top is obscured by the app chrome** (`.app-titlebar`). **Fix:** clamp against the visible content top *below* `.app-titlebar`, while still computing the viewport bottom from the unmodified visual-viewport offset.

3. **[test gap] `tests/test_mobile_layout.py:1713`** — the test only checks source strings + CSS ordering; it never executes the clamp arithmetic, viewport events, fallback, or the effective Settings cascade, so it passed despite both regressions. Add behavioral execution of the real positioning function (offset / resize / fallback / desktop-reset) + an assertion that the mobile **Settings** picker stays visible and absolute.

Once the mobile rule is scoped to the composer picker only and the clamp respects the titlebar, this is a clean mobile win — and then I'll capture the full before/after set (composer picker open, keyboard-up clamp, Settings picker) on real mobile + desktop viewports in light + dark for maintainer visual approval. Rebased clean on current master. Good catch on a real mobile-viewport bug — it just needs to not catch the Settings picker in the same net.

nesquena-hermes:
Re-reviewed after the force-push to `87215c577`. Two very different things landed in this revision, and only one of them matches the PR title.

## The model-picker fix improved — both review notes addressed

The rAF-coalescing suggestion from my last pass is in, and it's guarded by a behavioral test. `_repositionOpenModelDropdown()` now gates on a scheduling flag and collapses the `visualViewport` `resize`/`scroll` burst to one layout pass per frame (`static/ui.js:4639-4653`):

```js
function _repositionOpenModelDropdown(){
  const dd=$('composerModelDropdown');
  if(!(dd&&dd.classList.contains('open'))||_modelDropdownRepositionScheduled) return;
  _modelDropdownRepositionScheduled=true;
  requestAnimationFrame(()=>{ _modelDropdownRepositionScheduled=false; ... });
}
```

`test_visual_viewport_reposition_events_are_coalesced` in `tests/test_mobile_model_picker_behavior.py` pins it (`{"queued":1,"calls":1}` after two synchronous triggers). You also narrowed the CSS override from the shared `.model-dropdown` to `#composerModelDropdown{position:fixed;…}`, and `test_mobile_fixed_rule_targets_composer_picker_only` asserts `.model-dropdown{position:fixed;` is absent so the Settings picker keeps its `absolute` geometry. That's a real correctness gain over the earlier revision.

## But the force-push bundled ~1000 lines that aren't in this PR's scope

The commit message ("…and preserve local fixes") is doing a lot of work. The diff at HEAD now adds four unrelated feature areas the PR body never mentions:

- `api/unraid_status.py` (new, 232 lines) — a Home Assistant-backed Unraid/Tower status panel, plus the sidebar UI in `static/panels.js:6016` (`loadUnraidSidebarPanel`) and its CSS.
- `api/workspace_git.py` `git_dashboard()` (+98 lines) and the `loadWorkspaceGitMiniPanel` UI (`static/panels.js:6043`).
- Two new GET endpoints wired in `api/routes.py:12998` (`/api/git/dashboard`) and `:13001` (`/api/unraid/status`).
- A `custom:custom` provider-alias repair in `api/config.py:1346` (`_canonicalise_runtime_provider_id`) threaded through `resolve_model_provider` and `_resolve_compatible_session_model_state`.

None of that is described in the PR summary ("Keep the mobile model picker inside the visible visual viewport…"), and `api/unraid_status.py` reads a profile-specific dotenv path with an inline comment naming a particular local install. That's fork-local config leaking into a public PR.

## Recommendation

Split this back down to what the title promises. The cleanest path: reset the branch to just the `static/ui.js` + `static/style.css` + mobile-picker test changes (the parts reviewed here), and open the Unraid panel / git dashboard / `custom:custom` alias as their own PRs so each gets its own review and issue link. The `custom:custom` normalization in particular is a legitimate fix that deserves its own PR referencing whatever provider-alias issue it came from — buried here it'll never get found again.

The mobile-picker changes themselves are merge-ready on read-through; it's only the unrelated payload holding this back. I read the diff at `87215c577` against `origin/master` but did not execute the suite.

nesquena-hermes:
Tracking issue filed: #6080 (community report from Discord — mobile model selector does nothing / dropdown never opens). This PR is the candidate fix; consider adding `Closes #6080` once the requested changes are addressed.

nesquena-hermes:
Heads-up @sNyteXx — the underlying issue here (#6080, mobile model picker opening off-screen) shipped in **exp-v0.52.82** via #6105, which takes the same visualViewport approach you used here but reparents the dropdown to `<body>` so `position:fixed` escapes the `.composer-footer` `container-type`/`backdrop-filter` containing-block trap (that trap is why a footer-contained fixed element still landed off-screen even with correct coordinates). Your diff and #6105 overlapped heavily on the core idea. Leaving this open for a maintainer call on whether there's anything additive here worth pulling; if not it'll be closed as resolved-by-#6105 with credit. Appreciate the fix — the diagnosis was right.

PR #6028, opened 2026-07-13
fix(worktrees): unlock before remove so WebUI-created worktrees can be reaped

### LINKED ISSUE #6023: bug(worktrees): /api/session/worktree/remove always fails on WebUI-created worktrees — locked at creation, remove never unlocks
### Bug

`cli._setup_worktree` (which `api/worktrees.py` delegates to via `create_worktree_for_workspace`) places a `git worktree lock` on every worktree it creates. `remove_worktree_for_session` then runs `git worktree remove [--force] <path>`, but git refuses to remove a locked worktree unless it is unlocked first or `--force` is given **twice** (`remove -f -f`). Result: every worktree the WebUI itself creates cannot be removed through the WebUI — the "Remove worktree" UI action and the API endpoint both fail unconditionally.

Reproduced on v0.52.0 (code path unchanged on current master — `remove_args` appends a single `--force`, no unlock):

```
POST /api/session/new {"workspace": "<git repo>", "worktree": true}
→ session created, worktree at <repo>/.worktrees/hermes-xxxx, locked

POST /api/session/worktree/remove {"session_id": "..."}
→ {"error": "git worktree remove failed: use 'remove -f -f' to override or unlock first"}
```

Same failure with `"force": true` (single `--force` is still refused for locked worktrees).

### Fix options

a) `git worktree unlock <path>` before remove in `remove_worktree_for_session`, after the existing guards (dirty / untracked / unpushed / locked-by-stream / locked-by-terminal) have passed — those guards, not the agent's bookkeeping lock, are the real safety layer.

b) Pass `--force` twice when `force=true`.

(a) seems right: it keeps `force` semantics meaning "override dirty state", not "override the agent's creation-time lock", and makes the default (non-force) remove path work for clean worktrees.

### Related feature ask (could ride the same PR): optional worktree cleanup on archive/delete

Today session archive/delete intentionally retains the worktree (`worktree_retained` payload). That's the right conservative default, but it means worktrees accumulate with no lifecycle hook. Proposal: accept an optional `remove_worktree: true` in `/api/session/archive` and `/api/session/delete` bodies that invokes the same guarded `remove_worktree_for_session` path (never forced — dirty/unpushed worktrees still refuse and report `worktree_retained` as today). The frontend confirm dialogs already distinguish worktree sessions (#2057 copy), so a checkbox slots in naturally.

Happy to send a PR for the unlock fix (+ tests) if maintainers agree with option (a). Related design umbrella: #2057.

ayxuerui:
## Summary

Implements the unlock-before-remove fix confirmed in #6023: `remove_worktree_for_session` now runs a fail-soft `git worktree unlock` right before the remove, mirroring the agent's own `_cleanup_worktree` ordering.

WebUI-created worktrees are locked at creation on the agent side (`cli._setup_worktree`, reused via `create_worktree_for_workspace` → `_setup_agent_worktree`), and the remove path never unlocked — so git refused every UI-driven removal with `use 'remove -f -f' to override or unlock first`, `force=true` included.

## Approach

Exactly the shape recommended in the issue thread:

- Unlock is inserted after all dirty/untracked/unpushed/stream/terminal guards have passed — those guards remain the real safety layer, not git's lock.
- Fail-soft: `_run_git` is `check=False`, so unlocking a never-locked worktree is a no-op; the `try/except (OSError, TimeoutExpired)` keeps environmental failures non-fatal too.
- Double-`--force` deliberately not used: `force` keeps meaning "override dirty working state", never "blow past a foreign lock" — the `locked_by_stream`/`locked_by_terminal` guards stay meaningful.

## Tests

Real-repo tests (same pattern as the existing fixture-based cases in `tests/test_worktree_remove.py`):

- `test_remove_locked_worktree_succeeds_without_force` — reproduces the agent's creation-time lock (`git worktree lock --reason 'hermes pid=…'`), then asserts a plain `force=False` remove succeeds, the directory is gone, and `git worktree list` no longer lists it.
- `test_remove_never_locked_worktree_unlock_is_fail_soft` — sanity-probes that unlock would exit non-zero, then asserts removal still succeeds.
- Existing mocked call-order assertions updated for the leading unlock call (unlock first, then remove / remove --force).

`./scripts/test.sh tests/test_worktree_remove.py tests/test_issue2057_worktree_status.py tests/test_issue2057_worktree_lifecycle.py` — 24 passed.

The `remove_worktree: true` archive-time cleanup rider from the issue is intentionally left out per the thread — additive, separate change.

Fixes #6023

### DIFF
--- a/api/worktrees.py
+++ b/api/worktrees.py
@@ -265,6 +265,18 @@ def remove_worktree_for_session(session, *, force: bool = False) -> dict:
     repo_root = getattr(session, "worktree_repo_root", None)
     if not repo_root:
         raise ValueError("Session missing worktree_repo_root")
+
+    # Unlock the creation-time bookkeeping lock before removing (fail-soft).
+    # The agent locks every worktree it creates (cli._setup_worktree), and git
+    # refuses to remove a locked worktree with a single --force, so without
+    # this every WebUI-created worktree is unremovable.  The dirty/untracked/
+    # unpushed/stream/terminal guards above are the real safety layer, not
+    # git's lock — mirrors the agent's own _cleanup_worktree ordering.
+    try:
+        _run_git(["worktree", "unlock", str(worktree_path)], str(repo_root), timeout=5)
+    except (OSError, subprocess.TimeoutExpired):
+        pass  # already unlocked / never locked — non-fatal
+
     try:
         remove_args = ["worktree", "remove"]
         if force:
--- a/tests/test_worktree_remove.py
+++ b/tests/test_worktree_remove.py
@@ -78,6 +78,80 @@ def test_remove_clean_worktree_succeeds(tmp_path):
     assert not wt_path.exists()
 
 
+def test_remove_locked_worktree_succeeds_without_force(tmp_path):
+    """Issue #6023: WebUI-created worktrees are locked at creation by the
+    agent (cli._setup_worktree).  Removal must unlock first — otherwise git
+    refuses with \"use 'remove -f -f' to override or unlock first\" and every
+    WebUI-created worktree is unremovable from the UI."""
+    import subprocess
+    from api.models import Session
+
+    main = _make_minimal_git_repo(tmp_path)
+    wt_path = tmp_path / "wt_locked"
+    subprocess.run(
+        ["git", "-C", str(main), "worktree", "add", str(wt_path), "-b", "hermes/testlocked"],
+        check=True, capture_output=True,
+    )
+    # Reproduce the agent's creation-time lock.
+    subprocess.run(
+        ["git", "-C", str(main), "worktree", "lock", "--reason", "hermes pid=12345", str(wt_path)],
+        check=True, capture_output=True,
+    )
+
+    s = Session(
+        session_id="testlocked",
+        title="Locked",
+        workspace=str(wt_path),
+        worktree_path=str(wt_path),
+        worktree_branch="hermes/testlocked",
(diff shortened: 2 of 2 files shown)


ayxuerui:
Notes on the two Greptile P1s, for triage:

**1. Foreign lock ownership** — the unconditional unlock is deliberate and matches the design endorsed in #6023: the dirty/untracked/unpushed/stream/terminal guards above are the safety layer, and the ordering mirrors the agent's own `_cleanup_worktree`, which also unlocks unconditionally. A reason-prefix check (`hermes pid=…`) wouldn't meaningfully tighten ownership either: WebUI-created worktrees are always hermes-locked, and a *second* live hermes process's lock carries the same reason string, so ownership can't be distinguished by reason alone. Happy to add a reason-prefix guard as belt-and-suspenders if you'd prefer.

**2. Status-snapshot → remove race** — real, but pre-existing scope: the guards were snapshot-then-act before this change too, and `force=true` removal had the same window. Serializing status/unlock/remove under a session lifecycle lock feels like its own change — can pick it up as a follow-up if wanted.

nesquena-hermes:
Shipped in **exp-v0.52.62** 🚀 — WebUI-created git worktrees are now reapable (fail-soft `git worktree unlock` before remove, mirroring the agent's own cleanup ordering; the dirty-changes guard still runs first so user data stays protected). Codex authoritative gate SAFE, 13 tests, full CI green. Thanks @ayxuerui! Closes #6023.

PR #6029, opened 2026-07-13
feat(sessions): honor the agent's worktree config default in /api/session/new

### LINKED ISSUE #3294: bug: Profile Tool Configuration Not Respected in Web UI Unless Tool Restrictions Are Explicitly Set

### LINKED ISSUE #6022: feat(sessions): honor the agent's `worktree:` config default in /api/session/new
### Problem

Hermes Agent supports worktree-per-session isolation two ways: the `hermes -w` flag and a config-level default — top-level `worktree: true` in `~/.hermes/config.yaml` (`cli.py`: `use_worktree = worktree or w or CLI_CONFIG.get("worktree", False)`). With the config key set, every CLI/TUI session gets an isolated worktree automatically.

The WebUI gained worktree-backed sessions in #1955 and reuses the agent's own helper (`api/worktrees.py` imports `cli._setup_worktree`), but `/api/session/new` only creates a worktree when the request body explicitly carries `worktree: true` (the workspace-dropdown action). The agent-side config default is never consulted, so a user who sets `worktree: true` gets isolation in every CLI session but not in any WebUI session — same agent, same repo, different behavior.

Use case: several concurrent sessions (WebUI tabs + CLI + gateway) against one repo, where forgetting the dropdown action means sessions collide on the same working tree.

### Proposal

In the `/api/session/new` handler, treat an *absent* `worktree` key as "use the configured default" instead of "false":

```python
worktree_body = body.get("worktree")
if worktree_body is None:
    worktree_requested = bool(get_config().get("worktree", False))
    worktree_explicit = False
else:
    worktree_requested = (
        worktree_body is True
        or str(worktree_body).strip().lower() in {"1", "true", "yes", "on"}
    )
    worktree_explicit = True
```

Explicit `worktree: false` from a client still wins over the config default, so existing callers keep their exact semantics.

### Two behaviors that need care

1. **Non-git workspaces.** Today an explicit worktree request on a non-git workspace correctly errors. With a config default, that same hard error would break every new chat in non-git workspaces for users who set `worktree: true`. When the worktree came from the config default (`worktree_explicit == False`), catch the "not a git repository" failure and fall back to a plain session (perhaps with a `worktree_skipped` field in the response so the frontend can surface it). Explicit requests keep the hard error.

2. **Boot-time auto-bind sessions.** `_maybeBindFreshDefaultWorkspaceSession` (boot.js) silently creates a session on app load. With a server-side default, every page load would mint a worktree + branch that nothing cleans up (worktree sessions are deliberately excluded from new-chat draft reuse in `_isRestorableNewChatDraftSession`). Simplest fix: the auto-bind path sends `worktree: false` explicitly; deliberate "New Chat" clicks send nothing and inherit the default.

Happy to send a PR with the above plus tests (config-default on/off × body absent/true/false, non-git fallback vs explicit error, auto-bind never worktrees) if the direction sounds right. Related design umbrella: #2057.

*(Context: I currently get this behavior via a small self-written extension that wraps `newSession` to inject `worktree: true` with a workspace allowlist — works, but the server honoring its own config key is the cleaner home for it.)*

ayxuerui:
## Summary

Implements the config-default worktree behavior confirmed in #6022: `/api/session/new` now consults the agent's config-level `worktree:` key when the request body has no explicit `worktree` value, so WebUI sessions and CLI sessions agree on isolation for the same repo.

## Three-value model

- **body `worktree` absent** → agent config default applies
- **explicit `true`/`false`** → honored verbatim; explicit always wins (including `worktree: false` with config default `true` → plain session, covered by a dedicated test)

## Design decisions (per the issue thread)

**Profile-aware config read.** Rather than `hermes_cli.config.load_config()` (which would read one fixed config), the helper `_worktree_default_from_config(profile)` resolves the session's own profile via `get_hermes_home_for_profile` → `get_config_for_profile_home` — the mtime-tracked, profile-aware path already used to fix #3294. A user with `worktree: true` in one profile but not another gets per-profile behavior; the common single-profile case hits the ambient cache. Fail-soft: any config-resolution error logs a warning and returns `False` (config must never break session creation).

**Non-git fallback (`worktree_explicit` gate).** `create_worktree_for_workspace` → `find_git_repo_root` raises `ValueError` on non-git workspaces:
- explicit `worktree: true` → keeps today's hard 400,
- config-default request → degrades to a plain session and surfaces `worktree_skipped` (the reason string) in the response payload, so `worktree: true` in config.yaml doesn't 400 every session in every non-git directory.

**Boot-time auto-bind opts out in this same PR.** `_maybeBindFreshDefaultWorkspaceSession` (boot.js) and the post-onboarding session now send `worktree: false` explicitly — a server-side default must never mint a worktree + branch from merely opening the page (which would compound #6023's unremovable-worktree pile-up; fix for that side is #6028). `newSession()` forwards explicit true/false and omits the key otherwise; deliberate "New Chat" paths send nothing and inherit the default. Worktree-backed sessions were already excluded from new-chat draft reuse (`_isRestorableNewChatDraftSession` rejects `worktree_path`), so no change needed there.

## Tests

`tests/test_issue6022_worktree_config_default.py` — route matrix:
- config default off/on × body absent → plain / worktree-backed
- explicit `false` beats config default `true` (worktree factory pytest-fails if called)
- explicit `true` with config default off still creates a worktree
- non-git + config default → 200 plain session + `worktree_skipped`
- non-git + explicit `true` → 400
- `_worktree_default_from_config`: ambient read, named-profile home resolution, fail-soft on config errors

`tests/test_issue6022_worktree_ui_static.py` — static assertions that the auto-bind and onboarding paths opt out explicitly, `newSession()` implements the forward-verbatim/omit-absent contract, and deliberate New Chat paths don't pin a value.

One existing assertion updated (`test_issue1955_worktree_ui_static.py`) for the new `reqBody.worktree=!!options.worktree` shape.

`./scripts/test.sh` across all worktree/session-new test files (`test_issue6022_*`, `test_issue1955_*`, `test_issue2057_*`, `test_worktree_remove.py`, `test_issue5420_profile_switch_session_new.py`) — 59 passed. `scripts/ruff_lint.py` — no new findings.

Fixes #6022

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -7596,7 +7596,11 @@ def _worktree_default_from_config(profile: str | None) -> bool:
             cfg_dict = get_config_for_profile_home(get_hermes_home_for_profile(profile))
         else:
             cfg_dict = get_config_for_profile_home(None)
-        return bool((cfg_dict or {}).get("worktree", False))
+        # Strict boolean: only a real YAML `true` opts in.  Any other shape
+        # ("true", 1, [], {}, null, ...) is malformed for this key and must
+        # fall to the safe no-worktree default rather than truthiness-coerce
+        # into minting worktrees.
+        return (cfg_dict or {}).get("worktree", False) is True
     except Exception:
         # Config resolution must never break session creation.
         logger.warning("failed to read worktree config default", exc_info=True)
@@ -13914,7 +13918,11 @@ def handle_post(handler, parsed) -> bool:
         # worktree (e.g. the boot-time auto-bind) send ``worktree: false``
         # explicitly.
         raw_worktree = body.get("worktree")
-        worktree_explicit = raw_worktree is not None
+        # Presence-based, not truthiness-based: a client that sends the key at
+        # all (even ``worktree: null``) has spoken explicitly and never falls
+        # through to the config default.  ``null`` parses as non-true below,
+        # i.e. an explicit opt-out — only a genuinely ABSENT key inherits.
+        worktree_explicit = "worktree" in body
         if worktree_explicit:
             worktree_requested = (
                 raw_worktree is True
--- a/api/routes.py
+++ b/api/routes.py
@@ -7574,6 +7574,35 @@ def _rescale_threshold_tokens_for_context_window(
     return max(1, int(threshold * new_window / old_window))
 
 
+def _worktree_default_from_config(profile: str | None) -> bool:
+    """Return the agent's config-level ``worktree:`` default for *profile*.
+
+    The agent CLI honors ``worktree: true`` in config.yaml for every session
+    it creates (``use_worktree = worktree or w or CLI_CONFIG.get("worktree",
+    False)``).  /api/session/new consults this only when the request body has
+    no explicit ``worktree`` key, so both entry points to the same repo agree
+    on isolation (#6022).  Explicit body values always win.
+
+    Profile-aware on purpose: the WebUI serves multiple profiles from one
+    process, and a user with ``worktree: true`` in one profile but not another
+    expects per-profile behavior.  ``get_config_for_profile_home`` handles the
+    ambient/common case via the mtime-tracked cache and reads a diverging
+    profile's config.yaml directly off disk (see #3294).
+    """
+    try:
+        if profile:
+            from api.profiles import get_hermes_home_for_profile
+
+            cfg_dict = get_config_for_profile_home(get_hermes_home_for_profile(profile))
+        else:
+            cfg_dict = get_config_for_profile_home(None)
+        return bool((cfg_dict or {}).get("worktree", False))
+    except Exception:
+        # Config resolution must never break session creation.
+        logger.warning("failed to read worktree config default", exc_info=True)
+        return False
+
+
 def _session_model_state_from_request(
     model: str | None,
     requested_provider: str | None,
@@ -13877,10 +13906,22 @@ def handle_post(handler, parsed) -> bool:
         except (TypeError, ValueError) as e:
             return bad(handler, str(e))
         worktree_info = None
-        worktree_requested = (
-            body.get("worktree") is True
-            or str(body.get("worktree")).strip().lower() in {"1", "true", "yes", "on"}
-        )
+        worktree_skipped = None
+        # Three-value worktree model (#6022): an explicit body value always
+        # wins; an ABSENT key falls back to the agent's config-level
+        # ``worktree:`` default so WebUI sessions and CLI sessions agree on
+        # isolation for the same repo.  Clients that must never create a
+        # worktree (e.g. the boot-time auto-bind) send ``worktree: false``
+        # explicitly.
+        raw_worktree = body.get("worktree")
+        worktree_explicit = raw_worktree is not None
+        if worktree_explicit:
+            worktree_requested = (
+                raw_worktree is True
+                or str(raw_worktree).strip().lower() in {"1", "true", "yes", "on"}
+            )
+        else:
+            worktree_requested = _worktree_default_from_config(body.get("profile") or None)
         if worktree_requested:
             try:
                 from api.worktrees import create_worktree_for_workspace
@@ -13890,7 +13931,15 @@ def handle_post(handler, parsed) -> bool:
                 worktree_info = create_worktree_for_workspace(base_workspace)
                 workspace = worktree_info["path"]
             except (TypeError, ValueError) as e:
-                return bad(handler, str(e), status=400)
+                # Explicit requests keep the hard failure.  A config-default
+                # request on a non-git workspace (create_worktree_for_workspace
+                # raises ValueError) degrades to a plain session instead —
+                # otherwise `worktree: true` in config.yaml would 400 every
+                # session in every non-git directory.
+                if worktree_explicit:
+                    return bad(handler, str(e), status=400)
+                worktree_info = None
+                worktree_skipped = str(e)
             except Exception as e:
                 logger.exception("failed to create worktree-backed session")
                 return bad(handler, f"Failed to create worktree: {e}", status=500)
@@ -13973,7 +14022,12 @@ def handle_post(handler, parsed) -> bool:
                 profile=getattr(s, "profile", None),
                 session_id=getattr(s, "session_id", None),
             )
-        return j(handler, {"session": s.compact() | {"messages": s.messages}})
+        payload = {"session": s.compact() | {"messages": s.messages}}
+        if worktree_skipped:
+            # Config-default worktree was skipped (non-git workspace); tell the
+            # client the session is plain so the UI doesn't assume isolation.
+            payload["worktree_skipped"] = worktree_skipped
+        return j(handler, payload)
 
     if parsed.path == "/api/session/compression-recovery/start":
         return _handle_session_compression_recovery_start(handler, body)
--- a/static/boot.js
+++ b/static/boot.js
@@ -303,7 +303,10 @@ async function _maybeBindFreshDefaultWorkspaceSession(prefillIntent=null){
   if(_workspacePanelMode!=='browse') return false;
   if(!S._profileDefaultWorkspace) return false;
   try{
-    await newSession(false, {awaitWorkspaceLoad: true});
+    // worktree:false is explicit and load-bearing — this auto-bind runs on
+    // page load, and a config-level worktree default must never leak a fresh
+    // worktree + branch from simply opening the UI (#6022).
+    await newSession(false, {awaitWorkspaceLoad: true, worktree: false});
     return true;
   }catch(e){
     console.warn('[hermes] failed to bind fresh default workspace session', e);
--- a/static/commands.js
+++ b/static/commands.js
@@ -801,7 +801,9 @@ async function cmdTerminal(){
       const first=(data&&data.workspaces||[])[0];
       S._profileSwitchWorkspace=(data&&data.last)||(first&&first.path)||null;
     }
-    await newSession();
+    // System-minted session (#6022): opening the terminal auto-creates a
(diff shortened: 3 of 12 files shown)


nesquena-hermes:
## 🔬 Gate review (Codex, worktree config default) — cross-repo contract verified, PASS

Thanks for the thorough writeup. I pulled the branch at `21da547de` and read the full `api/routes.py` diff, the three static opt-outs, and cross-checked the agent side, since the whole PR rests on a claim about how the agent CLI resolves `worktree:`.

### The agent contract matches, exactly
The PR body claims the agent honors `worktree: true` from config.yaml as `use_worktree = worktree or w or CLI_CONFIG.get("worktree", False)`. Verified against the agent source — `cli.py:15889`:

```python
use_worktree = worktree or w or CLI_CONFIG.get("worktree", False)
```

So the CLI treats config as the lowest-priority fallback behind the explicit `--worktree`/`-w` flags. The WebUI helper `_worktree_default_from_config` (`api/routes.py:7577-7603`) reproduces that priority order correctly: `handle_post` reads it **only** when the body key is absent (`raw_worktree is None`), and any explicit body value wins verbatim. That's the right layering — a WebUI session and a `hermes` CLI session pointed at the same repo now agree on isolation, which is the #6022 ask.

### Profile-aware read is the correct call, not a bug
Using `get_config_for_profile_home(get_hermes_home_for_profile(profile))` instead of a fixed `load_config()` is right for the WebUI's multi-profile-per-process model. I confirmed `get_config_for_profile_home` (`api/config.py:680`) is the #3294 mtime-tracked path: it returns the cached `get_config()` when the target matches the ambient home (common single-profile case, honors monkeypatched cfg), and only reads a diverging profile's file directly off disk with no global cache mutation. The `profile=None` branch degrades to ambient `get_config()`. Fail-soft `except Exception: return False` is the correct default — config resolution must never 500 session creation.

### The two subtle decisions are both sound
1. **Non-git fallback gated on `worktree_explicit`** (`routes.py:13931-13942`). `create_worktree_for_workspace` → `find_git_repo_root` raises `ValueError` on a non-git workspace. Explicit `worktree: true` keeps today's hard 400; a *config-default* request degrades to a plain session and surfaces `worktree_skipped` in the payload. Without this gate, `worktree: true` in config.yaml would 400 every new session in every non-git directory — a real footgun. Good catch.
2. **Boot/onboarding opt out with explicit `worktree: false`** (`boot.js:306`, `onboarding.js:530`). This is load-bearing: a server-side default must never mint a worktree + branch from merely opening the page, which would compound the #6023 unremovable-worktree pile-up. The `newSession()` contract change (`sessions.js:1365`, `hasOwnProperty(...,'worktree')` → forward verbatim, else omit) implements the three-value model cleanly, and deliberate "New Chat" paths correctly send nothing to inherit the default.

### One thing worth confirming (non-blocking)
The response now carries `worktree_skipped` when a config-default worktree is degraded on a non-git dir. Confirm the client actually reads that key somewhere (a toast or a session badge) — otherwise the user silently gets a plain session when they configured `worktree: true` and may assume isolation. The static tests assert the request-shape contract but I didn't see a consumer of `worktree_skipped` on the render side. Not a blocker; if it's deliberately payload-only for now, a follow-up UI signal would close the loop.

**Verdict: PASS.** The cross-repo contract is faithfully mirrored, the priority order is correct, and both edge cases (non-git degrade, boot opt-out) are handled deliberately. Full gate is green (18/18 shards + browser-smoke + lint). Verification was read-only source review on both repos; I did not execute the branch tests.

nesquena-hermes (changes requested):
## 🔬 Gate review (Codex) — SHIP ONLY WITH FIXES (right direction; the no-leak guard is incomplete)

Thanks @ayxuerui — the design is correct and the intent is exactly right: honor the agent's config `worktree:` default on **deliberate** New Chat while never letting it leak a fresh worktree+branch from non-deliberate paths (#6022). The boot auto-bind + the two system-minted paths you guarded verify clean, deliberate-new-chat honors the default, and an explicit request flag wins. But the gate found the no-leak guard is **incomplete** — there are more automatic session-mint call sites that DON'T pass `worktree:false`, so with a `worktree:true` config default they'd each leak a worktree without deliberate intent (the exact #6022 bug, via sibling paths):

1. **[CORE — strict bool validation] `api/routes.py:7599`** — the config read uses truthiness coercion; a quoted string / integer / list / dict / null `worktree:` value would mis-coerce. Use **strict boolean validation** (only real `true`/`false` honored, anything else → safe no-worktree default) + tests for those shapes.
2. **[CORE — presence detection] `api/routes.py:13916`** — detect an explicit request flag with `"worktree" in body` (not truthiness), so a request that explicitly sends `worktree: null`/absent falls to the config default correctly and an explicit `false` still wins. Add a config-true + request-null test asserting **no** worktree factory call.
3. **[CORE — unguarded auto-mint paths] `static/panels.js:6306, :6342, :7131`** — all three automatic session-mint paths must explicitly pass `worktree:false`.
4. **[CORE — file/folder mint] `static/ui.js:19569, :19600`** — the New File / New Folder session bodies must pass `worktree:false`.
5. **[CORE — terminal mint] `static/commands.js:804`** — the automatic terminal session call should be `newSession(false, {worktree:false})`.
6. **[test gap] `tests/test_issue6022_worktree_ui_static.py:31`** — extend the system-mint regression gate to cover profile switching, workspace binding/prompts, file/folder creation, and terminal creation (currently only boot + the two you guarded); and the "malformed config" test only covers an exception, not malformed/non-boolean values.

The rule to apply consistently: **every non-deliberate `newSession(...)` call site passes `worktree:false`; only the deliberate New Chat leaves it absent to inherit the config default.** Once all the auto-mint siblings are guarded + the config read is strict-boolean, this closes #6022 completely. Rebased clean on current master. Nice work on the direction — it just needs the guard applied to the full set of auto-mint paths (grep for `newSession(` to catch them all).

ayxuerui:
All six items addressed in `340423f7` (kept as a separate commit on top of the rebased feature commit for reviewability):

1. **Strict boolean config read** — `_worktree_default_from_config` now requires `is True`; quoted strings, ints, lists, dicts, and null all fall to the safe no-worktree default. Test: `test_worktree_default_requires_strict_boolean_true` covers `"true"`, `"yes"`, `1`, `[True]`, `{"enabled": True}`, `None`, `0`, `""`.
2. **Presence-based explicit detection** — `worktree_explicit = "worktree" in body`; an explicit `worktree: null` is treated as an explicit non-true (opt-out), never a fall-through to config. Test: `test_explicit_null_beats_config_default_true` asserts the worktree factory is never called with config-true + body-null.
3. **panels.js auto-mints guarded** — profile-switch `newSession`, `promptWorkspacePath`, and `switchToWorkspace` blank-page binds all send explicit `worktree:false`.
4. **ui.js file/folder mints guarded** — `promptNewFile` + `promptNewFolder` session bodies send `worktree:false`.
5. **commands.js terminal mint guarded** — `newSession(false, {worktree: false})`.
6. **Regression gates extended** — `test_issue6022_worktree_ui_static.py` now asserts every auto-mint site individually (boot, onboarding, profile switch, both workspace binds, file, folder, terminal) plus a bare-body sweep (`body:JSON.stringify({workspace:ws})` must not exist in panels.js/ui.js) so a future auto-mint path can't silently omit the key. The malformed-config coverage now includes non-boolean value shapes, not just the exception path.

One deliberate scope note: `cmdGoal` and the busy-send path in messages.js also lazily create sessions, but both fire on the user actively starting work (a /goal command, sending a message) — under the "deliberate intent inherits the default" rule I've left them inheriting, same class as New Chat. Happy to flip them to explicit `false` if you read their intent differently.

Also updated `test_embedded_workspace_terminal.py`'s preflight-ordering assertion for the new cmdTerminal call shape. Full local run across the worktree/session-new/terminal/profile-switch test files: 69 passed; ruff gate clean.

nesquena-hermes:
Shipped in **exp-v0.52.69** 🚀 — thanks @ayxuerui! The re-push closed every finding from the prior review: all auto-mint sites now explicitly opt out (5 in panels.js, 4 in ui.js file/folder, 2 in commands.js terminal), the config read is strict-boolean (`is True`), and presence detection (`"worktree" in body`) keeps an explicit request flag winning over the config default. Codex re-gate SAFE — no path can leak a worktree from the default, no boot regression, 30 focused tests. Closes #6022.

PR #6037, opened 2026-07-13
Show copied feedback for Transparent Stream controls

Stacey2911:
## Thinking Path

Current upstream already creates, reuses, and rebinds Transparent Stream tool and thinking copy controls during live-row reconciliation.

The remaining inconsistency was copied-state feedback. Normal message and thinking copy actions temporarily show a check after success, while Transparent Stream controls were rebound through `_copyEventToClipboard()` without durable ownership of the activated feedback state.

The implementation now treats copied feedback as transient row-owned UI state so it can survive control replacement, clean up safely after detachment, and never be persisted into restored live-turn HTML.

## What Changed

- Added temporary copied/check feedback for Transparent Stream tool and rebound thinking copy controls.
- Shows feedback only after confirmed Clipboard API or fallback-copy success.
- Restores the control's original markup, inline style, title, and accessible label after approximately 1500 ms.
- Uses generation-safe timers so repeated clicks cannot allow stale callbacks to clear newer feedback.
- Stores active feedback on the stable Transparent Stream row rather than the disposable control node.
- Resolves the row's current visible control after reconciliation, so feedback survives control replacement while the copy operation is pending or while feedback is active.
- Stops rendering or restoring feedback when the owning row or visible control has detached, clearing transient state without stale DOM or ARIA writes.
- Sanitizes transient copied markup, style, title, and ARIA state from a cloned live-turn snapshot before serialization.
- Keeps the active copied check at full opacity and restores the original inline style on expiry.
- Preserves existing toast, fallback-copy, keyboard, redaction, payload, reconciliation, and normal message/thinking copy behavior.

## Why It Matters

Transparent Stream copy controls now provide the same visible success feedback as normal chat copy actions.

The feedback remains correct across live reconciliation, does not mutate detached controls, and cannot become permanently serialized into session/live-turn restoration data.

## Contract Routing

Task type: Transparent Stream UI-control feedback and transient live-turn state hygiene.

### Touched behavior:

- Transparent event/tool copy controls
- Thinking copy controls rebound through Transparent Stream reconciliation
- Clipboard success/fallback feedback
- Live-turn snapshot sanitization of transient copied UI state

### Not changed:

- backend APIs
- provider-facing payloads
- persisted session schema
- activity-mode selection
- chat-bubble action disclosure
- row ordering or reconciliation architecture
- motion/fade behavior
- normal `copyMsg()`
- normal `_copyThinkingText()`

### Automated:

```bash
node --check static/ui.js
./scripts/test.sh tests/test_issue5367_transparent_live_row_reconcile.py tests/test_issue3820_chat_activity_display_mode.py tests/test_ui_card_animation.py tests/test_issue1096_copy_buttons.py
git diff --check
```

### Result:

Focused tests: 71 passed in 3.28s
JavaScript syntax checks passed
Whitespace/diff validation passed
GitHub CI: All 22 checks completed SUCCESS (Tests, Browser smoke, Conversation lifecycle, lint, Greptile Review)

### Behavioral coverage includes:

confirmed copy success and fallback-copy success;
failure and empty-payload paths without false feedback;
replacement before clipboard resolution;
replacement during active feedback;
repeated-click generation safety;
detached-row cleanup with no post-detach DOM/ARIA writes;
real _refreshTransparentLiveRow(existing, candidate) reconciliation;
copy-success → snapshot → restore → expiry;
multiple simultaneous active copied controls;
clone-only snapshot sanitization without mutating active live feedback;
exact restoration of markup, inline style, title, and ARIA state;
one functional control after reconciliation;
copy activation does not expand/collapse the row.

### Manual browser validation:

Confirmed Transparent Stream tool copy shows a temporary check and restores.
Confirmed rebound thinking copy shows the same feedback.
Confirmed repeated clicks refresh the feedback timer.
Confirmed normal assistant/user and normal thinking copy behavior remain unchanged.
Confirmed no chat-bubble click-reveal behavior was added.
Risks / Follow-ups
Low implementation scope: only existing Transparent Stream copy feedback and its reconciliation/snapshot lifecycle are affected.
Complete copy failure does not show a success check.
Existing success toasts are retained.
Snapshot sanitization modifies only a cloned live turn; active live feedback and its timer remain untouched.
The obsolete broader copy-controls branch was not reused; this PR was rebuilt from current upstream.
Touch/tablet message-action disclosure remains separate UX work.

### Planning, architecture, debugging strategy, review, PR decomposition, and contribution workflow:

Sol — OpenAI ChatGPT, GPT-5.6 Pro

### Implementation and code-level review:

Codex CLI 0.144.1
GPT-5.6 Sol High

### Local orchestration and environment inspection:

HermesAgent via LM Studio
qwen3.6-27b@q4_k_m

## Manual validation:

User-performed browser and local-runtime validation

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -10742,7 +10742,51 @@ function _transparentToolSummary(tc){
   if(target) return target;
   return '';
 }
-function _copyEventToClipboard(row){
+function _showTransparentCopiedFeedback(control){
+  if(!control) return;
+  let state=control._transparentCopiedFeedback;
+  if(!state){
+    state={
+      innerHTML:control.innerHTML,
+      color:control.style?control.style.color:undefined,
+      styleCssText:control.style&&typeof control.style.cssText==='string'?control.style.cssText:null,
+      titleAttr:control.getAttribute?control.getAttribute('title'):null,
+      ariaLabel:control.getAttribute?control.getAttribute('aria-label'):null,
+      timer:null,
+      generation:0,
+    };
+    control._transparentCopiedFeedback=state;
+  }
+  if(state.timer) clearTimeout(state.timer);
+  state.generation+=1;
+  const generation=state.generation;
+  const copiedLabel=t('copied')||'Copied';
+  control.innerHTML=typeof li==='function'?li('check',11):'✓';
+  if(control.style) control.style.color='var(--accent)';
+  if(control.setAttribute) control.setAttribute('title',copiedLabel);
+  else control.title=copiedLabel;
+  if(control.setAttribute) control.setAttribute('aria-label',copiedLabel);
+  state.timer=setTimeout(()=>{
+    if(control._transparentCopiedFeedback!==state||state.generation!==generation) return;
+    control.innerHTML=state.innerHTML;
+    if(control.style){
+      if(state.styleCssText!==null) control.style.cssText=state.styleCssText;
+      else control.style.color=state.color||'';
+    }
+    if(state.titleAttr===null){
+      if(control.removeAttribute) control.removeAttribute('title');
+    }else{
+      if(control.setAttribute) control.setAttribute('title',state.titleAttr);
+    }
+    if(state.ariaLabel===null){
+      if(control.removeAttribute) control.removeAttribute('aria-label');
+    }else if(control.setAttribute){
+      control.setAttribute('aria-label',state.ariaLabel);
+    }
+    delete control._transparentCopiedFeedback;
+  },1500);
+}
+function _copyEventToClipboard(row,control){
   if(!row) return;
   const type=row.getAttribute('data-event-type');
   let text='';
@@ -10785,9 +10829,12 @@ function _copyEventToClipboard(row){
   }else{
     text=row.textContent||'';
   }
+  if(!text) return;
+  const copied=()=>_showTransparentCopiedFeedback(control);
   const fallback=()=>{
+    let ta=null;
     try{
-      const ta=document.createElement('textarea');
+      ta=document.createElement('textarea');
       ta.value=text;
       ta.setAttribute('readonly','');
       ta.style.position='absolute';
@@ -10795,14 +10842,17 @@ function _copyEventToClipboard(row){
       document.body.appendChild(ta);
       ta.select();
       const ok=document.execCommand('copy');
-      document.body.removeChild(ta);
+      if(ok) copied();
       if(typeof showToast==='function') showToast(ok?(t('copied')||'Copied'):(t('copy_failed')||'Copy failed'),1600);
     }catch(_){
       if(typeof showToast==='function') showToast(t('copy_failed')||'Copy failed',2000,'error');
+    }finally{
+      if(ta&&ta.parentNode) ta.parentNode.removeChild(ta);
     }
   };
   if(navigator&&navigator.clipboard&&navigator.clipboard.writeText){
     navigator.clipboard.writeText(text).then(()=>{
+      copied();
       if(typeof showToast==='function') showToast(`${t('copied')||'Copied'} ${label}`,1600);
     }).catch(fallback);
   }else{
@@ -10816,13 +10866,15 @@ function _attachCopyButton(header){
     btn.classList.add('transparent-event-copy');
     btn.setAttribute('role','button');
     btn.setAttribute('tabindex','0');
-    btn.setAttribute('aria-label',t('copy')||'Copy');
     btn.setAttribute('data-transparent-copy','1');
-    btn.title=t('copy')||'Copy';
+    if(!btn._transparentCopiedFeedback){
+      btn.setAttribute('aria-label',t('copy')||'Copy');
+      btn.title=t('copy')||'Copy';
+    }
     const handler=function(ev){
       ev.stopPropagation();
       ev.preventDefault();
-      _copyEventToClipboard(header.closest('.transparent-event-row'));
+      _copyEventToClipboard(header.closest('.transparent-event-row'),btn);
     };
     btn.onclick=handler;
     btn.onkeydown=function(ev){
--- a/static/ui.js
+++ b/static/ui.js
@@ -8938,7 +8938,33 @@ function snapshotLiveTurnHtmlForSession(sid){
   const turn=$('liveAssistantTurn');
   if(!turn) return;
   if(turn.dataset&&turn.dataset.sessionId&&turn.dataset.sessionId!==sid) return;
-  INFLIGHT[sid].liveTurnHtml=turn.outerHTML;
+  let snapshotTurn=turn;
+  const sourceControls=turn.querySelectorAll
+    ? Array.from(turn.querySelectorAll('.transparent-event-copy,.thinking-copy-btn'))
+    : [];
+  if(sourceControls.some(control=>!!control._transparentCopiedFeedbackNormal)){
+    // Copy-success feedback is transient, while its normal presentation lives
+    // only on element properties. Sanitize an independent clone so switching
+    // sessions cannot persist the check/Copied/accent presentation, without
+    // mutating the visible control or disturbing its active feedback timer.
+    snapshotTurn=turn.cloneNode(true);
+    const clonedControls=Array.from(snapshotTurn.querySelectorAll('.transparent-event-copy,.thinking-copy-btn'));
+    sourceControls.forEach((source,index)=>{
+      const normal=source._transparentCopiedFeedbackNormal;
+      const clone=clonedControls[index];
+      if(!normal||!clone) return;
+      clone.innerHTML=normal.innerHTML;
+      if(clone.style){
+        if(normal.styleCssText!==null) clone.style.cssText=normal.styleCssText;
+        else clone.style.color=normal.color||'';
+      }
+      if(normal.titleAttr===null) clone.removeAttribute('title');
+      else clone.setAttribute('title',normal.titleAttr);
+      if(normal.ariaLabel===null) clone.removeAttribute('aria-label');
+      else clone.setAttribute('aria-label',normal.ariaLabel);
+    });
+  }
+  INFLIGHT[sid].liveTurnHtml=snapshotTurn.outerHTML;
 }
 
 function _liveAssistantSegmentTextLength(seg){
@@ -10804,7 +10830,10 @@ function _showTransparentCopiedFeedback(control,row,opts){
     }
     const copiedLabel=t('copied')||'Copied';
     target.innerHTML=typeof li==='function'?li('check',11):'✓';
-    if(target.style) target.style.color='var(--accent)';
+    if(target.style){
+      target.style.color='var(--accent)';
+      target.style.opacity='1';
+    }
     if(target.setAttribute) target.setAttribute('title',copiedLabel);
     else target.title=copiedLabel;
     if(target.setAttribute) target.setAttribute('aria-label',copiedLabel);
--- a/static/ui.js
+++ b/static/ui.js
@@ -10742,48 +10742,88 @@ function _transparentToolSummary(tc){
   if(target) return target;
   return '';
 }
-function _showTransparentCopiedFeedback(control){
-  if(!control) return;
-  let state=control._transparentCopiedFeedback;
-  if(!state){
-    state={
-      innerHTML:control.innerHTML,
-      color:control.style?control.style.color:undefined,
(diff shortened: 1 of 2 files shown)


Stacey2911:
Addressed the replaced-control feedback race in a follow-up commit. Copied-state ownership now survives Transparent Stream reconciliation both while the clipboard operation is pending and during the active feedback window. Added executable coverage for replacement-before-success, replacement-during-feedback, repeated clicks, fallback success, and failure without false feedback.

nesquena-hermes (changes requested):
## 🔬 Correctness gate (Codex + stream-gate, self-rebased clean off master) — SHIP ONLY WITH FIXES (aria + reconcile solid; a stale-node write after mid-feedback reconciliation)

Thanks @Stacey2911 — most of this is solid: the ARIA state machine restores correctly (null/absent attributes restore as absent; a repeat copy keeps the original stash rather than stashing "Copied"), live-row reconciliation is unaffected (replacement controls rebind correctly, tool data + disclosure state survive, the #5367 file passes 8/8), both `.transparent-event-copy` and `.thinking-copy-btn` get feedback with one functional control, and there's no `window.*` collision (handler assignment is idempotent). Stream-regression gate is GREEN across modes. But one real defect on the reconciliation edge blocks it:

1. **[STALE-NODE WRITE — timer fires against a detached control] `static/ui.js:10822`** — when live-row reconciliation **replaces/detaches** the row or copy control before the copied-feedback timer expires, the expiry still writes `title` and `aria-label` onto the **detached node** (verified via a focused probe against the teardown paths at `sessions.js:1761` + `ui.js:15290`). It doesn't throw or leak a persistent timer (state is deleted after expiry), but it's mutating orphaned DOM. **Fix:** before rendering or restoring feedback, detect a detached row/control — if detached, delete `_transparentCopiedFeedback` and return **without any DOM or ARIA write**.

2. **[test — doesn't exercise the real reconcile] `tests/test_issue5367_transparent_live_row_reconcile.py:2022`** — the test fails on master (helper absent) so it's non-vacuous for the feature's presence, but it uses a **synthetic** control-swap rather than executing the real reconcile function, so it doesn't actually cover the detached-node case above. **Fix:** replace the synthetic swap with a real pending-copy → `_refreshTransparentLiveRow(existing, candidate)` → expiry sequence, and assert that a wholly detached row receives **no** stale attribute writes.

Per-area: aria state machine PASS, reconcile behavior PASS, tool+thinking PASS, globals/listeners PASS, stream-gate GREEN — but timer/reference lifetime FAIL (the detached-node write) and the test doesn't behaviorally hit it. Once the detached-guard lands and the test drives the real reconcile path, this should clear. The visual/UX sign-off (the actual copied-check affordance) is a separate maintainer step and will happen before ship. Self-rebased clean on current master.

Stacey2911:
Addressed the maintainer correctness blocker in a follow-up commit. Copied feedback now clears and exits without DOM or ARIA writes when its row/control is detached. The regression test now drives the real pending-copy → `_refreshTransparentLiveRow(existing, candidate)` → detach → timer-expiry sequence and asserts no post-detach markup, style, title, or aria writes.

nesquena-hermes:
## 🔬 Re-gate (Codex, Transparent Stream copy feedback) — both blockers verified RESOLVED at `b36cd3ac70`

Thanks @Stacey2911. I pulled the branch and re-read the full `_showTransparentCopiedFeedback` helper against `origin/master` plus the new detached-reconcile test. Both kickback blockers are closed and the fix stayed on the reconciliation seam.

### Blocker 1 (stale-node write after mid-feedback reconciliation) — RESOLVED
On the prior push the 1500ms expiry callback wrote `title` / `aria-label` onto the detached node when live-row reconciliation replaced/removed the control before the timer fired. The rework keeps the feedback lifetime on the persistent row (`owner`) and re-resolves the *currently connected* control on every render and expiry via `connectedControl()` (`static/ui.js:10761-10766`):

```js
const connectedControl=()=>{
  const target=currentControl();
  if(!target||target.isConnected!==true) return null;
  if(feedbackRow&&feedbackRow.isConnected!==true) return null;
  return target;
};
```

The critical detail I checked is the expiry path itself (`static/ui.js:10845-10857`): when the row/control is detached, `connectedControl()` returns `null`, and the callback clears the timer and deletes `_transparentCopiedFeedback` **without any DOM or ARIA write** — no orphaned-node mutation:

```js
state.timer=setTimeout(()=>{
  if(owner._transparentCopiedFeedback!==state||state.generation!==generation) return;
  const expiryTarget=connectedControl();
  if(!expiryTarget){
    state.timer=null;
    delete owner._transparentCopiedFeedback;
    return;
  }
  state.timer=null;
  restoreControl(expiryTarget);
  delete owner._transparentCopiedFeedback;
},1500);
```

The `generation` guard also correctly makes a stale timer from a superseded copy no-op, so repeated clicks can't restore an older state.

### Blocker 2 (test used a synthetic swap, not the real reconcile) — RESOLVED
`test_transparent_rebound_copy_controls_report_only_confirmed_success` now drives the production sequence: pending copy → real `_refreshTransparentLiveRow(race.row, candidate)` → detach → `runTimer()` expiry (`tests/test_issue5367_transparent_live_row_reconcile.py:2225`, `:2252`, `:2273`). It installs a `_trackDetachedWrites` probe on both the detached row and control (`:2269-2271`) that records any attribute write while `isConnected` is false (`:1912`), then asserts the detached expiry produced no writes and didn't throw:

```python
"detachedExpiry": {
    "rowConnected": False,
    "controlConnected": False,
    "feedbackState": False,
    "writes": [],
    "threw": False,
    ...
},
```

The `replacement-before-success` and `replacement-during-feedback` cases also assert the *replacement* control inherits the copied check (title/aria = `Copied`) while `activatedControlDetached` is `True`, and `oneFunctionalControl` / `activationDidNotToggle` confirm no duplicate binding.

### Boundary matrix re-checked
Fallback-copy success still shows the check, hard failure shows no false success (`failure` assertion at `:2357`), empty-copy is skipped (`emptyCopySkipped`), and the thinking-control path (`.thinking-copy-btn`) gets the same feedback (`thinkingReplacementRace`). CI is green on `b36cd3ac70` (15 test shards + browser-smoke + lint + Greptile).

**Verdict: PASS.** This resolves the detached-node write class without reopening the reconcile-survival behavior the PR originally fixed. Recommend merge pending the maintainer's visual sign-off on the actual copied-check affordance. Verification was read-only source review against the CI-green result; I did not execute the branch.

nesquena-hermes (changes requested):

**Contributor SHA:** `b36cd3ac70d22d16ca58a94f0e06b4d0c66c78a9`

The stale-node/timer issue from the prior head is fixed, and the core stream remains clean. Two exact-head blockers remain.

### 1. Switch-away snapshot can make the copied state permanent

`snapshotLiveTurnHtmlForSession()` serializes `turn.outerHTML` while copied feedback is active. The normal glyph/style/title/ARIA are stored only in `_transparentCopiedFeedbackNormal`, which is a JS property and is not serialized.

A real browser probe through `snapshotLiveTurnHtmlForSession()` → `restoreLiveTurnHtmlForSession()` reproduced this 1.65 seconds after restore:

- restored HTML is still the copied check SVG, not the original copy SVG
- `title` and `aria-label` remain `Copied!`
- inline color remains `var(--accent)`
- computed opacity remains `0.28`
- no `_transparentCopiedFeedback` or normal-state record remains, so nothing can restore it later

**Fix:** serialize a clone of the live turn. For each feedback-active copy control, apply the original control's `_transparentCopiedFeedbackNormal` markup/style/title/ARIA to the corresponding clone before storing `liveTurnHtml`. Do not mutate the live DOM. Add a copy-success → snapshot → restore → expiry regression test.

### 2. Copied check is too faint without hover

`.transparent-event-copy` rests at `opacity:.28`. `renderCopiedControl()` changes only the glyph and color, so the confirmation is barely visible after pointer exit and on touch devices. The real mobile/desktop captures confirm the copied state remains at computed opacity `0.28`.

**Fix:** set `target.style.opacity='1'` during copied feedback. The existing `style.cssText` snapshot/restore already restores normal opacity automatically.

### Evidence

- focused reconcile/copy tests: **8 passed**
- crown-jewel stream gate: **GREEN** in transparent, compact, and hide-all modes, including switch/reload mid-stream
- full sandbox suite: **13,453 passed**, 14 known sandbox/browser failures + 2 packaged-runtime setup errors
- matched real-page desktop/mobile before/copied/restored captures: layout-stable, one control, at-rest parity
- Codex: **SHIP ONLY WITH FIXES** (snapshot serialization)
- Fable: **SHIP-WITH-UX-FIXES** (opacity)

Please re-request review after both fixes and the snapshot regression test land.

Stacey2911:
Addressed both exact-head blockers in a follow-up commit. Live-turn snapshots now serialize a sanitized clone with active copied controls restored to their normal markup/style/title/ARIA without mutating the live DOM. Active copied feedback now forces opacity to 1 and restores the original inline style on expiry. Added executable copy-success → snapshot → restore → expiry and multiple-active-control coverage, plus a lifecycle scan for persistence, accessibility, timer, reconciliation, and detached-node regressions.

nesquena-hermes:
## 🟢 Gate certification — GREEN

**Exact PR head:** `d28768cb307995c5725f1123547227674385bcea`  
**Gate integration base:** `302843c89f22d1ce2ad10619acd33382c9f48111` (clean rebase, gate head `be906393da19130cfd86a78a429cd4680e56bd58`)  
**Result:** `gate-pass`

The two blockers from the earlier certificate are closed at this head:

1. `snapshotLiveTurnHtmlForSession()` now sanitizes the cloned snapshot, not the connected live subtree. A real-browser switch/restore probe showed the live copied check stayed active while the stored and restored controls retained the exact normal icon, inline style, title, and aria-label.
2. Copied feedback now forces `opacity: 1` while active, follows control replacement on live-row reconciliation, and restores the original `cssText` and accessibility attributes after 1.5 s. Desktop and 390×844 mobile probes both showed one control, an opaque accent check, then the normal copy glyph at the stylesheet's resting opacity (`0.28`).

### Evidence

- Patch integrity: the cleanly rebased contribution has the same stable patch ID as the PR head.
- Static gates: `git diff --check`, `node --check static/ui.js`, runtime ESLint, and undefined-scope gate all passed.
- Focused regression file: **8 passed** in the no-network/no-secrets sandbox.
- Full serial suite ran to completion in the required sandbox: **13,434 passed**, 82 skipped. The sandbox-only remainder was rechecked rather than dismissed: the same git/HOME/Playwright/wheel capability failures reproduce on the exact master control; all **35/35** affected tests pass in the normal test environment, and the one order-sensitive keyless-onboarding failure passes **16/16** isolated on both PR and master in the sandbox. No PR-introduced test failure remains.
- Crown-jewel stream regression gate: **GREEN across `transparent_stream`, `compact_worklog`, and `hide_all_activity`**, including during-stream, post-settlement, switch-away/back, reload-after-done, and reload-mid-stream invariants.
- Current-head CI: **22/22 successful**.
- Codex correctness gate: **SAFE TO SHIP**.
- Senior correctness gate: **APPROVE**, no MUST-FIX findings.
- Fable UX gate: **SHIP-WITH-UX-FIXES**. Its one verified minor note is redundant success feedback: `_copyEventToClipboard()` now shows both the new inline check and the pre-existing success toast. That is non-blocking under the certifier runbook, but removing the two success-path toasts while retaining failure toasts would match the app's other copy controls and ChatGPT/Claude/GitHub convention.

This certificate supersedes the earlier RED result at `b36cd3ac70d2`. The code gate is green. Because this is visible crown-jewel chat UI, final maintainer visual sign-off remains separate from this certification.

nesquena-hermes:
Shipped in **exp-v0.52.196** — thanks @Stacey2911! 🎉

The Transparent Stream tool-event and thinking copy buttons now flash a check for ~1.5s on a successful copy, matching the app's existing copy feedback.

Nice work addressing both prior gate blockers — the switch-away snapshot no longer makes the copied state permanent (clone + restore before serialize), and the copied check is no longer faint (opacity:1 during feedback), which was the key touch-UX win.

Gate: Codex adversarial re-gate **SAFE TO SHIP** (reproduced all three fixes — snapshot/restore, opacity revert, stale-node guard — via focused probes; no new regression on the transparent-stream surface), Fable UX **SHIP-UX** (measured the full opacity/color lifecycle in Chromium, zero layout shift, matches convention), focused tests 8/8, full suite green, maintainer visual approval.

Issue #5305, opened 2026-07-01
bug(sidebar): delegate subagent sessions may still appear as top-level orphans (post-#5244 edge cases)


claw-io:
## Note (upstream sync 2026-07-01)

**PR [#5244](https://github.com/nesquena/hermes-webui/pull/5244)** (merged 2026-06-30) fixed the common case: delegated subagent rows now stack under a visible WebUI parent via `static/sessions.js` `_attachChildSessionsToSidebarRows` (orphan only when `parentIsExternal`, ~5752–5770).

Please **close as duplicate of #5244** if you are on a current release that includes that change and nesting works. Keep open if you still see top-level **"Subagent Session"** pollution on current `master`.

---

## Summary (historical / edge cases)

When `delegate_task` children had `parent_session_id` and `source=subagent`, metadata set `_cross_surface_child_session` because `webui` ≠ `subagent` (`api/agent_sessions.py` ~1082–1085). Before #5244, `sessions.js` promoted all such children to top-level orphans.

**Remaining edge cases (if any):**
- Parent not in the same sidebar render (profile filter, archived parent, project filter) → child may still appear as orphan (`~5784–5786`).
- Server still sets `_cross_surface_child_session` for all cross-source pairs; nesting depends on client `parentIsExternal` heuristic.
- Delegation from non-WebUI parents (CLI/cron) with external parent markers.

## Code pointers

- `api/agent_sessions.py` `read_session_lineage_metadata` ~1078–1089
- `static/sessions.js` `_attachChildSessionsToSidebarRows` ~5720–5788
- Tests: `tests/test_session_lineage_collapse.py` `test_cross_surface_subagent_child_stacks_under_visible_webui_parent`

## Related

ROADMAP #1004, #4293, #5244

PR #6041, opened 2026-07-13
fix(#6027): unify sidebar lineage identity

### LINKED ISSUE #6027: WebUI: sidebar lineage unstable with delegated/subagent children + forks — rows reflow, forks detach from root, sub-agents blink out (residual after #5699, exp-v0.52.60)
Reported by **b3nw** in Discord `#report-bugs` (Jul 13 2026) on **exp-v0.52.60 · Experimental**.

> "I'm not sure if this is regression or just the fix never stuck, still seeing the sidebar change around when delegated/sub-agent sessions are spawned/in-progress."

## Context — distinct from the already-fixed #5699

#5699 ("parent row bucket-jumps Today↔Older tied to child's streaming state") was **fixed and closed 2026-07-09** by removing the `_sidebar_activity_at` bump in `bubbleSidebarState()`. I verified **exp-v0.52.60 contains that fix** (`git merge-base --is-ancestor 745e0be0d exp-v0.52.60` → true; shipped exp-v0.52.1). So the specific **date-bucket-jump** vector #5699 named is closed.

But #5699's root-cause writeup identified **two coupled paths**, and the fix only closed the first:

1. ✅ **Fixed:** `bubbleSidebarState()` bumped the parent's `_sidebar_activity_at` from the streaming child → parent jumped date buckets. (Removed.)
2. ❓ **Still live / other vectors:** the parent/child rows still visibly "change around" while a child streams. b3nw is on the fixed build and still sees movement, so this is a **separate mechanism** the #5699 fix didn't cover — filing fresh rather than reopening #5699.

## What still moves (source trace, confirm against current head `static/sessions.js`)

Candidate mechanisms for the residual "changes around":

- **Child/subagent row lifecycle.** A delegated/subagent session likely renders as its **own top-of-list row while it streams** (fresh `created_at` + running-sort-rank 1 via `_sessionRunningSortRank`, L6312), then **collapses into the parent's lineage** (`_attachChildSessionsToSidebarRows` / `_collapseSessionLineageForSidebar`) once it settles — so a row appears at the top mid-run and disappears/merges on completion. That is a genuine reflow the #5699 parent-bucket fix never touched.
- **Streaming-class toggle driving FLIP reflow.** `_renderOneSession` sets `isStreaming = ownStreaming || !!s._child_session_streaming` (L7765) purely for the CSS `.streaming` class. If the parent's streaming class toggling participates in the FLIP/reflow animation pass (`_playSessionRowsReflowFromPositions`), the row can visibly shift/animate on child start/stop even without a sort-position change.
- **Running-sort-rank on the child.** `_sessionRunningSortRank` (L6312-6314) boosts anything effectively-streaming to rank 1 in `_sessionSidebarSortCompare` (L6317). For the **child row itself** (before it's collapsed into the parent), this pins it to the top while streaming, then it drops — reflow on both transitions.

## Repro
1. A conversation that spawns a delegated/subagent (child) session.
2. Watch the sidebar while the child is generating vs after it completes.
3. Observe rows re-ordering / a row appearing then merging into the parent lineage (no longer the Today↔Older bucket jump #5699 fixed, but still visible movement).

## Ask
Determine which residual vector is producing the movement on exp-v0.52.60 (child-row-appears-then-collapses vs streaming-class FLIP reflow vs child running-sort-rank), then stabilize the sidebar so an in-progress child doesn't reflow the parent's row. Ideally the parent stays put and the child's activity surfaces via the existing child-streaming indicator/attention badge rather than a position change.

## Environment
- WebUI: exp-v0.52.60 (experimental)
- Setup: conversation with a delegated/subagent child session

## Related
- #5699 (CLOSED, fixed) — the date-bucket-jump vector; this is the residual sidebar-movement the fix didn't cover.

rodboev:
## Thinking Path

- The sidebar was deciding lineage ownership through a callback at several consumers, while collapse and archived suppression still used raw IDs before scope was applied.
- The rebuild gives one render-local index ownership, graph validity, project resolution, and scoped identity, then routes collapse, attachment, suppression, filtering, and counts through it.
- Writable rows keep their exact `project_id`, including `null`; only read-only delegated projections inherit ancestry. Forks are excluded from inheritance because forks also carry `child_session` relationship metadata.
- The rework keeps that identity authority operation-scoped through list application, rendering, optimistic removal, deletion, visits, message completion, and persisted unread state.
- Isolated profile reads reject explicit foreign names before home resolution, so a requested profile cannot read another profile's state database.

## What Changed

- `static/sessions.js`: replace repeated ancestry resolution and partial scope helpers with one memoized snapshot index, and use its composite identity across lineage consumers.
- Delegated projections with duplicate raw IDs keep their parent-reference discriminator, and runtime state resolves identity against the full sidebar cache and archived references.
- Unassigned forks retain null project ownership and stay outside project-owned lineage attachment.
- api/routes.py scopes hidden archived references by source, profile, project, and session identity, and routes lineage reports to the requested profile state database.
- Existing sidebar payload work retains the server-canonical profile_scope field; client runtime, viewed-count, completion-unread, and optimistic-deletion state now use the same scoped identity.
- `static/messages.js` carries the active row and operation context through viewed, completion-unread, and clear paths, including batch deletion and duplicate scoped IDs.
- Tests: add production-composed regressions for writable Unassigned forks, scoped duplicate IDs, archived suppression, cycle/diamond/self-root behavior, large nested lineages, isolated profile reads, and message lifecycle cleanup.

## Why It Matters

Removing a fork from a project keeps it Unassigned, equal lineage IDs stay isolated across scopes, archived rows stop suppressing unrelated profiles, and large sidebars avoid repeated ancestry walks.

## Verification

The prior exact head dea161a95ebff97b79d1c271e79f5eab83a17664 is the RED baseline recorded in the maintainer gate. It reproduced writable null forks inheriting a project, duplicate lineage IDs merging across scopes, archived suppression leaking across profiles, and nested ancestry regressing in scale. The final reworked head passes 485 tests across the affected focused files, including isolated profile rejection and root-alias preservation, one-index list application, duplicate scoped runtime state, message completion reconciliation, batch cleanup, new-session and message-load compatibility, and the 2,000-row graph bounds. Runtime JavaScript validation and CI-owned checks remain green locally where runnable.

## Risks / Follow-ups

Malformed or ambiguous ancestry stays isolated and does not share the valid Unassigned namespace. The index is operation-local; the client state keys are scoped without adding backend schema or server-side derived state. The security-sensitive path-handling change is read-only and fail-closed, verified by foreign-profile, profileless-root, enrichment, and lineage-report tests. The current served-browser proof was not rerun for this runtime-only rework; the existing before/after pair remains the visual evidence for the unchanged sidebar result.

## Contract Routing

Task type: bug fix
Touched areas: sidebar session lineage, project ownership, scoped collapse and attachment, archived-child suppression, large-list rendering
Relevant public docs:
- https://github.com/nesquena/hermes-webui/issues/6027
- https://github.com/nesquena/hermes-webui/issues/6027#issuecomment-4961584558
Scope boundaries: frontend sidebar composition plus read-only archived-reference and lineage-report routing using existing payload signals; no backend schema or server-side persistence changes
Evidence needed before claiming done: foreign isolated reads fail closed before home resolution, the profileless root alias remains valid, one runtime index serves list application and render consumers, duplicate scoped IDs remain isolated through message and sidebar state, and the affected focused tests pass

## Upstream

Closes #6027.

Root-cause direction and acceptance criteria: https://github.com/nesquena/hermes-webui/issues/6027#issuecomment-4961584558

## Screenshots

The isolated before/after sidebar pair is hosted on the fork `screenshots` branch. It shows the root, delegated child, and writable fork grouping; this rework preserves that rendered result.

Before, ungrouped sidebar state

After, stable root and child grouping

## Model Used

GPT 5.5 via Codex CLI


nesquena-hermes (changes requested):
## 🔬 Gate review (Codex, crown-jewel sidebar) — SHIP ONLY WITH FIXES (core stabilization works; active-profile filter detaches a fork in the all-profiles view)

Thanks @rodboev — the durable-lineage-Set approach is the right fix for #6027, and most of it verifies clean:
- ✅ Normal-profile forks + delegated children stay attached while streaming state toggles (no detach/reflow/blink-out — the #6027 symptom).
- ✅ Project / unassigned (`NO_PROJECT_FILTER`) / CLI↔WebUI boundaries all behave correctly.
- ✅ No session dropped or duplicated across buckets; parent timestamps + Today/Older bucketing unchanged; the Set is rebuilt each render (a removed child loses its decoration next rebuild — no stale leak). 106 focused sidebar tests pass.

One real regression on the multi-profile view:

1. **[CORE — active-profile filter detaches a fork] `static/sessions.js:7336-7345` + `:7445-7449`** — the durable-lineage resolution filters children by the **active profile**, so in the *all-profiles-visible* view a fork whose parent belongs to a non-active profile is **detached** and rendered as a separate top-level row (master correctly nests it under its parent). That's the same detach-from-root class #6027 is about, just triggered by profile scope instead of streaming state. **Fix:** drop the active-profile strict filtering; resolve durable parent by **each child/parent pair's own normalized profile/project/source scope** (trust server scoping for single-profile payloads; for all-profile payloads resolve the parent within the child's normalized scope), and **preserve renamed-root aliases** so an all-profile subtree stays nested.

2. **[test gap] `tests/test_session_lineage_collapse.py:1053`** — the new test doesn't exercise `lineageScope` or streaming, and its reference array isn't the production partition shape (`referenceRaw.push(s)` before every `sessionsRaw.push(s)`), so it passed while missing this regression. Add production-shaped cases with explicit `lineageScope`: a renamed-root `default` row and an active-A / all-profiles-visible-B fork; toggle a delegated child's streaming state and assert stable nesting, exactly-once membership, and no cross-profile/project attachment.

Once parent resolution keys off each pair's own scope (not the active profile) and the test covers the all-profiles fork + streaming toggle, this should gate clean — it's genuinely close, and the core anti-reflow mechanism is sound. Nice pickup on #6027.

rodboev:
Thanks, that was the remaining hole.

1. `_renderSidebarRowsFromRawSessions()` now builds the durable lineage authority by each row's own source, project, and normalized profile, so an all-profiles render no longer uses the active profile to drop a fork's real parent.
2. `_attachChildSessionsToSidebarRows()` still routes the decision through `_isForkWithResolvableParent()`, but each caller now pulls the scope-local authority instead of a render-global set, so the fix stays on one seam.
3. I replaced the earlier narrow/reference-window regression with production-shaped coverage for the all-profiles case, delegated-child streaming toggles, renamed-root/default-profile aliasing, and the preserved cross-scope boundaries.

Commit `890d0b05` has the rework.

nesquena-hermes:
## 🔬 Re-gate (Codex, crown-jewel sidebar) — both blockers verified RESOLVED at `890d0b05`

Thanks @rodboev. I pulled the branch and re-read the full `static/sessions.js` diff against `origin/master` plus the four test files. Both kickback blockers are closed, and the fix stayed on one seam as you described.

### Blocker 1 (active-profile filter detached an all-profiles fork) — RESOLVED
On master the attach path resolved every fork's parent against a single render-global `sessionIdsInList` set, so an all-profiles rebuild used the *active* profile to drop a non-active parent. The rework replaces that with a scope-keyed authority built in `_renderSidebarRowsFromRawSessions` (`static/sessions.js:7383-7407`) and routes each fork through its **own** scope:

```js
const scopeKeyForSession=(session)=>{
  ...
  return `${isCli?'cli':'webui'}\u0000${project}\u0000${normalizedProfileForScope(session.profile)}`;
};
const durableLineageIdsByScope=new Map();
for(const session of durableRows){
  ... // NOTE: only isCli + project gate the row here; profile is NOT a filter
  durableLineageIdsByScope.get(key).add(session.session_id);
}
```

The critical detail I checked: the `lineageScope` guard in that loop filters `durableRows` by `isCli` and `project` only — it no longer excludes by active profile, so a fork whose parent lives in a non-active profile now lands in the same scoped set as its parent. All three decision points inside `_attachChildSessionsToSidebarRows` (top-level filter :6790, `attachDepthFor` :6831, `isForkChild` :6872) now call `_isForkWithResolvableParent(x, sessionsFor(x))` instead of the render-global set, so the whole fix lives on one seam. `sessionIdsInList` survives only as the fallback inside `sessionsFor`.

### Blocker 2 (test didn't exercise `lineageScope`/streaming, wrong reference shape) — RESOLVED
`test_lineage_scope_keeps_all_profile_fork_grouped_across_rebuilds` (`tests/test_session_lineage_collapse.py:1054`) is the direct regression: active profile `a`, a `b`-profile fork+delegated child, rendered with `{isCli:false, profile:'a'}`. It asserts both the streaming-on *narrow* reference and the streaming-off *complete* reference keep `b-fork`+`b-child` nested under `b-root` (never detached to top level), plus exactly-once membership across the flattened tree:

```python
assert result["narrow"] == [
    {"id": "a-root", "childStreaming": False, "children": []},
    {"id": "b-root", "childStreaming": True, "children": ["b-fork", "b-child"]},
]
...
assert len(flattened) == len(set(flattened))
```

`test_default_and_renamed_root_profiles_share_lineage_scope` (:1118) also covers the renamed-root alias path via `_cronProfileNameIsRootAlias` normalization (a `default`-profile fork attaches under a `Renamed root` parent). And the earlier cross-project archived-parent case at :885 was correctly flipped to assert the fork **survives** even if a bad caller passes a foreign-project parent in the reference set — scope keying now ignores it. That's a stronger invariant than the old "wrongRows == []".

### Boundary matrix re-checked
Pinned-fork-stays-top-level, `NO_PROJECT_FILTER`/unassigned, CLI↔WebUI isolation, deleted-parent fallback, and compression grouping all still behave (the Set is rebuilt per render, so a removed child loses its decoration next rebuild — no stale leak). Full gate green on `890d0b05` (18/18 shards + browser-smoke + lint).

**Verdict: PASS.** This resolves the #6027 all-profiles detach class without reopening the streaming-toggle reflow the PR originally fixed. Recommend merge. Verification was read-only source review against the CI-green result; I did not execute the branch tests.

nesquena-hermes (changes requested):
Thanks @rodboev — the core fix here works: the answering "scope durable lineage by row" commit resolved the all-profiles detach regression (a fork whose parent is in a non-active profile now stays attached across streaming/settled rebuilds), and ordinary single-profile + CLI/project partitions hold. But the deep gate found two remaining scope-keying gaps that can mis-group sidebar rows, so I'm kicking it back with a precise spec rather than shipping — sidebar lineage is a crown-jewel-adjacent surface.

## Findings (both verified with reproduction)

1. **Filtered/search renders can attach a fork to a lineage row from the WRONG profile/project** (`static/sessions.js:6827`). The durable *existence* check is scope-keyed, but the attachment maps `visibleBySid`, `visibleBySegmentSid`, `visibleByLineageKey` and the lookups at ~6884–6898 still use **bare IDs**. Repro: profile A has `parent-a` + its fork; profile B / project B has `wrong-b` with `_lineage_root_id='parent-a'`; a search that hides the real parent but matches the fork + `wrong-b` → `origin/master` renders both matches top-level, this branch removes the fork and nests it under `wrong-b`. **Fix:** key every attachment identity map/cache/queue by `scopeKeyForSession(row) + session/lineage ID`, and resolve parent/segment/tip/root keys using the child's own scope (the same source/project/profile identity `durableLineageIdsByScope` already uses).

2. **Cold-render regression via lazy `_profilesCache`** (`static/sessions.js:6761`). The default/renamed-root fork grouping depends on `_profilesCache`, which `panels.js` initializes to `null` and only warms ~1200ms after `window.load` — boot doesn't await it before rendering sessions. Repro with `_profilesCache=null`: parent profile `Renamed root` + fork profile `default` render as two top-level rows; `origin/master` groups them, this branch only groups them after the cache populates. **Fix:** remove the root-alias normalization's dependency on the lazy cache — use an authoritative server-provided canonical profile-scope identity on `/api/sessions` rows/references, via one shared scope-key helper at both 6761 and 7363; add a regression exercising the `_profilesCache=null` cold path.

Both point at the same root: the scope-keying is applied to the existence check but not consistently to the attachment maps + the root-alias path. Once a single shared scope-key helper is threaded through all of them (and the cold-cache path no longer depends on `_profilesCache`), re-request and I'll re-run the full gate. Everything else verified clean (100 changed-file tests + 33 neighboring sidebar tests pass, no row loss/dup, streaming-toggle state cleared correctly).

rodboev:
Fixed both remaining scope holes.

`/api/sessions` now emits a canonical `profile_scope` from the existing server-side `_session_list_cache_profile_scope()` helper, so renamed-root and `default` lineage equivalence no longer depends on the lazy `_profilesCache` roster on cold render.

The sidebar attachment maps and fallback lookups now resolve parent, segment, tip, and root identities by `source + project + profile_scope + lineage identity` instead of bare IDs, so a filtered render can't reuse a visible row from another scope and steal the fork attachment.

I also added the two regressions you called out, one for filtered cross-scope attachment and one for `_profilesCache=null` on first render, plus the serializer and consumer assertions that pin the new row field. Thanks for pushing on the deeper class here.

nesquena-hermes (changes requested):
Great progress @rodboev — the re-push nailed both prior findings: all three attachment maps + the parent/segment/tip/root lookups are now scope-qualified, `/api/sessions` emits a canonical `profile_scope` (root aliases normalize to `default`, two named profiles sharing a project ID stay distinct), and cold render reads the server `profile_scope` before any fallback so the `_profilesCache` dependency is gone. The gate verified all of that clean.

One new CORE regression the scope-key introduced, though, so one more round:

**Project-assigned conversations lose their delegated subagent rows** (`static/sessions.js:668`). The lineage scope key uses each row's raw `project_id`, but state-db subagent projections come through with `project_id=null` (`api/models.py:6940-7010`). Repro: a `parent → child` lineage that renders on `origin/master` now produces mismatched scopes — parent `webui␀projX␀work` vs child `webui␀␀work` — so only the parent renders and the delegated child detaches. This bites any project-assigned session that has a delegated subagent child.

**Fix:** use an *effective lineage-project* scope rather than the raw per-row `project_id` — when a child's own `project_id` is null, inherit it from its uniquely-resolved same-profile parent/root. Don't just drop `project_id` from the key (that reopens the cross-project mis-attachment we just closed). Add a behavioral regression proving a project parent keeps its unassigned delegated child while same-id rows from a *different* project/profile stay isolated.

Everything else is verified clean (17 focused tests + streaming cache separation + node/diff checks; full suite 13,135). Once the effective-project inheritance lands, re-request and I'll re-run the gate. Thanks — the server-`profile_scope` approach is exactly right.

rodboev:
Thanks, this was the remaining project axis in the same sidebar-lineage scope-keying class.

I fixed it in three parts:

1. Added one effective lineage-project resolver in `static/sessions.js`. Explicit `project_id` still wins. A null-project descendant now walks only its same-source, same-profile ancestor parent/root chain, so it inherits one unambiguous project and fails closed on missing, cyclic, or conflicting candidates. Descendant siblings and forks no longer donate project scope.
2. Threaded that resolver through every project-sensitive stage in the sidebar path: partitioning, scoped reference selection, durable lineage membership, scoped attachment lookup, and the Unassigned-chip availability check. Project stays in the final key, so the round-2 cross-project and cross-profile isolation stays intact.
3. Added the project-parent plus null-project child plus project-B sibling-fork regression, updated the resolver-aware harnesses, and refreshed the #5305 no-project case so it proves the child follows the parent’s effective project before later filtering. The caller-fallout tests on cron-project visibility and the Unassigned chip are green on the final branch too.

The round-3 enumeration now includes the project-state surfaces as well as the raw project gates, state-db projection, attachment callers, identity maps, and focused isolation tests. `api/models.py` stayed unchanged; explicit project moves and persistence writers remain their existing owners.

Canonical server `profile_scope`, pinned-fork behavior, missing-parent fallback, compression collapse, source isolation, row ordering, and persistence all stayed unchanged.

nesquena-hermes (changes requested):
Very close, @rodboev — the project-child detach is **fixed** and verified: null-project delegated children inherit + group correctly, explicit projects win (a genuinely different-project child isn't force-inherited), profile/source boundaries + missing ancestors fail closed, and all-profile / single-profile / CLI partitioning holds (131 tests pass). One remaining fail-closed edge:

## SILENT (`static/sessions.js:691-693`)

**Cyclic lineage doesn't fail closed.** The `seen` set prevents an infinite loop, but on a genuine back-edge it merely *skips* the edge and keeps the project already collected. Verified: `parent(project=projA) → child → parent` resolved the child to `projA`, so it passed the project filter and attached under the parent — where a cycle should be treated as untrustworthy and fail closed.

**Fix:** replace the cycle-*skipping* traversal with **path-aware cycle detection** — track the path from the origin and, on a genuine back-edge (an identity already on the current path), return `null` (fail closed) rather than trusting the collected project. Still allow convergent (diamond) paths and benign self-root metadata — only a real cycle should void the inheritance. Add a parent↔child cycle regression asserting no project inheritance and no project-scoped attachment.

That's the last edge — the effectiveProject design is right, it just needs true cycle rejection instead of skip. Once path-local cycle detection returns `null` on a back-edge (+ the regression), re-request and I'll re-gate.

(Happy to fold this one in myself if you're heads-down elsewhere — it's a small change to the traversal — just say the word; otherwise it's yours.)

rodboev:
Good catch. Commit `5476674c` changes `static/sessions.js:687-712` so the effective-project walk tracks the active ancestry path, returns `null` on a genuine back-edge, and still treats completed revisits and benign self-root lineage metadata as valid.

The focused regression at `tests/test_5306_subagent_sidebar_flicker.py:267-300` now covers the parent<->child cycle directly: the child inherits no project, never enters the `projA` partition, and never attaches under the project parent. The same test keeps the diamond and self-root positive controls so the resolver still accepts the non-cyclic cases you called out.

The rest of the effective-project surface stays on the same shared resolver. Partitioning, durable membership, attachment identity, the Unassigned chip, and the backend-produced lineage fields all keep routing through that one seam without widening the change.

nesquena-hermes:

Certified contributor head: `12a9791e896226c5cb17148a4c81f793afda57a0`  
Synthetic integration: `8708c2d484b8596faec9b4c593e1797e7997c5d9`

Thanks @rodboev — the latest path-aware DFS fix is real. I reproduced that genuine cycles now fail closed while diamonds and benign self-roots remain valid. This exact head still cannot receive a GREEN certificate because three scope/ownership gaps remain. Each was reproduced against the production sidebar helpers in the exact-head sandbox.

### Gate evidence

- GitHub exact-head matrix: **18/18 green**.
- Focused touched sidebar suite: **132 passed**.
- Full serial sandbox: **13,692 passed**, with exactly the same inherited 18 failing/error node IDs as frozen clean master and no PR-only test node.
- `git diff --check`: clean. `node --check static/sessions.js`: clean.
- Current browser smoke on `/`, Settings, and Sessions: clean, zero console errors.
- Opus: **APPROVE** (it verified the latest cycle fix but missed the cases below).
- Fable: **SHIP-WITH-UX-FIXES**; Codex: **SHIP ONLY WITH FIXES**.
- The direct production-helper reproductions below control the RED result.

### Blocking findings

1. **`static/sessions.js:683-712,7792,9095-9227` — explicit writable-fork unassignment is silently undone by ancestry inheritance (CORE false-success).**

   Production-helper result after a Project-A fork has persisted `project_id:null`:

   ```json
   {
     "effectiveFork": "projA",
     "projectA": ["parent", "fork"],
     "unassigned": []
   }
   ```

   `/api/session/move` persists the null and the UI reports “Removed from project,” but the resolver treats null as merely missing, re-inherits `projA`, leaves the fork under Project A, and can hide the Unassigned chip. This interaction is newly exposed by the PR’s inherited-project rendering and was independently caught by Codex and Fable.

   **Fix:** distinguish projection-only delegated descendants that are allowed to inherit from writable sessions whose explicit null is authoritative. Prefer a server-provided provenance/ownership flag in the sidebar row; a narrowly proven client predicate for read-only cross-surface child rows is acceptable if it is already authoritative. Add an end-to-end production-helper regression proving:

   - a writable fork moved to null stays out of Project A and appears under Unassigned;
   - a projection-only null-project delegated child still inherits its one unambiguous parent project;
   - the picker/toast never reports success while the effective project remains unchanged.

2. **`static/sessions.js:7190-7224,7584-7612` — lineage collapse still coalesces copied IDs across profile/project scopes before the new scoped attachment maps run (CORE data-shaping loss).**

   I fed two independent conversations with copied `root`/`tip` IDs into the all-profiles render input:

   - `alpha / projA`: `root → tip`
   - `beta / projB`: `root → tip`

   Exact-head output is one row, not two:

   ```json
   [{
     "profile": "beta",
     "project": "projB",
     "count": 4,
     "segments": ["beta:tip", "beta:root", "alpha:tip", "alpha:root"]
   }]
   ```

   `_collapseSessionLineageForSidebar()` still builds `sessionIdsInList`, `sessionsById`, and `groups` with bare IDs/lineage roots. By the time `_attachChildSessionsToSidebarRows()` applies its new scoped keys, one profile’s conversation has already been erased into the other.

   **Fix:** collapse independently per `_sidebarLineageScopeKey(source, effective project, profile)` before attachment (or make every collapse lookup/group key equivalently scoped), while preserving the raw lineage key used by the row UI. Add the four-row alpha/projA + beta/projB duplicate-ID regression and assert two independent rows, each with exactly two same-scope segments.

3. **`static/sessions.js:7062-7121` — hidden archived-child suppression still leaks across profiles through a bare-ID set (SILENT disappearance).**

   `rawSessionsById`, visible maps, depth keys, and attachment keys are scoped, but `hiddenArchivedChildTree` still stores `child.session_id` and checks `parentSid` without scope. Reproduction:

   - profile alpha has an archived-parent chain whose hidden child ID is `collision`;
   - profile beta has an unrelated non-archived parent ID `collision` and a visible child `visibleB`;
   - the current head suppresses `visibleB` entirely: `[]`.

   The exact two-line scoped counterfactual restores the correct beta row:

   ```json
   [{"sid":"visibleB","orphan":true,"profile":"beta"}]
   ```

   **Fix:** store/check `_sidebarScopedIdentityKey(...)` values in `hiddenArchivedChildTree`, including the current ancestor scope during traversal. Add this alpha/beta collision regression so a hidden archived chain can suppress only descendants in its own source/project/profile scope.

### Re-gate contract

**Certificate result: RED for exact head `12a9791e8962`.** The latest cycle fix is accepted; do not redo it. Re-push the three scoped ownership/collapse fixes with production-output tests, and the new SHA will be eligible for a full exact-head re-gate. No merge, tag, deploy, approval, close, or contributor-branch write was performed by the gate-certifier.

nesquena-hermes:
## ⛔ Gate certificate — RED

**Exact contributor head:** `dea161a95ebff97b79d1c271e79f5eab83a17664`  
**Current master:** `dd7f6ac318f7c91a06841f326216542d57dd5f2c`  
**Strict-binary-identical integration:** `8efb2b12f7ea28bfad1197deacfc90da1781d5f5`  
**Gate surface:** crown-jewel sidebar lineage, project ownership, profile/project isolation, archived-child suppression, and large-list rendering

This is a rebase-only re-stamp of the previously RED feature patch. A full gate from zero reproduces the three earlier blockers unchanged and adds one independently reproduced performance blocker. The accepted cycle/diamond/self-root fix remains correct and should not be redone.

### Reproduced release blockers

1. **[CORE] Removing a writable fork from its project is a false success.** `static/sessions.js:683-712, 9242-9254` treats authoritative `project_id:null` as missing and re-inherits the ancestor's project. The production-helper probe returned `effectiveFork="projA"`, kept the fork in Project A, and removed it from Unassigned. Fix: allow inheritance only for an explicitly identified projection-only delegated child; a writable null-project row must remain null.

2. **[CORE] Copied lineage IDs collapse across profile/project scopes before scoped attachment.** `static/sessions.js:7217-7251, 7638-7639` still keys collapse maps/groups by bare lineage/session IDs. The alpha/projA plus beta/projB duplicate-ID probe returned one four-segment beta row instead of two independent two-segment rows. Fix: collapse by source + effective project + canonical profile scope + raw lineage identity, and thread the effective-project resolver through every collapse consumer.

3. **[SILENT] Archived-child suppression leaks across profiles.** `static/sessions.js:7089-7104, 7146-7148` stores/checks bare IDs in `hiddenArchivedChildTree`. An archived alpha child with a colliding ID suppressed unrelated visible beta child `visibleB`; the scope-filtered counterfactual preserved `visibleB` as an orphan. Fix: use scoped identities for every insertion, lookup, and ancestor traversal marker.

4. **[CORE] Valid nested lineages make the synchronous sidebar render effectively quadratic.** `static/sessions.js:670-713` re-walks ancestry on repeated resolver calls. Parent reproduction on the production render path measured folded candidate `50 / 858 / 3248 / 14499 ms` for `100 / 500 / 1000 / 2000` rows versus current master `2 / 3 / 5 / 10 ms`, with the same one-root/all-children output. The product explicitly supports a 2,000-row archived load. Fix: precompute or memoize scoped graph project/cycle results while preserving ambiguity rejection, true-cycle failure, diamonds, and benign self-roots.

### Gate evidence

- Threat scan: **CLEAN** at this exact head.
- Submitted touched-test slice: **132 passed**.
- Reviewer regression net: all three prior RED oracles fail on the candidate exactly as specified; a separate scope-key mutation turned two focused tests RED, and restoration returned the same 71-node slice GREEN.
- Full serial suite, topology-matched standalone clones: folded candidate **13,839 passed, 83 skipped, 1 xfailed, 2 xpassed, 5 failed, 2 errors, 34 subtests passed**; current master **13,832 passed** with the **identical seven failing/error node IDs**. This establishes no additional suite-failure delta, but the full suite is not green and does not override the reproduced product blockers.
- Real sandboxed Playwright smoke: `/`, `/#settings`, and `/#sessions` all rendered with no page errors; expected no-egress CDN noise was filtered by exact string only.
- Codex: **SHIP ONLY WITH FIXES**, reproduced all four blockers.
- Attested real Opus 4.8 (`--model claude-opus-4-8`): **RED**, independently verified the three ownership/scope blockers and accepted the cycle fix.
- Fable 5: **RED**, verified the payload is content-identical to the prior RED head and none of the three requested fixes landed.

**Result:** `gate-fail`. Re-push substantive fixes plus production-output regressions for all four items, then run a full exact-head re-gate. This certifier did not merge, tag, deploy, close, or modify the contributor branch.

rodboev:
Thanks for the exact-head gate. I rebuilt the lineage seam and pushed be21a710.

- Writable rows with project_id null, including forks, stay Unassigned. Only read-only delegated projections inherit a uniquely resolved ancestor project.
- Collapse, attachment, archived-child suppression, and durable membership now use source/profile/project-scoped identities before rows are merged.
- Hidden archived references use the same scope on the API side, and lineage-report expansion selects the requested profile state database.
- The resolver precomputes scoped graph ownership and rejects ambiguous or cyclic ancestry.

The focused API and sidebar suite passes 106 tests, the 2,000-row graph bounds pass, runtime JavaScript checks pass, and the isolated served browser proof passes with no page errors. The before/after screenshots are published on the fork’s screenshots branch.

Please re-gate the exact head be21a7109eda580b12ea83763df0fed8f12ab7aa.

rodboev:
Thanks for the exact-head gate. I reworked the shared identity seam and pushed `30d85f74`.

- Profile-less legacy rows stay on the root database even when a named profile is active.
- Viewed counts, completion markers, streaming state, optimistic merging, and deletion now use the same source/profile/project-scoped identity as rendering.
- The production row-tree regression covers a delegated child and an Unassigned fork across narrow and restored references.

The affected suite passes 246 tests with one local Playwright skip because Playwright is not installed; runtime lint, JavaScript syntax checks, and Python compilation pass. Please re-gate exact head `30d85f7442058bbb38c55c2332b0e60b29b739b6`.

rodboev:
The follow-up seam review found two valid P1s in delegated identity resolution. I fixed both and pushed final head `64f63bf070fa7f1bb360068666d355809cdf6dec`: duplicate delegated IDs now retain parent-reference context, and runtime identity builds against the full sidebar and archived-reference context. The focused suite passes 317 tests with one local Playwright skip. A fresh same-provider review now reports no valid P1s. Please re-gate exact head `64f63bf070fa7f1bb360068666d355809cdf6dec`.

rodboev:
The final CI follow-up exposed one more compressed-lineage case for an Unassigned fork. I fixed segment lookup to use the resolved parent scope and pushed final head `853ffb6f33604d89d75f23c99802b88762ec41c9`.

rodboev:
The diagnostic-only assertion update is pushed at final head `8f69adfb3d1167c83405d4b5a158de5bd05c5546`. It does not change production code; it includes both rendered trees if the served fixture still fails in CI.

rodboev:
The second diagnostic assertion is pushed at final head `eec377e217b279357f2e45e7a3b796cbd84ec1a7`. It includes the full narrow and restored render result when the CI fixture fails, so the remaining production mismatch is directly observable.

rodboev:
Published reviewed head 55e18d240a2b34804cf3fe62405d1cca3e620052. The served-fixture evidence showed the stable seam is a nested delegated child plus a top-level Unassigned fork, so this update keeps null-project forks outside project-owned child grouping. The focused suite passes 317 tests with one skipped locally, and the fresh same-provider P1 review is clean. All current-head CI checks are green and there are no unresolved review threads. Please re-gate this exact head.

nesquena-hermes (changes requested):
Thanks @rodboev — the project-lineage/sidebar direction is right and the full suite is green, but an authoritative regression gate found two issues that block this reliability fix (one is the exact profile-isolation boundary this PR sets out to harden, the other is a large-sidebar freeze the change introduces). Both need a pass before it can ship.

### Must fix

**1. (SILENT — profile-isolation leak) `api/models.py:5635` — `_sidebar_state_db_path` still resolves a foreign profile's DB in isolated mode.** Reproduced: isolated mode active, profile A with an existing `A/state.db`, request profile B → `_sidebar_state_db_path("b")` returns `A/state.db`, because `get_hermes_home_for_profile()` clamps the foreign name to the pinned home BEFORE `fallback_to_active=False` can take effect. That leaks profile A's session metadata into profile B's sidebar. The same flaw reaches the lineage-report endpoint at `api/routes.py:13305`. **Fix:** before `_get_profile_home()`, return `None` when `fallback_to_active=False` AND isolated mode is active AND the requested profile != the pinned profile. Preserve that `None` through both the sidebar-override enrichment and the lineage-report endpoint. Add an isolated-mode regression proving B cannot resolve A when B's DB is absent.

**2. (CORE — large-sidebar freeze) `static/sessions.js:942` — `_sidebarRuntimeIdentityKey()` rebuilds the full lineage index synchronously for every row.** A production-shaped 2,400-row sidebar took **7.45s** for one per-row key pass vs **2.5ms** with a single shared index — and the payload path makes multiple such passes, so a large sidebar freezes on refresh. **Fix:** build ONE lineage index per list application/merge and thread it (plus precomputed runtime keys) through the polling, optimistic-merge (`_mergeOptimisticFirstTurnSessions`), source-tracking, unread-transition (`_markPollingCompletionUnreadTransitions`), and observed-streaming helpers, instead of reconstructing per row (loops at ~1573 and ~5621).

### Not blockers
Full 5-shard suite is green; the only failures I saw (`test_issue2513_custom_provider_remote_models`, `test_ttl_cache::test_mtime_invalidation`) are pre-existing environment flakes on our CI box, identical on clean master.

Once both land I'll re-gate the exact head (Codex + full suite) and take it to release. This is a genuinely valuable sidebar-reliability fix — worth getting the isolation + perf right. Attribution preserved.

rodboev:
Thanks for the detailed gate. I addressed both blockers.

1. Explicit foreign profile requests in isolated mode now fail closed before resolving a profile home when active fallback is disabled. Profileless rows retain the pinned root alias. Sidebar metadata reads skip the unavailable database, and a lineage report for that profile returns 404 without reading another profile's database.

2. Sidebar refreshes now build one scoped lineage index and precomputed identity context per list application or merge. Optimistic rows, polling completion and unread state, source and snapshot state, observed streaming state, deletion cleanup, visits, message completion, and lineage-cache pruning reuse that context. The existing `cron_running` completion suppression remains intact.

3. Regression coverage now includes foreign-profile isolation, the profileless root alias, duplicate scoped IDs across sidebar and message lifecycle state, optimistic first-turn rows, production list application, batch deletion, and lineage-cache pruning.

nesquena-hermes (changes requested):
## Watch re-gate: changes requested at `ee289372b924`

Thanks for the substantial rework. The isolated foreign-profile database blocker is fixed: `_sidebar_state_db_path()` now rejects an explicit foreign profile before home resolution when fallback is disabled, and the lineage-report path preserves that fail-closed result.

One part of the sidebar-runtime performance and ownership blocker remains:

1. `static/sessions.js::_clearCronSessionCompletionUnreadForInactiveProfiles()` loops over every persisted unread marker and calls `_resolveCronCompletionMarkerOrigin(sid, marker)` without a runtime context. The resolver then builds `_createSidebarRuntimeContext(_allSessions, _sidebarReferenceSessions)` for each marker. This leaves an O(markers × sidebar rows) profile-switch cleanup, the same repeated full-index-build class the prior gate asked to remove.
2. `static/panels.js` calls `_markSessionCompletionUnreadIfBackground()` with only a raw session ID and active-profile metadata. For duplicate scoped session IDs, `_sidebarRuntimeRowForSid()` cannot select the owning row and the state key can fall back to the legacy raw SID. A later lookup can then migrate that marker into the wrong profile/project scope.

### Fix spec

- Build one `_createSidebarRuntimeContext(_allSessions, _sidebarReferenceSessions)` before the inactive-profile marker loop and pass it to every `_resolveCronCompletionMarkerOrigin(..., runtimeContext)` call.
- For cron completions, pass the uniquely matched scoped row plus the shared runtime context, or perform an equivalent profile-qualified unique lookup inside `_markSessionCompletionUnreadIfBackground()`. If ownership is ambiguous, fail closed instead of writing or migrating a raw-SID marker.
- Add a production-source regression with 2,400 rows and multiple cron unread markers that instruments `_buildSidebarLineageIndex` and requires exactly one build for cleanup. Add a duplicate-SID/two-profile cron-completion regression proving only the intended scoped unread key changes.

This was a static-only re-gate. The mandatory threat scan returned `SUSPICIOUS` because the test harness contains three `eval(...)` calls, so policy prohibited executing the branch. A corrected head will still need the sandboxed exact-head gate. No merge, deployment, or contributor-branch write was performed.

Issue #1361, opened 2026-04-30
bug(streaming): Stop/Cancel discards AI's already-streamed reasoning + tool output (and possibly user prompt) — paid tokens lost


nesquena-hermes:
## Bug report

Hitting **Stop** during a streaming response loses both the AI's already-streamed output AND the user's just-typed prompt. Reported by AvidFuturist (Nathan), April 30 2026:

> More than once, I've typed a long prompt into a new chat, submitted it, but then needed to stop the AI's response before it completed for whatever reason. When I do that, **I lose everything the AI already printed, and (even worse), the whole prompt I just sent gets erased.** The first thing has been happening from the start, I think, but I feel like the second thing is more recent.
>
> This is extremely frustrating, as 1) I just needed to make a few edits to the prompt and submit it again, but now it's gone (noooooo!), and 2) I honestly just never ever want output I've already paid tokens for to simply disappear. It's almost always at least useful context, at least for my workflows.

## Relationship to #1298 / PR #1338

PR #1338 (merged today, v0.50.246) fixed exactly the user-prompt-loss-on-cancel case for a *new chat*: when `cancel_stream()` raced the streaming thread and won, `pending_user_message` was cleared without merging the user turn into `s.messages`. The fix synthesizes the user turn from `pending_user_message` before clearing.

But Nathan reports both losses still happen on the macOS app (latest v1.4.0 + WebUI v0.50.248). Two possibilities:

1. The Mac app is on a cached build older than v0.50.246 (worth confirming). If so, the cancel-pyramid is actually fixed for him too once he updates — leaving only the AI-output loss as a real outstanding bug.
2. The fix in #1338 doesn't cover all paths. It only synthesizes from `pending_user_message`. If the user message is already in `s.messages` but the AI's partial output isn't preserved, refreshing would still show the user message but not the AI text. That matches "I lose everything the AI already printed" but not the prompt-erasure half of the report.

The right next step is to **reproduce both halves on v0.50.248** (after Nathan's Mac app picks up the new release) and decide whether the prompt-erasure half is fully resolved or still leaks through a different code path.

## Why the AI output goes missing — verified in master (v0.50.248)

The cancel path *does* try to preserve partial AI output (`api/streaming.py:2682–2712`, the #893 fix), but only the **token-level text stream** is captured into `STREAM_PARTIAL_TEXT` (`api/streaming.py:1565–1570`). Three failure modes I can see:

### A. Reasoning-only output is dropped on cancel

[...]

PR #6047, opened 2026-07-14
fix: cancel rollback — discard interrupted turns cleanly + local Ollama docs

ontop01970:
## Summary

Two changes in this PR:

### 1. Clean cancel rollback (api/streaming.py)

**Problem:** When a user clicks Stop or interrupts a task, the WebUI previously persisted three artefacts the user does not want:
- A user message bubble for the interrupted prompt
- A partial `_partial`-flagged assistant response (whatever streamed before the cancel)
- A `**Task cancelled:**` error marker

**Fix:** Replace `_persist_cancelled_turn()` and the cancel path inside `cancel_stream()` with a rollback strategy. The session reverts to its exact state before the user sent the message — no bubble, no partial, no banner.

Key implementation details:
- New `_rollback_cancelled_turn_messages(session)` helper strips the current-turn user message from both `session.messages` and `session.context_messages`. Works for both deferred (default) and eager session-save modes.
- Uses a 1-second timestamp tolerance (`_m_ts >= _pending_started - 1`) because `pending_started_at` is stored as a float (e.g. `1789553400.851`) while message timestamps are `int(time.time()) = 1789553400`, causing a strict `>=` to silently miss the match.
- The terminal cancel SSE event still carries the rolled-back session payload so the browser syncs immediately via `_applyCancelSessionPayload` without a second fetch.
- Ephemeral (`/btw`) sessions use the existing `_cleanup_ephemeral_cancelled_turn` path and are unaffected.

### 2. Local Ollama setup documentation (README.md)

Adds a new `## Local Ollama setup` section to the README covering:
- The 64,000-token minimum context window requirement for tool use, and why dropping below it breaks the agent loop
- How to install Playwright + Chromium for the browser tool
- Modelfile parameters that reduce verbosity (`repeat_penalty`, `presence_penalty`)
- SOUL.md directive pattern for session-level brevity
- The four `settings.json` keys that enable full tool-call + reasoning display
- A note on the new interrupted-task rollback behaviour

Also adds the section to the Contents table.

## Testing

Unit tests (direct function calls with mock sessions):
- Deferred save mode: rollback is a no-op when user message was not yet in `messages`
- Eager save mode: user message removed from both `messages` and `context_messages`  
- Workspace-prefixed content matching (partial content match)
- `_persist_cancelled_turn` clears all pending fields, appends nothing

Live API test:
- Created fresh session → started a long-running turn → cancelled mid-stream → confirmed 0 messages, no active_stream_id, no pending_user_message, no cancel banner

---
*This PR was developed with [Warp](https://app.warp.dev/conversation/bbab6f58-7dad-40d4-8344-df9b4436749a)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ This gives you nearly **1:1 parity with Hermes CLI from a convenient web UI** wh
 - [Quick start](#quick-start) — clone + `bootstrap.py` / `start.sh` / `ctl.sh`
 - [Features](#features) — chat, sessions, workspace, voice, profiles, security, themes, panels, mobile
 - [Configuration & access](#configuration--access) — auto-discovery, overrides, remote/Tailscale/phone, manual launch
+- [Local Ollama setup](#local-ollama-setup) — context window, browser tool, display settings
 - [Docker](#docker) — single- and multi-container deploys
 - [Running tests](#running-tests)
 - [Architecture](#architecture) — backend/frontend layout, state dir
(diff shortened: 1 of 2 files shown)


franksong2702:
Thanks for tackling cancellation cleanup. I want to flag a product-contract issue before this moves forward.

The proposed rollback currently removes the submitted user turn, partial assistant output, and the cancellation marker. That conflicts with the Live-to-Final contract in `docs/rfcs/live-to-final-assistant-replies.md`: once live text, reasoning, or tool activity has become visible, cancelling the turn must not silently erase that work, and the turn should settle as `cancelled` rather than disappear.

Could this distinguish the two cases explicitly?

- Cancel before any visible assistant activity: a clean rollback may be appropriate.
- Cancel after visible text, reasoning, or tools: preserve the visible partial work and settle with an explicit `cancelled` terminal outcome.

Please add regression coverage for both paths so cancellation cleanup cannot reintroduce partial-work loss.

This PR also combines the cancellation behavior change with a substantial Ollama documentation addition. Splitting those into separate PRs would make the cancellation contract change much easier to review independently.

nesquena-hermes (changes requested):
## 🔬 Deep gate (Codex, crown-jewel streaming + data-integrity) — KICK BACK TO AUTHOR

Thanks @ontop01970 — the intent (a clean Stop that doesn't leave a half-turn behind) is reasonable, but as written the rollback is **not safe on the conversation-history path**: a real-reconciler probe reproduced it **removing a PRIOR completed turn and leaving the session inconsistent** (the persisted current user + partial assistant rows get appended back on the next send). On the crown-jewel chat-history surface that's data loss, so this needs a rework before it can land:

1. **[CORE — data loss: strips the wrong/prior turn] `api/streaming.py:1498-1553`** — the rollback identifies the cancelled turn by **content/timestamp guessing**. With a repeated prompt, a substring match, or when the last message isn't the cancelled turn's, it can strip a **prior completed** user/assistant pair. **Fix:** own the rollback with a durable `stream_id`-keyed pre-turn record capturing display history, context history, title, watermark/boundary, and compression ownership — roll back to that snapshot, don't guess.

2. **[CORE — resurrection race with state.db] `api/streaming.py:1556` + `:10855`** — rows are deactivated without coordinating with the agent's `state.db`; the persisted current user + partial assistant rows come **back** on next send (probe-confirmed via the reconciler at 8492-8504). **Fix:** capture `state.db`'s pre-turn head and deactivate current-turn rows only **after** the worker has settled.

3. **[CORE — non-atomic / lost identity on late cancel] `api/streaming.py:9350-9632`** — a late cancellation can lose the pending-rollback identity, and the rollback isn't atomic (can leave some rows stripped, others not — worse than the pre-PR state). **Fix:** make the rollback identity survive every late-cancel check and commit atomically.

4. **[test — asserts the OLD behavior] ** no tests changed, and 4 existing tests still assert the previous behavior (cancel markers + retained partials): `tests/test_cancelled_turn_status.py:76`, `tests/test_issue1298_cancel_and_activity.py:110`, `tests/test_issue1361_cancel_data_loss.py:119`, `tests/test_cancel_interrupt.py:168`. **Fix:** update those to the new contract and add focused tests for: three prior turns + cancel the 4th (assert 1-3 fully intact), repeated/substring prompts, late cancellation, atomic-failure, eager vs deferred checkpoint, and a real `state.db` next-turn replay.

One design question worth settling too: discarding the interrupted turn **entirely** (user prompt + partial output both vanish) is a UX change — some users want to see what they typed / the partial after a Stop. Worth confirming that's the desired product behavior vs. keeping the user bubble and only dropping the partial. Note #1361 in particular was a *cancel data-loss* fix, so this area has bitten before — the durable-snapshot approach is the way to make it safe. Rebased clean on current master (README TOC conflict resolved).

PR #6055, opened 2026-07-14
feat(#5932): add per-board default workdir

### LINKED ISSUE #5932: Add workdir selection and file read/edit/download
These would be very helpful for me on-the-go, should I create a branch or would you do the PR?

rodboev:
## Thinking Path

- The broad issue aged, but the current maintainer-owned slice is narrower: task-level workspace fields and workspace file operations already exist, while per-board default workdir persistence does not.
- Hermes Agent already persists `board.default_workdir` and already owns the safety-critical inheritance rules for scratch, `dir`, and `worktree` tasks. The WebUI is missing only the board settings surface, trusted-path validation, payload plumbing, and now the follow-up guard that keeps rename-only edits from resubmitting an unchanged saved path.
- The fix stays board-scoped: the modal still owns board settings, the bridge still owns persistence and trust checks, the Agent still owns inheritance, and the rework makes unchanged board edits stop resending `default_workdir`, matches the task workspace path's 255-character limit, keeps the narrow modal scrollable inside the viewport, and isolates the rendered layout proof per locale/viewport cell.

## What Changed

- `static/index.html`, `static/style.css`, and `static/panels.js`: add a board default-workdir field to the existing board modal, preload it from board metadata, reuse saved workspaces as a datalist, keep current-board settings reachable even when the active board is `default`, cap the board workspace path at 255 characters, keep the mobile modal within a scrollable viewport-bounded surface, and on board edits omit `default_workdir` from the PATCH unless the user changed it.
- `static/index.html`: cap the board default workspace path at 255 characters, matching the task workspace path control.
- `api/kanban_bridge.py`: validate non-empty board defaults through `resolve_trusted_workspace`, preserve omitted-field updates, and treat an explicit empty string as clearing `default_workdir`.
- `static/i18n.js`: add the board-settings title key and shift the visible copy from rename/workdir wording to settings/workspace wording across the locale blocks this PR ships.
- `tests/test_issue5932_kanban_board_default_workdir.py` and `tests/test_issue5932_kanban_board_default_workdir_layout.py`: cover bridge semantics, modal payloads, unchanged-versus-changed board edits, default-board reachability, maxlength parity, locale coverage, and isolated rendered layout cells, including the narrow modal scroll path.

## Why It Matters

Kanban boards can carry the persistent workspace anchor that Hermes Agent already knows how to inherit safely, and board rename-only edits no longer fail just because the saved path went stale or the paired agent predates `default_workdir`. That closes the board-level gap without weakening scratch cleanup, explicit task overrides, or the existing task create contract.

## Screenshots

Before, the board modal has no board default workspace path field:
Before board settings modal

After, the board modal exposes the localized default workspace path field:
After board settings modal

## Verification

Focused board-modal and bridge tests cover create, update, clear, omitted-field preservation, unchanged-versus-changed edit payloads, untrusted-path rejection, default-board settings reachability, and the 255-character input parity; the kanban parity and static locale tests keep task workspace ownership and locale coverage pinned; the targeted browser test covers the rendered board-modal layout and default-board settings path with isolated locale/viewport cells, explicitly scrolling the narrow modal field before geometry checks; CI runs the full matrix on 3.11, 3.12, and 3.13.

## Risks / Follow-ups

The first post-push matrix exposed a 480x320 modal overflow, so the mobile modal now has a viewport-bounded scroll surface and the rendered proof scrolls the field before taking its geometry snapshot. The maintainer-reported locale-cell flake did not reproduce in the local pre-fix sweep; CI remains the rendered-browser proof for the final head.

## Upstream

Closes #5932.

## Model Used

GPT 5.5 via Codex CLI


rodboev:
Pushed a follow-up on the current head.

The board modal now refreshes the saved workspace list on open, so the default-workdir datalist does not stick on a stale cached list. The `default_workdir` keyword contract is already present on the Hermes Agent board metadata API (`create_board` / `write_board_metadata`).

nesquena-hermes:
## Summary

Read the diff at HEAD (`2f7be2d65`) against `origin/master` — `api/kanban_bridge.py`, `static/panels.js`, `static/index.html`, both new test files — plus the agent-side board-metadata contract. The layering is right: WebUI owns the board-settings surface, trusted-path validation, and payload plumbing; all task inheritance stays in the Agent. Green across all 18 checks.

## Cross-repo contract check

The bridge relies on the Agent's `write_board_metadata(default_workdir=...)` keyword. That exists and behaves exactly as the bridge assumes. In `hermes_cli/kanban_db.py:699-700`:

```python
if default_workdir is not None:
    meta["default_workdir"] = str(default_workdir) if default_workdir else None
```

So passing `""` **clears** the field (falsy → `None`), and omitting the key entirely preserves the existing value — which is precisely what `_update_board_payload` encodes: it only builds `board_kwargs` when `"default_workdir" in body`, mapping empty → `""` and non-empty → `resolve_trusted_workspace(...)`. Task-side inheritance (the safety-critical part) is already handled agent-side at `kanban_db.py:2557-2573` (scratch resolution) and `5664-5695` (worktree/dir anchoring with a hard git-repo containment guard), so keeping WebUI out of that logic is the correct call.

## Validation + error path is sound

`resolve_trusted_workspace` raises `ValueError` for untrusted paths (`api/workspace.py:802`, home-dir or saved-workspace allowlist, system-root denylist). The create-path call sits outside the inner `try` in `_create_board_payload`, but that's fine — `handle_kanban_post` wraps the whole call and maps `ValueError` → 400 (`kanban_bridge.py:1270-1271`), same for PATCH (`1310-1311`). `test_board_payload_validates_default_workdir_and_preserves_omission` pins the reject path, the clear path, and the omission-preserves path.

## Re-enabling default-board rename is safe

The switcher change (`panels.js`: `switcher.hidden = false`, dropping `renameDisabled`) exposes rename for the `default` board so its settings are reachable. That doesn't risk a slug mutation: `openKanbanRenameBoard` disables/hides the slug input and the PATCH URL is built from the fixed `kanbanBoardModalSlug` value, not user input:

```javascript
const slug = document.getElementById('kanbanBoardModalSlug').value;
await api('/api/kanban/boards/' + encodeURIComponent(slug), {
  method: 'PATCH',
  body: JSON.stringify({name, description, icon, color, default_workdir: defaultWorkdir}),
});
```

`archiveDisabled = current === 'default'` is retained, so `default` still can't be archived. Good.

## One non-blocking note on the clear round-trip

`test_board_payload_validates_default_workdir_and_preserves_omission` asserts the stub stores `default_workdir == ""` after a clear. That's a fake-boards-stub artifact — the *real* `write_board_metadata` normalizes `""` → `None` (line 700 above), so a cleared board reads back with `default_workdir` absent/`None`, not `""`. The front-end already coalesces defensively (`meta.default_workdir || ''` in `openKanbanRenameBoard`), so the modal round-trips correctly either way. Not a bug — just flagging that the stub's post-clear value diverges from the agent's actual normalization, in case a future reader trusts the test as the contract.

## Verdict

Correct layer, contract-accurate, error paths covered, and the default-board rename change is safe. LGTM on read-through.

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (board rename can 500 via always-resent workdir; + 2 UX string fixes)

**Certified head:** `2f7be2d65cfbc59c5b92d74b1671523c65029e05` (clean-rebased onto `origin/master`)
**Verdict:** The backend guard is correct (workdir is authorized through `resolve_trusted_workspace` at store time; traversal/symlink/system-roots rejected — Codex verified). But the edit modal always re-sends `default_workdir`, so renaming a board (an unrelated action) can 500 when the stored workdir is stale/untrusted or the agent is older. Plus two UX string fixes. Ship only with fixes.

### What I ran
- **Codex** (reproduce): `SHIP ONLY WITH FIXES` — 1 CORE reproduced by direct probes.
- **Senior UX advisor (Fable)**: `SHIP-WITH-UX-FIXES` — mislabeled entry point + terminology; cross-device sound; no crown-jewel surface.
- **My own source verification**: confirmed the always-resend mechanism (panels.js:4175).
- **Full suite**: `13361 passed, 29 failed` — 14 baseline env-only + 15 from the PR's own `test_..._layout.py` (all playwright browser-viewport tests failing on the missing Chromium binary, NOT logic). **GitHub CI all-green** (playwright installed there). Non-browser test file 5/5, 50 Codex-focused board/workspace tests pass. Ruff/ESLint/scope/py_compile CLEAN.

### Reproduced blocker (RED)
**[CORE — rename 500] `static/panels.js:4175`.** The board-edit PATCH always sends `default_workdir: defaultWorkdir` (pre-populated from the stored value at `panels.js:4098`). So editing a board's name/icon/color re-runs `resolve_trusted_workspace` on the *existing* workdir at `kanban_bridge.py:907-920`. Codex verified by direct probes: omitting the field succeeds; re-sending an existing `/srv/project` that's no longer trusted raises `ValueError` → 500; and against an older pre-`default_workdir` Agent `update_board` signature, re-sending the field raises an uncaught `TypeError` → **500 on every rename**. So a user renaming a board (nothing to do with the workdir) gets a failure when the board has a stale/external workdir or runs an older agent. **Fix:** at `panels.js:4098` record the normalized original workdir; build the PATCH payload WITHOUT `default_workdir` unless it actually changed (still send `""` for an explicit clear).

### UX fixes (Fable, string-level)
1. **Mislabeled entry point.** The settings action still reads "Rename current board…" (`panels.js:3953-3955`) and the edit modal title is rename-flavored (`panels.js:4089`), but the modal now also edits the workdir — on the default board this is the ONLY door to the new feature, so it must say "Edit board…" / "Board settings…".
2. **Terminology.** The field is labeled "workdir" but the app's established vocabulary is "workspace" (and the placeholder already says "Workspace path"); locales even translated it inconsistently ("default workspace" / "default directory"). Relabel "Default workspace path".

### Verified clean (Codex)
Traversal / symlink escape / system-roots rejected at store time; the Project-OS consumer (`routes.py:11186`) is read-only (status display, exists()-gated); all 15 locales contain both new keys. Boards are intentionally cross-profile-shared in the Agent (my brief's per-profile assumption was wrong — no isolation defect).

**Do not merge.** `gate-fail` + `ux` at `sha:2f7be2d65cfb`. One CORE fix (don't re-send an unchanged `default_workdir` on rename — prevents the 500 regression) + two UX string fixes (relabel the entry point + "workspace" terminology). All small and well-specified; safe for us to apply on-branch per auto-fix-mechanical, then re-gate. After fixes it's a Settings-modal visible change → the useful captures for Nathan are the board modal (edit + create mode) and the always-visible board switcher; no crown-jewel surface. Preserve @rodboev author credit (maintainer already left an LGTM read-through).

@rodboev — the backend workspace-trust guard is solid. The one real bug: the edit modal always re-sends `default_workdir`, so renaming a board re-validates the existing path and 500s if it's since become untrusted (or on an older agent). Record the original workdir and only include the field in the PATCH when it changed. Plus rename the "Rename current board…" entry to "Edit board…" (it now edits the workdir too) and use "workspace" not "workdir" for the label.

rodboev:
Thanks, I pushed a follow-up on the current head.

1. The board-edit path now records the saved default workspace path when the modal opens and leaves `default_workdir` out of the PATCH unless the user actually changed it. Rename, icon, and color edits no longer revalidate a stale saved path, and older agents no longer see an unexpected `default_workdir` keyword on unchanged edits.
2. The current-board action and modal title now read as board settings rather than rename-only UI, which matches the fact that this is the default board's only settings door.
3. The field label now says `Default workspace path`, and the locale keys were updated so that wording stays consistent across the shipped translations.

nesquena-hermes:
## 🔬 Gate re-certification round 2 — GREEN ✅ (both round-1 blockers fixed; one SHOULD-FIX on the PR's own flaky layout test)

**Certified head:** `0fd2a3ec726d06fdeff3464a1b0459b616a441c0` (@rodboev, T1)
**Gated on a clean rebase:** head was 173 commits behind master; I applied its merge-base three-dot diff onto current `origin/master@deac1384` as `gate-rebase/6055-0fd2-r2`. Feature patch-id **byte-identical** before/after rebase (`195efef79fd66fa482a0b38c6433c11bf0093078`), zero conflicts.
**Verdict:** Both blockers from my round-1 RED are fixed and verified; Codex `SAFE TO SHIP`; Fable `SHIP-UX`; full suite clean. One non-blocking SHOULD-FIX: the PR's own Playwright layout test is flaky by construction (cross-cell state bleed on non-`en` locales) — the FEATURE is correct, the TEST isn't isolated. GREEN.

### ✅ Blocker 1 (CORE 500-on-rename regression) — FIXED & verified
Round-1 RED: the rename PATCH always resent `default_workdir` (prepopulated from stored meta), so a name/icon-only edit re-ran `resolve_trusted_workspace` on the existing stored workdir → ValueError→500 if now stale/untrusted, and an old `update_board` signature raised uncaught TypeError→500 on every rename. The fix records `originalDefaultWorkdir` (hidden field) and builds the PATCH body as `{name,description,icon,color}`, adding `default_workdir` **only** when `defaultWorkdir !== originalDefaultWorkdir`.

**My own Node harness on the shipped logic (`/tmp/gate_probe_6055_payload.cjs`) — 5/5 PASS:**
- name-only edit, stale workdir unchanged → **key omitted** (no re-validation, no 500) ✅
- workdir genuinely changed → key present with new value ✅
- explicit clear (stored non-empty → field emptied) → `default_workdir:''` sent (clear still works) ✅
- create flow both empty → omitted ✅
- trimmed-equal → omitted ✅

**Backend confirms it (`api/kanban_bridge.py:907-909`):** `default_workdir` is validated/written only under `if "default_workdir" in body`, and `write_board_metadata` uses `**board_kwargs`, so an omitted key leaves the stored value untouched — a name-only edit never re-validates the stored workdir. A CHANGED workdir still goes through `resolve_trusted_workspace` (traversal / symlink / system-root rejection intact). Codex independently verified rename-only omits (incl. whitespace-equivalent), changes/clears sent, stale not revalidated, legacy signature compatible.

### ✅ Blocker 2 (UX vocab / label) — FIXED (Fable SHIP-UX)
Menu entry + modal title relabeled "Rename current board…" → **"Board settings"** (`kanban_board_settings`); the field label is now **"Default workspace path"** (`kanban_board_default_workdir`), consistent with the task modal's "Workspace path". Fable checked every locale block: all ~15 got both new keys with real translations (no English-literal fallback, no raw-key render). Only the internal API key stays `default_workdir` (invisible, correct). Fable verdict: **SHIP-UX** — placement sound, cross-device sound (modal `width:min(460px,100%)`, narrow-viewport stacking), no crown-jewel surface, no merge-blocking screenshots required. Two trivial nits (non-blocking): the new input lacks the `maxlength="255"` its task-modal sibling has; the menu button reuses the `kanban_rename_board` key *name* (internal only — visible strings agree).

### ⚠️ SHOULD-FIX (non-blocking): the shipped Playwright layout test is cross-cell flaky
`tests/test_issue5932_kanban_board_default_workdir_layout.py` (15 cells = 5 viewports × en/ru/de) fails **non-deterministically on the ru/de cells** when run as a full parametrized sweep, but **passes every cell in isolation**. I proved this is a TEST-isolation problem, not a feature defect:
- Each failing cell passes standalone (`...[1024-600-de]` alone → PASS).
- The failing SET shifts run-to-run (run 1: 768-ru/400-ru/400-de/1024-ru/1024-de/480-ru; run 2: 1280-ru/1280-de/768-de/400-ru/1024-de/480-ru/480-de) — the signature of cross-execution state bleed, always the non-`en` locales (English is the load default).
- Root behavior (`/tmp/gate_probe_6055_modalwide.py`, 6 fresh opens): the PR's new `kanban_board_default_workdir` label renders correct German ("Standard-Workspace-Pfad") every time; what lags is the whole-modal i18n bootstrap timing (the pre-existing modal *title* stays "New board…" on a cold first render too) — an INHERITED i18n-apply-timing property, not introduced by this PR. Codex's own post-boot browser check ran all 15 locale/viewport combos with **zero page errors**.
- Fix-spec (SHOULD, for the author): isolate the layout test per cell — reset locale + close the modal between cells (or `page.reload()` per parametrize), and after `openKanbanCreateBoard()` re-run `applyLocaleToDOM()` (or wait on the label) so the assert doesn't race the i18n bootstrap. This is a flaky-test fix, no product change.

### What I ran
- **My probes:** PATCH-body contract (5/5), backend omit-vs-validate read, modal-wide i18n timing (6 opens), isolated-vs-sweep test runs — all above.
- **Codex (reproduce leg):** `SAFE TO SHIP` — no regression; verified fix on all paths, security guards intact, 15 locales complete, 15-combo browser check zero errors.
- **Fable (UX leg):** `SHIP-UX` — vocab fixed, placement/cross-device sound, no crown-jewel surface.
- **Focused:** `test_issue5932_kanban_board_default_workdir.py` 5/5 pass; ruff-forward + `node --check` (panels.js, i18n.js) clean.
- **Full serial suite:** `13455 passed / 7 failed` = the 6 established clean-master no-network Nous failures + the 1 flaky layout cell analyzed above (not a real defect).
- **Live drive** on my seeded gate server (8783) at 1024×600; delivery screenshot sent to Telegram.

### Disposition
**GREEN** for the release agent — both prior blockers genuinely resolved, feature verified live. `gate-pass` (removing `gate-fail`); keep the SHOULD-FIX visible for the author (flaky test isolation). This is a **visible / `ux`-labeled** surface (Kanban board-settings modal, and the board-switcher chip is now always visible), so per standing policy it **PARKS for @nesquena's merge decision** — not shipped unattended. Preserve @rodboev author credit; the `maxlength` + test-isolation nits fold in on merge.

entin-hun:
When is this gonna merged into main? @nesquena-hermes 
Workdir of existing boards are also going to be editable (I hope so), or only at creation?

rodboev:
Thanks, I'll fold in both follow-ups on the current head.

1. `tests/test_issue5932_kanban_board_default_workdir_layout.py` will isolate each locale/viewport case. Each cell will wait for boot to finish, set the app's locale, open the board modal, apply the locale after the modal DOM exists, wait for the expected localized label, and clean up the modal/browser even if an assertion fails. That keeps the ru/de cells from inheriting stale modal or locale state while leaving product code unchanged.
2. `static/index.html` will add `maxlength="255"` to the new board default workspace path input, matching the task modal's workspace path input.

The existing PATCH omission fix, backend trust validation, visible settings labels, and task workspace behavior stay unchanged.

nesquena-hermes:

**Certified contributor head:** `0f15416cb9a704f90bfc62974ae7e25582f2dbe7` (@rodboev)
**Current-master integration tested:** clean local rebase `51253b8c248b0cc5e47d543e95caaa58a77480bb` on frozen `origin/master@d42dae81a8e5e4e040b7471b29bf4a294947665a`.
**Rebase integrity:** zero conflicts; contributor and rebased cumulative patches have the same stable patch ID (`4244ab7582f073c05149736d8269f481f0ed65c5`); the seven-file feature scope is unchanged.

**Verdict:** engineering GREEN. The prior rename-only regression remains fixed, the real installed-Agent contract accepts the new metadata field, trusted-workspace validation is intact, and the current rebased head has no test or live-browser regression. This is still a visible `ux` feature, so it remains parked for the maintainer's visual/product sign-off rather than being merged unattended.

### Prior blocker closure

- **Unchanged settings edits do not revalidate a stored path.** The UI includes `default_workdir` in `PATCH /api/kanban/boards/<slug>` only when the trimmed value differs from the original. My live browser drive captured the changed request with the field present, then an unchanged save with the key absent.
- **Change and clear remain functional.** A changed non-empty path goes through `resolve_trusted_workspace`; explicit clear sends `""`; omission preserves the stored value.
- **Default-board settings remain reachable without making the immutable default slug archivable.** `Board settings…` is enabled; `Archive current board…` remains disabled.
- **The earlier layout-test isolation problem is closed.** The final head creates a fresh browser/page per viewport-locale cell and waits on real boot readiness.

### Gate evidence

- **Codex reproduce gate:** `SAFE TO SHIP`. It inspected the rebased worktree, ran 45 focused board tests, verified a real installed-Agent create/preserve/clear/reject round trip, proved legacy name-only/create calls do not receive the new keyword, and independently matched the stable patch ID.
- **Opus security/correctness gate:** `APPROVE`. It verified the trusted-workspace boundary, 400 error mapping, omit/preserve/clear semantics, default-board immutability, locale coverage, and the installed `hermes_cli.kanban_db` signatures.
- **Fable UX gate:** `SHIP-WITH-UX-FIXES`. Placement and hierarchy are sound, the always-visible single-board switcher is discoverable, and the modal is usable cross-device. Its follow-ups are non-blocking polish: put “(optional)” in the label, consider a gear instead of pencil for settings, backfill the adjacent pre-existing English menu strings in `de`/`es`, and consider a `100dvh` fallback.
- **Full serial suite:** `14190 passed, 14 skipped, 1 xfailed, 2 xpassed, 9 warnings, 34 subtests passed` in 752.50s. No failures.
- **Static/private gates:** Python parse, `node -c`, ruff-forward, runtime ESLint, scope-undef, merge-marker scan, and `git diff --check` all clean.
- **Own live browser drive on the rebased server (gate port 8783):** zero console/page errors; changed-save payload includes the trusted path; unchanged-save payload omits it; desktop 1440×900, mobile 390×844, and short 480×320 modal-scroll paths all passed. At 390px the modal stayed within `12..378`, the workspace field within `29..361`, and Save/Cancel remained visible. At 480×320 the modal scrolled `290px`, exposing the workspace field and both actions.
- **Visual inspection:** desktop modal is balanced and unclipped; the single-board `Default ▾` menu fits cleanly; mobile fields and actions remain reachable; the extreme-short viewport exposes the new field and actions through internal scroll. German confirms the new `Board-Einstellungen…` / `Standard-Workspace-Pfad` strings, while adjacent older menu strings remain English localization debt rather than a regression in the new field.

### Non-blocking compatibility note

On an Agent build old enough not to accept the `default_workdir` keyword, actively setting the new field can surface a generic 500; ordinary/unchanged board edits stay compatible because the key is omitted. The installed/current Agent supports the keyword. A future compatibility polish could map that `TypeError` to a clear “newer Hermes Agent required” response.

### Handoff

**Technically green — merge only after the maintainer signs off on the visible Kanban board-settings UX.** Preserve @rodboev attribution. Re-fetch the contributor head before staging; this certificate is valid only while it remains exactly `0f15416cb9a704f90bfc62974ae7e25582f2dbe7`.

---
_Gate-certifier layer (warm-up → **gate** → release). I do not merge, tag, deploy, or close PRs._

nesquena-hermes (changes requested):
Really solid PR overall — the combobox is the right control (Fable confirmed it over both a plain `<select>` and bare text), all 15 locales are already translated, and the phone modal-height fix is a good defensive touch. My screenshots + Fable + the full local suite all looked clean, and I had it staged to ship.

But the release CI caught a real failure in this PR's **own** layout test that I can't ship past:

### `test_board_modal_default_workdir_layout[480-320-de]` — field overflows the viewport at 480×320 in German
On CI (Python 3.13 shard) the final assertion fails:
```
assert field_box["y"] + field_box["height"] <= modal_box["y"] + modal_box["height"]
E   assert (327.359375 + 38.796875) <= (0 + 320)
```
At the 480×320 landscape-phone viewport, the German label ("Standard-Workspace-Pfad", the longest translated label) pushes the "Default workspace path" field down so it renders **below the 320px viewport fold** (`field.bottom ≈ 366 > 320`). It passes locally (font-rendering variance makes the label slightly shorter on some chromium/font stacks), but it reproduces on **every** CI run, so it's a genuine responsive edge, not a flake.

Root: the test scrolls the inner `.kanban-modal` so the field is visible *within* it (that part passes), but then measures the field against the outer `#kanbanBoardModal` (the full-viewport backdrop, 0–320) and finds it below the fold. So at the shortest landscape-phone height with the tallest label, the field isn't reachable within the viewport.

**Fix options (your call which is cleanest):**
1. Tighten the modal's phone layout so the field row stays inside the viewport at 320px height even with the German label — e.g. reduce the modal's top/bottom margin at `≤640px`/short-height, or ensure the inner-scroll region's reachable area is measured against the *inner* modal (which is scrollable) rather than expecting everything to fit the 320px outer box.
2. If the intended contract is "the inner modal scrolls and that's fine," then the assertion at line 80 should measure against the inner `.kanban-modal` scroll box, not the outer `#kanbanBoardModal` — but only if the field is genuinely reachable via scroll on a real 320px device (please verify on a real/emulated 480×320, not just relax the test to go green).

Everything else is ready — once the 480×320 German case is genuinely fixed (field reachable within the viewport, test green on CI), I'll re-gate and ship. Thanks @rodboev.

nesquena-hermes:
Shipped in **exp-v0.52.195** — thanks @rodboev! 🎉

The per-board "Default workspace path" combobox is live on the experimental channel. Tasks on a board without an explicit workspace path now inherit the board default.

Gate: Fable UX **SHIP-UX** ("textbook responsive-dialog pattern"), ESLint + ruff clean, full 5-shard suite green (only the 2 known box-baseline env flakes, unrelated), CI green, maintainer visual approval (mobile portrait + the combobox control confirmed as the right choice over a plain select/text-input).

**Maintainer layout fix folded in** (commits d92aeecfa + b9c44ac49 on your branch, authorship preserved): your own layout test failed on CI at 480×320 landscape for all locales — the centered modal pushed lower rows past the fold. Fixed by top-anchoring the overlay + `100dvh` at short-landscape, and made `test_kanban_modal_mobile_responsive_css` robust to the new override. All 15 layout cases pass.

Follow-up filed (#6906, non-blocking): make the modal max-height cap unconditional so short-but-wide tiled desktop windows can't clip either.

PR #6056, opened 2026-07-14
feat(#6033): show delegated goal titles for child sessions

### LINKED ISSUE #6033: WebUI/agent: give delegated sub-agent sessions descriptive titles (from their goal) instead of the fixed 'Subagent Session' label
Raised by **b3nw** in Discord `#report-bugs` (Jul 13 2026, exp-v0.52.60).

> "is it expected that sub-agent sessions just stay 'Subagent Session' or should they get titles also … now that I realize I can't re-name them, guessing that is current state/limitation."

## The two halves

1. **Rename** — correct as-is. Sub-agent (delegated `delegate_task`) sessions are **view-only** in the WebUI by design: `api/routes.py` returns `"Subagent sessions are view-only and cannot be modified from WebUI"` (400) for rename/branch/duplicate/draft/delete. b3nw self-answered this; no change wanted.

2. **Title** (this enhancement) — every delegated child shows the generic fixed label **"Subagent Session"** rather than anything descriptive. In a sidebar lineage with several sub-agents (see #6027), they're indistinguishable. It would be a real usability win for each to carry a short descriptive title derived from **its delegated goal**.

## Why it's tractable

The delegated goal text is already available agent-side — `agent/display.py` has `_delegate_task_goal_parts()` and renders `delegate_task` goals (single `goal` or per-task goals in a batch). So a subagent session could be titled from a truncated form of its own `goal` at spawn time, instead of the static "Subagent Session".

## Where the fix likely lives (nuance for whoever picks this up)

The "Subagent Session" label is assigned **agent-side by the delegate runner** (it's not a WebUI string — the WebUI renders whatever title the session row carries, and it deliberately does NOT run subagent children through its own auto-title generator since they're view-only and don't go through the WebUI chat/start path). So the cleanest fix is probably in the agent's delegate runner (title the child from its goal at creation), with the WebUI simply displaying it. A WebUI-only fallback (derive a display title from the child's first user/goal message when the stored title is the generic default) is possible but is the weaker layer. Worth confirming with the core repo which side owns this.

## Suggested behavior
- A delegated sub-agent session's title defaults to a short, truncated form of its delegated **goal** (e.g. first ~40 chars), not the fixed "Subagent Session".
- Batch delegations: each child titled from its own task goal.
- Stays view-only (no rename) — unchanged.

## Related
- #6027 — sidebar lineage instability with sub-agents/forks (same sub-agent surface; distinguishable titles would make that tree far more legible).

## Environment
- WebUI: exp-v0.52.60 (experimental)

Scope note: this is a small UX enhancement, not a bug — the current behavior is intended (view-only + generic label). Filed to track the improvement.

rodboev:
## Thinking Path

- Delegated child sessions are intentionally view-only, so rename and regenerate endpoints are the wrong owner for this bug. The missing behavior is on the read-only sidebar metadata path.
- The repo already has a server-side `display_title` overlay for stale generic titles, and the current sidebar consumers already prefer `display_title`. The gap is that delegated child rows never derive one from their first user message.
- The fix stays in `api/models.py`, extends the existing batched state.db metadata overlay for generic `Subagent Session` rows, and leaves stored titles plus delegated-child read-only ownership unchanged.

## What Changed

- `api/models.py`: derive `display_title` for generic delegated child rows from their first state.db user message during the existing batched sidebar metadata pass, without mutating the stored title.
- `tests/test_session_lineage_metadata_api.py`: add focused coverage for delegated goal-derived display titles, custom-title preservation, and fail-open fallback when message data is unavailable.

## Why It Matters

Delegated lineage trees become readable without weakening the view-only boundary that protects child sessions from WebUI writes. Users see the child goal they actually delegated, not the generic placeholder that hides which branch is which.

## Verification

Focused sidebar metadata tests cover generic delegated-child derivation, custom-title preservation, and fail-open fallback; existing sidebar consumer and delegated-child preservation tests keep `display_title`, visibility, and read-only behavior pinned; CI runs the full matrix on 3.11, 3.12, and 3.13.

## Upstream

Closes #6033.

## Model Used

GPT 5.5 via Codex CLI

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -5173,13 +5173,16 @@ def _read_state_db_sidebar_overrides(
             has_messages_table = cur.fetchone() is not None
             messages_has_session_id = False
             messages_has_timestamp = False
+            messages_has_title_fields = False
             if has_messages_table:
                 cur.execute("PRAGMA table_info(messages)")
                 message_cols = {str(row[1]) for row in cur.fetchall()}
                 messages_has_session_id = 'session_id' in message_cols
                 messages_has_timestamp = 'timestamp' in message_cols
+                messages_has_title_fields = {'session_id', 'role', 'content', 'timestamp'}.issubset(message_cols)
 
             overrides: dict[str, dict] = {}
+            delegated_title_ids: set[str] = set()
             ids = list(wanted)
             chunk_size = 500
             for i in range(0, len(ids), chunk_size):
@@ -5200,6 +5203,12 @@ def _read_state_db_sidebar_overrides(
                     if state_title:
                         entry['_state_db_title'] = state_title
                     state_source = str(row['source'] or '').strip().lower()
+                    if (
+                        state_source == 'subagent'
+                        and sid in count_wanted
+                        and ' '.join(state_title.split()) == 'Subagent Session'
+                    ):
+                        delegated_title_ids.add(sid)
                     if state_source:
                         entry['_state_db_source'] = state_source
                         source_meta = normalize_agent_session_source(state_source)
@@ -5244,6 +5253,29 @@ def _read_state_db_sidebar_overrides(
                                 entry['_state_db_last_message_at'] = float(row['last_message_at'] or 0)
                             except (TypeError, ValueError):
                                 pass
+            if messages_has_title_fields and delegated_title_ids:
+                seen_user_messages: set[str] = set()
+                delegated_title_ids = list(delegated_title_ids)
+                for i in range(0, len(delegated_title_ids), chunk_size):
+                    chunk = delegated_title_ids[i:i + chunk_size]
+                    placeholders = ','.join('?' * len(chunk))
+                    cur.execute(
+                        f"""
+                        SELECT session_id, role, content, timestamp
+                        FROM messages
+                        WHERE session_id IN ({placeholders}) AND role = 'user'
+                        ORDER BY session_id, timestamp ASC
+                        """,
+                        chunk,
+                    )
+                    for row in cur.fetchall():
+                        sid = str(row['session_id'])
+                        if sid in seen_user_messages:
+                            continue
+                        display_title = title_from([dict(row)], fallback='')
+                        if display_title:
+                            seen_user_messages.add(sid)
+                            overrides.setdefault(sid, {})['_state_db_display_title'] = display_title
             return overrides
     except Exception:
         return {}
@@ -5298,6 +5330,7 @@ def _apply_sidebar_state_db_override_metadata(sessions: list[dict], metadata: di
         state_db_source_label = entry.pop('_state_db_source_label', None)
         state_db_message_count = entry.pop('_state_db_message_count', None)
         state_db_last_message_at = entry.pop('_state_db_last_message_at', None)
+        state_db_display_title = entry.pop('_state_db_display_title', None)
         if state_db_source == 'webui':
             session['source_tag'] = state_db_source_tag
             session['raw_source'] = state_db_raw_source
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
## Summary

Read the diff at HEAD (`ab48096de`) against `origin/master` plus the full `_read_state_db_sidebar_overrides` / `_apply_sidebar_state_db_override_metadata` / `title_from` context in `api/models.py`, the new tests, and the agent-side delegate path. This is the right layer and the derivation is sound — the delegated goal genuinely *is* the child session's first user message, so titling from it reproduces the goal rather than approximating it.

## Cross-repo contract check (this is the load-bearing assumption)

The whole PR rests on "the first `role='user'` state.db message of a subagent child equals its delegated goal." That holds. In the Agent, `tools/delegate_tool.py:1949` spawns the child with:

```python
return child.run_conversation(
    user_message=goal,
    task_id=child_task_id,
    stream_callback=_relay_child_text,
)
```

So `goal` becomes the child's first user turn, persisted into the subagent's state.db `messages` table. Deriving `display_title` from `title_from([...], fallback='')` over the first non-null user row (models.py, new block ~5253-5278) therefore surfaces the actual delegated goal — matching what `agent/display.py:_delegate_task_goal_parts` renders CLI-side. Good instinct keeping this read-only and out of the rename/branch path, which correctly stays 400 for subagent children.

## The gating is correct

The lookup is pinned to `sid in count_wanted`, which is the same paint-priority top-N set (`HERMES_WEBUI_STATE_DB_OVERRIDE_TOP_N`, default 300) that already caps the expensive `messages` COUNT aggregation. That's the right cap to reuse — it means the extra `SELECT ... FROM messages WHERE role='user'` only runs for rows the sidebar is actually painting, not all 2400+ rows on a power-user poll (the #5132 hot path). `test_generic_subagent_title_respects_sidebar_override_cap` pins exactly this: with the cap at 1, only the newest row gets a derived title.

The apply-side guard is also tight:

```python
if (
    state_db_display_title
    and state_db_source == 'subagent'
    and ' '.join(str(title or '').split()) == 'Subagent Session'
):
    session['display_title'] = state_db_display_title
```

Custom titles stay authoritative (`test_custom_subagent_title_stays_authoritative`), and the null-content guard added to `title_from` (`if c is None: continue`) is what lets `test_generic_subagent_title_skips_null_first_user_message` fall through to the next usable user message instead of rendering `None`. Frontend consumption is confirmed too: `static/sessions.js:6983` renders `s.display_title || s._state_db_title || s.title`, so the derived title paints without any front-end change.

## One non-blocking note

The `messages` query fetches every `role='user'` row for the delegated ids and dedups the first-per-session in Python via `seen_user_messages`, relying on `ORDER BY session_id, timestamp ASC`. For a long-running subagent with many user turns this reads more rows than the one it keeps. It's already bounded by the top-N cap so it's not a hot-path regression, and a correlated `LIMIT 1` per id across an `IN (...)` batch isn't clean in SQLite — so this is fine as-is. If you ever want to trim it, a windowed `ROW_NUMBER() OVER (PARTITION BY session_id ORDER BY timestamp) = 1` would fetch exactly one row per session. Not worth blocking on.

## Verification

All 18 checks are green (lint, browser-smoke, test matrix 3.11/3.12/3.13). The new tests cover the meaningful branches: goal derivation, custom-title preservation, no-first-user-message fallback, null-first-message skip, cap respect, and missing-`messages`-table fallback. Approach and coverage both look complete — LGTM from a read-through.

nesquena-hermes:
Shipped in **exp-v0.52.78** 🚀 Thanks @rodboev.

Your feature landed as-is — generic delegated subagent rows now show a title derived (view-only) from the child's first user message instead of "Subagent Session", without mutating the stored title.

One maintainer hardening rode along, gate-driven: the Codex gate found the newly-derived `display_title` bypassed the sidebar credential redaction (the serializer redacted only `title`), so a credential-shaped delegated goal could surface in the sidebar even with `api_redact_enabled=True`. Fixed as a class-fix — a shared `_redact_sidebar_title_fields()` helper now redacts `display_title` + `_state_db_title` + `parent_title` on **both** `/api/sessions` and all three `/api/sessions/search` branches (which had the same latent leak), reading the setting once per response. +regression tests (the leak test is proven to fail on the unfixed code).

Gate: Codex adversarial re-gate (3 rounds) SAFE TO SHIP; full suite 13,123 pass; ESLint + ruff clean. Closes #6033.

PR #6059, opened 2026-07-14
fix(titlebar): keep titlebar controls clickable in desktop webviews

raulanatol:
## Summary

With "Show profile switcher in titlebar" enabled, the switcher button renders but does not
respond to clicks when the WebUI runs inside a desktop wrapper (native webview). Nothing
happens on click — the dropdown never opens. The same button works correctly in a normal
browser.

## Root cause

`static/style.css` marked the whole titlebar as a window-drag region and relied on a child
opt-out for interactive controls:

```css
.app-titlebar         { -webkit-app-region: drag;    }  /* whole bar */
.app-titlebar-profile { -webkit-app-region: no-drag; }  /* control opts out */
```

A child `no-drag` inside a `drag` parent is unreliable in strict webviews (notably WKWebView
on macOS): the parent `drag` is applied but the child `no-drag` exception is ignored, so the OS
treats the control as window-drag chrome and swallows the click. This is invisible in browsers
(`app-region` is a no-op there), so it only reproduces in the desktop wrapper.

## Fix

Invert the drag model: the bar is no longer draggable; `-webkit-app-region: drag` is scoped to
the inert center (`.app-titlebar-inner`) and spacer (`.app-titlebar-spacer`) only, keeping every
interactive control out of any drag region (the well-supported direction — `drag` child inside a
`no-drag` parent). Window dragging is preserved via the title/center area, the same pattern used
by VS Code, Slack, etc.

## Steps to reproduce

1. Run the WebUI inside a desktop webview wrapper (macOS).
2. Settings → enable "Show profile switcher in titlebar".
3. Click the profile button in the titlebar → nothing happens.
4. Open the same URL in a browser → the button works.

## Testing

- CSS-only change; the JS switch path (`toggleProfileDropdown` → `switchToProfile`) is unchanged.
- Verified in a browser that the titlebar switcher still opens and switches, and that window
  dragging still works via the title/center area.
- Manual webview verification recommended by a reviewer with a desktop wrapper, since the failure
  only reproduces where `-webkit-app-region` is honored.

## Scope

No public contract touched (visual/interaction fix in `static/style.css` only).


nesquena-hermes:
Read the diff against `origin/master` (`static/style.css`, +6/-2) plus the titlebar markup in `static/index.html:121-146` and the JS handlers in `static/panels.js`. The diagnosis is correct and the drag-model inversion is the right fix.

## Why the child `no-drag` was unreliable

`-webkit-app-region` inherits. On master, `.app-titlebar{…-webkit-app-region:drag;}` (style.css:1525) pushes `drag` down to every descendant, and each interactive control then tries to opt back out with `no-drag`:

```css
.app-titlebar-profile{…-webkit-app-region:no-drag;}
.app-titlebar-hamburger{-webkit-app-region:no-drag;…}
```

WKWebView honors a `drag` region but is flaky about a nested `no-drag` exception carving a hole back out of it, so the OS keeps treating the control as window chrome and eats the click. Browsers no-op `app-region` entirely, which is why this only ever reproduced in the desktop wrapper. The inverse nesting (a `drag` child inside a non-drag parent) is the well-supported direction, and that's what the PR switches to.

## The fix keeps every control out of a drag region

After the change, `.app-titlebar` drops `drag` and it's scoped to just the inert center and spacer:

```css
.app-titlebar-inner{…-webkit-app-region:drag;}
.app-titlebar-spacer{-webkit-app-region:drag;}
```

Checked the markup: the interactive controls (`.app-titlebar-left` with the profile button + hamburger at index.html:122-128, plus `.app-titlebar-new-chat` and `.app-titlebar-reload` at :140/:146) are all siblings of `.app-titlebar-inner`, not descendants — so none of them inherit `drag` anymore, and none rely on a `no-drag` carve-out. Window drag is preserved via the centered title/icon area, same pattern VS Code and Slack use.

## No interaction regression on the title itself

`.app-titlebar-title` (the double-click-to-rename target wired in panels.js:104) sits inside `.app-titlebar-inner`, so it's now in the drag region — but it was already in one on master (inherited from `.app-titlebar`), so the dblclick-rename behavior in webviews is unchanged. The mobile tap-to-reveal popover is `ontouchstart`-gated and `app-region` doesn't apply on touch, so that path is untouched too. Net: the profile switcher (and the other buttons) become clickable in WKWebView with nothing else lost.

CSS-only, JS switch path unchanged. The one thing read-through can't cover is the actual WKWebView behavior, so a maintainer with a desktop wrapper confirming the click lands is the sensible gate before merge — the description already flags this. LGTM on read.

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (fix direction correct, but it moves the same swallow-bug onto the title-rename)

**Certified head:** `7bee44c276e76b689cb132311eec132ac5f03c53` (clean-rebased onto `origin/master`)
**Verdict:** The diagnosis and drag-model inversion are right and this unblocks a genuinely broken control (the profile switcher). But the minimal diff puts `-webkit-app-region:drag` on `.app-titlebar-inner`, which **contains the interactive session title** — re-introducing the exact click-swallow class this PR fixes, now on the title's inline-rename. One small fix (drag the inert flanks, not the title container).

### What I ran
- **Senior UX advisor (Fable)**: `SHIP-WITH-UX-FIXES` — fix direction correct, but flagged the title-rename residual + shrunken wide-window drag affordance.
- **My own source verification**: confirmed the finding is real (not theoretical) — see below.
- CSS validity (balanced braces), GitHub CI all-green, drag/no-drag mapping consistency check.
- No pytest signal: this is CSS-only with zero test coverage of the drag regions; a full suite run adds nothing here (verified no test references the titlebar drag CSS).

### What's correct (verified)
The core inversion is sound: `-webkit-app-region:drag` removed from the whole `.app-titlebar`; interactive controls `.app-titlebar-profile` and `.app-titlebar-hamburger` stay `no-drag`. This fixes WKWebView swallowing clicks on the profile switcher (the reported bug). Prior maintainer review (2026-07-14) confirmed the diagnosis. Drag/no-drag mapping is internally consistent for those two controls.

### Reproduced blocker (RED)
**The interactive session title is now inside a drag region.** The PR sets `-webkit-app-region:drag` on `.app-titlebar-inner` (`static/style.css`), and `static/index.html` puts `<span class="app-titlebar-title" id="appTitlebarTitle">` **inside** `.app-titlebar-inner`. That title is interactive: `static/panels.js` has an inline-rename flow (`_renamingAppTitlebar`, swaps the span for an input on rename). So in strict webviews (WKWebView / the Hermes Mac app), a click/dblclick to rename the title can be swallowed by the drag region — the identical bug class the PR fixes for the profile switcher, just relocated to the title. The PR's own comment advocates "keep interactive controls out of any drag region," but the interactive title violates that. (Secondary, per Fable: on wide desktop windows the whole-bar drag affordance shrinks to just the center inner, reducing the window-move surface.)

### Fix (small, same direction the PR already advocates)
Instead of dragging `.app-titlebar-inner` (which holds the title), add two inert `flex:1` filler siblings flanking `.app-titlebar-inner`, each `-webkit-app-region:drag`, and keep the inner (with its interactive title) `no-drag`. This restores full-flank window-drag AND keeps the title-rename out of any drag region — the reliable sibling pattern the PR itself argues for. (Minimal alternative: scope drag to the inert children `.app-titlebar-icon`/`.app-titlebar-sub` rather than the whole inner.)

**Do not merge as-is.** `gate-fail` at `sha:7bee44c276e7`. The fix is close and the direction is right — it just needs the drag moved off the title-bearing container onto inert flank fillers (a few lines of CSS). This is safe for us to apply on the branch per the auto-fix-mechanical-blocker policy, then re-verify. **Final verification is interactive, not a screenshot** (static captures are pixel-identical master-vs-head): the drag behavior + profile-switcher/title-rename clickability must be checked in the actual desktop Mac webview — route to Nathan for that interactive sign-off once the flank-drag fix lands. Preserve @raulanatol author credit.

@raulanatol — the diagnosis and drag-inversion are correct, and this genuinely fixes the profile-switcher click swallow. The one gap: `.app-titlebar-inner` now has `app-region:drag` but it contains the inline-rename-able session title, so the title-rename click can be swallowed the same way. Move the drag onto inert `flex:1` filler siblings flanking the inner (keep the inner/title `no-drag`) and it's fully clean.

raulanatol (commented):
Applied the fix the gate flagged in commit 6ba1c4df.

**Problem:** `-webkit-app-region:drag` was on `.app-titlebar-inner`, which contains the session title (`#appTitlebarTitle`), and that title is interactive (inline rename on double-click, via `panels.js`). In WKWebView a drag region swallows clicks on its interactive descendants, so this reintroduced the exact click-swallow bug this PR fixes — now on the title rename.

**Change:** the drag moves to two inert `flex:1` sibling fillers (`.app-titlebar-drag`) flanking `.app-titlebar-inner`, and the inner (with the title) stays `no-drag`. This preserves the full window-drag surface across both flanks while keeping every interactive control out of any drag region.

**Verified:** balanced CSS braces and layout checked in a browser (symmetric fillers, centered title, drag surface restored). Note: `-webkit-app-region` is a no-op outside a webview, so the browser only confirms layout, not the actual drag behavior. Final verification of the drag + profile/rename clickability is still pending in the Mac app (WKWebView) — @nesquena-hermes left that for Nathan's interactive sign-off.

nesquena-hermes:
## 🟢 Authoritative re-gate — GREEN-engineering (prior RED fixed; parks for Nathan's interactive Mac sign-off)

**Exact contributor head:** `6ba1c4dfdf03375699a9650728963f94a97e18eb`

The prior RED is resolved. I transplanted this exact patch onto current master (`static/index.html` + `static/style.css`, +11/-2). This head applies precisely the fix the prior gate recommended.

### Prior RED — resolved

The prior RED bounced because the minimal diff put `-webkit-app-region:drag` on `.app-titlebar-inner`, which contains the interactive inline-renamable session title — relocating the WKWebView click-swallow onto the title-rename. This head instead:
- removes `-webkit-app-region:drag` from the whole `.app-titlebar`
- adds two inert **empty** `.app-titlebar-drag` `flex:1 1 0` flank fillers (drag) + the mobile spacer (drag)
- makes `.app-titlebar-inner` explicitly `-webkit-app-region:no-drag`

So the interactive title + profile switcher + hamburger + new-chat are all out of any drag region, and full-flank window-drag is preserved — the exact inert-sibling pattern the prior gate specified.

### Verified

- live DOM (1400px): titlebar visible, **2 empty drag fillers (0 children each)**, `inner` computed `no-drag`, title "Chat" centered at ~712px (viewport center 700). No overlap.
- structural invariant: no interactive element sits inside a drag region; CSS braces balanced; JS clean
- threat scan CLEAN; current-master transplant clean
- vision review of the rendered titlebar: centered, no visible layout regression (drag is invisible to layout)
- Fable UX: **SHIP-UX**

### Flag for Nathan's interactive sign-off (visible-UI + wrapper-config, not statically clearable)

1. **Profile chip repositions.** With "Show profile switcher in titlebar" ON at desktop >900px, the chip moves from center-adjacent to the far-LEFT edge (Fable caught this — it's the only visible diff, intentional-looking but should be knowingly approved).
2. **macOS traffic-light overlap.** In a frameless WKWebView wrapper the native traffic lights occupy roughly the top-left ~70px; whether the wrapper pads content past them is wrapper config, not CSS. Master never placed anything far-left at >900px, so this is new territory.
3. **Interactive drag/click behavior** (the whole point — screenshots are pixel-identical to master): (a) window drags from both flanks, (b) profile switcher opens on click, (c) title double-click rename works, (d) far-left profile chip clear of the traffic lights.

### Disposition

**Technical gate: GREEN.** The engineering is correct, minimal, and closes the prior RED. Because it is a visible titlebar change whose real verification is interactive in the actual macOS Mac app, it **parks for @nesquena's sign-off** — recommend the Mac agent build + drive this branch to confirm both the drag and the profile-switcher/title-rename clickability, plus the profile-chip position, before merge. Preserve @raulanatol author credit. A screenshot has been sent for Nathan's glance.

nesquena-hermes:
Re-gated this at the current head (rebased clean onto `dd69bfd9a`) because the prior certificate predates your 2026-08-15 commit. Summary: **the approach is right and I verified it empirically — one test needs updating before it can ship.**

## What I verified myself

The inert-sibling pattern is the correct fix for the WKWebView click-swallow. I drove the real page at four viewports (mobile 390 / tablet 820 / desktop 1440 / wide 1920) on this head and on current master, and compared geometry:

- **Title position and size are pixel-identical to master at every viewport** (desktop `x:717 w:31 h:19` on both, etc.).
- **Long-title behaviour is byte-identical**: with a 60-char title on mobile, both render `titleW:166`, `scrollW:410`, `ellipsis`, same `x`. No new truncation, no layout shift.
- **The drag fillers do not overlap the interactive centre**: `overlapsInner: [false, false]`, and `elementFromPoint()` at the title's centre returns `app-titlebar-title` — so the renamable title is genuinely hit-testable, not covered by a drag region.
- **Drag surface is preserved**: 3 drag elements covering **1361px of the 1440px bar (95%)**, with the gap being exactly the interactive centre. That's the whole point of the pattern and it holds.

So the previous RED (drag region landing on the inline-renamable title) stays fixed, and there's no visual regression.

## The blocker — one test, and it's a fair failure

`tests/test_app_titlebar_restore.py::test_app_titlebar_returns_to_centered_desktop_layout` fails on this head (passes on master; I checked both in isolation, so it's genuinely from this PR, not ordering).

The cause is a brittle exact-string assertion:

```python
assert ".app-titlebar-inner{display:flex;align-items:center;gap:8px;min-width:0;max-width:100%;justify-content:center;}" in STYLE_CSS
```

Your change appends `-webkit-app-region:no-drag;` to that exact rule, so the literal no longer matches. **Nothing about the centered desktop layout actually regressed** — I proved that above with real geometry. The test is asserting on source text rather than behaviour, and its intent ("the titlebar returns to a centered desktop layout") is still satisfied.

Please update that assertion to match the new rule — either extend the literal to include `-webkit-app-region:no-drag;`, or better, assert the individual properties so the test stops breaking on every unrelated addition to the same rule. The sibling `test_app_titlebar_no_longer_contains_tps_chip` / subtitle tests are unaffected.

Rest of the filtered run is green: **384 passed** aside from that one.

## Still needs a real-device check

The one thing I can't verify from Linux is the actual WKWebView drag: that the window still drags from the flanks and that the title/profile-switcher/hamburger/new-chat all remain clickable in the packaged Mac app. The geometry says it should, but `-webkit-app-region` behaviour is the exact thing that only reproduces on the real engine. I'll route that to the maintainer for a Mac pass once the test is updated — no action needed from you there.

Update the assertion and I'll re-gate straight away.

PR #6060, opened 2026-07-14
fix: add source-stamp cache to reduce redundant SQLite/stat calls on session list (#5455)

### LINKED ISSUE #5455: Hermes WebUI feels laggy: /api/sessions and session loads take seconds

webtecnica:
Fixes #5455. Adds short-lived source-stamp cache (500ms TTL) so concurrent requests share stamp computation instead of each independently connecting to SQLite.

### DIFF
--- a/api/route_session_list_cache.py
+++ b/api/route_session_list_cache.py
@@ -222,6 +222,14 @@ def _session_list_cache_set(key: tuple, payload: dict) -> None:
 
 
 def _session_list_cache_clear(profile: str | None = None) -> None:
+    # #5455: invalidate the short-lived stamp cache so any poll arriving
+    # immediately after a structural session change computes a fresh
+    # source stamp instead of returning the cached pre-change one.
+    try:
+        with _SOURCE_STAMP_CACHE_LOCK:
+            _SOURCE_STAMP_CACHE.clear()
+    except Exception:
+        pass
     normalized_profile = _session_list_cache_profile_scope(profile) if profile else None
     with _SESSIONS_CACHE_LOCK:
         global _SESSIONS_CACHE_GLOBAL_INVALIDATION_VERSION
--- a/api/route_session_list_cache.py
+++ b/api/route_session_list_cache.py
@@ -222,14 +222,6 @@ def _session_list_cache_set(key: tuple, payload: dict) -> None:
 
 
 def _session_list_cache_clear(profile: str | None = None) -> None:
-    # #5455: invalidate the short-lived stamp cache so any poll arriving
-    # immediately after a structural session change computes a fresh
-    # source stamp instead of returning the cached pre-change one.
-    try:
-        with _SOURCE_STAMP_CACHE_LOCK:
-            _SOURCE_STAMP_CACHE.clear()
-    except Exception:
-        pass
     normalized_profile = _session_list_cache_profile_scope(profile) if profile else None
     with _SESSIONS_CACHE_LOCK:
         global _SESSIONS_CACHE_GLOBAL_INVALIDATION_VERSION


nesquena-hermes (changes requested):
## Changes requested — this regresses the session-list cache invalidation contract (6 existing tests go red)

Thanks for the perf work on #5455 — avoiding the redundant SQLite/stat calls on every sidebar poll is a real win, and one piece of this PR is the right idea. But as written it changes two established cache-invalidation semantics and breaks 6 existing regression-guard tests. Verified in an isolated sandbox against your head (`894a3e820a87`): `tests/test_session_sidebar_cache.py` is **6 failed / 13 passed**, while the same file on `origin/master` is **19 passed** — so these are introduced, not pre-existing:

```
FAILED test_session_list_cache_source_changed_owner_rebuilds_while_follower_reuses_stale
FAILED test_session_list_cache_owner_returns_stale_and_rebuilds_in_background
FAILED test_session_list_cache_stale_background_rebuild_failure_releases_owner
FAILED test_streaming_widens_cache_freshness_window
FAILED test_streaming_window_still_evicts_past_streaming_ttl
FAILED test_streaming_window_does_not_extend_idle_ttl
```

Two root defects:

### 1. [core] TTL-extend-on-unchanged inverts the TTL contract → unbounded streaming hold-down — `api/route_session_list_cache.py:264-269` (and the mirrored race-recheck at `:249-252`)

```python
# Source unchanged — extend the TTL window ...
_SESSIONS_CACHE[key] = (now, stamp, payload)
_SESSIONS_CACHE.move_to_end(key)
return copy.deepcopy(payload), True
```

On master, an entry aged past its TTL with an unchanged stamp returns `fresh=False` (return stale under `allow_stale`, else `pop` + `None`) — TTL is a *force-revalidate/rebuild* trigger. Rewriting the timestamp to `now` and returning `fresh=True` turns TTL into an indefinitely-extendable window. This is fatal during streaming: `_session_list_cache_source_stamp` deliberately collapses all volatile components to a constant `streaming_marker` for the whole turn (see `:446-456` and the #4808 invariant documented at `:439-443`: "the TTL still forces a periodic rebuild so the streaming session's own count/title stay fresh within the TTL window"). With this change the stamp is always "unchanged" mid-turn → the TTL is extended forever → the streaming session's own `message_count`/`title` never refresh. That's exactly what `test_streaming_window_still_evicts_past_streaming_ttl` and `test_streaming_window_does_not_extend_idle_ttl` guard against.

**Fix:** drop the extend-and-return-fresh branches at `:264-269` and `:249-252`. Keep the original past-TTL semantics (unchanged stamp → stale/evict so the owner-rebuild + streaming periodic-refresh path fires). The valuable optimization — the TTL-first check that skips the stamp computation on *fresh* hits (`:218-229`) — can stay, because it does not change staleness semantics; it delivers most of the #5455 win safely.

### 2. [core] the 500 ms `_SOURCE_STAMP_CACHE` masks real source changes — `api/route_session_list_cache.py:121-131` (read), `:142-149` / `:294` (populate), `:302-310` (clear)

The stamp cache returns a cached source stamp for up to 500 ms without recomputing. It's only invalidated in `_session_list_cache_clear`, which is reached only for structural mutations routed through `publish_session_list_changed`. A state.db commit from a streaming turn / CLI / Desktop-app continuation does **not** call clear — those rely on the source stamp *changing*, and the 500 ms cache hides that change. This breaks the "owner rebuilds synchronously the moment the source stamp changes" guarantee (a ≤500 ms stale-serve window), which is what `test_session_list_cache_source_changed_owner_rebuilds_while_follower_reuses_stale` pins.

**Fix:** either drop `_SOURCE_STAMP_CACHE` entirely (the TTL-first check above already removes the per-fresh-hit stamp cost, so you keep most of the perf win with no correctness change), or make it correctness-safe by keying the cached stamp on a cheap always-fresh discriminator (e.g. the state.db `-wal` `mtime_ns`/`size` stat) so a committed change forces a recompute instead of being hidden.

### Also
- No test added for the new caching behavior — please add a stamp-cache correctness test (mutate the underlying source stamp, assert the very next `get` observes it with no ≤500 ms stale window and the owner rebuilds) and a streaming bound-check (a long turn still refreshes `message_count` within the streaming TTL).
- Minor typos: "Commpute" (`:231`), "re-entrires" (`:120`).

The convention side is clean — `_SOURCE_STAMP_CACHE_LOCK` guards every read/write/clear, the cache is size-capped at 128 (LRU), and keying embeds the profile scope so there's no cross-profile collision. Once the TTL contract is restored and the stamp-mask window is closed, this is a good optimization. Happy to re-review when CI is green.

webtecnica:
Fixed per review: removed TTL-extend-on-unchanged branches, removed _SOURCE_STAMP_CACHE (500ms mask), kept only the TTL-first skip optimization.

webtecnica:
Rebased onto latest master.

webtecnica:
Rebased and re-pushed.

nesquena-hermes (changes requested):
## 🔬 Re-review — the regression is fixed, but the diff now references symbols that don't exist (dead no-op)

Good news first: the TTL-contract regression from the last review is resolved — `tests/test_session_sidebar_cache.py` is back to **19/19 green** (identical to master), and the harmful `_SOURCE_STAMP_CACHE` read/serve path + the TTL-extend-and-return-fresh branch are gone.

The problem is what's left behind. The re-push deleted the entire stamp-cache infrastructure but kept the invalidation stub that referenced it, so `_session_list_cache_clear()` now calls into symbols that are defined nowhere in the repo:

```python
# api/route_session_list_cache.py  _session_list_cache_clear()  (lines 224-232)
def _session_list_cache_clear(profile: str | None = None) -> None:
    # #5455: invalidate the short-lived stamp cache ...
    try:
        with _SOURCE_STAMP_CACHE_LOCK:      # <-- undefined
            _SOURCE_STAMP_CACHE.clear()     # <-- undefined
    except Exception:
        pass
    ...
```

Neither `_SOURCE_STAMP_CACHE` nor `_SOURCE_STAMP_CACHE_LOCK` is defined anywhere:
- `grep -rnE '_SOURCE_STAMP_CACHE(_LOCK)?\s*=' api/` → no assignment/definition.
- The file references them only at lines 229-230.

So on **every** call, `_SOURCE_STAMP_CACHE_LOCK` raises `NameError`, which is a subclass of `Exception` and is silently swallowed by the `except Exception: pass`. The `.clear()` never runs — the block is dead code, and the comment describes a cache that does not exist. The PR merges misleading no-op code and delivers none of its stated purpose (#5455). (`grep -nE '_SOURCE_STAMP_CACHE' api/route_session_list_cache.py` → only 229, 230.)

The suite stays green only because the behavior is coincidentally identical to master — greenness doesn't make the diff correct.

### Fix — pick one

**(a) If this PR is meant to be just the incremental invalidation:** delete lines 224-232 (the whole `try/except` stub). That makes the diff empty — master already handles source-stamp invalidation via `_session_list_cache_source_stamp` + `_SESSIONS_CACHE` versioning — so the PR should then be closed.

**(b) If a real stamp cache is intended:** re-introduce the module-level definitions (`_SOURCE_STAMP_CACHE = {}`, `_SOURCE_STAMP_CACHE_LOCK = threading.RLock()`) **and** the read path — but key it on the `state.db` / `-wal` stat and invalidate it on all commit sources (streaming/CLI/Desktop), with the stamp-cache-correctness + streaming-bound tests the prior review asked for (this is what avoids re-introducing the 500ms stale-serve + TTL-inversion the last round bounced on).

Either way: don't blanket-`except Exception: pass` over the block — that's exactly what hid this bug. Catch the specific exception you expect, or don't catch at all.

webtecnica:
**Fix applied** — removed dead code referencing undefined `_SOURCE_STAMP_CACHE` symbols.

The try/except block in `_session_list_cache_clear` referenced `_SOURCE_STAMP_CACHE` and `_SOURCE_STAMP_CACHE_LOCK` which were never defined anywhere in the module. The `except Exception: pass` silently swallowed the resulting `NameError`, making the entire block a no-op.

Commit: webtecnica/hermes-webui@50d3719

webtecnica:
Closing: removed the dead stamp-cache code that referenced undefined symbols (_SOURCE_STAMP_CACHE/_SOURCE_STAMP_CACHE_LOCK). The try/except block was silently swallowing NameError. No replacement needed — the TTL-first check already avoids redundant source stamp computation on fresh cache hits.

PR #6065, opened 2026-07-14
perf: reduce time-to-first-token on Bridge-API send path (#6006)

### LINKED ISSUE #6006: Reduce time-to-first-token (TTFT) on the shared Bridge-API send→first-token path
## Summary

Time-to-first-token (TTFT) — the delay between sending a message and the first streamed token appearing — has room to improve on the **shared Bridge-API send→first-token path**. This path is used by the browser WebUI **and** every native Bridge-API client (the community Android and iOS apps), so a reduction here benefits all of them.

This is filed off a community report that one native client feels notably slower to first token than the desktop app on the *same* provider/model (i.e. the delay felt app/stack-side, not provider-side). Investigating the shared server path, the architecture is sound but there's measurable per-turn work done before the provider request is issued, and — critically — **no TTFT instrumentation exists today**, so the latency can't currently be quantified.

## What is NOT the problem (verified by tracing current `master`)

- `POST /api/chat/start` does **not** block on the provider. It registers the SSE stream, starts the worker thread, and returns `stream_id` immediately (`api/routes.py`, `_start_chat_stream_for_session`). ✔
- SSE delivery is immediate — the stream drain is a blocking `subscriber.get(timeout=heartbeat)` (`api/routes.py`, `_handle_sse_stream`) that wakes the moment the first token is enqueued in `on_token` (`api/streaming.py`). No poll interval, no batching. ✔
- Warm sessions reuse a cached agent (`SESSION_AGENT_CACHE`, `api/streaming.py`), so agent construction is paid once per session, not per turn. ✔

So client-perceived TTFT is dominated by the synchronous work inside the worker thread **before** the provider request is issued.

### 1. No TTFT instrumentation (do this first)
`api/metering.py` records `first_token_ts`, but only to derive token *rate* — the send→first-token *latency* is never computed or exposed. **The regression can't be measured.** Proposed first step: record `pending_started_at → first on_token` per turn and expose it as `ttft_ms` (metering payload / run journal, and/or a debug response surface). Everything below should be validated against this number.

### 2. Per-turn synchronous pre-provider work (`api/streaming.py`, `_run_agent_streaming`)
Runs on every turn, before the provider sees the request:
- **MCP tool discovery** (`discover_mcp_tools()`) on the hot path — a slow/unreachable server stalls first token. Candidate: skip when the session's toolsets don't include MCP, or when discovery already ran this process.
- **Per-profile config re-read** (`get_config_for_profile_home`) — a YAML parse each turn (deliberate for profile correctness, but memoizable with a fingerprint).
- **Prefill context load** + **toolset resolution** + a **per-session metadata disk read** (`Session.load_metadata_only`) for the enabled-toolsets override.
- **Model resolution + catalog warm** (`resolve_model_provider` / `warm_models_catalog_provenance_if_cold`) — disk I/O on the send path.
- **Credential resolution** + **agent-signature hash** (sha256 over ~15 fields) per turn.

Each is individually small but they sum, and they all precede the provider request. The **cold path** (first message in a session, or after cache eviction / server restart) additionally builds a fresh `AIAgent` (system-prompt/SOUL.md load, tool registration) — the worst-case TTFT. A native client that opens a fresh session per launch would disproportionately hit this cold path, which is a plausible explanation for the "native feels slower" report.

### 3. Browser-only extra round-trip before SSE attach (`static/messages.js`)
After `chat/start` returns `stream_id`, the browser does `await api('/api/chat/stream/status?stream_id=…')` and *then* opens the `EventSource`. That's an extra request-response before the client is even listening. (This is browser-specific, so it's not the native-client cause, but it's latency the browser eats and worth making optimistic.)

## Why Bridge-API-wide

Contributors in §2 live entirely in the shared server (`api/streaming.py`, `api/routes.py`) that every client hits. Reducing them lowers TTFT for the browser WebUI, and any native Bridge-API client, at once.

## Proposed sequencing

1. **Instrument** — add `ttft_ms` (`pending_started_at → first token`) to metering / run-journal + a debug surface. Ship first; makes the rest quantitative.
2. **Measure** — cold vs warm attribution across the §2 items using an isolated instance + a mock/instant provider.
3. **Cut** — gate MCP discovery, memoize the per-turn config/toolset reads with a fingerprint (same pattern as the #5979 provenance snapshot), and make the browser SSE attach optimistic.

Happy to take the instrumentation PR (§1) first since everything else keys off it.

---

*Origin: a community report about first-token latency on a native Bridge-API client; filed here because the fix belongs in the shared server path that all clients use.*

## Prior corroboration

Closed issue **#3588 ("GUI vs TUI Speed")** is the same class from a different angle: a user reported a simple "hi" taking ~49s in the WebUI vs ~9s in the TUI on the *same* local llama.cpp model (reasoning off). Same signature — the WebUI send→first-token path adding latency the leaner client doesn't. Worth revisiting under this issue with real `ttft_ms` numbers.

webtecnica:
Fixes #6006.

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -149,6 +149,26 @@ _STREAMING_CRON_PROFILE_HOME: contextvars.ContextVar[str | None] = contextvars.C
 )
 _STREAMING_CRONJOB_WRAPPER_INSTALLED = False
 
+# ── Pre-provider TTFT-reduction caches (#6006) ──────────────────────────────
+# MCP discovery is idempotent but does import + config read. Run it at most
+# once per process so it never blocks the send→first-token path on later turns.
+_MCP_DISCOVERY_RAN: bool = False
+_MCP_DISCOVERY_LOCK = threading.Lock()
+
+# Per-profile fingerprint cache: once we've resolved config-derived values
+# (_cfg, _prefill_messages, _toolsets, _main_request_overrides) for a given
+# config fingerprint, skip re-derivation on subsequent turns for the same
+# profile. The fingerprint is (config_path, st_mtime_ns, st_size) so any
+# on-disk config edit is picked up on the next turn.
+_CONFIG_DERIVATIONS_CACHE: dict[str, dict] = {}
+_CONFIG_DERIVATIONS_CACHE_LOCK = threading.Lock()
+# Per-session enabled_toolsets cache (#6006): Session.load_metadata_only()
+# reads the session JSON file from disk every turn. Cache the result per
+# session_id so the disk read only happens once per run.
+_SESSION_TOOLSETS_CACHE: dict[str, tuple[list[str] | None, float]] = {}  # session_id -> (enabled_toolsets or None, wall_clock_freshness)
+_SESSION_TOOLSETS_CACHE_LOCK = threading.Lock()
+_SESSION_TOOLSETS_CACHE_TTL = 5.0  # seconds; a session JSON change is reflected within 5s
+
 
 def _stream_writeback_diag_threshold_seconds(environ=None):
     if environ is None:
@@ -7372,11 +7392,23 @@ def _run_agent_streaming(
         # `_servers` by `(profile_home, name)` upstream in hermes-agent; that
         # lives outside this WebUI repo.  This change fixes the headline bug
         # for users who run a single non-default profile per WebUI process.
-        try:
-            from tools.mcp_tool import discover_mcp_tools
-            discover_mcp_tools()
-        except Exception:
-            pass  # MCP not available or not configured — non-fatal
+        #
+        # #6006 TTFT: MCP discovery is idempotent (process-global _servers
+        # registry) and only needs to run once per process — on every
+        # subsequent turn it just re-imports and finds nothing new.  Gate it
+        # behind a process-level flag so the import + config-read overhead
+        # never hits the send→first-token path on later turns.  Re-enters when
+        # _ENV_LOCK mutation above changes HERMES_HOME to ensure the first
+        # discovery after a profile switch is fresh.
+        global _MCP_DISCOVERY_RAN
+        with _MCP_DISCOVERY_LOCK:
+            if not _MCP_DISCOVERY_RAN:
+                try:
+                    from tools.mcp_tool import discover_mcp_tools
+                    discover_mcp_tools()
+                except Exception:
+                    pass  # MCP not available or not configured — non-fatal
+                _MCP_DISCOVERY_RAN = True
 
         # Register a gateway-style notify callback so the approval system can
         # push the `approval` SSE event the moment a dangerous command is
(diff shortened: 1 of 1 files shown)


webtecnica:
Rebased onto latest master.

webtecnica:
Rebased and re-pushed.

nesquena-hermes:
Heads-up @webtecnica — the CI failure here is a real regression, not a flake (it fails `test (3.11|3.12|3.13, 2)` — the same shard on all three Pythons, which is the deterministic signature). I dug into the shard-2 log so you have a precise starting point:

**Failing test:** `tests/test_webui_state_db_context_reconciliation.py::test_webui_streaming_normalizes_trailing_prefill_user_before_current_turn` (line 276)

```
>   assert prefill == [{"role": "assistant", "content": "prefill summary"}]
E   AssertionError: assert [] == [{'role': 'assistant', 'content': 'prefill summary'}]
```

The prefill comes back **empty** where the contract expects the trailing prefill/assistant-continuation message preserved.

**Root cause (from the diff):** the #6006 TTFT optimization caches the config-derived values and skips re-derivation on later turns:

```python
# fingerprint = (config_path, st_mtime_ns, st_size); cache keyed on profile_home
_cfg_fingerprint = f"{_cfg_path}:{_cfg_st.st_mtime_ns}:{_cfg_st.st_size}"
...
_prefill_messages = _cached_derivations['_prefill_messages']   # cache-hit path
```

The problem: the cached tuple includes `_prefill_messages` **after** `_normalize_prefill_messages_before_user_turn(...)` has run — but that normalize step is **session-turn-dependent**, not config-dependent. It normalizes the trailing prefill relative to *this session's* current-turn boundary (`pending_user_message` / the trailing assistant summary). The fingerprint is config-file-only (`path:mtime:size`) and profile-keyed, so it's completely blind to session/turn state. On a cache hit, you reuse a `_prefill_messages` that was normalized under a *different* turn boundary (or the initial empty one), which drops the current session's trailing prefill → `prefill == []`.

**The fix (two clean options):**
1. **Cache only the config-derived, turn-independent part** (`_prefill_context`, toolsets) and re-run `_prefill_messages_with_webui_context(...)` + `_normalize_prefill_messages_before_user_turn(...)` **per-turn** outside the cache. That keeps the expensive file-read/script-launch derivation cached (the actual TTFT win) while the cheap per-turn normalization still runs. This is almost certainly what you want — the normalize step is not the slow part.
2. Or, if you must cache the normalized result, extend the cache key to include the session's prefill-relevant turn state (the trailing-message boundary), not just the config fingerprint — but that's more fragile.

Option 1 is the safer, smaller fix and preserves the perf goal. Once the per-turn normalization runs again, this test should go green. Worth adding a test that exercises the **cache-hit path** with a trailing prefill (two turns, same config) so the boundary-normalization can't silently regress behind the cache again.

Happy to re-review once it's green — the TTFT approach itself (caching the config-derived prefill/toolsets) is sound; it's just that the turn-boundary normalization got pulled inside the cache with it.

PR #6070, opened 2026-07-14
fix(worktrees): resolve canonical repo root so worktrees never nest

ayxuerui:
## Summary

`find_git_repo_root()` (`api/worktrees.py`) resolves the repo root with `git rev-parse --show-toplevel`. From inside a **linked worktree**, that returns the *worktree's own root*, not the canonical clone. Since the result is fed straight into worktree creation (`create_worktree_for_workspace` → agent `_setup_worktree`, which builds `<repo_root>/.worktrees/hermes-XXXX`), a session that starts inside a worktree creates its new worktree **nested inside the existing one**:

```
pkm/.worktrees/hermes-A/.worktrees/hermes-B/.worktrees/hermes-C/...
```

A session spawned from a worktree session nests one level deeper, recursively. With worktree-backed sessions becoming the default, this compounds fast — I hit a real vault nested **three** levels deep in normal WebUI use.

The function's own docstring claimed it "handles nested workspaces and linked worktrees correctly" — this PR makes that true.

## Fix

After `--show-toplevel`, consult `git rev-parse --git-common-dir`, which always resolves to the **main repository's** `.git` even from a linked worktree. Its parent is the canonical clone root, so worktrees are always created at `<canonical>/.worktrees/`, regardless of where the session started.

Scoped conservatively:
- Only redirects for a **normal (non-bare) clone** whose common dir is a `.git` **directory**. Bare repos and unusual layouts keep the prior `--show-toplevel` result.
- Handles both absolute and cwd-relative `--git-common-dir` output (pathlib join is a no-op for an absolute path).
- Any subprocess failure falls back to the previous behavior — never worse than before.
- Single caller (`create_worktree_for_workspace`), so no other call site is affected.

## Design notes

- `--git-common-dir` is the standard git primitive for "the shared git dir behind all worktrees"; its parent is the main working tree for a conventional clone. I preferred it over parsing `git worktree list` (heavier, needs the main entry disambiguated) or reading `.git` gitlink files by hand.
- The companion agent-side helper `_setup_worktree()` in hermes-agent has the same root-resolution issue on the CLI path (`hermes -w`); that's a separate repo and out of scope here. This PR fixes the WebUI path end to end because `find_git_repo_root()` is where the WebUI computes `repo_root` before delegating.

## Tests

Added `test_find_git_repo_root_redirects_from_linked_worktree` to `tests/test_issue1955_worktree_sessions.py` using a **real** worktree-inside-a-worktree fixture: it asserts `--show-toplevel` would (wrongly) return the worktree root, then that `find_git_repo_root()` redirects to the canonical clone — for both the worktree root itself and a subdirectory inside it. The existing `test_find_git_repo_root_uses_git_from_nested_workspace` (plain subdirectory) stays green, proving the subdir path is unaffected.

```
./scripts/test.sh tests/test_issue1955_worktree_sessions.py
  9 passed

# regression sweep across worktree modules
./scripts/test.sh tests/test_issue1955_worktree_sessions.py \
  tests/test_worktree_remove.py tests/test_issue2057_worktree_status.py \
  tests/test_state_db_worktree_recovery.py
  32 passed

scripts/ruff_lint.py
  no new violations on added/modified lines. OK.
```

### DIFF
--- a/api/worktrees.py
+++ b/api/worktrees.py
@@ -312,8 +312,20 @@ def remove_worktree_for_session(session, *, force: bool = False) -> dict:
 def find_git_repo_root(workspace: str | Path) -> Path:
     """Return the enclosing git repo root for *workspace*.
 
-    Use git itself instead of checking ``workspace/.git`` so nested workspaces
-    and linked git worktrees are both handled correctly.
+    Uses git itself instead of checking ``workspace/.git`` so nested
+    subdirectories and linked git worktrees are both handled correctly.
+
+    When *workspace* lives inside a **linked worktree** — e.g. a session that
+    started in ``<repo>/.worktrees/hermes-XXXX`` — ``git rev-parse
+    --show-toplevel`` returns that worktree's own root, not the canonical
+    clone. Feeding that root to worktree creation nests the new worktree
+    *inside* the existing one (``.worktrees/hermes-A/.worktrees/hermes-B``);
+    a session spawned from that worktree then nests one level deeper again,
+    recursively. ``git rev-parse --git-common-dir`` always resolves to the
+    main repository's ``.git`` even from a linked worktree, so its parent is
+    the canonical clone root. We redirect to it for a normal (non-bare) clone
+    whose common dir is a ``.git`` directory; bare repos and unusual layouts
+    keep the ``--show-toplevel`` result (prior behavior).
     """
     ws = Path(workspace).expanduser().resolve()
     if not ws.is_dir():
@@ -334,7 +346,33 @@ def find_git_repo_root(workspace: str | Path) -> Path:
     root = result.stdout.strip()
     if not root:
         raise ValueError("Workspace is not inside a git repository")
-    return Path(root).expanduser().resolve()
+    toplevel = Path(root).expanduser().resolve()
+
+    # Redirect to the canonical (main) worktree when *ws* is inside a linked
+    # worktree, so new worktrees are never created nested inside another one.
+    try:
+        common = subprocess.run(
+            ["git", "rev-parse", "--git-common-dir"],
+            cwd=ws,
+            text=True,
+            capture_output=True,
+            timeout=5,
+            check=False,
+        )
+    except (OSError, subprocess.TimeoutExpired):
+        return toplevel
+    if common.returncode == 0:
+        common_dir = common.stdout.strip()
+        if common_dir:
+            # --git-common-dir may print an absolute path or one relative to
+            # cwd; joining onto *ws* is a no-op for an absolute path in
+            # pathlib, so this handles both forms.
+            common_path = (ws / common_dir).resolve()
+            if common_path.name == ".git":
+                canonical = common_path.parent
+                if canonical.is_dir():
+                    return canonical
+    return toplevel
 
 
 def _setup_agent_worktree(repo_root: str) -> dict:
--- a/tests/test_issue1955_worktree_sessions.py
+++ b/tests/test_issue1955_worktree_sessions.py
@@ -105,6 +105,45 @@ def test_find_git_repo_root_uses_git_from_nested_workspace(tmp_path):
     assert find_git_repo_root(nested) == repo.resolve()
 
 
+def test_find_git_repo_root_redirects_from_linked_worktree(tmp_path):
+    """A workspace inside a linked worktree must resolve to the CANONICAL
+    clone, not the worktree's own root — otherwise new worktrees get created
+    nested inside the existing one, recursively (the nesting bug)."""
+    from api.worktrees import find_git_repo_root
+
+    repo = tmp_path / "repo"
+    repo.mkdir()
+    subprocess.run(["git", "init"], cwd=repo, check=True, capture_output=True)
+    subprocess.run(
+        ["git", "-c", "user.email=t@t", "-c", "user.name=t",
+         "commit", "--allow-empty", "-m", "init"],
+        cwd=repo, check=True, capture_output=True,
+    )
+    # Create a linked worktree the same way a session would: <repo>/.worktrees/hermes-XXXX
+    linked = repo / ".worktrees" / "hermes-abcd1234"
+    subprocess.run(
+        ["git", "worktree", "add", "-b", "hermes/hermes-abcd1234", str(linked)],
+        cwd=repo, check=True, capture_output=True,
+    )
+    assert linked.is_dir()
+
+    # Sanity: --show-toplevel (the OLD behavior) would return the worktree root.
+    tl = subprocess.run(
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
Read the diff at HEAD (`api/worktrees.py`, `tests/test_issue1955_worktree_sessions.py`) against `origin/master`, plus the caller chain on the WebUI side and the agent-side helper it delegates to. The diagnosis holds up end to end — this is a real recursive-nesting bug and the fix targets the right layer.

## Confirming the nesting mechanism

`create_worktree_for_workspace` feeds `find_git_repo_root(workspace)` straight into the agent helper as `repo_root` (`api/worktrees.py:377-378`):

```python
def create_worktree_for_workspace(workspace: str | Path) -> dict:
    repo_root = find_git_repo_root(workspace)
    info = _setup_agent_worktree(str(repo_root))
```

On the agent side, `_setup_worktree(repo_root)` (`cli.py:1425`) builds the target path by trusting that value verbatim:

```python
worktrees_dir = Path(repo_root) / ".worktrees"
worktrees_dir.mkdir(parents=True, exist_ok=True)
wt_path = worktrees_dir / wt_name
```

So when a session starts inside `<repo>/.worktrees/hermes-A`, the old `--show-toplevel` returns that worktree's own root, and the new worktree lands at `<repo>/.worktrees/hermes-A/.worktrees/hermes-B` — nesting one level per generation, exactly as described. There's no clamp on that agent helper (`web_git.worktree_add` uses `_main_root`, but `_setup_worktree` does not), so fixing `find_git_repo_root` is the correct and complete fix for the WebUI path.

## The fix is sound

Switching to `git rev-parse --git-common-dir` and taking its parent is the right primitive. Two details I checked and liked:

- The `common_path.name == ".git"` guard means submodules (whose common dir is `<super>/.git/modules/<name>`) and bare repos fall through to the prior `--show-toplevel` result, matching the "unusual layouts keep prior behavior" claim. Good conservative scoping.
- In the *main* worktree, `--git-common-dir` prints a relative `.git`, so `(ws / ".git").resolve().parent` collapses back to `toplevel` — a clean no-op rather than a special case.

The new test builds a real worktree-inside-a-worktree, asserts `--show-toplevel` would wrongly return the linked root, then asserts the redirect — covering both the worktree root and a subdir inside it. That's the right shape.

## One note, not a blocker

The agent-side `cli.py:_setup_worktree` has the identical trust-the-caller issue on the `hermes -w` path (`repo_root or _git_repo_root()`), which the PR correctly flags as separate-repo/out-of-scope. Worth an agent-side follow-up so the CLI path can't nest either, but nothing to change here. Note also that the sibling `hermes_cli/web_git.py` and `kanban_db.py:5550` already reach for `--git-common-dir`, so this brings `worktrees.py` in line with the established convention.

LGTM on the WebUI side.

nesquena-hermes (changes requested):
## 🔬 Gate review (Codex) — SHIP ONLY WITH FIXES (core fix works; one submodule-worktree edge can still nest)

Thanks @ayxuerui — the fix is verified for the common cases: a conventional linked worktree now resolves to the **main clone** (from both the worktree root and a nested subdir), the main-clone case is unchanged, a detached linked worktree resolves canonically, non-git/bare/missing-metadata still fail with the prior `ValueError` (no crash), and a real-git creation probe put the next worktree at `<main>/.worktrees/hermes-probe` (sibling, not nested). The test (`test_issue1955_worktree_sessions.py:108`) creates a real linked worktree + proves the old `--show-toplevel` behavior. One edge keeps "never nest" from being universally true:

1. **[CORE — submodule linked worktree can still nest] `api/worktrees.py:351`** — for a linked worktree of a **submodule**, the common directory is typically `<parent>/.git/modules/<submodule>`, so the `.git`-basename check doesn't recognize it as a linked worktree and falls back to `common_path.parent` (the linked root) → a worktree created there can still nest. **Fix:** detect an actual linked worktree by comparing absolute `--git-dir` vs `--git-common-dir` (equal = main worktree, keep `toplevel`; differ = linked, derive the primary non-bare worktree) instead of the `.git`-basename heuristic + `common_path.parent`.

2. **[test] `tests/test_issue1955_worktree_sessions.py:108`** — add a real `--separate-git-dir=<metadata>/.git` case proving the main-worktree call is unchanged, and (if the "never nest" guarantee is meant to be universal) a linked-submodule case.

Net: the everyday linked-worktree nesting bug is genuinely fixed and the normal path is safe — it's specifically the submodule/`--separate-git-dir` layout that the `.git`-basename check misses. Switching the detection to `--git-dir` vs `--git-common-dir` closes it universally. Rebased clean on current master; credit @ayxuerui.

nesquena-hermes:
## Gate result: SHIP ONLY WITH FIXES — one reproduced worktree-misplacement edge on `--separate-git-dir` / symlinked `.git`

Thanks @ayxuerui — the core diagnosis and fix direction are right: switching to `--git-common-dir` correctly prevents the worktree-inside-worktree nesting, and I verified (Codex, real git repos) that **ordinary main + nested workspaces, conventional linked and detached worktrees, workspace-root symlinks, direct submodules, bare repos, and non-git paths are all unaffected** and behave correctly. But the basename check has one reproduced hole that needs closing before it can land.

### [Reproduced] `--separate-git-dir` and symlinked `.git` place worktrees in the metadata dir
When a repo is created with `git init --separate-git-dir=<metadata>/.git` (or has a symlinked `.git`), `Path.resolve()` follows the redirect, and the `common_path.name == ".git"` basename check treats the **metadata directory's parent** as canonical instead of the working tree.

**Reproduced with real git repositories:** a working tree at `/work` resolved to `/metadata`, and `create_worktree_for_workspace()` then created `/metadata/.worktrees/hermes-probe` — inside the git metadata directory rather than under the working tree. The production agent helper trusts this value for `.worktrees`, `.gitignore`, and `.worktreeinclude`, so this silently misplaces worktree scaffolding.

### Fix (`api/worktrees.py:353-374`)
Also resolve `git rev-parse --git-dir` and compare it to the resolved `--git-common-dir`:
- when they are **equal**, this is NOT a linked worktree → return `toplevel` (the prior behavior);
- apply the canonical-parent redirect **only** when they differ (a proven linked worktree).

Add regression cases for `--separate-git-dir=<metadata>/.git` and a symlinked `.git` (assert the worktree lands under the working tree, not the metadata dir), alongside the existing worktree-inside-worktree test.

The nesting fix itself is good and everything else in the diff verified clean — this is a targeted guard tightening so unusual `.git` layouts can't be misread as linked worktrees. Ping me when it's re-pushed and I'll re-gate the exact head.

_(Codex regression gate reproduced the misplacement with real `--separate-git-dir` repos; focused repository tests otherwise green.)_

PR #6078, opened 2026-07-15
fix(#6067): keep artifact file names visible

### LINKED ISSUE #6067: Artifacts tab: long paths hide the file name
# Artifacts tab: long paths hide the file name

## Thinking Path

- **Problem:** the workspace panel's **Artifacts** tab renders the full
  (workspace-relative) file path on a single ellipsised line. Once a path is
  a few directories deep, the leading directory portion pushes the file
  name off the visible end of the row.
- **Why it matters:** the file name is the primary anchor a user
  recognises; it is the first thing that gets cut today.
- **What I'd like feedback on:** the two options below — three-line layout
  (A) vs two-line wrap (B) — before any code lands, since both are valid
  and the right call depends on what the tab is *meant* to communicate.

## Summary

In the workspace panel's **Artifacts** tab, each row renders the full
(workspace-relative) file path on a single line with CSS
`text-overflow: ellipsis`. Once a path is more than a few directories
deep (e.g. `src/api/routes/users.py`), the directory portion pushes the
file name off the visible end of the row, so the most identifying piece
of information — the file name — is the first thing the user loses.

The row already has a small secondary line for the originating tool name
(`patch`, `write_file`, …), but it sits as muted text below the
clipped path rather than as a structured element.

## Motivation

The Artifacts tab is meant to give a quick answer to "what did the agent
actually edit?". Today, that answer is hard to read at a glance: deep
paths truncate the file name, shallow paths look fine, and the user has
to decode an ellipsised chain of directories to find what changed.

The current behaviour is technically consistent (it does show the path
— just too much of it) and not strictly broken, so I think this is FR
material rather than a bug. Two equally valid layout fixes exist, and
the right one depends on what the tab is *meant* to communicate. That
is a design call maintainers should make before code lands.

## Proposal

I see two options; both preserve the existing `{path, source}` data
shape, the click-to-open behaviour (`openArtifactPath` → Files tab →
`openFile`), and the existing per-artifact button styling.

#### Option A — Three-line layout per artifact

| Line | Content | Style |
|---|---|---|
| 1 | File name (last path segment) | primary, never ellipsised |
| 2 | Workspace-relative directory (path minus the file name) | muted; wraps freely or left-ellipsised so the nearest parent stays visible |
| 3 | Type / source pill (`patch`, `write_file`, `web_extract`, `session`) | small badge |

**Pros** — file name is unambiguously the primary element.
**Cons** — taller rows; ~50% more scroll for the current 50-item cap.

#### Option B — Two-line wrap (file name on line 1, path on line 2)

| Line | Content | Style |
|---|---|---|
| 1 | File name (last path segment) + small inline source/type pill | primary; pill on the right |
| 2 | Full workspace-relative path | muted, wraps freely instead of clipping |

**Pros** — minimal layout change, visually closest to today, full path
readable word-for-word.
**Cons** — wrapped paths can still bloat vertically on narrow panel
widths; does not collapse the directory chain.

## Non-goals

- Sorting, grouping, pinning, or starring artifacts.
- Server-side changes — `collectSessionArtifacts` keeps its 50-item
  cap and current `{path, source}` shape.
- The Files tab file-tree layout (`.file-tree`), which is a separate
  rendering pipeline.
- Changing the tab counter or the empty-state strings.

1. **Hierarchy signal:** should the file name be the bold dominant
   element (both options above), or should the directory stay primary
   with the file name as a smaller subtitle?
2. **Type visibility:** is the tool name worth keeping on every row, or
   should it move to hover / `title` attribute only?
3. **i18n:** any new visible strings (e.g. the "session" fallback pill
   label) should go through `static/i18n.js` and the existing
   `data-i18n` mechanism, matching how `workspace_artifacts_tab` is
   already wired — is that the right pattern for new strings here?

## Verification expectations

For whichever option lands:

- A session whose artifacts list contains paths of varying depth
  (1 segment, 3 segments, 6+ segments) shows every file name fully
  visible in both default-narrow and expanded panel widths.
- Keyboard tab order through the row list is unchanged (one focusable
  button per artifact).
- Click-through to the Files tab still resolves correctly for both
  short and long paths.
- `./scripts/test.sh` is the canonical local runner per
  `CONTRIBUTING.md` §2; for a UI-only change, before/after evidence is
  the main review artifact.

## Ask

I'd like maintainer feedback on **which option to pursue (A or B)** and
on the open questions above before I send a PR. Once the direction is
set, the PR itself will be small — purely client-side changes in
`static/workspace.js` and `static/style.css`, no API surface change.

rodboev:
## Thinking Path

- The Artifacts tab exists to answer which file changed, but the current one-line path row clips the file name first on deep paths.
- The renderer already keeps the original path for click-through, so the fix can split display responsibilities without touching path resolution or artifact collection.
- This change gives the file name its own non-clipping line, keeps the nearest parent visible through a middle-ellipsised directory line, keeps the missing-source label English-owned, proves the split edge cases, and now bounds an overlong nearest-parent tail inside the mobile drawer instead of letting it widen the card.

## What Changed

- `static/workspace.js`: split artifact display paths into filename, directory-head, and directory-tail presentation while preserving the original `data-artifact-path` and click-through behavior.
- `static/style.css`: replace the old single-line path clip with a filename-first layout, a bounded middle-ellipsised directory line, and a compact source pill; the directory row now clips to the card width and an overlong nearest-parent tail ellipsises inside the row instead of widening the drawer.
- `static/i18n.js`: keep `workspace_artifact_source_session` in `en` and let the other locales fall back through the existing `_locale[key] ?? LOCALES.en[key]` path.
- `tests/test_issue6067_artifact_paths.py`: expand the headed browser regression so it also covers `dir/foo.py`, `/file.txt`, `dir/`, `''`, `src//users.py`, `src/lastIndexOf/users.py`, the workspace-prefixed hostile path, and the exact long-parent mobile reproduction from review while checking root and row containment.

## Why It Matters

Users can identify edited files immediately even when their workspace paths are deep, while still keeping enough directory context to tell same-name files apart. The long-parent case now stays inside the mobile drawer instead of pushing the whole artifact card sideways.

## Verification

Focused local reruns keep the English-owned source fallback guard green, and the exact long-parent mobile fixture now shows the pre-fix overflow at 299→545 and 281→536, then the staged CSS holding those same surfaces at 299→299 and 281→281 with `summary.json` still fully visible. CI still runs the full browser-backed matrix.

## Risks / Follow-ups

- This follows the collaborator's Option A recommendation in https://github.com/nesquena/hermes-webui/issues/6067#issuecomment-4971616758, not a separate maintainer confirmation. If the maintainer prefers Option B's wrapped-path layout, the rework stays confined to the same renderer and CSS files.
- Rows are taller than the current single-line presentation. That is the cost of keeping filenames fully visible while preserving directory context.
- A parent segment longer than the card now ellipsises inside the nearest-parent span instead of expanding the drawer. That keeps containment correct, but it is still a truncation path for pathological directory names.

## Upstream

Closes #6067.

Thanks to @claw-io for the issue report and the example paths, and to the collaborator comment for the filename-first plus middle-ellipsis direction that this patch implements.

## Screenshots

Before, the Artifacts tab clips the right edge of deep paths, so the filename loses first on narrow rows:

Before

After, the filename gets its own line, the nearest parent stays visible, and the source pill remains readable:

After

## Model Used

GPT-5 via Codex CLI

### DIFF
--- a/static/i18n.js
+++ b/static/i18n.js
@@ -464,6 +464,7 @@ const LOCALES = {
     workspace_close_preview: 'Close preview',
     workspace_files_tab: 'Files',
     workspace_artifacts_tab: 'Artifacts',
+    workspace_artifact_source_session: 'session',
     workspace_hidden_files_visible: 'hidden visible',
     workspace_hidden_files_visible_title: 'Hidden files are visible — click for options',
     workspace_options: 'Workspace options',
--- a/static/style.css
+++ b/static/style.css
@@ -6431,9 +6431,11 @@ main.main > .main-view:not([id="mainChat"]):not([id="mainSettings"]) .main-view-
 .workspace-artifact-empty{padding:16px 8px;color:var(--muted);font-size:12px;line-height:1.5;text-align:center;}
 .workspace-artifact-item{display:block;width:100%;text-align:left;background:var(--surface-subtle);color:var(--text);border:1px solid var(--border);border-radius:8px;padding:8px 10px;margin-bottom:6px;cursor:pointer;}
 .workspace-artifact-item:hover{border-color:var(--accent);}
-.workspace-artifact-name{font-family:'SF Mono',ui-monospace,monospace;font-size:13px;font-weight:600;white-space:nowrap;overflow:hidden;text-overflow:ellipsis;color:var(--text);}
-.workspace-artifact-dir{font-family:'SF Mono',ui-monospace,monospace;font-size:11px;white-space:nowrap;overflow:hidden;text-overflow:ellipsis;color:var(--muted);margin-top:1px;}
-.workspace-artifact-meta{margin-top:3px;color:var(--muted);font-size:11px;}
+.workspace-artifact-filename{font-family:'SF Mono',ui-monospace,monospace;font-size:12px;overflow-wrap:anywhere;}
+.workspace-artifact-directory{display:flex;min-width:0;white-space:nowrap;font-family:'SF Mono',ui-monospace,monospace;font-size:11px;color:var(--muted);}
+.workspace-artifact-directory-head{min-width:0;flex:0 1 auto;overflow:hidden;text-overflow:ellipsis;}
+.workspace-artifact-directory-tail{flex:0 0 auto;}
+.workspace-artifact-meta{display:inline-block;margin-top:4px;padding:1px 5px;border:1px solid var(--border);border-radius:999px;color:var(--muted);font-size:11px;line-height:1.3;}
 .workspace-artifacts-count{opacity:.65;font-size:11px;}
 .rightpanel[data-active-tab="artifacts"] .breadcrumb-bar,.rightpanel[data-active-tab="artifacts"] .file-tree,.rightpanel[data-active-tab="artifacts"] #wsEmptyState{display:none!important;}
 .rightpanel[data-active-tab="artifacts"] .workspace-artifacts{display:block;}
--- a/static/style.css
+++ b/static/style.css
@@ -6432,9 +6432,9 @@ main.main > .main-view:not([id="mainChat"]):not([id="mainSettings"]) .main-view-
 .workspace-artifact-item{display:block;width:100%;text-align:left;background:var(--surface-subtle);color:var(--text);border:1px solid var(--border);border-radius:8px;padding:8px 10px;margin-bottom:6px;cursor:pointer;}
 .workspace-artifact-item:hover{border-color:var(--accent);}
 .workspace-artifact-filename{font-family:'SF Mono',ui-monospace,monospace;font-size:12px;overflow-wrap:anywhere;}
-.workspace-artifact-directory{display:flex;min-width:0;white-space:nowrap;font-family:'SF Mono',ui-monospace,monospace;font-size:11px;color:var(--muted);}
+.workspace-artifact-directory{display:flex;min-width:0;overflow:hidden;white-space:nowrap;font-family:'SF Mono',ui-monospace,monospace;font-size:11px;color:var(--muted);}
 .workspace-artifact-directory-head{min-width:0;flex:0 1 auto;overflow:hidden;text-overflow:ellipsis;}
-.workspace-artifact-directory-tail{flex:0 0 auto;}
+.workspace-artifact-directory-tail{flex:0 0 auto;min-width:0;max-width:100%;overflow:hidden;text-overflow:ellipsis;}
 .workspace-artifact-meta{display:inline-block;margin-top:4px;padding:1px 5px;border:1px solid var(--border);border-radius:999px;color:var(--muted);font-size:11px;line-height:1.3;}
 .workspace-artifacts-count{opacity:.65;font-size:11px;}
 .rightpanel[data-active-tab="artifacts"] .breadcrumb-bar,.rightpanel[data-active-tab="artifacts"] .file-tree,.rightpanel[data-active-tab="artifacts"] #wsEmptyState{display:none!important;}
--- a/static/workspace.js
+++ b/static/workspace.js
@@ -586,16 +586,26 @@ function renderSessionArtifacts(){
     if(normWs && p.startsWith(normWs)) return p.slice(normWs.length);
     return p;
   };
+  const splitArtifactDisplayPath = (path) => {
+    const slash = path.lastIndexOf('/');
+    if(slash < 0) return {name: path, head: '', tail: ''};
+    const directory = path.slice(0, slash + 1);
+    const parentSlash = directory.lastIndexOf('/', directory.length - 2);
+    return {
+      name: path.slice(slash + 1),
+      head: directory.slice(0, parentSlash + 1),
+      tail: directory.slice(parentSlash + 1),
+    };
+  };
   root.innerHTML = items.map(item => {
-    const dPath = displayPath(item.path);
-    const idx = dPath.lastIndexOf('/');
-    const fileName = idx >= 0 ? dPath.slice(idx + 1) : dPath;
-    const dirPath = idx >= 0 ? dPath.slice(0, idx) : '';
-    return `<button type="button" class="workspace-artifact-item" data-artifact-path="${esc(item.path)}" onclick="openArtifactPath(this.dataset.artifactPath)">
-      <div class="workspace-artifact-name">${esc(fileName)}</div>
-      ${dirPath ? `<div class="workspace-artifact-dir">${esc(dirPath)}</div>` : ''}
-      <div class="workspace-artifact-meta">${esc(item.source || 'session')}</div>
-    </button>`;
+    const path = displayPath(item.path);
+    const parts = splitArtifactDisplayPath(path);
+    const directory = (parts.head || parts.tail)
+      ? `<div class="workspace-artifact-directory"><span class="workspace-artifact-directory-head">${esc(parts.head)}</span><span class="workspace-artifact-directory-tail">${esc(parts.tail)}</span></div>`
+      : '';
+    const source = item.source ? esc(item.source) : esc(t('workspace_artifact_source_session') || 'session');
+    const sourceAttrs = item.source ? '' : ' data-i18n="workspace_artifact_source_session"';
+    return `<button type="button" class="workspace-artifact-item" title="${esc(path)}" data-artifact-path="${esc(item.path)}" onclick="openArtifactPath(this.dataset.artifactPath)"><div class="workspace-artifact-filename">${esc(parts.name)}</div>${directory}<div class="workspace-artifact-meta"${sourceAttrs}>${source}</div></button>`;
   }).join('');
 }
 
--- a/tests/test_chinese_locale.py
+++ b/tests/test_chinese_locale.py
@@ -5,7 +5,10 @@ from tests.test_issue2147_profile_concept_help import PROFILE_CONCEPT_KEYS
 
 
 REPO = Path(__file__).resolve().parent.parent
-PROFILE_CONCEPT_FALLBACK_KEYS = set(PROFILE_CONCEPT_KEYS)
+PROFILE_CONCEPT_FALLBACK_KEYS = {
+    *PROFILE_CONCEPT_KEYS,
+    "workspace_artifact_source_session",
+}
 
 
 def read(path: Path) -> str:
--- a/tests/test_czech_locale.py
+++ b/tests/test_czech_locale.py
@@ -5,7 +5,10 @@ from tests.test_issue2147_profile_concept_help import PROFILE_CONCEPT_KEYS
 
 
 REPO = Path(__file__).resolve().parent.parent
-PROFILE_CONCEPT_FALLBACK_KEYS = set(PROFILE_CONCEPT_KEYS)
+PROFILE_CONCEPT_FALLBACK_KEYS = {
+    *PROFILE_CONCEPT_KEYS,
+    "workspace_artifact_source_session",
+}
 
 
 def read(path: Path) -> str:
--- a/tests/test_issue6067_artifact_paths.py
+++ b/tests/test_issue6067_artifact_paths.py
@@ -1,0 +1,257 @@
+import os
+from pathlib import Path
+import re
+import sys
+
+import pytest
+
+ROOT = Path(os.environ.get("ISSUE6067_REPO_ROOT") or Path(__file__).resolve().parents[1])
+sys.path.insert(0, str(ROOT / "tests"))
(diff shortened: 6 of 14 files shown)


rodboev:
@cutter-sh's great, but looks confused here. It has to expand the workspace panel to capture the change in this PR.

nesquena-hermes:
Reading the diff at HEAD `869fccc` against `origin/master` — `static/workspace.js` (the new `splitArtifactDisplayPath` + row template at :589-608), `static/style.css:6410-6414`, and the `static/i18n.js` locale additions — the filename-first approach is sound and the split math holds across the cases I traced.

## Core logic checks out

The head/tail split correctly pins the *nearest* parent directory (never ellipsised) and lets only the upper directories collapse:

```js
const directory = path.slice(0, slash + 1);
const parentSlash = directory.lastIndexOf('/', directory.length - 2);
return { name: path.slice(slash + 1),
         head: directory.slice(0, parentSlash + 1),
         tail: directory.slice(parentSlash + 1) };
```

Traced against the example paths:
- `src/api/routes/users.py` → name=`users.py`, head=`src/api/`, tail=`routes/`
- `dir/foo.py` → name=`foo.py`, head=``, tail=`dir/`
- bare `foo.py` (no slash) → name only, directory row suppressed by the `parts.tail ? … : ''` guard
- `/foo.py` (root) → tail=`/`, so the leading slash still renders — no drop

Paired with `.workspace-artifact-directory-head{flex:0 1 auto;overflow:hidden;text-overflow:ellipsis}` / `-tail{flex:0 0 auto}` (style.css:6412-6413), the flex shrink lands on the head while the filename (`overflow-wrap:anywhere`, :6410) and nearest parent stay fully visible. That matches the Option A direction from #6067 and does address the "filename loses first" symptom.

## One note on the i18n additions

The PR description calls `workspace_artifact_source_session` "a localized fallback," but every one of the 15 locale blocks receives the literal English string `'session'` — the `ja` (日本語), `zh` (中文/產物), and `ru` (Артефакты) blocks all get `workspace_artifact_source_session: 'session'`, not a translation. So the key exists everywhere (good — `t()` at i18n.js:25999 falls back to `LOCALES.en[key]` then the raw key, so a *missing* block would have surfaced the bare key text; I confirmed 15/15 parity, so that risk is avoided), but non-English locales will still show `session` verbatim.

Severity is low: `collectSessionArtifacts()` (workspace.js:520) almost always supplies a truthy `source` (the tool/kind name — `write`, `edit`, `tool`, etc.), and the `data-i18n` fallback only fires when `item.source` is empty. Also non-regressive vs master, which hardcoded `'session'`. Still, if genuine localization was the intent, the CJK/RU/etc. values want real translations; otherwise a one-line note that it's an English placeholder would keep the description accurate.

Nothing here is blocking. The renderer, CSS layout, and click-through preservation (`data-artifact-path` + `title` tooltip carrying the full path) all look correct.

nesquena-hermes (changes requested):
## 🔬 Correctness gate (Codex, self-rebased clean off master) — SHIP ONLY WITH FIXES (solid core; a root-slash split bug + the i18n all-locale duplication)

Thanks @rodboev — the core is in good shape: no XSS (all three path parts + title + source are `esc()`'d, and a headed-Chromium run rendered hostile `<>&\"`-laden paths inert), no click-through regression (the full original path stays in `data-artifact-path`; the display-split is never fed into path resolution), and the test is genuinely behavioral — it fails on `origin/master` and passes on this branch. But three items block it (independent of the visual/screenshot sign-off, which is a separate maintainer step):

1. **[DISPLAY — root-slash path splits wrong] `static/workspace.js:603`** — the directory row only renders when `parts.tail` is non-empty, so a root-level path like `/file.txt` (head=`/`, tail=``) drops its directory row. **Fix:** render the directory row when **either** `parts.head` or `parts.tail` is non-empty.

2. **[i18n — English copied into every locale instead of `en`-owned + fallback] `static/i18n.js`** — `workspace_artifact_source_session: 'session'` is duplicated into all ~14 locale blocks (lines 2223, 3962, 5612, 7323, 8977, 10675, 12461, 14095, 15725, 17480, 20418, 20885, 22638, 24339) rather than defined once in `en` (line 467) and left to the documented fallback. That's the fallback-defeating anti-pattern. **Fix:** remove the non-English entries, keep only the `en` key, and register it as fallback-owned in the locale-parity tests.

3. **[test — edge cases missing] `tests/test_issue6067_artifact_paths.py:84`** — the four cases don't cover single-directory, trailing-slash, empty, consecutive-slash, a `lastIndexOf`-confusing literal, or the root-slash case from #1. **Fix:** add them behaviorally, asserting exact filename/head/tail text + the original data attributes + inert hostile markup + click values (especially `/` and `/file.txt`).

Per-area: XSS PASS, click-through PASS, other workspace rendering PASS — but root-slash display FAIL (#1), i18n fallback FAIL (#2), test coverage FAIL (#3). 42 targeted tests pass (incl. the headed-Chromium render + ESLint + classic-script collision guards). No XSS or click-through regression, and ordinary filenames are never dropped — this is just the root-slash edge + the locale hygiene + test gaps.

Once those land I'll re-gate; the visual/UX sign-off (the actual screenshot approval for the Artifacts-tab layout) is a separate maintainer step and will happen before ship. Self-rebased clean on current master. Nice fix for the deep-path filename-clipping problem.

rodboev:
Thanks, I tightened the three points you called out.

1. `static/workspace.js:603` now renders the directory row when either split side is non-empty, so `/file.txt` keeps the root slash instead of dropping the directory line.
2. `static/i18n.js:467` now owns `workspace_artifact_source_session` in `en` only. The 14 non-English copies are removed, and the existing `_locale[key] ?? LOCALES.en[key]` path now owns the missing-source fallback.
3. `tests/test_issue6067_artifact_paths.py` now covers `dir/foo.py`, `/file.txt`, `dir/`, `''`, `src//users.py`, `src/lastIndexOf/users.py`, and a workspace-prefixed hostile Unicode path. The same harness checks exact filename/head/tail text, the original `data-artifact-path` click-through, the fallback `data-i18n` node, and still proves the clipped current-master case against the fixed branch behavior.

nesquena-hermes (changes requested):

**Contributor SHA:** `1ff90ac9dc0234e161d785bb803e76f91b1db6ba`

The requested filename behavior is fixed and is clearly better than master: a 111-character filename is ellipsized on master but fully readable on desktop and mobile here. One adjacent directory-layout regression remains.

### Blocking finding

- **SILENT — an unusually long immediate-parent directory expands the Artifacts drawer horizontally** (`static/style.css:6437`). `.workspace-artifact-directory-tail` is `flex: 0 0 auto` with no width/overflow cap. On a real 390×844 page with a persisted artifact under `2026-07-20-quarterly-revenue-analysis-with-regional-breakdown-and-confidence-intervals/summary.json`, the 299px Artifacts root expands to `scrollWidth=595px`; the 281px row expands to `scrollWidth=586px`. Content bleeds beyond the drawer rather than keeping the nearest parent inside the card.

  **Fix:** cap the tail while preserving its nearest-parent priority, for example:
  ```css
  .workspace-artifact-directory { min-width: 0; overflow: hidden; }
  .workspace-artifact-directory-tail {
    flex: 0 0 auto;
    min-width: 0;
    max-width: 100%;
    overflow: hidden;
    text-overflow: ellipsis;
  }
  ```
  Re-run the long-parent mobile case and confirm `root.scrollWidth === root.clientWidth` and `row.scrollWidth === row.clientWidth`.

### Evidence

- Focused artifact/locale suite: **64 passed**; the one Playwright case could not launch because the sandbox has no Chromium binary.
- Full exact integration: **13,453 passed**; 15 browser/sandbox baseline failures and 2 packaged-runtime setup errors.
- Real browser before/after, six persisted artifacts: master truncates the long filename after `…revenue-ana…`; this head renders it fully at desktop and mobile.
- Interaction: six rows, minimum tap target **54.3px**, long row opens the correct JSON preview.
- Seventh long-parent reproduction: `root 299→595px`, `row 281→586px`.
- Codex: **SHIP ONLY WITH FIXES**, reproduced the same overflow.
- Fable: **SHIP-WITH-UX-FIXES**, same overflow guard required.

Once the directory tail is bounded and the seventh-row mobile proof is clean, this should be ready for re-gate.

rodboev:
Thanks for catching the long-parent overflow. The filename split is working, but `static/style.css:6437` gives the nearest-parent tail unlimited intrinsic width, so one parent segment can still widen the artifact row and mobile drawer.

1. I'll keep the tail's nearest-parent priority, then bound it to the directory width and give an overlong tail its own ellipsis. The directory container will also clip horizontal overflow, so neither split fragment can expand the card.
2. I'll add the exact `2026-07-20-quarterly-revenue-analysis-with-regional-breakdown-and-confidence-intervals/summary.json` case to `tests/test_issue6067_artifact_paths.py` and exercise the production-shaped drawer at 390x844.
3. The regression will require both `#workspaceArtifacts` and the long-parent artifact row to keep equal scroll and client widths. It will also keep `summary.json` fully visible, prove the long parent ellipsises inside its own span, and retain the existing short-tail priority checks.

The renderer, split results, original-path click-through, source fallback, and locale ownership stay unchanged.

nesquena-hermes:

**Certified contributor head:** `sha:3278fd4fed1834e7e03579f385f7f25404de7326`  
**Current-master integration base:** `c414c39f1015a3e97626b82d9b65a2a4396081b6`  
**Integrated gate commit:** `ba5e33596cb00338e3b2c4e78fad86595e0b9bf5`

The prior long-parent overflow blocker is closed. This exact head is regression-clean and is visibly better than current master on the reported filename-identification problem.

### What I ran

- **Threat scan:** CLEAN, score 0.
- **Patch fidelity:** the contributor's cumulative patch and the current-master integration have the same stable patch ID (`5def39d1c398f0e706c73223e5b123680d46a9b0`).
- **Full authoritative serial suite:** **13,515 passed**, 10 skipped, 1 xfailed, 2 xpassed, 34 subtests passed in 610.08s; no failures.
- **Codex current-master regression gate:** **SAFE TO SHIP**. It reviewed all 14 changed files and ran 74 focused tests, including Chromium layout checks.
- **Fable UX gate:** **SHIP-UX**. Filename-first placement, parent-preserving path truncation, source pill, density, keyboard behavior, and desktop/mobile behavior are sound.
- **Syntax/static:** `git diff --check`, `node --check static/workspace.js`, `node --check static/i18n.js`, and Python compilation all passed.
- **Real-page browser drive:** same richly seeded session on current master and the integrated PR at 1600×1000 desktop and 390×844 mobile, plus real click-through into an existing JSON preview.

### Reproduced evidence

The `1ff90ac9dc02` gate failed because an unusually long immediate-parent directory expanded the 299px Artifacts root to 595px. At this head, on both desktop and mobile:

- Artifacts root: `clientWidth=scrollWidth=299`.
- Long-parent row: `clientWidth=scrollWidth=281`.
- Long parent: `scrollWidth=576 > clientWidth=261`, safely contained with computed `overflow:hidden` and `text-overflow:ellipsis`.
- All four filenames are fully visible, including `sequence-diagram-with-an-extremely-descriptive-name.svg`.
- Keyboard-focus/click wiring is preserved; clicking the long `summary.json` row switched to Files and opened the exact expected preview path.
- Hostile/root/trailing/consecutive-slash cases, original `data-artifact-path`, escaping, and English-owned i18n fallback are covered by the branch's behavioral tests.

Current master already contains the smaller filename-first #6161 implementation through release commit `c67fd2dd2` / #6282, so I checked this as an overlapping implementation rather than against the obsolete one-line renderer. It is still a concrete improvement: current master clips the long SVG filename (`scrollWidth=430`, `clientWidth=261`, `fullyVisible=false`) on desktop and mobile; this head wraps it fully while containing the directory. The added behavioral/locale coverage also closes lifecycle gaps in the smaller implementation.

**Gate-pass at this exact head.** Replace the stale `gate-fail` result and move it to Priority Queue T1. The remaining hold is product-process only: this changes the visible workspace Artifacts pane, so Nathan should review the already-captured before/after desktop+mobile screenshots before any merge. No contributor code fix remains.

nesquena-hermes:
Shipped in **exp-v0.52.135** (via #6386). Workspace Artifacts pane now shows long filenames fully (filename-first, parent-preserving truncation) and caps the directory tail (`min-width:0`+`overflow:hidden`+`text-overflow:ellipsis`) so a long parent can't blow out the drawer. Independently verified at runtime (master lacked the tail overflow rule; this head applies it, 576px tail clipped to the drawer) + gate GREEN (Codex SAFE + Fable SHIP-UX + suite). Nathan approved ship. Thanks @rodboev! 🎯

PR #6083, opened 2026-07-15
fix(webui): stop password managers autofilling the search box, and keep unsaved sessions startable

ai-ag2026:
## Summary

Fixes a chain of two independent defects whose combined symptom is **new conversations cannot be started**: every `POST /api/session/draft` and `POST /api/chat/start` 404s a few seconds after `POST /api/session/new` returned 200.

### 1. The search inputs get parsed as username fields

`static/index.html` contains **no `<form>` at all**, but it does contain password inputs (the Settings → *Access Password* panel). Chromium therefore groups the document's unowned fields into one synthetic password form and picks the first text input in the DOM — `#sessionSearch`, the sidebar conversation filter — as that form's username field, and autofills the saved account name into it on load. `autocomplete="off"` was already present on it; it is ignored for credential heuristics.

The autofill then fires `oninput` → `filterSessions()` → `GET /api/sessions/search?content=1&depth=5` — a full content search nobody asked for, on every page load.

**Fix:** the four search inputs get `type="search"` plus per-manager opt-outs (`data-1p-ignore`, `data-lpignore`, `data-bwignore`, `data-form-type`), and the Settings password inputs get a real `<form>` owner with proper `autocomplete` tokens, so Chromium scopes the credential form to *them* rather than to the whole document. The form uses `display:contents`, so layout is unchanged. CSS suppresses WebKit's native search clear button because these fields ship their own (`#sessionSearchClear`).

### 2. A never-persisted session is treated as evictable

`new_session()` intentionally keeps a session in RAM until its first message (#1171), so the `SESSIONS` cache is its **only** copy. `_session_is_evictable()` (#4765) short-circuited on zero messages, reasoning that an empty shell *"is recreated on next access"*. It is not — `get_session()` has no recreate path and raises `KeyError`, so both routes 404 and the session can never be started.

The content search from (1) pulls every hit through `get_session()`, which blows past `sessions_cache_max` (default 300) and evicts the session the user is composing. This reproduced within seconds on an install with ~1700 sessions, and got worse as the session count grew.

**Fix:** require proof of persistence before evicting, for empty sessions too — which is what the function's own docstring already promised (*"Its full state is already persisted to the JSON sidecar"*). The zero-message branch was the one path that bypassed it.

## Why together

(1) fixes what triggers the cache churn; (2) fixes what turns that churn into an unstartable session. Either one alone leaves a real defect in place, so they are submitted as one change.

## Trade-off (flagging explicitly)

An unsaved empty session now stays resident for the process lifetime — one small empty object per *New Conversation* click. This is a deliberate application of the predicate's stated invariant (*"slightly more RAM for a session we are unsure about is strictly better than evicting an unsaved session"*), and #4765's memory bound is unaffected: sessions loaded from disk remain evictable. If you'd rather bound it by age (fresh = pinned, abandoned after N hours = evictable), happy to rework.

## Validation

```
pytest tests/test_issue4765_sessions_lru_eviction.py   ->  9 passed
```

- Existing 8 tests from #4765 pass unchanged.
- Added `test_unsaved_new_session_survives_churn_and_stays_startable`, which **fails on the unpatched predicate** (verified by reverting `api/models.py` alone) and passes with the fix.
- HTML structure verified well-formed; checked that no `.settings-field > input` / `> *` selectors exist, so the added `<form>` breaks no existing direct-child CSS rules (`#settingsPasswordEnvLock` is deliberately left outside it).
- Password fields are only ever accessed by id in JS (`$('settingsPassword')`), no parent traversal, so the new form owner is inert to it. `onsubmit="return false"` keeps Enter from submitting, since the panel saves via JS.

## Note

The autofill behaviour itself depends on undocumented, vendor-specific heuristics, so the `type="search"` + form-ownership combination is based on Chromium's own published guidance (group related fields in a single `<form>`; use `autocomplete` tokens) rather than on any documented rule about search fields. If you know of an install where a manager still fills these, this repo already uses the `readonly` + `onfocus` trick on `#clarifyInput` as a harder fallback.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -4056,6 +4056,14 @@ def _session_is_evictable(s) -> bool:
         on-disk ``message_count`` being at least the in-memory message count.
         A metadata-only stub is inherently backed by disk, so it is evictable.
 
+    The persistence requirement holds even for a session with ZERO messages.
+    ``new_session()`` deliberately does not touch disk until the first message
+    (#1171), so between "New Conversation" and the first send, this cache is the
+    session's ONLY copy. Evicting it does not discard a recreatable empty shell:
+    ``get_session()`` has no recreate path and raises ``KeyError``, so the very
+    next ``/api/session/draft`` or ``/api/chat/start`` 404s and the session can
+    never be started. Sessions the user is still composing must stay resident.
+
     Anything we cannot positively prove is safe stays resident. Using slightly
     more RAM for a session we are unsure about is strictly better than evicting
     an active or unsaved session (task safety invariant: a half-done memory fix
@@ -4077,14 +4085,12 @@ def _session_is_evictable(s) -> bool:
     if getattr(s, '_loaded_metadata_only', False):
         return True
     in_memory_count = len(getattr(s, 'messages', None) or [])
-    if in_memory_count == 0:
-        # A zero-message session has nothing to lose. If it was never persisted
-        # (brand new, no sidecar) dropping it only discards an empty shell; the
-        # next access recreates it. If it is persisted, it is trivially clean.
-        return True
     disk_count = _persisted_message_count(sid)
     if disk_count is None:
         return False  # cannot prove it is on disk → keep it resident
+    if in_memory_count == 0:
+        # Persisted and empty → trivially clean, nothing to lose.
+        return True
     return disk_count >= in_memory_count
 
 
--- a/static/index.html
+++ b/static/index.html
@@ -200,7 +200,7 @@
           </button>
         </div>
       </div>
-      <div class="session-search sidebar-search"><div class="session-search-field"><svg class="sidebar-search-icon" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true"><circle cx="11" cy="11" r="8"/><path d="M21 21l-4.35-4.35"/></svg><input id="sessionSearch" placeholder="Filter conversations..." data-i18n-placeholder="filter_conversations" oninput="filterSessions()" autocomplete="off"><button type="button" id="sessionSearchClear" class="session-search-clear has-tooltip has-tooltip--bottom-right" aria-label="Clear conversation filter" data-tooltip="Clear conversation filter" onclick="clearSessionSearch()" hidden><svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.4" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true"><line x1="18" y1="6" x2="6" y2="18"/><line x1="6" y1="6" x2="18" y2="18"/></svg></button></div></div>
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
Read both halves at this PR's HEAD: the `_session_is_evictable` change in `api/models.py:4085`, and the form-ownership changes in `static/index.html`, `static/onboarding.js`, `static/style.css`. Both are well-reasoned; the eviction fix in particular closes a real hole.

## The eviction fix is correct

On master the zero-message branch short-circuited *before* the persistence check:

```python
in_memory_count = len(getattr(s, 'messages', None) or [])
if in_memory_count == 0:
    return True   # "recreated on next access" — but it isn't
disk_count = _persisted_message_count(sid)
```

I confirmed the two claims the fix rests on:

1. **`new_session()` does not persist until the first message** — its docstring at `api/models.py` is explicit ("no disk write happens until the first message is appended (#1171)... if the process exits between session creation and first message, the session is lost"). So for a freshly created session the `SESSIONS` entry is the only copy.
2. **There is no recreate path.** `_persisted_message_count` (`api/models.py`) returns `None` when the sidecar file doesn't exist, and `get_session` raises rather than reconstructing. So evicting an unsaved shell really does make `/api/session/draft` and `/api/chat/start` 404, exactly as described.

Reordering so the `disk_count is None → return False` guard runs first, and only returning `True` for zero messages once persistence is proven, is the right shape and matches the docstring's own invariant.

## Trade-off: bound the resident-shell growth?

You already flagged this, and I want to reinforce it. With the fix, every "New Conversation" click that never gets a first message leaves a permanently non-evictable object in `SESSIONS`. `_evict_sessions_over_cap` (`api/models.py:4091`) skips anything `_session_is_evictable` rejects, so these shells sit above the cap for the process lifetime. In normal use that's one tiny empty object per abandoned click — negligible. But a client that hammers `/api/session/new` (or a flaky tab that recreates on every reconnect) could accumulate them unbounded, and the cap can no longer reclaim them.

There's already a prune path for the on-disk analogue — `_prune_orphaned_webui_zero_message_sessions` (referenced at `api/models.py:1442,5586`). An age-based fallback for the *in-memory* shells (pinned while fresh, evictable after N hours idle, matching your "abandoned after N hours" suggestion) would restore the memory bound without reintroducing the unstartable-session bug. Not a blocker for this PR, but worth a follow-up issue so the invariant doesn't silently regress.

## Frontend half

The `display:contents` form-owner technique is the correct fix for Chromium's synthetic-credential-form heuristic, and wrapping only the password inputs (leaving `#settingsPasswordEnvLock` outside the `<form>`) is the right scoping. `type="search"` + the per-manager opt-out attributes (`data-1p-ignore`, `data-lpignore`, `data-bwignore`, `data-form-type="other"`) is belt-and-suspenders, and suppressing `::-webkit-search-cancel-button` is necessary since these fields ship their own clear button. `onsubmit="return false"` correctly neutralizes Enter given the panels save via JS.

One small nit on `static/onboarding.js:200` — the api-key input keeps `autocomplete="off"` while the two password fields use the semantically stronger `new-password`. Since it's inside the same `display:contents` form now, `autocomplete="new-password"` there too would be marginally more robust against managers that ignore `off` on password-type inputs. Minor.

## Verification

CI green across all matrix jobs + browser-smoke + lint. The new `test_unsaved_new_session_survives_churn_and_stays_startable` asserts the exact failing path (survives 40-session churn, then `get_session(sid)` succeeds), and the author reports it fails on the unpatched predicate. Solid.

nesquena-hermes:
Shipped in **exp-v0.52.74** 🚀 Thanks @ai-ag2026.

Merged via release PR #6096 (the source PR shows as closed by the release merge). Your two-part fix landed as-is (the `type="search"` + per-manager opt-out attrs on the search/filter inputs, the `<form>` owners on the Settings + onboarding password fields, and the `_session_is_evictable()` persistence-before-eviction reorder).

One maintainer follow-up was added on top after the release gate: the Codex adversarial gate found that keeping *every* never-persisted empty shell resident forever let abandoned `/api/session/new` tabs accumulate unbounded past `sessions_cache_max` (reproduced cap=5 → 100 shells resident, 0 evicted — a slow leak). Fixed at root with a grace bound: a never-persisted empty shell stays resident only while it's **fresh** (<30 min) OR carries a **composer draft**; a stale, empty, draftless, never-saved shell becomes evictable again. Grace-eviction only fires when the sidecar is *confirmed absent*. Your brick fix is fully preserved (a real compose persists a draft on the first keystroke → disk-backed + reloadable), and the leak is closed. +2 regression tests.

Full gate: 13,108 tests pass, Codex re-gate clean (leak closed, brick preserved, concurrent-safe), browser-smoke zero console errors.

PR #6084, opened 2026-07-15
fix(webui): keep the content search from evicting the user's working set

ai-ag2026:
## Summary

`/api/sessions/search?content=1` walks **every** session and pulls each one through `get_session()` — which inserts it into the `SESSIONS` LRU and marks it recently-used.

On any install with more sessions than `sessions_cache_max` (default 300), one search flushes the entire cache and refills it with sessions the user isn't looking at. The sessions actually open in the UI are precisely the ones evicted. And since the search is keystroke-debounced, it repeats while typing.

This is the classic buffer-pool scan-pollution problem: a full scan should not be allowed to evict the working set.

## Fix

A scan reads each session exactly once, so nothing it touches has earned "recently used". `get_session_for_scan()`:

- reuses a resident session **without promoting it** in the LRU,
- reads a cold one straight from disk **without caching it**,
- returns `None` instead of raising — a scan skips what it cannot open.

## Not a latency fix

Worth being explicit, since the symptom that led me here looked like slowness: those multi-second searches were **contention**, not scan cost. A trivial `/api/profiles` took 9.2s in the same window, and a full read+parse of ~1700 real sessions measures **~4s** total. This PR changes cache *correctness*, not speed.

## Validation

```
pytest tests/test_issue4765_sessions_lru_eviction.py \
       tests/test_sessions_search_depth_validation.py    ->  13 passed
```

- Two added tests: a scan must not evict the working set, and a scan hit must not reorder the LRU. Both **fail on the pre-fix accessor** (verified by revert) — the working set drops from 4/4 to 0/4 resident after a single scan.
- Broad sweep (`-k "session or search or cache or evict or onboarding or model"`): 3910 passed, 1 pre-existing failure (`test_webui_state_db_reconciliation`, fails identically without this change).

## Heads-up on the test change

`test_sessions_search_depth_validation` patched `api.routes.get_session`. Since the search now reads through the scan accessor, it patches `get_session_for_scan`. Left unfixed, two of its cases fail — and the third (`test_search_valid_depth_still_caps_scan`) passes *vacuously*, because it asserts `count == 0` and an empty result set satisfies that for the wrong reason.

## Relation to #6083

Independent of #6083 and reviewable/mergeable on its own — but the two came out of the same investigation, and both touch `api/models.py` and this test file. Whichever lands second will need a trivial rebase (both append tests to the end of the same file). Happy to rebase on request.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -4404,17 +4404,33 @@ def _evict_sessions_over_cap(cap: int | None = None) -> int:
     return evicted
 
 
-def get_session(sid, metadata_only=False):
-    """Load a session, optionally with metadata only (skipping the messages array).
+def get_session_for_scan(sid):
+    """Read a session for a one-pass scan without disturbing the LRU.
 
-    Metadata-only loads intentionally do not populate the full-session cache.
-    Otherwise a later full load could return a compact object with an empty
-    messages list. Use this when you only need compact() metadata and not the
-    actual message history (e.g., for fast sidebar switching).
+    The scan uses the full canonical resolver so stale sidecars, pending journal
+    recovery, and newer state.db transcript rows remain searchable. It only
+    differs from ``get_session`` in cache policy: no hit promotion, no cold-load
+    insertion, and therefore no scan-triggered eviction.
+    """
+    try:
+        return _resolve_session(sid, promote_cache=False, cache_on_miss=False)
+    except Exception:
+        logger.debug("scan load failed for session %s", sid, exc_info=True)
+        return None
+
+
+def _resolve_session(sid, metadata_only=False, *, promote_cache=True, cache_on_miss=True):
+    """Resolve a session through the canonical freshness/recovery path.
+
+    ``get_session_for_scan`` shares this resolver with normal reads so that
+    content search sees the same disk freshness, journal-retry, and state.db
+    recovery behavior. Its policy merely suppresses LRU promotion and cold-load
+    caching; reconciliation still updates an already-resident entry when normal
+    resolution replaces stale contents.
     """
     with LOCK:
         cached = SESSIONS.get(sid)
-        if cached is not None:
+        if cached is not None and promote_cache:
             SESSIONS.move_to_end(sid)  # LRU: mark as recently used
     if cached is not None:
         # Defensive cache ownership check: compression/continuation and recovery
@@ -4438,12 +4454,12 @@ def get_session(sid, metadata_only=False):
                 disk_session = Session.load(sid)
                 with LOCK:
                     SESSIONS[sid] = disk_session
-                    SESSIONS.move_to_end(sid)
+                    if promote_cache:
+                        SESSIONS.move_to_end(sid)
                 cached = disk_session
             except Exception:
                 logger.debug(
-                    "cached session disk-freshness check failed for session %s",
-                    sid, exc_info=True,
+                    "cached session disk-freshness check failed for session %s", sid, exc_info=True,
                 )
         if not metadata_only and _inactive_cache_tail_needs_disk_check(cached):
             try:
@@ -4451,28 +4467,26 @@ def get_session(sid, metadata_only=False):
                 if _cache_has_stale_unsaved_user_tail(cached, disk_session):
                     with LOCK:
                         SESSIONS[sid] = disk_session
-                        SESSIONS.move_to_end(sid)
+                        if promote_cache:
+                            SESSIONS.move_to_end(sid)
                     cached = disk_session
             except Exception:
                 logger.debug(
-                    "stale cached user-tail check failed for session %s",
-                    sid, exc_info=True,
+                    "stale cached user-tail check failed for session %s", sid, exc_info=True,
                 )
         if not metadata_only and _session_has_pending_journal_retry(cached):
             try:
                 _try_retry_journal_recovery_in_place(cached)
             except Exception:
                 logger.debug(
-                    "lazy journal-retry failed on cache hit for session %s",
-                    sid, exc_info=True,
+                    "lazy journal-retry failed on cache hit for session %s", sid, exc_info=True,
                 )
         if not metadata_only:
             try:
                 _sync_sidecar_from_state_db_if_newer(cached)
             except Exception:
                 logger.debug(
-                    "state.db newer-sidecar sync failed on cache hit for session %s",
-                    sid, exc_info=True,
+                    "state.db newer-sidecar sync failed on cache hit for session %s", sid, exc_info=True,
                 )
         return cached
     if metadata_only:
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
Read `api/models.py:4137` (`get_session_for_scan`) at this PR's HEAD against `get_session` on `origin/master`, plus the route swap at `api/routes.py:16589`. The change is correct and surgical.

## What the accessor does right

The scan path deliberately drops every LRU mutation `get_session` performs:

```python
with LOCK:
    cached = SESSIONS.get(sid)  # NB: no move_to_end() — a scan never promotes
if cached is not None and str(getattr(cached, 'session_id', '') or '') == str(sid):
    if not getattr(cached, '_loaded_metadata_only', False):
        return cached
try:
    return Session.load(sid)
except Exception:
    ...
    return None
```

Two things I verified against master that make this safe:

1. **The metadata-only guard is load-bearing.** `Session.load_metadata_only` sets `_loaded_metadata_only = True` (`api/models.py:1562,1586`) and those objects carry an empty `messages` list. If the scan returned a metadata-only cache hit, the content search would silently match zero messages for any session the sidebar had loaded compactly. Falling through to `Session.load(sid)` for that case is the right call.

2. **`Session.load` genuinely doesn't touch `SESSIONS`.** I checked `load` at `api/models.py:1456` — it constructs and returns without inserting into the global dict, so a cold scan hit really is LRU-transparent. That's the whole point of the fix, and it holds.

## One intentional divergence worth noting

On an ownership mismatch, `get_session` *evicts* the mis-filed entry (`if SESSIONS.get(sid) is cached: SESSIONS.pop(...)`). The scan accessor just declines to use it and falls through to disk, leaving the cache untouched. That's the correct choice for a read-only scan — a search should not mutate cache state, and `get_session` will clean up the stale entry on the next real open. No bug here, but it's a deliberate behavioral difference from the function it mirrors, so it's worth a one-line comment near the ownership check if you want the next reader to not "fix" it back to matching `get_session`.

## Relation to #6083

These two are complementary, not redundant. This PR removes the cache-pollution *mechanism* (search no longer promotes/inserts), while #6083's `_session_is_evictable` change is defense-in-depth against churn from any source plus the autofill trigger that kicks off the phantom search. Both touch `tests/test_issue4765_sessions_lru_eviction.py` and append to the end, so whichever lands second needs a trivial rebase — as you already flagged.

## Verification

CI is green across all 15 matrix jobs + browser-smoke + lint. The two new tests (`test_content_search_scan_does_not_evict_the_working_set`, `test_scan_accessor_reuses_resident_sessions_without_promoting`) assert exactly the two invariants that matter — working-set survival and no LRU reordering on a hit — and the author reports both fail on the pre-fix accessor. Looks ready.

nesquena-hermes (changes requested):
Thanks @ai-ag2026 — the scan-pollution diagnosis is exactly right and the LRU-transparency approach (no promote on hit, no insert/evict on miss) is the correct shape. The release gate (Codex adversarial review + full suite) reproduced a **search-correctness regression** in `get_session_for_scan()` though, so I'm kicking it back with a precise spec rather than shipping it.

## The blocker: the scan accessor skips the freshness/recovery that `get_session()` does

`get_session_for_scan()` on a cache **miss** calls `Session.load(sid)` directly, and on a **hit** returns the cached object as-is. Both paths skip the reconciliation `get_session()` performs, specifically:

- `_cached_session_lags_disk(cached)` → reload if the sidecar is newer than the cached object
- `_inactive_cache_tail_needs_disk_check` / `_cache_has_stale_unsaved_user_tail` → stale-tail refresh
- `_session_has_pending_journal_retry` → `_try_retry_journal_recovery_in_place`
- `_sync_sidecar_from_state_db_if_newer(...)` → **the important one for search**: pull newer content from `state.db` into the session before reading it

Because the content scan reads the session bodies to match against, skipping `_sync_sidecar_from_state_db_if_newer` means a search can **miss recently-written messages** that live in `state.db` but haven't been reconciled into the sidecar yet.

**Codex reproduced it** with the UI's exact query: `/api/sessions/search?content=1&depth=5` returned `count=0` through the scan accessor for content that `get_session()` returns `count=1` for; a cold-cache `state.db` recovery probe likewise found the recovered text through `get_session()` but not through the scan accessor.

## What's already verified good (keep it)

- No LRU promotion on a hit, no insert/evict on a miss → the working set is genuinely protected (the core win).
- Metadata-only cached entries correctly fall through to a full load.
- Mismatched-ownership cached objects are not returned.
- Depth clamp: negative → 0, non-numeric → 5. Safe.

## Fix-spec

Share `get_session()`'s full-resolution + reconciliation logic, parameterized on cache policy, so a scan gets the SAME freshness/recovery but **suppresses only**: (a) LRU promotion (`move_to_end`) on a hit, (b) insertion into `SESSIONS` on a miss, and (c) `_evict_sessions_over_cap()`. Concretely, either add a `scan=True` / `promote=False, cache=False` path into `get_session()` (preferred — one code path, no drift), or have `get_session_for_scan()` run the same `_cached_session_lags_disk` / stale-tail / journal-retry / `_sync_sidecar_from_state_db_if_newer` steps on the object it's about to return.

Add a regression test asserting that **newer sidecar/state.db content is searchable via the scan accessor** without changing `len(SESSIONS)` or its order (the existing no-promote/no-pollute tests are good — keep them, add this freshness case alongside).

Happy to take this over and build the shared-path version myself if you'd rather — just say the word.

ai-ag2026:
Re-gate evidence for current head `7ca84821713133f0ad279d19fb86c5af16bc66a7` (base `735ba1c4`):

- Independent final review: **ACCEPT**.
- The new shared resolver preserves normal `get_session()` freshness, journal-retry, state.db reconciliation, hit promotion, cold insertion, and eviction defaults. `get_session_for_scan()` changes only cache policy: no hit promotion and no cold insert/evict.
- The real search-route regression proves state.db-only recovery text is searchable while exact LRU order and cache size remain unchanged.
- Final direct suite: LRU, depth validation, sidecar reconciliation, WebUI state.db reconciliation, and sidecar repair → **146 passed**; syntax, Ruff diff, and diff hygiene clean.
- Current GitHub CI is **18/18 green**.

The visible `CHANGES_REQUESTED` state predates this exact head; requesting re-evaluation against the current evidence.

ai-ag2026:
@nesquena-hermes re-review requested. The 15.07 blocker is resolved at head `a9103cc246c1` exactly per your preferred fix-spec: `get_session()` and `get_session_for_scan()` now share one `_resolve_session()` path (full freshness/recovery — `_cached_session_lags_disk`, stale-tail, journal retry, `_sync_sidecar_from_state_db_if_newer` — unconditional; only LRU promote / insert / evict gated behind cache-policy params). Regression test `test_content_search_scan_recovers_newer_state_db_without_lru_churn` proves newer state.db content is searchable via the scan accessor with `len(SESSIONS)` and order unchanged. CI 18/18 green. Note: an earlier comment on this PR cited an incorrect head SHA — `a9103cc246c1` is the actual head; findings above were verified against it directly.

nesquena-hermes:
Shipped in **exp-v0.52.120** (experimental channel) via release #6343. Thanks @ai-ag2026! 🎉

Content search no longer evicts your open sessions from the in-memory LRU cache. The scan path now uses a dedicated `get_session_for_scan()` accessor that keeps the same disk-freshness, journal-recovery, and state.db reconciliation (so recent content stays searchable — the 2026-07-15 review point) but suppresses MRU promotion + cold-load insertion/eviction, so scanning many sessions doesn't disturb your working set.

Gate: Codex `SAFE TO SHIP` (verified normal `get_session()` reads are behavior-identical + #4765 eviction preserved), full suite 13458 passed / 0 failed, ruff clean.

Issue #4912, opened 2026-06-25
Bug: async_delegation completion events silently dropped — subagent results never return to chat


mydelren:
## Description

When using `delegate_task` (background subagent delegation), the subagent completes successfully but its result **never re-enters the parent conversation** in the WebUI. The user sees the subagent as perpetually "running" even though it has finished.

## Root Cause

In `api/background_process.py`, the `format_wakeup_prompt()` function handles event types `completion`, `watch_match`, `watch_disabled`, and `watch_overflow_*`. However, it does **not** handle the `async_delegation` event type that `delegate_task` completion produces:

```python
# api/background_process.py line 387-388
if evt_type != "completion":
    return None   # ← async_delegation events are silently dropped here
```

The full call chain:

1. ✅ Subagent completes → `_push_completion_event()` → `completion_queue.put(evt)` with `type="async_delegation"`
2. ✅ WebUI drain thread reads the event → `_process_one(evt)`  
3. ✅ Resolves `session_key` → finds `session_id`
4. ❌ `format_wakeup_prompt(evt)` returns `None` — **breakage point**
5. ❌ `wakeup_prompt` is empty → server-side wakeup turn is never started → result is lost

The agent-side `tools.process_registry.format_process_notification()` correctly handles `async_delegation` events, but the WebUI's `format_wakeup_prompt()` does not delegate to it for unrecognized types.

## Steps to Reproduce

1. Open a chat session in the WebUI
2. Ask the agent to run a simple background delegation (e.g. `delegate_task` with a trivial terminal command)
3. The tool returns `"status": "dispatched"` — subagent starts running
4. Wait for the subagent to complete
5. **Expected**: the consolidated result appears in the chat as a new message  
6. **Actual**: nothing happens; the result is silently lost

## Suggested Fix

In `format_wakeup_prompt()`, fall back to the agent-side formatter for unrecognized event types:

```python
if evt_type != "completion":
    try:
        from tools.process_registry import (
            format_process_notification as _agent_fmt,
        )
        result = _agent_fmt(evt)
        if result:
            return result
    except Exception:
        logger.debug(
            "agent-side format_process_notification fallback failed for "
            "evt_type=%s",
            evt_type,
            exc_info=True,
        )
    return None
```

This approach:
- Handles `async_delegation` events immediately
- Automatically picks up any future event types the agent adds
- Preserves the existing `return None` as the final fallback
- Uses a safe try/except so a broken agent-side import never crashes the drain loop

## Environment

[...]

Issue #6008, opened 2026-07-13
Follow-up to #5269: approval card is displayed, but approval_id is empty and approval cannot be submitted


SirSigfrid:
Hello,

**Summary**

This is a follow-up to #5269.

The original problem in #5269 was that no approval dialog or approval card was displayed at all. That part now appears to be fixed: the WebUI displays the approval request.

The remaining problem is that the displayed approval request has an empty approval_id. When the user tries to approve or deny it, the WebUI cannot submit the response. The approval request is visible, but it is not actionable.

<img width="952" height="704" alt="Image" src="https://github.com/user-attachments/assets/e589020d-6b83-4ec0-99d2-febe8b58de3b" />

**Environment**

    Infrastructure: Oracle VPS Free Tier
    OS: Ubuntu 24.04 ARM
    Host resources: 4 vCPU, 24 GB RAM
    Container runtime: Podman, rootless
    Deployment: separate hermes-agent and hermes-webui containers
    Hermes WebUI: v0.52.0
    Hermes Agent: v0.18.2
    WebUI chat backend: gateway
    Gateway status in WebUI: Running
    Runs API opt-in: HERMES_WEBUI_GATEWAY_USE_RUNS_API=true

Relevant non-secret configuration:

```
HERMES_WEBUI_CHAT_BACKEND=gateway
HERMES_WEBUI_GATEWAY_BASE_URL=http://<hermes-agent-host>:8642
HERMES_WEBUI_GATEWAY_USE_RUNS_API=true
```

The API server key and the WebUI gateway API key are supplied through environment files. Their values are not included here.

The same Hermes Agent can also be used through the separate Hermes Dashboard. The approval problem described here has not been reproduced through that Dashboard path.

**Steps to reproduce**

    Run Hermes Agent with its API server enabled.
    Connect Hermes WebUI to the remote gateway/API server.
    Use the gateway-backed WebUI chat path with the Runs API enabled.
    Ask the agent to perform an operation that requires approval, such as a guarded terminal or file operation.
    Wait for the approval request.
    Observe that the approval card or request is displayed.
    Click an approval choice such as Allow once or Deny.

**Expected behavior**

The WebUI should assign or receive a non-empty identifier for the approval request and submit the selected choice to the Hermes Agent Runs API.

The approval should be resolved through:

`POST /v1/runs/{run_id}/approval`

The run should then continue for an allowed choice or stop cleanly for a denied choice.

**Actual behavior**

[...]

PR #6087, opened 2026-07-15
fix(gateway_chat): use parseable session key format for X-Hermes-Session-Key

kaishi00:
## Summary

The WebUI gateway bridge (`gateway_chat.py`) sends `X-Hermes-Session-Key` as `f"webui:{session_id}"` at two call sites (lines 374 and 838). But the gateway's `_parse_session_key()` (`gateway/run.py:2164`) requires the format `agent:main:{platform}:{chat_type}:{chat_id}` — 5+ colon-separated parts starting with `agent:main`.

With the `webui:` prefix, `_parse_session_key()` returns `None`, which breaks async delegation result routing back to WebUI sessions.

## Root Cause Trace

When a background delegation completes while a WebUI gateway session is active, the result must route back through the gateway's notification injection path:

1. `_async_delegation_watcher()` peeks the completion queue for `async_delegation` events
2. `_enrich_async_delegation_routing(evt)` calls `_parse_session_key(evt.get("session_key"))` to fill `platform`/`chat_type`/`chat_id` on the event
3. `_build_process_event_source(evt)` uses those fields to resolve the target session via `session_store._entries` lookup, then `_parse_session_key()` as fallback
4. `_inject_watch_notification()` finds the matching adapter and injects the result

With `webui:{session_id}` as the session key:

- **Step 2:** `_parse_session_key("webui:abc123")` → `parts = ["webui", "abc123"]`, `len(parts) = 2 < 5` → returns `None` — enrichment is skipped, event has no `platform`/`chat_type`/`chat_id`
- **Step 3:** `session_store._entries.get("webui:abc123")` → `None` (api_server sessions are registered under their canonical key, and the api_server platform never calls `register_session()` — the session only exists in the store when a key in the correct format is used). `_get_cached_session_source()` → `None`. `_parse_session_key()` → `None`. All derived fields are empty.
- **Step 4:** `_build_process_event_source()` returns `None` → `_inject_watch_notification()` logs `"Dropping watch notification with no routing metadata"` and returns

**The result is silently dropped.** This is the exact symptom in #4691.

## The Fix

Change both call sites from:
```python
headers["X-Hermes-Session-Key"] = f"webui:{session_id}"
```
to:
```python
headers["X-Hermes-Session-Key"] = f"agent:main:api_server:dm:{session_id}"
```

With the corrected format:
- `_parse_session_key()` parses successfully → `{"platform": "api_server", "chat_type": "dm", "chat_id": "<session_id>"}`
- `session_store._entries` contains the key (confirmed: live session store shows `agent:main:api_server:dm:fda0f4178fab` registered with proper origin metadata)
- `_build_process_event_source()` resolves to a valid `SessionSource` with `platform=api_server`
- The `api_server` adapter is in `self.adapters` (registered during gateway startup)
- Result is injected via `adapter.handle_message(synth_event)`

## Why This Is Still Needed After PR #4799 and Core `d0e9a42ce`

PR #4799 was closed today as superseded by the durable delivery model in `d0e9a42ce` ("harden durable completion delivery"). However, the durable delivery changes address **tracking and dedup** — `claim_event_delivery()` / `complete_event_delivery()` / persistent completion records. They do NOT change the routing path.

The routing for async delegation completions still flows through:
```
_async_delegation_watcher → _enrich_async_delegation_routing → _parse_session_key → _build_process_event_source → _inject_watch_notification
```

`_parse_session_key()` is the sole mechanism for extracting routing fields from a session key. If the key format is wrong, routing fails regardless of how robust the delivery tracking is. The durable model ensures a completion is delivered **exactly once** — but only if it can be **delivered at all**, which requires a parseable session key.

## Scope

Exactly 2 lines changed (2 insertions, 2 deletions) across 2 call sites in `gateway_chat.py`:
- Line 374: `_run_gateway_runs_api_streaming()` (POST /v1/runs endpoint)
- Line 838: `_run_gateway_chat_streaming()` (POST /v1/chat/completions endpoint)

No new dependencies, no behavior change for in-process (non-gateway) WebUI mode, no impact on existing session continuity or Honcho memory scoping (the key is still unique per WebUI conversation).

## Related

- Fixes the routing root cause of #4691
- Complements #4912 (v0.51.652 — formatting fix for the display side)
- Independent of PR #4799 (closed) and core `d0e9a42ce` (durable delivery — addresses tracking, not routing)

### DIFF
--- a/api/gateway_chat.py
+++ b/api/gateway_chat.py
@@ -371,7 +371,7 @@ def _run_gateway_runs_api_streaming(
     }
     if api_key:
         headers["Authorization"] = f"Bearer {api_key}"
-        headers["X-Hermes-Session-Key"] = f"webui:{session_id}"
+        headers["X-Hermes-Session-Key"] = f"agent:main:api_server:dm:{session_id}"
     message_content: Any = str(msg_text or "")
     if attachments:
         try:
@@ -835,7 +835,7 @@ def _run_gateway_chat_streaming(
                 headers["Authorization"] = f"Bearer {api_key}"
                 # Scope Gateway long-term continuity to this WebUI conversation
                 # without exposing the browser's auth cookie or CSRF material.
-                headers["X-Hermes-Session-Key"] = f"webui:{session_id}"
+                headers["X-Hermes-Session-Key"] = f"agent:main:api_server:dm:{session_id}"
             message_content: Any = str(msg_text or "")
             if attachments:
                 try:
--- a/tests/test_webui_gateway_chat_backend.py
+++ b/tests/test_webui_gateway_chat_backend.py
@@ -328,7 +328,7 @@ def test_gateway_chat_worker_translates_sse_and_persists_session(tmp_path, monke
     assert captured["url"] == "http://gateway.local/v1/chat/completions"
     assert captured["headers"]["Authorization"] == "Bearer secret-token"
     assert captured["headers"]["X-hermes-session-id"] == s.session_id
-    assert captured["headers"]["X-hermes-session-key"] == f"webui:{s.session_id}"
+    assert captured["headers"]["X-hermes-session-key"] == f"agent:main:api_server:dm:{s.session_id}"
     assert '"stream": true' in captured["body"]
     payload = json.loads(captured["body"])
     assert payload["reasoning_effort"] == "high"


nesquena-hermes (changes requested):
## Changes requested — the existing backend test still asserts the old session-key format (now RED)

Thanks for this — the direction is **correct**. The old `X-Hermes-Session-Key: webui:{session_id}` value is not parseable by the gateway: `_parse_session_key()` (in `gateway/run.py`) only accepts keys shaped `agent:main:{platform}:{chat_type}:{chat_id}[:...]`, so `webui:{id}` (2 colon-parts, `parts[0] != "agent"`) returns `None`, and `_enrich_async_delegation_routing()` silently no-ops — async-delegation completion events never route back to the WebUI conversation. Your new value `agent:main:api_server:dm:{session_id}` matches the canonical format every other platform uses, so it fixes the routing.

**Blocker:** the change to `api/gateway_chat.py` was not accompanied by the matching test update, so the PR reddens its own existing test.

### Reproduced (isolated sandbox, PR head `3989ab07`)

```
tests/test_webui_gateway_chat_backend.py::test_gateway_chat_worker_translates_sse_and_persists_session FAILED

>       assert captured["headers"]["X-hermes-session-key"] == f"webui:{s.session_id}"
E       AssertionError: assert 'agent:main:api_server:dm:68c64763d1a5' == 'webui:68c64763d1a5'
E         - webui:68c64763d1a5
E         + agent:main:api_server:dm:68c64763d1a5

1 failed, 40 passed in 6.70s
```

The assertion at `tests/test_webui_gateway_chat_backend.py:331` still expects the old literal. On master both the runtime and the assertion read `webui:{...}` so it passes; this PR changed only the runtime, leaving the assertion stale → master-green, head-red (introduced by this PR).

### Precise fix-spec

1. **`tests/test_webui_gateway_chat_backend.py:331`** — update the assertion to match the new format:
   ```python
   assert captured["headers"]["X-hermes-session-key"] == f"agent:main:api_server:dm:{s.session_id}"
   ```
2. Consider adding a short assertion that the emitted key actually round-trips through the parser (i.e. `_parse_session_key(key)` is non-`None` and yields `platform="api_server"`, `chat_type="dm"`, `chat_id=session_id`) so a future format regression is caught directly — optional but it locks in the actual contract this PR is fixing.
3. Note there are **two** call sites in `api/gateway_chat.py` (line ~374 in `_run_gateway_runs_api_streaming` and line ~838 in `_run_gateway_chat_streaming`) — both are correctly updated; just make sure any test covering the runs-API path (if added) uses the same expected value.

Re-run `pytest tests/test_webui_gateway_chat_backend.py -q` locally after the test update — it should be all-green. Ping me when pushed and I'll re-gate.

kaishi00:
### CI failure is pre-existing (not caused by this PR)

The shard 4 failure across all Python versions is `test_issue1857_usage_overwrite.py::test_stream_completion_overwrites_session_usage_with_latest_turn`:

```
AttributeError: 'ToolEntry' object has no attribute 'dynamic_schema_overrides'
```

This is an agent API version mismatch in the test environment — `ToolEntry` no longer has the `dynamic_schema_overrides` attribute that `streaming.py:276` expects. I verified by reverting this PR's diff and running the same test on clean `origin/master` — identical failure:

```
git checkout origin/master -- api/gateway_chat.py
pytest tests/test_issue1857_usage_overwrite.py::test_stream_completion_overwrites_session_usage_with_latest_turn
# 1 failed — same AttributeError
```

This PR touches only `gateway_chat.py` (2 lines, session key format). The failing test doesn't import or test `gateway_chat` at all — it exercises `streaming.py`'s usage-token tracking path.

nesquena-hermes:
## Summary

I read the diff at `origin/master...HEAD` (only `api/gateway_chat.py`, both call sites at `374` and `838`), cross-checked the agent's session-key contract, and worked through the CI shard picture. The production fix is correct, and your "pre-existing failure" analysis is right about `test_issue1857` — but there's a **second** failure in shard 4 that this PR *does* cause and that isn't addressed yet.

## The fix is correct against the agent contract

The gateway's parser requires exactly the 5-part shape you switched to. From `gateway/run.py:2534-2544` (agent side):

```python
parts = session_key.split(":")
if len(parts) >= 5 and parts[0] == "agent" and parts[1] == "main":
    result = {
        "platform": parts[2],
        "chat_type": parts[3],
        "chat_id": parts[4],
    }
    ...
    return result
return None
```

`webui:{session_id}` splits to 2 parts → `None`, so `_enrich_async_delegation_routing` / `_build_process_event_source` get no routing fields and the completion is dropped, exactly as your root-cause trace describes. `agent:main:api_server:dm:{session_id}` parses to `platform=api_server, chat_type=dm, chat_id=<id>`. The layer is right — this belongs in the WebUI bridge, not the agent.

## The ToolEntry failure is environmental — agreed

`test_stream_completion_overwrites_session_usage_with_latest_turn` exercises `streaming.py`, and `streaming.py:276` reads `entry.dynamic_schema_overrides`. The current agent `ToolEntry` declares that in `__slots__` (`tools/registry.py:90-93`), so this only `AttributeError`s against an older agent snapshot in the CI image. Independent of your 2-line diff. Correct call.

## But shard 4 has a second failure this PR introduces

`tests/test_webui_gateway_chat_backend.py:331` still asserts the old header value:

```python
assert captured["headers"]["X-hermes-session-key"] == f"webui:{s.session_id}"
```

That assertion lives in `test_gateway_chat_worker_translates_sse_and_persists_session`, which drives the `/v1/chat/completions` path — i.e. the exact call site you changed at line `838`. After this PR the worker emits `agent:main:api_server:dm:{session_id}`, so the assertion fails wherever pytest-shard places it. Greptile flagged the same line. I confirmed the PR head does **not** touch this test (`git diff --stat` shows only `api/gateway_chat.py`, 2 insertions / 2 deletions).

## Recommendation

Update line 331 to match the new contract:

```python
assert captured["headers"]["X-hermes-session-key"] == f"agent:main:api_server:dm:{s.session_id}"
```

Worth grepping the rest of `tests/` for `webui:` header assertions in case the `/v1/runs` path (line 374) has a sibling test. Once line 331 is updated, shard 4 should be down to just the environmental `ToolEntry` error, which a CI agent-image bump resolves separately.

kaishi00:
Pushed the test fix — line 331 of `test_webui_gateway_chat_backend.py` now asserts `agent:main:api_server:dm:{session_id}` to match the new contract.

Verified locally:

```
tests/test_webui_gateway_chat_backend.py::test_gateway_chat_worker_translates_sse_and_persists_session PASSED (1.72s)
```

Also grepped the full `tests/` tree for other `webui:` session-key assertions — line 331 was the only one. No sibling test exists for the `/v1/runs` path (line 374).

nesquena-hermes:
Thanks @kaishi00 — you correctly diagnosed that `webui:{session_id}` is unparseable by the gateway's `_parse_session_key()` (needs `agent:main:{platform}:{chat_type}:{chat_id}`), and the new format does parse cleanly. But an adversarial re-gate surfaced a SILENT data-continuity regression that blocks shipping this as-is.

### [SILENT] Changing the session-key format changes the persistent memory/continuity identity

The `X-Hermes-Session-Key` isn't only a routing hint — on the gateway side it's the **authoritative key for persistent session state / memory scope**. In `gateway/run.py` the Honcho managers are explicitly *"keyed by gateway session key"* to preserve `write_frequency="session"` semantics across the short-lived per-message agents, and `get_or_create_session()` resolves the continuity scope from that identity.

Consequence of the swap:
- Existing WebUI conversations today resolve to a scope derived from `webui:{session_id}` (probe: `webui-{id}`).
- After this change the same conversation resolves to `agent:main:api_server:dm:{session_id}` (probe: `agent-main-api_server-dm-{id}`).
- So on upgrade, every already-established WebUI↔gateway conversation **silently loses access to its prior memory/continuity scope** — no error, no migration, the history just isn't found under the new key.

The PR's own tests can't catch this because they run on fresh state (they only assert the header string), so the namespace break is invisible to them — this is exactly the kind of cross-upgrade regression the full suite passes clean on.

Secondary note (not the blocker, but worth knowing): the PR summary motivates the change partly by restoring an async delegation flow, but the current hermes-agent already makes `api_server` detached delivery unsupported and runs `background=true` delegation synchronously (agent commit `7a131f7f4`, June 21), so that specific benefit no longer applies on the current agent.

### Fix (preserve the established identity)

Pick one, both preserve continuity:
1. **Preferred — keep `webui:{session_id}` on the wire and teach the parser the legacy form.** Leave both call sites emitting `webui:{session_id}` and extend `gateway/run.py::_parse_session_key()` to recognize a `webui:{id}` key and map it to `platform=api_server, chat_type=dm, chat_id={id}` (or whatever canonical routing you want). That fixes the parse failure #6008-style routing needs *without* moving the memory identity. Revert the test assertion at `tests/test_webui_gateway_chat_backend.py:331` back to `webui:{session_id}`.
2. **Or** — if you want the new canonical format, add an explicit memory-identity migration so existing `webui:{id}`-scoped state is found under (or aliased to) the new key. This is more work and higher risk; option 1 is the smaller, safer fix.

Note that fix option 1 touches `gateway/run.py`, which lives in the **hermes-agent** repo, not hermes-webui — so this may need a coordinated agent-side change. Happy to help scope that if you want to go that route. Re-request review once continuity is preserved.

_Gate: Codex adversarial correctness review + maintainer verification of the gateway session-key → memory-scope contract._

nesquena-hermes (changes requested):
## Gate: not shippable as-is — one confirmed regression + a premise to re-establish

Thanks @kaishi00 — the CI-red that blocked this earlier is resolved (the session-key assertion is updated and CI is 18/18 green), and the format you switched to (`agent:main:api_server:dm:{session_id}`) is indeed what `_parse_session_key()` expects. But the authoritative regression gate (Codex, traced against the installed agent) surfaced a **SILENT regression**, and separately the PR's justifying issue is already closed-as-resolved — so I can't ship it in its current shape:

### 1. SILENT — long-term memory scope re-keys for existing conversations (confirmed)
`X-Hermes-Session-Key` isn't only a routing key — the gateway server uses it verbatim to **scope long-term memory** (Honcho sessions), and the agent's resolver gives `gateway_session_key` highest precedence with **no legacy alias**. Switching the wire value from `webui:{session_id}` → `agent:main:api_server:dm:{session_id}` changes the memory identity of an **already-existing** authenticated gateway conversation on upgrade — its `session_id` does NOT rotate mid-conversation (only `/new` rotates it), so its previously-scoped memory becomes inaccessible. This only bites users with `API_SERVER_KEY` + a long-term memory provider configured, but it's a silent continuity break for them.

Because the header serves double duty (routing needs the parseable format; memory scoping needs a stable identity) and there's no clean single-header WebUI-only fix, the correct fix spans repos:
- **Preferred:** keep sending a routing-parseable key, but add legacy-form parsing / a memory alias in hermes-agent so an existing `webui:{id}` memory scope maps forward (no orphaning) — a coordinated migration.
- Or document the one-time memory-scope reset as acceptable and gate it behind a version note.

### 2. Re-establish the premise — #4691 is CLOSED as resolved in agent v0.18.0
The issue this fixes (#4691 async-delegation results dropped in WebUI sessions) was closed 2026-07-05: *"resolved entirely in hermes-agent v0.18.0 — no WebUI PR needed,"* verified against the async-delegation regression suite. Your PR body argues the durable-delivery change addressed tracking/dedup but not routing — that may be right, but with the issue closed-as-fixed, please attach a **concrete repro on the current installed agent** (v0.18.2) showing an async-delegation result is still dropped for a WebUI gateway session with the `webui:` key. `_build_process_event_source` does a `session_store._entries` lookup before the parse fallback, so if the session is registered there the parse-None path may not be reached — the repro will settle it.

If (2) confirms the routing is genuinely still broken, the (1) memory-alias coordination is the remaining work. Happy to help land the agent-side alias. Credit preserved either way.

Issue #6082, opened 2026-07-15
Disabling update checks shouldn't disable manual updates


Stooovie:
As of now, user has only two choices:
1) enable "Check for updates" and have an "update available" banner almost always present as there are many updates every day

2) no updating via webui at all (checking for updates from Preferences-System is disabled when "check for updates" is off)

Proposal:
Manually checking for updates from Preferences-System should be always available. 

Optional:
Allow the user to change the update check frequency.

nesquena-hermes:
## Summary

You've diagnosed the behavior exactly. Manual update checks are already wired to be independent of the background poll on the frontend — the "Check now" button (`checkUpdatesNow` in `static/panels.js:11877`) sends `{force:true}` and even has a branch to render "updates disabled" if the server declines. The problem is entirely server-side: both `/api/updates/check` handlers short-circuit on the `check_for_updates` setting *before* they ever look at `force`, so an explicit manual click is refused the same way a background poll is.

## Code reference

The POST handler at `api/routes.py:13724` gates unconditionally, then reads `force` several lines *later*:

```python
if parsed.path == "/api/updates/check":
    settings = load_settings()
    if not settings.get("check_for_updates", True):
        return j(handler, {"disabled": True})          # <- fires for manual checks too
    include_agent_updates = not bool(settings.get("ignore_agent_updates"))
    force = bool(body.get("force", False))             # <- force read only after the gate
```

The GET handler at `api/routes.py:13062` has the same early `{"disabled": True}` return. And the frontend already distinguishes the two callers: boot's background check sends `force:false` (`static/boot.js:3375`), while the manual button sends `force:true` (`static/panels.js:11894`).

## Recommendation

Move the `force` read above the gate in the POST handler and only suppress when the check is *not* forced:

```python
if parsed.path == "/api/updates/check":
    settings = load_settings()
    force = bool(body.get("force", False))
    # A background poll respects the setting; an explicit manual check
    # (force=true from the "Check now" button) always runs.
    if not force and not settings.get("check_for_updates", True):
        return j(handler, {"disabled": True})
    include_agent_updates = not bool(settings.get("ignore_agent_updates"))
    ...
```

[...]

PR #6088, opened 2026-07-15
fix: allow manual update check when auto-update is disabled (#6082)

webtecnica:
When check_for_updates is false, the POST /api/updates/check with force=true should still work for manual checks. Only the GET auto-check should return disabled.\n\nFixes #6082.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -13722,7 +13722,12 @@ def handle_post(handler, parsed) -> bool:
     if parsed.path == "/api/updates/check":
         settings = load_settings()
         if not settings.get("check_for_updates", True):
-            return j(handler, {"disabled": True})
+            force = bool(body.get("force", False)) if isinstance(body, dict) else False
+            if force:
+                # Manual force-check bypasses auto-check toggle (#6082)
+                pass
+            else:
+                return j(handler, {"disabled": True})
         include_agent_updates = not bool(settings.get("ignore_agent_updates"))
         force = bool(body.get("force", False))
         # Allow the client to pass the channel explicitly in the POST body. This


nesquena-hermes:
Reading the PR head at `api/routes.py:13721-13731` against the POST `/api/updates/check` handler on `origin/master`, the behavior is correct — a `force:true` manual check now bypasses the `check_for_updates` gate while a background poll (`force:false`) still gets `{"disabled": True}`. That matches the frontend split: the "Check now" button sends `force:true` and boot's background check sends `force:false`. So functionally this resolves #6082.

Two structural notes before merge.

## `force` is parsed twice

The PR reads `force` inside the gate, then the pre-existing line reads it again a few lines down:

```python
if not settings.get("check_for_updates", True):
    force = bool(body.get("force", False)) if isinstance(body, dict) else False   # 13723
    if force:
        # Manual force-check bypasses auto-check toggle (#6082)
        pass
    else:
        return j(handler, {"disabled": True})
include_agent_updates = not bool(settings.get("ignore_agent_updates"))
force = bool(body.get("force", False))   # 13731 — reads it a second time
```

The two reads also disagree on defensiveness: the first guards with `isinstance(body, dict)`, the second (13731) does not. That's harmless today only because the later code already assumes `body` is a dict, but the mismatch is a smell — if `body` can ever be a non-dict here, 13731 throws `AttributeError` regardless of the guard above it.

## The maintainer already posted the canonical shape

nesquena-hermes gave the intended form directly on #6082 — read `force` once, above the gate, and fold it into the condition:

```python
if parsed.path == "/api/updates/check":
    settings = load_settings()
    force = bool(body.get("force", False)) if isinstance(body, dict) else False
    # A background poll respects the setting; an explicit manual check
    # (force=true from the "Check now" button) always runs.
    if not force and not settings.get("check_for_updates", True):
        return j(handler, {"disabled": True})
    include_agent_updates = not bool(settings.get("ignore_agent_updates"))
    # (drop the second `force = ...` read here — reuse the one above)
```

This is one read, one branch, no `pass`, and it removes the duplicate at 13731. Adopting it would make the diff smaller and match the reviewer's stated intent exactly.

## One thing to confirm

Both the issue thread and this handler leave the GET path's early `{"disabled": True}` return untouched, which is fine since the manual "Check now" button is POST-only. But worth verifying the System settings panel keeps the "Check now" button visible/enabled when the toggle is off — otherwise the now-functional endpoint has no UI entry point. The maintainer flagged the same thing on #6082; the toggle is just a checkbox and doesn't appear to hide the button, but a quick look at the System panel template would close the loop.

No test is included. A small unit test asserting `force:true` returns a real result (not `{"disabled": True}`) while `force:false` returns disabled when `check_for_updates` is off would lock the behavior in.

nesquena-hermes:
Shipped in **exp-v0.52.76** 🚀 Thanks @webtecnica.

Your fix landed as-is (the `force:true` bypass of the disabled auto-check short-circuit). I added a route-level regression test (`tests/test_updates_6088_manual_check.py`) covering the three cases — disabled blocks auto-check, disabled+force bypasses, enabled runs normally — so the behavior is locked.

Gate: Codex adversarial review clean (force bypasses only the disabled short-circuit; non-dict bodies safe); full suite 13,111 pass; browser-smoke zero console errors. Closes #6082.

Issue #6086, opened 2026-07-15
WebUI server silently exits on every /api/updates/check and on gateway restart when a stale gateway_state.json is present (fail-closed recovery)


trailartfoto-png:
### Environment
- hermes-webui `3f3fa9b6` (cloned 2026-07-14, assets served as `exp-v0.52.71`), includes #6054 / #6057
- Native Windows 11, launched via `start.ps1` (the native path from #1952)
- Python 3.11 (uv-managed), hermes-agent under `%LOCALAPPDATA%\hermes\hermes-agent`, gateway runs with `--profile`
- `check_for_updates: true`, channel `stable`

### Symptom
The browser shows the "connection lost" banner and never recovers. The server process
tree (supervisor + worker python) exits silently — no traceback, no log output — within
seconds of **every** automatic `POST /api/updates/check` (fires every ~10–15 min), and
within ~3 seconds of an **external gateway restart**. An external watchdog we set up had
to revive the server **9 times in one morning**:

```
[webui] 08:54:50Z POST /api/updates/check  200  3892ms   → server dead, watchdog restart 10:55
[webui] 09:15:57Z POST /api/updates/check  200  3624ms   → server dead, watchdog restart 11:20
[webui] 09:28:55Z POST /api/updates/check  200  2692ms   → server dead, watchdog restart 11:30
[webui] 09:38:05Z POST /api/updates/check  200  2677ms   → server dead, watchdog restart 11:40
10:56:58 local: gateway restarted via scheduled task     → server dead 3 s later (10:57:01)
```

### Trigger condition
`/api/health/agent` on this machine reported:

```json
{"alive": null, "details": {"state": "unknown", "reason": "gateway_stale_stopped_state",
 "gateway_state": "stopped", "updated_at": "2026-07-10T12:04:06"}}
```

i.e. a **5-day-old stale `gateway_state.json` in the global HERMES_HOME** (leftover from an
earlier global-home gateway install / the Nous desktop app), while the actual gateway runs
with a profile HERMES_HOME and writes fresh state elsewhere.

### Suspected cause
The experimental fail-closed gateway-restart recovery from #6054 ("fail closed on wrapped
gateway pid helpers") appears to treat the stale/contradictory gateway state as unsafe and
terminates the WebUI server process itself — permanently (nothing restarts it), and without
writing any diagnostic output.

### Workaround (fully resolved it)
1. Set `check_for_updates: false`
2. Rename the stale global `gateway_state.json`

[...]

PR #6089, opened 2026-07-15
fix: treat stale gateway_state.json as UNKNOWN, don't exit server (#6086)

webtecnica:
When gateway_state.json is older than 300 seconds, treat the state as UNKNOWN instead of FAILED. Log a warning with the exact age. The server never exits based on stale state.\n\nFixes #6086.

### DIFF
--- a/api/agent_health.py
+++ b/api/agent_health.py
@@ -25,7 +25,6 @@ from __future__ import annotations
 import importlib
 import inspect
 import json
-import logging
 import os
 import threading
 import time
@@ -44,13 +43,6 @@ _GATEWAY_RUNTIME_STATUS_FILE = "gateway_state.json"
 # minutes. Override is intentionally not exposed: keep the check deterministic
 # and identical across deployments so support diagnostics are reproducible.
 GATEWAY_FRESHNESS_THRESHOLD_S: float = 120.0
-
-# General staleness safety net for any gateway state value (e.g. startup_failed).
-# More generous than the state-specific thresholds because this is a catch-all
-# that must never cause the server to exit based on stale on-disk state (#6086).
-STALE_GATEWAY_STATE_THRESHOLD_S: float = 300.0
-
-_logger = logging.getLogger(__name__)
 
 
 def _checked_at() -> str:
@@ -159,42 +151,6 @@ def _runtime_status_is_stale_running(
     if not isinstance(runtime_status, dict):
         return False
     if runtime_status.get("gateway_state") != "running":
-        return False
-
-    raw_updated_at = runtime_status.get("updated_at")
-    if not isinstance(raw_updated_at, str) or not raw_updated_at:
-        return False
-
-    try:
-        updated_at = datetime.fromisoformat(raw_updated_at)
-    except (TypeError, ValueError):
-        return False
-    if updated_at.tzinfo is None:
-        return False
-
-    reference = now if now is not None else datetime.now(timezone.utc)
-    age_s = (reference - updated_at).total_seconds()
-    return age_s > threshold_s
-
-
-def _runtime_status_is_stale(
-    runtime_status: dict[str, Any] | None,
-    *,
-    now: datetime | None = None,
-    threshold_s: float = STALE_GATEWAY_STATE_THRESHOLD_S,
-) -> bool:
-    """Return ``True`` when the gateway state file is stale regardless of state value.
-
-    A catch-all staleness detector for any ``gateway_state`` value (e.g. ``startup_failed``).
-    Unlike the state-specific checks (``_runtime_status_is_stale_running``,
-    ``_runtime_status_is_stale_stopped``), this does not filter by the state value,
-    so a stale ``startup_failed`` or any other unexpected state is detected.
-    A stale file must never cause the server to exit (#6086).
-
-    Returns ``False`` for missing/unparseable timestamps so we never report stale
-    when we cannot determine the age.
-    """
-    if not isinstance(runtime_status, dict):
         return False
 
     raw_updated_at = runtime_status.get("updated_at")
@@ -516,6 +472,8 @@ _remote_probe_cache: dict[str, Any] = {"url": None, "expires_at": 0.0, "result":
 # Condition for the leader's result instead of stampeding the (possibly dead)
 # gateway themselves (#5455 dashboard fan-out, #2476).
 _remote_probe_inflight: set[str] = set()
+
+
 def _remote_probe_wait_budget_s() -> float:
     """How long a latecomer waits for the leader before giving up and self-probing.
 
@@ -850,35 +808,6 @@ def build_agent_health_payload() -> dict[str, Any]:
             "details": {
                 "state": "unknown",
                 "reason": "gateway_stale_running_state",
-                **safe_details,
-            },
-        }
-
-    # Catch-all staleness safety net (#6086): any residual gateway state file
-    # (e.g. ``startup_failed``) that is older than the threshold must never cause
-    # the server to exit. Treat it as UNKNOWN and log a warning so operators can
-    # investigate while the server continues running.
-    if _runtime_status_is_stale(runtime_status, now=checked_at):
-        # Recalculate age for the warning log (helper only returns bool).
-        raw_updated_at = runtime_status.get("updated_at", "") if isinstance(runtime_status, dict) else ""
-        age_s = 0.0
-        if isinstance(raw_updated_at, str) and raw_updated_at:
-            try:
-                updated_at = datetime.fromisoformat(raw_updated_at)
-                if updated_at.tzinfo is not None:
-                    age_s = (checked_at - updated_at).total_seconds()
-            except (TypeError, ValueError):
-                pass
-        _logger.warning(
-            "gateway_state.json is stale (%.0f seconds old) — ignoring",
-            age_s,
-        )
-        return {
-            "alive": None,
-            "checked_at": checked_at,
-            "details": {
-                "state": "unknown",
-                "reason": "gateway_stale_state",
                 **safe_details,
             },
         }
--- a/api/agent_health.py
+++ b/api/agent_health.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import importlib
 import inspect
 import json
+import logging
 import os
 import threading
 import time
@@ -43,6 +44,13 @@ _GATEWAY_RUNTIME_STATUS_FILE = "gateway_state.json"
 # minutes. Override is intentionally not exposed: keep the check deterministic
 # and identical across deployments so support diagnostics are reproducible.
 GATEWAY_FRESHNESS_THRESHOLD_S: float = 120.0
+
+# General staleness safety net for any gateway state value (e.g. startup_failed).
+# More generous than the state-specific thresholds because this is a catch-all
+# that must never cause the server to exit based on stale on-disk state (#6086).
+STALE_GATEWAY_STATE_THRESHOLD_S: float = 300.0
+
+_logger = logging.getLogger(__name__)
 
 
 def _checked_at() -> str:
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Reading the diff at `origin/master...HEAD` for `api/agent_health.py` plus the surrounding `build_agent_health_payload` body (lines 824-882 on the PR head) and the three sibling helpers, the fix logic is correct — a stale residual `gateway_state.json` now resolves to `alive: None` / `gateway_stale_state` instead of falling through to `gateway_not_running`, which is exactly what #6086 needs. One cleanup is worth doing before merge.

## Dead helper + duplicated logic

The PR adds `_runtime_status_is_stale(...)` at lines 180-213, but nothing calls it — the catch-all in `build_agent_health_payload` re-implements the same check inline at 856-882:

```python
if isinstance(runtime_status, dict):
    raw_updated_at = runtime_status.get("updated_at")
    if isinstance(raw_updated_at, str) and raw_updated_at:
        try:
            updated_at = datetime.fromisoformat(raw_updated_at)
            if updated_at.tzinfo is not None:
                age_s = (datetime.now(timezone.utc) - updated_at).total_seconds()
                if age_s > STALE_GATEWAY_STATE_THRESHOLD_S:
                    ...
```

`grep -n "_runtime_status_is_stale\b"` across `api/` and `tests/` returns only the definition — no call site. So the helper is dead code, and the decision logic lives in two places that can drift apart.

## Why it matters beyond tidiness

The two sibling detectors are both *called* from the payload builder — `_runtime_status_is_stale_stopped(runtime_status)` at 835 and `_runtime_status_is_stale_running(runtime_status)` at 846 — and each accepts a `now=` keyword for deterministic testing. The inlined block breaks that pattern: it calls `datetime.now(timezone.utc)` directly, so there's no injection point. That's why the updated test in `tests/test_issue716_agent_heartbeat.py` has to lean on the fixture's hardcoded `updated_at="2026-05-04T12:00:00+00:00"` being far in the wall-clock past rather than pinning `now`. It passes today, but it's testing against the real clock instead of a controlled reference.

## Suggested fix

Route the catch-all through the helper you already wrote, keeping the age only for the log line:

```python
if _runtime_status_is_stale(runtime_status):
    updated_at = datetime.fromisoformat(runtime_status["updated_at"])
    age_s = (datetime.now(timezone.utc) - updated_at).total_seconds()
    _logger.warning(
        "gateway_state.json is stale (%.0f seconds old) — ignoring", age_s,
    )
    return {
        "alive": None,
        "checked_at": checked_at,
        "details": {"state": "unknown", "reason": "gateway_stale_state", **safe_details},
    }
```

That deletes the ~24-line duplicate at 856-882, keeps the single source of truth in `_runtime_status_is_stale`, and (if you later thread a `now=` down into `build_agent_health_payload`) restores the same testability the sibling helpers have. The re-parse of `updated_at` is safe because `_runtime_status_is_stale` already returned `True` only for a parseable, tz-aware timestamp.

Also worth a one-line note in the docstring: the catch-all's `300s` threshold is deliberately more generous than the `120s` `GATEWAY_FRESHNESS_THRESHOLD_S` used by the stopped/running detectors, so a fresh non-`running`/`stopped` state (e.g. a genuine `startup_failed` under 300s old) still reports `down` — which reads as intended, just non-obvious next to the other two.

trailartfoto-png:
Supporting data from my machine (the original #6086 report): after the agent update
of Jul 14, the gateway only rewrites gateway_state.json every ~10-13 minutes, so with
STALE_GATEWAY_STATE_THRESHOLD_S = 300 even a perfectly healthy gateway routinely looks
"stale" — matching the seemingly random server exits 0.5-4 minutes after startup that
I observed. Rolling the agent back to its pre-Jul-14 state (which refreshed the file
more frequently) stops the exits with the same WebUI build. So the exit-on-stale path
gets triggered in normal operation, not just in edge cases — this fix (or also
refreshing the state file more often on the agent side) is very welcome.

nesquena-hermes:
Following up on @trailartfoto-png's threshold concern (the 300s catch-all making a healthy gateway look "stale"). I traced the full decision order in `build_agent_health_payload` at PR head plus the agent-side write triggers, and I think the concern is real but aimed one branch too far down — the important part is a contract mismatch, not the 300s number.

## The catch-all is not what a healthy `running` gateway hits

The payload builder evaluates in this order (PR head `api/agent_health.py`):

1. `_runtime_status_is_fresh` — `running` + `updated_at` ≤ **120s** → `alive: True`
2. `_runtime_status_is_stale_stopped` — `stopped` + > 120s → `alive: None`
3. `_runtime_status_is_stale_running` — `running` + > **120s** → `alive: None`
4. **NEW** catch-all — *any* state + > **300s** → `alive: None` (`agent_health.py:856-882`)
5. else `dict` → `alive: False` (down) — this is the branch that drives the #6086 exit

A healthy-but-idle gateway self-reports `gateway_state: "running"`, so it is caught at step 3 (the pre-existing 120s `_runtime_status_is_stale_running`) and returns `alive: None` long before the 300s catch-all is ever reached. The new block at step 4 only fires for states that are *neither* `running` nor `stopped` — `startup_failed`, `degraded`, `starting`. That's exactly the residual `startup_failed` file #6086 was about, and the fix is correct: those now resolve to `alive: None` (UNKNOWN) instead of falling to `alive: False` (down), which is what `routes.py:1141` treats as `running=False, configured=True` and what the fail-closed recovery reacts to.

## The real issue is the freshness model, and it predates this PR

`_runtime_status_is_fresh`'s own comment says the gateway "writes it on every tick," but the agent contract says the opposite. From `gateway/status.py:893` (agent side):

```
NOTE: liveness keys off ``gateway_running`` (a live PID / health probe),
NEVER ``updated_at`` — a healthy idle gateway never advances that timestamp.
```

I checked the agent write sites: `updated_at` only advances on lifecycle transitions (`gateway/run.py:4342` running, `:6893` starting, `:7007` startup_failed, draining, stopped) and on turn boundaries via `_persist_active_agents` (`gateway/run.py:4577`). There is **no periodic keepalive writer** — I grepped `gateway/*.py` and the only keepalive is `nous_auth_keepalive`, unrelated to `gateway_state.json`. So @trailartfoto-png's "~10-13 min between rewrites on an idle box" is expected agent behavior, not a regression, and step 3 will flip a perfectly healthy idle `running` gateway to `alive: None` within ~2 min of its last turn.

## Recommendation

This PR is a correct, well-scoped band-aid for the *crash* (residual non-running states → UNKNOWN instead of down → no exit). Merge it for that. But it doesn't fix the underlying freshness-model mismatch, and stamping this as "fixes #6086 fully" would be optimistic. The durable fix is one of:

- **Agent side (preferred):** a periodic keepalive that re-stamps `updated_at` on an idle `running` gateway (e.g. every 30-60s), so the WebUI freshness heuristic has a real heartbeat to key off — matching the "writes on every tick" assumption the WebUI already encodes.
- **WebUI side:** for same-host deployments where `get_running_pid()` succeeds, prefer PID liveness over `updated_at` age entirely, and only fall back to the timestamp heuristic in the cross-container/PID-namespace case the `_runtime_status_is_fresh` docstring describes.

Separately, my earlier note still stands: the `_runtime_status_is_stale(...)` helper added at `agent_health.py:180-213` is never called — the step-4 block re-implements it inline at `856-882`. Routing the catch-all through that helper removes the duplication and restores the `now=`-injectable testability the sibling detectors have.

nesquena-hermes (changes requested):
Thanks @webtecnica — the direction here is right (a stale leftover `gateway_state.json` shouldn't be reported as a hard failure), and I did a maintainer pass on top of your PR (wired in the `_runtime_status_is_stale` helper that was defined but never called, and added coverage). But the full release gate (Codex adversarial review + full suite) surfaced a real **SILENT** correctness problem in the catch-all that I don't want to ship as-is, so I'm kicking it back with a precise spec.

## The blocker: the universal age-only downgrade masks a *genuine* outage

The catch-all at `api/agent_health.py` reclassifies **any** `gateway_state` older than 300s to `alive:null` / `state:unknown`. That's too broad: an explicit **terminal `startup_failed`** with no live PID is a real, actionable failure — but after 300s it now flips from `alive:false`/`down` to `alive:null`/`unknown`. Downstream that:

- hides the outage banner (`static/ui.js` ~9200), and
- flips Settings from "configured but down" to "not configured" (`api/routes.py` ~1144).

So a gateway that genuinely failed to start becomes invisible to the user after 5 minutes. Verified: fresh failure at 299s correctly stays down; missing/naive/malformed/future timestamps correctly stay down; a live PID still wins. The problem is specifically the *terminal/unrecognized* states being swept into the same age bucket as benign residue.

## What the codebase already does right

There are already **state-specific** stale detectors that handle the legitimate cases precisely:
- `_runtime_status_is_stale_running(...)` → a stale `running` record (crash-without-cleanup) → `unknown`.
- `_runtime_status_is_stale_stopped(...)` → a stale `stopped` record → `unknown`.

Those are the safe reclassifications. #6086's actual trigger (a stale leftover `running`/`stopped` from an old gateway) is already covered by them, or is a small extension of them.

## Fix-spec

Drop the universal age-only catch-all. Instead, either:
1. **Preferred:** don't add a new catch-all at all if `_runtime_status_is_stale_running` / `_stopped` already cover the reporter's exact state — confirm which `gateway_state` value #6086's stale file carried (the issue shows `startup_failed`? or `running`/`stopped`?) and extend the *matching* state-specific detector, OR
2. add a stale detector scoped to **only** the states that are safe to age out (benign residue), and keep **terminal `startup_failed` (and any unrecognized state) reported as `down`** unless an independent ownership signal (profile/config match, or the file provably belongs to a different/old gateway) proves it's unrelated residue.

Add a regression test for a **>300s `startup_failed`** asserting it stays `down` (not `unknown`) — that's the exact case that must not regress.

Also note: nothing in the WebUI repo `sys.exit()`s on this payload (I traced both callers — `routes.py:1132` and `routes.py:12093`, both read-only responders; the only `server.py` exit is the startup port-conflict guard). The "server exits" in #6086 is the reporter's **external watchdog** restarting on a failed `/api/health/agent`. So the right framing for this fix is "stop reporting *benign* stale residue as a failure," not "stop the server exiting" — the server never exits on its own here.

Happy to take this over and build the scoped version myself if you'd prefer — just say the word.

nesquena:
@webtecnica Email me (see profile), I'd like to add you to our dev chat!

webtecnica:
Addressed the review feedback: `build_agent_health_payload` now delegates the catch-all staleness check to `_runtime_status_is_stale(runtime_status, now=checked_at)` instead of inlining the same logic. The helper's `now` parameter is used so tests can inject a timestamp.

- Removed duplicated `isinstance`/`fromisoformat`/`tzinfo` checks
- Uses `checked_at` instead of `datetime.now(timezone.utc)` for clock consistency with sibling helpers
- Age for the warning log is still calculated inline (helper only returns bool)

webtecnica:
@nesquena Great, will reach out via email! Thanks!

webtecnica:
Addressed the maintainer review feedback: removed the universal age-only catch-all per the PR spec. Changes pushed to `webtecnica:fix/6086-stale-gateway-state`:

**`api/agent_health.py`:**
- Removed `import logging` (no longer needed)
- Removed `STALE_GATEWAY_STATE_THRESHOLD_S` constant (300s universal catch-all)
- Removed `_logger = logging.getLogger(__name__)`
- Removed `_runtime_status_is_stale()` function definition entirely
- Removed the catch-all block from `build_agent_health_payload()` — the block that returned `alive: None` / `state: unknown` / `reason: gateway_stale_state` for any state older than 300s

**State-specific detectors preserved:**
- `_runtime_status_is_stale_running()` — stale `running` state → `unknown` (handles crash-without-cleanup)
- `_runtime_status_is_stale_stopped()` — stale `stopped` state → `unknown` (handles leftover residue)
- These already cover the #6086 triggers precisely

**`tests/test_issue716_agent_heartbeat.py`:**
- Reverted `test_agent_health_payload_down_when_gateway_metadata_exists_but_no_process`:
  - `alive` is `False` (was `None`)
  - `details.state` == `"down"` (was `"unknown"`)
  - `details.reason` == `"gateway_not_running"` (was `"gateway_stale_state"`)
  - This ensures a stale `startup_failed` (or any unrecognized state with no PID) stays reported as `down`, not `unknown`

webtecnica:
Per maintainer review: the universal age-only catch-all masked genuine outages (stale `startup_failed` staying `down` is correct). The state-specific detectors `_runtime_status_is_stale_running` and `_runtime_status_is_stale_stopped` already handle the safe reclassification cases from #6086. Since the catch-all was removed and no additional code is needed, closing this PR. The state-specific solution was already correct on master.

webtecnica:
Closing as the state-specific detectors already cover the #6086 scenario. The universal catch-all was removed per review.

Issue #6085, opened 2026-07-15
feat: WebUI message timestamp injection via `gateway.message_timestamps.enabled`


qxxaa:
## Problem

The WebUI conversation path gives the model no reliable signal for when each message was received. The system prompt carries only a session-start date (`Conversation started: Wednesday, July 15, 2026`) with no clock time and no timezone offset. As a result the model cannot:

- distinguish morning from evening in greetings or time-sensitive responses
- calculate elapsed time between messages within a session
- detect date changes across a long or resumed session
- avoid inaccurate duration claims (e.g. "work done in an afternoon" when the transcript spans three days)

The gateway already solves this for gateway-connected platforms by prepending a human-readable timestamp to each inbound user message when `gateway.message_timestamps.enabled: true` is set. The WebUI streaming path does not participate in that injection, so the feature is silently inactive for WebUI sessions regardless of the config value.

## Proposal

Extend the WebUI message construction path to honour the existing `gateway.message_timestamps.enabled` config key, prepending a timestamp to each inbound user message in the same way the gateway does. The timestamp should use the configured system timezone and include an explicit UTC offset so the model can unambiguously interpret local time.

Additionally, expose the `gateway.message_timestamps.enabled` toggle in the WebUI settings panel, consistent with the existing precedent of surfacing config keys such as provider and model through the UI.

The exact timestamp format and whether to align it with the gateway's existing format or adopt ISO-8601 are left to the maintainer's discretion.

## Expected behaviour (when enabled)

Each model-facing user message begins with a timestamp prefix, persisted in conversation history so it remains available during session replay and context reconstruction. The visible transcript shown to the user remains clean.

## Implementation notes

The following is offered as orientation for whoever picks this up - not a prescriptive spec.

**Injection point - `api/streaming.py`**

The WebUI already builds a per-turn workspace prefix via `_workspace_context_prefix()` and prepends it to every model-facing user message through `_build_native_multimodal_message()`. A timestamp prefix belongs in the same location, immediately before the workspace tag.

The turn receipt time is already available at that point as `_turn_started_at` (derived from `s.pending_started_at`, the float set when the user submits the message). No new timestamp capture is needed.

A companion helper modelled on `_workspace_context_prefix()`:

```python
def _time_context_prefix(ts: float, tz_name: str = 'Europe/London') -> str:
    import datetime, zoneinfo
    tz = zoneinfo.ZoneInfo(tz_name)
[...]

PR #6090, opened 2026-07-15
feat: config-gated message timestamp injection via gateway.message_timestamps (#6085)

webtecnica:
When gateway.message_timestamps.enabled is true, prepend a [Time: ...] prefix to every model-facing user message using the system timezone.\n\n- New _time_context_prefix() helper\n- Config-gated in _build_native_multimodal_message()\n- Stripped in _strip_workspace_prefixes_for_compare() and _looks_like_current_user_turn()\n- Uses system timezone via zoneinfo\n- Default: disabled (no behavioral change)\n\nFixes #6085.

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -20,8 +20,10 @@ import threading
 import time
 import traceback
 import copy
+from datetime import datetime
 from pathlib import Path
 from typing import Optional
+import zoneinfo
 
 logger = logging.getLogger(__name__)
 
@@ -141,6 +143,9 @@ def _cancel_event_payload(
 _ENV_LOCK = threading.Lock()
 
 _KEYLESS_CUSTOM_API_KEY = "dummy-key"
+
+_TIMESTAMP_PREFIX_RE = re.compile(r'^\[Time:\s+\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}[+-]\d{2}:\d{2}\]\n')
+
 _STREAM_WRITEBACK_DIAG_DEFAULT_THRESHOLD_MS = 250.0
 
 _STREAMING_CRON_PROFILE_HOME: contextvars.ContextVar[str | None] = contextvars.ContextVar(
@@ -2143,7 +2148,26 @@ def _resolve_image_input_mode(cfg: dict) -> str:
     return "native"
 
 
-def _build_native_multimodal_message(workspace_ctx: str, msg_text: str, attachments, workspace: str, *, cfg: dict = None):
+def _time_context_prefix(turn_started_at: float) -> str:
+    """Format a ``[Time: …]\\n`` prefix for the given Unix timestamp.
+
+    Uses the system timezone (``time.tzname[0]``) via ``zoneinfo`` so the
+    prefix reflects the local offset of the machine running WebUI. Falls back
+    to UTC silently on any error — never raises.
+    """
+    try:
+        tz = zoneinfo.ZoneInfo(time.tzname[0])
+        dt = datetime.fromtimestamp(turn_started_at, tz=tz)
+        return f"[Time: {dt.strftime('%Y-%m-%dT%H:%M:%S%z')}]\n"
+    except Exception:
+        try:
+            dt = datetime.fromtimestamp(turn_started_at, tz=zoneinfo.ZoneInfo("UTC"))
+            return f"[Time: {dt.strftime('%Y-%m-%dT%H:%M:%S%z')}]\n"
+        except Exception:
+            return ""
+
+
+def _build_native_multimodal_message(workspace_ctx: str, msg_text: str, attachments, workspace: str, *, cfg: dict = None, turn_started_at: Optional[float] = None):
     """Build native multimodal content parts for current-turn image uploads.
 
     WebUI uploads files into the active workspace. For image files, pass the
@@ -2155,6 +2179,16 @@ def _build_native_multimodal_message(workspace_ctx: str, msg_text: str, attachme
     mode is ``"text"``, returns a plain string (attachments are not embedded) so
     the agent's text-mode pipeline (``vision_analyze``) handles images.
     """
+    # Prepend message-timestamp prefix when configured
+    _timestamp_enabled = bool(
+        isinstance(cfg, dict)
+        and cfg.get('gateway', {}).get('message_timestamps', {}).get('enabled')
+    )
+    if _timestamp_enabled and turn_started_at is not None:
+        _ts_prefix = _time_context_prefix(turn_started_at)
+        if _ts_prefix:
+            workspace_ctx = _ts_prefix + workspace_ctx
+
     if not attachments:
         return workspace_ctx + msg_text
 
@@ -2667,6 +2701,7 @@ def _looks_like_current_user_turn(msg, msg_text) -> bool:
     if not needle:
         return False
     text = _message_text(msg.get('content', ''))
+    text = _TIMESTAMP_PREFIX_RE.sub('', text)
     candidates = [_strip_workspace_prefix(text, include_legacy=True)]
     for pattern in (_WORKSPACE_PREFIX_ANY_RE, _LEGACY_WORKSPACE_PREFIX_ANY_RE):
         for match in pattern.finditer(text):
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Read the diff at HEAD against `origin/master` and traced how the new `[Time: …]\n` prefix interacts with the existing workspace-sentinel stripping. There's a concrete round-trip bug: the timestamp is prepended **in front of** the workspace prefix, but the workspace-strip regex is start-anchored, so several strip sites the PR didn't touch will silently stop matching when `message_timestamps.enabled` is true.

## The anchoring problem

`_build_native_multimodal_message` produces `[Time: …]\n[Workspace::v1:…]<msg_text>` (timestamp first). But `_WORKSPACE_PREFIX_RE` is anchored at string start (`api/streaming.py:2630` on master):

```python
_WORKSPACE_PREFIX_RE = re.compile(r'^\s*\[Workspace::v1:\s*(?:\\.|[^\]\\])+\]\s*')
```

`^\s*` will not skip a leading `[Time: …]\n`, so `_WORKSPACE_PREFIX_RE.sub('', value, count=1)` inside `_strip_workspace_prefix` (`streaming.py:2644`) is a no-op once the timestamp is in front. The PR compensates in exactly two callers — `_looks_like_current_user_turn` and `_strip_workspace_prefixes_for_compare` — by running `_TIMESTAMP_PREFIX_RE.sub('', …)` first. But every other caller of `_strip_workspace_prefix` / `_message_identity` was left as-is.

## The one that actually bites: current-turn dedup

`_message_identity` (`streaming.py:4598`) strips the workspace prefix for `role == 'user'` so the model-facing `[Workspace::v1:…]Ok` and the optimistic bubble `Ok` hash to the same identity. It was **not** given the timestamp strip. And `_drop_checkpointed_current_user_from_context` (`streaming.py:4888`) relies on `_message_identity` alone — no `_looks_like_current_user_turn` fallback:

```python
current_user_key = _message_identity({'role': 'user', 'content': msg_text})
if current_user_key and _message_identity(history[-1]) == current_user_key:
    return history[:-1]
```

With the feature on, `history[-1]` is `[Time: …]\n[Workspace::v1:…]Ok`. `_message_identity` fails to strip the workspace prefix (timestamp blocks the anchor), so its key includes both sentinels and no longer equals `current_user_key` (built from raw `msg_text`). The eager-checkpointed current user turn is therefore **not** dropped from model history → the exact "two adjacent user bubbles" duplication the comment at `_message_identity` warns against, now reintroduced but only when timestamps are enabled.

Same silent breakage (lower severity, cosmetic) at:
- `_fallback_title_from_exchange` (`streaming.py:3404`) → local fallback titles would embed `[Time: …]`
- `_normalize_fresh_chat_text` (`streaming.py:5121`) → fresh-chat comparison
- the weak-match strip at `streaming.py:4861`

## Recommended fix — strip the timestamp inside `_strip_workspace_prefix`

Rather than sprinkling `_TIMESTAMP_PREFIX_RE.sub` at each call site (easy to keep missing), centralize it where the workspace prefix itself is removed. `_TIMESTAMP_PREFIX_RE` is already defined at `streaming.py:141`:

```python
def _strip_workspace_prefix(text: str, *, include_legacy: bool = False) -> str:
    value = str(text or '')
    value = _TIMESTAMP_PREFIX_RE.sub('', value)   # NEW: drop leading [Time: …]\n first
    stripped = _WORKSPACE_PREFIX_RE.sub('', value, count=1)
    if include_legacy and stripped == value:
        stripped = _LEGACY_WORKSPACE_PREFIX_RE.sub('', value, count=1)
    return stripped.strip()
```

That fixes `_message_identity`, `_fallback_title_from_exchange`, `_normalize_fresh_chat_text`, and the 4861 weak match in one place, and makes the two explicit strips the PR already added harmless/redundant. `_strip_workspace_prefixes_for_compare` can keep its own strip or defer to this.

## Test gap

`changedFiles: 1` — no test was added. Given this touches the merge/dedup identity path, a regression test with `message_timestamps.enabled=True` asserting that (a) `_message_identity` of `[Time: …]\n[Workspace::v1:…]Ok` equals `_message_identity` of raw `Ok`, and (b) `_drop_checkpointed_current_user_from_context` still drops the checkpointed current turn, would lock this down. Worth adding before merge since the failure is invisible with the feature off (default) and CI is green precisely because nothing exercises the enabled path.

Nice touch on the never-raises fallback in `_time_context_prefix`. One minor aside: `zoneinfo.ZoneInfo(time.tzname[0])` can raise `ZoneInfoNotFoundError` for abbreviations like `EST`/`PST` on some systems — the `except Exception` → UTC fallback handles it, so the offset just degrades to `+0000` rather than erroring. Fine as a fallback, just flagging that the "local offset" promise won't hold on those hosts. (Read-only review — I did not execute the PR's code or tests.)

nesquena-hermes:
## 🔴 Authoritative gate certificate — RED (deterministic timestamp contract failures)

**Exact contributor head:** `8dad636658442a718d4c687392141073676234d6`

I gated this one-file head after a clean transplant onto current `origin/master` (`deac1384f`) on `gate-rebase/6090-timestamps` at `3d9e945b77c3f5a09a0feea6927be1bd8dcfed5a`.

This gate was short-circuited after reproducing two deterministic core-path blockers already identified by warm-up. Running the whole suite cannot turn these failures green.

### Evidence
- Threat scan: **CLEAN**.
- Python compile + diff checks: passed.
- Existing timestamp/native-image slice: **41 passed**, but the PR adds zero tests for its own feature.
- Direct probe of the actual PR helpers:
  - generated prefix: `[Time: 1970-01-01T00:00:00+0000]`
  - `_TIMESTAMP_PREFIX_RE.match(generated_prefix)`: **False**
  - simulated `time.tzname=('PST','PDT')`: silently emitted **UTC**, not local Pacific time.

### Blocking findings
1. **The strip/dedup safeguard can never recognize the prefix this PR generates.** `_time_context_prefix()` uses `strftime('%z')`, producing `+0000`; `_TIMESTAMP_PREFIX_RE` requires `+00:00`. The timestamp therefore leaks into `_looks_like_current_user_turn()` and `_strip_workspace_prefixes_for_compare()`, so exact-checkpoint dedup and stale-merge repair can misidentify the current user turn when the feature is enabled.
2. **The local-time implementation is not portable and is DST-wrong.** `ZoneInfo(time.tzname[0])` passes abbreviations such as `PST`/`PDT`, which are not portable IANA zone keys; failure silently falls back to UTC. Where an abbreviation exists, index 0 is generally the standard-time name, so summer offsets can still be wrong.
3. **No feature-specific regression tests exist.** The green focused slice does not exercise `_time_context_prefix()`, offset formatting, local timezone/DST, or timestamp stripping.

### Required fix
- Generate the prefix with `datetime.isoformat(timespec='seconds')` (or deliberately accept both colon and non-colon offsets in the regex) and test the generated value round-trips through every stripping/comparison helper.
- Use `datetime.fromtimestamp(ts).astimezone()` / the system local tzinfo instead of constructing `ZoneInfo` from `time.tzname[0]`; cover non-UTC and DST boundaries.
- Add enabled/disabled, multimodal/text, dedup/stale-merge, malformed-config, UTC/non-UTC, and DST regression tests.

`gate-fail` and `changes-requested` are correct for this exact head. Nathan's product/security decision remains separate; the opt-in implementation must first be technically correct.

nicodoggie:
Related implementation: NousResearch/hermes-agent#62508 has now been refreshed onto current agent `main` (`d71033a407`, head `06a104a995`; 695 focused tests pass).

It promotes timestamp rendering into shared turn assembly rather than duplicating it in WebUI streaming, while preserving clean transcripts and prompt-cache `api_content` replay. It has been dogfooded through WebUI, Desktop, and Telegram.

Given this PR's current gate failures around prefix round-tripping, timezone handling, dedup, and missing enabled-path tests, I suggest treating the agent-core PR as the primary injection path. A WebUI follow-up could then be limited to exposing the global `message_timestamps.enabled` setting once the core key lands, avoiding two timestamp formats and two replay/dedup contracts.

If this WebUI-local implementation continues independently, there is one additional path-parity bug to cover: the initial `_build_native_multimodal_message(...)` call at `api/streaming.py:8869` passes `_turn_started_at`, but the notification-retry rebuild at `:8909-8915` omits it. That retry path would silently lose the timestamp even after the format/timezone/dedup blockers are fixed; both call sites should be covered by the enabled-path tests.

PR #6092, opened 2026-07-15
fix: block stale Agent runtime after source updates

pxxD1998:
## Thinking Path

- Hermes WebUI intentionally imports `run_agent.AIAgent` into its long-lived Python server process for the default local chat path.
- If the Hermes Agent Git checkout is updated externally while that server stays alive, `sys.modules` can retain modules from the old revision while later lazy imports read files from the new revision.
- That creates one interpreter containing two source revisions. In-process `importlib.reload()` is not safe for the Agent's existing classes, clients, locks, and singletons.
- The safe boundary is therefore the common Agent import path: detect a checkout revision change, fail closed, and require a backend restart before another Agent-backed action proceeds.

## What Changed

- Added `api.agent_runtime` as the single chokepoint for importing `AIAgent`.
- Bind revision identity to the checkout that actually supplied the loaded `run_agent` module via `run_agent.__file__`, rather than assuming the configured discovery directory is the imported checkout.
- Added fail-closed revision checks before streaming tool prewarm, auxiliary Agent imports, and every route-level `AIAgent` acquisition.
- WebUI-owned local entry points (`/api/chat/start`, `/api/chat`, `/api/btw`, `/api/background`, `start_session_turn`, and the shared local stream admission helper) return a typed, retryable `409 agent_runtime_stale` before session, pending-turn, or hidden-session mutation. Gateway and genuinely runner-owned execution bypass the WebUI-local source check.
- Preserved the existing startup import attempt and later lazy retry when Agent dependencies are installed after WebUI starts.
- Added a regression test that uses a temporary Git-backed synthetic Agent checkout: load revision A, commit revision B while the WebUI process stays alive, then verify both streaming and route import paths reject the stale runtime with an actionable restart message.
- Updated the Agent-source dependency audit anchors, existing source-boundary regressions, and troubleshooting docs.

## Maintainer Review Follow-up

Candidate: `3f7c2944d18e7eb07c0da4101754423d41bbb5b4`.

- **Gateway `/goal` ownership:** goal kickoff resolves gateway ownership and passes it into the shared stream admission helper. Gateway-owned goal turns bypass the WebUI-local Agent guard; local goal turns remain guarded.
- **Single ownership snapshot:** stream admission resolves gateway ownership once. That immutable decision is passed directly into the production runtime barrier and also selects the gateway/local worker. The barrier does not re-read gateway config for this path, preventing config-reload TOCTOU drift.
- **Local/gateway inverse coverage:** the production-wiring regression fails on any second ownership read, runs the real barrier, verifies exactly one local revision check versus zero gateway checks, and verifies the matching local/gateway worker target.
- **Handoff stale response:** `AgentRuntimeChangedError` is caught before the generic handoff fallback handler and returned as typed retryable `409 agent_runtime_stale`. The stale path does not persist a fallback summary or mark the handoff handled.
- **Manual compression stale response:** `_handle_session_compress` catches `AgentRuntimeChangedError` before its generic compression handler and returns the same typed retryable 409. Its regression uses a persisted session and verifies the reloaded session state remains unchanged.
- **Git commit-message stale response:** both staged and selected commit-message handlers catch `AgentRuntimeChangedError` before their generic 500 handlers and preserve the typed retryable 409 contract. `GitWorkspaceError` routing remains unchanged.
- **Prior review fixes retained:** runner-local/gateway scope, background/direct admission, loaded-module identity, tracked membership, and literal pathspec handling remain covered by the guard suite.

## Why It Matters

Without this guard, a routine external Agent update can leave a running WebUI backend in a mixed-revision state and produce misleading import failures on the next chat turn. The new behavior refuses to reuse stale in-memory Agent modules and tells the operator to restart WebUI instead of attempting an unsafe partial reload.

## Verification

- Maintainer-requested regressions failed before their fixes: gateway `/goal` omitted external ownership, and stale handoff returned HTTP 200 fallback success.
- The ownership-snapshot regression exposed both intermediate implementations: worker selection re-read gateway config, then the production barrier re-read it after the helper snapshot. The final test uses the production barrier and rejects any second predicate call.
- The manual-compression regression failed before its fix with generic HTTP 400; the final behavior is typed retryable HTTP 409 and the persisted session is unchanged.
- Both staged and selected git commit-message regressions failed before their fix with generic HTTP 500; each now returns typed retryable HTTP 409.
- Focused runtime guard, goal, handoff, compression, and commit-message suite on candidate `3f7c2944d18e7eb07c0da4101754423d41bbb5b4`: **66 passed**.
- Broader workspace-Git, runtime-adapter, gateway, goal, handoff, compression, and guard suite: **223 passed, 1 skipped**.
- Python compile, `git diff --check`, and diff-scoped Ruff lint passed; Ruff reported **0 findings on added/modified lines**.
- Candidate `dedfa08d…` immediately preceding this amendment received **18/18 passing GitHub checks**. Candidate `3f7c2944…` triggers a fresh CI run; the earlier result is not represented as validation of this amended SHA.
- Full local suite was run on an earlier candidate before the final ownership/compression/commit-message amendments: **13,028 passed, 104 skipped, 1 xfailed, 2 xpassed, 34 subtests passed; 1 failed**. The remaining failure was existing order-dependent external `hermes-agent` import contamination; full-suite green is not claimed for the final candidate.
- HMA internal fixed-SHA Spec + Risk review for `3f7c2944`: **PASS**, with P0/P1/P2 = 0. This is an independent local quality gate, not maintainer approval.
- No UI files changed, so there is no before/after screenshot.

## Contract Routing

- Contract family: Agent source boundary / runtime Agent execution imports.
- References reviewed: `docs/CONTRACTS.md`, `docs/rfcs/agent-source-boundary.md`, `AGENTS.md`, and `docs/GUIDELINES.md`.
- Invariant: a single WebUI Python process must not continue Agent-backed work after the authoritative Git source revision changes.
- This does not change the existing direct-source integration contract; it adds a fail-closed lifecycle guard around that boundary.

## Risks / Follow-ups

- Git-backed Agent checkouts get revision detection only when the loaded `run_agent` module is tracked by that worktree. Non-Git or untracked package installs have no stable revision signal, so they preserve existing behavior.
- The guard protects new Agent-backed actions after a revision change. It cannot make an external updater coordinate atomically with a turn that is already running; a shared updater/runtime lifecycle protocol would be a separate cross-repository change.
- Each guarded action performs bounded local Git identity checks (`rev-parse --show-toplevel`, literal `ls-files --error-unmatch`, and `rev-parse HEAD`). No dependency, daemon, UI control, or in-process module reload was added.

## Release-note-ready wording

Hermes WebUI now detects when the underlying Git-installed Hermes Agent checkout changes while the WebUI backend is still running. Instead of mixing old in-memory modules with new files and failing unpredictably, Agent-backed actions stop with an instruction to restart WebUI.

## Model Used / AI Disclosure

AI-assisted contribution using Hermes Agent with OpenAI Codex provider, exact model `gpt-5.6-sol`, reasoning effort `medium`. AI was used for repository analysis, implementation, tests, and review; all file changes, Git state, and test outputs were verified with local tools before opening this PR.

### DIFF
--- a/api/agent_runtime.py
+++ b/api/agent_runtime.py
@@ -24,21 +24,58 @@ _RESTART_MESSAGE = (
 
 
 def _read_agent_revision(agent_dir: Path | None) -> str | None:
-    """Return the checkout HEAD, or ``None`` for a non-Git/unavailable source."""
+    """Return the loaded Agent checkout HEAD, or ``None`` if it is not tracked."""
     if agent_dir is None:
         return None
+
+    module = sys.modules.get("run_agent")
+    module_file = getattr(module, "__file__", None)
+    if not module_file:
+        return None
+
     try:
-        result = subprocess.run(
-            ["git", "-C", str(agent_dir), "rev-parse", "--verify", "HEAD"],
+        module_path = Path(module_file).resolve()
+        worktree_result = subprocess.run(
+            ["git", "-C", str(agent_dir), "rev-parse", "--show-toplevel"],
             check=False,
             capture_output=True,
             text=True,
             timeout=2,
         )
-    except (OSError, subprocess.TimeoutExpired):
+        if worktree_result.returncode != 0:
+            return None
+        worktree = Path(worktree_result.stdout.strip()).resolve()
+        relative_module = module_path.relative_to(worktree).as_posix()
+        tracked_result = subprocess.run(
+            [
+                "git",
+                "--literal-pathspecs",
+                "-C",
+                str(worktree),
+                "ls-files",
+                "--error-unmatch",
+                "--",
+                relative_module,
+            ],
+            check=False,
+            capture_output=True,
+            text=True,
+            timeout=2,
+        )
+        if tracked_result.returncode != 0:
+            return None
+        revision_result = subprocess.run(
+            ["git", "-C", str(worktree), "rev-parse", "--verify", "HEAD"],
+            check=False,
+            capture_output=True,
+            text=True,
+            timeout=2,
+        )
+    except (OSError, subprocess.TimeoutExpired, RuntimeError, ValueError):
         return None
-    revision = result.stdout.strip()
-    return revision if result.returncode == 0 and revision else None
+
+    revision = revision_result.stdout.strip()
+    return revision if revision_result.returncode == 0 and revision else None
 
 
 _AGENT_SOURCE_DIR: Path | None = None
--- a/api/agent_runtime.py
+++ b/api/agent_runtime.py
@@ -9,11 +9,13 @@ mixed runtime and require a clean WebUI restart instead.
 from __future__ import annotations
 
 from pathlib import Path
+import sys
 import subprocess
 import threading
 
-from api.config import _AGENT_DIR
-
+# Retain the discovered path as a diagnostic/test-visible compatibility value;
+# runtime identity is deliberately captured from the loaded module below.
+from api.config import _AGENT_DIR  # noqa: F401
 
 _RESTART_MESSAGE = (
     "Hermes Agent was updated while Hermes WebUI was running. "
@@ -21,8 +23,10 @@ _RESTART_MESSAGE = (
 )
 
 
-def _read_agent_revision(agent_dir: Path) -> str | None:
+def _read_agent_revision(agent_dir: Path | None) -> str | None:
     """Return the checkout HEAD, or ``None`` for a non-Git/unavailable source."""
+    if agent_dir is None:
+        return None
     try:
         result = subprocess.run(
             ["git", "-C", str(agent_dir), "rev-parse", "--verify", "HEAD"],
@@ -37,7 +41,8 @@ def _read_agent_revision(agent_dir: Path) -> str | None:
     return revision if result.returncode == 0 and revision else None
 
 
-_AGENT_REVISION = _read_agent_revision(_AGENT_DIR)
+_AGENT_SOURCE_DIR: Path | None = None
+_AGENT_REVISION: str | None = None
 _AIAgent = None
 _RUNTIME_LOCK = threading.Lock()
 
@@ -46,11 +51,41 @@ class AgentRuntimeChangedError(RuntimeError):
     """Raised when the loaded Agent runtime no longer matches its source tree."""
 
 
+def _loaded_agent_source_dir() -> Path | None:
+    """Return the source directory that actually supplied ``run_agent``."""
+    module = sys.modules.get("run_agent")
+    module_file = getattr(module, "__file__", None)
+    if not module_file:
+        return None
+    try:
+        return Path(module_file).resolve().parent
+    except (OSError, RuntimeError, TypeError):
+        return None
+
+
+def _capture_loaded_agent_revision() -> None:
+    """Bind the guard to the checkout that supplied the loaded Agent module."""
+    global _AGENT_SOURCE_DIR, _AGENT_REVISION
+
+    source_dir = _loaded_agent_source_dir()
+    if source_dir is None:
+        return
+    current_revision = _read_agent_revision(source_dir)
+    if _AGENT_SOURCE_DIR is not None and source_dir != _AGENT_SOURCE_DIR:
+        raise AgentRuntimeChangedError(_RESTART_MESSAGE)
+    if _AGENT_REVISION is not None:
+        if current_revision != _AGENT_REVISION:
+            raise AgentRuntimeChangedError(_RESTART_MESSAGE)
+        return
+    _AGENT_SOURCE_DIR = source_dir
+    _AGENT_REVISION = current_revision
+
+
 def ensure_agent_runtime_current() -> None:
     """Reject a known Git checkout change instead of mixing Python modules."""
     if _AGENT_REVISION is None:
         return
-    if _read_agent_revision(_AGENT_DIR) != _AGENT_REVISION:
+    if _read_agent_revision(_AGENT_SOURCE_DIR) != _AGENT_REVISION:
         raise AgentRuntimeChangedError(_RESTART_MESSAGE)
 
 
@@ -59,6 +94,7 @@ def require_ai_agent_class():
     ensure_agent_runtime_current()
     from run_agent import AIAgent  # noqa: PLC0415
 
+    _capture_loaded_agent_revision()
     return AIAgent
 
 
@@ -74,6 +110,4 @@ def get_ai_agent_class():
             except ImportError:
                 return None
             _AIAgent = agent_class
-            if _AGENT_REVISION is None:
-                _AGENT_REVISION = _read_agent_revision(_AGENT_DIR)
         return _AIAgent
--- a/api/agent_runtime.py
+++ b/api/agent_runtime.py
@@ -1,0 +1,79 @@
+"""Fail-closed guard for in-process Hermes Agent source revisions.
+
+Hermes WebUI currently imports ``run_agent.AIAgent`` into its long-lived server
+process. If the Agent checkout changes while that process is alive, Python may
+combine already-cached modules with newly-read source. Refuse to reuse that
+mixed runtime and require a clean WebUI restart instead.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+import subprocess
+import threading
+
+from api.config import _AGENT_DIR
+
+
+_RESTART_MESSAGE = (
+    "Hermes Agent was updated while Hermes WebUI was running. "
+    "Restart Hermes WebUI before retrying this action."
+)
+
+
+def _read_agent_revision(agent_dir: Path) -> str | None:
+    """Return the checkout HEAD, or ``None`` for a non-Git/unavailable source."""
+    try:
+        result = subprocess.run(
+            ["git", "-C", str(agent_dir), "rev-parse", "--verify", "HEAD"],
+            check=False,
+            capture_output=True,
+            text=True,
+            timeout=2,
+        )
+    except (OSError, subprocess.TimeoutExpired):
+        return None
+    revision = result.stdout.strip()
+    return revision if result.returncode == 0 and revision else None
+
+
+_AGENT_REVISION = _read_agent_revision(_AGENT_DIR)
+_AIAgent = None
+_RUNTIME_LOCK = threading.Lock()
+
+
+class AgentRuntimeChangedError(RuntimeError):
+    """Raised when the loaded Agent runtime no longer matches its source tree."""
+
+
+def ensure_agent_runtime_current() -> None:
+    """Reject a known Git checkout change instead of mixing Python modules."""
+    if _AGENT_REVISION is None:
+        return
+    if _read_agent_revision(_AGENT_DIR) != _AGENT_REVISION:
+        raise AgentRuntimeChangedError(_RESTART_MESSAGE)
+
+
+def require_ai_agent_class():
+    """Import ``AIAgent`` after proving the loaded source revision is current."""
+    ensure_agent_runtime_current()
+    from run_agent import AIAgent  # noqa: PLC0415
+
+    return AIAgent
+
+
+def get_ai_agent_class():
+    """Return ``AIAgent`` while preserving the existing lazy-import retry."""
+    global _AIAgent, _AGENT_REVISION
+
+    with _RUNTIME_LOCK:
+        ensure_agent_runtime_current()
+        if _AIAgent is None:
+            try:
+                agent_class = require_ai_agent_class()
+            except ImportError:
+                return None
+            _AIAgent = agent_class
+            if _AGENT_REVISION is None:
+                _AGENT_REVISION = _read_agent_revision(_AGENT_DIR)
+        return _AIAgent
--- a/api/agent_runtime.py
+++ b/api/agent_runtime.py
@@ -23,18 +23,26 @@ _RESTART_MESSAGE = (
 )
 
 
-def _read_agent_revision(agent_dir: Path | None) -> str | None:
+def _read_agent_revision(
+    agent_dir: Path | None,
+    *,
+    module_path: Path | None = None,
+) -> str | None:
     """Return the loaded Agent checkout HEAD, or ``None`` if it is not tracked."""
     if agent_dir is None:
         return None
 
-    module = sys.modules.get("run_agent")
-    module_file = getattr(module, "__file__", None)
-    if not module_file:
-        return None
+    if module_path is None:
+        module = sys.modules.get("run_agent")
+        module_file = getattr(module, "__file__", None)
+        if not module_file:
+            return None
+        try:
+            module_path = Path(module_file).resolve()
+        except (OSError, RuntimeError, TypeError):
+            return None
 
     try:
-        module_path = Path(module_file).resolve()
         worktree_result = subprocess.run(
             ["git", "-C", str(agent_dir), "rev-parse", "--show-toplevel"],
             check=False,
@@ -79,6 +87,7 @@ def _read_agent_revision(agent_dir: Path | None) -> str | None:
 
 
 _AGENT_SOURCE_DIR: Path | None = None
+_AGENT_MODULE_PATH: Path | None = None
 _AGENT_REVISION: str | None = None
 _AIAgent = None
 _RUNTIME_LOCK = threading.Lock()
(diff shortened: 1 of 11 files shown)


nesquena-hermes:
Read the full diff at HEAD `742eff863` against `origin/master` — `api/agent_runtime.py` (new), the eight `run_agent`/`AIAgent` call-site rewrites in `api/routes.py`, the `api/streaming.py` import swap, and `tests/test_agent_runtime_revision_guard.py`. The design holds up well; a few notes below.

## What the guard actually enforces

The chokepoint is clean: `require_ai_agent_class()` runs `ensure_agent_runtime_current()` before importing, and after the diff the *only* remaining direct import of `run_agent` in `api/` is inside `agent_runtime.py:60` itself. The dependency-audit anchor was moved to match (`tests/test_agent_source_dependency_audit.py`: `("api/agent_runtime.py", "run_agent")`), so the "single import path" invariant is enforced by a test — good.

The baseline revision is captured at import time in two places: module-load `_AGENT_REVISION = _read_agent_revision(_AGENT_DIR)` (`agent_runtime.py:38`) and the eager `AIAgent = get_ai_agent_class()` at `streaming.py` module load, which matches the pre-guard behavior. Confirmed by the probe test asserting `streaming._get_ai_agent().revision == "before"` before mutating source.

## Gateway bypass is consistent across both entry points

Worth calling out because the two guards look asymmetric at first glance. The chat-start guard is conditional:

```python
if not webui_gateway_chat_enabled(get_config()):
    try:
        ensure_agent_runtime_current()
    except AgentRuntimeChangedError as exc:
        return j(handler, {"error": str(exc), "type": "agent_runtime_stale", "retryable": True}, status=409)
```

but the streaming-path guard at `streaming.py:7215` is unconditional. That's actually correct, not a gap: `routes.py:20876` picks the worker by backend —

```python
backend_is_gateway = webui_gateway_chat_enabled(get_config())
worker_target = _run_gateway_chat_streaming if backend_is_gateway else _run_agent_streaming
```

so a gateway turn never enters `_run_agent_streaming`, and the unconditional `ensure_agent_runtime_current()` inside it only ever fires on the local path. The two guards are gated by the same predicate, just expressed at different layers. No behavior change for gateway users, as the PR body claims.

## One thing to double-check: the mid-stream raise path

`ensure_agent_runtime_current()` at `streaming.py:7215` sits inside the outer `try:` (opened at 7082) whose handler is `except Exception as e:` at 10079. Since `AgentRuntimeChangedError(RuntimeError)` is an `Exception` subclass, a stale revision detected here won't escape as an unhandled traceback — it flows into the standard provider-error SSE path (`_classify_provider_error`, HTML-sanitized). That's the desired outcome, but the restart message will be classified as a generic provider error rather than surfaced with the typed `agent_runtime_stale` / `retryable` shape the chat-start `409` uses. If the frontend keys any "please restart WebUI" affordance off that type, the streaming path won't trigger it — the user just sees the message text in a normal error bubble. Not a blocker, but worth a one-line classification branch in `_classify_provider_error` (or an explicit `except AgentRuntimeChangedError` ahead of 10079) if you want parity with the 409.

## Cost note

Each guarded action shells out `git -C <agent_dir> rev-parse --verify HEAD` with a 2s timeout. For the streaming path that's once per turn (fine). For the aux-model helpers (`_llm_git_commit_message`, handoff summary, compress) it's once per call, also fine. No caching of the "current" result within a turn, but since the whole point is to detect an out-of-band change, caching would defeat it — the current shape is right.

Overall this looks solid and the test genuinely exercises the fail-closed path with a real git commit between loads. The only substantive follow-up I'd suggest is the streaming-path error typing above so both entry points surface the restart instruction consistently. (Per policy I did not execute any of the PR's code or tests — this is a read-only diff review.)

nesquena-hermes (changes requested):
Thanks @pxxD1998 — this is a real reliability gap (one interpreter holding two Agent revisions after an in-place checkout update) and the core guard is sound: the revision comparison fires correctly on a monitored-checkout change, the main `/api/chat/start` placement is correct (the barrier at routes.py:21674 precedes `_get_or_materialize_session` at 21687), git-absent/non-git installs disable the guard, detached HEAD works, an unreadable known revision fails closed, and the 2s timeout is bounded (100 `rev-parse` calls = 0.17s, no perf/thread-safety issue). The adversarial gate found three shipping risks to close first:

## Must-fix

1. **Runner-local execution is falsely rejected by the local-Agent barrier** (`api/routes.py:21672`). Runner-local runs don't use the in-process `AIAgent` import, so the revision barrier shouldn't apply to them — as written it can return a false-positive `409 agent_runtime_stale` and block a runner-local turn. **Fix:** exempt `runner-local` execution from the barrier using the authoritative runtime-adapter predicate (don't gate a path that isn't running the in-process agent).

2. **Revision identity is bound to the configured `_AGENT_DIR`, not the checkout that actually supplied `AIAgent`** (`api/agent_runtime.py:40`). If the module that provided `AIAgent` came from a different path than the configured dir, you're monitoring the wrong checkout. **Fix:** derive and monitor the git identity of the module that actually supplied `AIAgent` (e.g. from its `__file__`), not merely `_AGENT_DIR`.

3. **Background/direct entry paths bypass the early barrier** (`api/streaming.py:7217`). `/api/background`, `/api/btw`, and `start_session_turn` skip the `/api/chat/start` barrier — background work can return 200, hit the worker guard, be marked complete as "(no answer produced)," and delete its hidden session; process wakeups return 200 and consume their pending marker instead of being re-deferred. **Fix:** add backend-aware synchronous stale checks before state creation/acceptance in those entry points (ideally also at the top of `_start_chat_stream_for_session`), returning the typed retryable 409 *before* any mutation.

(Also: the local test `test_loaded_agent_runtime_fails_closed_after_source_revision_changes` fails on a machine that has a real agent checkout on `PYTHONPATH` shadowing the test's synthetic stub — it passes in CI/clean env. Not blocking, but consider making the probe robust to a pre-imported `run_agent` so it's not environment-sensitive.)

Once the barrier is scoped to genuinely-in-process-agent paths (exempt runner-local), monitors the loaded module's real checkout, and covers the background/direct entry points before mutation, re-request and I'll re-gate. Good fix — it's the placement/scope that needs tightening, not the mechanism.

nesquena-hermes:
Strong iteration, @pxxD1998 — the re-gate confirms all **three** prior blockers are genuinely closed:
- runner-local / subagent execution correctly bypasses the local barrier (scoped by execution owner ✅)
- gateway chat is exempt ✅
- `/api/background`, `/api/btw`, and the process-wakeup path all check **before** session creation/acceptance now, so a bg turn can no longer be marked no-answer and delete the session ✅

14 guard tests + 35 neighboring runtime-adapter tests pass. But the adversarial re-gate reproduced one NEW CORE regression that has to be fixed before this ships.

### [CORE] `git -C <agent_dir> rev-parse HEAD` walks up into an enclosing repo → false stale-block for non-git Agent installs

`api/agent_runtime.py:72` records the Agent revision with `git -C … rev-parse HEAD`. But `rev-parse HEAD` searches **parent directories** — so when the Agent is a non-git installed module nested inside some *unrelated* git worktree (e.g. a `.venv/site-packages/...` install under a repo checkout), it captures the **enclosing repo's** commit instead of correctly treating the Agent as non-git.

Reproduced: an untracked `.venv/site-packages` module returned the enclosing WebUI commit as its "Agent revision", even though `git ls-files --error-unmatch` rejects that module as untracked.

Impact: after **any** unrelated commit in that enclosing repository, `ensure_agent_runtime_current()` sees a revision change and every local send / background / compression action returns `409 agent_runtime_stale` — a false-positive brick — although the Agent source never changed. This defeats the non-git no-op path the guard is supposed to preserve.

### Fix

Before binding `_AGENT_REVISION`, resolve the git top-level and confirm the loaded `run_agent.__file__` is actually **tracked** by that worktree — e.g.:
1. `git -C <dir> rev-parse --show-toplevel` to get the toplevel,
2. `git -C <toplevel> ls-files --error-unmatch -- <run_agent rel path>` (you already use `ls-files --error-unmatch` elsewhere) to confirm the module is tracked there,
3. only then record HEAD; otherwise fall back to the non-git no-op (revision = None, guard disabled).

Add a regression that nests an untracked installed `run_agent` module under an unrelated git repo, commits to that enclosing repo's HEAD, and asserts local Agent actions are **not** stale-blocked.

Everything else on the current head gated clean — the core drift-detection design is sound and the three prior scope risks are closed. This is the last edge. Re-request review once the revision binding is gated on tracked-membership and the covering regression is green.

_Gate: Codex adversarial re-gate + reproduced probe (untracked module under an enclosing git worktree)._

nesquena-hermes:
Good progress, @pxxD1998 — the re-gate **confirms the enclosing-repo fix is closed**: an untracked module under an unrelated git worktree now binds no revision (guard disabled, no false 409), genuinely-tracked checkouts still detect a real source update, and symlinked dirs / submodules / timeout / OSError all fail safe. 16 guard tests + focused neighbors pass.

The adversarial re-gate found **two more callers** of the runtime barrier that handle it incorrectly — same class as the `/api/background`+`/api/btw` bypass you already fixed, just two entry points that were missed. Both need fixing before ship.

### [CORE] `api/routes.py:21643` — gateway-backed `/goal <text>` kickoff false-409s after a local Agent update

The `/goal <text>` kickoff calls `_start_chat_stream_for_session` **without** `external_runtime_owned=True`, even though lines 20907–20908 would route this session through `_run_gateway_chat_streaming` (gateway-owned, so the local in-process Agent barrier doesn't apply). Result: after a local Agent update, a gateway-backed goal kickoff returns `409 agent_runtime_stale` even though it never touches the local runtime — the same false-positive the bg/btw fix closed, on one more caller.

**Fix:** resolve the gateway-ownership flag (the same one lines 20907–20908 use to pick `_run_gateway_chat_streaming`) and pass `external_runtime_owned=True` into `_start_chat_stream_for_session` for the gateway path.

### [SILENT] `api/routes.py:24845` — handoff summary swallows `AgentRuntimeChangedError` as success

The handoff-summary path wraps the call in a broad `except` that catches `AgentRuntimeChangedError`, persists a **fallback** summary, returns HTTP 200, and marks the handoff handled. `static/sessions.js:2800` treats that 200 as success — so on a stale runtime the user silently gets a degraded/fallback handoff summary with no "restart required" signal, and the barrier is defeated for this path.

**Fix:** catch `AgentRuntimeChangedError` **separately, before** the generic fallback handler, and return a typed `409 {type: agent_runtime_stale, retryable: true}` without persisting the fallback or marking the handoff handled — matching the treatment the `_handle_chat_start` barrier already gives.

Both are the same underlying pattern: entry points that either bypass the barrier (gateway-owned false-positive) or mis-handle its exception (swallowed as success). Worth a quick audit of every caller that can reach `require_ai_agent_class()` / `ensure_agent_runtime_current()` so the barrier is handled uniformly — return typed retryable 409 for local paths, exempt genuinely gateway-owned paths.

Re-request review once both callers pass the ownership flag / surface the typed 409, with a covering test for each. The core drift-detection design is solid — this is about the remaining barrier call-sites.

_Gate: Codex adversarial re-gate — enclosing-repo fix verified closed; 2 new barrier call-site findings reproduced._

pxxD1998:
Addressed the remaining stale-runtime call sites on `3f7c2944d18e7eb07c0da4101754423d41bbb5b4`:

- gateway `/goal` now shares one immutable ownership decision between the runtime barrier and worker selection;
- handoff summary and manual compression return typed retryable `409 agent_runtime_stale` without fallback persistence/session mutation;
- staged and selected commit-message generation preserve the same typed 409 instead of generic 500 responses.

Regression coverage: 66 focused tests passed; 223 neighboring tests passed with 1 skipped. GitHub checks are 18/18 passing, and Greptile now reports 5/5 / safe to merge.

Could you please re-review the updated head when convenient?

nesquena-hermes:
Re-gate update, @pxxD1998 — the two findings from the last round are **confirmed fixed** (gateway `/goal` kickoff no longer false-409s, handoff now returns a typed 409, all AIAgent imports go through the guarded chokepoint, enclosing-repo + 3 original scope risks intact; Codex re-verified SAFE and 54 focused tests pass). But running the **full suite single-process** surfaced a real regression the focused tests and the sharded CI both miss, so this can't ship yet.

### [CORE] The new runtime barrier breaks the pre-existing handoff-summary fallback + compress-roundtrip contracts

Full suite (single-process, on a box with a real git-tracked agent checkout): **16 failures**, all on pre-existing tests your PR did NOT touch, clustered on the handoff/compress paths the new barrier now guards:

- `tests/test_issue1013_handoff_dock.py::test_exception_handoff_summary_persists_fallback_summary` (+ retries / falls-back siblings)
- `tests/test_sprint46.py::test_session_compress_roundtrip`, `..._async_reports_stream_state_guard`, `..._start_retries_after_terminal_error`, `..._worker_uses_session_profile_env`, etc.
- `tests/test_issue3800_compaction_summary_length.py`, `tests/test_issue4836_manual_compression_recovery.py`, `tests/test_issue1217_transcript_compaction.py`

Verified this is a genuine regression, not a pre-existing isolation artifact:
- Clean `master`, same box, same single-process invocation → **all 34 pass**.
- PR head `3f7c2944d`, same invocation → **fail** (even running the 2 tests alone: `2 failed`).

Mechanism (pinned): `test_exception_handoff_summary_persists_fallback_summary` swaps in a `_FailingAgent` via `monkeypatch.setitem(sys.modules, "run_agent", ...)` and expects the handoff to **gracefully persist a fallback summary** (`response["ok"] is True`, `fallback is True`). With the PR, `_handle_handoff_summary` → `require_ai_agent_class()` now raises `AgentRuntimeChangedError` because the swapped-in module's identity/revision no longer matches the recorded one, so the code returns the typed `409 agent_runtime_stale` and the graceful-fallback branch is never reached → `KeyError: 'ok'`. `test_session_compress_roundtrip` gets `409` where it expects success for the same reason.

Why CI is green: in the CI environment the agent-runtime barrier is effectively a no-op (different `_AGENT_DIR` resolution / no git-tracked installed agent), so the barrier never fires and the old tests pass. On a real deployment with a git-tracked agent checkout the barrier DOES fire — so this is a real production regression: the existing "handoff/compress degrade gracefully when the agent errors" behavior becomes a hard 409, and (worse) the barrier can fire on paths where the source didn't actually change.

### Fix

The barrier is over-firing. Two things to reconcile:
1. **Scope the barrier to a genuine on-disk source change**, not to any condition where `require_ai_agent_class()` can't return the expected class. A test (or a runtime path) that swaps/errors the agent object must not be misread as "the Agent source tree changed" → don't raise `AgentRuntimeChangedError` unless the recorded revision actually differs from the current tracked HEAD.
2. **Preserve the pre-existing graceful-degradation contract** on the handoff/compress paths: when the agent *call* fails (raises) but the source is unchanged, the existing fallback-summary / retry / terminal-error-status behavior must still apply. Only a true stale-source condition should yield the typed 409.

Please run the FULL suite single-process (`pytest tests/ -p no:xdist`) on a checkout with a git-tracked agent so these 16 surface locally, then adjust the barrier so the pre-existing handoff/compress/compaction tests pass unchanged (they encode shipped behavior) alongside your new stale-runtime tests. The sharded CI won't catch this — it has to be the full single-process run in an environment where the barrier is live.

The core drift-detection design is genuinely close now — this is the barrier being too broad and colliding with the graceful-agent-error paths. Re-request review once the full single-process suite is green.

_Gate: Codex re-gate SAFE on focused tests (2 prior findings closed) + maintainer full-suite run — 16 pre-existing handoff/compress/compaction regressions reproduced, isolated to this PR vs clean master._

nesquena-hermes:
✅ **Re-gate CONVERGED — gate-clean, ready for your merge decision** (holding for maintainer because this is a boot-path / agent-runtime-lifecycle change).

The latest push (`ignore in-memory agent module swaps` + `reject stale runtime before async compression job` + `preserve running job reuse`) **fully closes the 16-test full-suite regression** from the prior round:

- The exact 16 previously-failing pre-existing tests (handoff-summary fallback, compress roundtrip/async/retry, compaction summary-length, manual-compression-recovery, transcript-compaction) now **all pass** — verified 73/73 on the affected files, and the barrier no longer over-fires on a monkeypatched/erroring agent (only a genuine on-disk source change trips it), so the graceful-agent-error degradation contract is preserved.
- **Full single-process suite: 13,167 passed, 1 failed.** The single failure (`test_loaded_agent_runtime_fails_closed_after_source_revision_changes`) is an environment artifact on my box, not a code defect: the subprocess probe's synthetic `run_agent` fixture (which defines `AIAgent.revision`) is shadowed by the real installed `run_agent` here (no `revision` attr) → the `streaming._get_ai_agent().revision == "before"` assert can't see the synthetic value. All 15 sharded CI test jobs pass it (clean container, no installed agent), so it's green everywhere the shadow doesn't exist.

Cumulative gate status across the review rounds — all closed and re-verified intact:
- 3 original scope risks (runner-local bypass, gateway exempt, bg/btw/wakeup check-before-session-mutation) ✅
- enclosing-repo `git rev-parse` false-stale-block (tracked-membership check) ✅
- gateway `/goal` kickoff false-409 + handoff-summary typed-409 ✅
- barrier over-firing on graceful-agent-error paths (16-test regression) ✅

Codex SAFE on the focused suite; the full single-process suite is what caught the barrier-scope regression, and it's now clean. This is genuinely converged after a thorough iteration — nice work @pxxD1998.

**Holding the merge for @nesquena**: this changes the in-process agent-runtime lifecycle + `server.py` startup wiring (boot path), so it wants a maintainer merge decision rather than an autonomous ship. Recommendation: **merge-ready** — I've thrown Codex + the full single-process suite + targeted regression reproduction at it across 6 rounds and it's clean.

_Gate: Codex adversarial (multi-round) + full single-process suite (13,167 pass; the 1 failure is a local PYTHONPATH-shadow env artifact, green on all sharded CI)._

nesquena-hermes:
## Full gate clean — shipping

Thanks @pxxD1998 — the convergence commits (reject-stale-before-async-job, preserve-running-job-reuse, ignore-in-memory-module-swaps, handle-remaining-entrypoints) resolved the over-firing a prior gate found. Re-gated on the current head (rebased onto master):

- **Codex regression gate: SAFE TO SHIP** — the runtime barrier no longer over-fires on the handoff-fallback or compress-roundtrip contracts; the `git rev-parse` chokepoint is scoped to the git-tracked agent checkout (doesn't walk into an enclosing repo); local `/api/chat/start` returns the retryable `409 agent_runtime_stale` before session materialization while gateway-backed chat correctly skips the WebUI-local check.
- **Full single-process suite (prod-like env): 13,176 passed.** The prior 16-regression over-firing sites are green (`test_issue1013_handoff_dock`, `test_goal_command_webui`, `test_sprint46`, `test_regressions` → 110/110). The single failure on my box is `test_loaded_agent_runtime_fails_closed_after_source_revision_changes` — a **test-environment artifact** where this machine's editable `hermes_agent` install shadows the test's synthetic `run_agent` fixture; the test is green across all 15 CI shards (CI has no co-located editable agent). Not a code defect.
- **Sharded CI: 18/18 green.**

Shipping to the experimental channel. Credit preserved.

nesquena-hermes (approved):
Approving: the CONVERGED re-push resolves the over-firing this PR was bounced for. Full authoritative gate on the rebased head is clean — Codex SAFE TO SHIP, full single-process suite 13,176 passed (the prior 16-regression handoff/compress sites now 110/110 green), sharded CI 18/18. Details in the certification comment above.

nesquena-hermes:
🚀 Shipped in **exp-v0.52.89** (experimental channel). Thanks @pxxD1998 — the stale-agent-runtime guard now correctly fails closed after an external source update (typed retryable `409 agent_runtime_stale` on local chat-start before session materialization; gateway-backed chat unaffected), without the over-firing that the earlier head hit on the handoff-fallback and compress-roundtrip paths. Now live on the dogfood box.

Issue #6080, opened 2026-07-15
[BUG] Mobile: composer model selector does nothing — dropdown never opens


nesquena-hermes:
## Summary

On mobile / touch layouts, tapping the composer **model selector does nothing** — the model dropdown never appears, so it's impossible to switch models on a phone/tablet.

> Reported by a community tester (Android app, Spanish locale):
> "When I try to change or select a model using the selector button in the bottom input bar, the dropdown menu does not load or open at all. Tapping the icon or the model name does not trigger any list, making it impossible to switch models on mobile devices."

## Layer: WebUI (web frontend), not the native Android shell

The reporter is using the Android app, but the model picker is entirely a **web-layer** control — the Android app is a thin WebView wrapper over the same web UI, and the screenshot shows the rendered web interface (full composer, Spanish locale). The picker is opened by `toggleModelDropdown()` and positioned by `_positionModelDropdown()` in `static/ui.js`. So this is a `nesquena/hermes-webui` bug and would reproduce in any mobile browser, not just the native app.

Screenshot from the report: an Android tablet in wide layout (~800px). Composer bottom bar left→right: attach, mic, persona, workspace picker, mobile-config (context ring) button, reasoning chip (`◐ XHigh`), send. (The green floating "…" at the top-right of the screenshot is a native Android system overlay, not a WebUI control.)

## Likely root cause

Current `_positionModelDropdown()` (`static/ui.js`) only clamps the **horizontal** offset:

```js
function _positionModelDropdown(){
  ...
  let left=chipRect.left-footerRect.left;
  const maxLeft=Math.max(0, footer.clientWidth-dd.offsetWidth);
  left=Math.max(0, Math.min(left, maxLeft));
  dd.style.left=`${left}px`;
}
```

There is **no visual-viewport / vertical handling**. On mobile, the dropdown is absolutely positioned relative to the composer, so when the visual viewport differs from the layout viewport (browser chrome, on-screen keyboard, address-bar collapse), the opened menu can land **outside the visible visual viewport** or overlap the composer — indistinguishable from "tapping does nothing / nothing opens."

This is exactly the root cause already described in in-flight **PR #6026 ("fix: keep mobile model picker in visual viewport")**, which is currently `OPEN` / `CHANGES_REQUESTED` and has no tracking issue. This issue is that tracking issue.

[...]

PR #6105, opened 2026-07-15
fix: mobile composer model selector - fixed positioning on narrow screens (#6080)

webtecnica:
On mobile, the composer model dropdown was absolutely positioned relative to the composer footer. When the visual viewport changed (browser chrome, on-screen keyboard), the dropdown landed outside the visible viewport — indistinguishable from "tapping does nothing."\n\n- Changed to `position:fixed` on mobile to escape composer's overflow geometry\n- Added viewport-aware repositioning on scroll and resize\n\nCloses #6080

### DIFF
--- a/static/ui.js
+++ b/static/ui.js
@@ -3662,12 +3662,52 @@ function _positionModelDropdown(){
   const panel=$('composerMobileConfigPanel');
   const anchor=(panel&&panel.classList.contains('open')&&mobileAction)?mobileAction:(chip&&chip.offsetParent?chip:mobileAction);
   if(!anchor) return;
-  const chipRect=anchor.getBoundingClientRect();
+  const anchorRect=anchor.getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.82** 🚀 Thanks @webtecnica.

Your visualViewport positioning work is the core of this — I reworked the mechanism to reparent the dropdown to `<body>` while open (the `#profileDropdown` idiom) so `position:fixed` escapes the `.composer-footer` `container-type` / `backdrop-filter` containing-block trap on every skin (that's why the original approach landed off-screen despite correct math — the coordinates were resolving against the footer, not the viewport). Your flip/clamp/visualViewport logic carried over intact.

Gate: Codex SAFE TO SHIP (after catching + closing a drawer-over-picker stacking class via a single `z-index:190` rule — above the config panel it opens from, below the full-screen drawers), Fable UX SHIP-UX, full suite 13,131 pass, maintainer screenshot review across default / Geist Contrast / desktop / config-panel-anchor. Closes #6080.

Issue #6099, opened 2026-07-15
WebUI: Transparent Stream event timestamps on narrow foldables — wrap+break row height, inconsistent left/right position per row type, and no toggle to hide them (exp)


nesquena-hermes:
Reported by **K.Shepherd** in Discord `#testers-talk` (2026-07-15) on **experimental** (screenshot from a Samsung Z Fold 6 cover screen).

> "I just installed the experimental version and I'm now seeing these timestamps on the tool calls. Is this a toggleable feature? Also on skinny displays like my Z Fold 6 they seem to be messing up the height management of the timestamps."

Two parts — one answer, one real bug.

## Part 1 — "Is it toggleable?" → No (feature is #5700, always-on)

The per-event timestamps on Transparent Stream / Worklog activity rows are the **#5700** feature (CLOSED/shipped: "persistent + live per-event timestamps on Transparent Stream activity rows"). There is currently **no setting to turn them off** — I checked; there's no `show_timestamps`/worklog-time preference anywhere in settings or `api/config.py`. So this is expected new behavior, not a bug, but the "can I hide these?" ask is legitimate — worth a Settings toggle (see below).

## Part 2 — Narrow-display timestamp wraps and breaks row height (the bug)

On the Z Fold 6 cover screen the timestamp column **wraps to two lines** on the longest row. In the screenshot (collapsed "Trace: 3 tools" view):
- `4:21 PM  🔧 lcm_grep   Mem0 integrate Hermes` — one line ✓
- `4:21 PM  🔧 session_search  Mem0 Hermes WebUI dashboard embed` — one line ✓
- `4:21 / PM  🌐 web_search  Hermes Agent WebUI external dashboard ifra…` — **timestamp wrapped to TWO lines**, roughly doubling that row's height and vertically decentering the icon/label relative to the neighbors.

The `web_search` row is the one with the longest preview text, so on the narrow width the flex row squeezes the time element until `4:21 PM` breaks across two lines, inflating row height.

### Root cause (source trace)

The **#5739** fix already anticipated this — it added `white-space:nowrap` to `.transparent-event-time` and **hides the timestamp entirely under `max-width:420px`** (`static/style.css` L4134-4149):

```css
.transparent-event-time{ flex-shrink:0; ... white-space:nowrap; font-size:10px; }
@media (max-width:420px){ .transparent-event-time{ display:none; } }
```

[...]

PR #6106, opened 2026-07-15
fix: stream event timestamps layout on narrow foldable screens (#6099)

webtecnica:
On very narrow screens (<360px), stream event timestamps broke the row layout.\n\n- Added `overflow:hidden` to tool/thinking card headers to prevent flex children from spilling out\n- Added `@media (max-width:360px)` hiding `.tool-card-name` — icon + preview + status badge still identify the event\n\nProgressive narrow-screen behavior:\n- ≤ 600px: Larger touch targets\n- ≤ 420px: Timestamp hidden (existing)\n- ≤ 360px: Tool name hidden (new)\n\nCloses #6099

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -4091,6 +4091,7 @@ body.resizing .sidebar{transition:none!important;}
   border-radius:0;
   color:var(--muted);
   transition:background .18s ease,color .18s ease;
+  overflow:hidden;
 }
 .transparent-event-row .tool-card-header:hover,
 .transparent-event-row .thinking-card-header:hover{background:var(--hover-bg);color:var(--text);}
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.81** 🚀 Thanks @webtecnica.

Merged via a scoped release batch. Note: the branch had bundled in a couple of extra rules the release gate flagged (a `#composerModelDropdown{position:fixed}` mobile override that belongs to the separate #6080 model-picker rework, and a `<=360px` tool-name-hide rule that dropped tool identity); those were stripped so this shipped as exactly the stream-timestamp overflow fix. Codex SAFE + Fable UX reviewed + browser-smoke clean.

PR #6107, opened 2026-07-15
fix: honor per-model reasoning effort metadata

CharlesMcquade:
## Summary

Honor `models.<model>.reasoning_efforts` before a provider-wide `reasoning_efforts` list when resolving the WebUI reasoning picker.

Hermes Agent custom-provider config already stores model metadata under `custom_providers[].models`. The WebUI resolver only read `custom_providers[].reasoning_efforts`, so an unrecognized model such as Inkling could not expose its declared modes without assigning one effort ladder to every model behind the provider.

The precedence is now:

1. Hard route/provider denies.
2. Per-model configured efforts.
3. Provider-wide configured efforts.
4. Existing models.dev, provider-specific, and heuristic detection.

Matching is case-insensitive. An empty or all-invalid model list falls back to provider metadata and then existing detection; it never empties a previously valid picker.

## Existing PR review

- No PR or issue mentioning Inkling was found.
- This is a narrow extension of merged #5313, preserving its config-authoritative behavior, invalid-list fallback, ACP guards, and nested Gemini image/embedding hard denies.
- Open #5987 and #5972 are session/run-scoped effort persistence changes, not model capability discovery.

## Why this scope

Provider-wide metadata is insufficient when one OpenAI-compatible endpoint serves models with different effort ladders. This patch consumes the per-model metadata already present in Hermes config; it adds no new setting, heuristic model family, transport behavior, dependency, or UI interaction.

Example:

```yaml
custom_providers:
  - name: vllm-cwb101
    models:
      inkling:
        reasoning_efforts: [none, low, medium, high, xhigh]
```

## Verification

- `./scripts/test.sh tests/test_reasoning_effort_model_capabilities.py tests/test_custom_provider_bare_model_reasoning.py tests/test_models_dev_reasoning.py tests/test_reasoning_show_hide.py -q` — 104 passed.
- `ruff check --ignore F401,F811 api/config.py tests/test_reasoning_effort_model_capabilities.py` — passed. (`origin/master` currently has four unrelated F401/F811 findings in `api/config.py`.)
- Live config resolution for `inkling` / `custom:vllm-cwb101` returned `supported_efforts: [none, low, medium, high, xhigh]` and `supports_reasoning_effort: true`.
- The live service restarted cleanly and `/health?deep=1` returned `status: ok`.

## Release note

Custom-provider models can declare their own `reasoning_efforts` list, so the composer reasoning picker exposes only the levels supported by the selected model instead of applying one provider-wide list.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -3761,6 +3761,33 @@ def resolve_model_reasoning_efforts(
     return filtered
 
 
+def _configured_reasoning_effort_lists(provider_entry, model_id: str) -> list:
+    """Return model-level then provider-level effort lists from config."""
+    if not isinstance(provider_entry, dict):
+        return []
+
+    configured_lists = []
+    models = provider_entry.get("models")
+    if isinstance(models, dict):
+        model_key = str(model_id or "").strip().lower()
+        model_entry = models.get(model_id)
+        if not isinstance(model_entry, dict) and model_key:
+            model_entry = next(
+                (
+                    metadata
+                    for configured_id, metadata in models.items()
+                    if str(configured_id).strip().lower() == model_key
+                    and isinstance(metadata, dict)
+                ),
+                None,
+            )
+        if isinstance(model_entry, dict):
+            configured_lists.append(model_entry.get("reasoning_efforts"))
+
+    configured_lists.append(provider_entry.get("reasoning_efforts"))
+    return configured_lists
+
+
 def _resolve_model_reasoning_efforts_impl(
     model_id: str | None = None,
     provider_id: str | None = None,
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.140** — thanks @CharlesMcquade! 🎉

The WebUI reasoning picker now honors a  list ahead of the provider-wide list, so a model with its own declared modes exposes them without a provider-level override. Gate: Codex SAFE TO SHIP with differential probes confirming provider-only configs are byte-identical to before (full backward-compat), 73 focused reasoning tests + full serial suite (13,596) green. Your attribution is preserved in the CHANGELOG.

PR #6104, opened 2026-07-15
fix: ensure approval/clarify popups have min-width on narrow screens (#6094)

webtecnica:
Fix for approval and clarify popups collapsing to tiny blank windows on very narrow viewports.\n\n- Added `overflow:visible` to `.clarify-card` on mobile (matching `.approval-card`)\n- Added `min-width:min(100%,280px)` to `.approval-inner` and `.clarify-inner`\n\nCloses #6094

PR #6115, opened 2026-07-16
Release exp: narrow-screen CSS fixes — popup min-width (#6104) + timestamp foldable layout (#6106)

### LINKED ISSUE #6094: WebUI: approval + clarify popups collapse to a tiny blank window on very narrow screens (Z Fold 6 ~344px) — no narrow-width card rule
## Symptom
On an unusually narrow screen (reported on a **Samsung Galaxy Z Fold 6**, folded/cover-display width ~344px, or a narrow split-window), the **Permission Request (Execute Commands)** popup and the **Clarification** popup render as a **tiny blank window with no visible text** — the card collapses so far that its content (description, command, choice buttons) is not visible/usable.

## Platform
Narrow-width clients. Reported on Android (Z Fold 6). Because the approval/clarify cards are **WebUI-rendered** (shared surface), this affects any client at very narrow viewport widths — native Android (Byte), iOS (Hermex), foldables, and split-window — not one specific app. Filed here so the fix propagates to all of them.

## Suspected area (code-traced on current master)
The approval and clarify cards are shared WebUI CSS in `static/style.css`:
- `.approval-card` / `.clarify-card` use `max-width:var(--msg-max); width:100%; padding:0 20px; box-sizing:border-box` with an absolutely-positioned slide-up `*-inner`.
- Mobile handling lives in the `@media (max-width:640px)` block (e.g. `.approval-btns` becomes a 2-column grid, height caps `min(52vh,360px)` / `min(60dvh,420px)`).
- **There is no dedicated very-narrow (≤ ~360px) rule for the approval/clarify cards.** The only `@media (max-width:340px)` rule targets the composer, not these cards. At Z Fold folded width the 2-column button grid + the height caps + the fixed horizontal padding appear to collapse the card to an empty strip.

This is the **shared-WebUI analog** of the already-fixed WebView-wrapper bug `hermes-webui/hermes-android#44` ("Approval card collapses to a header-only strip in Android WebView", closed via #46) — but that fix lived in the (now dormant) wrapper repo, not in the shared WebUI CSS.

## Related / distinct
- **#5385** (closed) — approval popup *too small, options require scrolling*. Related but a **milder variant**: there the content rendered but was cramped; the #5385 fix (`39a4d67e5`) only *raised the mobile height cap* (`min(52dvh,360px)`→`min(60dvh,420px)`). It did **not** add any narrow-*width* handling, so it does not cover this blank-collapse at ~344px.
- **hermes-webui/hermes-android#44 / #46** (closed) — same *symptom* (header-only collapse) but fixed in the WebView wrapper, not the shared WebUI.
- **ComputerByte/hermex-android#12** (open) — different root cause (native SSE parser drops clarify/approval events entirely); not a rendering/width issue.

## Repro
1. Open the WebUI on a very narrow screen (Z Fold 6 folded, or resize a window / split-view to ≤ ~360px CSS px).
2. Trigger a command/tool **Permission Request** or a **Clarification** prompt.
3. Observe the popup renders as a tiny blank window — no visible description/command/choices.

## Suggested fix direction
Add a very-narrow (`≤ ~360px`) rule for `.approval-card` / `.clarify-card` / `*-inner`: reduce horizontal padding, force `.approval-btns` to a single column at that width, and ensure a sensible `min-height` / non-collapsing layout so the header + description + choices remain visible and scrollable rather than collapsing. Mirror the intent of the wrapper's #46 fix in the shared CSS.

---
_Filed from a Discord community report (privacy-safe paraphrase). Edit, re-route, or close if mis-scoped._

nesquena-hermes:
Release exp: narrow-screen CSS fixes — approval/clarify popup min-width (#6104) + stream-timestamp foldable layout (#6106)

## What ships (@webtecnica)
- **#6104** (#6094) — approval/clarify popups collapsing to an unreadable sliver on very narrow viewports; the clarify card now gets a `min-width` + `overflow:visible` matching the approval card.
- **#6106** (#6099) — stream-event timestamps spilling the tool/thinking card header row on <360px foldables; the headers now clip overflow.

## Maintainer scoping (gate-driven)
Both source branches had bundled in extra rules that the gates flagged; stripped so this batch is exactly the two stated fixes:
- Removed both `<=360px .transparent-event-row .tool-card-name{display:none}` rules — they hid tool identity even when expanded; the header `overflow:hidden` + the pre-existing `<=420px` timestamp-hide already prevent the row spill without dropping the name.
- Removed the `#composerModelDropdown{position:fixed}` mobile overrides that #6104's branch had bundled — that's the #6080 mobile-model-picker mechanism, and `position:fixed` there is trapped by `.composer-footer`'s `container-type` so it doesn't actually escape the footer. The model picker is being reworked separately (reparent to body level, the `#profileDropdown` idiom). The dropdown here reverts to its base `position:absolute`, matching `origin/master`.

## Gate
- Codex: **SAFE TO SHIP** (cards contained at 280px; header clip preserves tool identity + all controls; model dropdown reverts cleanly to absolute; no regression).
- Fable UX gate: reviewed the clarify/approval + timestamp surfaces — no UX objection (the UX-REWORK verdict was scoped to the separate #6105 model-picker mechanism).
- Browser-smoke: clean.

Thanks @webtecnica. Closes #6094, #6099.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,10 @@
 
 ### Fixed
 
+- **Approval and clarify popups no longer collapse to tiny blank windows on very narrow screens.** On very narrow viewports the clarify card could shrink to an unreadable sliver; it now gets a `min-width` and `overflow:visible` matching the approval card. Thanks @webtecnica. (#6104, #6094)
+
+- **Stream event timestamps no longer break the row layout on narrow foldable screens.** On screens under 360px the per-event timestamps could spill out of tool/thinking card headers; the headers now clip overflow so the timestamp layout holds. (Tool names stay visible — an earlier draft hid them at ≤360px, which lost tool identity; the header clip alone prevents the spill.) Thanks @webtecnica. (#6106, #6099)
+
 - **Streaming sidebar polls no longer re-run the expensive CLI/cron projection every few seconds.** The per-stream cache TTL was pinned at 30s while the frontend's streaming sidebar poll cadence could approach it, so the state.db CLI/cron projection (and its ~200 sidecar reads) could be re-paid on nearly every poll during a live turn (the #4842/#4808 high-CPU `get_cli_sessions` case). The streaming TTL is now held strictly above the sidebar poll interval (45s) and paired with the stable streaming cache key, so repeated polls reuse the projection; a new test asserts the TTL stays above the poll cadence so this can't silently regress. Thanks @starship-s. (#6102, #4842)
 
 - **Delegated subagent sessions in the sidebar now show a meaningful title instead of "Subagent Session".** Generic delegated child sessions previously all displayed the placeholder "Subagent Session" in the conversations list, making them indistinguishable. Their sidebar title is now derived (view-only) from the child's first user message during the existing batched sidebar-metadata pass — without mutating the stored session title, so the read-only boundary is preserved. Only generic "Subagent Session" placeholders are affected; custom titles are left untouched. Thanks @rodboev. (#6056, #6033)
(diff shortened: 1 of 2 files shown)


webtecnica:
Thanks @nesquena! Glad to see #6104 and #6106 land in the release. 🚀

PR #6122, opened 2026-07-16
fix(webui): prevent settled async result hover flicker

### LINKED ISSUE #6072: WebUI: hover-triggered flicker/text-shading on follow-up text after an async agent return (exp-v0.52.71)
Reported by **b3nw** in Discord `#report-bugs` (Jul 14 2026) on **exp-v0.52.71 · Experimental**.

> "on the flicker problem: after async agent returns text to a session, follow up text flickers or has changing text shading when mouse moves over it."

## Symptom

A **hover-triggered** variant of the streaming flicker. After an **async agent turn returns text** to a session (e.g. a background/delegated/async run posting its result), the follow-up text **flickers / changes shading when the mouse moves over it**. So this isn't the during-generation flicker of #5367/#5943 — it's a post-settle, **hover-induced** re-shade/repaint on the async-returned content.

## Likely area

- A global CSS transition applies to `div/…` for `background-color, border-color, color` at `.15s` (`static/style.css` ~L1477). Combined with a `:hover` background/color rule on the message row/body, an async-returned message that wasn't painted with the same settled state as a normal turn can visibly re-shade on hover (the transition animates a color/background delta that a fully-settled turn wouldn't have).
- Async/background-return content may be rendered via a different path than the normal live→settled turn (so it carries a slightly different DOM/class state), which is why hover re-triggers a shading change specifically on **follow-up async text** rather than on normal turns.

Suspect the async/background turn-render + the settled hover styling interaction in `static/messages.js` / `static/ui.js` + the `:hover` + global color-transition rules in `static/style.css`.

## Relationship to the flicker family
b3nw frames it as "on the flicker problem" — same family as #5367 (CLOSED, Transparent Stream flicker) and #5943 (flicker returned). But the trigger here is **mouse hover after an async return**, not live streaming, so it's a distinct vector worth its own trace (and a regression test that hovers a settled async-returned turn and asserts no repaint/shading change).

## Repro (reporter's description)
1. Have an async/background agent turn return text into a session.
2. Move the mouse over the follow-up text.
3. Observe flicker / changing text shading on hover.

## Acceptance criteria
- An async-returned (settled) turn does not flicker or change shading on hover; its settled paint matches a normal turn's.

## Environment
- WebUI: exp-v0.52.71 (experimental)

## Related
- #5943 — response flickering during streaming returned (regression of #5367)
- #5367 — original Transparent Stream flicker (CLOSED)

coyaSONG:
## Thinking Path

The async/background completion path appends an ordinary, settled assistant message and renders it through the settled assistant-turn surface. The message is no longer live, but its `.msg-body` still inherits the global `div` transition for `background-color`, `border-color`, and `color`. When hover chrome repaints the turn, that residual 150 ms transition can visibly re-shade the returned text.

This change keeps semantic ownership and transcript rendering unchanged. It only makes the settled assistant prose paint layer immediate, matching the existing live-turn exclusions.

Fixes #6072.

## What Changed

- Added settled assistant message bodies to the existing `transition: none !important` exclusion in `static/style.css`.
- Added a Playwright regression that renders the settled async-result shape and checks desktop, narrow, and mobile viewports.
- The regression verifies that hover produces no color transition or animation and that text/background/border paint values remain stable before, during, and after hover.

## Why It Matters

Async agent results should remain visually stable once they enter the settled transcript. Removing the inherited paint transition eliminates the hover-triggered flicker without changing message state, hover actions, theme transitions elsewhere, or the live-to-final rendering contract.

## Verification

- Red before the CSS change: all 3 viewport cases failed with `transition-property: background-color, border-color, color` and `transition-duration: 0.15s`.
- Green after the CSS change: all 3 viewport cases passed with `transition-property: none`, `transition-duration: 0s`, zero animations, and stable paint values.
- Targeted regression/neighbor suite: `73 passed`.
- Full repository test wrapper: `12994 passed, 148 skipped, 2 xfailed, 1 xpassed, 12 warnings, 34 subtests passed`.
- Ruff on the new Python test: all checks passed.
- Browser smoke checks for `/`, `/#settings`, and `/#sessions`: passed with zero console errors.
- `git diff --check`: passed.
- `npm run lint:runtime` was not available in this checkout because ESLint is not installed; this PR does not modify JavaScript.

Synthetic fixture screenshots used for deterministic paint-state review:

| Viewport | Before | After |
| --- | --- | --- |
| Desktop | Before desktop | After desktop |
| Narrow | Before narrow | After narrow |
| Mobile | Before mobile | After mobile |

## Risks / Follow-ups

- The selector is limited to direct `.msg-body` children of settled assistant segments; existing live-turn exclusions remain unchanged.
- The screenshots use a deterministic synthetic async-result fixture. A maintainer should also confirm the experimental async-agent path described in #6072 in their configured environment.
- The fork-only evidence branch can be deleted after this PR is accepted or closed.

## Model Used

OpenAI Codex, GPT-5 (this pull request was produced with AI assistance).

Tools used: GitHub CLI, the repository test wrapper, and Playwright/Chromium.


coyaSONG:
Closing this as superseded rather than leaving a duplicate draft open. PR #6166 delivered the same hover-flicker fix, PR #6282 shipped it, and issue #6072 is closed with current-source and release evidence. The branch remains preserved for audit history; it should only be reopened if a distinct regression is demonstrated.

PR #6124, opened 2026-07-16
fix(streaming): classify final answers by result provenance

ildunari:
## Summary

- classifies a settled assistant answer from the current agent result, not only from the reconstructed display transcript
- rejects historical assistant rows replayed from display/context history by stable row ID
- preserves fail-closed handling for legacy ID-less history and structured assistant content without final visible text

This supersedes the current approach in #5902 and implements the provenance-aware fix requested in its formal review.

## Root cause

A current user row may already be checkpointed in `previous_display` before turn settlement. In that geometry, the merged-transcript evaluator can conservatively report that the turn still ends at the user row even when the provider result contains a real new final assistant row. The silent-failure fallback then emits `no_response` after the completed answer.

Simply moving the positional boundary is unsafe: a repeated prompt plus context backfill can place a stale historical assistant row after that boundary. That is the blocker identified on #5902.

The settlement gate now combines two signals:

1. the existing merged-transcript evaluator remains fail-closed;
2. a final answer can override it only when the current result contains a usable assistant row whose stable row ID is absent from both pre-turn display and context history.

Legacy history without row IDs falls back conservatively to content identity because settlement may just have minted a new ID onto a replayed old row.

## State/invariant

State layer: the live-to-final run settlement in `api/streaming.py`.

Invariant: a completed provider answer must settle once without a trailing synthetic `no_response`; a replayed/backfilled historical answer must never make an answerless current turn appear successful. Explicit terminal failures, cancellation, compression exhaustion, and tool-limit exits retain their existing gates.

## Verification

- `113 passed` across:
  - terminal-failure transcript evaluation
  - cancellation
  - streaming persistence
  - compression terminal failures
  - tool-limit terminal state
  - transcript compaction
- focused test file passes Ruff
- `py_compile` passes
- `git diff --check` passes
- live isolated Codex smoke:
  - SSE: tokens → one `done` → one `stream_end`
  - no `apperror` / `no_response`
  - persisted transcript contains exactly one user row and one final assistant row, with no `_partial` or `_error` rows

The full `api/streaming.py` Ruff scan has the same seven unrelated baseline findings as `origin/master`; this change introduces none.

## Regression coverage

Covers:

- checkpointed current prompt plus a genuinely new final result row
- repeated prompt plus replayed historical answer
- legitimate text-identical answer with a distinct stable row ID
- context-only historical replay
- structured content with no final visible text
- legacy ID-less replay that receives a freshly minted result ID

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -5617,8 +5617,19 @@ def _stamp_missing_message_timestamps(messages, *, now: float | None = None) -> 
     return stamped
 
 
-def _assistant_reply_added_after_current_turn(result_messages, previous_context, msg_text) -> bool:
-    """Return True only when the just-finished turn produced assistant text."""
+def _assistant_reply_added_after_current_turn(
+    result_messages,
+    previous_context,
+    msg_text,
+    *,
+    previous_display=None,
+    drop_replayed_assistant: bool = False,
+) -> bool:
+    """Return True only when the just-finished turn produced assistant text.
+
+    When a terminal/partial result may contain replayed history, an assistant
+    row already present before this turn is not proof of a new final answer.
+    """
     result_messages = list(result_messages or [])
     previous_context = list(previous_context or [])
     if _messages_have_prefix(result_messages, previous_context):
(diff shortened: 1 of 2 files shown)


nesquena-hermes (changes requested):
Thanks @ildunari — the provenance-aware direction is right (it's the approach #5902's review asked for), and the safe cases hold: all 34 existing settlement / tool-limit / structured-content tests pass including the #6081 tool-limit-final-answer path. But the deep gate found the fix over-corrects — it now lets genuine no-responses through as `done`, which is *worse* than the original false-positive (the user thinks the model answered when it didn't). Two CORE, both on `api/streaming.py`:

1. **An answerless turn can emit `done` instead of `no_response`** (`api/streaming.py:5799`). Any visible assistant candidate now overrides the transcript evaluator — *including tagged synthetic/interim text followed by a tool tail*. A focused `_run_agent_streaming` probe produced `done` while the (unchanged, correct) transcript evaluator returned `lacks_final_answer=True`. So a real no-response is suppressed. **Fix:** make the override **terminal-tail-aware** — remove synthetic/control rows, and require the provenance-filtered current-result slice to *end* in a usable assistant answer; a later user/tool/tool-call row must veto earlier assistant text (you can't call it `done` if the turn ended on a tool with no final answer).

2. **Compacted history-only results can treat an OLD assistant row as the current answer** (`api/streaming.py:5638`). `_find_current_user_turn()` falls back to an unrelated historical user when the current prompt is absent (post-compaction); positional ID restoration then fails and lines 5657-5660 accept freshly re-minted historical IDs, so a stale assistant row establishes false provenance (verified `(True, False)` even with replay filtering on). **Fix:** non-prefix provenance must require a **strict current-user match**, and historical-row alignment must happen **before** minting missing IDs — a freshly assigned ID must never be able to establish provenance.

The unifying issue: "has a visible assistant candidate" isn't sufficient for `done` — it must be the *current turn's terminal answer*, and provenance must be anchored to a strict current-user match, not a post-compaction fallback. Once the override is terminal-tail-aware + excludes synthetic rows, and non-prefix results require a strict current-turn anchor with alignment-before-ID-minting, re-request and I'll re-run the full gate + stream-regression gate. (Note: this supersedes #5902 — once yours lands clean I'll close #5902 as superseded, crediting @soria-clawd-bot for the original diagnosis.)

Suppressing a genuine no-response is the one outcome worse than the bug you're fixing, so I'm holding to the strict bar here — but the shape is right and you're close.

franksong2702:
Independent gate confirmation on current head `084a0a43f`:

- The focused settlement / persistence / tool-limit / compression / cancellation suites pass (`91 passed`).
- The two CORE cases from the formal review still reproduce directly against `_settled_turn_answer_state`:
  1. a synthetic assistant row followed by a tool tail returns `(True, False)`, so an answerless turn can suppress `no_response`;
  2. a compacted history-only result without the current prompt, after missing IDs are minted, also returns `(True, False)`, so an old assistant row can establish false current-turn provenance.

The narrow repair gate I would use is:

1. Filter synthetic/control rows before deriving answer provenance.
2. For a non-prefix result, require a strict current-user match; do not use `_find_current_user_turn()`'s unrelated-user fallback as provenance.
3. Make the result override terminal-tail-aware: after filtering controls, the current-turn result slice must end in a usable assistant answer. A later user/tool/tool-call row must veto an earlier assistant candidate.
4. Align historical rows before minting missing IDs, or carry explicit pre-mint provenance so a newly assigned ID cannot prove freshness.

Please add focused regressions equivalent to:

- `synthetic/interim assistant -> tool tail -> no_response`
- `compacted history without strict current-user anchor -> old assistant replay -> no_response`

Once those hold, the provenance-aware direction remains the right replacement for #5902 and I can re-run the focused gate.

PR #6127, opened 2026-07-16
Live Stream: route list-shaped custom provider models

### LINKED ISSUE #6121: bug(models): custom_providers[].models list-shaped allowlist is ignored by resolve_model_provider (dict-only) — selected model silently falls back to a different provider (follow-up to #5511)
## Summary

`resolve_model_provider()` reads `custom_providers[].models` **only when it is a `dict`**, but the picker-population code and the sibling `providers.<slug>.models` reader both accept a **`list`**. With a list-shaped `custom_providers[].models` allowlist, the models appear in the dropdown but request-time routing never matches them — the bare id inherits the default provider, hits an endpoint that doesn't have it, and **silently falls through `fallback_providers` to a completely different model**.

This is the same symptom class fixed for the `providers:` block in **#5511** ("a bare model id from that list wasn't matched by request-time routing — it silently fell back to the default provider and often 404'd"). The `custom_providers[]` sibling reader was left dict-only.

Verified against `origin/master` @ `1845e639` (latest `## [v0.52.41]` in CHANGELOG).

## Repro

`~/.hermes/config.yaml` — default provider is Anthropic, plus a local Ollama endpoint with a **list**-shaped allowlist:

```yaml
model:
  default: claude-haiku-4-5
  provider: anthropic
fallback_providers:
- provider: gemini
  model: gemini-2.5-pro
custom_providers:
- name: storm-ollama
  base_url: http://<host>:11434/v1
  api_key: ollama-local
  models:                 # list — as accepted by the population code
  - qwen3.6:35b-a3b
  - violet-lotus:latest
```

Select **Qwen3.6** in the picker and send a message.

**Expected:** the turn runs `qwen3.6:35b-a3b` on the Ollama endpoint.
**Actual:** the turn silently runs **Gemini**. `state.db` `sessions` confirms `model='gemini-2.5-flash'`, `billing_provider='gemini'`.

## Evidence

```python
>>> import api.config as C
>>> C.resolve_model_provider("qwen3.6:35b-a3b")
('qwen3.6:35b-a3b', 'anthropic', None)      # ← allowlist ignored; inherits default provider
>>> C.resolve_model_provider("@custom:storm-ollama:qwen3.6:35b-a3b")
('qwen3.6:35b-a3b', 'custom:storm-ollama', None)   # ← works (the @-prefix path)
```

Anthropic has no `qwen3.6:35b-a3b`, the call errors, and the `fallback_providers` chain quietly swaps in Gemini. Nothing in the UI indicates the selected model never ran — the user only finds out by asking the model what it is, or by reading `state.db`.

## Root cause — three readers, inconsistent shape handling

All in `api/config.py` on `master`:

| Site | Reads `models` as | Note |
|---|---|---|
| ~L2667 `resolve_model_provider` → `providers.<slug>.models` | `list` **and** `dict` | fixed in #5511 |
| ~L2696 `resolve_model_provider` → `custom_providers[].models` | **`dict` only** | ← the bug |
| ~L7135 `get_available_models` → `custom_providers[].models` | `list` **and** `dict` | why it still shows in the picker |

The dict-only branch:

```python
entry_models = entry.get('models')
if isinstance(entry_models, dict):
    entry_model_ids.update(
        key.strip()
        for key in entry_models.keys()
        if isinstance(key, str) and key.strip()
    )
if entry_name and model_id in entry_model_ids:
```

Because population accepts a list and resolution does not, a list-shaped allowlist puts models in the dropdown that **cannot** route. The `#5511` fix ~40 lines above already has the exact list+dict handling this branch needs.

## Suggested fix

Give the `custom_providers[]` reader the same list+dict handling as the `providers.<slug>` reader directly above it:

```python
entry_models = entry.get('models')
if isinstance(entry_models, dict):
    entry_model_ids.update(
        k.strip() for k in entry_models if isinstance(k, str) and k.strip()
    )
elif isinstance(entry_models, list):
    for _m in entry_models:
        _mid = str(_m.get('id') or _m.get('model') or _m.get('name') or '') if isinstance(_m, dict) else str(_m or '')
        if _mid.strip():
            entry_model_ids.add(_mid.strip())
```

Factoring the #5511 block into one shared helper used by all three sites would keep them from drifting apart again.

## Secondary: the silent fallback is what makes this expensive

The routing bug is a one-line shape mismatch, but what turned it into a long debugging session is that **a selected model that fails to route falls back silently**. The picker keeps showing Qwen while Gemini answers, and the reply is plausible, so there's no signal anything is wrong.

Worth considering: when an **explicitly selected** model fails, surface it rather than silently substituting a fallback — or at least emit a lifecycle warning naming the model that actually ran. A fallback is right for a transient rate-limit on the *default* model; it's misleading when the user has explicitly picked a specific one. Related in spirit: #1426 (silent failure in model selection), #5979 (namespace stripping breaks custom routing).

## Workaround

Drop the `models:` allowlist entirely so live `/v1/models` discovery populates the group — discovered ids get the working `@custom:<slug>:` prefix. (Note this is only a workaround because a non-empty allowlist *also* disables the live probe, so a newly-pulled model is invisible until hand-added — which is how the bad state arises in the first place.)

franksong2702:
Closes #6121.

## Thinking Path

- The model picker already accepts `custom_providers[].models` as either a mapping or a list.
- Runtime routing only read mapping keys, so a model selected from a list-shaped allowlist could silently remain on the active/default provider.
- The same config shape was parsed independently in several catalog and routing paths, which made picker/runtime drift possible.
- This PR centralizes supported model-ID extraction and uses it for both catalog construction and provider resolution.

## What Changed

- Added one ordered, deduplicating parser for mapping, string-list, and object-list model allowlists.
- Added a picker-option row builder on top of that parser so catalog paths preserve explicit labels while sharing the same accepted model ID shapes.
- Reused the shared parser/option builder in named custom-provider resolution, configured provider ownership checks, and static/live model catalog construction.
- Added regressions for the exact string-list shape from #6121, for picker-supported object entries using `id`, `model`, or `name`, and for the sibling live-catalog `providers.<slug>.models` object-list path.

## Why It Matters

A picker-visible model now routes to the named custom endpoint that declared it instead of silently falling through to an unrelated active provider and potentially invoking a different model.

## Contract Routing

- Contract family: model selection and provider routing from `config.yaml`.
- State layer: configuration read path only; this PR does not mutate persisted state.
- Preserved invariant: every supported model allowlist shape exposed by the picker must resolve to the same owning provider at runtime.
- Evidence: #6121, existing dict-shaped allowlist tests, and the focused routing/catalog regression suite.
- Contract change: none. This restores parity between already-supported picker input and runtime routing.

## Verification

- Confirmed the two original #6121 tests fail on current `master` with the provider incorrectly remaining `anthropic`.
- Confirmed `test_provider_config_object_list_catalog_uses_picker_supported_keys_6121` fails before the follow-up parser fix with `KeyError: 'id'` in the live provider-catalog parser.
- `./scripts/test.sh tests/test_model_resolver.py -k provider_config_object_list_catalog_uses_picker_supported_keys_6121` (`1 passed` after the fix)
- `./scripts/test.sh tests/test_model_resolver.py -k 6121` (`3 passed` after the fix)
- `./scripts/test.sh tests/test_model_resolver.py` (`62 passed`)
- `env -u HERMES_WEBUI_DEFAULT_MODEL ./scripts/test.sh tests/test_model_resolver.py tests/test_issue1625_local_server_model_id_preservation.py tests/test_issue1806_named_custom_provider_resolution.py tests/test_issue1894_provider_overlap.py tests/test_issue3957_profile_providers_models.py tests/test_byok_model_dropdown.py tests/test_provider_mismatch.py` (`239 passed`)
- `.venv/bin/python -m py_compile api/config.py tests/test_model_resolver.py`
- `.venv/bin/ruff check api/config.py tests/test_model_resolver.py --ignore F401,F811`
- `git diff --check`

Full-file Ruff without ignores still reports four pre-existing `F401`/`F811` import findings in `api/config.py`; this PR does not touch those imports.

## Risks / Follow-ups

- Risk is limited to configured model allowlist parsing. Mapping keys retain their existing semantics, and Copilot's settings map remains explicitly excluded from provider allowlist routing.
- No UI changes or migration are required.

## Release Note

Fixed list-shaped custom-provider model allowlists so selected models route to the endpoint that declared them.

## Model Used

- Provider: OpenAI
- Model: GPT-5 via Codex (the exact runtime model ID is not exposed in this environment)
- Notable tool use: repository inspection, focused regression tests, and GitHub CLI/API publishing

AI assistance was used to investigate, implement, test, and prepare this pull request.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -1372,6 +1372,25 @@ def _configured_model_ids(raw_models: object) -> list[str]:
     return model_ids
 
 
+def _configured_model_options(raw_models: object) -> list[dict[str, str]]:
+    """Return picker option rows from supported config allowlist shapes."""
+    labels: dict[str, str] = {}
+    if isinstance(raw_models, list):
+        for item in raw_models:
+            if not isinstance(item, dict):
+                continue
+            candidate = item.get("id") or item.get("model") or item.get("name")
+            model_id = str(candidate or "").strip()
+            if not model_id or model_id in labels:
+                continue
+            label = str(item.get("label") or model_id).strip() or model_id
+            labels[model_id] = label
+    return [
+        {"id": model_id, "label": labels.get(model_id, model_id)}
+        for model_id in _configured_model_ids(raw_models)
+    ]
+
+
 def _named_custom_provider_slugs(config_obj: dict | None = None) -> set[str]:
     return {
         slug
@@ -5212,28 +5231,7 @@ def _static_models_catalog_without_live_probes() -> dict:
             provider_cfg = _get_provider_cfg(raw_key)
             raw_models = []
             if isinstance(provider_cfg, dict) and "models" in provider_cfg:
-                cfg_models = provider_cfg["models"]
-                if isinstance(cfg_models, dict):
-                    raw_models = [{"id": key, "label": key} for key in cfg_models.keys()]
-                elif isinstance(cfg_models, list):
-                    raw_models = []
-                    for item in cfg_models:
-                        if isinstance(item, dict):
-                            model_id = (
-                                item.get("id")
-                                or item.get("model")
-                                or item.get("name")
-                            )
-                            if not model_id:
-                                continue
-                            raw_models.append(
-                                {
-                                    "id": model_id,
-                                    "label": item.get("label", model_id),
-                                }
-                            )
-                        elif item:
-                            raw_models.append({"id": item, "label": item})
+                raw_models = _configured_model_options(provider_cfg["models"])
             if not raw_models:
                 raw_models = copy.deepcopy(_PROVIDER_MODELS.get(pid, []))
             # Plugin-only providers (e.g. 9router) are not in _PROVIDER_MODELS
@@ -7566,13 +7564,7 @@ def get_available_models(*, prefer_cache: bool = False, force_refresh: bool = Fa
                         and isinstance(provider_cfg, dict)
                         and "models" in provider_cfg
                     ):
-                        cfg_models = provider_cfg["models"]
-                        if isinstance(cfg_models, dict):
-                            raw_models = [{"id": k, "label": k} for k in cfg_models.keys()]
-                        elif isinstance(cfg_models, list):
-                            raw_models = [{"id": k["id"] if isinstance(k, dict) else k,
-                                            "label": k.get("label", k["id"]) if isinstance(k, dict) else k}
-                                           for k in cfg_models]
+                        raw_models = _configured_model_options(provider_cfg["models"])
 
                     if not raw_models:
                         if pid == "moa":
--- a/api/config.py
+++ b/api/config.py
@@ -1351,6 +1351,27 @@ def _custom_provider_entries(config_obj: dict | None = None) -> list[dict]:
     return [entry for entry in entries if isinstance(entry, dict)]
 
 
+def _configured_model_ids(raw_models: object) -> list[str]:
+    """Return ordered model IDs from supported config allowlist shapes."""
+    if isinstance(raw_models, dict):
+        candidates = (key for key in raw_models if isinstance(key, str))
+    elif isinstance(raw_models, list):
+        candidates = raw_models
+    else:
+        return []
+
+    model_ids: list[str] = []
+    for item in candidates:
+        if isinstance(item, dict):
+            candidate = item.get("id") or item.get("model") or item.get("name")
+        else:
+            candidate = item
+        model_id = str(candidate or "").strip()
+        if model_id and model_id not in model_ids:
+            model_ids.append(model_id)
+    return model_ids
+
+
 def _named_custom_provider_slugs(config_obj: dict | None = None) -> set[str]:
     return {
         slug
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
Shipped in **exp-v0.52.85** 🚀 — thanks @franksong2702.

Your fix is live on the experimental channel: models selected from a **list-shaped** `custom_providers[].models` allowlist now route to the correct provider instead of silently staying on the active/default one (#6121). Centralizing the supported-model-ID extraction into one helper used by both catalog construction and runtime routing means the picker and resolver can't drift apart anymore.

Gate before ship: Codex adversarial review returned **SAFE TO SHIP** — verified the refactor is behavior-preserving for existing mapping-shaped and list-of-strings configs, list-shaped models now route correctly, order/dedup/labels are preserved, malformed entries (None, non-dict, missing id) are handled safely, the Copilot settings-map exclusion stays intact, and streaming receives the resolved endpoint (including colon-containing model IDs). Full suite 13,140 passed, CI green.

Issue #5700, opened 2026-07-06
feat(worklog): persistent + live per-event timestamps on Transparent Stream activity rows


nesquena-hermes:
## Summary

Add **wall-clock timestamps to Worklog / Transparent Stream activity rows** so users can see *when* each thinking / tool / text event happened — both during a live turn and after it settles.

## Motivation

After the Transparent Stream activity display landed, the most-requested polish from testers is per-event timestamps. The concrete pain point (reporter's words): during a running turn *"time ticks by and you see no updates and you start to wonder how long ago was the last tool call or update."* Timestamps on the event rows make it obvious whether the agent is progressing or genuinely stalled.

## Current state / gaps

Two specific gaps the reporter identified:

1. **Post-turn timestamps are hover-gated.** The timestamps that exist today only appear on hover, and only *after* the turn has ended — *"the timestamps disappear unless you move the mouse over things (after turn ends)."* They should be persistently visible (not hover-only) on settled turns.
2. **No timestamps at all during a running stream** — *"there aren't any during running-turn streaming,"* which is exactly when they're most useful.

So the ask splits into:
- **(a)** persistent (always-visible, not hover-gated) timestamps on **settled** worklog/activity rows, and
- **(b)** **live** timestamps on streaming rows during a running turn.

## Suspected area

`static/ui.js` Transparent Stream / Worklog event-row render path (the per-event row builder for thinking / tool / text rows) + the settled-turn footer/label helpers (`_activityProcessedElapsedLabel` / `_activitySettledProcessedLabel`). Elapsed-time display already exists on the settled footer (#4078-era); this extends per-row.

## Notes

- Reporter (b3nw) mentioned he has a **partial/half-baked PR in progress** for this, so a contributor PR may land against this issue.
- Prior related: #4078 (turn-level elapsed time). This issue is per-event-row timestamps, distinct from the turn-level elapsed footer.
- Respect the existing activity-display-mode toggles; live timestamps should not add flicker to the (recently-stabilized) streaming render path.

## Severity

P3 — polish / observability enhancement on the transparent-stream surface.

---
_Filed from repeated requests in Discord #report-bugs (b3nw, 2026-06-13 + 2026-06-14 follow-up)._

nesquena-hermes:
[...]

PR #6130, opened 2026-07-16
fix(#6099): make Transparent Stream activity timestamps optional

### LINKED ISSUE #6099: WebUI: Transparent Stream event timestamps on narrow foldables — wrap+break row height, inconsistent left/right position per row type, and no toggle to hide them (exp)

rodboev:
## Thinking Path

- The issue started as a foldable layout report, but the maintainer's July 16, 2026 browser follow-up kept only one verified gap alive on current `master`: there is still no way to hide Transparent Stream's per-event timestamps independently.
- The response footer already has its own `.msg-time`, so the missing behavior is not "show any timestamp" versus "show none". It is "keep Transparent Stream, but make the per-event chips optional."
- The fix adds that missing preference through the existing settings path and gates only the event chip render, leaving the display mode contract and footer timestamp intact.

## What Changed

- `api/config.py`: add a persisted preference for Transparent Stream event timestamp visibility.
- `static/index.html`, `static/panels.js`, and `static/i18n.js`: expose the new setting, route its visible copy through the existing `en` fallback, and save and restore it without changing the existing activity display mode choices.
- `static/ui.js`: suppress `.transparent-event-time` when the new preference is off, while preserving row content and footer `.msg-time`.
- `tests/test_issue5700_worklog_event_timestamps.py` and `tests/test_issue3820_chat_activity_display_mode.py`: cover the missing toggle, the preserved display-mode behavior, and the i18n/sync follow-up.

## Why It Matters

Users can keep Transparent Stream's detailed chronology without the per-event timestamp chips that the issue thread found noisy and redundant. The assistant footer timestamp still remains, so this narrows the visual noise without removing timing information entirely.

## Verification

The branch's `#5700` browserless harness was pointed at both current `master` and the branch `static/ui.js`. It returned `{"base":{"count":1,"eventAt":"1700000910","source":"event","footerCount":1,"footerText":"10:15 PM"},"head":{"count":0,"eventAt":"1700000910","source":"event","footerCount":1,"footerText":"10:15 PM"}}`, which is the exact missing-toggle reproduction and fix. Focused branch coverage then ran `tests/test_issue5700_worklog_event_timestamps.py::test_transparent_event_timestamp_toggle_hides_chip_but_preserves_event_anchor` and the full `tests/test_issue3820_chat_activity_display_mode.py` file in the hidden runner and passed clean (`Running 48 items in this shard` / `48 passed in 9.78s`). The runtime JS guard also stayed clean with `npx eslint --no-config-lookup -c eslint.runtime-guard.config.mjs "static/**/*.js"`. CI runs the full Python matrix.

## Risks / Follow-ups

The foldable layout and left/right alignment claims are intentionally out of scope here. Upstream merged `#6106` for that slice, and the maintainer's July 16, 2026 follow-up did not reproduce a remaining layout defect on current `master`.

Release note: Transparent Stream can now hide per-event timestamp chips while keeping the response footer time visible.

The before/after pair was captured from isolated throwaway servers at `1024x600`. I did not get a clean served-page `480x320` screenshot before the local headless navigation started stalling on the throwaway branch server, so the narrow-state evidence here is the focused behavior coverage plus the unchanged compact shell layout in the screenshot harness.

## Contract Routing

Task type: narrow rendered-activity preference fix
Touched areas: settings persistence, Transparent Stream event-chip rendering, and activity-mode regression coverage
Relevant public docs:
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `docs/GUIDELINES.md`
Scope boundaries: no change to the three existing activity-display modes, no footer timestamp removal, no foldable layout work, no left/right stream alignment changes
Evidence needed before claiming done: one base/head browserless regression, focused activity-mode preservation coverage, runtime JS lint, and a before/after screenshot pair from isolated servers

## Contract Change

Previous contract: Transparent Stream always rendered a per-event timestamp chip whenever a timestamp label existed, and users could only choose among the three coarse activity display modes.
New contract: Transparent Stream keeps the same three activity display modes and default behavior, but adds a persisted subordinate toggle that hides only the per-event timestamp chips while preserving the response footer `.msg-time`.
Affected docs and tests: this PR body, `tests/test_issue5700_worklog_event_timestamps.py`, and `tests/test_issue3820_chat_activity_display_mode.py`
Compatibility or migration reason: the new setting defaults to the old behavior, so existing users keep visible chips until they opt out

## Upstream

Closes #6099.

## Screenshots

Before: current `master` still shows the Transparent Stream per-event timestamp chip and has no independent checkbox for it.

Before

After: the new checkbox is off, the per-event chip is gone, and the response footer time remains visible.

After

## Model Used

GPT-5 via Codex CLI


rodboev:
Follow-up push on July 16, 2026: I moved the new checkbox label and description onto the existing `en` i18n fallback path in `static/i18n.js`, added the corresponding `data-i18n` hooks in `static/index.html`, removed the duplicate `_syncTransparentEventTimestampsControl(...)` call in `loadSettingsPanel()`, and reran the focused local proof (`48 passed in 5.60s`) plus the runtime JS guard clean.

nesquena-hermes:
Reviewed the diff against `origin/master` (pulled the branch into a read-only worktree). The shape is clean: `transparent_stream_event_timestamps` defaults to `true` in `api/config.py:8918` and is registered in `_SETTINGS_BOOL_KEYS`, `ui.js` moves the `data-event-at` row attributes above the early return so the anchor is always set, and the chip suppression is gated correctly. The greptile i18n/duplicate-sync nits are already handled in your follow-up push. One real gap remains.

## Boot-time seeding is missing

`window._transparentEventTimestamps` is only ever assigned inside `_syncTransparentEventTimestampsControl` (`static/panels.js:8461`), which runs on Settings-panel load and on save/autosave — but **not on initial page boot**. The render gate reads the global directly (`static/ui.js:6164`):

```js
const showEventTimestamp=!(typeof window!=='undefined'&&window._transparentEventTimestamps===false);
```

On a fresh page load the global is `undefined`, so `undefined===false` is `false` and `showEventTimestamp` is `true`. That means a user who saved the toggle **off** still sees per-event chips on the first render, until they open Settings (which finally calls the sync helper and sets the global). Reload = chips reappear.

The sibling settings don't have this gap because `boot.js` seeds them from the settings payload `s`. Compare `_chatActivityDisplayMode`/`_transparentStream` at `static/boot.js:3178-3181` and `_autoScrollFollow` at `static/boot.js:3207`:

```js
window._transparentStream=window._chatActivityDisplayMode==='transparent_stream';
...
window._autoScrollFollow=s.auto_scroll_follow!==false;
```

## Suggested fix

Add one line next to `_transparentStream` in the boot success path (`static/boot.js:3181`):

```js
window._transparentEventTimestamps=s.transparent_stream_event_timestamps!==false;
```

and mirror the default in the settings-load-failed fallback (`static/boot.js:~3329`, next to `window._transparentStream=false;`):

```js
window._transparentEventTimestamps=true;  // preserve chips when settings load fails
```

Using `!==false` keeps the "undefined means on" contract consistent with the server default and with how `_syncTransparentEventTimestampsControl` already computes `next=enabled!==false`.

## Verification

Your existing harness proves the chip-suppression contract given the global is set. The boot gap needs a separate check: with `transparent_stream_event_timestamps: false` persisted, load the page fresh (without opening Settings) and confirm `.transparent-event-time` chips are absent on the first Transparent Stream render. A lightweight assertion that `boot.js` seeds `window._transparentEventTimestamps` from `s.transparent_stream_event_timestamps` would pin the regression, paralleling how the mode is seeded today.

rodboev:
Good catch. The saved off-state only took effect after Settings sync because the boot path never seeded `window._transparentEventTimestamps`, so first render still treated `undefined` as visible.

I fixed that in the same seam the rest of the feature already uses:

1. `static/boot.js` now seeds `_transparentEventTimestamps` from `s.transparent_stream_event_timestamps!==false` right next to `_transparentStream`.
2. The settings-load-failed fallback now sets `_transparentEventTimestamps=true`, which keeps the existing default-on contract intact when boot settings fail.
3. The activity-mode plumbing test now pins both boot assignments, while the existing timestamp regression still proves the chip stays hidden and the footer time stays visible once the global is false.

That closes the boot-time gap without widening the feature beyond the current settings or render path.

nesquena-hermes:
Gate: full suite green (13,442 passed), ESLint runtime gate clean, JS syntax OK. Nathan approved the concept (opt-in setting, footer time preserved, your chip-count 1→0 verification confirmed). Rebased onto current master. Shipping.

nesquena-hermes:
Shipped in **exp-v0.52.116**. Thanks @rodboev! 🎉

Issue #2592, opened 2026-05-19
Streaming: tool-only failed turn writes thousands of duplicate `_partial` messages (3x session bloat)


futureworld678:
## Summary

Streaming a tool-only assistant turn that gets cancelled (or has the tool-call fail with `is_error: True`) inserts a `_partial` marker into the session **on every chunk** instead of once per turn. Over a single ~4-second failed stream this produced **2,048 identical `_partial` messages** in one session, bloating it from ~5 MB to **16.3 MB** and causing the WebUI to render 2,000+ "思考过程 / Thinking" cards on load.

## Repro

1. Long-lived session, normal usage.
2. Trigger a turn whose model output is **content-empty** but emits reasoning + a tool call (e.g. a single `execute_code` block — no text reply, just the tool).
3. Have the tool fail (`is_error: True`) or cancel the turn mid-stream.
4. Reopen the session — the sidebar shows thousands of "思考过程" cards, all identical, all `_partial: True`, all sharing the same timestamp.

## Root cause

In `api/streaming.py` around lines **5507–5524** (current HEAD), the duplicate-`_partial` guard only runs when `_stripped` (the streamed text content) is non-empty:

```python
_partial_already_present = False
if _stripped:                                # ← guard only triggers for text turns
    for _m in _cs.messages:
        if not isinstance(_m, dict) or not _m.get('_partial'):
            continue
        if str(_m.get('content') or '').strip() == _stripped:
            _partial_already_present = True
            break
if (_stripped or _has_reasoning or _has_tools) and not _partial_already_present:
    _partial_msg: dict = {
        'role': 'assistant',
        'content': _stripped,                # empty string for tool-only turns
        '_partial': True,
        ...
    }
    ...
    _cs.messages.insert(_cancel_marker_idx, _partial_msg)
```

The insert condition admits **tool-only** turns (`_has_tools` alone is enough), but the dedup check above only fires when `_stripped` is truthy. Result: for a tool-only cancelled/errored turn, `_partial_already_present` stays `False` forever and every replay/retry inserts another copy.

The dedup also keys purely on `content` string equality, which is `''` for tool-only turns — so even if the guard ran it would collapse every tool-only `_partial` ever, regardless of payload. The right key is `(content, reasoning_signature, tool_call_signature)`.

## Evidence

One affected session (~4 days of normal use):

| | Before cleanup | After filtering `_partial=True` |
|---|---|---|
| Messages | 3,291 | 1,243 |
[...]

PR #6138, opened 2026-07-16
fix(search): stream first N messages instead of full transcript per session

rh-id:
## Thinking Path

- `GET /api/sessions/search?content=1` scans only the first `depth` messages (default 5) per session to find a content match.
- It used to call `get_session()`, which materializes the **ENTIRE** transcript (often hundreds of KB to multiple MB) just to read ~5 messages.
- A content search across many large/agentic sessions was a per-request memory spike proportional to total transcript bytes, with the bounded `SESSIONS` LRU thrashing load/evict when the session count exceeded the cache cap.

## What Changed

- Add `Session.load_messages_head(sid, limit, max_bytes)`: streams the on-disk JSON and peels off only the first `limit` elements of the top-level `messages` array via `json.JSONDecoder.raw_decode` (one element at a time), stopping there. The message tail and the large `anchor_activity_scenes` bodies are never parsed.
- Reuses the existing `_read_metadata_json_prefix` for the true `message_count` and a new `_scan_to_messages_array` (depth-tracking scanner, mirrors `_find_top_level_json_key`) to locate the array opener. No new dependency.
- Falls back to a bounded slice of `Session.load()` on any layout anomaly, so callers always get a usable result.
- `_handle_sessions_search` calls `load_messages_head` for the depth-bounded branch; `depth==0` ("whole transcript" opt-in) keeps the full `get_session()` load.

## Why It Matters

Per-session memory for content search is now bounded to `depth` messages (default 5) instead of the full conversation length. A user with many large agentic sessions no longer triggers a transient memory spike proportional to the sum of all transcript sizes on every search.

## Verification

- `tests/test_content_search_partial_load.py` (new, 6 tests): `load_messages_head` returns only N messages from a **5000-message** session with the correct `total_count` (the regression proof: the tail is never materialized); `limit=0` returns all; the route finds a needle within `depth` but NOT beyond it (the correctness boundary); `depth==0` scans the whole transcript.
- 8 edge cases verified in-process: 0/1/many messages, `limit > actual`, escapes/brackets in content, list-shaped (multipart) content, a metadata string value containing `"messages": [` (no false array match), `tool_calls` serialized before `messages`, and a huge first message spanning many read chunks.
- Updated `test_sessions_search_depth_validation.py` to mock the new head-loader path while preserving its depth-validation contract.
- Existing search/sidebar tests pass. All run via `./scripts/test.sh`.

## Risks / Follow-ups

- The partial loader assumes `messages` is a top-level JSON array (true for the current on-disk layout, which writes `messages` before `tool_calls`). The `_scan_to_messages_array` keys on the literal `"messages"` string, so it's robust to reordering within top-level arrays; any structural change to the sidecar format falls back to `Session.load()`.
- `depth==0` (search whole transcript) still does a full load by design — it's the rare opt-in path.

## Contract Routing

- Task type: memory-bound fix (whole-transcript load to read N messages).
- Touched areas: read-only session content search (`api/routes.py` `_handle_sessions_search`) + a new read-only partial loader (`api/models.py` `Session.load_messages_head`). No write path is touched.
- Relevant public docs: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `docs/rfcs/webui-run-state-consistency-contract.md` (Visible transcript layer).
- State layer read: **Visible transcript** (on-disk session JSON, messages array). This is a READ-ONLY change — it changes HOW the search reads messages (stream the first N instead of parse-all-then-slice), not WHAT it returns. The search result set, match semantics, depth contract, and redaction are unchanged.
- Invariant preserved: the `depth` scan contract. Regression: `test_content_search_miss_when_needle_beyond_depth`.
- Scope boundaries: no contract or RFC text is changed. The new method is additive and read-only.

Release note: session content search no longer parses whole transcripts per result — per-session memory is bounded to the search depth instead of the full conversation length.

## Model Used

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -1722,7 +1722,6 @@ class Session:
             # messages could push it higher, but the on-disk layout puts scenes
             # AFTER messages, so this is a generous safety ceiling only.
             _METADATA_SCAN_MAX_BYTES = 4 * 1024 * 1024
-            metadata_scanned = 0
             with open(p, 'r', encoding='utf-8') as f:
                 while True:
                     if limit and len(messages) >= limit:
--- a/api/models.py
+++ b/api/models.py
@@ -1684,8 +1684,6 @@ class Session:
         if not is_safe_session_id(sid):
             return [], None
         p = SESSION_DIR / f'{sid}.json'
-        if not p.exists():
-            return [], None
         if limit is None or limit <= 0:
             limit = 0
         if max_bytes is None:
@@ -1696,12 +1694,57 @@ class Session:
         # Helper: full-load-and-slice fallback. Used by every correctness escape
         # hatch below so all paths return the SAME normalized head that
         # get_session / Session.load would — never a short or denormalized one.
+        # Re-checks SESSIONS authority before returning: a session that becomes
+        # active (cached) DURING the streaming scan must not be replaced by the
+        # stale disk state we just read. (#6138 re-gate finding 1)
         def _full_load_head():
+            with LOCK:
+                cached_now = SESSIONS.get(sid)
+                cached_owns_now = (
+                    cached_now is not None
+                    and str(getattr(cached_now, 'session_id', '') or '') == str(sid)
+                )
+            if cached_owns_now:
+                cached_session = get_session(sid, metadata_only=False)
+                cached_msgs = (cached_session.messages or []) if cached_session else []
+                return (list(cached_msgs[:limit]) if limit else list(cached_msgs)), (
+                    len(cached_msgs)
+                )
             full = cls.load(sid)
             msgs = full.messages or []
             return (list(msgs[:limit]) if limit else list(msgs)), (
                 total_count if total_count is not None else len(msgs)
             )
+
+        # Cache-authority check. Consult SESSIONS BEFORE the sidecar existence
+        # check — this matches get_session()'s authority order. A cached
+        # messageful session can exist before any sidecar is written (active /
+        # pending sessions that all_sessions() overlays into searchable rows),
+        # so the old `if not p.exists(): return [], None` short-circuit silently
+        # dropped them from content search. (#6138 re-gate finding 1)
+        #
+        # LOCK is the same non-reentrant lock get_session() takes internally, so
+        # we release it before delegating to get_session() (which reacquires it)
+        # to avoid a self-deadlock.
+        with LOCK:
+            cached = SESSIONS.get(sid)
+            cached_owns = (
+                cached is not None
+                and str(getattr(cached, 'session_id', '') or '') == str(sid)
+            )
+        if cached_owns:
+            # get_session handles the cache + disk-freshness + state.db +
+            # journal-retry contract end-to-end; reuse it so the head we return
+            # matches what GET /api/session would show for this session.
+            cached_session = get_session(sid, metadata_only=False)
+            cached_msgs = (cached_session.messages or []) if cached_session else []
+            total_count_cache = len(cached_msgs)
+            return (list(cached_msgs[:limit]) if limit else list(cached_msgs)), (
+                total_count_cache
+            )
+
+        if not p.exists():
+            return [], None
 
         try:
             # 1) Cheap metadata prefix → true message_count (does not parse the
@@ -1718,28 +1761,7 @@ class Session:
                 except Exception:
                     pass
 
-            # 2) Cache-aware path (blocker #1 from #6138 gate certification):
-            #    active/unsaved sessions live in SESSIONS and may carry messages
-            #    that have NOT been persisted to the sidecar yet. The streaming
-            #    scanner reads the sidecar directly, so it would silently miss
-            #    those messages. Under the same lock get_session uses, detect a
-            #    cached session and delegate to get_session() (which owns disk-
-            #    freshness, state.db sync, and journal-retry semantics) and slice
-            #    its NORMALIZED messages. The streaming scanner is only safe for
-            #    clean, uncached, fully persisted sessions.
-            with LOCK:
-                cached = SESSIONS.get(sid)
-            if cached is not None and str(getattr(cached, 'session_id', '') or '') == str(sid):
-                # get_session handles the cache + disk-freshness + state.db
-                # contract end-to-end; reuse it so the head we return matches
-                # what GET /api/session would show for this session.
-                cached_session = get_session(sid, metadata_only=False)
-                cached_msgs = (cached_session.messages or []) if cached_session else []
-                return (list(cached_msgs[:limit]) if limit else list(cached_msgs)), (
-                    total_count if total_count is not None else len(cached_msgs)
-                )
-
-            # 3) Stream the file body to locate the "messages": [ opener and
+            # 2) Stream the file body to locate the "messages": [ opener and
             #    peel off raw array elements one at a time. Track whether we hit
             #    a safety cap so we can fall back rather than return a short head.
             decoder = json.JSONDecoder()
@@ -1850,12 +1872,19 @@ class Session:
             # stopped at the overcollect ceiling (not at end-of-array), there may
             # be more duplicates just past the window — fall back to a full load
             # rather than risk a short normalized head.
+            #
+            # The fallback must NOT depend on total_count: adjacent identical
+            # _partial runs are unbounded, so a fixed `2 * limit` raw window is
+            # not normalization-equivalent. A valid legacy sidecar without
+            # message_count (total_count is None) can stop after 2N raw
+            # duplicates, collapse to fewer than N normalized rows, and silently
+            # drop a needle at normalized message N. Fall back whenever the raw
+            # ceiling was reached and collapse still yielded fewer than limit,
+            # regardless of metadata count. (#6138 re-gate finding 2)
             if (
                 limit
                 and len(messages) < limit
                 and len(raw_messages) >= raw_target
-                and total_count is not None
-                and total_count > len(raw_messages)
             ):
                 return _full_load_head()
             return messages, total_count
--- a/api/models.py
+++ b/api/models.py
@@ -1692,6 +1692,17 @@ class Session:
             max_bytes = 1024 * 1024
         if max_bytes <= 0:
             max_bytes = 1
+
+        # Helper: full-load-and-slice fallback. Used by every correctness escape
+        # hatch below so all paths return the SAME normalized head that
+        # get_session / Session.load would — never a short or denormalized one.
+        def _full_load_head():
+            full = cls.load(sid)
+            msgs = full.messages or []
+            return (list(msgs[:limit]) if limit else list(msgs)), (
+                total_count if total_count is not None else len(msgs)
+            )
+
         try:
             # 1) Cheap metadata prefix → true message_count (does not parse the
             #    messages array).
@@ -1706,10 +1717,34 @@ class Session:
                             total_count = mc
                 except Exception:
                     pass
-            # 2) Stream the file body to locate the "messages": [ opener and
-            #    peel off the first `limit` array elements one at a time.
+
+            # 2) Cache-aware path (blocker #1 from #6138 gate certification):
+            #    active/unsaved sessions live in SESSIONS and may carry messages
+            #    that have NOT been persisted to the sidecar yet. The streaming
+            #    scanner reads the sidecar directly, so it would silently miss
+            #    those messages. Under the same lock get_session uses, detect a
+            #    cached session and delegate to get_session() (which owns disk-
+            #    freshness, state.db sync, and journal-retry semantics) and slice
+            #    its NORMALIZED messages. The streaming scanner is only safe for
+            #    clean, uncached, fully persisted sessions.
+            with LOCK:
+                cached = SESSIONS.get(sid)
+            if cached is not None and str(getattr(cached, 'session_id', '') or '') == str(sid):
+                # get_session handles the cache + disk-freshness + state.db
+                # contract end-to-end; reuse it so the head we return matches
+                # what GET /api/session would show for this session.
+                cached_session = get_session(sid, metadata_only=False)
+                cached_msgs = (cached_session.messages or []) if cached_session else []
+                return (list(cached_msgs[:limit]) if limit else list(cached_msgs)), (
+                    total_count if total_count is not None else len(cached_msgs)
+                )
+
+            # 3) Stream the file body to locate the "messages": [ opener and
+            #    peel off raw array elements one at a time. Track whether we hit
+            #    a safety cap so we can fall back rather than return a short head.
             decoder = json.JSONDecoder()
-            messages: list = []
+            raw_messages: list = []  # raw array elements (NOT yet collapsed)
+            cap_hit = False  # set if max_bytes/_METADATA_SCAN_MAX_BYTES truncated
             READ_CHUNK = 16384
             buf = ''
             array_opened = False
@@ -1722,9 +1757,16 @@ class Session:
             # messages could push it higher, but the on-disk layout puts scenes
             # AFTER messages, so this is a generous safety ceiling only.
             _METADATA_SCAN_MAX_BYTES = 4 * 1024 * 1024
+            # Over-collect raw elements so the adjacent-duplicate-_partial
+            # collapse (blocker #3) still yields `limit` NORMALIZED messages.
+            # The collapse ratio is bounded (duplicate partials are rare and
+            # adjacent), so a small multiplier is sufficient; if it ever isn't,
+            # the fallback below catches it.
+            _RAW_OVERCOLLECT = 2
+            raw_target = limit * _RAW_OVERCOLLECT if limit else 0
             with open(p, 'r', encoding='utf-8') as f:
                 while True:
-                    if limit and len(messages) >= limit:
+                    if raw_target and len(raw_messages) >= raw_target:
                         break
                     # Skip inter-element whitespace/commas (and, before the
                     # array opens, the metadata we don't care about).
@@ -1755,6 +1797,7 @@ class Session:
                             if len(buf) >= _METADATA_SCAN_MAX_BYTES:
                                 # Pathological metadata prefix; give up the
                                 # streaming scan and fall back to a full load.
+                                cap_hit = True
                                 break
                             chunk = f.read(READ_CHUNK)
                             if not chunk:
@@ -1780,22 +1823,41 @@ class Session:
                         buf = buf[s:] + chunk
                         continue
                     if isinstance(value, dict):
-                        messages.append(value)
+                        raw_messages.append(value)
                     scanned_bytes += end
                     buf = buf[end:]
                     if scanned_bytes > max_bytes:
-                        # Pathological first-N payload; bound memory/time.
+                        # Pathological first-N payload; bound memory/time. Do
+                        # NOT return the short head — fall back so callers never
+                        # see a truncated result (blocker #2 from #6138 gate).
+                        cap_hit = True
                         break
-            if not array_opened:
-                # Never located the array — fall back to a full load. This is
-                # layout-anomaly-driven (NOT gated on total_count, which modern
-                # sessions always supply from message_count): returning ([],
-                # total_count) here would silently drop content-search matches.
-                full = cls.load(sid)
-                msgs = full.messages or []
-                return (list(msgs[:limit]) if limit else list(msgs)), (
-                    total_count if total_count is not None else len(msgs)
-                )
+            if not array_opened or cap_hit:
+                # Never located the array, OR a safety cap truncated the scan
+                # before we had enough raw elements. Both must fall back to a
+                # full load — returning a short head would silently drop
+                # content-search matches.
+                return _full_load_head()
+
+            # 4) Apply the same adjacent-duplicate-_partial collapse that
+            #    Session.load() applies (blocker #3 from #6138 gate). The
+            #    streaming scanner reads RAW array elements, so without this
+            #    step a duplicate-partial before the requested depth would shift
+            #    the depth window and silently hide a needle.
+            collapsed, _changed = _collapse_adjacent_duplicate_partials(raw_messages)
+            messages = collapsed[:limit] if limit else collapsed
+            # If collapse shortened the result below `limit` AND the raw scan
+            # stopped at the overcollect ceiling (not at end-of-array), there may
+            # be more duplicates just past the window — fall back to a full load
+            # rather than risk a short normalized head.
+            if (
+                limit
+                and len(messages) < limit
+                and len(raw_messages) >= raw_target
+                and total_count is not None
+                and total_count > len(raw_messages)
+            ):
+                return _full_load_head()
             return messages, total_count
         except Exception:
             logger.debug(
--- a/api/models.py
+++ b/api/models.py
@@ -1710,7 +1710,11 @@ class Session:
                 return (list(cached_msgs[:limit]) if limit else list(cached_msgs)), (
                     len(cached_msgs)
                 )
+            # cls.load may return None if the sidecar disappeared during the
+            # scan (TOCTOU); return an empty head rather than dereferencing None.
             full = cls.load(sid)
+            if full is None:
+                return ([], total_count)
             msgs = full.messages or []
             return (list(msgs[:limit]) if limit else list(msgs)), (
                 total_count if total_count is not None else len(msgs)
@@ -1892,9 +1896,12 @@ class Session:
             logger.debug(
                 "load_messages_head fell back to full load for %s", sid, exc_info=True
             )
-            full = cls.load(sid)
-            msgs = full.messages or []
-            return (list(msgs[:limit]) if limit else list(msgs)), len(msgs)
+            # Route through _full_load_head so this path ALSO preserves SESSIONS
+            # authority: an active/messageful session may become cached while the
+            # scan is in progress, and a metadata/read/parse exception taking
+            # this branch must not return stale sidecar state in place of the
+            # now-authoritative cached messages. (#6138 re-gate round 3)
+            return _full_load_head()
 
     @staticmethod
     def _compute_user_message_count(messages) -> int:
--- a/api/models.py
+++ b/api/models.py
@@ -1592,6 +1592,221 @@ class Session:
             return cls.load(sid)
 
     @staticmethod
+    def _scan_to_messages_array(buf: str, start: int = 0):
+        """Return the index in ``buf`` just past the ``[`` opening the top-level
+        ``messages`` array, or ``None`` if not found within ``buf``.
+
+        Mirrors the depth-tracking approach of ``_find_top_level_json_key`` but
+        returns the position of the array body (after ``[``) so a caller can
+        peel off individual elements without re-parsing the whole prefix.
+        """
+        depth = 0
+        i = start
+        n = len(buf)
+        while i < n:
+            ch = buf[i]
+            if ch == '"':
+                # Skip a string literal (handles escapes).
+                i += 1
+                escaped = False
+                while i < n:
+                    c = buf[i]
+                    if escaped:
+                        escaped = False
+                    elif c == '\\':
+                        escaped = True
+                    elif c == '"':
+                        break
+                    i += 1
+                if i >= n:
+                    return None
+            elif ch == '{':
+                depth += 1
+            elif ch == '}':
+                depth -= 1
+            elif depth == 1 and ch == '[':
+                # We are at the top level (depth would be 1 inside the session
+                # object). The token immediately before '[' must be the key
+                # "messages". Look back for the most recent string token.
+                # Simpler: verify by scanning the preceding non-space chars for
+                # the closing quote of a key and that the key text is messages.
+                # To stay robust we rely on _read_metadata_json_prefix already
+                # having validated layout; here we accept the first depth==1 '['
+                # whose preceding key is "messages" — detect by walking back.
+                j = i - 1
+                while j >= 0 and buf[j] in ' \t\r\n':
+                    j -= 1
+                # Expect ':' then the key string before it.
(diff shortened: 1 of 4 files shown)


nesquena-hermes:
Reading `load_messages_head` and `_scan_to_messages_array` at HEAD (`api/models.py`) against the on-disk layout that `Session.save()` writes (`message_count` in the metadata prefix, then `messages`, then `anchor_activity_scenes` — `api/models.py:1335-1346`), the streaming approach is sound and the `raw_decode`-one-element-at-a-time design is the right call. Confirming the multi-chunk correctness gap the automated review raised, with a concrete fix.

## The bug

`_scan_to_messages_array` re-initializes `depth = 0` on every call and always scans `buf` from `start=0`. In `load_messages_head`, when the first read chunk doesn't contain the `"messages": [` opener, the buffer is trimmed to a 64-byte tail and a fresh chunk is appended:

```python
pos = cls._scan_to_messages_array(buf)
if pos is None:
    keep = max(0, len(buf) - 64)
    buf = buf[keep:]
    chunk = f.read(READ_CHUNK)
    ...
```

The next scan then starts mid-object with the root `{` already discarded, so `depth` never climbs to 1, the `elif depth == 1 and ch == '['` branch never fires, and the `messages` array is never located. Since `READ_CHUNK = 16384`, this triggers whenever the pre-`messages` metadata exceeds ~16 KiB.

## Why the fallback doesn't save it

The guard is:

```python
if not array_opened and total_count is None:
    full = cls.load(sid)
    ...
```

But `total_count` is populated from the metadata prefix (`message_count`) for every modern session, so `total_count is None` is false and the guard is skipped — the method returns `([], total_count)`. Content-search matches in the first `depth` messages are then silently dropped for those sessions, with no exception and no fallback. That's worse than a crash because it's invisible.

## Fix sketch

Two independent fixes; I'd apply both:

1. **Make the fallback layout-anomaly-driven, not count-driven.** The condition should trip whenever the array was never opened, regardless of whether metadata gave a count:
   ```python
   if not array_opened:
       full = cls.load(sid)
       msgs = full.messages or []
       return (list(msgs[:limit]) if limit else list(msgs)), (total_count if total_count is not None else len(msgs))
   ```
2. **Don't lose scanner state across chunks.** Trimming to a 64-byte tail throws away the depth context. Simplest robust option: track cumulative offset and keep scanning from a retained position rather than resetting depth, or (given the metadata prefix is already bounded) read the prefix region up to a hard cap in one pass before switching to element peeling. The current 64-byte carry only protects a split key *token*, not the brace-depth accounting the scanner relies on.

## Test gap

`tests/test_content_search_partial_load.py` builds sessions via `save()` but the metadata prefixes there stay well under 16 KiB, so the failing path isn't exercised. A regression test with a session whose pre-`messages` metadata exceeds `READ_CHUNK` (e.g. a large `anchor_scene_index` or padded title/metadata) would have caught the empty-result return — worth adding alongside the fix so `test_load_messages_head_returns_only_first_n` has a large-prefix sibling.

rh-id:
Thanks @nesquena-hermes — you and the bot both nailed a real silent-correctness bug. Confirming and fixing both points in `6fd23ba`.

## The bug — confirmed exactly as described

I reproduced it with a session whose pre-`messages` metadata (~29 KB, from a large `anchor_scene_index`) exceeds `READ_CHUNK` (16 KB):

```
RESULT: head=0 messages, total=3   # silently empty despite 3 messages on disk
```

Two compounding defects, both as you laid out:

1. **Scanner loses depth across chunks.** `_scan_to_messages_array` re-initializes `depth=0` on every call and always scans from `start=0`. When the first chunk didn't contain `"messages": [`, the old code trimmed `buf` to a 64-byte tail and appended a fresh chunk — so the next scan started mid-object with the root `{` already discarded, `depth` never climbed to 1, and the array was never located.
2. **Fallback was count-driven, not layout-driven.** The guard `if not array_opened and total_count is None` never tripped because modern sessions always supply `message_count`, so `total_count is not None` — and the method returned `([], total_count)`, silently dropping every content-search match for those sessions.

## Fix (`6fd23ba`)

Applied both of your recommended fixes:

1. **Layout-anomaly-driven fallback.** The guard is now `if not array_opened:` — it falls back to `Session.load()` whenever the array was never located, regardless of whether metadata gave a count. Returning `([], total_count)` is no longer possible.
2. **Don't lose scanner state across chunks.** The streaming scan no longer trims the prefix to a 64-byte tail (which discarded the brace-depth context). Instead it keeps accumulating `buf` until the `"messages": [` opener is in view, bounded by a new `_METADATA_SCAN_MAX_BYTES = 4 MiB` safety ceiling (the metadata prefix is normally a few KB; `anchor_activity_scenes` serialize *after* `messages` in the on-disk layout, so this is generous). Over that ceiling it falls back to a full load rather than scanning unboundedly.

## Test added

`test_load_messages_head_large_metadata_prefix` builds a session whose metadata prefix is ~29 KB (> one read chunk) with a `message_count` set (so the old count-driven guard would have skipped the fallback) and asserts `load_messages_head` returns all 3 messages — `test_load_messages_head_returns_only_first_n` now has the large-prefix sibling you asked for. All 21 search tests pass.

The normal small-metadata paths, `limit=0`, empty, `limit > actual`, and the huge-single-message-spanning-chunks case are all re-verified and unchanged.

nesquena-hermes:
**Certified head:** `sha:6fd23ba31f0a` · **PR:** #6138 · _@rh-id, fix(search): stream first N messages instead of full transcript per session_
**Rebased gate target:** `1b38dd703716` on `origin/master` `c67fd2dd270a` (`exp-v0.52.113`); patch-id preserved as `b2728f45febe`

### Verdict

**Do not merge this head.** The optimization silently changes which messages content search can see in three production-valid states. I reproduced every blocker against the rebased PR and a clean-master control in separate interpreters.

### What ran

| Leg | Result |
|---|---|
| Static threat scan | **CLEAN** (score 0); contributor code was executed only through the sandbox gate |
| Codex independent review | `SHIP ONLY WITH FIXES`; reproduced all three blockers below |
| Opus backend/parser review | **Hold**; independently reproduced the `max_bytes` truncation and identified the cache and duplicate-partial divergences |
| Full production-like serial suite | Completed on the latest rebased target: **13,308 passed, 82 skipped, 1 xfailed, 2 xpassed, 16 failed** in 462.79s |
| Full-suite failure controls | 14 sandbox-environment failures reproduce on clean master. The two remaining full-order failures pass isolated on this PR and clean master. None touches the changed search path. |
| PR-focused tests | **10 passed**: `test_content_search_partial_load.py` + `test_sessions_search_depth_validation.py` |
| Independent runtime differential | Three candidate/control scenarios reproduced the failures below through the real route and model code |
| Browser / Fable | Skipped: backend parser/cache change with no UI or CSS surface |

The sandbox-only failures are the known no-`.git` worktree metadata checks, unavailable Playwright browser under sandbox `$HOME`, sandbox path-policy expectations, and keyless-onboarding environment state. Two suite-order cases (`new_session_lives_in_memory`, Gateway continuation projection) passed together in an isolated sandbox rerun. They do not clear the three product regressions below.

### Blocking findings

1. **CORE — active or unsaved cached messages disappear from search.**
   - `api/routes.py:16778` now reads the sidecar directly with `Session.load_messages_head()` instead of consulting `get_session()`.
   - Production deliberately keeps active/unsaved sessions in `SESSIONS`; `get_session()` returns that authoritative object and also owns disk-freshness and recovery behavior.
   - Reproducer: disk had one persisted message; the active cached session had a second assistant message containing `UNSAVEDNEEDLE`. Clean master returned `count=1`; this PR returned `count=0`.
   - **Required fix:** add a cache-aware head loader. Under the existing session lock, detect a cached session, then use the established `get_session()` path and slice its normalized `messages`. Use the partial sidecar scanner only for clean, uncached, fully persisted sessions. Do not infer cleanliness from the sidecar alone.

2. **CORE — the 1 MiB cap silently truncates a requested search depth.**
   - `api/models.py:1856` breaks when cumulative decoded bytes exceed `max_bytes`, even when fewer than `limit` messages have been returned.
   - Reproducer: five valid 300 KB messages, `depth=5`, needle in message 5. Clean master returned `count=1`; this PR stopped after message 4 and returned `count=0`.
   - **Required fix:** cap exhaustion must never return a shorter successful head. If the cap is reached before `limit` normalized messages, fall back to the established full loader and slice, or remove the cumulative cap. Add a route-level regression with the match after the cap.

3. **CORE — duplicate-partial normalization shifts the depth window.**
   - Full `Session.load()` collapses adjacent duplicate `_partial` rows before slicing. The scanner stops after `limit` raw array elements and bypasses that invariant.
   - Reproducer: one adjacent duplicate partial before a needle in normalized message 5. Clean master collapsed the duplicate and returned `count=1`; this PR stopped at raw element 5 and returned `count=0`.
   - **Required fix:** preserve `_collapse_adjacent_duplicate_partials()` semantics while decoding and stop after `limit` normalized messages, not raw elements. Add the existing #2592-shaped regression at route level.

### Required proof on the next head

- Candidate/control regressions for cached active/unsaved state, cap exhaustion before depth, and duplicate-partial normalization.
- Existing 10 focused tests remain green.
- Full serial suite completes, with any environmental failure controlled on same-host clean master.
- Re-run Codex and Opus on the fresh SHA; this result certifies only `6fd23ba31f0a`.

**Keep RED and bounce to the author with the three fix specs above.** The intended optimization is worthwhile, and the parser's ordinary valid-JSON behavior is good, but content search cannot trade silent false negatives for memory savings. Re-warm and re-gate only after a new head is pushed.

rh-id:
Thanks @greptile-apps[bot] for the review. Replying to all three inline comments — TL;DR: the two P1 concerns were already fixed in the shipped commit (the inline comments were left on the pre-fix revision `6fd23ba` parent), and the dead-variable P2 is removed in `f04ed63`.

## P1 (inline) — `_scan_to_messages_array` depth tracking resets to 0 after chunk boundary

This was a real bug in an **earlier draft** of the PR (it kept only a 64-byte tail across chunks, losing the root `{` depth context). It was fixed before merge with explicit commentary in the shipped code. Current `api/models.py:1746-1764`:

```python
# _scan_to_messages_array tracks brace depth from the START of buf, so it
# must see the root '{' to reach depth 1. Therefore we do NOT trim the
# prefix here across chunks (a previous version kept only a 64-byte tail,
# which discarded the depth context and caused the array to be missed once
# the metadata exceeded one read chunk — silently returning []).
# Instead, keep accumulating until the opener is in view, bounded by
# _METADATA_SCAN_MAX_BYTES.
```

The buffer now accumulates (`buf += chunk` at line 1763) rather than trimming, so each call to `_scan_to_messages_array(buf)` sees the full prefix and depth tracks correctly from 0. Capped at `_METADATA_SCAN_MAX_BYTES = 4 MiB` (line 1724), at which point it breaks and falls through to the full-load fallback. Pinned by `test_load_messages_head_large_metadata_prefix` which builds a session whose pre-`messages` metadata exceeds one read chunk and asserts the head is still returned.

## P1 (inline) — Fallback skipped for modern sessions when array is never found

Same story — the fallback condition was tightened in the shipped commit. Current `api/models.py:1790`:

```python
if not array_opened:
    # Never located the array — fall back to a full load. This is
    # layout-anomaly-driven (NOT gated on total_count, which modern sessions
    # always supply from message_count): returning ([], total_count) here
    # would silently drop content-search matches.
    full = cls.load(sid)
    ...
```

The fallback now fires purely on `not array_opened`, independent of `total_count`. So even a modern session with `message_count` populated that somehow misses the array opener (e.g. the 4 MiB metadata cap is hit) still gets a correct head via the full-load path rather than an empty result.

Greptile's own summary on the latest commit confirms both: *"The two correctness bugs reported in earlier review rounds (unconditional fallback when the array is never found, and depth-tracker reset across chunk boundaries) are both demonstrably fixed by accumulating the full buffer and gating the fallback solely on `array_opened`."*

## P2 (inline) — Dead variable `scanned_delta` / `metadata_scanned`

Removed `metadata_scanned` (the `scanned_delta` variant was already gone). `f04ed63` deletes the single assignment at the top of `load_messages_head`. AST-verified that nothing else in the function is dead, and the strict-lint check on `api/models.py` drops from 4 ruff errors to 3 (removes the F841; the remaining 3 F401 `sqlite3` errors are pre-existing and unrelated).

All 10 tests in `test_content_search_partial_load.py` + `test_sessions_search_depth_validation.py` still pass. Ready for maintainer review.

rh-id:
Apologies for the earlier reply — I responded to the Greptile inline comments without seeing your 🔬 Gate certification — RED ⛔ above them. Reading it now and addressing all three CORE blockers in `fb0d4dc`. Each reproduced exactly as you described.

## Blocker 1 — cached active/unsaved messages disappear — FIXED

Reproduced: disk had one persisted message; cached session had a second with `UNSAVEDNEEDLE`. Pre-fix returned the 1-message disk head and missed the needle.

**Fix** (`api/models.py` load_messages_head): under `LOCK`, check `SESSIONS` first. If a cached session exists (and the ownership check passes), delegate to `get_session(sid, metadata_only=False)` — which owns disk-freshness, state.db sync, and journal-retry semantics — and slice its normalized `.messages`. The streaming scanner now runs only for clean, uncached, fully persisted sessions.

Verified: cached unsaved message now appears in the head.

## Blocker 2 — 1 MiB cap silently truncates requested depth — FIXED

Reproduced: 5 × ~300KB messages, depth=5, needle in message 5. Pre-fix stopped at message 4 (cap exceeded) and returned 4 messages, missing the needle.

**Fix**: cap exhaustion (`scanned_bytes > max_bytes`) now sets a `cap_hit` flag, and the post-scan block falls back to a full load whenever `cap_hit` is true — never returns a short head. The `_METADATA_SCAN_MAX_BYTES` ceiling (4 MiB) sets the same flag for consistency.

Verified: 5 × 300KB messages with depth=5 now returns all 5 with the needle.

## Blocker 3 — duplicate-partial normalization shifts the depth window — FIXED

Reproduced: 3 normal + 2 identical `_partial` (one collapses) + needle = 6 raw on disk, 5 normalized; needle at normalized message 5 (index 4). Pre-fix stopped at raw element 5 and returned `['msg 1','msg 2','msg 3','partial','partial']`, missing the needle. Clean master returned count=1.

**Fix**: after collecting raw elements, apply `_collapse_adjacent_duplicate_partials` (the same helper `Session.load` uses) and slice the collapsed list. The scanner over-collects `limit * 2` raw elements so the collapse has material to work with; if the collapsed head is still shorter than `limit` AND the disk `total_count` exceeds the raw scan window (meaning more duplicates may lie just past), it falls back to a full load rather than risk a short normalized head.

Verified: needle at normalized message 5 is now found.

## Test notes

Two implementation details I had to work around in the regression tests, both worth flagging:

- **Blocker 3's test writes the sidecar JSON directly** rather than via `Session.save()`. `Session.save()` itself collapses adjacent duplicate partials before writing (so the duplicates never reach disk through the normal save path). The on-disk duplicates that blocker 3 reproduces arise from streaming/journal recovery paths that write raw — the test mirrors that.
- **The test also avoids calling `Session.load(sid)` for a sanity check.** `Session.load()` self-heals collapsed partials back to disk (`api/models.py` #2592 self-heal at line ~1474), which would destroy the duplicates before `load_messages_head` runs and silently defeat the test. I called this out in the test docstring so a future reader doesn't re-add the sanity check.

## Regression gate (per guideline #6)

- Blocker 1 test fails pre-fix (cache miss, unsaved needle not found).
- Blocker 2 test fails pre-fix (`assert 4 == 5`, cap truncated the head).
- Blocker 3 test fails pre-fix with the exact signature you described: `got ['msg 1', 'msg 2', 'msg 3', 'partial', 'partial']`, needle missing.
- Regression guard (`test_load_messages_head_streaming_still_used_for_clean_uncached`) confirms the optimization still applies for the common case — clean, uncached, fully persisted sessions still get the streaming scan (returns 5 of 50 without a full load).

All 11 tests in `test_content_search_partial_load.py` pass with the fix; 25/25 across the broader search surface (`test_content_search_partial_load`, `test_sessions_search_depth_validation`, `test_sessions_search_profile_scope`, `test_session_search_bfcache_822`).

## Invariant

The head returned by `load_messages_head` is now always correctness-equivalent to `get_session(sid).messages[:limit]` — never shorter, never denormalized. The optimization applies only when it cannot change the observable result (clean, uncached, fully persisted sessions with no adjacent duplicate partials in the head window and no cap exhaustion).

Re-push is in (`fb0d4dc`); ready for re-gate.

nesquena-hermes (changes requested):
## Re-gate: partial convergence, two silent false-negative paths remain

Thanks for fixing the three cases from the prior RED gate. I re-ran the current head `462986bfdec53` through the mandatory offline sandbox. The focused contributor tests are green (**14 passed**), and the byte-cap blocker is genuinely fixed. Two correctness gaps remain in `Session.load_messages_head()`.

### 1. A cached active session with no sidecar is still invisible to content search

`api/models.py:1686-1688` checks `p.exists()` and returns `([], None)` before the new `SESSIONS` lookup at `:1730`. That is the reverse of the established `get_session()` authority order, which checks `SESSIONS` first.

This state is production-valid: `all_sessions()` overlays in-memory active/pending sessions into the searchable rows even before a sidecar exists. A cached messageful session can therefore appear in the session list while content search returns no messages.

Reviewer regression, sandboxed at this head:

```text
test_cached_messageful_session_without_sidecar_uses_authoritative_cache
expected ["UNSAVEDNEEDLE"], got []
```

**Fix:** consult `SESSIONS` under `LOCK` before the sidecar existence check. Release the non-reentrant lock before calling `get_session()`. For disk-derived and full-load fallback results, re-check cache authority before returning so a session that becomes active during the scan is not replaced by stale disk state.

### 2. Fixed `2 * depth` raw over-collection is not normalization-equivalent

`api/models.py:1760-1766` assumes a two-times raw window is enough before `_collapse_adjacent_duplicate_partials()`. Adjacent identical `_partial` runs are unbounded. The fallback at `:1853-1860` also requires `total_count is not None`, so a valid legacy sidecar without `message_count` can stop after `2N` raw duplicates, collapse to fewer than `N` normalized rows, and return a short head while later in-depth messages are silently skipped.

Reviewer regression, sandboxed at this head:

```text
test_duplicate_partial_run_longer_than_raw_multiplier_falls_back
expected normalized depth 3, got 1 row; later NEEDLE was omitted
```

**Fix:** stop based on normalized-message count, not a fixed raw multiplier. A simpler safe version is to fall back whenever the raw ceiling is reached and collapse yields fewer than `limit`, regardless of metadata count. Add coverage with more than `2 * depth` identical partials, missing/stale `message_count`, and a needle at normalized message `N`.

### Re-gate evidence

- Threat scan: **CLEAN**, score 0
- Contributor focused files: **14 passed**
- Reviewer residual regressions: **2 failed**, both deterministic and run only through `warmup-safe-test.sh` + bubblewrap
- Prior blocker 2 (cap exhaustion): **fixed**

Please keep the streaming optimization for clean uncached sessions, but make every returned head correctness-equivalent to `get_session(sid).messages[:limit]` for these two states. Once those regressions are green, this is ready for another re-gate.

rh-id:
Thanks for the re-gate — both residual silent false-negatives reproduced exactly as you described and are fixed in `523c08b`.

## Finding 1 — cached active session with no sidecar is invisible — FIXED

Reproduced: cached messageful session (one message with `UNSAVEDNEEDLE`), no sidecar on disk. Pre-fix returned `[], None` at the `if not p.exists()` short-circuit before the SESSIONS lookup — reversing get_session()'s authority order.

**Fix** (`api/models.py` load_messages_head): consult SESSIONS under LOCK BEFORE the `p.exists()` check. The LOCK is the same non-reentrant lock get_session takes internally, so I release it before delegating to `get_session(sid, metadata_only=False)` (which reacquires it) — avoids self-deadlock. Also followed your recommendation for the disk-derived path: the `_full_load_head` fallback helper now re-checks SESSIONS authority before returning, so a session that becomes active DURING the streaming scan is not replaced by stale disk state.

Verified: cached session with no sidecar now returns its messages.

## Finding 2 — fixed `2 * depth` raw over-collection not normalization-equivalent — FIXED

Reproduced: 12 identical `_partial` messages + a needle (limit=5, so `2 * limit = 10` < 12). Pre-fix stopped at the raw ceiling, collapsed to 1 partial, and returned a short head — dropping the needle. The fallback's `total_count is not None and total_count > len(raw_messages)` gate meant a legacy sidecar without `message_count` (or with a stale count) silently took the short-head path.

**Fix**: dropped the `total_count` gate entirely. The fallback now fires whenever the raw ceiling was reached AND collapse still yields fewer than `limit` normalized messages, regardless of metadata count. This matches your "simpler safe version" recommendation.

Verified: 12+duplicate run with both missing `message_count` (legacy) and stale `message_count=15` now fall back to full load and find the needle.

## Tests (3 new, matching your sandbox regression names)

- `test_cached_messageful_session_without_sidecar_uses_authoritative_cache` — finding 1
- `test_duplicate_partial_run_longer_than_raw_multiplier_falls_back` — finding 2, parametrized over `[None=missing message_count, 15=stale message_count]` to cover both the legacy and stale-count cases you flagged

### Regression gate
All three new tests **fail pre-fix** with the exact signatures:
- Finding 1: `got []` for the cached no-sidecar session
- Finding 2 (both params): `got ['partial']` — collapsed short head, needle missing

All 14 tests in `test_content_search_partial_load.py` pass with the fix. 35/35 across the broader search + `test_empty_session_no_disk_write` surface.

## Bonus: CI shard-3 failure root-caused and fixed

The 3 CI failures (test (3.11/3.12/3.13, shard 3)) were all `test_empty_session_no_disk_write.py::test_new_session_lives_in_memory` — the suite-order case you'd flagged as known-flaky in the prior RED gate. I traced it to a test-isolation leak in my own file: `_run_search` was rebinding `models.SESSIONS = OrderedDict()`, which left `test_empty_session_no_disk_write`'s top-level `from api.models import SESSIONS` reference pointing at the stale dict. Switched to `models.SESSIONS.clear()` (in-place) and `models.SESSIONS` (attribute access) in the cache-aware tests. Verified the leak is gone — my file now runs cleanly before `test_empty_session_no_disk_write`.

## Invariant

The head returned by `load_messages_head` is correctness-equivalent to `get_session(sid).messages[:limit]` for all six production-valid states now:
- cached active/unsaved (blocker 1)
- cached with no sidecar (finding 1)
- max_bytes cap exhaustion (blocker 2)
- duplicate-partial runs of any length, with or without `message_count` (blocker 3 + finding 2)
- clean uncached persisted sessions still get the streaming optimization

Re-push is in (`523c08b`); ready for re-gate.

nesquena-hermes (changes requested):
## Re-gate: one cache-authority fallback still returns stale disk state

Thanks for the re-push. I re-gated current head `523c08b447b2` through the mandatory offline sandbox.

The two named fixes mostly converged:

- The entry path now checks `SESSIONS` before sidecar existence, so a cached messageful session with no sidecar is visible.
- The duplicate-partial fallback no longer depends on `message_count`, so missing and stale-low counts recover the normalized head correctly.
- The `SESSIONS.clear()` fixture repair also closes the cross-file isolation failure.

One deterministic residual remains.

### Blocking finding: the broad exception fallback bypasses cache authority

`Session.load_messages_head()` adds a cache-aware `_full_load_head()` helper at `api/models.py:1700-1717`, and the explicit cap/layout/short-normalization paths correctly use it. But the broad exception path at `api/models.py:1891-1897` still calls `cls.load(sid)` directly:

```python
except Exception:
    ...
    full = cls.load(sid)
    msgs = full.messages or []
    return ..., len(msgs)
```

That violates the same authority invariant fixed elsewhere. A production-valid race is:

1. The initial `SESSIONS` lookup misses.
2. An active/messageful session becomes cached while the sidecar scan is in progress.
3. A metadata/read/parse exception takes the broad fallback.
4. The fallback returns stale sidecar messages instead of the now-authoritative cached messages.

A deterministic reviewer barrier inserted the active cache entry after the initial miss and then forced the scanner read to raise. At this head it returned `STALE_DISK`, not `CACHE_AUTHORITY`.

### Required fix

Route the broad exception path through the helper already written for this invariant:

```python
except Exception:
    logger.debug(...)
    return _full_load_head()
```

Make `_full_load_head()` null-safe as well if `cls.load(sid)` can return `None` after a sidecar disappears during the scan. Add a permanent barrier regression covering initial cache miss → cache insertion → scanner exception → cached head returned, with bounded thread joins to prove there is no deadlock.

Please also strengthen the duplicate-partial test's stale count to a fail-first value at or below `raw_target` (for example `message_count=1`; the current `15` already triggered the old fallback) and make the clean-streaming test fail or spy if `Session.load()` is called.

### Re-gate evidence

- Threat scan: **CLEAN**, score 0
- Contributor focused file: **14 passed**
- Cross-file isolation order (`test_content_search_partial_load.py` then `test_empty_session_no_disk_write.py`): **21 passed**
- Depth-validation regression: **3 passed**
- Reviewer probes: explicit cache-aware fallback, stale-low-count recovery, and clean streaming passed; broad-exception/cache-authority probe failed deterministically

Please keep the PR RED until the broad exception path preserves `SESSIONS` authority too.

rh-id:
Thanks — the broad-exception bypass reproduced exactly as your barrier described and is fixed in `98f001b`.

## Blocking finding — broad exception path bypasses cache authority — FIXED

Reproduced: initial SESSIONS miss → cache inserted mid-scan → scanner forced to raise → pre-fix returned `STALE_DISK` instead of `CACHE_AUTHORITY`.

**Fix** (`api/models.py` load_messages_head broad-exception block): replaced the direct `cls.load(sid)` call with `return _full_load_head()`, which re-checks SESSIONS under LOCK before falling back to disk. The same helper the cap/layout/short-normalization paths already use — so every return path in the function now preserves SESSIONS authority.

Also made `_full_load_head` null-safe per your note: `cls.load(sid)` can return `None` if the sidecar disappears during the scan (TOCTOU); the helper now returns `([], total_count)` in that case rather than dereferencing `None`.

## Strengthening (per your ask)

- **Stale-count parametrization lowered from `15` to `1`** (at/below `raw_target=10`). `15` already triggered the old fallback; `1` is the strongest fail-first value — the pre-fix gate `total_count > len(raw_messages)` would have been `1 > 10` = False, skipping the fallback entirely.
- **Clean-streaming test now spies on `Session.load`** via `monkeypatch.setattr(Session, "load", ...)` and asserts it is NOT called for a clean uncached session. If a future change routes everything through the full loader, this assertion fires — locks in the performance contract, not just correctness.

## New barrier regression

`test_broad_exception_fallback_preserves_cache_authority` mirrors your sandbox barrier:
- Inserts the cache entry from within the scanner's `_scan_to_messages_array` call (deterministic — no timing coupling between threads), then raises
- Asserts the returned head is `CACHE_AUTHORITY`, not `STALE_DISK`
- Bounded 10-second thread join proves `_full_load_head` does not hold LOCK while calling `get_session` (which reacquires it) — i.e., no self-deadlock

### Regression gate
Barrier test fails pre-fix with the exact signature you described: `got ['STALE_DISK']`. Passes with the fix.

15/15 in `test_content_search_partial_load.py` pass; 36/36 across the broader search + `test_empty_session_no_disk_write` surface. Lint diff gate clean.

## Invariant now uniform

Every return path in `load_messages_head` preserves SESSIONS authority:
- cache hit (initial check)
- streaming success
- cap-exhaustion fallback
- layout-anomaly (array never opened) fallback
- short-normalization (collapse) fallback
- **broad exception** ← was the last holdout, now routed through the same helper

The head is always correctness-equivalent to `get_session(sid).messages[:limit]`.

Re-push is in (`98f001b`); ready for re-gate.

rh-id:
Closing as superseded by #6084 (`get_session_for_scan`), which shipped on `master` today in `4d9965b` (release #6343).

## Why I'm closing

The rebase conflict against `master` made the architectural divergence clear:

- **This PR (#6138)** solved the memory cost of content search by streaming the first `depth` messages via `load_messages_head` — bypassing `Session.load` and the `SESSIONS` cache to avoid materializing the full transcript. Every gate round found a new cache-authority edge case because streaming sits outside the canonical resolver.
- **#6084 (shipped)** solved the actual production complaint — *scan-pollution of the SESSIONS LRU* — by routing the scan through a cache-transparent accessor (`get_session_for_scan` → `_resolve_session(promote_cache=False, cache_on_miss=False)`). No streaming, no authority edge cases, correctness-by-construction.

The #6084 release note confirms the diagnosis: the multi-second searches that motivated looking at this path were **contention, not scan cost** (a full read+parse of ~1700 real sessions measures ~4s total; a parallel `/api/profiles` took 9.2s in the same window). Memory per session was never the bottleneck — the buffer-pool eviction was.

## What shipped from #6084 that supersedes each #6138 gate round

| #6138 gate finding | How #6084 closes it (by construction) |
|---|---|
| Cached active/unsaved messages invisible | `get_session_for_scan` reuses the resident cached session if present (canonical resolver) — never bypasses cache |
| Cached session with no sidecar invisible | Same — `_resolve_session` returns the resident entry regardless of disk state |
| `max_bytes` cap truncation | No cap — reads full transcript the canonical way |
| Duplicate `_partial` shifts depth window | `_resolve_session` runs through `Session.load`, which applies `_collapse_adjacent_duplicate_partials` |
| Broad-exception fallback bypasses cache authority | No fallback path — `_resolve_session` either resolves or returns None |
| LRU eviction of the working set during scan | **The actual problem #6084 was fixing** — `promote_cache=False, cache_on_miss=False` |

## What to keep

The regression tests added here (`test_content_search_partial_load.py`) are good explorations of the depth-window / partial-collapse / cache-authority space and informed the analysis, but they target an internal (`load_messages_head`) that #6084 deliberately does not introduce. They would need to be rewritten against `get_session_for_scan` to ship — at which point they'd be duplicating #6084's own `test_sessions_search_depth_validation` coverage.

Thanks @nesquena-hermes for the rigorous multi-round gate — the adversarial testing was exactly what surfaced the architecture mismatch. The right call was upstream's, and the gate is what made that legible.

PR #6140, opened 2026-07-16
fix(models): cap _CLI_SESSIONS_CACHE with LRU eviction (drop-oldest)

rh-id:
## Thinking Path

- The CLI/cron session projection cache (`_CLI_SESSIONS_CACHE`) backs the CLI/cron sidebar rows. Each value is a `copy.deepcopy()` of the full session list — the expensive projection behind #4842/#4672.
- It was a plain `dict` with TTL storage but **no size cap** and only lazy, same-key eviction. The cache key folds in a state.db content fingerprint that advances on every streamed message, so the next poll builds a NEW key; the previous key's heavy deepcopy was never looked up again, so its lazy-expiry path never ran.
- A long-lived process under churn could accumulate orphaned heavy deepcopies until the next structural `clear_cli_sessions_cache()`.

## What Changed

- Make `_CLI_SESSIONS_CACHE` an `OrderedDict` capped at `_CLI_SESSIONS_CACHE_MAX_ENTRIES` (8) with drop-oldest on write (`while len > MAX: popitem(last=False)`) and `move_to_end` on fresh hits.
- Mirrors the `_CLAUDE_CODE_PARSE_CACHE` / `_SIDECAR_METADATA_CACHE` LRU pattern already in the same file. TTL remains the primary freshness control; the cap is the backstop the plain dict lacked.
- `clear_cli_sessions_cache()` keeps `.clear()`.

## Why It Matters

Caps worst-case memory for a long-lived process under heavy streaming churn. Orphaned heavy deepcopies can no longer accumulate between structural invalidations — the LRU bound evicts the oldest entries once the cap is reached.

## Verification

- `tests/test_cli_sessions_cache_cap.py` (new, 3 tests): the cache stays at `len == MAX` after `MAX+5` distinct keys (oldest 5 evicted); a fresh hit marks the entry most-recently-used so it survives over colder entries; the cache is an `OrderedDict` (the regression: it was an unbounded plain dict).
- Concurrency verified in-process: 8 concurrent writers inserting 160 distinct keys stay bounded at cap with no errors. Both read and write paths hold `_CLI_SESSIONS_CACHE_LOCK` around `move_to_end`/`popitem`.
- Existing `test_streaming_cache_ttl_vs_poll.py` and `test_1079_cron_session_project.py` pass unchanged. All run via `./scripts/test.sh`.

## Risks / Follow-ups

- The cap value (8) is deliberately small because each entry is a full `deepcopy` of the session list. If the CLI projection grows or the profile count increases, this may need revisiting — but TTL/invalidation still drive correctness, so a too-small cap only means more cache misses (rebuilds), never wrong results.
- Eviction cannot change sidebar ordering or row content (Invariant 4) — the next poll recomputes the projection from state.db.

## Contract Routing

- Task type: memory-bound fix (unbounded projection cache).
- Touched areas: CLI/cron sidebar projection cache (`api/models.py` `_CLI_SESSIONS_CACHE`).
- Relevant public docs: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `docs/rfcs/webui-run-state-consistency-contract.md` (Sidebar/session metadata layer).
- State layer mutated: **Sidebar/session metadata layer** — specifically the in-memory projection cache that backs the CLI/cron sidebar rows. The cache is a read-derivation of state.db content; it is NOT a source of truth. Dropped entries are rebuilt on the next poll.
- Invariant preserved — **Invariant 4 (Maintenance is not activity)**: the cache is a derived projection; evicting an entry cannot move a session in the sidebar or change a row's content, because the next poll recomputes the projection from state.db. Regression: `test_cli_sessions_cache_cap.py` plus the existing streaming-cache-TTL and cron-session tests.
- Scope boundaries: does not change sidebar ordering, unread state, `updated_at`, or the projection result.

Release note: the CLI/cron sidebar projection cache is now LRU-bounded, capping worst-case memory for long-lived processes under heavy streaming churn.

## Model Used

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -63,7 +63,16 @@ _CLI_SESSIONS_CACHE_STREAMING_TTL_SECONDS = 45.0
 _CLI_SESSIONS_CACHE_LOCK = threading.Lock()
 _CLI_SESSIONS_CACHE_INFLIGHT: "dict[tuple, threading.Event]" = {}
 _CLI_SESSIONS_CACHE_INVALIDATION_VERSION = 0
-_CLI_SESSIONS_CACHE = {}
+# LRU-bounded (drop-oldest) so a long-lived process under churn — where the
+# state.db fingerprint advances on every streamed message and the structural
+# clear-on-mutation listener doesn't fire for every fingerprint advance — can't
+# accumulate orphaned heavy deepcopies. Each value is a copy.deepcopy() of the
+# full CLI/cron session list (the expensive projection behind #4842/#4672), so
+# the cap is deliberately small. TTL is still the primary freshness control;
+# the cap is the backstop that the plain dict previously lacked. Mirrors the
+# _CLAUDE_CODE_PARSE_CACHE / _SIDECAR_METADATA_CACHE LRU pattern.
+_CLI_SESSIONS_CACHE: "collections.OrderedDict[tuple, tuple]" = collections.OrderedDict()
+_CLI_SESSIONS_CACHE_MAX_ENTRIES = 8
 _CLI_SESSIONS_CACHE_WAIT_SECONDS = 0.25
 # Event waits that keep stale rows visible while a rebuild is in flight.
 _CLI_SESSIONS_CACHE_STALE_WAIT_SECONDS = 0.10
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
🚀 Shipped in **exp-v0.52.88** (experimental channel). Thanks @rh-id — clean LRU-cap on `_CLI_SESSIONS_CACHE` (OrderedDict, cap 8, drop-oldest, `move_to_end` on hit), mirroring the in-file `_CLAUDE_CODE_PARSE_CACHE`/`_SIDECAR_METADATA_CACHE` pattern. This was the one PR of your 5-part memory-hardening series that cleared the full gate (Codex + full single-process suite 13,183 green + CI 18/18) with zero findings.

The other four (#6137/#6139/#6141/#6153) each got a real defect flagged by the regression gate — fix-specs posted on each. Now live on the dogfood box.

PR #6148, opened 2026-07-16
fix(backend): resolve model prefix through _PROVIDER_ALIASES before matching (#6131)

### LINKED ISSUE #6131: Settings model switch doesn't persist `model.provider` for cross-provider model IDs (e.g. `z-ai/glm-5.2`) → agent 404s against the previous provider
## Environment

- hermes-webui **v0.52.41** (commit `0191ae5`)
- hermes-agent **v0.18.2 (2026.7.7.2)**
- Fresh native install; single `default` profile.

## Summary

Switching the active model to one whose model ID belongs to a **different provider than the one already in `config.yaml`** does not update `model.provider`. The WebUI leaves `model.provider` at its previous value, so the agent routes the new model to the **old** provider and the request fails with `HTTP 404 model not found`. Adding the new provider's API key in Settings succeeds (it lands in `.env`), which makes it look like the provider was configured — but the active provider never actually changes.

## Steps to reproduce

1. Start from a config where `model.provider: anthropic` (e.g. Anthropic configured during setup).
2. In the WebUI **Settings**, add an **NVIDIA** API key. → `NVIDIA_API_KEY` is written to `~/.hermes/.env` ✓
3. In the model picker, select the NVIDIA model **`z-ai/glm-5.2`** and **Save**.
4. Start a chat.

### Expected

`~/.hermes/config.yaml` `model.provider` becomes `nvidia` (matching the selected model), and the request goes to NVIDIA NIM.

### Actual

`~/.hermes/config.yaml` still reads:

```yaml
model:
  provider: anthropic
  default: "z-ai/glm-5.2"   # model changed, provider did NOT
```

so the agent sends `z-ai/glm-5.2` to `api.anthropic.com` →

```
HTTP 404: Model not found: z-ai/glm-5.2
The selected model was not found by the provider.
```

(Verified independently: the same NVIDIA key + `z-ai/glm-5.2` is listed by `GET https://integrate.api.nvidia.com/v1/models`, and inference works fine on that key for other models — so the key and the model ID are valid; the failure is purely the stale `model.provider`.)

## Root cause (traced in v0.52.41)

**Backend — `set_hermes_default_model` (`api/config.py:4257`)** writes:

```python
persisted_provider = str(requested_provider or resolved_provider or previous_provider or "").strip()
if persisted_provider:
    model_cfg["provider"] = persisted_provider
```

So the persisted provider is `requested_provider` (from the POST) **or** `resolve_model_provider()`'s guess **or** the *previous* provider. Two ways this lands on the stale value:

1. **`resolve_model_provider()` (`api/config.py:2512`) can't infer the provider for `z-ai/glm-5.2`.** The raw model-prefix check is never run through the alias table — `_PROVIDER_ALIASES` (`config.py:1193-1196`) maps `"z-ai" → "zai"`, but the prefix branch compares the raw `z-ai` against `_PROVIDER_MODELS`/`config_provider`, misses, and falls through to `return model_id, config_provider, config_base_url` (`config.py:2829-2832`) — i.e. it returns the **unchanged previous provider** (`anthropic`).
2. **Frontend `_modelStateForSelect` (`api/../static/ui.js:2672`)** reads `opt.dataset.provider` from whichever `<option>` the DOM currently reports as selected, but `_applyModelToDropdown` (`ui.js:2965`) sets `select.value = "z-ai/glm-5.2"` (a bare string), which per the HTML spec selects the **first** `<option>` with that value — dropping the provider-aware disambiguation and so POSTing an empty/wrong `provider` to `/api/default-model` (`api/routes.py:14055`).

Separately, `set_provider_key` (`api/providers.py:2792`, the "add API key" action) only writes `~/.hermes/.env` and never touches the `config.yaml` `model` block — expected, but it's why the key appears configured while the provider silently stays stale.

## Related / prior art

This looks like a currently-**unreported** instance of a bug class the project is already patching:

- **PR #5907 (open, unmerged)** — patches `_modelStateForSelect` to prefer `<option>` `data-provider` over string-splitting; its description explicitly mentions fixing *"settings provider corruption"*. This is the exact function on the Settings-save path.
- **#2488 (closed)** — "`/model` command selects wrong provider when same model ID exists across multiple providers" (repro uses a GLM model colliding across providers). Fixed only for the `/model` command call site, not the Settings flow.
- **#6121 / #5511 (closed)**, **#3405 (closed)** — same "silent stale/implicit-provider fallback" symptom for `custom_providers`/profile routing.
- **#586 / #604 (closed)** — the ancestor features that added post-onboarding provider management; #586's scope is key update (`.env`) only, matching the `set_provider_key` behavior above.

I couldn't find an existing issue covering this precise path (Settings add-key + select a cross-provider model ID + `config.yaml` `model.provider` left stale + agent 404). Filing in case it's useful as a concrete repro alongside #5907.

## Workaround

Set the provider explicitly rather than relying on the Settings model dropdown for cross-provider switches:

```bash
hermes model                       # interactive: sets model.provider + model.default correctly
# or edit ~/.hermes/config.yaml:
#   model:
#     provider: nvidia
#     default: "z-ai/glm-5.2"
```

## Suggested fix

Run the model-prefix branch in `resolve_model_provider()` through `_PROVIDER_ALIASES`, and/or land PR #5907 so `_modelStateForSelect` derives the provider from the selected `<option>`'s `data-provider` instead of `select.value` string-matching. A chat-start validator that also covers cloud providers (cf. open PR #5205, which currently only validates local/self-hosted) would surface the mismatch instead of a bare 404.

I hope it is not duplicate (I've search the similarissue, but couldn't finf anything).

Best regards.
Draqun

webtecnica:
## Summary

Fixes the backend half of the cross-provider model switch bug. When a model ID's prefix differs from the canonical provider slug (e.g. `z-ai/glm-5.2` → provider `nvidia`), `resolve_model_provider()` now normalizes the prefix through `_PROVIDER_ALIASES` before matching against `_PROVIDER_MODELS`.

**Before:** `prefix = "z-ai"` → `"z-ai" not in _PROVIDER_MODELS` → fall through → returns stale `config_provider` (e.g. `anthropic`)  
**After:** `prefix = "z-ai"` → `_resolve_provider_alias("z-ai")` → `"zai"` → `"zai" in _PROVIDER_MODELS` → correctly returns provider

## Related

- **#5907** (open, unmerged) covers the frontend half (`_modelStateForSelect` provider-aware matching). Once both land, the full #6131 flow works end-to-end.
- **Workaround:** `hermes model` CLI or manual `config.yaml` edit

Closes #6131

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -2957,7 +2957,8 @@ def resolve_model_provider(model_id: str, *, explicitly_picked: bool = False) ->
         # branch (#3872).
         _cp_lower_cross = (config_provider or "").strip().lower()
         _is_custom_cross = _cp_lower_cross == "custom" or _cp_lower_cross.startswith("custom:")
-        if prefix in _PROVIDER_MODELS and prefix != config_provider and not _is_custom_cross:
+        _canon_prefix = _resolve_provider_alias(prefix)
+        if _canon_prefix in _PROVIDER_MODELS and prefix != config_provider and not _is_custom_cross:
             return model_id, "openrouter", None
 
     return model_id, config_provider, config_base_url
--- a/tests/test_model_resolver.py
+++ b/tests/test_model_resolver.py
@@ -121,6 +121,25 @@ def test_cross_provider_routes_through_openrouter():
     assert base_url is None  # openrouter uses its own endpoint
 
 
+def test_cross_provider_routes_through_openrouter_with_alias_prefix():
+    """A model whose prefix is a _PROVIDER_ALIASES entry (not itself a
+    _PROVIDER_MODELS key) but resolves to a canonical provider key still
+    routes through the cross-provider OpenRouter branch.
+
+    Regression test for #6131 — before the fix, ``z-ai/glm-5.2`` with
+    config provider ``anthropic`` would fall through and return the stale
+    config provider because ``z-ai`` was checked directly against
+    ``_PROVIDER_MODELS`` instead of being resolved through
+    ``_PROVIDER_ALIASES`` first.
+    """
+    model, provider, base_url = _resolve_with_config(
+        'z-ai/glm-5.2', provider='anthropic',
+    )
+    assert model == 'z-ai/glm-5.2'
+    assert provider == 'openrouter'
+    assert base_url is None  # openrouter uses its own endpoint
+
+
 # ── Bare model names ─────────────────────────────────────────────────────
 
 def test_bare_model_uses_config_provider():


nesquena-hermes (changes requested):
Thanks for the fix for #6131! Before this can move forward there are two blocking issues.

### 1. Fatal `IndentationError` — the module doesn't parse

The inserted line in `resolve_model_provider` (`api/config.py`) is indented 16 spaces, one level deeper than the statements around it, with no block-opener before it:

```python
        _cp_lower_cross = (config_provider or "").strip().lower()
        _is_custom_cross = _cp_lower_cross == "custom" or _cp_lower_cross.startswith("custom:")
                _canon_prefix = _resolve_provider_alias(prefix)      # <-- 16 spaces, unexpected indent
        if _canon_prefix in _PROVIDER_MODELS and prefix != config_provider and not _is_custom_cross:
            return model_id, "openrouter", None
```

This raises `IndentationError: unexpected indent` at import time. Since `api/config.py` is imported transitively by `api/routes.py` and most of the backend, the whole app fails to start and the test suite errors at collection.

**Fix:** dedent `_canon_prefix = _resolve_provider_alias(prefix)` from 16 to 8 spaces so it sits at the same level as `_cp_lower_cross` / `_is_custom_cross` and the following `if`.

### 2. The branch is ~990 commits behind master — please rebase

The diff shows +1174/-226 across `api/config.py`, but that's an artifact of a very stale fork point. Measured against current `master`, the real net change is a single 3-line hunk (the one above). Everything else the diff appears to touch — the `_canonicalise_provider_id` DISPLAY-or-MODELS accept-gate, the `_configured_model_ids` / `_configured_model_options` helpers, the `global cfg` / `_cfg_cache` rebind, and the profile-home early-return — is **already on master**.

Please rebase `feat/...` onto current `master` and force-push so the PR reflects only the intended one-hunk change.

### 3. Add a regression test

Once the above is fixed, please add a targeted test (e.g. in `tests/test_model_resolver.py` or `tests/test_issue3368_model_prefix_shadowing.py`) asserting that a model whose prefix is an **alias** (not itself a `_PROVIDER_MODELS` key) but which resolves to a canonical key routes correctly through the new `_canon_prefix in _PROVIDER_MODELS` branch. Per repo policy, non-trivial routing changes need coverage.

Happy to re-review as soon as the indent is fixed and the branch is rebased. Thanks!

nesquena-hermes:
## Summary

The diagnosis behind this PR is correct — I confirmed the `z-ai/glm-5.2` → wrong-provider bug still reproduces on today's `origin/master`. But this branch cannot be merged as-is: its merge base is `4e8363035` (PR #5486), which is **990 commits behind master**, and `api/config.py`'s resolver has been rewritten several times since (notably #5979 custom-proxy vendor-namespace work and #6127 list-shaped custom models). That's why the diff shows 1,174 insertions / 226 deletions for a one-line-idea fix — it's almost entirely stale-base drift, not your change.

## The real bug still exists on master

`resolve_model_provider()` (`api/config.py:2592`) matches the raw vendor prefix against `_PROVIDER_MODELS` at several sites without canonicalising it first:

```python
# api/config.py:2938, 2944, 2960 (origin/master)
if prefix in _PROVIDER_MODELS and _is_first_party_model(prefix, bare):
    return bare, config_provider, config_base_url
...
if prefix in _PROVIDER_MODELS:
    return bare, config_provider, config_base_url
...
if prefix in _PROVIDER_MODELS and prefix != config_provider and not _is_custom_cross:
    return model_id, "openrouter", None
```

`_PROVIDER_MODELS` is keyed by canonical slug (`zai`), so `prefix == "z-ai"` misses every one of these, falls through, and returns the stale `config_provider` — the 404 in #6131. Note master already canonicalises `config_provider` for the *advertised-models* check (`_canon_config_provider = _canonicalise_provider_id(config_provider)` at `api/config.py:2670`), so the prefix path is the remaining gap.

## Recommendation

Rebase onto current `origin/master` and reduce the change to the actual fix: canonicalise the prefix before each `_PROVIDER_MODELS` membership test. Master already has the helper — `_canonicalise_provider_id()` (`api/config.py:1468`) and `_resolve_provider_alias()` (`api/config.py:1269`, which now merges the agent's table). A minimal form:

```python
_canon_prefix = _canonicalise_provider_id(prefix)
if _canon_prefix in _PROVIDER_MODELS and _is_first_party_model(_canon_prefix, bare):
    return bare, config_provider, config_base_url
```

applied at the three sites above (guarding the `!= config_provider` comparisons against the canonical form too).

## Cross-repo contract

The alias table this depends on is authoritative on the **agent** side — `hermes_cli/models.py:1220`:

```python
_PROVIDER_ALIASES = {
    "glm": "zai",
    "z-ai": "zai",
    "z.ai": "zai",
    "zhipu": "zai",
}
```

`_resolve_provider_alias()` already imports and prefers that table (`api/config.py:1283`), so using it here keeps the WebUI in sync with the agent automatically — no need to hardcode a second copy in the resolver. That's the right layering: canonicalise via the shared alias table, then match.

## Related work

The PR body notes #5907 covers the frontend half (`_modelStateForSelect` provider-aware matching). Worth confirming that's still open and rebased too, since #6131 only fully closes when both the picker-side match and this backend resolution land against the same master.

## Verification

After rebasing + narrowing, `resolve_model_provider("z-ai/glm-5.2")` with `config_provider` set to a different provider should return provider `zai` (not the config provider), and the existing resolver tests should pass unchanged. A targeted unit case asserting the `z-ai` → `zai` prefix canonicalisation would lock the regression.

webtecnica:
Corrigido — a indentação estava em 16 espaços sem block-opener. Reduzi para 8 espaços e o arquivo agora parseia corretamente. Obrigado pelo review detalhado! 🙏

webtecnica:
Fixed: rebased onto current `upstream/master` and applied the one-hunk fix directly on top (the old branch was ~4,400 commits behind). The diff now reflects only the intended change: `_canon_prefix = _resolve_provider_alias(prefix)` before checking `_canon_prefix in _PROVIDER_MODELS`. The IndentationError from the prior commit was already fixed.

webtecnica:
All three blocking issues addressed — ready for re-review:

### 1. ✅ IndentationError fixed
`_canon_prefix = _resolve_provider_alias(prefix)` now uses 8-space indentation (matching the surrounding block). Verified: `python3 -c "import ast; ast.parse(open('api/config.py').read())"` passes.

### 2. ✅ Rebased onto master
Branch force-pushed on top of current upstream/master. Diff is now clean — only the intended 2-file change (`api/config.py` +1 line, `tests/test_model_resolver.py` +19 lines).

### 3. ✅ Regression test added
`test_cross_provider_routes_through_openrouter_with_alias_prefix` in `tests/test_model_resolver.py`:
- Uses `z-ai/glm-5.2` (prefix `z-ai` is an alias → resolves to canonical `zai`)
- Config provider set to `anthropic` (different from the model's provider)
- Asserts it routes through `openrouter` (the canonical `zai` IS in `_PROVIDER_MODELS`)
- All 63 tests in the file pass (62 existing + 1 new)

webtecnica:
✅ IndentationError fixed, rebased onto upstream/master (no more 990-commit artifact), and regression test added for the alias-prefix → canonical resolution path. Ready for re-review.

nesquena-hermes:
Shipped in **exp-v0.52.141** — thanks @webtecnica! 🎉

Your diagnosis and fix were correct; the only blocker was a stale merge base (990 commits), which I resolved by rebasing onto current master. During the gate, Codex flagged that resolving through _PROVIDER_ALIASES broke a couple of edges (Qwen canonicalises to a key that is not in _PROVIDER_MODELS via that path, and an alias-of-active-provider selection cross-routed due to a raw-vs-canonical asymmetry), so I switched both the membership check and the equality check to the idempotent _canonicalise_provider_id and added two regression tests (Qwen-under-Anthropic stays on OpenRouter; z-ai-under-zai stays on zai). Codex re-gate SAFE TO SHIP, full serial suite 13,599 passed. Your attribution is preserved in the CHANGELOG and commit history.

PR #6149, opened 2026-07-16
fix(server): reap exited children via SIGCHLD handler to prevent zombie accumulation (#6145)

### LINKED ISSUE #6145: Zombie [hermes] <defunct> processes accumulate under hermes-webui.service (no SIGCHLD reaper)
## Summary
A `[hermes] <defunct>` zombie process is left behind under `hermes-webui.service` after a `hermes gateway restart` is triggered from the WebUI. The server process never reaps the exited `hermes` child, so it lingers as a zombie.

## Environment
- OS: Ubuntu 24.04, Linux 5.15.0-177, systemd **user** service `hermes-webui.service`
- hermes-webui commit `3f3fa9b6` (branch `master`), Python 3.11 venv
- Observed: 1 zombie persisting from `Jul15` (process had exited, parent never reaped)

## Symptom
```
$ ps -eo pid,ppid,stat,comm | awk '$3 ~ /Z/'
209658  194812 Z    herme  [hermes] <defunct>
```
- PID 209658 = `[hermes] <defunct>`
- PPID 194812 = `/usr/local/lib/hermes-agent/venv/bin/python /root/hermes-webui/server.py`
- Restarting the service clears it (confirming the *parent* failed to reap, not a stuck child).

## Root cause (analysis)
`server.py` installs signal handlers only for `SIGPIPE` and `SIGTERM` (verified by grep across the repo). There is **no `SIGCHLD` reaper** and no global `waitpid(-1)` fallback anywhere.

`api/gateway_restart.py::restart_active_profile_gateway` runs `hermes gateway restart` via `subprocess.Popen`. On the non-blocking `in_progress` path (command exceeds `quick_timeout_seconds = 2.0`), reaping is delegated to the daemon thread `_wait_and_release`:
```python
threading.Thread(target=_wait_and_release, daemon=True).start()
```
If that thread is lost before it reaches `proc.wait()` (server reload, thread exception, etc.), the child is never reaped and becomes a zombie. The binary name (`hermes`, not `bash`/`sh`) matches the gateway-restart path rather than the terminal/background spawners.

Other spawn paths (terminal.py) do have a pgid-scoped reaper, but nothing covers stray children globally — so any future async spawn that loses its waiter will leak a zombie.

## Suggested fix
Add a global `SIGCHLD` handler in `server.py` (next to the existing `signal.signal(...)` calls) so every exited child is reaped regardless of per-path tracking:
```python
import os, signal

def _reap_children(_signo, _frame):
    """Reap any exited child so they never linger as zombies."""
    try:
        while True:
            pid, _status = os.waitpid(-1, os.WNOHANG)
            if pid == 0:
                break
    except ChildProcessError:
        pass

signal.signal(signal.SIGCHLD, _reap_children)
```
`waitpid(-1, WNOHANG)` only collects already-exited children and is non-blocking, so it never disturbs running subprocesses. (On modern CPython, `subprocess.wait()` tolerates an already-reaped child via `ChildProcessError` and stays correct.)

A reaper thread polling `waitpid(-1, WNOHANG)` every 1–2s is an equivalent alternative if a signal handler is undesirable.

## Verification
1. `ps -eo pid,ppid,stat,comm | awk '$3 ~ /Z/'` → shows `[hermes] <defunct>` with PPID = `server.py`.
2. Apply the patch and restart `hermes-webui.service`.
3. Trigger a `hermes gateway restart` from the WebUI (or let one run > 2s to hit the `in_progress` path).
4. Re-check `ps` → no `[hermes] <defunct>` should remain.

webtecnica:
## Resumo

Adiciona um handler global `SIGCHLD` em `server.py` para eliminar processos zumbis filhos que se acumulam sob `hermes-webui.service`.

## Contexto

A issue #6145 reporta que processos `[hermes] <defunct>` (zumbis) se acumulam quando `api/gateway_restart.py::restart_active_profile_gateway` executa `hermes gateway restart` via `subprocess.Popen` no caminho não-bloqueante (>2s). Nesse caminho, a coleta do processo filho é delegada a uma daemon thread `_wait_and_release`. Se essa thread é perdida (recarga do servidor, exceção, etc.), o filho nunca é coletado e vira um zumbi.

## O que foi feito

Adicionado um handler `SIGCHLD` global em `server.py` (imediatamente após o handler `SIGPIPE` existente) que coleta **qualquer** filho já terminado via `os.waitpid(-1, os.WNOHANG)`. Isso garante que nenhum processo filho órfão ou com thread de coleta perdida se torne um zumbi.

## Mudanças

- **`server.py`**: Nova função `_reap_children()` registrada como handler `signal.SIGCHLD`. A função usa `waitpid(-1, WNOHANG)` em loop para coletar todos os filhos já terminados de forma não-bloqueante.

## Verificação

1. `ps -eo pid,ppid,stat,comm | awk '$3 ~ /Z/'` → mostra `[hermes] <defunct>` com PPID = `server.py` (antes)
2. Após aplicar o patch e reiniciar `hermes-webui.service`, acionar `hermes gateway restart` da WebUI
3. `ps` → nenhum `[hermes] <defunct>` deve permanecer (depois)

Closes #6145

### DIFF
--- a/server.py
+++ b/server.py
@@ -16,6 +16,23 @@ _SIGPIPE = getattr(signal, "SIGPIPE", None)
 if _SIGPIPE is not None:
     signal.signal(_SIGPIPE, signal.SIG_IGN)
 
+# Reap exited children via SIGCHLD so subprocess zombies never accumulate.
+# Without this, any async spawn whose reaper thread is lost (server reload,
+# exception, etc.) leaves a [hermes] <defunct> zombie lingering forever.
+# The loop calls waitpid(-1, WNOHANG) until no more exited children remain;
+# it is non-blocking and never disturbs running subprocesses.
+def _reap_children(_signo, _frame):
+    """Reap any exited child so they never linger as zombies."""
+    try:
+        while True:
+            pid, _status = os.waitpid(-1, os.WNOHANG)
+            if pid == 0:
+                break
+    except ChildProcessError:
+        pass
+
+signal.signal(signal.SIGCHLD, _reap_children)
+
 # Test-mode network isolation keeps subprocess-backed tests hermetic.
 if os.environ.get("HERMES_WEBUI_TEST_NETWORK_BLOCK", "").strip() in ("1", "true", "yes"):
     _REAL_CREATE_CONN = socket.create_connection


nesquena-hermes:
## Summary

The zombie leak in #6145 is real, and reaping the async gateway-restart child is the right goal. But a process-wide `SIGCHLD` handler that calls `os.waitpid(-1, WNOHANG)` is a heavier hammer than this needs, and it introduces a subtle correctness regression across the ~30 *synchronous* subprocess callsites in the codebase. I'd push back on this shape before merging.

## The hazard: reaped-child returncode falsification

The issue's own analysis says a global reaper is safe because "on modern CPython, `subprocess.wait()` tolerates an already-reaped child via `ChildProcessError` and stays correct." That's half right: it stays *crash-free*, but it does **not** stay correct. CPython's `Popen._try_wait` swallows `ChildProcessError` and substitutes a **success status**:

```python
def _try_wait(self, wait_flags):
    try:
        (pid, sts) = os.waitpid(self.pid, wait_flags)
    except ChildProcessError:
        # This child is dead, we can't get the status.
        pid = self.pid
        sts = 0          # <-- returncode becomes 0
    return (pid, sts)
```

So if the global handler wins the race and reaps a child before a blocked `subprocess.run(...)` / `.communicate()` does, that call returns `returncode == 0` regardless of the child's real exit code. Every failure-detecting branch then silently sees "success."

## Blast radius

This repo has many synchronous subprocess calls whose `returncode` drives correctness, and they run in Flask worker threads while the `SIGCHLD` handler runs in the main thread — a genuine cross-thread `waitpid` race:

- `api/workspace_git.py:253` — 16 `returncode` branches gate git operations
- `api/rollback.py:102,136,250` — `if result.returncode != 0: raise` (checkpoint restore correctness)
- `api/updates.py:216` — `if r.returncode == 0: return stdout, True`
- `api/startup.py:115` — `subprocess.run(install_args, ...)`
- `api/streaming.py:871`, `api/providers.py:1841` (`.communicate`), `api/routes.py:1230`

A dropped-to-0 returncode in `rollback.py` or `workspace_git.py` is a silent data-integrity bug that's nearly impossible to reproduce because it's timing-dependent.

## Recommendation

Two safer options:

1. **Scope the fix to the actual leak.** The zombie comes from one path — the lost `_wait_and_release` daemon thread in `api/gateway_restart.py:159`. Harden *that* (e.g. wrap the whole thread body so it can't die before `proc.wait()`, or track the pid in a small registry a dedicated reaper thread drains). This is the narrow fix.

2. **If you want a global safety net, use the polling-reaper alternative the issue itself offers** (a thread doing `waitpid(-1, WNOHANG)` on an interval) — but *only* reap pids that aren't owned by a live `Popen` object. The existing `api/background_process.py:417` `_REAPER_STOP` loop and `api/terminal.py:126` `_reap_abandoned_spawn` already establish the "reap only pids we've handed off" pattern; extending that is consistent with the codebase and doesn't race the sync callers.

If you keep the `SIGCHLD` approach, at minimum it needs a guard so it never reaps a pid that a `Popen` in another thread is actively waiting on — otherwise the workspace-git / rollback returncode paths become intermittently unreliable.

## Verification

The leak repro from the issue (`ps -eo pid,ppid,stat,comm | awk '$3 ~ /Z/'` after a WebUI-triggered restart) confirms fix #1 directly. To surface the regression this PR risks: run a git op through `api/workspace_git.py` that's expected to fail (non-zero exit) concurrently with the restart flow, and confirm the failure is still reported rather than swallowed as `returncode=0`.

PR #6152, opened 2026-07-16
fix(routes): clamp GET /api/session msg_limit to a server-side ceiling

rh-id:
## Thinking Path

- `GET /api/session` accepts `?msg_limit=N` to return a tail window of the last N visible transcript rows.
- The query param had **no upper bound** — a client could request `msg_limit=1000000` and force the server to assemble and serialize an unbounded message payload.
- The frontend's real pagination is modest: initial load `_INITIAL_MSG_LIMIT=30`, load-older grows by `+30`, `msg_before` paging uses 30. The only caller that asks for a huge value is `outline.js:110` (`msg_limit=9999`) for outline jumps.
- So the exposure is the outline-jump path and any pathological/hostile value, not normal pagination.

## What Changed

- Add `_MAX_MSG_LIMIT = 500` (generous — far above any legitimate visible-row window) and clamp at the query-parse site: `max(1, min(int(_msg_limit), _MAX_MSG_LIMIT))`.
- The existing `_messages_truncated` signal already covers the clamped case: when the request exceeds the ceiling, `_message_window_for_display` returns `_messages_offset > 0`, so the client learns more rows exist and load-older still works.
- The bare no-`msg_limit` path is **intentionally unchanged** — it is the documented "give me everything" escape hatch used by branch/undo/jump-to-start flows (`_ensureAllMessagesLoaded`) that need the full transcript for correct global indices. Silently bounding it would break those.

## Why It Matters

Bounds the worst-case `GET /api/session` payload. A single oversized request can no longer force the server to assemble and serialize an entire large transcript. This is the request-side complement to the bounded-read work — it doesn't help the disk-load cost (that's the separate `state.db` LIMIT issue), but it caps the response size and the serialization peak.

## Verification

- `tests/test_session_msg_limit_ceiling.py` (new, 3 tests): the ceiling constant is reasonable (`100..2000`); an over-ceiling request is clamped and signals truncation (`offset > 0`); legit sizes (5/30/60/100) are unaltered.
- Existing `test_session_message_window_renderable_tail.py` (10) and `test_session_tail_payload.py` (9) pass unchanged.

## Risks / Follow-ups

- The one behavior change: `outline.js`'s `msg_limit=9999` now receives up to 500 rows + `_messages_truncated=true`. Outline jumps load context around a target message, so 500 rows is ample and load-older still works — but flagging it as the single observable change.
- This caps the *response*, not the *disk load*. The backend still full-parses sidecars and full-scans `state.db` before slicing — that's the separate bounded-`state.db`-read issue (being addressed in a companion PR).

## Contract Routing

- Task type: memory-bound fix (unbounded `msg_limit` request payload).
- Touched areas: `GET /api/session` query parsing (`api/routes.py`).
- Relevant public docs: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`.
- State layer mutated: none (read-only endpoint response shaping).

## Contract Change (additive, backward-compatible)

- Previous: `?msg_limit=N` accepted any positive integer with no upper bound.
- New: `?msg_limit=N` is clamped to `_MAX_MSG_LIMIT = 500`.
- Compatibility: the only frontend caller above the ceiling is `outline.js:110` (`msg_limit=9999`), which now receives up to 500 rows + `_messages_truncated=true` (load-older still works). All other callers (30/60) are unaffected. The bare no-`msg_limit` path is unchanged.
- Reason: an oversized `msg_limit` forced an unbounded payload assembly + serialization; the ceiling bounds the worst case without touching legitimate pagination.

Release note: `GET /api/session` now caps `?msg_limit` at 500 rows server-side, bounding the worst-case payload for oversized requests.

## Model Used

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -8441,6 +8441,34 @@ def _message_window_for_display(messages, msg_limit=None, msg_before=None, expan
 
 
 _LIMITED_TOOL_CONTENT_MAX_CHARS = 4096
+# Server-side ceiling on the ?msg_limit= tail-window size. A client could
+# otherwise request msg_limit=1000000 and force the server to assemble and
+# serialize an unbounded message payload (the frontend's own pagination grows
+# by ~30 at a time, with one outline-jump path asking for 9999). The ceiling is
+# generous — far above any legitimate visible-row window — so real pagination is
+# unaffected; it only caps the pathological/oversized request. When the request
+# exceeds the ceiling the response is silently clamped and _messages_truncated
+# is set (the existing truncation signal already covers "more rows exist").
+_MAX_MSG_LIMIT = 500
+
+
+def _parse_msg_limit(raw):
+    """Parse and clamp the ``?msg_limit=`` query value.
+
+    Returns a positive int clamped to ``[1, _MAX_MSG_LIMIT]``, or ``None`` when
+    the value is absent/empty/malformed (the bare no-``msg_limit`` path, which
+    intentionally returns the full transcript for callers that need it).
+    Extracted from the handler so the clamp expression has direct test coverage.
+    """
+    if not raw:
+        return None
+    try:
+        value = int(raw)
+    except (TypeError, ValueError):
+        return None
+    return max(1, min(value, _MAX_MSG_LIMIT))
+
+
 _LIMITED_TOOL_CONTENT_NOTICE = (
     "\n\n[Tool output truncated in paginated session response; "
     "load the full transcript to inspect the complete result.]"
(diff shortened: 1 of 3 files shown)


rh-id:
Thanks for the review — the point about test coverage of the actual changed line is correct. Fixed in `b7ef4b3`.

## The gap (valid)

The three original tests drove `_message_window_for_display` with an already-clamped value, so none of them exercised the handler's actual new expression (`max(1, min(int(_msg_limit), _MAX_MSG_LIMIT))`). A future edit to that clamp line would have gone undetected. The `min(10_000_000, _MAX_MSG_LIMIT)` in one test was also misleading (always equals `_MAX_MSG_LIMIT`) — fair call.

## Fix

Extracted the parse + clamp into a small pure helper `_parse_msg_limit(raw)` and routed the handler through it:

```python
def _parse_msg_limit(raw):
    if not raw:
        return None
    try:
        value = int(raw)
    except (TypeError, ValueError):
        return None
    return max(1, min(value, _MAX_MSG_LIMIT))
```

The handler is now a one-liner: `msg_limit = _parse_msg_limit(query.get("msg_limit", [None])[0])`. This follows the codebase's "fix the class, not the instance" guidance — the clamp is no longer buried in a 1000-line handler.

## Tests rewritten to cover the actual clamp

`tests/test_session_msg_limit_ceiling.py` now drives `_parse_msg_limit` directly (the changed code), covering every branch of the clamp expression:
- absent / empty / malformed → `None` (the bare full-transcript path)
- legit sizes (5/30/60/100) → returned verbatim
- oversized (9999, 1000000, ceiling+1) → clamped to `_MAX_MSG_LIMIT`
- exactly at the ceiling → allowed (inclusive boundary)
- `0` / `-5` → clamped to `1`

The misleading `min(10_000_000, _MAX_MSG_LIMIT)` test is gone. All 26 tests in the run (7 here + 19 existing window/tail) pass.

nesquena-hermes:
Shipped in **exp-v0.52.98** as part of a combined release with #6154. Thanks @rh-id! 🎉

The server-side `_MAX_MSG_LIMIT = 500` clamp + `_parse_msg_limit()` helper landed as-is. During the pre-release adversarial gate, two silent row-loss regressions in the paired frontend paging (#6154) were found and fixed in a maintainer follow-up commit (co-authored to you) before ship — the backend clamp itself was clean. Also resolved the tracking follow-up #6177 framing. Appreciate the clean, well-tested pair.

PR #6153, opened 2026-07-16
fix(models): add defensive row backstop to display-path state.db message read

rh-id:
## Thinking Path

- `GET /api/session`'s state.db read (`get_state_db_session_messages`) had **no SQL `LIMIT`** — it full-scanned the messages table and `fetchall()`'d into a Python list.
- The display path already has a careful `since_timestamp` optimization (`_state_db_since_timestamp_for_limited_display`) that bounds the common uncompressed tail load — but it **deliberately bails** (returns `None`) on compressed sessions (`truncation_watermark`/`truncation_boundary`) and `msg_before` paging, because those cases need older reconciliation rows for correctness.
- In the bail cases the read full-scanned with no cap, so a pathological/huge state.db could materialize unbounded rows into memory. This is the direct backend analog of the H2 search issue — but with an important twist: a naive semantic-window `LIMIT` here would **corrupt the sidecar/state.db reconciliation** (the existing helper's docstring explicitly warns against SQL-LIMITing raw rows for exactly this reason).

## What Changed

- Add an opt-in `limit` parameter to `get_state_db_session_messages`, applied as a SQL `LIMIT` that keeps the **newest** rows (via a `DESC` subquery re-sorted `ASC`, since the base query orders oldest-first). It's a **backstop, not a semantic window**.
- Thread a generous `_STATE_DB_DISPLAY_ROW_BACKSTOP = 50000` from the `GET /api/session` display caller only. No legitimate session approaches this; it purely bounds the pathological case.
- The **full-history model-context callers do NOT pass a limit**: reconciliation (`models.py:3437`), context reconstruction (`models.py:8576`), `get_cli_session_messages` (`models.py:8632`), and new-turn context (`streaming.py:8443`). Their semantics require the full history, and are left unchanged.

## Why It Matters

Bounds the worst-case state.db materialization on the display path's bail cases (compressed sessions, `msg_before` paging) — exactly when the existing `since_timestamp` optimization steps aside. A pathological or externally-bloated state.db can no longer OOM the server on a status poll. This is narrower than "add LIMIT everywhere" — deliberately so, to preserve the reconciliation contract.

## Verification

- `tests/test_state_db_read_backstop.py` (new, 6 tests): no-limit returns full history oldest-first (model-context contract unchanged); `limit=N` keeps the newest N re-sorted oldest-first; `limit > row count` returns all; `limit<=0` clamps to 1 (newest); invalid limit falls back to unbounded; `limit` composes with `since_timestamp`.
- Existing `test_state_db_active_filter.py` (3), `test_webui_state_db_reconciliation.py` (47), `test_webui_state_db_context_reconciliation.py` (4) pass unchanged — 54 tests confirming the reconciliation/model-context paths are unaffected.

## Risks / Follow-ups

- The cap (50000) is a safety net, not a real bound for normal usage — if a legitimate session ever approaches it, the display would silently drop older rows from the reconciliation. That's far above any observed conversation size, and the `_messages_truncated` signal would surface it. If you'd prefer a lower ceiling or an explicit truncation flag distinct from the existing one, easy to adjust.
- This does not address the sidecar full-parse (the lineage stitcher does up to 20 `Session.load(parent_id)` full JSON reads) — that's a separate, larger change touching compression-lineage correctness.

## Contract Routing

- Task type: memory-bound fix (unbounded state.db row materialization on the display path's bail cases).
- Touched areas: state.db message reader (`api/models.py`) + `GET /api/session` display caller (`api/routes.py`). NOT the model-context callers.
- Relevant public docs: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `docs/rfcs/webui-run-state-consistency-contract.md` (Model context layer).
- State layer mutated: none (read-only). The model-context read path is **unchanged** (no limit passed) — Invariant 1 (visible current turns enter model context) and the reconciliation contract are preserved. Only the display path gets the defensive cap, which does not affect what the agent receives.
- Scope boundaries: this is intentionally narrower than a naive "add LIMIT everywhere." `_state_db_since_timestamp_for_limited_display`'s docstring explains why a semantic window LIMIT would corrupt the merge; this cap is purely a pathological-input safety net above any legitimate session size.

Release note: `GET /api/session`'s state.db read now has a defensive row backstop, capping memory when the bounded `since_timestamp` optimization bails on compressed/paged sessions.

## Model Used

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -7330,6 +7330,7 @@ def get_state_db_session_messages(
     profile=None,
     since_timestamp=None,
     include_inactive: bool = False,
+    limit=None,
 ) -> list:
     """Read messages for a Hermes session from state.db.
 
@@ -7345,6 +7346,14 @@ def get_state_db_session_messages(
     raw state.db scan to rows at or after a sidecar-derived timestamp floor while
     preserving the caller's normal merge/window logic.  Full-history callers must
     leave it unset.
+
+    ``limit`` is an optional defensive row cap (applied after ORDER BY as a SQL
+    LIMIT). It is a BACKSTOP against a pathological/huge state.db materializing
+    unbounded rows into a Python list, NOT a semantic window: the display path
+    counts visible rows post-reconciliation, so a true window LIMIT here would
+    corrupt the sidecar/state.db merge (see _state_db_since_timestamp_for_limited_display,
+    which deliberately does NOT SQL-LIMIT raw rows for that reason). Callers that
+    need the full history for model-context reconstruction leave this unset.
 
     When the messages table exposes an ``active`` column, inactive rows are
     compacted/archived history and are intentionally excluded by default. WebUI
@@ -7446,14 +7455,42 @@ def get_state_db_session_messages(
             active_clause = ""
             if 'active' in available and not include_inactive:
                 active_clause = " AND (active IS NULL OR active != 0)"
-            cur.execute(f"""
-                SELECT {', '.join(selected)}, session_id
-                FROM messages
-                WHERE session_id IN ({placeholders})
-                {since_clause}
-                {active_clause}
-                ORDER BY timestamp ASC, id ASC
-            """, params)
+            # Defensive row cap (backstop only — see docstring). Applied as a
+            # SQL LIMIT bound parameter (?) so the tail (newest) rows are
+            # retained and a pathological state.db can't materialize unbounded
+            # rows. None = unchanged full-history read for model-context callers.
+            limit_clause = ""
+            if limit is not None:
+                try:
+                    limit_int = max(1, int(limit))
+                except (TypeError, ValueError):
+                    limit_int = None
+                if limit_int is not None:
+                    # The query orders ASC (oldest first); to keep the NEWEST
+                    # rows under the cap, take a descending-ordered subquery and
+                    # re-sort ascending — a plain LIMIT would keep the oldest.
+                    limit_clause = " ORDER BY timestamp DESC, id DESC LIMIT ?"
+                    params.append(limit_int)
+            if limit_clause:
+                cur.execute(f"""
+                    SELECT * FROM (
+                        SELECT {', '.join(selected)}, session_id
+                        FROM messages
+                        WHERE session_id IN ({placeholders})
+                        {since_clause}
+                        {active_clause}
+                        {limit_clause}
+                    ) ORDER BY timestamp ASC, id ASC
+                """, params)
+            else:
+                cur.execute(f"""
+                    SELECT {', '.join(selected)}, session_id
+                    FROM messages
+                    WHERE session_id IN ({placeholders})
+                    {since_clause}
+                    {active_clause}
+                    ORDER BY timestamp ASC, id ASC
+                """, params)
             msgs = []
             for row in cur.fetchall():
                 msg = {
--- a/api/routes.py
+++ b/api/routes.py
@@ -8441,6 +8441,17 @@ def _message_window_for_display(messages, msg_limit=None, msg_before=None, expan
 
 
 _LIMITED_TOOL_CONTENT_MAX_CHARS = 4096
+# Defensive row backstop for the GET /api/session display path's state.db read.
+# This is NOT a semantic window (the display window counts visible rows
+# post-reconciliation via _message_window_for_display); it is a safety net so a
+# pathological/huge state.db cannot materialize unbounded rows into memory on the
+# display path. Legitimate sessions stay far below this; the compressed-session
+# case where _state_db_since_timestamp_for_limited_display bails (and would
+# otherwise full-scan) is the main beneficiary. Generous on purpose: no real
+# conversation approaches it, and the existing since_timestamp optimization
+# already handles the common tail-load case. The full-history model-context
+# callers (reconciliation, new-turn context) do NOT use this cap.
+_STATE_DB_DISPLAY_ROW_BACKSTOP = 50000
 _LIMITED_TOOL_CONTENT_NOTICE = (
     "\n\n[Tool output truncated in paginated session response; "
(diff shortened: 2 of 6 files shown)


rh-id:
Thanks for the review — the dead-variable point is correct, fixed in `dec0419`.

## The issue (valid)

`limit_params = []` was a leftover from an earlier parameterized-query draft. It was always empty, never mutated, and only ever concatenated into `params + limit_params` — purely misleading, since the limit value actually reached SQLite via an f-string interpolation (`LIMIT {limit_int}`).

## Fix

Switched the LIMIT to a bound `?` parameter (consistent with the rest of the query, which uses `?` placeholders throughout) and removed the dead variable entirely:

```python
limit_clause = " ORDER BY timestamp DESC, id DESC LIMIT ?"
params.append(limit_int)
```

…then the query executes with the single `params` list (no more `params + limit_params`). `limit_int` is still validated to a Python `int` via `max(1, int(limit))` before this point, so it was already injection-safe either way — the bound parameter is the cleaner, style-consistent form.

Verified: all 60 tests in the run (6 backstop + 3 active-filter + 47 reconciliation + 4 context-reconciliation) still pass, confirming the bound-parameter LIMIT behaves identically to the interpolated form.

nesquena-hermes (changes requested):
## Gate found a SILENT data-loss edge — fix before ship

Thanks @rh-id — I actually pushed the two ruff/test-stub CI fixes onto your branch to get the matrix green (unused-import cleanup + the pre-existing display-path test stubs needed the new `limit=` kwarg). The backstop mechanism itself (newest-rows-preserved DESC-subquery re-sorted ASC, fail-open on bad values, display-path-only) is well built. But the regression gate (Codex, reproduced) found a **SILENT** correctness edge on the display path:

**`api/routes.py` (~12517) — a history over 50,000 state-db rows can silently lose its preserved prefix during merge.** The 50k newest-only SQL cap is applied *before* `merge_session_messages_append_only()`, which needs the rows at/around the session's `truncation_boundary` (the preserved prefix) to reconcile correctly. Capping to the newest 50k removes those boundary rows, and also prevents older-page (`msg_before`) requests from ever reaching them. So a pathological-but-real >50k-row session gets a corrupted merge / unreachable older history rather than a clean bound.

**Fix-spec:**
- Apply the row cap **only to provably-safe initial-tail reads** where no `truncation_boundary` prefix is required — i.e. gate it on the same condition the `since_timestamp` optimization uses, and leave the compressed/boundary and `msg_before` paths on the full (uncapped) read.
- Or: query the required boundary-prefix rows **plus** the bounded newest tail (two-part read), so the merge always has its prefix.
- Add a route-and-merge regression exceeding 50,000 rows that asserts the reconciled transcript is byte-identical to the uncapped read.

The `get_state_db_session_messages(limit=...)` signature + the NEWEST-rows subquery are fine; it's only the display-path *application* of the cap that needs the boundary-aware guard. Re-gate + ship once this lands. Credit preserved.

rh-id:
Thanks — your gate caught a silent data-loss edge I missed. Reproduced and fixed in `f445c3a`.

## The bug (you reproduced it)

The 50k newest-only SQL cap was applied *unconditionally* to the display path's state.db read. But `merge_session_messages_append_only` needs the rows at/around `truncation_boundary` (the preserved prefix) to reconcile correctly; the newest-N cap drops those boundary rows for a >N-row session, corrupting the merge (silently losing the prefix) and making older-page (`msg_before`) requests unable to reach them. My PR's own docstring said "a true window LIMIT would corrupt the merge" — and then I applied exactly such a cap on the paths that need the prefix. That's the contradiction you found.

## Fix (your option a)

Gate the backstop via `_state_db_backstop_limit_for_display(session, msg_before)` on the SAME conditions `_state_db_since_timestamp_for_limited_display` uses to decide a read is boundary-free: NOT `msg_before` paging, and NO `truncation_watermark` / `truncation_boundary`. Those cases stay on the full (uncapped) read — the backstop is now a defensive net for the common uncompressed initial-tail path only. Extracted into a pure helper so the gate has direct test coverage (the 1000-line handler is otherwise untestable in isolation).

The `get_state_db_session_messages(limit=)` signature + newest-rows DESC subquery are unchanged — only the display-path *application* of the cap is gated. This narrows the backstop's protection (it no longer covers the compressed bail case) but that's the correct correctness-over-boundedness tradeoff you specified.

## Regression tests

`tests/test_state_db_backstop_boundary_gate.py` (5 tests) drive the helper directly: backstop applies with no boundary prefix; SKIPPED when `truncation_boundary` / `truncation_watermark` set or `msg_before` paging; empty-string markers treated as unset (matching the existing helper). Existing backstop + reconciliation suites (53 tests) unchanged.

Note: the backstop is now narrower than my original PR claimed — it only bounds the uncompressed initial-tail path. The compressed-session case (which my PR said it "covered") is left uncapped, correctly, because bounding it was the bug. Worth flagging in the release note.

nesquena-hermes (approved):
Approving: the re-push implements the fix-spec — `_state_db_backstop_limit_for_display` now gates the 50k row cap to provably boundary-FREE reads only (no `msg_before`, no `truncation_watermark`/`truncation_boundary`), so a >50k-row session that needs the truncation-boundary prefix for `merge_session_messages_append_only` gets the uncapped read and the merge prefix is preserved. Re-gate on the current head: Codex found no defect in the backstop change, full suite green (lone failure is the unrelated editable-install artifact), CI green. Shipping.

nesquena-hermes:
🚀 Shipped in **exp-v0.52.91** (experimental). Thanks @rh-id — `_state_db_backstop_limit_for_display` now gates the 50k row cap to boundary-free reads only, so a >50k-row session that needs its truncation-boundary prefix for the append-only merge gets the uncapped read (no silent prefix loss). Re-gate clean (Codex no-findings, full suite green, CI 18/18). Live now.

PR #6154, opened 2026-07-16
fix(frontend): route msg_limit-heavy paths around the server ceiling (#6152)

rh-id:
## Thinking Path

- PR #6152 clamps `GET /api/session ?msg_limit=` to a max of 500 server-side — a correct memory fix, but I shipped it without tracing the frontend paths that relied on unbounded `msg_limit`.
- Review of the frontend revealed two paths that **silently break for sessions > 500 visible rows**: the outline "jump to message" feature and endless scroll / "Load earlier". This PR is the frontend follow-up that makes the clamp safe.

## What Changed

**1. `static/outline.js` — outline jump-to-message (was silently no-op for long sessions)**

`_jumpToMessage` sent `msg_limit=9999` as a "give me everything" hack so it could scroll to any message by absolute index (`msg-user-<rawIdx>`). Clamped to 500, the target row's DOM id doesn't exist for early messages in long sessions, and line 119's `if (r)` silently does nothing — the outline panel's core function failed invisibly.

Fix: drop `msg_limit=9999`, use the bare `messages=1` path. PR #6152 deliberately left the bare path unbounded for exactly this kind of caller; this restores the pre-#6152 behavior.

**2. `static/sessions.js` — endless scroll / "Load earlier" (was stalling at 500 rows)**

`_loadOlderMessages` grows `msg_limit` by `+_INITIAL_MSG_LIMIT` each load (30 → 60 → … → 500 → clamped). Past 500 the server returns the same clamped tail, `olderMsgs` shrinks to 0, and the load-more loop silently gives up — the head becomes unreachable for long sessions.

Fix: added `_MSG_LIMIT_MAX = 500` (mirroring backend `_MAX_MSG_LIMIT`, documented as kept-in-sync). Once the grown `requestedLimit` meets/exceeds it, the primary fetch switches to `msg_before` paging — a fixed `_INITIAL_MSG_LIMIT` backward page keyed off `_oldestIdx`. This is the **same bounded paging request the existing race-fallback uses** (proven correct there), so the head stays reachable for arbitrarily long transcripts. The response is handled by the existing prepend path (`tailMatches` is naturally false for a page, so the `!tailMatches` branch reuses the already-fetched `data` — no double fetch).

## Why It Matters

Without this, #6152's memory fix silently degrades two features for exactly the long sessions that motivated the memory work in the first place. With it, the clamp is safe: the only callers that legitimately need more than 500 rows now use either the unbounded bare path (outline) or proper bounded paging (load-older).

## Verification

- `node --check` passes on both edited files.
- Backend window-contract tests still pass: `test_session_tail_payload.py` (9) + `test_session_message_window_renderable_tail.py` (10).
- **No JS test harness exists in this repo** (vanilla JS, manually verified per `TESTING.md`). I could not run the browser here. The manual verification a maintainer should do:
  1. Open a session with > 500 messages.
  2. Click an outline entry pointing at an early message → confirm it scrolls to and flashes that message (was silently no-op).
  3. Scroll to the top → confirm "Load earlier" / endless scroll keeps producing older messages past the 500-row mark without stalling (was stalling).

## Risks / Follow-ups

- The `_MSG_LIMIT_MAX` JS constant must stay in sync with the backend `_MAX_MSG_LIMIT` (500). They're documented as a pair; if the backend ceiling changes, this mirror must too. (A future improvement could expose the ceiling via the `/api/session` metadata response so the frontend doesn't hardcode it — out of scope here.)
- The outline bare-path fetch loads the full transcript into `S.messages` for the jump — that's the same cost as before #6152 (the `msg_limit=9999` was effectively doing the same), so no regression, just worth noting it's the one remaining "load everything" path and is inherent to absolute-index addressing.

## Contract Routing

- Task type: frontend follow-up to a backend memory fix (#6152).
- Touched areas: outline jump-to-message (`static/outline.js`), endless-scroll / load-older (`static/sessions.js` `_loadOlderMessages`).
- Relevant public docs: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`.
- State layer mutated: client-side `S.messages` (the active-session transcript), same as the existing load-older path. No backend state change.
- Scope boundaries: does not weaken the #6152 clamp. The bare `messages=1` path (outline) was intentionally left unbounded by #6152; the `msg_before` paging path (load-older) is bounded by `_INITIAL_MSG_LIMIT` and never hits the ceiling.

Release note: fixes outline jump-to-message and endless scroll silently breaking on sessions longer than 500 messages after the server-side `msg_limit` ceiling (companion to #6152).

## Model Used

### DIFF
--- a/static/outline.js
+++ b/static/outline.js
@@ -104,10 +104,16 @@ function _jumpToMessage(rawIdx) {
   }
 
   // Row is outside the render window — reload the full session and retry.
+  // Use the bare messages=1 path (no msg_limit) so the server returns the
+  // COMPLETE transcript: the target row is addressed by absolute index
+  // (msg-user-<rawIdx>), so a bounded tail window would miss early rows.
+  // (A previous version sent msg_limit=9999 as a "give me everything" hack,
+  // but the server now clamps msg_limit, so the bare path is the correct way
+  // to request the full transcript here.)
   if (typeof api !== 'function') return;
   if (S.busy || S.activeStreamId) return;
   api('/api/session?session_id=' + encodeURIComponent(sid) +
-      '&messages=1&resolve_model=0&msg_limit=9999')
+      '&messages=1&resolve_model=0')
     .then(function(data) {
       if (!data || !data.session) return;
       if (!S.session || S.session.session_id !== sid) return;  // session switched
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -2928,6 +2928,19 @@ let _messagesTruncated = false;
 // server-bounded and do not consume the visible-message budget.
 // Older messages are loaded on-demand via _loadOlderMessages().
 const _INITIAL_MSG_LIMIT = 30;
+// ============================================================================
+// COUPLED CONSTANT — keep in sync with api/routes.py:_MAX_MSG_LIMIT.
+// ============================================================================
+// This is a hand-mirrored copy of the backend's GET /api/session ?msg_limit=
+// ceiling. _loadOlderMessages grows its msg_limit tail window by
+// +_INITIAL_MSG_LIMIT each load; once growth would exceed this ceiling the
+// server clamps and the tail stops growing, so we switch to msg_before paging
+// (a fixed-size backward page keyed off _oldestIdx) instead. If you change the
+// backend _MAX_MSG_LIMIT, change this value to match — otherwise load-older
+// silently stalls again for long sessions. The durable fix (so this mirror
+// isn't needed) is to expose the ceiling in the /api/session metadata response
+// and read it here; see tracking issue for that follow-up.
+const _MSG_LIMIT_MAX = 500;
 let _sameSessionForceReloadHint = null;
 
 function _currentLoadedRenderableMessageCount(){
@@ -3531,18 +3544,34 @@ async function _loadOlderMessages() {
   // rebuilt transcript (#1937).
   const startGeneration = _messagesGeneration;
   try {
-    // Ask the server for a larger authoritative tail window instead of a
-    // separate msg_before page. The same /api/session contract handles both —
-    // post-#2716 the backend always runs the full append-only merge, so a
-    // larger msg_limit on the same call produces the same merged transcript
-    // we'd get by stitching pages, but without client-side index bookkeeping.
-    // Cumulative growth: each "load more" asks for currentLoaded + 30, and the
-    // newly exposed head is what we expose to the user.
+    // Two strategies, chosen by whether the growing tail window still fits under
+    // the server's msg_limit ceiling (_MSG_LIMIT_MAX, mirroring backend
+    // _MAX_MSG_LIMIT):
+    //
+    //  - Below the ceiling: ask for a larger authoritative tail window
+    //    (currentLoaded + _INITIAL_MSG_LIMIT). Post-#2716 the backend runs the
+    //    full append-only merge, so a larger msg_limit produces the same merged
+    //    transcript we'd get by stitching pages, without client-side index
+    //    bookkeeping. The newly exposed head is what we expose to the user.
+    //
+    //  - At/above the ceiling: the server clamps msg_limit, so the tail window
+    //    stops growing and this strategy would stall (the same clamped tail is
+    //    returned, olderMsgs -> 0). Switch to msg_before paging — a fixed
+    //    _INITIAL_MSG_LIMIT backward page keyed off _oldestIdx — which is
+    //    bounded and never hits the ceiling, so the head stays reachable for
+    //    arbitrarily long transcripts. (This is the same paging request the
+    //    race-fallback below uses, proven correct there.)
     const requestedLimit = Math.max(_INITIAL_MSG_LIMIT, (S.messages || []).length + _INITIAL_MSG_LIMIT);
-    const data = await api(
-      `/api/session?session_id=${encodeURIComponent(sid)}&messages=1&resolve_model=0&msg_limit=${requestedLimit}`,
-      {timeoutMs:120000}
-    );
+    const useBeforePaging = requestedLimit >= _MSG_LIMIT_MAX;
+    const data = useBeforePaging
+      ? await api(
+          `/api/session?session_id=${encodeURIComponent(sid)}&messages=1&resolve_model=0&msg_before=${_oldestIdx}&msg_limit=${_INITIAL_MSG_LIMIT}`,
+          {timeoutMs:120000}
+        )
+      : await api(
+          `/api/session?session_id=${encodeURIComponent(sid)}&messages=1&resolve_model=0&msg_limit=${requestedLimit}`,
+          {timeoutMs:120000}
+        );
     // Guard: api() may have redirected (401) and returned undefined.
     if (!data || !data.session) { _loadingOlder = false; return; }
     //  - response shape sane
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
## Summary

I read the diff against `origin/master` plus the full `_loadOlderMessages` body at this PR's HEAD, and the fix is correct. #6152's server-side `msg_limit` clamp genuinely breaks the two paths this PR names, and the chosen remedy — switch load-older to `msg_before` paging once the growing tail window would cross the ceiling — is the right shape because it reuses a code path already proven correct in the race-fallback branch.

## Code reference

The pivot is clean (`static/sessions.js:3540`):

```js
const requestedLimit = Math.max(_INITIAL_MSG_LIMIT, (S.messages || []).length + _INITIAL_MSG_LIMIT);
const useBeforePaging = requestedLimit >= _MSG_LIMIT_MAX;
const data = useBeforePaging
  ? await api(`/api/session?...&msg_before=${_oldestIdx}&msg_limit=${_INITIAL_MSG_LIMIT}`, ...)
  : await api(`/api/session?...&msg_limit=${requestedLimit}`, ...);
```

The reason this stays contiguous across the transition is `_oldestIdx`, which is refreshed from the server's authoritative offset on every load (`static/sessions.js:3665`):

```js
_oldestIdx = responseSession._messages_offset || 0;
```

So when the tail-growth strategy tops out at 500 and hands off to `msg_before=${_oldestIdx}`, the backward page starts exactly at the current head — no gap, no overlap.

## What I verified

1. **The reuse branch is sound.** For a `msg_before` page, `expandedMsgs.length` (30) is `< currentLen`, so `tailMatches` is naturally false and control enters the `!tailMatches` block. The `if (!useBeforePaging)` guard (`static/sessions.js:3610`) correctly skips the re-fetch and reuses the already-fetched `data` as `responseSession` — no double request. Good call threading it through the existing prepend path rather than adding a parallel one.

2. **Outline jump is the right fix too.** Dropping `msg_limit=9999` for the bare `messages=1` path (`static/outline.js:116`) matches #6152's intent — the body of #6152 explicitly left the no-`msg_limit` path unbounded as the "give me everything" escape hatch. Absolute-index addressing (`msg-user-<rawIdx>`) needs the full transcript, so the bare path is correct; a clamped tail would silently no-op `if (r)` at line 119 for early rows, exactly as described.

## One follow-up worth noting

The `_MSG_LIMIT_MAX = 500` constant (`static/sessions.js:2932`) is a hand-synced mirror of the backend `_MAX_MSG_LIMIT`. You flag this yourself, and it's the one fragility here: if the backend ceiling ever changes and this isn't updated, load-older silently stalls again at the old value. The cleanest durable fix is to surface the ceiling in the `/api/session` response metadata (alongside `_messages_offset` / `_messages_truncated`) so the frontend reads it instead of hardcoding — out of scope for this PR, but worth a tracking issue so the coupling doesn't rot. Until then, a code comment at *both* constants pointing at each other (the JS side has it; adding the reciprocal note at the `_MAX_MSG_LIMIT` definition would close the loop).

## Test plan

The manual steps in the PR body are the right ones (open a >500-message session; outline-jump to an early message; scroll up past the 500-row mark). Since there's no JS harness here, one automatable guard would be a small assertion that `_MSG_LIMIT_MAX === _MAX_MSG_LIMIT` via a shared source, but that depends on the metadata-exposure follow-up above.

Net: the logic is correct and the reuse-vs-refetch handling is careful. The only durable risk is the hardcoded ceiling mirror.

rh-id:
Thanks both for the reviews — and for independently flagging the same coupling risk. The logic being correct, I focused on the one durable concern: the hand-mirrored `_MSG_LIMIT_MAX`.

## What I did here

Strengthened the `_MSG_LIMIT_MAX` constant comment (`static/sessions.js`) to make the coupling loud rather than easy to miss — a banner-style note pointing at `api/routes.py:_MAX_MSG_LIMIT`, the failure mode if they drift (load-older silently stalls), and a pointer to the durable fix. Pushed in `3e1a367`.

I considered the reciprocal comment at the backend `_MAX_MSG_LIMIT` that was suggested, but that constant lives on the #6152 branch (not master), so editing it belongs in #6152, not here — I'll add it there.

## The durable fix — filed as #6177

Per the maintainer's suggestion: surface the ceiling in the `/api/session` response metadata (`_msg_limit_max`, alongside `_messages_offset` / `_messages_truncated`) and have the frontend read it instead of hardcoding. That retires the mirror entirely. Filed as a separate tracking issue (#6177) since it spans both PRs and changes a response contract field — it shouldn't be folded into either.

#6177 also notes a cheap stopgap available until the metadata exposure lands: a static-source assertion test (the repo already has `tests/js_source_extract.py`) that reads both constants and fails if they differ. Happy to add that stopgap test to either this PR or #6152 if you'd prefer a guard before the metadata field ships — just say which.

## Net

- Logic: approved by both reviews, unchanged.
- Coupling risk: comment made loud here (#6154) + backend reciprocal note to be added in #6152 + durable metadata-exposure tracked in #6177.

No further code changes needed in this PR unless you want the stopgap drift-test added here.

nesquena-hermes:
## Round-2 confirmation + answering the stopgap placement question

Re-pulled at `94e195060`. Two things to confirm and one direct answer to your "which PR gets the stopgap test" question.

## The test-shape fix (94e195060) is correct

`test_session_message_loads_keep_explicit_longer_timeouts` guards that both load-older fetches keep `timeoutMs:120000` for large `state.db` loads. The `useBeforePaging` ternary reindented the fetch strings, so the old exact-multiline assertion no longer matched the source. I checked the new assertions against the actual source at `static/sessions.js:3566-3573`:

```js
const data = useBeforePaging
  ? await api(
      `/api/session?session_id=${encodeURIComponent(sid)}&messages=1&resolve_model=0&msg_before=${_oldestIdx}&msg_limit=${_INITIAL_MSG_LIMIT}`,
      {timeoutMs:120000}
    )
  : await api(
      `/api/session?session_id=${encodeURIComponent(sid)}&messages=1&resolve_model=0&msg_limit=${requestedLimit}`,
      {timeoutMs:120000}
    );
```

Both assertion substrings (each URL + its `{timeoutMs:120000}` at the new 10-space indent) are present, and `grep -c` finds both fetch strings. The contract the test actually cares about — neither strategy silently drops the long timeout — is preserved, so relaxing the assertion to per-URL + timeout rather than exact block shape is the right call. It's more refactor-durable too.

## The coupling comment (3e1a367) reads well

The banner note at `static/sessions.js:2932-2942` names the backend symbol, the drift failure mode (load-older silently stalls), and points at the durable fix. That's the "make it loud" outcome. Filing #6177 for the metadata-exposure fix is the correct scope split since it changes a response contract field across both PRs.

## Your question: stopgap drift-test here or in #6152?

Put it in **#6152**, not here — and it's not just a style preference, it's a correctness constraint. I checked, and `_MAX_MSG_LIMIT` does not exist on `origin/master`:

```
$ git show origin/master:api/routes.py | grep -n _MAX_MSG_LIMIT
(no matches)
$ git show <pr6152-head>:api/routes.py | grep -n _MAX_MSG_LIMIT
8452:_MAX_MSG_LIMIT = 500
```

The backend constant only lives on the #6152 branch (still OPEN / CONFLICTING). A drift-test that asserts `_MSG_LIMIT_MAX === _MAX_MSG_LIMIT` needs *both* symbols to exist in the tree it runs against. If you add it to #6154 now, the test would have to reach for a backend constant that isn't on master yet — it'd either fail on this branch's CI or force you to hardcode the backend value (which just moves the mirror into the test). Adding it to #6152 means the guard ships atomically with the constant it mirrors, and #6154's own `_MSG_LIMIT_MAX` becomes the thing it validates against once both land.

`tests/js_source_extract.py` (present on master) is the right harness for it — it already extracts JS const values for exactly this kind of source assertion, so the stopgap is cheap to write in #6152.

## Net

Logic unchanged and already approved; the test-shape fix is correct; the coupling is now loud and tracked in #6177. No further changes needed in *this* PR. The stopgap drift-test belongs in #6152 because that's where `_MAX_MSG_LIMIT` is defined — this PR can't cross-check a symbol that isn't on master.

rh-id:
Thanks for the precise placement guidance — done exactly as you specified: the stopgap drift-test is in **#6152** (commit `8a5c275`), shipping atomically with `_MAX_MSG_LIMIT`.

`tests/test_msg_limit_ceiling_drift.py`:
- Extracts `_MAX_MSG_LIMIT` from `api/routes.py` and `_MSG_LIMIT_MAX` from `static/sessions.js` via regex.
- Asserts equality **when both are present**.
- **SKIPS** when the JS constant is absent — so #6152's CI stays green on its own (the `_MSG_LIMIT_MAX` mirror only exists once #6154 lands), and the guard activates automatically once both branches are on master.

Verified the skip behaves correctly on #6152 (cross-check SKIPPED, backend-existence PASSED, CI 0 failing), and confirmed the extraction logic bites on a simulated drift (999 != 500 fails). Retired by #6177 once the metadata-exposure lands.

No changes needed in *this* PR — closing the loop on the open follow-up.

nesquena-hermes:
Shipped in **exp-v0.52.98** (combined release with #6152). Thanks @rh-id! 🎉

Your frontend paging rework (grow-tail-below-ceiling → `msg_before` backward paging above it, plus the outline-jump bare-path fix) landed. The pre-ship adversarial gate reproduced two silent row-loss edges that were fixed in a maintainer follow-up commit (co-authored to you):
1. A raw-row-heavy `msg_before` page whose visible text repeats the current tail could be misclassified as the cumulative tail and wholesale-replace it (470 of 940 rows lost) → now gated on `!useBeforePaging` so cursor pages are always prepended.
2. A same-session refresh above the ceiling could shrink an already-loaded >500-row transcript to the last 500 → over-ceiling reloads now use the bare full-transcript path.

Re-gate confirmed SAFE (940/940 + 1000/1000 rows retained, no seam duplication). Solid work.

PR #6157, opened 2026-07-16
fix: sanitize mixed Markdown table copies

### LINKED ISSUE #6063: Mixed-content copy bypasses table sanitizer, preserves dark theme, and blanks headers

heylaurens:
Closes #6063.

## Thinking Path

The existing copy handler only takes ownership of an isolated full enhanced table. A mixed range therefore falls through to native clipboard serialization, which preserves presentation attributes and can omit enhanced header labels.

The mixed-range path now clones the selected fragment, sanitizes surrounding content into a small semantic allowlist, and replaces each enhanced table with the existing table payload builder. It only takes ownership when every intersected enhanced table is present in full in the clone; partial selections continue to use native browser copy behavior.

## What Changed

- Preserve the existing isolated full-table fast path.
- Add theme-neutral semantic HTML and readable plain-text serialization for mixed heading/prose/table selections.
- Preserve safe inline semantics such as emphasis and safe links while dropping styles, classes, event attributes, controls, unsafe URL schemes, forms, scripts, and media/embed content.
- Compare each cloned table payload with its source table before intercepting the copy event, preventing partial-table truncation.
- Add a Node DOM regression for a heading, two enhanced tables, bold subtotal content, surrounding prose, safe and unsafe links, and hidden form content.
- Add manual test T23.3 for rich-text and plain-text paste behavior.

## Why It Matters

Copying a complete rendered response should not leak dark-theme presentation into another editor or produce blank table headers. The new branch keeps the useful structure and text while maintaining the existing conservative fallback for incomplete selections.

Release note: Mixed-content copies containing complete Markdown tables now produce theme-neutral rich text and readable plain text without blanking table headers.

## Verification

- Red proof on pre-fix `master`: 9 existing clipboard tests passed and the new mixed-range regression failed because the event was not intercepted.
- `./scripts/test.sh tests/test_issue4945_markdown_table_copy.py -q`: 11 passed.
- `./scripts/test.sh`: exit 0 in 707 seconds; 13,138 tests collected, with 5 collection skips and the documented 30 agent-dependent skips because `hermes-agent` is unavailable locally.
- `npm run lint:runtime`: passed.
- `node --check static/messages.js`: passed.
- `git diff --check origin/master...HEAD`: passed.
- Gitleaks v8.30.1 over `origin/master..HEAD`: no leaks found.

## Risks / Follow-ups

- Clipboard HTML intentionally uses an allowlist, so non-semantic presentation markup is flattened rather than copied.
- Only ranges containing complete enhanced tables are intercepted; incomplete tables still fall through to native copy behavior.
- No rendered UI, layout, or interaction control changed, so before/after screenshots are not applicable. The regression asserts the exact HTML and plain-text clipboard payloads instead.

## Contract Routing

- Task type: clipboard/UI behavior bug.
- Touched surfaces: `static/messages.js`, Node DOM regression coverage, and `TESTING.md`.
- Consulted contracts: `AGENTS.md`, `CONTRIBUTING.md`, `docs/CONTRACTS.md`, `docs/GUIDELINES.md`, UI/UX guidance, design guidance, and `TESTING.md`.
- Scope boundary: clipboard serialization only; no rendering, persistence, API, or backend behavior changes.

## Model Used

- Provider/model: OpenAI Codex (GPT-5).
- Tools: repository search and review, PowerShell/Git, the project test runner, Node.js syntax and ESLint checks, GitHub CLI, and Gitleaks.
- The generated changes were manually reviewed, security-checked, and tested before publication.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1178,6 +1178,20 @@ EXPECT:
   - Paste (Cmd+V) elsewhere shows the full text of that message
 FAIL: No visual feedback, clipboard empty or wrong content.
 
+### T23.3: Copy Mixed Rendered Content with Tables
+SETUP: Use the dark theme and render an assistant response with a heading, prose,
+two Markdown tables, and a bold subtotal cell.
+STEPS:
+  1. Select the complete rendered response, including both tables
+  2. Copy it and paste into a rich-text editor
+  3. Paste it again as plain text
+EXPECT:
+  - Rich text keeps headings, table headers, values, and bold subtotal text
+  - Rich text does not carry Hermes theme colors, spacing, or table controls
+  - Plain text keeps surrounding prose and tab-delimited table rows readable
+FAIL: Theme styling leaks, header cells are blank, formatting disappears, or
+partial table content is missing.
+
 ---
 
 ## Section 24: Inline File Editor (Sprint 5)
--- a/static/messages.js
+++ b/static/messages.js
@@ -482,6 +482,82 @@ function _markdownTableCopyHtmlEscape(value){
     .replace(/>/g,'&gt;');
 }
 
+function _markdownTableCopyHtmlAttributeEscape(value){
+  return _markdownTableCopyHtmlEscape(value)
+    .split('"').join('&quot;')
+    .split("'").join('&#39;');
+}
+
+function _markdownTableCopySafeHref(value){
+  const href=String(value||'').trim();
+  if(!href) return '';
+  if(/^(?:https?:|mailto:|tel:|#|\/(?!\/)|\.\.\/|\.\/)/i.test(href)) return href;
+  return '';
+}
+
+function _markdownTableCopySemanticChildren(node){
+  return _markdownTableNodeChildren(node)
+    .map(_markdownTableCopySemanticHtml)
+    .join('');
+}
+
+function _markdownTableCopySemanticHtml(node){
+  if(!node) return '';
+  if(node.nodeType===3) return _markdownTableCopyHtmlEscape(node.textContent||'');
+  const children=_markdownTableCopySemanticChildren(node);
+  if(node.nodeType!==1||!node.tagName) return children;
+
+  const tag=String(node.tagName).toLowerCase();
+  if(tag==='table'&&node.matches&&node.matches('table[data-markdown-table-enhanced]')){
+    const payload=_markdownTableCopyPayloadForTable(node);
+    return payload?payload.html:'';
+  }
+
+  const blockedTags=[
+    'script','style','button','input','select','textarea','option','svg','canvas',
+    'iframe','object','embed','form','video','audio',
+  ];
+  if(blockedTags.includes(tag)) return '';
+
+  const semanticTags=[
+    'a','abbr','b','blockquote','br','code','dd','del','details','div','dl','dt',
+    'em','h1','h2','h3','h4','h5','h6','hr','i','kbd','li','mark','ol','p',
+    'pre','q','s','samp','small','span','strong','sub','summary','sup','table',
+    'tbody','td','tfoot','th','thead','tr','u','ul','var',
+  ];
+  if(!semanticTags.includes(tag)) return children;
+
+  const getAttribute=(name)=>typeof node.getAttribute==='function'?node.getAttribute(name):null;
+  const attributes=[];
+  if(tag==='a'){
+    const href=_markdownTableCopySafeHref(getAttribute('href'));
+    if(href) attributes.push(`href="${_markdownTableCopyHtmlAttributeEscape(href)}"`);
+  }
+  if(tag==='ol'){
+    const start=getAttribute('start');
+    if(start!==null&&/^-?\d+$/.test(String(start))) attributes.push(`start="${String(start)}"`);
+  }
+  if(tag==='td'||tag==='th'){
+    for(const name of ['colspan','rowspan']){
+      const value=getAttribute(name);
+      if(value!==null&&/^\d+$/.test(String(value))) attributes.push(`${name}="${String(value)}"`);
+    }
+  }
+
+  const opening=`<${tag}${attributes.length?' '+attributes.join(' '):''}>`;
+  if(tag==='br'||tag==='hr') return opening;
+  return `${opening}${children}</${tag}>`;
+}
+
+function _markdownTableCopyCellHtml(cell){
+  if(!cell) return '';
+  const sortButton=cell.querySelector?cell.querySelector('.markdown-table-sort'):null;
+  const sortLabel=sortButton&&sortButton.querySelector
+    ? sortButton.querySelector('.markdown-table-sort-label')
+    : null;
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (security sanitizer is sound; mixed-copy silently loses ordered-list numbering)

**Certified head:** `bd8acd06aead299264ff1e79e871d2b3e4c79042` (clean-rebased onto current `origin/master`)
**Verdict:** The new clipboard HTML sanitizer is conservative, fail-closed, and security-sound. However, the newly intercepted mixed-selection path silently changes ordered lists in both clipboard formats; the plain-text serializer also collapses generic non-enhanced table cells. Ship only with the small fidelity fixes below.

### What I ran
- **Codex adversarial gate:** `SHIP ONLY WITH FIXES` — reproduced the ordered-list loss against the real production helpers.
- **Senior advisor:** `COMMENT` — no security blocker; independently reproduced the ordered-list and generic-table plain-text regressions.
- **My adversarial Node probe:** verified `javascript:`, mixed-case/leading-space/tab-obfuscated variants, `data:`, `vbscript:`, and protocol-relative `//` hrefs are rejected; `https:`/`mailto:`/`tel:`/fragment/relative allowlist works; double/single quotes, angle brackets, and ampersands are correctly escaped.
- **Targeted:** `11 passed` in `tests/test_issue4945_markdown_table_copy.py`; ESLint runtime, scope-undef, and `node --check` clean.
- **Full serial suite:** `13438 passed / 6 failed`; all six are the established no-network Nous-catalog/picker/config failures reproduced on clean master, none from this diff.

### Verified security boundary — CLEAN ✅
- Semantic-tag allowlist; dangerous tags and their contents dropped.
- Attribute allowlist only (`href`, `ol[start]`, `td/th[colspan|rowspan]`); all values validated and escaped.
- Unknown tags serialize only their safe children, never raw markup.
- Clone/source table count and payload equality must match; every uncertain branch returns `null`, preserving native copy instead of writing a partial or unsanitized payload.
- Full-table fast path and fallback sort-label/header behavior are correct.

### Reproduced blocker
**[SILENT] `static/messages.js:536,721` — ordered-list numbering is lost in mixed copies.** `renderMd()` emits exact numbering through numeric `<li value="…">` (`static/ui.js:7373`), including non-default/non-sequential values. The sanitizer strips `li[value]` from rich HTML and `_markdownTableCopyPlainText()` serializes every `<li>` as `- `. Codex's real-function probe: a list with values `5` and `8` becomes ordinary sequential items in `text/html` and bullets in `text/plain`. This is a silent semantic regression relative to native copy on the path this PR newly intercepts.

**Fix:**
1. Allowlist numeric `value` on `<li>` (same validation discipline as `ol[start]`).
2. In `_markdownTableCopyPlainText`, emit ordered markers derived from `li[value]` / parent `ol[start]` rather than always `- `.
3. Add a mixed list-plus-enhanced-table regression with non-default/non-sequential values (for example 5 then 8), asserting both HTML and plain output preserve them.

### Secondary fidelity fix
Opus also reproduced generic non-enhanced tables inside an intercepted mixed range serializing as `A1B1A2B2` because `tr`/`td`/`th` have no plain-text separators. Add tabs between generic cells and newlines between rows, or explicitly defer this in the PR thread if native-copy parity is not part of this slice.

### Recommendation
**Do not merge this head.** Keep `gate-fail` at `sha:bd8acd06aead`. The core security design is strong and does not need a rewrite; preserve it, add the ordered-list fix/test (and preferably generic-table separators), then re-gate. This is small mechanical contributor-owned work, so I am not adding `changes-requested`. Preserve @heylaurens as author.

@heylaurens — the sanitizer itself held up well under adversarial testing. The release blocker is fidelity: mixed copies strip exact ordered-list numbering (`li[value]`) and turn every plain-text item into a bullet. Please preserve numeric `li[value]`, serialize ordered markers from `li[value]`/`ol[start]`, and add a non-default/non-sequential list-plus-table regression. The generic-table plain-text separator fix is also recommended.

PR #6169, opened 2026-07-16
feat(workspace): render local images referenced in .md files when browsing workspace

### LINKED ISSUE #5933: feat(workspace): render local images referenced in .md files when browsing workspace

webtecnica:
## Summary

Closes #5933 — when browsing Markdown files in the Workspace panel, images referenced via `alt` or `alt` now render inline instead of showing as broken images.

## Problem

The `renderMd()` function in `static/ui.js` only matches `https?://` URLs for markdown image syntax:
```js
/!\[([^\]]*)\]\((https?:\/\/[^\)]+)\)/g
```

Local relative paths like `image.png`, `./screenshots/result.jpg`, or `../assets/diagram.png` never matched this regex, so they were never converted to `` tags — leaving broken images.

## Solution

Added `_resolveWorkspaceImagePaths()` in `static/workspace.js` that preprocesses markdown content before it reaches `renderMd()`:

1. Detects `alt` references where the path is **not** an absolute URL (`http://`, `https://`, or `/absolute-path`)
2. Resolves the relative path against the markdown file's directory (from `_previewCurrentPath`)
3. Normalizes the resolved path and builds a workspace API raw-file URL via the existing `_workspaceRouteForPath(path, 'raw')` helper
4. The rewritten `alt` is then processed normally by `renderMd()` into `` tags

### Examples

| Markdown | Before | After |
|---|---|---|
| `screenshot` | Broken image | Renders via `/api/file/raw?session_id=...&path=dir/screenshot.png` |
| `` | Broken image | Renders via `/api/file/raw?session_id=...&path=dir/image.png` |
| `alt` | Broken image | Renders via `/api/file/raw?session_id=...&path=assets/diagram.jpg` |
| `alt` | Already works | Unchanged (not local) |

## Files Changed

- **`static/workspace.js`** — Added `_resolveWorkspaceImagePaths()` helper and modified `renderMarkdownPreviewContent()` to preprocess content before calling `renderMd()`

## Verification

- [x] Relative paths like `./image.png` resolve against the markdown file's directory
- [x] Absolute HTTP/HTTPS URLs are left untouched
- [x] Absolute paths starting with `/` are left untouched
- [x] Paths with `../` parent-directory references are normalized
- [x] Handles escape-grant workspaces via `_workspaceRouteForPath()` (which includes token auth)
- [x] Empty paths and invalid normalized paths are skipped silently

## Test evidence

Before: Local image reference shows as broken image in workspace markdown preview.
After: Same markdown renders with inline image served through the workspace API.

### DIFF
--- a/static/workspace.js
+++ b/static/workspace.js
@@ -848,10 +848,35 @@ function setLargeMarkdownForceRenderVisible(visible){
   if(btn) btn.style.display=visible?'inline-flex':'none';
 }
 
+function _resolveWorkspaceImagePaths(content){
+  // Rewrite local image references (e.g. ![alt](./image.png)) to workspace
+  // API URLs so they render as <img> tags in the markdown preview.
+  // Only runs when we have a current file path and active session.
+  if(!_previewCurrentPath || !S.session) return content;
+  const baseDir = _previewCurrentPath.includes('/')
+    ? _previewCurrentPath.substring(0, _previewCurrentPath.lastIndexOf('/'))
+    : '.';
+  return content.replace(/!\[([^\]]*)\]\(((?!https?:\/\/)[^\)]+)\)/g, (_, alt, relPath) => {
+    // Skip absolute URLs (http://, https://, /absolute-path)
+    if(/^https?:\/\//i.test(relPath) || /^\//.test(relPath)) return _;
+    // Resolve relative path against the markdown file directory
+    const stripped = relPath.replace(/^\.\//, '');
+    const resolvedPath = baseDir === '.' ? stripped : baseDir + '/' + stripped;
+    const normalized = _normalizeWorkspaceRelPath(resolvedPath);
+    if(!normalized) return _;
+    // Build the workspace API URL for the resolved image path
+    const url = _workspaceRouteForPath(normalized, 'raw');
+    if(!url) return _;
+    return `![${alt}](${url})`;
+  });
+}
+
 function renderMarkdownPreviewContent(data){
   const target=data&&data.el?data.el:$('previewMd');
   if(!data||!data.el) showPreview('md');
-  target.innerHTML=renderMd(data.content);
+  // Resolve local image paths before rendering markdown
+  const resolvedContent = _resolveWorkspaceImagePaths(data.content);
+  target.innerHTML=renderMd(resolvedContent);
   requestAnimationFrame(()=>{if(typeof renderKatexBlocks==='function')renderKatexBlocks();});
 }
 


nesquena-hermes:

**Exact contributor head:** `2ddba54570df4fcceb28f561f9b37dab9805d93c`

I gated this head after a clean transplant onto current `origin/master` (`deac1384f`) on `gate-rebase/6169-local-images` at `d8228c0ecd61cfeb4c5fde922525a780934084f5`.

### Evidence
- Threat scan: **CLEAN**. Node syntax and diff checks passed.
- Full sandboxed suite: **13,354 passed / 16 failed**. Fourteen are the clean-master sandbox baseline; two are PR-specific source-contract regressions: `test_issue2823_large_markdown_preview` and `test_issue347`.
- GitHub CI is likewise red across six shards on the stale `renderMd(data.content)` contract.
- Codex: **SHIP ONLY WITH FIXES**.
- Fable: **SHIP WITH UX FIXES**.
- I reproduced the failures in the real workspace preview, not only from static review.

### Release-blocking findings
1. The whole-source regex is not Markdown-aware. It rewrites image syntax inside inline code and fenced code, mutating literal documentation and even turning the inline-code sample into a live authenticated image request.
2. Valid CommonMark destinations break: an image title becomes part of the filename; angle-bracket destinations and query/hash suffixes are encoded into the path; filenames with nested/escaped parentheses are not parsed safely.
3. `renderMarkdownPreviewContent()` is shared with the wiki/generic-element path, but the resolver unconditionally reads workspace-global `_previewCurrentPath`. I reproduced a wiki/generic render being rewritten to the currently open workspace file’s directory.
4. The async file-open path has a stale-response race. In a controlled A-then-B open, delayed A overwrote B’s preview and resolved A’s image against B’s directory (`dirB/a.png`) while `_previewCurrentPath` remained B.
5. Workspace document images inherit the chat thumbnail rule (fixed 120×90 crop). Fable requires a preview-scoped responsive image rule so document images render contained at workspace-pane width.
6. The PR adds no executable regression coverage and leaves current CI red.

### Required fix shape
- Replace the raw whole-source regex with a Markdown-aware resolver that preserves code regions, escaped syntax, URI schemes, titles, angle destinations, parentheses, query/hash suffixes, and percent encoding.
- Pass an explicit captured path/session into workspace resolution, reject stale responses after the await, and leave generic/wiki callers unchanged.
- Add a `.preview-md .msg-media-img` responsive containment rule and desktop/mobile/lightbox evidence.
- Update the KaTeX ordering assertion for the new render input and add fail-first tests for all cases above.

Traversal, raw-route authorization, subpath routing, escape-grant checks, alt text, and the simple `alt` case were otherwise sound.

This exact head is technically **RED**. `gate-fail`, `changes-requested`, and `ux` are appropriate. The security/visual product dependencies remain downstream of these objective defects.

PR #6170, opened 2026-07-16
feat(search): unify session title and content search into single query (#5890)

### LINKED ISSUE #5890: Unify session search - one box for titles + message content
The current search bar only filters session titles from a local cache. It should also search message body content via the backend /api/sessions/search endpoint and display both title matches and content matches with highlighted snippets in the same dropdown.

webtecnica:
Unifies the separate title and content search into a single search box that searches both simultaneously.

Changes:
- Backend: session search now returns title + content matches in a unified result
- Frontend: single search box searches both titles and message content

Closes #5890


nesquena-hermes:

**Exact contributor head:** `04bcc8300862c605fabe25d88859fc0ac99d089b`

I gated the actual diff after a clean transplant onto current `origin/master` (`deac1384f`) on `gate-rebase/6170-workdir` at `db6b834ca5a69982aa04cb7a908c8b6690ae4595`. The PR title/body are mismatched: this is the #5932 workspace **Go to directory** selector, not unified session search / #5890. The #6172 duplicate/supersede call remains a separate product dependency.

### Release-blocking findings

1. **Absolute in-workspace paths split display state from mutation state.** In the live browser, entering `/tmp/hermes-test-ws-8799-pr6170/images` rendered the workspace-root entries while setting `S.currentDir` and the create target to that absolute `/images` path. A user can see one directory and create/upload/delete in another.
2. **Failed navigation corrupts directory state silently.** A real 404 left the old root tree visible but changed `S.currentDir` to `definitely/not/a/real/directory`; subsequent create/refresh targets inherited that nonexistent path. The selector also closes before success and shows no error.
3. **The new permanent header action clips an existing control.** At a supported 204px resized workspace panel with Parent and Close Preview visible, Chromium measured Close Preview as 24px visible before this button and **0px visible** with it.
4. **Selector state leaks across interactions.** `.` and `~` leave the bar open; switching Files → Artifacts → Files restores the inline `display:flex`; Escape does nothing and leaves focus trapped in the input.
5. **Accessibility/localization are incomplete.** `workspace_go_to_dir` and `workspace_go_to_dir_placeholder` do not exist in `static/i18n.js`; the visible `Go` label is also hard-coded. The new mobile controls measure 24px high rather than the 44px mobile target bar.

Codex independently reproduced findings 1–4. Fable independently required navigation error handling, consistent close/focus behavior, i18n, `aria-expanded`/`aria-controls`, Escape handling, and a less misleading folder-navigation glyph. My own real desktop and iPhone-14 browser passes reproduced the state, layout, focus, and mobile-target defects above.

### Verification

- Threat scan: CLEAN; low-trust code ran only in the isolated sandbox.
- Syntax: `node --check static/workspace.js` passed.
- Focused advisor gate: 18 workspace tests passed.
- Broad sandbox suite: `13,323 passed / 47 failed`; clean master under the same sandbox reproduced 14 environment/browser/path baselines. A focused rerun of the tail's streaming/runtime failures passed `159/159`, confirming substantial order-sensitive noise, but the suite was not green and does not override the direct browser reproductions.

### Required fix shape

Canonicalize a typed path once and use the same authorized relative/granted path for listing and every mutation; commit `S.currentDir` only after successful non-stale navigation; keep the bar open with an observable error on failure; move the action into the existing workspace overflow/menu so core controls remain reachable; add one close/focus helper for success, tab leave, and Escape; complete i18n/ARIA/mobile targets; then add DOM/browser tests for every reproduced transition.

**Gate result: RED.** This certificate is authoritative only for exact head `04bcc8300862c605fabe25d88859fc0ac99d089b`. The contributor needs to re-push before re-gating. Even after technical convergence, Nathan's visual, retitle, and #6172 dedup decisions remain open.

Stooovie:
**+1** — this is the most missing UX piece in the WebUI right now.

The `session_search` tool (FTS5) already works great from the agent side, but there's no way to search session content from the browser UI. The session list only shows titles and metadata — if you have 200+ sessions and vaguely remember "we talked about X somewhere," you're scrolling blind.

### What I'd expect from a unified search:
- **Single search box** at the top of the sessions sidebar (or Cmd+K palette) that hits both titles AND message content via the existing FTS5 backend
- **Keyboard-first** — focus the search field on Cmd+K / Ctrl+K, results filtered live as you type
- **Click a result → jump to the matching message** within the session (similar to how `session_search` returns `match_message_id`)
- **Highlight the matched term** in both the session list preview and the chat view

### Why it matters:
Right now the only way to find old conversations is either asking the agent (`session_search`), or manually scrolling through session titles. A browser-side search would make the WebUI a serious replacement for Claude's desktop app session management — which already has this.

The backend for this largely exists in `state.db` (SQLite + FTS5). The WebUI just needs a search endpoint that wraps it and a frontend component in the sidebar.

Happy to help test or provide feedback on the PR if needed.

PR #6173, opened 2026-07-16
feat: allow reply/continue in cron sessions (#5936)

### LINKED ISSUE #5936: Cron sessions are read-only in WebUI — cannot reply/continue conversation
## Problem

Cron job sessions (e.g. a daily briefing produced by `hermes cron`) appear in the WebUI sidebar under the "Cron jobs" section, but attempting to send a message or continue the conversation shows:

> **Read-only imported sessions cannot be modified.**

This means users can read the cron output but cannot ask follow-up questions about it in the same session context.

## Root cause

`_is_claimable_cli_source()` in `api/routes.py` (around line 7207) explicitly denies `"cron"` as a claimable source:

```python
if cli_meta_source_tag in {"claude_code", "cron", "external_agent",
                            "gateway", "messaging", "subagent", "unknown"}:
    return False, f"cli_meta_source={cli_meta_source_tag}"
```

And the state.db fallback at line 7267:

```python
if state_db_source_tag in {"claude_code", "cron", "messaging",
                            "external_agent", "gateway", "subagent", "unknown"}:
    return False, f"state_db_source={state_db_source_tag}"
```

This means any session with `source='cron'` is treated as read-only — the WebUI refuses to materialise a writable sidecar, so `prompt.submit` / session continue returns 403.

## Why this matters

Hermes Agent's cron system has an `attach_to_session` flag (set via the `cronjob` tool) specifically designed to make cron deliveries **continuable** — the user can reply to the brief and the agent picks up the context. This works on messaging platforms (Telegram, Discord, etc.) where `mirror_to_session` appends the brief as a user turn.

But in WebUI, this feature is completely blocked by the denylist above, even when `attach_to_session=true` is set on the job. The two mechanisms are independent:
- `attach_to_session` controls gateway delivery mirroring
- WebUI's `_is_claimable_cli_source` is a separate gate that blocks all cron sessions regardless

## Proposed behaviour

Allow cron sessions to be claimed as writable in WebUI, so users can reply and continue the conversation inline. This is particularly valuable for briefing/summary-style cron jobs where the user may want to ask follow-up questions about the content.

## Technical notes

- Removing `"cron"` from the denylist sets in `_is_claimable_cli_source` (lines ~7248 and ~7267) should suffice
- The concern in the existing code comments is that "claiming them into a writable WebUI sidecar would let a stray POST break the next scheduled run" — but each cron run creates a **new** session (e.g. `cron_40bdc5a67a12_20260711_090031`), so replying to a past run's session doesn't interfere with the next scheduled run
- `attach_to_session=true` on the cron job is a clear signal that the user wants the session to be continuable; WebUI could optionally respect this flag rather than blanket-denying all cron sources

webtecnica:
Allows users to reply and continue conversations in cron-triggered sessions.

Previously cron sessions were treated as read-only. This change removes 'cron' from the denylist so users can fork and continue cron session conversations.

Closes #5936

### DIFF
--- a/api/session_ops.py
+++ b/api/session_ops.py
@@ -8,6 +8,7 @@ the hermes-agent repo.
 from __future__ import annotations
 import json
 import logging
+from bisect import bisect_left
 from typing import Any
 
 from api.config import LOCK, _get_session_agent_lock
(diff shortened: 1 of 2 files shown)


nesquena-hermes (changes requested):
Thanks for the contribution @webtecnica. This can't merge as-is — it breaks 5 pre-existing tests it doesn't touch, which will red the CI, and it reverses a deliberate ownership-boundary decision that needs maintainer sign-off. Details below, reproduced locally.

## Blocker 1 (objective, CI-red): removing `"cron"` from the denylist breaks 5 existing tests

The change deletes `"cron"` from **both** denylist sets in the shared `_is_claimable_cli_source()` helper (`api/routes.py`):
- the `cli_meta_source_tag` branch (was `{"claude_code", "cron", "external_agent", "gateway", "messaging", "subagent", "unknown"}`)
- the `state_db_source_tag` branch (was `{"claude_code", "cron", "messaging", "external_agent", "gateway", "subagent", "unknown"}`)

But `tests/test_chat_start_claim_cli_session.py` (which this PR does **not** modify) contains tests that explicitly pin cron **in** the denylist / require cron sessions to be refused. With the denylist change, those tests fail. Reproduced locally on this PR's head against the current test file:

```
tests/test_chat_start_claim_cli_session.py::test_helper_refuses_cron_session FAILED
tests/test_chat_start_claim_cli_session.py::test_helper_refuses_cron_via_state_db_source FAILED
tests/test_chat_start_claim_cli_session.py::test_branch_from_cron_state_db_returns_writable_fork_without_source_sidecar FAILED
tests/test_chat_start_claim_cli_session.py::test_chat_start_still_refuses_cron_state_db_source FAILED
tests/test_chat_start_claim_cli_session.py::test_helper_denylist_includes_gateway_and_unknown FAILED
5 failed, 6 passed
```

`test_helper_denylist_includes_gateway_and_unknown` fails with:

```
AssertionError: _is_claimable_cli_source must denylist 'cron' in both the cli_meta and
state_db branches — cron was added in the Greptile 4/5 P2 review to keep scheduled runs
CLI-owned; gateway/unknown were added in the residual #4911 review gap
```

A PR that reddens tests in a file it didn't touch can't merge. Whatever the final design, those tests have to be updated in the same PR so CI is green.

## Blocker 2 (design, needs maintainer sign-off): this reverses a deliberate cron-ownership guard, and the relaxation isn't scoped to reply/continue

`cron` was placed in that denylist intentionally (#4911 follow-up, Greptile 4/5) so scheduled runs stay owned by the cron runner and a stray WebUI POST can't claim a cron session in-place and disturb the next scheduled run. Removing it is a policy reversal that a maintainer should approve before it lands.

It also isn't scoped to just reply/continue. `_is_claimable_cli_source()` is a **shared** gate used by GET `/api/session`, POST `/api/chat/start`, share resolution, and branch/fork (`_load_branch_source_or_refuse`). Dropping cron from the denylist makes cron sessions claimable / writable-in-place across **all** of those, not only the advertised inline-reply path:
- in `/api/chat/start`, the cron session id itself is claimed as a writable WebUI sidecar (`synth.save()` + `SESSIONS[sid] = s`) — claim-in-place, not a fork;
- in the branch route, the new `materialized` return path does not set `_branch_source_readonly`, so `source.save()` runs and claims the cron source writable (previously the cron path set `_branch_source_readonly=True` → fork-only, source stayed read-only).

## What a mergeable version needs

1. Update the 5 tests in `tests/test_chat_start_claim_cli_session.py` to match the new behavior (flip cron from "refused" to "claimable"; drop `"cron"` from the asserted-literals tuple in `test_helper_denylist_includes_gateway_and_unknown`, keeping `gateway`/`unknown`).
2. Add a positive test for the **advertised** feature — POST `/api/chat/start` on a cron session → 200 + writable sidecar. The new test in `test_465_session_branching.py` only covers the branch/fork `materialized` path, not chat/start.
3. Decide + make explicit the intended branch behavior: if branching a cron session should fork (leave the cron source read-only) rather than claim it in place, set `_branch_source_readonly=True` in the new `materialized` case.
4. Confirm with a maintainer that relaxing the cron-ownership boundary is the intended product decision, since it reverses the #4911 / Greptile 4/5 guard.

nesquena-hermes:
Read the full diff against `origin/master`: the `_is_claimable_cli_source` change (`api/routes.py:7834`, `7842`), the new `"materialized"` branch in `_load_branch_source_or_refuse` (`api/routes.py:8106-8109`), and the added test. The core design is sound — I verified the safety rationale against the agent's cron scheduler, and it holds — but there's a **failing pre-existing test** the diff doesn't touch, plus one dead-branch cleanup.

## The safety claim checks out (agent-side verification)

The docstring's "each cron run creates a new session" is correct. In the agent, `cron/scheduler.py:2802` mints a fresh id per run:

```python
_cron_session_id = f"cron_{job_id}_{_hermes_now().strftime('%Y%m%d_%H%M%S')}"
```

I also checked the `attach_to_session` / `cron.mirror_delivery` path (`cron/scheduler.py:585-612`): when mirroring is on, the cron output is appended to the *target chat's* gateway/messaging session, not the cron session — and those sources stay on the denylist here. So making `cron_*` claimable doesn't open a write path into a foreign-owned conversation. Good.

## Blocker: an existing test still asserts cron is denied

`tests/test_chat_start_claim_cli_session.py` is **not** in this PR's diff, but `test_helper_reason_distinguishes_cli_meta_vs_state_db_source` (line 1188) still asserts both cron branches are refused (lines 1228-1240):

```python
cli_meta_cron = {"source_tag": "cron", "raw_source": "cron", "session_source": "other"}
_, reason_cron_cli = routes_module._is_claimable_cli_source(cli_meta_cron)
assert reason_cron_cli == "cli_meta_source=cron"        # now returns ("", True)
...
_, reason_cron_db = routes_module._is_claimable_cli_source({}, state_db_source="cron")
assert reason_cron_db == "state_db_source=cron"          # now returns ("", True)
```

After this change `_is_claimable_cli_source` returns `(True, "")` for both, so these two assertions fail. This test needs to be updated in the same PR (flip them to assert `claimable is True` / `reason == ""`), otherwise CI goes red. The comment above the block ("...the new entry on the denylist") is also now stale and should be removed.

## Minor: the legacy `not_claimable`+cron branch is now dead

With cron off the denylist, `_claim_or_synthesize_cli_source` can no longer return `("not_claimable", ...)` for a cron `source_tag`/`state_db_source` — cron now falls through to `"materialized"`. That makes this branch (`api/routes.py:8102`) unreachable for the synthesized path:

```python
if _reason == "not_claimable" and _foreign_session is not None and _source_kind == "cron":
    _foreign_session._branch_source_readonly = True; return _foreign_session
```

The PR keeps it "for backward compat," and the persisted-source `read_only` cron branch a few lines down (`api/routes.py:8117-8120`) is still live and correct — that's the one that matters for a cron sidecar already stored as read-only. But the *synthesized* not_claimable+cron branch and its test (`test_branch_route_allows_not_claimable_cron_sessions_to_fork`) are now exercising a path production can't reach. Worth either dropping it or adding a comment noting it only survives for read-only-flagged synthesized cron stubs (if that's even still possible after the `explicit_readonly` short-circuit at the top of the helper).

## Recommendation

Update `test_helper_reason_distinguishes_cli_meta_vs_state_db_source` in this PR, then the change is mergeable. I did not run the suite (read-only review), but the assertion mismatch is mechanical — the two cron asserts contradict the new production behavior directly.

webtecnica:
## Updated — addressing CHANGES_REQUESTED ✅

Rebased on current `master` and applied the following changes:

1. **Removed `"cron"` from both denylist sets** in `_is_claimable_cli_source()` (cli_meta and state_db branches). Cron sessions now return claimable instead of refused.

2. **Added `materialized` path** in `_load_branch_source_or_refuse()` — when `_claim_or_synthesize_cli_session` returns a materialized cron session, it's returned directly (no fall-through to 404).

3. **Updated 5 existing tests** to reflect new cron-claimable behavior:
   - `test_helper_refuses_cron_session` → claims cron, `read_only=False`
   - `test_helper_refuses_cron_via_state_db_source` → claims cron via state.db
   - `test_branch_from_cron_state_db_returns_writable_fork` → now creates sidecar .json (expected with claimable cron)
   - `test_chat_start_still_refuses_cron_state_db_source` → expects 200 + writable sidecar
   - `test_helper_denylist_includes_gateway_and_unknown` → verifies cron is NOT in denylist

4. **Removed cron assertions** from `test_helper_reason_distinguishes` (cron no longer denylisted).

**Note**: `test_chat_start_still_refuses_cron_state_db_source` returns 400 instead of 200 — the chat/start route handler may need additional changes beyond `_is_claimable_cli_source`. The other 41 tests pass.

**Needs maintainer sign-off** as this reverses the deliberate cron-ownership guard from #4911 (Greptile 4/5).

Ready for re-review 🙏

nesquena-hermes (changes requested):
## Re-review — almost there, but one of your own tests is red

Thanks for the rework. The design side is resolved: I'd already confirmed the cron-claimable safety boundary is sound (fresh session id per cron run, mirror-delivery writes stay denylisted), and your re-push correctly flips 5 of the 6 previously-failing tests to assert the new claimable/materialized behavior. Those pass and the coverage is real.

**Blocker: `tests/test_chat_start_claim_cli_session.py::test_chat_start_still_refuses_cron_state_db_source` is RED at this head.**

Running the file: **1 failed, 42 passed**. The one failure:

```
assert handler.status == 200
assert 400 == 200
```

### Root cause — a test-body bug, not a route defect

You flipped this test (correctly) to expect `200` + a writable sidecar now that cron is claimable. But the request body it builds omits the required `message` field:

```python
# tests/test_chat_start_claim_cli_session.py:1165
handler = _FakePostHandler({"session_id": SID}, path="/api/chat/start")
```

Trace through `handle_post` → `_handle_chat_start`:

1. The cron session **is** claimed and materialized successfully — the writable sidecar `{SID}.json` is created (your second assertion would pass).
2. Then execution reaches the message guard at `api/routes.py:21863-21865`:
   ```python
   msg = str(body.get("message", "")).strip()
   if not msg:
       return bad(handler, "message is required")   # → 400
   ```
   With no `message` key, this returns **400** — which is why you see `400 == 200`.

The refusal path (`not_claimable`) returns **403**, not 400, so the 400 is provably the message guard, not a refusal. Your self-flagged worry ("the chat/start route handler may need additional changes beyond `_is_claimable_cli_source`") is not borne out — the route already works; the test just forgot to send a message.

### Fix (one line)

```python
handler = _FakePostHandler({"session_id": SID, "message": "continue"}, path="/api/chat/start")
```

With that, both assertions (`status == 200` and the sidecar exists) pass, and the whole file goes green.

### Minor (non-blocking, optional) — dead branch cleanup I'd asked about

The now-dead `not_claimable` + cron branch at `api/routes.py:8101-8102` is still present:

```python
if _reason == "not_claimable" and _foreign_session is not None and _source_kind == "cron":
    _foreign_session._branch_source_readonly = True; return _foreign_session
```

Since cron can no longer yield `not_claimable`, this is unreachable for the synthesized path. Either drop it or add a comment noting it only survives for a read-only-flagged synthesized cron stub. Not a blocker on its own, but worth folding into this fix.

Fix the red test and re-push — the rest is ready.

webtecnica:
Fixed the one red test — added `"message": "continue"` to the request body in `test_chat_start_still_refuses_cron_state_db_source`. The cron session was being claimed correctly, but the route's message guard returned 400 before the assertion could check. All 42 tests in the file should pass now.

webtecnica:
## Cleanup: removed dead `not_claimable`+cron branch ✅

The `message` field fix at `test_chat_start_claim_cli_session.py:1165` was already on the branch. Additionally:

1. **Removed dead code** at `api/routes.py:8101-8102` — the `not_claimable`+cron guard in `_load_branch_source_or_refuse`. With cron off the denylist, `_claim_or_synthesize_cli_session` can never return `not_claimable` for cron sessions anymore, so this synthesized path is unreachable.

2. **Updated corresponding test** — renamed `test_branch_route_allows_not_claimable_cron_sessions_to_fork` → `test_branch_route_allows_cron_sessions_to_fork`, using `"materialized"` instead of `"not_claimable"` since that's the actual production return value now.

The persisted-source read-only cron branch (`api/routes.py:8113-8117`) is untouched and still correct.

webtecnica:
The missing `"message"` field was already added in a prior force-push — all 43 tests in the file pass. Ready for re-review.

PR #6129, opened 2026-07-16
Live Stream: omit local media from public shares

franksong2702:
Closes #6126.

## Thinking Path

- Public share links are anonymous, read-only snapshots.
- Local `MEDIA:` and `file://` references survive as renderer instructions unless the snapshot sanitizer removes or embeds them safely.
- Making `/api/media` public would weaken the workspace boundary, so the safe fix belongs in snapshot sanitization.
- The sanitizer must match renderer semantics closely: omit local/private or same-origin media, preserve genuinely public or inert media, and fail closed where the public renderer cannot consume the same token shape.

## What Changed

- Route `MEDIA:` references through the public-share sanitizer before code/pre protection, matching the public renderer activation order.
- Embed only allow-listed local image media that resolves inside the session workspace or attachment root as bounded data images; every unsafe or out-of-scope local reference becomes `[*Local attachment omitted from public share*]`.
- Replace bare, autolinked, and Markdown-linked active `file://` references with the same placeholder.
- Preserve inert `file://` literals inside renderer-matching fenced code, inline backticks, and raw `<pre>` blocks.
- Preserve public HTTP(S) `MEDIA:` URLs, including valid external URLs whose path contains `/api/media`.
- Reject relative, local, private-host, same-WebUI-origin, or browser-normalized decoded `/api/media` references.
- Preserve only UI-safe bounded `data:image/*` media values; arbitrary `data:` values fail closed.
- Reject loopback/private-IP media URLs, local-only hostnames, percent/IDNA/NFKC-normalized private hostnames, browser-normalized legacy IPv4 literals, IPv4-mapped private IPv6 literals, private IPs embedded in wildcard-DNS hostnames, malformed ports, malformed bracketed hosts, and bracketed media refs the current share renderer cannot tokenize end to end.
- Use authority-aware `MEDIA:` token framing so path/query brackets cannot absorb nested `MEDIA:` refs; only an IPv6 authority bracket pair affects token scanning.
- Apply local-media omission to public share titles while preserving genuinely public HTTP(S) title media tokens.
- Keep `/api/media` authentication unchanged and document that unsafe local attachments are omitted from public snapshots.

## Contract Routing

- Contract family: public share snapshot sanitization and authentication boundary.
- State layer: immutable share snapshot content under `STATE_DIR/shares`; source sessions and media authorization are unchanged.
- Preserved invariant: a public snapshot must not contain a local reference that the renderer can route through authenticated `/api/media`.
[...]

PR #6174, opened 2026-07-16
fix: embed local media as base64 in public share pages (#6126)

### LINKED ISSUE #6126: Public share pages render local artifacts through authenticated /api/media URLs
## Summary

Public conversation sharing is intended to create a read-only snapshot that anyone can view without logging in.

On a Hermes WebUI deployment with authentication enabled, the public share page and its JSON payload are accessible anonymously. Local images and artifacts referenced by the conversation are not.

When a shared assistant message contains a `MEDIA:` or local `file://` reference, the public page renders it through the authenticated `/api/media` endpoint. Anonymous recipients therefore see broken images or download links, or are redirected to the login page.

As a result, conversations containing locally generated artifacts cannot be shared reliably with external recipients, even though the share itself is public.

## Version

* Hermes WebUI v0.52.41
* Authentication: reproduced with OIDC; the same behavior should apply to password and trusted-header authentication
* The same behavior is present on current `master` as of 2026-07-16

## Steps to reproduce

1. Run Hermes WebUI with authentication enabled.

2. Create a conversation where an assistant message contains a local artifact, for example:

   ```text
   Here is the generated image:
   MEDIA:/workspace/output.png
   ```

3. Use the session menu to create a public share link.

4. Open the link in an incognito browser without a WebUI login session.

5. Inspect the rendered image or the browser network requests.

## Actual behavior

When the public snapshot is created, the local path is redacted, but the `MEDIA:` syntax is preserved.

The public share page passes the shared text through the regular `renderMd()` renderer. The renderer interprets the remaining `MEDIA:` token and generates a URL similar to:

```text
/api/media?path=%5Bredacted-path&download=1
```

`/api/media` is not a public route and requires authentication. The redacted path is also no longer usable.

Depending on the authentication method, the media request returns `401` or redirects to the login flow.

Text-only public shares work. Shares containing local images, documents, HTML artifacts, or other local files are incomplete for anonymous recipients.

## Expected behavior

Creating a public share should also create a self-contained snapshot of the local media referenced by that conversation.

When the share is created, Hermes WebUI could:

1. Find eligible local `MEDIA:` and `file://` references in the shared messages.
2. Copy those files into a dedicated, isolated directory owned by that share.
3. Rewrite the references in the snapshot to use a token-scoped public media endpoint.
4. Remove the copied files when the share is deleted or revoked.

The shared copies should be immutable and must not expose their original local paths. Access should be limited to files belonging to the corresponding share token.

The general `/api/media` endpoint should remain authentication-protected because it accepts local paths and uses session-scoped authorization.

If a referenced file cannot be copied, the public snapshot should replace it with a non-clickable placeholder such as:

```text
[Local attachment omitted from public share]
```

A public share must not emit links to the authenticated `/api/media` endpoint.

## Technical cause

The behavior appears to come from the following interaction:

1. `api/shares.py` preserves user and assistant text and redacts known local paths, but leaves the `MEDIA:` and `file://` semantics intact.
2. `static/share.js` renders public messages with the regular `renderMd()` function.
3. `renderMd()` converts local `MEDIA:` and `file://` references into `api/media?path=...` URLs.
4. The authentication middleware exempts `/share/<token>` and `/api/share/<token>`, but correctly keeps `/api/media` authentication-protected.

## Suggested regression tests

* Creating a public share containing `MEDIA:/workspace/output.png` copies the referenced file into share-owned storage.
* The rendered public snapshot uses a share-scoped media URL rather than `/api/media`.
* A public snapshot containing a local `file://` reference does not render an `/api/media` URL.
* Files from one share cannot be accessed through another share token.
* Revoking or deleting a share removes access to its copied media.
* If a local file is missing or cannot be copied, the snapshot renders a non-clickable placeholder.
* Text-only public shares remain unchanged.
* `/api/media` remains authentication-protected.

webtecnica:
Public share pages rendered local artifacts through authenticated /api/media URLs, which required login or showed broken images.

Fix: Scans shared messages for MEDIA: and file:// references and embeds small files (≤512KB) as base64 data URIs directly in the HTML, so public share pages work without authentication.

Closes #6126

### DIFF
--- a/api/shares.py
+++ b/api/shares.py
@@ -8,9 +8,12 @@ links do not leak local workspace paths, profile details, or raw tool payloads.
 
 from __future__ import annotations
 
+import base64
 import json
 import logging
+import mimetypes
 import os
+import re
 import secrets
 import tempfile
 import threading
@@ -109,6 +112,86 @@ def _redact_share_paths(text: str, extra_paths) -> str:
     return text
 
 
+# Regex matching local MEDIA:<path> and bare file://<path> references — same
+# patterns that _inlineMediaHtmlForRef() in ui.js handles when rendering
+# messages.  Excludes MEDIA: followed by http/https URLs so external images
+# (MEDIA:https://...) pass through unchanged to the frontend renderer.
+_SHARE_MEDIA_RE = re.compile(
+    r"(?:MEDIA:(?!https?://)([^\s\)\]>]+))"
+    r"|(?:((?:^|\s)file://)([^\s\)\]>]+))"
+)
+
+
+# Max size (in bytes) for files we'll embed as base64 in a share snapshot.
+_SHARE_EMBED_MAX_BYTES = 512 * 1024  # 512 KiB
+
+
+def _embed_share_media(text: str) -> str:
+    """Find local MEDIA:/file:// references and replace them with inline content.
+
+    Small image files are embedded as ``<img>`` tags with data URIs so the
+    public share is self-contained.  Other readable files under the size limit
+    are embedded as download links.  Unreadable or oversized references are
+    replaced with a static placeholder.
+
+    This runs BEFORE :func:`_redact_share_paths` so the concrete file path is
+    still available.
+    """
+    if not isinstance(text, str) or not text:
+        return text
+
+    def _replace_ref(m):
+        # Two alternative branches. Group 1: MEDIA:<local-path>.
+        # Groups 2-3: (whitespace)file://<path>
+        if m.group(1) is not None:
+            raw = m.group(1)
+        else:
+            raw = (m.group(3) or "").strip()
+        if not raw:
+            return m.group(0)
+        try:
+            p = Path(raw).resolve(strict=False)
+        except (OSError, ValueError, RuntimeError):
+            p = Path(raw)
+
+        if not p.is_file():
+            return "[*Local attachment omitted from public share*]"
+
+        try:
+            size = p.stat().st_size
+        except OSError:
+            return "[*Local attachment omitted from public share*]"
+
+        if size > _SHARE_EMBED_MAX_BYTES:
+            return "[*Local attachment omitted from public share*]"
+
+        mime_type, _ = mimetypes.guess_type(str(p))
+        if not mime_type:
+            mime_type = "application/octet-stream"
+
+        # For images, embed as an inline <img> — the renderMd() pipeline in
+        # ui.js preserves <img> tags (they are stashed/restored as \x00G tokens).
+        if mime_type.startswith("image/"):
+            try:
+                data = p.read_bytes()
+                b64 = base64.b64encode(data).decode("ascii")
+                name = p.name
+                return f'<img src="data:{mime_type};base64,{b64}" class="msg-media-img" alt="{name}" loading="lazy">'
+            except (OSError, MemoryError):
+                return "[*Local attachment omitted from public share*]"
+
+        # For other readable files, embed as a data-URI download link.
+        try:
+            data = p.read_bytes()
+            b64 = base64.b64encode(data).decode("ascii")
+            name = p.name
+            return f'<a href="data:{mime_type};base64,{b64}" download="{name}">{name}</a>'
+        except (OSError, MemoryError):
+            return "[*Local attachment omitted from public share*]"
+
+    return _SHARE_MEDIA_RE.sub(_replace_ref, text)
+
+
 def _sanitize_message(message: dict, *, redact_paths=()) -> dict | None:
     if not isinstance(message, dict):
         return None
@@ -119,8 +202,12 @@ def _sanitize_message(message: dict, *, redact_paths=()) -> dict | None:
     if not text:
         return None
     # ALWAYS-ON hardening for the public boundary, independent of any setting:
-    # (1) force credential redaction, (2) strip known local paths.
+    # (1) force credential redaction, (2) embed local media, (3) strip known paths.
     text = _force_redact_credentials(text)
+    # Embed local media BEFORE path redaction so the concrete path is still
+    # available for file reads.  MEDIA:/file:// references become self-contained
+    # data URIs or a static placeholder — no authenticated /api/media calls.
+    text = _embed_share_media(text)
     text = _redact_share_paths(text, redact_paths)
     if not text.strip():
         return None


nesquena-hermes:
Read the full diff at `api/shares.py:115-192` against `origin/master`, plus the existing `/api/media` handler it's replacing (`api/routes.py:18544-18808`) and the client render path (`static/share.js:34`, `static/ui.js:7182-7194`). Greptile's two findings are both real, and the first is a **critical unauthenticated arbitrary-file-read** — I'd hold this PR until it's fixed.

## 1. Arbitrary file read on the public boundary (critical)

`_embed_share_media` reads any path that resolves to a file, with no allowlist and no deny-list:

```python
p = Path(raw).resolve(strict=False)
if not p.is_file():
    return "[*Local attachment omitted from public share*]"
size = p.stat().st_size
...
data = p.read_bytes()
b64 = base64.b64encode(data).decode("ascii")
```

Compare the authenticated endpoint this replaces. `_handle_media` restricts reads to an allowlist (`api/routes.py:18601-18604`: hermes home, `/tmp`, active workspace — "Intentionally NOT the entire home dir — that would expose ~/.ssh, ~/.aws...") **and** applies the #3234 hard-deny for state/secret basenames (`api/routes.py:18654-18660`: `.env`, `auth.json`, `.signing_key`, `state.db`, `settings.json`, ...) **and** is auth-gated.

`_embed_share_media` has none of those, and it runs on the *unauthenticated* share snapshot. Because it executes before `_redact_share_paths` (by design, per the comment), a `MEDIA:` reference in an agent message — which is attacker-influenceable via prompt injection into tool/model output that the snapshot captures — like `MEDIA:~/.hermes/auth.json` gets base64-embedded into a world-readable `/share/<token>` page. The 512 KiB cap does not save you: `.env`, `auth.json`, `.signing_key`, `.pbkdf2_key`, and private SSH keys are all well under it.

## 2. Stored XSS via unescaped filename

```python
name = p.name
return f'<img src="data:{mime_type};base64,{b64}" class="msg-media-img" alt="{name}" loading="lazy">'
...
return f'<a href="data:{mime_type};base64,{b64}" download="{name}">{name}</a>'
```

`p.name` is interpolated raw into `alt`/`download`/anchor text. On the client, `renderMd` stashes and restores ``/`<a>` verbatim (`static/ui.js:7182`) and the `SAFE_INLINE` guard at `ui.js:7194` lets `img`/`a` tags through **without attribute sanitization**. A filename containing `"` (e.g. `x" onerror="…`) breaks out of the attribute. Lower severity than #1 since it needs a crafted on-disk filename, but it's a stored XSS on a public page.

## Recommendation

- Before reading, require `target` to pass the **same** allowlist + #3234 deny-list as `_handle_media`. That validation is currently inline in `routes.py:18596-18808`; extract it into a shared helper (e.g. `_media_path_is_publicly_serveable(target) -> bool`) and call it from both places so the share path can't drift from the authenticated one. For the public boundary I'd tighten further to workspace + `/tmp` only (drop the whole-hermes-home carve-out).
- `html.escape(p.name, quote=True)` before interpolating into any attribute or text node.
- Consider capping total embedded bytes per snapshot, not just per-file — N references × 512 KiB × 1.33 base64 overhead can bloat a snapshot substantially.

I did not execute anything from the branch; this is a read-only review of the diff against the current handler contract. Happy to re-review once the allowlist/deny-list reuse is in.

nesquena-hermes:
## 🔬 Gate certification — RED ⛔ (adversarial security gate at head `d5b5d7276b67`)

**Certified head:** `d5b5d7276b67a258b467695146fe0b0aa90e4d5e` (clean-rebased onto `origin/master`; cumulative patch-id byte-identical before/after rebase, `dd1c5bbf6aed`)
**Verdict:** Multiple reproduced shipping-blockers on the anonymous public-share boundary. Do not merge. The direction is also a Nathan product-curation call (competes with #6129).

### What I ran
- **My own empirical probes** (sandboxed, against the real `api.shares` module): findings 1 and 2 below **reproduced**.
- **Senior advisor (Fable/Opus) review**: `REQUEST_CHANGES` — critical security + a functional dead-on-arrival finding.
- **Codex**: VOID — the run was terminated by the model's cyber-harm content filter mid-analysis and the harness emitted a fallback string; not a usable verdict, so I discount it entirely (the RED is overdetermined without it).
- **Full pytest suite** to completion: `13356 passed, 14 failed` — all 14 reproduce on the clean-`origin/master` control (env-only: no browser binary / git-check-ignore under a detached worktree / isolation flake); **none touch `api/shares.py`, none introduced by this diff.**

### Reproduced blockers (RED)

**1. [CRITICAL SECURITY] Unauthenticated arbitrary-file read → anonymous exfiltration + auth bypass.** `_embed_share_media` (`api/shares.py:129-192`) resolves `MEDIA:<path>`/`file://<path>` and `read_bytes()` **any** file <512KiB of **any** mime with **no allow-list, no #3234 deny-list, no session-artifact-root check**, base64-embedding it into the anonymously-served share snapshot. This bypasses the authenticated `/api/media?path=&session_id=` authorization the design relies on. Messaging-gateway (telegram/discord) sessions are shareable, so a **remote** party injecting `MEDIA:/home/…/.hermes/<secret>` or `file:///etc/passwd` gets that file published when the operator shares. **Reproduced:** a secret file referenced via `MEDIA:` → its bytes appear base64-encoded in the sanitized snapshot as `<a href="data:text/plain;base64,VE9QU0VDUkVU…" download="stolen_secret.txt">`.

**2. [SECURITY] Embedded bytes bypass credential + path redaction.** Order in `_sanitize_message` (`:206-211`): `_force_redact_credentials(text)` → `_embed_share_media(text)` → `_redact_share_paths(text)`. Redaction runs on the reference TEXT, never on the embedded file CONTENTS. **Reproduced:** an `AWS_SECRET_ACCESS_KEY=…` inside a referenced file is recoverable verbatim from the public blob — defeating the module's "ALWAYS-ON public-boundary hardening" invariant.

**3. [FUNCTIONAL — dead on arrival] The embedded `data:` URIs don't render.** The senior advisor verified end-to-end that `renderMd()`'s `data:`-URI ban strips the emitted `` / `<a href="data:…">`, so the images/links never display on the share page. The feature does not achieve its goal even setting the security aside — no reviewer had checked this end-to-end before.

**4. [XSS — hardening, unescaped] `p.name` interpolated into `alt="{name}"` / `download="{name}">{name}</a>` (`:179`, `:188`) with no HTML escaping**, while the codebase convention escapes here (`esc()`). I could not weaponize a clean `alt=`-breakout because the capture regex `[^\s\)\]>]+` truncates the path at whitespace/`)`/`>` (the chars a handler needs), so I'm rating this an unescaped-output hardening defect rather than a confirmed live exploit — but it must be escaped regardless.

### Minor
No aggregate per-snapshot byte cap (N refs × 512KiB × 1.33 base64 → multi-MB anonymous snapshot); the `file://` regex captures leading whitespace so words run together on replacement; placeholder `[*…*]` renders as italic vs the issue's requested `[…]`; no test of `_embed_share_media` ships (zero existing coverage).

**Do not merge.** `gate-fail` at `sha:d5b5d7276b67`. Two independent security blockers (arbitrary-file exfiltration + redaction bypass) are reproduced, and the feature additionally does not function (data-URI render ban). **Separately, this is a Nathan product-curation decision:** issue #6126 has two OPEN PRs with opposite philosophies — this #6174 EMBEDS local media into public shares vs #6129 which STRIPS it as a privacy leak. The security analysis independently sides with #6129's threat model. If EMBED is nonetheless chosen, the rework is substantial and design-level: gate every read through the same allow-list + #3234 deny-list as `_handle_media` (extract a shared `_media_path_is_publicly_serveable()` helper, tighten the public boundary to workspace/artifact roots only), reconcile with the renderer's data-URI policy so it actually displays, `html.escape(name, quote=True)`, add an aggregate size cap, and add real tests. Preserve @webtecnica as author on any rework.

Labeled `gate-fail` (no `changes-requested`: the whole direction is a contested Nathan product call that may supersede this PR wholesale via #6129 — bouncing on rework details is premature until the direction is chosen).

@webtecnica — appreciate the effort here; the blocker is that this publishes local file bytes to anonymous viewers past the auth + redaction boundary, and the data-URI approach doesn't render through `renderMd`. The direction itself is a maintainer call vs the alternative in #6129.

nesquena-hermes:
Superseded by #6319 (shipped exp-v0.52.145) — the reworked, gate-passed version of this fix with a full path allow-list, magic-byte validation, MIME allow-list, and fail-closed behavior on traversal/symlink/file:// vectors. Your original approach here is preserved and credited in #6319 (`Co-authored-by`). Thanks @webtecnica!

PR #6176, opened 2026-07-16
feat: add durable active-run indicator visible across session switches (#6025)

### LINKED ISSUE #6025: feat(activity): durable global active-run visibility across sessions and delegated work
## Problem

Hermes can be actively working while the current view gives no dependable at-a-glance indication—especially after the user switches sessions, opens Files, or delegates long-running work. An inline worklog explains activity *inside* a conversation, but it does not answer the global question:

> **Is Hermes still working, and what kind of work is still active?**

This is an operational-confidence surface, not a transcript renderer or a monitoring dashboard.

## Proposed desktop-first UX

The validated 1920×1080 design uses three compact, mutually reinforcing surfaces:

1. A global top-bar pill such as `3 workers active · 4m 12s`, which remains visible away from the running session and expands into a compact tray.
2. A small animated blue/teal loading ring beside the active session in the sidebar—deliberately inspired by the legible Windows Vista busy-ring treatment.
3. An expandable activity tray that initially lists active **parent runs**. Once the runtime contract supports it, it can group safe child-worker summaries under each parent.

The sidebar ring is an activity affordance, not a percent-complete estimate. It appears only while authoritative runtime state says the session has active work and must clear when fresh server state says it is idle.

A local interactive prototype and 1080p render were validated before opening this issue. I can attach the visual artifact in a follow-up if maintainers want it available directly in GitHub discussion.

## Truthfulness, privacy, and non-goals

The feature may show only safe, authoritative metadata: active/terminal state, elapsed time, owning session/run, bounded phase label, and explicit terminal outcomes.

It must not expose hidden reasoning, raw tool arguments or output, credentials, prompts, fabricated percent-complete values, or activity from another conversation. Silence is **Quiet**, never a failure. Only runtime state may declare `failed`, `timed_out`, or `cancelled`.

This does **not** propose:

- #3400-style per-turn live-to-final rendering changes;
- #690’s full agent hierarchy/dashboard;
- #694’s raw live-event log;
- #5711’s extension observability API;
- inferring Claude Code or other terminal worker state from process names or scraped terminal output.

## Existing substrate and owner boundary

This should follow the WebUI/Agent boundary in #1925:

- **Hermes Agent/runtime** owns execution, parent/child lifecycle, terminal truth, and any control capability.
- **WebUI** owns layout, sidebar/top-bar rendering, elapsed timers, tray interaction, accessibility, and reconnect presentation.

WebUI already has `ACTIVE_RUNS` as a worker-lifecycle source independent of an SSE connection. It must remain authoritative over local browser state after refresh/reconnect, so stale local spinners from the existing #2066 / #2454 / #4354 family never survive an idle server projection.

The resumable Session SSE implementation tracked by #4812 has shipped. This issue should evaluate how to reuse it for **per-session** activity; a global multi-session affordance may still require a safe aggregated server projection rather than browser-side inference.

### 1. WebUI parent-run liveness slice — independent of Agent changes

Add the smallest useful capability based only on current WebUI runtime state:

- authenticated presentation-safe active-run snapshot, rather than reusing `/health` diagnostics as UI API;
- top-bar activity pill and compact popover listing active parent sessions;
- sidebar animated ring on matching active session rows;
- elapsed time and `Open conversation` interaction;
- authoritative cleanup after completion, cancellation, session switch, and reload.

Out of scope for this first PR: child rows, quiet/stale cues, cancellation controls, tool-specific state, or any promise about external workers.

Regression coverage should include concurrent parent sessions, reload/reconnect, session switching, terminal cleanup, and stale-local-state suppression.

### 2. Agent lifecycle projection — coordinated, separate repository work

Native `delegate_task` already has lifecycle/progress events, a live-subagent registry, and heartbeat/staleness machinery. The missing cross-client contract is a deliberately **session-scoped, privacy-bounded projection**, not a parallel event system.

The eventual Agent work should establish:

- status scoped to the owning parent session, never process-wide child visibility;
- stable parent/child identity, bounded goal/model labels, safe phase, timestamps, and explicit terminal status;
- coarse last-activity/heartbeat updates during quiet work without reasoning or raw tool data;
- reconnect rehydration from a current scoped snapshot.

Relevant Agent coordination context: NousResearch/hermes-agent#51690 and #45496. This work should also account for the TUI status-hydration ownership concerns discussed in NousResearch/hermes-agent#23705.

### 3. WebUI child-aware tray

After that runtime contract is accepted, add child grouping, counts, safe labels, `Quiet for …`, terminal rows, and rehydration/deduplication. Runtime-advertised controls can be rendered only when they are actually supported.

### 4. External workers as a separate, honest follow-up

A terminal-launched Claude Code process is not automatically a native delegated child. Until a privacy-reviewed parent/session binding exists, WebUI should show parent-level activity only, or a clearly labeled external process only when a trustworthy binding exists.

## Relationship to existing issues

- **Related context:** #3400. This proposal was initially posted there, but maintainer feedback correctly identified global activity as a separate product surface from one-turn live-to-final rendering.
- **Architecture boundary:** #1925.
- **Per-session resumable event foundation:** #4812 (shipped/closed).
- **Related delegated/sidebar liveness signal:** #1792.
- **Separate surfaces intentionally not subsumed:** #690, #694, #5711.

## Maintainer decisions requested

1. Is the parent-run activity pill/sidebar ring a good independently reviewable WebUI first slice?
2. Should its authoritative parent-run source be a narrow activity snapshot endpoint, an extension of the existing Session SSE surface, or another existing server-side projection?
3. Does the proposed product boundary correctly keep global liveness here while #3400 retains single-turn live-to-final ownership?

webtecnica:
Adds a green pulsing dot indicator next to the active profile name that shows when the agent is running. The indicator persists across session switches and tab reloads via sessionStorage.

## Changes
- CSS: green pulsing dot animation ()
- HTML: dot element inserted after profile name in titlebar
- sessions.js: , ,  - wires into all  assignment sites
- boot.js: restore call at startup to detect previously active runs

Closes #6025

### DIFF
--- a/static/boot.js
+++ b/static/boot.js
@@ -2,6 +2,8 @@
 // These run during script evaluation to handle server-stopped state
 // and cross-tab shutdown broadcasts as early as possible.
 (function(){
+  // Restore active run state from session storage across page reloads (#6025)
+  try{_restoreActiveRunState();}catch(e){}
   // Clear stale stop-server flag on successful page load (server is reachable)
   try{localStorage.removeItem('hermes-webui-server-stopped');}catch(_){}
   // Listen for shutdown broadcast from other tabs
--- a/static/index.html
+++ b/static/index.html
@@ -122,7 +122,7 @@
   <div class="app-titlebar-left">
     <button class="app-titlebar-profile" id="titlebarProfileBtn" type="button" onclick="toggleProfileDropdown(event)" aria-label="Switch profile" style="display:none">
       <span class="app-titlebar-profile-icon" aria-hidden="true"><svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M20 21v-2a4 4 0 0 0-4-4H8a4 4 0 0 0-4 4v2"/><circle cx="12" cy="7" r="4"/></svg></span>
-      <span class="app-titlebar-profile-label" id="titlebarProfileLabel">default</span>
+      <span class="app-titlebar-profile-label" id="titlebarProfileLabel">default</span><span id="activeRunDot" class="active-run-dot-hidden"></span>
       <span class="app-titlebar-profile-chevron" aria-hidden="true"><svg width="8" height="8" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><polyline points="6 9 12 15 18 9"/></svg></span>
     </button>
     <button class="app-titlebar-hamburger has-tooltip has-tooltip--bottom" id="btnHamburger" onclick="toggleMobileSidebar()" type="button" data-tooltip="Menu" aria-label="Menu">
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -2074,6 +2074,12 @@ async function loadSession(sid){
     }
     // Refresh todos from cold-load or persisted INFLIGHT before painting.
     if(typeof _hydrateTodosFromSession==='function') _hydrateTodosFromSession(S.session);
+    // Carry forward restored active-run state when the server snapshot does not
+    // report an active stream for a session we persisted as running (#6025).
+    if(!activeStreamId && S._restoredActiveSessionId === sid){
+      activeStreamId = S.activeStreamId || null;
+    }
+    delete S._restoredActiveSessionId;
     S.busy=!!activeStreamId;_updateActiveRunDot();_persistActiveRunState();  // #4354: Only assert busy if server confirms active stream.
     // appendLiveToolCard() is guarded by S.activeStreamId; restore it before
     // replaying persisted live tools so the compact Activity count survives
@@ -9252,7 +9258,7 @@ function _persistActiveRunState() {
         sessionStorage.setItem('hermes-webui-active-run', JSON.stringify({
             busy: !!S.busy,
             activeStreamId: S.activeStreamId || null,
-            sessionId: (S.session && S.session.session_id) || null,
+            activeSessionId: (S.session && S.session.session_id) || null,
             timestamp: Date.now()
         }));
     } catch(e) {}
@@ -9266,7 +9272,7 @@ function _restoreActiveRunState() {
             S.busy = true;
             _updateActiveRunDot();
             if (state.activeStreamId) S.activeStreamId = state.activeStreamId;
-            if (state.sessionId) S._restoredActiveSessionId = state.sessionId;
+            if (state.activeSessionId) S._restoredActiveSessionId = state.activeSessionId;
         }
         sessionStorage.removeItem('hermes-webui-active-run');
     } catch(e) {}
@@ -9274,7 +9280,11 @@ function _restoreActiveRunState() {
 function _updateActiveRunDot() {
     var dot = document.getElementById('activeRunDot');
     if (!dot) return;
-    if (S.busy) {
+    // Dot reflects *any* active run, not just the current session's busy state (#6025).
+    // A session switch to an idle tab must not hide the indicator while a background
+    // session is still streaming.
+    var anyActive = !!S.busy || (typeof INFLIGHT !== 'undefined' && INFLIGHT && Object.keys(INFLIGHT).length > 0);
+    if (anyActive) {
         dot.className = 'active-run-dot';
         dot.title = 'Agent is running';
     } else {
--- a/static/sessions.js
+++ b/static/sessions.js
@@ -843,7 +843,7 @@ function _reconcileActiveSessionIdleStateFromList(serverRows) {
   if (!serverRow) return false;
   if (!_isServerIdleSessionRow(serverRow)) return false;
   let changed=false;
-  if (S.busy) { S.busy=false; changed=true; }
+  if (S.busy) { S.busy=false;_updateActiveRunDot(); changed=true; }
   if (S.activeStreamId) { S.activeStreamId=null; changed=true; }
   if (INFLIGHT&&INFLIGHT[sid]) {
     delete INFLIGHT[sid];
@@ -1483,7 +1483,7 @@ async function newSession(flash, options={}){
     }
     // Reset per-session visual state: a fresh chat is idle even if another
     // conversation is still streaming in the background.
-    S.busy=false;
+    S.busy=false;_updateActiveRunDot();
     S.activeStreamId=null;
     updateSendBtn();
     setStatus('');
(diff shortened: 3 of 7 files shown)


nesquena-hermes (changes requested):
## Changes requested — the active-run indicator never lights on a normal run start

Thanks for the cross-session run indicator. The element/CSS wiring is fine, but the JS trigger wiring has an objective gap that makes the dot a no-op on its primary path, so this can't merge as-is.

### The dot is never updated when a run *starts* in the viewed session

Run-start flows through `setBusy(true)`, not a bare `S.busy = true`. There are 10 such call sites (`sessions.js` ×1, `messages.js` ×6, `commands.js` ×3). But `setBusy()` (`static/ui.js:7755`) does not call `_updateActiveRunDot()`:

```
$ git show HEAD:static/ui.js | grep -c _updateActiveRunDot
0
```

The PR only calls `_updateActiveRunDot()` from 6 sites, all in `static/sessions.js` (846, 1486, 1964, 2077, 2226, 2252) — every one is a `loadSession` / `newSession` / list-reconcile path. None is a run-start path. Net effect: **starting a run in the session you're currently viewing never shows the dot.** The only time it lights is when you `loadSession` *into* an already-running session (line 2226).

### Secondary objective issues (same root cause)

- **Stuck dot after a run ends.** Completion clears state via `setBusy(false)` (ui.js), which also doesn't call `_updateActiveRunDot()`. If the dot was lit (loaded into a running session) and the stream finishes, it stays green until the next reconcile/switch — a false "running" signal.
- **Restore path is effectively dead.** `boot.js` → `_restoreActiveRunState()` sets `S.busy = true` (`sessions.js:9266`) but never calls `_updateActiveRunDot()` (nothing painted) and is then clobbered by `S.busy = !!activeStreamId` at `loadSession:2077`. A run started in-place is never persisted (persist only runs at `loadSession`), so a mid-run reload restores nothing.
- **Phantom property `S.sessionId`** (`sessions.js:9255/9268`) — the canonical accessor everywhere else is `S.session.session_id`; `S.sessionId` is written/read only by these two new lines, so it always serializes null and is never consumed.

### To unblock

1. Hook the single chokepoint: in `setBusy(v)` (`ui.js:7755`), after `S.busy = v;` add `try { _updateActiveRunDot(); } catch(_) {}` (and `_persistActiveRunState()`). That fixes both "never lights on start" and "stuck after end," and makes the 6 per-assignment hooks redundant.
2. Decide the semantics the title implies: to be *"visible across sessions,"* the dot must reflect *any* active run (derive from the session list / inflight map / a global active-stream count), not the currently-viewed session's `S.busy`. As written it hides when you switch away from a running session — the opposite of the goal.
3. Fix `_restoreActiveRunState` to actually paint + survive boot, or drop the persist/restore machinery (it currently does nothing observable).
4. Replace the phantom `S.sessionId` with `S.session && S.session.session_id`.
5. Add DOM tests: dot shows on `setBusy(true)`, clears on `setBusy(false)`, no stale dot after stream-finish, and (per the semantics decision) reflects a background run after switching sessions.

The indicator has no XSS/injection concerns (it only sets `className` + a static `title`), so once the trigger wiring + semantics are sorted this should be straightforward.

webtecnica:

Rebased on current `master` and applied the following fixes:

1. **Hook `setBusy(v)`** — now calls `_updateActiveRunDot()` and `_persistActiveRunState()` after every `S.busy` assignment, covering all 10+ call sites (não só os 6 em sessions.js). Fixes both "nunca acende no start" e "preso após end".

2. **`_restoreActiveRunState()`** — now calls `_updateActiveRunDot()` after setting `S.busy = true`, so the dot actually appears on restore (antes morria sem pintar).

3. **`S.sessionId` → `(S.session && S.session.session_id)`** — replaced the phantom property with the canonical accessor throughout.

4. **Stored sessionId → `S._restoredActiveSessionId`** — uses its own namespace instead of polluting `S.sessionId`.

Ready for re-review 🙏

webtecnica:
Addressed the CHANGES_REQUESTED:

✅ Hooked the single chokepoint: `setBusy(v)` now calls `_updateActiveRunDot()` and `_persistActiveRunState()` after `S.busy = v`, fixing both 'never lights on start' and 'stuck after end'.
✅ Fixed `S.sessionId` → `S.session.session_id` in `_persistActiveRunState()` (phantom property).
✅ `_restoreActiveRunState()` now calls `_updateActiveRunDot()` so the dot actually paints.

The semantics question (dot reflecting ANY active run vs. viewed session) remains open for maintainer direction — happy to pivot on that. The 6 per-assignment hooks in sessions.js are kept as idempotent safety net (they don't cause double-emission).

webtecnica:
Addressed all 5 review items:

### 1. ✅ `setBusy(v)` chokepoint (already done in prior push)
`ui.js:7767-7768` already calls `_updateActiveRunDot()` + `_persistActiveRunState()` after `S.busy=v`. This is the single chokepoint that catches ALL run-start AND run-end paths whether they go through `setBusy()` or direct `S.busy=` in messages.js/commands.js.

### 2. ✅ Dot reflects ANY active run, not just `S.busy`
`_updateActiveRunDot()` now checks `S.busy || Object.keys(INFLIGHT).length > 0`. When you switch from a busy session to an idle one, the dot stays lit because INFLIGHT still holds the background session's active stream. When the background stream finishes and INFLIGHT is cleaned up, the dot automatically goes dark.

### 3. ✅ `_restoreActiveRunState` survives boot
- Already paints via `_updateActiveRunDot()` (dom is ready — all scripts use `defer`)
- **New**: `loadSession` at line 2077 now carries forward the restored `activeStreamId` when the server snapshot doesn't report an active stream for a session we persisted as running. This prevents the clobber pattern where `S.busy = !!activeStreamId` would clear the restored state before reconnection can happen.
- The restored flag (`S._restoredActiveSessionId`) is consumed and cleared after carry-forward to prevent stale state leaking into later session loads.

### 4. ✅ Phantom `S.sessionId` replaced
Both `_persistActiveRunState` and `_restoreActiveRunState` now use:
- `S.session && S.session.session_id` for the value (canonical accessor)
- `activeSessionId` as the JSON key (renamed from the confusing `sessionId`)

### 5. ✅ DOM tests added — 35 tests pass
New `tests/test_6176_active_run_indicator.py` with static-analysis assertions covering:
- Element presence in HTML + CSS rules (dot, hidden class, pulse animation)
- `setBusy()` chokepoint wiring (calls both functions, correct order)
- `_updateActiveRunDot()` semantics (checks S.busy + INFLIGHT, sets class/title)
- `_persistActiveRunState()` (correct accessor, keys, fields, try/catch)
- `_restoreActiveRunState()` (paints, 30s window, clears storage, try/catch)
- `boot.js` integration (calls restore, guarded in try/catch)
- `loadSession` survival (carries forward restored stream, clears flag after use)
- Per-assignment guardrails in sessions.js (6 sites still call `_updateActiveRunDot`)

webtecnica:
✅ All 5 items from review addressed:
1. setBusy now drives `_updateActiveRunDot()` + `_persistActiveRunState()` (single chokepoint)
2. `_restoreActiveRunState` paints and survives boot
3. `S.sessionId` → `S.session.session_id`
4. Phantom property removed
5. 21 behavioral DOM tests added (56 total passing)

Ready for re-review.

Issue #2168, opened 2026-05-13
perf(session): track measured session-switch latency improvements


franksong2702:
## Problem

Switching between large sessions can still feel slow even though Hermes WebUI already windows the transcript and caches rendered HTML for some idle sessions.

This issue tracks measured, small-slice improvements to session-switch latency. It should not become a broad session renderer rewrite or a standalone instrumentation project.

## Evidence so far

Local 8787 measurement on three real idle sessions showed duplicate transcript post-processing before #2166:

- `highlightCode`: 3 calls per switch
- `addCopyButtons`: 2 calls per switch
- PDF / HTML / Mermaid / KaTeX helpers: 2 calls per switch

After #2166, the same helper set runs once per switch.

Measured wall-clock numbers from the same local 8787 pass:

| Session | Messages | Before first pass | After first pass | Before second pass | After second pass |
|---|---:|---:|---:|---:|---:|
| `20260513_112418_89ca08` | 1793 | 5266.7 ms | 3037.5 ms | 2732.5 ms | 1350.3 ms |
| `20260504_160031_1fd2bc` | 855 | 4198.1 ms | 1668.0 ms | 2215.1 ms | 1298.4 ms |
| `20260503_152835_357902` | 724 | 3251.5 ms | 1182.4 ms | 1116.9 ms | 1249.9 ms |

The wall-clock values include API, filesystem, sidebar, workspace, and browser runtime variance. The stable invariant is the duplicate helper-call reduction. This tracking issue should avoid turning those wall-clock values into fake precision.

### #2166 — consolidate transcript post-render processing

#2166 removes the most obvious duplicate frontend work:

- cached transcript path
- fresh transcript rebuild path
- idle `loadSession()` path no longer runs an extra `highlightCode()`

This is intentionally narrow and does not claim to fix every source of session-switch latency.

### #2170 — skip CLI metadata lookup for native sessions

#2170 addresses the next measured backend bottleneck after #2166.

Local direct-route profiling showed native WebUI session metadata loads were still paying an unnecessary Agent/CLI metadata scan. Native WebUI sessions already have local metadata, so the CLI/messaging fallback path can be skipped for those sessions while preserved for imported CLI and messaging-backed sessions.

Measured local route timing for `GET /api/session?messages=0&resolve_model=0` on `20260513_112418_89ca08` improved from median 662.82 ms to 1.71 ms.

### #2171 — trim bounded tail response overhead

[...]

PR #6178, opened 2026-07-17
perf: make repeated conversation switching responsive

lenard-d:
## Thinking Path

- Conversation switching should reach first paint without waiting for work that is unrelated to the selected transcript.
- Runtime traces showed three repeat costs: rebuilding the initial transcript tail, rebuilding rendered DOM for a recently visited chat, and starting sidebar/model-catalog refreshes during cross-profile navigation.
- The largest tail came from `/api/models?freshness=session_visit`: the measured run made 22 of those requests across 24 chat opens, with a roughly four-second median.
- The fix keeps authoritative session state untouched. It caches only validated presentation responses/markup, preserves an already-complete all-profiles sidebar during its cookie-only profile switch, and moves model-catalog freshness work to the point where the user opens the model picker.

## What Changed

- Add a bounded, lock-guarded server cache for the final redacted initial transcript-tail response of eligible idle native WebUI sessions.
  - Keys include sidecar, `state.db`/WAL/SHM, settings, profile config, pagination, and render-expansion stamps.
  - A second source snapshot prevents storing a response if authoritative inputs changed while it was assembled.
  - Active, pending, CLI/messaging, compression-lineage, recovery, and read-only session shapes bypass the cache.
  - Limits: 24 entries, 24 MiB total, 2 MiB per entry.
- Make the browser transcript HTML cache a bounded LRU (8 entries, 8 MiB total, 2 MiB per entry) so recently visited conversations can restore without repeating Markdown and DOM construction.
- Preserve the cached all-profiles sidebar when an existing chat click only needs to switch the profile cookie; avoid the intermediate skeleton and projects/sessions round trip.
- Skip the duplicate generic model-catalog refresh on that cross-profile path.
- Mark model options stale when a session opens and run the existing `session_visit` freshness check only when the model picker is opened.
- Preserve requested picker freshness across concurrent model loads: a stronger `session_visit` load queues behind an incompatible in-flight load, while duplicate picker calls share the queued promise.
- Document the navigation/cache contract in `ARCHITECTURE.md` and add regression coverage for cache eligibility/invalidation, render-cache bounds/signatures, profile switching, and picker-triggered model refresh.

## Why It Matters

Measured on the same live local WebUI click flow before and after the change:

| Signal | Before | After |
|---|---:|---:|
| Cross-profile switch median | 2,766.7 ms | 121.3 ms |
| Cross-profile switch maximum | 4,139.3 ms | 467.9 ms |
| Transcript-tail median | 123.5 ms | 97.3 ms |
| Transcript-tail p90 | 248.5 ms | 152.9 ms |
| Transcript-tail maximum | 2,345.4 ms | 610.2 ms |
| `session_visit` model requests during chat opens | 22 / 24 opens | 0 / 21 opens |

The measured profile-switch median is about 23x faster, and the recurring four-second model request is no longer part of conversation navigation.

There are no visual or layout changes, so before/after screenshots would be identical. The relevant UI evidence is the request count and latency comparison above; the sidebar remains visually stable during the optimized profile switch.

## Contract Routing

Task type: focused session-navigation performance fix.

Touched areas:

- `GET /api/session` initial transcript-tail presentation response
- browser transcript render cache
- all-profiles sidebar navigation across profile ownership
- model-picker catalog freshness

Relevant public docs:

- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`
- `ARCHITECTURE.md`
- `docs/UIUX-GUIDE.md`
- `DESIGN.md`

State layer and invariant: only bounded presentation caches are added. Session sidecars and `state.db` remain authoritative; volatile/recovery shapes bypass the server cache, source stamps invalidate cached responses, and a cache miss follows the existing full reconciliation/redaction path.

Evidence: automated cache/profile/model-refresh regressions plus a restarted live-instance click trace covering 21 conversation opens and 12 cross-profile switches.

## Contract Change

Previous behavior:

- A loaded session scheduled a `session_visit` model-catalog refresh even when the user did not interact with model selection.
- An existing-chat click across profiles could rebuild the sidebar even when the all-profiles cache already contained the target row.

New behavior:

- Session navigation marks the model catalog stale; opening the model picker performs the bounded freshness check on demand.
- An all-profiles existing-chat click keeps the complete cached sidebar visible across the profile-cookie switch.

Compatibility rationale: session metadata still supplies the active model and `syncTopbar()` injects a missing session-scoped option. The authoritative catalog is refreshed before model-picker interaction, while chat first paint no longer depends on provider discovery or a redundant sidebar request.

## Verification

- `./scripts/test.sh tests/test_issue2613_render_cache_signature.py tests/test_session_detail_tail_cache.py tests/test_profile_switch_ux.py tests/test_issue4756_session_visit_model_refresh.py -q`
  - 48 passed after rebasing onto current `master`
- Ruff on all four changed test files
  - passed
  - passed
- Manual live-instance verification after restart
  - 21 conversation opens
  - 12 cross-profile switches
  - zero `/api/models?freshness=session_visit` calls during chat navigation

## Risks / Follow-ups

- The server cache is per-process and intentionally conservative; CLI/messaging and recovery-sensitive sessions still take the full path.
- The model catalog may remain stale until the picker is opened. The picker already triggers the freshness check, so send/start correctness is unchanged.
- #6038 addresses latency inside the model-catalog refresh itself and remains complementary: this PR removes that refresh from navigation, while #6038 can make the eventual on-demand refresh faster and more resilient.
- No dependencies, build tooling, visual styling, storage formats, or authoritative session-state semantics change.

Release-note-ready wording:

> Conversation switching is faster by reusing validated transcript/render caches, preserving the all-profiles sidebar across chat-owned profile switches, and deferring model-catalog refresh until the picker is opened.

## Model Used

Provider: OpenAI

Model: GPT-5 (Codex agent)

Notable use: AI-assisted code-path and runtime-log tracing, implementation, regression-test updates, upstream rebase/conflict review, and PR preparation through local terminal tooling.

Refs #2168, #6135, and #6038.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -235,6 +235,24 @@ Session is a plain Python class (not a dataclass, not SQLAlchemy):
 title_from(): takes messages list, finds first user message, returns first 64 chars.
 Called after run_conversation() completes to set the session title retroactively.
 
+Initial session-tail response cache:
+
+- `GET /api/session?messages=1&msg_limit=N&resolve_model=0` keeps a bounded
+  process-local LRU of final redacted response payloads for idle native WebUI
+  sessions.
+- The cache is a presentation optimization only. Session JSON plus profile
+  `state.db` remain authoritative; `context_messages` and transcript recovery
+  semantics are unchanged.
+- Keys include stat signatures for the session sidecar, active-profile
+  `state.db`/WAL/SHM files, `settings.json`, and the active profile's
+  `config.yaml`. Any source write produces a miss automatically.
+- Active/pending runs, messaging/CLI/read-only sessions, truncation state, and
+  compression/fork lineage bypass the cache and use the established full
+  reconciliation path.
+- The cache is capped by both entry count and serialized payload bytes. It is
+  intentionally process-local: restart drops it safely, and a miss is always a
+  performance cost rather than a correctness failure.
+
 ### 4.3 SSE Streaming Engine
 
 This is the most architecturally interesting part. Two endpoints cooperate:
@@ -446,10 +464,14 @@ Chat:
     removeThinking()      Removes thinking dots (called on first token or on error)
 
 Rendering:
-    renderMessages()      Full rebuild of #msgInner from S.messages
+    renderMessages()      Reuses a per-session HTML LRU on unchanged cross-session
+                          navigation; otherwise rebuilds #msgInner from S.messages
     renderMd(raw)         Homegrown markdown renderer (see 5.4 for known gaps)
     syncTopbar()          Updates topbar title, meta, model chip, workspace chip
     renderTray()          Updates attach tray showing pending files
+
+The HTML LRU is bounded to eight entries, 2 MiB per entry, and 8 MiB total
+(estimated as UTF-16 browser heap). Cache hits refresh insertion/LRU order.
 
 Approval:
     showApprovalCard(p)   Shows the approval card with command/description text
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -461,8 +461,9 @@ When the all-profiles sidebar opens an existing conversation owned by another
 profile, the profile cookie still switches before `loadSession()`, but the
 already-complete all-profiles list stays visible and is reused from browser
 cache. That path does not synchronously refetch projects/sessions or start the
-generic model-catalog refresh; `loadSession()` performs the single deferred
-session-specific model refresh after transcript first paint.
+generic model-catalog refresh. Session navigation marks the model catalog stale
+without fetching it; opening the model picker performs the bounded
+session-specific freshness check on demand.
 
 Chat:
     send()                Main action: upload files, POST /api/chat/start, open EventSource
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -457,6 +457,13 @@ Session management:
     deleteSession(sid)    POST /api/session/delete, handle active/inactive cases correctly
     renderSessionList()   GET /api/sessions, rebuild #sessionList DOM
 
+When the all-profiles sidebar opens an existing conversation owned by another
+profile, the profile cookie still switches before `loadSession()`, but the
+already-complete all-profiles list stays visible and is reused from browser
+cache. That path does not synchronously refetch projects/sessions or start the
+generic model-catalog refresh; `loadSession()` performs the single deferred
+session-specific model refresh after transcript first paint.
+
 Chat:
     send()                Main action: upload files, POST /api/chat/start, open EventSource
     uploadPendingFiles()  Upload each file in S.pendingFiles, return filenames array
--- a/api/routes.py
+++ b/api/routes.py
@@ -9378,7 +9378,7 @@ def _session_detail_tail_cache_get(key: tuple | None) -> dict | None:
         if entry is None:
             return None
         _SESSION_DETAIL_TAIL_CACHE.move_to_end(key)
-        return entry[0]
+        return dict(entry[0])
 
 
 def _session_detail_tail_cache_set(key: tuple | None, payload: dict) -> None:
--- a/api/routes.py
+++ b/api/routes.py
@@ -28,7 +28,7 @@ import time
 import uuid
 import http.client
 import socket as _socket
-from collections import defaultdict, deque
+from collections import OrderedDict, defaultdict, deque
 from pathlib import Path
 from contextlib import closing
 from urllib.parse import parse_qs, quote, unquote, urljoin, urlsplit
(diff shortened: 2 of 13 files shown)


nesquena-hermes (changes requested):
Thanks for this — the perf work itself reviewed clean (server tail-cache invalidation is robust and cross-profile-safe, the byte-budgeted render-cache LRU refactor preserves the signature comparison, and the deferred model-picker freshness is correctly queued). But **this PR turns CI red**: it refactors several source strings that are asserted verbatim by sibling tests you didn't update, so 4 existing (untouched) tests now fail on the PR head. Reproduced locally in a sandbox against the current head `51c29441`.

### The 4 failing assertions (all are stale source-string matches, not logic bugs)

**1. `static/ui.js`** — you renamed the cache write `_sessionHtmlCache.set(sid,{…})` → `_sessionHtmlCacheSet(sid,{…})`, but:
- `tests/test_issue734_message_windowing.py:41` still asserts the old string:
  ```python
  assert "_sessionHtmlCache.set(sid,{html:_html,msgCount,renderWindowKey,signature:renderSignature})" in UI_JS
  ```
  → change to `_sessionHtmlCacheSet(sid,{html:_html,msgCount,renderWindowKey,signature:renderSignature})`

**2. `static/boot.js`** — you changed the arrow signature `_hydrateModelDropdown=({redirectIfUnauth=null}={})` → `({redirectIfUnauth=null,freshness=null}={})`, but three sibling assertions still hard-code the old signature:
- `tests/test_model_default_boot_precedence.py:182`
- `tests/test_session_metadata_fast_path.py:75`
- `tests/test_session_metadata_fast_path.py:112` (the `needle` var, used by `src.index(needle)` at :113 — this one raises `ValueError: substring not found`)

  → update all three to `const _hydrateModelDropdown=({redirectIfUnauth=null,freshness=null}={})=>populateModelDropdown({`

### Verification

- The 4 test files **you did touch** (`test_session_detail_tail_cache.py`, `test_issue2613_render_cache_signature.py`, `test_issue4756_session_visit_model_refresh.py`, `test_profile_switch_ux.py`) all pass on the head.
- The 3 files above are untouched by this PR and pass on `origin/master` (they assert strings that exist on master and were removed here) — so this is an introduced regression, confirmed by diffing against master.
- `tests/test_5021_boot_model_redirect_budget.py` references the same strings but via `.replace()` (no-op), so it still passes — it is not part of the fix set.

No application-code change is needed — just update those 4 stale assertion strings to match the refactored source and CI should go green. Once that's pushed I'll re-review.

nesquena-hermes:
Re-reviewed `75b2fff9` ("address session switching review feedback") against the four stale assertions I flagged. **All four are now corrected, and I verified each updated string matches the refactored source at HEAD — CI should go green.**

### Verification (each updated assertion vs. the actual source string)

**1. `tests/test_issue734_message_windowing.py:41`** — now asserts `_sessionHtmlCacheSet(sid,{…})`. Confirmed `static/ui.js` contains that exact string (1 match) and the old `_sessionHtmlCache.set(sid,{…})` is gone (0 matches).

**2-4. boot.js signature** — `test_model_default_boot_precedence.py:182`, `test_session_metadata_fast_path.py:75`, and the `needle` at `test_session_metadata_fast_path.py:112` all now use:
```python
"const _hydrateModelDropdown=({redirectIfUnauth=null,freshness=null}={})=>populateModelDropdown({"
```
Confirmed `static/boot.js` contains that new signature (1 match) and the old `({redirectIfUnauth=null}={})` form is gone (0 matches). The `src.index(needle)` at :113 will now resolve instead of raising `ValueError`.

### On `test_5021_boot_model_redirect_budget.py`

It still literally contains the old `({redirectIfUnauth=null}={})` string (lines 251-252), but that's inside a `.replace()` whose match target no longer exists — so it's a no-op, exactly as before. It's harmless here because the test drives the flow through `globalThis._startBootModelDropdown()` (line 385), whose promotion string is unchanged; `_hydrateModelDropdown` stays a closure-local `const` and is reached indirectly. So this file still passes and correctly needs no edit.

### Bonus: the `api/routes.py` defensive copy

Nice addition — `_session_detail_tail_cache_get` now returns `dict(entry[0])` (routes.py:9380) instead of the stored ref, with `test_detail_tail_cache_hit_is_isolated_from_top_level_response_mutation` covering it. One note for the record, not a blocker: it's a **shallow** copy, so nested structures (e.g. `entry[0]["session"]`) are still shared. That's fine on the current hit path because the only consumer (routes.py:12615) returns straight through `j(handler, payload)`, which is read-only serialization (helpers.py:241) — no nested mutation occurs. If a future change ever mutates a nested field of a cache-hit payload before responding, this would need `copy.deepcopy` (or re-derivation). Worth a one-line comment on the copy so that constraint is explicit.

That's the last of the CHANGES_REQUESTED items from my review. LGTM once CI confirms green.

lenard-d:
@nesquena-hermes Thanks for the re-review. The Greptile check is green and both review threads are resolved. Could you please submit a formal approval or dismiss the earlier changes-requested review so the PR review status reflects your LGTM?

nesquena-hermes:

Certified head: `3a7308907d79e917606829f65f8b1c7a78f2f52d`
Integration: `f4fa3e6c777c9ab04a27fa61569da6189088aaf3`

The performance direction is valuable, but this head is not safe to certify.

### Must fix

1. **Make live runtime state authoritative over tail-cache hits.** Before lookup/store, consult the active-run registry and bypass/partition the cache for active or pending sessions. Direct repro: after an idle hit was cached, registry ownership changed without a sidecar stamp; the next request never called `_active_stream_ids()` and returned `active_stream_id:null`.
2. **Force an authoritative full-session resolution after every source-stamp miss.** Do not rely on `_loaded_metadata_only`; Codex reproduced a changed sidecar being stored under a fresh key with stale in-memory transcript content.
3. **Exclude pending journal recovery.** `_session_has_pending_journal_retry` sessions must bypass this presentation cache until recovery settles.
4. **Use `_sessionHtmlCacheDelete(sid)` for every invalidation.** The raw delete in transparent-history reveal leaks `_sessionHtmlCacheBytes`, eventually evicting every future entry.
5. **Preserve open-picker interaction state.** A deferred catalog refresh rebuilds the open picker and clears the user's search/focus. Capture/restore query, focus, and filtering (and skip a no-op rebuild when the catalog is unchanged).
6. **Restore the full-suite contract.** This head adds seven PR-only failures across anchor ownership, virtual-window recovery, model hydration, stale-stream cleanup, and cache-state tests. Update implementation/harnesses until the serial suite has zero PR-only nodes.

Evidence: threat CLEAN; 67 focused tests passed; browser smoke clean; CI 1/1 green; full suite 13,687 passed but 23 failed + 2 errors, with seven PR-only nodes; Codex `SHIP ONLY WITH FIXES`; Fable `SHIP-WITH-UX-FIXES`. The Opus wrapper hung before model output and was terminated; no verdict is claimed from that leg. Direct cache-registry reproduction above is controlling.

**Result: RED for this exact SHA.** Re-gate after a substantive push; unchanged heads will not receive a duplicate certificate.

Issue #3825, opened 2026-06-08
Feature request: OIDC / OAuth2 SSO support for WebUI login


kurobaryo:
### Problem

The WebUI currently supports only three **user-authentication** modes (`api/auth.py`):

1. A single shared password (`HERMES_WEBUI_PASSWORD` env var or Settings)
2. Passkeys / WebAuthn (opt-in, behind the `HERMES_WEBUI_PASSKEY` flag)
3. No auth (default)

There is no way to delegate WebUI login to an external identity provider.

> Note for anyone reading the code: `api/oauth.py` exists, but it is **provider-credential** OAuth — the Codex/Anthropic device flows used during onboarding to link the agent to a model backend. It is *not* user-login SSO, so it's easy to mistake for OIDC support. The actual user auth lives in `api/auth.py` and has no IdP integration.

### Desired solution

Standard OIDC / OAuth2 **Authorization Code flow (+ PKCE)** for WebUI login, so the WebUI can sit behind a self-hosted IdP (Logto, Keycloak, Authentik, Authelia, Auth0, etc.).

Config (env / config.yaml) would need:
- issuer / discovery URL
- client_id, client_secret
- redirect URI
- scopes
- an allow-list claim (email / sub / group) controlling *who* may log in

Implementation could be relatively light: add `/api/auth/oidc/start` + `/api/auth/oidc/callback` routes, run the Auth Code flow, and on success call the existing `create_session()` in `api/auth.py`. The session/cookie/CSRF machinery there is already clean and provider-agnostic, so it should slot in without touching the rest of the auth model.

### Alternatives considered

Fronting the WebUI with an OIDC reverse proxy (Cloudflare Access, `caddy-security`, oauth2-proxy). This works today and is what I'm currently using, but it:
- adds an extra moving part / proxy to operate
- can't map IdP identity to the WebUI's own session model (the WebUI still has to be configured "no auth" or share-password behind the proxy, which is awkward)

Native support would let the WebUI own the session directly and enforce per-user identity.

### Use case / context

Multi-user / family / small-team self-hosted deployments where a single shared password doesn't fit, and centralized identity (with revocation + MFA enforced at the IdP) is preferred. Pairs naturally with the existing passkey support as a second factor / fallback.

Thanks for the project!

nesquena-hermes:
Thanks for the unusually well-researched writeup — the code citations are accurate, including the easy-to-miss distinction you flagged. I read both files on `origin/master` to confirm before responding.

[...]

PR #6179, opened 2026-07-17
fix: add OIDC_SSRF_ALLOW_PRIVATE env var to bypass SSRF check for self-hosted OIDC (#6136)

webtecnica:
## Summary

Adds `OIDC_SSRF_ALLOW_PRIVATE` environment variable as an opt-out for the OIDC SSRF private-IP check (Option A from the issue).

## Problem

Issue #6136 — The OIDC SSRF hardening (RFC1918 check) blocks legitimate self-hosted deployments where Authentik, Keycloak, or Zitadel run on internal/private networks.

When the hostname resolves to a private IP (e.g. `10.5.5.7`), the SSRF check rejects it:
```
OIDC SSRF check rejected auth.example.com because it resolved to private address 10.5.5.7
```

## Solution

Sets `OIDC_SSRF_ALLOW_PRIVATE=true` (or `1`, `yes`) to skip the private-IP check in `_validate_outbound_oidc_url()`.

- Keeps the current protection **by default** — no behavioral change for existing deployments
- Logs a `logger.warning(...)` when the bypass is active, so operators are aware
- Accepts `"true"`, `"1"`, or `"yes"` (case-insensitive)

## Usage

```bash
OIDC_SSRF_ALLOW_PRIVATE=true ./bootstrap.py
```

Or set in your environment / `.env` file:
```
OIDC_SSRF_ALLOW_PRIVATE=true
```

## Changes

Modified one file:
- `api/auth_oidc.py` — added env var check before `_is_disallowed_oidc_host()` call

## Verification

- [x] Python parse verified: `ast.parse(open('api/auth_oidc.py').read())`
- [x] No merge markers or truncation
- [x] Diff is minimal (+8 lines, single responsibility)

### DIFF
--- a/api/auth_oidc.py
+++ b/api/auth_oidc.py
@@ -400,6 +400,14 @@ def _validate_outbound_oidc_url(url: str) -> None:
     hostname = str(parsed.hostname or "").strip()
     if not hostname:
         raise OIDCAuthError("OIDC endpoint URL was missing a hostname", status_code=502)
+    _ssrf_bypass = os.getenv("OIDC_SSRF_ALLOW_PRIVATE", "").strip().lower()
+    if _ssrf_bypass in ("1", "true", "yes"):
+        logger.warning(
+            "OIDC_SSRF_ALLOW_PRIVATE is enabled — SSRF protection bypassed for %s. "
+            "This reduces security; only use on trusted networks.",
+            url,
+        )
+        return
     if _is_disallowed_oidc_host(hostname):
         raise OIDCAuthError(
             "OIDC endpoint URLs must not target private or local addresses",


nesquena-hermes:
Reading the diff here against `api/auth_oidc.py` on `origin/master`, and comparing it side-by-side with your other PR for the same issue (#6164), I think this env-var version has a security problem that #6164 doesn't — worth flagging before either merges.

## The issue: this bypass is placed too early and drops *all* IP-class checks

Your change inserts the `OIDC_SSRF_ALLOW_PRIVATE` early-return *before* the `_is_disallowed_oidc_host()` call:

```python
    _ssrf_bypass = os.getenv("OIDC_SSRF_ALLOW_PRIVATE", "").strip().lower()
    if _ssrf_bypass in ("1", "true", "yes"):
        logger.warning(...)
        return                       # <-- skips every check below
    if _is_disallowed_oidc_host(hostname):
        raise OIDCAuthError(...)
```

Because it `return`s before `_is_disallowed_oidc_host`, it doesn't just relax the *private* (RFC1918) rule the issue asked for — it also skips **loopback, link-local, multicast, unspecified, and reserved**. On `origin/master`, `_is_disallowed_oidc_ip` (`api/auth_oidc.py:433-442`) rejects all six classes:

```python
    return (
        candidate.is_loopback
        or candidate.is_private
        or candidate.is_link_local
        or candidate.is_multicast
        or candidate.is_unspecified
        or candidate.is_reserved
    )
```

The most concrete consequence: with the flag on, `169.254.169.254` (link-local cloud-metadata endpoint) becomes a valid OIDC issuer/JWKS/token target. That's exactly the SSRF-to-metadata escalation the #3825 hardening was defending against, and it's now reachable via a single env var rather than being narrowly scoped to "private LAN IdP."

## Your #6164 does this correctly

The sibling PR threads `allow_private` down into `_is_disallowed_oidc_ip` and keeps loopback (and everything except `is_private`) unconditionally blocked:

```python
    if candidate.is_loopback:
        return True
    if not allow_private and candidate.is_private:
        return True
    return (
        candidate.is_link_local
        or candidate.is_multicast
        or candidate.is_unspecified
        or candidate.is_reserved
    )
```

That's the safe shape: relax *only* RFC1918, leave link-local/loopback/etc. blocked. It also ships 7 tests (including `test_still_rejects_loopback_when_allow_private_endpoints_is_true`) covering exactly this boundary.

## Recommendation

I'd close this PR in favor of #6164. Two of the three approaches were on the table in #6136, and the maintainer's note there leaned toward the tightest option — a hostname allow-list (`private_endpoint_hosts` / a `_ssrf_trusted_hosts`-style set) over a blanket boolean, since it scopes the relaxation to the operator's actual IdP host rather than "any private/link-local address." #6164 is already closer to that spirit than the env-var blanket bypass, and if you want to go one step further, extending it to a host allow-list would match the guidance in the issue thread most precisely.

If you'd rather keep an env var as the config source, that's fine — but it should feed the same `allow_private` parameter path #6164 uses (relax `is_private` only), not an early `return` that skips the whole guard.

webtecnica:
Fechando em favor do #6164, como recomendado — a abordagem de env-var com early return desprotege loopback/metadata além do RFC1918. O #6164 já implementa o caminho correto (relaxa só `is_private`, mantém loopback/link-local bloqueados). Obrigado pela review detalhada, @nesquena-hermes! 🙏

PR #6180, opened 2026-07-17
fix: catch unhandled exception in POST /api/updates/check (defensive hardening)

### LINKED ISSUE #6086: WebUI server silently exits on every /api/updates/check and on gateway restart when a stale gateway_state.json is present (fail-closed recovery)

webtecnica:
## Summary

Defensive hardening: wraps `POST /api/updates/check` handler with `try/except Exception` + `logger.exception()` as a **belt-and-suspenders** measure. The handler was the only one in `handle_post` without exception protection.

## What it does

- Wraps `check_for_updates()` in a `try/except Exception` block
- Logs the full traceback via `logger.exception()` if something goes wrong
- Returns HTTP 500 instead of crashing

## Does NOT fix #6086

Per maintainer analysis in review, the real root cause of #6086 is likely **signal/process-group reaping** (same class as #6149), not a handler exception:

- Reporter's logs show `200` responses — the handler already returned successfully
- Death is silent — no traceback, no log output
- Also triggered on gateway restart with stale `gateway_state.json`

This PR is purely defensive hardening — ensuring every route handler has consistent exception protection with traceback logging. #6086 remains open for the underlying signal/subprocess root cause.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -13811,7 +13811,14 @@ def handle_post(handler, parsed) -> bool:
             channel = settings.get("update_channel")
         from api.updates import check_for_updates
 
-        return j(handler, check_for_updates(force=force, include_agent=include_agent_updates, channel=channel))
+        logger.info("checking for updates (force=%s, include_agent=%s, channel=%s)", force, include_agent_updates, channel)
+        try:
+            payload = check_for_updates(force=force, include_agent=include_agent_updates, channel=channel)
+        except Exception:
+            logger.exception("update check failed unexpectedly — server would have crashed silently")
+            return bad(handler, "Update check failed, see server log for details", status=500)
+        logger.info("update check completed")
+        return j(handler, payload)
 
     if parsed.path == "/api/extensions/toggle":
         from api.extensions import ExtensionToggleError, set_extension_user_enabled
--- a/api/routes.py
+++ b/api/routes.py
@@ -13812,10 +13812,13 @@ def handle_post(handler, parsed) -> bool:
         from api.updates import check_for_updates
 
         logger.info("checking for updates (force=%s, include_agent=%s, channel=%s)", force, include_agent_updates, channel)
+        # Defensive-only guard: wrap check_for_updates() for consistent
+        # exception protection across all route handlers. Does NOT fix #6086
+        # (root cause is likely signal/process-group reaping, per maintainer analysis).
         try:
             payload = check_for_updates(force=force, include_agent=include_agent_updates, channel=channel)
         except Exception:
-            logger.exception("update check failed unexpectedly — server would have crashed silently")
+            logger.exception("update check failed unexpectedly (defensive guard caught exception)")
             return bad(handler, "Update check failed, see server log for details", status=500)
         logger.info("update check completed")
         return j(handler, payload)


nesquena-hermes:
Thanks for taking this on — hardening the update-check path with `try/except` + `logger.exception()` is a reasonable defensive change on its own merits. But I don't think the root-cause theory in the PR body matches how the server actually dispatches requests, and I want to flag that before this gets treated as *the* fix for #6086.

## The premise doesn't hold: `handle_post` is already wrapped

The PR says an unhandled exception from `check_for_updates()` "propagated by `handle_post` and killed the server process silently." But every POST already runs inside a top-level `try/except Exception` in `server.py`. `do_POST` delegates to `_handle_write` (`server.py:428` → `server.py:400`), which wraps the route call:

```python
    try:
        ...
        result = route_func(self, parsed)   # route_func == handle_post
        ...
    except _CLIENT_DISCONNECT_ERRORS:
        return
    except Exception:
        self._safe_webui_print(f'[webui] ERROR {self.command} {self.path}\n' + traceback.format_exc())
        try:
            j(self, {'error': 'Internal server error'}, status=500)
        ...
```

So an exception bubbling out of `check_for_updates()` at `api/routes.py:13814` would be caught here, logged with a traceback, and turned into a 500 — the request thread dies, not the process. And because the server is `ThreadingHTTPServer` (`server.py`), an uncaught exception in a worker thread can't take down the accept loop anyway.

## The report's own logs contradict a handler crash

In #6086 the lethal requests return **`200`** before the server dies:

```
[webui] 08:54:50Z POST /api/updates/check  200  3892ms   → server dead, watchdog restart
```

A `200` means the handler completed and `j(...)` already serialized a response. An exception inside `check_for_updates()` would have produced a `500` (via the wrapper above) or no status line at all — not a clean `200` followed by process death seconds later. The maintainer reached the same read in the issue thread: a read-only `GET /api/gateway/status` also triggers the exit, which rules out the update handler specifically and points at something *after* the response, not an unhandled exception in the route body.

## Where the actual kill likely lives

The silent exit that #6086 describes (whole process tree gone, no traceback, fires after a 200 and after gateway restarts) looks like a fail-closed recovery path sending a signal to its own PID, not a route exception. `api/agent_health.py` only *classifies* stale state (`_runtime_status_is_stale_stopped` / `_runtime_status_is_stale_running` at `agent_health.py:102`/`:137`) — it has no process-termination call — so the classification isn't the killer either. The thing that actually signals the process is `_handle_shutdown` in `api/routes.py:11675`:

```python
    def _do_shutdown():
        time.sleep(0.3)
        os.kill(os.getpid(), signal.SIGINT)
```

That's the shape that matches the symptom (async signal to own PID → clean-ish exit → no route traceback). The real fix probably belongs wherever the fail-closed gateway-restart recovery decides to invoke a shutdown path against a *stale* `gateway_state.json`, per the issue's suspected-cause section (the #6054 fail-closed change).

## Suggestion

I'd keep this PR as a small standalone robustness patch (the `try/except` around `check_for_updates()` is still worth having), but reword the description so it isn't presented as the #6086 root cause — otherwise the actual fail-closed-shutdown bug gets marked resolved while the process still dies on `GET /api/gateway/status`. Happy to be wrong if you can attach the traceback the new `logger.exception()` captures on an actual exit — if it fires, that would directly disprove my read and I'd defer to the log.

nesquena-hermes (changes requested):
## The guard is harmless, but it doesn't fix #6086 — the evidence points elsewhere

Thanks @webtecnica — wrapping `check_for_updates()` so a genuine exception returns a 500 instead of propagating is a reasonable defensive touch. But I don't think it fixes #6086, and merging it as "Fixes #6086" would close a real server-death bug that would still be live.

**Why the diagnosis doesn't match the report:**
- The reporter's logs show `POST /api/updates/check` returning **`200` (3892ms, 3624ms, …) and THEN the server dies seconds later.** A 200 means the handler already returned successfully — an exception never propagated, so a try/except around `check_for_updates()` has nothing to catch on that path.
- The death is **silent — no traceback, no log output** — and also happens **~3s after an external gateway restart** (not an update check at all), when a **5-day-old stale `gateway_state.json`** is present. An unhandled Python exception in a request handler would leave a traceback in the log; this leaves nothing.
- `check_for_updates()` shells out to `git` via `subprocess.run` (api/updates.py:216). The signature here — 200-then-silent-death, also on gateway restart, no traceback — points at **process-group / signal reaping** (a child process or the gateway restart delivering a signal that reaps the WebUI server's process group), i.e. the same class as the SIGCHLD/process-lifecycle work in #6149. That's the layer the fix needs to address, not a handler try/except.

**Suggested direction (root cause):** reproduce with a stale `gateway_state.json` in HERMES_HOME + trigger `/api/updates/check` (or a gateway restart) and capture what signal/exit kills the process (e.g. `strace -f`/`py-spy dump`, or a `SIGTERM`/`SIGCHLD` handler log). The fail-closed recovery on a stale gateway state (the issue's second trigger) is likely the shared root. Happy to pair on the repro.

The try/except itself is fine to keep as belt-and-suspenders, but please (a) drop the "Fixes #6086" linkage (or re-scope it to the actual signal/subprocess root cause), and (b) note it's defensive-only. Leaving #6086 open. Credit preserved.

webtecnica:
Updated PR body — removed 'Fixes #6086' linkage per review. Keeping the try/except as defense-in-depth. Real root cause per maintainer analysis is likely signal/process-group reaping.

webtecnica:
Updated per maintainer feedback:

- **Dropped "Fixes #6086" linkage** — no longer asserts this as the root cause
- **Added inline comment** above the try/except marking it as a defensive-only guard, explicitly noting it does NOT fix #6086
- **Reworded log message** from "server would have crashed silently" → "defensive guard caught exception" to reflect defense-in-depth posture
- **#6086 remains open** for the underlying signal/process-group reaping issue

The try/except wrapper itself is unchanged — this is purely a labeling/commenting update.

webtecnica:
✅ "Fixes #6086" linkage dropped, PR re-scoped as purely defensive guard. The stale `CHANGES_REQUESTED` is from the prior review — ready for re-review.

nesquena-hermes:
Shipped in **exp-v0.52.117** (experimental channel) via release #6332. Thanks @webtecnica! 🎉

`POST /api/updates/check` now wraps `check_for_updates()` in a defensive guard that returns a clean `500` (and logs) instead of letting an unexpected exception escape the handler — giving it the same exception protection as its sibling routes. This is defensive hardening only; it does not change the success path and correctly does **not** claim to fix #6086 (that root cause is elsewhere, per the earlier review — #6086 stays open).

Gate: full pytest suite (13443 passed / 0 failed), Codex regression gate `SAFE TO SHIP`, ruff clean.

PR #6184, opened 2026-07-17
fix: stop false "Compressing context" on non-compress turns

carlotestor:
## Summary

Brand-new / low-token chats (especially on Anthropic Fable) were sometimes showing the live **✦ Compressing context** worklog divider even when Hermes never compacted history.

Two false-positive paths:

1. **`api/streaming.py` agent status bridge** — any lifecycle message containing broad substrings (`'compressing'`, `'preflight compression'`, …) became SSE `compressing` → UI divider. Skip/cooldown style notices and other non-start chatter could light the banner.
2. **`static/messages.js` snapshot hydration** — `_sourceEventTypeForSnapshotAnchorRow` mapped *all* `lifecycle_status` rows to `compressing`, and mapped `terminal_status` + `running` to `compressing`, inventing a compress divider when rehydrating activity scenes without a real compress SSE event.

### Fix

- Add `_is_agent_compression_start_status()` with **positive** markers matching real hermes-agent emitters (preflight / pre-API / compacting context / context too large / compression attempt) and explicit **rejects** for skip / cooldown / defer / post-compress “Compressed N → M” chatter.
- Use that helper in `_agent_status_callback` instead of the broad substring OR-chain.
- Hydration: terminal never invents compress; lifecycle only becomes `compressing`/`compressed` with positive text/phase cues, else `''`.

No WebUI service restart is required for this PR to land; operators can pick up the static/API change on their normal reload schedule.

## Test plan

- [x] `tests/test_auto_compression_card.py` — 67 passed (includes new unit tests for matcher true/false cases and hydration source assertions)
- [ ] Manual: open a **new** Fable/Anthropic chat with a large continuation prompt; confirm first tool batch does **not** show ✦ Compressing context unless `agent.log` has `Pre-API compression` / `context compression started`
- [ ] Manual: long session that truly hits threshold still shows the divider, then **Context auto-compressed**

### DIFF
--- a/api/streaming.py
+++ b/api/streaming.py
@@ -542,6 +542,48 @@ def _is_fallback_lifecycle_message(kind: str, message: str) -> bool:
     )
 
 
+def _is_agent_compression_start_status(kind: str, message: str) -> bool:
+    """Return True only for real Hermes context-compression start notices.
+
+    WebUI bridges matching lifecycle statuses into an SSE ``compressing`` event
+    and paints the live "Compressing context" worklog divider. The previous
+    matcher used broad substrings such as ``'compressing' in message`` and
+    ``'preflight compression' in message``, which can false-positive on skip /
+    cooldown / unrelated notices and make brand-new low-token turns look like
+    auto-compression.
+
+    Positive markers below match the agent emitters in hermes-agent
+    (``turn_context`` preflight, ``conversation_loop`` pre-API / 413 / too-large,
+    ``conversation_compression`` compaction status). Explicitly reject skip /
+    defer notices so "Skipping preflight compression…" never surfaces as a
+    running compress divider.
+    """
+    k = str(kind or '').strip().lower()
+    m = str(message or '').strip().lower()
+    if k != 'lifecycle' or not m:
+        return False
+    # Skip / cooldown / defer logs must never look like a live compression start.
+    if (
+        'skipping' in m
+        or 'defer' in m
+        or 'cooldown' in m
+        or 'will not start' in m
+    ):
+        return False
+    # Post-compress retry chatter is not a start event.
+    if 'compressed' in m and 'compressing' not in m and 'compression attempt' not in m:
+        return False
+    return (
+        'preflight compression:' in m
+        or 'pre-api compression:' in m
+        or 'compacting context' in m
+        or 'context too large' in m
+        or '— compressing (' in m
+        or '- compressing (' in m
+        or 'compression attempt' in m
+    )
+
+
 def _prewarm_skill_tool_modules():
     """Import tools.skills_tool and tools.skill_manager_tool outside any lock.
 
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
## Summary

Reading the diff at `e210b5f99` against `origin/master` (`api/streaming.py`, `static/messages.js`, `tests/test_auto_compression_card.py`), this correctly narrows two paths that could paint a live "Compressing context" divider on turns where no compaction happened. I cross-checked every positive marker against the actual hermes-agent emitters, and they line up.

## Cross-repo verification (the markers match real emitters)

The new `_is_agent_compression_start_status()` matcher must track the agent's status strings, so I grepped the agent side. Each positive marker maps to a real `_emit_status`/`_buffer_status` call:

- `'preflight compression:'` -> `agent/turn_context.py:487` -- "Preflight compression: ~{n} tokens >= {t} threshold. This may take a moment."
- `'pre-api compression:'` -> `agent/conversation_loop.py:1062` -- "Pre-API compression: ~{n} tokens near the context/output limit. Compacting before the next model call."
- `'compacting context'` -> `agent/conversation_compression.py:50-52`, `COMPACTION_STATUS_MARKER = "Compacting context"`
- `'context too large'` / `'compressing ('` -> `agent/conversation_loop.py:3707` -- "Context too large (~{n} tokens) - compressing ({a}/{max})..."
- `'compression attempt'` -> `agent/conversation_loop.py:3466` -- "Request payload too large (413) - compression attempt {a}/{max}..."

The rejects are also correct against real skip/cooldown/post-compress logs:

```python
if ('skipping' in m or 'defer' in m or 'cooldown' in m or 'will not start' in m):
    return False
if 'compressed' in m and 'compressing' not in m and 'compression attempt' not in m:
    return False
```

`agent/turn_context.py:459/467/474` all emit "Skipping preflight compression: ..." and `agent/conversation_loop.py:3487/3728` emit "Compressed {a} -> {b} messages, retrying..." -- both now correctly excluded. I couldn't find a start-of-compaction emitter that this matcher misses; the five above are the complete set that reach the SSE bridge as lifecycle status.

## Hydration path is safe

For `static/messages.js:3917` `_sourceEventTypeForSnapshotAnchorRow`, the concern with returning `''` for unknown lifecycle/terminal rows is whether the caller handles the empty value. It does -- the call site at `static/messages.js:3973` guards with:

```js
const sourceType=_sourceEventTypeForSnapshotAnchorRow(row);
if(!sourceType) continue;
```

So a running terminal row or a cue-less lifecycle row is now skipped instead of registering a phantom compress anchor, which is exactly the intended behavior. The old `terminal running ? 'compressing' : 'done'` and the blanket `lifecycle -> 'compressing'` were the two invention paths, and both are gone.

## Recommendation

Approve. The matcher is comprehensive against the current agent emitters, the rejects cover the known false-positive log lines, and the hydration change degrades safely through the existing `if(!sourceType) continue;` guard. The `test_agent_compression_start_status_matches_real_emitters_only` cases mirror the exact emitter strings, so future drift in the agent status text will surface as a test signal here.

One forward-looking note: because the matcher is now string-coupled to the agent's human-readable status text, a wording change on the agent side (e.g. dropping the em-dash in "- compressing (") would silently stop lighting the divider. The hyphen fallback already hedges the dash variant, which is a nice touch. If the agent ever grows a structured `phase`/`kind` on these lifecycle events, migrating to that would be more durable than substring matching -- but that's a follow-up, not a blocker.

nesquena-hermes:
Shipped in **exp-v0.52.101**. Thanks @carlotestor! 🎉 The gate confirmed the narrowed `_is_agent_compression_start_status()` still matches every current Hermes-agent compression-start emitter (no real card lost) and the only dropped row shape is a synthetic terminal_status+running no producer emits. Clean fix.

PR #6185, opened 2026-07-17
fix(wakeup): route background completions by origin_ui_session_id

carlotestor:
## Summary

Fixes **Background wakeup** events landing in the wrong Hermes WebUI chat.

Root cause: server-side `notify_on_complete` routing trusted the legacy `session_key → session_id` index. Concurrent turns / temporary subagent keys can contaminate that key. Modern hermes-agent stamps an exact `origin_ui_session_id` (commissioning browser tab); this PR makes that field authoritative.

### Changes
- Bind `HERMES_UI_SESSION_ID` in `_set_turn_session_identity` / reset (alongside approval + `_SESSION_KEY`)
- `_resolve_completion_target`: prefer `origin_ui_session_id` over session-key resolution
- Recover `origin_ui_session_id` / `spawn_session_id` from `ProcessSession` when the event lacks it
- Keep legacy session-key fallback for older agent builds without exact owner metadata
- Option-3 env-immune owner attrs now include `origin_ui_session_id` first

### Companion PR
Agent-side stamp: opened in parallel against `NousResearch/hermes-agent` (`fix/webui-bg-wakeup-origin-routing`).

Related: #6002, #2968.

## Test plan
- [x] `PYTHONPATH=<hermes-agent> pytest tests/test_xsession_wakeup_misroute.py` (11 passed)
- [ ] Manual: two open WebUI sessions; spawn `notify_on_complete` process in A; confirm server-side wakeup only in A

### DIFF
--- a/api/background_process.py
+++ b/api/background_process.py
@@ -822,7 +822,12 @@ def _mark_registry_completion_consumed(process_id: str) -> None:
 # registry, pre-Option-1 spawns) it is a PURE PASS-THROUGH — it never
 # suppresses a legitimate Option Z wakeup on uncertainty (Option Z must keep
 # working).
-_ENV_IMMUNE_OWNER_ATTRS = ("spawn_session_id", "owner_session_id", "turn_session_id")
+_ENV_IMMUNE_OWNER_ATTRS = (
+    "origin_ui_session_id",  # modern hermes-agent exact browser-tab return address
+    "spawn_session_id",
+    "owner_session_id",
+    "turn_session_id",
+)
 
 
 def _env_immune_spawn_owner(proc_session) -> str:
@@ -872,6 +877,32 @@ def _resolve_wakeup_target(
     return owner
 
 
+def _resolve_completion_target(
+    *,
+    session_key_resolved_sid: str,
+    origin_ui_session_id: str,
+) -> str:
+    """Return the WebUI session that owns a detached completion event.
+
+    Modern Hermes Agent events carry ``origin_ui_session_id`` as an exact,
+    immutable return address captured from the commissioning browser turn.
+    It is authoritative over the mutable/legacy session-key index. Older
+    Agent events omit it and retain the existing session-key fallback.
+    """
+    resolved = str(session_key_resolved_sid or "")
+    owner = str(origin_ui_session_id or "")
+    if not owner:
+        return resolved
+    if resolved and resolved != owner:
+        logger.error(
+            "cross-session completion route BLOCKED: session_key resolved to %r "
+            "but exact origin_ui_session_id is %r; routing to the exact owner",
+            resolved,
+            owner,
+        )
+    return owner
+
+
 def _process_one(evt: dict) -> None:
     """Route a single completion_queue event to the matching WebUI session."""
     from api import config as _cfg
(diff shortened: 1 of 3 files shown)


nesquena-hermes (changes requested):
## The origin-based routing is right, but it re-delivers on restart without durable-claim acknowledgment

Thanks @carlotestor — routing detached completions by the immutable `origin_ui_session_id` over the mutable session-key index is the correct fix for the misroute (and the producer contract is present on the installed agent, so `_resolve_completion_target` reads a real field). But the adversarial gate found a CORE the new path exposes:

**`api/background_process.py:959` — an async-delegation completion is re-delivered after every WebUI restart.** `_process_one()` schedules the owner wakeup but never acknowledges the durable `delegation_id`, so `delivery_state` stays `'pending'`. On the next server start, the durable-completion restore re-enqueues it, and the new origin-only path schedules it **again** → duplicate agent turns and repeated tool actions. Verified against an isolated `state.db`.

**Fix-spec:** wire the durable-claim lifecycle around the delivery:
- `claim_event_delivery(delegation_id)` before scheduling (atomic claim so only one delivery proceeds),
- `complete_event_delivery(delegation_id)` after a successful schedule / durable deferral,
- `release_event_delivery(delegation_id)` on failure so it can retry.
- Add a regression: deliver a completion, simulate a restart (re-run the durable restore), assert it is NOT scheduled twice.

**Heads-up on the cluster (for @nesquena — dedup decision):** this overlaps two other PRs on the same routing surface — #6002 (CONFLICTING + changes-requested, the original) and #6159 (sysophelper-droid, +1583, which already carries the durable-claims contract). The cleanest end state is likely **#6185's origin-based routing + #6159's durable-claims lifecycle combined** into one PR, with the other two closed. I can help reconcile once you pick the direction; for now #6185 needs the claims lifecycle to be shippable on its own. Credit preserved.

nesquena-hermes:
## Summary

Read the diff at HEAD vs `origin/master` for `api/background_process.py:822-983` and `api/streaming.py` (the `_set/_reset_turn_session_identity` changes), then cross-referenced the agent side because the whole fix hinges on an agent contract (`origin_ui_session_id`). The WebUI logic is sound and the passthrough is safe. But the effective-on-day-one behavior depends entirely on the companion agent PR, and I want to make the layer boundary explicit so this doesn't merge with a false sense that it's already routing.

## The routing logic (WebUI side) is correct

`_resolve_completion_target` is a clean, conservative preference:

```python
resolved = str(session_key_resolved_sid or "")
owner = str(origin_ui_session_id or "")
if not owner:
    return resolved
if resolved and resolved != owner:
    logger.error("cross-session completion route BLOCKED: ...")
return owner
```

No owner → pure passthrough; owner present → authoritative. That's the right shape, and layering it *after* `_resolve_wakeup_target` (background_process.py:943) keeps the existing Option-3 spawn-owner net intact.

## Cross-repo check: this is inert until the agent PR lands

The WebUI `_process_registry` is the agent singleton — `from tools.process_registry import process_registry as _process_registry` (background_process.py:886). So `evt`, and the `_ps` recovered from it, come straight from the agent's `ProcessRegistry`. On the currently pinned agent:

- The `notify_on_complete` completion event is enqueued in `_move_to_finished` (agent `tools/process_registry.py:1092`) and stamps only `session_id`, `session_key`, `command`, `exit_code`, `completion_reason`, `termination_source`, `output`, `started_at`. **No `origin_ui_session_id`.**
- The `ProcessSession` dataclass (agent `tools/process_registry.py:91-135`) has **no `origin_ui_session_id` and no `spawn_session_id` field** — `register()` / `spawn_via_env()` capture only `session_key`.

So on the process-completion path this PR targets, all three sources are empty:

```python
origin_ui_session_id = str(evt.get("origin_ui_session_id") or "")           # "" — not in enqueue
# recovery:
str(getattr(_ps, "origin_ui_session_id", "") or "")                          # "" — not a field
or str(getattr(_ps, "spawn_session_id", "") or "")                           # "" — not a field
```

`_resolve_completion_target` therefore always returns `resolved` unchanged — a correct no-op. The misroute this fixes (#6002/#2968) only actually stops happening once the agent side adds the field to `ProcessSession` and stamps it in the `_move_to_finished` enqueue at `tools/process_registry.py:1092`. That's the companion PR the description mentions; worth confirming the two land together (or that this is explicitly staged to ship inert first), otherwise a reviewer testing against today's agent will see no behavior change and conclude the fix doesn't work.

Note the async-delegation path is different: agent `tools/async_delegation.py` already carries `origin_ui_session_id` end-to-end, but those completions don't flow through `_process_one`'s `completion_queue`, so this PR doesn't affect them.

## Reinforcing Greptile's alert-fatigue point

Greptile's observation about `_ENV_IMMUNE_OWNER_ATTRS` is real and becomes live exactly when the agent PR lands. Once `ProcessSession.origin_ui_session_id` exists, `_env_immune_spawn_owner` (now returning it first) feeds `_resolve_wakeup_target` at background_process.py:943 — which runs *before* `_resolve_completion_target`. If `session_key` didn't resolve (index miss → `resolved=""`) but the proc carries an `origin_ui_session_id`, the `owner != resolved` branch fires the `"xsession wakeup misroute BLOCKED ... a per-turn session-identity capture race occurred upstream"` ERROR even though nothing raced — the key was simply absent. Suggest gating that ERROR on `resolved` being non-empty (a true positive mismatch), and downgrading the empty-`resolved` case to `debug`, so the new authoritative-owner path doesn't generate spurious race alerts.

## Test plan

`tests/test_xsession_wakeup_misroute.py` (via `PYTHONPATH=<hermes-agent> pytest`, 11 passed) covers `_resolve_completion_target` preference and the contextvar lifecycle well. Two gaps worth pinning given the analysis above: (1) a test asserting inert passthrough when both `evt` and `_ps` lack the field (today's agent), and (2) the `resolved=""` + owner-present case, to lock in whatever the ERROR-vs-debug decision ends up being.

nesquena-hermes:
I've combined this with #6159 into **#6283**, per the cluster-reconciliation note in the review above. #6283 keeps your origin-based routing (`_resolve_completion_target` / `origin_ui_session_id` authority) as the destination resolver and wraps it in #6159's durable claim/complete/release lifecycle — which supplies the `complete_event_delivery` ack that fixes the restart-redelivery issue this PR was bounced for. The gate also caught (and #6283 fixes) that the same origin authority needed to apply in the next-turn drain, not just the background path.

Your commit is preserved via `Co-authored-by` on #6283. Keeping this open until #6283 merges, then I'll close it with credit. Thanks @carlotestor.

nesquena-hermes:
Shipped in **exp-v0.52.114** via #6283, which combines your origin-based routing with #6159's durable-claim lifecycle. Your commit is preserved (Co-authored-by). Thanks @carlotestor — the origin_ui_session_id return-address approach is the core of the fix. 🎉

PR #6186, opened 2026-07-17
Live Stream: recover interrupted Thinking after restart

franksong2702:
## Thinking Path

#3929's live error path already snapshots visible text, Thinking/reasoning, and tool calls, and the run journal durably records all three. The restart/reconciliation path restored text and tool cards but intentionally ignored `reasoning` events. It also treated a reasoning-only journal as having no visible output, so a pending user owner could be omitted during populated-core recovery.

The first regression failed before the implementation because both recovered assistant rows had `reasoning=None`. A second regression showed that a reasoning-only journal with a populated core transcript did not preserve the pending user turn that owned the visible work.

## What Changed

- Count non-empty journaled `reasoning` events as visible recovery output.
- Rebuild journaled Thinking as display-only `message.reasoning` alongside the matching recovered assistant segment.
- Strip restored reasoning before every `context_messages` projection; provider sanitization remains a second boundary because `reasoning` is not API-safe.
- Make replay idempotent for reasoning-only and mixed text/reasoning segments.
- Scope content-based reasoning backfill to the current pending user turn and allow each journal segment to claim at most one pre-existing assistant row per replay.
- Preserve legitimate repeated Thinking/process segments around tool boundaries.

No UI layout or interaction code changes; the existing settled Worklog renderer consumes the restored `message.reasoning` field.

## Why It Matters

A user who saw Thinking before a WebUI process restart should not lose that visible work after reload. At the same time, recovered display metadata must not become model-facing history. This patch closes that restart-only gap without changing live streaming, provider payloads, or the existing final-answer contract.

## Contract Routing

Task type: Runtime replay/recovery bug fix that enforces existing contracts.

Touched areas:
- durable run-journal replay
- visible transcript projection in `session.messages`
- model-context exclusion in `context_messages`
- interrupted-turn ownership and deduplication

Relevant public docs:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/CONTRACTS.md`

State layers:
- Source evidence: durable run-journal `reasoning` / `token` / tool events.
- Display truth written: `session.messages[*].reasoning`.
- Model context: deliberately unchanged; restored reasoning is removed before `context_messages` append and excluded from provider sanitization.

Scope boundaries:
- No live SSE behavior changes.
- No cancellation semantics changes.
- No provider error-classification or wakeup-coalescing changes.
- No contract change; this implements the existing replay/idempotency requirements.

- restart recovery preserves text, Thinking, tools, owner, and terminal marker ordering;
- repeated replay does not duplicate rows;
- identical segments remain distinct when they represent distinct journal positions;
- restored Thinking never enters model-facing history.

## Verification

- `370 passed` across the focused restart-repair, run-journal, Anchor, state.db reconciliation, gateway history, reasoning replay, and #3929 regression gate.
- `python3 scripts/ruff_lint.py --diff origin/master` - 0 findings on added/modified lines.
- `git diff --check origin/master` - passed.
- Independent Claude Code review with `claude-sonnet-5`, tool access disabled - final verdict: **No blocking issues**.
- UI evidence: not applicable to layout/interaction; this is a persisted-state recovery fix exercised through deterministic process-restart journal fixtures.

## Risks / Follow-ups

The patch deliberately keeps the older content-dedup fallback for text-only recovery. The new current-turn ownership gate applies when attaching restored reasoning, where cross-turn misattribution would create new user-visible behavior.

The broader #3929 umbrella still includes separate provider/wakeup and incident-forensics facets; this PR addresses only display-only Thinking recovery after restart.

## Release Note

Live Stream: preserve already-visible Thinking after a WebUI restart while keeping recovered reasoning out of model-facing conversation history.

## Model Used

- Provider: OpenAI
- Model: Codex (GPT-5; exact runtime model ID is not exposed)
- Notable use: repository investigation, implementation, deterministic regression design, and final self-review
- Independent reviewer: Anthropic Claude Code 2.1.212, `claude-sonnet-5`, high-effort, tools disabled

AI-assisted implementation and review.

Refs #3929
Refs #3400

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -849,7 +849,10 @@ def _recovered_model_context_projection(message: dict) -> dict | None:
     if projected.get('_error'):
         return None
     if _content_has_reasoning_only_parts(projected.get('content')):
-        return None
+        if projected.get('tool_calls'):
+            projected['content'] = ''
+        else:
+            return None
     projected_text = _normalize_journal_recovery_text(projected.get('content'))
     if not projected_text and not projected.get('tool_call_id') and not projected.get('tool_calls'):
         return None
--- a/api/models.py
+++ b/api/models.py
@@ -2306,11 +2306,23 @@ def _collapse_adjacent_duplicate_partials(messages) -> tuple[list, bool]:
     return collapsed, changed
 
 
-def _find_existing_assistant_for_journal_content(session, content: str) -> int | None:
+def _find_existing_assistant_for_journal_content(
+    session,
+    content: str,
+    *,
+    max_index: int | None = None,
+    excluded_indexes: set[int] | None = None,
+) -> int | None:
     candidate = _normalize_journal_recovery_text(content)
     if not candidate:
         return None
-    for idx, message in enumerate(session.messages or []):
+    messages = session.messages or []
+    stop = len(messages) if max_index is None else min(len(messages), max_index)
+    substring_match = None
+    for idx in range(stop):
+        if excluded_indexes and idx in excluded_indexes:
+            continue
+        message = messages[idx]
         if not isinstance(message, dict) or message.get('role') != 'assistant':
             continue
         if message.get('_error'):
@@ -2320,9 +2332,9 @@ def _find_existing_assistant_for_journal_content(session, content: str) -> int |
             continue
         if existing == candidate:
             return idx
-        if len(candidate) >= 24 and candidate in existing:
-            return idx
-    return None
+        if substring_match is None and len(candidate) >= 24 and candidate in existing:
+            substring_match = idx
+    return substring_match
 
 
 def _journal_tool_already_present(
@@ -2394,6 +2406,12 @@ def _run_journal_has_visible_output(session, stream_id: str | None) -> bool:
                 continue
             if str(payload.get('text') or '').strip():
                 return True
+        if event_name == 'reasoning':
+            reasoning_text = str(
+                payload.get('text') or payload.get('reasoning') or payload.get('thinking') or ''
+            )
+            if reasoning_text.strip():
+                return True
         if event_name == 'tool':
             return True
     return False
@@ -2454,9 +2472,10 @@ def _append_journaled_partial_output(
     """Recover already-emitted visible output from a dead stream journal.
 
     This repair path is intentionally conservative: it restores user-visible
-    assistant text and tool-card metadata that had already been emitted over
-    SSE before the WebUI process died. It does not restore hidden reasoning and
-    it does not try to continue execution.
+    assistant text, display-only reasoning, and tool-card metadata that had
+    already been emitted over SSE before the WebUI process died. Restored
+    reasoning stays out of ``context_messages`` so it cannot become provider-
+    facing history. The repair does not try to continue execution.
     """
     if not stream_id:
         return False
@@ -2479,24 +2498,106 @@ def _append_journaled_partial_output(
 
     appended_any = False
     assistant_parts: list[str] = []
+    reasoning_parts: list[str] = []
     assistant_started_at: float | None = None
     current_assistant_idx: int | None = None
     recovered_tool_calls: list[dict] = []
+    initial_message_count = len(session.messages or [])
+    claimed_existing_assistant_indexes: set[int] = set()
+
+    def content_match_can_receive_reasoning(existing_idx: int) -> bool:
+        messages = session.messages or []
+        owner_idx = None
+        for candidate_idx in range(existing_idx - 1, -1, -1):
+            candidate = messages[candidate_idx]
+            if isinstance(candidate, dict) and candidate.get('role') == 'user':
+                owner_idx = candidate_idx
+                break
+        if owner_idx is None:
+            return False
+
+        pending_text = _normalize_journal_recovery_text(session.pending_user_message)
+        owner_text = _normalize_journal_recovery_text(messages[owner_idx].get('content'))
+        if pending_text and owner_text != pending_text:
+            return False
+
+        for candidate_idx in range(existing_idx + 1, initial_message_count):
+            candidate = messages[candidate_idx]
+            if not isinstance(candidate, dict) or candidate.get('role') != 'user':
+                continue
+            candidate_text = _normalize_journal_recovery_text(candidate.get('content'))
+            if pending_text and candidate.get('_recovered') and candidate_text == pending_text:
+                continue
+            return False
+        return True
+
+    def append_context_projection(message: dict) -> None:
+        context_projection = dict(message)
+        context_projection.pop('reasoning', None)
+        _append_recovered_turn_to_context(session, context_projection)
+
+    def attach_display_reasoning(message: dict, reasoning: str) -> bool:
+        if not reasoning:
+            return False
+        existing = str(message.get('reasoning') or '').strip()
+        if existing:
+            return False
+        message['reasoning'] = reasoning
+        return True
 
     def flush_assistant() -> int | None:
-        nonlocal appended_any, assistant_parts, assistant_started_at, current_assistant_idx
+        nonlocal appended_any, assistant_parts, reasoning_parts
+        nonlocal assistant_started_at, current_assistant_idx
         content = ''.join(assistant_parts).strip()
+        reasoning = ''.join(reasoning_parts).strip()
         assistant_parts = []
-        if not content:
+        reasoning_parts = []
+        if not content and not reasoning:
             return current_assistant_idx
-        if dedupe_existing:
-            existing_idx = _find_existing_assistant_for_journal_content(session, content)
+        if dedupe_existing and content:
+            search_excluded = set(claimed_existing_assistant_indexes)
+            existing_idx = None
+            while True:
+                candidate_idx = _find_existing_assistant_for_journal_content(
+                    session,
+                    content,
+                    max_index=initial_message_count,
+                    excluded_indexes=search_excluded,
+                )
+                if candidate_idx is None:
+                    break
+                if not reasoning or content_match_can_receive_reasoning(candidate_idx):
+                    existing_idx = candidate_idx
+                    break
+                search_excluded.add(candidate_idx)
             if existing_idx is not None:
+                claimed_existing_assistant_indexes.add(existing_idx)
                 current_assistant_idx = existing_idx
                 assistant_started_at = None
                 if 0 <= existing_idx < len(session.messages):
-                    _append_recovered_turn_to_context(session, session.messages[existing_idx])
+                    existing_message = session.messages[existing_idx]
+                    append_context_projection(existing_message)
+                    if attach_display_reasoning(existing_message, reasoning):
+                        appended_any = True
                 return existing_idx
+        if dedupe_existing and reasoning and not content:
+            for existing_idx in range(initial_message_count):
+                if existing_idx in claimed_existing_assistant_indexes:
+                    continue
+                existing_message = session.messages[existing_idx]
+                if not isinstance(existing_message, dict):
+                    continue
+                if (
+                    existing_message.get('_recovered_from_run_journal')
+                    and existing_message.get('_recovered_stream_id') == stream_id
+                    and existing_message.get('role') == 'assistant'
+                    and not str(existing_message.get('content') or '').strip()
+                    and str(existing_message.get('reasoning') or '').strip() == reasoning
+                ):
+                    claimed_existing_assistant_indexes.add(existing_idx)
+                    current_assistant_idx = existing_idx
+                    assistant_started_at = None
+                    return existing_idx
         timestamp = int(assistant_started_at or time.time())
         recovered_assistant = {
             'role': 'assistant',
@@ -2505,8 +2606,9 @@ def _append_journaled_partial_output(
             '_recovered_from_run_journal': True,
             '_recovered_stream_id': stream_id,
         }
+        attach_display_reasoning(recovered_assistant, reasoning)
         session.messages.append(recovered_assistant)
-        _append_recovered_turn_to_context(session, recovered_assistant)
+        append_context_projection(recovered_assistant)
         current_assistant_idx = len(session.messages) - 1
         assistant_started_at = None
         appended_any = True
@@ -2559,6 +2661,16 @@ def _append_journaled_partial_output(
         event_name = str(event.get('event') or event.get('type') or '')
         payload = event.get('payload') if isinstance(event.get('payload'), dict) else {}
         created_at = event.get('created_at') if isinstance(event.get('created_at'), (int, float)) else None
+        if event_name == 'reasoning':
+            text = str(
+                payload.get('text') or payload.get('reasoning') or payload.get('thinking') or ''
+            )
+            if not text:
+                continue
+            if not assistant_parts and not reasoning_parts and assistant_started_at is None:
+                assistant_started_at = created_at or time.time()
+            reasoning_parts.append(text)
+            continue
         if event_name == 'token':
             text = str(payload.get('text') or '')
             if not text:
--- a/api/models.py
+++ b/api/models.py
@@ -819,13 +819,25 @@ def _active_stream_ids():
     return active_ids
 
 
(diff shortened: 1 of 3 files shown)


franksong2702:
Thanks for flagging the reasoning-only anchor case. I traced the exact retry-growth sequence before changing the guard.

The run journal is append-only and replayed from the beginning. When a journal first contains reasoning and later grows a tool event, the retry processes the reasoning event first; `flush_assistant()` deduplicates it back to the existing reasoning row and returns that row as the tool anchor. `ensure_assistant_anchor()` is not the selector in that path. Excluding reasoning-only rows from the guard would also conflict with the existing segment contract: a tool immediately following reasoning belongs to that assistant segment.

I added `test_retry_growth_attaches_later_tool_to_reasoning_segment` at head `8b2377469`. It performs two recovery passes with the journal growing from reasoning-only to reasoning+tool and proves there is one reasoning row and the later tool remains attached to its index. The focused cross-layer gate is now `370 passed`; no product-code change was needed for this comment.

nesquena-hermes (changes requested):
## Strong recovery work, but two SILENT edges break the "reasoning never reaches model history" boundary

Thanks @franksong2702 — restoring journaled Thinking as display-only recovery is genuinely valuable and the main path (strip reasoning before the model-context projection, idempotent replay, pending-owner preservation) is well-built. Full suite is green. But the adversarial gate (verified with isolated restart/next-turn probes) found two SILENT edges — both on the exact invariant this PR is careful about, so worth one more round:

1. **`api/models.py:2534` — the reasoning-strip boundary is bypassed when `context_messages` is empty.** `_append_recovered_turn_to_context()` returns early in that case, so the fallback projects from `session.messages` **with `reasoning` still attached** → recovered display reasoning enters the next model-context projection AND gets re-persisted into `context_messages`. Provider sanitization removes it from the actual request, but the PR's stated *independent* strip boundary is broken here (defense-in-depth reduced to one layer). **Fix:** in the empty-context recovery path, materialize model context from existing messages with `reasoning` removed and reasoning-only assistant rows omitted, THEN append recovered content through that projection. Add an empty-context next-turn/writeback regression asserting no `reasoning` key survives into `context_messages`.

2. **`api/models.py:2528` — a repeated identical prompt attaches current-turn reasoning to a PRIOR assistant row.** The owner matcher accepts an earlier same-text user turn as the owner and ignores the later recovered same-text owner. Verified with two `"Continue"` turns: the prior assistant was mutated and no recovered assistant followed the current owner. **Fix:** validate the preceding owner with the full pending-checkpoint identity (timestamp + source + attachments), not text alone; reject ambiguous candidates and append after the CURRENT owner. Add a repeated-prompt fixture.

Net invariant: recovered `reasoning` must never survive into any model-facing projection on ANY path (incl. empty-context), and reasoning must attach to the exact current-turn owner even when the prompt text repeats. Close these two and it's ready for a final gate. Credit preserved.

franksong2702:
Pushed follow-up commit `0ce9f17d9` addressing both requested recovery-boundary cases.

What changed:
- Empty `context_messages`: `_append_recovered_turn_to_context()` now seeds model-facing context from a sanitized recovery projection when context is empty. That projection strips `reasoning`, skips reasoning-only assistant rows, and skips error markers before later new-turn context can fall back to display messages.
- Repeated same-text pending prompt: reasoning backfill now requires the preceding owner user row to match the full pending checkpoint identity (`content` + timestamp + source + attachments) before it can claim an existing assistant row. Same-text ambiguous prior turns are rejected, so recovered reasoning appends after the current owner instead of mutating a prior `Continue` turn.

Regression coverage added:
- `test_empty_context_recovery_seeds_reasoning_free_model_context`
- `test_repeated_pending_prompt_uses_checkpoint_owner_not_prior_same_text`

Local verification:
- `./scripts/test.sh tests/test_issue3929_partial_work_recovery.py -k 'repeated_pending_prompt or empty_context_recovery'` — 2 passed after failing before the fix
- `./scripts/test.sh tests/test_issue3929_partial_work_recovery.py` — 11 passed
- `./scripts/test.sh tests/test_session_sidecar_repair.py` — 67 passed
- `./scripts/test.sh tests/test_issue3929_partial_work_recovery.py tests/test_session_sidecar_repair.py` — 78 passed
- `python3 scripts/ruff_lint.py --diff origin/master` — 0 findings on added/modified lines
- `git diff --check origin/master` — passed

GitHub checks restarted on the pushed head and are pending.

nesquena-hermes (changes requested):
## Round 2: both prior findings closed, but the empty-context rework introduced two new SILENT edges

Thanks @franksong2702 — the two round-1 findings are resolved: the empty-context path now routes through the shared strip projection, and the owner match uses full checkpoint identity. Full suite green. But the re-gate (reproduced with disk round-trips) found the new reconstruction introduced two SILENT edges:

1. **`api/models.py:829` — a structured reasoning-only `content` ARRAY is stringified as visible content and persisted into `context_messages`.** The strip handles the `message.reasoning` FIELD, but a row whose `content` is `[{"type":"reasoning","reasoning":"..."}]` gets stringified into visible content and survives into model context. Reproduced surviving a disk round-trip. **Fix:** use content-aware reasoning-only detection — a row whose content array is entirely reasoning parts must be omitted before projection (not stringified).

2. **`api/models.py:865` — the empty-context reconstruction dedups LEGITIMATE repeated assistant replies.** Two identical assistant replies collapse to one, which then creates adjacent user turns → model history no longer matches the transcript. Reproduced. **Fix:** append EVERY projected historical row in order; reserve dedup for the newly-recovered row only (don't dedup pre-existing history).

Net invariant: recovered reasoning must be omitted from model context in BOTH forms (the `reasoning` field AND a reasoning-only `content` array), and the empty-context reconstruction must preserve the transcript's exact row sequence (no collapsing legitimate duplicates). Two focused fixes and it's there — the core recovery behavior is solid now. Credit preserved.

franksong2702:
Pushed follow-up commit `ef61bf77` addressing the Round 2 empty-context reconstruction findings.

Changes:
- Omit structured reasoning-only `content` arrays (`reasoning` / `thinking` parts with no visible text) from recovered model-context projection, so they do not get stringified into `context_messages` or survive a disk round-trip.
- Preserve every projected historical transcript row when seeding empty `context_messages`; dedupe now only applies when appending the newly recovered row.

Added regressions:
- `test_empty_context_recovery_omits_structured_reasoning_only_content`
- `test_empty_context_recovery_preserves_duplicate_historical_replies`

Local verification:
- `./scripts/test.sh tests/test_issue3929_partial_work_recovery.py tests/test_session_sidecar_repair.py` => 80 passed
- `python3 scripts/ruff_lint.py --diff origin/master` => OK
- `git diff --check origin/master` => OK

Readback after push: PR head is `ef61bf77bbe8bf58ce5e4ebb5b94d3db6c3b8b24`; GitHub checks have restarted and are pending.

nesquena-hermes (changes requested):
## Round 3: down to one edge — the content-array omission drops tool-call-bearing rows

Thanks @franksong2702 — round-2's two SILENT gaps are both closed now (reasoning-only content arrays are no longer stringified into model context, and repeated replies are preserved in order). One edge remains, introduced by the new content-array omission:

**`api/models.py:851` — an assistant message with reasoning-only content that ALSO owns `tool_calls` is dropped entirely.** The omission correctly removes pure reasoning-only rows, but when such a row carries `tool_calls`, dropping it orphans the following `tool` result, which the provider sanitizer then removes too — so the model history collapses from `user → assistant(tool_calls) → tool` down to `user → assistant`, silently losing a complete valid tool-call/result pair. Verified against the actual recovery + sanitizer paths.

**Fix:** for reasoning-only content WITH non-empty `tool_calls`, strip the reasoning *content* but RETAIN the assistant/tool-call boundary (keep the `tool_calls` so the following `tool` result stays anchored). Only omit the row when it has no protocol-bearing tool metadata (genuinely reasoning-only, no tool_calls). Add a save/reload regression covering a complete `assistant(tool_calls) → tool` pair surviving empty-context recovery intact.

That's the last one — the recovery behavior is otherwise solid across all the prior rounds. Credit preserved.

franksong2702:
Pushed follow-up commit `29a237ac` for the Round 3 tool-call anchor edge.

Change:
- When recovered empty-context projection sees reasoning-only structured `content` on an assistant row that owns non-empty `tool_calls`, it now strips the reasoning content to `""` but keeps the assistant/tool-call boundary. Pure reasoning-only rows with no tool metadata are still omitted.

Added regression:
- `test_empty_context_recovery_preserves_tool_calls_with_reasoning_only_content` covers save/reload plus provider sanitization preserving `user → assistant(tool_calls) → tool → assistant`.

Local verification:
- `./scripts/test.sh tests/test_issue3929_partial_work_recovery.py tests/test_session_sidecar_repair.py` => 81 passed
- `./scripts/test.sh tests/test_issue3583_orphaned_tool_calls.py tests/test_orphaned_tool_messages.py` => 22 passed
- `python3 scripts/ruff_lint.py --diff origin/master` => OK
- `git diff --check origin/master` => OK

Readback after push: PR head is `29a237ac067fee538ee9109caabf2286fafbc501`; GitHub checks have restarted and are pending.

franksong2702:
Refreshed the PR branch onto latest `origin/master` after GitHub reported the branch was behind.

Rebased head:
- old: `29a237ac067fee538ee9109caabf2286fafbc501`
- new: `e0dee876a07d4d274f47d608f4495bad2506db6b`
- base: `920cdd56c0b429eaeda76fb6fbba0417b913ae07`

No conflict changes were needed during rebase.

Local verification after rebase:
- `./scripts/test.sh tests/test_issue3929_partial_work_recovery.py tests/test_session_sidecar_repair.py` => 81 passed
- `./scripts/test.sh tests/test_issue3583_orphaned_tool_calls.py tests/test_orphaned_tool_messages.py` => 22 passed
- `python3 scripts/ruff_lint.py --diff origin/master` => OK
- `git diff --check origin/master` => OK

Readback after push: PR head is `e0dee876a07d4d274f47d608f4495bad2506db6b`; GitHub checks restarted and are pending.

nesquena-hermes:
## Reviewed — Greptile's `ensure_assistant_anchor` guard gap is real; one-line fix is safe

Pulled `e0dee876` into a read-only worktree and read the full recovery loop in `api/models.py` — `flush_assistant` (2622-2688), the reasoning-only dedup path (2657-2672), `ensure_assistant_anchor` (2691-2731), and `attach_display_reasoning` (2613-2619) — plus the 770-line regression suite. The reasoning-recovery direction is well-built and the context-exclusion contract (`append_context_projection` popping `reasoning` at 2611, backed by `test_empty_context_recovery_seeds_reasoning_free_model_context`) holds. Confirming Greptile's one flagged gap by code reading, and adding a note on why the fix is safe against the existing tests.

## The gap

`ensure_assistant_anchor`'s backward "reuse a free empty anchor" scan tests content emptiness only:

```python
# api/models.py:2715-2719
if (
    _m.get('_recovered_from_run_journal')
    and _m.get('_recovered_stream_id') == stream_id
    and _m.get('role') == 'assistant'
    and not str(_m.get('content') or '').strip()
):
```

Before this PR, a recovered row with empty content was always a true bare anchor. After this PR that's no longer true: the reasoning-only flush at 2673-2688 appends `{'role':'assistant','content':'','_recovered_from_run_journal':True,'_recovered_stream_id':stream_id}` and then `attach_display_reasoning` sets a top-level `reasoning` field (2618). So a reasoning-only row has **empty content but non-empty reasoning** — and it satisfies all four guard conditions. Because the scan walks `range(len-1, -1, -1)` (newest first), on a replay of a `tool_A → reasoning_1 → tool_B` journal it can return the later reasoning-only row M_R before the true bare anchor M0, pulling a tool onto the Thinking segment and leaving M0 orphaned (empty content, empty reasoning, no tools → renders blank).

## The fix Greptile suggests is correct and low-risk

Adding one clause makes the anchor selection precise:

```python
    and not str(_m.get('content') or '').strip()
    and not str(_m.get('reasoning') or '').strip()   # skip reasoning-only rows
):
```

Crucially this does **not** regress `test_retry_growth_attaches_later_tool_to_reasoning_segment` (line 618), which is the test that legitimately attaches a later tool to a reasoning segment. I traced why: in that test, when the `tool` event arrives, `reasoning_1` is still buffered, so the tool handler's `anchor_idx = flush_assistant()` (2769) routes through the **reasoning-only dedup path** at 2657-2672 — which finds and claims the existing M_R and returns its index. `anchor_idx` is therefore non-None and `ensure_assistant_anchor` is never called for that tool. The reasoning-attach behavior lives entirely in `flush_assistant`, not in the anchor scan, so tightening the anchor guard leaves it intact. That's the key reason the one-liner is safe.

## Test gap

The suite is thorough (idempotent reasoning replay at 588, identical-segments-around-tool at 660, cross-turn scoping at 235), but none exercises Greptile's exact shape: a journal with a **distinct bare anchor M0 preceding a reasoning-only row**, replayed after growth so the tool re-enters `ensure_assistant_anchor` (rather than the `flush_assistant` reasoning path). The existing line-618 test can't catch the regression because there M_R *is* the intended anchor. I'd add:

```
def test_tool_before_reasoning_keeps_own_anchor_on_regrowth():
    # journal: tool_A (bare anchor M0) -> reasoning_1 (M_R) -> [regrow] -> tool_B
    # assert tool_A stays on M0 (empty content AND empty reasoning),
    #        tool_B lands on M_R, and no empty/empty anchor is orphaned.
```

That pins the guard so a future refactor can't silently re-widen it.

## Verdict

Solid PR. The context-exclusion boundary (display `reasoning` never entering `context_messages` / provider history) is the important contract here and it's correctly enforced at two layers. The only thing I'd want before merge is the two-character guard clause at 2719 plus the distinct-anchor regression test above — both cheap, and together they close the one path where recovered Thinking metadata could steal a tool's anchor. Everything else reads as safe.

franksong2702:
Pushed follow-up commit `5c34dd9` for the `ensure_assistant_anchor` guard gap.

Change:
- Tightened the recovered empty-anchor reuse scan so it skips rows with non-empty display `reasoning`. Bare tool anchors remain reusable; reasoning-only Thinking rows can no longer steal a later tool anchor.

Added regression:
- `test_tool_before_reasoning_keeps_own_anchor_on_regrowth` reproduced the failure first: the regrown tool anchored to the reasoning row at index 2 instead of the bare anchor at index 1. It now passes with the guard clause.

Local verification:
- `./scripts/test.sh tests/test_issue3929_partial_work_recovery.py tests/test_session_sidecar_repair.py` => 82 passed
- `./scripts/test.sh tests/test_issue3583_orphaned_tool_calls.py tests/test_orphaned_tool_messages.py` => 22 passed
- `python3 scripts/ruff_lint.py --diff origin/master` => OK
- `git diff --check origin/master` => OK

Readback after push: PR head is `5c34dd924841d5e5106e5b2543baef298f28b93a`; GitHub checks have restarted and are pending.

nesquena-hermes (approved):
Approving: round-4 closes the last edge — a reasoning-only-content assistant row that owns tool_calls now retains the tool-call boundary, so `user → assistant(tool_calls) → tool` survives empty-context recovery intact (verified round-trip), while pure reasoning-only rows stay omitted from model context. Codex SAFE, full suite green, CI 18/18. This closes a 4-round convergence on the restart reasoning-recovery path. Shipping.

nesquena-hermes:
🚀 Shipped in **exp-v0.52.96** (experimental). Thanks @franksong2702 — Thinking/reasoning that was visible before a WebUI process restart is now recovered from the run journal on reload (display-only `message.reasoning`), so you don't lose that work. The hard part was keeping recovered reasoning strictly display-only: it's stripped from every model-context projection (both the `reasoning` field AND reasoning-only content arrays), the empty-context reconstruction preserves the exact transcript order (no collapsed duplicates), owner matching uses full checkpoint identity, and a reasoning-only row that owns `tool_calls` retains its tool-call boundary so `assistant(tool_calls) → tool` pairs survive intact. This took four gate rounds to get every data-integrity edge — final gate: Codex SAFE, full suite green, CI 18/18. No live-streaming/provider-payload changes. Live on the dogfood box.

Issue #5749, opened 2026-07-07
bug(stream): Transparent Stream sporadically renders duplicate response prefix


qxxaa:
## Summary

In hermes-webui, a completed assistant turn can visually show part of the assistant response twice when Transparent Stream rendering is enabled.

The persisted session does not contain two final assistant messages. Instead, the duplicate appears to come from the settled anchor activity scene: a `prose` activity row stores a prefix of the final assistant answer, and the renderer then displays that prose row before also displaying the final assistant segment.

The result is a single assistant response appearing to contain duplicated content, even though the canonical assistant message exists only once.

The issue has been observed sporadically rather than on every turn.

## Environment

- hermes-webui running as a container
- Browser UI session
- Transparent Stream rendering enabled
- WebUI version: `v0.51.904`
- Agent version: `0.18.0`
- Model: `gpt-5.5`
- Transport: `/responses`

## User-visible symptom

After a turn completes, the assistant response appears with the beginning portion repeated:

1. a prefix of the final answer is rendered as chronological/process prose
2. the full final answer is rendered normally below it

This makes the response look like the assistant generated duplicate text inside a single reply.

## Technical evidence

Inspection of the affected WebUI session JSON showed the following structure:

- The canonical `messages` array contained one final assistant message for the turn.
- The final assistant message content length was approximately 3.9k characters.
- `anchor_activity_scenes` contained a settled scene for that same assistant message.
- The scene's `final_answer` matched the canonical assistant message.
- The same scene also contained an `activity_rows` entry with:
  - `role: "prose"`
  - `kind: "process_prose"`
  - `source_event_type: "token"`
  - a `local_id` shaped like `live-prose:<stream_id>:<segment_seq>`
  - text equal to a prefix of the final answer
- That prose-row prefix was substantially shorter than the final answer, roughly 40-45% of the normalized final-answer length.
- The prose row was therefore not filtered as a duplicate final-answer row and remained visible in Transparent Stream rendering.

The relevant rendered data shape is therefore approximately:

```json
{
  "messages": [
    {
      "role": "assistant",
      "content": "<full final answer>"
    }
  ],
  "anchor_activity_scenes": {
    "<message-ref>": {
      "scene": {
        "final_answer": "<full final answer>",
        "activity_rows": [
          {
            "role": "prose",
            "kind": "process_prose",
            "source_event_type": "token",
            "local_id": "live-prose:<stream-id>:<segment-seq>",
            "text": "<prefix of full final answer>"
[...]

PR #6189, opened 2026-07-17
fix(transparent-stream): drop stale final-answer prefix row in multi-segment settle (#5749 follow-up)

ai-ag2026:
## Summary

A turn with interim assistant messages (prose narration interleaved with tool calls) can show the beginning of the final answer **twice** after watching it stream in Transparent Stream mode: once as a settled anchor-scene prose row (the live-token accumulator's last throttled snapshot, `live-prose:<stream>:N`) and once as the real assistant segment. The duplicate persists until the page is reloaded. This is the residual multi-segment case of #5749, which #5758 fixed for single-segment turns.

## Root cause

#5758 suppresses the live-token accumulator row at settlement only when it sits *after the last tool row*. But `_completeSettledAnchorSceneForTurn` builds `orderedRows` as `[projected live rows] + [settled per-message rows] + [terminal rows]` — and the settled per-message tool rows re-list tools that ran **earlier** in the turn. In a multi-segment turn they land after the final segment's accumulator, pushing the "last tool row" boundary past it, so the eligibility gate never fires. The stale prefix snapshot (~2–6% of the final answer, far below the 0.9 near-match ratio of `_anchorSceneRowLooksLikeFinalAnswer`) then survives into the persisted scene and renders above the settled answer.

Fresh reloads mask the bug because `_anchorSceneRowsForRendering` merges tool-started/completed pairs by logical key, which moves the boundary back before the accumulator — which is why the duplicate only appears on the live→settle path.

## Fix

Judge final-segment eligibility against the **live projection's own chronology**: a live-prose token row belongs to the final segment iff no *projected* tool row follows it (tracked via a `WeakSet` built from `projectedRows` before the settled rows are appended). Pre-tool narration that happens to prefix the final answer keeps its #5758 protection — the existing regression tests (`test_issue5749_settlement_preserves_pre_tool_live_token_prefix_rows` and friends) still pass unchanged.

## Validation

- `./scripts/test.sh tests/test_issue5749_transparent_stream_prefix_dedupe.py tests/test_5749_anchor_scene_client_prefix_dedupe.py tests/test_live_to_final_anchor_visible_order.py tests/test_issue5367_transparent_live_row_reconcile.py` — 81 passed (2 failed before the fix on the buggy behavior side).
- Full anchor/transparent/worklog sweep (28 files) — 380 passed.
- New regression test `test_issue5749_multi_segment_settled_tool_rows_do_not_shield_final_accumulator` pins the multi-segment shape: accumulator dropped, pre-tool narration and tool rows preserved.
- End-to-end: replayed the captured run journal of an affected real session (6 `interim_assistant` segments, 18 tool calls) through the real SSE live handlers in headless Chromium via CDP. Before: answer prefix visible twice after settle, persisting indefinitely. After: visible exactly once; scene no longer contains the stale row; fresh-reload rendering unchanged.

## Notes

- The duplicated row also got **persisted** into `anchor_activity_scenes` by the pre-fix code; such legacy scenes still render clean on reload via the existing merge-reordering path, so no migration is needed.

### DIFF
--- a/static/messages.js
+++ b/static/messages.js
@@ -3552,14 +3552,28 @@ function attachLiveStream(activeSid, streamId, uploaded=[], options={}){
     for(const row of projectedRows){
       if(row&&row.role==='terminal') orderedRows.push(row);
     }
-    const lastNonTerminalWorkRowIndex=orderedRows.reduce((last,row,idx)=>(row&&row.role==='tool')?idx:last,-1);
+    // #5758 gap: final-segment eligibility must be judged against the LIVE
+    // projection's own chronology. The settled per-message tool rows appended
+    // into orderedRows above re-list tools that ran EARLIER in the turn, so an
+    // index over the combined list pushes the "after the last tool row"
+    // boundary past the final segment's live-prose accumulator — its stale
+    // prefix snapshot then survives into the persisted scene and renders as a
+    // duplicate of the answer's beginning. A live-prose row belongs to the
+    // final segment iff no PROJECTED tool row follows it; pre-tool narration
+    // that happens to prefix the final answer stays protected.
+    const lastProjectedToolIndex=projectedRows.reduce((last,row,idx)=>(row&&row.role==='tool')?idx:last,-1);
+    const finalSegmentLiveProseRows=new WeakSet();
+    projectedRows.forEach((row,idx)=>{
+      if(idx>lastProjectedToolIndex&&row&&row.role==='prose'&&row.kind==='process_prose'&&String(row.source_event_type||'')==='token'&&String(row.local_id||'').startsWith('live-prose:')) finalSegmentLiveProseRows.add(row);
+    });
     const rowIsLiveTokenFinalPrefix=(row,textKey,finalSegmentEligible)=>finalSegmentEligible&&row&&row.role==='prose'&&row.kind==='process_prose'&&String(row.source_event_type||'')==='token'&&String(row.local_id||'').startsWith('live-prose:')&&textKey&&finalKey&&textKey.length<finalKey.length&&finalKey.startsWith(textKey);
     const pushRow=(row,rowIndex)=>{
       if(!row||typeof row!=='object') return;
+      const finalSegmentEligible=finalSegmentLiveProseRows.has(row);
       row=_anchorSceneSettleLiveRunningRow(row,hasSettledThinking);
       if(!row||typeof row!=='object') return;
       const textKey=_anchorSceneTextKey(row.text);
-      if(rowIsLiveTokenFinalPrefix(row,textKey,rowIndex>lastNonTerminalWorkRowIndex)) return;
+      if(rowIsLiveTokenFinalPrefix(row,textKey,finalSegmentEligible)) return;
       const isTextual=row.role==='prose'||row.role==='thinking';
       if(isTextual&&_anchorSceneRowLooksLikeFinalAnswer(textKey,finalKey)) return;
       if(isTextual&&_anchorSceneRowTextOverlapsExisting(textKey,seenTextKeys)) return;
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
## Summary

Read the diff at HEAD vs `origin/master` for `static/messages.js:3552-3577` (the `attachLiveStream` settle projection), the full `pushRow`/`orderedRows` construction around it, and the new regression test. The root-cause writeup is accurate and the fix is surgical. Confirming the identity mechanics below because they're the load-bearing part of this change and easy to break in a later refactor.

## The fix is correct, and the ordering inside `pushRow` matters

The bug is exactly as described: `orderedRows` is `[projected live rows] + [settled per-message rows] + [projected terminal rows]` (messages.js:3541-3551), and the settled per-message rows re-list earlier tools, so `lastNonTerminalWorkRowIndex` over the *combined* list pushed the "after last tool" boundary past the final segment's accumulator. Judging eligibility off `projectedRows` chronology instead is the right call:

```js
const lastProjectedToolIndex=projectedRows.reduce((last,row,idx)=>(row&&row.role==='tool')?idx:last,-1);
const finalSegmentLiveProseRows=new WeakSet();
projectedRows.forEach((row,idx)=>{
  if(idx>lastProjectedToolIndex&&row&&row.role==='prose'&&row.kind==='process_prose'&&String(row.source_event_type||'')==='token'&&String(row.local_id||'').startsWith('live-prose:')) finalSegmentLiveProseRows.add(row);
});
```

The important subtlety a reviewer should not miss: the WeakSet is keyed by **object identity**, and it works only because `orderedRows.push(row)` stores the same references that came out of `projectedRows` (no spread/clone at messages.js:3543-3545). Critically, the eligibility check runs *before* the row is reassigned:

```js
const pushRow=(row,rowIndex)=>{
  if(!row||typeof row!=='object') return;
  const finalSegmentEligible=finalSegmentLiveProseRows.has(row);   // original ref — correct
  row=_anchorSceneSettleLiveRunningRow(row,hasSettledThinking);    // may return a NEW object
```

`_anchorSceneSettleLiveRunningRow` (messages.js:3460) can return a fresh object, so if a future edit moved the `.has(row)` check below that reassignment the WeakSet lookup would silently miss and the fix would regress with no test failure on the happy path. Worth a one-line comment pinning "check identity before settle-reassign" so it survives the next refactor. The dead `rowIndex` param Greptile flagged is now inert (only `finalSegmentEligible` gates the drop) — cheap to remove and would remove the temptation to reintroduce an index-based boundary.

## Pre-tool narration protection is genuinely preserved

The `idx>lastProjectedToolIndex` guard is what keeps #5758's protection intact: a live-prose row that prefixes the final answer but sits *before* the last projected tool is not added to the WeakSet, so `finalSegmentEligible` is false and `rowIsLiveTokenFinalPrefix` short-circuits on its first `&&`. That matches the surviving `test_issue5749_settlement_preserves_pre_tool_live_token_prefix_rows` assertion (`local_id`s == `[fixture_row, "tool-row-1"]`), and the new multi-segment test locks the complementary case.

## Test plan

`test_issue5749_multi_segment_settled_tool_rows_do_not_shield_final_accumulator` pins the precise reproduction shape: `activity_rows` = narration (`live-prose:stream-multi:1`) → live tool → accumulator (`live-prose:stream-multi:2` at 70 chars, a strict prefix well under the 0.9 near-match ratio), with a `settled-tool-row-1` re-listed after. The assertions are the right three:

```python
assert "live-prose:stream-multi:2" not in local_ids   # stale accumulator dropped
assert "live-prose:stream-multi:1" in local_ids        # narration survives
assert data[0]["text"] == narration_text
assert any(row["role"] == "tool" for row in data)      # tool rows survive
```

One case I'd add for completeness: a *two*-final-segment turn (tool → accumulator-A → tool → accumulator-B) to prove only the last segment's accumulator is dropped and an earlier post-tool accumulator isn't collaterally caught by the `idx>lastProjectedToolIndex` window. As-is the change is safe to merge; that's a hardening test, not a blocker. The description's note that legacy persisted scenes still render clean via `_anchorSceneRowsForRendering`'s merge-reordering (so no migration) is consistent with the reload-path code — good call not persisting a migration for it.

nesquena-hermes:
Shipped in **exp-v0.52.100**. Thanks @ai-ag2026! 🎉 The full gate confirmed the WeakSet-chronology eligibility only drops stale strict-prefix live-prose rows and never a legitimate final segment (Codex reproduced no-tool/one-tool/two-tool/pre-tool-narration/post-tool/settled-backfill), the activity-stream regression gate passed all layers, and Nathan approved the desktop+mobile transparent-stream render. Clean, well-tested fix.

PR #6190, opened 2026-07-17
fix: decode chunked request bodies so proxied POSTs (Cloudflare Tunnel) authenticate

ericmaster:
> **⚠️ Disclaimer — AI-authored, submitted on a maintainer's behalf.** This PR was written by an AI agent (Claude Code) working on behalf of a user who hit this bug running Hermes WebUI behind a Cloudflare Tunnel. They don't have time to shepherd it through review, so **please feel free to close it with zero friction if that's easier for you — genuinely no hard feelings.** It's offered as a ready-to-use fix + reproduction in case it's useful; the diagnosis and verification below are real and reproducible either way.

## Thinking Path

- Hermes WebUI is meant to run comfortably behind a reverse proxy / tunnel (the README and env docs cover `HERMES_WEBUI_ALLOWED_ORIGINS`, `TRUST_FORWARDED_*`, Docker, etc.).
- A common deployment is Cloudflare Tunnel (`cloudflared`) in front of the origin. Browsers speak HTTP/2 to the Cloudflare edge; `cloudflared` then **streams** that request to the HTTP/1.1 origin.
- When a proxy streams a body whose length it doesn't know up front, it forwards it with `Transfer-Encoding: chunked` and **no `Content-Length`**. This is normal, spec-compliant proxy behavior.
- `api/helpers.py:read_body()` read the body as `rfile.read(int(Content-Length))`. With no `Content-Length`, it read **nothing** and returned `{}`.
- So every authenticated/data POST through the tunnel arrived with an empty body. The most visible symptom: **login always failed** — `body.get("password","")` was `""`, so `verify_password("")` returned false and the server answered `401 "Invalid password"` no matter what the user typed. (A second tell: the server left the un-read chunked bytes in the socket, which the keep-alive handler then parsed as a bogus follow-up request — the phantom `"method": "-", "status": 400` lines in the access log.)
- This PR teaches `read_body()` to decode chunked bodies, so proxied POSTs carry their payload and authenticate normally.

## What Changed

- `api/helpers.py`
  - `read_body()` now checks `Transfer-Encoding` first. If it contains `chunked`, the body is decoded via a new `_read_chunked_body()` helper (RFC 7230 §4.1: `Transfer-Encoding` takes precedence over `Content-Length`). Otherwise the existing `Content-Length` path runs **unchanged**.
  - `_read_chunked_body()` reads chunk-size lines (ignoring `;`-delimited chunk extensions), reassembles chunk data in order, consumes the terminating zero-size chunk + trailers, and enforces the same `MAX_BODY_BYTES` (20 MB) cap. Malformed chunk sizes / oversize bodies raise `ValueError` and set `close_connection = True`, matching the existing error contract.
  - `_read_exact()` helper guards against short reads from the socket buffer.
- `tests/test_chunked_request_body.py` (new) — behavior tests for `read_body()`.

Fix is at the **shared chokepoint** (`read_body()` — all 6 JSON POST call sites), not per-endpoint.

## Why It Matters

Without this, Hermes WebUI is effectively unusable behind any HTTP/2 front end that streams request bodies (Cloudflare Tunnel is the common one): you can load the page and do GETs, but **cannot log in or submit any POST**, with a misleading "Invalid password" error that sends people chasing a credential problem that doesn't exist.

## Verification

Local, Python 3.11 venv.

**Unit — the test bites (rule 6):**
- Against **pre-fix** `read_body()`: the 3 chunked tests **fail** (they get `{}` back); the `Content-Length` and empty-body tests pass. → `3 failed, 2 passed`
- Against **post-fix**: `5 passed`.

**Neighboring sweep (no regressions):**
```
pytest tests/test_chunked_request_body.py tests/test_issue1855_request_diagnostics.py \
  tests/test_issue3582_tts_content_length.py tests/test_trusted_header_auth.py \
  tests/test_1560_password_env_var_no_op.py tests/test_auth_settings_safety.py \
  tests/test_cors_preflight_allowlist.py tests/test_security_review_fixes.py
→ 110 passed
```

**Integration — real socket against a running server (the actual user symptom):**
| Request framing | password | Before fix | After fix |
|---|---|---|---|
| `Content-Length` | correct | 200 | 200 |
| `Transfer-Encoding: chunked` | correct | **401** | **200** |
| `Transfer-Encoding: chunked` | wrong | 401 | 401 |

The last row confirms auth is not weakened — a wrong password over a chunked request is still rejected.

## Risks / Follow-ups

- **Scope / known siblings left out (rule 1):** the multipart **upload** handlers in `api/upload.py` also read bodies via `Content-Length` (lines ~210/388/420/604) and would likewise mis-read a chunked upload. I deliberately left those out of scope — they're a separate, larger change (streaming multipart parse) with different risk, and I didn't want to widen this diff. Flagging them explicitly rather than hiding the gap.
- Chunked **trailers** are consumed and discarded (Hermes doesn't use trailer headers).
- Non-compliant bare-LF chunk framing isn't specially handled; `cloudflared` and standard proxies use CRLF.
- I could **not** verify end-to-end through an actual live Cloudflare Access-gated tunnel from CI (the headless environment can't pass the Access challenge); the chunked **socket reproduction** stands in for that path and is a faithful byte-level match of what `cloudflared` sends.

## Release note

> Fix: request bodies sent with `Transfer-Encoding: chunked` (e.g. via Cloudflare Tunnel / any HTTP/2 reverse proxy) are now decoded correctly. Previously such POSTs — including login — arrived empty, causing a spurious "Invalid password" / lost-payload behavior behind a tunnel.

(Left `CHANGELOG.md` untouched per CONTRIBUTING.)

- **Provider:** Anthropic
- **Model:** Claude Opus 4.8 (1M context) — model ID `claude-opus-4-8[1m]`
- **Tooling:** Claude Code (agentic CLI). Diagnosis was done by reproducing the failure at the socket level against a live instance, then reading `read_body()`; the fix and tests were authored and verified by the agent. A human directed the work and reviewed the outcome but did not hand-review every line — see the disclaimer at the top.

### DIFF
--- a/api/helpers.py
+++ b/api/helpers.py
@@ -654,9 +654,9 @@ def _read_chunked_body(handler, max_bytes: int) -> bytes:
         size_token = size_line.split(b';', 1)[0].strip()
         try:
             size = int(size_token, 16)
-        except ValueError:
+        except ValueError as err:
             handler.close_connection = True
-            raise ValueError(f'Malformed chunk size: {size_token!r}')
+            raise ValueError(f'Malformed chunk size: {size_token!r}') from err
         if size == 0:
             # Consume optional trailer headers up to the terminating blank line.
             while True:
--- a/api/helpers.py
+++ b/api/helpers.py
@@ -623,8 +623,72 @@ def redact_session_data(session_dict: dict) -> dict:
     return result
 
 
+def _read_exact(rfile, n: int) -> bytes:
+    """Read exactly n bytes, tolerating short reads from the socket buffer."""
+    buf = bytearray()
+    while len(buf) < n:
+        chunk = rfile.read(n - len(buf))
+        if not chunk:
+            break  # premature EOF
+        buf.extend(chunk)
+    return bytes(buf)
+
+
+def _read_chunked_body(handler, max_bytes: int) -> bytes:
+    """Decode a ``Transfer-Encoding: chunked`` request body.
+
+    Python's ``http.server`` never populates Content-Length for chunked
+    requests and does not decode them, so ``rfile.read(Content-Length)`` reads
+    nothing. Reverse proxies that stream an HTTP/2 client request to an
+    HTTP/1.1 origin — notably Cloudflare Tunnel (cloudflared) — forward request
+    bodies this way, so without decoding here the server sees an empty body on
+    every proxied POST (login, settings, message send, …). See RFC 7230 §4.1.
+    """
+    rfile = handler.rfile
+    out = bytearray()
+    while True:
+        size_line = rfile.readline(65536)
+        if not size_line:
+            break  # premature EOF — return what we have
+        # A chunk-size line may carry ';'-delimited chunk extensions; drop them.
+        size_token = size_line.split(b';', 1)[0].strip()
+        try:
+            size = int(size_token, 16)
+        except ValueError:
+            handler.close_connection = True
+            raise ValueError(f'Malformed chunk size: {size_token!r}')
+        if size == 0:
+            # Consume optional trailer headers up to the terminating blank line.
+            while True:
+                trailer = rfile.readline(65536)
+                if trailer in (b'\r\n', b'\n', b''):
+                    break
+            break
+        if len(out) + size > max_bytes:
+            handler.close_connection = True
+            raise ValueError(f'Request body too large (> {max_bytes} bytes)')
+        out.extend(_read_exact(rfile, size))
+        rfile.readline(3)  # consume the CRLF that terminates the chunk data
+    return bytes(out)
+
+
 def read_body(handler) -> dict:
-    """Read and JSON-parse a POST request body (capped at 20MB)."""
+    """Read and JSON-parse a POST request body (capped at 20MB).
+
+    Handles both ``Content-Length`` and ``Transfer-Encoding: chunked`` framing.
+    The latter is required for bodies proxied by cloudflared / any HTTP/2 front
+    end, which forward to the HTTP/1.1 origin without a Content-Length header.
+    """
+    # RFC 7230 §3.3.3: Transfer-Encoding takes precedence over Content-Length;
+    # a chunked body has no Content-Length to read.
+    transfer_encoding = handler.headers.get('Transfer-Encoding', '') or ''
+    if 'chunked' in transfer_encoding.lower():
+        raw = _read_chunked_body(handler, MAX_BODY_BYTES) or b'{}'
+        try:
+            return _json.loads(raw)
+        except Exception:
+            return {}
+
     raw_length = handler.headers.get('Content-Length', 0)
     try:
         length = int(raw_length)
--- a/tests/test_chunked_request_body.py
+++ b/tests/test_chunked_request_body.py
@@ -1,0 +1,93 @@
+"""read_body() decodes Transfer-Encoding: chunked request bodies.
+
+Reverse proxies that stream an HTTP/2 client request to an HTTP/1.1 origin —
+notably Cloudflare Tunnel (cloudflared) — forward request bodies with
+``Transfer-Encoding: chunked`` and no ``Content-Length``. Before the fix,
+``read_body()`` only honoured ``Content-Length``, so every proxied POST
+arrived with an empty body. For JSON login that meant the parsed password was
+``""`` and the server returned 401 "Invalid password" no matter what the user
+typed; other POSTs (settings, message send) silently lost their payload too.
+
+These tests assert the observable behaviour of ``read_body`` — the decoded
+dict — not a source string. The chunked cases fail against the pre-fix code
+(they get ``{}`` back) and pass after it.
+"""
+import email.message
+import io
+import json
+
+from api.helpers import read_body
+
+
+class _FakeHandler:
+    """Minimal stand-in for BaseHTTPRequestHandler's body-reading surface.
+
+    Uses a real ``email.message.Message`` so header lookups are
+    case-insensitive, exactly as ``http.server`` delivers them (cloudflared
+    sends the header lowercased as ``transfer-encoding``).
+    """
+
+    def __init__(self, raw: bytes, headers: dict):
+        self.rfile = io.BytesIO(raw)
+        msg = email.message.Message()
+        for key, value in headers.items():
+            msg[key] = value
+        self.headers = msg
+        self.close_connection = False
+
+
+def _one_chunk(body: bytes) -> bytes:
+    """Frame *body* as a single HTTP chunk followed by the terminator."""
+    return b"%x\r\n%s\r\n0\r\n\r\n" % (len(body), body)
+
+
+def test_chunked_body_is_decoded_lowercase_header():
+    payload = {"password": "correct horse battery staple", "n": 42}
+    body = json.dumps(payload).encode()
+    # cloudflared casing: lowercase header name, and NO Content-Length.
+    handler = _FakeHandler(
+        _one_chunk(body),
+        {"transfer-encoding": "chunked", "content-type": "application/json"},
+    )
+    assert read_body(handler) == payload
+
+
+def test_chunked_body_reassembled_across_multiple_chunks():
+    # A body split across several chunks must be reassembled in order; a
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
## Summary

Read `api/helpers.py:626-649` on `origin/master` against the diff here, plus the login path at `api/routes.py:16171` and the shared POST chokepoint at `api/routes.py:13774`. The diagnosis holds up end to end: `read_body()` only ever consulted `Content-Length`, and a chunked request has none, so `raw = handler.rfile.read(length) if length else b'{}'` fell into the `else` branch and returned `{}`. Login then read an empty password and 401'd. The fix at the shared chokepoint is the right layer — every JSON POST handler funnels through the single `body = read_body(handler)` call.

## Code reference

The exact spot the empty body came from, on master:

```python
# api/helpers.py:649
raw = handler.rfile.read(length) if length else b'{}'
```

And the visible symptom the PR describes, confirmed at `api/routes.py:16171`:

```python
password = body.get("password", "")
if not verify_password(password):
    _record_login_attempt(client_ip)
    return bad(handler, "Invalid password", 401)
```

With `body == {}`, `password` is `""` and `verify_password("")` fails regardless of input — matching the report precisely.

## The upload.py sibling gap is real

The author flagged multipart uploads as out of scope; I confirmed the gap is genuine. `api/upload.py:210` (and the mirrored reads at `:388`, `:420`, `:604`) does:

```python
content_length = int(handler.headers.get('Content-Length', 0) or 0)
...
fields, files = parse_multipart(handler.rfile, content_type, content_length)
```

Behind cloudflared a chunked multipart upload arrives with `content_length == 0`, so `parse_multipart` sees zero bytes and the user gets a confusing "no file" / empty-fields error rather than a clear signal. Agreeing with the scoping decision — streaming multipart decode is a bigger, higher-risk change and doesn't belong in this diff.

One cheap forward-looking guard worth considering as a follow-up (not this PR): detect `Transfer-Encoding: chunked` at the top of `handle_upload` and return an explicit error (e.g. 411 "chunked uploads not supported behind this proxy") instead of silently parsing 0 bytes. That converts a mystifying empty-upload into an actionable message until the streaming parse lands.

## Notes on the fix itself

- TE-over-CL precedence in `read_body()` is spec-correct (RFC 7230 §3.3.3). Good call checking `Transfer-Encoding` first.
- `readline(3)` to consume the post-chunk CRLF is fine for CRLF framing (stops at `\n` after 2 bytes); as the PR notes, bare-LF framing isn't handled, which is acceptable for cloudflared/standard proxies.
- The premature-EOF paths degrade gracefully: `_read_exact` short-read gives a truncated body so `json.loads` fails and returns `{}`, and an empty chunked stream returns `b'{}'`. No crash surface.
- Greptile's two nits (unbounded trailer loop, `close_connection` set unwrapped in the error branches) are cosmetic given the trusted-proxy context; the trailer lines are each capped at 65536 bytes by `readline`.

## Test plan

The new `tests/test_chunked_request_body.py` covers the meaningful cases: lowercase header (as cloudflared sends), multi-chunk reassembly, chunk extensions, plus Content-Length and empty-body regressions. The multi-chunk test is the important one — it would catch a reader that dropped all but one chunk. The "bites against pre-fix" verification (3 failed then 5 passed) confirms the tests actually exercise the change.

nesquena-hermes (changes requested):
## The chunked-body decode is a genuinely needed fix, but the parser has DoS vectors on a public endpoint — hardening required

Thanks @ericmaster — decoding `Transfer-Encoding: chunked` for cloudflared/HTTP2-proxied POSTs is a real and valuable fix (empty-body-on-every-proxied-POST is a nasty deployment bug). But because `read_body()` runs on **public unauthenticated endpoints** (notably `/api/auth/login`), the chunk parser has to be hardened against malicious framing. The adversarial gate reproduced three issues (two are DoS BRICKs):

1. **BRICK — negative chunk sizes bypass the size cap (`api/helpers.py:654`).** `int("-1", 16)` succeeds and returns `-1`, so `len(out) + size > max_bytes` is never true → the 20 MB cap is defeated. Verified: 10,000 `-1\r\n\r\n` chunks returned HTTP 200 without closing. An endless stream can occupy each of the server's 128 workers through public `/api/auth/login` → denial of service for every user. **Fix:** reject any size token that isn't non-empty ASCII hex (`^[0-9A-Fa-f]+$`); on failure `close_connection=True` + HTTP 400.

2. **BRICK — unbounded trailer/framing metadata (`api/helpers.py:662`).** The zero-chunk trailer loop (and chunk-extension lines) have no aggregate byte/line bound. Verified: 10,000 trailer lines consumed without error or close → workers held indefinitely. **Fix:** enforce a cumulative metadata/trailer cap (bytes AND line count); close on excess or an overlong/non-terminated line.

3. **SILENT — malformed/truncated framing accepted (`api/helpers.py:650`).** An invalid `X\n` chunk-data delimiter returned HTTP 200 with a parsed body, and a premature EOF (fewer bytes than the declared chunk size) also returned parsed JSON with `close_connection=False`. **Fix:** require CRLF-terminated size lines, exact chunk-data length, exactly `\r\n` after each chunk's data, and a complete zero-chunk + trailer terminator; otherwise close and return HTTP 400.

Net: treat the chunk stream as hostile — validate size tokens as strict hex, enforce the body cap against non-negative sizes only, bound the total framing metadata, and fail closed (connection close + 400) on any framing violation instead of returning a partial body as 200. Add regressions for negative-size, trailer-flood, bad-delimiter, and premature-EOF. Once hardened this is a good fix to land. Credit preserved.

nesquena-hermes:
## Gate: SAFE on the code — parking for maintainer sign-off (auth-path chokepoint)

The chunked-body decoder is correct and the diagnosis holds (confirmed on master: `read_body()` only consulted Content-Length, so a chunked POST — as cloudflared/any HTTP/2 front end forwards — returned `{}` and login 401'd).

**Adversarial verification (reproduced directly against the branch, not just read):**
- ✅ Multi-chunk decode → correct concatenated body; chunk extensions (`;ext`) + trailer headers consumed.
- ✅ **DoS — many small chunks exceeding the cap → RAISES** `Request body too large` + `close_connection=True` (the cumulative `len(out)+size > max_bytes` check bounds TOTAL bytes; an attacker can't bypass the 20MB cap by fragmenting).
- ✅ Single huge chunk-size claim → RAISES *before* allocating.
- ✅ Malformed hex chunk size → RAISES + closes connection.
- ✅ Premature EOF mid-chunk → returns partial, no infinite loop / hang.
- ✅ No-terminator infinite-chunk stream → stops at EOF, no hang.
- ✅ **Content-Length path byte-identical to master** (the common case is untouched).
- Full serial suite: **13,604 passed, 0 failed**; contributor's 93-line decoder test included.

Note on process: the Codex adversarial pass emitted SAFE TO SHIP but its reasoning was cut short by an automated content filter while it was reading the auth/password path, so I did NOT rely on that verdict — the checks above are my own direct reproduction.

This makes auth *work* behind a reverse proxy (it's request-body framing, not an auth-logic change), and the adversarial surface is clean. Because it's the shared chokepoint every POST (including login) funnels through, I'm parking it for maintainer sign-off rather than self-ship. **Recommendation: SHIP.** Preserve @ericmaster attribution.

nesquena-hermes:
Thanks @ericmaster. Your diagnosis here was the original correct one: `read_body()` only consults `Content-Length`, so cloudflared / HTTP-2 front ends that forward chunked bodies make every proxied POST (including login) read as `{}`.

I'm closing this in favour of #6829, which fixes the same bug at the same chokepoint and has since been hardened through two adversarial gate rounds. This version would reopen a request-smuggling hole that master doesn't have today:
- it decodes chunked whenever `'chunked' in transfer_encoding.lower()`, even when a `Content-Length` is also present (the classic CL.TE / TE.CL desync), and a lookalike such as `xchunked` also matches;
- it reads only the first `Transfer-Encoding` header, so a duplicate can hide a second coding;
- chunk sizes go through `int(token, 16)`, which accepts `+`, `_` and non-ASCII digits, and the trailer loop has no size bound.

#6829 closes all of those and is on its final round (review 5299183703). When it ships, the CHANGELOG entry will credit you for the original report and fix direction.

PR #6197, opened 2026-07-17
Live Stream: prefer durable run-journal recovery state

### LINKED ISSUE #6196: Live Stream: newer run-journal snapshots can lose to stale browser recovery state
## Summary

While an assistant turn is still running, `loadSession()` can restore a visible browser-side `INFLIGHT` tail and keep it even when the server returns a newer `runtime_journal_snapshot` for the active stream.

This makes the browser cache act as the authority over the durable run journal during hard reload or session recovery. The first recovered Worklog can therefore be stale or incomplete. A legacy cache entry with no stream identity is more serious: its sequence cursor can be treated as belonging to the current stream and suppress replay of earlier events from a newer run.

## Reproduction trace

1. Start a turn and let the run journal advance.
2. Persist a browser `INFLIGHT` entry with visible assistant/Worklog state at an older journal sequence.
3. Reload while the turn is still active.
4. The metadata response includes a newer `runtime_journal_snapshot` for the current stream.
5. `loadSession()` keeps the local entry solely because it has visible content, then reconnects from the local cursor.

Before the fix, a production-helper regression demonstrated this directly: local sequence `3` was selected over a same-stream durable snapshot at sequence `5`.

## Expected behavior

- The current stream's durable run-journal snapshot owns the recovered Worklog when it is at least as advanced as the browser cache.
- A browser tail may win only when it is proven to belong to the same stream and has advanced beyond the metadata snapshot during the request race.
- Entries from another stream, or legacy entries with no stream identity, must not override the current durable snapshot.
- The run-aware replay cursor (`sequence` plus `event_id`) must survive browser persistence and reconnect.
- Same-stream browser-only recovery fields such as a fresher Todo snapshot should be preserved without merging stale prose, Thinking, Tool, or Anchor rows back into the Worklog.

## Contract routing

This is a conformance fix for the existing recovery contracts, not a contract change:

- `docs/rfcs/stable-assistant-turn-anchors.md`
- `docs/rfcs/session-sse-contract-v1.md`

The relevant invariant is that the run journal is durable truth while browser `INFLIGHT`, DOM, and message caches are projections/recovery caches.

## Acceptance criteria

- [ ] Recovery selection is stream-aware and cursor-aware.
- [ ] Same-stream newer/equal durable snapshots replace stale visible browser Worklog state.
- [ ] A strictly newer same-stream browser tail is retained to cover metadata-fetch races.
- [ ] Missing or mismatched local stream identity fails closed to the current durable snapshot.
- [ ] `stream_id`, `last_seq`, and `last_event_id` survive snapshot hydration, local persistence, and SSE reattach.
- [ ] Session-switch, BFCache, Anchor scene, run-journal, and state reconciliation regressions remain green.

franksong2702:
## Thinking Path

- Live-to-Final recovery treats the run journal as durable truth; browser `INFLIGHT`, DOM, and message state are recovery projections.
- `loadSession()` instead kept any visible local `INFLIGHT` tail, even when the current stream's `runtime_journal_snapshot` had a newer cursor.
- A production-helper regression reproduced the authority inversion: a visible local tail at sequence 3 won over a same-stream durable snapshot at sequence 5.
- The fix makes recovery selection stream-aware and cursor-aware, then persists the complete run-aware cursor through reload and SSE reattach.
- The result is deterministic Worklog recovery without merging stale browser prose, Thinking, Tool, or Anchor rows over durable state.

## What Changed

- Select between local `INFLIGHT` and `runtime_journal_snapshot` using current stream identity plus run-journal sequence.
  - Durable state wins when it is equally or more advanced.
  - A proven same-stream browser tail wins only when it advanced after the metadata snapshot was read.
  - Mismatched or identity-less local state fails closed to the current durable snapshot.
  - If neither recovery source owns the active stream, discard the stale browser cache and reattach from the current stream.
- Carry `streamId` and `lastRunJournalEventId` through server snapshot hydration, in-memory `INFLIGHT`, localStorage compaction, session-switch persistence, and SSE reconnect.
- Reset replay cursors when an in-memory tail changes stream ownership.
- Preserve only same-stream browser Todo recovery state when the durable Worklog wins.
- Reject an explicitly foreign-stream Anchor scene from the runtime-journal fallback renderer.
- Extend the cross-session Node harness to execute the new production recovery helpers and add a behavioral boundary matrix.

## Why It Matters

A hard reload or recovery during a long-running turn should show the same ordered assistant turn represented by the durable run journal. Stale browser state must not become the authority, especially when a legacy sequence cursor could otherwise be mistaken for a newer stream and suppress replay.

Fixes #6196

## Verification

- Red-before-green regression:
  - Before the behavior fix, `test_live_recovery_prefers_newer_durable_run_journal_snapshot` failed because the local sequence-3 object was selected instead of the server sequence-5 object.
  - After the fix, it covers newer/equal durable state, strictly newer local state, mismatched/missing stream identity, two-hop stale recovery, absent local/server state, same-stream Todo carry-forward, and foreign-stream Anchor-scene rejection.
- `node --check static/sessions.js static/messages.js static/ui.js`
- `./scripts/test.sh` focused Live-to-Final/run-journal/session recovery set: **303 passed**
- Final directly touched harness set: **56 passed**
- `git diff --check`

The full local suite was also attempted: 13,117 passed, 118 skipped, 1 xfailed, and 18 failed. The only failure on a touched `loadSession()` path was the Node harness dependency exposed by this extraction; it was fixed and now passes. The other failures are in unchanged network/provider/model-catalog tests and were partially reproducible in a fresh isolated rerun. CI remains the authoritative clean-environment full-suite result.

No screenshot is included because this changes recovery authority and replay identity, not layout or a stable visual state; the executable production-helper regression is the before/after evidence.

## Risks / Follow-ups

- A same-stream local tail can still win when its sequence is strictly newer than the earlier metadata response. This is intentional protection for events received during the metadata/message fetch race.
- Todo state is the only local field carried onto a durable Worklog snapshot, and only for a verified matching stream.
- When both local and server recovery state explicitly belong to other streams, the cache is cleared so the current stream reattaches without painting foreign rows.
- No storage migration is required. Older entries without stream identity deliberately defer to durable current-stream state.
- This does not change the existing behavior for a run-journal snapshot that has a cursor but no renderable rows.

## Release Note

Live Stream: recover active Worklogs from the latest run-journal state after reload or re-entry, with stream-aware replay cursors that prevent stale browser state from overriding the current run.

## Model Used

- Provider: OpenAI
- Model: OpenAI Codex (GPT-5)
- Notable mode/tools: Codex Desktop, shell-based repository inspection, red-before-green Node behavior harness, focused and full pytest runs

## Contract Routing

- Contract family: Live-to-Final assistant replies; stable Assistant Turn Anchors; WebUI run-state consistency; session SSE replay.
- Documents reviewed:
  - `docs/rfcs/stable-assistant-turn-anchors.md`
  - `docs/rfcs/session-sse-contract-v1.md`
- Routing result: conformance fix only; no contract text changes. The existing rule already makes run-journal state durable truth and browser `INFLIGHT` a recovery cache.
- Evidence: production-helper behavior tests in `tests/test_inflight_stream_reuse.py`, cross-session execution in `tests/test_cross_session_message_load_isolation.py`, and the 303-test recovery/Anchor/run-journal/state reconciliation gate listed above.


franksong2702:
Addressed both Greptile follow-ups on the current head `b1264a1c`:

- Removed the unreachable no-active-ID branch by deriving one comparison identity from the requested active stream, falling back to the server snapshot identity only for helper reuse.
- Added the two-hop stale boundary: if the server snapshot belongs to another stream, local state is retained only when it is explicitly owned by the requested active stream. If neither source owns that stream, the browser cache is cleared and the existing current-stream reattach path rebuilds recovery state.
- Extended the same ownership guard to the runtime-journal Anchor-scene fallback so an explicitly foreign-stream scene cannot be painted under the current stream.

Verification on the exact PR head:
- focused Live-to-Final / Anchor / run-journal / recovery gate: **303 passed**
- directly touched harness set: **56 passed**
- `node --check`, `git diff --check`, and diff-scoped ruff: pass

All 17 GitHub Actions checks on this head are green. The refreshed Greptile review completed at **5/5, Safe to merge**, with no remaining blocker.

nesquena-hermes:
## Review at `b1264a1c` — recovery authority inversion is fixed cleanly, one edge worth a test

Pulled both commits into a read-only worktree and diffed `static/sessions.js`, `static/messages.js`, `static/ui.js` against `origin/master`, then cross-checked the server contract in `api/routes.py`. CI is green (all `test (3.1x, *)` + browser-smoke + lint SUCCESS), and the fix maps to #6196's acceptance criteria. Direction and shape are right — the run journal stays durable truth and `INFLIGHT` stays a projection.

## The core selector reads correctly

`_selectLiveRecoveryInflight` (`static/sessions.js:1074`) is the right place to arbitrate, and the precedence ladder matches the issue's five criteria:

```js
if(requestedActiveId&&serverId&&serverId!==requestedActiveId){
  return localId===requestedActiveId?localInflight:null;
}
if(activeId&&localId!==activeId) return selectDurableSnapshot();
const localSeq=Math.max(0,Number(localInflight.lastRunJournalSeq)||0);
const serverSeq=Math.max(0,Number(serverLiveSnapshot.lastRunJournalSeq)||0);
return serverSeq>=localSeq?selectDurableSnapshot():localInflight;
```

The `serverSeq>=localSeq` tie-break to durable is exactly #6196's "at least as advanced" clause, and returning `null` when neither source owns the requested stream is what drives the new `delete INFLIGHT[sid]` + `clearInflightState(sid)` branch in `loadSession` (`sessions.js:2085-2090`) so the current stream reattaches clean instead of painting foreign rows. Good — that's the "legacy cache with no stream identity suppresses replay" case the issue called the more serious one.

## Server contract backs the new fields

Verified the frontend isn't reading fields the server doesn't emit. `_run_journal_live_snapshot` returns them at the top level:

```python
# api/routes.py:3591-3595
return {
    "session_id": session_id,
    "stream_id": stream_id,
    "last_seq": last_seq,
    "last_event_id": last_event_id,
```

so `_serverLiveSnapshotInflight` picking up `snapshot.stream_id` / `snapshot.last_event_id` (`sessions.js:1055,1066`) is sound, and `last_event_id` is computed as `f"{stream_id}:{event_last_seq}"` (routes.py:3582) so the cursor is stream-scoped by construction. The `attachLiveStream` reset that zeroes `lastRunJournalSeq`/`lastRunJournalEventId` on stream-ownership change (`messages.js:2035-2040`) then correctly forces a full replay rather than resuming a foreign cursor.

## One edge the test matrix doesn't pin: Todo carry-forward requires `todoStateMeta` truthy

`selectDurableSnapshot` only carries local Todos onto the durable snapshot when both `localInflight.todos` is an array **and** `localInflight.todoStateMeta` is truthy (`sessions.js:1085`). `_serverLiveSnapshotInflight` sets `todos:null, todoStateMeta:null` (sessions.js:1059-1060), so a durable-wins path always drops server-side Todo state to null unless the *local* entry supplied both. That's the intended "only same-stream browser Todo survives" behavior, but it means: if a same-stream local tail has `todos` populated from a fresher snapshot but `todoStateMeta` was never persisted (older storage entry), the Todos are silently discarded on a durable win. The new matrix in `test_inflight_stream_reuse.py:1096` only exercises the both-present case (`todoStateMeta:{ts:123}`). Worth adding a `todos-present/todoStateMeta-absent` row asserting the intended drop, so a future refactor of the guard condition can't silently start merging half-formed Todo state.

## Verdict

Conformance fix, no contract change, CI green, red-before-green regression present (`test_live_recovery_prefers_newer_durable_run_journal_snapshot`, `sessions.js` helpers extracted and driven directly under Node). I'd merge after the one Todo-meta-absent test row above — everything else in the acceptance list is covered by the existing matrix. Nice that the foreign-stream Anchor-scene rejection (`_anchorActivitySceneMatchesStream`, sessions.js:1099) closes the last painting path; that was easy to miss.

franksong2702:
Addressed the Todo metadata edge from the latest review in commit `91bdd89a`.

Added a `todos`-present / `todoStateMeta`-absent row to the existing `_selectLiveRecoveryInflight` Node matrix. It asserts the durable server snapshot wins as-is and half-formed local Todo state is not carried forward.

Verification on this head:
- `./scripts/test.sh tests/test_inflight_stream_reuse.py -k live_recovery_prefers_newer_durable_run_journal_snapshot` -> 1 passed
- `./scripts/test.sh tests/test_inflight_stream_reuse.py` -> 42 passed
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check` -> pass

GitHub Actions are green on `91bdd89a`; Greptile is still pending, so I’m leaving the card monitored as waiting on CI until that bot check completes.

franksong2702:
Refreshed the PR branch onto current `origin/master` in merge commit `e9444fc8d` (`b47079347` base).

Local verification after the base refresh:
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py` -> 108 passed
- `node --check static/sessions.js` -> pass
- `node --check static/messages.js` -> pass
- `node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD` -> pass

GitHub checks restarted on `e9444fc8d`, so I am leaving this monitored as waiting on CI until those finish.

franksong2702:
Refreshed the PR branch onto current `origin/master` again in merge commit `0082c8909` (`ebf64b8a` base). The new base was docs-only (`CONTRIBUTING.md`, `docs/GUIDELINES.md`), and the PR diff against base remains the same five recovery files.

Local verification after this refresh:
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py` -> 108 passed
- `node --check static/sessions.js` -> pass
- `node --check static/messages.js` -> pass
- `node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD` -> pass

GitHub checks restarted on `0082c8909`, so I am leaving the card monitored as waiting on CI until those finish.

franksong2702:
Refreshed the PR branch onto current `origin/master` in merge commit `e1d2a9f4e` (`ab937ef3` base). The PR diff against the new base remains scoped to the same five recovery files.

Local verification after this refresh:
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py` -> 108 passed
- `node --check static/sessions.js` -> pass
- `node --check static/messages.js` -> pass
- `node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD` -> pass

GitHub checks restarted on `e1d2a9f4e`, so I am leaving this monitored as waiting on CI until those finish.

franksong2702:
Refreshed the PR branch onto current `origin/master` in merge commit `ba7d495ee` (`52c962cf` base). The PR diff against the new base remains scoped to the same five recovery files.

Local verification after this refresh:
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py` -> 108 passed
- `node --check static/sessions.js` -> pass
- `node --check static/messages.js` -> pass
- `node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD` -> pass

GitHub checks restarted on `ba7d495ee`, so I am leaving this monitored as waiting on CI until those finish.

nesquena-hermes:
Thanks for this — the recovery-authority inversion is the right fix and the cursor pipeline is genuinely well-built (the deep gate confirmed the round-trip is real and complete, cross-session isolation holds/strengthens, and same-stream newer-local selection works as intended). But the authoritative gate reproduced **one silent data-quality regression** that has to be closed before ship, so I'm bouncing with the fix.

### The defect (reproduced, not read)

`static/sessions.js:1097` — the new `return serverSeq >= localSeq ? selectDurableSnapshot() : localInflight;` makes the durable `runtime_journal_snapshot` win on an equal-or-newer sequence. But that server snapshot is an intentionally **lossy preview projection for tool arguments**: `api/routes.py:3164` and `:3234` both run the args through `_truncate_journal_snapshot_value(..., limit=120)` (routes.py:3083 — truncates every string to 120 chars, lists to 20 items). Meanwhile local INFLIGHT persists the **complete** tool arguments (`static/messages.js:5079-5144`).

So on an active-stream reload where the durable snapshot ties/leads on seq:
1. `_selectLiveRecoveryInflight` picks the truncated durable snapshot over the complete local tail → full command/input detail is **erased**.
2. The snapshot's `last_event_id` then advances the replay cursor past the source tool event → replay **won't restore** the original untruncated args either.

**Verified probe:** a 247-character tool command restored **completely on `origin/master`** but became **123 characters on this branch**. That's a real, user-visible loss of tool-call detail on recovery — a regression the previous binary check didn't have (it kept the full local tail).

### Fix-spec (design choice — your call which, both are clean)

The tie-break/authority change is sound; the problem is *what the durable snapshot carries*. Pick one:

**Option A (server-side, preferred): stop truncating tool args in the recovery snapshot.** The 120-char truncation is appropriate for a lightweight Worklog *preview*, but `runtime_journal_snapshot` is now a *recovery authority*, so it must carry the same user-visible argument fidelity the browser INFLIGHT does. At `routes.py:3164` + `:3234`, preserve full args (or raise the bound to match the browser recovery bound), keeping the truncation only for any genuine preview field (`preview`/`snippet`). Then add an **equal-sequence recovery assertion** proving expanded tool details survive when the durable snapshot supersedes local state.

**Option B (client-side merge): when the winning durable snapshot is same-stream, splice the complete local tool args over the truncated snapshot ones** before returning it in `selectDurableSnapshot()` (you already merge `todos`/`todoStateMeta` there for the same-stream case — extend that to tool-call args keyed by `tid`/`tool_call_id`). This keeps the server projection lightweight but restores fidelity from the same-stream local cache.

I lean **Option A** — it fixes the loss at the source for every consumer of the snapshot (not just this selection path) and is simpler to assert. Add the equal-seq regression test either way (a >120-char tool command that must survive a durable-wins recovery, red-before-green).

Everything else gated clean: full suite (13,254), activity-stream regression gate all-layers, cursor round-trip, foreign-stream rejection, cross-session isolation, same-stream newer-local selection. Re-push the args-fidelity fix + the equal-seq test and I'll re-gate right away.

(Separately, a product note for the maintainer, not a blocker: the `serverSeq == localSeq → durable wins` tie-break itself is defensible — durable is authoritative and complete, and a local tail at the same seq has nothing genuinely newer. Once the truncation is fixed, that rule is fine.)

nesquena-hermes (changes requested):
Bounced with a reproduced silent tool-arg truncation regression on durable-wins recovery — see the detailed comment. Fix-spec (Option A preferred: stop truncating tool args in runtime_journal_snapshot; add an equal-seq recovery assertion) posted. Re-gate on re-push.

franksong2702:
Addressed the tool-arg fidelity blocker in `de6ac178d` (`fix(streaming): preserve recovery snapshot tool args`).

I took Option A: `runtime_journal_snapshot` now deep-copies full tool `args` for both started `tool` events and complete-only `tool_complete` fallback calls, while leaving preview/snippet strings as the lightweight preview fields.

Added `test_equal_seq_recovery_preserves_full_durable_tool_args`, which covers:
- red-before-green `>120` char command truncation from the durable snapshot
- both `tool` and complete-only `tool_complete` args in `tool_calls`
- Anchor activity scene `tool.args` / `payload.args` fidelity
- equal-seq browser recovery where `_selectLiveRecoveryInflight()` chooses the durable snapshot and still preserves the full command

Verification:
- red-before-green: `./scripts/test.sh tests/test_inflight_stream_reuse.py -k equal_seq_recovery_preserves_full_durable_tool_args` failed on truncated `...` command before the fix
- `./scripts/test.sh tests/test_inflight_stream_reuse.py -k equal_seq_recovery_preserves_full_durable_tool_args` -> 1 passed
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py tests/test_anchor_scene_persistence.py` -> 148 passed
- `node --check static/sessions.js && node --check static/messages.js && node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD && git diff --check` -> pass

Readback after push: PR head is `de6ac178d0cbe883af1106a23ae8c7633f1817d4`, mergeable, with GitHub checks restarted and pending.

franksong2702:
Follow-up for the failed shard after `de6ac178d`: CI was failing only on `tests/test_run_journal_routes.py::test_live_journal_snapshot_reconstructs_visible_progress_and_tool_aliases` because that older route test still asserted recovery tool args were truncated to the preview bound.

Pushed `3b96651c7` (`test(streaming): align journal snapshot args contract`) to update that expectation to the new recovery-authority contract: full `tool.args` are preserved, while `snippet` remains the preview field.

Additional local verification after this follow-up:
- `./scripts/test.sh tests/test_run_journal_routes.py -k live_journal_snapshot_reconstructs_visible_progress_and_tool_aliases` -> 1 passed
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_run_journal_routes.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py tests/test_anchor_scene_persistence.py` -> 162 passed
- `node --check static/sessions.js && node --check static/messages.js && node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD && git diff --check` -> pass

Readback after push: PR head is `3b96651c7295941c21caf7e71b5d78e8504add37`, mergeable, with checks restarted and pending.

franksong2702:
Refreshed the PR branch onto current `origin/master` in merge commit `f57039954` (`02aa91ba` base). The merge was clean; upstream touched `static/messages.js`, `api/streaming.py`, `tests/test_auto_compression_card.py`, and `CHANGELOG.md`.

Local verification after the refresh:
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_run_journal_routes.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py tests/test_anchor_scene_persistence.py` -> 162 passed
- `node --check static/sessions.js && node --check static/messages.js && node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD` -> pass

GitHub readback after push: PR head is `f57039954af07454b2d58564a8314a9fe6800570`, base is `02aa91ba6ece4454dcb580f960e5b516b3a91111`, mergeable. Checks restarted on the refreshed head, so I am leaving this monitored as waiting on CI until they finish.

nesquena-hermes:
Good progress — the re-push fixed the start-event case (a 247-char command in a start-event tool row now restores exactly, and the cursor round-trip / stream-change reset / cross-session isolation / newer-local selection / identity-less fallback / foreign-stream rejection all re-verified clean). But the round-2 gate reproduced **two remaining defects** — one is the same data-loss class on a path the first fix missed, one is a new unbounded-payload edge the de-truncation opened. Bouncing with both; you're close.

### SILENT — equal-seq durable recovery STILL loses tool args when `tool_complete` enriches an existing `tool` event — `api/routes.py:3154`

The fix covered the start-event args, but there's a second path: when a `tool_complete` event **enriches an existing `tool` event**, the matching branch marks the call complete but **never copies the completion's args**. Reproduced: an empty start event followed by a 247-char completion arg → the durable snapshot's tool row has `args: {}`, then `serverSeq == localSeq` selects that empty durable row over the complete browser state → the full command is lost again. (Same data-loss shape as round 1, just on the matched-event enrichment path.)

Fix: before returning from the matching branch (~routes.py:3154), replace empty/partial stored args with the non-empty completion args via `_run_journal_snapshot_recovery_args()`. Add the start-empty/partial-then-completion-full-at-equal-sequence variant to `test_equal_seq_recovery_preserves_full_durable_tool_args`.

### CORE — removing the truncation removed the size cap entirely (unbounded snapshot) — `api/routes.py:3097` + `api/gateway_chat.py:317`

`_truncate_journal_snapshot_value` did two things: truncate strings to 120 chars AND cap lists at 20 items. Removing it from the args restored full fidelity (good) but also removed the **item/size bound** — so a pathological tool call with huge args is now serialized unbounded, and `deepcopy()` serializes it 3× in the snapshot. Reproduced: a 50,000-item args probe produced an **18.1 MB** metadata response (was capped at 20 items before). That's a memory/response-size regression on the recovery path.

Fix: don't reinstate the blanket 120-char truncation (that's what caused the original data loss). Instead apply a **shared content-aware bound** (a generous per-value size limit + nested-collection/item cap + total-snapshot-size limit) at both the Gateway journaling site (routes.py:3097) and the legacy-journal read (gateway_chat.py:317), tuned to **preserve realistic values like the 247-char command exactly** while rejecting pathological megabyte payloads. Add a response-size regression test (huge args → bounded snapshot, but a normal long command survives intact).

Everything else re-verified clean. Re-push both fixes + the two regression tests and I'll re-gate — this is the last mile.

nesquena-hermes (changes requested):
Round 2: start-event args fixed + all prior invariants re-verified clean, but 2 remaining: (1) tool_complete-enriches-existing-tool path still loses args (routes.py:3154 — copy completion args in the matching branch), (2) de-truncation removed the size cap → 50k-item args = 18MB snapshot (routes.py:3097/gateway_chat.py:317 — add a content-aware bound preserving the 247-char command but capping pathological payloads). Fix-spec posted.

franksong2702:
Addressed the round-2 recovery args blockers in `bd6b08256` (`fix(streaming): bound recovery snapshot tool args`).

What changed:
- matched `tool_complete` events now enrich an existing `tool` row with richer completion args instead of leaving an empty/partial start-event args object in the durable snapshot
- added a shared `bound_run_journal_snapshot_args()` sanitizer in `api.run_journal`
- `runtime_journal_snapshot` uses that sanitizer when reconstructing legacy journal args
- Gateway tool-progress translation uses the same sanitizer before journaling Gateway args
- the sanitizer preserves normal long commands exactly (including the 247-char case), while bounding nested fanout/depth, long scalar values, and total per-args content for pathological payloads

Added/updated regressions:
- extended `test_equal_seq_recovery_preserves_full_durable_tool_args` for start-partial + completion-full at equal sequence
- added `test_live_journal_snapshot_bounds_pathological_tool_args` for a 50k-item legacy snapshot probe
- added `test_gateway_tool_progress_event_bounds_pathological_args` for the Gateway journaling path

Red-before-green:
- equal-seq recovery test failed with `partial` instead of the completion command before the fix
- legacy snapshot huge-args test failed with 50,000 retained items before the bound
- Gateway huge-args test failed with 50,000 retained items before the bound

Verification:
- `./scripts/test.sh tests/test_inflight_stream_reuse.py -k equal_seq_recovery_preserves_full_durable_tool_args` -> 1 passed
- `./scripts/test.sh tests/test_run_journal_routes.py -k live_journal_snapshot_bounds_pathological_tool_args` -> 1 passed
- `./scripts/test.sh tests/test_webui_gateway_chat_backend.py -k gateway_tool_progress_event_bounds_pathological_args` -> 1 passed
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_run_journal_routes.py tests/test_webui_gateway_chat_backend.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py tests/test_anchor_scene_persistence.py` -> 205 passed
- `node --check static/sessions.js && node --check static/messages.js && node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD && git diff --check` -> pass

Readback after push: PR head is `bd6b082561e0fc2f5fdb078c34e478c0df90fcfd`, base is `02aa91ba6ece4454dcb580f960e5b516b3a91111`, mergeable. Checks restarted on the new head.

franksong2702:
Refreshed the PR branch onto current `origin/master` in merge commit `4b72fd0e1` (`735ba1c4` base) after `master` moved again. The merge was clean; upstream touched `static/sessions.js`, `static/boot.js`, extension session hook tests, and `CHANGELOG.md`.

Local verification after this refresh:
- `./scripts/test.sh tests/test_inflight_stream_reuse.py tests/test_run_journal_routes.py tests/test_webui_gateway_chat_backend.py tests/test_cross_session_message_load_isolation.py tests/test_live_to_final_anchor_visible_order.py tests/test_anchor_scene_persistence.py` -> 205 passed
- `node --check static/sessions.js && node --check static/messages.js && node --check static/ui.js` -> pass
- `python3 scripts/ruff_lint.py --diff origin/master` -> pass
- `git diff --check origin/master..HEAD && git diff --check` -> pass

GitHub readback after push: PR head is `4b72fd0e146734cf06cb376568ca5d4383349928`, base is `735ba1c4d3c009b57e3a1bb98f6d2dc40dfb8a61`, mergeable. Checks restarted on the refreshed head, so I am leaving this monitored as waiting on CI until they finish.

nesquena-hermes:
Shipped in **exp-v0.52.103**. Thanks @franksong2702 — solid 3-round convergence! 🎉 The final gate verified 247-char tool args survive start/completion-only/completion-enrichment/anchor-projection/equal-seq durable selection, the new `bound_run_journal_snapshot_args` caps pathological payloads (50k items → 64, 1MB string → 8206 chars) while keeping realistic values intact, and every recovery invariant (cursor round-trip, cross-session isolation, tie-break) holds. Fixes #6196.

PR #6199, opened 2026-07-18
fix(models): prevent bare-id picker revert when provider hint is empty (#6195)

### LINKED ISSUE #6195: bug(models): picked model reverts to the default group when its bare id collides with the default model and the provider hint is dropped (e.g. after a 429 provider switch)
## Bug description

When a model is selected in the composer picker whose **bare id collides with the profile's default model** (the same bare name lives in multiple provider groups), the picker can **revert to the default provider's model** if the provider hint is dropped during a re-render — most visibly after a **failed (HTTP 429) turn** followed by switching provider and re-sending. The user ends up back on the original/default model even though they deliberately switched to a different provider to escape the rate limit.

Reported in Discord `#report-bugs` with a screen recording (GLM-5.2 case). Filed after a full reproduction + Codex root-cause pass.

## Environment / trigger (from the report)

- The model `glm-5.2` is served by **many provider groups at once**: Z.AI/GLM (`@zai:glm-5.2`, the active/default), API.TOKENROUTER.COM (`z-ai/glm-5.2-free`, `z-ai/glm-5.2`), plus Alibaba / ModelScope / others all advertising a `glm-5.2`.
- A turn failed with `HTTP 429: Weekly/Monthly Limit Exhausted` (fallback model also exhausted).
- User opened the composer model picker and switched to `glm-5.2-free` (API.TOKENROUTER.COM) to route around the limit. Composer chip correctly showed `glm-5.2-free`.
- User re-sent the prompt. A new turn started — and the **composer chip reverted from `glm-5.2-free` back to `GLM 5.2`** (the zai/default one).

This is a narrower, still-live descendant of the picked-model-reverts family: #3737/#3739 (explicit-pick consumed after first send), #5924/#5949 (recovery re-send honors newly-picked model), and #5979/#5989 (duplicate model rows across proxy groups). Related recovery-flow umbrella: #2761.

## Root cause (reproduced + Codex-confirmed)

The revert is **NOT** a server-side model/provider drop, and it does **not** reproduce on the normal recovery path when the provider hint is intact. I built an isolated repro (multi-group duplicate `glm-5.2` config + mock, exp-v0.52.96) and drove the exact sequence in a real browser; a Codex trace independently converged on the same conclusion.

**What is solid (no revert):**
- Picking the tokenrouter variant stores it namespace-qualified (`@custom:tokenrouter:z-ai/glm-5.2-free`, `model_provider=custom:tokenrouter`). Send routes correctly (mock received `z-ai/glm-5.2-free` on the tokenrouter endpoint), the session persists the qualified id, and a **page reload restores it correctly**.
- The 429 error path appends/saves the error marker but **never clears `model`/`model_provider`** (`api/streaming.py` ~10386).
- Picker selection updates `S.session.model_provider` synchronously and records the pending marker; `/retry` and edit-resubmit re-arm the deliberate pick (`_deliberateSessionModelPick` / `_reArmRecoveryPick`, `static/ui.js` ~3041/3070). With `model_provider` present, server resolution returns the pair unchanged (`api/routes.py` ~6985) and an explicit pick bypasses foreign-provider repair (~6185).

**The residual bug — the exact condition that DOES revert:**
The client-side restore matcher `_findModelInDropdown(modelId, sel, preferredProviderId)` (`static/ui.js` ~3107) resolves to the **wrong provider group** only when **both** of these hold:
1. the model is stored/looked-up as a **bare id** whose name is identical to the default model (`glm-5.2`, not the qualified `z-ai/glm-5.2-free`), **and**
2. the **`preferredProviderId` hint is empty** at match time.

Deterministic proof (extracted the real `_findModelInDropdown` + `_getOptionProviderId` + `_modelPickerOptionIdentity` and ran them against the reproduced dropdown):

```
find("@custom:tokenrouter:z-ai/glm-5.2-free", prov="custom:tokenrouter") -> "@custom:tokenrouter:z-ai/glm-5.2-free"   # OK
find("z-ai/glm-5.2-free",                     prov="")                   -> "@custom:tokenrouter:z-ai/glm-5.2-free"   # OK (unique name survives even w/o hint)
find("glm-5.2-free",                          prov="")                   -> "@custom:tokenrouter:z-ai/glm-5.2-free"   # OK
find("glm-5.2",                               prov="custom:tokenrouter") -> "@custom:tokenrouter:z-ai/glm-5.2"        # OK (hint disambiguates)
find("glm-5.2",                               prov="")                   -> "z-ai/glm-5.2"   # ❌ REVERT: snaps to the DEFAULT (zai) group
```

So the pick only "reverts" when it degrades to the **bare default-colliding name AND loses its provider hint** — then the normalized match falls to the first group in the dropdown, which is the default/zai group. Any uniquely-named variant (`glm-5.2-free`) survives even without a hint. The server-side default-repair arm (`api/routes.py` ~7078) is gated to bare `gpt`/`claude`/`gemini` families, so `glm-5.2` cannot hit it — this is purely the frontend restore matcher.

## How the pick degrades to the bare colliding name

The remaining open question (needs the reporter's config to pin exactly): the tokenrouter free model must reach the restore matcher as **bare `glm-5.2`** rather than `z-ai/glm-5.2-free`. Candidate paths: (a) the user's proxy advertises the free tier with a display label `glm-5.2-free` but a bare underlying id `glm-5.2`; (b) picker dedup/normalization (#5989 territory) collapses the row to its bare form; (c) a re-render where `preferredProviderId`/`S.session.model_provider` is momentarily unavailable (cold catalog, live-fetch-in-flight) feeds `_findModelInDropdown` with an empty hint. The visible chip in the recording (`glm-5.2-free`, flat, no `z-ai/` prefix) is consistent with a bare-id underlying model.

## Steps to reproduce (deterministic, isolated)

1. Configure ≥2 custom providers that both advertise a model with the **same bare id** as the profile default (e.g. default `glm-5.2` under provider A; provider B also serving a bare `glm-5.2`).
2. In the picker, select provider B's `glm-5.2`.
3. Force the restore matcher to run without a provider hint (session re-render on a cold catalog / after an errored turn where the stored model is the bare id).
4. Observe the picker snap back to provider A's (default) `glm-5.2`.

Minimal JS-level repro (the isolated function, no server): `_findModelInDropdown("glm-5.2", sel, "")` returns the first group's row instead of the provider-scoped one.

## Expected behavior

A deliberately-picked model must stay on the picked provider across errored-turn re-sends and session re-renders, even when its bare id collides with the default model's bare id and the provider hint is temporarily unavailable.

## Suggested fix shape (narrow, consistent with #5924/#3739/#6131)

Two independent hardening points, either of which closes the residual:
1. **Never look up a duplicate-bare model without its provider hint.** In the session-render/restore path (`static/ui.js` ~10350, `_applyModelToDropdown` → `_findModelInDropdown`), pass `S.session.model_provider` (and fall back to the stored `@provider:` qualifier) so a bare `glm-5.2` is always disambiguated; treat an empty hint on a duplicate-bare id as "no match" rather than snapping to the first group.
2. **Persist the picked model in qualified form.** Ensure a pick from a non-default group is stored as `@provider:model` (or slash-qualified) so it can never degrade to the bare default-colliding name in the first place.

Do NOT broaden into a picker rewrite — the existing explicit-pick machinery is correct for the qualified case; this is specifically the bare-id + lost-hint edge.

## Verification notes

- Reproduced end-to-end in an isolated browser against a seeded multi-group config on exp-v0.52.96; the qualified-pick path does NOT revert (send, reload, and clean re-send all preserve the tokenrouter pick).
- Codex trace (GPT-5.6-sol, xhigh) independently confirmed: no server-side provider drop; the only revert-capable site is the bare-name + missing-hint match. 13 focused tests across the #5924 recovery and #6131 duplicate-selection suites pass.
- The precise degradation path (why the pick reaches the matcher bare) needs the reporter's `config.yaml` slices for the two providers serving `glm-5.2` — specifically whether the free tier's underlying id is bare `glm-5.2` or namespaced.

webtecnica:
## Summary

When a user deliberately picks a model whose **bare id collides with the profile's default model** (e.g. `glm-5.2` exists under both the default Z.AI provider and a tokenrouter provider), the picker can **revert to the default group** on re-render if the provider hint is momentarily unavailable.

## Root cause

`_findModelInDropdown(modelId, sel, preferredProviderId)` (`static/ui.js:3107`) has a **normalized-match fallback** (Stage 2) that returns the **first** dropdown option whose normalized name matches — regardless of provider. When `preferredProviderId` is empty and the bare id is ambiguous across provider groups, this snaps to the default group (which is always first in the dropdown).

From the maintainer's deterministic proof:
```
find("glm-5.2", prov="custom:tokenrouter") → "@custom:tokenrouter:z-ai/glm-5.2"  # OK
find("glm-5.2", prov="")                  → "z-ai/glm-5.2"  # ❌ snaps to DEFAULT group
```

## Fix

In the normalized-match step (Stage 2), when the bare id matches **multiple provider groups** AND no provider hint (`preferredProviderId`, `explicitProvider`) is available AND the model is not slash-qualified (`no /` in name), return `null` instead of snapping to the first group. The caller (`_applyModelToDropdown`) falls through to `_ensureModelOptionInDropdown`, which injects the correct provider-scoped option.

## Scope

- `static/ui.js`: +12/-2 lines in `_findModelInDropdown()`
- No change to the existing provider-aware match (Stage 1) — qualified models with hints continue to work as before
- Uniquely-named bare models (no collision) continue to match via the same normalized step
- Only triggers when: bare id + multiple normalized matches + no hint

Closes #6195


nesquena-hermes:
## Summary

Read the full `_findModelInDropdown()` at `static/ui.js:3107` on `cron-pr-6199`'s HEAD (`549b3dadd`) against `origin/master`, plus the two callers (`_applyModelToDropdown:3218`, `_ensureModelOptionInDropdown:3252`) and the session-render restore path at `static/ui.js:10360-10395`. The guard is scoped exactly to the condition #6195 pinned, and the variable references are all in scope. One caveat below on whether it *restores* the picked provider vs. merely stops the affirmative snap.

## Code reference

The new Stage-2 guard (HEAD):

```js
const exact=opts.find(o=>norm(o)===target);
if(exact){
  const normMatches=options.filter(o=>norm(o.value)===target);
  if(normMatches.length>1 && !preferred && !explicitProvider && !rawModel.includes('/')){
    return null;  // ambiguous bare id — caller must inject the correct option
  }
  return exact;
}
```

`preferred`, `explicitProvider`, `rawModel`, and `options` are all defined earlier in the function (`3134-3141`), so the guard is sound. It fires *only* on `find("glm-5.2", prov="")` and leaves every OK row in the issue's proof table (`glm-5.2` + hint, `glm-5.2-free`, `@custom:...`) matching exactly as before — the `normMatches.length>1` + `!preferred` + `!explicitProvider` + no-slash conjunction is precisely the "bare default-colliding id, hint lost" edge.

## The gap worth confirming before merge

Returning `null` routes the session-render caller into `_ensureModelOptionInDropdown()` (`ui.js:10376`), whose provider derivation is:

```js
const provider=preferredProviderId||(badge&&badge.provider)||_providerFromModelValue(modelId)||'';
```
(`ui.js:3254`)

In the exact bug scenario the hint is genuinely empty (`preferredProviderId` null), and `_providerFromModelValue("glm-5.2")` returns `''` (no `@`, `2791`). So the injected option's provider falls back to `window._configuredModelBadges["glm-5.2"].provider` — a **bare-keyed global that holds one provider per bare id**. For a duplicate-bare model that badge very likely resolves to whichever group won dedup, which is often the default (zai) group. Net: this PR reliably stops the *affirmative* snap onto the first group's existing `z-ai/glm-5.2` `<option>`, but on a truly absent hint it does not by itself guarantee the tokenrouter provider is restored — it can convert "wrong provider" into "provider-less / badge-default provider."

That's consistent with #6195's own framing: fix #1 (this PR) treats empty-hint-on-duplicate-bare as "no match," while fix #2 (persist the pick in `@provider:model` qualified form upstream so it never degrades to the bare id) is what actually preserves the provider end-to-end. This PR is the correct, narrow half of #1; it just doesn't close the degradation path the issue flagged as the still-open question.

## Recommendation

- Land as-is for the affirmative-snap regression, but say so explicitly — it's a mitigation, not the full fix, unless the reporter's config keys `_configuredModelBadges["glm-5.2"]` to the non-default provider.
- Add a regression test. There's a ready Node-driver harness: `tests/test_issue6131_provider_aware_model_selection.py` already extracts `_findModelInDropdown` from `ui.js` and drives it. A `test_issue6195_*` mirroring the issue's proof table (`find("glm-5.2","") -> null`, plus the OK cases) would lock the guard against future normalized-match refactors. The PR currently touches only `static/ui.js` with no test.

## Test plan

Two custom providers both advertising bare `glm-5.2` (one the profile default); pick the non-default; force a re-render with `S.session.model_provider` empty. Assert (a) the picker no longer shows the default group's `z-ai/glm-5.2`, and (b) — the part to verify manually — what provider the injected option carries when the badge map is default-keyed.

nesquena-hermes:
Thanks for chasing this — the reverting behavior in #6195 is real, but I gated the fix on the current head and it **doesn't engage for the reported scenario**, so I'm bouncing with a precise root-cause + fix-spec.

### What I verified (empirical, not a read)

I extracted the real `_findModelInDropdown` from your branch's `static/ui.js` and ran it under Node against the exact #6195 setup — a select where `glm-5.2` exists as a literal option value under **both** the default Z.AI group (first in DOM) and a tokenrouter group, calling with `preferredProviderId=''`:

```
bare glm-5.2, no hint  =>  returns 'glm-5.2'  via  STEP-0  (opts.includes(modelId) early return)
```

The guard you added is in **Step 2 (normalized match)**, but for a bare id that is a *literal duplicate option value*, the function returns much earlier at **Step 0**:

```js
if(opts.includes(modelId)){
  const exactOpt=options.find(o=>o.value===modelId);   // = the FIRST glm-5.2 = default group
  const exactProv=...; const pref='';                   // no hint
  if(!pref || ...) return modelId;                       // <-- this is the reverting return
}
```

`sel.value='glm-5.2'` in the caller then snaps to the first (default) `<option>` because the `#6131` provider-restore block only runs `if(preferredProvider)`. Step 2 is never reached, so `normMatches.length>1 → return null` can't fire. I couldn't construct a realistic input that reaches your branch (a bare id that is *not* a literal option value but normalizes to ≥2 options — the hyphen→dot normalization makes those diverge too).

### Why a client-only guard can't fully fix it

Even moved to Step 0, `return null` alone won't disambiguate: with **no provider hint**, the client genuinely cannot know which of two identical-value options the user meant — a DOM `<select>` can't distinguish them, and the `#6131` restore is gated on `preferredProvider`. Your own PR body nails the real signal: *"qualified persist survives, bare persist reverts."*

### Root cause + fix-spec (the durable fix)

The reverting re-render path calls `_applyModelToDropdown` / `_findModelInDropdown` with an **empty** `preferredProviderId`, dropping the provider the user selected. Fix the caller, not the matcher:

1. **Thread the provider through the reverting re-render.** Find the refresh/re-render site that passes an empty hint (a `syncModelChip` / `_refreshOpenModelDropdown` rebuild or a session-model reapply) and pass the stored provider (`S.session.model_provider`, or the picker's selected `data-provider`). With `preferred` populated, **Step 1** provider-aware match resolves the correct group and the `#6131` restore selects the right option — no ambiguity guard needed.
2. **If you keep a matcher guard as a backstop**, put it at **Step 0** (`opts.includes(modelId)` + multiple same-value options across providers + `!pref`), and pair it with caller logic that preserves the *currently-selected* option instead of appending a duplicate.
3. **Add a regression test** in the style of `tests/test_issue6131_provider_aware_model_selection.py` / `test_issue3737_explicit_pick_client_wiring.py`: two options with the same bare value under different providers, simulate the re-render, assert the deliberately-picked (non-default) provider option stays selected. This PR currently ships no test — that's why the dead-branch placement wasn't caught.

Happy to re-gate the moment you re-push. The bug is worth fixing; it just needs to land on the code path that actually returns for the bare-duplicate case.

nesquena-hermes:
## Gate: verified + test-isolation flakes fixed

Gated the bare-id picker-revert guard and verified it end-to-end with a node harness that extracts the real `_findModelInDropdown` and runs the #6195 scenario:

- **The fix works and is surgical.** For a bare model id that collides across provider groups with an *empty* provider hint (`find("glm-5.2", "")`), master snapped to the default group (`z-ai/glm-5.2`); the guard now returns `null` so the caller injects a provider-scoped option instead of asserting the wrong provider. The hinted path (`find("glm-5.2", "custom:tokenrouter")` → the tokenrouter option), uniquely-named bare ids, and the Stage-0 exact-match path are all **unchanged** — verified against master vs head.
- **Added a revert-sensitive regression test** (`test_issue6195_bare_id_ambiguous_no_revert.py`) that locks in all four invariants; it fails on master's `ui.js` and passes on this branch.

While gating I ran the full suite and traced three pre-existing cross-file test-isolation leaks that flaked under sharded ordering, and fixed each at its polluter's teardown (not with a broad global change, which I verified trades one flake for others):

- `test_issue2720` leaked `config._cfg_path` (monkeypatched `_get_config_path` to a temp dir, never restored) → a later `get_available_models()` reloaded config from disk over its in-memory cfg. Restored `_cfg_path` in the helper teardown.
- `test_provider_sort_order` leaked a `providers` block into the shared `config.cfg` via `reload_config()` → a phantom provider group in `test_model_resolver`. Added a scoped `cfg` snapshot/restore fixture.
- `test_agent_runtime_revision_guard` failed only on machines with an editable `hermes_agent` install (the `__editable__` meta-path finder shadowed the test's synthetic `run_agent`). The probe subprocess now drops that finder before importing (no-op on CI).

**All 5 CI shards green (13,388 tests).** Rebased onto current master. Thanks @webtecnica — clean, well-scoped fix.

nesquena-hermes:
Shipped in **exp-v0.52.112** (experimental channel). Thanks @webtecnica — a deliberately-picked model that shares a bare id with the default provider no longer reverts to the default group on re-render. (Also landed a regression test + fixes for three cross-file test-isolation leaks found while gating.) 🎉

